PR #2837, opened 2017-10-26
[jvm-packages] Rank is not a classification task

ebernhardson:
Training a model with the experimental rank:ndcg objective incorrectly
returns a Classification model. Adjust the classification check to
not recognize rank:* objectives as classification.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -194,8 +194,8 @@ object XGBoost extends Serializable {
     val objective = params.getOrElse("objective", params.getOrElse("obj_type", null))
     objective != null && {
       val objStr = objective.toString
-      objStr == "classification" || (!objStr.startsWith("reg:") && objStr != "count:poisson" &&
-        objStr != "rank:pairwise")
+      objStr != "regression" && !objStr.startsWith("reg:") && objStr != "count:poisson" &&
+        !objStr.startsWith("rank:")
     }
   }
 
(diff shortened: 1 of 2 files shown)


jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -195,7 +195,7 @@ object XGBoost extends Serializable {
     objective != null && {
       val objStr = objective.toString
       objStr == "classification" || (!objStr.startsWith("reg:") && objStr != "count:poisson" &&
-        objStr != "rank:pairwise")
+        !objStr.startsWith("rank:pairwise"))
```
superbobry: Sorry, I'm missing the difference here. Are there any other built-in rank objectives?

ebernhardson: rank:ndcg [@n], which is still experimental (doesn't work with hist algo, for one) but gives significantly better results on my datasets

superbobry: Okay, thanks for the explanation. Am I right that you, in fact, want to check `!objStr.startsWith("rank:")`?

ebernhardson: doh, yes. I'm certain i built and tested with that in my analytics network, but i guess when i redid the patch locally to upload i completely spaced out. New patch incoming.

ebernhardson:
patch updated for startsWith("rank:")

superbobry:
Could you add a unit test please?

ebernhardson:
Patch updated with tests. This also identified that obj_type -> regression was incorrectly identified as a classification task, so the function was updated to pass the tests.

PR #2532, opened 2017-07-19
[jvm-packages] Added baseMargin to ml.dmlc.xgboost4j.LabeledPoint

superbobry:
Previously, in order to use baseMargin the user should've provided an RDD with *exactly* the same partitioning as the training RDD. This is hard/impossible to achieve in the general case. This PR integrates baseMargin into `LabeledPoint` making the partitioning requirement irrelevant. See individual commit messages for motivation on each change.

PR #2841, opened 2017-10-28
[jvm-packages] Repair spark model eval

ebernhardson:
In the refactor to add base margins, #2532, all of the labels were lost
when creating the dmatrix. This became obvious as metrics like ndcg
always returned 1.0 regardless of the results.


ebernhardson:
Somewhat related, although I didn't look too closely because I don't use base margins, but it seems if you did train a model using base margins then you will also need to use them when evaluating the model or the results will be incorrect? It could be they are making it through and i simply didn't notice though.

superbobry:
> [...] if you did train a model using base margins then you will also need to use them when evaluating the model or the results will be incorrect.

I think in that case you'd need access to the model which produced the base margins, otherwise, the output would be miscalibrated. Wdyt @CodingCat?

PR #2846, opened 2017-10-30
xgbfi C-API

Far0n:
- C++ implementation of [xgbfi ](https://github.com/Far0n/xgbfi) 
- XGBoosterGetFeatureInteractions C-API

### DIFF
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -17,6 +17,7 @@
 #include "../common/math.h"
 #include "../common/io.h"
 #include "../common/group_data.h"
+#include "../analysis/xgbfi.h"
 
 namespace xgboost {
 // booster wrapper for backward compatible reason.
@@ -943,6 +944,32 @@ XGB_DLL int XGBoosterSaveRabitCheckpoint(BoosterHandle handle) {
   } else {
     rabit::CheckPoint(bst->learner());
   }
+  API_END();
+}
+
+XGB_DLL int XGBoosterGetFeatureInteractions(BoosterHandle handle,
+                                            int max_fi_depth,
+                                            int max_tree_depth,
+                                            int max_deepening,
+                                            int ntrees,
+                                            const char *fmap,
+                                            xgboost::bst_ulong *out_len,
+                                            const char ***out_fi_array) {
+  API_BEGIN();
+  Booster* bst = static_cast<Booster*>(handle);
+  std::vector<std::string>& str_vecs = XGBAPIThreadLocalStore::Get()->ret_vec_str;
+  std::vector<const char*>& charp_vecs = XGBAPIThreadLocalStore::Get()->ret_vec_charp;
+  str_vecs = xgbfi::GetFeatureInteractions(*(bst->learner()),
+                                           max_fi_depth,
+                                           max_tree_depth,
+                                           max_deepening,
+                                           ntrees, fmap);
+  charp_vecs.resize(str_vecs.size());
+  for (size_t i = 0; i < str_vecs.size(); ++i) {
+    charp_vecs[i] = str_vecs[i].c_str();
+  }
+  *out_fi_array = dmlc::BeginPtr(charp_vecs);
+  *out_len = static_cast<xgboost::bst_ulong>(charp_vecs.size());
   API_END();
 }
 
--- a/amalgamation/xgboost-all0.cc
+++ b/amalgamation/xgboost-all0.cc
@@ -53,6 +53,9 @@
 #include "../src/tree/updater_histmaker.cc"
 #include "../src/tree/updater_skmaker.cc"
 
+//xgbfi
+#include "../src/analysis/xgbfi.cc"
+ 
 // global
 #include "../src/learner.cc"
 #include "../src/logging.cc"
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -578,4 +578,24 @@ XGB_DLL int XGBoosterLoadRabitCheckpoint(
  */
 XGB_DLL int XGBoosterSaveRabitCheckpoint(BoosterHandle handle);
 
+/*!
+* \brief Xgbfi: XGBoost Feature Interactions & Importance
+* \param handle BoosterHandle
+* \param max_fi_depth upper bound for depth of interactions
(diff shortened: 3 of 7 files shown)


src/c_api/c_api.cc
```diff
@@ -946,5 +947,45 @@ XGB_DLL int XGBoosterSaveRabitCheckpoint(BoosterHandle handle) {
...
+      LOG(CONSOLE) << "Warning: unable to read feature map: \"" << fmap_file << "\", "
+        "feature names wont be mapped";
+    }
+  }
+  Booster* bst = static_cast<Booster*>(handle);
+  xgbfi::XgbModelDump dump = bst->learner()->DumpModel(fmap, true, "text");
+  xgbfi::XgbModel model = xgbfi::XgbModelParser::GetXgbModelFromDump(dump, ntrees);
+  xgbfi::FeatureInteractions fi = model.GetFeatureInteractions(max_fi_depth,
+                                                               max_tree_depth,
+                                                               max_deepening);
```
RAMitchell: I think you don't need all of these classes here. Is it possible to have one single function (not method) GetFeatureInteractions() here? That way you do not need to define these classes in xgbfi.h, you can instead define them in the .cc file. This is cleaner as these classes are only specific to xgbfi and are not needed in other code.

Far0n: sure, will do.

Far0n:
Thank you for the timely review @RAMitchell! I'm wanted to ship unit tests along with the Python bindings in a 2nd commit to this PR. I do not plan to add R bindings here.

A feature interaction has the structure of
```
struct {
  vector<string> feature_names;  // varying size
  double gain;
  double fscore;
  ... // or
  vector<float> stats;
}
```
One reason for strings is, that I want to avoid feature id ->feature names mapping in python and that this csv-like format is easily converted into a pandas DataFrame.

Far0n:
@RAMitchell Could you please review the changes.

I don't know what to do with the following lint-error, because I think the method get_feature_interactions is at the right place:
```
 python-package/xgboost/core.py:706: refactor (R0904, too-many-public-methods, Booster) Too many public methods (21/20)
```

Thank you!

RAMitchell:
Looks good. My inclination would be to disable the lint-error. I agree with you that the method is in the correct place. For the R build you probably need to add the new c++ file to the amalgamation. One of the Python failures is also due to flake8 warnings so also address these.

@khotilov can I get a second review?

khotilov:
I'll try to take a better look this weekend, but so far I've had just under an hour to look through the code and to write some notes and questions.

@Far0n My first question is about understanding your intentions better: do you see this addition as sort of a "plugin" to xgboost that you would like to have here, and you would like to keep it as similar as reasonably possible in structure and function to your original tool (which you want to keep developing)? Or would you prefer this to be absorbed/integrated into xgboost? I'm asking because I see that your code is currently very loosely coupled with the core (just by passing a text dump of a model).

- The code has little in common with the stuff under src/tree/, and not only based on the code itself but also based on the purpose (keep in mind that the src/tree/ is about *building a tree*). Maybe it would make more sense to create an src/analysis/ directory?
- It would be better to have the stuff in StringUtils under common/. Some of it can be reused.
- Note that instead of having multiple separate small classes, xgboost's c++ style tends to use struct for fully public classes, and to nest the internal auxiliary classes/data structures within the classes that utilize them.
- I'm not sure why there must be a *static* XgbModelParser::xgb_node_lists_ container.
- A fair amount of code could be saved by dealing with GBTreeModel directly instead of parsing a model dump.
- But if the model dump parsing is to stay, why not to have a functionality to restore a GBTreeModel from it?
- An interface to filter specific trees would be useful (e.g., to determine interactions that are important to a specific class in a multiclass setting).

Again, those are just my observations so far, which are more for a discussion than for action.

Far0n:
@khotilov 

For obv reasons, I don't like this dump parsing as well, but my journey starting within c_api hit a private wall in booster -> learner. So how to get the model without changing core code? I like the idea to keep it loosely coupled, because I have high hopes regarding sth like treelite to provide it for any tree based model. GBTreeModel seems also very optimzed for internal use.

khotilov:
Yes, the tree model is well hidden, for the reason of having the same interface in all GradientBoosters. However, we can change the core code for a good cause. Doing various tree model analysis tasks is a good cause. I did it previously for my own hacking experiments, but we'll think about some clean option.

The treelite is more for deployment than analysis. There's dmlc/treelite/issues/5 about a possibility of doing some transformations though... Maybe @hcho3 can comment.

hcho3:
@Far0n Yes, treelite is currently geared toward deployment. I am actually debating whether there should be a separate project for model analysis and transformation. The workflow would be:

```
   Model (xgboost, lightgbm, scikit-learn etc.)
=> Common model schema
=> Transformed model (pruning etc)
=> Optimize for deployment
=> Compiled prediction library, for deployment
```

The original plan was to have transformation step as part of treelite, but it may be beneficial to have a separate project for transformation and analysis. Let me get back to you on this decision.

hcho3:
Also, it would be nice if we can set a time to chat about what kind of model analysis you'd like to do. Can we schedule a time for a Google hangout?

hcho3:
Also, it would be nice if we can set a time to chat about what kind of model analysis you'd like to do. Can we schedule a time for a Google hangout?

RAMitchell:
@hcho3 if you are able to come to h2oworld @Far0n will also be there.

hcho3:
@RAMitchell That's great! I really want to come now. I'll let you know by end of this week whether I'd be able to attend.

Far0n:
make it happen @hcho3 :)

hcho3:
@Far0n I am able to attend h2oworld. I will see you next month.

Far0n:
awesome @hcho3, looking forward to!

khotilov:
So, what was the outcome of you all meeting in December?

hcho3:
@khotilov @Far0n I am currently working to separate out the front-end of treelite to enable model analysis and editing. I've been swamped by other duties as a maintainer, but I'll try to get it done by next month.

Far0n:
Thank you @hcho3. I'm looking forward to it. Take your time though. :)

Issue #2843, opened 2017-10-28
[jvm-packages] Get number of rounds from XGBoost Spark Cross Validation


xiaozhouwang:
First of all, I want to say that XGBoost Spark is greatly developed and thank you so much for all xgboost contributors' hard work!!

Now back to the question I want to ask.:)
So I am following the example here 
https://docs.databricks.com/_static/notebooks/xgboost.html

Specifically the "Tune Model using MLlib Cross Validation" part to do cross validation. So in the end, 
we get something like this:
```
cvModel.bestModel.extractParamMap
res4: org.apache.spark.ml.param.ParamMap =
{
	XGBoostRegressionModel_f3425b374335-alpha: 0.0,
	XGBoostRegressionModel_f3425b374335-booster: gbtree,
	XGBoostRegressionModel_f3425b374335-colsample_bylevel: 1.0,
	XGBoostRegressionModel_f3425b374335-colsample_bytree: 1.0,
	XGBoostRegressionModel_f3425b374335-eta: 0.6,
	XGBoostRegressionModel_f3425b374335-featuresCol: features,
	XGBoostRegressionModel_f3425b374335-gamma: 0.0,
	XGBoostRegressionModel_f3425b374335-grow_policy: depthwise,
	XGBoostRegressionModel_f3425b374335-labelCol: label,
	XGBoostRegressionModel_f3425b374335-lambda: 1.0,
	XGBoostRegressionModel_f3425b374335-lambda_bias: 0.0,
	XGBoostRegressionModel_f3425b374335-max_bin: 16,
	XGBoostRegressionModel_f3425b374335-max_delta_step: 0.0,
	XGBoostRegressionModel_f3425b374335-max_depth: 7,
	XGBoostRegressionModel_f3425b374335-min_child_weight: 1.0,
	XGBoostRegressionModel_f3425b374335-normalize_type: tree,
	XGBoostRegressionModel_f3425b374335-predictionCol: prediction,
	XGBoostRegressionModel_f3425b374335-rate_drop: 0.0,
	XGBoostRegressionModel_f3425b374335-sample_type: uniform,
	XGBoostRegressionModel_f3425b374335-scale_pos_weight: 1.0,
	XGBoostRegressionModel_f3425b374335-sketch_eps: 0.03,
	XGBoostRegressionModel_f3425b374335-skip_drop: 0.0,
	XGBoostRegressionModel_f3425b374335-subsample: 1.0,
[...]

PR #2850, opened 2017-10-31
[jvm-packages] #2843 Add "round" and "objective" as model parameters

sethah:
This is a simple patch that would address the problem in https://github.com/dmlc/xgboost/issues/2843. A new trait `BoostingAlgorithmParams` is created and holds other boosting-level (or ensemble-level) params (different from "booster" params) that should also be available in the model. `XGBoostEstimator` and `XGBoostModel` then mix in this trait so that the parameters are owned by both.

Obviously, there may be other parameters that should be transferred to the model, so we can discuss that here as well. If there is some design history that I've missed about why these parameters weren't copied originally, please let me know.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/BoostingAlgorithmParams.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/BoostingAlgorithmParams.scala
@@ -1,0 +1,48 @@
+/*
+ Copyright (c) 2014 by Contributors
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+package ml.dmlc.xgboost4j.scala.spark.params
+
+import scala.collection.immutable.HashSet
+
+import org.apache.spark.ml.param._
+
+trait BoostingAlgorithmParams extends Params {
+
+  /**
+   * The number of rounds for boosting
+   */
+  val round = new IntParam(this, "num_round", "The number of rounds for boosting",
+    ParamValidators.gtEq(1))
+
+  /**
+   * Specify the learning task and the corresponding learning objective.
+   * options: reg:linear, reg:logistic, binary:logistic, binary:logitraw, count:poisson,
+   * multi:softmax, multi:softprob, rank:pairwise, reg:gamma. default: reg:linear
+   */
+  val objective = new Param[String](this, "objective", "objective function used for training," +
+    s" options: {${BoostingAlgorithmParams.supportedObjective.mkString(",")}",
+    (value: String) => BoostingAlgorithmParams.supportedObjective.contains(value))
+
+  setDefault(objective -> "reg:linear", round -> 1)
+
+}
+
+private[spark] object BoostingAlgorithmParams {
+  val supportedObjective = HashSet("reg:linear", "reg:logistic", "binary:logistic",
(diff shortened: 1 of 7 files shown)


sethah:
cc @xiaozhouwang

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/BoostingAlgorithmParams.scala
```diff
@@ -0,0 +1,48 @@
...
+  val objective = new Param[String](this, "objective", "objective function used for training," +
+    s" options: {${BoostingAlgorithmParams.supportedObjective.mkString(",")}",
+    (value: String) => BoostingAlgorithmParams.supportedObjective.contains(value))
+
+  setDefault(objective -> "reg:linear", round -> 1)
+
+}
+
+private[spark] object BoostingAlgorithmParams {
+  val supportedObjective = HashSet("reg:linear", "reg:logistic", "binary:logistic",
```
superbobry: Minor nitpick: wdyt about using plural here? `supportedObjectives`. Also, any reason for using `HashSet` instead of `Set`?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/BoostingAlgorithmParams.scala
```diff
@@ -0,0 +1,48 @@
...
+import scala.collection.immutable.HashSet
+
+import org.apache.spark.ml.param._
+
+trait BoostingAlgorithmParams extends Params {
+
+  /**
+   * The number of rounds for boosting
+   */
+  val round = new IntParam(this, "num_round", "The number of rounds for boosting",
```
superbobry: I wish we could rename it to numRound...

I actually have a branch doing that, but the diff is huge.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/BoostingAlgorithmParams.scala
```diff
@@ -0,0 +1,48 @@
...
+ limitations under the License.
+ */
+
+package ml.dmlc.xgboost4j.scala.spark.params
+
+import scala.collection.immutable.HashSet
+
+import org.apache.spark.ml.param._
+
+trait BoostingAlgorithmParams extends Params {
```
superbobry: (Feel free to ignore) I think we can drop `Algorithm` from the name, as `Boosting` already makes it pretty obvious.

sethah: Done.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/BoostingAlgorithmParams.scala
sethah: Yeah, `numIterations` or `numRounds` sounds better to me, but it would break compatibility

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/BoostingAlgorithmParams.scala
sethah: Changed to `supportedObjectives`. Using `HashSet` to follow conventions in `BoosterParams`, for example.

sethah:
Initially, I thought it would be better to just add `LearningTaskParams` to the model as well, but it doesn't work because the parameter `numClasses` conflicts with the class member val `numClasses` for `XGBoostClassificationModel`. So, now `BoostingParams` and `LearningTaskParams` seem to cover common ground. We could probably rethink the organization of the params traits to make it clearer, but perhaps in another PR.

superbobry:
Hey @sethah do you plan to continue working on this PR or should we merge it as-is?

lbustelo:
Are there any work-arounds at the moment? Can I dump the model and look at that array to tell the number of rounds?

RAMitchell:
Closing as inactive for more than six months

PR #2866, opened 2017-11-08
Improved gpu_hist_experimental algorithm

RAMitchell:
- Implement colsampling, subsampling for gpu_hist_experimental

 - Optimised multi-GPU implementation for gpu_hist_experimental

 - Make nccl optional

 - Add Volta architecture flag

 - Optimise RegLossObj

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ msvc_use_static_runtime()
 
 # Options
 option(USE_CUDA  "Build with GPU acceleration") 
+option(USE_NCCL "Build using NCCL for multi-GPU. Also requires USE_CUDA") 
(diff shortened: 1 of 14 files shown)


pseudotensor:
Hi @RAMitchell , can we also go back to double for the accumulator.  Arno and I tried to do it (bc25ee6762f1839e85812755d64df4de7b043872 and d93b3edf2e9c8e63140d8ae375c568ab1162d57a) but it's not working correctly as far as we can tell.  We get larger errors and some tests of the scale invariance of the error don't work.  Feel free to use our code as a starting point, but you probably have a better idea.

pseudotensor:
The changes/improvements look awesome.

Issue #2867, opened 2017-11-09
Pandas unit tests failing


RAMitchell:
The pandas continuous integration tests are currently failing with the following message:

```
ERROR: test_pandas (test_with_pandas.TestPandas)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/dmlc/xgboost/tests/python/test_with_pandas.py", line 45, in test_pandas
    dm = xgb.DMatrix(df, label=pd.Series([1, 2]))
  File "/home/travis/build/dmlc/xgboost/python-package/xgboost/core.py", line 265, in __init__
    feature_types)
  File "/home/travis/build/dmlc/xgboost/python-package/xgboost/core.py", line 190, in _maybe_pandas_data
    for i in data.columns
  File "/home/travis/build/dmlc/xgboost/python-package/xgboost/core.py", line 190, in <listcomp>
    for i in data.columns
TypeError: 'int' object is not iterable
======================================================================
ERROR: test_with_sklearn.test_feature_importances
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/miniconda/envs/python3/lib/python3.5/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/dmlc/xgboost/tests/python/test_with_sklearn.py", line 91, in test_feature_importances
    xgb_model = xgb.XGBClassifier(seed=0).fit(X, y)
  File "/home/travis/build/dmlc/xgboost/python-package/xgboost/sklearn.py", line 501, in fit
    missing=self.missing, nthread=self.n_jobs)
  File "/home/travis/build/dmlc/xgboost/python-package/xgboost/core.py", line 265, in __init__
    feature_types)
  File "/home/travis/build/dmlc/xgboost/python-package/xgboost/core.py", line 190, in _maybe_pandas_data
    for i in data.columns
  File "/home/travis/build/dmlc/xgboost/python-package/xgboost/core.py", line 190, in <listcomp>
    for i in data.columns
TypeError: 'int' object is not iterable
```

This may be due to a software update unrelated to XGBoost. 

I am not familiar with Pandas so perhaps someone can suggest a fix.

PR #2872, opened 2017-11-10
Fix MultiIndex detection (breaks for latest pandas==0.21.0).

RAMitchell:
Fix for #2867


marcelflygare:
Just noticed, that it breaks for pandas==0.22 on importing MultiIndex.

PR #2878, opened 2017-11-14
AVX gradients

RAMitchell:
Accelerate basic gradient calculations using AVX instructions. Maybe optionally turned on via cmake. Speedups are in the range of 6-7x for logistic regression calculation on an AWS machine.

Added basic AVX wrapper class to simplify usage of AVX in future.


superbobry:
Have you considered using [boost.simd](https://github.com/NumScale/boost.simd)? It allows to abstract away the specific architecture and support SSE2/AVX/AVX2 etc.

RAMitchell:
I have considered it but boost is very monolithic. I think these changes are very lightweight and AVX is almost universally supported.

src/common/avx_helpers.h
```diff
@@ -0,0 +1,104 @@
...
+inline Float8 operator/(Float8 lhs, const Float8& rhs) {
+  lhs /= rhs;
+  return lhs;
+}
+
+// https://codingforspeed.com/using-faster-exponential-approximation/
+inline Float8 Exp4096(Float8 x) {
+  Float8 fraction(1.0 / 4096.0);
+  Float8 ones(1.0f);
+  x.x = _mm256_fmadd_ps(x.x, fraction.x, ones.x);  // x = 1.0 + x / 4096
```
khotilov: Note that the `_mm256_fmadd_ps` is technically not a part of AVX, but is from the FMA3 instruction set which is available starting from Haswell arch, while AVX is available starting from Sandy Bridge. I would suggest replacing it with more conservative
```
  x *= fraction;
  x += ones;
```
I've checked that it makes no difference in performance.

khotilov:
@RAMitchell I'm curious: does the gradient calculation become a significant contributor to runtime in some situations? I'm not really happy about the increasing messiness of the code, but if it helps, then why not.

@superbobry So far this project managed to stay clean from the dependency on boost. Besides, boost.simd would pull in practically all the boost. While there are some other, even header-only, I believe, simd libraries, I lean towards @RAMitchell's lightweight approach.

I did some testing of the AVX sigmoid and exp code using a quick Rcpp wrapper
https://gist.github.com/khotilov/3dcf8ce42631880bbaf8d03d457d0eb2

```R
curve(approximate_sigmoid(x) - 1/(1 + exp(-x)), -12, 12, n = 1000); grid()
```
image

The exp4096 mostly underestimates the exp:
```R
curve(log(exp4096(x)) - x, -9, 9, n = 1000); grid()
```
image
```R
# verify the monotonicity
x <- seq(from=-12, to=12, length.out = 8*100)
min(diff(approximate_sigmoid(x)))
## 0
min(diff(exp4096(x)))
## 1.84265e-07

library(microbenchmark)
x <- runif(1000000)
microbenchmark(exp4096(x), exp(x))
## Unit: milliseconds
##        expr       min        lq      mean    median        uq       max neval cld
##  exp4096(x)  5.636702  5.842971  8.801399  6.394068  9.333585 133.17423   100  a 
##      exp(x) 50.410999 51.238929 53.008278 51.859162 54.684212  75.79283   100   b
microbenchmark(approximate_sigmoid(x), 1/(1+exp(x)))
## Unit: milliseconds
##                    expr       min        lq      mean    median        uq       max neval cld
##  approximate_sigmoid(x)  5.719951  5.929213  7.354263  6.180529  9.400727  14.57115   100  a 
##          1/(1 + exp(x)) 54.512440 55.593252 59.545110 56.124107 58.769688 187.05631   100   b
```

RAMitchell:
@khotilov I'm currently working on optimising for p3 aws instances. 50% of the runtime was  gradient calculations for binary classification when I started optimising. I got it down from 50 seconds to about 2.7 seconds now.

I agree with you that its a little messy at the moment. I might be able to come up with something better.

Thanks for the great charts!

tqchen:
SIMD is great, to implement it cleanly. perhaps we can use some packet templates that defers to scalr version when the simd is not available. The packet math in mshadow might be a good reference

tqchen:
I see float8 is already doing that in some sens, so only a fallback implementionn is needed

RAMitchell:
I have considerably improved the accuracy of the exp to around 1e-6. The sigmoid should be around 1e-7.

sigmoid
exp

Also the code is now far less messy in src\objective\regression_obj.cc.

PR #2791, opened 2017-10-16
[GPU-plugin] added GPU objective function

teju85:
Objective functions can now also be run on GPUs.
Currently, reg:linear, reg:logistic, binary:logistic and binary:logitraw functions have been ported.
To select their GPU counterparts, prefix 'gpu:' in their names. (Refer to parameter.md for more info)

PR #2880, opened 2017-11-15
Added GPU versions of multiple objective functions.

teju85:
Hi,
1. reg:linear, reg:logistic, binary:logistic and binary:logitraw have GPU versions now
2. All comments from the previously closed PR #2791 have been addressed here.
3. Our testing shows ~9-10% end-to-end training time improvement using the benchmark.py (on a Quadro GP100 workstation)

Some observations:
Most of the time in the objective function evaluation is actually spent in copying data between host and GPU. We need to update the interface of objective function evaluation to help eliminate these copies. This should further reduce training time.

Let us know in case you need any more inputs from us.

Regards,
Thejaswi

### DIFF
--- a/doc/parameter.md
+++ b/doc/parameter.md
@@ -165,6 +165,9 @@ Specify the learning task and the corresponding learning objective. The objectiv
   - "reg:logistic" --logistic regression
   - "binary:logistic" --logistic regression for binary classification, output probability
   - "binary:logitraw" --logistic regression for binary classification, output score before logistic transformation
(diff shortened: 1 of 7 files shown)


RAMitchell:
I prefer to use AVX for this for now as it does not require any extension of the API.

I agree that we need interfaces capable of working with device memory to efficiently use GPUs for some parts of XGBoost. I'm not sure exactly what this would look like currently.

Perhaps we can revisit this after developing such an interface.

PR #2836, opened 2017-10-26
[jvm-packages] Exposed train-time evaluation metrics

superbobry:
They are accessible via `XGBoostModel.summary`. The summary is not serialized with the model and is only available after the training.

PR #2885, opened 2017-11-17
[jvm-packages] Fixed a subtle bug in train/test split

superbobry:
Iterator.partition (naturally) assumes that the predicate is deterministic
but this is not the case for

    r.nextDouble() <= trainTestRatio

therefore sometimes the DMatrix(...) call got a NoSuchElementException
and crashed the JVM due to lack of exception handling in
XGBoost4jCallbackDataIterNext.


superbobry:
Merged into #2836.

PR #2888, opened 2017-11-19
[R] maintenance Nov 2017; SHAP plots

khotilov:
Fixes for a few outstanding issues.

SHAP contribution dependency plots (an alternative to partial dependency plots). So far only 1D dependency plots are implemented, and no ggplot version. E.g., for a specific class in multiclass classification:
```R
set.seed(123)
x <- as.matrix(iris[, -5])
is.na(x[sample(nrow(x) * 4, 30)]) <- TRUE # introduce some missing values - will be shown in purple
mbst <- xgboost(data = x, label = as.numeric(iris$Species) - 1,
                nrounds = 20, max_depth = 2, eta = 0.1, subsample = .5, nthread = 2,
                objective = "multi:softprob", num_class = 3, verbose = 1)
# for class 1 (versicolor)
xgb.plot.shap(x, model = mbst, trees = seq(from=1, by=3, length.out=20), target_class = 1,
                top_n = 4, n_col = 2, col = rgb(0, 0, 1, 0.5), pch = 16, pch_NA = 17)
```
image


khotilov:
@pommedeterresautee @hetong007 could one of you take a look, please?

R-package/R/xgb.plot.shap.R
```diff
@@ -0,0 +1,208 @@
...
+#' The similarity to partial dependency plots is that they also give an idea for how feature values
+#' affect predictions. However, in partial dependency plots, we usually see marginal dependencies
+#' of model prediction on feature value, while SHAP contribution dependency plots display the estimated
+#' contributions of a feature to model prediction for each individual case.
+#' 
+#' When \code{plot_loess = TRUE} is set, feature values are rounded to 3 significant digits and
+#' weighted LOESS is computed and plotted, where weights are the numbers of data points
+#' at each rounded value.
+#' 
+#' Note: SHAP contributions are shown on the scale of model margin. E.g., for a logistic binomial objective,
```
hetong007: Would you mind to add the original paper `A Unified Approach to Interpreting Model Predictions` as a reference under `@references`?

khotilov: Yes, it looks like now it's a NIPS paper!

@slundberg we were previously referring to https://arxiv.org/abs/1706.06060. Should the "A Unified Approch ..." NIPS paper supersede it, or does it make sense to keep references to both?

slundberg: Great work! Both would make since because the newer arXiv contains the fast algorithm we are using in XGBoost and the original NIPS paper describes the actual values in more detail.

khotilov: Ok, I've added the references to both.

Issue #815, opened 2016-02-10
Specifying Group ID in rank:pairwise Python?


tokestermw:
I am trying to fit a pairwise ranking model. If I'm understanding the algorithm correctly, it needs a group (or query) id so it can rank things relative to the group / query. It looks like the R wrapper can specify the group / query id but I can't seem to find it in the Python wrapper.

Usage in R (has a `group=` option)
https://github.com/dmlc/xgboost/issues/413

Usage in C++ (uses the Microsoft Learning to Rank data which has query id, `qid`)
https://github.com/dmlc/xgboost/blob/master/demo/rank/runexp.sh

I wrote a quick script but it seems like it's treating everything in a single group. How do we train the entire model on multiple groups? Thanks a bunch.

    gbm = XGBRegressor(objective="rank:pairwise")

    X =  np.random.normal(0, 1, 1000).reshape(100, 10)
    y = np.random.randint(0, 5, 100)

    group = [50, 50] # -- where to put this????

    gbm.fit(X, y)

    # should be in reverse order
    print y[gbm.predict(X).argsort()][::-1]

tokestermw:
Looks like there is a `set_group` method in `DMatrix`. https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/core.py#L415

I just need to add this to `sklearn.py`.

PR #2937, opened 2017-12-08
Add grouping to Python sklearn interface

munro:
Adds support to pass through `groups` & `group_sizes` to XGBoost's `sklearn` interface.

~~This is more of a proof of concept, I want to rework the interface so that `groups` is used the way the `sklearn` does things.  With `sklearn` groups should be a group identifier, while for `xgboost` it expects the data to be sorted by group, and then given group sizes.  I was thinking about making a `group_sizes` parameter, and then having a `groups` parameter that works the `sklearn` way.  Let me know if that sounds good!~~

This adds ability to build ranking models through the `sklearn` API, by setting `.fit(X, y, groups=groups)`, which will now:
1) resort the data so groups are next to each other (based on `groups`)
2) on the created dmatrix, it will call `DMatrix.set_group(...)` with the calculated group sizes
3) goes through existing `Booster.predict(...)` code path
4) resorts the results back to what they original were
5) returns back to what called the `sklearn` API

I also passed through `group_sizes` if the programmer would rather use the XGBoost style of handing grouping, instead of how sklearn handles it (which is to assign a group number for every row).

#815 seems related, but it looks like they just jumped down to using `DMatrix` & `train` directly... but I have a bunch of parameter tuning setup so I don't want to rework that code. 😄

### DIFF
--- a/python-package/xgboost/compat.py
+++ b/python-package/xgboost/compat.py
@@ -52,6 +52,7 @@ try:
     from sklearn.base import BaseEstimator
     from sklearn.base import RegressorMixin, ClassifierMixin
     from sklearn.preprocessing import LabelEncoder
+    from sklearn.utils import safe_indexing
     try:
         from sklearn.model_selection import KFold, StratifiedKFold
     except ImportError:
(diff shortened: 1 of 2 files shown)


RAMitchell:
Closing as inactive for more than six months

Issue #2319, opened 2017-05-19
R package failed R CMD check under Windows for arch 'i386'


hetong007:
## Environment info
Operating System: Windows 10

`xgboost` version used: 29289d2

The R `sessionInfo()`:
```
> sessionInfo()
R version 3.4.0 (2017-04-21)
Platform: i386-w64-mingw32/i386 (32-bit)
Running under: Windows >= 8 x64 (build 9200)

Matrix products: default

locale:
[1] LC_COLLATE=Chinese (Simplified)_China.936 
[2] LC_CTYPE=Chinese (Simplified)_China.936   
[3] LC_MONETARY=Chinese (Simplified)_China.936
[4] LC_NUMERIC=C                              
[5] LC_TIME=Chinese (Simplified)_China.936    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] compiler_3.4.0 tools_3.4.0
```

## Steps to reproduce

1. (under Linux/OSX) run `make Rbuild` to generate the tarball
2. Under Windows with R 3.4.0 and both x64/i386 installed, run 
`R CMD check --multiarch xgboost_0.6.4.4.tar.gz`
3. When checking examples and tests for arch `i386`, there will be errors with popup windows (look like corrupted memory). No cause of error in the print-out messages from `R CMD check`.

## What have you tried?

1. Remove all files under `tests/testthat` folder except `test_basic.R`, then simplify its content (no multithreading, no testthat wrapper, no sparse matrix):
```
require(xgboost)

train = matrix(10,5,2)
label = c(0,1,0,0,1)

bst <- xgboost(data = train, label = label, max_depth = 2,
               eta = 1, nthread = 1, nrounds = 2, objective = "binary:logistic")
```
The error remains. So the `xgboost()` call triggered the error. 
2. However, if replace the `xgboost()` call with `xgb.DMatrix`, then there's no error triggered. 
3. Not sure how to reproduce out of the `R CMD check` yet. Tried `gctorture(TRUE)` but with no error.

Could you guys please help to reproduce or shed some light on it? @thirdwing @khotilov

khotilov:
I've just tried: installing the 32 bit components of Rtools34 and R (I need the patched version since the non-patched one has a bug related to reading file time info on my system)
```
> sessionInfo()
R version 3.4.0 Patched (2017-05-08 r72665)
Platform: i386-w64-mingw32/i386 (32-bit)
[...]

PR #2954, opened 2017-12-17
[R] AppVeyor CI for R package

khotilov:
Two additional appveyor jobs that build the R package:
- cmake + VC2015 x64 build + unit tests
- make + Rtools/mingw x64 build + R check
I've disabled the multiarch build for mingw since xgboost is crashing 32 bit R at exit on windows #2319


khotilov:
The following are available as AppVeyor artifacts:
- pre-built binary packages for Win64, built with Rtools (within the "rmingw" job) and VC2015 ("rmsvc" job)
- source package 
- R check log file
- when R check fails, its whole output would be kept in failure.zip

The latter three are only done in the "rmingw" job.

PR #2964, opened 2017-12-21
[jvm-packages] Saving models into a tmp folder every a few rounds

Yunni:
**What's in this PR**
This PR expose a Spark API to save models into a tmp hdfs folder every a few rounds, and when the job restart with the same setting, it will train based on the saved model instead of training from scratch.

**Why do we need this change**
Since the Rabit checkpoint could not be recovered in case of Spark executor lost, saving models every a few rounds could be a good way to improve the reliability of the job.

**How is this PR tested**
The change is tested on agaricus dataset inside the demo, which is also included in the unit test.
A test on larger dataset is done on HIGGS dataset, which is illustrated in the gDoc: https://docs.google.com/document/d/1TTHiePSX4bVMUu-5B0VIogNaHVhUTEbe_4AGRdVmDno/edit

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/TmpBoosterManager.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/TmpBoosterManager.scala
@@ -1,0 +1,138 @@
+/*
+ Copyright (c) 2014 by Contributors
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+package ml.dmlc.xgboost4j.scala.spark
+
+import ml.dmlc.xgboost4j.scala.Booster
+import org.apache.commons.logging.LogFactory
+import org.apache.hadoop.fs.{FileSystem, Path}
+import org.apache.spark.SparkContext
+
+/**
+  * A class which allows user to save temporary boosters every a few rounds. If a previous job
+  * fails, the job can restart training from a saved booster instead of from scratch. This class
+  * provides interface and helper methods for the tmp booster functionality.
+  *
+  * @param sc the sparkContext object
+  * @param boosterTmpPath the hdfs path to store temporary boosters
+  */
+class TmpBoosterManager(sc: SparkContext, boosterTmpPath: String) {
+  private val logger = LogFactory.getLog("XGBoostSpark")
+  private val modelSuffix = ".model"
+
+  private def getPath(version: Int) = {
+    s"$boosterTmpPath/$version$modelSuffix"
+  }
+
+  private def getExistingVersions: Seq[Int] = {
+    val fs = FileSystem.get(sc.hadoopConfiguration)
+    if (boosterTmpPath.isEmpty || !fs.exists(new Path(boosterTmpPath))) {
+      Seq()
+    } else {
+      fs.listStatus(new Path(boosterTmpPath)).map(_.getPath.getName).collect {
+        case fileName if fileName.endsWith(modelSuffix) => fileName.stripSuffix(modelSuffix).toInt
+      }
+    }
+  }
+
+  /**
+    * Load existing booster with the highest version.
+    *
+    * @return the booster with the highest version.
+    */
+  private[spark] def loadBooster: Booster = {
+    val versions = getExistingVersions
+    if (versions.nonEmpty) {
+      val version = versions.max
+      val fullPath = getPath(version)
+      logger.info(s"Start training from previous booster at $fullPath")
+      val model = XGBoost.loadModelFromHadoopFile(fullPath)(sc)
+      model.booster.booster.setVersion(version)
+      model.booster
+    } else {
+      null
+    }
+  }
+
+  /**
+    * Clean up all previous models and save a new model
+    *
+    * @param model the xgboost model to save
+    */
+  private[spark] def updateModel(model: XGBoostModel): Unit = {
+    val fs = FileSystem.get(sc.hadoopConfiguration)
+    val prevModelPaths = getExistingVersions.map(version => new Path(getPath(version)))
+    val fullPath = getPath(model.version)
+    logger.info(s"Saving temporary model with version ${model.version} to $fullPath")
+    model.saveModelAsHadoopFile(fullPath)(sc)
+    prevModelPaths.foreach(path => fs.delete(path, true))
+  }
+
+  /**
+    * Clean up tmp boosters with version higher than or equal to the round.
+    *
+    * @param round the number of rounds in the current training job
+    */
+  private[spark] def cleanUpHigherVersions(round: Int): Unit = {
+    val higherVersions = getExistingVersions.filter(_ / 2 >= round)
+    higherVersions.foreach { version =>
+      val fs = FileSystem.get(sc.hadoopConfiguration)
+      fs.delete(new Path(getPath(version)), true)
+    }
+  }
+
+  /**
+    * Calculate a list of checkpoint rounds to save tmp boosters based on the savingFreq and
+    * total number of rounds for the training. Concretely, the saving rounds start with
+    * prevRounds + savingFreq, and increase by savingFreq in each step until it reaches total
+    * number of rounds
+    *
+    * @param savingFreq the increase on rounds during each step of training
+    * @param round the total number of rounds for the training
+    * @return a seq of integers, each represent the index of round to save the tmp booster
+    */
+  private[spark] def getSavingRounds(savingFreq: Int, round: Int): Seq[Int] = {
+    if (boosterTmpPath.nonEmpty && savingFreq > 0) {
+      val prevRounds = getExistingVersions.map(_ / 2)
+      val firstSavingRound = (0 +: prevRounds).max + savingFreq
+      (firstSavingRound until round by savingFreq) :+ round
+    } else if (savingFreq <= 0) {
+      Seq(round)
+    } else {
+      throw new IllegalArgumentException("parameters \"booster_tmp_path\" should also be set.")
+    }
+  }
+}
+
+object TmpBoosterManager {
+
+  private[spark] def extractParams(params: Map[String, Any]): (String, Int) = {
+    val boosterTmpPath: String = params.get("booster_tmp_path") match {
+      case None => ""
+      case Some(path: String) => path
+      case _ => throw new IllegalArgumentException("parameter \"booster_tmp_path\" must be" +
+        " an instance of String.")
+    }
+
+    val savingFreq: Int = params.get("saving_frequency") match {
+      case None => 0
+      case Some(freq: Int) => freq
+      case _ => throw new IllegalArgumentException("parameter \"saving_frequency\" must be" +
+        " an instance of Int.")
+    }
+    (boosterTmpPath, savingFreq)
+  }
+}
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -331,6 +331,7 @@ object XGBoost extends Serializable {
     val (boosterTmpPath, savingRounds) = extractParamsForCheckpoint(sc, params, round)
 
     val partitionedData = repartitionForTraining(trainingData, nWorkers)
+    var prevBooster = loadPrevBooster(partitionedData.sparkContext, boosterTmpPath)
     // Train for every ${savingRound} rounds and save the partially completed booster
     savingRounds.map {
       savingRound: Int =>
@@ -338,7 +339,6 @@ object XGBoost extends Serializable {
         try {
           val parallelismTracker = new SparkParallelismTracker(sc, timeoutRequestWorkers, nWorkers)
           val overriddenParams = overrideParamsAccordingToTaskCPUs(params, sc)
-          val prevBooster = loadPrevBooster(partitionedData.sparkContext, boosterTmpPath)
           val boostersAndMetrics = buildDistributedBoosters(partitionedData, overriddenParams,
             tracker.getWorkerEnvs, savingRound, obj, eval, useExternalMemory, missing, prevBooster)
           val sparkJobThread = new Thread() {
@@ -359,6 +359,7 @@ object XGBoost extends Serializable {
               params.getOrElse("num_class", "2").toString.toInt
           }
           if (boosterTmpPath.nonEmpty && savingRound < round) {
+            prevBooster = model.booster
             saveTmpBooster(sc, boosterTmpPath, model)
           }
           model
@@ -387,17 +388,16 @@ object XGBoost extends Serializable {
     }
     if (boosterTmpPath.nonEmpty && savingFreq > 0) {
       val prevRound = getHighestVersion(sparkContext, boosterTmpPath) / 2
-      if (prevRound > round) {
+      if (prevRound >= round) {
         val fs = FileSystem.get(sparkContext.hadoopConfiguration)
         fs.delete(new Path(boosterTmpPath), true)
       }
       val savingRounds: Seq[Int] = (prevRound + savingFreq until round by savingFreq) :+ round
       (boosterTmpPath, savingRounds)
-    } else if (boosterTmpPath.isEmpty && savingFreq <= 0) {
-      ("", Seq(round))
+    } else if (savingFreq <= 0) {
+      (boosterTmpPath, Seq(round))
     } else {
-      throw new IllegalArgumentException("parameters \"booster_tmp_path\" and " +
-        "\"saving_frequency\" must be set simultaneously")
+      throw new IllegalArgumentException("parameters \"booster_tmp_path\" should also be set.")
     }
   }
 
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -20,13 +20,12 @@ import java.io.File
 
 import scala.collection.mutable
 import scala.util.Random
-
 import ml.dmlc.xgboost4j.java.{IRabitTracker, Rabit, XGBoostError, RabitTracker => PyRabitTracker}
 import ml.dmlc.xgboost4j.scala.rabit.RabitTracker
 import ml.dmlc.xgboost4j.scala.{XGBoost => SXGBoost, _}
 import ml.dmlc.xgboost4j.{LabeledPoint => XGBLabeledPoint}
 import org.apache.commons.logging.LogFactory
-import org.apache.hadoop.fs.{FSDataInputStream, Path}
+import org.apache.hadoop.fs.{FSDataInputStream, FileSystem, Path}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.Dataset
 import org.apache.spark.ml.feature.{LabeledPoint => MLLabeledPoint}
@@ -54,6 +53,7 @@ object TrackerConf {
 
 object XGBoost extends Serializable {
   private val logger = LogFactory.getLog("XGBoostSpark")
+  private val modelSuffix = ".model"
 
   private def removeMissingValues(
       denseLabeledPoints: Iterator[XGBLabeledPoint],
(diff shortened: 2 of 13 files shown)


CodingCat:
will look into this, @Yunni

Yunni:
@CodingCat Thank you!

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
```diff
@@ -446,14 +451,15 @@ int loadRabitCheckpoint() throws XGBoostError {
    */
   void saveRabitCheckpoint() throws XGBoostError {
     XGBoostJNI.checkCall(XGBoostJNI.XGBoosterSaveRabitCheckpoint(this.handle));
+    version += 1;
```
CodingCat: we might also need to update the comments to indicate that this method update version number

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -104,20 +117,24 @@ public static Booster train(
     }
 
     //initialize booster
-    Booster booster = new Booster(params, allMats);
-
-    int version = booster.loadRabitCheckpoint();
+    if (booster == null) {
+      booster = new Booster(params, allMats);
+      booster.loadRabitCheckpoint();
+    } else {
```
CodingCat: add comments to specify this is the branch for training based on an existing model

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -76,6 +76,19 @@ public static Booster train(
           IObjective obj,
           IEvaluation eval,
           int earlyStoppingRound) throws XGBoostError {
+    return train(dtrain, params, round, watches, metrics, obj, eval, earlyStoppingRound, null);
+  }
+
+  public static Booster train(
```
CodingCat: would you help to add comments in these train() methods?

jvm-packages/xgboost4j/src/main/scala/ml/dmlc/xgboost4j/scala/XGBoost.scala
```diff
@@ -52,13 +52,14 @@ object XGBoost {
       metrics: Array[Array[Float]] = null,
       obj: ObjectiveTrait = null,
       eval: EvalTrait = null,
-      earlyStoppingRound: Int = 0): Booster = {
+      earlyStoppingRound: Int = 0,
+      booster: JBooster = null): Booster = {
```
CodingCat: update comments

jvm-packages/xgboost4j/src/test/java/ml/dmlc/xgboost4j/java/BoosterImplTest.java
```diff
@@ -347,4 +344,53 @@ public void testCV() throws XGBoostError {
...
+    // Save tempBooster to bytestream and load back
+    ByteArrayInputStream in = new ByteArrayInputStream(tempBooster.toByteArray());
+    tempBooster = XGBoost.loadModel(in);
+    in.close();
+
+    // Continue training using tempBooster
+    round = 4;
+    Booster booster2 = XGBoost.train(trainMat, paramMap, round, watches, null, null, null, 0, tempBooster);
+    float booster2error = eval.eval(booster2.predict(testMat, true, 0), testMat);
+    TestCase.assertTrue(booster1error == booster2error);
```
CodingCat: did we test on larger datasets?

CodingCat:
would you extend the changes to Spark part, e.g. use a broadcast variable to pass models to executors and train based on that?

Yunni:
Will extend the changes to Spark and test it on larger dataset

jvm-packages/xgboost4j/src/main/scala/ml/dmlc/xgboost4j/scala/XGBoost.scala
Yunni: Updated

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
Yunni: Done.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
Yunni: Done.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
```diff
@@ -446,14 +451,15 @@ int loadRabitCheckpoint() throws XGBoostError {
...
   }
 
   /**
    * Internal initialization function.
```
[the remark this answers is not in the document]
Yunni: Actually I was wrong, this creates a new booster and should not be called for existing Booster.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
Yunni: Done.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
```diff
@@ -433,10 +438,10 @@ public void saveModel(OutputStream out) throws XGBoostError, IOException {
    * @return the stored version number of the checkpoint.
    * @throws XGBoostError
    */
-  int loadRabitCheckpoint() throws XGBoostError {
+  void loadRabitCheckpoint() throws XGBoostError {
     int[] out = new int[1];
     XGBoostJNI.checkCall(XGBoostJNI.XGBoosterLoadRabitCheckpoint(this.handle, out));
-    return out[0];
+    version = out[0];
```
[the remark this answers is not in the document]
Yunni: Done.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -331,34 +333,66 @@ object XGBoost extends Serializable {
...
+    }
+
+    val savingFreq: Int = params.get("saving_frequency") match {
+      case None => 0
+      case Some(freq: Int) => freq
+      case _ => throw new IllegalArgumentException("parameter \"saving_frequency\" must be" +
+        " an instance of Int.")
     }
+
+    val savingRounds: Seq[Int] = if (hdfsTmpPath == null || savingFreq <= 0){
```
CodingCat: we should not swallow the user's input error here, i.e. we should deal with hdfsTmpPath == null as a wrong param in the helper function above 

and personally, I think the following way to regulate these parameters is more natural to the user

1. if the user sets a > 0 value for savingFreq, that means they want to the checkpoint 
2. if the sets a > 0 value for savingFreq, they must specify hdfsTmpPath, otherwise, we should stop training due to the missing parameter

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -405,6 +435,29 @@ object XGBoost extends Serializable {
     xgBoostModel.setPredictionCol(predCol)
   }
 
+  private def loadPrevBooster(sparkContext: SparkContext, hdfsTmpPath: String):
```
CodingCat: we may want to do some cleanup when loading the latest checkpoint, e.g. use an async thread to delete all old files. If failed, just put a warning

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -405,6 +435,29 @@ object XGBoost extends Serializable {
...
+      null
+    } else {
+      val fs = FileSystem.get(sparkContext.hadoopConfiguration)
+      val tempPath = new Path(hdfsTmpPath)
+      val versions = fs.listStatus(tempPath).map(_.getPath.getName).collect {
+        case fileName if fileName.endsWith(modelSuffix) => fileName.stripSuffix(modelSuffix).toInt
+      }
+      if (versions.nonEmpty) {
+        val version = versions.max
+        val fullPath = s"$hdfsTmpPath/$version$modelSuffix"
```
CodingCat: we should not duplicate the path pattern in two places to facilitate the further modification

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -331,34 +333,66 @@ object XGBoost extends Serializable {
...
+          if (hdfsTmpPath != null && savingRound < round) {
+            val fullPath = s"$hdfsTmpPath/${model.version}$modelSuffix"
+            logger.info(s"Saving temporary model to $fullPath")
+            model.saveModelAsHadoopFile(fullPath)(sc)
+          }
+          model
+      } finally {
+        tracker.stop()
+      }
+    }.last
```
CodingCat: we may want to clean up all temporary files when we finish the last round training,

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -405,6 +439,28 @@ object XGBoost extends Serializable {
...
 
+  private[spark] def loadPrevBooster(sparkContext: SparkContext, hdfsTmpPath: String):
+      Booster = {
+    val fs = FileSystem.get(sparkContext.hadoopConfiguration)
+    if (hdfsTmpPath == null || !fs.exists(new Path(hdfsTmpPath))) {
+      null
+    } else {
+      val versions = fs.listStatus(new Path(hdfsTmpPath)).map(_.getPath.getName).collect {
+        case fileName if fileName.endsWith(modelSuffix) => fileName.stripSuffix(modelSuffix).toInt
+      }
```
CodingCat: in some cases (though rare in XGBoost case), when we have a lot of files under the same directory, this listStatus can be very slow...

so we may want to clean those old files for each time we successfully load the latest checkpoint

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostGeneralSuite.scala
```diff
@@ -335,4 +336,35 @@ class XGBoostGeneralSuite extends FunSuite with PerTest {
...
+    val tmpPath = Files.createTempDirectory("model1").toAbsolutePath.toString
+    val paramMap = List("eta" -> "1", "max_depth" -> 2, "silent" -> "1",
+      "objective" -> "binary:logistic", "hdfs_tmp_path" -> tmpPath,
+      "saving_frequency" -> 2).toMap
+    val finalModel = XGBoost.trainWithRDD(trainingRDD, paramMap, round = 5,
+      nWorkers = numWorkers)
+    def error(model: XGBoostModel): Float = eval.eval(
+      model.booster.predict(testSetDMatrix, outPutMargin = true), testSetDMatrix)
+    val model4 = XGBoost.loadModelFromHadoopFile(s"$tmpPath/4.model")
+    val model8 = XGBoost.loadModelFromHadoopFile(s"$tmpPath/8.model")
```
CodingCat: why is it 4 and 8?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -106,13 +106,15 @@ object XGBoost extends Serializable {
       obj: ObjectiveTrait,
       eval: EvalTrait,
       useExternalMemory: Boolean,
-      missing: Float): RDD[(Booster, Map[String, Array[Float]])] = {
+      missing: Float,
+      hdfsTmpPath: String): RDD[(Booster, Map[String, Array[Float]])] = {
```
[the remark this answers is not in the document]
superbobry: Could also be `checkpointPath`?

Yunni:
I did some AWS load testing with a larger dataset and put it in gDoc. I need to update it after this round of CR comments. Will put it in the PR description shortly.

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostGeneralSuite.scala
Yunni: Because the version increment twice for each round. So if the temp models are saved in round 2 and round 4, the model will be named as "4.model" and "8.model".
See here for more details: https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java#L112

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
Yunni: Good Point.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
Yunni: I think clean old files during saving makes more sense, because we actually can keep exactly 1 model for job recovering.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
Yunni: Cleaned during temp model saving.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -331,34 +333,62 @@ object XGBoost extends Serializable {
...
+          logger.info(s"Rabit returns with exit code $trackerReturnVal")
+          val model = postTrackerReturnProcessing(trackerReturnVal, boostersAndMetrics,
+            sparkJobThread, isClsTask)
+          if (isClsTask){
+            model.asInstanceOf[XGBoostClassificationModel].numOfClasses =
+              params.getOrElse("num_class", "2").toString.toInt
+          }
+          if (hdfsTmpPath != null && savingRound < round) {
+            val fullPath = s"$hdfsTmpPath/${model.version}$modelSuffix"
+            logger.info(s"Saving temporary model to $fullPath")
```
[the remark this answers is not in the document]
Yunni: Done.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -331,34 +333,66 @@ object XGBoost extends Serializable {
...
+      case Some(freq: Int) => freq
+      case _ => throw new IllegalArgumentException("parameter \"saving_frequency\" must be" +
+        " an instance of Int.")
     }
+
+    val savingRounds: Seq[Int] = if (hdfsTmpPath == null || savingFreq <= 0){
+      Seq(round)
+    } else {
+      (savingFreq until round by savingFreq) :+ round
+    }
```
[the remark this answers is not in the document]
Yunni: Done.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -331,34 +333,62 @@ object XGBoost extends Serializable {
...
+      case _ => throw new IllegalArgumentException("parameter \"saving_frequency\" must be" +
+        " an instance of Int.")
     }
+
+    val savingRounds: Seq[Int] = if (savingFreq <= 0){
+      Seq(round)
+    } else {
+      (savingFreq until round by savingFreq) :+ round
+    }
+    savingRounds.map {
```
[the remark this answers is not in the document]
Yunni: Done.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
Yunni: Done.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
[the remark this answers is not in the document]
Yunni: I prefer `boosterTmpPath` as the checkpoint might be confused with Rabit checkpoints.

CodingCat:
@Yunni test failed...

Yunni:
@CodingCat @superbobry I have pasted the gDoc link in the description. Feel free to comment.
@CodingCat That might be a transient error. Let's see if it fails again.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -405,6 +458,44 @@ object XGBoost extends Serializable {
...
+        versions.max
+      } else {
+        0
+      }
+    }
+  }
+
+  private def saveTmpBooster(sc: SparkContext, boosterTmpPath: String, model: XGBoostModel) = {
+    val fs = FileSystem.get(sc.hadoopConfiguration)
+    fs.delete(new Path(boosterTmpPath), true)
```
CodingCat: sorry, missed this one, we should delete only after we successfully save the latest model, otherwise, we can lose everything

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -331,33 +327,90 @@ object XGBoost extends Serializable {
...
+
+    val savingFreq: Int = params.get("saving_frequency") match {
+      case None => 0
+      case Some(freq: Int) => freq
+      case _ => throw new IllegalArgumentException("parameter \"saving_frequency\" must be" +
+        " an instance of Int.")
+    }
+    if (boosterTmpPath.nonEmpty && savingFreq > 0) {
+      val prevRound = getHighestVersion(sparkContext, boosterTmpPath) / 2
+      if (prevRound >= round) {
```
CodingCat: delete something in a method named "extractParamsForCheckpoint" seems scary

is this branch just for the case that user want to train x more iterations after they have got a model which is completed after y iterations, where y > x?

if this is the case, did we save a model after successfully training for y iterations? (in https://github.com/dmlc/xgboost/pull/2964/files#diff-095250777d030169ae83818b37ab44b1R361) looks like no

and also, does it mean the user has to name their model as x.model (where x is an integer).....

CodingCat: we may want to make this more user-friendly

Yunni: This is actually something else. When user reduces the number of rounds of training, we want to clean up all models with higher version. Please take a look at `TmpBoosterManager.cleanUpHigherVersions`

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
Yunni: Good point. I have made fix and renamed the method into `TmpBoosterManager.updateModel`

Yunni:
@CodingCat I understand the previous code is confusing. As a result, I moved all helper methods to a new class `TmpBoosterManager` along with documentations and unit tests. Let me know if it helps.

CodingCat:
Looks like something similar to the last time failed again

left some comments

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/TmpBoosterManager.scala
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+  * A class which allows user to save temporary boosters every a few rounds. If a previous job
+  * fails, the job can restart training from a saved booster instead of from scratch. This class
+  * provides interface and helper methods for the tmp booster functionality.
+  *
+  * @param sc the sparkContext object
+  * @param boosterTmpPath the hdfs path to store temporary boosters
+  */
+class TmpBoosterManager(sc: SparkContext, boosterTmpPath: String) {
```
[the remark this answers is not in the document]
Yunni: Done.

I though checkpoint might be confused with Rabit checkpoint. But since of both you thinks it makes sense, let's go this way.

CodingCat:
LGTM, merged thanks

PR #2985, opened 2017-12-27
delete duplicated code in python-package

csgwma:
delete duplicated code in python-package.
The line 142 and line 158 in the python-package/xgboost/sklearn.py are the same code, delete line 142.


csgwma:
What' more, please check The Note "**kwargs is unsupported by Sklearn.". I think **kwargs is supported by sklearn now!

PR #2990, opened 2017-12-29
[jvm-packages] fix numAliveCores in SparkParallelismTracker when WebUI is disabled

CodingCat:
the current tests succeed but with strange NPE like the one below

image

The reason is that when we disabled webui of spark in test, the current code returns null as url of rest end point leading to a NPE inside jackson

This PR fixes this


CodingCat:
@Yunni Just found this class was implemented by you, would you please have a review on this

PR #2994, opened 2017-12-31
[R] fix for the 32 bit windows issue

khotilov:
@hetong007 @thirdwing 
Fix for #2319 . Win32 R (or the specific mingw version in Rtools) doesn't seem to handle the destruction of `thread_local`-based storage properly, so I've disabled it only for this specific platform.

The dual 64+32 architecture check and build of a binary package (as an appveyor "artifact") for windows are now enabled in CI.


thirdwing:
Great! Thank you for handling this.

PR #2999, opened 2018-01-02
[jvm-packages] Reduce the workload of a unit test

Yunni:
I found this unit test is unstable on Mac. By reducing the number of rounds, the stability significantly improves. This might be the root cause of Jenkins build failures.


Yunni:
@CodingCat @superbobry Sorry I think I have involved an unstable test. Would you plz take a look at this PR? Thanks!

CodingCat:
do we know the root cause? reducing training rounds from 5 to 3 will be more stable looks like a partial fix....

Yunni:
@CodingCat The segment error happens only on Mac system. I think it might caused by high memory usage but I don't have ways to verify it. Do you have any suggestions here?

CodingCat:
on your mac or travis's?

Yunni:
@CodingCat Both. The memory consumption is around 500MB and the probability of segfault is around 20%

CodingCat:
I can understand in travis containers the flaky test might be caused by high memory usage, but in your mac book, 500 MB usage in the native memory can lead to this is not that convincing.......

Yunni:
@CodingCat That's a very good point. I have narrowed down the problem to be related to Object serialization on Mac. Will send out another PR to fix the issue shortly.

PR #3000, opened 2018-01-02
[jvm-packages] Unify the serde of Booster

Yunni:
It turns out there is a transient error of SIGSEGV when running unit tests on MacOS. After some investigation, I found that the issue is related to serialization of Booster. Currently Booster has different ways to serialize/deserialize as Serializable and KryoSerializable. This PR unifies the way of serialization and should resolve the transient error in unit tests.


superbobry:
Hmm, do you have a theory on why it failed the tests?

Yunni:
@superbobry My theory is that `byte[] bytes = (byte[])in.readObject();` does not always read the correct number of bytes, which caused the following JNI call on L513 to fail.

CodingCat:
> My theory is that byte[] bytes = (byte[])in.readObject(); does not always read the correct number of bytes, which caused the following JNI call on L513 to fail.

it's not true, any class implementing serializable interface should be read correctly by this method

PR #3005, opened 2018-01-04
[jvm-packages] Prevent dispose being called on unfinalized JBooster

Yunni:
There are Scala Booster (SBooster) and Java Booster(JBooster). JBooster is a class variable of SBooster. When SBooster finalize, it calls JBooster.dispose(). However, JBooster could still be alive and referenced at that time. This will cause any following JNI calls crash with SIGSEGV.


Yunni:
Further Info:
As the Jenkins log report the issue is inside `XGBoosterGetModelRaw`, I suspect `handle` in `XGBoosterGetModelRaw` of `c_api.cc`  could be a nullptr. So I added the following line to the Java code: 
```
    System.out.println(handle);
    XGBoostJNI.checkCall(XGBoostJNI.XGBoosterGetModelRaw(this.handle, bytes));
```
In the cases it crashed, it printed out the following log.
```
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.67.4, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
140634665597936
140634665646384
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.67.4, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
140634665646384
140634665646384
140634665646384
140634665646384
140634665646384
140634650185712
140634666116896
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.67.4, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
140634666116896
140634666116896
140634666116896
140634666116896
140634666116896
140634665714528
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.67.4, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
140634666612656
140634666612656
140634666612656
140634666612656
140634666612656
0
#
# A fatal error has been detected by the Java Runtime Environment:
#
...
```

The only place where `handle` is set to 0 is inside `Booster.dispose()` so I think `dispose` is likely to be the issue. Do you buy this theory @CodingCat @superbobry ?

CodingCat:
you are suggesting that someone uses java booster after scala booster has been reclaimed .....

would you give more details about how this happened?

Yunni:
I am not familiar with how objects are finalized in Spark local tests. But we can verify this theory by the following:

- In `JBooster.toByteArray()`:
```
  public byte[] toByteArray() throws XGBoostError {
    byte[][] bytes = new byte[1][];
    System.out.println("Calling toByteArray on Booster " + this.hashCode());
    XGBoostJNI.checkCall(XGBoostJNI.XGBoosterGetModelRaw(this.handle, bytes));
    return bytes[0];
  }
```
- In `SBooster.finalize()`:
```
  override def finalize(): Unit = {
    println(s"Dispose Booster ${booster.hashCode()}")
    super.finalize()
    dispose
  }
```
- We got the following log in case of SIGSEGV:
```
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.68.106, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
Calling toByteArray on Booster 2083255801
Calling toByteArray on Booster 1757933443
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.68.106, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
Calling toByteArray on Booster 1757933443
Calling toByteArray on Booster 1757933443
Dispose Booster 1783158786
Dispose Booster 912824903
Dispose Booster 1867337681
Dispose Booster 804867161
Dispose Booster 1628506771
Dispose Booster 818493874
Dispose Booster 2083255801
Dispose Booster 295397222
Calling toByteArray on Booster 1757933443
Calling toByteArray on Booster 1757933443
Calling toByteArray on Booster 1757933443
Calling toByteArray on Booster 1046790284
Calling toByteArray on Booster 103394766
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.68.106, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
Calling toByteArray on Booster 103394766
Calling toByteArray on Booster 103394766
Dispose Booster 231813807
Dispose Booster 930172202
Dispose Booster 449180708
Dispose Booster 1506919064
Dispose Booster 1046790284
Dispose Booster 1683795884
Dispose Booster 2050963654
Dispose Booster 339567342
Dispose Booster 533754929
Dispose Booster 231813807
Dispose Booster 449180708
Dispose Booster 339567342
Dispose Booster 1683795884
Dispose Booster 533754929
Dispose Booster 1046790284
Dispose Booster 2050963654
Dispose Booster 1506919064
Calling toByteArray on Booster 103394766
Calling toByteArray on Booster 103394766
Calling toByteArray on Booster 103394766
Calling toByteArray on Booster 2101519237
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.68.106, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
Calling toByteArray on Booster 2128859746
Calling toByteArray on Booster 2128859746
Calling toByteArray on Booster 2128859746
Calling toByteArray on Booster 2128859746
Calling toByteArray on Booster 2128859746
Dispose Booster 1266732748
Dispose Booster 1358779668
Dispose Booster 43806916
Dispose Booster 576386241
Dispose Booster 461724876
Dispose Booster 1894567389
Dispose Booster 1692807811
Dispose Booster 273815506
Dispose Booster 879401726
Dispose Booster 1525173820
Dispose Booster 1300538821
Dispose Booster 916561256
Dispose Booster 789672186
Dispose Booster 114766155
Dispose Booster 2101519237
Dispose Booster 1377260303
Dispose Booster 968275380
Dispose Booster 916561256
Dispose Booster 879401726
Dispose Booster 1300538821
Dispose Booster 1377260303
Dispose Booster 1525173820
Dispose Booster 114766155
Dispose Booster 789672186
Dispose Booster 2101519237
Calling toByteArray on Booster 1266732748
#
# A fatal error has been detected by the Java Runtime Environment:
#
......
```
We can see JBooster 1266732748 is disposed by the SBooster.finalize() before `JBooster.toByteArray()` is called. @CodingCat Does this make sense?

CodingCat:
it's no guarantee that hashCode is always different for different boosters, 

Dispose Booster 1377260303 appeared for twice, which does not make sense if we assume hashCode is always unique (finalize should not be called for twice in JVM)

can you replace hashCode with Thread.currentThread.name() (or similar name, forgot )? I want to see if any thread (Spark Task) call finalize first and then toByteArray

Yunni:
@CodingCat That's a good catch. I should use `System.identityHashCode` instead, like the following:
```
// Booster.java
  public byte[] toByteArray() throws XGBoostError {
    byte[][] bytes = new byte[1][];
    System.out.println("Calling toByteArray of Booster " + System.identityHashCode(this) + " on thread " + Thread.currentThread().getName());
    XGBoostJNI.checkCall(XGBoostJNI.XGBoosterGetModelRaw(this.handle, bytes));
    return bytes[0];
  }
// Booster.scala
  override def finalize(): Unit = {
    println(s"Dispose Booster ${System.identityHashCode(booster)} on thread ${Thread.currentThread.getName}")
    super.finalize()
    dispose
  }
```
Here is the log:
```
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.68.106, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster 1386918939 on thread Executor task launch worker-7
Calling toByteArray of Booster 1681993012 on thread ScalaTest-run-running-XGBoostGeneralSuite
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.68.106, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster 1681993012 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster 1681993012 on thread dag-scheduler-event-loop
Dispose Booster 1463111286 on thread Finalizer
Dispose Booster 195099247 on thread Finalizer
Dispose Booster 1785185288 on thread Finalizer
Dispose Booster 1386918939 on thread Finalizer
Dispose Booster 1433733252 on thread Finalizer
Dispose Booster 1307587876 on thread Finalizer
Dispose Booster 171595210 on thread Finalizer
Dispose Booster 64422361 on thread Finalizer
Calling toByteArray of Booster 1681993012 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster 1681993012 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster 1681993012 on thread dag-scheduler-event-loop
Calling toByteArray of Booster 998829261 on thread Executor task launch worker-0
Calling toByteArray of Booster 730955696 on thread ScalaTest-run-running-XGBoostGeneralSuite
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.68.106, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster 730955696 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster 730955696 on thread dag-scheduler-event-loop
Dispose Booster 962046526 on thread Finalizer
Dispose Booster 930172202 on thread Finalizer
Dispose Booster 1867472469 on thread Finalizer
Dispose Booster 998829261 on thread Finalizer
Dispose Booster 16379185 on thread Finalizer
Dispose Booster 1514726224 on thread Finalizer
Dispose Booster 1223781314 on thread Finalizer
Dispose Booster 975174892 on thread Finalizer
Dispose Booster 962046526 on thread Finalizer
Dispose Booster 808084661 on thread Finalizer
Dispose Booster 975174892 on thread Finalizer
Dispose Booster 16379185 on thread Finalizer
Dispose Booster 1867472469 on thread Finalizer
Dispose Booster 808084661 on thread Finalizer
Dispose Booster 1514726224 on thread Finalizer
Dispose Booster 1223781314 on thread Finalizer
Dispose Booster 998829261 on thread Finalizer
Calling toByteArray of Booster 730955696 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster 730955696 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster 730955696 on thread dag-scheduler-event-loop
Calling toByteArray of Booster 620053911 on thread Executor task launch worker-1
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.68.106, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster 1188390859 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster 1188390859 on thread dag-scheduler-event-loop
Calling toByteArray of Booster 1188390859 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster 1188390859 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster 1188390859 on thread dag-scheduler-event-loop
Dispose Booster 1191647229 on thread Finalizer
Dispose Booster 172485900 on thread Finalizer
Dispose Booster 722958990 on thread Finalizer
Dispose Booster 1007007984 on thread Finalizer
Dispose Booster 453901713 on thread Finalizer
Dispose Booster 1138779831 on thread Finalizer
Dispose Booster 986474635 on thread Finalizer
Dispose Booster 1504951211 on thread Finalizer
Dispose Booster 968275380 on thread Finalizer
Dispose Booster 620053911 on thread Finalizer
Dispose Booster 1711244882 on thread Finalizer
Dispose Booster 1604911595 on thread Finalizer
Dispose Booster 1882815447 on thread Finalizer
Dispose Booster 620053911 on thread Finalizer
Dispose Booster 1711244882 on thread Finalizer
Dispose Booster 1624223185 on thread Finalizer
Dispose Booster 2061074734 on thread Finalizer
Dispose Booster 734856003 on thread Finalizer
Dispose Booster 1289212445 on thread Finalizer
Dispose Booster 1604911595 on thread Finalizer
Dispose Booster 1882815447 on thread Finalizer
Dispose Booster 1289212445 on thread Finalizer
Dispose Booster 2061074734 on thread Finalizer
Dispose Booster 1624223185 on thread Finalizer
Dispose Booster 734856003 on thread Finalizer
Calling toByteArray of Booster 986474635 on thread Executor task launch worker-0
#
# A fatal error has been detected by the Java Runtime Environment:
#
...
```

CodingCat:
no guarantee on System.identifyHashCode either, it is still a hash code https://stackoverflow.com/questions/4930781/how-do-hashcode-and-identityhashcode-work-at-the-back-end

so, I would suggest you change the code a bit to allow each booster has a name, then you can assign the name to be the stageid + task_id creating it, and then identify if any booster is released before serialized

Yunni:
Sure. Here is how I assign the name to each Booster

```
// XGBoost.buildDistributedBoosters() for new Booster
        val suffix = s"-$stageId-$taskId"
        if (prevBooster != null) {
          prevBooster.booster.name += suffix
        }
        val booster = SXGBoost.train(watches.train, params, round,
          watches.toMap, metrics, obj, eval,
          earlyStoppingRound = numEarlyStoppingRounds, prevBooster, "Booster" + suffix)

// Booster.readObject() for deserialized Booster
  private void readObject(java.io.ObjectInputStream in)
          throws IOException, ClassNotFoundException {
    try {
      this.init(null);
      this.version = in.readInt();
      this.name = "Loaded-" + this.version;
```
And here is the log:
```
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.68.106, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster Booster-0-0 on thread Executor task launch worker-5
Calling toByteArray of Booster Loaded-4 on thread ScalaTest-run-running-XGBoostGeneralSuite
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.24.68.106, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster Loaded-4 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster Loaded-4 on thread dag-scheduler-event-loop
Dispose Booster Booster-0-3 on thread Finalizer
Dispose Booster Booster-0-6 on thread Finalizer
Dispose Booster Booster-0-4 on thread Finalizer
Dispose Booster Booster-0-1 on thread Finalizer
Dispose Booster Booster-0-7 on thread Finalizer
Dispose Booster Booster-0-2 on thread Finalizer
Dispose Booster Booster-0-5 on thread Finalizer
Dispose Booster Booster-0-0 on thread Finalizer
Dispose Booster Loaded-4-2-2 on thread Finalizer
Dispose Booster Loaded-4-2-0 on thread Finalizer
Dispose Booster Loaded-4-2-6 on thread Finalizer
Calling toByteArray of Booster Loaded-4 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster Loaded-4 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster Loaded-4 on thread dag-scheduler-event-loop
Calling toByteArray of Booster Loaded-4-2-0 on thread Executor task launch worker-0
#
# A fatal error has been detected by the Java Runtime Environment:
#
```

CodingCat:
I look at the code and log and got a theory on what was happening, just in case I confuse you, can you help to change code further as

When  prevbooster is not null, we make a deep copy for it and update the copy in the further training?

My theory is basically blaming spark’s context cleaner which reclaimed prevbooster which is shipped as part of a broadcast variable for task running

Yunni:
For sure, here is how I changed it:
```
        val prevBoosterCopy = if (prevBooster != null) {
          // Copy the booster with new name
          val os = new ByteArrayOutputStream()
          prevBooster.saveModel(os)
          os.close()
          val is = new ByteArrayInputStream(os.toByteArray)
          val copy = SXGBoost.loadModel(is, name)
          is.close()
          copy.booster.name = name
          copy
        } else prevBooster
```
And the log:
```
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=10.0.1.13, DMLC_TRACKER_PORT=9092, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster Booster-0-0 on thread Executor task launch worker-7
Calling toByteArray of Booster beforeCopy-4 on thread ScalaTest-run-running-XGBoostGeneralSuite
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=10.0.1.13, DMLC_TRACKER_PORT=9092, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster beforeCopy-4 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster beforeCopy-4 on thread dag-scheduler-event-loop
Dispose Booster Booster-0-5 on thread Finalizer
Dispose Booster Booster-0-0 on thread Finalizer
Dispose Booster Booster-0-4 on thread Finalizer
Dispose Booster Booster-0-2 on thread Finalizer
Dispose Booster Booster-0-6 on thread Finalizer
Dispose Booster Booster-0-1 on thread Finalizer
Dispose Booster Booster-0-7 on thread Finalizer
Dispose Booster Booster-0-3 on thread Finalizer
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-1
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-7
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-0
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-4
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-5
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-2
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-6
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-3
Calling toByteArray of Booster beforeCopy-4 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster beforeCopy-4 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster beforeCopy-4 on thread dag-scheduler-event-loop
Calling toByteArray of Booster Booster-2-0 on thread Executor task launch worker-5
Calling toByteArray of Booster beforeCopy-8 on thread ScalaTest-run-running-XGBoostGeneralSuite
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=10.0.1.13, DMLC_TRACKER_PORT=9092, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster beforeCopy-8 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster beforeCopy-8 on thread dag-scheduler-event-loop
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-2
Dispose Booster beforeCopy-4 on thread Finalizer
Dispose Booster beforeCopy-4 on thread Finalizer
Dispose Booster Booster-2-5 on thread Finalizer
Dispose Booster Booster-2-6 on thread Finalizer
Dispose Booster Booster-2-4 on thread Finalizer
Dispose Booster Booster-2-1 on thread Finalizer
Dispose Booster Booster-2-0 on thread Finalizer
Dispose Booster Booster-2-6 on thread Finalizer
Dispose Booster Booster-2-2 on thread Finalizer
Dispose Booster Booster-2-5 on thread Finalizer
Dispose Booster Booster-2-4 on thread Finalizer
Dispose Booster Booster-2-3 on thread Finalizer
Dispose Booster Booster-2-7 on thread Finalizer
Dispose Booster Booster-2-7 on thread Finalizer
Dispose Booster Booster-2-0 on thread Finalizer
Dispose Booster Booster-2-1 on thread Finalizer
Dispose Booster Booster-2-3 on thread Finalizer
Dispose Booster Booster-2-2 on thread Finalizer
Dispose Booster beforeCopy-4 on thread Finalizer
Dispose Booster beforeCopy-4 on thread Finalizer
Dispose Booster beforeCopy-4 on thread Finalizer
Dispose Booster beforeCopy-4 on thread Finalizer
Dispose Booster beforeCopy-4 on thread Finalizer
Dispose Booster beforeCopy-4 on thread Finalizer
Dispose Booster beforeCopy-4 on thread Finalizer
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-3
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-1
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-0
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-6
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-4
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-5
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-7
Calling toByteArray of Booster beforeCopy-8 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster beforeCopy-8 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster beforeCopy-8 on thread dag-scheduler-event-loop
Calling toByteArray of Booster Booster-4-0 on thread Executor task launch worker-1
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=10.0.1.13, DMLC_TRACKER_PORT=9092, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster  on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster  on thread dag-scheduler-event-loop
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-3
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-2
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-1
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-0
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-7
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-4
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-5
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-6
Calling toByteArray of Booster  on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster  on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster  on thread dag-scheduler-event-loop
Dispose Booster Booster-6-5 on thread Finalizer
Dispose Booster Booster-6-2 on thread Finalizer
Dispose Booster Booster-6-0 on thread Finalizer
Dispose Booster Booster-6-1 on thread Finalizer
Dispose Booster Booster-6-4 on thread Finalizer
Dispose Booster Booster-6-6 on thread Finalizer
Dispose Booster Booster-6-3 on thread Finalizer
Dispose Booster Booster-6-7 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster beforeCopy-8 on thread Finalizer
Dispose Booster Booster-4-0 on thread Finalizer
Dispose Booster Booster-4-7 on thread Finalizer
Dispose Booster Booster-4-1 on thread Finalizer
Dispose Booster Booster-4-5 on thread Finalizer
Dispose Booster Booster-4-3 on thread Finalizer
Dispose Booster Booster-4-7 on thread Finalizer
Dispose Booster Booster-4-6 on thread Finalizer
Dispose Booster Booster-4-0 on thread Finalizer
Dispose Booster Booster-4-4 on thread Finalizer
Dispose Booster Booster-4-2 on thread Finalizer
Dispose Booster Booster-4-4 on thread Finalizer
Dispose Booster Booster-4-5 on thread Finalizer
Dispose Booster Booster-4-2 on thread Finalizer
Dispose Booster Booster-4-6 on thread Finalizer
Dispose Booster Booster-4-1 on thread Finalizer
Dispose Booster Booster-4-3 on thread Finalizer
Calling toByteArray of Booster Booster-6-0 on thread Executor task launch worker-2
#
# A fatal error has been detected by the Java Runtime Environment:
#
```
Some more findings: Are you wondering why the finalizer is called twice? Actually there are 2 SBooster sharing 1 JBooster. With the following code:
```
 // Booster.scala
  override def finalize(): Unit = {
    println(s"Dispose JBooster ${booster.name} in SBooster ${this.hashCode()} on thread ${Thread.currentThread.getName}")
    super.finalize()
    dispose
  }
```
You get the following log:
```
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=10.0.1.13, DMLC_TRACKER_PORT=9092, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster Booster-0-0 on thread Executor task launch worker-2
Calling toByteArray of Booster beforeCopy-4 on thread ScalaTest-run-running-XGBoostGeneralSuite
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=10.0.1.13, DMLC_TRACKER_PORT=9092, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster beforeCopy-4 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster beforeCopy-4 on thread dag-scheduler-event-loop
Dispose JBooster Booster-0-0 in SBooster 626528334 on thread Finalizer
Dispose JBooster Booster-0-6 in SBooster 1996640407 on thread Finalizer
Dispose JBooster Booster-0-7 in SBooster 18422757 on thread Finalizer
Dispose JBooster Booster-0-5 in SBooster 2132254810 on thread Finalizer
Dispose JBooster Booster-0-2 in SBooster 921487462 on thread Finalizer
Dispose JBooster Booster-0-1 in SBooster 1501164087 on thread Finalizer
Dispose JBooster Booster-0-4 in SBooster 1049159513 on thread Finalizer
Dispose JBooster Booster-0-3 in SBooster 472460885 on thread Finalizer
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-6
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-5
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-1
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-0
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-3
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-2
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-4
Calling toByteArray of Booster beforeCopy-4 on thread Executor task launch worker-7
Calling toByteArray of Booster beforeCopy-4 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster beforeCopy-4 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster beforeCopy-4 on thread dag-scheduler-event-loop
Calling toByteArray of Booster Booster-2-0 on thread Executor task launch worker-6
Calling toByteArray of Booster beforeCopy-8 on thread ScalaTest-run-running-XGBoostGeneralSuite
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=10.0.1.13, DMLC_TRACKER_PORT=9092, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster beforeCopy-8 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster beforeCopy-8 on thread dag-scheduler-event-loop
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-5
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-1
Dispose JBooster Booster-2-6 in SBooster 930172202 on thread Finalizer
Dispose JBooster beforeCopy-4 in SBooster 968275380 on thread Finalizer
Dispose JBooster beforeCopy-4 in SBooster 1032398426 on thread Finalizer
Dispose JBooster beforeCopy-4 in SBooster 2022424463 on thread Finalizer
Dispose JBooster beforeCopy-4 in SBooster 1160684969 on thread Finalizer
Dispose JBooster beforeCopy-4 in SBooster 1228514221 on thread Finalizer
Dispose JBooster beforeCopy-4 in SBooster 1031202598 on thread Finalizer
Dispose JBooster beforeCopy-4 in SBooster 413417664 on thread Finalizer
Dispose JBooster beforeCopy-4 in SBooster 681029008 on thread Finalizer
Dispose JBooster Booster-2-5 in SBooster 741820388 on thread Finalizer
Dispose JBooster Booster-2-4 in SBooster 645057171 on thread Finalizer
Dispose JBooster Booster-2-0 in SBooster 485113053 on thread Finalizer
Dispose JBooster Booster-2-3 in SBooster 159709693 on thread Finalizer
Dispose JBooster Booster-2-2 in SBooster 530224974 on thread Finalizer
Dispose JBooster Booster-2-7 in SBooster 797761660 on thread Finalizer
Dispose JBooster Booster-2-1 in SBooster 1649229188 on thread Finalizer
Dispose JBooster Booster-2-0 in SBooster 1462268781 on thread Finalizer
Dispose JBooster Booster-2-1 in SBooster 915719666 on thread Finalizer
Dispose JBooster Booster-2-4 in SBooster 732236722 on thread Finalizer
Dispose JBooster Booster-2-6 in SBooster 808963713 on thread Finalizer
Dispose JBooster Booster-2-3 in SBooster 488937565 on thread Finalizer
Dispose JBooster Booster-2-7 in SBooster 13688294 on thread Finalizer
Dispose JBooster Booster-2-5 in SBooster 492350816 on thread Finalizer
Dispose JBooster Booster-2-2 in SBooster 785655869 on thread Finalizer
Dispose JBooster beforeCopy-4 in SBooster 1277721377 on thread Finalizer
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-6
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-3
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-2
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-0
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-7
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-4
Calling toByteArray of Booster beforeCopy-8 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster beforeCopy-8 on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster beforeCopy-8 on thread dag-scheduler-event-loop
Calling toByteArray of Booster Booster-4-0 on thread Executor task launch worker-2
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=10.0.1.13, DMLC_TRACKER_PORT=9092, DMLC_NUM_WORKER=8}
Calling toByteArray of Booster  on thread ScalaTest-run-running-XGBoostGeneralSuite
Calling toByteArray of Booster  on thread dag-scheduler-event-loop
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-3
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-5
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-6
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-2
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-0
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-1
Calling toByteArray of Booster beforeCopy-8 on thread Executor task launch worker-4
Dispose JBooster beforeCopy-8 in SBooster 130262285 on thread Finalizer
Dispose JBooster beforeCopy-8 in SBooster 1325673097 on thread Finalizer
Dispose JBooster beforeCopy-8 in SBooster 1143906288 on thread Finalizer
Dispose JBooster beforeCopy-8 in SBooster 1263535399 on thread Finalizer
Dispose JBooster beforeCopy-8 in SBooster 1369516685 on thread Finalizer
Dispose JBooster beforeCopy-8 in SBooster 1067389522 on thread Finalizer
Dispose JBooster beforeCopy-8 in SBooster 268360526 on thread Finalizer
Dispose JBooster beforeCopy-8 in SBooster 1145867947 on thread Finalizer
Dispose JBooster beforeCopy-8 in SBooster 484520429 on thread Finalizer
Dispose JBooster Booster-4-0 in SBooster 1529593249 on thread Finalizer
Dispose JBooster Booster-4-5 in SBooster 578169121 on thread Finalizer
Dispose JBooster Booster-4-1 in SBooster 40882706 on thread Finalizer
Dispose JBooster Booster-4-6 in SBooster 387249643 on thread Finalizer
Dispose JBooster Booster-4-7 in SBooster 290755613 on thread Finalizer
Dispose JBooster Booster-4-3 in SBooster 41453623 on thread Finalizer
Dispose JBooster Booster-4-0 in SBooster 933709015 on thread Finalizer
Dispose JBooster Booster-4-7 in SBooster 1009274503 on thread Finalizer
Dispose JBooster Booster-4-3 in SBooster 1935177881 on thread Finalizer
Dispose JBooster Booster-4-4 in SBooster 1740917146 on thread Finalizer
Dispose JBooster Booster-4-6 in SBooster 2018489878 on thread Finalizer
Dispose JBooster Booster-4-5 in SBooster 46564774 on thread Finalizer
Dispose JBooster Booster-4-2 in SBooster 1422318458 on thread Finalizer
Dispose JBooster Booster-4-1 in SBooster 1092658679 on thread Finalizer
Dispose JBooster Booster-4-2 in SBooster 2122003752 on thread Finalizer
Dispose JBooster Booster-4-4 in SBooster 1324367009 on thread Finalizer
.................
```
You see JBooster-4-0 is first disposed when SBooster 1529593249 is finalized and second time when SBooster 933709015 is finalized. As the hash code is different, they must be different SBooster.
`

CodingCat:
if you make a copy to for each booster, isn't it eliminating the sharing?

CodingCat:
would you upload your current code?

Yunni:
@CodingCat I find out why: A new `SBooster` is always created at the end of the training, even if the `JBooster` inside is already referenced by `prevBooster`.
https://github.com/dmlc/xgboost/pull/3005/files#diff-578d7adb219f0724c983cdd6cc45a39aL69

CodingCat:
it's a nice finding!!! I just validated that, in Executor side, since the task binary has to be deserialized `Per Task`, ContextCleaner should not mess up stuffs

We should clean the changes in this PR now, see my inlined comments

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
```diff
@@ -432,6 +432,7 @@ public void setVersion(int version) {
    */
   public byte[] toByteArray() throws XGBoostError {
     byte[][] bytes = new byte[1][];
+    checkHandle();
```
CodingCat: let's make jvm crash, I just do not want to downgrade the issue which has a root cause in native layer

jvm-packages/xgboost4j/src/main/scala/ml/dmlc/xgboost4j/scala/Booster.scala
```diff
@@ -204,11 +202,6 @@ class Booster private[xgboost4j](private[xgboost4j] var booster: JBooster)
     booster.dispose()
   }
 
-  override def finalize(): Unit = {
-    super.finalize()
-    dispose
-  }
-
```
CodingCat: with the latest change, this should be kept, otherwise, we are relying on GC in our implementation which is not a good practice

jvm-packages/xgboost4j/src/test/java/ml/dmlc/xgboost4j/java/BoosterImplTest.java
```diff
@@ -395,4 +395,15 @@ public void testTrainFromExistingModel() throws XGBoostError, IOException {
     TestCase.assertTrue(booster1error == booster2error);
     TestCase.assertTrue(tempBoosterError > booster2error);
   }
+
+  @Test(expected = XGBoostError.class)
+  public void testThrowErrorOnDisposedBooster() throws XGBoostError {
```
CodingCat: we should make the test consistent what we fixed, i.e. we should test after reading the previous checkpoint, 

the previous booster and the current generated booster is `==`

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
Yunni: But SIGSEGV is super-hard to debug, right? I would suggest we throw some meaningful exception when calling API on an reclaimed Booseter.

I can make another PR to throw the exception at native level. Do you think that makes sense?

jvm-packages/xgboost4j/src/main/scala/ml/dmlc/xgboost4j/scala/Booster.scala
Yunni: Sorry, do you mean this change should be kept or this overriden method should be kept? FYI, `dispose` is already been called in `JBooster.finalize()` https://github.com/dmlc/xgboost/pull/3005/files/8853a856d71242bc2986a42cba08e73933aee0cc#diff-e2f4a7e14a33d6f1c152a7f1b1a7a084L519

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
CodingCat: well....the issue is that you can identify a booster is reclaimed, because we set handle to 0 explicitly in java code. So using this approach is not a real (comprehensive) fix

if you really want to tackle this problem, we need to find a way to justify a reclaimed booster without the assistance in our java code

jvm-packages/xgboost4j/src/main/scala/ml/dmlc/xgboost4j/scala/Booster.scala
CodingCat: overridden method should be kept

my idea is that


* since you have ensured that no two scala boosters are sharing the same java booster, so java booster's finalize would be called for only once => we are safe to keep it 

* if you do not do that, we are actually saying that, "java booster" is a member variable of scala booster, we have guaranteed that it's one-one mapping, but we do not want to reclaim any resources used by this member variable instead we rely on the fact that java booster is not referred by anyone else, GC will do it eventually. 

While the second item is true, but introducing a dependency on GC for resource management is not a good practice.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
CodingCat: nvm, it's fine to use `handle = 0` to identify a reclaimed booster, can you do it in another PR?

jvm-packages/xgboost4j/src/main/scala/ml/dmlc/xgboost4j/scala/Booster.scala
Yunni: But the current implementation (on master) calls `JBooster.dispose()` twice. Once in `JBooster.finalize()` and once in `SBooster.finalize()`

I definitely agree relying on gc and override `finalize()` is not a good idea. I think user should control when to dispose() a booster, similarly as closing a IOStream.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
Yunni: Sorry, I don't get what do you mean by `justify a reclaimed booster`?

Yunni: Sure. Sounds good~

jvm-packages/xgboost4j/src/main/scala/ml/dmlc/xgboost4j/scala/Booster.scala
Yunni: BTW. Object.finalize() has been deprecated in JDK-9:
https://bugs.openjdk.java.net/browse/JDK-8165641

CodingCat: @Yunni the native resources of a xgboost booster is a resource owned by scala booster, the reason for the java booster to exist is just to reuse the code. The scala user should not be aware of java booster in the ideal case

when scala booster is reclaimed, it should proactively reclaim its resources, i.e. bytes in native layer, instead of relying on the fact that java booster has a reference on that and eventually java booster would be gone and the resources would be gone eventually

for JDK 9, you would refer to the API doc: https://docs.oracle.com/javase/9/docs/api/java/lang/Object.html, which says `Classes whose instances hold non-heap resources should provide a method to enable explicit release of those resources, and they should also implement AutoCloseable if appropriate.`

since making native booster as something like AutoCloseable and call in try block sounds a bit weird, our way (to reclaim resources as fast as possible) looks a bit closer to it's philosophy...in future, if we really move to JDK9, the only way might be explicitly manage boosters by building a channel between driver and executor, once the booster is downloaded to driver, we send something to executor side to say, you can release all boosters now

Yunni: @CodingCat `the native resources of a xgboost booster is a resource owned by scala booster` This's definitely not the implementation we have today. Right now SBooster is a wrapper around JBooster and JBooster inits and dispose native resource.

If we go with the way you are talking about, we should make `SBooster extends JBooster`. In this case, SBooster directly controls native resources instead of through the `jbooster` inside it.

In either ways, from the JVM perspective, as long as the object `jbooster` is not explicitly disposed or reclaimed, the native resources should not be reclaimed. `JBooster.dispose()` should be called exactly once. Otherwise we have to manually keep 1-1 mapping between `jbooster` and `sbooster` to avoid this kind of issue.

CodingCat: You can try to extend from java booster, you will find why we choose the current way to implement, i.e. make it as a variable but hidden from the user

You will find, if you extend, all apis are still with java-implemented classes, which does not make a real “scala” package

In this sense, scala booster has the resource, but manage it through a java class for code reusability

When scala booster is reclaimed, it should reclaim the resources, instead of making it expensive to implement a real scala package because we will have a indefinite length to really reclaim the native resource

CodingCat: JBooster.dispose has been idempotent...since it has that if condition

the real issue is not dispose() being called for 1+ time, it is because you assign an management util belonging to a scala booster to another one. The latter unfortunately gets a garbage, and is trying to behave good (being collected by the user) ...not allowed

Yunni: OK. So here's how it happens
```
val a = new SBooster(jbooster)
val b = new SBooster(jbooster)
a.method() // Crash sometimes because b could be gc-ed
```
I agree performance-wise this piece of code is not optimized enough, but I don't think the bug should be root caused at this level. Do you really prefer manual enforcement of 1-1 mapping between SBooster and JBooster? As our codebase keeps growing, this rule is very implicit and hard to catch.

I don't get what do you mean by `all apis are still with java-implemented classes`? Most Java API are Scala-friendly, right? If we do `SBooster extends JBooster`, I think we only need to implement a few APIs for scala Maps. 

@superbobry Would like to hear your thoughts on this, mind sharing?

CodingCat: Let me explain step by step

1. is `this rule is very implicit and hard to catch.`?

no and yes, 

why no:

```
val a = new SBooster(jbooster)
val b = new SBooster(jbooster)
a.method() // Crash sometimes because b could be gc-ed
```

(end-)user  cannot call this directly, since we have make constructor of scala booster private. The only way for them to create a booster is XGBoost.train. The only place to introduce this vulnerability is XGBoost.train and some tests (we will not encourage test private API anyway)

now, why yes, 

with the code base grows, any rule is becoming hard to capture unless you have something to prevent it. For example, you explicitly have something like auto_ptr, as no two auto_ptr objects should own the same element. This is like `can you ensure that there is no memory leak caused by unpaired new/delete in a huge c++ lib?". Since you have done an excellent job to find this issue, in future code review, we should be aware of this so that we can 

 * ensure that Scala booster's constructor should be called only in XGBoost.train

 * ensure that no many-to-one mapping of scala and java booster there

(I would suggest adding comments in scala booster class to explicitly claim this requirement)


2. `all apis are still with java-implemented classes` 

if you extend scala booster from java you will get access to java booster's public APIs. For example, 

```java
public void update(DMatrix dtrain, int iter) throws XGBoostError
```

This DMatrix is from java package. Now you have two options

* do nothing and ask the scala user to use this API: the issue is that scala user has to always switching between java/scala language APIs (e.g. java/scala collection APIs) to bridge java xgboost APIs and scala classes in their code. For instance, DMatrix in java package receives java.util.iterator as a constructor parameter, to use this, scala user has to convert (explicitly/implicitly) their scala collection to java collection and then get the iterator. This is apparently not a user-friendly design 

* implement scala.DMatrix and corresponding update API: the issue is that, every time the user programs in any IDE/editor+plugin, they will get auto-completion hints for two update APIs, one is with java.DMatrix, one is with scala.DMatrix. They have to eyeball to figure out which one to use. Another unfriendly design

Yunni: Since it's an open source project, I would assume developers are free to fork/extend their own version of the codebase. So if you really want to keep the overridden finalize method, I would add a package-private flag to prevent jbooster being shared by multiple sbooster besides adding the warning message.

`every time the user programs in any IDE/editor+plugin, they will get auto-completion hints for two update APIs` That's pretty common in scala development. Is that really considered user-unfriendly? I think maybe we should open another threads to discuss this topic.

CodingCat: Isn’t private flag there?

Can you point me to a serious scala project which has this issue but widely accepted as a non-issue?

CodingCat: let's discuss offline

PR #3044, opened 2018-01-17
c_api.cc variable declared inapproiate

mouendless:
In line 461, the "size_t offset = 0;" should be declared before any calculation, otherwise will cause compilation error. 

When using CMake for mvn package under jvm-package, will cause error below:
```
I:\Libraries\xgboost\src\c_api\c_api.cc(416): error C2146: Missing ";" before "offset" [I:\Libraries\xgboost\build\objxgboost.vcxproj]
```

and with the modification in the PR the problem solved.
Not sure with higher CMaker will fix the problem, but in C standard, the variable maybe better to be declared on the top in the function.

### DIFF
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -406,6 +406,7 @@ void prefixsum_inplace(size_t *x, size_t N) {
       suma[0] = 0;
     }
     size_t sum = 0;
+    size_t offset = 0;
 #pragma omp for schedule(static)
     for (omp_ulong i = 0; i < N; i++) {
       sum += x[i];
@@ -413,7 +414,6 @@ void prefixsum_inplace(size_t *x, size_t N) {
     }
     suma[ithread+1] = sum;
 #pragma omp barrier
-    size_t offset = 0;
     for (omp_ulong i = 0; i < static_cast<omp_ulong>(ithread+1); i++) {
       offset += suma[i];
     }


mouendless:
to be added, the C89 standard CMake would cause the issue above, while C99 and c++ maker could pass. I think it's still better to fix it for better compatibility

hcho3:
I'm not sure if it is C89 issue. For one thing, we have this line
```cpp
size_t sum = 0;
```
that is in the middle of the function.

Does your compiler support OpenMP? If not, we could have a case where OpenMP pragmas behaving weirdly.

mouendless:
yes we have ```size_t offset = 0```, but in c89 standard, declaration should come before any operation, otherwise it will not be recognized, while the original one come after ```suma[ithread+1] = sum;```.

hcho3:
Does not [this line](https://github.com/dmlc/xgboost/blob/d878c36c84edd625e93a6ba1635cba4f69f3defa/src/c_api/c_api.cc#L405) count as "operation"?

At any rate, I am not sure whether `c_api.cc` should even by compiled by a C89 compiler. I under that a C89 compiler may see `c_api.h`, but the actual implementation of C API layer should still be compiled by a C++ compiler.

mouendless:
I think omp operations is ultimate by the compiler and would not count.
Whether should be count may lay in your decision. 
I'm just telling I was using my platform with a certain cmaker and vs version and encountering this problem with compiling xgboost-jvm failure. And after this modification the compilation could pass.

hcho3:
Which Visual Studio version are you using? If it's a problem with Visual Studio version, I would just merge it for the sake of compatibility.

mouendless:
vs 2015 as I remember.
It's my hometown computer, I can provide the detail environment next week if you need.

hcho3:
Since it's a small change, I'll just go ahead and merge it

PR #3062, opened 2018-01-23
(Minor) Fix in SparkParallelTracker to work with Spark2.3

tomasatdatabricks:
This PR changes how xgboost registers and removes Listeners to SparkContext so that it works on Spark2.3 while it continues to work on versions of Spark >= 2.0.

Spark 2.3 includes changes in LiveListenerBus which are incompatible with xgboost's SparkParallelTracker - the SparkContext::listenerBus member variable changed type and it is no longer possible to register or deregister listeners by directly accessing list of listeners in it. I've changed  it so that listener is registered via public method on SparkContext. The public method for removing listener has been added only in Spark2.3 and in order to run on older versions of Spark I still call the method on the listenerBus instance.

This change means xgboost will not compile against/ run on spark version < 2.0.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
@@ -76,11 +76,12 @@ class SparkParallelismTracker(
   }
 
   private[this] def safeExecute[T](body: => T): T = {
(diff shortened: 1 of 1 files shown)


CodingCat:
we do not support spark 1.x in anyway since we have integrated with Spark 2.x ML pipeline

superbobry:
We'll need to detail this in the docs, I guess. 

Thanks @tomasatdatabricks!

PR #3068, opened 2018-01-25
Fixed the bug with illegal memory access in test_large_sizes.py with 4 GPUs.

canonizer:
- thrust::copy() called from dvec::copy() for gpairs invoked a GPU kernel instead of
  cudaMemcpy()
- this resulted in illegal memory access if the GPU running the kernel could not access
  the data being copied
- new version of dvec::copy() for thrust::device_ptr iterators calls cudaMemcpy(),
  avoiding the problem.

### DIFF
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -369,6 +369,16 @@ class dvec {
     }
     thrust::copy(begin, end, this->tbegin());
   }
+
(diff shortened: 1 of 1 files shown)


RAMitchell:
Thanks Andy I will test this tomorrow. Looks like the continuous integration is failing on something unrelated to this PR.

PR #3061, opened 2018-01-23
A fix for CRAN submission of version 0.7-0

hetong007:
This PR fix the current Error/Warning from `R CMD check --as-cran`, in order to have a new CRAN submission.

- The R-package version number increases to `0.7-0`, following the change of the main repository.
- Resolving https://github.com/dmlc/xgboost/issues/2960 and  https://github.com/dmlc/xgboost/issues/2948 .

The check logs from https://win-builder.r-project.org can be viewed at:

- R 3.3.3: https://win-builder.r-project.org/8xI5cw7al9UR/00check.log
- R 3.4.3: https://win-builder.r-project.org/Af73bVzp8BDq/00check.log
- R devel: https://win-builder.r-project.org/CC2L5vstoeP1/00check.log

Issue #2960, opened 2017-12-19
A new CRAN submission to solve the reported 'noLD' issue.


hetong007:
This was notified by **Prof Brian Ripley** through email:

> This concerns packages
> 
> CBPS ExtDist GGIR RInSp RSQLite drtmle expss gamlss.inf lakemorpho lhs partitionComparison pre recipes tableone vpc xgboost xpose
> 
> which have 'noLD' 'Additional issues' reported on their CRAN results pages.  And to quote the manual
> 
> <quote>
> Only test the accuracy of results if you have done a formal error analysis. Things such as checking that probabilities numerically sum to one are silly: numerical tests should always have a tolerance. That the tests on your platform achieve a particular tolerance says little about other platforms. R is configured by default to make use of long doubles where available, but they may not be available or be too slow for routine use. Most R platforms use ‘ix86’ or ‘x86_64’ CPUs: these use extended precision registers on some but not all of their FPU instructions. Thus the achieved precision can depend on the compiler version and optimization flags—our experience is that 32-bit builds tend to be less precise than 64-bit ones. But not all platforms use those CPUs, and not all which use them configure them to allow the use of extended precision. In particular, ARM CPUs do not (currently) have extended precision nor long doubles, and long double was 64-bit on HP/PA Linux.
> 
> If you must try to establish a tolerance empirically, configure and build R with --disable-long-double and use appropriate compiler flags (such as -ffloat-store and -fexcess-precision=standard for gcc, depending on the CPU type) to mitigate the effects of extended-precision calculations.
> </quote>
> 
> Packages partitionComparison and vpc do test for sums without a tolerance ....
> 
> Please submit a corrected update before **Jan 24** to safely retain the package on CRAN.

**Proposal**

As we are currently having some discussions about a new CRAN version (e.g.  https://github.com/dmlc/xgboost/issues/2948), it may be a good chance to work on a new version on CRAN. The difficulty for it mainly comes from https://github.com/dmlc/xgboost/issues/2319.

@khotilov @thirdwing, I would like to have your input on the following options:

1. Only fix the 'noLD' issue for the CRAN version and not introducing any changes on github.
2. A walkaround: fix the 'noLD' issue and merge our recent changes in R code to the CRAN version. However this may result in a different version from github.
3. Fix the 'noLD' issue and try again to solve https://github.com/dmlc/xgboost/issues/2319 . The benefit is obvious: this will sync our github repo and CRAN again, but the cause of the crash is still unclear.

[...]

PR #3076, opened 2018-01-28
Replacing cout with LOG

khotilov:
Submission of the R package to CRAN forbids the use of std::cout #2960, #3061. All the console output needs to go through the LOG facilities which are then captured in https://github.com/dmlc/xgboost/blob/master/R-package/src/xgboost_custom.cc

I did the change in a couple of files that are currently included in the R package submission to CRAN. @RAMitchell There's still some cout use in device_helpers.cuh and updater_gpu_hist.cu that should be preferably also changed to LOG. If there is interest, I can define, e.g., LOG_VERBATIM in https://github.com/dmlc/xgboost/blob/master/include/xgboost/logging.h that would allow using LOG(VERBATIM) very much as cout (it would not prefix its output with timestamp even when XGBOOST_LOG_WITH_TIME is defined, and would not have automatic newline).

### DIFF
--- a/src/common/quantile.h
+++ b/src/common/quantile.h
@@ -281,10 +281,10 @@ struct WQSummary {
   // helper function to print the current content of sketch
   inline void Print() const {
     for (size_t i = 0; i < this->size; ++i) {
-      LOG(INFO) << "[" << i << "] rmin=" << data[i].rmin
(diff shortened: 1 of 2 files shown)


hadjipantelis:
Any luck with the 0.7.0 find it's way to CRAN?

PR #3079, opened 2018-01-29
xgboost r

hblu:


khotilov:
I see that it's an H2O fork, but why are you trying to merge it here without any explanation?

hcho3:
I suggest that we close this issue until a further explanation is provided.

PR #3085, opened 2018-01-30
'hist': Montonic Constraints

redditur:
I have extended the monotonic constraints functionality so that they are now compatible with the 'hist' tree method. I have also added tests based on the repo by @madrury (https://github.com/madrury/xgboost-monotonic-testing).

### DIFF
--- a/src/tree/param.h
+++ b/src/tree/param.h
@@ -376,7 +376,7 @@ struct NoConstraint {
   inline static void Init(TrainParam *param, unsigned num_feature) {
     param->monotone_constraints.resize(num_feature, 0);
   }
-  inline double CalcSplitGain(const TrainParam &param, bst_uint split_index,
+  inline double CalcSplitGain(const TrainParam &param, int constraint,
                               GradStats left, GradStats right) const {
     return left.CalcGain(param) + right.CalcGain(param);
   }
--- a/src/tree/param.h
+++ b/src/tree/param.h
@@ -421,6 +421,7 @@ template <typename param_t>
 template <typename param_t>
   XGBOOST_DEVICE inline double CalcSplitGain(const param_t &param, int constraint,
                               GradStats left, GradStats right) const {
+    const double negative_infinity = -std::numeric_limits<double>::infinity();
     double wleft = CalcWeight(param, left);
     double wright = CalcWeight(param, right);
     double gain =
@@ -429,9 +430,9 @@ template <typename param_t>
     if (constraint == 0) {
       return gain;
     } else if (constraint > 0) {
-      return wleft < wright ? gain : 0.0;
+      return wleft <= wright ? gain : negative_infinity;
     } else {
-      return wleft > wright ? gain : 0.0;
+      return wleft >= wright ? gain : negative_infinity;
     }
   }
 
--- a/doc/tutorials/monotonic.md
+++ b/doc/tutorials/monotonic.md
@@ -76,3 +76,15 @@ Some other examples:
 
 - ```(1,0)```: An increasing constraint on the first predictor and no constraint on the second.
 - ```(0,-1)```: No constraint on the first predictor and a decreasing constraint on the second.
+
+**Choise of tree construction algorithm**. To use monotonic constraints, be
+sure to set the `tree_method` parameter to one of `'exact'`, `'hist'`, and
+`'gpu_hist'`.
+
+**Note for the `'hist'` tree construction algorithm**.
+If `tree_method` is set to either `'hist'` or `'gpu_hist'`, enabling monotonic
+constraints may produce unnecessarily shallow trees. This is because the
+`'hist'` method reduces the number of candidate splits to be considered at each
+split. Monotonic constraints may wipe out all available split candidates, in
+which case no split is made. To reduce the effect, you may want to increase
+the `max_bin` parameter to consider more split candidates.
--- a/src/tree/updater_fast_hist.cc
+++ b/src/tree/updater_fast_hist.cc
@@ -982,7 +982,7 @@ class FastHistTreeUpdaterSwitch : public TreeUpdater {
     inner_->Init(args);
   }
 
-  void Update(const std::vector<bst_gpair>& gpair,
+  void Update(HostDeviceVector<bst_gpair>* gpair,
               DMatrix* data,
               const std::vector<RegTree*>& trees) override {
     CHECK(inner_ != nullptr);
--- a/src/tree/updater_fast_hist.cc
+++ b/src/tree/updater_fast_hist.cc
@@ -870,13 +870,13 @@ class FastHistMaker: public TreeUpdater {
             if (d_step > 0) {
               // forward enumeration: split at right bound of each bin
               loss_chg = static_cast<bst_float>(
-                  constraint.CalcSplitGain(param, fid, e, c) -
+                  constraint.CalcSplitGain(param, param.monotone_constraints[fid], e, c) -
                   snode.root_gain);
               split_pt = cut_val[i];
             } else {
               // backward enumeration: split at left bound of each bin
               loss_chg = static_cast<bst_float>(
-                  constraint.CalcSplitGain(param, fid, c, e) -
+                  constraint.CalcSplitGain(param, param.monotone_constraints[fid], c, e) -
                   snode.root_gain);
               if (i == imin) {
                 // for leftmost bin, left bound is the smallest feature value
@@ -961,10 +961,45 @@ class FastHistMaker: public TreeUpdater {
   std::unique_ptr<TreeUpdater> pruner_;
 };
 
+// simple switch to defer implementation.
+class FastHistTreeUpdaterSwitch : public TreeUpdater {
+ public:
+  FastHistTreeUpdaterSwitch() : monotone_(false) {}
+  void Init(const std::vector<std::pair<std::string, std::string> >& args) override {
+    for (auto &kv : args) {
+      if (kv.first == "monotone_constraints" && kv.second.length() != 0) {
+        monotone_ = true;
+      }
+    }
+    if (inner_.get() == nullptr) {
+      if (monotone_) {
+        inner_.reset(new FastHistMaker<GradStats, ValueConstraint>());
+      } else {
+        inner_.reset(new FastHistMaker<GradStats, NoConstraint>());
+      }
+    }
+
+    inner_->Init(args);
+  }
+
+  void Update(const std::vector<bst_gpair>& gpair,
+              DMatrix* data,
+              const std::vector<RegTree*>& trees) override {
+    CHECK(inner_ != nullptr);
+    inner_->Update(gpair, data, trees);
+  }
+
+ private:
+  //  monotone constraints
+  bool monotone_;
+  // internal implementation
+  std::unique_ptr<TreeUpdater> inner_;
+};
+
 XGBOOST_REGISTER_TREE_UPDATER(FastHistMaker, "grow_fast_histmaker")
 .describe("Grow tree using quantized histogram.")
 .set_body([]() {
-    return new FastHistMaker<GradStats, NoConstraint>();
+    return new FastHistTreeUpdaterSwitch();
   });
 
 }  // namespace tree
--- a/tests/python/test_monotone_constraints.py
+++ b/tests/python/test_monotone_constraints.py
@@ -1,0 +1,85 @@
+import numpy as np
+import xgboost as xgb
+import unittest
+
+
+def is_increasing(y):
+    return np.count_nonzero(np.diff(y) < 0.0) == 0
+
+
+def is_decreasing(y):
+    return np.count_nonzero(np.diff(y) > 0.0) == 0
+
+
+def is_correctly_constrained(learner):
+    n = 100
+    variable_x = np.linspace(0, 1, n).reshape((n, 1))
+    fixed_xs_values = np.linspace(0, 1, n)
+
+    for i in range(n):
+        fixed_x = fixed_xs_values[i] * np.ones((n, 1))
+        monotonically_increasing_x = np.column_stack((variable_x, fixed_x))
+        monotonically_increasing_dset = xgb.DMatrix(monotonically_increasing_x)
+        monotonically_increasing_y = learner.predict(
+            monotonically_increasing_dset
+        )
+
+        monotonically_decreasing_x = np.column_stack((fixed_x, variable_x))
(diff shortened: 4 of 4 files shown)


aldanor:
Looks like the build failed for unrelated reasons.

@redditur If you squash your commits (or just the most recent one) and force-push the branch it will retrigger the travis build.

aldanor:
You can also [close the PR and then open it again](https://github.com/travis-ci/travis-ci/issues/576#issuecomment-65317915).

aldanor:
R build failed, again for unrelated reasons.

Could this be possibly merged?

(@RAMitchell? Among other things, this fixes a bug introduced when monotonic constraints for gpu_hist were added and `CalcSplitGain()` signature was changed)

redditur:
Yeah, it's related to a recent issue discussed on stackoverflow;(https://stackoverflow.com/questions/48545650/cannot-install-devtools-properly-without-digest-r)
As far as I can tell it should be a simple fix- editing the travis build config.

src/tree/param.h
```diff
@@ -429,9 +430,9 @@ template <typename param_t>
     if (constraint == 0) {
       return gain;
     } else if (constraint > 0) {
-      return wleft < wright ? gain : 0.0;
+      return wleft <= wright ? gain : negative_infinity;
     } else {
-      return wleft > wright ? gain : 0.0;
+      return wleft >= wright ? gain : negative_infinity;
```
khotilov: What is a reason for using -Inf here?

redditur: To ensure that the split never happens.

khotilov: I understand the general idea, but I'm curious why zero is not sufficient now, and whether there might potentially be any corner cases when the negative value might matter.

redditur: As far as I understand- and I may be completely wrong- the split is calculated on the basis of the change in loss (loss_chg). You can see here;

(https://github.com/dmlc/xgboost/blob/master/src/tree/updater_fast_hist.cc#L869)

that the loss change (at least in the fast histogram updater) is calculated as the gain of the split minus the gain of the current node, hence the loss change can be negative.

Therefore, the negative infinity ensures that the loss change is as small as possible for any split which violates the constraints.

khotilov: My memory could be wrong, but I thought that the negative loss change wasn't allowed by design.
However, if using -Inf doesn't break anything I'm ok with that, except please make the shortcut defined as const.

And I'm not sure whether that java_test fail is real or if travis is being a flaky primadonna again @superbobry ?

redditur:
There's no obvious reason to me why this should break the code. 

Okay, that makes sense. I'll set the shortcut to a constant.

src/tree/param.h
redditur: There's no obvious reason to me why this should break the code.

Okay, that makes sense. I'll set the shortcut to a constant.

khotilov: @redditur were you going to set the const?

khotilov:
@hcho3 Does the approach in this PR look good to you?

hcho3:
@khotilov I just got back from a conference (SysML 2018). I will write a review in a day or two.

src/tree/param.h
hcho3: @redditur Let me argue why the change is not strictly necessary, in terms of algorithmic correctness. The functions `CalcGain` and `CalcGainGivenWeight` always return nonnegative values by design. With algebra we can show that
```
Gain = G^2 / (H + lambda)
```
where G and H are the sums of first- and second-order gradients, respectively. Since all second-order gradients are nonnegative, H is also nonnegative and so is the gain. Since all gains are nonnegative, assigning gain 0 for invalid splits is okay.

At this point, we are arguing about aesthetics rather than correctness. The fact I just described is implicit and not well documented in code. So I'd still include `-Inf` for the sake of clarity.

hcho3: @redditur And maybe I'm nitpicking here, but I'd use
```cpp
return wleft > wright ? gain : negative_infinity
```
instead of
```cpp
return wleft >= wright ? gain : negative_infinity
```
The existing code invalidates the split whenever `wleft == wright`, so we should honor this behavior.

hcho3:
@aldanor When you said
> Among other things, this fixes a bug introduced when monotonic constraints for gpu_hist were added and CalcSplitGain() signature was changed)
Which bug are you referring to?

hcho3:
@khotilov It looks good to me. We should merge it in once a few comments are addressed.

hcho3:
@redditur Let me know whether you agree with our comments. If you do, I can make the necessary changes and merge it.

aldanor:
@hcho3 

https://github.com/redditur/xgboost/blob/220e6d9dd0affb61d87154470acc593e7f424b6a/src/tree/updater_fast_hist.cc#L870

When `gpu_hist` was implemented, the `CalcSplitGain()` signature was changed (here: https://github.com/dmlc/xgboost/pull/2904/files), however not all of its uses were updated in the coded. Specifically, in `updater_fast_hist.cc`, the `fid` index was wrongfully passed in place of a monotonic constraint.

aldanor:
@hcho3 

> The functions CalcGain and CalcGainGivenWeight always return nonnegative values by design

IIRC, the problem was that because of how `loss_chg` is calculated (split-gain minus gain), just setting the split-gain to 0 is not sufficient to prevent it from being used in case monotonic constraints are violated. As such, it's not an 'aesthetical' but rather a functional fix, @redditur may have some actual examples of that happening.

hcho3:
@aldanor 
> When monotonic constraints support for gpu_hist was implemented, the CalcSplitGain() signature was changed (here: https://github.com/dmlc/xgboost/pull/2904/files), however not all of its uses were updated in the code. Specifically, in `updater_fast_hist.cc`, the `fid` index was wrongfully passed in place of a monotonic constraint.

I notice that`NoConstraint::CalcSplitGain()` and `ValueConstraint::CalcSplitGain()` now have different signatures. The code still ended up working so far because 1) the hist updater only supported `NoConstraint` and 2) `NoConstraint::CalcSplitGain()` does not use the second argument at all.

Now that we are supporting `ValueConstraint` in the hist updater, we should fix the signature for `NoConstraint::CalcSplitGain()`.

hcho3:
@aldanor 
> just setting the split-gain to 0 is not sufficient to prevent it from being used in case monotonic constraints are violated

You would be right if there exists a case where all candidate splits have the split-gain of 0 and thus an invalid split (which also has split-gain of 0) may be chosen.

khotilov:
@hcho3 The discussion about the loss change positivity reminded me about seeing some sporadic `assert self.non_decreasing(res['train']['auc'])` failures in tests/python/test_fast_hist.py. Setting `np.random.seed(4)` in there and then running with `nosetests  tests/python/test_fast_hist.py` seems to make it reproducible. I didn't do any more digging, but I thought there might be some link with this.

As a side-note: many python tests do not fix or use the numpy random generator state properly.

hcho3:
@khotilov I had previously thought that the `min_split_loss` parameter would prevent negative loss change, but I was wrong. What's really happening is that the hist updater is allowed to make splits with negative loss change, and then later the pruner prunes the trees post-hoc using `min_split_loss` criteria. So for the purpose of monotonic constraints, we must not allow invalid splits at all, by setting negative infinity as the loss gain.

aldanor:
> So for the purpose of monotonic constraints, we must not allow invalid splits at all, by setting negative infinity as the loss gain.

Yes, exactly.

hcho3:
@aldanor I now stand corrected. We are indeed addressing a functional defect. One last thing: would you be in favor of replacing the line
```cpp
return wleft >= wright ? gain : negative_infinity
```
with
```cpp
return wleft > wright ? gain : negative_infinity
```
? I think we should preserve the behavior when `wleft == wright`.

redditur:
>  @redditur may have some actual examples of that happening.

The tests I've written are sufficient examples to see this happening.

> we should fix the signature for NoConstraint::CalcSplitGain().

Okay, I'll do this.

> As a side-note: many python tests do not fix or use the numpy random generator state properly.

This is a side-issue, but I'm not comfortable with tests setting seeds; it's just asking for hackery to be built into the tests (i.e. people setting specific seeds so that their tests pass). A test should pass regardless of the seed, so why set it?

> What's really happening is that the hist updater is allowed to make splits with negative loss change, and then later the pruner prunes the trees post-hoc using min_split_loss criteria. 

This is my understanding, but, still, those splits should be pruned and shouldn't exist at the end of fitting, correct? Yet, this is not the case. Try the tests using 0 instead of -infinity and with a non-zero gamma and you'll see them failing sometimes. So, the pruning step doesn't quite work as one would expect.

> I think we should preserve the old behavior when wleft == wright.

Firstly, the old behaviour is wrong, so I'm not sure why it should be preserved. Secondly, I'm not sure what you mean by preserve the old behaviour? In the previous incarnation, when the weights were equal the CalcSplitGain method would have outputted 0. The code snippet

`return wleft > wright ? gain : negative_infinity`

on the other hand would output negative_infinity. So, at least in that sense, the behaviour wouldn't be preserved. Besides I think to anyone reading the code, the suggested code snippet looks strange;  why have a strict comparison?

hcho3:
@redditur I meant to say that splits with `wleft == wright` should be invalidated.

redditur:
> splits with wleft == wright should be invalidated

@hcho3 

What's the reasoning for this though? It would only make the learner stricter than it needs to be, which could hinder prediction performance unnecessarily.

tqchen:
I think it is fine to remove constraint of for cases when ```wleft == wright```. In terms of loss comparison, it is better to return 0 as loss change can never be non-negative.

redditur:
@tqchen 
> I think it is fine to remove constraint of for cases when wleft == wright

By which you mean keep the line 

`return wleft >= wright ? gain : negative_infinity`

as is?

> loss change can never be non-negative.

This is not true; see the above discussion.

tqchen:
The point is that if the loss change ever becomes negative, the growing algorithm will choose not to split anyway, so setting things to 0 will be OK

tqchen:
I do mean we can keep the equality

hcho3:
>if the loss change ever becomes negative, the growing algorithm will choose not to split anyway, so setting things to 0 will be OK

I thought so too, but the parameter `min_split_loss` does not get applied at the time of tree construction.

hcho3:
@aldanor @redditur I am almost convinced about merging this PR in. Just to make sure, would you be able to provide us with an example where the old behavior results in an invalid split? Thanks!

redditur:
> Just to make sure, would you be able to provide us with an example where the old behavior results in an invalid split?

@hcho3 

Sure. Take the pull request as it is at the minute and change negative_infinity to 0. Then in the tests set numpy's seed,

`seed = 1234`
`np.random.seed(seed)`

and set the same seed in the parameters for the constrained_hist_method; that is,

`params_for_constrained_hist_method = {`
`    'tree_method': 'hist', 'silent': 1,`
`    'monotone_constraints': '(1, -1)', 'seed': seed`
`}`

The test for the constrained hist method will then fail.

hcho3:
@redditur I managed to reproduce the bug for 'hist'. Thanks! Do you also have an example where the old behavior fails for 'exact' method?

hcho3:
@redditur I looked into the code and found out that for some reason `root_gain` is being set to negative value. This phenomenon doesn't occur for the `exact` method. I'm trying to find out whether this behavior is legitimate or not.

redditur:
> By any chance, do you also have an example where the old behavior fails for 'exact' method?

@hcho3 

No, I don't. I've never seen the 'exact' method fail a monotonicity constraint and so think this is a 'hist' method specific issue.

hcho3:
Here is my findings, as for why the old behavior is broken:
* When monotonic constraints are applied, gain values may possibly be negative.
* Thus, it is possible that all valid splits get negative gains (@redditur gave us an example where this actually occurs); since invalid splits were assigned a gain of 0, an invalid split got picked as the "best".
So I am now in favor of merging @redditur 's PR.

Before we merge in this code, we must decide on one issue: how should we handle the scenario where all valid splits result in negative loss change?
1) Do not split at all
2) Choose the best valid split, even with a negative loss change.
Right now, this scenario is handled incorrectly: the split is set to `(feature 0, threshold 0.0)` because the accumulator variable `snode[nid].best` gets initialized to `(feature 0, threshold 0.0, loss_chg 0.0)` and never gets updated afterwards.

khotilov:
@hcho3 1. Do not split at all sounds reasonable to me. The monotonicity requirement seems to act as a kind of regularization that might result in situations where making any split would be detrimental.

hcho3:
@khotilov I agree. I'll make a change to reflect this decision.

redditur:
> how should we handle the scenario where all valid splits result in negative loss change?

@hcho3 

I feel that we are in danger of this pull request diverging from its intended purpose. I think, therefore, that this should be opened as a separate issue. 

However, for what it's worth, I couldn't possibly decide between;

>1. Do not split at all
>2. Choose the best valid split, even with a negative loss change.

without understanding in more detail some of the decisions already taken in codebase.

Firstly, I don't understand what the loss change actually refers to and therefore don't understand why splits are compared based on this quantity. The name suggests that it refers to the change in some loss function brought about by a split, but the code doesn't suggest this is the case and, moreover, if this were the case, why would this quantity ever be negative?

Why aren't splits chosen based on the output of CalcSplitGain, which, if named appropriately, is the gain brought about by a proposed split?

hcho3:
> I feel that we are in danger of this pull request diverging from its intended purpose.

The intended purposr of this PR is to support monotonic constraints in 'hist' method. Your addition revealed a corner case that was not properly handled before. Fixing this is crucial for correct behavior of monotonic constraints, so I think it's reasonable to deal with it here.

>The name suggests that it refers to the change in some loss function brought about by a split ... if
this were the case, why would this quantity ever be negative?

Indeed the name does suggest the change in the loss function brought about by the split. It is actually possible that this quantity be negative ... because the leaf weight is modified by the monotonic constraints. (Without monotonic constraints, you do get always positive loss change, but with then you don't.)

hcho3:
@redditur Now that I think about it, I'm actually fine with opening another PR. Why don't you merge your work on top of latest master? And then I'll propose my bug fix. Thanks!

hcho3:
@redditur I went ahead and deleted my addition from this PR. I'll merge it when tests pass.

aldanor:
Looks like the tests pass ok?

hcho3:
Done. Thank you all!

PR #3093, opened 2018-02-04
[jvm-packages] Declared Spark as provided in the POM

### LINKED ISSUE #2891: Spark Version
xgboost does not support spark 2.2, perhaps because of <spark.version>2.1.0</spark.version> in below:

https://github.com/dmlc/xgboost/blob/master/jvm-packages/pom.xml

It would be great to upgrade or make it flexible for compile.
Currently using xgboost based on Spark 2.1 results in trainRDD working, but trainDataFrame not working.

superbobry:
Closes #2891.


CodingCat:
merged, thanks

Issue #3099, opened 2018-02-07
Linear modelling in XGBoost


RAMitchell:
I am planning the following work on improving linear modelling in Xgboost. Comments and feedback welcome. The goal is to make xgboost very competitive with other packages for linear modelling.

- [ ]  Refactor gblinear and add linear updater classes to allow different linear model implementations
- [ ]  Normalise l1 and l2 penalties to be consistent with number of training examples (same as sklear/glmnet)
- [ ]  Faster more accurate training with new coordinate descent algorithm

- [ ]  Early stopping based on user specified tolerance parameter. Early stopping based on eval metric is not accurate enough and doesn't consider regularisation

- [ ] More unit tests. Add tests for Poisson/Gamma (I suspect these will not work with current shotgun implementation)

- [ ] AVX/GPU acceleration

- [ ]  Add Newton method or bfgs based solver

khotilov:
I agree that the linear booster is an underdog that needs some good care. And GPU might really accelerate things here, especially if some of the Nvidia's matrix algebra could be used.

BTW, I have a branch that I started working on in Aug https://github.com/dmlc/xgboost/compare/master...khotilov:gblinear_dev?expand=1
that has a penalty normalization options and adds a top-N components stagewise algorithm. I never got time to finish the work to a clean state (add the updater interface, etc.). Do you mind me merging that later in the evening after doing some minimal reconciliation (for the predictor interface)?

And I'm not really sure what that early stopping issue is. The tolerance could be set in the wrapper interface.

RAMitchell:
I'm going to PR my work in the next couple of days. Let's just PR both and see what makes sense from there.

I want to do early stopping based on the delta weight. I.e. if the largest weight update in an iteration falls below tolerance. Sklearn uses a more sophisticated stopping criterea based on the duality gap that could also be an option. I'm open to ideas how to do that. My current thought was to add a return value to the DoBoost  c api function indicating if the algorithm has converged, then the python/r code can stop iterating.

PR #3103, opened 2018-02-09
Refactor linear modelling and add new coordinate descent updater

RAMitchell:
Improvements to linear modelling as discussed in issue #3099


khotilov:
I didn't have much time to review yet, but the restructuring looks nice!
I've updated [my branch](https://github.com/dmlc/xgboost/compare/master...khotilov:gblinear_dev) to working condition, but it wouldn't make much sense to PR it. I'll add my stuff after this would get merged.

src/linear/updater_shotgun.cc
```diff
@@ -0,0 +1,127 @@
+/*!
+ * Copyright 2018 by Contributors
+ * \author Tianqi Chen
```
hcho3: You should put your name on here

src/gbm/gblinear.cc
```diff
@@ -9,188 +9,98 @@
...
 // training parameter
 struct GBLinearTrainParam : public dmlc::Parameter<GBLinearTrainParam> {
   /*! \brief learning_rate */
-  float learning_rate;
-  /*! \brief regularization weight for L2 norm */
-  float reg_lambda;
-  /*! \brief regularization weight for L1 norm */
-  float reg_alpha;
-  /*! \brief regularization weight for L2 norm in bias */
-  float reg_lambda_bias;
```
hcho3: I notice that this parameter is gone. Is either of the two updaters different from the original algorithm in `GBLinear`?
Also, you should update `parameter.md` to document the new parameters.

src/gbm/gblinear.cc
```diff
@@ -9,188 +9,98 @@
...
 /*!
  * \brief gradient boosted linear model
  */
 class GBLinear : public GradientBooster {
  public:
   explicit GBLinear(bst_float base_margin)
-      : base_margin_(base_margin) {
-  }
+      : base_margin_(base_margin),
+        sum_instance_weights(0),
```
hcho3: You should rename this variable to `sum_instance_weight` (without the `s` in the end) to match the rest of the code (e.g. see `Update()` interface).

include/xgboost/linear_updater.h
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * \fn  virtual void LinearUpdater::Update(std::vector<bst_gpair>* in_gpair, DMatrix* data, gbm::GBLinearModel* model, double sum_instance_weight) = 0;
+   *
+   * \brief Updates linear model given gradients.
+   *
+   * \param in_gpair            The gradient pair statistics of the data.
+   * \param data                Input data matrix.
+   * \param model               Model to be updated.
+   * \param sum_instance_weight The sum instance weight.
```
hcho3: It took me a while to understand why the sum instance weight had to be part of the generic interface. Why don't you copy the comment line 27 of the file `coordinate_common.h`, which says
```
* \param sum_instance_weight the sum instance weight, used to normalise L1/L2 penalty
```
You should also add to `parameter.md` a clarification of L1/L2 penalty: that they will be normalized by the number of training instances, to follow the convention of sklearn/glmnet.

include/xgboost/linear_updater.h
```diff
@@ -0,0 +1,68 @@
...
+   */
+
+  virtual void Update(std::vector<bst_gpair>* in_gpair, DMatrix* data,
```
hcho3: We had discussed re-factoring `Update` interface to use `HostDeviceVector` consistently. Would this code work okay after the re-factor? My guess is that a few additional lines would be all that would be needed, but I'd like a confirmation from you.

hcho3: Or we could keep `HostDeviceVector` for tree updaters only, should we decide against future GPU support for linear boosting. Let me know what you think.

src/gbm/gblinear.cc
RAMitchell: I removed the reg_lambda_bias parameter. My justification for this is that it was undocumented in the first place and this parameter is not used in competing implementations. We could trivially leave it in but  I dont personally think it's necessary.

include/xgboost/linear_updater.h
RAMitchell: Sure, can do. I was thinking of creating a new docs page with info about linear model implementations as a few people have been asking about it.. Perhaps that would be a good place to clarify this.

include/xgboost/linear_updater.h
RAMitchell: I think it will trivially work by changing to HostDeviceVector. I do have plans for a gpu version in the near future.

RAMitchell:
One interesting thing I also noticed is that changing the dmatrix column iterator to NOT sort features results in a 2-3x speedup of the entire algorithm. This is due to significantly improved cache utilisation when accessing and modifying the gpair vector. I will push something when I figure out a clean way of implementing it.

hcho3:
> changing the dmatrix column iterator to NOT sort features results in a 2-3x speedup of the entire algorithm
Nice! I assume linear boosting does not require a sorted access to features, unlike tree boosting?

include/xgboost/linear_updater.h
hcho3: That's good to know.

include/xgboost/linear_updater.h
hcho3: Along with it, can we also make a demo as well?

src/gbm/gblinear.cc
hcho3: Fair enough then. So, is the `coordinate` algorithm same as the original algorithm, save for the missing `reg_lambda_bias` parameter?

src/gbm/gblinear.cc
```diff
@@ -317,60 +226,51 @@ class GBLinear : public GradientBooster {
   }
 
  protected:
-  inline void Pred(const RowBatch::Inst &inst, bst_float *preds, int gid, bst_float base) {
+  bool CheckConvergence() {
+    if (is_converged) return true;
```
khotilov: Is_converged || param.tolerance == 0.

src/linear/updater_coordinate.cc
```diff
@@ -0,0 +1,133 @@
...
+    DMLC_DECLARE_FIELD(coordinate_selection)
+        .set_default("cyclic")
+        .describe("Coordinate selection algorithm");
+    DMLC_DECLARE_FIELD(maximum_weight)
+        .set_lower_bound(0.0f)
+        .set_default(1e15f)
+        .describe(
+            "Maximum absolute value of linear model weights. Prevents "
+            "coefficients increasing to infinity in separable logistic "
+            "regression problems.");
```
khotilov: I'd rather have an error thrown when complete separation happens and things go out of wack. I can't think of any practical context when such clipping could be useful, as the regularization is a more proper tool to deal with this.

src/gbm/gblinear.cc
```diff
@@ -9,188 +9,98 @@
...
-    } else {
-      return std::min(-(sum_grad + reg_lambda * w - reg_alpha) / (sum_hess + reg_lambda), -w);
-    }
-  }
-  // given original weight calculate delta bias
-  inline double CalcDeltaBias(double sum_grad, double sum_hess, double w) const {
-    return - (sum_grad + reg_lambda_bias * w) / (sum_hess + reg_lambda_bias);
+    DMLC_DECLARE_FIELD(linear_updater)
+        .set_default("updater_shotgun")
+        .describe("Update algorithm for linear model.");
```
khotilov: It would be very helpful when descriptions would include a list of current available options.
Another approach that would keep things clear is to use enum.

src/linear/updater_coordinate.cc
```diff
@@ -0,0 +1,133 @@
...
+        .set_lower_bound(0.0f)
+        .set_default(0.0f)
+        .describe("L2 regularization on weights.");
+    DMLC_DECLARE_FIELD(reg_alpha)
+        .set_lower_bound(0.0f)
+        .set_default(0.0f)
+        .describe("L1 regularization on weights.");
+    DMLC_DECLARE_FIELD(coordinate_selection)
+        .set_default("cyclic")
+        .describe("Coordinate selection algorithm");
```
khotilov: Either list the available options or use enum.

src/gbm/gblinear.cc
```diff
@@ -9,188 +9,98 @@
...
-          }
-        }
-      }
+    this->LazySumWeights(p_fmat);
+
+    if (!this->CheckConvergence()) {
+      updater->Update(in_gpair, p_fmat, &model, sum_instance_weights);
     }
+
+    iteration++;
```
khotilov: I don't see this `iteration` member being used anywhere.

src/linear/updater_coordinate.cc
```diff
@@ -0,0 +1,133 @@
...
+ */
+
+class CoordinateUpdater : public LinearUpdater {
+ public:
+  // set training parameter
+  void Init(
+      const std::vector<std::pair<std::string, std::string> > &args) override {
+    param.InitAllowUnknown(args);
+    selector.reset(CoordinateSelector::Create(param.coordinate_selection));
+    monitor.Init("GBLinear ", param.debug_verbose);
```
khotilov: A second "GBLinear " Monitor? Would it make sense to pass GBLinear's monitor to updaters?

src/gbm/gblinear.cc
hcho3: @khotilov As an aside, are you aware of any guideline about variable naming? Do we follow British spelling or American spelling? (initialised vs initialized) Not terribly important, but just curious.

include/xgboost/linear_updater.h
hcho3: Good point. I agree with your suggestion about having a parameter for the legacy behavior.

src/gbm/gblinear.cc
khotilov: In general, we should follow https://google.github.io/styleguide/cppguide.html#Variable_Names

include/xgboost/linear_updater.h
```diff
@@ -0,0 +1,68 @@
...
+  virtual ~LinearUpdater() {}
+  /*!
+   * \brief Initialize the updater with given arguments.
+   * \param args arguments to the objective function.
+   */
+  virtual void Init(
+      const std::vector<std::pair<std::string, std::string> >& args) = 0;
+
+  /**
+   * \fn  virtual void LinearUpdater::Update(std::vector<bst_gpair>* in_gpair, DMatrix* data, gbm::GBLinearModel* model, double sum_instance_weight) = 0;
```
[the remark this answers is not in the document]
RAMitchell: Ah yes. I will remove it :)

src/gbm/gblinear.cc
RAMitchell: I think IsConverged should always return false if tol == 0. The updater does not run if IsConverged== true.

src/gbm/gblinear.cc
```diff
@@ -9,188 +9,98 @@
...
-  }
-  // given original weight calculate delta bias
-  inline double CalcDeltaBias(double sum_grad, double sum_hess, double w) const {
-    return - (sum_grad + reg_lambda_bias * w) / (sum_hess + reg_lambda_bias);
+    DMLC_DECLARE_FIELD(linear_updater)
+        .set_default("updater_shotgun")
+        .describe("Update algorithm for linear model.");
+    DMLC_DECLARE_FIELD(tolerance)
+        .set_lower_bound(0.0f)
+        .set_default(1e-4f)
```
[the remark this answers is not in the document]
RAMitchell: Fair point but note that this is the convergence criteria sklearn uses for its elastic net..

src/linear/updater_coordinate.cc
RAMitchell: I forgot to change its name when I copied it. I think it's fine for the updater to have its own monitor as its not really necessary for it to be in the interface.

RAMitchell:
Regarding the change in default behaviour with normalisation - I think we don't need a switch as the user can just multiply alpha/lambda by number of training examples. I would consider adding normalisation more like fixing a bug than changing behaviour as i can't think of why you would ever want the unnormalised penalty. I think probably most users would have been unaware that it was not normalised.

khotilov:
While the normalization of likelihood and penalties as in sklearn/glmnet is the most common, other conventions exist out there. What was used in gblinear for a few years was a *specific convention*, while not without some issues, was not a bug. People were still able to use it and to optimize the penalty factors for their data. 

When introducing breaking changes (which, BTW, needs to be marked as such in the NEWS) and not giving an easy option to fall back to the old behavior, at least, you have to be rather specific in describing the change. So, in addition to mentioning that the current convention is consistent with sklearn/glmnet, providing a brief description for how to convert the penalty factors from the previous convention would be a nice thing to do.

src/gbm/gblinear.cc
khotilov: Yep, my bad...  But shouldn't then the line be `if (param.tolerance == 0.f) return false;`?

RAMitchell: Yes thanks

khotilov:
I was trying to do some test runs, but something ain't right with the shotgun updater. E.g.:
```
[1]	train-auc:0.720909	test-auc:0.721105 
[2]	train-auc:0.383117	test-auc:0.381982 
[3]	train-auc:0.643373	test-auc:0.644198 
[4]	train-auc:0.417603	test-auc:0.416076 
[5]	train-auc:0.593602	test-auc:0.594589 
[6]	train-auc:0.435729	test-auc:0.433470 
[7]	train-auc:0.593219	test-auc:0.594222 
[8]	train-auc:0.436250	test-auc:0.433684 
[9]	train-auc:0.593775	test-auc:0.594641 
[10]	train-auc:0.436242	test-auc:0.433523 
...
```
In the same data, AUC was quickly converging to 76% with the old version and with the cyclic or greedy coordinate descent. The random coordinate descent did converge, but with poor AUC ~ 67%.

Other observations that I've got while using it:
- could we rename the "linear_updater" parameter to "updater", just as for the gbtree?
- also, to keep in spirit with how the updaters are named for gbtree, instead of "updater_shotgun" and "updater_coordinate", could they be named, e.g., as "shotgun" and "coord_descent"? ("coordinate" alone doesn't carry that sense of closure if you get what I mean)
- I would prefer renaming "coordinate_selection" to maybe "feature_selector" or "order_selector"? as I would like to reuse this parameter for the shotgun, as well as to keep the "doer" semantics consistency. 
- Prediction caching does shave-off some time - good idea! And, surely, turning the column sorting off makes things ~2x faster.

RAMitchell:
That all sounds good. I will try to add some basic tests for shotgun. It's kind of annoying to test because it never gives the same answer.

RAMitchell:
@khotilov I have added some unit tests for shotgun but am unable to reproduce this divergent behaviour.

Can you provide a reproducible example?

khotilov:
Sure. It's a different simulated d/s, but shotgun still behaves badly:
```R
library(xgboost)
set.seed(1)
n <- 200000
p <- 500
x <- matrix(runif(n*p), n)
b <- rnorm(50)
y = rbinom(n, 1, plogis(x[, 10*(1:50)] %*% b + rnorm(n)))

itr <- sample(n, floor(0.8*n))
dtr <- xgb.DMatrix(x[itr,], label=y[itr])
dts <- xgb.DMatrix(x[-itr,], label=y[-itr])

bst <- xgb.train(data=dtr, watchlist=list(train=dtr, test=dts), nrounds=30,
                 booster='gblinear', objective = "binary:logistic", eval_metric="auc",
                 updater='coord_descent',
                 eta=0.5, nthread = 6, lambda=0.001, alpha=0.001)
```
```
[1]	train-auc:0.790928	test-auc:0.794081 
[2]	train-auc:0.803214	test-auc:0.805937 
[3]	train-auc:0.812966	test-auc:0.815227 
[4]	train-auc:0.820719	test-auc:0.822490 
...
[30]	train-auc:0.853822	test-auc:0.853279 
```
```R
bst <- xgb.train(data=dtr, watchlist=list(train=dtr, test=dts), nrounds=30,
                 booster='gblinear', objective = "binary:logistic", eval_metric="auc",
                 updater='shotgun',
                 eta=0.5, nthread = 6, lambda=0.001, alpha=0.001)
```
```
[1]	train-auc:0.842163	test-auc:0.845236 
[2]	train-auc:0.838214	test-auc:0.841844 
[3]	train-auc:0.824600	test-auc:0.823935 
[4]	train-auc:0.689022	test-auc:0.696358 
[5]	train-auc:0.557932	test-auc:0.557538 
[6]	train-auc:0.506578	test-auc:0.506170 
[7]	train-auc:0.500000	test-auc:0.500000 
[8]	train-auc:0.500000	test-auc:0.500000 
...
[30]	train-auc:0.500000	test-auc:0.500000
```

RAMitchell:
@khotilov this should be fixed now.

khotilov:
Great! I'll PR my additions over the weekend.

Issue #2766, opened 2017-10-03
[jvm-packages] evalSet returning constant value for all iterations.


mhmoudr:
For bugs or installation issues, please provide the following information.
The more information you provide, the more easily we will be able to offer
help and advice.

## Environment info
Operating System: Linux 
Package used (python/R/jvm/C++): jvm
`xgboost` version used: latest
The commit hash (`git rev-parse HEAD`): 178517524fd983ff1672d8929bd68ecfb4a0d730

## Steps to reproduce

1. try to build a tree model on a training dataset 
2. try to call evalSet for all iterations on hold-out dataset using code similar to:

```
(0 to iterations).foreach(i => {
    val evals = model.evalSet(Array(holdoutData), Array("holdout"), iter = i)
    println(evals)
})
```
then the result will be something similar to: 
```
[0]	holdout-rmse:22.571465
[1]	holdout-rmse:22.571465
[2]	holdout-rmse:22.571465
[3]	holdout-rmse:22.571465
[4]	holdout-rmse:22.571465
[5]	holdout-rmse:22.571465
[6]	holdout-rmse:22.571465
[7]	holdout-rmse:22.571465
[8]	holdout-rmse:22.571465
```
while I am expecting each iteration to give a different error, as I have tried to calculate the error manually and error was varying, also worth to mention that the error provided above was matching the first iteration error. 

## Expected result 
using the following code to calculate error for each iteration "not efficient enough as this code re calculate predictions from scratch"

```
(0 to iterations).foreach(i => {
    val lbl = holdoutData.getLabel
    val preds = model.predict(holdoutData,treeLimit = i).map(_(0))
    val mse = lbl.zip(preds).map(v=>Math.pow(v._1-v._2,2)).sum / lbl.length.toDouble
    val rmse = Math.sqrt(mse)
    println(s"$i: $rmse")
    rmse
})
```

Result:

```
0 :22.571464303032133
1 :22.84230245222776
2 :22.840524887185357
3 :22.838731954973362
4 :22.836923434431508
5 :22.83509939134127
6 :22.833259625287162
7 :22.831403881749665
8 :22.829532237854362
9 :22.827644608033697
10 :22.825740744783133
```

mhmoudr:
probably this issue is back to the code that exists in learner.cc as EvalOneIter does not do any prediction relates to the provided iteration, my reading that iter parameter is used for logging only (correct me if I am wrong!)
```
[...]

PR #3108, opened 2018-02-11
Fix evalOneIter function and use iteration index for predicting the value

mhmoudr:
More details in the ticket, reference #2766


mhmoudr:
closed as changes breaks internal training code, refer to discussion in ticket #2766 for more details

PR #3116, opened 2018-02-16
Replaced std::vector-based interfaces with HostDeviceVector-based interfaces.

canonizer:
- replacement was performed in the learner, boosters, predictors,
  updaters, and objective functions
- only interfaces used in training were replaced;
  interfaces like PredictInstance() still use std::vector
- refactoring necessary for replacement of interfaces was also performed,
  such as using HostDeviceVector in prediction cache


hcho3:
Nice! Let me review your code and get back to you in a day or two.

src/c_api/c_api.cc
```diff
@@ -705,14 +705,15 @@ XGB_DLL int XGBoosterBoostOneIter(BoosterHandle handle,
...
                                   bst_float *hess,
                                   xgboost::bst_ulong len) {
-  std::vector<bst_gpair>& tmp_gpair = XGBAPIThreadLocalStore::Get()->tmp_gpair;
+  HostDeviceVector<bst_gpair>& tmp_gpair = XGBAPIThreadLocalStore::Get()->tmp_gpair;
   API_BEGIN();
   Booster* bst = static_cast<Booster*>(handle);
   std::shared_ptr<DMatrix>* dtr =
       static_cast<std::shared_ptr<DMatrix>*>(dtrain);
-  tmp_gpair.resize(len);
+  tmp_gpair.resize(len, -1);
```
RAMitchell: This resize function looks like it is assigning the value -1 if you are used to the std vector interface. Perhaps you can do it like:
```
void resize( size_type count, T value = T(), int device = -1);
```

src/cli_main.cc
```diff
@@ -324,15 +324,15 @@ void CLIPredict(const CLIParam& param) {
...
+  HostDeviceVector<bst_float> preds;
   learner->Predict(dtest.get(), param.pred_margin, &preds, param.ntree_limit);
   if (param.silent == 0) {
     LOG(CONSOLE) << "writing prediction to " << param.name_pred;
   }
   std::unique_ptr<dmlc::Stream> fo(
       dmlc::Stream::Create(param.name_pred.c_str(), "w"));
   dmlc::ostream os(fo.get());
-  for (bst_float p : preds) {
+  for (bst_float p : preds.data_h()) {
```
RAMitchell: How does range based iteration work here? I thought you had to implement begin and end methods to make this work.

src/common/host_device_vector.cu
```diff
@@ -17,6 +17,16 @@ struct HostDeviceVectorImpl {
       data_h_.resize(size);
     }
   }
+  HostDeviceVectorImpl(std::initializer_list<T> init, int device)
```
hcho3: I like how you are supporting `std::initializer_list<T>` so that it's possible to use braced array constants to initialize `HostDeviceVector`. Here is a suggestion: why don't we also suggest initialization from `std::vector<T>` as well? This way, users can initialize `HostDeviceVector` with constants as well as vectors.

To minimize code duplication, we can have `HostDeviceVectorImpl` constructor to accept `std::vector<T>` and then expose two interfaces from `HostDeviceVector` that constructs `HostDeviceVectorImpl` objects.
```cpp
template <typename T>
HostDeviceVector<T>::HostDeviceVector(const std::vector<T>& init, int device)
  : impl_(new HostDeviceVectorImpl<T>(init)) {}

template <typename T>
HostDeviceVector<T>::HostDeviceVector(std::initializer_list<T> init, int device)
  : impl_(new HostDeviceVectorImpl<T>(init)) {}

...

template <typename T>
struct HostDeviceVectorImpl {
  HostDeviceVectorImpl(size_t size, T val, int device) { ... }
  HostDeviceVectorImpl(const std::vector<T>& init, int device) { ... }
};
```

src/objective/regression_obj.cc
```diff
@@ -38,27 +38,29 @@ class RegLossObj : public ObjFunction {
...
+    const omp_ulong n = static_cast<omp_ulong>(preds_h.size());
+    auto gpair_ptr = out_gpair->ptr_h();
     avx::Float8 scale(param_.scale_pos_weight);
 
     const omp_ulong remainder = n % 8;
     int nthread = omp_get_max_threads();
     // Use a maximum of 8 threads
 #pragma omp parallel for schedule(static) num_threads(std::min(8, nthread))
-    for (omp_ulong i = 0; i < n - remainder; i += 8) {
+    for (int i = 0; i < n - remainder; i += 8) {
```
hcho3: Can we keep the loop variable to `omp_ulong`? Any reason to change?

src/common/host_device_vector.cu
canonizer: Done.

src/objective/regression_obj.cc
canonizer: Done.

src/c_api/c_api.cc
canonizer: Done. Actually, I defined it as `resize(size_t n, int device = -1, T v = T())`, with device first, since I've got use cases where I want to have a non-default device, but I don't have any where I have a non-default fill value.

I've also added `T v = T()` parameter to the constructor.

include/xgboost/predictor.h
```diff
@@ -186,41 +170,14 @@ class Predictor {
...
-   * out_preds, const gbm::GBTreeModel& model, unsigned ntree_limit = 0)
-   *
-   * \brief Attempt to predict from cache.
-   *
-   * \return  True if it succeeds, false if it fails.
-   */
-  bool PredictFromCache(DMatrix* dmat, std::vector<bst_float>* out_preds,
-                        const gbm::GBTreeModel& model,
-                        unsigned ntree_limit = 0);
-
```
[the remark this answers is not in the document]
canonizer: Currently, these functions are defined in the Predictor class, but used only in CPUPredictor. GPUPredictor defines and uses its own versions of these functions. Since these functions are not virtual, the implementation in GPUPredictor does not override them, but just hides them, and I wanted to avoid that. Therefore, I moved these functions from Predictor to CPUPredictor, where they are actually used.

The alternatives considered:

- define these functions in GPUPredictor under a different name; I also wanted to avoid that, since the GPUPredictor does not use functions defined in Predictor;
- make these functions virtual, together with DevicePredictInternal/PredLoopInternal, and have the common logic in Predictor; this, however, quickly amounts to a refactoring of the predictors, and is a topic for a different pull request.

I'm happy to move these functions back to Predictor and have the GPU-based versions of these functions named differently in GPUPredictor, if you actually prefer it that way.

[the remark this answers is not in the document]
hcho3: Let's just remove these functions from Predictor class entirely, as they seem to be more of an implementation detail rather than a public interface. We can do another PR in the future if the functions need to be exposed again.

src/common/host_device_vector.h
```diff
@@ -62,8 +62,9 @@ template <typename T> struct HostDeviceVectorImpl;
 template <typename T>
 class HostDeviceVector {
  public:
-  explicit HostDeviceVector(size_t size = 0, int device = -1);
+  explicit HostDeviceVector(size_t size = 0, int device = -1, T v = T());
```
RAMitchell: I think I still prefer it in the order (size, v, device) for consistency with std vector.

canonizer: Done.

PR #3120, opened 2018-02-20
Added back UpdatePredictionCache() in updater_gpu_hist.cu.

canonizer:
Added back UpdatePredictionCache() in updater_gpu_hist.cu.

- it had been there before, but wasn't ported to the new version
  of updater_gpu_hist.cu


src/tree/updater_gpu_hist.cu
```diff
@@ -876,14 +926,22 @@ class GPUHistMaker : public TreeUpdater {
     omp_set_num_threads(nthread);
   }
```
RAMitchell: Do we need this old version of UpdatePredictionCache without HostDeviceVector?

src/tree/updater_gpu_hist.cu
```diff
@@ -876,14 +926,22 @@ class GPUHistMaker : public TreeUpdater {
...
+  bool UpdatePredictionCache
+    (const DMatrix* data, std::vector<bst_float>* p_out_preds) override {
     return false;
   }
 
-  bool UpdatePredictionCache(
-      const DMatrix* data, HostDeviceVector<bst_float>* p_out_preds) override {
-    return false;
+  bool UpdatePredictionCache
+  (const DMatrix* data, HostDeviceVector<bst_float>* p_out_preds) override {
```
RAMitchell: I notice that the fast histogram updater implementation of this function has a number of checks that are lacking here:
```
// p_last_fmat_ is a valid pointer as long as UpdatePredictionCache() is called in
      // conjunction with Update().
      if (!p_last_fmat_ || !p_last_tree_ || data != p_last_fmat_) {
        return false;
      }
```
@hcho3 can you comment on the purpose of these checks and if we need them here?

It looks like they are to ensure that the cached predictions inside the current updater match the input DMatrix.

src/tree/updater_gpu_hist.cu
hcho3: I don't think we need this, especially because the interface `TreeUpdater` now has only one version of `UpdatePredictionCache`, with HostDeviceVector.

src/tree/updater_gpu_hist.cu
hcho3: @RAMitchell According to `tree_updater.h`, `UpdatePredictionCache` should return `false` whenever the updater lacks the capability to update the prediction cache for a given data matrix:
```
  /*!
   * \brief determines whether updater has enough knowledge about a given dataset
   *        to quickly update prediction cache its training data and performs the
   *        update if possible.
   * \param data: data matrix
   * \param out_preds: prediction cache to be updated
   * \return boolean indicating whether updater has capability to update
   *         the prediction cache. If true, the prediction cache will have been
   *         updated by the time this function returns.
   */
```
This check is essential for correct operations of the predictor. The prediction cache is a shortcut that's only valid when the updater "knows" about the dataset for which prediction is being calculated. So you cannot use the shortcut for datasets other than the training data.

So yes, you do need the check.

RAMitchell:
@canonizer any updates?

canonizer:
Working on it.

src/tree/updater_gpu_hist.cu
canonizer: Done.

pseudotensor:
Were there bug fixes beyond the multi-GPU-ification of the eval_set ?  I see slightly different results in our reproducibility tests with this update, which says this update changed things even for 1 GPU.

canonizer:
This update is about re-introducing `UpdatePredictionCache()` for the GPU histogram updater, both for single and multiple GPUs. The implementation supports multiple GPUs. Apart from that, there were no other bug fixes.

Regarding reproducibility: could you point me to the tests that produce slightly different results for 1 GPU?

PR #3134, opened 2018-02-26
Additional improvements for gblinear

khotilov:
It's a followup to #3103. Details:
- Added a `FeatureSelector::Setup` method: it allows to set up the internal state of a FeatureSelector prior to looping through the features.
- Renamed `FeatureSelector::SelectNextFeature` to `FeatureSelector::NextFeature`.
- Changed the `feature_selector` parameter to enum. That allows having separate sets of allowed selectors in different updaters.
- Added a "shuffle" FeatureSelector: it could be seen either as a cyclic selector with random feature reshuffling prior to each update or as a random selector without replacement. It's helpful for reducing the bias from column order (note that when running "shotgun" without shuffling, features in the front would get priority in cyclic weight updates at every boosting round).
- Changed the meaning of the `FeatureSelector::NextFeature` return value of -1 to be "no feature was selected and no more expected for this group".
- Added a "thrifty" FeatureSelector: which is an approximately-greedy selector with linear complexity in the number of features. Parallel univariate weight estimation at the beginning of iteration supplies the order for feature selection, with an option to restrict the selection to top-k features (e.g., top_k=1 is a classic componentwise linear boosting that is known to approximate Lasso when the learning rate is small; also, top_k=1 is sort of an "unrolled" greedy updater where prediction gradients are updated after each feature).  With sufficiently small k, it has faster time-per-iteration than the shotgun, but it needs more iterations to converge. This updater would be a good option when only a relatively small fraction out of a large number of features is expected to remain in a model.
- Added the top_k restriction parameter to `GreedyFeatureSelector`. In addition to reducing the complexity per iteration to O(num_feature*top_k), not looping through all the features at each boosting round would help to reduce the "rotting" of prediction gradients for weak features, and would keep zero coefficients as zeros.
- Added the missing `eta` alias for `learning_rate` for CoordinateUpdater.
- For the shotgun updater, changed the learning rate default to be 0.5. I've frequently seen convergence issues with the default eta=1 (and even with etas > 0.5) in noisy datasets with many features.
- Made use of `GetBiasGradientParallel` and `UpdateBiasResidualParallel` within the `ShotgunUpdater` code.
- Since GetGradient, GetGradientParallel, and UpdateResidualParallel all work with a single feature, restricted the `ColIterator` within them to a single feature.
- alpha and lambda are now denormalized only once per update.
- Changed to have a consistent "alpha, lambda" parameter order in all the methods.
- Moved the `DumpModel` implementation into `GBLinearModel`.
- Added an R callback for collecting the history of linear coefficients through iterations.

Need to discuss the following:
- Consider getting rid of the `random` FeatureSelector in favor of `shuffle`. The former usually has poor performance, esp. in datasets with lots of non-sparse features.
- Consider setting the "shuffle" feature selector as default for the shotgun.
- Maybe consider changing the default eta from 1 to something lower for the coord_descend updater as well. While it's much less prone to convergence issues than the shotgun, avoiding the coefficients "explosions" during initial iterations is usually beneficial, esp. when top_k is not strongly restricted.
- Should there be two special return values for `FeatureSelector::NextFeature`? e.g.,
  - `-1`: "no feature was selected and no more expected for this group"
  - `-2`: "skip this feature"
- Consider switching to the column-wise parallelism in `GreedyFeatureSelector`, as it seems to be faster (sometimes quite faster) than the row-wise.

An illustration:
```R
library(magrittr)
# add 2nd order interactions
x <- model.matrix(Species ~ .^2, iris)[,-1]
dtrain <- xgb.DMatrix(scale(x), label = 1*(iris$Species == "versicolor"))
param <- list(booster = "gblinear", objective = "reg:logistic", eval_metric = "auc",
             lambda = 0.0003, alpha = 0.0003, nthread = 4)
# For 'shotgun', the default linear updater, using high eta values may result in
# unstable behaviour in some datasets. With this simple dataset, however, the high learning
# rate does not break the convergence, but allows to illustrate the typical pattern of
# "stochastic explosion" behaviour of this lock-free algorithm at early boosting iterations.
bst <- xgb.train(param, dtrain, list(tr=dtrain), nrounds = 300, eta = 1.,
                callbacks = list(cb.gblinear.history()))
# Extract the coefficients' path and plot them vs boosting iteration number:
coef_path <- xgb.gblinear.history(bst)
matplot(coef_path, type = 'l')
```

```R
# things look much more tame with smaller eta
xgb.train(param, dtrain, list(tr=dtrain), nrounds = 300, eta = .5,
          callbacks = list(cb.gblinear.history())) %>%
  xgb.gblinear.history %>%
  matplot(type = 'l')
```

```R
# selection of single feature per iteration results in Lasso-like paths
xgb.train(param, dtrain, list(tr=dtrain), nrounds = 600, eta = .9,
          updater = 'coord_descent', feature_selector = 'thrifty', top_k = 1,
          callbacks = list(cb.gblinear.history())) %>%
  xgb.gblinear.history %>%
  matplot(type = 'l')
```


RAMitchell:
> Consider getting rid of the random FeatureSelector in favor of shuffle. The former usually has poor performance, esp. in datasets with lots of non-sparse features.

Agreed.
> Consider setting the "shuffle" feature selector as default for the shotgun.

I am okay with this. What kind of performance improvement are you seeing?

> Maybe consider changing the default eta from 1 to something lower for the coord_descend updater as well. While it's much less prone to convergence issues than the shotgun, avoiding the coefficients "explosions" during initial iterations is usually beneficial, esp. when top_k is not strongly restricted.

I haven't seen any convergence issues yet so I will hold off on this.

> Should there be two special return values for FeatureSelector::NextFeature? e.g.,
-1: "no feature was selected and no more expected for this group"
-2: "skip this feature"

Yes this is fine.
> Consider switching to the column-wise parallelism in GreedyFeatureSelector, as it seems to be faster (sometimes quite faster) than the row-wise.
Yes this is fine. I didn't optimise this function as I'm not actually expecting anyone to use it, it was just for academic comparison. I was planning to implement a StochasticGreedyFeatureSelector that uses a small sample of training instances to estimate greedy selection. This would actually be usable. Not sure how close this is to your "thrifty" selector.

I suggest we keep all changes relating to feature selection out of the public parameter API for a month or two until they become stable.

Do you want to contribute to a paper? I am working on a set of benchmarks here https://github.com/RAMitchell/GLM-Benchmarks/blob/master/benchmark.py that will form the basis of a comparison against glmnet, sklearn and liblinear. The novel contributions will be the GPU accelerated coordinate descent as well as feature selection algorithms.

khotilov:
>> Consider setting the "shuffle" feature selector as default for the shotgun.

> I am okay with this. What kind of performance improvement are you seeing?

In a "worst case scenario" simulated experiment with significant features being in the last columns, I've seen the "shuffle" converging noticeably faster.

WRT the default learning rate being = 1 for the coordinate descent updater:

> I haven't seen any convergence issues yet so I will hold off on this.

I've seen some cases where convergence was faster with smaller eta. But yes, we can hold that off.

WRT the additional return code for `FeatureSelector::NextFeature` with the meaning of "skip this feature": after sleeping on it, I don't see it as potentially useful anymore...

WRT GreedyFeatureSelector:

> I didn't optimise this function as I'm not actually expecting anyone to use it, it was just for academic comparison.

For cases with a few dozens of features it's not that bad, and it takes quite less of boosting rounds to converge. Also the use of top_k makes it more manageable; and subsampling would make it faster too.

> I was planning to implement a StochasticGreedyFeatureSelector that uses a small sample of training instances to estimate greedy selection. This would actually be usable. Not sure how close this is to your "thrifty" selector.

When there are high correlations among features, the up-front univariate feature ranking estimate in the thrifty selector is not so good beyond a few first ones. So, an option for subsampling for greedy ranking should in principle be useful. While even for normalized penalties there's some dependence of their optimal values on the size of a sample, that shouldn't really matter for univariate best feature selection.

> I suggest we keep all changes relating to feature selection out of the public parameter API for a month or two until they become stable.

Ok.

> Do you want to contribute to a paper? I am working on a set of benchmarks here https://github.com/RAMitchell/GLM-Benchmarks/blob/master/benchmark.py that will form the basis of a comparison against glmnet, sklearn and liblinear. The novel contributions will be the GPU accelerated coordinate descent as well as feature selection algorithms.

Yes, I'm interested. I have some other ideas that might be worth considering.

src/linear/coordinate_common.h
```diff
@@ -98,16 +93,15 @@ inline std::pair<double, double> GetGradient(
...
-    int group_idx, int num_group, int fidx,
-
-    const std::vector<bst_gpair> &gpair, DMatrix *p_fmat) {
+inline std::pair<double, double> GetGradientParallel(int group_idx, int num_group, int fidx,
+                                                     const std::vector<bst_gpair> &gpair,
+                                                     DMatrix *p_fmat) {
   double sum_grad = 0.0, sum_hess = 0.0;
-  dmlc::DataIter<ColBatch> *iter = p_fmat->ColIterator();
+  const std::vector<bst_uint> fset(1u, static_cast<bst_uint>(fidx));
+  dmlc::DataIter<ColBatch> *iter = p_fmat->ColIterator(fset);
```
RAMitchell: This can use a list initialiser to save one line:
dmlc::DataIter<ColBatch> *iter = p_fmat->ColIterator({static_cast<bst_uint>(fidx)});

khotilov:
The delay is b/c I've realized that the processing of multiple SparsePages/data in external memory is broken - I'm fixing those.

khotilov:
The last two commits are the new ones.

I've added a col-wise parallel processing to the greedy selector. That made it in some cases more than twice faster. And also I've restructured a bit the computation to accommodate sparse pages. In principle, those updaters could even be made to work in a distributed manner.

Some observations about _external memory processing_:
- The default max_row_perbatch here is set to 65536 rows.
- The shotgun updater becomes a per-batch learner - lowering learning rate is important - something to keep in mind.
- The coordinate descent updater is still exact since it computes whole-sample gradient sums.
- There is some interface disconnect for the `max_row_perbatch` parameter: it is a DMatrix parameter, but it is passed during the creation of a booster. So, once `InitColAccess` was performed for a DMatrix, re-training a new booster with different `max_row_perbatch` would not change the existing batch structure of a DMatrix. Also, I saw some odd thing happening sometimes when the `max_row_perbatch` parameter was changed in when sharding was used (e.g., `DMatrix("x.csv?format=csv&label_column=0#shard1:shard2")`): `HaveColAccess` would start returning false at each iteration. I did not figure out the pattern yet.

RAMitchell:
> There is some interface disconnect for the max_row_perbatch parameter: it is a DMatrix parameter, but it is passed during the creation of a booster. So, once InitColAccess was performed for a DMatrix, re-training a new booster with different max_row_perbatch would not change the existing batch structure of a DMatrix. Also, I saw some odd thing happening sometimes when the max_row_perbatch parameter was changed in when sharding was used (e.g., DMatrix("x.csv?format=csv&label_column=0#shard1:shard2")): HaveColAccess would start returning false at each iteration. I did not figure out the pattern yet.

I would definitely like to get more clarity around this. I was thinking of re-factoring the learner LazyInitDMatrix method by making it call booster->LazyInitDMatrix(), in order to allow each booster type to make decisions about how the DMatrix should be configured instead of the learner.

Also, is it possible to add some unit tests the external memory version?

khotilov:
It turned out, HaveColAccess returning false was not really related to shards, but in general, when cached data was being loaded again, the `sorted` flag was not set for `col_iter_`. Setting it within `TryInitColData` seems to be a better spot.

One thing that I don't like about gblinear is that its convergence speed is currently very sensitive to having the columns centered (which is not really good for sparse data). Will need to think about how to improve that.

tests/python/test_linear.py
```diff
@@ -90,20 +94,43 @@ def train_classification_multi(param_in):
...
+    bst.__del__()
+    for f in glob.glob("tmptmp_*"):
+        os.remove(f)
 
 
 # Enumerates all permutations of variable parameters
 def assert_updater_accuracy(linear_updater, variable_param):
-    param = {'booster': 'gblinear', 'updater': linear_updater, 'eta': 1., 'tolerance': 1e-8}
+    param = {'booster': 'gblinear', 'updater': linear_updater, 'eta': 1.,
+             'top_k': 10, 'tolerance': 1e-5, 'ntread': 3}
```
RAMitchell: nthread is misspelled here.

PR #3137, opened 2018-02-26
Documenting CSV loading into DMatrix

hcho3:
We'd just need to expose the CSV parser in dmlc-core to the Python wrapper.

~~TODO: let user specify the column index containing the true label.~~ See below.


khotilov:
While it's not documented, CSVParser is already accessible via URI parameters, e.g.,
`xgb.DMatrix('data.csv?format=csv&label_column=0')`

PR #3139, opened 2018-02-28
Specify a value for num_class even if n_classes_ <= 2

cchan:
The error message produced was:
`xgboost.core.XGBoostError: value 0for Parameter num_class should be greater equal to 1`
This was due to not specifying a value for num_class.

The fix moves the setting of num_class out of the if statement.

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -465,6 +465,8 @@ class XGBClassifier(XGBModel, XGBClassifierBase):
         if self.n_classes_ > 2:
             # Switch to using a multiclass objective in the underlying XGB instance
             xgb_options["objective"] = "multi:softprob"
+
+        if self.n_classes_ >= 2:
             xgb_options['num_class'] = self.n_classes_
 
         feval = eval_metric if callable(eval_metric) else None


cchan:
It seems like some tests are failing. Should I modify the tests or am I doing something wrong?

khotilov:
Multiclass classification is sometimes guessed from the presence of a num_class parameter. 

Why is this change is necessary? Where was that error produced?

cchan:
@khotilov The error is produced when doing binary classification using XGBClassifier. It does do some guessing and sets the value, but the place where it sets the value is in an `if self.n_classes_ > 2`, so it never sets the value of `xgb_options['num_class']` and it defaults to zero I guess.

I've

cchan:
@khotilov I'm pretty confused about the test output; it seems that some of the data we're testing on has only one class. How can XGBClassifier over one class be meaningful? It just trivially predicts that one class, right?

cchan:
@khotilov Fixed the tests by only setting `xgb_options['n_classes']` if `n_classes_ >= 2`. Don't know what the desired behavior is when `n_classes < 2` but apparently setting `xgb_options['n_classes']` in that case was breaking the tests.

Let me know if you think I should add anything else to this PR! :)

cchan:
I'll look into it more tomorrow. Still not sure why this fixes it on my end but breaks the Travis build.

khotilov:
I still don't see the problem you trying to solve here.

num_class is only necessary for more than two classes. For binary classification objectives it's assumed that there are two classes.

hcho3:
Closing for now. If you think this PR is still relevant, feel free to re-open it.

Issue #3140, opened 2018-02-28
Early stopping is implemented incorrectly in XGBoost.java:203 (jvm-packages)


yvirin-apixio:
## Environment info

`xgboost` version used: 0.8
 
The commit hash (`git rev-parse HEAD`):
11bfa8584d8ddba2060ae2731f8599184e65f3b9

## The description of the problem
In the file: https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java line 203:

seems like the "&=" is not the correct operator to use, and instead it should be "|=" with decreasing initialized to false, in order to make sure consecutive "increases" in the loss function during evaluation rounds have been observed. With the current implementation the code stops after first increase, which is wrong. Probably it is a hard bug to catch in practice, I was just inspecting the implementation of early stopping and found this issue.
The intended usage of earlyStoppingRound parameter is described in line 121 (comment).

superbobry:
Feel free to submit a PR fixing this. There is also #2982.

yvirin-apixio:
@superbobry Thanks for connecting to the other issue. Somewhat related question: currently there is no way to know which is the optimal treeLimit to use in predictions, correct? Presumably it would be the tree which performed the best on evaluation during training.

hcho3:
Closing this for now. Feel free to re-open for any updates.

PR #3141, opened 2018-03-01
fix issue #3140: fix early stopping

yanvirin:
Fixing the unit test for early stopping, and then fixed the actual early stopping code to make the new unit test pass. This is to fix: https://github.com/dmlc/xgboost/issues/3140

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
@@ -197,13 +197,13 @@ public class XGBoost {
           metrics[i][iter] = metricsOut[i];
         }
 
-        boolean decreasing = true;
+        boolean decreasing = false;
         float[] criterion = metrics[metrics.length - 1];
         for (int shift = 0; shift < Math.min(iter, earlyStoppingRound) - 1; shift++) {
-          decreasing &= criterion[iter - shift] <= criterion[iter - shift - 1];
+          decreasing |= criterion[iter - shift] <= criterion[iter - shift - 1];
         }
 
-        if (!decreasing) {
+        if (!decreasing && earlyStoppingRound > 0) {
           Rabit.trackerPrint(String.format(
                   "early stopping after %d decreasing rounds", earlyStoppingRound));
           break;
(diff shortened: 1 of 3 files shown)


yanvirin:
I see that tests in CI sporadically fail. Does anyone know why?

xyzhu12:
in line 179 BoosterImplTest.java, we need to know the iter we stopped, which is not earlyStoppingRound. We need to get the last stop iter from XGBoost.train.

xyzhu12:
@yvirin-apixio Let make it clear with an example. Assume we set earlyStoppingRound=3, and round=7, the measurement is classification error. The error in each iteration is as follows:

iter=1     0.8
iter=1     0.8
iter=1     0.8
iter=1     0.8
iter=1     0.8

yvirin-apixio:
@xyzhu12 Is your comment about how I test the fix, or about what else should be done to make early stopping useful in xgboost? If its the former, than I think the test is doing ok - it is similar to what it was and I do not want to rely on the version of the booster to check the iteration. If that's the former, then I think this is something for another change. I wanted just to fix the way we stop, and the test is working correctly in that respect. I think currently it is the consumers responsibility to examine all the evaluation results once training is done and choose the best one.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -197,13 +197,13 @@ public static Booster train(
           metrics[i][iter] = metricsOut[i];
         }
 
-        boolean decreasing = true;
+        boolean decreasing = false;
         float[] criterion = metrics[metrics.length - 1];
         for (int shift = 0; shift < Math.min(iter, earlyStoppingRound) - 1; shift++) {
-          decreasing &= criterion[iter - shift] <= criterion[iter - shift - 1];
+          decreasing |= criterion[iter - shift] <= criterion[iter - shift - 1];
```
superbobry: I'm not sure this is correct. The training must stop after `earlyStoppingRound` consecutive increases in the criterion. Your PR changes the semantics to *at least a single decrease*.

yvirin-apixio: if there is at least one decrease, it means it is not increasing throughout the consecutive period. Pay attention to the fact that it checks "!decreasing" to stop early. So if it is decreasing, then it is fine, meaning if it found at least one decreasing round, then continue. If all rounds are strictly increasing, then it is overfitting and you should stop. That's the logic.

hcho3:
Closing as stale PR.

PR #3150, opened 2018-03-06
Handle an edge case for monotonic 'hist'

hcho3:
This is a follow-up of PR #3085.

When monotonic constraints are applied in conjunction with the 'hist' algorithm, it is often the case that all valid split candidates (those satisfying the monotonic constraints) result in negative loss change. This commit handles this edge case correctly, by refusing to split when any valid split results in negative loss change.

See discussion at https://github.com/dmlc/xgboost/pull/3085#pullrequestreview-100365400 for more details.

TODO: create a test case demonstrating the edge case.

### DIFF
--- a/doc/tutorials/monotonic.md
+++ b/doc/tutorials/monotonic.md
@@ -77,7 +77,7 @@ Some other examples:
 - ```(1,0)```: An increasing constraint on the first predictor and no constraint on the second.
 - ```(0,-1)```: No constraint on the first predictor and a decreasing constraint on the second.
 
-**Choise of tree construction algorithm**. To use monotonic constraints, be
+**Choice of tree construction algorithm**. To use monotonic constraints, be
 sure to set the `tree_method` parameter to one of `'exact'`, `'hist'`, and
 `'gpu_hist'`.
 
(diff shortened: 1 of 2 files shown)


hcho3:
It turns out this PR is unnecessary: [this line](https://github.com/dmlc/xgboost/blob/d5f1b74ef5b33ebb64863f66fd976d6c6efe0667/src/tree/updater_fast_hist.cc#L191) already eliminates all split candidates with zero or negative gains.

redditur:
@hcho3 
I would like to re-open this issue as I still believe this bug exists.

If, as you say, 

>  this line already eliminates all split candidates with zero or negative gains

then setting the gain to be 0 when the constraints are violated would be sufficient. However, as discussed in #3085, this is not the case.

Setting the gain to be -inf in #3085 was a hack to get that PR working; it's not a long-term solution.

hcho3:
With -inf in #3085, I was unable to trigger the edge case. That was why I decided not to proceed with this PR.

redditur:
> Why do you see this as a "bug", if it is never triggered?

Because -inf is a hack really. As discussed in #3085 setting the gain to be 0 when the constraints aren't satisfied should be fine. It isn't because, as discussed there, splitting happens even if the loss change is non-positive. 

I think the changes you proposed there- which I assume correspond to the changes in this PR- make sense. No splits should occur if all the loss changes are non-positive.

As a sanity check it probably makes sense to add to this PR the change of setting -inf to 0 when the monotonic constraints are violated to check if the tests still pass.

hcho3:
Closing this as a new method for monotonic constraints is introduced in #3335.

Issue #3156, opened 2018-03-08
Instability in gpu_hist


pseudotensor:
```
import xgboost
import pickle
import numpy as np
pathpickle='c92b4c2f-38c3-47ba-94f9-95af4058dcad_xgbobj2.dat'

model, XX_pd, y, _, kwargs = pickle.load(open(pathpickle, 'rb'))
del kwargs['sample_weight_eval_set']
del kwargs['early_stopping_threshold']
del kwargs['early_stopping_limit']
#del kwargs['early_stopping_rounds']
#kwargs['early_stopping_rounds'] = 30
#kwargs['eval_metric'] = "logloss"
#del kwargs['eval_set']
#kwargs['eval_set'] = (kwargs['eval_set'][0], (XX_pd, y))
kwargs['eval_set'] = ((XX_pd, y),)
#kwargs['verbose'] = False
#model.kwargs['tree_method'] = 'hist'
model.fit(XX_pd, y, **kwargs)
#model.fit(XX_pd, y)
```

[c92b4c2f-38c3-47ba-94f9-95af4058dcad_xgbobj2.zip](https://github.com/dmlc/xgboost/files/1791795/c92b4c2f-38c3-47ba-94f9-95af4058dcad_xgbobj2.zip)

If any type of eval set is passed, even the training set as above, still happens.  The training score does not descrease monotonically.  Using "hist" works correctly.

pseudotensor:
@RAMitchell

RAMitchell:
Fixed this bug in #3167

Turns out both hist and gpu_hist were incorrect.

PR #3167, opened 2018-03-14
Fix incorrect minimum value in quantile generation

RAMitchell:
There was a bug for both hist and gpu_hist where the lowest quantile could have an upper bound of zero and lower bound of zero. This fixes it so that the minimum value is strictly less than all other cut values.


RAMitchell:
It can be reproduced by running the script at #3156.

Specifically it occurred in a column that only contained the values 0 and nan.

hcho3:
I see. The old behavior doesn't work when the minimum value is exactly zero. I'll approve this PR.

PR #3171, opened 2018-03-14
Fixed supposed performance bug

Red-Portal:
I think this was meant to be this way.
The fix will result in very minor performance improvements


hcho3:
@RAMitchell Can you take a look at this and make sure this is what is intended?

PR #3182, opened 2018-03-17
Fix memory leak in XGDMatrixCreateFromMat_omp()

### LINKED ISSUE #3161: Memory leak in XGDMatrixCreateFromMat_omp()
## Environment info
Operating System: Ubuntu Artful in a Docker container (under an Ubuntu Artful host)

Compiler: Target: x86_64-linux-gnu
gcc version 7.2.0 (Ubuntu 7.2.0-8ubuntu3.2) 

Package used (python/R/jvm/C++): C

`xgboost` version used: From Git: d5f1b74ef5b33ebb64863f66fd976d6c6efe0667

## Steps to reproduce

1. Comment out the "avoid openmp line" (simply to get to call this function)
```
  // avoid openmp unless enough data to be worth it to avoid overhead costs
  //if (nrow*ncol <= 10000*50) {
  //  return(XGDMatrixCreateFromMat(data, nrow, ncol, missing, out));
  //}
```
2. Call `XGDMatrixCreateFromMat_omp()`
3. Run under valgrind (3.13.0 is what I'm using)

## What have you tried?

1. Nothing beyond verifying it

## Output
```
==2954== 4 bytes in 1 blocks are definitely lost in loss record 1 of 4
==2954==    at 0x4C3089F: operator new[](unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==2954==    by 0x5034285: XGDMatrixCreateFromMat_omp (c_api.cc:456)
==2954==    by 0x10A1F6: predict (main.c:468)
==2954==    by 0x10A0FC: thread (main.c:440)
==2954==    by 0x109EAE: start_threaded_predictions (main.c:391)
==2954==    by 0x10924D: main (main.c:104)
```

## Cause
We allocate memory with `new` but do not free it: `badnan = new int[nthread];` [Line](https://github.com/dmlc/xgboost/blob/master/src/c_api/c_api.cc#L456)

## Background
I found this by building xgboost using Clang's [static analyzer](https://clang-analyzer.llvm.org/scan-build.html) (from Clang 5.0.0-3). There were 13 things reported by it. As far as I can tell they were all false positives or not very important other than this one.

Thank you!

horgh:
This replaces the array allocated by new with a std::vector.

Fixes #3161

### DIFF
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -452,11 +452,8 @@ XGB_DLL int XGDMatrixCreateFromMat_omp(const bst_float* data,
   // Check for errors in missing elements
   // Count elements per row (to avoid otherwise need to copy)
   bool nan_missing = common::CheckNAN(missing);
-  int *badnan;
-  badnan = new int[nthread];
-  for (int i = 0; i < nthread; i++) {
-    badnan[i] = 0;
-  }
+  std::vector<bool> badnan;
+  badnan.resize(nthread, false);
 
 #pragma omp parallel num_threads(nthread)
   {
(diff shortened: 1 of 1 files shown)


RAMitchell:
Thanks for addressing this. Can you please change from vector<bool> to vector<int>.

It is actually unsafe for multiple threads to write to vector<bool> because internally it uses a bit packed representation.

vector<bool> is generally regarded as a bit of a mistake in the standard library and should often be avoided.

horgh:
Interesting. Thanks for the tip! I don't work with C++ very much and had no idea about that.

PR #3183, opened 2018-03-17
Allow compilation with -Werror=strict-prototypes

horgh:
Without this, with gcc 7.2.0, we see:

/xgboost/include/xgboost/c_api.h:98:1: error: function
declaration isn't a prototype [-Werror=strict-prototypes]
 XGB_DLL const char *XGBGetLastError();
  ^~~~~~~

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -95,7 +95,7 @@ XGB_EXTERN_C typedef int XGBCallbackDataIterNext(
  *  this function is thread safe and can be called by different thread
  * \return const char* error information
  */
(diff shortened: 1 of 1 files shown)


RAMitchell:
Can you resolve this error by instead writing:
```
XGB_DLL const char *XGBGetLastError(void);
```
In general I would prefer to deal with warnings/errors by addressing the problem instead of disabling the warning.

horgh:
I'm not sure I understand. Isn't that what this change is already doing?

RAMitchell:
Sorry ignore what I said :) For some reason I assumed you had changed the compiler flag.

PR #3184, opened 2018-03-17
Improve .gitignore patterns

horgh:
In particular, the patterns involving 'xgboost' seemed overly broad. We shouldn't ignore /include/xgboost, I think.


hcho3:
Looks good to me. Thanks!

PR #3185, opened 2018-03-19
'hist': improved subsampling

redditur:
The current subsampling approach used in xgboost is;

1. computationally inefficient.
2. highly dependent on the seed; specifically, certain seeds could lead to massive oversampling/undersampling.
3. unaligned with the documentation; specifically in the parameters section of the docs it (incorrectly) states;

"subsample [default=1]
subsample ratio of the training instance. Setting it to 0.5 means that XGBoost randomly collected half of the data instances to grow trees".

I would like to replace the approach with an approach that doesn't suffer these drawbacks; in particular I think it should be possible to apply the following idea;

1. do one step of a Fisher-Yates shuffle.
2. calculate the statistics of the first subsampled point.
3. if the hessian is positive then include this point as a datapoint.
4. repeat.

This will avoid the need to calculate and query the statistics for each datapoint; thereby reducing a, per tree, O(number of datapoints) operation to a, per tree, O(subsample size) operation.

I've already added some code in this pull request which implements the Fisher-Yates shuffle. However, elsewhere, the code still needs to be changed so that the statistics (the hessian and gradient) aren't calculated for each datapoint and only calculated on the fly to get the indices that are part of the subsample. 

I need help- from someone who knows the codebase in depth- for this last part because I'm not 100% sure on where/how best to do this (@khotilov @hcho3).

### DIFF
--- a/include/xgboost/data.h
+++ b/include/xgboost/data.h
@@ -292,6 +292,15 @@ class DMatrix {
   virtual float GetColDensity(size_t cidx) const = 0;
   /*! \return reference of buffered rowset, in column access */
   virtual const RowSet& buffered_rowset() const = 0;
+  // get subsampled indices.
+  virtual std::vector<uint64_t> get_subsampled_indices(
+      const float subsample,
+      const std::vector<xgboost::bst_gpair>& gpair
+  ) {
+      std::vector<uint64_t> subsampled_indices;
+      return subsampled_indices;
+  };
+
   /*! \brief virtual destructor */
   virtual ~DMatrix() {}
   /*!
--- a/src/data/simple_dmatrix.h
+++ b/src/data/simple_dmatrix.h
@@ -13,6 +13,7 @@
 #include <algorithm>
 #include <cstring>
 #include "./sparse_batch_page.h"
+#include "../common/random.h"
 
 namespace xgboost {
 namespace data {
@@ -20,7 +21,7 @@ namespace data {
 class SimpleDMatrix : public DMatrix {
  public:
   explicit SimpleDMatrix(std::unique_ptr<DataSource>&& source)
-      : source_(std::move(source)) {}
+      : source_(std::move(source)), shuffled_indices_(initialise_shuffled_indices_()) {}
 
   MetaInfo& info() override {
     return source_->info;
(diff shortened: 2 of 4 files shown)


RAMitchell:
I don't really see a problem with the current sampling scheme. The current scheme selects very close to the same number of instances with very high probability. It's also almost certainly faster than shuffling due to cache effects. Subsample selection is also not a bottleneck in my recent benchmarks. 

I can see a use case for subsampling before gradient calculation to avoid redundant work at gradient calculation time. Even then the gradient calculation is just  not that significant, particularly if avx is enabled.

redditur:
> The current scheme selects very close to the same number of instances with very high probability. 

Agreed, this is probably a non-issue.

> It's also almost certainly faster than shuffling due to cache effects.

I'm no computer scientist, but I doubt this very much. The cost from the extra random number generation- which is also probably not entirely cache friendly- is almost certainly more than the cost from the extra cache misses.

> Subsample selection is also not a bottleneck in my recent benchmarks.

I don't doubt this. However, why, for instance, query 1 million points 200 times when you only want 50 000?

On a side note, I'm not sure why points with negative hessians are excluded. I'm not convinced that such points are detrimental to fitting.

redditur:
> I suggest timing things before optimising to challenge your assumptions.

Unless I'm missing something major, the following code demonstrates that there is an inefficiency in subsampling in xgboost- subsampling should be slightly more computationally costly, not ~20 times more;

`import numpy as np`
`import xgboost as xgb`
`import sklearn.datasets as dset`
`import time`

`np.random.seed(12345)`

`# 1 million datapoints with a 5% subsample.`
`subsample = 0.05`
`n = int(1e6)`
`n_round = 100`

`param_sub = dict(`
    `max_depth=6, eta=1, silent=1,`
    `objective='binary:logistic', subsample=subsample`
`)`
`param = dict(`
    `max_depth=6, eta=1, silent=1,`
    `objective='binary:logistic', subsample=1.0`
`)`

`m = 10`
`ts_sub = np.empty(m, dtype=np.int64)`
`ts = np.empty(m, dtype=np.int64)`

`for i in range(m):`
    
    `# generate dataset.`
    `Xtrain, ytrain = dset.make_classification(n)`
    `dtrain = xgb.DMatrix(Xtrain, label=ytrain)`

    `# train on all data with subsampling.`
    `t0 = time.time()`
    `bst = xgb.train(param_sub, dtrain, n_round)`
    `t1 = time.time()`
    `ts_sub[i] = t1 - t0`

    `# get subsampled data.`
    `idxs = np.random.choice(np.arange(len(Xtrain)), size=int(subsample*n), replace=False)`
    `Xtrain, ytrain = Xtrain[idxs], ytrain[idxs]`
    `dtrain = xgb.DMatrix(Xtrain, label=ytrain)`

    `# train on subsampled data.`
    `t0 = time.time()`
    `bst = xgb.train(param, dtrain, n_round)`
    `t1 = time.time()`
    `ts[i] = t1 - t0`

`av = np.mean(ts)`
`std = np.std(ts)`
`av_sub = np.mean(ts_sub)`
`std_sub = np.std(ts_sub)`

`print('Average time taken for all data with subsampling: {} +- {}'.format(av_sub, std_sub))`
`# I get with 4 cores; 123.1 +- 1.758`
`print('Average time taken for subsampled data: {} +- {}'.format(av, std))`
`# I get with 4 cores; 5.1 +- 0.539`

RAMitchell:
Training on a preselected sample of a dataset is not the same thing as training on the full data set with sub sampling. With subsampling a new sample is selected at each boosting iteration. 

Closing for now, we can reopen if you want to do more work on this.

Issue #3194, opened 2018-03-22
CANNOT INSTALL XGBOOST: Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/wj/5tbzs3bx7892f45x20qsx9fh0000gn/T/pip-build-de5hi0sz/xgboost/


anthonygekko:
I am using OSX
I am installing xgboost but failed after putting 'pip3 install xgboost'
Python version: 3.6

"Collecting xgboost
  Using cached xgboost-0.7.post3.tar.gz
    Complete output from command python setup.py egg_info:
    ++ pwd
    + oldpath=/private/var/folders/wj/5tbzs3bx7892f45x20qsx9fh0000gn/T/pip-build-xrklq5op/xgboost
    + cd ./xgboost/
    + [[ darwin13 == \d\a\r\w\i\n* ]]
    + LIB_XGBOOST=libxgboost.dylib
    + make lib/libxgboost.dylib -j4
    Makefile:31: MAKE [/Library/Developer/CommandLineTools/usr/bin/make] - checked OK
    c++ -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -I/private/var/folders/wj/5tbzs3bx7892f45x20qsx9fh0000gn/T/pip-build-xrklq5op/xgboost/xgboost/dmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -msse2 -fPIC -fopenmp -MM -MT build/logging.o src/logging.cc >build/logging.d
    c++ -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -I/private/var/folders/wj/5tbzs3bx7892f45x20qsx9fh0000gn/T/pip-build-xrklq5op/xgboost/xgboost/dmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -msse2 -fPIC -fopenmp -MM -MT build/learner.o src/learner.cc >build/learner.d
    c++ -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -I/private/var/folders/wj/5tbzs3bx7892f45x20qsx9fh0000gn/T/pip-build-xrklq5op/xgboost/xgboost/dmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -msse2 -fPIC -fopenmp -MM -MT build/c_api/c_api.o src/c_api/c_api.cc >build/c_api/c_api.d
[...]

Issue #3118, opened 2018-02-17
Issue with Pip Install


Andreilys:
Downloading via pip install xgboost on my mac with python 3.5.4 I encounter this error:

    ++ pwd
    + oldpath=/private/tmp/pip-build-w_r84zub/xgboost
    + cd ./xgboost/
    + [[ darwin17 == \d\a\r\w\i\n* ]]
    + LIB_XGBOOST=libxgboost.dylib
    + make lib/libxgboost.dylib -j4
    Makefile:31: MAKE [/Applications/Xcode.app/Contents/Developer/usr/bin/make] - checked OK
    c++ -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -I/private/tmp/pip-build-w_r84zub/xgboost/xgboost/dmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -msse2 -fPIC -fopenmp -MM -MT build/learner.o src/learner.cc >build/learner.d
    c++ -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -I/private/tmp/pip-build-w_r84zub/xgboost/xgboost/dmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -msse2 -fPIC -fopenmp -MM -MT build/logging.o src/logging.cc >build/logging.d
    c++ -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -I/private/tmp/pip-build-w_r84zub/xgboost/xgboost/dmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -msse2 -fPIC -fopenmp -MM -MT build/c_api/c_api.o src/c_api/c_api.cc >build/c_api/c_api.d
    c++ -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -I/private/tmp/pip-build-w_r84zub/xgboost/xgboost/dmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -msse2 -fPIC -fopenmp -MM -MT build/c_api/c_api_error.o src/c_api/c_api_error.cc >build/c_api/c_api_error.d
    clang: error: unsupported option '-fopenmp'clang: error: unsupported option '-fopenmp'clang: error: unsupported option '-fopenmp'clang: error: unsupported option '-fopenmp'

    make: *** [build/logging.o] Error 1
    make: *** Waiting for unfinished jobs....
    make: *** [build/c_api/c_api.o] Error 1
[...]

Issue #2426, opened 2017-06-19
Impossible to install XGBoost on mac 


JeremieGuez:
Hi guys,

I am trying to install xgboost on my mac (10.12.5) and after several attempts I did not succeed. I follow different tuto trying to modifi gcc variable by adding gcc-6 but I have the same problem. 

So what I did:

1. git clone --recursive https://github.com/dmlc/xgboost
2. cd xgboost; cp make/minimum.mk ./config.mk; make -j4

I have the following error:

```
/usr/local/bin/g++ -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -Idmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -msse2 -fPIC -DDISABLE_OPENMP -MM -MT build/learner.o src/learner.cc >build/learner.d
/usr/local/bin/g++ -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -Idmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -msse2 -fPIC -DDISABLE_OPENMP -MM -MT build/logging.o src/logging.cc >build/logging.d
/usr/local/bin/g++ -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -Idmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -msse2 -fPIC -DDISABLE_OPENMP -MM -MT build/c_api/c_api.o src/c_api/c_api.cc >build/c_api/c_api.d
/usr/local/bin/g++ -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -Idmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -msse2 -fPIC -DDISABLE_OPENMP -MM -MT build/c_api/c_api_error.o src/c_api/c_api_error.cc >build/c_api/c_api_error.d
/bin/sh: /usr/local/bin/g++: No such file or directory
/bin/sh: /usr/local/bin/g++: No such file or directory
make: *** [build/learner.o] Error 127
make: *** Waiting for unfinished jobs....
/bin/sh: /usr/local/bin/g++: No such file or directory
make: *** [build/logging.o] Error 127
make: *** [build/c_api/c_api.o] Error 127
/bin/sh: /usr/local/bin/g++: No such file or directory
make: *** [build/c_api/c_api_error.o] Error 127
```
My gcc version is:

```
[...]

PR #3199, opened 2018-03-25
Fix up `make pippack` command for building source package for PyPI

hcho3:
* Now `make pippack` works without any manual action: it will produce `xgboost-[version].tar.gz`, which one can use by typing `pip3 install xgboost-[version].tar.gz`. I tested it on MacOS and Linux.
* Detect OpenMP-capable compilers (clang, gcc-5, gcc-7) on MacOS

This PR will fix recent reports of failed installation on MacOS, such as #3118, #3194, #2426, provided that a new release is issued using `make pippack`. I plan to make release 0.71 once this is merged.


hcho3:
@aldanor I am about to release version 0.71. Could you verify that `make pippack` produces a tarball that works with conda?

PR #3200, opened 2018-03-25
Release version 0.71

hcho3:
This is a follow-up of #3199. Users will be able to perform `pip install xgboost` on Linux and MacOS.
**Do not merge this PR until #3217 is merged first.**

EDIT: Also, need to include an update in R-package version.


lev112:
shouldn't the versions in the jvm pom files be updates as well?
they are currently 0.8-SNAPSHOT

hcho3:
@lev112 I think JVM package have its own versioning scheme.

CodingCat:
I am not going to have special versioning for jvm package...if you would like to, we can make release of xgb more regulated...

hcho3:
> if you would like to, we can make release of xgb more regulated...

Any suggestion on this?

CodingCat:
basically, we can follow the common practice in other open source project, versioning with x.y.z, x (major), y(minor), z (maintenance)

0. all PR should go to master branch

1. have regular release cycle for minor release, and have major release when we have significant update...in usual, only release maintenance version for the last minor one

2. when have a minor/major release, cut the branch from master

hcho3:
> have regular release cycle for minor release, and have major release when we have significant update

I like this idea. Certainly, we don't want the past situation where we were stuck with 0.6 for 1.5 years. Would quarterly release (every 3 months) be a reasonable interval? I will discuss this with my manager to ensure that I have sufficient bandwidth to do quarterly minor releases.

> when have a minor/major release, cut the branch from master

Can you clarify this part? Currently, all PRs get merged into the master, and each release is simply a tag in the linear history of master.

> versioning with x.y.z, x (major), y(minor), z (maintenance)

I had an extended discussion with R package maintainers about versioning scheme (see #3203). Here are some things we thought were reasonable:

* Use two digits to indicate the version of the core lib: `0.71`, `0.72`, `0.80` etc. The first digit will indicate the major release and the second the minor.
* For the R package, use a decimal to indicate package releases: `0.71.1`, `0.71.2` etc, since it is common to have multiple package releases that share the same core lib.

In light of this, I am wondering where maintenance releases would fit in.

hcho3:
@CodingCat Also, what is the policy for setting version for the JVM packages? Currently, `pom.xml` has `0.8-SNAPSHOT`.

CodingCat:
> I like this idea. Certainly, we don't want the past situation where we were stuck with 0.6 for 1.5 years. Would bi-monthly release (every 2 months) be a reasonable interval? I will discuss this with my manager to ensure that I have sufficient bandwidth to do bi-monthly minor releases.

yes, I think 2-months release cycle, even 3-months is reasonable

> Can you clarify this part? Currently, all PRs get merged into the master, and each release is simply a tag in the linear history of master.

e.g. say next release target is 0.8, then we should merge all patches to master until we decide it's time for 0.8 code freeze. At the time of freeze, we cut a new branch from master, say branch-0.8 and only  the following urgent fix should be to this branch-0.8 and other patches should still go to 0.8

and later, if we decide something is a very important  to be backport, we merge that to branch-0.8, otherwise it should go to master

> I had an extended discussion with R package maintainers about versioning scheme (see #3203). Here are some things we thought were reasonable: Use two digits to indicate the version of the core lib: 0.71, 0.72, 0.80 etc. The first digit will indicate the major release and the second the minor. For the R package, use a decimal to indicate package releases: 0.71.1, 0.71.2 etc, since it is common to have multiple package releases that share the same core lib.

I have 0 knowledge of R, which does package release mean here?

looks like your "minor" in the above comments equivalent to the maintenance release I referred, your major means minor in my mind. For major release, my idea actually means it jumps from 0.x to 1.x, 1.x to 2.x, etc.

> Also, what is the policy for setting version for the JVM packages? Currently, pom.xml has 0.8-SNAPSHOT.

yes, I was following what I described above when upgrading version number in jvm, 

say after we release 0.7, we should upgrade master branch version number to 0.8-SNAPSHOT, and 0.7 branch version number to 0.7.1-SNAPSHOT, and right before release, we should update by removing SNAPSHOT string there.... or we can skip the step of appending -SNAPSHOT

hcho3:
@CodingCat Thanks for your reply. I don't think we'd have any major release (in the sense of going up to 1.x, 2.x etc) for a while, until the existing features become mature. For now, I will make an announcement for a bi-monthly release cycle, with an estimated release dates for 0.72, 0.73, 0.80 etc. I will have to talk to other people about a written policy for "code freeze," since it's been a long while since this repo had a branch other than the master.

> I have 0 knowledge of R, which does package release mean here?

For Python and R packages, it may be possible to have a few package releases that share the same core lib version. This is especially common for the R package, since the R wrapper needs to be often updated to fit CRAN convention (see [history](https://github.com/dmlc/xgboost/commits/master/R-package) for a sample of R-specific maintenance commits).

Issue #3189, opened 2018-03-20
Question about version numbers / tags.


mingwandroid:
I built XGBoost (libxgboost, py-xgboost and r-xgboost) for the Anaconda Distribution recently.

I made a bit of a mistake with the versioning I think because I released it as version 0.7, whereas the one I built a year ago was version 0.60, numerically 7 is < 60 so our solver is preferring 0.60.

What is the *correct* version number here? I see differences between 'series' and 'tags' over at https://github.com/dmlc/xgboost/releases

Is the 0.7 versioned release a mistake that I should correct by changing it to 0.70 or should 0.60 really have been 0.6.0?

hcho3:
Hmm, apparently, some conventions (notably [this](https://setuptools.readthedocs.io/en/latest/setuptools.html#specifying-your-project-s-version)) treats `0.7` as older than `0.60`.
This is an oversight of my part; it should have been `0.70` to satisfy those conventions.

Would releasing a minor release `0.71` fix the issue? I don't think rewriting the repo history is a good idea, so it is infeasible to rename the `0.7` tag .

mingwandroid:
Sure! 0.71 would be just great.

hcho3:
I've filed PRs for release 0.71: #3199, #3200

mingwandroid:
Thank you.

PR #3203, opened 2018-03-25
Replace cBind by cbind

hetong007:
The package `Matrix` will be submitted to CRAN on May 25th, and deprecate its `cBind` and `rBind` by `cbind` and `rbind` respectively with a warning message.

We need a new CRAN submission for this change.

Source: https://stat.ethz.ch/pipermail/r-devel/2018-March/075737.html


hcho3:
> We need a new CRAN submission for this change.

Does the CRAN submission use the same version as the main repository? I have proposed a minor release 0.71 (primarily to fix broken Python installation), and I wonder if the release should wait until this PR gets merged.

hetong007:
We can wait for a week or two I believe.

hcho3:
@hetong007 I will wait then. Thanks!

hetong007:
@hcho3 Oh sorry, I guess I totally misunderstooded what you tried to say. Were you asking that if the release needs to include it?

If so, I'd say yes and I'll try to fix and merge this PR as soon as possible.

I believe R-package follows the main repository's version, according to the [contributor's guide](https://xgboost.readthedocs.io/en/latest/how_to/contribute.html#r-package-versioning). So, at the time of the release, is it desired to bump up the version number for all interfaces at the same commit?

Thanks!

hetong007:
@khotilov Can I have your kind input on this PR? Thanks!

hetong007:
@hcho3 This is merged.

hcho3:
@hetong007 Thanks for letting me know.

> I believe R-package follows the main repository's version, according to the contributor's guide.
I'm waiting a little more until I commit the 0.71 release. Which file should be updated to change R-package version?

terrytangyuan:
@hcho3 Could you also bump the date in DESCRIPTION? https://github.com/dmlc/xgboost/blob/master/R-package/DESCRIPTION#L5
Thanks!

mingwandroid:
Would `0.71.1` not be a better version number for the R package?

hetong007:
@mingwandroid our rules for R package versioning: https://xgboost.readthedocs.io/en/latest/how_to/contribute.html#r-package-versioning

mingwandroid:
I read those. It says:

> core_major.core_minor

core_major is 0 at present and core_minor is 71. That's what started this whole thing, because you need the numbers to be sane for every release, so 0.7 was a mistake given that 0.60 was already released in the past.

hcho3:
@mingwandroid @hetong007 @terrytangyuan #3189 notes that some versioning schemes in the Python ecosystem regard `0.7` as older than `0.60`. This is one of the reason why I have proposed the new release `0.71`.

Do you have a similar situation for the R package? If not, `0.7.1.1` seems more consistent with the existing guideline.

mingwandroid:
Yes, we [released](https://anaconda.org/r/r-xgboost/files) 0.7 unfortunately. I think 0.7 needs to be considered a mis-step (at least from a package builder's perspective) and 0.71 should be the new version that corrects this.

hetong007:
As long as users can 
1. see that this is a newer R package
2. track the underlying core version

I'd say the version number makes sense. 

Besides, R package and python package are not exactly the same, and probably won't be compared side-by-side. They are related only through the underlying core implementation. 

Can't really say which rule is better than the other, so we just follow what's already there for the R package.

mingwandroid:
> They are related only through the underlying core implementation.

I don't really see any good reason not to align the version numbers against that though? It makes things more simple.

mingwandroid:
FWIW, I build libxgboost, py-xgboost and r-xgboost all at once from a single recipe and libxgboost is shared between the python and the R packages.

https://github.com/AnacondaRecipes/xgboost-feedstock/blob/master/recipe/meta.yaml

hetong007:
I'd say it may confuse users if we change the format, and still with the current format we can match the underlying core version.

hcho3:
I'm inclined to stick with the current scheme for the R-package, unless there is a compelling reason to change (e.g. versioning convention that says `0.60 > 0.7`)

mingwandroid:
Yes, the Anaconda Distribution's version scheme (and IMHO most sane ones!) will consider `0.60 > 0.7`.

mingwandroid:
Please reconsider, this will cause our users *considerable* pain.

mingwandroid:
I do not think this is changing scheme at all, you are just jumping ahead a lot on the minor version to realign and to create an incrementing version story with 0.7 being a misstep.

hcho3:
> Please reconsider, this will cause our users considerable pain.

@mingwandroid Does all package on Anaconda follow the same versioning scheme, irrespective of languages (Python vs R)?

hcho3:
@mingwandroid I see that your proposal makes sense for Anaconda. It is indeed a "compelling reason". Only reservation I have is whether your proposal makes sense for CRAN as well. I'll wait to hear from @hetong007 and other R-package maintainers.

mingwandroid:
@hcho3, ok thanks. I believe other linux distros (or homebrew on macOS) that package R packages will end up in the same situation too though.

My fingers are crossed that this is acceptable to everyone!

hetong007:
@khotilov can we have your input on this?

mingwandroid:
> R users usually just check the version by simply comparing the strings in R like this:

This assumes that no external package manager is involved in installing the R package though. People who use system package managers are at the mercy of what those package managers do.

hetong007:
Is it possible to have a list of all external package managers involving the R-package?

khotilov:
While it feels odd that we need to resort to huge version jump because of a mistake, as long as the approach would be sound, settled and is consistently followed across the project, I would be happy. IMO, the versioning strategy that is currently in place for the R package is simple, well defined, rather universal and should be compatible with the majority of package managers. There are pros and cons to using a single version for different language interfaces to a single core library. But since the interfaces for Python and R are managed within the same repo, it would not be much of an effort to maintain a combined version number.

As it is with other sane version interpretation approaches, R doesn't do a lexicographic comparison of digits, and it compares integer vectors (after splitting the string by '[.-]' regex and converting to integers). 
```R
> package_version("0.60.4.1") > package_version("0.7.1.1")
[1] TRUE
```

khotilov:
However, with the single version, we would not be able to use the 3rd number to denote package releases, as it might be not feasible to release R and Python packages at the same time.

khotilov:
@hetong007 what do you see as a possible reason for breaking? Any sane modern package manager would treat 0.60 > 0.7 as true. 

@hcho3 what you are saying is what I have in mind and is what I wrote previously in https://xgboost.readthedocs.io/en/latest/how_to/contribute.html#r-package-versioning . But perhaps I was too pessimistic by saying "we would not be able" in the previous comment, since it is not really a requirement for package releases to have no gaps in versions. But the gaps would still look sloppy...

On unrelated note: was there any idea in mind about when the major version would change from 0 to 1?

hetong007:
@khotilov I think that, since what we do is to bump to 0.71, it won't break anything as 0.71 > 0.6. Therefore I'm not against it now.

hcho3:
@hetong007 @khotilov @mingwandroid @terrytangyuan I will go ahead and up the R version to `0.71.1`, if there is no objection. Thanks all!

hcho3:
I've updated PR #3200 to update the R version to `0.71.1`. In the future, let us stick with two digits for major core releases (`0.71`, `0.72`, `0.80`, etc).

hcho3:
> was there any idea in mind about when the major version would change from 0 to 1?

My guess is that 1.0 would represent the moment when the latest features (e.g. GPU support) becomes mature. @RAMitchell and I are currently working to set up a CI server for GPU code to this end.

Issue #3162, opened 2018-03-12
BUG: gpu_predictor in Python


dangoldner:
Hello,
I am getting similar behavior to #2965, but in Python. 

## Environment info
Operating System: linux (aws p2.xlarge)

Compiler: gcc

Package used (python/R/jvm/C++): python

`xgboost` version used: 0.7

If installing from source, please provide 

1. The commit hash (`git rev-parse HEAD`) - a1b48afa41f888

2. Logs will be helpful (If logs are large, please upload as attachment).

If you are using jvm package, please 

1. add [jvm-packages] in the title to make it quickly be identified
2. the gcc version and distribution

If you are using python package, please provide

1. The python version and distribution: Python 3.6.4 Anaconda (default Jan 16 2018, 18:10:19)

2. The command to install `xgboost` if you are not installing from source - Installed from source

If you are using R package, please provide

1. The R `sessionInfo()`
2. The command to install `xgboost` if you are not installing from source

## Steps to reproduce

```
gpu_params = {'tree_method':'gpu_hist', 
              'predictor':'cpu_predictor',
              'n_jobs': -1,
              'seed': 123
             }
m = xgb.XGBRegressor(**gpu_params).fit(X_ts, y_ts)
m.score(X_ts, y_ts) 
m.score(X_test, y_test)
```

Then do exactly the same thing changing only "cpu_predictor" to "gpu_predictor". Test score changes dramatically, predicted values change dramatically. With cpu, the values I get are sensible and verified.

RAMitchell:
@dangoldner I need your complete script including dataset to reproduce this.

dangoldner:
That's fair. Data are proprietary-I'll see if I can cook up some data that do the same thing.

dangoldner:
Still happening for me but I can't find a vanilla reproduction. I'll reopen if I can find a good example.

dangoldner:
Ok got it: 

Complete script is at https://gist.github.com/38a8c1d16d26283c9e7759f06be0fcd5. 

It seems to be related to the size of the problem: If `m, n = 10000, 7` and `tr_size = 6000`, the bug does _not_ occur. With larger sizes as shown in the gist, it does. 

Also note that the problem appears starting with the _second_ prediction, not the first.

dangoldner:
I pulled and rebuilt xgboost, but the behavior persists. 

Commit hash 00d9728

I ran the same script (below), from the terminal: 
[...]

PR #3217, opened 2018-04-03
Fix logic in GPU predictor cache lookup

RAMitchell:
Attempt #2 at fixing the phantom bug in #3162.

@dangoldner can you please try this fix and check it works before I merge this time.


dangoldner:
@RAMitchell Sorry just seeing this -- saw that the merge happened so tested master (8d66243f). Seems to work now: 
screen shot 2018-04-03 at 10 51 53 pm
Thanks for sticking with it - hope it doesn't resurface! :)

PR #3222, opened 2018-04-04
Clang-tidy static analysis

RAMitchell:
Add clang tidy based static analysis to the XGBoost code base.
http://clang.llvm.org/extra/clang-tidy/

As an experiment I have checked the file learner.cc using the "modernise" checks. This results in the following output:
```
/home/travis/build/RAMitchell/xgboost/src/learner.cc:144:24: warning: pass by value and use std::move [modernize-pass-by-value]
  explicit LearnerImpl(const std::vector<std::shared_ptr<DMatrix> >& cache)
                       ^
/home/travis/build/RAMitchell/xgboost/src/learner.cc:172:23: warning: escaped string literal can be written as a raw string literal [modernize-raw-string-literal]
      LOG(CONSOLE) << "Tree method is selected to be \'hist\', which uses a "
                      ^
                      R"(Tree method is selected to be 'hist', which uses a single updater )"
/home/travis/build/RAMitchell/xgboost/src/learner.cc:343:7: warning: use auto when declaring iterators [modernize-use-auto]
      std::map<std::string, std::string>::const_iterator it =
      ^
      auto
/home/travis/build/RAMitchell/xgboost/src/learner.cc:483:16: warning: escaped string literal can be written as a raw string literal [modernize-raw-string-literal]
            << "Tree method is automatically selected to be \'approx\'"
               ^
               R"(Tree method is automatically selected to be 'approx')"
/home/travis/build/RAMitchell/xgboost/src/learner.cc:486:16: warning: escaped string literal can be written as a raw string literal [modernize-raw-string-literal]
            << " set tree_method to \'exact\'";
               ^
               R"( set tree_method to 'exact')"
/home/travis/build/RAMitchell/xgboost/src/learner.cc:491:25: warning: escaped string literal can be written as a raw string literal [modernize-raw-string-literal]
        LOG(CONSOLE) << "Tree method is selected to be \'approx\'";
                        ^
                        R"(Tree method is selected to be 'approx')"
/home/travis/build/RAMitchell/xgboost/src/learner.cc:523:5: warning: use range-based for loop instead [modernize-loop-convert]
    for (size_t i = 0; i < cache_.size(); ++i) {
```

This will allow us to have much more powerful checks than a simple C++ style check. 

I would like to add some more checks and began expanding this to the rest of the c++ files and eventually cuda if possible.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,22 @@
+Checks: 'modernize-*,-modernize-make-*,-modernize-raw-string-literal,google-*,-google-default-arguments,-clang-diagnostic-#pragma-messages,readability-identifier-naming'
+CheckOptions:
+  - { key: readability-identifier-naming.ClassCase,           value: CamelCase  }
+  - { key: readability-identifier-naming.StructCase,          value: CamelCase  }
+  - { key: readability-identifier-naming.TypeAliasCase,          value: CamelCase  }
+  - { key: readability-identifier-naming.TypedefCase,          value: CamelCase  }
+  - { key: readability-identifier-naming.TypeTemplateParameterCase,          value: CamelCase  }
+  - { key: readability-identifier-naming.LocalVariableCase,          value: lower_case  }
+  - { key: readability-identifier-naming.MemberCase,          value: lower_case  }
+  - { key: readability-identifier-naming.PrivateMemberSuffix,           value: '_'  }
+  - { key: readability-identifier-naming.ProtectedMemberSuffix,           value: '_'  }
+  - { key: readability-identifier-naming.EnumCase,          value: CamelCase  }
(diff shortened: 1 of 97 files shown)


RAMitchell:
@khotilov I think using is generally prefered because it supports templated types. I think it reads better also so I would like to keep it if possible.

I will finish the google code checks next week then standardise all of the function/variable naming conventions.

After this I would like to implement at least some of the tougher corecpp guidelines checks. These prevent things like pointer arithmetic and unguarded indexing into arrays so should improve safety a lot, in particular for my gpu code.

RAMitchell:
@khotilov I still need to manually check over the last 5-6 files next week, but apart from that this should be mergable. 

Let me know if any more issues.

PR #3254, opened 2018-04-18
[jvm-packages] upgrade spark version to 2.3

CodingCat:


CodingCat:
@hcho3 looks like R pkg test failed in windows, any idea on why?

image

hcho3:
@CodingCat It appears that the package 'data.table' is unavailable as Windows binary. I made a small fix (#3257) to fall back on source install whenever Windows binaries are unavailable. (Read about the option `type='both'` in [documentation](https://www.rdocumentation.org/packages/utils/versions/3.4.3/topics/install.packages).)

CodingCat:
I see, thanks @hcho3

PR #3275, opened 2018-04-26
[jvm-packages] Fix broken link for xgboost-spark example.

yanboliang:
The link is broken. As DataFrame is the first class API for Spark, so I point the the link to DataFrame based example.

cc @CodingCat


CodingCat:
merged thanks

PR #3276, opened 2018-04-26
[jvm-packages] scripts to cross-build and deploy artifacts to github

CodingCat:
currently only support linux/mac


superbobry:
IIUC Mac support is only there if the host OS is macOS, right? How about going the Travis/Appveyor way as in [criteo-forks/xgboost-jar](https://github.com/criteo-forks/xgboost-jars)?

CodingCat:
yes, it relies on a mac machine to publish artifacts, (which seems fine to me, Mac is not that hard to find and this operation is just to be done by XGB developers not users)

travis-based solution looks more like a manual download of uber jar from travis if I didn't misunderstand it?

jvm-packages/dev/build-docker.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+set -x
+
+sudo docker run --rm -m 4g -e JAVA_OPTS='-Xmx6g' --attach stdin --attach stdout --attach stderr --volume `pwd`/../:/xgboost codingcat/xgboost:latest /xgboost/jvm-packages/crossbuild/build.sh
```
CodingCat: some path are messed up here, will fix it this evening

superbobry:
The current setup in xgboost-jars if semi-automatic. The build is triggered for every new tag. Both Travis and Appveyor upload the artifacts produced by the build to a GitHub release corresponding to the tag. One can then assemble the combined release via the [`download_latest_release.py`](https://github.com/criteo-forks/xgboost-jars/blob/master/download_latest_release.py) script which fetches all files from the latest GitHub release and merges them into a single JAR ready for publication.

CodingCat:
isn't it much more complicated than publishing to maven and let user add via pom.xml/build.sbt?

the only issue is the missing of windows support which could be done but I personally feel that the ROI there is pretty low

CodingCat:
the next thing here would be to fix the compiler versions and make the document

superbobry:
It is much more complicated indeed, but it offers support for all three platforms, which I think is the right thing to do (even though the Windows user base might seem marginal).

CodingCat:
for windows, there are some way to achieve that, e.g. use virtualbox instead of docker to build the dll file

I am still the holding the point that, let's put it to later, I didn't see many users desire a windows release (most of them working with windows are data scientists who like python obviously, but if you consider data scientists from companies like Uber which are MacOS-oriented companies, that would be another story)

for JVM users, a lot of them are just for Spark....

superbobry:
Late point: some of my colleagues prefer to develop on Windows for Linux-based production. Not having Windows DLLs in the official JARs would make development impossible for these people (since XGBoost does not have a pure Java fallback).

CodingCat:
it's not exclusive, you can still compile windows packages in your company's repo, and serve your development

for public XGBoost at this moment, I would suggesting not using a much more complicated release process just for covering a very small amount of users

PR #3273, opened 2018-04-25
Allow lazy init of GPU pointer, so even if compiled with GPU support, won't access GPU unless run GPU algorithm like gpu_hist.

pseudotensor:

PR #3287, opened 2018-04-30
Multi-GPU HostDeviceVector.

canonizer:
- HostDeviceVector instances can now span multiple devices, defined by GPUSet struct
- the interface of HostDeviceVector has been modified accordingly
- GPU objective functions are now multi-GPU
- GPU predicting from cache is now multi-GPU
- avoiding omp_set_num_threads() calls
- other minor changes

### DIFF
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -1004,14 +1004,29 @@ class AllReducer {
 
 template <typename T, typename FunctionT>
 void ExecuteShards(std::vector<T> *shards, FunctionT f) {
-  auto previous_num_threads = omp_get_max_threads();
-  omp_set_num_threads(shards->size());
-#pragma omp parallel
-  {
-    auto cpu_thread_id = omp_get_thread_num();
-    f(shards->at(cpu_thread_id));
+#pragma omp parallel for schedule(static, 1)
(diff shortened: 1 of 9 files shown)


RAMitchell:
@canonizer I have created a dh::ExecuteShards method for operating on data structures representing a gpu shard using a single thread each: https://github.com/dmlc/xgboost/blob/e1f57b4417146b3a11e14619b00851c67dffa0a3/src/linear/updater_gpu_coordinate.cu#L259

Can we generalise this to use it in your code also?

RAMitchell:
Looks good. Would be good to also apply ExecuteShards to gpu_hist when we do that pr. 

Re. Clang tidy, we could try including cuda toolkit  includes. I'm currently just ignoring clang tidy compile errors in the CI checks.

Can you also have a look at https://github.com/dmlc/xgboost/pull/3273. If we changed the thrust device vec member to some kind of lazy pointer we can use cuda binary on non cuda systems.

canonizer:
Now using `dh::ExecuteShards()` in GPU histogram updater as well.

Regarding #3273: I added my comments there. However, #3273 is superseded by this pull request, as the shards of HostDeviceVector are only initialized when `devices_` becomes non-empty. This happens either when object construction or `Reshard()` is called with a non-empty GPUSet.

PR #3298, opened 2018-05-07
Increase precision of bst_float values in tree dumps

samronsin:
`DumpRegTree` writes floats with default 6-digits precision which causes precision loss. 
This PR aims at setting the precision to a level where the conversion back to float is lossless.


hcho3:
Looks good to me. Thanks!

PR #3307, opened 2018-05-10
[jvm-packages] update xgboost4j cross build script to be compatible with older glibc

CodingCat:


CodingCat:
@dbtsai you can check this, I have uploaded the version compiled with GLIBC 2.12 and tested with 2.19 environment,

dbtsai:
+@lindblombr and @

Thanks @CodingCat for this PR. How can I test it if it run in our older Linux runtime? Will you cut a release or snapshot?

Thanks.

lev112:
@CodingCat I'm using the published xgboost 0.72, and getting the following error when running on my hadoop cluster

`/lib64/libm.so.6: version `GLIBC_2.23' not found`

the version of glibc on my cluster is 2.17

wasn't this PR supposed to solve this issue?
does 2.17 should work?

thanks

PR #3309, opened 2018-05-11
Add cuda 8/9.1 centos 6 builds, test GPU wheel on CPU only container.

RAMitchell:
@hcho3 these GPU enabled binaries are able to run with no dependencies in a CPU only centos 6 container.


RAMitchell:
@mmalohlava could you please review this.

PR #3311, opened 2018-05-11
CRAN Submission for 0.71.1

hetong007:
Several changes are made according to the feedback from CRAN's manual check.

1. When calling `xgb.dump` in examples/tests/demo/vignettes, only dump to `tempdir()`, instead of current working directory.

> Please ensure that your functions do not write by default or in your examples/vignettes/tests in the user's home filespace. That is not allow by CRAN policies. Please only write/save files if the user has specified a directory. In your examples/vignettes/tests you can write to tempdir().

2. Adding a list of authors for the base XGBoost implementation.

> From the CRAN policies you agreed to:
> 
> "The ownership of copyright and intellectual property rights of all components of the package must be clear and unambiguous (including from the authors specification in the DESCRIPTION file). Where code is copied (or derived) from the work of others (including from R itself), care must be taken that any copyright/license statements are preserved and authorship is not misrepresented.
> 
> Preferably, an ‘Authors@R’ would be used with ‘ctb’ roles for the authors of such code. Alternatively, the ‘Author’ field should list these authors as contributors.
> 
> Where copyrights are held by an entity other than the package authors, this should preferably be indicated via ‘cph’ roles in the ‘Authors@R’ field, or using a ‘Copyright’ field (if necessary referring to an inst/COPYRIGHTS file)."

I'm not keeping track of everyone's email address, so it is encouraged to update your record in [R-package/DESCRIPTION](https://github.com/dmlc/xgboost/blob/master/R-package/DESCRIPTION )


hcho3:
> Committers are people who have made substantial contribution to the project and granted write access to the project.

Is there a formal process to declare someone as committer?

hcho3:
I see. Thanks!

PR #3319, opened 2018-05-14
GPU binning and compression.

canonizer:
- binning and index compression are done inside the DeviceShard constructor
- in case of a DMatrix with multiple row batches, it is first converted into a single row batch


src/tree/updater_gpu_hist.cu
```diff
@@ -282,63 +440,109 @@ struct DeviceShard {
...
     int max_elements_row = 0;
-    for (auto i = row_begin; i < row_end; i++) {
-      max_elements_row =
-          (std::max)(max_elements_row,
-                     static_cast<int>(gmat.row_ptr[i + 1] - gmat.row_ptr[i]));
-    }
+    int block = 256;
+    thrust::device_vector<int> max_row_size_d(1, 0);
+    max_row_size_k<<<dh::DivRoundUp(n_rows, block), block>>>
+      (max_row_size_d.data().get(), row_ptr_d.data().get(), n_rows);
```
RAMitchell: Could this kernel be replaced with something like this:
```
int n = 10;
thrust::device_vector<size_t> offsets(n + 1);
thrust::sequence(offsets.begin(), offsets.end(), 0, 5);
auto segments = thrust::make_zip_iterator(
    thrust::make_tuple(offsets.begin(), offsets.begin() + 1));
auto get_size = [=] __device__(thrust::tuple<size_t, size_t> segment) {
  return thrust::get<1>(segment) - thrust::get<0>(segment);
};
using TransformT = thrust::transform_iterator<decltype(get_size),
                                              decltype(segments), size_t>;
TransformT row_size = TransformT(segments, get_size);
size_t max =
    thrust::reduce(row_size, row_size + n, 0, thrust::maximum<size_t>());
fprintf(stderr, "max %d\n", max);
```
If you used a cub reduce this would give state-of-the-art performance and be much safer. This function could be slightly modified to accept a binary op so you could do a max reduce: https://github.com/dmlc/xgboost/blob/49b9f39818d9002c3fffcd00da0bec6d573f4f17/src/common/device_helpers.cuh#L810

src/tree/updater_gpu_hist.cu
```diff
@@ -224,6 +224,163 @@ struct CalcWeightTrainParam {
...
+  }
+  size_t ielement = (irow + base_row) * row_stride + ifeature;
+  size_t ibit_start = ielement * nbits, ibit_end = (ielement + 1) * nbits - 1;
+  size_t ibyte_start = ibit_start / 8, ibyte_end = ibit_end / 8;
+
+  bin <<= 7 - ibit_end % 8;
+  for (ptrdiff_t ibyte = ibyte_end; ibyte >= (ptrdiff_t)ibyte_start; --ibyte) {
+    atomicOrByte(buffer, ibyte, bin & 0xff);
+    bin >>= 8;
+  }
```
RAMitchell: This bit compression code could be a function somewhere in this file: https://github.com/dmlc/xgboost/blob/master/src/common/compressed_iterator.h

Bit manipulation is extremely prone to errors so I would like to see some unit tests for this.

src/tree/updater_gpu_hist.cu
```diff
@@ -224,6 +224,163 @@ struct CalcWeightTrainParam {
...
+    return;
+  int row_size = static_cast<int>(row_ptrs[irow + 1] - row_ptrs[irow]);
+  uint bin = null_gidx_value;
+  if (ifeature < row_size) {
+    RowBatch::Entry entry = entries[row_ptrs[irow] - row_ptr_begin + ifeature];
+    int feature = entry.index;
+    float fvalue = entry.fvalue;
+    const float *feature_cuts = &cuts[cut_rows[feature]];
+    int ncuts = cut_rows[feature + 1] - cut_rows[feature];
+    bin = upper_bound(feature_cuts, ncuts, fvalue);
```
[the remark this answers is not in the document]
canonizer: No, as `thrust::upper_bound()` allocates memory and launches a kernel even in a single-element version on the device.

src/tree/updater_gpu_hist.cu
canonizer: Done (using thrust).

src/tree/updater_gpu_hist.cu
canonizer: Done.

src/tree/param.h
```diff
@@ -76,6 +76,8 @@ struct TrainParam : public dmlc::Parameter<TrainParam> {
   int gpu_id;
   // number of GPUs to use
   int n_gpus;
+  // number of rows in a single GPU batch
```
RAMitchell: I would like to see this be automatically calculated and the parameter removed. It is hard to see a user understanding and manipulating this parameter effectively.

src/tree/updater_gpu_hist.cu
```diff
@@ -224,6 +227,53 @@ struct CalcWeightTrainParam {
         learning_rate(p.learning_rate) {}
 };
 
+// index of the first element in cuts greater than v, or n none;
```
RAMitchell: I use something like this in the GPU predictor, we should look at standardising this function in device helpers. Not necessary for this PR though.

src/tree/updater_gpu_hist.cu
```diff
@@ -270,75 +320,119 @@ struct DeviceShard {
 
   dh::CubMemory temp_memory;
 
+  // TODO(canonizer): do add support multi-batch DMatrix here
```
RAMitchell: Unnecessary TODO based on our conversation about removing support for external memory?

tests/cpp/helpers.h
```diff
@@ -23,6 +23,8 @@ long GetFileSize(const std::string filename);
 
 std::string CreateSimpleTestData();
 
+std::string CreateBigTestData(size_t n_iters);
```
RAMitchell: I think this parameter should be called "n_entries"?

src/tree/param.h
canonizer: Done.

src/tree/updater_gpu_hist.cu
canonizer: Done.

tests/cpp/helpers.h
canonizer: Done.

RAMitchell:
@canonizer FYI uint is not defined in the C++ standard and does not work with MSVC. I have also modified to transform iterator code so it will compile on Windows.

Issue #3306, opened 2018-05-09
Very poor "GPU" performance with latest refactor, more extreme sensitivity to n_jobs setting


pseudotensor:
https://s3.amazonaws.com/h2o-release/h2o4gpu/snapshots/ai/h2o/h2o4gpu/slowgpu.zip

Isolated the problem.  Run this and then run with n_jobs=20 or 10 (instead of 39 or 40).  Whatever you guys did in the last 6 days with @canonizer led to extreme openmp problems.
￼￼￼￼￼
Before that commit, n_jobs=39 or 40 for 1 gpu run was no problem.  Now this issue ruins the GPU performance.  I thought things were only more isolated away from using the CPU, but this seems to suggest some major problems left over.

pseudotensor:
@RAMitchell @canonizer

pseudotensor:
for 1080ti: 

9.743741s for n_jobs=39
1.58s for n_jobs=10

When watching GPU usage and CPU usage, CPUs are used more and more heavily when n_jobs grows beyond ~20.

The same thing happens with "hist".  Things used to be good even if using n_jobs=40, but now they are not.

canonizer:
Looked at it. You are using `objective='binary:logistic'`, which is evaluated on the CPU, and causes additional copies between host and device. You should use `objective='gpu:binary:logistic'` to evaluate the objective on the GPU.

I'll take a closer look, but this is just the first thing that is noticeable.

canonizer:
Did a run of slowgpu.zip on a 6-core machine with 2 GP100 (only one of them was used). The run with n_jobs=40 takes about 15% more time than with n_jobs=10. However, I did not observe the 6x difference that you mentioned.

Could you tell me more about the hardware on which you did the run?

pseudotensor:
Dual  Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz with 40 hyper-threading cores.  Dual 1080ti. Ubuntu 16.04.2 LTS . NVIDIA-SMI 387.34                 Driver Version: 387.34 . cuda9.0.

pseudotensor:
Also, we can't use gpu:binary:logistic or gpu:reg:linear due to another bug where those return different shapes than the normal non-gpu versions.  I've yet to file that bug.

canonizer:
I'll take another look tomorrow.

[...]

PR #3324, opened 2018-05-16
Fix for issue 3306.

teju85:
Submitting PR for fixing issue #3306 on behalf of @canonizer .
@RAMitchell @pseudotensor can you guys review?


RAMitchell:
@teju85 @canonizer what was the actual problem? Was it not launching parallel threads or just running slowly for some reason when n_gpus=1?

@pseudotensor can you confirm that this fixes the problem before I merge.

pseudotensor:
Yes, thanks!  That fixed it.

This new version gets 3.5s consistently on the same script (issue #3306) with 40 cores, and 3.2s with 10 cores.  So not a crazy discrepancy anymore.

Thanks!

Issue #3208, opened 2018-03-29
java.lang.NoSuchMethodError: org.apache.spark.scheduler.LiveListenerBus.removeListener


dimon777:
For bugs or installation issues, please provide the following information.
The more information you provide, the more easily we will be able to offer
help and advice.

## Environment info
Operating System: MacOSX $ 10.12.6

Compiler: gcc-7

Package used (python/R/jvm/C++): python

`xgboost` version used: commit b08762066113db81b6872dfcc2c8f253ac092da4

If installing from source, please provide 

1. The commit hash (`git rev-parse HEAD`)
b08762066113db81b6872dfcc2c8f253ac092da4

2. Logs will be helpful (If logs are large, please upload as attachment).

1. add [jvm-packages] in the title to make it quickly be identified

2. the gcc version and distribution

If you are using python package, please provide

1. The python version and distribution: python 3.6

2. The command to install `xgboost` if you are not installing from source

## Steps to reproduce

1. Compile xgboost from source on MacOSX
2. Install spark 2.3 standalone 
3. Install https://github.com/apache/incubator-toree
4. Install https://github.com/jupyter-scala/jupyter-scala
5. Fire up jupyter-notebook, choose apache toree scala kernel

Observe error below

## What have you tried?

1. Installing apache-toree via pip and compiling it form source

The problem seems to be caused by XGBoost.

ERROR:

numRound: Int = 10
numWorkers: Int = 4
paramMap: scala.collection.immutable.Map[String,Any] = Map(min_child_weight -> 3.0, base_score -> 0.005, colsample_bylevel -> 0.9, silent -> 1, seed -> 49, subsample -> 1.0, max_depth -> 10, objective -> reg:linear, eval_metric -> auc, eta -> 0.023, colsample_bytree -> 0.82)
java.lang.NoSuchMethodError: org.apache.spark.scheduler.LiveListenerBus.removeListener(Ljava/lang/Object;)V
  at org.apache.spark.SparkParallelismTracker.safeExecute(SparkParallelismTracker.scala:84)
  at org.apache.spark.SparkParallelismTracker.execute(SparkParallelismTracker.scala:107)
  at ml.dmlc.xgboost4j.scala.spark.XGBoost$$anonfun$trainDistributed$4.apply(XGBoost.scala:354)
  at ml.dmlc.xgboost4j.scala.spark.XGBoost$$anonfun$trainDistributed$4.apply(XGBoost.scala:337)
[...]

Issue #3151, opened 2018-03-06
java.lang.NoSuchMethodError: org.apache.spark.scheduler.LiveListenerBus.listeners()Ljava/util/concurrent/CopyOnWriteArrayList


deniska2010:
Hi!
I'm trying to run the following code to train my Xgboost model:
```
val indexerAfterIF = stringIndexer(dataNotAnomalies)
val dataNotAnomaliesWithDoubleOkveds = indexerAfterIF.transform(dataNotAnomalies)
val trackerConf = new TrackerConf(0L, "scala")
    val numRound = 500
    val numWorkers = 10
    // training parameters
    val paramMap = List(
      "eta" -> 0.2f,
      "max_depth" -> 4,
      "min_child_weight" -> 3.0,
      "subsample" -> 0.55,
      "colsample_bytree" -> 0.56,
      "eval_metric" -> "mlogloss",
      "tracker_conf" -> trackerConf,
      "silent" -> 1,
      "objective" -> "multi:softmax",
      "num_class" -> nClasses,
      "lambda" -> 3.3000000000000003,
      "gamma" -> 4.7,
      "alpha" -> 0.9
    ).toMap
 val xgBoostModelWithDF = {
      XGBoost
        .trainWithDataFrame(dataNotAnomaliesWithDoubleOkveds,
          paramMap,
          round = numRound,
          nWorkers = numWorkers,
          useExternalMemory = true)
    }
```

But I got the following error:

`Exception in thread "main" java.lang.NoSuchMethodError: org.apache.spark.scheduler.LiveListenerBus.listeners()Ljava/util/concurrent/CopyOnWriteArrayList; at org.apache.spark.SparkParallelismTracker.safeExecute(SparkParallelismTracker.scala:75) at org.apache.spark.SparkParallelismTracker.execute(SparkParallelismTracker.scala:102) at ml.dmlc.xgboost4j.scala.spark.XGBoost$.trainDistributed(XGBoost.scala:350) at ml.dmlc.xgboost4j.scala.spark.XGBoostEstimator.train(XGBoostEstimator.scala:139) at ml.dmlc.xgboost4j.scala.spark.XGBoostEstimator.train(XGBoostEstimator.scala:36) at org.apache.spark.ml.Predictor.fit(Predictor.scala:96) at ml.dmlc.xgboost4j.scala.spark.XGBoost$.trainWithDataFrame(XGBoost.scala:199)` 

my spark version is 2.2.0

[...]

Issue #1927, opened 2017-01-03
[jvm-package] Add WatchList to xgboost4j-spark trainWithDataFrame / trainWithRDD 


mydpy:
For distributed training it would be useful to calculate a group of items to be evaluated during training to watch performance on the validation set. This is offered as a watch in single-node training. 

Limitations of metrics collection in Spark are severe. However, presently I will need to modify the XGBoost source code in order to add my own accumulators. It would be helpful to have a structure in place for me to define them in the training method. 

I suspect this is not feasible. If so, let's document that here.

xydrolase:
Since the model up to each iteration is synchronized to each distributed worker, so it is possible to compute metric on an evaluation set, so long as it is small enough so that it won't take too long to evaluate. 

I'm actually also interested in this, and I'll look into it and see if I can make a PR for this feature.

xydrolase:
@mydpy I'll start working on this feature this week. Still contemplating what's the best strategy to run evaluation. I can either let the user pass a RDD of evaluation set, so that each worker performs its own evaluation on a portion of the validation set; Or, use broadcast to send the whole evaluation set to the workers, but only let one to perform evaluation. Will experiment with some proof of concepts and post updates as I make progress.

buptzzl:
I am trouble with not watchlist in xgboost-spark,does any news come out?

CodingCat:
@xydrolase I am thinking about either use the original rabit to allreduce metrics, or use your akka-based rabit tracker (add some application layer message) + accumulator to achieve this....

scutlinhao:
Add "WatchList" to xgboost4j-spark， does any news come out now？

chaithu123:
Any update on the watchlist usage in xgboost4j-spark?

superbobry:
There's API to subsample the input DataFrame into train and test, see #2711.

wangzhilong:
@superbobry 
hi, i want to add "WatchList" to xgboost4j-spark, so i use latest master version。
when i run,  some error happened: 
`Exception in thread "main" java.lang.NoSuchMethodError: org.apache.spark.scheduler.LiveListenerBus.listeners()Ljava/util/concurrent/CopyOnWriteArrayList;
	at org.apache.spark.SparkParallelismTracker.safeExecute(SparkParallelismTracker.scala:75)
	at org.apache.spark.SparkParallelismTracker.execute(SparkParallelismTracker.scala:102)
[...]

PR #3325, opened 2018-05-17
[jvm-packages] Fix Spark 2.2 Support (Amending #3062)

thesuperzapper:
This pull request amends the broken #3062 allow Spark 2.2 to work.

Please note this won't work in Spark <=2.1 as sc.removeSparkListener was implemented in Spark 2.2. (So perhaps a more general method is better, although that is what was attempted in #3062)

This PR fixes: #3208, #3151 and the discussion in #1927.

I do find it strange that #3062 dose not work in Spark 2.2, it's probably due to some sort of public/private issue in the org.apache.spark.scheduler.LiveListenerBus class inheritance (In Spark itself). The error is: `java.lang.NoSuchMethodError: org.apache.spark.scheduler.LiveListenerBus.removeListener(Ljava/lang/Object;)V`

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
@@ -81,7 +81,7 @@ class SparkParallelismTracker(
     try {
       body
     } finally {
-      sc.listenerBus.removeListener(listener)
+      sc.removeSparkListener(listener)
     }
   }
 


CodingCat:
thanks for this PR which LGTM

before I merge, I would like to setup a TravisCI test matrix covering Spark 2.2 and 2.3, would you be interested to do that?

CodingCat:
example:https://github.com/graphframes/graphframes/blob/master/.travis.yml

thesuperzapper:
Could someone also try master with Spark 2.1, and see if this `java.lang.NoSuchMethodError` occurs.

CodingCat:
I do not think we will support Spark 2.1 as a too-old version

thesuperzapper:
@CodingCat LGTM? 

In its current state XGBoost-0.72 only supports Spark 2.3+, but with this PR we at least support Spark 2.2+.

An alternative is to do a bit of a hacky case statement and treat pre/post 2.2 separately, do you want me to do that?

With regards to the Travis test matrix, I don't have much experience with that, but its something we need to do.

thesuperzapper:
@CodingCat, this is still an issue on master, if we want 0.8 to support Spark 2.2 this should be merged, be aware that this PR breaks Spark <=2.1.

There is a small chance the existing (non PR) code works if you build XGBoost with 2.2 specified, but I can't get master to build on Spark 2.2 right now.

(I get these errors on build)
```
[INFO] Compiling 14 source files to /xgboost/jvm-packages/xgboost4j-spark/target/classes at 1533555864068
[ERROR] /xgboost/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala:43: error: trait HasWeightCol in package shared cannot be accessed in package org.apache.spark.ml.param.shared
[ERROR]   with BoosterParams with HasWeightCol with HasBaseMarginCol with HasNumClass with ParamMapFuncs
[ERROR]                           ^
[ERROR] /xgboost/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala:60: error: not found: value weightCol
[ERROR]   def setWeightCol(value: String): this.type = set(weightCol, value)
[ERROR]                                                    ^
[ERROR] /xgboost/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala:166: error: not found: value weightCol
[ERROR]     val weight = if (!isDefined(weightCol) || $(weightCol).isEmpty) lit(1.0) else col($(weightCol))
[ERROR]                                 ^
[ERROR] /xgboost/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala:166: error: not found: value weightCol
[ERROR]     val weight = if (!isDefined(weightCol) || $(weightCol).isEmpty) lit(1.0) else col($(weightCol))
[ERROR]                                                 ^
[ERROR] /xgboost/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala:166: error: not found: value weightCol
[ERROR]     val weight = if (!isDefined(weightCol) || $(weightCol).isEmpty) lit(1.0) else col($(weightCol))
[ERROR]                                                                                         ^
[ERROR] /xgboost/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressor.scala:44: error: not found: type HasWeightCol
[ERROR]   with LearningTaskParams with HasBaseMarginCol with HasWeightCol with HasGroupCol
[ERROR]                                                      ^
[ERROR] /xgboost/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressor.scala:62: error: not found: value weightCol
[ERROR]   def setWeightCol(value: String): this.type = set(weightCol, value)
[ERROR]                                                    ^
[ERROR] /xgboost/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressor.scala:160: error: not found: value weightCol
[ERROR]     val weight = if (!isDefined(weightCol) || $(weightCol).isEmpty) lit(1.0) else col($(weightCol))
[ERROR]                                 ^
[ERROR] /xgboost/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressor.scala:160: error: not found: value weightCol
[ERROR]     val weight = if (!isDefined(weightCol) || $(weightCol).isEmpty) lit(1.0) else col($(weightCol))
[ERROR]                                                 ^
[ERROR] /xgboost/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressor.scala:160: error: not found: value weightCol
[ERROR]     val weight = if (!isDefined(weightCol) || $(weightCol).isEmpty) lit(1.0) else col($(weightCol))
[ERROR]                                                                                         ^
[ERROR] 10 errors found
```

CodingCat:
@thesuperzapper I don't think 0.8 version would be support 2.2 anymore

because we used classes in `org.apache.spark.ml.param.shared` package which were still unreasonably limiting access in 2.3- versions

hcho3:
@CodingCat So 8.0 requires Spark 2.3+?

CodingCat:
yes

hcho3:
Got it. I'll modify http://xgboost.readthedocs.io/en/latest/jvm/index.html, which currently says "Spark 2.0+ required."

thesuperzapper:
@CodingCat in that case, can we merge this, as it won't affect 2.3, but it will allow 2.2 to work (mostly) with the jars built for 2.3.

CodingCat:
@thesuperzapper I don't think moving models to a new package is a good idea, it will break all existing code, 

I think a possible approach is to merge this PR to 0.72 branch which should work with Spark 2.2 without package namespace issue @hcho3 what do you think?

hcho3:
@CodingCat I like the idea. Can you update 0.72 JARs on Maven Central?

CodingCat:
@hcho3 , we cannot update jar for maven, we need to release a new version, either 0.72.1 (sounds strange) or 0.73

thesuperzapper:
@CodingCat @hcho3 just for reference, there is little point backporting this to 0.72, as if you build 0.72 with Spark 2.2 specified in the pom.xml,  the error doesn't happen, (so perhaps we should just publish a Spark 2.2 built jar to maven?)

What I was suggesting was that we merge this into 0.8, so that the 0.8 jars we build for Spark 2.3 work on Spark 2.2, as in my testing, models still successfully train/run on Spark 2.2 if this PR is applied.

hcho3:
@thesuperzapper You previously said that

> In its current state XGBoost-0.72 only supports Spark 2.3+, but with this PR we at least support Spark 2.2+.

So wouldn't backporting this to 0.72 have some benefit?

thesuperzapper:
> > In its current state XGBoost-0.72 only supports Spark 2.3+, but with this PR we at least support Spark 2.2+.

> So wouldn't backporting this to 0.72 have some benefit?

@hcho3 Yea, I was wrong, what I should have said was it lets the jars built for 2.3 work with 2.2. (I.e. the maven jars)

@CodingCat I agree that we should merge this and distributed jars built on 2.3, with a proviso that 2.2 may not behave as expected.

hcho3:
Since this commit is a one-liner, I'm going to go ahead and merge this. The JVM doc will still specify Spark 2.3+ as minimum requirement.

Issue #3322, opened 2018-05-15
Updated dmlc-core submodule blocks R CRAN submission


hetong007:
With https://github.com/dmlc/xgboost/pull/3221, we have `dmlc-core` updated.

This change updated several headers, thus introduces the following `NOTE` to the CRAN check:

```
  Files which contain pragma(s) suppressing diagnostics:
    'src/dmlc-core/include/dmlc/any.h'
    'src/dmlc-core/include/dmlc/concurrency.h'
    'src/dmlc-core/include/dmlc/concurrentqueue.h'
    'src/dmlc-core/include/dmlc/logging.h'
```

I'm submitting R package 0.71.1 to CRAN with the old `dmlc-core` headers. However it is just a one-time compromise. 

@hcho3  Do you see these pragmas temporarily or stable? Thanks.

hcho3:
@hetong007 The updated `dmlc-core` has an important usability improvement for R users (see #2849), so we should find a way to keep the latest `dmlc-core`.

As for the pragmas, here are the PRs that introduced them: [#236](https://github.com/dmlc/dmlc-core/pull/236), [#357](https://github.com/dmlc/dmlc-core/pull/357), [#243](https://github.com/dmlc/dmlc-core/pull/243). They all appear to be there to remove gratuitous compiler warnings. Does CRAN completely disallow removal of warnings?

hcho3:
@hetong007 And I understand that MXNet also has R binding, and MXNet I think has pretty recent version of `dmlc-core`. We should find out how MXNet-R gets around this problem.

hetong007:
I am with the latest `dmlc-core`, just the pragmas may block CRAN submission in the future. 

Now CRAN mark pragmas suppressing diagnostics as `NOTE`, but still won't allow a submission with this `NOTE`.

On the MXNet side, the R package has never been submitted to CRAN, since it actually has other severe issues (GPU compatibility, etc).

hcho3:
I see two ways to move forward:

1. Submit a PR to remove pragma's. This may increase gratuitous compiler warnings on some platforms.

OR

2. Write a script that removes pragma's before CRAN submission.

RAMitchell:
As an aside, will it be possible to submit GPU enabled binaries to CRAN? We are now producing linux binaries that can run seamlessly on GPU or non-GPU systems via Jenkins CI.

hetong007:
@RAMitchell To have a GPU version submitted to CRAN, we may need to:

[...]

PR #3329, opened 2018-05-17
For CRAN submission, remove all #pragma's that suppress compiler warnings

hcho3:
A few headers in dmlc-core contain #pragma's that disable compiler warnings, which is against the CRAN submission policy. Fix the problem by removing the offending #pragma's as part of the command `make Rbuild`.

This addresses issue #3322.


hcho3:
@hetong007 I got this message from `rmingw` test. Should I move the script?
```
* checking top-level files ... NOTE
Non-standard file/directory found at top level:
  'remove_warning_suppression_pragma.sh'
```

hetong007:
No need to move it, please instead edit [cleanup](https://github.com/dmlc/xgboost/blob/master/R-package/cleanup).

Issue #3302, opened 2018-05-08
Recipe for target 'lib/libxgboost.so' failed 


SachinKalsi:
I am getting the following error. Please help

```
makefile:173: recipe for target 'xgboost' failed 
make: *** [xgboost] Error 1 
make: *** Waiting for unfinished jobs.... 
Makefile:164: recipe for target 'lib/libxgboost.so' failed
```

## Environment info
Operating System: Ubuntu 16.04

Compiler: gcc 5.4

Package used (python/R/jvm/C++): python

`xgboost` version used: 0.71

If installing from source, please provide 

1. The commit hash (`git rev-parse HEAD`): b8a0d66fe6f9f8acc3b27afa81d2fc8ef074633d

. the gcc version and distribution: gcc 5.4

If you are using python package, please provide

1. The python version and distribution: 3.6.5 Anaconda
2. The command to install `xgboost` if you are not installing from source

## Steps to reproduce

1. git clone --recursive https://github.com/dmlc/xgboost
2. cd xgboost; make -j4

## What have you tried?

1. git clone --recursive https://github.com/dmlc/xgboost
2. cd xgboost; make -j4

hcho3:
Can you post the full compiler error? The message `recipe for target 'xgboost' failed` doesn't have enough information for us to act on it.

hcho3:
@SachinKalsi The `rabit` submodule ([dmlc/rabit](https://github.com/dmlc/rabit)) failed to compile because, on your system, the compiler flag `-msse2` is not supported. Are you using an ARM processor?

Currently, xgboost points to an old version of `rabit`, which had `-msse2` hard-coded in its build. Latest version of `rabit` does not have this problem. For now, you should update `rabit` by running
```
cd rabit
git pull origin master
```

hcho3:
Closing this as it is a duplicate of #3075, #3244.

SachinKalsi:
@hcho3 : No, I am using INTEL. I had to  remove `-msse2 ` option from the Makefile to make `xgboost` compile successfully. I also did try what you suggested & it was compiling successfully too.
Thanks for the heads up.

Issue #3244, opened 2018-04-13
Problem with hardcoded -msse2 compiler option in xgboost/rabit/


mjos:
Hi,

there is a problem with compilation of librabit on non-intel platforms (such as power linux on ppc64le) due to hardcoded usage of -msse2 compiler flag. On other places there is USE_SSE variable that controls this, but not in librabit Makefiles. If USE_SSE would be used, it would be possible  to compile the package, however it would be best to autodetect target platform (perhaps using -march=native ? I'm not expert in any way).

Thanks and best regards,

Michal

## Environment info
Operating System: RedHat 7.3 on ppce64le

Compiler: gcc 4.8.5-11

Package used (python/R/jvm/C++): python

`xgboost` version used: 0.7.post3

If you are using python package, please provide

1. The python version and distribution: python 3.6.4
2. The command to install `xgboost` if you are not installing from source
pip download -d pypi --no-binary :all: xgboost
pip install --no-index --find-links pypi xgboost

If you are using R package, please provide

1. The R `sessionInfo()`
2. The command to install `xgboost` if you are not installing from source

## Steps to reproduce

1. tar xzf xgboost-0.7.post3
2. cd xgboost-0.7.post3
3. xgboost/build-python.sh

## What have you tried?

1. export USE_SSE=0
2. xgboost/build-python.sh
3. Eventually I removed the hardcoded -msse2 options from Makefiles and built xgboost succesfully.

Laurae2:
`-mtune=native` would be more reasonable if the library is built to be used elsewhere.

RAMitchell:
There was a discussion about this flag in xgb here: https://github.com/dmlc/xgboost/issues/2446

I don't believe the flag is necessary to enable sse and could simply be removed. You could submit a PR for this in rabit.

mjos:
I'll prepare the PR. Thanks.

mjos:
It turned out that the problem was already solved in rabit by 41c96a25a9ae1886ca0de5a044d3e3cc2e5e9643 and 2eb1a1a371f41a0e0ef3127255aef4ea4f52c461. So the PR is just updating the rabit submodule to latestest version. See #3246. Is it ok like this?

Issue #3075, opened 2018-01-28
Xgboost compiles with old rabbit code without support for ARM causing failure due to msse2 flag


ParadoxShmaradox:
For bugs or installation issues, please provide the following information.
The more information you provide, the more easily we will be able to offer
help and advice.

## Environment info
Operating System:
RASPBIAN STRETCH

Compiler:
gcc (Raspbian 6.3.0-18+rpi1) 6.3.0 20170516

Package used (python/R/jvm/C++):
python
`xgboost` version used:
0.7.post3
If installing from source, please provide 

1. The commit hash (`git rev-parse HEAD`)
2. Logs will be helpful (If logs are large, please upload as attachment).
[xgboost_0.7.log](https://github.com/dmlc/xgboost/files/1671435/xgboost_0.7.log)

If you are using python package, please provide

1. The python version and distribution
2. The command to install `xgboost` if you are not installing from source
python 2.7

## Steps to reproduce

1. pip install xgboost==0.7.post3
2.
3.

## What have you tried?

1.
2.
3.

It seems like rabbit is failing to compile because of the -msse2 flag. I've looked at the latest code for rabbit and seems like support for the USE_SSE flag was added with [https://github.com/dmlc/rabit/pull/46](url)

Looks like xgboost code is still compiled with an older revision for rabbit without support for the flag.

headupinclouds:
This sounds similar to a `rabit` + `dmlc` header dependency mismatch I encountered in the current xgboost submodule sources, although I haven't looked at this particular issue to know if it is directly related.  These two submodules introduce conflicting header files, namely `dmlc-core/include/dmlc/base.h` and a mismatched bundled version of the same file in rabit here `rabit/include/dmlc/base.h`.  These have different `ANDROID` preprocessor logic (at least), that was causing some build problems for android toolchains. 

As one example:

https://github.com/dmlc/dmlc-core/blob/c553350aafc701a7a501d40f3445e9b5657812ad/include/dmlc/base.h#L44-L45
```
/*!
 * \brief Whether to print stack trace for fatal error,
 * enabled on linux when using gcc.
 */
#if (defined(__GNUC__) && !defined(__MINGW32__)\
     && !(defined __MINGW64__) && !(defined __ANDROID__))
#if (!defined(DMLC_LOG_STACK_TRACE))
#define DMLC_LOG_STACK_TRACE 1
#endif
#if (!defined(DMLC_LOG_STACK_TRACE_SIZE))
[...]

PR #3330, opened 2018-05-18
Point rabit submodule at latest commit from master.

### LINKED ISSUE #3328: Rabit C API pointed to by submodule in master and 0.71 missing recent fixes
I'm working on some new bindings to XGboost using the C API, and building them from a fresh checkout of XGBoost's `master` or `0.71` tag fails with:

    xgboost/rabit/include/rabit/c_api.h:35:1: error: unknown type name 'RABIT_EXTERN_C'
    xgboost/rabit/include/rabit/c_api.h:35:11: error: expected identifier or '('

On further investigation, this is because both `master` and the `0.71` tag contain `rabit` as a submodule pointing at https://github.com/dmlc/rabit/tree/a764d45cfb438cc9f15cf47ce586c02ff2c65d0f .

This is missing a lot of fixes that are in `rabit/master`, including this one (https://github.com/dmlc/rabit/commit/0b406754fa6c5d90e54d55f992045bde45f447ea#diff-d513b1282a73bf588d02ec8dd60216da), which fixes the build issue I'm having.

So I can fix the above by checking out XGboost, then doing a submodule update to pull the latest change from Rabit, then building, but it would be useful it builds worked from a fresh XGboost checkout (i.e. if the rabit submodule pointed to a more recent rabit branch/tag).

davechallis:
The rabit submodule pointed to a very old point in the rabit tree, so was missing quite a lot of fixes/improvements from the past year or two.

This was causing some attempts to build against XGBoost's C API to fail, without updating the submodule first. This commit means that building against the C API will succeed with a fresh checkout (fixes #3328).


hcho3:
Thanks for filing the PR.

@RAMitchell I noticed that in #3178 you updated `rabit` to a later version but not all the way to the latest. Were there any issues associated with latest `rabit`?

hcho3:
I'd like to merge this in, as it solves a host of issues: #3328, #3302, #3244, #3075.

PR #3136, opened 2018-02-26
[TREE] Interaction constraints

BlueTea88:
I would like to incorporate interaction constraints functionality in tree building.
It currently only supports the API in R and in the exact greedy algorithm on multi-core.

Please see #3135 for more information.

First time trying to contribute so let me know if there are any issues or you require more information.

PR #3335, opened 2018-05-23
Refactor of FastHistMaker to allow for custom regularisation methods

henrygouk:
Currently only FastHistMaker has been modified to use this new regularisation framework. I figured I should get some feedback before I put in the effort of refactoring the other updaters.

Future PRs will focus on novel regularisation methods.

**Things that need to be checked:**
- [x] L2 weight penalty is applied correctly
- [x] Monotonic constraints are applied correctly
- [x] Speed is not impacted noticeably

### DIFF
--- a/src/tree/split_evaluator.cc
+++ b/src/tree/split_evaluator.cc
@@ -6,6 +6,7 @@
 #include <dmlc/registry.h>
 #include "split_evaluator.h"
 #include "param.h"
+#include "../common/common.h"
 #include "../common/host_device_vector.h"
 
 namespace dmlc {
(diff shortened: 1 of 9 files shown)


hcho3:
Nice! I like how split evaluation logic is now being separated out from the updaters. There is a pending PR (#3136) to implement feature interaction constraints. Do you see a way to incorporate feature interaction constraints in the SplitEvaluator construct?

src/tree/split_evaluator.cc
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  bst_float ComputeSplitLoss(bst_uint parentID,
+                             bst_uint featureID,
+                             const GradStats& left,
+                             const GradStats& right) const override {
+    bst_float ret = 0.0f;
+
+    for(auto& e : m_inner) {
+      ret += e->ComputeSplitLoss(parentID, featureID, left, right);
```
RAMitchell: Could the composition multiple return values here cause NaN?

src/tree/split_evaluator.cc
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  SplitEvaluator* GetHostClone() const override {
+    RidgePenalty* r = new RidgePenalty();
+    r->m_params = this->m_params;
+
+    return r;
+  }
+
+  bst_float ComputeLoss(bst_uint parentID, const GradStats& stats) const override {
```
RAMitchell: Is this a loss function or a gain function?

henrygouk: This should be the loss function.

src/tree/split_evaluator.cc
henrygouk: I think this will only happen if one evaluator returns Inf and another returns -Inf, or one of them returns NaN. I can't think of a situation where a loss function would evaluate to -Inf, so this maybe isn't too much of an issue?

henrygouk:
@hcho3 It should be possible to incorporate that into this framework. The efficient way to do it would be to add another method to the SplitEvaluator that is called each time a split is applied to the tree, which would update some internal state of the SplitEvaluator. E.g., which features have already been split on.

BlueTea88:
I might be wrong, but it looks like the monotonicity constraints only holds at the local split level.  It might not guarantee monotonicity for subsequent splits of the same tree which the current code does - through lower bounds and upper bounds?

hcho3:
@henrygouk Are you dropping `AdditiveSplitEvaluator`? If so, we may be able to eliminate virtual function calls by having `SplitEvaluator` as a template argument to the tree updater. See `TConstraint` for an existing example. (In particular, `spliteval_` should be of type `SplitEvaluator`, not `std::unique_ptr<SplitEvaluator>.) Only downside is that the updater code gets duplicated by however many evaluators there are. If you'd like, I can propose a template version of your code and you can run some benchmark.

henrygouk:
If the code currently in master is using the templated approach would it be suitable to just compare the performance of this PR to master?

RAMitchell:
@hcho3 I prefer virtual functions unless there is a visible speed difference.

henrygouk:
using the benchmarking script, tests/benchmark/benchmark_tree.py, I've come up with these numbers

This PR:
tree_method=hist Train Time: 60.4643552303 seconds
tree_method=exact Train Time: 586.678107023 seconds

master:
tree_method=hist Train Time: 66.2623929977 seconds
tree_method=exact Train Time: 621.636813879 seconds

demo/binary_classification/mushroom.conf
```diff
@@ -4,6 +4,8 @@ booster = gbtree
 # choose logistic regression loss function for binary classification
 objective = binary:logistic
 
+# updater = "grow_fast_histmaker"
```
RAMitchell: Was this accidentally committed?

RAMitchell:
There are some failures on Windows relating to the dumped tree outputs: 
`` == testthat results  ===========================================================
  OK: 532 SKIPPED: 4 FAILED: 7
  1. Failure: xgb.dump works (@test_helpers.R#44) 
  2. Failure: xgb.dump works (@test_helpers.R#53) 
  3. Failure: xgb.model.dt.tree works with and without feature names (@test_helpers.R#213) 
  4. Failure: xgb.importance works with and without feature names (@test_helpers.R#239) 
  5. Failure: updating the model works (@test_update.R#32) 
  6. Failure: updating the model works (@test_update.R#41) 
  7. Failure: updating the model works (@test_update.R#64)```

You can find these tests here: https://github.com/dmlc/xgboost/tree/master/R-package/tests/testthat
Curiously these tests pass on Linux. 

If this can be solved I think this PR is mergable.

hcho3:
@henrygouk Curiously, the R-MinGW test fails for Windows R 32-bit. Investigating why.

hcho3:
@henrygouk @RAMitchell It turns out that the split evaluator produces a **substantially different** model when XGBoost is compiled for 32-bit instead of the usual 64-bit. I'm attaching the model dumps made on 32-bit and 64-bit.
[win32.txt](https://github.com/dmlc/xgboost/files/2129390/win32.txt)
[win64.txt](https://github.com/dmlc/xgboost/files/2129391/win64.txt)

henrygouk:
I think I've figured out the problem.

By changing all occurrences of 'bst_float' in split_evaluator.cc and split_evaluator.h to be 'double' I can get the R tests to pass when running in 32-bit mode. This is because the x87 floating point unit is used in the x86 ISA, whereas in AMD64 the scalar SSE instructions are used. Something that is kind of annoying is that all floating point arithmetic done with the x87 unit uses 80 bit registers. The values are then rounded when stored in the 32/64 float/double variables in memory. Therefore, expressions will carried out with much higher precision in 32-bit mode, and the precision loss will only occur when the values are stored in memory.

Changing the bst_float occurrences to double is not really a good solution to this problem. 64-bit floating point arithmetic will certainly be closer to the 80-bit arithmetic done using the x87 units, but it is still possible to get different answers. I think the correct way to solve this problem would be to find a way to build the 32-bit R library with SSE2 enabled, however I have absolutely no experience with building R bindings, so perhaps someone else should look into this.

hcho3:
> I think the correct way to solve this problem would be to find a way to build the 32-bit R library with SSE2 enabled.

I've added `-msse2` flag to Win32 R build. Hope it solves the problem.

hcho3:
@hetong007 I am worried that adding `-msse2` for Win32 would break installation for users of old machines that do not support SSE2. Are we committed to supporting machines without SSE2?

hetong007:
I'm not award of SSE2 related regulation from CRAN. Maybe it is good to test it through https://win-builder.r-project.org/.

The best practice could be that we by default compile with SSE2, and provide a switch in `Makevars.win` for users who want to compile without it.

hcho3:
@hetong007 It appears that most Windows machines today would have SSE2-capable CPUs (see [this article](https://www.gizmodo.com.au/2018/06/microsoft-quietly-drops-support-for-non-sse2-cpus-in-windows-7/) about Windows 7 dropping support for non-SSE2 machines).

How does a typical R user specify a switch for compilation? Does `install.packages` or `devtools::install()` let the user pass extra flags?

hcho3:
@henrygouk I rebased the whole branch against the latest master, so that the GPU tests would pass. Once all tests pass, I will merge the PR. Thanks for your contribution.

hetong007:
@hcho3 that sounds good still. I think we can always take a look back if that raise a significant issue.

hcho3:
@BlueTea88 @henrygouk Now that this PR has been merged, let's discuss how we can add feature interaction constraints.

BlueTea88:
@hcho3 Awesome, I can try drafting up a new method for interaction constraints in split_evaluator as @henrygouk mentioned.

PR #3338, opened 2018-05-24
Fix print.xgb.Booster for R

r-suzuki:
I always get "Handle is invalid!" message with the R package xgboost 0.71.1. I found that `valid_handle` is `TRUE` when `x$handle` is (a kind of ) `NULL`, but it seems to be the opposite.

### DIFF
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -37,11 +37,14 @@ xgb.handleToBooster <- function(handle, raw = NULL) {
 # Check whether xgb.Booster.handle is null
 # internal utility function
 is.null.handle <- function(handle) {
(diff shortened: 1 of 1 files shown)


r-suzuki:
I have also included a fix for `is.null.handle` to treat NULL handle.

If we load a model object from .RData file (not using `xgb.save`) the `handle` object is `NULL`. While `is.null.handle` is expected to return `TRUE` for `NULL` handle, `is.null(handle)` is never reached since it stops when checking the `class`. So I changed the sequence a bit.

So this PR includes:

- Fix `print.xgb.Booster` to properly check whether `handle` is `NULL` or not.
- Avoid stopping the `print` function when the `handle` is `NULL`

PR #3346, opened 2018-05-29
[jvm-packages] Fix checkpoint_interval parameter name

giurim:
The parameter in the param map should be checkpoint_interval instead of checkpointInterval to work correctly.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostDFSuite.scala
+++ b/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostDFSuite.scala
@@ -145,8 +145,10 @@ class XGBoostDFSuite extends FunSuite with PerTest with TableDrivenPropertyCheck
     assert(xgbEstimator.get(xgbEstimator.checkpointPath).get === "checkpointPath1" )
 
     val xgbEstimatorCopy = xgbEstimator.copy(ParamMap.empty)
-    assert(xgbEstimatorCopy.fromParamsToXGBParamMap("checkpoint_path").toString === "checkpointPath1")
-    assert(xgbEstimatorCopy.fromParamsToXGBParamMap("checkpoint_interval").toString.toDouble === 13.0)
+    assert(xgbEstimatorCopy
+      .fromParamsToXGBParamMap("checkpoint_path").toString === "checkpointPath1")
+    assert(xgbEstimatorCopy
+      .fromParamsToXGBParamMap("checkpoint_interval").toString.toDouble === 13.0)
   }
 
   test("eval_metric is configured correctly") {
--- a/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostDFSuite.scala
+++ b/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostDFSuite.scala
@@ -138,6 +138,17 @@ class XGBoostDFSuite extends FunSuite with PerTest with TableDrivenPropertyCheck
     assert(xgbEstimatorCopy.fromParamsToXGBParamMap("objective").toString === "binary:logistic")
   }
 
+  test("checkpoint parameters configured correctly") {
+    val xgbParamMap = Map("checkpoint_path" -> "checkpointPath1", "checkpoint_interval" -> "13")
+    val xgbEstimator = new XGBoostEstimator(xgbParamMap)
+    assert(xgbEstimator.get(xgbEstimator.checkpointInterval).get === 13.0)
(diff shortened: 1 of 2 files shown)


jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostDFSuite.scala
```diff
@@ -138,6 +138,19 @@ class XGBoostDFSuite extends FunSuite with PerTest with TableDrivenPropertyCheck
     assert(xgbEstimatorCopy.fromParamsToXGBParamMap("objective").toString === "binary:logistic")
   }
 
+  test("checkpoint parameters configured correctly") {
```
CodingCat: we generally do not recommend have a separate test case for parameters like checkpoint_interval, would you please have a test case which runs with a specified checkpoint_interval and check if the checkpoint file is eventually there?

CodingCat:
good catch! I left some comments here

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostDFSuite.scala
giurim: @CodingCat Thx, I changed the test.

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostDFSuite.scala
```diff
@@ -138,6 +141,27 @@ class XGBoostDFSuite extends FunSuite with PerTest with TableDrivenPropertyCheck
     assert(xgbEstimatorCopy.fromParamsToXGBParamMap("objective").toString === "binary:logistic")
   }
 
+  test("test checkpointing with dataframe-based training") {
```
CodingCat: would you update updateCheckpoint() method in CheckPointManager class with a switch controlling if it clears the previous version of checkpoint 

and then you disable checkpoint cleaning in the test, and count the files and check file names in this test?

giurim: As we do not have access to the CheckPointManager directly because it is instantiated inside the estimator, this would require to add a new xgboost parameter to switch off clearing previous checkpoint versions. I think it is a good idea as that is an actual feature we might want to use but I feel it would be cleaner to put that into another pull-request as this only intend to fix a typo in the parameter name. What do you think @CodingCat ?

CodingCat: I think it's fine to put this into the same PR, as the major point for this feature is to validate the correctness of checkpoint_interval

CodingCat: ping

CodingCat:
ping?

giurim:
Hi,

just to make sure, because it is quite a lot of work:
- You suggest to have a completely new xgboost parameter (it is not just a constructor param on CheckpointManager because we do not call the constructor directly) for testing which disables the deletion of previous checkpoint models. As far as I see to achieve this we should also change the checkpoint logic a bit or at least check we do not interfere with that by leaving previous models there.

Best,

Gyuri

CodingCat:
another way to avoid xgb parameter is that you have a companion object of CheckpointManager in which you have a variable to control whether it cleans previous cp file

and in test you just turn on/off that switch

giurim:
Can you provide a small example to understand how that would work?

CodingCat:
```scala
object CheckpointManager {
  private var cleanExpiredCP = true
}

class CheckpointManager {
    //...
   private[spark] def updateCheckpoint(checkpoint: XGBoostModel): Unit = {
    val fs = FileSystem.get(sc.hadoopConfiguration)
    val prevModelPaths = getExistingVersions.map(version => new Path(getPath(version)))
    val fullPath = getPath(checkpoint.version)
    logger.info(s"Saving checkpoint model with version ${checkpoint.version} to $fullPath")
    checkpoint.saveModelAsHadoopFile(fullPath)(sc)
    if (CheckpointManager.cleanExpiredCP) {
        prevModelPaths.foreach(path => fs.delete(path, true))
    }
  }

  //...
}

class TestSuite extends BeforeAndAfter {
  override def before() {
    CheckpointManager.cleanExpiredCP = true
  }

  test("test checkpoint manager") {
     CheckpointManager.cleanExpiredCP = false
     //...
  }
}


```

CDHales:
Ping @giurim - How's this coming along? I'm very interested.

CodingCat:
ping

CodingCat:
this PR is not valid anymore as we move to 0.8, thanks!

PR #3358, opened 2018-06-02
Update 0.8 version num

CodingCat:
@hcho3 I think in future, if you want, you can just go to jvm-packages/ and run `dev/change_version.sh 0.72-SNAPSHOT 0.80-SNAPSHOT` when you prepare the release so that we can merge the commit with the formal release one

### DIFF
--- a/jvm-packages/pom.xml
+++ b/jvm-packages/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>ml.dmlc</groupId>
     <artifactId>xgboost-jvm</artifactId>
-    <version>0.72-SNAPSHOT</version>
+    <version>0.80-SNAPSHOT</version>
     <packaging>pom</packaging>
     <properties>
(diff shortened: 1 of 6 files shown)


hcho3:
@CodingCat Got it. So the latest master should be labeled as `xxx-SNAPSHOT` where `xxx` is the next upcoming release?

CodingCat:
yes, that's right

pvillacorta-stratio:
Excuse me, I've just compiled the jvm package with `mvn -DskipTests=true -Dspark.version=2.2.0 package` but it is not creating the jar-with-dependencies file, only xgboost4j-spark-0.80-SNAPSHOT.jar. I remember to have created it without problems a month ago. Am I missing something? Thanks

jzhu-apixio:
@CodingCat, will 0.8.0 work with spark 2.3?  Thanks so much.

PR #3363, opened 2018-06-05
added python doc string for nthreads to dmatrix

ngoyal2707:


python-package/xgboost/core.py
```diff
@@ -254,6 +254,8 @@ def __init__(self, data, label=None, missing=None,
             Set names for features.
         feature_types : list, optional
             Set types for features.
+        nthread : integer, optional
+            Number of threads to use for loading data from numpy array.
```
RAMitchell: I believe -1 means use all threads. Maybe just add a note for this?

ngoyal2707: Sweet, added that, thanks.

Issue #3370, opened 2018-06-08
[jvm-packages] PySpark Support Checklist


thesuperzapper:
## Overview:
This is a meta issue for implementing PySpark support.

## Related PRs:

- https://github.com/dmlc/xgboost/pull/4656

## TODO:
- [ ] __Decide on the Python package name__
    - PR https://github.com/dmlc/xgboost/pull/4656 uses `sparkxgb` (which is pretty widely used at this point)
    - Alternativly, we could use: `spark-xgboost`, `pyspark-xgboost` or `pyspark-xgb`
- [ ] __Develop PySpark wrapper code__
   - PR https://github.com/dmlc/xgboost/pull/4656 pretty much has this covered
   - However, it probably needs to be rebased for XGBoost 1.5, as it was based on 0.9, lol!
- [ ] __Decide how to package the code__
   - PR https://github.com/dmlc/xgboost/pull/4656 embeds the PySpark API into the XGBoost spark jar
   - This is how the [Spark-NLP](https://github.com/JohnSnowLabs/spark-nlp) and [GraphFrames](https://github.com/graphframes/graphframes) PySpark packages work, so we can probably just start with that (rather than using PyPi)
- [ ] __Create a testing framework__
   - The tests can be put into a `run_tests.py` file under `./jvm-packages/xgboost4j-spark/src/main/resources/sparkxgb/` 
   - These tests can be run in CI by adding a new `spark-submit` command to [`./xgboost/blob/master/tests/ci_build/test_jvm_cross.sh#L38`](https://github.com/dmlc/xgboost/blob/master/tests/ci_build/test_jvm_cross.sh#L38)
- [ ] __Write API Documentation__
   - Since we are adding a whole new API to XGBoost, we will need to write some docs for the ["JVM Package" section](https://xgboost.readthedocs.io/en/stable/jvm/index.html) of the website
   - The source for these pages on the website is found at [`./doc/jvm/`](https://github.com/dmlc/xgboost/tree/master/doc/jvm)

CodingCat:
would you please send a WIP version of PR when you are doing the work so that we can discuss some problems undergoing if necessary

thesuperzapper:
@CodingCat there are a few other non-essential things I think we should support.
* We should fix `saveAsHadoopFile(` in the base Spark XGBoost API, so that there is at least rudimentary interoperablity with the Base Python XGBoost API. (See: #2480).
[...]

Issue #2480, opened 2017-07-03
[jvm-packages] Models saved using xgboost4j-spark cannot be loaded in Python xgboost


ssimeonov:
A similar problem was reported in [this issue](https://github.com/dmlc/xgboost/issues/1115), which was closed without any verification. The [page](http://xgboost.readthedocs.io/en/latest//tutorials/aws_yarn.html#analyze-the-model) cited as a reason for closing without verification claims there should be no problem, yet the claim flies in the face of multiple people having experienced the problem.

Here I'll attempt to provide specific steps to reproduce the problem based on the instructions for [using XGBoost with Spark](https://docs.databricks.com/_static/notebooks/xgboost.html) from Databricks. The steps should be reproducible in the [Databricks Community Edition](https://accounts.cloud.databricks.com/registration.html#signup/community).

The instructions in the Scala notebook work sufficiently well for `xgboostModel.save("/tmp/myXgboostModel")` to generate `/tmp/myXgboostModel/data` and `/tmp/myXgboostModel/metadata/part-00000` (and the associated `_SUCCESS` file) using [saveModelAsHadoopFile()](https://github.com/dmlc/xgboost/blob/d53534045984031cc8fc6ecea31ed33279ca63ad/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostModel.scala#L299) under the covers. 

The `data` file ([download it](https://www.dropbox.com/s/f83m76irgugna9j/myXgboostModel_data?dl=1)) is 90388 bytes in my environment and begins with `??_reg_??features??label?`. 

The metadata file is:

```json
[...]

PR #3376, opened 2018-06-11
[WIP] Initial PySpark Support

thesuperzapper:
This is a WIP PR for PySpark  (DO NOT MERGE UNTIL #3370 has been resolved)

This implements a few objects:
- XGBoostEstimator
   - A wrapper for "ml.dmlc.xgboost4j.scala.spark.XGBoostEstimator"
- XGBoostClassificationModel
   - An object intended to wrap "ml.dmlc.xgboost4j.scala.spark.XGBoostClassificationModel"
- XGBoostRegressionModel
   - An object intended to wrap "ml.dmlc.xgboost4j.scala.spark.XGBoostRegressionModel"
- XGBoostPipeline
   - A class derived from Spark's default Pipeline, which is able to save/load XGBoost pyspark estimators. 
- XGBoostPipelineModel
   - A class derived from Spark's default PipelineModel, which is able to save/load XGBoost pyspark models.

I have NOT included support for the following xgboost4j-spark features in this commit:
- custom_eval
- custom_obj
- numEarlyStoppingRounds
- trainTestRatio
- groupData

In terms of completeness, the estimator pipeline objects work as expected in my testing on Spark 2.2 with Python 2.7. (If #3325 is applied first)


haiy:
@thesuperzapper this PR is Great! But I think there are  few more questions to be clarified,
 - 1 as a xgb official support, perhaps it's better to support `ParamGridBuilder` which is a basic hyperparameter tuning and give an example
- 2 because xgb has its own summary function, maybe we can also support it in python 
- 3 `saveAsHadoopFile` is great to export the booster for analysis, can we support it either?
- 4 PEP8 is a general python coding style, so .

Besides above,  can we just create a file like `ml/dmlc/xgboost4j/scala/spark.py` and implement our
`XGBoostClassifier,XGBoostClassificationModel`  in it to avoid rewrite so many functions?

thesuperzapper:
@haiy 
1. This works already, but is not documented yet as this is only the initial commit.

2. Can you specify the specific summary function you want?

3. Yea, this is easy. However in my experimentation the model saved by saveAsHadoopFile is not interoperable with the Base XGBoost Python API (See: #2480), instead you have to call `.booster.saveModel("XXX")`, which dose not support Hadoop. Personally, I think we should fix saveModelAsHadoopFile. 
   * (In this vein, some proper interoperablity with the Python api would be awesome, e.g. we could have a `toPython(` method which returns the base XGBoost Python estimator/model)

4. Can you be specific around where I have violated PEP?

5. I am not sure how creating a file in `ml/dmlc/xgboost4j/scala/spark.py` will reduce the function overwriting. The approach I have taken here is very similar to H2O's sparkling water and GraphFrames, as the Python developer api for object writing was not properly expanded till Spark 2.3.
   * And for proper pipeline support, each object has to be able to save/load to disk.

thesuperzapper:
It's also worth noting that I have used a bit of a hack to bring the parameters from the Scala objects into these Python objects. I programmatically go through parameters found in the Scala objects, and create parameters in the python ones. (This is similar to [SPARK-10931](https://issues.apache.org/jira/browse/SPARK-10931), and is [found here ](https://github.com/thesuperzapper/xgboost/blob/f99a14269167f1d6a49661c0a3dee53b0cb73b94/jvm-packages/python/sparkxgb/xgboost.py#L31) in this PR.)

This means that every parameter in the Scala object is present in these Python ones.

jvm-packages/python/sparkxgb/util.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        Returns the full class name of the Java XGBoost class. In this case,
+        we prepend "ml.dmlc.xgboost4j.scala.spark" to the name of the class.
+        """
+
+        # Deal with Python locations for XGBoost objects.
+        if clazz.__name__ in ("XGBoostEstimator", "XGBoostClassificationModel", "XGBoostRegressionModel"):
+            java_package = clazz.__module__.replace("sparkxgb", "ml.dmlc.xgboost4j.scala.spark")
+
+        # Our Pipeline object calls down to Spark's Pipeline object.
```
haiy: Is it really necessary to separate each else branch with a comment? Or just give our logic note ?

jvm-packages/python/sparkxgb/xgboost.py
```diff
@@ -0,0 +1,254 @@
...
+                 # Column Params
+                 baseMarginCol="baseMargin", featuresCol="features", labelCol="label", predictionCol="prediction",
+                 weightCol="weight",
+
+                 # Booster Params
+                 base_score=0.5, booster="gbtree", eval_metric="error", num_class=2, num_round=2,
+                 objective="binary:logistic", seed=None,
+
+                 # Tree Booster Params
+                 alpha=0.0, colsample_bytree=1.0, colsample_bylevel=1.0, eta=0.3, gamma=0.0, grow_policy='depthwise',
```
haiy: Although the  grouped  category parameters seems easier for us, maybe  it's better to make  them a litter compact because they are class init arguments. Please correct me if I'm wrong.

jvm-packages/python/sparkxgb/xgboost.py
```diff
@@ -0,0 +1,254 @@
...
+    def __init__(self, java_model=None):
+        """
+        Override the __init__ from JavaModel.
+        """
+
+        super(XGBoostClassificationModel, self).__init__(java_model)
+        if java_model is not None:
+            # Get parameters only present in the model object.
+            self._create_params_from_java()
+
```
haiy: Does this blank line should be removed?

haiy:
hi @thesuperzapper , thanks for your answers!
as to `saveAsHadoopFile ` ,  can we rewrite a python version ?  I guess the bug is caused by `SparkContext`. 

Your hack for the parameter is fine.  I didn't get it through your comment .

haiy:
we can also add some examples now!

CodingCat:
hmmmm....can you rebase with spark_dev_do_not_delete branch (there is a huge refactor), will it make you have to redo a lot of work?

CodingCat:
@yanboliang would you please drive the review here, let's target it to 0.9

yanboliang:
Sure, I will take a look soon. Thanks.

ecampana:
I am using `xgboost4j-spark-0.80.jar` along with `xgboost4j-0.80.jar`, as well as the wrapper functions stored in `sparkxgb.zip` which is compatible with version 0.80 of xgboost. I am able to make predictions after fitting the data with a list of features and target value. I would like to now extract the information gain of each feature, but I have not managed to figure out a way to obtain these results. Would someone be able to point me in the right direction. Much appreciated!

```
prediction = ['prediction']

features = ['feature_1', 'feature_2',  'feature_3', 'feature_4']

vectorAssembler = VectorAssembler().setInputCols(features).setOutputCol('features')

xgboost = XGBoostClassifier(
    featuresCol='features',
    labelCol='label',
    predictionCol='prediction'
)

pipeline = Pipeline().setStages([vectorAssembler, xgboost])

model = pipeline.fit(train)

results = model.transform(test).select(features+prediction)

results.select('prediction').distinct().collect()
[Row(prediction=0.0), Row(prediction=1.0)]
```
I have tried accessing the below with no luck.
```
print(model.stages[1].nativeBooster.getFeatureScore.__doc__)

Help on class Booster in package ml.dmlc.xgboost4j.scala:

Booster implements scala.Serializable, com.esotericsoftware.kryo.KryoSerializable {
|  
|  Methods defined here:
|  
|  getFeatureScore(String) : Map
|  
|  ------------------------------------------------------------
|  Fields defined here:
|  
|  ------------------------------------------------------------
|  Internal classes defined here:
|  
}
```

jvm-packages/python/sparkxgb/xgboost.py
```diff
@@ -0,0 +1,210 @@
...
+        return py_stage
+
+
+class XGBoostClassifier(JavaParamsOverrides, JavaEstimator, JavaMLWritable, XGBoostReadable):
+    """
+    A PySpark implementation of ml.dmlc.xgboost4j.scala.spark.XGBoostClassifier.
+    """
+
+    @keyword_only
+    def __init__(self, alpha=0.0, baseMarginCol=None, baseScore=0.5, checkpointInterval=-1, checkpointPath='',
```
akhilbatra898: Does it support the addition of booster params other than the general params like interaction constraints and monotone_constraints?

The 0.8 snapshot of xgboost supports it. I am planning to use this wrapper for my project, if it supports addition of other params.

ecampana: Will this permit the extraction of information gain for each feature the model ends up using?

ecampana: Will this permit the extraction of information gain for each feature the model ends up using?

hcho3:
@CodingCat @yanboliang Can we review this PR at some point?

AugustKarlstedt:
+1

CodingCat:
see this in 1.0

ecampana:
@CodingCat do you mean to say that extracting feature gain is now supported in version 1.0?

CodingCat:
@ecampana ? I mean official support on pyspark

thesuperzapper:
@CodingCat, sorry I lost track of this a while ago, are you saying you have made some progress on this?

If not, we really need to:
*  _setup a unit testing framework for pyspark_ 
* setup the CI to package this
* look into using a more sensible way of saving the trained models to disk (I think there is a pyspark api for model persistence/writing after 2.3) [or make some helper scala classes]

I have a bit more time to work on this, so we should agree on an approach.

CodingCat:
@thesuperzapper I am just saying I finally find enough time to ensure that this can be released in XGBoost 1.0 may only have time to review 

it's super nice if you can still work on it 

all things you mentioned are very important

CodingCat:
@thesuperzapper are you going to resume the work here soon?

hcho3:
@CodingCat This PR has been around for a while. Can we merge this and mark it as an experimental feature?

thesuperzapper:
@CodingCat I am happy to resume work, but I think we should agree on how we will ship the PySpark code:

We have a few options: (We should probably do more than one)
* PyPi
* Conda
* Putting the python code in the jar (for --packages in spark-submit)
* Offering a zip file

CodingCat:
@hcho3 I would prefer make it as a milestone in 1.0 release

@thesuperzapper good question

my answer would be let's start from one, and I would choose pypi to keep consistent with the official way to install pyspark itself http://spark.apache.org/downloads.html (though we can also install pyspark conda)

jvm-packages/python/sparkxgb/xgboost.py
```diff
@@ -0,0 +1,210 @@
...
+        return self._set(**kwargs)
+
+    def _input_kwargs_processed(self):
+        """
+        Until consensus on parameter names can be achieved, we must rename kwargs which would break Python.
+        """
+
+        kwargs = self._input_kwargs
+        if "reg_lambda" in kwargs:
+            kwargs["lambda"] = kwargs.pop("reg_lambda")
```
trivialfis: `lambda` is actually an alias of `reg_lambda`.

CodingCat:
@thesuperzapper do you have more questions?

thesuperzapper:
I am going to make a new PR, as I have screwed this one up a bit. LOL

CodingCat:
@thesuperzapper any update on this?

thesuperzapper:
@CodingCat yep, still working on this in my down time, the only issue I ran into so far was that you will need to update your dev build image for the jars. I recommend you move to the new CI pipeline images here: https://github.com/dmlc/xgboost/blob/release_0.90/tests/ci_build/Dockerfile.jvm

I should have a new PR in a few days, but it probably wont be ready to merge straight away, because we will need lots of testing and docs (Aiming for 1.0). The main thing holding me up is trying to clean up model persistence to not need a custom pipeline object.

CodingCat:
What’s the problem with docker image? The image I used is with a very low version of glibc to ensure compatibility

thesuperzapper:
Yep, I agree with using an old GLIBC, [but this change was made recently](https://github.com/dmlc/xgboost/pull/4538), so we need to ensure GCC is newer than 5.0 (GLIBC can still be old if building on centos6)

CodingCat:
I see, thanks for raising it up, I will look into it to make image consistent in release and test env

trivialfis:
@CodingCat @thesuperzapper @hcho3 @RAMitchell Since PySpark support will add a new sklearn like interface, is it possible that we consider **not** exporting any non-pyspark related parameters to the interface and just accept an `**argv` dictionary?  We are moving to implement parameters validation in C++, albeit slowly, but I think it's possible that accepting a dictionary like `xgb.train` would make things easier.

thesuperzapper:
@trivialfis I am not sure that's a good idea if we want to keep interactions with core spark functionality. 
For example, we want to support things like ParamGridBuilder(), which allow grid searches of the parameter space.

Cant users just explode their own dictionary when they initialize the PySpark XGBoost model object?

thesuperzapper:
@CodingCat the issues I am running into:
* Similar to the spark-nlp package I think we should have the pyspark code in the spark jar (so we can have - --packages work, because the jar can act as a python zip), how should we do this?
  1. Place the pyspark code under src/main/resources 
  2. Somehow specify it in maven (I am not that versed in maven as I usually use SBT
* I am really not sure what the best approach to getting the package on PyPi is, how do we intergrate it with CI, or do we do it manually?

I am still doing a bit of cleanup on the actual wrapper too, and implementing some intergration tests.

PR #3377, opened 2018-06-11
Upgrade NCCL to 2.0

RAMitchell:
Disable NCCL from Jenkins build until binaries become available for Nvidia centos 6 images.


RAMitchell:
I am postponing the upgrade to 2.0 because there is no measurable performance improvement and Nvidia has put NCCL 2.0 installers behind a membership wall. This makes it impossible to script the install for CI or other purposes. Will revisit if this changes.

Issue #2561, opened 2017-07-28
Weighting instances in ranking model using xgboost


dengjn:
Does anyone know if I can weight instances in ranking model ('objective': 'rank:pairwise') in XGBoost? I know R GBM can only weight the whole query group instead of each instance within the query group. Does XGBoost also have this limitation? 

thanks,
jd

pommedeterresautee:
related to https://github.com/dmlc/xgboost/issues/2231

dengjn:
Thank you! But I mean how to use weight to calculate the loss function in ranking model, not use weight to calculate the evaluation metric such as mrr, ndcg. Does anyone have experience in this?

ngoyal2707:
@dengjn https://github.com/dmlc/xgboost/pull/3379 PR should solve this issue, which is now merged to master.

Issue #2460, opened 2017-06-27
Rank:pairwise  instance weight not supported?


shengyu-guo:
Hi,

I am trying to use lambdamart for pairwise rank.

Refer to here:https://github.com/dmlc/xgboost/blob/master/doc/input_format.md#instance-weight-file.  I can give the instance weight in xxx.weight file

After I set the weight for some instance, I found the model, however,  still look the same compared to the model trained by all instance weighted the same. 

I check out the code in rank_obj.cc and I do not find the weight is used to construct or set the pair.

So...Could you tell me the if the instance weight is supported for LambdaMart?

Thanks!

Attach the code I've checked:
```
// get lambda weight for the pairs
        this->GetLambdaWeight(lst, &pairs);
        // rescale each gradient and hessian so that the lst have constant weighted
        float scale = 1.0f / param_.num_pairsample;
        if (param_.fix_list_weight != 0.0f) {
          scale *= param_.fix_list_weight / (gptr[k + 1] - gptr[k]);
        }
        for (size_t i = 0; i < pairs.size(); ++i) {
          const ListEntry &pos = lst[pairs[i].pos_index];
          const ListEntry &neg = lst[pairs[i].neg_index];
          const bst_float w = pairs[i].weight * scale;
          const float eps = 1e-16f;
          bst_float p = common::Sigmoid(pos.pred - neg.pred);
          bst_float g = p - 1.0f;
          bst_float h = std::max(p * (1.0f - p), eps);
          // accumulate gradient and hessian in both pid, and nid
          gpair[pos.rindex].grad += g * w;
          gpair[pos.rindex].hess += 2.0f * w * h;
          gpair[neg.rindex].grad -= g * w;
          gpair[neg.rindex].hess += 2.0f * w * h;
}
...
class PairwiseRankObj: public LambdaRankObj{
 protected:
  void GetLambdaWeight(const std::vector<ListEntry> &sorted_list,
                       std::vector<LambdaPair> *io_pairs) override {}
};
```

ngoyal2707:
Thanks @shengyu-guo for reporting this issue, https://github.com/dmlc/xgboost/pull/3379 PR should solve it, which is now merged to master.

PR #3379, opened 2018-06-12
added code for instance based weighing for rank objectives

ngoyal2707:
This PR uses sample_weights per list instance for all the ranking objectives. This is somewhat important for some use cases where in production systems we want to give more weights for some sample instance compared to other. (In this case, sample instance is query entire document set).

Testing Done:
1) Used python wrapper to train on weighted loss and evaluate on weighted loss. Performs much better than just training on unweighted loss.
usage:
`train_data_xgboost.set_weight(train_pairs_weight)`

The weights need to be normalized externally as per current implementation as following:
```
train_pairs_weight /= train_pairs_weight.sum()
train_pairs_weight *= train_pairs_weight.shape[0]
```
which is essentially doing:
w_i = w_i*N / (sum (w_i))

If you think I should do the normalization inside the source code, let me know, I can easily change the PR with that.  

EDIT:
This PR should resolve both of following issues:
https://github.com/dmlc/xgboost/issues/2460
https://github.com/dmlc/xgboost/issues/2561


ngoyal2707:
Hey @hcho3 , @tqchen : wonder if you guys can take a look at this and give your opinions.

ngoyal2707:
@hcho3 Thanks for looking into the PR, as per your comments I have added changes with:

1) Put weight normalization inside the rank_obj.cc, so there's no need of externally normalize them.
2) Added tests showing that putting weights works with "rank:pairwise" objective.

Questions:
1) Is there any alternative to looping over whole dataset serially to find sum_weights inside GetGradients as it seems like it should be done only once, although I saw that other objectives were following the same  https://github.com/dmlc/xgboost/blob/master/src/objective/regression_obj.cc#L219 , so guess it's not too slow? 


Side Note:
1) I think there's a small bug (or weirdness) with the implementation of how rank_obj.cc creates pairs. It can technically create same pair twice as the way [loop](https://github.com/dmlc/xgboost/blob/master/src/objective/rank_obj.cc#L77) is written, it never checks for pair already added. I don't think it will make much difference, but do you think it is intentional vs needs to be corrected?

ngoyal2707:
@hcho3 your changes of doing weight normalization looks good to me. Can you please merge them to master now?

hcho3:
@ngoyal2707 I ended up removing my latest change. The normalization factor would need to be re-computed whenever the objective function object is re-used for a different dataset, and I could not find a good way to detect the change of dataset. For now, I think we should just compute the normalization factor to be on the safe side.

ngoyal2707:
@hcho3 I see, good point, sorry I didn't catch that. 
Can you please merge my original request then? I can look into the if I can move that to constructor as follow up refactor.

hcho3:
@ngoyal2707 Yes, I will merge it once all tests pass.

ejalonas:
One dumb question @ngoyal2707, can you clarify the structure of the `train_pairs_weight` in your initial example? Are those weights based on a group of results, or on each result within a group (e.g. if I have DMatrix with 5 groups of 10 results a piece for pairwise ranking, when I go to set the weights, is the structure of it an array of 5 floats (weighting per group), and array of 50 entries (weighting each result) or an array of len 5 where each entry is an array of len 10 (weighting each result again).

ngoyal2707:
Hey @ejalonas , In my implementation `train_pairs_weight` is  at group level. So for your example of `5` groups of `10` results each, you will need to give `5` weights. 

If you want individual instance based weight, you can use `rank:ndcg` objective with giving magnitude of `label` as your weights, although `ndcg` scaling from magnitude of `label` to `weight` is not linear.
Currently "rank:pairwise" objective ignores magnitude of `label.

ejalonas:
Thanks @ngoyal2707 ! I really appreciate your clarification and contribution to this repo.

sunlei198911:
@ngoyal2707 r  which release version will begin to support the group weight?

PR #3387, opened 2018-06-15
[jvm-packages] XGBoost Spark integration refactor

yanboliang:
Combine a bunch of PRs into one, to merge dev branch to master.


CodingCat:
There are several PRs merged to master before this one, would you check if this PR contains those fixes

yanboliang:
@CodingCat  Yes, I have rebased master.

CodingCat:
@yanboliang sure, thanks, looks like some newly added test were failed....@RAMitchell would you mind sharing some insights on this?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -489,12 +308,29 @@ private class Watches private(
 
 private object Watches {
 
+  def buildGroups(groups: Seq[Int]): Seq[Int] = {
+    val output = mutable.ArrayBuffer.empty[Int]
+    var count = 1
+    var i = 1
+    while (i < groups.length) {
+      if (groups(i) != groups(i - 1)) {
```
a-johnston: Maybe related to https://github.com/dmlc/xgboost/issues/3406. Since this is a sequence made from an iterable the `apply` happening twice in `(groups(i) != groups(i - 1))` could be linear time causing this method to be quadratic (thus blowing up for larger datasets)?

CodingCat: @yanboliang can you look at this, I think @a-johnston 's point is valid

and additionally, loading the whole group as Seq still makes it vulnerable to OOM

a-johnston: Anecdotally, @ngoyal2707 and I were checking this out earlier and it worked for smaller datasets (tens of thousands) but did not for larger (tens of millions) without ever hitting OOM although we could train data without groups on the same framework. Instead it just hung forever processing, although we never connected a debugger to check what impl is actually used. I think it might even be cleaner to just do something like
```scala
var lastGroup = -1
for (group <- groups) {
    if (lastGroup != group) {
        lastGroup = group
<etc>
```

yanboliang: @a-johnston You have a good point and I think your suggestion should be more optimized, Scala ```apply``` may have some performance issue. But the current code only cause time from O(N) to O(2N) at worst, it's still linear not quadratic(O(N^2)).
I will do some experiments in the following days for further check, if you found more clues, please feel free to share them. Thanks.

a-johnston: @yanboliang looking a bit closer, I'm pretty sure that `groups` here is a `scala.collection.immutable.Stream$Cons` which would lead to linear time apply and then quadratic overall (including the `while` loop). Linear is necessary with this approach since it's streaming over the entire column. While I generally prefer this approach compared to the old `groupData`, this is definitely a degradation.

Also if you're too busy to dig more into this, I can open a PR for this later today.

yanboliang: @a-johnston I see, ```Seq.apply``` is linear time, so overall will be quadratic. You suggested code could solve the performance issue as well. What about change the input format from ```Seq``` to ```Iterator``` to make it less vulnerable to OOM? Please feel free to open a PR for this.
BTW, the new approach is align with #2749 , if that PR get merged, we can switch the underlying implementation to leverage it. Thanks.

PR #3388, opened 2018-06-16
Implement GPU version of softmax.

trivialfis:
During the process, I moved test_multiclass_metric.cc. I think it's misplaced.

For simple benchmark, I modified `demo/gpu_acceleration/cover_type.py' to use the new gpu:multi:softmax. I only ran 30 rounds since I am using a laptop GPU. Here is my results:

Hardware: GTX960m, Intel 4720HQ
Platform: Fedora 27 x86-64
CUDA: 9.1

With multi:softmax:
GPU Training Time: 295.1220784187317 seconds

With gpu:multi:softmax:
GPU Training Time: 264.19690442085266 seconds

I think it's worthy. Tell me your thoughts. And how can I improve my code. :-)


trivialfis:
Hey @hcho3 , @tqchen : It's ready for review now. :-) (Sorry for bothering, I don't know how to mark something is ready with Github)

src/objective/multiclass_obj_gpu.cu
```diff
@@ -0,0 +1,245 @@
...
+    CHECK_NE(info.labels_.size(), 0U) << "label set cannot be empty";
+    CHECK(preds->Size() ==
+          (static_cast<size_t>(param_.num_class) * info.labels_.size()))
+        << "SoftmaxMultiClassObj: label size and pred size does not match";
+
+    const int nclass = param_.num_class;
+    size_t n = preds->Size();
+    auto ndata = static_cast<size_t>(n / nclass);  // number of data points
+
+    preds->Reshard(devices_);
```
RAMitchell: I see a potential error here e.g. if num_class is 7 and the first 3 predictions for a training instance end up on the first device and the last 4 predictions end up on the second device the result will be wrong.

This is not your fault we just did not think that far ahead with HostDeviceVector :)

cc @canonizer we could just add an optional integer parameter to the Reshard method indicating that data may not be separated below some granularity.

src/objective/multiclass_obj_gpu.cu
```diff
@@ -0,0 +1,245 @@
...
+      max_data_.Resize(ndata);
+
+#pragma omp parallel for schedule(static, 1) if (devices_.Size() > 1)
+    for (int i = 0; i < devices_.Size(); ++i) {
+      int d = devices_[i];
+      dh::safe_cuda(cudaSetDevice(d));
+      const int block = 256;
+      size_t n = preds->DeviceSize(d);
+      size_t ndata = n / nclass;
+
```
RAMitchell: These two smaller kernels could be written in very few lines using the dh::launch_n() primitive. Search the code base for examples on how to use this.

.gitignore
```diff
@@ -91,3 +91,9 @@ lib/
 metastore_db
 
 plugin/updater_gpu/test/cpp/data
+
```
RAMitchell: I think this file should not have been committed.

src/objective/multiclass_obj_gpu.cu
trivialfis: Currently I can't contribute to the anything about multi-gpu. Since, well, I don't have multi-gpu. Hence, like the regression case, it's marked not implemented. 
But I will fix other stuffs soon. Thanks for the review. :)

trivialfis: Hi. I tried to use lambda with `dh::LaunchN` inside the `Transform`, which generated the following error:
```error: The enclosing parent function ("Transform") for an extended  __device__ lambda cannot have private or protected access within its class```

There is a problem with closure. Is it really necessary to use this helper function? Making a work around might make things quirky.

In the last commit (except for further commit when fixing clang-tidy is needed), I tried to define BLK_SIZE to represent the default block size for CUDA in xgboost (which is 256). With that, although I didn't use dh helper function, information duplication should not be increased because of a new file. :-)  If you don't like it or want it to be in a separated PR, I can revert that commit.

RAMitchell: Let's keep the kblock changes for a different PR - though this change does seem reasonable to me. Note also that the kAlign parameter is not related to block size but is a number of bytes so probably shouldn't be assigned as kBlock. Using dh::LaunchN was more of a suggestion, what you have now is fine.

trivialfis: The last commit is removed.

Issue #3402, opened 2018-06-23
[jvm-packages]Keep hitting updater_histmaker.cc error using the latest 0.8-snapshot version on Spark


qwemicheal:
I am testing xgboost on Spark 2.3 using the latest 0.8-snapshot built from master.
After loading train data in libsvm format to Spark dataframe and start training ,I kept hitting an old issue frequently raised on 2016 and 2017 like [this one](https://github.com/dmlc/xgboost/issues/1869)
The error message is as the following:
<img width="1636" alt="screen shot 2018-06-22 at 6 13 08 pm" src="https://user-images.githubusercontent.com/14201792/41809237-d623defe-771c-11e8-8a83-7f71799f94c3.png">
I have tried with train data from 20mb to 400mb and different parameters. They all failed with the exact same issue as in the above picture.

I have no idea why this error would happen so I switch back to  version 0.72 and everything works fine.
I am guessing the [3301](https://github.com/dmlc/xgboost/pull/3301) pull actually cause the error since it modified the code calculating the hist.

For bugs or installation issues, please provide the following information.
The more information you provide, the more easily we will be able to offer
help and advice.

## Environment info
Operating System:
Ubuntu 16.0.4
Compiler:

Package used (python/R/jvm/C++):

`xgboost` version used:
0.8-snapshot
If installing from source, please provide 

1. The commit hash (`git rev-parse HEAD`) 
34e3edfb1a6b075a00b49972faada70e4dbf7648
2. Logs will be helpful (If logs are large, please upload as attachment).

If you are using jvm package, please 

1. add [jvm-packages] in the title to make it quickly be identified

2. the gcc version and distribution

If you are using python package, please provide

1. The python version and distribution
2. The command to install `xgboost` if you are not installing from source

If you are using R package, please provide

1. The R `sessionInfo()`
2. The command to install `xgboost` if you are not installing from source

## Steps to reproduce

1. download the libsvm file 
[sample4.libsvm.zip](https://github.com/dmlc/xgboost/files/2129865/sample4.libsvm.zip)

2. load it using spark 
3. train a model using spark's trainWithDataFrame api  like the following
[train.txt](https://github.com/dmlc/xgboost/files/2129877/train.txt)

## What have you tried?

1. Change the parameter,  doesn't work 
2.
3.

CodingCat:
@yanboliang would you please take a look?

RAMitchell:
[...]

PR #3408, opened 2018-06-26
Correct mistake from dmatrix refactor

RAMitchell:
Issue: #3402

I believe I found and fixed the mistake with this PR but am unable to reproduce. @qwemicheal Are you able to build from this commit and see if the error persists?


qwemicheal:
@RAMitchell 
I cloned your PR and rebuild, the error still exist.
screen shot 2018-06-26 at 2 40 59 pm

The full container std err output is as following. 

[full_containr_stderr.txt](https://github.com/dmlc/xgboost/files/2135867/full_containr_stderr.txt)

But your modification seems to be taken effect. Before the error would bump up in less than 10 or 20 iterations. This time it bump up until the 62 iteration.

RAMitchell:
Thanks I will look deeper.

Issue #3412, opened 2018-06-26
[jvm-packages] buildGroups is unnecessarily expensive


a-johnston:
Added in https://github.com/dmlc/xgboost/pull/3387, the method relies on a loop followed by two `Seq.apply` calls:
```scala
while (i < groups.length) {
    if (groups(i) != groups(i - 1)) {
...
```
which for lazily generated sequences (when I checked, `groups` was an instance of `scala.collection.immutable.Stream$Cons`) is quadratic. This behavior can be seen by comparing the running times of the following accesses:
```scala
val s = Stream.continually(0).toSeq
s(0)
s(1000000)
s(100000000)
```
(I ctrl-C'd before that last one finished on my machine)

The method `buildGroups` should do a single pass and not rely on index access.

CodingCat:
@a-johnston are you interested in filing a PR?

a-johnston:
Yep, in progress

PR #3413, opened 2018-06-26
[jvm-packages] Avoid use of Seq.apply in buildGroups

a-johnston:
Fixes https://github.com/dmlc/xgboost/issues/3412

Regarding the concern with memory usage in https://github.com/dmlc/xgboost/pull/3387, since Seq is lazy I don't think there's much overhead (and it definitely isn't loading the whole thing in memory, otherwise the indexing wouldn't be an issue) compared to using the iterator directly.

Manually ran and checked that it 1) had the same output and 2) completed for a large input


jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -311,15 +311,15 @@ private object Watches {
   def buildGroups(groups: Seq[Int]): Seq[Int] = {
```
CodingCat: good to know that iterator.toSeq actually returns a Stream!

thanks, will merge once Travis is done

PR #3414, opened 2018-06-27
Fix wget for google tests in CI

hcho3:
CI tests were failing because wget prompts "the user" for a response whenever the google test archive is already on the disk.

Fix: Use `-nc` option to skip download when the archive already exists


hcho3:
For an example, see https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-3335/44/pipeline/41.

PR #3424, opened 2018-06-29
Build universal wheels using GPU CI

RAMitchell:
@hcho3 where did we get to with uploading GPU enabled wheels to pypi? I will be releasing a blog post soon it would be nice to have this ready if possible.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -97,7 +97,7 @@ def buildPlatformCmake(buildName, conf, nodeReq, dockerTarget) {
         sh """
         ${dockerRun} ${dockerTarget} ${dockerArgs} tests/ci_build/build_via_cmake.sh ${opts}
         ${dockerRun} ${dockerTarget} ${dockerArgs} tests/ci_build/test_${dockerTarget}.sh
-        ${dockerRun} ${dockerTarget} ${dockerArgs} bash -c "cd python-package; rm -f dist/*; python setup.py bdist_wheel"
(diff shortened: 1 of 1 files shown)


hcho3:
@RAMitchell  The current CI already tests the universal wheels on Linux? Yes, I'll upload the wheels on PyPI this weekend. Since Mac and Windows are not supported, we'll have to name the wheels "manylinux".

hcho3:
And let us transition to a CI setup in the future where you have multiple slaves, Linux and Windows. This way, we can generate Windows wheels. Only complication is that NVidia docker is Linux only, so we'll have to write scripts specific to Windows

RAMitchell:
Sounds good. I did a bit of testing today, note that the 9.1 wheels throw a runtime error when GPU algorithms are called on systems with cuda 8.0/9.0. We should upload the 8.0 wheel for maximum compatibility, this works fine on all systems I tested.

This PR is basically just to make the CI artifacts directly usable on a python 3 system.

hcho3:
@RAMitchell Do we really need the `--universal` flag? It appears to cause test failure (https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-3424/1/pipeline). I think we can simply use the artifacts given by the latest successfully Jenkins test, e.g. https://xgboost-ci.net/job/xgboost/job/master/47/.

RAMitchell:
The test failures just due to a file having its name changed - I will fix this. You should be able to use the artifacts that already exist but you will have to rename them to work on python 3.  e.g. 

xgboost-0.72-py2-none-any.whl -> xgboost-0.72-py2.py3-none-any.whl

Otherwise pip thinks the py2 wheel is incompatible with python 3.

PR #3429, opened 2018-07-01
Add LASSO

henrygouk:
Just noticed that the previous PR for the SplitEvaluator stuff didn't incorporate LASSO---this should fix that.

I've also tweaked things to make it easier to apply multiple constraints, by using composition of SplitEvaluators. I'll need to run some benchmarks before this is ready to be merged.

### DIFF
--- a/tests/python/test_elastic_net.py
+++ b/tests/python/test_elastic_net.py
@@ -1,0 +1,47 @@
+import numpy as np
+import unittest
+import xgboost as xgb
+
+from numpy.testing import assert_approx_equal
+
+train_data = xgb.DMatrix(np.array([[1]]), label=np.array([1]))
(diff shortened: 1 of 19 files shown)


tests/python/test_elastic_net.py
```diff
@@ -0,0 +1,47 @@
+import numpy as np
+import unittest
+import xgboost as xgb
+
+from numpy.testing import assert_approx_equal
+
+train_data = xgb.DMatrix(np.array([[1]]), label=np.array([1]))
+
```
RAMitchell: This filename is a little confusing as we also have linear models in XGBoost. May be call it test_tree_regularization.py and test_alpha/test_lambda?

PR #3430, opened 2018-07-01
fixed MinGW missed dll

### LINKED ISSUE #977: The procedure entry point could not be located in the dynamic link library libxgboost.dll
I install the xgboost package in python on Windows based on the answer from Disco4Ever in [http://stackoverflow.com/questions/33749735/how-to-install-xgboost-package-in-python-windows-platform]
After I run *_import xgboost as xgb *_ in python 3.5, this error appears. 
error

StrikerRUS:
Fixed #977. Very old issue, but is still presented while xgboost installation via MinGW.


hcho3:
Thanks!

PR #3432, opened 2018-07-03
Fix bug of using list(x) function when x is string

wenduowang:
list('abcdcba') = ['a', 'b', 'c', 'd', 'c', 'b', 'a'] ---> resulting in duplicate values

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -738,7 +738,7 @@ class DMatrix(object):
         if feature_names is not None:
             # validate feature name
             if not isinstance(feature_names, list):
-                feature_names = list(feature_names)
+                feature_names = [feature_names]
             if len(feature_names) != len(set(feature_names)):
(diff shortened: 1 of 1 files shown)


python-package/xgboost/core.py
```diff
@@ -738,7 +738,7 @@ def feature_names(self, feature_names):
         if feature_names is not None:
             # validate feature name
             if not isinstance(feature_names, list):
-                feature_names = list(feature_names)
+                feature_names = [feature_names]
```
hcho3: I don't think this is the best way to solve the problem. I think a better way is to first check `isinstance(feature_names, STRING_TYPES)` and have a special handling logic for the single string case.

hcho3:
Thanks a lot for your pull request. We appreciate your contribution. Can you address the review comment?

wenduowang:
Hey! Thanks for your comment. I have committed a new change and hope it would fix the problem.

Issue #2967, opened 2017-12-22
get_uint_info - RuntimeError: expected float pointer


KOLANICH:
When I call ```get_uint_info("root_index")``` I get the error ```RuntimeError: expected float pointer```. It seems that python API doesn't have it implemented right, since it assumes it to be always float.

The fix is:

```python
npToPointerTMapping={
	np.float32: ctypes.c_float,
	np.uint32: ctypes.c_uint,
}
def ctypes2numpy(cptr, length, dtype):
    """Convert a ctypes pointer array to a numpy array.
    """
    if dtype not in npToPointerTMapping:
        raise RuntimeError('Supported  types:', list(npToPointerTMapping.keys()))
    if not isinstance(cptr, ctypes.POINTER(npToPointerTMapping[dtype])):
        raise RuntimeError('expected '+str(npToPointerTMapping[dtype])+' pointer')
    res = np.zeros(length, dtype=dtype)
    if not ctypes.memmove(res.ctypes.data, cptr, length * res.strides[0]):
        raise RuntimeError('memmove failed')
    return res
```

KOLANICH:
@tqchen, is it fixed?

hcho3:
Addressed in #3442.

KOLANICH:
Thank you.

PR #3442, opened 2018-07-05
Fix get_uint_info()

hcho3:
Addresses #2967.


tqchen:
please add a regression test to prevent it from happening again

hcho3:
Done. The added test will call get_float_info() and get_uint_info(). The test will succeed if it does not crash.

Issue #3053, opened 2018-01-21
suggestion: default to ntree_limit = best_ntree_limit for predictions in scikit-learn API


kodonnell:
This seems like a good default. The main reason I'm raising this instead is that it'd make it easier to use tooling like `RandomizedSearchCV` which use the default score method for tuning hyperparameters - and this will be the last iteration, not the early stopping iteration. 

Workaround for anyone interested in a quick hack:

```python
class EarlyStoppingXGBClassifier(xgboost.XGBClassifier):
    
    def predict(self, data, output_margin=False, ntree_limit=0):
        if ntree_limit > 0 or not hasattr(self, 'best_ntree_limit'):
            return super().predict(data, output_margin, ntree_limit)
        print('predicting with best_ntree_limit')
        return super().predict(data, output_margin, self.best_ntree_limit)
            
    def predict_proba(self, data, output_margin=False, ntree_limit=0):
        if ntree_limit > 0 or not hasattr(self, 'best_ntree_limit'):
            return super().predict_proba(data, output_margin, ntree_limit)
        print('predicting with best_ntree_limit')
        return super().predict_proba(data, output_margin, self.best_ntree_limit)
```

Happy to submit PR if thought a good idea ...

aldanor:
You should probably still respect `ntree_limit` parameter if it's explicitly passed in and is non-zero, otherwise check if `best_ntree_limit` is set (and is non-zero?). This way, it could be probably be included straight into `XGBClassifier`.

kodonnell:
> You should probably still respect ntree_limit parameter

Good point for generic use. I'll update code example.

kodonnell:
Reason for closing @tqchen ?

tqchen:
This issue was closed due to the current triaging, we try to aggressive close issues from now on so opened issues like this can get attended to instead of being forgotten A PR is welcomed!

hcho3:
Consolidating to #3439. A new issue should be opened if someone decides to actively work on implementing this feature.

kodonnell:
PR submitted @hcho3 @tqchen - up to you if you wish to reopen.

PR #3445, opened 2018-07-05
python sklearn api: defaulting to best_ntree_limit if defined, otherwise current behaviour

kodonnell:
As per #3053. Pretty straightforward, so no tests written ... and not sure if it's enough to warrant adding to CONTRIBUTORS.md?

cc @aldanor


kodonnell:
I don't think the failing build has anything to do with my code, given I changed so little.

hcho3:
@kodonnell It was a minor style issue. One of the tests examine whether the code passes a set of style guidelines. One of the guidelines is that no line should have trailing whitespaces.

hcho3:
Merged. Thanks!

PR #3446, opened 2018-07-05
Replaced std::vector with HostDeviceVector in MetaInfo and SparsePage.

canonizer:
- added distributions to HostDeviceVector
- using HostDeviceVector for labels, weights and base margings in MetaInfo
- using HostDeviceVector for offset and data in SparsePage
- other necessary refactoring


canonizer:
Yes, the vectors are currently copied between the host and device even if not modified.

If you want to avoid this, I can add const-only versions of methods that get host or GPU data. E.g., `ConstHostVector()`, `ConstDevicePointer()` etc.

RAMitchell:
This seems like a good idea, otherwise they will be constantly syncing.

canonizer:
Done.

trivialfis:
Is it worth to store some extra information in one of the GPU related class like number of SMs?  Currently all the grid stride functions just use one grid.

RAMitchell:
@canonizer I have rebased this. Can you please review before I merge? It was a complicated process so there are probably mistakes. I have a backup of your PR before the rebase if necessary.

RAMitchell:
A couple of GPU tests are currently failing because I removed the behaviour of "wrapping around" when a number of devices is specified that is greater than the number of available devices. This was to be consistent with what @trivialfis did in his PR. We need to decide if we are going to "wrap" the device ordinal or throw some error. I'm thinking we should throw an error because if the user selected multiple GPUs they probably did not intend for it to simply run and multiple blocks on the same GPU.

We could check for a sufficient available devices inside of the GPUDistribution or GPUSet.

Issue #3135, opened 2018-02-26
[New Feature] Interaction constraints


BlueTea88:
Hi, I would like to add interaction constraints functionality to tree building.  
 
Basically, this would constrain the combination of variables in each tree based on constraints specified by the user.  Initial variables will still get selected in a greedy fashion (best variable at the time will be chosen) but subsequent variables will be limited to variables that have permitted interactions with the initial variables.

Potential benefits include:
* Better predictive performance from focusing on interactions that work - whether through domain specific knowledge or algorithms that rank interactions
* Less noise in predictions
* More control to the user on what the model can fit (for example, the user may want to exclude some interactions even if they perform well due to regulatory constraints)

The idea is discussed briefly in the paper:
Delta Boosting Machine and its Application in Actuarial Modelling (Antonio et al., 2015) 
https://actuaries.asn.au/Library/Events/ASTINAFIRERMColloquium/2015/AntonioEtAlDeltaBoostingPaper.pdf

BlueTea88:
I have an experimental version in the repo below.  
https://github.com/BlueTea88/xgboost/tree/int_cont

If possible, I would love to merge it to this main repository (see #3136).

It has two arguments:
* `int_constraints_flag` - TRUE/FALSE whether to impose interaction constraints
* `int_constraints_list` - permitted interactions specified as a list, where each item of the list is a set of column names that represents one permitted interaction (all column names listed in the set can be interacted with each other)

As an example, please see:
https://github.com/BlueTea88/xgboost/blob/int_cont/R-package/demo/interaction_constraints.R

## Limitations
* Currently only supports the exact greedy algorithm on multi-core
* API only updated for R

cassieqiao:
I followed your example and I assume int_constraints_list = list(c('V1','V2'),c('V3','V4','V5')) means V1 can only interact with V2 and vice versa. Same idea for V3, V4 and V5. But I got results like below. Basically, the constraints are not applied. Can you clarify it? Thanks.

temp.int
[[1]]
[1] "V3" "V5"

[[2]]
[1] "V3" "V4"

[[3]]
[1] "V4" "V5"

[[4]]
[1] "V3" "V4" "V5"

[[5]]
[1] "V1" "V3" "V4" "V5"

[[6]]
[1] "V2" "V3" "V4" "V5"

[[7]]
[1] "V3" "V4" "V5" "V6"

[[8]]
[1] "V4" "V5" "V8"

[[9]]
[...]

PR #3448, opened 2018-07-05
[TREE] Interaction Constraints again

BlueTea88:
Trying to incorporate interaction constraints again as mentioned in #3135 and using the SplitEvaluator structure created from #3335.

Currently doesn't support both monotonic and interaction constraints at the same time.  Not sure if there is a more elegant way to do this than creating another SplitEvaluator for both monotonic and interaction constraints at the same time.  @hcho3 and @henrygouk, any thoughts on this?

There are also other parts that are not that elegant: 
- infers the number of features from int_constraints and nint_constraints rather than getting it from the data properties
- creates a vector to store features from previous splits rather than obtaining them from the tree directly

Haven't yet done testing that it actually works in R.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,8 @@ if(USE_AVX)
   add_definitions(-DXGBOOST_USE_AVX)
 endif()
 
+# enable custom logging
+add_definitions(-DDMLC_LOG_CUSTOMIZE=1)
 
 # compiled code customizations for R package
 if(R_LIB)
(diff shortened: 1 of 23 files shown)


henrygouk:
> Currently doesn't support both monotonic and interaction constraints at the same time. Not sure if there is a more elegant way to do this than creating another SplitEvaluator for both monotonic and interaction constraints at the same time.

Unfortunately I didn't think about this when I first implemented SplitEvaluator, but I've got a small tweak in #3429 that makes this much easier.

BlueTea88:
I stuffed up something when trying to rebase my branch.  Will try this again in another pull request.

PR #3456, opened 2018-07-07
[jvm-packages] disable booster setup for xgboost4j-spark

CodingCat:
as gbtree is the only booster type having distributed updater (hist_maker,prune) we should disable the configuration of these two parameters to avoid confusing users


RAMitchell:
How hard would it be to just make the linear booster distributed? The coordinate descent updater just needs to comunicate its gradients at each feature update.

CodingCat:
@RAMitchell , I think it's doable to have a distributed gblinear booster, but I do not have bandwidth to deal with it for now...if anyone make it, I can just allow booster to be gbtree or gblinear in Spark

PR #3404, opened 2018-06-25
Upgrading to NCCL2

teju85:
Note that as part of this process, the nccl submodule has now been removed.

PR #3460, opened 2018-07-09
Upgrade cuda version to 9.2 for CI workflows

teju85:
- Needed by the issue #3404
 - as v9.1 doesn't have a nccl2 release

JFYI, please refer to @hcho3's comment regarding possible caching behavior of Jenkinsfile here: https://github.com/dmlc/xgboost/pull/3404.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,7 +7,7 @@
 dockerRun = 'tests/ci_build/ci_build.sh'
 
 def buildMatrix = [
-    [ "enabled": true,  "os" : "linux", "withGpu": true, "withNccl": true,  "withOmp": true, "pythonVersion": "2.7", "cudaVersion": "9.1" ],
+    [ "enabled": true,  "os" : "linux", "withGpu": true, "withNccl": true,  "withOmp": true, "pythonVersion": "2.7", "cudaVersion": "9.2" ],
     [ "enabled": true,  "os" : "linux", "withGpu": true, "withNccl": true,  "withOmp": true, "pythonVersion": "2.7", "cudaVersion": "8.0" ],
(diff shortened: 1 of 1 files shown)


hcho3:
@teju85 I'll merge it as soon as the tests pass. Thanks!

teju85:
Hi @hcho3 ,
That's where I'm unsure if CI workflow would be testing the "right" things here! Is there a way to have CI workflow pick up the Jenkinsfile changes in this PR, rather than the cached one?

teju85:
Ok, confirmed, the jenkins check above still seems to pick up from v9.1
Ref: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-3460/1/pipeline

hcho3:
@teju85 I think this PR will pass all the tests, since this PR does not yet introduce NCCL2. I don't think caching is the issue; it is rather that, when testing PRs, Jenkins uses old version of `Jenkinsfile`. See my comment at [#3404 (comment)](https://github.com/dmlc/xgboost/pull/3404#issuecomment-401694505).

teju85:
So, does this mean, once this PR gets merged, the checks for the next set of PRs will use the updated Jenkinsfile? (this was something unclear to me in your comment in PR #3404)

hcho3:
> So, does this mean, once this PR gets merged, the checks for the next set of PRs will use the updated Jenkinsfile?

Yes. When the `master` gets updated, then future PRs will use the updated `Jenkinsfile`.

teju85:
Thanks Philip for the confirmation!

hcho3:
@teju85 Now that this has been merged, you should re-base #3404 against the latest `master` to make use of the updated Jenkinsfile.

PR #3464, opened 2018-07-11
[JVM-packages] Support single instance prediction.

yanboliang:
Support single instance prediction for ```XGBoostClassificationModel``` and ```XGBoostRegressionModel```.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
@@ -230,7 +230,10 @@ class XGBoostClassificationModel private[ml](
     this
   }
 
-  // The performance of single instance prediction is not ideal, use it carefully!
+  /**
+   * Single instance prediction.
+   * Note: The performance is not ideal, use it carefully!
+   */
   override def predict(features: Vector): Double = {
     import DataUtils._
     val dm = new DMatrix(Iterator(features.asXGB))
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
@@ -16,6 +16,7 @@
 
 package ml.dmlc.xgboost4j.scala.spark
 
+import scala.collection.Iterator
 import scala.collection.JavaConverters._
 import scala.collection.mutable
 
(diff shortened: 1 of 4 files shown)


jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
```diff
@@ -229,30 +230,25 @@ class XGBoostClassificationModel private[ml](
...
-  private def probability(features: Vector): Array[Float] = {
-    import DataUtils._
-    val dm = new DMatrix(scala.collection.Iterator(features.asXGB))
-    _booster.predict(data = dm, outPutMargin = false)(0)
-  }
-
+  // The performance of single instance prediction is not ideal, use it carefully!
   override def predict(features: Vector): Double = {
-    throw new Exception("XGBoost-Spark does not support online prediction")
+    import DataUtils._
```
CodingCat: can we logwarn here?

yanboliang: This function would be called for each instance, very frequently. I think logging inside would flood log files.

CodingCat: doing this? https://github.com/CodingCat/spark/blob/554cd391b3ddb5fb3f7c52950610e832ad40047b/core/src/main/scala/org/apache/spark/scheduler/LiveListenerBus.scala#L136-L147

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
```diff
@@ -229,30 +230,25 @@ class XGBoostClassificationModel private[ml](
...
-    _booster.predict(data = dm, outPutMargin = true)(0)
-  }
-
-  private def probability(features: Vector): Array[Float] = {
-    import DataUtils._
-    val dm = new DMatrix(scala.collection.Iterator(features.asXGB))
-    _booster.predict(data = dm, outPutMargin = false)(0)
-  }
-
+  // The performance of single instance prediction is not ideal, use it carefully!
```
CodingCat: I think we should write it as part of scaladoc of this function (in parent class maybe) so that it is visible from scaladoc

yanboliang: SG.

mengxr:
when is the next release?

CodingCat:
@hcho3  I think the next one would be Aug 1st?

hcho3:
@CodingCat Yes.

PR #3467, opened 2018-07-13
Updates for GPU CI tests

RAMitchell:


RAMitchell:
@canonizer I have changed two statically allocated HostDeviceVector back to std::vector because they sometimes throw exceptions when they get freed after the cuda runtime.

PR #3477, opened 2018-07-17
Implementation of hinge loss for binary classification

henrygouk:
Here is my first attempt at an implementation for the binary hinge loss.

I've only run one fairly restricted benchmark on the HIGGS dataset, where it seems to be more resilient to overfitting compared to binary:logistic when the learning rate is high. 

For objective="binary:hinge"

```
[995]	train-error:0.231613	test-error:0.241256
[996]	train-error:0.231603	test-error:0.241226
[997]	train-error:0.23159	test-error:0.241192
[998]	train-error:0.23157	test-error:0.241154
[999]	train-error:0.23156	test-error:0.241184
```

For objective="binary:logistic"

```
[995]	train-error:0.22295	test-error:0.24493
[996]	train-error:0.222938	test-error:0.244944
[997]	train-error:0.222906	test-error:0.244978
[998]	train-error:0.222878	test-error:0.245008
[999]	train-error:0.222864	test-error:0.245094
```

Both of these were done with a learning rate of 1.

Things that need to be done before merging:

- [ ] Add test case
- [ ] Ensure bindings to other languages are working


RAMitchell:
Nice. I am thinking about removing AVX in future versions. I am not sure of the additional code complexity is worth the gains. Might be better to implement a GPU version.

henrygouk:
Looks like the cpp tests fail when built with make, but pass when built with cmake. Any insight, @RAMitchell?

RAMitchell:
Just add a tag here: https://github.com/dmlc/xgboost/blob/dd07c25d1227f7ebfe596d048be3b62c6fb97369/src/objective/objective.cc#L33

PR #3481, opened 2018-07-17
Improved library loading a bit

KOLANICH:


python-package/xgboost/core.py
```diff
@@ -114,10 +114,16 @@ def _get_log_callback_func():
...
-    lib_path = find_lib_path()
-    if len(lib_path) == 0:
+    lib_paths = find_lib_path()
+    if len(lib_paths) == 0:
         return None
-    lib = ctypes.cdll.LoadLibrary(lib_path[0])
+    pathBackup = os.environ['PATH']
+    for lib_path in lib_paths:
+      try:
+        os.environ['PATH'] = pathBackup + os.pathsep + os.path.dirname(lib_path) # needed when the lib is compiled with non-system-available dependencies
```
hcho3: "needed when the lib is compiled with non-system-available dependencies"
Which dependency was causing issue?

KOLANICH:
libc from MinGW. I have manually added its dlls to the dir of xgboost.dll since putting them into path would break the apps dependent on other version of libc.

hcho3:
@KOLANICH Can you fix the indentation of core.py? Existing Python code uses 4 spaces for indent.

python-package/xgboost/core.py
```diff
@@ -114,10 +114,16 @@ def _get_log_callback_func():
...
-    lib_path = find_lib_path()
-    if len(lib_path) == 0:
+    lib_paths = find_lib_path()
+    if len(lib_paths) == 0:
         return None
-    lib = ctypes.cdll.LoadLibrary(lib_path[0])
+    pathBackup = os.environ['PATH']
+    for lib_path in lib_paths:
+        try:
+            os.environ['PATH'] = pathBackup + os.pathsep + os.path.dirname(lib_path) # needed when the lib is compiled with non-system-available dependencies
```
hcho3: Can we put the comment in another line to pass the pylint?

python-package/xgboost/core.py
```diff
@@ -114,10 +114,16 @@ def _get_log_callback_func():
...
+            lib = ctypes.cdll.LoadLibrary(lib_path)
+        except:
```
hcho3: Let's specify the type of exception we'd like to catch here. (If the type is not known, just use `except Exception:`.

python-package/xgboost/core.py
KOLANICH: we can

python-package/xgboost/core.py
KOLANICH: OK

PR #3493, opened 2018-07-19
Fix relpath in setup.py on Windows

### LINKED ISSUE #3480: path is on mount '***:', start on mount 'C:' when building python package on Windows on the drive different from the one python is installed
The bug is only relevant to the systems having Windows paths.

```
  File "setup.py", line 19, in <module>
    LIB_PATH = [os.path.relpath(libfile, CURRENT_DIR) for libfile in libpath['find_lib_path']()]
  File "setup.py", line 19, in <listcomp>
    LIB_PATH = [os.path.relpath(libfile, CURRENT_DIR) for libfile in libpath['find_lib_path']()]
  File "***\Anaconda3\lib\ntpath.py", line 586, in relpath
    path_drive, start_drive))
ValueError: path is on mount '***:', start on mount 'C:'
```

The bisection showed to 88488fdbb9d915e8cc166fd620fcd4c2fe4bd1f1 .

hcho3:
Fixes #3480.

### DIFF
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -16,7 +16,14 @@ libpath_py = os.path.join(CURRENT_DIR, 'xgboost/libpath.py')
 libpath = {'__file__': libpath_py}
 exec(compile(open(libpath_py, "rb").read(), libpath_py, 'exec'), libpath, libpath)
 
-LIB_PATH = [os.path.relpath(libfile, CURRENT_DIR) for libfile in libpath['find_lib_path']()]
+LIB_PATH = []
+for libfile in libpath['find_lib_path']():
+  try:
+    relpath = os.path.relpath(libfile, CURRENT_DIR)
+    LIB_PATH.append(relpath)
+  except ValueError:
+    continue
+
 print("Install libxgboost from: %s" % LIB_PATH)
 # Please use setup_pip.py for generating and deploying pip installation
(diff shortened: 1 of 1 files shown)


python-package/setup.py
```diff
@@ -16,7 +16,14 @@
 libpath = {'__file__': libpath_py}
 exec(compile(open(libpath_py, "rb").read(), libpath_py, 'exec'), libpath, libpath)
 
-LIB_PATH = [os.path.relpath(libfile, CURRENT_DIR) for libfile in libpath['find_lib_path']()]
+LIB_PATH = []
+for libfile in libpath['find_lib_path']():
+  try:
+    relpath = os.path.relpath(libfile, CURRENT_DIR)
+    LIB_PATH.append(relpath)
```
KOLANICH:
```python
break
```
?

hcho3: Why?

KOLANICH: Because [the first path seem to be the only path we need](https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/libpath.py#L24 ).

hcho3: I think `find_lib_path()` is guaranteed to return only existing library files. Okay then, I'll add `break` to prevent duplicates.

PR #3494, opened 2018-07-20
Added configuration for python into .editorconfig

KOLANICH:


.editorconfig
```diff
@@ -4,4 +4,8 @@ root = true
 charset=utf-8
 indent_style = space
 indent_size = 2
-insert_final_newline = true
\ No newline at end of file
+insert_final_newline = true
+
+[*.py]
+indent_style = space
+indent_size = 2
```
hcho3: Should be ``indent_size = 4``, since pylint uses this setting

PR #3498, opened 2018-07-21
Add total_gain and total_cover importance measures

### LINKED ISSUE #3484: Inconsistency of feature importance between Python and R implementations
I've noticed that the Python and R implementations of feature importance behave a bit differently.  The Python `get_score(importance_type='gain')` function returns each feature's *average* gain over all of its splits, while the R `xgb.importance` function's `"Gain"` column gives each's feature's *total* gain.

I'm not an expert, but intuitively, the total gain seems much more useful as a measure of importance.  (Would you prefer a feature that gets used a single time with gain 100; or 100 times, each with gain 10?)  I'd be happy to submit a PR introducing an `importance_type='total_gain'` argument to `get_score`.  What do you think?

For what it's worth, this is what I see in the git history:
* [Dec 28, 2014](https://github.com/dmlc/xgboost/commit/3694772bde99af6b7df88e849ee5e09acea83c12): R implementation is added, returning total gain; however, the documentation implies that it returns average gain.
* [Apr 21, 2016](https://github.com/dmlc/xgboost/pull/1135/commits/a1e2b0652f718e883a4e97cd9beaf1e750c1da7b): Python implementation is added, returning average gain.  (Perhaps based on the R documentation?)
* [Jan 21, 2017](https://github.com/dmlc/xgboost/commit/2b5b96d760b43a6d37bc605c8e2c34f8a0471dd2?diff=split): R documentation is updated to make it clear that it returns total gain.

Thanks!

jqmp:
Add `'total_gain'` and `'total_cover'` as possible `importance_type`
arguments to `Booster.get_score` in the Python package.

`get_score` already accepts a `'gain'` argument, which returns each
feature's average gain over all of its splits.  `'total_gain'` does the
same, but returns a total rather than an average.  This seems more
intuitively meaningful, and also matches the behavior of the R package's
`xgb.importance` function.

I also added an analogous `'total_cover'` command for consistency.

This resolves #3484.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -1329,17 +1329,23 @@ class Booster(object):
         """Get feature importance of each feature.
         Importance type can be defined as:
             'weight' - the number of times a feature is used to split the data across all trees.
-            'gain' - the average gain of the feature when it is used in trees
-            'cover' - the average coverage of the feature when it is used in trees
+            'gain' - the average gain across all splits the feature is used in.
+            'cover' - the average coverage across all splits the feature is used in.
+            'total_gain' - the total gain across all splits the feature is used in.
+            'total_cover' - the total coverage across all splits the feature is used in.
 
         Parameters
         ----------
         fmap: str (optional)
-           The name of feature map file
+           The name of feature map file.
+        importance_type: str, default 'weight'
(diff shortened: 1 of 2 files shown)


hcho3:
Thanks!

PR #3502, opened 2018-07-25
build uber jar with dependency by default, this is the behavior of ol…

eshengit:
…d release

building fat jar is the old behavior and its good to be consistent.


CodingCat:
no, currently we publish xgboost4j on maven, https://github.com/dmlc/xgboost/tree/master/jvm-packages#access-release-version

and we cannot publish uber jar to maven

PR #3514, opened 2018-07-26
[WIP] Expose group info to C API + other wrappers

hcho3:
As pointed out in https://discuss.xgboost.ai/t/jvm-packages-getting-group-for-custom-rank-objective-and-eval/89, it is currently not possible to extract group info from DMatrix outside of the core lib. This is because the C API does not have a function for getting group info. (Weirdly, there is a C API function for setting group info.)

TODOs:
- [ ] Add C++ test
- [ ] Expose the function to Python wrapper
- [ ] Expose the function to Java/Scala wrapper
- [ ] Expose the function to R wrapper

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -295,15 +295,29 @@ XGB_DLL int XGDMatrixSetUIntInfo(DMatrixHandle handle,
                                  const unsigned *array,
                                  bst_ulong len);
 /*!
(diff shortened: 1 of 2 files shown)


Helw150:
I'll do my best to take this over on #3566 - though it may be slow (and stupid looking 🤣) going as I figure out the tasks

hcho3:
@Helw150  Thanks a lot!

Issue #3026, opened 2018-01-10
save/load model still fails with "count:poisson"


j-mark-hou:
It seems like there's still some error with saving/loading poisson regression Booster objects.  Saving and loading it once works, but if you then save that again, and then try to load it a second time, it fails. Also, if you try and continue training with the model that you loaded the first time, it also fails.  However, predicting with the loaded model is fine.

I've only tested this for Python on Windows currently.

```
import xgboost as xgb
import numpy as np
import pickle as pkl
import os
import platform
print(platform.system(), platform.release())
print(sys.version)
print(xgb.__version__)
# Windows 2012ServerR2
# 3.6.2 |Anaconda, Inc.| (default, Sep 19 2017, 08:03:39) [MSC v.1900 64 bit (AMD64)]
# 0.7


# synthetic data
n = 100
d = 10
train_cutoff = int(.6*n)
val_cutoff = int(.8*n)
X = np.random.normal(size=(n,d))
y = np.random.randint(0,20, size=n)

# create dmatrices
dmat_train = xgb.DMatrix(X[:train_cutoff,:], y[:train_cutoff])
dmat_val = xgb.DMatrix(X[train_cutoff:val_cutoff,:], y[train_cutoff:val_cutoff])
dmat_test = xgb.DMatrix(X[val_cutoff:,:], y[val_cutoff:])

# 1. Linear regression: no problems saving and loading
#train
evallist = [(dmat_train, 'train'), (dmat_val, 'val')]
params = {'learning_rate':.3, 
          'eval_metric':'rmse', 
          'objective':'reg:linear',
         }
model1 = xgb.train(params, dmat_train, evals=evallist, num_boost_round=20, verbose_eval=1, early_stopping_rounds=20)

# save and load a bunch of times. 
file1 = '1.model'
file2 = '2.model'
model1.save_model(file1)
model2 = xgb.Booster(model_file=file1)
model2.save_model(file2)
model3 = xgb.Booster(model_file=file2)
# no errors

# 2. Poisson regression: breaks
#train
evallist = [(dmat_train, 'train'), (dmat_val, 'val')]
params = {'learning_rate':.3, 
          'eval_metric':'rmse', 
          'objective':'count:poisson',
         }
model1 = xgb.train(params, dmat_train, evals=evallist, num_boost_round=20, verbose_eval=1, early_stopping_rounds=20)
# save and load a bunch of times.  things break
file1 = '1.model'
file2 = '2.model'
model1.save_model(file1)
[...]

PR #3515, opened 2018-07-26
Fix model saving for 'count:possion': max_delta_step as Booster attribute

### LINKED ISSUE #3509: error occurred when calling copy function use loaded model
The code is as follows:
``` python
import xgboost as xgb
import numpy as np

class BoosterCountPoissonExample():
    def __init__(self):
        data = np.random.rand(5, 10)  # 5 entities, each contains 10 features
        label = np.random.randint(2, size=5)  # binary target
        self.dtrain = xgb.DMatrix(data, label=label)

    def get_model_joblib(self):
        print("get model by joblib")
        param = {'max_depth': 2, 'eta': 1, 'silent': 1, 'objective': 'count:poisson'}
        plst = param.items()
        model = xgb.train(plst, self.dtrain)
        model_cp_1 = model.copy()
        print("before dump model, copy")

        from sklearn.externals import joblib
        path_name = "/Users/Jane/Desktop/model.pkl"
        joblib.dump(model, path_name)

        model_load = joblib.load(path_name)
        model_cp_2 = model_load.copy()
        print("after dump model, copy")

        return model

test = BoosterCountPoissonExample()
test.get_model_joblib()
```

Before dump the model, copy function works well. However, If I dump and then load the model,  I will get the following message when calling copy:
AssertError:read can not have position excceed buffer length. 
Below is the screen shot.
image
Could you please help on this? Thanks a lot. :)

hcho3:
Fixes #3509 and #3026, where `max_delta_step` parameter gets lost during serialization / pickling.

Fix: save `max_delta_step` as a Booster attribute, which is guaranteed to be saved correctly.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -5,7 +5,6 @@ CheckOptions:
   - { key: readability-identifier-naming.TypeAliasCase,          value: CamelCase  }
   - { key: readability-identifier-naming.TypedefCase,          value: CamelCase  }
   - { key: readability-identifier-naming.TypeTemplateParameterCase,          value: CamelCase  }
-  - { key: readability-identifier-naming.LocalVariableCase,          value: lower_case  }
   - { key: readability-identifier-naming.MemberCase,          value: lower_case  }
   - { key: readability-identifier-naming.PrivateMemberSuffix,           value: '_'  }
   - { key: readability-identifier-naming.ProtectedMemberSuffix,           value: '_'  }
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -25,7 +25,7 @@ namespace {
 
 const char* MAX_DELTA_STEP_DEFAULT_VALUE = "0.7";
 
-}  // namespace anonymous
+}  // anonymous namespace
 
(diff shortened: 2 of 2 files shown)


hcho3:
@RAMitchell Can you look at https://travis-ci.org/dmlc/xgboost/jobs/408745349? I get Clang-tidy error
```
/home/travis/build/dmlc/xgboost/src/learner.cc:43:8: warning: invalid case style for local variable '__i0' [readability-identifier-naming]
```
even though my code contains no such variable. Is there a way to disable it?

RAMitchell:
I think it's related to the global constant defined in learner.cc.

These are the rules that govern global constants: https://github.com/dmlc/xgboost/blob/e2f09db77a4ea1f759f9be585ab01a320d933900/.clang-tidy#L15

So maybe try camel case with k prefix:
const char* kMaxDeltaStepDefaultValue = "0.7";

RAMitchell:
I'm out of ideas you may have to remove the lint rule for local variables.

PR #3525, opened 2018-07-28
Enable building with sanitizers.

trivialfis:


hcho3:
I'm curious, in what situations would sanitizing XGBoost be useful?

trivialfis:
@hcho3
The idea of address sanitizer originates from @RAMitchell . Adding others as options is just convenience. The output from sanitizers shows some interesting warning, like memory leak and race conditions. I think it will be useful to help directing the development. :) WDYT?

hcho3:
@trivialfis I like the idea. Can you clarify what packages and tools that need to be installed to enable this feature?

trivialfis:
@hcho3 I added 3 CMake scripts finding those packages in cmake/modules. One for each sanitizer.
Sanitizers by itself are bundled in GCC or clang/llvm. But distributions might package them separately.

libasan for address sanitizer.
liblsan for leak sanitizer.
libtsan for thread sanitizer.

You can use above sanitizers with GCC >= 4.8 or LLVM >= 3.1.

Memory sanitizer is LLVM exclusive, hence not included in this PR.

hcho3:
@trivialfis Let's put this information in the documentation as well.

hcho3:
LGTM

PR #3527, opened 2018-07-29
Fix typo in ElasticNet threshold function

jqmp:
I think the `ThresholdL1` function in `split_evaluator.cc` has a typo: for small values of `g`, it makes more sense to return `0` rather than `g`.  This is what the version of `ThresholdL1` in `param.h` does.  (In fact, maybe it would be simplest to just call that one directly?)

But I could be missing something -- @henrygouk is the original author so maybe he can correct me.


henrygouk:
Yes, it should be returning zero in that case.

I think a sensible long term goal is to refactor the different updaters to use the SplitEvaluator functionality, which will result in the ThresholdL1 function being removed from `param.h` at some point.

PR #3529, opened 2018-07-30
[jvm-packages] the current version of xgboost does not consider missing value in prediction

CodingCat:
this PR is to fix it

@yanboliang @superbobry would you please help to review?


CodingCat:
thanks

PR #3532, opened 2018-07-30
[jvm-packages] enable predictLeaf/predictContrib/treeLimit in 0.8 

CodingCat:
missing the test cases

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
@@ -142,6 +142,12 @@ class XGBoostClassifier (
 
   def setCustomEval(value: EvalTrait): this.type = set(customEval, value)
 
+  def setLeafPredictionCol(value: String): this.type = set(leafPredictionCol, value)
+
+  def setContribPredictionCol(value: String): this.type = set(contribPredictionCol, value)
+
+  def setTreeLimit(value: Int): this.type = set(treeLimit, value)
+
   // called at the start of fit/train when 'eval_metric' is not defined
   private def setupDefaultEvalMetric(): String = {
     require(isDefined(objective), "Users must set \'objective\' via xgboostParams.")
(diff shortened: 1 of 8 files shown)


hcho3:
@CodingCat Should we shoot for merging this PR before releasing 0.80? If we need to, we can delay 0.80 release a bit.

CodingCat:
@yanboliang would you please take a look?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
```diff
@@ -313,7 +315,79 @@ class XGBoostClassificationModel private[ml](
...
+  }
+
+  private def produceResultIterator(
+      originalRowItr: Iterator[Row],
+      rawPredictionItr: Iterator[Row],
+      probabilityItr: Iterator[Row],
+      predLeafItr: Iterator[Row],
+      predContribItr: Iterator[Row]): Iterator[Row] = {
+    // the following implementation is to be improved
+    if (isDefined(leafPredictionCol) && isDefined(contribPredictionCol)) {
```
yanboliang: I have a little concern that users may set ```leafPredictionCol``` with ```""```, their intention should be not to output this column, but here we don't handle this case. At MLlib, we check ```$(leafPredictionCol).isEmpty``` as well. What do you think about it?

CodingCat: yes, good point, let me revise it

CodingCat:
@yanboliang please check the latest update

yanboliang:
Looks good, it seams the failure is not relevant.

hcho3:
I fixed the error in the Jenkins build. It should build successfully now.

CodingCat:
@hcho3 I merged this and cherry-picked to release_0.80 branch

hcho3:
@CodingCat Okay, I will do a code freeze using this commit. The branch will be re-based against the master at this commit.

As for the doc building, I'm going to place the build script on the CI server. The built doc will be placed to an S3 bucket, which will be then downloaded by ReadTheDocs. Expect a commit by end of today.

PR #3534, opened 2018-07-30
[jvm-packages] Tutorial of XGBoost4J-Spark

CodingCat:
joint work of @yanboliang  and @CodingCat


hcho3:
Thanks so much for the tutorial! This should be useful for new users. If you have a rendered output of the tutorial page, can you post it here?

CodingCat:
@hcho3 what do you mean by rendered output ? like HTML files?

hcho3:
@codingcat Actually never mind, I'll just render it on my side. Thanks again for your contribution.

hcho3:
Currently, I notice that `index.md` is under `xgboost4j-spark/docs/`. Is it common to put docs under JVM package directory? With current configuration, `index.md` will not be rendered along with the rest of documentation under `doc/`. This is an issue, since xgboost.readthedocs.io will only host files under `doc/`.

Also, let's consider using reST(restructuredtext) instead of Markdown. The former is similar to the latter, but the former gets better support by ReadTheDocs Sphinx. (ReadTheDocs renderer often botch up tables and relative links in Markdown.)

CodingCat:
I see, let me move that

CodingCat:
@hcho3 any recommended editor for rst? really hard to edit and view......

hcho3:
@CodingCat You can locally build the doc with `make html`. To display the built doc on browser, you can use https://github.com/GaretJax/sphinx-autobuild. (This package conveniently refreshes on every change made to each file.)

hcho3:
@CodingCat If you'd like, I can do the rst conversion for you.

CodingCat:
@hcho3 sure, thanks so much, that would be a great help if you can give a hand on this

hcho3:
@CodingCat Should we mention that it is possible to save models to HDFS?

doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -0,0 +1,473 @@
...
+A data scientist produces an ML model and hands it over to an engineering team for deployment in a production environment. Reversely, a trained model may be used by data scientists, for example as a baseline, across the process of data exploration. So it's important to support model persistence to make the models available across usage scenarios and programming languages.
+
+XGBoost4j-Spark supports saving and loading XGBoostClassifier/XGBoostClassificationModel and XGBoostRegressor/XGBoostRegressionModel. It also supports saving and loading a ML pipeline which includes these estimators and models.
+
+We can save the XGBoostClassificationModel to file system:
+
+.. code-block:: scala
+
+  val xgbClassificationModelPath = "/tmp/xgbClassificationModel"
+  xgbClassificationModel.write.overwrite().save(xgbClassificationModelPath)
```
CodingCat: @hcho3 I think we may not need to explicitly say we can save in HDFS, as we rely on Spark's ML framework to do that (as here), so when user use Spark (with a proper hadoop configuration), it's automatically to be HDFS file

hcho3: I noticed that the user needs to compile with USE_HDFS=ON to use nativeBooster.saveModel(). Can we make a note? Or alternatively, we can alter the JARs in your maven repo to support HDFS out of box.

CodingCat: good point, we may mention it in the section discussing how to interact across language bindings

doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -0,0 +1,478 @@
...
+
+Then we can load this model with single node Python XGBoost:
+
+.. code-block:: python
+
+  import xgboost as xgb
+  bst = xgb.Booster({'nthread': 4})
+  bst.load_model(nativeModelPath)
+
+When interacting with other language bindings, XGBoost also supports saving-models-to and loading-models-from file systems other than the local one:
```
CodingCat: @hcho3 please check this part

doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -286,16 +286,22 @@ Then we can load this model with single node Python XGBoost:
   bst = xgb.Booster({'nthread': 4})
   bst.load_model(nativeModelPath)
 
+When interacting with other language bindings, XGBoost also supports saving-models-to and loading-models-from file systems other than the local one:
+
+* You can build XGBoost4J-Spark with the steps described in `here <https://xgboost.readthedocs.io/en/latest/jvm/index.html#installation-from-source>`_, but turning `USE_HDFS <https://github.com/dmlc/xgboost/blob/e939192978a0c152ad7b49b744630e99d54cffa8/jvm-packages/create_jni.py#L18>`_ (or USE_S3, etc. in the same place) switch on. With this approach, you can reuse the above code example by replacing "nativeModelPath" with a HDFS path.
+
+However, if you build with USE_HDFS, etc. you have to ensure that the involved shared object file, e.g. libhdfs.so, is put in the LIBRARY_PATH of your cluster. To avoid the complicated cluster environment configuration, we recommend download file manually from HDFS and load with the pre-built version of XGBoost.
```
hcho3: What do you mean by "download file manually from HDFS"? Do you mean that the user should stick with native paths?

hcho3: I see what you mean here. I think you should also XGBoost4J needs USE_HDFS=ON to run `xgbClassificationModel.nativeBooster.saveModel()`. So USE_HDFS=ON is needed for both XGBoost4J-Spark and Python sessions.

CodingCat: yes, that's what the above two paragraphs for `With this approach, you can reuse the above code example by replacing "nativeModelPath" with a HDFS path.`

hcho3: I'm confused about the recommendated action, since running
```
local_file_path = download_hdfs(hdfs_path)
xgb.load(local_file_path)
```
won't solve the problem of XGBoost4J-Spark unable to save to HDFS to begin with.

CodingCat: I added description about how to save a pure booster to HDFS

doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -290,7 +290,20 @@ When interacting with other language bindings, XGBoost also supports saving-mode
 
 * You can build XGBoost4J-Spark with the steps described in `here <https://xgboost.readthedocs.io/en/latest/jvm/index.html#installation-from-source>`_, but turning `USE_HDFS <https://github.com/dmlc/xgboost/blob/e939192978a0c152ad7b49b744630e99d54cffa8/jvm-packages/create_jni.py#L18>`_ (or USE_S3, etc. in the same place) switch on. With this approach, you can reuse the above code example by replacing "nativeModelPath" with a HDFS path.
 
-However, if you build with USE_HDFS, etc. you have to ensure that the involved shared object file, e.g. libhdfs.so, is put in the LIBRARY_PATH of your cluster. To avoid the complicated cluster environment configuration, we recommend download file manually from HDFS and load with the pre-built version of XGBoost.
+However, if you build with USE_HDFS, etc. you have to ensure that the involved shared object file, e.g. libhdfs.so, is put in the LIBRARY_PATH of your cluster. To avoid the complicated cluster environment configuration, we recommend the following steps to pass models with HDFS, S3, etc. (taking HDFS as an example)
+
+* create a new file with **val outputStream = fs.create("hdfs_path")**
```
hcho3: Does `fs` here refer to an object of `org.apache.hadoop.fs.FileSystem`?

CodingCat: I think it's OK to have this assumption here.....

doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -290,7 +290,20 @@ When interacting with other language bindings, XGBoost also supports saving-mode
...
+
+* pass the returned OutputStream in the first step to **xgbClassificationModel.nativeBooster.saveModel(outputStream)**
+
+* download file in other languages from HDFS and load with the pre-built (without the requirement of libhdfs.so) version of XGBoost:
+
+.. code-block:: python
+
+  import xgboost as xgb
+  bst = xgb.Booster({'nthread': 4})
+  local_path = download_from_hdfs("hdfs_path")
```
hcho3: In which package do we get `download_from_hdfs()`?

doc/jvm/xgboost4j_spark_tutorial.rst
hcho3: Let's add a sentence about using a binding of Hadoop then. Something like

... we recommend that you use the language-specific binding of Hadoop as follows:
```
private val fs = {
    val conf = new Configuration()
    FileSystem.get(conf)
  }
val outputStream = fs.create("hdfs_path")
```

doc/jvm/xgboost4j_spark_tutorial.rst
CodingCat: no....it's a dummy helper function which is supposed to be implemented by the user

doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -0,0 +1,505 @@
...
+
+  There is a consistency issue between XGBoost4J-Spark and other language bindings of XGBoost.
+
+  When users use Spark to load training/test data in LIBSVM format with the following code snippet:
+
+  .. code-block:: scala
+
+    spark.read.format("libsvm").load("trainingset_libsvm")
+
+  Spark assumes that the dataset is using 1-based indexing (feature indices staring with 1). However, when you do prediction with other bindings of XGBoost (e.g. Python API of XGBoost), XGBoost assumes that the dataset is using 0-based indexing (feature indices starting with 0). It creates a pitfall for the users who train model with Spark but predict with the dataset in the same format in other bindings of XGBoost. The solution is to transform the dataset to 0-based indexing before you predict with, for example, Python API.
```
hcho3: > The solution is to transform the dataset to 0-based indexing before you predict with, for example, Python API.

Now that #3546 has been merged, do we want to mention the ability to import LIBSVM files with 1-based indexing? You add an `indexing_mode` parameter to the path as follows:
```
train.libsvm?indexing_mode=1
```

CodingCat: @hcho3 please check the latest update

hcho3:
LGTM. Let's merge it once tests pass.

CodingCat:
@hcho3 something screwed in python CI environment? 
image

hcho3:
@CodingCat Apparently one of the tests were throwing `std::bad_alloc` for the lack of memory. The test has passed now.

hcho3:
@CodingCat It's merge now. BTW, can you grant me access to the API docs (Java/Scala) so that they can be brought up to date?

CodingCat:
I think those stuffs are uploaded to https://github.com/dmlc/dmlc.github.io/tree/master/docs/scaladocs which I do not have write access either.....

hcho3:
Is there a command to build Javadoc/scaladoc? Let's host them on ReadTheDoc together with other docs so that they stay current.

CodingCat:
how about let me try in the weekend, there is a bunch of plugin change and profile isolation in pom.xml

not sure if the previous command works

PR #3535, opened 2018-07-31
Fix #3402: wrong fid crashes distributed algorithm

### LINKED ISSUE #3402: [jvm-packages]Keep hitting updater_histmaker.cc error using the latest 0.8-snapshot version on Spark

hcho3:
The bug was introduced by the recent DMatrix refactor (#3301). It was partially fixed by #3408 but the example in #3402 was still failing. The example in #3402 will succeed after this fix is applied.

TODO: Add a regression test.


hcho3:
@RAMitchell Any good idea for generating an example? It should be an example that triggers the bug.

RAMitchell:
Maybe it is appropriate in this case to add the file, it's not a blocking issue. In python I would generally try to use sklearn data generators or random numbers to generate a matrix reproducing the problem otherwise we could add a lot of data files over time to the repo.

hcho3:
@RAMitchell I went ahead and replace the regression test with a single Python file with the minimal example. The example fails without this PR but succeeds with the PR.

PR #3539, opened 2018-07-31
Use __CUDA__ macro with __NVCC__.

trivialfis:
* \_\_CUDA\_\_ is defined in clang. Making the change won't make clang
compile xgboost, but syntax checking for *.cu from clang is at least partially
working.

### DIFF
--- a/include/xgboost/base.h
+++ b/include/xgboost/base.h
@@ -69,7 +69,7 @@
 /*!
  * \brief Tag function as usable by device
  */
-#ifdef __NVCC__
+#if defined (__CUDA__) || defined(__NVCC__)
 #define XGBOOST_DEVICE __host__ __device__
 #else
 #define XGBOOST_DEVICE


hcho3:
Can you elaborate your reasoning for replacing `__NVCC__` with `__CUDA__`? Does this improve compatibility?

trivialfis:
@hcho3 To some degree. __CUDA__ is a macro defined by clang to indicate the current scope, __NVCC__ is a similar macro defined by NVCC. As you can see from the patch, it's not really replacing, just checking for the right one to use. With this patch clang can compile a larger portion of the code, but still not the whole project.

But the reason for myself is I can use better syntax checking and code static analyse tools from clang. Now I can see error/warning and do code navigation for CUDA interactively in my editor just like normal c++ code. NVCC provides very primitive error handling facility and confusing error message.

So the patch is mainly for easy development. :)

RAMitchell:
I'm not sure I like substituting macros with more macros. It can be a little confusing. What about doing this:

`#if defined(__NVCC__) || defined(__CUDA__)`

trivialfis:
@RAMitchell Done.

PR #3548, opened 2018-08-02
Span class.

trivialfis:
Implement span class from ISO++20 for safer pointer indexing.

TODOs:
- [x] Fix convert and container constructor.
- [x] Fix as_bytes.
- [x] Structure tests for CUDA tests.
- [x] Fix errors in test farm.
- [x] Make an example use in xgboost.
- [x] Make it work with MSVC.

### DIFF
--- a/src/common/span.h
+++ b/src/common/span.h
@@ -1,0 +1,633 @@
+/*!
+ * Copyright 2018 XGBoost contributors
+ * \brief span class based on ISO++20 span
+ *
+ * About NOLINTs in this file:
+ *
+ *   If we want Span to work with std interface, like range for loop, the
+ *   naming must be consistant with std, not XGBoost. Also, the interface also
+ *   conflicts with XGBoost coding style, specifically, the use of `explicit'
+ *   keyword.
+ *
+ *
+ * Some of the code is copied from Guidelines Support Library, here is the
+ * license:
+ *
+ * Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+ *
+ * This code is licensed under the MIT License (MIT).
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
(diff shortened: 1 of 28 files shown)


src/common/span.h
```diff
@@ -0,0 +1,524 @@
...
+  XGBOOST_DEVICE const_reverse_iterator crbegin() {
+    return const_reverse_iterator(cend());
+  }
+  XGBOOST_DEVICE const_reverse_iterator crend() {
+    return const_reverse_iterator{cbegin()};
+  }
+
+  XGBOOST_DEVICE reference operator[](index_type idx) const {
+    SPAN_CHECK(idx >= 0 && idx < size());
+    SPAN_CHECK(idx >= 0 && idx < size());
```
RAMitchell: Duplicate check here.

tests/cpp/common/test_span.cc
```diff
@@ -0,0 +1,418 @@
...
+
+TEST(Span, TestStatus) {
+  int status = 1;
+  TestTestStatus {&status}();
+  ASSERT_EQ(status, -1);
+}
+
+TEST(Span, Constructors) {
+  // Dynamic extent
+  {
```
RAMitchell: These two tests are identical.

tests/cpp/common/test_span.cc
```diff
@@ -0,0 +1,418 @@
...
+  }
+}
+
+TEST(Span, FromNullPtr) {
+  // dynamic extent
+  {
+    Span<float> s {nullptr, static_cast<Span<float>::index_type>(0)};
+    ASSERT_EQ(s.size(), 0);
+    ASSERT_EQ(s.data(), nullptr);
+
```
RAMitchell: More duplicated tests.

trivialfis:
@RAMitchell  I will re-visit all test cases, see what's been missing.
For the as_bytes, it did came across my mind. Maybe one needs to serialize a segment of memory, like a trained model? I don't know, so obliging to the standard is just me making a safe bet.

trivialfis:
Not ready yet, seems tests is still not thorough enough.

RAMitchell:
@hcho3 as noted by @trivialfis there is actually an array_view class in dmlc-core. After a brief look it seems it is used by the nvvm/tv projects. We could replace array_view with our span which is more powerful,  better tested and compliant with the upcoming standard. What do you think?

hcho3:
@RAMitchell You should file a pull request at dmlc-core and make a case for upgrading array_view with the span class. There seems to be a good amount of code using array_view already and it would take some work to replace all occurrences.

For the sake of lesser friction, you could simply add the span class to dmlc-core and have DMLC projects gracefully adopt it.

trivialfis:
@RAMitchell Ready for another review. The Travis failed fetching packages in Python test, it should be fine on a good day.

The last commit is just a toy show case for using Span in kernel code. It could be reverted for a more formal integration.

RAMitchell:
Looks good. One more thing to consider, span will incur a performance overhead when using the operator[], these changes will have worsened performance in places like this: https://github.com/dmlc/xgboost/blob/a96039141a8dd2d053ae8f415c8fc8e52223216e/src/data/simple_dmatrix.cc#L61

This can be fixed by changing this to a range based for loop, I believe this does not need to do any index checking.

RAMitchell:
DVec can be dealt with later. My idea is that we can gradually enforce the usage of span across the code base using the clang-tidy linter. The most important algorithm for us is gpu_hist as it is used a lot in production, so maybe we can start there when you are ready.

I think this should be good to merge when the CI tests pass.

trivialfis:
@RAMitchell So it passed. \0/

PR #3551, opened 2018-08-04
[jvm-packages] update java/scala docs

CodingCat:
please merge https://github.com/dmlc/xgboost/pull/3532 first


CodingCat:
@hcho3

CodingCat:
@hcho3 , if we build the doc on the fly, the doc for the released version would be replaced by the latest version for every update of the doc

this case would be tricky when we have some broken update of the API

hcho3:
@CodingCat Would you be okay with it if we maintain multiple versions, e.g. latest, 0.80, 0.81, etc.

CodingCat:
@hcho3 I am fine with maintaining multiple versions of API doc, 

is ReadTheDocs using `make html` to build the doc.....or I should just tell you manual steps, that will save much more time for us?

hcho3:
@CodingCat Can you provide me with the commands? I'd like to get it working before the release of 0.80.

CodingCat:
@hcho3 it's a multi-stepped process:

## Generate scala doc

1. run `mvn scala:doc -DskipTests` on the path of xgboost/jvm-packages

2. the above step would failed on xgboost4j-example but it does not matter

3. then you need to copy `xgboost4j/target/site/scaladocs/*`, `xgboost4j-spark/target/site/scaladocs/*`, `xgboost4j-flink/target/site/scaladocs/*` to different places and modify index.rst to create hyperlinks to index.html of each package

## Generate java doc

1. run ` mvn javadoc:javadoc`

2. the above step would failed on xgboost4j-example but it does not matter

3. then you need to copy `xgboost4j/target/site/apidocs/*` to a folder and modify index.rst to create hyperlinks to its index.html

PR #3557, opened 2018-08-05
Add travis sanitizers tests.

trivialfis:
Early PR. I need to see how's travis reacting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,8 @@ env:
     - TASK=cpp_test
     # distributed test
     - TASK=distributed_test
+    # address sanitizer test
+    - TASK=sanitizer_test
 
 matrix:
   exclude:
@@ -43,6 +45,8 @@ matrix:
       env: TASK=cpp_test
     - os: osx
       env: TASK=distributed_test
+    - os: osx
+      env: TASK=sanitizer_test
 
 # dependent apt packages
 addons:
@@ -62,6 +66,8 @@ addons:
       - graphviz
       - gcc-4.8
       - g++-4.8
+      - gcc-7
+      - g++-7
 
 before_install:
   - source dmlc-core/scripts/travis/travis_setup_env.sh
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,9 +14,10 @@ option(USE_NCCL "Build using NCCL for multi-GPU. Also requires USE_CUDA")
 option(JVM_BINDINGS "Build JVM bindings" OFF)
 option(GOOGLE_TEST "Build google tests" OFF)
 option(R_LIB "Build shared library for R package" OFF)
-option(USE_SANITIZER "Use santizer flags" OFF)
 set(GPU_COMPUTE_VER "" CACHE STRING
   "Space separated list of compute versions to be built against, e.g. '35 61'")
+option(USE_SANITIZER "Use santizer flags" OFF)
+option(SANITIZER_PATH "Path to sanitizes.")
 set(ENABLED_SANITIZERS "address" "leak" CACHE STRING
   "Semicolon separated list of sanitizer names. E.g 'address;leak'. Supported sanitizers are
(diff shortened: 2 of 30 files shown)


hcho3:
I suspect that old logic for locating GTest is lingering around somehow. The new logic should produce
```
-- Found GTest: /workspace/gtest/lib/libgtest.a 
```
whereas the old logic would produce
```
-- Found the GTest includes: /workspace/gtest/include
-- Found the GTest library: /workspace/gtest/lib/libgtest.a
```

Re-basing this PR against the latest upstream master in an attempt to be rid of the old logic.

trivialfis:
@RAMitchell Just tested asan on my machine with gcc-4.9 with Guix (where has no CUDA), which has libasan-1.0. In short, it doesn't work.

To confirm the testxgboost did linked against libasan, do:
``` bash
ASAN_OPTIONS=verbosity=1 ./testxgboost
```
I saw outputs like:
> ==22882==Parsed ASAN_OPTIONS: verbosity=1
 ==22882==AddressSanitizer: libc interceptors initialized
...

But no error is generated. So I think the problem is indeed on asan.

I attached the complete log for running ./testxgboost built by gcc-7, generated from:
``` bash
export ASAN_SYMBOLIZER_PATH=$(which llvm-symbolizer)
ASAN_OPTIONS=symbolize=1 ./testxgboost > log 2>&1
```
You can check the errors, see if it's worthy to upgrade Travis.

[log.txt](https://github.com/dmlc/xgboost/files/2271826/log.txt)

hcho3:
@trivialfis Thanks a lot for introducing me to sanitizers. I used address sanitizer to locate a double-free error in MXNet. This is super cool!

trivialfis:
@hcho3 Nah. Idea of putting it in XGBoost came from @RAMitchell while I did the leg work.  I guess he had enough with these bugs in XGBoost.
I am still trying to make it to Travis, lets wait and see.

trivialfis:
Voilà. asan on Travis sorted out.

Other than asan, the thread sanitizer reviews loads of data races, many of them relate to openmp. Some of them might be false positive since tsan still on beta, but seem pretty scary.  

Running thread sanitizer would take much longer than running address sanitizer (so long that I never cross the c_api.XGDMatrixCreateFromMat_omp test), and requires a lot more memory too. So it might not be a good fit for Travis.

@hcho3 , @RAMitchell  WDYT?

RAMitchell:
Very nice! Some of the data races may actually be intentional - for example the shotgun linear models solver uses hog wild parallelism. Can you post the log for tsan?

Let's try to fix some of these leaks in this PR.

It might also be interesting to run the python tests, although they are probably too time consuming to run on Travis.

trivialfis:
@RAMitchell The attached tsan log is generated without CUDA.  The time from shell says that it takes about 45 minutes to get it so you might want a better computer to deal with it. That make sense, shotgun did contribute about half of the warning.

[tsan.log](https://github.com/dmlc/xgboost/files/2292120/tsan.log)

src/c_api/c_api.cc
```diff
@@ -442,14 +442,14 @@ void PrefixSum(size_t *x, size_t N) {
   delete[] suma;
 }
 
-XGB_DLL int XGDMatrixCreateFromMat_omp(const bst_float* data,  // NOLINT
+XGB_DLL int XGDMatrixCreateFromMat_omp(const bst_float* _data,  // NOLINT
```
hcho3: Why are we renaming `data` to `_data`?

trivialfis: The first the I read the code got a little bit confused by namespace data, param data and mat.page_.data.  I will try to fix all leaks from tests in this PR and will change these thing back after done.

trivialfis:
If everything looks fine then I will rebase it into two commits, one concerning Travis and another concerning fixes.

trivialfis:
Ready now.

PR #3556, opened 2018-08-05
Fix #3485, #3540: Don't use dropout for predicting test sets

hcho3:
Dropout (for DART) should only be used at training time.

Fixes #3485.
Fixes #3540.

Issue #3540, opened 2018-07-31
loading pickle model is unstable as getting inconsistent prediction (10% abs. error)


guptaaditya13:
I saved the xgb regression model by pickle (protocol = 4). On loading the model by pickle and using .predict() it gives different output each time. 

training parameters : 
```
train_param = {
        'booster': 'dart',
        'sample_type': 'uniform',
        'normalize_type': 'tree',
        'rate_drop': 0.1,
        'skip_drop': 0.5,
        'objective': "reg:linear",
        'silent': False,
        'learning_rate': 0.1,
        'n_jobs': 18,
        'max_depth': 60,
        'tree_method': 'exact',
        'subsample': 0.75,
        'colsample_bytree': 0.8,
    }
```
number of trees = 30
Saving model by 
```
pickle.dump(model, open(str(model_name), "wb"),protocol=4)
```

testing : 
```
model = pickle.load(model_file.open('rb'))
print('trying batch of 5 for Predictions')
for x in range(0,5):
    print('iteration ' + str(x))
    for i in range(0,10,5):
        featureMatrix = features[i:i+5]
        originalPredictionsArray = originalPredictions[i:i+5]
        featuresDmatrix = xg.DMatrix(featureMatrix)
        predictions = model.predict(featuresDmatrix)
        for j in range(0,len(predictions)):
            percentageError = (predictions[j] - originalPredictionsArray[j])/(originalPredictionsArray[j]) * 100
            print('original : ' + str(originalPredictionsArray[j]) + ', predicted : ' + str(predictions[j]) + ', difference : ' + str(percentageError))
```
Output : 
```
trying batch of 5 for Predictions
iteration 0
original : 0.7610444, predicted : 0.7673624, difference : 0.8301770291664858
original : 2.5389864, predicted : 2.5028908, difference : -1.42165293710883
original : 2.5389864, predicted : 2.5028908, difference : -1.42165293710883
original : 0.4709874, predicted : 0.47068974, difference : -0.06319834516848555
original : 1.7682437, predicted : 1.7803591, difference : 0.6851685081183685
original : 1.7682437, predicted : 1.7321702, difference : -2.0400737641673303
original : 0.20668933, predicted : 0.22382358, difference : 8.289855681279548
original : 0.30097038, predicted : 0.31360516, difference : 4.198014276721492
original : 0.30097038, predicted : 0.31360516, difference : 4.198014276721492
[...]

PR #3563, opened 2018-08-06
Revert "Fix #3485, #3540: Don't use dropout for predicting test sets"

### LINKED ISSUE #3485: Inconsistent results with dart booster
I am getting inconsistent predictions using dart booster if I save it to file and then load again.
Using other boosters I have consistent predictions and all work as should.

I am using:
- Linux 4.17.5-1-ARCH
- Python 3.6.6
- xgboost 0.72.1
- numpy 1.14.5

```
import xgboost as xgb
import numpy as np

data = np.random.rand(10,1)
labels = np.random.rand(10,1) > 0.5
train = xgb.DMatrix(data, label=labels)
predict = xgb.DMatrix(np.random.rand(1,1))

param = {'booster': 'dart', 
         'silent':'1'}
bst = xgb.Booster()
bst = xgb.train(param.items(), train)
bst.save_model("xgb_test.bin")

for i in range(0,10): #run load model and prediction ten times to see difference
    bst = xgb.Booster()
    bst.load_model('xgb_test.bin')
    new_value = bst.predict(predict)
    print(new_value)
    if i == 0:
        last_value = new_value
    if last_value != new_value:
        raise SystemExit
```

hcho3:
Reverts dmlc/xgboost#3556, as it breaks the logic for custom evaluation and custom objective. I've documented the use of `ntree_limit` when predicting with test sets.

### DIFF
--- a/doc/parameter.rst
+++ b/doc/parameter.rst
@@ -12,6 +12,10 @@ Before running XGBoost, we must set three types of parameters: general parameter
 
   In R-package, you can use ``.`` (dot) to replace underscore in the parameters, for example, you can use ``max.depth`` to indicate ``max_depth``. The underscore parameters are also valid in R.
 
+.. contents::
+  :backlinks: none
+  :local:
+
 ******************
 General Parameters
 ******************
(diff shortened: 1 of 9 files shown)


khotilov:
Perhaps, here https://xgboost.readthedocs.io/en/latest/parameter.html#additional-parameters-for-dart-booster-booster-dart it would be a good place to add a sentence briefly explaining how predict works for DART. While it's not a "booster parameter" per se, it's an important parameter to use this booster properly.

hcho3:
@khotilov Not sure if a sentence would be sufficient. How about adding
```rst
.. note:: Using ``predict()`` with DART booster

  If the booster object is DART type, ``predict()`` will perform dropouts, i.e. only
  some of the trees will be evaluated. This will produce incorrect results if ``data`` is
  not the training data. To obtain correct results on test sets, set ``ntree_limit`` to
  a nonzero value, e.g.

    .. code-block:: python

      preds = bst.predict(dtest, ntree_limit=num_round)
```

RAMitchell:
Another way of dealing with this problem would be to cache the pointer to the training matrix and only use dropout when PredictBatch is called with the training matrix. This would automatically produce "correct" results without need for users to modify parameters. This behaviour would have to be carefully documented.

hcho3:
@RAMitchell For 0.80 release, I think we should retain the existing behavior. Let's come back to it after the release, to come up with a better design for DART booster.

PR #3564, opened 2018-08-07
Refactor parts of fast histogram utilities

RAMitchell:
I have made a few changes to the histogram utilities, mostly encapsulating functionality. This improves safety as it restricts direct access to pointers and provides a more defined interface to use the Column class.

I also added a few tests to check that the Column class was doing what I expected.

I am using this class as a part of a new experimental algorithm. 

@hcho3 A question: There is code for packing smaller bin ranges into Column<T> but this is not documented or used. Was this an experiment? Was it useful in the end?


hcho3:
@RAMitchell Yes, I toyed with the idea of packing small bins into Column, but in the end it didn't make much difference in performance. You can try using `uint8_t` or `uint16_t` with the (undocumented) parameter `colmat_dtype`.

hcho3:
Also I appreciate you taking time to read my old code. I wish I had time to get back to it but reality is such that I barely have time just to keep up with maintenance duty 😭

RAMitchell:
Would you be opposed to removing the bin packing/type switching? It is adding a fair amount of complexity that might be unnecessary.

hcho3:
I'll leave it to you to decide.

Ps. I had someone ask me if multi-GPU training increases amount of data that can be used. Does GPU hist partition the data over multiple GPUs or does the data get duplicated?

RAMitchell:
Data gets partitioned across GPUs, so more GPUs allows you to run bigger datasets.

hcho3:
@RAMitchell Should we include this in 0.80 release?

RAMitchell:
There should be no changes in behaviour for the end user. There is no need to include it but it probably won't hurt either.

PR #3569, opened 2018-08-08
[jvm-packages] Change the phase of exec maven plugin

CodingCat:
@hcho3 this would avoid compiling native libs for doc generation


hcho3:
Thanks!

hcho3:
So the native lib would be compiled for `mvn package` or `mvn compile` but not `mvn scala:doc`?

CodingCat:
Exactly

hcho3:
I still have to `mvn install` before `mvn scala:doc` because XGBoost4J-Spark depends on XGBoost4J and the former tries to locate the latter from the local Maven repository. Anyway out of this?

CodingCat:
ah, I know why it works in my side, because I once installed the package locally

I don't have a good solution for now....

hcho3:
@CodingCat It's not a show stopper, so I'll go ahead and merge the current PR as is. Closing this PR for now.

PR #3576, opened 2018-08-10
[jvm-packages] Avoid loosing precision when computing probabilities by converting to Double early

tovbinm:
Current computation of `Array(1 - probability(0), probability(0))` is done with `Float` which looses precision. This fix converts to `Double` first, then computes the probabilities.


tovbinm:
Not sure what's he problem with the build, but it seems unrelated. How do I restart it? @CodingCat

tovbinm:
@hcho3 this should be good to go

tovbinm:
@hcho3 this should be good to go

PR #3577, opened 2018-08-10
[jvm-packages] Make sure 'thresholds' are considered when executing predict method

tovbinm:
Suggestion: I would recommend implementing `predictRaw` and `raw2probabilityInPlace` correctly as well, so it would comply with all regular spark stages.


tovbinm:
Not sure what's he problem with the build, but it seems unrelated. How do I restart it? @CodingCat

tovbinm:
@hcho3 this should be good to go

PR #3599, opened 2018-08-16
Fix accessing DMatrix.handle before set.

### LINKED ISSUE #3597: [Python] AttributeError: 'DMatrix' object has no attribute 'handle'
Version: 725f4c36f26955d55baeb4c03f8ec955ca3cbaf3
Python: Python 3.6 virtualenv

The very first time I encountered the failure is on Travis test farm.  Doing nose tests locally reproduced it. In `test_dmatrix_numpy_init` there are some `assertRaises'`  The `ValueError` is thrown before `self.handle` is set.  

I can rearrange the order of those code, but I don't know why this bug is not deterministic on Travis.

trivialfis:
Close #3597.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -479,7 +479,7 @@ class DMatrix(object):
             nthread))
 
     def __del__(self):
-        if self.handle is not None:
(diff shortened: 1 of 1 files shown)


hcho3:
Thanks!

PR #3600, opened 2018-08-16
Add JSON model dump functionality documentation

grace-lam:
I wrote a native XGBoost to Vespa converter that uses XGBoost's JSON model dump as input (https://github.com/vespa-engine/vespa/pull/6577), so I added some missing documentation for the existing `dump_format` parameter in the CLI version.


hcho3:
Thanks!

PR #3560, opened 2018-08-06
sklearn api for ranking

Shiki-H:
This PR resolves #2859. My implementation is slightly different from OP's implementation as group information has to be separately fed to `fit` method, instead of combining it together with features. I believe that ranking is quite a different task compare to classification or regression, so this difference in interface is probably justified. By the way, this is currently how LightGBM implements its sklearn ranker interface. 
Regarding testing, I am not very sure what is the best way to test this interface. What I have done so far is to compare it with the results obtained from the `train` method, but this relies on the dataset used in `demo/rank`. If you believe there are better ways of testing, please let me know.

PR #3619, opened 2018-08-21
Update Python API doc

hcho3:
XGBRanker was added by #3560 and needs to be added to the Python API documentation.

@Shiki-H

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -14,6 +14,7 @@
 from subprocess import call
 from sh.contrib import git
 import urllib.request
+from urllib.error import HTTPError
 from recommonmark.parser import CommonMarkParser
 import sys
 import re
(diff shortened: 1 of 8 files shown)


StrikerRUS:
@hcho3 
I think it's appropriate place to notify you about the following. By default, Sphinx doesn't generate docstrings from parent class (`XGBModel`). That's why, for example, `XGBRegressor` at present has only documentation describing its constructor. `:inherited-members:` flag (http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) can solve the problem.

hcho3:
@StrikerRUS Thanks for letting us know. I went ahead and added `:inherited-members:` to XGBRegressor.

StrikerRUS:
No problem, @hcho3 !

And what about `XGBClassifier` and `XGBRanker`? For instance, `get_booster`, `get_params`, `feature_importances_`, etc. won't be shown at the docs page.

Also I've noticed that `XGBRanker`'s `predict` method will not be shown even with `:inherited-members:` flag because it has no own docstring but overrides parent's method. Possible workaround (if you don't want copy-paste parent's docstring or write decorator https://stackoverflow.com/a/13743316):
```
predict.__doc__ = XGBModel.predict.__doc__ 
```

Issue #3623, opened 2018-08-22
Segmentation fault when using GPU


yananchen1989:
```
dtrain = xgb.DMatrix('./part-0819.libsvm#dtrain.cache')
dtest = xgb.DMatrix('./part-0820.libsvm#dtest.cache')


#booster: gbtree(default), gblinear or dart
param = {'max_depth':7, 'eta': 0.01, 'silent':1, 'objective':'binary:logistic','eval_metric':['error','auc'],\
          'subsample': 0.7, 'min_child_weight':0.2, 'colsample_bytree':0.7,'scale_pos_weight':1.6,\
          'alpha':0.1,'lambda':0.1, 'max_bin':256,'disable_default_eval_metric':1} 
param['tree_method'] = 'gpu_hist'
param['gpu_id'] = 0
print (param)

evallist= [(dtrain, 'train'),(dtest, 'eval')]
bst = xgb.train(param, dtrain,  num_boost_round=3000, evals =evallist , early_stopping_rounds=100,\
	            verbose_eval=5) # xgb_model


```

GPU is tested by benchmark and it can work normally.
But when I tried this code above, the program failed with the error `Segmentation fault`. No other message come out.
And when I get rid of the gpu setting: 
```
param['tree_method'] = 'gpu_hist'
param['gpu_id'] = 0
```
The programme also failed with much error:
```
[15:14:52] /data01/xgboost/src/tree/updater_histmaker.cc:303: fv=0.418824, hist.size=51[15:14:52] /data01/xgboost/src/tree/updater_histmaker.cc:303: fv=0.783455, hist.size=53[15:14:52] /data01/xgboost/src/tree/updater_histmaker.cc:303: fv=0.943319, hist.size=58

[15:14:52] /data01/xgboost/src/tree/updater_histmaker.cc:303: fv=0.303076, hist.size=40
[15:14:52] /data01/xgboost/src/tree/updater_histmaker.cc:303: fv=0.223285, hist.size=48[15:14:52] /data01/xgboost/src/tree/updater_histmaker.cc:303: fv=0.460549, hist.size=58[15:14:52] /data01/xgboost/src/tree/updater_histmaker.cc:303: fv=0.195431, hist.size=50[15:14:52] /data01/xgboost/src/tree/updater_histmaker.cc:303: fv=0.36465, hist.size=59
[...]

PR #3624, opened 2018-08-23
Document CUDA requirement, lack of external memory on GPU

hcho3:
Addresses #3623 and #3625.


hcho3:
@RAMitchell Just checking with you whether CUDA 8.0+, Compute Capability 3.5+ is reasonable systems requirements going forward.

PR #3626, opened 2018-08-23
Merge generic device helper functions into gpu set.

trivialfis:
* Remove the use of old NDevices* functions.

@RAMitchell Some planed changes from #3608 extracted here to let that PR less messy.  To complete that PR, some functions need to be generalized for both CPU and GPU.  Since I can't test it on multi-gpu environment, I think it's best to split this as an independent PR to ease reviewing.

### DIFF
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -6,6 +6,8 @@
 #ifndef XGBOOST_COMMON_COMMON_H_
 #define XGBOOST_COMMON_COMMON_H_
 
+#include <xgboost/base.h>
+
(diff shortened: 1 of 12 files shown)


trivialfis:
@RAMitchell , @canonizer I think I need to rebase it on #3446 .  Otherwise this should be all of it once tests passed.  Please help reviewing, I can't test it on multi-gpu environment.  Thanks in advance. :)

RAMitchell:
LGTM

PR #3635, opened 2018-08-26
GPU memory usage fixes + column sampling refactor

RAMitchell:
@pseudotensor 

Address very large memory usage due to thrust copy calls on non-trivial objects.

Address very large memory usage in histogram construction when the number of features and number of leaves is large. The gpu_hist updater will now recycle memory when it exceeds some limit. This can make histogram construction slightly less efficient as we may not be able to use the subtraction truck.

Refactored column sampling to use the same simplified interface across multiple updaters. @hcho3 as a consequence of this the hist updater now supports column sampling by level.

Also no longer redundantly evaluate splits when using column sampling in gpu_hist.


pseudotensor:
Great!

PR #3637, opened 2018-08-27
Add google test for a column sampling, restore metainfo tests

RAMitchell:
Some meta infotests were accidentally removed in a previous commit due to failures with some Windows compilers.

### DIFF
--- a/tests/cpp/data/test_metainfo.cc
+++ b/tests/cpp/data/test_metainfo.cc
@@ -103,22 +103,32 @@ TEST(MetaInfo, LoadQid) {
     0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55, 60
   };
   const std::vector<xgboost::Entry> expected_data{
-    {1, 1}, {2, 1}, {3, 0}, {4, 0.2}, {5, 0},
-    {1, 0}, {2, 0}, {3, 1}, {4, 0.1}, {5, 1},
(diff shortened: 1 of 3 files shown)


tests/cpp/data/test_metainfo.cc
```diff
@@ -69,4 +69,66 @@ TEST(MetaInfo, SaveLoadBinary) {
...
+  const std::vector<xgboost::bst_uint> expected_group_ptr{0, 4, 8, 12};
+  CHECK(info.qids_ == expected_qids);
+  CHECK(info.group_ptr_ == expected_group_ptr);
+  CHECK_GE(info.kVersion, info.kVersionQidAdded);
+
+  const std::vector<size_t> expected_offset{
+    0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55, 60
+  };
+  const std::vector<xgboost::Entry> expected_data{
+      xgboost::Entry(1, 1),   xgboost::Entry(2, 1),   xgboost::Entry(3, 0),
```
hcho3: So using `{1, 1}` to initialize `xgboost::Entry` fails MSVC?

RAMitchell: The initialisation type is ambiguous according to visual studio 2015.

hcho3: Got it. Thanks for the fix!

PR #3643, opened 2018-08-28
Lanuch function for unifying CPU and GPU code. [Reopen]

trivialfis:
Continue on #3608, Using SFINA to mitigate the issue encountered.  Wait for further discussion. @RAMitchell


src/common/transform.h
```diff
@@ -0,0 +1,265 @@
...
+      LaunchRangeCUDA(_func, _range, _flags, _devices, _vectors...);
+    } else {
+      LaunchRangeCPU(_func, _range, _flags->HostPointer(), UnpackHDV(_vectors)...);
+    }
+  }
+
+ private:
+  template <typename... T>
+  void Reshard(GPUSet _devices, HostDeviceVector<T>*... _vectors) {
+    std::vector<HDVAny> vectors {_vectors...};
```
RAMitchell: Can you just use recursion on the variadic arguments to iterate through the vectors? This would make HDVAny unnecessary.

src/common/common.h
```diff
@@ -103,5 +134,99 @@ class Range {
 };
 
 }  // namespace common
+
+/* \brief set of devices across which HostDeviceVector can be distributed.
+ *
+ * Currently implemented as a range, but can be changed later to something else,
+ *   e.g. a bitset
+ */
+class GPUSet {
```
RAMitchell: What is the rationale for moving GPUSet here?

src/common/common.h
```diff
@@ -1,17 +1,48 @@
...
+#define WITH_CUDA true
+#else
+#define WITH_CUDA false
+#endif
+
+namespace dh {
+#if defined(__CUDACC__)
+/*
+ * Error handling  functions
+ */
```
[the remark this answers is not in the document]
trivialfis: Moved here, used by GPUSet, not duplicated.

src/common/transform.h
RAMitchell: I think it would be possible using std::thread but probably not omp. Not sure how important it is that this function is parallel.

src/common/common.h
RAMitchell: Okay that's fine

src/common/transform.h
trivialfis: I'm not sure either.  That seems to be the right job for OpenMP, one line and worry about nothing.

trivialfis:
The idea of segmented transform is dropped, now is simply transforming each element.

Shared memory support was added and removed, if there is a use case somewhere in the future, please let me know so that I can get it back.

For spiting up an Evaluator in Transform, it's just acting as a delimiter to make passing a bunch of arguments easier on the eye.

All regression and softmax objs are now generic for GPU and CPU.

RAMitchell:
It may be possible to implement softmax requiring no working memory. Also I think I prefer recursion for the reshard operation. We can leave it single threaded for now.

trivialfis:
Okay, I will try recursion.

The passed in preds pointer for get gradient is now const, I can't do inplace softmax.  What would be your suggestion?

RAMitchell:
Maybe do it in 3 passes? 1st pass find wmax, second pass find wsum, third pass use these values to calculate final results.

trivialfis:
@hcho3 Hi, could you clear the cache of Jenkins for me?  I am truly sorry about this.  Seems every time a file changes its name or removed the problem in Jenkins would arise.  The re-factoring about this PR should be complete now.

trivialfis:
@hcho3  Never mind, I think this time the problem is in the old nvcc. :(

trivialfis:
# Choosing devices
Currently the `Transform` class accepts a `GPUSet` parameter to determine devices to use. And all objectives using `Transform` pass `GPUSet::All()` as argument.  This implies that objectives using `Transform` will use GPU to calculate the result as long as there's available device, despite earlier step in the pipeline might be running on CPU.  One way to modify that is explicitly setting `n_gpu` to 0, then `Transform` will run on CPU.

Another method determining device would be looping through all input vectors to see if there is any one of them resided on GPU.  As can be done by:
``` c++
    template <typename T>
    bool HasDeviceVector(const HostDeviceVector<T>* vec) const {
      return !vec->Devices().IsEmpty();
    }
    template <typename Head, typename... Rest>
    bool HasDeviceVector(const HostDeviceVector<Head>* head,
                         const HostDeviceVector<Rest>*... rest) const {
      return !head->Devices().IsEmpty() || HasDeviceVector(rest...);
    }
```
But this way user(actual algorithm) of `Transform` won't be able to specify devices to use.  I think it's best to let the actual algorithms to make this decision.

## Performance
I ran a small benchmark using demo/cover_type.py, the CPU training time is used to compare two different methods since it will require memory copy when passing CPU data to GPU.  To generate the result, I ran 300 iterations with both methods.

CPU: Intel 4720HQ
Memory: 16GB DDR3
GPU: GTX 960M
CUDA: 9.2
Platform: Fedora 27 x86-64

With following result:

 Methods |  Training Time (seconds)
 --- | ---
Loop through |  220.12616324424744
By GPUSet |  216.3728747367859

The result doesn't really tell the difference, the acceleration from GPU evens out the time needed for copying.  With a decent non-mobile GPU, copying data should be worthy.

trivialfis:
@RAMitchell  Ready. :)

hcho3:
@trivialfis Any update on this?

ps. Please ignore the failed test `continuous-integration/jenkins/pr-head`. (This was due to misconfiguration of Jenkins CI server.) Only `continuous-integration/jenkins/pr-merge` is relevant.

trivialfis:
@hcho3  Thanks for noticing.  I will need to rebase it on master branch and possibly do some better testings and benchmarks on this.  Sorry for the long waiting time.

src/common/transform.h
```diff
@@ -0,0 +1,211 @@
...
+      if (reshard_)
+        UnpackReshard(distribution_, _vectors...);
+
+      GPUSet devices = distribution_.Devices();
+      size_t range_size = *range_.end() - *range_.begin();
+#pragma omp parallel for schedule(static, 1) if (devices.Size() > 1)
+      for (omp_ulong i = 0; i < devices.Size(); ++i) {
+        int d = devices.Index(i);
+        // Ignore other attributes of GPUDistribution for spliting index.
+        size_t shard_size =
```
canonizer: Could you use the shard size derived from `distribution_`? The `HostDeviceVector` objects are not necessarily block-distributed.

src/common/transform.h
```diff
@@ -0,0 +1,211 @@
...
+              typename... HDV>
+    void LaunchCUDA(Functor _func, HDV*... _vectors) const {
+      LOG(FATAL) << "Not part of device code. WITH_CUDA: " << WITH_CUDA();
+    }
+#endif
+
+    template <typename... HDV>
+    void LaunchCPU(Functor func, HDV*... vectors) const {
+      auto end = *(range_.end());
+#pragma omp parallel for schedule(static, 1)
```
canonizer: I doubt that scheduling in chunks of size 1 leads to the best performance. I think it is better to just omit the chunk size.

src/objective/multiclass_obj.cu
```diff
@@ -0,0 +1,179 @@
...
+    CHECK(preds.Size() == (static_cast<size_t>(param_.num_class) * info.labels_.Size()))
+        << "SoftmaxMultiClassObj: label size and pred size does not match";
+    out_gpair->Resize(preds.Size());
+    label_correct_.Fill(1);
+    const int nclass = param_.num_class;
+    const auto ndata = static_cast<int64_t>(preds.Size() / nclass);
+    common::ReshardAll(out_gpair, GPUDistribution::Block(devices_),
+                       &info.labels_, GPUDistribution::Block(devices_),
+                       &preds, GPUDistribution::Granular(devices_, nclass),
+                       &info.weights_, GPUDistribution::Block(devices_),
```
canonizer: Please avoid this, and just call `Reshard()` on each `HostDeviceVector` individually.

src/objective/multiclass_obj.cu
```diff
@@ -0,0 +1,179 @@
...
+      common::Transform<>::Init(
+          [=] XGBOOST_DEVICE(size_t _idx, common::Span<bst_float> _preds) {
+            common::Span<bst_float> point =
+                _preds.subspan(_idx * nclass, nclass);
+            common::Softmax(point.begin(), point.end());
+          },
+          common::Range{0, ndata}, GPUDistribution::Granular(devices_, nclass))
+        .Eval(io_preds);
+    } else {
+      common::ReshardAll(io_preds, GPUDistribution::Granular(devices_, nclass),
```
canonizer: Please call `Reshard()` on each `HostDeviceVector` individually.

src/common/transform.h
```diff
@@ -0,0 +1,211 @@
...
+
+template <typename HDV>
+void ReshardAll(HDV* vector, GPUDistribution dist) {
+  vector->Reshard(dist);
+}
+template <typename HDV, typename... HdvDist>
+void ReshardAll(HDV* vector, GPUDistribution dist, HdvDist... rest) {
+  vector->Reshard(dist);
+  ReshardAll(rest...);
+}
```
canonizer: Please remove this (I mean `ReshardAll`).

Functions like these improve neither performance nor readability of the code. On the contrary, they increase complexity by introducing yet another unnecessary level of abstraction.

Just calling `Reshard()` on each HostDeviceVector sequentially is clearer.

src/objective/hinge.cu
```diff
@@ -0,0 +1,109 @@
...
+                   HostDeviceVector<GradientPair> *out_gpair) override {
+    CHECK_NE(info.labels_.Size(), 0U) << "label set cannot be empty";
+    CHECK_EQ(preds.Size(), info.labels_.Size())
+        << "labels are not correctly provided"
+        << "preds.size=" << preds.Size()
+        << ", label.size=" << info.labels_.Size();
+
+    const bool is_null_weight = info.weights_.Size() == 0;
+    const size_t ndata = preds.Size();
+    out_gpair->Resize(ndata);
```
canonizer: If a `HostDeviceVector` with an empty distribution is resized, memory will be allocated for it on the host. If it is later resharded, this host memory allocation will remain.

Therefore, it is better to reshard a `HostDeviceVector` before resizing it, rather than the other way around.

src/common/common.cu
```diff
@@ -0,0 +1,18 @@
...
+
+namespace xgboost {
+
+int AllVisibleImpl::AllVisible() {
+  int n_visgpus = 0;
+  try {
+    dh::safe_cuda(cudaGetDeviceCount(&n_visgpus));
+  } catch(const std::exception& e) {
+    return 0;
+  }
```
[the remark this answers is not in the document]
trivialfis: The try/catch is for XGBoost compiled with CUDA but running on CPU,  in which case the cudaGetDeviceCount will fail and we return 0 as default.

src/common/common.h
```diff
@@ -1,17 +1,48 @@
...
+#include <limits>
+#include <type_traits>
 #include <vector>
 #include <string>
 #include <sstream>
 
+#if defined(__CUDACC__)
+#include <thrust/system/cuda/error.h>
+#include <thrust/system_error.h>
+#define WITH_CUDA() true
```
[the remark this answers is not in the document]
trivialfis: If you mean `XGBOOST_USE_CUDA`, it's a definition from CMake, which doesn't indicate whether this translation unit is being compiled by nvcc.

src/common/math.h
```diff
@@ -29,22 +29,26 @@ inline avx::Float8 Sigmoid(avx::Float8 x) {
...
-inline void Softmax(std::vector<float>* p_rec) {
-  std::vector<float> &rec = *p_rec;
-  float wmax = rec[0];
-  for (size_t i = 1; i < rec.size(); ++i) {
-    wmax = std::max(rec[i], wmax);
+template <typename Iterator>
+XGBOOST_DEVICE inline void Softmax(Iterator start, Iterator end) {
+  float wmax = *start;
+  for (Iterator i = start+1; i != end; ++i) {
+    wmax = fmaxf(*i, wmax);
```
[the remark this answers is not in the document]
trivialfis: Thanks, let me try a `static_assert`.

src/common/transform.h
```diff
@@ -0,0 +1,211 @@
...
+ *     will merge functions with same signature.
+ */
+template <bool CompiledWithCuda = WITH_CUDA()>
+class Transform {
+ private:
+  template <typename Functor>
+  struct Evaluator {
+   public:
+    Evaluator(Functor func, Range range, GPUSet devices, bool reshard) :
+        func_(func), range_{range}, reshard_{reshard},
```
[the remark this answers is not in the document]
trivialfis: It's dealing with old gcc and msvc,  they have problem initializing lambda a an object, {} doesn't work.

src/common/transform.h
trivialfis: Sorry about that, I made the changes in a local branch but haven't done the push yet. :)

src/common/transform.h
trivialfis: You are right.  These functions will be removed.

src/objective/hinge.cu
trivialfis: Thanks for the details.

src/common/transform.h
trivialfis: My mistake,  I tried to do that but gave up, since vectors are not necessarily come with the same distribution (different granularity for example).  The `shard_size` is defined for indexing thread, which should be fit the block distribution.

Issue #2796, opened 2017-10-17
[PYTHON][RABIT] calling rabit.init multiple times exits the process


TomAugspurger:
## Environment info
Operating System: macOS 10.12.6

Compiler: gcc

Package used (python/R/jvm/C++): python

`xgboost` version used: 0.6

If installing from source, please provide 

1. The commit hash (`git rev-parse HEAD`) 78c4188cec31425f708d238160ea3afb67a7250a

2. Logs will be helpful (If logs are large, please upload as attachment).

If you are using python package, please provide

1. The python version and distribution: 0.6, source

## Steps to reproduce

```python
In [1]: import xgboost as xgb
In [2]: xgb.rabit.init()

In [3]: xgb.rabit.init()
rabit::Init is already called in this thread
```
(and the process exits)

---

Would it be possible to avoid this, say by storing a global or threadlocal variable? It seems like https://github.com/dmlc/rabit/blob/a764d45cfb438cc9f15cf47ce586c02ff2c65d0f/python/rabit.py#L46-L52 intends to do something like that, but even then, [the init](https://github.com/dmlc/rabit/blob/a764d45cfb438cc9f15cf47ce586c02ff2c65d0f/python/rabit.py#L103) will still continue to run, since it doesn't check the result and continues unconditionally.

I'm happy to submit a patch for this, but I'm not sure whether to make it in the rabit repo or here. My local copy picks up the version of rabit in `xgboost/python-package/xgboost/rabit.py`, so I've made the issue here.

TomAugspurger:
Was this fixed in rabit?

hcho3:
This issue was closed due to the current triaging. We try to aggressive close issues from now on so opened issues can get attended to. I went ahead and re-opened it.

tqchen:
please send a PR to the rabit repo

PR #3656, opened 2018-08-31
Fix rabit initializing twice

TomAugspurger:
Closes https://github.com/dmlc/xgboost/issues/2796 (which is already closed, but not fixed).

### DIFF
--- a/python-package/xgboost/rabit.py
+++ b/python-package/xgboost/rabit.py
@@ -5,10 +5,15 @@
 from __future__ import absolute_import
 import sys
 import ctypes
+import threading
+import warnings
 import numpy as np
 
 from .core import _LIB, c_str, STRING_TYPES
 from .compat import pickle
+
+_local = threading.local()
+_local.initialized = False
 
 
 def _init_rabit():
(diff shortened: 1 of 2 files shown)


TomAugspurger:
What's the expected behavior of `rabit.finalize` here?

```
In [1]: import xgboost

In [2]: xgboost.rabit.init()

In [3]: xgboost.rabit.finalize()

In [4]: xgboost.rabit.init()
/Users/taugspurger/sandbox/xgboost/python-package/xgboost/rabit.py:33: UserWarning: rabit has already been initialized on thread 140736122418048.
  threading.current_thread().ident)

In [5]: xgboost.rabit.finalize()
rabit::Finalize engine is not initialized or already been finalized.
```

I would expect `finalize` to set `initialized` back to False, and not crash after calling finalize twice.

hcho3:
@TomAugspurger I think finalize() is supposed to be called only once:

https://github.com/dmlc/rabit/blob/87143deb4c0a34302f727ba35497e3380b2cced8/src/engine.cc#L56-L62

TomAugspurger:
Yes, it seems that way. It's not really clear to me what the expected behavior is here, so I'm going to close this.

PR #3660, opened 2018-09-03
Fix CRAN check by removing reference to std::cerr

hcho3:
Reference to std::cerr causes CRAN check to fail. The solution is to remove the reference when compiled with R.

### DIFF
--- a/R-package/src/Makevars.in
+++ b/R-package/src/Makevars.in
@@ -12,7 +12,7 @@ XGB_RFLAGS = -DXGBOOST_STRICT_R_MODE=1 -DDMLC_LOG_BEFORE_THROW=0\
 
 # disable the use of thread_local for 32 bit windows:
 ifeq ($(R_OSTYPE)$(WIN),windows)
-    XGB_RFLAGS += -DDMLC_CXX11_THREAD_LOCAL=0 -msse2 -mfpmath=sse
+    XGB_RFLAGS += -DDMLC_CXX11_THREAD_LOCAL=0
 endif
 $(foreach v, $(XGB_RFLAGS), $(warning $(v)))
 
--- a/R-package/src/Makevars.win
+++ b/R-package/src/Makevars.win
@@ -24,7 +24,7 @@ XGB_RFLAGS = -DXGBOOST_STRICT_R_MODE=1 -DDMLC_LOG_BEFORE_THROW=0\
 
 # disable the use of thread_local for 32 bit windows:
 ifeq ($(R_OSTYPE)$(WIN),windows)
-    XGB_RFLAGS += -DDMLC_CXX11_THREAD_LOCAL=0 -msse2 -mfpmath=sse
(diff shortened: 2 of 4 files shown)


hetong007:
Have you tried to check with the `--as-cran` flag:

```
R CMD check --as-cran xgboost*.tar.gz
```

hcho3:
Yes

hetong007:
My steps:

- Clone xgboost
- Apply your patch
- Modify `Makefile` to build vignettes and check as cran
- `make Rcheck`

Besides the vignettes part, do you see anything else different with yours?

hcho3:
Is it master + my patch or release_0.80 + my patch?

hetong007:
It is master + your patch

hcho3:
I think for 0.80 submission, we should use `release_0.80` branch, not `master`

hetong007:
OK now I see. For the warning I met, will you consider it as a separate issue with a later fix?

hcho3:
@hetong007 Can you try `release_0.80` with my patch to see if it's CRAN compliant? If so, we can come back to the warning later

hetong007:
I checkout to release_0.80 and updated `dmlc-core` submodule, still have the WARNING：

```
* checking compilation flags used ... WARNING
Compilation used the following non-portable flag(s):
  ‘-Wdate-time’ ‘-Werror=format-security’ ‘-Wformat’
```

hcho3:
I'm unable to re-produce it on my end. If you are using EC2 machine, can you grant me access to it?

hcho3:
It's weird, I managed to reproduce the warning by upgrading R from 3.4.4 to 3.5.1

hetong007:
From R win-builder, we get the following WARNING:

```
* checking compilation flags in Makevars ... WARNING
Non-portable flags in variable 'PKG_CPPFLAGS':
  -msse2 -mfpmath=sse
```

On the other hand, the previous warning on `non-portable flag(s)` are gone. It might be a local-only warning and a possible explanation can be found at https://stackoverflow.com/questions/50658198/cran-submission-r-cmd-check-warning-compilation-flags-used

hcho3:
@hetong007 #3335 introduced the SSE flag to ensure that Windows 32-bit and 64-bit behave consistently. Should we rewrite the tests so that we can remove the flag?

hetong007:
Can you elaborate on "tests", i.e. which tests are you referring to?

hcho3:
@hetong007 Here it is:
https://github.com/dmlc/xgboost/blob/0988fb191f6aacde50020b2d5e94c92df2ac4e58/R-package/tests/testthat/test_helpers.R#L43-L44

The problem with this test is that it expects the model dump to be exactly 200 lines long, so a slight variation causes it to fail.

hetong007:
@hcho3 Maybe we can mask that test on 32-bit platform.

hcho3:
Done

hetong007:
LGTM. Thanks @hcho3 , please ping me when it is backported.

hcho3:
@hetong007 Done.

PR #3666, opened 2018-09-04
Fix `demo/rank` data processing order

Shiki-H:
This is a very trivial issue. I noticed that with the original script, data processing is bundled with `run_exp.sh`, which now serves as a demo on how to use CLI. If one wants to play with python interface, he needs to execute `run_exp.sh` first. I figured that it might be better to process the data right after it is downloaded so one can run the python demos independently from CLI demo.


hcho3:
Thanks!

PR #3668, opened 2018-09-04
Add the missing max_delta_step

khotilov:
The `max_delta_step` parameter fell through the cracks during the recent change to `SplitEvaluator` for the `grow_colmaker` and `grow_fast_histmaker` updaters. It can be very useful for noisy datasets. I didn't see any noticeable timing performance change for the default `max_delta_step=0` case.

Also I've removed the unused reg_gamma parameter from ElasticNet, since it's only to be used during pruning.

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -223,3 +223,17 @@ test_that("train and predict with non-strict classes", {
   expect_error(pr <- predict(bst, train_dense), regexp = NA)
   expect_equal(pr0, pr)
 })
+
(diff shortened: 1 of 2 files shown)


hcho3:
@henrygouk Can you take a look?

henrygouk:
Looks good to me.

In future, constraints like this should probably be put into new SplitEvaluator subclasses to prevent a labyrinth of if statements from clogging up ElasticNet. That said, I think this particular constraint should be present in ElasticNet, since it is fixing backwards compatibility.

PR #3673, opened 2018-09-05
Update dmlc-core, to fix partitioned file loading

hcho3:
Updating dmlc-core brings the following benefits:

* **Fix a bug in reading partitioned data files in distributed environment** (dmlc/dmlc-core#452)
* Add OMPException class: can be used to wrap OpenMP loops, so that exceptions don't crash whole program (dmlc/dmlc-core#445)
* Handle error for S3 list (dmlc/dmlc-core#439)


hcho3:
@htgeis

Issue #3616, opened 2018-08-21
local variable 'lib' referenced before assignment


yananchen1989:
When I import xgboost in ipython of Anaconda, there comes the error:
```
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-1-e528dc1a32ff> in <module>()
----> 1 import xgboost

/root/anaconda2/lib/python2.7/site-packages/xgboost/__init__.py in <module>()
      9 import os
     10
---> 11 from .core import DMatrix, Booster
     12 from .training import train, cv
     13 from . import rabit                   # noqa

/root/anaconda2/lib/python2.7/site-packages/xgboost/core.py in <module>()
    134
    135 # load the XGBoost library globally
--> 136 _LIB = _load_lib()
    137
    138

/root/anaconda2/lib/python2.7/site-packages/xgboost/core.py in _load_lib()
    126         except OSError:
    127             continue
--> 128     lib.XGBGetLastError.restype = ctypes.c_char_p
    129     lib.callback = _get_log_callback_func()
    130     if lib.XGBRegisterLogCallback(lib.callback) != 0:

UnboundLocalError: local variable 'lib' referenced before assignment
```
Any solutions ?

hcho3:
Did you install from the source, or did you run `pip install xgboost`?

yananchen1989:
@hcho3  I just use pip install method, without any other operations.

hcho3:
Can you describe the OS environment? The binary wheel was built using CentOS 7.

yananchen1989:
I also tried the 'make -j4' method as the official docs said.
My os is 
NAME="Ubuntu"
VERSION="14.04.5 LTS, Trusty Tahr"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 14.04.5 LTS"
VERSION_ID="14.04"

```
Makefile:31: MAKE [make] - checked OK
g++ -DDMLC_LOG_CUSTOMIZE=1 -std=c++11 -Wall -Wno-unknown-pragmas -Iinclude   -Idmlc-core/include -Irabit/include -I/include -O3 -funroll-loops -msse2 -fPIC -fopenmp -MM -MT build/learner.o src/learner.cc >build/learner.d
[...]

PR #3690, opened 2018-09-10
Better error message for failed library loading

hcho3:
Addresses #3616.

On some systems, the XGBoost library (`libxgboost.so`, `xgboost.dll`, or `libxgboost.dylib`) may fail to load because of missing dependencies. Previously, it would result in a uninformative error
```
UnboundLocalError: local variable 'lib' referenced before assignment
```

This pull request is aimed at producing a better error message for this scenario.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -118,13 +118,26 @@ def _load_lib():
     if len(lib_paths) == 0:
         return None
     pathBackup = os.environ['PATH']
+    lib_success = False
+    os_error_list = []
     for lib_path in lib_paths:
         try:
             # needed when the lib is linked with non-system-available dependencies
             os.environ['PATH'] = pathBackup + os.pathsep + os.path.dirname(lib_path)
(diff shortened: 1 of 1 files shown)


python-package/xgboost/core.py
```diff
@@ -118,13 +118,26 @@ def _load_lib():
...
+        except OSError as e:
+            os_error_list.append(str(e))
             continue
+    if not lib_success:
+        libname = os.path.basename(lib_paths[0])
+        raise XGBoostError(
+            'XGBoost Library ({}) could not be loaded.\n'.format(libname) + \
+            'Likely causes:\n' + \
+            '  * OpenMP runtime is not installed ' + \
+            '(vcomp140.dll for Windows, libgomp.so for UNIX-like OSes)\n' + \
```
khotilov: ... or libgomp-1.dll for Windows

hcho3: This is for MinGW?

hcho3: Done

Issue #901, opened 2016-03-02
How to use xgboost to do lambdamart listwise ranking?


gusuperstar:
“rank:pairwise” –set XGBoost to do ranking task by minimizing the pairwise loss

do u mean this? Since lambdamart is a listwise approach, how can i fit it to listwise ranking? including commond, parameters, and training data format， and where can i set the lambda for lambdamart. 
could u give a brief demo or intro? many thanks!

gusuperstar:
ok, i see. XGBoost supports accomplishing ranking tasks. In ranking scenario, data are often grouped and we need the [group information file](../../doc/input_format.md#group-input-format) to s
        pecify ranking tasks. The model used in XGBoost for ranking is the LambdaRank, this function is not yet completed. Currently, we provide pairwise rank.


So, listwise learing is not supportted. Any plan?

tqchen:
use rank:ndcg for lambda rank with ndcg metric

gdf0:
Hi, I just tried to use both objective = 'rank:map' and objective = 'rank:ndcg', but none of them seem to work. The pairwise objective function is actually fine. I can see in the code that the LambdaMART objective function is still there, however I do not understand why it cannot be selected using the python API. Thanks.

kapild:
@tqchen can you comment if `rank:ndcg` or `rank:map` works for Python?

travisbrady:
This needs clarification in the docs.
Specifically:
1. The FAQ says "Yes, xgboost implements LambdaMART. Checkout the objective section in parameters" yet the parameters page contains no mention of LambdaMART whatsoever.
2. If LambdaMART does exist, there should be an example. I'm happy to submit a PR for this.
3. This is maybe just an issue of mixing of terms, but I'd recommend that if Xgboost wants to advertise LambdaMART on the FAQ that the docs and code then use that term also.

Sandy4321:
is it resolved?

hcho3:
@vatsan Looks like it was an oversight. Can you submit a pull request to update the parameter doc?

hcho3:
@vatsan @Sandy4321 @travisbrady I am adding all objectives to parameter doc: #3672

PR #3697, opened 2018-09-12
[jvm-packages] Add rank:ndcg and rank:map to Spark supported objectives

linjer:
According to the [Learning Task Parameters documentation](https://xgboost.readthedocs.io/en/latest/parameter.html#learning-task-parameters), `rank:ndcg` and `rank:map` learning objectives should be available.

There is a little discussion about if there was any specific reason why these objectives were previously omitted [here](https://github.com/dmlc/xgboost/issues/901) and [here](https://discuss.xgboost.ai/t/jvm-packge-does-xgboost-spark-support-rank-ndcg/229).


CodingCat:
LGTM, merged, thanks!

PR #3717, opened 2018-09-21
Fix #3702: do not round up integer thresholds for integer features in JSON dump

### LINKED ISSUE #3702: Prediction for some samples does not follow the json that is obtained by get_dump()
booster_98
This is the 99th booster of my model which I obtained from the get_dump() function.
When I perform model._Booster.predict(X_test, pred_leaf=True), then for the 99th estimator it gives output as 8. That means, for that estimator, output node number is 8, which is -0.094418548 here.

My input data is:
[cylinders] := 4.0 
[displacement] := 98.0 
[horsepower] := 63.0 
[weight] := 2051.0
[acceleration] := 17.0 
[model year] := 77.0
[origin] := 1.0 

So in the estimator, the first condition is model_year < 80, which is true. So it goes to true node.
Then second condition is horsepower < 69, which is also true. So again it goes to true node.
Then third condition is horsepower <  64, which is again true. So it should give true node value as 0.0231908988, whose node id is 7.

Then why model._Booster.predict(X_test, pred_leaf=True) function giving 8 as output for that estimator?

hcho3:
Fixes #3702 by not rounding up integer threshold for integer features  in JSON dump

### DIFF
--- a/src/tree/tree_model.cc
+++ b/src/tree/tree_model.cc
@@ -6,6 +6,7 @@
 #include <xgboost/tree_model.h>
 #include <sstream>
 #include <limits>
+#include <cmath>
 #include <iomanip>
 #include "./param.h"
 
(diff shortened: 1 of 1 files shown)


Nirmal-Neel:
My concern is that for any floating point number (e.g. 230.5), the json will not reflect the exact number. Instead it will take its floor and will add 1 to it. So we are loosing the information.

Am I correct?

hcho3:
The condition x < 230.5 is equivalent to x < 231 **if the feature x is known to be an integer.** So no information is lost.

Nirmal-Neel:
Got it. Thank you so much. Can you merge the code?

PR #3721, opened 2018-09-25
Implement devices to devices reshard.

trivialfis:
This PR attempts to fix some multi gpu issues.

* Do full data copy when reshard device memory.
* Remove calculating row_segments for gpu_hist and gpu_sketch.

### DIFF
--- a/src/common/common.cu
+++ b/src/common/common.cu
@@ -8,6 +8,8 @@ namespace xgboost {
 int AllVisibleImpl::AllVisible() {
   int n_visgpus = 0;
   try {
+    // When compiled with CUDA but running on CPU only device,
+    // cudaGetDeviceCount will fail.
     dh::safe_cuda(cudaGetDeviceCount(&n_visgpus));
   } catch(const std::exception& e) {
     return 0;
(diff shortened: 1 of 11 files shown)


src/tree/param.h
```diff
@@ -186,7 +186,7 @@ struct TrainParam : public dmlc::Parameter<TrainParam> {
         .describe("gpu to use for single gpu algorithms");
     DMLC_DECLARE_FIELD(n_gpus)
         .set_lower_bound(-1)
-        .set_default(1)
+        .set_default(-1)
```
RAMitchell: Changing the default number of gpus is a big change with significant performance implications. Lets talk more about why you want to do this. The multi-gpu version can be slower depending on the input.

trivialfis: OK, I will leave it unchanged.

trivialfis:
Please don't merge until I got more confidence multi-gpu is actually smoothed.

trivialfis:
@RAMitchell Ready for another review now. :)

hcho3:
@trivialfis Would be nice if we can add unit tests for multi-GPU code. I'm looking into adding a slave worker with multi-GPU instance (p2.8xlarge) to the CI server.

trivialfis:
@hcho3 Cool, I will try to make more tests for multi-gpu. :)

trivialfis:
@hcho3 Do we split single gpu tests and multi-gpu tests, or just try to use all available gpus in the tests?

hcho3:
@trivialfis In #3741, I am using a separate directory `python-mgpu` to indicate which tests should be run with a multi-GPU machine as supposed to a single-GPU machine.

PR #3739, opened 2018-09-28
[jvm-packages] Fixed the distributed updater check

superbobry:
The updater used in distributed training is grow_histmaker and not grow_colmaker as the error message stated prior to this commit.


superbobry:
https://github.com/dmlc/xgboost/blob/72cd1517d6b1d145c34e13a063fadd31b507b01d/src/learner.cc#L222-L224

https://github.com/dmlc/xgboost/blob/72cd1517d6b1d145c34e13a063fadd31b507b01d/src/learner.cc#L167-L169

It seems the C++ part does allow overriding dsplit, but I am not sure what would be the effect of doing that.

CodingCat:
c++ part is the reason for the test being succeeded even with the typo here?

superbobry:
The typo is only activated if the user provided a custom "updater". I think there is no test doing this: https://github.com/dmlc/xgboost/search?l=Scala&q=updater

CodingCat:
yeah, I mean even the user passed in a wrong updater, it would be overwritten by c++

Issue #3749, opened 2018-10-03
[jvm-packages] For training data with group, empty RDD partition threw exception


weitian:
when using spark training on training data with group:
XGBoost.trainDistributed(queriesRDD, params, round, nWorkers, obj, eval, useExternalMemory, missing, hasGroup=true)

The spark job failed with exception:
18/10/03 11:02:04 ERROR executor.Executor: Exception in task 117.0 in stage 2.0 (TID 532)
java.lang.ArrayIndexOutOfBoundsException: 0
	at ml.dmlc.xgboost4j.scala.spark.LabeledPointGroupIterator.next(XGBoost.scala:527)
	at ml.dmlc.xgboost4j.scala.spark.LabeledPointGroupIterator.next(XGBoost.scala:493)
	at scala.collection.Iterator$$anon$13.hasNext(Iterator.scala:462)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
	at org.apache.spark.util.collection.ExternalSorter.insertAll(ExternalSorter.scala:199)
	at org.apache.spark.shuffle.sort.SortShuffleWriter.write(SortShuffleWriter.scala:63)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:96)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:53)
	at org.apache.spark.scheduler.Task.run(Task.scala:108)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:338)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)

The cause is that my training data does have empty partitions.

weitian:
It is a quick fix, will submit a PR later

hcho3:
@weitian Thanks. You should add yourself to CONTRIBUTORS.md when you do so.

hcho3:
Closed by #3750

PR #3750, opened 2018-10-03
[Blocking][jvm-packages] For training data with group, empty RDD partition threw exception (#3749)

weitian:
private var isNewGroup = true

  override def hasNext: Boolean = {
    return base.hasNext || isNewGroup
 }

The initial value for isNewGroup should be false otherwise the hasNext is wrong with empty partition.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostGeneralSuite.scala
+++ b/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostGeneralSuite.scala
@@ -24,6 +24,7 @@ import ml.dmlc.xgboost4j.scala.DMatrix
 import ml.dmlc.xgboost4j.scala.rabit.RabitTracker
 import ml.dmlc.xgboost4j.scala.{XGBoost => SXGBoost, _}
 import org.apache.hadoop.fs.{FileSystem, Path}
+import org.apache.spark.TaskContext
 import org.apache.spark.ml.linalg.Vectors
 import org.apache.spark.sql._
 import org.scalatest.FunSuite
@@ -256,8 +257,16 @@ class XGBoostGeneralSuite extends FunSuite with PerTest {
     }
   }
 
+  test("repartitionForTrainingGroup with group data which has empty partition") {
+    val trainingRDD = sc.parallelize(Ranking.train, 5).mapPartitions(it => {
+      // make one partition empty for testing
+      it.filter(_ => TaskContext.getPartitionId() != 3)
+    })
+    XGBoost.repartitionForTrainingGroup(trainingRDD, 4)
+  }
+
   test("distributed training with group data") {
-    val trainingRDD = sc.parallelize(Ranking.train, 2)
+    val trainingRDD = sc.parallelize(Ranking.train, 5)
     val (booster, metrics) = XGBoost.trainDistributed(
       trainingRDD,
       List("eta" -> "1", "max_depth" -> "6", "silent" -> "1",
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -80,3 +80,4 @@ List of Contributors
   - liuliang01 added support for the qid column for LibSVM input format. This makes ranking task easier in distributed setting.
 * [Andrew Thia](https://github.com/BlueTea88)
   - Andrew Thia implemented feature interaction constraints
+* [Wei Tian](https://github.com/weitian)
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -494,7 +494,7 @@ private[spark] class LabeledPointGroupIterator(base: Iterator[XGBLabeledPoint])
   extends AbstractIterator[XGBLabeledPointGroup] {
 
   private var firstPointOfNextGroup: XGBLabeledPoint = null
-  private var isNewGroup = true
+  private var isNewGroup = false
 
   override def hasNext: Boolean = {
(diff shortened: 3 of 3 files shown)


hcho3:
Can you put your handle to the list of contributors in CONTRIBUTORS.md?

CodingCat:
can we have a test covering that?

BTW, since we have some kind of special logic in ranker training now, I think it makes sense to have a separate suite for all ranking tests

weitian:
> can we have a test covering that?
> 
> BTW, since we have some kind of special logic in ranker training now, I think it makes sense to have a separate suite for all ranking tests

Yes, we should add a XGBoostRankerSuite with more test cases.

For ranker training, "rank:pairwise" is a major reason we chose XGBoost over other machine learning platform.
I suggest we could add a XGBoostRanker to wrap all ranker related logic, too.

This could be a big topic all together, I suggest we create another issue and work on it? what do you think?

hcho3:
I'm curious, is rank:ndcg and rank:map supported in XGBoost4J-Spark? If so, how do they measure in practice compared to rank:pairwise?

CodingCat:
I think XGBoostRanker might be too large for a PR right before the release, but it doesn't prevent it to be a potentially good idea, 

But before you code I would like to discuss over a brief plan, so that we can prevent it from being an over-design...

CodingCat:
for this PR, we can focus on fix the issue and have a test guarding that

CodingCat:
@hcho3 yes, I think we have supported all objective functions https://github.com/dmlc/xgboost/blob/1878e0b6d152ca7bb643e94b7d20f2331cc2ffc3/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/LearningTaskParams.scala#L94-L102

hcho3:
@CodingCat @weitian I am thinking about migrating some tests from Travis to Jenkins, since they tend to fail due to lack of memory or connection problems.

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostGeneralSuite.scala
```diff
@@ -267,4 +268,14 @@ class XGBoostGeneralSuite extends FunSuite with PerTest {
 
     assert(booster != null)
   }
+
+  test("repartitionForTrainingGroup with empty partition") {
+    val trainingRDD = sc.parallelize(Ranking.train, 5).mapPartitions(it => {
+      // make one partition empty for testing
+      it.filter(_ => TaskContext.getPartitionId() != 3)
+    })
+    val traingGroupsRDD = XGBoost.repartitionForTrainingGroup(trainingRDD, 4)
```
CodingCat: can you call through trainDistributed API, which is the unified interface for XGBClassifier/Regressor 

and you may want to change repartitionForTrainingGroup to private

weitian: changed the test to call trainDistributed API.

repartitionForTrainingGroup is package level private already just like the repartitionForTraining method.
API user have no access to it, but it is needed to write unit test.

CodingCat: as the current unit test is kind of coarse-grained, i.e. we didn't test the most nested call of the function, but test everything along the calling path by calling trainDistributed()

I just noticed that the previous ranking trainer test are also calling repartitionForTrainingGroup()

so you should either change everything to call trainDistributed() and make repartitionForTrainingGroup as private, 

or you should have two unit test cases for everything you want to test, by calling trainDistributed and repartitionForTrainingGroup respecitively

weitian: > as the current unit test is kind of coarse-grained, i.e. we didn't test the most nested call of the function, but test everything along the calling path by calling trainDistributed()
> 
> I just noticed that the previous ranking trainer test are also calling repartitionForTrainingGroup()
> 
> so you should either change everything to call trainDistributed() and make repartitionForTrainingGroup as private,
> 
> or you should have two unit test cases for everything you want to test, by calling trainDistributed and repartitionForTrainingGroup respectively

1. There is another unit test calling trainDistributed(): test("distributed training with group data")
2. repartitionForTrainingGroup was defined as 'private[spark]' which follows the convention of the existing method 'private[spark] def repartitionForTraining()'
3. If you take a close look of 'test("repartitionForTrainingGroup with group data")', there is no way to verify whether the grouped data is correct or not by calling  trainDistributed() instead of repartitionForTrainingGroup()

CodingCat: the existence of this bug shows that test("distributed training with group data") doesn't cover the test with data containing empty partition, 

test("repartitionForTrainingGroup with group data")' should be changed to something like test("distributed training with group data containing empty partition")' 

because we have to ensure that  the code path (and the future change on them) in trainDistributed() before  `repartitionForTrainingGroup` has no impact on the things you are testing here... trainDistributed() is actually the entry for everything outside of this file 


if you intentionally want to test the data integrity produced by  `repartitionForTrainingGroup` keep it as private[spark] is fine and you may want to add back your original test in addition to the current one

weitian: add the original test back.

1. test("repartitionForTrainingGroup with group data"): test data integrity after re-grouping.
2. test("repartitionForTrainingGroup with group data which has empty partition"): test corner case with empty partition.
3. test("distributed training with group data which has empty partition"): test from trainDistributed() which is the API to the outside.

CodingCat:
LGTM, merged

CodingCat:
Thanks

PR #3751, opened 2018-10-04
Catch dmlc::Error.

trivialfis:
Fix #3643.


trivialfis:
I switched safe_cuda to use dmlc logging facility without changing the catch statements.  My bad...

hcho3:
Thanks a lot!

PR #3781, opened 2018-10-09
Fix #3523: Fix CustomGlobalRandomEngine for R

### LINKED ISSUE #3523: colsample_bytree seemingly not random
My understanding of `colsample_bytree` is that it randomly samples from the features (columns) for each tree, so it would be a way to limit dimensionality prior to constructing trees. However when I use values < 1, it seems to always take from the first columns (based on whatever percentage) and not randomly. Extending on the sample R code for xgb.train:

```

dtrain <- xgb.DMatrix(agaricus.train$data, label = agaricus.train$label)
dtest <- xgb.DMatrix(agaricus.test$data, label = agaricus.test$label)
watchlist <- list(train = dtrain, eval = dtest)

## A simple xgb.train example:
param <- list(max_depth = 2, eta = 1, silent = 1, nthread = 2, colsample_bytree = 1,
              objective = "binary:logistic", eval_metric = "auc")
bst <- xgb.train(param, dtrain, nrounds = 100, watchlist, verbose = 0)
xgb.importance(model = bst)

param <- list(max_depth = 2, eta = 1, silent = 1, nthread = 2, colsample_bytree = 0.01,
              objective = "binary:logistic", eval_metric = "auc")
bst <- xgb.train(param, dtrain, nrounds = 100, watchlist, verbose = 0)
xgb.importance(model = bst)
```
0.01 is equivalent to saying 1 feature and the last importance matrix always shows `cap-shape=bell`, the first column in the data. I wouldn't expect this if the column sampling were random. I see this with my own data too, where X for colsample_bytree always results in columns in the importance matrix being from the first X * 100% of columns in my data matrix. So if someone uses 0.8, they would essentially be throwing away the last 20% of their columns and I don't think that's what users would want or expect they're doing.

hcho3:
**Diagnosis** Apple Clang's implementation of `std::shuffle` doesn't work correctly when it is run with the random bit generator for R package:
```cpp
CustomGlobalRandomEngine::result_type
CustomGlobalRandomEngine::operator()() {
  return static_cast<result_type>(
      std::floor(unif_rand() * CustomGlobalRandomEngine::max()));
}
```

Minimial reproduction of failure (compile using Apple Clang 10.0):
```cpp
std::vector<int> feature_set(100);
std::iota(feature_set.begin(), feature_set.end(), 0);
    // initialize with 0, 1, 2, 3, ..., 99
std::shuffle(feature_set.begin(), feature_set.end(), common::GlobalRandom());
    // This returns 0, 1, 2, ..., 99, so content didn't get shuffled at all!!!
```

Note that this bug is platform-dependent; it does not appear when GCC or upstream LLVM Clang is used.

**Fix** Use a platform-independent implementation of `std::shuffle`. A header-only library called PCG (Apache license) is included for this purpose.

Closes #3523.

TODO: add a regression test, replace other occurrences of `std::shuffle` in XGBoost codebase

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -237,3 +237,28 @@ test_that("max_delta_step works", {
   expect_true(all(bst1$evaluation_log$train_logloss < bst2$evaluation_log$train_logloss))
   expect_lt(mean(bst1$evaluation_log$train_logloss)/mean(bst2$evaluation_log$train_logloss), 0.8)
 })
+
+test_that("colsample_bytree works", {
+  # Randomly generate data matrix by sampling from uniform distribution [-1, 1]
+  set.seed(1)
+  train_x <- matrix(runif(1000, min = -1, max = 1), ncol = 100)
+  train_y <- as.numeric(rowSums(train_x) > 0)
+  test_x <- matrix(runif(1000, min = -1, max = 1), ncol = 100)
+  test_y <- as.numeric(rowSums(test_x) > 0)
+  colnames(train_x) <- paste0("Feature_", sprintf("%03d", 1:100))
+  colnames(test_x) <- paste0("Feature_", sprintf("%03d", 1:100))
+   dtrain <- xgb.DMatrix(train_x, label = train_y)
+  dtest <- xgb.DMatrix(test_x, label = test_y)
+  watchlist <- list(train = dtrain, eval = dtest)
+   # Use colsample_bytree = 0.01, so that roughly one out of 100 features is
+  # chosen for each tree
+  param <- list(max_depth = 2, eta = 0, silent = 1, nthread = 2,
+                colsample_bytree = 0.01, objective = "binary:logistic",
+                eval_metric = "auc")
+   set.seed(2)
+  bst <- xgb.train(param, dtrain, nrounds = 100, watchlist, verbose = 0)
+  xgb.importance(model = bst)
+  # If colsample_bytree works properly, a variety of features should be used
+  # in the 100 trees
+  expect_gte(nrow(xgb.importance(model = bst)), 30)
+})
--- a/src/common/random.h
+++ b/src/common/random.h
@@ -33,14 +33,14 @@ using RandomEngine = std::mt19937;
 class CustomGlobalRandomEngine {
  public:
   /*! \brief The result type */
-  typedef size_t result_type;
+  using result_type = uint32_t;
   /*! \brief The minimum of random numbers generated */
   inline static constexpr result_type min() {
     return 0;
   }
   /*! \brief The maximum random numbers generated */
   inline static constexpr result_type max() {
-    return std::numeric_limits<size_t>::max();
+    return std::numeric_limits<result_type>::max();
   }
   /*!
    * \brief seed function, to be implemented


hcho3:
@neverfox @CodingCat @Laurae2 @ben519 @khotilov

hcho3:
@trivialfis No idea. One thing I learned is that random functions in `std` are NOT portable in their behavior on different platforms: https://www.reddit.com/r/cpp/comments/7i21sn/til_uniform_int_distribution_is_not_portable/
In the future, if random functions show undesirable behavior on certain platforms, we'll have to replace them with a custom implementation.

trivialfis:
Thanks for the link.  The current tests use random to generate dmatrix, see:

https://github.com/dmlc/xgboost/blob/1db28b8718081a63a55ecc1d861a8d21802d6cb5/tests/cpp/helpers.cc#L109

It may be helpful to make tests reproducible.

hcho3:
@trivialfis Sure, we can look into using a random generation library. [Boost.Random](https://www.boost.org/doc/libs/1_68_0/doc/html/boost_random.html) and [PCG](https://github.com/imneme/pcg-cpp) come to mind. The former increases build complexity greatly, whereas the latter doesn't compile well with MSVC currently. Any other suggestion?

trivialfis:
No, this is very new to me.  I am currently trying to add unittests for gpu-hist,  for now I might just hard code the generated matrix. :(

hcho3:
@trivialfis 
> I might just hard code the generated matrix. :(

You can do that, but it doesn't look elegant to me. Let me look into Boost.Random. It looks like we don't have to include entire Boost, as Boost.Random comes in a separate repository: https://github.com/boostorg/random 
Let me try it. If build is not too painful, I might just include Boost.Random into this PR.

trivialfis:
Great! Thanks.

hcho3:
@trivialfis I gave Boost.Random a shot, but Boost modules are so tightly-coupled that Boost.Random ended up requiring **18** Boost modules :(  Yes, a simple implementation of linear congruential generator is the way to go, since we can have reproducible tests without pulling insane amount of dependencies.

hcho3:
Also need to mention that the 18 Boost modules amount to 443K lines of code. This will potentially increase compilation time a lot. Therefore, I motion against including Boost as dependency.

trivialfis:
Great.  I should push the update tomorrow. :)

khotilov:
I wonder if the 64 bit random integers produced by the custom random engine while the apples clang's shuffle somehow expects them to be 32 bit here could be the root cause?

thvasilo:
It does seem _very_ weird that even such a minimal example wouldn't work, which basically means that `std::shuffle` is broken for Apple's clang, which is unlikely.

What I see in the code for `common::GlobalRandom()` is that it does not initialize the Mersenne twister with any seed, and is also not declared as thread-local.

Have you tried initializing the random engine like:

```c++
    std::random_device rd;
    std::mt19937 g(rd());

    std::shuffle(v.begin(), v.end(), g);
```

I would be very surprised if this didn't work on all platforms. That said, the result might still differ between platforms because it's implementation specific, so if we want consistency then PCG seems like the best option.

In general these things are easy to get wrong, I can definitely recommend Stephan Lavavej's talk on the subject: https://channel9.msdn.com/Events/GoingNative/2013/rand-Considered-Harmful
Especially the second part where he talks about the C++11 random engines and how to use them.

@khotilov I _think_ `std::mt19937` produces 32-bit numbers, there is `std::mt19937_64` for 64-bit.

hcho3:
@thvasilo `std::shuffle` works well if the Mersenne Twister (MT) is used. The bug being discussed is when the MT is not used. When XGBoost is compiled for R support, the macro `XGBOOST_STRICT_R_MODE` is enabled and
 the custom generator `CustomGlobalRandomEngine` is used instead of MT. The shuffle function doesn't work when the custom generator is given.

> also not declared as thread-local.

The random generator is wrapped in thread local storage: https://github.com/dmlc/xgboost/blob/d594b11f3590981c9c9aa3eaa9233027be312c6a/src/common/common.cc

hcho3:
@khotilov This is a possibility. Let me try changing `CustomGlobalRandomEngine` to 32-bit.

thvasilo:
@hcho3 OK, so the problem lies in the `CustomGlobalRandomEngine` gotcha.

tqchen:
Let us try to look more closer at least, it could be very possible due to reason @khotilov suggested and we just have to fix the CustomRandomEngine

hcho3:
@tqchen @khotilov Indeed, changing the return type of `CustomGlobalRandomEngine` to 32-bit fixes the issue. For now, I'll change the return type only for Apple Clang, so that we don't break anything.

hcho3:
The `colsample_bytree` test is failing for R + MSVC. Let's see if we'd need to use 32-bit for Windows as well.

khotilov:
It looks like R has only 32-bit random generators. E.g., here's the code for its default Mersenne Twister generator: https://github.com/wch/r-source/blob/bc124ff19cca95328e114f0fc2f068150ee2aa61/src/main/RNG.c#L633
So it does not really make much sense to cast it to 64 bit integer. So unless there are significant concerns with reproducibility, let's just use uint32_t everywhere. The xgboost's own default random engine is 32 bit std::mt19937 anyway.

PR #3785, opened 2018-10-11
Add basic unittests for gpu-hist method.

trivialfis:
* Split building histogram into separated class.
* Extract `InitCompressedRow` definition.
* Basic tests for gpu-hist.
* Document the code more verbosely.
* Removed HistCutUnit.
* Removed some duplicated copies in GPUHistMaker.

This PR is to add some basic unittests for gpu-hist method, for later refactor (using span, reducing memory requirement, etc.) needs.
Trying to create the unit tests somehow shows that each step in the algorithm depends on different and large amount of data pieces.  I hope we can come up with a better solution to handle these data (and parameters) dependencies before adding more extensive tests.

### DIFF
--- a/tests/cpp/tree/test_gpu_hist.cu
+++ b/tests/cpp/tree/test_gpu_hist.cu
@@ -1,86 +1,396 @@
-
 /*!
- * Copyright 2017 XGBoost contributors
+ * Copyright 2017-2018 XGBoost contributors
  */
+
 #include <thrust/device_vector.h>
 #include <xgboost/base.h>
+#include <random>
 #include "../helpers.h"
 #include "gtest/gtest.h"
 
 #include "../../../src/data/sparse_page_source.h"
 #include "../../../src/gbm/gbtree_model.h"
 #include "../../../src/tree/updater_gpu_hist.cu"
+#include "../../../src/tree/updater_gpu_common.cuh"
 #include "../../../src/common/common.h"
 
 namespace xgboost {
 namespace tree {
-TEST(gpu_hist_experimental, TestSparseShard) {
-  int rows = 100;
-  int columns = 80;
-  int max_bins = 4;
-  auto dmat = CreateDMatrix(rows, columns, 0.9f);
-  common::GHistIndexMatrix gmat;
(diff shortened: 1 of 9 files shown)


tests/cpp/tree/test_gpu_hist.cu
```diff
@@ -1,86 +1,446 @@
...
+    {0.417022, 0.997185},    {0.720325, 0.932557},
+    {0.000114381, 0.128124}, {0.302333, 0.999041},
+    {0.146756, 0.236089},    {0.0923386, 0.396581},
+    {0.18626, 0.387911},     {0.345561, 0.669746},
+    {0.396767, 0.935539},    {0.538817, 0.846311},
+    {0.419195, 0.313274},    {0.68522, 0.524548},
+    {0.204452, 0.443453},    {0.878117, 0.229577},
+    {0.0273876, 0.534414},   {0.670467, 0.913962},
+  };
+  thrust::device_vector<GradientPair> gpair (n_rows);
```
RAMitchell: You can copy a std::vector to device_vector directly by assignment.

tests/cpp/tree/test_gpu_hist.cu
```diff
@@ -1,86 +1,446 @@
...
+
+  ASSERT_EQ(res.size(), 1);
+  ASSERT_EQ(res[0].findex, 7);
+  ASSERT_NEAR(res[0].fvalue, 0.26, xgboost::kRtEps);
+}
+
+TEST(GpuHist, ApplySplit) {
+  GPUHistMaker hist_maker = GPUHistMaker();
+  int constexpr nid = 0;
+  int constexpr n_rows = 16;
```
RAMitchell: My preferred syntax is "constexpr int", this is strange to read.

hcho3:
@trivialfis Nice job creating LCG random generator. Maybe in the future we can re-use it for other purposes.

trivialfis:
@hcho3 maybe we can put it in dmlc-core.  But it's not a serious random generator, so I want to keep it for testing purpose only, in case others mis-use it in real algorithm.

Issue #2754, opened 2017-09-28
XGBoost Regression Accuracies Worst with gpu_hist, Slower with gpu_exact


saj1919:
I wanted to run xgb faster .. so I chose to go with gpu.
But I am getting disappointing results and accuracies. I thought same parameters would work and I have to just change the `tree_update` method. But not the case.

Following parameters are used commonly for all experiments.
```
params = {}
params["objective"] = "reg:linear"
params["booster"] = "gbtree"
params["eval_metric"] = "rmse"
params["eta"] = 0.05
params["subsample"] = 0.25
params["silent"] = 0
params["max_depth"] = 15
params["alpha"] = 0.001
params["min_child_weight"] = 10
param_list = list(params.items())
num_rounds = 100
```
Keeping above constant, I just changed the tree_method. 
Following are the results. 

```
Time : 30 sec.
params["tree_method"] = “gpu_hist" 
[0]	train-rmse:0.473405
[10]	train-rmse:0.283467
[20]	train-rmse:0.169742
[30]	train-rmse:26.6116
[40]	train-rmse:1039.72
[50]	train-rmse:627.03
[60]	train-rmse:441.193
[70]	train-rmse:303.379
[80]	train-rmse:1226.3
[90]	train-rmse:1257.13
[99]	train-rmse:897.348
```

```
Time : 1.5 min
params["tree_method"] = "hist" 
[0]	train-rmse:0.473405
[10]	train-rmse:0.283465
[20]	train-rmse:0.169748
[30]	train-rmse:0.101671
[40]	train-rmse:0.060921
[50]	train-rmse:0.036541
[60]	train-rmse:0.021971
[70]	train-rmse:0.013288
[80]	train-rmse:0.008154
[90]	train-rmse:0.005174
[99]	train-rmse:0.003641
```

```
Time : 5 mins
params["tree_method"] = “gpu_exact" 
[0]	train-rmse:0.473404
[10]	train-rmse:0.283459
[20]	train-rmse:0.169736
[30]	train-rmse:0.101651
[40]	train-rmse:0.060896
[50]	train-rmse:0.036505
[60]	train-rmse:0.021923
[70]	train-rmse:0.013223
[80]	train-rmse:0.008062
[90]	train-rmse:0.005052
[99]	train-rmse:0.003485
```

```
Time : 10 mins
params["tree_method"] = "exact"
[0]	train-rmse:0.473405
[10]	train-rmse:0.283465
[20]	train-rmse:0.169747
[...]

PR #3793, opened 2018-10-15
Address #2754, accuracy issues with gpu_hist

RAMitchell:
With min_child_weight and lambda set to 0 it is possible to have a divide by zero in the weight calculation leading to incorrect results.

This PR prevents the divide by zero, adds tests for this case and fixes a minor compilation issue with Windows.

### DIFF
--- a/src/objective/multiclass_obj.cu
+++ b/src/objective/multiclass_obj.cu
@@ -101,7 +101,8 @@ class SoftmaxMultiClassObj : public ObjFunction {
           for (int k = 0; k < nclass; ++k) {
             // Computation duplicated to avoid creating a cache.
             bst_float p = expf(point[k] - wmax) / static_cast<float>(wsum);
-            const bst_float h = fmax(2.0f * p * (1.0f - p) * wt, kRtEps);
+            const float eps = 1e-16f;
+            const bst_float h = fmax(2.0f * p * (1.0f - p) * wt, eps);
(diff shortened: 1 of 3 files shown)


src/objective/multiclass_obj.cu
```diff
@@ -101,7 +101,8 @@ class SoftmaxMultiClassObj : public ObjFunction {
           for (int k = 0; k < nclass; ++k) {
             // Computation duplicated to avoid creating a cache.
             bst_float p = expf(point[k] - wmax) / static_cast<float>(wsum);
-            const bst_float h = fmax(2.0f * p * (1.0f - p) * wt, kRtEps);
+            const float eps = 1e-16f;
```
trivialfis: There is a eps defined in:
https://github.com/dmlc/xgboost/blob/516457fadcd315554e052b24307392b6aad30e49/include/xgboost/base.h#L203

Not very important. :)

RAMitchell: This is actually the Windows compilation error. For some reason this constant is not defined in device code. Also this number was supposed to be 1e-16f, kRtEps is for something different.

hcho3: @RAMitchell I'd like to get Windows slaves up and running soon, so that we can automatically check compilation issues.

mayer79:
Thx for this fix - very sweet.

Issue #3794, opened 2018-10-15
[MGPU] python-gpu/test_large_sizes.py failed.


trivialfis:
During test, `weight_` and `labels_` inside `Metainfo` are resharded with a different and non-empty distribution, causing the CHECK inside `reshard` to fail.

The cause is pretty simple, the test first run on single GPU:
https://github.com/dmlc/xgboost/blob/516457fadcd315554e052b24307392b6aad30e49/tests/python-gpu/test_large_sizes.py#L100

Then later run on multi-gpu:
https://github.com/dmlc/xgboost/blob/516457fadcd315554e052b24307392b6aad30e49/tests/python-gpu/test_large_sizes.py#L106

While running multi-gpu test, the distribution of `weigth_` and `label_` remains unchanged.  Hence, to make the test pass we simply have to remove one of these two tests.

This issue is a reminder for me to sort out a better approach for multi-gpu.

hcho3:
@trivialfis Does #3738 solve this issue?

trivialfis:
No, this one needs be solved by handling changed parameter.

hcho3:
@trivialfis I'm confused, don't lines 100 and 106 create Booster object from scratch? Which objects are being shared?

hcho3:
Also, should I add this as a blocking issue?

trivialfis:
@hcho3 The weight and label from DMatrix are not resharded.  To fix this we need to come up with a solution handling  changing parameters, in this case, the number of GPUs . I think we can solve it along with the callback issue, so a known bug describing users currently can't change the number of GPUs in between two training/predict sessions might be more appropriate than a quick fix (reshard everything).

hcho3:
So add an item to "known issue" for the next release, then?

trivialfis:
Yes, please.

hcho3:
@trivialfis What would be the work-around for this issue? Should I delete the DMatrix object and re-load?

trivialfis:
Should work, but let me test it.

trivialfis:
@hcho3 Tested. 
Inserting:
```python
            ag_dtrain = xgb.DMatrix(X, y, nthread=40)
```
In between two training sessions in `/test_large_sizes.py` will let the test pass.  The above line essentially recreated DMatrix.  I will focus on #3825 and #3795 next.  After that we can try to find out a solution to these problems.  Feel free to let me know if you have any suggestion. :)

hcho3:
[...]

PR #3803, opened 2018-10-17
Fix #3579: Fix memory leak when learing rate callback is registered

### LINKED ISSUE #3579: XGBoost - hist + learning_rate decay memory usage

hcho3:
**Diagnosis** The learning rate callback function calls `XGBoosterSetParam()` to update the learning rate. The `XGBoosterSetParam()` function in turn calls `Learner::Configure()`, which resets and re-initializes each tree updater, calling `FastHistMaker::Init()`. The `FastHistMaker::Init()` function in turn
re-allocates internal objects that were meant to be recycled across iterations. Thus memory usage increases over time.

**Fix** The learning rate callback should call a new function `XGBoosterUpdateParamInPlace()`. The new function is designed so that no object is re-allocated.

Closes #3579.

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -374,6 +374,21 @@ XGB_DLL int XGBoosterSetParam(BoosterHandle handle,
                               const char *value);
 
 /*!
+ * \brief Update a parameter "in place", i.e. in a way that the underlying
+ *        objects (Booster, Learner, TreeUpdater, etc.) are preserved.
+ *        If updating the parameter would reset or re-create any of the
+ *        underlying objects, this function will fail. This function addresses
+ *        https://github.com/dmlc/xgboost/issues/3579.
+ * \param handle handle
+ * \param name parameter name
+ * \param value value of parameter
+ * \return 0 when success, -1 when failure happens
+ */
+XGB_DLL int XGBoosterUpdateParamInPlace(BoosterHandle handle,
+                                        const char *name,
+                                        const char *value);
+
+/*!
  * \brief update the model in one round using dtrain
  * \param handle handle
  * \param iter current iteration rounds
--- a/include/xgboost/gbm.h
+++ b/include/xgboost/gbm.h
@@ -44,6 +44,16 @@ class GradientBooster {
    */
   virtual void Configure(const std::vector<std::pair<std::string, std::string> >& cfg) = 0;
   /*!
+   * \brief Update a parameter "in place", i.e. in a way that the underlying
+   *        objects (Booster, Learner, TreeUpdater, etc.) are preserved.
+   *        If updating the parameter would reset or re-create any of the
+   *        underlying objects, this function will have no effect. This function
+   *        addresses https://github.com/dmlc/xgboost/issues/3579.
+   * \param name parameter name
+   * \param value value of parameter
+   */
+  virtual void UpdateParamInPlace(const std::string& name, const std::string& value) = 0;
+  /*!
    * \brief load model from stream
    * \param fi input stream.
    */
--- a/include/xgboost/learner.h
+++ b/include/xgboost/learner.h
@@ -54,6 +54,16 @@ class Learner : public rabit::Serializable {
    */
   virtual void Configure(const std::vector<std::pair<std::string, std::string> >& cfg) = 0;
   /*!
+   * \brief Update a parameter "in place", i.e. in a way that the underlying
+   *        objects (Booster, Learner, TreeUpdater, etc.) are preserved.
+   *        If updating the parameter would reset or re-create any of the
+   *        underlying objects, this function will have no effect. This function
+   *        addresses https://github.com/dmlc/xgboost/issues/3579.
+   * \param name parameter name
+   * \param value value of parameter
+   */
+  virtual void UpdateParamInPlace(const std::string& name, const std::string& value) = 0;
+  /*!
    * \brief Initialize the model using the specified configurations via Configure.
    *  An model have to be either Loaded or initialized before Update/Predict/Save can be called.
    */
--- a/include/xgboost/linear_updater.h
+++ b/include/xgboost/linear_updater.h
@@ -12,6 +12,7 @@
 #include <vector>
 #include "../../src/gbm/gblinear_model.h"
 #include "../../src/common/host_device_vector.h"
+#include "../../src/common/common.h"
 
 namespace xgboost {
 /*!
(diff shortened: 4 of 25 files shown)


hcho3:
@RAMitchell @trivialfis @khotilov The fix ended up introducing quite a bit of boilerplate code, since the runtime parameter update would need to be passed all the way to individual parameter structure. So far I'm not able to come up with a better alternative. Any suggestion is welcome.

trivialfis:
Ops, I missed this one, gotta be more careful next time. :)
@hcho3  Looks good to me.

trivialfis:
@hcho3 It seems to be a reasonably quick fix for this issue.  There are some other parameters that we want to implement.  For example, a better control for n_gpus, issues like https://github.com/dmlc/xgboost/issues/3794 really comes down to changing parameters.  I am not yet familiar with many parts of the code base.  See if we can come up something better.

trivialfis:
@hcho3 Is it possible to verify that some parameter needs to be updated during Init()?  So that we don't need to update everything.  It's not clear to me how to do it yet.

hcho3:
Not really, since Init() is supposed to be called once in the first iteration only, and the C++ code isn't aware of the Python callback functions that potentially change training parameters in the middle of the training.

Right now, callbacks call SetParam(), which calls Init() and cause memory leak. The new approach will have callbacks call UpdateParamInPlace() instead.

RAMitchell:
@hcho3 just had a quick glance, can review properly in a few days. I am assuming the problem is the histogram matrix being regenerated every time? The correct solution is probably to remove the histogram matrix from the tree updater and instead associate it with the dmatrix object, which maintains its state after changes in parameters. I realise this is probably a lot more work but it would address the fundamental problem. This must also be a problem in the gpu_hist updater so we should probably do the same there.

trivialfis:
@RAMitchell I was thinking about a more complicated parameters manager in dmlc-core, which can accept some optional functors(callbacks), and call it whenever a corresponding parameter has changed.

hcho3:
@trivialfis I don't think dmlc-core supports callback functionality for parameters. It would be a useful addition.

@RAMitchell I agree that histogram matrix should be merged to DMatrix class. In fact, in the long run we should have DMatrix class manage multiple representations (quantized? CSR / ELLPACK / CSC layouts?). Once this refactor is done, updaters would be "shallow" objects, storing no data. We could also remove UpdateParamInPlace() boilerplate as well. However, this kind of refactor is a lot of work, so I propose to return to it after 0.81 release.

RAMitchell:
@hcho3 you could list this as a known issue and simply delay the fix for future versions. I think the issue is not critical, e.g. you can use the learning rate decay with the exact tree method if necessary.

I am happy to merge this, but the risk is that unforeseen issues arise with the temporary fix and we end up sinking time into the temporary fix and not the long-term fix.

Just some thoughts, go with what you think is best.

hcho3:
@RAMitchell 

> you can use the learning rate decay with the exact tree method if necessary.

In fact, only `hist` is affected by this issue, since it stores extra data inside. Both `exact` and `approx` are not affected.

> the risk is that unforeseen issues arise with the temporary fix and we end up sinking time into the temporary fix and not the long-term fix.

I agree with this sentiment. In fact, I've been hesitant to merge this because of all the boilerplate required. I also agree that we should focus on the long-term, robust fix.

I will close this pull request now. Also, 
1. The 0.81 release note will indicate this as a known issue. Users will be asked to use `exact` / `approx` to avoid memory leak.
2. After 0.81, we should start working on adding multiple representations to DMatrix class. This also has ramifications to #3810 as well.

Issue #3737, opened 2018-09-28
Get NoClassDefFoundError when running xgboost on yarn demo


edwardzjl:
## environment

* **java**: java1.8.0
* **hadoop**: hadoop-2.6.0-cdh5.11.1
* **CLASSPATH**: /opt/cloudera/parcels/CDH/lib/hadoop-mapreduce:/opt/cloudera/parcels/CDH/lib/hadoop/hadoop-common.jar:/usr/java/current/lib:.:
* **LD_LIBRARY_PATH**: /root/programs/xgboost/lib:/opt/cloudera/parcels/CDH/lib64:/usr/java/current/jre/lib/amd64/server:

## problem

followed the instructions on [tutorial](https://xgboost.readthedocs.io/en/latest/tutorials/aws_yarn.html)

after compiling, test/distributed/runtests.sh works fine, but demo job fails immediately after job submitted.

job log says:

> Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/hadoop/io/DataOutputBuffer
	at java.lang.Class.getDeclaredMethods0(Native Method)
	at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
	at java.lang.Class.privateGetMethodRecursive(Class.java:3048)
	at java.lang.Class.getMethod0(Class.java:3018)
	at java.lang.Class.getMethod(Class.java:1784)
	at sun.launcher.LauncherHelper.validateMainClass(LauncherHelper.java:544)
	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:526)
Caused by: java.lang.ClassNotFoundException: org.apache.hadoop.io.DataOutputBuffer
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:338)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 7 more

Is there sth wrong with my CLASSPATH or LD_LIBRARY_PATH settings?

I'm using cdh5.11.1 and the file architecture is different from apache-hadoop. Maybe it's the reason?

hcho3:
[...]

PR #3807, opened 2018-10-18
[jvm-packages] Document incompatibility of CDH Hadoop with XGBoost4J-Spark.

hcho3:
See discussion at #3737.


hcho3:
@CodingCat What do you think? Is it reasonable to require Apache Hadoop specifically?

CodingCat:
according to the discussions there, they are not using xgboost4j but run on yarn directly, it is more about a problem in https://github.com/dmlc/dmlc-core/blob/649be18a8c55c48517861d67158a45dec54992ee/tracker/dmlc_tracker/yarn.py

hcho3:
@CodingCat Got it. Then it's more accurate to say that YARN doesn't work with CDH Hadoop. Closing it for now

CodingCat:
yeah, dmlc-core is compiled with vanilla version of hadoop https://github.com/dmlc/dmlc-core/blob/master/tracker/yarn/pom.xml

they need to build by their own

hcho3:
@edwardzjl

edwardzjl:
Ok I see, maybe we will give it a try later. Thanks for the information.

Issue #3631, opened 2018-08-24
[jvm-packages] xgboost4j 0.80 early stopping


xuchaoqun:
In class XGBoost.java, method 'train', the early stopping logic is too simple, if any criterion value is greater than that before, stop it.  But when evalMetric is 'auc', the stop condition must be reversed. So this code has not been finished yet?
```
        boolean decreasing = true;
        float[] criterion = metrics[metrics.length - 1];
        for (int shift = 0; shift < Math.min(iter, earlyStoppingRound) - 1; shift++) {
          **decreasing &= criterion[iter - shift] <= criterion[iter - shift - 1];**
        }

        if (!decreasing) {
          Rabit.trackerPrint(String.format(
                  "early stopping after %d decreasing rounds", earlyStoppingRound));
          break;
        }
```

CodingCat:
I think we need a parameter passed in that paramBag to indicate which direction we prefer when doing early stopping

PR #3808, opened 2018-10-18
[Blocking][jvm-packages] fix the early stopping feature

CodingCat:
added direction parameter and also fix the judge criterion 

closes https://github.com/dmlc/xgboost/issues/3631 

closes https://github.com/dmlc/xgboost/issues/3140

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -132,6 +132,14 @@ object XGBoost extends Serializable {
     try {
       val numEarlyStoppingRounds = params.get("num_early_stopping_rounds")
         .map(_.toString.toInt).getOrElse(0)
+      if (numEarlyStoppingRounds > 0) {
+        if (!params.contains("metrics_expected_direction") &&
+          (params("metrics_expected_direction") == "asc" ||
+            params("metrics_expected_direction") == "desc")) {
+          throw new IllegalArgumentException(
+            "metrics_expected_direction has to be specified as" + " either asc or desc")
+        }
+      }
       val metrics = Array.tabulate(watches.size)(_ => Array.ofDim[Float](round))
       val booster = SXGBoost.train(watches.train, params, round,
         watches.toMap, metrics, obj, eval,
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -133,11 +133,8 @@ object XGBoost extends Serializable {
       val numEarlyStoppingRounds = params.get("num_early_stopping_rounds")
         .map(_.toString.toInt).getOrElse(0)
       if (numEarlyStoppingRounds > 0) {
-        if (!params.contains("metrics_expected_direction") &&
-          (params("metrics_expected_direction") == "asc" ||
-            params("metrics_expected_direction") == "desc")) {
-          throw new IllegalArgumentException(
-            "metrics_expected_direction has to be specified as" + " either asc or desc")
+        if (!params.contains("maximize_evaluation_metrics")) {
+          throw new IllegalArgumentException("maximize_evaluation_metrics has to be specified")
         }
       }
       val metrics = Array.tabulate(watches.size)(_ => Array.ofDim[Float](round))
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -169,7 +169,7 @@ object XGBoost extends Serializable {
     require(tracker.start(trackerConf.workerConnectionTimeout), "FAULT: Failed to start tracker")
     tracker
   }
-
+  
   /**
    * Check to see if Spark expects SSL encryption (`spark.ssl.enabled` set to true).
    * If so, throw an exception unless this safety measure has been explicitly overridden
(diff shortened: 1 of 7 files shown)


CodingCat:
also @yanboliang , can you take a look?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -132,6 +132,14 @@ object XGBoost extends Serializable {
     try {
       val numEarlyStoppingRounds = params.get("num_early_stopping_rounds")
         .map(_.toString.toInt).getOrElse(0)
+      if (numEarlyStoppingRounds > 0) {
+        if (!params.contains("metrics_expected_direction") &&
+          (params("metrics_expected_direction") == "asc" ||
+            params("metrics_expected_direction") == "desc")) {
```
yanboliang: Can you explain the logic here? I think we just need to check ```params.contains("metrics_expected_direction")```. If users don't set ```metrics_expected_direction``` explicitly, then ```params``` doesn't contain it.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -217,6 +212,45 @@ public static Booster train(
...
+  }
+
+  private static boolean getMetricsExpectedDirection(
+      Map<String, Object> params, int earlyStoppingRound) {
+    boolean metricsExpectedDirection = false;
+    if (earlyStoppingRound > 0) {
+      if (params.get("metrics_expected_direction").equals("asc")) {
+        metricsExpectedDirection = true;
+      } else if (params.get("metrics_expected_direction").equals("desc")) {
+        metricsExpectedDirection = false;
```
yanboliang: The code here is a little confused. If we use boolean to represent ```asc``` or ```desc```, it's better to rename ```metricsExpectedDirection``` as ```metricsExpectedAsc```.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
yanboliang: BTW, I saw there is param validation check at the definition:
```
final val metricsExpectedDirection = new Param[String](this, "metricsExpectedDirection",
    "the expected direction of the metrics' change, asc or desc",
    (value: String) => value == "asc" || value == "desc")
```
This will be called automatically with the two params setting method(MLlib way and XGBoost param map way), so inside the code, we don't need to validate params. I think here we just need to check ```params.contains("metrics_expected_direction")```. If contained, it must be legal.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/LearningTaskParams.scala
```diff
@@ -87,6 +87,13 @@ private[spark] trait LearningTaskParams extends Params {
 
   final def getNumEarlyStoppingRounds: Int = $(numEarlyStoppingRounds)
 
+
+  final val metricsExpectedDirection = new Param[String](this, "metricsExpectedDirection",
+    "the expected direction of the metrics' change, asc or desc",
+    (value: String) => value == "asc" || value == "desc")
```
yanboliang: It's better to make it case-insensitive via ```value.toLowerCase```

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -118,7 +118,7 @@ public static Booster train(
    *                performance on the validation set.
    * @param metrics array containing the evaluation metrics for each matrix in watches for each
    *                iteration
-   * @param earlyStoppingRound if non-zero, training would be stopped
+   * @param earlyStoppingRounds if non-zero, training would be stopped
    *                           after a specified number of consecutive
    *                           increases in any evaluation metric.
```
yanboliang: ```increases``` is out of date? It depends on ```metrics_expected_direction```.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -196,16 +196,11 @@ public static Booster train(
...
         }
-
-        boolean decreasing = true;
-        float[] criterion = metrics[metrics.length - 1];
-        for (int shift = 0; shift < Math.min(iter, earlyStoppingRound) - 1; shift++) {
-          decreasing &= criterion[iter - shift] <= criterion[iter - shift - 1];
-        }
-
-        if (!decreasing) {
+        boolean onTrack = judgeIfTrainingOnTrack(params, earlyStoppingRounds, metrics, iter);
```
yanboliang: Can we put this check inside of ```earlyStoppingRounds > 0```? As if users don't  set early stopping, we don't need to pay effort to check it on track or not.

yanboliang:
BTW, do we need to port ```metrics_expected_direction``` to python xgboost?

hcho3:
@yanboliang Python XGBoost already has a list of metrics to be maximized:
https://github.com/dmlc/xgboost/blob/3564b68b988c373bb5a01812471e3f37a50ccd3c/python-package/xgboost/callback.py#L191-L201

deepxg:
Can I suggest that we make early stopping a policy-based API with some sort of callback? So instead of just a count and a direction, we supply an object which receives the metrics up to the current round, and whatever other information might be useful, and we have a default implementation that is just the normal early stopping after X rounds without increase/decrease. I ask because I suspect there might be circumstances where you care about more than just an absolute number of rounds, or you might for efficiency reasons want to stop training even while still decreasing, just because the returns appear to be diminishing. Externalising the decision and making it configurable would future-proof the API somewhat.

So, just something like `IStopping` with `boolean shouldStopAfter(float[][] metrics)` in the simplest case, but arguments could be made for more context. Making these objects stateful across training runs also simplifies the logic in the early stopping case, as you can just maintain the min/max as you go, and a count of how long ago it was.

CodingCat:
@deepxg it's a good suggestion. However, I think it is out of the scope of this PR:

1. 0.81 (the coming version) is a minor release version so it should not have break changes of previous API

2. and the current implementation is to keep consistent with all other language bindings...

we might evaluate and work on your suggestion in a more comprehensive way which covers all language bindings in the coming releases

CodingCat:
@yanboliang ping for review?

yanboliang:
I think @deepxg have a good point, we should consider it in the roadmap. Thanks.

PR #3818, opened 2018-10-22
update rabit submodule

chenqin:


hcho3:
@chenqin Can you add yourself to `CONTRIBUTORS.md`?

hcho3:
@chenqin Looks like the Java test is failing on Mac OSX: https://travis-ci.org/dmlc/xgboost/jobs/444862707. Can you take a look? You can locally run the Java test by running
```
mvn -q clean install -DskipTests -Dmaven.test.skip
mvn -q test
```

I'll look on my end as well.

CodingCat:
some behavior of poll() in mac makes Spark starts two trackers on the same port....I didn't go into the reason of this but obviously that's the root cause of the failure

chenqin:
looks like the root cause stay in shutdown process, tracker seems wait in big forloop for every slave to send it's shutdown message. For some reason, mac has different behavior and caused test hang for 10 mins before it timeout

chenqin:
Some investigation, looks like I need more time to understand rabit worker shutdown code path as well as recovery logic.

1. xgboost4j use librabit with failure recovery feature by default, it was configured in CMake file in xgboost folder
2. allreducereboust.cc and latest poll based socket comm approach has a bug only in macos where we observed shutdown and stop links of one child will trigger other non connected node try to reconnectlinks and cause infinite loop. In IDE debug mode, where I did step by step execution, this would pass. Seems like a unsynced race condition that only reproduceable in macos.
3. after change from robust to base all reduce implementation in jvm package, there are still some tests failure.

.gitmodules
```diff
@@ -3,7 +3,7 @@
 	url = https://github.com/dmlc/dmlc-core
 [submodule "rabit"]
 	path = rabit
-	url = https://github.com/dmlc/rabit
+	url = https://github.com/chenqin/rabit
```
CodingCat: can you update this to the official repo again? and run some integration test (with large scale dataset)?

make/config.mk
```diff
@@ -52,7 +52,7 @@ endif
 # Rabit library version,
 # - librabit.a Normal distributed version.
 # - librabit_empty.a Non distributed mock version,
-LIB_RABIT = librabit.a
+LIB_RABIT = librabit_base.a
```
CodingCat: this should also be reverted

make/travis.mk
```diff
@@ -20,7 +20,7 @@ USE_AZURE = 0
 # Rabit library version,
 # - librabit.a Normal distributed version.
 # - librabit_empty.a Non distributed mock version,
-LIB_RABIT = librabit.a
+LIB_RABIT = librabit_base.a
```
CodingCat: revert

PR #3820, opened 2018-10-23
Update R-package/R/xgb.ggplot.R (visual)

meztez:
Changed width parameter of var important ggplot from 0.05 to 0.5 to make it more visible when displaying more variables.


hcho3:
Thanks!

PR #3825, opened 2018-10-24
Simplify tree building code.

trivialfis:
Remove:
- GHistRow
- GHistEntry
- GHistIndexRow
- kSimpleStats
- CheckInfo
- SetLeafVec

Moved LossChangeMissing output common since they are not common..
Addressed some compiler warnings, the build message will be cleaner.

There are suggestions from @hcho3 and @RAMitchell we reduce the stored data in algorithm class #3803, and I also want to structure these algorithms in a nicer way, so this would be a multi-stage refactor.

At first I will try to use Span class, if the bound check in it hurts performance too much I will switch to a similar class without bound check.  Along the way I also want to create unittests, which might be a little time consuming.

### DIFF
--- a/src/tree/param.h
+++ b/src/tree/param.h
@@ -256,12 +256,12 @@ struct TrainParam : public dmlc::Parameter<TrainParam> {
 
 // functions for L1 cost
 template <typename T1, typename T2>
-XGBOOST_DEVICE inline static T1 ThresholdL1(T1 w, T2 lambda) {
-  if (w > +lambda) {
-    return w - lambda;
+XGBOOST_DEVICE inline static T1 ThresholdL1(T1 w, T2 alpha) {
+  if (w > + alpha) {
+    return w - alpha;
   }
-  if (w < -lambda) {
-    return w + lambda;
+  if (w < - alpha) {
+    return w + alpha;
   }
   return 0.0;
 }
@@ -271,9 +271,9 @@ XGBOOST_DEVICE inline static T Sqr(T a) { return a * a; }
 
 // calculate the cost of loss function
 template <typename TrainingParams, typename T>
-XGBOOST_DEVICE inline T CalcGainGivenWeight(const TrainingParams &p, T sum_grad,
-                                        T sum_hess, T w) {
-  return -(2.0 * sum_grad * w + (sum_hess + p.reg_lambda) * Sqr(w));
+XGBOOST_DEVICE inline T CalcGainGivenWeight(const TrainingParams &p,
+                                            T sum_grad, T sum_hess, T w) {
+  return -(T(2.0) * sum_grad * w + (sum_hess + p.reg_lambda) * Sqr(w));
 }
 
 // calculate the cost of loss function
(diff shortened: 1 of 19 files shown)


trivialfis:
The things I really want to remove are `ValueConstrains` class and all `CalcGain` `CalcWeight` it uses.  But until we figure out how to get split evaluator working on GPU or find a better replacement, the duplication seems lingering.

trivialfis:
It turns out, span does hurt the performace, following is the result from a simple benchmark.

Using dense dataset Epsilon, with following parameters

- nthread: 1
- tree_method: hist
- objective: binary:logistic

- training-data/full-dataset ratio set to 0.8 .

- CMake flags: -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUSE_OPENMP=ON
- Platform: Ubuntu 18.04 x86-64
- CPU: AMD Ryzen 7 2700 Eight-Core Processor

# With first 5000 lines
Test error converge to 0 around 20 iterations.
## With span
78.37750792503357 sec
78.31656765937805 sec
## Without span
68.44277930259705
69.19314885139465

# With first 50000 lines
## With span
231.58369183540344 sec
237.2560956478119 sec
## Without span
218.65958523750305 sec
215.93592476844788 sec

hcho3:
Sure, I'll look at it soon.

hcho3:
@trivialfis My work got really busy for the last week. The sprint will be over in a few days, then I should be able to add unit tests for the quantile hist updater. Can we merge this after that?

trivialfis:
@hcho3 No problem, take your time.  This is not something critical. :)

hcho3:
I've re-based this PR against the latest master. Let's come back to it after #3957 is merged.

trivialfis:
@hcho3 No problem, I looked into that PR, should be easy to rebase.

trivialfis:
@hcho3 Ready for review.

trivialfis:
Rebased.

trivialfis:
@hcho3 Can you review this?  Since it modifys tree code so any other future PRs can have conflicts with this one.

src/tree/param.h
```diff
@@ -271,54 +271,61 @@ XGBOOST_DEVICE inline static T Sqr(T a) { return a * a; }
...
       return Sqr(sum_grad) / (sum_hess + p.reg_lambda);
     } else {
       return Sqr(ThresholdL1(sum_grad, p.reg_alpha)) /
-             (sum_hess + p.reg_lambda);
+          (sum_hess + p.reg_lambda);
     }
   } else {
     T w = CalcWeight(p, sum_grad, sum_hess);
-    T ret = sum_grad * w + T(0.5) * (sum_hess + p.reg_lambda) * Sqr(w);
+    T ret = CalcGainGivenWeight(p, sum_grad, sum_hess, w);
```
RAMitchell: Are you absolutely sure this is equivalent? I would like a little more assurance here because of this function is wrong we would have a big problem.

RAMitchell: Thanks. Note that due to the non associativity of floating point operations the result may be very slightly different, as a consequence from this commit onwards xgboost may produce very slightly different results.

I think we should still merge this change, if we enforced ourselves to produce identical xgboost results in every release we would never have a chance to improve xgboost.

Issue #3828, opened 2018-10-25
Having trouble save and load model with sklearnAPI


jazztronomers:
Having trouble save and load XGBClassifier() model with sklearnAPI 

--------------------------------------------------------------------------------------
**source:**

from sklearn import datasets
from sklearn.model_selection import train_test_split
import xgboost as xgb

iris = datasets.load_iris()
X = iris.data[:, [2, 3]]
Y = iris.target

x_train, x_test, y_train, y_test = train_test_split(X, Y, test_size=0.3, random_state=0)
model = xgb.XGBClassifier()
model.fit(x_train, y_train)

print(model.score(x_test,y_test))
model.save_model('model.model')

newModel = xgb.XGBClassifier()
newModel.load_model('model.model')
print(newModel.score(x_test,y_test))

--------------------------------------------------------------------------------------
**error:**

Traceback (most recent call last):
  File "D:/pythonProjects/sklearnTest/decisionTreeTest.py", line 20, in <module>
    print(newModel.score(x_test,y_test))
  File "C:\Users\xxxxxxxx\AppData\Local\Programs\Python\Python35\lib\site-packages\sklearn\base.py", line 349, in score
    return accuracy_score(y, self.predict(X), sample_weight=sample_weight)
  File "C:\Users\xxxxxxxx\AppData\Local\Programs\Python\Python35\lib\site-packages\xgboost\sklearn.py", line 608, in predict
    return self._le.inverse_transform(column_indexes)
AttributeError: 'XGBClassifier' object has no attribute '_le'

--------------------------------------------------------------------------------------

could someone help..?

jazztronomers:
solved with pickle lib, it worked well

--------------------------------------------------------------------

pickle.dump(model, open("pima.pickle.dat", "wb"))

# load model from file
loaded_model = pickle.load(open("pima.pickle.dat", "rb"))

# make predictions for test data

print(loaded_model.score(x_test,y_test))

hcho3:
@jazztronomers Sorry for the confusion about `save_model()` / `load_model()`. I will update the Python doc to explicitly recommend pickling.

PR #3829, opened 2018-10-25
Recommend pickling as the way to save XGBClassifier / XGBRegressor / XGBRanker

hcho3:
The `save_model()` and `load_model()` method only saves the part of the model that's common to all language interfaces and do not preserve Python-specific attributes, such as `feature_names`. More crucially, label encoder is not preserved either; this is needed for the scikit-learn wrapper, since you may have string labels.

Fix: Explicitly recommend pickling as the way to save scikit-learn model objects.

Related issue: #3828

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -241,6 +241,13 @@ class XGBModel(XGBModelBase):
         """
         Save the model to a file.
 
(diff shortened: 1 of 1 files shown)


hcho3:
On a longer term, let's consider migrating to an extensible model format to store auxiliary information: JSON, Protocol Buffers, or INI config. LightGBM uses INI-like format, which works well for them.

PR #3835, opened 2018-10-27
merge latest rabit module

chenqin:
cleanup and apply fix to address OSX test failure on
https://github.com/dmlc/xgboost/pull/3818

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -85,3 +85,4 @@ List of Contributors
 * [Andrew Thia](https://github.com/BlueTea88)
   - Andrew Thia implemented feature interaction constraints
 * [Wei Tian](https://github.com/weitian)
+* [Chen Qin] (https://github.com/chenqin)


CodingCat:
can you do some integration test? it's a fundamental change and it is too close to the next release date...we need to be very cautious about whether we want to merge this to 0.81...

based on what we see in rabit, I am very hesitated to merge it to 0.81 which is supposed to be a minor release

chenqin:
make sense, that's also nice to have  regression tests baseline in term of prediction accuracy.

hcho3:
@CodingCat Just to double check, this is going to part of 0.81 release?

CodingCat:
yes, we have multiple tests in production, everything looks fine

hcho3:
Got it. Can you guys give me a short description (a sentence of two) of what this does?

CodingCat:
@chenqin ?

CodingCat:
@chenqin can you add the description here which is to be part of release note?

chenqin:
Fix OS file descriptor limit assertion error on large cluster, replace select based allreduce/broadcast with poll based implementation. Mitigate tracker “thundering herd” issue on large cluster, add exponential backoff retry when workers connect to tracker.

hcho3:
Thanks. I'll add it to the release note.

PR #3848, opened 2018-10-31
Allow XGBRanker sklearn interface to use other xgboost ranking objectives

RAMitchell:
```
clf = xgb.XGBRanker(objective="rank:map")
```
The above code returns "please use XGBRanker for ranking task", despite passing a valid ranking objective.

This PR allows any objective function with the substring "rank:" instead of just "rank:pairwise".


hcho3:
Thanks for correcting my mistake. It will be part of 0.81 release.

Issue #1832, opened 2016-12-02
Bring back distributed column training


AbdealiJK:
It would be nice to have a method to train a model in distributed mode.
Currently, the only way to distribute the model training is by using colmaker, which loads the memory in all the nodes and does the split computation in a distributed manner.

It would be nice to have the distributed colmaker to also load up only chunks of the data (columnar chunks). This was users can have a distributed model training without binning.

tqchen:
Yes! this is an already supported feature, though not used for a while, so I am not sure if it still works, the synchronization is done with bitmap.

We deleted the example at some point https://github.com/dmlc/xgboost/tree/df7c7930d09b707c93c21c3ebadae2e13769d808/multi-node/col-split

relevant code here 
https://github.com/dmlc/xgboost/blob/master/src/tree/updater_colmaker.cc#L781

would be nice to revive it:)

AbdealiJK:
Woah, thanks a lot!

I was afraid I'd have to implement major chunks of it.
I'll look into it after the CPP tests

AbdealiJK:
So the requirements for this task are:
 - Verify it still works, if not ... fix it
 - Write a demo application that uses this
 - Write CPP unittests for the functionality

AbdealiJK:
The DistColMaker part makes sense to me. But after making the first tree, the predictions need to be updated so that the gradients/hessians are calculated using the leaf weights from the first tree.

Here, https://github.com/dmlc/xgboost/blob/f512f08437bfa47621556f51e7e5789133825469/src/learner/learner-inl.hpp#L176 shows that the PredictRaw() and DoBoost() are being called one after the other. Hence, after the first tree is built, the PredictRaw should be updating the preds_ using the tree1.

But, as each node has only a small part of the columns, this cannot happen because PredictRaw() traverses the whole tree with the whole column data and would think that the columns not loaded in this machine are just missing (and hence use the default value) ...

@tqchen How was this being tackled before ? I can't seem to find any code which handles this

tqchen:
The new leaf value is done by GetLeafPosition function.

https://github.com/dmlc/xgboost/blob/6dabdd33e330bb0cda5b91c75516ff73d53ca316/src/gbm/gbtree.cc#L342

I removed this piece of code during last refactor, but this logic can be added by when necessary. Specifically, GetLeafPosition gives the position of each instances of training matrix(which is obtained through the distributed sync procedure)

AbdealiJK:
[...]

PR #3849, opened 2018-11-01
[Blocking] Fix #3840: Clean up logic for parsing tree_method parameter

### LINKED ISSUE #3840: [Blocking] "tree_method=approx" does not invoke the correct updater
First reported by @Laurae2 in #3810.

Currently, setting `tree_method=approx` does not invoke the algorithm in the `HistMaker` class; it invokes `ColMaker` class instead.

Minimal reproduction:
```python
import numpy as np
import xgboost
import time

N = 5*10**3
p = 350
np.random.seed(1)

X =  np.random.randn(N, p)
y =  np.random.randn(N, 1)

param = {'max_depth': 10, 'eta': 1, 'silent': 0, 'objective': 'reg:linear'}
param['eval_metric'] = 'rmse'
dtrain = xgb.DMatrix(X, label=y)
nrounds = 5

param['tree_method'] = 'approx'
model = xgboost.train(param, dtrain, nrounds)
param['tree_method'] = 'exact'
model = xgboost.train(param, dtrain, nrounds)
```
Observe that the two calls to `xgboost.train()` yields the identical sequence of log outputs from the pruner. This is no accident. After injecting extra diagnostic logs into the updater classes, here's what I get:
```
[11:04:20] /xgboost/src/learner.cc:203: tparam_.tree_method = 1
[11:04:20] /xgboost/src/learner.cc:162: ConfigureUpdaters(), cfg_ = {eta:1, max_depth:10, objective:reg:linear, seed:0, silent:0, tree_method:approx, }, tparam_.tree_method = 1
[11:04:20] /xgboost/src/tree/updater_colmaker.cc:27: Using tree_method='exact'
[11:04:20] /xgboost/src/tree/updater_prune.cc:74: tree pruning end, 1 roots, 784 extra nodes, 0 pruned nodes, max_depth=10
[11:04:20] /xgboost/src/tree/updater_prune.cc:74: tree pruning end, 1 roots, 444 extra nodes, 0 pruned nodes, max_depth=10
[11:04:20] /xgboost/src/tree/updater_prune.cc:74: tree pruning end, 1 roots, 312 extra nodes, 0 pruned nodes, max_depth=10
[11:04:20] /xgboost/src/tree/updater_prune.cc:74: tree pruning end, 1 roots, 472 extra nodes, 0 pruned nodes, max_depth=10
[11:04:20] /xgboost/src/tree/updater_prune.cc:74: tree pruning end, 1 roots, 342 extra nodes, 0 pruned nodes, max_depth=10
[11:04:20] /xgboost/src/learner.cc:203: tparam_.tree_method = 2
[11:04:20] /xgboost/src/learner.cc:162: ConfigureUpdaters(), cfg_ = {eta:1, max_depth:10, objective:reg:linear, seed:0, silent:0, tree_method:exact, }, tparam_.tree_method = 2
[11:04:20] /xgboost/src/tree/updater_colmaker.cc:27: Using tree_method='exact'
[11:04:20] /xgboost/src/tree/updater_prune.cc:74: tree pruning end, 1 roots, 784 extra nodes, 0 pruned nodes, max_depth=10
[11:04:20] /xgboost/src/tree/updater_prune.cc:74: tree pruning end, 1 roots, 444 extra nodes, 0 pruned nodes, max_depth=10
[11:04:21] /xgboost/src/tree/updater_prune.cc:74: tree pruning end, 1 roots, 312 extra nodes, 0 pruned nodes, max_depth=10
[11:04:21] /xgboost/src/tree/updater_prune.cc:74: tree pruning end, 1 roots, 472 extra nodes, 0 pruned nodes, max_depth=10
[11:04:21] /xgboost/src/tree/updater_prune.cc:74: tree pruning end, 1 roots, 342 extra nodes, 0 pruned nodes, max_depth=10
```

**So there is no connection between `LearnerTrainParam.tree_method` and the actual updater class being chosen.** On the other hand, `updater` parameter actually determines the choice of updater class.'
Fix. Overhaul initialization logic in `src/learner.cc` so that there is a clear one-to-one correspondence between `tree_method` and `updater` parameters.

@trivialfis

hcho3:
Try to consolidate logic concerning `tree_method` and `updater` into one place. Also use C++11 enum class for `tree_method` for extra safety. (For instance, compilers will generate warning if a `switch` statement doesn't handle all possible values of an enum class. This makes our code future-proof, in case someone decides to add another tree method.)

Closes #3840.

### DIFF
--- a/src/common/enum_class_param.h
+++ b/src/common/enum_class_param.h
@@ -71,12 +71,11 @@ class FieldEntry<EnumClass> : public FieldEntry<int> {  \
     has_default_ = true;  \
     return *this;  \
   }  \
-  /* NOLINTNEXTLINE */  \
   inline void Init(const std::string &key, void *head, EnumClass& ref) {  \
     Super::Init(key, head, *reinterpret_cast<int*>(&ref));  \
   }  \
 };  \
 }  /* namespace parameter */  \
-}  /* namespace dmlc */
+}  /* namespace dmlc */  // NOLINT
 
 #endif  // XGBOOST_COMMON_ENUM_CLASS_PARAM_H_
--- a/src/common/enum_class_param.h
+++ b/src/common/enum_class_param.h
@@ -19,7 +19,7 @@
  *
  * Usage:
  *
- * \code
+ * \code{.cpp}
  *
  *   // enum class must inherit from int type
  *   enum class Foo : int {
@@ -44,7 +44,7 @@
  *   };
  *
  *   DMLC_REGISTER_PARAMETER(MyParam);
- *
+ * \endcode
  */
 #define DECLARE_FIELD_ENUM_CLASS(EnumClass) \
 namespace dmlc {  \
--- a/src/common/enum_class_param.h
+++ b/src/common/enum_class_param.h
@@ -1,0 +1,45 @@
+/*!
+ * Copyright 2015-2018 by Contributors
+ * \file enum_class_param.h
+ * \brief macro for using C++11 enum class as DMLC parameter
+ * \author Hyunsu Philip Cho
+ */
+
+#ifndef XGBOOST_COMMON_ENUM_CLASS_PARAM_H_
+#define XGBOOST_COMMON_ENUM_CLASS_PARAM_H_
+
+#include <dmlc/parameter.h>
+#include <string>
+#include <type_traits>
+
+// specialization of FieldEntry for enum class (backed by int)
+#define DECLARE_FIELD_ENUM_CLASS(EnumClass) \
+template <>  \
+class dmlc::parameter::FieldEntry< EnumClass >  \
+  : public dmlc::parameter::FieldEntry<int> {  \
+ public:  \
+  FieldEntry<EnumClass>() {  \
+    static_assert(  \
+      std::is_same<int, typename std::underlying_type<EnumClass>::type>::value,  \
+      "enum class must be backed by int");  \
+    is_enum_ = true;  \
+  }  \
+  typedef FieldEntry<int> Super;  \
+  void Set(void *head, const std::string &value) const override {  \
+    Super::Set(head, value);  \
+  }  \
+  inline FieldEntry<EnumClass>& add_enum(const std::string &key, EnumClass value) {  \
+    Super::add_enum(key, static_cast<int>(value));  \
+    return *this;  \
+  }  \
+  inline FieldEntry<EnumClass>& set_default(const EnumClass& default_value) {  \
+    default_value_ = static_cast<int>(default_value);  \
+    has_default_ = true;  \
+    return *this;  \
+  }  \
+  inline void Init(const std::string &key, void *head, EnumClass& ref) {  \
+    Super::Init(key, head, *reinterpret_cast<int*>(&ref));  \
+  }  \
+};
+
+#endif  // XGBOOST_COMMON_ENUM_CLASS_PARAM_H_
--- a/src/common/enum_class_param.h
+++ b/src/common/enum_class_param.h
@@ -71,11 +71,11 @@ class FieldEntry<EnumClass> : public FieldEntry<int> {  \
     has_default_ = true;  \
     return *this;  \
   }  \
-  inline void Init(const std::string &key, void *head, EnumClass& ref) {  \
+  inline void Init(const std::string &key, void *head, EnumClass& ref) {  /* NOLINT */  \
     Super::Init(key, head, *reinterpret_cast<int*>(&ref));  \
   }  \
 };  \
 }  /* namespace parameter */  \
-}  /* namespace dmlc */  // NOLINT
+}  /* namespace dmlc */
 
 #endif  // XGBOOST_COMMON_ENUM_CLASS_PARAM_H_
--- a/src/common/enum_class_param.h
+++ b/src/common/enum_class_param.h
@@ -14,9 +14,10 @@
 
 // specialization of FieldEntry for enum class (backed by int)
 #define DECLARE_FIELD_ENUM_CLASS(EnumClass) \
+namespace dmlc {  \
+namespace parameter {  \
 template <>  \
-class dmlc::parameter::FieldEntry< EnumClass >  \
-  : public dmlc::parameter::FieldEntry<int> {  \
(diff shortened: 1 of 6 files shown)


hcho3:
@Laurae2 This should fix the issue with selecting correct `tree_method`.

src/learner.cc
```diff
@@ -479,21 +499,92 @@ class LearnerImpl : public Learner {
...
 
-    if (!p_train->SingleColBlock() && cfg_.count("updater") == 0) {
-      if (tparam_.tree_method == 2) {
-        LOG(CONSOLE) << "tree method is set to be 'exact',"
-                     << " but currently we are only able to proceed with "
-                        "approximate algorithm";
+    const TreeMethod current_tree_method = tparam_.tree_method;
+    if (rabit::IsDistributed()) {
+      /* Choose tree_method='approx' when distributed training is activated */
+      CHECK(tparam_.dsplit != 0)
```
trivialfis: Now that we are here, can dsplit be changed to enum class?

src/learner.cc
```diff
@@ -479,21 +499,92 @@ class LearnerImpl : public Learner {
...
+      switch (current_tree_method) {
+       case TreeMethod::kAuto:
+        LOG(CONSOLE) << "Tree method is automatically set to 'approx' "
+                        "since external-memory data matrix is used.";
+        break;
+       case TreeMethod::kApprox:
+        // things are okay, do nothing
+        break;
+       case TreeMethod::kExact:
+        LOG(CONSOLE) << "Tree method was set to be 'exact', "
```
trivialfis: Not a big issue, but here you use `LOG(CONSOLE)`, in the GPU case you use `LOG(FATAL)`, may I ask about the reason?

src/common/enum_class_param.h
```diff
@@ -0,0 +1,49 @@
...
+
+#ifndef XGBOOST_COMMON_ENUM_CLASS_PARAM_H_
+#define XGBOOST_COMMON_ENUM_CLASS_PARAM_H_
+
+#include <dmlc/parameter.h>
+#include <string>
+#include <type_traits>
+
+// specialization of FieldEntry for enum class (backed by int)
+#define DECLARE_FIELD_ENUM_CLASS(EnumClass) \
```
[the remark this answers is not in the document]
trivialfis: Not a request of change, just a random comment.

src/learner.cc
```diff
@@ -479,21 +499,92 @@ class LearnerImpl : public Learner {
...
-      if (tparam_.tree_method == 2) {
-        LOG(CONSOLE) << "tree method is set to be 'exact',"
-                     << " but currently we are only able to proceed with "
-                        "approximate algorithm";
+    const TreeMethod current_tree_method = tparam_.tree_method;
+    if (rabit::IsDistributed()) {
+      /* Choose tree_method='approx' when distributed training is activated */
+      CHECK(tparam_.dsplit != 0)
+        << "Precondition violated; dsplit cannot be zero in distributed mode";
+      if (tparam_.dsplit == 1) {
```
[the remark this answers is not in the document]
hcho3: @trivialfis According to #1832, `distcol` has been broken. Until we have a functional implementation, I prefer to hide this method.

src/common/enum_class_param.h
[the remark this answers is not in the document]
hcho3: @trivialfis Here, we need macros to define template specialization of `FieldEntry` for the enum class `EnumClass`. Originally, I tried creating a new class with the content of the macro and make subclasses of it, but it didn't work because the methods `add_enum`, `set_default`, and `Init` are not virtual.

src/learner.cc
hcho3: @trivialfis I'm not sure if it's okay to silently change `tree_method` from `gpu_hist` to `approx` or `hist`. Users may find it jarring.

hcho3: My intuition is that it isn't too big of a leap to change from `exact` to `approx`, since both run on CPU. On the other hand, it is a big leap to go from `gpu_hist` to `approx`, since users would presumably expected to benefit from their nice GPUs.

src/learner.cc
hcho3: Yes

hcho3: Done

src/common/enum_class_param.h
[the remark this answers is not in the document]
hcho3: I added a docstring for `DECLARE_FIELD_ENUM_CLASS` with an example.

src/learner.cc
```diff
@@ -151,7 +152,7 @@ DMLC_REGISTER_PARAMETER(LearnerTrainParam);
  * \brief learner that performs gradient boosting for a specific objective
  * function. It does training and prediction.
  */
-class LearnerImpl : public Learner {
+class LearnerImpl : public Learner, public LearnerTestHook {
```
trivialfis: This feels weird to me.  If I were you I would first add `std::map<std::string, std::string> Learner::GetConfiguration() const` abstract method to `Learner`, let `LearnerImpl` to override it.  Then call `Learner::Create()` to make a learner pointer in testing code, which can be used to obtain the configuration you are interested.

But that's just personal preference.

trivialfis: Another option would be declare `LearnerImpl` in a header, then create a `LearnerMock` in testing code, which inherits `Learner`.  Then you can use `LearnerMock` to access the internal of `LearnerImpl`.

src/common/enum_class_param.h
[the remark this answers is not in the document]
trivialfis: Nice.

src/learner.cc
hcho3: @trivialfis The options you presented are all valid. I chose this approach so that the change to `Learner` and `LearnerImpl` classes is minimized. Also, it would be easy to add access (from the test) to more internal stuff in Learner.

trivialfis:
@hcho3 Hi, I'm a little bit confused by the configuration logic, does linear models require configuration?

Another thing, updaters for linear model are `gpu_coord_descent`, `coord_descent` and `shotgun`.  which should be configured but configuring `updater` is marked as dangerous action.

hcho3:
@trivialfis Yes it was my oversight. The warning shouldn't have been displayed for gblinear.

Issue #3850, opened 2018-11-01
gpu_id error in R when it's non-zero(xgboost 0.81.0.1)


joegaotao:
I have installed the newest multi-GPU xgboost from the source code(0.81.01) and the server has 8 GPUS. But when I change `gpu_id` to non-zero, there are some errors. Parameter `n_gpus` is ok. the test code: 

```
library('xgboost')
# Simulate N x p random matrix with some binomial response dependent on pp columns
set.seed(111)
N <- 1000000
p <- 50
pp <- 25
X <- matrix(runif(N * p), ncol = p)
betas <- 2 * runif(pp) - 1
sel <- sort(sample(p, pp))
m <- X[, sel] %*% betas - 1 + rnorm(N)
y <- rbinom(N, 1, plogis(m))

tr <- sample.int(N, N * 0.75)
dtrain <- xgb.DMatrix(X[tr,], label = y[tr])
dtest <- xgb.DMatrix(X[-tr,], label = y[-tr])
wl <- list(train = dtrain, test = dtest)

# An example of running 'gpu_hist' algorithm
# which is
# - similar to the 'hist'
# - the fastest option for moderately large datasets
# - current limitations: max_depth < 16, does not implement guided loss
# You can use tree_method = 'gpu_exact' for another GPU accelerated algorithm,
# which is slower, more memory-hungry, but does not use binning.
param <- list(objective = 'reg:logistic', eval_metric = 'auc', subsample = 0.5, nthread = 4,
  max_bin = 64, tree_method = 'gpu_hist', gpu_id = 1)
pt <- proc.time()
bst_gpu <- xgb.train(param, dtrain, watchlist = wl, nrounds = 50)
proc.time() - pt
```
error:

```
terminate called after throwing an instance of 'dmlc::Error'
  what():  [12:58:22] /home/tgao/repos/xgboost/include/xgboost/./../../src/common/common.h:200: Check failed: Contains(device) 

Stack trace returned 10 entries:
[bt] (0) /home/tgao/rpkg/xgboost/libs/xgboost.so(dmlc::StackTrace[abi:cxx11]()+0x5a) [0x7f6d5395a0da]
[bt] (1) /home/tgao/rpkg/xgboost/libs/xgboost.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x28) [0x7f6d5395abd8]
[...]

PR #3851, opened 2018-11-01
Fix specifying gpu_id, add tests.

trivialfis:
Address #3850 .

I don't have the multi-gpu machine for testing currently.

### DIFF
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -194,8 +194,9 @@ class GPUSet {
   }
 
   size_t Size() const {
-    size_t res = *devices_.end() - *devices_.begin();
-    return res < 0 ? 0 : res;
+    GpuIdType size = *devices_.end() - *devices_.begin();
+    GpuIdType res = size < 0 ? 0 : size;
+    return (size_t)res;
   }
 
   /*
@@ -208,7 +209,7 @@ class GPUSet {
    * `Index' converts a device id to a zero-based index.
    */
   GpuIdType DeviceId(size_t index) const {
-    GpuIdType result = *devices_.begin() + index;
+    GpuIdType result = *devices_.begin() + static_cast<GpuIdType>(index);
     CHECK(Contains(result)) << "\nDevice " << result << " is not in GPUSet."
                             << "\nIndex: " << index
                             << "\nGPUSet: (" << *begin() << ", " << *end() << ")"
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -196,7 +196,7 @@ class GPUSet {
   size_t Size() const {
     GpuIdType size = *devices_.end() - *devices_.begin();
     GpuIdType res = size < 0 ? 0 : size;
-    return (size_t)res;
+    return static_cast<size_t>(res);
   }
 
   /*
@@ -220,7 +220,7 @@ class GPUSet {
     CHECK(Contains(device)) << "\nDevice " << device << " is not in GPUSet."
                             << "\nGPUSet: (" << *begin() << ", " << *end() << ")"
                             << std::endl;
-    size_t result = device - *devices_.begin();
+    size_t result = static_cast<size_t>(device - *devices_.begin());
     return result;
   }
 
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -147,61 +147,85 @@ struct AllVisibleImpl {
  */
 class GPUSet {
  public:
+  using GpuIndex = int;
+  static constexpr GpuIndex kAll = -1;
+
   explicit GPUSet(int start = 0, int ndevices = 0)
       : devices_(start, start + ndevices) {}
 
   static GPUSet Empty() { return GPUSet(); }
 
-  static GPUSet Range(int start, int ndevices) {
-    return ndevices <= 0 ? Empty() : GPUSet{start, ndevices};
+  static GPUSet Range(GpuIndex start, GpuIndex n_gpus) {
+    return n_gpus <= 0 ? Empty() : GPUSet{start, n_gpus};
   }
-  /*! \brief ndevices and num_rows both are upper bounds. */
-  static GPUSet All(int ndevices, int num_rows = std::numeric_limits<int>::max()) {
-    int n_devices_visible = AllVisible().Size();
-    if (ndevices < 0 || ndevices >  n_devices_visible) {
-      ndevices = n_devices_visible;
+  /*! \brief n_gpus and num_rows both are upper bounds. */
+  static GPUSet All(GpuIndex gpu_id, GpuIndex n_gpus,
+                    GpuIndex num_rows = std::numeric_limits<GpuIndex>::max()) {
+    CHECK_GE(gpu_id, 0) << "gpu_id must be >= 0.";
+    CHECK_GE(n_gpus, -1) << "n_gpus must be >= -1.";
+
+    GpuIndex const n_devices_visible = AllVisible().Size();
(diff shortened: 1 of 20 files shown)


hcho3:
@trivialfis We may want to list this as a known issue. Do you have a short description of the issue?

hcho3:
@trivialfis Got it. I'll trust your judgment on this.

trivialfis:
@hcho3 Tested with the R script provided by @joegaotao . Thanks.  I think it should now be fixed.

@RAMitchell But it's not clear to me, what's the point of having `normalised_device_index`.  Can we just use the device index obtained from CUDA?

trivialfis:
@hcho3 Not really.  Just talked to @RAMitchell , the current fix will disable dividing training sessions across GPUs.   Although the script won't fail but it defeats the purpose of specifying `gpu_id`.

trivialfis:
Let me keep trying.

trivialfis:
@RAMitchell Ready for a review.  I haven't figure out how to make a unified way to allocate GPU, but the current fix won't add any burden to the code base, see if we can merge it.

src/common/common.h
```diff
@@ -169,14 +169,15 @@ class GPUSet {
...
-  }
+
   /*! \brief Counting from gpu_id */
   GPUSet Normalised(int gpu_id) const {
+    int n_devices_visible = AllVisible().Size();
+    CHECK(Size() + gpu_id <= n_devices_visible)
+        << "n_gpu + gpu_id should less than or equal to total number of available devices."
+        << "\n n_gpu: " << Size()
+        << ", gpu_id: " << gpu_id
+        << ", number of available devices: " << n_devices_visible;
```
RAMitchell: Lets remove Unnormalised(), I think it is wrong as the range can be larger than the number of GPUs. This unnormalised/normalised terminology is confusing and I can see it somehow has different meanings across different parts of the code. From now on whenever a device index is referred to it means the physical device ordinal of the GPU.

src/common/common.h
```diff
@@ -169,14 +169,15 @@ class GPUSet {
...
-  /*! \brief Ensure gpu_id is correct, so not dependent upon user knowing details */
-  static int GetDeviceIdx(int gpu_id) {
-    auto devices = AllVisible();
-    CHECK(!devices.IsEmpty()) << "Empty device.";
-    return (std::abs(gpu_id) + 0) % devices.Size();
-  }
+
   /*! \brief Counting from gpu_id */
   GPUSet Normalised(int gpu_id) const {
+    int n_devices_visible = AllVisible().Size();
```
RAMitchell: Let's rename this function , without unnormalised/normalised terminology.

src/common/common.h
```diff
@@ -197,7 +198,9 @@ class GPUSet {
   bool IsEmpty() const { return Size() == 0; }
   /*! \brief Get un-normalised index. */
```
RAMitchell: As above let's remove the function Index().

src/tree/updater_gpu_hist.cu
```diff
@@ -251,15 +251,15 @@ struct DeviceHistogram {
   thrust::device_vector<GradientPairSumT::ValueT> data;
   const size_t kStopGrowingSize = 1 << 26;  // Do not grow beyond this size
   int n_bins;
-  int device_idx;
+  int normalised_device_idx;
```
RAMitchell: This should all be device_idx, as before let's remove all normalised/unnormalised terminology.

src/common/transform.h
```diff
@@ -125,19 +125,22 @@ class Transform {
...
+
+      // Extract index to deal with possible old OpenMP.
+      size_t device_beg = *(devices.begin());
+      size_t device_end = *(devices.end());
 #pragma omp parallel for schedule(static, 1) if (devices.Size() > 1)
-      for (omp_ulong i = 0; i < devices.Size(); ++i) {
-        int d = devices.Index(i);
+      for (omp_ulong device = device_beg; device < device_end; ++device) {  // NOLINT
         // Ignore other attributes of GPUDistribution for spliting index.
+        int d_unnormalised = devices.Index(device);
```
RAMitchell: In situations like this you need to have some index beginning from 0 you could get it from the enclosing for loop:
```
 for (omp_ulong i = 0; i < devices.Size(); ++i) {
    auto device_idx = devices[i];
    size_t shard_size =
            GPUDistribution::Block(devices).ShardSize(range_size, i);
```

src/tree/updater_gpu_hist.cu
```diff
@@ -1336,6 +1338,7 @@ class GPUHistMaker : public TreeUpdater {
   common::Monitor monitor_;
   dh::AllReducer reducer_;
   std::vector<ValueConstraint> node_value_constraints_;
+  /*! List storing normalised device index*/
```
RAMitchell: We should be able to remove this.

trivialfis:
@hcho3 I can't address these problems today.  I need some more time to completely remove the normalise/unnormalise stuffs.  So known issue it is.

hcho3:
@trivialfis Okay, I'll add it to the known issues. How should I summarize this issue (in a sentence)? How about "`gpu_id` cannot be set when one GPU is used"?

trivialfis:
"Specifying gpu-id is not yet supported." is more realistic from my perspective.

trivialfis:
Due to  #3794 , even if we try to support specifying `gpu_id`, users still need to have a copy their dataset for each GPU.

trivialfis:
@RAMitchell Ready for another review now. :)

src/common/common.h
```diff
@@ -147,61 +147,85 @@ struct AllVisibleImpl {
...
+  /*
+   * By default, we have two configurations of identifying device, one
+   * is the device id obtained from `cudaGetDevice'.  But we sometimes
+   * store objects that allocated one for each device in a list, which
+   * requires a zero-based index.
+   *
+   * Hence, `DeviceId' converts a zero-based index to actual device id,
+   * `Index' converts a device id to a zero-based index.
+   */
+  GpuIndex DeviceId(GpuIndex index) const {
```
RAMitchell: Suggest index is NOT of type GpuIndex. I assume GpuIndex is intended to refer to a physical device?

src/common/common.h
```diff
@@ -147,61 +147,85 @@ struct AllVisibleImpl {
...
+    GpuIndex result = *devices_.begin() + index;
+    CHECK(Contains(result)) << "\nDevice " << result << " is not in GPUSet."
+                            << "\nIndex: " << index
+                            << "\nGPUSet: (" << *begin() << ", " << *end() << ")"
+                            << std::endl;
+    return result;
+  }
+  GpuIndex Index(GpuIndex device) const {
```
RAMitchell: As above maybe this function should return an integer and not GpuIndex? You should be clear about what this type is and not use it for both these purposes.

src/common/common.h
trivialfis: Done. :)

src/common/common.h
trivialfis: Renamed `GpuIndex` to `GpuIdType', and use it along with `size_t` to indicate different types of indices.

hcho3:
@trivialfis I just added a new multi-GPU worker, but somehow I broke the old one. Let me reboot it.

trivialfis:
@hcho3 Thanks. :)

hcho3:
@trivialfis Both workers are now up and running.

PR #3853, opened 2018-11-01
[jvm-packages] Fix JVM doc build

hcho3:
To get around of the bug https://issues.apache.org/jira/browse/SUREFIRE-1588, set `useSystemClassLoader=false`.

### DIFF
--- a/jvm-packages/pom.xml
+++ b/jvm-packages/pom.xml
@@ -319,6 +319,7 @@
                 <version>2.19.1</version>
                 <configuration>
                     <skipTests>false</skipTests>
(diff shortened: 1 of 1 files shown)


hcho3:
@CodingCat @yanboliang Is this change reasonable?

CodingCat:
I think it does no harm to jvm building, but what's the error we see in doc side?

hcho3:
@CodingCat See https://xgboost-ci.net/blue/organizations/jenkins/restricted-xgboost/detail/master/81/pipeline. It's due to a bug in the Maven surefire plugin.

CodingCat:
I see....I am fine with this patch

PR #3856, opened 2018-11-02
Fix #3342 and h2oai/h2o4gpu#625: Save predictor parameters in model file

### LINKED ISSUE #3342: multi-GPU training is not adaptable to other GPU counts or CPU
It appears that predict uses the same GPU setup as train.  So if I do train on 3 GPUs, I can't do predict on that model with only 1 GPU or 2 GPUs (say if I pickled the model and then from that run on 1GPU).  I can force it to use CPU by changing the predictor.

I saw in the code that for predict it is not yet implemented to use n_gpus or gpu_id, but can it be done?  The ultimate goal is then to be able to do take any pickled state (that was trained on any system type or GPU count) and predict on that model with any system type.

hcho3:
This allows pickled models to retain predictor attributes, such as 'predictor' (whether to use CPU or GPU) and 'n_gpu' (number of GPUs to use). Related: h2oai/h2o4gpu#625

Closes #3342.

TODO. Write a test.


hcho3:
@pseudotensor It turns out we can actually save predictor parameters inside the binary model (and thus pickles).

pseudotensor:
Great!

hcho3:
@pseudotensor Can you take a quick look at the test and see if it's sufficient?

hcho3:
`GPUtil` somehow fails to measure GPU memory usage. I'll just directly query the kind of predictor being used.

pseudotensor:
Sure.   BTW, here's the code fragment we have to use to restore the booster stuff:

```
    # The booster, after being restored from pickled file, doesn't remember the predictor-specific C++ parameters: n_gpus, gpu_id, silent
    # So restore the model settings to the booster before running predict
    def restore_booster_params(self, model, experiment_id):
        booster = model.get_booster()
        attrs = [a for a in dir(model) if
                 not a.startswith('__') and not a.startswith('_') and not callable(getattr(model, a))]
        n_gpus = model.kwargs['n_gpus']
        skip_gpu_param = not config.use_gpu_for_prediction or \
                         get_num_gpus_per_experiment() == 0 or \
                         get_num_gpus_per_experiment() != n_gpus  # FIXME - shouldn't have to have exactly same number of GPUs

        for attr in attrs:
            value = getattr(model, attr)
            str_attr = str(attr)
            if str_attr == 'nthread':
                str_attr = 'n_jobs'
            if str_attr == 'seed':
                str_attr = 'random_state'
            if "gpu" in str_attr or isinstance(value, str) and "gpu" in value:
                if skip_gpu_param:
                    continue
            if str_attr != 'booster':
                booster.set_param({str_attr: getattr(model, str_attr)})
        for k, v in zip(model.kwargs.keys(), model.kwargs.values()):
            if isinstance(k, str) and "gpu" in k or isinstance(v, str) and "gpu" in v:
                if skip_gpu_param:
                    continue
            booster.set_param({k: v})
```

hcho3:
@pseudotensor Thanks, that was helpful. This PR handles `predictor`, `n_gpu`, and `gpu_id`. I didn't include `n_jobs` in this PR because pickling would preserve it:
https://github.com/dmlc/xgboost/blob/e04ab56b57a6e9b3eb0df0c38df8cb0a564e2ab4/python-package/xgboost/sklearn.py#L158-L161

I will go ahead and fix the test.

hcho3:
@trivialfis FYI, I took your previous advice and added abstract method `GetConfigurationArguments()` to `Learner` interface.

PR #3869, opened 2018-11-05
Make C++ unit tests run and pass on Windows

### LINKED ISSUE #3038: [jvm-packages] Location of dtrain_cache files
I'm using xgboost in a Spark environment and while it works just fine, I find that it creates lots of "dtrain_cache" files in the current directory.
In some cases, that directory is not writable which causes the training to fail. 
I believe it should be possible to specify the location where those files are created, but I could not find it in the documentation.

Here is a sample code that I use:

    val xgModel = XGBoost.trainWithDataFrame(
      sourceData.withColumn(labelColNormalizedName, labelColNormalized),
      paramMap,
      round = numRound,
      nWorkers = numWorkers,
      useExternalMemory = true,
      labelCol = labelColNormalizedName)

With this I get files created that follow this naming scheme:

    MyProgram-%d-dtrain_cache-%d
    MyProgram-%d-dtrain_cache-%d.col.meta
    MyProgram-%d-dtrain_cache-%d.col.page
    MyProgram-%d-dtrain_cache-%d.row.page
    MyProgram-%d-dtest_cache-%d
    MyProgram-%d-dtest_cache-%d.row.page

Looking at the source, it seems that it comes from `SparsePageSource::Create` but I can't find the link between this method and the `XGBoost.trainWithDataFrame` method and its params.
Any hints would be most welcome.

hcho3:
* Fix logic for external memory. The letter ':' is part of drive letter, so remove the drive letter before splitting on ':'.
* Cosmetic syntax changes to keep MSVC happy.

Fixes #3038.

### DIFF
--- a/src/data/sparse_page_source.cc
+++ b/src/data/sparse_page_source.cc
@@ -6,10 +6,35 @@
 #include <dmlc/timer.h>
 #include <xgboost/logging.h>
 #include <memory>
+#include <vector>
+#include <string>
+#include <locale>
 
 #if DMLC_ENABLE_STD_THREAD
 #include "./sparse_page_source.h"
 #include "../common/common.h"
+
+namespace {
+
+// Split a cache info string with delimiter ':'
+// If cache info string contains drive letter (e.g. C:), exclude it before splitting
+inline std::vector<std::string>
+GetCacheShards(const std::string& cache_info) {
+  if (cache_info.length() < 2) {
+    return {cache_info};
+  }
+  if (std::isalpha(cache_info[0], std::locale::classic())
+      && cache_info[1] == ':') {
+    std::vector<std::string> cache_shards
+      = xgboost::common::Split(cache_info.substr(2), ':');
+    cache_shards[0] = cache_info.substr(0, 2) + cache_shards[0];
+    return cache_shards;
+  } else {
+    return xgboost::common::Split(cache_info, ':');
+  }
+}
+
+}  // namespace anonymous
 
 namespace xgboost {
 namespace data {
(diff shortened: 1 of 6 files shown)


hcho3:
It turns out `std::regex` is not supported in GCC 4.8.2.

src/data/sparse_page_source.cc
```diff
@@ -6,19 +6,44 @@
...
+namespace {
+
+// Split a cache info string with delimiter ':'
+// If cache info string contains drive letter (e.g. C:), exclude it before splitting
+inline std::vector<std::string>
+GetCacheShards(const std::string& cache_info) {
+  if (cache_info.length() < 2) {
+    return {cache_info};
+  }
+  if (std::isalpha(cache_info[0], std::locale::classic())
```
trivialfis: I'm not sure, is it a guarantee that path like "A:xxx" and A must be drive letter?

trivialfis: Currently I'm quite limited to Linux.  I will need to spend some more time on Windows. :(

hcho3: In Windows, the colon is a reserved letter and thus it cannot appear inside the path. Only time it appears is as part of drive letter. See https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file#naming-conventions for more information about Windows paths.

To be safe, I'll add a guard to this check so that it only runs for Windows.

trivialfis: Thanks for the clarification.  With the guard then PR looks good to me. :)

Issue #3874, opened 2018-11-07
using booster='gblinear' will trigger GPU(R-xgboost 0.81)? 


joegaotao:
When I test gblinear and I found the default model would trigger GPU usage while I didn't specify any GPU parameters.

from R-package demo with minor change
```
library('xgboost')

# Simulate N x p random matrix with some binomial response dependent on pp columns
set.seed(111)
N <- 1000000
p <- 50
pp <- 25
X <- matrix(runif(N * p), ncol = p)
betas <- 2 * runif(pp) - 1
sel <- sort(sample(p, pp))
m <- X[, sel] %*% betas - 1 + rnorm(N)
y <- rbinom(N, 1, plogis(m))

tr <- sample.int(N, N * 0.75)
dtrain <- xgb.DMatrix(X[tr,], label = y[tr])
dtest <- xgb.DMatrix(X[-tr,], label = y[-tr])
wl <- list(train = dtrain, test = dtest)

param <- list(objective = 'reg:linear', subsample = 0.5,  nthread = 10L, 
  booster = "gblinear", lambda = 1e-4, updater = "coord_descent", feature_selector = "random")
pt <- proc.time()
bst_gpu <- xgb.train(param, dtrain, watchlist = wl, nrounds = 50)
proc.time() - pt
```
I found GPU has a process (gpu_id = 0)and memory usage

I just compile R-xgboost (0.81) with multi-GPU.

```
git clone --recursive https://github.com/dmlc/xgboost
mkdir build
cd build
cmake .. -DUSE_CUDA=ON -DUSE_NCCL=ON -DNCCL_ROOT=/usr/local/nccl -DR_LIB=ON
make install -j4
```

trivialfis:
Yes.  If XGBoost is compiled with CUDA support, it will try to run objective functions on GPU (1 GPU by default).  This reminds me that, I might need to add a switch to that.

joegaotao:
OK,  I find there is a src/gpu_coord_descent.cu, and `updater = "gpu_coord_descent"` also makes GPU works, but the [xgboost doc](https://xgboost.readthedocs.io/en/latest/parameter.html#parameters-for-linear-booster-booster-gblinear) seems to miss that specification, maybe should add it to the doc?

trivialfis:
I would love to organize the doc a bit.  It will also provide a good mental model for me to keep going.  Thanks for reminding.

PR #3452, opened 2018-07-06
add sum_instance of each node

USCYuandaDu:
Add  sum_instance into RTreeNodeState, this record the amount of data split into this node. Currently only suitable with gradient boosting tree. And add two tests.

PR #3878, opened 2018-11-08
Enable running objectives with 0 GPU.

trivialfis:
* Enable 0 GPU for objectives.
* Add doc for GPU objectives.
* Fix some objectives defaulted to running on all GPUs.

Brought up by #3874 .

### DIFF
--- a/doc/gpu/index.rst
+++ b/doc/gpu/index.rst
@@ -117,13 +117,9 @@ Most of the objective functions implemented in XGBoost can be run on GPU.  Follo
 | rank:map        | |cross|     |
 +-----------------+-------------+
 
-For multi-gpu support, objective functions also honor the ``n_gpus``
-parameter, which, by default is set to 1.  XGBoost has modularized
-implementation of objective functions. Such design enables flexibility of
-XGBoost, but prevents further interaction between different modules.  So even
-the tree method is selected to be a CPU algorithm, the objective function still
-try to run on GPU, which sometimes could be problematic due memory copying
-between host and devices.  To disable running on GPU, just set ``n_gpus`` to 0.
+For multi-gpu support, objective functions also honor the ``n_gpus`` parameter,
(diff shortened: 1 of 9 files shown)


trivialfis:
@hcho3  I have some difficulties dealing with the parameters.  I want to organize the parameters a little bit so that I can gain more confidence I'm not breaking anything,  but the first thing I encountered is how to save and load these parameters.  So I spent some time on learning how XGBoost saves its model in C++ level and found out that current binary dumping is not exactly formatted. In #3886, you talked about:

> We may want to consider a next generation model format

 Combined with `dump_model` in Python, I want to know what's your idea about next generation model format.  Thanks. :)

hcho3:
@trivialfis The current binary format saves LearnerModelParam, Objective, Metrics, and the decision trees (RegTrees). Extra attributes (string key-value pairs) are also saved. Other parameters are not saved, with the exception of predictor parameters I added in #3856. In general, parameters that are specific to training, such as LearnerTrainParam and TrainParam, are not saved.
The issue is backward compatibility. We are free to change any unsaved parameters but we are not free to change any saved parameters (exception: you can add extra Booster attributes). This is why we cannot change type of num_feature nor add instance count to tree nodes (#3452)

I'd love to adopt a new model format where we can add new fields easily without breaking backward compatibility. Protobuf, XML, JSON would qualify. But the transition is not going to be easy. We'll need to release 1.0 and provide converters for old model files.

hcho3:
A hacky way is to simply append all extra new fields to the end of the legacy format, but it doesn't handle per-tree or per-node information well. But for parameters, this hack may work.

trivialfis:
@hcho3 Thanks for the explanation.  Since we just released the 0.81,  we can consider organizing things a little bit before the next release hit.  I will try to learn more and think about it.  If you have some interesting ideas, feel free to open an issue so more voice can join. :)

Issue #3884, opened 2018-11-08
Invalid Parameter format for num_feature expect int but value='4294967258'


newnovice01:
[18:18:13] start ip-172-31-20-145:1
[18:18:13] Load part of data 1 of 4 parts
[18:18:45] Load part of data 1 of 4 parts
terminate called after throwing an instance of 'dmlc::ParamError'
  what():  Invalid Parameter format for num_feature expect int but value='4294967258'

This is with training of Criteo TB dataset for 1 day's worth in libsvm format. I was also using 1 days worth of data for test. 

https://www.csie.ntu.edu.tw/~cjlin/libsvmtools/datasets/binary.html#criteo

hcho3:
@newnovice01 This is a limitation of XGBoost: number of features cannot exceed `INT_MAX` (2147483647). For now, I will have the program throw exception when number of features is too big. We do want to remove the limitation eventually, however.

PR #3886, opened 2018-11-09
Document current limitation in number of features

hcho3:
See discussion at #3884. The `num_feature` variable uses 32-bit integer, which causes XGBoost to fail for datasets with more than 2147483647 features. For now, print a better error message.

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -674,11 +674,15 @@ class LearnerImpl : public Learner {
   inline void LazyInitModel() {
     if (this->ModelInitialized()) return;
     // estimate feature bound
+    // TODO(hcho3): Change num_feature to 64-bit integer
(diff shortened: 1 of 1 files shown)


hcho3:
@trivialfis FYI

trivialfis:
@hcho3 What's preventing using int64_t currently?

hcho3:
@trivialfis XGBoost saves the trained model as binary dump. So you can't easily add or change fields without breaking backward compatibility. As an aside, someone told me today that the use of binary dump is causing endian-compatibility issue as well (model created by little-endian machine doesn't load on big-endian machine).

khotilov:
@trivialfis Since dense structures are currently used in many places to index features, using such large indices would be consuming 2Gb times N of memory.

PR #3888, opened 2018-11-09
Remove unnecessary warning when 'gblinear' is selected

hcho3:
Rectify my mistake in #3849, according to discussion in https://github.com/dmlc/xgboost/pull/3849#issuecomment-436866139.


hcho3:
@trivialfis Good suggestion.

PR #3889, opened 2018-11-10
Minor refactor of split evaluation in gpu_hist

RAMitchell:
Move split evaluation logic into each device instead of having this logic "globally".  This is in preparation for a future PR where each device will propose splits based on its subset of the data, each device will vote and then the algorithm will globally calculate the optimal split for the selected feature.

Use span class in split evaluation for improved memory safety.

### DIFF
--- a/src/tree/updater_gpu_hist.cu
+++ b/src/tree/updater_gpu_hist.cu
@@ -149,7 +149,7 @@ template <int BLOCK_THREADS>
 __global__ void EvaluateSplitKernel(
     const GradientPairSumT* d_hist,  // histogram for gradients
     uint64_t n_features,
-    int* feature_set,  // Selected features
+    const int* feature_set,  // Selected features
     DeviceNodeStats node,
 
     const uint32_t* d_feature_segments,  // row_ptr form HistCutMatrix
@@ -451,11 +451,10 @@ struct DeviceShard {
   TrainParam param;
   bool prediction_cache_initialised;
 
-  // FIXME: Remove this
-  int64_t* tmp_pinned;  // Small amount of staging memory
-
   // Used to process nodes concurrently
-  std::vector<cudaStream_t> streams;
+  //std::vector<cudaStream_t> streams;
+
+  int64_t* tmp_pinned;  // Small amount of staging memory
 
   dh::CubMemory temp_memory;
 
@@ -473,7 +472,8 @@ struct DeviceShard {
     null_gidx_value(0),
     param(_param),
     prediction_cache_initialised(false),
-    tmp_pinned(nullptr) {}
+    tmp_pinned(nullptr)
+    {}
 
   /* Init row_ptrs and row_stride */
   void InitRowPtrs(const SparsePage& row_batch) {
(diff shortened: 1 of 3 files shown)


RAMitchell:
@hcho3 I am having issues with the Jenkins machines, do you know what's going on?

hcho3:
@RAMitchell Somehow auto scaling isn't working. For now, I manually provisioned workers.

RAMitchell:
Thanks added a description above. I will try to be a little more verbose in future.

src/tree/updater_gpu_hist.cu
```diff
@@ -550,16 +515,63 @@ struct DeviceShard {
...
+      best_split.Update(candidate, param);
+    }
+    return best_split;
+  }
+
+  /** \brief Builds both left and right hist with subtraction trick if possible.
+   */
+  void BuildHistWithSubtractionTrick(int nidx_parent, int nidx_left,
+                                     int nidx_right) {
+    auto smallest_nidx =
```
trivialfis: Nice, now we have a better naming.

src/tree/updater_gpu_hist.cu
```diff
@@ -550,16 +515,63 @@ struct DeviceShard {
     hist.Reset();
   }
 
+  DeviceSplitCandidate EvaluateSplit(int nidx,
+                                     const HostDeviceVector<int>& feature_set,
+                                     ValueConstraint value_constraint) {
+    dh::safe_cuda(cudaSetDevice(device_id_));
+    auto d_split_candidates = temp_memory.GetSpan<DeviceSplitCandidate>(feature_set.Size());
+    DeviceNodeStats node(node_sum_gradients[nidx], nidx, param);
+    feature_set.Reshard(GPUSet::Range(device_id_, 1));
```
trivialfis: I assume in the future there will be one `feature_set` for each device and the correspondence won't change, otherwise `Reshard` will get us into trouble.

src/tree/updater_gpu_hist.cu
```diff
@@ -43,14 +43,15 @@ using GradientPairSumT = GradientPairPrecise;
...
-__device__ GradientPairSumT ReduceFeature(const GradientPairSumT* begin,
-                                          const GradientPairSumT* end,
+__device__ GradientPairSumT ReduceFeature(common::Span<const GradientPairSumT> feature_histogram,
                                           TempStorageT* temp_storage) {
   __shared__ cub::Uninitialized<GradientPairSumT> uninitialized_sum;
   GradientPairSumT& shared_sum = uninitialized_sum.Alias();
 
   GradientPairSumT local_sum = GradientPairSumT();
   // For loop sums features into one block size
+  auto begin = feature_histogram.data();
```
trivialfis: Actually we might be able to use non-pointer iteration with this line:
https://github.com/dmlc/xgboost/blob/be0bb7dd908ba910f73ee33095df915df6a0c138/src/common/span.h#L144
change to use `<`,
and remove the check in this function:
https://github.com/dmlc/xgboost/blob/be0bb7dd908ba910f73ee33095df915df6a0c138/src/common/span.h#L177

The error report will be delayed to applying `* operator` and `-> operator`, but it will provide us some flexibility.  WDYT?

src/tree/updater_gpu_hist.cu
RAMitchell: Yes we will have to deal with this in the next PR. I am wondering if we can reshard it such that a copy exists on each device.

thvasilo:
Hello @RAMitchell, just curious will the future PR implement the PV-Tree described [here](https://arxiv.org/abs/1611.01276)?

RAMitchell:
@thvasilo Yes something like that, perhaps not exactly.

thvasilo:
@RAMitchell that's cool I'm interested in seeing how it turns out. The PV-Tree had issues with scaling and their accuracy experiments were limited unfortunately. Doing away completely with histogram communication still seems like a big challenge.

RAMitchell:
@thvasilo I would be interested if you have done some experiments to suggest pv-tree is not accurate enough. My intuition was that choosing slightly suboptimal features sometimes will not affect convergence much.

thvasilo:
@RAMitchell I'm mostly referring to the quality of evaluation in the paper. Only two closed-source datasets where used, and Figures 1-3 show deltas in NDCG and AUC in the order of 10^-2 to 10^-3. Whether that of difference is useful is up to the user/debatable.

For scalability see Section 5.2.1, Figure 2 is hard to interpret, a strong/weak scaling linear plot would have made more sense. In any case, just food for thought.

PR #3902, opened 2018-11-15
Fix Typo in learner.cc

theycallhimavi:


trivialfis:
@theycallhimavi Thanks.

PR #3910, opened 2018-11-16
[jvm-packages]support multiple validation datasets in Spark

CodingCat:
- [x] converge the current training/test split with multi validation datasets
- [x] add unit test for multiple validation set
- [x] add support for ranking training tasks
- [x] add unit test for ranking training
- [x] update tutorial
- [x] fix early stopping feature


CodingCat:
@yanboliang take a further look?

yanboliang:
Looks good to me overall, thanks.

CodingCat:
thanks @yanboliang

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/NonParamVariables.scala
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package ml.dmlc.xgboost4j.scala.spark.params
+
+import org.apache.spark.sql.DataFrame
+
+trait NonParamVariables {
+  protected var evalSetsMap: Map[String, DataFrame] = Map.empty
+
+  def setEvalSets(evalSets: Map[String, DataFrame]): Unit = {
```
linghaogu: I am not sure if it is the right place to comment, let me know if it is not correct.

I am a user of xgboost and I am using 0.82 for our production training. One thing I think we can improve is that the return value of this function is better to be this.type
Because as a scala user I will write code like:

    val xgb = new XGBoostClassifier(xgboostParam)
      .setFeaturesCol(MSDataSchema.FEATURE_VECTOR)
      .setLabelCol(MSDataSchema.RELEVANCE_LEVEL)
      .setEvalSets(Map("eval" -> testData))

xgb will be a None in this case

CodingCat:
Oh, it’s a typo or autocomplete by ide, feel free to file a PR addressing this!

linghaogu:
I will thank you!

PR #3920, opened 2018-11-18
[jvm-packages] update version to 0.82-SNAPSHOT

CodingCat:
will update again if we decide jump to 0.9x


CodingCat:
@hcho3 do we have any expectation to jump to 0.9x ?

PR #3928, opened 2018-11-21
[jvm-packages]Fix early stopping condition

CodingCat:


CodingCat:
@yanboliang ?

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -218,8 +218,15 @@ static boolean judgeIfTrainingOnTrack(
...
+    if (iter < earlyStoppingRounds - 1) {
+      return true;
+    }
     float[] criterion = metrics[metrics.length - 1];
-    for (int shift = 0; shift < Math.min(iter, earlyStoppingRounds) - 1; shift++) {
+    for (int shift = 0; shift < earlyStoppingRounds - 1; shift++) {
+      // the initial value of onTrack is false and if the metrics in any of `earlyStoppingRounds`
+      // iterations goes to the expected direction, we should consider these `earlyStoppingRounds`
+      // as `onTrack`
       onTrack |= maximizeEvaluationMetrics ?
```
mingyang: Does this `|=` mean that, if the metric is moving in the right direction *any* two consecutive steps within the `earlyStoppingRounds` from the current iteration, then this method will return `true`? 

This may not be what people normally expect from setting early stopping. For example, I'm getting a real training progress below with `earlyStoppingSteps` set to 20: the training should stop around iterations 120 since the maximum PR AUC was observed around iteration 100. But the current logic seems to look for *any* upward pieces within `earlyStoppingSteps` and keep training.
image

In the `python-package`(see [here](https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/callback.py#L246)), training stops if the current iteration is `earlyStoppingSteps` away from the best iteration. Should the spark version be consistent with the python implementation?

mingyang: I don't really write in Scala. So created this [issue](https://github.com/dmlc/xgboost/issues/4050) instead.

Issue #3931, opened 2018-11-22
[jvm-packages] The vector size of `rawPredictionCol` generated by XGBoostClassificationModel is incorrect


huafengw:
For binary classification, the vector in `rawPredictionCol` only contains one element while Spark's standard model will generate vectors having two elements. For example in GBTClassificationModel, the predictRaw function is like: 
```
  override protected def predictRaw(features: Vector): Vector = {
    val prediction: Double = margin(features)
    Vectors.dense(Array(-prediction, prediction))
  }
```

This problem will cause calling BinaryClassificationEvaluator failed when using `rawPredictionCol` to evaluate.

PR #3932, opened 2018-11-22
[jvm-packages] Fix vector size of 'rawPredictionCol' in XGBoostClassificationModel

huafengw:
Fix https://github.com/dmlc/xgboost/issues/3931

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
@@ -415,7 +415,9 @@ class XGBoostClassificationModel private[ml](
     var numColsOutput = 0
 
     val rawPredictionUDF = udf { rawPrediction: mutable.WrappedArray[Float] =>
-      Vectors.dense(rawPrediction.map(_.toDouble).toArray)
+      val raw = rawPrediction.map(_.toDouble).toArray
+      val rawPredictions = if (numClasses == 2) Array(-raw(0), raw(0)) else raw
+      Vectors.dense(rawPredictions)
     }
 
     val probabilityUDF = udf { probability: mutable.WrappedArray[Float] =>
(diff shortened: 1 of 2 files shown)


jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
```diff
@@ -415,7 +415,9 @@ class XGBoostClassificationModel private[ml](
     var numColsOutput = 0
 
     val rawPredictionUDF = udf { rawPrediction: mutable.WrappedArray[Float] =>
-      Vectors.dense(rawPrediction.map(_.toDouble).toArray)
+      val raw = rawPrediction.map(_.toDouble).toArray
+      val rawPredictions = if (numClasses == 2) Array(-raw(0), raw(0)) else raw
+      Vectors.dense(rawPredictions)
```
CodingCat: in XGBoost, rawPrediction means `margin` while in Spark, it means probability, so if you use BinaryClassifierEvaluator, you should pass in the value of `getProbabilityCol` instead of rawPredictionCol

huafengw: I think Spark also distinguishes `margin` and `probability`. For example in GBTClassificationModel the rawPrediction is `margin` and probability is the other thing. Please correct me if I'm wrong.

And if XGBoost has a different name convention, I think xgboost-spark should follow Spark's since it uses Spark's interface.

CodingCat: you are right, I cannot track back why I had the wrong but lasting impression that Spark doesn't have margin support... 

you need to fix the unit test though....

CodingCat: @yanboliang take a look at this?

Max1993Liu:
Hi, I'm running into the same issue with `binaryClassificationEvaluator` when using the following dependencies in my project: 
"ml.dmlc" % "xgboost4j-spark" % "0.81"
Since now it's been fixed, how can I use the fixed version in my project?

Thank you so much!

PR #3945, opened 2018-11-28
GPU performance logging/improvements

RAMitchell:
- Improved GPU performance logging
- Reintroduces the subtraction trick in the multi-GPU case
- Adds better performance logging
- Use only one "ExecuteShards" function.
- Use threads to launch NCCL AllReduce

### DIFF
--- a/src/tree/updater_gpu_hist.cu
+++ b/src/tree/updater_gpu_hist.cu
@@ -587,23 +587,6 @@ struct DeviceShard {
     return best_split;
   }
 
-  /** \brief Builds both left and right hist with subtraction trick if possible.
-   */
-  void BuildHistWithSubtractionTrick(int nidx_parent, int nidx_left,
-                                     int nidx_right) {
-    auto smallest_nidx =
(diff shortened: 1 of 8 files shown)


trivialfis:
From the look of it, it should be use `ExecuteIndexShard` only.  I will look more closely after Jenkins resumes.

src/tree/updater_gpu_hist.cu
```diff
@@ -1009,34 +992,66 @@ class GPUHistMaker : public TreeUpdater {
   void AllReduceHist(int nidx) {
     if (shards_.size() == 1) return;
 
-    reducer_.GroupStart();
-    for (auto& shard : shards_) {
+    monitor_.Start("AllReduce");
+    dh::ExecuteIndexShards(&shards_, [&](int idx, std::unique_ptr<DeviceShard>& shard) {
```
trivialfis: I see.  `ExecuteIndexShards` spawns multiple threads so the group calls are no longer needed.

PR #3965, opened 2018-12-05
Single precision histograms on GPU

RAMitchell:
This PR adds the ability to perform histogram summation on the GPU in single precision floating point. To do this I templated most of the code in gpu_hist. A side-effect will be modestly increased compile times and slightly larger binaries (with gcc, 153mb->155mb).

Another side-effect of this PR currently is that the GPUHistMaker will be reset every time updater->Init() is called. I'm not sure yet if this is exactly the behaviour we want.

I also simplified the gpu_hist test slightly to avoid combinatorial explosion of test cases, significantly reducing runtime.

### DIFF
--- a/src/tree/updater_gpu_hist.cu
+++ b/src/tree/updater_gpu_hist.cu
@@ -87,7 +87,7 @@ __device__ GradientSumT ReduceFeature(common::Span<const GradientSumT> feature_h
 
 /*! \brief Find the thread with best gain. */
 template <int BLOCK_THREADS, typename ReduceT, typename scan_t,
-          typename max_ReduceT, typename TempStorageT, typename GradientSumT>
+          typename MaxReduceT, typename TempStorageT, typename GradientSumT>
 __device__ void EvaluateFeature(
     int fidx,
     common::Span<const GradientSumT> node_histogram,
@@ -135,7 +135,7 @@ __device__ void EvaluateFeature(
     // Find thread with best gain
     cub::KeyValuePair<int, float> tuple(threadIdx.x, gain);
     cub::KeyValuePair<int, float> best =
-        max_ReduceT(temp_storage->max_reduce).Reduce(tuple, cub::ArgMax());
+        MaxReduceT(temp_storage->max_reduce).Reduce(tuple, cub::ArgMax());
 
     __shared__ cub::KeyValuePair<int, float> block_max;
     if (threadIdx.x == 0) {
@@ -361,7 +361,7 @@ __global__ void compress_bin_ellpack_k(
 }
 
 template <typename GradientSumT>
-__global__ void sharedMemHistKernel(size_t row_stride, const bst_uint* d_ridx,
+__global__ void SharedMemHistKernel(size_t row_stride, const bst_uint* d_ridx,
                                     common::CompressedIterator<uint32_t> d_gidx,
                                     int null_gidx_value,
                                     GradientSumT* d_node_hist,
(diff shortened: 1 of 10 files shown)


src/tree/updater_gpu_hist.cu
```diff
@@ -1308,6 +1357,37 @@ class GPUHistMaker : public TreeUpdater {
...
+
+  void Update(HostDeviceVector<GradientPair>* gpair, DMatrix* dmat,
+              const std::vector<RegTree*>& trees) override {
+    if (hist_maker_param_.single_precision_histogram) {
+      float_maker_->Update(gpair, dmat, trees);
+    } else {
+      double_maker_->Update(gpair, dmat, trees);
+    }
+  }
+
```
trivialfis: Is update prediction cache needed for this class?

src/tree/updater_gpu_hist.cu
```diff
@@ -1308,6 +1357,37 @@ class GPUHistMaker : public TreeUpdater {
   GPUDistribution dist_;
 };
 
+class GPUHistMaker : public TreeUpdater {
```
trivialfis: I think there are two choices here, one is adding a parameter `single_precision_histogram` and choose implementation at runtime, as done in this PR.  The cost is adding this wrapper class.

Another option is make it a separated updater, as done by something similar to:
``` c++
XGBOOST_REGISTER_TREE_UPDATER(GPUHistMaker<GradientPair>, "single_precision_grow_gpu_hist")

XGBOOST_REGISTER_TREE_UPDATER(GPUHistMaker<GradientPairPrecise>, "grow_gpu_hist")
```
This way those `override` modifiers can be preserved and with one less wrapper, also, a separated parameter class won't be necessary.  It makes sense to me since these two classes can indeed be considered as different updaters.

I don't mind your choice, but do you think it's worth considering?

src/tree/updater_gpu_hist.cu
```diff
@@ -69,10 +87,10 @@ __device__ GradientPairSumT ReduceFeature(common::Span<const GradientPairSumT> f
 
 /*! \brief Find the thread with best gain. */
 template <int BLOCK_THREADS, typename ReduceT, typename scan_t,
-          typename max_ReduceT, typename TempStorageT>
+          typename max_ReduceT, typename TempStorageT, typename GradientSumT>
```
trivialfis: The linter didn't catch it, but since we are here and there are some formatting changes in this PR, we can also rename the `max_ReduceT` to something like `MaxReduceT`.

src/tree/updater_gpu_hist.cu
```diff
@@ -30,7 +30,25 @@ namespace tree {
 
 DMLC_REGISTRY_FILE_TAG(updater_gpu_hist);
 
-using GradientPairSumT = GradientPairPrecise;
+// training parameters specific to this algorithm
+struct GPUHistMakerTrainParam
```
trivialfis: Please see comment for `GPUHistMaker`.

src/tree/updater_gpu_hist.cu
```diff
@@ -341,18 +360,17 @@ __global__ void compress_bin_ellpack_k(
   wr.AtomicWriteSymbol(buffer, bin, (irow + base_row) * row_stride + ifeature);
 }
 
-__global__ void sharedMemHistKernel(size_t row_stride,
-                                    const bst_uint* d_ridx,
+template <typename GradientSumT>
+__global__ void sharedMemHistKernel(size_t row_stride, const bst_uint* d_ridx,
```
trivialfis: So is this one, `SharedMemHistKernel`.

src/tree/updater_gpu_hist.cu
RAMitchell: I thought about this but I think it is more natural for a user to toggle a training parameter than to change the updater type. I don't want users to be changing the "updater" if possible, for example there may forget to add "prune" and then get a different behaviour from tree_method="gpu_hist".

src/tree/updater_gpu_hist.cu
RAMitchell: Yes, well spotted! This PR would have disabled prediction caching.

PR #3967, opened 2018-12-05
scikit-learn api section documentation correction

lyxthe:
The documentation is quite inconsistent in the scikit-learn api section since the fit paragraph tells that when early stopping rounds occurs, the last iteration is returned not the best one, but the predict paragraph tells that when the predict is called without ntree_limit specified, then ntree_limit is equals to best_ntree_limit.

Thus, when reading the fit part, one could think that it is needed to specify what is the best iter when calling the predict, but when reading the predict part, then the best iter is given by default, it is the last iter that you have to specify if needed.

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -631,11 +631,11 @@ class XGBClassifier(XGBModel, XGBClassifierBase):
         early_stopping_rounds : int, optional
             Activates early stopping. Validation error needs to decrease at
             least every <early_stopping_rounds> round(s) to continue training.
-            Requires at least one item in evals.  If there's more than one,
-            will use the last. Returns the model from the last iteration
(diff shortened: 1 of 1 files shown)


Edvard88:
What about this documentation 
https://xgboost.readthedocs.io/en/latest/python/python_intro.html?highlight=early%20stopping
will be it fixed?

@lyxthe 've written in  https://stackoverflow.com/questions/53483648/is-the-xgboost-documentation-wrong-early-stopping-rounds-and-best-and-last-it
if you fit with "best iteration" from early_stopping summary 

For example:
Stopping. Best iteration:
[109]	validation_0-auc:0.996667

fit with (109), you won't given the best score.
You should fit with "plus one" iteration, fit with 110 (because iterations starts from 0 ).
Then you'll get best score and it'll be best iteration.
Is it issue?

Issue #3913, opened 2018-11-16
xgboost.core.XGBoostError: b'[18:53:36] /root/repo/xgboost/src/common/host_device_vector.cu:355: Check failed: distribution_.IsEmpty() || distribution.IsEmpty()


pseudotensor:
```
Traceback (most recent call last):
  File "/home/jon/.pyenv/versions/3.6.4/lib/python3.6/site-packages/xgboost/sklearn.py", line 738, in fit
    callbacks=callbacks)
  File "/home/jon/.pyenv/versions/3.6.4/lib/python3.6/site-packages/xgboost/training.py", line 227, in train
    xgb_model=xgb_model, callbacks=callbacks)
  File "/home/jon/.pyenv/versions/3.6.4/lib/python3.6/site-packages/xgboost/training.py", line 74, in _train_internal
    bst.update(dtrain, i, obj)
  File "/home/jon/.pyenv/versions/3.6.4/lib/python3.6/site-packages/xgboost/core.py", line 1053, in update
    dtrain.handle))
  File "/home/jon/.pyenv/versions/3.6.4/lib/python3.6/site-packages/xgboost/core.py", line 170, in _check_call
    raise XGBoostError(_LIB.XGBGetLastError())
[...]

PR #3984, opened 2018-12-10
Unify used devices globally.

trivialfis:
Use global static `GPUSet` to represent devices and fail explicitly for changing devices.

Address #3913 .

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -175,7 +175,10 @@ def _check_call(ret):
         return value from API calls
     """
     if ret != 0:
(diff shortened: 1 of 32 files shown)


trivialfis:
Closing for now due to the fact that we have multiple sources of input dmatrix.  A singleton GPUSet might not be practical.  Sorry for the noise.

trivialfis:
I might have to find a Amazon instance to sort this out...

trivialfis:
@RAMitchell I don't think I can complete this PR is recent future and rebasing is gonna be painful.  Will close it for now.

PR #3990, opened 2018-12-12
[jvm-packages] upgrade spark to 2.4.0

rongou:
All the tests pass, and I've run the `SparkTraining` example successfully.

### DIFF
--- a/jvm-packages/pom.xml
+++ b/jvm-packages/pom.xml
@@ -34,7 +34,7 @@
         <maven.compiler.source>1.7</maven.compiler.source>
         <maven.compiler.target>1.7</maven.compiler.target>
         <flink.version>1.5.0</flink.version>
-        <spark.version>2.3.1</spark.version>
+        <spark.version>2.4.0</spark.version>
         <scala.version>2.11.12</scala.version>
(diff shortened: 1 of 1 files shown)


CodingCat:
@rongou thanks for the contribution !

my suggestion is 

1. we need more thorough integration test before we upgrade dependency

2. we need to test all versions we are supposed to support in Travis (there were some breaking changes when we upgrade from 2.2 to 2.3)

3. I am not going to take 2.4.0 but would prefer more 2.4.1 (2.4.0 is buggy especially in barrier scheduling)

rongou:
@CodingCat Thanks for the reply!

> 
> 1. we need more thorough integration test before we upgrade dependency

Are you talking about manual tests or automated tests? Is there an issue that tracks this?

> 2. we need to test all versions we are supposed to support in Travis (there were some breaking changes when we upgrade from 2.2 to 2.3)

Do you mean making `tests/travis/run_test.sh` to be able to switch spark versions?

> 3. I am not going to take 2.4.0 but would prefer more 2.4.1 (2.4.0 is buggy especially in barrier scheduling)

OK, this we just need to wait, right?

CodingCat:
I will close it for now

PR #3995, opened 2018-12-12
Combine TreeModel and RegTree

RAMitchell:
This PR contains a subset of less controversial changes from #3983 with more concrete reasoning for the changes. The high level goal is to reduce the complexity and size of the class and improve usability.

The changes are:

**Merge TreeModel and RegTree**
RegTree is a subclass of TreeModel, which has no other subclasses. It looks as if this was designed many years ago with the possibility of having other types of trees apart from regression trees. Todays xgboost only supports regression trees and it is difficult to see what other kind of tree we might support, whether this interface would even allow it and how this would fit into the current code base.

I believe this classifies as speculative generality (https://refactoring.guru/smells/speculative-generality). Removal will make the code base easier to understand, more flexible and easier to maintain. In the event that we do want to add additional types of tree in future we can reintroduce the base class, but I think it is better to do this with specific proposals in mind as opposed to keeping the current design that may or may not meet these needs.

**Remove TreeModel::AddRightChild method**
This method is unused and contributes to the bloat of the class.

**Remove second argument of TreeModel::GetDepth(int nid, bool pass_rchild = false)**
Argument is not used anywhere and has no clear purpose

**Move TreeModel::InitModel() to constructor**
There is no need for a delayed Init method here. It is safer to perform initialisation in the constructor so the object is always prepared for use. The previous design is less safe as if the user forgets to call InitModel it can result in a segfault.

**Remove TreeModel::Predict method**
This method is unused and contributes to the bloat of the class. If this is needed in future it can be trivially implemented.

**Move large functions to .cc file**
These functions are not frequently called (e.g. in a hot loop) and do not need to be inlined. This also improves compilation time.

### DIFF
--- a/include/xgboost/tree_model.h
+++ b/include/xgboost/tree_model.h
@@ -22,6 +22,8 @@
 
 namespace xgboost {
 
+struct PathElement;  // forward declaration
+
 /*! \brief meta parameters of the tree */
 struct TreeParam : public dmlc::Parameter<TreeParam> {
   /*! \brief number of start root */
@@ -62,18 +64,26 @@ struct TreeParam : public dmlc::Parameter<TreeParam> {
   }
 };
 
+/*! \brief node statistics used in regression tree */
+struct RTreeNodeStat {
+  /*! \brief loss change caused by current split */
+  bst_float loss_chg;
+  /*! \brief sum of hessian values, used to measure coverage of data */
+  bst_float sum_hess;
+  /*! \brief weight of current node */
+  bst_float base_weight;
+  /*! \brief number of child that is leaf node known up to now */
+  int leaf_child_cnt;
+};
+
 /*!
- * \brief template class of TreeModel
- * \tparam TSplitCond data type to indicate split condition
- * \tparam TNodeStat auxiliary statistics of node to help tree building
+ * \brief define regression tree to be the most common tree model.
+ *  This is the data structure used in xgboost's major tree models.
  */
-template<typename TSplitCond, typename TNodeStat>
-class TreeModel {
+class RegTree {
  public:
-  /*! \brief data type to indicate split condition */
-  using NodeStat = TNodeStat;
   /*! \brief auxiliary statistics of node to help tree building */
-  using SplitCond = TSplitCond;
+  using SplitCondT = bst_float;
   /*! \brief tree node */
   class Node {
    public:
@@ -111,7 +121,7 @@ class TreeModel {
       return (this->info_).leaf_value;
     }
     /*! \return get split condition of the node */
-    inline TSplitCond SplitCond() const {
+    inline SplitCondT SplitCond() const {
       return (this->info_).split_cond;
     }
     /*! \brief get parent of the node */
@@ -131,6 +141,13 @@ class TreeModel {
       return parent_ == -1;
     }
     /*!
+     * \brief set the left child
+     * \param nid node id to right child
+     */
+    inline void SetLeftChild(int nid) {
+      this->cleft_ = nid;
+    }
+    /*!
      * \brief set the right child
      * \param nid node id to right child
      */
@@ -143,7 +160,7 @@ class TreeModel {
      * \param split_cond  split condition
      * \param default_left the default direction when feature is unknown
      */
-    inline void SetSplit(unsigned split_index, TSplitCond split_cond,
+    inline void SetSplit(unsigned split_index, SplitCondT split_cond,
                           bool default_left = false) {
       if (default_left) split_index |= (1U << 31);
       this->sindex_ = split_index;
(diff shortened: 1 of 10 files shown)


trivialfis:
Looked briefly, I like the moving to cc file part.  One minor issue is can you take care the indent of function parameters?  Next time I step on the code my editor might try to correct these indent which brings some headaches.

include/xgboost/tree_model.h
```diff
@@ -517,15 +446,15 @@ class RegTree: public TreeModel<bst_float, RTreeNodeStat> {
...
+  void CalculateContributionsApprox(const RegTree::FVec& feat, unsigned root_id,
                                            bst_float *out_contribs) const;
   /*!
    * \brief get next position of the tree given current pid
    * \param pid Current node id.
    * \param fvalue feature value if not missing.
    * \param is_unknown Whether current required feature is missing.
    */
-  inline int GetNext(int pid, bst_float fvalue, bool is_unknown) const;
+  int GetNext(int pid, bst_float fvalue, bool is_unknown) const;
```
tqchen: This function need to be inlined

RAMitchell:
@trivialfis I'm not sure what you mean can you show an example please.

src/tree/tree_model.cc
```diff
@@ -169,4 +169,236 @@ std::string RegTree::DumpModel(const FeatureMap& fmap,
...
+    result  = this->FillNodeMeanValue(node.LeftChild()) * this->Stat(node.LeftChild()).sum_hess;
+    result += this->FillNodeMeanValue(node.RightChild()) * this->Stat(node.RightChild()).sum_hess;
+    result /= this->Stat(nid).sum_hess;
+  }
+  this->node_mean_values_[nid] = result;
+  return result;
+}
+
+void RegTree::CalculateContributionsApprox(const RegTree::FVec& feat, unsigned root_id,
+                                                  bst_float *out_contribs) const {
```
trivialfis: This line is incorrectly indented.  I don't know how it looks like on your IDE, but please view it on Github.

src/tree/tree_model.cc
```diff
@@ -169,4 +169,236 @@ std::string RegTree::DumpModel(const FeatureMap& fmap,
...
+  bst_float one_fraction;
+  bst_float pweight;
+  PathElement() = default;
+  PathElement(int i, bst_float z, bst_float o, bst_float w) :
+    feature_index(i), zero_fraction(z), one_fraction(o), pweight(w) {}
+};
+
+// extend our decision path with a fraction of one and zero extensions
+void ExtendPath(PathElement *unique_path, unsigned unique_depth,
+                       bst_float zero_fraction, bst_float one_fraction, int feature_index) {
```
trivialfis: So is this line.

src/tree/tree_model.cc
```diff
@@ -169,4 +169,236 @@ std::string RegTree::DumpModel(const FeatureMap& fmap,
...
+    unique_path[i].feature_index = unique_path[i+1].feature_index;
+    unique_path[i].zero_fraction = unique_path[i+1].zero_fraction;
+    unique_path[i].one_fraction = unique_path[i+1].one_fraction;
+  }
+}
+
+// determine what the total permuation weight would be if
+// we unwound a previous extension in the decision path
+bst_float UnwoundPathSum(const PathElement *unique_path, unsigned unique_depth,
+                                unsigned path_index) {
```
trivialfis: And this.

src/tree/tree_model.cc
```diff
@@ -169,4 +169,236 @@ std::string RegTree::DumpModel(const FeatureMap& fmap,
...
+      total += (unique_path[i].pweight / zero_fraction) / ((unique_depth - i)
+               / static_cast<bst_float>(unique_depth + 1));
+    }
+  }
+  return total;
+}
+
+// recursive computation of SHAP values for a decision tree
+void RegTree::TreeShap(const RegTree::FVec& feat, bst_float *phi,
+                              unsigned node_index, unsigned unique_depth,
```
trivialfis: And this, any many others in `tree_model.h`.

include/xgboost/tree_model.h
```diff
@@ -130,6 +140,13 @@ class TreeModel {
     inline bool IsRoot() const {
       return parent_ == -1;
     }
+    /*!
+     * \brief set the left child
+     * \param nid node id to right child
+     */
+    inline void SetLeftChild(int nid) {
```
trivialfis: Remove `inline` keyword.

include/xgboost/tree_model.h
```diff
@@ -143,7 +160,7 @@ class TreeModel {
      * \param split_cond  split condition
      * \param default_left the default direction when feature is unknown
      */
-    inline void SetSplit(unsigned split_index, TSplitCond split_cond,
+    inline void SetSplit(unsigned split_index, SplitCondT split_cond,
```
trivialfis: The `inline` is not useful here.  Any methods defined inside class is automatically inlined.

include/xgboost/tree_model.h
```diff
@@ -321,38 +299,28 @@ class TreeModel {
...
   /*!
    * \brief get current depth
    * \param nid node id
-   * \param pass_rchild whether right child is not counted in depth
    */
-  inline int GetDepth(int nid, bool pass_rchild = false) const {
+  inline int GetDepth(int nid) const {
     int depth = 0;
     while (!nodes_[nid].IsRoot()) {
-      if (!pass_rchild || nodes_[nid].IsLeftChild()) ++depth;
```
trivialfis: I don't quite understand the logic of this, so it's not used anymore?

RAMitchell: pass_rchild is never used in the code, therefore it is always false and this condition always evaluates to true. So I have removed both the function argument pass_rchild and the if statement so the behaviour is the same but the function is simpler.

PR #4001, opened 2018-12-15
Fix logger outputs.

trivialfis:
Introduced in #3982.  Logging switch statement is not breaking properly.  Sorry for the noise, my bad.


trivialfis:
Trying to trigger a re-build on Travis.  Recently Travis seems to be cranky.

PR #4002, opened 2018-12-16
Fix and optimize logger

trivialfis:
Re-opening, due to hanged Travis.

- Fixed incorrect switch statement.
- Use macro to prevent constructing stream when not needed, should have minor performance improvement.
- Add more tests.


trivialfis:
Since this is a small PR and contains fix, I'm going to merge it tomorrow if no objection is raised. :)

PR #4013, opened 2018-12-20
Merge duplicated linear updater parameters.

trivialfis:


RAMitchell:
Additionally to my review comments above, if we ever manage to implement a system that detects if parameters are unused, it would be better for the shared parameter structure to not contain parameters unless they are used by all updaters. Otherwise we will detect that the parameter has been initialised in the parameter struct but the algorithm never uses it.

trivialfis:
@RAMitchell 
> have a secondary parameters structure in each algorithm with more specific parameters

Agree.  I splited `top_k` out of `LinearTrainParam` for coordinate descent.  For other part of `XGBoost`, simply grepping the source code should be able to tell the uniqueness of parameters.  `n_gpus` and `gpu_id` should be removed in once #3984 is sorted out.

>  if we ever manage to implement a system that detects if parameters are unused

Yes, we will have that eventually.  My plan is we put the `InitAllowUnknown` from `dmlc::Parameter` into good use.   This method returns all parameters that are not known to that specific `dmlc::Parameter`.  We then pass it to other components.  One extra effort is we need to make sure future PR doesn't introduce duplicated parameters.  A simple test by parsing XGBoost's code based looking for `DMLC_DECLARE_FIELD` and `DMLC_DECLARE_ALIAS` should be able to get the job done.  The only blocker is the binary format, which copies some parameters like `num_class` from other places.  But I'm optimistic about that.

RAMitchell:
LGTM. It might be nice if we could use one parameter struct as a base class for another so then we end up with only one struct in the updater. I briefly tried this but it seems unlikely this would work without significant modification of the DMLC parameter manager. So let's just stick with having to two structs per updater.

trivialfis:
@RAMitchell Interesting.  I tried that too.  `dmlc::Parameter` lets parent class inherit child class, which makes thing a little bit tricky.

PR #4015, opened 2018-12-21
 Require leaf statistics when expanding tree

RAMitchell:
This PR extends the RegTree::ExpandNode() function to require summary statistics for the node being expanded as well as leaf weights for the left and right child. Calculating the leaf weights in advance for the child nodes required cacheing the left and right gradient sums from split calculation, so I have extended the SplitEntry class with this information.

There are several impacts to this change:
- Expanding a tree node is now a safe operation. After you call ExpandNode() the tree is complete and may be used immediately with all the required statistics and leaf nodes configured.
- As a consequence of the above, all code in tree updater algorithms that manually accesses RegTree internals such as tree nodes, statistics or leaf nodes becomes unnecessary (e.g. [this](https://github.com/dmlc/xgboost/blob/f75a21af25d1752648d78d8efae33b6cb1a7fac9/src/tree/updater_colmaker.cc#L124)). I have not remove these in this PR as it would make it too complicated, but I have verified that it is possible to remove them safely.
- Because the gradient sums in the newly created node children are cached, we no longer need to recalculate this. This means we can remove loops such as [this](https://github.com/dmlc/xgboost/blob/f75a21af25d1752648d78d8efae33b6cb1a7fac9/src/tree/updater_colmaker.cc#L214) and [this](https://github.com/dmlc/xgboost/blob/f75a21af25d1752648d78d8efae33b6cb1a7fac9/src/tree/updater_quantile_hist.cc#L631) for free. Again I will leave this for a future PR.

So in summary this will collapse a lot of the complexity around configuring a decision tree and will also improve performance. This is the way I have been doing things already in the GPU algorithms so it also helps me unify the code base.

### DIFF
--- a/src/tree/updater_gpu_hist.cu
+++ b/src/tree/updater_gpu_hist.cu
@@ -1182,42 +1182,35 @@ class GPUHistMakerSpecialised{
   }
 
   void ApplySplit(const ExpandEntry& candidate, RegTree* p_tree) {
-    // Add new leaves
     RegTree& tree = *p_tree;
-    tree.ExpandNode(candidate.nid, candidate.split.findex, candidate.split.fvalue,
-                    candidate.split.dir == kLeftDir);
-    auto& parent = tree[candidate.nid];
-    tree.Stat(candidate.nid).loss_chg = candidate.split.loss_chg;
-
-    // Set up child constraints
-    node_value_constraints_.resize(tree.GetNodes().size());
     GradStats left_stats(param_);
     left_stats.Add(candidate.split.left_sum);
     GradStats right_stats(param_);
     right_stats.Add(candidate.split.right_sum);
-    node_value_constraints_[candidate.nid].SetChild(
-        param_, parent.SplitIndex(), left_stats, right_stats,
-        &node_value_constraints_[parent.LeftChild()],
-        &node_value_constraints_[parent.RightChild()]);
-
-    // Configure left child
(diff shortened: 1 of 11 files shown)


RAMitchell:
Ping @tqchen @hcho3

src/tree/updater_gpu_hist.cu
```diff
@@ -1182,42 +1182,35 @@ class GPUHistMakerSpecialised{
...
-        param_, parent.SplitIndex(), left_stats, right_stats,
-        &node_value_constraints_[parent.LeftChild()],
-        &node_value_constraints_[parent.RightChild()]);
-
-    // Configure left child
+    GradStats parent_sum(param_);
+    parent_sum.Add(left_stats);
+    parent_sum.Add(right_stats);
+    node_value_constraints_.resize(tree.GetNodes().size());
+    auto base_weight = node_value_constraints_[candidate.nid].CalcWeight(param_, parent_sum);
```
hcho3: Do we need to re-compute `base_weight` here? Is it possible to re-use the leaf weight of `nid` node prior to expanding?

RAMitchell: The leaf weight is multiplied by learning rate, we would have to undo this with potential numerical issues.

hcho3: Ah I see. Then let's proceed as it is. Go ahead and merge?

PR #4020, opened 2018-12-25
Use Span in GPU exact updater.

trivialfis:
* Add a small test.
* Use p_impl for gpu_exact so the test doesn't need to include a cu file.


trivialfis:
> I don't like the usage of pimpl just for the test. This is non standard for the rest of the updater algos and I'm not sure there is a strong enough reason for it. Rest looks good.

Done.

PR #4028, opened 2018-12-28
[jvm-packages] make clarity of xgboost4j-example pom.xml

KyleLi1985:
Add needed dependency to make xgboost4j-example's pom.xml clarity


CodingCat:
it doesn't make sense, as spark has been the transient dependency . Why we need to list transient dependency here?

PR #4029, opened 2018-12-29
Use Span in gpu coordinate.

trivialfis:
* Use Span in device code.
* Fix shard size calculation.
  - Use lower_bound instead of upper_bound.
* Check empty devices.


trivialfis:
That's a bad news.  Another MGPU test failed I can't reproduce. :(

trivialfis:
@RAMitchell I would like to move these empty shards checking into `ExecuteIndexedShards`, but there are other complications like `gpu_predictor` uses overlapped data distribution.  So I'm leaving it for #3984 .

PR #4046, opened 2019-01-06
[jvm-packages] fix safe execution

CodingCat:
looks like we are capturing the wrong events in SparkParallelismTracker

fix it @Yunni


Yunni:
Good catch. Thanks!

PR #4049, opened 2019-01-07
[jvm-packages] For Performance consideration and Alignment input parameter of repartition function

KyleLi1985:


CodingCat:
merged thanks

PR #4066, opened 2019-01-19
Prevent training without setting up caches.

### LINKED ISSUE #4056: [Blocking] python kernel failed when call Booster().predict
Sometimes, the python kernel will be failed when I call Booster().predict method.
Here is an example:
```python
import xgboost as xgb
import numpy as np

d = xgb.DMatrix(np.random.rand(1000, 5))
tree = xgb.Booster({'max_depth': 3, 'eta': 0.1, 'base_score': 0.0})

tree.boost(d, np.random.rand(1000), np.ones(1000))
print('start...')
preds = tree.predict(d)
print('end...')
```
Then 
```
[19:35:05] src/tree/updater_prune.cc:74: tree pruning end, 1 roots, 14 extra nodes, 0 pruned nodes, max_depth=3
start...

Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)
```
However, when I modify the following line
```python
tree.boost(d, np.random.rand(1000), np.ones(1000))
```
into
```python
tree.boost(d, np.ones(1000), np.ones(1000))
```
Then every thing is ok.
```
start...
[19:49:16] src/tree/updater_prune.cc:74: tree pruning end, 1 roots, 0 extra nodes, 0 pruned nodes, max_depth=0
end...
```
How to fix this problem?

trivialfis:
* Add warning for internal functions.
* Check number of features.
close #4056 .

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -1041,8 +1041,8 @@ class Booster(object):
             _check_call(_LIB.XGBoosterSetParam(self.handle, c_str(key), c_str(str(val))))
 
     def update(self, dtrain, iteration, fobj=None):
-        """
-        Update for one iteration, with objective function calculated internally.
+        """Update for one iteration, with objective function calculated
+        internally.  This function should not be called directly by users.
 
         Parameters
         ----------
@@ -1052,6 +1052,7 @@ class Booster(object):
             Current iteration number.
         fobj : function
             Customized objective function.
+
         """
         if not isinstance(dtrain, DMatrix):
             raise TypeError('invalid training matrix: {}'.format(type(dtrain).__name__))
@@ -1066,8 +1067,9 @@ class Booster(object):
             self.boost(dtrain, grad, hess)
 
     def boost(self, dtrain, grad, hess):
-        """
-        Boost the booster for one iteration, with customized gradient statistics.
+        """Boost the booster for one iteration, with customized gradient
+        statistics.  Like update, This function should not be called
+        directly by users.
 
         Parameters
         ----------
@@ -1077,6 +1079,7 @@ class Booster(object):
             The first order of gradient.
         hess : list
             The second order of gradient.
+
         """
         if len(grad) != len(hess):
             raise ValueError('grad / hess length mismatch: {} / {}'.format(len(grad), len(hess)))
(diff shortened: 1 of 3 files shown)


trivialfis:
Calling `Booster.update` with first initializing `Booster` with cache for dtrain results in `num_feature` == 0.  I can raise an error in Python, but checking it in C++ seems more reliable.  Added documentation effectively marked `Booster.update` and `Booster.boost` as internal methods.

@RAMitchell , @hcho3  Could you take a look see if there is better decision?

trivialfis:
> Can we really say don't use something that is documented as a part of the public api? Maybe it is better to say 'power users only' or something.

Actually I don't like my solution.  Let's see if I can make new datasets part of the caches safely.

trivialfis:
@RAMitchell Turns out I can't make the incoming dataset for training to become part of the caches for the following reason:

In `c_api`, `DMatrix` handler is a `std::shared_ptr`,  while `Learner.UpdaterOneIter` accepts raw pointer.  Every time `c_api::XGBoostUpdaterOneIter` calls `Learner::UpdaterOneIter`, it first calls `shared_ptr::get()` to pass the raw pointer.  So there's no way to let the caches from `Learner` obtain ownership of this `DMatrix` without changing `Learner`'s interface, hence can not make it part of the caches (which is a vector of shared pointer).

And no, it's not "power user only".  I'm a power user ( I think :) ), I don't know how to make it work other than making a copy of the abstracted APIs.

trivialfis:
I will go ahead and merge this if no objections. @RAMitchell @hcho3

hcho3:
Go ahead.

python-package/xgboost/core.py
```diff
@@ -1066,8 +1067,9 @@ def update(self, dtrain, iteration, fobj=None):
             self.boost(dtrain, grad, hess)
 
     def boost(self, dtrain, grad, hess):
-        """
-        Boost the booster for one iteration, with customized gradient statistics.
+        """Boost the booster for one iteration, with customized gradient
+        statistics.  Like update, This function should not be called
```
thvasilo: Small typo
```suggestion
        statistics.  Like update, this function should not be called
```

Also we might want to add a [cross-ref](https://stackoverflow.com/a/21297186/209882) here to clearly show which update we're talking about.

trivialfis:
@thvasilo Thanks for the pointer.

Issue #4080, opened 2019-01-25
[jvm-packages] Train XGBoost in Spark, got errors, but failed to quit


FredYao:
Datasize was really small (about several MBs). 

Spark version: 2.3.0
yarn cluster
Num-executors: 4
Executor-cores 2
Executor-memory 4g
Executor-memory-overhead 2g

XGBoost param:
"eta" -> 0.1f,
"max_depth" -> 5,
"objective" -> "binary:logistic",
"num_round" -> 200,
"tracker_conf" -> TrackerConf(30000L, "scala"),
"num_workers" -> 3,
"nthread" -> 1

Scala code:

`
def train(...): PipelineModel = {
    val data = readData(...)
    val pipeline = new Pipeline().setStages(...)  // the last stage is an XGBoostClassifier
    val model = pipeline.fit(data)
    val result = model.transform(data)
    model
}

def main(String[] args): Unit = {
    val spark = //create a sparksession
    ...
    val a = Try(train(...))
    a match {
       case Success(model) => println("Successfully trained")
       case Failure(ex) => println("Failed to train " + ex)
    }
    
   spark.stop()
`

The training failed at 'foreachPartition' at XGBoost.scala:287. See below screens
![111](https://user-images.githubusercontent.com/4297204/51719341-f6d74200-1ffd-11e9-9fa8-1999892d7482.png)
![222](https://user-images.githubusercontent.com/4297204/51719343-fa6ac900-1ffd-11e9-873b-84227733af36.png)

The UI showed no active executors:
![333](https://user-images.githubusercontent.com/4297204/51719377-18382e00-1ffe-11e9-8388-cf21cd987f40.png)

The Spark application was hanging there and never quit. 

I knew that there were something wrong when training the XGBoost model, however, what I had expected was that the application would stop and quit normally. Does anybody know the causes? Thanks!

hcho3:
@FredYao Maybe the data was so small that one of the executors didn't get any work assigned?
@CodingCat Does distributed XGBoost handle edge cases where a worker doesn't get any work (because dataset is so small, and you have unbalanced trees where most of the data falls in one node)?

CodingCat:
@hcho3, in that case the application will fail

@FredYao , can you try from master branch? https://github.com/dmlc/xgboost/pull/4046 should fix the issue

CodingCat:
I have tested, #4046 resolved the issue

CodingCat:
[...]

PR #4108, opened 2019-02-06
[jvm-packages] better fix for shutdown applications 

CodingCat:
closes https://github.com/dmlc/xgboost/issues/4080


CodingCat:
tested in prod environment, now it shutdown sparkcontext directly instead of tweaking the types of exception

PR #4095, opened 2019-01-31
[REVIEW] Enable Multi-Node Multi-GPU functionality

mt-jones:
To enable multi-node multi-GPU functionality, this PR adjusts `dh::AllReducer::Init` to query all `rabit::GetRank()`s for GPU device information (additional node stats). This permits NCCL to construct a unique ID for a communicator clique across many rabit workers. Exposing this unique ID, also permits custom process schema for each GPU in the comm clique (one process per GPU, or one process for multiple GPUs, etc.).

In Dask, this would permit a user to construct a collection of XGBoost processes over an inhomogeneous device cluster. It would also permit resource management systems to assign a rank for each GPU in the cluster.

PR #4113, opened 2019-02-07
Use nccl group calls to prevent from dead lock.

### LINKED ISSUE #4107: xgboost for GPU hangs on AllReduce and CPU Barrier
Could be nccl issue, but I'd like to get a confirmation.
https://github.com/NVIDIA/nccl/issues/176

trivialfis:
* launch all reduce sequentially.
* Fix gpu_exact test memory leak.

close #4107


trivialfis:
#4095 should fix this in a better way.  But before that PR is fully baked, here is a quick fix.

Issue #4098, opened 2019-02-01
src/common/./quantile.h:396: Check failed: nbig < n quantile: too many large chunk  when group weights used


kretes:
When training python-xgboost on CPU with pairwise objective and calling set_weights on DMatrix - for some dataset size it blows with following exception:

```
[16:22:40] Tree method is selected to be 'hist', which uses a single updater grow_fast_histmaker.
[16:22:41] src/common/./quantile.h:392:  check quantile stats, nbig=30756, n=20479
[16:22:41] src/common/./quantile.h:393:  srcsize=40907, maxsize=20481, range=0.001, chunk=9.7661e-08
...
src/common/./quantile.h:396: Check failed: nbig < n quantile: too many large chunk

Stack trace returned 7 entries:
[bt] (0) /home/kraken/.conda/envs/kraken_full/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(dmlc::StackTrace()+0x42) [0x7fa3d4a8acf2]
[bt] (1) /home/kraken/.conda/envs/kraken_full/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x18) [0x7fa3d4a8b2f8]
[bt] (2) /home/kraken/.conda/envs/kraken_full/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(xgboost::common::WXQSummary<float, float>::SetPrune(xgboost::common::WQSummary<float, float> const&, unsigned long)+0xab2) [0x7fa3d4abb312]
[bt] (3) /home/kraken/.conda/envs/kraken_full/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(+0xb5568) [0x7fa3d4ab6568]
[bt] (4) /home/kraken/.conda/envs/kraken_full/lib/python3.6/site-packages/xgboost/./lib/../../../../libgomp.so.1(+0x10e25) [0x7fa3db09ae25]
[bt] (5) /lib/x86_64-linux-gnu/libpthread.so.0(+0x8184) [0x7fa3e4b7e184]
[bt] (6) /lib/x86_64-linux-gnu/libc.so.6(clone+0x6d) [0x7fa3e48ab03d]
```

This happens for a dataset size of 16M observations, 44 features and 300K groups.
For 25% less observations - the training goes on.
For 25% more observation - the training fails with segfault without any stacktrace 
When we are not passing any weights for groups - the training goes on.

[...]

Issue #4093, opened 2019-01-31
[Roadmap] external memory support for tree_method='hist'


CodingCat:
in the [doc](https://xgboost.readthedocs.io/en/latest/parameter.html) we mentioned that Distributed and external memory version only support tree_method=approx

but is it valid now? what's the problem with external memory + hist? I tested but didn't see any issue, did I miss anything?

CodingCat:
@hcho3

CodingCat:
hmmm....looks like the training converged too early....but didn't get the idea why it is related to external memory or the row.page has some problem....

trivialfis:
There's open issue related to external memory #4037.  I'm still struggling to understand it...

CodingCat:
@hcho3 , yes, our data size makes external memory a must-have......but I am a bit confused which part makes hist does not work with external memory....

@trivialfis what I observed is that when I setup external memory with hist tree_method, the metrics (train/test) doesn't change after 5 - 6 iterations, though the training still moves forward.....

CodingCat:
I have started looking into why external memory version of hist is messed up with accuracy

hcho3:
@CodingCat Is this issue blocking?

CodingCat:
no, it just means you need more memory to use distributed hist,

rongou:
@CodingCat Is this still an issue? As part of my work on gpu external memory support, I tested `hist` on the Higgs dataset and got identical AUC metrics. The external memory version is actually slightly faster.

|Mode|Time(seconds)|AUC|
|---|---:|---:|
|`hist`|1309.64|0.8393|
|`hist` external memory|1228.53|0.8393|

CodingCat:
I didn’t test it for a while...last time what I saw is the metrics stays
the same since the first few iteration

Maybe it has been fixed in some changes regarding dmatrix, etc.

PeterPanonGit:
@rongou In my experiments external memory seems to run much slower than the in-memory counterpart regardless of the tree_method I use. However in your benchmark it seems that hist and hist external memory speed is almost the same. Could you post a script that we can use to re-produce your benchmark?

rongou:
@PeterPanonGit I don't have the script anymore, but I wasn't doing anything special, just the standard approach for enabling external memory. Since in external memory mode a lot of data is written out to disk and read back, the speed of your disk might be a factor. Are you on a local desktop with a late model SSD? In the cloud environment I/O might be more constrained.

[...]

Issue #4039, opened 2019-01-03
[python-package] does objective = 'rank:ndcg' works fine in XGBRanker?


myotheone:
XGBRanker docstring says:
objective : string
            Specify the learning task and the corresponding learning objective.
            **Only "rank:pairwise" is supported currently.**
So does rank:ndcg works fine in XGBRanker?

myotheone:
docstring need to be updated

hcho3:
@myotheone @trivialfis Agreed that the docstring should be updated. I'll open a PR to fix it.

PR #4115, opened 2019-02-08
Make `HistCutMatrix::Init' be aware of groups.

trivialfis:
* Add checks for group size.

Address segfault in  #4098,  error in quantile is not resolved.

### DIFF
--- a/src/common/hist_util.cc
+++ b/src/common/hist_util.cc
@@ -24,7 +24,12 @@
 namespace xgboost {
 namespace common {
 
+HistCutMatrix::HistCutMatrix() {
+  monitor_.Init("HistCutMatrix");
+}
+
 void HistCutMatrix::Init(DMatrix* p_fmat, uint32_t max_num_bins) {
+  monitor_.Start("Init");
   const MetaInfo& info = p_fmat->Info();
 
   // safe factor for better accuracy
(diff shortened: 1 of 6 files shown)


hcho3:
@trivialfis Yes, to clarify, weights in ranking objects will have to be per-group basis. Does this PR make this assumption? For the quantile sketch, we can replicate the group weight to cover all data points in each group.

trivialfis:
@hcho3 This PR accepts both **per-group** and **per-instance** until the last commit `see failed java tests`.  I don't know how long does it take to get Travis running.

There was a test in jvm-package scala that inputs different number of weights and groups:

https://travis-ci.org/dmlc/xgboost/jobs/490537869#L1636

Please ignore the incorrect check, it's fixed in later commits.  But the showed error message seems to support **per-instance** weighting.

trivialfis:
@hcho3 Ha, there it's:
https://travis-ci.org/dmlc/xgboost/jobs/490661939#L1705

So it's a buggy test and I should remove/fix it?

hcho3:
@trivialfis Yes, and we'll need to be clear about the use of per-group weights. It really doesn't make sense to give different weights to data rows in the same query session, since only the relative ordering of the data rows are important (hence learning-to-rank).

trivialfis:
@hcho3 Thanks for the explanation.  That goes in my reading list.

hcho3:
@trivialfis A good place to start is to learn what the ranking metric (e.g. NDCG) does. This reading helped me: http://dalelane.co.uk/blog/?p=3403

trivialfis:
@hcho3 Thanks for the link!  That's useful.

trivialfis:
@hcho3 This should be ready now if the tests don't go nuts.  Currently I can't write the documentation for ranking #4039 .

hcho3:
@trivialfis Okay, I'll review.

src/common/hist_util.cc
```diff
@@ -35,30 +35,47 @@ void HistCutMatrix::Init(DMatrix* p_fmat, uint32_t max_num_bins) {
...
       CHECK_EQ(nthread, omp_get_num_threads());
       auto tid = static_cast<unsigned>(omp_get_thread_num());
       unsigned begin = std::min(nstep * tid, ncol);
       unsigned end = std::min(nstep * (tid + 1), ncol);
+
+      // Data groups, used in ranking.
+      std::vector<bst_uint> const& groups = info.group_ptr_;
+      size_t const num_groups = groups.size() == 0 ? 0 : groups.size() - 1;
+      size_t group_ind =  // index into groups
+          tid * static_cast<unsigned>((num_groups + nthread - 1) / nthread);
```
hcho3: I don't think this will work when `p_fmat->GetRowBatches()` returns multiple batches.

hcho3:
@trivialfis Can you look at my latest revision and see if it's good?

hcho3:
Oops, I think it's wrong. Let me try again

trivialfis:
@hcho3 Do we need to calculate different `group_ind` for each thread since they might be processing data belonging to different groups?  And, it might be worthy to use some sort of binary search to find the index instead of a linear search.

trivialfis:
I can try it again tomorrow unless you beat me to it. :)

hcho3:
@trivialfis Notice that parallelization happens over columns, not rows

trivialfis:
@hcho3 I see.  Sorry for the mistake.

hcho3:
@trivialfis No problem. I should probably add another test case with multiple batches.

trivialfis:
@hcho3 That would be great.  I wanted to add some tests for histogram, but it's quite weird since quantile is not well tested.  I don't know how to derive the solution for tests.

src/common/hist_util.cc
```diff
@@ -35,30 +35,62 @@ void HistCutMatrix::Init(DMatrix* p_fmat, uint32_t max_num_bins) {
...
+  std::vector<bst_uint> const& group_ptr = info.group_ptr_;
+  size_t const num_groups = group_ptr.size() == 0 ? 0 : group_ptr.size() - 1;
+  // Use group index for weights?
+  bool const use_group_ind = num_groups != 0 && weights.size() != info.num_row_;
+
   for (const auto &batch : p_fmat->GetRowBatches()) {
-    #pragma omp parallel num_threads(nthread)
+    size_t group_ind;
+    bool group_ind_found = false;
+    if (use_group_ind) {
```
trivialfis: Move this test out of for loop.  If we don't use group, then there's no need for this loop.

src/common/hist_util.cc
```diff
@@ -35,30 +35,62 @@ void HistCutMatrix::Init(DMatrix* p_fmat, uint32_t max_num_bins) {
...
+      // Find the group that contains first row of the data batch
+      size_t const base_rowid = batch.base_rowid;
+      for (size_t i = 0; i < num_groups; ++i) {
```
trivialfis: Maybe a binary search is more appropriate since `group_ptr` is sorted by default.

hcho3:
@trivialfis It turns out that currently you cannot enable external memory for fast quantile updater. See #4093. I've added a test anyway, so that we can later add external memory support to fast quantile.

src/common/hist_util.cc
hcho3: Not sure if we can do that, since this loop will have to run once for every batch (note the use of `batch.base_rowid`.

src/common/hist_util.cc
hcho3: Can you look at it?

trivialfis: Yes.

src/common/hist_util.cc
trivialfis: I will try it tomorrow, thanks for all the help!

hcho3:
@trivialfis Looks like I uncovered a bug in external memory. I'll remove the failing test and file an new issue.

trivialfis:
@hcho3 It bugs me to add such an overhead.

In the context of using external memory, does the data come in as original order?  If so we can avoid adding such search completely.

trivialfis:
@hcho3 I will keep looking at the external memory part later.  But if you know the answer then please let me know.  The data loading logic is quite confusing to me.  Many thanks!

trivialfis:
@hcho3 I added the monitor for profiling anyway.  I don't think the data will come in as original order.  And since the search performs in O(log n) with log(2^32) == 32, I'm gonna ignore it for now.  Sorry for the noise and ready for another review. :)

hcho3:
@trivialfis For now, it is safe to assume that the external pages are out-of-order. We'll have to dive deeper into external memory when we attempt to fix #4130.

trivialfis:
@hcho3 Could you help taking another look?  I don't think this PR has anything to do with failed R test.

trivialfis:
@hcho3 Mac tests in Travis is not looking good.  Is there anything we can do about it?

hcho3:
@trivialfis This time it’s a bug in Travis CI. Will file an issue today.

hcho3:
@trivialfis Eventually, I hope we find a better alternative for Mac builds, but unfortunately, we cannot get Mac workers with Jenkins since Amazon EC2 doesn't support it.

hcho3:
Filed a support ticket just now. Hope they get back to me soon.

hcho3:
@trivialfis Wow, they got back to me and fixed the issue right away. Builds are working now.

trivialfis:
@hcho3 That's really nice!

PR #4123, opened 2019-02-11
Update datatable usage

st-pasha:
This PR ensures that xgboost continues to work correctly with the upcoming changes to the `datatable` package.
- Prefer name `dt.Frame` instead of `dt.DataTable` (which was deprecated a year ago);
- `dt.Frame.internal` namespace will be removed; instead the underlying data pointers can be accessed using `datatable.internal.frame_column_data_r()` (which returns a `ctypes.c_void_p` object).

### DIFF
--- a/python-package/xgboost/compat.py
+++ b/python-package/xgboost/compat.py
@@ -49,7 +49,11 @@ except ImportError:
 
 # dt
 try:
-    from datatable import DataTable
+    import datatable
+    if hasattr(datatable, "Frame"):
+        DataTable = datatable.Frame
+    else:
+        DataTable = datatable.DataTable
     DT_INSTALLED = True
 except ImportError:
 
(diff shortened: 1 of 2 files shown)


CodingCat:
I am not a python expert, but can we have some tests to guard these code with compatibility? (we are missing many in various packages, python/jvm, not sure about R)

st-pasha:
@CodingCat Testing whether xgboost works with different versions of different libraries is possible, but not trivial. Because in python you can have only one version of a library installed at any given time, these tests would necessarily have to install one version, run a test, then reinstall another version, etc. Now, a test is not supposed to alter user's environment -- a better place to implement this entire process is from a CI such as Jenkins or Travis.

hcho3:
As @st-pasha points out, we should use CI to try out multiple versions of `datatable`.

trivialfis:
I'm going to merge this since all current tests are not changed and passed.  @hcho3  I doubt the possibility of testing each version of dependencies.  We have sklearn, pandas, datatable and later some for others GPU.  In the context of Python, easiest way is always staying with the latest version from pip and advice users to do the same (maybe with help from virtualenv or similar tools).

hcho3:
@trivialfis Yeah, you're right. Except for major dependencies (e.g. CUDA), we should just advise users to use the latest versions.

PR #4126, opened 2019-02-11
reformat benchmark_tree.py to get rid of lint errors

rongou:
Making the code a bit easier to read in IDEs like PyCharm. For now this is pure refactoring, but hoping to make this run faster to generate larger datasets.

@RAMitchell


rongou:
`r_test` is failing? Probably not my fault. :)

PR #4138, opened 2019-02-13
[jvm-packages] add `verbosity` param to make it easier to debug problems

rongou:
@CodingCat @RAMitchell


CodingCat:
merged, thanks

PR #4139, opened 2019-02-13
Fix for windows compilation

RAMitchell:
omp index variable must be a signed integral type.


RAMitchell:
Anyone have ideas on what's wrong with the R tests?

hcho3:
@RAMitchell It looks like a random failure of package installation. In general, we should migrate some tests to Jenkins, where we can pre-install packages for the sake of stability.

hcho3:
Re-based against latest master to fix the failing R test

Issue #4127, opened 2019-02-12
distributed `hist` tree method not working?


rongou:
I have a test that takes some rows from the mortgage dataset (https://rapidsai.github.io/demos/datasets/mortgage-data) and trains xgboost on spark. It uses the `auto/approx` tree method before and seems to work fine. I just built from master and tried the new `hist` method, but it seems to be giving me random predictions (AUC of 0.5). Has anyone tried the distributed `hist` method on a real dataset and does it work for you?

Here is the prediction from `auto/approx`:
```
+-----+--------------------+--------------------+--------------------+----------+
|label|            features|       rawPrediction|         probability|prediction|
+-----+--------------------+--------------------+--------------------+----------+
|    0|(27,[10,11,12,13,...|[0.00272440910339...|[1.00272440910339...|       0.0|
|    0|(27,[10,11,12,13,...|[6.00278377532959...|[1.00060027837753...|       0.0|
|    0|(27,[10,11,12,13,...|[3.54290008544921...|[1.00035429000854...|       0.0|
|    0|(27,[10,11,12,13,...|[9.58383083343505...|[1.00095838308334...|       0.0|
|    0|(27,[10,11,12,13,...|[-4.1690468788146...|[0.99958309531211...|       0.0|
|    0|(27,[10,11,12,13,...|[0.00151866674423...|[1.00151866674423...|       0.0|
|    0|(27,[10,11,12,13,...|[0.00215685367584...|[1.00215685367584...|       0.0|
|    0|(27,[10,11,12,13,...|[-0.0041129589080...|[0.99588704109191...|       0.0|
|    0|(27,[10,11,12,13,...|[-7.7697634696960...|[0.99922302365303...|       0.0|
|    0|(27,[10,11,12,13,...|[1.54614448547363...|[1.00015461444854...|       0.0|
|    0|(27,[10,11,12,13,...|[-0.0034057796001...|[0.99659422039985...|       0.0|
|    0|(27,[10,11,12,13,...|[-0.0017806887626...|[0.99821931123733...|       0.0|
|    0|(27,[10,11,12,13,...|[-0.0025569796562...|[0.99744302034378...|       0.0|
|    0|(27,[10,11,12,13,...|[-5.8969855308532...|[0.99941030144691...|       0.0|
|    0|(27,[10,11,12,13,...|[-3.9306282997131...|[0.99960693717002...|       0.0|
[...]

PR #4140, opened 2019-02-14
more correct way to build node stats in distributed fast hist

CodingCat:
the current approach is to choose the first feature's histogram to build node stats, 

however, when this feature is too sparse, it will lead to a tree without any growth

so this PR addresses the issue by 

(1) only syncing the root's stats

(2) utilizing the stats cache implemented in #4015 to get all the others' stats

closes https://github.com/dmlc/xgboost/issues/4127


CodingCat:
there is a bug when dealing with dense data layout................sigh....

hcho3:
@CodingCat Do you need help?

CodingCat:
@hcho3 thanks, it's fine now, though the logic for depthwise becomes a bit tricky

CodingCat:
it's ready to review now

CodingCat:
@trivialfis @RAMitchell @hcho3  I believe it is the last commit I would put in distributed hist....thanks for the review

RAMitchell:
Can you please explain a little bit in words the purpose of this PR and what led you to make these changes? I think I have an idea but I want to make sure I understand. 

I don't like using the last slot of the histogram to store the node stats. I believe if something has a distinct purpose it should be a separate variable. This can very easily create problems later given that the histogram array may not always be the same size depending on some state. I definitely recommend reading this as a reference for c++ architecture: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rp-direct

Some recent changes of mine (#4015) may actually mean that we can get the summary statistics from a cache after split finding and we do not need to recalculate these at every node, only the root. This might simplify your PR.

CodingCat:
@RAMitchell so the current approach in master branch is to use first feature's histogram to build the node's stats, which assumes that the feature is not "too sparse" and summing up all values in the bins across workers will make an approximate result....but it's not true for some cases, and it will lead to a no-grow tree as indicated in #4127


using the last bin to store stats is to mimic the behavior of `approx` which is motivated by saving a rabit call (which is slow and the main reason that we don't see a significant speedup when increasing worker number beyond some values) 

If I understand #4015 correctly, a lot of code (e.g. https://github.com/dmlc/xgboost/blob/d506a8bc633fea3c8d535c70a5c11047122f07b5/src/tree/updater_quantile_hist.cc#L757-L796) is unnecessary?

CodingCat:
I like the idea in #4015 @RAMitchell , this should save a significant amount of cpu cycles.....I would move forward to remove `InitNewNode` in `hist` here....

CodingCat:
@RAMitchell I have made the changes based on #4015, thanks for the review

hcho3:
@CodingCat Is this ready to merge?

CodingCat:
I think it's ready, waiting for confirm from anyone of you

PR #4142, opened 2019-02-14
Fix broken R test: Install Homebrew GCC

hcho3:
Missing GCC Fortran causes installation failure of a dependency package (igraph).

@trivialfis @CodingCat @RAMitchell This should fix the R test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,6 +68,7 @@ addons:
       - g++-7
   homebrew:
     packages:
+      - gcc@7
       - graphviz
     update: true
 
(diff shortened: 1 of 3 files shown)


trivialfis:
@hcho3 If this is a bug in some R dependencies, we might just report it to the upstream ...  Watching your commit history I can almost feel your pain. :)

hcho3:
@trivialfis It appears to be configuration problem, where the Fortran compiler is not located properly. Looks like there is some difference between my Macbook Pro and Travis CI.

trivialfis:
@hcho3 Not having the correct toolchain after "configure" phase is a bug in build system.   But it's autotools ...  We can still raise an issue: https://github.com/igraph/rigraph

hcho3:
@trivialfis And now other dependencies are randomly failing to download :(

hcho3:
Let's see if we can come up with a minimal reproduction of the build bug in igraph...

hcho3:
Phew, the R build is finally fixed.

trivialfis:
Shall we merge it?

PR #4147, opened 2019-02-15
Accept numpy array view.

### LINKED ISSUE #3841: [BLOCKING] [python] DMatrix setting incorrect values from sliced array
On xgboost 0.80 (tested on Windows with Python 2 and Python 3), creating a DMatrix from a sliced array will result in incorrect values being copied in some cases. Here is a minimal repro:
```python
import numpy as np
import xgboost as xgb

y = np.array([12, 34, 56], np.float32)[::2]
print(xgb.DMatrix([], y).get_label())  # prints [12, 34] instead of [12, 56]
print(xgb.DMatrix([], y + 0).get_label())  # prints [12, 56] as expected
```

This issue has popped up for LightGBM before (Microsoft/LightGBM#1205), for which their solution is to test `label.flags.c_contiguous` and make a copy of the array as necessary.

trivialfis:
close #3841 .

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -219,6 +219,17 @@ def c_array(ctype, values):
     return (ctype * len(values))(*values)
 
 
+def _get_length_and_stride(data):
+    "Return length and stride of 1-D data."
+    if isinstance(data, np.ndarray) and data.base is not None:
+        length = len(data.base)
+        stride = data.strides[0] // data.dtype.itemsize
+    else:
+        length = len(data)
+        stride = 1
+    return length, stride
+
+
 PANDAS_DTYPE_MAPPER = {'int8': 'int', 'int16': 'int', 'int32': 'int', 'int64': 'int',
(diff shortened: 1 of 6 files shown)


python-package/xgboost/core.py
```diff
@@ -715,9 +737,7 @@ def set_group(self, group):
         group : array like
             Group size of each group
         """
-        _check_call(_LIB.XGDMatrixSetGroup(self.handle,
-                                           c_array(ctypes.c_uint, group),
-                                           c_bst_ulong(len(group))))
+        self.set_uint_info('group', group)
```
hcho3: Are you sure this is equivalent to `XGDMatrixSetGroup()`?

trivialfis: Good point.  `SetInfo` does a copy first, while `SetGroup` builds the `group_ptr` on the fly, which should be slightly faster.  But other than this, the code is duplicated.  It's a little bit tricky to add a test, since we can't get the group back without another function transforming `group_ptr` into the original `group`.  We can postpone it to #4150.  But I am happy to add such test if you think it's better.

PR #4148, opened 2019-02-15
Added SKL-like random forest Python API.

canonizer:
- added XGBRFClassifier and XGBRFRegressor classes to Scikit-Learn-like xgboost API
- added documentation describing how to use xgboost for random forests,
  as well as existing caveats

### DIFF
--- a/doc/rf.rst
+++ b/doc/rf.rst
@@ -58,9 +58,9 @@ A random forest model can then be trained as follows::
   bst = train(params, dmatrix, num_boost_round=1)
 
 
-*****************
-With SKL-Like API
-*****************
+**************************
+With Scikit-Learn-Like API
+**************************
 
 ``XGBRFClassifier`` and ``XGBRFRegressor`` are SKL-like classes that provide random forest
 functionality. They are basically versions of ``XGBClassifier`` and ``XGBRegressor`` that
--- a/doc/rf.rst
+++ b/doc/rf.rst
@@ -1,0 +1,89 @@
+#########################
+Random Forests in XGBoost
+#########################
+
+XGBoost is normally used to train gradient-boosted decision trees and other gradient
+boosted models. Random forests use the same model representation and inference, as
+gradient-boosted decision trees, but a different training algorithm. There are XGBoost
+parameters that enable training a forest in a random forest fashion.
+
+
+****************
+With XGBoost API
+****************
+
+The following parameters must be set to enable random forest training.
+
+* ``booster`` should be set to ``gbtree``, as we are training forests. Note that as this
+  is the default, this parameter needn't be set explicitly.
+* ``subsample`` must be set to a value less than 1 to enable random selection of training
+  cases (rows).
+* One of ``colsample_by*`` parameters must be set to a value less than 1 to enable random
+  selection of columns. Normally, ``colsample_bynode`` would be set to a value less than 1
+  to randomly sample columns at each tree split.
+* ``num_parallel_tree`` should be set to the size of the forest being trained.
+* ``num_boost_round`` should be set to 1. Note that this is a keyword argument to
+  ``train()``, and is not part of the parameter dictionary.
+* ``eta`` (alias: ``learning_rate``) must be set to 1 when training random forest
+  regression.
+* ``random_state`` can be used to seed the random number generator.
+
+  
+Other parameters should be set in a similar way they are set for gradient boosting. For
+instance, ``objective`` will typically be ``reg:linear`` for regression and
+``binary:logistic`` for classification, ``lambda`` should be set according to a desired
+regularization weight, etc.
+
+If both ``num_parallel_tree`` and ``num_boost_round`` are greater than 1, training will
+use a combination of random forest and gradient boosting strategy. It will perform
+``num_boost_round`` rounds, boosting a random forest of ``num_parallel_tree`` trees at
+each round. If early stopping is not enabled, the final model will consist of
+``num_parallel_tree`` * ``num_boost_round`` trees.
+
+Here is a sample parameter dictionary for training a random forest on a GPU using
+xgboost::
+
+  params = {
+    'colsample_bynode': 0.8,
+    'learning_rate': 1,
+    'max_depth': 5,
+    'num_parallel_tree': 100,
+    'objective': 'binary:logistic',
+    'subsample': 0.8,
+    'tree_method': 'gpu_hist'
+  }
+
+A random forest model can then be trained as follows::
+
+  bst = train(params, dmatrix, num_boost_round=1)
+
+
+*****************
+With SKL-Like API
+*****************
+
+``XGBRFClassifier`` and ``XGBRFRegressor`` are SKL-like classes that provide random forest
+functionality. They are basically versions of ``XGBClassifier`` and ``XGBRegressor`` that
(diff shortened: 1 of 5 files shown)


canonizer:
I've removed `n_gpus` and `gpu_id`. However, I've left `tree_method` and `colsample_bynode`. Please indicate if you want me to remove those two (note that other `colsample_by*` parameters are already part of SKL-like classes).

I've also removed `self.mode`, and now handle it using subclass methods.

RAMitchell:
@trivialfis can I get your review.

@canonizer xgboost's default tree_method is currently 'exact'. This PR would change the sklearn interface to use 'hist' as the default. This would be a huge change for our user base. Arguably we should in fact make this change, but that is a big decision and that we need some internal discussion on.

trivialfis:
@RAMitchell Will review shortly.

doc/rf.rst
```diff
@@ -0,0 +1,89 @@
...
+
+* ``booster`` should be set to ``gbtree``, as we are training forests. Note that as this
+  is the default, this parameter needn't be set explicitly.
+* ``subsample`` must be set to a value less than 1 to enable random selection of training
+  cases (rows).
+* One of ``colsample_by*`` parameters must be set to a value less than 1 to enable random
+  selection of columns. Normally, ``colsample_bynode`` would be set to a value less than 1
+  to randomly sample columns at each tree split.
+* ``num_parallel_tree`` should be set to the size of the forest being trained.
+* ``num_boost_round`` should be set to 1. Note that this is a keyword argument to
```
trivialfis: In non-boosting scenario?

doc/rf.rst
```diff
@@ -0,0 +1,89 @@
...
+    'tree_method': 'gpu_hist'
+  }
+
+A random forest model can then be trained as follows::
+
+  bst = train(params, dmatrix, num_boost_round=1)
+
+
+*****************
+With SKL-Like API
```
trivialfis: Scikit-Learn in the header might be better.  This is actually really the first time I see this abbreviation.

python-package/xgboost/sklearn.py
```diff
@@ -863,12 +874,76 @@ def evals_result(self):
...
+
+    def __init__(self, max_depth=3, learning_rate=1, n_estimators=100, silent=True,
+                 objective="binary:logistic", tree_method='hist',
+                 n_jobs=1, nthread=None, gamma=0, min_child_weight=1,
+                 max_delta_step=0, subsample=0.8, colsample_bytree=1, colsample_bylevel=1,
+                 colsample_bynode=0.8, reg_alpha=0, reg_lambda=1, scale_pos_weight=1,
+                 base_score=0.5, random_state=0, seed=None, missing=None, **kwargs):
+        super(XGBRFClassifier, self).__init__(
+            max_depth=max_depth, learning_rate=learning_rate, n_estimators=n_estimators,
+            silent=silent, objective=objective, booster='gbtree', tree_method=tree_method,
```
trivialfis: Let's use `verbosity` instead of `slient`.  The later is deprecated.

python-package/xgboost/sklearn.py
```diff
@@ -863,12 +874,76 @@ def evals_result(self):
...
         + '\n'.join(XGBModel.__doc__.split('\n')[2:])
 
 
+class XGBRFRegressor(XGBRegressor):
+    # pylint: disable=missing-docstring
+    __doc__ = "Implementation of the scikit-learn API "\
+              + "for XGBoost random forest regression.\n\n"\
+              + '\n'.join(XGBModel.__doc__.split('\n')[2:])
+
+    def __init__(self, max_depth=3, learning_rate=1, n_estimators=100, silent=True,
```
trivialfis: Same here, use `verbosity` instead of `silent`.

python-package/xgboost/sklearn.py
```diff
@@ -61,14 +61,17 @@ class XGBModel(XGBModelBase):
     learning_rate : float
         Boosting learning rate (xgb's "eta")
     n_estimators : int
-        Number of boosted trees to fit.
+        Number of trees to fit. If the mode is gbm, this is the number of boosted rounds.
```
trivialfis: I don't understand why do we need to split it into two modes?  In my conceptual model,  XGBoost is a boosting framework, `tree`, `linear` and `random forest` are different base models of `XGBoost` that being boosted.  Adding `XGBoostRF*` makes sense to me since it creates a better view of different base models, we can apply the same to `linear` model and `tree` model so we have `XGBoostLinear*` and `XGBoostTree*`.  But I want to keep the semantic of parameters like `n_estimators` unchanged and mean the same thing in different contexts.  We can declare `n_estimator` exclusively in `XGBoostRF*` so that other two interfaces won't accept this parameters accidentally.

I'm open to discussions on this.

python-package/xgboost/sklearn.py
```diff
@@ -132,10 +137,10 @@ class XGBModel(XGBModelBase):
     """
 
     def __init__(self, max_depth=3, learning_rate=0.1, n_estimators=100,
-                 silent=True, objective="reg:linear", booster='gbtree',
-                 n_jobs=1, nthread=None, gamma=0, min_child_weight=1, max_delta_step=0,
-                 subsample=1, colsample_bytree=1, colsample_bylevel=1,
-                 reg_alpha=0, reg_lambda=1, scale_pos_weight=1,
+                 silent=True, objective="reg:linear", booster='gbtree', tree_method='hist',
```
trivialfis: Is changing `tree_method` a must?  We do want to change our default tree method in recent future, but preferably through low level `c++` code.

doc/rf.rst
canonizer: Yes, as xgboost API was designed around boosting, and the default value for `num_boost_round` is 10.

It is also possible to set both `num_boost_round` and `num_parallel_tree` to a value larger than 1. In this case, the classifier would be trained using a combination of gradient boosting and random forests.

doc/rf.rst
canonizer: Done.

python-package/xgboost/sklearn.py
canonizer: Done.

python-package/xgboost/sklearn.py
canonizer: Done.

python-package/xgboost/sklearn.py
canonizer: I actually replaced `silent` with `verbosity` throughout this file.

python-package/xgboost/sklearn.py
canonizer: `n_estimators` always has the meaning of the number of simple classifiers used (in this case, trees). The only difference is in how they are trained.

I removed the comment regarding the mode, as it no longer applies.

python-package/xgboost/sklearn.py
canonizer: Removed `tree_method`.

trivialfis:
@RAMitchell @hcho3 WDYT?

trivialfis:
I will make a separated PR for marking this experimental later.  Let's merge it now. @canonizer @RAMitchell @hcho3 .

canonizer:
#4255 brings back the silent parameter and marks it deprecated.

PR #4149, opened 2019-02-15
Fix clang-tidy.

trivialfis:


trivialfis:
@hcho3 Sorry for flushing the tests farm,  but I have to see Jenkins in action. :(

trivialfis:
Finally closing to complete...  @hcho3 @RAMitchell Here are some status reports:

- clang-tidy is not working perfectly.  There's a missing namespace declaration of `cub` in `thrust`.  My suggestion is patching thrust with one line `using namespace cub`.  CUDA 10.0/1 are not tested yet, which might be interesting.

- Due to above mentioned problem, some warnings in gpu side appears on my machine doesn't appear on Jenkins.  Two of them are noteworthy,  in `updater_gpu`, there's memory address raw casting which might create "garbage value" according to clang's term. `atomicAdd` in `gpu_updater_common` is not properly handled by `clang-tidy` but OK with clang ...

- I opened a PR in `dmlc-core` to workaround a clang bug, see https://github.com/dmlc/dmlc-core/pull/513.

- As of writing, tidy for unitests code is disabled, will add it back at next commit.  Don't worry, warnings in there are addressed.

- GTest is upgraded to 1.8 for clang-tidy to reduce number of unfiltered warnings.

trivialfis:
This is rather huge and tedious, so maybe it's a good idea to review it piece at a time ...

src/common/host_device_vector.cu
```diff
@@ -46,8 +46,8 @@ template <typename T>
 struct HostDeviceVectorImpl {
   struct DeviceShard {
     DeviceShard()
-        : proper_size_(0), device_(-1), start_(0), perm_d_(false),
-        cached_size_(~0), vec_(nullptr) {}
+        : proper_size_{0}, device_{-1}, start_{0}, perm_d_{false},
+      cached_size_{0}, vec_{nullptr} {}
```
RAMitchell: cached_size_(~0) -> cached_size_{0}

Doesn't this change the behaviour due to removing the '~' operator ?

src/common/host_device_vector.cc
```diff
@@ -93,16 +98,16 @@ common::Span<const T> HostDeviceVector<T>::ConstDeviceSpan(int device) const {
 }
 
 template <typename T>
-std::vector<T>& HostDeviceVector<T>::HostVector() { return impl_->data_h_; }
+std::vector<T>& HostDeviceVector<T>::HostVector() { return impl_->Vec(); }
 
 template <typename T>
 const std::vector<T>& HostDeviceVector<T>::ConstHostVector() const {
-  return impl_->data_h_;
+  return impl_->Vec();
```
hcho3: I see lots of replacement of member access with getters. Does clang-tidy always prefer getters/setters to direct member access?

tests/ci_build/clang_tidy.sh
```diff
@@ -1,12 +1,15 @@
 #!/bin/bash
 
-rm -rf gtest googletest-release-1.7.0
-wget -nc https://github.com/google/googletest/archive/release-1.7.0.zip
-unzip -n release-1.7.0.zip
-mv googletest-release-1.7.0 gtest && cd gtest
-cmake . && make
-mkdir lib && mv libgtest.a lib
+export GTEST_NAME=googletest-release-1.8.1
```
hcho3: Can we also use a variable for the zip file name?
```bash
export GTEST_PKG_NAME=release-1.8.1
export GTEST_NAME=googletest-${GTEST_PKG_NAME}
...
wget -nc https://github.com/google/googletest/archive/${GTEST_PKG_NAME}.zip
```

src/common/host_device_vector.cc
trivialfis: No.  Private member is required to add a "_" suffix while public member should not contain such suffix.  So sometimes I change public member into private in order to preserve the suffix.

trivialfis:
@hcho3 Changed sh script for preparing GTest.

trivialfis:
@RAMitchell @hcho3 Thanks!

PR #4152, opened 2019-02-16
Added trees_to_dataframe() method for Booster class

pford221:
Worked alongside @hshujuan for this PR to add nearly equivalent methodology to python package as R for parsing a dumped text model to a pandas DataFrame. Testing on model built on agaricus dataset included...

- Parse 10,000 trees of `max_depth` 5 in  about 3 seconds
- Parse `gbtree` and `dart` boosters
- Output equivalent to R's `xgb.model.dt.tree()` function
- Can handle trees that have no splits (i.e. leaf only trees)

If any thing about this PR is not adequate, please let me know.

Thanks,
Patrick

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -4,11 +4,13 @@
 """Core XGBoost Library."""
 from __future__ import absolute_import
 import collections
+from IPython.core.debugger import set_trace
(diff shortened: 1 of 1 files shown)


pford221:
Yes. I will do that now.

PR #4153, opened 2019-02-16
Added trees_to_df() method for Booster class

pford221:
Worked alongside @hshujuan for this PR to add nearly equivalent methodology to python package as R for parsing a dumped text model to a pandas DataFrame. Testing on model built on agaricus dataset included...

* Parse 10,000 trees of `max_depth` 5 in  about 3 seconds
* Parse `gbtree` and `dart` boosters
* Output equivalent to R's `xgb.model.dt.tree()` function
* Can handle trees that have no splits (i.e. leaf only trees)

If any thing about this PR is not adequate, please let me know.

Thanks,
Patrick

### DIFF
--- a/python-package/.pylintrc
+++ b/python-package/.pylintrc
@@ -3,6 +3,7 @@
 ignore=tests
 
 extension-pkg-whitelist=numpy
+generated-members=pandas.*
 
 disiable=unexpected-special-method-signature,too-many-nested-blocks
 
--- a/python-package/.pylintrc
+++ b/python-package/.pylintrc
@@ -3,7 +3,7 @@
 ignore=tests
 
 extension-pkg-whitelist=numpy
-generated-members=pandas.*
+generated-members=pandas.DataFrame.*
 
 disiable=unexpected-special-method-signature,too-many-nested-blocks
 
--- a/python-package/.pylintrc
+++ b/python-package/.pylintrc
@@ -3,7 +3,6 @@
 ignore=tests
 
 extension-pkg-whitelist=numpy
-generated-members=pandas.DataFrame.*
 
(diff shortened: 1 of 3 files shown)


pford221:
Hi,

I'm having trouble understanding the logs for why there was an error. Can someone guide me on what needs to be changed?

Thanks,
Patrick

hcho3:
@pford221 The error is due to the Python coding style convention.

hcho3:
The tests should pass now.

pford221:
Thank you, @hcho3 , for cleaning up the line indentations and for contemplating backwards compatibility for older versions of pandas. Let me know if there's something I can do. My team at work uses the functionality at work frequently so I think it would be a great addition to the python library.

Patrick

trivialfis:
Could you add a simple test in tests/python ?

pford221:
Sure. Would confirming that the sum of all the gain in the dataframe (post processing) and in the dumped text model between (pre processing) are the same? Is it OK to use the agaricus dataset for this test?

trivialfis:
@pford221 Yes.  We plan to revise the dumping logic in the future.  Just to make sure we won't break your code.

trivialfis:
@pford221 For the failed linting, rebasing on master branch should resolve the problem.

pford221:
Hi,

I added  test_parse_tree.py to tests/python and squashed all the commits into a single commit. I studied the other tests and tried to write this test in the same style, but I have to admit this is my first test writing a test like this, so let me know if I did anything wrong.

Thanks,
Patrick

pford221:
Thanks for fixing the formatting @hcho3. Is there anything else needed for this PR?

hcho3:
@pford221 Thanks for your contribution. It is now merged.

PR #4156, opened 2019-02-17
demo's doc: Update README.MD to fix typo

keni-chi:
I noticed "Benckmark" was used at the demo/kaggle-higgs/README.MD instead of "Benchmark".
If that was intentional then ignore this change.

### DIFF
--- a/demo/kaggle-otto/README.MD
+++ b/demo/kaggle-otto/README.MD
@@ -1,4 +1,4 @@
-Benckmark for Otto Group Competition
+Benchmark for Otto Group Competition
 =========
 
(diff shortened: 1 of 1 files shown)


hcho3:
@trivialfis @RAMitchell  I'm seeing this error lately in unit test:
```
[ RUN      ] GpuHist.ApplySplit
unknown file: Failure
C++ exception with description "vector::_M_range_check" thrown in the test body.
```

Any idea what's going on?

trivialfis:
That's weird.  Seems not deterministic.  Will try looking into it.

trivialfis:
@keni-chi Thanks!

PR #4161, opened 2019-02-19
Fix incorrect device in multi-GPU algorithm

RAMitchell:
@trivialfis here is a challenge for you :) Find a way for us to eliminate bugs completely due to incorrect device selection.

I feel we will keep encountering these mistakes unless we come up with a better way.

### DIFF
--- a/src/tree/updater_gpu_hist.cu
+++ b/src/tree/updater_gpu_hist.cu
@@ -300,6 +300,7 @@ struct DeviceHistogram {
   void AllocateHistogram(int nidx) {
     if (HistogramExists(nidx)) return;
 
+    dh::safe_cuda(cudaSetDevice(device_id_));
     if (data.size() > kStopGrowingSize) {
       // Recycle histogram memory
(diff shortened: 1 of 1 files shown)


hcho3:
@RAMitchell How did you spot the bug?

I have interesting idea: write the multi-GPU algorithm as a formal spec and validate it with a model checker. While formal verification won’t prevent software bugs, it will prevent incorrect design.

https://www.amazon.com/Practical-TLA-Planning-Driven-Development/dp/1484238281

trivialfis:
Will think about that.  :)

RAMitchell:
@hcho3 I spotted it by looking at the area of code throwing an unrelated exception, I have seen this before so I was looking for it. In general though these multi gpu bugs are nondeterminustic and extremely hard to diagnose.

Your link looks interesting I have never done formal verification before.

hcho3:
@RAMitchell Got it. Yeah, this kind of bug is really hard to detect. I want to spend some time to learn formal verification to reduce the chance of concurrency bugs.

Issue #4150, opened 2019-02-16
Eliminate extra dataset copy in Python.


trivialfis:
During construction of DMatrix, Python wrapper might duplicate the dataset because it's not contiguous or not of the right data type. But we can handle these situations inside c++ code to avoid constructing an extra copy of dataset.

My goal is not to share the underlying buffer with Python data structure, but to eliminate the extra copies constructed during conversion inside Python wrapper.

hcho3:
Fixed in #4147

PR #4165, opened 2019-02-19
Fix #4163: always copy sliced data

### LINKED ISSUE #4163: [BLOCKING] /root/repo/xgboost/src/objective/regression_obj.cu:65: Check failed: preds.Size() == info.labels_.Size() (904 vs. 3616) labels are not correctly providedpreds.size=904, label.size=3616
1dac5e24108890c946c5a4673f0d1fa4d6125943 on fresh build off dmlc

[failxgbregression.zip](https://github.com/dmlc/xgboost/files/2881889/failxgbregression.zip)

@hcho3 seems to be **major** regression after recent fixes.  Seems all regression problems have the issue.

hcho3:
#4147 attempted to read sliced NumPy array directly into the C++ backend, but it relied on a NumPy internal variable (`data.base`) whose meaning we are not able to determine precisely. In particular, `len(data.base)` can sometimes mean the number of bytes in the NumPy array and other times mean the number of elements in the array.

Rather than dealing with undocumented NumPy internals, we'll take the approach of Microsoft/LightGBM#1210, i.e. always copy sliced data. This approach increases memory consumption slightly but the effect is not too bad. After all, the data matrix gets always copied when the input is a slice:
https://github.com/dmlc/xgboost/blob/c8c472f39aa08d5b8f180076a429f09317451cc7/python-package/xgboost/core.py#L487-L490

Closes #4163

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -219,17 +219,6 @@ def c_array(ctype, values):
     return (ctype * len(values))(*values)
 
 
-def _get_length_and_stride(data):
-    "Return length and stride of 1-D data."
-    if isinstance(data, np.ndarray) and data.base is not None:
-        length = len(data.base)
-        stride = data.strides[0] // data.dtype.itemsize
-    else:
-        length = len(data)
-        stride = 1
-    return length, stride
-
-
 PANDAS_DTYPE_MAPPER = {'int8': 'int', 'int16': 'int', 'int32': 'int', 'int64': 'int',
                        'uint8': 'int', 'uint16': 'int', 'uint32': 'int', 'uint64': 'int',
                        'float16': 'float', 'float32': 'float', 'float64': 'float',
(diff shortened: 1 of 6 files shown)


hcho3:
We'll need to investigate a better way to eliminate copies, per #4150.

python-package/xgboost/core.py
```diff
@@ -637,15 +631,14 @@ def set_uint_info(self, field, data):
...
-        _check_call(_LIB.XGDMatrixSetUIntInfoStrided(
-            self.handle,
-            c_str(field),
-            c_data,
-            c_bst_ulong(stride),
-            c_bst_ulong(length)))
+        if getattr(data, 'base', None) is not None and \
+           data.base is not None and isinstance(data, np.ndarray) \
+           and isinstance(data.base, np.ndarray) and (not data.flags.c_contiguous):
+            raise ValueError('Sliced data is not allowed for uint property')
```
trivialfis: Why not?

python-package/xgboost/core.py
```diff
@@ -595,14 +584,15 @@ def set_float_info(self, field, data):
         data: numpy array
             The array of data to be set
         """
+        if getattr(data, 'base', None) is not None and \
+           data.base is not None and isinstance(data, np.ndarray) \
+           and isinstance(data.base, np.ndarray) and (not data.flags.c_contiguous):
+            raise ValueError('Should use set_float_info_npy2d() to use sliced data')
```
trivialfis: Can we just redirect that to that call?

python-package/xgboost/core.py
hcho3: Right now, there is only one `uint` type field called `root_index`. Do we need slicing for this?

trivialfis: We are here anyway, no need to leave something can be done in one line I think.  But go as you think best.

hcho3: Addressed in the latest commit

CodingCat:
this one is ready to go?

Issue #3499, opened 2018-07-23
Multi Node GPU Support ?


SemiCode1220:
Hi guys !

I wanted to know if anyone is working on extending GPU support to the multi-node distributed setting? i.e: "gpu_approx"

As I understand, current gpu acceleration is only supported for the exact and hist versions.

RAMitchell:
cc @teju85 

I think there were some plans to build a distributed version using NCCL - not sure if they are high priority. Are you asking from the perspective of a user or potential contributor?

teju85:
@RAMitchell @SemiCode1220 
We certainly are exploring this as a possibility.
Hence, the PR #3404 to upgrade to NCCLv2.
I'll comment back here, once we have more to share.

SemiCode1220:
Could you point me towards the issue I should consult to able to contribute towards distributed-GPU support ?

RAMitchell:
I don't think we have any such issue so we can discuss here. Feel free to propose a solution to achieve this. I have not worked with the distributed algorithms before but presumably there needs to be some kind of AllReduce operation using rabit every time a histogram is generated in gpu_hist. We would also need to make sure each node is using the same quantile cuts. 

Ideally the AllReduce for histograms would be able to take place in device memory using NCCL but this is not currently possible in the rabit API. For now it would be a good proof of concept to show a distributed version that reduces the histograms in host memory only.

trivialfis:
About distributed xgboost on MPI based APIs, the MPI related code needs some rework. Currently xgboost doesn't build on latest MPI3.0, which dropped support for C++ interface that employed in xgboost. Maybe we can work on MPI and NCCL support altogether.

hcho3:
@trivialfis I think we'd need to modify `rabit` to work with MPI 3.0.

hcho3:
Multi-Node, Multi-GPU support was added in #4095.

Issue #2983, opened 2017-12-26
[jvm-packages] Support for GPU in Spark package


troszok:
Is there a way to enable GPU support in Spark package? 

When I tried to force GPU tree method I am getting: 
```
Exception in thread "main" java.lang.IllegalArgumentException: XGBoostEstimator_d9a9bf44f8ae parameter tree_method given invalid value gpu_hist
```

I also found in the documentation:
```
This plugin currently works with the CLI, python and R - see installation guide for details.
```
But I see some issues related to xgboost4j-jvm (for example: https://github.com/dmlc/xgboost/issues/2324) so people are using it.
 
Is it planned anytime soon or if there is a way to enable it?

RAMitchell:
I have added the call for contribution label. Would appreciate it if someone else could take this on!

mhmoudr:
This ticket has been closed, but I cannot see any PR linked to it, isn't there a planned support for GPUs in JVM packages?

hcho3:
@mhmoudr See #3499. Basically, the GPU algorithms need to support distributed (multi-node) training. Currently, the GPU algorithms only work on a single node.

PR #4175, opened 2019-02-22
add `gpu_hist` support to Spark

rongou:
Two parts to the PR:
* On the Spark/Scala side, it's mostly plumbing.
* On the C++/CUDA side, create the NCCL communicator for distributed mode, with support for different GPUs per node (e.g. a mixture of DGX-1s and DGX-2s). Partially based on #4095.

I've tested this on GCP with a 20-node Spark Standalone cluster, 1 T4 GPU per node.

@RAMitchell @canonizer @CodingCat @mt-jones

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/BoosterParams.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/BoosterParams.scala
@@ -145,11 +145,13 @@ private[spark] trait BoosterParams extends Params {
   final def getAlpha: Double = $(alpha)
 
   /**
-   * The tree construction algorithm used in XGBoost. options: {'auto', 'exact', 'approx'}
+   * The tree construction algorithm used in XGBoost. options: {'auto', 'exact', 'approx', 'hist',
+   * 'gpu_hist', 'gpu_exact'}
    *  [default='auto']
    */
   final val treeMethod = new Param[String](this, "treeMethod",
-    "The tree construction algorithm used in XGBoost, options: {'auto', 'exact', 'approx', 'hist'}",
+    "The tree construction algorithm used in XGBoost, options: {'auto', 'exact', 'approx', " +
+    "'hist', 'gpu_hist', 'gpu_exact'}",
     (value: String) => BoosterParams.supportedTreeMethods.contains(value))
 
   final def getTreeMethod: String = $(treeMethod)
@@ -284,7 +286,7 @@ private[spark] object BoosterParams {
 
   val supportedBoosters = HashSet("gbtree", "gblinear", "dart")
 
-  val supportedTreeMethods = HashSet("auto", "exact", "approx", "hist")
+  val supportedTreeMethods = HashSet("auto", "exact", "approx", "hist", "gpu_hist", "gpu_exact")
 
   val supportedGrowthPolicies = HashSet("depthwise", "lossguide")
 
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/GeneralParams.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/GeneralParams.scala
@@ -57,6 +57,14 @@ private[spark] trait GeneralParams extends Params {
   final def getNthread: Int = $(nthread)
 
   /**
+   * number of GPUs to use per worker: -1=use all GPUs. default 1
+   */
+  final val nGpus = new IntParam(this, "nGpus", "number of GPUs to use per worker",
+    ParamValidators.gtEq(-1))
+
+  final def getNGpus: Int = $(nGpus)
+
+  /**
    * whether to use external memory as cache. default: false
    */
   final val useExternalMemory = new BooleanParam(this, "useExternalMemory",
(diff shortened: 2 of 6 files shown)


CodingCat:
Thanks for the contribution

First of all, we should not merge this at least for 0.82 release, as it is never tested by anyone except the author 

Second, I don't think it's ready to support GPU in XGBoost-Spark since Spark itself is far from being ready:

1. I didn't see there is a clear agreement in the Spark community about how to support GPU (e.g. https://github.com/apache/spark/pull/23347)

2. there are a lot of issues to be resolved to get a mature support for GPU in Spark per our colleagues in Uber, e.g. when I run a task for a stage with CPU, I will release right after the task is done, however with GPU, the GPU resources have to be allocated for the whole life cycle of Spark application (even for Databricks, they admit everything is in beta https://docs.databricks.com/user-guide/clusters/gpu.html )

3. even Spark does not have particular CI tests for covering its integration with GPU.....it's hard for us to ensure the quality of XGBoost-Spark as a downstream library with this feature.....

so I didn't see I will agree to have this fancy feature in near future

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/GeneralParams.scala
```diff
@@ -171,7 +179,7 @@ private[spark] trait GeneralParams extends Params {
 
   final def getSeed: Long = $(seed)
 
-  setDefault(numRound -> 1, numWorkers -> 1, nthread -> 1,
+  setDefault(numRound -> 1, numWorkers -> 1, nthread -> 1, nGpus -> 1,
```
CodingCat: this should be 0 at least

rongou:
@CodingCat Thanks for the quick review! I don't know about fancy, XGBoost GPU support was added in 2016. :)

Yeah it's fine to hold off until 0.82 is released, and I totally agree not everything is ready, but isn't this the whole point of open source development, the Cathedral and the Bazaar, release early and release often, given enough eyeballs all bugs are shallow, etc.

To address your specific concerns:
1. The issue with  apache/spark#23347 is mainly about how to custom docker images for Spark in general, not specific to GPU support. There is a JIRA issue to figure this out: https://issues.apache.org/jira/browse/SPARK-24655.
1. Yes currently Spark is not really aware of any accelerators, but I understand Databricks is working on it. Until that is ready, maybe XGBoost-Spark can have very limited GPU support that's not ideal or perfect in any way but at least a working solution? It can also provide an important use case for Databricks to optimize.
1. I can work on making the current CI happy, and perhaps add some tests for distributed GPU training if Jenkins supports it.

Looking through past issues, people have asked for this (https://github.com/dmlc/xgboost/issues/2983, https://github.com/dmlc/xgboost/issues/3499), so at least there is some demand.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/GeneralParams.scala
rongou: I'd be happy to change it, but it's only used when `tree_method` is set to `gpu_hist`, in which case the user probably expects to grab a GPU. It's also the default on the C++ side (https://github.com/dmlc/xgboost/blob/master/src/tree/param.h#L200), so having this default seems less surprising and more consistent.

CodingCat:
> XGBoost GPU support was added in 2016. :)

XGBoost GPU != XGBoost-Spark GPU

> release early and release often, given enough eyeballs all bugs are shallow, etc.

It's not the same situation, the current issue we know there is a bug and we push this known bug to the user? 

>  The issue with  [apache/spark#23347](https://github.com/apache/spark/pull/23347) is mainly about how to custom docker images for Spark

here are something copied from that thread 

"One other concern is that adding another option that we say is ready to run out of the box for GPUs, is that we have to maintain this mode and ensure it is tested in CI." - this is from mccheah, which is one of the main persons behind spark@k8s which is the only way to make Spark "run" with GPU

> Until that is ready, maybe XGBoost-Spark can have very limited GPU support that's not ideal or perfect in any way but at least a working solution

no, this is not the definition of "working" especially nowadays XGBoost is not a research project anymore...if you track how Spark accepts features, it is more and more conservative along the way

> I can work on making the current CI happy, and perhaps add some tests for distributed GPU training if Jenkins supports it.

I mean Spark, as the base of XGBoost-Spark, should prove that it is supporting GPU in a mature way, so feel free to work with Spark community on this

> Looking through past issues, people have asked for this (#2983, #3499), so at least there is some demand.

I think https://github.com/dmlc/xgboost/issues/3499 is to be addressed in https://github.com/dmlc/xgboost/pull/4095

regarding #2983, since the release of XGBoost-GPU, how many issues are raised about GPU, and how many are about distributed GPU in Spark? I don't think that's a convincing number to trigger us take the risk of claiming it is supported in XGBoost-Spark even Spark didn't say that

CodingCat:
I would suggest, as you are working for NVIDIA, how about bringing a library based on XGBoost-Spark to support distributed GPU? as that's the major interests for NVIDIA, but as the community member, I have concern on the quality of code in the master branch

and on some day (I also hope it will happen ASAP) when Spark supports GPU better, I'd more than happy to work with you to bring that here

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/BoosterParams.scala
```diff
@@ -284,7 +286,7 @@ private[spark] object BoosterParams {
 
   val supportedBoosters = HashSet("gbtree", "gblinear", "dart")
 
-  val supportedTreeMethods = HashSet("auto", "exact", "approx", "hist")
+  val supportedTreeMethods = HashSet("auto", "exact", "approx", "hist", "gpu_hist", "gpu_exact")
```
canonizer: Most likely, `gpu_exact` doesn't support running in a distributed setting. Could you remove it from the list of supported tree methods?

srowen:
FWIW there is an actual design for GPU resource scheduling in Spark 3.0: https://issues.apache.org/jira/browse/SPARK-24615 It will probably go into 3.0 in some form. Yeah, that would be a good time to try to use GPU-aware scheduling. Anything else is a little hacky.

PR #4226, opened 2019-03-07
[xgboost4j-spark] Allow set the parameter "maxLeaves".

shishaochen:
When tree_method="hist" and grow_policy="lossguide", max_leaves is needed.


CodingCat:
sorry, I missed it , with 0.82, you can pass with paramMap like https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-example/src/main/scala/ml/dmlc/xgboost4j/scala/example/spark/SparkTraining.scala#L66-L72

can you also add to XGBoostRegressor?

shishaochen:
@CodingCat Sure! I have added "setMaxLeaves" to XGBoostRegressor in the new commit.
Acutally, xgboost4j-spark only support a subset of [all parameters](https://xgboost.readthedocs.io/en/latest/parameter.html). Then, most users may check whether a parameter is allowed by scanning setter methods in XGBoost scala wrappers.

CodingCat:
merged, thanks

PR #4232, opened 2019-03-08
Downgrade Jenkins agent to CUDA 9.2, to fix broken NCCL2 installation

hcho3:
See https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-4231/2/pipeline/51.

The current scripts downloads NCCL2 from URL `https://developer.download.nvidia.com/compute/redist/nccl/v2.2/nccl_2.2.13-1%2Bcuda${CUDA_VERSION}_x86_64.txz`. This URL works fine when `CUDA_VERSION` is 8.x or 9.x but fails for 10.1.

~~TODO. Either find a link that's not behind a login wall, OR upload installers to a private S3 bucket.~~
EDIT. I think it would be best if we place NCCL2 installers inside the worker image. This way, we don't generate unnecessary HTTP traffic.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,7 +14,7 @@ def dockerRun = 'tests/ci_build/ci_build.sh'
 def utils
 
 def buildMatrix = [
-    [ "enabled": true,  "os" : "linux", "withGpu": true, "withNccl": true,  "withOmp": true, "pythonVersion": "2.7", "cudaVersion": "10.1", "multiGpu": true],
+    [ "enabled": true,  "os" : "linux", "withGpu": true, "withNccl": true,  "withOmp": true, "pythonVersion": "2.7", "cudaVersion": "9.2", "multiGpu": true],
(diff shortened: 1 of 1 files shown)


hcho3:
I created a new issue to discuss improvements to Jenkins: #4234. I'd like to create a major refactor of Jenkins, for the sake of simplicity, de-duplication, and wider coverage.

RAMitchell:
Hmmm we can probably get the new nccl binary up soon. Suggest merging this though so we can continue to use CI, we can change it back to 10.1 later.

trivialfis:
Now that nccl is open sourced, can we compile it ourselves?  I'm doing this for 10.1 setting

rongou:
That's a very old version of nccl. Maybe upgrade to a newer version?

hcho3:
Merging for now to unblock CI. Will follow up with another PR to support 10.1 properly

PR #4238, opened 2019-03-08
Improved multi-node multi-GPU random forests.

canonizer:
- removed rabit::Broadcast() from each invocation of column sampling
- instead, syncing the PRNG seed when a ColumnSampler() object is constructed
- this makes non-trivial column sampling significantly faster in the distributed case
- refactored distributed GPU tests
- added distributed random forests tests


tests/distributed-gpu/test_gpu_basic_1x4.py
```diff
@@ -1,51 +1,15 @@
...
-                    xgb.rabit.finalize()
-                    raise Exception('Worker models diverged: test.model.1x4.{} differs from test.model.1x4.{}'.format(i, j))
-
-if (rank != 0) and (fail):
-    xgb.rabit.finalize()
-
-# Notify the tracker all training has been successful
-# This is only needed in distributed training.
-xgb.rabit.finalize()
+import distributed_gpu as dgpu
```
RAMitchell: I think these python files can be combined into a single file using command line arguments to configure the test.

canonizer: Done.

Issue #4239, opened 2019-03-08
Error memory access in gpu compressed iterator.


trivialfis:
@RAMitchell I think I might have found (one of) the cause of failing tests.  Running `cuda-memcheck` with gtest:
```
========= Invalid __global__ write of size 4
=========     at 0x00000330 in /usr/local/cuda/include/device_atomic_functions.hpp:158:void dh::LaunchNKernel<xgboost::common::WriteSymbolFunction>(unsigned long, unsigned long, xgboost::common::WriteSymbolFunction)
=========     by thread (231,0,0) in block (0,0,0)
=========     Address 0x7fe7e7c00080 is out of bounds
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame:/usr/lib/x86_64-linux-gnu/libcuda.so.1 (cuLaunchKernel + 0x2cd) [0x24d9dd]
=========     Host Frame:../xgboost/testxgboost [0xa88172]
=========     Host Frame:../xgboost/testxgboost [0xa88367]
=========     Host Frame:../xgboost/testxgboost [0xab70f5]
=========     Host Frame:../xgboost/testxgboost [0x66eaca]
=========     Host Frame:../xgboost/testxgboost [0x66e21f]
=========     Host Frame:../xgboost/testxgboost [0x66e274]
=========     Host Frame:../xgboost/testxgboost (_ZN2dh13LaunchNKernelIN7xgboost6common19WriteSymbolFunctionEEEvmmT_ + 0x27) [0x66f22e]
=========     Host Frame:../xgboost/testxgboost (_ZN2dh7LaunchNILi8ELi256EN7xgboost6common19WriteSymbolFunctionEEEvimT1_ + 0xd3) [0x66f056]
=========     Host Frame:../xgboost/testxgboost (_ZN7xgboost6common31CompressedIterator_TestGPU_Test8TestBodyEv + 0x2bb) [0x66d799]
=========     Host Frame:../xgboost/testxgboost (_ZN7testing8internal38HandleSehExceptionsInMethodIfSupportedINS_4TestEvEET0_PT_MS4_FS3_vEPKc + 0x65) [0x89737f]
=========     Host Frame:../xgboost/testxgboost (_ZN7testing8internal35HandleExceptionsInMethodIfSupportedINS_4TestEvEET0_PT_MS4_FS3_vEPKc + 0x5a) [0x891bed]
[...]

PR #4259, opened 2019-03-15
Fix multi-GPU test failures

RAMitchell:
I believe the issue was that calls to ->ConstHostVector() can modify the underlying vector, allocating or de allocating memory. Multiple threads were calling this in parallel resulting in strange memory errors.

@trivialfis I used ordinary gdb to find this as cuda-gdb seems to just crash.


trivialfis:
@RAMitchell Should we forbit calling `ExecuteIndexShards` inside another `ExecuteIndexShards`?

I will try to verify the problem is indeed solved and close #4239 later.

RAMitchell:
It may not be related to #4239. This is a good idea if possible.

trivialfis:
This is making me crazy ... My CUDA is once again broken ...

PR #4262, opened 2019-03-15
[jvm-packages] Allow supression of Rabit output in Booster::train in xgboost4j

harrybraviner:
There is currently no option to suppress the output from Rabit in xgboost4j. Redirecting `System.out` temporarilly does not help either (since the print happens in native code). This can result in a large amount of printing to the screen, resulting in other important messages being lost.

This PR changes the behavior to the following:
* Setting "silent" in this map to "true", "True", a non-zero integer, or a string that can be parsed to such an int will prevent printing.
* Setting "verbose_eval" to "False" or "false" will prevent printing.
* Setting "verbose_eval" to an int (or a String parseable to an int) `n` will result in printing every n steps, or no printing is `n` is zero.
    
This is to match the python behaviour described here: https://www.kaggle.com/c/rossmann-store-sales/discussion/17499

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
@@ -228,13 +228,56 @@ public class XGBoost {
             break;
           }
         }
-        if (Rabit.getRank() == 0) {
-          Rabit.trackerPrint(evalInfo + '\n');
+        if (Rabit.getRank() == 0 && shouldPrint(params, iter)) {
+          if (shouldPrint(params, iter)){
+            Rabit.trackerPrint(evalInfo + '\n');
+          }
         }
       }
       booster.saveRabitCheckpoint();
     }
     return booster;
+  }
+
+  static Integer tryGetIntFromObject(Object o) {
+    if (o instanceof Integer) {
+      return (int)o;
+    } else if (o instanceof String) {
+      try {
+        return Integer.parseInt((String)o);
+      } catch (NumberFormatException e) {
+        return null;
+      }
+    } else {
+      return null;
+    }
+  }
+
+  static boolean shouldPrint(Map<String, Object> params, int iter) {
+    Object silent = params.get("silent");
+    Integer silentInt = tryGetIntFromObject(silent);
+    if (silent != null) {
+      if (silent.equals("true") || silent.equals("True")
+              || (silentInt != null && silentInt != 0)) {
+        return false;  // "silent" will stop printing, otherwise go look at "verbose_eval"
+      }
+    }
+
+    Object verboseEval = params.get("verbose_eval");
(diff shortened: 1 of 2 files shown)


harrybraviner:
Note that one can test this behaviour manually by fiddling with the maps in the scala tests of xgboost4j. I'm not really sure how to add a unit test, since I don't know how to intercept the output of Rabit. I suppose we could instead test the behaviour of the `shouldPrint` method.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -228,15 +228,58 @@ public static Booster train(
...
+            Rabit.trackerPrint(evalInfo + '\n');
+          }
         }
       }
       booster.saveRabitCheckpoint();
     }
     return booster;
   }
 
+  static Integer tryGetIntFromObject(Object o) {
```
CodingCat: this should be given private accessibility

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -228,15 +228,58 @@ public static Booster train(
...
+        return Integer.parseInt((String)o);
+      } catch (NumberFormatException e) {
+        return null;
+      }
+    } else {
+      return null;
+    }
+  }
+
+  static boolean shouldPrint(Map<String, Object> params, int iter) {
```
CodingCat: this should be given private accessibility

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
harrybraviner: Agreed, I've made this change.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
harrybraviner: Agreed, I've made this change.

CodingCat:
@harrybraviner would you please rebase with the master to grab the fix for the failed R test?

harrybraviner:
@CodingCat I *think* I've done what you wanted. Let's see if the tests pass now.

I did have to do a force-push - am I correct in thinking that that's OK, but only because this is my branch and no-one else is using it (other than that this PR exists).

harrybraviner:
I'm a bit confused about why this build is failing - as you can see my diff changes nothing outside of the xgboost4j package. I'm about to successfully build and install this on my system.

hcho3:
@harrybraviner A URL fetch failed spriously. I have restarted the tests.

CodingCat:
thanks for the contribution! merged

PR #4267, opened 2019-03-17
 Deprecate `reg:linear' in favor of `reg:squarederror'.

### LINKED ISSUE #3929: Change name of objective:linear_regression
Inspired from [this issue](https://discuss.xgboost.ai/t/boosted-tree-for-regression/456) on the discussion board, right now we are calling the squared error objective linear regression which, as was the case in the linked issue is confusing to the users, especially I assume when combined with the fact that XGBoost offers a linear learner as well.

So I'm curious on the community's opinion on (eventually, after a couple of deprecating releases) changing the naming of the "reg:linear" objective to something like "reg:squarederror".

trivialfis:
close #3929.

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -28,12 +28,12 @@ NVL <- function(x, val) {
 # Merges booster params with whatever is provided in ...
 # plus runs some checks
 check.booster.params <- function(params, ...) {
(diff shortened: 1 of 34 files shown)


trivialfis:
@hcho3 Thanks~

PR #4275, opened 2019-03-18
Allow unique prediction vector for each input matrix

RAMitchell:
When training with multiple matrices, such as a test and train matrix, the preds_ member variable is resized at least twice per boosting iteration.

When the GPU is being used this resize is particularly costly, as copies between the host and device must be performed.

This PR uses std::map to provide a unique predictions vector for each dmatrix so the predictions vectors never need to be resized.

This PR results in around 10% improvement in runtime for tests/benchmark/benchmark_tree.py.

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -485,10 +485,10 @@ class LearnerImpl : public Learner {
     this->PerformTreeMethodHeuristic(train);
 
     monitor_.Start("PredictRaw");
-    this->PredictRaw(train, &preds_);
+    this->PredictRaw(train, &preds_[train]);
     monitor_.Stop("PredictRaw");
     monitor_.Start("GetGradient");
-    obj_->GetGradient(preds_, train->Info(), iter, &gpair_);
+    obj_->GetGradient(preds_[train], train->Info(), iter, &gpair_);
     monitor_.Stop("GetGradient");
     gbm_->DoBoost(train, &gpair_, obj_.get());
     monitor_.Stop("UpdateOneIter");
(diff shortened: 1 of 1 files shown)


src/learner.cc
```diff
@@ -771,7 +772,7 @@ class LearnerImpl : public Learner {
   // name of objective function
   std::string name_obj_;
   // temporal storages for prediction
```
thvasilo: Can take the chance and fix this comment as well, I assume it's meant to say "temporary"?

thvasilo:
Hello @RAMitchell, I'm curious about how this affects the overall memory usage, when you say the `preds_` variable was resized, where did this happen and how does this PR fix it?
My assumption is that the storage used by the `preds_` vector was "shared" so if we predicted on the training data first it would have size N_train, then on the test data it would be resized to N_test.

What I assume this PR does is keep both preds vectors in memory pointed to by the map?

Am I correct to assume the values in this map get populated by reference?

RAMitchell:
> Hello @RAMitchell, I'm curious about how this affects the overall memory usage, when you say the preds_ variable was resized, where did this happen and how does this PR fix it?
My assumption is that the storage used by the preds_ vector was "shared" so if we predicted on the training data first it would have size N_train, then on the test data it would be resized to N_test.

> What I assume this PR does is keep both preds vectors in memory pointed to by the map?

Yes this is correct.

> Am I correct to assume the values in this map get populated by reference?

I don't understand this question. The map entries each contain a pointer uniquely identifying each DMatrix as well as a vector for the predictions.

> @RAMitchell Is it possible to use only one prediction vector with largest size? The memory is not freed in each session so a CV search might end up using lots of memory..

It would be possible but IMO it's not even worth it. We actually already do a similar thing in [prediction cacheing](https://github.com/dmlc/xgboost/blob/359ed9c5bc8b83f6b5d6c10e130d07d4886c9fec/include/xgboost/predictor.h#L191). The extra amount of memory will be equal to 4 bytes times the number of rows in the second matrix. We already use much more memory than this for storing things like gradients (8 bytes per row), labels, weights, prediction cacheing. Not to mention memory costs proportional to m rows are typically dwarfed by the matrix itself which costs m*n.

RAMitchell:
Does anyone have further thoughts on this? @hcho3?

Again I think the impact is minimal and is consistent with an approach we have taken previously, so I would like to merge.

PR #4281, opened 2019-03-20
jenkins build for cuda 10.0

rongou:
Also removed the hack to work around the cuda 10.1 compiler bug. We can try 10.1 again once update 1 is out and the bug is fixed.

@RAMitchell @trivialfis

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,7 +14,7 @@ def dockerRun = 'tests/ci_build/ci_build.sh'
 def utils
 
 def buildMatrix = [
-    [ "enabled": true,  "os" : "linux", "withGpu": true, "withNccl": true,  "withOmp": true, "pythonVersion": "2.7", "cudaVersion": "9.2", "multiGpu": true],
+    [ "enabled": true,  "os" : "linux", "withGpu": true, "withNccl": true,  "withOmp": true, "pythonVersion": "2.7", "cudaVersion": "10.0", "multiGpu": true],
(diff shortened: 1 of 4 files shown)


RAMitchell:
The dockerfile searches for nccl binaries here (https://github.com/dmlc/xgboost/blob/09bd9e68cf369e759052e4aae2cce6871cb996fd/tests/ci_build/Dockerfile.gpu#L24).

I'm not sure if 10.0 binaries are available at this link. We may have to move back to a nccl submodule to have a more sustainable way of building multi-GPU xgboost.

rongou:
OK, I guess Jenkins doesn't actually use the `Jenkinsfile` in the PR?

Is there a reason to stick with CUDA 8 and Centos 6? If we drop CUDA 8 and move to Centos 7 there would be more options.

RAMitchell:
I think updates to the Jenkinsfile do not immediately take effect due to security reasons. Possibly it works for members, I'm not sure.

I set up the builds on Centos 6 and cuda 8.0 for maximum compatibility binary wheels (https://opensource.com/article/19/2/manylinux-python-wheels). Potentially we could upgrade cuda soon. What particular features do you need?

hcho3:
@RAMitchell I changed the security configuration for Jenkinsfile, so that changes are visible. Let me try again.

rongou:
Looks like Jenkins has passed with CUDA 10.0? Please take another look. @RAMitchell @hcho3

hcho3:
LGTM

PR #4283, opened 2019-03-20
Further optimisations for gpu_hist.

RAMitchell:
- Fuse final update position functions into a single more efficient kernel

- Refactor gpu_hist with a more explicit ellpack matrix representation

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -521,7 +521,7 @@ class LearnerImpl : public Learner {
     }
     for (size_t i = 0; i < data_sets.size(); ++i) {
       DMatrix * dmat = data_sets[i];
-      this->PredictRaw(dmat, &preds_[dmat]);
+      this->PredictRaw(data_sets[i], &preds_[dmat]);
       obj_->EvalTransform(&preds_[dmat]);
       for (auto& ev : metrics_) {
         os << '\t' << data_names[i] << '-' << ev->Name() << ':'
@@ -771,8 +771,7 @@ class LearnerImpl : public Learner {
   std::string name_gbm_;
   // name of objective function
   std::string name_obj_;
-  // temporal storages for prediction, store a unique prediction vector for each
-  // matrix to avoid constant resizing
+  // temporal storages for prediction
   std::map<DMatrix*, HostDeviceVector<bst_float>> preds_;
   // gradient pairs
   HostDeviceVector<GradientPair> gpair_;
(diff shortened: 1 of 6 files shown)


RAMitchell:
A quick review of my optimisation work over the last month. My improvements have been:
- Combine or remove small host/device transfers
- Use streams to overlap kernels in split evaluation
- Use streams to overlap memory transfers in position updating
- Fuse final position updating kernels into a single kernel
- Find and remove unnecessary host/device transfers between boosting iterations

On a 10M*100 dense input matrix, boosting for 500 iterations, the performance improvement is approximately:
1 gpu: 28.2%
2 gpus: 33.5%
8 gpus: 40.9%

In particular multi-GPU scalability seems to have improved considerably.

src/learner.cc
```diff
@@ -485,10 +485,10 @@ class LearnerImpl : public Learner {
     this->PerformTreeMethodHeuristic(train);
 
     monitor_.Start("PredictRaw");
-    this->PredictRaw(train, &preds_);
+    this->PredictRaw(train, &preds_[train]);
```
canonizer: Given that DMatrix pointers are increasingly used as cache indices, the parameter should probably be changed from `DMatrix*` to `shared_ptr<DMatrix>` in all those places. We can then use `weak_ptr<DMatrix>` as the index into the cache.

This can be done in another pull request, however.

src/tree/updater_gpu_hist.cu
```diff
@@ -397,27 +488,25 @@ __global__ void CompressBinEllpackKernel(
...
-                                    common::CompressedIterator<uint32_t> d_gidx,
-                                    int null_gidx_value,
+__global__ void SharedMemHistKernel(ELLPackMatrix matrix, const bst_uint* d_ridx,
                                     GradientSumT* d_node_hist,
                                     const GradientPair* d_gpair,
                                     size_t segment_begin, size_t n_elements) {
   extern __shared__ char smem[];
   GradientSumT* smem_arr = reinterpret_cast<GradientSumT*>(smem); // NOLINT
-  for (auto i : dh::BlockStrideRange(0, null_gidx_value)) {
+  for (auto i : dh::BlockStrideRange(0, matrix.null_gidx_value)) {
```
canonizer: Could you add a function like `matrix.BinCount()`, just to make the code more readable? `null_gidx_value` can then be used in cases where it means 'no value'

PR #4284, opened 2019-03-21
initial support for external memory in gpu_predictor

rongou:
This seems to work on the mushroom dataset using a single GPU (I changed the page size to 4KB to get multiple pages). The batch objects seem to be reused, causing the `CHECK` to fail.

@canonizer  @RAMitchell

### DIFF
--- a/src/common/host_device_vector.cu
+++ b/src/common/host_device_vector.cu
@@ -601,7 +601,7 @@ void HostDeviceVector<T>::Shard(const GPUDistribution &distribution) const {
 }
 
 template <typename T>
-void HostDeviceVector<T>::Reshard(const GPUDistribution &distribution) {
+void HostDeviceVector<T>::Reshard(const GPUDistribution &distribution) const {
   impl_->Reshard(distribution);
 }
 
--- a/src/common/host_device_vector.cu
+++ b/src/common/host_device_vector.cu
@@ -601,7 +601,7 @@ void HostDeviceVector<T>::Shard(const GPUDistribution &distribution) const {
 }
 
 template <typename T>
-void HostDeviceVector<T>::Reshard(const GPUDistribution &distribution) const {
+void HostDeviceVector<T>::Reshard(const GPUDistribution &distribution) {
   impl_->Reshard(distribution);
 }
 
--- a/src/data/sparse_page_source.cc
+++ b/src/data/sparse_page_source.cc
@@ -104,6 +104,10 @@ void SparsePageSource::BeforeFirst() {
   }
 }
 
+SparsePage& SparsePageSource::Value() {
+  return *page_;
+}
+
 const SparsePage& SparsePageSource::Value() const {
   return *page_;
 }
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -329,7 +329,7 @@ class GPUPredictor : public xgboost::Predictor {
     for (const auto &batch : dmat->GetRowBatches()) {
       CHECK(i_batch == 0 || devices_.Size() == 1) << "External memory not supported for multi-GPU";
       // out_preds have been sharded and resized in InitOutPredictions()
-      batch.offset.Reshard(GPUDistribution::Overlap(devices_, 1));
+      batch.offset.Shard(GPUDistribution::Overlap(devices_, 1));
       std::vector<size_t> device_offsets;
       DeviceOffsets(batch.offset, &device_offsets);
       batch.data.Reshard(GPUDistribution::Explicit(devices_, device_offsets));
@@ -337,8 +337,7 @@ class GPUPredictor : public xgboost::Predictor {
         shard.PredictInternal(batch, dmat->Info(), out_preds, batch_offset, model,
                               h_tree_segments, h_nodes, tree_begin, tree_end);
       });
-      size_t num_rows = batch.Size();
-      batch_offset += num_rows * model.param.num_output_group;
+      batch_offset += batch.Size() * model.param.num_output_group;
       i_batch++;
     }
     monitor_.StopCuda("DevicePredictInternal");
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -326,7 +326,7 @@ class GPUPredictor : public xgboost::Predictor {
 
     size_t i_batch = 0;
     size_t batch_offset = 0;
-    for (const auto &batch : dmat->GetRowBatches()) {
+    for (auto &batch : dmat->GetRowBatches()) {
       CHECK(i_batch == 0 || devices_.Size() == 1) << "External memory not supported for multi-GPU";
       // out_preds have been sharded and resized in InitOutPredictions()
       batch.offset.Shard(GPUDistribution::Overlap(devices_, 1));
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -246,6 +246,7 @@ class GPUPredictor : public xgboost::Predictor {
     void PredictInternal
     (const SparsePage& batch, const MetaInfo& info,
      HostDeviceVector<bst_float>* predictions,
+     const size_t batch_offset,
      const gbm::GBTreeModel& model,
      const thrust::host_vector<size_t>& h_tree_segments,
      const thrust::host_vector<DevicePredictionNode>& h_nodes,
@@ -284,8 +285,8 @@ class GPUPredictor : public xgboost::Predictor {
                                                  data_distr.Devices().Index(device_));
 
       PredictKernel<BLOCK_THREADS><<<GRID_SIZE, BLOCK_THREADS, shared_memory_bytes>>>
-        (dh::ToSpan(nodes_), predictions->DeviceSpan(device_), dh::ToSpan(tree_segments_),
-         dh::ToSpan(tree_group_), batch.offset.DeviceSpan(device_),
+        (dh::ToSpan(nodes_), predictions->DeviceSpan(device_).subspan(batch_offset),
+         dh::ToSpan(tree_segments_), dh::ToSpan(tree_group_), batch.offset.DeviceSpan(device_),
          batch.data.DeviceSpan(device_), tree_begin, tree_end, info.num_col_,
          num_rows, entry_start, use_shared, model.param.num_output_group);
     }
@@ -324,18 +325,20 @@ class GPUPredictor : public xgboost::Predictor {
     }
 
     size_t i_batch = 0;
-
+    size_t batch_offset = 0;
     for (const auto &batch : dmat->GetRowBatches()) {
-      CHECK_EQ(i_batch, 0) << "External memory not supported";
+      CHECK(i_batch == 0 || devices_.Size() == 1) << "External memory not supported for multi-GPU";
       // out_preds have been sharded and resized in InitOutPredictions()
-      batch.offset.Shard(GPUDistribution::Overlap(devices_, 1));
+      batch.offset.Reshard(GPUDistribution::Overlap(devices_, 1));
       std::vector<size_t> device_offsets;
       DeviceOffsets(batch.offset, &device_offsets);
-      batch.data.Shard(GPUDistribution::Explicit(devices_, device_offsets));
+      batch.data.Reshard(GPUDistribution::Explicit(devices_, device_offsets));
       dh::ExecuteIndexShards(&shards_, [&](int idx, DeviceShard& shard) {
-        shard.PredictInternal(batch, dmat->Info(), out_preds, model,
+        shard.PredictInternal(batch, dmat->Info(), out_preds, batch_offset, model,
                               h_tree_segments, h_nodes, tree_begin, tree_end);
       });
+      size_t num_rows = batch.Size();
+      batch_offset += num_rows * model.param.num_output_group;
       i_batch++;
     }
     monitor_.StopCuda("DevicePredictInternal");
--- a/tests/cpp/helpers.h
+++ b/tests/cpp/helpers.h
@@ -154,7 +154,7 @@ class SimpleRealUniformDistribution {
 std::shared_ptr<xgboost::DMatrix> *CreateDMatrix(int rows, int columns,
                                                  float sparsity, int seed = 0);
 
-std::unique_ptr<DMatrix> CreateSparsePageDMatrix();
+std::unique_ptr<DMatrix> CreateSparsePageDMatrix(size_t n_entries, size_t page_size);
 
 gbm::GBTreeModel CreateTestModel();
 
--- a/include/xgboost/data.h
+++ b/include/xgboost/data.h
@@ -284,6 +284,7 @@ class BatchIteratorImpl {
  public:
   virtual ~BatchIteratorImpl() {}
   virtual BatchIteratorImpl* Clone() = 0;
+  virtual SparsePage& operator*() = 0;
   virtual const SparsePage& operator*() const = 0;
   virtual void operator++() = 0;
   virtual bool AtEnd() const = 0;
@@ -305,6 +306,11 @@ class BatchIterator {
   void operator++() {
     CHECK(impl_ != nullptr);
     ++(*impl_);
+  }
+
+  SparsePage& operator*() {
+    CHECK(impl_ != nullptr);
+    return *(*impl_);
   }
 
   const SparsePage& operator*() const {
--- a/src/common/host_device_vector.h
+++ b/src/common/host_device_vector.h
@@ -259,7 +259,7 @@ class HostDeviceVector {
   /*!
    * \brief Change memory distribution.
    */
-  void Reshard(const GPUDistribution &distribution) const;
+  void Reshard(const GPUDistribution &distribution);
 
   void Resize(size_t new_size, T v = T());
 
--- a/src/common/host_device_vector.h
+++ b/src/common/host_device_vector.h
@@ -259,7 +259,7 @@ class HostDeviceVector {
   /*!
    * \brief Change memory distribution.
    */
-  void Reshard(const GPUDistribution &distribution);
+  void Reshard(const GPUDistribution &distribution) const;
 
   void Resize(size_t new_size, T v = T());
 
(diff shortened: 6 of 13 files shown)


src/common/host_device_vector.cu
```diff
@@ -361,7 +361,6 @@ struct HostDeviceVectorImpl {
 
   void Reshard(const GPUDistribution& distribution) {
     if (distribution_ == distribution) { return; }
-    CHECK(distribution_.IsEmpty() || distribution.IsEmpty());
```
canonizer: This check should remain.

RAMitchell:
Looks good in general. If we are explicitly supporting this feature there must be a test to enforce that it continues to work.

src/common/host_device_vector.cu
rongou: Done.

rongou:
@RAMitchell @canonizer @sriramch Tests are passing now. Please take another look.

trivialfis:
@rongou multi-gpu tests didn't pass.  ;-)

sriramch:
isn't the fundamental issue here that the accumulated prediction vector (over multiple batches) have been created once and sharded based on the *total* number rows in the dataset as opposed to the ones  in the individual batches - like [here](https://github.com/dmlc/xgboost/blob/a11f8ca44690b112527417c1204e2f854216cff3/src/predictor/gpu_predictor.cu#L377)

if so, wouldn't this work? basically, we create and compute batch predictions per page on the gpu and then accumulate them into the global prediction vector at the end of each batch.

```
diff --git a/src/predictor/gpu_predictor.cu b/src/predictor/gpu_predictor.cu
index 39b5c87..7f56543 100644
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -178,7 +178,6 @@ __device__ float GetLeafWeight(bst_uint ridx, const DevicePredictionNode* tree,
   return n.GetWeight();
 }
 
-template <int BLOCK_THREADS>
 __global__ void PredictKernel(common::Span<const DevicePredictionNode> d_nodes,
                               common::Span<float> d_out_predictions,
                               common::Span<size_t> d_tree_segments,
@@ -283,8 +282,10 @@ class GPUPredictor : public xgboost::Predictor {
       size_t entry_start = data_distr.ShardStart(batch.data.Size(),
                                                  data_distr.Devices().Index(device_));
 
-      PredictKernel<BLOCK_THREADS><<<GRID_SIZE, BLOCK_THREADS, shared_memory_bytes>>>
-        (dh::ToSpan(nodes_), predictions->DeviceSpan(device_), dh::ToSpan(tree_segments_),
+      PredictKernel<<<GRID_SIZE, BLOCK_THREADS, shared_memory_bytes>>>
+        (dh::ToSpan(nodes_),
+         predictions->DeviceSpan(device_),
+         dh::ToSpan(tree_segments_),
          dh::ToSpan(tree_group_), batch.offset.DeviceSpan(device_),
          batch.data.DeviceSpan(device_), tree_begin, tree_end, info.num_col_,
          num_rows, entry_start, use_shared, model.param.num_output_group);
@@ -323,20 +324,47 @@ class GPUPredictor : public xgboost::Predictor {
                 h_nodes.begin() + h_tree_segments[tree_idx - tree_begin]);
     }
 
-    size_t i_batch = 0;
+    // Accumulate all predictions from multiple batches into this
+    std::vector<float> &hvec = out_preds->HostVector();
+    std::vector<float>::iterator hitr = hvec.begin();
 
     for (const auto &batch : dmat->GetRowBatches()) {
-      CHECK_EQ(i_batch, 0) << "External memory not supported";
       // out_preds have been resharded and resized in InitOutPredictions()
       batch.offset.Reshard(GPUDistribution::Overlap(devices_, 1));
       std::vector<size_t> device_offsets;
       DeviceOffsets(batch.offset, &device_offsets);
       batch.data.Reshard(GPUDistribution::Explicit(devices_, device_offsets));
+
+      // Gather the batch predictions; but, first reset the number of rows
+      // to what is present in the batch. The local batch prediction should
+      // be resharded based on the number of devices present already
+      struct RowReset {
+         public:
+            RowReset(MetaInfo &minfo, const SparsePage &r_batch)
+               : minfo_(minfo) {
+               orig_nrows_ = minfo_.num_row_;
+               minfo.num_row_ = r_batch.Size();
+            }
+                                                    
+            ~RowReset() {
+               minfo_.num_row_ = orig_nrows_;
+            }
+         private:
+            MetaInfo &minfo_;
+            size_t orig_nrows_;
+      }rowgd(dmat->Info(), batch);                                      
+
+      HostDeviceVector<bst_float> batchPreds;
+      this->InitOutPredictions(dmat->Info(), &batchPreds, model);
+
       dh::ExecuteIndexShards(&shards_, [&](int idx, DeviceShard& shard) {
-        shard.PredictInternal(batch, dmat->Info(), out_preds, model,
+        shard.PredictInternal(batch, dmat->Info(), &batchPreds, model,
                               h_tree_segments, h_nodes, tree_begin, tree_end);
       });
-      i_batch++;
+
+      // Copy the batch predictions to the global prediction array
+      std::copy(batchPreds.ConstHostVector().begin(), batchPreds.ConstHostVector().end(), hitr);
+      std::advance(hitr, batchPreds.Size());
     }
     monitor_.StopCuda("DevicePredictInternal");
   }
```

i haven't run any unit tests to see if this will fly; but some local testing on a dataset works across single and multiple gpus.

rongou:
@trivialfis thanks, I can reproduce the error now in the unit test by increasing the dmatrix size. Working on a fix.

@sriramch yes same goal different approaches.

rongou:
Hmm, the multi-gpu external memory test fails on Jenkins but passes on my dev instance. Need to investigate more.

src/predictor/gpu_predictor.cu
```diff
@@ -323,20 +324,26 @@ class GPUPredictor : public xgboost::Predictor {
...
                               h_tree_segments, h_nodes, tree_begin, tree_end);
       });
-      i_batch++;
+      for (int i = 0; i < devices_.Size(); i++) {
+        if (batch.offset.Devices().Contains(i)) {
+          size_t num_rows = batch.offset.DeviceSize(i) - 1;
+          batch_offsets[i] += num_rows * model.param.num_output_group;
+        }
+      }
+      batch.offset.Reshard(empty_distribution);
```
sriramch: i'm wondering why we need this. is it to make [this](https://github.com/dmlc/xgboost/blob/master/src/common/host_device_vector.cu#L363) check happy?

more fundamentally, why is that check present? is it not plausible to first shard a device vector based on a gpu distribution config, only to later reshard with another distribution with a different configuration? for my own edification, what is forcing one of the distribution to be always non-empty?

trivialfis: @sriramch I think it was added to prevent accidental data movement.

sriramch: can you please expand? wouldn't reshard recreate a new distribution (reinitialize the shards) based on the new distribution config, thus effectively wiping out the old one? when would there be a `data movement`?

trivialfis: If memory serves I think it was added by @canonizer .  And it's to make sure if one want to move data from one device to another, or to gather/scatter , it has to be done explicitly.  We might be doing this accidentally due to wrong parameters, data granularity etc.

rongou:
Hmm, the problem with the current approach is if batches cannot be divided evenly between multiple GPUs, then we cannot rely on the existing sharding. As @canonizer suggested, we need a new block cyclic gpu distribution. Will do that in a separate PR and come back to this later.

rongou:
@canonizer suggested limiting the external memory support for single gpu only for now.

@canonizer @RAMitchell @sriramch PTAL.

sriramch:
@rongou here is what i was referring to. please try this when you get a chance:

```
diff --git a/src/common/host_device_vector.h b/src/common/host_device_vector.h
index 425cbff..eda35fc 100644
--- a/src/common/host_device_vector.h
+++ b/src/common/host_device_vector.h
@@ -132,7 +132,8 @@ class GPUDistribution {
     if (size == 0) { return 0; }
     if (offsets_.size() > 0) {
       // explicit offsets are provided
-      CHECK_EQ(offsets_.back(), size);
+      //if (index >= offsets_.size()) return 0;
+      //CHECK_EQ(offsets_.back(), size);
       return offsets_.at(index);
     }
     // no explicit offsets
@@ -145,7 +146,9 @@ class GPUDistribution {
     if (size == 0) { return 0; }
     if (offsets_.size() > 0) {
       // explicit offsets are provided
-      CHECK_EQ(offsets_.back(), size);
+      //if (index >= offsets_.size()-1) return 0;
+      //CHECK_EQ(offsets_.back(), size);
+      if (!offsets_.at(index+1)) return 0;
       return offsets_.at(index + 1)  - offsets_.at(index) +
         (index == devices_.Size() - 1 ? overlap_ : 0);
     }
diff --git a/src/predictor/gpu_predictor.cu b/src/predictor/gpu_predictor.cu
index 39b5c87..98c3868 100644
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -178,7 +178,6 @@ __device__ float GetLeafWeight(bst_uint ridx, const DevicePredictionNode* tree,
   return n.GetWeight();
 }
 
-template <int BLOCK_THREADS>
 __global__ void PredictKernel(common::Span<const DevicePredictionNode> d_nodes,
                               common::Span<float> d_out_predictions,
                               common::Span<size_t> d_tree_segments,
@@ -231,9 +230,16 @@ class GPUPredictor : public xgboost::Predictor {
       auto data_span = data.DeviceSpan(device);
       dh::safe_cuda(cudaSetDevice(device));
       // copy the last element from every shard
+      if (!data_span.size()) {
+      size_t zero_size = 0;
+      dh::safe_cuda(cudaMemcpy(&offsets.at(shard + 1),
+                               &zero_size,
+                               sizeof(size_t), cudaMemcpyHostToHost));
+      } else {
       dh::safe_cuda(cudaMemcpy(&offsets.at(shard + 1),
                                &data_span[data_span.size()-1],
                                sizeof(size_t), cudaMemcpyDeviceToHost));
+      }
     }
   }
 
@@ -283,8 +289,10 @@ class GPUPredictor : public xgboost::Predictor {
       size_t entry_start = data_distr.ShardStart(batch.data.Size(),
                                                  data_distr.Devices().Index(device_));
 
-      PredictKernel<BLOCK_THREADS><<<GRID_SIZE, BLOCK_THREADS, shared_memory_bytes>>>
-        (dh::ToSpan(nodes_), predictions->DeviceSpan(device_), dh::ToSpan(tree_segments_),
+      PredictKernel<<<GRID_SIZE, BLOCK_THREADS, shared_memory_bytes>>>
+        (dh::ToSpan(nodes_),
+         predictions->DeviceSpan(device_),
+         dh::ToSpan(tree_segments_),
          dh::ToSpan(tree_group_), batch.offset.DeviceSpan(device_),
          batch.data.DeviceSpan(device_), tree_begin, tree_end, info.num_col_,
          num_rows, entry_start, use_shared, model.param.num_output_group);
@@ -323,20 +331,52 @@ class GPUPredictor : public xgboost::Predictor {
                 h_nodes.begin() + h_tree_segments[tree_idx - tree_begin]);
     }
 
-    size_t i_batch = 0;
+    // Accumulate all predictions from multiple batches into this
+    std::vector<float> &hvec = out_preds->HostVector();
+    std::vector<float>::iterator hitr = hvec.begin();
 
     for (const auto &batch : dmat->GetRowBatches()) {
-      CHECK_EQ(i_batch, 0) << "External memory not supported";
       // out_preds have been resharded and resized in InitOutPredictions()
       batch.offset.Reshard(GPUDistribution::Overlap(devices_, 1));
+
       std::vector<size_t> device_offsets;
       DeviceOffsets(batch.offset, &device_offsets);
+
       batch.data.Reshard(GPUDistribution::Explicit(devices_, device_offsets));
+
+      // Gather the batch predictions; but, first reset the number of rows
+      // to what is present in the batch. The local batch prediction should
+      // be resharded based on the number of devices present already
+      struct RowReset {
+         public:
+            RowReset(MetaInfo &minfo, const SparsePage &r_batch)
+               : minfo_(minfo) {
+               orig_nrows_ = minfo_.num_row_;
+               minfo.num_row_ = r_batch.Size();
+            }
+                                                    
+            ~RowReset() {
+               minfo_.num_row_ = orig_nrows_;
+            }
+         private:
+            MetaInfo &minfo_;
+            size_t orig_nrows_;
+      }rowgd(dmat->Info(), batch);                                      
+
+      HostDeviceVector<bst_float> batchPreds;
+      this->InitOutPredictions(dmat->Info(), &batchPreds, model);
+      // Starting from hitr copy batch size * model.param.num_output_group
+      // that were previously present into batchPreds
+      std::copy(hitr, hitr + (batch.Size() * model.param.num_output_group), batchPreds.HostVector().begin());
+
       dh::ExecuteIndexShards(&shards_, [&](int idx, DeviceShard& shard) {
-        shard.PredictInternal(batch, dmat->Info(), out_preds, model,
+        shard.PredictInternal(batch, dmat->Info(), &batchPreds, model,
                               h_tree_segments, h_nodes, tree_begin, tree_end);
       });
-      i_batch++;
+
+      // Copy the batch predictions to the global prediction array
+      std::copy(batchPreds.ConstHostVector().begin(), batchPreds.ConstHostVector().end(), hitr);
+      std::advance(hitr, batchPreds.Size());
     }
     monitor_.StopCuda("DevicePredictInternal");
   }
```

sriramch:
please refer to the multi gpu external memory predictor fix [here](https://gist.github.com/sriramch/72986f49b74f379282f5d13776f17cc6)

rongou:
@canonizer @RAMitchell @hcho3 I think this is in a reasonable shape now, please take another look. We'll send a followup PR to enable multi-gpu external memory.

src/data/sparse_page_source.cc
```diff
@@ -67,9 +67,10 @@ SparsePageSource::SparsePageSource(const std::string& cache_info,
     size_t fbegin = fi->Tell();
     prefetchers_[i].reset(new dmlc::ThreadedIter<SparsePage>(4));
     prefetchers_[i]->Init([&fi, &fmt] (SparsePage** dptr) {
-        if (*dptr == nullptr) {
-          *dptr = new SparsePage();
+        if (*dptr != nullptr) {
+          delete *dptr;
```
canonizer: Any particular reason for this?

tests/cpp/helpers.h
```diff
@@ -154,7 +154,8 @@ class SimpleRealUniformDistribution {
 std::shared_ptr<xgboost::DMatrix> *CreateDMatrix(int rows, int columns,
                                                  float sparsity, int seed = 0);
 
-std::unique_ptr<DMatrix> CreateSparsePageDMatrix();
+std::unique_ptr<DMatrix> CreateSparsePageDMatrix(const size_t n_entries=12,
```
canonizer: Is it possible to drop the default values?

Also, no need for `const size_t` in parameters, just `size_t` is fine.

rongou: Done.

src/data/sparse_page_source.cc
rongou: If `SparsePage`s are reused, `data` in the last page might require different sharding because the offsets are different. The current `HostDeviceVector::Reshard()` method doesn't allow this because it assumes sharding only happens once. Since we don't really care about the device side data once we are done with a `SparsePage`, it seems simpler to just delete it.

Alternatively we could rename the current `Reshard` method to `Shard`, and add a new method `Reshard` to explicitly reshard an already sharded `HostDeviceVector`. Let me know if you prefer this.

rongou: BTW I created #4396 if we want to take the second approach.

RAMitchell: This approach looks much easier. There may be a performance to the current CPU code however. If the batch size is consistent the old implementation reuses memory in the SparsePage object when calling fmt->Read(). If you can do an experiment to show that the impact is negligible this approach is strongly preferred.

hcho3:
Rebasing against the latest master to fix tests

hcho3:
@rongou @RAMitchell @canonizer What's the status of this pull request? Is it ready to merge?

rongou:
@hcho3 we found that the extra memory allocations cause significant slowdowns in CPU code, so adding the resharding method (#4396) so that we can reuse the `SparsePage`s. I'll mark this as WIP for now and wait for the other PR.

hcho3:
@rongou Now that #4396 is merged, can we consider this PR once again?

src/data/sparse_page_source.cc
rongou: No longer need this.

rongou:
@hcho3 yes. @canonizer @RAMitchell please take another look.

src/predictor/gpu_predictor.cu
```diff
@@ -324,18 +325,20 @@ class GPUPredictor : public xgboost::Predictor {
...
 
     size_t i_batch = 0;
-
+    size_t batch_offset = 0;
     for (const auto &batch : dmat->GetRowBatches()) {
-      CHECK_EQ(i_batch, 0) << "External memory not supported";
+      CHECK(i_batch == 0 || devices_.Size() == 1) << "External memory not supported for multi-GPU";
       // out_preds have been sharded and resized in InitOutPredictions()
-      batch.offset.Shard(GPUDistribution::Overlap(devices_, 1));
+      batch.offset.Reshard(GPUDistribution::Overlap(devices_, 1));
```
canonizer: Why do we need `Reshard()` here? Won't the distribution be the same as long as `devices_` stays the same?

src/predictor/gpu_predictor.cu
```diff
@@ -324,18 +325,20 @@ class GPUPredictor : public xgboost::Predictor {
...
       std::vector<size_t> device_offsets;
       DeviceOffsets(batch.offset, &device_offsets);
-      batch.data.Shard(GPUDistribution::Explicit(devices_, device_offsets));
+      batch.data.Reshard(GPUDistribution::Explicit(devices_, device_offsets));
       dh::ExecuteIndexShards(&shards_, [&](int idx, DeviceShard& shard) {
-        shard.PredictInternal(batch, dmat->Info(), out_preds, model,
+        shard.PredictInternal(batch, dmat->Info(), out_preds, batch_offset, model,
                               h_tree_segments, h_nodes, tree_begin, tree_end);
       });
+      size_t num_rows = batch.Size();
```
canonizer: `num_rows` is used only once. Consider replacing it with a call to `batch.Size()`.

src/predictor/gpu_predictor.cu

src/predictor/gpu_predictor.cu

hcho3:
Nice work, everyone!

PR #4293, opened 2019-03-25
Retire DVec class in favour of c++20 style span for device memory.

RAMitchell:
Both classes serve the same purpose - a non-owning memory container. Span is more generic and better designed.

### DIFF
--- a/src/linear/updater_gpu_coordinate.cu
+++ b/src/linear/updater_gpu_coordinate.cu
@@ -19,18 +19,18 @@ namespace linear {
 
 DMLC_REGISTRY_FILE_TAG(updater_gpu_coordinate);
 
-void RescaleIndices(size_t ridx_begin, dh::DVec<xgboost::Entry> *data) {
-  auto d_data = data->GetSpan();
-  dh::LaunchN(data->DeviceIdx(), data->Size(),
-              [=] __device__(size_t idx) { d_data[idx].index -= ridx_begin; });
+void RescaleIndices(int device_idx, size_t ridx_begin,
+                    common::Span<xgboost::Entry> data) {
+  dh::LaunchN(device_idx, data.size(),
+              [=] __device__(size_t idx) { data[idx].index -= ridx_begin; });
 }
 
 class DeviceShard {
   int device_id_;
-  dh::BulkAllocator<dh::MemoryType::kDevice> ba_;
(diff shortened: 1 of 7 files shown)


trivialfis:
@RAMitchell I haven't looked this in detail yet.  Glancing through the changes it seems to make sure we access the right pointer for each device there are two conditions:

- You only use it in `DeviceShard` which can keep device id as part of it.
- Remember calling `cudaSetDevice` before doing anything.

I have been reluctant to make the change since the above two conditions are hard to guarantee, and the bugs are hard to fix ... Could you explain why do you think it's solved in this PR? I'm just want to feel a little bit assurance.

RAMitchell:
I think the above problem still exists with the DVec class - we were not calling cudaSetDevice() on every method call of DVec even when we did have the device index information available. 

This problem of device selection might be easier to approach by enforcing that each device shard has its respective device selected every time it calls a method, instead of trying to explicitly associate memory with a device.

src/linear/updater_gpu_coordinate.cu
```diff
@@ -87,8 +87,10 @@ class DeviceShard {
 
   void UpdateGpair(const std::vector<GradientPair> &host_gpair,
                    const gbm::GBLinearModelParam &model_param) {
-    gpair_.copy(host_gpair.begin() + ridx_begin_ * model_param.num_output_group,
-                host_gpair.begin() + ridx_end_ * model_param.num_output_group);
+    dh::safe_cuda(cudaMemcpy(
```
trivialfis: async?

src/tree/updater_gpu_hist.cu
```diff
@@ -727,13 +728,16 @@ struct DeviceShard {
       dh::safe_cuda(cudaMemsetAsync(left_counts.data().get(), 0,
                                     sizeof(int64_t) * left_counts.size()));
     }
-    thrust::sequence(ridx.CurrentDVec().tbegin(), ridx.CurrentDVec().tend());
+    thrust::sequence(
+        thrust::device_pointer_cast(ridx.CurrentSpan().data()),
```
trivialfis: I thought any iterator would do,  why creating a wrapper?

RAMitchell: If you pass a raw pointer thrust thinks the memory is on the host and crashes.

Issue #3720, opened 2018-09-24
Roadmap: Fix flaky tests


hcho3:
This issue is aimed at keeping track of flaky tests.

Rationale: some of existing tests tend to fail randomly, even when the pull request / branch update has nothing do to the tests. **Since flaky tests bewilder community contributors and hamper developer productivity, they should be quickly addressed.** 

**Note to maintainers: If you spot a flaky test, please report it here.**

**List of existing flaky tests**
- [x] [R-package/tests/testthat/test_update.R](https://github.com/dmlc/xgboost/blob/79d854c695aa5d32c022a2372902597fead1a23a/R-package/tests/testthat/test_update.R): Test that model updates work.
  * Lines [32](https://github.com/dmlc/xgboost/blob/79d854c695aa5d32c022a2372902597fead1a23a/R-package/tests/testthat/test_update.R#L32), [41](https://github.com/dmlc/xgboost/blob/79d854c695aa5d32c022a2372902597fead1a23a/R-package/tests/testthat/test_update.R#L41) and [64](https://github.com/dmlc/xgboost/blob/79d854c695aa5d32c022a2372902597fead1a23a/R-package/tests/testthat/test_update.R#L64) randomly fail for 32-bit Windows + MinGW target.
  * Build log (AppVeyor): https://ci.appveyor.com/project/tqchen/xgboost/build/1.0.2949/job/7i0d2lim73hhy44f#L2960, https://ci.appveyor.com/project/tqchen/xgboost/build/1.0.2954/job/l2icjm6yq4kc2k3s#L3093
- [ ] [tests/python/regression_test_utilities.py](https://github.com/dmlc/xgboost/blob/58d783df16d54361a50b98e211e98a36d4d876ac/tests/python/regression_test_utilities.py#L130-L153): Test suite
  * Randomly throws `std::bad_alloc` exception due to memory shortage
  * Build log (Travis CI): https://travis-ci.org/dmlc/xgboost/jobs/427691975#L2250

khotilov:
I could not reproduce the 1st one, but I will need to try with this specific version R-3.5.1patched later.

hcho3:
@khotilov I think the issue only appears for 32-bit R.

hcho3:
For tests/python/regression_test_utilities.py, we could solve it by migrating Python test to Jenkins, where we can use slave workers with more memory.

khotilov:
[...]

PR #4301, opened 2019-03-27
- make test fails to find testxgboost, if a build directory is created

sriramch:
on which cmake is run. this appends the build directory to the test binary
- make the assignments of HostDeviceVector exception safe
- rm storing a dummy GPUDistribution instance for CPU based code

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -340,7 +340,7 @@ if(GOOGLE_TEST)
       $<TARGET_OBJECTS:objxgboost>)
   endif ()
 
-  set_output_directory(testxgboost ${PROJECT_SOURCE_DIR})
+  set_output_directory(testxgboost ${CMAKE_BINARY_DIR})
   target_include_directories(testxgboost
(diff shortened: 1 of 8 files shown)


rongou:
@RAMitchell

sriramch:
#3720 reference to bad_alloc

sriramch:
> @sriramch Your assignment operator is not exception safe since `new` can throw. https://stackoverflow.com/questions/3279543/what-is-the-copy-and-swap-idiom might help a little. But I would leave it to another PR. ;-)

@trivialfis what this patch is attempting to do is that when a target hostvectorimpl (other) is assigned to a source (*this), any intermediary exception that *may* occur during this assignment does not alter the initial state of the source. it is *not* trying to make the assignment operator not throw any exceptions at all.  this entails simply doing the copy work on the side and when that operation is deemed successful, simply swaps the guts of it with the source. if the side work fails ofr some reason (say due to new throwing as you state), source object (*this) isn't impacted in any way

Issue #4162, opened 2019-02-19
Enforce having more threads than GPUs.


trivialfis:
Brought up in #4076 .  Having less threads than GPUs will lead to undefined behaviors.  Another example is a hang I encountered when working on #3974.  #4095 might be able to decouple the number of threads and number of GPUs.  But until then, I want to make a last minute fix that when user specified `nthread < n_gpus`, we either decrease `n_gpus` to `nthreads` or increase `nthreads` to `n_gpus`.  @RAMitchell WDYT?

mt-jones:
In an effort to provide some additional insight:

Currently https://github.com/dmlc/xgboost/pull/4095 creates two modes of operation:
1. A NCCL communicator clique where a single process manages a collection of GPUs
2. A NCCL communicator clique where there is a single process (PID) for each GPU involved in the clique

(2) is triggered by `rabit::GetWorldSize()`. Meaning, if within a node you assign an MPI rank per physical GPU, you will create a process for each GPU.

(1) is triggered by calling an XGBoost process with OpenMP threads, rather than MPI ranks.

In short, it is possible to adjust the code so that multiple processes are created in the NCCL communicator clique which performs a check against `nthread < n_gpus`

trivialfis:
@mt-jones Thanks for the insight.

> it is possible to adjust the code so that multiple processes are created in the NCCL communicator clique which performs a check against `nthread < n_gpus`

Are you saying that, in non-distributed mode, when user specifies `nthread < n_gpus`, we fallback to a slightly more complicated setup that uses process instead of threads.  But on any other time we default to using threads?

mt-jones:
> @mt-jones Thanks for the insight.
> 
> > it is possible to adjust the code so that multiple processes are created in the NCCL communicator clique which performs a check against `nthread < n_gpus`
> 
> Are you saying that, in non-distributed mode, when user specifies `nthread < n_gpus`, we fallback to a slightly more complicated setup that uses process instead of threads. But on any other time we default to using threads?

I'm saying that I've removed the flag for `distributed`. Instead, using `rabit::GetWorldSize()` to make an inference on whether XGB is being executed in a distributed manner.

Effectively, a check is performed to see if `rabit::GetWorldSize() == 1`. If so, create a NCCL communicator clique with one process; else, create a NCCL communicator clique with multiple processes.

trivialfis:
@mt-jones Ah I see.  Will look into the detail.  Thanks!

[...]

PR #4302, opened 2019-03-27
 Use `std::thread` instead of `OMP` for GPUs.

trivialfis:
Use std::thread in ExecuteIndexShards.

* Span for atomic write symbol.
* Use `ExecutePerDevice` instead of OpenMP loop.
* Use `SaveCudaContext` to make sure nothing can change master thread's device.

This decouples parameter `nthread` with `n_gpus`.  See  #4162.

I added some debugging utilities in device_helpers.cuh, please let me keep them ...

### DIFF
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -175,8 +175,7 @@ inline void CheckComputeCapability() {
   }
 }
 
-DEV_INLINE void AtomicOrByte(xgboost::common::Span<unsigned int> buffer,
-                             size_t ibyte, unsigned char b) {
+DEV_INLINE void AtomicOrByte(unsigned int* __restrict__ buffer, size_t ibyte, unsigned char b) {
   atomicOr(&buffer[ibyte / sizeof(unsigned int)], (unsigned int)b << (ibyte % (sizeof(unsigned int)) * 8));
 }
 
@@ -543,8 +542,7 @@ struct CubMemory {
 // Load balancing search
 
 template <typename CoordinateT, typename SegmentT, typename OffsetT>
-void FindMergePartitions(int device_idx,
-                         xgboost::common::Span<CoordinateT> d_tile_coordinates,
+void FindMergePartitions(int device_idx, CoordinateT *d_tile_coordinates,
                          size_t num_tiles, int tile_size, SegmentT segments,
                          OffsetT num_rows, OffsetT num_elements) {
   dh::LaunchN(device_idx, num_tiles + 1, [=] __device__(int idx) {
@@ -566,7 +564,7 @@ void FindMergePartitions(int device_idx,
 template <int TILE_SIZE, int ITEMS_PER_THREAD, int BLOCK_THREADS,
           typename OffsetT, typename CoordinateT, typename FunctionT,
           typename SegmentIterT>
-__global__ void LbsKernel(xgboost::common::Span<CoordinateT> d_coordinates,
+__global__ void LbsKernel(CoordinateT *d_coordinates,
                           SegmentIterT segment_end_offsets, FunctionT f,
                           OffsetT num_segments) {
   int tile = blockIdx.x;
@@ -631,8 +629,8 @@ void SparseTransformLbs(int device_idx, dh::CubMemory *temp_memory,
   CHECK(num_tiles < std::numeric_limits<unsigned int>::max());
 
   temp_memory->LazyAllocate(sizeof(CoordinateT) * (num_tiles + 1));
-  xgboost::common::Span<CoordinateT> tmp_tile_coordinates {
-    temp_memory->GetSpan<CoordinateT>(sizeof(CoordinateT) * (num_tiles + 1))};
+  CoordinateT *tmp_tile_coordinates =
+      reinterpret_cast<CoordinateT *>(temp_memory->d_temp_storage);
 
   FindMergePartitions(device_idx, tmp_tile_coordinates, num_tiles,
                       BLOCK_THREADS, segments, num_segments, count);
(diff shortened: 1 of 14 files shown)


trivialfis:
@hcho3 Can we upgrade the compiler at some point ... What the convention of supported compiler?

hcho3:
@trivialfis Right now, we assume that C++11 is supported. Do you want something from C++14 or C++17?

trivialfis:
@hcho3 No.  Just a compiler that has complete support for c++11.  For example you found `<regex>` is not available in g++-4.8.2,  and here the variadic arguments capturing is not working in 4.8.2.  See: https://stackoverflow.com/questions/20006574/workaround-for-variadic-lambda-capture

I might be able to workaround this later, but if we were to claim c++11 support, 4.8.2 might be too old.

hcho3:
@trivialfis Got it, we should probably bump the version. Any suggestion?

trivialfis:
@hcho3 I'm not sure about what's the status in commercial deployment.  Can you find an example for currently active distribution that uses 4.x?  If not, I would suggest sticking to one of those "popular stable distributions" like debian.

hcho3:
@trivialfis 
> I would suggest sticking to one of those "popular stable distributions" like debian.

The reason for using CentOS 6 is to maximize the compatibility of the binary wheel in many Linux distributions. Otherwise, XGBoost wheel may depend on recent versions of GLIBC and other system libraries that other distributions may lack.

That said, we can certainly upgrade the GCC version. We just need to use later version of `devtoolset` package.

trivialfis:
@hcho3 I see.  Just checked out CentOS, EOL is at the end of 2020 .....

Let's hold it and I will try to work around it for this PR.  Thanks.

trivialfis:
One lesson I learned from this PR is, under no circumstance should one change the device in master thread.  I wish I can add a test for that.

trivialfis:
@RAMitchell

src/common/device_helpers.cuh
```diff
@@ -382,13 +433,17 @@ class BulkAllocator {
...
-    for (size_t i = 0; i < d_ptr_.size(); i++) {
-      if (!(d_ptr_[i] == nullptr)) {
-        safe_cuda(cudaSetDevice(device_idx_[i]));
-        safe_cuda(cudaFree(d_ptr_[i]));
-        d_ptr_[i] = nullptr;
+    SaveCudaContext {
+      [&]() {
+        for (size_t i = 0; i < d_ptr_.size(); i++) {
+          if (!(d_ptr_[i] == nullptr)) {
+            safe_cuda(cudaSetDevice(device_idx_[i]));
```
RAMitchell: I thought we established that setting the device was unnecessary for cudaFree?

src/common/device_helpers.cuh
```diff
@@ -575,8 +631,8 @@ void SparseTransformLbs(int device_idx, dh::CubMemory *temp_memory,
   CHECK(num_tiles < std::numeric_limits<unsigned int>::max());
 
   temp_memory->LazyAllocate(sizeof(CoordinateT) * (num_tiles + 1));
-  CoordinateT *tmp_tile_coordinates =
-      reinterpret_cast<CoordinateT *>(temp_memory->d_temp_storage);
+  xgboost::common::Span<CoordinateT> tmp_tile_coordinates {
+    temp_memory->GetSpan<CoordinateT>(sizeof(CoordinateT) * (num_tiles + 1))};
```
RAMitchell: Changes relating to span should be in a different PR

src/linear/updater_gpu_coordinate.cu
```diff
@@ -53,17 +53,13 @@ class DeviceShard {
...
-      };
-      auto column_begin =
-          std::lower_bound(col.cbegin(), col.cend(),
-                           xgboost::Entry(row_begin, 0.0f), cmp);
-      auto column_end =
-          std::lower_bound(col.cbegin(), col.cend(),
-                           xgboost::Entry(row_end, 0.0f), cmp);
-      column_segments.emplace_back(
-          std::make_pair(column_begin - col.cbegin(), column_end - col.cbegin()));
+      auto cmp = [](Entry e1, Entry e2) { return e1.index < e2.index; };
```
RAMitchell: These formatting changes make the PR confusing. It's not immediately obvious if the changes are only formatting and this file will be included in the diff even when nothing has changed relating to the commit message.

hcho3:
@trivialfis For some reason, one of the GPU agents failed with error "Remote call on Linux GPU slave (i-0c8b6da83669c814e) failed". I restarted the agent.

trivialfis:
@hcho3 Thanks for helping out.

@RAMitchell Done removing unrelated changes.

RAMitchell:
Let me do a performance test then I will merge it if everything is okay.

RAMitchell:
@trivialfis there is a performance regression here. Using the following command with 8x Tesla V100-SXM2-32GB:
```
python xgboost/tests/benchmark/benchmark_tree.py --rows 10000000 --params "{'n_gpus':8, 'debug_verbose':0}"
```
Time appears to have gone from around 20s to 50s. According to my recent profiling poor scaling on multi-GPU seems related to the way we are using threads.

Another way of achieving what you are trying to do while staying with omp would be to store the omp global number of threads (set according to nthreads parameter) to a temporary variable, change the number of threads to the number of GPUs just for ExecuteShards, then change it back afterwards.

trivialfis:
@RAMitchell I wonder what happened ...  Let me see if I can reproduce the slow down.   BTW, `debug_verbose` is deprecated.

trivialfis:
With #4454 it's possible to get rid of the global OpenMP parameter, I will close this one.

PR #4271, opened 2019-03-18
[jvm-packages] logging version number

CodingCat:
in a shared cluster environment, it's nice to have explicit version information to track which version the user is using

PR #4303, opened 2019-03-28
[jvm-packages] do not filter shared library files

rongou:
The recent change #4271 causes resources to be filtered, including shared library files that are copied there (see https://github.com/dmlc/xgboost/blob/master/jvm-packages/create_jni.py#L102), corrupting them in the process:
```sh
OpenJDK 64-Bit Server VM warning: You have loaded library /tmp/libxgboost4j7996472599554222538.so which might have disabled stack guard. The VM will try to fix the stack guard now.
It's highly recommended that you fix the library with 'execstack -c <libfile>', or link it with '-z noexecstack'.
```

@CodingCat

### DIFF
--- a/jvm-packages/xgboost4j/pom.xml
+++ b/jvm-packages/xgboost4j/pom.xml
@@ -71,6 +71,18 @@
                   </execution>
               </executions>
           </plugin>
+          <plugin>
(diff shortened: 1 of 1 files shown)


CodingCat:
Thanks! I didn't expect it will also touch non-text files, will merge after CI is passed

PR #4306, opened 2019-03-28
[CI] Upgrade to GCC 5.3.1, CMake 3.6.0

hcho3:
Upgrade the Jenkins CI container to use GCC 5.3.1, in order to obtain complete C++11 support (along with `std::regex`),

### DIFF
--- a/tests/ci_build/Dockerfile.gpu
+++ b/tests/ci_build/Dockerfile.gpu
@@ -7,15 +7,16 @@ ENV DEBIAN_FRONTEND noninteractive
 # Install all basic requirements
 RUN \
     yum -y update && \
-    yum install -y tar unzip wget xz git && \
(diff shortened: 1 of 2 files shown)


hcho3:
Looks like GCC 5 has a newer version of libgomp, and it’s causing issues on Ubuntu 14.04 (version used by Travis CI). Should we revert this? @RAMitchell @trivialfis

trivialfis:
@hcho3 please don't revert.  They will upgrade the OS.  It would be painful to add back those workarounds only to remove it few months later.

trivialfis:
@hcho3 Besides, even if we revert it, people still need to compile XGBoost themselves unless we push another release immediately.  But the problem wouldn't be here if they compile XGBoost at the first place(the mismatch wouldn't happen).  So reverting won't be much of help.

hcho3:
@trivialfis Got it. I won’t revert it then

PR #4308, opened 2019-03-28
add a test for cpu predictor using external memory

rongou:
@RAMitchell @hcho3


RAMitchell:
Is there a way we can test external memory without so much data? I feel like this type of test is much slower than it really needs to be.

rongou:
@RAMitchell hmm with the current APIs it seems quite hard to test external memory without the large data. Let me know if you have any ideas. The test runs around 800 ms on my machine, seems inline with some the existing tests.

rongou:
@RAMitchell @hcho3 ok, made some minimal api changes to allow testing on smaller external memory pages. The test now runs in a few milliseconds. Please take another look.

RAMitchell:
Looks good, and this is a much more sustainable way to write tests for external memory. I will wait to see if anyone else has comments before merging.

Issue #4296, opened 2019-03-27
[jvm-packages] XGBoost Spark 0.82 crashed when handling Float.NaN


forest-sky-sea:
I saw someone has mentioned the same problem before in https://github.com/dmlc/xgboost/issues/2286

My test data is generated as follows, using Float.NaN to represent missing values in the data:
```scala
import spark.implicits._
val testDF = Seq(
   (1.0f, 0.0f, Float.NaN, 1.0),
   (1.0f, 0.0f, 1.0f, 1.0),
   (0.0f, 1.0f, 0.0f, 0.0),
   (1.0f, 0.0f, 1.0f, 1.0),
   (1.0f, Float.NaN, 0.0f, 0.0),
   (0.0f, 0.0f, 0.0f, 0.0),
   (0.0f, 1.0f, 0.0f, 2.0),
   (Float.NaN, 0.0f, 0.0f, 2.0)
 ).toDF("col1", "col2", "col3", "label")
```
Then use `VectorAssembler` to assemble feature columns:
```scala
val vectorAssembler = new VectorAssembler()
   .setInputCols(Array("col1", "col2", "col3"))
   .setOutputCol("features")
   .setHandleInvalid("keep")
val inputDF = vectorAssembler.transform(testDF).select("features", "label")
```
Finally my data is as follows:
```scala
inputDF.show()
+-------------+-----+
|     features|label|
+-------------+-----+
|[1.0,0.0,NaN]|  1.0|
|[1.0,0.0,1.0]|  1.0|
|[0.0,1.0,0.0]|  0.0|
|[1.0,0.0,1.0]|  1.0|
|[1.0,NaN,0.0]|  0.0|
|    (3,[],[])|  0.0|
|[0.0,1.0,0.0]|  2.0|
|[NaN,0.0,0.0]|  2.0|
+-------------+-----+
```
Then, I try to train a model using xgboost4j-spark 0.82
```scala
val paramMap = Map("max_depth" -> 5)
val model = new XGBoostClassifier(paramMap)
   .setLabelCol("label")
   .setFeaturesCol("features")
   .setObjectiveType("classification")
   .setObjective("multi:softprob")
   .setNumClass(3)
   .fit(inputDF)
```
The training step crashes, full stack trace:
```
[11:20:52] WARNING: /xgboost/src/learner.cc:621: Tree method is automatically selected to be 'approx' for distributed training.
[11:20:52] Tree method is automatically selected to be 'approx' for distributed training.
[...]

Issue #4294, opened 2019-03-26
[jvm-package] Inconsistent predictions using XGBoost4j-Spark with model trained in R


greg-rychlewski:
Hi,

I'm using an XGBoost model trained in R to make predictions using the xgboost4j-spark package. 

When I compare the probability predictions made by R with the ones made with xgboost4j-spark they always differ when one of the features is missing.

I'm not sure if this is a bug in the package or if some of the missing value parameters need to be adjusted. Right now I'm using the default missing value options (R's predict function defaults to NA, it looks like xgboost4j-spark's transform function defaults to NaN).

Here are the relevant parts of the code I'm using:

```
val input = assembler
      .setHandleInvalid("keep")
      .transform(data)

val model = new XGBoostClassificationModelBridge("test", 2, modelPath)
      .xgbClassificationModel
val predictions = model.transform(input)
```

I'm using version 0.82 of xgboost4j-spark with spark 2.4.0.

greg-rychlewski:
I did some more digging and found that I can get the correct probability in spark when I access the booster directly and pass in a dmatrix created from a float array:

`val data = Array(
 0.0,0.0,56.39,353.0,306.0,218793.63,0.0,99999.0,0.0,15.07,99999.0,0.0,0.0,265177.16,11.0,-2.0,-2.0,200.0,6459.0,4276.1366,2.0,1809.72,80000.0,0.0,5.3,346238.0,-4.0,999.0,305.0,305.0,0.0,1.0,50.0,0.0,82.0,999.0,0.0,165.0,999.0,26.0,150000.0,-1.0,23.0,0.0,0.0,0.0,0.0,Double.NaN).map(_.toFloat)
    val dMatrix = new DMatrix(data, nrow = 1, ncol = 48, missing = Float.NaN )
    val predict_dmatrix = bst.predict(dMatrix)
    predict_dmatrix.foreach( x => println(x.map(_.toDouble).mkString) )
\\ output = 9.949778177542612E-5 (correct)`

When I convert the same array to a dense vector and use it to create a dataframe so that I can use the mllib transform method, I get the incorrect probability:

` val v = Vectors.dense(data.map(_.toDouble))
    val df = spark.createDataFrame(Seq((0, v))).toDF("id", "features")
    val predict_df = model.transform(df)
    predict_df.show(false) 
 \\ output = 1.8168642418459058E-4 (incorrect)`

CodingCat:
[...]

PR #4309, opened 2019-03-29
[jvm-packages] handle NaN as missing value explicitly

CodingCat:
closes https://github.com/dmlc/xgboost/issues/4296 https://github.com/dmlc/xgboost/issues/4294


CodingCat:
@hcho3 somehow python test is flaky again

PR #4278, opened 2019-03-19
Performance optimizations for CPUs [part 2]

SmirnovEgorRu:
This PR should contains performance improvement for multi-core CPUs. Most of these optimizations have been taken from [Intel® DAAL](https://github.com/intel/daal) and adopted for XGBoost. Here are following changes:
1) Added nested parallelism by nodes for grow_policy=depthwise on CPU. For this QuantileHistMaker has been reworked to be thread-safe. It provides better scaling by number HW cores and number of levels in a tree.
2) Additional parallelism with grow_policy=lossguide.
3) General optimizations for ApplySplit, EvaluateSplit, BuildHist, InitData functions.
4) Moved computation of gradient statistics from InitNewNode to BuildHist function, where all required data is in hot cache. Here is good boost for sparse data.

Performance measurements and more details will be provided by me a little bit later.

Issue #3810, opened 2018-10-18
Call for contribution: improve multi-core CPU performance of 'hist'


hcho3:
It is about time to tackle the elephant in the room: performance on multi-core CPUs.

# Description of Problem
Currently, the `hist` tree-growing algorithm (`tree_method=hist`) scales poorly on multi-core CPUs: for some datasets, **performance deteriorates as the number of threads is increased**. This issue was discovered by [@Laurae2](https://github.com/Laurae2)'s [Gradient Boosting Benchmark](https://sites.google.com/view/lauraepp/new-benchmarks).

The scaling behavior is as follows for [Bosch dataset](https://www.kaggle.com/c/bosch-production-line-performance/data):
![Performance scaling on C5.9xlarge](https://raw.githubusercontent.com/hcho3/xgboost-fast-hist-perf-lab/master/scaling.png)

# Call for Contribution

I have identified the performance bottleneck of the 'hist' algorithm and put it in a small repository: [hcho3/xgboost-fast-hist-perf-lab](https://github.com/hcho3/xgboost-fast-hist-perf-lab). You can try to improve the performance by revising [src/build_hist.cc](https://github.com/hcho3/xgboost-fast-hist-perf-lab/blob/master/src/build_hist.cc).

## Some ideas
* Change the data matrix layout from [CSR](https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_row_(CSR,_CRS_or_Yale_format)) to other layouts such as [ellpack](https://docs.nvidia.com/cuda/cusparse/index.html#ellpack-itpack-format-ell)
* Try to distribute work more equally among worker threads. Work imbalance is caused by irregular sparse patterns of the data matrix.
* Group complementary features together, a common scenario for one-hot-encoded data.

hcho3:
@Laurae2 Thank you for preparing the GBT benchmark. It has been helpful in identifying the problem spot.

trivialfis:
@hcho3 Does OpenMP `guided` schedule help load balancing?  If so the ellpack won't be very useful.

hcho3:
My guess is that static allocation of work using ellpack would achieve balanced workload with lower overhead than `guided` or `dynamic` mode of OpenMP. With `dynamic`, you get runtime overhead of maintaining work stealing queue

CodingCat:
might be a bit off-topic, do we have benchmark results of `approx`?

[...]

PR #4310, opened 2019-03-29
Optimizations of pre-processing for 'hist' tree method

SmirnovEgorRu:
This commit is a part of this [pull-request](https://github.com/dmlc/xgboost/pull/4278).
Performance before this commit:

Data Set | First iteration | 50 iterations
-- | -- | --
Higgs1m, ms | 900 | 3889
Higgs10m, ms | 8923 | 33638
9m x 45, ms | 6093 | 19389

Performance results for these changes:

Data Set | First iteration | 50 iterations
-- | -- | --
Higgs1m, ms | 672 | 3502
Higgs10m, ms | 7834 | 30121
9m x 45, ms | 4369 | 15960

HW: Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, 2 sockets, 14 cores per socket.

### DIFF
--- a/src/common/column_matrix.h
+++ b/src/common/column_matrix.h
@@ -71,7 +71,7 @@ class ColumnMatrix {
   // construct column matrix from GHistIndexMatrix
   inline void Init(const GHistIndexMatrix& gmat,
                    double  sparse_threshold) {
-    const auto nfeature = static_cast<bst_uint>(gmat.cut.row_ptr.size() - 1);
+    const int32_t nfeature = static_cast<int32_t>(gmat.cut.row_ptr.size() - 1);
     const size_t nrow = gmat.row_ptr.size() - 1;
 
     // identify type of each column
@@ -86,7 +86,7 @@ class ColumnMatrix {
 
     gmat.GetFeatureCounts(&feature_counts_[0]);
     // classify features
-    for (bst_uint fid = 0; fid < nfeature; ++fid) {
+    for (int32_t fid = 0; fid < nfeature; ++fid) {
       if (static_cast<double>(feature_counts_[fid])
                  < sparse_threshold * nrow) {
         type_[fid] = kSparseColumn;
@@ -100,7 +100,7 @@ class ColumnMatrix {
     boundary_.resize(nfeature);
     size_t accum_index_ = 0;
     size_t accum_row_ind_ = 0;
-    for (bst_uint fid = 0; fid < nfeature; ++fid) {
+    for (int32_t fid = 0; fid < nfeature; ++fid) {
       boundary_[fid].index_begin = accum_index_;
       boundary_[fid].row_ind_begin = accum_row_ind_;
       if (type_[fid] == kDenseColumn) {
@@ -126,7 +126,7 @@ class ColumnMatrix {
     // pre-fill index_ for dense columns
 
     #pragma omp parallel for
-    for (bst_uint fid = 0; fid < nfeature; ++fid) {
+    for (int32_t fid = 0; fid < nfeature; ++fid) {
       if (type_[fid] == kDenseColumn) {
         const size_t ibegin = boundary_[fid].index_begin;
         uint32_t* begin = &index_[ibegin];
--- a/src/common/column_matrix.h
+++ b/src/common/column_matrix.h
@@ -124,6 +124,8 @@ class ColumnMatrix {
     }
 
     // pre-fill index_ for dense columns
+
+    #pragma omp parallel for
     for (bst_uint fid = 0; fid < nfeature; ++fid) {
       if (type_[fid] == kDenseColumn) {
         const size_t ibegin = boundary_[fid].index_begin;
@@ -184,8 +186,8 @@ class ColumnMatrix {
 
   std::vector<size_t> feature_counts_;
   std::vector<ColumnType> type_;
-  std::vector<uint32_t> index_;  // index_: may store smaller integers; needs padding
-  std::vector<size_t> row_ind_;
+  SimpleArray<uint32_t> index_;  // index_: may store smaller integers; needs padding
+  SimpleArray<size_t> row_ind_;
   std::vector<ColumnBoundary> boundary_;
 
   // index_base_[fid]: least bin id for feature fid
--- a/src/common/hist_util.cc
+++ b/src/common/hist_util.cc
@@ -11,6 +11,7 @@
 #include "./column_matrix.h"
 #include "./hist_util.h"
 #include "./quantile.h"
+#include "./../tree/updater_quantile_hist.h"
 
 #if defined(XGBOOST_MM_PREFETCH_PRESENT)
   #include <xmmintrin.h>
@@ -49,7 +50,7 @@ void HistCutMatrix::Init(DMatrix* p_fmat, uint32_t max_num_bins) {
   constexpr int kFactor = 8;
   std::vector<WXQSketch> sketchs;
 
-  const int nthread = omp_get_max_threads();
+  const size_t nthread = omp_get_max_threads();
 
   unsigned const nstep =
       static_cast<unsigned>((info.num_col_ + nthread - 1) / nthread);
@@ -67,34 +68,86 @@ void HistCutMatrix::Init(DMatrix* p_fmat, uint32_t max_num_bins) {
   // Use group index for weights?
   bool const use_group_ind = num_groups != 0 && weights.size() != info.num_row_;
 
-  for (const auto &batch : p_fmat->GetRowBatches()) {
-    size_t group_ind = 0;
-    if (use_group_ind) {
-      group_ind = this->SearchGroupIndFromBaseRow(group_ptr, batch.base_rowid);
-    }
-#pragma omp parallel num_threads(nthread) firstprivate(group_ind, use_group_ind)
-    {
-      CHECK_EQ(nthread, omp_get_num_threads());
-      auto tid = static_cast<unsigned>(omp_get_thread_num());
-      unsigned begin = std::min(nstep * tid, ncol);
-      unsigned end = std::min(nstep * (tid + 1), ncol);
+  if (use_group_ind)
+  {
+    for (const auto &batch : p_fmat->GetRowBatches()) {
+      size_t group_ind = this->SearchGroupIndFromBaseRow(group_ptr, batch.base_rowid);
+      #pragma omp parallel num_threads(nthread) firstprivate(group_ind, use_group_ind)
+      {
+        CHECK_EQ(nthread, omp_get_num_threads());
+        auto tid = static_cast<unsigned>(omp_get_thread_num());
+        unsigned begin = std::min(nstep * tid, ncol);
+        unsigned end = std::min(nstep * (tid + 1), ncol);
 
-      // do not iterate if no columns are assigned to the thread
-      if (begin < end && end <= ncol) {
-        for (size_t i = 0; i < batch.Size(); ++i) { // NOLINT(*)
-          size_t const ridx = batch.base_rowid + i;
-          SparsePage::Inst const inst = batch[i];
-          if (use_group_ind &&
-              group_ptr[group_ind] == ridx &&
-              // maximum equals to weights.size() - 1
-              group_ind < num_groups - 1) {
-            // move to next group
-            group_ind++;
+        // do not iterate if no columns are assigned to the thread
+        if (begin < end && end <= ncol) {
+          for (size_t i = 0; i < batch.Size(); ++i) { // NOLINT(*)
+            size_t const ridx = batch.base_rowid + i;
+            SparsePage::Inst const inst = batch[i];
+            if (group_ptr[group_ind] == ridx &&
+                // maximum equals to weights.size() - 1
+                group_ind < num_groups - 1) {
+              // move to next group
+              group_ind++;
+            }
+            for (auto const& entry : inst) {
+              if (entry.index >= begin && entry.index < end) {
+                size_t w_idx = group_ind;
+                sketchs[entry.index].Push(entry.fvalue, info.GetWeight(w_idx));
+              }
+            }
           }
-          for (auto const& entry : inst) {
-            if (entry.index >= begin && entry.index < end) {
-              size_t w_idx = use_group_ind ? group_ind : ridx;
-              sketchs[entry.index].Push(entry.fvalue, info.GetWeight(w_idx));
+        }
+      }
+    }
+  } else {
+    for (const auto &batch : p_fmat->GetRowBatches()) {
+      const size_t size = batch.Size();
+      const size_t block_size = 512;
+      const size_t block_size_iter = block_size * nthread;
+      const size_t n_blocks = size / block_size_iter + !!(size % block_size_iter);
+
+      std::vector<std::vector<std::pair<float, float>>> buff(nthread);
+      for(size_t tid = 0; tid < nthread; ++tid) {
+        buff[tid].resize(block_size * ncol);
+      }
+
+      std::vector<size_t> sizes(nthread * ncol, 0);
+
+      for (size_t iblock = 0; iblock < n_blocks; ++iblock) {
+        #pragma omp parallel num_threads(nthread)
+        {
+          int tid = omp_get_thread_num();
+
+          const size_t ibegin = iblock * block_size_iter + tid * block_size;
+          const size_t iend = std::min(ibegin + block_size, size);
+
+          auto* p_sizes = sizes.data() + ncol * tid;
+          auto* p_buff = buff[tid].data();
+
+          for(size_t i = ibegin; i < iend; ++i) {
+            size_t const ridx = batch.base_rowid + i;
+            bst_float w = info.GetWeight(ridx);
+            SparsePage::Inst const inst = batch[i];
+
+            for (auto const& entry : inst) {
+              const size_t idx = entry.index;
+              p_buff[idx * block_size + p_sizes[idx]] = { entry.fvalue, w };
+              p_sizes[idx]++;
+            }
+          }
+          #pragma omp barrier
+          #pragma omp for schedule(static)
+          for(size_t icol = 0; icol < ncol; ++icol) {
+            for(size_t tid = 0; tid < nthread; ++tid) {
+              auto* p_sizes = sizes.data() + ncol * tid;
+              auto* p_buff = buff[tid].data() + icol * block_size;
+
+              for(size_t i = 0; i < p_sizes[icol]; ++i) {
+                sketchs[icol].Push(p_buff[i].first,  p_buff[i].second);
+              }
+
+              p_sizes[icol] = 0;
             }
           }
         }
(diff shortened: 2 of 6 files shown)


src/tree/updater_quantile_hist.cc
```diff
@@ -420,25 +420,74 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
     // mark subsample and build list of member rows
+
     if (param_.subsample < 1.0f) {
       std::bernoulli_distribution coin_flip(param_.subsample);
       auto& rnd = common::GlobalRandom();
+      size_t j = 0;
       for (size_t i = 0; i < info.num_row_; ++i) {
         if (gpair[i].GetHess() >= 0.0f && coin_flip(rnd)) {
-          row_indices.push_back(i);
+          p_row_indices[j++] = i;
```
CodingCat: why adding values to the underlying array directly is performance friendly?

SmirnovEgorRu: There are 2 points:
1) push_back has at least 1 "if" to check current capacity with new size, even if reserve() was called. It leads to additional branching and performance slowdown.
2) constructions like this:
```cpp
if (a[i] > c)
   b[j++] = a[i];
```
Can be easily vectorized by AVX-512 instructions - vcompressps/vcompresspd and some mask-instructions for comparison, even with compiler auto-vectorization. I know, that now XGBoost from a box supports only sse4.2, however AVX-512 will be possibly available in the future.
With push_back() it doesn't work.

SmirnovEgorRu:
I see failed tests for R. But logs contain these messages:
Error in download.file(url, destfile, method, mode = "wb", ...) : 
  cannot open URL 'http://cran.rstudio.com/bin/macosx/el-capitan/contrib/3.5/gh_1.0.1.tgz'
ERROR: dependencies ‘callr’, ‘cli’, ‘digest’, ‘httr’, ‘jsonlite’, ‘memoise’, ‘pkgbuild’, ‘pkgload’, ‘rcmdcheck’, ‘remotes’, ‘sessioninfo’ are not available for package ‘devtools’

Looks like these errors don't depend on my commit. May I take that tests are passed?

CodingCat:
I have retriggered the test, should pass this time

SmirnovEgorRu:
@CodingCat thanks. Looks like it is okay now.

src/common/hist_util.cc
```diff
@@ -177,22 +229,66 @@ uint32_t HistCutMatrix::GetBinIdx(const Entry& e) {
...
+    new_size += batch.Size();
+  }
+
+  row_ptr.resize(new_size);
+  row_ptr[0] = 0;
+
+  size_t rbegin = 0;
+  size_t prev_sum = 0;
+
+  for (const auto &batch : p_fmat->GetRowBatches()) {
```
hcho3: Note to @trivialfis, @CodingCat: we are not really sure if 'hist' method works when external memory is enabled (i.e. `GetRowBatches()` returns more than one batch). We will need to test 'hist' against external memory in a future pull request.

src/common/hist_util.cc
```diff
@@ -177,22 +229,66 @@ uint32_t HistCutMatrix::GetBinIdx(const Entry& e) {
...
+        for (size_t i = ibegin; i < iend; ++i) {
+          row_ptr[rbegin + 1 + i] += p_part[tid];
+        }
+      }
     }
+
     index.resize(row_ptr.back());
 
     CHECK_GT(cut.cut.size(), 0U);
-    CHECK_EQ(cut.row_ptr.back(), cut.cut.size());
```
hcho3: @SmirnovEgorRu Why are we removing this line?

src/common/hist_util.h
```diff
@@ -19,6 +19,70 @@
 namespace xgboost {
 namespace common {
 
+template<typename T>
```
hcho3:
```suggestion
/*
 * \brief A thin wrapper around dynamically allocated C-style array.
 * Make sure to call resize() before use.
 */
template<typename T>
```

src/common/hist_util.h
```diff
@@ -19,6 +19,70 @@
...
+
+  void resize(size_t n) {
+    T* ptr = static_cast<T*>(malloc(n*sizeof(T)));
+    memcpy(ptr, ptr_, n_ * sizeof(T));
+    free(ptr_);
+    ptr_ = ptr;
+    n_ = n;
+  }
+
+  T& operator[](size_t idx) {
```
hcho3: What is the advantage of using `SimpleArray` over `std::vector`? We still have one extra indirection each time we index the array.

src/tree/updater_quantile_hist.h
```diff
@@ -28,6 +28,38 @@
 #include "../common/column_matrix.h"
 
 namespace xgboost {
+
+template<typename T, size_t MaxStackSize>
```
hcho3:
```suggestion
/*!
 * \brief A C-style array with in-stack allocation. As long as the array is smaller than MaxStackSize, it will be allocated inside the stack. Otherwise, it will be heap-allocated.
 */
template<typename T, size_t MaxStackSize>
```

hcho3: @SmirnovEgorRu Is stack allocation more performant than heap allocation?

src/tree/updater_quantile_hist.cc
```diff
@@ -420,25 +420,74 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
         }
       }
+      row_indices.resize(j);
     } else {
-      for (size_t i = 0; i < info.num_row_; ++i) {
-        if (gpair[i].GetHess() >= 0.0f) {
-          row_indices.push_back(i);
+      MemStackAllocator<int, 128> buff(this->nthread_);
+      int* p_buff = buff.Get();
```
```suggestion
      bool* p_buff = buff.Get();
```

src/tree/updater_quantile_hist.cc
```diff
@@ -420,25 +420,74 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
-          row_indices.push_back(i);
+          p_row_indices[j++] = i;
         }
       }
+      row_indices.resize(j);
     } else {
-      for (size_t i = 0; i < info.num_row_; ++i) {
-        if (gpair[i].GetHess() >= 0.0f) {
-          row_indices.push_back(i);
+      MemStackAllocator<int, 128> buff(this->nthread_);
```
```suggestion
      MemStackAllocator<bool, 128> buff(this->nthread_);
```

src/tree/updater_quantile_hist.cc
```diff
@@ -420,25 +420,74 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
+          if (gpair[i].GetHess() < 0.0f) {
+            buff.Get()[tid] = true;
+            break;
+          }
+        }
+      }
+
+      bool has_heg_hess = false;
+      for (size_t tid = 0; tid < this->nthread_; ++tid) {
+        if (p_buff[tid] == true) {
```
```suggestion
        if (p_buff[tid]) {
```

src/tree/updater_quantile_hist.cc
hcho3: @SmirnovEgorRu Did you intend to use `int` to avoid false sharing? Or can we use `bool` instead?

src/tree/updater_quantile_hist.cc
```diff
@@ -420,25 +420,74 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
+
+        for (size_t i = ibegin; i < iend; ++i) {
+          if (gpair[i].GetHess() < 0.0f) {
+            buff.Get()[tid] = true;
+            break;
+          }
+        }
+      }
+
+      bool has_heg_hess = false;
```
```suggestion
      bool has_neg_hess = false;
```

src/tree/updater_quantile_hist.cc
```diff
@@ -420,25 +420,74 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
+      for (size_t tid = 0; tid < this->nthread_; ++tid) {
+        if (p_buff[tid] == true) {
+          has_heg_hess = true;
+        }
+      }
+
+      if (has_heg_hess) {
+        size_t j = 0;
+        for (size_t i = 0; i < info.num_row_; ++i) {
+          if (gpair[i].GetHess() >= 0.0f) {
```
hcho3: Note to @trivialfis, @RAMitchell: I wonder if we should just forbid negative Hessian values. All the non-negativity checks in the updater codebase are not only eye-sores but also reduces parallelism.

src/common/row_set.h
```diff
@@ -59,7 +59,6 @@ class RowSetCollection {
   }
   // clear up things
   inline void Clear() {
-    row_indices_.clear();
```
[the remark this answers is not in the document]
SmirnovEgorRu: I did some measurements with my all optimizations and InitData() function spends ~15% of tree building time.
This block of code spends all of this time:
```cpp
for (size_t i = 0; i < info.num_row_; ++i) {
  if (gpair[i].GetHess() >= 0.0f) {
    row_indices.push_back(i);
```
There are 2 reasons: std::vector - not performance friendly object, push_back contains at least one "if" to check current size. Also, this code can't be parallelized and it is sequential.

I have changed this code to use C-style partition:
```cpp
j = 0;
for (size_t i = 0; i < info.num_row_; ++i) {
  if (gpair[i].GetHess() >= 0.0f) {
    p_row_indices[j++] = i;
```
But in this case I need to do row_indices_->resize(num_row) each time after row_indices_->clear(). But resize(num_rows) - very expensive operations if current size=0, it will fill data inside by default value (in our case - 0) by not performance manner. So, I deleted clear() and use resize(nrows) only one time.

After my changes InitData() spends less then 1% now.

P.S. std::vector - really not performance friendly :)

src/common/hist_util.cc
```diff
@@ -67,34 +68,85 @@ void HistCutMatrix::Init(DMatrix* p_fmat, uint32_t max_num_bins) {
...
-          if (use_group_ind &&
-              group_ptr[group_ind] == ridx &&
-              // maximum equals to weights.size() - 1
-              group_ind < num_groups - 1) {
-            // move to next group
-            group_ind++;
+  } else {
+    for (const auto &batch : p_fmat->GetRowBatches()) {
+      const size_t size = batch.Size();
+      const size_t block_size = 512;
```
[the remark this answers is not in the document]
SmirnovEgorRu: No, but this value influences on size of allocated buffer: it should not be too large from the one point of view and be not too small (useful work of omp-task should be more than omp-overhead) from the second point of view.

By experiments 512 have been chosen as optimal.

src/common/hist_util.cc
SmirnovEgorRu: Now row_ptr is allocated to final length from the start. So if we have a lot of batches - back() for not final batch will return some trash.

src/tree/updater_quantile_hist.cc
SmirnovEgorRu: False sharing should not affect this place

src/tree/updater_quantile_hist.cc
SmirnovEgorRu: committed, thanks

src/tree/updater_quantile_hist.cc
SmirnovEgorRu: commited

src/tree/updater_quantile_hist.cc
SmirnovEgorRu: committed

SmirnovEgorRu:
@hcho3 
Yes, blocking is common practice, even if you don't do it explicitly - omp will do blocking for you implicitly.
On practice prefetch is not needed for modern CPUs, only in some places when order of memory-access is not trivial. In case of BuildHist it helps, but don't see when it can be applied more.

src/common/column_matrix.h
```diff
@@ -71,7 +71,7 @@ class ColumnMatrix {
   // construct column matrix from GHistIndexMatrix
   inline void Init(const GHistIndexMatrix& gmat,
                    double  sparse_threshold) {
-    const auto nfeature = static_cast<bst_uint>(gmat.cut.row_ptr.size() - 1);
+    const int32_t nfeature = static_cast<int32_t>(gmat.cut.row_ptr.size() - 1);
```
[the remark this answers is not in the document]
SmirnovEgorRu: Windows requires that integer type for count inside "#pragma omp paralell for" should have signed type.
Before nfeature wasn't used for parallel loop, but now it is used. So it was changed from unsigned to signed

trivialfis:
Sorry for the delay, will review shortly.

src/tree/updater_quantile_hist.h
trivialfis: @hcho3 Yes, stack allocation should be more performant, help allocation requires some searching/merging to find available memory block.  `malloc` itself manages some sort of memory pool.

src/common/hist_util.cc
CodingCat: AFAICT, it doesn't work in distributed setting, even after @trivialfis 's fix in 0.82

RAMitchell:
@SmirnovEgorRu I have a slightly off topic question: for your experiments do you use hyper threading? In my work I found that xgboost can perform very poorly and with high variance when using all threads (default nthreads setting). 

This is my CPU:
```
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                80
On-line CPU(s) list:   0-79
Thread(s) per core:    2
Core(s) per socket:    20
Socket(s):             2
NUMA node(s):          2
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 79
Model name:            Intel(R) Xeon(R) CPU E5-2698 v4 @ 2.20GHz
Stepping:              1
CPU MHz:               1656.875
CPU max MHz:           3600.0000
CPU min MHz:           1200.0000
BogoMIPS:              4391.55
Virtualization:        VT-x
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              51200K
NUMA node0 CPU(s):     0-19,40-59
NUMA node1 CPU(s):     20-39,60-79
```

I am wondering if this is a problem that affects others and if so how can we solve it.

@Laurae2 did you also experience this in your xgboost experiments?

Laurae2:
@RAMitchell xgboost fast histogram has poor scalability, therefore hyperthreads is beneficial only with a small number of threads (such as on a small desktop or a laptop). For instance on a large enough dataset, my i7-7700HQ gets a 40% performance boost from hyperthreads, but my Dual E5-2699v4 (and my Dual Xeon 6154) gets from 50% slowdown to 2x slowdown (the bigger the dataset, the smaller the slowdown).

I recommend (@SmirnovEgorRu also recommends) to disable hyperthreading. Another solution is to use OpenMP environment parameters to pin threads to specific cores when using xgboost with many threads.

Also note if using gcc or icc. I noticed icc provides the same performance as gcc for xgboost for gcc recent versions (before @SmirnovEgorRu performance improvements). For later versions of xgboost, I did not test on icc. More info here: https://sites.google.com/view/lauraepp/benchmarks/intel-compiler-may-2018

src/common/hist_util.cc
```diff
@@ -67,34 +68,85 @@ void HistCutMatrix::Init(DMatrix* p_fmat, uint32_t max_num_bins) {
   // Use group index for weights?
   bool const use_group_ind = num_groups != 0 && weights.size() != info.num_row_;
 
-  for (const auto &batch : p_fmat->GetRowBatches()) {
-    size_t group_ind = 0;
-    if (use_group_ind) {
-      group_ind = this->SearchGroupIndFromBaseRow(group_ptr, batch.base_rowid);
+  if (use_group_ind) {
```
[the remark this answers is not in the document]
SmirnovEgorRu: I replaced std::vector by SimpleArray because std::vector has to slow resize() function, it fill memory by very not performance manner.
Amortization of  allocation/deallocation is not needed here, because I need to call resize() only one time.

src/common/hist_util.cc
```diff
@@ -67,34 +68,85 @@ void HistCutMatrix::Init(DMatrix* p_fmat, uint32_t max_num_bins) {
...
-            // move to next group
-            group_ind++;
+  } else {
+    for (const auto &batch : p_fmat->GetRowBatches()) {
+      const size_t size = batch.Size();
+      const size_t block_size = 512;
+      const size_t block_size_iter = block_size * nthread;
+      const size_t n_blocks = size / block_size_iter + !!(size % block_size_iter);
+
+      std::vector<std::vector<std::pair<float, float>>> buff(nthread);
```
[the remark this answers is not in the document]
SmirnovEgorRu: Buffs are not needed for better cache locality.
It requires for collecting local features-values by each thread and after push them to global queue without conflicts

src/common/hist_util.cc
```diff
@@ -67,34 +68,85 @@ void HistCutMatrix::Init(DMatrix* p_fmat, uint32_t max_num_bins) {
...
+              p_sizes[idx]++;
+            }
           }
-          for (auto const& entry : inst) {
-            if (entry.index >= begin && entry.index < end) {
-              size_t w_idx = use_group_ind ? group_ind : ridx;
-              sketchs[entry.index].Push(entry.fvalue, info.GetWeight(w_idx));
+          #pragma omp barrier
+          #pragma omp for schedule(static)
+          for (int32_t icol = 0; icol < static_cast<int32_t>(ncol); ++icol) {
```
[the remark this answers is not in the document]
SmirnovEgorRu: OMP must be here :)
Each OMP task will push around 512*nthreads/ncols elements, it is enough to have gain from OMP. Also, OMP threads have already created on 117 line.

src/tree/updater_quantile_hist.h
SmirnovEgorRu: @hcho3, yes, because even with single-thread application stack allocation - really cheap.
But most important point here that in multi-threaded application calling of common system malloc() in different threads at the same moment leads to threads locking. In some workloads it can became really hotspot.

For memory allocation with threading better to use tbb_malloc or jemalloc - they are pretty good in this case. In XGBoost we don't have them, so I used stack allocator for small buffers.

Some more details about advantages of scalable mallocs can be found here:
https://www.infoworld.com/article/3201285/why-effective-parallel-programming-must-include-scalable-memory-allocation.html

src/tree/updater_quantile_hist.h
```diff
@@ -28,6 +28,38 @@
...
+class MemStackAllocator {
```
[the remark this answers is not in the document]
SmirnovEgorRu: std::vector - very not efficient, but probably in this case it will work.
if you come back tbb - they have std-style ScalableAllocator which can be useful here.

src/common/hist_util.h
SmirnovEgorRu: I replaced std::vector by SimpleArray because std::vector has to slow resize() function, it fill memory by very not performance manner.
Amortization of allocation/deallocation is not needed here, because I need to call resize() only one time.

SmirnovEgorRu:
@RAMitchell, I agree with @Laurae2, in case of XGBoost better to disable HyperThreading.
I have already written it [here](https://github.com/dmlc/xgboost/issues/3810#issuecomment-459061348)

However, after all my changes (https://github.com/dmlc/xgboost/pull/4278) I observed that HyperThreading is able to provide improvements for some data sets, but need investigate this fact,

SmirnovEgorRu:
@hcho3 @CodingCat @trivialfis,
Hi, do I need to do something else for this PR?

CodingCat:
LGTM

src/tree/updater_quantile_hist.cc
trams: I am very interested in trying to recompile with AVX-512 and testing it in a real world

PR #4313, opened 2019-03-30
Revert "[jvm-packages] handle NaN as missing value explicitly"

CodingCat:
mistakenly merged......Reverts dmlc/xgboost#4309


CodingCat:
instead of revert, will revise based on the previous merge

PR #4316, opened 2019-03-31
Update cmake with native cuda support

RAMitchell:
FindCuda.cmake is deprecated, making it a good time to switch.

Native cuda support in cmake allows me to enable compilation dynamic parallelism in future (which I couldn't do with old cuda cmake) as well as enables multithreaded compilation on Windows, saving me a lot of development time.

@trivialfis I think we discussed this in the past and I was hesitant to switch because of problems, but this time when I tried it everything went smoothly so I guess I just did something wrong last time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required (VERSION 3.2)
-project(xgboost)
+project(xgboost LANGUAGES CXX)
 include(cmake/Utils.cmake)
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/modules")
 find_package(OpenMP)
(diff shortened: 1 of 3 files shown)


trivialfis:
> I think we discussed this in the past

@RAMitchell Yap.  There were two concerns around this.  Other than the one you mentioned above, newer version of CMake is rarely packaged in distributions.  Here is a search result from Ubuntu 16.04LTS: https://packages.ubuntu.com/search?suite=xenial&searchon=names&keywords=cmake

So we might need to think it through ...

trivialfis:
@RAMitchell  have a branch for a complete rewrite of CMake: https://github.com/trivialfis/xgboost/tree/cmake  .  I don't want to pull it in because it still uses old CUDA wrapper.  But if we were to dump to CMake version,  I would like to merge it here, and enable distributing C-APIs along the way.

RAMitchell:
@trivialfis I think I am ok with Ubuntu 16.04LTS users having to upgrade their cmake. We have prepackaged binary releases including nccl now so I think building from source is less common.

What do you want me to do with this PR? Should I finish it or would you prefer to do it yourself as a part of the cmake rewrite?

trivialfis:
@RAMitchell  I would like to make it part of my rewrite.

trivialfis:
@RAMitchell Feel free to reopen.

PR #4322, opened 2019-04-02
Lift gpu_hist implementation into header.

trivialfis:
Test file no longer includes a .cu file.  

This change is to prepare CMake upgrade.  After introducing CUDA as a language in CMake, the duplicated definitions of dmlc registrations and CUDA kernels are caught by linker.

If we were to link the test executable to `libxgboost.so` instead of using CMake OBJECT target, the `gpu_hist` will register twice, which also results in an error.

No logical change is introduced.

@RAMitchell The change is not preferred since it overrides git history.  I guess it's a technical debt we now have to pay.  Suggestions are welcomed.


RAMitchell:
Are you sure this is necessary? In the case of duplicating dmlc registrations for the google tests I was able to get around this in the past by enclosing the registration in a macro that checks if google test is included.

trivialfis:
@RAMitchell Let's see how #4323 goes.

PR #4323, opened 2019-04-02
Refactor CMake scripts.

trivialfis:
* Demo for using CMake with XGBoost.
* Document for using CMake with XGBoost.
* Use CMake CUDA language.
* Use CMake to handle Doxygen.
* Split up CMakeList.
* Export install target.
* Remove build.sh
* Workaround for gpu_hist test.
* Remove build_config.h.in
* Bump CMake for GPU to 3.12
* Move compilation database related code to tidy.py

TODOs:
- [x] Bring back clang-tidy.
- [x] Enable CLI test in Jenkins.
- [x] Get rabit build on Windows.
- [x] Clean up.
- [x] Fix R installation hack.
- [x] Correct tidy flags.


trivialfis:
@RAMitchell Thanks for the suggestion, I added your workaround for dmlc registrations.  But the added `inline` keyword for GPU kernel generates a warning.

jschueller:
Hi, it could install the python package in a dir relative to CMAKE_INSTALL_PREFIX (lib/pythonx.ysite-packages on unix and Lib/site-packages on win32).

trivialfis:
@jschueller That would be part of Python setuptools.   I don't think CMake is the right tool to install Python package.

jschueller:
Correct me if I'm wrong but it seems the python bindings is just made of a bunch of .py files, I guess cmake could copy them to the site dir.

trivialfis:
@jschueller setuptools does a lot more than copying scripts.  It helps compile Python scripts into bytecodes, creates a Packages and handles its dependencies.

trivialfis:
@hcho3 Related to #4331.  I bumped CMake to 3.8 but another attempt by @RAMitchell uses 3.9, see #4316.  I'm not sure this PR will work out smoothly as you can see from the tests, my build failed in various ways depending on different platforms, while built perfectly on my machine.  I believe there are lots of different bugs from different platforms ...  So please don't put too much weight on this PR for now,  I'm not sure I can workaround all of them ...

hcho3:
@trivialfis Okay, in that case, should we keep the current CMake version (3.6.0) for now?

trivialfis:
@hcho3 Yep, thanks.  I will upgrade it when it's done.

trivialfis:
List of known issues in CMake CUDA lang:
- Chosen compiler by exporting `CC` and `CXX` doesn't not get propagated to CUDA.
- `CXX_STANDARD` does not get propagated to CUDA.
- Flags like `-Xcompiler xxx` or `-gencode arch,code ...` that contain space does not do well with `target_compile_options`.

trivialfis:
As mentioned by @RAMitchell and @hcho3, I should write a short introduction about rationale about this rewrite.  Here it's. ;-)

# Goal
tl;dr: Enable C APIs with easy to use build system interface:
```
find_package(xgboost REQUIRED)
target_link_libraries(mytarget xgboost::xgboost)
```
The above two lines should get include path and linking done.

# Why it takes a rewrite to accomplish this
There's a trend about [Modern CMake](https://cliutils.gitlab.io/modern-cmake/), where one specify build options target by target.  The extra effort is not only to make the code look nicer, but a fundamental requirement for making a usable shared library.  Since we are building C APIs, user of the interface should not know anything about what libraries xgboost need, what extra flags are required to build a project with xgboost, nor the CXX standard we use.  For instance, users of xgboost should not need to pass `OpenMP` flags to their projects.  That means we need to seal all these stuffs behind the build system.

With `CMake`, all requirements of using a library is abstracted in a installed config file like `xgboost-config.cmake` which is used by `find_package` command.  And the target based style `CMake` is to make sure we only export absolute necessary build specification to our interface.  In xgboost, that's only include directory and the path of linkable object.  Commands like `include_directories`, `link_libraries` and global variables that don't distinguish between *interface* and *internal* should **not** be used any more.

# What problems this won't solve
Symbols visibility.  For a shared library, providing restricted but stable ABI is important for easy distribution.  In some cases one would pass flags like `gcc -fvisibility=hidden` to make all symbols hidden from interface except the ones marked by `__attribute__(visibility("default"))`.  There are other complications in this like whether `dmlc::Error` should be visible to other users.  This PR doesn't address this problem, and not likely in the future since we have to go through `dmlc-core` and `rabit` to make a clear sense about the implications.

trivialfis:
@RAMitchell Could you take a look at the above known issues?  I felt like there are too many things to work around ...

hcho3:
@trivialfis Can we just use a recent version of CMake to avoid known issues?

trivialfis:
@hcho3 Currently I build it with CMake 3.12, which ships with Ubuntu 18.10.  Do you think it's a little too aggressive? @RAMitchell

trivialfis:
Currently it's the last item
>  Failed to link cuda runtime before CMake@3.12. Probably due to our use of object library.

I can't work around.  I don't know how to link CUDA runtime manually.  Another way to do it might be using static library with PIC for libxgboost at first then convert it into shared library later ..  But it's getting more and more weird.

hcho3:
@trivialfis I'm okay with using CMake 3.12 for CUDA code.

RAMitchell:
@trivialfis I am ok with requiring cmake 3.12. I don't have any experience with your particular issue linking the run-time, but in the past I have built the project by combining GPU code into a static library and linking this into the xgboost shared library. This was a workaround for linking issues due to dmlc register, at the time I could not make it work using object libraries.

demo/regression/machine.conf
```diff
@@ -1,30 +1,28 @@
 # General Parameters, see comment for each definition
 # choose the tree booster, can also change to gblinear
 booster = gbtree
-# this is the only difference with classification, use reg:squarederror to do linear classification
+# this is the only difference with classification, use reg:linear to do linear classification
```
hcho3: Can you remove `demo/regression/machine.conf` from this pull request? It's not relevant to the CMake refactor.

tests/ci_build/Dockerfile.gpu
```diff
@@ -15,8 +15,8 @@ RUN \
     wget https://repo.continuum.io/miniconda/Miniconda2-4.3.27-Linux-x86_64.sh && \
     bash Miniconda2-4.3.27-Linux-x86_64.sh -b -p /opt/python && \
     # CMake
-    wget -nv -nc https://cmake.org/files/v3.6/cmake-3.6.0-Linux-x86_64.sh --no-check-certificate && \
-    bash cmake-3.6.0-Linux-x86_64.sh --skip-license --prefix=/usr
+    wget -nv -nc https://cmake.org/files/v3.12/cmake-3.12.0-Linux-x86_64.sh --no-check-certificate && \
+    bash cmake-3.12.0-Linux-x86_64.sh --skip-license --prefix=/usr
```
hcho3: Don't forget to update `tests/ci_build/Dockerfile.clang_tidy` as well.

demo/regression/machine.conf
trivialfis: @hcho3 It's used in a test for CLI.  But I can split it into another PR.

trivialfis:
@hcho3 @RAMitchell Ready for another review.  The build system is quite huge by now,  CMake has its limitation.  If you find anything that is not clear or over complicated in the code, please let me know.  I don't want anyone get stumbled by the build system.

trivialfis:
Added a short intro in `demo/c-api/readme.md` for using CMake with xgboost.

trivialfis:
@RAMitchell Would be be able to try it on Windows with CUDA when time allows?

cmake/Doc.cmake
```diff
@@ -0,0 +1,16 @@
+function (run_doxygen)
+  find_package(Doxygen REQUIRED)
+
+  if (NOT DOXYGEN_DOT_FOUND)
+    message(FATAL_ERROR "Command `dot` not found.  Please install graphviz.")
```
RAMitchell: "Please install graphviz." Should this be doxygen?

demo/c-api/README.md
```diff
@@ -0,0 +1,30 @@
+C-APIs
+===
+
+**XGBoost** implements a C API originally designed for various language
+bindings.  For detailed reference, please check xgboost/c_api.h.  Here is a
+demonstration of using the API.
+
+# CMake
```
RAMitchell: We need tests for this demo using cmake in a future PR. Functionality that is not tested is likely to be broken accidentally.

cmake/Doc.cmake
trivialfis: @RAMitchell No.  It's the `dot` command used to generate class hierarchy.

demo/c-api/README.md
trivialfis: @hcho3 is changing the ci I think?  To test this we need to create two dummy projects one for each case.

hcho3: @trivialfis Yes, I'm going to review this PR right now, and then file a PR to refactor Jenkins.

CMakeLists.txt
```diff
@@ -1,359 +1,205 @@
-cmake_minimum_required (VERSION 3.2)
-project(xgboost)
+cmake_minimum_required(VERSION 3.2)
+project(xgboost LANGUAGES CXX C VERSION 0.82)
```
hcho3: Any reason why we put the version number here?

cmake/Utils.cmake
```diff
@@ -29,25 +28,37 @@ function(msvc_use_static_runtime)
...
       endforeach()
+      set(variables
+          CMAKE_CUDA_FLAGS_DEBUG
+          CMAKE_CUDA_FLAGS_MINSIZEREL
+          CMAKE_CUDA_FLAGS_RELEASE
+          CMAKE_CUDA_FLAGS_RELWITHDEBINFO
+      )
+      foreach(variable ${variables})
+          if(${variable} MATCHES "-MD")
+              string(REGEX REPLACE "-MD" "-MT" ${variable} "${${variable}}")
```
hcho3: What is this section for? Doesn't the previous section already replace `/MD` with `/MT` for MSVC?

hcho3: @RAMitchell Does MSVC use both `/MD` and `-MD` flags?

CMakeLists.txt
trivialfis: @hcho3 It will be exported to the CMake install file, used by client CMake script `find_package(xgboost 0.82)`

cmake/Utils.cmake
trivialfis: @RAMitchell My guess is nvcc uses Unix like `-` syntax on Windows instead of `/`?  So one section for CXX and another for CUDA.  Notice the names of processed variables. `CXX` vs `CUDA`.

CMakeLists.txt
hcho3: @trivialfis Would a user want to install multiple versions of XGBoost? Otherwise, the version of XGBoost is determined by the git commit of cloned XGBoost repository.

cmake/Utils.cmake
hcho3: Got it.

CMakeLists.txt
trivialfis: @hcho3 
> Would a user want to install multiple versions of XGBoost

I don't know.  But it seems weird that there's no way to determine the version of a package?  When you try to find your dependency, making sure it's greater than some version seems to be a common practice right?  Like using `find_package(Boost 1.64)` to make sure boost is **at least** 1.64 ...

It's possible to call `git log` then process the strings to get the correct version.  But let leave it as the last resort.

hcho3: @trivialfis Ah so the user can specify the minimum required version of XGBoost. Got it.

cmake/Utils.cmake
```diff
@@ -69,19 +80,23 @@ function(format_gencode_flags flags out)
...
+  set(${out} "${${out}}--generate-code=arch=compute_${ver},code=compute_${ver};")
+
   set(${out} "${${out}}" PARENT_SCOPE)
 endfunction(format_gencode_flags flags)
 
 # Assembles the R-package files in build_dir;
 # if necessary, installs the main R package dependencies;
 # runs R CMD INSTALL.
 function(setup_rpackage_install_target rlib_target build_dir)
+  # backup cmake_install.cmake
```
hcho3: What is this for?

trivialfis: @hcho3 The next line removes whole R-package build directory, resulting in a missing `cmake_install.cmake` file.  That will cause installation failure.

src/tree/updater_gpu_hist.cu
```diff
@@ -47,8 +49,9 @@ struct GPUHistMakerTrainParam
                   "-1 to use all rows assignted to a GPU, and 0 to auto-deduce");
   }
 };
-
+#if !defined(GTEST_TEST)
```
hcho3: So we are excluding parameter registration when `GTEST_TEST` is defined? What benefit does it bring? Is this related to Google Test?

trivialfis: @hcho3 No benefit.  Just to workaround the gpu hist test without overriding the file's git history.  See #4322

src/CMakeLists.txt
```diff
@@ -0,0 +1,107 @@
...
+  CXX_STANDARD_REQUIRED ON)
+target_compile_definitions(objxgboost
+  PRIVATE
+  -DDMLC_LOG_CUSTOMIZE=1  # enable custom logging
+  $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:_MWAITXINTRIN_H_INCLUDED>
+  ${XGBOOST_DEFINITIONS})
+if (XGBOOST_MM_PREFETCH_PRESENT)
+  target_compile_definitions(objxgboost
+    PRIVATE
+    -DXGBOOST_MM_PREFETCH_PRESENT=1)
```
hcho3: What is the benefit of removing `build_config.h.in` and `configure_file()` ?

trivialfis: @hcho3 I prefer config file too.  One benefit of using something like `build_config.h` is to define macros that are visible to users, making them part of the interface.  If we were to employ config file instead of definitions, we should use it consistently for all defined macros and generate that file to build tree instead of source tree.  But that can not be done without first deprecating makefiles since make doesn't have a concept of `build tree vs source tree`.

So, using definitions for all macros is a temporary but consistent solution.

trivialfis:
Ping @CodingCat

trivialfis:
Thanks for all the help!

Issue #4240, opened 2019-03-08
Create C API doc and tutorial


hcho3:
I have seen multiple users asking for an example of C API usage. Documentation would be useful for people wanting to use XGBoost from a C++ program.

- [x] Generate API doc for C API functions
- [ ] Add an option in CMake to build a static library instead of shared
- [x] Create a tutorial with an end-to-end example
- [x] Export proper cmake install targets
- [x] Remove all deprecated functions

trivialfis:
@hcho3 For exposing C APIs:

- [ ] Export proper cmake install targets.
- [ ] Remove all deprecated functions.

hcho3:
@trivialfis Do we want to add an option to build a static library? Is it possible to build static libs with CUDA?

trivialfis:
Haven't try it yet.  What's the benefit of using static library?

hcho3:
You can statically link XGBoost with your application, so that you can distribute one binary instead of two.

trivialfis:
@hcho3 I think it's possible, but there are always surprises.  Just replace `add_library(xgboost SHARED)` with `add_library(xgboost STATIC)` and remove `-fPIC` flags.  I think this is mostly for Windows users and I don't have access to a Windows machine for now, so will leave it to others who can test it.

hcho3:
Will keep track of progress on static library in #4483

PR #4333, opened 2019-04-04
C API example

canonizer:
- can be useful for #4240


RAMitchell:
Maybe just add a test to make sure we dont break it.. In our jenkins we could add a script test_demos.sh and start adding the other demos eventually too.

RAMitchell:
After looking at the build system, I think we need to wait for refactoring and add tests for the demos in a later PR with a little more thought.

PR #4340, opened 2019-04-08
Fix document about colsample_by* parameter

jokerkeny:
Correct the calculation mistake in colsample_by* example.


hcho3:
Thanks! merged.

PR #4343, opened 2019-04-08
 Combine thread launches into single launch per tree for gpu_hist

### LINKED ISSUE #4332: Deprecation warnings with CUDA 9 or newer
Compiling XGBoost with CUDA 9 or newer causes deprecation warnings because the deprecated warp intrinsic `__ballot` is used, e.g.
```
xgboost/src/tree/updater_gpu_hist.cu(581): warning: function "__ballot"
cuda_10.0.130/include/sm_20_intrinsics.h(405): here was declared deprecated ("__ballot() is deprecated in favor of __ballot_sync() and may be removed in a future release (Use -Wno-deprecated-declarations to suppress this warning).")

ptxas /tmp/tmpxft_0000c481_00000000-10_updater_gpu_hist.compute_50.ptx, line 15761; warning : Instruction 'vote' without '.sync' is deprecated since PTX ISA version 6.0 and will be discontinued in a future PTX ISA version
ptxas /tmp/tmpxft_0000c481_00000000-10_updater_gpu_hist.compute_50.ptx, line 17379; warning : Instruction 'vote' without '.sync' is deprecated since PTX ISA version 6.0 and will be discontinued in a future PTX ISA versio
``` 
Instead of `__ballot` `__ballot_sync` should be used: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-vote-functions 

Expectation is that XGBoost compile warning free with CUDA 9 and newer.

RAMitchell:
This PR contains major changes in the design of the gpu_hist algorithm. Previously we repeatedly launched groups of omp threads with one thread per GPU to execute some unit of work on the GPU. During a single tree construction there would be hundreds of thread launches. This can cause stalls as shown here:
multi-gpu-stall
This PR now launches one thread per GPU and this thread is used to construct the entire tree, only exiting when tree construction is finished. Each device shard still needs to communicate with other shards, but this is achieved via NCCL.

So we now no longer need to create and destroy many threads per iteration (or rely on the omp implementation to manage the thread pool efficiently) and we now avoid any unnecessary CPU thread synchronisation.

This PR also fixes #4332

Potential issues to look out for:
- Each device shard redundantly builds its own tree. If these trees were to somehow get out of sync, the algorithm would fail. Given that the histograms have been synced, I expect these to be identical. This potential issue also applies to the distributed GPU version even before this PR.
- Correct use of NCCL. I believe we have had some issues with deadlocks when using NCCL from multiple threads before. I have never experienced this myself. 

@canonizer @trivialfis can you please review.

### DIFF
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -12,6 +12,7 @@
 #include "span.h"
 
 #include <algorithm>
+#include <omp.h>
 #include <chrono>
 #include <ctime>
 #include <cub/cub.cuh>
@@ -1043,11 +1044,15 @@ class AllReducer {
 template <typename T, typename FunctionT>
 void ExecuteIndexShards(std::vector<T> *shards, FunctionT f) {
   SaveCudaContext{[&]() {
+    // Temporarily turn off dynamic so we have a guaranteed number of threads
+    bool dynamic = omp_get_dynamic();
+    omp_set_dynamic(false);
     const long shards_size = static_cast<long>(shards->size());
 #pragma omp parallel for schedule(static, 1) if (shards_size > 1)
     for (long shard = 0; shard < shards_size; ++shard) {
       f(shard, shards->at(shard));
     }
+    omp_set_dynamic(dynamic);
   }};
 }
 
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -777,8 +777,18 @@ class AllReducer {
                  allreduce_calls_(0) {}
 
   /**
-   * \fn  void Init(const std::vector<int> &device_ordinals)
-   *
+   * \brief If we are using a single GPU only
+   */
+  bool IsSingleGPU() {
+#ifdef XGBOOST_USE_NCCL
+    CHECK(device_counts.size() > 0) << "AllReducer not initialised.";
+    return device_counts.size() <= 1 && device_counts.at(0) == 1;
+#else
+    return true;
+#endif
+  }
+
+  /**
    * \brief Initialise with the desired device ordinals for this communication
    * group.
    *
@@ -954,6 +964,21 @@ class AllReducer {
       dh::safe_cuda(cudaStreamSynchronize(streams[i]));
     }
 #endif
+  };
+
+  /**
+   * \brief Synchronizes the device 
+   *
+   * \param device_id Identifier for the device.
+   */
+  void Synchronize(int device_id) {
+#ifdef XGBOOST_USE_NCCL
+    dh::safe_cuda(cudaSetDevice(device_id));
+    int idx = std::find(device_ordinals.begin(), device_ordinals.end(),device_id) - device_ordinals.begin();
+    CHECK(idx < device_ordinals.size());
+    dh::safe_cuda(cudaStreamSynchronize(streams[idx]));
+#endif
+
   };
 
 #ifdef XGBOOST_USE_NCCL
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -752,6 +752,29 @@ void Gather(int device_idx, T *out, const T *in, const int *instId, int nVals) {
                                        });
 }
 
+class SaveCudaContext {
+ private:
+  int saved_device_;
+
+ public:
+  template <typename Functor>
+  explicit SaveCudaContext (Functor func) : saved_device_{-1} {
+    // When compiled with CUDA but running on CPU only device,
+    // cudaGetDevice will fail.
+    try {
+      safe_cuda(cudaGetDevice(&saved_device_));
+    } catch (const dmlc::Error &except) {
+      saved_device_ = -1;
+    }
+    func();
+  }
+  ~SaveCudaContext() {
+    if (saved_device_ != -1) {
+      safe_cuda(cudaSetDevice(saved_device_));
+    }
+  }
+};
+
 /**
  * \class AllReducer
  *
@@ -973,12 +996,13 @@ class AllReducer {
    */
   void Synchronize(int device_id) {
 #ifdef XGBOOST_USE_NCCL
-    dh::safe_cuda(cudaSetDevice(device_id));
-    int idx = std::find(device_ordinals.begin(), device_ordinals.end(),device_id) - device_ordinals.begin();
-    CHECK(idx < device_ordinals.size());
-    dh::safe_cuda(cudaStreamSynchronize(streams[idx]));
+    SaveCudaContext([&]() {
+      dh::safe_cuda(cudaSetDevice(device_id));
+      int idx = std::find(device_ordinals.begin(), device_ordinals.end(), device_id) - device_ordinals.begin();
+      CHECK(idx < device_ordinals.size());
+      dh::safe_cuda(cudaStreamSynchronize(streams[idx]));
+    });
 #endif
-
   };
 
 #ifdef XGBOOST_USE_NCCL
@@ -1003,29 +1027,6 @@ class AllReducer {
     return id;
   }
 #endif
-};
-
-class SaveCudaContext {
- private:
-  int saved_device_;
-
- public:
-  template <typename Functor>
-  explicit SaveCudaContext (Functor func) : saved_device_{-1} {
-    // When compiled with CUDA but running on CPU only device,
-    // cudaGetDevice will fail.
-    try {
-      safe_cuda(cudaGetDevice(&saved_device_));
-    } catch (const dmlc::Error &except) {
-      saved_device_ = -1;
-    }
-    func();
-  }
-  ~SaveCudaContext() {
-    if (saved_device_ != -1) {
-      safe_cuda(cudaSetDevice(saved_device_));
-    }
-  }
 };
 
 /**
--- a/src/tree/updater_gpu_hist.cu
+++ b/src/tree/updater_gpu_hist.cu
@@ -621,6 +621,7 @@ template <typename GradientSumT>
 struct DeviceShard {
   int n_bins;
   int device_id;
+  int shard_idx;  // Position in the local array of shards
 
   dh::BulkAllocator ba;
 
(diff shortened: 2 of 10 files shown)


src/tree/updater_gpu_hist.cu
```diff
@@ -1243,277 +1446,41 @@ class GPUHistMakerSpecialised{
...
+    // Threads will cooperatively build the tree, synchronising over histograms.
+    // Each thread will redundantly
+    // build its own copy of the tree
+    dh::ExecuteIndexShards(
+        &shards_,
+        [&](int idx, std::unique_ptr<DeviceShard<GradientSumT>>& shard) {
+          // Only the first shard will actually modify the tree
+          RegTree* tree = idx == 0 ? p_tree : &dummy_trees.at(idx);
+          shard->UpdateTree(gpair, p_fmat, tree,
+                          &reducer_, column_sampler_);
```
canonizer: This means that each tree will start with the same state of the column sampler, which is probably something to avoid.

Also, copying of ColumnSampler in multiple threads may fail to work, as it uses `shared_ptr`, for which the copying is not thread-safe.

A better idea may be to pack the seed and colsample_by* parameters, and then pass them to the individual shards, so that they can initialize their column samplers.

src/tree/updater_gpu_hist.cu
```diff
@@ -649,18 +660,32 @@ struct DeviceShard {
...
       : device_id(_device_id),
+        shard_idx(shard_idx),
         row_begin_idx(row_begin),
         row_end_idx(row_end),
         n_rows(row_end - row_begin),
         n_bins(0),
         param(std::move(_param)),
-        prediction_cache_initialised(false) {}
+        prediction_cache_initialised(false) {
+    column_sampler.Init(num_columns, param.colsample_bynode,
```
canonizer: This initialization gets overwritten by the `Reset()` method.

src/tree/updater_gpu_hist.cu
```diff
@@ -1243,277 +1446,41 @@ class GPUHistMakerSpecialised{
       this->InitDataOnce(dmat);
       monitor_.StopCuda("InitDataOnce");
     }
-
-    column_sampler_.Init(info_->num_col_, param_.colsample_bynode,
```
canonizer: This ColumnSampler (in the GPUHistMakerSpecialised) never gets initialized. As the result, colsample_by* parameters are not passed to the shards.

src/tree/updater_gpu_hist.cu
```diff
@@ -649,18 +660,32 @@ struct DeviceShard {
...
 
+  using ExpandQueue =
+      std::priority_queue<ExpandEntry, std::vector<ExpandEntry>,
+                          std::function<bool(ExpandEntry, ExpandEntry)>>;
+  std::unique_ptr<ExpandQueue> qexpand;
+
   // TODO(canonizer): do add support multi-batch DMatrix here
-  DeviceShard(int _device_id, bst_uint row_begin, bst_uint row_end,
-              TrainParam _param)
+  DeviceShard(int _device_id, int shard_idx, bst_uint row_begin,
```
canonizer: It would be nice to have tests similar to https://github.com/dmlc/xgboost/blob/master/tests/distributed/distributed_gpu.py but testing that shards within a single process build the same model.

src/tree/updater_gpu_hist.cu
```diff
@@ -1243,277 +1446,41 @@ class GPUHistMakerSpecialised{
...
+    std::vector<RegTree> dummy_trees(shards_.size());
+    for (auto& tree : dummy_trees) {
+      tree = *p_tree;
     }
+    gpair->Reshard(dist_);
 
-    monitor_.StartCuda("FinalisePosition");
-    this->FinalisePosition(p_tree);
-    monitor_.StopCuda("FinalisePosition");
+    // Launch one thread for each device "shard" containing a subset of rows.
```
trivialfis: > If these trees were to somehow get out of sync

Is there a way to test this?  It can totally happen in the future since there are many parameters affecting tree construction.

RAMitchell:
Am still unsatisfied with the column sampling design, will try to come up with something better with tests in the next few days.

src/tree/updater_gpu_hist.cu
RAMitchell: I just added a check macro at runtime. I don't think there will be a performance penalty for something so small.

src/tree/updater_gpu_hist.cu
RAMitchell: Done.

src/tree/updater_gpu_hist.cu
```diff
@@ -715,7 +740,18 @@ struct DeviceShard {
   }
 
   // Reset values for each update iteration
-  void Reset(HostDeviceVector<GradientPair>* dh_gpair) {
+  // Note that the column sampler must be passed by value because it is not
```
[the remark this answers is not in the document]
RAMitchell: I added a basic test for column sampler using different threads, as well as checking the trees are equivalent after construction. I think this should cover this case.

RAMitchell:
@trivialfis @canonizer ready for a second review.

src/tree/updater_gpu_hist.cu
```diff
@@ -1264,277 +1469,46 @@ class GPUHistMakerSpecialised{
...
-                                   timestamp++));
-        qexpand_->push(ExpandEntry(right_child_nidx,
-                                   tree.GetDepth(right_child_nidx),
-                                   splits.at(1), timestamp++));
-        monitor_.StopCuda("EvaluateSplits");
-      }
+    // Check the shards produced identical trees
+    for (auto i = 1ull; i < dummy_trees.size(); i++) {
+      CHECK(*p_tree == dummy_trees.at(i));
     }
```
canonizer: Depending on the tree size and the number of GPUs, this may be quite an overhead. Also note that it runs serially for any number of GPUs. It also doesn't check that the trees are the same in a distributed setting.

I wonder if it is possible to put this check behind a parameter.

Also, is there a python test that runs with multiple GPUs in a single process?

tests/cpp/common/test_random.cc
```diff
@@ -45,7 +46,23 @@ TEST(ColumnSampler, Test) {
...
+TEST(ColumnSampler, ThreadSynchronisation) {
+  const int64_t num_threads = 10;
+  int seed = 7;
+  int n = 128;
+  std::vector<std::vector<int>> results(num_threads);
+#pragma omp parallel for schedule(static, 1)
+  for (int64_t i = 0; i < num_threads; ++i) {
+    ColumnSampler cs(seed);
+    cs.Init(n, 0.5f, 0.5f, 0.5f);
+    results.at(i) = cs.GetFeatureSet(0)->ConstHostVector();
```
canonizer: Consider testing several levels for several iterations here.

src/tree/updater_gpu_hist.cu
```diff
@@ -736,7 +761,16 @@ struct DeviceShard {
   }
 
   // Reset values for each update iteration
-  void Reset(HostDeviceVector<GradientPair>* dh_gpair) {
+  // Note that the column sampler must be passed by value because it is not
+  // thread safe
```
canonizer: This comment about passing the column sampler no longer applies.

src/common/device_helpers.cuh
```diff
@@ -752,6 +752,29 @@ void Gather(int device_idx, T *out, const T *in, const int *instId, int nVals) {
                                        });
 }
 
+class SaveCudaContext {
```
canonizer: What about calling it `WithDevice` and passing the new device number in? In this way, it can save the old device and switch to the new one, and then switch back in the destructor. Ideally, this would enable avoiding calling `cudaSetDevice` directly in xgboost code.

Also, what about making it a function instead of a class?

RAMitchell: I will leave this to @trivialfis

tests/cpp/common/test_random.cc
RAMitchell: Done

src/tree/updater_gpu_hist.cu
RAMitchell: I added a parameter for this and extended the check to distributed mode using rabit broadcast. The python multi-GPU tests will cover the case of a single process with multiple GPUs.

src/common/device_helpers.cuh
trivialfis: Ideally,  we should avoid changing device in master thread.  So maybe working toward that end would be better.

trivialfis:
Seems the added tests paid off.  Will review once tests passed.

src/tree/updater_gpu_hist.cu
```diff
@@ -38,6 +38,7 @@ struct GPUHistMakerTrainParam
   bool single_precision_histogram;
   // number of rows in a single GPU batch
   int gpu_batch_nrows;
+  bool debug_synchronize;
```
trivialfis: Not to cause you more troubles.  But can we use existing `verbosity:debug` flag?  We have `debug_verbose` before and somehow people take it as part of the interface.  Even though I deprecated by introducing `verbosity` parameter, I'm still seeing its use in issues.  Be careful when adding flags like these...

trivialfis: You can use `ConsoleLogger::ShouldLog`, see its use in `timer.h::Monitor` for an example.

RAMitchell:
@trivialfis @canonizer I had this working and then it suddenly stopped working - after two days I finally realised the cmake refactor had disabled openmp in all cuda code. The new cmake native support does not automatically propagate host compiler flags to nvcc via -Xcompiler, contrary to the old FindCuda.cmake. Just something to be aware of in future.

src/tree/updater_gpu_hist.cu
RAMitchell: I don't want this to be a part of verbose due to the hidden performance penalty. This flag may not be a permanent fixture.

trivialfis:
@RAMitchell Any update?  If you still have pending work on this PR then I can extract the CMake fix out.

PR #4349, opened 2019-04-09
[BREAKING][jvm-packages] fix the non-zero missing value handling

CodingCat:
it's a successive PR for https://github.com/dmlc/xgboost/commit/ad4de0d718e33b98d9ae32bb3858a57a989ef5f8

Spark's vector assembler transformer only accepts 0 as the missing value, which creates problems when the user takes 0 as the meaningful value and there are enough number of 0-valued features leading vector assembler to use SparseVector to represent the feature vector. The reason is that those 0-valued features has been filtered out by Spark. 

It's fine if the user's DataFrame only contains DenseVector, as the 0 is kept.

This PR  changes the behavior of XGBoost-Spark as:

when the user sets a non-zero missing value and we detect there is sparse vector, we stop the application to prevent error

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -18,9 +18,7 @@ package ml.dmlc.xgboost4j.scala.spark
 
 import java.io.File
 import java.nio.file.Files
-import java.util.Properties
 
-import scala.collection.mutable.ListBuffer
 import scala.collection.{AbstractIterator, mutable}
 import scala.util.Random
 
@@ -32,8 +30,8 @@ import org.apache.commons.io.FileUtils
 import org.apache.commons.logging.LogFactory
 
 import org.apache.spark.rdd.RDD
-import org.apache.spark.{SparkContext, SparkException, SparkParallelismTracker, TaskContext}
-import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.{SparkContext, SparkParallelismTracker, TaskContext}
+import org.apache.spark.sql.SparkSession
 import org.apache.spark.storage.StorageLevel
 
 
@@ -75,8 +73,8 @@ object XGBoost extends Serializable {
     if (missing != 0.0f) {
       xgbLabelPoints.map(labeledPoint => {
         if (labeledPoint.indices != null) {
-            throw new RuntimeException("you can only specify missing value as 0.0 when you have" +
-              " SparseVector as your feature format")
+            throw new RuntimeException(s"you can only specify missing value as 0.0 (set value" +
+              s" $missing) when you have SparseVector or Empty vector as your feature format")
         }
         labeledPoint
       })
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -105,7 +105,8 @@ object XGBoost extends Serializable {
       removeMissingValues(verifyMissingSetting(xgbLabelPoints, missing),
         missing, (v: Float) => v != missing)
     } else {
-      removeMissingValues(xgbLabelPoints, missing, (v: Float) => !v.isNaN)
+      removeMissingValues(verifyMissingSetting(xgbLabelPoints, missing),
+        missing, (v: Float) => !v.isNaN)
     }
   }
 
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -73,8 +73,9 @@ object XGBoost extends Serializable {
     if (missing != 0.0f) {
       xgbLabelPoints.map(labeledPoint => {
         if (labeledPoint.indices != null) {
-            throw new RuntimeException(s"you can only specify missing value as 0.0 (set value" +
-              s" $missing) when you have SparseVector or Empty vector as your feature format")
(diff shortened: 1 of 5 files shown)


CodingCat:
@yanboliang the PR is changed to this

superbobry:
I'm sorry, I'm not an active user of XGBoost any more, so I don't feel confident reviewing this. @alois-bissuel, do you want to have a look?

alois-bissuel:
A small comment on the aim of the review, as I did not have enough time to look at the code in details (and won't have time before next Tuesday).
I might be mistaken, but I think using `SparseVector` is a way (without specifying a special number which signals a missing value) to correctly handle missing value. 
There is a specific use of it if one has only integer (and obviously ordinal) features (there is no NaN for ints).
I also looked at the code of `VectorAssembler` in Spark MLlib, and it seems that it correctly handles null features only since Spark v2.4 (see [VectorAssembler.scala#L282](https://github.com/apache/spark/blob/7a8efc8fe32abb51a5c8803c6afceadf838e46e6/mllib/src/main/scala/org/apache/spark/ml/feature/VectorAssembler.scala#L282) in Spark 2.4).
So after this review, there should be no other way of handling missing values than using a special value (either Nan or set it through the parameter "missing")

CodingCat:
@alois-bissuel thanks for the review

and to clarify the point of the PR and help the further review, this PR is to prevent the following case happening

* user sets parameter `missing` as -1 (anything other than 0 and NaN) and use 0 to represent some meaningful feature

* vector assembler will filter those 0 values https://github.com/apache/spark/blob/7a8efc8fe32abb51a5c8803c6afceadf838e46e6/mllib/src/main/scala/org/apache/spark/ml/feature/VectorAssembler.scala#L262

* vector assembler decided to use sparse vector to represent that feature https://github.com/apache/spark/blob/branch-2.4/mllib-local/src/main/scala/org/apache/spark/ml/linalg/Vectors.scala#L165-L173

* then we lost those 0-valued features in step 3 and also filter -1 in XGBoost-Spark

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/MissingValueHandlingSuite.scala
```diff
@@ -0,0 +1,147 @@
...
+      "objective" -> "binary:logistic", "missing" -> -1.0f, "num_workers" -> 1).toMap
+    val model = new XGBoostClassifier(paramMap).fit(inputDF)
+    model.transform(inputDF).collect()
+  }
+
+  test("specify a non-zero missing value and meet an empty vector we should" +
+    " stop the application") {
+    val spark = ss
+    import spark.implicits._
+    val testDF = Seq(
```
alois-bissuel: Maybe a small comment on why the vector is empty would be nice here

CodingCat: thx, added

CodingCat:
@hcho3 I think windows test is broken somehow?

hcho3:
@CodingCat I’ll take a look

CodingCat:
merged, thanks for the review @alois-bissuel

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -107,7 +106,8 @@ object XGBoost extends Serializable {
       removeMissingValues(verifyMissingSetting(xgbLabelPoints, missing),
         missing, (v: Float) => v != missing)
     } else {
-      removeMissingValues(xgbLabelPoints, missing, (v: Float) => !v.isNaN)
+      removeMissingValues(verifyMissingSetting(xgbLabelPoints, missing),
```
rongou: Is this change correct? In the `else` clause `missing.isNaN` is true, yet `verifyMissingSetting()` would throw an exception. Seems like the `else` clause would always throw an exception. Is that expected?

CodingCat: yes, it's expected 

if you have NaN as missing value, that means you think 0 is a meaningful value in your dataset, however, vectorAssembler has produced a sparsevector for some of your rows (by filtering 0, since vectorassembler only thinks 0 as missing value)

instead of proceeding with this incompatible issue between XGBoost/VectorAssembler, we have to stop the training to avoid messing up the accuracy

else clause does not always throw exception, because when you have every row as dense vector, it would exceed (because we didn't find any 0 filtered by vector assembler...)

cpfarrell:
This is coming in pretty late but would it be possible to add a parameter option that would avoid this check? It's possible to construct a SparseVector yourself that includes zeros in the SparseVector and has the values be missing from it indicate a missing value. Doing this allows you to solve the problem mentioned in the "Note" at https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html#dealing-with-missing-values. I understand the reason for adding this check as most users will have used the VectorAssembler class to construct their feature vector but if you have handled that separately it seems like it would be nice to be able to specify a parameter to the model saying essentially "I know what I'm doing" and have an easy way of handling feature sets that have both missing values and zeros.

If this isn't the right place to be asking this question I'd be happy to file a separate feature request ticket. I'd also be happy to work on this myself if open to the change.

CodingCat:
> This is coming in pretty late but would it be possible to add a parameter option that would avoid this check? It's possible to construct a SparseVector yourself that includes zeros in the SparseVector and has the values be missing from it indicate a missing value. Doing this allows you to solve the problem mentioned in the "Note" at https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html#dealing-with-missing-values. I understand the reason for adding this check as most users will have used the VectorAssembler class to construct their feature vector but if you have handled that separately it seems like it would be nice to be able to specify a parameter to the model saying essentially "I know what I'm doing" and have an easy way of handling feature sets that have both missing values and zeros.
> 
> If this isn't the right place to be asking this question I'd be happy to file a separate feature request ticket. I'd also be happy to work on this myself if open to the change.

feel free to open a PR

Issue #4342, opened 2019-04-08
[jvm-packages] XGBoost4j can't load the native library on Java 12


Craigacp:
XGBoost fails to load the JNI library when running on Java 12. It loads with a warning on Java 9 through 11.

In 12 the ability to modify fields on the ClassLoader was turned off - https://bugs.openjdk.java.net/browse/JDK-8210522. And so in 12 you get:
```
[04/08/19 16:59:50:371] SEVERE NativeLibLoader.addNativeDir usr_paths
[04/08/19 16:59:50:371] SEVERE DMatrix.<clinit> load native library failed.
[04/08/19 16:59:50:372] SEVERE DMatrix.<clinit> java.io.IOException: Failed to get field handle to set library path
Exception in thread "main" java.lang.UnsatisfiedLinkError: ml.dmlc.xgboost4j.java.XGBoostJNI.XGDMatrixCreateFromCSREx([J[I[FI[J)I
	at ml.dmlc.xgboost4j.java.XGBoostJNI.XGDMatrixCreateFromCSREx(Native Method)
	at ml.dmlc.xgboost4j.java.DMatrix.<init>(DMatrix.java:122)
```

whereas in 11 it prints:
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by ml.dmlc.xgboost4j.java.NativeLibLoader (file:/local/Repositories/Tribuo/Classification/XGBoost/target/mlrg-classification-xgboost-3.0.0-SNAPSHOT-jar-with-dependencies.jar) to field java.lang.ClassLoader.usr_paths
WARNING: Please consider reporting this to the maintainers of ml.dmlc.xgboost4j.java.NativeLibLoader
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```
and execution continues. In Java 8 it executes without a warning.

CodingCat:
yes...expected, xgboost4j's jvm support is essentially following Spark's......

Craigacp:
If I can work out a fix that doesn't break Java 8 will you accept it?

CodingCat:
> If I can work out a fix that doesn't break Java 8 will you accept it?

more than happy

Craigacp:
[...]

PR #4351, opened 2019-04-09
[jvm-packages] Fixing the NativeLibLoader on Java 9+

Craigacp:
Fix for issue #4342.

The old NativeLibLoader had a short-circuit load path which modified java.library.path and attempted to load the xgboost library from outside the jar first, falling back to loading the library from inside the jar. This path is a no-op every time when using XGBoost outside of it's source tree. Additionally it triggers an illegal reflective access warning in the module system in 9, 10, and 11.

On Java 12 the ClassLoader fields are not accessible via reflection (separately from the illegal reflective acces warning), and so it fails in a way that isn't caught by the code which falls back to loading the library from inside the jar.

This commit removes that code path and always loads the xgboost library from inside the jar file as it's a valid technique across multiple JVM implementations and works with all versions of Java.

I tested it on macOS Mojave using JDK 11 & 12, Linux using JDK 8, 11 and 12, and Windows using an older build using JDK 9 and 12.

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java
@@ -30,14 +30,19 @@ class NativeLibLoader {
   private static final Log logger = LogFactory.getLog(NativeLibLoader.class);
 
   private static boolean initialized = false;
-  private static final String nativePath = "../../lib/";
   private static final String nativeResourcePath = "/lib/";
   private static final String[] libNames = new String[]{"xgboost4j"};
 
   static synchronized void initXGBoost() throws IOException {
     if (!initialized) {
       for (String libName : libNames) {
-        smartLoad(libName);
+        try {
+          String libraryFromJar = nativeResourcePath + System.mapLibraryName(libName);
+          loadLibraryFromJar(libraryFromJar);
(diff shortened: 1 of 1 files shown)


CodingCat:
LGTM, let's merge it when Travis is ready and after @hcho3 bring up the java worker, we can have test across jdk versions there...(don't need to work on Travis now which looks like to be eventually discarded? @hcho3 )

CodingCat:
@hcho3 just to confirm before I merge it, Travis will eventually gone?

hcho3:
@CodingCat Sort of. We may want to keep a few tests targeted for Mac however.

CodingCat:
@hcho3 got it, thx!

@Craigacp do you mind adding some profile in pom.xml which only compiles xgboost4j module, and add some cross jdk test in travis then?

hcho3:
@CodingCat @Craigacp I'm not sure if cross-jdk test is needed on Travis, since Travis will only have Mac targets after migration.

CodingCat:
@hcho3 we can only do that in linux/windows, but we still need some profile for only compiling xgboost4j now or in near future 

so I don't mind merging it now

hcho3:
@CodingCat Yes, let's merge this PR.

Craigacp:
Do you still want me to make the pom.xml profile for xgboost4j only before merging?

CodingCat:
@Craigacp no, thanks! we can merge it now!

PR #4352, opened 2019-04-09
[rabit harden] use dmlc-core in rabit

chenqin:
rabit already includes dmlc-core submodule while in xgboost, dmlc-core project can use rabit dmlc-core project instead.

apply dmlc/rabit after https://github.com/dmlc/rabit/pull/90 merged

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
-[submodule "dmlc-core"]
-	path = dmlc-core
-	url = https://github.com/dmlc/dmlc-core
 [submodule "rabit"]
 	path = rabit
-	url = https://github.com/dmlc/rabit
+	url = https://github.com/chenqin/rabit
 [submodule "cub"]
 	path = cub
 	url = https://github.com/NVlabs/cub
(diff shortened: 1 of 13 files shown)


RAMitchell:
Not sure if I agree with this change. What if we want to update the dmlc core version in xgboost but can't because of compatibility issues in rabit? What is the downside of having each library have their own version of dmlc core?

hcho3:
@RAMitchell I am also wary as well. Sometimes, a change in XGBoost necessitates change in dmlc-core too.

chenqin:
> Not sure if I agree with this change. What if we want to update the dmlc core version in xgboost but can't because of compatibility issues in rabit? What is the downside of having each library have their own version of dmlc core?

I see your point @RAMitchell @hcho3 Having a always master branch compatible dmlc-core which gated by rabit update version was the goal. In fact, I started to add xgb regression tests on rabit just to make sure rabit always able to run xgboost. 

In another venue, I don't think having separate dmlc-core sub-modules in both xgboost as well as rabit sounds like a big issue. Likely there are cases having to update dmlc-core due to a bug fix or urgent feature directly will be more preferable than go through yet another project.

So I plan to just remove duplicated includes from rabit/dmlc-core to upgrade rabit in xgboost.

PR #4358, opened 2019-04-11
refactor tests to get rid of duplication

rongou:
@RAMitchell @hcho3

### DIFF
--- a/tests/cpp/helpers.cc
+++ b/tests/cpp/helpers.cc
@@ -148,8 +148,8 @@ std::unique_ptr<DMatrix> CreateSparsePageDMatrix() {
   dmlc::TemporaryDirectory tempdir;
   const std::string tmp_file = tempdir.path + "/big.libsvm";
   CreateBigTestData(tmp_file, 12);
-  DMatrix * dmat = DMatrix::Load(
-      tmp_file + "#" + tmp_file + ".cache", true, false, "auto", 64UL);
+  std::unique_ptr<DMatrix> dmat = std::unique_ptr<DMatrix>(DMatrix::Load(
+      tmp_file + "#" + tmp_file + ".cache", true, false, "auto", 64UL));
   EXPECT_TRUE(FileExists(tmp_file + ".cache.row.page"));
 
   // Loop over the batches and count the records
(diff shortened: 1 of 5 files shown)


tests/cpp/helpers.cc
```diff
@@ -142,4 +143,38 @@ std::shared_ptr<xgboost::DMatrix>* CreateDMatrix(int rows, int columns,
   return static_cast<std::shared_ptr<xgboost::DMatrix> *>(handle);
 }
 
+std::unique_ptr<DMatrix> CreateSparsePageDMatrix() {
+  // Create sufficiently large data to make two row pages
+  dmlc::TemporaryDirectory tempdir;
+  const std::string tmp_file = tempdir.path + "/big.libsvm";
+  CreateBigTestData(tmp_file, 12);
+  DMatrix * dmat = DMatrix::Load(
+      tmp_file + "#" + tmp_file + ".cache", true, false, "auto", 64UL);
```
hcho3: Let's wrap pointer in the smart pointer immediately, for the sake of exception safety.
```cpp
std::unique_ptr<DMatrix> dmat = std::unique_ptr<DMatrix>(DMatrix::Load(
      tmp_file + "#" + tmp_file + ".cache", true, false, "auto", 64UL));
```

tests/cpp/helpers.cc
```diff
@@ -142,4 +143,38 @@ std::shared_ptr<xgboost::DMatrix>* CreateDMatrix(int rows, int columns,
...
+  int64_t batch_count = 0;
+  int64_t row_count = 0;
+  for (const auto &batch : dmat->GetRowBatches()) {
+    batch_count++;
+    row_count += batch.Size();
+  }
+  EXPECT_EQ(batch_count, 2);
+  EXPECT_EQ(row_count, dmat->Info().num_row_);
+
+  return std::unique_ptr<DMatrix>(dmat);
```
hcho3:
```suggestion
  return dmat;
```

tests/cpp/helpers.cc
rongou: Done.

tests/cpp/helpers.cc
rongou: Done.

rongou:
@RAMitchell @hcho3 do you know why the python test is failing? This PR only changes c++ tests, so shouldn't affect python tests at all, right?

hcho3:
@rongou The failure is due to out of memory error.
I am taking a day off tomorrow to sort out CI issues. In particular, all Linux tests should be moved to Jenkins to use machines with bigger memory.

PR #4365, opened 2019-04-13
add a new predict method: predictPackage

sperlingxx:
Add a new predict method: **predictPackage**, which regard various kinds of predictions as a package.
By **predictPackage**,  we can get leaf indices, scores(margin values) and final results(transformed values) at once, which avoid unnecessary repeated tree searching.
I will add jvm and python support later.


trivialfis:
Closing for now.

Issue #4350, opened 2019-04-09
[jvm-packages] cross-version spark support 


CodingCat:
@hcho3 I am going to work on support spark 2.4.1 and have compatibility test over spark 2.3

my plan is to trigger two builds for spark 2.4/2.3 respectively and also have version-specific test to ensure the compatibility

shall I wait for java worker to be ready in jenkins or I should work on travis?

hcho3:
Let me add Java workers to Jenkins. Can you provide commands to compile JARs?

CodingCat:
yes, just `mvn package`

srowen:
That's great @CodingCat -- would be great to get a 2.4.x build going as 2.3.x is EOL in a few months. I suspect you have this well in hand, but if you're hitting weird problems updating to 2.4 (shouldn't be much) I'd be happy to try to debug.

CodingCat:
I would limit the definition of cross-version support to "support loading models trained in previous version" in XGBoost

I have done several experiments on running a spark-2.4-built xgboost with spark 2.3 or vice versa. The most significant problem is from the library which Spark depends on and brings some breaking changes by their own. In that way we cannot guarantee a spark 2.4 built version can be run with spark 2.3 runtimes 

even to "support loading models trained in previous version", we need some code to handle (1) breaking changes in XGBoost parameters, e.g. reg:linear doesn't exist anymore; (2) breaking changes in Spark, e.g. vectorAssembler will fail with Float.NaN by default...

srowen:
Personally, I'd relax that condition if needed -- require the same version of xgboost / Spark to read/write things.

Issue #3986, opened 2018-12-11
Tracking deprected features.


trivialfis:
|  Feature                                                 | PR(or time)|                                                             | Removed |
|----------------------------------------------------------|------------|-------------------------------------------------------------|----|
| GPU Objectives.                                          | #3643      | Now all objectives having GPU port will use GPU by default. | ✘  |
| `grow_fast_histmaker`                                    | #3836      | Now is renamed to `grow_quantile_histmaker`.                | ✘  |
| Some of R parameters documented in `R/xgb.importance.R`. | #1964      | ??                                                          | ?? |
| `GBTreeModelParam::num_pbuffer_deprecated`               | ??         | ??                                                          | ✘  |
| `python/training.py: train::learning_rates`              | #1797      | Now use use callback API.                                   | ✘  |
| `python/sklearn.py: seed`                                | ??         | Now use random_state.                                       | ✘  |
| `python/sklearn.py: nthread`                             | ??         | Now use n_jobs.                                             | ✘  |
| `XGDMatrixCreateFromCSC`                                 | #1600      | ??                                                          | ✘  |
| `XGDMatrixCreateFromCSR`                                 | #1600      | `XGDMatrixCreateFromCSR_omp`                                | ✘  |
| `DMatrix::SaveToLocalFile`                               | ?? In 2015 | ??                                                          | ✘  |
|                                                          |            |                                                             |    |
|                                                          |            |                                                             |    |

trivialfis:
Inspried by @thvasilo in #3929.(Thanks)  Lets keep track of deprecated features here.  Great chance that I missed something, anyone who notice please help adding it back. :)  @hcho3  Please help filling the `??` if it's still possible (please don't spend too much time on it, it's deprecated), otherwise I will summit a PR to remove everything that loss track.

hcho3:
Let's discuss whether `num_roots` should be kept or not. I think it is not documented and experimental and should be deprecated. Already, I decided not to support it when I first wrote `hist` updater.

trivialfis:
@hcho3 I tried to find out what it does.  It turns out `num_roots` is not documented.  Is there any reference for it?

tqchen:
I think it is fine dropping them, but let us also keep the backward compatibility in mind. It might be useful to first introduce json exchange format, and then allow upgrade via json

hcho3:
Just found out that ``TreeParam.max_depth`` is never set and never used:
[...]

PR #4377, opened 2019-04-16
[jvm-packages] support spark 2.4 and compatibility test with previous xgboost version

CodingCat:
closes https://github.com/dmlc/xgboost/issues/4350


jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/DefaultXGBoostParamsReader.scala
```diff
@@ -22,12 +22,17 @@ import org.json4s.JsonAST.JObject
...
+import org.apache.spark.ml.param.{Param, Params}
 import org.apache.spark.ml.util.MLReader
 
 // This originates from apache-spark DefaultPramsReader copy paste
 private[spark] object DefaultXGBoostParamsReader {
 
+  private val paramNameCompatibilityMap: Map[String, String] = Map("silent" -> "verbosity")
+
+  private val paramValueCompatibilityMap: Map[String, Map[Any, Any]] =
+    Map("objective" -> Map("reg:linear" -> "reg:squarederror"))
```
CodingCat: @trivialfis @hcho3  do we have other deprecated or brokenly changed parameters?

jvm-packages/pom.xml
```diff
@@ -34,7 +34,7 @@
         <maven.compiler.source>1.7</maven.compiler.source>
         <maven.compiler.target>1.7</maven.compiler.target>
         <flink.version>1.5.0</flink.version>
-        <spark.version>2.3.3</spark.version>
+        <spark.version>2.4.1</spark.version>
```
hcho3: @CodingCat Is it still possible to compile XGBoost4J with Spark 2.3.x with this change?

CodingCat: if you skip tests, yes, 

if you want to run the test, no, because the API in https://github.com/dmlc/xgboost/pull/4377/files#diff-158851aba72186f53a0ff66dd4a792c8R264 does not exist in Spark 2.3, and we have to add in Spark 2.4 to ensure that Spark can handle NaN as expected

hcho3: @CodingCat Is there a way to mask tests not compatible with Spark 2.3.x? I don't think we want to suddenly drop support for Spark 2.3.x

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/DefaultXGBoostParamsReader.scala
hcho3: @CodingCat See #3986

jvm-packages/pom.xml
CodingCat: I don't think we can mask a single test. The way to skip test is to do something like `mvn package -DskipTests` which skips the compilation of test classes. Once you compile them, e.g. with `mvn package -Dspark.version=2.3.3`, it will fall into cannot find symbol error (since the API does not exist)

According to @srowen from Databricks, the community is going to drop 2.3.x soon, my personal thinking is to move forward (as we have stay in 2.3.x for at least two versions, further, we are going to bring some breaking changes https://github.com/dmlc/xgboost/pull/4349)

hcho3: @CodingCat So I should just use Spark 2.4.1 in the new CI pipeline (will file PR today)?

CodingCat: yes, thanks

hcho3: Nice, one fewer things to worry about

srowen: Yes, Spark 2.3.x will be technically EOL in August. If you want my opinion: I'd keep it simple and forego Spark compatibility with 2.3 if that helps.

That said, what's the runtime issue? maybe I have an easier workaround. There _shouldn't_ be breaking changes between 2.3 and 2.4 (not in the public API) but it is possible either by mistake or exception. If it's that a deprecated method has gone away, maybe there are alternatives.

hcho3: @srowen Mainly the issue is that one of the test being added in this PR will not be compatible with Spark 2.3.x. As @CodingCat said, it is not possible to exclude a single test when running `mvn package`.

srowen: Can the test simply check the Spark version and skip it if it's not Spark 2.3? IIRC you can check `org.apache.spark.SPARK_VERSION` for example.

srowen: ... or if it's actually a compile-time incompatibility (what is it?) you _could_ make a separate test source tree that is only activated by a profile or specific spark.version in the build. Not sure if it's worth it or what.

CodingCat: it's a compilation-time change this line in https://github.com/dmlc/xgboost/pull/4377/files#diff-158851aba72186f53a0ff66dd4a792c8R264 just doesn't compile in 2.3 

and for breaking changes in 2.3-2.4, the parse() method https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/DefaultXGBoostParamsWriter.scala#L73 in jackson is brokenly changed in two versions depended by 2.3/2.4....

I don't think 2.3 is that important to us given it's EOL soon

srowen: OK sounds good. You could work around the first one with reflection maybe, but working around Jackson differences is harder. Well you have my vote for just making the next version focus on Spark 2.4.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/DefaultXGBoostParamsReader.scala
CodingCat: thx, looks like these are the only two params needing special cares

Issue #4379, opened 2019-04-16
[VOTE] Ending Python 2.x support in next release (0.83)


hcho3:
Given that Python 2 will reach its end of life at the end of this year, we should drop Python 2 support. That is, contributors would no longer be required to write Python code that's compatible with Python 2.7. Also, Python distributions of XGBoost on PyPI will be Python 3 exclusive.

Indicate your support by replying "+1" to this thread. I want to get 4-5 votes from the committers to proceed with this decision. (Others, feel free to voice support with +1 reply as well, although only committers will have binding votes.)

@dmlc/xgboost-committer

terrytangyuan:
Definitely +1

trivialfis:
+1.  

We need to upgrade Python on test farm then.

RAMitchell:
+1

CodingCat:
+1

hcho3:
+1

pommedeterresautee:
+1

tqchen:
+1

hcho3:
With 7 votes from committers, it is decided that **the upcoming release (0.83) will support Python 3 exclusively.** I filed #4381 to migrate pylint to Python 3, and #4380 will use Python 3 exclusively for all CI tests. Thanks!

CAM-Gerlach:
Given the major status of this project and the major importance of this project to the scientific Python community, you should consider adding your name and logo to [Python3Statement.org](https://python3statement.org/) of projects planning to drop support for Python 2 on or before Jan 1 2020, to maximize public visibility of your support timeline in a central location, and help send the message that remaining Python 2 users should start transitioning now to minimize the pain later. To add XGBoost, you can just [submit a simple PR](https://github.com/python3statement/python3statement.github.io/pulls) to the Python3Statement repo with your name and logo—a SVG or transparent PNG of this would be ideal:

![image](https://user-images.githubusercontent.com/17051931/57117419-e8e63b00-6d29-11e9-90d9-a09e931b7cfe.png)

Or, if you like, I can create a PR for you there and @hcho3 can sign off, if you want. Thanks!

hcho3:
@CAM-Gerlach Yes, we'd love to put our logo there. Can you go ahead and create a PR on our behalf? I'll sign it off.

CAM-Gerlach:
@hcho3 Sure; I'll go ahead and tag you there. Thanks!

CAM-Gerlach:
[...]

PR #4380, opened 2019-04-17
[CI] Refactor Jenkins CI pipeline + migrate tests to Jenkins

hcho3:
See #4234 for details.

Overview of Jenkins CI pipeline:
Screen Shot 2019-04-16 at 8 17 43 PM

PR #4381, opened 2019-04-17
Migrate pylint check to Python 3

hcho3:
As we are about to deprecate Python 2 (#4379), we need to migrate pylint (Python coding style check) to Python 3. There are some minor changes to meet the new requirements.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -1,6 +1,6 @@
 # coding: utf-8
 # pylint: disable=too-many-arguments, too-many-branches, invalid-name
-# pylint: disable=too-many-branches, too-many-lines, W0141
+# pylint: disable=too-many-branches, too-many-lines, too-many-locals
 """Core XGBoost Library."""
 from __future__ import absolute_import
 import collections
@@ -30,7 +30,6 @@ c_bst_ulong = ctypes.c_uint64
 
 class XGBoostError(Exception):
     """Error thrown by xgboost trainer."""
-    pass
 
 
 class EarlyStopException(Exception):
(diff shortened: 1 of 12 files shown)


python-package/xgboost/core.py
```diff
@@ -1721,7 +1714,7 @@ def get_split_value_histogram(self, feature, fmap='', bins=None, as_pandas=True)
         xgdump = self.get_dump(fmap=fmap)
         values = []
         regexp = re.compile(r"\[{0}<([\d.Ee+-]+)\]".format(feature))
-        for i in range(len(xgdump)):
+        for i, _ in enumerate(xgdump):
```
trivialfis: Just out of curiosity, why using `enumerate`.

python-package/xgboost/training.py
```diff
@@ -462,7 +462,7 @@ def cv(params, dtrain, num_boost_round=10, nfold=3, stratified=False, folds=None
                                rank=0,
                                evaluation_result_list=res))
         except EarlyStopException as e:
-            for k in results.keys():
+            for k in results:
```
trivialfis: Does this change anything related to code logic?

hcho3: No, the change is to comply with the latest pylint guideline (PyCQA/pylint#699). Logic-wise nothing changed.

python-package/xgboost/core.py
hcho3: To comply with pylint style guideline: PyCQA/pylint#684

hcho3:
@trivialfis I think this is ready. I'd like to get this merged soon, since #4380 depends on it.

Issue #4307, opened 2019-03-28
[jvm-packages] xgboost4j-spark Prediction Optimization


mingyang:
# Motivation
Recently, I've been testing whether xgboost4j-spark could be a viable solution in production. One problem surfaced when *multiple models* were applied to the same dataset in sequence. 

In an industrial setting, it's very likely that a practitioner builds multiple models using the same dataset, with same features but different labels for different tasks. For example, one might want to predict the demographic status of her users (gender, age, marital status, etc.) with the same set of base features.

# Ideal Situation
Once multiple models have been trained on the labeled dataset, one wants to make predictions on a (potentially) much bigger unlabeled dataset. In pseudo code, it looks like 
```
for model in trained_models; do
  data = model.transform(data).withColumnRenamed('prediction', 'prediction_from_model_i')
data.writeoutput
```
Ideally, this only loops through the dataset *ONCE* while applying all models.

# Current Observations
## In a toy setup
I've tried to apply 4 models, using xgboost4j-spark and spark logistic regression (LR), to a small dataset and monitored the IO. What's far from ideal is when applying multiple xgboost4j-spark models, the number of input records grows *exponentially* with the number of models applied:

model|#models|#input|prediction_time
-|-|-|-
xgboost4j-spark|0|11952 (size of the data)|
xgboost4j-spark|1|23904|6
xgboost4j-spark|2|47808|9
xgboost4j-spark|3|95616|22
xgboost4j-spark|4|192002|23
spark-LR|0|10912 (size of the data)|
spark-LR|1|10912|12
spark-LR|2|10912|15
spark-LR|3|10912|15
spark-LR|4|10912|16

In other words, if there are N models, then spark will read the dataset 2^N times. On the contrary, native spark LR simply reads the dataset *once* no matter how many models applied.

Even when just applying one model, the current implementation requires *2X of the dataset*.

## In a more realistic setup
This is where I first noticed the problem: 
- Predicting using one model reads in ~1TB data (again 2X of the dataset size), while applying 8 models together read in 127.4 TB (~2^8 X of the dataset size). 
- Prediction time went up from 6-7mins using one model to 7-9 hours using all eight, as compared to 7 X 8 ~ 1 hour if we do predictions separately in sequence.

# Suspected Reasons
[...]

PR #4388, opened 2019-04-20
[BLOCKING][jvm-packages] fix non-deterministic order within a partition  (in the case of an upstream shuffle) on prediction 

### LINKED ISSUE #4387: [jvm-packages] bug of XGBoostModel.transformInternal
ZipPartitions action at XGBoostModel.transformInternal will cause a mismatch between originalColumns and resultColumns.

`val resultRDD = inputRDD.zipPartitions(predictionRDD, preservesPartitioning = true) {
 `

In here, inputRDD and predictionRDD will be evaluated separated. Co-dependencies of these two RDDs will be computed twice, which may produce different row orders inside each partition if previous actions are not strictly keeping the same order (like ShuffleRead).

There is a tiny test below, which will fail because of column mismatching : 

> 

    val source = ss.sparkContext.parallelize((1 to 20).toList)
    val rnd = source.mapPartitions { iter =>
      iter.map(x => x -> Random.nextDouble())
        .toList
        .sortBy(_._2)
        .iterator
    }
    val negative = rnd.map { case (i, x) => i -> -x }
    val zipped = rnd.zip(negative)
    zipped.collect().foreach { case (x, y) =>
      x._1 shouldEqual y._1
    }

sperlingxx:
fix issue #4387 by replacing zip RDDs with caching original data in closure.

Corresponding unit tests have been added.

closes #4387 

closes https://github.com/dmlc/xgboost/issues/4307

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
@@ -290,6 +290,10 @@ class XGBoostClassificationModel private[ml](
 
         private val batchIterImpl = rowIterator.grouped(
           XGBoostClassificationModel.PREDICTION_BATCH_SIZE).flatMap { batchRow =>
+<<<<<<< HEAD
+=======
+
+>>>>>>> regressor impl
           if (batchCnt == 0) {
             val rabitEnv = Array("DMLC_TASK_ID" -> TaskContext.getPartitionId().toString).toMap
             Rabit.init(rabitEnv.asJava)
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
@@ -287,14 +287,20 @@ class XGBoostClassificationModel private[ml](
     val bBooster = dataset.sparkSession.sparkContext.broadcast(_booster)
     val appName = dataset.sparkSession.sparkContext.appName
 
-    val inputRDD = dataset.asInstanceOf[Dataset[Row]].rdd
-    val predictionRDD = dataset.asInstanceOf[Dataset[Row]].rdd.mapPartitions { rowIterator =>
+    val resultRDD = dataset.asInstanceOf[Dataset[Row]].rdd.mapPartitions { rowIterator =>
       if (rowIterator.hasNext) {
         val rabitEnv = Array("DMLC_TASK_ID" -> TaskContext.getPartitionId().toString).toMap
         Rabit.init(rabitEnv.asJava)
-        val featuresIterator = rowIterator.map(row => row.getAs[Vector](
-          $(featuresCol))).toList.iterator
+
         import DataUtils._
+
+        val rowBuffer = mutable.ArrayBuffer[Row]()
+        val featureBuffer = mutable.ArrayBuffer[XGBLabeledPoint]()
+        rowIterator.foreach { row =>
+          rowBuffer += row
+          featureBuffer += row.getAs[Vector]($(featuresCol)).asXGB
+        }
+
         val cacheInfo = {
           if ($(useExternalMemory)) {
             s"$appName-${TaskContext.get().stageId()}-dtest_cache-${TaskContext.getPartitionId()}"
@@ -302,30 +308,22 @@ class XGBoostClassificationModel private[ml](
             null
           }
         }
+
         val dm = new DMatrix(
-          XGBoost.processMissingValues(featuresIterator.map(_.asXGB), $(missing)),
+          XGBoost.processMissingValues(featureBuffer.iterator, $(missing)),
           cacheInfo)
         try {
           val Array(rawPredictionItr, probabilityItr, predLeafItr, predContribItr) =
             producePredictionItrs(bBooster, dm)
           Rabit.shutdown()
-          Iterator(rawPredictionItr, probabilityItr, predLeafItr,
-            predContribItr)
+          produceResultIterator(rowBuffer.iterator,
+            rawPredictionItr, probabilityItr, predLeafItr, predContribItr)
         } finally {
           dm.delete()
         }
       } else {
         Iterator()
       }
-    }
-    val resultRDD = inputRDD.zipPartitions(predictionRDD, preservesPartitioning = true) {
-      case (inputIterator, predictionItr) =>
-        if (inputIterator.hasNext) {
-          produceResultIterator(inputIterator, predictionItr.next(), predictionItr.next(),
-            predictionItr.next(), predictionItr.next())
-        } else {
-          Iterator()
-        }
     }
 
     bBooster.unpersist(blocking = false)
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
@@ -19,6 +19,7 @@ package ml.dmlc.xgboost4j.scala.spark
 import scala.collection.Iterator
 import scala.collection.JavaConverters._
 import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
 
 import ml.dmlc.xgboost4j.java.Rabit
 import ml.dmlc.xgboost4j.scala.{Booster, DMatrix, XGBoost => SXGBoost}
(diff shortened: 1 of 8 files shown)


CodingCat:
thanks, yeah, this is a bug which is not easy to fix

this PR actually falls back to the previous problem about memory footprint, you can check https://github.com/dmlc/xgboost/commit/f368d0de2b6d60cdba67fed8291e7a3ce402850d#diff-a435450e9c28607f848ccf3246944a44 

let me think about what is the right way to fix, if we go to sort ,we need to do significant perf benchmarking before merging

sperlingxx:
@CodingCat 
Thank you for replying me that quickly. 
Actually, this solution is quite like v0.81's implementation.
I have several ideas to improve this(maybe not good ideas = =)
1. do prediction in _MiniBatch_ style
    Maybe we can add a param like batch size?
2. add params _idColumn_ and _appendColumns_
   We filter other columns to reduce memory overhead of caching original data.

CodingCat:
@sperlingxx can you elaborate more on the second approach?

CodingCat:
at the same time I am benchmarking what if we sortWithinPartition beforehand

CodingCat:
I tested with a prototype to sort each partition before hand, 

without sorting, we need 6+mins to finish the prediction over 120G input

with sort, it grows to 12+mins to finish the same task

CodingCat:
I am experimenting with several potential solutions and finding more problems in our implementation, will update soon

sperlingxx:
@CodingCat 
The second approach, we keep as fewer columns as we can, before the batch prediction stage(transformInternal).  So, we can cache fewer data of original DataFrame(inputRDD).

 something like:
>   dataset.toDF().select($(appendColumns) : _*)

CodingCat:
ok, so I essentially tried three approaches to resolve the issue and finding more problems in XGBoost. 

0. approaches

I tried <b>sorting</b> the DataFrame before feeding to transformInternal(), <b>duplicate</b> dataset like the implementation here and <b>miniBatch</b>

1. benchmark of different approaches

I trained a model based on an internal dataset having 1.5~b rows and 20~ features and load the model to predict the training dataset in a separate Spark application.

To scale the test, I manually duplicate the dataset in Spark application and the benchmark results only counts the time spent on prediction stage

used code: https://github.com/CodingCat/xgboost4j-spark-scalability/blob/master/src/main/scala/me/codingcat/xgboost4j/PureXGBoostPredictor.scala

2. benchmark results

image

resources I used `--num-executors 100 --executor-memory 14g --executor-cores 8`

3. problems here, 

booster's prediction method is and has to be an synchronized method

CodingCat:
the actions here

1. I will submit a PR to @sperlingxx 's branch soon 

2. fix the sharing issue in booster after this is merged (not blocking tho)

hcho3:
@CodingCat BTW, you have permission to directly modify all PRs as a maintainer

CodingCat:
@hcho3 how to modify with a bunch of changes?

hcho3:
@CodingCat You can get a local clone of this PR by running
```
git clone --recursive https://github.com/sperlingxx/xgboost -b hot_fix_spark_estimator
```

CodingCat:
I see.......thx

sperlingxx:
@CodingCat 
 I'm not sure whether the implementation here will work in a minibatch way with the whole pipeline. 
So, I rewrote the classification part in a pure lazy(iterator) style. I hope it will be helpful :)

CodingCat:
@sperlingxx can you explain how your implementation is a `pure lazy(iterator) style` and why my suggested implementation is not?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
```diff
@@ -287,46 +284,54 @@ class XGBoostClassificationModel private[ml](
...
-          if ($(useExternalMemory)) {
-            s"$appName-${TaskContext.get().stageId()}-dtest_cache-${TaskContext.getPartitionId()}"
-          } else {
-            null
+    val resultRDD = dataset.asInstanceOf[Dataset[Row]].rdd.mapPartitions { rowIterator =>
+      new AbstractIterator[Row] {
+        private var batchCnt = 0
+
+        private val batchIterImpl = rowIterator.grouped(
+          XGBoostClassificationModel.PREDICTION_BATCH_SIZE).flatMap { batchRow =>
```
CodingCat: batchRow has been a Seq[Row] instead of Iterator[] here, so it's not lazy evaluated and stays in memory until this batch is finished

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
```diff
@@ -287,46 +284,54 @@ class XGBoostClassificationModel private[ml](
...
+          if (batchCnt == 0) {
+            val rabitEnv = Array("DMLC_TASK_ID" -> TaskContext.getPartitionId().toString).toMap
+            Rabit.init(rabitEnv.asJava)
+          }
+
+          val features = batchRow.map(row => row.getAs[Vector]($(featuresCol)))
```
CodingCat: regarding the memory footprint, you have put two Seq, one for Seq[Row], one for Seq[Vector] in memory, 

you can compare with my implementation, it only keeps a Seq[Row] due to iterator.duplicate()

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
```diff
@@ -287,46 +284,54 @@ class XGBoostClassificationModel private[ml](
...
-          Rabit.shutdown()
-          Iterator(rawPredictionItr, probabilityItr, predLeafItr,
-            predContribItr)
-        } finally {
-          dm.delete()
+
+        override def hasNext: Boolean = batchIterImpl.hasNext
+
+        override def next(): Row = {
+          val ret = batchIterImpl.next()
```
CodingCat: ah, I see, the key point is here, IIUC, with this implementation you don't need to locate prediction results in memory for a complete partition,

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
CodingCat: but we should think about the memory footprint in this place as the "grouped iterator" has been put in memory for twice

CodingCat:
this is a significant change regarding performance and we need to be very careful about the correctness as well

can you also use internal dataset for evaluation?

CodingCat:
ok, here is my latest benchmark results, looks like with the current implementation is slower than the diff I pushed yesterday

image


My theory is still on the synchronized method in Booster

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
sperlingxx: Yes. And, I've tried to mock your solution with the below script. 

```scala
object MiniBatchDebug {
  import scala.collection.Iterator
  import scala.collection.mutable.ListBuffer

  private val batchSize = 3
  private val dataSize = 9

  def f(iter: Iterator[Array[Int]]): Iterator[(Array[Int], Int)] = {
    var batchCnt = 0
    val results = new ListBuffer[Iterator[(Array[Int], Int)]]
    while (iter.hasNext) {
      println(s"fetch batch num: $batchCnt")
      val (x, y) = iter.take(batchSize).duplicate
      // mock the reading behavior of batch compute
      val ret = y.toArray.map(_.sum).iterator
      results += x.zip(ret)
      batchCnt += 1
    }
    results.reduce((i1, i2) => i1 ++ i2)
  }

  def main(args: Array[String]): Unit = {
    val dataSource = Seq.tabulate(dataSize)(x => Array(x, x * 2))
    val sourceIter: Iterator[Array[Int]] = dataSource.iterator
    val result = f(sourceIter).map { case (x, y) => x.sum + y }
    (1 to dataSize).foreach { i =>
      println(s"fetch element $i")
      result.next()
    }
  }
}
```

Here is the print info, which demonstrated it will run all batches before real data fetching :
>  fetch batch num: 0
fetch batch num: 1
fetch batch num: 2
fetch element 1
fetch element 2
fetch element 3
fetch element 4
fetch element 5
fetch element 6
fetch element 7
fetch element 8
fetch element 9

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
sperlingxx: Oh, it's an unnecessary footprint. Maybe it can be replaced by 
```scala
val features = batchRow.iterator.map(row => row.getAs[Vector]($(featuresCol)))
```

sperlingxx:
> ok, here is my latest benchmark results, looks like with the current implementation is slower than the diff I pushed yesterday
> 
> image
> 
> My theory is still on the synchronized method in Booster

Thanks for benchmarking!

I'm a little confused about the __context switches__ caused by _synchronized prediction_ . Is it because there are multiple spark tasks running on each executor concurrently, and their booster share the same handle?

CodingCat:
Yes, because we are using a broadcast booster which is singleton per executor, and regarding why we use broadcasted booster, you can check my previous comments

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifierSuite.scala
```diff
@@ -27,11 +27,12 @@ import org.apache.spark.Partitioner
 
 class XGBoostClassifierSuite extends FunSuite with PerTest {
 
-  test("XGBoost-Spark XGBoostClassifier ouput should match XGBoost4j") {
+  test("XGBoost-Spark XGBoostClassifier output should match XGBoost4j") {
     val trainingDM = new DMatrix(Classification.train.iterator)
     val testDM = new DMatrix(Classification.test.iterator)
     val trainingDF = buildDataFrame(Classification.train)
     val testDF = buildDataFrame(Classification.test)
+    val randSortedTestDF = buildDataFrameWithRandSort(Classification.test)
```
CodingCat: let's separate them into two test to highlight the randsorting version and normal version of the test

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressorSuite.scala
```diff
@@ -25,11 +25,12 @@ import org.scalatest.FunSuite
 
 class XGBoostRegressorSuite extends FunSuite with PerTest {
 
-  test("XGBoost-Spark XGBoostRegressor ouput should match XGBoost4j: regression") {
+  test("XGBoost-Spark XGBoostRegressor output should match XGBoost4j: regression") {
```
CodingCat: make the test name consistent with classifier part

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
```diff
@@ -47,7 +47,7 @@
    *                  the prediction of these DMatrices will become faster than not-cached data.
    * @throws XGBoostError native error
    */
-  Booster(Map<String, Object> params, DMatrix[] cacheMats) throws XGBoostError {
+  public Booster(Map<String, Object> params, DMatrix[] cacheMats) throws XGBoostError {
```
CodingCat: this might no be necessary, maybe my bad

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
```diff
@@ -287,46 +284,58 @@ class XGBoostClassificationModel private[ml](
...
+    val resultRDD = dataset.asInstanceOf[Dataset[Row]].rdd.mapPartitions { rowIterator =>
+      new AbstractIterator[Row] {
+        private var batchCnt = 0
+
+        private val batchIterImpl = rowIterator.grouped(
+          XGBoostClassificationModel.PREDICTION_BATCH_SIZE).flatMap { batchRow =>
+<<<<<<< HEAD
+=======
+
+>>>>>>> regressor impl
```
CodingCat: my bad

CodingCat:
I left more comments there, @sperlingxx would you please move forward with the PR

my suggestion is, use your way for the next version, and look at how to resolve shared properties among boosters after that

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifierSuite.scala
sperlingxx: okay!

CodingCat:
LGTM, thanks, will merge after CI is happy

Issue #4406, opened 2019-04-26
[jvm-packages] SparkMLlibPipeline example in xgboost4j-example causes error "rabit::Init is already called in this thread"


hcho3:
I ran spark-submit on two Spark examples in xgboost4j-example as follows:
```
spark-submit --class ml.dmlc.xgboost4j.scala.example.spark.SparkTraining  \
  --master 'local[8]' \
  ./target/xgboost4j-tester-1.0-SNAPSHOT-jar-with-dependencies.jar \
  /workspace/jvm-packages/xgboost4j-tester/iris.csv
spark-submit --class ml.dmlc.xgboost4j.scala.example.spark.SparkMLlibPipeline  \
  --master 'local[8]' \
  ./target/xgboost4j-tester-1.0-SNAPSHOT-jar-with-dependencies.jar  \
  /workspace/jvm-packages/xgboost4j-tester/iris.csv  \
  /workspace/jvm-packages/xgboost4j-tester/native_model  \
  /workspace/jvm-packages/xgboost4j-tester/pipeline_model
```

The first spark-submit ran successfully, but the second spark-submit crashes with error
```
rabit::Init is already called in this thread
```

See the full log at https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-4401/15/pipeline/204#step-249-log-1296.

CodingCat:
we have a bug in calling rabit.init() in spark, 

I believe even you only run the second test, SparkMLlibPipeline, it also failed

hcho3:
@CodingCat Was this bug around for a while?

CodingCat:
@hcho3 I think so, I am verifying that, because we usually will not evaluate the dataframe containing predictions for twice so it is not triggered......

CodingCat:
this bug was introduced in https://github.com/dmlc/xgboost/commit/2d875ec0197d5a83e7d585daf472b8201aa97c51

CodingCat:
working on a fix

hcho3:
@CodingCat Thanks!

CodingCat:
Rabit part fix: https://github.com/dmlc/rabit/pull/92

PR #4407, opened 2019-04-27
[jvm-packages] allow partial evaluation of dataframe before prediction

CodingCat:
closes https://github.com/dmlc/xgboost/issues/4406

@sperlingxx FYI, because we call rabit.shutdown only when the partition is exhausted https://github.com/dmlc/xgboost/pull/4388/files#diff-a435450e9c28607f848ccf3246944a44R330, if we partially evaluate a DataFrame with APIs like show(), it will leave some rabit in initialized but not shutdown state

as a result, when we run a new task in the same thread, rabit will crash since we cannot call init() for twice, 

I have fixed rabit in https://github.com/dmlc/rabit/pull/92 and added a test in xgboost guarding it

@hcho3, I am resuming the disabled spark test in this PR, hopefully everything fine

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,7 +78,7 @@ else ()
 endif (MINGW OR R_LIB)
 add_library(rabit STATIC ${RABIT_SOURCES})
 target_include_directories(rabit PRIVATE
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/rabit/include>
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/dmlc-core/include>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/rabit/include/rabit>)
 set_target_properties(rabit
   PROPERTIES
--- a/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/RabitRobustnessSuite.scala
+++ b/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/RabitRobustnessSuite.scala
@@ -219,7 +219,11 @@ class RabitSuite extends FunSuite with PerTest {
     val model = new XGBoostClassifier(paramMap ++ Array("num_round" -> 10,
       "num_workers" -> numWorkers)).fit(trainingDF)
     val prediction = model.transform(trainingDF)
+    // a partial evaluation of dataframe will cause rabit initialized but not shutdown in some
+    // threads
     prediction.show()
-    prediction.show()
+    // a full evaluation here will re-run init and shutdown all rabit proxy
+    // expecting no error
+    prediction.collect()
   }
 }
--- a/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/RabitRobustnessSuite.scala
+++ b/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/RabitRobustnessSuite.scala
@@ -16,14 +16,73 @@
 
 package ml.dmlc.xgboost4j.scala.spark
 
+import java.util.concurrent.LinkedBlockingDeque
+
+import scala.util.Random
+
 import ml.dmlc.xgboost4j.java.{IRabitTracker, Rabit, RabitTracker => PyRabitTracker}
 import ml.dmlc.xgboost4j.scala.rabit.{RabitTracker => ScalaRabitTracker}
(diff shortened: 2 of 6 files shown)


CodingCat:
@chenqin I am upgrading rabit in this PR as I need to bring my fix on rabit here

CodingCat:
@hcho3 I am trying to get familiar with the new CI system, I think the unit test of jvm packages is not really run?

the only mvn package command I found is 

image

which is under xgboost4j-tester directory...did I miss anything?

hcho3:
@CodingCat My bad. In this line, https://github.com/dmlc/xgboost/blob/ea850ecd209f647f737f8268d2ba527281feb85d/tests/ci_build/build_jvm_packages.sh#L12
`-DskipTests` should not be there. I forgot to remove it after testing.

hcho3:
Note that there are two places where unit tests are run:
1. `build-jvm-packages` (all unit tests)
2. `test-jvm-jdk*` (unit tests for ml.dmlc.xgboost4j)

CodingCat:
> Note that there are two places where unit tests are run:
> 
> 1. `build-jvm-packages` (all unit tests)
> 2. `test-jvm-jdk*` (unit tests for ml.dmlc.xgboost4j)

got it, thanks!

hcho3:
Looks like all unit tests are passing now. Thanks for catching my mistake :)

chenqin:
I think we just need to make sure passing same id to rabit as one of
- rabit_task_id
- DMLC_TASK_ID
- mapred_tip_id
In spark use
TaskContext.partitionId() 

also 
- mapred_task_id

In spark use TaskContext.attemptNumber()

The reason is 
AllreduceBase::ConnectTracker(void) will report task_id to tracker and it's better make sure tracker aware this is reconnect instead of bootstrap.

CodingCat:
> I think we just need to make sure passing same id to rabit as one of
> 
> * rabit_task_id
> * DMLC_TASK_ID
> * mapred_tip_id
>   In spark use
>   TaskContext.partitionId()
> 
> also
> 
> * mapred_task_id
> 
> In spark use TaskContext.attemptNumber()
> 
> The reason is
> AllreduceBase::ConnectTracker(void) will report task_id to tracker and it's better make sure tracker aware this is reconnect instead of bootstrap.

NO, it doesn't touch anything about what is passed to rabit, check https://github.com/dmlc/rabit/pull/92/files#diff-e9982813ddad38a491ec44b1331647c1L48

chenqin:
> > I think we just need to make sure passing same id to rabit as one of
> > 
> > * rabit_task_id
> > * DMLC_TASK_ID
> > * mapred_tip_id
> >   In spark use
> >   TaskContext.partitionId()
> > 
> > also
> > 
> > * mapred_task_id
> > 
> > In spark use TaskContext.attemptNumber()
> > The reason is
> > AllreduceBase::ConnectTracker(void) will report task_id to tracker and it's better make sure tracker aware this is reconnect instead of bootstrap.
> 
> NO, it doesn't touch anything about what is passed to rabit, check https://github.com/dmlc/rabit/pull/92/files#diff-e9982813ddad38a491ec44b1331647c1L48
> 
> and there is no tracker in this case
this rabit change https://github.com/dmlc/rabit/pull/92 will cause tracker assign random rank if the envs were not passed in.

CodingCat:
> > > I think we just need to make sure passing same id to rabit as one of
> > > 
> > > * rabit_task_id
> > > * DMLC_TASK_ID
> > > * mapred_tip_id
> > >   In spark use
> > >   TaskContext.partitionId()
> > > 
> > > also
> > > 
> > > * mapred_task_id
> > > 
> > > In spark use TaskContext.attemptNumber()
> > > The reason is
> > > AllreduceBase::ConnectTracker(void) will report task_id to tracker and it's better make sure tracker aware this is reconnect instead of bootstrap.
> > 
> > 
> > NO, it doesn't touch anything about what is passed to rabit, check https://github.com/dmlc/rabit/pull/92/files#diff-e9982813ddad38a491ec44b1331647c1L48
> > and there is no tracker in this case
> 
> this rabit change [dmlc/rabit#92](https://github.com/dmlc/rabit/pull/92) will cause tracker assign random rank if the envs were not passed in.

how to contact a tracker without passing envs?????

chenqin:
humm, nvm, I see what you mean.
`
 String nodeHost = container.getNodeId().getHost();
        env.put("DMLC_NODE_HOST", nodeHost);
        env.put("DMLC_TASK_ID", String.valueOf(task.taskId));
        env.put("DMLC_ROLE", task.taskRole);
        env.put("DMLC_NUM_ATTEMPT", String.valueOf(task.attemptCounter));
        // ctx.setUser(userName);
        ctx.setEnvironment(env);
`

PR #4411, opened 2019-04-27
[jvm-packages] fix compatibility problem of spark version

sperlingxx:
I found that `VectorAssembler. setHandleInvalid`,  which used in `MissingValueHandlingSuite.scala`, 
is newly added in spark_2.4.
So, I tried to replace`VectorAssembler` by creating `DenseVector` directly.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/MissingValueHandlingSuite.scala
+++ b/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/MissingValueHandlingSuite.scala
@@ -16,12 +16,13 @@
 
 package ml.dmlc.xgboost4j.scala.spark
 
-import scala.util.Random
 import ml.dmlc.xgboost4j.java.XGBoostError
-import org.scalatest.FunSuite
 import org.apache.spark.ml.feature.VectorAssembler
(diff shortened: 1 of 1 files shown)


CodingCat:
I don't think I would agree with the changes here

1. first of all, we officially only support Spark 2.4+ now, 

2. second, when we write tests, we should simulate user's behavior with the best effort. With this principle, we should not skip vector assembler

3. skip vector assembler is risky to bring hard-to-notice inconsistency, e.g.https://github.com/dmlc/xgboost/pull/4411/files#diff-76870d19095de7c9de6d6c703efe2346R56, if you produce this vector with vector assembler, it cannot be a dense one

sperlingxx:
> I don't think I would agree with the changes here
> 
> 1. first of all, we officially only support Spark 2.4+ now,
> 2. second, when we write tests, we should simulate user's behavior with the best effort. With this principle, we should not skip vector assembler
> 3. skip vector assembler is risky to bring hard-to-notice inconsistency, e.g.https://github.com/dmlc/xgboost/pull/4411/files#diff-76870d19095de7c9de6d6c703efe2346R56, if you produce this vector with vector assembler, it cannot be a dense one

So, I reverted the ` vectorAssembler`. And call `setHandleInvalid` by runtime reflection only when spark >= 2.4, so as to keep this Suite compatible with earlier version.

CodingCat:
merged, thanks

Issue #4398, opened 2019-04-24
Implement bart.


trivialfis:
Hi all,

## Introduction
Recently I have been reading **[Bayesian Additive Regression Tree](https://arxiv.org/pdf/0806.3286.pdf)** and started to implement it on top of XGBoost.  It's a tree ensemble model with MCMC(modified gibbs) sampling for structure learning.  Recent years some variances have been proposed, mostly tuning the sampling method to get higher performance or lower computation cost.  Currently I'm still trying to get a grasp for the original version.

Although XGBoost itself is a gradient boosting framework, but exception like random forest is also implemented in the library.  I would like to know the interest in this subject.  If the algorithm is accepted as part of XGBoost, I will try to implement it as a plugin for initial support.

## Existing implementation
The algorithm has been implemented many times as R package:
- [bart machine](https://github.com/cran/bartMachine)
- [bart](https://github.com/cran/BART.git)
- [bart bma](https://github.com/BelindaHernandez/BART-BMA.git)
- and the original implementation.

Also one WIP Python project:
- [bartpy](https://github.com/JakeColtman/bartpy)

I don't quite understand the R packages' source code so am only familiar with bartpy.

## My plan
Even though currently I'm still at "trying to do it", but already have over 1000 lines of code added so it's not trivial.  If the model is accepted I will first submit a naive implementation that can be reviewed.

@tqchen @hcho3 @RAMitchell @CodingCat

hcho3:
@trivialfis This is interesting. I will need to give myself some time to review the paper.

I think you should do either 1) write a RFC to explain your (software) design, in the form similar to dmlc/tvm#2889 OR 2) write a draft pull request that can be easily reviewed.

PR #4412, opened 2019-04-27
Initial commit for bart.

trivialfis:
Introduced in #4398, this is a draft implementation of  **[Bayesian Additive Regression Tree](https://arxiv.org/pdf/0806.3286.pdf)**.  It's a tree ensemble model with MCMC(CGM98 gibbs) as structure learning algorithm.  Although  XGBoost is a gradient boosting library, which is quite different than sampling methods.  There are Bayesian methods that use gradient, but to my knowledge non has apply it on tree model yet.  I'm interested in this subject and want to work on it in the future if possible.  I hope that implementing bart can open up the possibility of Bayesian view in XGBoost.

Currently it's far from usable, notably missing a predictor, classification support, burn in support and optimization is non-exist.  Also, please ignore the styling.

Other than these, there are a few design issues to be addressed:
- gibbs sampling doesn't use gradient at all so there will be conflict with current internal structure.
- sampling based method is kind of hard to test.
- missing data handling is difficult since we can't just accumulate some statistic.

In this PR a `GibbsUpdater` is created to coordinate the sampling iteration, while `TreeMutation` class is an enhanced version of `RegTree` with specific sampling features.  New model lives in plugin for now.  I would like to know the opinion from community about this added feature. The development will continue if it's accepted as an useful model and suitable to live in XGBoost.

@tqchen @hcho3 @RAMitchell @CodingCat and anyone who's interested. ;-)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,7 @@ address, leak and thread.")
 ## Plugins
 option(PLUGIN_LZ4 "Build lz4 plugin" OFF)
 option(PLUGIN_DENSE_PARSER "Build dense parser plugin" OFF)
+option(PLUGIN_BART "Build bart plugin" ON)
 
 ## Deprecation warning
 if (USE_AVX)
@@ -54,6 +55,10 @@ if (USE_CUDA)
   format_gencode_flags("${GPU_COMPUTE_VER}" GEN_CODE)
   message(STATUS "CUDA GEN_CODE: ${GEN_CODE}")
 endif (USE_CUDA)
+
+if (PLUGIN_BART)
+  add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/plugin/bart)
+endif (PLUGIN_BART)
 
 # dmlc-core
 add_subdirectory(${PROJECT_SOURCE_DIR}/dmlc-core)
--- a/include/xgboost/tree_model.h
+++ b/include/xgboost/tree_model.h
@@ -495,7 +495,7 @@ class RegTree {
    */
   void FillNodeMeanValues();
 
- private:
+ protected:
   // vector of nodes
   std::vector<Node> nodes_;
   // free node space, used during training process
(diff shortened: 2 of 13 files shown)


CodingCat:
It may take a while for me to read the paper first...

CodingCat:
@yinlou looks like relevant to your research

yinlou:
Overall structure makes sense to me. It would be interesting to include some benchmark results on open datasets to show the benefits of BART.

trivialfis:
I have a local implementation that is more complete and testable.  By now I have some experience with various Bayesian methods.  I think one of the problems of these models that rarely mentioned in literatures, is their heavy usage of `log` and `exp`.  It's very difficult to avoid numeric overflow/underflow.

trivialfis:
Nah, let's stay with GB.  For anyone interested in where did this go, see https://github.com/trivialfis/kapok .

trivialfis:
Here's a simple explanation why I dropped the idea of merging it into XGBoost, for anyone who's interested.  And sorry for the noise.

BART, like many other ensemble model, one first train a weak model, ask it for prediction, then subtract its prediction from true labels (y) and get the error.  After that we train another weak model, use previous error as the new target, add the second model's prediction to the first one to reduce total error, and so on.  Instead of using greedy exploration, BART uses sampling method, with each stage's error combined with other factors as the sampling mean, specifically mean of Gaussian distribution for leaves' value.

During each stage, it draws leaves' value non-deterministically, which means it can draw a good tree that reduces the error.  But it's also possible that it draw a bad tree that somehow enlarge the error.  Mathematically the error should still converge to zero.  But in my experiments, due to its heavy use of 'log', 'exp' and division, numeric overflows happens easily.  If in one of the training iterations we get a numeric overflow for one leaf (we prefer small trees so every leaf matters), all the previous training is basically ruined.  And the next iteration will use the mean containing this ridiculous error to sample new values, therefore it won't be corrected quickly.  In general it's not a very robust way to train a model.

A possible fix at the algorithm level is, we somehow add rejection sampling for leaves' value and related factors, that accepts drawn values probabilistically conditioned on new errors.  But that would be a research project for the future. ;-)  Thanks for the attention on this.

@yinlou @RAMitchell @CodingCat @hcho3

PR #4419, opened 2019-04-30
[BLOCKING] Make CMakeLists.txt compatible with CMake 3.3

### LINKED ISSUE #4415: XGBoost fails to build because of missing -pthread and -fopenmp cmake directives
1. -pthread is not included into CMAKE_CXX_FLAGS
2. -fopenmp is not included into CMAKE_CXX_FLAGS

To fix the error; had to modify CMakeLists.txt to include the following

set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")

if (USE_OPENMP)
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp")
endif (USE_OPENMP)

hcho3:
* Export OpenMP as a PUBLIC target
* Upgrade CMake to 3.3, to use the generator expression `COMPILE_LANGUAGE:CXX`. For the same reason, require CMake 3.11 for Visual Studio
* Use CMake 3.3 in the `build-cpu` task of Jenkins
* Remove `-funroll-loops` flag when compiled with MSVC

Closes #4415

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,14 @@
-cmake_minimum_required(VERSION 3.2)
+cmake_minimum_required(VERSION 3.3)
 project(xgboost LANGUAGES CXX C VERSION 0.82)
 include(cmake/Utils.cmake)
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/modules")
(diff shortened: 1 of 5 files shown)


hcho3:
Closing for now. I've made things quite too complicated. OpenMP doesn't really need to be a target

PR #4425, opened 2019-04-30
[jvm-packages] Tutorial on handling missing values

Daniel8hen:
add tutorial on missing values and how to handle those within XGBoost.

### DIFF
--- a/doc/jvm/xgboost4j_spark_tutorial.rst
+++ b/doc/jvm/xgboost4j_spark_tutorial.rst
@@ -153,16 +153,22 @@ Now, we have a DataFrame containing only two columns, "features" which contains 
 "sepal length", "sepal width", "petal length" and "petal width" and "classIndex" which has Double-typed
 labels. A DataFrame like this (containing vector-represented features and numeric labels) can be fed to XGBoost4J-Spark's training engine directly.
 
-Dealing with missing values (NaN / Null):
+Dealing with missing values (NaN / Null)
 --------------------------
 
-In case a feature (column) contains missing values (NaN / NULLs) for any reason (could be related to business logic / wrong data ingestion process / any other reason), one should handle it in the below manner, following the below steps:
+In case a feature (column) contains missing values (NaN / NULLs) for any reason (could be related to business logic / wrong data ingestion process / etc.), one should decide on a strategy of how to handle it.
 
-1. Set some irregular value to that feature.
+Possible ways are - 
+1. One can ignore it in the VectorAssembler (using setHandleInvalid = "skip"); 
+2. keep it (using setHandleInvalid = "keep"), and then XGBoost will handle it.
+
+In case a user chose strategy #2, it should handle it in the below manner:
+
+2.1. Set some irregular value to that feature.
 For example: A feature that contains values in the range of 0 to 1, we can set the missing values as -999.
 (Note: The missing value should be replaced with the same value on whole of the dataset, and not only on a specific feature)
 
-2. Once we have "padded" our dataframe with irregular values for the missing values, we should tell our XGBoostClassifier / XGBoostRegressor about it, via a parameter called "missing", which represents the missing value.
+2.2. Once we have "padded" our dataframe with irregular values for the missing values, we should tell our XGBoostClassifier / XGBoostRegressor about it, via a parameter called "missing", which represents the missing value.
 The algorithm will automatically learn what's the ideal direction to go when a value is missing, based on that value. 
 
 Example:
@@ -182,6 +188,18 @@ Example:
 
 3. Then, once we transform (test) our XGBoost model, we should see that indeed the number of records match the number we "fed" our model with.
 
+
+Dealing with 0's as missing values
+--------------------------
+In case a dataset contains 0 as a missing value, and those are meaning values (i.e. represent some kind of a behavior in the data), one should handle it, due to the fact that Spark's VectorAssembler transformer only accepts 0 as the missing values. This issue creates a problem when the user has 0 as meaningful value, plus there are enough 0's to use SparseVector.
+
+In case the dataset is represented by a DenseVector, the 0 is kept.
+
+In order to handle it, one should follow the below manner:
+
+1. Transform 0 to other value, for example: A feature that contains values in the range of 0 to 1, we can set the missing values as -999.
+
+2. Note that 0 values will be filtered in case Spark's VectorAssembler uses SparseVector (and not DenseVector).
 
 Training
 ========
--- a/doc/jvm/xgboost4j_spark_tutorial.rst
+++ b/doc/jvm/xgboost4j_spark_tutorial.rst
@@ -153,7 +153,7 @@ Now, we have a DataFrame containing only two columns, "features" which contains 
 "sepal length", "sepal width", "petal length" and "petal width" and "classIndex" which has Double-typed
 labels. A DataFrame like this (containing vector-represented features and numeric labels) can be fed to XGBoost4J-Spark's training engine directly.
 
-Dealing with missing values:
+Dealing with missing values
 ~~~~~~~~~~~~~~~~~~~~~~
 
 In the case that a feature column contains missing values for any reason (could be related to business logic / wrong data ingestion process / etc.), the user should decide on a strategy of how to handle it. 
--- a/doc/jvm/xgboost4j_spark_tutorial.rst
+++ b/doc/jvm/xgboost4j_spark_tutorial.rst
@@ -155,10 +155,6 @@ labels. A DataFrame like this (containing vector-represented features and numeri
 
 Dealing with missing values
 ~~~~~~~~~~~~~~~~~~~~~~
-SparseVector VS. DenseVector in Spark and XGBoost:
-due to the fact that Spark's VectorAssembler transformer only accepts 0 as a missing values, this one creates a problem when the user has 0 as meaningful value, plus there are enough 0's to use SparseVector.
-
-However, In case the dataset is represented by a DenseVector, the 0 is kept.
 
 Strategies to handle missing values (and therefore overcome issues as above):
 
@@ -193,6 +189,11 @@ Example of setting a missing value (e.g. -999) to the "missing" parameter in XGB
         setFeaturesCol("features").
         setLabelCol("classIndex")
 
+.. note:: Using 0 to represent meaningful value
+
+Due to the fact that Spark's VectorAssembler transformer only accepts 0 as a missing values, this one creates a problem when the user has 0 as meaningful value plus there are enough 0's to use SparseVector (However, In case the dataset is represented by a DenseVector, the 0 is kept)
+
+In this case, users are also supposed to transform 0 to some other values to avoid the issue.
 
 Training
 ========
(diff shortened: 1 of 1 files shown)


doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -153,6 +153,36 @@ Now, we have a DataFrame containing only two columns, "features" which contains
 "sepal length", "sepal width", "petal length" and "petal width" and "classIndex" which has Double-typed
 labels. A DataFrame like this (containing vector-represented features and numeric labels) can be fed to XGBoost4J-Spark's training engine directly.
 
+Dealing with missing values (NaN / Null):
+--------------------------
+
+In case a feature (column) contains missing values (NaN / NULLs) for any reason (could be related to business logic / wrong data ingestion process / any other reason), one should handle it in the below manner, following the below steps:
```
CodingCat: it's not true actually, if the feature is `NaN` instead of non-exist or as NULL, 

users should choose to ignore it in vector assembler or keep it to let xgboost handle

Daniel8hen:
Hey guys, I'm updating per your chnages and will ping once all is done for the 2nd review.
Thanks,
Daniel

Daniel8hen:
@hcho3 , @CodingCat I've update the tutorial accordingly. Waiting for your input once again.

doc/jvm/xgboost4j_spark_tutorial.rst
Daniel8hen: @CodingCat note I've fixed it and added it to the document.

Daniel8hen:
> We also need to mention the case where 0's are meaningful values, i.e. 0 != missing

@hcho3 note it was fixed and added.

Daniel8hen:
@hcho3 @CodingCat guys any update? What do you think about the last version? Thank you!

doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -153,6 +153,54 @@ Now, we have a DataFrame containing only two columns, "features" which contains
...
+
+In case a feature (column) contains missing values (NaN / NULLs) for any reason (could be related to business logic / wrong data ingestion process / etc.), one should decide on a strategy of how to handle it.
+
+Possible ways are - 
+1. One can ignore it in the VectorAssembler (using setHandleInvalid = "skip"); 
+2. keep it (using setHandleInvalid = "keep"), and then XGBoost will handle it.
+
+In case a user chose strategy #2, it should handle it in the below manner:
+
+2.1. Set some irregular value to that feature.
```
CodingCat: I think it's still some kind of inaccurate, 

for NaN, even users decide to keep it in VectorAssembler stage, they can set missing parameter as NaN (which is the default) to ignore it instead of transforming

only for NULL, users have to transform the dataset

doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -153,6 +153,54 @@ Now, we have a DataFrame containing only two columns, "features" which contains
...
+  val xgbClassifier = new XGBoostClassifier(xgbParam).
+        setFeaturesCol("features").
+        setLabelCol("classIndex")
+
+3. Then, once we transform (test) our XGBoost model, we should see that indeed the number of records match the number we "fed" our model with.
+
+
+Dealing with 0's as missing values
+--------------------------
+In case a dataset contains 0 as a missing value, and those are meaning values (i.e. represent some kind of a behavior in the data), one should handle it, due to the fact that Spark's VectorAssembler transformer only accepts 0 as the missing values. This issue creates a problem when the user has 0 as meaningful value, plus there are enough 0's to use SparseVector.
```
CodingCat: 0 as `feature` value???

Daniel8hen:
Thank you @hcho3 .

doc/jvm/xgboost4j_spark_tutorial.rst
Daniel8hen: yes, i.e. a value inside a feature within the dataset. A meaningful value.
Would you like me to re-phrase it better?

doc/jvm/xgboost4j_spark_tutorial.rst
Daniel8hen: @CodingCat , we didn't talk about NaN during our discussion so I would appreciate you shed more light and then I will update it accordingly. Definitely willing to do any change so everything will be accurate here.

If you'd like, I can add the below for the "dealing with Null" section ->
****
"Another case is NaN (Not-a-Number). In case users decide to keep those in the VectorAssembler stage (using "setHandleInvalid(keep)), they will need to set the "missing" parameter as NaN (and, it is the default), so XGBoost will handle it as the missing logic, and won't train nor transform it.

The case where users need to deal with changing the values to something else is relevant mostly for Null."
****
What do you think @CodingCat ?

Just let me know and I'll add this as well, and do let me know if you think we need more cycles here.

Thanks,
Daniel

doc/jvm/xgboost4j_spark_tutorial.rst
Daniel8hen: @CodingCat per below note, in case it's NaN - I will refer the user to the below (new text, still not committed) section. In case it's Null - I will keep the above text.

WDYT?

doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -153,6 +153,54 @@ Now, we have a DataFrame containing only two columns, "features" which contains
 "sepal length", "sepal width", "petal length" and "petal width" and "classIndex" which has Double-typed
 labels. A DataFrame like this (containing vector-represented features and numeric labels) can be fed to XGBoost4J-Spark's training engine directly.
 
+Dealing with missing values (NaN / Null)
```
CodingCat: Let's make `dealing with missing values` as a subsection of `Transform Raw Iris Dataset` instead a peer section

doc/jvm/xgboost4j_spark_tutorial.rst
CodingCat: In the case that a feature column contains missing values for any reason (could be related to business logic / wrong data ingestion process / etc.), the user should decide on a strategy of how to handle it. The choice of approach depends on the value representing 'missing' which fall into four categories: (1) 0; (2) NaN; (3) NULL; (4) any other value besides the earlier three cases.

We introduce the following approaches dealing with missing value and their fitting scenarios:

1. skip VectorAssembler (using setHandleInvalid = "skip") directly: used in (2), (3);
2. keep it (using setHandleInvalid = "keep"), and set `missing` parameter in XGBClassifier/XGBRegressor as the value representing `missing`: used in (2) and (4);
3. keep it (using setHandleInvalid = "keep") and transform to other irregular values: used in (3)
4. transform to other values: for case (1), blah blah, blah

doc/jvm/xgboost4j_spark_tutorial.rst
Daniel8hen: @CodingCat done, is part of updated pull request (hopefully final pull req.).

doc/jvm/xgboost4j_spark_tutorial.rst
Daniel8hen: @CodingCat done, is part of updated pull request (hopefully final pull req.).

doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -153,6 +153,39 @@ Now, we have a DataFrame containing only two columns, "features" which contains
...
+(1) 0
+(2) NaN 
+(3) NULL
+(4) any other value which is not mentioned in (1) / (2) / (3)
+
+We introduce the following approaches dealing with missing value and their fitting scenarios:
+(*) Skip VectorAssembler (using setHandleInvalid = "skip") directly. Used in (2), (3).
+(*) Keep it (using setHandleInvalid = "keep"), and set the "missing" parameter in XGBClassifier/XGBRegressor as the value representing missing. Used in (2) and (4).
+(*) Keep it (using setHandleInvalid = "keep") and transform to other irregular values. Used in (3).
+(*) Transform it to other values. Used in (1).
```
CodingCat: you can keep your original description about why it is tricky with (1)

doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -153,6 +153,39 @@ Now, we have a DataFrame containing only two columns, "features" which contains
 "sepal length", "sepal width", "petal length" and "petal width" and "classIndex" which has Double-typed
 labels. A DataFrame like this (containing vector-represented features and numeric labels) can be fed to XGBoost4J-Spark's training engine directly.
 
+Dealing with missing values:
```
CodingCat: @hcho3 is it the right way to make it as a subsection?

Daniel8hen: You are right, fixed again before pull req. @CodingCat

Daniel8hen:
@CodingCat I'm done. Let me know if you see anything else, will be happy to fix that as well. Reviewd both header and content few times and saw it fits plus users understand. Thank you.

doc/jvm/xgboost4j_spark_tutorial.rst
```diff
@@ -153,6 +153,41 @@ Now, we have a DataFrame containing only two columns, "features" which contains
...
+2. NaN.
+3. Null.
+4. any other value which is not mentioned in (1) / (2) / (3).
+
+We introduce the following approaches dealing with missing value and their fitting scenarios:
+
+1. Skip VectorAssembler (using setHandleInvalid = "skip") directly. Used in (2), (3).
+2. Keep it (using setHandleInvalid = "keep"), and set the "missing" parameter in XGBClassifier/XGBRegressor as the value representing missing. Used in (2) and (4).
+3. Keep it (using setHandleInvalid = "keep") and transform to other irregular values. Used in (3).
+4. Transform it to other values. Used in (1).
```
CodingCat: "due to the fact that Spark's VectorAssembler transformer only accepts 0 as the missing values. This issue creates a problem when the user has 0 as meaningful value, plus there are enough 0's to use SparseVector.	

 In case the dataset is represented by a DenseVector, the 0 is kept."

I mean this paragraph from your original commits is informative, you can keep it

doc/jvm/xgboost4j_spark_tutorial.rst
Daniel8hen: Never mind, already updated it and the message to the user is basically the same.

doc/jvm/xgboost4j_spark_tutorial.rst
Daniel8hen: ah I see. I will add it as well then.

CodingCat: I changed accordingly based on your current structure

Daniel8hen: OK, thanks for the changes from your side as well. When expected to see it in tutorial, @CodingCat ?
Thank you

hcho3: The tutorial will be available right after this PR is merged.

Daniel8hen: Thank u @hcho3 . I saw failures on Travis CI - should I ignore it for now @CodingCat ?

hcho3: @Daniel8hen The failed test is most likely due to flaky internet connection. I re-started the test.

Daniel8hen: Thx @hcho3 . So merge will be automatically once all tests and builds are done?

hcho3: Yes, I'll merge it. @CodingCat Are we good with what we have here now?

CodingCat: yes, thanks

hcho3:
@Daniel8hen Here is the rendered tutorial: https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html#dealing-with-missing-values

Thanks everyone!

PR #4436, opened 2019-05-03
Make AUCPR work with multiple query groups

### LINKED ISSUE #4431: AUCPR metric fails for learning-to-rank
Consider the following toy data for learning-to-rank:
```python
import numpy as np
import xgboost

Xrow = np.array([1, 2, 6, 8, 11, 14, 16, 17])
Xcol = np.array([0, 0, 1, 1,  2,  2,  3,  3])
X = csr_matrix((np.ones(shape=8), (Xrow, Xcol)), shape=(20, 4))
y = np.array([0.0, 1.0, 1.0, 0.0, 0.0,
              0.0, 1.0, 0.0, 1.0, 0.0,
              0.0, 1.0, 0.0, 0.0, 1.0,
              0.0, 1.0, 1.0, 0.0, 0.0])    # 1: relevant, 0: not relevant
group = np.array([5, 5, 5, 5], dtype=np.uint)
dtrain = xgboost.DMatrix(X, label=y)
dtrain.set_group(group)

params = {'eta': 1, 'tree_method': 'exact',
          'objective': 'rank:pairwise', 'eval_metric': 'aucpr',
          'max_depth': 1}
bst = xgboost.train(params, dtrain, 10, evals=[(dtrain, 'train')])
```

This example fails because of the following error in the AUCPR metric function:
```
  File "test.py", line 19, in <module>
    bst = xgboost.train(params, dtrain, 10, evals=[(dtrain, 'train')])
  File "/Users/chyunsu/Desktop/xgboost/python-package/xgboost/training.py", line 216, in train
    xgb_model=xgb_model, callbacks=callbacks)
  File "/Users/chyunsu/Desktop/xgboost/python-package/xgboost/training.py", line 84, in _train_internal
    bst_eval_set = bst.eval_set(evals, i, feval)
  File "/Users/chyunsu/Desktop/xgboost/python-package/xgboost/core.py", line 1172, in eval_set
    ctypes.byref(msg)))
  File "/Users/chyunsu/Desktop/xgboost/python-package/xgboost/core.py", line 176, in _check_call
xgboost.core.XGBoostError: [01:55:51] /Users/chyunsu/Desktop/xgboost/src/metric/rank_metric.cc:442: 
Check failed: !auc_gt_one: AUC-PR: AUC > 1.0
```

hcho3:
Fixes #4431.

I read [the paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3961324/) referenced by the AUCPR implementation, and as far as I can tell, the `EvalAucPR` seems to be consistent with the paper.

The issue was that `EvalAucPR` was assuming the existence of only a single query group. More precisely, it throws an exception when `sum_auc >= 1.0`, where it should have checked `sum_auc >= ngroup` instead. Since all tests for AUCPR was performing binary classification (where there is only one group), the bug went undetected.

Fix. Have a proper check on the AUC value, in presence of multiple query groups. Also add a test with learning-to-rank example.

### DIFF
--- a/src/metric/rank_metric.cc
+++ b/src/metric/rank_metric.cc
@@ -436,15 +436,16 @@ struct EvalAucPR : public Metric {
       }
     }
     CHECK(!auc_error) << "AUC-PR: the dataset only contains pos or neg samples";
-    CHECK_LE(sum_auc, static_cast<double>(ngroup)) << "AUC-PR: AUC > 1.0";
     /* Report average AUC across all groups */
     if (distributed) {
       bst_float dat[2];
       dat[0] = static_cast<bst_float>(sum_auc);
       dat[1] = static_cast<bst_float>(ngroup);
       rabit::Allreduce<rabit::op::Sum>(dat, 2);
+      CHECK_LE(dat[0], dat[1]) << "AUC-PR: AUC > 1.0";
       return dat[0] / dat[1];
     } else {
+      CHECK_LE(sum_auc, static_cast<double>(ngroup)) << "AUC-PR: AUC > 1.0";
       return static_cast<bst_float>(sum_auc) / ngroup;
     }
   }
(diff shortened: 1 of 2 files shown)


hcho3:
I'm merging this, since it involves the change of a single assertion and doesn't change the actual implementation of AUCPR.

PR #4437, opened 2019-05-03
prediction with external memory on multiple gpus...

sriramch:
this is a follow up to #4284 which now supports prediction on multiple gpu's using external memory.

@RAMitchell @canonizer @rongou please review.


rongou:
I have a version that uses explicit offsets to slide over the prediction vector. I'll send a separate PR and we can discuss which approach is better.

sriramch:
superseded by #4438

PR #4438, opened 2019-05-03
mgpu predictor using explicit offsets

rongou:
Alternative approach to #4437. Uses explicit offsets of slide over the out prediction vector. I feel this is slightly cleaner, but I'm ok with either approach.

@canonizer @RAMitchell @sriramch

### DIFF
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -291,7 +291,6 @@ class GPUPredictor : public xgboost::Predictor {
 
       const int BLOCK_THREADS = 128;
       size_t num_rows = batch.offset.DeviceSize(device_) - 1;
-      if (num_rows < 1) { return; }
       const int GRID_SIZE = static_cast<int>(dh::DivRoundUp(num_rows, BLOCK_THREADS));
 
       int shared_memory_bytes = static_cast<int>
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -361,6 +361,7 @@ class GPUPredictor : public xgboost::Predictor {
       DeviceOffsets(batch.offset, &device_offsets);
       batch.data.Reshard(GPUDistribution::Explicit(devices_, device_offsets));
 
+      // TODO(rongou): only copy the model once for all the batches.
       dh::ExecuteIndexShards(&shards_, [&](int idx, DeviceShard& shard) {
         shard.PredictInternal(batch, dmat->Info(), out_preds, model,
                               h_tree_segments, h_nodes, tree_begin, tree_end);
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -242,7 +242,7 @@ class GPUPredictor : public xgboost::Predictor {
     auto& offsets = *out_offsets;
     size_t n_shards = devices_.Size();
     offsets.resize(n_shards + 2);
-    size_t rows_per_shard = (batch_size + n_shards - 1) / n_shards; 
+    size_t rows_per_shard = (batch_size + n_shards - 1) / n_shards;
     for (int shard = 0; shard < devices_.Size(); ++shard) {
       offsets[shard] = batch_offset + shard * rows_per_shard * n_classes;
     }
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -225,13 +225,14 @@ class GPUPredictor : public xgboost::Predictor {
     auto& offsets = *out_offsets;
     offsets.resize(devices_.Size() + 1);
     offsets[0] = 0;
+    const size_t last = data.ConstHostVector().back();
 #pragma omp parallel for schedule(static, 1) if (devices_.Size() > 1)
     for (int shard = 0; shard < devices_.Size(); ++shard) {
       int device = devices_.DeviceId(shard);
       auto data_span = data.DeviceSpan(device);
       dh::safe_cuda(cudaSetDevice(device));
       if (data_span.size() == 0) {
-        offsets[shard + 1] = data.ConstHostVector().back();
+        offsets[shard + 1] = last;
       } else {
         // copy the last element from every shard
         dh::safe_cuda(cudaMemcpy(&offsets.at(shard + 1),
@@ -346,10 +347,13 @@ class GPUPredictor : public xgboost::Predictor {
 
     size_t batch_offset = 0;
     for (auto &batch : dmat->GetRowBatches()) {
-      std::vector<size_t> out_preds_offsets;
-      PredictionDeviceOffsets(out_preds->Size(), batch_offset, batch.Size(),
-                              model.param.num_output_group, &out_preds_offsets);
-      out_preds->Reshard(GPUDistribution::Explicit(devices_, out_preds_offsets));
+      bool is_external_memory = batch.Size() < dmat->Info().num_row_;
+      if (is_external_memory) {
+        std::vector<size_t> out_preds_offsets;
+        PredictionDeviceOffsets(out_preds->Size(), batch_offset, batch.Size(),
+                                model.param.num_output_group, &out_preds_offsets);
+        out_preds->Reshard(GPUDistribution::Explicit(devices_, out_preds_offsets));
+      }
 
       batch.offset.Shard(GPUDistribution::Overlap(devices_, 1));
       std::vector<size_t> device_offsets;
@@ -400,6 +404,7 @@ class GPUPredictor : public xgboost::Predictor {
     size_t n_classes = model.param.num_output_group;
     size_t n = n_classes * info.num_row_;
     const HostDeviceVector<bst_float>& base_margin = info.base_margin_;
+    out_preds->Shard(GPUDistribution::Granular(devices_, n_classes));
     out_preds->Resize(n);
     if (base_margin.Size() != 0) {
       CHECK_EQ(out_preds->Size(), n);
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -221,18 +221,19 @@ class GPUPredictor : public xgboost::Predictor {
   };
 
  private:
-  void DeviceOffsets(const HostDeviceVector<size_t>& data, std::vector<size_t>* out_offsets) {
+  void DeviceOffsets(const HostDeviceVector<size_t>& data,
+                     size_t total_size,
+                     std::vector<size_t>* out_offsets) {
     auto& offsets = *out_offsets;
     offsets.resize(devices_.Size() + 1);
     offsets[0] = 0;
-    const size_t last = data.ConstHostVector().back();
 #pragma omp parallel for schedule(static, 1) if (devices_.Size() > 1)
     for (int shard = 0; shard < devices_.Size(); ++shard) {
       int device = devices_.DeviceId(shard);
       auto data_span = data.DeviceSpan(device);
       dh::safe_cuda(cudaSetDevice(device));
       if (data_span.size() == 0) {
-        offsets[shard + 1] = last;
+        offsets[shard + 1] = total_size;
       } else {
         // copy the last element from every shard
         dh::safe_cuda(cudaMemcpy(&offsets.at(shard + 1),
@@ -357,7 +358,7 @@ class GPUPredictor : public xgboost::Predictor {
 
       batch.offset.Shard(GPUDistribution::Overlap(devices_, 1));
       std::vector<size_t> device_offsets;
-      DeviceOffsets(batch.offset, &device_offsets);
+      DeviceOffsets(batch.offset, batch.data.Size(), &device_offsets);
       batch.data.Reshard(GPUDistribution::Explicit(devices_, device_offsets));
 
       // TODO(rongou): only copy the model once for all the batches.
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -268,6 +268,9 @@ class GPUPredictor : public xgboost::Predictor {
      const thrust::host_vector<size_t>& h_tree_segments,
      const thrust::host_vector<DevicePredictionNode>& h_nodes,
      size_t tree_begin, size_t tree_end) {
+      if (predictions->DeviceSize(device_) == 0) {
+        return;
+      }
       dh::safe_cuda(cudaSetDevice(device_));
       nodes_.resize(h_nodes.size());
       dh::safe_cuda(cudaMemcpyAsync(dh::Raw(nodes_), h_nodes.data(),
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -340,17 +340,15 @@ class GPUPredictor : public xgboost::Predictor {
     for (auto &batch : dmat->GetRowBatches()) {
       GPUSet devices = GPUSet::All(param_.gpu_id, param_.n_gpus, batch.Size());
       ConfigureShards(devices);
-
-      batch.offset.Reshard(GPUDistribution::Overlap(devices_, 1));
-
-      std::vector<size_t> device_offsets;
-      DeviceOffsets(batch.offset, &device_offsets);
-      batch.data.Reshard(GPUDistribution::Explicit(devices_, device_offsets));
-
       std::vector<size_t> out_preds_offsets;
       PredictionDeviceOffsets(out_preds->Size(), batch_offset, batch.Size(),
                               model.param.num_output_group, &out_preds_offsets);
       out_preds->Reshard(GPUDistribution::Explicit(devices_, out_preds_offsets));
+
+      batch.offset.Reshard(GPUDistribution::Overlap(devices_, 1));
+      std::vector<size_t> device_offsets;
+      DeviceOffsets(batch.offset, &device_offsets);
+      batch.data.Reshard(GPUDistribution::Explicit(devices_, device_offsets));
 
       dh::ExecuteIndexShards(&shards_, [&](int idx, DeviceShard& shard) {
         shard.PredictInternal(batch, dmat->Info(), out_preds, model,
@@ -358,6 +356,11 @@ class GPUPredictor : public xgboost::Predictor {
       });
       batch_offset += batch.Size() * model.param.num_output_group;
     }
+
+    GPUSet devices = GPUSet::All(param_.gpu_id, param_.n_gpus, dmat->Info().num_row_);
+    ConfigureShards(devices);
+    out_preds->Reshard(GPUDistribution::Granular(devices_, model.param.num_output_group));
+
     monitor_.StopCuda("DevicePredictInternal");
   }
 
@@ -393,7 +396,6 @@ class GPUPredictor : public xgboost::Predictor {
     size_t n_classes = model.param.num_output_group;
     size_t n = n_classes * info.num_row_;
     const HostDeviceVector<bst_float>& base_margin = info.base_margin_;
-    out_preds->Shard(GPUDistribution::Granular(devices_, n_classes));
     out_preds->Resize(n);
     if (base_margin.Size() != 0) {
       CHECK_EQ(out_preds->Size(), n);
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -237,14 +237,19 @@ class GPUPredictor : public xgboost::Predictor {
     }
   }
 
+  // This function populates the explicit offsets that can be used to create a window into the
+  // underlying host vector. The window starts from the `batch_offset` and has a size of
+  // `batch_size`, and is sharded across all the devices. Each shard is granular depending on
+  // the number of output classes `n_classes`.
   void PredictionDeviceOffsets(size_t total_size, size_t batch_offset, size_t batch_size,
                                int n_classes, std::vector<size_t>* out_offsets) {
     auto& offsets = *out_offsets;
     size_t n_shards = devices_.Size();
     offsets.resize(n_shards + 2);
-    size_t rows_per_shard = (batch_size + n_shards - 1) / n_shards;
-    for (int shard = 0; shard < devices_.Size(); ++shard) {
-      offsets[shard] = batch_offset + shard * rows_per_shard * n_classes;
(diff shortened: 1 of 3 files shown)


src/predictor/gpu_predictor.cu
```diff
@@ -324,21 +336,27 @@ class GPUPredictor : public xgboost::Predictor {
...
+      GPUSet devices = GPUSet::All(param_.gpu_id, param_.n_gpus, batch.Size());
+      ConfigureShards(devices);
+
+      batch.offset.Reshard(GPUDistribution::Overlap(devices_, 1));
+
       std::vector<size_t> device_offsets;
       DeviceOffsets(batch.offset, &device_offsets);
       batch.data.Reshard(GPUDistribution::Explicit(devices_, device_offsets));
+
+      std::vector<size_t> out_preds_offsets;
```
sriramch: by changing the distribution of `out_preds` this way, callers expectation of the distribution within this host device vector - its granularity, size etc. may be violated *after* the predict api is invoked for any post processing such as transformation via objective function etc.

rongou: If we need the whole vector back on GPUs for further processing, we can always reshard it.

src/predictor/gpu_predictor.cu
```diff
@@ -324,21 +336,27 @@ class GPUPredictor : public xgboost::Predictor {
...
-      // out_preds have been sharded and resized in InitOutPredictions()
-      batch.offset.Shard(GPUDistribution::Overlap(devices_, 1));
+      GPUSet devices = GPUSet::All(param_.gpu_id, param_.n_gpus, batch.Size());
+      ConfigureShards(devices);
+
+      batch.offset.Reshard(GPUDistribution::Overlap(devices_, 1));
+
       std::vector<size_t> device_offsets;
       DeviceOffsets(batch.offset, &device_offsets);
       batch.data.Reshard(GPUDistribution::Explicit(devices_, device_offsets));
```
trivialfis: Is the granularity handled correctly?  Could you make a test for predicting multiclass output showing the correctness for granularity? Is the resharding in `InitOutPredictions()` still needed?

src/predictor/gpu_predictor.cu
sriramch: wouldn't this break existing code? let's say if the user does this
```
predictor->predict(out_predictions)
get_gradients(out_predictions, gradients)
do_tree_boost(gradients);
```
if the follow-up steps implicitly assumed the predictions to be shared based on number_of_rows/classes etc., will this change not break that code without the user having to recode to reshard?

rongou: Add a final resharding step to preserve the current behavior.

src/predictor/gpu_predictor.cu
rongou: Should be as `n_classes` is taken into consideration for the explicit offsets. I'll look at adding a test for multiclass.

Removed the sharding in `InitOutPredictions()`.

RAMitchell:
I like this PR based on its simplicity. @rongou @sriramch can one of you please do an experiment comparing the performance of both of your PRs checking for large differences in performance. It sometimes hard to see exactly how many memory copies are occurring when using HostDeviceVector.

src/predictor/gpu_predictor.cu
```diff
@@ -324,22 +336,31 @@ class GPUPredictor : public xgboost::Predictor {
...
+
       dh::ExecuteIndexShards(&shards_, [&](int idx, DeviceShard& shard) {
-        shard.PredictInternal(batch, dmat->Info(), out_preds, batch_offset, model,
+        shard.PredictInternal(batch, dmat->Info(), out_preds, model,
                               h_tree_segments, h_nodes, tree_begin, tree_end);
       });
       batch_offset += batch.Size() * model.param.num_output_group;
-      i_batch++;
     }
+
```
sriramch: how about extracting the existing distribution before iterating over the batches and reinstating it after?

before looping:
```
const GPUDistribution dist = out_preds->Distribution();
```

after looping:
```
out_preds->Reshard(dist);
```

this way, this method need not worry about how it was sharded before.

src/predictor/gpu_predictor.cu
```diff
@@ -237,6 +237,19 @@ class GPUPredictor : public xgboost::Predictor {
     }
   }
 
+  void PredictionDeviceOffsets(size_t total_size, size_t batch_offset, size_t batch_size,
```
sriramch: i do not know how b3bdc4e stabilized the build by reduced the test data size. the number of rows inside the dmatrix that the test exercises are tiny, that removing and/or adding dmatrixes should not have made any difference. it is likely that the tests were run on slaves that have lesser gpu's that are masking issues.

src/predictor/gpu_predictor.cu
```diff
@@ -237,6 +237,19 @@ class GPUPredictor : public xgboost::Predictor {
...
+                               int n_classes, std::vector<size_t>* out_offsets) {
+    auto& offsets = *out_offsets;
+    size_t n_shards = devices_.Size();
+    offsets.resize(n_shards + 2);
+    size_t rows_per_shard = (batch_size + n_shards - 1) / n_shards;
+    for (int shard = 0; shard < devices_.Size(); ++shard) {
+      offsets[shard] = batch_offset + shard * rows_per_shard * n_classes;
+    }
+    offsets[n_shards] = batch_offset + batch_size * n_classes;
```
sriramch: if the batch_size is fairly close to the number of gpu's (say, slightly more), then `offsets[n_shards]` can be less than the one before it. this is possible for the last batch in the dmatrix

src/predictor/gpu_predictor.cu
```diff
@@ -324,22 +336,31 @@ class GPUPredictor : public xgboost::Predictor {
...
     for (auto &batch : dmat->GetRowBatches()) {
-      CHECK(i_batch == 0 || devices_.Size() == 1) << "External memory not supported for multi-GPU";
-      // out_preds have been sharded and resized in InitOutPredictions()
-      batch.offset.Shard(GPUDistribution::Overlap(devices_, 1));
+      GPUSet devices = GPUSet::All(param_.gpu_id, param_.n_gpus, batch.Size());
+      ConfigureShards(devices);
+      std::vector<size_t> out_preds_offsets;
+      PredictionDeviceOffsets(out_preds->Size(), batch_offset, batch.Size(),
+                              model.param.num_output_group, &out_preds_offsets);
+      out_preds->Reshard(GPUDistribution::Explicit(devices_, out_preds_offsets));
```
sriramch: i *think* this should consider the possibility that some shards may not end up getting any rows (0), and that code that relies on a non trivial number for the number of rows have to change to accommodate this possibility, which is what the other pr does.

sriramch:
@RAMitchell thanks for your review. this 'pr' can optimize away those copies for the batch prediction which the other 'pr' does. but, i think this 'pr' still has to work for smaller batch sizes. i have provided some comments to this effect earlier.

src/predictor/gpu_predictor.cu
```diff
@@ -237,6 +237,19 @@ class GPUPredictor : public xgboost::Predictor {
     }
   }
 
+  void PredictionDeviceOffsets(size_t total_size, size_t batch_offset, size_t batch_size,
+                               int n_classes, std::vector<size_t>* out_offsets) {
+    auto& offsets = *out_offsets;
+    size_t n_shards = devices_.Size();
+    offsets.resize(n_shards + 2);
+    size_t rows_per_shard = (batch_size + n_shards - 1) / n_shards;
```
canonizer: `dh::DivRoundUp()` does that.

src/predictor/gpu_predictor.cu
```diff
@@ -237,6 +237,19 @@ class GPUPredictor : public xgboost::Predictor {
...
+    for (int shard = 0; shard < devices_.Size(); ++shard) {
+      offsets[shard] = batch_offset + shard * rows_per_shard * n_classes;
+    }
+    offsets[n_shards] = batch_offset + batch_size * n_classes;
+    offsets[n_shards + 1] = total_size;
```
canonizer: Doesn't this mean that the last device gets a disproportionately large number of rows? Won't it lead to out-of-memory errors?

src/predictor/gpu_predictor.cu
```diff
@@ -324,22 +336,31 @@ class GPUPredictor : public xgboost::Predictor {
...
+      out_preds->Reshard(GPUDistribution::Explicit(devices_, out_preds_offsets));
+
+      batch.offset.Reshard(GPUDistribution::Overlap(devices_, 1));
       std::vector<size_t> device_offsets;
       DeviceOffsets(batch.offset, &device_offsets);
       batch.data.Reshard(GPUDistribution::Explicit(devices_, device_offsets));
+
       dh::ExecuteIndexShards(&shards_, [&](int idx, DeviceShard& shard) {
-        shard.PredictInternal(batch, dmat->Info(), out_preds, batch_offset, model,
+        shard.PredictInternal(batch, dmat->Info(), out_preds, model,
```
canonizer: This actually copies the trees to the GPU for every SparsePage. Don't we want to copy them once instead?

src/predictor/gpu_predictor.cu
```diff
@@ -324,22 +336,31 @@ class GPUPredictor : public xgboost::Predictor {
...
     }
 
-    size_t i_batch = 0;
     size_t batch_offset = 0;
     for (auto &batch : dmat->GetRowBatches()) {
-      CHECK(i_batch == 0 || devices_.Size() == 1) << "External memory not supported for multi-GPU";
-      // out_preds have been sharded and resized in InitOutPredictions()
-      batch.offset.Shard(GPUDistribution::Overlap(devices_, 1));
+      GPUSet devices = GPUSet::All(param_.gpu_id, param_.n_gpus, batch.Size());
+      ConfigureShards(devices);
```
canonizer: For GPUPredictor, shards don't have any data-dependent state, only the trees and the device number. It would be clearer to remove `ConfigureShards()`, and filter out cases where a shard has 0 rows to predict.

src/predictor/gpu_predictor.cu
rongou: Changed the test a bit to test multi-class output.

src/predictor/gpu_predictor.cu
rongou: Not sure if we need to do this, as we used to shard `out_preds` ourselves in `InitOutPredictions()`.

src/predictor/gpu_predictor.cu
rongou: Agreed, not sure what the CI is doing.

src/predictor/gpu_predictor.cu
sriramch: the existing code may have assumed this behavior (or explicitly shard with with similar configuration, which may break) - i.e. the resultant out_preds has been sharded based on the number of rows and number of classes. although, this specific change of extracting and reinstating the distribution may no longer be required, as you removed the sharding from `InitOutPredictions`, it should be required to preserve current behavior; else, clients may not be able to use this unless they reshard it based on previous assumptions.

rongou:
@RAMitchell @sriramch I ran some experiments on these two PRs using 1 billion rows, written to a tmpfs directory on a GCP VM with 4x T4 GPUs.

Timing of the `PredictBatch()` method:

This PR:
18.972 s
17.855 s
18.054 s

#4437:
20.842 s
20.123 s
19.931 s

Looks like this PR is slightly faster.

src/predictor/gpu_predictor.cu
rongou: Right, the final resharding after all the batches are done should preserve current behavior.

src/predictor/gpu_predictor.cu
rongou: Now limit it to `batch_size`.

src/predictor/gpu_predictor.cu
rongou: Added checks.

src/predictor/gpu_predictor.cu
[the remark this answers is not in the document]

src/predictor/gpu_predictor.cu
rongou: I'm actually creating a vector of size `n_shards + 2`, so the last shard ends at the second to last element.

src/predictor/gpu_predictor.cu

src/predictor/gpu_predictor.cu

src/predictor/gpu_predictor.cu
rongou: Good point. Added a TODO, will send a followup PR.

rongou:
@canonizer @RAMitchell @sriramch @trivialfis all the comments have been addressed. Please take another look.

src/predictor/gpu_predictor.cu
```diff
@@ -225,16 +225,39 @@ class GPUPredictor : public xgboost::Predictor {
...
     for (int shard = 0; shard < devices_.Size(); ++shard) {
       int device = devices_.DeviceId(shard);
       auto data_span = data.DeviceSpan(device);
       dh::safe_cuda(cudaSetDevice(device));
-      // copy the last element from every shard
-      dh::safe_cuda(cudaMemcpy(&offsets.at(shard + 1),
-                               &data_span[data_span.size()-1],
-                               sizeof(size_t), cudaMemcpyDeviceToHost));
+      if (data_span.size() == 0) {
+        offsets[shard + 1] = host_data[data.Size() - 1];
```
sriramch: nitpick: wouldn't `data.ConstHostVector().back()` work?

src/predictor/gpu_predictor.cu
```diff
@@ -246,11 +269,13 @@ class GPUPredictor : public xgboost::Predictor {
     void PredictInternal
     (const SparsePage& batch, const MetaInfo& info,
      HostDeviceVector<bst_float>* predictions,
-     const size_t batch_offset,
      const gbm::GBTreeModel& model,
      const thrust::host_vector<size_t>& h_tree_segments,
      const thrust::host_vector<DevicePredictionNode>& h_nodes,
      size_t tree_begin, size_t tree_end) {
+      if (predictions->DeviceSize(device_) == 0) {
```
sriramch: the couple of lines from line# 296 can be hoisted out and moved here before the trees are copied to the gpu.

tests/cpp/predictor/test_gpu_predictor.cu
```diff
@@ -233,6 +231,33 @@ TEST(gpu_predictor, MGPU_Test) {
...
+  std::unique_ptr<Predictor> gpu_predictor =
+      std::unique_ptr<Predictor>(Predictor::Create("gpu_predictor"));
+  gpu_predictor->Init({std::pair<std::string, std::string>("n_gpus", "-1")}, {});
+
+  gbm::GBTreeModel model = CreateTestModel();
+  const int n_classes = 3;
+  model.param.num_output_group = n_classes;
+  std::vector<std::unique_ptr<DMatrix>> dmats;
+  dmats.push_back(CreateSparsePageDMatrix(9, 64UL));
+  dmats.push_back(CreateSparsePageDMatrix(128, 128UL));
```
sriramch: perhaps, you could add the other dmatrix (that was removed earlier) here now.

src/predictor/gpu_predictor.cu
```diff
@@ -225,16 +225,39 @@ class GPUPredictor : public xgboost::Predictor {
     auto& offsets = *out_offsets;
     offsets.resize(devices_.Size() + 1);
     offsets[0] = 0;
+    auto host_data = data.ConstHostVector();
```
RAMitchell: I believe this copies the data, is this intended? Use of auto is tricky when dealing with references: https://stackoverflow.com/questions/7138588/c11-auto-what-if-it-gets-a-constant-reference

src/predictor/gpu_predictor.cu

src/predictor/gpu_predictor.cu
rongou: Hmm, the offsets are sharded using the standard `Overlap` distribution. Looks like that doesn't handle the last batch correctly. This check should be equivalent.

tests/cpp/predictor/test_gpu_predictor.cu
rongou: Done.

src/predictor/gpu_predictor.cu
rongou: No longer doing this as per @sriramch 's suggestion.

src/predictor/gpu_predictor.cu
sriramch: oh - what i meant was to add that check also here and bail out early, as opposed to building the tree on gpu, check and bailing out. sorry for the super unhelpful comment earlier.

rongou: The second check is actually no longer needed. Removed.

rongou:
@hcho3 looks like the build machine ran out of disk space?
```
[2019-05-09T02:17:46.935Z] tar: .m2/repository/commons-httpclient/commons-httpclient/3.1/commons-httpclient-3.1.pom: Cannot write: No space left on device
```

src/predictor/gpu_predictor.cu
```diff
@@ -230,11 +230,33 @@ class GPUPredictor : public xgboost::Predictor {
       int device = devices_.DeviceId(shard);
       auto data_span = data.DeviceSpan(device);
       dh::safe_cuda(cudaSetDevice(device));
-      // copy the last element from every shard
-      dh::safe_cuda(cudaMemcpy(&offsets.at(shard + 1),
-                               &data_span[data_span.size()-1],
-                               sizeof(size_t), cudaMemcpyDeviceToHost));
+      if (data_span.size() == 0) {
+        offsets[shard + 1] = data.ConstHostVector().back();
```
canonizer: `ConstHostVector()` (and sync to host, in general), is not safe to call in multi-threaded code.

src/predictor/gpu_predictor.cu
```diff
@@ -375,7 +400,6 @@ class GPUPredictor : public xgboost::Predictor {
     size_t n_classes = model.param.num_output_group;
     size_t n = n_classes * info.num_row_;
     const HostDeviceVector<bst_float>& base_margin = info.base_margin_;
-    out_preds->Shard(GPUDistribution::Granular(devices_, n_classes));
```
canonizer: If `out_preds` are not sharded, copying/filling will take place on host. This is something we'd like to avoid, at least when not using external memory.

src/predictor/gpu_predictor.cu
canonizer: I've looked into it in more detail. This actually satisfies the check that the last element of the offsets must be equal to the size of the HostDeviceVector. The check that `CHECK_EQ(offsets.size() + 1, devices.Size())` has been implied, but apparently never inserted. Your approach enables to avoid having GPU memory corresponding to the tail of HostDeviceVector.

This looks a bit hacky to me. However, if you decide to proceed with this option, please document this in `HostDeviceVector`, and add a test so that this option is not broken.

A more general alternative would be to have only a window of the whole HostDeviceVector visible on the GPU. However, this would make the implementation even more complicated.

src/predictor/gpu_predictor.cu
```diff
@@ -324,22 +344,27 @@ class GPUPredictor : public xgboost::Predictor {
...
 
-    size_t i_batch = 0;
     size_t batch_offset = 0;
     for (auto &batch : dmat->GetRowBatches()) {
-      CHECK(i_batch == 0 || devices_.Size() == 1) << "External memory not supported for multi-GPU";
-      // out_preds have been sharded and resized in InitOutPredictions()
+      std::vector<size_t> out_preds_offsets;
+      PredictionDeviceOffsets(out_preds->Size(), batch_offset, batch.Size(),
+                              model.param.num_output_group, &out_preds_offsets);
+      out_preds->Reshard(GPUDistribution::Explicit(devices_, out_preds_offsets));
```
canonizer: 1. Resharding will also copy back to host predictions from previous batches. However, this should be fine if your performance measurements don't show any problems.

2. A distribution with offsets is not equal to an implicit distribution with the same offsets. Please add avoiding extra copying in this situation in `Shard()` and `Reshard()`. No additional copying should happen when external memory is not used.

src/predictor/gpu_predictor.cu
rongou: Extracted out the value. Would this work?

src/predictor/gpu_predictor.cu
rongou: Added back.

src/predictor/gpu_predictor.cu
rongou: For 1, I think the copying needs to happen since we are sliding over the predictions on the GPUs.

For 2, added a guard to avoid unnecessary resharding.

hcho3:
@rongou I will go ahead and expand disk space for the slave workers.

src/predictor/gpu_predictor.cu
rongou: Yeah I was thinking we could add a `Windowed` distribution to `GPUDistribution` to support sliding over a vector from GPUs. Would that make sense? Probably in a followup PR.

src/predictor/gpu_predictor.cu
```diff
@@ -225,13 +225,14 @@ class GPUPredictor : public xgboost::Predictor {
     auto& offsets = *out_offsets;
     offsets.resize(devices_.Size() + 1);
     offsets[0] = 0;
+    const size_t last = data.ConstHostVector().back();
```
canonizer: This may still cause unnecessary creation of a CPU copy if the data is originally present only on the GPU.

On the other hand, this value is equal (at least in the single-batch case) to `batch.data.Size()`, as it is the total number of entries. Could you use that value instead?

src/predictor/gpu_predictor.cu
canonizer: `Windowed` would go against 1:1 (mostly) correspondence between the contents on the host and on the GPU, which is something we'd like to keep.

Instead, something like a block-cyclic distribution may be better.

However, please add documentation to HostDeviceVector or GPUDistribution that you are using the case where `offsets.size() + 1 != devices.Size()`.

rongou: Added a note to `GPUDistribution::Explicit()`.

hcho3:
@rongou @RAMitchell I think this PR should be part of 0.90, since it's a follow up to #4284. Can we merge this now?

Issue #4422, opened 2019-04-30
[jvm-packages] false should be the default value of maximize_evaluation_metrics


david-liu:
https://github.com/dmlc/xgboost/blob/253fdd8a42d5ec6b819788199584d27bf9ea6253/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala#L161

I use spark `XGBoostRegressor` to train model and set the NumEarlyStoppingRounds and the metric is ` rmse`, but my job throw exception:
```
19/04/30 01:42:13 WARN TaskSetManager: Lost task 44.0 in stage 69.0 (TID 5371, hdc42-mcc10-01-1110-4204-067-tess0029.stratus.rno.ebay.com, executor 52): java.lang.IllegalArgumentException: maximize_evaluation_metrics has to be specified
```
which means the `XGBoostRegressor`  did not set the default value for `maximize_evaluation_metrics`. I think  the default value should be `false`, considering most matrics are decresing the loss, event the use did not set it (Except `AUC`)

hcho3:
@david-liu This behavior is documented in https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html#early-stopping. I do agree that automatically setting `maximize_evaluation_metrics` would be a nice feature to have. Feel free to submit a pull request.

PR #4446, opened 2019-05-08
[jvm-packages] Automatically set maximize_evaluation_metrics if not explicitly given in XGBoost4J-Spark

shishaochen:
This pull request is aimed at resolving "[jvm-packages] false should be the default value of maximize_evaluation_metrics"(#4422).
The metric names to maximize are aligned with [xgboost.callback.early_stop](https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/callback.py#L189) in the Python package.
To avoid duplicating codes in both `XGBoostClassifier` and `XGBoostRegressor`, parameter `maximize_evaluation_metrics` is set in [XGBoost.scala](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala) when not provided by user.

closes https://github.com/dmlc/xgboost/issues/4422

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -24,6 +24,7 @@ import scala.util.Random
 
 import ml.dmlc.xgboost4j.java.{IRabitTracker, Rabit, XGBoostError, RabitTracker => PyRabitTracker}
 import ml.dmlc.xgboost4j.scala.rabit.RabitTracker
+import ml.dmlc.xgboost4j.scala.spark.params.LearningTaskParams
 import ml.dmlc.xgboost4j.scala.{XGBoost => SXGBoost, _}
 import ml.dmlc.xgboost4j.{LabeledPoint => XGBLabeledPoint}
 import org.apache.commons.io.FileUtils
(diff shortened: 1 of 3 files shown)


CodingCat:
thanks, would you please also update https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html#early-stopping accordingly?

shishaochen:
@CodingCat Document is updated in new commits.
When `custom_eval` is set, `maximize_evaluation_metrics` is still required.

CodingCat:
@hcho3 jenkins says `[2019-05-09T03:31:04.298Z] E: You don't have enough free space in /var/cache/apt/archives/.`?

hcho3:
@CodingCat Looks like we need to expand the volume for the slave node

shishaochen:
It seems that 2 failures of checks are not caused by my code change but the disk volume.
Coming back to this PR, is there anything else need I do? @CodingCat

hcho3:
I’m looking at the disk space issue.

hcho3:
@CodingCat @rongou I went ahead and increased the disk space.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -160,7 +160,7 @@ object XGBoost extends Serializable {
         .map(_.toString.toInt).getOrElse(0)
       val overridedParams = if (numEarlyStoppingRounds > 0 &&
           !params.contains("maximize_evaluation_metrics")) {
-        if params.contains("custom_eval") {
+        if (params.contains("custom_eval")) {
```
SpringHerald: when `num_early_stopping_rounds` is non-zero, `eval_metric` is a built-in metric,  the `custom_eval` is still in params, just the value is null, so it will still throw an exception. 
So the if condition should be `if (params.contains("custom_eval") && params("custom_eval") != null)` or remove `custom_eval` from params.

CodingCat: Hmmm....I think I missed some part of this PR

The reality is early stopping doesn’t work with custom metrics

PR #4447, opened 2019-05-08
Enable building with shared NCCL.

### LINKED ISSUE #4445: Allowing dynamic linking of NCCL
Appears that [NCCL's static library is used](  https://github.com/dmlc/xgboost/blob/ade3f302373cf630f5115aca26cb40ea9cac78ac/cmake/modules/FindNccl.cmake#L36 ) when building xgboost. Am curious if it would be reasonable to allow users to dynamically link instead or as an alternative to static linking.

trivialfis:
close #4445

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,8 +25,12 @@ set(NVTX_HEADER_DIR "" CACHE PATH "Path to the stand-alone nvtx header")
 ## CUDA
 option(USE_CUDA  "Build with GPU acceleration" OFF)
 option(USE_NCCL  "Build with NCCL to enable multi-GPU support." OFF)
+option(BUILD_WITH_SHARED_NCCL "Build with shared NCCL library." OFF)
 set(GPU_COMPUTE_VER "" CACHE STRING
   "Semicolon separated list of compute versions to be built against, e.g. '35;61'")
+if (BUILD_WITH_SHARED_NCCL AND (NOT USE_NCCL))
+  message(SEND_ERROR "Build XGBoost with -DUSE_NCCL=ON to enable BUILD_WITH_SHARED_NCCL.")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -25,8 +25,12 @@ set(NVTX_HEADER_DIR "" CACHE PATH "Path to the stand-alone nvtx header")
 ## CUDA
 option(USE_CUDA  "Build with GPU acceleration" OFF)
 option(USE_NCCL  "Build with NCCL to enable multi-GPU support." OFF)
+option(BUILD_WITH_SHARED_NCCL "Build with shared NCCL library." OFF)
 set(GPU_COMPUTE_VER "" CACHE STRING
   "Semicolon separated list of compute versions to be built against, e.g. '35;61'")
+if (BUILD_WITH_SHARED_NCCL AND (NOT USE_NCCL))
+  message(SEND_ERROR "Build XGBoost with -DUSE_NCCL=ON to enable BUILD_WITH_SHARED_NCCL.")
```
hcho3: Why `SEND_ERROR`, not `FATAL_ERROR`?

hcho3:
@jakirkham Can you try this pull request and see if it works for your use case?

CMakeLists.txt
trivialfis: @hcho3 FATAL_ERROR stops immediately, SEND_ERROR can continue other checking so that user can collect all errors in one go.

trivialfis:
@jakirkham Combined with `NCCL_ROOT` env variable, this should be able to use system NCCL.  If there are other issues in the platform you are packaging for please let me know, maybe I can help with that.

jakirkham:
Thanks for working on this. Sorry for being somewhat slow to reply. The patch does work nicely. Really appreciate the effort here.

PR #4448, opened 2019-05-09
add support for training on gpu with external memory

sriramch:
- the sparse pages are taken in batches and the compressed histogram indexes and
     the cut matrices are incrementally built
   - also snuck in some (perf) changes related to sketches aggregation amongst multiple
     features across multiple sparse page batches
- note: this is v0 of this feature, and i'm well aware that more could be done. there will
  be follow-up pr's that iteratively improves on this over time
- there is a boost in performance (during training initialization time) and reduction in
  peak memory usage when enabling external memory training and prediction - the latter at the
  expense of speed (which is to be expected). the boost is greater on wider datasets. 

@RAMitchell @canonizer @rongou please review this.


sriramch:
closing and reopening to trigger a build to fix [this](https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-4448/9/pipeline) transient build failure

sriramch:
another transient failure on cpu - hence, re-triggering it

src/common/hist_util.h
```diff
@@ -115,10 +117,24 @@ struct HistCutMatrix {
...
 };
 
+/*!
+ * \brief A container that holds the device sketches across all
+ *  sparse page batches which are distributed to different devices.
+ *  As sketches are aggregated by column, the atomic flag simulates
+ *  a lock in user land when multiple devices could be pushing
+ *  sketch summary for the same column across distinct rows.
+ */
+struct SketchContainer {
```
rongou: I wonder how this works in a distributed/multi-node setup. Have you tried to run this with multiple workers?

sriramch: is this specific to this code fragment or a question in general?

typically, in a distributed setup, wouldn't the dmatrix be appropriately created by splitting the rows across multiple workers, with different parts of the file going to different workers. at this phase, this only deals with the set of rows it has been handed over. 

i haven't tried this in a multi-node/distributed training environment.

src/common/hist_util.cu
```diff
@@ -324,9 +316,27 @@ struct GPUSketcher {
...
+          explicit ColLocker(std::atomic_flag *lock)
+            : col_lock_(lock) {
+              while (col_lock_->test_and_set()) {}
+            }
+          ~ColLocker() {
+            col_lock_->clear();
+          }
+         private:
+          std::atomic_flag *col_lock_;
+        }lock(sketch_container_->col_locks_[icol].get());
```
rongou: Hmm I guess I don't quite understand this code. Since you are looping through the columns, each thread is working on a different block of columns, why do you still need to lock on the column?

sriramch: rows in a batch are split across multiple devices and the rows across those devices can be aggregating sketches for the same column.

src/common/hist_util.h
rongou: Yeah I guess the question is not specific to this code fragment. But if in a distributed setup each worker only works with its local data shard, perhaps in external memory mode we also don't need to  accumulate quantiles across all the batches? If a single batch is representative enough, maybe that's all we need? Just a thought, not sure if it'll work out. :)

rongou:
LGTM overall, but @canonizer and @RAMitchell should review this more thoroughly. :)

src/common/hist_util.cu
rongou: Oh right, this is in a device shard. Since each shard needs to process all the columns, wouldn't this introduce a lot of contention?

hcho3:
This PR will want to wait after 0.90 version, which is coming out end of this week

src/common/hist_util.cu
sriramch: sure - if all the shards tries to push the summary for the same column at the same time. from whatever profiling i have done on xgboost for some of the datasets i tried, the hotspots in this method seem to be in computing the column cuts and presizing the quantile sketch summary. since we are spinning here, it will burn some cpu if they are contending as opposed to a more expensive lock acquisition via a mutex.

src/common/hist_util.cu
```diff
@@ -324,9 +316,27 @@ struct GPUSketcher {
...
+#pragma omp parallel for schedule(static)
       for (int icol = 0; icol < num_cols_; ++icol) {
-        summaries_[icol].MakeFromSorted(&cuts_h_[n_cuts_ * icol], n_cuts_cur_[icol]);
-        sketches_[icol].PushSummary(summaries_[icol]);
+        WXQSketch::SummaryContainer summary;
+        summary.Reserve(n_cuts_);
+        summary.MakeFromSorted(&cuts_h_[n_cuts_ * icol], n_cuts_cur_[icol]);
+
+        // Spin in user land until we we can grab this lock
+        struct ColLocker {
```
canonizer: Please use `std::mutex`.

src/tree/updater_gpu_hist.cu
```diff
@@ -1444,23 +1455,101 @@ class GPUHistMakerSpecialised{
                                           column_sampling_seed));
         });
 
-    // Find the cuts.
+    // Initialize Sketches across all batches
+    common::SketchContainer sketch_container;
+    sketch_container.sketches_.resize(info_->num_col_);
+    sketch_container.col_locks_.resize(info_->num_col_);
+#pragma omp parallel for schedule(static)
```
canonizer: This is a fairly simple initialization. Does it need OpenMP parallelization?

src/common/hist_util.h
```diff
@@ -115,10 +117,24 @@ struct HistCutMatrix {
...
+  std::vector<HistCutMatrix::WXQSketch> sketches_;
+  std::vector<std::unique_ptr<std::atomic_flag>> col_locks_;
```
canonizer: Please use `std::mutex`.

src/common/hist_util.h
canonizer: This structure definitely belongs to the internals of GPUSketcher, not to its public interface. Could you move it there?

src/tree/updater_gpu_hist.cu
```diff
@@ -1444,23 +1455,101 @@ class GPUHistMakerSpecialised{
...
+    size_t row_stride(0);  // Max row_stride across the entire dataset
+    size_t processed_size(0);
+
+    // Rows allocated to each GPU in a batch
+    std::vector<std::vector<size_t>> shard_allocations;
     monitor_.StartCuda("Quantiles");
-    common::DeviceSketch(batch, *info_, param_, &hmat_, hist_maker_param_.gpu_batch_nrows);
-    n_bins_ = hmat_.row_ptr.back();
+    for (const auto &batch : dmat->GetRowBatches()) {
+      struct RowReset {
```
canonizer: Please avoid declaring new data types inside function and method code, especially if those data types contain methods themselves. It makes understanding the code more difficult.

src/common/hist_util.cu
```diff
@@ -113,9 +112,9 @@ struct GPUSketcher {
 
    public:
     DeviceShard(int device, bst_uint row_begin, bst_uint row_end,
-                tree::TrainParam param) :
+                tree::TrainParam param, SketchContainer *sketch_container) :
```
canonizer: GPUSketcher already breaks SparsePage object into smaller batches. It should be possible to add another level of looping around that, so that quantile extraction with/without external memory is contained entirely within GPUSketcher.

RAMitchell:
Agree with @canonizer on keeping quantile generation self contained and splitting this into 2 PRs. Given some of your changes to quantile generation are performance related you should provide a high level description of how you have changed the algorithm and a few numbers to back up your optimisations. e.g. Quantile generation for large number of features is slow because xxx, modifying algorithm by xxx results in 2x speed up on this dataset and less memory usage.

sriramch:
a transient [build error](https://travis-ci.org/dmlc/xgboost/jobs/533334476) now. closing and reopening the pr to trigger a build.

sriramch:
training with external memory on cpu clearly has some memory corruption issues. the ci/cd pipeline demonstrates [this](https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-4448/19/pipeline/) and i have observed this in my testing as well. since this change has nothing to do with that, i'm triggering a build again, hoping(!) that this passes now.

sriramch:
@hcho3 thanks a lot for proactively looking over this and restarting the build to get over all those transient errors. i sincerely appreciate it. i wonder why these builds are so flaky. can some tests (specifically around cpu with external memory) be disabled until it is fixed? it takes some effort to stabilize the builds.

src/common/hist_util.cu
sriramch: done. i did this to avoid context switching when contention is greater and the work done in the critical section is smaller. i have changed this to keep it simple for now.

src/tree/updater_gpu_hist.cu
sriramch: i was trying to speed this up for wider datasets. i have changed this now to parallelize only when the number of features are > 1000. the openmp threads should be pre-created - right? i presume they aren't created/destroyed during every one of these parallel blocks.

src/common/hist_util.h
sriramch: done

src/tree/updater_gpu_hist.cu
sriramch: done - i have refactored this area of code.

src/common/hist_util.cu
sriramch: i have now made the quantile generation self contained within the gpu sketcher.

sriramch:
@canonizer @RAMitchell i have addressed your comments. please review them when you get a chance.

hcho3:
@sriramch Travis CI doesn't provide much memory, so tests often fail there. Fortunately, I managed to move almost all tests to Jenkins CI, where I can provision more memory. (The remaining two tests in Travis CI are for Mac OSX, so I couldn't move them.)

We definitely need to inspect external memory code to eliminate double-free errors.

src/tree/updater_gpu_hist.cu
```diff
@@ -1326,23 +1305,50 @@ inline void DeviceShard<GradientSumT>::InitCompressedData(
...
+  } else if (rows_per_batch == -1) {
+     gpu_batch_nrows = num_elems;
+  } else {
+     gpu_batch_nrows = std::min(num_elems, static_cast<size_t>(rows_per_batch));
+  }
+  const std::vector<Entry>& data_vec = row_batch.data.ConstHostVector();
 
   thrust::device_vector<Entry> entries_d(gpu_batch_nrows * row_stride);
-  size_t gpu_nbatches = dh::DivRoundUp(n_rows, gpu_batch_nrows);
+  size_t gpu_nbatches = dh::DivRoundUp(num_elems, gpu_batch_nrows);
```
RAMitchell: What does `num_elems` refer to here? It looks like it is referring to some number of rows but the word element normally refers to a single entry in a matrix not a row.

src/tree/updater_gpu_hist.cu
```diff
@@ -676,6 +677,7 @@ struct DeviceShard {
   bst_uint row_begin_idx;
   bst_uint row_end_idx;
   bst_uint n_rows;
+  bst_uint rows_processed;
```
RAMitchell: Any way to do this without a member variable? Its difficult for someone looking at the shard object to see what this does, what state it carries and if it is necessary to the rest of the algorithm.

src/tree/updater_gpu_hist.cu
```diff
@@ -1326,23 +1305,50 @@ inline void DeviceShard<GradientSumT>::InitCompressedData(
 
 template <typename GradientSumT>
 inline void DeviceShard<GradientSumT>::CreateHistIndices(
-    const SparsePage& row_batch, size_t row_stride, int null_gidx_value) {
+    const SparsePage &row_batch,
+    const common::HistCutMatrix &hmat,
+    const std::vector<size_t> &shard_allocations,
+    int rows_per_batch) {
+  // Has any been allocated for me in this batch?
+  size_t num_elems = shard_allocations[shard_idx];
```
RAMitchell: Nitpick: use vector.at() outside of hot loops for better error messages.

src/tree/updater_gpu_hist.cu
```diff
@@ -1444,23 +1453,64 @@ class GPUHistMakerSpecialised{
...
+
+    monitor_.StartCuda("BinningCompression");
+    size_t processed_size = 0;
+    // TODO(sriramch): Think of a way to utilize *all* the GPUs to build the
+    // compressed bins
+    for (const auto &batch : dmat->GetRowBatches()) {
+      std::vector<size_t> shard_allocations(shards_.size());
+      // Distribute the rows in this batch to the different shards
+      for (size_t i = 0; i < shards_.size(); ++i) {
+        // Does this batch pertain to me?
```
RAMitchell: This section of code is very difficult to read due to multiple nested conditions and min functions.

src/tree/updater_gpu_hist.cu
```diff
@@ -1444,23 +1453,64 @@ class GPUHistMakerSpecialised{
...
+        int64_t num_elems = processed_size + batch.Size() - shards_[i]->row_begin_idx;
```
RAMitchell: As before be careful using the word 'elems'.

src/tree/updater_gpu_hist.cu
sriramch: reworded.

src/tree/updater_gpu_hist.cu
sriramch: done without altering the device shard state

src/tree/updater_gpu_hist.cu
sriramch: are you suggesting because it has the potential to throw an exception?

src/tree/updater_gpu_hist.cu
sriramch: reworked the algorithm. hopefully it is clear now.

src/tree/updater_gpu_hist.cu
sriramch: done

sriramch:
>> As an extreme example you could even have a class who's job it is to manage partitioning of rows - not suggesting you necessarily do this but I think a cleaner implementation is possible.

[sc] i'll look into this in a future pr.

src/common/hist_util.cu
```diff
@@ -341,63 +357,91 @@ struct GPUSketcher {
...
-        shard->GetSummary(&summary, icol);
-        sketches[icol].PushSummary(summary);
-      }
+    size_t row_stride = 0;
+    for (const auto &shard : shards_) {
+      row_stride = std::max(row_stride, shard->GetRowStride());
     }
 
-    hmat->Init(&sketches, param_.max_bin);
+    return row_stride;
```
canonizer: Why are you accumulating the row stride here, and returning it?

src/common/hist_util.h
```diff
@@ -116,9 +116,9 @@ struct HistCutMatrix {
 };
 
 /*! \brief Builds the cut matrix on the GPU */
-void DeviceSketch
-  (const SparsePage& batch, const MetaInfo& info,
-   const tree::TrainParam& param, HistCutMatrix* hmat, int gpu_batch_nrows);
+size_t DeviceSketch
```
canonizer: Why are you returning a value from `DeviceSketch()`?

After some thinking, I presume that it is to avoid an extra pass through the data in external memory. If it is so, then this is a good decision. However, it is not obvious from the code, and definitely should be documented.

sriramch:
> Please, do split this pull request into two (one for quantiles, one for the histogram updater).

will do this - with quantiles coming first. i'll close this and create them

src/common/hist_util.cu
```diff
@@ -324,9 +335,14 @@ struct GPUSketcher {
 
       // add cuts into sketches
       thrust::copy(cuts_d_.begin(), cuts_d_.end(), cuts_h_.begin());
+#pragma omp parallel for schedule(static)
```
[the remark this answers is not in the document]
sriramch: i did this for wider datasets, and it was showing an improvement.

src/common/hist_util.h
```diff
@@ -116,9 +116,9 @@ struct HistCutMatrix {
...
+  (const tree::TrainParam& param, int gpu_batch_nrows, DMatrix *dmat,
+   std::vector<HistCutMatrix::WXQSketch> *sketches);
```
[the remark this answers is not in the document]
sriramch: will do...

src/common/hist_util.cu
```diff
@@ -341,63 +357,91 @@ struct GPUSketcher {
...
+
+    sketches->swap(sketch_container.sketches_);
+
+    return row_stride;
   }
 
  private:
   std::vector<std::unique_ptr<DeviceShard>> shards_;
-  tree::TrainParam param_;
-  GPUDistribution dist_;
```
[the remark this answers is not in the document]
sriramch: it is created now for every batch and passed in to the batch sketcher.

src/common/hist_util.cu
```diff
@@ -341,63 +357,91 @@ struct GPUSketcher {
...
+  const tree::TrainParam &param_;
```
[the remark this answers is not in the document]
sriramch: the training param is created by the caller and the life of the gpu sketcher is controlled by the caller. this indicates that the training param isn't owned by it and is immutable.

src/common/hist_util.cu
sriramch: this is needed for later during data initialization on the gpu. since we are going through the entire dmatrix here, it is easier to compute this in-situ as opposed to doing another iteration later only to compute this.

sriramch:
i'll split this into 2 pr's. hence, closing this...

Issue #4361, opened 2019-04-12
Unable to load model trained on multi-gpu machine, on single-gpu machine, 


kyoungrok0517:
I see the following error if I use below code to load the model. The model was trained on Linux multi-gpu machine. I'm trying to load the model on Windows single-gpu machine.

```
XGBoostError: b'[12:21:26] c:\\users\\administrator\\desktop\\xgboost\\src\\common\\common.h:182: Check failed: n_gpus <= n_available_devices (1 vs. 0) Starting from gpu id: 1, there are only 0 available devices, while n_gpus is set to: 1'
```

**[Code]**
```python
with open('./model/model_skipgram_2019-04-04T05-30-20.pkl', 'rb') as f:
    model = pickle.load(f)
```
**[Model]**
Model ([link](https://www.dropbox.com/s/iba259mom5m8usd/model_skipgram_2019-04-04T05-30-20.pkl?dl=0))

kyoungrok0517:
If I do `model.set_param('gpu_id', 0)` on multi-gpu machine then load on the single-gpu machine I could load the model. I think there's some bug in sanity checking code.

trivialfis:
@hcho3 Can we talk about our progress on Json at some point or give a roadmap on this?  I don't want to touch the parameter configuration before having a flexible model format  ..

hcho3:
@trivialfis I recall that parameter validation should be tied together with the JSON serialization. Did you get a chance to review the complete schema? Feel free to post at the JSON RFC thread if you have more detailed idea.

hcho3:
@trivialfis Right now, I am reviewing your CMake refactor PR and also revamping CI.

kyoungrok0517:
Any update on this issue??

hcho3:
@kyoungrok0517 I just tried loading the pickle you uploaded on p2.xlarge machine (single GPU), and I did not get the same error.

trivialfis:
closed by #4454 .

PR #4454, opened 2019-05-10
De-duplicate GPU parameters.

trivialfis:
* Only define `gpu_id` and `n_gpus` in Learner `TrainParam`
* Disable all GPU usage when GPU related parameters are not specified.


trivialfis:
@RAMitchell It turns out there are some numeric difference between CPU and CUDA transform weren't caught by the tests before.  After disabling GPU this test won't pass:
https://github.com/dmlc/xgboost/blob/be0f346ec907b9b8c76597c2ee308fcd53ba311a/tests/python-gpu/test_gpu_prediction.py#L13

I tried to diff the outputs from two different settings.  The difference mostly come from `GetGradient`, starts at 1e-6 and accumulates up.  To pass the test `rtol` needs to be changed to `1e-3` from `1e-5`.  Here is the mismatch rate with 5000x10 data set:
```
E               AssertionError: 
E               Not equal to tolerance rtol=0.0001, atol=0
E               
E               Mismatch: 0.04%
E                x: array([-0.108287,  0.163954,  0.096447, ..., -0.173721, -0.115342,
E                      -0.577273], dtype=float32)
E                y: array([-0.108287,  0.163954,  0.096447, ..., -0.173721, -0.115342,
E                      -0.577273], dtype=float32)
```

trivialfis:
Another issue is, I'm currently hacking the configuration in Learner TrainParam, honestly I don't fully understand what's happening in there.

trivialfis:
One last todo is either delay the configuration of Objective to after data is known, or resize the label checking result vector every time `GetGradient` is called.

trivialfis:
> it would be good to make sure the behaviour is expected upon serialisation/deserialisation of the model.

Good point, will try to make a test later.

> One feature that would be nice to have is to log information

Will add a `LOG(INFO)` in `GPUSet::Init`.

trivialfis:
@RAMitchell Actually I don't want to merge this PR.  I remember there were users who train different models on different GPUs.  Making a global variable will break it.  I need to think of something that doesn't break the functionality but also keep our implementation clean.

trivialfis:
Suggestions are welcomed.

trivialfis:
@hcho3 @RAMitchell I have a local branch that passes a pointer to `const LearnerTrainParam` to `Predictor`, `Metric`, `TreeUpdater`, `GBM` and `LinearUpdater` via `Create` method in respective class.  The change is quite massive and I would like to here about your opinions before proceeding.

The benefit of passing `LearnerTrainParam` is we can eliminate the duplication of `gpu_id`, `n_gpus` without creating a global variable.  As another benefit, `nthread` is also passed around,  creating an opportunity of eliminating global OpenMP variable.

The downside is the restructuring is massive, and all `Create` methods need to accept an addition parameter.

trivialfis:
Also, this is a blocker for JSon.

hcho3:
Current structure of parameter handling causes duplication, as learning parameters are duplicated into objectives, metrics, and updaters. So your proposal has merits. Let me think over this and get back to you.

trivialfis:
@hcho3 It also fixes the bug that XGBoost choosing GPU aggressively due to lack of global configuration.  Currently if the user didn't supply `ngpu=0` parameter XGBoost will run metrics and objectives on GPU by default.

tests/cpp/helpers.h
```diff
@@ -18,13 +18,22 @@
...
+#include "../../src/common/common.h"
 
 #if defined(__CUDACC__)
 #define DeclareUnifiedTest(name) GPU ## name
 #else
 #define DeclareUnifiedTest(name) name
 #endif
 
+#if defined(__CUDACC__)
+#define NGPUS() 1
```
canonizer: Could you use `#define NGPUS 1`, without `()`, as there are no parameters?

src/learner.cc
```diff
@@ -315,6 +261,23 @@ class LearnerImpl : public Learner {
...
+        << "Always call InitModel or LoadModel before any evaluation.";
+    this->ValidateDMatrix(dmat);
+    // Configure GPU parameters
+    // FIXME(trivialfis): How do we know dependent parameters are all set?
+    if (tparam_.tree_method == TreeMethod::kGPUHist ||
+        tparam_.tree_method == TreeMethod::kGPUExact ||
+        (cfg_.find("predictor") != cfg_.cend() &&
+         cfg_.at("predictor") == "gpu_predictor")) {
+      if (cfg_.find("n_gpus") == cfg_.cend()) {
+        tparam_.n_gpus = 1;
```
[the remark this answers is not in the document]
trivialfis: I think using a single GPU as default is safer.  Multiple GPUs might be slower if dataset is not large enough, or with a weird combo of GPU models.

include/xgboost/gbm.h
```diff
@@ -14,17 +14,22 @@
...
+#include <xgboost/generic_parameters.h>
 #include "../../src/common/host_device_vector.h"
 
 namespace xgboost {
 /*!
  * \brief interface of gradient boosting model.
  */
 class GradientBooster {
+ protected:
+  LearnerTrainParam const* learner_param_;
```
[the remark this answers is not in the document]
trivialfis: No copying is because `Learner`(booster) can change its parameters.  Users can try both GPU and CPU in a single script or REPL.

Smart pointers imply ownership sharing/transferring, but I don't want other components other than `Learner`(booster) to change the parameters, hence a raw pointer to constant.

include/xgboost/gbm.h
```diff
@@ -149,6 +154,7 @@ class GradientBooster {
    */
   static GradientBooster* Create(
       const std::string& name,
+      LearnerTrainParam const* gparam,
```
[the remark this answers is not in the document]
trivialfis: As above.

trivialfis:
@hcho3 Could you help taking a look in Jenkins' cache?  It seems clang-tidy is running on an old copy of this PR: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-4454/10/pipeline

In latest commit `gpu_predictor.cu:376:3` is not a constructor.

include/xgboost/gbm.h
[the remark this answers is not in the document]
canonizer: It looks like the same object can be pointed to by `std::shared_ptr<T>` and `std::shared_ptr<const T>`: constructor (9) (the templated version) in https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr appears to support this case.

Disclaimer: I'm not a C++ expert, so I may be wrong here.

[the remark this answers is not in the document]
hcho3: `std::shared_ptr<T>` indeed does not copy the underlying object. However, it performs reference counting, which involves atomic adds and subtracts, and thus adds overhead.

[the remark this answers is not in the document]
canonizer: Yes, it uses atomics, but only during initialization. Dereferencing and member access don't use them.

I'll leave the decision up to you.

trivialfis:
@RAMitchell @canonizer Could you please explain why shared pointer is better?

From my point of view,  objects like `Predictor` do not own `LearnerTrainParam`, so its "borrowing" a pointer from `Learner`.  That's like when you what to access the internal data of `std::vector` by calling `data()`, or `std::string` by calling `c_str()`, you get a raw (const) pointer.  With the returned raw pointer you are only borrowing its content, but should not manage the content of the pointer (deallocate) because the ownership is still in `std::string`/`std::vector`.

Passing `shared_ptr` means the ownership is shared.  It's a semantic issue.  See 
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-smart
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-ptr
 and related topics.

I'm open the suggestions, but I need more context to understand why a shared pointer is a better choice.

src/linear/updater_gpu_coordinate.cu
```diff
@@ -164,7 +164,7 @@ class GPUCoordinateUpdater : public LinearUpdater {
                       const gbm::GBLinearModelParam &model_param) {
     if (!shards_.empty()) return;
 
-    dist_ = GPUDistribution::Block(GPUSet::All(tparam_.gpu_id, tparam_.n_gpus,
+    dist_ = GPUDistribution::Block(GPUSet::All(learner_param_->gpu_id, learner_param_->n_gpus,
```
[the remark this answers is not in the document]
trivialfis: Agree, but I just realized that it requires moving `GPUSet` into headers.

trivialfis:
@hcho3 
>  Could you please help taking a look in Jenkins' cache?

Never mind, it works. Thanks.

src/learner.cc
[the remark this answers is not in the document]
trivialfis: Sorry for this.  GPUSet::All also accepts an additional parameter `num_rows` which limits the number of GPUs to be less than number of data rows.  So I think current way is more verbose but makes some logical sense. (I feel weird that `LearnerTrainParam` accept this parameter)

RAMitchell:
Obviously let's not merge this until after 0.9

trivialfis:
ping @hcho3

trivialfis:
close #4494
close #4361

hcho3:
@trivialfis Thanks for fixing the pickling issue

pseudotensor:
@hcho3 Yes, this works great.  Please merge! :)

hcho3:
@pseudotensor Thanks for trying it out. I will review this PR and approve it today.

src/learner.cc
```diff
@@ -713,6 +689,7 @@ class LearnerImpl : public Learner {
       CHECK_LE(num_col, static_cast<uint64_t>(std::numeric_limits<unsigned>::max()))
         << "Unfortunately, XGBoost does not support data matrices with "
         << std::numeric_limits<unsigned>::max() << " features or greater";
+      // FIXME(trivialfis): Do we need this max since num_feature is just unsigned 0?
```
hcho3: Here, we are computing the maximum number of features over multiple cached DMatrix objects.

src/learner.cc
```diff
@@ -725,13 +702,16 @@ class LearnerImpl : public Learner {
...
     cfg_["num_feature"] = common::ToString(mparam_.num_feature);
     CHECK(obj_ == nullptr && gbm_ == nullptr);
-    obj_.reset(ObjFunction::Create(name_obj_));
+    obj_.reset(ObjFunction::Create(&tparam_, name_obj_));
     obj_->Configure(cfg_.begin(), cfg_.end());
     // reset the base score
     mparam_.base_score = obj_->ProbToMargin(mparam_.base_score);
-    gbm_.reset(GradientBooster::Create(name_gbm_, cache_, mparam_.base_score));
+    gbm_.reset(GradientBooster::Create(name_gbm_,
+                                       static_cast<LearnerTrainParam* const>(&tparam_),
```
hcho3: I don't think we need `static_cast` to cast non-const pointer to const, since the conversion should occur automatically.

src/objective/multiclass_obj.cu
```diff
@@ -63,11 +55,14 @@ class SoftmaxMultiClassObj : public ObjFunction {
     const int nclass = param_.num_class;
     const auto ndata = static_cast<int64_t>(preds.Size() / nclass);
 
-    out_gpair->Shard(GPUDistribution::Granular(devices_, nclass));
-    info.labels_.Shard(GPUDistribution::Block(devices_));
-    info.weights_.Shard(GPUDistribution::Block(devices_));
-    preds.Shard(GPUDistribution::Granular(devices_, nclass));
-    label_correct_.Shard(GPUDistribution::Block(devices_));
+    auto devices = GPUSet::All(tparam_->gpu_id, tparam_->n_gpus, preds.Size());
```
hcho3: Why are we constructing new `GPUSet` here? Is this to allow the user to change number of GPUs in the middle of an active Python session?

src/objective/hinge.cu
```diff
@@ -75,31 +57,26 @@ class HingeObj : public ObjFunction {
           }
           _out_gpair[_idx] = GradientPair(g, h);
         },
-        common::Range{0, static_cast<int64_t>(ndata)}, devices_).Eval(
-            &label_correct_, out_gpair, &preds, &info.labels_, &info.weights_);
+        common::Range{0, static_cast<int64_t>(ndata)},
+        GPUSet::All(tparam_->gpu_id, tparam_->n_gpus, ndata)).Eval(
+            out_gpair, &preds, &info.labels_, &info.weights_);
```
hcho3: Note to myself: understand how `GPUSet::Eval()` works, so that we can implement ranking metrics (AUC/AUCPR/NDCG) on GPU in the future

tests/cpp/common/test_common.cu
```diff
@@ -1,5 +1,9 @@
-#include "../../../src/common/common.h"
 #include <gtest/gtest.h>
+#include "xgboost/logging.h"
```
hcho3:
```suggestion
#include <xgboost/logging.h>
```

tests/cpp/common/test_enum_class_param.cc
```diff
@@ -1,6 +1,6 @@
-#include "../../../src/common/enum_class_param.h"
 #include <dmlc/parameter.h>
 #include <gtest/gtest.h>
+#include "xgboost/enum_class_param.h"
```
hcho3:
```suggestion
#include <xgboost/enum_class_param.h>
```

tests/cpp/helpers.h
hcho3: @trivialfis Second this comment.

tests/cpp/metric/test_multiclass_metric.cc
```diff
@@ -4,17 +4,12 @@
...
-#if defined(__CUDACC__)
-#define N_GPU() Arg{"n_gpus", "1"}
-#else
-#define N_GPU() Arg{"n_gpus", "0"}
-#endif
-
-inline void TestMultiClassError(std::vector<Arg> args) {
-  xgboost::Metric * metric = xgboost::Metric::Create("merror");
-  metric->Configure(args);
+inline void TestMultiClassError(xgboost::GPUSet devices) {
```
hcho3:
```suggestion
inline void TestMultiClassError(xgboost::GPUSet const& devices) {
```
to be consistent with uses of `GPUSet` in the `Learner` class.

tests/cpp/metric/test_multiclass_metric.cc
```diff
@@ -28,12 +23,16 @@ inline void TestMultiClassError(std::vector<Arg> args) {
...
 TEST(Metric, DeclareUnifiedTest(MultiClassError)) {
-  TestMultiClassError({N_GPU()});
+  auto devices = xgboost::GPUSet::Range(0, NGPUS());
+  TestMultiClassError(devices);
 }
 
-inline void TestMultiClassLogLoss(std::vector<Arg> args) {
-  xgboost::Metric * metric = xgboost::Metric::Create("mlogloss");
-  metric->Configure(args);
+inline void TestMultiClassLogLoss(xgboost::GPUSet devices) {
```
hcho3:
```suggestion
inline void TestMultiClassLogLoss(xgboost::GPUSet const& devices) {
```

tests/cpp/objective/test_regression_obj.cc
```diff
@@ -195,9 +215,65 @@ TEST(Objective, DeclareUnifiedTest(TweedieRegressionGPair)) {
...
+  {
+    // CPU
+    lparam.n_gpus = 0;
+    obj->GetGradient(preds, info, 0, &cpu_out_preds);
+  }
+  {
+    // CUDA
+    lparam.n_gpus = 1;
+    obj->GetGradient(preds, info, 0, &cuda_out_preds);
+  }
```
hcho3: Nice demo of how `n_gpus` can be changed on the fly :)

tests/cpp/test_learner.cc
```diff
@@ -115,4 +116,118 @@ TEST(Learner, SLOW_CheckMultiBatch) {
...
+
+  std::vector<bst_float> labels(kRows);
+  for (size_t i = 0; i < labels.size(); ++i) {
+    labels[i] = i;
+  }
+  p_dmat->Info().labels_.HostVector() = labels;
+  std::vector<std::shared_ptr<DMatrix>> mat {p_dmat};
+
+  std::unique_ptr<Learner> learner {Learner::Create(mat)};
+  learner->Configure({Arg{"predictor", "gpu_predictor"},
```
hcho3: I'm assuming `tree_method` is set to `auto` (CPU algorithm) here? Let's explicitly write it out here

tests/cpp/test_learner.cc
```diff
@@ -115,4 +116,118 @@ TEST(Learner, SLOW_CheckMultiBatch) {
...
+  {
+    // Create a scope to close the stream before next read.
+    std::unique_ptr<dmlc::Stream> fo(dmlc::Stream::Create(fname.c_str(), "w"));
+    learner->Save(fo.get());
+  }
+
+  std::unique_ptr<dmlc::Stream> fi(dmlc::Stream::Create(fname.c_str(), "r"));
+  learner->Load(fi.get());
+  ASSERT_EQ(learner->GetLearnerTrainParameter().gpu_id, 0);
+  ASSERT_EQ(learner->GetLearnerTrainParameter().n_gpus, 0);
```
hcho3: Why do we get 0 for `n_gpus` after loading the model? Is this because `tree_method` is `auto`?

So is it correct to say that you'll get CPU predictor when loading a learner with CPU algorithm, unless you explicitly override `predictor` parameter?

tests/cpp/test_learner.cc
```diff
@@ -115,4 +116,118 @@ TEST(Learner, SLOW_CheckMultiBatch) {
...
+    ASSERT_EQ(learner->GetLearnerTrainParameter().n_gpus, 1);
+  }
+  {
+    // with CPU algorithm
+    std::unique_ptr<Learner> learner {Learner::Create(mat)};
+    learner->Configure({Arg{"tree_method", "hist"}});
+    learner->InitModel();
+    learner->UpdateOneIter(0, p_dmat.get());
+    ASSERT_EQ(learner->GetLearnerTrainParameter().gpu_id, 0);
+    ASSERT_EQ(learner->GetLearnerTrainParameter().n_gpus, 0);
```
hcho3: So it's expected that `n_gpus` is automatically set to 0 for CPU algorithms?

tests/python-gpu/load_pickle.py
```diff
@@ -0,0 +1,20 @@
+'''Loading a pickled model generated by test_pickling.py'''
+import pickle
+import unittest
+import os
+import xgboost as xgb
+import sys
+
+sys.path.append("tests/python")
```
hcho3: This is not needed, because tests are run after XGBoost is installed into the local Python environment.

tests/python-gpu/test_pickling.py
```diff
@@ -0,0 +1,53 @@
...
+                 'verbosity': 1}
+        bst = xgb.train(param, train_x)
+
+        with open(model_path, 'wb') as fd:
+            pickle.dump(bst, fd)
+        args = ["pytest",
+                "--verbose",
+                "-s",
+                "--fulltrace",
+                "./tests/python-gpu/load_pickle.py"]
```
hcho3: Nice job writing the pickling test.

src/objective/hinge.cu
trivialfis: @hcho3 I have been reading up about that.  Thanks for your pointer to that paper.

tests/cpp/test_learner.cc
trivialfis: @hcho3 Yes.  That's the plan, choosing GPU parameters based on algorithms/predictor.

tests/cpp/test_learner.cc
trivialfis: @hcho3 Not that complicated.  Configuration is done after data is known, in `ConfugurationWithKnownData`.  Until the model see actual data, the default is CPU.  (Remember that this PR fixes XGBoost choosing GPU too aggressively).

src/objective/multiclass_obj.cu
trivialfis: @hcho3 No, user cannot change gpu_id or n_gpus, it's a  known limitation in XGBoost to be solved. But we can switch between GPU and CPU. :)

src/learner.cc
```diff
@@ -562,10 +528,16 @@ class LearnerImpl : public Learner {
...
+    return tparam_;
+  }
+
   std::pair<std::string, bst_float> Evaluate(DMatrix* data,
                                              std::string metric) {
     if (metric == "auto") metric = obj_->DefaultEvalMetric();
-    std::unique_ptr<Metric> ev(Metric::Create(metric.c_str()));
+    std::unique_ptr<Metric> ev(Metric::Create(
+        metric.c_str(), static_cast<LearnerTrainParam* const>(&tparam_)));
+    this->ConfigurationWithKnownData(data);
```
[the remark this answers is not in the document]
trivialfis: @hcho3 Nah, that's the tests for.  It took me some time to correct all tests.  But it's more than worthy.

tests/python-gpu/load_pickle.py
trivialfis: @hcho3 Those tests aren't installed.  I don't think I can import `from test_pickling import build_dataset, model_path` without proper sys path.

hcho3: @trivialfis Ah never mind then.

pseudotensor:
` ModuleNotFoundError: No module named 'xgboost.training'`

hcho3:
@pseudotensor Can we have a repro script?

hcho3:
@trivialfis Currently, Python tests on Win64 are not idempotent, as they attempt to install and remove XGBoost package into the system Python environment. So running more than one jobs in the same machine will cause problems (one job installs XGBoost and another removes it). For now, I limited the number of jobs to 1 for Windows workers, but I'd like to make Win64 tests idempotent in a follow-up PR.

pseudotensor:
> @pseudotensor Can we have a repro script?

That's just from the CI, seems fixed.

trivialfis:
@pseudotensor sounds like the reason described by @hcho3 .

hcho3:
Thanks for clarification. I will follow up with a PR to stabilize Windows CI

Issue #4227, opened 2019-03-07
[jvm-packages] missing parameter problems in xgboost-spark


adobay:
features that are not presented in the sparse feature matrix are treated as ‘missing’ on xgboost-spark, while spark ml vectorAssembler determines dense or sparse on storage cost. so for rows of dense,its 'missing' is NaN, for rows of sparse, its NaN and 0.

hcho3:
Can you post a small example where this become a problem?

adobay:
import org.apache.spark.ml.feature.VectorAssembler
val vectorAssembler = new VectorAssembler().
  setInputCols(Array("sepal length", "sepal width", "petal length", "petal width")).
  setOutputCol("features")
val xgbInput = vectorAssembler.transform(labelTransformed).select("features", "classIndex")

this is the example on https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html.
after transform, in xgbInput, some rows will be denseVector, some rows will be sparseVector.
the same data in the two different formats will produce different result.

hcho3:
@adobay Try setting `missing` parameter to 0.

@CodingCat This thread reminds me of #3634. Can we add a note about missing value handling to the XGBoost4J-Spark tutorial?

CodingCat:
it might be a problem in terms of usability

but how the same data is interpreted differently by VectorAssembler?

adobay:
this is the compressed code in spark ml:
def compressed: Vector = {
    val nnz = numNonzeros
    // A dense vector needs 8 * size + 8 bytes, while a sparse vector needs 12 * nnz + 20 bytes.
    if (1.5 * (nnz + 1.0) < size) {
      toSparseWithSize(nnz)
    } else {
      toDense
    }
  }

my suggestion is that XGBoost4J-Spark read sparseVector just like denseVector, read in 0, not as missing.

adobay:
@CodingCat @hcho3

CodingCat:
With the above code, how spark can interpret the same data into, sometimes dense and sometimes sparse?

adobay:
//for each row, vectorAssembler call the assembleFunc
val assembleFunc = udf { r: Row =>
      VectorAssembler.assemble(lengths, keepInvalid)(r.toSeq: _*)
    }.asNondeterministic()

//the last row creates Sparse or Dense vector depending how many non-zero values 
[...]

Issue #3742, opened 2018-09-29
why call xgboost4j api and call xgboost4j-spark api get different result


dwscu:
hello, recently I use spark-shell to run xgboost code,.I use xgboost4j-spark to run a model and predict,get the correct results.But later,I use this model to call xgboost4j predict method,I get wrong results
Why does this model get two different results?Do I use xgboost wrong?

here is my code.

xxx@xxx:~/danw$ /data/spark2/bin/spark-shell --jars xgboost4j-spark-0.72.jar,xgboost4j-0.72.jar 
2018-09-29 15:02:16 WARN  NativeCodeLoader:62 - Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Setting default log level to "WARN".
To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).
Spark context Web UI available at http://hz-ecom-wordspark-01:4040
Spark context available as 'sc' (master = local[*], app id = local-1538204542657).
Spark session available as 'spark'.
Welcome to
      ____              __
     / __/__  ___ _____/ /__
    _\ \/ _ \/ _ `/ __/  '_/
   /___/ .__/\_,_/_/ /_/\_\   version 2.3.1
      /_/
         
Using Scala version 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_45)
Type in expressions to have them evaluated.
Type :help for more information.

scala> import org.apache.spark.ml.feature.LabeledPoint
import org.apache.spark.ml.feature.LabeledPoint

scala> import org.apache.spark.ml.linalg.Vectors
import org.apache.spark.ml.linalg.Vectors

scala> import org.apache.spark.mllib.util.MLUtils
import org.apache.spark.mllib.util.MLUtils

scala> import ml.dmlc.xgboost4j.scala.{XGBoost, DMatrix}
import ml.dmlc.xgboost4j.scala.{XGBoost, DMatrix}

scala> import ml.dmlc.xgboost4j.scala.spark.XGBoost
import ml.dmlc.xgboost4j.scala.spark.XGBoost

scala> val train = MLUtils.loadLibSVMFile(sc, "/home/bbtree_RD/danw/agaricus.txt.train").map{ x => LabeledPoint(x.label, Vectors.dense(x.features.toArray))}
train: org.apache.spark.rdd.RDD[org.apache.spark.ml.feature.LabeledPoint] = MapPartitionsRDD[7] at map at <console>:30

[...]

Issue #3634, opened 2018-08-26
[jvm-packages] XGBoost spark predictions not consistent between SparseVector and DenseVector


pekaalto:
Hi,
I have noticed that if I change the underlying feature-type from SparseVector to DenseVector the predictions change wildly. 

I suspect that the DenseVector is not working correctly and the underlying issue is that  `DMatrix` is not working properly when created with `LabeledPoint` with indices = `null`.   Those are created here:

https://github.com/dmlc/xgboost/blob/0988fb191f6aacde50020b2d5e94c92df2ac4e58/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala#L65

Below is a script to reproduce the issue and investigation about the `LabeledPoint`'s. 

```
import com.github.fommil.netlib.BLAS.{getInstance => blas}
import ml.dmlc.xgboost4j.LabeledPoint
import ml.dmlc.xgboost4j.scala.spark._
import ml.dmlc.xgboost4j.scala.{Booster, DMatrix, XGBoost}
import org.apache.spark.SparkContext
import org.apache.spark.ml.linalg.{DenseVector, SparseVector}
import org.apache.spark.sql.functions._
import org.apache.spark.sql.{Column, SQLContext, SQLImplicits, SparkSession}

import scala.reflect.runtime.universe._
import scala.util.Random

object XGBTests {

  val spark = {
    SparkSession.builder
      .master("local[2]")
      .config("spark.sql.shuffle.partitions", "2")
      .config("spark.default.parallelism", "2")
      .appName("pekkatest")
      .getOrCreate()
  }

  val sc: SparkContext = spark.sparkContext
  val sqlContext: SQLContext = spark.sqlContext

  import sqlContext.implicits._

  case class Obs(features: Array[Float], label: Float)

  def createArtificialData(numFeatures: Int, numObs: Int): Array[Obs] = {
    val weights = (1 to numFeatures).map { _ => Random.nextDouble() }.toArray

    (1 to numObs).map { id =>
      val x = (1 to numFeatures).map { _ =>
        if (Random.nextDouble() > 0.7) Random.nextDouble() + 0.1 else 0.0
      }.toArray
      val y = blas.ddot(numFeatures, x, 1, weights, 1) + 3 * Random.nextGaussian()
      Obs(x.map(_.toFloat), y.toFloat)
    }.toArray
  }

  val numFeatures = 20
  val dataSize = 1000

[...]

PR #4455, opened 2019-05-10
[jvm-packages]fix XGBoost-on-Spark SparseVector missing value problem

adobay:
treat SparseVector as DenseVector


adobay:
the problem
https://github.com/dmlc/xgboost/pull/4315
https://github.com/dmlc/xgboost/issues/3742
https://github.com/dmlc/xgboost/issues/3634
https://github.com/dmlc/xgboost/issues/4227

CodingCat:
treating everything as dense vector brings significant amount of waste on the memory space, and the problem has been fixed in https://github.com/dmlc/xgboost/pull/4349 , are you still observing something out of order?

hcho3:
@CodingCat I'll defer to you on whether this pull request should make it to 0.90 release. However, I'm concerned that this PR increases memory consumption significantly and caused some tests to fail.

shishaochen:
Memory issue is extemely important in distributed training on large datasets.
In my cases, thousands of slots are 
defined while only small set of them are used per training job. To not change feature index, sparse 
vectors are needed or a must.

CodingCat:
I will close this PR soon as the problem has been fixed by other PRs,

PR #4458, opened 2019-05-11
[CI] Properly cache Docker containers for R tests

hcho3:
There are two R test suites currently running, one for R 3.4.4 and another for R 3.5.3. They require two separate Docker containers, but currently only one of them is being cached; the other container is being built from scratch.

After this small fix, both R containers will be cached.


hcho3:
The logic for naming containers has become quite cryptic. I'll need to clean up `ci_build.sh` in a follow-up PR.

PR #4461, opened 2019-05-12
[CI] Add option to use Google Test lib that's bundled with dmlc-core

hcho3:
The dmlc-core submodule bundles google test lib when `GOOGLE_TEST` flag is set. This pull requests adds an option `USE_DMLC_GTEST` to have XGBoost use the bundled gtest lib.


hcho3:
The idea was interesting, but it doesn't work on on Windows. Somehow linking the gtest target against testxgboost binary results in re-definition of OMP functions.

PR #4473, opened 2019-05-15
Add native support for Dask

RAMitchell:
There has been some discussion about integrating dask-xgboost into the main xgboost repository (dask/dask-xgboost#39). This will make maintenance/packaging easier and ensure the documentation lives in the same place.

I have taken a different approach to the integration than the original dask-xgboost, which completely wraps everything into a new API. I have instead provided helper functions for launching dask workers. Workers launched in this way are automatically connected via rabit and may perform distributed training.

I believe this approach is very flexible as we do not need to modify the existing xgboost python API in any way, we just call these functions using the dask launcher. This approach was also extremely lightweight in terms of code.

See the demo for a worked example.

TODO:
- [x] Add python tests
- [x] Provide multi-GPU demo
- [x] Possibly support sklearn APIs

@mrocklin @mt-jones @TomAugspurger

### DIFF
--- a/demo/dask/README.md
+++ b/demo/dask/README.md
@@ -2,12 +2,19 @@
 
 [Dask](https://dask.org/) is a parallel computing library built on Python. Dask allows easy management of distributed workers and excels handling large distributed data science workflows.
 
-This demo shows how to train and make predictions for an xgboost model on a distributed dask environment. We make use of first-class support in xgboost for launching dask workers. Workers launched in this manner are automatically connected via xgboosts underlying communication framework, Rabit. The calls to `xgb.train()` and `xgb.predict()` occur in parallel on each worker and are synchronized.
+The simple demo shows how to train and make predictions for an xgboost model on a distributed dask environment. We make use of first-class support in xgboost for launching dask workers. Workers launched in this manner are automatically connected via xgboosts underlying communication framework, Rabit. The calls to `xgb.train()` and `xgb.predict()` occur in parallel on each worker and are synchronized.
+
+The GPU demo shows how to configure and use GPUs on the local machine for training on a large dataset.
 
 ## Requirements
 Dask is trivial to install using either pip or conda. [See here for official install documentation](https://docs.dask.org/en/latest/install.html).
 
-## Running the script
+The GPU demo requires [GPUtil](https://github.com/anderskm/gputil) for detecting system GPUs.
+
+Install via `pip install gputil` 
+
+## Running the scripts
 ```bash
-python dask_demo.py
+python dask_simple_demo.py
+python dask_gpu_demo.py
 ```
--- a/demo/dask/README.md
+++ b/demo/dask/README.md
@@ -5,7 +5,7 @@
 This demo shows how to train and make predictions for an xgboost model on a distributed dask environment. We make use of first-class support in xgboost for launching dask workers. Workers launched in this manner are automatically connected via xgboosts underlying communication framework, Rabit. The calls to `xgb.train()` and `xgb.predict()` occur in parallel on each worker and are synchronized.
 
 ## Requirements
-Dask is trivial to install from a conda environment. [See here for official install documentation](https://docs.dask.org/en/latest/install.html).
+Dask is trivial to install using either pip or conda. [See here for official install documentation](https://docs.dask.org/en/latest/install.html).
 
 ## Running the script
 ```bash
--- a/demo/dask/README.md
+++ b/demo/dask/README.md
@@ -1,0 +1,13 @@
+# Dask Integration
+
+[Dask](https://dask.org/) is a parallel computing library built on Python. Dask allows easy management of distributed workers and excels handling large distributed data science workflows.
+
+This demo shows how to train and make predictions for an xgboost model on a distributed dask environment. We make use of first-class support in xgboost for launching dask workers. Workers launched in this manner are automatically connected via xgboosts underlying communication framework, Rabit. The calls to `xgb.train()` and `xgb.predict()` occur in parallel on each worker and are synchronized.
+
+## Requirements
+Dask is trivial to install from a conda environment. [See here for official install documentation](https://docs.dask.org/en/latest/install.html).
+
+## Running the script
+```bash
+python dask_demo.py
+```
--- a/demo/dask/dask_demo.py
+++ b/demo/dask/dask_demo.py
@@ -8,10 +8,10 @@ import xgboost as xgb
 # Define the function to be executed on each worker
 def train(X, y):
     print("Start training with worker #{}".format(xgb.rabit.get_rank()))
-    # X,y are dask data frame objects distributed across the cluster.
+    # X,y are dask objects distributed across the cluster.
     # We must obtain the data local to this worker and convert it to DMatrix for training.
     # xgb.dask.create_worker_dmatrix follows the API exactly of the standard DMatrix constructor
-    # (xgb.DMatrix()), except that it 'unpacks' dask dataframe objects to obtain data local to
+    # (xgb.DMatrix()), except that it 'unpacks' dask distributed objects to obtain data local to
     # this worker
     dtrain = xgb.dask.create_worker_dmatrix(X, y)
 
@@ -30,15 +30,15 @@ def main():
     cluster = LocalCluster(n_workers=2, threads_per_worker=2)
     client = Client(cluster)
 
-    # Generate some small test data and convert to a distributed dask dataframe
+    # Generate some small test data as a dask array
     # These data frames are internally split into partitions of 20 rows each and then distributed
     #  among workers, so we will have 5 partitions distributed among 2 workers
-    # Note that the partition size MUST be consistent across different dask dataframes
+    # Note that the partition size MUST be consistent across different dask dataframes/arrays
     n = 10
     m = 100
     partition_size = 20
-    X = dd.from_array(da.random.random((m, n), partition_size))
-    y = dd.from_array(da.random.random(m, partition_size))
+    X = da.random.random((m, n), partition_size)
+    y = da.random.random(m, partition_size)
 
     # xgb.dask.run launches an arbitrary function and its arguments on the cluster
     # Here train(X, y) will be called on each worker
--- a/demo/dask/dask_demo.py
+++ b/demo/dask/dask_demo.py
@@ -1,0 +1,54 @@
+from distributed import Client, LocalCluster
+import dask.dataframe as dd
+import numpy as np
+import xgboost as xgb
+
+
+# Define the function to be executed on each worker
+def train(X, y):
+    print("Start training with worker #{}".format(xgb.rabit.get_rank()))
+    # X,y are dask data frame objects distributed across the cluster.
+    # We must obtain the data local to this worker and convert it to DMatrix for training.
+    # xgb.dask.create_worker_dmatrix follows the API exactly of the standard DMatrix constructor
+    # (xgb.DMatrix()), except that it 'unpacks' dask dataframe objects to obtain data local to
+    # this worker
+    dtrain = xgb.dask.create_worker_dmatrix(X, y)
+
+    # Train on the data. Each worker will communicate and synchronise during training. The output
+    #  model is expected to be identical on each worker.
+    bst = xgb.train({}, dtrain)
+    # Make predictions on local data
+    pred = bst.predict(dtrain)
+    print("Finished training with worker #{}".format(xgb.rabit.get_rank()))
+    # Get text representation of the model
+    return bst.get_dump()
+
+
+def main():
+    # Launch a very simple local cluster using two distributed workers with two CPU threads each
+    cluster = LocalCluster(n_workers=2, threads_per_worker=2)
+    client = Client(cluster)
+
+    # Generate some small test data and convert to a distributed dask dataframe
+    # These data frames are internally split into partitions of 20 rows each and then distributed
+    #  among workers, so we will have 5 partitions distributed among 2 workers
+    # Note that the partition size MUST be consistent across different dask dataframes
+    n = 10
+    m = 100
+    partition_size = 20
+    X = dd.from_array(np.random.random((m, n)), partition_size)
+    y = dd.from_array(np.random.random(m), partition_size)
+
+    # xgb.dask.run launches an arbitrary function and its arguments on the cluster
+    # Here train(X, y) will be called on each worker
+    # This function blocks until all work is complete
+    models = xgb.dask.run(client, train, X, y)
+
+    # models contains a dictionary mapping workers to results
+    # We expect that the models are the same over all workers
+    first_model = next(iter(models.values()))
+    assert all(model == first_model for worker, model in models.items())
+
+
+if __name__ == '__main__':
+    main()
--- a/demo/dask/dask_demo.py
+++ b/demo/dask/dask_demo.py
@@ -1,5 +1,6 @@
 from distributed import Client, LocalCluster
 import dask.dataframe as dd
+import dask.array as da
 import numpy as np
 import xgboost as xgb
 
@@ -36,8 +37,8 @@ def main():
     n = 10
     m = 100
     partition_size = 20
-    X = dd.from_array(np.random.random((m, n)), partition_size)
-    y = dd.from_array(np.random.random(m), partition_size)
+    X = dd.from_array(da.random.random((m, n), partition_size))
+    y = dd.from_array(da.random.random(m, partition_size))
 
     # xgb.dask.run launches an arbitrary function and its arguments on the cluster
     # Here train(X, y) will be called on each worker
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -36,10 +36,19 @@ def _start_tracker(n_workers):
     return env
 
 
-def _merge_partitions(data, begin_partition, end_partition, total_partitions):
-    if data.npartitions != total_partitions:
-        raise ValueError("Dask data must have the same partitions")
-
+def get_local_data(data):
+    """
+    Unpacks a distributed data object to get the rows local to this worker
+    :param data: A distributed dask data object
+    :return: Local data partition e.g. numpy or pandas
+    """
+    if isinstance(data, DaskArray):
+        total_partitions = len(data.chunks[0])
+    else:
+        total_partitions = data.npartitions
+    partition_size = int(math.ceil(total_partitions / rabit.get_world_size()))
+    begin_partition = partition_size * rabit.get_rank()
+    end_partition = min(begin_partition + partition_size, total_partitions)
     if isinstance(data, DaskArray):
         return data.blocks[begin_partition:end_partition].compute()
 
@@ -57,25 +66,19 @@ def create_worker_dmatrix(*args, **kwargs):
     :param args: DMatrix constructor args.
     :return: DMatrix object containing data local to current dask worker
     """
-    total_partitions = args[0].npartitions
-    partition_size = int(math.ceil(total_partitions / rabit.get_world_size()))
-    begin_partition = partition_size * rabit.get_rank()
-    end_partition = min(begin_partition + partition_size, total_partitions)
     dmatrix_args = []
     dmatrix_kwargs = {}
     # Convert positional args
     for arg in args:
         if isinstance(arg, (DaskDataFrame, DaskSeries, DaskArray)):
-            dmatrix_args.append(
-                _merge_partitions(arg, begin_partition, end_partition, total_partitions))
+            dmatrix_args.append(get_local_data(arg))
         else:
             dmatrix_args.append(arg)
 
     # Convert keyword args
     for k, v in kwargs.items():
         if isinstance(v, (DaskDataFrame, DaskSeries, DaskArray)):
-            dmatrix_kwargs[k] = _merge_partitions(v, begin_partition, end_partition,
-                                                  total_partitions)
+            dmatrix_kwargs[k] = get_local_data(v)
         else:
             dmatrix_kwargs[k] = v
 
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -4,12 +4,10 @@ import os
 import sys
 import math
 from threading import Thread
-import distributed
-import dask
 import numpy as np
-import pandas as pd
 from . import rabit
 from .core import DMatrix
+from .compat import (DataFrame, pandas, dask, distributed)
 
 # Try to find the dmlc tracker script
 
@@ -42,8 +40,8 @@ def __start_tracker():
 def __concat(data):
     if isinstance(data[0], np.ndarray):
         return np.concatenate(data, axis=0)
-    if isinstance(data[0], (pd.DataFrame, pd.Series)):
-        return pd.concat(data, axis=0)
+    if isinstance(data[0], (DataFrame, pandas.Series)):
+        return pandas.concat(data, axis=0)
 
     raise TypeError("Data must be either numpy arrays or pandas dataframes"
                     ". Got %s" % type(data[0]))
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -4,10 +4,9 @@ import os
 import sys
 import math
 from threading import Thread
-import numpy as np
 from . import rabit
 from .core import DMatrix
-from .compat import (DaskDataFrame, DaskSeries, DaskArray, distributed_get_client,
+from .compat import (DaskDataFrame, DaskSeries, DaskArray,
                      distributed_get_worker)
 
 # Try to find the dmlc tracker script
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -7,7 +7,8 @@ from threading import Thread
 import numpy as np
 from . import rabit
 from .core import DMatrix
-from .compat import (DataFrame, pandas, dask, distributed)
+from .compat import (DaskDataFrame, DaskSeries, distributed_get_client,
+                     distributed_get_worker)
 
 # Try to find the dmlc tracker script
 
@@ -18,9 +19,9 @@ sys.path.append(ALTERNATE_TRACKER_PATH)
 from tracker import RabitTracker  # noqa
 
 
-def __start_tracker():
+def _start_tracker():
     """ Start Rabit tracker """
-    client = distributed.get_client()
+    client = distributed_get_client()
     host = client.scheduler.address
     if '://' in host:
         host = host.rsplit('://', 1)[1]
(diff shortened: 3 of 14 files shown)


python-package/xgboost/dask.py
```diff
@@ -42,14 +40,26 @@ def __start_tracker():
...
                     ". Got %s" % type(data[0]))
 
 
-def create_worker_dmatrix(*args):
+def __merge_partitions(dd, begin_partition, end_partition, total_partitions):
+    if dd.npartitions != total_partitions:
+        raise ValueError("Dask data must have the same partitions")
+    # Get local partitions
+    partitions = [dd.partitions[i].compute() for i in
+                  range(begin_partition, end_partition)]
```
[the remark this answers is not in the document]
mrocklin: This second approach would also remove the need to think about concatenation, which would remove a coordiantion point for libraries like cudf, which would now need to contend with how xgboost does things.

python-package/xgboost/dask.py
```diff
@@ -65,18 +75,24 @@ def create_worker_dmatrix(*args):
     begin_partition = partition_size * rabit.get_rank()
     end_partition = min(begin_partition + partition_size, total_partitions)
     dmatrix_args = []
+    dmatrix_kwargs = {}
+    # Convert positional args
     for arg in args:
         if isinstance(arg, (dask.dataframe.core.DataFrame, dask.dataframe.core.Series)):
```
mrocklin: You can safely drop the `.core` from this type.  That will make things slightly less brittle if we ever choose to rearrange our internal directory structure.

tests/ci_build/Dockerfile.cpu
```diff
@@ -23,7 +23,8 @@ ENV GOSU_VERSION 1.10
 RUN \
     pip install pyyaml cpplint pylint astroid sphinx numpy scipy pandas matplotlib sh recommonmark guzzle_sphinx_theme mock \
                 breathe matplotlib graphviz pytest scikit-learn wheel kubernetes urllib3 && \
-    pip install https://h2o-release.s3.amazonaws.com/datatable/stable/datatable-0.7.0/datatable-0.7.0-cp37-cp37m-linux_x86_64.whl
+    pip install https://h2o-release.s3.amazonaws.com/datatable/stable/datatable-0.7.0/datatable-0.7.0-cp37-cp37m-linux_x86_64.whl && \
+    conda install dask distributed -c conda-forge
```
mrocklin: You can drop `distributed`.  The `dask` package is a meta package with includes distributed.  If you want just the core dask library then you would `conda install dask-core`

tests/python/test_with_dask.py
```diff
@@ -0,0 +1,73 @@
...
+    dmat = xgb.DMatrix([[0]], label=[xgb.rabit.get_rank()])
+    bst = xgb.train({"eta": 1.0}, dmat, 1)
+    return bst.predict(dmat)
+
+
+def test_train():
+    # Train two workers, the first has label 0, the second has label 1
+    # If they build the model together the output should be 0.5
+    cluster = LocalCluster(n_workers=2, threads_per_worker=2)
+    client = Client(cluster)
```
mrocklin: You may want to take a look at our testing infrastructure: https://distributed.dask.org/en/latest/develop.html#writing-tests

These objects create and destroy processes and event loops.  If you do this a few times in CI systems then things can get a little messy.  Those testing harnesses should keep things relatively clean.  In particular you might want the `client` pytest fixture.


```python
from distributed.utils_test import client

def test_train(client):
    preds = xgb.dask.run(client, run_train)
    assert all(v[0] == 0.5 for v in preds.values())
````

File: tests/python/test_with_dask.py [hunk 5]
```diff
@@ -0,0 +1,73 @@
...
+
+
+def test_dask_dataframe():
+    cluster = LocalCluster(n_workers=2, threads_per_worker=2)
+    client = Client(cluster)
+    n = 10
+    m = 100
+    partition_size = 25
+    X = dd.from_array(np.random.random((m, n)), partition_size)
+    y = dd.from_array(np.random.random(m), partition_size)
```
mrocklin: You could also do `da.random.random(m, partition_size)` (though this isn't important).

demo/dask/README.md
```diff
@@ -0,0 +1,13 @@
+# Dask Integration
+
+[Dask](https://dask.org/) is a parallel computing library built on Python. Dask allows easy management of distributed workers and excels handling large distributed data science workflows.
+
+This demo shows how to train and make predictions for an xgboost model on a distributed dask environment. We make use of first-class support in xgboost for launching dask workers. Workers launched in this manner are automatically connected via xgboosts underlying communication framework, Rabit. The calls to `xgb.train()` and `xgb.predict()` occur in parallel on each worker and are synchronized.
+
+## Requirements
+Dask is trivial to install from a conda environment. [See here for official install documentation](https://docs.dask.org/en/latest/install.html).
```
mrocklin: it's also trivial to install with pip by the way

```python
pip install dask distributed
```

demo/dask/dask_demo.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # Generate some small test data and convert to a distributed dask dataframe
+    # These data frames are internally split into partitions of 20 rows each and then distributed
+    #  among workers, so we will have 5 partitions distributed among 2 workers
+    # Note that the partition size MUST be consistent across different dask dataframes
+    n = 10
+    m = 100
+    partition_size = 20
+    X = dd.from_array(np.random.random((m, n)), partition_size)
+    y = dd.from_array(np.random.random(m), partition_size)
```
mrocklin: We should probably use either `da.random.random` or `dask.datasets.timeseries` here.  Otherwise users will scale this up and ask why it doesn't work.  Allocating a large numpy array in local memory and then passing that to Dask is good for a quick demo, but a bad practice to give to impressionable users.

CodingCat:
No objection on this PR for now (considering where xgboost is )

But one thing to think about is what is the criterion to determine if we have native support to some data processing framework

Do we want to support in https://github.com/modin-project/modin in future? Why we choose dask instead of modin?

(I am not a fan of modin, never used...)

Just something we have to be clear as the maintainer of the project @hcho3, @trivialfis

trivialfis:
@CodingCat 
> Do we want to support in https://github.com/modin-project/modin in future? 

Not quite understand what's the advantage of using pandas: https://h2oai.github.io/db-benchmark/
There are also https://github.com/yandex/ClickHouse , https://github.com/rapidsai/cudf .  Just dropping a few links...

python-package/xgboost/dask.py
```diff
@@ -42,14 +40,26 @@ def __start_tracker():
 def __concat(data):
     if isinstance(data[0], np.ndarray):
         return np.concatenate(data, axis=0)
-    if isinstance(data[0], (pd.DataFrame, pd.Series)):
-        return pd.concat(data, axis=0)
+    if isinstance(data[0], (DataFrame, pandas.Series)):
+        return pandas.concat(data, axis=0)
```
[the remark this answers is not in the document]
mrocklin: Expanding on my note below.  I think that ideally xgboost wouldn't need to have a concat function at all.  I would hope that all of this code would go away.

python-package/xgboost/dask.py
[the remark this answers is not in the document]
mrocklin: I mean that in the second code snippet in my comment above xgboost is handed a pre-concatenated dataframe.  So while Dask still has to think about this, XGBoost doesn't have to.  It's one less thing for cudf and xgboost devs to have to talk about.  (or any other library author who wants to work with dask and xgboost together)

CodingCat:
> @CodingCat
> 
> > Do we want to support in https://github.com/modin-project/modin in future?
> 
> Not quite understand what's the advantage of using pandas: https://h2oai.github.io/db-benchmark/
> There are also https://github.com/yandex/ClickHouse , https://github.com/rapidsai/cudf . Just dropping a few links...

one of the biggest selling point of dask is it can scale pandas-styled analytics to distributed nodes, integrate Dask with XGB is also beneficial because we can have a seamless data preparation and model training now

my point is just based on the links you posted, what's the criterion to judge we should make native support of Dask, instead of any other one of them?

otherwise, XGBoost is becoming a downstream library of many other projects (big burden for syncing versions, and most of projects without a company support are just going to die at that point, do we want to clean code base by deleting code?)

mt-jones:
>my point is just based on the links you posted, what's the criterion to judge we should make native support of Dask, instead of any other one of them? or in future, ClickHouse's person come and say they have a ClickHouse-ML library to work with XGBoost, do we want to integrate to XGB code base?

I think there are perhaps two feasible options:
1. Refuse native integration for all such frameworks, and require that the code for interoperability live in that native framework. This would require XGBoost to adopt a specification for data ingest/egress so that external frameworks could properly dispatch their data into XGBoost.
2. Allow for native integration of external frameworks into XGBoost with a caveat: that the contributor associated with the interoperability layer for that framework maintain the that layer within XGBoost for all time; else, that code will face inevitable deprecation and removal. A criterion can be established for what "maintain" means in this case.

In either case, the onus for developing and maintaining the interoperability (contractual) interface should fall to the native framework contributor.

>I think we should be clear about it instead of integrating for showing friend-ness. Otherwise, XGBoost is becoming a downstream library of many other projects (big burden for syncing versions, and most of projects without a company support are just going to die, at that point, do we want to clean code base by deleting code?)

Regardless of what is done, I think code rot is inevitable. Deleting old code, and redesigning components regularly seems natural. I'm not sure if I've missed your point, though(?) If you're suggesting that XGBoost remove all interface-related code in favor of a single standard, I think that's a great idea.

Such an interface should be capable of dispatching types. For example, if you chose to integrate with Dask, you could design a conversion dispatcher _for XGBoost_, and people that wanted to design an interoperability layer with another layer could simply (with some care) register a source-to-target converter [much like concatenation is done](https://github.com/dask/dask/blob/0dbe40de09942c004d7b3bbe5a93b23897877cc4/docs/source/dataframe-extend.rst). Like (1), I think this requires XGBoost specify more formally an internal data structure so that the process of building these methods in other frameworks is simpler.

CodingCat:
> Regardless of what is done, I think code rot is inevitable. Deleting old code, and redesigning components regularly seems natural. I'm not sure if I've missed your point, though(?) 

let's imagine that we integrated with Hadoop's MapReduce years ago, at this time point, shall we still support that considering the current status of MapReduce project? most of users are moving to Spark, but some users are happily using XGBoost-MapReduce....or a living example is XGBoost-Flink, I am seriously considering if we want to bring it to XGBoost 1.0, AFAIK, Flink-Batch has a small but existing  group of users, and Flink's father company is bought by Alibaba now which makes many guesses to the future of project.....

support => continuous burden 

not support => some users left XGBoost community

tests/ci_build/Dockerfile.cpu
RAMitchell: This confuses me a little. I copied this line directly from your install documentation. Is this documentation out of date?

tests/python/test_with_dask.py
RAMitchell: When I do this I get `AttributeError: 'function' object has no attribute 'run_on_scheduler'

mt-jones:
I see. Is the proposition for 1.0 to simply drop all such code and avoid the maintenance burden? Or is this being weighed against lost communities?

CodingCat:
1.0 is more about a milestone of XGBoost's functionality and performance, but it's also a good chance to drop burdens if necessary...

mt-jones:
> 1.0 is more about a milestone of XGBoost's functionality and performance, but it's also a good chance to drop burdens if necessary...

Agree. So, correct me if I’m wrong, but your original question was more about how to define a burden. Yes?

RAMitchell:
@CodingCat very good points. I guess my justification for including dask is that it gives us a vehicle to launch distributed GPU algorithms, given that spark is not currently ready for this. As an active maintainer I am prepared to deal with this, especially given the relatively low overhead. Maybe we need a more well defined criteria for deprecating support for other frameworks.

tests/ci_build/Dockerfile.cpu
mrocklin: Where were you looking?

Official docs are here: https://docs.dask.org/en/latest/install.html

I wouldn't be surprised if other docs exist elsewhere though

CodingCat:
> > 1.0 is more about a milestone of XGBoost's functionality and performance, but it's also a good chance to drop burdens if necessary...
> 
> Agree. So, correct me if I’m wrong, but your original question was more about how to define a burden. Yes?

hmmmmm...........not exactly....burden comes from any integration, XGBoost-Spark also brings burden, we waited for more than 1 week for 2.4.3 release to mark XGBoost 0.9 is ready to go. But obviously, the original idea to bring XGBoost4J-Spark to XGBoost is a right one as it pushes XGBoost to production environment significantly.

I think my question is about the criterion for tradeoff between burden and benefits to XGBoost.

python-package/xgboost/dask.py
```diff
@@ -42,14 +40,26 @@ def __start_tracker():
...
-def create_worker_dmatrix(*args):
+def __merge_partitions(dd, begin_partition, end_partition, total_partitions):
+    if dd.npartitions != total_partitions:
+        raise ValueError("Dask data must have the same partitions")
+    # Get local partitions
+    partitions = [dd.partitions[i].compute() for i in
+                  range(begin_partition, end_partition)]
+    if not partitions:
+        raise ValueError("Worker " + str(
+            rabit.get_rank()) + " has no data. Try using smaller partitions")
```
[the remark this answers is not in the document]
RAMitchell: If one of the xgboost.train() calls has no data I believe it will just fail. I prefer to explicitly enforce all workers have data for now.

tests/python/test_with_dask.py
mrocklin: I recommend starting from here:

```python
from distributed.utils_test import client, loop, cluster_fixture
from dask.distributed import Client

def test_client_fixture(client):
    assert isinstance(client, Client)
    assert hasattr(client, 'run_on_scheduler')
```

```
py.test test_foo.py 
=============================================== test session starts ===============================================
platform linux -- Python 3.6.7, pytest-4.1.1, py-1.7.0, pluggy-0.8.0 -- /home/mrocklin/Software/anaconda/bin/python
cachedir: .pytest_cache
rootdir: /home/mrocklin/workspace/distributed, inifile: setup.cfg
plugins: xdist-1.23.2, timeout-1.2.1, forked-0.2, cov-2.5.1, asyncio-0.10.0
collected 1 item                                                                                                  

baz.py::test_client_fixture PASSED                                                                          [100%]

============================================ slowest 10 test durations ============================================
0.80s setup    baz.py::test_client_fixture
0.52s teardown baz.py::test_client_fixture

(0.00 durations hidden.  Use -vv to show these durations.)
============================================ 1 passed in 1.41 seconds =============================================
```

I should have mentioned that this assumes that you're using pytest, which I recognize now may not be a valid assumption.

CodingCat:
> @CodingCat very good points. I guess my justification for including dask is that it gives us a vehicle to launch distributed GPU algorithms, given that spark is not currently ready for this. As an active maintainer I am prepared to deal with this, especially given the relatively low overhead. Maybe we need a more well defined criteria for deprecating support for other frameworks.

agree, I think distributed GPU makes efforts on Dask more valuable instead of just bringing Dask here....

> Maybe we need a more well defined criteria for deprecating support for other frameworks.

also agree, I am drafting several documents for the community and I will ensure this is covered and bring it to discussion

mrocklin:
> I think my question is about the criterion for tradeoff between burden and benefits to XGBoost.

For what it's worth I totally understand the concerns around building connections to an external library inside of a core codebase.  In maintaining Dask we've been pretty conservative about what is allowed in core, what gets imported by default, and so on.  As a maintainer it's a hard decision to make.  

I don't maintain XGBoost, and I'm obviously biased towards Dask, so please don't trust anything I say here, but here are my thoughts.

-  Dask was intentionally designed to be infrastructure for other libraries and as such is fairly lightweight.  The core of the project is Pure Python and stdlib only (though to be clear, you're relying on more than core here).  
-  Other projects of similar size, like Scikit-Learn, maintain dask-specific connections in their codebases (well, in this case it's in joblib, which is a core component of Scikit-Learn)
-  The design that @RAMitchell has proposed intentionally has a very small surface area, and depends only on very mature APIs, so I would not expect the maintenance burden to be high, even if the two libraries undergo a lot of API churn (which seems unlikely at this stage of maturity)
-  In this case you do have @RAMitchell around to handle the maintenance

However 

-  If you can easily avoid it I wouldn't import dask by default as this PR currently does.

tests/ci_build/Dockerfile.cpu
RAMitchell: I think google led me to some different instructions :)

RAMitchell: http://distributed.dask.org/en/latest/install.html

mrocklin: Indeed, I just tried that and got there as a first hit.  We should probably start retiring a lot of content on that site.

python-package/xgboost/dask.py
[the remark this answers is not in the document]
mt-jones: I’m not sure I understand the scope of the suggestion and how `dd.partitions[begin_partition:end_partition].compute()` affects this. Are you saying this method calls a concatenation method in Dask under the hood?

tests/ci_build/Dockerfile.cpu
RAMitchell: I think this would help your usability a lot. When I first tried dask I had a few install frustrations just because of old and inconsistent documentation. The other significant usability issue that I came across was errors because I didn't start a client within a main function. This was not clear to me and I think most users experience this and have two Google for the answer.

mrocklin: - https://github.com/dask/dask/issues/4813
-  https://github.com/dask/distributed/issues/2708

python-package/xgboost/dask.py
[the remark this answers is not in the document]
mt-jones: Found my answer:
```
        This allows partitionwise slicing of a Dask Dataframe.  You can perform normal
        Numpy-style slicing but now rather than slice elements of the array you
        slice along partitions so, for example, ``df.partitions[:5]`` produces a new
        Dask Dataframe of the first five partitions.
```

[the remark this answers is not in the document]
mt-jones: Thank you for clarifying. I certainly missed your point about avoiding an explicit call to concat.

RAMitchell:
Another question here is should we be able to return the predictions as a distributed object? If we know the return type for the train function is always some local data collection (numpy/pandas) we can package it into a distributed dataframe. Not sure how this works if the user wants to return other types of objects.

mrocklin:
If you do something like the following:

```python 
result = df.map_partitions(local_predict_function)  # dataframe
result = x.map_blocks(local_predict_function) # array
```

Then Dask will figure out the correct parallel type regardless of what the user's predict function returns.

Would that solve things?

RAMitchell:
Tests are currently failing on CI but passing on my local machine. Maybe something to do with the permissions of the docker environment and the need for dask to read/write temporary files.

I have decided not to try to return dask objects from the run function at this time. Users may want to return multiple objects or objects of different types and it is not obvious to me how to cleanly handle this, so I will leave it to the user to concatenate their results back together.

This is ready for another review @mrocklin @mt-jones .

mrocklin:
```
_________________________ ERROR at setup of test_train _________________________
file /Users/travis/build/dmlc/xgboost/tests/python/test_with_dask.py, line 25
  def test_train(client):
E       fixture 'client' not found
>       available fixtures: cache, capfd, capfdbinary, caplog, capsys, capsysbinary, cov, doctest_namespace, monkeypatch, no_cover, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.
```

This seems to be an import error?  Perhaps this line isn't imported?

https://github.com/dmlc/xgboost/pull/4473/files#diff-bc381ee4ba446c889bda3fa1032f9a5dR7

mrocklin:
```
> [2019-05-20T02:33:56.685Z] 

[2019-05-20T02:33:56.685Z] >   raise Exception(response["text"])

[2019-05-20T02:33:56.685Z] E   Exception: I/O operation on closed file

[2019-05-20T02:33:56.685Z] 

[2019-05-20T02:33:56.685Z] /opt/python/lib/python3.7/site-packages/distributed/core.py:550: Exception
```

That error is a bit odd.  It's also coming through as text rather than as a proper traceback meaning that Python wasn't able to serialize the exception (perhaps because it had a file handle?)

Can you try a test that just starts a Dask client and then does a trivial amount of work with it so that we can isolate this between Dask-on-this-CI and Dask+XGBoost-on-this-CI?

Maybe something like this:

```python
def test_dask(client):
    assert client.submit(lambda x: x + 1, 10).result() == 11
```

RAMitchell:
```
distributed.protocol.pickle - INFO - Failed to deserialize b'\x80\x04\x95#\x00\x00\x00\x00\x00\x00\x00\x8c\x0cxgboost.dask\x94\x8c\x0e_start_tracker\x94\x93\x94.'
  File "/opt/python/lib/python3.7/site-packages/distributed/protocol/pickle.py", line 61, in loads
    return pickle.loads(x)
  File "/opt/python/lib/python3.7/site-packages/xgboost-0.83.dev0-py3.7.egg/xgboost/__init__.py", line 11, in <module>
    from .core import DMatrix, Booster
  File "/opt/python/lib/python3.7/site-packages/xgboost-0.83.dev0-py3.7.egg/xgboost/core.py", line 22, in <module>
    from .compat import (STRING_TYPES, PY3, DataFrame, MultiIndex, py_str,
  File "/opt/python/lib/python3.7/site-packages/xgboost-0.83.dev0-py3.7.egg/xgboost/compat.py", line 47, in <module>
    import datatable.Frame as DataTable
  File "/opt/python/lib/python3.7/site-packages/datatable/__init__.py", line 8, in <module>
    from .dt_append import rbind, cbind
  File "/opt/python/lib/python3.7/site-packages/datatable/dt_append.py", line 8, in <module>
    from datatable.utils.misc import plural_form as plural
  File "/opt/python/lib/python3.7/site-packages/datatable/utils/misc.py", line 8, in <module>
    from .typechecks import TImportError
  File "/opt/python/lib/python3.7/site-packages/datatable/utils/typechecks.py", line 14, in <module>
    from datatable.utils.terminal import term
  File "/opt/python/lib/python3.7/site-packages/datatable/utils/terminal.py", line 89, in <module>
    term = MyTerminal()
  File "/opt/python/lib/python3.7/site-packages/datatable/utils/terminal.py", line 20, in __init__
    super().__init__()
  File "/opt/python/lib/python3.7/site-packages/blessed/terminal.py", line 171, in __init__
    self._keyboard_fd = sys.__stdin__.fileno()
ValueError: I/O operation on closed file
distributed.core - ERROR - I/O operation on closed file
  File "/opt/python/lib/python3.7/site-packages/distributed/core.py", line 412, in handle_comm
    result = yield result
  File "/opt/python/lib/python3.7/site-packages/tornado/gen.py", line 729, in run
    value = future.result()
  File "/opt/python/lib/python3.7/site-packages/tornado/gen.py", line 209, in wrapper
    yielded = next(result)
  File "/opt/python/lib/python3.7/site-packages/distributed/worker.py", line 3245, in run
    function = pickle.loads(function)
  File "/opt/python/lib/python3.7/site-packages/distributed/protocol/pickle.py", line 61, in loads
    return pickle.loads(x)
  File "/opt/python/lib/python3.7/site-packages/xgboost-0.83.dev0-py3.7.egg/xgboost/__init__.py", line 11, in <module>
    from .core import DMatrix, Booster
  File "/opt/python/lib/python3.7/site-packages/xgboost-0.83.dev0-py3.7.egg/xgboost/core.py", line 22, in <module>
    from .compat import (STRING_TYPES, PY3, DataFrame, MultiIndex, py_str,
  File "/opt/python/lib/python3.7/site-packages/xgboost-0.83.dev0-py3.7.egg/xgboost/compat.py", line 47, in <module>
    import datatable.Frame as DataTable
  File "/opt/python/lib/python3.7/site-packages/datatable/__init__.py", line 8, in <module>
    from .dt_append import rbind, cbind
  File "/opt/python/lib/python3.7/site-packages/datatable/dt_append.py", line 8, in <module>
    from datatable.utils.misc import plural_form as plural
  File "/opt/python/lib/python3.7/site-packages/datatable/utils/misc.py", line 8, in <module>
    from .typechecks import TImportError
  File "/opt/python/lib/python3.7/site-packages/datatable/utils/typechecks.py", line 14, in <module>
    from datatable.utils.terminal import term
  File "/opt/python/lib/python3.7/site-packages/datatable/utils/terminal.py", line 89, in <module>
    term = MyTerminal()
  File "/opt/python/lib/python3.7/site-packages/datatable/utils/terminal.py", line 20, in __init__
    super().__init__()
  File "/opt/python/lib/python3.7/site-packages/blessed/terminal.py", line 171, in __init__
    self._keyboard_fd = sys.__stdin__.fileno()
ValueError: I/O operation on closed file
distributed.core - ERROR - I/O operation on closed file
```
This error message actually goes away if we don't import datatable. @st-pasha can you comment on this?

st-pasha:
It appears the error is coming from the following line in module `blessed` (dependency for `datatable`):
```
sys.__stdin__.fileno()
```
I presume that somehow on your environment the `__stdin__` stream is closed, which causes the exception. I've never seen this particular error before, but there's the first time for everything.

This is something that we could trivially fix in `datatable` (guard the use of blessed.Terminal with a try-catch block for example), but I'm not sure how urgent this problem is for you, and whether you'd be able to use the development version of datatable in your CI.
However, I believe the following workaround could do the trick: in the `compat.py` file before importing `datatable` add the following lines:
```
if sys.__stdin__.closed:
    sys.__stdin__ = None
```
we already test for the case when `sys.__stdin__` is completely missing and avoid using `blessed` in that case.

RAMitchell:
Thanks @st-pasha for the quick response. I'm not much of a Python expert but wouldn't this have side-effects?

st-pasha:
The standard stdin channel for python is `sys.stdin`, so "normal" programs are expected to use that. The `sys.__stdin__` is the "original" stdin, and according to docs should almost never be used (https://docs.python.org/3/library/sys.html#sys.__stdin__). The documentation even mentions that this variable could be `None`.

Of course, changing something at the `sys` level, even if this something is rarely used, is generally dangerous. However, in this case the `sys.__stdin__` is removed only in the case when the channel was already closed, so it was in a "unusable" state anyways. So, in this case I don't think this could do any harm (unless there is a python program that doesn't take into account possibility of `sys.__stdin__` being `None`, in which case it's better to fix that program anyways).

RAMitchell:
I am getting socket errors with rabit on Windows so will not support this platform at the moment.

CodingCat:
@cq ^^

RAMitchell:
@hcho3 So I have learned that setuptools packages xgboost in a completely different way on Windows - none of the source files get included. This means it does not contain the dmlc tracker scripts. The only solution I have that seems to make any sense is to add the tracker scripts as a package so they get included consistently in the right place.

hcho3:
@RAMitchell This is because symlinks are not supported in Windows. All source files are included in `python-package` directory as symlinks, and symlinks are silently ignored in Windows. One workaround is to have `setup.py` manually copy directories (use `shutil`).

PR #4474, opened 2019-05-16
Add support for cross-validation using query ID

### LINKED ISSUE #270: Group-aware CV Fold Creation for Rank Task
This was important for rank typed tasks. For rank type tasks, there is an group information to group a instances into a List. In cross validation, folds should be created such that each group was splitted either into train or test.

Currently This was not handled correctly, which means cv do not work properly for rank typed tasks.
Here are something that needs to do
- [ ] add a slice interface that is related to group
- [ ] in python and R, do group aware split when group information presents.

Since less user use cv for rank task so far. This is not an urgent issue, but I am opening this for now. To see if anyone want to take it over. This is likely need to have things to do with the C wrapper part as well as the language bindings.

bryan-woods:
Fixes #270.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -89,3 +89,5 @@ List of Contributors
 * [Sam Wilkinson](https://samwilkinson.io)
 * [Matthew Jones](https://github.com/mt-jones)
 * [Jiaxiang Li](https://github.com/JiaxiangBU)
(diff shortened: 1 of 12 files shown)


hcho3:
I will modify the example to use ZIP format, so that we don't need to use UNRAR utility

bryan-woods:
thank you for your help!

Issue #4480, opened 2019-05-20
Error in error message from `_maybe_pandas_data` : `TypeError: sequence item 0: expected str instance, int found`


srowen:
This is just a minor bug, I believe, in how `_maybe_pandas_data` tries to report an error. The error message itself generates an error.

My code ended up passing a DataFrame with bad type to xgboost, which triggers the error at https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/core.py#L239 This is all fine and expected.

However the `.join` in that line causes a different error to be thrown: 
```
  File "/databricks/python/lib/python3.6/site-packages/shap/explainers/tree.py", line 172, in shap_values
    X = xgboost.DMatrix(X)
  File "/databricks/python/lib/python3.6/site-packages/xgboost/core.py", line 384, in __init__
    feature_types)
  File "/databricks/python/lib/python3.6/site-packages/xgboost/core.py", line 241, in _maybe_pandas_data
    raise ValueError(msg + ', '.join(bad_fields))
TypeError: sequence item 0: expected str instance, int found
```

I think it's because in my case, the DataFrame column names are ints? they aren't otherwise given names. And Python doesn't let you join a list of ints.

I think the fix may just be ...

```
bad_fields = [str(data.columns[i]) for i, dtype in ...
```

If so happy to try opening a PR.

trivialfis:
@srowen Thanks for the issue.  Please go ahead and open a PR.

PR #4481, opened 2019-05-20
Ensure pandas DataFrame column names are treated as strings in type error message

srowen:
Closes https://github.com/dmlc/xgboost/issues/4480

If the DataFrame column names aren't strings, constructing the error message here will yield an error:
```
File "/databricks/python/lib/python3.6/site-packages/shap/explainers/tree.py", line 172, in shap_values
    X = xgboost.DMatrix(X)
  File "/databricks/python/lib/python3.6/site-packages/xgboost/core.py", line 384, in __init__
    feature_types)
  File "/databricks/python/lib/python3.6/site-packages/xgboost/core.py", line 241, in _maybe_pandas_data
    raise ValueError(msg + ', '.join(bad_fields))
TypeError: sequence item 0: expected str instance, int found
```

This simply ensures it's a string.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -231,7 +231,7 @@ def _maybe_pandas_data(data, feature_names, feature_types):
 
     data_dtypes = data.dtypes
     if not all(dtype.name in PANDAS_DTYPE_MAPPER for dtype in data_dtypes):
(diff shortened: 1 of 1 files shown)


trivialfis:
Thanks!

PR #4478, opened 2019-05-20
Simplify INI-style config reader using C++11 STL

fuhaoda:
Use C++11 STL to simplify the   "./src/common/config.h"  file.

PR #4486, opened 2019-05-22
- training with external memory part 1 of 2

sriramch:
- this pr focuses on computing the quantiles using multiple gpus on a
     dataset that uses the external cache capabilities
   - there will a follow-up pr soon after this that will support creation
     of histogram indices on large dataset as well
   - both of these changes are required to support training with external memory
   - the sparse pages in dmatrix are taken in batches and the the cut matrices
     are incrementally built
   - also snuck in some (perf) changes related to sketches aggregation amongst multiple
     features across multiple sparse page batches. instead of aggregating the summary
     inside each device and merged later, it is aggregated in-place when the device
     is working on different rows but the same feature


sriramch:
@canonizer @RAMitchell @rongou - please review

this is very similar to #4448 and i have split it up per the earlier review request

sriramch:
@hcho3 any reason why i do not see all the checks with this pr (typically there are ~ 10 of them)?

hcho3:
@sriramch Looks like GitHub webhook didn't get activated. Let me try to re-trigger

hcho3:
@sriramch See https://www.githubstatus.com/. Looks like there is an issue with notification delivery. We'll have to wait until it's fixed

sriramch:
> @sriramch See https://www.githubstatus.com/. Looks like there is an issue with notification delivery. We'll have to wait until it's fixed

thanks @hcho3 for looking and helping!

hcho3:
@sriramch For now, I've manually triggered all Jenkins tests (Linux/Win64).

sriramch:
> @sriramch For now, I've manually triggered all Jenkins tests (Linux/Win64).

@hcho3 thanks for your help - much appreciated!

hcho3:
@sriramch I'm having some issues with triggering Linux tests. Can we wait until the webhook issue is fixed?

sriramch:
> @sriramch I'm having some issues with triggering Linux tests. Can we wait until the webhook issue is fixed?

@hcho3 - no problem - we can wait!

src/common/hist_util.cu
```diff
@@ -82,6 +84,18 @@ __global__ void UnpackFeaturesK
...
+/*!
+ * \brief A container that holds the device sketches across all
+ *  sparse page batches which are distributed to different devices.
+ *  As sketches are aggregated by column, the mutex guards
+ *  multiple devices pushing sketch summary for the same column
+ *  across distinct rows.
+ */
+struct SketchContainer {
+  std::vector<HistCutMatrix::WXQSketch> sketches_;  // NOLINT
+  std::vector<std::unique_ptr<std::mutex>> col_locks_; // NOLINT
```
canonizer: Feel free to use `std::vector<std::mutex>`, unless there is a reason to use `std::unique_ptr`.

src/common/hist_util.cu
```diff
@@ -113,9 +127,13 @@ struct GPUSketcher {
...
    public:
     DeviceShard(int device, bst_uint row_begin, bst_uint row_end,
-                tree::TrainParam param) :
+                tree::TrainParam param, SketchContainer *sketch_container) :
       device_(device), row_begin_(row_begin), row_end_(row_end),
-      n_rows_(row_end - row_begin), param_(std::move(param)) {
+      n_rows_(row_end - row_begin), param_(std::move(param)), sketch_container_(sketch_container) {
+    }
+
+    inline size_t GetRowStride() {
```
canonizer: Any reason to have a method here? If the purpose is to make `row_stride_` accessible, consider making `DeviceShard` a struct.

src/common/hist_util.cu
```diff
@@ -136,20 +154,10 @@ struct GPUSketcher {
...
-      // initialize sketches
-      sketches_.resize(num_cols_);
-      summaries_.resize(num_cols_);
       constexpr int kFactor = 8;
       double eps = 1.0 / (kFactor * param_.max_bin);
       size_t dummy_nlevel;
-      WXQSketch::LimitSizeLevel(row_batch.Size(), eps, &dummy_nlevel, &n_cuts_);
-      // double ncuts to be the same as the number of values
-      // in the temporary buffers of the sketches
-      n_cuts_ *= 2;
```
canonizer: Is it possible to keep doubling `n_cuts_`?

src/common/hist_util.cu
```diff
@@ -341,63 +357,91 @@ struct GPUSketcher {
...
+                                                    decltype(counting), size_t>;
+      TransformT row_size_iter = TransformT(counting, get_size);
+      row_stride_ = thrust::reduce(row_size_iter, row_size_iter + n_rows_, 0,
+                                   thrust::maximum<size_t>());
     }
   };
 
-  void Sketch(const SparsePage& batch, const MetaInfo& info,
-              HistCutMatrix* hmat, int gpu_batch_nrows) {
+  size_t SketchBatch(const GPUDistribution &dist, const SparsePage &batch,
```
canonizer: The row stride can be definitely passed through a shard or class variable. No need to return a value here.

src/common/hist_util.cu
```diff
@@ -341,63 +357,91 @@ struct GPUSketcher {
...
+                     const MetaInfo &info, SketchContainer *sketch_container) {
```
canonizer: `sketch_container` should be a member of a class or shard.

src/common/hist_util.cu
```diff
@@ -341,63 +357,91 @@ struct GPUSketcher {
...
+  }
+
+  GPUSketcher(const tree::TrainParam &param, int gpu_nrows)
+    : param_(param), gpu_batch_nrows_(gpu_nrows) {
   }
 
-  GPUSketcher(tree::TrainParam param, size_t n_rows) : param_(std::move(param)) {
-    dist_ = GPUDistribution::Block(GPUSet::All(param_.gpu_id, param_.n_gpus, n_rows));
+  /* Builds the sketches on the GPU */
+  size_t Sketch(DMatrix *dmat, HistCutMatrix *hmat) {
```
canonizer: No need for a return value, just add an extra struct field to GPUSketcher.

src/common/hist_util.cu
```diff
@@ -341,63 +357,91 @@ struct GPUSketcher {
...
+    size_t row_stride = 0;
+    const MetaInfo &info = dmat->Info();
+
+    // Initialize Sketches for this dmatrix
+    SketchContainer sketch_container;
+    sketch_container.sketches_.resize(info.num_col_);
+    sketch_container.col_locks_.resize(info.num_col_);
+#pragma omp parallel for schedule(static) if (info.num_col_ > 1000)
+    for (int icol = 0; icol < info.num_col_; ++icol) {
```
canonizer: This looks like a constructor or initializer for SketchContainer.

src/common/hist_util.h
```diff
@@ -115,10 +115,13 @@ struct HistCutMatrix {
...
 
-/*! \brief Builds the cut matrix on the GPU */
-void DeviceSketch
-  (const SparsePage& batch, const MetaInfo& info,
-   const tree::TrainParam& param, HistCutMatrix* hmat, int gpu_batch_nrows);
+/*! \brief Builds the cut matrix on the GPU.
+ *   This returns the row stride across the entire dataset to avoid
+ *   further passes on the dmatrix to compute it.
+ */
+size_t DeviceSketch
```
canonizer: Passing the pointer to the row stride as a parameter may be more readable.

However, I'll leave this up to you.

src/common/hist_util.cu
canonizer: No need to pass `dist` as a parameter, it can be created inside this function.

src/common/hist_util.cu
```diff
@@ -324,9 +335,14 @@ struct GPUSketcher {
 
       // add cuts into sketches
       thrust::copy(cuts_d_.begin(), cuts_d_.end(), cuts_h_.begin());
+#pragma omp parallel for schedule(static)
```
canonizer: Consider doing it in a parallel loop only starting from a certain number of columns.

src/common/hist_util.cu
```diff
@@ -341,63 +357,91 @@ struct GPUSketcher {
...
+      auto get_size = [=] __device__(size_t row) {
+        return row_iter[row + 1] - row_iter[row];
+      }; // NOLINT
+
+      auto counting = thrust::make_counting_iterator(size_t(0));
+      using TransformT = thrust::transform_iterator<decltype(get_size),
+                                                    decltype(counting), size_t>;
+      TransformT row_size_iter = TransformT(counting, get_size);
+      row_stride_ = thrust::reduce(row_size_iter, row_size_iter + n_rows_, 0,
+                                   thrust::maximum<size_t>());
```
canonizer: This piece of code is fairly isolated. Consider putting it into a separate function.

src/tree/updater_gpu_hist.cu
```diff
@@ -1446,7 +1446,9 @@ class GPUHistMakerSpecialised{
 
     // Find the cuts.
     monitor_.StartCuda("Quantiles");
-    common::DeviceSketch(batch, *info_, param_, &hmat_, hist_maker_param_.gpu_batch_nrows);
+    // TODO(sriramch): The return value will be used when we add support for histogram
+    // index creation for multiple batches
+    (void)common::DeviceSketch(param_, hist_maker_param_.gpu_batch_nrows, dmat, &hmat_);
```
canonizer: No need to cast the return value to `void`. Just ignoring it is fine.

tests/cpp/common/test_gpu_hist_util.cu
```diff
@@ -1,50 +1,62 @@
...
+#include "../../../src/common/device_helpers.cuh"
+#include "../../../src/common/hist_util.h"
+
+#include "../helpers.h"
+
 namespace xgboost {
 namespace common {
 
-void TestDeviceSketch(const GPUSet& devices) {
+void TestDeviceSketch(const GPUSet& devices, bool use_external_memory = false) {
```
canonizer: Do you need a default parameter value here? Wouldn't updating calls to `TestDeviceSketch` be easier?

tests/cpp/common/test_gpu_hist_util.cu
```diff
@@ -1,50 +1,62 @@
...
-  DMatrixHandle dmat_handle;
-  XGDMatrixCreateFromMat(test_data.data(), nrows, 1, -1,
-                         &dmat_handle);
-  auto dmat = static_cast<std::shared_ptr<xgboost::DMatrix> *>(dmat_handle);
-
-  // parameters for finding quantiles
+  std::shared_ptr<xgboost::DMatrix> *dmat = nullptr;
+
+  size_t num_cols = 1;
+  if (!use_external_memory) {
```
canonizer: Consider `if (use_external_memory)` and swapping the branches.

tests/cpp/common/test_gpu_hist_util.cu
```diff
@@ -1,50 +1,62 @@
...
 
   // find quantiles on the CPU
   HistCutMatrix hmat_cpu;
   hmat_cpu.Init((*dmat).get(), p.max_bin);
 
   // find the cuts on the GPU
-  const SparsePage& batch = *(*dmat)->GetRowBatches().begin();
   HistCutMatrix hmat_gpu;
-  DeviceSketch(batch, (*dmat)->Info(), p, &hmat_gpu, gpu_batch_nrows);
+  (void)DeviceSketch(p, gpu_batch_nrows, (*dmat).get(), &hmat_gpu);
```
canonizer: No need to cast to `void`.

canonizer: `dmat->get()`

src/common/hist_util.cu
sriramch: elements within a vector have to be copy/assign-able or move-able. a mutex is neither of those.

src/common/hist_util.cu
sriramch: what is the issue with having this accessor? i can consider making it a const member function, if any, as the state isn't getting mutated. deviceshard already delineates what is public and private and i do not want to loosen the accessibility restrictions without fully understanding why it was made a class in the first place.

src/common/hist_util.cu
sriramch: will do

src/common/hist_util.cu
sriramch: will do

src/common/hist_util.cu
sriramch: will do.

src/common/hist_util.cu
sriramch: will do

src/common/hist_util.cu
sriramch: will do

src/common/hist_util.cu
sriramch: will do

src/tree/updater_gpu_hist.cu
sriramch: i'll clean all these up in a follow-up pr for the histogram. i added a todo and did this just to remind myself of this. the void return casting is to just to notate for now that the return value is ignored

tests/cpp/common/test_gpu_hist_util.cu
sriramch: i do not see an issue either way. i'll change it nonetheless.

tests/cpp/common/test_gpu_hist_util.cu
sriramch: will do

tests/cpp/common/test_gpu_hist_util.cu
sriramch: this is to indicate that the return value is ignored. i can do it c++ style if you want!

sriramch: will do...

src/common/hist_util.cu
canonizer: I think copying/assignment/moving only applies when resizing the vector. Creating a mutex vector of a fixed size should be possible.

I'm not a C++ expert here, however, so I may be wrong.

src/common/hist_util.cu
```diff
@@ -94,6 +94,17 @@ __global__ void UnpackFeaturesK
 struct SketchContainer {
   std::vector<HistCutMatrix::WXQSketch> sketches_;  // NOLINT
   std::vector<std::unique_ptr<std::mutex>> col_locks_; // NOLINT
+  SketchContainer(const tree::TrainParam &param, DMatrix *dmat) {
+    const MetaInfo &info = dmat->Info();
+    // Initialize Sketches for this dmatrix
+    sketches_.resize(info.num_col_);
+    col_locks_.resize(info.num_col_);
+#pragma omp parallel for schedule(static) if (info.num_col_ > 1000)
```
canonizer: This number occurs in more than one place. Consider defining an constant and using it instead.

src/common/hist_util.cu
```diff
@@ -372,69 +372,74 @@ struct GPUSketcher {
...
 
  private:
   std::vector<std::unique_ptr<DeviceShard>> shards_;
   const tree::TrainParam &param_;
   int gpu_batch_nrows_;
+  size_t row_stride_;
+  std::unique_ptr<SketchContainer> sketch_container_;
 };
 
 size_t DeviceSketch
```
canonizer: Please document what value is being returned.

Alternatively, consider using a pointer parameter instead of returning a value.

src/common/hist_util.cu
```diff
@@ -372,69 +372,74 @@ struct GPUSketcher {
...
-    return row_stride;
   }
 
   GPUSketcher(const tree::TrainParam &param, int gpu_nrows)
-    : param_(param), gpu_batch_nrows_(gpu_nrows) {
+    : param_(param), gpu_batch_nrows_(gpu_nrows), row_stride_(0) {
   }
 
-  /* Builds the sketches on the GPU */
+  /* Builds the sketches on the GPU for the dmatrix */
```
canonizer: Please document the return value.

tests/cpp/common/test_gpu_hist_util.cu
```diff
@@ -51,7 +51,7 @@ void TestDeviceSketch(const GPUSet& devices, bool use_external_memory = false) {
 
   // find the cuts on the GPU
   HistCutMatrix hmat_gpu;
-  (void)DeviceSketch(p, gpu_batch_nrows, (*dmat).get(), &hmat_gpu);
+  (void)DeviceSketch(p, gpu_batch_nrows, dmat->get(), &hmat_gpu);
```
canonizer: Could you also check that the correct row stride is returned?

src/tree/updater_gpu_hist.cu
canonizer: A TODO describing that you ignore the value is fine. No need to cast to `void`, ignoring return values in C/C++ is OK.

sriramch:
w.r.t. this:

> https://github.com/dmlc/xgboost/pull/4486#discussion_r288601814

to accommodate this, we have to explicitly make the type non-copy-able/move-able/assign-able. i have made just that to prevent such accidental usage.

src/common/hist_util.cu
sriramch: done

src/common/hist_util.cu
sriramch: done

tests/cpp/common/test_gpu_hist_util.cu
sriramch: done

hcho3:
@sriramch So is the external memory feature now complete? That is, can I take today's master branch and train with GPU and external memory data?

sriramch:
w.r.t. this:

> https://github.com/dmlc/xgboost/pull/4486#issuecomment-497498962

@hcho3 - not yet. as this pr states, it is only part 1 of 2. i'm working on the last part and hope to have the pr up soon. 

unfortunately, the last commit #4478 broke the xgboost cli, thus hampering my testing.

hcho3:
@sriramch It seems okay now: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/master/225/pipeline

sriramch:
@hcho3 - what i meant was that if i include this commit in my branch, then the cli just sits and spins chewing a lot of memory on a large dataset that has the external cache enabled, before it can even create those sparse pages on disk. this seems like a clear regression. hence, i reverted this commit in my private branch. i did not refer to the jenkins build error. sorry for not being clear earlier.

if you are interested, the full set of external memory training changes are [here](https://github.com/sriramch/xgboost/tree/full_training_extmem)

hcho3:
@sriramch I will need to look into the regression. Can you post a reproducible example?

My guess is that #4478 might have mishandled `#` in the data path that's used to enable external memory. Sorry for my oversight.

hcho3:
@sriramch Actually, your CLI config would be sufficient for me to diagnose the issue.

sriramch:
@hcho3 - please refer to this [gist](https://gist.github.com/sriramch/2174171d3bf8395b71222212fcb9594b)

yes, i think the '#' tokens are possibly ignored. is it possible to revert this for now?

hcho3:
@sriramch Indeed, your CLI config makes the problem very clear. I will resolve this issue very soon.

PR #4487, opened 2019-05-22
- fix issues with training with external memory on cpu

sriramch:
- use the batch size to determine the correct number of rows in a batch
   - use the right number of threads in omp parallalization if the batch size
     is less than the default omp max threads (applicable for the last batch)

@canonizer @RAMitchell @rongou please review.

### DIFF
--- a/src/common/hist_util.cc
+++ b/src/common/hist_util.cc
@@ -229,7 +229,7 @@ uint32_t HistCutMatrix::GetBinIdx(const Entry& e) {
 
 void GHistIndexMatrix::Init(DMatrix* p_fmat, int max_num_bins) {
   cut.Init(p_fmat, max_num_bins);
-  size_t nthread = omp_get_max_threads();
+  const size_t nthread = omp_get_max_threads();
   const uint32_t nbins = cut.row_ptr.back();
   hit_count.resize(nbins, 0);
   hit_count_tloc_.resize(nthread * nbins, 0);
@@ -247,20 +247,19 @@ void GHistIndexMatrix::Init(DMatrix* p_fmat, int max_num_bins) {
   size_t prev_sum = 0;
 
   for (const auto &batch : p_fmat->GetRowBatches()) {
-    auto bsize = static_cast<omp_ulong>(batch.Size());
-    nthread = std::min(bsize, nthread);
+    auto batch_size = static_cast<omp_ulong>(batch.Size());
 
     MemStackAllocator<size_t, 128> partial_sums(nthread);
     size_t* p_part = partial_sums.Get();
 
-    size_t block_size =  bsize / nthread;
+    size_t block_size =  batch_size / nthread;
 
     #pragma omp parallel num_threads(nthread)
     {
       #pragma omp for
       for (int32_t tid = 0; tid < nthread; ++tid) {
         size_t ibegin = block_size * tid;
-        size_t iend = (tid == (nthread-1) ? bsize : (block_size * (tid+1)));
+        size_t iend = (tid == (nthread-1) ? batch_size : (block_size * (tid+1)));
 
         size_t sum = 0;
         for (size_t i = ibegin; i < iend; ++i) {
@@ -272,15 +271,21 @@ void GHistIndexMatrix::Init(DMatrix* p_fmat, int max_num_bins) {
       #pragma omp single
       {
         p_part[0] = prev_sum;
-        for (int32_t i = 1; i < nthread; ++i) {
-          p_part[i] = p_part[i - 1] + row_ptr[rbegin + i*block_size];
+        if (batch_size < nthread) {
+          // Only the last thread handles it. Hence, the offset to the
+          // row_ptr shall be the number of entries accumulated thus far
+          p_part[nthread-1] = prev_sum;
+        } else {
+          for (int32_t i = 1; i < nthread; ++i) {
+            p_part[i] = p_part[i - 1] + row_ptr[rbegin + i*block_size];
+          }
         }
       }
 
       #pragma omp for
       for (int32_t tid = 0; tid < nthread; ++tid) {
         size_t ibegin = block_size * tid;
-        size_t iend = (tid == (nthread-1) ? bsize : (block_size * (tid+1)));
+        size_t iend = (tid == (nthread-1) ? batch_size : (block_size * (tid+1)));
 
         for (size_t i = ibegin; i < iend; ++i) {
           row_ptr[rbegin + 1 + i] += p_part[tid];
@@ -288,12 +293,12 @@ void GHistIndexMatrix::Init(DMatrix* p_fmat, int max_num_bins) {
       }
     }
 
-    index.resize(row_ptr[rbegin + bsize]);
+    index.resize(row_ptr[rbegin + batch_size]);
 
     CHECK_GT(cut.cut.size(), 0U);
 
     #pragma omp parallel for num_threads(nthread) schedule(static)
-    for (omp_ulong i = 0; i < bsize; ++i) { // NOLINT(*)
+    for (omp_ulong i = 0; i < batch_size; ++i) { // NOLINT(*)
       const int tid = omp_get_thread_num();
       size_t ibegin = row_ptr[rbegin + i];
       size_t iend = row_ptr[rbegin + i + 1];
@@ -316,8 +321,8 @@ void GHistIndexMatrix::Init(DMatrix* p_fmat, int max_num_bins) {
       }
     }
 
-    prev_sum = row_ptr[rbegin + bsize];
-    rbegin += bsize;
+    prev_sum = row_ptr[rbegin + batch_size];
+    rbegin += batch_size;
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/common/hist_util.cc
```diff
@@ -248,17 +247,20 @@ void GHistIndexMatrix::Init(DMatrix* p_fmat, int max_num_bins) {
   size_t prev_sum = 0;
 
   for (const auto &batch : p_fmat->GetRowBatches()) {
+    auto bsize = static_cast<omp_ulong>(batch.Size());
```
rongou: `batch_size`?

src/common/hist_util.cc
```diff
@@ -229,8 +229,7 @@ uint32_t HistCutMatrix::GetBinIdx(const Entry& e) {
 
 void GHistIndexMatrix::Init(DMatrix* p_fmat, int max_num_bins) {
   cut.Init(p_fmat, max_num_bins);
-  const int32_t nthread = omp_get_max_threads();
-  // const int nthread = 1;
+  size_t nthread = omp_get_max_threads();
```
rongou: Make this `const`?

src/common/hist_util.cc
```diff
@@ -248,17 +247,20 @@ void GHistIndexMatrix::Init(DMatrix* p_fmat, int max_num_bins) {
   size_t prev_sum = 0;
 
   for (const auto &batch : p_fmat->GetRowBatches()) {
+    auto bsize = static_cast<omp_ulong>(batch.Size());
+    nthread = std::min(bsize, nthread);
```
rongou: Wouldn't this make `nthread` too small if batch size fluctuates? Probably should declare a new variable.

src/common/hist_util.cc
sriramch: will do

src/common/hist_util.cc
sriramch: it is being changed later

src/common/hist_util.cc
sriramch: i think this is going to be pertinent only for the last batch. for the ones before that, nthread will be the default omp max. for the last batch, *if* it happens to be lesser (typically ~ 64), then it scales it down. the nthread is mostly pegged to the batch size in this block.

rongou: From tracing through some tests, the batch size doesn't seem to be constant and can fluctuate up and down. Seems safer to not make nthread a running min.

src/common/hist_util.cc
rongou: See comment below. Probably safer to declare another variable there.

src/common/hist_util.cc
sriramch: i have left the number of threads as is and adjusted the partial sums if the batch size is smaller than the number of threads.

the test should also now test all possible scenarios with batch size and thread pool size.

src/common/hist_util.cc
```diff
@@ -270,29 +271,34 @@ void GHistIndexMatrix::Init(DMatrix* p_fmat, int max_num_bins) {
       #pragma omp single
       {
         p_part[0] = prev_sum;
-        for (int32_t i = 1; i < nthread; ++i) {
-          p_part[i] = p_part[i - 1] + row_ptr[rbegin + i*block_size];
+        if (batch_size < nthread) {
```
RAMitchell: Would this also solve your problem?
```
const size_t nthread = std::min(batch_size, omp_get_max_threads());
```

src/common/hist_util.cc
```diff
@@ -248,17 +247,19 @@ void GHistIndexMatrix::Init(DMatrix* p_fmat, int max_num_bins) {
   size_t prev_sum = 0;
 
   for (const auto &batch : p_fmat->GetRowBatches()) {
+    auto batch_size = static_cast<omp_ulong>(batch.Size());
+
     MemStackAllocator<size_t, 128> partial_sums(nthread);
     size_t* p_part = partial_sums.Get();
 
-    size_t block_size =  batch.Size() / nthread;
```
RAMitchell: Why not use batch.Size()? Is this expected to change?

sriramch: since it is used in multiple places, i only invoke this once and use that variable elsewhere.

src/common/hist_util.cc
sriramch: sure - that is how i'd it initially ca904a8. then, i realized that not all thread blocks are pegged to the batch size. hence, i left the desired number of threads as-is.

sriramch:
w.r.t. this:
> https://github.com/dmlc/xgboost/pull/4487#discussion_r287441627

the batch size is referenced in multiple places in this block. the change wasn't done to somehow make it performant. it was simply done to localize getting the batch size in a single place, such that it is easy to change in the future, in case the means to retrieve it changes later.

src/common/hist_util.cc
```diff
@@ -229,8 +229,7 @@ uint32_t HistCutMatrix::GetBinIdx(const Entry& e) {
 
 void GHistIndexMatrix::Init(DMatrix* p_fmat, int max_num_bins) {
   cut.Init(p_fmat, max_num_bins);
-  const int32_t nthread = omp_get_max_threads();
-  // const int nthread = 1;
+  const size_t nthread = omp_get_max_threads();
```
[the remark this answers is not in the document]
sriramch: since the max_thread count cannot be a negative number, i made the return type reflect that.

src/common/hist_util.cc
```diff
@@ -270,29 +271,34 @@ void GHistIndexMatrix::Init(DMatrix* p_fmat, int max_num_bins) {
...
+        size_t iend = (tid == (nthread-1) ? batch_size : (block_size * (tid+1)));
 
         for (size_t i = ibegin; i < iend; ++i) {
           row_ptr[rbegin + 1 + i] += p_part[tid];
         }
       }
     }
 
-    index.resize(row_ptr.back());
+    index.resize(row_ptr[rbegin + batch_size]);
```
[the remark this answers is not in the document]
sriramch: this is the most important change in this pr. the index and the row_ptr is incrementally built with every batch. without this index will be an empty vector, if it has multiple batches.

tests/cpp/common/test_column_matrix.cc
```diff
@@ -50,5 +50,23 @@ TEST(DenseColumnWithMissing, Test) {
...
+    GHistIndexMatrix gmat; \
+    gmat.Init(dmat.get(), 256); \
+  }
+
+  // Vary the number of threads to make sure that the last batch
+  // is distributed properly to the available number of threads
+  // in the thread pool
+  TEST_HIST_IDX_MATRIX(20)
+  TEST_HIST_IDX_MATRIX(30)
+  TEST_HIST_IDX_MATRIX(40)
```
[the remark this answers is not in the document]
sriramch: i'll change this to a function

sriramch:
close and reopen to fix [this](https://travis-ci.org/dmlc/xgboost/jobs/536890523) transient error.

hcho3:
@sriramch Let's ignore the Mac OSX tests for the time being. Somehow we are getting the error
```
MD5MismatchError: Conda detected a mismatch between the expected content and downloaded content
```

hcho3:
@sriramch Mac OSX tests are no longer required for merging. Please ignore Mac OSX tests until I find time to fix them

sriramch:
@hcho3 the issue seems to be with python tests. see here - https://travis-ci.org/dmlc/xgboost/builds/536902957

it has happened twice - similar issue. wouldn't this be deemed a build failure then?

hcho3:
@sriramch No, because the failure is due to failed installation of Miniconda Python environment. Look for the error message
```
MD5MismatchError: Conda detected a mismatch between the expected content and downloaded content
```

src/common/hist_util.cc
RAMitchell: `batch.Size()` is the "most true" source of this information so you should use it directly if there is no performance penalty. This may seem pedantic but adding layers between effects readability and creates assumptions. For example that batch.Size() has not changed since you first called it - in this case maybe a reasonable assumption, but in general this as a reason not to do this.

Hope this makes sense.

src/common/hist_util.cc
RAMitchell: So the PR does not work if you do this, or there are performance side-effects? I am asking because I prefer simple solutions.

src/common/hist_util.cc
sriramch: i do not want to make assumptions on performance implications without profiling first. hence, this wasn't done due to performance reasons.

i'm trying to understand how using a variable named `batch_size` as opposed to using `batch.Size()` impacts readability, given that this is used in multiple places in the code block. by using the variable here, we are also conveying the intent that a new value isn't used at the call site and that a constant value computed at the beginning of the block is used.  further, this also localizes the means to compute the batch size in a single place, such that *if* the way to compute the retrieve the batch size changes later, the change is limited to a single place.

src/common/hist_util.cc
sriramch: the original intent was that the number of threads do not change, and i tried to preserve that intent. if the number of threads is pegged to the batch size everywhere, then this change would have preserved that intent. however, it is also pegged to the number of bins later. hence, if you'd a large number of bins and a smaller batch size, then that could be impacted with a change like this. hence, i decided to leave it as-is.

RAMitchell: So can't you just uniquely define the number of threads before each parallel block?

sriramch: sure, there could be multiple ways to solve this. doing this would also mean that someone would have to remember to readjust the number of threads every time a block here is added/moved; when the block logic changes to parallelize on something other than batch_size or number of bins etc.

Issue #4169, opened 2019-02-20
GPU Support for Feature Interaction Constraints


laceyg:
Is there any active work or plan to support [Feature Interaction Constraints](https://xgboost.readthedocs.io/en/latest/tutorials/feature_interaction_constraint.html) on GPUs with ```gpu_hist``` and ```gpu_exact```?  Is there anything fundamental about this feature that would violate a parallel implementation?

hcho3:
> Is there anything fundamental about this feature that would violate a parallel implementation?

This is more due to the fact that we have separate code path for GPU and CPU implementations of the training algorithm. The feature interaction constraints require `SplitEvaluator` class, which is only available from the CPU implementation currently.

trivialfis:
@laceyg Closing as now interaction constraint supports `gpu_hist`, and `gpu_exact` is deprecated.

Issue #4354, opened 2019-04-10
[RFC] Possible DMatrix refactor


RAMitchell:
As we add different types of algorithms to xgboost and as the performance of these algorithms  improves, the DMatrix class may also be improved in order to improve performance and memory usage for these particular cases.

**Current DMatrix**
The current DMatrix class has two sub classes. The first is the primary in memory representation where the entire dataset is ingested and converted into a CSR format with values stored as 32-bit floats. The entire dataset may be transposed in memory into a column format for exact style algorithms.

The second is the external memory representation which constructs a number of binary page files (32mb in size by default) which are streamed from disk. If a column format is requested, new transposed pages will be generated and these will be streamed from disk.

The end user does not directly choose the type of DMatrix and these subclasses are not exposed via external APIs. The type of underlying DMatrix is automatically selected according to if the user supplies a cache prefix in the constructor.

Currently a batch of data is stored inside the DMatrix as a 'SparsePage' class. This uses HostDeviceVector in order to expose data to both the CPU and GPU.

**Current fast histogram algorithms**
The current static histogram algorithms (hist, gpu_hist) convert the DMatrix object on the fly inside of their respective TreeUpdater implementations, not as a part of DMatrix implementation, although I believe that it was eventually intended to be integrated with DMatrix (#1673). 'hist' converts the data into a CSR matrix with integers instead of floats. 'gpu_hist' converts the data into an ELLPACK format with integers instead of floats and additionally applies bitwise compression at run time to both the matrix elements and indices, commonly resulting in 4-5x compression over the standard CSR matrix. In gpu_hist we avoid ever copying the training DMatrix to the GPU if prediction cacheing is used.

 **Some desirable features for DMatrix**
Here I will list some features that we would like to have. Not all of these will be practical under the same design.
- The ability to train entirely on a compressed quantised DMatrix. We could save around ~4x memory and be much more competitive with LightGBM on memory usage.
- Support for dense matrices. We would directly halve memory usage by using dense formats where appropriate.
- DMatrix as a thin wrapper over an already allocated data matrix. e.g. a numpy ndarray. Given a dense numpy input, we copy this into DMatrix format resulting in 3n memory usage. Simply using the underlying numpy array as our storage uses 1n memory.
- Support for ingest of data from GPU memory (see #3997)
- Support for external memory for all of the above

 **Notes on possible implementation**
[...]

PR #4488, opened 2019-05-22
Feature interaction constraint for GPU Hist.

trivialfis:
This is still WIP,  my approach is to just drop all features that don't comply to constraints before entering split evaluation.  Not sure about how to merge it with CPU SplitEvaluator yet.

GPU Exact will be left to another PR.  Initially I tried to implement a kernel side evaluator that rejects features during split evaluation, see: https://github.com/dmlc/xgboost/commit/70a59367ec9c48a798e475d7a8c4f08c89133a32 , which is much more complicated and now replaced, but might still be useful for GPU EXACT.

related to #4169 

@RAMitchell


RAMitchell:
I don't really care about gpu_exact and am willing to deprecate the algorithm in future. Probably after we do some work to improve the performance of sparse data sets for gpu_hist.

trivialfis:
I'm not sure about why would VC failed to build generated host stub from `host_device_vector.cu`.

hcho3:
@trivialfis I found https://issues.jenkins-ci.org/plugins/servlet/mobile#issue/JENKINS-9104. Looks like we have issues running multiple MSBuild jobs running in parallel.

trivialfis:
@hcho3 Could you help working around it?  From what Andreas said, it seems to be a matter of setting envs.

hcho3:
Let me see if I can improve stability of Windows tests

trivialfis:
@RAMitchell 
> Probably after we do some work to improve the performance of sparse data sets for gpu_hist.

I'm thinking about that too.  Perhaps the feature bundling from lgb would help, but it has memory usage issue as mentioned by @hcho3 in https://github.com/dmlc/xgboost/issues/4354#issuecomment-495025316

trivialfis:
This PR uses `SplitEvaluator` to enable feature interaction support for GPU Hist.  This avoids creating another interaction constraint implementation,  but still we need to think about how to merge the code path of other split evaluators.  One thing worth noting is that, interaction constraint is different from other split evaluators in two ways, it has dynamic size buffer to track tree splits, and it doesn't have to be used during actual split evaluation. So in the future we might want to distinguish it from others.

RAMitchell:
I think we need to build a DeviceSplitEvaluator class that shadows the functionality of SplitEvaluator but uses device memory internally and may be updated on the device (e.g. when registering new splits I do not want to copy memory up and down, this is a very large performance penalty). Do you think this is possible?

trivialfis:
@RAMitchell 

For feature interaction constraint (FIC), it won't be necessary since feature sampling has already been using host memory, FIC does not do any extra host device copying. 

For other split evaluators, I don't think the current abstract interface of `SplitEvaluator` is useful, it's just making things more complicated than necessary.  There are only 3 of them, technically FIC is better used for feature selection than gradient gain evaluation.  So 2 `if else` are all we need ...

RAMitchell:
Its not strictly true that its already doing a copy - I think if column sampling is 1.0 the column sampler always returns the same HostDeviceVector and no copy occurs. The cost of introducing a single small memcpy between host and device for each node may be about 30% increased runtime.

trivialfis:
@RAMitchell Sorry for the ambiguity.  I'm aware of this.  The correct term is: no more than using feature sampling.

The difficulties of implementing this on device are: 

- It needs to use `set`. Children merge their parents' constraint set during split.  Ordering doesn't matter, but no duplication is allowed.  Situations like `[[0, 2], [1, 3, 4], [5, 6], [3, 5]]` where `3` appears in two different set can happen.  Say root splits at `1`, and its left child splits at `5`.  During split the two set `[1, 3, 4], [3, 5]` needs to be merged into child because children should be able to interact with existing parent's split (also this is what the current implementation do).

- FIC can come in with varying size, from single element `[[0]]` to all features.

-  Also device memory allocation is inevitable since every node needs to have its own set.  Combined with first difficulty, to avoid copying, we need to 1. compute the size of set on device (the information is not copied to host). 2. Allocate memory on host. 3 Come back to device and compute the actual set.  I assume this is not much better than copying memory.

Currently I don't have any good idea for how to meet your requirement yet.  But suggestions are welcomed.  If it's any consolation, I plan to bring back feature grouping to reduce sparsity #4501, which should bring some nice improvement for GPU Hist.

RAMitchell:
To implement a set I would use a boolean (bit?) vector of length n_features. You should know how much memory to allocate ahead of time because the maximum number of nodes is always constrained in the GPU algorithm and we know how many interaction sets there are. Should be fun to try and implement :) 

Looking at monotone constraints and feature interaction constraints at a higher level, it seems to me that there is a desire from users to directly influence the optimisation process of tree construction. I wonder if there is a more general way of specifying this as an interface? Probably not but interesting to think about.

trivialfis:
@RAMitchell Let me wrap up other PRs first so I can focus on this.

trivialfis:
Closing for now, will open a new PR when it's ready.

PR #4499, opened 2019-05-24
Add dll_path for cygwin users

yellowdolphin:
When built under cygwin, the Makefile produces 'cygxgboost.dll', which dll_path.py did not cover.

### DIFF
--- a/python-package/xgboost/libpath.py
+++ b/python-package/xgboost/libpath.py
@@ -37,6 +37,8 @@ def find_lib_path():
         dll_path = [os.path.join(p, 'libxgboost.so') for p in dll_path]
     elif sys.platform == 'darwin':
         dll_path = [os.path.join(p, 'libxgboost.dylib') for p in dll_path]
+    elif sys.platform == 'cygwin':
+        dll_path = [os.path.join(p, 'cygxgboost.dll') for p in dll_path]
 
     lib_path = [p for p in dll_path if os.path.exists(p) and os.path.isfile(p)]
 


trivialfis:
@yellowdolphin I never used cygwin before, what's the result from CMake?

yellowdolphin:
cmake is all good, except
-- Looking for fopen64 - not found
-- Looking for backtrace - not found
-- Could NOT find Backtrace (missing:  Backtrace_LIBRARY Backtrace_INCLUDE_DIR)

Also, under cygwin fopen64 is not replaced in xgboost/rabit/include/rabit/internal/utils.h:
#if !defined(__GNUC__) || defined(__FreeBSD__)
#define fopen64 std::fopen
#endif  // !defined(__GNUC__) || defined(__FreeBSD__)

When I activate that "#define", make runs through as well.

Apart from the dll name issue, setup.py is happy and the python API runs fine :-)

trivialfis:
@yellowdolphin Thanks for the information.  Also for the PR.  Would you like to fix the CMake build too? :)

RAMitchell:
If we want to support Cygwin it would be good to add some CI tests (either to Appveyor or Jenkins). Otherwise you may find the build gets broken in future.

yellowdolphin:
Concerning the built and cygwin support, the code is now somewhat inconsistent: I looked into the fopen64 issue and found that dmlc-core already considers cygwin and quite a list of other OSs where fopen64 is not present. But in rabit (rabit/include/rabit/internal/utils.h), only for FreeBSD and apple fopen64 is replaced by fopen (if glic is used). I wonder if this does not break the build for other OS's as well.

I guess the quick solution would be to apply the same logic in rabit/dmlc-core whether to use fopen64. On the other hand, I read that fopen64 is obsolete and one should instead set
_FILE_OFFSET_BITS=64
which then for 32-bit linux would map fopen to fopen64 to allow reading >4GB files.

https://www.gnu.org/software/libc/manual/html_mono/libc.html#Opening-Streams
https://stackoverflow.com/questions/4357570/use-file-offset64-vs-file-offset-bits-64

So, would it be safe to get rid of fopen64 altogether?

I can prepare a PR for whatever you prefer, but for the 2nd solution, some testing might be appropriate and I don't have a 32-bit system with 4GB mem...

trivialfis:
@yellowdolphin Thanks for digging into it.  I will open a new issue concerning `fopen64`.

Issue #2326, opened 2017-05-20
Fast histogram algorithm exhibits long start-up time and high memory usage


Laurae2:
Two issues in one:

* Fast histogram xgboost does not start training after a while (stuck in building matrix?)
* Fast histogram xgboost uses a massive amount of memory (reaching over 50GB RAM when exact xgboost takes only a peak of about 10GB RAM)

Before running, to know:

* Will not work using Microsoft R versions (incompatible due to Microsoft R spamming messages about sending data collection to Microsoft while compiling packages)
* Dataset download is 240MB, uncompressed is 2GB, sum of RDS files are 650MB
* Make sure to have about 25GB RAM free to fast histogram (25GB extra seem committed in memory but never used for unknown reasons)
* Creating the dataset (3GB, 7,744,201,107,060 elements) requires using my svmlight package in R
* My scripts can run copy&paste just by changing the folders appropriately in `setwd`
* I expected fast histogram to be slower than exact xgboost due to sparsity, but not by that much

Created files info:

```
  File: url_svmlight.tar.gz
CRC-32: c152f632
   MD4: b05d0a58ad6f53f9ad20a5f350c25df2
   MD5: 2eb74f59d06807a3845f90ed36fe8ded
 SHA-1: 605121dde12ce7baf098e30cd8856ef4ed6d5e69
```

```
  File: reput_sparse.rds
CRC-32: 846b1907
   MD4: 2ea36b208118b7d46d18ccb5fad8da98
   MD5: ce371f418463ebc799367ed41f5fd654
 SHA-1: 7f2a73609d72ff61a24a6d176c44752b68954ea6
```

## Environment info
Operating System: Windows Server 2012 R2 (baremetal) - OS doesn't matter, happens also in Ubuntu 17.04

Compiler: MinGW 7.1

Package used (python/R/jvm/C++): R 3.4

`xgboost` version used: from source, latest commit of today (e5e7217)

## Computer specs:
* Quad E7-88xx v4 (72 cores), but 3 CPUs were disabled to get rid of NUMA
* 1TB RAM 1866MHz RAM (256GB RAM available to get rid of NUMA)
* Not virtualized

Used 8 threads for xgboost because it was for a benchmark comparison.

## Steps to reproduce

1. Download this dataset: http://archive.ics.uci.edu/ml/datasets/URL+Reputation (svmlight format, 121 parts, 2396130 rows, 3231961 features
2. Load data using my svmlight parser
[...]

PR #4501, opened 2019-05-25
Small optimization for block hist matrix.

trivialfis:
This is an early WIP.  Please ignore copied and pasted code.  Initial benchmark with HIGGS:

* Platform : Ubuntu 18.10
* CPU : AMD Ryzen 7 2700 Eight-Core Processor

Results:

- Time spent in `GHistIndexBlockMatrix::Init`

| Before | Now |
|----------|-------|
| 7.56465 s | 4.09427 s  (-45.87%) |

- High water marked (peak) memory usage with `GHistIndexBlockMatrix::Init`:

|  | Before | Now |
|--|----------|-------|
| Child high-water RSS                                     |   12138724 KiB | 8359052 KiB (-31.13%) |
| Recursive and acc. high-water RSS+CACHE |   19922500 KiB | 8371424 KiB (-57.98%) |

This is mostly for addressing memory usage issue.  The benchmark is carried out by `tests/benchmark/block_hist_init.cc`.  Hopefully with more effort we can bring back feature grouping, which might be beneficial to GPU Hist too.

One particular thing worth noting is that, I tried injecting OMP into many places, it ends up single thread version is the fastest one.


trivialfis:
Sadly this is not even close to fixing #2326 .

trivialfis:
@hcho3 The benchmarking code is not necessary, but very useful for  #4503 .  You can decide whether we should keep it here.

hcho3:
@trivialfis I'll review this PR later in detail. I think we should keep the benchmark code. Later, we can expand it as part of a regular performance benchmark.

trivialfis:
@hcho3 Take your time.  This is part of memory optimization for hist and is not the most critical one.

PR #4433, opened 2019-05-02
CPU optimizations

SmirnovEgorRu:
This PR is changes from https://github.com/dmlc/xgboost/pull/4278 without pre-processing optimizations (it has been merged yet https://github.com/dmlc/xgboost/pull/4310)

In addition I refactored the code to make it simple. In previews version I used omp tasks, but they are not supported in MSVS by default, so I replaced them by simple omp parallel for.

Code changes looks massive, but it is hard to split them on small parts.

PR #4502, opened 2019-05-25
Revert hist init optimization.

trivialfis:
@SmirnovEgorRu @hcho3 @RAMitchell

See https://github.com/dmlc/xgboost/pull/4433#issuecomment-495951502 for detail.

### DIFF
--- a/src/common/hist_util.cc
+++ b/src/common/hist_util.cc
@@ -50,7 +50,7 @@ void HistCutMatrix::Init(DMatrix* p_fmat, uint32_t max_num_bins) {
   constexpr int kFactor = 8;
   std::vector<WXQSketch> sketchs;
 
-  const size_t nthread = omp_get_max_threads();
+  const int nthread = omp_get_max_threads();
 
   unsigned const nstep =
       static_cast<unsigned>((info.num_col_ + nthread - 1) / nthread);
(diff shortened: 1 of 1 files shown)


hcho3:
I think we should pick a few datasets with different characteristics and run benchmarks:
* Dense dataset with small number of columns
* Sparse dataset with high number of columns
* Sparse dataset whose columns are genera by a high-cardinality categorical variables

trivialfis:
That would be a very good idea.

thvasilo:
@hcho3 The datasets we used in our paper would be good candidates for that.

Bosch: Dense with ~970 features: https://www.kaggle.com/c/bosch-production-line-performance/data
URL: sparse with 3.2M features, generated from text as well as numerical features http://www.sysnet.ucsd.edu/projects/url/
Avazu: Sparse with 1M features https://www.csie.ntu.edu.tw/~cjlin/libsvmtools/datasets/binary.html and https://www.kaggle.com/c/avazu-ctr-prediction/data
RCV1: Sparse with ~47k features, also available at https://www.csie.ntu.edu.tw/~cjlin/libsvmtools/datasets/binary.html

I'm not aware of high-dimensional sparse datasets that were not generated by high-cardinality categorical variables.

One thing to note about URL vs. Avazu is perhaps that avazu strangely uses the value 0.258 (instead of 1) to indicate the presence of a variable, I'm not sure if that affects the parsing (i.e. treating it as a dense variable where it isn't)

Issue #4503, opened 2019-05-26
Improve sparse dataset memory usage for hist tree method.


trivialfis:
@hcho3  @RAMitchell 

As part of refactoring DMatrix,  I'm trying to profile memory usage issue across XGBoost.  One glaring issue in hist method is that it uses feature index to build the histogram cut matrix.  For dense dataset this is optimal, but for sparse dataset this results in very huge waste of memory.  One particular reproducible dataset would be [URL](http://archive.ics.uci.edu/ml/datasets/URL+Reputation)

The linked dataset has total size of 2.2 GB in text form.  After some optimizations and limiting to first 450000 lines of dataset, it still costs 28.6 GB of memory.  Here is a simple report from valgrind:

![Screenshot from 2019-05-26 21-14-27](https://user-images.githubusercontent.com/16746409/58382312-5f115100-7ffb-11e9-8a2a-38c64142578e.png)

In some selected rows of above dataset, largest feature index is 3231962, but actual number of none zero columns is just 394.  Is there any way we can compress the histogram cut building into a form that doesn't depend on feature index?

trivialfis:
And I need to see if it's possible to do feature grouping before building histogram cut.

CodingCat:
When I was working on distributed hist, I f found hist were still built for all features even with col sampling. I thought this is the bottleneck and work on it

At that time, I added something like a featureIdToHistIndex (but I didn’t merge it since I found the real bottleneck is rabit)...is the similar approach enough here?

trivialfis:
@CodingCat This is still at quantiles cut building stage.  I'm not sure about hist updater yet.  Your approach is my initial idea.  But maybe there are many other problems that I don't know.  So I figured I should ask for some suggestions here first.

RAMitchell:
Yes we should definitely look into cut building scalability with number of features. URL is a great test case. I would like to apply what we learn to the GPU algorithms as well.

Regarding training, I believe LightGBM has the concept of a sparse histogram. Another approach that I have seen is to switch to exact style split search when the number of training examples low (e.g. less than the number of features).

thvasilo:
I think this is quite related to the problem we tried to solve with @hcho3 in our [recent paper on block-distributed GBTs](https://arxiv.org/abs/1904.10522).

Using a sparse histogram representation gave us orders of magnitude improvements in histogram size. We could/should try to find a way for an efficient sparse histogram representation that does works with Rabbit's limitation of knowing the object size in advance.

[...]

PR #4508, opened 2019-05-28
Provide example of custom distributed evaluation function via dask

RAMitchell:
This is an example of how to write a distributed evaluation function from the Python side. @CodingCat this is what I was thinking of in the discussion of #4280. It does require understanding what AllReduce does but apart from that I think it's quite simple. This approach is fully general, a user could implement any possible evaluation function.

### DIFF
--- a/demo/dask/dask_custom_eval_demo.py
+++ b/demo/dask/dask_custom_eval_demo.py
@@ -1,0 +1,45 @@
+from dask.distributed import Client, LocalCluster
+import dask.dataframe as dd
+import dask.array as da
+import numpy as np
+import xgboost as xgb
+
+
+# Define rmse function using Python that mimics the internal xgboost rmse function
+def custom_rmse(preds, dtrain):
+    labels = dtrain.get_label()
+    weights = dtrain.get_weight()
+
+    mse_local_sum = sum(np.square(labels - preds) * weights)
+    weights_local_sum = sum(weights)
+    # Pass the sum mse and weights to an allreduce call as a numpy array
+    result = xgb.rabit.allreduce(np.asarray([mse_local_sum, weights_local_sum]), 2)
+    # Result now contains the sum over all workers
+    return 'my-error', np.sqrt(result[0] / result[1])
+
+
+def train(X, y, weights):
+    dtrain = xgb.dask.create_worker_dmatrix(X, y, weight=weights)
+
+    evals_result = {}
+    bst = xgb.train({}, dtrain, evals=[(dtrain, 'train')], evals_result=evals_result,
+                    feval=custom_rmse)
+    # Check our custom function outputs roughly the same result as the built-in version
+    assert np.allclose(evals_result['train']['rmse'], evals_result['train']['my-error'])
+
+
+def main():
+    cluster = LocalCluster(n_workers=2, threads_per_worker=2)
+    client = Client(cluster)
+
+    n = 10
+    m = 100
+    partition_size = 20
+    X = da.random.random((m, n), partition_size)
+    y = da.random.random(m, partition_size)
+    weights = da.random.random(m, partition_size)
+    xgb.dask.run(client, train, X, y, weights)
+
+
+if __name__ == '__main__':
+    main()


CodingCat:
I agree the code would not be complicated if we look at an individual case

But for the cases in our current view, users essentially have to write the same piece of code like here to implement each custom function

And in xgb, we have provide the same functionality in c++, my point is why we put this burden to user again (for many users I talked with I do not even believe they can write it correctly)

trivialfis:
@RAMitchell @CodingCat I think it's possible to do allreduce inside `Learner`.

trivialfis:
I'm curious about whether we can call back to c++ for all reduce automatically given a customized metric.  We can add a C API and let language bindings call it after calculating customized metric.  It might not be general enough, but seems reasonable to think about.

hcho3:
@trivialfis Do you mean something like #4280?

trivialfis:
Much less complicated I think.

demo/dask/dask_custom_eval_demo.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Define rmse function using Python that mimics the internal xgboost rmse function
+def custom_rmse(preds, dtrain):
+    labels = dtrain.get_label()
+    weights = dtrain.get_weight()
+
+    mse_local_sum = sum(np.square(labels - preds) * weights)
+    weights_local_sum = sum(weights)
+    # Pass the sum mse and weights to an allreduce call as a numpy array
+    result = xgb.rabit.allreduce(np.asarray([mse_local_sum, weights_local_sum]), 2)
```
trivialfis: @CodingCat My question is why calling these 2 lines requires that much of code in #4280 .  Personally I prefer the code in this PR @RAMitchell  since it seems much simpler,  I glanced through the code and understood how to do it immediately.  And I believe I can debug my code much easier since I can just put a print in here and get to know everything.

Revisiting comments in #4280 I get that you want to mask the complexity in dealing with rabit and multiclass/rank metrics @hcho3 .  I like the idea of making things simple for users, but can the layer be done as simple as these two lines?

trivialfis: @CodingCat Also I think there's no need to mask the `sum` operation in #4280 .  What if I want to use `mean` or `max` instead of `sum`?

RAMitchell:
Still need to do a couple of things on this. The python rabit API is not actually documented, there are no tests, there are no enums defined for AllReduce ops. Would also be good to get some agreement and consistency with spark before we go down this route (if at all).

CodingCat:
To me, exposing rabit to users is equivalent to ask users to learn how NCCL/Arrow works when they want to use RAPIDS. It's not my taste to develop a software for users. Blocking something for some non-concrete reason is not my taste to manage software development either.

Not masking `sum` operation, etc. can be resolved by providing more general abstractions in c++ layer, why #4280 contains a lot of code can be explained as (1) most of changes are moving stuffs; (2) I need to create abstractions instead of just write two lines of plain code because Java/Scala accepts a class instead of a function

I am too busy to continue the discussion here which is essentially a duplication of #4280. I don't think I would change my taste about what is to be exposed to the users just because of this. 

So please go ahead with the approach here. I will close #4280 and leave the community or myself (when I have time) to make it consistent with python part

hcho3:
@CodingCat @trivialfis @RAMitchell 

I don't think exposing Rabit to users is the best path forward.

1. I am noticing that users are trying to make their customized metrics work and fail, especially for multi-class classification, learning-to-rank, and survival tasks. We don't have to add yet another obstacle.
2. We may want to adopt other collective primitives other than AllReduce. For example, in https://github.com/dmlc/xgboost/issues/4503#issuecomment-496181250, we may want to adopt a sparse representation to reduce memory consumption for sparse, high-dimensional datasets, and current (dense) AllReduce may not be optimal.

Moving forward: I think we can still merge this PR. Later, we should take time to decide on a good, high-level metric interface for different tasks (binary classification, multi-class classification, regression, learning-to-rank, survival, counting) and implement the high-level interface on top of this PR. So for most users, AllReduce call would be automated (as @trivialfis suggested).

Note to @CodingCat: I liked the general idea of #4280. I will leave my comments about the metric interface soon.

demo/dask/dask_custom_eval_demo.py
hcho3: @trivialfis I think the user can define `getFinal()` to implement a custom reduce op

hcho3: Actually no, `getFinal()` is not a custom reduce op. Sorry about that. My brain is fried after long day of work

RAMitchell:
> We may want to adopt other collective primitives other than AllReduce. For example, in #4503 (comment), we may want to adopt a sparse representation to reduce memory consumption for sparse, high-dimensional datasets, and current (dense) AllReduce may not be optimal.

I don't think dimensionality of the data matters for evaluation because we only consider rows. I can't think of a case where the AllReduce would be problematic. Maybe for ranking where you have a huge number of groups?

> I don't think exposing Rabit to users is the best path forward.

I can try implementing @CodingCat's higher-level interfaces here and see what it looks like. There is no rush to merge this. My feeling is we can do this without touching C++ code. Performance should be adequate using numpy and it's only proportional to number of rows so evaluation doesn't tend to be a bottleneck.

hcho3:
@RAMitchell 

> I don't think dimensionality of the data matters for evaluation because we only consider rows

This is true for metrics. So AllReduce will be appropriate for the customized distributed metrics. The link I put up was referring to the use of AllReduce in gradient histogram calculation.

> My feeling is we can do this without touching C++ code

I personally favor baking *good* (emphasis on "good") metric interface into the C API. That way, we don't have to duplicate the distributed code across multiple language wrappers. However, I'm pretty flexible on this point: as long as the user can stick to the high-level API without touching Rabit, I'm fine with putting all things in the wrapper level.

Aside: Let's also consider a path to supporting user-defined objective (loss) function in distributed setting.

avinashbarnwal:
@hcho3 Survival modeling, We can use Discrimination indexes such as C-index, Brier Score along with negative log-likelihood that is currently present in the Cox-PH model. This document looks good. 
https://bmcmedresmethodol.biomedcentral.com/articles/10.1186/1471-2288-12-102

PR #4510, opened 2019-05-28
- fix crash with approx tree method on cpu

sriramch:
fix crash with approx tree method on cpu when training is enabled with external memory

@trivialfis  - please review

### DIFF
--- a/tests/cpp/data/test_data.cc
+++ b/tests/cpp/data/test_data.cc
@@ -55,13 +55,26 @@ TEST(SparsePage, PushCSC) {
 }
 
 TEST(SparsePage, PushCSCAfterTranspose) {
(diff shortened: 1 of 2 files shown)


tests/cpp/data/test_data.cc
```diff
@@ -52,4 +53,15 @@ TEST(SparsePage, PushCSC) {
     ASSERT_EQ(inst[i].index, indices_sol[i % 3]);
   }
 }
+
+TEST(SparsePage, PushCSCAfterTranspose) {
```
trivialfis: Can we have specific expected results other than not crashing in this test?

sriramch: i have modified the test. please check now...

sriramch:
@trivialfis thanks for your review. can you merge once the build passes? i need this for my tests

Issue #4513, opened 2019-05-29
h2o4gpu test failures


pseudotensor:
on master after GPU dup merge,  I get from h2ogpu testing these failures that don't allow me to merge to master:

`tests/python/open_data/gbm/test_gpu_prediction_pickledmodel.py`

```
[2019-05-29T18:40:33.963Z] E           xgboost.core.XGBoostError: [18:21:05] /root/repo/xgboost/src/predictor/gpu_predictor.cu:384: Check failed: devices.Size() != 0 (0 vs. 0) : 

[2019-05-29T18:40:33.963Z] E           Stack trace:

[2019-05-29T18:40:33.963Z] E             [bt] (0) /opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x22) [0x7f13c7dce0b2]

[2019-05-29T18:40:33.963Z] E             [bt] (1) /opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(xgboost::predictor::GPUPredictor::PredictBatch(xgboost::DMatrix*, xgboost::HostDeviceVector<float>*, xgboost::gbm::GBTreeModel const&, int, unsigned int)+0x534) [0x7f13c7f8aa64]

[2019-05-29T18:40:33.963Z] E             [bt] (2) /opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(xgboost::LearnerImpl::Predict(xgboost::DMatrix*, bool, xgboost::HostDeviceVector<float>*, unsigned int, bool, bool, bool, bool) const+0x67) [0x7f13c7e4ece7]

[2019-05-29T18:40:33.963Z] E             [bt] (3) /opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(XGBoosterPredict+0x113) [0x7f13c7ddc533]

[2019-05-29T18:40:33.963Z] E             [bt] (4) /opt/h2oai/h2o4gpu/python/lib/python3.6/lib-dynload/../../libffi.so.6(ffi_call_unix64+0x4c) [0x7f142a74cec0]

[...]

PR #4516, opened 2019-05-30
Fix prediction from loaded pickle.

trivialfis:
Fix https://github.com/dmlc/xgboost/issues/4513 .

@pseudotensor @sh1ng

### DIFF
--- a/include/xgboost/learner.h
+++ b/include/xgboost/learner.h
@@ -119,7 +119,7 @@ class Learner : public rabit::Serializable {
                        bool pred_leaf = false,
                        bool pred_contribs = false,
                        bool approx_contribs = false,
-                       bool pred_interactions = false) const = 0;
+                       bool pred_interactions = false) = 0;
 
   /*!
    * \brief Set additional attribute to the Booster.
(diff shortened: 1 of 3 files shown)


trivialfis:
The test is adopted from https://github.com/h2oai/h2o4gpu/blob/master/tests/python/open_data/gbm/test_gpu_prediction_pickledmodel.py#L189-L222

trivialfis:
@hcho3 I ended up removing the const from predict method.  One way to keep the const is to make train_param a pointer even in Learner, but that might not be really useful.

pseudotensor:
Thanks @trivialfis !

pseudotensor:
FYI, the idea is we typically build a model on the GPU, then for scoring the system could be arbitrarily different, with more or less GPUs or even none.  And, even if have GPUs when scoring, we still want to be able to choose to use only CPU.

trivialfis:
@pseudotensor See last part of tests/python-gpu/test_pickling.py in this PR.  I tested switching to cpu.

pseudotensor:
Yes, I saw thanks.  I was just explaining the diversity of what we would like to do in case context helps.

trivialfis:
Aha, got it.

pseudotensor:
merge?

PR #4519, opened 2019-05-31
 Smarter choice of histogram construction for distributed gpu_hist

RAMitchell:
This PR addresses a performance regression that was introduced to the distributed GPU algorithms in #4343, after which we no longer intelligently selected whether to build the histogram on the left or right node based on how much data was in each partition. The performance regression only applies to the distributed or multi-GPU version.

To solve this I needed to construct a simple max AllReduce operation that works over multiple threads as rabit is not thread safe.

### DIFF
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -7,6 +7,7 @@
 #include <thrust/system/cuda/error.h>
 #include <thrust/system_error.h>
 #include <xgboost/logging.h>
+#include <rabit/rabit.h>
 
 #include "common.h"
 #include "span.h"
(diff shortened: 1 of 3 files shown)


mt-jones:
Using XGBoost before https://github.com/dmlc/xgboost/pull/4343 and after: 50s >> 1min 50s on a common training set (mortgage workflow in RAPIDS). See notebook [here](https://github.com/rapidsai/notebooks-extended/blob/master/advanced/E2E/mortgage/mortgage_e2e.ipynb). 

When using this PR in the aforementioned workflow, I measure training times back at 50s. This certainly addresses the performance regression we observed. Thanks @RAMitchell !

PR #4521, opened 2019-05-31
Fix C++11 config parser

hcho3:
In https://github.com/dmlc/xgboost/pull/4486#issuecomment-497509172, we found out that #4478 broke the CLI parser when the training data path contained a `#` letter. That is, the new C++11 parser could not differentiate between
```
data = train.libsvm    # cache
```
and
```
data = "train.libsvm#cache"
```

I did an overhaul of parser logic (with `std::regex`!) to fix this issue and others. I also wrote bunch of tests.

**Why not just revert #4478?**  I like the idea of using C++11. Also, the old parser silently ignored backslashes (`\`), which breaks Windows paths.

cc @sriramch

TODO. Update the CLI doc to describe the grammar of configuration file. See some examples now in `tests/cpp/common/test_config.cc`.

### DIFF
--- a/src/common/config.h
+++ b/src/common/config.h
@@ -30,10 +30,10 @@ class ConfigParser {
    */
   explicit ConfigParser(const std::string& path)
     : line_comment_regex_("^#"),
-      key_regex_("^([^#\"'=\\r\\n\\t ]+)[\\t ]*="),
-      key_regex_escaped_("^([\"'])([^\"'=\\r\\n]+)\\1[\\t ]*="),
-      value_regex_("^([^#\"'=\\r\\n\\t ]+)[\\t ]*(?:#.*){0,1}$"),
-      value_regex_escaped_("^([\"'])([^\"'=\\r\\n]+)\\1[\\t ]*(?:#.*){0,1}$") {
+      key_regex_(R"rx(^([^#"'=\r\n\t ]+)[\t ]*=)rx"),
+      key_regex_escaped_(R"rx(^(["'])([^"'=\r\n]+)\1[\t ]*=)rx"),
+      value_regex_(R"rx(^([^#"'=\r\n\t ]+)[\t ]*(?:#.*){0,1}$)rx"),
+      value_regex_escaped_(R"rx(^(["'])([^"'=\r\n]+)\1[\t ]*(?:#.*){0,1}$)rx") {
     NormalizeEOL(path);
     fi_.open(path);
     CHECK(!fi_.fail()) << "Cannot open file " << path;
--- a/src/common/config.h
+++ b/src/common/config.h
@@ -7,11 +7,12 @@
 #ifndef XGBOOST_COMMON_CONFIG_H_
 #define XGBOOST_COMMON_CONFIG_H_
 
-#include <dmlc/logging.h>
+#include <xgboost/logging.h>
 #include <cstdio>
 #include <string>
-#include <istream>
 #include <fstream>
+#include <istream>
+#include <sstream>
 #include <vector>
 #include <regex>
 #include <iterator>
@@ -33,10 +34,35 @@ class ConfigParser {
       key_regex_(R"rx(^([^#"'=\r\n\t ]+)[\t ]*=)rx"),
       key_regex_escaped_(R"rx(^(["'])([^"'=\r\n]+)\1[\t ]*=)rx"),
       value_regex_(R"rx(^([^#"'=\r\n\t ]+)[\t ]*(?:#.*){0,1}$)rx"),
-      value_regex_escaped_(R"rx(^(["'])([^"'=\r\n]+)\1[\t ]*(?:#.*){0,1}$)rx") {
-    NormalizeEOL(path);
-    fi_.open(path);
-    CHECK(!fi_.fail()) << "Cannot open file " << path;
+      value_regex_escaped_(R"rx(^(["'])([^"'=\r\n]+)\1[\t ]*(?:#.*){0,1}$)rx"),
+      path_ {path} {}
+
+  std::string LoadConfigFile(const std::string& path) {
+    std::ifstream fin(path, std::ios_base::in | std::ios_base::binary);
+    CHECK(fin) << "Failed to open: " << path;
+    std::string content{std::istreambuf_iterator<char>(fin),
+                        std::istreambuf_iterator<char>()};
+    return content;
+  }
+
+  /*!
+   * \brief Normalize end-of-line in a file so that it uses LF for all line
+   *        endings.
+   * This is needed because some OSes use CR or CR LF instead.
+   * So we replace all CR with LF. Replacement will be performed in-place.
+   * \param p_config_str pointer to configuration
+   */
+  std::string NormalizeConfigEOL(std::string const& config_str) {
+    std::string result;
+    std::stringstream ss(config_str);
+    for (size_t i = 0; i < config_str.size(); ++i) {
+      if (config_str[i] == '\r') {
+        result.push_back('\n');
+        continue;
+      }
+      result.push_back(config_str[i]);
+    }
+    return result;
   }
 
   /*!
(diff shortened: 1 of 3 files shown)


src/common/config.h
```diff
@@ -2,84 +2,161 @@
...
+      // Key has a whitespace and/or #; it has to be wrapped around a pair of
+      // single or double quotes. Example: "foo bar"  'foo#bar'
+      CHECK_EQ(m.size(), 3);
+      *key = m[2].str();
+    } else {
+      LOG(FATAL) << "This line is not a valid key-value pair: " << str;
+    }
+
+    /* Match value */
+    buf.erase(buf.begin(), m[0].second);
```
hcho3: This line fails to compile on GCC 5.x but works fine on GCC 7.x. Investigating

src/common/config.h
```diff
@@ -2,84 +2,161 @@
...
-   * \param cfgFileName name of configure file
+   * \brief Constructor for INI-style configuration parser
+   * \param path path to configuration file
    */
-  explicit ConfigParse(const std::string &cfgFileName) {
-    fi_.open(cfgFileName);
-    if (fi_.fail()) {
-      LOG(FATAL) << "cannot open file " << cfgFileName;
-    }
+  explicit ConfigParser(const std::string& path)
```
sriramch: please document the grammar. it is a bit difficult to grok all this without poring over this in greater detail.

src/common/config.h
```diff
@@ -2,84 +2,161 @@
...
+    : line_comment_regex_("^#"),
```
sriramch: what about lines that do not start with '#' but a sequence of spaces followed by '#' - would that be treated as comments as well? if not, should it be?

how about comments at the end of a config item - a la: `key = value # this is a value of key`

i'm not suggesting that we support new syntax - just making sure that we do not regress on the existing behavior, if it indeed is supported today.

tests/cpp/common/test_config.cc
```diff
@@ -0,0 +1,163 @@
...
+  // Note: '#' in path won't be accepted correctly unless the whole path is
+  // wrapped with quotes. This is important for external memory.
+  ASSERT_TRUE(parser.ParseKeyValuePair("data = dmatrix.libsvm#dtrain.cache",
+                                       &key, &value));
+  ASSERT_EQ(key, "data");
+  ASSERT_EQ(value, "dmatrix.libsvm");  // cache was silently ignored
+
+  // 5. Wrapping key/value with quotes
+  // Any key or value containing '#' needs to be wrapped with quotes
+  ASSERT_TRUE(parser.ParseKeyValuePair("data = \"dmatrix.libsvm#dtrain.cache\"",
```
sriramch: try some with absolute paths as well `data = file#/absolute/path/to/cache/file`

src/common/config.h
```diff
@@ -2,84 +2,161 @@
...
+      key_regex_(R"rx(^([^#"'=\r\n\t ]+)[\t ]*=)rx"),
+      key_regex_escaped_(R"rx(^(["'])([^"'=\r\n]+)\1[\t ]*=)rx"),
+      value_regex_(R"rx(^([^#"'=\r\n\t ]+)[\t ]*(?:#.*){0,1}$)rx"),
```
sriramch: the value payload may also contain token `/` `-` etc - `hdfs://` `s3://` `.fmt-` etc. they may need to be handled as well

src/common/config.h
```diff
@@ -2,84 +2,161 @@
...
+  static void NormalizeEOL(const std::string& path) {
+    std::FILE* fp = std::fopen(path.c_str(), "rb+");
+    CHECK(fp) << "Cannot open file " << path;
+    int ch;
+    while ((ch = std::fgetc(fp)) != EOF) {
+      if (ch == static_cast<int>('\r')) {
+        CHECK_EQ(std::fseek(fp, -1, SEEK_CUR), 0)
+          << "Could not normalize line ending in file " << path;
+        std::fputc(static_cast<int>('\n'), fp);
+        CHECK_EQ(std::fseek(fp, 0, SEEK_CUR), 0)
```
sriramch: what is the purpose of this fseek? doesn't the previous fputc advance the file pointer anyways?

src/common/config.h
```diff
@@ -2,84 +2,161 @@
...
+   * This is needed because some OSes use CR or CR LF instead.
+   * So we replace all CR with LF. Replacement will be performed in-place.
+   * \param path path to configuration file
+   */
+  static void NormalizeEOL(const std::string& path) {
+    std::FILE* fp = std::fopen(path.c_str(), "rb+");
+    CHECK(fp) << "Cannot open file " << path;
+    int ch;
+    while ((ch = std::fgetc(fp)) != EOF) {
+      if (ch == static_cast<int>('\r')) {
```
sriramch: would this be corrupting user-data, if 'cr' happens to be in it? perhaps, a pathological scenario. is it safe to see if this happens only at the end of line?

src/common/config.h
hcho3: > what about lines that do not start with '#' but a sequence of spaces followed by '#' - would that be treated as comments as well?

Yes.

> how about comments at the end of a config item

Yes, it would be supported.

src/common/config.h
hcho3: I will go ahead and write a doc for the grammar.

src/common/config.h
hcho3: These letters are fine, because in the value, all letters except `#`, `"`, `'`, `=`, `\r`, `\n`, `\t`, and ` ` (space) are allowed. (`#`, `\t`, and ` ` may be inside the value if the value is quoted)

src/common/config.h
hcho3: This line is supposed to be a no-op. We need `std::fseek` here because the C standard requires it when `fgetc()` and `fputc()` are interleaved ([link](https://en.cppreference.com/w/c/io/fopen)):
> In update mode ('+'), both input and output may be performed, but output cannot be followed by input without an intervening call to fflush, fseek, fsetpos or rewind, and input cannot be followed by output without an intervening call to fseek, fsetpos or rewind, unless the input operation encountered end of file.

src/common/config.h
hcho3: In the old parser, `\r` was assumed to be a newline character, so we are preserving its behavior.

src/common/config.h
hcho3: See an example of a comment at the end of an item: https://github.com/dmlc/xgboost/pull/4521/files/f376752b4a128519ad12246f50d162eaca0c0b22#diff-0c20fe270f3630af98c7b4428807af8cR110

src/common/config.h
hcho3: And if we assume that the text file is encoded in UTF-8, then `\r` will never show up as part of a multi-type character.

src/common/config.h
sriramch: please document it. it is not quite intuitive when you read it the first time.

also, an fflush may be a better option here - at least, it conveys the impression that the streams are sync'd after a write for the readers to see the new data. an fseek from the current position does not quite convey the intent.

trivialfis: I'm not an expert in file IO.  But is it possible to load the config file in one go then modify its content like:
```
  std::string content;
  fin.seekg(0, std::ios::end);
  content.resize(fin.tellg());
  fin.seekg(0, std::ios::beg);
  fin.read(&content[0], content.size());
  fin.close();
```

trivialfis:
@hcho3 Could you review my commit?  It prevents creating side effect, which might cause surprising behaviours on the user side.

trivialfis:
@hcho3 Hi I'm curious if you have some experience of encounter failed line reading on OSX/Windows.  From this [SO](https://stackoverflow.com/questions/7372012/line-feed-and-carriage-return), it seems there's no need for handling carriage.  Also the currently failed test on Jenkins suggests so.

trivialfis:
@sriramch @hcho3  Thanks.

PR #4526, opened 2019-05-31
- training with external memory - part 2 of 2

sriramch:
- when external memory support is enabled, building of histogram indices are
     done incrementally for every sparse page
   - the entire set of input data is divided across multiple gpu's and the relative
     row positions within each device is tracked when building the compressed histogram buffer
   - the external memory is used for building the compressed buffer *only*. it is required for the compressed buffer to fit entirely into the gpu memory
   - this was tested using a mortgage dataset containing ~ 670m rows before 4xt4's could be
     saturated

@RAMitchell @canonizer @rongou - please review

### DIFF
--- a/src/tree/updater_gpu_hist.cu
+++ b/src/tree/updater_gpu_hist.cu
@@ -627,10 +627,10 @@ struct DeviceShard;
 // entire dataset across multiple sparse page batches. This keeps track of the number
 // of rows to process from a batch and the position from which to process on each device.
 struct RowStateOnDevice {
-  const size_t n_rows; // Number of rows assigned to this device
-  size_t n_rows_processed; // Number of rows processed thus far
-  size_t batch_n_rows; // Number of rows to process from the current sparse page batch
-  size_t row_offset; // Offset from the current sparse page batch to begin processing
+  const size_t n_rows;  // Number of rows assigned to this device
+  size_t n_rows_processed;  // Number of rows processed thus far
+  size_t batch_n_rows;  // Number of rows to process from the current sparse page batch
+  size_t row_offset;  // Offset from the current sparse page batch to begin processing
 
   explicit RowStateOnDevice(size_t nrows)
     : n_rows(nrows), n_rows_processed(0), batch_n_rows(0), row_offset(0) {
--- a/src/tree/updater_gpu_hist.cu
+++ b/src/tree/updater_gpu_hist.cu
@@ -623,6 +623,33 @@ __forceinline__ __device__ void CountLeft(int64_t* d_count, int val,
 template <typename GradientSumT>
 struct DeviceShard;
 
+// Instances of this type are created while creating the histogram bins for the
+// entire dataset across multiple sparse page batches. This keeps track of the number
+// of rows to process from a batch and the position from which to process on each device.
+struct RowStateOnDevice {
+  const size_t n_rows_; // Number of rows assigned to this device
+  size_t n_rows_processed_; // Number of rows processed thus far
+  size_t batch_n_rows_; // Number of rows to process from the current sparse page batch
+  size_t row_offset_; // Offset from the current sparse page batch to begin processing
+
+  RowStateOnDevice(size_t nrows)
+    : n_rows_(nrows), n_rows_processed_(0), batch_n_rows_(0), row_offset_(0) {
+  }
+
+  // Advance the row state by the number of rows processed
+  void Advance() {
+    n_rows_processed_ += batch_n_rows_;
+    CHECK_LE(n_rows_processed_, n_rows_);
+    batch_n_rows_ = row_offset_ = 0;
+  }
+
+  static RowStateOnDevice CreateRowStateOnDevice(size_t total_rows, size_t batch_rows) {
+    RowStateOnDevice rstate(total_rows);
+    rstate.batch_n_rows_ = batch_rows;
+    return rstate;
+  }
+};
+
 template <typename GradientSumT>
 struct GPUHistBuilderBase {
  public:
@@ -666,8 +693,6 @@ struct DeviceShard {
   /*! \brief Sum gradient for each node. */
   std::vector<GradientPair> node_sum_gradients;
   common::Span<GradientPair> node_sum_gradients_d;
-  /*! \brief row offset in SparsePage (the input data). */
-  thrust::device_vector<size_t> row_ptrs;
   /*! \brief On-device feature set, only actually used on one of the devices */
   thrust::device_vector<int> feature_set_d;
   thrust::device_vector<int64_t>
@@ -696,7 +721,6 @@ struct DeviceShard {
                           std::function<bool(ExpandEntry, ExpandEntry)>>;
   std::unique_ptr<ExpandQueue> qexpand;
 
-  // TODO(canonizer): do add support multi-batch DMatrix here
   DeviceShard(int _device_id, int shard_idx, bst_uint row_begin,
               bst_uint row_end, TrainParam _param, uint32_t column_sampler_seed)
       : device_id(_device_id),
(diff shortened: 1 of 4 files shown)


hcho3:
@sriramch Let me put up with a guide to run formatting tests locally.

sriramch:
>> https://github.com/dmlc/xgboost/pull/4526#issuecomment-497839137

@hcho3 - thank you! i also feel that some of these rules are overly restrictive. is there a way to comment on it? further, in our earlier projects, we used to run clang-format (either manually or enforce it via jenkins) that keeps the styling consistent. we typically check-in the rules in the repo and simply run that tool that enforces it. this can save multiple round trips with these pushes.

hcho3:
@sriramch It turns out that clang-tidy supports automatically fixing some errors. So we should review style constraints and only keep those that can be applied automatically.

cc @RAMitchell @trivialfis

trivialfis:
@hcho3 Due to the fact that there are some places in thrust that trick clang, also we need to use a different libstd to avoid some bugs.  Auto fix might not be reliable enough for now.

RAMitchell:
Agree, the auto fix is not reliable with cuda. Documentation on how to locally run formatting lint checks would help a lot. This could be difficult to dependence on specific llvm versions.

src/tree/updater_gpu_hist.cu
```diff
@@ -1328,23 +1329,47 @@ inline void DeviceShard<GradientSumT>::InitCompressedData(
...
+  if (!device_row_state.batch_n_rows) return;
+
+  size_t row_stride = this->ellpack_matrix.row_stride;
+
+  const auto &offset_vec = row_batch.offset.ConstHostVector();
+  /*! \brief row offset in SparsePage (the input data). */
+  thrust::device_vector<size_t> row_ptrs(device_row_state.batch_n_rows+1);
+  thrust::copy(
+    offset_vec.data() + device_row_state.row_offset,
+    offset_vec.data() + device_row_state.row_offset + device_row_state.batch_n_rows + 1,
```
RAMitchell: I would put a check here to make sure we don't read off the end of offset_vec.

src/tree/updater_gpu_hist.cu
```diff
@@ -1328,23 +1329,47 @@ inline void DeviceShard<GradientSumT>::InitCompressedData(
...
+    row_ptrs.begin());
+
   int num_symbols = n_bins + 1;
   // bin and compress entries in batches of rows
-  size_t gpu_batch_nrows =
+  size_t gpu_batch_nrows = 0;
+  if (rows_per_batch == 0) {
```
RAMitchell: Let's have some comments to explain these three conditions.

src/tree/updater_gpu_hist.cu
```diff
@@ -1355,17 +1380,20 @@ inline void DeviceShard<GradientSumT>::CreateHistIndices(
...
     CompressBinEllpackKernel<<<grid3, block3>>>
         (common::CompressedBufferWriter(num_symbols),
          gidx_buffer.data(),
          row_ptrs.data().get() + batch_row_begin,
          entries_d.data().get(),
-         gidx_fvalue_map.data(), feature_segments.data(),
-         batch_row_begin, batch_nrows,
+         gidx_fvalue_map.data(),
+         feature_segments.data(),
+         device_row_state.n_rows_processed + batch_row_begin,
```
RAMitchell: There are too many similarly named to variables here. e.g. device_row_state.batch_n_rows, batch_nrows. I want to be very clear what each of these variables is doing through descriptive naming and comments. I don't mind if the names are long, it's more important to know what is happening.

src/tree/updater_gpu_hist.cu
```diff
@@ -1355,17 +1380,20 @@ inline void DeviceShard<GradientSumT>::CreateHistIndices(
...
+         batch_nrows,
          row_ptrs[batch_row_begin],
-         row_stride, null_gidx_value);
+         row_stride,
+         hmat.row_ptr.back());
```
RAMitchell: Do not pass `hmat.row_ptr.back()` as the argument to null_gidx_value. This does not very clearly explain what is going on.

src/tree/updater_gpu_hist.cu
```diff
@@ -623,6 +623,33 @@ __forceinline__ __device__ void CountLeft(int64_t* d_count, int val,
...
+  }
+
+  // Advance the row state by the number of rows processed
+  void Advance() {
+    n_rows_processed += batch_n_rows;
+    CHECK_LE(n_rows_processed, n_rows);
+    batch_n_rows = row_offset = 0;
+  }
+
+  static RowStateOnDevice CreateRowStateOnDevice(size_t total_rows, size_t batch_rows) {
```
RAMitchell: I don't understand this static method. Why not just use a constructor?

src/tree/updater_gpu_hist.cu
```diff
@@ -1375,6 +1403,58 @@ inline void DeviceShard<GradientSumT>::CreateHistIndices(
   entries_d.shrink_to_fit();
 }
 
+// An instance of this type is created which keeps track of total number of rows to process,
+// rows processed thus far, rows to process and the offset from the current sparse page batch
+// to begin processing on each device
+class DeviceHistogramBuilderState {
```
RAMitchell: I am wondering if there is another way to do this without having this class to manage the process. For example is it possible for each Shard to independently calculate its row state?

src/tree/updater_gpu_hist.cu
sriramch: will do...

src/tree/updater_gpu_hist.cu
sriramch: will do...

src/tree/updater_gpu_hist.cu
sriramch: i'll try and come up with some names.

src/tree/updater_gpu_hist.cu
sriramch: will do...

src/tree/updater_gpu_hist.cu
sriramch: this is added primarily for the tests as a convenience method to create an instance of this type which can then be passed to the histogram index creation method. the type is constructed and its state initialized in 2 phases, and thus it takes only a single argument (as all the information isn't available during type construction). i added this to prevent the clutter in the tests, which may then require multiple steps to initialize and to set its state in the absence of this.

src/tree/updater_gpu_hist.cu
sriramch: i wanted to make this a part of the shard state to begin with, but decided against it. the state that is maintained by this type is transient in nature - meaning, this is needed only for creating the compressed bins during the tree initialization time. this isn't needed later during the tree construction/boosting time. further, the state maintained by this time repeatedly changes for each batch, and isn't even a true reflection of the entire dmatrix. thus, i decoupled this to be a separate type, as it isn't a part of the device shard, rather one used to construct its state.

hcho3:
@sriramch @trivialfis To improve stability of Linux tests, I increased the memory of CPU workers from 16 GB to 32 GB.

sriramch:
> @sriramch @trivialfis To improve stability of Linux tests, I increased the memory of CPU workers from 16 GB to 32 GB.

thanks @hcho3. it is a bit ironic though that these external memory tests are consuming more memory - perhaps something to look at later!

src/tree/updater_gpu_hist.cu
RAMitchell: I'm ok with this.

src/tree/updater_gpu_hist.cu
RAMitchell: If it's only needed in the tests, create a function in the tests. Also if you have to do this, having a second constructor overload is preferred to using a static method.

src/tree/updater_gpu_hist.cu
canonizer: There's a lot of transient state in a shard. I think putting it inside a shard will make the code cleaner.

I'll leave it up to you, however.

src/tree/updater_gpu_hist.cu
canonizer: The reason we have this for quantiles is that the batch size affects the exact quantile values that are found.

There is no such problem here, i.e. the result does not depend on the batch size. Therefore, there's no need for a complicated selection of `rows_per_batch`, the automatically selected value will do, as `row_stride` is already computed based on all batches and shards.

sriramch: done

src/tree/updater_gpu_hist.cu
sriramch: done

PR #4527, opened 2019-06-02
Deprecate gpu_exact, bump required cuda version in docs

RAMitchell:
Document deprecation of the exact GPU algorithm. I am no longer willing to support this algorithm, users should be encouraged to use 'gpu_hist' in all cases.

Reduction in old code will help reduce maintenance burden and refactoring overhead for those of us currently working on GPU algorithms.


trivialfis:
> I am no longer willing to support this algorithm

Neither do I.  But I don't see it needs many maintenance either other than the current `gamma` parameter bug.  I will leave it for you to decide.

PR #4528, opened 2019-06-02
Refactor histogram building code for gpu_hist

RAMitchell:
This PR removes an abstract base class for histogram building and two separate code paths for building histograms with and without shared memory. Instead it fuses the two code paths at the cost of a couple of if statements.

### DIFF
--- a/src/tree/updater_gpu_hist.cu
+++ b/src/tree/updater_gpu_hist.cu
@@ -515,28 +515,38 @@ __global__ void CompressBinEllpackKernel(
 }
 
 template <typename GradientSumT>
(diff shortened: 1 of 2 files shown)


src/tree/updater_gpu_hist.cu
```diff
@@ -515,28 +515,38 @@ __global__ void CompressBinEllpackKernel(
 }
 
 template <typename GradientSumT>
-__global__ void SharedMemHistKernel(ELLPackMatrix matrix, const bst_uint* d_ridx,
+__global__ void SharedMemHistKernel(ELLPackMatrix matrix,
```
trivialfis: Maybe change the name too.

PR #4529, opened 2019-06-02
 Optimizations for CPU

SmirnovEgorRu:
It is this PR https://github.com/dmlc/xgboost/pull/4433. But I have merged all commits in once and applied comment in the review.
My next steps here:
- Make some coding cleanup and add comments where needed
- Attach performance data

### DIFF
--- a/src/tree/param.h
+++ b/src/tree/param.h
@@ -279,7 +279,7 @@ XGBOOST_DEVICE inline T CalcGain(const TrainingParams &p, T sum_grad, T sum_hess
     }
   } else {
     T w = CalcWeight(p, sum_grad, sum_hess);
-    T ret = CalcGainGivenWeight(p, sum_grad, sum_hess, w);
+    T ret = CalcGainGivenWeight<TrainingParams, T>(p, sum_grad, sum_hess, w);
     if (p.reg_alpha == 0.0f) {
       return ret;
     } else {
@@ -299,7 +299,7 @@ template <typename TrainingParams, typename T>
 XGBOOST_DEVICE inline T CalcGain(const TrainingParams &p, T sum_grad, T sum_hess,
                                  T test_grad, T test_hess) {
   T w = CalcWeight(sum_grad, sum_hess);
-  T ret = CalcGainGivenWeight(p, test_grad, test_hess);
+  T ret = CalcGainGivenWeight<TrainingParams, T>(p, test_grad, test_hess);
   if (p.reg_alpha == 0.0f) {
     return ret;
   } else {
@@ -338,15 +338,16 @@ XGBOOST_DEVICE inline float CalcWeight(const TrainingParams &p, GpairT sum_grad)
 }
 
 /*! \brief core statistics used for tree construction */
-struct XGBOOST_ALIGNAS(16) GradStats {
+struct  GradStats {
+  typedef double GradType;
   /*! \brief sum gradient statistics */
-  double sum_grad;
+  GradType sum_grad;
   /*! \brief sum hessian statistics */
-  double sum_hess;
+  GradType sum_hess;
 
  public:
-  XGBOOST_DEVICE double GetGrad() const { return sum_grad; }
-  XGBOOST_DEVICE double GetHess() const { return sum_hess; }
+  XGBOOST_DEVICE GradType GetGrad() const { return sum_grad; }
+  XGBOOST_DEVICE GradType GetHess() const { return sum_hess; }
 
   XGBOOST_DEVICE GradStats() : sum_grad{0}, sum_hess{0} {
     static_assert(sizeof(GradStats) == 16,
@@ -356,7 +357,7 @@ struct XGBOOST_ALIGNAS(16) GradStats {
   template <typename GpairT>
   XGBOOST_DEVICE explicit GradStats(const GpairT &sum)
       : sum_grad(sum.GetGrad()), sum_hess(sum.GetHess()) {}
-  explicit GradStats(const double grad, const double hess)
+  explicit GradStats(const GradType grad, const GradType hess)
       : sum_grad(grad), sum_hess(hess) {}
   /*!
    * \brief accumulate statistics
@@ -381,7 +382,7 @@ struct XGBOOST_ALIGNAS(16) GradStats {
   /*! \return whether the statistics is not used yet */
   inline bool Empty() const { return sum_hess == 0.0; }
   /*! \brief add statistics to the data */
-  inline void Add(double grad, double hess) {
+  inline void Add(GradType grad, GradType hess) {
     sum_grad += grad;
     sum_hess += hess;
   }
@@ -411,7 +412,7 @@ struct ValueConstraint {
 
   template <typename ParamT>
   XGBOOST_DEVICE inline double CalcGain(const ParamT &param, GradStats stats) const {
-    return CalcGainGivenWeight(param, stats.sum_grad, stats.sum_hess,
+    return CalcGainGivenWeight<ParamT, float>(param, stats.sum_grad, stats.sum_hess,
                                CalcWeight(param, stats));
   }
 
@@ -422,8 +423,8 @@ struct ValueConstraint {
     double wleft = CalcWeight(param, left);
     double wright = CalcWeight(param, right);
     double gain =
-        CalcGainGivenWeight(param, left.sum_grad, left.sum_hess, wleft) +
-        CalcGainGivenWeight(param, right.sum_grad, right.sum_hess, wright);
+        CalcGainGivenWeight<ParamT, float>(param, left.sum_grad, left.sum_hess, wleft) +
+        CalcGainGivenWeight<ParamT, float>(param, right.sum_grad, right.sum_hess, wright);
     if (constraint == 0) {
       return gain;
     } else if (constraint > 0) {
@@ -468,6 +469,7 @@ struct SplitEntry {
   bst_float split_value{0.0f};
   GradStats left_sum;
   GradStats right_sum;
+  bool default_left{true};
 
   /*! \brief constructor */
   SplitEntry()  = default;
@@ -482,7 +484,11 @@ struct SplitEntry {
    * \param split_index the feature index where the split is on
    */
   inline bool NeedReplace(bst_float new_loss_chg, unsigned split_index) const {
-    if (this->SplitIndex() <= split_index) {
+    if (!std::isfinite(new_loss_chg)) {  // in some cases new_loss_chg can be NaN or Inf,
+                                         // for example when lambda = 0 & min_child_weight = 0
+                                         // skip value in this case
+      return false;
+    } else if (this->SplitIndex() <= split_index) {
       return new_loss_chg > this->loss_chg;
     } else {
       return !(this->loss_chg > new_loss_chg);
@@ -500,6 +506,7 @@ struct SplitEntry {
       this->split_value = e.split_value;
       this->left_sum = e.left_sum;
       this->right_sum = e.right_sum;
+      this->default_left = e.default_left;
       return true;
     } else {
       return false;
@@ -514,13 +521,11 @@ struct SplitEntry {
    * \return whether the proposed split is better and can replace current split
    */
   inline bool Update(bst_float new_loss_chg, unsigned split_index,
-                     bst_float new_split_value, bool default_left,
+                     bst_float new_split_value, bool new_default_left,
                      const GradStats &left_sum, const GradStats &right_sum) {
     if (this->NeedReplace(new_loss_chg, split_index)) {
       this->loss_chg = new_loss_chg;
-      if (default_left) {
-        split_index |= (1U << 31);
-      }
+      this->default_left = new_default_left;
       this->sindex = split_index;
       this->split_value = new_split_value;
       this->left_sum = left_sum;
@@ -536,9 +541,9 @@ struct SplitEntry {
     dst.Update(src);
   }
   /*!\return feature index to split on */
-  inline unsigned SplitIndex() const { return sindex & ((1U << 31) - 1U); }
+  inline unsigned SplitIndex() const { return sindex; }
   /*!\return whether missing value goes to left branch */
-  inline bool DefaultLeft() const { return (sindex >> 31) != 0; }
+  inline bool DefaultLeft() const { return default_left; }
 };
 
 }  // namespace tree
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -566,7 +566,7 @@ void QuantileHistMaker::Builder::BuildHistsBatch(const std::vector<ExpandEntry>&
     const int32_t node_idx = task_node_idx[itask];
 
     common::GradStatHist::GradType* data_local_hist;
-    common::GradStatHist* grad_stat;
+    common::GradStatHist* grad_stat;  // total gradient/hessian value for node `nid`
     std::tie(data_local_hist, grad_stat) = GetHistBuffer(&(*hist_is_init)[node_idx],
         &grad_stats[node_idx], block_id, nthread, tid,
         &(*hist_buffers)[node_idx], hist_size);
@@ -1082,9 +1082,10 @@ void QuantileHistMaker::Builder::EvaluateSplitsBatch(
   // parallel enumeration
   #pragma omp parallel for schedule(guided)
   for (int32_t i = 0; i < tasks.size(); ++i) {
+    // node_idx : offset within `nodes` list
     const int32_t  node_idx    = tasks[i].first;
     const size_t   fid         = tasks[i].second;
-    const int32_t  nid         = nodes[node_idx].nid;
+    const int32_t  nid         = nodes[node_idx].nid;  // usually node_idx != nid
     const int32_t  sibling_nid = nodes[node_idx].sibling_nid;
     const int32_t  parent_nid  = nodes[node_idx].parent_nid;
 
@@ -1110,6 +1111,9 @@ void QuantileHistMaker::Builder::EvaluateSplitsBatch(
       const bool compute_backward = this->EnumerateSplit(+1, gmat, hist_[nid], snode,
           info, &splits[i].first, fid, nid);
 
+      // Sometimes, we don't need to enumerate backward because forward and backward
+      // enumeration will give same loss values. This is the case if the particular feature 
+      // column contains no missing values. So enumerate backward only if it's necessary.
       if (compute_backward) {
         this->EnumerateSplit(-1, gmat, hist_[nid], snode, info,
             &splits[i].first, fid, nid);
@@ -1161,6 +1165,10 @@ void QuantileHistMaker::Builder::InitNewNode(int nid,
 }
 
 // enumerate the split values of specific feature
+// d_step: +1 or -1, indicating direction at which we scan candidate thresholds in order
+// fid: feature for which we seek to pick best threshold
+// Returns false if we don't need to enumerate in opposite direction.
+//   This is the case if the particular feature (fid) column contains no missing values.
 bool QuantileHistMaker::Builder::EnumerateSplit(int d_step,
                                                 const GHistIndexMatrix& gmat,
                                                 const GHistRow& hist,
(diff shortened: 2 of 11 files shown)


SmirnovEgorRu:
Hi @chenqin @trivialfis ,
I have done huge work on improvement of code quality like adding checkers, comments, splitting code to separate smaller functions and other code cleaning.
Can I count on a review soon?

Current performance results for training stage are following:
depth-wise mode:

Data Set | This PR | XGB 0.9
-- | -- | --
MSLR (3017514 x 137, 5 classes), s | 62.9 | 267.3
HIGGS1m (1m x 28, 2 classes), s | 4.8 | 13.0

lossguide mode:

Data Set | This PR | XGB 0.9
-- | -- | --
MSLR (3017514 x 137, 5 classes), s | 183.1 | 274.6
HIGGS1m (1m x 28, 2 classes), s | 6.1 | 13.6

HW: Intel(R) Xeon(R) Gold 5120 CPU @ 2.2GHz, 14 cores per socket, 2 sockets, HT: on

Also, could you, please, run code coverage job? My expectation - code coverage will be similar to which was before. Because mostly I didn't add new functionality, but just changed implementation of some functions and existing tests should be enough.

SmirnovEgorRu:
Now I see following performance gaps:
- Time for hist computation was reduced well and now computation of new gradients/hessians on each iteration spends significant time. Here are many opportunities to use AVX/AVX512 jointly with mask instructions.
- Prediction between iterations, especially for cases when we train a few trees on a iteration (dont use caches in this cases). In this worst case I see that prediction takes similar time as training. I dont see reason why we cant use cache for prediction, potentially it can be added.
- Some pre-processing. I hope it will addressed in this pr https://github.com/dmlc/xgboost/pull/4501
- Also, for large data sets usage of int8/int16 instead of int32 for bin-matrix improves performance significantly. In some cases more than in 2x time.

trivialfis:
@SmirnovEgorRu Thanks for the huge effort.  Will review soon.

chenqin:
Thanks for huge effort! 

A few questions in general

- proactive prefetch data into all level of cache can be great help speeding up. why T0(all level)? L3 is abundant enough I assume.

- given compiler also do optimization, do we plan to test turn on flags and benchmark against https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

> * Time for hist computation was reduced well and now computation of new gradients/hessians on each iteration spends significant time. Here are many opportunities to use AVX/AVX512 jointly with mask instructions

SmirnovEgorRu:
@chenqin Hi!
I prefetch data which will be used very soon (after reading of the next 10 rows of bin-matrix), so better to prefetch this to L1. Also this data can be already in L3 and we need to fetch this to L1 to reduce latency.

> given compiler also do optimization, do we plan to test turn on flags and benchmark against

Here I mean - vector instruction sets mostly. Now XGBoost uses sse2 only, but for gradient/hessians computations (as an example) AVX512 can be used as well:
- Some of objective functions contain many code branches. They can be replaced by vector masks instructions from AVX512. From time to time it provides 10x speedup or even more.
- Usage of wider vector registers also should help.

One problem here - that compiler in some cases can't auto-vectorize your code, instrinsic-functions help a lot here. Also, if we want to support different instructions sets (for example AVX512 and sse2) in one library-file - we need to have separate paths of code for them. It can be done, but requires some changes in build system.

I think, if topic of gradient boosting optimizations are interested - we can write some article for web site (as it was done for GPU). From my side I can provide materials about this topic. @chenqin, @hcho3 What do you think?

SmirnovEgorRu:
Hi @hcho3, 
Do you have any estimations when you will be free to review this PR? :)

hcho3:
Reviewing now

trivialfis:
@hcho3 I'm only approving the PR because the tests passed, and I think optimizing histogram build will be beneficial to future improvement.  But if we merge this someone needs to bring a huge refactor to quantile_hist and clarify how all these are put together.

Personally I prefer optimize this by thinking about the whole structure.  And I think there's lots of opportunities in optimization if we look at how the whole procedure is run, I mean how many of the steps are necessary if we do some preprocessing to the data like calculating the number of non-zero columns, implement a sparse histogram you mentioned, if it's possible to do grouping block before constructing quantile cuts etc.  Going instruction level optimization is cool, but to me quantile hist has a lot more room to be improved on other places and vectorization might not be the critical point.  I don't have the time to do it myself right now, it just my opinions.

src/tree/updater_quantile_hist.cc
```diff
@@ -184,23 +324,366 @@ void QuantileHistMaker::Builder::EvaluateSplits(
...
+  // 2. Build partial histograms for each node
+  #pragma omp parallel for schedule(guided)
+  for (int32_t itask = 0; itask < n_hist_buidling_tasks; ++itask) {
+    const size_t tid = omp_get_thread_num();
+    const int32_t nid = task_nid[itask];
+    const int32_t block_id = task_block_idx[itask];
+    const int32_t node_idx = task_node_idx[itask];
+
+    common::GradStatHist::GradType* data_local_hist;
+    common::GradStatHist* grad_stat;
```
```suggestion
    common::GradStatHist* grad_stat;  // total gradient/hessian value for node `nid`
```

src/tree/updater_quantile_hist.cc
```diff
@@ -184,23 +324,366 @@ void QuantileHistMaker::Builder::EvaluateSplits(
...
+  const GradientPair::ValueT* const pgh =
+      reinterpret_cast<const GradientPair::ValueT*>(gpair.data());
+
+  // 2. Build partial histograms for each node
+  #pragma omp parallel for schedule(guided)
+  for (int32_t itask = 0; itask < n_hist_buidling_tasks; ++itask) {
+    const size_t tid = omp_get_thread_num();
+    const int32_t nid = task_nid[itask];
+    const int32_t block_id = task_block_idx[itask];
+    const int32_t node_idx = task_node_idx[itask];
```
```suggestion
    // node_idx : location of node `nid` within the `nodes` list. In general, node_idx != nid
    const int32_t node_idx = task_node_idx[itask];
```

src/tree/updater_quantile_hist.cc
```diff
@@ -88,94 +89,233 @@ bool QuantileHistMaker::UpdatePredictionCache(
...
+    int32_t nid) {
+
+  // add constraints
+  if (!(*p_tree)[nid].IsLeftChild() && !(*p_tree)[nid].IsRoot()) {
+    auto parent_id = (*p_tree)[nid].Parent();
+    // it's a right child
+    auto left_sibling_id = (*p_tree)[parent_id].LeftChild();
+    auto parent_split_feature_id = snode_[parent_id].best.SplitIndex();
+
+    {
```
hcho3: I don't think we need an extra layer of scoping here. Can you put `spliteval_->AddSplit(*)` in the same level as the rest of code?

src/tree/updater_quantile_hist.cc
```diff
@@ -549,290 +1055,113 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
-  } else {
-    ApplySplitSparseData(rowset, gmat, &row_split_tloc_, column, lower_bound,
-                         upper_bound, split_cond, default_left);
-  }
+  // partial results
+  std::vector<std::pair<SplitEntry, SplitEntry>> splits(tasks.size());
+  // parallel enumeration
+  #pragma omp parallel for schedule(guided)
+  for (int32_t i = 0; i < tasks.size(); ++i) {
+    const int32_t  node_idx    = tasks[i].first;
```
```suggestion
    // node_idx : offset within `nodes` list
    const int32_t  node_idx    = tasks[i].first;
```

src/tree/updater_quantile_hist.cc
```diff
@@ -549,290 +1055,113 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
+    const size_t   fid         = tasks[i].second;
+    const int32_t  nid         = nodes[node_idx].nid;
```
```suggestion
    const int32_t  nid         = nodes[node_idx].nid;  // usually node_idx != nid
```

src/tree/updater_quantile_hist.cc
```diff
@@ -871,27 +1200,45 @@ void QuantileHistMaker::Builder::EnumerateSplit(int d_step,
...
+        if (c.sum_hess >= param_.min_child_weight) {
+          bst_float loss_chg;
+          bst_float split_pt;
+          {
+            loss_chg = static_cast<bst_float>(spliteval_->ComputeSplitScore(nodeID,
+              fid, e, c) - snode.root_gain);
+          }
+
           split_pt = cut_val[i];
           best.Update(loss_chg, fid, split_pt, d_step == -1, e, c);
```
```suggestion
           best.Update(loss_chg, fid, split_pt, false, e, c);
```
since `d_step` is known to be 1 in this scope

src/tree/updater_quantile_hist.cc
```diff
@@ -902,8 +1249,14 @@ void QuantileHistMaker::Builder::EnumerateSplit(int d_step,
         }
```
```cpp
best.Update(loss_chg, fid, split_pt, true, c, e);
```
since `d_step` is known to be -1 in this scope.

src/tree/updater_quantile_hist.cc
```diff
@@ -871,27 +1200,45 @@ void QuantileHistMaker::Builder::EnumerateSplit(int d_step,
...
+  if (d_step == 1) {
+    for (int32_t i = ibegin; i < iend; i++) {
+      e.Add(hist[i].GetGrad(), hist[i].GetHess());
+      if (e.sum_hess >= param_.min_child_weight) {
+        c.SetSubstract(snode.stats, e);
+        if (c.sum_hess >= param_.min_child_weight) {
+          bst_float loss_chg;
+          bst_float split_pt;
+          {
+            loss_chg = static_cast<bst_float>(spliteval_->ComputeSplitScore(nodeID,
```
hcho3: I don't think we need an extra layer of scoping here.

src/tree/updater_quantile_hist.cc
```diff
@@ -871,27 +1200,45 @@ void QuantileHistMaker::Builder::EnumerateSplit(int d_step,
...
+        if (c.sum_hess >= param_.min_child_weight) {
+          bst_float loss_chg;
+          bst_float split_pt;
+
           // backward enumeration: split at left bound of each bin
-          loss_chg = static_cast<bst_float>(
-              spliteval_->ComputeSplitScore(nodeID, fid, c, e) -
-              snode.root_gain);
+          {
+            loss_chg = static_cast<bst_float>(
```
hcho3: I don't think we need an extra layer of scoping here.

src/tree/updater_quantile_hist.cc
```diff
@@ -549,290 +1055,113 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
+    snode->root_gain = static_cast<float>(
+      spliteval_->ComputeScore(parentid, snode->stats,
+      snode->weight));
   }
-  builder_monitor_.Stop("InitNewNode");
 }
 
 // enumerate the split values of specific feature
-void QuantileHistMaker::Builder::EnumerateSplit(int d_step,
+bool QuantileHistMaker::Builder::EnumerateSplit(int d_step,
```
```suggestion
// d_step: +1 or -1, indicating direction at which we scan candidate thresholds in order
// fid: feature for which we seek to pick best threshold
// Returns false if we don't need to enumerate in opposite direction.
//   This is the case if the particular feature (fid) column contains no missing values.
bool QuantileHistMaker::Builder::EnumerateSplit(int d_step,
```

src/tree/updater_quantile_hist.cc
```diff
@@ -549,290 +1055,113 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
-        } else {
-          right.push_back(rid[k]);
-        }
-      } else {
-        if (static_cast<int32_t>(rbin[k] + column.GetBaseIdx()) <= split_cond) {
-          left.push_back(rid[k]);
-        } else {
-          right.push_back(rid[k]);
-        }
+      if (compute_backward) {
```
```suggestion
      // Sometimes, we don't need to enumerate backward because forward and backward
      // enumeration will give same loss values. This is the case if the particular feature 
      // column contains no missing values. So enumerate backward only if it's necessary.
      if (compute_backward) {
```

src/tree/updater_quantile_hist.cc
```diff
@@ -549,290 +1055,113 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
-                        default_left);
-  } else {
-    ApplySplitSparseData(rowset, gmat, &row_split_tloc_, column, lower_bound,
-                         upper_bound, split_cond, default_left);
-  }
+  // partial results
+  std::vector<std::pair<SplitEntry, SplitEntry>> splits(tasks.size());
+  // parallel enumeration
+  #pragma omp parallel for schedule(guided)
+  for (int32_t i = 0; i < tasks.size(); ++i) {
```
hcho3: I really like how we are evaluating both node `nid` and its sibling in the same thread. Nicely done!

src/tree/updater_quantile_hist.cc
```diff
@@ -88,94 +89,233 @@ bool QuantileHistMaker::UpdatePredictionCache(
...
+    BuildNodeStat(gmat, p_fmat, p_tree, gpair_h, nid);
+    if (sibling_nid > -1) {
+      BuildNodeStat(gmat, p_fmat, p_tree, gpair_h, sibling_nid);
+    }
+  }
+  perf_monitor.UpdatePerfTimer(TreeGrowingPerfMonitor::timer_name::INIT_NEW_NODE);
+}
+
+template<typename RowIdxType, typename IdxType>
+std::pair<size_t, size_t> PartitionDenseLeftDefaultKernel(const RowIdxType* rid,
```
hcho3: Should we inline this function?

src/tree/updater_quantile_hist.cc
```diff
@@ -88,94 +89,233 @@ bool QuantileHistMaker::UpdatePredictionCache(
...
+    } else {
+      p_right[iright++] = rid[i];
+    }
+  }
+
+  return { ileft, iright };
+}
+
+template<typename RowIdxType, typename IdxType>
+std::pair<size_t, size_t> PartitionDenseRightDefaultKernel(const RowIdxType* rid,
```
hcho3: Should we inline this function?

src/tree/updater_quantile_hist.cc
```diff
@@ -88,94 +89,233 @@ bool QuantileHistMaker::UpdatePredictionCache(
...
+      p_right[iright++] = rid[i];
+    } else {
+      p_left[ileft++] = rid[i];
+    }
+  }
+  return { ileft, iright };
+}
+
+template<typename RowIdxType, typename IdxType>
+std::pair<size_t, size_t> PartitionSparseKernel(const RowIdxType* rowid,
```
hcho3: Should we inline this function?

src/tree/param.h
```diff
@@ -338,15 +338,16 @@ XGBOOST_DEVICE inline float CalcWeight(const TrainingParams &p, GpairT sum_grad)
 }
 
 /*! \brief core statistics used for tree construction */
-struct XGBOOST_ALIGNAS(16) GradStats {
```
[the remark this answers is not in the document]
SmirnovEgorRu: alignas(X) means that sizeof(GradStats) == X.
But  if you allocate memory for array of GradStats - pointer to array can be not aligned by X.
Below this class has check for sizeof(GradStats) == 16, so I dont see profits from usage of alignas.

src/tree/updater_quantile_hist.h
```diff
@@ -262,6 +264,75 @@ class QuantileHistMaker: public TreeUpdater {
...
+                        int32_t nid);
+
+    void BuildNodeStatBatch(
+        const GHistIndexMatrix &gmat,
+        DMatrix *p_fmat,
+        RegTree *p_tree,
+        const std::vector<GradientPair> &gpair_h,
+        const std::vector<ExpandEntry>& nodes);
+
+    void CreateNewNodes(const GHistIndexMatrix &gmat,
```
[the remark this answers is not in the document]
SmirnovEgorRu: Done

src/tree/updater_quantile_hist.cc
```diff
@@ -209,24 +692,34 @@ void QuantileHistMaker::Builder::ExpandWithDepthWidth(
...
-    SyncHistograms(starting_index, sync_count, p_tree);
-    BuildNodeStats(gmat, p_fmat, p_tree, gpair_h);
-    EvaluateSplits(gmat, column_matrix, p_fmat, p_tree, &num_leaves, depth, &timestamp,
-                   &temp_qexpand_depth);
+
+    // buffer to store partial histograms
+    std::vector<std::vector<common::GradStatHist::GradType*>> hist_buffers;
+    // uint8_t is used instead of bool due to read/write
+    // to std::vector<bool> - thread unsafe
+    std::vector<std::vector<uint8_t>> hist_is_init;
```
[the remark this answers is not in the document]
SmirnovEgorRu: Amount of writing into hist_is_init is not so big, also it is split to separate std::vectors. VTune also say that it is not a hotspot. So, it should be okay.

src/tree/updater_quantile_hist.cc
```diff
@@ -184,23 +324,366 @@ void QuantileHistMaker::Builder::EvaluateSplits(
...
+    task_nid->resize(task_nid->size() + n_local_blocks, nid);
+    for (size_t j = 0; j < n_local_blocks; ++j) {
+      task_node_idx->push_back(i);
+      task_block_idx->push_back(j);
+    }
+
+    (*hist_buffers)[i].clear();
+    for (size_t j = 0; j < n_local_histograms; j++) {
+      (*hist_buffers)[i].push_back(
+        reinterpret_cast<common::GradStatHist::GradType*>(hist_buff_[i_hist++].data()));
```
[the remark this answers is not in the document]
SmirnovEgorRu: It has check:
  GradStatHist() : sum_grad{0}, sum_hess{0} {
    static_assert(sizeof(GradStatHist) == 8,
                  "Size of GradStatHist is not 8 bytes.");
  }
So, we are sure that GradStatHist  is 2 floats

src/tree/updater_quantile_hist.cc
SmirnovEgorRu: Done

src/tree/updater_quantile_hist.cc
SmirnovEgorRu: Done

src/tree/updater_quantile_hist.cc
SmirnovEgorRu: Done

src/tree/updater_quantile_hist.cc
SmirnovEgorRu: I dont see benefits from this, but I added this

src/tree/updater_quantile_hist.cc
SmirnovEgorRu: Done

src/tree/updater_quantile_hist.cc
SmirnovEgorRu: Done

src/tree/updater_quantile_hist.cc
SmirnovEgorRu: Done

SmirnovEgorRu:
@hcho3, thanks for review. I have applied all comments, CI is green.

hcho3:
Looks like the R project website (https://cloud.r-project.org) is currently down, causing Windows R tests to fail. For now, we will ignore the failed tests.

SmirnovEgorRu:
@hcho3, thank you!
Do I need to do smt else for merging this PR?

hcho3:
@SmirnovEgorRu It’s merged now. Thanks!

coding-komek:
Thanks for including these changes! Is there an ETA as to when a release can be expected?

arnocandel:
Are results (model, predictions) reproducible for same machine, same random state?

SmirnovEgorRu:
@arnocandel I tested this for multiple data sets - it was reproducible even for multicore systems,
Also, XGBoost has very sensitive tests (like for features importance), If you do smth wrong in terms of results сhanging - they would be failed.

SmirnovEgorRu:
Hi @hcho3, @trivialfis,
Do you know when the next release will be available?

hcho3:
See #4680

PR #4532, opened 2019-06-04
Overload device memory allocation

RAMitchell:
This PR creates a custom thrust device memory allocator. It serves as an entry point to use custom allocators in the future (such as cacheing pool allocators) and provides logging for memory usage statistics.

There should be no change in behaviour here.

The logging output looks like this:
```
[16:39:52] ======== Global Device Memory Allocations:  ========
[16:39:52] Peak memory usage: 1114mb
[16:39:52] Number of allocations: 198
```
TODO:
- [x] Thread safety

- [x] Test with multiple GPUs

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,4 +1,4 @@
-file(GLOB_RECURSE CPU_SOURCES *.cc)
+file(GLOB_RECURSE CPU_SOURCES *.cc *.h)
 list(REMOVE_ITEM CPU_SOURCES ${PROJECT_SOURCE_DIR}/src/cli_main.cc)
 
 include(CheckCXXSourceCompiles)
@@ -33,7 +33,7 @@ endif (PLUGIN_DENSE_PARSER)
 # Object library is necessary for jvm-package, which creates its own shared
 # library.
 if (USE_CUDA)
-  file(GLOB_RECURSE CUDA_SOURCES *.cu)
+  file(GLOB_RECURSE CUDA_SOURCES *.cu *.cuh)
   add_library(objxgboost OBJECT ${CPU_SOURCES} ${CUDA_SOURCES} ${PLUGINS_SOURCES})
   target_compile_definitions(objxgboost
     PRIVATE -DXGBOOST_USE_CUDA=1)
@@ -118,5 +118,9 @@ endif (USE_OPENMP)
 # replace /MD with /MT. See https://github.com/dmlc/xgboost/issues/4462
 # for issues caused by mixing of /MD and /MT flags
 msvc_use_static_runtime()
+
+# This grouping organises source files nicely in visual studio
+auto_source_group("${CUDA_SOURCES}")
+auto_source_group("${CPU_SOURCES}")
 
 #-- End object library
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -225,6 +225,67 @@ inline void LaunchN(int device_idx, size_t n, L lambda) {
   LaunchN<ITEMS_PER_THREAD, BLOCK_THREADS>(device_idx, n, nullptr, lambda);
 }
 
+class MemoryLogger {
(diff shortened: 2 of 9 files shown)


trivialfis:
Actually I kept my very old profiling branch that logs trace for memory allocation.  It's difficult to rebase now.  But you can see if there's anything reusable: https://github.com/trivialfis/xgboost/tree/profiling

src/common/device_helpers.cuh
```diff
@@ -225,6 +221,97 @@ inline void LaunchN(int device_idx, size_t n, L lambda) {
...
+        std::max(peak_allocated_bytes, currently_allocated_bytes);
+      num_allocations++;
+    }
+    void RegisterDeallocation(void *ptr) {
+      num_deallocations++;
+      currently_allocated_bytes -= device_allocations[ptr];
+      device_allocations.erase(ptr);
+    }
+  };
+  std::map<int, DeviceStats>
```
trivialfis: You can just use a vector with size of all visible GPUs obtained from GPUSet.  This way you can avoid RB tree look up and lock guard used in registration.

src/CMakeLists.txt
```diff
@@ -33,7 +33,7 @@ endif (PLUGIN_DENSE_PARSER)
 # Object library is necessary for jvm-package, which creates its own shared
 # library.
 if (USE_CUDA)
-  file(GLOB_RECURSE CUDA_SOURCES *.cu)
+  file(GLOB_RECURSE CUDA_SOURCES *.cu *.cuh)
```
trivialfis: It's not usual to include headers since CMake will track those automatically.  I'm not sure if it's gonna affect CMake.  Is it only for source grouping?

src/common/device_helpers.cuh
```diff
@@ -225,6 +221,97 @@ inline void LaunchN(int device_idx, size_t n, L lambda) {
...
+};
+};
+
+// Declare xgboost allocator
+// Replacement of allocator with custom backend should occur here
+template <typename T>
+using XGBDeviceAllocator = detail::XGBDefaultDeviceAllocator<T>;
+/** \brief Specialisation of thrust device vector using custom allocator. */
+template <typename T>
+using device_vector = thrust::device_vector<T,  XGBDeviceAllocator<T>>;
```
trivialfis: It would be more future proof if we use a different name and disable any usage of device_vector in linting.

src/predictor/gpu_predictor.cu
```diff
@@ -373,7 +373,7 @@ class GPUPredictor : public xgboost::Predictor {
   }
 
  public:
-  GPUPredictor()
+  GPUPredictor()  // NOLINT
```
trivialfis: I'm curious what's the error message.

src/CMakeLists.txt
RAMitchell: There is some discussion about it here: https://stackoverflow.com/questions/36174499/why-add-header-files-into-add-library-add-executable-command-in-cmake?rq=1

Header files don't get included in the IDE project unless they are included here. I don't see any reason not to do it

src/common/device_helpers.cuh
RAMitchell: This way is more general, works during testing and doesn't depend on global state. Your approach also assumes multiple threads will not make an allocation for a single GPU. I'm not worried about look up performance in a map when verbosity 3 is enabled.

src/predictor/gpu_predictor.cu
RAMitchell: Something about using `=default` for default constructors, I think it was a false positive.

src/common/device_helpers.cuh
RAMitchell: Is it okay if I do this in a future PR? I'm working on a second stage of this now, but I do agree with your comment.

trivialfis: @RAMitchell No problem.

PR #4534, opened 2019-06-06
Feature interaction for GPU Hist.

trivialfis:
Continuation of #4488.

This PR removed the previous attempt of reusing CPU `SplitEvaluator` to avoid device/host memory copying during tree update, as suggested by @RAMitchell .  The downside is now the memory usage on GPU side is of size `O(n_features^2) + O(n_features * n_nodes)`, let alone some other buffer allocations.  One possible improvement is to parse the input data during DMatrix construction to optimize sparse dataset memory usage, but that is beyond the scope of this PR.

TODOs:
- [x] Fuse some kernels.
- [x] Finish tests.
- [x] Figure out why is it completely broken on Windows.

### DIFF
--- a/tests/cpp/tree/test_constraints.cu
+++ b/tests/cpp/tree/test_constraints.cu
@@ -1,0 +1,281 @@
+/*!
+ * Copyright 2019 XGBoost contributors
+ */
+#include <gtest/gtest.h>
+#include <thrust/copy.h>
+#include <thrust/device_vector.h>
+#include <cinttypes>
+#include <string>
+#include <bitset>
+#include <set>
+#include "../../../src/tree/constraints.cuh"
+#include "../../../src/tree/param.h"
+#include "../../../src/common/device_helpers.cuh"
+
+namespace xgboost {
+
+struct FConstraintWrapper : public FeatureInteractionConstraint {
+  common::Span<BitField> GetNodeConstraints() {
+    return FeatureInteractionConstraint::s_node_constraints_;
+  }
+  FConstraintWrapper(tree::TrainParam param, int32_t n_features) :
+      FeatureInteractionConstraint(param, n_features) {}
+
+  dh::device_vector<int32_t> const& GetDSets() const {
+    return d_sets_;
+  }
+  dh::device_vector<int32_t> const& GetDSetsPtr() const {
+    return d_sets_ptr_;
+  }
+};
+
+
+std::string GetConstraintsStr() {
+  std::string const constraints_str = R"constraint([[1, 2], [3, 4, 5]])constraint";
+  return constraints_str;
+}
+
+tree::TrainParam GetParameter() {
+  std::vector<std::pair<std::string, std::string>> args{
+    {"interaction_constraints", GetConstraintsStr()}
+  };
+  tree::TrainParam param;
+  param.Init(args);
+  return param;
+}
+
+void CompareBitField(BitField d_field, std::set<uint32_t> positions) {
+  std::vector<BitField::value_type> h_field_storage(d_field.bits_.size());
+  thrust::copy(thrust::device_ptr<BitField::value_type>(d_field.bits_.data()),
+               thrust::device_ptr<BitField::value_type>(
+                   d_field.bits_.data() + d_field.bits_.size()),
+               h_field_storage.data());
+  BitField h_field;
+  h_field.bits_ = {h_field_storage.data(), h_field_storage.data() + h_field_storage.size()};
+
+  for (size_t i = 0; i < h_field.Size(); ++i) {
+    if (positions.find(i) != positions.cend()) {
+      ASSERT_TRUE(h_field.Check(i, true));
+    } else {
+      ASSERT_FALSE(h_field.Check(i));
+    }
+  }
+}
+
+TEST(FeatureInteractionConstraint, Init) {
+  int32_t constexpr kFeatures = 6;
+  {
+    tree::TrainParam param = GetParameter();
+    FConstraintWrapper constraints(param, kFeatures);
+    ASSERT_EQ(constraints.Features(), kFeatures);
+    common::Span<BitField> s_nodes_constraints = constraints.GetNodeConstraints();
+    for (BitField const& d_node : s_nodes_constraints) {
+      std::vector<BitField::value_type> h_node_storage(d_node.bits_.size());
+      thrust::copy(thrust::device_ptr<BitField::value_type>(d_node.bits_.data()),
+                   thrust::device_ptr<BitField::value_type>(
+                       d_node.bits_.data() + d_node.bits_.size()),
+                   h_node_storage.data());
+      BitField h_node;
+      h_node.bits_ = {h_node_storage.data(), h_node_storage.data() +  h_node_storage.size()};
+      // no feature is attached to node.
+      for (size_t i = 0; i < h_node.Size(); ++i) {
+        ASSERT_FALSE(h_node.Check(i));
+      }
+    }
+  }
+
+  {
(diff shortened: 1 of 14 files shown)


trivialfis:
@RAMitchell The initialization code is quite a mess due to the fact that I need a lots of auxiliary data structures to pass all information to device.  Feel free to reach me if got stuck.

Currently I use atomic operations to save the troubles.  Later I can map each thread to one `uint64_t` to avoid those operations.

trivialfis:
@RAMitchell I can try and implement a sparse set to drop the `O(n_feat^2)` part, based on maximum length of all interaction sets, turning it into `O(n_feat * max(length))`.  Although `O(max(length)) == n_feat` but I believe it should help a lot in practice, could you take a look at current approach before I proceed? since this optimization is just adding a look up table.

`O(n_feat * n_nodes)` part is still here.

RAMitchell:
Another way of expressing this is to pre-compute allowed feature interactions for each feature.

Lets use an example set of constraints:
`[[0,1,2],[0,4]]`

Where `->` means allowed to interact with and `*` means any feature:
```
0->0,1,2,4
1->0,1,2
2->0,1,2
3->*
4->0,4
```
If you have this available as a data structure it does not need to change or have sets merged as you grow the tree. Have I understood this correctly, is this possible? One problem with doing this is the memory requirements of expressing large subsets increases quadratically e.g. [[0, 1, 2, 3, 4, 5...]] is not efficiently represented using my approach.

One other observation about this feature interactions business is that it is possible to input a highly redundant set of feature interactions.
`[[0,1],[0,1,2],[1,2]]` is the same as `[[0,1,2]]`
Any interaction set that exists as a subset of another set is redundant.

trivialfis:
@RAMitchell 

I created a mapping: `feature -> sets it belongs to -> features in those sets`.  So the memory usage should be better.  All initialization is done on CPU side then copied to device, it makes things easier.  The mapping is done on device side and fully paralleled, so computation speed should not hurt too much.

> One other observation about this feature interactions business is that it is possible to input a highly redundant set of feature interactions.
`[[0,1],[0,1,2],[1,2]]` is the same as `[[0,1,2]]`
Any interaction set that exists as a subset of another set is redundant.

This is handled in the mapping, but not optimized.  I think a better document could be useful to prevent this kind of mistake.  It can happen since I have to dig through the source code to understand exact behaviour of this constraint. (And I can't memorize it even after writing the implementation).

The bit field is kept, 8 times less is still quite a save, especially for large trees.  If we encounter any obscured bugs of it in the future then I will just replace it with a boolean vector.

trivialfis:
> What do you think about combining this with monotone constraints into a class implementing similar methods to SplitEvaluator? It shares the need to log tree structure after applying splits.

Let me think about it, see how many code they can share.

> 
> Another question: do you think the amount of code, tests an additional complexity for the bitfield justifies the 8x space savings? Is this space saving likely to matter in practice?

 It's really due to the fact that I can't run XGBoost with the URL dataset, many of my recent effort including looking at DMatrix and CPU histogram are with that dataset in mind.  So every time an allocation depends on number of features I got sweaty.  If we can introduce some form of compressed dataset, I will happily remove the bit field class.

trivialfis:
@RAMitchell I think the two constraints are quite different regarding:
- Interaction constraint doesn't need to be passed into split evaluation kernel, monotonic constraint does.
- Interaction constraint accept a list, then return another list of features.  Monotonic constraint compute scores feature by feature.
- They both record the tree structure, but interaction constraint records the id of feature attached to nodes, while monotonic records values.

I will try to unify GPU constraints in another PR, but I don't think I can design a good interface without moving other constraints around first.  So I want this PR to focus on making interaction constraint work without touching other code (It's quite complicated already).

trivialfis:
Finally. ;-)

Thanks @RAMitchell .

RAMitchell:
Please document this also.

datametrician:
https://xgboost.readthedocs.io/en/latest/tutorials/feature_interaction_constraint.html?highlight=interaction also needs to get updated @trivialfis. Great work!

trivialfis:
@RAMitchell @datametrician Will document it along with some more details in another PR.  I'm thinking how to express the exact detail of interaction constraint.  As commented above, I only know these by digging through the original implementation.

trivialfis:
I got a segfault somewhere making sphinx doc ...

rongou:
Looks like tests are failing at head?
```shell
$ ./testxgboost --gtest_filter=FeatureInteractionConstraint.Query
Note: Google Test filter = FeatureInteractionConstraint.Query
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from FeatureInteractionConstraint
[ RUN      ] FeatureInteractionConstraint.Query
/home/rou/src/xgboost/tests/cpp/tree/test_constraints.cu:225: Failure
Expected equality of these values:
  h_output.size()
    Which is: 1
  solution.size()
    Which is: 2
/home/rou/src/xgboost/tests/cpp/tree/test_constraints.cu:225: Failure
Expected equality of these values:
  h_output.size()
    Which is: 3
  solution.size()
    Which is: 4
/home/rou/src/xgboost/tests/cpp/tree/test_constraints.cu:225: Failure
Expected equality of these values:
  h_output.size()
    Which is: 0
  solution.size()
    Which is: 1
/home/rou/src/xgboost/tests/cpp/tree/test_constraints.cu:225: Failure
Expected equality of these values:
  h_output.size()
    Which is: 0
  solution.size()
    Which is: 1
[  FAILED  ] FeatureInteractionConstraint.Query (360 ms)
[----------] 1 test from FeatureInteractionConstraint (360 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (360 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] FeatureInteractionConstraint.Query

 1 FAILED TEST
```

hcho3:
@rongou That's weird. The test is passing in the CI: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/master/251/pipeline/205

rongou:
What kind of gpu is used on the ci worker? I have two Titan V's on my desktop. It's also failing for @sriramch who is using T4.

trivialfis:
@rongou @sriramch I'm running gtx 1080 and 1080 TI.

trivialfis:
Is the error deterministic? If not, then @RAMitchell got a point maybe I should try removing the bit field.

rongou:
It fails consistently on my desktop.

trivialfis:
@rong  How about other tests:
` --gtest_filter='Feature*';`
` --gtest_filter='BitField*';`

tests/cpp/tree/test_constraints.cu
```diff
@@ -0,0 +1,321 @@
...
+     */
+
+    std::vector<int32_t> h_input_feature_list {0, 1, 2, 3, 4, 5};
+    dh::device_vector<int32_t> d_input_feature_list (h_input_feature_list);
+    common::Span<int32_t> s_input_feature_list = dh::ToSpan(d_input_feature_list);
+
+    auto s_output = constraints.Query(s_input_feature_list, 1);
+    CompareFeatureList(s_output, {0, 1, 2});
+    s_output = constraints.Query(s_input_feature_list, 2);
+    CompareFeatureList(s_output, {1, 2});
```
sriramch: @trivialfis this is the beginning of the failure.

note: if i remove line 250, then this check will fail if i query on node 1 as well

trivialfis:
I have an idea now.  Will open a new PR for testing.

PR #4538, opened 2019-06-06
make sure that xgboost has gcc 5.x at the very least to build on gcc tool chain...

sriramch:
- pr #4521 started using regex's whose support is iffy at best on gcc 4.8.x which comes standard
   with core stock centos7
- prevent someone from building with this gcc toolchain and require at least 5.x

@hcho3 @trivialfis please review

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,10 @@ if (MSVC)
   cmake_minimum_required(VERSION 3.11)
 endif (MSVC)
 
+if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.0)
+  message(FATAL_ERROR "GCC version must be at least 5.0!")
+endif()
+
 set_default_configuration_release()
 
 #-- Options
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,10 +9,6 @@ if (MSVC)
   cmake_minimum_required(VERSION 3.11)
 endif (MSVC)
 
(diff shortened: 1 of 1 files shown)


sriramch:
@trivialfis / @hcho3 can you please merge when you get a chance.

trivialfis:
@sriramch Thanks for the PR.

thesuperzapper:
@sriramch this is causing me some grief, as when using spark-xgboost, people often need to run on Hadoop clusters which use redhat7.

Even if we cant build on gcc-4.8, can we support running with older versions of GLIBC?

trivialfis:
I thought cent 7 is the same with RH 7?

thesuperzapper:
@trivialfis, yeah, for most things, but that's not the issue here. We need to support redhat7 or many real world usages of xgboost will not work.

Currently if your forced to build with g++-5 this means you can't build on redhat7. (Which you need to do to get the older dynamic GLIBC-2.17, which it needs)

This really isn't my area of expertise, but with the the current master, I have to undo this PR to build xgboost4j for redhat. Can we find another way to do whatever is requiring g++-5?

sriramch:
@thesuperzapper is it not possible to use the devtoolset-6 or 7 on rh7? 

the core stock centos supports glibc-2.17 by default and i was able to build using this and devtoolset-6 on it.  the devtoolset will pull the newer cxxabi's internally and can still work with the libcxx abi present on that host - in the cent os case, the libcxx abi that came with gcc 4.8.5. 

the reason why this is required is documented in this pr.

PR #4539, opened 2019-06-06
Add c_api callbacks for direct DMatrix construction

RAMitchell:
This PR proposes allowing direct construction of a DMatrix object via an external callback function. This is in the same spirit as [this existing API function](https://github.com/dmlc/xgboost/blob/59ae42a1792697165507a04f4ef4879487aba2ca/include/xgboost/c_api.h#L125), but where this function copies the data entirely three times (!!), my function does it once. It cannot be used to create an external memory DMatrix, only a simple in memory DMatrix.

The idea here is to give libraries such as cudf and python datatable a pathway to efficiently create a DMatrix without explicitly creating API functions for each new library. I can trivially add a device memory version of this function.

One possible downside is that this relies on us keeping the in memory format constant. If we change the internal data layout or types for SimpleCSRSource this would break.

### DIFF
--- a/demo/c-api/README.md
+++ b/demo/c-api/README.md
@@ -5,6 +5,22 @@ C-APIs
 bindings.  For detailed reference, please check xgboost/c_api.h.  Here is a
 demonstration of using the API.
 
+# Train
+This example shows how to load data into a DMatrix, train and make predictions.
+
+To run the training example from this directory:
+```bash
+cd train
+make
+# Make sure the system can find the xgboost shared library
+export LD_LIBRARY_PATH=${LD_LIBRARY_PATH=}:${PWD}/../../../lib
+./c-api-demo
+```
+This demo assumes a Linux system with make and that the xgboost library is compiled.
+
+# Custom DMatrix creation
+This example shows how an external library can create a DMatrix object using function callbacks.
+
 # CMake
 If you use **CMake** for your project, you can either install **XGBoost**
 somewhere in your system and tell CMake to find it by calling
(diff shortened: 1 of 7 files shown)


trivialfis:
@RAMitchell Can you explain a little bit more?  I actually don't understand why does it need a call back to initialize dmatrix for non-batched data.  I haven't look into cudf's data representation yet, but it have to be one of those CSR/CSC/COO/Dense ... right?  Can we "just copy" it?

RAMitchell:
You can see the previous attempt at integration in #3997. The conversion is nontrivial and doesn't generalise, the same goes for datatable. This would essentially push the custom code for each library out of xgboost back onto the other libraries.

We are also looking at possible ways to integrate based on the `__cuda_array_interface__` attribute. It's not exactly clear what the best way to integrate other data containers easily into xgboost without imposing maintenance burden on xgboost.

trivialfis:
@RAMitchell Thanks for the explanation.  I didn't know it was so intrusive.   Just read the documentation of cudf,  it makes me feel weird that cudf has all those `to_*` methods ranging from `to_numpy`, `to_gpu_matrix` to `to_pandas`, but chooses to implement a `to_csr` in XGBoost? ;-)  I need some time for getting familiar with cudf.

RAMitchell:
To update, we are currently looking at another approach based on [`__cuda_array_interface__`](https://numba.pydata.org/numba-doc/dev/cuda/cuda_array_interface.html) to achieve integration with external device memory data structures. We would serialise this python dict as a JSON string and pass it to the C API. As a byproduct of this we could also be able to directly ingest data from numba, cupy and pytorch. I will leave this PR open for now as another option.

trivialfis:
@RAMitchell Please take #4147, #4150 into consideration while designing the interface.

PR #4545, opened 2019-06-10
Fix doc and demo for Python package.

trivialfis:
@RAMitchell Could you help verifying the Window pip issue in this file?  If it still persist I will move the note to doc/build.rst, otherwise I just remove all notes.


RAMitchell:
Just did a quick test from pip install, everything seems fine.

trivialfis:
@RAMitchell Thanks.  That's a good news.

PR #4547, opened 2019-06-11
[RFC] Disscussion for how to refactor DMatrix.

trivialfis:
Hi, @RAMitchell , since we are all working on DMatrix related code, I want to be on the same page how should we proceed to avoid conflicting work.

## Background
This PR is only a demo for how a histogram indices matrix is built.  Previously `hist_utils.cc` iterates on rows by calling `GetRowBatches`, which is complicated and memory demanding.  Here I created `data/histogram.cc` which contains 73 lines of code but is already capable of building a histogram indices (no consideration for weight groups).  And it consumes **much less** memory for sparse dataset since it understands number of rows for each column.  Also, I can further optimize it based on number of non-empty columns.  The key is by using `GetColumnBatches` (CSC instead of CSR) instead.

## The problem
DMatrix converts everything into CSR, so calling `GetColumnBatches` is quite slow, also waste in memory.  I'm positive that once the barrier is removed, our initialization stage will be much more efficient in both memory and speed.  Can we discuss a plan for bridging between backend formats including `libsvm`, `csv`, `csr`, `csc`, `dense` and `__cuda_array_interface__` to frontend outputs including `histogram indices`, `csc`, `csr`, `sorted csc`?

### DIFF
--- a/src/common/hist_util.h
+++ b/src/common/hist_util.h
@@ -114,7 +114,7 @@ struct HistCutMatrix {
 };
 
 /*! \brief Builds the cut matrix on the GPU.
- *  
+ *
  *  \return The row stride across the entire dataset.
  */
 size_t DeviceSketch
--- a/src/data/histogram.cc
+++ b/src/data/histogram.cc
@@ -1,0 +1,73 @@
+#include <xgboost/logging.h>
+#include "histogram.h"
+
+namespace xgboost {
+
+CutMatrix::CutMatrix() : column_ptrs_{0} {}
+
+void CutMatrix::Build(DMatrix* dmat, uint32_t const max_num_bins) {
+  using WXQSketch = common::WXQuantileSketch<bst_float, bst_float>;
+  auto& info = dmat->Info();
+  constexpr float kFactor = 8;
+
+  for (auto const& batch : dmat->GetColumnBatches()) {
+    for (uint32_t col_id = 0; col_id < batch.Size(); ++col_id) {
+      WXQSketch sketch;
+      common::Span<xgboost::Entry const> column = batch[col_id];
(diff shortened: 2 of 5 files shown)


trivialfis:
@thvasilo @CodingCat  @hcho3 @RAMitchell  Suggestions are welcomed.

hcho3:
Would love to review this next Monday, when I come back from vacation. Looking forward to it :)

trivialfis:
Also, I glanced through quantiles histogram method, there is no need for getting row batches during update.  So the column matrix class can be removed, which should also result in a big speed up.   The essential part is how do we construct the DMatrix to suite algorithm's need.

RAMitchell:
This looks like a very clean interface, I like it a lot. What is your plan for avoiding calling `GetColumnBatches`?

One of the things I have struggled with in thinking about a DMatrix refactor is how to reuse these derived matrix types (such as a histogram matrix) across different parts of xgboost code without recreating them and without simply adding a method to DMatrix (column batches is an example of this). For example what if I want to make predictions using the histogram matrix without creating it from scratch.

Right now I am thinking of using a factory singleton. We give this singleton a pointer to the DMatrix and it lazily gives us the histogram matrix by looking up if it exists already in a map. All derived matrix types including column transformations and GPU matrix data structures would have their own singleton generator.

Each of these derived matrix types would need to implement their own methods for serialisation in order to provide external memory training.

hcho3:
Didn’t have time to check correctness, but I approve the proposed interface.

trivialfis:
@hcho3 this is just a demo for the necessity of building different dmatrix for different algorithms.  The purpose of this PR is to show that for histogram method, CSC is better than CSR.  But I'm not sure about how to design the dmatrix with flexible underlying formats.

trivialfis:
Will open a real PR once I have a concrete plan.

Issue #4542, opened 2019-06-07
[RFC] Optimize configurations in Learner.


trivialfis:
Currently the configuration steps in Learner is bloated in both size and complexity.  I think we have all come to agreement that it needs better organization.  Other than simplifying future maintenance, there are other benefits on this:

- Parameter validation, the configuration happens in both initialization and during each update so copies of parameters are stored in various places, including both `Learner` and `GradientBooster`.  So we don't know exactly when a parameter should be used nor which copy is actually used.

- Smooth out the path for our plan of transitioning to JSON format since one big issue is keeping compatibility, and I want the JSON format to be a fresh start.

To do so, we need to draft a testing suit for current binary model IO so that later refactoring won't break the existing model.  Then we need to decide what style of parameters management we want:
- A global one like most of other gbm libraries, all parameters stored in one `config` class, then we pass it through the whole library.
- Many local parameter structs just like current XGBoost.

Personally I don't have a preference.

I am struggling whether should JSON happen first or optimization for Learner.  But parameter validation is definitely coming after we reorganize those configurations.

@RAMitchell @CodingCat @hcho3

trivialfis:
One funny thing, I tried to move tree updater configuration into GBM, the `Learner.SLOW_CheckMultiBatch` test gave me a dead lock in `col_maker`.

trivialfis:
During many previous fixes and refactoring PRs including #4553,  I came to realize that how should we proceed on parameter validations.  There are two fundamental problems in current XGBoost internal parameters handling:

- The semantic of `Configure` and `Init` is not clear, and they depend on each others.  In many places one has to allocate memory for components (Init) to do further configurations, example are gbm asking updaters to configure themselves.  But it's also not possible to do `Init` without first configuring stuffs out, example like creating updaters after selecting tree methods.
- Some configurations has to be done after seeing the DMatrix, like `num_feature`, `n_gpus`(to be deprecated).  While our abstract `Configure` is not prepared for this, it has no parameter for DMatrix.
- Most of the configurations can be changed during each boost iteration.  Our configuration is also not prepared for this.  It doesn't know whether those string parameters are changed or not, so we just plug some partial configuration into `Update` or `DoBoost`.

My proposal is, 
- Firstly we clear up the semantic of `Init` and `Configure`.  There should be no public `Init` method.  All initialization should be a part of `Configure`.  And the configuration should be capable of detecting whether parameters has changed, like comparing new and old `dmlc::Parameter`.
[...]

PR #4553, opened 2019-06-12
Offload some configurations into GBM.

trivialfis:
* Move tree heuristic configurations into GBTree.
* Split up declarations and definitions of GBTree for testing.
* Implement UseGPU in gbm to detect whether we are using GPU algorithm.

See  #4542 for more comments and plan.  This is part one of it, one or two more PRs are coming.


trivialfis:
@RAMitchell Yes, added a test for that.

PR #4558, opened 2019-06-13
Set reg_lambda=0 for scikit-learn-like random forest classes.

canonizer:


RAMitchell:
These are the experiments that led to this, soon to be published in a blog. L2 penalty only increases bias in random forest models, whereas boosting models can compensate for this bias by boosting more rounds.
bias_variance_decomposition_gradient_boosting_lambda_(l2_penalty)
bias_variance_decomposition_random_forest_lambda_(l2_penalty)

I think we should set lambda to 1e-5 instead of 0 to prevent numerical instability in weight calculation -G/(H+ lambda) as the hessian tends to 0.

canonizer:
Done.

PR #4569, opened 2019-06-16
add C_API about incremental DMatrix building

sperlingxx:
Add C_API for data::DataSource, by which,  data can be fetched gradually by appending to the tail of DataSource. And Then,  DMatrix can be created by data::DataSource.

I think, these APIs is helpful to build _DMatrix Builder_ for other languages (like python). Since now, when we build a DMatrix in python, we need cache all data in format like pandas.DataFrame. With _DMatrix Builder_ , additional cache can be avoided. 

 I'm willing to write _DMatrix Builder_  for python later.

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -35,6 +35,8 @@ typedef void *BoosterHandle;  // NOLINT(*)
 typedef void *DataIterHandle;  // NOLINT(*)
 /*! \brief handle to a internal data holder. */
 typedef void *DataHolderHandle;  // NOLINT(*)
+/*! \brief handle to a data source. */
+typedef void *DataSourceHandle;  // NOLINT(*)
 
 /*! \brief Mini batch used in XGBoost Data Iteration */
 typedef struct {  // NOLINT(*)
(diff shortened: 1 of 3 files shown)


sperlingxx:
@trivialfis  Can you help to review this?  :)

trivialfis:
@RAMitchell Could you please take a look?  Looks like it's related to #4539 .

RAMitchell:
@sperlingxx please take a look at my callback implementation in #4539 and see if it meets your needs. Also look at the function `XGDMatrixCreateFromDataIter` from the C API for loading data in batches.

sperlingxx:
> @sperlingxx please take a look at my callback implementation in #4539 and see if it meets your needs. Also look at the function `XGDMatrixCreateFromDataIter` from the C API for loading data in batches.

I think `XGDMatrixCreateFromDataIter` isn't so friendly to other languages as to Java(JNI).

#4539 is cool, but it also need to allocate/manage continous memory space out of callback, which may be not so easy in other language?

trivialfis:
@sperlingxx See comment in https://github.com/dmlc/xgboost/pull/4539#issuecomment-500609198 .  WDYT?

sperlingxx:
@trivialfis  In my opinion, `DataSourceAppend` seems more directly/transparently than `callbacks`.

RAMitchell:
I don't know if the data source class is really designed to be exposed to end users. What about an append method for the DMatrix itself?

sperlingxx:
@RAMitchell  I agree with this idea.

RAMitchell:
Closing for now, lets try to come up with a solution that does not expose DataSource to the end user.

sperlingxx:
@RAMitchell  Okay, do you have any ideas? Maybe we can establish a new class like `DMatrixBuilder`?

RAMitchell:
I don't think we need a new class. What about an append method that allows you to concatenate two dmatrix objects together?
```
dmatrix1.append(dmatrix2)
```
You would need to check the dimensions are all correct. Internally in the C++ code I would probably implement this as an append method on the dmatrix interface, then implement this for SimpleCSRSource, throw a not implemented error for SparsePageDMatrix (I'm not sure if appending makes sense for external memory).

First I would want to be very sure it's not possible to implement this using `XGDMatrixCreateFromDataIter`.

sperlingxx:
@RAMitchell `dmatrix1.append(dmatrix2)` sounds good to me.

I think `XGDMatrixCreateFromDataIter` is tightly binding with jvm-packages. For example,  group columns is ingored in `XGDMatrixCreateFromDataIter`, which is handled in java codes and appended into DMatrix by `set_float_info`.

PR #4572, opened 2019-06-17
- do not create device vectors for the entire sparse page while compu…

sriramch:
…ting histograms...

   - while creating the compressed histogram indices, the row vector is created for the entire
     sparse page batch. this is needless as we only process chunks at a time based on a slice
     of the total gpu memory
   - this pr will allocate only as much as required to store the appropriate row indices and the entries

@RAMitchell please review.

### DIFF
--- a/src/tree/updater_gpu_hist.cu
+++ b/src/tree/updater_gpu_hist.cu
@@ -695,8 +695,6 @@ struct DeviceShard {
   /*! \brief Sum gradient for each node. */
   std::vector<GradientPair> node_sum_gradients;
   common::Span<GradientPair> node_sum_gradients_d;
-  /*! \brief On-device feature set, only actually used on one of the devices */
-  dh::device_vector<int> feature_set_d;
   dh::device_vector<int64_t>
       left_counts;  // Useful to keep a bunch of zeroed memory for sort position
   /*! The row offset for this shard. */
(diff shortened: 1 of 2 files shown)


src/tree/updater_gpu_hist.cu
```diff
@@ -1339,35 +1328,36 @@ inline void DeviceShard<GradientSumT>::CreateHistIndices(
...
+    /*! \brief row offset in SparsePage (the input data). */
+    dh::device_vector<size_t> row_ptrs(batch_nrows+1);
+    thrust::copy(
+      offset_vec.data() + device_row_state.row_offset_in_current_batch + batch_row_begin,
+      offset_vec.data() + device_row_state.row_offset_in_current_batch + batch_row_end + 1,
+      row_ptrs.begin());
+
     // number of entries in this batch.
-    size_t n_entries = row_ptrs[batch_row_end] - row_ptrs[batch_row_begin];
+    size_t n_entries = row_ptrs.back() - row_ptrs.front();
```
RAMitchell: I never noticed this before, but isn't the dereferencing elements of `row_ptrs` on the host invoking a bunch of tiny memory copies from the device to host?

sriramch: yes, it will. i have made the requisite changes; please take 1 more look.

fwiw, there is no perceptible change in performance with this.

RAMitchell:
@sriramch it would help me if you can format the text of the PR title and the commit messages as you want them to appear to other developers. Normally I will remove commit messages about lint fixes, so don't worry about those, but if the title and other commit messages can be descriptive and well formatted that would be great.

sriramch:
> @sriramch it would help me if you can format the text of the PR title and the commit messages as you want them to appear to other developers. Normally I will remove commit messages about lint fixes, so don't worry about those, but if the title and other commit messages can be descriptive and well formatted that would be great.

@RAMitchell thanks for the tip. i'll remember this for the future pr's. i see that you already changed the title for this one.

PR #4574, opened 2019-06-17
upgrade to Scala 2.12

koertkuipers:
hello!

spark 2.4 will be last spark release to support scala 2.11 and is also available for scala 2.12

flink 1.7 is available for scala 2.11 and scala 2.12. i am not familiar enough with flink or its community to say when they will drop scala 2.11

this pullrequest builds the jvm libs using scala 2.12. to make the switch i had to change java version to 1.8 (required for scala 2.12) and upgrade flink and akka.

to be able to publish artifacts for both scala 2.11 and 2.12 (scala is not compatible across minor versions) i modified the artifacts to have scala version in the name. its not pretty :( but it seems to be the convention.

to publish artifacts for scala 2.11 one just has to make a few small modifications to pom files. this can probably easily be automated with a bash script. e.g. `jvm-packages/dev/change_scala_version.sh`. i am not aware of an approach with maven where both scala 2.11 and scala 2.12 artifacs can be created more elegantly.

### DIFF
--- a/tests/ci_build/Dockerfile.jvm_cross
+++ b/tests/ci_build/Dockerfile.jvm_cross
@@ -22,7 +22,7 @@ RUN \
     ln -s /opt/apache-maven-3.6.1/ /opt/maven && \
     # Spark
     # This should be: wget https://archive.apache.org/dist/spark/spark-$SPARK_VERSION/spark-$SPARK_VERSION-bin-scala2.12-hadoop2.7.tgz
-    python -m awscli s3 cp s3:///xgboost-ci-jenkins-artifacts/spark-$SPARK_VERSION-bin-scala2.12-hadoop2.7.tgz . && \
+    python -m awscli s3 cp s3://xgboost-ci-jenkins-artifacts/spark-$SPARK_VERSION-bin-scala2.12-hadoop2.7.tgz . && \
     tar xvf spark-$SPARK_VERSION-bin-scala2.12-hadoop2.7.tgz -C /opt && \
     ln -s /opt/spark-$SPARK_VERSION-bin-scala2.12-hadoop2.7 /opt/spark
 
--- a/tests/ci_build/Dockerfile.jvm_cross
+++ b/tests/ci_build/Dockerfile.jvm_cross
@@ -20,9 +20,10 @@ RUN \
     tar xvf apache-maven-3.6.1-bin.tar.gz -C /opt && \
     ln -s /opt/apache-maven-3.6.1/ /opt/maven && \
     # Spark
-    wget https://archive.apache.org/dist/spark/spark-$SPARK_VERSION/spark-$SPARK_VERSION-bin-hadoop2.7.tgz && \
-    tar xvf spark-$SPARK_VERSION-bin-hadoop2.7.tgz -C /opt && \
-    ln -s /opt/spark-$SPARK_VERSION-bin-hadoop2.7 /opt/spark
+    # This should be: wget https://archive.apache.org/dist/spark/spark-$SPARK_VERSION/spark-$SPARK_VERSION-bin-scala2.12-hadoop2.7.tgz && \
+    wget https://tresata-public.s3.amazonaws.com/spark-$SPARK_VERSION-bin-scala2.12-hadoop2.7.tgz && \
+    tar xvf spark-$SPARK_VERSION-bin-scala2.12-hadoop2.7.tgz -C /opt && \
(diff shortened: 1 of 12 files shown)


tests/ci_build/Dockerfile.jvm_cross
```diff
@@ -20,9 +20,10 @@ RUN \
     tar xvf apache-maven-3.6.1-bin.tar.gz -C /opt && \
     ln -s /opt/apache-maven-3.6.1/ /opt/maven && \
     # Spark
-    wget https://archive.apache.org/dist/spark/spark-$SPARK_VERSION/spark-$SPARK_VERSION-bin-hadoop2.7.tgz && \
-    tar xvf spark-$SPARK_VERSION-bin-hadoop2.7.tgz -C /opt && \
-    ln -s /opt/spark-$SPARK_VERSION-bin-hadoop2.7 /opt/spark
+    # This should be: wget https://archive.apache.org/dist/spark/spark-$SPARK_VERSION/spark-$SPARK_VERSION-bin-scala2.12-hadoop2.7.tgz && \
```
CodingCat: @hcho3 shall we put this file to the image of our CI workers?

CodingCat:
isn't this PR dropping the support of 2.11 instead of support both 2.11 and 2.12?

koertkuipers:
> isn't this PR dropping the support of 2.11 instead of support both 2.11 and 2.12?

i made sure to select libraries and syntax that work well for both scala 2.11 and 2.12. 

maven doesnt support building artifacts for two scala versions. but artifacts for scala 2.11 can be build and tested by making a few very small changes to the pom file. this could be automated in a bash script if desired. 

i think the same probably applies to the automated testing infrastructure, although it will be a little more work. for example `appveyor.yml` and `tests/ci_build/Dockerfile.jvm_cross` would have to be modified dynamically. i assume/hope environment variables can be used for this.

CodingCat:
> > isn't this PR dropping the support of 2.11 instead of support both 2.11 and 2.12?
> 
> i made sure to select libraries and syntax that work well for both scala 2.11 and 2.12.
> 
> maven doesnt support building artifacts for two scala versions. but artifacts for scala 2.11 can be build and tested by making a few very small changes to the pom file. this could be automated in a bash script if desired.
> 
> i think the same probably applies to the automated testing infrastructure, although it will be a little more work. for example `appveyor.yml` and `tests/ci_build/Dockerfile.jvm_cross` would have to be modified dynamically. i assume/hope environment variables can be used for this.

it's fine, we can safely just support 2.12 as we are moving to the next 1.0 release

mhlr:
What is the time frame for a release containing this?

tests/ci_build/Dockerfile.jvm_cross
hcho3: @CodingCat Apache doesn't distribute artifacts with 2.12?

CodingCat: 2.12 is in maven central, but no downloadable artifact

hcho3: @CodingCat Got it, we can put it in a S3 bucket and have Jenkins CI workers access it

CodingCat: @hcho3 would you mind putting this file to a place which is under your control instead of using @koertkuipers's personal account? after that I can merge it with some changes

CodingCat: ping @hcho3

hcho3: @CodingCat Done

hcho3:
@dre-hh FYI, we are working to compile XGBoost with Scala 2.12.

tests/ci_build/Dockerfile.jvm_cross
CodingCat: thanks @hcho3

CodingCat:
as 2.12 would be our major goal, I don't think we still need to change the artifact name, 

what do you think @koertkuipers

koertkuipers:
> as 2.12 would be our major goal, I don't think we still need to change the artifact name,
> 
> what do you think @koertkuipers

i prefer to change artifact name to have scala version in it (as ugly as it is...)

folks might want to cross-build for 2.11 and 2.12 (or for 2.12 and 2.13 at some point for spark 3). for them life will be easier if you have scala version in artifact name.

CodingCat:
make sense

CodingCat:
merged, thanks

Issue #4531, opened 2019-06-03
[RFC] Remove support for single process multi-GPU


RAMitchell:
This issue proposes removing support for single process multi-GPU in xgboost. What this means is that we would no longer support the 'n_gpus' parameter and the ability to train with multiple GPUs without some distributed framework.

We would focus on multi-GPU support through distributed frameworks such as Dask (see #4473) and Spark, where the framework would be expected to assign a worker process for each GPU independently.

The reasoning is that it would greatly simplify all GPU code in xgboost. Currently GPU code works by accepting an input DMatrix and then manually partitioning rows across each available GPU. This partitioning as well as managing multiple GPUs via threads imposes significant complexity on developers. The result is more code that is harder to read, more bug prone (we have many issues documenting this) and less flexible when developing new features. It is difficult to overstate how beneficial this would be for developers currently working on GPU code.

By doing this we are effectively deferring the task of partitioning data for a single machine with multiple GPUs onto the higher level distributed framework.

In the short term this will be a direct removal of an existing feature, but I believe in the medium term (1 year) it will result in a significantly better product.

We would provide a path for users to transition to new multi-GPU support. This will be better for users in the long run also as these distributed solutions are arbitrarily scalable. Between making the decision to deprecate this functionality and removing code we would wait at least three months.

This is still very much up for debate and feedback is welcome from developers and users, particularly those who currently use multi-GPU support as a part of their workflow.

TimZaman:
Yes! This is one of reasons why in DL land horovod got so popular. The single process story of TensorFlow was too complicated, and didn't scale to multinode either. Nowadays most multi gpu code I see works with horovod and torch.distributed. Both are in the multiprocess+single-gpu-per-process paradigm. This also naturally scales to multiple nodes without any code editting and provides the user (and dev) with a simpler multigpu programming model.

trivialfis:
I like the idea of decoupling data distribution from algorithm.  But I insist that this is better done in c++ instead of high level libraries like dask, there are R and Julia packages that don't attract much attentions as others.  I want least maintenance burden added to language bindings.  Those thick bindings are much harder to maintain than current slightly bloated internal code.

trivialfis:
@TimZaman I thought the way TF distributing subgraph is quite elegant.  Seems I'm a little behind on this. :)

datametrician:
[...]

PR #4579, opened 2019-06-18
deprecate single node multi-gpu mode

rongou:
First step for #4531 

@canonizer @RAMitchell @trivialfis


RAMitchell:
Can we add a `LOG(WARNING)` maybe in the initialisation of gpu_hist to detect if someone is using this?

rongou:
@RAMitchell added warning message.

PR #4591, opened 2019-06-20
Add instruction to run formatting checks locally [skip ci]

hcho3:
cc @sriramch

Aside: now it is possible to add [skip ci] in the commit message to skip Jenkins tests for a single pull request.

### DIFF
--- a/doc/contribute.rst
+++ b/doc/contribute.rst
@@ -10,6 +10,12 @@ Everyone is more than welcome to contribute. It is a way to make the project bet
 **Guidelines**
 
 * `Submit Pull Request`_
+* `Running Formatting Checks Locally`_
+
+  - `Linter`_
+  - `Clang-tidy`_
+  - `Running checks inside a Docker container (Recommended)`_
+
 * `Git Workflow Howtos`_
 
   - `How to resolve conflict with master`_
@@ -19,7 +25,6 @@ Everyone is more than welcome to contribute. It is a way to make the project bet
 * `Documents`_
 * `Testcases`_
 * `Sanitizers`_
-* `clang-tidy`_
 * `Examples`_
 * `Core Library`_
 * `Python Package`_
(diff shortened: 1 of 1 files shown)


sriramch:
👍 @hcho3 thanks a lot for doing this.

doc/contribute.rst
```diff
@@ -44,6 +45,24 @@ Submit Pull Request
...
+*********************************
+
+Once you submit a pull request to `dmlc/xgboost <https://github.com/dmlc/xgboost>`_, we perform
+automatic checks to enforce coding style conventions (we use `pylint <https://github.com/PyCQA/pylint>`_,
+`cpplint <https://github.com/cpplint/cpplint>`_, and `clang-tidy <https://clang.llvm.org/extra/clang-tidy/>`_).
+You can run formatting checks locally using the following commands (Docker required):
+
+.. code-block:: bash
+
+  tests/ci_build/ci_build.sh clang_tidy docker --build-arg CUDA_VERSION=9.2 tests/ci_build/clang_tidy.sh
```
trivialfis: We already have documentation for tidy in this file.

I don't think it takes a docker to run the lint for normal contributors.  It's the safest way to go, but maybe add a note for running it without docker and just let the command line error report whatever is missing?

hcho3:
@dmlc/xgboost-committer Note to maintainers: `[skip ci]` should be used sparingly. For most pull requests, all tests should run. One legitimate case for using `[skip ci]` is document updates (Markdown/RST).

hcho3:
For some reason, Skip CI plugin in Jenkins doesn't work for the Win64 pipeline. I may want to merge Win64 pipeline into the main pipeline later, but for now, this is good enough.

doc/contribute.rst
hcho3: The CI server will run clang-tidy with both `--cpp` and `--cuda` enabled, and for `--cuda`, we need the CUDA toolkit. (Don't need a physical GPU however.)

hcho3: @trivialfis Let me think over a bit and add a sentence about running the checks without Docker. My original intention was to exactly reproduce whatever the Jenkins CI server is doing, but I see how Docker may be too high requirement for some contributors.

hcho3:
@trivialfis Can you look at this again?

PR #4596, opened 2019-06-21
Update doc for feature constraints.  Fix some warnings.

trivialfis:
I'm not confident about my writing, please help verifying.  Otherwise it might be very embarrassing ...

I am not sure about how to fix the following warnings yet:
```
/home/fis/Workspace/xgb/xgboost/doc/gpu/index.rst:89: WARNING: Inline interpreted text or phrase reference start-string without end-string.
/home/fis/Workspace/xgb/xgboost/doc/jvm/index.rst:3: WARNING: Duplicate explicit target name: "here".
/home/fis/Workspace/xgb/xgboost/doc/jvm/xgboost4j_spark_tutorial.rst:3: WARNING: Duplicate explicit target name: "here".
/home/fis/Workspace/xgb/xgboost/doc/jvm/xgboost4j_spark_tutorial.rst:3: WARNING: Duplicate explicit target name: "here".
/home/fis/Workspace/xgb/xgboost/python-package/xgboost/plotting.py:docstring of xgboost.to_graphviz:19: WARNING: Unexpected indentation.
/home/fis/Workspace/xgb/xgboost/python-package/xgboost/plotting.py:docstring of xgboost.to_graphviz:24: WARNING: Unexpected indentation.

```

PR #4598, opened 2019-06-22
Doc for customized objective/metric.

trivialfis:
Depends on #4596.


trivialfis:
@thvasilo Thanks!  This proves a point, watching television doesn't improve English.

Actually I was hesitating about this PR, I think there are rooms for us to improve multi-classes and distributed objectives/metrics.  But if this PR goes into official document, it would be very difficult for us to change anything.

trivialfis:
@thvasilo Thanks for the review.

PR #4602, opened 2019-06-24
Implement tree model dump with code generator.

### LINKED ISSUE #4228: `plot_tree` failed to parse tree nodes with feature names.
Here is a sample output from `plot_tree`:
```
[04:53:30] DEBUG: /home/fis/Workspace/xgb/xgboost/src/c_api/c_api.cc:1039: 0:[Init_Win_bytes_backward<28960] yes=1,no=2,missing=2
	1:[Total Length of Fwd Packets<25145] yes=7,no=8,missing=8
		7:[Init_Win_bytes_backward<235] yes=15,no=16,missing=15
			15:leaf=-0.0792390853
			16:[Init_Win_bytes_backward<237] yes=17,no=18,missing=18
				17:[Flow Bytes/s<969.591125] yes=19,no=20,missing=20
					19:[Destination Port<88] yes=21,no=22,missing=21
						21:[Flow IAT Min<160] yes=23,no=24,missing=23
							23:leaf=0.0835221782
							24:leaf=-0.0530516468
						22:leaf=-0.0624887124
					20:leaf=-0.0714369118
```

Used parameters are rather simple:
``` python
params = {'tree_method': 'gpu_hist',
          'verbosity': 3,
          'n_gpus': '2',
          'max_depth': 8,
          'grow_policy': 'lossguide',
          'num_parallel_tree': 4}

bst = xgb.train(
    params, dtrain, evals=[(dtest, 'dtest')], num_boost_round=8)
```

Failed with error:
```
  File "/home/fis/Workspace/general-python/lib/python3.6/site-packages/xgboost/plotting.py", line 138, in _parse_node
    raise ValueError('Unable to parse node: {0}'.format(text))
ValueError: Unable to parse node: 1:[Total
```

### LINKED ISSUE #4561: ValueError: Unable to parse node: 0:
Plot tree encounter errors. How to fix the problems.
image
image

trivialfis:
* Split up builders.
* Use pattern matching.
* Implement generator for Graphviz.

Closes #4228, Closes #4561

There are few benefits of this PR:
- After introducing proper code templates in generator, the semantic of generated text files become clearer.  Currently we don't have documentation around what exactly a dumped model should look like.
- Eliminates parsing the text dump again in language bindings.  One can generate very weird output by parsing the dumped text, as in above linked issues.  This PR implements a generator for Graphviz, which directly parses the tree model instead of text output.
- All bindings of Graphviz can accept dot file strings, so it should be easier for scala and jvm (and any other languages) to implement plotting.  @CodingCat. (I did a simple grep in the jvm package didn't find any implementation).

But there's one downside:
- Now we need to pass extra parameters into c++ for Graphviz.

@CodingCat  @RAMitchell @hcho3 Could you please take a quick look? See if the added complication is worthy for fixing above issues?

### DIFF
--- a/python-package/xgboost/plotting.py
+++ b/python-package/xgboost/plotting.py
@@ -4,7 +4,6 @@
 """Plotting Library."""
 from __future__ import absolute_import
 
-import re
 from io import BytesIO
 import numpy as np
 from .core import Booster
@@ -117,56 +116,11 @@ def plot_importance(booster, ax=None, height=0.2,
     return ax
 
 
-_NODEPAT = re.compile(r'(\d+):\[(.+)\]')
-_LEAFPAT = re.compile(r'(\d+):(leaf=.+)')
-_EDGEPAT = re.compile(r'yes=(\d+),no=(\d+),missing=(\d+)')
-_EDGEPAT2 = re.compile(r'yes=(\d+),no=(\d+)')
-
-
-def _parse_node(graph, text, condition_node_params, leaf_node_params):
-    """parse dumped node"""
-    match = _NODEPAT.match(text)
-    if match is not None:
-        node = match.group(1)
-        graph.node(node, label=match.group(2), **condition_node_params)
-        return node
-    match = _LEAFPAT.match(text)
-    if match is not None:
-        node = match.group(1)
-        graph.node(node, label=match.group(2), **leaf_node_params)
-        return node
-    raise ValueError('Unable to parse node: {0}'.format(text))
-
-
-def _parse_edge(graph, node, text, yes_color='#0000FF', no_color='#FF0000'):
-    """parse dumped edge"""
-    try:
-        match = _EDGEPAT.match(text)
-        if match is not None:
-            yes, no, missing = match.groups()
-            if yes == missing:
-                graph.edge(node, yes, label='yes, missing', color=yes_color)
-                graph.edge(node, no, label='no', color=no_color)
-            else:
-                graph.edge(node, yes, label='yes', color=yes_color)
-                graph.edge(node, no, label='no, missing', color=no_color)
-            return
-    except ValueError:
-        pass
-    match = _EDGEPAT2.match(text)
-    if match is not None:
-        yes, no = match.groups()
-        graph.edge(node, yes, label='yes', color=yes_color)
-        graph.edge(node, no, label='no', color=no_color)
-        return
-    raise ValueError('Unable to parse edge: {0}'.format(text))
-
-
-def to_graphviz(booster, fmap='', num_trees=0, rankdir='UT',
-                yes_color='#0000FF', no_color='#FF0000',
+def to_graphviz(booster, fmap='', num_trees=0, rankdir=None,
+                yes_color=None, no_color=None,
                 condition_node_params=None, leaf_node_params=None, **kwargs):
-    """Convert specified tree to graphviz instance. IPython can automatically plot the
-    returned graphiz instance. Otherwise, you should call .render() method
+    """Convert specified tree to graphviz instance. IPython can automatically plot
+    the returned graphiz instance. Otherwise, you should call .render() method
     of the returned graphiz instance.
 
     Parameters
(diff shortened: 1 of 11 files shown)


RAMitchell:
Looks like a nice reorganisation of code. We don't need any changes to the C API right, because it already accepts a string specifying format?

trivialfis:
@KOLANICH I remember sometimes ago you mentioned allowing saving model in other formats, could this be one of the building blocks for your purpose?

trivialfis:
Will try to write the production rules for text file later.

trivialfis:
@hcho3 Now `plotting.to_graphviz` in Python package returns a `graphviz.Source` instead of `graphviz.Digraph` , which means user can't modified the returned object like adding new node to the graph manually.  I can't think of a good reason to do such thing so I'm just gonna make the change.

Can you think of some use cases for modifying returned graph?  If so I will close this PR to avoid the change.

KOLANICH:
Thanks for mentioning me.

trivialfis:
Will add detailed user document in another PR.  But I won't go over language bindings for now.  Let me fix some piled up issues first.

src/tree/tree_model.cc
```diff
@@ -17,158 +22,601 @@ DMLC_REGISTER_PARAMETER(TreeParam);
...
-      fo << " }";
-    } else {
-      fo << nid << ":leaf=" << std::setprecision(float_max_precision) << tree[nid].LeafValue();
-      if (with_stats) {
-        fo << ",cover=" << std::setprecision(float_max_precision) << tree.Stat(nid).sum_hess;
-      }
-      fo << '\n';
+
+  static std::string Match(std::string const& input,
+                           std::map<std::string, std::string> replacements) {
```
hcho3: Can we consider using fmtlib (https://fmt.dev/latest/index.html) for string templating? It basically lets you use Python-like formatting in C++11. I've used it extensively for dmlc/treelite and so far I've had good experience.

```cpp
std::string result = fmt::format("{foo} {bar}",
                                 "foo"_a = "hello", "bar"_a = 32)
```

Of course, every additional dependency causes more burden, so you should decide whether Python-like formatting would be sufficiently useful to worth the cost.

hcho3: If you'd like, I can quickly make an example with fmtlib.

trivialfis: Thanks, I looked into treelite  and interested in GPU algorithm, but doesn't have the right now.  :)

I don't think I need to use fmt for this simple purpose, I benchmarked my code and it can generate a thousand trees within second.  

If you think the readability would be better with fmt, I can create something similar but simplified.  I am OK with dependencies, but reluctant to use git submodules for dependencies management ...

hcho3: @trivialfis I do think that using fmtlib would improve readability (due to Python-like syntax) and ease formatting for floating-point values (no call to `ToStr()` needed). But I'll defer to you as to whether that benefit is worth adding a new git submodule.

python-package/xgboost/plotting.py
```diff
@@ -184,64 +137,77 @@ def to_graphviz(booster, fmap='', num_trees=0, rankdir='UT',
...
-    kwargs :
-        Other keywords passed to graphviz graph_attr
+        .. code-block:: python
+
+        {'shape': 'box',
+         'style': 'filled',
+         'fillcolor': '#e48038'}
+
+    kwargs : Other keywords passed to graphviz graph_attr, E.g.:
+        `graph [ {key} = {value} ]`
```
hcho3:
```suggestion
        ``graph [ {key} = {value} ]``
```

src/tree/tree_model.cc
hcho3:
```suggestion
                           std::map<std::string, std::string> const& replacements) {
```

src/tree/tree_model.cc
```diff
@@ -17,158 +22,601 @@ DMLC_REGISTER_PARAMETER(TreeParam);
...
+    std::string result = input;
+    for (auto const& kv : replacements) {
+      auto pos = result.find(kv.first);
```
hcho3: Are we assuming that each keyword will occur only once in the template? As it is,
```cpp
Match("{foo} {foo} {bar}", { { "{foo}", "1" }, { "{bar}", "0" } })
```
will only replace the first occurrence of `{foo}` and produce `"1 {foo} 0"`.

src/tree/tree_model.cc
```diff
@@ -17,158 +22,601 @@ DMLC_REGISTER_PARAMETER(TreeParam);
...
+}  // namespace xgboost
+
+
+namespace dmlc {
+DMLC_REGISTRY_ENABLE(::xgboost::TreeIOReg);
+}  // namespace dmlc
+
+namespace xgboost {
+
+TreeGenerator* TreeGenerator::Create(std::string attrs, FeatureMap const& fmap,
```
hcho3:
```suggestion
TreeGenerator* TreeGenerator::Create(std::string const& attrs, FeatureMap const& fmap,
```

src/tree/tree_model.cc
```diff
@@ -17,158 +22,601 @@ DMLC_REGISTER_PARAMETER(TreeParam);
...
+    auto result = SuperT::Match(
+        kTreeTemplate,
+        {{"{nodes}", this->BuildTree(tree, 0, 0)}});
+    ss_ << result;
+  }
+};
+
+XGBOOST_REGISTER_TREE_IO(TextGenerator, "text")
+.describe("Dump text representation of tree")
+.set_body([](FeatureMap const& fmap, std::string attrs, bool with_stats) {
```
hcho3:
```suggestion
.set_body([](FeatureMap const& fmap, std::string const& attrs, bool with_stats) {
```

src/tree/tree_model.cc
```diff
@@ -17,158 +22,601 @@ DMLC_REGISTER_PARAMETER(TreeParam);
...
+  static DMLC_ATTRIBUTE_UNUSED ::xgboost::TreeIOReg&                    \
+  __make_ ## TreeIOReg ## _ ## UniqueId ## __ =                         \
+      ::dmlc::Registry< ::xgboost::TreeIOReg>::Get()->__REGISTER__(Name)
+
+
+class TextGenerator : public TreeGenerator {
+  using SuperT = TreeGenerator;
+
+ public:
+  TextGenerator(FeatureMap const& fmap, std::string attrs, bool with_stats) :
```
hcho3:
```suggestion
  TextGenerator(FeatureMap const& fmap, std::string const& attrs, bool with_stats) :
```

src/tree/tree_model.cc
```diff
@@ -17,158 +22,601 @@ DMLC_REGISTER_PARAMETER(TreeParam);
...
+
+DMLC_REGISTER_PARAMETER(GraphvizParam);
+
+class GraphvizGenerator : public TreeGenerator {
+  using SuperT = TreeGenerator;
+  std::stringstream& ss_;
+  GraphvizParam param_;
+
+ public:
+  GraphvizGenerator(FeatureMap const& fmap, std::string attrs, bool with_stats) :
```
hcho3:
```suggestion
  GraphvizGenerator(FeatureMap const& fmap, std::string const& attrs, bool with_stats) :
```

src/tree/tree_model.cc
```diff
@@ -17,158 +22,601 @@ DMLC_REGISTER_PARAMETER(TreeParam);
...
+         {"{indent}", Indent(depth)},
+         {"{nodes}",  tree[nid].IsLeaf() ? this->LeafNode(tree, nid, depth) :
+                                           this->SplitNode(tree, nid, depth)}});
+    return result;
+  }
+};
+
+XGBOOST_REGISTER_TREE_IO(JsonGenerator, "json")
+.describe("Dump json representation of tree")
+.set_body([](FeatureMap const& fmap, std::string attrs, bool with_stats) {
```
hcho3:
```suggestion
.set_body([](FeatureMap const& fmap, std::string const& attrs, bool with_stats) {
```

src/tree/tree_model.cc
```diff
@@ -17,158 +22,601 @@ DMLC_REGISTER_PARAMETER(TreeParam);
...
+  // Only indicator is different, so we combine all different node
+  // types into this function.
+  std::string PlainNode(RegTree const& tree, int32_t nid, uint32_t depth) override {
+    auto split = tree[nid].SplitIndex();
+    auto cond = tree[nid].SplitCond();
+    static std::string const kNodeTemplate = "    {nid} [ label=\"{label}\" {params}]\n";
+    static std::string const kLabelTemplate = R"({fname}{<}{cond})";
+
+    // Indicator only has fname.
+    bool has_less = split > fmap_.Size() || (split < fmap_.Size() && fmap_.type(split));
```
hcho3: We should also disallow the case where `split` is equal to `fmap_.Size()`.
```suggestion
    bool has_less = split >= fmap_.Size() || (split < fmap_.Size() && fmap_.type(split));
```

src/tree/tree_model.cc
trivialfis: Yes.  I will add a comment for this, if I place another loop inside then I might need to start optimizing the procedure.

For a real code generator I think treelite is the better place to go unless you want to merge the functionality into XGBoost, then I will put more thoughts into designing a tree compiler.

hcho3: @trivialfis Makes sense. Yes, you should indicate this limitation in the comment.

> For a real code generator I think treelite is the better place to go unless you want to merge the functionality into XGBoost, then I will put more thoughts into designing a tree compiler.

I agree. And I think it's best to keep Treelite as a separate project, given its complexity and scope (e.g. it supports LightGBM).

trivialfis:
@hcho3 Thanks for the review.

PR #4608, opened 2019-06-26
Fix doc for customized objective/metric

hcho3:
Follow-up to #4598. The example training log was not being displayed because of a syntax issue.

### DIFF
--- a/doc/tutorials/custom_metric_obj.rst
+++ b/doc/tutorials/custom_metric_obj.rst
@@ -122,7 +122,7 @@ parameter:
 
 We will be able to see XGBoost printing something like:
 
-.. code-block::
+.. code-block:: none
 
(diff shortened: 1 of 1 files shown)


hcho3:
Before: 
Screen Shot 2019-06-26 at 1 38 08 PM

After:
Screen Shot 2019-06-26 at 1 38 37 PM

trivialfis:
@hcho3 What's the version of sphinx used in server?  I displayed the doc fine with 2.1.2:

d

trivialfis:
I need to match my version with the one on server.

hcho3:
@trivialfis It turns out that ReadTheDocs was using an old version of Sphinx (1.8.1). Let me go ahead and change that to latest version (2.1.2).

hcho3:
See #4609

PR #4609, opened 2019-06-26
Use Sphinx 2.1+ to compile documentation

hcho3:
ReadTheDocs service (it hosts https://xgboost.readthedocs.io/en/latest/) has been previously using Sphinx 1.8.1, which is quite an old version. Upgrade Sphinx to 2.1+.

cc @trivialfis

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,4 +1,4 @@
-sphinx
+sphinx>=2.1
 mock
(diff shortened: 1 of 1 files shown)


hcho3:
Example output: https://hcho3-xgboost.readthedocs.io/en/doc_fix/tutorials/custom_metric_obj.html

PR #4610, opened 2019-06-26
Add to documentation how to run tests locally

trams:
A documentation change.
I added a chapter about how to build native unit tests and launch them to the Contributors docs

It was not at all obvious to me when I started on-boarding on this project so my guess someone else could benefit too

Tell me what do you think

### DIFF
--- a/doc/contribute.rst
+++ b/doc/contribute.rst
@@ -19,7 +19,7 @@ Everyone is more than welcome to contribute. It is a way to make the project bet
 * `Running Unit Tests Locally`_
 
   - :ref:`python_tests_pytest`
-  - `Google Tests`_
+  - `Google Test`_
   - `Running tests inside a Docker container (Recommended)`_
 
 * `Git Workflow Howtos`_
@@ -168,9 +168,9 @@ In addition, to build and test CUDA code, run:
 
   If you are running into issues running the command above, consider using our Docker container. See :ref:`running_tests_inside_docker`.
 
-Google Tests
-============
-To build and run C++ unit tests, install Google Test library with headers
+Google Test
+===========
+To build and run C++ unit tests, install `Google Test <https://github.com/google/googletest>`_ library with headers
 and then enable tests while running CMake:
 
 .. code-block:: bash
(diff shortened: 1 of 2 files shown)


hcho3:
Nice, thanks!

hcho3:
@trams I added instructions for running Python tests and for using Docker containers for convenience. Can you take a look?

trams:
Nice. I didn't yet built python tests so I would try to follow your instructions and I'll you a feedback

trams:
I think running `pytest -v -s --fulltrace tests/python` requires building|installing a python package which is not done for some reason on my machine. I will investigate this tomorrow and come back to you

hcho3:
@trams I think you should set `PYTHONPATH` so that Python can find the XGBoost package.

trams:
I tried running python tests one more time.
One need to install few packager for them to pass but otherwise it works

I tried running unit tests in Docker.
Native one worked like a charm
Python didn't work until I manually create a wheel using setup.py
After I created they mostly worked (5 failed)

trams:
As for the structure.
Given that you recommend to run tests in Docker then I suggest to put the chapter about running them in Docker on top. What do you think?

trams:
Thank you so much for improving dramatically my tiny pull request. Thanks to your addition I was able to start poking around the python part of this project

hcho3:
> I tried running unit tests in Docker.
Native one worked like a charm
Python didn't work until I manually create a wheel using setup.py
After I created they mostly worked (5 failed)

I found a small problem with `test_python.sh` script. After the fix, you should be able to run Python tests without having to build a wheel.

hcho3:
@trams I found out why the 5 tests were failing. I will submit a pull request soon.

hcho3:
> Given that you recommend to run tests in Docker then I suggest to put the chapter about running them in Docker on top. What do you think?

I think it's still good to explain what individual tests do before showing Docker commands.

hcho3:
@trams dmlc/xgboost#4617 should fix the 5 tests that failed.

PR #4614, opened 2019-06-27
remove caching allocator/vector and fix issues relating to inability to train large datasets

sriramch:
- the caching device allocator caches device allocations in bins of different sizes, which can be reused for future allocations. however, this has regressed on xgboost's capability to train on large datasets
- device memory cached by this allocator cannot be seen by other non caching allocators
- as a consequence, types that use regular allocators (device_malloc_allocator and such) will attempt to carve out more device memory when needed, eventhough plenty may be available for use inside the caching allocators
- bottom line: this kind of allocator choice can't be made in a piecemeal fashion

i have thus resorted to using a consistent allocator type across for now and fixed issues arising out of additional allocations introduced by #4554. in the process, the unit test failures have been fixed as well (it was hanging on tesla t4's earlier).

ideally, the core stock xgboost should use the regular device allocators, and use-cases that need sophisticated (de)allocation techniques should dynamically load device allocators via `LD_PRELOAD` or some dll injection mechanism on windows, quite similar to how one can replace the glibc allocator with jemalloc/tcmalloc/vespamalloc and its likes.

it is also possible to use a caching device allocator across entire xgboost; but first, its implications have to be well understood, including, but not limited to how the different allocation bins can be configured from within xgboost for different use-cases. from what i can see from the code, it uses (whose impact on performance is unknown at present):

- a global mutex across *all* devices to retrieve and to cache allocations
- it sweeps through pertinent bin to release all free blocks and retrying, when a request for device allocation can't be satisfied, which can be fairly expensive
- it pads every device allocation request to a bin that geometrically progresses by the growth factor, thus entailing significant wastage (more for bigger device memory chunks)
 
my naive attempt to alias all `device_vector` types to a `caching_device_vector` failed as well (not investigated on this yet).

@RAMitchell @rongou - please review...

### DIFF
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -310,11 +310,11 @@ struct XGBDefaultDeviceAllocatorImpl : thrust::device_malloc_allocator<T> {
   };
   pointer allocate(size_t n) {
     pointer ptr = super_t::allocate(n);
-    GlobalMemoryLogger().RegisterAllocation(ptr.get(), n * sizeof(T));
+    GlobalMemoryLogger().RegisterAllocation(ptr.get(), n);
     return ptr;
   }
   void deallocate(pointer ptr, size_t n) {
-    GlobalMemoryLogger().RegisterDeallocation(ptr.get(), n *sizeof(T));
+    GlobalMemoryLogger().RegisterDeallocation(ptr.get(), n);
     return super_t::deallocate(ptr, n);
   }
 };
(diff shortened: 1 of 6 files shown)


rongou:
Probably shouldn't include rabit changes?

src/common/device_helpers.cuh
```diff
@@ -310,11 +310,11 @@ struct XGBDefaultDeviceAllocatorImpl : thrust::device_malloc_allocator<T> {
   };
   pointer allocate(size_t n) {
     pointer ptr = super_t::allocate(n);
-    GlobalMemoryLogger().RegisterAllocation(ptr.get(), n);
+    GlobalMemoryLogger().RegisterAllocation(ptr.get(), n * sizeof(T));
     return ptr;
   }
   void deallocate(pointer ptr, size_t n) {
-    GlobalMemoryLogger().RegisterDeallocation(ptr.get(), n);
+    GlobalMemoryLogger().RegisterDeallocation(ptr.get(), n *sizeof(T));
```
rongou: Needs space.

src/common/device_helpers.cuh
```diff
@@ -334,19 +334,19 @@ struct XGBCachingDeviceAllocatorImpl : thrust::device_malloc_allocator<T> {
...
                                                 n * sizeof(T));
-     pointer thrust_ptr = thrust::device_ptr<T>(ptr);
-     GlobalMemoryLogger().RegisterAllocation(thrust_ptr.get(), n);
+     pointer thrust_ptr(ptr);
+     GlobalMemoryLogger().RegisterAllocation(thrust_ptr.get(), n * sizeof(T));
      return thrust_ptr;
    }
    void deallocate(pointer ptr, size_t n) {
-     GlobalMemoryLogger().RegisterDeallocation(ptr.get(), n);
+     GlobalMemoryLogger().RegisterDeallocation(ptr.get(), n *sizeof(T));
```
rongou: space

sriramch:
let me reopen another one. i do not know how those rabit related changes ended up here.

PR #4615, opened 2019-06-27
- remove caching allocator/vector and fix issues relating to inability to train large datasets

sriramch:
- the caching device allocator caches device allocations in bins of different sizes, which can be reused for future allocations. however, this has regressed on xgboost's capability to train on large datasets
- device memory cached by this allocator cannot be seen by other non caching allocators
- as a consequence, types that use regular allocators (device_malloc_allocator and such) will attempt to carve out more device memory when needed, eventhough plenty may be available for use inside the caching allocators
- bottom line: this kind of allocator choice can't be made in a piecemeal fashion

i have thus resorted to using a consistent allocator type across for now and fixed issues arising out of additional allocations introduced by #4554. in the process, the unit test failures have been fixed as well (it was hanging on tesla t4's earlier).

ideally, the core stock xgboost should use the regular device allocators, and use-cases that need sophisticated (de)allocation techniques should dynamically load device allocators via `LD_PRELOAD` or some dll injection mechanism on windows, quite similar to how one can replace the glibc allocator with jemalloc/tcmalloc/vespamalloc and its likes.

it is also possible to use a caching device allocator across entire xgboost; but first, its implications have to be well understood, including, but not limited to how the different allocation bins can be configured from within xgboost for different use-cases. from what i can see from the code, it uses (whose impact on performance is unknown at present):

- a global mutex across *all* devices to retrieve and to cache allocations
- it sweeps through pertinent bin to release all free blocks and retrying, when a request for device allocation can't be satisfied, which can be fairly expensive
- it pads every device allocation request to a bin that geometrically progresses by the growth factor, thus entailing significant wastage (more for bigger device memory chunks)
 
my naive attempt to alias all `device_vector` types to a `caching_device_vector` failed as well (not investigated on this yet).

@RAMitchell @rongou - please review...

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,4 +1,4 @@
-Checks: 'modernize-*,-modernize-make-*,-modernize-use-auto,-modernize-raw-string-literal,google-*,-google-default-arguments,-clang-diagnostic-#pragma-messages,readability-identifier-naming'
+Checks: 'modernize-*,-modernize-make-*,-modernize-use-auto,-modernize-raw-string-literal,-modernize-avoid-c-arrays,google-*,-google-default-arguments,-clang-diagnostic-#pragma-messages,readability-identifier-naming'
 CheckOptions:
   - { key: readability-identifier-naming.ClassCase,           value: CamelCase  }
   - { key: readability-identifier-naming.StructCase,          value: CamelCase  }
--- a/R-package/DESCRIPTION
+++ b/R-package/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: xgboost
 Type: Package
 Title: Extreme Gradient Boosting
-Version: 0.82.0.1
-Date: 2019-03-11
+Version: 0.90.0.1
+Date: 2019-05-18
 Authors@R: c(
   person("Tianqi", "Chen", role = c("aut"),
          email = "tianqi.tchen@gmail.com"),
--- a/R-package/R/xgb.cv.R
+++ b/R-package/R/xgb.cv.R
@@ -133,8 +133,15 @@ xgb.cv <- function(params=list(), data, nrounds, nfold, label = NULL, missing = 
 
   # Check the labels
   if ( (inherits(data, 'xgb.DMatrix') && is.null(getinfo(data, 'label'))) ||
-       (!inherits(data, 'xgb.DMatrix') && is.null(label)))
+       (!inherits(data, 'xgb.DMatrix') && is.null(label))) {
     stop("Labels must be provided for CV either through xgb.DMatrix, or through 'label=' when 'data' is matrix")
+  } else if (inherits(data, 'xgb.DMatrix')) {
+    if (!is.null(label))
+      warning("xgb.cv: label will be ignored, since data is of type xgb.DMatrix")
+    cv_label = getinfo(data, 'label')
+  } else {
+    cv_label = label
+  }
 
   # CV folds
   if(!is.null(folds)) {
@@ -144,7 +151,7 @@ xgb.cv <- function(params=list(), data, nrounds, nfold, label = NULL, missing = 
   } else {
     if (nfold <= 1)
       stop("'nfold' must be > 1")
-    folds <- generate.cv.folds(nfold, nrow(data), stratified, label, params)
+    folds <- generate.cv.folds(nfold, nrow(data), stratified, cv_label, params)
   }
 
   # Potential TODO: sequential CV
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -191,6 +191,20 @@ test_that("xgb.cv works", {
   expect_false(is.null(cv$call))
 })
 
+test_that("xgb.cv works with stratified folds", {
+  dtrain <- xgb.DMatrix(train$data, label = train$label)
+  set.seed(314159)
+  cv <- xgb.cv(data = dtrain, max_depth = 2, nfold = 5,
+               eta = 1., nthread = 2, nrounds = 2, objective = "binary:logistic",
+               verbose=TRUE, stratified = FALSE)
+  set.seed(314159)
+  cv2 <- xgb.cv(data = dtrain, max_depth = 2, nfold = 5,
+                eta = 1., nthread = 2, nrounds = 2, objective = "binary:logistic",
+                verbose=TRUE, stratified = TRUE)
+  # Stratified folds should result in a different evaluation logs
+  expect_true(all(cv$evaluation_log[, test_error_mean] != cv2$evaluation_log[, test_error_mean]))
+})
+
 test_that("train and predict with non-strict classes", {
   # standard dense matrix input
   train_dense <- as.matrix(train$data)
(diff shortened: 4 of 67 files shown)


RAMitchell:
I'm not convinced there is anything wrong with using caching allocators in general. I think as long as you don't use them for too large allocations it should be fine. The benefit from a software architecture perspective is significant and I want to extend their use across xgboost to create useful data structures. I also don't see how using caching allocators and standard allocators are mutually exclusive.

Why did the caching allocator in RowPartitioner cause such a regression in memory performance? It is allocating memory proportional to the number of rows which I would expect to be considerably less than the dataset size. Did the memory allocated for this purpose get successfully reused or did something go wrong leading to it being allocated multiple times? Was the memory allocated inefficiently used due to the bin sizes being a power of 2? There are some configurable options in the allocator I would like to explore first.

The outcome of this may be that we cannot use the caching allocator for memory sizes this large, but that does not mean this will not be useful in a lot of other areas.

```a global mutex across all devices to retrieve and to cache allocations
it sweeps through pertinent bin to release all free blocks and retrying, when a request for device allocation can't be satisfied, which can be fairly expensive```

Do you have any performance measurements showing that this actually impacts runtime? Standard cudaMalloc is also expensive as it needs to manage a heap of unified memory across devices.

**Comment by sriramch** [comment]
> I'm not convinced there is anything wrong with using caching allocators in general. I think as long as you don't use them for too large allocations it should be fine. 

i wasn't alluding to that as such. my broader point was that it can't be done in isolation (more on this a bit later). 

> I also don't see how using caching allocators and standard allocators are mutually exclusive.

this issue exemplifies this. the caching allocators have usable memory which can't be accessed by others. this becomes more imperative, when there are layers above xgboost. for instance, spark may impose its own memory requirements on the gpu (during data parsing/processing), and it may reason that what they may require is held up inside this. thus, it may be useful to use a single allocator that different components may use, and provide a configurable means within xgboost to use one.

> Why did the caching allocator in RowPartitioner cause such a regression in memory performance? It is allocating memory proportional to the number of rows which I would expect to be considerably less than the dataset size. Did the memory allocated for this purpose get successfully reused or did something go wrong leading to it being allocated multiple times? Was the memory allocated inefficiently used due to the bin sizes being a power of 2? 

see the change in the reset method. when a `unique_ptr` is reset with a new instance, the existing ptr isn't ~ first (the argument to reset is evaluated first). the new instance will first try to allocate more memory (and fail), as the soon to be ~ instance still holds them needlessly.

second, parts of xgboost were not using a caching allocator. hence, even when row partitioner instances are ~, creation of device vectors can still fail (as they use a regular allocators), and needed memory is being held in the internal caches of the caching allocator.

> There are some configurable options in the allocator I would like to explore first.

this may have to be empirically determined for different use-cases. further, it still does not solve this issue, unless we force this type of allocator throughout due to reasons mentioned above.

> 
> The outcome of this may be that we cannot use the caching allocator for memory sizes this large, but that does not mean this will not be useful in a lot of other areas.

certainly, and that wasn't the point i was making. i'm sorry if my earlier comments gave this impressions.

> 
> > a global mutex across all devices to retrieve and to cache allocations
> > it sweeps through pertinent bin to release all free blocks and retrying, when a request for device allocation can't be satisfied, which can be fairly expensive
> 
> Do you have any performance measurements showing that this actually impacts runtime? Standard cudaMalloc is also expensive as it needs to manage a heap of unified memory across devices.

since i couldn't train larger datasets with this, i couldn't really asses its impacts. on smaller ones that i tried there wasn't that big of a difference. the numbers between using a regular vs caching allocators were fluctuating quite a bit.

**Comment by RAMitchell** [comment]
> see the change in the reset method. when a unique_ptr is reset with a new instance, the existing ptr isn't ~ first (the argument to reset is evaluated first). the new instance will first try to allocate more memory (and fail), as the soon to be ~ instance still holds them needlessly.

This explains a lot thanks. Are you able to train successfully without memory issues with only this fix? I would still like to use the caching allocator in the row partitioner if possible. I am trying to get away from maintaining persistent memory buffers and instead clearly create and destroy objects as they are needed.

The general idea is that the caching allocator should not be grabbing significant amounts of memory, only "working memory" type allocations. This would allow easily designing dynamic data structures that can grow and change or that need temporary memory buffers. As an example think of a decision tree representation on the device that needs to dynamically expand as new nodes are added. We currently deal with these kind of situations by allocating up front the largest possible tree, but this is not a flexible approach.

**Comment by sriramch** [comment]
> > see the change in the reset method. when a unique_ptr is reset with a new instance, the existing ptr isn't ~ first (the argument to reset is evaluated first). the new instance will first try to allocate more memory (and fail), as the soon to be ~ instance still holds them needlessly.
> 
> This explains a lot thanks. Are you able to train successfully without memory issues with only this fix?

no - due to reasons mentioned earlier. this is necessary, but not sufficient.

> The general idea is that the caching allocator should not be grabbing significant amounts of memory, only "working memory" type allocations. 

for larger datasets, it will end up carving out larger chunks. say 500m rows will require 2gb just for the row indices (with double buffers, positional indices etc., this will quickly add up)

> As an example think of a decision tree representation on the device that needs to dynamically expand as new nodes are added. We currently deal with these kind of situations by allocating up front the largest possible tree, but this is not a flexible approach.

well, having the allocator maintain a scarce and an expensive resource that other parts needs will not work either. i started going down the rat hole of using a caching allocator throughout. however, the tests failed right off the bat (and i didn't investigate why). further, doing this will also require exposing some configurable hooks for this allocator, as it cannot be statically determined, as there may not be a one-size fits all approach.

perhaps, one approach could be to create a caching allocator at the beginning of the training and release it soon after. this way it can be customized and localized just for this use-case. i haven't really tried or thought about this - just something that came off the top of my head while writing this.

**Comment by RAMitchell** [comment]
After looking at it a little more, I think the bin growth = 8 factor is the problem. This means any allocation could use up to 8x memory in the worst case. Try changing the parameters as follows:
```
static cub::CachingDeviceAllocator *allocator = new cub::CachingDeviceAllocator(2, 9, 29);
```
This would be growing by a factor of 2x for each bin with a minimum of 512 bytes, and a maximum of around 500mb.

sriramch:
> 
> ```
> static cub::CachingDeviceAllocator *allocator = new cub::CachingDeviceAllocator(2, 9, 29);
> ```
> 
> This would be growing by a factor of 2x for each bin with a minimum of 512 bytes, and a maximum of around 500mb.

by having global allocators such as this, gpu memory will be held up inside xgboost, which no one will be able to tap into. users *may* want to do post processing on gpu after training, and they will have no way to access this latent memory. assume i'm training 100m rows on a gpu. with this kind of allocation strategy, we'll be leaving ~ 2gb of gpu memory on the side - `(100m rows * 4 bytes - rounded to 512m) * 4`  (assuming no/negligible additional overhead for the container to carry this data) for the row and positional indices. we could play with these numbers; but lesser the growth factor, or capping the number of bins, the greater the chance that this allocator becomes a no-op, as larger allocation sizes results in not reusing previous chunks.

if we restrict this to just the row partitioner, what do you think about creating a caching allocator at the beginning of the training session and freeing it up right after to keep this localized? the allocations are still reused within a training session, but not across?

RAMitchell:
The tree updater class is actually never deleted or cleaned up until the booster class is deleted. This has always been a problem - the internal c++ code has no way to know that training is "finished". Only the language wrapper knows. I would be interested in solving this but I don't have any ideas at the moment.

sriramch:
@RAMitchell please take 1 more look...

PR #4629, opened 2019-07-02
sum_weights by passes last group

gdupret:
the weight of the last group is never added to the total sum

### DIFF
--- a/src/objective/rank_obj.cc
+++ b/src/objective/rank_obj.cc
@@ -63,7 +63,7 @@ class LambdaRankObj : public ObjFunction {
       std::vector<ListEntry>  lst;
       std::vector< std::pair<bst_float, unsigned> > rec;
       bst_float sum_weights = 0;
-      for (bst_omp_uint k = 0; k < ngroup; ++k) {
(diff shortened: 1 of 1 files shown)


src/objective/rank_obj.cc
```diff
@@ -63,7 +63,7 @@ class LambdaRankObj : public ObjFunction {
       std::vector<ListEntry>  lst;
       std::vector< std::pair<bst_float, unsigned> > rec;
       bst_float sum_weights = 0;
-      for (bst_omp_uint k = 0; k < ngroup; ++k) {
+      for (bst_omp_uint k = 0; k <= ngroup; ++k) {
```
hcho3: Indexing in C++ is 0-based, so we want to iterate from 0 to `ngroups - 1`.

gdupret: const auto ngroup = static_cast<bst_omp_uint>(gptr.size() - 1);

I suspect this is related to the reference to gprt[k+1] line 74

hcho3: `gptr` vector is defined to have length `[number of groups] + 1`.

Issue #4517, opened 2019-05-30
Multiclass classification load model problem


Leonhoohoo:
I trained six-class model and predict my training data .
Then, I save the model.
But there has some problem when I load the model and predict the training data.
The result is wrong.
 
I test the two-class model is okay.
But  multiclass has some problem when I save the model then predict.

```
clf = xgb.XGBClassifier(n_jobs=-1)
model = clf.fit(x_train, y_train)
x = model.predict_proba(x_train)
model.save_model('tmp_model/0013')
print(x)

array([[0.04758353, 0.01750289, 0.5136149 , 0.06486283, 0.04118178,
        0.31525403],
       [0.01309447, 0.4156407 , 0.38199472, 0.0167549 , 0.12109441,
        0.0514208 ],
       [0.01633686, 0.19335738, 0.4924989 , 0.04608943, 0.18769601,
        0.06402139],
       ...,
       [0.14344662, 0.07964588, 0.02425297, 0.68924797, 0.03783285,
        0.02557368],
       [0.42988327, 0.11303075, 0.09305003, 0.28722036, 0.03225137,
        0.04456424],
       [0.00883721, 0.47754583, 0.0562663 , 0.01131444, 0.30832848,
        0.13770771]], dtype=float32)


--------------------------------------------------------------------------
# load model and predict

clf = xgb.XGBClassifier(n_jobs=-1)
booster = xgb.Booster()
booster.load_model(f'tmp_model/0013')
clf._Booster = booster
pred = clf.predict_proba(x_train)
print(pred)

array([[0.9524165 , 0.9869055 , 0.98366314, ..., 0.14344662, 0.42988327,
        0.00883721],
       [0.9824971 , 0.5843593 , 0.80664265, ..., 0.07964588, 0.11303075,
        0.47754583],
       [0.4863851 , 0.6180053 , 0.5075011 , ..., 0.02425297, 0.09305003,
        0.0562663 ],
       [0.93513715, 0.9832451 , 0.9539106 , ..., 0.68924797, 0.28722036,
        0.01131444],
       [0.9588182 , 0.8789056 , 0.812304  , ..., 0.03783285, 0.03225137,
        0.30832848],
       [0.68474597, 0.9485792 , 0.9359786 , ..., 0.02557368, 0.04456424,
        0.13770771]], dtype=float32)
```
Please, help me.
Thanks

trivialfis:
@Leonhoohoo Could you provide the XGBoost version or commit hash?

[...]

PR #4632, opened 2019-07-03
Add warning when save_model() is called from scikit-learn interface

hcho3:
Many methods of `XGBClassifier` relies on the existence of certain fields, such as `self.objective`, `self.missing`, and `self.n_classes_` (these fields exist to ensure interoperability with scikit-learn) and these fields are **not** preserved by the call to `XGBClassifier.save_model()`, as the binary serialization format is not aware of these fields. See #4517, where `XGBClassifier.save_model()` breaks `predict_proba()` for multi-class classifiers.

Fix: Display prominent warning whenever `save_model()` is called from `XGBClassifier`. The recommendation is not to use it, unless the purpose is to export the model to a non-Python environment (e.g. Java).

EDIT. Maybe we should just deprecate `save_model()`, as its name is quite misleading, and call it `export_model_for_external_use()` or something like that.

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -278,6 +278,11 @@ class XGBModel(XGBModelBase):
         fname : string
             Output file name
         """
+        warnings.warn("save_model: Useful attributes in the Python " +
+                      "object {} will be lost. ".format(type(self).__name__) +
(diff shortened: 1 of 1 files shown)


hcho3:
I opened a new issue: #4639

PR #4633, opened 2019-07-03
Empty evaluation list in early stopping should produce meaningful error message

### LINKED ISSUE #4443: Early Stop in Python (IndexError: list index out of range)
https://github.com/dmlc/xgboost/blob/5de7e12704b38d81a46471e02a791d4c586a7286/python-package/xgboost/callback.py#L225

While using `TimeSeriesSplit` as cv for `GridSearchCV`,  `early_stop.callback` function may receive empty `evaluation_result_list`.
When it occur error is raised (`IndexError: list index out of range`)

hcho3:
Closes #4443, by ignoring iterations with empty evaluation result.

### DIFF
--- a/python-package/xgboost/callback.py
+++ b/python-package/xgboost/callback.py
@@ -222,7 +222,11 @@ def early_stop(stopping_rounds, maximize=False, verbose=True):
 
     def callback(env):
         """internal function"""
-        score = env.evaluation_result_list[-1][1]
+        if not env.evaluation_result_list:
+            score = float('-inf') if maximize_score else float('inf')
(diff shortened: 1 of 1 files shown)


hcho3:
@trivialfis Okay, let me add an informative error message instead.

hcho3:
There is actually an informative error message already: https://github.com/dmlc/xgboost/pull/4633/files#diff-0b3dbbf617dc3d1d93080d0f54b584e1L184

We just need to ensure that the check is performed before `env.evaluation_result_list[-1]` is accessed.

PR #4638, opened 2019-07-04
Fix early stopping in the Python package

### LINKED ISSUE #4421: custom evaluation function is not used if name is lexicographically smaller than default
Dear xgboost developers,
I have found the following issue.
When I try to use a custom evaluation function, it is only working when I either use parameter disable_default_eval_metric or choose a name for my evaluation metric that is lexicographically larger than the name of the default metric.
If you know these workarounds, they are acceptable, but it took me some time to find out how to overcome the issue. Hence, it would be nice if this bug could be fixed.
I have attached a jupyter notebook that can be used to reproduce the bug. I needed to zip it to be able to upload it -:)

[xgboost_bug.zip](https://github.com/dmlc/xgboost/files/3130499/xgboost_bug.zip)

### LINKED ISSUE #4630: xgb.cv with early stopping returns wrong best iteration when multiple metrics are used
I've been using xgb.cv with early stopping to determine the best number of training rounds. The [documentation for early-stopping](https://xgboost.readthedocs.io/en/latest/python/python_intro.html#early-stopping) only mentions `xgb.train` and `param['eval_metric']` and says that 

> if you specify more than one evaluation metric the last one in param['eval_metric'] is used for early stopping

I assumed that the same would be true for `xgb.cv` and the parameter `metrics`. However, when using multiple metrics it does not return the correct number for the best iteration.

Following minimal code example demonstrates the encountered problem:

```
import sklearn.datasets as datasets
import xgboost as xgb

print(xgb.__version__)

X, y = datasets.load_breast_cancer(return_X_y=True)
train = xgb.DMatrix(X, y)

params = {'objective':'binary:logistic'}

metrics = [['auc'], ['error'], ['logloss'], ['logloss', 'auc'], ['logloss', 'error'], ['error', 'logloss']]
for i, m in enumerate(metrics):
    result = xgb.cv(params, train, 1000, 5, stratified=True, metrics=m, early_stopping_rounds=50, seed=42)
    print(f'\n{i+1}: metrics = {m}\n', result.tail(1)[[f"test-{x}-mean" for x in m]])
```
Output:

```
0.82

1: metrics = ['auc']
      test-auc-mean
170       0.993797

2: metrics = ['error']
     test-error-mean
69         0.031612

3: metrics = ['logloss']
     test-logloss-mean
64           0.096297

4: metrics = ['logloss', 'auc']
     test-logloss-mean  test-auc-mean
64           0.096297       0.993461

5: metrics = ['logloss', 'error']
     test-logloss-mean  test-error-mean
64           0.096297         0.033382

6: metrics = ['error', 'logloss']
     test-error-mean  test-logloss-mean
64         0.033382           0.096297
```

In cases 1-3 with a single metric we get different values for the best number of iteration, i.e. 170, 69, 64, respectively. Cases 4-6 have two metrics where the second metric is the same as in cases 1-3. Therefore, the expected values for the best number of iterations should be the same as in 1-3. Instead, it's always 64 - which corresponds to the best number of iterations of the 'logloss' metric.

hcho3:
* Consistently choose the last validation set + last metric combination as the early stopping criteria.
* The order between metrics should be preserved, so that the user will have certainty of which metric will be used for early stopping.
* Clarify the semantics of early stopping in the Python API doc. Preview: https://hcho3-xgboost.readthedocs.io/en/fix_early_stopping_python/python/python_api.html#xgboost.train

Closes #4630
Closes #4421

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -325,7 +325,7 @@ class XGBModel(XGBModelBase):
             If a str, should be a built-in evaluation metric to use. See
             doc/parameter.rst.
             If a list of str, should be the list of multiple built-in evaluation metrics
-            to use. 
+            to use.
             If callable, a custom evaluation metric. The call
             signature is ``func(y_predicted, y_true)`` where ``y_true`` will be a
             DMatrix object such that you may need to call the ``get_label``
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -304,8 +304,7 @@ class XGBModel(XGBModelBase):
             early_stopping_rounds=None, verbose=True, xgb_model=None,
             sample_weight_eval_set=None, callbacks=None):
         # pylint: disable=missing-docstring,invalid-name,attribute-defined-outside-init
-        """
-        Fit the gradient boosting model
+        """Fit gradient boosting model
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


hcho3:
In the future, I may want to overhaul how the evaluation metrics are passed from C++ to Python/R/Java. Currently, the C++ function `XGBoosterEvalOneIter()` is returning a string consisting of metric names and values all combined. This means that Python/R/Java bindings have to parse the returned string. It may be better to have the C++ function return a structured object instead.

hcho3:
@sriramch @rongou

sriramch:
> Currently, the C++ function `XGBoosterEvalOneIter()` is returning a string consisting of metric names and values all combined. This means that Python/R/Java bindings have to parse the returned string. It may be better to have the C++ function return a structured object instead (something like a key-value store).

@hcho3 are there plans to use protobuf or similar serialized data formats for things such as this? this may also have a nice side effect of supporting compatibility when the return data evolves, as opposed to a binary structs.

trivialfis:
@sriramch There is a RFC of using JSON for everything.  But currently I'm not sure about its priority.

PR #4645, opened 2019-07-08
Fix auc error in distributed mode caused by imbalanced dataset

sperlingxx:
Running xgboost distributedly with imbalanced datasets may trigger the exception: `AUC: the dataset only contains pos or neg samples` ,  especially when world_size is large.  Because sub datasets held by some workers only contains no pos or neg samples,  which will cause local exceptions.

This PR modified the response of `auc_error`.  In distribute mode, workers whose `auc_error > 0` will be ignored in AUC average rather than raising an error.

### DIFF
--- a/src/metric/rank_metric.cc
+++ b/src/metric/rank_metric.cc
@@ -197,18 +197,21 @@ struct EvalAuc : public Metric {
       // this is the AUC
       sum_auc += sum_pospair / (sum_npos * sum_nneg);
     }
-    CHECK(!auc_error)
-      << "AUC: the dataset only contains pos or neg samples";
-    /* Report average AUC across all groups */
-    if (distributed) {
-      bst_float dat[2];
+
+    // Report average AUC across all groups
+    // In distributed mode, workers which only contains pos or neg samples
+    // will be ignored when aggregate AUC.
+    bst_float dat[2] = {0.0f, 0.0f};
+    if (!auc_error) {
       dat[0] = static_cast<bst_float>(sum_auc);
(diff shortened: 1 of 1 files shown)


trivialfis:
So the fix is by delaying checks to after reduction?  LGTM.

sperlingxx:
> So the fix is by delaying checks to after reduction? LGTM.

Yes, after this PR,  the exception(`AUC: the dataset only contains pos or neg samples`) will be thrown only when there is no pos or neg samples in global dataset.

trivialfis:
Thanks for the fix.

hcho3:
@sperlingxx This fix is good for now. However, in the long term, we should investigate a more robust method for computing AUC in distributed setting. The current approach, where AUC is computed in each local partition and then averaged via AllReduce, may introduce significant bias compared to the exact global AUC, since we are essentially creating a noisy estimate of AUC based on a single local partition. The AUC estimate may be inaccurate if one partition gets a lion's share of the positive label, or if the positive label is highly rare across the board.

The robust approach would use the approximate quantile sketches instead, so that we have some theoretical bounds on our AUC estimates.

tinkle1129:
sorry to disturb you.  
There was a dataset for ranking task,  and most samples are negative. I tried to use train it in distributed mode and found that all workers contains at least one group which had only pos or neg samples. As the result, I got the exception: AUC: the dataset only contains pos or neg samples.
Besides, I found that "auc_error" would be changed into "1" if there was a group full of negative or positive samples in each worker.

trivialfis:
I think there's a fix for the exception.  Not such about the second issue yet.

PR #4652, opened 2019-07-10
[python-package] ctypes2numpy supports more dtypes

sperlingxx:


trivialfis:
Is XGBoost returning these types from C APIs?

sperlingxx:
@trivialfis  So far, there isn't. I want to make it more generally, so it can be used as a general util.

trivialfis:
I will merge it when there's an actual use case.  Otherwise XGBoost returning one of added types is considered as a bug.  Also I have been trying to narrow the range of used types, dynamically casting types is dangerous and expensive.

sperlingxx:
@trivialfis  Fine, I closed this.

trivialfis:
@sperlingxx Thanks for the effort.

PR #4655, opened 2019-07-10
Improve setup.py by adding extra dependencies

trams:
I did some python code reading and I noticed that we have a lot of optional dependencies which are not fully exposed though setup.py.

So I added extra dependencies.
Now one can use

  pip install -e .[dask]

to install all xgboost supporting dask

### DIFF
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -43,6 +43,13 @@ setup(name='xgboost',
           'numpy',
           'scipy',
       ],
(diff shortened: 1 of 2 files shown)


hcho3:
@trivialfis Done. Note that currently, setup_pip.py is exclusively used for source distribution (sdist).

trivialfis:
@trams @hcho3 Thanks.

PR #4658, opened 2019-07-12
WIP: skeleton code for saving compressed features to disk

rongou:
Initial idea on how to save the compressed features to disk. Haven't worked out all the details yet, so not sure if it'll work out, please let me know what you think.

@RAMitchell @sriramch @trivialfis

### DIFF
--- a/src/data/simple_dmatrix.cc
+++ b/src/data/simple_dmatrix.cc
@@ -81,12 +81,12 @@ BatchSet SimpleDMatrix::GetSortedColumnBatches() {
   return BatchSet(begin_iter);
 }
 
-BatchSet SimpleDMatrix::GetTransformedBatches(Transformer* transformer){
+BatchSet SimpleDMatrix::GetTransformedBatches(Transformer* transformer) {
   // Transformed page doesn't exist, generate it
   if (!transformed_page_) {
     transformer->Init(this);
-    auto cast = dynamic_cast<SimpleCSRSource*>(source_.get());
-    auto page = transformer->Transform(cast->page_);
+    auto source = dynamic_cast<SimpleCSRSource*>(source_.get());
+    auto page = transformer->Transform(source->page_);
     transformed_page_.reset(&page);
   }
   auto begin_iter = BatchIterator(new SimpleBatchIteratorImpl(transformed_page_.get()));
--- a/src/data/simple_dmatrix.cc
+++ b/src/data/simple_dmatrix.cc
@@ -81,6 +81,18 @@ BatchSet SimpleDMatrix::GetSortedColumnBatches() {
   return BatchSet(begin_iter);
 }
 
+BatchSet SimpleDMatrix::GetTransformedBatches(Transformer* transformer){
+  // Transformed page doesn't exist, generate it
+  if (!transformed_page_) {
+    transformer->Init(this);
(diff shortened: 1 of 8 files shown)


trivialfis:
The prototype looks good to me.  Just note that, if you think the existing structure is not suitable or too complex, you can roll out your own.  I don't understand how the current external memory works anyway.

src/data/simple_dmatrix.cc
```diff
@@ -81,6 +81,18 @@ BatchSet SimpleDMatrix::GetSortedColumnBatches() {
   return BatchSet(begin_iter);
 }
 
+BatchSet SimpleDMatrix::GetTransformedBatches(Transformer* transformer){
+  // Transformed page doesn't exist, generate it
+  if (!transformed_page_) {
+    transformer->Init(this);
+    auto cast = dynamic_cast<SimpleCSRSource*>(source_.get());
```
sriramch: can we call the return value as `source?`

sriramch:
i'm also curious as to how and where these transformed (compressed) pages are going to be used? is this going to be used during binning, used in lieu of the compressed bins (which is where the biggest gpu memory consumption is) etc.? it *may* be prohibitively expensive if we are bringing in these pages into gpu during tree construction batch by batch to compute the feature quantile bin the feature belongs to, even if we do this only when external memory is enabled.

rongou:
My current plan is in `Transformer::Init(DMatrix* dmat)` we'll loop through the dmatrix once to generate the quantiles and keep them in memory. Then the `DMatrix::GetTransformedBatches()` will loop through the row batches, compress each batch via `Transformer::Transform(batch)`, and write the compressed features to disk. During tree construction, the compressed features will be streamed back to GPU.

Yes it'll probably be very slow, but I think it's a necessary waypoint towards the full solution. We can then look at overlapping data copying and computation, importance sampling etc.

sriramch:
would it be easier (and possibly quicker) to assess the impact of streaming these into gpu in batchwise fashion like this, or to even assess if such a thing is possible for tree construction, without doing this dmatrix refactoring first? the compressed feature bins for the entire data is in gpu today. it should be possible to copy them to host memory and to bring them in batches - right?

rongou:
@sriramch if you have some time to run the experiment, that would be great!

I'm not really refactoring dmatrix; it's more leveraging the existing code to make it possible to support external memory mode on gpus. I'll keep pursuing this to see where that leads us.

hcho3:
@rongou Take a look at https://xgboost.readthedocs.io/en/latest/contribute.html#running-formatting-checks-locally to run formatting checks locally.

src/data/simple_dmatrix.cc
rongou: Done.

RAMitchell:
The transformer class looks like the [visitor design pattern](https://en.wikipedia.org/wiki/Visitor_pattern). I like this idea.

What I don't understand is how the transformer class can deal with compressed matrix representations using integer features. It seems to only be able to transform a SparsePage into another SparsePage, which is strictly a CSR (or CSC if transposed) floating point format.

rongou:
Yeah I think we need to have different types of SparsePages. I have some idea to make this work, will send out an update.

rongou:
I think I have a path forward now. Closing this and will send a separate PR.

PR #4646, opened 2019-07-08
contribute to community doc

CodingCat:

PR #4659, opened 2019-07-15
Re-organize contributor's guide

hcho3:
Follow-up to #4646.

* Expose the new community guideline `contrib/` to the global Table of Contents.
* Split the existing contributor's guide (`contribute.rst`) into multiple pages, as it's gotten quite too long.


hcho3:
@sriramch @rongou @trams @thvasilo Would you take a look at the new contributor's guide? See the rendered output at https://hcho3-xgboost.readthedocs.io/en/refactor_contribute_doc/contrib/index.html.

doc/contrib/unit_tests.rst
```diff
@@ -0,0 +1,232 @@
...
+
+  If you are having difficulty running the commands below (e.g. due to missing packages), consider using our Docker container. See :ref:`running_tests_inside_docker`.
+
+To run Python unit tests, first install `pytest <https://docs.pytest.org/en/latest/contents.html>`_ package:
+
+.. code:: bash
+
+  pip3 install --user pytest
+
+Then compile XGBoost:
```
trivialfis: We don't need to repeat the build instructions here.  There's `build.rst`.

doc/contrib/unit_tests.rst
```diff
@@ -0,0 +1,232 @@
...
+
+.. note:: Having issue? Try Docker container
+
+  If you are having difficulty running the commands below (e.g. due to missing packages), consider using our Docker container. See :ref:`running_tests_inside_docker`.
+
+To run Python unit tests, first install `pytest <https://docs.pytest.org/en/latest/contents.html>`_ package:
+
+.. code:: bash
+
+  pip3 install --user pytest
```
trivialfis: Just `pip3 install pytest`.  I use virtualenv so the `--user` option will install pytest to the wrong place.  If you weren't using virtualenv, OS will report a permission error and as a user you should try to look up related doc.

doc/contrib/unit_tests.rst
```diff
@@ -0,0 +1,232 @@
...
+  cd build
+  cmake ..
+  make
+  cd ..
+
+Now invoke pytest at the project root directory:
+
+.. code:: bash
+
+  export PYTHONPATH=./python-package
```
trivialfis: Have you tested whether this works as expected?  I usually install XGBoost first then run the tests.  We can later revise setup.py a little bit to utilize develop option of setuptools.

doc/contrib/unit_tests.rst
```diff
@@ -0,0 +1,232 @@
...
+  pytest -v -s --fulltrace tests/python
+
+In addition, to build and test CUDA code, run:
+
+.. code:: bash
+
+  cd build
+  cmake -DUSE_CUDA=ON -DUSE_NCCL=ON ..
```
trivialfis: Same as above, we have build.rst.  Please try to cite that file instead of repeating it here.

doc/contrib/unit_tests.rst
```diff
@@ -0,0 +1,232 @@
...
+
+One can also run all unit test using ctest tool which provides higher flexibility. For example:
+
+.. code-block:: bash
+
+  ctest --verbose
+
+.. _running_tests_inside_docker:
+
+Running tests inside a Docker container (Recommended)
```
trivialfis: Actually having people to fail on tests are good thing.  Many occasions some bugs that weren't caught on test farm showed up on my machine, or sometimes the other way around.  Encouraging diversity of environment may help us catching more bugs.  I don't see the necessity of promoting Docker since we have a fairly simple dependency tree.

doc/contrib/unit_tests.rst
```diff
@@ -0,0 +1,232 @@
...
+
+How to use sanitizers with CUDA support
+=======================================
+Runing XGBoost on CUDA with address sanitizer (asan) will raise memory error.
+To use asan with CUDA correctly, you need to configure asan via ASAN_OPTIONS
+environment variable:
+
+.. code-block:: bash
+
+  ASAN_OPTIONS=protect_shadow_gap=0 ${BUILD_DIR}/testxgboost
```
sriramch: is this tested? i never had success in using asan with cuda...

doc/contrib/coding_guide.rst
```diff
@@ -0,0 +1,158 @@
...
+################
+
+**Contents**
+
+.. contents::
+  :backlinks: none
+  :local:
+
+********************
+C++ Coding Guideline
```
sriramch: are the guidelines around abi compatibility documented somewhere? if not, should we? ideally, we wouldn't want to break abi (for external facing api's) until a major release (or until a minor version release bump). if we go down that path, perhaps, we should enforce abi checks as a part of pull request as well...

trivialfis:
@sriramch Asan is only able to detect host side error.  Could you be more specific about the failures you encountered?

sriramch:
> @sriramch Asan is only able to detect host side error. Could you be more specific about the failures you encountered?

@trivialfis - yes, i'm aware. i was trying to run the unit tests with this option enabled while building (`USE_SANITIZER=ON`). i *think* there was a hang of some sorts - it has been a while though since i tried enabling asan. i'll get back with more specific info. the next time i enable it.

doc/contrib/coding_guide.rst
CodingCat: @sriramch yes, we had some discussion in https://github.com/dmlc/xgboost/pull/4646/files#diff-0ced3dfd44ba156272e6db34a5247462, '

we will add binary compatibility tools after releasing 1.0 (i.e. taking 1.0.0 as the ground truth)

trivialfis:
@sriramch Have you tried `ASAN_OPTIONS=protect_shadow_gap=0 ./testxgboost` ?

doc/contrib/index.rst
```diff
@@ -1,14 +1,28 @@
-###################
+#####################
 Contribute to XGBoost
-###################
+#####################
 
+XGBoost has been developed by community members. Everyone is welcomed to contribute. We value all forms of contributions, including, but not limited to:
```
rongou: welcome

doc/contrib/coding_guide.rst
```diff
@@ -0,0 +1,158 @@
...
+********************
+- Follow `Google style for C++ <https://google.github.io/styleguide/cppguide.html>`_.
```
rongou: Are we following the 80-character line length? I see a lot of files with longer lines. Perhaps 100?

hcho3: We actually enforce 100-character line length instead: https://github.com/dmlc/dmlc-core/blob/f1ff6cc117f4e95169a9f62be549c8fe3e15c20f/scripts/lint.py#L72

Let me update the doc. I should also mention that we allow the use of C++ exception.

doc/contrib/coding_guide.rst
hcho3: @sriramch Let me clarify the release policy. We plan to guarantee ABI compatibility for all C API functions in [c_api.h](https://xgboost.readthedocs.io/en/latest/dev/c__api_8h.html), but not for any other C++ functions or constructs. This is because only the former is exposed to other language bindings, such as Python, R, and JVM packages.

doc/contrib/unit_tests.rst
hcho3: Okay, let me go ahead and remove the Docker instructions.

doc/contrib/unit_tests.rst
hcho3: Yes, I use it all the time.

hcho3:
@sriramch @trams I updated the Release Policy to include a note about ABI compatibility. Can you take a look and see if it's sufficiently clear?

sriramch:
>> Also note that a change in the MAJOR version signifies a breaking change in the binary. Thus, an application designed to use the shared library libxgboost.so from XGBoost 1.x is not guaranteed to work with libxgboost.so from XGBoost 2.x.

@hcho3 we can easily break this for the consumers *without* this explicit note. i think it is incumbent about us to version the shared objects when we break abi with the major release. hence, what is libxgboost.so in 1.x will be libxgboost.so.1 in 2.x. a similar versioning scheme can exist for windows as well. this is enabled via the soname directives on linux's. hence, binaries built using 1.x shall even fail to come up with 2.x without rebuilding, as they will look for the unversioned shared object and fail.

hcho3:
@sriramch I see your point about re-naming shared libraries, but I still think it's helpful to state our policy explicitly.

trivialfis:
Let's delay the plan for binary stability.  XGBoot are moving quite fast currently, keeping ABI stable will add some restrictions.   We can make a check list and see what we want to accomplish before making the guarantee.

hcho3:
@trivialfis Do we plan to make any change in the C API?

trivialfis:
Currently I'm not sure.  Let's see how cudf integration goes.  Also do you want to have a set of API dealing with XGBoost trees for treelite?  The point is let's plan for stable ABI only if we are sure that it's actually stable.

hcho3:
Ah I forgot about cudf. Then let's delay the binary stability for now, until after cudf integration work is done.

> Also do you want to have a set of API dealing with XGBoost trees for treelite?

I don't think it's necessary, since Treelite is already able to parse XGBoost model files directly.

trivialfis:
Not if we replace binary with json.

hcho3:
@trivialfis We can discuss after you're done with the JSON prototype.

trams:
@hcho3 thank you for a clear answer. I appreciate. We can live without ABI compatibility. I would just prefer we state this explicitly

sriramch:
@trivialfis 

> i was trying to run the unit tests with this option enabled while building (`USE_SANITIZER=ON`). i _think_ there was a hang of some sorts - it has been a while though since i tried enabling asan. i'll get back with more specific info. the next time i enable it.

just following up on this - did not mean to pollute this pr with this. we can take this offline, if you wish. i'll investigate this later. here are some snippets after enabling asan builds with the appropriate ASAN_OPTIONS set before running the unit tests.

```
[ RUN      ] GPUSet.Verbose
unknown file: Failure
C++ exception with description "[21:12:06] /home/sriramc_nvidia_com/xgboost.master/src/common/common.cc:40: Check failed: n_gpus <= n_devices_visible (1 vs. 0) : 
Stack trace:
  [bt] (0) ./testxgboost() [0x414081]
  [bt] (1) ./testxgboost() [0x41474d]
  [bt] (2) ./testxgboost() [0x667e03]
  [bt] (3) ./testxgboost() [0x55c119]
  [bt] (4) ./testxgboost() [0xa50084]
  [bt] (5) ./testxgboost() [0xa4b515]
  [bt] (6) ./testxgboost() [0xa31390]
  [bt] (7) ./testxgboost() [0xa31c38]
  [bt] (8) ./testxgboost() [0xa32327]

" thrown in the test body.
[  FAILED  ] GPUSet.Verbose (0 ms)
[ RUN      ] GPUSet.MGPU_GPUBasic
unknown file: Failure
C++ exception with description "[21:12:06] /home/sriramc_nvidia_com/xgboost.master/src/common/common.cc:40: Check failed: n_gpus <= n_devices_visible (1 vs. 0) : 
Stack trace:
  [bt] (0) ./testxgboost() [0x414081]
  [bt] (1) ./testxgboost() [0x41474d]
  [bt] (2) ./testxgboost() [0x667e03]
  [bt] (3) ./testxgboost() [0x55c54b]
  [bt] (4) ./testxgboost() [0xa50084]
  [bt] (5) ./testxgboost() [0xa4b515]
  [bt] (6) ./testxgboost() [0xa31390]
  [bt] (7) ./testxgboost() [0xa31c38]
  [bt] (8) ./testxgboost() [0xa32327]

" thrown in the test body.
[  FAILED  ] GPUSet.MGPU_GPUBasic (1 ms)
[----------] 4 tests from GPUSet (3374 ms total)

[----------] 2 tests from CompressedIterator
[ RUN      ] CompressedIterator.Test
[       OK ] CompressedIterator.Test (2574 ms)
[ RUN      ] CompressedIterator.TestGPU
unknown file: Failure
C++ exception with description "[21:12:09] /home/sriramc_nvidia_com/xgboost.master/tests/cpp/common/../../../src/common/common.h:41: /home/sriramc_nvidia_com/xgboost.master/tests/cpp/common/test_gpu_compressed_iterator.cu: 35: out of memory
Stack trace:
  [bt] (0) ./testxgboost() [0x414081]
  [bt] (1) ./testxgboost() [0x41474d]
  [bt] (2) ./testxgboost() [0x5630b8]
  [bt] (3) ./testxgboost() [0x5805f5]
  [bt] (4) ./testxgboost() [0xa50084]
  [bt] (5) ./testxgboost() [0xa4b515]
  [bt] (6) ./testxgboost() [0xa31390]
  [bt] (7) ./testxgboost() [0xa31c38]
  [bt] (8) ./testxgboost() [0xa32327]

" thrown in the test body.
[  FAILED  ] CompressedIterator.TestGPU (0 ms)

...

[----------] 1 test from Monitor
[ RUN      ] Monitor.Logging
Aborted (core dumped)
```

run on ubuntu 16.04 with cuda 10

PR #4662, opened 2019-07-15
add tests symbol link for pycharm users

jq:
this links to tests folder and allow pycharm users open python-package as the root folder

### DIFF
--- a/python-package/xgboost/tests
+++ b/python-package/xgboost/tests
@@ -1,0 +1,1 @@
+../../tests/python/


hcho3:
> allow pycharm users open python-package as the root folder

Can you explain why you’d want to open python-package as the root folder? XGBoost is not a pure Python package, and the main logic is contained in the native C++ codebase, and opening python-package as the root will hide all C++ code from the view.

jq:
for python coders focus on python development, it is easier to just open python-package as the root folder.

PR #4664, opened 2019-07-16
Fix maven warnings

trams:
* exec plugin was missing a version
* reportPlugins has been deprecated:
  see https://maven.apache.org/plugins/maven-site-plugin/maven-3.html#Classic_configuration_Maven_2__3


CodingCat:
merged, thanks

PR #4667, opened 2019-07-16
[jvm-packages] allowing chaining prediction

CodingCat:
this PR is to include  https://github.com/dmlc/rabit/pull/99

the use case is like following 

```
val df1 = xgbmodel.transform(training)
val df2 = xgbmodel2.transform(df1)
```

because we shutdown rabit in both of df1 and df2 (the task will processing df1's partition first and then df2's ), the original code explicitly prohibit such usage pattern, 

I added a test to guard this usage pattern 

@sperlingxx would you mind helping to review ?


sperlingxx:
LGTM.

CodingCat:
@sperlingxx thanks

PR #4670, opened 2019-07-17
upgrade version num

CodingCat:


hcho3:
CMake doesn't allow non-integers in version number

CodingCat:
> CMake doesn't allow non-integers in version number

good to know, thanks

PR #4676, opened 2019-07-18
[jvm-packages] Launch dev jvm image under my user

trams:
I am actually very interested to use dev jvm Docker image on my linux machine cause otherwise the binaries I built are linked to a very new GLIBC and they cannot be used on my Spark cluster

But the thing is on linux Docker behaves differently from other platforms.
One thing which is different is that on linux before this change the Docker container would produce a root owned files. This makes it difficult to switch from local dev to this container and back.

In this change I employ `entrypoint.sh` script which other Docker images use which create a user inside the container and it uses gosu to launch a build|test job under this user. This leads for the script to produce files which are owned by my host machine user.

I also changed a bit the way I share a maven M2 cache cause after this change you can't easily use /root folder

exec ensures that the last command is not launched in a subshell


trams:
I forgot to reference the relevant pull request: #4640 
cc @thesuperzapper

thesuperzapper:
@trams have you tested that the new Dockerfile works with `.\dev\build-linux.cmd` on Windows? 

And that --skip-tests still works?

trams:
I just tested the build-linux.cmd on Windows and it worked. But note that I did not actually changed the Windows part at all

trams:
On Windows I testsed

CodingCat:
@thesuperzapper can you help to review this?

trams:
I forgot to tell that on Linux I made sure that --skip-tests still work

trams:
@thesuperzapper, please help review this change

PR #4683, opened 2019-07-19
[WIP] Initial commit for JSON.

trivialfis:
Currently depends on #4577 .  This is a monolithic PR for early review.  I will try to split it up into 2 or 3 dependent PRs later once it's ready.  I can create stacked PRs if pushing to origin is allowed. :-)

TODOs:
- [ ] Revise schema.
- [x] Locales
- [x] Better error reporting.
- [x] Check for implicit copying.
- [x] Post profiling results.
- [ ] Enable GPU device configuration for varying environments.
- [ ] Tree node stat instance count.
- [ ] Enable pickling support.
- [ ] Enable rabit checkpoint.
- [x] `default_left` is not saved.

@hcho3 @RAMitchell  Sorry for the long wait.

## Notes to myself

`leaf_child_cnt` is not yet omitted.
`num_boosted_rounds` is not saved.

## Merged parts

Initial implementation #4708
Slightly improved version #4739

### DIFF
--- a/doc/json-schema.rst
+++ b/doc/json-schema.rst
@@ -1,0 +1,710 @@
+################################
+XGBoost JSON Schema, Version 1.0
+################################
+
+Preface
+=======
+This document contains an exhaustive description of the XGBoost JSON schema, a
+mapping between XGBoost object classes and JSON objects and arrays. We aim to
+store a **complete representation** of all XGBoost objects.
+
+Representation of ``dmlc::Parameter`` objects
+---------------------------------------------
+Every object of subclasses of ``dmlc::Parameter`` are to be represented as JSON
+objects. We create a bridge method to seamlessly convert ``dmlc::Parameter`` objects into
+JSON objects, such that proper value types are used.
+
+For example, consider the parameter class
+
+.. code-block:: cpp
+
+  struct MyParam : public dmlc::Parameter<MyParam> {
+    float learning_rate;
+    int num_hidden;
+    int activation;
+    std::string name;
+    DMLC_DECLARE_PARAMETER(MyParam) {
+      DMLC_DECLARE_FIELD(num_hidden);
+      DMLC_DECLARE_FIELD(learning_rate);
+      DMLC_DECLARE_FIELD(activation).add_enum("relu", 1).add_enum("sigmoid", 2);
+      DMLC_DECLARE_FIELD(name);
+    }
+  };
+
+and an object created by the following initialization code:
+
+.. code-block:: cpp
+
+  MyParam param;
+  param.learning_rate = 0.1f;
+  param.num_hidden = 10;
+  param.activation = 2;  // sigmoid
+  param.name = "MyNet";
+
+This collection is naturally expressed as the following JSON object:
+
+.. parsed-literal::
+
+  {
+    "learning_rate": "0.1",
+    "num_hidden": "10",
+    "activation": "sigmoid",
+    "name": "MyNet"
+  }
+
+Notice the value fields are strings even for numeric values.
+
+Notations
+---------
+
+In the following sections, the schema for each XGBoost class is shown as a JSON
+object. Fields whose keys are marked with *italic* are optional and may be
+absent in some models. The hyper-linked value indicate that the value shall be
+the JSON representation of another XGBoost class. The *italic* value indicate
+that the value shall be a primitive type (string, integer, floating-point etc).
+Every mention of ``floating-point`` refers to single-precision floating point
+(32-bit), unless explicitly stated otherwise.  Every mention of ``integer``
+refers to 32-bit integer unless stated otherwise.
+
+Full content of the schema
+==========================
+
+Note: Click :ref:`here <example>` for a minimal example of the current schema.
+
+.. contents:: :local:
+
+XGBoostModel
+------------
+This is the root object for XGBoost model.
+
+.. parsed-literal::
+
+  {
+    "version" : ["1", "0"],
+    "learner" : Learner_
+  }
+
+Learner
+-------
+.. parsed-literal::
+
+  {
+    "gradient_booster" : GradientBooster_,
+    "learner_train_param" : LearnerTrainParam_,
+    "learner_model_param": LearnerModelParam_,
+    "metrics" : [ *array of* Metric_ ],
+    "objective" : Objective_
+  }
+
+The ``learner_train_param`` field stores (hyper)parameters used for training, while
+``learner_model_param`` field stores model related properties.
+
+The ``gradient_booster`` field stores an gradient boosted ensemble consisting of
+models of certain type (e.g. tree, linear).
+
+The ``metrics`` field is used to store evaluation metrics.
+
+The ``objective`` field stores the objective (loss) function used to train the
+ensemble model.
+
+LearnerTrainParam
+-----------------
+This class is a subclass of ``dmlc::Parameter``.
+
+.. parsed-literal::
+
+  {
+    "dsplit": *string*,
+    "disable_default_eval_metric": *boolean*,
+    "booster": "string",
+    "objective": "string",
+    "num_boosting_round" : *integer*
+  }
+
+The ``dsplit`` field indicates the data partitioning mode for distributed
+learning. Its value should be one of ``auto``, ``col``, and ``row``. The value
+should be set to ``auto`` when only a single node is used for training.
+
+The ``num_boosting_round`` field stores the number of boosting rounds performed.
+This number is different from the number of trees if ``num_parallel_tree`` of
+GBTreeTrainParam_ is greater than 1.
+
+LearnerModelParam
+-----------------
+This class is a subclass of ``dmlc::Parameter``.
+
+.. parsed-literal::
+
+   {
+     "base_score": "*integer*",
+     "num_feature" : "*integer*",
+     *"num_class"* : "*integer*"
+   }
+
+
+The ``num_class`` is used only for multi-class classification task, in which it
+indicates the number of output classes.
+
+GenericParameter
+----------------
+This class is a subclass of ``dmlc::Parameter``.
+
+.. parsed-literal::
+
+  {
+    "seed": "*integer*",
+    "seed_per_iteration": "*boolean*",
+    "nthread": "*integer*"
+    *"gpu_id"*: "*integer*",
+    *"n_gpus"*: "*integer*"
+  }
+
+The ``gpu_id`` and ``n_gpus`` fields are used to set the GPU(s) to use for
+training and prediction. If no GPU is used, the fields should be omitted.
+
+GradientBooster
+---------------
+Currently, we may choose one of the three subclasses for the gradient boosted
+ensemble:
+
+* GBTree_: decision tree models
+* Dart_: DART (Dropouts meet Multiple Additive Regression Trees) models
+* GBLinear_: linear models
+
+We can determine which subclass was used by looking at the ``name`` field
+of each subclass.
+
+Metric
+------
+.. parsed-literal::
+
+  *string*
+
+For the time being, every metric is fully specified by a single string. In the
+future, we may want to add extra parameters to some metrics. When that happens,
+we should add subclasses of ``Metric``.
+
+The string must be a valid metric name as specified by the `parameter
+doc <https://xgboost.readthedocs.io/en/latest/parameter.html#learning-task-parameters>`_.
+
+Objective
+---------
+Currently, we may choose one of the 10 subclasses for the objective function:
+
+* RegLossObj_
+* SoftmaxMultiClassObj_
+* HingeObj_
+* LambdaRankObj_
+* PairwiseRankObj_
+* LambdaRankObjNDCG_
+* LambdaRankObjMAP_
+* PoissonRegression_
+* CoxRegression_
+* GammaRegression_
+* TweedieRegression_
+
+GBTree
+------
+The ``GBTree`` class stores an ensemble of decision trees that are produced
+via gradient boosting. It is a subclass of GradientBooster_.
+
+.. parsed-literal::
+
+  {
+    "name" : "gbtree",
+    "gbtree_train_param" : GBTreeTrainParam_,
+    "model" : GBTreeModel_
+  }
+
+The ``gbtree_train_param`` field is the list of training parameters specific to
+``GBTree``.
+
+GBTreeTrainParam
+----------------
+This class is a subclass of ``dmlc::Parameter``.
+
+.. parsed-literal::
+
+  {
+    "num_parallel_tree": *integer*,
+    "predictor": *string*,
+    "process_type": *string*,
+    "tree_method": string,
+    "updater": string,
+  }
+
+The ``num_parallel_tree`` field denotes the number of parallel trees constructed
+during each iteration. It is used to support boosted random forest.
+
+The ``updater`` field stores a comma separated list of updater names that was provided at
+the beginning of training. This field may not necessarily match the sequence given in the
+``updaters`` field of GBTree_ or Dart_ due to auto configuration.
+
+The ``process_type`` field denotes whether to create new trees (``default``) or
+to update existing trees (``update``) during the boosting process. The field's
+value must be either ``default`` or ``update``. Keep in mind that ``update`` is
+highly experimental; most use cases will use ``default``.
+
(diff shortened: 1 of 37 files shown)


doc/json-schema.rst
```diff
@@ -0,0 +1,1059 @@
+################################
+XGBoost JSON Schema, Version 1.0
```
RAMitchell: What is the plan to maintain this? I can see this going out of date quickly without some kind of CI enforcement. This looks like we are exposing the complete serialisation format such that someone could construct a working xgboost model completely external to xgboost. This could be a very useful feature, but also with some cost to support. Note that up until now we have not provided a schema for our binary format, only tried to enforce backwards compatibility.

hcho3: @RAMitchell We can automate the schema enforcement if we express our model schema using JSON Schema. There are tools available to validate JSON files against a given JSON schema.

hcho3:
@RAMitchell I agree that NIH (JSON parsing code) should eventually be merged into dmlc-core. For now, let's keep it in XGBoost and make sure it works.

hcho3:
>  I can create stacked PRs if pushing to origin is allowed. :-)

Why don’t you create a new branch dev_json and push your code there?

trivialfis:
@hcho3 Jenkins seems to return a 404 at first load.  It works fine after a refresh.  I have been experiencing the error in few days so it might not an incident.

Screenshot from 2019-07-22 04-14-06

include/xgboost/tree_model.h
```diff
@@ -320,6 +331,8 @@ class RegTree {
     fo->Write(dmlc::BeginPtr(stats_), sizeof(RTreeNodeStat) * nodes_.size());
   }
 
+  void Save(Json* out) const;
```
trams: What do you think about using a smart pointer (I think unique_ptr could work here but I am fine with shared_ptr too)?

I think usage of smart pointers could make our code a bit safer. What do you think?

src/c_api/c_api.cc
```diff
@@ -891,23 +895,51 @@ XGB_DLL int XGBoosterPredict(BoosterHandle handle,
 XGB_DLL int XGBoosterLoadModel(BoosterHandle handle, const char* fname) {
   API_BEGIN();
   CHECK_HANDLE();
-  std::unique_ptr<dmlc::Stream> fi(dmlc::Stream::Create(fname, "r"));
```
trams: I am wandering whether this stream supports peek() method. If so we may peak at the header and see whether it is binary format or not?

I am a bit fuzzy on details how binary serialization works now. Do we have some kind of known header?

src/gbm/gblinear_model.cc
```diff
@@ -0,0 +1,99 @@
...
+  auto& out = *p_out;
+  out["model_param"] = toJson(param);
+  std::string raw;
+  raw.reserve(weight.size() * 4 /*floats*/ + weight.size() /*comma*/ );
+  raw += '[';
+  size_t const n_weights = weight.size();
+  for (size_t i = 0; i < n_weights; ++i) {
+    auto w = weight[i];
+    if (i != n_weights - 1) {
+      raw += std::to_string(w) + ',';
```
trams: I am concerned that std::to_string is locale dependent (https://en.cppreference.com/w/cpp/string/basic_string/to_string)
and in some locales (russian or ukrainian) they use ',' as a floating point which may break the output.

What do you think about using a locale independent function (if we can use C++17 we should consider std::to_chars)

trams: Also as far as I understood documentation std::to_string just calls fprintf (https://en.cppreference.com/w/cpp/io/c/fprintf) which by default outputs floating point numbers with precision 6 which is not always enough to reconstruct the same number

src/tree/tree_model.cc
```diff
@@ -618,6 +622,138 @@ std::string RegTree::DumpModel(const FeatureMap& fmap,
...
+
+  void ParseStat(int32_t n_nodes, std::vector<RTreeNodeStat>* p_stats) {
+    p_stats->resize(n_nodes);
+    auto& stats = *p_stats;
+    GetChar('[');
+    for (int32_t i = 0; i < n_nodes; ++i) {
+      GetChar('[');
+      auto& s = stats[i];
+      size_t pos {0};
+      s.loss_chg = std::stod(raw_str_.substr(cursor_.Pos(), 17), &pos);
```
trams: I cannot figure out why do we take 17 letters here. How do we know the size of the field?

Is it the size of serialized double?
What do you think about using std::numeric_limits<double>::max_digits10 ?

include/xgboost/tree_model.h
trivialfis: It's pretty safe for current purpose.  The Json object is a stack allocated object in C_API, going through the model and then expired due to being out of scope.

src/c_api/c_api.cc
trivialfis: > I am a bit fuzzy on details how binary serialization works now. Do we have some kind of known header?

Not really.  I want to deprecate the binary format, it might still take a long time to actually do it.  Anyway I myself consider it as a burden and don't want to deal with it anymore. ;-)

src/gbm/gblinear_model.cc
trivialfis: Good point.  I will look into `std::to_chars`, thanks for the suggestion.

src/tree/tree_model.cc
trivialfis: `std::numeric_limits::max_digits10` is a good choice.  Again, thanks for the suggestion.

trivialfis:
> How close is your format to xgboost json dump?

Not close.  The dump prints only trees, but prints with much more readable structure.  While this one dumps **everything**, but in a much more compact way for serialization.

src/gbm/gblinear_model.cc
trivialfis: Implementing `std::{from, to}_chars` for integer is easy, but floating point number is messy.  I think I'm just gonna use `std::locale` to work around it instead.

hcho3:
@trivialfis It looks like the result page is inaccessible from anonymous (not logged-in) users. Let me take a look.

src/gbm/gblinear_model.cc
trams: I think worst case we can set locale at the top of conversion.

I am more concern that we should output at least numeric_limit<float>::max_digits10 digits so we can reconstruct the same float and I do not believe that the default does this (I think it outputs only 6 instead of 9 digits)

src/c_api/c_api.cc
trams: I see. Good to know that native format will become a thing from the past eventually

include/xgboost/tree_model.h
trams: I forgot about object on the stack. Yes. You are right, raw pointers are the best solution in this case

src/gbm/gblinear_model.cc
hcho3: @trivialfis How about using [`std::sprintf`](https://en.cppreference.com/w/cpp/io/c/fprintf)? This function is governed by the locale set in [`std::setlocale()`](https://en.cppreference.com/w/cpp/locale/setlocale), and the standard guarantees that the default C locale is `"C"`.

include/xgboost/tree_model.h
AnthonyTruchet: Well references are an even better option in this case...

src/gbm/gblinear_model.cc
AnthonyTruchet: Are a sanity rules, locales are meant for interaction with a human user not for serialization format. serialization format SHOULD use the invariant C locale and the numerical precision SHOULD enable roundtrip. I've already lost hour debugging ML code that was not doing that, let avoid this nightmare ;)

AnthonyTruchet: This might be useful but I've not done C++ for a few year and might be outdated: https://www.boost.org/doc/libs/1_59_0/libs/locale/doc/html/std_locales.html

include/xgboost/tree_model.h
trivialfis: @AnthonyTruchet It's by google coding standard, just pass a pointer if it's meant to be changed.  So ...

src/gbm/gblinear_model.cc
hcho3: @trivialfis FYI, fmtlib is becoming part of C++20: http://www.zverovich.net/2019/07/23/std-format-cpp20.html

trivialfis: @hcho3 Good to know.

trams: BTW, as a general rule what version of a standard are we using for xgboost?
C++14?

hcho3: @trams We adhere to C++11 standard.

src/gbm/gblinear_model.cc
```diff
@@ -0,0 +1,114 @@
...
+
+void GBLinearModel::Save(Json* p_out) const {
+  using WT = std::remove_reference<decltype(std::declval<decltype(weight)>().back())>::type;
+  using JT = Number::Float;
+  static_assert(std::is_same<WT, JT>::value, "");
+  auto& out = *p_out;
+  out["model_param"] = toJson(param);
+  std::string raw;
+  raw.reserve(
+      weight.size() * std::numeric_limits<Number::Float>::max_digits10 /*floats*/ +
```
trams: Note that std::numeric_limits<Number::Float>::max_digits10 is just amount of digits one need to perfect serialization-deserialization without lost of data. The resulting string would be longer because we also need to output dot ('.') and possible minus ('-') for negative numbers if any

So I suggest to add +1 here and even +2 if negative numbers are possible

src/gbm/gblinear_model.h
```diff
@@ -10,6 +10,8 @@
 #include <string>
 #include <cstring>
 
+#include "xgboost/json.h"
```
trams: I would suggest to put a forward declaration here. `class Json;` should be enough
Otherwise `json.h` header could "escape" though this header.

src/gbm/gbtree.cc
```diff
@@ -565,5 +604,51 @@ XGBOOST_REGISTER_GBM(Dart, "dart")
...
+  TreeSelectRaw(StringView str, size_t* pos) :
+      JsonReader{str, *pos}, pre_pos_{pos} {}
+
+  Json ParseRaw() {
+    SkipSpaces();
+    auto beg = cursor_.Pos();
+    while (true) {
+      char ch = GetNextChar();
+      while (ch != ']' && ch != -1) { ch = GetNextChar(); }
+      ch = GetNextNonSpaceChar();
```
trams: As far as I understand after line 622 we would have
ch == ']' or ch == -1
the first one is non white space. The second case indicates that the stream has been at the end.
Either way we do not need to skip non space chars. Or do I miss something here?

src/tree/tree_model.cc
```diff
@@ -618,6 +622,165 @@ std::string RegTree::DumpModel(const FeatureMap& fmap,
...
+                   };
+
+  tree_raw += '[';
+  for (size_t i = 0; i < stats_.size(); ++i) {
+    auto const& s = stats_[i];
+    tree_raw += '[';
+    AppendFloat(s.loss_chg, false);
+    AppendFloat(s.sum_hess, false);
+    AppendFloat(s.base_weight, false);
+    AppendInt(s.leaf_child_cnt, true);;
```
trams: Typo: two semicolons

src/tree/tree_model.cc
```diff
@@ -618,6 +622,165 @@ std::string RegTree::DumpModel(const FeatureMap& fmap,
...
+void RegTree::Save(Json* p_out) const {
+  auto& out = *p_out;
+  CHECK_EQ(param.num_nodes, static_cast<int>(nodes_.size()));
+  CHECK_EQ(param.num_nodes, static_cast<int>(stats_.size()));
+  out["tree_param"] = toJson(param);
+  CHECK_EQ(get<String>(out["tree_param"]["num_nodes"]), std::to_string(param.num_nodes));
+
+  std::string tree_raw;
+  tree_raw.reserve(
+      stats_.size() * std::numeric_limits<Number::Float>::max_digits10 * 2);
```
trams: I am trying to figure out why you reserve this amount of memory.
It seems each stat has 3 floats and one integer so I would expect that the total length needed is

stats_.size() * ( 2 + float_length*3 + int_length + 3) + 2
2 is []
3 are commas
float_length = std::numeric_limits<Number::Float>::max_digits10 + 1
int_length ...

src/tree/tree_model.cc
```diff
@@ -618,6 +622,165 @@ std::string RegTree::DumpModel(const FeatureMap& fmap,
...
+    tree_raw += ']';
+    if (i != stats_.size() - 1) {
+      tree_raw += ',';
+    }
+  }
+  tree_raw += ']';
+  out["stats"] = JsonRaw(std::move(tree_raw));
+
+  tree_raw.clear();
+  tree_raw.reserve(nodes_.size() * 9 * 2);
```
trams: I guess 9 comes from amount of digits one need to save while serializing float. Why do we have 2 here? We have 5 ints and 1 float below

I see that the logic for preallocating buffers for arrays is quite complex and I think this is not the only place we have it. What do you think about extracting this into separate functions?

It is not very critical. It is rather the optimization.

tests/cpp/common/test_json.cc
```diff
@@ -0,0 +1,360 @@
...
+
+TEST(Json, TestParseObject) {
+  std::string str = R"obj({"TreeParam" : {"num_feature": "10"}})obj";
+  auto json = Json::Load(StringView{str.c_str(), str.size()});
+}
+
+TEST(Json, ParseNumber) {
+  std::string str = "31.8892";
+  auto json = Json::Load(StringView{str.c_str(), str.size()});
+  ASSERT_NEAR(get<JsonNumber>(json), 31.8892, kRtEps);
```
trams: I think we use float everywhere (not double) so I would suggest to add `f` suffix to the literal: 31.8892f
This is rather minor and nitpicking

trivialfis:
@trams Thanks for the review.  I'm no expert with string manipulations so your suggestions are really helpful.

src/common/json.cc
```diff
@@ -0,0 +1,609 @@
...
+
+  Number::Float number{0};
+  number = std::stof(substr, &pos);
+  for (size_t i = 0; i < pos; ++i) {
+    GetNextChar();
+  }
+  return Json(number);
+}
+
+Json JsonReader::ParseBoolean() {
```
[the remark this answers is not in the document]
trivialfis: @trams I don't think it can reach `ParseBoolean` at the first place if current token leads with a `x`.

src/common/json.cc
```diff
@@ -0,0 +1,609 @@
...
+    } else if (ch == '\t') {
+      buffer += "\\t";
+    } else if (static_cast<uint8_t>(ch) <= 0x1f) {
+      // Unit separator
+      char buf[8];
+      snprintf(buf, sizeof buf, "\\u%04x", ch);
+      buffer += buf;
+    } else {
+      buffer += ch;
+    }
```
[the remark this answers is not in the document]
trivialfis: @trams I do want to support UTF-8.  We want to upstream this into `dmlc-core` once mature, so I'm trying not to make assumption.  It will just have to take some more time, which is a good thing since model serialization is important and deserves the time.

src/common/json.cc
```diff
@@ -0,0 +1,609 @@
...
+  this->Write(StringView{str.c_str(), str.size()});
+  convertor_.str("");
+}
+
+void JsonWriter::Visit(JsonRaw const* raw) {
+  auto const& str = raw->getRaw();
+  this->Write(str);
+}
+
+void JsonWriter::Visit(JsonNull const* null) {}
```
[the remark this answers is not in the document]
trivialfis: @trams Thanks for pointing that out.  Fixed.

trivialfis:
Extracting implementation to #4732 for easier review.

trivialfis:
Sorry folks, this is no longer blocking 1.0 as I want think it's necessary to improve `dmlc::Parameter` first before finishing JSON integration.  See #4755 .  I want to do it right so it might take some more time.  Please understand.

PR #4693, opened 2019-07-22
dynamically allocate gpu device

wbo4958:
Prerequisite: GPU must be configured as EXCLUSIVE_PROCESS by
"nvidia-smi -c EXCLUSIVE_PROCESS"

when detecting that gpu will be used and gpu_id is not be specified as well,
then auto allocate a GPU for current worker.

### DIFF
--- a/demo/dask/dask_gpu_auto_demo.py
+++ b/demo/dask/dask_gpu_auto_demo.py
@@ -1,0 +1,43 @@
+from dask.distributed import Client, LocalCluster
+import dask.dataframe as dd
+import dask.array as da
+import numpy as np
+import xgboost as xgb
+import GPUtil
+import time
+
+
+# Define the function to be executed on each worker
+def train(X, y):
+    dtrain = xgb.dask.create_worker_dmatrix(X, y)
+    # Specify the GPU algorithm
+    # If set GPU to be EXCLUSIVE mode by nvidia-smi -c EXCLUSIVE_PROCESS, native will
+    # auto-allocate gpu for each worker
+    params = {"tree_method": "gpu_hist"}
+    print("Worker {} starting training on {} rows".format(xgb.rabit.get_rank(), dtrain.num_row()))
+    start = time.time()
+    xgb.train(params, dtrain, num_boost_round=500)
+    end = time.time()
+    print("Worker {} finished training in {:0.2f}s".format(xgb.rabit.get_rank(), end - start))
+
+
+def main():
+    max_devices = 16
+    # Check which devices we have locally
+    available_devices = GPUtil.getAvailable(limit=max_devices)
+    # Use one worker per device
+    cluster = LocalCluster(n_workers=len(available_devices), threads_per_worker=4)
+    client = Client(cluster)
+
+    # Set up a relatively large regression problem
+    n = 100
+    m = 10000000
+    partition_size = 100000
+    X = da.random.random((m, n), partition_size)
+    y = da.random.random(m, partition_size)
+
+    xgb.dask.run(client, train, X, y)
+
+
+if __name__ == '__main__':
+    main()
--- a/src/common/common.cc
+++ b/src/common/common.cc
@@ -28,6 +28,10 @@ GlobalRandomEngine& GlobalRandom() {
 int AllVisibleImpl::AllVisible() {
   return 0;
 }
+
+int AllVisibleImpl::AllocateGPUDeviceId() {
+  return -1;
+}
 #endif  // !defined(XGBOOST_USE_CUDA)
 
 constexpr GPUSet::GpuIdType GPUSet::kAll;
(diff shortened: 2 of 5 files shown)


wbo4958:
@rongou @RAMitchell @trivialfis Could you guys help to review this?

rongou:
What happens if EXCLUSIVE_PROCESS is not set?

RAMitchell:
I want to understand a bit more about the requirement for this from the spark side. How does this help configure the cluster? Is this information not available in language bindings?

This PR raises the question of whose job is it to schedule gpu resources - user code or internal xgboost code?

I see that it could bring some convenience to users to do it automatically this way.

Some of your comments about race conditions worry me though. It must be 100% robust for production use. Even failure 1/million times is not ok.

wbo4958:
> What happens if EXCLUSIVE_PROCESS is not set?

If EXCLUSIVE_PROCESS is not set, the primary GPU will be used.

tgravescs:
the race condition comment is purely that you lost in acquiring the GPU.  If 2 processes happen to be trying to acquire the same gpu at the same time, one will get it, one won't.  The one that doesn't lost the race and then just tries the next GPU and loops twice and eventually will just give up and exit.

The TODO comment there was purely if you wanted to add more checks like I want X type of Gpu, or a gpu with Y memory, etc..  That comment should probably be removed.

We could add in a flag to disable it doing the allocation, that way if someone has an environment they don't want it, it can cleanly be disabled.  Is the gpu_id parameter now properly working and passed down everywhere?

RAMitchell:
I would like to change the "gpu_id" parameter to a string with default "auto", otherwise a number  provided by a user. Currently this would produce errors unless exclusive process is set. One way to solve this is to use nvml in xgboost as a dependency. I will try this when I get back from travelling.

If this is too heavyweight for us to include as a dependency it is likely this PR will be rejected and we will require Spark side to handle gpu scheduling.

tgravescs:
We are adding GPU scheduling to Spark but that won't be until Spark 3.0 which won't be released for a few more months. Of course it also relies on your cluster manager supporting it as well.  This works now and with any cluster manager as long as you set the GPU's in PROCESS EXCLUSIVE mode.

trivialfis:
@RAMitchell I went through the source code of GPUtil and I think it's of similar approach, which to me is a bad thing ...  

To summarize, GPUtil uses `nvidia-smi` (which is based on `nvml`) to get a point estimate (with only one output sample) for GPU runtime information.  Then it allocate GPU based on some tunable heuristics like whether the memory is used more than 50% ...  If  there's no GPU that meets the requirement of those heuristics then it will sleep for a period and try again unless number of attempts is set to one.

Is it possible to bind ID of GPU to `Rabit::GetRank()`?

tgravescs:
what exactly do you mean bind the gpu to the rank?  The rank is going to be application specific - meaning 1 XGboost application would have its own rank and a second XGBoost application is going to have its own rank and those might conflict when looking at them together. So if you are running on a cluster manager like YARN where multiple applications might be on the same node, you can't just rely on both getting say gpu id 1, or really any particular GPU id for that matter.  If you just rely on the # of GPUs it needs then you are essentially back to something like this PR.  But maybe I'm missing something.

rongou:
Ideally the cluster manager should be aware of GPUs. That way, if I want to launch, say, an 8-gpu training job in a cluster with 4-gpu nodes, the cluster manager would allocate 2 full nodes for me. Then I can figure out how to allocate a gpu for each worker process based on the rabit rank:
```python
GPUS_PER_NODE = 4
gpu_id = xgb.rabit.get_rank() % GPUS_PER_NODE
```
Or the cluster manager could just launch 8 containers with a gpu attached to each, and the xgboost process would always use gpu 0.

This can be done fairly easily with Kubernetes. I guess this is where we are going with Spark 3.0 as well?

In the mean time, since Spark 2.x is not aware of GPUs, this PR is sort of a workaround. I think it's fine to put it in for now, with the understanding we can take it out once Spark has better support for GPU scheduling.

One problem I see with this approach is that in a busy Spark cluster with many GPU training jobs, it's likely they would step on each and fail to grab the needed number of GPUs. Maybe that's acceptable until the cluster manager provides some gang scheduling capabilities.

RAMitchell:
Im not going to push temporary workarounds to dmlc master branch. If this can be made to work without relying on nvidia-smi exclusive process I think its a compelling feature though.

@trivialfis my plan is to use nvml to detect if any process is using the device ([nvmlDeviceGetComputeRunningProcesses](https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceQueries.html#group__nvmlDeviceQueries_1g46ceaea624d5c96e098e03c453419d68))

trivialfis:
Closing for now.  We will come up with something later if Spark 3.0 doesn't beat up to it.

Issue #4498, opened 2019-05-24
Consider migrating tests from AppVeyor/Travis CI to Azure Pipelines


hcho3:
Consider migrating all tests from AppVeyor/Travis CI to [Azure Pipelines](https://azure.microsoft.com/en-us/services/devops/pipelines/).

Why migrate tests?
* OSX tests on Travis CI are quite slow and have proven flaky from time to time (e.g. #4487,  #4448).
* AppVeyor allows only a single job to run at a time; Azure Pipelines allows for 10 parallel jobs
* Azure Pipelines can run both Windows and Mac OSX tests (note: no GPUs.)
* Azure Pipelines is provided free of charge to open-source projects

TODO. Test Azure Pipelines for performance and stability

cc @sriramch

CodingCat:
+10000 on getting rid of AppVeyor

hcho3:
I played with Azure Pipelines a bit, and so far it's been pleasure to work with.
* Builds start almost instantaneously and run fast.
* In a single YAML file, I can use Linux, Mac OSX, and Windows agents.
* Jobs can run in parallel, but some jobs can be made dependent on other jobs.

See example: [azure-pipelines.yml](https://github.com/dmlc/treelite/blob/600afd55d1fa9bb94fc88fd3a3043cb2d5b20651/azure-pipelines.yml), [build result](https://dev.azure.com/hcho3/treelite/_build/results?buildId=49)

trivialfis:
Can't wait.

trams:
I am also interested in seeing Azure Pipelines instead of AppVeyor

hcho3:
We've set up GitHub Actions (#5917) and moved some tests from AppVeyor to GitHub Actions (#5922). There is currently a nondeterministic test that's blocking a further migration away from AppVeyor. 

We are keeping Travis CI because we need access to MacOS High Sierra to build XGBoost Python wheel. (Expectation is that it will be compatible with last three major releases, High Sierra, Mojave, and Catalina.) GitHub Actions only provides MacOS Catalina.

Closing this in favor of #5891.

trivialfis:
@hcho3 I don't think we are running 3 versions of osx on travis right?

trivialfis:
And another thing is, our test on Travis now take 1.5+ hour to run for each PR.  And multiple PRs cannot be run in parallel.

hcho3:
> And multiple PRs cannot be run in parallel.

[...]

PR #4697, opened 2019-07-23
remove RowSet which is no longer being used

rongou:
@hcho3 @RAMitchell @trivialfis

### DIFF
--- a/include/xgboost/data.h
+++ b/include/xgboost/data.h
@@ -349,41 +349,6 @@ class DataSource : public dmlc::DataIter<SparsePage> {
 };
 
 /*!
- * \brief A vector-like structure to represent set of rows.
- * But saves the memory when all rows are in the set (common case in xgb)
- */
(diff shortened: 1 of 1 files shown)


hcho3:
@rongou It looks like the CI worker doesn't have proper access to the S3 object `s3://xgboost-ci-jenkins-artifacts/spark-$SPARK_VERSION-bin-scala2.12-hadoop2.7.tgz`. I'm puzzled why this is happening now, since #4574 passed tests successfully. Let me look.

trivialfis:
@rongou Would you like to rebase it onto master so the tests can pass?

rongou:
@hcho3 looks like the python tests are still failing, but for a different reason now?

hcho3:
@rongou See #4498. We'd like to remove away from Travis CI asap.

rongou:
@hcho3 ok finally passed.

PR #4698, opened 2019-07-23
- optimize gpu memory usage

sriramch:
- this pr provides users the option to train larger datasets by enabling external
     memory *and* peforming prediction related ops on cpu (by using cpu predictor)
      - yes, it will be slow, and that is the trade off here
   - it moves operations such as base prediction computation (and by association the input matrix
     labels and weights), metrics computation etc. out of gpu memory into system memory and cpu
   - it also provides a csr type data layout for the histogram indices, which saves a good chunk
     of gpu memory, with savings nearly proportional to the sparsity of the matrix

- testing done
   - no regression when external memory isn't enabled. it is slightly faster due to prevention of
     some copies (gradient pair, prediction cache vector etc.)
   - it will be slower when external memory is enabled; but can now train larger data sets
      - we can now train 1 billion rows of the mortgage dataset on a system that has 4 t4's. it
        used to be ~670m earlier


sriramch:
@RAMitchell @trivialfis @rongou please review. i'll ask why the pr checks aren't starting

sriramch:
@hcho3 any reason why the pr checks aren't kicking off?

hcho3:
@sriramch Looks like there is a permission issue with status update: https://xgboost-ci.net/job/xgboost/job/PR-4698/1/console

hcho3:
@sriramch I had to re-enter my GitHub credentials so that Jenkins would have permission to push status updates. It should work now.

hcho3:
For your reference: https://xgboost.readthedocs.io/en/latest/contrib/coding_guide.html#running-formatting-checks-locally

sriramch:
> @sriramch I had to re-enter my GitHub credentials so that Jenkins would have permission to push status updates. It should work now.

[sc] thanks @hcho3 

> It would be nice if I can get some tips from you guys at NVIDIA about dev ops with Jenkins. https://gpuci.gpuopenanalytics.com looks pretty nice to me.

[sc] i think some folks in china may be looking/working on it. i'll check and see if i can find someone.

RAMitchell:
Im curious what data distribution leads to CSR being much more efficient than ellpack. It would have to have an uneven number of nonzero elements in each row right? Maybe we can just use CSR by default, let me think about it.

I wonder if there is a cleaner interface for expressing guidelines on how to run algorithms (cpu or gpu).

sriramch:
> Im curious what data distribution leads to CSR being much more efficient than ellpack. It would have to have an uneven number of nonzero elements in each row right? 

[sc] if the total number of features across the rows exhibits a larger span, and there are fewer rows skewed towards the end range, then the alternate representation can be efficient; or, a handful of training instances that are much wider relative to others etc. for instance, the number of features in the mortgage dataset across the rows range between 8 and 28 features. the number of rows in the higher end are fairly minimal compared to ones centered around the mean. a similar data distribution exists for others i have looked at as well - mnist8m, rcv1 etc.

> Maybe we can just use CSR by default, let me think about it.

[sc] right now, it chooses the most compact representation dynamically based on the training dataset. however, i have enabled this *only* when we are training using external memory, as my intent wasn't to cause performance regressions. the alternate representation would require 2 lookups, and thus can have small, yet a measurable impact to performance. i haven't measured it though.

sriramch:
@hcho3 would you be able to restart the build? the errors seem to be transient to me. better yet, is it possible for you to give us privileges to restart the builds via - https://xgboost-ci.net/job/xgboost/job/PR-x, if we want to do it ourselves without bothering you?

hcho3:
@sriramch Looks like the error is persisting even after restarting the test.

hcho3:
@sriramch @rongou @hetong007 Many pull requests are currently failing CI because of the following error
```
        elif issubclass(cls, ABCDataFrame):
            # Same for DataFrame and SparseDataFrame
            arith_flex = _arith_method_FRAME
            comp_flex = _flex_comp_method_FRAME
            arith_special = _arith_method_FRAME
            comp_special = _comp_method_FRAME
            bool_special = _arith_method_FRAME
>       return arith_flex, comp_flex, arith_special, comp_special, bool_special
E       UnboundLocalError: local variable 'arith_flex' referenced before assignment

/opt/python/lib/python3.7/site-packages/pandas/core/ops/__init__.py:719: UnboundLocalError
!!!!!!!!!!!!!!!!!!! Interrupted: 20 errors during collection !!!!!!!!!!!!!!!!!!!
```
and I just managed to re-produce this error locally using Docker. I will submit a PR to fix this soon.

RAMitchell:
> [sc] right now, it chooses the most compact representation dynamically based on the training dataset. however, i have enabled this only when we are training using external memory, as my intent wasn't to cause performance regressions. the alternate representation would require 2 lookups, and thus can have small, yet a measurable impact to performance. i haven't measured it though.

When you say two lookups do you mean the row size pointers? My intuition is that this will not be significant.

I dont want to cram multiple data structures into one using if statements, its hard to maintain. Im ok with doing this sort of thing using polymorphism but of course we cant use it on the device. Im thinking the way forward is to use CSR.

How much memory do you save from the CSR change and how much from the changes to use CPU algos for gradient/prediction? I want to do a cost benefit analysis for these changes.

sriramch:
> I dont want to cram multiple data structures into one using if statements, its hard to maintain. Im ok with doing this sort of thing using polymorphism but of course we cant use it on the device. Im thinking the way forward is to use CSR.

[sc] i'll think of someway to abstract this. i believe we need to have both representations, as dense and close to dense matrices will not benefit from the alternate representation (csr).

> 
> How much memory do you save from the CSR change and how much from the changes to use CPU algos for gradient/prediction? I want to do a cost benefit analysis for these changes.

[sc] if we use cpu algos for prediction/gradient computation, then on a per device, we can save 12 times the number of rows assigned to that device, if instances do not have weights; else, it is 16 times the number of rows assigned on the device. so, for ~100m rows/device, the savings is ~ 1.1-1.5gb

savings from csr isn't as straight forward, as it depends on the number of features in the dataset, number of histogram bins/feature, number of training instances assigned to the device etc. it could be in the order of several gb's. if you take the mortgage dataset for instance, the savings are in the order of ~ 2.5gb per device with the new representation for ~ 670m training instances. of course, this increases with more instances.

sriramch:
> @sriramch @rongou @hetong007 Many pull requests are currently failing CI because of the following error
> and I just managed to re-produce this error locally using Docker. I will submit a PR to fix this soon.

thanks @hcho3 - much appreciated!

trivialfis:
I can imagine the saved memory for extremely sparse dataset like URL with using CSR.   Let me take a closer look tomorrow.

RAMitchell:
> [sc] i'll think of someway to abstract this. i think we need to have both representations, as dense and close to dense matrices will not benefit from the alternate representation (csr).

Ellpack still uses 2n memory for dense matrices as it stores the column indices. Its not that different from CSR really. I can see it being useful to have dense and csr though if you can create the abstraction.

> [sc] if we use cpu algos for prediction/gradient computation, then on a per device basis, we can save 12 times the number of rows assigned to that device, if instances do not have weights; else, it is 16 times the number of rows assigned on the device. so, for ~100m rows/device, the savings is ~ 1.1-1.5gb

Im still dubious about added complexity for these gains. Trying to save memory per row will have little effect on wider problems.

sriramch:
> Ellpack still uses 2n memory for dense matrices as it stores the column indices. Its not that different from CSR really. I can see it being useful to have dense and csr though if you can create the abstraction.

[sc] for dense matrices the compressed buffer contains n_rows * row_stride (which should be number of features) number of entries. for the csr implementation, it would have to contain those many elements, plus additional overhead to carry the row indices which is needless. i have now removed those enums and kept a single state that specifies whether row indices are needed to stride over the compressed gidx buffer. trying to abstract this in multiple classes with just using templates without polymorphism on *device* is tricky, given that partially specialized types have to be known to the compiler and can't rely on dynamic values (if i have to decide the class type based on the layout type dynamically). 

> 
> > [sc] if we use cpu algos for prediction/gradient computation, then on a per device basis, we can save 12 times the number of rows assigned to that device, if instances do not have weights; else, it is 16 times the number of rows assigned on the device. so, for ~100m rows/device, the savings is ~ 1.1-1.5gb
> 
> Im still dubious about added complexity for these gains. Trying to save memory per row will have little effect on wider problems.

[sc] this is restricted only when cpu prediction is used *and* when external memory is enabled (which  aren't the default modes). this can be helpful when more data needs to be trained at the expense of degraded latency. it can save up to 2.5 to 3.5 gb's, which is a good chunk of gpu memory for large datasets. it can be tool in the toolbox that may be used, if needed.

RAMitchell:
I've been thinking about polymorphism on device - there will be a way to achieve this somehow it's just a question of how ugly it will be. There would be a lot of benefits - we could implement any kind of matrix data structure with minimal overhead including fully dense matrices. I have some other uses in mind too, like split evaluation.

sriramch:
> I've been thinking about polymorphism on device - there will be a way to achieve this somehow it's just a question of how ugly it will be. There would be a lot of benefits - we could implement any kind of matrix data structure with minimal overhead including fully dense matrices. I have some other uses in mind too, like split evaluation.

[sc] @RAMitchell - i'll try and take a crack at it and see how far i get with it. i'm not debating the benefits of such an approach; just not sure how feasible and elegant it may be. i tried to do this with template specializations to begin with, but this got out of hand fairly quickly. if there is a way, i can try and demonstrate the approach for this use-case first, and perhaps broaden it up for adoption in other places later. can i work on it in a following pr? if so, can you merge this if it looks ok to you?

RAMitchell:
Im not comfortable with specialising behaviour for external memory in this way and the benefit is proportional to number of rows only. It's also not a configurable behaviour for a user - if they run external memory these functions occur on the CPU regardless of if they have enough memory. I don't know if this is the optimal behaviour for every situation.

I do want to pursue different data structures though I think this will have a 
big impact.

RAMitchell:
Im not comfortable with specialising behaviour for external memory in this way and the benefit is proportional to number of rows only. It's also not a configurable behaviour for a user - if they run external memory these functions occur on the CPU regardless of if they have enough memory. I don't know if this is the optimal behaviour for every situation.

I do want to pursue different data structures though I think this will have a 
big impact.

RAMitchell:
Re. polymorphism, the object just needs to be allocated in local kernel memory for the vtable to work. I was thinking we pass the base class in as a kernel parameter then use its copy constructor to reinstantiate it inside the kernel. The reinstantiation could be hidden inside a function

src/metric/elementwise_metric.cu
```diff
@@ -346,7 +346,10 @@ struct EvalEWiseBase : public Metric {
         << "hint: use merror or mlogloss for multi-class classification";
     const auto ndata = static_cast<omp_ulong>(info.labels_.Size());
     // Dealing with ndata < n_gpus.
-    GPUSet devices = GPUSet::All(tparam_->gpu_id, tparam_->n_gpus, ndata);
+    // Perform the metrics computation on CPU if device list isn't present in prediction vector
```
trivialfis: I think there can a better heuristic on this configuration.  Please try to document the proposed behavior.

src/objective/hinge.cu
```diff
@@ -38,6 +38,8 @@ class HingeObj : public ObjFunction {
     const bool is_null_weight = info.weights_.Size() == 0;
     const size_t ndata = preds.Size();
     out_gpair->Resize(ndata);
+    GPUSet devices = (preds.Devices().IsEmpty() && tparam_->external_memory)
```
trivialfis: You can make a policy for this configuration, and abstract the policy into a method of `GPUSet` or another class.  But again I think it's not the best heuristic.

src/objective/multiclass_obj.cu
```diff
@@ -59,14 +59,17 @@ class SoftmaxMultiClassObj : public ObjFunction {
     const int nclass = param_.num_class;
     const auto ndata = static_cast<int64_t>(preds.Size() / nclass);
 
-    auto devices = GPUSet::All(tparam_->gpu_id, tparam_->n_gpus, preds.Size());
-    out_gpair->Shard(GPUDistribution::Granular(devices, nclass));
-    info.labels_.Shard(GPUDistribution::Block(devices));
-    info.weights_.Shard(GPUDistribution::Block(devices));
-    preds.Shard(GPUDistribution::Granular(devices, nclass));
+    GPUSet devices = (preds.Devices().IsEmpty() && tparam_->external_memory)
```
trivialfis: What's wrong with resharding to empty device?

src/tree/updater_gpu_hist.cu
```diff
@@ -139,49 +139,95 @@ __forceinline__ __device__ int BinarySearchRow(
  * device. Does not own underlying memory and may be trivially copied into
  * kernels.*/
 struct ELLPackMatrix {
+  /*! \brief How is the compressed data laid out? */
```
trivialfis: I'm ok with specialization as optimization is never easy.  But I would like to have a very distinguishable and organized implementations for each specialization.  Polymorphism is one way to do it, another way may be by creating a dispatcher.

sriramch:
> and the benefit is proportional to number of rows only. It's also not a configurable behaviour for a user 

[sc] please see the next response. 8-12 bytes per row can quickly add up for larger number of rows (1.1-1.5gb for 100m rows), which can be used to accommodate a slightly larger dataset.

> - if they run external memory these functions occur on the CPU regardless of if they have enough memory. I don't know if this is the optimal behaviour for every situation.

[sc] very valid point. i agree that it is not best to overload the external memory configuration for this. i'll introduce a config option to move this to the cpu on a need-be basis.

> 
> I do want to pursue different data structures though I think this will have a
> big impact.

[sc] can we keep the changes to the ellpack matrix then? i'll work on the polymorphism on device in a follow-up pr?

src/objective/hinge.cu
sriramch: @trivialfis - agree that this is not a better heuristic. i'll add another configuration item to control this.

src/objective/multiclass_obj.cu
sriramch: possibly nothing. however, sharding without any devices (or to no device) doesn't seem right, and hence i made this change.

sriramch:
@RAMitchell @trivialfis - *summary*

- i'll add a configuration item to move base predictions/metric computation on the cpu on a need-be basis without reusing the external memory configuration for this

- i'll work on polymorphism or other means to clearly abstract different data layout representations into its own type in a *follow up* pr?

does this sound ok to you?

RAMitchell:
I still dont think the configuration changes are worth it, and the CSR change is too messy.

I think we should try to achieve memory reduction by allowing different data structures via #4686. If we merged your CSR changes how would it interact with that? Could we have a GPUCSRPage and GPUEllPackPage?  Another key question is can we select the optimal page type at runtime given that @rongou has gone with the template approach?

sriramch:
> I think we should try to achieve memory reduction by allowing different data structures via #4686. If we merged your CSR changes how would it interact with that? Could we have a GPUCSRPage and GPUEllPackPage? 

[sc] i have cleaned up the monolithic ellpackmatrix implementation to now create different matrix types, which can then be manipulated polymorphically *on the device*. please look.

honestly, i do not have a clue as to how #4686 is going to progress. i'm presuming that compressed bins are going to be written (to memory/disk), which are then going to be streamed in to build the trees via `matrix->GetBatches<CompressedPage>`? is it going to be in batches or one page for the whole matrix - i'm not sure. the compress kernel may have to be used to do the binning, which can then create an appropriate matrix type on the gpu based on the representation that is compact on the gpu.

> Another key question is can we select the optimal page type at runtime given that @rongou has gone with the template approach?

[sc] even with the template based approach, it would have to be coded in at least one place to invoke the get batches method with the appropriate page type based on some configuration/heuristic (there has to be an if block somewhere).

PR #4701, opened 2019-07-24
Remove VC-2013 support.

trivialfis:
* Removing it as it is not fully c++11 compliance.


trivialfis:
@hcho3 @RAMitchell @CodingCat

trivialfis:
Any objection? ;-)

PR #4706, opened 2019-07-25
[jvm-package] jvm test should clean up after themselfs

trams:
I noticed that some tests do not fully clean up their tmp files so I fixed that.
Nothing else

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManagerSuite.scala
+++ b/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManagerSuite.scala
@@ -17,12 +17,11 @@
 package ml.dmlc.xgboost4j.scala.spark
 
 import java.io.File
-import java.nio.file.Files
 
-import org.scalatest.{BeforeAndAfterAll, FunSuite}
+import org.scalatest.FunSuite
 import org.apache.hadoop.fs.{FileSystem, Path}
 
-class CheckpointManagerSuite extends FunSuite with PerTest with BeforeAndAfterAll {
+class CheckpointManagerSuite extends FunSuite with TmpFolderPerSuite with PerTest {
 
   private lazy val (model4, model8) = {
     val training = buildDataFrame(Classification.train)
@@ -33,7 +32,7 @@ class CheckpointManagerSuite extends FunSuite with PerTest with BeforeAndAfterAl
   }
 
   test("test update/load models") {
-    val tmpPath = Files.createTempDirectory("test").toAbsolutePath.toString
+    val tmpPath = createTmpFolder("test").toAbsolutePath.toString
     val manager = new CheckpointManager(sc, tmpPath)
     manager.updateCheckpoint(model4._booster)
     var files = FileSystem.get(sc.hadoopConfiguration).listStatus(new Path(tmpPath))
@@ -49,7 +48,7 @@ class CheckpointManagerSuite extends FunSuite with PerTest with BeforeAndAfterAl
   }
 
   test("test cleanUpHigherVersions") {
-    val tmpPath = Files.createTempDirectory("test").toAbsolutePath.toString
+    val tmpPath = createTmpFolder("test").toAbsolutePath.toString
     val manager = new CheckpointManager(sc, tmpPath)
     manager.updateCheckpoint(model8._booster)
     manager.cleanUpHigherVersions(round = 8)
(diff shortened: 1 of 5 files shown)


CodingCat:
which test doesn't clean? the PR looks like extracting temp dir creation to a class instead of adding anything?

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PersistenceSuite.scala
```diff
@@ -124,6 +98,8 @@ class PersistenceSuite extends FunSuite with PerTest with BeforeAndAfterAll {
   }
 
   test("test persistence of MLlib pipeline with XGBoostClassificationModel") {
+    val tempDir = createTmpFolder("MLLib").toFile
```
CodingCat: why we need to create tempDir in each test manually?

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/TmpFolderPerSuite.scala
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package ml.dmlc.xgboost4j.scala.spark
+
+import java.nio.file.{Files, Path}
+
+import org.apache.spark.network.util.JavaUtils
+import org.scalatest.{BeforeAndAfterAll, FunSuite}
+
+trait TmpFolderPerSuite extends BeforeAndAfterAll { self: FunSuite =>
```
CodingCat: what's the issue to put it in PerTest?

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PersistenceSuite.scala
trams: As far as I know in scalatest different tests are not isolated.
I create a temp folder once per suite. So one need to create temp folders in each test explicitly to make them isolated.

Alternative solution is to change a trait to create a temp folder per test

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/TmpFolderPerSuite.scala
trams: I wanted to have an extra flexibility if we want to have a test which create temp files but we do not need SparkSession

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PersistenceSuite.scala
CodingCat: but what's the problem with reusing the created dir here?

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManagerSuite.scala
```diff
@@ -17,12 +17,11 @@
...
 
 import java.io.File
-import java.nio.file.Files
 
-import org.scalatest.{BeforeAndAfterAll, FunSuite}
+import org.scalatest.FunSuite
 import org.apache.hadoop.fs.{FileSystem, Path}
 
-class CheckpointManagerSuite extends FunSuite with PerTest with BeforeAndAfterAll {
+class CheckpointManagerSuite extends FunSuite with TmpFolderPerSuite with PerTest {
```
CodingCat: I may miss something, why this change can address the issue?

is it just moving the beforeAll&afterAll to somewhere else?

trams: Before this change in CheckpointManagerSuite each test created the temp directory in /tmp (on linux) and never removed it.

(Note that PerTest neither did not nor does not now create any temp directories. It just create a SparkSession to use)

After this change CheckpointManagerSuite will include TmpFolderPerSuite .
TmpFolderPerSuite  will create a temp directory and then it will be removed.
Also after this change all temp folders are created as children of this root temp folder (see line 35 in this file)

CodingCat: oh. yes, we manually created "test" dir there....

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PersistenceSuite.scala
CodingCat: I would prefer keeping writing test easy by forcing devs to remember call createTmpFolder manually,

if reusing the directory doesn't bring any issue, can we just reuse the created one in beforeAll()?

CodingCat: @trams would you please address this so that we can merge it?

trams:
Yes, I will address it

trams:
@CodingCat , I updated review as you asked. I also rebased on a fresh master and CI gives a green light.

CodingCat:
thanks, merged!

PR #4702, opened 2019-07-24
[WIP] Initial commit for cudf integration.

trivialfis:
closes #4696.  Let's keep that PR as a reference for now.

PR #4708, opened 2019-07-25
A simple Json implementation for future use.

trivialfis:
This will enable #4702 and #4683 (extracted from there).  Other than serialization,  the JSon implementation will be used to consume [numpy array interface](https://docs.scipy.org/doc/numpy/reference/arrays.interface.html), and [cuda array interface](numba.pydata.org/numba-doc/latest/cuda/cuda_array_interface.html).  This will benefit XGBoost as creating the generic interface will enable introp with lots of existing libraries (~arrow and friends~, cupy, numba etc..) without creating specific C backend source code.

A json implementation will the first part of on going effort for above described features.  This PR doesn't touch any existing code base, can we consider merging it although it's not used by other code?

### DIFF
--- a/include/xgboost/json.h
+++ b/include/xgboost/json.h
@@ -1,0 +1,530 @@
+/*!
+ * Copyright (c) by Contributors 2019
+ */
+#ifndef XGBOOST_JSON_H_
+#define XGBOOST_JSON_H_
+
+#include <xgboost/logging.h>
+
+#include <string>
+
+#include <map>
+#include <memory>
+#include <vector>
+#include <functional>
+#include <utility>
+
+namespace xgboost {
+
+class Json;
+class JsonReader;
+class JsonWriter;
+
+class Value {
+ public:
+  /*!\brief Simplified implementation of LLVM RTTI. */
+  enum class ValueKind {
+    String,
+    Number,
+    Integer,
+    Object,  // std::map
+    Array,   // std::vector
+    Raw,
+    Boolean,
+    Null
+  };
+
+  explicit Value(ValueKind _kind) : kind_{_kind} {}
+
+  ValueKind Type() const { return kind_; }
+  virtual ~Value() = default;
+
+  virtual void Save(JsonWriter* writer) = 0;
+
+  virtual Json& operator[](std::string const & key) = 0;
+  virtual Json& operator[](int ind) = 0;
+
+  virtual bool operator==(Value const& rhs) const = 0;
+  virtual Value& operator=(Value const& rhs) = 0;
+
+  std::string TypeStr() const;
+
+ private:
+  ValueKind kind_;
+};
+
+template <typename T>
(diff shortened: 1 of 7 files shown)


trivialfis:
@rongou @sriramch Please help reviewing.

src/common/json.cc
```diff
@@ -0,0 +1,608 @@
...
+    } else if (ch == '\r') {
+      buffer += "\\r";
+    } else if (ch == '\t') {
+      buffer += "\\t";
+    } else if (static_cast<uint8_t>(ch) <= 0x1f) {
+      // Unit separator
+      char buf[8];
+      snprintf(buf, sizeof buf, "\\u%04x", ch);
+      buffer += buf;
+    } else {
```
trams: I would suggest to have another branch static_cast<uint8_t>(ch) > 127 in which we would fail serialization. Supporting UTF8 encoded Unicode strings is hard and I suggest at least for now limit ourself to the realm of simple strings

src/common/json.cc
```diff
@@ -0,0 +1,608 @@
...
+
+  if (ch == 't') {
+    for (size_t i = 0; i < 3; ++i) {
+      buffer.push_back(GetNextNonSpaceChar());
+    }
+    if (buffer != u8"rue") {
+      Error("Expecting boolean value \"true\".");
+    }
+    result = true;
+  } else {
```
trams: Beware this function would parse "xalse" as false

include/xgboost/json.h
```diff
@@ -0,0 +1,530 @@
...
+  Value const& GetValue() const & { return *ptr_; }
+  Value const& GetValue() &&      { return *ptr_; }
+  Value&       GetValue() &       { return *ptr_; }
+
+  bool operator==(Json const& rhs) const {
+    return *ptr_ == *(rhs.ptr_);
+  }
+
+ private:
+  std::shared_ptr<Value> ptr_;
```
trams: What do you think about using std::unique_ptr here
I looked though this pull request and your another (Json format) pull request and it seems we never actually have two Json objects pointing to the same instance of Value. We always move them around.

Benefits of unique_ptr is
1. easier to reason
2. no cycles
3. shared ownership could lead to hard-to-debug problems

trivialfis:
@trams Yes.  I need to split that PR into a few dependent PR.  I will address your comment from that PR here.  :-). Thanks.

include/xgboost/json.h
trivialfis: I plan to implement a pool allocator in combination with dmlc:: Any in the future for better performance.

trivialfis:
@RAMitchell Removed the friend class.

trivialfis:
Since it doesn't modify existing code base,  I incline to merge it here with all related comments addressed as now I'm working on the cudf intergration which will make use of this Json implementation.  Any objection?  @hcho3 @CodingCat

Issue #4679, opened 2019-07-18
[BLOCKING] Per-node sync slows down distributed training with 'hist'


CodingCat:
looks like latest branch has per-node sync again https://github.com/dmlc/xgboost/blob/4ef6d216b9989f0ea7101fcbecefe25f6125c195/src/tree/updater_quantile_hist.cc#L615-L617

CodingCat:
and the grads is also synced per node 

https://github.com/dmlc/xgboost/blob/4ef6d216b9989f0ea7101fcbecefe25f6125c195/src/tree/updater_quantile_hist.cc#L643 (only needed for root actually , check https://github.com/dmlc/xgboost/pull/4140)

trivialfis:
Seems the optimization PR didn't make a clean rebase.

CodingCat:
more discussion at https://github.com/dmlc/xgboost/pull/4716

hcho3:
Re-opening this, since #4716 fixed correctness but did not address performance. @thvasilo points out that syncing for every nodes leads to significant slow-down due to extra communication.

thvasilo:
Copying from the other issue:  While the second per-node sync seems straightforward to fix (see [this WIP](https://github.com/thvasilo/xgboost/commit/8196a16e5732b579addb5e5b7ac3339bd0545758#diff-10aac201acca00858b390ff0d679434fR655)) getting rid of the [first loop](https://github.com/dmlc/xgboost/blob/f0064c07ab70f1c770bb0e9eb64fe85afb3a3d73/src/tree/updater_quantile_hist.cc#L616) could entail changes to how we build and use the `hist_` container (of HistCollection type) which would require a lot of effort to untangle.

hcho3:
@thvasilo Note that this issue is not present in release 0.90. I believe #4529 introduced it. A proper fix would need to first analyze #4529 in detail and selectively revert the parts that cause per-node sync. In 0.90, `hist` performs per-level sync (unless `grow_policy` is specifically set to `lossguide`).

PR #4716, opened 2019-07-30
[HOTFIX] distributed training with hist method 

sperlingxx:
Problem: 
- Distributed training with hist method will stop working when synchronizing histograms of different machines.

Debug Stack:
-   Size of `expand_nodes` in different workers are not equal, because one node may be a leaf node on one worker, but it is still splitable on other workers.
-  Results of `SplitEvaluation` are inconsistent on different workers (machines).
-  Inconsistent split results will be observed only when `nthread > 1`.
-  When I changed OPENMP schedule policy of `EvaluateSplitsBatch` from `schedule(guided)` to `schedule(dynamic) num_threads(nthread)`,  everything works.

### DIFF
--- a/src/common/hist_util.cc
+++ b/src/common/hist_util.cc
@@ -48,6 +48,7 @@ void HistogramCuts::Build(DMatrix* dmat, uint32_t const max_num_bins) {
     DenseCuts cuts(this);
     cuts.Build(dmat, max_num_bins);
   }
+  LOG(INFO) << "Total number of hist bins: " << cut_ptrs_.back();
 }
 
 bool CutsBuilder::UseGroup(DMatrix* dmat) {
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -1079,7 +1079,8 @@ void QuantileHistMaker::Builder::EvaluateSplitsBatch(
   // partial results
   std::vector<std::pair<SplitEntry, SplitEntry>> splits(tasks.size());
   // parallel enumeration
-#pragma omp parallel for schedule(guided)
+  const auto nthread = omp_get_num_threads();
+  #pragma omp parallel for schedule(dynamic) num_threads(nthread)
   for (omp_ulong i = 0; i < tasks.size(); ++i) {
     // node_idx : offset within `nodes` list
     const int32_t  node_idx    = tasks[i].first;
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -556,7 +556,7 @@ void QuantileHistMaker::Builder::BuildHistsBatch(const std::vector<ExpandEntry>&
       reinterpret_cast<const GradientPair::ValueT*>(gpair.data());
 
   // 2. Build partial histograms for each node
-  #pragma omp parallel for schedule(guided)
+  #pragma omp parallel for schedule(static)
   for (int32_t itask = 0; itask < n_hist_buidling_tasks; ++itask) {
     const size_t tid = omp_get_thread_num();
     const int32_t nid = task_nid[itask];
@@ -856,7 +856,7 @@ bool QuantileHistMaker::Builder::UpdatePredictionCache(
     }
   }
 
-#pragma omp parallel for schedule(guided)
+#pragma omp parallel for schedule(static)
   for (omp_ulong k = 0; k < tasks_elem.size(); ++k) {
     const RowSetCollection::Elem rowset = tasks_elem[k];
     if (rowset.begin != nullptr && rowset.end != nullptr && rowset.node_id != -1) {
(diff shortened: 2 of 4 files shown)


CodingCat:
is it only happened in master branch?

sperlingxx:
> is it only happening in master branch?

@CodingCat  Yes, I have tested branch `release_0.9`,  everything is all right.  I think some modifications in PR #4529 cause the problem.

CodingCat:
> is it only happening in master branch?

I think it is related to issue at https://github.com/dmlc/xgboost/issues/4679

in 0.9 branch, the node stats is synced for only once when working on the root and left/right should be calculated from cache (check https://github.com/dmlc/xgboost/pull/4140) so I think everything should be fine

@hcho3 @trivialfis @RAMitchell I believe it is a blocking issue for 1.0?

CodingCat:
> > is it only happening in master branch?
> 
> @CodingCat Yes, I have tested branch `release_0.9`, everything is all right. I think some modifications in PR #4529 cause the problem.

Bingo! check my previous comment

trivialfis:
> I believe it is a blocking issue for 1.0?

Could you please add this to the roadmap as you have better idea for what's happening.

hcho3:
@CodingCat Yes, I think this is blocking.

@sperlingxx Thanks for the report. FYI, see https://xgboost.readthedocs.io/en/latest/contrib/unit_tests.html for locally running tests

CodingCat:
I have updated 1.0.0 roadmap

sperlingxx:
@hcho3  Thanks for doc link, I will tune unit tests in local.

sperlingxx:
@hcho3 @trivialfis @CodingCat 
I've tested codes of master branch with below script:
```python
import xgboost as xgb
import unittest
import numpy as np


class TestOMP(unittest.TestCase):
    def test_omp(self):
        dpath = 'demo/data/'
        dtrain = xgb.DMatrix(dpath + 'agaricus.txt.train')
        dtest = xgb.DMatrix(dpath + 'agaricus.txt.test')

        param = {'booster': 'gbtree',
                 'grow_policy': 'depthwise',
                 'tree_method': 'hist',
                 'eval_metric': 'error',
                 'max_depth': 5,
                 'min_child_weight': 0}

        watchlist = [(dtest, 'eval'), (dtrain, 'train')]
        num_round = 5

        def run_trial():
            res = {}
            bst = xgb.train(param, dtrain, num_round, watchlist, evals_result=res)
            metrics = [res['train']['error'][-1], res['eval']['error'][-1]]
            preds = bst.predict(dtest)
            return metrics, preds

        def consist_test(title, n):
            auc, pred = run_trial()
            for i in range(n-1):
                auc2, pred2 = run_trial()
                try:
                    assert auc == auc2
                    assert np.array_equal(pred, pred2)
                except Exception as e:
                    print('-------test %s failed, num_trial: %d-------' % (title, i))
                    raise e
                auc, pred = auc2, pred2
            return auc, pred

        print('test approx ...')
        param['tree_method'] = 'approx'

        param['nthread'] = 1
        auc_1, pred_1 = consist_test('approx_thread_1', 100)

        param['nthread'] = 2
        auc_2, pred_2 = consist_test('approx_thread_2', 100)

        param['nthread'] = 3
        auc_3, pred_3 = consist_test('approx_thread_3', 100)

        assert auc_1 == auc_2 == auc_3
        assert np.array_equal(auc_1, auc_2)
        assert np.array_equal(auc_1, auc_3)

        print('test hist ...')
        param['tree_method'] = 'hist'

        param['nthread'] = 1
        auc_1, pred_1 = consist_test('hist_thread_1', 100)

        param['nthread'] = 2
        auc_2, pred_2 = consist_test('hist_thread_2', 100)

        param['nthread'] = 3
        auc_3, pred_3 = consist_test('hist_thread_3', 100)

        assert auc_1 == auc_2 == auc_3
        assert np.array_equal(auc_1, auc_2)
        assert np.array_equal(auc_1, auc_3)
```

`approx method`  passed all the tests, but `hist method` always failed in consistent tests when thread num > 1. (If we reduce trial num, sometimes test of `hist method` will be passed). 

After I changed all OPENMP scheduling policys in updater_quantile_hist.cc  to `static` (from `guided` ,  consistent tests of `hist method` can be passed.

There exists three call of OPENMP parallel for in updater_quantile_hist.cc :
https://github.com/dmlc/xgboost/blob/cb9a80ca900429335f565c2a811e9c8aee66dd87/src/tree/updater_quantile_hist.cc#L559 

https://github.com/dmlc/xgboost/blob/cb9a80ca900429335f565c2a811e9c8aee66dd87/src/tree/updater_quantile_hist.cc#L859

https://github.com/dmlc/xgboost/blob/cb9a80ca900429335f565c2a811e9c8aee66dd87/src/tree/updater_quantile_hist.cc#L1082

trivialfis:
@hcho3 Do you have time to investigate impact from previous optimization PR?  I have a feeling that we might need to revert some problematic parts of it.

hcho3:
@trivialfis I'll probably have to take time to investigate. I'll let you know when I do so. My org has interest in fixing this problem as well.

PR #4730, opened 2019-08-03
Specify version macro in CMake.

trivialfis:


CMakeLists.txt
```diff
@@ -13,6 +13,12 @@ if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.0)
   message(FATAL_ERROR "GCC version must be at least 5.0!")
 endif()
 
+message(STATUS "xgboost VERSION: ${xgboost_VERSION}")
```
hcho3: Where is this variable defined?

trivialfis: @hcho3 Defined by `project` command.  See https://cmake.org/cmake/help/latest/command/project.html .

Issue #4653, opened 2019-07-10
[python-package] include dmlc_tracker into python-package


sperlingxx:
For now,  `dmlc_tracker/tracker.py` is copied into python-package on the fly, during the python installation. 

```python
# Get dmlc tracker script
shutil.copy('../dmlc-core/tracker/dmlc_tracker/tracker.py', 'xgboost/')
```

Maybe there exists a better approach, include dmlc_tracker into xgboost python-package via `__init__.py`.  Something like: 

```python
try:
    from .dmlc-core.tracker.dmlc_tracker.tracker import RabitTracker
    ....
```

sperlingxx:
@RAMitchell @hcho3  What do you think?  I am grad to work on this.

trivialfis:
@sperlingxx Feel free to open a PR.  :)

hcho3:
Solved by #4731

PR #4731, opened 2019-08-04
[python package] include dmlc-tracker into xgb python pkg

sperlingxx:
This PR is a response to issue #4653 .

I made a specialized copy of `dmlc_tracker/tracker.py` in xgboost, because:
1.  symbolic link doesn't work well under  _Windows_
2.  original version of `dmlc_tracker/tracker.py` has a lot of linter problem
3.  Future modifications can be applied on this copy without considering compatibility.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,6 +96,3 @@ plugin/updater_gpu/test/cpp/data
 # files from R-package source install
 **/config.status
 R-package/src/Makevars
-
-# Python install
-python-package/xgboost/tracker.py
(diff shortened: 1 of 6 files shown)


sperlingxx:
It seems that something irrelevant with this PR make tests failed, how can I rerun it?

trivialfis:
Restarted, seems to be a network issue?

trivialfis:
@sperlingxx The socket seems pointing only to local host port?

sperlingxx:
@trivialfis I made some silly mistakes while fixing linter problem of RabitTracker.

Issue #4639, opened 2019-07-04
Deprecate XGBRegressor.save_model() / XGBClassifier.save_model()


hcho3:
Many methods of `XGBClassifier` relies on the existence of certain fields, such as `self.objective`, `self.missing`, and `self.n_classes_` (these fields exist to ensure interoperability with scikit-learn) and these fields are **not** preserved by the call to `XGBClassifier.save_model()`, as the binary serialization format is not aware of these fields. See #4517, where `XGBClassifier.save_model()` breaks `predict_proba()` for multi-class classifiers.

We should deprecate  `XGBRegressor.save_model()` / `XGBClassifier.save_model()` / `XGBRanker.save_model()` method, because its name is really misleading. Instead, we should re-name it to `export_model_for_external_binding()`.

trivialfis:
I would like to first understand the difference between languages.  I think I might have a chance to join the development of JVM package in near future, so can we hold this off a bit and give me a chance to understand what's the best way to save the model?

trivialfis:
@hcho3 I made the differences clear between model IO and memory based serialization in https://github.com/dmlc/xgboost/pull/4732  and with a RFC here: #4855 .  Feel free to reopen as I'm currently tracking down all related issues can be resolved by JSON PR.

hcho3:
@trivialfis Does your RFC address the Python objects used specifically for scikit-learn binding? We may want to port these objects into the C++ core if we want to aim cross-language compatibility.

trivialfis:
@hcho3 We can add extra code in scikit learn interface for restoring these objects.

hcho3:
@trivialfis What I mean is that semantics of the extra Python objects are lost when the model is serialized and then deserialized into, say, XGBoost4J-Spark.

trivialfis:
@hcho3 Let me try a fix.

trivialfis:
@hcho3 After we have JSON IO, we can append extra fields into the model like `scikit-learn-model`.  This won't effect internal model IO as core C++ doesn't look beyond what's needed.  Just to note that Python JSON implementation has a fast parser but a slow generator.

trivialfis:
Attributes in booster are still lost, but has nothing to do with skl now.  Closing.

Issue #4618, opened 2019-06-28
Check failed: n_gpus <= n_available_devices (1 vs. -4) Starting from gpu id: 6, there are only -4 available devices, while n_gpus is set to: 1


pseudotensor:
```
xgboost.core.XGBoostError: [12:18:20] /root/repo/xgboost/include/xgboost/../../src/common/common.h:193: Check failed: n_gpus <= n_available_devices (1 vs. -4) Starting from gpu id: 6, there are only -4 available devices, while n_gpus is set to: 1

Stack trace returned 10 entries:
[bt] (0) /home/stefanp/RB/scoring-pipeline/env/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(dmlc::StackTrace(unsigned long)+0x54) [0x7fc7fdd22604]
[bt] (1) /home/stefanp/RB/scoring-pipeline/env/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x1d) [0x7fc7fdd22d9d]
[bt] (2) /home/stefanp/RB/scoring-pipeline/env/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(xgboost::GPUSet::All(int, int, int)+0x527) [0x7fc7fdf18927]
[bt] (3) /home/stefanp/RB/scoring-pipeline/env/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(xgboost::obj::RegLossObj<xgboost::obj::LinearSquareLoss>::Configure(std::vector<std::pair<std::string, std::string>, std::allocator<std::pair<std::string, std::string> > > const&)+0x233) [0x7fc7fdf4ec33]
[bt] (4) /home/stefanp/RB/scoring-pipeline/env/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(void xgboost::ObjFunction::Configure<std::_Rb_tree_iterator<std::pair<std::string const, std::string> > >(std::_Rb_tree_iterator<std::pair<std::string const, std::string> >, std::_Rb_tree_iterator<std::pair<std::string const, std::string> >)+0xd1) [0x7fc7fdda49d1]
[bt] (5) /home/stefanp/RB/scoring-pipeline/env/lib/python3.6/site-packages/xgboost/./lib/libxgboost.so(xgboost::LearnerImpl::Load(dmlc::Stream*)+0xc9a) [0x7fc7fddb2f6a]
[...]

Issue #4072, opened 2019-01-21
Model Serialization and Endianness


DoanHuuJacques:
I run XGBoost dans save the model on Big Endian platform (eg: SUSE/PPC64 Big Endian).
This native serialisation cannot be reloaded on Little Endian platforms (eg: Win64/Linux64 on Intel).
Indeed, when looking at how XGBoost saves the model via Learner::Save() method, the internal C structures are serialized as is with endianness awareness by using helper function from serializer.h
I expect to have a neutral model binary format regardless of the running platform/chipset.

Would you confirm that issue?
Thanks

trivialfis:
@DoanHuuJacques We expect solving it by #3980 .  But not much progress has been made so far.  See https://github.com/hcho3/xgboost-json-schema/pull/5

DoanHuuJacques:
@trivialfis 
JSON format is actually a better/substainable option than raw binary format. Would you confirm that the current serialisation is not endianness aware?
Thanks

trivialfis:
@DoanHuuJacques It's **not** endianness aware.  @hcho3 received similar report before I think. :)

DoanHuuJacques:
@trivialfis , @hcho3 
Hi,
Is there an ETA for the JSON serialisation support?
Is backward compatibility of current binary format supported? typically, can a runtime 0.81 load a serialized model 0.72?
Thanks

trivialfis:
> Is there an ETA for the JSON serialisation support?

Not yet.

> Is backward compatibility of current binary format supported?

Yes.  Binary format should linger around.

trivialfis:
I believe the endianness issue is resolved in dmlc-core.

DoanHuuJacques:
@trivialfis 
would you mind elaborating a bit on how the endianness has been resolved in dmlc-core? Do you have a pull request related to this fix
Thanks

hcho3:
@trivialfis Can you point me to a relevant pull request?

@DoanHuuJacques @trivialfis submitted a series of pull requests recently to implement JSON serialization. I believe this will solve the problem with endianness.

DoanHuuJacques:
thanks @hcho3 for the information
I'm aware that @trivialfis  is working on JSON serialization and it seems that this feature is still under development (not yet merged)

trivialfis:
@hcho3 https://github.com/dmlc/dmlc-core/pull/406 https://github.com/dmlc/xgboost/issues/3916#issuecomment-443111346

[...]

Issue #3980, opened 2018-12-08
RFC: JSON as Next-Generation Model Serialization Format


hcho3:
# RFC: JSON as Next-Generation Model Serialization Format

1. Motivation: Why a new serialization format?
2. Proposal: Use JSON serialization to ensure backward and forward compatibility
3. Semantic versioning
4. Other benefits of JSON
5. Implementation of the proposal with DMLC JSON parser
6. Addressing potential objections
7. Appendix A: Full 1.0 Schema
8. Appendix B: Representation of missing values for various data types

In this RFC document, I propose that XGBoost project would eventually migrate to JSON for serializing decision tree models.

Special thanks to @trivialfis for initially coming up with the idea of using JSON.

## Motivation: Why a new serialization format?
The current method of serialization used in XGBoost is a binary dump of model parameters. For instance, the decision tree class (`RegTree`) is serialized as follows:

```cpp
/* Adopted from include/xgboost/tree_model.h; simplified for presentation */
void Save(dmlc::Stream* fo) {
  // Save tree-wide parameters
  fo->Write(&param, sizeof(TreeParam));
  // Save tree nodes
  fo->Write(nodes_.data(), sizeof(Node) * nodes_.size());
  // Save per-node statistics
  fo->Write(stats_.data(), sizeof(NodeStat) * stats_.size());
  // Save optional leaf vector
  if (param.size_leaf_vector != 0) {
    fo->Write(leaf_vector_);
  }
}
void Load(dmlc::Stream* fi) {
  // Load tree-wide parameters
  fi->Read(&param, sizeof(TreeParam));
  nodes_.resize(param.num_nodes);
  stats_.resize(param.num_nodes);
  // Load tree nodes
  fi->Read(nodes_.data(), sizeof(Node) * nodes_.size());
  // Load per-node statistics
  fi->Read(stats_.data(), sizeof(NodeStat) * stats_.size());
  // Load optional leaf vector
  if (param.size_leaf_vector != 0) {
    fi->Read(&leaf_vector_);
  }
}
```
The binary serialization method has a few benefits: it is straight-forward to implement, has low memory overhead (since we can read from and write to file streams directly), is fast to run (no pre-processing required), and produces compact model binaries.

[...]

PR #4732, opened 2019-08-04
Add JSON IO for various components.

trivialfis:
* Split up model IO and serialization.
* Unify the use of `num_feature`, `base_score` and `num_output_group` across XGBoost with `learner_model_param_` (`LearnerModelParam`).  This is passed as a pointer to const across XGBoost.  Deprecate all duplicated parameter.  Mark original model parameter as legacy.
* Unify `num_output_group` and `num_class`.
* Add JSON for both model IO and serialisation.
* Remove duplicated parameter in split evaluator.
* Fix wrong predict contribution tests.
* Add tests for JSON IO in both cxx and Python.
* Save updaters for carrying out training continuation.
* Rigorous tests for training continuation.
* Define prediction cache in `gbtree` instead of `predictor`, as the cache needs to be in sync between `cpu_predictor` and `gpu_predictor`.
* Add basic documentation for the serialisation format.
* Enabled save/load config in Python pickle.

Todos:
- [x] Use a model parameter which contains `num_output_group`, `num_feature`, `base_score` across XGBoost.
- [x] Extract interaction constraint into an independent evaluator.
- [x] Remove num roots. #5059
- [x] Merge `num_output_group` and `num_class`, remove related configurations.
- [x] Address reviewers' comments.
- [x] Add configuration for runtime related parameters including `gpu_id`. Closes  #5072 .
- [ ] Add configuration for `nthreads`.  This should be done together with #4975.
- [x] Remove `gpu_id` and `gpu_predictor` from binary model.
- [x] Fix prediction cache for Multi-Class.
- [x] Track down related issues:
  + [x] #4878
  + [x] #4877 [Won't fix as that's related to CUDA context]
  + [x] #4855
  + [x] #4755
  + [x] #4758 [Should be fixed, but have to confirm]
  + [x] #4639 [Will fix in next PR]
  + [x] #4618
  + [x] #4072 [Fixed in both binary and JSON]
  + [x] #3980
  + [x] #3771 [Will not fix as this is a feature request]
  + [x] #5003
  + [x] #5023
  + [x] https://github.com/slundberg/shap/pull/520 [Will fix after we merge charconv and rewrite their parser).
  + [x] #4837 [Waiting for feedback]


trivialfis:
@hetong007 Do you have any suggestion for releasing CRAN source distribution with CMake instead of Make?

trivialfis:
Also @pommedeterresautee ;-).

hetong007:
I'm probably the least experienced developer of XGBoost in terms of CMake. Adding @khotilov as another potential source of suggestions.

src/c_api/c_api.cc
```diff
@@ -891,23 +895,51 @@ XGB_DLL int XGBoosterPredict(BoosterHandle handle,
...
   API_END();
 }
 
 XGB_DLL int XGBoosterSaveModel(BoosterHandle handle, const char* fname) {
   API_BEGIN();
   CHECK_HANDLE();
-  std::unique_ptr<dmlc::Stream> fo(dmlc::Stream::Create(fname, "w"));
-  auto *bst = static_cast<Learner*>(handle);
-  bst->Save(fo.get());
+  common::Monitor monitor;
```
trivialfis: Note to myself: make this a model parameter instead of condition code.

src/common/common.h
```diff
@@ -64,6 +64,20 @@ inline std::vector<std::string> Split(const std::string& s, char delim) {
   return ret;
 }
 
+inline uint32_t stoui(std::string const& str, size_t* pos) {
```
trivialfis: Note to myself: Remove this now we have a general number parsing procedure.

src/common/json.cc
```diff
@@ -585,11 +685,9 @@ class GlobalCLocale {
   }
 };
 
-Json Json::Load(StringView str, bool ignore_specialization) {
+Json Json::Load(StringView str) {
   GlobalCLocale guard;
```
trivialfis: Note to myself: No longer needed now we have deterministic number parsing.

src/learner.cc
```diff
@@ -300,6 +302,65 @@ class LearnerImpl : public Learner {
     this->configured_ = true;
   }
 
+  void Save(Json* p_out) const override {
+    // FIXME(trivialfis); num_boosting_round
```
trivialfis: @hcho3 I forgot why we decided to save `num_boosted_rounds` (number of training iterations done so far) before.  Could you remind me a little?

hcho3: @trivialfis See https://xgboost-json-schema.readthedocs.io/en/latest/#gbtree

trivialfis: @hcho3 Yes I see that it's there, but I don't remember why it's there.  Currently it's not save and XGBoost doesn't seem to have any problem from recovering.

hcho3:
@trivialfis @hetong007 I'm quite nervous about requiring CMake in the R package. There are lots of different machines the R package get tested, and we have to pass every single one of them in order to keep the package on CRAN. See https://cran.r-project.org/web/checks/check_results_xgboost.html for the list of machines. **My concern is that not every one of the testing machines would have CMake installed.**

Of course, we could simply ask users to compile from source and not bother with CRAN, but then we'd lose the ability to install XGBoost with `install.packages('xgboost')`

trivialfis:
@hcho3 Thanks for the link.  Let's keep them there for now.  Will open a new issue when I have something concrete.

src/learner.cc
hcho3: @trivialfis I think it's because we store the array of tree objects, and the DMLC serializer for arrays stores the length of the array.

trivialfis: @hcho3 I will remove it in the schema then.

pommedeterresautee:
Hi, there is another point to take into account, I have made a [PR](https://github.com/r-lib/remotes/pull/408) on `remotes` package and when (soon I hope) it will be merged, XGBoost R package from `Github` will be installable in 1 line directly on R with `remotes` or `devtools` instead of current process (git clone taking care of submodules, etc.) requiring a working toolchain outside of the R environment. 

I am wondering if switching to Cmake would break things or even worst make it harder for new comers to setup the toolchain to be able to install github code.

trivialfis:
@trams Please help reviewing.  This is extracted from previous monolithic PR with specialization and schema removed.

Currently the performance is quite bad, it takes about 1.5s to load and 2.0s to save.  Running valgrind most of the time spent on memory allocation (std::vector from JsonArray).  I plan to add a custom pool allocator for all objects or use input string as buffer in  a later PR.  For now I'm going to prioritize on correctness.

Will fix the failing tests.  Unsurprisingly, handling those dynamic parameters is difficult.  But the general structure for JSON integration should be ready for review.

trivialfis:
@pommedeterresautee Thanks for the information.  Let me see if I can simplify those makefiles.

trams:
@trivialfis, thank you for mentioning me in this review.
Right now (till the end of the week) I am attending the conference and I am not sure I will have time to give a comprehensive review

As for performance I recently learned about https://github.com/lemire/simdjson a very fast json parser (~2-3Gb per second). I do not know your reasons to implement your own json parser vs reusing the third party but either way I suggest to dive into their code for tips on how they implemented high performance json parser (see here one of their benchmarks https://lemire.me/blog/2019/08/02/json-parsing-simdjson-vs-json-for-modern-c/)

trivialfis:
@trams Thanks, I'm award of SIMD json.  But requiring avx is not feasible in XGBoost.  I will try simpler method like iterative parsing, single allocator etc.  I'm not too worried about that.

We don't want to add dependencies.

trivialfis:
Emm.  I might need to add an `UpdateAllowUnknown` method to `dmlc::Parameter` so that I don't have to save the parameter cache.

trivialfis:
@RAMitchell Let me try to extract a even smaller part of this first to push the cudf PR forward.

trivialfis:
@hcho3 @RAMitchell The added `Load` method is very much like our `Configure` method, except for Json is structured while configuration is not.  I wonder if there's a way to merge two of them.

include/xgboost/gbm.h
```diff
@@ -10,6 +10,7 @@
 
 #include <dmlc/registry.h>
 #include <xgboost/base.h>
+#include <xgboost/json.h>
```
trams: Do we really need to include json.h here? It seems like a forward declaration of `class Json;` should be enough

include/xgboost/tree_model.h
```diff
@@ -216,9 +226,9 @@ class RegTree {
     };
     // pointer to parent, highest bit is used to
     // indicate whether it's a left child or not
-    int parent_;
+    int parent_{-1};
```
trams: [Optional] Maybe it is a good place to use `constexpr missing_value = -1`

python-package/xgboost/compat.py
```diff
@@ -16,14 +16,18 @@
...
     STRING_TYPES = (str,)
 
-
     def py_str(x):
         """convert c string back to python string"""
-        return x.decode('utf-8')
+        try:
+            s = x.decode('utf-8')
+        except UnicodeDecodeError as e:
+            print('Failed to decode error message, please file a bug report.')
```
trams: 1. I am not sure it is very useful. I think the message should contain a bad string unless you expect it to happen often and the string is big
2. I would rather use logging library and pass and exception there as is. It usually produces a better output and one can configure it.
3. If you insist on using print can I suggest to use `stderr` instead of `stdout`?

src/gbm/gblinear_model.cc
```diff
@@ -0,0 +1,39 @@
...
+#include <utility>
+#include <limits>
+
+#include "gblinear_model.h"
+
+namespace xgboost {
+namespace gbm {
+
+void GBLinearModel::Save(Json* p_out) const {
+  using WT = std::remove_reference<decltype(std::declval<decltype(weight)>().back())>::type;
```
trams: Optional. What do you think about using WeightType or Weight instead of WT? It seems there is only one usage so we can use longer names which are easier to read.

src/gbm/gblinear_model.cc
```diff
@@ -0,0 +1,39 @@
...
+  using JT = Number::Float;
+  static_assert(std::is_same<WT, JT>::value, "");
```
trams: Optional. What do you think about adding an error message: Weight and JT (what is that?) are supposed to be of the same type
or something like this :)

src/gbm/gblinear_model.cc
```diff
@@ -0,0 +1,39 @@
...
+#include <limits>
+
+#include "gblinear_model.h"
+
+namespace xgboost {
+namespace gbm {
+
+void GBLinearModel::Save(Json* p_out) const {
+  using WT = std::remove_reference<decltype(std::declval<decltype(weight)>().back())>::type;
+  using JT = Number::Float;
```
trams: Is JT JsonType? JsonNumberType?

src/gbm/gblinear_model.h
```diff
@@ -10,6 +10,8 @@
 #include <string>
 #include <cstring>
 
+#include "xgboost/json.h"
```
trams: I think it is possible to substitute it with forward declaration: class Json.

src/gbm/gbtree.cc
```diff
@@ -254,6 +253,22 @@ void GBTree::CommitModel(std::vector<std::vector<std::unique_ptr<RegTree>>>&& ne
...
 
+void GBTree::Load(Json const& in) {
+  fromJson(in["gbtree_train_param"], &tparam_);
+  model_.Load(in["model"]);
+}
+
+void GBTree::Save(Json* p_out) const {
+  auto& out = *p_out;
+  out["name"] = String("gbtree");
+  out["gbtree_train_param"] = toJson(tparam_);
```
trams: I do not remember whether our index operation takes temp objects on the left side (&&). Does it?

src/gbm/gbtree.cc
```diff
@@ -274,13 +289,36 @@ class Dart : public GBTree {
       fi->Read(&weight_drop_);
     }
   }
+  void Load(Json const& in) override {
+    auto const& gbtree = in["gbtree"];
+    GBTree::Load(gbtree);
+    auto j_weight_drop = get<Array>(in["weight_drop"]);
```
trams: Will this auto without & trigger a copy?

src/gbm/gbtree_model.cc
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  tree_info.resize(param.num_trees);
+}
+
+void GBTreeModel::Save(Json* p_out) const {
+  auto& out = *p_out;
+  CHECK_EQ(param.num_trees, static_cast<int>(trees.size()));
+  out["model_param"] = toJson(param);
+  std::vector<Json> trees_json;
```
trams: Theoretically we know how many trees are and we could ask std::vector to allocate enough space

src/learner.cc
```diff
@@ -300,13 +308,49 @@ class LearnerImpl : public Learner {
...
+    }
+    if (tparam_.model_format == "json") {
+      LOG(WARNING) << "JSON serialization is at experimental stage. "
+                   << "Output schema is subject to change in the future.";
+      auto  ReadStream =
+          [](dmlc::Stream* fi) {
+            std::string buffer;
+            std::string temp;
+            size_t read {0};
+            size_t size {1};
```
trams: I would start with something bigger then 1 byte. I think 4K is a reasonable default

src/learner.cc
```diff
@@ -300,13 +308,49 @@ class LearnerImpl : public Learner {
...
+    monitor_.Start(__func__);
+    if (cfg_.find("model_format") != cfg_.cend()) {
+      tparam_.model_format = cfg_.at("model_format");
+    }
+    if (tparam_.model_format == "json") {
+      LOG(WARNING) << "JSON serialization is at experimental stage. "
+                   << "Output schema is subject to change in the future.";
+      auto  ReadStream =
+          [](dmlc::Stream* fi) {
+            std::string buffer;
```
trams: I suggest not to use strings as arrays of bytes.
As far as I remember C++ standard only for std::vector guarantees that
std::vector x;
&x[0] would actually point to a consecutive array of x.size() elements

while most implementation of std::string do the same I would still advice to use std::vector<char> given that we do not use any string specific functions anyway

src/learner.cc
```diff
@@ -319,6 +363,82 @@ class LearnerImpl : public Learner {
...
     }
 
+    if (tparam_.model_format == "json") {
+      LOG(WARNING) << "JSON serialization is at experimental stage. "
+                   << "Output schema is subject to change in the future.";
+      Json model { Object() };
+      this->Save(&model);
+      std::stringstream ss;
+      Json::Dump(model, &ss);
+      size_t length = ss.str().size();
```
trams: I would actually switch 373 and 374 lines. And write size_t length = str.size()
I am not sure that it is guaranteed that ss.str() consecutive calls return the same object
Otherwise it is just a bit wasteful :)

src/learner.cc
```diff
@@ -319,6 +363,82 @@ class LearnerImpl : public Learner {
...
+      fo->Write(str.c_str(), length);
+    } else {
+      LOG(INFO) << "Saving binary model";
+      this->SaveBinary(fo);
+    }
+  }
+
+  void Save(Json* p_out) const override {
+    Json& out = *p_out;
+    Integer::Int major{XGBOOST_VER_MAJOR}, minor{XGBOOST_VER_MINOR}, patch{XGBOOST_VER_PATCH};
```
trams: I see why you need versions...

src/objective/rank_obj.cc
```diff
@@ -170,14 +173,32 @@ class LambdaRankObj : public ObjFunction {
...
 
 class PairwiseRankObj: public LambdaRankObj{
  protected:
   void GetLambdaWeight(const std::vector<ListEntry> &sorted_list,
                        std::vector<LambdaPair> *io_pairs) override {}
+
+  void Save(Json* p_out) const override {
+    auto&  out = *p_out;
+    out["name"] = String("rank:pairwise");
+    out["lambda_rank_param"] = toJson(LambdaRankObj::param_);
```
trams: I noticed that here the key name is `lambda_rank_param` but above `SoftmaxMultiClassParam` is used. Do we have a naming convention we want to follow for the produced json?
I would personally prefer `lower_case` as opposite to `CamelCase` but I would rather have the single convention then none :)

src/tree/tree_model.cc
```diff
@@ -617,6 +620,82 @@ std::string RegTree::DumpModel(const FeatureMap& fmap,
...
+    }
+    CHECK_EQ(param.num_nodes, stats_.size());
+  }
+
+  {
+    std::vector<Json> const& v_j_nodes = get<Array>(in["nodes"]);
+    nodes_.resize(v_j_nodes.size());
+
+    for (size_t i = 0; i < nodes_.size(); ++i) {
+      auto const& j_node = get<Array const>(v_j_nodes[i]);
```
trams: I would insert
CHECK_EQ(6, j_node.size());
here

src/tree/tree_model.cc
```diff
@@ -617,6 +620,82 @@ std::string RegTree::DumpModel(const FeatureMap& fmap,
...
+    std::vector<Json> j_node(6);
+    j_node[0] = static_cast<I>(n.LeftChild());
+    j_node[1] = static_cast<I>(n.RightChild());
+    j_node[2] = static_cast<I>(n.Parent());
+    j_node[3] = static_cast<I>(n.SplitIndex());
+    j_node[4] = n.SplitCond();
+    j_node[5] = n.DefaultLeft();
+    CHECK(IsA<Boolean>(j_node[5]));
+
+    v_j_nodes[i] = j_node;
```
trams: Is here an implicit constructor call to Array? If so could you make it explicit to make it easier to read?

src/tree/tree_model.cc
```diff
@@ -617,6 +620,82 @@ std::string RegTree::DumpModel(const FeatureMap& fmap,
...
+    j_stats[3] = static_cast<I>(s.leaf_child_cnt);
+
+    v_j_stats[i] = std::move(j_stats);
+  }
+  out["stats"] = std::move(v_j_stats);
+
+  std::vector<Json> v_j_nodes(nodes_.size());
+  for (size_t i = 0; i < nodes_.size(); ++i) {
+    auto const& n = nodes_[i];
+    std::vector<Json> j_node(6);
```
trams: This may be a perfect case for using C++11 initializer_list

src/tree/tree_model.cc
```diff
@@ -617,6 +620,82 @@ std::string RegTree::DumpModel(const FeatureMap& fmap,
...
+    j_stats[1] = s.sum_hess;
+    j_stats[2] = s.base_weight;
+    j_stats[3] = static_cast<I>(s.leaf_child_cnt);
+
+    v_j_stats[i] = std::move(j_stats);
+  }
+  out["stats"] = std::move(v_j_stats);
+
+  std::vector<Json> v_j_nodes(nodes_.size());
+  for (size_t i = 0; i < nodes_.size(); ++i) {
```
trams: What is a motivation of representing a tree node as an array?

I see the following pros and cons.

Pros.
1. Less space
2. Faster to parse
3. Easier to parse: no need to deal with missing attributes

Cons
1. It is harder to evolve. What if we want to add another attribute? Just add it to the end? What is we want to deprecate the element? Should we still put a dummy value?
2. The code is much harder to read. We need to remember to which index represent which attribute

So in the end we can evolve the schema even with this being an array (much harder to remove attributes, though)
And we can minimize burden on code reader by using constants for indexes which
1. would have a name
2. ensure there is no mix up

src/tree/tree_model.cc
```diff
@@ -850,10 +929,9 @@ void RegTree::CalculateContributions(const RegTree::FVec &feat,
 
   // Preallocate space for the unique path data
   const int maxd = this->MaxDepth(root_id) + 2;
-  auto *unique_path_data = new PathElement[(maxd * (maxd + 1)) / 2];
+  std::vector<PathElement> unique_path((maxd * (maxd + 1)) / 2);
```
trams: Good change! Thank you for cleaning up the code

tests/cpp/test_learner.cc
```diff
@@ -190,6 +191,77 @@ TEST(Learner, IO) {
...
 }
 
+TEST(Learner, Json_IO) {
+  size_t constexpr kRows = 10;
+  int32_t constexpr kIters = 4;
+  std::vector<bst_float> labels(kRows);
+  for (size_t i = 0; i < labels.size(); ++i) {
+    labels[i] = i;
+  }
+  auto pp_dmat = CreateDMatrix(kRows, 10, 0);
```
trams: [Optional] [Minor] Should we use a unique_ptr here?

include/xgboost/gbm.h
trams: A friendly ping. A forward definition help to reduce code interdependence and decrease a compile time.
It is not a major issue but if a forward declaration is indeed enough I would rather use it

include/xgboost/learner.h
```diff
@@ -9,13 +9,13 @@
...
 
 #include <rabit/rabit.h>
-
 #include <xgboost/base.h>
 #include <xgboost/gbm.h>
 #include <xgboost/metric.h>
 #include <xgboost/objective.h>
 #include <xgboost/feature_map.h>
 #include <xgboost/generic_parameters.h>
+#include <xgboost/json.h>
```
trams: Can we substitute this with a forward declaration
class Json;
?

It seems Json is only used as a pointer and reference in this header

python-package/xgboost/compat.py
trams: A friendly ping, what do you think about using a logging library?

src/gbm/gblinear.cc
```diff
@@ -73,9 +75,25 @@ class GBLinear : public GradientBooster {
   void Load(dmlc::Stream* fi) override {
     model_.Load(fi);
   }
+  void Load(Json const& in) override {
+    fromJson(in["gblinear_train_param"], &param_);
```
trams: Should we add an assert that in["name"] == "gblinear" ?

src/gbm/gbtree.cc
```diff
@@ -286,6 +287,22 @@ void GBTree::CommitModel(std::vector<std::vector<std::unique_ptr<RegTree>>>&& ne
   monitor_.Stop("CommitModel");
 }
 
+void GBTree::Load(Json const& in) {
+  fromJson(in["gbtree_train_param"], &tparam_);
```
trams: should we add an assert that in["name"] == "gbtree"?

trivialfis:
Related issues: #4878 , #4877, #4855, #4755,  #4758, #4639, #4618, #4072, #3980, #3771

Just leaving a note here.

trivialfis:
It seems I broke the linter...

include/xgboost/gbm.h
trivialfis: Done.

trivialfis:
Still quite early.  This is 3 RFCs in one PR ...

src/gbm/gblinear.cc
trivialfis: @trams Since this is inside GBLinear, I don't think the object can be initialized in the first place if `name != gblinear`.

trivialfis:
Note to myself: Objective is part of the model since `PredTransform` is required after load.

include/xgboost/gbm.h
trams: I do not want to nitpick. I am just a bit confused. It seems this include is still here. Do I miss some other pull request or updates?

include/xgboost/learner.h
```diff
@@ -191,5 +178,57 @@ class Learner : public rabit::Serializable {
...
+  /* \brief number of features  */
+  unsigned num_feature;
+  /* \brief number of classes, if it is multi-class classification  */
+  int num_class;
+  /*! \brief Model contain additional properties */
+  int contain_extra_attrs;
+  /*! \brief Model contain eval metrics */
+  int contain_eval_metrics;
+  /*! \brief reserved field */
+  int reserved[29];
```
trams: Is there any reason to make it this size?
I think this deserves a comment and maybe a static_assert checking that the size of a structure is not too big

include/xgboost/parameter.h
```diff
@@ -78,4 +79,28 @@ class FieldEntry<EnumClass> : public FieldEntry<int> {  \
...
 }  /* namespace dmlc */
 
-#endif  // XGBOOST_ENUM_CLASS_PARAM_H_
+namespace xgboost {
+template <typename Type>
+struct XGBoostParameter : public dmlc::Parameter<Type> {
+  bool initialised_ {false};
+
+ public:
+  bool& Initialised() { return initialised_; }
```
trams: This looks a bit unexpected
1. Do we want to permit to change initialized_ flag from outside?
2. If so shouldn't we create a separate method for that?
Also note that one cannot read this flag if the instance of XGBoostParameter is constant

src/common/io.cc
```diff
@@ -6,14 +6,96 @@
...
+    std::memcpy(dptr, dmlc::BeginPtr(buffer_) + buffer_ptr_, size);
+    buffer_ptr_ += size;
+    return size;
+  }
+}
+
+size_t PeekableInStream::PeekRead(void* dptr, size_t size) {
+  size_t nbuffer = buffer_.length() - buffer_ptr_;
+  if (nbuffer < size) {
+    buffer_ = buffer_.substr(buffer_ptr_, buffer_.length());
```
trams: Line 37-41 seems to be just a potential optimization (shrink_to_fit).
I suggest to consider
1. extracting to a separate method so it would be easier to follow the logic
2. potentially removing it cause it may lead to unexpected to the client actions

If you want you can give more control to the client by exposing the method shrink

What do you think?

src/common/io.cc
```diff
@@ -6,14 +6,96 @@
...
+    std::memcpy(dptr, dmlc::BeginPtr(buffer_), buffer_.length());
+    return buffer_.length();
+  } else {
+    std::memcpy(dptr, dmlc::BeginPtr(buffer_) + buffer_ptr_, size);
+    return size;
+  }
+}
+
+FixedSizeStream::FixedSizeStream(PeekableInStream* stream) : PeekableInStream(stream), pointer_{0} {
+  size_t size {1}, total {0};
```
trams: Should we set the initial size to some bigger number like 32 or 16? I believe a lot of allocators do not allocate less then 8 bytes anyway so we won't actually save any memory by doing .resize(1) or .resize(2) but we would incur 4-5 additional calls

src/common/io.cc
```diff
@@ -6,14 +6,96 @@
...
+  return read;
+}
+
+// size_t FixedSizeStream::Read(std::string* str) {
+//   return this->Read(&((*str)[0]), str->size());
+// }
+
+size_t FixedSizeStream::PeekRead(void* dptr, size_t size) {
+  if (size >= buffer_.size() - pointer_)  {
+    LOG(WARNING) << "Read exceeded";
```
trams: Why do we have a warning here but not line 37? Should one write logs in both cases for consistency?

Also I am not sure the line "Read exceeded" will be that useful without any additional information

src/gbm/gbtree_model.cc
```diff
@@ -0,0 +1,77 @@
...
+  trees_to_update.clear();
+  for (int i = 0; i < param.num_trees; ++i) {
+    std::unique_ptr<RegTree> ptr(new RegTree());
+    ptr->LoadModel(fi);
+    trees.push_back(std::move(ptr));
+  }
+  tree_info.resize(param.num_trees);
+  if (param.num_trees != 0) {
+    CHECK_EQ(
+        fi->Read(dmlc::BeginPtr(tree_info), sizeof(int) * param.num_trees),
```
trams: Note that if we would write on machine with sizeof(int) == 8 and read on machine with sizeof(int) == 4 this may fail

src/gbm/gbtree_model.h
```diff
@@ -37,12 +45,16 @@ struct GBTreeModelParam : public dmlc::Parameter<GBTreeModelParam> {
   int reserved[32];
   /*! \brief constructor */
   GBTreeModelParam() {
-    std::memset(this, 0, sizeof(GBTreeModelParam));
+    std::memset(this, 0, sizeof(GBTreeModelParam));  // FIXME(trivialfis): Why?
```
trams: C++ standard does not require a compiler to generate initialization of integers (does not set them to zero).
So they will have a value whatever was in memory before allocating

include/xgboost/learner.h
trivialfis: @trams No idea.  It's legacy code for binary IO.  This PR is setting up everything to remove it.

trivialfis: @trams 

> I do not want to nitpick. I am just a bit confused. It seems this include is still here. Do I miss some other pull request or updates?

Sorry about that.  It's not entering the "clean up" phase yet.  I kept finding new problems in the old binary IO and parameter logics.  Excited to get them fixed.

include/xgboost/parameter.h
trivialfis: @trams Will fix it.  This combines an new commit in `dmlc-core`.  Just got it merged today.  Will clean them up.

src/gbm/gbtree_model.h
trivialfis: @trams I'm aware of this.  Just don't like mixing c style and c++ style of doing things.  Plus it's not idea to have multiple initialization methods used in one place.  So I should remove it when starting cleaning up.

src/gbm/gbtree_model.cc
trivialfis: Will fix it.  This portion of code is legacy and copied from header to avoid "recompiling the world".  Thanks for the suggestion.

src/common/io.cc
trivialfis: @trams Nah, that's just me debugging.  Will not appear in final code.  It's an WIP.  @RAMitchell suggested to split this PR into smaller portions.  Will try doing that once I can get everything sorted our and document the whole logic clearly.

src/common/io.cc
trivialfis: @trams Let me think about it.  This fixed stream might not be necessary.

src/common/io.cc
trivialfis: @trams It's copied code from header to avoid "recompiling the world".  It troubles me that `dmlc-core` has a complete different framework for IO with self defined streams instead of aligning with standard c++ or c.  Will think about your suggestion.  Thanks for the suggestion!

src/objective/rank_obj.cc
trivialfis: @trams Will unify it.

include/xgboost/learner.h
trivialfis: @trams I moved the `ModelParameter` into this header.  Maybe I should move it into an independent header.

src/gbm/gbtree.cc
trivialfis: @trams It's a long way:

`gbtree_train_param` is not in the map, so `std::map::operator[]` initializes one for us, returns it as an reference of `Json` object all the way back to here,  then the move assignment operator is invoked ...

trivialfis: Will submit an optimization PR for JSON.  Right now I just want to get parameter IO sorted out for distributed settings.  So that I don't have to fix any more bugs for it in the next major release. ;-)

tests/cpp/test_learner.cc
trivialfis: @trams The generator uses c_api to do the job.  It's an headache but so far I don't have the time to refactor our unittests ...

include/xgboost/gbm.h
trivialfis: It's removed right now, along with other headers.

src/objective/rank_obj.cc
trivialfis: @trams  I modified to snake case in  #4945 .  I'm also tracking your comment in this PR.  #4945 and #4936 will lighten this PR a lot, but the critical part is still in this PR (saving the model itself).  Also, all your comments are either not related to those PRs or addressed.  Thanks for taking your time with this.  Let me know your thoughts.

trivialfis:
Note to myself: https://github.com/dmlc/xgboost/pull/4947

include/xgboost/learner.h
trams: I see.
It is not very obvious that json.h has something besides json handling

trivialfis:
Possibly related: https://github.com/slundberg/shap/pull/520

trivialfis:
Todos:
- [ ] Use a model parameter which contains `num_output_group`, `num_feature`, `base_score` across XGBoost.
- [ ] Add a proper data generator so that we don't copy any more code in tests.
- [ ] Extract interaction constraint into an independent evaluator.
- [ ] Merge `num_output_group` and `num_class`, remove related configurations.
- [ ] Address reviewers' comments.

src/gbm/gbtree_model.cc
trivialfis: @trams Will leave that to the optimization PR.

trivialfis:
@hcho3 @CodingCat @trams @RAMitchell I want to merge this PR first then #5046 after next release.  This way we can get a working pipeline for JSON serialization.  The PR is coming to mature now, I will track down rest of comments that are not related to performance concern.  As  #5046 gives me some confidence that we can have a fast IO routine with JSON, I removed all the hacks originally in this PR.

I'm really excited about coming to this stage, as looking at the code base, once binary IO is gone, we can remove ALL the hacks, clumsy configurations, duplicated parameters, obscure IO logics, and have a robust serialisation method in distributed environment.  Consider most of the technical debt paid in once.

Originally I refactored some binary IO logic, but now I just leave it as it is.

tests/cpp/tree/test_gpu_hist.cu
```diff
@@ -251,6 +251,9 @@ void TestHistogramIndexImpl() {
 
   // Build 2 matrices and build a histogram maker with that
   tree::GPUHistMakerSpecialised<GradientPairPrecise> hist_maker, hist_maker_ext;
+  auto pp_dmat = xgboost::CreateDMatrix(kNRows, kNCols, 0);
```
RAMitchell: I feel like `xgboost::CreateDMatrix` should just return a shared pointer. You don't have to do this in this PR.

include/xgboost/learner.h
```diff
@@ -180,16 +169,83 @@ class Learner : public Model, public rabit::Serializable {
...
+            " this parameter will be automatically detected by learner.");
+    DMLC_DECLARE_FIELD(num_class).set_default(0).set_lower_bound(0).describe(
+        "Number of class option for multi-class classifier. "
+        " By default equals 0 and corresponds to binary classifier.");
+  }
+};
+
+// base_score is changed by obj.ProbToMargin, which creates some headache when performing
+// Model IO.  This is an immutable copy with transformed `changed _score' used internally.
+struct LearnerModelParamImpl {
```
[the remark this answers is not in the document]
trivialfis: @RAMitchell Changed.  I want to remove the original one but impossible with binary IO still here.  Will clarify it in more details.

tests/cpp/tree/test_gpu_hist.cu
trivialfis: @RAMitchell Actually I did it in this PR previously then reverted.  The code change became quite scary.

include/xgboost/parameter.h
```diff
@@ -80,6 +80,23 @@ class FieldEntry<EnumClass> : public FieldEntry<int> {  \
 }  /* namespace dmlc */
 
 namespace xgboost {
+
+template <typename P, typename Base, typename Container>
+Args UpdateAllowUnknownImpl(P* parameter, Base* base,
```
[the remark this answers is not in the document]
trivialfis: @RAMitchell Split up the base class from derived class.

trivialfis:
Update:

- Define prediction cache in gbtree instead of predictor, as the cache needs to be in sync between `cpu_predictor` and `gpu_predictor`.
- Add basic documentation for the serialisation format.

trivialfis:
@hcho3 I added a vomit CMake option paired with an `Observer` class in c++.  Can we consider https://github.com/dmlc/xgboost/issues/4837 being resolved?  As you can extend the `Observer` class very easily and apply it to most of the objects in XGBoost.

hcho3:
@trivialfis Awesome! Let me look at it today.

trivialfis:
@RAMitchell  @hcho3  I enabled part of the JSON PR in Python pickle.  It might be a little risky to enable JSON in something so important.  But at least this way the pickling is actually working.

include/xgboost/generic_parameters.h
```diff
@@ -12,6 +12,9 @@
 
 namespace xgboost {
 struct GenericParameter : public XGBoostParameter<GenericParameter> {
+  static int32_t constexpr kCpuId = -1;
```
RAMitchell: What is this used for?

src/learner.cc
```diff
@@ -134,36 +172,127 @@ struct LearnerTrainParam : public XGBoostParameter<LearnerTrainParam> {
...
+ * - has_cache ? 1 : 0 (means whether configuration cache `cfg_` has any new entry
+ *                      that hasn't been passed through XGBoost)
+ *
+ * Combining above two into {0, 0} {0, 1} {1, 0} {1, 1}, below is the state transition.
+ *
+ * The difficult part is XGBoost allows changing parameters per-iteration, which means
+ * weird usage patterns like setting a different global bias for each iteration is valid
+ * and must be properly handled.
+ */
+class ModelStatus {
```
RAMitchell: It seems a bit extreme to have to design a finite state machine in order to track the status of initialisation.

CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ option(USE_OPENMP "Build with OpenMP support." ON)
 option(JVM_BINDINGS "Build JVM bindings" OFF)
 option(R_LIB "Build shared library for R package" OFF)
 ## Dev
+option(USE_DEBUG_VOMIT "Dump internal training results like gradients and predictions to stdout.
```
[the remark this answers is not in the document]
trivialfis: @RAMitchell i wanted to limit its usage as much as possible, so no one can accidentality turn it on.  There's a limit of buffer size on most of terminals, it will flush your screen.   I can change to the wording, it's picked up from some other projects. ;-)

include/xgboost/generic_parameters.h
trivialfis: @RAMitchell Configuring GPU ID, originally in https://github.com/dmlc/xgboost/pull/5072 which I decided not to merge.  Besides it's better to replace repeated usage of `-1` with a named symbol.

src/learner.cc
trivialfis: @RAMitchell Removed.  I implemented it for handling edge cases like save immediately after load etc.  Now I just enforce there's a configure step before save/dump.

src/learner.cc
trivialfis: @trams See https://stackoverflow.com/questions/1986966/does-s0-point-to-contiguous-characters-in-a-stdstring .  Using string eliminates a copy and plays nice with `dmlc::Stream`.  Should be safe here, not nice looking though.

trivialfis:
@RAMitchell I removed the state machine.  Changing cache location is required for training continuation.   The debug output is only convenient for debugging training continuation.  Without it it's really difficult to notice what's changed after load.

trivialfis:
@RAMitchell  

Update: Disabled prediction cache explicitly for multi-class case.  Which breaks training continuation for some updaters.  See those serialisation tests, it should be really thorough for JSON.  This turned out to be another bug cleaning PR.

src/tree/param.cc
```diff
@@ -0,0 +1,79 @@
+/*!
+ * Copyright by Contributors 2019
+ */
+#include <iostream>
+#include <vector>
+
+#include "param.h"
+
+namespace std {
+std::istream &operator>>(std::istream &is, std::vector<int> &t) {
```
trivialfis: This is not only moving to cc file.  But also fixed parsing empty vector.

trivialfis:
@hcho3 Please help reviewing.  4 days later it will be one year anniversary of https://github.com/dmlc/xgboost/issues/3980 . ;-)

hcho3:
@trivialfis Thanks a lot for your valiant effort. I will start reviewing this week

src/tree/tree_model.cc
trivialfis: Done.

trivialfis:
Re-enabled prediction cache for multi-class.

trivialfis:
@RAMitchell To summarize, other than bug fixes, this PR does 4 things:

- Add JSON IO for XGBoost.
- Replace duplicated parameters, deprecate the old one when deleting is not feasible.
- Handle prediction cache syncing.
- Exhausted tests for training continuation, including boosted trees, boosted forest, multi-class, logit.

So why is that so huge?  Because the JSON IO creates a framework for easy testing.  It's meant to help creating reproducible train, where binary IO is not feasible.  Also with JSON, you can access and verify all internal parameters of XGBoost.  Hence easy testing for second item.

trivialfis:
@RAMitchell The R tests actually passes on Jenkins ... I'm gonna ignore the error I found.

``` R
> require(data.table)
Loading required package: data.table
data.table 1.12.2 using 4 threads (see ?getDTthreads).  Latest news: r-datatable.com
> a <- data.table(c=1:3, d=2:4)
> str(a)
Classes ‘data.table’ and 'data.frame':	3 obs. of  2 variables:
 $ c: int  1 2 3
 $ d: int  2 3 4
 - attr(*, ".internal.selfref")=<externalptr> 
```

Notice the single quote around `data.table` on my machine is Unicode, weirdly it doesn't seem to be the case for `data.frame`.  The Unicode breaks `testthat::expect_output`'s regex searching.

hcho3:
@trivialfis Can we close this, or is there more work to be done in JSON serialization?

trivialfis:
Will close it once R package is updated.

trivialfis:
Sorry for being slow.

PR #4733, opened 2019-08-04
Remove makefiles.

trivialfis:
Just trying to see what will happen on test farm since R-package use amalgamation other than in tree makefiles.


trivialfis:
Seems tricky.  Sorry for the noise.

PR #4737, opened 2019-08-05
Move bitfield into common.

trivialfis:
To be used by cudf support.

The validity mask from Apache arrow memory layout is passed in as `unsigned char`, but the specification also said that it must be of 8 byte aligned and the length should be padded to be a multiple of 8 byte.  So `uint64_t` used in BitField fits.


trivialfis:
@wbo4958 Feel free to provide suggestion as we go. ;-)

trivialfis:
@RAMitchell Merging as this is only moving code section, should be safe.  Feel free to reach me for any doubt.

PR #4738, opened 2019-08-06
fix objective parameter issue

wbo4958:
Issue repro step:
1. load a model (naming A) from disk,
2. invoke A.setParam() to set a parameter
3. Learner.configured_ will be set to false
4. A.Predict()
5. learner.Configure()
    then tparam_.objective will set to default (reg:squarederror)
6. learner.ConfigureObjective() will use incorrect objective

For example
in train phase, objective is multi:softprob
but after before 6 steps, in predict phase, objective will be multi:softmax

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -289,6 +289,7 @@ class LearnerImpl : public Learner {
     }
     cfg_["num_class"] = common::ToString(mparam_.num_class);
     cfg_["num_feature"] = common::ToString(mparam_.num_feature);
+    cfg_["objective"] = common::ToString(tparam_.objective);
 
     gbm_->Configure({cfg_.cbegin(), cfg_.cend()});
     obj_->Configure({cfg_.begin(), cfg_.end()});
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -287,9 +287,12 @@ class LearnerImpl : public Learner {
         metrics_.emplace_back(Metric::Create(name, &generic_param_));
       }
     }
-    cfg_["num_class"] = common::ToString(mparam_.num_class);
-    cfg_["num_feature"] = common::ToString(mparam_.num_feature);
-    cfg_["objective"] = common::ToString(tparam_.objective);
+
+    auto m = mparam_.__DICT__();
+    cfg_.insert(m.cbegin(), m.cend());
+
+    auto n = tparam_.__DICT__();
+    cfg_.insert(n.cbegin(), n.cend());
 
     gbm_->Configure({cfg_.cbegin(), cfg_.cend()});
     obj_->Configure({cfg_.begin(), cfg_.end()});
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -288,8 +288,8 @@ class LearnerImpl : public Learner {
       }
     }
 
-    auto m = mparam_.__DICT__();
-    cfg_.insert(m.cbegin(), m.cend());
+    cfg_["num_class"] = common::ToString(mparam_.num_class);
+    cfg_["num_feature"] = common::ToString(mparam_.num_feature);
 
     auto n = tparam_.__DICT__();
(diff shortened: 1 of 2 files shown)


wbo4958:
@trivialfis @RAMitchell Could you help to review the change?

trivialfis:
Will try to understand what's happening there.

trivialfis:
@wbo4958 Let me work on `dmlc::Parameter` a little.

trivialfis:
The problem is after loading model from disk, parameters cache is empty so every call to `set_param()` will invalidate loaded model.  @wbo4958 Do you think we can fill the configuration cache `cfg_` with everything loaded including `mparam_` and `tparam_`?  You can do so easily by calling:
```
  cfg_.emplace(mparam_.__DICT__().cbegin(), mparam_.__DICT__().cend());
```

wbo4958:
> The problem is after loading model from disk, parameters cache is empty so every call to `set_param()` will invalidate loaded model. @wbo4958 Do you think we can fill the configuration cache `cfg_` with everything loaded including `mparam_` and `tparam_`? You can do so easily by calling:
> 
> ```
>   cfg_.emplace(mparam_.__DICT__().cbegin(), mparam_.__DICT__().cend());
> ```

Yes, That's the problem, Do we need all params? previous ,we only need num_class/num_feature

hcho3:
@wbo4958 Did the issue exist in 0.90 or older versions?

hcho3:
@trivialfis In #4577, we removed the `Booster` class from c_api.cc and that class used to cache all configuration parameters in `cfg_`.

trivialfis:
@hcho3 But those in old `Booster` aren't saved in binary model?  Honestly I don't really have a good idea about all these use cases.  Hopefully JSON can make our lives easier.

wbo4958:
> @wbo4958 Did the issue exist in 0.90 or older versions? ~My guess is that older versions would have had the same problem.~ See my next comment.

0.90 is good.

hcho3:
@trivialfis Don't merge this just yet. I'd like to add a test case to cover @wbo4958's use case.

trivialfis:
@hcho3 @wbo4958 The reason I want to put everything into `cfg_` is because I can imagine `booster` having same issue, along with other parameters.

wbo4958:
@hcho3 @trivialfis  pls help to review the newest code.

trivialfis:
@wbo4958 Tests are failing.

wbo4958:
@trivialfis Seems we shouldn't involve base_score from mparam_ into cfg_

in LogisticRegression, 

```
  static bst_float ProbToMargin(bst_float base_score) {
    CHECK(base_score > 0.0f && base_score < 1.0f)
      << "base_score must be in (0,1) for logistic loss";
    return -logf(1.0f / base_score - 1.0f);
  }
```

`base_score` will be re-calculated in ConfigureObjective
`mparam_.base_score = obj_->ProbToMargin(mparam_.base_score);`

So if base_score is 0.5, after ProbToMargin, base_score will be set to 0, when save model, mParam_.base_score will be saved.

after Load model. if we involve base_score into cfg_, then when re-configure, mParam_.base_score will be 0. and it will fail in ProbToMargin check.

wbo4958:
@hcho3 I locally can build successfully, why it failed in this Jenkins linux build?

trivialfis:
@wbo4958 We run sanitizer, you have memory leak in your test, specifically the dmatrix shared pointer.

wbo4958:
> @wbo4958 We run sanitizer, you have memory leak in your test, specifically the dmatrix shared pointer.

Oops, my fault.

Fixed.

wbo4958:
@trivialfis All checks passed, Could you help to merge this?

trivialfis:
@wbo4958 Thanks, merging.

PR #4739, opened 2019-08-06
Add Json integer, remove specialization.

trivialfis:
Extracted from #4732, which in turns is extracted from #4683 to reduce PR size for quicker review.

### DIFF
--- a/include/xgboost/json.h
+++ b/include/xgboost/json.h
@@ -4,8 +4,9 @@
 #ifndef XGBOOST_JSON_H_
 #define XGBOOST_JSON_H_
 
-#include <xgboost/logging.h>
+#include <dmlc/io.h>
 
+#include <xgboost/logging.h>
 #include <string>
 
 #include <map>
@@ -29,7 +30,6 @@ class Value {
     Integer,
     Object,  // std::map
     Array,   // std::vector
-    Raw,
     Boolean,
     Null
   };
@@ -63,9 +63,9 @@ T* Cast(U* value) {
   if (IsA<T>(value)) {
     return dynamic_cast<T*>(value);
   } else {
-    throw std::runtime_error(
-        "Invalid cast, from " + value->TypeStr() + " to " + T().TypeStr());
+    LOG(FATAL) << "Invalid cast, from " + value->TypeStr() + " to " + T().TypeStr();
   }
+  return dynamic_cast<T*>(value);  // supress compiler warning.
 }
 
 class JsonString : public Value {
@@ -123,32 +123,6 @@ class JsonArray : public Value {
   }
 };
 
-class JsonRaw : public Value {
-  std::string str_;
-
- public:
-  explicit JsonRaw(std::string&& str) :
-      Value(ValueKind::Raw),
-      str_{std::move(str)}{}  // NOLINT
-  JsonRaw() : Value(ValueKind::Raw) {}
-
-  std::string const& getRaw() &&      { return str_; }
-  std::string const& getRaw() const & { return str_; }
-  std::string&       getRaw()       & { return str_; }
-
-  void Save(JsonWriter* writer) override;
-
-  Json& operator[](std::string const & key) override;
-  Json& operator[](int ind) override;
-
-  bool operator==(Value const& rhs) const override;
-  Value& operator=(Value const& rhs) override;
-
-  static bool isClassOf(Value const* value) {
-    return value->Type() == ValueKind::Raw;
-  }
-};
-
 class JsonObject : public Value {
   std::map<std::string, Json> object_;
 
@@ -185,7 +159,9 @@ class JsonNumber : public Value {
 
  public:
   JsonNumber() : Value(ValueKind::Number) {}
-  JsonNumber(double value) : Value(ValueKind::Number) {  // NOLINT
+  template <typename FloatT,
+            typename std::enable_if<std::is_same<FloatT, Float>::value>::type* = nullptr>
+  JsonNumber(FloatT value) : Value(ValueKind::Number) {  // NOLINT
     number_ = value;
   }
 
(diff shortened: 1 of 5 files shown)


trivialfis:
Integer is used to read pointer value

src/common/json.cc
```diff
@@ -527,15 +525,118 @@ Json JsonReader::ParseObject() {
 }
 
 Json JsonReader::ParseNumber() {
-  std::string substr = raw_str_.substr(cursor_.Pos(), kMaxNumLength);
-  size_t pos = 0;
+  // Adopted from sajson with some simplifications and small optimizations.
```
hcho3: Have you considered using `dmlc::strtof()` to parse floating-point numbers?
https://github.com/dmlc/dmlc-core/blob/808f485387f9a03f78fa9f1159f387d0d91b7a28/include/dmlc/strtonum.h#L99

Some differences:
* `dmlc::strtof()` contains more checks for overflow and underflow conditions
* `dmlc::strtof()` supports NAN and INF
* `dmlc::strtof()` avoids the use of `exp10` and `exp` by repeating `*10`: 
https://github.com/dmlc/dmlc-core/blob/808f485387f9a03f78fa9f1159f387d0d91b7a28/include/dmlc/strtonum.h#L184-L199

I think your implementation is reasonable. I'll defer to you to choose between your implementation or `dmlc::strtof()`.

src/common/json.cc
```diff
@@ -527,15 +525,118 @@ Json JsonReader::ParseObject() {
...
+  if (exponent) {
+    CHECK(is_float);
+    // If d is zero but the exponent is huge, don't
+    // multiply zero by inf which gives nan.
+    if (f != 0.0) {
+      // Only use exp10 from libc on gcc+linux
+#if !defined(__GNUC__) || defined(_WIN32) || defined(__APPLE__)
+#define exp10(val) std::pow(10, (val))
+#endif  // !defined(__GNUC__) || defined(_WIN32) || defined(__APPLE__)
+      f *= exp10(exponent);
```
hcho3: Is `exp10` faster?

src/common/json.cc
```diff
@@ -585,39 +686,23 @@ class GlobalCLocale {
   }
 };
 
-Json Json::Load(StringView str, bool ignore_specialization) {
+Json Json::Load(StringView str) {
   GlobalCLocale guard;
-  LOG(WARNING) << "Json serialization is still experimental."
```
hcho3: Why are we removing this warning?

tests/cpp/common/test_json.cc
```diff
@@ -143,9 +143,26 @@ TEST(Json, TestParseObject) {
 }
 
 TEST(Json, ParseNumber) {
-  std::string str = "31.8892";
-  auto json = Json::Load(StringView{str.c_str(), str.size()});
-  ASSERT_NEAR(get<JsonNumber>(json), 31.8892f, kRtEps);
+  {
```
hcho3: Take a look at https://github.com/dmlc/dmlc-core/blob/808f485387f9a03f78fa9f1159f387d0d91b7a28/test/unittest/unittest_param.cc#L18 and see if we should adopt some of the test cases.

tests/cpp/common/test_json.cc
```diff
@@ -333,18 +351,19 @@ TEST(Json, LoadDump) {
   Json::Dump(origin, &fout);
   fout.close();
 
-  buffer = common::LoadSequentialFile(path);
```
hcho3: I notice that `common::LoadSequentialFile()` has a special logic for POSIX target. Is there significant speed-up when you use `posix_fadvise()` followed by POSIX `read()`?

include/xgboost/json.h
```diff
@@ -256,15 +262,16 @@ class JsonBoolean : public Value {
 };
 
 struct StringView {
```
hcho3: Aside: we may want to contribute this class into dmlc-core (`dmlc::StringView`)

src/common/json.cc
trivialfis: @hcho3 Will put it into `Learner::Load` and `Save`.  This implementation will also be used to consume `__array_interface__` from numpy and `__cuda_array_interface__` from cudf and numba and friends.

include/xgboost/json.h
trivialfis: @hcho3 There are quite a lot of requirements to make it really useful.  One from them is `{from, to}_chars` from c++17.  I tried to adopt it from `libcxx` from llvm project, but a few thousands lines of code is unavoidable.

tests/cpp/common/test_json.cc
trivialfis: @hcho3 I will remove it in next PR as it's not really useful for loading model from memory buffer, and conflicts with rabit check point logic.

But to answer your question, yes, it's a huge speed up.

src/common/json.cc
trivialfis: @hcho3 The problem is we also support integer parsing.  It's not in json specification but added by most of implementations.  Also we have lots of integers in tree node index.  So I need a function that can determine whether the string is an integer or a float.

hcho3: @trivialfis Got it. In that case, let's just keep `JsonReader::ParseNumber()` as it is.

In the future, when we decide to move your JSON serializer into dmlc-core, we can discuss how to combine the logic in `JsonReader::ParseNumber()` and `dmlc::ParseFloat()` for code de-duplication.

src/common/json.cc
trivialfis: @hcho3 Sorry I never benchmark this particular function knowing that the newly revised `ParseNumber` is no longer a bottle neck.

trivialfis:
>  I previously read your JSON implementation at #3930.

Ah, that was almost a thousand tickets ago. ;-)

tests/cpp/common/test_json.cc
trivialfis: @hcho3 Looking into those tests, JSON number is much simpler.  Like not having `f` suffix, no `inf` or `NAN`.

number

hcho3: @trivialfis Interesting, so NAN and INF are not allowed in JSON. That's awesome

trivialfis:
Merging as no additional logic change to original code base.

trivialfis:
@hcho3 Thanks for the review.

PR #4743, opened 2019-08-07
remove some unused functions as reported by cppcheck

rongou:
After removing `gpu_hist`, I ran cppcheck to see if it can find more dead code, but it found a few random functions.

@RAMitchell @trivialfis @hcho3

### DIFF
--- a/src/common/bitmap.h
+++ b/src/common/bitmap.h
@@ -47,7 +47,7 @@ struct BitMap {
     for (bst_omp_uint i = 0; i < nsize; ++i) {
       uint32_t res = 0;
       for (int k = 0; k < 32; ++k) {
-        int bit = vec[(i << 5) | k];
+        uint32_t bit = vec[(i << 5) | k];
         res |= (bit << k);
       }
       data[i] = res;
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -31,13 +31,13 @@ const char* kMaxDeltaStepDefaultValue = "0.7";
 
 inline bool IsFloat(const std::string& str) {
   std::stringstream ss(str);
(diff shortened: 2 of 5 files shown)


src/common/bitmap.h
```diff
@@ -47,7 +47,7 @@ struct BitMap {
     for (bst_omp_uint i = 0; i < nsize; ++i) {
       uint32_t res = 0;
       for (int k = 0; k < 32; ++k) {
-        int bit = vec[(i << 5) | k];
+        uint32_t bit = vec[(i << 5) | k];
```
rongou: cppcheck was complaining about `shifting signed 32-bit value by 31 bits is undefined behaviour`.

src/learner.cc
```diff
@@ -31,13 +31,13 @@ const char* kMaxDeltaStepDefaultValue = "0.7";
 
 inline bool IsFloat(const std::string& str) {
   std::stringstream ss(str);
-  float f;
```
rongou: Uninitialized variable

src/learner.cc
```diff
@@ -31,13 +31,13 @@ const char* kMaxDeltaStepDefaultValue = "0.7";
...
+  float f{};
   return !((ss >> std::noskipws >> f).rdstate() ^ std::ios_base::eofbit);
 }
 
 inline bool IsInt(const std::string& str) {
   std::stringstream ss(str);
-  int i;
```
rongou: Uninitialized variable

src/common/bitmap.h
trivialfis: Good to know. Thanks.

PR #4744, opened 2019-08-07
remove the qids_ field in MetaInfo

rongou:
I'm not sure how much we care about saved `MetaInfo` binaries. If you want I can add the versioning back.

Part of the DMatrix refactoring #4354 

@trivialfis @RAMitchell


trivialfis:
@rongou I'm afraid we have to maintain its stability.  Just give a warning for loading old binary.

rongou:
@trivialfis added the version. Please take another look.

PR #4747, opened 2019-08-07
- fix external memory documentation [skip ci]

sriramch:
- to state that it is supported now on gpu algorithms

@RAMitchell @trivialfis @rongou - minor fix for external memory documentation.


trivialfis:
@sriramch I think it's still very much experimental.

sriramch:
> @sriramch I think it's still very much experimental.

@trivialfis - indeed. reinstated that note...

Issue #4680, opened 2019-07-18
[Roadmap] XGBoost 1.0.0 Roadmap


CodingCat:
@dmlc/xgboost-committer please add your items here by editing this post. Let's ensure that 

* each item has to be associated with a ticket 

* major design/refactoring are associated with a RFC before committing the code 

* blocking issue must be marked as blocking

* breaking change must be marked as breaking

for other contributors who have no permission to edit the post, please comment here about what you think should be in 1.0.0

I have created three new types labels, 1.0.0, Blocking, Breaking

- [x] Improve installation experience on Mac OSX (#4477)
- [x] Remove old GPU objectives.
- [x]  Remove gpu_exact updater (deprecated) #4527
- [x]  Remove multi threaded multi gpu support (deprecated) #4531
- [x]  External memory for gpu and associated dmatrix refactoring #4357 #4354
- [ ] Spark Checkpoint Performance Improvement (https://github.com/dmlc/xgboost/issues/3946) 
- [x] [BLOCKING] the sync mechanism in hist method in master branch is broken due to the inconsistent shape of tree in different workers (https://github.com/dmlc/xgboost/pull/4716, https://github.com/dmlc/xgboost/issues/4679)
- [x] Per-node sync slows down distributed training with 'hist' (#4679)
- [x] Regression tests including binary IO compatibility, output stability, performance regressions.

thesuperzapper:
Not a committer, but can we please target PySpark API for 1.0? 
Issue: #3370 
Current PR: #4656

CodingCat:
for other contributors who have no permission to edit the post, please comment here about what you think should be in 1.0.0

thesuperzapper:
Also, should we target moving exclusively to the Scala based Rabit tracker (for Spark) in 1.0?

trams:
I am also not a committer but me and the company I work in is very interested in fixing the performance issue with checkpointing (or at least mitigate it) #3946

thesuperzapper:
There is also the secondary consideration, that we might not be ready for 1.0, and the API guarantees that come with that, for example, we could instead do 0.10.0 next?

RAMitchell:
Added some gpu related items.

chenqin:
would like to get native xgb fix included.
https://github.com/dmlc/xgboost/issues/4753

trivialfis:
JSON is removed from the list.  See https://github.com/dmlc/xgboost/pull/4683#issuecomment-520485615

[...]

PR #4749, opened 2019-08-07
[BREAKING] prevent multi-gpu usage

rongou:
See RFC #4531
Part of 1.0.0. roadmap #4680 

### DIFF
--- a/include/xgboost/generic_parameters.h
+++ b/include/xgboost/generic_parameters.h
@@ -41,9 +41,9 @@ struct GenericParameter : public dmlc::Parameter<GenericParameter> {
     DMLC_DECLARE_FIELD(n_gpus)
         .set_default(0)
         .set_range(0, 1)
-        .describe("Deprecated, please use distributed training with one "
-                  "process per GPU. "
-                  "Number of GPUs to use for multi-gpu algorithms.");
+        .describe("Deprecated. Single process multi-GPU training is no longer supported. "
+                  "Please switch to distributed training with one process per GPU. "
+                  "This can be done using Dask or Spark.");
   }
 };
 }  // namespace xgboost
--- a/include/xgboost/generic_parameters.h
+++ b/include/xgboost/generic_parameters.h
@@ -40,7 +40,7 @@ struct GenericParameter : public dmlc::Parameter<GenericParameter> {
         .describe("The primary GPU device ordinal.");
     DMLC_DECLARE_FIELD(n_gpus)
         .set_default(0)
-        .set_lower_bound(0)
+        .set_range(0, 1)
         .describe("Deprecated, please use distributed training with one "
                   "process per GPU. "
                   "Number of GPUs to use for multi-gpu algorithms.");
--- a/include/xgboost/generic_parameters.h
+++ b/include/xgboost/generic_parameters.h
@@ -40,7 +40,7 @@ struct GenericParameter : public dmlc::Parameter<GenericParameter> {
         .describe("The primary GPU device ordinal.");
     DMLC_DECLARE_FIELD(n_gpus)
         .set_default(0)
-        .set_lower_bound(-1)
+        .set_lower_bound(0)
         .describe("Deprecated, please use distributed training with one "
                   "process per GPU. "
                   "Number of GPUs to use for multi-gpu algorithms.");
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -580,8 +580,14 @@ class LearnerImpl : public Learner {
     }
     gbm_->Configure(args);
 
-    if (this->gbm_->UseGPU() && cfg_.find("n_gpus") == cfg_.cend()) {
-      generic_param_.n_gpus = 1;
+    if (this->gbm_->UseGPU()) {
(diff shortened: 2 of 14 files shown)


tests/cpp/predictor/test_gpu_predictor.cu
```diff
@@ -68,6 +68,41 @@ TEST(gpu_predictor, Test) {
   delete dmat;
 }
 
+TEST(gpu_predictor, MoreTest) {
```
rongou: This was the `MGPU_Test`, similar to `Test`, but tests more dmatrix sizes.

tests/cpp/predictor/test_gpu_predictor.cu
```diff
@@ -89,10 +124,43 @@ TEST(gpu_predictor, ExternalMemoryTest) {
   }
 }
 
-#if defined(XGBOOST_USE_NCCL)
+TEST(gpu_predictor, MoreExternalMemoryTest) {
```
rongou: similar

tests/python-gpu/test_large_sizes.py
```diff
@@ -25,7 +25,7 @@ def eprint(*args, **kwargs):
 # reduced to fit onto 1 gpu but still be large
 rows3 = 5000  # small
 rows2 = 4360032  # medium
-rows1 = 42360032  # large
+rows1 = 32360032  # large
```
rongou: The original size crashes on my Titan V, so lowering it a bit.

tests/cpp/predictor/test_gpu_predictor.cu
```diff
@@ -68,6 +66,41 @@ TEST(gpu_predictor, Test) {
   delete dmat;
 }
 
+TEST(gpu_predictor, MoreTest) {
```
sriramch: can this be combined with test which can then be invoked with varying n_cols and n_rows

include/xgboost/generic_parameters.h
```diff
@@ -40,7 +40,7 @@ struct GenericParameter : public dmlc::Parameter<GenericParameter> {
         .describe("The primary GPU device ordinal.");
     DMLC_DECLARE_FIELD(n_gpus)
         .set_default(0)
-        .set_lower_bound(-1)
+        .set_lower_bound(0)
```
sriramch: i'm hoping no one uses the moving target (the head version) in their testing, as this is a change in behavior in their configuration.

rongou: Yes this is a breaking change if you are using multi-gpu.

tests/cpp/predictor/test_gpu_predictor.cu
rongou: Done.

trivialfis:
Please provide some informative messages instead of setting range.  Like it's removed in 1.0, what are the alternative options or a reference to related doc page.  I think the parameter is quite popular despite its limitations.  :-)

rongou:
The parameter description refers to distributed training with one process per GPU. Do we have a canonical document/tutorial for that? Maybe using dask?

trivialfis:
Demo yes.  But not official document as the interface is not mature.

trivialfis:
But sure, we should mention dask at least.  I will work on it after sorting out current PRs, assuming no one else beats me to it.

rongou:
Happy to point it when you have it ready. Do you want me to change the wording on the current PR?

rongou:
Finally got the CI to pass. Please take another look. @RAMitchell @trivialfis

src/learner.cc
```diff
@@ -580,8 +580,14 @@ class LearnerImpl : public Learner {
...
     gbm_->Configure(args);
 
-    if (this->gbm_->UseGPU() && cfg_.find("n_gpus") == cfg_.cend()) {
-      generic_param_.n_gpus = 1;
+    if (this->gbm_->UseGPU()) {
+      if (cfg_.find("n_gpus") == cfg_.cend()) {
+        generic_param_.n_gpus = 1;
+      }
+      if (generic_param_.n_gpus != 1) {
+        LOG(FATAL) << "Multi-GPU training is no longer supported. "
```
RAMitchell: I think this error message could mislead a user to think multi-GPU training is completely removed, perhaps it is better to say multi-GPU training using threads is no longer supported, and that this can now be achieved instead using distributed training.

trivialfis:
Please mention distributed training with dask.  :)

src/learner.cc
rongou: Done.

rongou:
Reworded the messages, PTAL.

src/learner.cc
```diff
@@ -585,8 +585,9 @@ class LearnerImpl : public Learner {
         generic_param_.n_gpus = 1;
       }
       if (generic_param_.n_gpus != 1) {
```
trivialfis: Not needed for this PR.  But after removing `n_gpus`, the `UseGPU` function should follow ;-).

trivialfis:
@RAMitchell What do you think?  Ready to go?

PR #4754, opened 2019-08-08
[jvm-packages] cleaning checkpoint file after a successful training

CodingCat:


CodingCat:
@trams would you please help to review?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManager.scala
```diff
@@ -127,7 +133,7 @@ private[spark] class CheckpointManager(sc: SparkContext, checkpointPath: String)
 
 object CheckpointManager {
 
-  private[spark] def extractParams(params: Map[String, Any]): (String, Int) = {
+  private[spark] def extractParams(params: Map[String, Any]): (String, Int, Boolean) = {
```
trams: I think it may be a time to create a case class

case class Params(checkpointPath: String, checkpointInverval: Int, skip_clean_checkpoint: Boolean)

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManager.scala
```diff
@@ -141,6 +147,13 @@ object CheckpointManager {
       case _ => throw new IllegalArgumentException("parameter \"checkpoint_interval\" must be" +
         " an instance of Int.")
     }
-    (checkpointPath, checkpointInterval)
+
+    val skipCheckpointFile: Boolean = params.get("skip_clean_checkpoint") match {
+      case None => false
```
trams: I am a bit worried that it changes "API".
Before this change xgboost-spark does not clean it checkpoint folder
After this change it will do the cleaning by default.

What do you think about creating a param
"clean_checkpoint" instead of "skip_clean_checkpoint"
Those who wants can enable cleaning.

One use case when cleaning checkpoint folder may not be a good idea is if we desire to train N trees, optionally validate it and then continue training (may be with different hyperparameters) another M trees.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManager.scala
```diff
@@ -53,6 +53,12 @@ private[spark] class CheckpointManager(sc: SparkContext, checkpointPath: String)
     }
   }
 
+  def cleanPath(): Unit = {
+    if (checkpointPath != "") {
+      FileSystem.get(sc.hadoopConfiguration).delete(new Path(checkpointPath), true)
```
trams: This assumes that CheckpontManager owns the folder (i.e. all files in this folder has been created by this or earlier CheckpointManager) so it is safe to remove the whole folder.

This is true for our use case. I am not sure it is actually true for everybody. At least we should update the docs (and 1.0 changelog) to mention this

One way to solve this problem would be to actually reuse `cleanUpHigherVersions`. We can pass it here to clean all versions. After that we can remove the folder non recursively. That would remove the empty directory if any

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -473,6 +475,11 @@ object XGBoost extends Serializable {
             tracker.stop()
           }
       }.last
+      // we should delete the checkpoint directory after a successful training
+      if (!skipCleanCheckpoint) {
```
trams: [Really minor] I am not sure whether xgboost has java|scala coding style. I generally prefer to have `cleanCheckpoint` flag not `skipCleanCheckpoint`. That would avoid adding extra "not" which makes reading the code slightly harder
P.S. This is really bike shadding

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManagerSuite.scala
```diff
@@ -67,4 +68,51 @@ class CheckpointManagerSuite extends FunSuite with TmpFolderPerSuite with PerTes
...
+      assert(files.length == 1)
+      assert(files.head.getPath.getName == "8.model")
+      val tmpModel = SXGBoost.loadModel(s"$tmpPath/8.model")
+      // Train next model based on prev model
+      val nextModel = new XGBoostClassifier(paramMap ++ Seq("num_round" -> 8)).fit(training)
+      assert(error(tmpModel) > error(prevModel._booster))
+      assert(error(prevModel._booster) > error(nextModel._booster))
+      assert(error(nextModel._booster) < 0.1)
+    }
+    if (!skipCleanCheckpoint) {
```
trams: Should we just put "else" here?

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManagerSuite.scala
```diff
@@ -67,4 +68,51 @@ class CheckpointManagerSuite extends FunSuite with TmpFolderPerSuite with PerTes
...
+    val paramMap = Map("eta" -> "1", "max_depth" -> 2,
+      "objective" -> "binary:logistic", "checkpoint_path" -> tmpPath,
+      "checkpoint_interval" -> 2, "num_workers" -> numWorkers) ++ cacheDataMap ++
+      skipCleanCheckpointMap
+
+    val prevModel = new XGBoostClassifier(paramMap ++ Seq("num_round" -> 5)).fit(training)
+    def error(model: Booster): Float = eval.eval(
+      model.predict(testDM, outPutMargin = true), testDM)
+
+    if(skipCleanCheckpoint) {
```
trams: [Optional] [Minor] Missing space between if and brackets

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManagerSuite.scala
```diff
@@ -67,4 +68,51 @@ class CheckpointManagerSuite extends FunSuite with TmpFolderPerSuite with PerTes
...
+      assert(!FileSystem.get(sc.hadoopConfiguration).exists(new Path(tmpPath)))
+    }
+  }
+
+  test("training with checkpoint boosters") {
+    trainingWithCheckpoint(cacheData = false, skipCleanCheckpoint = true)
+  }
+
+  test("training with checkpoint boosters with cached training dataset") {
+    trainingWithCheckpoint(cacheData = false, skipCleanCheckpoint = true)
```
trams: I think this line should be `cacheData = true`.
Otherwise I do not see the difference between 112 and 108

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManager.scala
CodingCat: actually the previous implementation is buggy, even you wanted N trees, the leftover of checkpoint is the one produced after N-1 iterations, 

the reason I want to make cleaning as a default behavior is that I encountered several times that the left over makes my successive training starts with a checkpoint instead of from scratch if I didn't change checkpoint path

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManager.scala
CodingCat: done

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
CodingCat: as explained above, cleaning checkpoint is the desired behavior. skipCleanCheckpoint is there mainly for testing

regarding the use cases you want to continue training, I think that belongs to another feature that you can start training from an existing model

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManagerSuite.scala
CodingCat: fixed

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManagerSuite.scala
CodingCat: fixed

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManagerSuite.scala
CodingCat: fixed

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/CheckpointManager.scala
trams: 👍 Now I understand you motivation

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
trams: Agree on your proposed feature: train starting from an existing model

PR #4757, opened 2019-08-11
add os.PathLike support for file paths to DMatrix and Booster Python classes

EvanKepner:
Hi - this PR adds support for `os.PathLike` file paths through the `DMatrix` and `Booster` classes in the Python implementation. As of Python 3.4, `pathlib` has become part of the standard library. The reason for this patch is to provide pathlike file system support similar to `numpy` e.g., `numpy.genfromtxt()` which can take `os.PathLike` objects.

Standard library references:
1. `os.fspath`: https://docs.python.org/3/library/os.html#os.fspath
2. `os.PathLike`: https://docs.python.org/3/library/os.html#os.PathLike
3. `pathlib`: https://docs.python.org/3/library/pathlib.html#module-pathlib

A "prior art" version of this support exists in `numpy`, and I've made the attribution to the reference I used in `compat.py` which will allow for the multiple Python version compatibility: https://github.com/numpy/numpy/blob/v1.17.0/numpy/compat/py3k.py#L188-L247

Major modifications:

1. `compat.py` has the back-port compatible version of `os.PathLike` and `os.fspath` implemented as `os_PathLike` and `os_fspath`. I've cited the `numpy` implementation.

2. `core.py` imports `os_PathLike` and `os_fspath` from `compat.py`. In instances where a file path is used the string support remains, but now a pathlike object can be used. `os_fspath` is used to transform paths into strings (and if a string is passed, it is returned without modification). Doc-strings have been updated throughout.

3. `test_basic_pathlike.py` is a new test file that uses an identical setup to `test_basic.py`. The modification is to use `Path` objects throughout with the overloaded `/` operator for path construction in the tests. These tests are skipped if the `from pathlib import Path` import fails (as marked in `testing.py`).

Thanks for your consideration :+1:


EvanKepner:
Looks like the [appveyor failure](https://ci.appveyor.com/project/tqchen/xgboost/builds/26614127/job/cxxjimqfp5t0vspk) is related to some form of network or DNS error, but all other status checks pass. Given that, I won't add any additional commits until review (assuming this is functionality you want). From a user perspective this PR enables changes like:

```python
from pathlib import Path

dpath = Path('demo') / 'data'
dtrain = xgb.DMatrix(dpath / 'agaricus.txt.train')
```

trivialfis:
Thanks.  Will review soon.

trivialfis:
@hcho3 Do we have a lowest supported Python 3.x version?  Checking setup.py it's a list of Python versions including a `Python :: 3` entry?

hcho3:
https://github.com/dmlc/xgboost/blob/198f3a6c4a1f9b315395e8c3fbee5d829199b425/python-package/setup.py#L66

trivialfis:
@hcho3 Oops sorry, my mistake.  Thanks for the pointer.

python-package/xgboost/compat.py
```diff
@@ -24,6 +26,84 @@ def py_str(x):
...
+# pathlib
+try:
+    from pathlib import Path, PurePath  # noqa
+except ImportError:
+    Path = PurePath = None
+
+# Attribution:
+#https://github.com/numpy/numpy/blob/v1.17.0/numpy/compat/py3k.py#L188-L247
+# backport abc.ABC
+if sys.version_info[:2] >= (3, 4):
```
trivialfis: This test is not needed as we require >= 3.4.  Just use ABC.

tests/python/test_basic_pathlike.py
```diff
@@ -0,0 +1,115 @@
...
+except ImportError:
+    pass
+
+pytestmark = pytest.mark.skipif(**tm.no_pathlike())
+
+rng = np.random.RandomState(1994)
+
+class TestBasicPathLike(unittest.TestCase):
+
+    def test_basic_PathLike(self):
```
trivialfis: I don't think it takes a complete run of training to test this.  Try unittest.

tests/python/test_basic_pathlike.py
```diff
@@ -0,0 +1,115 @@
...
+        # save model
+        bst.save_model(Path('xgb.model'))
+        # load model and data in
+        bst2 = xgb.Booster(model_file=Path('xgb.model'))
+        dtest2 = xgb.DMatrix(Path('dtest.buffer'))
+        preds2 = bst2.predict(dtest2)
+        # assert they are the same
+        assert np.sum(np.abs(preds2 - preds)) == 0
+
+    def test_record_results_PathLike(self):
```
trivialfis: Ditto.

tests/python/test_basic_pathlike.py
```diff
@@ -0,0 +1,115 @@
...
+# instead of string file paths throughout.
+dpath = 'demo/data/'
+
+try:
+    from pathlib import Path
+    dpath = Path(dpath)
+except ImportError:
+    pass
+
+pytestmark = pytest.mark.skipif(**tm.no_pathlike())
```
trivialfis: skip if `no_pathlike` returns true?  Are you sure it's testing the one in here instead of the one in library?

python-package/xgboost/compat.py
EvanKepner: Good catch - will do.

tests/python/test_basic_pathlike.py
EvanKepner: Since 3.4 is the minimum expectation I can remove this conditional entirely. Pathlib will always be present given that min version: https://docs.python.org/3.4/library/pathlib.html

tests/python/test_basic_pathlike.py
EvanKepner: Will do, I was aiming for consistency with `test_basic` since these are relatively fast, but agreed a full run likely isn't needed. I'll remove this separated file and integrate with the other tests directly.

EvanKepner:
@trivialfis thanks for the fast review! Since Python 3.4 is the minimum expected version I'll do the refactoring you highlighted and add a new commit. I'll put a comment here for the second review to let you know.

For use case, if a user of `xgboost` wants to use `pathlib` to manage their files s/he will not experience a difference between Python 3.4 and 3.6 (or 3.7) in the interaction. `Pathlib` is in the standard library as of 3.4, but 3.6 introduced `os.fspath` and `os.PathLike` which is why the backport is needed in `compat.py`. This patch continues to support string paths as well, so if a user prefers that route it will continue to operate as-is.

If the minimum Python version of `xgboost` moves to be 3.6 the backport part of `compat.py` could be removed and `os.fspath` / `os.PathLike` could be used directly.

EvanKepner:
@trivialfis I've applied the discussed changes. This is ready for a second review at your convenience. The Jenkins checks passed. Here are the big updates:

1. Removed the `abc` import check if the version is under py3.4 in `compat.py`.
2. Small changes to the `os_fspath` definition for when the Python version is under 3.6 to remove the `pylint disable` comments (removing redundant `else` statements and checks for `Path` being `None`). 
3. Removed my added `test_basic_pathlike.py` file and opted for a new `TestBasicPathLike` class in `test_basic.py` that is more unit-test style around the specific functionality for path support. This also removes the need for a separate function in `testing.py`.

In general, I'm importing `pathlib` directly without a `try` block since 3.4 is the minimum expected version, and that includes `pathlib` in the standard library. The only version specific check is in `compat.py` for the back-port of `os_fspath` and `os_PathLike` which are introduced in 3.6.

trivialfis:
Thanks!

RAMitchell:
Nice work @EvanKepner, you are welcome to submit more PRs in future.

EvanKepner:
Thanks! Appreciate it, I'll keep my eyes open :+1:

PR #4759, opened 2019-08-11
add c_api: XGDMatrixCopyDataToCSR

sperlingxx:
The purpose of this PR is adding a c_api(XGDMatrixCopyDataToCSR), with which we can fetch data of `DMatrix` in language bindings other than C/C++.

Below is API,
```c++
XGB_DLL int XGDMatrixCopyDataToCSR(const DMatrixHandle handle,
                                   size_t **out_row_ptr,
                                   unsigned **out_indices,
                                   float **out_data);
```
Memory of copied CSR data isn't owned by this API, but owned by who (language bindings other than C/C++) call this API.

I will add api for python and java later, if this PR is approved.


sperlingxx:
@trivialfis Can you help to review this?

trivialfis:
Will look into it.

src/c_api/c_api.cc
```diff
@@ -758,6 +758,39 @@ XGB_DLL int XGDMatrixGetUIntInfo(const DMatrixHandle handle,
...
 
+XGB_DLL int XGDMatrixCopyDataCSR(const DMatrixHandle handle,
+                                 size_t **out_row_ptr,
+                                 unsigned **out_indices,
+                                 float **out_data) {
+  API_BEGIN();
+  CHECK_HANDLE();
+  const auto dmat = static_cast<std::shared_ptr<DMatrix>*>(handle)->get();
+
+  *out_row_ptr = reinterpret_cast<size_t *>(
```
trivialfis: You need to specify who owns the memory.  If XGBoost owns the memory then we also need to create a `free` function.  On the other hand if the user owns this memory we should have a way to tell how should they allocate memory, like size of `values` and `row_ptr_`.  I personally prefer the second option.

src/c_api/c_api.cc
```diff
@@ -758,6 +758,39 @@ XGB_DLL int XGDMatrixGetUIntInfo(const DMatrixHandle handle,
   API_END();
 }
 
+XGB_DLL int XGDMatrixCopyDataCSR(const DMatrixHandle handle,
+                                 size_t **out_row_ptr,
+                                 unsigned **out_indices,
```
trivialfis: Use `uint32_t`.

include/xgboost/c_api.h
```diff
@@ -307,6 +307,19 @@ XGB_DLL int XGDMatrixGetUIntInfo(const DMatrixHandle handle,
...
+
+/*!
+ * \brief copy data of matrix in CSR format
+ * \param handle a instance of data matrix
+ * \param out_row_ptr pointer to row offsets
+ * \param out_indices pointer to findex
+ * \param out_data pointer to fvalues
+ * \return 0 when success, -1 when failure happens
+ */
+XGB_DLL int XGDMatrixCopyDataCSR(const DMatrixHandle handle,
```
trivialfis: Use `CopyDataToCSR`, I don't mind verbose code.

tests/cpp/c_api/test_c_api.cc
```diff
@@ -63,3 +63,39 @@ TEST(c_api, XGDMatrixCreateFromMat_omp) {
     delete dmat;
   }
 }
+
+TEST(c_api, XGDMatrixCopyDataCSR) {
+  omp_set_num_threads(2);
```
trivialfis: Store the original value first, and change it back at the end.  Don't assume it's running single threaded.  You can create a context class with constructor and destructor doing these jobs.

tests/cpp/c_api/test_c_api.cc
```diff
@@ -63,3 +63,39 @@ TEST(c_api, XGDMatrixCreateFromMat_omp) {
...
+  for (int i = 0; i < row_ptr.size(); ++i) {
+    ASSERT_EQ(row_ptr[i], row_ptr_[i]);
+  }
+  for (size_t i = 0; i < row_ptr.back(); ++i) {
+    ASSERT_EQ(indices[i], indices_[i]);
+    ASSERT_EQ(data[i], data_[i]);
+  }
+
+  delete dmat;
+  free(row_ptr_);
```
trivialfis: As in above, If XGBoost owns the memory you need to create an abstracted `free` method.   In the user side, calling `delete` and `free` are two different things.  In XGBoost by default `malloc` is from libc, but the user can be using jemalloc ...

Also as in c++, choose `std::vector` first, then `new/delete`, then `malloc` unless you are creating your own memory pool.

src/c_api/c_api.cc
```diff
@@ -758,6 +758,39 @@ XGB_DLL int XGDMatrixGetUIntInfo(const DMatrixHandle handle,
...
+
+  *out_row_ptr = reinterpret_cast<size_t *>(
+      std::malloc(dmat->Info().num_row_ * sizeof(size_t)));
+  *out_indices = reinterpret_cast<unsigned *>(
+      std::malloc(dmat->Info().num_nonzero_ * sizeof(unsigned)));
+  *out_data = reinterpret_cast<float *>(
+      std::malloc(dmat->Info().num_nonzero_ * sizeof(float)));
+
+  size_t row_offset = 0;
+  for (const auto &batch : dmat->GetBatches<SparsePage>()) {
```
trivialfis: What if the data is external memory?

src/c_api/c_api.cc
sperlingxx: Thanks for detailed comments. Since these method will be created language bindings other than C/C++, maybe the first approach is better? After read all you comments, I think this PR need to be refactored into a class like `DMatrixView`?  In this way, memory can be freed  
 by destructor.

trivialfis:
@sperlingxx Please try to carry it out.

sperlingxx:
@trivialfis I will refine this PR asap, but these days, I am quite busy with org business.

trivialfis:
@sperlingxx no problem.  It's good thing to have be occupied with something.

trivialfis:
Still don't understand how can I break the linter ...

PR #4767, opened 2019-08-13
Fix unstable results in hist method

SmirnovEgorRu:
After my [changes](https://github.com/dmlc/xgboost/pull/4529) I observed some small differences in results from run to run.
For example, log-loss for HIGGS data set:
1 run: 0.807342768946606615
2 run: 0.807342768708895875

The reason was so easy - guided scheduling instead of static in OpenMP loop. It leads to non-deterministic operations with floating-point numbers.

I have fixed it and now log-loss:
1 run: 0.807342767697801778
2 run: 0.807342767697801778
3 run: 0.807342767697801778

Training time is the same as it were before:
Before time: 2951ms
After time: 2903ms

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -2,9 +2,9 @@ Contributors of DMLC/XGBoost
 ============================
 XGBoost has been developed and used by a group of active community. Everyone is more than welcomed to is a great way to make the project better and more accessible to more users.
 
-Project Management Committee(PMC) 
+Project Management Committee(PMC)
 ----------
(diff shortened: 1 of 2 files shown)


SmirnovEgorRu:
Ci contains following error:
tests/python/test_linear.py::TestLinear::test_shotgun terminate called after throwing an instance of 'std::bad_alloc'
[2019-08-13T11:59:10.252Z]   what():  std::bad_alloc

I don't sure, that it is my problem, probably some infrastructure issue. Can we restart this?

hcho3:
See #4716, which already removed `guided` scheduling.

PR #4769, opened 2019-08-14
[native xgb distributed training] allow failed worker retry

chenqin:
The goal of this pr is to enable native xgb worker retry on both approx and fast hist tree_method, 
https://github.com/dmlc/rabit/pull/98

- via enable rabit_cache=1 setting, user can retry failed xgb worker without restart entire cluster
- add rabit_recovery test case to traivs
- add missing parameters to checkpoint payload
- detect older model to be backward compatible 

Misc: I will update gitmodules once dependency pr landed.


trivialfis:
Is it possible to use CMake instead of Make?

chenqin:
> Is it possible to use CMake instead of Make?

Looks like building librabit_mock.a with rabit CMake will not build correctly. xgboost cmake has separate logic to build librabit.a. Looking...

chenqin:
> Is it possible to use CMake instead of Make?
@trivialfis 
hum, switched to cmake and passed travis-ci, jenkinsis still failing not sure why

trivialfis:
@chenqin clang-tidy?

chenqin:
moved to https://github.com/dmlc/xgboost/pull/4808

PR #4773, opened 2019-08-14
make HostDeviceVector single gpu only

### LINKED ISSUE #4531: [RFC] Remove support for single process multi-GPU

rongou:
Closes #4531 
Part of 1.0.0 roadmap #4680 

Finally got the CI to pass. I can do more cleanups, but thought you guys should take a look first.


trivialfis:
@rongou I haven't looked into this in detail yet.  But seems `GPUDistribution` can be removed too?

rongou:
@trivialfis yes there is a whole bunch of stuff that can be removed or cleaned up. I didn't want to make this bigger than it already is, but I'll do some of the more obvious ones.

trivialfis:
@rongou Nice.  Please ping me when it's ready.

rongou:
It's ready now.

src/linear/updater_gpu_coordinate.cu
```diff
@@ -167,33 +167,28 @@ class GPUCoordinateUpdater : public LinearUpdater {
                       const gbm::GBLinearModelParam &model_param) {
     if (!shards_.empty()) return;
 
-    dist_ = GPUDistribution::Block(GPUSet::All(learner_param_->gpu_id, learner_param_->n_gpus,
-                                               p_fmat->Info().num_row_));
-    auto devices = dist_.Devices();
+    device_ = learner_param_->gpu_id;
 
-    size_t n_devices = static_cast<size_t>(devices.Size());
     size_t row_begin = 0;
```
sriramch: i think this entire block can be removed and `0, num_row` directly passed to `DeviceShard` constructor; even better if `DeviceShard` does not use row indices any more (and uses num_row)

src/predictor/gpu_predictor.cu
```diff
@@ -352,40 +310,45 @@ class GPUPredictor : public xgboost::Predictor {
     InitModel(model, tree_begin, tree_end);
 
     size_t batch_offset = 0;
+    auto* preds = out_preds;
+    std::unique_ptr<HostDeviceVector<bst_float>> batch_preds{nullptr};
```
sriramch: instead of dual copy: 
- copying input prediction vector into a temp one based on batch size and, 
- copying back the output prediction vector from the kernel back into `out_preds` 

is it not possible to:
-  slice and dice the `out_preds` based on batch size - basically, get a device span from `out_preds` and get a subspan based on the batch size and batch offset
- pass the subspan to predict internal as a pass through to the kernel

src/predictor/gpu_predictor.cu
```diff
@@ -500,25 +463,27 @@ class GPUPredictor : public xgboost::Predictor {
...
     shards_.clear();
-    shards_.resize(devices_.Size());
+    shards_.resize(1);
     dh::ExecuteIndexShards(&shards_, [=](size_t i, DeviceShard& shard){
-        shard.Init(devices_.DeviceId(i));
+        shard.Init(device_);
       });
   }
 
   std::vector<DeviceShard> shards_;
```
sriramch: similar to earlier and applies to all: this need not be an array anymore

src/common/device_helpers.cuh
```diff
@@ -72,30 +72,6 @@ const T *Raw(const thrust::device_vector<T> &v) {  //  NOLINT
...
-}
-
-inline void SynchronizeAll() {
-  for (int device_idx : xgboost::GPUSet::AllVisible()) {
-    safe_cuda(cudaSetDevice(device_idx));
-    safe_cuda(cudaDeviceSynchronize());
-  }
-}
-
-inline size_t AvailableMemory(int device_idx) {
```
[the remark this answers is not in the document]
rongou: Added back.

src/common/hist_util.cu
```diff
@@ -401,17 +401,15 @@ struct GPUSketcher {
...
 
   void SketchBatch(const SparsePage &batch, const MetaInfo &info) {
-    GPUDistribution dist =
-      GPUDistribution::Block(GPUSet::All(generic_param_.gpu_id, generic_param_.n_gpus,
-                                         batch.Size()));
+    auto device = generic_param_.gpu_id;
 
     // create device shards
-    shards_.resize(dist.Devices().Size());
+    shards_.resize(1);
```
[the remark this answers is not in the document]
rongou: Done.

src/common/hist_util.cu
```diff
@@ -401,17 +401,15 @@ struct GPUSketcher {
...
     dh::ExecuteIndexShards(&shards_, [&](int i, std::unique_ptr<DeviceShard>& shard) {
```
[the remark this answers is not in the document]
rongou: Done.

src/common/host_device_vector.cu
```diff
@@ -210,58 +208,54 @@ struct HostDeviceVectorImpl {
   }
 
   void InitShards() {
-    int ndevices = distribution_.devices_.Size();
+    int ndevices = device_ >= 0 ? 1 : 0;
     shards_.resize(ndevices);
     dh::ExecuteIndexShards(&shards_, [&](int i, DeviceShard& shard) {
```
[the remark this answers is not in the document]
rongou: Done.

src/common/host_device_vector.cu
```diff
@@ -362,35 +356,26 @@ struct HostDeviceVectorImpl {
...
   }
 
-  void Shard(GPUSet new_devices) {
-    if (distribution_.Devices() == new_devices) { return; }
-    Shard(GPUDistribution::Block(new_devices));
-  }
-
-  void Reshard(const GPUDistribution &distribution) {
-    if (distribution_ == distribution) { return; }
+  void Reshard(int device) {
```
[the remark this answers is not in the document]
rongou: Done.

src/linear/updater_gpu_coordinate.cu
rongou: Done.

src/predictor/gpu_predictor.cu

src/predictor/gpu_predictor.cu
rongou: That'll assume the whole predictor vector will be on device, right? I think the current approach doesn't put the whole vector on device in external memory mode, although I haven't 100% verified it.

rongou:
@RAMitchell @trivialfis @sriramch I think this is as far as this PR would go. Please take another look.

I still want to remove the sharding in `updater_gpu_hist.cu`, but it seems a bit more involved, so probably leave it to a follow up PR.

src/predictor/gpu_predictor.cu
sriramch: even if the device copy is prevented here, it'll be copied right back (soon after this) in the objective functions while computing the gradients, or later when the prediction cache is updated during training.

PR #4778, opened 2019-08-15
Remove `UseGPU` in GBM.

trivialfis:
Clean up after removing `n_gpus`.


trivialfis:
Let's don't rush into this for now.  The dependent on `GPUSet` is quite risky.

PR #4789, opened 2019-08-18
Remove plugin, cuda related code in automake & autoconf files

### LINKED ISSUE #4627: Plugin example not working for python
The plugin example (https://github.com/dmlc/xgboost/tree/master/plugin/example) does not work. When I compile from source as a python package after following instructions for the plugin example I get the error "xgboost.core.XGBoostError: b'[03:34:57] /home/willie/xgboost/src/objective/objective.cc:23: Unknown objective function mylogistic\n\n".

trivialfis:
Closes #4627.

Default to CMake instead.

Breaking: Makefile will no longer work with plugins.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,11 +116,12 @@ if (R_LIB)
 endif (R_LIB)
 
 # core xgboost
-add_subdirectory(${PROJECT_SOURCE_DIR}/src)
+add_subdirectory(${xgboost_SOURCE_DIR}/plugin)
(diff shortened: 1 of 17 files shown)


hcho3:
So we are requiring CMake for CUDA code? I’m up for it, as long as it is clearly documented.

trivialfis:
@hcho3 I don't think it works before this PR.  I can't find any reference of `nvcc` `CUDA` or `*.cu`.

Issue #4600, opened 2019-06-23
New way of specifying device.


trivialfis:
Following our deprecation of `n_gpus` parameter, we may need a new way to specify device XGBoost will run on.  There are a few scenarios we need to consider:

- When training, the device is automatically selected based on `tree_method`.  Do we still want to support running predictor and metrics on CPU while updater running on GPU?
- During prediction, how do we specify what device to run on like CPU/ GPU:0/ GPU:1 ... or even multiple CPUs?  It seems putting everything on `gpu_id` is ambiguous.

@RAMitchell

RAMitchell:
These are all good questions, for which I have no answers right now :)

We can introduce a `device` parameter indicating the preferred device to run any given function on. Do you have any other suggestions?

trivialfis:
Personally I really like the tensorflow's device context, but I'm not sure about how the other languages would implement something like that.  E.g.
```
with device('/gpu:0'):
    ...
```
Maybe `booster.set_device('/gpu:0')`, I'm not sure yet, hence the issue.

RAMitchell:
I've been thinking that this is not a critical issue for now. Evaluation can't run on the gpu when called independent of training but this is not a big deal as evaluation is rarely a bottleneck (simple calculations proportional only to number of rows). Prediction can still occur on device via setting the booster params.

PR #4795, opened 2019-08-20
Prevent copying data to host.

### LINKED ISSUE #4751: CPU predictor crashes when training with data stored in GPU memory.
We are working on the GPU training pipeline, to load data into GPU memory and choose `gpu_hist` to run the training locally, non-distributed.

After some basic investigation, found that the first `PredictRaw` in `Learner` will use `cpu_predcitor`, then it crashes at line 71 in file https://github.com/dmlc/xgboost/blob/master/src/predictor/cpu_predictor.cc 

And the crash happens only when parallel is enabled. Error message just before crash is followed

> double free or corruption (out)
double free or corruption (!prev)
double free or corruption (!prev)

What' more interesting is it can pass after even only adding a log print as below,
` 69         }`
 `70         for (int k = 0; k < kUnroll; ++k) {`
 **`71           LOG(WARNING) << "debug == tid = " << tid;`**
 `72           inst[k] = batch[i + k];`
` 73         }`

I think running parallel here may be too fast for GPU -> CPU case, since it may need a little time to copy the batch data from GPU to CPU memory, what's more the sync is lazy per code below from file https://github.com/dmlc/xgboost/blob/master/src/common/host_device_vector.cu . 
`  const std::vector<T>& ConstHostVector() {`
`    LazySyncHost(GPUAccess::kRead);`
`    return data_h_;`
`  }`

Any idea to fix it ? or to fix why cpu_predictor is used when doing the first `PredictRaw` even specifying the tree method as 'gpu_hist'.

trivialfis:
Closes #4751 .

### DIFF
--- a/src/gbm/gbtree.cc
+++ b/src/gbm/gbtree.cc
@@ -57,6 +57,14 @@ void GBTree::Configure(const Args& cfg) {
 
   monitor_.Init("GBTree");
 
+  if (tparam_.tree_method == TreeMethod::kGPUHist &&
+      std::none_of(cfg.cbegin(), cfg.cend(),
+                   [](std::pair<std::string, std::string> const& arg) {
+                     return arg.first == "predictor";
+                   })) {
+    tparam_.predictor = "gpu_predictor";
+  }
+
   configured_ = true;
 }
 
@@ -88,7 +96,7 @@ void GBTree::PerformTreeMethodHeuristic(std::map<std::string, std::string> const
     // set, since only experts are expected to do so.
     return;
   }
-
+  // tparam_ is set before calling this function.
   if (tparam_.tree_method != TreeMethod::kAuto) {
     return;
   }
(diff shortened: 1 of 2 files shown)


trivialfis:
There are still small copying scattered over the place.  Including `ColumnSampler` and `label_correct_` in regression obj.  But the big ones are addressed.

src/gbm/gbtree.cc
```diff
@@ -57,6 +57,14 @@ void GBTree::Configure(const Args& cfg) {
 
   monitor_.Init("GBTree");
 
+  if (tparam_.tree_method == TreeMethod::kGPUHist &&
+      std::none_of(cfg.cbegin(), cfg.cend(),
+                   [](std::pair<std::string, std::string> const& arg) {
+                     return arg.first == "predictor";
```
RAMitchell: Is this check correct? Are you sure this change won't cause the cpu_predictor param to be overwritten?

trivialfis: @RAMitchell 

- If "predictor" is specified by user, then it should appear in this configuration and be found by this `std::none_of`, hence nothing changes.
- If `predictor` is loaded from binary model, it should also appear in this configuration, nothing changes.
- If `predictor` is never set, and `tree_method` is not `gpu_hist`, nothing changes.
- If `predictor` is never set, but `tree_method` is `gpu_hist`, then `predictor` **is changed** to `gpu_predictor`.

I most certainly don't like the configuration.  In an ideal world there should be only one `hist` method and only one `predictor`, just like our objective functions and metrics.  We dispatch them by `gpu_id` or by #4600 .  But I don't think we can do that in near future ...

RAMitchell: My mistake, I thought this was checking the second part of the pair for some reason.

PR #4798, opened 2019-08-21
Fix auc error in rank task

tinkle1129:
There was a dataset for ranking task, and most samples are negative. I tried to use train it in distributed mode and found that all workers contains at least one group which had only pos or neg samples. As the result, I got the exception: AUC: the dataset only contains pos or neg samples.Besides, I found that "auc_error" would be changed into "1" if there was a group full of negative or positive samples in each worker.
This PR modified the response of auc_error. In distribute mode, auc_error is regarded as the number of invalid groups in each worker. Workers will reserve the number of valid groups and calculate average auc among all valid groups from all workers.


trivialfis:
@hcho3 Could you please help taking a look?

trivialfis:
Thanks @tinkle1129 @hcho3

PR #4801, opened 2019-08-22
Restrict access to `cfg_` in gbm.

trivialfis:
For parameter validation, I need to remove the copy of configuration.  Or at least copy only what's necessary.  After this PR, the copied configuration will only be used by initializing `updaters` and `TreeParam`.  I plan to deprecate `TreeParam` after introducing JSON.

### DIFF
--- a/include/xgboost/gbm.h
+++ b/include/xgboost/gbm.h
@@ -154,6 +154,12 @@ class GradientBooster {
       GenericParameter const* gparam,
       const std::vector<std::shared_ptr<DMatrix> >& cache_mats,
       bst_float base_margin);
+
+  static void AssertGPUSupport() {
+#ifndef XGBOOST_USE_CUDA
+    LOG(FATAL) << "XGBoost version not compiled with GPU support.";
+#endif  // XGBOOST_USE_CUDA
+  }
 };
 
 /*!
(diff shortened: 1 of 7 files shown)


src/gbm/gblinear.cc
```diff
@@ -65,7 +65,11 @@ class GBLinear : public GradientBooster {
     updater_.reset(LinearUpdater::Create(param_.updater, learner_param_));
     updater_->Configure(cfg);
     monitor_.Init("GBLinear");
+    if (param_.updater == "gpu_coord_descent") {
```
chenqin: Can we have a helper function which does those translation and checks instead?

include/xgboost/gbm.h
```diff
@@ -154,6 +154,12 @@ class GradientBooster {
       GenericParameter const* gparam,
       const std::vector<std::shared_ptr<DMatrix> >& cache_mats,
       bst_float base_margin);
+
+  static void AssertGPUSupport() {
```
chenqin: Again, we should move to a helper function and do checks and validation against permutation of configurations

src/gbm/gblinear.cc
trivialfis: @chenqin I'm working toward that.  Full validation for all parameters like whether some parameters are not used.  But right now I need to eliminate the copies of these parameters.

trivialfis:
Interesting.  I actually found a bug after adding this validation.

trivialfis:
@hcho3 Copied from commit message:

Remove `grow_global_histmaker`
    
    This updater is the same as `grow_histmaker`.  The former is not in our
    document so we just remove it.

PR #4802, opened 2019-08-22
Restrict dependency versions to ones that support Python2.

A87P:
Do not merge it to the `master`. It only serves for the purpose of patching up dependency versions for the `v0.82` version of the `xgboost`.


A87P:
Should not be merged to master.

Issue #4727, opened 2019-08-02
[jvm-package] Add documentation warning for saving Spark trained model with non-default missing value in native format


cpfarrell:
When training XGBoost models on Spark it is possible to set the value of "missing" as part of the parameters of the model (more description at https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html#dealing-with-missing-values). If you save the model in the format needed to be loaded in other bindings (via https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html#interact-with-other-bindings-of-xgboost) then this missing parameter gets dropped (in general parameters not being included in the model is discussed in https://github.com/dmlc/xgboost/issues/4104). If the native model is then loaded (either on another platform or even again into Spark) the absence of this missing parameter will cause predictions to be inaccurate. An especially confusing aspect of this is that the missing parameter in python is a property of the DMatrix and so is a property of the dataset fed to XGBoost, in Spark however it's part of the model's parameters and so a property of the model. It can then be easy to forget to set the parameter correctly when constructing your DMatrix in python since it seems like it would be already baked into the model.

I imagine including the value of the missing parameter along with the model is likely blocked by https://github.com/dmlc/xgboost/issues/3980 but would it be possible to add to the documentation page about dealing with missing values (https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html#dealing-with-missing-values) that care needs to be taken to set the missing parameter correctly on the other side if saving the model in native format?

hcho3:
Resolved by #4727

PR #4805, opened 2019-08-23
[jvm-packages] Allow for bypassing spark missing value check

cpfarrell:
This pull request adds a new parameter in Spark XGBoost that allows for bypassing assertion added as part of https://github.com/dmlc/xgboost/pull/4349/files (related comment https://github.com/dmlc/xgboost/pull/4349#issuecomment-517506892)

The reason for this change is that I have my own VectorAssembler implementation that allows for creating sparse vectors where any value can be encoded as missing. I had been using this assembler to create feature vectors with NaN encoded as missing which integrated well with XGBoost (both in Spark and saving the models in binary form for use on other platforms). This change would let me continue doing that by providing this special flag, essentially its a way of addressing the "Note" section at https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html#dealing-with-missing-values. I plan on providing my VectorAssembler implementation here in case any one else might be interested in it but will take just a little bit of time to do that.

For the actual code here: I mostly just copied what was happening for the missing parameter and passed my new variable around the same way. I tried to follow the style of the rest of the code base but I've never written Scala before so likely have a few style issues would be happy to hear about.

Probably the biggest thing here is how to make it clear why this exists, I am not at all sure on my naming so welcome feedback. Also should maybe document more what it does and how it should be used (including when it shouldn't be used).

For testing, I ran jvm-packages/dev/build-linux.sh and it succeeded. Would like to try actually running this inside a Spark job to verify it works how I think it does.

Let me know thoughts on this.

### DIFF
--- a/doc/jvm/xgboost4j_spark_tutorial.rst
+++ b/doc/jvm/xgboost4j_spark_tutorial.rst
@@ -156,24 +156,9 @@ labels. A DataFrame like this (containing vector-represented features and numeri
 Dealing with missing values
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Strategies to handle missing values (and therefore overcome issues as above):
-
-In the case that a feature column contains missing values for any reason (could be related to business logic / wrong data ingestion process / etc.), the user should decide on a strategy of how to handle it.
-The choice of approach depends on the value representing 'missing' which fall into four different categories:
-
-1. 0
-2. NaN
-3. Null
-4. Any other value which is not mentioned in (1) / (2) / (3)
-
-We introduce the following approaches dealing with missing value and their fitting scenarios:
-
-1. Skip VectorAssembler (using setHandleInvalid = "skip") directly. Used in (2), (3).
-2. Keep it (using setHandleInvalid = "keep"), and set the "missing" parameter in XGBClassifier/XGBRegressor as the value representing missing. Used in (2) and (4).
-3. Keep it (using setHandleInvalid = "keep") and transform to other irregular values. Used in (3).
-4. Nothing to be done, used in (1).
-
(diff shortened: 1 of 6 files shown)


cpfarrell:
@CodingCat FYI since talked with you about this in https://github.com/dmlc/xgboost/pull/4349#issuecomment-517529693

cpfarrell:
Thaks, and yea will do on updating docs.

Mind if I do https://github.com/dmlc/xgboost/issues/4727 at the same time since would be modifying the same place in documentation? Or could do it in separate branch if you want.

CodingCat:
Sure, let’s do together

cpfarrell:
Due to various things I probably won't be able to get back to this for 3-4 weeks but am still planning on updating the docs so can push this

cpfarrell:
I just updated review with a new version after rebasing on master and updating documentation for how to handle missing values. I ended up rewriting the documentation a decent amount as I've seen this issue cause lots of confusion with people trying to use spark on xgboost and I'm not sure what was there before was exactly right. Open to reverting those changes if desired though (or changed in any way). Also open to adding more explanation about what makes this so tricky (I've found it hard to explain to folks) but that might become kind of long in the documentation.

Also, any thought of having an implementation of a VectorAssembler that would keep zeros checked into this project so could directly point people at it to use and thus make integration easier

hcho3:
@CodingCat Is this ready to be merged?

CodingCat:
In traveling, no laptop, will look at early this next week

CodingCat:
merged, thanks for the contribution!

sriramch:
@cpfarrell @hcho3 this is failing builds. can someone please help looking into this?

hcho3:
@sriramch It appears a minor infraction in Scala style check.

hcho3:
See #5134

Issue #4786, opened 2019-08-17
[RFC] Design of Checkpoint Mechanism in XGBoost-Spark


CodingCat:
## Checkpoint Cleanup

The current implementation in XGBoost-Spark does not clean up the checkpoint file after a successful training. As a result, the user may get a wrong training process without a careful setup. For instance, the following the process will lead to a wrongly-formed model in the second run.

User starts training by setting checkpoint interval as 2 iterations and round as 10 iterations
After a successful training, the user starts a new training job with a new group of parameters (except the changing of checkpoint directory)

The second training job will finish “too fast” for two reasons:

The checkpoint built at the 8th iteration in the first was left in the checkpoint directory
Without a different checkpoint dir setup, the second run will load the left checkpoint file and only run for 2 iterations 

Therefore, we propose to always cleanup the checkpoint directory with a successful training.

## Deterministic Partitioning

Based on the definition of gradient boosting, we should ensure that the input dataset to a booster is fixed for each iteration. However, this would be hard to achieve in the distributed training with Spark.

The current partitioning mechanism in XGBoost-Spark is to ensure that the input data can be dispatched to the correct number of partitions. The number of partitions is controlled by the user’s configuration on parameter numWorkers.

The design of deterministic partitioning is to ensure that given a fixed input dataset the partitioning of the strategy is always the same. This goal is interpreted as follows:

We rely on the user to ensure that the data feed to XGBoost-Spark is fixed for each checkpoint interval
XGBoost-Spark guarantees that the partitioning of the given dataset is deterministic

The current partitioning strategy in XGBoost-Spark cannot achieve the second part of the goal. The current strategy is based on the repartition API in Spark RDD which is implemented to ensure an even distribution. The execution of the current partitioning is like the following

We have A partitions and would like to repartition to B partitions 
Each element in each partition of A is dispatched to B partitions but randomly start from one of B partitions.

Because of the random start, we cannot guarantee the fixed partitioning strategy for each checkpoint interval. 

We propose the following mechanism to have a deterministic partitioning and achieve load balancing with the best effort

partitionId = math.abs(preventOverflow(row.getHashCode + row.getAs[Float](math.abs(row.getHashCode) % number_non_zero_features)) % numWorkers

## Avoid Multiple Jobs for Checkpointing

[...]

PR #4807, opened 2019-08-25
[jvm-packages] enable deterministic repartitioning when checkpoint is enabled

CodingCat:
part of https://github.com/dmlc/xgboost/issues/4786

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
@@ -18,6 +18,7 @@ package ml.dmlc.xgboost4j.scala.spark
 
 import ml.dmlc.xgboost4j.{LabeledPoint => XGBLabeledPoint}
 
+import org.apache.spark.HashPartitioner
 import org.apache.spark.ml.feature.{LabeledPoint => MLLabeledPoint}
 import org.apache.spark.ml.linalg.{DenseVector, SparseVector, Vector, Vectors}
 import org.apache.spark.ml.param.Param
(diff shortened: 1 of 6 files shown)


CodingCat:
@trams

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostEstimatorCommon.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import ml.dmlc.xgboost4j.scala.spark.params._
+
+import org.apache.spark.ml.param.shared.HasWeightCol
+
+private[spark] sealed trait XGBoostEstimatorCommon extends GeneralParams with LearningTaskParams
+  with BoosterParams with ParamMapFuncs with NonParamVariables {
+
+  def needDeterministicRepartitioning: Boolean = {
+    getCheckpointPath.nonEmpty && getCheckpointInterval > 0
```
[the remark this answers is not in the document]
CodingCat: > it our code checkpoint_interval is set to a very big number by default and checkpoint path is always set

shall Spark detect that someone enabled logging but redirect the logging from stdout/stderr to /dev/null?

same thing here, we don't want to spend time on such stuffs

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
```diff
@@ -73,12 +74,81 @@ object DataUtils extends Serializable {
...
+  private def calculatePartitionKey(row: Row, numPartitions: Int): Int = {
+    val Row(_, features: Vector, _, _) = row
+    val rowHashCode = row.hashCode()
+    val featureValue = features match {
+      case denseVector: DenseVector =>
+        featureValueOfDenseVector(rowHashCode, denseVector)
+      case sparseVector: SparseVector =>
+        featureValueOfSparseVector(rowHashCode, sparseVector)
+    }
+    (rowHashCode.toLong + featureValue).toInt % numPartitions
```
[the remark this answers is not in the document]
CodingCat: > I still do not understand why do we need to add featureValue to already "quite random" hash code?

you cannot guarantee row.hashCode is conflict-free, if you go into the details of how MurmurHash3 is implemented. The conflict will come from two parts: (1) hashcode conflict of the column value; (2) the lower 26 bits conflicts in the mixed value of 'Seq' and the hash code of column value;

I didn't see any problem by bring an additional 32 bits to avoid conflict

>  I am a bit concern that featureValue + rowHashCode could be more then MAX_INT.

I made a mistake here, I will address it

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
```diff
@@ -73,12 +74,81 @@ object DataUtils extends Serializable {
...
+  private def calculatePartitionKey(row: Row, numPartitions: Int): Int = {
+    val Row(_, features: Vector, _, _) = row
+    val rowHashCode = row.hashCode()
+    val featureValue = features match {
+      case denseVector: DenseVector =>
+        featureValueOfDenseVector(rowHashCode, denseVector)
+      case sparseVector: SparseVector =>
+        featureValueOfSparseVector(rowHashCode, sparseVector)
+    }
+    math.abs((rowHashCode.toLong + featureValue).toString.hashCode % numPartitions)
```
CodingCat: @trams please check this 

this has been `Long + Double and toString and hashCode `

it shall prevent overflow

CodingCat:
@trivialfis any idea on the flaky gpu test?

trivialfis:
Looks like network failure ...

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
CodingCat: Long + Double is a double, mod a double with an integer and then to int creates many conflicts in partition number

e.g. any value in [1.0, 2.0) will be assigned to partition 1....

(actually rowHashCode.toDouble + featureValue is more helpful for reading

trams: I see your point but I do not think that any feature value with `[1.0, 2.0)` will be assigned a partition 1 given that this featureValue should have different hash codes

I do not mind toString if it is better but I would prefer to do without it.

What about using `rowHashCode.toLong + featureValue.toLong`

P.S. I really do not have an experience building hash functions and I am always paranoid to accidentally take the bad one so I am fine with toString. I just want to make sure we looked at all options

trams: Also what do you think about completely different option: "Use murmur to combine rowHashCode and featureValue into a good hash value"

P.S. I am just spitballing here so the idea may be completely stupid

CodingCat: AFK, just quick response to why [1,2) will be assigned to the same partition

Your propsoal is 
math.abs((rowHashCode.toLong + featureValue) % numPartitions.toLong).toInt

Because of toInt here, 1.1.toInt will be 1, 1.6.toInt will be the same

trams: I understand that 1.1.toInt = 1.6.toInt = 1 but it is not obvious to be it is a problem to apply `.toInt` to `featureValue`. Yes in case
1.rowHashCode.toLong = x, featureValue = 1.1 we would have featureValue.toInt = 1 the result will be (x + 1) % numPartitions
2. .rowHashCode.toLong = y, featureValue = 1.6 we would have featureValue.toInt = 1 the result will be (y + 1) % numPartitions

given that x != y because featureValue is part of Row and different it should (key word should) not be a problem

But I am pretty sure I am missing something else here. I do not want to bikeshed here and prematurely optimize (my concern with toString is mostly about performance and excessive allocations of Strings) cause I think this change is awesome and will solve a lot of problems

All this discussion made me think. Do we have any tests on "real" data which shows that this hash function is "uniform" to some extend?

PR #4808, opened 2019-08-25
[rabit_bootstrap_cache ] failed xgb worker recover from other workers

chenqin:
[copy from xgboost/pull/4769] The goal of this pr is to enable failed native xgb worker retry and restore on **approx** tree_method detail underlaying implementation can be found  https://github.com/dmlc/rabit/pull/98

Summary:
- via enable rabit_bootstrap_cache=1 setting, user can retry failed xgb worker without restart entire job
- add xgb_recovery test case to traivs to simulate multiple xgb worker failure  on approx three_method with prediction accuracy on par
- add needed cfg_ to native checkpoint payload if user set rabit_bootstrap_cache=1
- backward compatible with old model or disabled/not set rabit_bootstrap_cache case.

Note: per convo with @CodingCat fast histgram in master is still not ready to support.


hcho3:
Running tests locally, for faster experimentation: https://xgboost.readthedocs.io/en/latest/contrib/unit_tests.html

chenqin:
I did tests of commenting extra config in learner, it pass failed jvm tests on model save/load . So looks like jvm saved model were impacted with extra configs. @CodingCat F.Y.I
std::set<std::string> saved_configs_ = {};
  //{"max_depth", "tree_method", "dsplit",
  //  "seed", "silent", "gamma", "min_child_weight"};

chenqin:
@trivialfis since you were working on organizing trainer parameters, can you help review this change. We allow failed xgb worker retry with additional saved_params IF user opt-ed in rabit_bootstrap_cache

cc @hcho3 @CodingCat

trivialfis:
Yes.  I'm also playing with rabit recently.  Will review soon.  Thanks for mentioning me.

chenqin:
image

trivialfis:
@chenqin Is this currently critical?  I wrote the CMake file before, the reason I build rabit in XGBoost is because it fails to build on Windows.  Would it be inconvenient for you if I take some time to open a PR for rabit cmake build file?

src/learner.cc
```diff
@@ -332,8 +339,15 @@ class LearnerImpl : public Learner {
       }
     }
     {
-      // Write `predictor`, `gpu_id` parameters as extra attributes
-      for (const auto& key : std::vector<std::string>{"predictor", "gpu_id"}) {
+      std::vector<std::string> saved_params{"predictor", "n_gpus", "gpu_id"};
```
trivialfis: Why is `n_gpus` here.  Is it from an old commit?

src/learner.cc
```diff
@@ -648,6 +662,10 @@ class LearnerImpl : public Learner {
   std::vector<std::shared_ptr<DMatrix> > cache_;
 
   common::Monitor monitor_;
+
+  /*! \brief saved config keys used to restore failed worker */
+  std::set<std::string> saved_configs_ = {"max_depth", "tree_method", "dsplit",
+    "seed", "silent", "num_round", "gamma", "min_child_weight"};
```
trivialfis: How many of these are new?

CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ option(GOOGLE_TEST "Build google tests" OFF)
 option(USE_DMLC_GTEST "Use google tests bundled with dmlc-core submodule (EXPERIMENTAL)" OFF)
 option(USE_NVTX "Build with cuda profiling annotations. Developers only." OFF)
 set(NVTX_HEADER_DIR "" CACHE PATH "Path to the stand-alone nvtx header")
+option(RABIT_MOCK "Build rabit with mock" OFF)
```
trivialfis: We only do this on windows right?

src/learner.cc
```diff
@@ -601,7 +615,7 @@ class LearnerImpl : public Learner {
       num_feature = std::max(num_feature, static_cast<unsigned>(num_col));
     }
     // run allreduce on num_feature to find the maximum value
-    rabit::Allreduce<rabit::op::Max>(&num_feature, 1);
+    rabit::Allreduce<rabit::op::Max>(&num_feature, 1, nullptr, nullptr, true);
```
trivialfis: That;s really hacky ....

src/learner.cc
```diff
@@ -332,8 +339,15 @@ class LearnerImpl : public Learner {
       }
     }
     {
-      // Write `predictor`, `gpu_id` parameters as extra attributes
-      for (const auto& key : std::vector<std::string>{"predictor", "gpu_id"}) {
+      std::vector<std::string> saved_params{"predictor", "n_gpus", "gpu_id"};
+      // check if rabit_bootstrap_cache were set to non zero before adding to checkpoint
+      if (cfg_.find("rabit_bootstrap_cache") != cfg_.end() &&
```
trivialfis: Have you tested backward compatibility?

src/learner.cc
chenqin: legacy code.

src/learner.cc
chenqin: all. those are missing configs to rerun failed rabit worker. missing these will break recovery.

CMakeLists.txt
chenqin: we don't do it anywehre except testing.

src/learner.cc
chenqin: if we don't explicit set rabit_bootstrap_cache to 1, yes, it was tested read/write old format with new code path.

If we set rabit_bootstrap_cache to 1, and read old format or rabit_bootstrap_cache=0, yes, it passed all tests except failed xgb worker recovery. That is because additional parameters were empty exactly same format as we have before this pr.

In both cases, it can load old checkpoint format and pass tests.

src/learner.cc
chenqin: This is a trade off we made to allow adhoc/one time allreduce sync. We might optimize this out after we unify "adhoc/one-time" operations with "inloop" operations with help from caller signature. But this is not a trivial task to decide which operation result should be evict verses which should be kept. One thing at a time.
 
detail design here
https://github.com/dmlc/xgboost/files/3482936/Rabit.bootstrap.operations.support.pdf

pr
https://github.com/dmlc/rabit/pull/98

chenqin:
> @chenqin Is this currently critical? I wrote the CMake file before, the reason I build rabit in XGBoost is because it fails to build on Windows. Would it be inconvenient for you if I take some time to open a PR for rabit cmake build file?

https://github.com/dmlc/rabit/blob/master/CMakeLists.txt

src/learner.cc
trivialfis: @chenqin But it's not in the master branch now?

src/learner.cc
chenqin: update, I manage to get rid of this, after merge to master api will be same as before and user don't need to worry about cache or non cache
```
rabit::Allreduce<rabit::op::Max>(&num_feature, 1);
```
https://github.com/dmlc/rabit/pull/102/commits/69789ec4510d019d25127c62973acef10d10c6fc

src/learner.cc
chenqin: removed.

src/learner.cc
chenqin: new as stored in checkpoint? all of those.
new as part of configuration? none.
checkpoint load overwrite configuration read so easiest way to fix this is to make checkpoint self contained.

chenqin:
can we rerun tests, not sure why ci point to old commit instead of 884736d
https://travis-ci.org/chenqin/xgboost/builds/581732104

chenqin:
@trivialfis

trivialfis:
@chenqin Sorry for the late reply.  Will review this once @CodingCat approves https://github.com/dmlc/rabit/pull/102 .

CodingCat:
> I did tests of commenting extra config in learner, it pass failed jvm tests on model save/load . So looks like jvm saved model were impacted with extra configs. @CodingCat F.Y.I
> std::setstd::string saved_configs_ = {};
> //{"max_depth", "tree_method", "dsplit",
> // "seed", "silent", "gamma", "min_child_weight"};
> 
> update, this is actually xgboost4j not xgboost4j-spark, I can help clean up in upcoming jvm focused pr.

not true....nothing to do with jvm-layer, offline conversation finished

chenqin:
@trivialfis :)

src/common/hist_util.cc
```diff
@@ -303,7 +303,9 @@ void DenseCuts::Init
   }
   CHECK_EQ(summary_array.size(), in_sketchs->size());
   size_t nbytes = WXQSketch::SummaryContainer::CalcMemCost(max_num_bins * kFactor);
-  sreducer.Allreduce(dmlc::BeginPtr(summary_array), nbytes, summary_array.size());
+  // TODO(chenqin): after remove is_bootstrap, hist break rabit assumeption
```
trivialfis: So what's the thing needs to be done in this todo?

src/common/hist_util.cc
```diff
@@ -303,7 +303,9 @@ void DenseCuts::Init
...
+  sreducer.Allreduce(dmlc::BeginPtr(summary_array), nbytes, summary_array.size(),
+    nullptr, nullptr, "DenseCuts");
```
trivialfis: Just a suggestion for future.  We can count the number of rows for each column, sync the maximum one in all workers.  Then we can remove this "DenseCuts" and use sparse version instead.

src/common/quantile.h
```diff
@@ -651,6 +651,9 @@ class QuantileSketchTemplate {
      * \param src the source summary
      * \param max_nbyte maximum number of byte allowed in here
      */
+#ifndef _WIN32
+    #pragma omp declare simd uniform(src, max_nbyte)
```
trivialfis: Why is this windows specific?

src/learner.cc
```diff
@@ -648,6 +662,10 @@ class LearnerImpl : public Learner {
   std::vector<std::shared_ptr<DMatrix> > cache_;
 
   common::Monitor monitor_;
+
+  /*! \brief saved config keys used to restore failed worker */
+  std::set<std::string> saved_configs_ = {"max_depth", "tree_method", "dsplit",
```
trivialfis: The IO is driving me crazy ...  Do you think it's possible to modify rabit serializable object from `Load(dmlc::Stream*)` to `Load(FixedLengthBuffer )`?  It's a little bit ugly to use stream with Json.  And besides, I don't think it's possible to check point a stream with unlimited size.

tests/cli/runxgb.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+#run make in rabit/test to generate librabit_mock
+#update config.mk and build xgboost using mock
+export DMLC_SUBMIT_CLUSTER=local
+
+submit="python3 ../../dmlc-core/tracker/dmlc-submit"
```
trivialfis: Maybe in the future we can merge this into our interface.

src/common/hist_util.cc
chenqin: per offline convo with @CodingCat, we need to figure out a way to move this one time sync before loadcheckpoint. 
image
https://github.com/dmlc/rabit/blob/master/README.md

More detailed reason is we removed is_boostrap flag and use checkpoint load as flag to differentiate cache v.s seqno based allreduce results recovery https://github.com/dmlc/rabit/commit/9a7ac85d7eb65b1a0b904e1fa8d5a01b910adda4

src/common/hist_util.cc
chenqin: I agree, I recall there was a bug raised before due to testing dataset have different column size as training dataset.

src/common/quantile.h
chenqin: in short visual studio don't support omp 4.5, this is legacy code. Will remove this in favor of compiler hinted approach. 

cc  @SmirnovEgorRu

chenqin: removed.

src/learner.cc
chenqin: as long as we make change to this interface and add new interface. I recall what they do is to write to a file and expose stream handler.
https://github.com/dmlc/dmlc-core/blob/master/include/dmlc/io.h#L132

tests/cli/runxgb.sh
chenqin: sounds good. can you specify which interface you mentioned.

src/common/hist_util.cc
chenqin: sounds like this should be a follow up pr

src/learner.cc
chenqin: It's a bit more work, so what happens because of nested call to save, a stream handler were called by various of classes and append payload onto stream and flush to disk. Load is just reverse of what save does.

chenqin: I see your point, in order to decode json, we need to load all into memory first...

chenqin: following with separate pr and link to https://github.com/dmlc/xgboost/issues/3980

chenqin: https://github.com/dmlc/xgboost/issues/4852

chenqin:
Addressed feedbacks
- remove omp declarative
- explained TODO, follow up would move histogram init before loadcheckpoint.
- explained allreduce in DMatrix::Load, follow up with fix of remove duplicated colmn size check in train/eval datasets

thvasilo:
I'll add here my understanding of the original PR in rabit, hopefully it helps other maintainers with reviewing and understanding the changes. @CodingCat and @chenqin can correct any mistakes in my explanation.

After going through the design doc, my understanding of the purpose of [dlmc/rabit#98](https://github.com/dmlc/rabit/pull/98) is allowing for single worker recovery vs. the current fail-all recovery (explained below). The doc also mentions handling cases where workers fail before the first checkpoint, during bootstrap actions like getting the number of features which is done through an allreduce at the beginning of training or broadcasting the column sampler seed value. It's not 100% clear to me why these are special cases, it's hard to grok the doc for that section (and I'm jet lagged).

If my understanding is correct, in the current implementation, in case a failure happens all workers are shut down and restarted ("fail-all" in the doc), and learning resumes from the last checkpoint. This involves requesting resources from the scheduler (e.g. Yarn) and shuffling all the data again from scratch. 
If we have `W` workers and `F` failures happen (where a failure is defined as at least one worker failed during a distinct iteration) , we would need to request `W*(F+1)` new instances: W workers to start the job, then W workers again after each failure.
Since all workers, including healthy ones are shut down in case of a failure, we need to transfer the data again to the new workers. In case of W workers and F failures that would mean shuffling data `W*F` times, according to the doc. 

Putting my PhD hat on, I'll note here that I don't fully agree with this definition of a shuffle in the doc, which treats each partition (i.e. worker) as a "shuffle". A "shuffle" (in Spark terms) is a single distribution of the data amongst all workers in the cluster. Each distinct failure currently requires one shuffle, so I'd rather say that F failures require F shuffles, rather than F*W. However we need the per worker granularity later, so I'd say that currently each failure requires us to send `W` _data partitions_ over the network.

In any case, when we have massive datasets and hundreds of workers, both these operations, requesting resources and shuffling the data across the network can be very costly and block training for extended periods of time, especially in multi-tenant clusters.

The proposed solution then is not to kill all workers and start training from the last checkpoint, but rather do a single node recovery: when a node fails, only that one is restarted, the rest of the cluster waits for it to be bootstrapped, and then continues learning.
Compared to the existing approach, a failure would only mean requesting `f` new instances from the scheduler, where `f` is the number of failed workers in the current iteration. Accordingly we would only need to send `f` data partitions over the network (one for each failed worker), which can be highly beneficial when `f << W`.

The doc continues to explain how the recovery is handled, but I haven't gone through that part in detail. Hope this helps explain the purpose of the PR!

chenqin:
Appreciate your help on explain things in much detail way! 

I agree the wording may needs more thoughts.  F*W assume every failure cause entire job retry hence lead to data full reshuffle (before loadcheckpoint) and we have M failures one at a time in job life cycle. In real life, those dataset may not saved in HDFS instead generated from various table join and feature extraction stages.  

In this example, along with techniques of external shuffle service as well as determined partitioned scheme. We are moving towards limited impact of single failure towards less than full shuffle. So yes, the comparison is lower-bound of estimation where ideally just move 1/M dataset. Worst case should be same as current approach, when cluster lost track of critical mass of datasets so that it needs to redo everything from beginning.


> Putting my PhD hat on, I'll note here that I don't fully agree with this definition of a shuffle in the doc, which treats each partition (i.e. worker) as a "shuffle". A "shuffle" (in Spark terms) is a single distribution of the data amongst all workers in the cluster. Each distinct failure currently requires one shuffle, so I'd rather say that F failures require F shuffles, rather than F*W. However we need the per worker granularity later, so I'd say that currently each failure requires us to send `W` _data partitions_ over the network.

chenqin:
Can we rerun jvm tests, seems flaky

trivialfis:
Restarted.  Will merge once we have tests passed.

trivialfis:
@chenqin Merged, big thanks!

PR #4810, opened 2019-08-26
further cleanup of single process multi-GPU code

rongou:
Address review feedback in #4773 


trivialfis:
Will look into soon

rongou:
@trivialfis looks like calling `SetParam()` would trigger a reconfiguration of the learner, so in theory setting a different `gpu_id` should work, although I don't know if we have a test for it.

trivialfis:
@rongou Thanks!

PR #4816, opened 2019-08-28
Update cudf integration based on cudf 0.9 __cuda_array_interface__

kkraus14:
Note: This will change the `__cuda_array_interface__` of the mask as the typestr will be `<t1` and the size will be the number of bits as opposed to the number of bytes. I'm unsure of the downstream changes that may be needed to handle this.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -236,27 +236,15 @@ def _extract_interface_from_cudf(df, is_info):
                          'refer to specific API.')
 
     def get_interface(obj):
-        return obj.mem.__cuda_array_interface__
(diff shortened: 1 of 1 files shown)


RAMitchell:
So we will require cudf >= 0.9?

kkraus14:
> So we will require cudf >= 0.9?

Yes, this will only work with cudf >= 0.9.

trivialfis:
Thanks.  Let me try it and change c++ layer in a few days.

kkraus14:
> Thanks. Let me try it and change c++ layer in a few days.

Let me know if I can be of any help or I can answer any questions / concerns.

PR #4656, opened 2019-07-11
[jvm-packages] initial pyspark api (WIP)

thesuperzapper:
This is a cleaned up version of the PySpark API, which previously had the PR #3376.

**There are numerous improvements:**
* You no longer need to use a special Pipeline object, the default spark ones work.
* Parameter getters/setters are dynamically generated from the parameters present in the Scala object.
* I have moved the code under `xgboost4j-spark/src/main/resources` which means the python code will end up in the .jar file
  * This means users can use the jar as if it were a python package zip.
  * This means that --packages should make the python library available automatically.
* Internal code for saving/loading is more sensible.

**Here is an example:** (Don't forget to spark-submit the `xgboost4j.jar` and `xgboost4j-spark.jar`)
```
# Initialise a SparkSession
from pyspark.sql import SparkSession
spark = SparkSession.builder.getOrCreate()

# Add the jar file (as if it were a Python zip)
spark.sparkContext.addPyFile("/PATH/TO/JAR/xgboost4j-spark-0.90.jar")

# Import Objects 
from sparkxgb import XGBoostClassifier, XGBoostClassificationModel

# Read some data
dataPath = "/PATH/TO/SPARK_HOME/data/mllib/sample_binary_classification_data.txt"
data = spark.read.format("libsvm").option("vectorType", "dense").load(dataPath)
dataSplit = data.randomSplit([0.8, 0.2], seed = 1000)
dataTrain = dataSplit[0]
dataTest = dataSplit[1]

# Define the model
paramMap = {
    "eta": 0.1,
    "maxDepth": 2,
    "objective": "binary:logistic",
    "numRound": 5,
    "numWorkers": 2
}
xgbClassifier = XGBoostClassifier(**paramMap) \
    .setFeaturesCol("features") \
    .setLabelCol("label")

# Write/Read object
xgbClassifier.write().overwrite().save("xgboost_python")
xgbClassifier_new = XGBoostClassifier().load("xgboost_python")

# Train the model
xgboostModel = xgbClassifier_new.fit(dataTrain)

# Display model parameters
for param, value in xgboostModel.extractParamMap().items():
    print((param.name, value))

# Write/Read model
xgboostModel.write().overwrite().save("xgboost_python.model")
xgboostModel_new = XGBoostClassificationModel().load("xgboost_python.model")

# Transform some data
xgboostModel_new.transform(dataTest).show()
```

**Still to do:**
* Setup PyPi to distribute (For those who want to PIP install rather than use the .jar)
[...]

PR #4819, opened 2019-08-30
Rewrite Dask interface.

### LINKED ISSUE #4814: Making sure to have right mapping between predictors and output for dask interface.
Dask makes small partitions for data objects.  Currently XGBoost collects those partitions into local workers based on `rabit.get_rank()`.  But I am not sure that's the correct way of collecting partitions.  It's possible that we end up collecting different portions of whole dataset.

Quoting from the document of [internal design](https://docs.dask.org/en/latest/dataframe-design.html#partitions)
```
When our index is sorted and we know the values of the divisions of our partitions, then we can be clever and efficient with expensive algorithms (e.g. groupby’s, joins, etc…).
```

Say for example if I loaded a dataset, split it up into predictors and labels and do some processing to each of them separately.  It's not promised that they will have the same number of total partitions, nor the order of partitions stays the same.

@RAMitchell

trivialfis:
This PR rewrites the dask interface based on `dask-xgboost`, with added support for evaluation metrics and slightly different interface.

* Implement functional train and predict.
* Implement Scikit-Learn wrapper for regression and classification.  No ranking yet.
* Implement DaskDMatrix as a proxy for underlying distributed DMatrix.

Closes #4814.


trivialfis:
@RAMitchell @hcho3 @CodingCat  I can further abstract `DMatrix` to accept `dask.dataframe`, but with considerably more changes.  WDYT?

trivialfis:
I would like to see how many works are in common with https://github.com/dmlc/xgboost/pull/4656 .  @thesuperzapper Could you also join the review after this PR no longer being WIP?

RAMitchell:
> @RAMitchell @hcho3 @CodingCat I can further abstract DMatrix to accept dask.dataframe, but with considerably more changes. WDYT?

No rush on this I think. Lets stabilise the dask interface first.

Whats your plan for assigning gpu_id in this interface?

trivialfis:
@RAMitchell I have another branch that uses a utility I implemented.  I need to somehow merge it here.

python-package/xgboost/distributed.py
```diff
@@ -0,0 +1,369 @@
...
+    def __enter__(self):
+        rabit.init(self.args)
+        logging.debug('-------------- rabit say hello ------------------')
+
+    def __exit__(self, *args):
+        rabit.finalize()
+        logging.debug('--------------- rabit say bye ------------------')
+
+
+def concat(L):
```
mt-jones: Instead of dispatching concatenation through `isinstance` checks, I recommend using `dask.dataframe.multi.concat`, here. This way Dask handles the types checks.

[Here's a link to the documentation](https://docs.dask.org/en/latest/dataframe-api.html?highlight=concat#dask.dataframe.multi.concat).

python-package/xgboost/distributed.py
```diff
@@ -0,0 +1,369 @@
...
+        return dmatrix
+
+    def num_row(self):
+        return self.n_rows
+
+    def num_col(self):
+        return self.n_cols
+
+
+def train(worker_train,
```
mt-jones: Recommend adding a keyword argument for `client`, here. Like this:
```suggestion
def train(client=get_client(),
          worker_train,
```
This way, `get_client()` is the default client; otherwise, it can be overridden by the user.

python-package/xgboost/distributed.py
```diff
@@ -0,0 +1,369 @@
...
+    '''Train XGBoost model.
+
+    worker_train: Should be xgboost.train.
+    '''
+    if platform.system() == 'Windows':
+        logging.warning(
+            'Windows is not officially supported for dask/xgboost,' +
+            ' contribution are welcomed.')
+
+    client: Client = get_client()
```
mt-jones: This syntax is experimental in Python 3.6. I would recommend avoiding it, for now.
```suggestion

```

python-package/xgboost/distributed.py
mt-jones: If you do this, then you can add a check `if client == None` based on previous import statements.

mrocklin:
@RAMitchell asked me for my thoughts on the use of the `distributed_dispatch` decorator here.  

Strictly from a Dask user's perspective I think that it's nice to have first class support for Dask within XGBoost.  However, I can see how from an XGBoost maintainer's point of view, this might be concerning.

In general I think that API extensibility is good.  For example it's nice that I can use `np.sum` on a numpy array, cupy array, dask array, and so on.  However usually when making an API extensible it's common to establish a contract that anyone can implement, without having to directly modify code within the core project.  For example maybe `xgb.train` looks for a `.__xgb_train__` method on the passed-in arguments and calls that.   That way, if some other project comes along and implements things correctly then they can be first class as well, even if they don't have strong ties to the XGBoost maintainers.  Protocols like this tend to level the playing field.

That's all maybe a bit philosophical.  I'm happy to get more into details here if folks want.

mrocklin:
I'll also say, it would be nice to find protocols that would allow us to pass in Dask dataframes or Dask arrays directly, without creating the `DaskDMatrix` object.  I'm not sure how easy that would be.  Maybe using something like `functools.singledispatch` or `multipledispatch`?

python-package/xgboost/distributed.py
```diff
@@ -0,0 +1,369 @@
...
+                 weight=None,
+                 feature_names=None,
+                 feature_types=None):
+        self._feature_names = feature_names
+        self._feature_types = feature_types
+        self._missing = missing
+
+        assert len(X.shape) == 2, ('Expecting 2 dimensions input, got: ' +
+                                   str(X.shape))
+        self.n_rows = X.shape[0]
```
mrocklin: Note that dask dataframes don't know their number of rows without computing, which can be expensive.  Is this strictly necessary for XGBoost?

python-package/xgboost/distributed.py
```diff
@@ -0,0 +1,369 @@
...
+        self.n_cols = X.shape[1]
+
+        self.X = X
+        self.y = y
+        self.weights = weight
+
+        assert type(X) in (DDF, DA)
+        assert type(y) in (DDF, DS, DA)
```
mrocklin: I had to look these up in the imports above.  I wonder if it might be clearer to use `da.Array` and `dd.DataFrame`.  These are longer, yes, but they're also terms that will be well understood by anyone who uses Dask.

python-package/xgboost/distributed.py
```diff
@@ -0,0 +1,369 @@
...
+
+        client = get_client()
+
+        if isinstance(X, DDF):
+            client.sync(self.from_dataframe)
+        elif isinstance(X, DA):
+            client.sync(self.from_array)
```
mrocklin: Are you intending to support asynchronous usage?  

If s, does this do what you expect when you make this within an async function?

mrocklin:
The actual implementation looks like what we've all done historically.  It's not super clean but there doesn't seem to be a better solution at the moment.  I think it would be worth thinking about how to improve these sorts of workflows within Dask.  

My guess is that that's out of scope for folks here (some core Dask devs probably need to spend some serious time thinking about that).  Mostly I mention this so that you don't grow too attached to this implementation, and remain open to some day replacing it with something else.

python-package/xgboost/distributed.py
trivialfis: @mt-jones Will remove it.  I only use it for code completion during development.

python-package/xgboost/distributed.py
trivialfis: @mt-jones Maybe I need to make this `train` method as another interface.

python-package/xgboost/distributed.py
trivialfis: @mt-jones Thanks for the pointer.

python-package/xgboost/distributed.py
trivialfis: @mrocklin It's a lazy object I think.  To actually get the number of rows I need to call `compute()`.  I put it here just for verifying the rows accumulated from partitions are correct.

python-package/xgboost/distributed.py
trivialfis: @mrocklin Sure.  Will refactor it properly.

trivialfis:
> My guess is that that's out of scope for folks here 

I'm interested in getting to know more about dask.  So please keep me in the loop.

> Mostly I mention this so that you don't grow too attached to this implementation

That won't be a problem.  If there's an improvement we can change it anytime.  I don't expect one commit can get  everything right. ;-)

As for sklearn interface.  I'm working on it.

python-package/xgboost/distributed.py
trivialfis: @mrocklin 

> Are you intending to support asynchronous usage?

No.  It's more like a result of try and errors ...  Asking `who_has` returns a coroutine, also the `parts` so I called `await` ...

python-package/xgboost/distributed.py
trivialfis: @mt-jones I put `multi.concat` at the end of my type checkings since I still need to check for array instead of dataframe.  cudf should go into that route.  Also I got some weird hanging in `pytest` with `multi.concat` but can't reproduce outside of those fixtures used in tests.

trivialfis:
@mrocklin @RAMitchell @mt-jones Please take a look again.

python-package/xgboost/dask.py
```diff
@@ -1,121 +1,589 @@
...
+        msg = 'Windows is not officially supported for dask/xgboost,'
+        msg += ' contribution are welcomed.'
+        logging.warning(msg)
+
+    if 'evals_result' in kwargs.keys():
+        raise ValueError(
+            'evals_result is not supported in dask interface.',
+            'The evaluation history is returned as result of training.')
+
+    if client is None:
```
mt-jones: The code in `predict` uses `if not client`. Can we make these identical?

Also, if I call `xgb.dask.train(params, dtrain)` [*] will this trigger the `get_client()` call? Or do I need to explicitly write `xgb.dask.train(None, params, dtrain)`?

If would be good to support (*) in addition to `xgb.dask.train(client, params, dtrain)`. Should we just make this a keyword argument?

python-package/xgboost/dask.py
```diff
@@ -1,121 +1,589 @@
...
+    elif isinstance(model, dict):
+        booster = model['booster']
+    else:
+        raise TypeError(_expect([Booster, dict], type(model)))
+
+    if not isinstance(data, DaskDMatrix):
+        raise TypeError(_expect([DaskDMatrix], type(data)))
+
+    worker_map = data.worker_map
+    if not client:
```
mt-jones: Ref to previous comment for `xgb.dask.train`.

demo/dask/gpu_training.py
```diff
@@ -0,0 +1,41 @@
...
+
+    # DaskDMatrix acts like normal DMatrix, works as a proxy for local
+    # DMatrix scatter around workers.
+    dtrain = DaskDMatrix(client, X, y)
+
+    # Use train method from xgboost.dask instead of xgboost.  This
+    # distributed version of train returns a dictionary containing the
+    # resulting booster and evaluation history obtained from
+    # evaluation metrics.
+    output = xgb.dask.train(client,
```
mt-jones: This is super clean! Really like this.

python-package/xgboost/dask.py
mt-jones: This amounts to an API that behaves like this:
```
xgb.dask.train(params, dtrain)
```

or

```
xgb.dask.train(params, dtrain, client=my_client)
```

If we don't like the look of the position in the latter, I'm fine with leaving it as it is.

trivialfis: @mt-jones I tried this.  There are other arguments defined as `*args` and passed into `xgboost.train`.  I don't feel comfortable adding an extra at the middle of `*args` and `**kwargs`.  It's quite confusing to me along with duck typing ...

trivialfis: But will address other comments.

trivialfis:
@mrocklin @TomAugspurger  The original implementation of dask-xgboost is acknowledged in both doc and code header.  Thank you so much!

trivialfis:
@CodingCat @hcho3 Would you like to take a look before merging?

trivialfis:
Merging as we will have more time to test it.  @hcho3 @CodingCat Try it out, I have grown to like dask now. ;-)

trivialfis:
@mt-jones @mrocklin @RAMitchell Thanks!

PR #4824, opened 2019-09-01
Optimizations of distributed 'hist' mode for CPUs

SmirnovEgorRu:
I observed that distributed XGBoost on 1 node is slower than non-distributed (Batch) XGBoost. Difference is 1.5-3x times depending on data set and parameters. The reasons of this:
- We took always left tree node for complex histogram computation and right node - computed by subtraction trick. But I added choosing of the smallest tree node across of computation nodes.
- Poor threading for histogram reduction in distributed case - also fixed.

Performance measurements:

Data set  | higgs1m
-- | --
1 node   (before), s | 5.29
1 node (this   PR), s | 3.66
Batch mode, s | 3.53

So, we see 1.5x speedup against prev version and similar time as in Batch mode.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -102,3 +102,5 @@ List of Contributors
 * [Haoda Fu](https://github.com/fuhaoda)
 * [Evan Kepner](https://github.com/EvanKepner)
   - Evan Kepner added support for os.PathLike file paths in Python
+* [Egor Smirnov](https://github.com/SmirnovEgorRu)
+  - Egor optimized 'hist' tree method for Intel CPUs
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -455,13 +455,27 @@ void QuantileHistMaker::Builder::CreateNewNodesBatch(
     const int32_t right_id = node.RightChild();
     row_set_collection_.AddSplit(nid, n_left, left_id, right_id);
 
-    if (rabit::IsDistributed() ||
-        row_set_collection_[left_id].Size() < row_set_collection_[right_id].Size()) {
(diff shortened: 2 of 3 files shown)


CodingCat:
except optimizing, could you check https://github.com/dmlc/xgboost/pull/4716 which involves a correctness issue in the current syncing strategy

hcho3:
@CodingCat Good point, we cannot let the sync issue unaddressed before 1.0.0. I'll see if there's anyone in my org to help. (Also will try to get some time myself too.)

SmirnovEgorRu:
@CodingCat @hcho3 I have found an issue that distributed workers can build different trees and I have fixed this.

I also checked log-loss for higgs (4 local workers)- it is exactly the same for 2 runs:

SmirnovEgorRu:
@thvasilo it was done to reuse this->histred_ rabit::Reducer object and don't create new once special for size_t. So, we just compute sum of all samples in each tree node using floats instead of size_t.

thvasilo:
> @thvasilo it was done to reuse this->histred_ rabit::Reducer object and don't create new once special for size_t. So, we just compute sum of all samples in each tree node using floats instead of size_t.

I'm afraid this would confuse potential future maintainers, so a comment should be added at least.

Is it possible to use one of the existing aggregation ops (Sum) here, [like in the rabit usage example](https://github.com/dmlc/rabit/blob/9a7ac85d7eb65b1a0b904e1fa8d5a01b910adda4/guide/basic.cc#L30)? We shouldn't need a Reducer object to sum/max some integers I think.

SmirnovEgorRu:
Close since it's not relevant code. Anyway, optimizations idea can be used in farther commits.

CC @ShvetsKS

PR #4851, opened 2019-09-11
[blocking] fix parallel eval_split of hist updater

sperlingxx:
In #4716,  we modified schedule policy of OPENMP from `guided` to `static`. But I caught same problem after this fix.  
After a lot of debug work, I found the `rabit::IsDistributed` method, which is in parallel body of OPENMP,  is not thread safe.  Because [rabit::GetEngine](https://github.com/dmlc/rabit/blob/dba32d54d1668033356a2ad505c239411d660821/src/engine.cc#L74) is thread local.


sperlingxx:
@hcho3  @trivialfis  @SmirnovEgorRu  @CodingCat  Can you help to review this?

hcho3:
Thanks!

trivialfis:
Thanks.

PR #4858, opened 2019-09-15
[jvm-packages] refine numAliveCores method of SparkParallelismTracker 

sperlingxx:
This PR propose a more stable approach to query total number of alive cores; fetch executor information via spark API `sc.statusStore.executorList` directly rather than using restful API of spark WebUI.

In addition,  I found a bug about `waitForCondition`. Query thread wrapped by `Future` will never exit if condition is not satisfied.  Because `Await.ready` will throw an exception without stopping the`Future` thread when timeout is reached. 

@CodingCat can you help to look this?

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
@@ -19,10 +19,6 @@ package org.apache.spark
 import org.apache.commons.logging.LogFactory
 import org.apache.spark.scheduler._
 
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.duration._
-import scala.concurrent.{Await, Future, TimeoutException}
-
 /**
  * A tracker that ensures enough number of executor cores are alive.
  * Throws an exception when the number of alive cores is less than nWorkers.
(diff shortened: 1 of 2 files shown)


jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
```diff
@@ -43,39 +37,34 @@ class SparkParallelismTracker(
...
-        Int.MaxValue
-      }
-    } catch {
-      case ex: Throwable =>
-        logger.warn(s"Unable to read total number of alive cores from REST API." +
-          s"Health Check will be ignored.")
-        ex.printStackTrace()
-        Int.MaxValue
-    }
+<<<<<<< Updated upstream
```
hcho3: Unclean git rebase?

jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
```diff
@@ -43,39 +33,28 @@ class SparkParallelismTracker(
...
-        Int.MaxValue
-      }
-    } catch {
-      case ex: Throwable =>
-        logger.warn(s"Unable to read total number of alive cores from REST API." +
-          s"Health Check will be ignored.")
-        ex.printStackTrace()
-        Int.MaxValue
-    }
+    sc.statusStore.executorList(true). map(_.totalCores).sum
```
CodingCat: extra space?

CodingCat:
merged thanks

PR #4862, opened 2019-09-17
Fix data loading

trivialfis:
* Fix config regex.
* Try to explain the error better in exception. (not really useful since dmlc csv parser can parse libsvm file without generating any error.)
* Update doc.

### DIFF
--- a/doc/contrib/docs.rst
+++ b/doc/contrib/docs.rst
@@ -28,5 +28,3 @@ Examples
 * We are super excited to hear about your story, if you have blogposts,
   tutorials code solutions using XGBoost, please tell us and we will add
   a link in the example pages.
-
-
--- a/doc/python/python_intro.rst
+++ b/doc/python/python_intro.rst
@@ -18,6 +18,8 @@ To verify your installation, run the following in Python:
 
   import xgboost as xgb
 
+.. _python_data_interface:
+
 Data Interface
 --------------
 The XGBoost python module is able to load data from:
@@ -50,7 +52,7 @@ The data is stored in a :py:class:`DMatrix <xgboost.DMatrix>` object.
 
   .. note:: Categorical features not supported
 
-    Note that XGBoost does not support categorical features; if your data contains
+    Note that XGBoost does not provide specialization for categorical features; if your data contains
     categorical features, load it as a NumPy array first and then perform
     `one-hot encoding <http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html>`_.
 
--- a/doc/tutorials/input_format.rst
+++ b/doc/tutorials/input_format.rst
@@ -7,6 +7,9 @@ Basic Input Format
 ******************
 XGBoost currently supports two text formats for ingesting data: LibSVM and CSV. The rest of this document will describe the LibSVM format. (See `this Wikipedia article <https://en.wikipedia.org/wiki/Comma-separated_values>`_ for a description of the CSV format.)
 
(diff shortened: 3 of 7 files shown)


trivialfis:
Somehow the libsvm parser can parse a csv without generating any error ...

trivialfis:
@hcho3 Do you know the use case of optional value field in libsvm?  In the following code, it's possible for a libsvm `<feature>:<value>` pair to have only feature but no value. https://github.com/dmlc/dmlc-core/blob/808f485387f9a03f78fa9f1159f387d0d91b7a28/src/data/libsvm_parser.h#L146

Could you think of a case I might break compatibility if I enforce having a complete pair?

> Thanks for correcting my mistake :)

Just don't get mad when I create new ones ..

thvasilo:
@trivialfis AFAIK there's no formal format for libsvm files, however there is a program that checks for dataset validity [checkdata.py](https://github.com/cjlin1/libsvm/blob/master/tools/checkdata.py).

I tried a simple dataset with that

```
1 2:1 3:4
1 3
1 4:
```

where the second line does not a value, and the third has the colon but no value, and both were marked as errors.
So for "pure" libsvm files this would be wrong. 

But taking a look at the [blame](https://github.com/dmlc/dmlc-core/blame/808f485387f9a03f78fa9f1159f387d0d91b7a28/src/data/libsvm_parser.h#L146) these were added explicitly, [one](https://github.com/dmlc/dmlc-core/commit/e8824983113e809546cebdcb844db1e8db4cafb9) to support the [ranklib format](https://sourceforge.net/p/lemur/wiki/RankLib%20File%20Format/) and one to add the [weight support](https://github.com/dmlc/dmlc-core/commit/fbf75b7f410b389df20573e6bef205d1b69fed75).

So i'd be hesitant to remove until we are sure why they were added in the first place.

trivialfis:
@thvasilo Thanks for the reply.  From the links you provided, neither ranklib nor [having weights](https://xgboost.readthedocs.io/en/latest/tutorials/input_format.html#instance-weights) would allow missing value with an non-missing index.  I would like to take the risk ...

trivialfis:
Leave a note for the extended libsvm format here:
```
<line> -> <label> <QID> <feature>: <value> ... <feature>: <value> <Comment>
<label> -> <positive integer>: <float> | <positive integer>
<QID> -> qid: <positive integer> | e
<Comment> -> # <String> | e
```

trivialfis:
@hcho3 I can't iterate enough time to rewrite the parser right now.  Can I just fix this config path issue then open an issue for better documentation around our uri format?

hcho3:
@trivialfis Yes, let’s reduce the scope of this PR, to fixing path parsing in config files. Let’s return to data parsing later.

trivialfis:
@hcho3 Would you like to take another if the tests pass?

sriramch:
@trivialfis - this commit broke the build in my environment. more specifically, the assertion on ofstream added in dmatrix uri test. i have gtest 1.7 environment, which does not have this - https://github.com/google/googletest/blob/master/googletest/include/gtest/gtest.h#L293

is there a minimal gtest-dev requirement? if so, we either need to enforce this, or revert the assertion on ofstream explicitly and check explicitly if good bits are set on the stream.

hcho3:
@sriramch We require gtest 1.8.1 and higher. See https://github.com/dmlc/dmlc-core/pull/542

sriramch:
@hcho3 thanks for the clarification and the doc. pr.

PR #4866, opened 2019-09-17
ignore vscode and clion files

rongou:
Ignore files generated by IDEs (VSCode, IntelliJ, CLion, etc.).

@RAMitchell @trivialfis @CodingCat

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,10 +69,8 @@ config.mk
 /xgboost
 *.data
 build_plugin
-.idea
 recommonmark/
 tags
-*.iml
 *.class
 target
 *.swp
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -95,3 +93,11 @@ metastore_db
 # files from R-package source install
 **/config.status
 R-package/src/Makevars
+
+# Visual Studio Code
+/.vscode/
+
+# IntelliJ/CLion
+/.idea/
```
trivialfis: I don't use it, but why is an extra `/`?

rongou: That's what CLion does when you ask it to ignore a directory, it's slightly more precise, a directory at the repository root only. Changed it back in case anyone creates a project in a subdirectory.

PR #4867, opened 2019-09-17
remove device shards

rongou:
Remove `DeviceShard` classes since we no longer support single process multi-GPU.

In `updater_gpu_hist.cu`, I only renamed the `DeviceShard` to `GPUHistMakerDevice` since it's fairly self contained. Let me know if you have a better name for it.

Hopefully the last cleanup PR for #4531.

### DIFF
--- a/src/common/hist_util.cu
+++ b/src/common/hist_util.cu
@@ -27,9 +27,11 @@ namespace common {
 
 using WXQSketch = DenseCuts::WXQSketch;
 
-__global__ void FindCutsK
-(WXQSketch::Entry* __restrict__ cuts, const bst_float* __restrict__ data,
- const float* __restrict__ cum_weights, int nsamples, int ncuts) {
+__global__ void FindCutsK(WXQSketch::Entry* __restrict__ cuts,
+                          const bst_float* __restrict__ data,
+                          const float* __restrict__ cum_weights,
+                          int nsamples,
+                          int ncuts) {
   // ncuts < nsamples
   int icut = threadIdx.x + blockIdx.x * blockDim.x;
   if (icut >= ncuts) {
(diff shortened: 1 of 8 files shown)


trivialfis:
Sorry for the long wait.  Will give a proper review tomorrow.

src/common/hist_util.cu
```diff
@@ -27,9 +27,11 @@ namespace common {
```
trivialfis: @RAMitchell is doing sketching algorithm anyway, these code is becoming legacy sooner or later.  It's not very motivating to change them ...

src/linear/updater_gpu_coordinate.cu
```diff
@@ -235,29 +145,100 @@ class GPUCoordinateUpdater : public LinearUpdater {  // NOLINT
...
+        gpair_.data(),
+        host_gpair.data(),
+        gpair_.size() * sizeof(GradientPair), cudaMemcpyHostToDevice));
+  }
+
   // training parameter
   LinearTrainParam tparam_;
   CoordinateParam coord_param_;
-  int device_{};
+  int device_{-1};
```
trivialfis: We should use the `generic_parameter` in the future.

src/tree/updater_gpu_hist.cu
```diff
@@ -1076,26 +1074,26 @@ class GPUHistMakerSpecialised {
...
       const DMatrix* data, HostDeviceVector<bst_float>* p_out_preds) {
-    if (shard_ == nullptr || p_last_fmat_ == nullptr || p_last_fmat_ != data) {
+    if (maker_ == nullptr || p_last_fmat_ == nullptr || p_last_fmat_ != data) {
       return false;
     }
     monitor_.StartCuda("UpdatePredictionCache");
     p_out_preds->SetDevice(device_);
-    dh::safe_cuda(cudaSetDevice(shard_->device_id));
-    shard_->UpdatePredictionCache(p_out_preds->DevicePointer());
+    dh::safe_cuda(cudaSetDevice(maker_->device_id));
```
trivialfis: Maybe in the future we can move this into `maker_`.

src/linear/updater_gpu_coordinate.cu
rongou: Done.

src/tree/updater_gpu_hist.cu
rongou: Done.

trivialfis:
Merging as so we can have more time to test it.

PR #4872, opened 2019-09-18
Add check for length of weights.

### LINKED ISSUE #4869: sample_weight crashes the interpreter
Hi all,

today I encountered an issue with xgboost that crashed my python interpreter. I tried to  train a multi-class classifier with a sparse matrix. Without sample_weight, the process goes fine. When I set the parameter to the weights that I generated using sklearn.utils.class_weight, It crashed the interpreter.

Here's a sample code:

```
class_weights = list(class_weight.compute_class_weight('balanced', np.unique(y), y))
# X is a sparse csr_matrix
X_train, X_test, Y_train, Y_test = train_test_split(X, y, test_size=0.15, random_state=1338)

xgb = XGBClassifier(n_estimators=2000, max_depth=12,
                    objective='multi:softprob', nthread=4, 
                    seed=1338, gamma=5.5, learning_rate=0.0001,
                   colsample_bytree=0.5, subsample=0.5,
                   num_class=len(np.unique(Y_train))

xgb.fit(X_train, Y_train, sample_weight=class_weights)
```

I'm thinking it could be due to the train test split, which could remove some low-occurrence classes and then we get a size mismatch - however this should not crash the interpreter. I will update the issue as I get more information.

trivialfis:
Closes #4869 .

### DIFF
--- a/demo/guide-python/boost_from_prediction.py
+++ b/demo/guide-python/boost_from_prediction.py
@@ -1,4 +1,5 @@
 #!/usr/bin/python
+import numpy as np
 import xgboost as xgb
 
 dtrain = xgb.DMatrix('../data/agaricus.txt.train')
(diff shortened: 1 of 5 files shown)


trivialfis:
@RAMitchell ping for review.

trivialfis:
Let's make a constructor for meta info.  :-)

Issue #4707, opened 2019-07-25
parameter gamma doesn't work at all when using gpu_hist


panranyang:
I use "hist" and "gpu_hist" to train my model. When gamma=15 for both, "hist" return a model whose sum of features' fscore is 0, which means all nodes were pruned.While "gpu_hist" return a normal model. Then I use gamma=0 for "hist", the result becomes normal. Though I use gamma=1e4 for "gpu_hist", the result is still just like gamma=0.

I  just wonder if "gamma" does not work with "gpu_hist" or if I use wrong parameter when using "gpu_hist"

GPU: GTX1080
xgboost version: 0.82
GPU driver version:396.37
CUDA version: 9.2.148

RAMitchell:
It doesn't currently because it doesn't use the pruner. This can possibly be fixed.

panranyang:
Thank you!

PR #4874, opened 2019-09-19
Support gamma in GPU_Hist.

trivialfis:
@hcho3 @RAMitchell I just prevent the split when `loss_chg` < `gamma`.  Could you think of a case adding an explicit `pruner` is useful?  One additional thing I found in `pruner` is a syncing operation (broadcasting rank 0 tree), but I don't think it's necessary for `gpu_hist`.


trivialfis:
I will add tests if the method is reasonable.

RAMitchell:
I can't think of any reason why not, let's do it.

trivialfis:
@RAMitchell Please take a look again.

trivialfis:
@panranyang Fixed. #4707

Issue #4895, opened 2019-09-25
[Feature request] Python-like interface for modern C++


hcho3:
#4894 proposes to expose C++ headers to 3rd-party applications. As an extension, we should also consider building a Python-like frontend in modern C++. I'm thinking something like https://pytorch.org/cppdocs/frontend.html#end-to-end-example, where PyTorch offers a complete modern C++ API that's at feature parity with the Python package.

In particular, features like early stopping should be available from a C++ application.

trivialfis:
Yeah, I love the interface from pytorch.  Too bad the torch script is based on Python.

trivialfis:
It doesn't necessarily like Python though, we can improve upon it if needed.

trams:
Looks good! I am all for moving common code from python|java|R to C++

hcho3:
To new contributors: it would be great if we can have an experimental prototype for the C++ frontend. For the first pull request, it is okay to have only basic features (load data, train). Post comment here you'd like to work on this feature, so that we can discuss a possible design. I am available for help.

hcho3:
I'd like to get to this issue, hopefully before end of this year.

PR #4897, opened 2019-09-26
Export c++ headers in CMake installation.

### LINKED ISSUE #4893: Refactor to eliminate relative include paths into source dir
There are a lot of places in the code where the following pattern of including paths outside the current dirs directory tree is used.
https://github.com/dmlc/xgboost/blob/82ee2317e80f93465685529e1ab9acff52b683c7/include/xgboost/data.h#L22-L24
https://github.com/dmlc/xgboost/blob/82ee2317e80f93465685529e1ab9acff52b683c7/include/xgboost/gbm.h#L24
I think this is very bad practice and I don’t understand the rational behind this. 
E.g. if you think `span.h` is something considered a private implementation detail than it should get included only by *.cc  files within the src/ tree. 
If you ever need to include this into some of the headers in `include/` it’s probably not a private implementation detail anymore but something you need to share with who ever is using includes from the `include/` dir. In that case `span.h` should also live in `include/`.

This anti-pattern repeats throughout a many sources, not only in xgboost, but also in rabit.

### LINKED ISSUE #4894: [RFC] Expose C++ headers as part of the CMake target
For more context, see discussion at https://discuss.xgboost.ai/t/is-there-a-c-api-document-and-usage-examples/592/10.

@codingforfun suggested that the C++ headers in XGBoost should be exposed to 3rd-party applications via the CMake target `xgboost`. I initially objected because future refactoring would change C++ interfaces, but @codingforfun replied that API stability is not as important in his use case:

> Also I don’t think your concerns about API stability really matter from the perspective of a project directly accessing the C++ interface.
> * ABI stability – I think this is the number one reason for C-Interfaces. You don’t have that anyway when using C++.  [My note: it means that C++ users will re-compile XGBoost periodically.]
> * **A lot of stuff that can be done in C++ easily is PITA or impossible using the C-Interface. This outweights the need to refactoring for interface changes.** [emphasis mine] Example: Currently we
>    - Derive from DataSource to feed data in memory into XGB
>    - Derive from ObjFunction to define our own objective
>    - Derive from Metric to define our own metric
>    - Derive from dmlc::Stream to get the trained classifier into QByteArray

I'm now persuaded that the C++ headers in XGBoost should be exposed as part of the CMake target, on the condition that we clearly document future possibility of API changes. ("We do not make any guarantee about API stability, so your application may not work with future XGBoost releases".) In our experience, we tend to be less conservative when it comes to changing C++ interfaces, compared to changing C API.

cc @trivialfis @RAMitchell @chenqin @CodingCat @thvasilo Let me know what you think.

trivialfis:
Closes #4894. Closes #4893.

Some notes:
- This doesn't mean we have a c++ API,  they are just headers.
- Unlike c API, which is well abstracted.  Those c++ headers have `dmlc` dependency.  User needs to link and include `dmlc` manually.
- There are forward declarations in headers.  That means there are things one cannot do even with the headers.
- On the other hand, this PR cleans up many header include paths, it's been a headache for quite a long time now.  Since we will have to do the refactoring anyway, doing it earlier so we can avoid adding more git history to those moved files.

All in all, these headers don't form an API and might only be accessible to to someone who code for fun, like @codingforfun ;-).  But never the less it's a first step.  Thanks @hcho3 @codingforfun


src/common/bitfield.h
```diff
@@ -13,7 +13,12 @@
 #include <string>
 #include <vector>
 
-#include "span.h"
+#if defined(__CUDACC__)
+#include <thrust/copy.h>
+#include <thrust/device_ptr.h>
+#endif  // defined(__CUDACC__)
+
+#include "xgboost/common/span.h"
```
hcho3: Sometimes you use `<>` for XGBoost headers and sometimes you use `""`. Any rationale for this?

hcho3:
I really need to find time to stabilize tests...

hcho3:
Also, we should put a warning about possibility of future change in the C++ headers.

trivialfis:
Or try to reduce the memory usage in hist :-), I think there are plenty of room for that.

Yep, will add a short note in doc.

src/common/bitfield.h
trivialfis: @hcho3 I use `""` in internal code and `<>` in public headers.  The difference becomes important when you export an interface.  With proper configuration in CMake, headers in system path are included with compiler flag `-isystem` instead of `-I`.  Using `""` will fail.

hcho3: Interesting. So `<>` will cause the compiler to look into system paths.

codingforfun:
>     * This doesn't mean we have a c++ API,  they are just headers.
Well, headers are an API. Maybe not the one you wished to have, but they are. I still don’t really get why this feels so bad for you guys. For me this feels like a somewhat artificial distinction. 

>     * Unlike c API, which is well abstracted.  Those c++ headers have `dmlc` dependency.  User needs to link and include `dmlc` manually.
That is an interesting point. Currently I just patch CMakeLists of XGB to locally install dmlc API ;-) along XGB. 
Do you see DMLC as something independent that may be installed separately on a system in any real world scenario or is it merely a building block of XGB. In the latter case I would suggest to either check if it can be avoided in the XGB interface or just install it along XGB headers. 

>     * There are forward declarations in headers.  That means there are things one cannot do even with the headers.
Have to check this, but with v0.80 we had no problems. (After patching access to install for installing all your headers.)

> All in all, these headers don't form an API and might only be accessible to to someone who code for fun, like @codingforfun ;-). But never the less it's a first step. Thanks @hcho3 @codingforfun
The name doesn’t mean we are not doing serious programming. Actually we are using this for real world problems ;-)

To summaries: My whole point here is: There is a lot of interesting things to do with the C++ library of XGB and it’s already all there so in my opinion there is no point in trying to suppress usage of it. Every interface is an interface. Maybe not the best possible one, but you can’t claim you don’t have one and trying to keep it secrets just creates pain on your side and for your users.

trivialfis:
> Well, headers are an API. Maybe not the one you wished to have, but they are. I still don’t really get why this feels so bad for you guys. For me this feels like a somewhat artificial distinction.

An API sure, an user interface, not really.  It needs to be tested, have demos, documents, release version control, reasonably stable etc ...  We currently don't have the resource to do all that.  For example, I'm trying to revise our IO logic to better handle distributed setting.  The c++ headers might got a some changes (enforcing a new IO functions with pure functions) while C/Python/R/Scalar/Java/Julia/Ruby API stay intact.

> Do you see DMLC as something independent that may be installed separately on a system in any real world scenario or is it merely a building block of XGB. In the latter case I would suggest to either check if it can be avoided in the XGB interface or just install it along XGB headers.

It's a "common directory" for dmlc projects, another big user is mxnet.  With C API we hide it behind the scene,  but c++ headers use it extensively.  To give you an example, in latest release and master branch, you can use XGBoost as a CMake target like `find_package(xgboost) target_link_library(awesome xgboost::xgboost)`. `dmlc/openmp/cuda` are not directly linked into your binary so you don't even need to think about them. Only `libxgboost.so` will appear in your `ldd libawesome.so`. But this limits to C API.  There's a demo in `xgboost/demo/c-api` .

> The name doesn’t mean we are not doing serious programming. Actually we are using this for real world problems

More fun. ;-)

> Every interface is an interface.

Since we have a plugin system (see `xgboost/plugin/example`), you can consider using it, not stable though.

trivialfis:
@hcho3 Added some notes, please review.

hcho3:
@codingforfun As noted in the doc, we currently do not promise any kind of stability from version to version when it comes to the C++ header interface. However, if we ever come around implementing a C++ "frontend" in #4895, then we will designate this subset of C++ interface as more stable.

doc/c.rst
```diff
@@ -0,0 +1,9 @@
+#################
+XGBoost C Package
+#################
+
+XGBoost implements a set of C API designed for various bindings, we maintain the its
```
codingforfun: small typo. remove the _the_

doc/c++.rst
```diff
@@ -0,0 +1,9 @@
+###############
+XGBoost C++ API
+###############
+
+We installs all c++ headers during ``CMake`` build installation after 1.0 release.  But
```
hcho3:
```suggestion
Starting from 1.0 release, CMake will generate installation rules to export all C++ headers. But
```

doc/c++.rst
```diff
@@ -0,0 +1,9 @@
...
+the c++ interface is much closer to the internal of XGBoost than other language bindings.
+As a result it's changing quite often and we don't maintain its stability.  Along with the
+plugin system (see ``plugin/example`` in XGBoost's source tree), users can utilize some
+existing c++ headers for gaining more access to the internal of XGBoost.
```
hcho3:
```suggestion
existing c++ headers for gaining more access to the internal of XGBoost.

* `C++ interface documentation (latest master branch) <https://xgboost.readthedocs.io/en/latest/dev/files.html>`_
* `C++ interface documentation (last stable release) <https://xgboost.readthedocs.io/en/stable/dev/files.html>`_
```

doc/c.rst
```diff
@@ -0,0 +1,9 @@
...
+stability and the CMake/make build interface.  See ``demo/c-api/README.md`` for an
+overview and related examples.  Also one can generate doxygen document by providing
+``-DBUILD_C_DOC=ON`` as parameter to ``CMake`` during build, or simply look at function
+comments in ``include/xgboost/c_api.h``.
```
hcho3:
```suggestion
comments in ``include/xgboost/c_api.h``.

* `C API documentation (latest master branch) <https://xgboost.readthedocs.io/en/latest/dev/c__api_8h.html>`_
* `C API documentation (last stable release) <https://xgboost.readthedocs.io/en/stable/dev/c__api_8h.html>`_
```

doc/c.rst
trivialfis:
```suggestion
XGBoost implements a set of C API designed for various bindings, we maintain the its
```

trivialfis:
@RAMitchell I removed the `tbegin` from `HostDeviceVector` and put it into `device_helpers.cuh` to avoid specialized macros.   It's now `dh::tbegin(vector)`, similar to `std::begin`.

trivialfis:
Don't merge right now.  Will do some re-organization tomorrow.

hcho3:
@trivialfis Can we merge this for now?

trivialfis:
Just want to remove the common subdirectory in headers

trivialfis:
@hcho3 Done.  Sorry for the long wait.

trivialfis:
ping @RAMitchell

PR #4902, opened 2019-09-27
Support feature names/types for cudf.

### LINKED ISSUE #4901: [BUG] `DMatrix` loosing feature names when created with `cudf`.
#### `DMatrix` looses feature names when created with `cudf`

We seem to loose feature names when creating dmatrix from a cudf dataframe.  

This can lead to silent as the order of columns might change. 

#### Example:

```python
import xgboost as xgb
import cudf
import pandas as pd
```

##### Cudf output:

```
xgb.DMatrix(cudf.DataFrame({'x':[1,2,3]}),cudf.DataFrame({'y':[1,2,3]})).feature_names
```
```
['f0']
```

##### Pandas Output
```
xgb.DMatrix(pd.DataFrame({'x':[1,2,3]}),pd.DataFrame({'y':[1,2,3]})).feature_names
```

```
['x']
```

#### Additional Context: 

We can probably modify the `pandas` class that extracts the feature names to work with `cudf` too.

https://github.com/dmlc/xgboost/blob/2fa8b359e06783ed5824824ec8ffb32c1d4e1383/python-package/xgboost/core.py#L289-L292

trivialfis:
@VibhuJawa I added a test, could you please review? Closes  #4901

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -17,10 +17,11 @@ import json
 import numpy as np
 import scipy.sparse
 
-from .compat import (STRING_TYPES, PY3, DataFrame, MultiIndex, py_str,
-                     PANDAS_INSTALLED, DataTable,
-                     CUDF_INSTALLED, CUDF_DataFrame, CUDF_Series,
-                     os_fspath, os_PathLike)
+from .compat import (
+    STRING_TYPES, PY3, DataFrame, MultiIndex, py_str,
+    PANDAS_INSTALLED, DataTable,
+    CUDF_INSTALLED, CUDF_DataFrame, CUDF_Series, CUDF_MultiIndex,
+    os_fspath, os_PathLike)
 from .libpath import find_lib_path
 
 
@@ -243,8 +244,11 @@ def c_array(ctype, values):
 
 
 def _use_columnar_initializer(data):
-    '''Whether should we use columnar format initializer (pass data in as
-    json string).  Currently cudf is the only valid option.'''
+    '''Whether should we use columnar format initializer (pass data in as json
+    string).  Currently cudf is the only valid option.  For other dataframe
+    types, use their sepcific API instead.
+
+    '''
     if CUDF_INSTALLED and (isinstance(data, (CUDF_DataFrame, CUDF_Series))):
         return True
     return False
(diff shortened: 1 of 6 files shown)


python-package/xgboost/core.py
```diff
@@ -445,13 +481,9 @@ def __init__(self, data, label=None, missing=None,
...
-                                                                feature_types)
-
-        data, feature_names, feature_types = _maybe_dt_data(data,
-                                                            feature_names,
-                                                            feature_types)
+        data, feature_names, feature_types = _convert_dataframes(
+            data, feature_names, feature_types
+        )
 
         label = _maybe_pandas_label(label)
```
VibhuJawa: We might need to implement the `cudf` equivalent for  below  too as when it does not raise a  exception when i provide it with a  label `dataframe` with multiple columns .

```
ValueError: DataFrame for label cannot have multiple columns
```

https://github.com/dmlc/xgboost/blob/4be4484f966096b3a8a3df277dbb07aa1bcce8d1/python-package/xgboost/core.py#L322-L338

python-package/xgboost/core.py
```diff
@@ -337,6 +338,24 @@ def _maybe_pandas_label(label):
     return label
 
 
+def _maybe_cudf_dataframe(data, feature_names, feature_types):
+    if not (CUDF_INSTALLED and isinstance(data,
+                                          (CUDF_DataFrame, CUDF_Series))):
+        return data, feature_names, feature_types
+    if feature_names is None:
+        if isinstance(data, CUDF_Series):
```
VibhuJawa: We may want to extend this function to support multi-indexes as currently pandas does. (So some porting their pandas code to cudf might face problems . )

We can approach this later but still might be good to throw a error stating that multi-index is not supported.

trivialfis: @VibhuJawa Let me get it done in this PR.

python-package/xgboost/core.py
trivialfis: @VibhuJawa There will be an error in c++ level.  The c++ implementation should handle much as possible so that we don't have too much duplicated code across language bindings.

trivialfis:
@VibhuJawa Done.  Let me know your thoughts.

VibhuJawa:
> @VibhuJawa Done. Let me know your thoughts.

Looks good to go for me. Thanks a ton for the fix.

trivialfis:
@VibhuJawa Thanks for the review.

PR #4687, opened 2019-07-20
[jvm-packages] Move to SBT

trams:
I took an abandoned pull request https://github.com/dmlc/xgboost/pull/2767 and adapted to current master (cc @superbobry)

Checklist:
1. JNI compilation
2. Tests
3. Scalastyle and Checkstyle
4. Cross-compilation wrt Scala version
5. Uber JAR

I welcome your suggestions.

One side effect of this pull request I am very interested in is an ability to build against Scala 2.11
I think it would be nice to support Scala 2.11 as long as our dependencies support it

What do you think?

PR #4903, opened 2019-09-27
Fix jvm tests.

trams:
2 things are done

1. all data files are read from resources not by relative path in a source tree (this enables easy move to SBT or other build system)
2. fixed DTDs

This is an extraction from this pull request https://github.com/dmlc/xgboost/pull/4687

### DIFF
--- a/jvm-packages/checkstyle.xml
+++ b/jvm-packages/checkstyle.xml
@@ -16,8 +16,8 @@
   -->
 
 <!DOCTYPE module PUBLIC
-          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
-          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
+    "https://checkstyle.org/dtds/configuration_1_3.dtd">
 
 <!--
 
--- a/jvm-packages/create_jni.py
+++ b/jvm-packages/create_jni.py
@@ -110,16 +110,13 @@ if __name__ == "__main__":
        "xgboost4j/src/main/resources")
 
     print("copying train/test files")
-    maybe_makedirs("xgboost4j-spark/src/test/resources")
+    maybe_makedirs("xgboost4j/src/test/resources")
     with cd("../demo/regression"):
         run("{} mapfeat.py".format(sys.executable))
         run("{} mknfold.py machine.txt 1".format(sys.executable))
 
-    for file in glob.glob("../demo/regression/machine.txt.t*"):
-        cp(file, "xgboost4j-spark/src/test/resources")
-    for file in glob.glob("../demo/data/agaricus.*"):
-        cp(file, "xgboost4j-spark/src/test/resources")
-
     maybe_makedirs("xgboost4j/src/test/resources")
+    for file in glob.glob("../demo/regression/machine.txt.t*"):
+        cp(file, "xgboost4j/src/test/resources")
     for file in glob.glob("../demo/data/agaricus.*"):
         cp(file, "xgboost4j/src/test/resources")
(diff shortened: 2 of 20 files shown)


jvm-packages/checkstyle.xml
```diff
@@ -16,8 +16,8 @@
   -->
 
 <!DOCTYPE module PUBLIC
-          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
-          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
+    "https://checkstyle.org/dtds/configuration_1_3.dtd">
```
CodingCat: just question, what happened to them?

jvm-packages/create_jni.py
```diff
@@ -110,16 +110,13 @@ def normpath(path):
...
         run("{} mapfeat.py".format(sys.executable))
         run("{} mknfold.py machine.txt 1".format(sys.executable))
 
-    for file in glob.glob("../demo/regression/machine.txt.t*"):
-        cp(file, "xgboost4j-spark/src/test/resources")
-    for file in glob.glob("../demo/data/agaricus.*"):
-        cp(file, "xgboost4j-spark/src/test/resources")
-
     maybe_makedirs("xgboost4j/src/test/resources")
+    for file in glob.glob("../demo/regression/machine.txt.t*"):
```
CodingCat: if machine.txt.t* is only used for xgboost4j-spark, why we put it in xgboost4j?

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PersistenceSuite.scala
```diff
@@ -20,10 +20,9 @@ import java.io.File
 import java.util.Arrays
 
 import scala.io.Source
+import ml.dmlc.xgboost4j.scala.{Classification, DMatrix, Regression}
```
CodingCat: why we have many adding of imports but never use them?

jvm-packages/xgboost4j/src/test/scala/ml/dmlc/xgboost4j/scala/TrainTestData.scala
```diff
@@ -14,12 +14,14 @@
  limitations under the License.
  */
 
-package ml.dmlc.xgboost4j.scala.spark
+package ml.dmlc.xgboost4j.scala
```
CodingCat: if you move this file to xgboost4j, there is a significant part duplicate with TrainTestData.java and Classification.java

additionally, most of methods here are only called by spark package, why we need to move?

jvm-packages/checkstyle.xml
trams: Frankly I do not know.
It may be worth reporting to the owner of puppycrawl.com

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PersistenceSuite.scala
trams: I checked. We use all of those added imports.
One needs to understand why they are not in this diff

CodingCat: ah, I think you moved Classification and Regression to xgboost4j leading to these changes

jvm-packages/xgboost4j/src/test/scala/ml/dmlc/xgboost4j/scala/TrainTestData.scala
trams: This is a good point. I remember that I moved it because there was at least one test in xgboost4j which used this data but now I can't seem to figure out which one.

So I will probably try to move it back.

hcho3:
Closing, as I don't think we'd move to SBT any time soon.

PR #4908, opened 2019-10-01
Check deprecated `n_gpus`.

trivialfis:


trivialfis:
If you set `n_gpus`  to greater than 1, it still throws an error.  Throws warning only when `n_gpus == 1`.

PR #4914, opened 2019-10-06
Resolve dask performance issues

RAMitchell:
Resolves performance bug where dask attempts to generate hashes from nested functions with large arguments in client.map.

On the YearPredictMSD dataset this reduces runtime from 90s to 45s when using dask.

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -369,6 +369,7 @@ def train(client, params, dtrain, *args, evals=(), **kwargs):
 
     futures = client.map(dispatched_train,
                          range(len(worker_map)),
+                         pure=False,
                          workers=list(worker_map.keys()))
     results = client.gather(futures)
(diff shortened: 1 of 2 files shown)


trivialfis:
Let me try some others.  I'm convinced that using simple functions without any capturing is a better choice.

RAMitchell:
After some digging we found that DaskDMatrix.n_rows variable was causing this class to store the entire dataset when serialised by dask, where it should be just a lightweight wrapper class. This brings the performance of the dask API back to normal.

hcho3:
And now S3 downloads are timing out :(
https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-4914/4/pipeline

hcho3:
I noticed that the S3 bucket for storing artifacts and stashes exceeded 200 GB in size. I emptied the bucket and will restart the test.

RAMitchell:
Thanks Philip! Let me know if there is anything I can do.

PR #4924, opened 2019-10-09
Update dmlc-core.

trivialfis:
* Fixed some threading errors.
* Allow updating parameters.


hcho3:
Re-invoking Jenkins as a test

PR #4927, opened 2019-10-10
[CI] Run cuDF tests in Jenkins CI server

### LINKED ISSUE #4792: Set up test for cudf.
After merging the cudf PR,  I need some help for setting up corresponding Python test on Jenkins since on the one hand cudf doesn't support pip, but on the other hand our Jenkins test farm doesn't use conda.

Could you please help?  @mt-jones @hcho3

hcho3:
Closes #4792

@trivialfis @RAMitchell Feel free to tag more people who can help

### DIFF
--- a/tests/ci_build/test_python.sh
+++ b/tests/ci_build/test_python.sh
@@ -36,7 +36,8 @@ case "$suite" in
     ;;
 
   cudf)
-    pytest -v -s --fulltrace tests/python-gpu/test_from_columnar.py
+    source activate cudf_test
+    python -m pytest -v -s --fulltrace tests/python-gpu/test_from_columnar.py tests/python-gpu/test_gpu_with_dask.py
     ;;
 
   cpu)
--- a/tests/ci_build/test_python.sh
+++ b/tests/ci_build/test_python.sh
@@ -35,6 +35,10 @@ case "$suite" in
     ./runtests-gpu.sh
     ;;
 
+  cudf)
+    pytest -v -s --fulltrace tests/python-gpu/test_from_columnar.py
+    ;;
+
   cpu)
     pytest -v -s --fulltrace tests/python
     cd tests/distributed
@@ -42,7 +46,7 @@ case "$suite" in
     ;;
 
   *)
-    echo "Usage: $0 {gpu|mgpu|cpu}"
+    echo "Usage: $0 {gpu|mgpu|cudf|cpu}"
     exit 1
     ;;
 esac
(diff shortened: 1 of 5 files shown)


tests/ci_build/test_python.sh
```diff
@@ -35,14 +35,18 @@ case "$suite" in
     ./runtests-gpu.sh
     ;;
 
+  cudf)
+    pytest -v -s --fulltrace tests/python-gpu/test_from_columnar.py
```
trivialfis:
```suggestion
    pytest -v -s --fulltrace tests/python-gpu/test_from_columnar.py tests/python-gpu/test_gpu_with_dask.py
```

hcho3:
@trivialfis `test_from_columnar.py` is passing (Hooray!) but `test_gpu_with_dask.py` is not yet passing. At least now cuDF is working. Would you be able to look at it?

https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-4927/9/pipeline/#step-325-log-220

hcho3:
`test_gpu_with_dask.py` has been hanging for one hour: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-4927/9/pipeline/#step-325-log-246 I'm going to kill it. Can you look at the test so that it doesn't hang upon failure?

trivialfis:
That's weird.  I haven't encountered that.  Let's skip that test.  Will try to reproduce it.  Maybe it's due to preemptive?

trivialfis:
@hcho3 Sorry for pinging you on this again.  Could  you give an estimation of how good a chance that we can't allocate any GPU on Jenkins?

hcho3:
@trivialfis We should have exclusive access, as long as the EC2 instance is launched. Note that currently the cuDF suite has access to a single GPU. Does the Dask test require more than one GPU? If that’s the case, let’s look at it after I’m back.

trivialfis:
@hcho3 No.  Any number of GPU is fine.  But currently neither this PR nor https://github.com/dmlc/xgboost/pull/4928 is getting any GPU.

hcho3:
@trivialfis I see that multi-GPU tests are timing out because multi-GPU worker is not allocated in time.

I went ahead and raised the limit for multi-GPU workers to 3. (*) Hopefully this will resolve the problem.

(*) Now that we are using G4, the cost is lower, so we can afford to use a higher limit.

trivialfis:
@hcho3 Thanks for taking your time on this.  Seems resolved. ;-)

> Now that we are using G4, the cost is lower, so we can afford to use a higher limit.

That's a really good news.

PR #4928, opened 2019-10-10
Ignore columnar alignment requirement.

trivialfis:
* Ignore the 8 bytes requirement.
* Fix stride size.
* tests.


trivialfis:
Don't merge it right now.

PR #4930, opened 2019-10-11
[WIP] Avoid copying data to host in ellpack init.

trivialfis:
Data is still copied during sketching.  This PR only tries to create specialization for ellpack when data is coming from device. E.g. cudf.

@rongou This might has some conflicts with your PR.  Preferably we should merge #4879 first.  Would you like to take a look?


trivialfis:
@rongou Closing for now.  Will wait for all the refactoring.

PR #4935, opened 2019-10-12
Span: use `size_t' for index_type,  add `front' and `back'.

trivialfis:
* Remove a batch of `static_cast'.

Originally I implemented `Span` with some conventions from GSL, where uses `int64_t` as `index_type`.  But `span` in c++20 actually uses `size_t`, and during the development I found `size_t` should be much more consistent with the rest of the world.  So this PR changes `Span` to align with std.

* Added front and back from c++20 for completeness.

### DIFF
--- a/include/xgboost/span.h
+++ b/include/xgboost/span.h
@@ -31,7 +31,8 @@
 
 #include <xgboost/logging.h>  // CHECK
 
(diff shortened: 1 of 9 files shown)


trivialfis:
Jenkins is not giving me any log ...

PR #4937, opened 2019-10-12
Warn when _validate_features sets feature_names

### LINKED ISSUE #4854: [python-package] running .predict() on a loaded model silently sets the feature_names attribute
When I load a model with `bst.load_model()` that was saved using `bst.save_model()`, and then call `bst.predict(X)` for the first time, the model silently is given the `feature_names` of `X`, which can be dangerous behavior if the `X` isn't actually the full size of the model. This is a consequence of the following code in `Booster._validate_features(self, data)`:

```python
if self.feature_names is None:
    self.feature_names = data.feature_names
    self.feature_types = data.feature_types
```

I know the documentation recommends pickling, but I think a warning should be displayed if `_validate_features()` is actually *setting* `feature_names` rather than validating.

Example:

```python
In [1]: import xgboost as xgb

In [2]: import numpy as np

In [5]: X_train = np.random.rand(5,10)

In [8]: bst = xgb.XGBRegressor()

In [13]: y_train = np.random.rand(5,1)

In [14]: bst.fit(X_train, y_train)
[08:23:04] WARNING: src/objective/regression_obj.cu:152: reg:linear is now deprecated in favor of reg:squarederror.
Out[14]:
XGBRegressor(base_score=0.5, booster='gbtree', colsample_bylevel=1,
             colsample_bynode=1, colsample_bytree=1, gamma=0,
             importance_type='gain', learning_rate=0.1, max_delta_step=0,
             max_depth=3, min_child_weight=1, missing=None, n_estimators=100,
             n_jobs=1, nthread=None, objective='reg:linear', random_state=0,
             reg_alpha=0, reg_lambda=1, scale_pos_weight=1, seed=None,
             silent=None, subsample=1, verbosity=1)

In [17]: bst.save_model('model')

In [25]: bst_loaded = xgb.XGBRegressor()

In [26]: bst_loaded.load_model('model')
[08:26:24] WARNING: src/objective/regression_obj.cu:152: reg:linear is now deprecated in favor of reg:squarederror.

In [27]: X_pred_1 = np.random.rand(1, 9) # oops! wrong input shape

In [28]: bst_loaded.predict(X_pred_1) # still returns prediction
Out[28]: array([0.25939447], dtype=float32)

In [29]: bst_loaded.get_booster().feature_names # silently has defined `feature_names`
Out[29]: ['f0', 'f1', 'f2', 'f3', 'f4', 'f5', 'f6', 'f7', 'f8']

In [30]: X_pred_2 = np.random.rand(1, 10) # now try correct input shape

In [31]: bst_loaded.predict(X_pred_2)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-31-7e193d6f83c4> in <module>
----> 1 bst_loaded.predict(X_pred_2)

~\AppData\Local\Continuum\miniconda3\envs\mpopp_infra\lib\site-packages\xgboost\sklearn.py in predict(self, data, output_margin, ntree_limit, validate_features)
    454                                           output_margin=output_margin,
    455                                           ntree_limit=ntree_limit,
--> 456                                           validate_features=validate_features)
    457
    458     def apply(self, X, ntree_limit=0):

~\AppData\Local\Continuum\miniconda3\envs\mpopp_infra\lib\site-packages\xgboost\core.py in predict(self, data, output_margin, ntree_limit, pred_leaf, pred_contribs, approx_contribs, pred_interactions, validate_features)
   1282
   1283         if validate_features:
-> 1284             self._validate_features(data)
   1285
   1286         length = c_bst_ulong()

~\AppData\Local\Continuum\miniconda3\envs\mpopp_infra\lib\site-packages\xgboost\core.py in _validate_features(self, data)
   1688
   1689                 raise ValueError(msg.format(self.feature_names,
-> 1690                                             data.feature_names))
   1691
   1692     def get_split_value_histogram(self, feature, fmap='', bins=None, as_pandas=True):

ValueError: feature_names mismatch: ['f0', 'f1', 'f2', 'f3', 'f4', 'f5', 'f6', 'f7', 'f8'] ['f0', 'f1', 'f2', 'f3', 'f4', 'f5', 'f6', 'f7', 'f8', 'f9']
training data did not have the following fields: f9
```

Happy to submit a PR that issues a warning when `Booster._validate_features(self, data)` is called with `self.feature_names is None`.

awbirdsall:
- Current behavior silently sets feature_names of Booster to
  match data when Booster.predict(X) is called on a loaded model
- Add warning to relevant section of Booster._validate_features
  so user is aware it is happening.
- Warning text: "Booster's feature_names did not exist. Booster's
  feature_names and feature_types are being set to match the data."
- Closes #4854

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -1819,6 +1819,9 @@ class Booster(object):
         Set feature_names and feature_types from DMatrix
         """
         if self.feature_names is None:
+            warnings.warn("Booster's feature_names did not exist. Booster's "
+                          "feature_names and feature_types are being set to "
+                          "match the data.")
             self.feature_names = data.feature_names
             self.feature_types = data.feature_types
         else:


trivialfis:
Looks good.  Thanks.  I wonder if there's an easy way to store utf-8 feature names in c++ ...

trivialfis:
A little hesitate to merge it since the message seems inevitable for most use cases.

trivialfis:
May be able to remove the validation if we can resolve https://github.com/dmlc/xgboost/pull/4594

awbirdsall:
Yes, I agree that would be ideal. In the meantime no matter how common the use case (or even *because* of how common the use case!) I think it’s important for there to be some kind of guardrail against a mismatch between the trained model and the data. Displaying a warning seems like one reasonable way to prevent a “validate features” step from actually silently coercing features to match, but I could also imagine a more involved fix.

trivialfis:
@awbirdsall You are definitely right about the warning.  But let me give a shot of fixing the above mentioned issue first.  It might take a few days if I weren't able to do so I will merge this one. ;-)  Thanks for the patience.

awbirdsall:
No problem, thanks for the work on the library! :)

trivialfis:
Progress on https://github.com/dmlc/xgboost/pull/4954 .

hcho3:
@trivialfis Should we consider saving the feature names in the Booster, like how we save them in the `DMatrix`? Otherwise, we can't do a reliable feature validation after saving and loading the model.

trivialfis:
Thought about it before, but not sure yet.  I don't really like xgboost generating pseudo feature names.

hcho3:
@trivialfis We should consider removing the feature name validation option, since due to the lack of serialization of feature names, the option is only half working.

trivialfis:
Got it.  We may remove the feature name generation, and save valid feature names obtained from pandas and alike.  WDYT?

hcho3:
@trivialfis Sounds good, if your proposal also entails saving feature names in the C++ layer.

trivialfis:
Yup.  Always prefer doing things in c++ when possible.

hcho3:
@trivialfis I opened #6520 to keep track of the proposal.

@awbirdsall I am closing this pull request in favor of #6520. We want a more robust solution to store the feature names in the model files.

awbirdsall:
@hcho3 thanks for the update and chasing after this! The improvements in the new issue look like they should be very helpful 👍

PR #4941, opened 2019-10-13
Fix dask prediction.

trivialfis:


RAMitchell:
I am coming up against this assertion on some training runs also `assert X_parts.shape[1] == 1`

trivialfis:
Em...  I can't reproduce what you said.  Will keep looking.

trivialfis:
Let me create some benchmark scripts for prediction.

trivialfis:
@RAMitchell 

That's the reason I added so many insertions .. Here are 3 insertions you might fail:
- `inconsistent` insertion in this PR:  This might due to imbalanced data, you can cal `data.rechunk` or `data.repartition` if it's a dataframe to make it work.  But this throw performance out of window.  Luckily this don't usually happen in real dataset since you load the label and data altogether with one csv.  I added the suggestion into error msg.
- `cols in (0, c)`.  The number of columns for each dataset must be the same.  Your input is not correct, might due to sparsity in your dataset.  Added a hint in error message.
- `X_parts.shape[1] == 1`.  No idea.  Could you give a reproducible script so I can check?

RAMitchell:
Here was the problem, I had the following
```python
partition_size = 1000
X = da.from_array(data.X_train, partition_size)
```
The reason this fails is the dataset had 2000 features - more than the partition size. When you specify a single value for partition size it actually sets all dimensions to this value not just the first.

This fixes it
```python
partition_size = 1000
X = da.from_array(data.X_train, (partition_size, data.X_train.shape[1]))
```

I think you can reproduce this bug in your demo (https://github.com/dmlc/xgboost/blob/master/demo/dask/gpu_training.py) by setting `n = 1001`.

I found this unintuitive, how can we make this better? We could coerce the data into the correct dimensions and log a warning instead, or just fail with a better message. The demos will need updating also.

Maybe something like:
"Warning: Data should be partitioned by row, re-partitioning. To avoid this specify the number of columns for your dask DataFrame/Array explicitly. e.g. `chunks=(partition_size, X.shape[1])`"

Issue #4855, opened 2019-09-14
[RFC] Distinguish model, booster and runtime when performing IO operations.


trivialfis:
## Background

I will use Python package as the example for below discussion, but it should be equally applicable to other language bindings.  Right now when XGBoost saves it's "model", it saves the `tree_model` or `linear_model`, plus some attributes like `max_delta_step` and `predictor`, which are not strictly related to the "model" itself.  Take neural network as an example, the "model" should be its weight and structure, and should has nothing to do with objective function or optimizer.  Such boundary also exists in XGBoost, where the concept of model needs to be restricted to trees or linear weights only.  I believe this is how the current IO interface was designed.

But now we have some more complicated situations, like the `Booster` is used in pickle, transferred between workers for distributed learning, used as checkpoing for model recovery etc.  In some of these scenarios, optimizer(or the term used in XGBoost, `tree_method`) and some objective parameters are crucial components in the saved file (or memory buffer, you know what I mean).  So with time we added some of training parameters (like the `max_delta_step` hyper-parameter) into the output file, and some more are coming with the excellent work by @chenqin for better rabit recovery procedures.

Also, there are generic parameters that define the runtime configuration, like `gpu_id`, `nthread` and `verbosity`, these are even less correlated to the concept of "model".  Saving them to disk is not really making sense.  But it's sometime necessary to serialize them in distributed setting.  For instance, right now the distributed monitor actually doesn't work as expected in distributed environment due to the fact that `verbosity` is lost in transferred booster.

## Old Proposal
I proposed saving the "complete snapshot" of XGBoost before while trying to introduce the json serialization format, which should mitigate the above issues.  But as described, most of those parameters are not strictly related to the concept of "model", when user is only trying to save the resulting model and use it for prediction or want to share it over internet, those parameters are not only useless, but also adds additional complexity.  

## New one
So here I suggest we split the model IO into two different logics, one for the model itself, like trees, while the other for a complete closure. 

For the first logic, it should be used in normal model IO, like `Booster.save_model`. We output only the trees and strictly related parameters like `num_features`.  No `max_delta_step`, no `predictor` nor `gpu_id`. So it **can not** be used for model recovery or transferred between workers during distributed training.  When users want to continue training on saved model, they need to set the parameters again via `booster.set_param`.  Since a tree is a model, it doesn't know how to train itself.  And we promise to keep the stability of this output.

[...]

Issue #4755, opened 2019-08-09
[RFC] Redesign `dmlc::Parameter`.


trivialfis:
There are a few problems in `dmlc::Parameter` I want to address:
# Issues
## Parameter update
Currently `dmlc::Parameter` uses a static manager (`__MANAGER()__` method) class to carry out all the initialization steps.  The problem is this manager is created per-class instead of per-instance, so there is no way that it can understand whether a particular parameter instance has been initialized.  As a side effect, we cannot do update operation on it, which results in every call of `Configure` a complete re-initialization. See #4738 .  Once JSON is here the problem will spread through all other parameters.

- Adding a `initialized_` field to `dmlc::Parameter` won't work as it will break all binary IO logic (size is changed).
- Adding a `std::map<Parameter*, bool> initialized_` field to the manager won't work, as distinguishing instance by pointer breaks easily with stack allocated objects:
``` c++
{
  Param p;
}
{
  Param p0;  // Same address as p
}
```

- Adding non-trivial constructor/destructor to `dmlc::Parameter` calling `__MANAGER()__` won't work.  I tried and get a dead lock from static variables, stuck at :
```
__cxa_guard_acquire 
```
- Adding `std::map<std::string, FieldAccessEntry*>>` to manager won't work, again it operates on class instead of instance.

## Initialization

Another issue is the parameter can't initialize itself.  It troubles me and @RAMitchell  for quite a while now as those primitive types have undefined values at initialization.  The stack example above, `param0` might has same value as `param` (actually compiler just reuse `param` for `param0` even in debugging compilation).  Also, one can imagine calling any method from `__MANAGER__` is quite dangerous.

## Description is not very useful.

The one most duplicated code in XGB is those descriptions of parameters, as there's no way we can query the document from C++ level, resulting in those descriptions being copied across XGBoost.

## Namespace

As XGBoost has so many parameters that can form a small embedded language, putting all of them into one `std::vector` makes precise validation really difficult.

# What's next
I don't have any concrete idea yet.  But discussions are welcomed.  @RAMitchell @CodingCat @hcho3 Please join.

hcho3:
Yes! This is a long-due discussion. See #3803, for example, where we have no way to detect runtime update of a field in `dmlc::Parameter` and thus leak memory.

RAMitchell:
Thanks @trivialfis. As this functionality lives in dmlc-core it can be very difficult for us to change as we will impose changes on mxnet etc. 

I see two paths forward:
[...]

PR #4945, opened 2019-10-14
Add Model and Configurable interface.

trivialfis:
* Apply Configurable to objective functions.
* Apply Model to Learner and Regtree, gbm.
* Add Load/SaveConfig to objs.
* Refactor obj tests to use smart pointer.
* placeholders for Save/Load Model.

A small portion of #4732.  Also related to #4855.

### DIFF
--- a/include/xgboost/generic_parameters.h
+++ b/include/xgboost/generic_parameters.h
@@ -7,7 +7,7 @@
 
 #include <dmlc/parameter.h>
 #include <xgboost/logging.h>
-#include <xgboost/enum_class_param.h>
+#include <xgboost/parameter.h>
 
 #include <string>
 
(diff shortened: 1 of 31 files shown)


trivialfis:
Closing for now, see if I can remove some burdens on Jenkins.

trivialfis:
@RAMitchell I have two RFC #4855 , #4755 . The UpdateAllowUnknown came from the second one.  Does that help?

trivialfis:
But yeah, I can provide an example, a few mins.

trivialfis:
@RAMitchell `tweedie_variance_power` in `TweedieRegression` is not saved in learner.  So in dask if you pickled to another worker, it will happily use the default value instead of one provided by user.

And I have these many related and opened issues: https://github.com/dmlc/xgboost/pull/4732#issuecomment-535485608

PR #4947, opened 2019-10-15
Don't `set_params` at the end of `set_state`.

trivialfis:
* Also fix another issue found in dask prediction.

### DIFF
--- a/tests/python/test_pickling.py
+++ b/tests/python/test_pickling.py
@@ -1,0 +1,48 @@
+import pickle
+import numpy as np
+import xgboost as xgb
+import os
+
+
+kRows = 100
+kCols = 10
+
+
+def generate_data():
+    X = np.random.randn(kRows, kCols)
+    y = np.random.randn(kRows)
+    return X, y
+
+
+def test_model_pickling():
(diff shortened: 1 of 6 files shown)


tests/python/test_pickling.py
```diff
@@ -0,0 +1,47 @@
...
+    with open(filename, 'rb') as fd:
+        bst = pickle.load(fd)
+
+    with open(filename, 'wb') as fd:
+        pickle.dump(bst, fd)
+
+    with open(filename, 'rb') as fd:
+        bst = pickle.load(fd)
+
+    assert bst.get_dump(dump_format='json')
```
JohnZed: should we assert that this is identical to the first model dump not just nonempty?

trivialfis: Yup.  We have continuous training test, but adding it would be useful.

trivialfis:
Also partially revert the previous fix.  My brain is fried ...

RAMitchell:
For others following, this appears to be a fundamental error in python serialisation where the model can be pickled and unpickled multiple times, resulting in a different model.

PR #4954, opened 2019-10-16
[JVM-Package][WIP]  Add missing value as parameter for DMatrix.

trivialfis:
Continue #4594 .


trivialfis:
@CodingCat I got some kind help for narrowing the test failure to this:

https://github.com/dmlc/xgboost/blob/185e3f19169d79b0910b848de5fe8055434e675e/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala#L256

Could you elaborate on why Scala needs to removing missing value itself?

trivialfis:
@CodingCat I think I found the error, here this dataframe is passed as multiple data iterator:

https://github.com/dmlc/xgboost/blob/010b8f1428a3302d7500e7adfd933ba1d00b5f5c/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/MissingValueHandlingSuite.scala#L52

The first row is passed as a standalone iter (hence creating a DMatrix with single row), and Scala side removes the `float.NaN` before pushing it into c++, so there's only 2 columns left in there.

I think the primary reason of doing this is because creating `SimpleCSRSource` from parser does handle missing value, https://github.com/dmlc/xgboost/blob/010b8f1428a3302d7500e7adfd933ba1d00b5f5c/src/data/simple_csr_source.cc#L73.  I can implement the missing value handling in C++.  Could you help removing the scala version of missing value handling?

trivialfis:
I can't go all  the way up to scala ...

jvm-packages/xgboost4j/src/main/scala/ml/dmlc/xgboost4j/scala/DMatrix.scala
```diff
@@ -35,11 +35,12 @@ class DMatrix private[scala](private[scala] val jDMatrix: JDMatrix) {
     *  init DMatrix from Iterator of LabeledPoint
     *
     * @param dataIter An iterator of LabeledPoint
+    * @param missing the specified value to represent the missing value
     * @param cacheInfo  Cache path information, used for external memory setting, null by default.
     * @throws XGBoostError native error
     */
-  def this(dataIter: Iterator[LabeledPoint], cacheInfo: String = null) {
-    this(new JDMatrix(dataIter.asJava, cacheInfo))
```
CodingCat: instead of doing this, can we just add a new constructor?

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/DMatrix.java
```diff
@@ -40,18 +40,20 @@
...
     if (iter == null) {
       throw new NullPointerException("iter: null");
     }
     // 32k as batch size
     int batchSize = 32 << 10;
     Iterator<DataBatch> batchIter = new DataBatch.BatchIterator(iter, batchSize);
     long[] out = new long[1];
-    XGBoostJNI.checkCall(XGBoostJNI.XGDMatrixCreateFromDataIter(batchIter, cacheInfo, out));
+    XGBoostJNI.checkCall(XGBoostJNI.XGDMatrixCreateFromDataIterEx(
+            batchIter, missing, cacheInfo, out));
```
CodingCat: let's keep the original one and add new constructor

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/MissingValueHandlingSuite.scala
```diff
@@ -48,6 +48,7 @@ class MissingValueHandlingSuite extends FunSuite with PerTest {
   test("handle Float.NaN as missing value correctly") {
     val spark = ss
     import spark.implicits._
+    println("handle Float.NaN as missing value correctly")
```
CodingCat: and remove printlns

trivialfis:
@CodingCat  So glad that you are here ..

trivialfis:
@CodingCat 

> I can help to remove missing value handling code in Scala

How do you want me to proceed?  Should I simply add the missing value handling in c++ in this PR and remove other changes?  There's a check in this PR that will fail the current missing value handling test in scala.

PR #4959, opened 2019-10-18
Follow Docstring Conventions

TheJacobKim:
## PR Description
I have noticed an inconsistent convention used throughout xgboost's core Python file. So I made some changes that will make the core file to respect docstring conventions according to [PEP 257](https://www.python.org/dev/peps/pep-0257/).

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -108,7 +108,7 @@ def from_cstr_to_pystr(data, length):
 
 
 def _expect(expectations, got):
-    '''Translate input error into string.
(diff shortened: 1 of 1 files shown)


RAMitchell:
Thanks, is it possible to enforce this via pylint?

trivialfis:
Thanks.

TheJacobKim:
I am not sure if there is a linter for this, but there should be one.

trivialfis:
Em.  Usually I just build the doc myself and if there's no warning then I will submit it.

PR #4960, opened 2019-10-18
Catch exception in transform function omp context.

trivialfis:
Related to #4869.

### DIFF
--- a/tests/cpp/common/test_transform_range.cc
+++ b/tests/cpp/common/test_transform_range.cc
@@ -50,13 +50,26 @@ TEST(Transform, DeclareUnifiedTest(Basic)) {
   out_vec.Fill(0);
 
   Transform<>::Init(TestTransformRange<bst_float>{},
-	                Range{0, static_cast<Range::DifferenceType>(size)},
-	                TRANSFORM_GPU)
+                    Range{0, static_cast<Range::DifferenceType>(size)},
+                    TRANSFORM_GPU)
       .Eval(&out_vec, &in_vec);
   std::vector<bst_float> res = out_vec.HostVector();
 
   ASSERT_TRUE(std::equal(h_sol.begin(), h_sol.end(), res.begin()));
 }
+
+#if !defined(__CUDACC__)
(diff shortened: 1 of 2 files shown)


tests/cpp/common/test_transform_range.cc
```diff
@@ -50,13 +50,26 @@ TEST(Transform, DeclareUnifiedTest(Basic)) {
...
-	                TRANSFORM_GPU)
+                    Range{0, static_cast<Range::DifferenceType>(size)},
+                    TRANSFORM_GPU)
       .Eval(&out_vec, &in_vec);
   std::vector<bst_float> res = out_vec.HostVector();
 
   ASSERT_TRUE(std::equal(h_sol.begin(), h_sol.end(), res.begin()));
 }
 
+#if !defined(__CUDACC__)
```
hcho3: Why is this macro guard needed?

trivialfis: @hcho3 This file is included in another cu file so we can have test in both gpu and cpu. (that's what the transform for, define one function and run it on both sides).  But this test only runs on CPU since it's related to openmp, so we don't compile it on cuda mode.

tests/cpp/common/test_transform_range.cc
```diff
@@ -50,13 +50,26 @@ TEST(Transform, DeclareUnifiedTest(Basic)) {
...
+TEST(Transform, Exception) {
+  size_t const kSize {16};
+  std::vector<bst_float> h_in(kSize);
+  const HostDeviceVector<bst_float> in_vec{h_in, -1};
+  EXPECT_ANY_THROW({
+    Transform<>::Init([](size_t idx, common::Span<float const> _in) { _in[idx + 1]; },
+                      Range(0, static_cast<Range::DifferenceType>(kSize)), -1)
+        .Eval(&in_vec);
```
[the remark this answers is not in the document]
trivialfis: @hcho3 Out of range index, Span catches it and throws.

trivialfis:
@hcho3 Thanks.

PR #4961, opened 2019-10-18
Use `UpdateAllowUnknown' for non-model related parameter.

trivialfis:
Model parameter can not pack an additional boolean value due to binary IO
format.  This commit deals only with non-model related parameter configuration.

### DIFF
--- a/tests/ci_build/tidy.py
+++ b/tests/ci_build/tidy.py
@@ -244,7 +244,7 @@ if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Run clang-tidy.')
     parser.add_argument('--cpp', type=int, default=1)
     parser.add_argument('--cuda', type=int, default=1)
-    parser.add_argument('--use-dmlc-gtest', type=int, default=0,
+    parser.add_argument('--use-dmlc-gtest', type=int, default=1,
                         help='Whether to use gtest bundled in dmlc-core.')
     args = parser.parse_args()
 
--- a/tests/ci_build/tidy.py
+++ b/tests/ci_build/tidy.py
@@ -26,17 +26,22 @@ def call(args):
 
 
 class ClangTidy(object):
-    '''
-    clang tidy wrapper.
+    ''' clang tidy wrapper.
     Args:
-      cpp_lint: Run linter on C++ source code.
-      cuda_lint: Run linter on CUDA source code.
+      args:  Command line arguments.
+          cpp_lint: Run linter on C++ source code.
+          cuda_lint: Run linter on CUDA source code.
+          use_dmlc_gtest: Whether to use gtest bundled in dmlc-core.
     '''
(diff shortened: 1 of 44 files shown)


hcho3:
Is there a problem that you see only in the test farm?

trivialfis:
@hcho3 It should be fixed now.  Let's see if the tests pass.

trivialfis:
@hcho3 I thing I didn't think of, is clang-tidy doesn't understand static initialization of `dmlc::Parameter`.  It believes that the `entry_map_` in parameter manager is empty so uses of it might trigger a clang-tidy error.

trivialfis:
@hcho3 Done.

tests/ci_build/tidy.py
```diff
@@ -234,11 +239,13 @@ def test_tidy():
     parser = argparse.ArgumentParser(description='Run clang-tidy.')
     parser.add_argument('--cpp', type=int, default=1)
     parser.add_argument('--cuda', type=int, default=1)
+    parser.add_argument('--bundle', type=int, default=1,
```
hcho3: Not sure how useful this argument is going to be, since currently, `tidy.py` doesn't let the user specify the path of google test. Either we should add `--gtest-path` option or remove `--bundle` option. My personal preference is to always use the bundled gtest.

tests/cpp/common/test_parameter.cc
```diff
@@ -53,3 +57,44 @@ TEST(EnumClassParam, Basic) {
...
+    // When it's not initialized, perform set_default.
+    ASSERT_NEAR(p.f, 11.0f, kRtEps);
+    ASSERT_NEAR(p.d, 2.71828f, kRtEps);
+
+    p.d = 3.14149;
+
+    p.UpdateAllowUnknown(xgboost::Args{{"f", "2.71828"}});
+    ASSERT_NEAR(p.f, 2.71828f, kRtEps);
+
+    // p.d is un-effected by the update.
```
hcho3: Nice!

tests/ci_build/tidy.py
trivialfis: I installed gtest on my system path so CMake always handles it for me...

trivialfis:
@RAMitchell Not sure about this.  We can discuss it after having model parameter (these are runtime parameters).

tests/ci_build/tidy.py
hcho3: In this case, let's rename the option to `--use-system-gtest`

trivialfis:
@hcho3 I added a `use-dmlc-gtest` option and set it to default.  The naming is consistent with our CMake parameter.

trivialfis:
Now I see the problem on Jenkins...

hcho3:
@trivialfis Let's enable bundled gtest for CI tasks.

trivialfis:
@hcho3 Done.

PR #4973, opened 2019-10-22
[CI] Use latest dask

hcho3:
Dask tests are stuck and timing out due to dask/dask#5465: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-4971/2/pipeline/40/#step-75-log-1677. The cause of the timeout is because one of dependencies of Dask, `distributed`, changed its semantics in recent versions in ways that are not compatible with Dask 2.0.0. Fix: Pin version of `distributed`.

### DIFF
--- a/tests/ci_build/Dockerfile.cpu
+++ b/tests/ci_build/Dockerfile.cpu
@@ -24,7 +24,7 @@ RUN \
     pip install pyyaml cpplint pylint astroid sphinx numpy scipy pandas matplotlib sh recommonmark guzzle_sphinx_theme mock \
                 breathe matplotlib graphviz pytest scikit-learn wheel kubernetes urllib3 && \
     pip install https://h2o-release.s3.amazonaws.com/datatable/stable/datatable-0.7.0/datatable-0.7.0-cp37-cp37m-linux_x86_64.whl && \
-    pip install "dask[complete]==2.6.0"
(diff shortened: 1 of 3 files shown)


trivialfis:
I thought 2.5 is available now?  Can we remove the version number and stick with the latest version?

hcho3:
@trivialfis What is our commitment with Dask support? Is it sufficient to use the latest all the time, or should we consider users of older versions? (I'm assuming we'd need at least 2.x)

trivialfis:
No.  Just keep running with the latest version.  Dask is very light weight and use small releases.  Also we don't use any internal API so if something breaks we can fix it in dask.

hcho3:
@trivialfis  When Dask 2.0 was released, it renamed an internal variable and broke XGBoost, so I had to submit #4617. I wonder whether something like this could happen again. We could choose the middle groun and only specify the major release of Dask, e.g. `dask>=2.0.0,<3.0.0`.

trivialfis:
I wouldn't worry about compatibility with old dask.  We haven't release 1.0 yet and our dask interface went through a complete rewrite during these few months.  So I would very much like to start with everything clean.

________________________________

hcho3:
Thanks for clarification.

trivialfis:
Thanks!

Issue #4843, opened 2019-09-09
Memory leak in subsequent model fitting?


GuidoBartoli:
Hi,
I created a XGBoost model with data subsampling (`subsample < 1`) and I implemented a loop to train it using different `random_state` values and choose the best performing initialization.

This is the simplified Python code:
```
# setup train_x, train_y, eval_set, eval_metric #
accuracies = []
best = None
for i in range(100):
    variation = xgb.XGBClassifier(random_state=i, n_jobs=-1)
    variation.fit(train_x, train_y, eval_set=eval_set, eval_metric=eval_metric)
    accuracy = 1 - variation.evals_result()['validation_1']['error'][-1]
    if best is None or accuracy > max(accuracies):
        best = variation
    accuracies.append(accuracy)
```

It works as expected, but I noticed that if I monitor the RAM during the process, the usage continues to grow after each iteration even if I assigned a new XGBClassifier to the same variation variable. Python should free the unreferenced memory location every time and maintain the memory usage constant, right?

This is the screenshot of System Monitor Resources:
![memory](https://user-images.githubusercontent.com/16103676/64523864-8adfea80-d2fd-11e9-85c6-b2e01edca355.jpg)
Computation starts at the first CPU spike at 100% on the left and a new iteration starts at each drop in CPU usage while RAM usage steps up accordingly, but never goes down.

Is this a bug in resource deallocation inside `XGBClassifier` or am I missing something?

Many thanks in advance,
Cheers

hcho3:
Can you try inserting `del best` under the condition `if best_model is None or accuracy > max(accuracies)`, so as to explicitly remove the object?

GuidoBartoli:
Thanks for the suggestion! However I cannot delete `best`, because it is needed in the code following the snippet... maybe you are referring to the temporary `variation`?

I modified the code like this:
```
# setup train_x, train_y, eval_set, eval_metric #
accuracies = []
best = None
for i in range(100):
    variation = xgb.XGBClassifier(random_state=i)
    variation.fit(train_x, train_y, eval_set=eval_set, eval_metric=eval_metric)
    accuracy = 1 - variation.evals_result()['validation_1']['error'][-1]
    if best is None or accuracy > max(accuracies):
        best = variation
    accuracies.append(accuracy)
    del variation
```
But unfortunately RAM usage continues... :(
[...]

Issue #4619, opened 2019-06-28
External memory race condition.


trivialfis:
Hi, @RAMitchell @hcho3 @CodingCat 

I encountered an error on master branch a30176907f5628020f8238464c2f58907ee724d1 by running Python test:
``` bash
pytest -s --fulltrace --verbose ./tests/python/test_updaters.py -k colmaker
```

The error is not deterministic, sometimes it's a segfault, while other times it's some failing checks,  I attached a few of them at the end.  Also note that if I limit XGBoost to single thread all tests run smoothly.  So I believe that it's a race error.  I ran tsan but pytest got stuck at collecting tests, asan is not really helpful for debugging producer/consumer thread model.  Could you provide some suggestions? I can't think of a place to start looking...

```
[02:09:12] WARNING: /home/fis/Workspace/xgb/xgboost/src/gbm/gbtree.cc:95: DANGER AHEAD: You have manually specified `updater` parameter. The `tree_method` parameter will be ignored. Incorrect sequence of updaters will produce undefined behavior. For common uses, we recommend using `tree_method` parameter instead.
[02:09:12] WARNING: /home/fis/Workspace/xgb/xgboost/src/gbm/gbtree.cc:95: DANGER AHEAD: You have manually specified `updater` parameter. The `tree_method` parameter will be ignored. Incorrect sequence of updaters will produce undefined behavior. For common uses, we recommend using `tree_method` parameter instead.
[02:09:12] SparsePage::Writer Finished writing to tmptmp_.sorted.col.page
[02:09:12] SparsePageSource: Finished writing to tmptmp_
[02:09:12] SparsePage::Writer Finished writing to tmptmp_.col.page
terminate called after throwing an instance of 'dmlc::Error'
  what():  [02:09:12] /home/fis/Workspace/xgb/xgboost/dmlc-core/include/dmlc/threadediter.h:189: Check failed: !producer_sig_processed_: 
Stack trace:
  [bt] (0) /home/fis/Workspace/xgb/lib/python3.6/site-packages/xgboost-0.90-py3.6.egg/xgboost/./lib/libxgboost.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x4a) [0x7f9ed8329030]
[...]

PR #4975, opened 2019-10-23
[WIP] Clarify default number of threads.

trivialfis:
While looking into https://github.com/dmlc/xgboost/issues/4843 , I found the configuration of nthreads is not unified nor clear.  This PR clarify the behaviour of `nthreads` and creates a simple function that used through out XGBoost with tests.

### DIFF
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -6,15 +6,18 @@
 #ifndef XGBOOST_COMMON_COMMON_H_
 #define XGBOOST_COMMON_COMMON_H_
 
-#include <xgboost/base.h>
-#include <xgboost/logging.h>
+#include <dmlc/omp.h>
 
+#include <algorithm>
 #include <exception>
 #include <limits>
 #include <type_traits>
 #include <vector>
 #include <string>
 #include <sstream>
+
+#include "xgboost/base.h"
+#include "xgboost/logging.h"
 
 #if defined(__CUDACC__)
 #include <thrust/system/cuda/error.h>
@@ -142,6 +145,15 @@ class Range {
 };
 
 int AllVisibleGPUs();
+
+inline int OmpDefaultThreads(int32_t threads) {
+  if (threads <= 0) {
+    threads = std::max(omp_get_num_procs() / 2, 1);
+  }
+  return threads;
+}
+
+// int
 }  // namespace common
 }  // namespace xgboost
 #endif  // XGBOOST_COMMON_COMMON_H_
--- a/doc/parameter.rst
+++ b/doc/parameter.rst
@@ -34,9 +34,11 @@ General Parameters
     configurations based on heuristics, which is displayed as warning message.
     If there's unexpected behaviour, please try to increase value of verbosity.
 
-* ``nthread`` [default to maximum number of threads available if not set]
+* ``nthread`` [default to half number of processors available if not set]
 
-  - Number of parallel threads used to run XGBoost
+  - Number of parallel threads used to run XGBoost, only used if XGBoost is compiled with
+    OpenMP (default).  Internally uses `omp_get_num_procs` to determine number of
+    processors online on current device.
 
 * ``disable_default_eval_metric`` [default=0]
 
--- a/include/xgboost/generic_parameters.h
+++ b/include/xgboost/generic_parameters.h
@@ -41,8 +41,9 @@ struct GenericParameter : public XGBoostParameter<GenericParameter> {
             "Seed PRNG determnisticly via iterator number, "
             "this option will be switched on automatically on distributed "
             "mode.");
-    DMLC_DECLARE_FIELD(nthread).set_default(0).describe(
-        "Number of threads to use.");
+    DMLC_DECLARE_FIELD(nthread).set_default(0)
+        .set_lower_bound(-1)
+        .describe("Number of threads to use.");
     DMLC_DECLARE_ALIAS(nthread, n_jobs);
 
     DMLC_DECLARE_FIELD(gpu_id)
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -469,8 +469,9 @@ class DMatrix(object):
         feature_types : list, optional
             Set types for features.
         nthread : integer, optional
-            Number of threads to use for loading data from numpy array. If -1,
-            uses maximum threads available on the system.
+            Number of threads to use for loading data from numpy array. If -1
+            or 0, uses half number of processors available on system.
+
         """
         # force into void_p, mac need to pass things in as void_p
         if data is None:
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -73,9 +73,11 @@ class XGBModel(XGBModelBase):
         available.  It's recommended to study this option from parameters
         document.
     n_jobs : int
-        Number of parallel threads used to run xgboost.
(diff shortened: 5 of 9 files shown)


trivialfis:
@hcho3 Now we need to face the inevitable: #4619

hcho3:
@RAMitchell We have an example in dmlc/tvm: https://github.com/dmlc/tvm/blob/cffb4fba03ea582417e2630bd163bca773756af6/src/runtime/threading_backend.cc#L226-L230

trivialfis:
@RAMitchell @hcho3 I see the significance of this change.  Marked as WIP for further discussion.

RAMitchell:
I'm thinking let's take your approach and document clearly. For the majority of cases it is correct and  the best we can do, given no way to reliably detect hyperthreads from c++.

trivialfis:
@RAMitchell In c++ land no, on Linux yes.

trivialfis:
Also, right now (before this PR) we use whatever OMP is set outside of XGBoost context, so user may have a global configuration of threads.  This will change the behaviour.  Now that I think about it, it's evil to "be smart" and add configurations.  Quoting from Python:
``` python
import this
```

```
In the face of ambiguity, refuse the temptation to guess.
There should be one-- and preferably only one --obvious way to do it.
Although that way may not be obvious at first unless you're Dutch.
```

So more inputs are welcomed.  @trams @hcho3 ;-)

RAMitchell:
It would be good to leave some hints on the docs about the evils of hyperthreading and suggest manually configuring nthreads for optimal performance.

trams:
I like the general direction of this pull request. Your code change makes it very clear that by default we are using NUM_OF_LOGICAL_CPU / 2 and I think one should add a comment that it is done because of Hyper Threading (it is not obvious)

Now here is my feedback
1. We run xgboost sometimes in hosted notebook. So the kernel is actually running inside Mesos container with restriction to 2 or 4 CPUs. I always set `n_thread` manually so I do not know what will be default value. Either way, I am not sure that ignoring hyper threading is particularly good idea in containers. I am not sure whether Linux CFS and its share limits (Kuberneris and Mesos uses this mechanism)  aware of hyperthreading. In cloud we also usually have muti processor machines so it makes things even more complex
2. I suggest to add some logging. If `n_thread` is not set get a reasonable default and output it to stdout or stderr so a user can see this picked value
3. I suggest to create a python function `get_default_nthread` to expose this value to a user (to give an ability to troubleshoot) like wrf-python did https://wrf-python.readthedocs.io/en/latest/user_api/generated/wrf.omp_get_num_procs.html

Meanwhile I will try to launch training in this hosted solution and see how it works in `0.90`

trams:
Also I forgot to share two interesting links
1. https://www.openmp.org/spec-html/5.0/openmpsu114.html it seems the function will return number of processors (logical?) so it won't return container size. I do not know how to fetch a container size :(
2. http://mesos.apache.org/documentation/latest/isolators/cgroups-cpu/#effects-on-application-when-using-cfs-bandwidth-limiting CFS Bandwidth limiting has interesting side effects.

trivialfis:
@trams Thanks for your suggestions and pointers.  They are really helpful!  

> Either way, I am not sure that ignoring hyper threading is particularly good idea in containers.

I think none of us has any good idea on how to make the default value good across different platform.  That's why I quoted the Python Easter eggs and thinking just stick with whatever omp is default and let the user choose it explicitly.

> If n_thread is not set get a reasonable default and output it to stdout or stderr so a user can see this picked value

I'm trying to avoid warning, in many issues people choose to ignore warning by setting `verbosity = 1`, which suppresses many truly important warning like not setting `updater` parameter.

> I suggest to create a python function get_default_nthread to expose this value to a user

Aside from this PR, I want to create a global config store for XGBoost like `xgb.config.gpu_id = 0` with some language sugars like in Python:

``` python
with xgb.config(gpu_id=0, nthread=16) as xgb_config:
    xgb.train({'tree_method`: 'gpu_hist`, ... }, dtrain)
```

The configuration maybe available for per-booster and global.  But right now it's not on the top of to-do list.

>https://www.openmp.org/spec-html/5.0/openmpsu114.html it seems the function will return number of processors (logical?) so it won't return container size.

I tried this and it returns number of threads after hyper-threading.

> CFS Bandwidth limiting has interesting side effects.

There are other side effects like memory bandwidths.  Building histogram is very sensitive to this one.  So having more threads with limited memory bandwidth is also harmful.

trams:
> 
> 
> @trams Thanks for your suggestions and pointers. They are really helpful!
> 
> > Either way, I am not sure that ignoring hyper threading is particularly good idea in containers.
> 
> I think none of us has any good idea on how to make the default value good across different platform. That's why I quoted the Python Easter eggs and thinking just stick with whatever omp is default and let the user choose it explicitly.
Yes. I like this principle. Explicit is better then implicit
> 
> > If n_thread is not set get a reasonable default and output it to stdout or stderr so a user can see this picked value
> 
> I'm trying to avoid warning, in many issues people choose to ignore warning by setting `verbosity = 0`, which suppresses many truly important warnings like not to set the `updater` parameter.
Fair enough but I still strongly suggest
1. output it to debug log at least
2. Expose this in xgboost-spark (or even fail if it is not explicitly set in xgboost-spark) as Spark Accumulator or Spark Config so one can see it in UI
> 
> > I suggest to create a python function get_default_nthread to expose this value to a user
> 
> Aside from this PR, I want to create a global config store for XGBoost like `xgb.config.gpu_id = 0` with some language sugars like in Python:
> 
> ```python
> with xgb.config(gpu_id=0, nthread=16) as xgb_config:
>     dtrain = xgb.DMatrix(X_train, label=y_train)
>     xgb.train({'tree_method`: 'gpu_hist', ... }, dtrain)
> ```
> 
> The configuration maybe available for per-booster and global. But right now it's not on the top of to-do list.
> 
> > https://www.openmp.org/spec-html/5.0/openmpsu114.html it seems the function will return number of processors (logical?) so it won't return container size.
> 
> I tried this and it returns number of threads after hyper-threading.
I think I failed to communicate my concerned.
My concern is not whether the function returns number of physical processors or logical (hyper threads). My concern how does this function will behave inside the Mesos container (or any other container).

Let me give you some context.

In my company we have a hosted notebook solution which allows to launch a notebook inside the cloud. Behind the scene it launches a kernel inside a Mesos container (with 8 cpus by default). From a user perspective amount of logical cores one has is the size of the container (so it will be 8) but ... as far as I read Mesos documentation CPU limit implemented using CFS Bandwidth limiting which results in OMP thinking that it have access to all logical cores of the machine (which is 72 in our case) instead of actual amount of "available" core in the container (8 in our case)
This will lead in increased overhead for handling all these extra threads for nothing

This Mesos container thing already bit me few times. That's why I strongly suggest to either have some logging or expose the default in python library so a hypothetical user can just look it up and check whether it is sane
> 
> > CFS Bandwidth limiting has interesting side effects.
> 
> There are other side effects like memory bandwidths. Building histogram is very sensitive to this one. So having more threads with limited memory bandwidth is also harmful.

src/common/common.h
```diff
@@ -142,6 +145,15 @@ class Range {
 };
 
 int AllVisibleGPUs();
+
+inline int OmpDefaultThreads(int32_t threads) {
```
trams: I suggest to make it not inline function and expose it in C API and Python API

PR #4976, opened 2019-10-23
[CI] Upload nightly builds to S3

### LINKED ISSUE #4813: Nightly releases
I have had requests from some users for access to nightly releases from our CI. I believe it would simplify the workflow for many groups. This thread is to track the issue and discuss implementation.

As far as I know we do have packages being produced on master branch, I'm not sure if they are stored in any kind of persistent storage. I would imagine it would just be a case of archiving binaries on S3 and providing a link in documentation?

hcho3:
Closes #4813.

Expected usage: pip3 install https:\/\/s3-us-west-2.amazonaws.com/xgboost-nightly-builds/xgboost-1.0.0+[commit hash]-py2.py3-none-manylinux1_x86_64.whl

Listing is available at https://s3-us-west-2.amazonaws.com/xgboost-nightly-builds/list.html

cc @randerzander @JohnZed

### DIFF
--- a/tests/ci_build/rename_whl.py
+++ b/tests/ci_build/rename_whl.py
@@ -1,0 +1,45 @@
+import sys
+import os
+from contextlib import contextmanager
+
+def normpath(path):
+    """Normalize UNIX path to a native path."""
+    normalized = os.path.join(*path.split("/"))
+    if os.path.isabs(path):
+        return os.path.abspath("/") + normalized
+    else:
+        return normalized
+
+@contextmanager
+def cd(path):
+    path = normpath(path)
+    cwd = os.getcwd()
+    os.chdir(path)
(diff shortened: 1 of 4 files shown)


tests/ci_build/rename_whl.py
```diff
@@ -0,0 +1,45 @@
+import sys
+import os
+from contextlib import contextmanager
+
+def normpath(path):
```
trivialfis: Can this be done by https://docs.python.org/3/library/os.path.html#os.path.normpath ?

hcho3:
Note: artifacts from pull requests are pull in a separate directory, e.g. `PR-4976`. Only artifacts made with the `master` branch will be uploaded at the root directory.

PR #4984, opened 2019-10-25
Fix dart usegpu.

trivialfis:


trivialfis:
@hcho3. Is there any way I can look into the stalled Windows executor?  Like whether it's truly failed allocation or some plugins gone wrong?

hcho3:
@trivialfis The EC2 plugin in Jenkins is to blame here. The plugin lost connection to the agent but somehow the agent still counted toward the instance cap.

trivialfis:
Please let me know how do I help.

hcho3:
@trivialfis For now, you can log into the EC2 console (I gave you credential) and watch if there is a P2 instance running that's not showing up in https://xgboost-ci.net/computer/. If so, you should terminate it so that a new instance can be launched.

In the long run, we need something like https://github.com/apache/incubator-mxnet-ci/tree/master/services/jenkins-autoscaling, a Lambda function that queries the job queue in the Jenkins server and provisions EC2 workers. There are lots of moving parts here, but I'll try to get it done by end of this year.

hcho3:
@trivialfis https://github.com/pycontribs/jenkinsapi seems useful too

trivialfis:
@hcho3 Huge thanks!  That will be great.  Seems to be quite a project by itself.  I need to learn some basic about operating Jenkins.

hcho3:
@trivialfis I decided to take small steps instead of making a big script in one go, since small scripts are much easier to test and debug.

Issue #2776, opened 2017-10-07
Custom multiclass objectives in R


tvatter:
Hi everyone,

I am wondering whether it is possible to create custom multiclass objectives for xgboost in R.

Below is an MWE:

```
rm(list=ls())
require(xgboost)
set.seed(0)

# Simulate some dummy data
n <- 1e2
num_features <- 2
num_class <- 3
y <- apply(rmultinom(n, 1, rep(1, num_class)), 2, function(yy) which(yy != 0))
x <- matrix(rnorm(n*num_features), n, num_features)

# Format for xgboost
dtrain <- xgb.DMatrix(x, label = y)

# A dummy custom objective to understand what is going on
dummy_obj <- function(preds, dtrain) {
  browser()
}

# Feed to xgboost
params_dummy <- list(objective = dummy_obj, 
                     num_class = num_class)
model_dummy <- xgb.train(params_dummy, data = dtrain, nrounds = 1)

```

On my computer, executing the code above allowed me to do:

```
Called from: obj(pred, dtrain)
Browse[1]> preds
  [1] 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
 [40] 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
 [79] 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
Browse[1]> Q
```

Having read the source code of [multiclass_obj.cc](https://github.com/dmlc/xgboost/blob/master/src/objective/multiclass_obj.cc), I was expecting for preds a vector of length n x num_class, or a matrix with n rows and num_class columns, or something like that.

So, my question is: am I missing something obvious (very likely) or is it currently not possible to implement custom multiclass objectives?

Below are my environment informations, although I doubt they will be useful in this case.

---------
Operating System:

OS X 10.12.6

Compiler:

Not relevant

Package used (python/R/jvm/C++):

R

`sessionInfo()`:

R version 3.4.1 (2017-06-30)
Platform: x86_64-apple-darwin16.6.0 (64-bit)
Running under: macOS Sierra 10.12.6

Matrix products: default
[...]

Issue #4288, opened 2019-03-22
Custom multiclass objective receives transformed predictions


jeffzi:
Hi,

I'm having issues with custom multiclass objective function receiving transformed predictions. The problem has already been raised in issue https://github.com/dmlc/xgboost/issues/2776

I believe the objective function should expect `preds` as a (N, K) array, N = #data, K = #classes, but it's not the case by default in both python and R. The previous issue points to https://github.com/dmlc/xgboost/blob/263e2038e94fe8a0bb5e86ba3aef4e1feed150f1/src/objective/multiclass_obj.cu#L154-L157

In python, I managed to get the correct `preds` shape by adding `'objective': 'multi:softprob'` to the parameters on top of my custom objective function. Here is a MWE:

```python
import numpy as np
from sklearn import datasets
from sklearn.preprocessing import OneHotEncoder
import xgboost as xgb

iris = datasets.load_iris()
X, y = iris.data, iris.target
dtrain = xgb.DMatrix(X, label=y)

# builtin
params = {'objective': 'multi:softprob', 'num_class': 3}
model_builtin = xgb.train(params, dtrain, num_boost_round = 1)
preds_builtin = model_builtin.predict(dtrain)

# custom
def obj(preds, dtrain):
    labels = dtrain.get_label().reshape(-1, 1)
    labels = OneHotEncoder(sparse=False, categories='auto').fit_transform(labels)
    grad = preds - labels
    hess = 2.0 * preds * (1.0 - preds)
    return grad.flatten(), hess.flatten()

params = {'objective': 'multi:softprob', 'num_class': 3}
model_custom = xgb.train(params, dtrain, num_boost_round = 1, obj = obj)
preds_custom = model_custom.predict(dtrain)

# assert approaches give same results
assert np.sum(np.abs(preds_custom - preds_builtin)) == 0

## fails without objective 'multi:softprob because preds.shape == (150,)
#params = {'num_class': len(np.unique(y))}
#model = xgb.train(params, dtrain, num_boost_round = 1, obj = obj)

import sinfo
sinfo.sinfo()
#> -----
#> numpy     	1.16.1
#> sklearn   	0.20.3
#> xgboost   	0.82
#> -----
#> Python 3.7.2 (default, Feb 12 2019, 08:15:36) [Clang 10.0.0 (clang-1000.11.45.5)]
#> Darwin-18.2.0-x86_64-i386-64bit
#> 4 logical CPU cores, i386
```
[...]

PR #4996, opened 2019-10-30
Custom multiclass objectives in R

Someone894:
Added workaround described in https://github.com/dmlc/xgboost/issues/2776

with this change the https://github.com/pmontman/fforma package works fine.


trivialfis:
Could you please help taking a look @pommedeterresautee ?

pommedeterresautee:
Tks, I check today or tomorrow.

pommedeterresautee:
@Someone894 Hi, I have made few checks and it seems to not break simple cases (I was worried about the reshaping).
However, I am wondering how `outputmargin` set to TRUE will never impact in some way the training? (in some other cases that your own custom objective).
I am asking because the function `xgb.iter.update` is obviously central and I am under the impression you have already performed a deep work for `fforma`, so may be you can help me in better understand that point.

Someone894:
The `FFORMA`-System needs the custom objective in order to work. So @pmontman created a custom-version (https://github.com/pmontman/customxgboost), which is not maintained anymore. This got me worried about the old code causing some trouble in the future. So I talked to him in order to understand what his changes to the basic XGBoost-package where. Then I implemented these changes and tested the `FFORMA`-System with them and it worked fine. but to be honest I did not understand the changes in the  XGBoost-package since it feels really complicated  to me.

pommedeterresautee:
Thank you a lot for these information.
Basically the function `xgb.iter.update` is quite important.
It is called at each iteration of the boosting process.
https://github.com/dmlc/xgboost/blob/5e97de6a4186cd616bda3387c8cf9c5ec1904189/R-package/R/xgb.train.R#L331
Its purpose is to apply the objective to the predictions of the model being built and compute the loss.
https://github.com/dmlc/xgboost/blob/9536a7e0d216bc3f0487b91ede4597acc391272c/R-package/R/utils.R#L149
This PR modifies the predictions by setting `outputmargin` to `True`, basically the predictions are now the margin directly computed by XGBoost (the original untransformed predictions).
My feeling is that in several cases this will have unwanted effect on the objective, and the quality of the model.

Would it be possible for you to contact @pmontman to see if there are reasons why such change would be Ok?

Otherwise, I am thinking that another approach may be more appropriate. 
May be another hook function or some information provided through parameters of `xgb.iter.update`  making possible for third party objectives to retrieve the margin.

Regarding the reshape thing, it should be easy to perform in the objective itself.

jeffzi:
I encountered the same issue several months ago while looking at FFORMA as well (see #4288).

I had to make the same modification in xgb.iter.eval to fix xgb.cv as well.
https://github.com/dmlc/xgboost/blob/e089e16e3da6637c5374eb1e2e0a3b9fd92907ba/R-package/R/utils.R#L175

pommedeterresautee:
Thanks @jeffzi . I understand the purpose of the change, my question is how this change may not have any impact on other use cases? because here we are wondering if it makes sense to apply the change to the main library.
I may miss something.

Someone894:
I tried to reach @pmontman to also have a look at this topic, but nothing happened, as you can see.

trivialfis:
@Someone894 @pommedeterresautee It make sense to me to use raw prediction as objective function should handle this transformation itself.  WDYT?

jeffzi:
It makes sense to me to use raw prediction, since it reflects what happens when no custom objective is given.

https://github.com/dmlc/xgboost/blob/ad4a1c732cfad6d6dc5be012145a2dac37413cc7/R-package/R/utils.R#L145-L151

`UpdateOneIter`  uses raw prediction:

https://github.com/dmlc/xgboost/blob/ad4a1c732cfad6d6dc5be012145a2dac37413cc7/src/learner.cc#L561-L567

seangupta:
Hi, I've encountered this problem as well. Will there be an update soon? Thanks.

trivialfis:
I incline to merge this PR(or new one with tests).  @hcho3 WDYT?

PR #5004, opened 2019-11-01
- ndcg ltr implementation on gpu

sriramch:
- this is a follow-up to the pairwise ltr implementation
  - one more is remaining after this (which will follow this pr)
  - i'll work on getting the performance numbers


hcho3:
@sriramch Which reference did you use for the ranking algorithm? I was looking at [Chris J.C. Burges's paper](https://www.microsoft.com/en-us/research/publication/from-ranknet-to-lambdarank-to-lambdamart-an-overview/) and I personally find it hard to understand. I'm wondering if you have another reference I can look at.

sriramch:
> @sriramch Which reference did you use for the ranking algorithm? I was looking at [Chris J.C. Burges's paper](https://www.microsoft.com/en-us/research/publication/from-ranknet-to-lambdarank-to-lambdamart-an-overview/) and I personally find it hard to understand. I'm wondering if you have another reference I can look at.

@hcho3 - the primary ones i looked at are the following (for pairwise):

http://wwwconference.org/proceedings/www2011/proceedings/p377.pdf
http://159.226.40.238/~junxu/publications/IRJ2010_LETOR.pdf

that said, i'm primarily looking at the cpu implementation as a reference implementation to model this.

i haven't come across the paper you referenced.

sriramch:
# performance numbers
## test environment
   * 1 socket
   * 6 cores/socket
   * 2 threads/core
   * 80 gb system memory
   * v100 gpu

## test
   * tree method is gpu_hist
   * uses all cpu threads
   * 100 trees
   * objective is `rank:ndcg`

## small [mslr](https://www.microsoft.com/en-us/research/project/mslr/) dataset
   * ~ 3.6m training instances
   * distributed over ~ 10k groups
   * consuming 3.9 gb disk space
   * master does the gradient computation on cpu
   * fork does the gradient computation on gpu - i.e. this pr

| branch | peak_memory_usage(mb) | time(s) / map-metric | time(s) / auc-metric | time(s) / rmse-metric |
| :---: | :---: | :---: |  :---: |  :---: |
| master | 2163 | 9.75s / .642747 | 9.82s / 1.104549 | 9.97s / 1.007689 |
| fork | 2177 | 0.61s / .642742 | 0.61s / 1.109999 | 0.62s / 1.004864 |

## large [mslr](https://www.microsoft.com/en-us/research/project/mslr/) dataset
   * ~ 11.3m training instances
   * distributed over ~ 95k groups
   * consuming 13 gb disk space
   * master does the gradient computation on cpu
   * fork does the gradient computation on gpu - i.e. this pr

| branch | peak_memory_usage(mb) | time(s) / map-metric | time(s) / auc-metric | time(s) / rmse-metric |
| :---: | :---: | :---: |  :---: |  :---: |
| master | 4328 | 28.63s / .651674 | 28.2s / 1.084873 | 28.71 / 1.037019 |
| fork | 4371 | 1.62s / .650833 | 1.62s / 1.081308 | 1.62s / 1.035145 |

sriramch:
@hcho3 - is the [last](https://travis-ci.org/dmlc/xgboost/jobs/606125015) test failure a transient error that plausibly could resolve itself with a restart?

the pre-lint/clang [commit](https://travis-ci.org/dmlc/xgboost/jobs/606099557) passed this stage, and thus the question.

hcho3:
@sriramch I restarted it. We have a backlog item to remove Travis CI: #4498

sriramch:
> @sriramch I restarted it. We have a backlog item to remove Travis CI: #4498

@hcho3 thanks for your help!

sriramch:

src/objective/rank_obj.cu
```diff
@@ -96,16 +274,154 @@ struct PairwiseLambdaWeightComputer {
...
+  // dindexable_sorted_preds_pos_ptr_:  8 0 9 1 7 2 6 3 4 5
+  // at indices:                        0 1 2 3 4 5 6 7 8 9
+  // Basically, swap the previous 2 arrays, sort the indices and reorder positions
+  // for an O(1) lookup using the position where the sorted label exists
+  void CreateIndexableSortedPredictionPositions(const uint32_t *dsorted_preds_pos) {
+    int device_id = -1;
+    dh::safe_cuda(cudaGetDevice(&device_id));
+    uint32_t *dindexable_sorted_preds_pos_ptr = dindexable_sorted_preds_pos_ptr_;
+    // Sort the positions (as indices), and group its indices as sorted prediction positions
+    dh::LaunchN(device_id, dindexable_sorted_preds_pos_->size(),
```
RAMitchell: I think this is just a thrust scatter operation. Prefer thrust for readability. 

Not sure if you have used permutation iterators before, but they are also useful for these types of situations. You can create a 'virtual' permutation on an array without actually moving data around.

tests/cpp/objective/test_ranking_obj.cc
```diff
@@ -76,4 +76,33 @@ TEST(Objective, DeclareUnifiedTest(PairwiseRankingGPairSameLabels)) {
   ASSERT_NO_THROW(obj->DefaultEvalMetric());
 }
 
+TEST(Objective, DeclareUnifiedTest(NDCGRankingGPair)) {
```
RAMitchell: These tests look a little sparse considering the amount of changes. e.g. what if problems occur only when more than one GPU block is launched?

Consider some more c++ tests and integration tests in python (or scala if you prefer).

src/objective/rank_obj.cu
```diff
@@ -483,6 +723,11 @@ class SortedLabelList {
       const GradientPair in_neg_gpair(-g * weight, 2.0f * weight * h);
       dh::AtomicAddGpair(&out_gpair[original_pos[neg_idx]], in_neg_gpair);
     });
+
+    // Wait until the computations done by the kernel is complete
+    dh::safe_cuda(cudaStreamSynchronize(nullptr));
+
+    weight_computer.ReleaseResources();
```
RAMitchell: `weight_computer.ReleaseResources();`

I am trying to move away from having objects in half configured states in xgboost. For example if you use an object in the wrong way or at the wrong time you would get a seg fault or some cryptic change of behaviour.

Instead we want to create an object on the stack when we need it, guarantee that it is in a usable state after construction, then let it be destroyed. This was previously hard to do without performance penalties because GPU memory allocation is expensive, but I think it is possible with caching allocators.

src/objective/rank_obj.cu
```diff
@@ -46,6 +46,184 @@ struct LambdaRankParam : public XGBoostParameter<LambdaRankParam> {
...
+  // Original position of the items before they are sorted descendingly within its groups
+  dh::caching_device_vector<uint32_t> doriginal_pos_;
+
+  // Segments within the original list that delineates the different groups
+  dh::caching_device_vector<uint32_t> group_segments_;
+
+  // Need this on the device as it is used in the kernels
+  dh::caching_device_vector<uint32_t> dgroups_;       // Group information on device
+
+  dh::XGBCachingDeviceAllocator<char> alloc_;         // Allocator to be used by sort for managing
```
RAMitchell: Just a note that this allocator does not necessarily need to be a member variable, it can be created and destroyed whenever you need without performance penalty.

trivialfis:
Will look into this tomorrow.

trivialfis:
@hcho3 

> Which reference did you use for the ranking algorithm? 

I read this: https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/tr-2007-40.pdf to get a sense for what's learning to rank, then the paper you referred.  Hope that helps.

trivialfis:
Wrong link.  Fixed.  Sorry.  Some slides are [here](https://wwwconference.org/www2009/pdf/T7A-LEARNING%20TO%20RANK%20TUTORIAL.pdf) if you are not into the whole book.

src/objective/rank_obj.cu
sriramch: please look. i have introduced an abstraction that lets the types be reference counted in fcb55e1. this will let the objects themselves keep track of how many copies they have.

src/objective/rank_obj.cu
sriramch: done

sriramch:
@hcho3 is there a way to get the nvidia driver version on windows? i looked at the build logs, but couldn't (could only find nvcc compiler version). typically, `nvidia-smi` provides this information.

src/objective/rank_obj.cu
trivialfis: @sriramch That's impressive.  Never thought about adding a intrusive ptr in kernel.

trivialfis:
@sriramch You can add a line to https://github.com/dmlc/xgboost/blob/b29b8c2f34cd0be58e8cadd9bb388e85898d4ae4/Jenkinsfile-win64#L73

sriramch:
> @sriramch You can add a line to
> 
> https://github.com/dmlc/xgboost/blob/b29b8c2f34cd0be58e8cadd9bb388e85898d4ae4/Jenkinsfile-win64#L73

thanks @trivialfis for your tip. i'll try now...

RAMitchell:
I think I see what you are doing with RefCountable. The problem is that we have a memory owning class on the host that we would like to use in a kernel, but the device vector objects cannot be shallow copied and used in the kernel unless we construct them as pointers. My solution to this is normally:

Create a nested class inside NDCGLambdaWeightComputer, call it NDCGLambdaWeightComputerDeviceAccessor. This subclass contains pointers to the device vectors in the parent class and methods to be called inside the kernel. This class is trivially copyable.

Before launching the kernel call
`auto accessor = weight_computer.GetAccessor()`
And copy this into the kernel by lambda capture.

So basically we create thin wrapper classes for use in kernels with some clear semantics on how they are related to memory owning classes that exist on the host. 

WDYT?

sriramch:
> So basically we create thin wrapper classes for use in kernels with some clear semantics on how they are related to memory owning classes that exist on the host.
> 
> WDYT?

[sc] sounds good to me. using managed memory needs a little bit more work to be used seamlessly x-platform, as i realized after reading the cuda documentation, that data movement between host and device isn't as transparent on windows as it is on linux's. the unified memory model on windows is pre-6.x, even when its compute capability is 6.x or greater. the upshot is that periodic `cudaDeviceSynchronize` may be required to move data between the kernel and host.

sriramch:
@RAMitchell i have addressed your comments. please look.

the build failure seems to be a transient error, which i presume will go away with a restart of the build pipeline (or that specific piece of that pipeline).

@trivialfis thanks for your review.

trivialfis:
Restarted the build.

src/objective/rank_obj.cu
```diff
@@ -46,6 +46,185 @@ struct LambdaRankParam : public XGBoostParameter<LambdaRankParam> {
...
+    group_segments_.resize(num_elems);
+
+    dgroups_ = groups;
+
+    // Launch a kernel that populates the segment information for the different groups
+    uint32_t *gsegs = group_segments_.data().get();
+    const uint32_t *dgroups = dgroups_.data().get();
+    uint32_t ngroups = dgroups_.size();
+    int device_id = -1;
+    dh::safe_cuda(cudaGetDevice(&device_id));
```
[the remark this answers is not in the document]
sriramch: wouldn't learner configure the generic parameter, which is then passed to the objective function when it is created? further, this method is invoked on a cpu thread that already sets the device by extracting it from the generic parameter in `ComputeGradientsOnGPU`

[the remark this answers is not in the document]
trivialfis: Yes.  But I remember the learner class doesn't set device with cuda as it's a cc file.  So a little worry.  I guess it's fine to recover it from thread local cuda state if you are sure it's set.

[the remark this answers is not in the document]
RAMitchell: Can one of you please confirm this before we merge, otherwise I'm happy.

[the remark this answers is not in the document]
sriramch: @RAMitchell - yes, this works as expected. i explicitly tried setting the gpu_id to a non-zero value on a multi gpu node and was able to see the right gpu being used

Issue #4915, opened 2019-10-07
Unnecessary memory allocations for `hist` method?


thvasilo:
The hist method makes heavy use of a [HistCollection](https://github.com/dmlc/xgboost/blob/f0064c07ab70f1c770bb0e9eb64fe85afb3a3d73/src/common/hist_util.h#L440) container that is supposed to maintain for each node (leaf) id a collection of histograms.

The backing data structure is [a vector of vectors of `GradStatHist`](https://github.com/dmlc/xgboost/blob/f0064c07ab70f1c770bb0e9eb64fe85afb3a3d73/src/common/hist_util.h#L476) objects.

Now, whenever we add a new node id to this collection, we resize this vector to hold enough elements so that _the highest nid provide_ is a valid index.

From some print debugging, printing out nodeid [here](https://github.com/dmlc/xgboost/blob/f0064c07ab70f1c770bb0e9eb64fe85afb3a3d73/src/tree/updater_quantile_hist.cc#L615) (printing the node id of every node being synced), it seems like node ids _are non-contiguous_. For example they might be [1,3,5,7,9,11,13]. However whenever we expand the size of the container, [it looks like](https://github.com/dmlc/xgboost/blob/f0064c07ab70f1c770bb0e9eb64fe85afb3a3d73/src/common/hist_util.h#L462) we allocate memory for _all_ node ids, regardless of whether they represent an actual node or not. Is this the case, or are node ids actually contiguous, thereby it's necessary to always pre-populate every element?

As far as I understand the 2D data vector has dimensions `<MAX_NODE_ID> x <NBINS>`, where [uint32_t nbins = gmat.cut.Ptrs().back();](https://github.com/dmlc/xgboost/blob/f0064c07ab70f1c770bb0e9eb64fe85afb3a3d73/src/tree/updater_quantile_hist.cc#L536). For example for the `rcv1` dataset that has ~42k features creates 748,378 bins per node, which each correspond I think to two float values per bin. So every time we add a redundant item in the `_hist` vector we are wasting num_bins * 2 float values, which, when we have say millions of features can be very significant.

So my impression is that we are currently using twice as much memory as is necessary, because we initialize every vector in the 2D vector of histogram values with `nbins` values, regardless of the fact that half of the node ids are not used (?).

[...]

Issue #4952, opened 2019-10-16
[RFC] 1Tb benchmark


trams:
Hello nice people,

I would like to discuss with you building a big distributed benchmark so we can assess GPU/CPU performance. During 2019 Q4 I will have few days to pursue this topic and I would like to hear your opinion.

Is it useful? Will you run it from time to time?

Here is a gist of my idea.
I want to build a reproducible benchmark which uses Criteo 1Tb machine learning dataset
(more about it here https://labs.criteo.com/2013/12/download-terabyte-click-logs-2/)

The dataset has 39 features: 13 quantative and 26 categorical features
So for milestone 1 I would just drop all categorical features and use it as a training dataset.
I sized it recently and I've got ~100Gb if I store it in memory as Spark Dataframe which is big enough to test a distributed scenario but small enough to test it on 1 machine too (especially if downsampled)

As a milestone 2 (work in progress) I wanted to apply a simple version of mean encoding (something like https://towardsdatascience.com/why-you-should-try-mean-encoding-17057262cd0). I have a working prototype. I think I will just apply the mean transformation once and publish the resulting dataset somewhere.

This should give us a bigger version of a benchmark. I expect it to be ~300-400Gb

What do you think? Do you want to contribute in any way? Will you relaunch it from time to time (from release to release for example, or every month)

Do you have any cloud budget for this kind of things or should I also try to find a corporate sponsor?

trivialfis:
@rongou

trivialfis:
@wbo4958

hcho3:
> Do you have any cloud budget for this kind of things or should I also try to find a corporate sponsor?

I'm in the process of enabling finer grained financial control over the CI server. This is also necessary to enable regression tests, to ensure consistency of training performance between releases.

trivialfis:
@hcho3 We are currently trying to scale GPU algorithm with spark to tackle this dataset.  I don't think it's possible to run it on Jenkins.  Maybe we (nvidia) can run it on a regular basis.

hcho3:
@trivialfis It would be awesome if NVIDIA can host this benchmark.

trivialfis:
@hcho3 Not really. That's not my decision to make.  But I can run it when needed.

hcho3:
Well, if you are running it using NVIDIA's resources, then from my perspective NVIDIA is "hosting" the benchmark. That's what I intended.

trivialfis:
Yup.  Definitely.  :-)

[...]

PR #5005, opened 2019-11-03
Use global leave index, small clean up for cpu hist.

trivialfis:
## Add LeaveIndexCache in TreeUpdater.

Now for both CPU Hist and GPU Hist, the leave cache is stored in gbtree and uses format similar to
original GPU Hist implementation.  Now the cache can be reused by other components that requires the leaf mapping.  As I couldn't find similar structure in approx and exact, so only hist method is implemented.

- Remove node_id from Elem.

In order to align with GPU Hist, also have small drop of memory usage.

- Remove row set collection.

Use LeaveIndexCache instead.

## Small clean up for CPU Hist.

- Remove pruner in CPU Hist.

Use pre-pruning instead of post-pruning.  Possibly fixing #4915 .

- Remove PerfMonitor.

Use common::Monitor instead.

- Remove duplicated dense partition.
- Remove unused variables.

### DIFF
--- a/include/xgboost/tree_updater.h
+++ b/include/xgboost/tree_updater.h
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2014-2019 by Contributors
+ * Copyright 2014-2020 by Contributors
  * \file tree_updater.h
  * \brief General primitive for tree learning,
  *   Updating a collection of trees given the information.
@@ -34,7 +34,7 @@ class TreeUpdater : public Configurable {
 
  public:
   /*! \brief virtual destructor */
-  virtual ~TreeUpdater() = default;
+  ~TreeUpdater() override = default;
   /*!
    * \brief Initialize the updater with given arguments.
    * \param args arguments to the objective function.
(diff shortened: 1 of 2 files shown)


src/tree/gpu_hist/row_partitioner.cu
```diff
@@ -60,21 +63,29 @@ void RowPartitioner::SortPosition(common::Span<TreePositionT> position,
...
+
   ridx_b.resize(num_rows);
   position_a.resize(num_rows);
   position_b.resize(num_rows);
-  ridx = dh::DoubleBuffer<RowIndexT>{&ridx_a, &ridx_b};
+
+  common::Span<RowIndexT> s_ridx_buffer = common::Span<RowIndexT>{
+    ridx_b.data().get(), ridx_b.size()};
+  common::Span<RowIndexT> s_ridx_a = leaf_index_->RowIndices().DeviceSpan();
+  ridx = dh::DoubleBuffer<RowIndexT>{s_ridx_a, s_ridx_buffer};
```
RAMitchell: So `leaf_index_` now provides the storage for RowPartitioner?

include/xgboost/tree_updater.h
```diff
@@ -66,11 +65,124 @@ class TreeUpdater {
...
   virtual char const* Name() const = 0;
 
+  class LeaveIndexCache;
   /*!
    * \brief Create a tree updater given name
    * \param name Name of the tree updater.
    */
-  static TreeUpdater* Create(const std::string& name, GenericParameter const* tparam);
+  static TreeUpdater* Create(const std::string& name,
+                             std::shared_ptr<LeaveIndexCache> cache,
```
RAMitchell: `LeaveIndexCache` should be `LeafIndexCache`

trivialfis:
@RAMitchell As you can see the storage is shared by different components to eliminate the extra O(n) memory copy.  I can try to stored it in tree updater only and add a method in tree updater for asking it explicitly.

trivialfis:
@hcho3 Those like `Create..TasksBatch` has to go.  I can't allocate time to do a full clean up for CPU Hist.   I might prefer maintainability over performance, after banging my head on the monitor, I started to think about reverting the CPU optimization.  WDYT?

hcho3:
@trivialfis Yes, it’s quite bad. Also there’s the issue with per node syncing in distributed training: https://github.com/dmlc/xgboost/issues/4679

hcho3:
@trivialfis Please review #5008, which hopefully should return the 'hist' updater to a legible state.

RAMitchell:
O(n) memory copy once per boosting iteration is not significant. I dont think you need to optimise this. Maybe for memory savings it could be justified? After looking at this more, I think you need to implement the objective function first, even if its slow and has redundant computation and then move forward with this when you have benchmark numbers.

trivialfis:
> Maybe for memory savings it could be justified?

@RAMitchell Yup.  That's what I'm concerned with, not computation performance.  I will try to add a better interface so things get more explicit instead of a hidden variable.  But I would generally try to avoid any extra memory allocation, as many of us are trying to squeeze as much more room as possible for datasets like #4952 .

hcho3:
Can we close this? I don't think it's still relevant to the current codebase.

trivialfis:
It's still relevant, I will close it once I have a replacement PR

trivialfis:
Replaced by https://github.com/dmlc/xgboost/pull/7812 .

Issue #3826, opened 2018-10-24
Leafs with no training events in them


Denisevi4:
Very rarely I incur this instance of having leafs in the model that have cover=0.

Here is a dump of such a tree (sorry. don't know how to format it)

0:[C19<-0.00326049794] yes=1,no=2,missing=1,gain=6.68777793e-05,cover=205821
	1:[C49<13.9841671] yes=3,no=4,missing=3,gain=0.000157496921,cover=20660
		3:[C104<1.53036571] yes=7,no=8,missing=7,gain=0.000128577725,cover=18218
			7:[C10<39774.3789] yes=15,no=16,missing=15,gain=0.000146228616,cover=16352
				15:leaf=6.29458373e-05,cover=8496
				16:leaf=-0,cover=7856
			8:[C147<18.836647] yes=17,no=18,missing=17,gain=0.000132655929,cover=1866
				17:leaf=-0,cover=1138
				18:leaf=-0.000121151919,cover=728
		4:[C140<7.49999973e-09] yes=9,no=10,missing=9,gain=0.000101037891,cover=2442
			9:leaf=-0,cover=1434
			10:[C49<13.9841671] yes=19,no=20,missing=19,gain=6.39266946e-05,cover=1008
				**19:leaf=-nan,cover=0**
				20:leaf=-9.8133547e-05,cover=1008
	2:[C71<3.32500008e-07] yes=5,no=6,missing=5,gain=9.50862304e-05,cover=185161
		5:[C127<0.237853229] yes=11,no=12,missing=11,gain=0.000120236931,cover=180340
			11:[C102<1.97036135] yes=21,no=22,missing=21,gain=0.000182247895,cover=173194
				21:leaf=-0,cover=160209
				22:leaf=-3.38961036e-05,cover=12985
			12:[C146<6.92263126] yes=23,no=24,missing=23,gain=0.000333737611,cover=7146
				23:leaf=-3.61185957e-05,cover=1833
				24:leaf=7.79525362e-05,cover=5313
		6:[C16<0.0269251391] yes=13,no=14,missing=13,gain=0.000221637121,cover=4821
			13:[C115<0.623490334] yes=25,no=26,missing=25,gain=2.03016607e-05,cover=2146
				25:leaf=-0,cover=1280
				26:leaf=4.27486339e-05,cover=866
			14:[C13<0.00162636698] yes=27,no=28,missing=27,gain=0.000202318741,cover=2675
[...]

PR #5010, opened 2019-11-04
attempt to fix repeated split and 0 cover nodes

cowden:
This patch addresses the issue raised in https://github.com/dmlc/xgboost/issues/3826

I was able to determine that this situation arises when there is "spike" of data at a single floating point number having a slight increase in gain to split away.  The lack of numerical precision causes the split point calculation to fail to resolve between the observed values.  Then, the "<" logic of scoring the tree is different from the sort and split logic of the search.  

In this patch, I've simply tried to catch the loss of resolution and propose a different split in that situation.  Testing this on the example provided in the issue above, the split leading to the 0-cover nodes is shifted slightly (the tree structure below is similar to the previous tree), and the nodes on the other side of the tree are left unaffected.  See the new split values below.

0:C2 < -0.00652099587
1: **C1 < 27.9683361**
3:C0 < 526.223633
7:C0 < 2.30000001e-08
4:C0 < 1.49999995e-08
10:C2 < -0.00678067794
2:C1 < 23.9683456
5:C1 < 19.1684093
11:C0 < 9.99999972e-10
12:C2 < 0.006448247
6:C1 < 26.7018356
13:C2 < 0.0112791462
14:C0 < 0.524977624

For reference, here is the original tree structure:
0:C2 < -0.00652099587
1: **C1 < 27.9683342**
2:C1 < 23.9683456
3:C2 < -0.0164765697
4:C0 < 1.49999995e-08
5:C1 < 19.1684074
6:C1 < 26.7018356
8:C0 < 2.30000001e-08
10: **C1 < 27.9683342**
11:C0 < 9.99999972e-10
12:C2 < 0.006448247
13:C2 < 0.0112791462
14:C0 < 0.524977624

### DIFF
--- a/src/tree/updater_colmaker.cc
+++ b/src/tree/updater_colmaker.cc
@@ -413,7 +413,7 @@ class ColMaker: public TreeUpdater {
               if ( proposed_split == fvalue ) {
                 e.best.Update(loss_chg, fid, e.last_fvalue,
                             d_step == -1, c, e.stats);
-              } else { 
+              } else {
                 e.best.Update(loss_chg, fid, proposed_split,
                             d_step == -1, c, e.stats);
               }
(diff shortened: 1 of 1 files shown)


trivialfis:
Thanks @cowden !  Closes  #3826 .

PR #5016, opened 2019-11-06
Improve operation efficiency for single predict

KerryJi:
Improve operation efficiency for single predict


RAMitchell:
This looks like a major bug. The unsigned integer in the omp for loop is underflowing. How long does it take to do a single predict? I would have thought this would crash.

hcho3:
@RAMitchell I'm not sure if it's actually underflowing. If `nsize` is smaller than `kUnroll`, then `nsize - rest` would simply be 0.

KerryJi:
> This looks like a major bug. The unsigned integer in the omp for loop is underflowing. How long does it take to do a single predict? I would have thought this would crash.

It takes about 40 ms. if you add this judgment code, it only takes 0.4 ms。

Issue #5030, opened 2019-11-11
Cannot use random state passed by IterativeImputer


david-cortes:
If I try to use `XGBRegressor` in [SciKit Learn's Iterative Imputer](https://scikit-learn.org/stable/modules/generated/sklearn.impute.IterativeImputer.html), it will fail due to the random state that is passed from the imputer - example:
```
estimator = make_pipeline(
    IterativeImputer(XGBRegressor()),
    Ridge()
)
```
```
XGBoostError: Invalid Parameter format for seed expect int but value='<mtrand.RandomState object at 0x7f865f1d30f0>'
```

Would be nice if xgboost's scikit-learn API classes could accept this type of random state like sklearn's own classifiers/regressors.

Full example:
```python
import numpy as np
from xgboost import XGBRegressor
from sklearn.experimental import enable_iterative_imputer
from sklearn.impute import IterativeImputer
from sklearn.pipeline import make_pipeline
from sklearn.model_selection import cross_val_score

np.random.seed(1)
X = np.random.normal(size = (100, 5))
X[np.random.randint(100, size = 10), np.random.randint(5, size = 10)] = np.nan
y = np.random.normal(size = 100)
estimator = make_pipeline(
    IterativeImputer(XGBRegressor(n_jobs=16)),
    XGBRegressor()
)
cross_val_score(
    estimator, X, y, scoring='neg_mean_squared_error',
    cv=5
)
```

trivialfis:
@david-cortes Would you like to open a PR as you already familiar with it?

david-cortes:
I’m not really familiar with how it works, but have been checking it in a bit more detail.

I’ve now realized that the class RandomState is from numpy ([link](https://github.com/numpy/numpy/blob/9ee262b5bf89e8c866a507e4d62b78532361adc2/numpy/random/mtrand.pyx)), and as of v1.17, it is a container for a random state in a Mersenne-Twister number generator, which is supposed to be used by the classifier/regressor for random number generators, with its state becoming changed as it generates random numbers, and this new state then passed to the next classifier/regressor.

Since xgboost doesn’t use this python class for random number generation, and it is able to use something other than C++’s MT19937, I think it might not be feasible to simply convert python-C++ states back and forth. Even then, NumPy might change the default RNG from MT19937 to something else in the future too.

[...]

PR #5032, opened 2019-11-12
Allow using RandomState object from NumPy

david-cortes:
Trying to fix issue: https://github.com/dmlc/xgboost/issues/5030

Sometimes, SciKit Learn will pass a parameter `random_state` to model objects which is not an integer, but rather an instance of NumPy's MT19937 random state class. Objects from SciKit-Learn such as `RandomForestRegressor` are able to use this type of random state in addition to random seeds, but XGB at the time of writing only takes integers.

This is particularly problematic when trying to use XGBoost in pipelines - e.g. `IterativeImputer` from SciKit-Learn will fail to fit with objects from XGBoost's sklearn module.

This quick fix "solves" the issue by using the NumPy RandomState object to draw a random integer which is used as seed for XGBoost, rather than trying to copy this random state back and forth with the C++ MT19937 which is used by default by XGBoost.

(I wasn't able to compile and test, so I'm hoping that this PR will pass automatic tests once it's created)

### DIFF
--- a/tests/python/test_with_sklearn.py
+++ b/tests/python/test_with_sklearn.py
@@ -444,11 +444,21 @@ def test_split_value_histograms():
 
 
 def test_sklearn_random_state():
+    from sklearn.datasets import load_iris
+
     clf = xgb.XGBClassifier(random_state=402)
     assert clf.get_xgb_params()['random_state'] == 402
 
     clf = xgb.XGBClassifier(random_state=401)
     assert clf.get_xgb_params()['random_state'] == 401
+
+    random_state = np.random.RandomState(seed = 403)
+    iris = load_iris()
+    clf = xgb.XGBClassifier(booster='gblinear', n_estimators=100, random_state=random_state)
+    clf.fit(iris.data, iris.target)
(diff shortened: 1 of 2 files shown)


tests/python/test_with_sklearn.py
```diff
@@ -444,12 +444,22 @@ def test_split_value_histograms():
...
     clf = xgb.XGBClassifier(random_state=402)
     assert clf.get_xgb_params()['random_state'] == 402
 
     clf = xgb.XGBClassifier(random_state=401)
     assert clf.get_xgb_params()['random_state'] == 401
 
+    random_state = np.random.RandomState(seed = 403)
+    iris = load_iris()
+    clf = xgb.XGBClassifier(booster='gblinear', n_estimators=100, random_state=random_state)
+    clf.fit(iris.data, iris.target)
```
trivialfis: Let me try to simplify this into a unittest.

david-cortes:
Closing as there’s an improved PR opened: https://github.com/dmlc/xgboost/pull/5049

PR #4102, opened 2019-02-05
Separate Depthwidth and Lossguide growing policy in fast histogram

CodingCat:
it's a following PR of https://github.com/dmlc/xgboost/pull/4011 which separates depthwidth and loss guide growing policy in fast histogram per our discussion in https://github.com/dmlc/xgboost/issues/4077

closes https://github.com/dmlc/xgboost/issues/4077

@trivialfis @hcho3 @RAMitchell please help to review

PR #5034, opened 2019-11-13
Extract interaction constraint from split evaluator.

trivialfis:
* Extract interaction constraints from split evaluator.

The reason for doing so is mostly for model IO, where `num_feature` and `interaction_constraints` are copied in split evaluator.  Also interaction constraint by itself is a feature selector, acting like column sampler and it's inefficient to bury it deep in the evaluator chain.  Lastly removing one another copied parameter is a win.

* Enable inc for approx tree method.

As now the implementation is spited up from evaluator class, it's also enabled for approx method.

* Removing obsoleted code in colmaker.

They are never documented nor actually used in real world.  Also there isn't a single test for those code blocks.

* Unifying the types used for row and column.

As the size of input dataset is marching to billion, incorrect use of `int` is subject to overflow, also singed integer overflow is undefined behaviour.  This PR starts the procedure for unifying used index type to unsigned integers.  There's optimization that can utilize this undefined behaviour, but after some testings I don't see the optimization is beneficial to XGBoost.

Related to #4732 .


src/tree/updater_quantile_hist.cc
```diff
@@ -170,6 +173,8 @@ void QuantileHistMaker::Builder::BuildNodeStats(
       auto parent_split_feature_id = snode_[parent_id].best.SplitIndex();
       spliteval_->AddSplit(parent_id, left_sibling_id, nid, parent_split_feature_id,
```
trivialfis: @hcho3 I'm not entirely sure why there are two splits on the pipeline.

src/tree/updater_quantile_hist.cc
```diff
@@ -298,6 +303,7 @@ void QuantileHistMaker::Builder::ExpandWithLossGuide(
       bst_uint featureid = snode_[nid].best.SplitIndex();
       spliteval_->AddSplit(nid, cleft, cright, featureid,
```
trivialfis: @hcho3 Here is another one

src/tree/updater_quantile_hist.cc
hcho3: Interesting, I have no idea either. Let's dig into git blame.

trivialfis: @hcho3 Seems to be #4102 .

trivialfis:
@hcho3 @RAMitchell  I enforced the row index to be `uint64_t` as `size_t` is different for 32 bit system and 64 bit system.  This might change memory usage on 32 bit system, does that seem to be a reasonably change?

trivialfis:
Note:
On OSX:
```
typedef unsigned int         uint32_t;
typedef unsigned long long   uint64_t;
typedef unsigned long       __darwin_size_t;
```

PR #5044, opened 2019-11-18
Demo of external data adapters

RAMitchell:
This is a proof of concept for an adapter class for external data. The idea is to create a common interface to access data external to xgboost. The interface returns batches of COO tuples, this way we can create a DMatrix class regardless of what format the external data is actually in. This demo shows how a SimpleDMatrix can be constructed from a dense matrix, CSR matrix and a CSC matrix without ever knowing the actual type of the input. The idea would be to replace all existing DMatrix construction with this interface.

The use of polymorphism creates a performance concern. Here are some numbers for creating a 100000*1000 matrix from a dense input:

```
[16:01:40] Adapter - 1 thread:   1.735766s
[16:01:41] Adapter - all threads:        0.489645s
[16:01:41] OMP:  0.519433s
[16:01:43] Default:      1.304615s
```
The single threaded adapter version is approximately 0.4s slower than the existing single threaded constructor ` XGDMatrixCreateFromMat`. The multithreaded adapter version is on par with `XGDMatrixCreateFromMat_omp`. This seems within an acceptable range, if not I think it can be optimised more.

The use of these adapters can greatly simplify our data pipeline. A DMatrix class only needs an adapter to construct itself, and does not care about the underlying format. New external data sources just need to implement an adapter, then any DMatrix class can use it. We can implement quantile DMatrix classes that can perform sketching directly over external data, completely avoiding memory for an intermediate CSR representation. This is particularly useful for our GPU algorithms.

We can unify the way threads are used to construct a DMatrix. Currently some c_api constructors silently use all threads, some specify nthreads as an argument.

We can also unify missing value handling across input sources. Currently we allow specifying missing values for dense matrices but ignore this for CSR/CSC.


hcho3:
Nice, so this should help with memory consumption as well, yes?

sriramch:
@RAMitchell thanks for clearly expressing the approach via this prototype. i have a few questions as under:

- doesn't this new approach require the *entire* data to be available somewhere (on the host memory, for instance), for it to be then fed in via these iterators
- presently, the 'etl' data brought into the device incrementally, is simply sinked into a dmatrix on host, which is finally given for training. it sounds like we need a similar sink which will then be strided over by an external adapter
- can the quantiles be incrementally built using the *partial foreign* data already on the device without any intermediary sinks. the quantile sketching can be done in batches even today. ideally, there could be user provided hooks that is called which can then release the foreign data after the sketching process is complete, so that we still have room in the device to accommodate unprocessed foreign etl data and processed internal data that may be used for training

sriramch:
> I think it would be better if we keep the iterator idea with templates as it's simpler in code. (No runtime polymorphism)

is the adapter interface going to be exposed via the 'c' interface (as a void pointer)? if so, then the template approach *may* not be possible.

RAMitchell:
> I think it would be better if we keep the iterator idea with templates as it's simpler in code. (No runtime polymorphism)

Templates are not necessarily simpler, but we may have to use them for the device variant because of difficulties using polymorphism in kernels.

> doesn't this new approach require the entire data to be available somewhere (on the host memory, for instance), for it to be then fed in via these iterators

Are you asking about loading something like a single file, or with respect to incremental building from many sources. In theory, if data is being loaded from a file, the adapter can buffer and stream the data. 

I am not sure how we can deal with incremental building yet.

> is the adapter interface going to be exposed via the 'c' interface (as a void pointer)? if so, then the template approach may not be possible.

I hadn't planned to expose it as an interface but it may be possible in a similar way that callbacks are currently implemented in the c_api.

sriramch:
> > I think it would be better if we keep the iterator idea with templates as it's simpler in code. (No runtime polymorphism)
> 
> Templates are not necessarily simpler, but we may have to use them for the device variant because of difficulties using polymorphism in kernels.

[sc] performance of device polymorphism is terrible.

> 
> > doesn't this new approach require the entire data to be available somewhere (on the host memory, for instance), for it to be then fed in via these iterators
> 
> Are you asking about loading something like a single file, or with respect to incremental building from many sources. In theory, if data is being loaded from a file, the adapter can buffer and stream the data.
> 
> I am not sure how we can deal with incremental building yet.

[sc] i was referring specifically to incremental building, as we have a solution today that syncs all data to host and uses that for training.

src/c_api/adapter.h
```diff
@@ -0,0 +1,275 @@
...
+  CSRAdapter(const size_t* row_ptr, const unsigned* feature_idx,
+             const float* values, size_t num_rows, size_t num_elements,
+             size_t num_features)
+      : ExternalDataAdapter(num_features, num_rows, num_elements),
+        row_ptr(row_ptr),
+        feature_idx(feature_idx),
+        values(values) {}
+
+ private:
+  class CSRAdapterBatch {
```
trivialfis: Why do we need an intermedia type `Batch`?

src/c_api/adapter.h
```diff
@@ -0,0 +1,275 @@
...
+   private:
+    size_t row_idx;
+    size_t size;
+    const unsigned* feature_idx;
+    const float* values;
+  };
+
+ public:
+  size_t Size() const { return num_rows; }
+  const CSRAdapterBatch operator[](size_t idx) const {
```
trivialfis: Could you write a `\brief` for what is the semantic meaning of `operator[]`, first glance it wasn't clear to me what exactly is a batch for different adapters.

src/c_api/adapter.h
RAMitchell: We need to access the data in parts on the host when using sparse formats because we cannot efficiently perform random access reads into a sparse matrix. If I tried to randomly access an element in a CSR matrix, the floating point value and column index are immediately available but I need to perform binary search to find the row index. 

So it is efficient to access elements by iterating through the rows of a CSR matrix or the columns of a CSC matrix, but not by random access. 

On the GPU we will simply do the binary search because we have a lot more computing power.

RAMitchell:
> Originally I though maybe an iterator can be a thinner layer without any form of runtime polymorphism, and can be directly put into thrust algorithms. With various adapters I'm not sure how to do that. Reading your code I believe you know better than me about iterators. But could you help me understand why you choose this route?

Current code uses templates and not run-time polymorphism. This entire exercise is just to provide a uniform view of data, considering that we have many emerging formats. 

One interesting side effect of using these adapters is that we can create adapters over DMatrix classes. This means we could construct any DMatrix class from any other DMatrix class without needing to write a constructor.

RAMitchell:
I decided to test my assumptions that random access into sparse matrices is too slow and it seems to be true
```
[ RUN      ] c_api.RandomAccessPerformance
[14:07:27] Default:      1.543194s
[14:07:31] Random access:        4.214347s
```
This is for matrix construction with random access into the input sparse matrix. Every time we access an element it requires binary search.

src/c_api/adapter.h
```diff
@@ -0,0 +1,478 @@
...
+ * for a CSC matrix. Typically in sparse matrix formats we can efficiently
+ * access subsets of elements at a time, but cannot efficiently lookups elements
+ * by random access, hence the "line" abstraction, allowing the sparse matrix to
+ * return subsets of elements efficiently. Individual elements are described by
+ * a COO tuple (row index, column index, value).
+ *
+ * This abstraction allows us to read through different sparse matrix formats
+ * using the same interface. In particular we can write a DMatrix constructor
+ * that uses the same code to construct itself from a CSR matrix, CSC matrix,
+ * dense matrix or libsvm file. To see why this is necessary, imagine we have 5
```
trivialfis: How about csr file.

src/c_api/adapter.h
```diff
@@ -0,0 +1,478 @@
...
+ * needs to be inferred during construction. Other sparse formats may specify a
+ * number of rows/columns, but we can encounter entirely sparse rows or columns,
+ * leading to disagreement between the inferred number and the meta-info
+ * provided. To resolve this, adapters have methods specifying the number of
+ * rows/columns expected, these methods may return zero where these values must
+ * be inferred from data. A constructed DMatrix should agree with the input
+ * source on numbers of rows/columns, appending empty rows if necessary.
+ *  */
+
+struct COOTuple {
```
trivialfis: Do you have an expected effect on memory usage now that we have an intermediate type?

src/c_api/adapter.h
```diff
@@ -0,0 +1,478 @@
...
+      return DTType::kInt64;
+    } else {
+      LOG(FATAL) << "Unknown data table type.";
+      return DTType::kUnknown;
+    }
+  }
+
+  class Line {
+    float DTGetValue(const void* column, DTType dt_type, size_t ridx) const {
+      float missing = std::numeric_limits<float>::quiet_NaN();
```
trivialfis: So DT has fixed missing value?

src/common/group_data.h
```diff
@@ -71,23 +71,23 @@ struct ParallelGroupBuilder {
     // set rptr to correct size
     for (std::size_t tid = 0; tid < thread_rptr_.size(); ++tid) {
       if (rptr_.size() <= thread_rptr_[tid].size()) {
-        rptr_.resize(thread_rptr_[tid].size() + 1);  // key + 1
+        rptr_.resize(thread_rptr_[tid].size() + 1, rptr_.back());  // key + 1
       }
     }
     // initialize rptr to be beginning of each segment
-    std::size_t start = 0;
+    std::size_t count = 0;
```
trivialfis: Thanks for cleaning it up.

src/common/group_data.h
```diff
@@ -71,23 +71,23 @@ struct ParallelGroupBuilder {
     // set rptr to correct size
     for (std::size_t tid = 0; tid < thread_rptr_.size(); ++tid) {
       if (rptr_.size() <= thread_rptr_[tid].size()) {
-        rptr_.resize(thread_rptr_[tid].size() + 1);  // key + 1
+        rptr_.resize(thread_rptr_[tid].size() + 1, rptr_.back());  // key + 1
```
trivialfis: Is it possible that `rptr_` has 0 size?  `back()` will segfault.

src/data/simple_dmatrix.h
```diff
@@ -26,6 +28,116 @@ class SimpleDMatrix : public DMatrix {
...
+    if (offset_vec.size() - 1 < adapter->NumRows() && adapter->NumRows() > 0) {
+      while (offset_vec.size() - 1 < adapter->NumRows()) {
+        offset_vec.emplace_back(offset_vec.back());
+      }
+      mat.info.num_row_ = adapter->NumRows();
+    } else {
+      mat.info.num_row_ = offset_vec.size() - 1;
+    }
+
+    mat.info.num_nonzero_ = offset_vec.back();
```
trivialfis: Unless you are sure about there must be at least 1 element, don't use `front` `back`.

src/c_api/adapter.h
RAMitchell: This type is only created when we access elements one at a time, the data is never stored using this type.

src/c_api/adapter.h
RAMitchell: It has defined missing values for each type. This is separate from xgboost's missing value handling which is applied on top of this in the DMatrix constructor.

src/common/group_data.h
RAMitchell: Well spotted. Fixed and added tests.

RAMitchell:
This is ready for further review.

The issue with the empty dmatrix tests on dask occurred because number of columns were not synchronised among distributed workers in the new code path. I plan to add more extensive checks for correct DMatrix initialisations in distributed mode once I have all DMatrix construction happening through a unified code path.

To do after this: change the sparse page matrix constructor to use adapters, then remove all other DMatrix::Create() factory methods, leaving just xgboost::CreateDMatrix(AdapterT* adapter).

Once this is complete we can look at modifying sketching algorithms to use adapters as input. This leads to quantile algorithms with dramatically lower memory usage.

src/data/simple_dmatrix.h
```diff
@@ -26,6 +28,118 @@ class SimpleDMatrix : public DMatrix {
...
+    uint64_t default_max = std::numeric_limits<uint64_t>::max();
+    uint64_t last_group_id = default_max;
+    bst_uint group_size = 0;
+    auto& offset_vec = mat.page_.offset.HostVector();
+    auto& data_vec = mat.page_.data.HostVector();
+    uint64_t inferred_num_columns = 0;
+
+    adapter->BeforeFirst();
+    // Iterate over batches of input data
+    while (adapter->Next()) {
```
trivialfis: I'm not sure why you want to move the data handling logic into SimpleDMatrix.  Unless you want to merge `*Source` and `*DMatrix`, there is a (vague) line drawn between them.  The source handles raw data as backend class, while DMatrix is a frontend class that provides user interface for manipulating processed data, like transforming from CSR to CSC and vice versa.  I sometime visualize a workflow to understand the code, previously I thought it should be:
```
numpy/cudf  -> c_api -> (*)adapter -> (*)source -> (*)dmatrix   # (*) means various subclasses
```
But if the function is placed in `SimpleDmatrix`, this workflow doesn't apply.  This kind of visualization is rather important to me as I can't remember the code, but understanding the workflow is easy.

src/data/simple_dmatrix.h
```diff
@@ -26,6 +28,118 @@ class SimpleDMatrix : public DMatrix {
...
+            builder.Push(element.row_idx, Entry(element.column_idx, element.value),
+              tid);
+          }
+        }
+      }
+
+      // Append meta information if available
+      if (batch.Labels() != nullptr) {
+        auto& labels = mat.info.labels_.HostVector();
+        labels.insert(labels.end(), batch.Labels(), batch.Labels() + batch.Size());
```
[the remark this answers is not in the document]
RAMitchell: This, like much of this code is copied from the original implementation. I would like to leave it as this code is clear and I dont think performance matters for this line.

src/c_api/adapter.h
```diff
@@ -0,0 +1,478 @@
+/*!
+ *  Copyright (c) 2019 by Contributors
+ * \file adapter.h
```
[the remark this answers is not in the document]
RAMitchell: Agree

src/data/simple_dmatrix.h
RAMitchell: I do want to merge them. They are so closely tied that i dont think we need both. I want to have self contained constructors for the dmatrix.

src/c_api/adapter.h
```diff
@@ -0,0 +1,478 @@
...
+    row_offset = 0;
+  }
+  bool Next() override {
+    bool next = parser->Next();
+    batch.reset(new FileAdapterBatch(&parser->Value(), row_offset));
+    row_offset += parser->Value().size;
+    return next;
+  }
+  // Indicates a number of rows/columns must be inferred
+  size_t NumRows() const { return 0; }
```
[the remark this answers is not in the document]
RAMitchell: Agree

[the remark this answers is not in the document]
RAMitchell: By far a better way to do this is to use a 'maybe' type but we don't have this in c++11.

sriramch:
@RAMitchell - while working on the last ranking algorithm, i sync'd up my branch with the master today that pulled in this pr (in addition to several others). it looks like dmatrix construction for large datasets (say one containing ~ 10m instances) is taking much much longer than the previous versions (in the order of several minutes from a few seconds). i wasn't sure which specific pr was the culprit, and i believe this seems to be the one.

from my initial analysis, the root cause seems to be the parallel group builder repeatedly resizing the thread vectors while iterating over the batches using the adapters to accommodate the incremental rows from the batch, even though it has no use for the row indices from the previous batches. as such, i feel the data and the offsets for the batches have to be computed locally and accumulated, in the process also reusing the thread vectors across multiple batches.

i just stumbled upon it this afternoon and wanted to alert you. i have started patching this and wanted to give you a heads up. i have to leave in a bit now, but hopefully should be able to provide you an update on this tomorrow.

RAMitchell:
Thanks @sriramch, I can see how the push backs might be causing performance problems.

PR #5057, opened 2019-11-21
[jvm-packages] Support specifying features via multiple columns

firestarman:
**Why this PR?**
There are always more features than one in a classification/regression task, but currently the API to specify feature columns in Classifier/Regressor only supports one single column, which requires users to assemble the multiple feature columns into a `org.apache.spark.ml.linalg.Vector` before fitting to train/transform. It looks like somewhat complicated.

This PR aims at eliminating the step of vectorization for feature columns, then users can fit the loaded dataset directly by specifying the feature columns as an `Array[String]`. It not only makes the XGBoost-Spark easier to use, but also saves a lot of time to learn other transformers for vectorization, especially for novices at ML.

**What this PR?**
This PR introduces a new API `setFeaturesCol(value: Array[String]`, along with its implementation and related support.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
@@ -180,4 +180,43 @@ object DataUtils extends Serializable {
     repartitionRDDs(deterministicPartition, numWorkers, arrayOfRDDs)
   }
 
+  private[spark] def convertDataFrameToXGBLabeledPointRDDs(
+      labelCol: Column,
+      featuresCols: Array[Column],
+      weight: Column,
+      baseMargin: Column,
+      group: Option[Column],
+      numWorkers: Int,
+      deterministicPartition: Boolean,
+      dataFrames: DataFrame*): Array[RDD[XGBLabeledPoint]] = {
+    // Order is [label, weight, margin, features, group]
+    val hasGroup = group.nonEmpty
+    val commonColumns = Seq(
+      labelCol.cast(FloatType),
+      weight.cast(FloatType),
+      baseMargin.cast(FloatType)) ++ featuresCols.map(_.cast(FloatType))
+    val selectedColumns = if (hasGroup) {
+      commonColumns :+ group.get.cast(IntegerType)
+    } else {
+      commonColumns
+    }
+    val featureSize = featuresCols.length
+    val fStartPos = 3
+    val arrayOfRDDs = dataFrames.toArray.map {
+      df => df.select(selectedColumns: _*).rdd.map (row => {
+        val (label, weight, baseMargin) = (row.getFloat(0), row.getFloat(1), row.getFloat(2))
+        val values = (fStartPos until fStartPos + featureSize).map(row.getAs[Float](_)).toArray
+        val xgbLp = hasGroup match {
+          // All are treated as Dense Data
+          case true => val group = row.getInt(row.size - 1)
+            XGBLabeledPoint(label, null, values, weight, group, baseMargin)
+          case false =>
+            XGBLabeledPoint(label, null, values, weight, baseMargin = baseMargin)
+        }
+        attachPartitionKey(row, deterministicPartition, numWorkers, xgbLp)
+      })
+    }
+    repartitionRDDs(deterministicPartition, numWorkers, arrayOfRDDs)
+  }
+
 }
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
@@ -182,7 +182,7 @@ object DataUtils extends Serializable {
 
   private[spark] def convertDataFrameToXGBLabeledPointRDDs(
       labelCol: Column,
-      featuresCols: Array[Column],
+      featuresCols: Seq[Column],
       weight: Column,
       baseMargin: Column,
       group: Option[Column],
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
@@ -17,7 +17,6 @@
 package ml.dmlc.xgboost4j.scala.spark
 
 import ml.dmlc.xgboost4j.{LabeledPoint => XGBLabeledPoint}
-
 import org.apache.spark.HashPartitioner
 import org.apache.spark.ml.feature.{LabeledPoint => MLLabeledPoint}
 import org.apache.spark.ml.linalg.{DenseVector, SparseVector, Vector, Vectors}
@@ -26,6 +25,8 @@ import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.{Column, DataFrame, Row}
 import org.apache.spark.sql.functions.col
 import org.apache.spark.sql.types.{FloatType, IntegerType}
+
+import scala.collection.mutable
 
 object DataUtils extends Serializable {
   private[spark] implicit class XGBLabeledPointFeatures(
@@ -180,7 +181,38 @@ object DataUtils extends Serializable {
     repartitionRDDs(deterministicPartition, numWorkers, arrayOfRDDs)
   }
 
+  private[spark] def compressValues(
+      rawValues: Seq[Any],
+      missing: Float): (Array[Int], Array[Float]) = {
+    val indices = mutable.ArrayBuilder.make[Int]
+    val values = mutable.ArrayBuilder.make[Float]
+    var featureIdx = 0
+    def keepCondition = (v: Float) => if (missing.isNaN) !v.isNaN else v != missing
+    // Build sparse array based on missing value directly from the raw rows
+    rawValues.foreach {rv =>
+      if (rv != null) {
+        val typedValue = rv match {
+          case _: Float => rv.asInstanceOf[Float]
+          case _: Double => rv.asInstanceOf[Double].toFloat
+          case _: Byte => rv.asInstanceOf[Byte].toFloat
+          case _: Short => rv.asInstanceOf[Short].toFloat
+          case _: Int => rv.asInstanceOf[Int].toFloat
+          case _: Long => rv.asInstanceOf[Long].toFloat
+          case _ => throw new RuntimeException(s"Unsupported type of feature value @index" +
+            s" $featureIdx, only numeric is allowed.")
+        }
+        if (keepCondition(typedValue)) {
+          indices += featureIdx
+          values += typedValue
+        }
+      } // else { Always remove null values }
+      featureIdx += 1
+    }
+    (indices.result, values.result)
+  }
+
   private[spark] def convertDataFrameToXGBLabeledPointRDDs(
+      missing: Float,
       labelCol: Column,
       featuresCols: Seq[Column],
       weight: Column,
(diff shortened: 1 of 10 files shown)


jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
```diff
@@ -180,4 +180,43 @@ object DataUtils extends Serializable {
...
+  private[spark] def convertDataFrameToXGBLabeledPointRDDs(
+      labelCol: Column,
+      featuresCols: Seq[Column],
+      weight: Column,
+      baseMargin: Column,
+      group: Option[Column],
+      numWorkers: Int,
+      deterministicPartition: Boolean,
+      dataFrames: DataFrame*): Array[RDD[XGBLabeledPoint]] = {
+    // Order is [label, weight, margin, features, group]
```
CodingCat: how we prevent users messing up by passing an unexpected schema?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressor.scala
```diff
@@ -274,9 +318,18 @@ class XGBoostRegressionModel private[ml] (
...
-          val features = batchRow.iterator.map(row => row.getAs[Vector]($(featuresCol)))
-
           import DataUtils._
+          val features = batchRow.iterator.map(row =>
+            if ($(featuresCols).nonEmpty) {
+              // Prefer multiple columns for features
+              val values = row.schema.filter(f => $(featuresCols).contains(f.name))
+                .map(f => row.get(row.schema.fieldIndex(f.name)).toString.toFloat).toArray
+              // All are treated as dense data
+              XGBLabeledPoint(0.0f, null, values)
```
CodingCat: IIUC, if users do not want to use vectorAssembler, they treat every row as dense vector?

CodingCat: if that's the case, can we filter missing value here to enable the usage of sparse vector?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
firestarman: Actually the order in this comment is defined by us via `DataFrame.select` below @line 206. Not meaning the order provided in schema by users.

We also has some check for the schema [here](https://github.com/dmlc/xgboost/blob/47444e1fb9fa53950e79276726f5d4069e750c9a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostEstimatorCommon.scala#L35)

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressor.scala
firestarman: Good findings and good suggestion, However XGBoost already does this to filter missing values from both dense and sparse data, please see [here](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala#L274). So we don't need to do it again.

CodingCat: no, I was saying that

1. because of the usage of vector assembler, xgboost-spark can only accept 0 as the missing value when sparse vector is presented

2. you have only two ways to workaround the limitation in 1 currently, (1) replace all 0s with some impossible values in your dataset; (2) before feed dataframe to xgboost, translate all sparse vector to dense manually.  Both approaches will bring higher memory footprint due to the usage of dense vectors

3. since this PR works on skipping vector assembler, you get some facilities to directly build spark vectors by filtering values (specified as `missing`) other than 0 before feeding them to XGBoost training part

firestarman: Actually this PR has nothing to do the sparse vector i think, but introduces a new training pipeline independently.

1. Current pipeline: **Raw DataFrame => VectorAssembler => XGBoost**.
2. A New pipeline of this PR: **Raw DataFrame => XGBoost**.

I believe it is `VectorAssembler` invovles sparse vector and dense vector. But our "multiple columns for feature" uses the raw feature columns directly, so all of their types are required to be numeric (here is the[ check](https://github.com/dmlc/xgboost/blob/47444e1fb9fa53950e79276726f5d4069e750c9a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostEstimatorCommon.scala#L39)), not sparse or dense vector. When users choose to use the new pipeline, XGBoost is fitted with the same raw `DataFrame ` as `VectorAssembler `, and build `XGBLabeledPoint`s directly from the raw feature columns and label column. What's mroe, the raw feature values are always "dense". 

As far as I know, there is NO limitation on missing value for dense feature values, then the removing logic [here ](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala#L274)will works for this new pipeline.

CodingCat:
Could you point me to my words indicating that “this PR has something to do
with sparse vector”?

firestarman:
Sorry for confusion, updated my last comment after i got what you mean. I keep " this PR has nothing to do the sparse vector" as an additional explanations of the PR.

firestarman:
Hi Reviewers, 
May i know what's the next action by me ?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressor.scala
CodingCat: I was saying ` Both approaches will bring higher memory footprint due to the usage of dense vectors`

I am repeating again "do NOT force everything to be dense, otherwise, you will have very HIGH memory footprint"

is it still not clear?

CodingCat:
again, if you modify code in jvm package, add [jvm-package] in the title

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressor.scala
firestarman: Thanks but two questions:
1) Is it correct to always remove 0 to save memory for sparse vector? Even this is forced by VectorAssembler.  I mean when 0 is a valid value for feature, removing all the 0s will affect the gradient computation and may build a different model after trainning. ( @trivialfis Please correct me if wrong. ) Since the native will not restore the missing value from the parameter "missing".
2) If it is yes to question1, is it enough to remove 0s to build a sparse vector here? Since missing value will be removed next by [here](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressor.scala#L290).

firestarman:
> again, if you modify code in jvm package, add [jvm-package] in the title

Updated.

CodingCat:
That’s what I was saying

This PR bypassed vector assembler, you can remove what user specified as
missing, instead of must be 0 in the case of vector assembler, to build
sparse vector.....

firestarman:
Thanks.
Your proposal of removing user specified missing value seems to be already supported per my understanding of the function [`processMissingValues`](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala#L289).

The feature values are treated as dense just after getting from a row, but will become sparse after calling `processMissingValues` and before building the DMatrix.

For the case of my PR, all the `XGBLabeledPoint`s are dense, meaning `labeledPoint.indices == null`, so no exception will be thrown even missing value is not 0, per the function below.
```
private def verifyMissingSetting(xgbLabelPoints: Iterator[XGBLabeledPoint], missing: Float):
      Iterator[XGBLabeledPoint] = {
    if (missing != 0.0f) {
      xgbLabelPoints.map(labeledPoint => {
        if (labeledPoint.indices != null) {
            throw new RuntimeException(s"you can only specify missing value as 0.0 (the currently" +
              s" set value $missing) when you have SparseVector or Empty vector as your feature" +
              " format")
        }
        labeledPoint
      })
    } else {
      xgbLabelPoints
    }
  }
```
Next all the dense `XGBLabeledPoint`s become sparse by removing the user specified missing values in the function `removeMissingValues`.
```
  private def removeMissingValues(
      xgbLabelPoints: Iterator[XGBLabeledPoint],
      missing: Float,
      keepCondition: Float => Boolean): Iterator[XGBLabeledPoint] = {
    xgbLabelPoints.map { labeledPoint =>
      val indicesBuilder = new mutable.ArrayBuilder.ofInt()
      val valuesBuilder = new mutable.ArrayBuilder.ofFloat()
      for ((value, i) <- labeledPoint.values.zipWithIndex if keepCondition(value)) {
        indicesBuilder += (if (labeledPoint.indices == null) i else labeledPoint.indices(i))
        valuesBuilder += value
      }
      labeledPoint.copy(indices = indicesBuilder.result(), values = valuesBuilder.result())
    }
  }
```
and the `keepCondition` is provided in `processMissingValues`.
```
  private[spark] def processMissingValues(
      xgbLabelPoints: Iterator[XGBLabeledPoint],
      missing: Float): Iterator[XGBLabeledPoint] = {
    if (!missing.isNaN) {
      removeMissingValues(verifyMissingSetting(xgbLabelPoints, missing),
        missing, (v: Float) => v != missing)
    } else {
      removeMissingValues(verifyMissingSetting(xgbLabelPoints, missing),
        missing, (v: Float) => !v.isNaN)
    }
  }
```

CodingCat:
Please read my comments carefully before helping me to understand the code
written by myself, thanks!

I have said for MANY times: It is about memory footprint!

Again, PLEASE read my comments carefully

CodingCat:
I am rethinking about this PR

1. this PR is proposing to accept multiple instead of only one input col to fit XGBoost model. 

2. the motivation of PR is stated as `currently the API to specify feature columns in Classifier/Regressor only supports one single column, which requires users to assemble the multiple feature columns into a org.apache.spark.ml.linalg.Vector before fitting to train/transform. It looks like somewhat complicated.`

I kind of agree with the above items but I would put  more thoughts about "who should resolve the problem"

XGBoostClassifier accepts a single col because it extends from the base class like `ProbabilisticClassifier` which in turn extends from `HasFeatureCol`. i.e. Classifier/Regressor takes a single col as input is a design from Spark side. To meet this requirement, Spark provides a tool called VectorAssembler to help users achieve this. As the author said, it actually enforces an additional stage in pipeline.

Why we have to resolve this `additional stage` in a downstream machine learning library instead of Spark itself? and in a way that most of Spark ML users are not used to (most of users, if not all, have been using vector assembler) ?

If we merge it now, what if Spark resolved this `additional stage`, we have to bring in a breaking change in XGBoost to leverage Spark's fix?

regarding missing value, I actually prefer the approach https://github.com/dmlc/xgboost/pull/4805/files

CodingCat:
I will be in traveling in the next few days, expect some delay of the reply

anfeng:
@CodingCat Almost all Spark XGBoost use cases have multiple feature columns. It's hard to make such statement at Spark ML level. Let's introduce this convenience method at Spark XGBoost 1st. Once we see significant adoption of this method in XGBoost, we could reengage Spark ML community to introduces this convenience method, and adjust XGBoost accordingly. The implementation change from XGBoost to Spark ML will not impact XGBoost apps.

CodingCat:
What do you mean by multiple feature col?

Of course users have multiple features, but they eventually use a single
col which is typed as vector

This is a data preprocessing stage which should go to data processing
framework, i.e. Spark

We should make XGBoost as clear as a ml library and XGBoost-Spark as an
integration in XGBoost side which only accept Spark’s definition of data
interface instead of introducing more and more data processing
functionalities

CodingCat:
Regarding ‘ Once we see significant adoption of this method in XGBoost, we
could reengage Spark ML community to introduces this convenience method,
and adjust XGBoost accordingly.’

Since we haven’t known whether it has a rich user base or say whether it is
useful, we should be more conservative to merge since merge means long-term
maintenance . Some functionality with a very small user base after a long
time is even more troublesome, it puts maintainers to dilemmas to remove or
keep outdated code

firestarman:
Updated the PR to enable the sparse array of values when building XGBLabeledPoints from raw rows.

Verifided it locally in standalone mode by training with 4G data and transforming with 1G data, their memory usage are very close. Memory checking is done by the built-in app "System Monitor" of Ubuntu. 
Spark single column: https://drive.google.com/open?id=1XqBdt8p3OeBcLEYd10r91WxO66RQ9CNw
Our multi columns: https://drive.google.com/open?id=1L1aYSuoddvIso65MZE8HyMRh9Po9Me_2

For bypassing the missing value limitation, will update accordingly once PR https://github.com/dmlc/xgboost/pull/4805 being merged, to avoid conflication.

Could you please review it again? Thanks in advance.

The main intent of this PR is to provide another (more eaiser) way to write a XGBoost Spark application by trying to avoid introducing any unnecessary data processing stage. I believe users will like it.

firestarman:
Training data is 4 copies of the data file below, (7275113 rows,  28 cols, sparsity: 0.534108115276992)
https://drive.google.com/open?id=12aF6_vTZbA1BiS0spdzZp8nU5Cw1xgl2

Transforming data is 5 copies of the data file below, (1819567 rows,  28 col, sparsity: 0.5244729142388586 )
https://drive.google.com/open?id=1aCD7l0ORW3qRr_F02iwCjrQqmLUb-JTj

PR #5064, opened 2019-11-24
Added new train_folds parameter for xgb.cv()

iblumin:
New parameter and code allows user to select which indices will be used for training in each cross-validation fold if desired. This is to be used in conjunction with the existing folds parameter. 

Currently, xgb.cv() assumes that if folds is not NULL, the given indices are to be used for the validation sets for each cross-validation fold, and all remaining indices should be used for training in each respective fold. Adding a train_folds parameter to specify which indices are to be used for the training set for each fold allows the implementation of a walk-forward or rolling window approach to cross-validation. 

I came across this code on stackoverflow, where it was originally contributed by @RolandASc.
https://stackoverflow.com/questions/32433458/how-to-specify-train-and-test-indices-for-xgb-cv-in-r-package-xgboost/51412073

### DIFF
--- a/src/gbm/gbtree_model.cc
+++ b/src/gbm/gbtree_model.cc
@@ -40,7 +40,7 @@ void GBTreeModel::Load(dmlc::Stream* fi) {
 void GBTreeModel::SaveModel(Json* p_out) const {
   auto& out = *p_out;
   CHECK_EQ(param.num_trees, static_cast<int>(trees.size()));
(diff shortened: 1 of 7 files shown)


trivialfis:
@iblumin Sorry you used your master branch and somehow I evened it.  I created a new branch for your PR in https://github.com/dmlc/xgboost/pull/5114 .

trivialfis:
And added your commit there.

PR #5078, opened 2019-11-29
[CI] Locate vcomp140.dll from System32 directory

hcho3:
After I re-built Jenkins testing pipeline from scratch, I found out that I cannot always expect the path `C:\\Program Files (x86)\\Microsoft Visual Studio\\Shared\\14.0\\VC\\redist\\x64\\Microsoft.VC140.OpenMP\\vcomp140.dll` to exist. Since vcomp140.dll is a system lib, it's better to use `C:\\Windows\\System32\\vcomp140.dll`

### DIFF
--- a/tests/ci_build/insert_vcomp140.py
+++ b/tests/ci_build/insert_vcomp140.py
@@ -7,7 +7,7 @@ if len(sys.argv) != 2:
     print('Usage: {} [wheel]'.format(sys.argv[0]))
     sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


trivialfis:
Windows is really out of my reach, maybe @RAMitchell can help here . :-)

hcho3:
I found the reason why `testxgboost.exe` was crashing. I had to install a system lib called [ Microsoft Visual C++ Redistributable for Visual Studio](https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads).

Issue #5061, opened 2019-11-23
[CI] Jenkins is down


hcho3:
The Jenkins master node (xgboost-ci.net) has been down (*) for a few days, according to @trivialfis. **Impact: Substantial amount of test suite is rendered unavailable.** I'm re-building the master from scratch. Sorry for inconvenience caused. @dmlc/xgboost-committer 

PS. This seems to be a good opportunity to document all steps it takes to set up the Jenkins master node from scratch. See hcho3/xgboost-devops#6. Also #4958 

(*) I cannot either ping or SSH into the master instance.

trivialfis:
Travis seems also down as there is a permission error.  Will try to resolve it later.

trivialfis:
Travis is fixed in #5062

hcho3:
@trivialfis Thanks for fixing Travis! I'll get Jenkins back up very soon.

hcho3:
Update: Linux pipeline is now back up. I still need to work on the Windows pipeline.

trivialfis:
Okay.  I messed up a little.

hcho3:
@trivialfis I updated https://github.com/hcho3/xgboost-devops/issues/6 to include some details about the Windows pipeline. I'm also closing this issue, as Jenkins is now fully back up.

hcho3:
Sorry, looks like I called it too early. Windows pipeline is still broken.

hcho3:
Fixed in #5078

PR #5080, opened 2019-11-29
Update python install script to fix MacOS build error (#4933).

yage99:
The build documentation and python installation script is updated to resolve the installation issue in MacOS versions higher than High Sierra. see #4933.

### DIFF
--- a/doc/build.rst
+++ b/doc/build.rst
@@ -98,11 +98,11 @@ Building on OSX
 Install with pip: simple method
 --------------------------------
 
-First, obtain ``gcc-8`` with Homebrew (https://brew.sh/) to enable multi-threading (i.e. using multiple CPU threads for training). The default Apple Clang compiler does not support OpenMP, so using the default compiler would have disabled multi-threading.
+First, obtain ``gcc-9`` and ``OpenMP`` with Homebrew (https://brew.sh/) to enable multi-threading (i.e. using multiple CPU threads for training). The default Apple Clang compiler does not support OpenMP, so using the default compiler would have disabled multi-threading.
 
 .. code-block:: bash
 
-  brew install gcc@8
+  brew install gcc@9 libomp
 
 Then install XGBoost with ``pip``:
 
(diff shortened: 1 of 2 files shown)


trivialfis:
Do you think it's possible to use CMake?  Currently we are desperately trying to restrict the usage of makefile to R package cran distribution.

trivialfis:
If not we will have to consider putting some efforts on making those makefiles on par with CMake.

yage99:
The goal of this patch is to do a quick fix for the current issue.
CMake maybe an option, which will enable OpenMP with clang. So users don't need to install gcc. But this migration may break the current stability and needs some tests.
If you want I can help do this work recently.

yage99:
I'm not sure what use CMake mean. Python use dmlc-core/CMakeLists.txt to compile library, currently clang is not fully compatible with OpenMP. So install gcc@9 and libomp may be the easiest way to compile XGBoost on MacOS.

yage99:
Hi, @trivialfis I have figured out the CMake stuff. Sorry I didn't notice the python script was using Makefile to compile the library. The install script is updated to use CMakeList.txt instead to compile XGBoost library. And I have tested this patch in MacOS and Linux (Ubuntu 18.04).

trivialfis:
@yage99 Thanks.  The doc looks good to me.  Will try out the build myself once #5061 is resolved.

trivialfis:
@hcho3 Could you please help taking a look?

> FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:\\Program Files (x86)\\Microsoft Visual Studio\\Shared\\14.0\\VC\\redist\\x64\\Microsoft.VC140.OpenMP\\vcomp140.dll'

https://xgboost-ci.net/blue/organizations/jenkins/xgboost-win64/detail/PR-5080/3/pipeline/

hcho3:
@trivialfis The issue was fixed in #5078. Please rebase against latest master.

yage99:
I have rebased my changes against the new master. Please check if this works well. @trivialfis

trivialfis:
@yage99  Thanks!

PR #5082, opened 2019-12-01
[jvm-packages] do not use multiple jobs to make checkpoints

CodingCat:


CodingCat:
@trams mind taking a look?

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/DeprecatedCheckpointManagerSuite.scala
```diff
@@ -18,54 +18,71 @@ package ml.dmlc.xgboost4j.scala.spark
...
 
-  private lazy val (model4, model8) = {
-    val training = buildDataFrame(Classification.train)
-    val paramMap = Map("eta" -> "1", "max_depth" -> "2", "silent" -> "1",
-      "objective" -> "binary:logistic", "num_workers" -> sc.defaultParallelism)
-    (new XGBoostClassifier(paramMap ++ Seq("num_round" -> 2)).fit(training),
-    new XGBoostClassifier(paramMap ++ Seq("num_round" -> 4)).fit(training))
+  private def produceParamMap(checkpointPath: String, checkpointInterval: Int):
+  Map[String, Any] = {
+    Map("eta" -> "1", "max_depth" -> "2", "silent" -> "1",
```
trivialfis: Silent is deprecated.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/ExternalCheckpointManager.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  public Booster loadCheckpointAsBooster() throws IOException, XGBoostError {
+    List<Integer> versions = getExistingVersions();
+    if (versions.size() > 0) {
+      int latestVersion = versions.stream().max(Comparator.comparing(Integer::valueOf)).get();
+      String checkpointPath = getPath(latestVersion);
+      InputStream in = fs.open(new Path(checkpointPath));
+      logger.info("loaded checkpoint from " + checkpointPath);
+      Booster booster = XGBoost.loadModel(in);
```
trivialfis: I have a RFC for distinguishing `model` and `checkpoint`: https://github.com/dmlc/xgboost/issues/4855 .  Which is implemented in #4732  with simple documentation.  The basic idea is when you perform `SaveModel` only the trees and objective are saved, while `SaveCheckPoints` also saves all hyper-parameters like `eta, max_depth, tree_method` etc that's guaranteed to carry out the previous training.  By carrying out previous training, I mean:
``` 
 for i in range(4):
    bst.update_one_iter()

  bst.save()
  bst.load()

  for i in range(4):
    bst.update_one_iter()
```
Should equals to following when prediction cache is not involved:
```
  for i in range(8):
    bst.updater_one_iter()
```
It seems `XGBoosterLoadRabitCheckpoint` is more appropriate here.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/ExternalCheckpointManager.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+
+  public void updateCheckpoint(Booster boosterToCheckpoint) throws IOException, XGBoostError {
+    List<String> prevModelPaths = getExistingVersions().stream()
+            .map(this::getPath).collect(Collectors.toList());
+    String eventualPath = getPath(boosterToCheckpoint.getVersion());
+    String tempPath = eventualPath + "-" + UUID.randomUUID();
+    OutputStream out = fs.create(new Path(tempPath), true);
+    boosterToCheckpoint.saveModel(out);
```
trivialfis: Same as above.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/ExternalCheckpointManager.java
CodingCat: hmmm....in JVM part, we actually not that care about it, as the training parameters are corrected by what users passed in training API


e.g. 

1. we load a booster with default parameter at https://github.com/dmlc/xgboost/pull/5082/files#diff-095250777d030169ae83818b37ab44b1R530

2. we continue training with https://github.com/dmlc/xgboost/pull/5082/files#diff-095250777d030169ae83818b37ab44b1R543-R544

3. in buildDistributedBoosters https://github.com/dmlc/xgboost/pull/5082/files#diff-095250777d030169ae83818b37ab44b1R442-R443 we will call Java API like https://github.com/dmlc/xgboost/pull/5082/files#diff-095250777d030169ae83818b37ab44b1R354-R357

4. before we start  training, we'll set params https://github.com/dmlc/xgboost/pull/5082/files#diff-314932ca1cabd6bd148e8f8bd85b8947R190

trivialfis: @CodingCat Can we consider replacing all these with the JSON implementation after I manage to merge that PR? Maybe next release after I tested it enough on dask side?  You can load/save only hyper-parameters in that PR.

trivialfis: @CodingCat I provided a sample output of `XGBoosterSaveJsonParameters` in #4732 's document.  Pasted below.  I believe it's more robust than language binding implementation as it actually walks through all parameters.  Also with it we can build some consistency between language bindings.


``` json
{
      "Learner": {
        "generic_parameter": {
          "enable_experimental_json_serialization": "0",
          "gpu_id": "0",
          "gpu_page_size": "0",
          "n_jobs": "0",
          "random_state": "0",
          "seed": "0",
          "seed_per_iteration": "0"
        },
        "gradient_booster": {
          "gbtree_train_param": {
            "num_parallel_tree": "1",
            "predictor": "gpu_predictor",
            "process_type": "default",
            "tree_method": "gpu_hist",
            "updater": "grow_gpu_hist",
            "updater_seq": "grow_gpu_hist"
          },
          "name": "gbtree",
          "updater": {
            "grow_gpu_hist": {
              "gpu_hist_train_param": {
                "debug_synchronize": "0",
                "gpu_batch_nrows": "0",
                "single_precision_histogram": "0"
              },
              "train_param": {
                "alpha": "0",
                "cache_opt": "1",
                "colsample_bylevel": "1",
                "colsample_bynode": "1",
                "colsample_bytree": "1",
                "default_direction": "learn",
                "enable_feature_grouping": "0",
                "eta": "0.300000012",
                "gamma": "0",
                "grow_policy": "depthwise",
                "interaction_constraints": "",
                "lambda": "1",
                "learning_rate": "0.300000012",
                "max_bin": "256",
                "max_conflict_rate": "0",
                "max_delta_step": "0",
                "max_depth": "6",
                "max_leaves": "0",
                "max_search_group": "100",
                "refresh_leaf": "1",
                "sketch_eps": "0.0299999993",
                "sketch_ratio": "2",
                "subsample": "1"
              }
            }
          }
        },
        "learner_train_param": {
          "booster": "gbtree",
          "disable_default_eval_metric": "0",
          "dsplit": "auto",
          "objective": "reg:squarederror"
        },
        "metrics": [],
        "objective": {
          "name": "reg:squarederror",
          "reg_loss_param": {
            "scale_pos_weight": "1"
          }
        }
      },
      "version": [1, 0, 0]
    }
```

jvm-packages/xgboost4j/pom.xml
```diff
@@ -13,6 +13,16 @@
     <packaging>jar</packaging>
 
     <dependencies>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-hdfs</artifactId>
+            <version>2.7.3</version>
```
trams: This may be problematic in our (Criteo) environment and in general too.
The problem is that currently `xgboost-spark` depends on Spark 2.4.3 (https://github.com/dmlc/xgboost/blob/551db2269f037337dc4ef464e8e278c5f1eca607/jvm-packages/pom.xml#L37)
and it depends on Hadoop 2.6.5
So I suggest two things
1. change it to 2.6.5 unless you really need 2.7.3
2. Move this to a parent pom as a property and use it here

Also note you are including this as a compile dependency which may be not ideal at least in some environments but here I am no expert. I am familiar with our Hadoop|Yarn cluster only and our way of managing Spark jobs which is a bit weird :)

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/ExternalCheckpointManager.java
```diff
@@ -0,0 +1,116 @@
...
+      return null;
+    }
+  }
+
+  public void updateCheckpoint(Booster boosterToCheckpoint) throws IOException, XGBoostError {
+    List<String> prevModelPaths = getExistingVersions().stream()
+            .map(this::getPath).collect(Collectors.toList());
+    String eventualPath = getPath(boosterToCheckpoint.getVersion());
+    String tempPath = eventualPath + "-" + UUID.randomUUID();
+    OutputStream out = fs.create(new Path(tempPath), true);
```
trams: We can try to use try-with-resources here to make it a bit clearer and cleaner
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

It is the same concept as RAII from C++ but in Java

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/ExternalCheckpointManager.java
```diff
@@ -0,0 +1,116 @@
...
+      }
+    });
+  }
+
+  public void cleanUpHigherVersions(int currentRound) throws IOException {
+    getExistingVersions().stream().filter(v -> v / 2 >= currentRound).forEach(v -> {
+      try {
+        fs.delete(new Path(getPath(v)), true);
+      } catch (IOException e) {
+        e.printStackTrace();
```
trams: I suggest to use logger.error here. You can copy the message from line 79

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -108,42 +112,47 @@ public static Booster train(
...
+          Booster booster,
+          int iter,
+          Set<Integer> checkpointIterations,
+          ExternalCheckpointManager ecm) throws XGBoostError {
+    try {
+      if (checkpointIterations.contains(iter)) {
+        ecm.updateCheckpoint(booster);
+      }
+    } catch (Exception e) {
+      logger.error("failed to save checkpoint in XGBoost4J", e);
```
trams: I suggest to add to the message `iter`. It will greatly the quality of an expection

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -108,42 +112,47 @@ public static Booster train(
...
+      Map<String, Object> params,
+      int numRounds,
+      Map<String, DMatrix> watches,
+      float[][] metrics,
+      IObjective obj,
+      IEvaluation eval,
+      int earlyStoppingRounds,
+      Booster booster,
+      int checkpointInterval,
+      String checkpointPath,
```
trams: I am a bit confused. Is there a reason why we don't use Hadoop Path object here. I would also replace Filesystem on line 143 to Configuration object instead and create Filesystem on demand like it is done in scala wrapper

P.S. Is there a reason why do we have java and scala wrappers at the same time?

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -239,6 +253,44 @@ public static Booster train(
...
+          IEvaluation eval,
+          int earlyStoppingRounds,
+          Booster booster) throws XGBoostError {
+    try {
+      return trainAndSaveCheckpoint(dtrain, params, round, watches, metrics, obj, eval,
+              earlyStoppingRounds, booster,
+              -1, null, null);
+    } catch (IOException e) {
+      logger.error("training failed in xgboost4j", e);
+      throw new XGBoostError("training failed in xgboost4j " + e);
```
trams: I suggest to include `e` as a cause when creating an exception.

Instead of `new RuntimeException("message " + e)` one can do `new RuntimeException("message ", e)`
this way the stack trace of the cause will be passed down and it will resurface to the developer faster.

To archive the same using XGBoostError I suggest to add a parameter `cause` and pass it to the parent constructor (see https://docs.oracle.com/javase/7/docs/api/java/lang/Exception.html#Exception(java.lang.String,%20java.lang.Throwable) ) for more details

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/ExternalCheckpointManager.java
CodingCat: sure

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -342,14 +341,25 @@ object XGBoost extends Serializable {
     rabitEnv.put("DMLC_TASK_ID", taskId)
     rabitEnv.put("DMLC_NUM_ATTEMPT", attempt)
     rabitEnv.put("DMLC_WORKER_STOP_PROCESS_ON_ERROR", "false")
-
+    val numRounds = xgbExecutionParam.numRounds
+    val makeCheckpoint = xgbExecutionParam.checkpointParam.checkpointPath != null &&
```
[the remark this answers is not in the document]
CodingCat: sure, done

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
CodingCat: done

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
CodingCat: I think Scala wrapper also accepts string?

the major reason to have two wrappers is to return Scala Booster

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
CodingCat: done

jvm-packages/xgboost4j/pom.xml
CodingCat: I changed to provided and keep 2.7.3 , the reason is that 2.7.3 is kind of standard support in spark, if you look at spark, it only provides pre-build for 2.7

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -342,14 +341,25 @@ object XGBoost extends Serializable {
...
-      val booster = SXGBoost.train(watches.toMap("train"), xgbExecutionParam.toMap, round,
-        watches.toMap, metrics, obj, eval,
-        earlyStoppingRound = numEarlyStoppingRounds, prevBooster)
+      val metrics = Array.tabulate(watches.size)(_ => Array.ofDim[Float](numRounds))
+      val externalCheckpointParams = xgbExecutionParam.checkpointParam
+      val booster = if (makeCheckpoint) {
+        SXGBoost.trainAndSaveCheckpoint(
+          watches.toMap("train"), xgbExecutionParam.toMap, numRounds,
+          watches.toMap, metrics, obj, eval,
+          earlyStoppingRound = numEarlyStoppingRounds, prevBooster, Some(externalCheckpointParams))
```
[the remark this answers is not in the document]
CodingCat: changed

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -342,14 +341,25 @@ object XGBoost extends Serializable {
     rabitEnv.put("DMLC_TASK_ID", taskId)
     rabitEnv.put("DMLC_NUM_ATTEMPT", attempt)
     rabitEnv.put("DMLC_WORKER_STOP_PROCESS_ON_ERROR", "false")
-
+    val numRounds = xgbExecutionParam.numRounds
+    val makeCheckpoint = xgbExecutionParam.checkpointParam.checkpointPath != null &&
+      xgbExecutionParam.checkpointParam.checkpointPath.nonEmpty &&
+      taskId.toInt == 0
```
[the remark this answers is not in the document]
CodingCat: I think we can do it later if we find the perf issue

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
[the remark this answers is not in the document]
CodingCat: done

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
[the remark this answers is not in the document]
trams: Agree

jvm-packages/xgboost4j/pom.xml
trams: I forgot that you target Spark 3.0 in this release which supports only Hadoop 2.7
Making dependency `provided` should solve Criteo's problem

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
trams: I see. It makes sense

CodingCat:
ok, finally got a change to do integration test, everything looks fine 

single job

image

cp created 

image

PR #5083, opened 2019-12-02
[WIP] [CI] Add CUDA 10.2

hcho3:
* Add CUDA 10.2
* Remove CUDA 10.0: mostly to conserve CI resource. Also: CUDA 10.0 is incompatible with G4 instance type on Windows, and we'd like to use G4 if at all possible (much better bang for the buck than P2)

@rongou @sriramch

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -64,8 +64,8 @@ pipeline {
             'build-cpu': { BuildCPU() },
             'build-cpu-rabit-mock': { BuildCPUMock() },
             'build-gpu-cuda9.0': { BuildCUDA(cuda_version: '9.0') },
-            'build-gpu-cuda10.0': { BuildCUDA(cuda_version: '10.0') },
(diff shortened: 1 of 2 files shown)


sriramch:
@hcho3 our releases go out primarily on 10.0. i see the reason behind removing 10.0 (due to the need to use g4). is it possible to still retain this?
should 9.x be removed instead, if resource savings is another goal?

rongou:
@hcho3 we can probably remove 10.1 if we still want to keep cuda 9.

trivialfis:
@RAMitchell Found some issues with CUDA 10.2?

hcho3:
Maybe we should wait a bit, given that CUDA 10.2 is still pretty new?

hcho3:
Closing for now. Will re-visit later.

RAMitchell:
I had segfaults on 10.2, not sure if its local to my machine. Will do some more investigation.

PR #5072, opened 2019-11-27
[WIP] Hack for GPU ID config

trivialfis:
Currently depends on #5066.  This is a very ugly hack for saving making sure pickled model understand what GPU should be used after load.  Before I can get the new model IO merged, this unblocks some issues.

Closes #5003 .

PR #5086, opened 2019-12-04
Hack for GPU ID config

trivialfis:
Continues https://github.com/dmlc/xgboost/pull/5072 .  The JSON PR works out everything, but being too huge.  So continue the band-aid for rapids' release.  @RAMitchell @JohnZed 

* Define prediction cache in GBTree.
* Initialise prediction cache on first iteration.
* Save gpu_id into binary model when CUDA is enabled.

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -295,7 +295,6 @@ class LearnerImpl : public Learner {
 #ifdef XGBOOST_USE_CUDA
           if (saved_param == "predictor" || saved_param == "gpu_id") {
             cfg_[saved_param] = kv.second;
-            LOG(WARNING) << saved_param << " value: " << kv.second;
             LOG(INFO)
               << "Parameter '" << saved_param << "' has been recovered from "
               << "the saved model. It will be set to "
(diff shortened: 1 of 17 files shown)


src/learner.cc
```diff
@@ -287,6 +316,8 @@ class LearnerImpl : public Learner {
             cfg_["predictor"] = "cpu_predictor";
             kv.second = "cpu_predictor";
             LOG(INFO) << "Switch gpu_predictor to cpu_predictor.";
+          } else {
```
JohnZed: Probably I'm just missing it, but why not `else if (is_gpu_predictor)` so it doesn't trigger in the cpu_predictor case?

trivialfis:
I might implement a state machine for this configuration in next PR.

trivialfis:
@RAMitchell I will revert the changes in Learner if I can't get JSON in soon enough.

trivialfis:
Seems on Windows pipeline, if I stop the previous run, the next run will fail at getting sources.

trivialfis:
Back ported the changes in prediction cache.  Added a small Python test.

PR #5087, opened 2019-12-04
Fix parsing empty vector in parameter.

trivialfis:
Small part of https://github.com/dmlc/xgboost/pull/4732 .

### DIFF
--- a/src/tree/param.cc
+++ b/src/tree/param.cc
@@ -3,18 +3,20 @@
  */
 #include <iostream>
 #include <vector>
+#include <utility>
 
 #include "param.h"
 
 namespace std {
 std::istream &operator>>(std::istream &is, std::vector<int> &t) {
+  t.clear();
   // get (
   while (true) {
     char ch = is.peek();
     if (isdigit(ch)) {
       int idx;
       if (is >> idx) {
-        t.assign(&idx, &idx + 1);
+        t.emplace_back(idx);
       }
       return is;
     }
@@ -73,7 +75,7 @@ std::istream &operator>>(std::istream &is, std::vector<int> &t) {
       return is;
     }
   }
-  t.assign(tmp.begin(), tmp.end());
+  t = std::move(tmp);
   return is;
 }
 }  // namespace std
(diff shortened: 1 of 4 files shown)


src/tree/param.h
```diff
@@ -2,7 +2,7 @@
  * Copyright 2014 by Contributors
  * \file param.h
  * \brief training parameters, statistics used to support tree construction.
- * \author Tianqi Chen
+n * \author Tianqi Chen
```
hcho3: Remove the stray letter in the beginning

src/tree/param.cc
```diff
@@ -0,0 +1,79 @@
...
+
+namespace std {
+std::istream &operator>>(std::istream &is, std::vector<int> &t) {
+  // get (
+  while (true) {
+    char ch = is.peek();
+    if (isdigit(ch)) {
+      int idx;
+      if (is >> idx) {
+        t.assign(&idx, &idx + 1);
```
hcho3: Just curious, is this better than `t = { idx };`?

src/tree/param.cc
```diff
@@ -0,0 +1,79 @@
...
+        break;
+      }
+    } else if (ch == ')') {
+      break;
+    } else {
+      is.setstate(std::ios::failbit);
+      return is;
+    }
+  }
+  t.assign(tmp.begin(), tmp.end());
```
hcho3: Is this better than `t = tmp`?

src/tree/param.cc
trivialfis: Honestly I don't know what assign does.  It's simply old code moved to a cc file to avoid recompiling half of XGBoost when debugging.  I only added a few lines handling empty vector.

hcho3: Got it. In that case, let’s use `=` operator for the sake of clarity.

src/tree/param.h
trivialfis: Done.

src/tree/param.cc
trivialfis: @hcho3 Used `emplace_back` with `clear` added at the beginning.

src/tree/param.cc
trivialfis: @hcho3 Used moved assign operator.

PR #5088, opened 2019-12-04
Implement training observer.

### LINKED ISSUE #4837: [RFC] Callback interface for logging internal information, to aid debugging
Recently, I've seen many users report regressions in model performance (e.g. accuracy, AUC) when they upgraded their XGBoost to latest version. For example, see https://discuss.xgboost.ai/t/learning-with-xgboost-0-90-vs-1-0-0/1068/5.

I'd like to work with @chenqin to set up [regression tests](https://github.com/dmlc/xgboost/issues/4680#issuecomment-524034688). As part of that, we need more than just traces of evaluation set metrics (accuracy, AUC) etc, since the metrics only capture the aggregate information. Many useful information such as gradient values, split scores, and quantile sketches remain hidden deep within the C++ codebase.

I propose to create a **callback interface** so as to expose useful information for verbose logging. Once we can log these pieces of information, we can use them to **detect potential regressions and troubleshoot anomalies**. Logging the **traces of non-aggregate quantities** will allow us to have closer look at the internal workings of XGBoost. We can even build a "debugger" for users to find out why their XGBoost models are performing terribly.

For each of the quantities listed above, XGBoost will call the callback function with the quantity as an argument. Here are the possible callback interfaces (there’s more than one since quantities are of different types):

* `void callback(const char* keyword, const float* array, const uint64_t* shape, int dim)`
* `void callback(const char* keyword, const double* array, const uint64_t* shape, int dim)`
* `void callback(const char* keyword, const GradientPair* array, const uint64_t* shape, int dim)`
    - GradientPair is defined to be the C structure consisting of two float fields.

Here is what I envision to be the user experience:
```python
import xgboost

# Note. some details are elided for brevity

arrays = {}

def handle_float_array(keyword, array, shape, dim):
  arrays[keyword] = np.array(array, shape=shape)

bst = xgboost.train( params={ ... },
                     callbacks=[ ..., debug_logging_callback(handle_float_array)] )

# Now arrays should contain a bunch of arrays extracted from XGBoost internals
```

Each quantity will have a unique keyword associated, e.g. `"gradient"`, `"hessian"`, `"histogram"`, `"quantile_sketch"`, etc.

Requirements

* Don't pay for what you don't use: This feature shall impose no runtime overhead unless it is turned on.
* Should work in distributed environments, e.g. Spark, Dask, Kubernetes

List of quantities we should log

* `yhat_i`: Predicted labels (model outputs) for data points (Per data point, Per boosting iteration) 
* `x_i` - Input
* `l(y_i, yhat_i)`: Loss value (“residual”) computed from true and predicted labels (Per data point, Per boosting iteration)
  - We can use this to highlight rows that each decision tree handles well and not well.
 * Gradient and hessian values computed from the loss value (Per data point, Per boosting iteration)
* Histogram (partial sums) over gradient and hessian values (one histogram per individual split node)
  - We can use this to figure out why and how individual split is chosen (which feature? which threshold?)
* Quantile sketches (Per feature)
  - We can visualize estimated feature quantiles and determine if the estimates are reasonable given the actual distribution of data. Quantile estimates are used to speed up training.

trivialfis:
Closes #4837  Small part of https://github.com/dmlc/xgboost/pull/4732 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,8 @@ option(USE_OPENMP "Build with OpenMP support." ON)
 option(JVM_BINDINGS "Build JVM bindings" OFF)
 option(R_LIB "Build shared library for R package" OFF)
 ## Dev
+option(USE_DEBUG_OUTPUT "Dump internal training results like gradients and predictions to stdout.
+Should only be used for debugging." OFF)
 option(GOOGLE_TEST "Build google tests" OFF)
 option(USE_DMLC_GTEST "Use google tests bundled with dmlc-core submodule" OFF)
 option(USE_NVTX "Build with cuda profiling annotations. Developers only." OFF)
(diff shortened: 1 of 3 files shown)


hcho3:
This is going to be quite useful for developers who want to implement new objectives and metrics. Thanks!

trivialfis:
Please note that as a small part of JSON PR, you can't use this right now as the observer uses JSON for dumping tree model and XGBoost parameter.

hcho3:
@RAMitchell Recently I have been working to add a new objective, and often I would find it useful to print out intermediate variables like gradients. The benefit is that you get trace of a variable while training is proceeding as usual.

trivialfis:
@hcho3 If you use the observer, you don't have to enable the cmake flag and recompile the whole world, just change this particular file.  The CMake flag is mostly for new comer.

Issue #4357, opened 2019-04-10
[RFC] External memory support for GPU


rongou:
## Motivation
XGBoost has experimental support for [external memory](https://xgboost.readthedocs.io/en/latest/tutorials/external_memory.html) that allows for training on datasets that don’t fit in main (a.k.a. host) memory. However, it’s not available with GPU algorithms. To train on large datasets with GPUs, we have to either downsample the data, which defeats the purpose, or scale out to multi-gpu and multi-node settings, which have their own complexities and limitations. It’s desirable to make the GPU algorithms more flexible by adding external memory support.

## Goals
* Support external memory for the `gpu_hist` tree method.
* Provide reasonable performance.

## Non-Goals
This proposal is for XGBoost only, and doesn’t apply to other libraries in RAPIDS such as cuDF/cuML.

## Assumptions
* We’ll build on top of the existing external memory version in XGBoost, which only supports importing from the `libsvm` format.
* We’ll rely on existing cache systems to leverage host memory, for example, Linux page cache for local files, FS-Cache for network filesystems.
* Although the approach should work if data is spilled to disk, for reasonable performance, we assume the whole training dataset fits in host memory, thus is cached after the first round.

## Risks
* A naive implementation may be unacceptably slow.
* A fully tuned implementation that is asynchronous and fully pipelined may introduce a lot of complexity to the code base.

### Existing Code
The current XGBoost code allows the user to specify the input data as
```
filename#cacheprefix
```
where `filename` is the normal path to a `libsvm` file, and `cacheprefix` is a path to a cache file that XGBoost will use for external memory cache. During training, the data is read in, parsed, and written out to the cache in 32MB pages. For each round, the pages are read back in and fed to the learner as batches.

Currently the `gpu_hist` tree method only accepts the first batch and errors out if there are more than one batch. In the training loop, the data is initialized once by going through feature quantile generation and data compression. The compressed feature quantiles are kept in GPU memory during training.

The GPU algorithm is described in detail in the [paper](https://arxiv.org/abs/1806.11248).

Adding external memory support can be split into the following phases.

### Phase 1: Basic Correctness
Here we aim to support external memory in the simplest way:
[...]

PR #5093, opened 2019-12-06
Gradient based sampling for external memory mode on GPU

rongou:
In GPU external memory mode, rely on gradient-based sampling to allow for bigger datasets. The final step for #4357.

The main idea is from https://arxiv.org/abs/1803.00841.

High level design:
* At the beginning of each round, take the gradient pairs, perform weighted sampling without replacement based on the absolute value of the gradient.
* Compact all the sampled rows in the DMatrix into a single ELLPACK page.
* Construct the tree the same way as in-memory mode, using the compacted page that's kept in memory.
* After the tree is constructed, loop through all the original pages to finalize the position of all the rows.
* Repeat for each round.

On a generated synthetic dataset (1 million rows, 50 features), gradient-based sampling still works down to about 10% of the total rows sampled with loss of accuracy, while uniform random sampling only works at 50% of data, and completely fails to converge at 10%.

On a larger synthetic dataset (30 million rows, 200 features), CPU external memory mode actually runs out of memory on my 32GB desktop, while GPU external memory mode works fine on a Titan V (12GB).

Some benchmark numbers for the synthetic 1 million row, 50 feature dataset:
```
number of rounds: 500
max_depth = 0
max_leaves = 256
grow_policy = lossguide
```

|Mode|Eval Error|Time (seconds)|
|---|---:|---:|
|CPU in-core|0.0083|508.40|
|CPU external memory|0.0096|513.58|
|GPU in-core|0.0086|31.13|
|GPU external memory, sample all rows|0.0093|46.30|
|GPU external memory, sample 50%|0.0081|124.44|
|GPU external memory, sample 10%|0.0085|121.29|
|GPU external memory, sample 5%|0.0098|120.48|


hcho3:
@rongou Can you provide a summary of what this pull request does? Why a new kind of sampling?

trivialfis:
@hcho3 I think it's this one: https://arxiv.org/abs/1901.09047

rongou:
@hcho3 @trivialfis I'll write a more detailed summary once I'm confident that it's actually working properly. The main reason for this is that the naive implementation of external memory mode on GPU requires reading back all the pages for every tree node, which is pretty expensive since data are moved over the PCIe bus. By doing some kind of smart sampling and keeping the sampled page in GPU memory, we can hope to get reasonable performance without degrading the accuracy too much. The main idea is from https://arxiv.org/abs/1803.00841, but LightGBM also does something similar.

include/xgboost/base.h
```diff
@@ -193,6 +193,24 @@ class GradientPairInternal {
...
+  XGBOOST_DEVICE GradientPairInternal<T> operator*(float multiplier) const {
+    GradientPairInternal<T> g;
+    g.grad_ = grad_ * multiplier;
+    g.hess_ = hess_ * multiplier;
+    return g;
+  }
+
+  XGBOOST_DEVICE GradientPairInternal<T> operator/(float divider) const {
+    GradientPairInternal<T> g;
+    g.grad_ = grad_ / divider;
```
trivialfis: Be careful for 0 division.

hcho3:
Thanks a lot for the paper references.

hcho3:
@rongou @trivialfis I'm at NeurIPS 2019 and am actually meeting the author of https://arxiv.org/abs/1901.09047 today. Wish me luck.

src/tree/gpu_hist/gradient_based_sampler.cu
```diff
@@ -0,0 +1,225 @@
...
+                                           SamplingMethod sampling_method)
+    : batch_param_(batch_param), info_(info), sampling_method_(sampling_method) {
+  monitor_.Init("gradient_based_sampler");
+
+  // If `subsample` is not specified, try to figure out how many rows to sample based on available
+  // free GPU memory.
+  if (subsample == 0.0f || subsample == 1.0f) {
+    sample_rows_ = MaxSampleRows(n_rows);
+  } else {
+    sample_rows_ = n_rows * subsample;
```
sriramch: what if there isn't enough memory to honor the user-defined subsample % of rows? should this instead be the following in *addition* to this?

`sample_rows_ = std::min(sample_rows_, MaxSampleRows(sample_rows_));`

followed by some info message saying that the user thresholds aren't being honored, if it is being scaled down?

src/tree/gpu_hist/gradient_based_sampler.cu
```diff
@@ -0,0 +1,225 @@
...
+                 thrust::counting_iterator<size_t>(n_rows),
+                 dh::tbegin(row_index_));
+  }
+}
+
+// Determine the maximum number of rows that can fit into the available GPU memory.
+size_t GradientBasedSampler::MaxSampleRows(size_t n_rows) {
+  size_t available_memory = dh::AvailableMemory(batch_param_.gpu_id);
+  // Subtract row_weight_, row_index_, and sample_row_index_.
+  available_memory -= n_rows * (sizeof(float) + 2 * sizeof(size_t));
```
sriramch: if `available_memory` is less to accommodate even these indices for the number of rows in the dataset, this can return a negative number which will then be casted into an unsigned value resulting in a weird 2 complement value messing with the estimation.

src/tree/gpu_hist/gradient_based_sampler.cu
```diff
@@ -0,0 +1,225 @@
...
+  size_t usable_memory = available_memory * 0.7;
```
sriramch: is 70% a best guess? is this to accommodate additional overhead incurred during training? iow, would this hold true even for much much larger data sets? should this be made configurable, or determined based on some other heuristic?

src/tree/gpu_hist/gradient_based_sampler.cu
```diff
@@ -0,0 +1,225 @@
...
+                    dh::tbegin(row_weight_),
+                    CalculateWeight(common::GlobalRandom()()));
+  return WeightedSampling(gpair, dmat);
+}
+
+// Perform sampling after the weights are calculated.
+GradientBasedSample GradientBasedSampler::WeightedSampling(
+      common::Span<xgboost::GradientPair> gpair, DMatrix* dmat) {
+  // Sort the gradient pairs and row indexes by weight.
+  thrust::sort_by_key(dh::tbegin(row_weight_), dh::tend(row_weight_),
```
sriramch: to sample a requisite number of rows from the dataset, i see a number of items that are still pegged to `O(num_rows)` *here*.  for instance, this specific line (radix sort) isn't sorting it in-place - meaning, this is going to need auxiliary device memory to sort the row weights and to associate them with the corresponding gradient pairs and its row indices. this will still perhaps need *at least* O(num_rows) * sizeof(item) additional device memory for each entity being sorted and associated.

a simple back-of-the-envelope calculation for 200m rows will require ~ 3.7gb to hold the row weights, indices etc. ~ 3gb of temp buffer during sort. this will only go much higher for bigger datasets.

perhaps, heap or other sorting algorithm may help, if we can trade time for space. but, even more fundamentally, maintaining the entire meta-data as a function of the size of the dataset in the device memory may not scale for larger ones. perhaps, doing this on host on bringing only the sample row indices may alleviate this, at the expense of speed, if it's tolerable.

src/tree/gpu_hist/gradient_based_sampler.cu
```diff
@@ -0,0 +1,225 @@
...
+                         dh::tbegin(sample_row_index_));
+  thrust::transform(dh::tbegin(gpair), dh::tend(gpair),
+                    dh::tbegin(sample_row_index_),
+                    dh::tbegin(sample_row_index_),
+                    ClearEmptyRows());
+
+  // Compact the ELLPACK pages into the single sample page.
+  thrust::fill(dh::tbegin(page_->gidx_buffer), dh::tend(page_->gidx_buffer), 0);
+  for (auto& batch : dmat->GetBatches<EllpackPage>(batch_param_)) {
+    page_->Compact(batch_param_.gpu_id, batch.Impl(), sample_row_index_);
```
sriramch: is there some benefit to bailing out sooner once we have the target sample number of rows compacted in, by having the compact method return number of rows compacted in this batch, and using that to decide? if these samples are going to be distributed near evenly across all pages, this may not help.

src/tree/gpu_hist/gradient_based_sampler.cu
```diff
@@ -0,0 +1,225 @@
...
+    rng.discard(i);
+    return dist(rng);
+  }
+};
+
+GradientBasedSample GradientBasedSampler::UniformSampling(common::Span<GradientPair> gpair,
+                                                          DMatrix* dmat) {
+  // Generate random weights.
+  thrust::transform(thrust::counting_iterator<size_t>(0),
+                    thrust::counting_iterator<size_t>(0) + gpair.size(),
```
sriramch: how about simply `thrust::counting_iterator<size_t>(gpair.size())`

src/tree/gpu_hist/gradient_based_sampler.cuh
```diff
@@ -0,0 +1,83 @@
...
+
+  /*! \brief Sample from a DMatrix based on the given gradients. */
+  GradientBasedSample Sample(common::Span<GradientPair> gpair, DMatrix* dmat);
+
+ private:
+  static const SamplingMethod kDefaultSamplingMethod = kSequentialPoissonSampling;
+
+  GradientBasedSample NoSampling(common::Span<GradientPair> gpair, DMatrix* dmat);
+  GradientBasedSample SequentialPoissonSampling(common::Span<GradientPair> gpair, DMatrix* dmat);
+  GradientBasedSample UniformSampling(common::Span<GradientPair> gpair, DMatrix* dmat);
```
sriramch: is this used someplace, or simply kept so that it can be used for comparison when needed?

src/tree/updater_gpu_hist.cu
```diff
@@ -540,13 +551,19 @@ struct GPUHistMakerDevice {
...
               GradientPair());
+
+    if (use_gradient_based_sampling) {
+      auto sample = sampler->Sample(dh_gpair->DeviceSpan(), dmat);
+      n_rows = sample.sample_rows;
+      page = sample.page;
+      gpair = sample.gpair;
+    } else {
+      gpair = dh_gpair->DeviceSpan();
+      SubsampleGradientPair(device_id, gpair, param.subsample);
```
sriramch: this *may* alter gradient pairs. hence, if the caller is expecting this to be unaltered, this may result in a behavioral change. the boost methods are assuming this to be unaltered - more so, when there is more than a single updater. i'm not sure if such a configuration is practical - but, it looks to be there nonetheless.

my guess is that this is perhaps the reason why it is being copied in this specific tree updater.

rongou:
Here is the training graph for Higgs (split 0.95/0.05):
```
    'max_depth': 8,
    'learning_rate': 0.1,
    'reg_lambda': 1,
    'max_leaves': 256,
    'eval_metric': 'auc',
    'tree_method': 'gpu_hist',
    'gpu_id': 1,
```
Using the MVS formula `weight = sqrt(grad^2 + 0.1 * hess^2)`.

Looks like 10% sampling flattens out after a while, but everything else outperforms the full sample.
MVS

rongou:
Here is the same chart with the first 100 rounds removed.
higgs

hcho3:
@rongou It is very interesting that we can get superior convergence rate with subsampling than not. I wonder if XGBoost was overfitting prior to introduction of sampling, and sampling is acting as regularization.

hcho3:
https://arxiv.org/abs/1901.09047 has a lot of goodies: 1) better sampling methods; 2) early stopping with sequential hypothesis testing. Time to study statistics, I suppose.

rongou:
@hcho3 My latest idea came from this NeurIPS paper: https://papers.nips.cc/paper/9645-minimal-variance-sampling-in-stochastic-gradient-boosting.

rongou:
The Higgs dataset probably needs some hyperparameter tuning. I ran the `subsample=0.6` one to 5,000 rounds, looks like the AUC was still climbing (final round 0.8538).

trivialfis:
@rongou 

Am I miss understanding something?


GPU external memory, sample all rows | 0.0093 | 46.30
-----------------------------------------------------| --------- | -------

Sampling all rows is faster than smaller sample rate?

hcho3:
@rongou Thanks for the link. I actually did see that poster at NeurIPS last week. I will read the paper.

src/data/ellpack_page.cuh
```diff
@@ -71,6 +71,11 @@ struct EllpackInfo {
                        size_t row_stride,
                        const common::HistogramCuts& hmat,
                        dh::BulkAllocator* ba);
+
+  /*! \brief Return the total number of symbols (total number of bins plus 1 for not found). */
+  inline size_t NumSymbols() const {
```
trivialfis: No need for inline.

src/data/ellpack_page.cu
```diff
@@ -96,6 +110,85 @@ EllpackPageImpl::EllpackPageImpl(DMatrix* dmat, const BatchParam& param) {
...
+    size_t dst_offset = dst_row * row_stride;
+    size_t src_offset = row_id * row_stride;
+    for (size_t j = 0; j < row_stride; j++) {
+      cbw.AtomicWriteSymbol(dst_data_d, src_iterator_d[src_offset + j], dst_offset + j);
+    }
+  }
+};
+
+// Compacts the data from the given EllpackPage into the current page.
+void EllpackPageImpl::Compact(int device, EllpackPageImpl* page, common::Span<size_t> row_indexes) {
```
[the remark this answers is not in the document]
rongou: I guess the meanings are slightly different. In `Push()` we are taking a whole page and adding to the current one, but here we are only selectively adding some rows.

src/tree/gpu_hist/row_partitioner.cuh
```diff
@@ -125,7 +125,6 @@ class RowPartitioner {
       idx += segment.begin;
       RowIndexT ridx = d_ridx[idx];
       bst_node_t new_position = op(ridx);  // new node id
-      if (new_position == kIgnoredTreePosition) return;
```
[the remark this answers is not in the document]
rongou: In `UpdatePosition()` we are back to only dealing with a single `EllpackPage` now.

src/tree/gpu_hist/gradient_based_sampler.cu
rongou: As @RAMitchell suggested, maybe we should just get rid of guessing available memory and let user set `subsample` explicitely.

src/tree/gpu_hist/gradient_based_sampler.cu
rongou: Added a check.

src/tree/gpu_hist/gradient_based_sampler.cu
rongou: Yeah just a guess from a few experiments. Probably should just get rid of this method.

src/tree/gpu_hist/gradient_based_sampler.cu
rongou: Yes it's not completely "out of core", but I think we are going to run out of other resources (e.g. main memory) before hitting the GPU memory limit. As it stands now, I actually can't generate a large enough dataset on my desktop to test the external memory mode, had to spin up a high-memory instance on GCP.

src/tree/gpu_hist/gradient_based_sampler.cu
rongou: It seems unlikely we won't have a row in a 32MB page.

src/tree/gpu_hist/gradient_based_sampler.cu
rongou: Done.

src/tree/gpu_hist/gradient_based_sampler.cuh
rongou: Not used by real code.

src/tree/updater_gpu_hist.cu
rongou: Probably should get rid of this in favor of the new gradient-based sampling.

src/data/ellpack_page.cuh
rongou: Done.

rongou:
@RAMitchell @trivialfis @sriramch @hcho3 Addressing all the questions together.

> What is the user interface to this functionality? Does it have parameters, is this mode exclusive to other parameter settings? Does it need to be used with external memory mode? Also note that automatically configuring parameters (e.g. trying to guess available memory to determine sample rows) is something that has caused us a lot of problems in the past.

For now this is only enabled in gpu external memory mode. The only parameter is `subsample`. We can get rid of the guessing available memory part and let user explicitly set `subsample`. What do you think?

> How can we achieve maximal conceptual separation between the idea of sampling, the learning algorithm and the input data source? Ideally the tree algorithm would have no idea if the incoming data is being sampled or not, it just gets given data to process. Can we use this algorithm with SimpleDMatrix as well as SparsePageDMatrix or potentially some new format?

This is more or less the case already. The only exception is in `FinalisePosition()` we have to go through the whole, unsampled data. One way to make it more transparent is to have a separate `RowPartitioner` for the whole dataset, but that'll consume more memory.

I've left the existing uniform random sampling code for SimpleDMatrix alone as I'm not sure if there are existing users. If you think we should use the new approach for both DMatrixes, I can get rid of that code.

> it may be useful to run some benchmarks on larger (not necessarily wider) datasets, as this specifically tries to make it work out-of-core, to see what the limits are.

The purpose of the 30 million row 200 feature dataset was to test the limit. I can generate another dataset with fewer features (maybe 50?).

> Sampling all rows is faster than smaller sample rate?

When "sampling" all rows, the whole ELLPACK matrix is kept in memory, not resampled/reconstructed every round, so it's faster.

> Is the old external memory mode removed?

There is not an "old" external memory mode per se. This is really the last step in making GPU external mode usable in practice.

RAMitchell:
How about this as the interface, you have the `subsample` parameter which acts normally and an enum parameter `subsample_method` that allows your method to be selected. Ideally it should work regardless of external memory mode or not.

If we could make predictions using EllPack matrices in `gpu_predictor` we wouldn't need to call `FinalisePosition()` and rely on prediction caching, although I'm not sure how to achieve that just yet.

trivialfis:
Would you like to write a tutorial for it in doc?  For advertising and our education?

src/tree/gpu_hist/gradient_based_sampler.cu
sriramch: system memory usage is typically less when external memory mode is enabled, as the entire dataset isn't made resident (specifically during sketching), and only parts are streamed in incrementally. i think it may be worthwhile to check how much more data we can accommodate over and above the gpu memory with this pr. i still think that maintaining the meta data for the dataset in the device memory this way to sample will turn out to be more expensive.

perhaps, the same dataset could be concatenated a few times to create a large dataset just for the purposes of determining the limits.

rongou:
@RAMitchell that sounds like a good interface. One question, do you think we need to implement gradient-based sampling for CPU tree updaters? Looks like sampling is done in several places. If the answer is yes, can that wait until a follow-up PR?

@trivialfis what I meant was the external memory mode for GPU had been a work-in-progress and wasn't very usable without sampling. The interface for external memory mode (both CPU and GPU) is still the same so the existing guide is still valid. I can add some wording there for enabling gradient-based sampling, would that be enough?

trivialfis:
Yup.  I was referring to the introduction of used algorithms etc.

rongou:
@RAMitchell I added a new `sampling_method` parameter, it works for both in-core and external memory modes now. Please take another look.

trivialfis:
Just parsed the papers today.  Will dig into the details tomorrow.

src/tree/gpu_hist/row_partitioner.cuh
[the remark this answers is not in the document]
RAMitchell: Does this mean you are abandoning external memory in the traditional sense?

src/tree/gpu_hist/gradient_based_sampler.cu
```diff
@@ -0,0 +1,286 @@
...
+    : original_page_(page),
+      batch_param_(batch_param),
+      is_external_memory_(page->matrix.n_rows != n_rows),
+      subsample_(subsample),
+      is_sampling_(subsample < 1.0),
+      sampling_method_(sampling_method),
+      sample_rows_(n_rows * subsample) {
+  monitor_.Init("gradient_based_sampler");
+
+  if (is_external_memory_) {
```
RAMitchell: If you have to have a bool and an if statement in every method, I think it suggests that you need two separate classes.

src/tree/gpu_hist/gradient_based_sampler.cu
```diff
@@ -0,0 +1,286 @@
...
+  thrust::fill(dh::tbegin(sample_row_index_), dh::tbegin(sample_row_index_) + sample_rows_, 1);
+  thrust::fill(dh::tbegin(sample_row_index_) + sample_rows_, dh::tend(sample_row_index_), 0);
+
+  // Sort the gradient pairs and sample row indexes by the original row index.
+  thrust::sort_by_key(dh::tbegin(row_index_), dh::tend(row_index_),
+                      thrust::make_zip_iterator(thrust::make_tuple(dh::tbegin(gpair),
+                                                                   dh::tbegin(sample_row_index_))));
+
+  // Compact the non-zero gradient pairs.
+  thrust::copy_if(dh::tbegin(gpair), dh::tend(gpair), dh::tbegin(gpair_), IsNonZero());
```
RAMitchell: Is it possible that less values are copied than sample_rows_?

tests/cpp/tree/test_gpu_hist_sampling.cc
```diff
@@ -0,0 +1,128 @@
...
+        DMatrix::Load(ext_mem_file, true, false, "auto", kPageSize));
+  }
+
+  static void TearDownTestCase() {
+    dmat.reset();
+    dmat_ext.reset();
+    delete temp_dir;
+  }
+
+  static void VerifyPredictionMean(const std::shared_ptr<DMatrix>& dtrain,
```
rongou: @RAMitchell is this close to what you had in mind to verify samples provide unbiased estimate?

src/tree/updater_colmaker.cc
```diff
@@ -202,6 +202,8 @@ class ColMaker: public TreeUpdater {
         }
         // mark subsample
         if (param_.subsample < 1.0f) {
+          CHECK_EQ(param_.sampling_method, TrainParam::kUniform)
+            << "Only uniform sampling is supported";
```
trivialfis: Or `gradient based sampling is only supported by GPU Hist.`

src/tree/gpu_hist/row_partitioner.cuh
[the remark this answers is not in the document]
trivialfis: @RAMitchell @rongou By looking at the code, this two should be orthogonal.  Correct me if I'm wrong.

src/tree/gpu_hist/gradient_based_sampler.cu
```diff
@@ -0,0 +1,286 @@
...
+  }
+};
+
+/*! \brief A functor that clears the row indexes with empty gradient. */
+struct ClearEmptyRows : public thrust::binary_function<GradientPair, size_t, size_t> {
+  XGBOOST_DEVICE size_t operator()(const GradientPair& gpair, size_t row_index) const {
+    if (gpair.GetGrad() != 0 || gpair.GetHess() != 0) {
+      return row_index;
+    } else {
+      return SIZE_MAX;
```
trivialfis: Maybe use `std::numeric_limit` instead.

tests/cpp/tree/test_gpu_hist_sampling.cc
```diff
@@ -0,0 +1,128 @@
...
+    VerifyPredictionMean(dmat, subsample, "gradient_based");
+  }
+}
+
+TEST_F(GpuHistSamplingTest, GradientBasedSampling_ExternalMemory) {
+  for (int i = 1; i < 10; i++) {
+    float subsample = static_cast<float>(i) / 10.0f;
+    VerifyPredictionMean(dmat_ext, subsample, "gradient_based");
+  }
+}
```
trivialfis: As usual,  please add a null test with 0 number of rows. ;-)

trivialfis:
@rongou Let us know when it's ready.

src/tree/gpu_hist/gradient_based_sampler.cu
rongou: I changed the in-memory implementation to match external memory. The only special handling now is when not sampling, in external memory mode we concatenate the pages together.

src/tree/gpu_hist/row_partitioner.cuh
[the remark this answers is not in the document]
rongou: @RAMitchell I guess you could say that. Streaming the whole dataset for every tree node just seems too expensive, at least with the current PCIe bandwidth/latency. If the GPU is more tightly integrated with the CPU (e.g. IBM Power9) it might be worth reconsidering.

src/tree/gpu_hist/gradient_based_sampler.cu
rongou: Done.

src/tree/gpu_hist/gradient_based_sampler.cu
rongou: Yes it's possible. Added a line to fill `gpair_` with empty gpairs.

src/tree/updater_colmaker.cc
rongou: Done.

tests/cpp/tree/test_gpu_hist_sampling.cc
rongou: I'm not sure these tests are testing the right things. Removed for now.

RAMitchell:
Even if it's ready, I think we won't merge this until after 1.0 release due to some risky changes to our core algorithms. This shouldn't be too far away though.

rongou:
The code is ready, but looks like there is a failing test involving `XGBRFClassifier`. Need to do some debugging.

rongou:
@RAMitchell @trivialfis I refactored the code, the existing behavior is preserved for uniform sampling. This PR actually reverts some of the changes made to `gpu_hist` in my last PR, so it's arguably less risky.

trivialfis:
@rongou The next release should be really close.  1 last remaining blocker being model IO of Scikit-Learn interface.  I will try to merge this PR once we can split up the branch.  It should be fine as I believe we can make it for the next rapids release.

src/tree/updater_gpu_hist.cu
```diff
@@ -719,35 +714,46 @@ struct GPUHistMakerDevice {
...
-                  position = node.LeftChild();
-                } else {
-                  position = node.RightChild();
-                }
-              }
-              node = d_nodes[position];
-            }
-            return position;
-          });
+    if (row_partitioner->GetRows().size() != p_fmat->Info().num_row_) {
```
trivialfis: What's the context of this condition?

rongou: When sampling in external memory mode, the EllpackPage is compacted (rows with zero gradients are removed).

trivialfis:
@RAMitchell I think it's ready for merging.  WDYT?

trivialfis:
Restarted the test.  Will merge unless @rongou has other comments regarding @RAMitchell 's review.

src/tree/gpu_hist/gradient_based_sampler.cu
```diff
@@ -0,0 +1,380 @@
...
+ExternalMemoryNoSampling::ExternalMemoryNoSampling(EllpackPageImpl* page,
+                                                   size_t n_rows,
+                                                   const BatchParam& batch_param)
+    : batch_param_(batch_param),
+      page_(new EllpackPageImpl(batch_param.gpu_id, page->matrix.info, n_rows)) {}
+
+GradientBasedSample ExternalMemoryNoSampling::Sample(common::Span<GradientPair> gpair,
+                                                     DMatrix* dmat) {
+  if (!page_concatenated_) {
+    // Concatenate all the external memory ELLPACK pages into a single in-memory page.
```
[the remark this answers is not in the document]
rongou: I think this option is here mostly for completeness' sake. If the whole dataset fits in GPU memory, then presumably you wouldn't want to use external memory; if it doesn't fit, you probably want to play around with sampling.

I have noticed that writing out external pages and then concatenating them together might allow you to train on slightly larger datasets versus keeping everything in memory, probably because of lower working memory requirement. Not sure how useful that is though.

PR #5094, opened 2019-12-06
Implement JSON IO for updaters

trivialfis:
* Remove parameters in split evaluator.

Another small fraction of #4732 .

### DIFF
--- a/src/tree/split_evaluator.cc
+++ b/src/tree/split_evaluator.cc
@@ -46,7 +46,7 @@ SplitEvaluator* SplitEvaluator::Create(const std::string& name) {
 }
 
 // Default implementations of some virtual methods that aren't always needed
-void SplitEvaluator::Init(const Args& args) {}
+void SplitEvaluator::Init(const TrainParam* param) {}
 void SplitEvaluator::Reset() {}
 void SplitEvaluator::AddSplit(bst_uint nodeid,
                               bst_uint leftid,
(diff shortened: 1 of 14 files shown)


src/tree/split_evaluator.cc
```diff
@@ -175,22 +146,6 @@ XGBOOST_REGISTER_SPLIT_EVALUATOR(ElasticNet, "elastic_net")
     return new ElasticNet(std::move(inner));
   });
 
-/*! \brief Encapsulates the parameters required by the MonotonicConstraint
-        split evaluator
-*/
-struct MonotonicConstraintParams
-    : public XGBoostParameter<MonotonicConstraintParams> {
-  std::vector<bst_int> monotone_constraints;
```
hcho3: Is this field now part of TrainParam?

hcho3: Indeed it is: https://github.com/dmlc/xgboost/blob/df9bdbbcb9713389759293a67e225b0985a52346/src/tree/param.h#L75

trivialfis: @hcho3 Yes: https://github.com/dmlc/xgboost/blob/2dcb62ddfb2e10d58358195992df829e6d558c83/src/tree/param.h#L75

I think the one in TrainParam is older than the one in split evaluator.

hcho3: Nice job deduplicating this parameter. Having one source of parameters will make our life easier.

trivialfis:
@hcho3 

> Having one source of parameters will make our life easier.

After JSON PR, the only duplicated parameter is `max_delta_step` in `TrainParam` and `PoissonRegressionParam`.  We will be ready for some sort of parameter validation.

PR #5097, opened 2019-12-06
- performance fixes for dmatrix construction using adapters

sriramch:
- the parallel group builder isn't amenable for piece wise construction of a dmatrix
    from batches of data. it relies on accumulating data on each thread and consolidating
    at the end of it all. further, it also repeatedly resizes the internal vectors as and
    when new data trickles in. the resizes are extremely expensive, if a row 'x' is attempted
    to be added, where 'x' is large, and when there are little to no elements before 'x'
  - the piece-wise construction exemplifies this issue, when large batches of data are taken in.
    the future batches with large row indices will assume that there are going to be data
    coming in for older rows, and will try to make room for those, by carving out large
    chunks of memory, only to be thrown away very soon when smaller batches of data 
    are processed with large row indices
  - this approach will control the state of the parallel builder by pre-allocating the thread
    vectors and reusing them for every batch. it will further treat each batch as though it
    is the entire data, and once the build is complete, it will consolidate the batch data and the
    offset vectors into the dmatrix's data and offset vectors

without this fix, dmatrix construction of a dataset containing 100m rows takes hours
(i aborted the process after 4 hours). with this fix, it takes 30-40s.

### DIFF
--- a/src/data/simple_dmatrix.h
+++ b/src/data/simple_dmatrix.h
@@ -47,20 +47,28 @@ class SimpleDMatrix : public DMatrix {
     auto& data_vec = mat.page_.data.HostVector();
     uint64_t inferred_num_columns = 0;
 
+    size_t num_lines_processed = 0;
+
+    using OffsetVecT = std::remove_reference<decltype(offset_vec)>::type::value_type;
+    std::vector<std::vector<OffsetVecT>> tmp_thread_ptr;
+
     adapter->BeforeFirst();
     // Iterate over batches of input data
     while (adapter->Next()) {
       auto &batch = adapter->Value();
-      common::ParallelGroupBuilder<
-        Entry, std::remove_reference<decltype(offset_vec)>::type::value_type>
-        builder(&offset_vec, &data_vec);
-      builder.InitBudget(0, nthread);
+
+      // Batch data and its offsets
+      std::vector<OffsetVecT> batch_offset_vec;
+      std::vector<Entry> batch_data_vec;
+      common::ParallelGroupBuilder<Entry, OffsetVecT> builder(&batch_offset_vec,
+                                                              &batch_data_vec,
+                                                              &tmp_thread_ptr);
 
       // First-pass over the batch counting valid elements
       size_t num_lines = batch.Size();
+      builder.InitBudget(num_lines, nthread);
 #pragma omp parallel for schedule(static)
-      for (omp_ulong i = 0; i < static_cast<omp_ulong>(num_lines);
-        ++i) {  // NOLINT(*)
+      for (omp_ulong i = 0; i < static_cast<omp_ulong>(num_lines); ++i) {  // NOLINT(*)
         int tid = omp_get_thread_num();
         auto line = batch.GetLine(i);
         for (auto j = 0ull; j < line.Size(); j++) {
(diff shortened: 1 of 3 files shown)


trivialfis:
@sriramch I will run some tests tomorrow at work, thanks for fixing so fast!

src/data/simple_dmatrix.h
```diff
@@ -47,20 +47,28 @@ class SimpleDMatrix : public DMatrix {
...
+      // Batch data and its offsets
+      std::vector<OffsetVecT> batch_offset_vec;
+      std::vector<Entry> batch_data_vec;
+      common::ParallelGroupBuilder<Entry, OffsetVecT> builder(&batch_offset_vec,
+                                                              &batch_data_vec,
+                                                              &tmp_thread_ptr);
 
       // First-pass over the batch counting valid elements
       size_t num_lines = batch.Size();
+      builder.InitBudget(num_lines, nthread);
```
RAMitchell: `num_lines` != number of rows. I built this as an abstraction that can deal with both CSC/CSR sparse matrices. In the case of CSC num_lines will be the number of columns.

RAMitchell: Perhaps instead of this we can use heuristic to initialise the expected number of rows (that should be exact for all current cases). Take the last element of the last "line" and use its row index.

src/data/simple_dmatrix.h
```diff
@@ -69,23 +77,22 @@ class SimpleDMatrix : public DMatrix {
...
-        ++i) {  // NOLINT(*)
+      for (omp_ulong i = 0; i < static_cast<omp_ulong>(num_lines); ++i) {  // NOLINT(*)
         int tid = omp_get_thread_num();
         auto line = batch.GetLine(i);
         for (auto j = 0ull; j < line.Size(); j++) {
           auto element = line.GetElement(j);
           if (!common::CheckNAN(element.value) && element.value != missing) {
-            builder.Push(element.row_idx, Entry(element.column_idx, element.value),
-              tid);
+            builder.Push(element.row_idx - num_lines_processed,
```
RAMitchell: Again, num_lines_processed != num_rows_processed.

src/data/simple_dmatrix.h
trivialfis: @RAMitchell Should we revert the commit and prepare a new one?  Seems to be cleaner way.

sriramch: the csX adapters that are present today seem to be single batch adapters, in that they return only one batch in total. hence, the offsets to the row indices by the number of rows processed thus far may at best be moot. this *perhaps* could be the reason why all these tests are passing.

that said, it doesn't preclude someone to create an adapter that can return a csc formatted data as multiple batches in the future, and we should handle this, which the original version does not.

please check the new version, as this is being handled slightly differently. i have also added a csc adapter test that forces the adapters to infer sizes, thus in a way also ascertaining the integrity of the parallel build.

src/data/simple_dmatrix.h
sriramch: i also wonder if the parallel builder was build to handle this type of use-cases. i see them typically used for building csc type matrices. i do not know the history here, and thus the question. are we retrofitting this to solve the needs of the adapter (to also build csr style matrices)?

src/data/simple_dmatrix.h
RAMitchell: @trivialfis reverting is unnecessary.

src/data/simple_dmatrix.h
RAMitchell: Yes we are retrofitting it, it just wasn't designed to process more than one batch. Let me produce another PR for comparison that modifies the group builder to more efficiently handle this case.

PR #5098, opened 2019-12-07
Group builder modified for incremental building

RAMitchell:
This is an alternative fix to #5097, it addresses the issue of slow performance when building large files due to the group builder repeatedly creating large temporary vectors.

It adds the optional parameter of base_row_offset_, so when building incrementally, temporary thread vectors are sized relative to this instead of 0.

I used the following test (not committed) to measure performance:
```c++
TEST(SimpleDMatrix, Performance) {
  std::string filename = "test.libsvm";
  CreateBigTestData(filename, 100000000);
  std::unique_ptr<dmlc::Parser<uint32_t>> parser(
    dmlc::Parser<uint32_t>::Create(filename.c_str(), 0, 1, "auto"));
  data::FileAdapter adapter(parser.get());
  common::Timer t;
  data::SimpleDMatrix dmat(&adapter, std::numeric_limits<float>::quiet_NaN(),
    1);
  t.Stop();
  t.PrintElapsed("SimpleDMatrix");
  }
}
```

In this example it goes from 25s->6.9s.

@sriramch can you check performance on your example please?

### DIFF
--- a/src/data/simple_dmatrix.h
+++ b/src/data/simple_dmatrix.h
@@ -51,9 +51,10 @@ class SimpleDMatrix : public DMatrix {
     // Iterate over batches of input data
     while (adapter->Next()) {
       auto &batch = adapter->Value();
+
       common::ParallelGroupBuilder<
         Entry, std::remove_reference<decltype(offset_vec)>::type::value_type>
-        builder(&offset_vec, &data_vec);
+          builder(&offset_vec, &data_vec, offset_vec.size() - 1);
       builder.InitBudget(0, nthread);
 
       // First-pass over the batch counting valid elements
--- a/src/data/simple_dmatrix.h
+++ b/src/data/simple_dmatrix.h
@@ -52,10 +52,20 @@ class SimpleDMatrix : public DMatrix {
     while (adapter->Next()) {
       auto &batch = adapter->Value();
 
+      size_t base_row_offset = offset_vec.empty() ? 0 : offset_vec.size() - 1;
       common::ParallelGroupBuilder<
         Entry, std::remove_reference<decltype(offset_vec)>::type::value_type>
-          builder(&offset_vec, &data_vec, offset_vec.size() - 1);
-      builder.InitBudget(0, nthread);
+          builder(&offset_vec, &data_vec, base_row_offset);
+      // Estimate expected number of rows by using last element in batch
+      // This is not required to be exact but prevents unnecessary resizing
+      size_t expected_rows = 0;
+      if (batch.Size() > 0) {
+        auto last_line = batch.GetLine(batch.Size() - 1);
+        if (last_line.Size() > 0) {
+          expected_rows = last_line.GetElement(last_line.Size() - 1).row_idx;
+        }
+      }
+      builder.InitBudget(expected_rows, nthread);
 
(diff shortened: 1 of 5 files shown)


src/data/simple_dmatrix.h
```diff
@@ -51,9 +51,10 @@ class SimpleDMatrix : public DMatrix {
     // Iterate over batches of input data
     while (adapter->Next()) {
       auto &batch = adapter->Value();
+
       common::ParallelGroupBuilder<
         Entry, std::remove_reference<decltype(offset_vec)>::type::value_type>
-        builder(&offset_vec, &data_vec);
+          builder(&offset_vec, &data_vec, offset_vec.size() - 1);
```
sriramch: check for empty offset vec to pass 0.

tests/cpp/data/test_simple_dmatrix.cc
```diff
@@ -126,3 +126,32 @@ TEST(SimpleDMatrix, EmptyRow) {
   CHECK_EQ(dmat.Info().num_row_, 2);
   CHECK_EQ(dmat.Info().num_col_, 2);
```
sriramch: you may also consider changes that went into [test_adapter.cc](https://github.com/dmlc/xgboost/pull/5097/files#diff-75016bcb82696e4ac5cdb4a33ba99a8a)

RAMitchell:
@sriramch thanks, I will grab the tests. My rationale for moving the changes into the group builder comes from my newer PR #5092. I moved the code for pushing an adapter batch onto a sparse page into SparsePage.Push() (https://github.com/dmlc/xgboost/blob/cd0ddc52be58350291dda3a2b671a2d9a692b345/src/data/data.cc) so I could reuse it for sparse page DMatrix. It is easier to reuse if the group builder is self-contained in each operation and I don't have to share state.

I can apply the heuristic to resize the temporary thread vectors and I think that will deal with any potential problems with dynamically resizing those vectors.

RAMitchell:
@sriramch I'm not too worried about reallocating the temporary vectors for each batch. It seems like premature optimisation to try to reuse these across batches without a demonstrated case where this impacts performance.

Let me know what you think of the updated version.

sriramch:
> My rationale for moving the changes into the group builder comes from my newer PR #5092. I moved the code for pushing an adapter batch onto a sparse page into SparsePage.Push() (https://github.com/dmlc/xgboost/blob/cd0ddc52be58350291dda3a2b671a2d9a692b345/src/data/data.cc) so I could reuse it for sparse page DMatrix. It is easier to reuse if the group builder is self-contained in each operation and I don't have to share state.

[sc] this explanation makes the reason for the change into group builder much more clearer.

> 
> I can apply the heuristic to resize the temporary thread vectors and I think that will deal with any potential problems with dynamically resizing those vectors.

[sc] thanks and the new change looks good to me.

sriramch:
> @sriramch I'm not too worried about reallocating the temporary vectors for each batch. It seems like premature optimisation to try to reuse these across batches without a demonstrated case where this impacts performance.

[sc] i'll try and see if i can hack up something in dmlc-core such that the file parser carves out larger chunks of rows, which can then be fed to these iterators as row blocks.

PR #5101, opened 2019-12-08
Pass pointer to model parameters.

trivialfis:
This PR de-duplicates most of the model parameters except for the one in
`tree_model.h`.  One difficulty is `base_score` is a model property but can be
changed at runtime by objective function.  Hence when performing model IO, we
need to save the one provided by users, instead of the one transformed by
objective.  Here we created an immutable version of `LearnerModelParam` that
represents the value of model parameter after configuration.

Not so small part of #4732 .

### DIFF
--- a/src/common/hist_util.cc
+++ b/src/common/hist_util.cc
@@ -44,15 +44,11 @@ void HistogramCuts::Build(DMatrix* dmat, uint32_t const max_num_bins) {
   float constexpr kSparsityThreshold = 0.0005;
   // FIXME(trivialfis): Distributed environment is not supported.
   if (sparsity < kSparsityThreshold && (!rabit::IsDistributed())) {
+    LOG(INFO) << "Building quantile cut on a sparse dataset.";
     SparseCuts cuts(this);
     cuts.Build(dmat, max_num_bins);
   } else {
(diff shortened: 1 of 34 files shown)


src/common/hist_util.cc
```diff
@@ -44,11 +44,15 @@ void HistogramCuts::Build(DMatrix* dmat, uint32_t const max_num_bins) {
   float constexpr kSparsityThreshold = 0.0005;
   // FIXME(trivialfis): Distributed environment is not supported.
   if (sparsity < kSparsityThreshold && (!rabit::IsDistributed())) {
-    LOG(INFO) << "Building quantile cut on a sparse dataset.";
     SparseCuts cuts(this);
     cuts.Build(dmat, max_num_bins);
   } else {
-    LOG(INFO) << "Building quantile cut on a dense dataset or distributed environment.";
+    if (rabit::IsDistributed() && sparsity < kSparsityThreshold) {
```
RAMitchell: I don't understand how this relates to this PR?

trivialfis: Em.  Let me remove that.  I found the message not precise enough while testing model reproducibility.

PR #5105, opened 2019-12-09
Remove dead code in colmaker.

trivialfis:


hcho3:
@trivialfis How did you locate dead code? Does your IDE do this?

trivialfis:
@hcho3 No.  Sometimes `clangd` (a language server implementation) does that for me for unused function/variable, but not in this case.

PR #5111, opened 2019-12-11
JSON configuration IO.

trivialfis:
The "Configuration" part of https://github.com/dmlc/xgboost/pull/4732 .  See the `saving_model.rst` in #4732 for what does the configuration look like.

* Enable using JSON configuration in Python pickle.


trivialfis:
Made the naming more consistent across XGBoost.

hcho3:
I wonder why tests are failing. Is this WIP?

trivialfis:
I made some extra cleanup and found some problems in existing code.

hcho3:
Got it. Let me know if I can help in any way.

trivialfis:
I thought the JSON can enable a fully restored model from pickle, except I missed the prediction cache.  The DMatrix cache is missing.  It generates a very small (<1e-7) floating difference due to different floating point errors.  Also slower performance.

include/xgboost/parameter.h
```diff
@@ -80,6 +80,23 @@ class FieldEntry<EnumClass> : public FieldEntry<int> {  \
 }  /* namespace dmlc */
 
 namespace xgboost {
+
+template <typename P, typename Base, typename Container>
+Args UpdateAllowUnknownImpl(P* parameter, Base* base,
+                            Container const& kwargs, bool* out_changed = nullptr) {
+  static_assert(std::is_base_of<Base, P>::value, "");
+  if (parameter->GetInitialised()) {
```
hcho3: Nit: This is British spelling. An American would write `GetInitialized`. Shall we add the use of British spelling as part of the contributor's guide? My opinion: Since British English is what the rest of the world uses, we should enforce British spelling. Maybe a linter can help with this.

Some different spellings:

British | American
------|------
Initialise | Initialize
Colour | Color
Labour | Labor
Centre | Center
Metre | Meter

trivialfis: @hcho3 I think it's fine to use any of them.  I don't really notice where are contributors coming from.

hcho3: I suppose a good IDE should catch a misspelled variable before it causes compilation error.

src/gbm/gbtree.cc
```diff
@@ -289,6 +299,47 @@ void GBTree::CommitModel(std::vector<std::vector<std::unique_ptr<RegTree>>>&& ne
   monitor_.Stop("CommitModel");
 }
 
+void GBTree::LoadConfig(Json const& in) {
+  CHECK_EQ(get<String>(in["name"]), "gbtree");
+  fromJson(in["gbtree_train_param"], &tparam_);
+  int32_t const n_gpus = xgboost::common::AllVisibleGPUs();
+  if (n_gpus == 0 && tparam_.predictor == PredictorType::kGPUPredictor) {
+    tparam_.UpdateAllowUnknown(Args{{"predictor", "auto"}});
```
hcho3: YES! This is so much better than my previous hack in #3856

src/learner.cc
```diff
@@ -370,43 +425,9 @@ class LearnerImpl : public Learner {
...
-              << RenderParamVal(kv.second) << " for prediction. To "
-              << "override the predictor behavior, explicitly set '"
-              << saved_param << "' parameter as follows:\n"
-              << "  * Python package: bst.set_param('"
-              << saved_param << "', [new value])\n"
-              << "  * R package:      xgb.parameters(bst) <- list("
-              << saved_param << " = [new value])\n"
-              << "  * JVM packages:   bst.setParam(\""
-              << saved_param << "\", [new value])";
-          }
```
hcho3: (cry of joy)

include/xgboost/parameter.h
RAMitchell: I think it's a bit of a non-issue, it will affect workflow barely if at all.

python-package/xgboost/core.py
```diff
@@ -1076,28 +1076,44 @@ def __init__(self, params=None, cache=(), model_file=None):
...
         _check_call(_LIB.XGBoosterCreate(dmats, c_bst_ulong(len(cache)),
                                          ctypes.byref(self.handle)))
-        self.set_param({'seed': 0})
         self.set_param(params or {})
         if (params is not None) and ('booster' in params):
             self.booster = params['booster']
         else:
             self.booster = 'gbtree'
-        if model_file is not None:
+        if isinstance(model_file, Booster):
```
RAMitchell: I would love to see some in-line comments as to what is going on here.

trivialfis:
On a second thought, let me lower down the serialization code into c++, might get rid of one big FIXME.

trivialfis:
@hcho3 @RAMitchell I moved the concatenation into c++ from Python.  This saves me from the trouble of handling R code, also formalize the different between "Saving model" and "Serialization".

trivialfis:
@RAMitchell With the term "Model" it means saving only the tree, with the term "Config" it means all internal parameters, with the term "Serialization" it means save everything, the two of above combined.  

Saving config is just an extra bonus for showing the internal parameters of XGBoos t.

PR #5112, opened 2019-12-11
Small refinements for JSON model.

trivialfis:
* Addresses comments from @hcho3 in https://github.com/dmlc/xgboost/pull/5110 .
* Making the naming scheme more consistent across XGBoost.

### DIFF
--- a/tests/cpp/gbm/test_gblinear.cc
+++ b/tests/cpp/gbm/test_gblinear.cc
@@ -37,15 +37,15 @@ TEST(GBLinear, Json_IO) {
 
   model = Json::Load({model_str.c_str(), model_str.size()});
   ASSERT_TRUE(IsA<Object>(model));
-  model = model["model"];
 
   {
+    model = model["model"];
     auto weights = get<Array>(model["weights"]);
     ASSERT_EQ(weights.size(), 17);
   }
 
   {
-    model = Json::Load({model_str.c_str(), model_str.size()});
+    model = Json::Load(StringView{model_str.c_str(), model_str.size()});
     model = model["model"];
     auto weights = get<Array>(model["weights"]);
     ASSERT_EQ(weights.size(), 17);  // 16 + 1 (bias)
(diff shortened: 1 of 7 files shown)


trivialfis:
@hcho3 @RAMitchell

tests/cpp/gbm/test_gblinear.cc
```diff
@@ -37,15 +37,15 @@ TEST(GBLinear, Json_IO) {
...
 
   {
+    model = model["model"];
     auto weights = get<Array>(model["weights"]);
     ASSERT_EQ(weights.size(), 17);
   }
 
   {
-    model = Json::Load({model_str.c_str(), model_str.size()});
+    model = Json::Load(StringView{model_str.c_str(), model_str.size()});
```
hcho3: Still not sure why you are loading from `model_str` twice. I don't see `model_str` modified in the middle, so it's not a round-trip test.

trivialfis: @hcho3 Thanks for catching it.  Removed in latest commit.  That's an oversight.  As for the string view,  I have a better implementation in #5046 .  Will consider merging some parts of that optimization PR so that we have better performance without scarifying the API.

tests/cpp/test_learner.cc
```diff
@@ -143,7 +143,7 @@ TEST(Learner, Json_ModelIO) {
     for (int32_t iter = 0; iter < kIters; ++iter) {
       learner->UpdateOneIter(iter, p_dmat.get());
     }
-    learner->SetAttr("bset_score", "15.2");
+    learner->SetAttr("best_score", "15.2");
```
hcho3:
```suggestion
    learner->SetAttr("base_score", "15.2");
```

trivialfis: @hcho3 No, it's actually best score or best iteration.

trivialfis: @hcho3 Used by binding to set this training information.  The binary IO is misusing it for storing other parameters.

trivialfis: @hcho3 See : https://github.com/dmlc/xgboost/blob/208ab3b1ff22ac2d5d767961468016a17240268c/python-package/xgboost/callback.py#L219

hcho3: Got it.

PR #5050, opened 2019-11-18
Use DART tree weights when computing SHAPs

Kodiologist:
Fixes #5036.

I edited `CPUPredictor::PredictContribution` to multiply each tree's contributions by the tree's DART weight before adding them. This makes SHAP sums equal predictions, as expected. ~~The bug likely still exists for the GPU predictor and for interaction SHAPs.~~ Both non-interaction SHAPs and interaction SHAPs are affected.

This PR contains a commit from #5058 in order to get it to compile. The commit can be rebased out once #5058 is merged.

PR #5115, opened 2019-12-12
[Breaking] Don't drop trees during DART prediction by default

Kodiologist:
This is another attempt at #3556 to avoid the issue of people forgetting to set `ntreelimit` / `ntree_limit` (or not being aware that they have to) and getting incorrect predictions under DART.

I am aware that a change like this was previously reverted (#3563) because "it breaks the logic for custom evaluation and custom objective". However, no tests seem have been added or examples provided that demonstrate this breakage, so I have no idea whether this new PR recapitulates it. What I can say is that this PR passes the test suite for the R package, including with an assignment to `ntreelimit` removed.

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -145,7 +145,7 @@ xgb.iter.update <- function(booster_handle, dtrain, iter, obj = NULL) {
   if (is.null(obj)) {
     .Call(XGBoosterUpdateOneIter_R, booster_handle, as.integer(iter), dtrain)
   } else {
-    pred <- predict(booster_handle, dtrain)
+    pred <- predict(booster_handle, dtrain, training = TRUE)
     gpair <- obj(pred, dtrain)
     .Call(XGBoosterBoostOneIter_R, booster_handle, dtrain, gpair$grad, gpair$hess)
   }
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -288,7 +288,7 @@ xgb.Booster.complete <- function(object, saveraw = TRUE) {
 #' @export
 predict.xgb.Booster <- function(object, newdata, missing = NA, outputmargin = FALSE, ntreelimit = NULL,
                                 predleaf = FALSE, predcontrib = FALSE, approxcontrib = FALSE, predinteraction = FALSE,
-                                reshape = FALSE, ...) {
+                                reshape = FALSE, training = FALSE, ...) {
 
   object <- xgb.Booster.complete(object, saveraw = FALSE)
   if (!inherits(newdata, "xgb.DMatrix"))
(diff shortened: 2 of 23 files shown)


trivialfis:
@Kodiologist dart tests are failing.  I'm not entirely sure about this PR.  Will revisit dart later.

Kodiologist:
I've been working on this, but making slow progress. Apparently, `Dart::PredictBatch` is used for training as well as prediction with a trained model, so removing dropout from it entirely is a no-go. The new attempt I've pushed is passing the Python test, but now fails the SHAP-sum tests I provided for #5050.

trivialfis:
@Kodiologist Let me know if there's anything I can help.  I'm familiar with both C++ and Python layers.

hcho3:
> no tests seem have been added or examples provided that demonstrate this breakage

See https://github.com/dmlc/xgboost/pull/3556#issuecomment-410851113

hcho3:
@Kodiologist I think I fixed the failing R test. DART object internally keeps track of a sample of trees that are "dropped out", and this sample needs to be emptied whenever deterministic prediction is requested.

There's still an issue of enabling customized objective. Let's see if we can add a special flag for this use case.

Kodiologist:
I've tried to add a test for using custom evaluation functions with DART (see my last commit), but apparently, I can't get it to work: I've set it to return 3 in place of the actual negative log-likelihood and the predictions are coming out the same. This feature doesn't seem to be quite documented, but based on `tests/python/test_eval_metrics.py` and error messages I've gotten, it looks like you need to set `eval_metric` to the name `m` of an already existing metric, and then `feval` needs to return a tuple `(m, EXPR)` where `EXPR` is the computed evaluation; `feval` is then supposed to override the default implementation of `m`.

trivialfis:
Will look into this in these few days, see if I can help.  Thanks for the effort.

trivialfis:
Revisited the paper, will work on this PR tomorrow.

trivialfis:
Em.  So drop out prediction is required at training, but not inference.  But we need to expose this because of customized objective function.  Quite an headache.

trivialfis:
@Kodiologist I thought about this, the problem is prediction differs between training and inference.  This PR adds an extra parameter `drop_tree`, can we rename it to `training`, and set it the `True` when used with custom objective function?

trivialfis:
@hcho3 WDYT?

Kodiologist:
That makes sense to me.

RAMitchell:
@Kodiologist im happy to go ahead with this if you change the semantics per @trivialfis 's suggestion.

hcho3:
@trivialfis Yes, your suggestion makes sense.

Kodiologist:
@RAMitchell I'm happy to do so, but I can't seem to test it as I described above, so I won't know if it's correct any better than I do the current PR.

trivialfis:
Could you describe the issues you encountered while you are trying to run tests?

Kodiologist:
> I've tried to add a test for using custom evaluation functions with DART (see my last commit), but apparently, I can't get it to work: I've set it to return 3 in place of the actual negative log-likelihood and the predictions are coming out the same.

trivialfis:
Feel try to push the test here so I can take a look.

Kodiologist:
I did. Like I said, see my last commit, c9d7804c2f38e41547aa2fc62134b84cf6b50a1f.

Kodiologist:
I have, at least, renamed the parameters as requested.

trivialfis:
Let me try to fix it.

trivialfis:
@RAMitchell @hcho3 Please help taking a look.

trivialfis:
* A training parameter is passed down from language interface to dart prediction functions.
* Enabled it for R/Python, not Jvm.

trivialfis:
Ideally I should use predictor for non-training prediction, but right now the cache is not initialized and requires some refactoring to enable it.  Let's make it a to-do item.

trivialfis:
I will open a new PR for directly scaling tree leaf's weight.

trivialfis:
@Kodiologist Thanks for opening the PR!

PR #5119, opened 2019-12-15
Fix visual studio output library directories

RAMitchell:
Using the RELWITHDEBINFO build on visual studio results in the library being put on the incorrect location.


hcho3:
@RAMitchell Do you personally use RELWITHDEBINFO and MINSIZEREL builds? What are they used for?

RAMitchell:
I was using release with debug info for profiling. It optimises the code but you can run a profiler and still trace through all the functions/variables.

PR #5120, opened 2019-12-15
Tests and documents for new JSON routines.

trivialfis:


trivialfis:
For model schema, see https://github.com/hcho3/xgboost-json-schema/pull/11 .

PR #5129, opened 2019-12-16
implementation of map ranking algorithm on gpu

sriramch:
- also effected necessary suggestions mentioned in the earlier ranking pr's
  - made some performance improvements to the ndcg algo as well
  - i'll work on documenting the perf. numbers next


sriramch:
# performance numbers
## test environment
   * 1 socket
   * 6 cores/socket
   * 2 threads/core
   * 80 gb system memory
   * v100 gpu

## test
   * tree method is gpu_hist
   * uses all cpu threads
   * 100 trees
   * objective is `rank:map`

## small [mslr](https://www.microsoft.com/en-us/research/project/mslr/) dataset
   * ~ 3.6m training instances
   * distributed over ~ 10k groups
   * consuming 3.9 gb disk space
   * master does the gradient computation on cpu
   * fork does the gradient computation on gpu - i.e. this pr
   * gradient computation time and (inferred additional) memory overhead below

  |   |   | master |   |   |   | fork |  
:--: | :--: | :--: | :--: | :--: | :--: | :--: | :--: | :--:
metric |   | value | time (s) | memory (MiB) |   | value | time (s) | memory (MiB)
  |   |   |   |   |   |   |   |  
map |   | 0.648909 | 7 | 2177 |   | 0.648607 | 0.57 | 2191
auc |   | 0.992987 | 6.96 | " |   | 0.996506 | 0.57 | "
rmse |   | 1.05868 | 6.99 | " |   | 1.053611 | 0.57 | "
ndcg |   | 0.738857 | 6.99 | " |   | 0.739546 | 0.57 | "

## large [mslr](https://www.microsoft.com/en-us/research/project/mslr/) dataset
   * ~ 11.3m training instances
   * distributed over ~ 95k groups
   * consuming 13 gb disk space
   * master does the gradient computation on cpu
   * fork does the gradient computation on gpu - i.e. this pr
   * gradient computation time and (inferred additional) memory overhead below

  |   |   | master |   |   |   | fork |  
:--: | :--: | :--: | :--: | :--: | :--: | :--: | :--: | :--:
metric |   | value | time (s) | memory (MiB) |   | value | time (s) | memory (MiB)
  |   |   |   |   |   |   |   |  
map |   | 0.656295 | 23.12 | 4371|   | 0.656354 | 1.77 | 4414
auc |   | 1.026710 | 23.42 | " |   | 1.022204 | 1.76 | "
rmse |   | 0.994965 | 22.60 | " |   | 0.992704 | 1.76 | "
ndcg |   | 0.709523 | 22.95 | " |   | 0.710803 | 1.78 | "

sriramch:
upstream merge of #4805 into this fork is failing builds. can someone please look?

sriramch:

trivialfis:
Will look into this at weekend.  Thanks!

src/common/device_helpers.cuh
```diff
@@ -142,31 +142,91 @@ DEV_INLINE void AtomicOrByte(unsigned int* __restrict__ buffer, size_t ibyte, un
...
+ * \param v value for which to find the upper bound
+ * \param comp determines how the items are sorted ascending/descending order - should conform
+ *        to ordering semantics
+ * \return the smallest index i that has a value >= v, or n if none is larger
+ *         when sorted ascendingly
+ *         or, an index i with a value <= v, or 0 if none is smaller when sorted descendingly
+*/
+// Preserve existing default behavior of upper bound
+template <typename T, typename Comp = thrust::less<T>>
+XGBOOST_DEVICE __forceinline__ uint32_t LowerBound(const T *__restrict__ items,
```
RAMitchell: It's a bit unusual to use a comparison operator in this way to define the sort direction of the input. For this kind of code I'm inclined to stay as close to the STL/thrust as possible so it's immediately obvious whats happening.

A more 'STL-like' way to achieve this could be to pass a reverse iterator to the standard lower bound function.

src/objective/rank_obj.cu
```diff
@@ -465,105 +525,100 @@ struct NDCGLambdaWeightComputer {
...
+   * \param p_map_stats a vector containing the accumulated precisions for each position in a list
+   * \param map_stats_size size of the accumulated precisions vector
    */
-  inline static bst_float GetLambdaMAP(const std::vector<ListEntry> &sorted_list,
-                                       int index1, int index2,
-                                       std::vector<MAPStats> *p_map_stats) {
-    std::vector<MAPStats> &map_stats = *p_map_stats;
-    if (index1 == index2 || map_stats[map_stats.size() - 1].hits == 0) {
+  XGBOOST_DEVICE inline static bst_float GetLambdaMAP(
+    int ppred_idx, int npred_idx,
```
RAMitchell: According to google style variable names should not be abbreviated, except for common conventions (idx, num is fine). I don't know what ppred/npred means. Longer variable names are ok.

src/objective/rank_obj.cu
```diff
@@ -601,26 +651,127 @@ struct MAPLambdaWeightComputer {
...
+    };  // NOLINT
+
+    thrust::transform(thrust::make_counting_iterator(static_cast<uint32_t>(0)),
+                      thrust::make_counting_iterator(nitems),
+                      this->dmap_stats_.begin(),
+                      ComputeItemPrecisionLambda);
+
+    // Lastly, compute the accumulated precisions for all the items segmented by groups.
+    // The precisions are accumulated within each group
+    thrust::inclusive_scan_by_key(group_segments.begin(), group_segments.end(),
```
RAMitchell: `thrust::inclusive_scan_by_key` allocates memory, so benefits from using a temporary allocator.

src/objective/rank_obj.cu
```diff
@@ -356,52 +427,40 @@ struct NDCGLambdaWeightComputer {
...
-  inline const dh::caching_device_vector<uint32_t> &GetSortedPredPos() const {
-    return dindexable_sorted_preds_pos_;
+    : IndexablePredictionSorter(dpreds, segment_label_sorter),
+      dgroup_dcg_(segment_label_sorter.GetNumGroups(), 0.0f),
+      weight_multiplier_(segment_label_sorter, *this) {
+    const auto &group_segments = segment_label_sorter.GetGroupSegments();
+
+    // Compute each elements DCG values and reduce them across groups concurrently.
+    auto end_range =
+      thrust::reduce_by_key(group_segments.begin(), group_segments.end(),
```
RAMitchell: `thrust::reduce_by_key` is notoriously slow, it has to dynamically work out how many keys exist then allocate memory accordingly. If you know the number of keys in advance [cub::DeviceSegmentedReduce](https://nvlabs.github.io/cub/structcub_1_1_device_segmented_reduce.html) has much better performance.

If this section of code is not performance critical then using thrust is fine.

src/common/device_helpers.cuh
RAMitchell: Nevermind I just saw that they have a comparison operator in thrust :)

src/objective/rank_obj.cu
sriramch: i instrumented the code to compute the overheads and here is what i found:

- for the small mslr dataset, the fusing of transform and reduction (which is done here) took 18.2ms for 100 rounds, which is roughly ~ 182 microseconds per tree building round. out of this, half the time is spent on the transform operation - which is computing the item's dcg weight. hence, the overhead of reduction is ~ 90 microseconds per call

- for the large dataset it is ~ 400 microseconds for the transform and reduction of which reduction took ~ half the time

src/objective/rank_obj.cu
sriramch: done

src/objective/rank_obj.cu
sriramch: done, although the performance barely budged. there were ~ 200 more allocation this entailed while performing this.

src/common/device_helpers.cuh
sriramch: i modeled this api to be similar to the std/thrust's lower and upper bound, which also takes in a comparator. the only different is that this api returns the position as opposed to an iterator (as that is how it was used before in xgboost), where an item could be slotted without violating the order of the items determined by the comparator (which determines which one comes before the other).

sriramch:
> Overall looks good to me.
> 
> Is this comment still valid:
> 
> https://github.com/dmlc/xgboost/blob/ccaa819a7a01af5c9dc434060d8434c0f73d4e49/src/objective/rank_obj.cu#L1056
> 
> ?

fixed the debug message.

src/objective/rank_obj.cu
RAMitchell: So you are saying the performance impact is negligible, is that correct?

src/objective/rank_obj.cu
RAMitchell: Ok, I still want to make it a habit to use temporary allocators consistently with thrust.

Issue #5100, opened 2019-12-08
Extending XGBClassifier with **kwargs gives error about missing parameters


mglowacki100:
Hi,
I've tried to extend XGBClassifier with code as below. 
Basically, it adds handling feature names when data is pandas dataframe and fix issues with child class signature. There is a problem with `__init__`, mix-ins are not important. 
Analogous code for e.g. LogisticRegression from scikit works fine.
```
import copy
import pandas as pd
import xgboost as xgb


#mix-ins: 
class FeatureNames(object):
    def get_feature_names(self):
        if not isinstance(self.feature_names, list):
            raise ValueError('Must fit data first!')
        else:
            return self.feature_names


class FeatureImportances(object):
    def get_feature_importances(self):
        return dict(zip(self.get_feature_names(), self.feature_importances_))

        
class XGBClassifierN(FeatureImportances, FeatureNames, xgb.XGBClassifier):
    def __init__(self, **kwargs): 
        super().__init__(**kwargs)
        self.feature_names = None
   
    #hack from https://stackoverflow.com/questions/51430484/how-to-subclass-a-vectorizer-in-scikit-learn-without-repeating-all-parameters-in
    def get_params(self, deep=True):
        params = super().get_params(deep)
        cp = copy.copy(self)
        cp.__class__ = xgb.XGBClassifier
        params.update(cp.__class__.get_params(cp, deep))
        return params

    def fit(self, X, y=None):
        self.feature_names = list(X.columns)
        return super().fit(X, y)


XGBClassifierN()
print(XGBClassifierN())
```
Error message:
```
runfile('/home/mglowacki/Desktop/xg_example.py', wdir='/home/mglowacki/Desktop')
Traceback (most recent call last):

  File "<ipython-input-1-693451a0d6aa>", line 1, in <module>
    runfile('/home/mglowacki/Desktop/xg_example.py', wdir='/home/mglowacki/Desktop')

  File "/home/mglowacki/anaconda3/envs/py37/lib/python3.7/site-packages/spyder_kernels/customize/spydercustomize.py", line 827, in runfile
    execfile(filename, namespace)

  File "/home/mglowacki/anaconda3/envs/py37/lib/python3.7/site-packages/spyder_kernels/customize/spydercustomize.py", line 110, in execfile
    exec(compile(f.read(), filename, 'exec'), namespace)

  File "/home/mglowacki/Desktop/xg_example.py", line 48, in <module>
    print(XGBClassifierN())

[...]

Issue #4725, opened 2019-07-31
XGBoost version not compiled with GPU support, even though it is


leedrake5:
It has been a long road to installation, but here's what works for Ubuntu 18.04 and R 3.6.1:

#Cuda 10.0 + NCCL 2.4.7
###install OS agnostic NCCL and put it in /usr/local/nccl-2.4.7
#remove all symlinks to usr/local/cuda and usr/bin/nvcc
```
cd /home/<user>/Software
sudo rm -R xgboost

git clone --recursive https://github.com/dmlc/xgboost
cd xgboost
mkdir build
cd build

cmake -DUSE_CUDA=ON -DR_LIB=ON -DR_LIB=ON -DUSE_NCCL=ON -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda-10.0 -DNCCL_ROOT=/usr/local/nccl-2.4.7 ..
sudo make -j4
```

This installs successfully. After this, I compile the R package

```
cd /home/<user>/Software/xgboost
sudo R CMD INSTALL R-package

```

This all works, and to my knowledge I've done everything right to enable GPU support. But when I run test data:

```
library(xgboost)
library(rBayesianOptimization)
data(agaricus.train, package = "xgboost")
 dtrain <- xgb.DMatrix(agaricus.train$data,
                       label = agaricus.train$label)
 cv_folds <- KFold(agaricus.train$label, nfolds = 5,
                   stratified = TRUE, seed = 0)
 xgb_cv_bayes <- function(max_depth, min_child_weight, subsample) {
   cv <- xgb.cv(params = list(booster = "gbtree", eta = 0.01,
                              max_depth = max_depth,
                              min_child_weight = min_child_weight,
                              subsample = subsample, colsample_bytree = 0.3,
                              lambda = 1, alpha = 0,
                              objective = "reg:linear",
                              eval_metric = "rmse"),
                data = dtrain, nround = 10,
                folds = cv_folds, prediction = TRUE, showsd = TRUE,
                early_stopping_rounds = 5, tree_method = 'gpu_hist', n_gpus=-1,maximize = TRUE, verbose = 0)
   list(Score = cv$evaluation_log$test_rmse_mean[cv$best_iteration],
        Pred = cv$pred)
 }
 OPT_Res <- BayesianOptimization(xgb_cv_bayes,
                                 bounds = list(max_depth = c(2L, 6L),
                                               min_child_weight = c(1L, 10L),
[...]

Issue #5099, opened 2019-12-08
[Documentation problem] XGBoostClassifier doesn't specify if monotone_constraints are usable


alexbaron10:
https://xgboost.readthedocs.io/en/latest/python/python_api.html#module-xgboost.sklearn

Doesn't specify if monotone constraints are usable (it seems like they are but i'm not entirely sure since it doesn't explicitly specify.

Thanks in advance

EDIT: neither does it appear here:

https://github.com/dmlc/xgboost/blob/master/doc/parameter.rst

trivialfis:
Em.  Let me check all parameters for SKL interface.  It should work as they are all passed into `xgb.train` as part of `**kwargs`.  I wish we can generate these parameters based on c++ code.

alexbaron10:
> Em. Let me check all parameters for SKL interface. It should work as they are all passed into `xgb.train` as part of `**kwargs`. I wish we can generate these parameters based on c++ code.

Thanks for the effort. I've been trying to use them with DART but i can't see if they are working or not either (documentation doesn't specify if they apply when using DART), that's another problem. Maybe a checkable way could be added in the future that tells if monotone_constrains are being used?

trivialfis:
Depends on https://github.com/dmlc/xgboost/pull/5130 .

PR #5130, opened 2019-12-17
[Breaking] Remove Scikit-Learn default parameters

mglowacki100:
PR started as: https://github.com/dmlc/xgboost/issues/5100
Summary (edited):
- main point is refactoring of `get_params` in `XGBModel` thats allow less repetition in `__init__` of children classes like XGBClassifier (MixIn was not required, it was based on https://stackoverflow.com/questions/59248211) and now is more clear which parameters defaults values are directly inherited
- I've deleted in `get_params` in `XGBModel` two lines that were giving me error in my issue:
` if params['missing'] is np.nan:`      
`     params['missing'] = None  # sklearn doesn't handle nan. see #4725)`
tests are passing, so either there is no specific test-case, or everything is correct ;-)
- in original default for `n_jobs` in `XGBRanker` is `-1` (accidental inconsistency?)
- `XGBRegressor` now has its own `__init__`
- I moved up `gpu_id` to the `XGBModel`, so less repetition and I think it makes sense
- I was thinking about setting default `objective` in `XGBModel` to `None`, to make `XGBModel` more abstract estimator in similar fashion scikit-learn, but it gives error at cpp level about unrecognized type of objective, so I've decided to keep it as it it is

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -487,13 +487,6 @@ class XGBModel(XGBModelBase):
           If you want to run prediction using multiple thread, call ``xgb.copy()`` to make copies
           of model object and then call ``predict()``.
 
-        .. note:: Using ``predict()`` with DART booster
-
-          If the booster object is DART type, ``predict()`` will perform dropouts, i.e. only
-          some of the trees will be evaluated. This will produce incorrect results if ``data`` is
-          not the training data. To obtain correct results on test sets, set ``ntree_limit`` to
-          a nonzero value, e.g.
-
           .. code-block:: python
 
             preds = bst.predict(dtest, ntree_limit=num_round)
@@ -607,7 +600,8 @@ class XGBModel(XGBModelBase):
         feature_importances_ : array of shape ``[n_features]``
 
         """
-        if getattr(self, 'booster', None) is not None and self.booster not in {'gbtree', 'dart'}:
+        if getattr(self, 'booster', None) is not None and self.booster not in {
+                'gbtree', 'dart'}:
             raise AttributeError('Feature importance is not defined for Booster type {}'
                                  .format(self.booster))
         b = self.get_booster()
@@ -774,14 +768,6 @@ class XGBClassifier(XGBModel, XGBClassifierBase):
           If you want to run prediction using multiple thread, call
           ``xgb.copy()`` to make copies of model object and then call
           ``predict()``.
-
-        .. note:: Using ``predict()`` with DART booster
-
-          If the booster object is DART type, ``predict()`` will perform
-          dropouts, i.e. only some of the trees will be evaluated. This will
-          produce incorrect results if ``data`` is not the training data. To
-          obtain correct results on test sets, set ``ntree_limit`` to a nonzero
-          value, e.g.
 
           .. code-block:: python
 
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -1,14 +1,15 @@
 # coding: utf-8
 # pylint: disable=too-many-arguments, too-many-locals, invalid-name, fixme, E0012, R0912, C0302
 """Scikit-Learn Wrapper interface for XGBoost."""
+import copy
 import warnings
 import json
 import numpy as np
 from .core import Booster, DMatrix, XGBoostError
 from .training import train
 
-# Do not use class names on scikit-learn directly.
-# Re-define the classes on .compat to guarantee the behavior without scikit-learn
+# Do not use class names on scikit-learn directly.  Re-define the classes on
+# .compat to guarantee the behavior without scikit-learn
 from .compat import (SKLEARN_INSTALLED, XGBModelBase,
                      XGBClassifierBase, XGBRegressorBase, XGBLabelEncoder)
 
@@ -48,18 +49,17 @@ def _objective_decorator(func):
     return inner
 
 
-class XGBModel(XGBModelBase):
-    # pylint: disable=too-many-arguments, too-many-instance-attributes, invalid-name
-    """Implementation of the Scikit-Learn API for XGBoost.
+__estimator_doc = '''
+    n_estimators : int
+        Number of gradient boosted trees.  Equivalent to number of boosting
+        rounds.
+'''
 
-    Parameters
-    ----------
+__model_doc = '''
     max_depth : int
         Maximum tree depth for base learners.
     learning_rate : float
         Boosting learning rate (xgb's "eta")
-    n_estimators : int
-        Number of trees to fit.
     verbosity : int
         The degree of verbosity. Valid values are 0 (silent) - 3 (debug).
     objective : string or callable
@@ -75,7 +75,8 @@ class XGBModel(XGBModelBase):
     n_jobs : int
         Number of parallel threads used to run xgboost.
     gamma : float
-        Minimum loss reduction required to make a further partition on a leaf node of the tree.
+        Minimum loss reduction required to make a further partition on a leaf
+        node of the tree.
     min_child_weight : int
         Minimum sum of instance weight(hessian) needed in a child.
     max_delta_step : int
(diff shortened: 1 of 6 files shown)


trivialfis:
Please rebase your branch to latest master first.

mglowacki100:
I've followed: https://xgboost.readthedocs.io/en/latest/contrib/git_guide.html and get `Everything up-to-date` after `git push --force` so I hope now is correct.

hcho3:
To fix broken builds, merge the latest master.

trivialfis:
> To fix broken builds, merge the latest master.

Do it again. ;-)  The link for maven was broken.

mglowacki100:
@trivialfis 
Tests are passing. I've updated first comment with changes summary.

python-package/xgboost/sklearn.py
```diff
@@ -143,8 +144,8 @@ class XGBModel(XGBModelBase):
 
     def __init__(self, max_depth=3, learning_rate=0.1, n_estimators=100,
                  verbosity=1, objective="reg:squarederror",
-                 booster='gbtree', tree_method='auto', n_jobs=1, gamma=0,
-                 min_child_weight=1, max_delta_step=0, subsample=1,
+                 booster='gbtree', tree_method='auto', n_jobs=1, gpu_id=-1,
```
trivialfis: Thanks for putting `gpu_id` here.

python-package/xgboost/sklearn.py
```diff
@@ -598,28 +604,8 @@ class XGBClassifier(XGBModel, XGBClassifierBase):
...
-            min_child_weight=min_child_weight,
-            max_delta_step=max_delta_step, subsample=subsample,
-            colsample_bytree=colsample_bytree,
-            colsample_bylevel=colsample_bylevel,
-            colsample_bynode=colsample_bynode,
-            reg_alpha=reg_alpha, reg_lambda=reg_lambda,
-            scale_pos_weight=scale_pos_weight,
-            base_score=base_score, random_state=random_state, missing=missing,
-            **kwargs)
+    def __init__(self, objective="binary:logistic", **kwargs):
```
trivialfis: I looked into the source code of inspection, not sure if this might break sklearn grid search.  Will help adding some tests here.

mglowacki100: `test_kwargs_grid_search` for `XGBClassifier` and `test_parameter_tuning` for `XGBRegressor` uses `GridSearchCV`. Please let me know if you have more specific case in my mind. 
Btw. just an idea for separate task if not already done: compare XGBoost test-cases with LightGBM for Sklearn - a lot of overlap, but maybe there are interesting cases that  are missing.

trivialfis: Definitely.  I'm seeing more and more users preferring sklearn interface over more primitive train and predict.  Will try to make it my default interface for development in the future.

python-package/xgboost/sklearn.py
```diff
@@ -881,32 +853,20 @@ class XGBRegressor(XGBModel, XGBRegressorBase):
...
-            colsample_bytree=colsample_bytree,
-            colsample_bylevel=colsample_bylevel,
-            colsample_bynode=colsample_bynode,
-            reg_alpha=reg_alpha, reg_lambda=reg_lambda,
-            scale_pos_weight=scale_pos_weight,
-            base_score=base_score, random_state=random_state, missing=missing,
-            **kwargs)
+    def __init__(self, learning_rate=1, subsample=0.8, colsample_bynode=0.8,
+                 reg_lambda=1e-5, **kwargs):
+        super().__init__(learning_rate=learning_rate, subsample=subsample,
```
trivialfis: Some parameters doesn't make sense for RF wrapper.  For example, the `n_estimators` means number of trees in random forest, while for other regressor `n_estimators` means number of boosting rounds.

python-package/xgboost/sklearn.py
```diff
@@ -881,32 +853,20 @@ class XGBRegressor(XGBModel, XGBRegressorBase):
...
+                         colsample_bynode=colsample_bynode,
```
trivialfis: While other parameters are largely the same.

python-package/xgboost/sklearn.py
```diff
@@ -221,13 +223,17 @@ def set_params(self, **params):
...
         return self
 
-    def get_params(self, deep=False):
+    def get_params(self, deep=True):
+        # pylint: disable=attribute-defined-outside-init
         """Get parameters."""
-        params = super(XGBModel, self).get_params(deep=deep)
+        #Based on: https://stackoverflow.com/questions/59248211
+        params = super().get_params(deep)
+        cp = copy.copy(self)
```
trivialfis: Copying involves a serialization and unserialization of model, performance is one thing, another drawback is all caches of `DMatrix` are removed during copy.  Will look into the technique further.

python-package/xgboost/sklearn.py
mglowacki100: I'm not sure, but I think this behavior also occurs in original code. Maybe it is better to remove `n_estimators` from `XGBModel` and add it only to classifiers, but for regressors create separate parameter `n_boosting_rounds`?

python-package/xgboost/sklearn.py
mglowacki100: I was thinking that `DMatrix` is involved only during `fit`, so `__init__` is light-weight but for sure it need to be checked.

trivialfis:
Note to myself: I will push some more commits to this branch, make sure the author at final squashing doesn't change (if we merge this PR).

trivialfis:
Tried printing out the procedure, seems `get_params` is called multiple times internally.

mglowacki100:
Yes, `get_params` is called recursively to gather parameters from hiearchy class and to create full signature. Also with this trick you can't use abstract classes. 
There is some work to deal with it on `scikit` level: https://github.com/scikit-learn/scikit-learn/issues/13555 , I've not digested it fully yet but it seems that is based on the same trick. I plan to look at `CatBoost`  and check libraries like `autoclass`.

trivialfis:
Thanks for the reference.  Really is the dark corner of Python.  Let me keep digging into it.  Your PR is really important (at least to me),  I just want to make sure it doesn't break anything and I can document the technique for future reference.

mglowacki100:
If there will be some problems with this, there is probably an alternative approach with boiler-plate code generation by engine template (e.g. ninja2, https://github.com/minimaxir/automl-gs).
Btw. with this PR extensions for handling feature names from pandas DataFrame are quite compact:
``` 
class XGBClassifierPD(FeatureImportances, FeatureNames, xgb.XGBClassifier):
    """
    XGBoost Classifier for Pandas DataFrame
    """
    def __init__(self, **kwargs): 
        super().__init__(**kwargs)
        self.feature_names = None
   
    def fit(self, X, y=None):
        self.feature_names = list(X.columns)
        return super().fit(X, y)


class FeatureNames():
    def get_feature_names(self):
        if not isinstance(self.feature_names, list):
            raise ValueError('Must fit data first!')
        else:
            return self.feature_names

class FeatureImportances():
    def get_feature_importances(self):
        return dict(zip(self.get_feature_names(), self.feature_importances_))
```
There is ongoing work to handle feature names within `sklearn` , but for now this is my stopgap solution.

trivialfis:
@RAMitchell @hcho3  Please help taking a look.

trivialfis:
Breaking in the sense that default values are now consistent with c++ and hence basic `train` method, but not the ones originally presented in skl wrappers.  I reordered some parameters in the base class to inspect the changes, which can be reverted back, but not sure how useful would that be as normally one would use keyword arguments ...

trivialfis:
@RAMitchell kwargs doesn't work with scikit-learn interface.  It inspects the parameter to `__init__` function, kwargs will break the inspect procedure.

trivialfis:
@mglowacki100 The interesting thing happened, @RAMitchell did a commit: https://github.com/dmlc/xgboost/commit/5d6baed998b3a3287006148670d10ec04565f5f1#diff-2c197a11c1b576e821f5942be9eab74c that can use kwargs with grid search.  The hack might not be necessary now.

trivialfis:
`max_depth`: 3 -> 6.
`eta`: 0.1 -> 0.3

Might be easier to overfit.

trivialfis:
The reason this doesn't trigger a serialization is because skl wrappers don't have `__copy__` method.  There needs to have some way to make sure we won't run into troubles by calling serialization repeatedly in the future.

trivialfis:
The basic flow in `get_params` is:
- Return parameters in subclass first, by calling inspect.
- Return parameters in `XGBModel` (the base class).
- Return whatever in `**kwargs`.
- Merge them.

trivialfis:
@hcho3

trivialfis:
Reminder to myself:  Add doc for https://github.com/dmlc/xgboost/issues/5099 .

trivialfis:
@hcho3 Friendly ping.  :-)

python-package/xgboost/sklearn.py
```diff
@@ -1,6 +1,7 @@
-# coding: utf-8
+# coding-8
```
hcho3: ??? We should keep `coding: utf-8`

python-package/xgboost/sklearn.py
```diff
@@ -717,15 +771,17 @@ def predict(self, data, output_margin=False, ntree_limit=None,
...
-          some of the trees will be evaluated. This will produce incorrect results if ``data`` is
-          not the training data. To obtain correct results on test sets, set ``ntree_limit`` to
-          a nonzero value, e.g.
+          If the booster object is DART type, ``predict()`` will perform
+          dropouts, i.e. only some of the trees will be evaluated. This will
+          produce incorrect results if ``data`` is not the training data. To
+          obtain correct results on test sets, set ``ntree_limit`` to a nonzero
+          value, e.g.
 
           .. code-block:: python
```
hcho3: Side comment: this note is no longer necessary, since `predict()` will no longer drop trees by default. See #5115 and #5204. Also need to remove the same note from `XGBModel.predict()` too.

python-package/xgboost/sklearn.py
trivialfis: Good cache.

trivialfis:
@mglowacki100 Thank you so much for introducing the work around.  It's been a huge headache for those duplicated parameters.

PR #5137, opened 2019-12-19
Example JSON model parser and Schema.

trivialfis:
* Add schema.
* Add an example model parser.
* Use `jsonschema` package to validate the schema and generated model.


trivialfis:
@hcho3 @KOLANICH I merged the schema to here.  The display is not so nice but should suffice for now.  We can move all discussion to main XGBoost repo for future improvement.

trivialfis:
Note to myself: add a ref to demo.

demo/json-model/json_parser.py
```diff
@@ -0,0 +1,180 @@
...
+            nid = stacks.pop()
+
+            node['node id'] = nid
+            node['gain'] = self.loss_change(nid)
+            node['cover'] = self.sum_hessian(nid)
+            nodes.append(node)
+
+            if not self.is_leaf(nid) and not self.is_deleted(nid):
+                left = self.left_child(nid)
+                rigth = self.right_child(nid)
```
hcho3:
```suggestion
                right = self.right_child(nid)
```

demo/json-model/json_parser.py
```diff
@@ -0,0 +1,180 @@
...
+                stacks.append(left)
+                stacks.append(rigth)
```
hcho3:
```suggestion
                stacks.append(right)
```

tests/python/test_basic_models.py
```diff
@@ -11,6 +13,20 @@
 rng = np.random.RandomState(1994)
 
 
+def json_model():
```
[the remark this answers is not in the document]
trivialfis: @hcho3 Done.

Issue #5104, opened 2019-12-09
CPU optimizations - 'hist' method


SmirnovEgorRu:
Moved discussion from [PR #5008](https://github.com/dmlc/xgboost/pull/5008) to the issue. After the PR - performance drop on CPU is observed. I prepared a plan how to establish optimizations from [PR #4529](https://github.com/dmlc/xgboost/pull/4529) in master branch again.

I tried to split large changes to separate PRs:

|№|Description|Status|
|--|--|--|
|1| Extend Performance Timer to track overall performance of each algorithmic stage for whole training|[**Merged**](https://github.com/dmlc/xgboost/pull/5107)|
|2| Add benchmarks to track performance changes (~3 data sets are enough for the start, it can be extended later)|[**Merged**](https://github.com/dmlc/xgboost-bench/commit/c787a590e0ece4898a65c0026ad9dce21e201876)|
|3| Add nested parallelism in EvaluateSplit function + tests |[**Merged**](https://github.com/dmlc/xgboost/pull/5138)|
|4| Add nested parallelism in BuildHist function + tests |[**Merged**](https://github.com/dmlc/xgboost/pull/5156)|
|5| Add nested parallelism in ApplySplit function + tests |[**Merged**](https://github.com/dmlc/xgboost/pull/5244)|

Expected timelines for creation the all PRs - EOY.

trivialfis:
I would love to see the performance optimization back and sorry for the oversight.  Feel free to ping me if you need any help.

hcho3:
> Add benchmarks to track performance changes (~3 data sets are enough for the start, it can be extended later)

For now, a script that we can run is okay, so that performance results can be re-produced. I plan to design and build an automatic system for performance benchmarking.

SmirnovEgorRu:
All PRs are merged, final results:

Dataset | higgs1m | Letters | Airline-ohe | MNIST | MSRank-30K | Mortgage
-- | -- | -- | -- | -- | -- | --
Before commit https://github.com/dmlc/xgboost/pull/5008, sec | 15.5 | 10.3 | 55.3 | 69.5 | 99.1 | 18.3
After commit https://github.com/dmlc/xgboost/pull/5008, sec | 123.5 | 144.6 | 133.2 | Inf | 332.2 | 53.4 
Current Master, sec | 14.7 | 10.1 | 59.4 | 80.5 | 111.6 | 19.0

[...]

PR #5138, opened 2019-12-19
Optimization of EvaluateSplit function 

SmirnovEgorRu:
EvaluateSplit was optimized by processing many tree nodes in the same time.
Foe details look at the issue https://github.com/dmlc/xgboost/issues/5104

To implement this ParallelFor2d was introduced to support nested parallelism. This function will be used for optimizations of other functions BuildHist and ApplySplit.
In previous implementation the main source of code complexity was blocking for nested parallelism. Now common code for this was aggregated and moved in common threading utils. So, further PRs should be more light-weight.

I will provide performance numbers a bit later.

CC: @hcho3, @RAMitchell, @trivialfis

### DIFF
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -28,6 +28,8 @@
 #include "../common/hist_util.h"
 #include "../common/row_set.h"
 #include "../common/column_matrix.h"
+#include "../common/threading_utils.h"
+
 
 namespace xgboost {
 namespace tree {
@@ -188,9 +190,11 @@ void QuantileHistMaker::Builder::EvaluateSplits(
     int depth,
     unsigned *timestamp,
     std::vector<ExpandEntry> *temp_qexpand_depth) {
+  this->EvaluateSplit(qexpand_depth_wise_, gmat, hist_, *p_fmat, *p_tree);
+
   for (auto const& entry : qexpand_depth_wise_) {
     int nid = entry.nid;
-    this->EvaluateSplit(nid, gmat, hist_, *p_fmat, *p_tree);
+
     if (snode_[nid].best.loss_chg < kRtEps ||
         (param_.max_depth > 0 && depth == param_.max_depth) ||
         (param_.max_leaves > 0 && (*num_leaves) == param_.max_leaves)) {
@@ -220,7 +224,8 @@ void QuantileHistMaker::Builder::ExpandWithDepthWise(
   int num_leaves = 0;
 
   // in depth_wise growing, we feed loss_chg with 0.0 since it is not used anyway
-  qexpand_depth_wise_.emplace_back(ExpandEntry(0, p_tree->GetDepth(0), 0.0, timestamp++));
+  qexpand_depth_wise_.emplace_back(ExpandEntry(ExpandEntry::kRootNid,
+      p_tree->GetDepth(ExpandEntry::kRootNid), 0.0, timestamp++));
   ++num_leaves;
   for (int depth = 0; depth < param_.max_depth + 1; depth++) {
     int starting_index = std::numeric_limits<int>::max();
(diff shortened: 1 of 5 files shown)


RAMitchell:
This looks much better than before, also PR is about the right size. I will leave others to comment on algorithm correctness.

SmirnovEgorRu:
Looks like errors in CI contain are not related to my changes:
[2019-12-19T21:54:43.782Z] Connecting to apache.osuosl.org (apache.osuosl.org)|64.50.236.52|:80... connected.
[2019-12-19T21:54:44.174Z] HTTP request sent, awaiting response... 404 Not Found
[2019-12-19T21:54:44.174Z] 2019-12-19 21:54:43 ERROR 404: Not Found.

trivialfis:
Please rebase it to master branch or merge the latest commits.

SmirnovEgorRu:
@hcho3, here is a breakdown for Airline data set, now performance for EvaluateSplit is similar to how it was before reverting:

airline-ohe | ApplySplit | EvaluateSplit | BuildHist | SyncHistogram | Prediction | Total, sec
-- | -- | -- | -- | -- | -- | --
Master | 26 | **27** | 67 | 12 | 2 | 157
Before   reverting | 9.0 | **6.1** | 28.8 | 0.0 | 0.7 | 63.7
This PR | 25.8 | **5.4** | 67.3 | 11.6 | 1.5 | 135.3

src/tree/updater_quantile_hist.cc
```diff
@@ -554,40 +562,77 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
   builder_monitor_.Stop("InitData");
 }
 
-void QuantileHistMaker::Builder::EvaluateSplit(const int nid,
+// if sum of statistics for non-missing values in the node
```
trivialfis: @hcho3 I thought missing value is removed during construction of DMatrix?

src/tree/updater_quantile_hist.cc
```diff
@@ -259,9 +263,11 @@ void QuantileHistMaker::Builder::ExpandWithLossGuide(
 
   this->InitNewNode(0, gmat, gpair_h, *p_fmat, *p_tree);
 
-  this->EvaluateSplit(0, gmat, hist_, *p_fmat, *p_tree);
-  qexpand_loss_guided_->push(ExpandEntry(0, p_tree->GetDepth(0),
-                                         snode_[0].best.loss_chg, timestamp++));
+  ExpandEntry node(0, p_tree->GetDepth(0), snode_[0].best.loss_chg, timestamp++);
```
trivialfis: The depth of root node is determined I believe.

src/tree/updater_quantile_hist.cc
```diff
@@ -259,9 +263,11 @@ void QuantileHistMaker::Builder::ExpandWithLossGuide(
...
+  this->EvaluateSplit({node}, gmat, hist_, *p_fmat, *p_tree);
+  node.loss_chg = snode_[0].best.loss_chg;
```
trivialfis: Why do you need to set it here?  Is it set before this PR?

src/tree/updater_quantile_hist.cc
hcho3: @trivialfis DMatrix does not store missing values physically, but the split-finding algorithm still needs to take account of missing values.

Explanation for this piece of function: Let `NM` represent the set of all data points whose (particular) feature `f` is not missing, and let `S` represent the set of data points associated with the particular node `snode`. Then we want to ask whether two sets `intersection(NM, S)` and `S` are identical or not. There are two possibilities:

1. `intersection(NM, S) == S`. That is, `S` is a subset of `NM` and all data points in `S` has a non-missing value for the particular feature `f`.
2. `intersection(NM, S) != S`. In this case, there exists one data point in `S` which has a missing value for feature `f`.

trivialfis: @hcho3 Thanks for the detailed explanation.  I forgot this optimization.

src/tree/updater_quantile_hist.cc
trivialfis: I see the passed in parameter is no longer a struct.

src/tree/updater_quantile_hist.cc
SmirnovEgorRu: Changed 0 to ExpandEntry::kRootNid const, if I correctly understand the comment.

src/tree/updater_quantile_hist.cc
```diff
@@ -554,40 +565,77 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
+    }
   }
-  GHistRow node_hist = hist[nid];
 
-#pragma omp parallel for schedule(dynamic) num_threads(nthread)
-  for (bst_omp_uint i = 0; i < nfeature; ++i) {  // NOLINT(*)
-    const auto feature_id = static_cast<bst_uint>(feature_set[i]);
+  // Create 2D space (# of nodes to process x # of features to process)
+  // to process them in parallel
+  common::BlockedSpace2d space(n_nodes_in_set, [&](size_t fid) {
```
```suggestion
  common::BlockedSpace2d space(n_nodes_in_set, [&](size_t nid_in_set) {
```
to make it clear that the functor is measuring the size of the feature set associated with a particular node ID.

src/tree/updater_quantile_hist.cc
```diff
@@ -554,40 +565,77 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
+      return features_sets[fid]->Size();
```
```suggestion
      return features_sets[nid_in_set]->Size();
```

src/tree/updater_quantile_hist.cc
```diff
@@ -831,8 +879,10 @@ void QuantileHistMaker::Builder::InitNewNode(int nid,
   builder_monitor_.Stop("InitNewNode");
 }
 
+
 // enumerate the split values of specific feature
```
```suggestion
// Enumerate the split values of specific feature.
// Returns the sum of gradients corresponding to the data points that contains a non-missing value
// for the particular feature fid.
```

trivialfis:
Will run some bench on my own today.  For other reviews I will leave it up to @hcho3

src/tree/updater_quantile_hist.h
```diff
@@ -230,7 +231,8 @@ class QuantileHistMaker: public TreeUpdater {
                      const RegTree& tree);
 
     // enumerate the split values of specific feature
```
hcho3:
```suggestion
    // Enumerate the split values of specific feature
    // Returns the sum of gradients corresponding to the data points that contains a non-missing
    // value for the particular feature fid.
```

hcho3:
@trivialfis Can we merge this now?

trivialfis:
Will try merging it today.

trivialfis:
Running benchmarks.

trivialfis:
@SmirnovEgorRu Thanks for the optimization!

PR #5144, opened 2019-12-21
Note for `DaskDMatrix`.

trivialfis:
I received many questions around why the `DaskDMatrix` is slow and buggy, while most of the time `DaskDMatrix` is not the one to be blamed.  Here's some short notes for what's happening inside `DaskDMatrix`.


trivialfis:
This is weird, `xgboost.dask.train` doesn't show up in final rendering of doc,

trivialfis:
@hcho3. Thanks!  It would have taken me lots of time to figure out this!

Issue #4190, opened 2019-02-27
Cannot use DMatrix's set_base_margin method in SciKit-Learn API


tmbluth:
Though this can be done in the xgboost.train framework, the SciKit-Learn API framework only allows X and y as inputs which are internally fed to DMatrix. However from my research looking through the code and online there is no way to access DMatrix methods from the SciKit-Learn wrapper, thereby rendering this functionality untouchable. The base margin does not seem to fit anywhere in XGBClassifier or XGBRegressor, though the user can boost off an entire model with the same inputs. 

Is there a way to add this functionality to allow SciKit-Learn fans to boost from predictions instead of boosting from an entire first model?

In my case, using the first model is not possible, but using predictions from that first model as a starting point is possible, since the second phase model uses a super-set of the first model's attributes.

trivialfis:
@tmbluth Seems you spent some time dig through the code,  do you have any suggestion how we can approach this?

tmbluth:
Unfortunately I am more of an applied data scientist that came from a research background and less of a software developer, so my expertise does not yet extend there. I'm also primarily an R user that is picking up Python, so I think there are others more qualified.

trivialfis:
> Unfortunately I am more of an applied data scientist that came from a research background

Cool.  Let me think about it later.

littlebeanfang:
come across the same trouble

zdwhite:
I believe I have the same issue.

Essentially I have to use the scikit learn wrapper to have access to the "predict_probas" method. This is frustrating as I lose all the functionality of the dmatrix when building the classifier using the scikitlearn wrapper.

I thought I could try and manipulate the dmatrix structure and use it as an input for the scikit learn wrapper however I can't seem to find a way of doing this.

trivialfis:
Maybe add an extra parameter in `fit`?

PR #5151, opened 2019-12-23
Add base margin to sklearn interface.

trivialfis:
Closes https://github.com/dmlc/xgboost/issues/4190 .


trivialfis:
Note that the sklearn interface is much more slower than primitive interface as DMatrix is constructed for every call of `fit` and `predict`.  Will address after next release.

Issue #5143, opened 2019-12-21
[Roadmap] DMatrix refactoring


RAMitchell:
This issue is a roadmap and checklist for ongoing work re-factoring DMatrix (see RFC #4354).

The first steps are to use a common interface to external data, unifying the way DMatrix objects are constructed and simplifying the process of adding new external data sources.

- [x] Create adapter interface to external data, implement constructors for SimpleDMatrix (#5044)
- [x] Extend data adapters to SparsePageDMatrix (#5092)
- [x] Extend data adapters to cudf data loading (#5108)
- [x] Add support for cupy (#5206)
- [x] Implement DMatrix slice via adapter (#5198)

After the above all DMatrix constructors will be happening via adapters, missing value handling and use of threads will be consistent.

Then I plan to start reducing the number of classes associated with DMatrix.
- [x] Remove SimpleCSRSource into SimpleDMatrix (#5315)
- [x] Remove SparsePageSource into SparsePageDMatrix

The final goal is to save memory by constructing histogram matrices for the `hist` and `gpu_hist` algorithms directly from external data using adapters. We will need some discussion on the interface e.g. if a user wants to build the histogram DMatrix directly, specify an enum to the constructor indicating  DMatrix type
- [x] Develop interface for instantiating histogram DMatrix directly
- [x] Build constructor for EllPack matrix directly using adapters

Lastly:
- [x] Enable weighted sketching for `DeviceQuantileDMatrix`.

dattatreyam:
Hi, I am trying to generate a DaskDMatrix object from large dataframes and it is super slow. Is it expected to be slow ? If so, can you please help me understand why this is the case.

dtrain = DaskDMatrix(client, large_dask_dataframe_x, large_dask_dataframe_y) 

my large_dask_dataframe has few million examples ...

trivialfis:
@dattatreyam Dask's computation is lazy, and `DaskDMatrix`'s constructor forces all lazy computation to materialize.  That means all your previous operations on dask's data structure like dataframe or array are carried out in the constructor of `DaskDMatrix`.

JohnZed:
@RAMitchell - the steps look good to me, but I see you allude to implementing DMatrix slice functionality. Is this like selecting a subset of rows? Or columns? Sounds promising but I'd be curious to hear more what you have in mind for it.

RAMitchell:
@JohnZed its functionality that already exists but needs to be ported to new code before all of the old code can be removed. It is a small job in this case.

[...]

PR #5153, opened 2019-12-23
Quick fix for memory leak in CPU Hist.

trivialfis:
Closes https://github.com/dmlc/xgboost/issues/3579 .

The proper fix is putting quantiles in `DMatrix`, which we have already done for GPU Hist.  See https://github.com/dmlc/xgboost/issues/5143 for related future roadmap.  This PR is a band-aid for fixing the old issue as we want a fully working 1.0 release.

### DIFF
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -53,8 +53,7 @@ void QuantileHistMaker::Configure(const Args& args) {
 void QuantileHistMaker::Update(HostDeviceVector<GradientPair> *gpair,
                                DMatrix *dmat,
                                const std::vector<RegTree *> &trees) {
-  if (is_gmat_initialized_.find(dmat) == is_gmat_initialized_.cend() ||
-      is_gmat_initialized_.at(dmat) == false) {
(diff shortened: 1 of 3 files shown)


trivialfis:
Looking back, this PR is only made possible by many previous refactorings.

trivialfis:
@hcho3 @CodingCat  Is there any particular bug for hist that's consider critical?  We @RAMitchell  would like to make `hist` as the default algorithm.  1.0 seems to be a right opportunity.

trivialfis:
Ping @hcho3

hcho3:
@trivialfis I’m okay with it. We do need to make further improvements, to make it faster and more memory efficient.

trivialfis:
@hcho3 Well, improvement is endless unless there's a better algorithm succeeds `hist` (like what `hist`) does to `exact`.  But it would be the right time to make `hist` as the default now.

Are you Okay with this PR?

hcho3:
@trivialfis I was referring to future improvements.

trivialfis:
@hcho3 Got it.  Will create a new PR for setting it as default.

Would you like to review this fix?

Issue #4211, opened 2019-03-05
[RFC] Training parameter validation


jphelps13:
Hey :)

**Issue**
We have recently found that we mistyped a parameter in our `xgb.train` implementation. Instead of `colsample_bytree` we used `colsample_by_tree`. 

We didn't receive any warnings or errors, and we're unable to find any lookup functions for the parameter names to use as validation.

**Feature Request**
We can see the parameter list here:
https://xgboost.readthedocs.io/en/latest/parameter.html

However, we are worried on how we will maintain our parameter set in case of changes, or human errors like this. Is it possible to provide one of the following, please?

1. An exported function to extract the parameter names for `xgb.train` that we can use for validation
2. A warning/error option if any name in `params` argument doesn't exist

Any suggestions would be appreciated, thank you :)

hcho3:
Option 2 would be really nice to have, but that requires more work. We can go with Option 1 for now. I'm not that familiar with R, so I'll need some help from R experts. For now, I'll prepare a PR for Python interface.

hcho3:
Proposed interface: the user is able to call a special function (`xgboost.Booster.get_params()`) for the purpose of parameter validation:

```python
bst = xgboost.train(...)
params_used = bst.get_params()
```

This is not entirely satisfactory, since the user still needs to perform validation manually. Ideally, XGBoost should warn users about un-used parameters.

hcho3:
How to warn users about un-used parameters: modify all subclasses of `dmlc::Parameter` so as to require a reference to a central parameter registry. All un-used parameters will be tracked by that registry.

```cpp
template<typename PType>
struct TrackedParameter : public dmlc::Parameter<PType> {
   TrackedParameter(UnusedParameterRegistry* registry) : registry_(registry) {}

   template<typename Container>
   inline std::vector<std::pair<std::string, std::string> >
   InitAllowUnknown(const Container &kwargs) {
     auto unused = dmlc::Parameter<PType>::InitAllowUnknown(kwargs);
     registry_->RegisterUnusedParameters(unused);
     return unused;
   }
}
```

RAMitchell:
This will dramatically improve the usability of xgboost. I see a lot of scripts online with mistyped parameters.

CodingCat:
One side note, do we want to address those typos in xgb params?

like num_round, nthread, etc.

hcho3:
@CodingCat Do you mean auto-correcting? For now, warning the user about un-used parameter should be sufficient.

[...]

PR #5157, opened 2019-12-25
Parameter validation

trivialfis:
Related issue: https://github.com/dmlc/xgboost/issues/4211

Using the first option described in: https://github.com/dmlc/xgboost/issues/4211#issuecomment-568661064 

Note the parameter validation is not strict, because we have a lots of parameters packed into `TrainParam` and each tree method uses only a selected subset.  But to me this is a good starting point for future development.  The difficult part of getting it work is refactoring `approx`.

I didn't split the parameters exclusive to CPU Hist, as I think some of them are useful to GPU in the future.

### DIFF
--- a/src/gbm/gbtree.cc
+++ b/src/gbm/gbtree.cc
@@ -359,9 +359,7 @@ class Dart : public GBTree {
 
   void Configure(const Args& cfg) override {
     GBTree::Configure(cfg);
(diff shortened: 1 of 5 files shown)


trivialfis:
Nah, one way or the other we will have the same problem.  We can refactor the approx after this release.

PR #5160, opened 2019-12-26
Fix wrapping GPU ID.

trivialfis:
* Removed some data copying.
* Make sure `gpu_id` is valid before any configuration is carried out.

Closes https://github.com/dmlc/xgboost/issues/4618 .

### DIFF
--- a/src/data/simple_dmatrix.cc
+++ b/src/data/simple_dmatrix.cc
@@ -41,7 +41,7 @@ BatchSet<SparsePage> SimpleDMatrix::GetRowBatches() {
 BatchSet<CSCPage> SimpleDMatrix::GetColumnBatches() {
   // column page doesn't exist, generate it
   if (!column_page_) {
-    auto page = dynamic_cast<SimpleCSRSource*>(source_.get())->page_;
+    auto const& page = dynamic_cast<SimpleCSRSource*>(source_.get())->page_;
     column_page_.reset(new CSCPage(page.GetTranspose(source_->info.num_col_)));
   }
   auto begin_iter =
@@ -52,7 +52,7 @@ BatchSet<CSCPage> SimpleDMatrix::GetColumnBatches() {
 BatchSet<SortedCSCPage> SimpleDMatrix::GetSortedColumnBatches() {
   // Sorted column page doesn't exist, generate it
   if (!sorted_column_page_) {
-    auto page = dynamic_cast<SimpleCSRSource*>(source_.get())->page_;
+    auto const& page = dynamic_cast<SimpleCSRSource*>(source_.get())->page_;
     sorted_column_page_.reset(
         new SortedCSCPage(page.GetTranspose(source_->info.num_col_)));
     sorted_column_page_->SortRows();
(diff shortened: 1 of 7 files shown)


src/data/simple_dmatrix.cc
```diff
@@ -41,7 +41,7 @@ BatchSet<SparsePage> SimpleDMatrix::GetRowBatches() {
 BatchSet<CSCPage> SimpleDMatrix::GetColumnBatches() {
   // column page doesn't exist, generate it
   if (!column_page_) {
-    auto page = dynamic_cast<SimpleCSRSource*>(source_.get())->page_;
+    auto const& page = dynamic_cast<SimpleCSRSource*>(source_.get())->page_;
```
RAMitchell: Isnt this a pointer, so there is no copy?

src/learner.cc
```diff
@@ -275,7 +275,8 @@ class LearnerImpl : public Learner {
       // `verbosity` in logger is not saved, we should move it into generic_param_.
       // FIXME(trivialfis): Make eval_metric a training parameter.
       if (kv.first != "num_feature" && kv.first != "verbosity" &&
-          kv.first != "num_class" && kv.first != kEvalMetric) {
+          kv.first != "num_class" && kv.first != "num_output_group" &&
```
RAMitchell: What is this change for?

tests/cpp/data/test_sparse_page_dmatrix.cc
```diff
@@ -51,14 +51,14 @@ TEST(SparsePageDMatrix, ColAccess) {
   EXPECT_EQ(dmat->GetColDensity(1), 0.5);
 
   // Loop over the batches and assert the data is as expected
-  for (auto col_batch : dmat->GetBatches<xgboost::SortedCSCPage>()) {
+  for (auto const& col_batch : dmat->GetBatches<xgboost::SortedCSCPage>()) {
```
RAMitchell: I wonder if we can delete the copy constructor to prevent this in future.

tests/python-gpu/test_gpu_pickling.py
```diff
@@ -31,6 +34,12 @@ def load_pickle(path):
 
 
 class TestPickling(unittest.TestCase):
+    args_tempale = [
```
RAMitchell: Should be `args_template`?

src/data/simple_dmatrix.cc
trivialfis: No, it's not a pointer.  See the tailing `page_`

src/learner.cc
trivialfis: Oh, I missed this parameter is automatically added when using multi class model.

tests/cpp/data/test_sparse_page_dmatrix.cc
trivialfis: Yes.  I tried to delete the copy constructor of Host device vector and found these.  Will get a PR with another fix.

tests/python-gpu/test_gpu_pickling.py
trivialfis: Good catch.

src/data/simple_dmatrix.cc
RAMitchell: That is quite bad that this could happen. Good job finding it.

Issue #5164, opened 2019-12-27
Removing num_roots breaks tree lite.


trivialfis:
Echoing from @sperlingxx , tree lite checks num_roots parameter internally, we should keep maintenance for this parameter in recent future.

hcho3:
@trivialfis Treelite only checks if the number of roots is 1. It can be removed.

trivialfis:
@hcho3  It wasn't too difficult to keep it, we can leave it there and deprecate it along with binary IO when JSON is production ready.  This way we can show a proper warning.

PR #5165, opened 2019-12-27
Fix num_roots to be 1.

trivialfis:
Closes https://github.com/dmlc/xgboost/issues/5164 .

~Need to verify.~


trivialfis:
Tested against master branch of treelite.

PR #5166, opened 2019-12-28
Fix span constructor.

trivialfis:
* Make sure copy constructor is used.
* Constructor for container no-longer available for device code.


trivialfis:
@RAMitchell This is a bug fix, the constructors use compile time dispatching, which choose one constructor when input itself is a Span, and use another constructor for general container like std::vector.  Before this PR the constructor for general container is always used.

PR #5169, opened 2019-12-29
Don't use `modernize-use-trailing-return-type`.

trivialfis:


hcho3:
What is the rationale of this change?

trivialfis:
@hcho3 Newer clang-tidy has this option for forcing
``` c++
auto foo() -> int {
}
```
style of writing code, I don't think we will do a whole refactor to accommodate it right?  As it's more a styling issue instead of solving any real problem.

trivialfis:
It shows up in my editor, kind of distracting.

PR #5171, opened 2019-12-30
[BLOCKING] [jvm-packages] add gpu_hist and enable gpu scheduling

wbo4958:
Add gpu_hist tree_method to leverage GPU to accelerate XGBoost running on spark


wbo4958:
> Why is this fixed to be 1?

No big deal here, It's just an example. The reason here is it's easy to find **one** GPU to test.

hcho3:
Not sure if this is right time, since without Spark 3.0 there isn’t a good way to allocate GPUs in a cluster.

wbo4958:
> These checks seems to be duplicated with C++ code.

Yes, But we can early detect the error in Driver.

hcho3:
Also see discussion in https://github.com/dmlc/xgboost/pull/4175. Can you explain how this PR addresses the concerns raised in #4175?

wbo4958:
> Not sure if this is right time, since without Spark 3.0 there isn’t a good way to allocate GPUs in a cluster.

Yes, but for now, we can fully leverage the single GPU on each worker (running on the different host). And then we can add GPU allocation supporting when spark 3.0 releases.

hcho3:
> but for now, we can fully leverage the single GPU on each worker (running on the different host)

This seems like an assumption that may not hold in practice. There is no assurance that every machine will have one GPU, and no assurance that each worker will be in a different physical node. That is, what if two workers try to access the same GPU? So I think GPU query and allocation mechanism will be still necessary.

wbo4958:
> > but for now, we can fully leverage the single GPU on each worker (running on the different host)
> 
> This seems like an assumption that may not hold in practice. There is no assurance that every machine will have one GPU, and no assurance that each worker will be in a different physical node. That is, what if two workers try to access the same GPU? So I think GPU query and allocation mechanism will be still necessary.

If we directly submit XGboost application to the existing spark, Yes, I agree that will be a problem. But if we re-deploy spark, just like I said we can deploy spark and configure application paras to make XGBoost run on single GPU on each worker running on the different host.

Currently, **it's SNAPSHOT**, So why don't we first add gpu_hist supporting to let users can use GPU accelerating. 

And we all know Spark has absorbed GPU Scheduling in upcoming spark 3.0 release. Once spark 3.0 releases, we can add GPU allocation something.

AFAIK, The first version of GPU scheduling is one whole GPU allocation, and then evolves to the GPU fraction allocation, and finally evolves Stage Level scheduling. Gpu scheduling is not completed in one commit.

wbo4958:
> Also see discussion in #4175. Can you explain how this PR addresses the concerns raised in #4175?

No, This PR can't address the concerns of #4175. But currently the good news is that XGBoost is in SNAPSHOT, and spark 3.0 SNAPSHOT has supported GPU scheduling. So we can add GPU related one by one.

trivialfis:
Is there any guard against misuse?

hcho3:
My suggestion is to delay this PR **after** 1.0 release.

wbo4958:
> Personally I’d still like to wait for Spark 3.0 release, but I can live with this PR being merged into the codebase, as long as the experimental nature is made super clear. At the very minimum we should delay merging of this PR **until after** 1.0 release, given the experimental nature of the PR.

Thx. I'm fine with this. 
BTW, Will 1.0 release be based on Spark 3.0?

hcho3:
Spark 3.0 is currently in preview, and to my knowledge no date has been set yet (correct me if I’m wrong). So most likely XGBoost 1.0 would be out before Spark 3.0

wbo4958:
> Spark 3.0 is currently in early preview, and to my knowledge no date has been set yet (correct me if I’m wrong). So most likely XGBoost 1.0 would be out before Spark 3.0

Got it. Thx for the info.

CodingCat:
the current XGB is compiled with 2.4 which is very likely binary incompatible with 3.0, 

can we really merge this PR and ship to the user? based on my knowledge, summer of next year is the expected date for Spark 3.0 leading me to put less hope to make 3.0 as the default spark version for xgb 1.0

hcho3:
@CodingCat I made it clear that this PR cannot be part of 1.0.

wbo4958:
> the current XGB is compiled with 2.4 which is very likely to be binary incompatible with 3.0,
> 
> can we really merge this PR and ship to the user? based on my knowledge, the summer of next year is the expected date for Spark 3.0 leading me to put less hope to make 3.0 as the default spark version for xgb 1.0

Thx for the info :)

wbo4958:
added gpu scheduling

wbo4958:
Hi @CodingCat @hcho3 @trivialfis. XGBoost has updated the spark dependency to Spark 3.0 which has already supported GPU scheduling. So I think maybe it's the time to take advantage of GPU's power to speed up XGBoost4j.

In this PR, I added the gpu_hist tree_method which can allow the user to choose GPU alogrithim of XGBoost library. and I also added the GPU scheduling to get the GPU device from Spark.

I've done some tests for this PR for both classification(Mortgage dataset) and regression(taxi dataset) in spark 3.0.0

| spark cluster   |  node|    gpus      |  workers | result |
|----------|:-------------:|:-------------:|:-------------:|------:|
| spark local | 1 node|  1 gpu  | 1 xgb worker | pass |
| spark standalone | 1 node|   2 gpus   |  2 xgb worker | pass |
| spark standalone | 2 nodes|   (1 + 2) gpus   |  3 xgb worker | pass |
| spark yarn | 5 nodes|   1 gpu per executor   |  2 xgb worker | pass |

Pls help to review.

wbo4958:
Hi @CodingCat @hcho3 @trivialfis I changed the XGBoost4j jars compiling behaviour a little bit. eg. to enable CUDA for XGBoost4j jars. For now, we can run below command
```
mvn -Duse.cuda=ON package
```

hcho3:
@wbo4958 If separating build from test is too difficult, feel free to build and test GPU-enabled JVM packages in one step.

wbo4958:
Thx @hcho3 , I will run locally first.

wbo4958:
Hi @hcho3 @CodingCat, I have added the CI build and test for this PR, pls help to review.

jvm-packages/create_jni.py
```diff
@@ -7,6 +7,10 @@
 import sys
 from contextlib import contextmanager
 
+# usage
+# ./create_jni.py [ON|OFF]
+#  ON : USE_CUDA=ON
+#  OFF: USE_CUDA=OFF
```
hcho3, line 13: Let's use `argparse` Python module to provide a proper CLI interface:
```
python create_jni.py --use-cuda ON
python create_jni.py --use-cuda OFF
python create_jni.py  # CUDA is off by default
```

If you'd like, I can push to this branch.

hcho3:
cc @WeichenXu123 @trams @sperlingxx Would you like to take a look at proposed GPU extension of XGBoost4J-Spark? Now that we moved to Spark 3.0, we have a well-defined mechanism to allocate GPUs in clusters.

CodingCat:
> Changes look good to me. Some comments:
> 
> * This PR only adds integration tests for `gpu_hist`. We should add unit tests as well. Do you think they can be added to this PR in reasonable amount of time?
> * We should get approval from those more knowledgeable about JVM packages.

one thing to think about is, if we add gpu based unit test,  it will be hard for devs who has no gpu hardware in their dev environment to build xgb

hcho3:
@CodingCat That's a good point. Is there a way to add a separate suite of unit tests that could run optionally? In Python, it is possible to select a set of unit test suites to run.

CodingCat:
> @CodingCat That's a good point. Is there a way to add a separate suite of unit tests that could run optionally? In Python, it is possible to select a set of unit test suites to run.

one possible way might be using different profiles and ignore some tests with the approach https://stackoverflow.com/questions/4267852/ignoring-maven-scala-unit-test-files (not-verified)

wbo4958:
@hcho3 @CodingCat Thx, I will try to add some unit tests for gpu_hist

jvm-packages/create_jni.py
wbo4958: Hi @hcho3, Pls help with that.

wbo4958:
Hi @hcho3, when compiling jvm packages, the added gpu unit tests will run and will require the GPU. But from the log, seems the building node didn't contain a GPU. The error log can be found as follows:
```
ml.dmlc.xgboost4j.java.XGBoostError: [11:25:42] /workspace/src/gbm/gbtree.cc:459: Check failed: common::AllVisibleGPUs() >= 1 (0 vs. 1) : No visible GPU is found for XGBoost.
```

jenkins node("linux && gpu") https://github.com/dmlc/xgboost/pull/5171/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R288

The error log can be found in  https://xgboost-ci.net/blue/rest/organizations/jenkins/pipelines/xgboost/branches/PR-5171/runs/18/nodes/116/log/?start=0

Could you help with that?

Jenkinsfile
```diff
@@ -282,6 +284,28 @@ def BuildCUDA(args) {
   }
 }
 
+def BuildJVMPackagesWithCUDA(args) {
+  node('linux && gpu') {
+    unstash name: 'srcs'
+    echo "Build XGBoost4J-Spark with Spark ${args.spark_version}, CUDA ${args.cuda_version}"
+    def container_type = "jvm_gpu_build"
+    def docker_binary = "docker"
```
hcho3, line 292:
```suggestion
    def docker_binary = "nvidia-docker"
```

hcho3: This change will make GPUs available inside the Docker image. @wbo4958

jvm-packages/pom.xml
```diff
@@ -52,7 +53,65 @@
...
+                        <groupId>org.scalatest</groupId>
+                        <artifactId>scalatest-maven-plugin</artifactId>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <!-- gpu-with-gpu-tests profile with only gpu test suites -->
+        <profile>
+            <id>gpu-with-gpu-tests</id>
```
CodingCat, line 98: so we really need two profile? mvn does support something like testing a particular suite

wbo4958: Hi @CodingCat, Thx for the review. 

I suppose "mvn does support something like testing a particular suite" is specifying like 
```
 mvn test -Dsuites=ml.dmlc.xgboost4j.scala.DMatrixSuite
```

Yeah, Seems mvn can specify which test suite to run, Suppose the test suite **must be** package name. So if PR has added another gpu related suites, then developer needs to change the CI build script to add another gpu suite.

This profile is only for CI used, and it will pick any suite tagged with GpuTestSuite to run. So the only thing for developer to do is just adding a GpuTestSuite tag to the new gpu suites, developer does not need to touch CI part,  then everything will work as expected.

wbo4958:
@hcho3 @CodingCat Could you review again. CI passed

jvm-packages/create_jni.py
hcho3: I pushed a new commit.

CodingCat:
merged, thanks!

wbo4958:
Thx @CodingCat

PR #5173, opened 2019-12-30
Lazy initialization of device vector.

### LINKED ISSUE #5162: Fix running CUDA compiled XGBoost on CPU only machine
Recently I found XGBoost compiled with latest CUDA broken on CPU only machine.  Specifically the gradient vector tries to initialize CUDA context and throws an error during destruction due to `thrust::device_vector` trying to call device function to destroy elements.

trivialfis:
* Fix #5162.

* Disable copy constructor of HostDeviceVector.  Prevents implicit copying.

### DIFF
--- a/include/xgboost/data.h
+++ b/include/xgboost/data.h
@@ -59,8 +59,25 @@ class MetaInfo {
    * can be used to specify initial prediction to boost from.
    */
   HostDeviceVector<bst_float> base_margin_;
+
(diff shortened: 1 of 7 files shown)


trivialfis:
@RAMitchell 
``` c++
this->data_ = HostDeviceVector{1, 2, 3, 4, 5};
```

Not important, but convenient for writing tests.

PR #5178, opened 2020-01-01
[WIP] [Breaking] Default to hist `tree_method`.

trivialfis:
Currently XGBoost defaults to exact and approx for larger dataset.  But as `hist` tree method is more efficient, we may want to set it as default in 1.0.

- Set `hist` to default.
- Don't test external memory training on unsupported tree methods.
- Requires users to set appropriate tree method in distributed environment and external memory training, removing auto configurations.
- Add `refresh` as a tree method.

Passed all tests in C++/Python/R on my local machine, expecting some glitches on JVM side.

### DIFF
--- a/R-package/tests/testthat/test_update.R
+++ b/R-package/tests/testthat/test_update.R
@@ -15,17 +15,18 @@ test_that("updating the model works", {
   watchlist = list(train = dtrain, test = dtest)
 
   # no-subsampling
-  p1 <- list(objective = "binary:logistic", max_depth = 2, eta = 0.05, nthread = 2)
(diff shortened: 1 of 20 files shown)


trivialfis:
@CodingCat Is it expected for jvm package to have different number of columns for each worker?

I got a hanging test `"infrequent features"` which launch 2 workers and one has 2 columns while the other has 3 columns.

trivialfis:
@CodingCat 

There are a few issues here:
- https://github.com/dmlc/xgboost/pull/4954#issuecomment-544367296
- the `checkCall` is performed in java instead of c++, so when XGBoost has caught an error, the binding code in `xgboost4j.cpp` continues the execute.  Which may lead to segfault as the output of C API might not be properly set due to exception control flow.

PR #5180, opened 2020-01-03
Add new lines for spark tutorial missing values section

cpfarrell:
Realized from looking at https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html#dealing-with-missing-values that the options were kind of hard to look at as didn't provide new lines so added them in. Not sure if better way to format this/not really sure how to preview them beforehand.

Also, hadn't really known doc changes would go live immediately, I thought would only go live once new option was available in a version


trivialfis:
Thanks!

trivialfis:
It's will be visible at master doc after building.

PR #5191, opened 2020-01-09
Enforce correct data shape.

trivialfis:
Some related fixes ported from https://github.com/dmlc/xgboost/pull/5178 .

* Added a check call macro in jvm package, prevents executing other functions from jvm when error occurred in XGBoost.  For example, when prediction fails jvm should not try to allocate memory based on the output prediction size.
* Update doc to recommend `hist`/`gpu_hist` algorithm.
* Add a parameter `validate_features` and enable it for `R/Python/C` except for `jvm`.
* Sync maximum number of columns in DMatrix across workers.  This also fixes empty dask cudf dataframe.

### DIFF
--- a/doc/parameter.rst
+++ b/doc/parameter.rst
@@ -112,18 +112,26 @@ Parameters for Tree Booster
 
   - The tree construction algorithm used in XGBoost. See description in the `reference paper <http://arxiv.org/abs/1603.02754>`_.
   - XGBoost supports  ``approx``, ``hist`` and ``gpu_hist`` for distributed training.  Experimental support for external memory is available for ``approx`` and ``gpu_hist``.
-  - Choices: ``auto``, ``exact``, ``approx``, ``hist``, ``gpu_hist``
+
+  - Choices: ``auto``, ``exact``, ``approx``, ``hist``, ``gpu_hist``, this is a
+    combination of commonly used updaters.  For other updaters like ``refresh``, set the
+    parameter ``updater`` directly.
 
     - ``auto``: Use heuristic to choose the fastest method.
 
-      - For small to medium dataset, exact greedy (``exact``) will be used.
-      - For very large dataset, approximate algorithm (``approx``) will be chosen.
-      - Because old behavior is always use exact greedy in single machine,
-        user will get a message when approximate algorithm is chosen to notify this choice.
+      - For small dataset, exact greedy (``exact``) will be used.
+      - For larger dataset, approximate algorithm (``approx``) will be chosen.  It's
+        recommended to try ``hist`` and ``gpu_hist`` for higher performance with large
+        dataset.  The reasons they aren't default yet are because XGBoost has some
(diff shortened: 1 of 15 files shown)


doc/parameter.rst
```diff
@@ -112,18 +112,26 @@ Parameters for Tree Booster
...
     - ``auto``: Use heuristic to choose the fastest method.
 
-      - For small to medium dataset, exact greedy (``exact``) will be used.
-      - For very large dataset, approximate algorithm (``approx``) will be chosen.
-      - Because old behavior is always use exact greedy in single machine,
-        user will get a message when approximate algorithm is chosen to notify this choice.
+      - For small dataset, exact greedy (``exact``) will be used.
+      - For larger dataset, approximate algorithm (``approx``) will be chosen.  It's
+        recommended to try ``hist`` and ``gpu_hist`` for higher performance with large
+        dataset.  The reasons they aren't default yet are because XGBoost has some
```
RAMitchell: > The reasons they aren't default yet are because XGBoost has some
        existing tests tied to ``approx``, and only GPU version of ``hist``
        (``gpu_hist``)has support for ``external memory``.

I don't think documentation needs to contain these details related to roadmaps and issues.

doc/parameter.rst
```diff
@@ -112,18 +112,26 @@ Parameters for Tree Booster
...
+        (``gpu_hist``)has support for ``external memory``.
 
-    - ``exact``: Exact greedy algorithm.
+      - Because old behavior is always use exact greedy in single machine, user will get a
+        message when approximate algorithm is chosen to notify this choice.
+    - ``exact``: Exact greedy algorithm.  Enumerates all split candidates.
     - ``approx``: Approximate greedy algorithm using quantile sketch and gradient histogram.
-    - ``hist``: Fast histogram optimized approximate greedy algorithm. It uses some performance improvements such as bins caching.
+    - ``hist``: Faster histogram optimized approximate greedy algorithm. It uses some
+      performance improvements such as bins caching.
```
RAMitchell: > bins caching

I dont really know what this means.

jvm-packages/xgboost4j/src/native/xgboost4j.cpp
```diff
@@ -22,6 +22,11 @@
 #include <vector>
 #include <string>
 
+#define JVM_CHECK_CALL(__errcode)                   \
```
RAMitchell: This JVM stuff needs to be separated and reviewed by JVM contributers.

doc/parameter.rst
trivialfis: @hcho3

jvm-packages/xgboost4j/src/native/xgboost4j.cpp
trivialfis: It comes up when enforcing the number of columns.  Will split it up.

PR #5197, opened 2020-01-13
Explicitly use UTF-8 codepage when using MSVC

hcho3:
One of C++ tests contain non-ASCII letters:
https://github.com/dmlc/xgboost/blob/8cbcc53ccbd3d2c4c74ef35fcec018be64d084cd/tests/cpp/common/test_config.cc#L154-L156

Microsoft Visual C++ (MSVC) uses the system codepage by default when compiling C++ sources. This resulted in error `Error C2001: Newline in constant` when I compiled XGBoost on my Windows laptop, which had Korean (CP949) locale.

This PR explicitly sets the compilation flag `/utf-8` so that MSVC always uses UTF-8 encoding. See related discussion at https://github.com/libusb/libusb/issues/207.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -33,6 +33,12 @@ if (USE_CUDA)
     $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=${OpenMP_CXX_FLAGS}>
   )
 
+  if (MSVC)
+    target_compile_options(objxgboost PRIVATE
+      $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=/utf-8>
+    )
+  endif (MSVC)
+
   set_target_properties(objxgboost PROPERTIES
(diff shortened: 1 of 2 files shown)


trivialfis:
Consider using `u8 ` string prefix.

hcho3:
@trivialfis Using `u8` prefix alone is insufficient, because MSVC will try to parse the source file using the system codepage unless `/utf-8` is specified. This will result in error on some locales, such as Korean (CP949).

> Makefile is also related if we rely on build system.

No, since the problem is limited to Visual Studio on Windows, for which we exclusively use CMake already. GCC and other compilers do not suffer from this issue.

hcho3:
https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=vs-2019

PR #5198, opened 2020-01-13
Implement slice via adapters

RAMitchell:
This PR allows us to consolidate further constructions of DMatrix into a single constructor path.

### DIFF
--- a/src/data/adapter.h
+++ b/src/data/adapter.h
@@ -8,6 +8,8 @@
 #include <limits>
 #include <memory>
 #include <string>
+#include <utility>
+#include <vector>
 
 namespace xgboost {
 namespace data {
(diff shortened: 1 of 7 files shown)


src/data/adapter.h
```diff
@@ -481,6 +485,92 @@ class FileAdapter : dmlc::DataIter<FileAdapterBatch> {
...
+  // Fetch metainfo values according to sliced rows
+  template <typename T>
+  std::vector<T> Gather(const std::vector<T>& in) {
+    if (in.empty()) return {};
+
+    std::vector<T> out(this->Size());
+    for (auto i = 0ull; i < this->Size(); i++) {
+      out[i] = in[ridx_set[i]];
+    }
+    return std::move(out);
```
trivialfis: I usually get a warning from clang that says this prevents copy elision and just return the value.

PR #5199, opened 2020-01-13
[jvm-packages] Add JVM_CHECK_CALL.

trivialfis:
* Add a check call macro in jvm package, prevents executing other functions
from jvm when error occurred in XGBoost. For example, when prediction fails jvm
should not try to allocate memory based on the output prediction size.


trivialfis:
Don't understand why, so now the C API wrapper is returning correct error code but failure handling tests are all failing?

trivialfis:
@CodingCat ping.

trivialfis:
@CodingCat Friendly ping. ;-)

PR #5202, opened 2020-01-14
Add number of columns to native data iterator.

trivialfis:
This helps mitigating imputing sparse dataset from JVM package, where some columns might be completely missing.  @wbo4958 will submit a follow up PR on JVM side.

For consistency and simplicity, `NativeDataIter` is turned into an adapter.

As a new data field is added, we break the ABI.

### DIFF
--- a/src/data/adapter.h
+++ b/src/data/adapter.h
@@ -1,18 +1,26 @@
 /*!
- *  Copyright (c) 2019 by Contributors
+ *  Copyright (c) 2019~2020 by Contributors
  * \file adapter.h
  */
 #ifndef XGBOOST_DATA_ADAPTER_H_
 #define XGBOOST_DATA_ADAPTER_H_
 #include <dmlc/data.h>
+
+#include <cstddef>
+#include <functional>
 #include <limits>
 #include <memory>
 #include <string>
 #include <utility>
 #include <vector>
 
+#include "xgboost/logging.h"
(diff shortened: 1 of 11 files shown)


trivialfis:
@CodingCat

trivialfis:
We do something similar on dask, but delegate the task to np/pd.

trivialfis:
Pls don't merge.

trivialfis:
Will wait for the corresponding jvm PR and merge them together.

src/data/adapter.h
```diff
@@ -485,6 +493,112 @@ class FileAdapter : dmlc::DataIter<FileAdapterBatch> {
...
+  // override functions
+  void BeforeFirst() override {
+    CHECK(at_first_) << "Cannot reset IteratorAdapter";
+  }
+
+  bool Next() override {
+    if ((*next_callback_)(
+            data_handle_,
+            [](void *handle, XGBoostBatchCSR batch) -> int {
+              API_BEGIN();
```
RAMitchell: Would be nice if we can avoid bringing c api headers here somehow.

trivialfis: @RAMitchell I can push the function into a CC file.  `XGBoostBatchCSR` is needed in this adapter.

trivialfis: Actually I can't.  The callback function signature is also needed.

PR #5206, opened 2020-01-15
Support dmatrix construction from cupy array

RAMitchell:
A few things left to do:

- [x] Correctly construct labels/info
- [x] Add an integration test for training
- [x] Mention cupy in documentation

@trivialfis I removed the check for array interface version as I found I could get version 0 by installing cupy from conda or 2 by installing from pip, but not 1. Does this break anything for cudf?

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -20,7 +20,7 @@ import scipy.sparse
 from .compat import (
     STRING_TYPES, DataFrame, MultiIndex, Int64Index, py_str,
     PANDAS_INSTALLED, DataTable,
-    CUDF_INSTALLED, CUDF_DataFrame, CUDF_Series, CUDF_MultiIndex, CUPY_Array,
+    CUDF_INSTALLED, CUDF_DataFrame, CUDF_Series, CUDF_MultiIndex,
     os_fspath, os_PathLike)
 from .libpath import find_lib_path
 
@@ -502,7 +502,7 @@ class DMatrix(object):
             self._init_from_dt(data, nthread)
         elif _use_columnar_initializer(data):
             self._init_from_columnar(data, missing, nthread)
-        elif isinstance(data, CUPY_Array):
+        elif hasattr(data, "__cuda_array_interface__"):
             self._init_from_cupy(data, missing, nthread)
         else:
             try:
@@ -709,14 +709,11 @@ class DMatrix(object):
                                                c_bst_ulong(len(data))))
 
     def set_interface_info(self, field, data):
-        """Set info type peoperty into DMatrix."""
-        if isinstance(data, CUPY_Array):
-            interfaces = bytes(json.dumps([data.__cuda_array_interface__], indent=2), 'utf-8')
-        else:
-            interfaces = _extract_interface_from_cudf(data)
+        """Set info type property into DMatrix."""
+        interface = bytes(json.dumps([data.__cuda_array_interface__], indent=2), 'utf-8')
         _check_call(_LIB.XGDMatrixSetInfoFromInterface(self.handle,
                                                        c_str(field),
-                                                       interfaces))
+                                                       interface))
 
     def set_float_info_npy2d(self, field, data):
         """Set float type property into the DMatrix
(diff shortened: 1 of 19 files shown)


trivialfis:
No.  But would be nice the have it somewhere in the path.  I remember last time someone has a super old numba with cudf and caused a failure.

tests/python/testing.py
```diff
@@ -1,6 +1,6 @@
 # coding: utf-8
 from xgboost.compat import SKLEARN_INSTALLED, PANDAS_INSTALLED, DT_INSTALLED
-from xgboost.compat import CUDF_INSTALLED, DASK_INSTALLED
+from xgboost.compat import CUDF_INSTALLED, CUPY_INSTALLED, DASK_INSTALLED
```
trivialfis: Can we merge this after the release?  I would like to investigate lazy library loading.  These libraries are all heavy in size.

RAMitchell: Yep, I just measured it takes 0.6s to import cupy vs 1.73s for xgboost.

RAMitchell:
@trivialfis I got this working without any imports. It might be possible to do the same with cudf, particularly if they fix the incomplete `__cuda_array_interface__` attribute.

jakirkham:
> It might be possible to do the same with cudf, particularly if they fix the incomplete `__cuda_array_interface__` attribute.

@RAMitchell, could you please raise an issue on the cuDF repo documenting in which ways it is incomplete?

RAMitchell:
@jakirkham it was my misunderstanding. I thought the mask attribute was supposed to be a `__cuda_array_interface__` , looking closer at the standard it should be an object that has the attribute `__cuda_array_interface__`.

python-package/xgboost/core.py
```diff
@@ -862,7 +848,7 @@ def set_group(self, group):
         group : array like
             Group size of each group
         """
-        if _use_columnar_initializer(group):
+        if hasattr(group, "__cuda_array_interface__"):
```
trivialfis: Does a cudf dataframe with single column have the array interface?  Also does a cudf series have array interface?  Can you be sure that the cudf test is using correct initializer?

src/c_api/c_api.cu
```diff
@@ -7,14 +7,28 @@
...
   std::string json_str{c_json_strs};
   data::CudfAdapter adapter(json_str);
-  *out = new std::shared_ptr<DMatrix>(DMatrix::Create(&adapter, missing, 1));
+  *out =
+      new std::shared_ptr<DMatrix>(DMatrix::Create(&adapter, missing, nthread));
   API_END();
 }
+
+XGB_DLL int XGDMatrixCreateFromArrayInterface(char const* c_json_strs,
+                                                     bst_float missing,
```
trivialfis: Wrong indent.

python-package/xgboost/core.py
trivialfis: > Does a cudf dataframe with single column have the array interface?

If it does, then we need to make sure the cudf DataFrame is not initialized with `cupy` initializer.

python-package/xgboost/core.py
```diff
@@ -500,8 +461,10 @@ def __init__(self, data, label=None, weight=None, base_margin=None,
             self._init_from_npy2d(data, missing, nthread)
         elif isinstance(data, DataTable):
             self._init_from_dt(data, nthread)
-        elif _use_columnar_initializer(data):
-            self._init_from_columnar(data, missing)
+        elif hasattr(data, "__cuda_array_interface__"):
+            self._init_from_array_interface(data, missing, nthread)
+        elif isinstance(data, CUDF_DataFrame):
```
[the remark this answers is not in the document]
RAMitchell: That sounds great if possible.

python-package/xgboost/core.py
RAMitchell: cudf with one column does not have array interface. I will modify it to accept this for metainfo.

jakirkham:
> it was my misunderstanding. I thought the mask attribute was supposed to be a `__cuda_array_interface__` , looking closer at the standard it should be an object that has the attribute `__cuda_array_interface__`.

No worries. Just want to make sure when we are discovering issues they get raised so we can address them 🙂

PR #5208, opened 2020-01-15
Add Optuna badge to README.md

Crissman:
Hi!

[Optuna]( (optuna.org) ) is a new hyperparameter optimization library, and we've written an integration module for XGBoost to make it easy to use Optuna to search for good hyperparameter settings and prune unpromising trials. We're looking for ways to let users of XGboost know this is available and thought a badge to show Optuna integration is available would be helpful.

If there are other ways you would recommend reaching out to XGBoost users to let them know about Optuna, please let us know.

Thanks!


Crissman:
Here's our example of using the pruning integration with XGBoost: https://github.com/optuna/optuna/blob/9942ec023f19cb1d3f1e053f9573792e43fcdedd/examples/pruning/xgboost_integration.py

PR #5212, opened 2020-01-17
Added static functions

gaurav1086:


trivialfis:
Why not call instance instead?

trivialfis:
Closing as I believe using non-static function is more flexible.  It's actually a "common" code used in debugging, I sometimes define a local instance with string name, static function will disable that.

PR #5215, opened 2020-01-18
Fix compilation error in c-api-demo.c

mattn:


trivialfis:
Oops, forgot the demo, I should add it as part of test.

trivialfis:
Thanks!

PR #5218, opened 2020-01-21
Add dart to JSON schema.

trivialfis:


trivialfis:
The schema describes output model.  If you change the output model then the schema needs to be updated.

Issue #4774, opened 2019-08-14
XGBoost Spark training quite slow - Good practices


kumarprabhu1988:
Hello!

I posted a question about some OOM errors I was facing with training here: https://discuss.xgboost.ai/t/xgboost4j-spark-fails-with-oom-errors/1054. Thankfully, I was able to resolve these issues and found a configuration that works. However, it takes ~16 min for 1 round. I'm looking for tips to speed this up. 

My data is ~120GB before transforming into a vector. I'm using spark 2.3.2 and xgboost 0.82. My configuration ensures most of the training and validation data are present in memory. I'm using `spark.memory.storageFraction = 0.16` which is quite low. Execution apparently requires lot of memory. I get OOM errors if I increase this value. I've tried increasing `spark.memory.fraction` as well and I get OOM errors if I increase it to 0.7. 

Here's my full spark configuration:
```
spark.executor.cores             8
spark.driver.memory              11171M
spark.executor.memory            10000M
spark.executor.memoryOverhead    15600M
spark.default.parallelism        1280
spark.sql.shuffle.partitions     2000
spark.memory.fraction            0.6
spark.memory.storageFraction     0.16
spark.task.cpus                  4
```

For xgboost, I use this configuration:
```
val booster = new XGBoostClassifier(
  Map(
    "missing" -> -999999.0,
    "booster" -> "gbtree",
    "objective" -> "binary:logistic",
    "eval_metric" -> "logloss",
    "tree_method" -> "approx",
    "eta" -> 0.2,
    "gamma" -> 1.0,
    "alpha" -> 20,
    "max_depth" -> 4,
    "num_round" -> 1800,
    "num_workers" -> 160,
    "nthread" -> 4,
    "timeout_request_workers" -> 60000L
  )
).setLabelCol(targetVar)
.setEvalSets(evalSet)
.setUseExternalMemory(true)
.setCheckpointInterval(2)
.setCheckpointPath("checkpoints_path")
```

FYI, I can train this data on a single super large machine and it takes ~1 min per iteration (though the first iteration takes more than 1 hour in addition to 0.5 hours for loading data) on this machine. The goal is to move this whole training process to `xgboost-spark` so it can scale with the data and we don't have to get larger machines. 

Posting here because I didn't get any responses on the discussion forum. 

Any help will be appreciated.
Thank you!
Kumar

chenqin:
@kumarprabhu1988 please try 0.9, there was a critical bug in .82

[...]

Issue #4482, opened 2019-05-20
Ability to drop references to underlying DMatrix objects while keeping the booster object alive


vladpetric:
Our DMatrix objects consume a lot of memory. Because the booster class keeps a reference to the DMatrix objects used for training and validation (evals), they never get released (destructed) while the booster object is alive.

We confirmed this by running our test setup under gdb. We observed that C++ shared_ptr references to DMatrix objects grow, but because we keep the Python booster objects around, the shared_ptr references don't decrease to zero.

We'd like to be able to use the boosters without necessarily keeping around the DMatrixes used to train it.

Thank you!

Version: 0.82 (pip installed)
Python: 3.6

trivialfis:
@RAMitchell related to #4354 .  WDYT?

vladpetric:
It certainly seems related to #4354 , though I should say that this is more about how DMatrixes are used and reference-counted, rather than how they are implemented.

trivialfis:
@vladpetric You are right and I agree with your proposal, booster and DMatrix should be decoupled.  But that requires a thorough inspection of current code base, so putting it into the DMatrix refactor plan seems appropriate.

vladpetric:
Thanks!

ivinogra:
I can confirm that this hack will allow deleting a DMatrix (from python):
```
m = xgb.train(params, dmatrix, **more_params)
pd.to_pickle(m, YOUR_PATH)
del m
del dmatrix
m = pd.read_pickle(YOUR_PATH)
```
It is also necessary to `del more_params['evals']` if this parameter was used, and accordingly for other parameters. While DMatrix code is being refactored, is it reasonable to note in the comments to the code that releasing a DMatrix may require deleting any boosters that have trained on it?

trivialfis:
@ivinogra I'm not sure yet.

trivialfis:
Done.

Issue #3946, opened 2018-11-28
[xgboost4j-spark] Checkpointing performance issue


alois-bissuel:
I just noticed a weird performance issue when using checkpoints in the Spark wrapper for XGBoost.
Before the first checkpoint, the learning time is constant per tree, but after the first checkpoint, the time needed to learn a new tree is much higher, and it grows with the number of trees. 

Here is the parameter map used by XGBoost:
`Map(alpha -> 0.0, min_child_weight -> 1.0, sample_type -> uniform, base_score -> 0.5, colsample_bylevel -> 1.0, grow_policy -> depthwise, skip_drop -> 0.0, lambda_bias -> 0.0, silent -> 0, scale_pos_weight -> 1.0, seed -> 0, features_col -> features, num_early_stopping_rounds -> 0, label_col -> label, num_workers -> 5, subsample -> 1.0, lambda -> 1.0, max_depth -> 5, probability_col -> probability, raw_prediction_col -> rawPrediction, tree_limit -> 0, custom_eval -> null, rate_drop -> 0.0, max_bin -> 16, train_test_ratio -> 1.0, use_external_memory -> false, objective -> binary:logistic, eval_metric -> logloss, num_round -> 500, timeout_request_workers -> 1800000, missing -> NaN, checkpoint_path -> viewfs://root/user/XXX/YYYY/checkpoint, tracker_conf -> TrackerConf(0,python), tree_method -> auto, max_delta_step -> 0.0, eta -> 0.3, colsample_bytree -> 1.0, normalize_type -> tree, custom_obj -> null, gamma -> 0.0, sketch_eps -> 0.03, nthread -> 4, prediction_col -> prediction, checkpoint_interval -> 100)`

Attached is a small graph, where this issue is shown on the same dataset, with different checkpointing number. Please disregard the discrepancy between the two run, as they might not have been launched with the same multithreading parameters.

<img width="877" alt="image" src="https://user-images.githubusercontent.com/41194547/49159616-35fe5500-f325-11e8-94ad-fa3b8950da2d.png">

[...]

PR #5220, opened 2020-01-22
Move prediction cache to Learner.

trivialfis:
## Clean-ups

* Remove duplicated cache in Learner and GBM.
* Remove ad-hoc fix of invalid cache.
* Remove `PredictFromCache` in predictors.
* Remove prediction cache for linear altogether, as it's only moving the prediction into training process but doesn't provide any actual overall speed gain.
* The cache is now unique to Learner, which means the ownership is no longer shared by any other components.

## Changes

* Add version to prediction cache.
* Use weak ptr to check expired DMatrix.
* Pass shared pointer instead of raw pointer.

## Risks

Before the DMatrix is tied to Booster with a shared pointer, which means the access to it is always safe as it can't expire before the booster.  Now we need to make sure the safety ourselves.

## Advantages

As now we don't store the shared pointer any more,  this should reduce the memory usage when doing grid search with libraries like Scikit-Learn (during evaluation the training matrix can be released).   Also, performance for training continuation (incremental training, checkpoint recovery) will be significantly better as we cache all encountered DMatrix.

Last thing is this PR is intended to ease the development of our DMatrix refactoring plan, which requires an easy to access and reliable prediction cache, so that we can use quantile values as a proxy to actual data.

### PRs:
#5272

### Issues:
#3946
#4786
#4774
#4482

### DIFF
--- a/include/xgboost/predictor.h
+++ b/include/xgboost/predictor.h
@@ -1,5 +1,5 @@
 /*!
- * Copyright by Contributors
+ * Copyright 2017-2020 by Contributors
  * \file predictor.h
  * \brief Interface of predictor,
  *  performs predictions for a gradient booster.
@@ -32,47 +32,83 @@ namespace xgboost {
  * \brief Contains pointer to input matrix and associated cached predictions.
  */
 struct PredictionCacheEntry {
-  std::shared_ptr<DMatrix> data;
+  // A storage for caching prediction values
   HostDeviceVector<bst_float> predictions;
+  // The version of current cache, corresponding number of layers of trees
+  uint32_t version;
+  // A weak pointer for checking whether the DMatrix object has expired.
+  std::weak_ptr< DMatrix > ref;
+
+  PredictionCacheEntry() : version { 0 } {}
+  /* \brief Update the cache entry by number of versions.
+   *
+   * \param v Added versions.
+   */
+  void Update(uint32_t v) {
+    version += v;
+  }
+};
+
+/* \brief A container for managed prediction caches.
+ */
+class PredictionContainer {
+  std::unordered_map<DMatrix *, PredictionCacheEntry> container_;
+  void ClearExpiredEntries();
+
+ public:
+  PredictionContainer() = default;
+  /* \brief Add a new DMatrix to the cache, at the same time this function will clear out
+   *        all expired caches by checking the `std::weak_ptr`.  Caching an existing
+   *        DMatrix won't renew it.
+   *
+   *  Passing in a `shared_ptr` is critical here.  First to create a `weak_ptr` inside the
+   *  entry this shared pointer is necessary.  More importantly, the life time of this
+   *  cache is tied to the shared pointer.
+   *
+   *  Another way to make a safe cache is create a proxy to this entry, with anther shared
+   *  pointer defined inside, and pass this proxy around instead of the real entry.  But
+   *  seems to be too messy.  In XGBoost, functions like `UpdateOneIter` will have
+   *  (memory) safe access to the DMatrix as long as it's passed in as a `shared_ptr`.
+   *
+   * \param m shared pointer to the DMatrix that needs to be cached.
+   * \param device Which device should the cache be allocated on.  Pass
+   *               GenericParameter::kCpuId for CPU or positive integer for GPU id.
+   *
+   * \return the cache entry for passed in DMatrix, either an existing cache or newly
+   *         created.
+   */
+  PredictionCacheEntry& Cache(std::shared_ptr<DMatrix> m, int32_t device);
+  /* \brief Get a prediction cache entry.  This entry must be already allocated by `Cache`
+   *        method.  Otherwise a dmlc::Error is thrown.
+   *
+   * \param m pointer to the DMatrix.
+   * \return The prediction cache for passed in DMatrix.
+   */
+  PredictionCacheEntry& Entry(DMatrix* m);
+  /* \brief Get a const reference to the underlying hash map.  Clear expired caches before
+   *        returning.
+   */
+  decltype(container_) const& Container();
 };
 
 /**
  * \class Predictor
  *
- * \brief Performs prediction on individual training instances or batches of
- * instances for GBTree. The predictor also manages a prediction cache
- * associated with input matrices. If possible, it will use previously
- * calculated predictions instead of calculating new predictions.
- *        Prediction functions all take a GBTreeModel and a DMatrix as input and
- * output a vector of predictions. The predictor does not modify any state of
- * the model itself.
+ * \brief Performs prediction on individual training instances or batches of instances for
+ *        GBTree. Prediction functions all take a GBTreeModel and a DMatrix as input and
+ *        output a vector of predictions. The predictor does not modify any state of the
+ *        model itself.
  */
-
 class Predictor {
  protected:
   /*
    * \brief Runtime parameters.
    */
   GenericParameter const* generic_param_;
-  /**
-   * \brief Map of matrices and associated cached predictions to facilitate
-   * storing and looking up predictions.
-   */
-  std::shared_ptr<std::unordered_map<DMatrix*, PredictionCacheEntry>> cache_;
-
-  std::unordered_map<DMatrix*, PredictionCacheEntry>::iterator FindCache(DMatrix const* dmat) {
-    auto cache_emtry = std::find_if(
-        cache_->begin(), cache_->end(),
-        [dmat](std::pair<DMatrix *, PredictionCacheEntry const &> const &kv) {
-          return kv.second.data.get() == dmat;
-        });
-    return cache_emtry;
-  }
 
  public:
-  Predictor(GenericParameter const* generic_param,
-            std::shared_ptr<std::unordered_map<DMatrix*, PredictionCacheEntry>> cache) :
-      generic_param_{generic_param}, cache_{cache} {}
+  explicit Predictor(GenericParameter const* generic_param) :
+      generic_param_{generic_param} {}
   virtual ~Predictor() = default;
 
   /**
@@ -91,12 +127,11 @@ class Predictor {
    * \param           model       The model to predict from.
    * \param           tree_begin  The tree begin index.
    * \param           ntree_limit (Optional) The ntree limit. 0 means do not
-   * limit trees.
+   *                              limit trees.
    */
-
-  virtual void PredictBatch(DMatrix* dmat, HostDeviceVector<bst_float>* out_preds,
+  virtual void PredictBatch(DMatrix* dmat, PredictionCacheEntry* out_preds,
                             const gbm::GBTreeModel& model, int tree_begin,
-                            unsigned ntree_limit = 0) = 0;
+                            uint32_t const ntree_limit = 0) = 0;
 
   /**
    * \fn  virtual void Predictor::UpdatePredictionCache( const gbm::GBTreeModel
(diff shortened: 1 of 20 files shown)


trivialfis:
Requires some more thoughts to make a clean design for the cache.

trivialfis:
https://github.com/dmlc/xgboost/issues/4482

trivialfis:
@hcho3 We talked about adding a custom segfault handler like: https://github.com/bombela/backward-cpp 

I added a simple one in this PR for debugging on Jenkins (and it was helpful).  But it breaks JVM package's error handling/recovery logic.

include/xgboost/predictor.h
```diff
@@ -32,47 +32,83 @@ namespace xgboost {
...
+   *        DMatrix won't renew it.
+   *
+   *  Passing in a `shared_ptr` is critical here.  First to create a `weak_ptr` inside the
+   *  entry this shared pointer is necessary.  More importantly, the life time the this
+   *  cache is tied to the shared pointer.
+   *
+   *  Another way to make a safe cache is create a proxy to this entry, with anther shared
+   *  pointer defined inside, and pass this proxy around instead of the real entry.  But
+   *  seems to be too messy.  In XGBoost, functions like `UpdateOneIter` will have
+   *  (memory) safe acceess to the DMatrix as long as it's passed in as a `shared_ptr`.
```
trivialfis: Note to myself: `life time the` -> `life time of`, `acceess` -> `access`.

include/xgboost/predictor.h
```diff
@@ -32,47 +32,83 @@ namespace xgboost {
...
+   *               GenericParameter::kCpuId for CPU or positive integer for GPU id.
+   *
+   * \return the cache entry for passed in DMatrix, either an existing cache or newly
+   *         created.
+   */
+  PredictionCacheEntry& Cache(std::shared_ptr<DMatrix> m, int32_t device);
+  /* \brief Get a prediction cache entry.  This entry must be already allocated by `Cache`
+   *        method.  Otherwise a dmlc::Error is thrown.
+   *
+   * \param m pointer the DMatrix.
```
trivialfis: To the

src/learner.cc
```diff
@@ -187,9 +190,13 @@ void GenericParameter::ConfigureGpuId(bool require_gpu) {
...
 class LearnerImpl : public Learner {
  public:
-  explicit LearnerImpl(std::vector<std::shared_ptr<DMatrix> >  cache)
-      : need_configuration_{true}, cache_(std::move(cache)) {
+  explicit LearnerImpl(std::vector<std::shared_ptr<DMatrix> > cache)
+      : need_configuration_{true},
+        cache_{new PredictionContainer} {
     monitor_.Init("Learner");
+    for (std::shared_ptr<DMatrix> const& d : cache) {
+      cache_->Cache(d, GenericParameter::kCpuId);
```
RAMitchell: So the shared pointer doesn't get copied here?

src/learner.cc
```diff
@@ -741,13 +748,19 @@ class LearnerImpl : public Learner {
...
-      this->PredictRaw(data_sets[i], &preds_[dmat], false);
-      obj_->EvalTransform(&preds_[dmat]);
+      std::shared_ptr<DMatrix> dmat = data_sets[i];
+      auto& predt = this->cache_->Cache(dmat, generic_parameters_.gpu_id);
+      this->ValidateDMatrix(dmat.get());
+      this->PredictRaw(data_sets[i].get(), &predt, false);
+
+      predictions_.SetDevice(generic_parameters_.gpu_id);
+      predictions_.Resize(predt.predictions.Size());
+      predictions_.Copy(predt.predictions);
```
RAMitchell: Is there an extra copy here that wasn't there before?

src/learner.cc
```diff
@@ -816,14 +829,18 @@ class LearnerImpl : public Learner {
...
     } else if (pred_interactions) {
-      gbm_->PredictInteractionContributions(data, &out_preds->HostVector(), ntree_limit,
+      gbm_->PredictInteractionContributions(data.get(), &out_preds->HostVector(), ntree_limit,
                                             approx_contribs);
     } else if (pred_leaf) {
-      gbm_->PredictLeaf(data, &out_preds->HostVector(), ntree_limit);
+      gbm_->PredictLeaf(data.get(), &out_preds->HostVector(), ntree_limit);
     } else {
-      this->PredictRaw(data, out_preds, training, ntree_limit);
+      auto* predts = &cache_->Cache(data, generic_parameters_.gpu_id);
```
RAMitchell: What is `predts`? More descriptive name maybe.

src/learner.cc
```diff
@@ -816,14 +829,18 @@ class LearnerImpl : public Learner {
...
+      this->PredictRaw(data.get(), predts, training, ntree_limit);
+      out_preds->SetDevice(generic_parameters_.gpu_id);
+      out_preds->Resize(predts->predictions.Size());
+      out_preds->Copy(predts->predictions);
```
RAMitchell: Again, is there extra copying happening here?

src/learner.cc
```diff
@@ -965,22 +983,24 @@ class LearnerImpl : public Learner {
...
   HostDeviceVector<GradientPair> gpair_;
   bool need_configuration_;
 
  private:
   /*! \brief random number transformation seed. */
   static int32_t constexpr kRandSeedMagic = 127;
-  // internal cached dmatrix
-  std::vector<std::shared_ptr<DMatrix> > cache_;
+  // internal cached dmatrix for prediction.
+  std::unique_ptr< PredictionContainer > cache_;
```
RAMitchell: Why is this inside a unique pointer?

src/tree/updater_prune.cc
```diff
@@ -86,9 +87,6 @@ class TreePruner: public TreeUpdater {
         npruned = this->TryPruneLeaf(tree, nid, tree.GetDepth(nid), npruned);
       }
     }
-    LOG(INFO) << "tree pruning end, "
```
RAMitchell: Why is this removed?

trivialfis:
Wil revert some changes, I sometimes added formatting changes because clangd is warning with clang-tidy checks and it's quite difficult to focus.

src/learner.cc
trivialfis: @RAMitchell We don't keep the shared pointer in booster after this PR.  This way the DMatrix object can be destructed without first deleting the booster.

src/learner.cc
trivialfis: Will replace it with a normal variable.

trivialfis:
> Changing the DMatrix pointer arguments to shared pointers could potentially be extracted.

This might not be possible as we need the shared pointer for constructing `std::weak_ptr`.

RAMitchell:
You could do the shared pointers first. This way we get past some harmless changes before we look at the core logic.

src/learner.cc
trivialfis: @RAMitchell It's a copied moved from `Predictor::PredictFromCache`.  I removed this function altogether as now the cache is managed by `Learner`, hence the copying is also done by `Learner`.  It's 1 copy less as I also removed the temporary vector in C API.

src/learner.cc
trivialfis: Renamed.  But seems to be a convention in XGBoost.

src/predictor/cpu_predictor.cc
```diff
@@ -156,60 +135,78 @@ class CPUPredictor : public Predictor {
...
   void UpdatePredictionCache(
       const gbm::GBTreeModel& model,
       std::vector<std::unique_ptr<TreeUpdater>>* updaters,
-      int num_new_trees) override {
+      int num_new_trees,
+      DMatrix* m,
+      PredictionCacheEntry* predts) override {
     int old_ntree = model.trees.size() - num_new_trees;
     // update cache entry
-    for (auto& kv : (*cache_)) {
```
trivialfis: It might be worth noting, before this function update the cache for all DMatrix, including validation datasets.  Now it only update the training DMatrix, as I don't think the `updater` can/should have information about validation datasets in the future (otherwise that would be peeking and violates the purpose of validation).

trivialfis:
Not sure why, putting `HostDeviceVector` into thread local store crashes Python on Windows at the end of execution.  @RAMitchell Have you ever experienced this?

RAMitchell:
If you use device vectors as static variables they can get destructed after the cuda context has been released, leading to errors.

trivialfis:
This might make #5207 difficult to implement...

trivialfis:
@RAMitchell Anyway problem for another day.   I cleaned up the PR per your request.  Formatting changes are reverted.  But the change in linear is kept, as gbm no longer has access to global prediction cache, the code removal for linear is necessary.

src/predictor/cpu_predictor.cc
```diff
@@ -156,60 +135,78 @@ class CPUPredictor : public Predictor {
...
     }
-    if (cache_entry->second.predictions.Size() == 0) {
-      // See comment in GPUPredictor::PredictBatch.
-      InitOutPredictions(cache_entry->second.data->Info(),
-                         &(cache_entry->second.predictions), model);
-      cache_entry->second.predictions.Copy(*out_preds);
+    uint32_t const beg_version = predts->version;
+    CHECK_LE(beg_version, end_version);
+
+    if (beg_version < end_version) {
```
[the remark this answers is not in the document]
trivialfis: Yes.  So there's no need for predict from cache.

trivialfis:
@RAMitchell 

> If I use two differently sized evaluation matrices, the predictions vector will get resized twice every single iteration. This is a very common case.

Before this PR, Learner stores another prediction vector for each DMatrix (hence 2 prediction vectors for each DMatrix, one for cache one for temporary), I can restore that to avoid the resize.  Another way around is I can modify the input to metric into a `span` and resize the prediction vector when it doesn't have sufficient space for current DMatrix.

trivialfis:
I don't have a better name for `version`.  If you are boosting forest with 4 trees per-forest, the updated version each round is 4.  This is non-ideal, but right now there's not much we can do.

As another problem regarding refresh updater I mentioned offline with @RAMitchell , it should be fine as training continuation requires at least 1 serialization.  The cache is cleared.

mli:
# [Codecov](https://codecov.io/gh/dmlc/xgboost/pull/5220?src=pr&el=h1) Report
> Merging [#5220](https://codecov.io/gh/dmlc/xgboost/pull/5220?src=pr&el=desc) into [master](https://codecov.io/gh/dmlc/xgboost/commit/29eeea709aff275fa97187858906fce711b4ae34&el=desc) will **not change** coverage by `%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/dmlc/xgboost/pull/5220?src=pr&el=tree)

```diff
@@           Coverage Diff           @@
##           master    #5220   +/-   ##
=======================================
  Coverage   83.76%   83.76%           
=======================================
  Files          11       11           
  Lines        2409     2409           
=======================================
  Hits         2018     2018           
  Misses        391      391           
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/dmlc/xgboost/pull/5220?src=pr&el=continue).
> Powered by [Codecov](https://codecov.io/gh/dmlc/xgboost/pull/5220?src=pr&el=footer). Last update [29eeea7...4fcc43e](https://codecov.io/gh/dmlc/xgboost/pull/5220?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

trivialfis:
The copying is annoying.  I don't have any thoughts on to get rid of them yet.

PR #5222, opened 2020-01-22
Config for linear updaters.

trivialfis:


trivialfis:
@RAMitchell The valuation result seems to have slight different between numpy input and cudf input.

PR #5228, opened 2020-01-24
Cleanup quantile.h

RAMitchell:
I found it difficult to understand what is going on in quantile.h, so I went through and removed dead code, tried to simplify some relationships between classes and tried to make as many methods/variables private as possible. This PR should cause no change in behaviour.

In particular:
- WQSummary is removed as everything seems to use WXQSummary
- QuantileSketchTemplate has been collapsed into WXQuantileSketch as it was the only instance being used

Ideally I wanted something like an interface where I could hide away the internals but still use it in a sketching algorithm. This can't happen yet because this part https://github.com/dmlc/xgboost/blob/44469a0ca9375f3f44bc9b827a38e0eeed37d475/src/tree/updater_basemaker-inl.h#L366 
accesses internal members in many places.

### DIFF
--- a/src/common/hist_util.cc
+++ b/src/common/hist_util.cc
@@ -79,18 +79,17 @@ void SparseCuts::SingleThreadBuild(SparsePage const& page, MetaInfo const& info,
   p_cuts_->min_vals_.resize(end_col - beg_col, 0);
 
   for (uint32_t col_id = beg_col; col_id < page.Size() && col_id < end_col; ++col_id) {
-    // Using a local variable makes things easier, but at the cost of memory trashing.
-    WXQSketch sketch;
     common::Span<xgboost::Entry const> const column = page[col_id];
     uint32_t const n_bins = std::min(static_cast<uint32_t>(column.size()),
                                      max_num_bins);
+    // Using a local variable makes things easier, but at the cost of memory trashing.
+    WXQSketch sketch(info.num_row_, 1.0 / (n_bins * kFactor));
     if (n_bins == 0) {
       // cut_ptrs_ is initialized with a zero, so there's always an element at the back
       p_cuts_->cut_ptrs_.emplace_back(p_cuts_->cut_ptrs_.back());
       continue;
     }
 
-    sketch.Init(info.num_row_, 1.0 / (n_bins * kFactor));
     for (auto const& entry : column) {
       uint32_t weight_ind = 0;
       if (use_group_ind) {
@@ -241,10 +240,7 @@ void DenseCuts::Build(DMatrix* p_fmat, uint32_t max_num_bins) {
   unsigned const nstep =
       static_cast<unsigned>((info.num_col_ + nthread - 1) / nthread);
   unsigned const ncol = static_cast<unsigned>(info.num_col_);
-  sketchs.resize(info.num_col_);
-  for (auto& s : sketchs) {
-    s.Init(info.num_row_, 1.0 / (max_num_bins * kFactor));
-  }
+  sketchs.resize(info.num_col_, WXQSketch(info.num_row_, 1.0 / (max_num_bins * kFactor)));
 
   // Data groups, used in ranking.
   std::vector<bst_uint> const& group_ptr = info.group_ptr_;
--- a/src/common/hist_util.cc
+++ b/src/common/hist_util.cc
@@ -233,14 +233,13 @@ void DenseCuts::Build(DMatrix* p_fmat, uint32_t max_num_bins) {
 
   // safe factor for better accuracy
   constexpr int kFactor = 8;
-  std::vector<WXQSketch> sketchs;
 
   const int nthread = omp_get_max_threads();
 
-  unsigned const nstep =
-      static_cast<unsigned>((info.num_col_ + nthread - 1) / nthread);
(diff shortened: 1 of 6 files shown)


trivialfis:
Plan to add unit tests?

src/common/quantile.h
```diff
@@ -9,21 +9,23 @@
...
 #include <cstring>
-#include <algorithm>
 #include <iostream>
+#include <vector>
 
 namespace xgboost {
 namespace common {
+
 /*!
  * \brief experimental wsummary
```
trivialfis: We don't call it experimental anymore.

src/common/hist_util.cc
```diff
@@ -234,17 +233,13 @@ void DenseCuts::Build(DMatrix* p_fmat, uint32_t max_num_bins) {
...
 
-  unsigned const nstep =
-      static_cast<unsigned>((info.num_col_ + nthread - 1) / nthread);
-  unsigned const ncol = static_cast<unsigned>(info.num_col_);
-  sketchs.resize(info.num_col_);
-  for (auto& s : sketchs) {
-    s.Init(info.num_row_, 1.0 / (max_num_bins * kFactor));
-  }
+  unsigned const nstep = static_cast<unsigned>((info.num_col_ + nthread - 1) /
+  nthread); unsigned const ncol = static_cast<unsigned>(info.num_col_);
```
trivialfis: That's a weird way to format code.

src/common/quantile.h
```diff
@@ -418,18 +204,21 @@ struct WXQSummary : public WQSummary<DType, RType> {
           size_t i = bid;
           RType maxdx2 = src.data[end].RMaxPrev() * 2;
           for (; k < n; ++k) {
-            RType dx2 =  2 * ((k * mrange) / n + begin);
+            RType dx2 = 2 * ((k * mrange) / n + begin);
             if (dx2 >= maxdx2) break;
             while (i < end &&
-                   dx2 >= src.data[i + 1].rmax + src.data[i + 1].rmin) ++i;
+                   dx2 >= src.data[i + 1].rmax + src.data[i + 1].rmin)
+              ++i;
```
trivialfis: Add a `{}`.

src/common/quantile.h
```diff
@@ -686,19 +447,18 @@ class QuantileSketchTemplate {
    * \param maxn maximum number of data points can be feed into sketch
    * \param eps accuracy level of summary
    */
-  inline void Init(size_t maxn, double eps) {
```
trivialfis: Great that we can remove the `Init` method.

tests/cpp/data/test_quantile.cc
```diff
@@ -0,0 +1,40 @@
+// Copyright by Contributors
+
+#include <gtest/gtest.h>
+#include "../../../src/common/quantile.h"
+
+namespace xgboost {
+namespace common {
+TEST(Quantile, Basic) {
```
trivialfis: Do we need a null test?

src/common/quantile.h
```diff
@@ -444,150 +233,124 @@ struct WXQSummary : public WQSummary<DType, RType> {
...
+        ++b;
       } while (b != b_end);
     }
-    CHECK(dst == data + size) << "bug in combine";
+    this->size = dst - data;
+    const RType tol = 10;
+    RType err_mingap, err_maxgap, err_wgap;
+    this->FixError(&err_mingap, &err_maxgap, &err_wgap);
+    if (err_mingap > tol || err_maxgap > tol || err_wgap > tol) {
+      LOG(INFO) << "mingap=" << err_mingap << ", maxgap=" << err_maxgap
```
trivialfis: Is it necessary to log this?

src/common/quantile.h
```diff
@@ -444,150 +233,124 @@ struct WXQSummary : public WQSummary<DType, RType> {
...
+        *dst = Entry(b->rmin + aprev_rmin, b->rmax + a->RMaxPrev(), b->wmin,
+                     b->value);
+        bprev_rmin = b->RMinNext();
+        ++dst;
+        ++b;
       }
     }
     if (a != a_end) {
-      RType bprev_rmax = (b_end - 1)->rmax;
+      RType brmax = (b_end - 1)->rmax;
```
trivialfis: Please change it to a more verbose name.

src/common/quantile.h
```diff
@@ -408,7 +193,8 @@ struct WXQSummary : public WQSummary<DType, RType> {
     }
```
trivialfis: > CHECK(nbig < n) << "quantile: too many large chunk";

I recently run into this error when weight is sampled from a Gaussian distribution.  Not sure what's happening here.

src/common/hist_util.cc
RAMitchell: I just auto formatted with clang-format. Will clean it up.

src/common/quantile.h
RAMitchell: I actually have no idea what any of this code does, it's just copy pasted. I will spend a bit more time learning the algorithms to see if I can clean it up more.

trivialfis:
Recently I have been learning boost accumulator library.   We might want to borrow some concept from there.

PR #5229, opened 2020-01-24
Make some GPU tests deterministic

RAMitchell:
We observed an intermittent test failure, with a small difference and accuracy between two subsequent training calls.

This PR makes random numbers consistent, prevents weights from getting close to 0 and enforces a single thread.


trivialfis:
Oops. :-)

Issue #5117, opened 2019-12-14
error: no matching conversion for functional-style cast from 'std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000> >::rep' (aka 'long long') to 'xgboost::Integer' (aka 'xgboost::JsonInteger')


yurivict:
```
/usr/ports/misc/xgboost/work/xgboost-0.90-235-gad4a1c73/src/common/timer.cc:47:25: error: no matching conversion for functional-style cast from 'std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000> >::rep' (aka 'long long') to 'xgboost::Integer' (aka 'xgboost::JsonInteger')
    j_pair["elapsed"] = Integer(std::chrono::duration_cast<std::chrono::microseconds>(
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/ports/misc/xgboost/work/xgboost-0.90-235-gad4a1c73/include/xgboost/json.h:189:7: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000> >::rep' (aka 'long long') to 'const xgboost::JsonInteger' for 1st argument
class JsonInteger : public Value {
      ^
/usr/ports/misc/xgboost/work/xgboost-0.90-235-gad4a1c73/include/xgboost/json.h:189:7: note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000> >::rep' (aka 'long long') to 'xgboost::JsonInteger' for 1st argument
/usr/ports/misc/xgboost/work/xgboost-0.90-235-gad4a1c73/include/xgboost/json.h:200:3: note: candidate template ignored: requirement 'std::is_same<long long, long>::value' was not satisfied [with IntT = long long]
  JsonInteger(IntT value) : Value(ValueKind::Integer), integer_{value} {} // NOLINT
  ^
```

clang-8 on FreeBSD

hcho3:
Thanks for your report. We generally aim to make XGBoost POSIX-compliant, but currently our testing infrastructure uses Linux exclusively.

trivialfis:
@yurivict Would you like to open a PR as you have better testing environment for BSD?  A simple static_cast should get it working.

godasunil:
[...]

PR #5233, opened 2020-01-27
Support FreeBSD

hcho3:
* `exp10` is a GNU extension and is missing on FreeBSD. Add `__FreeBSD__` guard to disable use of `exp10`. In the long run, we may want to add a feature testing for `exp10` to the build system.
* `std::chrono::duration_cast<...>.count()` implementation returns `long long` type, which `JsonInteger` does not accept. Explicitly cast the result into `int64_t`.

Depends on https://github.com/dmlc/rabit/pull/133.

Closes https://github.com/dmlc/xgboost/issues/5117.

### DIFF
--- a/src/common/json.cc
+++ b/src/common/json.cc
@@ -69,7 +69,7 @@ void JsonWriter::Visit(JsonNull const* null) {
   this->Write("null");
 }
 
-void JsonWriter::Visit(JsonString const* str) {
+fvoid JsonWriter::Visit(JsonString const* str) {
   std::string buffer;
   buffer += '"';
   auto const& string = str->getString();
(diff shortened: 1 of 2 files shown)


trivialfis:
>  In the long run, we may want to add a feature testing for exp10 to the build system.

I have an implementation using lookup table.  Will start splitting up the optimized JSON PR after this release.

hcho3:
> I have an implementation using lookup table. Will start splitting up the optimized JSON PR after this release.

Awesome!

hcho3:
For now, I'm leaving out the rabit update from this PR. I will file a separate PR for updating rabit.

trivialfis:
There are a handful of BSDs out there, we may want to use `__linux__` instead.

Issue #5235, opened 2020-01-27
Model compatibility with version 0.9 seems broken


sh1ng:
Loading an old model is not possible anymore 
```
Traceback (most recent call last):
  File "tests/python/open_data/gbm/test_gpu_prediction_pickledmodel.py", line 270, in <module>
    TestGPUPredict().test_predict_sklearn_frompickle(None)
  File "tests/python/open_data/gbm/test_gpu_prediction_pickledmodel.py", line 246, in test_predict_sklearn_frompickle
    model = load_obj("./tests/python/open_data/gbm/model_saved.pkl")
  File "tests/python/open_data/gbm/test_gpu_prediction_pickledmodel.py", line 32, in load_obj
    return pickle.load(f)
  File "/home/sh1ng/dev/.venv/lib/python3.6/site-packages/xgboost/core.py", line 1096, in __setstate__
    _LIB.XGBoosterUnserializeFromBuffer(handle, ptr, length))
  File "/home/sh1ng/dev/.venv/lib/python3.6/site-packages/xgboost/core.py", line 189, in _check_call
    raise XGBoostError(py_str(_LIB.XGBGetLastError()))
xgboost.core.XGBoostError: [15:21:38] /home/sh1ng/dev/h2o4gpu/xgboost/src/learner.cc:677: Check failed: header == serialisation_header_ (

```
https://github.com/h2oai/h2o4gpu/blob/master/tests/python/open_data/gbm/model_saved.pkl

it's get loaded with
```
    with open(name, 'rb') as f:
        return pickle.load(f)
```

Is it intentional? What's recommended way to migrate to a new version?

trivialfis:
Yes, it's intentional.  I wrote a tutorial explaining the difference between pickle and `bst.save_model`: https://xgboost.readthedocs.io/en/latest/tutorials/saving_model.html  It applies to binary (what you are using currently) pickle too.

The key idea is, `bst.save_model` is stable across different versions, but not pickle, as the later captures all internal parameters.  If your previous model is saved as pickle, you will have to load it back with same version of XGBoost, then call `save_model` to export the model (trees) for later version of XGBoost.  This creates some inconveniences,  suggestions are welcomed.

sh1ng:
Current api recommends to use pickle . [**If you are using only the Python interface, we recommend pickling the model object for best results.** ](https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBRegressor.load_model)

[...]

PR #5236, opened 2020-01-27
Better error message about loading pickled model.

trivialfis:
Related https://github.com/dmlc/xgboost/issues/5235 .


trivialfis:
Running some more tests.

Issue #5090, opened 2019-12-04
[Feature Request] Add score_tree_interval during early stopping


kylejn27:
## Request:

Add a `score_tree_interval` option so that when you're building with really large data, the model doesn't eval on each tree. 

Similar to this:
http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/algo-params/score_tree_interval.html

## Purpose:

With large data, scoring every iteration on the validation set is extremely costly. Currently with `early_stopping_rounds` the behavior is to score on every early_stopping_round round. 
https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBRegressor.fit. 

It would be nice to be able to space out the width of the early stopping rounds, as H2O does, see H2O's stopping_rounds parameter for example. You can tell H2O to score every 20 trees and if the model hasn't improved in 5 scoring iterations (e.g., 100 trees), then XGBoost would stop training.

It would also be nice if you could also enable different early stopping at different points in the tree. For example, suppose you wanted to not score on the eval_set until the 1000th tree, and then score on every tree. This would help make the training time more efficient if you knew before hand (say, from prior modeling runs) how long the tree took before converging.

I'm primarily concerned with the python implementation of this library, I don't _think_ this has been implemented elsewhere already

trivialfis:
Em.  I wanted to implement stage_predict for XGBoost.

kylejn27:
happy to take a stab at it, though due to certain restrictions I can't PR into this repository for a while but I can collaborate here in the meantime

trivialfis:
Sorry for the late reply.  I will take a look into this option today.

kylejn27:
no problem, not super high priority for me. Mostly interested in getting familiar with the internals of this library and I think attempting to implement this kind of thing would teach me a lot

trivialfis:
Give it a go, and feel free to reach me if you need any help.  One personal advice, be careful of the prediction cache, it bites.

kylejn27:
great, thanks!

kylejn27:
@trivialfis Just getting some time to start looking at this today. I think there was a misunderstanding on what I'd be implementing. I was intending on only scoring certain iterations of the model during model fit to improve model fit performance (speed not accuracy). staged_predict is on the prediction side. 

I don't know of any parameter following the scikit learn implementation that enables this kind of thing so it would have to be an xgboost unique parameter

[...]

PR #5238, opened 2020-01-27
Add score tree interval

kylejn27:
**What**

Adds `eval_start` to control the start iteration for model evaluation during model training
Adds `eval_interval` to control the interval that model evaluation occurs during model training

**Why**

Implements https://github.com/dmlc/xgboost/issues/5090


python-package/xgboost/sklearn.py
```diff
@@ -1153,7 +1172,9 @@ def _dmat_init(group, **params):
                               evals=evals,
                               evals_result=evals_result, feval=feval,
                               verbose_eval=verbose, xgb_model=xgb_model,
-                              callbacks=callbacks)
+                              callbacks=callbacks,
```
trivialfis: Thanks for submitting a PR for this!  I like the idea of selecting range of iterations for evaluation.  But I really hope that we can implement it as a simple callback instead of parameters + callback.  Might need some refactoring.

kylejn27: Thanks for the feedback! I think that's a good idea, I'll see what I can do

kylejn27: @trivialfis Finally have some time to get back to this. 

I agree with you that it would be nice for this to be implemented as only a callback.  Its entirely possible that I'm missing something obvious but I'm not sure if it can be without making significant changes to the code. 

The main functionality added from this PR is the ability to skip the [bst.eval_set](https://github.com/dmlc/xgboost/blob/08dada08251e71ebd4e34a9ee788617f0a4c16e4/python-package/xgboost/training.py#L86-L93) method if the current iteration doesn't occur within the specified range (`score_tree_interval`). 

I suppose modifying `_train_internal` to  run [bst.eval_set](https://github.com/dmlc/xgboost/blob/08dada08251e71ebd4e34a9ee788617f0a4c16e4/python-package/xgboost/training.py#L86-L93) inside of a callback might work. This new score callback could be initialized with `eval_start` and `eval_interval` defaulting to 0 and 1 respectively. If the user wishes to change the defaults the user could manually add this callback with different initialization parameters. The only problem with this, would be that the state of `evaluation_result_list` would have to to be passed around between the different callbacks and the order that the callbacks are executed in would matter (this new callback would have to go first)

Happy to chat about this further, perhaps moving where the model evaluation occurs is the best solution

trivialfis: @kylejn27  Let me take a deeper look.

trivialfis:
New callback is landed, I will revisit this feature.

PR #5232, opened 2020-01-26
[R] Robust endian detection in CRAN xgboost build

hcho3:
CMake doesn't appear to be a well accepted alternative in the R ecosystem, in particular CRAN. Without CMake, detecting endianness has been a hairy endeavor:
```cpp
 #if defined(__APPLE__) || defined(_WIN32)
    #define DMLC_LITTLE_ENDIAN 1
  #elif defined(__GLIBC__) || defined(__GNU_LIBRARY__) \
        || defined(__ANDROID__) || defined(__RISCV__)
    #include <endian.h>
    #define DMLC_LITTLE_ENDIAN (__BYTE_ORDER == __LITTLE_ENDIAN)
  #elif defined(__FreeBSD__) || defined(__OpenBSD__)
    #include <sys/endian.h>
    #define DMLC_LITTLE_ENDIAN (_BYTE_ORDER == _LITTLE_ENDIAN)
  #elif defined(__EMSCRIPTEN__) || defined(__hexagon__)
    #define DMLC_LITTLE_ENDIAN 1
  #elif defined(__sun) || defined(sun)
    #include <sys/isa_defs.h>
    #if defined(_LITTLE_ENDIAN)
      #define DMLC_LITTLE_ENDIAN 1
    #else
      #define DMLC_LITTLE_ENDIAN 0
    #endif
  #else
    #error "Unable to determine endianness of your machine; use CMake to compile"
  #endif
```
(https://github.com/dmlc/dmlc-core/blob/cce75c1f1eb6153d79e2c244374de9dd5d564e2e/include/dmlc/endian.h#L14-L36)

#5131 revealed the inadequacy of the macro cascades. So in this PR, I used Autoconf (already used by R) to detect the endianness of the system at build time. **Now we can detect endianness robustly when we install XGBoost from CRAN (i.e. `install.packages('xgboost')`**.

In `configure.ac`, I added the macro `AC_RUN_IFELSE()` in order to compile and run the following C program:
```c
#include <stdint.h>

int main(void) {
  const uint16_t endianness = 256;
  return !!(*(const uint8_t *)&endianness);
    // returns 0 for little endian
    // returns 1 for big endian
}
```
Depending on the return value of the program, we set the compilation symbol `DMLC_CMAKE_LITTLE_ENDIAN` to either 1 (little endian) and 0 (big endian).

I tested this change as follows:
* I tried running `R CMD INSTALL .` both with Ubuntu and Alpine Linux.
[...]

PR #5240, opened 2020-01-28
[R] Enable OpenMP with AppleClang in XGBoost R package

hcho3:
Follow-up to #5232. This PR uses Autoconf to detect the availability of OpenMP library (`libomp`) on Mac OSX. This is done by compiling the following program with command `${CC} -o conftest conftest.c /usr/local/lib/libomp.dylib -Xclang -fopenmp`:

```c
/* conftest.c */
#include <omp.h>

int main(void) {
  return (omp_get_max_threads() <= 1);
    // returns 0 when there's only one thread
    // returns 1 when there's more than one thread
}
```

**Implication: Now Mac users can simply run `install.packages('xgboost')` and enjoy the performance of multiple threads,** as long as they first install `libomp` package from Homebrew. 

I tested this change as follows:
* I ran `R CMD INSTALL .` on my Macbook Pro. A sample XGBoost training job runs with all 4 threads:
```r
library(xgboost)
train <- agaricus.train
test <- agaricus.test
bst <- xgboost(data = train$data, label = train$label, max.depth = 1, eta = 1,
               nthread = 4, nrounds = 1000, objective = "binary:logistic", verbose = 1,
               eval.metric = 'auc')
```
Screen Shot 2020-01-28 at 12 34 11 AM

* Then I uninstalled `libomp` Homebrew and ran `R CMD INSTALL .` again. A prominent warning shows up, as I intended:
```
checking whether OpenMP will work in a package... no
*****************************************************************************************
WARNING: OpenMP is unavailable on this Mac OSX system. Training speed may be suboptimal.
         To use all CPU cores for training jobs, you should install OpenMP by running

             brew install libomp
*****************************************************************************************
```
Aside from the warning, XGBoost compiles successfully. The sample job ran without a problem, albeit this time with a single thread.

Note to reviewer: you should only review `configure.ac`. The `configure` script is generated from `configure.ac` when `autoconf` is run.

### DIFF
--- a/R-package/configure.ac
+++ b/R-package/configure.ac
@@ -4,6 +4,9 @@ AC_PREREQ(2.62)
 
 AC_INIT([xgboost],[0.6-3],[],[xgboost],[])
 
+# Use this line to set CC variable to a C compiler
+AC_PROG_CC
+
 ### Check whether backtrace() is part of libc or the external lib libexecinfo
 AC_MSG_CHECKING([Backtrace lib])
 AC_MSG_RESULT([])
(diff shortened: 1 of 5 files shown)


R-package/configure.ac
```diff
@@ -4,6 +4,9 @@ AC_PREREQ(2.62)
 
 AC_INIT([xgboost],[0.6-3],[],[xgboost],[])
 
+# Use this line to set CC variable to a C compiler
+AC_PROG_CC
```
trivialfis: ?  Why ?

hcho3: The feature testing script needs access to `CC` variable.

trivialfis: @hcho3 Yup, but why setting it to a `C` complier?  I assume that means changing `g++` to `gcc` or `clang++` to `clang`.

hcho3: `CC` refers to a C compiler and `CXX` refers to a C++ compiler.

trivialfis: Yup.   I thought we use `CXX` var exclusively so are compilation tests ... Never mind,  I don't know anything about autotools.  Sorry for the noise.

hcho3: Don't worry, I don't know much about autotools either. I wait for the day where we can use CMake for CRAN submissions.

PR #5146, opened 2019-12-22
Enable OpenMP with Apple Clang (Mac default compiler)

hcho3:
Simplify logic for importing OpenMP into the CMake build by treating OpenMP as a target. We no longer have to set compiler flags; all it takes is to import target `OpenMP::OpenMP_CXX` via `target_link_libraries()`. See https://cliutils.gitlab.io/modern-cmake/chapters/packages/OpenMP.html for more information.

This will be especially useful for Mac OSX users. The old way requires them to use Homebrew GCC, which is a heavy dependency (*). The new way allows them to use Apple Clang (default system compiler) instead, provided that they install OpenMP library via Homebrew (`brew install libomp`). Furthermore, we can release XGBoost 1.0.0 on Homebrew with OpenMP enabled.

CMake version is raised to ~3.9~ 3.12 to access `OpenMP::OpenMP_CXX` target. 3.12 is needed to use `target_link_libraries()` with object target `objxgboost`. Note that we've been already requiring 3.12 for the GPU algorithms.

Closes #4477.
Depends on https://github.com/dmlc/dmlc-core/pull/586

(*) See discussion at Homebrew/homebrew-core#43246, where Homebrew maintainers asked to remove GCC dependency from XGBoost.

PR #5241, opened 2020-01-28
Make `pip install xgboost*.tar.gz` work by fixing build-python.sh

hcho3:
This PR makes a few fixes so that users can install Python XGBoost from a source tarball `xgboost-1.0.0-SNAPSHOT.tar.gz`. This is important for users of Mac OSX, FreeBSD, and other platforms where we do not offer pre-built wheels.

Note that #5146 enabled the use of AppleClang (default compiler on OSX), so we no longer require Homebrew GCC. Thus, we can simplify the installation doc considerably.

How to test:
```bash
# Pack XGBoost source file into a tarball
make pippack 
# Install XGBoost source distribution
pip3 install xgboost-1.0.0-SNAPSHOT.tar.gz --user
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,19 +3,32 @@ sudo: required
 
 # Enabling test OS X
 os:
+  - linux
   - osx
 
 osx_image: xcode10.3
+dist: bionic
 
 # Use Build Matrix to do lint and build seperately
 env:
   matrix:
     # python package test
     - TASK=python_test
+    # test installation of Python source distribution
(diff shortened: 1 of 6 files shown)


hcho3:
I added a test in Travis.

hcho3:
The sdist test keep failing for the Linux target, but I cannot reproduce it on my Linux machine.

hcho3:
I will start a new branch for 1.0.0 release after this PR is merged.

PR #5245, opened 2020-01-29
Save Scikit-Learn attributes into learner attributes.

trivialfis:
This PR modifies the behavior of saving Scikit-Learn model, pickle is no longer
recommended.  ~Instead an additional JSON file is produced for saving
Scikit-Learn specific attributes.  When using JSON output format in XGBoost,
the additional file is subsumed into main model file.~

Instead the attributes are saved into `Learner::attributes_`.


hcho3:
Is this blocking?

trivialfis:
@hcho3 I hope so, but will value your opinion.  It's gonna be painful for everyone who uses pickle before.  So let's get started sooner as possible.

trivialfis:
@hcho3 @RAMitchell Please help with this PR as we will have to maintain it for a very long time as we promise the stability of output model, which includes this meta file.  If you have suggestions or doubt, please share.

trivialfis:
I saved the additional attributes into `Learner::attributes_` instead of a separated file.  It's a hack, but seems nicer than producing another file or parsing the model multiple times.

hcho3:
Serialization has been a headache for Python objects in sklearn interface, and this is a welcome corrective.

RAMitchell:
I was looking at the older commit, I like your new version much better than the file. The additional attribute is harmless right? I'm ok with this and defer to the two of you.

hcho3:
I'd like to merge it soon, if there's no objection.

PR #5247, opened 2020-01-30
Cleanup warnings.

trivialfis:
From clang-tidy-9 and gcc-7: Invalid case style, narrowing definition, wrong
initialization order, unused variables.

### DIFF
--- a/include/xgboost/model.h
+++ b/include/xgboost/model.h
@@ -15,6 +15,7 @@ namespace xgboost {
 class Json;
 
 struct Model {
+  virtual ~Model() = default;
   /*!
    * \brief load the model from a json object
    * \param in json object where to load the model from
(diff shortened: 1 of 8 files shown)


hcho3:
Why are we removing `num_elements` from adapters?

RAMitchell:
Its unused I think. Any way to make these checks part of CI? That way you won't have to cleanup after me in future :)

hcho3:
@RAMitchell CI already runs clang-tidy, but an older version of it (7):
https://github.com/dmlc/xgboost/blob/ea850ecd209f647f737f8268d2ba527281feb85d/tests/ci_build/Dockerfile.clang_tidy#L10

Maybe we should upgrade clang-tidy to 9?

RAMitchell:
Sounds good!

trivialfis:
It requires adding latest llvm distribution to CI I believe.  Will do that after the release.

PR #5248, opened 2020-01-30
Test model compatibility.

trivialfis:
The total size of uploaded models is 100K, which should not be too much a burden for git.

I also tested 0.90 with binary model (no JSON back then), but didn't upload the models here.

I also uploaded the script used for generating these models, please review. ;-)


hcho3:
I'm quite wary of adding binary files to the git repo. Maybe CI can run the generation script on the fly?

RAMitchell:
+1 to avoiding binary files. Can we upload them somewhere or make use of `pip install xgboost=0.9` to generate them on the fly?

trivialfis:
> There is another very good reason for keeping binary files out of your repository: They are usually much bigger.

That's not true, in 100K files, JSON takes up more than a half.

> Data about changes in binary files just makes the commits impossible to read.

We should not change it very often as it will be breaking.

One way to do it is we checkout a old commit in CI and generate these models.  But that seems to be very slow ...  Another option is rely on pip.

trivialfis:
@hcho3 

> If you'd like, I can add the round-trip test to Jenkins CI

Please.  :-). We can do this after the release.  Currently generated models are tagged as snapshot.

hcho3:
Okay, for now, I want to get the release candidate out, so let's merge this. I will come after the release and have CI generate models on the fly, so that we can remove the binary files from the repo. As long as we remove the binary files right away, the size bloat effect would be minimal.

trivialfis:
@hcho3. If you like we can merge it in the new release branch rather than master branch.

hcho3:
@trivialfis That's even better. Let me create a new branch now.

hcho3:
Changed the base branch of this PR to `release_1.0.0`.

trivialfis:
@hcho3 Merge as you see fit.

Issue #5095, opened 2019-12-06
Threshold effect due to multiple cut point strategies in histogram mode


DoanHuuJacques:
Hi,

looking at AddCutPoint() method, I noticed a heuristic with size threshold (16) which determines the strategy to define cut point values.
Is there a particular reason (except to avoid division operations and to be a bit faster) to not always consider the mid-point strategy even for large cardinality?

Indeed, when I encode my categorical features based on the category frequency, if the cardinality is greater than 16, the last 2 categories (most frequent ones) and the first 2 categories cannot be split. The consequence is that if 2 categories at extreme boundaries are significantly discriminative to improve the model accuracy, the split cannot occur between them.
Such issue is more effective with frequency encoding as most frequent categories are placed at the upper boundary.

Thanks a lot for clarification.
Regards

Jacques

void AddCutPoint(WXQSketch::SummaryContainer const& summary) {
    if (summary.size > 1 && summary.size <= 16) {
      /* specialized code categorial / ordinal data -- use midpoints */
      for (size_t i = 1; i < summary.size; ++i) {
        bst_float cpt = (summary.data[i].value + summary.data[i - 1].value) / 2.0f;
        if (i == 1 || cpt > p_cuts_->cut_values_.back()) {
          p_cuts_->cut_values_.push_back(cpt);
        }
      }
    } else {
      for (size_t i = 2; i < summary.size; ++i) {
        bst_float cpt = summary.data[i - 1].value;
        if (i == 2 || cpt > p_cuts_->cut_values_.back()) {
          p_cuts_->cut_values_.push_back(cpt);
        }
      }
    }
  }

trivialfis:
@hcho3

hcho3:
> Is there a particular reason (except to avoid division operations and to be a bit faster) to not always consider the mid-point strategy even for large cardinality?

The cut point is part of a quantile sketch, so it represents an estimate of quantile value. The midpoint strategy introduces bias by shifting the estimate upward. The shifting is done for categorical or binary features because often a quantile estimate coincides with another integer feature value, so shifting is necessary to discriminate between two adjacent discrete feature values.

The number 16 was chosen as a trade-off: if we enable the midpoint strategy for all cases, we would affect splits for continuous features and thus incur significant bias. And currently we don’t have a good way to distinguish between discrete and continuous feature columns, so hence the heuristic. Perhaps letting the user explicitly list discrete features will address the issue better.

DoanHuuJacques:
[...]

Issue #5096, opened 2019-12-06
[Improvement] Support of discrete vs continuous metadata + AddCutPoint() adaptation


DoanHuuJacques:
Hi,

as per discussed in the close issue thread (https://github.com/dmlc/xgboost/issues/5095), give the possibility to provide per feature whether it's a continuous or discrete.

Indeed, XGBoost learning strategy can be adapted accordingly.
The typical benefit (mentioned in the above issue entry about AddCutPoint() method) is to apply the proper cut point strategy for either discrete or continuous feature, instead of relying on a predefined cardinality threshold (16).

Thanks

RAMitchell:
As of #5251, I think this is resolved such that categorical feature values >16 do not get lost. We can reopen this discussion if there are any further issues with quantile cut points and categorical features.

PR #5251, opened 2020-01-31
Testing hist_util

RAMitchell:
This PR is for adding more tests for quantile generation in hist/gpu_hist. Part of the motivation is to create some absolute tests of accuracy for GPU sketching, instead of comparing against existing CPU algorithms.

So far I have tested the following expected behaviours with respect to a single feature:
- The histogram cut `min_value` should be less than all inputs
- The last histogram cut value should be greater than all inputs
- The rank of quantile cuts from the sketch should not exceed error ~0.01 as compared to rank from sorted values
- And input of k < num_bins unique values should output k unique bins. i.e. it should be possible to split over each unique value

I also expected that number of cuts (excluding the minimum value) should equal the number of bins requested if the unique input size is greater than num_bins. This is true for categorical features with <= unique values but not for continuous features. This currently means that if you ask for 256 bins you get 254 actual bins, which seems like a bug. Two of the samples get lost here (https://github.com/dmlc/xgboost/blob/472ded549d2a335bc807819ddb09349d5831610f/src/common/hist_util.h#L208), one value gets added afterwards to be larger than everything. Another sample seems to get lost here (https://github.com/dmlc/xgboost/blob/472ded549d2a335bc807819ddb09349d5831610f/src/common/hist_util.cc#L319), where 256 samples go into SetPrune and 255 come out, despite max_num_bins being set at 256.

After resolving this I also want to add tests comparing the rank of the output quantiles against the correct rank by sorting the input.

I would also like to clarify the need for specialised cut points for categorical variables, implemented here (https://github.com/dmlc/xgboost/blob/472ded549d2a335bc807819ddb09349d5831610f/src/common/hist_util.h#L200) and discussed in #5095.

@hcho3 any comments would be much appreciated.

### DIFF
--- a/src/common/hist_util.h
+++ b/src/common/hist_util.h
@@ -147,7 +147,7 @@ class HistogramCuts {
 
   size_t TotalBins() const { return cut_ptrs_.back(); }
 
-  BinIdx SearchBin(float value, uint32_t column_id) {
+  BinIdx SearchBin(float value, uint32_t column_id) const {
     auto beg = cut_ptrs_.at(column_id);
     auto end = cut_ptrs_.at(column_id + 1);
     auto it = std::upper_bound(cut_values_.cbegin() + beg, cut_values_.cbegin() + end, value);
@@ -171,7 +171,7 @@ class HistogramCuts {
  */
 class CutsBuilder {
  public:
-  using WXQSketch = common::WXQuantileSketch<bst_float, bst_float>;
+  using WQSketch = common::WQuantileSketch<bst_float, bst_float>;
 
  protected:
   HistogramCuts* p_cuts_;
@@ -195,8 +195,9 @@ class CutsBuilder {
     return group_ind;
   }
 
-  void AddCutPoint(WXQSketch::SummaryContainer const& summary) {
-    for (size_t i = 1; i < summary.size; ++i) {
+  void AddCutPoint(WQSketch::SummaryContainer const& summary, int max_bin) {
+    int required_cuts = std::min(static_cast<int>(summary.size), max_bin);
+    for (size_t i = 1; i < required_cuts; ++i) {
       bst_float cpt = summary.data[i].value;
       if (i == 1 || cpt > p_cuts_->cut_values_.back()) {
         p_cuts_->cut_values_.push_back(cpt);
@@ -240,7 +241,7 @@ class DenseCuts  : public CutsBuilder {
       CutsBuilder(container) {
     monitor_.Init(__FUNCTION__);
   }
-  void Init(std::vector<WXQSketch>* sketchs, uint32_t max_num_bins);
+  void Init(std::vector<WQSketch>* sketchs, uint32_t max_num_bins);
   void Build(DMatrix* p_fmat, uint32_t max_num_bins) override;
 };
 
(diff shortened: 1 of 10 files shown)


RAMitchell:
I added tests for eps accuracy of sketch for combinations of number of bins and input sizes. I also fixed an off by one error where for categorical data the number of histogram bins would always be one less than the unique inputs, so certain categorical values were never used. 

It still seems possible for summary.SetPrune() to unnecessarily remove cuts in some cases, but this is not a major issue and I'm not confident to change that function, so will leave that for another day.

I also did some experiments with specialised logic for categorical splits, I don't think they have any effect on accuracy, maybe someone else can provide an example of why this is necessary?

src/common/hist_util.h
```diff
@@ -196,20 +196,10 @@ class CutsBuilder {
...
-          p_cuts_->cut_values_.push_back(cpt);
-        }
-      }
-    } else {
-      for (size_t i = 2; i < summary.size; ++i) {
-        bst_float cpt = summary.data[i - 1].value;
-        if (i == 2 || cpt > p_cuts_->cut_values_.back()) {
-          p_cuts_->cut_values_.push_back(cpt);
-        }
+    for (size_t i = 1; i < summary.size; ++i) {
```
trivialfis: @hcho3 I vaguely remembered that you have some use cases for this specialization.

RAMitchell: I'm still working through a bunch of corner cases for this.

trivialfis: @RAMitchell My preference is either we don't support it, or provide full blown implementation.

hcho3: https://github.com/dmlc/xgboost/issues/5095 and https://github.com/dmlc/xgboost/issues/5096 may be relevant.

RAMitchell:
So WXQSummary.SetPrune() seems to be quite broken. Sometimes it unnecessarily removes elements from a set before it reaches capacity and my tests show incorrect distributions of elements in its quantile ranges. 

I replaced this with the simpler function WQSummary.SetPrune() for the hist and gpu_hist algorithms and everything is working as expected. I believe this function only gets called once towards the end of quantile calculation, so I am not expecting this to affect speed but will check with some benchmarks shortly.

RAMitchell:
@rongou I noticed your sampling tests were quite volatile. The reason is because quantiles calculated on external memory end up slightly different compared to in memory.

The fix is to set max_bins equal to the number of training rows so that the quantiles end up the same for both external memory and in memory (each row has its own bin). Your tests comparing predictions from external memory/in memory now pass to a much higher accuracy.

rongou:
@RAMitchell that's great! Thanks!

src/common/hist_util.h
RAMitchell: The specialisation is unnecessary, we can use the same code to get correct results for both categorical and continuous. After this PR, if the number of unique feature values is less than or equal to `max_bin`, then each value will get its own bin and can be used for splitting. Before this PR, if the number of unique values was > 16 but less than `max_bin`, categorical features would get lost.

RAMitchell:
WQSketch vs  WXQSketch, 5 features
I checked the performance of the different pruning methods to compute DenseCuts and there does not seem to be a meaningful difference. If anything WQSketch is faster.

Issue #5254, opened 2020-01-31
Improving compatibility with Scikit-learn


daviddiazvico:
According to the [Scikit-learn development guide](https://scikit-learn.org/stable/developers/develop.html#parameters-and-init), the `__init__` method should not do any parameter validation or set any "trailing _" parameter.

In order to have an estimator compatible with the automated hyper-parameter search objects (GridSearchCV, RandomizedSearchCV, skopt.BayesSearchCV, etc.), that should be handled by the `fit` method.

In particular, the validation of `missing` and the instantiation of `_Booster` should be moved from `__init__` to the beginning of `fit` in the sklearn wrapper.

Related PR: #5255 

Thanks!

trivialfis:
PR is merged.  Thanks!  The guide will help a lot in the future.

PR #5255, opened 2020-01-31
Improved sklearn compatibility

daviddiazvico:
Related issue: #5254

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -105,9 +105,8 @@ __model_doc = '''
            Using gblinear booster with shotgun updater is nondeterministic as
            it uses Hogwild algorithm.
 
-    missing : float, optional
-        Value in the data which needs to be present as a missing value. If
-        None, defaults to np.nan.
+    missing : float, default np.nan
+        Value in the data which needs to be present as a missing value.
     num_parallel_tree: int
         Used for boosting random forest.
     monotone_constraints : str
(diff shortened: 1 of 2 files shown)


trivialfis:
I wasn't aware of there's a guide.  Thanks!

python-package/xgboost/sklearn.py
```diff
@@ -476,6 +466,9 @@ def fit(self, X, y, sample_weight=None, base_margin=None,
 
                 [xgb.callback.reset_learning_rate(custom_rates)]
         """
+        self.missing = self.missing if self.missing is not None else np.nan
```
StrikerRUS: It is not allowed to change public attributes (without trailing or leading `_`):
https://github.com/scikit-learn/scikit-learn/blob/b194674c42d54b26137a456c510c5fdba1ba23e0/sklearn/utils/estimator_checks.py#L949-L955
Instead, `self.missing_` should be created in `fit()` and used later.

daviddiazvico: You are right, I think it's fixed now.

trivialfis:
@hcho3 Could you reproduce the failure on Travis?

hcho3:
I have no idea why `distributed` package is causing trouble.

daviddiazvico:
If I'm teading the log correctly, it is just a timeout error:


    _____________________ ERROR at teardown of test_classifier _____________________
    ...
    E               concurrent.futures._base.TimeoutError

Maybe just re-running the test will yield a different result.

trivialfis:
@hcho3 Seems I fixed the timeout.

PR #5258, opened 2020-02-01
[WIP] Handle degenerate cases that result in NAN's in leaf outputs

hcho3:
See discussion at https://discuss.xgboost.ai/t/still-getting-unexplained-nans-new-replication-code/1383?u=hcho3.

In certain degenerate cases, both the hessian sum and `reg_lambda` can be zero, leading to NANs in leaf outputs.

This PR eliminates NANs, by returning 0.0 whenever both hessian sum and `reg_lambda` are zero.

~It also adds a check in the Learner class to ensure that there is no INF or NAN in the prediction result vector.~ I left out the check for now, since performance impact is uncertain.

~Note: I think NANs in the prediction is a serious bug, so its fix should be part of 1.0.0 release.~ Will return after 1.0.0 release.

@dan-stromberg

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -826,6 +826,7 @@ class LearnerImpl : public Learner {
       gbm_->PredictLeaf(data, &out_preds->HostVector(), ntree_limit);
     } else {
       this->PredictRaw(data, out_preds, training, ntree_limit);
+      this->AssertPredictionIsFinite(out_preds);  // Scan for INFs and NANs
       if (!output_margin) {
         obj_->PredTransform(out_preds);
       }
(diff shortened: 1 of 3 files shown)


hcho3:
@trivialfis No, since the goal is to prevent a zero denominator. A denominator of 1e-6f is valid and thus not degenerate.

hcho3:
The `test_regressor` test is throwing `TimeoutError` exception at teardown. I will come back to it later.

trivialfis:
I mean replacing 0 with Eps as denominator, instead returning 0 as result.

hcho3:
So apply a clipping operation to the denominator then? I see now that it would be more a reasonable choice.

trivialfis:
I'm not sure what does "clipping" mean in this context.  I think the heuristic adopted elsewhere is replacing 0 with a small float, so the value is changed from `NaN` to something close to `inf`.

trivialfis:
Let me go through it one more time.

trivialfis:
We need to make sure that, when hessian is invalid (after regularization), the split is guaranteed to be invalid.

hcho3:
@trivialfis Actually, in this degenerate case, no split meets the standard of `min_child_weight` specified. So this line

https://github.com/dmlc/xgboost/blob/c67163250e4b8a815a7ec16dc0ccaf49e3d58954/src/tree/updater_quantile_hist.cc#L1034

is never called, and `best` probably contains all zeros.

I cannot think of any other scenarios where the hessian would be zero.

trivialfis:
@hcho3 If it's not clear then we don't have to fix it right now.

hcho3:
@trivialfis Also the issue only appears when the regularization parameter `reg_lambda` is set to zero.

For now, can we just have the function return a zero weight whenever `sum_hess` is zero? I'd like to return to it after the release.

hcho3:
Closing for now. I will direct the OP to set a non-zero `reg_lambda`.

trivialfis:
There are a few other places employing this equation, fixing one of them the other will diverse.  It might be better that we can investigate it further.

PR #5260, opened 2020-02-01
Add a general parameter for XGBoost4j-Spark to allow skipping missing value processing

zhangzhang10:
This PR adds a boolean parameter 'skipProcessingMissing' to allow skipping missing value processing when the input is known to contain no missing values. If, before being passed to XGBoost, the input is already pre-processed so that missing values are removed or transformed, then there is no need for XGBoost to search for and process missing values again. In our experiment using a pre-processed "mortgage" dataset (600M+ rows, 46 columns), this parameter can save the run time substantially on Spark clusters:

Cluster size | Run time reduced on average
------------  | ---------------------------------
3 nodes, 21 workers | 55 seconds
4 nodes, 28 workers | 41 seconds

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -69,6 +69,7 @@ private[this] case class XGBoostExecutionParams(
     obj: ObjectiveTrait,
     eval: EvalTrait,
     missing: Float,
+    skipProcessingMissing: Boolean,
     allowNonZeroForMissing: Boolean,
     trackerConf: TrackerConf,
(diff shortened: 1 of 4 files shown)


CodingCat:
what's the total training time?

zhangzhang10:
> what's the total training time?

Total training time depends on training parameters and cluster configurations. As an example, training the aforementioned dataset using the 'hist' tree method, 'lossguide' grow policy, 21 workers (3 nodes, 7 workers each), 8 threads per worker, for 20 rounds, I observed average total training time as follows:

with 'skipProcessingMissing' | without 'skipProcessingMissing' | time save (seconds)
------------------------------- | ----------------------------------- | ----------------------
392                                         | 449                                              | 57

Thanks.

CodingCat:
it sounds like only a 10% improvement, and 449/20 rounds itself is a very small training case, when you work on larger dataset, processingMissing will hardly be the bottleneck...

in conclusion, I didn't see significant value for this

PR #5277, opened 2020-02-04
[WIP] Fix loading old logit model, helper for converting old pickle.

trivialfis:
#5276

PR #5279, opened 2020-02-04
Don't define label encoder for binary logistic.

trivialfis:
During testing #5277 , I found the generated model is extremely huge and slow.  The result is binary logistic doesn't require label to be strictly in {0, 1}.  Anything fits into sigmoid is accepted.  So running label encoding will simply return the labels in worse case.

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -738,7 +738,7 @@ class XGBModel(XGBModelBase):
     "Implementation of the scikit-learn API for XGBoost classification.",
     ['model', 'objective'])
 class XGBClassifier(XGBModel, XGBClassifierBase):
-    # pylint: disable=missing-docstring,invalid-name,too-many-instance-attributes
+    # pylint: disable=missing-docstring,too-many-instance-attributes
     def __init__(self, objective="binary:logistic", **kwargs):
         super().__init__(objective=objective, **kwargs)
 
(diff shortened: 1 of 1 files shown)


python-package/xgboost/sklearn.py
```diff
@@ -749,7 +749,10 @@ def fit(self, X, y, sample_weight=None, base_margin=None,
         # pylint: disable = attribute-defined-outside-init,arguments-differ
 
         evals_result = {}
-        self.classes_ = np.unique(y)
+        if self.objective != 'binary:logistic':
+            self.classes_ = np.array([0.0, 1.0])
```
hcho3: ??? Why 0 and 1 here?

trivialfis: @hcho3  Sorry, still writing tests.  Just raised the PR early.

trivialfis:
@hcho3 It's a little bit tricky.  If we don't run `np.unique`, we don't know how many classes are there, hence we can't decide the objective.  But if we run it on a binary classification model, `np.unique` is not accurate.

hcho3:
> But if we run it on a binary classification model, np.unique is not accurate

Why? You’d still get two classes in this case?

trivialfis:
@hcho3 The issue is, `y` is not necessarily integers or categorical.  It can be floating point numbers.

trivialfis:
@hcho3 We can throw a warning for non-discrete types.  But since the input varies as there are numpy, pandas, and possibly others may also works with Python duck typing.  Checking types is not easy either.  Do you have any suggestion?

PR #5280, opened 2020-02-04
Fix CMake + MSVC build with setuptools.

trivialfis:
Follow up fix for #5271 .

### DIFF
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -95,6 +95,7 @@ class BuildExt(build_ext.build_ext):
     def build(self, src_dir, build_dir, generator, build_tool=None, use_omp=1):
         '''Build the core library with CMake.'''
         cmake_cmd = ['cmake', src_dir, generator]
+
         for k, v in USER_OPTIONS.items():
             arg = k.replace('-', '_').upper()
             value = str(v[2])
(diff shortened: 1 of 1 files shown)


trivialfis:
There are `sdist` tests on travis for OSX and Linux.  But no such test for Windows yet.

hcho3:
Usually, Windows users do not have CMake or MSVC installed on their machines. We should document what the user needs to use `build_ext` on Windows.

trivialfis:
@hcho3  We have documents for building C++ library.  Also I added some notes here: https://xgboost.readthedocs.io/en/latest/build.html#python-package-installation at last PR.  Do you think it's enough?

hcho3:
@trivialfis I have not had a chance to look at #5271 thoroughly. The updated doc looks good to me. I think we should add a note about needing MSVC specifically for Windows Python.

trivialfis:
@hcho3  Got it.  I will add some notes.  Feel free to provide your feedback on the new setuptools script.

hcho3:
Actually, never mind. For now we can keep the build doc as it is. I do want to re-organize the build doc later, something like https://rapids.ai/start.html

trivialfis:
> Actually, never mind. For now we can keep the build doc as it is. I do want to re-organize the build doc later, something like https://rapids.ai/start.html

That's not a link to rapids build doc?  BTW, how do I remove the use of `silent` parameter in the `about` section of our landing page: https://xgboost.ai/about

I think there's minor difference around using CMake between 3.13 and later versions with MSVC.  I use CMake 3.16 for this PR.

hcho3:
> how do I remove the use of silent parameter in the about section of our landing page

You should submit a patch to https://github.com/xgboost-ai/xgboost-ai.github.io.

> That's not a link to rapids build doc?

You can select "Source" in the selector and it display a link to build instruction. MXNet (https://mxnet.apache.org/get_started) does a similar thing. The idea is to make the installation doc page less overwhelming, by letting the user filter the doc by the use case they have in mind.

hcho3:
> I think there's minor difference around using CMake between 3.13 and later versions with MSVC. I use CMake 3.16 for this PR.

Interesting. I was not aware of the difference.

trivialfis:
Got it.  Thanks for the pointers.

trivialfis:
@hcho3 I read this answer https://stackoverflow.com/questions/28350214/how-to-build-x86-and-or-x64-on-windows-from-command-line-with-cmake

I just took the simplest one (<3.13) and it works even I'm running 3.16.  Not sure about the implications.

PR #5283, opened 2020-02-05
Updated Windows build docs

ankane:
Adds instructions for Visual Studio 15 and 16 and  the `cmake --build` command.


trivialfis:
Thanks!

ankane:
Thank you both!

PR #5031, opened 2019-11-12
re-enable recovery test after rabit allow duplicated allreduce calls with same signature

chenqin:
Reeanble rabit tests after fix 
https://github.com/dmlc/rabit/pull/128 merged

Rabit assume bootstrap allreduce has unique signature, in dmlc/xgboost#5012 we find it failed to satisfy such constraint.

The goal of this pr is to loosen such constraint and allow last write overwrite results with same signature during bootstrap phase. Here is example https://github.com/dmlc/xgboost/pull/4990/files#diff-5fabd019d3f1af7572baa4a6301cf076R36

PR #5311, opened 2020-02-15
Enable Rabit test.

trivialfis:
Closes https://github.com/dmlc/xgboost/pull/5031.


trivialfis:
Closing to avoid spamming the list.

@chenqin Could you please help taking a look into #5031 when time allows?  I can't push to your branch as you are using master.

Issue #5313, opened 2020-02-16
Saving pkl model changes default encoding


sh1ng:
xgboost 1.0.0rc2

```
import xgboost as xgb
import numpy as np
import pickle
import locale


def makeXy():
    np.random.seed(1)
    X = np.random.randn(10, 10)
    y = [0, 1] * int(10 / 2)
    return X, y


X, y = makeXy()

dm = xgb.DMatrix(X, label=y)

param = {
    "objective": "binary:logistic",
    'eval_metric': 'auc',
}

model = xgb.train(param, dm, 1)

print(locale.getpreferredencoding(False))

with open("model.pkl", 'wb') as f:
    pickle.dump(obj=model, file=f)

print(locale.getpreferredencoding(False))
```

```
UTF-8
ANSI_X3.4-1968
```

I'm not quite sure, but could it be related to dmlc-core and changing locale?

sh1ng:
Thanks

sh1ng:
@trivialfis are you also going to merge it into `release_1` branch?

trivialfis:
It's merged there.

PR #5314, opened 2020-02-16
Fix changing locale.

trivialfis:
Fixes https://github.com/dmlc/xgboost/issues/5313

Don't do anything special if the original locale is `classic`.  I'm not sure if it's a quality fix, as the default global locale in Python is `utf-8` while `C` in c++.  However changing locale in c++ also changes locale in Python ...  So they have different getters but same setter..


trivialfis:
Update: Just don't set the locale, as number parsing is implemented in XGBoost.

PR #5315, opened 2020-02-17
Remove SimpleCSRSource

RAMitchell:
This PR removes the SimpleCSRSource class to reduce complexity of data handling classes according to the roadmap in #5143. MetaInfo is moved to be a member of SimpleDMatrix and DMatrix serialisation logic is moved to methods of SimpleDMatrix.


src/c_api/c_api.cc
```diff
@@ -324,12 +321,16 @@ XGB_DLL int XGDMatrixFree(DMatrixHandle handle) {
...
+XGB_DLL int XGDMatrixSaveBinary(DMatrixHandle handle, const char* fname,
                                 int silent) {
   API_BEGIN();
   CHECK_HANDLE();
-  static_cast<std::shared_ptr<DMatrix>*>(handle)->get()->SaveToLocalFile(fname);
+  auto dmat = static_cast<std::shared_ptr<DMatrix>*>(handle)->get();
+  if (data::SimpleDMatrix* derived = dynamic_cast<data::SimpleDMatrix*>(dmat)) {
+    derived->SaveToLocalFile(fname);
+  } else {
+    LOG(FATAL) << "Legacy binary saving only supported by SimpleDMatrix";
```
trivialfis: So it's legacy now?  Please open an issue to  make sure it's not used, like not used in spark recovery.  It seems to be a handy method during similar senarios.

RAMitchell: I don't have any plans to remove it, I'm just following the language of the existing comments.

If we wanted to more comprehensively support this across DMatrix types we would lift SaveToLocalFile up to be a virtual method and then provide implementations for other DMatrix types.

RAMitchell: Updated string

PR #5322, opened 2020-02-18
Remove duplicated matplotlib installation

Wei-1:
In `Dockerfile.cpu` `matplotlib` was installed twice.
Removing the duplication for cleanness

### DIFF
--- a/tests/ci_build/Dockerfile.cpu
+++ b/tests/ci_build/Dockerfile.cpu
@@ -22,7 +22,7 @@ ENV GOSU_VERSION 1.10
 # Install Python packages
 RUN \
     pip install pyyaml cpplint pylint astroid sphinx numpy scipy pandas matplotlib sh \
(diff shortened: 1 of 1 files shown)


Wei-1:
Error log shows that `Client: not connected` while testing `xgboost.dask`
Is this an occasional failure?

trivialfis:
No.  Let me take a look, there is a fix by removing use of test fixture in master branch, I'm not sure if your branch has that commit.

Wei-1:
I can see the latest commit is: Display Sponsor button, link to OpenCollective (#5325)
l will rebase this PR to that.

trivialfis:
That's weird ...  Only this PR is having trouble.  @hcho3 Have you seen similar issue in other build?

Wei-1:
Will the order of the `matplotlib` installation make a difference?
Maybe caused by version dependency or something?

Wei-1:
Close because fixed by #5336

PR #5324, opened 2020-02-19
Remove unnecessary DMatrix methods

RAMitchell:
Two methods are removed that do not provide broadly useful functionality:
DMatrix::GetColumnDensity
DMatrix::SingleColumnBlock

Column density is only needed by the exact algorithm, so I compute it in the updater itself. Single column block can be established by other means.

### DIFF
--- a/src/gbm/gbtree.cc
+++ b/src/gbm/gbtree.cc
@@ -27,6 +27,7 @@
 #include "../common/common.h"
 #include "../common/random.h"
 #include "../common/timer.h"
+#include "../data/sparse_page_dmatrix.h"
 
 namespace xgboost {
 namespace gbm {
@@ -130,7 +131,7 @@ void GBTree::PerformTreeMethodHeuristic(DMatrix* fmat) {
       "Tree method is automatically selected to be 'approx' "
       "for distributed training.";
     tparam_.tree_method = TreeMethod::kApprox;
-  } else if (!fmat->SingleColBlock()) {
+  } else if (dynamic_cast<data::SparsePageDMatrix*>(fmat)) {
     LOG(WARNING) << "Tree method is automatically set to 'approx' "
                     "since external-memory data matrix is used.";
     tparam_.tree_method = TreeMethod::kApprox;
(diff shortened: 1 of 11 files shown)


src/linear/updater_gpu_coordinate.cu
```diff
@@ -58,10 +58,15 @@ class GPUCoordinateUpdater : public LinearUpdater {  // NOLINT
 
     num_row_ = static_cast<size_t>(p_fmat->Info().num_row_);
 
-    CHECK(p_fmat->SingleColBlock());
-    SparsePage const& batch = *(p_fmat->GetBatches<CSCPage>().begin());
+    auto batch_iter = p_fmat->GetBatches<CSCPage>().begin();
```
trivialfis: I'm not sure why you want to remove this method and then created two different approaches to obtain the check.  One in gbm with dynamic cast another one here.

RAMitchell: I can use the dynamic cast approach here for consistency.

trivialfis: It seems that single method is more simple, we don't need to include extra header here, it breaks the abstraction.

RAMitchell: I reverted the change. Note that the abstraction is broken just by having a method in a base class that differentiates behaviour based on the subclass.

src/gbm/gbtree.cc
```diff
@@ -27,6 +27,7 @@
 #include "../common/common.h"
 #include "../common/random.h"
 #include "../common/timer.h"
+#include "../data/sparse_page_dmatrix.h"
```
trivialfis: Revert this also.

src/linear/updater_gpu_coordinate.cu
```diff
@@ -14,6 +14,7 @@
 #include "../common/device_helpers.cuh"
 #include "../common/timer.h"
 #include "./param.h"
+#include "../data/sparse_page_dmatrix.h"
```
trivialfis: And this.

PR #5326, opened 2020-02-19
ranking metric computation accelaration on gpu

sriramch:
- this pr accelerates `map`, `ndcg`, `precision`, `auc` and `aucpr` metrics on gpu
   - further, it also accelerates the `auc` and `aucpr` metrics on cpu

the performance numbers for ranking and non-ranking datasets are below.

@RAMitchell @trivialfis @rongou - please review


sriramch:
# performance numbers

## test environment
- 1 socket
- 6 cores/socket
- 2 threads/core
- 80 gb system memory
- v100 gpu

## test
- uses all cpu threads
- builds 100 trees
- runs the different objective functions `rank:pairwise`, `rank:ndcg` and `rank:map`
- computes metrics `map`, `ndcg`, `pre@2`, `auc` and `aucpr`
- uses the small and a large [mslr](https://www.microsoft.com/en-us/research/project/mslr/) ranking benchmark datasets having the following characteristics:
   - small mslr dataset containing ~ 3.6m training instances distributed over ~ 10k groups consuming 3.9 gb disk space
   - large mslr dataset containing ~ 11.3m training instances distributed over ~ 95k groups consuming 13 gb disk space
- metric eval times are reported below

## results
- no additional gpu memory was used
- all times are in **seconds** 
- first column is the eval time on cpu
- second column is the eval time on gpu

small dataset | gpu => train, rank | gpu => train, rank, eval | large dataset | gpu => train, rank | gpu => train, rank, eval
:--: | :--: | :--: | :--: | :--: | :--:
**algo/metric** | **cpu => eval** | **cpu => none** | **algo/metric** | **cpu => eval** | **cpu => none**
 |   |   |  |   |  
pairwise/map | 2.51 | 0.39 | pairwise/map | 9.69 | 1.47
pairwise/ndcg | 6 | 0.57 | pairwise/ndcg | 19.04 | 1.97
pairwise/pre@2 | 2.38 | 0.37 | pairwise/pre@2 | 9.02 | 1.4
pairwise/auc | 21.27 | 0.48 | pairwise/auc | 80.29 | 1.75
pairwise/aucpr | 27.92 | 0.61 | pairwise/aucpr | 115.1 | 1.87
 |   |   |  |   |  
ndcg/map | 2.57 | 0.38 | ndcg/map | 9.53 | 1.4
ndcg/ndcg | 6.02 | 0.56 | ndcg/ndcg | 18.86 | 1.9
ndcg/pre@2 | 2.31 | 0.35 | ndcg/pre@2 | 8.85 | 1.33
ndcg/auc | 21.09 | 0.47 | ndcg/auc | 79.01 | 1.68
ndcg/aucpr | 27.31 | 0.59 | ndcg/aucpr | 111.41 | 1.79
 |   |   |  |   |  
map/map | 2.43 | 0.38 | map/map | 9.57 | 1.41
map/ndcg | 5.84 | 0.56 | map/ndcg | 18.89 | 1.91
map/pre@2 | 2.36 | 0.35 | map/pre@2 | 8.82 | 1.34
map/auc | 21.06 | 0.47 | map/auc | 79.11 | 1.69
map/aucpr | 27.16 | 0.59 | map/aucpr | 111.46 | 1.79

trivialfis:
Is there anyway that we can dispatch the computation kernel based on `generic_parameter_.gpu_id`?  If we can't unify the kernel like other element wise metric/objective, I would like to have a clear separation between two implementations, with cu files containing only GPU kernels.

sriramch:
the following shows the `auc` and `aucpr` metric computation improvements on cpu for a *non ranking* dataset - in this case **higgs**. 

- same host spec as above and builds 100 trees
- train on gpu, metric computation/eval on cpu
- all times are in **seconds**

**metric** | **master** | **this pr**
:--: | :--: | :--: 
auc | 82.95 | 68.7
aucpr | 142.08 | 124.12

sriramch:
> And my point is it would be better if we can avoid the include trick.

@trivialfis  - are you asking if it is possible to separate the cpu and gpu implementations completely into separate files *without* having the `.cc` file include the `.cu` file?

the simplest thing i can think of (for metric `x`) is this:

- register metric `x` referencing the cpu implementation from the `.cc` file if it's a `cpu only` build
- register `x-cpu` referencing the cpu implementation from the `.cc` file if it's a `gpu` build and register metric `x` referencing the gpu implementation from the `.cu` file
   - the gpu implementation will then delegate responsibility to `x-cpu` when no device ordinal is set (by looking at the `generic_parameter_.gpu_id` that you mention)

trivialfis:
> are you asking if it is possible to separate the cpu and gpu implementations completely into separate files without having the .cc file include the .cu file?

Yup.  ;-)

My suggestion is simply two functions called `CPUAUCKernel` and `GPUAUCKernel` , very much like your `EvalRankOnGPU/CPU`, but instead of class members we define them as plain old functions.  Put the GPU function declaration in a header or the cc file, then put the implementation in a cu file.  May need to pass some extra arguments like `tparam_`, less nice but plain old.

@RAMitchell might have better suggestions.

trivialfis:
Another thing is we want to replace the `Allreduce` with mean value by using a sketch algorithm for approximating global sorting in distributed environment.  It's not strictly related to this PR, but seems appropriate to leave a note here.

sriramch:
@trivialfis i have now [decoupled](https://github.com/dmlc/xgboost/commit/8509685cef11a89f3f67c1ee429b7def8f553fde) the cpu and the gpu implementations  such that they reside independently without resorting to the include trick (as you allude). please see if this is palatable. if it is, then i can roll this to other areas where there is a significant digression between the two implementations.

@RAMitchell i would appreciate your input as well on this when you get a chance.

src/common/device_helpers.cuh
```diff
@@ -482,6 +484,176 @@ using device_vector = thrust::device_vector<T,  XGBDeviceAllocator<T>>;
...
+  caching_device_vector<uint32_t> group_segments_;
+
+  // Need this on the device as it is used in the kernels
+  caching_device_vector<uint32_t> dgroups_;       // Group information on device
+
+  // Where did the item that was originally present at position 'x' move to after they are sorted
+  caching_device_vector<uint32_t> dindexable_sorted_pos_;
+
+  // Initialize everything but the segments
+  void Init(uint32_t num_elems) {
```
RAMitchell: Prefer constructors over Init functions. Preferably create this object and destroy it when you need it, instead of using it as a member variable. Caching vectors are designed for this.

src/common/device_helpers.cuh
```diff
@@ -482,6 +484,176 @@ using device_vector = thrust::device_vector<T,  XGBDeviceAllocator<T>>;
...
+    };  // NOLINT
+
+    thrust::transform(thrust::make_counting_iterator(static_cast<uint32_t>(0)),
+                      thrust::make_counting_iterator(num_elems),
+                      group_segments_.begin(),
+                      ComputeGroupIDLambda);
+  }
+
+  // Accessors that returns device pointer
+  inline const T *GetItemsPtr() const { return ditems_.data().get(); }
```
RAMitchell: Pass Span rather than pointers.

src/common/device_helpers.cuh
```diff
@@ -482,6 +484,176 @@ using device_vector = thrust::device_vector<T,  XGBDeviceAllocator<T>>;
...
+  inline uint32_t GetNumItems() const { return ditems_.size(); }
+  inline const caching_device_vector<T> &GetItems() const {
```
RAMitchell: Span over const device vectors.

src/metric/metric.cc
```diff
@@ -26,32 +26,64 @@ Metric* Metric::Create(const std::string& name, GenericParameter const* tparam)
...
+  if (metric == nullptr) {
+    LOG(FATAL) << "Unknown metric function " << name;
+  }
+
+  metric->tparam_ = tparam;
+  return metric;
+}
+
+Metric *
+GPUMetric::CreateGPUMetric(const std::string& name, GenericParameter const* tparam) {
```
RAMitchell: Not sure why this is needed. I don't recall having to do this for other factory methods in GPU code.

src/metric/rank_metric.cc
```diff
@@ -1,21 +1,35 @@
 /*!
- * Copyright 2015 by Contributors
- * \file rank_metric.cc
- * \brief prediction rank based metrics.
- * \author Kailong Chen, Tianqi Chen
+ * Copyright 2020 XGBoost contributors
  */
+// When device ordinal is present, we would want to build the metrics on the GPU. It is possible
```
RAMitchell: This needs a little more thought from us if we want to use this approach going forward. Ideally we would have a consistent configuration logic when choosing GPU algorithms.

src/metric/rank_metric.cc
```diff
@@ -393,140 +380,230 @@ struct EvalMAP : public EvalRankList {
...
+          std::stable_sort(rec.begin(), rec.end(), common::CmpFirst);
+        } else {
+          XGBOOST_PARALLEL_SORT(rec.begin(), rec.end(), common::CmpFirst);
+        }
 
-    const std::vector<bst_float>& h_preds = preds.HostVector();
-    for (omp_ulong i = 0; i < ndata; ++i) {
-      exp_p_sum += h_preds[i];
+        // calculate AUC
+        double sum_pospair = 0.0;
```
RAMitchell: There is a bit much for me to review in one go. In particular when stuff gets moved its harder to see where the algorithm got changed in the diff. Can we try to separate this into a few PRs, maybe one for moving things around, one for logic for launching/configuring GPU metrics, one per changed metric. Given that a AUC is such a key metric I really want to digest whats going on here.

sriramch:
@RAMitchell - i'll go through this and see if i can split this into multiple digestible pr's. in the interim, i'll comment on the few things you have raised.

src/metric/metric.cc
sriramch: the main idea is to decouple the cpu and gpu implementations when there is significant digression *without* the cuda file include trick. @trivialfis raised this issue and the discussion is subsumed in this pr. more specifically, the following change may shed some light.

- [decouplement](https://github.com/dmlc/xgboost/commit/8509685cef11a89f3f67c1ee429b7def8f553fde)

the gpu implementation is now dynamically looked up from the registry and dispatched when there is a valid device present and if xgboost is gpu enabled.

src/common/device_helpers.cuh
sriramch: these are internal methods - not exposed. these types are indeed created and destructed on the fly.

Issue #4237, opened 2019-03-08
External memory cache files are not updated between runs


thvasilo:
Encountered this while trying to test external memory training for [this issue](https://discuss.xgboost.ai/t/model-does-not-train-on-data-in-external-memory/712) on the discussion board.

I was trying to train on RCV1 which by originally has the +1/-1 labels, causing XGB to crash with an error.
I fixed this and pointed to the new file, however the cache files where unchanged, so when I tried to train I got the same error.

This is because the cache files had already been created from my previous run, and when I ran the algo again, they were not generated again. I assume this is expected behavior, but thought we might want to mention that in the tutorial or something to avoid having users train on stale data (or even the wrong dataset) in case they re-use the same cache file names.

Meaning that once you run:

`dtrain = xgb.DMatrix("data/rcv1/data.test#dtrain.cache")`

then running 

`dtrain = xgb.DMatrix("data/another-dataset/data.test#dtrain.cache")` on a separate run will result in training on the old (RCV1) data unless the dtrain.cache files are deleted.

Are the external memory cache files supposed to be removed automatically at training end?

trivialfis:
@thvasilo After digging a lot into XGBoost,  I'm still not familiar with external memory ... I will try having a complete understanding of it before next release hits.

RAMitchell:
I ran into this problem also when I was playing with external memory. You have to manually delete the cache files between runs, otherwise it will silently continue to use previous cache files and produce incorrect results with no warning (!!).

I can see two ways of solving this:
- generate unique cache files for each run
- cleanup cache files in the dmatrix destructor

Option two is preferred so we don't fill up the systems memory. We should be very careful when deleting files though.

rongou:
This is a "feature" to avoid generating the cache files over and over as they are pretty expensive to write out for large datasets. It's this line here: https://github.com/dmlc/xgboost/blob/095de3bf5f9febba9908637197873630021c733a/src/data/data.cc#L265

We can get rid of that line, but I'm not sure if existing users might expect that caching behavior. A few possible solutions:
1. Don't cache the row pages between runs.
1. Add some configuration parameter to keep/delete caches.
1. Somehow be smart and make sure the source data files haven't changed. For example, maybe we can keep some kind of hash of the source data in `MetaInfo`, or check the timestamps of source data and cache files.

My vote is 1. What do you guys think?

@thvasilo @trivialfis @RAMitchell

thvasilo:
[...]

PR #5352, opened 2020-02-24
Partial rewrite EllpackPage

RAMitchell:
This PR tries to resolve a few issues with EllpackPage to allow some new features in future.
- Remove dependency between external objects (e.g. EllpackPageSource) and EllpackPage. Previously the memory for histogram cuts would be stored in some other object. This is dangerous and breaks encapsulation. Proposed design keeps a set of histogram cuts inside EllpackPage so the object is self-sufficient. This does require us to serialise and deserialise cuts for external memory, some performance testing is needed.
- HistogramCuts now uses HostDeviceVector for storage. This lets me use it in EllpackPage for storing the cuts.
- Remove the ability to half construct EllpackPage. Previously it was very easy to incorrectly construct this object, resulting in crashes. Now this mostly happens via safe constructors (excluding compact/copy operations).
- Remove duplicated variables and state inside EllpackPage. Also removed a SparsePage member variable that was there for some reason.
- Switch underlying storage to HostDeviceVector. This greatly simplifies the object state between host and device and means we don't need to use pimpl. No need for an impl->InitDevice() method, HostDeviceVector takes care of all of that.
- Massively simplify EllpackPageSource by sharing code, remove pimpl

### DIFF
--- a/src/common/hist_util.h
+++ b/src/common/hist_util.h
@@ -44,17 +44,35 @@ class HistogramCuts {
   using BinIdx = uint32_t;
   common::Monitor monitor_;
 
-  std::vector<bst_float> cut_values_;
-  std::vector<uint32_t> cut_ptrs_;
-  std::vector<float> min_vals_;  // storing minimum value in a sketch set.
-
  public:
+  HostDeviceVector<bst_float> cut_values_;
+  HostDeviceVector<uint32_t> cut_ptrs_;
+  HostDeviceVector<float> min_vals_;  // storing minimum value in a sketch set.
+
   HistogramCuts();
-  HistogramCuts(HistogramCuts const& that) = delete;
+  HistogramCuts(HistogramCuts const& that) {
+    cut_values_.Resize(that.cut_values_.Size());
+    cut_ptrs_.Resize(that.cut_ptrs_.Size());
+    min_vals_.Resize(that.min_vals_.Size());
+    cut_values_.Copy(that.cut_values_);
+    cut_ptrs_.Copy(that.cut_ptrs_);
+    min_vals_.Copy(that.min_vals_);
+  }
+
   HistogramCuts(HistogramCuts&& that) noexcept(true) {
     *this = std::forward<HistogramCuts&&>(that);
   }
-  HistogramCuts& operator=(HistogramCuts const& that) = delete;
+
+  HistogramCuts& operator=(HistogramCuts const& that) {
+    cut_values_.Resize(that.cut_values_.Size());
+    cut_ptrs_.Resize(that.cut_ptrs_.Size());
+    min_vals_.Resize(that.min_vals_.Size());
+    cut_values_.Copy(that.cut_values_);
+    cut_ptrs_.Copy(that.cut_ptrs_);
+    min_vals_.Copy(that.min_vals_);
+    return *this;
+  }
+
   HistogramCuts& operator=(HistogramCuts&& that) noexcept(true) {
     monitor_ = std::move(that.monitor_);
     cut_ptrs_ = std::move(that.cut_ptrs_);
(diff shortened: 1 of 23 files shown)


trivialfis:
> Remove the ability to half construct EllpackPage. 

What does this mean?

RAMitchell:
If you look at this line of code: https://github.com/dmlc/xgboost/blob/655cf17b60538a8655ca2abff29c74d3ce3a995c/tests/cpp/helpers.h#L280

You can take an existing page and call these public methods on it to change the internals in not very obvious ways. I've tried to prevent that by removing/privatising these methods.

RAMitchell:
@rongou would appreciate your review as well.

src/data/ellpack_page.cu
```diff
@@ -65,47 +61,45 @@ __global__ void CompressBinEllpackKernel(
...
   monitor_.StartCuda("Quantiles");
   // Create the quantile sketches for the dmatrix and initialize HistogramCuts.
-  common::HistogramCuts hmat;
-  size_t row_stride =
-      common::DeviceSketch(param.gpu_id, param.max_bin, param.gpu_batch_nrows, dmat, &hmat);
+  row_stride = common::DeviceSketch(param.gpu_id, param.max_bin,
+                                    param.gpu_batch_nrows, dmat, &cuts_);
   monitor_.StopCuda("Quantiles");
 
   monitor_.StartCuda("InitEllpackInfo");
```
rongou: No longer needed?

src/data/ellpack_page_raw_format.cu
```diff
@@ -17,26 +17,35 @@ class EllpackPageRawFormat : public SparsePageFormat<EllpackPage> {
...
+    return false;
   }
 
   void Write(const EllpackPage& page, dmlc::Stream* fo) override {
     auto* impl = page.Impl();
-    fo->Write(impl->matrix.n_rows);
-    auto buffer = impl->idx_buffer;
-    CHECK(!buffer.empty());
-    fo->Write(buffer);
+    fo->Write(impl->cuts_.cut_values_.ConstHostVector());
```
rongou: This is cool. Perhaps we should have an option to keep the ellpack pages between runs for things like hyperparameter tuning.

src/data/ellpack_page_source.cu
```diff
@@ -3,162 +3,85 @@
...
   auto* impl = page->Impl();
-  impl->matrix.info = ellpack_info_;
-  impl->Clear();
 
-  const MetaInfo& info = dmat->Info();
   size_t bytes_write = 0;
   double tstart = dmlc::GetTime();
   for (const auto& batch : dmat->GetBatches<SparsePage>()) {
-    impl->Push(device_, batch);
+    temp_host_page.Push(batch);
```
rongou: How expensive is this?

RAMitchell: This was being done before, except it was happening inside EllpackPage.

src/data/ellpack_page_raw_format.cu
RAMitchell: Consensus from #4237 was that we would delete cache after each run to prevent undefined behaviours.

src/data/ellpack_page.cu
```diff
@@ -174,198 +174,123 @@ struct CompactPage {
...
-                                        const SparsePage& row_batch,
-                                        const RowStateOnDevice& device_row_state) {
-  // Has any been allocated for me in this batch?
-  if (!device_row_state.rows_to_process_from_batch) return;
-
-  unsigned int null_gidx_value = matrix.info.n_bins;
-  size_t row_stride = matrix.info.row_stride;
+                                        const SparsePage& row_batch) {
+  if (row_batch.Size() == 0) return;
+  unsigned int null_gidx_value = NumSymbols() - 1;
```
trivialfis: This is `num_gidx_value` right?

trivialfis: NVM ... Had a second look.

src/common/hist_util.h
```diff
@@ -113,17 +113,35 @@ class HistogramCuts {
...
-
  public:
+  HostDeviceVector<bst_float> cut_values_;
+  HostDeviceVector<uint32_t> cut_ptrs_;
+  HostDeviceVector<float> min_vals_;  // storing minimum value in a sketch set.
+
   HistogramCuts();
-  HistogramCuts(HistogramCuts const& that) = delete;
+  HistogramCuts(HistogramCuts const& that) {
+    cut_values_.Resize(that.cut_values_.Size());
```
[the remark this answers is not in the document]
RAMitchell: Don't think so. Device gets set when the object gets used in device code.

tests/cpp/tree/gpu_hist/test_gradient_based_sampler.cu
```diff
@@ -29,20 +29,20 @@ void VerifySampling(size_t page_size,
...
   if (fixed_size_sampling) {
     EXPECT_EQ(sample.sample_rows, kRows);
-    EXPECT_EQ(sample.page->matrix.n_rows, kRows);
+    EXPECT_EQ(sample.page->n_rows, kRows);
     EXPECT_EQ(sample.gpair.size(), kRows);
   } else {
-    EXPECT_NEAR(sample.sample_rows, sample_rows, kRows * 0.016f);
-    EXPECT_NEAR(sample.page->matrix.n_rows, sample_rows, kRows * 0.016f);
-    EXPECT_NEAR(sample.gpair.size(), sample_rows, kRows * 0.016f);
+    EXPECT_NEAR(sample.sample_rows, sample_rows, kRows * 0.05f);
```
trivialfis: Why is it necessary to relax the tests?

src/data/ellpack_page.cu
trivialfis: +1

PR #5358, opened 2020-02-25
enable rabit test

chenqin:
reenable rabit failure recovery test


trivialfis:
Thanks for following up on this.

PR #5361, opened 2020-02-25
POC implementation of deterministic GPU histogram.

trivialfis:
* Use pre-rounding based method to obtain reproducible floating point
  summation.
* This makes GPU Hist more stable but not yet fully reproducible.

### DIFF
--- a/src/tree/gpu_hist/histogram.cu
+++ b/src/tree/gpu_hist/histogram.cu
@@ -1,0 +1,184 @@
+/*!
+ * Copyright 2020 by XGBoost Contributors
+ */
+#include <thrust/reduce.h>
+#include <thrust/iterator/transform_iterator.h>
+#include <algorithm>
+#include <ctgmath>
+#include <limits>
+
+#include "xgboost/base.h"
+#include "row_partitioner.cuh"
+
+#include "histogram.cuh"
+
+#include "../../data/ellpack_page.cuh"
+#include "../../common/device_helpers.cuh"
+
+namespace xgboost {
+namespace tree {
+// Following 2 functions are slightly modifed version of fbcuda.
+
+/* \brief Constructs a rounding factor used to truncate elements in a sum such that the
+   sum of the truncated elements is the same no matter what the order of the sum is.
+
+ * Algorithm 5: Reproducible Sequential Sum in 'Fast Reproducible Floating-Point
+ * Summation' by Demmel and Nguyen
+
+ * In algorithm 5 the bound is calculated as $max(|v_i|) * n$.  Here we use the bound
+ *
+ * \begin{equation}
+ *   max( fl(\sum^{V}_{v_i>0}{v_i}), fl(\sum^{V}_{v_i<0}|v_i|) )
+ * \end{equation}
+ *
+ * to avoid outliers, as the full reduction is reproducible on GPU with reduction tree.
+ */
+template <typename T>
+DEV_INLINE __host__ T CreateRoundingFactor(T max_abs, int n) {
+  T delta = max_abs / (static_cast<T>(1.0) - 2 * n * std::numeric_limits<T>::epsilon());
+
+  // Calculate ceil(log_2(delta)).
+  // frexpf() calculates exp and returns `x` such that
+  // delta = x * 2^exp, where `x` in (-1.0, -0.5] U [0.5, 1).
+  // Because |x| < 1, exp is exactly ceil(log_2(delta)).
+  int exp;
+  std::frexp(delta, &exp);
+
+  // return M = 2 ^ ceil(log_2(delta))
+  return std::ldexp(static_cast<T>(1.0), exp);
+}
+
+namespace {
+struct Pair {
+  GradientPair first;
+  GradientPair second;
+};
+DEV_INLINE Pair operator+(Pair const& lhs, Pair const& rhs) {
+  return {lhs.first + rhs.first, lhs.second + rhs.second};
+}
+}  // anonymous namespace
+
+struct Clip : public thrust::unary_function<GradientPair, Pair> {
+  static DEV_INLINE float Pclip(float v) {
+    return v > 0 ? v : 0;
+  }
+  static DEV_INLINE float Nclip(float v) {
+    return v < 0 ? abs(v) : 0;
+  }
+
+  DEV_INLINE Pair operator()(GradientPair x) const {
+    auto pg = Pclip(x.GetGrad());
+    auto ph = Pclip(x.GetHess());
+
+    auto ng = Nclip(x.GetGrad());
+    auto nh = Nclip(x.GetHess());
+
+    return { GradientPair{ pg, ph }, GradientPair{ ng, nh } };
+  }
+};
+
+template <typename GradientSumT>
+GradientSumT CreateRoundingFactor(common::Span<GradientPair const> gpair) {
+  using T = typename GradientSumT::ValueT;
+  dh::XGBCachingDeviceAllocator<char> alloc;
+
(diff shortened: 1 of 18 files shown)


trivialfis:
Experimenting on Year Prediction Dataset:

## Non-deterministic histogram build:
non-deterministic

## Deterministic histogram build:
deterministic

tests/cpp/common/test_gpu_hist_util.cu
```diff
@@ -76,6 +76,20 @@ void TestDeviceSketch(bool use_external_memory) {
     ASSERT_LT(fabs(hmat_cpu.Values()[i] - hmat_gpu.Values()[i]), eps * nrows);
   }
 
+  // Determinstic
```
trivialfis: Note: Typo

src/tree/gpu_hist/histogram.cu
```diff
@@ -0,0 +1,156 @@
...
+/* \brief Constructs a rounding factor used to truncate elements in a sum such that the
+   sum of the truncated elements is the same no matter what the order of the sum is.
+
+ * Follows Algorithm 5: Reproducible Sequential Sum in 'Fast Reproducible Floating-Point
+ * Summation' by Demmel and Nguyen
+
+ * @param max max(sum(positive gradient) + sum(abs(negative gradient)))
+ * @param n The number of elements for the sum, or an upper bound estimate
+ */
+DEV_INLINE __host__ float CreateRoundingFactor(float max, int n) {
```
RAMitchell: I was expecting you to do this with doubles, this would give an extra 29 bits of precision and might resolve your accuracy issues. We can make the check to only do the truncation if GradientSumT is double precision.

trivialfis:
Em I still get some non-determinism somewhere.  Need more investigation.

trivialfis:
I was able to obtain bit by bit reproducible result on 1M rows HIGGS dataset.  But started to get discrepancy on ~5M YearPrediction.

trivialfis:
There's minor discrepancy on 5M YearPrediction on may laptop but not on my desktop.   On 1M+ Higgs both have discrepancy, arguably much smaller than non-deterministic hist.

I added a new parameter so that we can switch if on certain datasets the accuracy drops too much.  But so far on YearPrediction and Higgs the accuracy loss is negligible.

trivialfis:
@RAMitchell After getting `dh::SumReduction` out of the way, I was able to obtain bit by bit reproducible result on both YearPrediction, Higgs and Epsilon for 100 iterations.  As for why `cub::DeviceReduce` produces non-deterministic result I'm not sure.

RAMitchell:
Awesome! Maybe uninitialised memory in `node_sum_gradients_d` or someone else?

trivialfis:
@RAMitchell 

> Maybe uninitialised memory in `node_sum_gradients_d`

That's not the cause.

trivialfis:
@RAMitchell  Time for training on Year Prediction is up from `1.46069` to `1.6410` for 100 iterations.  I don't have more commits coming unless there's further optimization required.  Please review when time allows.

trivialfis:
Actually maybe I can pre-compute the rounding.

mli:
# [Codecov](https://codecov.io/gh/dmlc/xgboost/pull/5361?src=pr&el=h1) Report
> Merging [#5361](https://codecov.io/gh/dmlc/xgboost/pull/5361?src=pr&el=desc) into [master](https://codecov.io/gh/dmlc/xgboost/commit/655cf17b60538a8655ca2abff29c74d3ce3a995c&el=desc) will **increase** coverage by `0.31%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/dmlc/xgboost/pull/5361?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master    #5361      +/-   ##
==========================================
+ Coverage   83.75%   84.07%   +0.31%     
==========================================
  Files          11       11              
  Lines        2413     2411       -2     
==========================================
+ Hits         2021     2027       +6     
+ Misses        392      384       -8     
```


| [Impacted Files](https://codecov.io/gh/dmlc/xgboost/pull/5361?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [python-package/xgboost/compat.py](https://codecov.io/gh/dmlc/xgboost/pull/5361/diff?src=pr&el=tree#diff-cHl0aG9uLXBhY2thZ2UveGdib29zdC9jb21wYXQucHk=) | `54.54% <0.00%> (+0.58%)` | :arrow_up: |
| [python-package/xgboost/core.py](https://codecov.io/gh/dmlc/xgboost/pull/5361/diff?src=pr&el=tree#diff-cHl0aG9uLXBhY2thZ2UveGdib29zdC9jb3JlLnB5) | `80.10% <0.00%> (+0.65%)` | :arrow_up: |

------

[Continue to review full report at Codecov](https://codecov.io/gh/dmlc/xgboost/pull/5361?src=pr&el=continue).
> Powered by [Codecov](https://codecov.io/gh/dmlc/xgboost/pull/5361?src=pr&el=footer). Last update [655cf17...e8348f0](https://codecov.io/gh/dmlc/xgboost/pull/5361?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

src/tree/gpu_hist/histogram.cu
```diff
@@ -0,0 +1,155 @@
...
+GradientPair CreateRoundingFactor(common::Span<GradientPair> gpair) {
+  auto positive = []__device__(float v) {
+    return v > 0 ? v : 0;
+  };  // NOLINT
+  auto negative = []__device__(float v) {
+    return v < 0 ? abs(v) : 0;
+  };  // NOLINT
+
+  dh::XGBCachingDeviceAllocator<char> alloc;
+  auto positive_sum = thrust::reduce(
```
RAMitchell: Not sure if it matters, but it is possible to fuse these two reductions into a single reduction e.g. apply transform iterator to gradient pair inputs that returns 2 GradientPair (pos/neg) and define an operator to sum them. Alternatively just use the absolute sum, although this is slightly less tight and we could lose up to one bit of precision.

src/tree/gpu_hist/histogram.cuh
```diff
@@ -0,0 +1,40 @@
...
+#define HISTOGRAM_CUH_
+#include <thrust/transform.h>
+#include "../../data/ellpack_page.cuh"
+
+namespace xgboost {
+namespace tree {
+
+GradientPair CreateRoundingFactor(common::Span<GradientPair> gpair);
+
+DEV_INLINE float TruncateWithRoundingFactor(float rounding_factor, float x) {
```
RAMitchell: I don't understand how this function could be so expensive that it slows down the algorithm if we use it in the histogram kernel.

src/tree/gpu_hist/histogram.cu
trivialfis: @RAMitchell I tried using `cub::DeviceReudce` with 2 cuda streams for this, the end result doesn't seem to be any different.  As for abs sum, it failed a Windows test for GPU ranking when using float.

src/tree/gpu_hist/histogram.cuh
trivialfis: @RAMitchell I pushed the change for computing truncation on the fly.  There are other functions like `GetHess`, `GetGrad` and constructors.  I'm not sure if the compiler can optimize them away.

trivialfis:
> I think we need a Python level integration test.

Done.

RAMitchell:
@sh1ng @pseudotensor you might find this feature useful.

Issue #5363, opened 2020-02-25
ValueError: underlying buffer has been detached


brydag:
Hello

Please help as soon as possible. I have the issue as below:

xgboost\compat.py", line 101, in <module>
    if sys.__stdin__ is not None and sys.__stdin__.closed:
ValueError: underlying buffer has been detached

Python version 3.7.4
Anaconda3 conda 4.8.2
xgboot 1.0.1
Windows Server2012R2

Packages are in conda virtual environment

brydag:
Your fix its work perfectly. Thank you very much.

PR #5364, opened 2020-02-25
Use type name for data type check.

trivialfis:
I replaced explicit import for datatable with simple string.  The trick should also work with other packages.  But I'm keeping this PR to be minimum so that others can raise concerns around the new compatibility check.

Should close https://github.com/dmlc/xgboost/issues/5363 .  @brydag Could you please try this patch?  I can't reproduce it as there are some specific libraries in your dependencies that are using `sys.__stdin__`.

@RAMitchell  Just tried importing dask + datatable, it seems they still don't play well with each others.  But I don't think XGBoost is a good place to workaround the limitation.

@hcho3

### DIFF
--- a/python-package/xgboost/compat.py
+++ b/python-package/xgboost/compat.py
@@ -79,6 +79,14 @@ else:
 # END NUMPY PATHLIB ATTRIBUTION
 ###############################################################################
 
+
+def lazy_isinstance(instance, module, name):
+    '''Use string representation to identify a type.'''
+    module = type(instance).__module__ == module
+    name = type(instance).__name__ == name
+    return module and name
+
+
 # pandas
 try:
     from pandas import DataFrame, Series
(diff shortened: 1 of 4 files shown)


brydag:
Thank you very much it is work perfect. I spent so much time to google it. Thank you much.

hcho3:
Should we include this fix in 1.0.2?

trivialfis:
@hcho3 I would like to hear about your opinion.  I want to merge it in 1.0.2 as I keep it small, but it will be great if you can help deciding about it as I sometimes can be over optimistic.

trivialfis:
@RAMitchell That's a risk we need to take.  For datatable the original code uses `hasattr` to dispatch so I assume it's a stable structure.  On the bright side, we have tests for all packages we support.

hcho3:
I think it's small enough to be included in the patch release.

trivialfis:
> I think it's small enough to be included in the patch release.

I will push into 1.0 release branch.

hcho3:
@brydag The 1.0.2 release is now available.

PR #5365, opened 2020-02-26
Sketching from adapters

RAMitchell:
Overhaul of GPU sketching code to support sketching on external data.

Todo:
Verify peak memory usage

Some performance charts:
AdapterDeviceSketch, DeviceSketch, DeviceSketchOld and WQSketch
Sketch performance varying number of columns


trivialfis:
I'm a little bit concerned about removing weight.

hcho3:
Weighted quantile sketch is one of prominent contribution of XGBoost paper (2016). Is it necessary to remove it?

RAMitchell:
We cant currently access weights via the adapter constructor - they get added after in different c_api functions. If we can change the c_api to get weights on dmatrix construction we can do it.

trivialfis:
Can we create an empty DMatrix handle at first?

RAMitchell:
Current implementation preserves existing behaviour, using weights to compute quantiles when available from the DMatrix. When we implement DMatrix from adapters we can revisit how to access weights in this case.

src/common/hist_util.cu
```diff
@@ -98,379 +43,413 @@ __global__ void UnpackFeaturesK(float* __restrict__ fvalues,
...
-    // copy the weights if necessary
-    if (has_weights_) {
-      const auto& weights_vec = info.weights_.HostVector();
-      dh::safe_cuda(cudaMemcpyAsync(weights_.data().get(),
-                                    weights_vec.data() + batch_row_begin,
-                                    batch_nrows * sizeof(bst_float),
-                                    cudaMemcpyDefault));
-    }
+template <typename ReturnT, typename IterT, typename FuncT>
+thrust::transform_iterator<FuncT, IterT, ReturnT> MakeTransformIterator(
```
trivialfis: I think thrust has a similar function.

src/common/hist_util.cu
```diff
@@ -98,379 +43,413 @@ __global__ void UnpackFeaturesK(float* __restrict__ fvalues,
...
+void ProcessBatch(AdapterT* adapter, size_t begin, size_t end, float missing,
+                  SketchContainer* sketch_container, int num_cuts) {
+  dh::XGBCachingDeviceAllocator<char> alloc;
+  adapter->BeforeFirst();
+  adapter->Next();
+  auto& batch = adapter->Value();
+  // Enforce single batch
+  CHECK(!adapter->Next());
+
+  auto batch_iter = MakeTransformIterator<data::COOTuple>(
```
trivialfis: How about `thrust::make_transform_iterator`?

src/common/hist_util.cu
```diff
@@ -98,379 +43,413 @@ __global__ void UnpackFeaturesK(float* __restrict__ fvalues,
...
+  size_t num_cuts;
+  WQuantileSketch<bst_float, bst_float>::LimitSizeLevel(
+      dmat->Info().num_row_, eps, &dummy_nlevel, &num_cuts);
+  num_cuts = std::min(num_cuts, dmat->Info().num_row_);
+  if (sketch_batch_num_elements == 0) {
+    sketch_batch_num_elements = dmat->Info().num_nonzero_;
+  }
+  dmat->Info().weights_.SetDevice(device);
+  for (const auto& batch : dmat->GetBatches<SparsePage>()) {
+    size_t batch_nnz = batch.data.ConstHostVector().size();
```
trivialfis: If the input is from device this will pull all data back to host.  Do you think we can just call `batch.data.Size()`?

src/common/hist_util.cu
```diff
@@ -9,85 +9,30 @@
...
 #include <memory>
 #include <mutex>
+#include <utility>
+#include <vector>
 
-#include "hist_util.h"
-#include "xgboost/host_device_vector.h"
+#include "../data/adapter.h"
+#include "../data/device_adapter.cuh"
+#include "../tree/param.h"
```
trivialfis: What's used in this header?

src/common/hist_util.cu
```diff
@@ -98,379 +43,413 @@ __global__ void UnpackFeaturesK(float* __restrict__ fvalues,
...
+                                      row_ptrs.end(), element_idx) -
+                  row_ptrs.begin() - 1;
+    d_temp_weights[idx] = weights[ridx + base_rowid];
+  });
+
+  // Sort
+  thrust::sort_by_key(thrust::cuda::par(alloc), sorted_entries.begin(),
+                      sorted_entries.end(), temp_weights.begin(),
+                      EntryCompareOp());
+  std::vector<Entry> entries_t(sorted_entries.begin(), sorted_entries.end());
```
trivialfis: Where is this used?

src/common/hist_util.cu
```diff
@@ -98,379 +43,413 @@ __global__ void UnpackFeaturesK(float* __restrict__ fvalues,
...
+
+  // Scan weights
+  thrust::inclusive_scan_by_key(thrust::cuda::par(alloc),
+                                sorted_entries.begin(), sorted_entries.end(),
+                                temp_weights.begin(), temp_weights.begin(),
+                                [=] __device__(const Entry& a, const Entry& b) {
+                                  return a.index == b.index;
+                                });
+  std::vector<float> weights_t(temp_weights.begin(), temp_weights.end());
```
trivialfis: Where is this used?

src/common/hist_util.cu
```diff
@@ -98,379 +43,413 @@ __global__ void UnpackFeaturesK(float* __restrict__ fvalues,
...
-    auto counting = thrust::make_counting_iterator(size_t(0));
-    using TransformT = thrust::transform_iterator<decltype(get_size), decltype(counting), size_t>;
-    TransformT row_size_iter = TransformT(counting, get_size);
-    size_t batch_row_stride =
-        thrust::reduce(row_size_iter, row_size_iter + n_rows_, 0, thrust::maximum<size_t>());
-    row_stride_ = std::max(row_stride_, batch_row_stride);
-  }
+// Count the entries in each column and exclusive scan
+void GetColumnSizesScan(int device,
+                        dh::caching_device_vector<size_t>* column_sizes_scan,
```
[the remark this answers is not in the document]
RAMitchell: Noted. If something is a sum I think it is reasonable to use size_t. For example a sum of `bst_row_t` can exceed 32 bits.

src/common/hist_util.cu
RAMitchell: I wrote this because of issues I used to have with msvc where it could not deduce the return type for the thrust version. I tried it without and it seems to be resolved now :)

src/common/hist_util.cu
RAMitchell: Ah, that was left behind from debugging.

src/common/hist_util.cu
RAMitchell: Actually I get a fatal error with thrust::make_transform_iterator so there seems to be problems still. I left this helper function in with a note.

trivialfis:
I don't think the maker function can be a cause of fatal error right?

trivialfis:
@RAMitchell 

This is the whole definition of `thrust::make_transform_iterator`:

``` cuda
template <class AdaptableUnaryFunction, class Iterator>
inline __host__ __device__
transform_iterator<AdaptableUnaryFunction, Iterator>
make_transform_iterator(Iterator it, AdaptableUnaryFunction fun)
{
  return transform_iterator<AdaptableUnaryFunction, Iterator>(it, fun);
} // end make_transform_iterator
```

Other than it's a host device function, there's no  difference with your implementation.  Could you take a deeper look into the fatal error you encountered?  I think it's worth the debugging effort.

RAMitchell:
My implementation explicitly specifies the return type. The issue has always beem automatic return type deduction on msvc. I can look further.

trivialfis:
Yup thanks!  If you just print the return type by debugger or use `typeid(RetType).name()`.  If that's the root cause than there's nothing we can do.  But I'm still interested in what other type can MSVC deduce. ;-)  It's just an unification algorithm, should not be way too weird.

RAMitchell:
These are the two return types for my version and thrust:
```c++
thrust::transform_iterator<__nv_dl_wrapper_t<__nv_dl_tag<void (__cdecl*)(xgboost::data::CupyAdapter *,unsigned __int64,unsigned __int64,float,xgboost::common::SketchContainer *,int),&xgboost::common::ProcessBatch<xgboost::data::CupyAdapter>,1>,xgboost::data::CupyAdapterBatch const >,thrust::counting_iterator<unsigned __int64,thrust::use_default,thrust::use_default,thrust::use_default>,xgboost::data::COOTuple,thrust::use_default>

thrust::transform_iterator<__nv_dl_wrapper_t<__nv_dl_tag<void (__cdecl*)(xgboost::data::CupyAdapter *,unsigned __int64,unsigned __int64,float,xgboost::common::SketchContainer *,int),&xgboost::common::ProcessBatch<xgboost::data::CupyAdapter>,2>,xgboost::data::CupyAdapterBatch const >,thrust::counting_iterator<unsigned __int64,thrust::use_default,thrust::use_default,thrust::use_default>,thrust::use_default,thrust::use_default>
```

trivialfis:
Let me take another look today.  It seems weird that the `GetRowStride` is defined twice and have to pull all data to host.

PR #5378, opened 2020-02-28
- move segment sorter to common

sriramch:
- this is the first of a handful of pr's that splits the larger pr #5326
  - it moves this facility to common (from ranking objective class), so that it can be
    used for metric computation
  - it wraps all the bald device pointers into span
  - it uses the device caching allocator where applicable

@RAMitchell @trivialfis @rongou - please review

there is no functionality change besides what has been explicitly called out.

i'll next create a pr for the cpu side metric changes

### DIFF
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -8,7 +8,9 @@
 #include <thrust/system/cuda/error.h>
 #include <thrust/system_error.h>
 #include <thrust/logical.h>
+#include <thrust/gather.h>
 
+#include <omp.h>
 #include <rabit/rabit.h>
 #include <cub/cub.cuh>
 #include <cub/util_allocator.cuh>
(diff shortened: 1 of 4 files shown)


src/common/device_helpers.cuh
```diff
@@ -1285,6 +1287,175 @@ thrust::device_ptr<T const> tcend(xgboost::common::Span<T> const& span) {
...
+    };  // NOLINT
+
+    thrust::transform(thrust::make_counting_iterator(static_cast<uint32_t>(0)),
+                      thrust::make_counting_iterator(num_elems),
+                      group_segments_.begin(),
+                      ComputeGroupIDLambda);
+  }
+
+  // Accessors that returns device pointer
+  inline uint32_t GetNumItems() const { return ditems_.size(); }
```
trivialfis: Just a note here.  All class method defined in-class are automatically inlined.  These inline specifiers are not necessary.

trivialfis:
Please rebase.

Issue #5338, opened 2020-02-21
When to drop Python 3.5


hcho3:
Python 3.5 is reaching its end-of-life on September this year. I'd like to start a discussion to decide when we should stop supporting Python 3.5.

Personally, I would really love to use f-strings to format strings, and f-string is only available in Python 3.6+.

Iqrar99:
Agree, using f-string is a good decision. I would like to work on this.

JohnZed:
Release 1.1 could be a good time to drop it?  Complete EOL of 3.5 is September this year, so certainly no lead to go that long...

trivialfis:
@hcho3 Let's pin it to 1.2.

hcho3:
1.2 sounds good.

hcho3:
As of #5388, we will require Python 3.6+.

PR #5388, opened 2020-03-04
Drop Python 3.5; Require Python 3.6+; Replace old-school string formatting with f-strings

Iqrar99:
Based on #5338 , I just replaced some old-school string formatting with f-strings because f-strings are readable, more concise, and less prone to error than other ways of formatting, and also faster.
It's still in progress.

- [x] demo
- [x] jvm-packages
- [x] python-package
- [x] tests


hcho3:
Note that there’s still discussion ongoing about when to drop Python 3.5. This pull request will have to wait until the discussion concludes.

Iqrar99:
okay, no problem. :)

hcho3:
@Iqrar99 We are moving forward with this PR.

hcho3:
Pylint throws warning for using f-strings in loggers. I removed f-strings from loggers to pass lint check.

hcho3:
I understand that this PR is intrusive, changing a large part of the codebase. I'd like to get approval from 2-3 committers before we merge this PR. An alternative is to merely allow f-strings in the future and keep old string formatting code as is.

hcho3:
I changed my mind now. This PR ended up being quite intrusive, touching all Python files. Whole-sale change like this may introduce unintended breakage.

hcho3:
I opened #5715 to replace this PR.

Issue #5339, opened 2020-02-21
[Discussion] XGBoost Predict is not threadsafe


kylejn27:
I would like to start a formal discussion on potential fixes for the thread safety issue of the xgboost predict method. It is noted in a few places that the predict method is not thread safe [ [in this issue](https://github.com/dmlc/xgboost/issues/3371) and [in the Python API](https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/core.py#L1349-L1366) and [in the C++ API](https://github.com/dmlc/xgboost/blob/master/include/xgboost/predictor.h#L132-L147) ]

A group of my colleagues at Capital One (@ahmadia, @gforsyth, @mmccarty, @stephenpardy) has come together and looked into a few possible suggested changes for this problem and would like to kick off a greater discussion on how best to improve user experience.

The best way to go about doing this would be to dive into the C++ layer and make the underlying methods thread safe. If anybody wishes to take this on, this would be the ideal change. However, this would be an intrusive set of code changes and a less intrusive just as effective change can be implemented at the Python layer. 

We propose the following:
- add a lock to the existing ctypes API call so that the predict method *cannot* be called in a multithreaded manner unintentionally. This prevents users from shooting themselves in the foot.
- build a new, threadsafe predict method or object. This method or object will use `threading.local` memory to store a copy of the booster object. The primary booster object should be automatically copied into thread local memory upon fork.

kylejn27:
Just as a heads up, I'll be unresponsive for about a week but the others that I mentioned will be able to respond. I figured I'd kick off this discussion now though.

trivialfis:
Personally I prefer changing the code in C++ to make it thread safe.  To provide some information about the subject, the problem is in our prediction cache.  If we can the this cache thread local then rest should be easy.

trivialfis:
For discussion purpose, would you like to share some context about usage of thread safe prediction?  

For example if you are using a thread management framework to handle incoming prediction queries, then there's no need for using a cache as once the prediction is done, the data is no longer useful for XGBoost.

But if you are doing incremental prediction, like first predict the first 4 trees then obtain the prediction from 8 trees to observe the difference, then the cache is still useful.

RAMitchell:
We will absolutely support contributions to fix this problem. Let us know what you propose in c++ layer and we can help work out a solution.

trivialfis:
[...]

Issue #5207, opened 2020-01-15
Feature request: allow predict to output on GPU device memory


JohnZed:
For users whose whole pipeline is in GPU memory, having predict return results on GPU device memory would be optimal. This particularly comes up in the case of hyperparameter sweeps, where a GPU-based scoring function is available (and fast), but we have to bring predictions to host memory.

Perhaps something like `booster.predict(test_dmatrix, device="gpu0")` would work as an API, with a default to `device="cpu"`?

PR #5389, opened 2020-03-05
Thread safe, inplace prediction.

trivialfis:
* Implementation for numpy, csr, cudf and cupy.
* Tie thread local to Learner. (merged)
* A placeholder for cache id.

Related: https://github.com/dmlc/xgboost/issues/5339 https://github.com/dmlc/xgboost/issues/5207

~Currently `ntree_limit` is not supported, nor `tree_begin`.  I can add these additional parameters, but it feels more  interesting if we can have a  method to "slice" the model somehow.~

The prediction type is specified as a string, and only normal prediction is supported.  I can change it to using explicit parameter with bits like `predict` function.  I just thought maybe string is cleaner.

Lastly, the C API has a lots of parameters as it has to accept most of the parameters from DMatrix and prediction.

### DIFF
--- a/include/xgboost/predictor.h
+++ b/include/xgboost/predictor.h
@@ -16,6 +16,7 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
+#include <mutex>
 
 // Forward declarations
 namespace xgboost {
@@ -54,6 +55,7 @@ struct PredictionCacheEntry {
 class PredictionContainer {
   std::unordered_map<DMatrix *, PredictionCacheEntry> container_;
   void ClearExpiredEntries();
+  std::mutex cache_lock_;
 
  public:
   PredictionContainer() = default;
@@ -133,6 +135,18 @@ class Predictor {
                             const gbm::GBTreeModel& model, int tree_begin,
                             uint32_t const ntree_limit = 0) = 0;
 
+  /**
+   * \brief Inplace prediction.
+   * \param           x                      Type erased data adapter.
+   * \param           model                  The model to predict from.
+   * \param           missing                Missing value in the data.
+   * \param [in,out]  out_preds              The output preds.
+   * \param           tree_begin (Optional) Begining of boosted trees used for prediction.
+   * \param           tree_end   (Optional) End of booster trees. 0 means do not limit trees.
+   */
+  virtual void InplacePredict(dmlc::any const &x, const gbm::GBTreeModel &model,
+                              float missing, PredictionCacheEntry *out_preds,
+                              uint32_t tree_begin = 0, uint32_t tree_end = 0) const = 0;
   /**
    * \brief online prediction function, predict score for one instance at a time
    * NOTE: use the batch prediction interface if possible, batch prediction is
--- a/src/gbm/gbtree.h
+++ b/src/gbm/gbtree.h
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2014-2019 by Contributors
+ * Copyright 2014-2020 by Contributors
  * \file gbtree.cc
  * \brief gradient boosted tree implementation.
  * \author Tianqi Chen
@@ -16,6 +16,7 @@
 #include <string>
 #include <unordered_map>
 
+#include "xgboost/base.h"
 #include "xgboost/data.h"
 #include "xgboost/logging.h"
 #include "xgboost/gbm.h"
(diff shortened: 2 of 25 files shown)


trivialfis:
@CodingCat Not sure if you are still interested in https://github.com/dmlc/xgboost/issues/4354#issuecomment-481847624  .

trivialfis:
I made that PR as it will help grid searching with dask ml so it's also useful for XGBoost to have the functionality.

trivialfis:
Added support for both tree begin and tree end.  Unlike the `predict` function, here they are specified as boosted rounds, which is unambiguous in the context of random forest and multi-output model.

Added a placeholder `cache_id` parameter for the future, where we might want to cache the prediction for inplace predict.  This is practically doable as there's a `id` intrinsic function in Python.

trivialfis:
It's a bit tricky for user to cache the prediction, as there's global bias and base_margin.

include/xgboost/predictor.h
```diff
@@ -133,6 +135,18 @@ class Predictor {
                             const gbm::GBTreeModel& model, int tree_begin,
                             uint32_t const ntree_limit = 0) = 0;
 
+  /**
+   * \brief Inplace prediction.
+   * \param           x                      Type erased data adapter.
+   * \param           model                  The model to predict from.
+   * \param           missing                Missing value in the data.
+   * \param [in,out]  out_preds              The output preds.
+   * \param           tree_begin (Optional) Begining of boosted trees used for prediction.
```
RAMitchell: This is layer begin in some places and tree_begin here, as this intentional? Just want to make sure there is clarity around function arguments specifying numbers of trees.

trivialfis:
@RAMitchell With a simple script:


``` python
import cudf
import xgboost as xgb
from time import time

data = cudf.read_csv('./covtype.data')
target = data.iloc[:, -1]
target -= 1
X = data.iloc[:, :-1]

m = xgb.DMatrix(X, target)

booster = xgb.train({'tree_method': 'gpu_hist', 'num_class': len(target.unique())}, dtrain=m, num_boost_round=100)


start = time()
booster.set_param({'predictor': 'gpu_predictor'})
for i in range(100):
    booster.inplace_predict(X)
end = time()
print(end - start)

booster.set_param({'predictor': 'gpu_predictor'})
start = time()
for i in range(100):
    m = xgb.DMatrix(X)
    booster.predict(m)
end = time()
print(end - start)
```

```
(XGBoost) fis@fis-ord:~/Workspace/XGBoost/XGBoostUtils/thread-safe$ python bench.py
/home/fis/.anaconda3/envs/XGBoost/lib/python3.7/site-packages/fsspec/implementations/local.py:33: FutureWarning: The default value of auto_mkdir=True has been deprecated and will be changed to auto_mkdir=False by default in a future release.
  FutureWarning,
11.088415622711182
58.56940412521362
```

src/gbm/gbtree.h
```diff
@@ -203,6 +204,22 @@ class GBTree : public GradientBooster {
                     bool training,
                     unsigned ntree_limit) override;
 
+  void InplacePredict(dmlc::any const &x, float missing,
+                      PredictionCacheEntry *out_preds,
+                      uint32_t layer_begin = 0,
+                      unsigned layer_end = 0) const override {
+    CHECK(configured_);
+    // From here on, layer becomes concrete trees.
```
trivialfis: @RAMitchell this function changes `layer` to `tree`.

trivialfis:
There seems to be a problem with dask-cudf when `map_partitions` reduces dimensions.  As 0.13 is getting out I will tackle it later.

RAMitchell:
Let me take a quick look at the cudf dmatrix constructor before we merge this to check its not slow for  a simple reason.

trivialfis:
No problem.  Won't merge until there's explicit approval.  The result is reproducible with cupy.

RAMitchell:
I ran this script on my machine with a 1080ti:
```python
import cupy as cp
import xgboost as xgb
from sklearn.datasets import fetch_covtype
from time import time

X, y = fetch_covtype(return_X_y=True)
X = cp.array(X)
y = cp.array(y) - 1

m = xgb.DMatrix(X, y)

try:
    booster = xgb.Booster(model_file='model.bin')
except:
    booster = xgb.train({'tree_method': 'gpu_hist', 'num_class': len(cp.unique(y))}, dtrain=m,
                        num_boost_round=100)
    booster.save_model('model.bin')

start = time()
booster.set_param({'predictor': 'gpu_predictor'})

iterations = 10
for i in range(iterations):
    booster.inplace_predict(X)
end = time()
print(end - start)

booster.set_param({'predictor': 'gpu_predictor'})
start = time()
for i in range(iterations):
    m = xgb.DMatrix(X)
    booster.predict(m, output_margin=True)
end = time()
print(end - start)
```

Initially inplace prediction is 3x faster. I profiled and commented out this one line and now inplace is 1.4x faster. I think we can get the same result with cudf as we have established cudf can be implicitly transformed to row major and use the same constructor path (with a small amount of effort).

Can you please try on your end.

trivialfis:
Yup.  I can confirm it's about 1.6 times faster with cudf.  What's that sync for?

RAMitchell:
Can't remember :) Whatever the reason I'm sure we can figure out how to avoid it.

trivialfis:
@RAMitchell I removed the sync and tests seem to be passing.  WDYT?

msoos:
Just a heads up -- as per #5450  I am getting around 100x slowdown due to the expired entry cleaning after each prediction. So the 1.4x/5.0x slowdown is not what I'm observing. This PR seems to solve the issue for me.

RAMitchell:
That's good enough for me, let's go ahead and add this feature.

trivialfis:
Thanks!

PR #5393, opened 2020-03-06
Move thread local entry into Learner.

trivialfis:
Extracted from #5389 .

* This is an attempt to workaround CUDA context issue in static variable, where
the CUDA context can be released before device vector.

* Fix training with GPU data on multi-threaded environment.  Calling `HostVector` in `Transform` causes race condition.

* Add PredictionEntry to thread local entry.

This eliminates one copy of prediction vector.

* Don't define CUDA C API in a namespace.

### DIFF
--- a/include/xgboost/learner.h
+++ b/include/xgboost/learner.h
@@ -11,6 +11,7 @@
 #include <rabit/rabit.h>
 #include <xgboost/base.h>
 #include <xgboost/feature_map.h>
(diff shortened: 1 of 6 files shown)


trivialfis:
Great, so this is the part that causes appvayor to fail.

trivialfis:
Closing to avoid flushing the CI.

PR #5397, opened 2020-03-06
Adding static library option

dpayne:
This adds a cmake option to build the library statically which is useful when linking into a static executable. The option is disabled by default.


mli:
# [Codecov](https://codecov.io/gh/dmlc/xgboost/pull/5397?src=pr&el=h1) Report
> Merging [#5397](https://codecov.io/gh/dmlc/xgboost/pull/5397?src=pr&el=desc) into [master](https://codecov.io/gh/dmlc/xgboost/commit/8d06878bf9b778db68ae98f68d99a3557c7ea885&el=desc) will **not change** coverage by `%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/dmlc/xgboost/pull/5397?src=pr&el=tree)

```diff
@@           Coverage Diff           @@
##           master    #5397   +/-   ##
=======================================
  Coverage   84.07%   84.07%           
=======================================
  Files          11       11           
  Lines        2411     2411           
=======================================
  Hits         2027     2027           
  Misses        384      384           
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/dmlc/xgboost/pull/5397?src=pr&el=continue).
> Powered by [Codecov](https://codecov.io/gh/dmlc/xgboost/pull/5397?src=pr&el=footer). Last update [8d06878...84817f0](https://codecov.io/gh/dmlc/xgboost/pull/5397?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

CMakeLists.txt
```diff
@@ -133,13 +134,23 @@ add_subdirectory(${xgboost_SOURCE_DIR}/plugin)
...
-#-- Shared library
-add_library(xgboost SHARED ${XGBOOST_OBJ_SOURCES})
-target_include_directories(xgboost
-  INTERFACE
-  $<INSTALL_INTERFACE:${CMAKE_INSTALL_PREFIX}/include>
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>)
-target_link_libraries(xgboost PRIVATE ${LINKED_LIBRARIES_PRIVATE})
+if (BUILD_STATIC_LIB)
+    #-- Static library
+    add_library(xgboost STATIC ${XGBOOST_OBJ_SOURCES})
```
RAMitchell: Can you replace this entire if statement with a cmake generator expression to determine the STATIC/SHARED keyword?

dpayne: I tried using generator expressions but couldn't make it work correctly with an cmake option. Another way is to drop setting the library type explicitly and to use the cmake property https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html. The user can then build statically with ` -DBUILD_SHARED_LIBS=OFF `.

One issue that I did find is the library does not seem to obey `set_output_directory` and leaves the compiles `libxgboost.a` file in the top level build directory. This was true for both attempts at this.

RAMitchell: Add the following to the function `set_output_directory` in cmake/Utils.cmake:
```
		ARCHIVE_OUTPUT_DIRECTORY ${dir}
		ARCHIVE_OUTPUT_DIRECTORY_DEBUG ${dir}
		ARCHIVE_OUTPUT_DIRECTORY_RELEASE ${dir}
		ARCHIVE_OUTPUT_DIRECTORY_RELWITHDEBINFO ${dir}
		ARCHIVE_OUTPUT_DIRECTORY_MINSIZEREL ${dir}
```
This will place the static library in the correct place.

RAMitchell: And yes exposing `BUILD_SHARED_LIBS` as an option seems to be the correct way to do this.

CMakeLists.txt
```diff
@@ -133,8 +133,8 @@ add_subdirectory(${xgboost_SOURCE_DIR}/plugin)
 add_subdirectory(${xgboost_SOURCE_DIR}/src)
 set(XGBOOST_OBJ_SOURCES "${XGBOOST_OBJ_SOURCES};$<TARGET_OBJECTS:objxgboost>")
 
-#-- Shared library
-add_library(xgboost SHARED ${XGBOOST_OBJ_SOURCES})
+#-- library
+add_library(xgboost ${XGBOOST_OBJ_SOURCES})
```
hcho3: CI is breaking because with this change static lib is generated by default. Is it possible to make SHARED lib a default?

dpayne: Ah that's interesting on my local, shared was generated by default. I'll investigate some more and follow up.

hcho3: We could declare `BUILD_SHARED_LIBS` as an option and give false as the default value.

dpayne: Yup, just made that change. Hopefully it passes now.

hcho3:
Let's not use `BUILD_SHARED_LIBS`, since it also seems to affect `libdmlc` (coming from dmlc-core). Can we revert to the first version?

trivialfis:
Thanks for the PR!

PR #5350, opened 2020-02-23
Split up `LearnerImpl`.

trivialfis:
This PR splits up `LearnerImpl` into 3 different components.  Starting from basic configuration to model IO and lastly the actual learner that performs training/prediction .  This is an attempt to modularize it so we can shrink it down in the future.

For instance, some configurations are no longer needed in current state of XGBoost, like the objective function configuration should be simplified.  Binary model IO is really difficult to work with and should be reduced to absolute minimum.  This PR splits the monolithic `LearnerImpl` so we can look into individual group of features more easily.

PR #5404, opened 2020-03-10
Split learner with composition.

trivialfis:
Another version of  #5350, use composition instead of inheritance as suggested by @RAMitchell .  Keeping the old branch for comparison.

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -194,7 +194,7 @@ void GenericParameter::ConfigureGpuId(bool require_gpu) {
 using XGBAPIThreadLocalStore =
     dmlc::ThreadLocalStore<std::map<Learner const *, XGBAPIThreadLocalEntry>>;
 
-struct LearnerAttributs {
+struct LearnerAttributes {
   static std::string const kEvalMetric;  // NOLINT
 
   PredictionContainer cache;
@@ -266,12 +266,12 @@ struct LearnerAttributs {
   }
 };
 
-std::string const LearnerAttributs::kEvalMetric {"eval_metric"};  // NOLINT
+std::string const LearnerAttributes::kEvalMetric {"eval_metric"};  // NOLINT
 
 
 class LearnerConfiguration {
  public:
-  void Configure(LearnerAttributs* attr) {
+  void Configure(LearnerAttributes* attr) {
     if (!attr->need_configuration) { return; }
 
     attr->monitor.Start("Configure");
@@ -336,7 +336,7 @@ class LearnerConfiguration {
     attr->monitor.Stop("Configure");
   }
 
-  void LoadConfig(Json const& in, LearnerAttributs* attr) {
+  void LoadConfig(Json const& in, LearnerAttributes* attr) {
     CHECK(IsA<Object>(in));
     Version::Load(in, true);
 
@@ -376,7 +376,7 @@ class LearnerConfiguration {
     attr->need_configuration = true;
   }
 
-  void SaveConfig(Json* p_out, LearnerAttributs const* attr) const {
+  void SaveConfig(Json* p_out, LearnerAttributes const* attr) const {
     CHECK(!attr->need_configuration) << "Call Configure before saving model.";
     Version::Save(p_out);
     Json& out { *p_out };
@@ -404,7 +404,7 @@ class LearnerConfiguration {
   }
 
  private:
-  void ValidateParameters(LearnerAttributs const* attr) {
+  void ValidateParameters(LearnerAttributes const* attr) {
     Json config { Object() };
     this->SaveConfig(&config, attr);
     std::stack<Json> stack;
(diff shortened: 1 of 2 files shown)


src/learner.cc
```diff
@@ -193,60 +190,129 @@ void GenericParameter::ConfigureGpuId(bool require_gpu) {
...
-  explicit LearnerImpl(std::vector<std::shared_ptr<DMatrix> > cache)
-      : need_configuration_{true} {
-    monitor_.Init("Learner");
-    for (std::shared_ptr<DMatrix> const& d : cache) {
-      cache_.Cache(d, GenericParameter::kCpuId);
+struct LearnerAttributes {
+  /*! \brief Constant string identifying a passed in parameter is a metric name. */
+  static std::string const kEvalMetric;  // NOLINT
+  /*! \brief A global storage for internal prediction. */
+  PredictionContainer cache;
```
RAMitchell: If any of these variables are only used by one of the learner classes and do not need to be shared, move them to be members of that class. The idea is that variables are not shared if they do not need to be.

src/learner.cc
```diff
@@ -193,60 +190,129 @@ void GenericParameter::ConfigureGpuId(bool require_gpu) {
...
-      omp_set_num_threads(generic_parameters_.nthread);
+    if (attr->generic_parameters.nthread != 0) {
+      omp_set_num_threads(attr->generic_parameters.nthread);
     }
 
     // add additional parameters
     // These are cosntraints that need to be satisfied.
-    if (tparam_.dsplit == DataSplitMode::kAuto && rabit::IsDistributed()) {
-      tparam_.dsplit = DataSplitMode::kRow;
+    if (attr->tparam.dsplit == DataSplitMode::kAuto && rabit::IsDistributed()) {
```
RAMitchell: Are we able to remove data split mode in the near future?

src/learner.cc
```diff
@@ -345,141 +481,148 @@ class LearnerImpl : public Learner {
...
 
-  void CheckDataSplitMode() {
-    if (rabit::IsDistributed()) {
-      CHECK(tparam_.dsplit != DataSplitMode::kAuto)
-        << "Precondition violated; dsplit cannot be 'auto' in distributed mode";
-      if (tparam_.dsplit == DataSplitMode::kCol) {
-        // 'distcol' updater hidden until it becomes functional again
-        // See discussion at https://github.com/dmlc/xgboost/issues/1832
-        LOG(FATAL) << "Column-wise data split is currently not supported.";
+  void ConfigureNumFeatures(LearnerAttributes* attr) {
```
RAMitchell: Can we remove this in the near future and use DMatrix as the source of truth?

src/learner.cc
```diff
@@ -345,141 +481,148 @@ class LearnerImpl : public Learner {
...
+    CHECK_NE(attr->mparam.num_feature, 0)
+        << "0 feature is supplied.  Are you using raw Booster interface?";
+    // Remove these once binary IO is gone.
+    attr->cfg["num_feature"] = common::ToString(attr->mparam.num_feature);
+    attr->cfg["num_class"] = common::ToString(attr->mparam.num_class);
+  }
+
+  void ConfigureGBM(LearnerAttributes *attr,
+                    LearnerTrainParam const &old, Args const &args) const {
+    if (attr->gbm == nullptr || old.booster != attr->tparam.booster) {
```
RAMitchell: If Learner had access to parameters on its construction, we could get rid of these chains of Configure methods.

src/learner.cc
```diff
@@ -572,44 +715,43 @@ class LearnerImpl : public Learner {
...
+    LearnerModelParamLegacy mparam = attr->mparam;  // make a copy to potentially modify
     std::vector<std::pair<std::string, std::string> > extra_attr;
     mparam.contain_extra_attrs = 1;
 
     {
       std::vector<std::string> saved_params;
       // check if rabit_bootstrap_cache were set to non zero before adding to checkpoint
-      if (cfg_.find("rabit_bootstrap_cache") != cfg_.end() &&
-        (cfg_.find("rabit_bootstrap_cache"))->second != "0") {
+      if (attr->cfg.find("rabit_bootstrap_cache") != attr->cfg.end() &&
```
RAMitchell: Not sure what `rabit_bootstrap_cache` is

src/learner.cc
trivialfis: Depends.  Do we still want to support different split mode in the future?

src/learner.cc
trivialfis: Once the JVM PR is merged.

src/learner.cc
trivialfis: It doesn't.  The parameters can be loaded from pickle.

src/learner.cc
trivialfis: It's from JVM package.

trivialfis:
I'm afraid the variables are indeed shared.

trivialfis:
@RAMitchell I prefer https://github.com/dmlc/xgboost/pull/5350 over this PR.  I know that jumping around inheritance is not ideal, but at least that way we don't need an extra `this` pointer (`attr` in this PR).  The split PR is only to split up the functionality so we can remove them as much as possible in the future.

RAMitchell:
Sounds good to me.

Issue #5376, opened 2020-02-28
Cannot incrementally train after pruning


stephenpardy:
This bug is related to #5297. The fix in #5335 allows you to prune a model multiple times (thanks for that fix!), but you encounter the same bug when trying to add incremental trees. The setup is familiar from the old issue:
```
from sklearn.datasets import make_classification
import xgboost as xgb
print(xgb.__version__)  # '1.1.0-SNAPSHOT'
X, y = make_classification()
xgb_data = xgb.DMatrix(X, label=y)
param_dict = {
    'booster':'gbtree',
    'objective':'binary:logistic',
    'slient': 0,
    'max_depth':2,
    'tree_method':'exact',
    'eval_metric':'logloss',
}

refresh_dict = {
    'booster':'gbtree',
    'objective':'binary:logistic',
    'slient': 0,
    'process_type':'update',
    'updater':'refresh,prune',
    'gamma': 1,
}
booster = xgb.train(param_dict, xgb_data, 1)

# prune and refresh is fine
pruned_booster = xgb.train(
    refresh_dict,
    xgb_data,
    1,
    xgb_model=booster,
)

# works fine
pruned2_booster = xgb.train(
    refresh_dict,
    xgb_data,
    1,
    xgb_model=pruned_booster,
)

#incremental training throws error
booster = xgb.train(param_dict, xgb_data, 1, xgb_model=pruned_booster)
```

The error is:
```
XGBoostError: [11:25:41] /Users/<USERNAME>/python_github_repos/xgboost/include/xgboost/tree_model.h:346: Check failed: node.IsLeaf(): 
Stack trace:
  [bt] (0) 1   libxgboost.dylib                    0x0000001a188b095e dmlc::LogMessageFatal::~LogMessageFatal() + 110
  [bt] (1) 2   libxgboost.dylib                    0x0000001a189d5d7d xgboost::RegTree::ExpandNode(int, unsigned int, float, bool, float, float, float, float, float, int) + 509
  [bt] (2) 3   libxgboost.dylib                    0x0000001a189d3ba7 xgboost::tree::ColMaker::Builder::FindSplit(int, std::__1::vector<int, std::__1::allocator<int> > const&, std::__1::vector<xgboost::detail::GradientPairInternal<float>, std::__1::allocator<xgboost::detail::GradientPairInternal<float> > > const&, xgboost::DMatrix*, xgboost::RegTree*) + 615
[...]

PR #5406, opened 2020-03-11
Check whether current updater can modify a tree.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/5376 .


trivialfis:
Also fixes https://github.com/dmlc/xgboost/issues/5376#issuecomment-597865759

PR #5414, opened 2020-03-14
[dask] Honor `nthreads` from dask worker.

trivialfis:
Check `nthreads` from dask worker.  This parameter can be set by:
``` python
    with LocalCluster(n_workers=7, threads_per_worker=1) as cluster:
```
or similar command line argument of `dask-worker`.


trivialfis:
@RAMitchell I'm not sure how to test it.

PR #5420, opened 2020-03-15
Device dmatrix

RAMitchell:
This is a proof of concept for a DMatrix class that only contains an EllPack page and may be created efficiently from cupy/cudf inputs. See #5143 for roadmap.

Issues to be resolved:
- The DeviceDMatrix currently has no way to get the parameter `max_bin` so is fixed on 256 bins. 
- I directly exposed the DMatrix class in python. Not sure if this is necessary or the best way to do things.
- 2 new C API functions are added. Again not sure if this is the best approach or we should pass parameters into existing functions to select DMatrix type.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -1001,7 +1001,7 @@ class DMatrix(object):
         self._feature_types = feature_types
 
 
-class DeviceDMatrix(DMatrix):
+class DeviceQuantileDMatrix(DMatrix):
     """Device memory Data Matrix used in XGBoost.
 
     DMatrix is a internal data structure that used by XGBoost
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -1002,10 +1002,13 @@ class DMatrix(object):
 
 
 class DeviceQuantileDMatrix(DMatrix):
-    """Device memory Data Matrix used in XGBoost.
+    """Device memory Data Matrix used in XGBoost for training with tree_method='gpu_hist'. Do not
+    use this for test/validation tasks as some information may be lost in quantisation. This
+    DMatrix is primarily designed to save memory in training and avoids intermediate steps,
+    directly creating a compressed representation for training without allocating additional
+    memory. Implementation does not currently consider weights in quantisation process(unlike
+    DMatrix).
 
-    DMatrix is a internal data structure that used by XGBoost
-    which is optimized for both memory efficiency and training speed.
     You can construct DeviceDMatrix from cupy/cudf
     """
 
@@ -1018,7 +1021,6 @@ class DeviceQuantileDMatrix(DMatrix):
         self.max_bin = max_bin
         if not (hasattr(data, "__cuda_array_interface__") or (
                 CUDF_INSTALLED and isinstance(data, CUDF_DataFrame))):
-            print(type(data))
             raise ValueError('Only cupy/cudf currently supported for DeviceDMatrix')
 
         super().__init__(data, label=label, weight=weight, base_margin=base_margin,
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -287,6 +287,18 @@ def _maybe_pandas_data(data, feature_names, feature_types,
     return data, feature_names, feature_types
 
 
+def _cudf_array_interfaces(df):
+    '''Extract CuDF __cuda_array_interface__'''
+    interfaces = []
+    for col in df:
+        interface = df[col].__cuda_array_interface__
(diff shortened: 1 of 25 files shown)


trivialfis:
I prefer the name `HistogramDMatrix`/`QuantileDMatrix`, as it's created for histogram algorithm.

RAMitchell:
Should we separate implementations for device and host? That would be `DeviceQuantileDMatrix`

trivialfis:
I think a general interface is better.

RAMitchell:
@trivialfis Tests are passing and I have added support for max_bin as a DMatrix parameter. I named the python class DeviceQuantileDMatrix, after some thought I don't want to hedge the CPU version of this class into the same python wrapper. I think it's better to have a separate class. I prefer being explicit about what it is upfront instead of doing automatic magic underneath to switch between functionality.

python-package/xgboost/core.py
```diff
@@ -1001,6 +1001,69 @@ def feature_types(self, feature_types):
         self._feature_types = feature_types
 
 
+class DeviceQuantileDMatrix(DMatrix):
+    """Device memory Data Matrix used in XGBoost.
+
+    DMatrix is a internal data structure that used by XGBoost
```
trivialfis: Do we need to mention `gpu_hist` explicitly ?

python-package/xgboost/core.py
```diff
@@ -1001,6 +1001,69 @@ def feature_types(self, feature_types):
...
+    def __init__(self, data, label=None, weight=None, base_margin=None,
+                 missing=None,
+                 silent=False,
+                 feature_names=None,
+                 feature_types=None,
+                 nthread=None, max_bin=256):
+        self.max_bin = max_bin
+        if not (hasattr(data, "__cuda_array_interface__") or (
+                CUDF_INSTALLED and isinstance(data, CUDF_DataFrame))):
+            print(type(data))
```
trivialfis: remove

src/data/device_dmatrix.cu
```diff
@@ -0,0 +1,251 @@
...
+#include "../common/math.h"
+#include "adapter.h"
+#include "device_adapter.cuh"
+#include "ellpack_page.cuh"
+#include "device_dmatrix.h"
+
+namespace xgboost {
+namespace data {
+
+struct IsValidFunctor : public thrust::unary_function<Entry, bool> {
```
trivialfis: Is this copied from `simple_dmatrix.cu`?

src/data/device_dmatrix.cu
```diff
@@ -0,0 +1,251 @@
...
+                                         device_accessor.NullValue(), idx);
+    }
+  });
+}
+
+// Does not currently support metainfo as no on-device data source contains this
+// Current implementation assumes a single batch. More batches can
+// be supported in future. Does not currently support inferring row/column size
+template <typename AdapterT>
+DeviceDMatrix::DeviceDMatrix(AdapterT* adapter, float missing, int nthread, int max_bin) {
```
trivialfis: Is the weight used?

python-package/xgboost/core.py
RAMitchell: Done, added a little more documentation.

src/data/device_dmatrix.cu
RAMitchell: Yes it's the same but I don't see a convenient place to put it.

src/data/device_dmatrix.cu
RAMitchell: No, I added some mention of this in the python docstring.

trivialfis: @RAMitchell We might need to be careful around this as changing the C api in the future is not easy.

python-package/xgboost/core.py
```diff
@@ -1001,6 +1001,71 @@ def feature_types(self, feature_types):
...
+                         missing=missing,
+                         silent=silent,
+                         feature_names=feature_names,
+                         feature_types=feature_types,
+                         nthread=nthread)
+
+    def _init_from_array_interface_columns(self, df, missing, nthread):
+        """Initialize DMatrix from columnar memory format."""
+        interfaces = []
+        for col in df:
```
trivialfis: Could you extract this as a function.  I believe there's a same portion of code in DMatrix.

src/data/device_dmatrix.cu
trivialfis: > Yes it's the same but I don't see a convenient place to put it.

`adapter.cuh`?

trivialfis:
Please rebase onto master for resolving failure on travis.

PR #5438, opened 2020-03-23
[R-package] fixed inconsistency in R -e calls in FindLibR.cmake

jameslamb:
In this PR, I want to propose cleaning up an inconsistency in `FindLibR.cmake` to make installation of the R package more reliable.

I think that the searches for `LIBR_INCLUDE_DIRS` and `LIBR_LIB_DIR` should use `--vanilla`  to ignore any `.Rprofile` files and the calling environment. This will reduce the chance that a user's particular R setup could cause failures in these calls.

Thanks for your consideration!

### DIFF
--- a/cmake/modules/FindLibR.cmake
+++ b/cmake/modules/FindLibR.cmake
@@ -103,12 +103,12 @@ else()
     )
     # ask R for the include dir
     execute_process(
-      COMMAND ${LIBR_EXECUTABLE} "--slave" "--no-save" "-e" "cat(R.home('include'))"
+      COMMAND ${LIBR_EXECUTABLE} "--slave" "--vanilla" "-e" "cat(R.home('include'))"
       OUTPUT_VARIABLE LIBR_INCLUDE_DIRS
     )
     # ask R for the lib dir
     execute_process(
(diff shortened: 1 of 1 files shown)


trivialfis:
Is it possible that some build environments rely on customized configuration?

jameslamb:
> Is it possible that some build environments rely on customized configuration?

By the time you reach the calls in this file, you've already found `R` / `R.exe`. If people are using `.Rprofile` or `.Renviron` to mess with the environment variables used for `R.home()`to locate one somewhere OTHER than the place known to that R executable, I'd consider that very nonstandard.

And the more common thing I think you'll run into is that people have printed messages in their `.Rprofile` using `cat()` or `print()`, which will break these checks.

```shell
Rscript -e "cat(R.home())"
# /Library/Frameworks/R.framework/Resources

echo "cat('hello')" > .Rprofile
Rscript -e "cat(R.home())"
# hello/Library/Frameworks/R.framework/Resources

rm .Rprofile
echo "cat('hello')" > .Rprofile
Rscript --vanilla -e "cat(R.home())"
# /Library/Frameworks/R.framework/Resource
```

trivialfis:
@jameslamb Thanks for the detailed explanation!

PR #5453, opened 2020-03-30
Prevent copying SimpleDMatrix.

### LINKED ISSUE #5462: [cmake warning] Calling host function from device function in SymbolBits
## Summary

When compiling from master I'm seeing the warning

```
src/data/../common/compressed_iterator.h(37): warning: 
  calling a __host__ function from a __host__ __device__ function is not allowed
```

This seems to come from [`compressed_iterator.h`](https://github.com/dmlc/xgboost/blob/master/src/common/compressed_iterator.h#L35)

```cpp
inline XGBOOST_DEVICE size_t SymbolBits(size_t num_symbols) {
  auto bits = std::ceil(log2(static_cast<double>(num_symbols)));
  return std::max(static_cast<size_t>(bits), size_t(1));
}
```

You include cmath at the top so I think this can be

```cpp
inline XGBOOST_DEVICE size_t SymbolBits(size_t num_symbols) {
  auto bits = ceil(log2(static_cast<double>(num_symbols)));
  return max(static_cast<size_t>(bits), size_t(1));
}
````

## Reproduce

```bash
git clone --recursive https://github.com/dmlc/xgboost
mkdir build
cd build
cmake .. -DUSE_CUDA=ON
make # wait for around 77%
```

I'm using gcc 8.3 though I'm guessing this is an everything thing

trivialfis:
* Set default dtor for SimpleDMatrix to initialize default copy ctor, which is
deleted due to unique ptr.

* Remove commented code.
* Remove warning for calling host function (std::max).  Fix  #5462  .
* Remove warning for initialization order.
* Remove warning for unused variables.

### DIFF
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -64,6 +64,11 @@ inline std::vector<std::string> Split(const std::string& s, char delim) {
   return ret;
 }
 
+template <typename T>
+XGBOOST_DEVICE T Max(T a, T b) {
+  return a < b ? b : a;
+}
+
 // simple routine to convert any data to string
 template<typename T>
 inline std::string ToString(const T& data) {
--- a/src/common/compressed_iterator.h
+++ b/src/common/compressed_iterator.h
@@ -8,6 +8,8 @@
 #include <cstddef>
 #include <algorithm>
 
+#include "common.h"
+
 #ifdef __CUDACC__
 #include "device_helpers.cuh"
 #endif  // __CUDACC__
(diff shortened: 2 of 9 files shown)


SteveBronder:
Can you add #5462 as a linked issue so it's auto closed?

trivialfis:
@SteveBronder Done. ;-)

SmirnovEgorRu:
@trivialfis, just for my understanding - does the PR reduce memory consumption?

trivialfis:
No.

PR #5460, opened 2020-03-31
[POC] Multi target tree.

trivialfis:
This is only a proof of concept implementation for exact tree method.  For anything real we still have a long way to go.

Please ignore all changes in colmaker, it's not intended to be part of final PR.

@RAMitchell @hcho3


trivialfis:
For testing classification please use this, as I haven't decide how to do configuration cleanly inside XGBoost yet so I did not write any interface or demo:
``` python
import xgboost as xgb
import pandas as pd
import numpy as np
from time import time
import pickle
import os
import argparse
import sys

if os.path.exists('covtype.pkl'):
    print('Load')
    with open('covtype.pkl', 'rb') as fd:
        X, y_enc, y, n = pickle.load(fd)
else:
    print('Encode')
    data = pd.read_csv('covtype.data', nrows=None)
    X = data.iloc[:, :-1]
    y = data.iloc[:, -1].values

    def encode(rows, y):
        unique = np.unique(y)
        ret = np.zeros((rows, len(unique)))
        for t in range(len(unique)):
            ret[:, t] = y == unique[t]
        return ret, len(unique)

    y_enc, n = encode(X.shape[0], y)
    with open('covtype.pkl', 'wb') as fd:
        pickle.dump((X, y_enc, y, n), fd)

n_estimators = 64


def bench(ot, is_ori):
    print('[bench', ot, '] ', 'Classes', n, 'y_enc:', y_enc.shape, 'X:',
          X.shape)
    sys.stdout.flush()
    if ot == 'multi':
        updater = 'multi_exact'
    else:
        if is_ori == 1:
            updater = 'grow_colmaker'
        else:
            updater = 'multi_exact'
    reg = xgb.XGBMultiRegressor(output_type=ot,
                                updater=updater,
                                objective='multireg:softprob',
                                num_class=n,
                                n_estimators=n_estimators,
                                n_jobs=16)
    s = time()
    if ot == 'multi':
        reg.fit(X,
                y_enc,
                eval_set=[(X, y_enc)],
                eval_metric=['mtlogloss', 'rmse'])
    elif ot == 'single':
        reg.fit(X,
                y_enc,
                eval_set=[(X, y_enc)],
                eval_metric=['mtlogloss', 'rmse'])
    else:
        raise ValueError('Unknown output_type')
    e = time()
    print('fit:', e - s)

    s = time()
    predt = reg.predict(X)
    e = time()
    print('predict:', e - s)
```

PR #5463, opened 2020-03-31
Fix uninitialized value bug in xgboost callback

### LINKED ISSUE #5424: Dictionary key referenced before assignment during early stop
When running xgboost.cv with early stopping rounds, after a few iterations the code breaks from a KeyError on the line below. There's some situation where state['bst_msg'] is not yet assigned, perhaps because it isn't set in init.

KeyError: 'best_msg'

https://github.com/dmlc/xgboost/blob/8ca06ab329b29f3fc3a92174bcc35ff9896d3a5c/python-package/xgboost/callback.py#L246

I was running my code with nrounds=500 and early_stopping_rounds=20 (though I've run it in the range of 10 to 200 early_stopping_rounds). 

    bst_cv = xgb.cv(params, dtrain, 500, nfold = 5, shuffle=True, early_stopping_rounds=20)

mjkohl32:
Issue: [Dictionary key referenced before assignment during early stop](https://github.com/dmlc/xgboost/issues/5424)  Closes #5424 .

Bug in early stopping rounds callback function when calling xgboost.cv() where dictionary value is never set. In the case where the model makes no improvements starting in the first iteration, state['best_msg'] is never set and there is a KeyError when the callback function returns after the early stopping rounds. I added an initial value (default message) for state['best_msg'] in the place where state['best_score'], etc get initial values.

### DIFF
--- a/python-package/xgboost/callback.py
+++ b/python-package/xgboost/callback.py
@@ -209,6 +209,10 @@ def early_stop(stopping_rounds, maximize=False, verbose=True):
             state['best_score'] = float('-inf')
         else:
             state['best_score'] = float('inf')
+        msg = '[%d]\t%s' % (
+              env.iteration,
+              '\t'.join([_fmt_metric(x) for x in env.evaluation_result_list]))
+        state['best_msg'] = msg
 
         if bst is not None:
             if bst.attr('best_score') is not None:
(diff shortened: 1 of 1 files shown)


RAMitchell:
The fix looks good, if you address the linter warning in CI I can merge.

Issue #5461, opened 2020-03-31
Huge increase in learning time introduced in version 0.81


ldesreumaux:
I recently updated my version of xgboost from 0.80 to 1.0.2 and I noticed a huge increase in learning time.

I investigated the problem and found that this increase is caused by the `Span` class introduced in 2c50278 and first released in version 0.81. Compared to the structure that was used before, this class does some checks (`SPAN_CHECK` instructions in `span.h`). If you remove these checks, you will notice a **huge decrease in learning time (> x2 !)**. In fact, the learning time gets back to what it was before version 0.81.

These checks should only be present in Debug mode.

SmirnovEgorRu:
@ldesreumaux, thank you for reporting the issue.
Could you, please, provide more details (e.g. parameters and dimensions) to reproduce the issue on my side?

ldesreumaux:
Here is a python script to reproduce the issue. With this script, there is a x2 improvement on my laptop when I remove the checks in version 1.0.2.

```
import numpy as np
import xgboost as xgb

import time

def generate_data():
    X = np.random.normal(5, 2, [n_obs, n_features],)
    y = np.random.gamma(2, 4, n_obs)
    return X, y

seed = 42
num_boost_round = 1000
early_stopping_rounds = 10
learning_rate = 0.05
tree_method = "hist"

n_obs = 5*10**6
n_features = 50

params = {
    "learning_rate": learning_rate,
    "tree_method": tree_method,
    "subsample": 0.9,
    "seed": seed,
    "silent": 1,
    "verbosity": 0
}

np.random.seed(seed)
X_train, y_train = generate_data()
X_test, y_test = generate_data()
dtrain = xgb.DMatrix(X_train, label=y_train)
dtest = xgb.DMatrix(X_test, label=y_test)

start_time = time.time()

model = xgb.train(
    params=params,
    dtrain=dtrain,
    num_boost_round=num_boost_round,
    early_stopping_rounds=early_stopping_rounds,
    evals=[(dtest, 'test')],
    verbose_eval=False
)

print("Time:", time.time() - start_time)
print("Trees:", model.best_iteration, "iterations")
```

trivialfis:
A bit surprised your bottle neck is in objective.  Is the slow down still in 1.0.2?

trivialfis:
Which compiler are you using?

ldesreumaux:
I am using Visual Studio 2015 on my Windows laptop.

[...]

PR #5464, opened 2020-03-31
Reduce span check overhead.

trivialfis:
Related: #5461

From my bench having span check is about 10% slower with higgs dataset (a dense dataset is be good enough for this case)  running https://github.com/NVIDIA/gbm-bench .  This PR drops it to 5% while still keeping the possibility of printing out error message.

I still want to keep the check as there are always surprises in real world.

### DIFF
--- a/include/xgboost/span.h
+++ b/include/xgboost/span.h
@@ -29,11 +29,13 @@
 #ifndef XGBOOST_SPAN_H_
 #define XGBOOST_SPAN_H_
 
-#include <xgboost/logging.h>  // CHECK
+#include <xgboost/base.h>
 
 #include <cinttypes>          // size_t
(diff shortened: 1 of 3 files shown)


trivialfis:
ping @SmirnovEgorRu

trivialfis:
@SmirnovEgorRu I'm only guessing based on experiments.  Creating any c++ class in that condition drops the performance down to using `dmlc::LOG` (which creates a std::ostream).  But using pain c style code increase the perf.

This check doesn't stand out in valgrind so I don't really understand the perf drop mentioned in #5461 .

trivialfis:
If you have better way to optimize it (like using a bounded pointer or somehow manipulate the cache), I can drop this PR.

PR #5481, opened 2020-04-03
Upgrade xgboost-spark to spark-3.0-preview2

WeichenXu123:
Upgrade xgboost-spark to spark-3.0-preview2


WeichenXu123:
Gently ping @CodingCat , Thanks!

CodingCat:
thanks for the contribution, but we cannot support a preview version 

and based on the current progress of 3.0, even it is formally released,  we may be in a place that support both 2.4.x and 3.x

Issue #4668, opened 2019-07-16
Allow releasing GPU memory


RAMitchell:
One common piece of feedback we receive about the GPU algorithms is that memory is not released after training. It may be possible to release memory by deleting the booster object but this is not a great user experience.

See
#4018 
#3083 
#2663
#3045 

The reason why we have not implemented this already is that the internal C++ code does not actually know when training is finished. The language bindings call each training iteration one by one and I don't believe we have any information inside the GPU training code to say if another training iteration is expected or not.

I see a few solutions:
1) We try to use some heuristic internally to decide if it is a good time to free all memory from inside GBTree. 
2) We implement an API function for cleanup - this function can be specific to GPU memory or it can just be a general hint for xgboost to delete any working memory or temporary data structures. I do not like this option as it will propagate through the entire code base - the learner, booster, updaters and predictor will all have to implement these methods.
3) We implement a method in the language bindings where the booster object  serializes itself and then deserializes from disk. Doing this will clear all temporary data structures and should leave the booster in a usable state to resume training or do prediction.

I am leaning towards option 3) but I think it relies on #3980 to make sure all parameters are correctly saved. Maybe it's still possible to do this with current serialization and not have any unexpected side-effects due to parameters not all being saved.

@trivialfis @sriramch @rongou

trivialfis:
Or we pass num boost round to c++?

seanthegreat7:
For those who look for a quick workaround till you fix it properly check [my solution here](https://stackoverflow.com/questions/56298728/how-do-i-free-all-memory-on-gpu-in-xgboost)

seanthegreat7:
For those who are looking for a quick workaround until you fix it properly you may want to look at [my solution here](https://stackoverflow.com/questions/56298728/how-do-i-free-all-memory-on-gpu-in-xgboost/56585662#56585662)

trivialfis:
@seanthegreat7 Thanks.  That's actually an interesting workaround.

Lauler:
None of the workarounds seem to be working on Windows 10. Tried deleting and loading the booster object (still crashed). 

Tried predicting in a subprocess similar to @seanthegreat7 (but for R instead of python). The subprocess just ran indefinitely without finishing.

Would indeed be greatly appreciated if you provided a solution for this issue!

jtromans:
I'm finding this very difficult especially when performing a wide parameter search in a loop of some kind. 

For example:
```
exp_models= []
[...]

PR #5484, opened 2020-04-04
Serialise booster after training to reset state

RAMitchell:
Potential fix for #4668

This resets the booster object after python training. It costs around 24 milliseconds to do this and does not seem to be affected by number of trees.
The below script immediately goes OOM without this fix, after the fix it uses a constant amount of memory regardless of number of iterations.

Testing script
```python
import xgboost as xgb
import numpy as np
import GPUtil

n = 10000
m = 100
X = np.random.randn(n, m)
y = np.random.randn(n)
exp_models = []
for i in range(200):
    clf = xgb.XGBRegressor(
        tree_method='gpu_hist', gpu_id=0, n_estimators=500)
    trained_model = clf.fit(X, y, verbose=False)
    exp_models.append(trained_model)
    print(GPUtil.getGPUs()[0].memoryUsed)
```

24ms seems like a small price to pay for this. Would be useful for CPU and GPU algorithms.

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -731,6 +731,12 @@ class LearnerIO : public LearnerConfiguration {
       tparam_.dsplit = DataSplitMode::kRow;
     }
 
+    // Do not load "process_type":"update", this causes all trees to be removed
+    // from the model on configuration
+    if (cfg_.find("process_type") == cfg_.end()) {
+      cfg_["process_type"] = "default";
+    }
+
     this->Configure();
   }
 
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -731,12 +731,6 @@ class LearnerIO : public LearnerConfiguration {
       tparam_.dsplit = DataSplitMode::kRow;
     }
 
-    // Do not load "process_type":"update", this causes all trees to be removed
-    // from the model on configuration
-    if (cfg_.find("process_type") == cfg_.end()) {
-      cfg_["process_type"] = "default";
-    }
-
     this->Configure();
   }
 
--- a/python-package/xgboost/training.py
+++ b/python-package/xgboost/training.py
@@ -109,7 +109,9 @@ def _train_internal(params, dtrain,
     else:
         bst.best_iteration = nboost - 1
     bst.best_ntree_limit = (bst.best_iteration + 1) * num_parallel_tree
-    return bst
+
+    # Copy to serialise and unserialise booster to reset state and free training memory
+    return bst.copy()
 
 
 def train(params, dtrain, num_boost_round=10, evals=(), obj=None, feval=None,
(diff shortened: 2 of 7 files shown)


RAMitchell:
@trivialfis there is a problem with configuration of the pruner after copying. After loading from pickle learner->configure() gets called and messes up the state of gbm. Not sure how to address this.

trivialfis:
Let me look into it.

trivialfis:
After some simple digging, here's what happen:

- Save the model with `process_type == kUpdate`.
- Load the model with `process_type == kUpdate`, which triggers `InitTreesToUpdate`, resulting into empty model.
- Save this model without any actual updating, so the trees in `trees_to_update` are not moved into `trees`, but they are also not saved.
- Load this empty model with `process_type == kUpdate`, resulting into having 0 trees to update, hence the error you see.

RAMitchell:
I am looking into this but there is not an obvious simple fix. `process_type == kUpdate` is hugely problematic with serialisation.

trivialfis:
Is it possible to update without moving the trees?

RAMitchell:
I tried two approaches so far:
- Push trees to update back into the model before saving.
Save methods are const, including the `rabit::Serializable` interface so we can't modify data structures before saving, at least without propagating changes through multiple libraries.
- Remove `trees_to_update` and operate on the model in place during the boosting process. 
Prediction then has no idea that it should only predict on part of the model, unless we introduce a new state and make predictor and anything else that uses the model aware of it.

One final approach I am trying is to only lazily pushed trees to `trees_to_update` during training instead of configuration.

RAMitchell:
> What happens if number of boosting rounds has greater value than number of trees? Is it possible that the vector got initialised again?

It should wrap around again.

> Also what happens if the number of boosting rounds has smaller value?

Extra trees on the end get deleted from the saved model. This is true of the old implementation.

I tried delaying the instantiation of `trees_to_update` as mentioned above and this creates a new bug where the first prediction iteration uses the full model instead of empty model.

I think that the update process is broken. No matter which approach we take there will be edge cases. This PR just highlighted problems that were already there.

trivialfis:
Setting backward compatibility and among of work aside, do you have an ideal design in mind?

RAMitchell:
I'm not sure. The internal gbtree is generally oblivious to which boosting iteration is happening or if training has finished, is ongoing or has just begun. These are important for error-checking any kind of update process. The gbtree should also in theory be able to save and resume its state at any iteration, it's hard to see how to do this for updating.

Also if a user runs an update process and we store state that it has finished updating all trees then tries to run another update process do we throw an error that updating has finished already? There is no way to know if the user wants to do a continuation of previous updating or start from the beginning.

src/learner.cc
```diff
@@ -731,6 +731,12 @@ class LearnerIO : public LearnerConfiguration {
       tparam_.dsplit = DataSplitMode::kRow;
     }
 
+    // Do not load "process_type":"update", this causes all trees to be removed
+    // from the model on configuration
+    if (cfg_.find("process_type") == cfg_.end()) {
```
trivialfis, line 736: I'm not comfortable with putting this in Learner.  
- `process_type` is a training parameter and is not a model parameter, which means you don't need this parameter to define what a model is or how it looks like.  So the hack should go into `LoadConfig`.
- `Load` and `Save` should produce exactly the same internal states, only this way can we guarantee a pickle thrown to another worker or recovered from failure during training can continue to work.
- Also `process_type` is a gbtree specific parameter.

PR #5491, opened 2020-04-06
Improve the histogram kernel performance

canonizer:
- opt into more dynamic shared memory
- better launch configuration

### DIFF
--- a/src/tree/gpu_hist/histogram.cu
+++ b/src/tree/gpu_hist/histogram.cu
@@ -150,21 +150,37 @@ void BuildGradientHistogram(EllpackDeviceAccessor const& matrix,
                             common::Span<GradientPair const> gpair,
                             common::Span<const uint32_t> d_ridx,
                             common::Span<GradientSumT> histogram,
-                            GradientSumT rounding, bool shared) {
-  const size_t smem_size =
-      shared
-      ? sizeof(GradientSumT) * matrix.NumBins()
-      : 0;
-  auto n_elements = d_ridx.size() * matrix.row_stride;
+                            GradientSumT rounding) {
+  // decide whether to use shared memory
+  int device = 0;
+  dh::safe_cuda(cudaGetDevice(&device));
+  int max_shared_memory = dh::MaxSharedMemoryOptin(device);
+  size_t smem_size = sizeof(GradientSumT) * matrix.NumBins();
+  bool shared = smem_size <= max_shared_memory;
+  smem_size = shared ? smem_size : 0;
 
-  uint32_t items_per_thread = 8;
-  uint32_t block_threads = 256;
-  auto grid_size = static_cast<uint32_t>(
-      common::DivRoundUp(n_elements, items_per_thread * block_threads));
+  // opt into maximum shared memory for the kernel if necessary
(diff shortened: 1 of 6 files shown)


src/tree/gpu_hist/histogram.cu
```diff
@@ -145,40 +145,84 @@ __global__ void SharedMemHistKernel(EllpackDeviceAccessor matrix,
   }
 }
 
+template <typename GradientSumT>
+void InitGradientHistogram(int device_idx, int n_bins, HistogramLaunchConfig* config) {
```
RAMitchell, line 149: Prefer returning by value directly instead of via pointer. This is the functional programming "avoid side-effects" style.

RAMitchell, line 149: How expensive is this function? I know some cuda API calls can be very expensive, but I want to do everything possible to avoid storing extra state.

If possible I would call this function before every single kernel invocation and never have to worry about the configuration being passed in an incorrect state.

src/tree/updater_gpu_hist.cu
```diff
@@ -427,7 +427,7 @@ struct GPUHistMakerDevice {
   TrainParam param;
   bool deterministic_histogram;
   bool prediction_cache_initialised;
-  bool use_shared_memory_histograms {false};
+  HistogramLaunchConfig histogram_config;
```
RAMitchell, line 430: Storing this state is a little fragile because developers will come and modify the program later, and if this is incorrectly initialised some strange things will happen.

Also consider the following (admittedly unlikely) cases:
- The type of GPU gets changed during training
- The number of bins or input data gets changed during training

src/tree/gpu_hist/histogram.cuh
```diff
@@ -17,12 +17,26 @@ DEV_INLINE T TruncateWithRoundingFactor(T const rounding_factor, float const x)
   return (rounding_factor + static_cast<T>(x)) - rounding_factor;
 }
 
+struct HistogramLaunchConfig {
+  unsigned grid_size;
+  unsigned block_threads;
+  bool shared;
+  HistogramLaunchConfig(unsigned grid_size, unsigned block_threads, bool shared) :
+    grid_size(grid_size), block_threads(block_threads), shared(shared) {}
+  HistogramLaunchConfig() : grid_size(0), block_threads(0), shared(false) {}
```
RAMitchell, line 26: What happens if this gets used with default parameters?

RAMitchell:
Would also be good to see an example that shows a demonstrable effect on runtime for this PR. This can be a 10 python script on randomly generated data, or use tests/benchmark/benchmark_tree.py.

src/tree/gpu_hist/histogram.cu
canonizer: No longer applies.

src/tree/updater_gpu_hist.cu
canonizer: Removed extra state.

src/tree/gpu_hist/histogram.cuh
canonizer: No longer applies.

PR #5497, opened 2020-04-08
Unify max nodes.

trivialfis:


src/tree/param.h
```diff
@@ -230,6 +230,20 @@ struct TrainParam : public XGBoostParameter<TrainParam> {
...
+
+  bst_node_t MaxNodes() const {
+    if (this->max_depth == 0 && this->max_leaves == 0) {
+      LOG(FATAL) << "Max leaves and max depth cannot both be unconstrained.";
+    }
+    bst_node_t n_nodes{0};
+    if (this->max_leaves > 0) {
+      n_nodes = this->max_leaves * 2 - 1;
+    } else {
+      n_nodes = (1 << (this->max_depth + 1)) - 1;
```
RAMitchell, line 242: Technically it should be the minimum specified by either max_leaves or max_depth in the case where both are specified.

trivialfis: Actually my preference is throw an error when it doesn't match.  But will take the minimum value, seems a bit more friendly.

PR #5504, opened 2020-04-09
Use ellpack for prediction only when sparsepage doesn't exist.

### LINKED ISSUE #5351: Issues in predicting on ellpack.
We implemented predicting on ellpack for reducing memory usage.  But after some more testing, I found the implementation is not thoroughly considered.  Here are a few more cases we need to put more thoughts on:

- What if the user change tree method but use the same DMatrix?
- What if the user train for another round but at the second round they put the training DMatrix as a validation DMatrix?

@RAMitchell

trivialfis:
Closes #5351 .


trivialfis:
Closes #5351 .

PR #5506, opened 2020-04-09
[CI] Use devtoolset-6 because devtoolset-4 is EOL and no longer available

viirya:
When running `build-linux.sh`, the building of docker image will fail due to devtoolset-4-xxx packages are not available now. We should change to use devtoolset-6-xxx.

Tested locally.

### DIFF
--- a/jvm-packages/dev/Dockerfile
+++ b/jvm-packages/dev/Dockerfile
@@ -24,7 +24,7 @@ RUN \
     yum install -y tar unzip wget xz git centos-release-scl yum-utils java-1.8.0-openjdk-devel && \
     yum-config-manager --enable centos-sclo-rh-testing && \
     yum -y update && \
-    yum install -y devtoolset-4-gcc devtoolset-4-binutils devtoolset-4-gcc-c++ && \
+    yum install -y devtoolset-6-gcc devtoolset-6-binutils devtoolset-6-gcc-c++ && \
     # Python
     wget https://repo.continuum.io/miniconda/Miniconda3-4.5.12-Linux-x86_64.sh && \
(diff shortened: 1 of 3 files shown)


viirya:
cc @CodingCat

hcho3:
@CodingCat @trams @viirya Would upgrading to devtoolset-6 change GLIBC version? We need to keep GLIBC version as low as possible to not break old clusters.

viirya:
I don't see significant difference with GLIBC versions between the two.

https://webcache.googleusercontent.com/search?q=cache:DNLdQqw0xQgJ:https://centos.pkgs.org/6/centos-sclo-rh-testing-x86_64/devtoolset-4-gcc-5.3.1-6.1.el6.x86_64.rpm.html+&cd=2&hl=en&ct=clnk&gl=us

https://centos.pkgs.org/6/centos-sclo-rh-x86_64/devtoolset-6-gcc-6.3.1-3.1.el6.x86_64.rpm.html

viirya:
Thanks @hcho3!

hcho3:
CUDA 9.0 doesn't work with GCC 6, with cryptic error
```
Internal Compiler Error (codegen): "there was an error in verifying the lgenfe output!"
```
Minimal example is available at https://gist.github.com/akors/e945c7099100f7069c3477275bbfdc04.

I don't think we want to drop support for CUDA 9.0 just yet. So we will use devtoolset-6 everywhere, except for the GPU build, where we use devtoolset-4. I managed to get devtoolset-4 working by specifying full URLs manually (see my latest commit).

viirya:
Thanks @hcho3

PR #5507, opened 2020-04-10
Remove `distcol` updater.

### LINKED ISSUE #5498: [RFC] Remove `distcol` updater.
The updater is long broken: https://github.com/dmlc/xgboost/issues/1832 and all other tree methods are distributed.  So maintaining this updater might not be necessary.

If we want to restore its support,  first thing needed to to is implementing a method for tree updaters to access a buffered node position for running prediction, like what's in the old code (check out linked issue above).  Then we need to re-enable column splitting in `DMatrix`.  Lastly we need to fix the `distcol`.

I'm not eliminating the possibility of spiting data by column in the future, just suggesting we remove the current `distcol`.   From above note, restoring the support might be best done from scratch to fit the more important updaters like histogram based algorithms.

@RAMitchell @hcho3

trivialfis:
Closes #5498.


src/learner.cc
```diff
@@ -888,8 +888,6 @@ class LearnerImpl : public LearnerIO {
       CHECK(tparam_.dsplit != DataSplitMode::kAuto)
         << "Precondition violated; dsplit cannot be 'auto' in distributed mode";
       if (tparam_.dsplit == DataSplitMode::kCol) {
```
RAMitchell, line 890: Can we deprecate `tparam_.dsplit`?

trivialfis: For now I'm keeping it here to include possible data splitting scheme in the future.  Let's think about removing it after benchmark for distributed algorithm is done.

PR #5511, opened 2020-04-10
Obtain num boosted rounds directly from booster.

trivialfis:
The PR tries to remove the hack for obtaining number of boosted rounds in Python.  I implemented the getter in c++ which directly counts the number of trees.

Most of the tests are already in `tests/python/test_training_continuation.py`, where `best_ntree_limit` is heavily tested.

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -356,6 +356,15 @@ XGB_DLL int XGBoosterCreate(const DMatrixHandle dmats[],
 XGB_DLL int XGBoosterFree(BoosterHandle handle);
 
 /*!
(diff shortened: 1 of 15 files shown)


trivialfis:
@RAMitchell

>  Can we get this information in another way?

 Right now we are using `len(get_dump())` and pick the parameters like `num_parallel_tree` and `num_class` in Python/R to calculate boosted rounds.

> I'm reluctant to create an API function for a single scalar variable unless there is no alternative.

As #5460 being sorted out,  I am thinking we give a `shape` attribute to the booster, like how many trees, how many output for each tree, how many groups, size of random forest etc ...  Along the way we create a friendly API for specifying these attributes and remove related configuration in `Learner`.

PR #5512, opened 2020-04-10
[dask] dask cudf inplace prediction.

### LINKED ISSUE #5207: Feature request: allow predict to output on GPU device memory

### LINKED ISSUE #5425: [FEA] get `dask_cudf.Series` from xgb.dask.predict()
**Is your feature request related to a problem? Please describe.**

I would like to obtain prediction result from  `prediction = xgb.dask.predict(client, output, dtrain)` as `dask_cudf.Series` instead of `dask.array.core.Array`.  I'd like to keep prediction results on GPU, as I do prediction on large number of records.

I am using, rapids 0.13 nightly in a conda env, and Dask 2.12.0.  Here is  a min rep example code:
```
import xgboost as xgb
import cudf, dask_cudf

cdf = cudf.DataFrame()
cdf['day'] = [15, 10, 20, 20,  21, 25, 28, 29]
cdf['hour'] = [19, 20, 20, 21, 18, 12, 15, 13]
cdf['passenger_count'] = [1, 1, 2, 2, 3, 3, 4, 2]
cdf['fare_amount'] = [5.0, 3.5, 12.5, 4.5, 9.0, 5.0, 3.5, 7.5]
ddf=dask_cudf.from_cudf(cdf, npartitions=2)

#prepare dtrain and call xgboost 
....
dtrain = xgb.dask.DaskDMatrix(client, X_train, Y_train)
trained_model = xgb.dask.train(client, params, dtrain,
                        num_boost_round=100, evals=[(dtrain, 'train')])
#prepare dtest
....
dtest = xgb.dask.DaskDMatrix(client, X_test, Y_test)

#do prediction
prediction = xgb.dask.predict(client, trained_model['booster'], dtest)

type(prediction)
dask.array.core.Array
```
**Describe the solution you'd like**
Keep prediction results on GPU when doing prediction on large number of records.
```
prediction = xgb.dask.predict(client, output, dtrain)
type(prediction)
dask_cudf.core.Series
```

trivialfis:
Now data returns on GPU.

Closes #5207 .  Closes #5425 .


tests/ci_build/Dockerfile.gpu
```diff
@@ -18,7 +18,7 @@ ENV PATH=/opt/python/bin:$PATH
 RUN \
     pip install numpy pytest scipy scikit-learn pandas matplotlib wheel kubernetes urllib3 graphviz && \
     pip install "dask[complete]" && \
-    conda install -c rapidsai -c nvidia -c numba -c conda-forge -c anaconda dask-cuda
+    conda install -c rapidsai -c conda-forge -c anaconda dask-cuda
```
hcho3, line 21: I don't think this line is necessary for running your test.

hcho3: Never mind. I found issues with the dockerfile that was causing Docker images to fail.

hcho3:
I managed to fix the Docker builds. It appears that mixing pip and conda is not a good idea, since that can mess up with dependency resolution. Both CUDF and GPU containers now use Conda exclusively, since `dask-cuda` package is not on PyPI.

Also I upgrade Miniconda to the latest version.

tests/ci_build/Dockerfile.cpu
```diff
@@ -29,7 +29,7 @@ RUN conda create -n py35 python=3.5 && \
 RUN \
     pip install pyyaml cpplint pylint astroid sphinx numpy scipy pandas matplotlib sh \
     		recommonmark guzzle_sphinx_theme mock breathe graphviz \
-		pytest scikit-learn wheel kubernetes urllib3 jsonschema boto3 && \
+		pytest scikit-learn wheel python-kubernetes urllib3 jsonschema boto3 && \
```
hcho3, line 32: Should use `kubernetes` with pip.

trivialfis: Oops, my bad.  Also what's the correct way of install matching cuda with conda?
```
[2020-04-14T09:03:59.184Z] PackagesNotFoundError: The following packages are not available from current channels:

[2020-04-14T09:03:59.184Z] 

[2020-04-14T09:03:59.184Z]   - cudatoolkit=9.0.176
```

hcho3: We can actually remove `cudatoolkit` from `Dockerfile.gpu`, since we don't use cuDF there. CuDF requires CUDA 10.0+, hence the error.

hcho3: Okay, I just ran Docker build on my Linux machine, and finally `test-python-gpu-cuda9.0` build is fixed.

trivialfis: Great!  I think cudatoolkit is a dependency of cudf and can be automatically installed?

hcho3: Let's keep it in the Dockerfile, since the official doc (https://rapids.ai/start.html) explicitly includes `cudatoolkit`.

trivialfis: Got it.  Thanks for all the help!

hcho3:
In the future when we drop CUDA 9.0, we can merge `Dockerfile.gpu` and `Dockerfile.cudf` and make CI simpler.

trivialfis:
@hcho3  It would be great if we can merge latest GPU test (like 10.1 currently) with `mgpu` test.

hcho3:
@trivialfis MGPU tests are already run with CUDA 10.1. The test suite is named `test-python-mgpu-cuda10.1`.

RAMitchell:
Dropping cuda 9.0 should be no problem, maybe at 1.2?

trivialfis:
@hcho3 That's a good news.  Do you think it's a good idea to merge two test suites?  I find it error prone that I need to check for skipped tests on CI.  If the `cuda 10.1` test suite can run all GPU tests then it might simplify things a bit.

trivialfis:
@RAMitchell 

> Dropping cuda 9.0 should be no problem, maybe at 1.2?

Sounds good to me.

hcho3:
@trivialfis Only tricky part is the provision of EC2 instances. Right now, the suite `test-python-gpu-cuda10.1` uses g4dn.xlarge instance (1 GPU) whereas the suite `test-python-mgpu-cuda10.1` uses g4dn.2xlarge instance (2 GPUs).

trivialfis:
@hcho3  Got it, thanks for the explanation.

hcho3:
Also, let's consider adding CUDA 10.2 to CI, now that cuDF supports CUDA 10.2.

trivialfis:
@hcho3 

> Also, let's consider adding CUDA 10.2 to CI, now that cuDF supports CUDA 10.2.

Let's pin it to 1.2.

tests/ci_build/Dockerfile.gpu
```diff
@@ -9,16 +9,16 @@ RUN \
...
 ENV PATH=/opt/python/bin:$PATH
 
 # Install Python packages
 RUN \
-    pip install numpy pytest scipy scikit-learn pandas matplotlib wheel kubernetes urllib3 graphviz && \
-    pip install "dask[complete]" && \
-    conda install -c rapidsai -c nvidia -c numba -c conda-forge -c anaconda dask-cuda
+    conda create -n gpu_test -c rapidsai -c nvidia -c conda-forge -c defaults \
+        python=3.7 dask dask-cuda numpy pytest scipy scikit-learn pandas \
+        matplotlib wheel python-kubernetes urllib3 graphviz dask-cudf cudf cupy
```
hcho3, line 21: Please do not install cuDF in Dockerfile.gpu. CuDF is not available for CUDA 9.0, so it will break test-python-gpu-cuda9.0.

trivialfis:
I have reverted the installation for cudf.  But the integration tests must be run on mgpu worker with both cudf and dask installed.  So right now they are not running on CI.

I also changed memory copying with cupy from using raw pointer into using some simple wrappers from cupy.  The end result is if there's a memory error, we can safely narrow down the issue to cupy internal.  As cupy is much better in memory management task so it will make our lives easier in the future.

hcho3:
> But the integration tests must be run on mgpu worker with both cudf and dask installed. So right now they are not running on CI.

I am adding them back to CI, by creating another test suite `mgpu-cudf`. It's kind of ugly, but we have to live with it until we drop CUDA 9.0.

hcho3:
@trivialfis The new multi-GPU tests are now functional: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-5512/19/pipeline/214. Finally :)

trivialfis:
@hcho3 Thanks for the help and demonstrating the process of changing CI pipeline.  I'm learning so that I don't have to trouble you for small changes. ;-)

PR #5513, opened 2020-04-11
Remove makefiles.

trivialfis:
Another attempt after https://github.com/dmlc/xgboost/pull/4733 .


hcho3:
This should be fine, since the R package has its own Makefile under R-package directory (it’s called Makevars.in).

trivialfis:
@hcho3 Yup, tests are passing.  I left a few targets in the makefile, but building XGBoost with make should be completely removed now.

PR #5514, opened 2020-04-11
Ensure that configured dmlc/build_config.h is picked up by Rabit and XGBoost

### LINKED ISSUE #5510: Build failure on musl (CMAKE)
Build environment: `alpine` [docker container](https://hub.docker.com/_/alpine)
XGBoost Version: 1.0.2 and master (works on 0.82)

Build Script:
```
git submodule init; git submodule update
mkdir build; cd build
cmake ..
build
```

Result: 
```
[  1%] Building CXX object rabit/CMakeFiles/rabit.dir/src/allreduce_base.cc.o
In file included from /xgboost/rabit/../dmlc-core/include/dmlc/././base.h:300,
                 from /xgboost/rabit/../dmlc-core/include/dmlc/./logging.h:16,
                 from /xgboost/rabit/../dmlc-core/include/dmlc/io.h:15,
                 from /xgboost/rabit/include/rabit/internal/utils.h:16,
                 from /xgboost/rabit/src/allreduce_base.h:18,
                 from /xgboost/rabit/src/allreduce_base.cc:14:
/xgboost/rabit/../dmlc-core/include/dmlc/build_config_default.h:31:27: fatal error: execinfo.h: No such file or director
y
   31 |   #define DMLC_EXECINFO_H <execinfo.h>
      |                           ^
compilation terminated.
```

The actual error (`execinfo`) is a bit of red herring. Notice that the file being included is `/build_config_default.h`. That file should not be included here.

https://github.com/dmlc/dmlc-core/blob/981b1c32f91668e669ee376856f92f36cfd2a351/include/dmlc/base.h#L297-L301

So going by the source above, `DMLC_CORE_USE_CMAKE` should be defined, but it is not. 

To show that this is the problem, the build does succeed if I set the following on the shell before calling cmake and make, however that is only a very hacky workaround:

`export CXXFLAGS="-DDMLC_CORE_USE_CMAKE -I/xgboost/build/dmlc-core/include"`

So to further troubleshoot this, I ran the same cmake build on a glibc system. To trigger the error, I added the following line to the top of `dmlc-core/include/dmlc/build_config_default.h`

```
#error "Should never load this build_config_default"
```

And predictably, the build fails with 

```
[  1%] Building CXX object rabit/CMakeFiles/rabit.dir/src/allreduce_base.cc.o
cd /workspace/srcdir/xgboost/build/rabit && /opt/bin/c++   -I/workspace/srcdir/xgboost/rabit/include -I/workspace/srcdir/xgboost/rabit/../dmlc-core/include  -O3 -DNDEBUG -fPIC   -std=gnu++11 -o CMakeFiles/rabit.dir/src/allreduce_base.cc.o -c /workspace/srcdir/xgboost/rabit/src/allreduce_base.cc
In file included from /workspace/srcdir/xgboost/rabit/../dmlc-core/include/dmlc/././base.h:300,
                 from /workspace/srcdir/xgboost/rabit/../dmlc-core/include/dmlc/./logging.h:16,
                 from /workspace/srcdir/xgboost/rabit/../dmlc-core/include/dmlc/io.h:15,
                 from /workspace/srcdir/xgboost/rabit/include/rabit/internal/utils.h:16,
                 from /workspace/srcdir/xgboost/rabit/src/allreduce_base.h:18,
                 from /workspace/srcdir/xgboost/rabit/src/allreduce_base.cc:14:
/workspace/srcdir/xgboost/rabit/../dmlc-core/include/dmlc/build_config_default.h:10:2: error: #error "Should never load
this build_config_default"
 #error "Should never load this build_config_default"
  ^~~~~
```

So, in summary, I think the cmake created configuration is not used by the xboost build currently. The builds succeed in a linux/glibc system since the default settings work in that environment, but using `musl` makes this issue apparent.

hcho3:
When we run `cmake ..` in a build directory, CMake generates a header `build/dmlc-core/dmlc/build_config.h` ([see here](https://github.com/dmlc/dmlc-core/blob/5df8305fe699d3b503d10c60a231ab0223142407/CMakeLists.txt#L161); note that the header is inside the build directory). **Right now, only `dmlc` target uses this generated header. Neither `rabit` nor `objxgboost` uses this header.** 

Instead, `rabit` and `objxgboost` targets use [`dmlc-core/include/dmlc/build_config_default.h`](https://github.com/dmlc/dmlc-core/blob/5df8305fe699d3b503d10c60a231ab0223142407/include/dmlc/build_config_default.h). This header was meant to be a fallback, to be used only when CMake is unavailable. The fallback header is good enough for common platforms like Linux but not so good for uncommon platforms such as Alpine Linux. See #5510.

Fix:
* Explicitly declare `dmlc` as a dependency for the `rabit` and `objxgboost` targets. This way, all relevant flags get propagated from `dmlc` to the targets, including:
  1. All header paths for `dmlc`, including `build/dmlc-core/dmlc/build_config.h`;
  2. Macros for `dmlc`, including `DMLC_CORE_USE_CMAKE`.
* Upgrade minimum CMake version to 3.13. See https://cmake.org/cmake/help/v3.13/policy/CMP0079.html.
* Add a test in CI where we attempt to compile XGBoost with the fallback header removed.
* Remove unit tests for R 3.4.4 from CI to conserve resources.

Closes #5510.


hcho3:
I really don't like this fix. Ideally, I'd like to revise CMakeLists.txt of Rabit to link DMLC explicitly, to indicate that whoever that consumes Rabit also must consume DMLC too.

hcho3:
CI is fixed in #5506

PR #5522, opened 2020-04-13
Optimizations for RNG in InitData kernel

ShvetsKS:
This PR contains optimizations for case when `subsample` parameter less than `1`.
In big data sets, when `subsample < 1` `InitData` kernel requires ~ 25% of full training computational time.

### DIFF
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -536,6 +536,63 @@ bool QuantileHistMaker::Builder::UpdatePredictionCache(
   return true;
 }
 
+void QuantileHistMaker::Builder::InitSampling(const std::vector<GradientPair>& gpair,
+                                                const DMatrix& fmat,
+                                                std::vector<size_t>* row_indices) {
+  const auto& info = fmat.Info();
+  auto& rnd = common::GlobalRandom();
+  std::vector<size_t>& row_indices_local = *row_indices;
+  size_t* p_row_indices = row_indices_local.data();
+#if XGBOOST_CUSTOMIZE_GLOBAL_PRNG
+  std::bernoulli_distribution coin_flip(param_.subsample);
+  size_t j = 0;
+  for (size_t i = 0; i < info.num_row_; ++i) {
+    if (gpair[i].GetHess() >= 0.0f && coin_flip(rnd)) {
+      p_row_indices[j++] = i;
+    }
+  }
+  /* resize row_indices to reduce memory */
+  row_indices_local.resize(j);
+#else
+  const size_t nthread = this->nthread_;
+  std::vector<size_t> row_offsets(nthread, 0);
+  /* usage of mt19937_64 give 2x speed up for subsampling */
+  std::vector<std::mt19937> rnds(nthread);
+  /* create engine for each thread */
+  for (std::mt19937& r : rnds) {
+    r = rnd;
+  }
+  const size_t discard_size = info.num_row_ / nthread;
+  #pragma omp parallel num_threads(nthread)
+  {
+    const size_t tid = omp_get_thread_num();
+    const size_t ibegin = tid * discard_size;
+    const size_t iend = (tid == (nthread - 1)) ?
+                        info.num_row_ : ibegin + discard_size;
+    std::bernoulli_distribution coin_flip(param_.subsample);
+
+    rnds[tid].discard(2*discard_size * tid);
+    for (size_t i = ibegin; i < iend; ++i) {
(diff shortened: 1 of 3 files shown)


ShvetsKS:
With this PR we can see  2x performance gain for InitData kernel. 

santander-190k (subsample = 0.5)| InitData | Full Training Time
-- | -- |--
Master | 47.24 | 179.71
This PR | 23.43 | 160.46

airline-ohe-1m (subsample = 0.5)| InitData
-- | -- 
Master | 24.5 
This PR | 11.68 

The sequence of random numbers was not changed at all.

**Note:** if we use `std::mt19937_64`  RNG engine we will see additional 2x speed up, and the accuracy will not change in this case (but sequence of random numbers will be different of course).

src/tree/updater_quantile_hist.cc
```diff
@@ -574,15 +574,55 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
       CHECK_EQ(param_.sampling_method, TrainParam::kUniform)
         << "Only uniform sampling is supported, "
         << "gradient-based sampling is only support by GPU Hist.";
-      std::bernoulli_distribution coin_flip(param_.subsample);
       auto& rnd = common::GlobalRandom();
+#if XGBOOST_CUSTOMIZE_GLOBAL_PRNG
```
SmirnovEgorRu, line 578: Could you, please, explain the reason to add the define?

src/tree/updater_quantile_hist.cc
```diff
@@ -574,15 +574,55 @@ void QuantileHistMaker::Builder::InitData(const GHistIndexMatrix& gmat,
...
+      const size_t nthread = this->nthread_;
+      std::vector<size_t> row_offsets(nthread, 0);
+      /* usage of mt19937_64 give 2x speed up for subsampling */
+      std::vector<std::mt19937> rnds(nthread);
+      /* create engine for each thread */
+      for (std::mt19937& r : rnds) {
+        r = rnd;
+      }
+      const size_t discard_size = info.num_row_ / nthread;
+      #pragma omp parallel num_threads(nthread)
```
SmirnovEgorRu, line 597: Overall LGTM, but just to improve - better to move this piece of code to a separate function. Because InitData becomes massive, especially with introducing conditional building.

Maybe it's paranoia, but did you check single-thread perf? Is it the same as before?

src/tree/updater_quantile_hist.cc
ShvetsKS: Introduced optimizations are applicable only  for std rng engines, but if `XGBOOST_CUSTOMIZE_GLOBAL_PRNG` was defined internal `CustomGlobalRandomEngine` is used (see https://github.com/dmlc/xgboost/blob/master/src/common/random.h#L37).
`CustomGlobalRandomEngine` doesn't have implementation of `discard` method.

src/tree/updater_quantile_hist.cc
ShvetsKS: Thanks. Fixed.
single-thread was checked for santander benchmark (but with 1000 n_iter).
309.4s vs 309.3s
And there should not be any changes for single-thread as seems no additional work was introduced :)

ShvetsKS:
@trivialfis yes, it was tested manually already, could you give some example where similar checks were already implemented (run with different number of OMP threads, as I select `nthreads` via `omp_get_num_threads()`) ?

SmirnovEgorRu:
@ShvetsKS, potenyially you can extend https://github.com/dmlc/xgboost/blob/master/tests/python/test_openmp.py this one to use subsampling.

trivialfis:
`omp_set_num_threads` maybe useful?  Also @SmirnovEgorRu some unittests for histogram utilities are failing if you set threads to 1.

trivialfis:
Let me reproduce it after fix building without omp.

trivialfis:
The error I mentioned is not related to this PR.  Looks good to me if the test can be added.

ShvetsKS:
@trivialfis , @SmirnovEgorRu  tests were added.

ShvetsKS:
Seems the reason of failed testing stages:
- Jenkins Linux: Test 
- Jenkins Linux: Deploy 

is not related to changes in this PR.
@trivialfis Should we restart it?

hcho3:
I've restarted tests.

hcho3:
#5537 should fix the failing tests.

trivialfis:
@SmirnovEgorRu Feel free to merge. ;-)

PR #5533, opened 2020-04-14
[jvm-packages] [CI] Create a Maven repository to host SNAPSHOT JARs

### LINKED ISSUE #4752: [jvm-packages] Jenkins CI should build snapshot JARs
The Jenkins CI (https://xgboost-ci.net) should build snapshot JARs for XGBoost4J and XGBoost4J-Spark.

* This simplifies our release process, since we just need to copy pre-built JARs.
* Adventurous users can download nightly builds, to use latest features and bug fixes.
* We can use Docker images with old Linux (CentOS 6) so that the JAR would use old version of GLIBC and LIBGOMP. I've seen many people try building XGBoost JAR themselves and see it fail on the cluster because the cluster has old GLIBC (and they don't have admin right to upgrade).

So in short, we should do what we are currently doing with the Python package: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/master/321/artifacts

@CodingCat @trams

### LINKED ISSUE #5494: [jvm-package] xgboost >1 not published to maven, or anywhere else
Hey team,

xgboost jars should be easily accessible for all, but they are not.

Maven central https://mvnrepository.com/artifact/ml.dmlc/xgboost4j only has up to version 0.90

The snapshot repo @ https://raw.githubusercontent.com/CodingCat/xgboost/maven-repo/  is broken as well

https://github.com/CodingCat/xgboost/tree/maven-repo/ml/dmlc/xgboost4j doesn't list 1.0 either

hcho3:
Closes #4752.
Closes #5494.

Jenkins will now deploy freshly made JAR packages to a public S3 bucket `xgboost-maven-repo`.  Users can now access SNAPSHOT JARs by adding
```xml
<repository>
  <id>XGBoost4J Snapshot Repo</id>
  <name>XGBoost4J Snapshot Repo</name>
  <url>https://s3-us-west-2.amazonaws.com/xgboost-maven-repo/snapshot/</url>
</repository>
```
to their pom.xml.

[Updated installation instructions](https://xgboost.readthedocs.io/en/release_1.0.0/jvm/index.html#installation-from-maven-repository)
[File listing of `xgboost-maven-repo`](https://s3-us-west-2.amazonaws.com/xgboost-maven-repo/list.html)

Pinging @trams @CodingCat @trivialfis @sperlingxx @wbo4958 @WeichenXu123 @chenqin who might be interested.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -94,6 +94,17 @@ pipeline {
         milestone ordinal: 4
       }
     }
+    stage('Jenkins Linux: Deploy') {
+      agent none
+      steps {
+        script {
+          parallel ([
+            'deploy-jvm-packages': { DeployJVMPackages(spark_version: '2.4.3') }
+          ])
+        }
+        milestone ordinal: 5
+      }
+    }
   }
 }
 
@@ -380,6 +391,21 @@ def TestR(args) {
     sh """
     ${dockerRun} ${container_type} ${docker_binary} ${docker_args} tests/ci_build/build_test_rpkg.sh || tests/ci_build/print_r_stacktrace.sh
     """
+    deleteDir()
+  }
+}
+
+def DeployJVMPackages(args) {
+  node('linux && cpu') {
+    unstash name: 'srcs'
+    if (env.BRANCH_NAME == 'master' || env.BRANCH_NAME.startsWith('release')) {
+      echo 'Deploying to xgboost-maven-repo S3 repo...'
+      def container_type = "jvm"
+      def docker_binary = "docker"
+      sh """
+      ${dockerRun} ${container_type} ${docker_binary} tests/ci_build/deploy_jvm_packages.sh ${args.spark_version}
+      """
+    }
     deleteDir()
   }
 }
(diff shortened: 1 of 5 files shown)


trams:
This is amazing. @alois-bissuel might be interested

hcho3:
@trams Do people still use Scala 2.11? I’ve had a use ask me to upload XGBoost4J artifact compiled with Scala 2.11. (The one that’s available was compiled with Scala 2.12.)

trams:
My current and my previous (Criteo) employees are using Scala 2.11
So I would it is still quite common

From my opinion if it not too much trouble it would be nice to provide 1.0 version with scala 2.11

If it is too much trouble it is reasonable to drop support. After all there is more or less common rule to support two last versions of scala only (as far as I know 2.13 has been released)

hcho3:
@trams Thanks. I went ahead and uploaded 1.0.0 with Scala 2.11. I think we can drop Scala 2.11 once we adopt Spark 3.0, which drops support for 2.11.

trams:
Great news!
And I agree there is no point to support 2.11 when you depend on Spark 3 which does not support 2.11

Issue #5401, opened 2020-03-09
NaN values and Scikit-Learn RFECV


benjaminvdb:
I couldn't find the issue associated with [this XGBoost forum topic](https://discuss.xgboost.ai/t/nan-values-and-scikit-learn-rfecv/1432), so I assume none was created. I can confirm this problem persists with the latest nightly of xgboost (`a38e7bd19c461e0bed7bd96ec72d56132157d4af`) and scikit-learn (`018c6dc57d21c89c7d1278c686c7d5d62f32ee48`).

I agree with Mike Creeth's statement in the previously mentioned forum post:

> I believe this is because RFECV does some checking based on the tags that it gets from the estimator. It uses the tag ‘allow_nan’ to determine whether or not to check X for NaN values. It seems that currently XGBoost simply inherits the default “allow_nan” tag value from the scikit-learn estimator class, which is False. As XGB does in fact handle null values in X, I believe this behavior is incorrect.

```
from xgboost import XGBClassifier
from sklearn.feature_selection import RFECV

estimator = XGBClassifier()
selector = RFECV(estimator, cv=3)
selector = selector.fit(X, y)
```

with `X` having one or more `np.nan` values, raises the following error:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-35-6d04ec0892c9> in <module>
     18 
     19 selector = RFECV(model, cv=3)#, scoring=neg_rmse)
---> 20 selector = selector.fit(X_train.values, y_train.values)

/local/burghbvander/miniconda3/envs/fastai/lib/python3.7/site-packages/sklearn/feature_selection/_rfe.py in fit(self, X, y, groups)
    498             X, y, accept_sparse="csr", ensure_min_features=2,
    499             force_all_finite=not tags.get('allow_nan', True),
--> 500             multi_output=True
    501         )
    502 

/local/burghbvander/miniconda3/envs/fastai/lib/python3.7/site-packages/sklearn/base.py in _validate_data(self, X, y, reset, **check_params)
    404             out = X
    405         else:
--> 406             X, y = check_X_y(X, y, **check_params)
    407             out = X, y
    408 

[...]

PR #5538, opened 2020-04-15
Fix skl nan tag.

trivialfis:
Closes https://github.com/dmlc/xgboost/issues/5401 .


trivialfis:
Weird that the Python test passes on Linux but fails on Windows ...

Jenkinsfile-win64
```diff
@@ -131,6 +135,10 @@ def TestWin64GPU(args) {
     bat """
     conda activate && for /R %%i in (python-package\\dist\\*.whl) DO python -m pip install "%%i"
     """
+    echo "Install Python dependencies..."
```
trivialfis, line 138: `Installing`.

PR #5557, opened 2020-04-17
Distributed optimizations

ShvetsKS:
This PR consists changes to optimize distributed mode of training.

1) depthwise changes: Possibility to build 'smaller' nodes was introduced for distributed mode (before this only left nodes were build)  for more balanced work. Than 'subtraction trick' is applied for 'larger' node.
2) lossguide changes: similar to above.
3) number of rabit::Allreduce calls is the same as before due to structured histograms adding. 
4) Timers were added to track most time consuming parts for distributed also.

**mortgage 3.5Gb** 
6 workers x 8 threads | loss | depth
-- | -- | --
Master | 76.6 | 68.5
This PR | 48.7 | 44

**higgs 10m,  2.7Gb** 
running line: python .../dmlc-core/tracker/dmlc-submit --cluster=local --num-workers=4 --worker-cores=12 python distr_higgs10m.py

4 workers x 12 threads | depth
-- | --
Master | 877.83
This PR | 223.66

### DIFF
--- a/src/tree/updater_prune.cc
+++ b/src/tree/updater_prune.cc
@@ -14,7 +14,7 @@
 #include "xgboost/json.h"
 #include "./param.h"
 #include "../common/io.h"
-
+#include "../common/timer.h"
 namespace xgboost {
 namespace tree {
 
@@ -25,6 +25,7 @@ class TreePruner: public TreeUpdater {
  public:
   TreePruner() {
     syncher_.reset(TreeUpdater::Create("sync", tparam_));
+    pruner_monitor_.Init("TreePruner");
   }
   char const* Name() const override {
     return "prune";
@@ -52,6 +53,7 @@ class TreePruner: public TreeUpdater {
   void Update(HostDeviceVector<GradientPair> *gpair,
               DMatrix *p_fmat,
               const std::vector<RegTree*> &trees) override {
+    pruner_monitor_.Start("PrunerUpdate");
     // rescale learning rate according to size of trees
     float lr = param_.learning_rate;
     param_.learning_rate = lr / trees.size();
@@ -60,6 +62,7 @@ class TreePruner: public TreeUpdater {
     }
     param_.learning_rate = lr;
     syncher_->Update(gpair, p_fmat, trees);
+    pruner_monitor_.Stop("PrunerUpdate");
   }
 
  private:
@@ -105,6 +108,7 @@ class TreePruner: public TreeUpdater {
   std::unique_ptr<TreeUpdater> syncher_;
   // training parameter
   TrainParam param_;
+  common::Monitor pruner_monitor_;
 };
 
 XGBOOST_REGISTER_TREE_UPDATER(TreePruner, "prune")
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -102,40 +102,82 @@ void QuantileHistMaker::Builder::SyncHistograms(
   builder_monitor_.Start("SyncHistograms");
 
   const bool isDistributed = rabit::IsDistributed();
-
   const size_t nbins = hist_builder_.GetNumBins();
-  common::BlockedSpace2d space(nodes_for_explicit_hist_build_.size(), [&](size_t node) {
-    return nbins;
-  }, 1024);
+  if (!isDistributed) {
+    common::BlockedSpace2d space(nodes_for_explicit_hist_build_.size(), [&](size_t node) {
+      return nbins;
+    }, 1024);
 
-  common::ParallelFor2d(space, this->nthread_, [&](size_t node, common::Range1d r) {
-    const auto entry = nodes_for_explicit_hist_build_[node];
-    auto this_hist = hist_[entry.nid];
-    // Merging histograms from each thread into once
-    hist_buffer_.ReduceHist(node, r.begin(), r.end());
+    common::ParallelFor2d(space, this->nthread_, [&](size_t node, common::Range1d r) {
+      const auto entry = nodes_for_explicit_hist_build_[node];
+      auto this_hist = hist_[entry.nid];
+      // Merging histograms from each thread into once
+      hist_buffer_.ReduceHist(node, r.begin(), r.end());
 
-    // Store posible parent node
-    auto this_local = phist_local_[entry.nid];
-    CopyHist(this_local, this_hist, r.begin(), r.end());
+      if (!(*p_tree)[entry.nid].IsRoot() && entry.sibling_nid > -1 /*&& !isDistributed*/) {
+        auto parent_hist = hist_[(*p_tree)[entry.nid].Parent()];
+        auto sibling_hist = hist_[entry.sibling_nid];
+        SubtractionHist(sibling_hist, parent_hist, this_hist, r.begin(), r.end());
+      }
+    });
+  } else {
+    common::BlockedSpace2d space(nodes_for_explicit_hist_build_.size(), [&](size_t node) {
+      return nbins;
+    }, 1024);
 
-    if (!(*p_tree)[entry.nid].IsRoot() && entry.sibling_nid > -1 && !isDistributed) {
-      auto parent_hist = hist_[(*p_tree)[entry.nid].Parent()];
-      auto sibling_hist = hist_[entry.sibling_nid];
-      SubtractionHist(sibling_hist, parent_hist, this_hist, r.begin(), r.end());
-    } else if (!(*p_tree)[entry.nid].IsRoot() && entry.sibling_nid > -1) {
-      auto sibling_hist = hist_[entry.sibling_nid];
-      auto parent_hist = phist_local_[(*p_tree)[entry.nid].Parent()];
-      SubtractionHist(sibling_hist, parent_hist, this_hist, r.begin(), r.end());
+    common::ParallelFor2d(space, this->nthread_, [&](size_t node, common::Range1d r) {
+      const auto entry = nodes_for_explicit_hist_build_[node];
+      auto this_hist = hist_[entry.nid];
+      // Merging histograms from each thread into once
+      hist_buffer_.ReduceHist(node, r.begin(), r.end());
+
       // Store posible parent node
-      auto sibling_local = phist_local_[entry.sibling_nid];
-      CopyHist(sibling_local, sibling_hist, r.begin(), r.end());
-    }
-  });
-  if (isDistributed) {
+      auto this_local = phist_local_[entry.nid];
+      CopyHist(this_local, this_hist, r.begin(), r.end());
+
+      if (!(*p_tree)[entry.nid].IsRoot() && entry.sibling_nid > -1) {
+        auto sibling_hist = hist_[entry.sibling_nid];
+        auto parent_hist = phist_local_[(*p_tree)[entry.nid].Parent()];
+        SubtractionHist(sibling_hist, parent_hist, this_hist, r.begin(), r.end());
+        // Store posible parent node
+        auto sibling_local = phist_local_[entry.sibling_nid];
+        CopyHist(sibling_local, sibling_hist, r.begin(), r.end());
+      }
+    });
+
     builder_monitor_.Start("SyncHistograms:histred_.Allreduce");
-    int m =  nodes_for_explicit_hist_build_.size() + nodes_for_subtraction_trick_.size();
-    this->histred_.Allreduce(hist_[starting_index].data(), hist_builder_.GetNumBins() * m);
+    this->histred_.Allreduce(hist_[starting_index].data(), hist_builder_.GetNumBins() * sync_count);
     builder_monitor_.Stop("SyncHistograms:histred_.Allreduce");
+
+    common::ParallelFor2d(space, this->nthread_, [&](size_t node, common::Range1d r) {
+      const auto entry = nodes_for_explicit_hist_build_[node];
+      if (!((*p_tree)[entry.nid].IsLeftChild())) {
+        auto this_hist = hist_[entry.nid];
+
+        if (!(*p_tree)[entry.nid].IsRoot() && entry.sibling_nid > -1) {
+          auto parent_hist = hist_[(*p_tree)[entry.nid].Parent()];
+          auto sibling_hist = hist_[entry.sibling_nid];
+          SubtractionHist(this_hist, parent_hist, sibling_hist, r.begin(), r.end());
+        }
+      }
+    });
+
+    common::BlockedSpace2d space2(nodes_for_subtraction_trick_.size(), [&](size_t node) {
+      return nbins;
+    }, 1024);
+
+    common::ParallelFor2d(space2, this->nthread_, [&](size_t node, common::Range1d r) {
+      const auto entry = nodes_for_subtraction_trick_[node];
+      if (!((*p_tree)[entry.nid].IsLeftChild())) {
+        auto this_hist = hist_[entry.nid];
+
+        if (!(*p_tree)[entry.nid].IsRoot() && entry.sibling_nid > -1) {
+          auto parent_hist = hist_[(*p_tree)[entry.nid].Parent()];
+          auto sibling_hist = hist_[entry.sibling_nid];
+          SubtractionHist(this_hist, parent_hist, sibling_hist, r.begin(), r.end());
+        }
+      }
+    });
   }
 
   builder_monitor_.Stop("SyncHistograms");
@@ -159,15 +201,17 @@ void QuantileHistMaker::Builder::BuildHistogramsLossGuide(
   int starting_index = std::numeric_limits<int>::max();
   int sync_count = 0;
 
-  AddHistRows(&starting_index, &sync_count);
+  AddHistRows(&starting_index, &sync_count, p_tree);
   BuildLocalHistograms(gmat, gmatb, p_tree, gpair_h);
   SyncHistograms(starting_index, sync_count, p_tree);
 }
 
 
-void QuantileHistMaker::Builder::AddHistRows(int *starting_index, int *sync_count) {
+void QuantileHistMaker::Builder::AddHistRows(int *starting_index, int *sync_count,
+                                             RegTree *p_tree) {
   builder_monitor_.Start("AddHistRows");
-  std::vector<size_t> merged_hist(nodes_for_explicit_hist_build_.size() +
+
+  std::vector<int> merged_hist(nodes_for_explicit_hist_build_.size() +
                                   nodes_for_subtraction_trick_.size());
   for (size_t i = 0; i < nodes_for_explicit_hist_build_.size(); ++i) {
     merged_hist[i] = nodes_for_explicit_hist_build_[i].nid;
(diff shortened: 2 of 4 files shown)


thvasilo:
Hello @ShvetsKS could you provide some details as to what the plan is here?

SmirnovEgorRu:
@thvasilo, 
The changes are close to this PR https://github.com/dmlc/xgboost/pull/4824. Currently we have different behavior for single-node and distributed modes in terms of what histograms we compute explicitly by BuildHist function and what nodes are build by SubtractionTrick.

Now for single-node it's more optimal, because we select smaller set of rows for BuildHist function and SubtractionTrick for larger, it brings performance gain. For distributed mode we just use BuildHist for left siblings always, in general it shows worse performance.

The PR introduce similar logic to distributed mode.

src/tree/updater_quantile_hist.cc
```diff
@@ -424,15 +439,12 @@ void QuantileHistMaker::Builder::ExpandWithLossGuide(
       ExpandEntry right_node(cright, cleft, p_tree->GetDepth(cright),
                             0.0f, timestamp++);
 
-      if (rabit::IsDistributed()) {
-        // in distributed mode, we need to keep consistent across workers
+      size_t node_sizes[2] = { row_set_collection_[cleft].Size(),
+                               row_set_collection_[cright].Size() };
+      if (node_sizes[0] < node_sizes[1]) {
```
SmirnovEgorRu, line 444: I think now you can use

```cpp
if (row_set_collection_[cleft].Size() < row_set_collection_[cright].Size()) {
...
}
```

src/tree/updater_quantile_hist.cc
```diff
@@ -305,31 +323,28 @@ void QuantileHistMaker::Builder::SplitSiblings(const std::vector<ExpandEntry>& n
...
-        small_siblings->push_back(entry);
-      } else if (!node.IsRoot() && !node.IsLeftChild() &&
-                 (row_set_collection_[nid].Size() <=
-                  row_set_collection_[(*p_tree)[node.Parent()].LeftChild()].Size())) {
+      const int32_t left_id = (*p_tree)[node.Parent()].LeftChild();
+      const int32_t right_id = (*p_tree)[node.Parent()].RightChild();
+      size_t node_sizes[2] = { row_set_collection_[left_id ].Size(),
+                               row_set_collection_[right_id].Size() };
+
+      if (nid == left_id && node_sizes[0] < node_sizes[1]) {
```
SmirnovEgorRu, line 338: The same, now it can be simplified

src/tree/updater_quantile_hist.cc
```diff
@@ -160,18 +167,28 @@ void QuantileHistMaker::Builder::BuildHistogramsLossGuide(
...
-    int nid = entry.nid;
+  std::vector<size_t> merged_hist(nodes_for_explicit_hist_build_.size() +
+                                  nodes_for_subtraction_trick_.size());
+  for (size_t i = 0; i < nodes_for_explicit_hist_build_.size(); ++i) {
+    merged_hist[i] = nodes_for_explicit_hist_build_[i].nid;
+  }
+  for (size_t i = 0; i < nodes_for_subtraction_trick_.size(); ++i) {
+    merged_hist[nodes_for_explicit_hist_build_.size() + i] =
+    nodes_for_subtraction_trick_[i].nid;
+  }
```
SmirnovEgorRu, line 178: We can use ```qexpand_depth_wise_``` because it contains all nodes to build already

src/tree/updater_quantile_hist.cc
ShvetsKS: done

src/tree/updater_quantile_hist.cc
ShvetsKS: No such vector for `lossguide`. But `AddHistRows` is common for both methods.

src/tree/updater_quantile_hist.cc
ShvetsKS: done

trivialfis:
A genuine question, do you have difficulty to read the hist code after many optimizations (I see that you are really good at spotting optimization opportunity so I may be not as skilled)?  I see that specializing for each case is good for performance, but could you try to arrange the source code into finer modules and create a easier to follow structure?  Maybe some of these can help others to understand the code?

- policy based http://loki-lib.sourceforge.net/index.php?n=Main.Policy-basedDesign
- strategy https://en.wikipedia.org/wiki/Strategy_pattern

trivialfis:
Also, if you specialize over something, then there needs to be a new test case for your specialization.

src/tree/updater_quantile_hist.cc
```diff
@@ -102,33 +104,82 @@ void QuantileHistMaker::Builder::SyncHistograms(
...
+      if (!(*p_tree)[entry.nid].IsRoot() && entry.sibling_nid > -1 /*&& !isDistributed*/) {
+        auto parent_hist = hist_[(*p_tree)[entry.nid].Parent()];
+        auto sibling_hist = hist_[entry.sibling_nid];
+        SubtractionHist(sibling_hist, parent_hist, this_hist, r.begin(), r.end());
+      }
+    });
+  } else {
+    common::BlockedSpace2d space(nodes_for_explicit_hist_build_.size(), [&](size_t node) {
+      return nbins;
+    }, 1024);
```
SmirnovEgorRu, line 128: Space is the same with line 109. So you can move this out of "if - else"

src/tree/updater_quantile_hist.cc
```diff
@@ -102,33 +104,82 @@ void QuantileHistMaker::Builder::SyncHistograms(
...
+
+    common::ParallelFor2d(space, this->nthread_, [&](size_t node, common::Range1d r) {
+      const auto entry = nodes_for_explicit_hist_build_[node];
+      auto this_hist = hist_[entry.nid];
+      // Merging histograms from each thread into once
+      hist_buffer_.ReduceHist(node, r.begin(), r.end());
+
+      if (!(*p_tree)[entry.nid].IsRoot() && entry.sibling_nid > -1 /*&& !isDistributed*/) {
```
SmirnovEgorRu, line 119:
```
/*&& !isDistributed*/
```
Is extra thing here, I suppose

src/tree/updater_quantile_hist.cc
```diff
@@ -102,33 +104,82 @@ void QuantileHistMaker::Builder::SyncHistograms(
   builder_monitor_.Start("SyncHistograms");
 
   const bool isDistributed = rabit::IsDistributed();
-
   const size_t nbins = hist_builder_.GetNumBins();
-  common::BlockedSpace2d space(nodes_for_explicit_hist_build_.size(), [&](size_t node) {
-    return nbins;
-  }, 1024);
+  if (!isDistributed) {
```
SmirnovEgorRu, line 108: This "if-else" is tooo large.
Need to split this in different methods or classes or apply another refactoring approach.

src/tree/updater_quantile_hist.cc
```diff
@@ -102,33 +104,82 @@ void QuantileHistMaker::Builder::SyncHistograms(
...
-      SubtractionHist(sibling_hist, parent_hist, this_hist, r.begin(), r.end());
-    }
-  });
+        if (!(*p_tree)[entry.nid].IsRoot() && entry.sibling_nid > -1) {
+          auto parent_hist = hist_[(*p_tree)[entry.nid].Parent()];
+          auto sibling_hist = hist_[entry.sibling_nid];
+          SubtractionHist(this_hist, parent_hist, sibling_hist, r.begin(), r.end());
+        }
+      }
+    });
```
SmirnovEgorRu, line 165: At least it can be moved in a separate function and reused in several places

src/tree/updater_quantile_hist.cc
```diff
@@ -152,26 +203,51 @@ void QuantileHistMaker::Builder::BuildHistogramsLossGuide(
...
 
-  AddHistRows(&starting_index, &sync_count);
+  AddHistRows(&starting_index, &sync_count, p_tree);
   BuildLocalHistograms(gmat, gmatb, p_tree, gpair_h);
   SyncHistograms(starting_index, sync_count, p_tree);
 }
 
 
-void QuantileHistMaker::Builder::AddHistRows(int *starting_index, int *sync_count) {
+void QuantileHistMaker::Builder::AddHistRows(int *starting_index, int *sync_count,
```
SmirnovEgorRu, line 212: Unfortunately this function became massive and needed to be refactored.
I think we can add firstly left and secondary right nodes - in simpler manner.

For example, you don't need to create tmp vector and sort it. It's enough to just have a loop which handle all required things.

src/tree/updater_quantile_hist.cc
```diff
@@ -152,26 +203,51 @@ void QuantileHistMaker::Builder::BuildHistogramsLossGuide(
...
+        phist_local_.AddHistRow(nid);
+      }
+    }
   }
 
+  (*sync_count) = merged_hist.size() / 2;
+  if (*sync_count == 0) {
+    (*sync_count) = 1;
+  }
+  CHECK_EQ(n_left, (*sync_count));
```
SmirnovEgorRu, line 250: It's equal to # of left nodes or 1 for the root. So, it can be simplified.

SmirnovEgorRu: Also, I'm not sure that a condition in the check will be always true.

src/tree/updater_quantile_hist.cc
```diff
@@ -480,8 +549,9 @@ void QuantileHistMaker::Builder::Update(const GHistIndexMatrix& gmat,
     p_tree->Stat(nid).base_weight = snode_[nid].weight;
     p_tree->Stat(nid).sum_hess = static_cast<float>(snode_[nid].stats.sum_hess);
   }
-
+  builder_monitor_.Start("PrunerUpdate");
   pruner_->Update(gpair, p_fmat, std::vector<RegTree*>{p_tree});
+  builder_monitor_.Stop("PrunerUpdate");
```
SmirnovEgorRu, line 554: Should monitors be a part of Pruner, not Builder?

src/tree/updater_quantile_hist.h
```diff
@@ -81,7 +81,9 @@ using xgboost::common::Column;
 /*! \brief construct a tree using quantized feature values */
 class QuantileHistMaker: public TreeUpdater {
  public:
-  QuantileHistMaker() = default;
+  QuantileHistMaker() {
+    updater_monitor_.Init("Quantile");
```
SmirnovEgorRu, line 85: "QuantileHistMaker": probably

src/tree/updater_quantile_hist.h
```diff
@@ -331,6 +333,9 @@ class QuantileHistMaker: public TreeUpdater {
     std::vector<NodeEntry> snode_;
     /*! \brief culmulative histogram of gradients. */
     HistCollection hist_;
+    /*! \brief culmulative local parent histogram of gradients. */
+    HistCollection phist_local_;
```
SmirnovEgorRu, line 337: prefix "p" means pointer. So, ```hist_local_```, probably, is better naming.
Or even ```hist_local_worker_```

src/tree/updater_quantile_hist.cc
```diff
@@ -102,33 +104,82 @@ void QuantileHistMaker::Builder::SyncHistograms(
...
+      if (!((*p_tree)[entry.nid].IsLeftChild())) {
+        auto this_hist = hist_[entry.nid];
+
+        if (!(*p_tree)[entry.nid].IsRoot() && entry.sibling_nid > -1) {
+          auto parent_hist = hist_[(*p_tree)[entry.nid].Parent()];
+          auto sibling_hist = hist_[entry.sibling_nid];
+          SubtractionHist(this_hist, parent_hist, sibling_hist, r.begin(), r.end());
+        }
+      }
+    });
```
SmirnovEgorRu, line 182: So, you just do Subtraction Trick for left nodes from 2 different queues, other things are the same.
You can merge them into one simply or extract to a new method. Anyway, we should reduce "copy-past" code.

src/tree/updater_quantile_hist.cc
ShvetsKS: done

src/tree/updater_quantile_hist.cc
ShvetsKS: done

src/tree/updater_quantile_hist.h
ShvetsKS: renamed, thanks

src/tree/updater_quantile_hist.h
ShvetsKS: done

src/tree/updater_quantile_hist.cc
ShvetsKS: unfortunately it should be sorted, without sorting I get wrong results(different evaluation error)

src/tree/updater_quantile_hist.cc
ShvetsKS: simplified

src/tree/updater_quantile_hist.cc
ShvetsKS: this timer is interesting in context of Builder measure as I understand.

SmirnovEgorRu: Yes, but it's not only one place where the pruner is called

src/tree/updater_quantile_hist.cc
ShvetsKS: method `ParallelSubtractionHist` was added

src/tree/updater_quantile_hist.cc
ShvetsKS: done, thanks

src/tree/updater_quantile_hist.cc
ShvetsKS: refactored

src/tree/updater_quantile_hist.cc
ShvetsKS: this timer was moved

src/tree/updater_quantile_hist.cc
```diff
@@ -152,24 +183,50 @@ void QuantileHistMaker::Builder::BuildHistogramsLossGuide(
...
+    }
   }
-  (*sync_count) = nodes_for_explicit_hist_build_.size();
 
-  for (auto const& node : nodes_for_subtraction_trick_) {
-    hist_.AddHistRow(node.nid);
+  if (n_left == 0) {
+    (*sync_count) = 1;
+  } else {
+    (*sync_count) = n_left;
```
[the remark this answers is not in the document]
ShvetsKS: done

ShvetsKS:
> Please don't merge until [#5557 (comment)](https://github.com/dmlc/xgboost/pull/5557#issuecomment-620240348) is resolved. Also can we bring some refactoring like [#5557 (comment)](https://github.com/dmlc/xgboost/pull/5557#issuecomment-620239758) before merging any further optimization ?

@trivialfis Thanks for your suggestion.
Some refactoring was done, I hope it's more clear now. (Not canonical strategy pattern was implemented for synchronization and adding hist rows. I think the role of 'Context' class is covered by 'QuantileHistMaker::Builder' already. )

Unit tests for new functionality is in progress.

ShvetsKS:
@trivialfis , @SmirnovEgorRu seems all major comments were applied, could you take a look ?

thvasilo:
Hello @ShvetsKS since these change aim at improving distributed code, have you tested against the master in a distributed (not local cluster) environment. 

My concern with previous changes being that they introduced many synchronization points. That's not happening any more I think, but I'd like to see the differences when run say over dask or Spark (or MPI) with network communication happening.

To stress test I'd suggest using the [URL dataset](https://www.csie.ntu.edu.tw/~cjlin/libsvmtools/datasets/binary.html#url) which should introduce large amounts of communication.

trivialfis:
@thvasilo We have specialized optimization for super sparse dataset, the opt is not enabled on distributed environment.  If @ShvetsKS is interested I can point out the details.

ShvetsKS:
> @thvasilo We have specialized optimization for super sparse dataset, the opt is not enabled on distributed environment. If @ShvetsKS is interested I can point out the details.

As far as I understand changes in this PR also should introduce some impact for sparse datasets (we have `BuildHist` specialized for sparse data, and Partition kernel). @trivialfis  Could you share the details please?

@thvasilo thanks for suggestions, measurements are in progress.

trivialfis:
@ShvetsKS This line: https://github.com/dmlc/xgboost/blob/535479e69fd298c2cbabcb853adebb97f2fdf515/src/common/hist_util.cc#L129

If you run URL without the `SparseCuts`, the memory usage will blow up (by blow up I mean we don't know how much it needs, no one has the machine with enough memory to try).  But it's not enabled for distributed environment as we need to sync the number of required entries, I wasn't familiar with distributed system back then ...

ShvetsKS:
@thvasilo
please see measurements (over Spark)
mortgage data set was used.
606M Rows on 4 SKL8180 nodes | depthwise, s | lossguide, s
-- | -- | --
Master | 473.64 | 749.62
This PR | 303.51| 399.67

src/tree/updater_quantile_hist.h
```diff
@@ -105,6 +109,12 @@ class QuantileHistMaker: public TreeUpdater {
   }
 
  protected:
+  friend class HistSynchronizer;
```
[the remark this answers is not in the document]
ShvetsKS: This classes are defined as `friends` as it requires access to protected field. Usage of visitor pattern does't eliminate this problem (get methods should be added, or `AddHistRows` and `SyncHistograms` methods should accept all required fields).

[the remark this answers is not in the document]
ShvetsKS: I don't think so. `HistSynchronizer`, ... classes were defined as `friends` to have access to protected fields, and visitor pattern doesn't eliminate this problem (we will need to add get methods for each required field, or pass its references into `AddHistRows`, `SyncHistograms` methods).

src/tree/updater_quantile_hist.h
```diff
@@ -367,14 +384,79 @@ class QuantileHistMaker: public TreeUpdater {
...
+                              RegTree *p_tree) override;
+};
+
+class DistributedHistSynchronizer: public HistSynchronizer {
+ public:
+  explicit DistributedHistSynchronizer(QuantileHistMaker::Builder* builder):
+                                       HistSynchronizer(builder) {}
+
+  void SyncHistograms(int starting_index,
+                              int sync_count,
```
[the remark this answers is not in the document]
ShvetsKS: fixed

src/tree/updater_quantile_hist.h
```diff
@@ -367,14 +384,79 @@ class QuantileHistMaker: public TreeUpdater {
...
+class HistRowsAdder {
+ public:
+  explicit HistRowsAdder(QuantileHistMaker::Builder* builder) : builder_(builder) {}
+  virtual void AddHistRows(int *starting_index, int *sync_count, RegTree *p_tree) = 0;
+  virtual ~HistRowsAdder() {
+    builder_ = nullptr;
+  }
+
+ protected:
+  QuantileHistMaker::Builder* builder_;
```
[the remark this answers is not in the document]
ShvetsKS: This pointer was moved to `SyncHistograms` and `AddHistRows` methods.
Sorry, I don't see benefits from applying visitor pattern (https://github.com/dmlc/xgboost/pull/5557#discussion_r426546418).

src/tree/updater_prune.cc
```diff
@@ -14,7 +14,7 @@
 #include "xgboost/json.h"
 #include "./param.h"
 #include "../common/io.h"
-
+#include "../common/timer.h"
```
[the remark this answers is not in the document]
ShvetsKS: I only wanted to track 'hidden' time.

src/tree/updater_quantile_hist.cc
```diff
@@ -95,43 +104,151 @@ bool QuantileHistMaker::UpdatePredictionCache(
...
+                                     int *starting_index, int *sync_count,
+                                     RegTree *p_tree) {
+  builder->builder_monitor_.Start("AddHistRows");
+
+  for (auto const& entry : builder->nodes_for_explicit_hist_build_) {
+    int nid = entry.nid;
+    builder->hist_.AddHistRow(nid);
+    (*starting_index) = std::min(nid, (*starting_index));
   }
+  (*sync_count) = builder->nodes_for_explicit_hist_build_.size();
```
SmirnovEgorRu, line 203: Minor comment: there is no need to initialize ```starting_index``` and ```sync_count``` by special value in cases of single-node mode. Since it's used only in distributed mode.

thvasilo:
> @ShvetsKS This line:
> 
> https://github.com/dmlc/xgboost/blob/535479e69fd298c2cbabcb853adebb97f2fdf515/src/common/hist_util.cc#L129
> 
> If you run URL without the `SparseCuts`, the memory usage will blow up (by blow up I mean we don't know how much it needs, no one has the machine with enough memory to try). But it's not enabled for distributed environment as we need to sync the number of required entries, I wasn't familiar with distributed system back then ...

That will be interesting to investigate in a future PR. 

Thanks @ShvetsKS for the measurements, great improvements!

src/tree/updater_quantile_hist.cc
ShvetsKS: fixed

PR #5558, opened 2020-04-17
gpu_hist performance fixes

RAMitchell:
There are a few fixes here:
 - #5534 introduced a performance regression. Caching device vectors were used as temporary vectors in the EvaluateSplit code, creation of these vectors introduces synchronisation, slowing down the algorithm. This PR introduces TemporaryArray as a simple data structure, still using the same allocator internally, but avoiding the synchronisation overhead.
- Remove calls to cudaSetDevice on a per node basis. This function is called over 100,000 times when building 500 trees, which is a little excessive. While it is relatively cheap, this does have a noticeable impact. I have removed all calls during each node construction, but left calls that occur only once per tree.
- Fixed a bug where the histogram storage would double on every call up to `size_t kStopGrowingSize = 1 << 26`, which allocates about 2GB of memory, regardless of how much memory is actually needed. This saves a huge amount of working memory and also makes the algorithm faster because we don't need to zero 2GB of memory every boosting iteration.

Running `python benchmark_tree.py` on my windows system with a 1080Ti, 500 iterations with 10M*50 training instances, training time goes from 18s to 10.5s. If we consider before the regression of #5534 it goes from 12.5s to 10.5s, an improvement of about 1.2x.

### DIFF
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -209,7 +209,6 @@ inline void LaunchN(int device_idx, size_t n, cudaStream_t stream, L lambda) {
   if (n == 0) {
     return;
   }
-  safe_cuda(cudaSetDevice(device_idx));
   const int GRID_SIZE =
       static_cast<int>(xgboost::common::DivRoundUp(n, ITEMS_PER_THREAD * BLOCK_THREADS));
   LaunchNKernel<<<GRID_SIZE, BLOCK_THREADS, 0, stream>>>(  // NOLINT
@@ -368,6 +367,7 @@ struct XGBCachingDeviceAllocatorImpl : thrust::device_malloc_allocator<T> {
      GlobalMemoryLogger().RegisterDeallocation(ptr.get(), n * sizeof(T));
      GetGlobalCachingAllocator().DeviceFree(ptr.get());
    }
+
   __host__ __device__
     void construct(T *)  // NOLINT
   {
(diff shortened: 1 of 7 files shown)


src/common/device_helpers.cuh
```diff
@@ -391,6 +391,24 @@ using device_vector = thrust::device_vector<T,  XGBDeviceAllocator<T>>;  // NOLI
 template <typename T>
 using caching_device_vector = thrust::device_vector<T,  XGBCachingDeviceAllocator<T>>;  // NOLINT
 
+// Faster to instantiate than caching_device_vector and invokes no synchronisation
+// Use this where vector functionality (e.g. resize) is not required
+template <typename T>
+class TemporaryArray {
```
trivialfis, line 397: Maybe `FixedSizeArray` ...

src/tree/updater_gpu_hist.cu
```diff
@@ -324,9 +319,9 @@ class DeviceHistogram {
   }
 
   void Reset() {
-    dh::safe_cuda(cudaMemsetAsync(
-        data_.data().get(), 0,
-        data_.size() * sizeof(typename decltype(data_)::value_type)));
+    auto d_data = data_.data().get();
+      dh::LaunchN(device_id_, data_.size(),
```
trivialfis, line 323: Is this just a `memset` or `thrust::fill`?

src/common/device_helpers.cuh
RAMitchell: Array already implies a fixed size: https://en.cppreference.com/w/cpp/container/array

src/tree/updater_gpu_hist.cu
RAMitchell: Yes, I wanted to avoid cudaMemset due to performance concerns and calls to thrust may soon be synchronous. I might be migrating a lot of thrust code in the near future to other options.

src/common/device_helpers.cuh
trivialfis: That's compile time.

RAMitchell: Nonetheless I would like to keep the name. This also follows thrust::detail::temporary_array.

PR #5560, opened 2020-04-19
Use non-synchronising scan

RAMitchell:
This switches the inclusive scan in the row partitioner from thrust to cub to avoid the synchronisation of thrust APIs from cuda 10.1 onwards

### DIFF
--- a/src/tree/gpu_hist/row_partitioner.cu
+++ b/src/tree/gpu_hist/row_partitioner.cu
@@ -63,7 +63,7 @@ struct WriteResultsFunctor {
 // Change the value type of thrust discard iterator so we can use it with cub
 class DiscardOverload : public thrust::discard_iterator<IndexFlagTuple> {
  public:
-  using value_type = IndexFlagTuple;
+  using value_type = IndexFlagTuple;  // NOLINT
 };
 
 // Implement partitioning via single scan operation using transform output to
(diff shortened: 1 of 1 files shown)


trivialfis:
I need to run some large deterministic tests for cub algorithms.

RAMitchell:
This one is just integers, but yes we need to watch out for that non deterministic floating point reduction we had with cub.

src/tree/gpu_hist/row_partitioner.cu
```diff
@@ -68,19 +76,21 @@ void RowPartitioner::SortPosition(common::Span<bst_node_t> position,
                                   int64_t* d_left_count, cudaStream_t stream) {
   WriteResultsFunctor write_results{left_nidx, position, position_out,
                                     ridx,      ridx_out, d_left_count};
-  auto discard_write_iterator = thrust::make_transform_output_iterator(
-      thrust::discard_iterator<int>(), write_results);
+  auto discard_write_iterator =
+      thrust::make_transform_output_iterator(DiscardOverload(), write_results);
```
trivialfis, line 80: Interesting, so this function works on Windows but not `make_transform_iterator` ...

RAMitchell:
It's a bigger job to comprehensively go through and change everything. This one is important because it gets called ~15000 times when building 500 trees, the others can wait I think.

PR #5570, opened 2020-04-21
Use cudaDeviceGetAttribute instead of cudaGetDeviceProperties

RAMitchell:
https://devblogs.nvidia.com/cuda-pro-tip-the-fast-way-to-query-device-properties/

This results in a nontrivial performance improvement on linux (~10%).


trivialfis:
How many times do we call this function?

RAMitchell:
~1000, for 500 trees, in gpu predictor.

PR #3723, opened 2018-09-25
Disable flaky tests in R-package/tests/testthat/test_update.R

hcho3:
Disable flaky tests in R-package/tests/testthat/test_update.R.
Lines 32, 41 and 64 randomly fail for 32-bit Windows + MinGW target.

Build log (AppVeyor): https://ci.appveyor.com/project/tqchen/xgboost/build/1.0.2949/job/7i0d2lim73hhy44f#L2960, https://ci.appveyor.com/project/tqchen/xgboost/build/1.0.2954/job/l2icjm6yq4kc2k3s#L3093

PR #5577, opened 2020-04-21
Port R compatibility patches from 1.0.0 release branch

hcho3:
I made two patches to `release_1.0.0` branch in order to comply with CRAN package submission policy. The two patches are not yet in the master branch.

### DIFF
--- a/R-package/tests/testthat/test_helpers.R
+++ b/R-package/tests/testthat/test_helpers.R
@@ -7,8 +7,8 @@ require(vcd, quietly = TRUE)
 
 float_tolerance = 5e-6
 
-# disable some tests for Win32
-win32_flag = .Platform$OS.type == "windows" && .Machine$sizeof.pointer != 8
+# disable some tests for 32-bit environment
+flag_32bit = .Machine$sizeof.pointer != 8
 
 set.seed(1982)
 data(Arthritis)
@@ -44,7 +44,7 @@ mbst.GLM <- xgboost(data = as.matrix(iris[, -5]), label = mlabel, verbose = 0,
 
 
 test_that("xgb.dump works", {
-  if (!win32_flag)
+  if (!flag_32bit)
     expect_length(xgb.dump(bst.Tree), 200)
   dump_file = file.path(tempdir(), 'xgb.model.dump')
   expect_true(xgb.dump(bst.Tree, dump_file, with_stats = T))
@@ -54,7 +54,7 @@ test_that("xgb.dump works", {
   # JSON format
   dmp <- xgb.dump(bst.Tree, dump_format = "json")
   expect_length(dmp, 1)
-  if (!win32_flag)
+  if (!flag_32bit)
     expect_length(grep('nodeid', strsplit(dmp, '\n')[[1]]), 188)
 })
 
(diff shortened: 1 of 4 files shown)


R-package/tests/testthat/test_helpers.R
```diff
@@ -7,8 +7,8 @@ require(vcd, quietly = TRUE)
 
 float_tolerance = 5e-6
 
-# disable some tests for Win32
-win32_flag = .Platform$OS.type == "windows" && .Machine$sizeof.pointer != 8
+# disable some tests for 32-bit environment
+flag_32bit = .Machine$sizeof.pointer != 8
```
trivialfis, line 11: What's wrong with our 32 bit build?

src/common/hist_util.cc
```diff
@@ -831,7 +831,11 @@ void GHistIndexBlockMatrix::Init(const GHistIndexMatrix& gmat,
  * \brief fill a histogram by zeros in range [begin, end)
  */
 void InitilizeHistByZeroes(GHistRow hist, size_t begin, size_t end) {
+#if defined(XGBOOST_STRICT_R_MODE) && XGBOOST_STRICT_R_MODE == 1
```
trivialfis, line 834: Could you run a small benchmark on this?  You can  call `std::fill(hist.data() + begin, hist.data() + end, tree::GradStats()` to avoid overhead created by `span`.  If the perf gain by `memset` is negligible, we can mainline it to master.

src/common/observer.h
```diff
@@ -20,11 +20,11 @@
 #define OBSERVER_PRINT LOG(INFO)
 #define OBSERVER_ENDL ""
 #define OBSERVER_NEWLINE ""
-#else
+#else  // defined(XGBOOST_STRICT_R_MODE) && XGBOOST_STRICT_R_MODE == 1
```
trivialfis, line 23: You can remove any use of this class (currently only called in learner), and don't define the class when compiling on R.

src/common/hist_util.cc
SmirnovEgorRu: We can use in both cases:
```cpp
using FPType = decltype(tree::GradStats::sum_grad);
FPType* pdata = reinterpret_cast<FPType*>(hist.data());
std::fill(pdata + 2 * begin, pdata + 2 * (begin + end), 0);
```

It should have the same perf as memset

hcho3: Let's come back to this after 1.1.0 release.

R-package/tests/testthat/test_helpers.R
hcho3: Some of the unit tests will fail because 32-bit platform uses x87 unit, whereas 64-bit uses SSE. See #3723 for context. In the long run, we should re-design tests so that they work equally well on 32-bit and 64-bit systems. 

Support for 32-bit systems is mandated by CRAN submission policy.

hcho3: Tracking issue: #5578

src/common/hist_util.cc
hcho3: Tracking issue: #5579

src/common/observer.h
hcho3: That would disable debug logging entirely for the R package. Not sure if we should do that.

PR #5581, opened 2020-04-22
Avoid rabit calls in learner configuration

RAMitchell:
I had an issue in #5484 where trying to save the model only on one distributed worker resulted in a rabit error, in that case I modified the tests, but came across the same error today when running the cover type dataset with gbm-bench. 

I traced this back to learner.Configure() being called when the model is saved and then calling rabit allreduce to synchronise number of features. This rabit call should be unnecessary as DMatrix constructors are responsible for making sure they correctly set number of columns in distributed environments.

PR #5585, opened 2020-04-23
[WIP] [jvm-packages] move dmatrix building into rabit env

wbo4958:
DMatrix needs to communicate with other partitions to make sure the
number of features is consistent.


wbo4958:
For now, there are some unit tests failing. one is feature size checking in DMatrix, the other one is rabit  timeout issue. Hi @trivialfis can you help on these issues?

RAMitchell:
Is it possible that the number of DMatrix objects on each worker becomes different? Or the creation of these objects is not synchronised across workers for some reason?

trivialfis:
Let me add a DMatrix check.

RAMitchell:
We merged a more moderate solution in #5581 given that there might be complications with this approach.

wbo4958:
Thx Rory, Closed this since #5581 has been merged.

PR #5589, opened 2020-04-23
[CI] Use Vault repository to re-gain access to devtoolset-4

hcho3:
* `tests/ci_build/Dockerfile.gpu_build`: devtoolset-4 is no longer available from mirror.centos.org. So use vault.centos.org repository instead.
* `tests/ci_build/Dockerfile.jvm`: devtoolset-6 has been removed from centos-sclo-rh-testing repository, not too long after devtoolset-4 had been removed there (#5506). Use vault.centos.org repository instead.
* Use manylinux2010 tag, since we use CentOS 6 to build the Python wheel. The wheel will now be called `xgboost-1.x.x-py3-none-manylinux2010_x86_64.whl`. Tensorflow uses manylinux2010 tag too.


hcho3:
In the long term, we should migrate to CentOS 7. Python community recently adopted PEP 599, which creates `manylinux2014` tag to designate wheels built with CentOS 7.

Issue #5592, opened 2020-04-24
Check failed: gpu_predictor_


rgbinventions:
Hey guys!
I've built xgboost with GPU support. Everything was ok. But when i'm training the classifier model i got several gpu_predictor warnings and then fatal error. Here is the full error list:
```
XGBoostError                              Traceback (most recent call last)
<ipython-input-5-f9ae4a94f7eb> in <module>
     28 
     29 gsearch = GridSearchCV(estimator=model, param_grid=params_grid, scoring='roc_auc', cv=5)
---> 30 gsearch.fit(X_train, y_train)
     31 
     32 predictions = gsearch.best_estimator_.predict(X_test)

C:\Anaconda3\lib\site-packages\sklearn\model_selection\_search.py in fit(self, X, y, groups, **fit_params)
    737             refit_start_time = time.time()
    738             if y is not None:
--> 739                 self.best_estimator_.fit(X, y, **fit_params)
    740             else:
    741                 self.best_estimator_.fit(X, **fit_params)

C:\Anaconda3\lib\site-packages\xgboost\sklearn.py in fit(self, X, y, sample_weight, base_margin, eval_set, eval_metric, early_stopping_rounds, verbose, xgb_model, sample_weight_eval_set, callbacks)
    822                               evals_result=evals_result, obj=obj, feval=feval,
    823                               verbose_eval=verbose, xgb_model=xgb_model,
--> 824                               callbacks=callbacks)
    825 
    826         self.objective = xgb_options["objective"]

C:\Anaconda3\lib\site-packages\xgboost\training.py in train(params, dtrain, num_boost_round, evals, obj, feval, maximize, early_stopping_rounds, evals_result, verbose_eval, xgb_model, callbacks)
    210                            evals=evals,
    211                            obj=obj, feval=feval,
--> 212                            xgb_model=xgb_model, callbacks=callbacks)
    213 
    214 

C:\Anaconda3\lib\site-packages\xgboost\training.py in _train_internal(params, dtrain, num_boost_round, evals, obj, feval, xgb_model, callbacks)
     73         # Skip the first update if it is a recovery step.
     74         if version % 2 == 0:
---> 75             bst.update(dtrain, i, obj)
     76             bst.save_rabit_checkpoint()
     77             version += 1

C:\Anaconda3\lib\site-packages\xgboost\core.py in update(self, dtrain, iteration, fobj)
   1366             _check_call(_LIB.XGBoosterUpdateOneIter(self.handle,
[...]

Issue #5011, opened 2019-11-05
GPU memory monitor fails to log allocation.


trivialfis:
When setting `verbosity` to 3 the memory logger should log all allocation and deallocation on GPU.  But I encountered the deallocatation error when input is cudf, saying that the pointer in SpasePage::data is never allocated, which is obviously not true.

This is a reminder for myself.  Also @RAMitchell might have some inputs here.

``` python
def test_cudf():
    X = np.random.randn(kRows, kCols)
    y = np.random.randn(kRows)

    df = pd.DataFrame(X)

    df = cudf.from_pandas(df)

    dtrain = xgb.DMatrix(df, label=y)
    bst = xgb.train({'tree_method': 'gpu_hist', 'verbosity': 2},
                    dtrain=dtrain,
                    evals=[(dtrain, 'X')],
                    num_boost_round=2)

    print('Pickle')
    import pickle
    with open('model.bst', 'bw') as fd:
        pickle.dump(bst, fd)

    with open('model.bst', 'br') as fd:
        bst: xgb.Booster = pickle.load(fd)
        dtest = xgb.DMatrix(df, label=y)
        bst.set_param({'gpu_id': '0'})
        bst.predict(dtest)
```

trivialfis:
Found the cause ... verbosity is not effective during DMatrix construction.

PR #5594, opened 2020-04-24
Better message when no GPU is found.

trivialfis:
https://github.com/dmlc/xgboost/issues/5592

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -187,7 +187,9 @@ void GenericParameter::ConfigureGpuId(bool require_gpu) {
   // raw model), number of available GPUs could be different.  Wrap around it.
   int32_t n_gpus = common::AllVisibleGPUs();
   if (n_gpus == 0) {
-    LOG(WARNING) << "No visible GPU is found, setting `gpu_id` to -1";
+    if (gpu_id != kCpuId) {
+      LOG(WARNING) << "No visible GPU is found, setting `gpu_id` to -1";
+    }
     this->UpdateAllowUnknown(Args{{"gpu_id", std::to_string(kCpuId)}});
   } else if (gpu_id != kCpuId && gpu_id >= n_gpus) {
     LOG(WARNING) << "Only " << n_gpus << " GPUs are visible, setting `gpu_id` to " << gpu_id;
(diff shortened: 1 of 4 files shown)


trivialfis:
@RAMitchell In a longer term,  I would like to have something similar to: https://www.tensorflow.org/api_docs/python/tf/config/list_physical_devices in XGBoost, WDYT?

RAMitchell:
@trivialfis seems like users could do this with many available third-party tools, why does xgboost need to do this?

trivialfis:
@RAMitchell We can use it in combined with many other things:

- Is XGBoost compiled with CUDA?
- What version of CUDA is XGBoost compiled with?
- Is XGBoost compiled with OpenMP?
- What's the default number of threads in XGBoost: #4975
- Verbosity #5011 .
- Is CPU prefetching enabled in XGBoost?

trivialfis:
@hcho3 @RAMitchell Could you please review?

src/learner.cc
```diff
@@ -187,8 +187,12 @@ void GenericParameter::ConfigureGpuId(bool require_gpu) {
   // raw model), number of available GPUs could be different.  Wrap around it.
   int32_t n_gpus = common::AllVisibleGPUs();
   if (n_gpus == 0) {
+    if (gpu_id != kCpuId) {
+      LOG(WARNING) << "No visible GPU is found, setting `gpu_id` to -1";
+    }
     this->UpdateAllowUnknown(Args{{"gpu_id", std::to_string(kCpuId)}});
   } else if (gpu_id != kCpuId && gpu_id >= n_gpus) {
+    LOG(WARNING) << "Only " << n_gpus << " GPUs are visible, setting `gpu_id` to " << gpu_id;
```
hcho3, line 195:
```suggestion
    LOG(WARNING) << "Only " << n_gpus << " GPUs are visible, setting `gpu_id` to " << (gpu_id % n_gpus);
```

tests/python-gpu/load_pickle.py
```diff
@@ -49,3 +55,16 @@ def test_wrap_gpu_id(self):
         test_x = xgb.DMatrix(x)
         res = bst.predict(test_x)
         assert len(res) == 10
+
+    def test_training_on_cpu_only_env(self):
+        assert os.environ['CUDA_VISIBLE_DEVICES'] == '-1'
+        print('RUnning')
```
hcho3, line 61:
```suggestion
```

PR #5603, opened 2020-04-25
Clarify docstring for XGBoosterPredict() C API

### LINKED ISSUE #5601: v1.1.0rc1 breaking API change: when should `training` be true in XGBoosterPredict(handle, dmat, option_mask, ntree_limit, training, ...)
I'm trying to update downstream libraries for v1.1.0rc1. There was a breaking change in the c_api.h to have an additional `training bool` parameter in this code change:

https://github.com/hcho3/xgboost/commit/f100b8d87871adaa693ea84548d69a598672d6f9
~~~
* Add `training` parameter for prediction method.
* [Breaking]: Add `training` to C API.
~~~

I'm looking at the patch but cannot tell when training should be true or false in the XGBoosterPredict() call. Could somebody offer some guidance?

I read that 
~~~
* \param training Whether the prediction value is used for training.
~~~

but that doesn't really tell me enough to understand what the consequences of training true (1) will be.

Thanks for you help.

hcho3:
Clarify the meaning of `training` parameter to the C API function `XGBoosterPredict()`.

Closes #5601.

@glycerine Can you review?

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -418,7 +418,8 @@ XGB_DLL int XGBoosterEvalOneIter(BoosterHandle handle,
  *          4:output feature contributions to individual predictions
  * \param ntree_limit limit number of trees used for prediction, this is only valid for boosted trees
  *    when the parameter is set to 0, we will use all the trees
- * \param training Whether the prediction value is used for training.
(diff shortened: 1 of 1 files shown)


hcho3:
Closing in favor of #5604

PR #5604, opened 2020-04-25
Clarify meaning of `training` parameter in XGBoosterPredict()

### LINKED ISSUE #5601: v1.1.0rc1 breaking API change: when should `training` be true in XGBoosterPredict(handle, dmat, option_mask, ntree_limit, training, ...)

glycerine:
fixes #5601 . Pre-empts questions about the breaking API change for v1.1.0 that downstream maintainers will have.

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -418,12 +418,14 @@ XGB_DLL int XGBoosterEvalOneIter(BoosterHandle handle,
  *          4:output feature contributions to individual predictions
  * \param ntree_limit limit number of trees used for prediction, this is only valid for boosted trees
  *    when the parameter is set to 0, we will use all the trees
- * \param training_dart_use_dropout Whether the prediction value is used for training under DART.
- *    For most use cases, set it to 0. Set this parameter to 1 only if ALL of the following
- *    conditions hold:
- *    1) you are writing a training loop (inference doesn't apply);
- *    2) you are using DART;
- *    3) you are using a customized objective function.
+ * \param training Whether the prediction function is used as part of a training loop.
+ *    1. If you are running prediction to obtain a metric (not gradient pairs), set training=False.
+ *    2. When should you set training=true?
+ *       Normally, you don't need to use XGBoosterPredict() in your training loop, since you are
+ *       probably using XGBoosterUpdateOneIter(), which calls XGBoosterPredict() internally. The one
+ *       case you should use XGBoosterPredict() in your training loop is when you wrote a customized
+ *       objective. In this case, call XGBoosterPredict() with training=true and then invoke your
+ *       customized objective fobj(predicted_label, true_label) to obtain the gradient pairs.
  * \param out_len used to store length of returning result
  * \param out_result used to set a pointer to array
  * \return 0 when success, -1 when failure happens
@@ -432,7 +434,7 @@ XGB_DLL int XGBoosterPredict(BoosterHandle handle,
                              DMatrixHandle dmat,
                              int option_mask,
                              unsigned ntree_limit,
-                             int training_dart_use_dropout,
+                             int training,
                              bst_ulong *out_len,
                              const float **out_result);
 /*
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -419,15 +419,11 @@ XGB_DLL int XGBoosterEvalOneIter(BoosterHandle handle,
  * \param ntree_limit limit number of trees used for prediction, this is only valid for boosted trees
  *    when the parameter is set to 0, we will use all the trees
  * \param training_dart_use_dropout Whether the prediction value is used for training under DART.
- *        If you are not using DART, just set it to 0 and ignore rest of this discussion.
- *        Setting training_dart_use_dropout=1 will cause the prediction function to perform 
- *        dropout, evaluating only a subset of the trees in the ensemble. Setting it to 0 will 
- *        evaluate all trees.
- *        XGBoosterPredict() is used in two places:
- *	  a) When you are training a tree ensemble. Here, you want to perform dropout.
- *        b) When you are done training and now would like to evaluate the model on a new dataset.
- *           In this second case, you do not want dropout.
- *	  Are you using XGBoosterPredict() after you are done training your model? Then set training=0 (false).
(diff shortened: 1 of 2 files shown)


hcho3:
@glycerine I simplified the explanation in the docstring. Can you review and see if it makes sense to you?

glycerine:
@hcho3 LGTM. Thanks for adding the clarification about custom objectives.

trivialfis:
This way the parameter is general enough to be reused in the future instead of being a dart only parameter.

hcho3:
@trivialfis Can you suggest a good way to document the meaning of `training` for people who use the C API function? Right now, it is not clear when `training` parameter should be set to 1.

hcho3:
```
 * \param training Whether the prediction function is used as part of a training loop.
 *    If you are running prediction on a model that's finished training, set `training` to false.
```

How about this?

glycerine:
YAGNI applies. I would say don't document it as general-purpose until it actual is. It is already used here in R in the general sense, even when DART may not be used.

https://github.com/dmlc/xgboost/blob/master/R-package/R/utils.R#L148

Is there harm is setting to true when not under DART?

trivialfis:
Maybe we can give `dart` as an example.

glycerine:
> If you are running prediction on a model that's finished training, set `training` to false.

A little too sparse. I would want a little more guidance about when it is okay to set it to true.

trivialfis:
@glycerine How about this, if the prediction result is used for calculating gradient for training, set it to true, otherwise set it to false.

trivialfis:
Also, it's a breaking change in 1.0.0.

hcho3:
```
* \param training Whether the prediction function is used as part of a training loop.
*    1. If you are running prediction on a model that's finished training, set training=false.
*    2. When should you set training=true?
*       Normally, you don't need to use XGBoosterPredict() in your training loop, since you are
*       probably using XGBoosterUpdateOneIter(), which calls XGBoosterPredict() internally. The one
*       case you should use XGBoosterPredict() in your training loop is when you wrote a customized
*       objective. In this case, call XGBoosterPredict() with training=true and then invoke your
*       customized objective fobj(predicted_label, true_label) to obtain the gradient pairs.
```

I tried elaborating the design behind the `training` parameter.

glycerine:
~~~
* \param training Whether the prediction function is used as part of a training loop.
*    1. If you are running prediction on a model that's finished training, set training=false.
*    2. When should you set training=true?
*       Normally, you don't need to use XGBoosterPredict() in your training loop, since you are
*       probably using XGBoosterUpdateOneIter(), which calls XGBoosterPredict() internally. The one
*       case you should use XGBoosterPredict() in your training loop is when you wrote a customized
*       objective. In this case, call XGBoosterPredict() with training=true and then invoke your
*       customized objective fobj(predicted_label, true_label) to obtain the gradient pairs.
~~~

@hcho3 I like it. Very clear and helpful.

trivialfis:
> *    1. If you are running prediction on a model that's finished training, set training=false.

I would omit this item.  We do prediction during training to use external metrics.

glycerine:
> I would omit this item. We do prediction during training to use external metrics.

@trivialfis That is the most important piece of guidance. Without that, we are lost. 

You really must describe detail when to set it true, and when to set it false. Give examples. Be precise. Without the line above, it is far too vague, and you will get tons of questions.

hcho3:
```
* \param training Whether the prediction function is used as part of a training loop.
*    When should you set training=true? Set training=true if 1) you are writing your training loop 
*    and 2) you are using a customized objective function to obtain gradient pairs.
*    Normally, you don't need to use XGBoosterPredict() in your training loop, since you are
*    probably using XGBoosterUpdateOneIter(), which calls XGBoosterPredict() internally. The one
*    case you should use XGBoosterPredict() in your training loop is when you wrote a customized
*    objective. In this case, call XGBoosterPredict() with training=true and then invoke your
*    customized objective fobj(predicted_label, true_label) to obtain the gradient pairs.
```

glycerine:
great. This is good in that it give guidance about when to use training=true.

It would also be good to add what change happens when you set training true, and that shows the intent. I'm guessing it has to do with dropout or not, as one example.

trivialfis:
Prediction can be run in 2 scenarios:
- Normal model inference, given a X, obtain the y_p from model.
- Obtain the prediction for computing gradients.  For example `dart` booster performs dropout during training, and the prediction result will be different from the one obtained by normal inference step.

The `training` parameter is used to distinguish these 2 scenarios.  Set to `false` for the first scenario and `true` for the second one.

glycerine:
@trivialfis Excellent.

include/xgboost/c_api.h
```diff
@@ -418,7 +418,14 @@ XGB_DLL int XGBoosterEvalOneIter(BoosterHandle handle,
...
  *    when the parameter is set to 0, we will use all the trees
- * \param training Whether the prediction value is used for training.
+ * \param training Whether the prediction function is used as part of a training loop.
+ *    Prediction can be run in 2 scenarios:
+ *    1. Given data matrix X, obtain prediction y_pred from the model.
+ *    2. Obtain the prediction for computing gradients. For example, DART booster performs dropout
+ *       during training, and the prediction result will be different from the one obtained by normal
+ *       inference step due to dropped trees.
+ *    Set training=false for the first scenario. Set training=true for the second scenario.
+ *    The second scenario applies when you are defining a custom objective function.
```
hcho3, line 428: Two scenarios are now listed. DART is used as an example. Also mentioning custom objective as an example of Scenario 2.

glycerine: looks great to me.

PR #5610, opened 2020-04-27
Refactor gpu_hist split evaluation

RAMitchell:
- Move split evaluation code out of updater_gpu_hist.cu
- Fuse left and right child split evaluation into a single kernel. (We can now run this kernel on a single stream, later I want to have each node construction running kernels asynchronously on its own stream).
- Remove dependency on EllPack data structure, accept spans instead so we can more easily test this component
- Clean up some unnecessary member variables in updater_gpu_hist.cu

Some possibilities for the next PR:
- Pass operators to split evaluation, handling gain calculation, constraints etc.
- Rewrite split evaluation using cub inclusive scan instead of custom kernels

### DIFF
--- a/src/tree/constraints.cuh
+++ b/src/tree/constraints.cuh
@@ -28,8 +28,8 @@ struct ValueConstraint {
   inline static void Init(tree::TrainParam *param, unsigned num_feature) {
     param->monotone_constraints.resize(num_feature, 0);
   }
-  template <typename ParamT>
-  XGBOOST_DEVICE inline double CalcWeight(const ParamT &param, tree::GradStats stats) const {
+  template <typename ParamT, typename GpairT>
+  XGBOOST_DEVICE inline double CalcWeight(const ParamT &param, GpairT stats) const {
     double w = xgboost::tree::CalcWeight(param, stats);
     if (w < lower_bound) {
       return lower_bound;
@@ -63,9 +63,9 @@ struct ValueConstraint {
       return wleft >= wright ? gain : negative_infinity;
     }
   }
-
-  inline void SetChild(const tree::TrainParam &param, bst_uint split_index,
-                       tree::GradStats left, tree::GradStats right, ValueConstraint *cleft,
+  template <typename GpairT>
+  void SetChild(const tree::TrainParam &param, bst_uint split_index,
+                       GpairT left, GpairT right, ValueConstraint *cleft,
                        ValueConstraint *cright) {
     int c = param.monotone_constraints.at(split_index);
     *cleft = *this;
--- a/src/tree/gpu_hist/evaluate_splits.cu
+++ b/src/tree/gpu_hist/evaluate_splits.cu
@@ -1,0 +1,246 @@
+/*!
+ * Copyright 2020 by XGBoost Contributors
+ */
+#include "evaluate_splits.cuh"
+
+namespace xgboost {
+namespace tree {
+
+// With constraints
+template <typename GradientPairT>
+XGBOOST_DEVICE float LossChangeMissing(
+    const GradientPairT& scan, const GradientPairT& missing,
+    const GradientPairT& parent_sum, const GPUTrainingParam& param,
+    int constraint, const ValueConstraint& value_constraint,
+    bool& missing_left_out) {  // NOLINT
+  float parent_gain = CalcGain(param, parent_sum);
+  float missing_left_gain = value_constraint.CalcSplitGain(
+      param, constraint, GradStats(scan + missing),
+      GradStats(parent_sum - (scan + missing)));
+  float missing_right_gain = value_constraint.CalcSplitGain(
+      param, constraint, GradStats(scan), GradStats(parent_sum - scan));
+
+  if (missing_left_gain >= missing_right_gain) {
+    missing_left_out = true;
+    return missing_left_gain - parent_gain;
+  } else {
+    missing_left_out = false;
+    return missing_right_gain - parent_gain;
+  }
+}
+
+/*!
+ * \brief
+ *
+ * \tparam ReduceT     BlockReduce Type.
+ * \tparam TempStorage Cub Shared memory
+ *
+ * \param begin
+ * \param end
+ * \param temp_storage Shared memory for intermediate result.
+ */
+template <int BLOCK_THREADS, typename ReduceT, typename TempStorageT, typename GradientSumT>
+__device__ GradientSumT ReduceFeature(common::Span<const GradientSumT> feature_histogram,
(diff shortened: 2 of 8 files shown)


RAMitchell:
I modified a couple of tests here that were giving false negatives. 

`tests/python-gpu/test_gpu_prediction.py` tests that AUC is not increasing with a logloss objective, which cannot be expected to always be true, I changed this to testing for a non-increasing logloss metric.

I changed the tolerance on `tests/cpp/tree/gpu_hist/test_gradient_based_sampler.cu` to be much higher. This test follows a binomial distribution and can be expected to fail with probability ~0.04 for any given random seed with n=4096, p=0.5, even when the algorithm is correct. The probability of false negative with the new tolerance is < 0.0001.

trivialfis:
Will take a look later today, github is being moody right now.

src/tree/constraints.cuh
```diff
@@ -28,8 +28,8 @@ struct ValueConstraint {
   inline static void Init(tree::TrainParam *param, unsigned num_feature) {
     param->monotone_constraints.resize(num_feature, 0);
   }
-  template <typename ParamT>
-  XGBOOST_DEVICE inline double CalcWeight(const ParamT &param, tree::GradStats stats) const {
+  template <typename ParamT, typename GpairT>
+  XGBOOST_DEVICE inline double CalcWeight(const ParamT &param, GpairT stats) const {
```
trivialfis, line 32: Is there any change for the gradient type?

src/tree/gpu_hist/evaluate_splits.cu
```diff
@@ -0,0 +1,261 @@
...
+// With constraints
+template <typename GradientPairT>
+XGBOOST_DEVICE float LossChangeMissing(const GradientPairT& scan,
+                                       const GradientPairT& missing,
+                                       const GradientPairT& parent_sum,
+                                       const GPUTrainingParam& param,
+                                       int constraint,
+                                       const ValueConstraint& value_constraint,
+                                       bool& missing_left_out) {  // NOLINT
+  float parent_gain = CalcGain(param, parent_sum);
```
trivialfis, line 19: I see, the `CalcGain` is using a gradient pair instead of `GradStats`, do you think this is the reason for changed training score?

src/tree/gpu_hist/evaluate_splits.cuh
```diff
@@ -0,0 +1,37 @@
...
+#include "../updater_gpu_common.cuh"
+
+namespace xgboost {
+namespace tree {
+
+template <typename GradientSumT>
+struct EvaluateSplitInputs {
+  int nidx;
+  GradientSumT parent_sum;
+  GPUTrainingParam param;
```
trivialfis, line 18: I think the normal `TrainParam` is just a `POD`, we don't need to define `GPUTrainingParam` for passing into kernel.

src/tree/gpu_hist/evaluate_splits.cu
```diff
@@ -0,0 +1,261 @@
...
+                                          const DeviceSplitCandidate& b) {
+  return a.loss_chg > b.loss_chg ? a : b;
+}
+
+template <typename GradientSumT>
+void EvaluateSplits(common::Span<DeviceSplitCandidate> out_splits,
+                    EvaluateSplitInputs<GradientSumT> left,
+                    EvaluateSplitInputs<GradientSumT> right) {
+  size_t combined_num_features =
+      left.feature_set.size() + right.feature_set.size();
```
trivialfis, line 206: Is there any test for column sampling?

src/tree/gpu_hist/evaluate_splits.cuh
```diff
@@ -0,0 +1,37 @@
...
+  common::Span<const bst_feature_t> feature_set;
+  common::Span<const uint32_t> feature_segments;
+  common::Span<const float> feature_values;
+  common::Span<const float> min_fvalue;
+  common::Span<const GradientSumT> gradient_histogram;
+  ValueConstraint value_constraint;
+  common::Span<const int> monotonic_constraints;
+};
+template <typename GradientSumT>
+void EvaluateSplits(common::Span<DeviceSplitCandidate> out_splits,
```
trivialfis, line 28: Out of curiosity, are you planing to generalize this function into more than 2 nodes?

src/tree/updater_gpu_hist.cu
```diff
@@ -520,74 +335,87 @@ struct GPUHistMakerDevice {
...
+                                      left_nidx);
+    auto right_sampled_features =
+        column_sampler.GetFeatureSet(tree.GetDepth(right_nidx));
+    right_sampled_features->SetDevice(device_id);
+    common::Span<bst_feature_t> right_feature_set =
+        interaction_constraints.Query(right_sampled_features->DeviceSpan(),
+                                      left_nidx);
+    auto matrix = page->GetDeviceAccessor(device_id);
+
+    EvaluateSplitInputs<GradientSumT> left{left_nidx,
```
trivialfis, line 390: Nitpick, can we keep the formatting of `left` and `right` consistent?

src/tree/gpu_hist/evaluate_splits.cuh
RAMitchell: It's too large.

src/tree/gpu_hist/evaluate_splits.cuh
RAMitchell: No

RAMitchell:
Yes the code will have to move to the header to use functors.

Let me see if I can find the cause of minor differences.

RAMitchell:
The cause for the slight differences was slightly different code for reducing over feature splits. Given two features with the same split gain, the original code would take the first one, I had inadvertently changed it to be the last. Its changed back in the latest commit.

src/tree/updater_gpu_hist.cu
RAMitchell: This is just what clang format decided, is it really a problem? It's quite annoying for me to have manually format one section of code when the rest is automatically formatted.

src/tree/gpu_hist/evaluate_splits.cu
RAMitchell: Added a test for this and for features with identical gain.

src/tree/gpu_hist/evaluate_splits.cu
RAMitchell: I don't think this matters.

src/tree/constraints.cuh
RAMitchell: I don't think any information is lost here.

Issue #5495, opened 2020-04-07
[FEA] DaskXGBClassifier Early Stopping


programmylife:
We (myself and @kylejn27) would like to update our model training to run with DaskXGBClassifier, but it requires early stopping. We would like DaskXGBClassifier to match the spec of XGBClassifier to include early stopping:

DaskXGBClassifier
```
class DaskXGBClassifier(DaskScikitLearnBase, XGBClassifierBase):
    # pylint: disable=missing-docstring
    _client = None
    def fit(self,
            X,
            y,
            sample_weights=None,
            eval_set=None,
            sample_weight_eval_set=None,
            verbose=True):
```

XGBClassifier
```
class XGBClassifier(XGBModel, XGBClassifierBase):
    # pylint: disable=missing-docstring,invalid-name,too-many-instance-attributes
    def __init__(self, objective="binary:logistic", **kwargs):
        super().__init__(objective=objective, **kwargs)
    def fit(self, X, y, sample_weight=None, base_margin=None,
            eval_set=None, eval_metric=None,
            early_stopping_rounds=None, verbose=True, xgb_model=None,
            sample_weight_eval_set=None, callbacks=None):
```

trivialfis:
Yup.  Will work on this, I just need to redactor the call back a little bit.

Issue #4955, opened 2019-10-16
Learning rate decay


guyko81:
I have been thinking about robust model building on Important Outliers - outliers that matters so we don't want the model to add predictions closer to the average value. That led me the idea that implementing a learning rate decay would add the model more flexibility as in the beginning it could catch the trivial rules (a trivial case decision tree in the beginning) due to high learning rate and later in the training it could fine tune the result. 

I have made my own GBTree with this additional feature and found that the model learnt much faster as well! Not a big coding but great improvement in performance. I have not tested the robustness (e.g. overfitting) of the model on many datasets but the one I used on showed promising results - no overfit.

I used learning_rate_start=0.5, learning_rate_min=0.01 and lr_decay=0.95. 
First iteration:
`lr = learning_rate_start`
At each iteration afterwards the following rule applies:
`lr = max(learning_rate_min, lr*lr_decay)`

trivialfis:
We have a callback for setting learning rate although I haven't really use it myself.  Would it help even simplifying things further?

guyko81:
My mistake, I was not aware of this callback.
However a rename might be useful, maybe learning_rate_schedule, rather than reset_learning_rate?

And an example code would be nice too.
Should the following work? (I can't try right now, no XGBoost on work computer)


```
def learning_rate_decay(boosting_round, num_boost_round):
    learning_rate_start = 0.5
    learning_rate_min = 0.01
    lr_decay = 0.95
    lr = learning_rate_start * np.power(lr_decay, boosting_round)
    return max(learning_rate_min, lr)

xgclassifier = xgb.train(params, xgb_dtrain, num_rounds,  callbacks = [xgb.callback.reset_learning_rate(learning_rate_decay)])
```

and with the new naming (I think it's more obvious):
`xgclassifier = xgb.train(params, xgb_dtrain, num_rounds,  callbacks = [xgb.callback.learning_rate_schedule(learning_rate_decay)])`

trivialfis:
@guyko81 Thanks for the suggestion.  I agree with your naming scheme, will keep this open so we can have better documentation for callbacks.

hcho3:
To new contributors: If you're reading this and interested in writing the document for learning rate decay, please comment here. Feel free to ping me with questions. I am available for help.

PR #5612, opened 2020-04-27
[POC] Implement callback for dask.

trivialfis:
Proof of concept implementation for dask training callback.  As it is not consistent with single node implementation, this is an early PR for review and discussion.

The rational behind a new style of callback function is that, the original one is modelled after R implementation.  The R impl uses language environment for storing states, while Python doesn't have such facility so the original callback creates an artificial one to be consistent with R.  I believe the new style is cleaner, native to Python, and easier customize.  With the new interface, https://github.com/dmlc/xgboost/pull/5238 and #5495 can be easily implemented.

To provide a short introduction to new callback function, here is its interface:
``` python
class Callback:
    def __init__(self):
        self.history = {}
        pass

    def before_training(self, model):
        pass

    def after_training(self, model):
        pass

    def before_iteration(self, model, epoch):
        return False

    def after_iteration(self, model, epoch):
        return False
```

All the other callbacks is sub-class of this `Callback`.  This interface is suitable for both single node and distributed training.  Function return value is whether training should be stopped,  `False` if not.  By using a return value, we can deprecate `EarlyStopException`, which is created for one particular callback.

To-dos:
- [ ] Make sure no data copy is performed.
- [ ] Get correct starting rounds: https://github.com/dmlc/xgboost/pull/5511 is closed so I need to figure out a way to get correct starting round.
- [x] Be compatible with original callback function.  The new `Callback` can be a wrapper of original callback, I need to port the implementation to non-dask training.
- [ ] Support user defined metric.
- [ ] Tests.

## Future plan

If the new design is approved, I would like to deprecate the following parameters:
- In scikit-learn interface `fit` function `eval_set`, `eval_metric`, `early_stopping_rounds`, `verbose`, `sample_weight_eval_set`, 
- From `train` function `evals`, `feval`, `maximize`, `early_stopping_rounds`, `evals_result`, `verbose_eval=True` 
- lastly corresponding parameters in dask.

## Related:

#4955
#3822

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -1,4 +1,4 @@
-# pylint: disable=too-many-arguments, too-many-locals
+# pylint: disable=too-many-arguments, too-many-locals, too-many-lines
 """Dask extensions for distributed training. See
 https://xgboost.readthedocs.io/en/latest/tutorials/dask.html for simple
 tutorial.  Also xgboost/demo/dask for some examples.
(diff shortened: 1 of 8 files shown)


python-package/xgboost/dask.py
```diff
@@ -365,14 +392,314 @@ def _get_rabit_args(worker_map, client):
     rabit_args = [('%s=%s' % item).encode() for item in env.items()]
     return rabit_args
 
+
+def evaluate(model, data, name):
```
trivialfis, line 396: Note to myself: Remove it, not used.

trivialfis:
@hcho3 Can we disable the c++ header ordering lint?

hcho3:
@trivialfis Why should we disable it? It is part of Google C++ coding style.

PR #5615, opened 2020-04-28
Add example of hyperparameter optimization library

Crissman:
Add Optuna as an example of a Python hyperparameter optimization library that can be used.


Crissman:
Closing used on response to #5616 5616

PR #5616, opened 2020-04-28
Update param_tuning.rst

Crissman:
Add Optuna as an example of a Python hyperparameter optimization library that can be used.

### DIFF
--- a/doc/tutorials/param_tuning.rst
+++ b/doc/tutorials/param_tuning.rst
@@ -58,3 +58,8 @@ This can affect the training of XGBoost model, and there are two ways to improve
 
   - In such a case, you cannot re-balance the dataset
   - Set parameter ``max_delta_step`` to a finite number (say 1) to help convergence
+
+**************************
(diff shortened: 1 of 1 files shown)


RAMitchell:
I think it would be more appropriate for you to add something to this page than directly to our documentation: https://github.com/dmlc/xgboost/blob/master/demo/README.md

Also there should be at least a link to some content that shows how to use it.

hcho3:
Also we already have a working demo for Optuna: https://github.com/dmlc/xgboost/blob/master/demo/aft_survival/aft_survival_demo_with_optuna.py

Crissman:
Thanks. Both of those suggestions--demo and mention Optuna on the demos README.md--look complete, so I'll close this issue.

If you could retweet the [Optuna blog](https://twitter.com/OptunaAutoML/status/1254686886062288901) announcement in @@XGBoostProject, that would be appreciated.

Thanks!

Issue #5571, opened 2020-04-21
Extensible call back function for DMatrix.


trivialfis:
This is an issue for tracking initializing data on distributed environment.  On both Spark side and Dask side we need to concatenate partitions into a single block of data to initialize the DMatrix.  Even with the new `DeviceQuantileDMatrix` the peak memory usage is still twice the size of actual data.

One solution is to implement some sorts of call back functions for internal data adapter to process these partitions inplace to avoid concatenating data.

PR #5623, opened 2020-04-30
Move device dmatrix construction code into ellpack.

trivialfis:
* Make ellpack more self contained.

To provide some contexts, I'm implementing #5571 .  The basic procedure for GPU is:

~- Define a callback function for each input type.~
- Use that call back function to iterate through each partition 3 times:
  + Accumulate total number of rows.
  + Sketch.
  + Build an ellpack for each partition.
- Concatenate all Ellpacks by gradient sampler, like current external memory implementation.

Hence I need `EllpackImpl` to be more independent.

### DIFF
--- a/include/xgboost/data.h
+++ b/include/xgboost/data.h
@@ -350,6 +350,8 @@ class EllpackPage {
   /*! \brief Destructor. */
   ~EllpackPage();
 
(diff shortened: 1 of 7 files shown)


trivialfis:
@RAMitchell Removed the macro from header, but kept macros in `cu` files.

PR #5629, opened 2020-05-05
Use DMatrix Proxy for implementing data callback.

### LINKED ISSUE #5583: [FEA][Python] Multi-GPU training with `DeviceQuantileDMatrix`
A helpful feature would be if the newly added  `DeviceQuantileDMatrix` could be extended to Multi-GPU/Multi-node training via dask backend.

trivialfis:
@RAMitchell @hcho3 The second prototype.

Closes #5583, #5571

Please ignore the code in `quantile`, it's copied and pasted from `hist_util.cu`.

Depends on #5623 .  First prototype is in #5630 .

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -451,7 +451,73 @@ def _maybe_np_slice(data, dtype=np.float32):
     return data
 
 
-class DMatrix(object):
+class DataIter:
+    '''The interface for user defined data iterator'''
+    def __init__(self):
+        proxy_handle = ctypes.c_void_p()
+        _check_call(_LIB.XGProxyDMatrixCreate(ctypes.byref(proxy_handle)))
+        self._handle = DMatrix(proxy_handle)
+
+    @property
+    def proxy(self):
+        '''Handler of DMatrix proxy.'''
+        return self._handle
+
+    def reset_wrapper(self, this):  # pylint: disable=unused-argument
+        '''A wrapper for user defined `reset` function.'''
+        self.reset()
+
+    def next_wrapper(self, this):  # pylint: disable=unused-argument
+        '''A wrapper for user defined `next` function.
+
+        `this` is not used in Python.  ctypes can handle `self` of a Python
+        member function automatically when converting a it to c function
+        pointer.
+
+        '''
+        def data_handle(data, label=None, weight=None, base_margin=None,
+                        group=None,
+                        label_lower_bound=None, label_upper_bound=None):
+            if lazy_isinstance(data, 'cudf.core.dataframe', 'DataFrame'):
+                self.proxy.set_data_from_cuda_columnar(data)
+            elif lazy_isinstance(data, 'cupy.core.core', 'ndarray'):
+                self.proxy.set_data_from_cuda_interface(data)
+            else:
+                raise TypeError(
+                    'Value type is not supported for data iterator:' +
+                    str(type(self._handle)))
+            self.proxy.set_info(label=label, weight=weight,
+                                base_margin=base_margin,
+                                group=group,
+                                label_lower_bound=label_lower_bound,
+                                label_upper_bound=label_upper_bound)
+
+        ret = self.next(data_handle)  # pylint: disable=not-callable
+        return ret
+
+    def reset(self):
+        '''Reset the data iterator.  Prototype for user defined function.'''
+        raise NotImplementedError()
+
+    def next(self, input_data):
+        '''Set the next batch of data.
+
+        Parameters
+        ----------
+
+        data_handle: callable
+            A function with same data fields like `data`, `label` with
+            `xgboost.DMatrix`.
+
+        Returns
+        -------
+        0 if there's no more batch, otherwise 1.
+
+        '''
+        raise NotImplementedError()
+
+
+class DMatrix:
     """Data Matrix used in XGBoost.
 
     DMatrix is a internal data structure that used by XGBoost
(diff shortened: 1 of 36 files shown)


python-package/xgboost/core.py
```diff
@@ -522,12 +570,42 @@ def __init__(self, data, label=None, weight=None, base_margin=None,
...
+                ctypes.c_void_p
+            )
+            next_callback = next_factory(data.next_wrapper)
+            _check_call(_LIB.XGDMatrixCreateFromCallback(
+                None,
+                reset_callback,
+                next_callback,
+                ctypes.c_float(missing),
+                ctypes.c_int(nthread),
+                ctypes.c_int(256),
```
hcho3, line 605: Are we okay with hardcoding 256 for `max_bin` here?

python-package/xgboost/dask.py
```diff
@@ -356,6 +370,118 @@ def get_worker_data_shape(self, worker):
         return (rows, cols)
 
 
+class DaskPartitionIter(DataIter):
+    '''A data iterator for `DaskDeviceQuantileDMatrix`.
+
+    '''
+    def __init__(self, data, label=None, weight=None, base_margin=None,
+                 label_lower_bound=None, label_upper_bound=None):
+        '''Generate some random data for demostration.
```
hcho3, line 379: Does this comment make sense here? This is not a demo.

src/data/iterative_device_dmatrix.cu
```diff
@@ -0,0 +1,109 @@
...
+#include "proxy_dmatrix.h"
+#include "device_adapter.cuh"
+
+namespace xgboost {
+namespace data {
+
+#define DISPATCH_MEM(__Proxy, __Fn)                                     \
+  [](DMatrixProxy const* proxy) -> decltype(                            \
+      (dmlc::get<CupyAdapterBatch>(proxy->Value())).__Fn()) {           \
+    if (proxy->Value().type() == typeid(CupyAdapterBatch)) {            \
```
hcho3, line 20: Is it correct to say that the proxy matrix performs type erasure for the batch, and now we're trying to recover the type of the batch dynamically?

From the way I see it, we have an implicit list of allowable types, namely `CupyAdapterBatch` and `CudfAdapterBatch`. It may be better to add a type ID field in the DMatrix proxy. [Take a look at PackedFunc](https://docs.tvm.ai/dev/runtime.html#packedfunc) from TVM project, which uses type erasure to expose a generic function type. The PackedFunc uses type ID to distinguish between underlying types.

Also, is it possible to avoid macro here? Why use macro?

src/data/ellpack_page.cuh
```diff
@@ -161,8 +161,16 @@ class EllpackPageImpl {
 
   template <typename AdapterT>
   explicit EllpackPageImpl(AdapterT* adapter, float missing, bool is_dense, int nthread,
-                           int max_bin, common::Span<size_t> row_counts_span,
+                           int max_bin,
+                           common::Span<size_t> row_counts_span,
                            size_t row_stride);
+
+  template <typename AdapterBatch>
+  explicit EllpackPageImpl(AdapterBatch batch, float missing, int device, bool is_dense, int nthread,
```
hcho3, line 169: What's your rationale for adding a second interface for `EllpackPageImpl()`? Is it to specify a custom `cuts`?

src/data/ellpack_page.cu
```diff
@@ -302,6 +301,35 @@ EllpackPageImpl::EllpackPageImpl(AdapterT* adapter, float missing, bool is_dense
...
+                                 common::Span<size_t> row_counts_span,
+                                 size_t row_stride, size_t n_rows, common::HistogramCuts cuts) {
+  dh::safe_cuda(cudaSetDevice(device));
+
+  *this = EllpackPageImpl(device, cuts, is_dense, row_stride, n_rows);
+  if (is_row_major) {
+    CopyDataRowMajor(batch, this, device, missing);
+  } else {
+    // CopyDataColumnMajor(adapter, batch, this, missing);
+    LOG(FATAL) << "Not implemented";
```
hcho3, line 317: What makes the second `EllpackPageImpl` different from the first `EllpackPageImpl`, such that column major is not supported?

include/xgboost/c/callback.h
```diff
@@ -0,0 +1,87 @@
...
+
+/*!
+ * \brief Create a DMatrix from a data iterator.
+ * \param data_handle The handle to the data.
+ * \param callback The callback to get the data.
+ * \param cache_info Additional information about cache file, can be null.
+ * \param out The created DMatrix
+ * \return 0 when success, -1 when failure happens.
+ */
+XGB_DLL int XGDMatrixCreateFromDataIter(
```
hcho3, line 71: We already have a custom callback data iterator interface here, used by the JVM package. Is it correct to say that the new proxy interface is more powerful than this existing interface? I do notice that the proxy lets you specify `reset()` and `next()` callbacks.

src/data/iterative_device_dmatrix.cu
trivialfis: I will look into the pack function today.

include/xgboost/c/callback.h
trivialfis: They are different.  The original one supports only CSR input format, also it doesn't reset data.

src/data/iterative_device_dmatrix.cu
hcho3: Cool, I felt that PackedFunc may give us hint for simplifying the proxy.

include/xgboost/c/callback.h
hcho3: Thanks for the explanation.

trivialfis:
One last piece would be the dask interface then the big series will be over.

PR #5631, opened 2020-05-05
Reduce device synchronisation

RAMitchell:
Creates a data structure called Driver which is a slightly more custom priority queue, returning batches of nodes that may be executed asynchronously.

With this we don't have to wait for the device to host copy at the end of split evaluation before starting construction on the next node in many cases.

I also fixed a slight performance bug in row_partitioner where a memory copy was not asynchronous because the memory was not pinned.

### DIFF
--- a/src/common/column_matrix.h
+++ b/src/common/column_matrix.h
@@ -82,14 +82,16 @@ template <typename BinIdxType>
 class DenseColumn: public Column<BinIdxType> {
  public:
   DenseColumn(ColumnType type, common::Span<const BinIdxType> index,
-              uint32_t index_base,
-              const std::vector<bool>::const_iterator missing_flags)
+              uint32_t index_base, const std::vector<bool>& missing_flags,
+              size_t feature_offset)
       : Column<BinIdxType>(type, index, index_base),
-        missing_flags_(missing_flags) {}
-  bool IsMissing(size_t idx) const { return missing_flags_[idx]; }
+        missing_flags_(missing_flags),
+        feature_offset_(feature_offset) {}
+  bool IsMissing(size_t idx) const { return missing_flags_[feature_offset_ + idx]; }
  private:
   /* flags for missing values in dense columns */
-  std::vector<bool>::const_iterator missing_flags_;
+  const std::vector<bool>& missing_flags_;
+  size_t feature_offset_;
 };
 
 /*! \brief a collection of columns, with support for construction from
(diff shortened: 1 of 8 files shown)


RAMitchell:
@ShvetsKS the use of vector<bool> in DenseColumn seems to be causing test failures on my system and on CI.

ShvetsKS:
@RAMitchell could you give link to failed tests for reproducing?

ShvetsKS:
@RAMitchell I don't have any objections to use `feature_offset_` instead of displaced iterator if we don't hide a real problem.
But it's not clear what is exact root cause of iterator crash.

RAMitchell:
@ShvetsKS See https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-5633/2/pipeline/108

Boolean vector is a strange thing. My guess is that the iterator contains a reference to the original and something happens when the iterator gets destructed.

ShvetsKS:
@RAMitchell  thanks for fixing this issue. It was just interesting for me why we didn't see this fail before this PR.

trivialfis:
Is this issue presented in 1.1 rc ?

trivialfis:
> Boolean vector

I don't mind forbidding its usage.

RAMitchell:
Benchmarks for this PR:
| Platform | GPU    | Grow policy | Depth | Before PR (s) | This PR (s) | Speedup     |
|----------|--------|-------------|-------|---------------|-------------|-------------|
| Windows  | 1080ti | depthwise   | 6     | 9.7           | 8.5         | 1.141176471 |
| Windows  | 1080ti | depthwise   | 8     | 22.5          | 16.7        | 1.347305389 |
| Windows  | 1080ti | lossguide   | 6     | 9.7           | 9.2         | 1.054347826 |
| Windows  | 1080ti | lossguide   | 8     | 21.7          | 20.3        | 1.068965517 |
| Ubuntu   | V100   | depthwise   | 6     | 4.9           | 4.6         | 1.065217391 |
| Ubuntu   | V100   | depthwise   | 8     | 9.3           | 8.1         | 1.148148148 |
| Ubuntu   | V100   | lossguide   | 6     | 4.8           | 4.8         | 1           |
| Ubuntu   | V100   | lossguide   | 8     | 9.3           | 8.8         | 1.056818182 |

RAMitchell:
@trivialfis they may still need it as the difference in memory consumption is 8x vs vector<int8_t>.

@hcho3 we should cherry pick the changes to src/common/column_matrix.h into the release. Let me move it to a separate PR.

PR #5648, opened 2020-05-08
[for test] make xgboost v1.0.0 compatible with spark-3.0

WeichenXu123:
* Upgrade kryo to 4.0.2
* make overridden predictRaw method public
* set version to be 1.0.0-spark3


hcho3:
Closing in favor of #5836

PR #5649, opened 2020-05-09
Upgrade to CUDA 10.0

### LINKED ISSUE #5632: XGBoost 1.1.0 SNAPSHOT gpu_hist still numerically unstable
Dear everyone,

according to https://github.com/dmlc/xgboost/issues/5023 a model using gpu_hist should be reproducible using 1.1.0

The version that I was using for these experiments is:
`https://s3-us-west-2.amazonaws.com/xgboost-nightly-builds/xgboost-1.1.0_SNAPSHOT%2B86beb68ce8ffa73a90def6a9862f0e8a917b58c3-py2.py3-none-manylinux1_x86_64.whl`

The code that I am using is:
```
#!/home/user/miniconda/envs/xgboost-1.0.2-cuda-10.1/bin/python
import click
import xgboost as xgb
import numpy as np
from sklearn.datasets import fetch_covtype, load_boston
from sklearn.model_selection import train_test_split
import time
import random
import os


@click.command()
@click.option('--seed', type=int, default=0)
@click.option('--epochs', type=int, default=25)
@click.option('--no-cuda', type=bool, default=False)
def train(seed, epochs, no_cuda, dataset):
    # Fetch dataset using sklearn
    dataset == 'covertype':
    dataset = fetch_covtype()
    param = {
        'objective': 'multi:softmax',
        'num_class': 8
        # 'single_precision_histogram': True
    }

    X = dataset.data
    y = dataset.target

    # Create 0.75/0.25 train/test split
    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.25, train_size=0.75, random_state=0)

    # Set random seeds
    random_seed(seed, param)
    param['subsample'] = 0.5
    param['colsample_bytree'] = 0.5
    param['colsample_bylevel'] = 0.5

    # Convert input data from numpy to XGBoost format
    dtest = xgb.DMatrix(X_test, label=y_test)

    # Set CPU or GPU as training device
    if no_cuda:
        param['tree_method'] = 'hist'
    else:
        param['tree_method'] = 'gpu_hist'

    # Train on the chosen device
    results = {}
    gpu_runtime = time.time()
    xgb.train(param, dtrain, epochs, evals=[(dtest, 'test')], evals_result=results)
    if not no_cuda:
        print(f'GPU Run Time: {str(time.time() - gpu_runtime)} seconds')
    else:
        print(f'CPU Run Time: {str(time.time() - gpu_runtime)} seconds')

def random_seed(seed, param):
    os.environ['PYTHONHASHSEED'] = str(seed) # Python general
    np.random.seed(seed)
    random.seed(seed) # Python random
    param['seed'] = seed


if __name__ == '__main__':
    train()
```

When training the covertype dataset with 1000 epochs I get the following results:

1.  [999]	test-merror:0.05985
     GPU Run Time: 470.45096254348755 seconds
2.  [999]	test-merror:0.05978
     GPU Run Time: 493.8201196193695 seconds
3. [999]	test-merror:0.05894
    GPU Run Time: 484.21098017692566 seconds
4. [999]	test-merror:0.05962
    GPU Run Time: 477.36872577667236 seconds
5. [999]	test-merror:0.06025
    GPU Run Time: 487.09354853630066 seconds

Is this still to be expected or why do I not get perfectly reproducible results?

Thank you very much!

hcho3:
Closes #5632

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -64,7 +64,7 @@ pipeline {
             'build-cpu': { BuildCPU() },
             'build-cpu-rabit-mock': { BuildCPUMock() },
             'build-cpu-non-omp': { BuildCPUNonOmp() },
-            'build-gpu-cuda9.0': { BuildCUDA(cuda_version: '9.0') },
+            'build-gpu-cuda9.0': { BuildCUDA(cuda_version: '9.2') },
             'build-gpu-cuda10.0': { BuildCUDA(cuda_version: '10.0') },
             'build-gpu-cuda10.1': { BuildCUDA(cuda_version: '10.1') },
             'build-jvm-packages': { BuildJVMPackages(spark_version: '2.4.3') },
@@ -251,8 +251,8 @@ def BuildCUDA(args) {
     ${dockerRun} ${container_type} ${docker_binary} ${docker_args} bash -c "cd python-package && rm -rf dist/* && python setup.py bdist_wheel --universal"
     ${dockerRun} ${container_type} ${docker_binary} ${docker_args} python3 tests/ci_build/rename_whl.py python-package/dist/*.whl ${commit_id} manylinux2010_x86_64
     """
-    // Stash wheel for CUDA 9.0 target
-    if (args.cuda_version == '9.0') {
+    // Stash wheel for CUDA 9.2 target
+    if (args.cuda_version == '9.2') {
       echo 'Stashing Python wheel...'
       stash name: 'xgboost_whl_cuda9', includes: 'python-package/dist/*.whl'
       path = ("${BRANCH_NAME}" == 'master') ? '' : "${BRANCH_NAME}/"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -64,7 +64,6 @@ pipeline {
             'build-cpu': { BuildCPU() },
             'build-cpu-rabit-mock': { BuildCPUMock() },
             'build-cpu-non-omp': { BuildCPUNonOmp() },
-            'build-gpu-cuda9.0': { BuildCUDA(cuda_version: '9.2') },
             'build-gpu-cuda10.0': { BuildCUDA(cuda_version: '10.0') },
             'build-gpu-cuda10.1': { BuildCUDA(cuda_version: '10.1') },
             'build-jvm-packages': { BuildJVMPackages(spark_version: '2.4.3') },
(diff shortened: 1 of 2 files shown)


hcho3:
Build is failing for CUDA 9.2.

trivialfis:
@hcho3 No, never seen it before.  I also failed to compile with 9.0 with `redefinition of "std"` in a container before doing some workarounds (I removed C++ 11 requirement).  I assume there needs to be some workarounds for 9.2 on specific platform (docker in this case).

trivialfis:
> Or maybe we should upgrade to CUDA 10.

I would get @RAMitchell in the loop.  I don't mind upgrading the dependency, but I remember CUDA 10.0 requires applying some binary patches to fix a compiler bug that reduces member function to free function.

hcho3:
> CUDA 10.0 requires applying some binary patches to fix a compiler bug that reduces member function to free function.

We are using Docker images maintained by NVIDIA, and the images are kept up to date, including critical patches.

hcho3:
@RAMitchell Are you okay with upgrading to CUDA 10.0 with this release? We don't want to give up on deterministic GPU algorithm, which is one of the major highlights of 1.1.0 release.

hcho3:
We should also add a unit test that checks whether GPU hist is deterministic.

RAMitchell:
I think its fine, as long as we are not ahead of rapids (lowest version is 10.0).

RAMitchell:
We have determinism tests although apparently they did not trigger, @trivialfis is there a different test you can add that reproduces the issue?

hcho3:
Are the existing determinism tests different from the script in https://github.com/dmlc/xgboost/issues/5632?

hcho3:
@trivialfis @RAMitchell I verified that using CUDA 10.0 indeed fixes nondeterminism: https://github.com/dmlc/xgboost/issues/5632#issuecomment-626258605.

trivialfis:
I can add a stronger test.  But it will be tricky as actually even with 9.0 , I can't reproduce the issue on RTX 8000.  It's very platform dependent.

hcho3:
@trivialfis I couldn't reproduce it either when I was using RTX 8000. But I could reproduce it using EC2 G4 instance.

trivialfis:
@hcho3 I will try to add a stronger test later.  I don't currently have a computer at hand.

trivialfis:
Oops, it's on release branch.

PR #5651, opened 2020-05-10
Update Python demos with tests.

trivialfis:
* Remove GPU memory usage demo.
* Add tests for demos.
* Remove `silent`.
* Remove shebang as it's not portable.


demo/gpu_acceleration/memory.py
```diff
@@ -1,51 +0,0 @@
-import xgboost as xgb
```
aktech, line 1: @trivialfis I am curious to know, why this file was removed, is it not a correct demo?

trivialfis: @aktech. It's done internally so users don't have to do it themselves.

Issue #5442, opened 2020-03-26
Contribute oneAPI support in XGBoost


SmirnovEgorRu:
Hi everyone,

oneAPI is a program currently developed by Intel to have one programming model for many types of HW like CPU, GPU, FGPA and other accelerators. The base of the platform is DPC++ language. At the moment it's on Beta stage.
More details are available by links:
- https://software.intel.com/en-us/oneapi
- [oneAPI Product Brief](https://software.intel.com/en-us/download/oneapi-product-brief)
- [oneAPI Programming Guide](https://software.intel.com/en-us/oneapi-programming-guide)

My team wants to contribute oneAPI support for XGBoost. Potentially it allows to support any kind of HW (for example future Intel GPUs and HW of other vendors) without writing separate code for each of them. Now we are working on proposal how it can be.

Before showing our initial work I want to understand what XGB community think about this?

hcho3:
@SmirnovEgorRu Does it support OpenCL? It would benefit many XGBoost users if they could use OpenCL and leverage Intel and AMD graphics.

SmirnovEgorRu:
@hcho3, DPC++ is based on OpenCL, but with more high-level C++ interfaces. So, the answer is yes.

napetrov:
@hcho3, @SmirnovEgorRu   I would slightly extend this answer. 
yes, DPC++ is based on sycl and can handle OpenCL, but long-term plan is that primary focus in is DPC++ kernels and not OpenCL. 

As for different hardware support - this will really depend on runtime availability for any particular hardware platform. Plan is to get all required specification open so new platforms can be enabled. 

And overall the longterm goal of oneAPI is to cover not just GPUs but other hardware platforms from different vendors and be an open standard.

hcho3:
@napetrov @SmirnovEgorRu I should clarify my question: Can DPC++ kernels be compiled into OpenCL?

> long-term plan is that primary focus in is DPC++ kernels and not OpenCL

But eventually DPC++ kernels will have to be lowered into something like OpenCL?

hcho3:
@trivialfis @dmlc/xgboost-committer I have scant knowledge about SYCL. If one of you have any context here, feel free to chime in.

I am inclined to support this effort, as long as it has potential to benefit many users. (That's why I asked about OpenCL and Intel/AMD graphics.) We could use plugin mechanism to enable one API prototype as an experimental basis.

RAMitchell:
[...]

PR #5659, opened 2020-05-12
Introducing DPC++-based plugin (updater, predictor, objective function) supporting oneAPI programming model

vepifanov:

Updater target is grow_quantile_histmaker.

Following process intended:
1. Start with existing implementation of grow_quantile_histmaker and continuously replace OpenMP parts into DPC++ kernels.
2. As soon as we get whole flow transformed into DPC++ kernels, start analyzing potential optimizations.

Current partial stoppers:
1. DPC++ compiler is required to build plugin sources and final linkage of the library, so additional compiler/linker flags should be added to building process (beyond the plugin itself).
2. Data passing between updater/prediction/getGradient calls is wrapped into HostDeviceVector which now supports only CPU/CUDA memory, so additional host-device transfers are required on each iteration.


trivialfis:
Just out of personal curiosity, oneapi seems pretty similar to sycl, is it going to be complete different frameworks in the future or it's more about branding?  A follow up question, if there's another vendor supports sycl as khronos sycl like Codeplay or Xilinx do we need to go over this again?

Also now that we have 3 backends, we need to come up with a way to manage them.

napetrov:
@trivialfis - thanks for questions
oneAPI - is a programming model, Intel is initial contributor but this is open initiative that welcome other contributors
sycl - this is both sycl language and corresponding khronos standard
DPC++ - this is language extension on top of sycl that used by oneAPI. It have conformance to certain sycl spec  but primarily this is covered by oneAPI specification in same way as other componets such as oneDAL(Classical ML)
OneAPI spec - https://spec.oneapi.com/versions/latest/index.html 

As for other vendors support - the overall idea that this should allow simple integration for other vendors.  
there was a discussion on this with more details on overall plan
https://github.com/dmlc/xgboost/issues/5442

For backend management - yes. Should we assume this as requirement a this point? or at this is something that make sense to define then this will be moved out of plugin stage?

RAMitchell:
In the plug in stage I don't think it's a requirement, but to integrate into the main code base we will need a clear plan for multiple backends in a maintainable way. It's something to start thinking about.

trivialfis:
You don't have to write a complete implementation at one drop right?  You can start from small but critical like histogram building, and expand to some other places in the future ...

PR #5633, opened 2020-05-06
[WIP] Upgrade to C++14

hcho3:
cc @canonizer @trivialfis

PR #5664, opened 2020-05-13
C++14 for xgboost

canonizer:
C++14 for xgboost

Closes #5633


RAMitchell:
Is it possible to set `CUDA_STANDARD 14` instead of setting compiler flags?

canonizer:
Using `CUDA_STANDARD` now.

trivialfis:
Out of curiosity, I don't recall c++ 14 has that many breaking changes?  It seems this is more difficult than I was expecting.

PR #5674, opened 2020-05-17
[CI] Remove CUDA 9.0 from Windows CI.

trivialfis:


hcho3:
It won't work yet, since Jenkins will use the old AMI image which has CUDA 9.0 installed. Since we don't have Docker on Windows yet, we'll need to re-create AMI image from scratch.

trivialfis:
Got it.

PR #5679, opened 2020-05-19
[jvm-packages] default objective to binary:logistic for XGBoostClassifier

wbo4958:
Currently, the default objective for both XGBoostClassifier and
XGBoostRegressor is "reg:squarederror". 

So this PR is to default objective to binary:logistic for XGBoostClassifier.


wbo4958:
@CodingCat Could you help to review this PR?

trivialfis:
How does it work when objective is supposed to be `multi:softmax`?

wbo4958:
> How does it work when objective is supposed to be `multi:softmax`?

Hi trivialfis. the previous default value is to default classifier objective to be regressor objective. which seems not to make sense. 

This PR defaults classifier objective to be classifier related objective, we can default to multi:softmax, that' ok.

For your question, it's not working. suppose it will throw exception.

CodingCat:
Why we have to make softmax as default?

wbo4958:
> Why we have to make softmax as default?
No, we don't

trivialfis:
On Python side we actually choose between these 2 objs depending on `num_class`.  It caused some headaches (output of logistic doesn't have `prob` and `max`) but still it works.  Let's figure out a way that can be more general.

PR #5689, opened 2020-05-20
Implement Python data handler.

trivialfis:
* Add a factory method for handling various data types.

This PR decouples DMatrix Python interface with various backends.

In the future I want to use the backends for all meta info setters.  This way we have have better support for non-numpy data structures for setting metainfo.

### DIFF
--- a/python-package/xgboost/data.py
+++ b/python-package/xgboost/data.py
@@ -35,6 +35,7 @@ class DataHandler(abc.ABC):
         other XGBoost API.'''
         return data, None, None
 
+    @abc.abstractmethod
     def handle_input(self, data, feature_names, feature_types):
         '''Abstract method for handling different data input.'''
 
@@ -49,8 +50,8 @@ class DMatrixDataManager:
         '''Register a data handler handling specfic type of data.'''
         self.__data_handlers['.'.join([module, name])] = handler
 
-    def register_handler_dly(self, func, handler):
-        '''Register a data handler that handles an opaque type of data.
+    def register_handler_opaque(self, func, handler):
+        '''Register a data handler that handles data with opaque type.
 
         Parameters
         ----------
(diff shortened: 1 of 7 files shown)


python-package/xgboost/data.py
```diff
@@ -0,0 +1,624 @@
...
+                             data.shape)
+        # flatten the array by rows and ensure it is float32.  we try to avoid
+        # data copies if possible (reshape returns a view when possible and we
+        # explicitly tell np.array to try and avoid copying)
+        flatten = np.array(data.reshape(data.size), copy=False,
+                           dtype=np.float32)
+        flatten = self._maybe_np_slice(flatten, np.float32)
+        self.check_complex(data)
+        handle = ctypes.c_void_p()
+        _check_call(_LIB.XGDMatrixCreateFromMat_omp(
```
RAMitchell, line 229: We could warn on copy by checking if the data pointer is the same. I think this could be quite useful for telling people their memory usage is unnecessarily high.

python-package/xgboost/sklearn.py
```diff
@@ -536,12 +537,16 @@ def fit(self, X, y, sample_weight=None, base_margin=None,
             else:
                 params.update({'eval_metric': eval_metric})
 
-        self._Booster = train(params, train_dmatrix,
-                              self.get_num_boosting_rounds(), evals=evals,
-                              early_stopping_rounds=early_stopping_rounds,
-                              evals_result=evals_result, obj=obj, feval=feval,
-                              verbose_eval=verbose, xgb_model=xgb_model,
-                              callbacks=callbacks)
+        try:
```
RAMitchell, line 540: What's the reason for the try/catch?

trivialfis: Scikit Learn doesn't recognize XBoost error.  Exception catching is a part of its interface.

python-package/xgboost/data.py
trivialfis: Or just use array interface.

PR #5693, opened 2020-05-21
[R-package] Reduce duplication in configure.ac

jameslamb:
I've been using `xgboost`'s `configure.ac` as a reference, and noticed an opportunity for reducing duplication. Right now the location for `libomp` and the `OPENMP_CXXFLAGS` are hard-coded in two places.

This small PR proposes removing the second hard-coding to reduce duplication and the risk of mistakes if that location or those flags are updated in the future.

### DIFF
--- a/R-package/configure
+++ b/R-package/configure
@@ -613,6 +613,7 @@ infodir
 docdir
 oldincludedir
 includedir
+runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
@@ -682,6 +683,7 @@ datadir='${datarootdir}'
 sysconfdir='${prefix}/etc'
 sharedstatedir='${prefix}/com'
 localstatedir='${prefix}/var'
+runstatedir='${localstatedir}/run'
 includedir='${prefix}/include'
 oldincludedir='/usr/include'
 docdir='${datarootdir}/doc/${PACKAGE_TARNAME}'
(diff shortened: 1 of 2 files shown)


trivialfis:
Do we need to re-generate the configure script using autotools?

jameslamb:
> Do we need to re-generate the configure script using autotools?

Ah yes I can do that if you'd like! I'd be sure to use `automake` 2.62 since I see that's the version you used before.

Let me know if you'd like me to do that.

trivialfis:
Sure.  Please help generating the config file.

jameslamb:
Updated in https://github.com/dmlc/xgboost/pull/5693/commits/d9a92f2ebc5734e00ee31e7ed4dd386286619133.

I originally used `autoconf` 2.62., but it looks like the last time this file was updated, `autoconf` 2.69 was used https://github.com/dmlc/xgboost/blob/cb3ed404cff3636248b04e7081bfccfefec3e3d8/R-package/configure#L3

So I decided to use 2.69, and I updated the `AC_PREREQ` call in `configure.ac`.

I updated using `autoconf` 2.69 installed in an Ubuntu 18.04 environment, using this command

```shell
autoconf \
  --verbose \
  --output configure \
  configure.ac
```

PR #5702, opened 2020-05-25
Fix `IsDense`.

trivialfis:
* Add tests.


hcho3:
What does this do? Does it fix an existing issue?

trivialfis:
@hcho3 Yes, I'm doing the dmatrix callback, it's a bug encountered by @wbo4958 .  This fixed the order of DeviceDMatrix initialization steps where calculating `num_nonzero_` must precede calling `IsDense`.  I also added a test.

hcho3:
Thanks for clarification.

PR #5716, opened 2020-05-27
[Doc] Fix typos in AFT tutorial

hcho3:
Survival labels are always non-negative, so the left-censored label is `[0, b]`, not `(-inf, b]`.


hcho3:
@avinashbarnwal Can you review?

avinashbarnwal:
Yes. This is correct as we are inputting survival times in the model not log of survival times.

hcho3:
Thanks!

PR #5719, opened 2020-05-27
Allow pass fmap to importance plot

kongzii:
This adds the option to pass `fmap` to Pythons `plot_importance` function. So resulting graph will show actual names of features instead of default ones, if fmap is provided. 

fmapplotimportance


hcho3:
Rerunning tests

Issue #5666, opened 2020-05-14
1.1.0RC2/RC1 Performance degradation


ShvetsKS:
Slowdown was discovered in case we compare training time obtained on release version - 1.1.0rc2(1.1.0rc1)https://pypi.org/project/xgboost/#history vs custom build on head (c42f533ae992f79728020939184a7a7002324e7c) of release branch(https://github.com/dmlc/xgboost/commits/release_1.1.0) we will see next results:

dataset | 1.1.0rc2(1.1.0rc1) | custom build on c42f533ae992f79728020939184a7a7002324e7c
-- | -- | --
higgs1m |  21.33s | 16.24s 
mortgage1Q | 22.0s | 17.67s

25% slowdown in average.

custom build was obtained by default instruction:
```
mkdir build
cd build
cmake ..
make -j4
```
gcc version: 
gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0

higgs1m is public benchmark (https://github.com/dmlc/xgboost-bench/tree/master/hist_method).

**note** we can see big difference from run to run measurements on rc1/2 build:
higgs1m: 
rc1: **21.3283** sec ( [19.96872500400059, 20.516250001004664, 21.93073065700446, 22.897386002005078, 32.79936764600279] ) 
vs
custom: **16.2467** sec ( [15.885650180993252, 15.897585067999898, 16.517098495001846, 16.686414559000696, 18.001827495994803] )

hcho3:
RC2 is actually at commit 8aaabce7c9540e242dc47dd0fe6490981f35d1cf of the same release branch. Two commits have been made on top of RC2. So it seems like these two commits actually improved the performance?

hcho3:
Another possibility is the build environment. The PyPI wheels are built using CentOS 6 + GCC 5.x. So your custom build may be using native platform-specific optimization that’s not available in the PyPI build.

SmirnovEgorRu:
What I see - new PRs shouldn't impact in the performance. 
Different env is the most possible root cause. Do we use a docker container to build PyPI build? If so - how to get this?

Thank you!

hcho3:
You can run the following command at the project root:
```
tests/ci_build/ci_build.sh gpu_build docker -it --build-arg CUDA_VERSION=9.0 \
  tests/ci_build/build_via_cmake.sh -DUSE_CUDA=ON -DUSE_NCCL=ON \
[...]

PR #5720, opened 2020-05-27
Fix release degradation

ShvetsKS:
fix related to #5666

this PR reduce gaps between -DUSE_CUDA=ON and -DUSE_CUDA=OFF

dataset |  -DUSE_CUDA=ON | -DUSE_CUDA=OFF
-- | -- | --
higgs1m | 16.07s  | 15.9s
mortgage1Q | 17.9s | 17.67s

After profiling it was discovered that `GetGradient` step computational time was increased ~5x what introduce 25% degradation for full training.
Fix is related to lambda capturing (when this function was compiled with cuda, its performance was worse than with g++ usage). Without capturing performance is similar.

### DIFF
--- a/src/objective/regression_obj.cu
+++ b/src/objective/regression_obj.cu
@@ -73,18 +73,26 @@ class RegLossObj : public ObjFunction {
           << "Number of weights should be equal to number of data points.";
     }
     auto scale_pos_weight = param_.scale_pos_weight;
-    common::Transform<>::Init(
-        [=] XGBOOST_DEVICE(size_t _idx,
+    HostDeviceVector<float> scale_pos_weight_;
+    scale_pos_weight_.Resize(1);
+    scale_pos_weight_.Fill(scale_pos_weight);
+    HostDeviceVector<int> is_null_weight_;
+    is_null_weight_.Resize(1);
+    is_null_weight_.Fill(is_null_weight);
+
+    common::Transform<>::Init([] XGBOOST_DEVICE(size_t _idx,
                            common::Span<int> _label_correct,
                            common::Span<GradientPair> _out_gpair,
                            common::Span<const bst_float> _preds,
                            common::Span<const bst_float> _labels,
-                           common::Span<const bst_float> _weights) {
+                           common::Span<const bst_float> _weights,
+                           common::Span<int> _is_null_weight,
+                           common::Span<float> _scale_pos_weight) {
           bst_float p = Loss::PredTransform(_preds[_idx]);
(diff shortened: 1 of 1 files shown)


ShvetsKS:
seems failed steps are not related to current changes, as in another PR we have the same:
https://xgboost-ci.net/blue/organizations/jenkins/xgboost-win64/detail/PR-5719/1/pipeline

src/objective/regression_obj.cu
```diff
@@ -73,18 +73,26 @@ class RegLossObj : public ObjFunction {
...
-    common::Transform<>::Init(
-        [=] XGBOOST_DEVICE(size_t _idx,
+    HostDeviceVector<float> scale_pos_weight_;
+    scale_pos_weight_.Resize(1);
+    scale_pos_weight_.Fill(scale_pos_weight);
+    HostDeviceVector<int> is_null_weight_;
+    is_null_weight_.Resize(1);
+    is_null_weight_.Fill(is_null_weight);
+
+    common::Transform<>::Init([] XGBOOST_DEVICE(size_t _idx,
```
SmirnovEgorRu, line 83: Is it not enough to change [=] to [&] in lambda?
Also, if we need to set scalars to the function - do we really need use common::Span instead of just scalars?

trivialfis: You can't pass a ref of host data to device.

ShvetsKS: seems scalars are also not applicable, as it isn't aligned with common interface of `Transform::Eval`

hcho3:
Rebasing to the master branch, to incorporate latest fixes to the CI system.

hcho3:
Another alternative is to ship two shared libs in the Python wheel: `libxgboost.so` with GPU code and `libxgboost_cpu.so` with CPU code only.

trivialfis:
Just a quick update that I reproduced the slow down.

ShvetsKS:
> Just a quick update that I reproduced the slow down.

slow down for GPU with current changes?

trivialfis:
Let me try a different way to do it later today.

ShvetsKS:
> This might be really bad for GPU perf. Resize and Fill are 2 different kernels. Let me take a look later today.

@trivialfis I tried to even reduce number of resizes. Please check it (I didn't notices any degradation on GPU).
I hope I did it correctly, as `label_correct_` size is always equal to 1 (it was checked for GPU also), and we can simplify this: https://github.com/dmlc/xgboost/pull/5720/files#diff-8dda8ac25f49e7fdb3357e89fd275ab4L101

Issue #5699, opened 2020-05-22
Check failed: base_score > 0.0f && base_score < 1.0f: base_score must be in (0,1) for logistic loss, got: -0 when copy underlying booster


sh1ng:
version 1.1

Looks similar to https://github.com/dmlc/xgboost/issues/5276 

```
[2020-05-22T15:21:46.406Z] xgboost.core.XGBoostError: [15:21:45] /root/repo/xgboost/src/objective/./regression_loss.h:89: Check failed: base_score > 0.0f && base_score < 1.0f: base_score must be in (0,1) for logistic loss, got: -0
[2020-05-22T15:21:46.406Z] Stack trace:
[2020-05-22T15:21:46.406Z]   [bt] (0) /opt/h2oai/dai/cpu-only/lib/python3.6/site-packages/xgboost/lib/libxgboost.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x6a) [0x1506d9a1377a]
[2020-05-22T15:21:46.406Z]   [bt] (1) /opt/h2oai/dai/cpu-only/lib/python3.6/site-packages/xgboost/lib/libxgboost.so(xgboost::obj::LogisticRegression::ProbToMargin(float)+0xbf) [0x1506d9cf650f]
[2020-05-22T15:21:46.406Z]   [bt] (2) /opt/h2oai/dai/cpu-only/lib/python3.6/site-packages/xgboost/lib/libxgboost.so(xgboost::LearnerIO::LoadModel(dmlc::Stream*)+0x14d2) [0x1506d9b23672]
[2020-05-22T15:21:46.406Z]   [bt] (3) /opt/h2oai/dai/cpu-only/lib/python3.6/site-packages/xgboost/lib/libxgboost.so(xgboost::LearnerIO::Load(dmlc::Stream*)+0x2c4) [0x1506d9b21c84]
[2020-05-22T15:21:46.407Z]   [bt] (4) /opt/h2oai/dai/cpu-only/lib/python3.6/site-packages/xgboost/lib/libxgboost.so(XGBoosterUnserializeFromBuffer+0x4e) [0x1506d9a1a74e]
[2020-05-22T15:21:46.407Z]   [bt] (5) /opt/h2oai/dai/lib/libffi.so.6(ffi_call_unix64+0x4c) [0x1507b54c6e2c]
[...]

PR #5724, opened 2020-05-28
Fix loading old model.

trivialfis:
Quick and dirty patch for fixing https://github.com/dmlc/xgboost/issues/5699.  But as there's warning about loading old model, and can be fixed by user saving the model again before doing any further operations.  So please comment should we do the hacky fix @hcho3 @RAMitchell @sh1ng ;-)

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -694,15 +694,24 @@ class LearnerIO : public LearnerConfiguration {
       warn_old_model = false;
     }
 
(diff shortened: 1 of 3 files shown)


hcho3:
@trivialfis Should this go into 1.1.1 patch release?

trivialfis:
@hcho3 Yup.  Let me back port it.

PR #5730, opened 2020-05-30
Don't use masked array in array interface.

trivialfis:
The implementation is incorrect with due to the line `array_interface_.valid.Check(row_idx)`, it didn't show up in the test because masked array itself is not supported by cupy.  So this PR makes it explicit that we are not supporting it either.


trivialfis:
Currently, the only way to create a masked array for tensor, is to use `numpy.ma` module.  So if we were to port array interface to host code, the code needs to be reiterated.

PR #5739, opened 2020-05-31
Update RabitTracker.java

cycchow:
comment out system.out.println


trivialfis:
Could you please explain the rationale of commenting code?  We don't usually leave commented code in the repo.

cycchow:
you have written a system.out.println, which same to logger.debug, the
above line. please remove all the system.out.println, and use logger for
better logging.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/RabitTracker.java
```diff
@@ -198,8 +198,8 @@ public int waitFor(long timeout) {
       return returnVal;
     } catch (InterruptedException e) {
       // we should not get here as RabitTracker is accessed in the main thread
-      e.printStackTrace();
-      logger.error("the RabitTracker thread is terminated unexpectedly");
+
```
trivialfis, line 201:
```suggestion

```

PR #5744, opened 2020-06-01
Add pkgconfig to cmake

kongzii:
Hi,

this will add (and enable by default) ability to add xgboost.pc into the system, so xgboost library will be findable by pkg-config.

The main motivation behind this is a new [SwiftXGBoost](https://github.com/kongzii/SwiftXGBoost). 
Currently, one needs to build xgboost with custom libdir `-DCMAKE_INSTALL_LIBDIR=/usr/lib` or add xgboost.pc manually, because the swift compiler is not able to find xgboost in default installation directories, but I believe it will be useful for other use-cases too.

I tested this on my MacOS 10.15 and on Ubuntu using this dockerfile:

```
FROM ubuntu:20.04

RUN apt-get update \
    && apt-get upgrade -y

RUN apt-get install -y cmake clang

RUN apt-get install -y pkg-config

WORKDIR /xgboost

COPY . .

RUN pkg-config --list-all

RUN mkdir build \
    && cd build \
    && cmake .. \
    # && cmake .. -DINSTALL_PKGCONFIG=OFF \
    && make \
    && make install

RUN find / -type f -name xgboost.pc
RUN pkg-config --list-all | grep xgboost
RUN cat /usr/local/lib/pkgconfig/xgboost.pc
```


hcho3:
Merged. Thanks!

PR #5747, opened 2020-06-02
Expose device sketching in header.

trivialfis:
Extracted from #5629 .


src/common/hist_util.cu
```diff
@@ -22,94 +22,14 @@
 #include "../data/device_adapter.cuh"
 #include "device_helpers.cuh"
```
RAMitchell, line 23: Is there any point in keeping this file now?

trivialfis: Em, let me remove it.  Thanks!

PR #5749, opened 2020-06-02
Reorder includes.

trivialfis:
@hcho3 

Address recent update of cpplint.

### DIFF
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -2,13 +2,16 @@
 #include <dmlc/logging.h>
 #include <dmlc/omp.h>
 #include <xgboost/c_api.h>
+
+#include <cstring>
+#include <cstdio>
 #include <vector>
 #include <string>
 #include <utility>
-#include <cstring>
-#include <cstdio>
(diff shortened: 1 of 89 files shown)


hcho3:
@trivialfis It is actually possible to disable the header ordering check. Let us discuss the merits of keeping the check or not. You seemed to want to remove it, and I am interesting to hearing your opinion on this topic.

trivialfis:
@hcho3 I'm fine with both options, XGBoost doesn't rely on header macros too much so ordering might not be a real issue.  Also should we consider dmlc and rabit headers as system headers (indicated by `-isystem` instead of `-I`)?  But on the other hand having organized headers is nice.

trivialfis:
I'm confused, the CI result is different from https://github.com/dmlc/xgboost/pull/5612

hcho3:
I've purged the Docker cache, so that the build env is built from scratch.

trivialfis:
Thanks!  Will continue the PR tomorrow.

hcho3:
You have other things on your plate. If you want, I can take over this PR.

PR #5752, opened 2020-06-03
Speed up python test

RAMitchell:
This PR speeds up tests to a more reasonable level, while maintaining most of the test coverage.

`pytest -s -v --profile tests/python-gpu/test_gpu_updaters.py` goes from around 120s to 22s.

I had a failing test for DeviceQuantileDMatrix as well where passing a label with type float64 was passing doubles directly into xgboost and reading them as floats. The root cause was the data handler refactor which was incorrectly handling types in this case. I also noticed that it is possible to create a DeviceQuantileDMatrix using numpy without causing the expected error. For now I have disabled numpy inputs to DeviceQuantileDMatrix altogether.

### DIFF
--- a/tests/python/regression_test_utilities.py
+++ b/tests/python/regression_test_utilities.py
@@ -4,6 +4,8 @@ import numpy as np
 import os
 import sys
 import xgboost as xgb
+from joblib import Memory
+memory = Memory('./cachedir', verbose=0)
 
 try:
     from sklearn import datasets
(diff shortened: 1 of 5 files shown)


tests/python/regression_test_utilities.py
```diff
@@ -73,14 +83,18 @@ def get_small_weights():
     return get_weights_regression(1e-6, 1e-5)
 
 
+@memory.cache
 def get_weights_regression(min_weight, max_weight):
-    rng = np.random.RandomState(199)
```
trivialfis, line 77: Why not rng?  Quoting from help(numpy.random.seed):

```
seed(...) method of numpy.random.mtrand.RandomState instance
    seed(self, seed=None)
    
    Reseed a legacy MT19937 BitGenerator
    
    Notes
    -----
    This is a convenience, legacy function.
```

PR #5754, opened 2020-06-03
Update swift docs

kongzii:
A little update for swift documentation.
Examples are tested at https://github.com/kongzii/SwiftXGBoost/blob/master/Tests/XGBoostTests/XGBoostDocsTests.swift.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -26,7 +26,7 @@ Contents
   R package <R-package/index>
   JVM package <jvm/index>
   Ruby package <https://github.com/ankane/xgb>
-  Swift package <https://github.com/kongzii/SwiftXGBoost>
+  Swift package <swift/index>
   Julia package <julia>
   C Package <c>
   C++ Interface <c++>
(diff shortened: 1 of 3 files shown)


trivialfis:
Hi @kongzii , it seems weird that the code lives in a different repo while the tutorial lives here.  @dmlc/xgboost-committer  Do we have some polices around adding language bindings in the main repo?

kongzii:
Yes, I can understand. The idea was to have here a few code examples and a little description about the package, while main documentation and more in-depth examples are with the swift package repository.

hcho3:
@trivialfis Take a look at discussion in #4473, especially @CodingCat's remark. He suggests that **two conditions** should be met before any integration code with third-party tools gets hosted in the main XGBoost repository:
* Integration brings benefits to a substantial portion of XGBoost users.
* One of more contributors are available to maintain the integration code.

I agree with this criteria. And my opinion is that Swift binding does not yet meet the bar, so the Swift binding should continue to be hosted in kongzii/SwiftXGBoost. We also have Julia and Ruby bindings and they too are hosted in separate repositories.

kongzii:
Yeah, no problem :) I will close this and create `readthedocs` from SwiftXGBoost repository with this content. And perhaps it will be merged in the future.

PR #5755, opened 2020-06-03
Remove column major specialization.

trivialfis:
As cuDF is actually dense data, removing all the specializations.

* Use shared pointer for inplace prediction.  Using `dmlc::any` with `ArrayInterface` might lead to some subtle bugs as `dmlc::any` needs to copy the underlying value, while `ArrayInterface` contains stack allocated memory `char type[3]`.

### DIFF
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -463,8 +463,8 @@ XGB_DLL int XGBoosterPredictFromDense(BoosterHandle handle, float *values,
   CHECK_EQ(cache_id, 0) << "Cache ID is not supported yet";
   auto *learner = static_cast<xgboost::Learner *>(handle);
 
-  auto x =
-      std::make_shared<xgboost::data::DenseAdapter>(values, n_rows, n_cols);
+  std::shared_ptr<xgboost::data::DenseAdapter> x{
+    new xgboost::data::DenseAdapter(values, n_rows, n_cols)};
   HostDeviceVector<float>* p_predt { nullptr };
   std::string type { c_type };
   learner->InplacePredict(x, type, missing, &p_predt);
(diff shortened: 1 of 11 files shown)


hcho3:
Do you need help with fixing build issue on Mac OSX?

trivialfis:
@hcho3 Yes please ... Thanks!

@RAMitchell Do you want me to keep the code here?  Honestly I have never used CSC as data matrix before.

RAMitchell:
I don't think it's a big deal, we can always remove it and look up the code on version control later.

hcho3:
So the issue is that the virtual destructor of `dmlc::DataIter<T>` (parent class) is declared  `nothrow(false)`, so it can potentially throw at destruction. The C++ runtime in Mac OSX assumes that the type `T` in the expression `std::make_shared<T>`  won't ever throw exceptions.

Solution is to not use `std::make_shared<T>`.

hcho3:
Indeed removing `std::make_shared<T>` fixed the build. Shall we add a linter check to prohibit the use of `std::make_shared<T>` in CPU code?

trivialfis:
@hcho3  Thanks a lot for fixing the issue!  Also thanks for providing references for related issues.

> Shall we add a linter check to prohibit the use of std::make_shared<T> in CPU code?

One of the osx build was passing even using `std::make_shared`, I think for working around language implementation quirks we just have to do it case by case.

hcho3:
Another alternative is to define a small utility function that does what `std::make_shared<T>` does but allows throwing destructors.

trivialfis:
@hcho3 Let stay with the workaround for now.  Implementing `std::make_shared` correctly is not trivial.

PR #5760, opened 2020-06-05
Implement weighted sketching for adapter.

trivialfis:
* Implemented weighted sketching for adapter.
* Implemented a sketching interface for adapter batch instead of adapter.

The implementation is mostly for  #5629 .  I plan to replace the internal implementation of `DeviceDMatrix` with the iterative version in  #5629.  So this PR doesn't export the interface to Python.

I tried to extract some common code into functions, but still not very satisfying.  One idea for the future is we may just have an adapter for SparsePage so we can unify all sketching code.

### DIFF
--- a/src/common/hist_util.cuh
+++ b/src/common/hist_util.cuh
@@ -136,12 +136,11 @@ void GetColumnSizesScan(int device, size_t num_columns,
                          column_sizes_scan->end(), column_sizes_scan->begin());
 }
 
-template <typename AdapterBatch>
-size_t SketchBatchNumElements(AdapterBatch batch, size_t sketch_batch_num_elements,
-                              size_t columns, int device,
-                              size_t num_cuts) {
+inline size_t SketchBatchNumElements(size_t sketch_batch_num_elements,
+                                     size_t columns, int device,
+                                     size_t num_cuts, bool has_weight) {
   if (sketch_batch_num_elements == 0) {
-    int bytes_per_element = 16;
+    size_t bytes_per_element = has_weight ? 24 : 16;
     size_t bytes_cuts = num_cuts * columns * sizeof(SketchEntry);
     size_t bytes_num_columns = (columns + 1) * sizeof(size_t);
     // use up to 80% of available space
@@ -329,8 +328,8 @@ HistogramCuts AdapterDeviceSketch(AdapterT* adapter, int num_bins,
   adapter->Next();
   auto& batch = adapter->Value();
   sketch_batch_num_elements = SketchBatchNumElements(
-      batch, sketch_batch_num_elements,
-      adapter->NumColumns(), adapter->DeviceIdx(), num_cuts);
+      sketch_batch_num_elements,
+      adapter->NumColumns(), adapter->DeviceIdx(), num_cuts, false);
 
   // Enforce single batch
   CHECK(!adapter->Next());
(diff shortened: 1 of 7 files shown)


tests/cpp/common/test_hist_util.cu
```diff
@@ -210,6 +224,19 @@ TEST(HistUtil, DeviceSketchMultipleColumnsExternal) {
   }
 }
 
+template <typename Adapter>
+void ValidateBatchedCuts(Adapter adapter, int num_bins, int num_columns, int num_rows,
+                         DMatrix* dmat) {
+  common::HistogramCuts batched_cuts;
+  SketchContainer sketch_container(num_bins, num_columns, num_rows);
+  AdapterDeviceSketch(adapter.Value(), num_bins, std::numeric_limits<float>::quiet_NaN(),
```
RAMitchell, line 232: This interface should return the entire sketch not some partially computed version.

src/common/hist_util.cuh
```diff
@@ -197,12 +343,54 @@ HistogramCuts AdapterDeviceSketch(AdapterT* adapter, int num_bins,
...
+    ProcessBatch(batch, adapter->DeviceIdx(), adapter->NumColumns(),
+                 begin, end, missing, &sketch_container, num_cuts);
   }
 
   dense_cuts.Init(&sketch_container.sketches_, num_bins, adapter->NumRows());
   return cuts;
 }
+
+template <typename Batch>
+void AdapterDeviceSketch(Batch batch, int num_bins,
```
RAMitchell, line 356: This is confusing to me, there are two functions with this name that return different things and accept different inputs. Instead of writing a whole new interface for sketching from batches why don't you upgrade the previous interface that accepts an adapter so that it can iterate over multiple batches?

RAMitchell: Another suggestion would be to differentiate variable names because we have two concepts of a batch (my fault). One comes from the adapter and the other is a subset of the adapter batch. If we used something like "batch_subset" for the second it might be clearer.

trivialfis: @RAMitchell 

>  why don't you upgrade the previous interface that accepts an adapter so that it can iterate over multiple batches

We don't have multiple batches in 1 adapter.  In  #5629 we have multiple batches from an external iterator.  Surely we can make an adapter for it.  But then every function have to iterate it independently, resulting into multiple times more iterations.

tests/cpp/common/test_hist_util.cu
trivialfis: See https://github.com/dmlc/xgboost/pull/5760#discussion_r436484897 .

src/common/hist_util.cuh
trivialfis: > If we used something like "batch_subset" for the second it might be clearer.

I changed it into `SlidingWindow`.  It's a name used in literature.

trivialfis: Also, I want to replace the previous implementation of device DMatrix to the new one in  #5629 .  If this is doable, we won't be needing the old function in next PR.

RAMitchell: Oh I see, the newer version will not accept an adapter? It's less of a problem if we can get rid of the old interface.

src/data/device_adapter.cuh
```diff
@@ -52,6 +52,9 @@ class CudfAdapterBatch : public detail::NoMetaInfo {
     return {row_idx, column_idx, value};
   }
 
+  XGBOOST_DEVICE bst_row_t NumRows() const { return num_elements_ / columns_.size(); }
```
hcho3, line 55: This line is causing build failure on my Linux machine.
```
/home/phcho/tmp2/xgboost/src/common/../data/device_adapter.cuh(55): error: identifier "num_elements_" is undefined

/home/phcho/tmp2/xgboost/src/data/device_adapter.cuh(55): error: identifier "num_elements_" is undefined
```

src/common/hist_util.cuh
RAMitchell: >  But then every function have to iterate it independently, resulting into multiple times more iterations.

What functions need to operate on a batch concurrently with sketching before iterating? I thought the sketching would need to iterate over all the data before other things happen?

trivialfis: @RAMitchell I can pack multiple things in 1 iteration.  Like getting number of non-zero elements for each batch, while at the same time run sketching on that batch.

trivialfis:
@RAMitchell Memory usage tests are added.

trivialfis:
Added lower bounds for memory usage.  Add comments about how are those calculated.

PR #5761, opened 2020-06-05
python plot_importance without feature name when using np.array for training data

machineCYC:
- add features_name parameter for using numpy array to train model


trivialfis:
I think the simpler way to add feature names is just to use DataFrame, or set the feature names manually to booster:

``` python
model.get_booster().feature_names = feature_names
```

Issue #5769, opened 2020-06-09
Hessians of "multi:softmax" and "multi:softprob" appear to be incorrect


mthorrell:
Copying (and enumerating) the relevant code from: [src/objective/multiclass_obj.cu](https://github.com/dmlc/xgboost/blob/d3a0efbf162f3dceaaf684109e1178c150b32de3/src/objective/multiclass_obj.cu#L103):
```
(103) bst_float p = expf(point[k] - wmax) / static_cast<float>(wsum);
(104) const float eps = 1e-16f;
(105) const bst_float h = fmax(2.0f * p * (1.0f - p) * wt, eps);
(106) p = label == k ? p - 1.0f : p;
(107) gpair[idx * nclass + k] = GradientPair(p * wt, h);
```
If `p` is the result of the softmax operation on tree outputs, `k`, (line 103), then the gradient is `p - 1` OR `p` depending on the label (line 106).  Therefore, the diagonal of the hessian (derivative of `p` wrt the tree output, `k`) is `p * (1-p)`, NOT `2 * p * (1-p)` as line 105 says. You can verify on [wolfram alpha too](https://www.wolframalpha.com/input/?i=d2%2Fdx2%5B+log%28Exp%5Bx%5D+%2F+%28Exp%5Bx%5D+%2B+Exp%5By%5D%29+%5D). You can also see the correct calculation in line 77 of the logistic loss [here](https://github.com/dmlc/xgboost/blob/d3a0efbf162f3dceaaf684109e1178c150b32de3/src/objective/regression_loss.h).

If line 105 is fixed, you get better performance on the MNIST dataset and others. To see this yourself on a randomly generated dataset, change line 68 in this [example](https://github.com/dmlc/xgboost/blob/9c1103e06cf72de4f02a0f3ed9a8c5691e91e748/demo/guide-python/custom_softmax.py) (also, beef up `kRows = 1000` so the difference is clearer).  In line 68, just change the `2.0` to a `1.0`.

Am I missing something? Is there a reason why that `2.0f` is there that I'm unaware of?

trivialfis:
Please see https://github.com/dmlc/xgboost/issues/1825

mthorrell:
Ah got it.  Two things don't add up if the reasoning is from the MM approach (rather than a Newton step).

[...]

PR #5770, opened 2020-06-09
Softmax hess fix

mthorrell:
Addresses Issue #5769. Just need to remove the `2.0f` and the hessian calculation should be correct.


trivialfis:
It's correct, I will find the reference and reply in your issue.

PR #5771, opened 2020-06-09
Revert "Reorder includes. (#5749)"

hcho3:
Cpplint 1.5.1 reverts the changes made in 1.5.0 which necessitated re-ordering headers in #5749.

Related: https://github.com/microsoft/LightGBM/issues/3132, https://github.com/dmlc/dmlc-core/pull/624


hcho3:
If cpplint gives too much headaches in the future, we should consider pinning its version.

trivialfis:
Software has bugs, I'm fine with that fact.  We can go more closely with upstream when it happens again. :-)

Issue #3916, opened 2018-11-18
[DISCUSSION] Adopting JSON-like format as next-generation model format


trivialfis:
As discussed in #3878 and #3886 , we might want a more extendable format for saving XGBoost model.

For now my plan is utilizing the `JSONReader` and `JSONWriter` implemented in `dmlc-core` to add **experimental** support for saving/loading model into Json file.  Due to the fact that related area of code is quite messy and is dangerous to change, I want to share my plan and possibly an early PR as soon as possible so that someone could point out my mistakes earlier(there will be mistakes), and we don't make duplicated work. :)

@hcho3

hcho3:
@trivialfis Are we thinking of something like `dump_model(dump_format='json')`, except that we'd be able to both read and write the new model format? (JSON dump is currently write-only)

trivialfis:
@hcho3 `Save()` and `Load()`, with `dmlc::JSONwriter *` instead of `dmlc::stream*` as parameter.

hcho3:
@trivialfis Awesome! We can add it as a plug-in.

trivialfis:
@hcho3 My plan is we replace the old binary `Save` and `Load` in a few releases, so currently I'm trying to patch to code directly.  When it's done, XGBoost will save to json by default, as long as we keep the old binary `Load` function in the code, new releases of XGBoost can still read old binary format.  WDYT?

trivialfis:
@hcho3 Good idea.  This way we don't even need to change the interface.

hcho3:
@trivialfis Cool. I'll look forward to your pull request.

thvasilo:
I'd argue for text models, because binary models introduce long-term compatibility and maintenance burden, [case in point](https://stackoverflow.com/questions/44582581/xgboost-binary-format-spec), and recently there have been requests for adding additional parameters to the model that cannot be satisfied easily, I think we might be breaking compatibility. [[1]](https://discuss.xgboost.ai/t/treeparam-is-it-safe-to-add-new-member-variables/368) and I remember users asking for the number of samples in each leaf.

A JSON model would make it much easier to extend, having a binary specification with explicit versions would also help greatly, see e.g. [DataSketches](https://github.com/DataSketches/sketches-core-cpp/blob/master/kll/include/kll_sketch.hpp#L490). Other options like exporting/importing to JPMML or ONNX (not sure if they support decision trees) could be left for other repositories.

KOLANICH:
[...]

PR #5772, opened 2020-06-09
Implement fast numeric printing routines.

trivialfis:
* Implement ryu algorithm for printing floating point.
* Implement a fast integer printing routine.
* Remove pretty printing to simplify the JSON implementation.

For saving 7000 trees, runtime drops from 3.262300729751587 to 1.2325658798217773 .

For loading the new model using both implementation it's about 10% faster.  But the comparison isn't fair as the new output model is significantly smaller due the shorter floating point representation.

Extracted from #5046 .


src/common/charconv.h
```diff
@@ -0,0 +1,78 @@
...
+
+template <typename T>
+struct NumericLimits;
+
+template <> struct NumericLimits<float> {
+  // Unlike std::numeric_limit<float>::max_digits10, which represents the **minimum**
+  // length of base10 digits that are necessary to uniquely represent all distinct values.
+  // This value is used to represent the maximum length.  As sign bit occupies 1 character:
+  // sign + len(str(2^24)) + decimal point + `E` + sign + len(str(2^8)) + '\0'
+  static constexpr size_t kMaxDigit10Len = 16;
```
hcho3, line 30: So `std::numeric_limit<float>::max_digits10` is not sufficient? Do you have a reference to this?

src/common/charconv.h
```diff
@@ -0,0 +1,78 @@
...
+  }
+  // first write '-' and convert to unsigned, then write the rest.
+  if (value == 0) {
+    *first = '0';
+    return {std::next(first), std::errc()};
+  }
+  uint64_t unsigned_value = value;
+  if (value < 0) {
+    *first = '-';
+    std::advance(first, 1);
```
hcho3, line 71: Why not `++first`?

trivialfis: No particular reason, just trying to mimic the standard library.

src/common/charconv.h
trivialfis: I don't have a reference, I calculated myself based on the algorithm.  Here is an example `1.9310392E-38`, `std::numeric_limit<float>::max_digits10` equals to `9`, which is not sufficient to print the value.

trivialfis:
> assuming that you are using the same set of tests that the original C implementation is using.

I enlarged the tests set actually.

trivialfis:
Found some issues during brute force test.  Will continue the PR.

trivialfis:
I brute force tested all floating numbers for roundtrip.  It should be ready to get JSON out of experimental stage.

include/xgboost/json_io.h
```diff
@@ -82,9 +68,9 @@ class JsonReader {
     return GetNextChar();
   }
 
-  char GetChar(char c) {
-    char result = GetNextNonSpaceChar();
-    if (result != c) { Expect(c, result); }
+  char GetConsecutiveChar(char c) {
+    char result = GetNextChar();
+    if (XGBOOST_EXPECT(result != c, false)) { Expect(c, result); }
```
RAMitchell, line 73: What is happening here?

src/common/charconv.cc
```diff
@@ -0,0 +1,944 @@
...
+constexpr uint64_t RyuPowLogUtils::kFloatPow5Split[47];
+
+class PowerBaseComputer {
+ private:
+  static uint8_t
+  ToDecimalBase(bool const accept_bounds, uint32_t const mantissa_low_shift,
+                MantissaInteval const base2, MantissaInteval *base10,
+                bool *mantissa_low_is_trailing_zeros,
+                bool *mantissa_out_is_trailing_zeros) noexcept(true) {
+    uint8_t last_removed_digit = 0;
```
RAMitchell, line 330: Just a note that if there is a problem with code in functions like this it will be extremely difficult and time-consuming for current maintainers to fix. This is because the code is highly specialised and requires domain knowledge.

src/common/charconv.h
```diff
@@ -0,0 +1,102 @@
...
+namespace xgboost {
+int32_t ToCharsFloatImpl(float f, char * const result);
+
+struct to_chars_result {  // NOLINT
+  char* ptr;
+  std::errc ec;
+};
+
+template <typename T>
+struct NumericLimits;
```
RAMitchell, line 27: Are all of these functions/classes needed in the xgboost namespace? Can you document the functions meant to be used in other files and hide the rest in a detail namespace or some other way?

tests/cpp/common/test_json.cc
```diff
@@ -432,4 +524,35 @@ TEST(Json, IntVSFloat) {
     ASSERT_EQ(ptr, 2503595760);
   }
 }
+
+TEST(Json, RoundTrip) {
+  uint32_t i = std::numeric_limits<uint32_t>::max() / 100 * 0;
```
RAMitchell, line 529: What is happening here? A comment on what this test is doing would help me.

src/common/charconv.h
trivialfis: I hid most of the functions into detail namespace.  Sadly the limit has to be available to others as I plan to use `to_chars` and `from_chars` in other places like model dump.  The required buffer size wasn't specified by the C++ ISO, so we have to define it.

tests/cpp/common/test_json.cc
trivialfis: Removed.  Previously I placed a progress bar while running brute force test.  The upstreamed test is not full coverage in order to reduce unittest time down to 1 sec.

include/xgboost/json_io.h
trivialfis: Added comment.

trivialfis:
> The sooner we can upstream this to dmlc-core the better.

For the numeric parsing part, yes.  It will be much more useful in there since we also want roundtrip on floating point parameters.

src/common/charconv.cc
hcho3: @RAMitchell I agree. When I was reviewing this PR, I just treated `charcov.cc` as a black-box.

It all comes down to the question of whether we should implement a feature ourselves or use a third-party library. In the original RFC discussion for JSON, we decided against including third-party libraries for JSON, e.g. see comment https://github.com/dmlc/xgboost/issues/3916#issuecomment-443273481. The reason was that we wanted to build XGBoost for multiple platforms and languages, and existence of 3rd-party dependencies like RapidJSON would make that difficult.

If we really do not want to include `charconv` in the XGBoost codebase, here are some alternatives:
* Statically link the original Ryu library. This is rather difficult, since it uses Bazel for building.
* Statically link the [fmtlib library](https://github.com/fmtlib/fmt). This library is header-only, so can be used with both CMake and Makefile. The library is slated to become part of C++20, and I've had great experience using this library in Treelite.

trivialfis: 1 is a no go.  It's more like a reference implementation, I don't think it's intended for these kind of usage.  How does fmtlib parse the value back?  I thought it's just printing.  If other maintainer really want to use 3 party options I'm happy to let him choose the next steps.

hcho3: > How does fmtlib parse the value back?

It does not. We can keep the JSON parser implementation (+ parsing) and just use fmtlib for printing. My understanding is that the specialized implementation is mostly the printing of floats.

My opinion is that a large third party libs like RapidJSON is a no go. I can accept something like fmtlib. But I’d rather get charconv merged and augment it with more documentation.

Let’s wait to hear from @RAMitchell.

hcho3: The C++ ecosystem is immensely decentralized with many competing build systems: Make, CMake, Autotools, Bazel, Meson, build2 and others. This makes a great pain to integrate with third party libs, especially because XGBoost already supports two build systems (Autotools for R, CMake for all else) and DMLC-Core supports three (+GNU Make). So if a third-party lib works with one build system but not with another, we cannot use it.

trivialfis: You need to have a correctly rounding algorithm for parsing number, which is also included in this PR (part of ryu , but I shrink the double table to float table). So it's best to have a roundtrip implementation.  Besides, since the functions here are defined by C++ 17,  we can eventually use that.

hcho3: Ah so Ryu is really a round trip algorithm?

trivialfis: Recently yes.

RAMitchell: I'm okay to continue with this is the other options we have discussed are not great. I only mention this to be explicit about the risks we are taking, and that we should try to avoid 'black box' situations where possible otherwise identify them and try to mitigate the problems.

hcho3: Got it. I should make time to actually read the Ryu paper then. Let’s see if we can add more fine grained unit tests for subcomponents.

trivialfis: I think it's not too bad.  I understand the concern, that's why I went ahead and just enumerate all floating point numbers to ensure it's correctness.  Maybe it will have bug in arm or some other places that we don't have test but other than that it's 100 coverage.

hcho3: @trivialfis Can you add the code for the brute force test? Is it written in pytest or Google Test? The CI won't run it but we can run it manually from time to time.

trivialfis: @hcho3 Change this line:

https://github.com/dmlc/xgboost/blob/38ee51478740cb1aae1cfa0791168d384c388082/tests/cpp/common/test_json.cc#L548

to `i+=1`.  Right now on master branch I use sampling to reduce the test time.

hcho3: Thanks. Let me submit a small PR so that we can optionally enable the brute force test.

Issue #5293, opened 2020-02-08
don't have "Split"  "RealCover" "RealCover" col  


Huaichao2018:
https://xgboost.readthedocs.io/en/latest/R-package/discoverYourData.html
importanceRaw <- xgb.importance(feature_names = sparse_matrix@Dimnames[[2]], model = bst, data = sparse_matrix, label = output_vector)

# Cleaning for better display
importanceClean <- importanceRaw[,`:=`(Cover=NULL, Frequency=NULL)]

head(importanceClean)
## 6:              Age         53.5 0.04620627        10   0.3571429

our results is : Feature       Gain
1:              Age 0.60965369
2: TreatmentPlacebo 0.34017103
3:          SexMale 0.02340126
4:      AgeDiscret6 0.01514658
5:      AgeDiscret4 0.01162745

PR #5773, opened 2020-06-09
Added conda environment file for building docs

elliothershberg:
The formatting for the R package docs is [broken](https://xgboost.readthedocs.io/en/latest/R-package/xgboostPresentation.html), as mentioned in #5293. I want to work on fixing this. To get started, I followed the [instructions for building the docs](https://xgboost.readthedocs.io/en/latest/R-package/xgboostPresentation.html). I had a few issues getting the dependencies installed correctly, and figured I would share my conda environment to make building the docs easier.

If this PR is accepted, I can update the instructions to mention how to use this environment file to install the dependencies for the docs. I will likely address the issues with the R package formatting in the original issue as well.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -6,3 +6,4 @@ sh>=1.12.14
 matplotlib>=2.1
 graphviz
 numpy
+recommonmark
--- a/doc/xgboost_doc.yml
+++ b/doc/xgboost_doc.yml
@@ -1,0 +1,15 @@
+name: xgboost_docs
+dependencies:
+  - python
+  - pip
+  - pygraphviz
+  - sphinx
+  - recommonmark
+  - mock
+  - sh
+  - matplotlib
+  - pip:
+    - breathe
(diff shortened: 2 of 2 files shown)


hcho3:
We already have a `requirements.txt` file: https://github.com/dmlc/xgboost/blob/master/doc/requirements.txt. Have you tried it?

elliothershberg:
To be honest at first I didn't see that 🙈 . I was just following the instructions in the docs and didn't notice the requirements.txt. But I just tried building using the requirements.txt and there are missing requirements that cause the build to fail. recommonmark isn't included in the requirements file.

Would you rather I just work on updating that file instead of using a conda environment?

hcho3:
Actually, let's offer both options, since some prefer to use pip and others Conda. Can you update the requirements to include recommonmark?

elliothershberg:
Sounds good. Yeah I can fix the requirements.txt to be up to date 👍

PR #5774, opened 2020-06-10
[jvm-packages] Switch Task mode to OpenMP mode after data conversion

FelixYBW:
One big issue of current xgboost training on Spark is that ETL usually run in task mode which sets spark.task.cpus = 1. But xgboost training is more efficient on openMP mode which sets spark.task.cpus = spark.executor.cores = xgbparam.nthread. We usually save the ETL result into a parquet file, then start a new spark context with task.cpus=nthreads, then we load the data, convert into DMatrix, start training. In this way the conversion has to run in little threads and takes very long time. e.g. a conversion of 150M Row x 46 columns takes about 870s on a 2x28 config, takes ~250s on a 7x8 config.  Even worse, the inference is more efficient on task mode, so we have to save the model and start another new context to do inference.

It's really not a good solution. The patch gives a way to swich task mode to OMP mode AFTER data conversion, then the training run in OMP mode, then the inference goes back to task mode. It solved the model swtich issue, accelerately the data conversion process greatly and be able to use one or two executor per node which is more efficient for training.

The basic logic is as below:
    1. repartition input data set to N x executor#. N is integer.
    2. Create DMatrix in each task thread of executor
    3. Force caching the Watches[DMatrix*] in executor's block manager
    4. Use a customized coalescePartitioner to combine the Watches[DMatrix*] into a large DMatrix. Set preferred location to the executor
    5. Train xgboost from the large DMatrix in task thread of an executor

Because we cached the pointer of DMatrix in block manager instead of the converted data itself, so we must make sure the training stage must be scheduled to the same executor. we can set spark.locality.wait.process to a large number for this. Spark3.0's stage level resource management can change task.cpus at run time but it doesn't solve the issue automatically. We still need to combine the converted DMatrix of each executor into a large one. It either can't solve the data locality issue, we still need to set wait time a large number.

Here is the result, with 7x8 configuration, the patch can decrease the data conversion time from 250s to 70s. with 2x28 configuration, it can decrease from 870s to 160s. It's possible to further optimize later.

image

It only support trainForNonRanking and evalSetsMap.isEmpty now. I will add other 3 cases.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostConfigureSuite.scala
+++ b/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostConfigureSuite.scala
@@ -29,15 +29,19 @@ class XGBoostConfigureSuite extends FunSuite with PerTest {
       .config("spark.serializer", "org.apache.spark.serializer.KryoSerializer")
       .config("spark.kryo.classesToRegister", classOf[Booster].getName)
 
-  test("nthread configuration must be no larger than spark.task.cpus") {
-    val training = buildDataFrame(Classification.train)
-    val paramMap = Map("eta" -> "1", "max_depth" -> "2", "verbosity" -> "1",
-      "objective" -> "binary:logistic", "num_workers" -> numWorkers,
-      "nthread" -> (sc.getConf.getInt("spark.task.cpus", 1) + 1))
-    intercept[IllegalArgumentException] {
-      new XGBoostClassifier(paramMap ++ Seq("num_round" -> 2)).fit(training)
-    }
-  }
+  /*
+   * Disable this test because PR 5774 allows nthread to be larger than
+   * spark.task.cpus.
+   */
+  // test("nthread configuration must be no larger than spark.task.cpus") {
+  //   val training = buildDataFrame(Classification.train)
+  //   val paramMap = Map("eta" -> "1", "max_depth" -> "2", "verbosity" -> "1",
+  //     "objective" -> "binary:logistic", "num_workers" -> numWorkers,
+  //     "nthread" -> (sc.getConf.getInt("spark.task.cpus", 1) + 1))
+  //   intercept[IllegalArgumentException] {
+  //     new XGBoostClassifier(paramMap ++ Seq("num_round" -> 2)).fit(training)
+  //   }
+  // }
 
   test("kryoSerializer test") {
     // TODO write an isolated test for Booster.
--- a/doc/jvm/xgboost4j_spark_tutorial.rst
+++ b/doc/jvm/xgboost4j_spark_tutorial.rst
@@ -492,10 +492,11 @@ This parameter controls how many parallel workers we want to have when training 
   By default, we allocate a core per each XGBoost worker. Therefore, the OpenMP optimization within each XGBoost worker does not take effect and the parallelization of training is achieved
   by running multiple workers (i.e. Spark tasks) at the same time.
 
-  If you do want OpenMP optimization, you have to
+  On the other hand, OpenMP parallelization can improve performance for training tasks on CPUs, especially when using large datasets. If you do want OpenMP parallelization, you can
 
   1. set ``nthread`` to a value larger than 1 when creating XGBoostClassifier/XGBoostRegressor
-  2. set ``spark.task.cpus`` in Spark to the same value as ``nthread``
+  2. set ``spark.executor.cores`` in Spark to the same value as ``nthread``
+  3. set ``spark.task.cpus`` in Spark to 1.
 
 Gang Scheduling
 ===============
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -364,6 +364,20 @@ XGB_DLL int XGDMatrixSliceDMatrixEx(DMatrixHandle handle,
                                     bst_ulong len,
                                     DMatrixHandle *out,
                                     int allow_groups);
+
+
+/*!
+ * \brief combine right dmatrix into left, return the left
+ * \param handle_left instance of data matrix to be merged into
+ * \param handle right instance of data matrix to be merged
+ * \param out the left instance of data matrix
+ * \return 0 when success, -1 when failure happens
+ */
+XGB_DLL int XGDMatrixCombineDMatrix(DMatrixHandle handle_left,
+                                  DMatrixHandle handle_right,
+                                  uint64_t total_size,
+                                  DMatrixHandle* out);
+
 /*!
  * \brief free space in data matrix
  * \return 0 when success, -1 when failure happens
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -569,6 +569,14 @@ XGB_DLL int XGDMatrixGetUIntInfo(const DMatrixHandle handle,
 XGB_DLL int XGDMatrixNumRow(DMatrixHandle handle,
                             bst_ulong *out);
 /*!
+ * \brief get data vector size.
+ * \param handle the handle to the DMatrix
+ * \param out The address to hold data vector size
+ * \return 0 when success, -1 when failure happens
+ */
+XGB_DLL int XGDMatrixDataVecSize(DMatrixHandle handle,
+                            bst_ulong *out);
+/*!
  * \brief get number of columns
  * \param handle the handle to the DMatrix
  * \param out The output of number of columns
--- a/include/xgboost/data.h
+++ b/include/xgboost/data.h
@@ -187,7 +187,8 @@ struct Entry {
   /*! \brief feature value */
   bst_float fvalue;
   /*! \brief default constructor */
-  Entry(){};
+  Entry() {
+  }
   /*!
    * \brief constructor with index and value
    * \param index The feature or row index.
--- a/include/xgboost/data.h
+++ b/include/xgboost/data.h
@@ -187,8 +187,7 @@ struct Entry {
   /*! \brief feature value */
   bst_float fvalue;
   /*! \brief default constructor */
-  Entry() {
-  }
+  Entry() = default;
   /*!
    * \brief constructor with index and value
    * \param index The feature or row index.
--- a/include/xgboost/data.h
+++ b/include/xgboost/data.h
@@ -187,7 +187,7 @@ struct Entry {
   /*! \brief feature value */
   bst_float fvalue;
   /*! \brief default constructor */
-  Entry() = default;
+  Entry(){};
   /*!
    * \brief constructor with index and value
    * \param index The feature or row index.
@@ -549,6 +549,11 @@ class DMatrix {
                          int max_bin);
 
   virtual DMatrix *Slice(common::Span<int32_t const> ridxs) = 0;
+  virtual DMatrix* Combine(DMatrix* right, uint64_t total_size) {
+    CHECK(false) << " this type of DMatrix not supported";
+    return this;
+  }
+
   /*! \brief page size 32 MB */
   static const size_t kPageSize = 32UL << 20UL;
 
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
@@ -92,10 +92,15 @@ object DataUtils extends Serializable {
       }
     } else {
       arrayOfRDDs.map(rdd => {
-        if (rdd.getNumPartitions % numWorkers != 0 ) {
-          val partitions = rdd.getNumPartitions +
-            numWorkers - (rdd.getNumPartitions % numWorkers)
-          rdd.map(_._2).repartition(partitions)
+        val nPartitions = if (rdd.sparkContext.isLocal) numWorkers else {
+          if (rdd.getNumPartitions % numWorkers != 0) {
+            rdd.getNumPartitions + numWorkers - (rdd.getNumPartitions % numWorkers)
+          } else {
+            rdd.getNumPartitions
+          }
+        }
+        if (rdd.getNumPartitions != nPartitions) {
+          rdd.map(_._2).repartition(nPartitions)
         } else {
           rdd.map(_._2)
         }
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
@@ -92,8 +92,10 @@ object DataUtils extends Serializable {
       }
     } else {
       arrayOfRDDs.map(rdd => {
-        if (rdd.getNumPartitions != numWorkers) {
-          rdd.map(_._2).repartition(numWorkers)
+        if (rdd.getNumPartitions % numWorkers != 0 ) {
+          val partitions = rdd.getNumPartitions +
+            numWorkers - (rdd.getNumPartitions % numWorkers)
+          rdd.map(_._2).repartition(partitions)
         } else {
           rdd.map(_._2)
         }
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -514,20 +514,22 @@ object XGBoost extends Serializable {
           partitionCoalescer = Some(new ExecutorInProcessCoalescePartitioner()))
 
       val reducedrdd = coalescedrdd.mapPartitions { iter =>
-          val totalsize = iter.foldLeft(Map("train" -> 0L, "test" -> 0L)) {
+          val matcharr = iter.toArray
+          val totalsize = matcharr.foldLeft(Map("train" -> 0L, "test" -> 0L)) {
            (l, r) =>
              val merged = l.toSeq ++ r.rowNumMap.toSeq
              merged.groupBy(_._1).mapValues(_.map(_._2).sum)
           }
           totalsize.foreach( iter => System.out.println("xgbtck reduce_total "
             + iter._1 + " " + iter._2 ))
-          Iterator( iter.reduce { (l, r) =>
+          Iterator( matcharr.reduce { (l, r) =>
             val rst = l.combine(r, totalsize)
             l.delete()
             r.delete()
             rst
             }
             )}.cache()
+
 //      reducedrdd.foreachPartition(() => _)
       watchrdd.unpersist()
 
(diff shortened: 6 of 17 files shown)


FelixYBW:
Added Ranking support.

zhangzhang10:
@hcho3 Hi can you please help take a look at PR5774? One of the CI checks (jenkins pr-merge) failed due to "time out" error. How shall we fix such a problem? Thanks!

hcho3:
@zhangzhang10 It probably is due to hitting [the daily budget limit in the CI server](https://xgboost-ci.net/dashboard/). I restarted the tests.

hcho3:
@zhangzhang10 Actually it appears that this unit test in the JVM package is hanging:
```
[2020-08-28T17:30:10.151Z] Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=172.17.0.3, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=4}
[2020-08-28T17:30:10.151Z]  xgbtck executorcoalesce Some(executor_d7f5674fd1f3_driver) : 0 1 2 3
```
We need to fix this test for this PR to be merged.

hcho3:
This is the test that's hanging:
https://github.com/dmlc/xgboost/blob/d7b45fbcafcce8295a94858a996482fdb9091f3e/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/ParameterSuite.scala#L26-L41

zhangzhang10:
@hcho3 Thank you for catching the failed test. I'll fix it!

CodingCat:
@FelixYBW thanks for the contribution, but I don't think this is a solution to handle this problem and this should be a problem handled in XGBoost side. My concerns come from 

1. "is saving transformed data to Parquet really a bad idea"?

When I started this project in school, I also thought eliminating the intermediate data is good. However, in prod case, training of XGB is "much more fragile" than ETL processing itself. Spot instance, preemption and various hw faults can take down anyone of your workers, and XGBoost training will just crash. If you couple ETL and training together, that means, you need to redo ETL in the case of a failed training .


2. IIUC, the key thing to make your method work is to schedule task to a particular executor saving DMatrix pointer. To make this happen, you are setting `spark.locality.wait.process to a large number for this`

first of all, even you set spark.locality.wait.process  to a large number, there is no guarantee that your task will eventually get scheduled to that executor. 

second, setting spark.locality.wait.process to a super large value will hurt performance of ETL in storage-compute disaggregated architecture (which is the trend in industry). In such an architecture, we usually set spark.locality.wait to 0s to avoid wasting of compute resources for pursuing locality which doesn't exist at all.

3. I do not think as a ML framework, XGBoost/XGBoost-Spark should tackle this problem. The underlying framework like Spark should consider per-stage resource claim , just like they should consider only allocating GPU to a particular stage of Spark applications instead of letting XGBoost-Spark to handle this for them.

BTW, I don't think Spark 3.0 has been able to set per-stage cpu core, the PR hasn't been merged (I might be wrong)

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostConfigureSuite.scala
```diff
@@ -29,15 +29,19 @@ class XGBoostConfigureSuite extends FunSuite with PerTest {
...
-    val training = buildDataFrame(Classification.train)
-    val paramMap = Map("eta" -> "1", "max_depth" -> "2", "verbosity" -> "1",
-      "objective" -> "binary:logistic", "num_workers" -> numWorkers,
-      "nthread" -> (sc.getConf.getInt("spark.task.cpus", 1) + 1))
-    intercept[IllegalArgumentException] {
-      new XGBoostClassifier(paramMap ++ Seq("num_round" -> 2)).fit(training)
-    }
-  }
+  /*
+   * Disable this test because PR 5774 allows nthread to be larger than
```
CodingCat, line 33: this is wrong. if you use more than spark.task.cpus cores in each worker, that means you silently oversubscribe cluster resources

zhangzhang10:
@CodingCat, you may have misunderstood this PR. Although @FelixYBW 's original comment mentioned ETL, the main point of this PR isn't about ETL or about doing away intermediate data. The main point is to provide improved overall training performance when OpenMP optimization is enabled. On Spark, OpenMP optimization is enabled by allocating multiple threads for each training task (controlled by parameter `nthread`). But the step before training, i.e. constructing DMatrix data structure from input data is still slow because it is not as well parallelized. With this PR, we can make the DMatrix construction step run faster by using a larger number of single-threaded tasks. Then, the DMatrix data structures created by tasks on the same executor are combined. The training step then takes advantage of OpenMP by using multi-threaded tasks, albeit a fewer number of them. 

From a resource management point of view, this PR, while maintaining the same amount of resource (i.e. compute cores), automatically shifts from using many single-threaded tasks to create DMatrix, to using fewer multi-threaded task for training. This way, it aims to provide a better overall training experience.

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostConfigureSuite.scala
zhangzhang10: For this PR to work, the XGBoost parameter `nthread` needs to be the same as `spark.executor.cores`, and `spark.task.cpus` is set to 1. This means `nthread` has to be larger than `spark.task.cpus`. This PR works by having one task per executor. This should not lead to oversubscription.

CodingCat: so I am seeing some assumptions here

1. users has to set spark.task.cpus to 1 to use xgb-spark

2. you assign one core to each executor process and actually fork many threads in each jvm process, and each thread are essentially compute-intensive for xgb task, what's the point to do that?

CodingCat:
> But the step before training, i.e. constructing DMatrix data structure from input data is still slow because it is not as well parallelized. With this PR, we can make the DMatrix construction step run faster by using a larger number of single-threaded tasks. Then, the DMatrix data structures created by tasks on the same executor are combined. The training step then takes advantage of OpenMP by using multi-threaded tasks, albeit a fewer number of them.

sure, but as I said, "even you set spark.locality.wait.process to a large number, there is no guarantee that your task will eventually get scheduled to that executor." and `setting spark.locality.wait.process to a super large value will hurt performance of ETL in storage-compute disaggregated architecture (which is the trend in industry).`



from the description `a conversion of 150M Row x 46 columns takes about 870s on a 2x28 config, ` training 150M on a single task seems pretty heavy, why not assigning more workers?

zhangzhang10:
> from the description `a conversion of 150M Row x 46 columns takes about 870s on a 2x28 config, ` training 150M on a single task seems pretty heavy, why not assigning more workers?

Let me answer this question first. The reason for not assigning more workers is to keep a lower memory overhead, so more memory can be given to the training tasks. Consequently, we can train a larger dataset this way. The advantage can be significant. We carefully benchmarked memory consumptions using a random dataset with 50 features to see how big a dataset we can train. The result is shown below (the experiments were done on an Intel Xeon 8180 node with 56 cores and 380GB memory):

|Config (num_workers x cores_per_worker)   |Max data size (num_rows)   |Peak mem (GB)   |
|:-:|:-:|:-:|
|2x28   |550 million   |360   |
|7x8   |300 million   |300   |

Using the 2x28 config we can train a dataset 1.8x larger than what we can train using the 7x8 config. The incentive for using fewer workers is obvious. However, without this PR the 2x28 config means a very long data conversion time, which diminishes its benefit. With this PR the data conversion time is substantially shortened so that users can easily reap the benefit.

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostConfigureSuite.scala
zhangzhang10: We are assigning `spark.executor.cores` cores to each executor. The value is typically larger than 1. For example, in the 7x8 config we assign 8 cores to each of the 7 executors. With `spark.task.cpu=1` we can start 8 single-threaded tasks on an executor. They create 8 per-task DMatrices. These DMatrices are then combined into one large DMatrix and the 8 tasks stop. Next, one training task starts on the executor, which has 8 cores at its disposal. The training task can now rely on OpenMP (starting 8 OpenMP threads) to achieve good parallelism. There aren't oversubscription issues. It is just the same resources being used in different ways in different stages. This is the essence of this  PR. Thanks.

carsonwang:
> sure, but as I said, "even you set spark.locality.wait.process to a large number, there is no guarantee that your task will eventually get scheduled to that executor." and `setting spark.locality.wait.process to a super large value will hurt performance of ETL in storage-compute disaggregated architecture (which is the trend in industry).`

@CodingCat , the tasks should be scheduled correctly even if the locality wait is 0.  Setting spark.locality.wait.process to 0 means the [getAllowedLocalityLevel](https://github.com/apache/spark/blob/0811666ab104b41cf189233439f4158b18bc8282/core/src/main/scala/org/apache/spark/scheduler/TaskSetManager.scala#L434) will always be ANY. But Spark schedules the tasks in [increasing order of locality levels](https://github.com/apache/spark/blob/0811666ab104b41cf189233439f4158b18bc8282/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala#L579) which is controlled by `maxLocality` and it will start from PROCESS_LOCAL. For example, if we have N executors and each executor has M cores. In the training stage, there are only N tasks and every task has a preferred location (executor). In this case, Spark will first schedule tasks using maxLocality = PROCESS_LOCAL. Even though  spark.locality.wait.process is 0, all N tasks will be scheduled to the preferred location correctly with the maxLocality constraint. Please correct if anything was missing here.

CodingCat:
> > from the description `a conversion of 150M Row x 46 columns takes about 870s on a 2x28 config, ` training 150M on a single task seems pretty heavy, why not assigning more workers?
> 
> Let me answer this question first. The reason for not assigning more workers is to keep a lower memory overhead, so more memory can be given to the training tasks. Consequently, we can train a larger dataset this way. The advantage can be significant. We carefully benchmarked memory consumptions using a random dataset with 50 features to see how big a dataset we can train. The result is shown below (the experiments were done on an Intel Xeon 8180 node with 56 cores and 380GB memory):
> 
> Config (num_workers x cores_per_worker)	Max data size (num_rows)	Peak mem (GB)
> 2x28	550 million	360
> 7x8	300 million	300
> Using the 2x28 config we can train a dataset 1.8x larger than what we can train using the 7x8 config. The incentive for using fewer workers is obvious. However, without this PR the 2x28 config means a very long data conversion time, which diminishes its benefit. With this PR the data conversion time is substantially shortened so that users can easily reap the benefit.

there are some missing information here

> Using the 2x28 config we can train a dataset 1.8x larger than what we can train using the 7x8 config. The incentive for using fewer workers is obvious.

the peak memory usage is 1.2 larger, right? using your word, this diminishes the benefit.

> without this PR the 2x28 config means a very long data conversion time

sure, can you quantify "very long" in an e2e way, how long it takes for overall training (# of iterations?),  is data conversion the bottleneck?

>  2x28 

are we sure embedding so many cores to a single worker is a good idea? https://github.com/szilard/GBM-perf/issues/41#issuecomment-696029422

CodingCat:
> > sure, but as I said, "even you set spark.locality.wait.process to a large number, there is no guarantee that your task will eventually get scheduled to that executor." and `setting spark.locality.wait.process to a super large value will hurt performance of ETL in storage-compute disaggregated architecture (which is the trend in industry).`
> 
> @CodingCat , the tasks should be scheduled correctly even if the locality wait is 0. Setting spark.locality.wait to 0 means the [getAllowedLocalityLevel](https://github.com/apache/spark/blob/0811666ab104b41cf189233439f4158b18bc8282/core/src/main/scala/org/apache/spark/scheduler/TaskSetManager.scala#L434) will always be ANY. But Spark schedules the tasks in [increasing order of locality levels](https://github.com/apache/spark/blob/0811666ab104b41cf189233439f4158b18bc8282/core/src/main/scala/org/apache/spark/scheduler/TaskSchedulerImpl.scala#L579) which is controlled by `maxLocality` and it will start from PROCESS_LOCAL. For example, if we have N executors and each executor has M cores. In the training stage, there are only N tasks and every task has a preferred location (executor). In this case, Spark will first schedule tasks using maxLocality = PROCESS_LOCAL. Even though spark.locality.wait is 0, all N tasks will be scheduled to the preferred location correctly with the maxLocality constraint. Please correct if anything was missing here.


I am a bit confused here,  the author of the PR said that we need to set localitywait to a very big value, but seems you are arguing this is not necessary?

in your comments, 

> In the training stage, there are only N tasks and every task has a preferred location 

this is another consumption, with this PR, you have to 

1. guarantee all executors saving  dmatrix pointer cannot be used by any other tasks other than the xgb worker

2. number of executors have to be larger than # of workers

to me, the first 1 can be easily broken if you have an interactive session where you triggered two jobs (maybe one for training, one for whatever reason) in a spark application asynchronously ...

the second one is a breaking change, it seems it completely disabled the use case that users want to locate >1 worker in a single executor

CodingCat:
so my general comments are that 

1. this PR is an very interesting to improve  some part of the e2e performance, but I don't think it is necessary given the cost we have to pay as well as the realistic constraints  e.g. the benchmark suggests packing 150MM rows in a single worker with 28 cores even multi-core scalability of xgb hasn't been addressed very well

2. the cost of the PR is that the behavior of XGB-Spark becomes hacky, e.g. (just an example, not comprehensive list) with this PR you will have an Spark application with spark.task.cpus = 1 but will trigger many threads internally at some phase

3. there are some assumptions which are not solid to me and can make XGB-Spark very fragile 


if you don't mind, please close the PR

FelixYBW:
@CodingCat I proposed the PR but @zhangzhang10 and @carsonwang actually working on this. This is why I didn't reply in time.

Some background here: The initial design of this PR is to solve a real world issue of one of our customer. In their usage, the ETL, training and prediction do run in the same spark context. We noted 1 xgb worker with multiple OMP threads on each node can train larger dataset.  But if we set the spark.task.cpus to nthread, the ETL and prediction work can't use all the CPU resource. If we set spark.task.cpus to 1 then use a repartition (worker#) before training, it actually can still work because spark prefer to schedule a task to executors in round robin by default. But the repartition and data conversion itself still need to complete in the single task threads which is a big hotspot with large dataset. 

 > I am a bit confused here, the author of the PR said that we need to set localitywait to a very big value, but seems you are arguing this is not necessary?

@CodingCat  You are right, My initial assumption of localitywait is wrong. But later @carsonwang confirmed it's not necessary on current spark. The only risk is if we start two jobs on the same spark context. But question is: does user really share the training spark context with others currently, especially with current solution where a user needs to set task.cpus.

> 
> in your comments,
> 
> > In the training stage, there are only N tasks and every task has a preferred location
> 
> this is another assumption, with this PR, you have to
> 
> 1. guarantee all executors saving  dmatrix pointer cannot be used by any other tasks other than the xgb worker
> 2. number of executors have to be larger than # of workers
> 
> to me, the first 1 can be easily broken if you have an interactive session or you triggered two jobs (maybe one for training, one for whatever reason) in a spark application asynchronously ...

The dmatrix pointer shouldn't be reused by other tasks other than the xgb worker. It's transformed for xgb training only.  The conversion itself is still part of the training. We didn't expose the dmatrix pointers.

> 
> the second one is a breaking change, it seems it completely disabled the use case that users want to locate >1 worker in a single executor

The basic idea here is a single rank on a node can be more efficient with performance and memory usage than multiple ranks. OMP threads is more efficient than rank's parallelism. We have proved both except that single rank has a performance issue of gmat.init which we are still optimizing. 

On the other side, it only make sense to locate >1 workers in one executor is the spark local mode as in the CI tests or for development. In this case, we use default coalesce to worker#, so we can locate >1 workers. On cluster mode, if user really need to try multiple ranks per node, it makes more sense to create multiple executors and use one worker per executor.

FelixYBW:
> so my general comments are that
> 
> 1. this PR is an very interesting to improve  some part of the e2e performance, but I don't think it is necessary given the cost we have to pay as well as the realistic constraints  e.g. the benchmark suggests packing 150MM rows in a single worker with 28 cores even multi-core scalability of xgb hasn't been addressed very well
> 2. the cost of the PR is that the behavior of XGB-Spark becomes hacky, e.g. (just an example, not comprehensive list) with this PR you will have an Spark application with spark.task.cpus = 1 but will trigger many threads internally at some phase

I don't see any cost of this PR compared to current solution. In stead setting task.cpus setting to nthread is an extra requirement. its default value is 1. The threads it triggers are the nthreads parameter. It's the same as current solution. In future, if dynamic resource allocation works, we can add that to the PR to solve the scheduling issue better.

> 3. there are some assumptions which are not solid to me and can make XGB-Spark very fragile

Can you create the cases which works on current solution but break in this PR? If so we can propose more comprehensive solution.

CodingCat:
@FelixYBW 

> ETL, training and prediction do run in the same spark context. 

ok, please check https://github.com/dmlc/xgboost/pull/5774#issuecomment-683476042, I think I was on the same page with you in terms of what is this PR doing but argued that, if you do large scale ETL, you should not put it in the same sparkcontext with training, later @zhangzhang10 said this PR is not about ETL or intermediate data. Could you guys get some agreement first?

>  does user really share the training spark context with others currently, especially with current solution where a user needs to set task.cpus.

they (including me) don't share sparkcontext with others, but in the same context, they may start multiple jobs asynchronously thus sharing the same set of executors

> The dmatrix pointer shouldn't be reused by other tasks other than the xgb worker. It's transformed for xgb training only. The conversion itself is still part of the training. We didn't expose the dmatrix pointers.

"Because we cached the pointer of DMatrix in block manager instead of the converted data itself, so we must make sure the training stage must be scheduled to the same executor. " this is what you mentioned in your description

the biggest risk here is, you have no 100% guarantee workers can only be scheduled to the executor storing the pointer

> The basic idea here is a single rank on a node can be more efficient with performance and memory usage than multiple ranks.

true, this PR tends on pack as many cores to the same worker as possible to "improve performance", but  you didn't answer my previous question, "did we resolve scalability of xgb on multi cores (>8) arch"?


> On the other side, it only make sense to locate >1 workers in one executor is the spark local mode as in the CI tests or for development. 

I am not saying that users schedule more workers to the same executor is in best practice or not, I am saying, it is a breaking change, we will release a version which breaks all users' system which originally run with, e.g., 10 executors (2 core per each, and set nworkers as 20), 


> I don't see any cost of this PR compared to current solution

every change comes with a cost which could be maintenance, breaking change, etc.. Paricularly this PR, as I said, involves many assumptions which are not solid, and breaking changes, I can give two costs which I have mentioned in some of my comments 1. the behavior of XGB is totally changed in a breaking way, 2. the reliability of XGB is compromised due to the assumption like you have to schedule task with exact PROCESS_LOCAL....

> task.cpus setting to nthread is an extra requirement

I hope you have read my previous comments , spark.task.cpus  means how many cores you allocated to each spark task, now, you are making the behavior of spark application inconsistent with the configuration. If you give me a Spark UI, it shows spark.task.cpus  = 1, I will assume every task will always spread 1 task, but now, I need to figure out, "oh, this is a XGB app, yeah, XGB changed its behavior at some version, each task could spread many threads even you saw spark.task.cpus = 1"

> Can you create the cases which works on current solution but break in this PR? If so we can propose more comprehensive solution.

I think I have been very clear, this PR is not in the right tech path, I would not spend too much time to rewrite the list (I have mentioned them in my comments)

CodingCat:
one of the charming parts of open source is that everyone can fork a new version, I have listed all reasons I think the idea of the PR is not in the right path (repeatedly and answer different authors with the same idea)

I do not think we could move forward here, feel free to customize xgb for your customer if you believe this is the right path, but for open source version, we need to at least keep XGB behavior clean , reliable and backward compatible which are all violated by the idea here.

PR #5778, opened 2020-06-11
[R] replace uses of T and F with TRUE and FALSE

jameslamb:
This pull request proposes replacing uses of  `T` and `F` in R code with `TRUE`  and  `FALSE`. `T` and  `F` are dangerous because they are only aliases and can be overwritten.

```r
TRUE <- 4
# Error in TRUE <- 4 : invalid (do_set) left-hand side to assignment

T <- 4
T
# [1] 4
```

### DIFF
--- a/R-package/demo/basic_walkthrough.R
+++ b/R-package/demo/basic_walkthrough.R
@@ -100,7 +100,7 @@ print(paste("test-error=", err))
 
 # You can dump the tree you learned using xgb.dump into a text file
 dump_path = file.path(tempdir(), 'dump.raw.txt')
-xgb.dump(bst, dump_path, with_stats = T)
+xgb.dump(bst, dump_path, with_stats = TRUE)
 
 # Finally, you can check which features are the most important.
 print("Most important features (look at column Gain):")
--- a/R-package/demo/caret_wrapper.R
+++ b/R-package/demo/caret_wrapper.R
@@ -9,7 +9,7 @@ require(e1071)
 # Load Arthritis dataset in memory.
 data(Arthritis)
 # Create a copy of the dataset with data.table package (data.table is 100% compliant with R dataframe but its syntax is a lot more consistent and its performance are really good).
-df <- data.table(Arthritis, keep.rownames = F)
+df <- data.table(Arthritis, keep.rownames = FALSE)
 
 # Let's add some new categorical features to see if it helps. Of course these feature are highly correlated to the Age feature. Usually it's not a good thing in ML, but Tree algorithms (including boosted trees) are able to select the best features, even in case of highly correlated features.
 # For the first feature we create groups of age by rounding the real age. Note that we transform it to factor (categorical data) so the algorithm treat them as independant values.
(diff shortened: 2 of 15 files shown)


hcho3:
Is `T` a synonym (alias) of `TRUE` in R?

jameslamb:
> Is `T` a synonym (alias) of `TRUE` in R?

It is! But the difference (like in that code sample) is that `T` can be overridden and  `TRUE` cannot. So all this code changed in this this PR looks like it is working, but it could behave in an unexpected way in the future if someone adds a patch that creates any variable named `T`.

jameslamb:
We use `{lintr}` in `LightGBM` to catch such issues (and many others): https://github.com/microsoft/LightGBM/blob/d42f43132df63ae64783d3bc4f6366dd325116af/.ci/lint_r_code.R#L53

I'd be happy to propose a pull request that adds similar linting in `xgboost` CI if it interests you.

Alternatively, you or other maintainers could run it manually to look for things, like this from the root of the repo:

```shell
Rscript lint_r_code.R $(pwd)
```

hcho3:
@jameslamb Thanks for your explanation. Here is the current list of R linters we use: https://github.com/dmlc/xgboost/blob/master/R-package/tests/testthat/test_lint.R. You should add `T_and_F_symbol_linter` in that list.

hcho3:
Hmm looks like R lint hasn't been enabled:
https://github.com/dmlc/xgboost/blob/1d22a9be1cdeb53dfa9322c92541bc50e82f3c43/R-package/tests/testthat/test_lint.R#L26

Let's try enabling it and see what happens.

jameslamb:
> Hmm looks like R lint hasn't been enabled:
> https://github.com/dmlc/xgboost/blob/1d22a9be1cdeb53dfa9322c92541bc50e82f3c43/R-package/tests/testthat/test_lint.R#L26
> 
> Let's try enabling it and see what happens.

Sure, I just pushed https://github.com/dmlc/xgboost/pull/5778/commits/600185a0b071da53ff9a03c9659d02a80f5e2000 which uncomments it and adds the `T_and_F_symbol_linter`. Could you point me to which CI build it should be running in?


It looks like that `test_lint.R` was added in 2015. A lot of new linters have been added since then, which might be worth considering. Some of those are not just style things but can find actual code-correctness issues such as the one in this PR. Maybe we should move this discussion off this narrowly-scoped PR and into an issue?

terrytangyuan:
@jameslamb Thanks James! We should definitely revamp the code style and hopefully uncomment most of those linters.

hcho3:
It seems like all lint checks are passing: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-5778/2/pipeline/221#step-246-log-364.

I agree that we should create a new issue to add more lint checks for R in the future.

jameslamb:
> It seems like all lint checks are passing: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-5778/2/pipeline/221#step-246-log-364, where all testthat tests are run.
> 
> I agree that we should create a new issue to add more lint checks for R in the future.

@hcho3 that log isn't evidence that they're passing, since the linting code has `skip`s and I'm not sure if any of them are being triggered.

https://github.com/dmlc/xgboost/blob/1d22a9be1cdeb53dfa9322c92541bc50e82f3c43/R-package/tests/testthat/test_lint.R#L3-L5

jameslamb:
> I agree that we should create a new issue to add more lint checks for R in the future.

opened #5779

hcho3:
I removed the skip statements except `skip_on_cran()`. Let's see if lint check passes.

PR #5782, opened 2020-06-12
Avoid including `c_api.h` in header files.

trivialfis:


hcho3:
The name of this pull request is a bit confusing, since we are moving `#include <xgboost/c_api.h>` from headers to sources, not removing them. I'm assuming this is to speed up builds?

hcho3:
In the future, we may want to use pre-compiled headers to speed up builds. [CMake supports it.](https://cmake.org/cmake/help/v3.16/command/target_precompile_headers.html)

trivialfis:
> The name of this pull request is a bit confusing

@hcho3 Yup, let me change it.

> In the future, we may want to use pre-compiled headers to speed up builds

That's mostly for external libraries right?  If you cache your own header it will be slower.  Also nvcc doesn't seem to support it.

hcho3:
> Also nvcc doesn't seem to support it.

Darn, I was about to suggest PCH for Cub and Thrust :(

PR #5783, opened 2020-06-12
Accept iterator in device dmatrix.

trivialfis:
Stacked on #5782 and #5781 .  Replacing #5629 .

Related https://github.com/dmlc/xgboost/issues/5143 .

Related #5583 .

* Accept iterator for device dmatrix.
A new iterative device dmatrix is implemented, replacing the old device dmatrix.
* No longer have test for column major, as we now treat them uniformly as dense data.
* Add a custom data iterator for handling single batch of data.  
* Now the device quantile dmatrix can do weighted sketching.


trivialfis:
> can we deprecate old APIs and move to new unified APIs

No. That will require a CPU implementation.  It's possible with the basic structure defined in this PR, but I won't carry it out, at least not in near future.


> What is the relationship between adapters and this new iterator API? Are adapters still needed and in what circumstances?

Iterator returns batches of data, adapter adapts each batch.

src/data/array_interface.h
```diff
@@ -230,6 +230,7 @@ class ArrayInterfaceHandler {
 
 // A view over __array_interface__
 class ArrayInterface {
+ private:
```
RAMitchell, line 233: This looks redundant.

src/common/quantile.h
```diff
@@ -475,30 +475,6 @@ class QuantileSketchTemplate {
         this->data = dmlc::BeginPtr(space);
       }
     }
-    /*!
```
RAMitchell, line 478: Not sure if this change is relevant.

trivialfis: Reverted

src/data/array_interface.h
trivialfis: Reverted

PR #5795, opened 2020-06-15
Split Features into Groups to Compute Histograms in Shared Memory

canonizer:
- split features into groups to compute histograms in shared memory
- currently for dense matrices only


src/tree/updater_gpu_hist.cu
```diff
@@ -238,6 +243,43 @@ struct GPUHistMakerDevice {
     }
   }
 
+  void InitFeatureGroups() {
```
RAMitchell, line 246: I'm trying to think of ways to get rid of initialisation and make the kernel self-contained. Is it possible for a thread block to do a binary search over bins/features inside the kernel to independently allocate work to itself? Bin/feature information is available inside EllPackDeviceAccessor.

src/tree/gpu_hist/histogram.cu
```diff
@@ -105,20 +105,27 @@ __global__ void SharedMemHistKernel(EllpackDeviceAccessor matrix,
                                     common::Span<const RowPartitioner::RowIndexT> d_ridx,
                                     GradientSumT* __restrict__ d_node_hist,
                                     const GradientPair* __restrict__ d_gpair,
-                                    size_t n_elements,
+                                    const int* feature_groups, const int* bin_groups,
                                     GradientSumT const rounding,
                                     bool use_shared_memory_histograms) {
   using T = typename GradientSumT::ValueT;
   extern __shared__ char smem[];
+  int group = blockIdx.y;
```
RAMitchell, line 113: I would like kernel code to be much more readable. Some suggestions:
- add comments defining what is a bin_group/feature_group
- add comments talking about your parallel decomposition of work e.g. how each thread block is given a submatrix from the original matrix, and that submatrix may have the width of the original matrix if the data is sparse or has few features

You could create a struct called BlockWorkDecomposition that contains start/end rows and features and move some of this code to its constructor.

Imagine you are trying to convey what is happening here to a new xgboost developer.

src/tree/gpu_hist/histogram.cu
```diff
@@ -169,33 +181,41 @@ void BuildGradientHistogram(EllpackDeviceAccessor const& matrix,
...
+  int num_groups = feature_groups.size() - 1;
   int n_mps = 0;
   dh::safe_cuda(cudaDeviceGetAttribute(&n_mps, cudaDevAttrMultiProcessorCount, device));
   int n_blocks_per_mp = 0;
   dh::safe_cuda(cudaOccupancyMaxActiveBlocksPerMultiprocessor
                 (&n_blocks_per_mp, kernel, block_threads, smem_size));
+  // TODO(canonizer): This is really a hack, find a better way to distribute the data
+  // among thread blocks
   unsigned grid_size = n_blocks_per_mp * n_mps;
+  grid_size = common::DivRoundUp(grid_size, common::DivRoundUp(num_groups, 4));
```
RAMitchell, line 193: I don't understand num_groups/4.

Needs comments here explaining the work decomposition.

src/tree/updater_gpu_hist.cu
canonizer: I've moved feature groups into separate data types, this also made writing tests easier.

src/tree/gpu_hist/histogram.cu
canonizer: Now that there is a separate data type for feature groups, I think the code is more readable.

src/tree/gpu_hist/histogram.cu
canonizer: Done.

src/tree/gpu_hist/feature_groups.cu
```diff
@@ -0,0 +1,71 @@
...
+#include "feature_groups.cuh"
+
+#include "../../common/device_helpers.cuh"
+#include "../../common/hist_util.h"
+
+namespace xgboost {
+namespace tree {
+
+template <typename GradientSumT>
+void FeatureGroups::Init(const common::HistogramCuts& cuts, bool is_dense,
```
RAMitchell, line 18: Using a constructor and storing this object inside a unique pointer is preferable to using init functions.

My reasoning is that if the user forgets/doesn't realise the object is uninitialised you get a deferred crash - something undefined happens later in the program.

If we use a unique pointer and it is not initialised you get a crash the first time you touch it and it is more obvious what the problem is.

Issue #5793, opened 2020-06-14
In place prediction with cupy on windows crashes


RAMitchell:
This is a note that the following test crashes on windows currently:
`pytest tests\python-gpu\test_gpu_prediction.py::TestGPUPredict::test_inplace_predict_cupy`

This needs to be fixed to run cupy tests on windows, otherwise we do not have coverage for device memory interfaces on windows.

trivialfis:
Do you have a log?  Is it only reproducible on CI?

trivialfis:
This might be caused by using CUDA memory in TLS on Windows. Gott an `cudaInitializationError` in the destructor of `HostDeviceVector` when I spawn a new thread for testing in https://github.com/dmlc/xgboost/pull/9362

PR #5797, opened 2020-06-16
Add cupy to Windows CI

RAMitchell:
Multithreaded in-place prediction with cupy does not work on Windows (#5793). I am unable to find a workaround so have just disabled one test.

### DIFF
--- a/Jenkinsfile-win64
+++ b/Jenkinsfile-win64
@@ -137,6 +137,7 @@ def TestWin64GPU(args) {
     conda activate && for /R %%i in (python-package\\dist\\*.whl) DO python -m pip install "%%i"
     """
     echo "Installing Python dependencies..."
+    def cuda_short_ver = args.cuda_target.replaceAll('_', '')
     bat """
      conda activate && conda install -y hypothesis && conda upgrade scikit-learn pandas numpy hypothesis  && python -m pip install cupy-${cuda_short_ver}
     """
--- a/Jenkinsfile-win64
+++ b/Jenkinsfile-win64
@@ -138,7 +138,7 @@ def TestWin64GPU(args) {
     """
     echo "Installing Python dependencies..."
     bat """
-     conda activate && conda install -y hypothesis cupy && conda upgrade scikit-learn pandas numpy hypothesis cupy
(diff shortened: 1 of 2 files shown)


Jenkinsfile-win64
```diff
@@ -138,7 +138,7 @@ def TestWin64GPU(args) {
     """
     echo "Installing Python dependencies..."
```
hcho3, line 139:
```suggestion
    echo "Installing Python dependencies..."
    def cuda_short_ver = args.cuda_target.replaceAll('_', '')
```

Jenkinsfile-win64
```diff
@@ -138,7 +138,7 @@ def TestWin64GPU(args) {
...
     bat """
-     conda activate && conda install -y hypothesis && conda upgrade scikit-learn pandas numpy hypothesis
+     conda activate && conda install -y hypothesis cupy && conda upgrade scikit-learn pandas numpy hypothesis cupy
```
hcho3, line 141: cuPy version on Conda is quite old. I suggest install cuPy from pip
```suggestion
     conda activate && conda install -y hypothesis && conda upgrade scikit-learn pandas numpy hypothesis  && python -m pip install cupy-${cuda_short_ver}
```

tests/python-gpu/test_gpu_prediction.py
```diff
@@ -147,8 +147,10 @@ def predict_dense(x):
             copied_predt = cp.array(booster.predict(d))
             return cp.all(copied_predt == inplace_predt)
 
-        for i in range(10):
-            run_threaded_predict(X, rows, predict_dense)
+        # Don't do this on Windows, see issue #5793
+        if not sys.platform.startswith("win"):
+            for i in range(10):
+                run_threaded_predict(X, rows, predict_dense)
```
hcho3, line 153: Let's mark the test as skipped explicitly, so that we can remember to fix it:
```suggestion
        # Don't do this on Windows, see issue #5793
        if sys.platform.startswith("win"):
            pytest.skip('Multi-threaded in-place prediction with cuPy is not working on Windows')
        for i in range(10):
            run_threaded_predict(X, rows, predict_dense)
```

PR #5799, opened 2020-06-16
Accept string for ArrayInterface constructor.

trivialfis:
Extracted from https://github.com/dmlc/xgboost/pull/5783 .

Restore some checks from original implementation for `ArrayInterfaceHandler`.

### DIFF
--- a/src/data/array_interface.h
+++ b/src/data/array_interface.h
@@ -1,7 +1,7 @@
 /*!
  * Copyright 2019 by Contributors
  * \file array_interface.h
- * \brief Basic structure holding a reference to arrow columnar data format.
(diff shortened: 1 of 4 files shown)


src/data/array_interface.h
```diff
@@ -231,8 +232,8 @@ class ArrayInterfaceHandler {
 class ArrayInterface {
  public:
   ArrayInterface() = default;
-  explicit ArrayInterface(std::map<std::string, Json> const &column,
-                          bool allow_mask = true) {
+  void Initialize(std::map<std::string, Json> const &column,
```
RAMitchell, line 235: Should this be private?

PR #5800, opened 2020-06-16
Implement extend method for meta info.

trivialfis:
* Implement extend for host device vector.

Extracted from https://github.com/dmlc/xgboost/pull/5783 .

### DIFF
--- a/src/common/host_device_vector.cu
+++ b/src/common/host_device_vector.cu
@@ -125,6 +125,25 @@ class HostDeviceVectorImpl {
     }
   }
 
+  void Extend(HostDeviceVectorImpl* other) {
+    auto ori_size = this->Size();
+    this->Resize(ori_size + other->Size(), T());
+    if (HostCanWrite() && other->HostCanRead()) {
+      auto& h_vec = this->HostVector();
+      auto& other_vec = other->HostVector();
+      CHECK_EQ(h_vec.size(), ori_size + other->Size());
+      std::copy(other_vec.cbegin(), other_vec.cend(), h_vec.begin() + ori_size);
+    } else {
+      auto ptr = other->ConstDevicePointer();
+      SetDevice();
+      CHECK_EQ(this->DeviceIdx(), other->DeviceIdx());
+      dh::safe_cuda(cudaMemcpyAsync(this->DevicePointer() + ori_size,
+                                    ptr,
+                                    other->Size() * sizeof(T),
+                                    cudaMemcpyDeviceToDevice));
+    }
+  }
+
   std::vector<T>& HostVector() {
     LazySyncHost(GPUAccess::kNone);
     return data_h_;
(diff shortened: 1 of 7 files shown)


src/common/host_device_vector.cu
```diff
@@ -125,6 +125,25 @@ class HostDeviceVectorImpl {
...
+    if (HostCanWrite() && other->HostCanRead()) {
+      auto& h_vec = this->HostVector();
+      auto& other_vec = other->HostVector();
+      CHECK_EQ(h_vec.size(), ori_size + other->Size());
+      std::copy(other_vec.cbegin(), other_vec.cend(), h_vec.begin() + ori_size);
+    } else {
+      auto ptr = other->ConstDevicePointer();
+      SetDevice();
+      CHECK_EQ(this->DeviceIdx(), other->DeviceIdx());
+      dh::safe_cuda(cudaMemcpyAsync(this->DevicePointer() + ori_size,
```
RAMitchell, line 140: Just a note that this makes this assumption of plain old data. This is true in other places in HostDeviceVector and might cause us problems if someone tried to use it with more complicated types.

hcho3: Shall we add a static assertion to HostDeviceVector? Something like
```cpp
template <typename T>
class HostDeviceVector {
   ...
   static_assert(std::is_pod<T>::value, "HostDeviceVector only admits POD types");
};
```

RAMitchell: I think for `std::is_pod<T>` to be true the class cannot have a user-defined constructor. This is inconvenient, for example our gradient pair would have to be all constructed using gpair{a,b} syntax.

hcho3: We can use `std::is_standard_layout` instead, which allows for user-defined constructors.

RAMitchell: I haven't seen that before, let's try it. Seems like it would prevent use of abstract classes at least. Would this prevent classes with pointers?

hcho3: @RAMitchell This document has good explanation of the standard layout: https://docs.microsoft.com/en-us/cpp/cpp/trivial-standard-layout-and-pod-types?view=vs-2019#standard-layout-types

trivialfis: I'm just gonna replace it with thrust function instead. :-)

RAMitchell: I would recommend not using thrust for copying as it will allocate memory and manually construct unless data is POD.

RAMitchell: std::is_standard_layout and cudaMemcpy seems like a performant compromise.

hcho3: Just pushed a small patch to add the static assertion. Please review.

PR #5804, opened 2020-06-17
Add an option to run brute-force test for JSON round-trip

chyunsu3:
Follow-up to #5772

Add an option to run brute-force test for JSON round-trip. It will test every single 32-bit floating-point values. The CI won't run the brute-force test; the expectation is that we'd run it manually from time to time.

To run:
```
XGBOOST_JSON_ROUNDTRIP_EXHAUSTIVE_TEST=1 ./testxgboost --gtest_filter=Json.RoundTrip
```

The use of policy-class pattern is a bit ugly; I wanted to avoid branching in the middle of a tight loop.

EDIT. Now the command is
```
./testxgboost --gtest_filter=Json.DISABLED_RoundTripExhaustive \
  --gtest_also_run_disabled_tests
```

### DIFF
--- a/tests/cpp/common/test_json.cc
+++ b/tests/cpp/common/test_json.cc
@@ -9,6 +9,7 @@
 #include "xgboost/json.h"
 #include "xgboost/logging.h"
 #include "xgboost/json_io.h"
+#include "xgboost/parameter.h"
 #include "../helpers.h"
 #include "../../../src/common/io.h"
 #include "../../../src/common/charconv.h"
@@ -525,7 +526,8 @@ TEST(Json, IntVSFloat) {
   }
 }
 
-TEST(Json, RoundTrip) {
+template <typename JsonRoundTripTestPolicy>
+inline void JsonRoundTripTestImpl() {
   uint32_t i = 0;
   SimpleLCG rng;
   SimpleRealUniformDistribution<float> dist(1.0f, 4096.0f);
@@ -545,10 +547,33 @@ TEST(Json, RoundTrip) {
     }
 
     auto t = i;
-    i+= static_cast<uint32_t>(dist(&rng));
+    i += JsonRoundTripTestPolicy::Next(dist, rng);
     if (i < t) {
       break;
     }
+  }
+}
+
+struct ExhaustivePolicy {
+  template <typename T, typename U>
+  inline uint32_t static Next(T& dist, U& rng) {
+    return 1;
+  }
+};
+
+struct SamplingPolicy {
+  template <typename T, typename U>
+  inline uint32_t static Next(T& dist, U& rng) {
+    return static_cast<uint32_t>(dist(&rng));
+  }
+};
+
+TEST(Json, RoundTrip) {
+  int exhaustive_flag = dmlc::GetEnv("XGBOOST_JSON_ROUNDTRIP_EXHAUSTIVE_TEST", 0);
+  if (exhaustive_flag == 1) {
+    JsonRoundTripTestImpl<ExhaustivePolicy>();
+  } else {
+    JsonRoundTripTestImpl<SamplingPolicy>();
   }
 }
 }  // namespace xgboost
--- a/tests/cpp/common/test_json.cc
+++ b/tests/cpp/common/test_json.cc
@@ -553,8 +553,7 @@ TEST(Json, RoundTrip) {
 }
 
 TEST(Json, DISABLED_RoundTripExhaustive) {
-#pragma omp parallel for schedule(static)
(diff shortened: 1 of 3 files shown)


trivialfis:
Quoting from http://cheezyworld.com/wp-content/uploads/2010/12/PlainGoogleQuickTestReferenceGuide1.pdf

```
## Temporarily Disabling Tests

If you have a broken test that you cannot fix right away, you can add the `DISABLED_` prefix
to its name.  This will exclude it from execution.*Note: This feature should only be used
for temporary pain-relief.  You still have to fix the disabled tests at a later time.  As
a reminder, Google Test will print a banner warning you if a test program contains
disabled tests.

## Temporarily Enabling Disabled Tests

To include disabled tests in test execution, just invoke the test program with
`--gtest_also_run_disabled_tests` or set the `GTEST_ALSO_RUN_DISABLED_TESTS` environment
variable to a value greater than 0
```

hcho3:
I removed the policy class because it's too complicated and doesn't save that many lines. It's simpler to copy and paste the code.

tests/cpp/common/test_json.cc
```diff
@@ -545,7 +545,34 @@ TEST(Json, RoundTrip) {
...
+    if (i < t) {
+      break;
+    }
+  }
+}
+
+TEST(Json, DISABLED_RoundTripExhaustive) {
+  uint32_t i = 0;
+  SimpleLCG rng;
+  SimpleRealUniformDistribution<float> dist(1.0f, 4096.0f);
```
trivialfis, line 558: No longer needed.

tests/cpp/common/test_json.cc
```diff
@@ -545,7 +545,34 @@ TEST(Json, RoundTrip) {
...
+      ASSERT_TRUE(std::isnan(get<Number const>(loaded)));
+    } else {
+      ASSERT_EQ(get<Number const>(loaded), f);
+    }
+
+    auto t = i;
+    ++i;
     if (i < t) {
       break;
     }
```
trivialfis, line 578: No longer needed.  You can't overflow the integer when adding 1.

hcho3: Actually we still need it, since we run the loop until `i` becomes equal to `std::numeric_limits<uint32_t>::max()`. Then running `++i` would set `i` to zero.

trivialfis (approved):
Thanks for helping on the test!  Fun thing that I actually put a progress bar in there when I first created the test. ;-)

hcho3:
I'm going to run it on my machine and see how long it takes :)

trivialfis:
@hcho3 Do it in release build. ;-)

tests/cpp/common/test_json.cc
trivialfis: Sure, you can change `<=` to `<`.  Dropping a branch.

hcho3:
It took 816 seconds on my machine. Not bad.

trivialfis:
@hcho3 It will be faster when intrusive ptr is merged in dmlc.

trivialfis:
Use omp to parallel the test.  On my machine it's:
```
[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (115711 ms total)
[  PASSED  ] 1 test.

real	1m55.714s
```

trivialfis:
@hcho3 Let me know if it make sense to run it on CI.

tests/cpp/common/test_json.cc
```diff
@@ -567,6 +566,11 @@ TEST(Json, DISABLED_RoundTripExhaustive) {
     } else {
       EXPECT_EQ(get<Number const>(loaded), f);
     }
+  };
+#pragma omp parallel for schedule(static)
+  for (uint32_t i = 0; i < std::numeric_limits<uint32_t>::max(); ++i) {
```
hcho3, line 571: This is going to exclude the max value `i < std::numeric_limits<uint32_t>::max()`. Let's make the loop variable `i` 64-bit integer.

hcho3: Oops never mind, didn't see the last line.

hcho3:
Visual Studio doesn't allow an unsigned loop variable in an OpenMP loop, so I changed `i` to signed 64-bit integer.

hcho3:
Let me enable it only Jenkins Linux worker, since it's got the most number of cores.

hcho3:
@trivialfis I don't think we should run the exhaustive test in CI. It takes 176 seconds on Linux and 252 seconds on Windows. It may be because the CI machine runs other tasks concurrently.

PR #5813, opened 2020-06-19
Swap byte-order in binary serializer to support big-endian arch

qzhang90:
This pull request is to solve an endian problem of xgboost models.

There is the following definition in `dmlc-core/include/dmlc/base.h`
```
/*!
 * \brief Use little endian for binary serialization
 *  if this is set to 0, use big endian.
 */
#ifndef DMLC_IO_USE_LITTLE_ENDIAN
#define DMLC_IO_USE_LITTLE_ENDIAN 1
#endif
```

In my use case, the model needs to be trained on a little-endian machine and used on a big-endian machine. Therefore, on the little-endian machine, I set  `#define DMLC_IO_USE_LITTLE_ENDIAN 0` for training. However, the model CANNOT be used on a big-endian machine.

With this pull request, as long as the user specifies `#define DMLC_IO_USE_LITTLE_ENDIAN 0`, no matter on which machine the model is trained (little-endian or big-endian), the model is able to be used on a big-endian machine, and vice versa.


trivialfis:
I don't mind merging it since I think we will keep the binary model for a while.  I need to help some other major down stream projects like treelite and shap to transition into JSON later.  After than we can consider putting a warning on binary model.

Just not sure how to test this compatibility.

hcho3:
@trivialfis I took a short nap and found a way to simplify this pull request. Originally it seemed like too much addition for what we gain, but we can shift the balance by reducing the amount of code being added. I'll push some changes shortly, and we can consider merging it.

hcho3:
Travis CI recently added support for s390x (IBM Z), so we can add some testing for big-endian platforms.

src/learner.cc
```diff
@@ -872,6 +917,11 @@ class LearnerIO : public LearnerConfiguration {
 )doc";
       int64_t sz {-1};
       CHECK_EQ(fp.Read(&sz, sizeof(sz)), sizeof(sz));
+
+      if (DMLC_IO_NO_ENDIAN_SWAP == 0){
+        sz = __builtin_bswap64(sz);
```
hcho3, line 922: We should avoid using compiler intrinsics like this, since we'd like to support Visual Studio as well. I'm looking into using [`dmlc::ByteSwap()`](https://github.com/dmlc/dmlc-core/blob/master/include/dmlc/endian.h#L51)

src/gbm/gbtree_model.cc
```diff
@@ -10,6 +10,12 @@ namespace gbm {
 
 void GBTreeModel::Save(dmlc::Stream* fo) const {
   CHECK_EQ(param.num_trees, static_cast<int32_t>(trees.size()));
+
+  if (DMLC_IO_NO_ENDIAN_SWAP == 0) {
+    GBTreeModelParam* param_ptr = const_cast<GBTreeModelParam*>(&param);
```
hcho3, line 15: We should not mutate the model state when serializing the model.

hcho3: See my commit.

src/learner.cc
hcho3: See my commit.

qzhang90:
> I don't mind merging it since I think we will keep the binary model for a while. I need to help some other major down stream projects like treelite and shap to transition into JSON later. After than we can consider putting a warning on binary model.
> 
> Just not sure how to test this compatibility.

trivialfis:
@hcho3 Can we split up the PR into supporting power and cross platform model loading?

hcho3:
@trivialfis This PR doesn’t actually add tests for cross-platform loading. I’m stuck because I don’t know why C++ tests are failing.

trivialfis:
Let's hold the PR.  It's very difficult to debug in the future since non of us have physical access to ppc.

I think we can first add ppc to test matrix, skip whatever tests that are failing, document the limitations according to failing tests.

Then we or anyone who's interested in can fix them later.

qzhang90:
Hi @hcho3 @trivialfis , thanks for making so much progress on this PR. 

> hcho3 commented on Jun 21
> @trivialfis This PR doesn’t actually add tests for cross-platform loading. I’m stuck because I don’t know why C++ tests are failing.
> 
Do I need to provide some tests for cross-platform loading? Anything that I can help to figure out why the C++ tests are failing?

> trivialfis commented on Jun 21 • 
> Let's hold the PR. It's very difficult to debug in the future since none of us have physical access to ppc.
> 
> I think we can first add ppc to test matrix, skip whatever tests that are failing, document the limitations according to failing tests.
> 
> Then we or anyone who's interested in can fix them later.
Can you advise why do we concern about not having physical access to ppc? Is it because Travis does not support ppc but we want to make sure the code will not break on ppc?
Thanks!

hcho3:
> Can you advise why do we concern about not having physical access to ppc? Is it because Travis does not support ppc but we want to make sure the code will not break on ppc?

He probably meant s390x. None of us developers have access to s390x (IBM Z).

hcho3:
@qzhang90 In my opinion, the XGBoost project currently lacks sufficient resource to support big-endian architectures. We neither have development machines nor regular contributors who understand big-endian archs. If Travis CI reports a failure in IBM Z, none of us will be able to fix it. Thus, as it is, I am inclined to reject this pull request.

We may reconsider if you or someone else commits to maintain big-endian support in the long term.

qzhang90:
> > Can you advise why do we concern about not having physical access to ppc? Is it because Travis does not support ppc but we want to make sure the code will not break on ppc?
> 
> He probably meant s390x. None of us developers have access to s390x (IBM Z). Travis CI supports it, but not having a local access hampers development and debugging.

@hcho3 I see. This one could be helpful https://github.com/multiarch/qemu-user-static. It allows running containers with different architectures, which can provide local access to the big-endian environment.

hcho3:
> I see. This one could be helpful https://github.com/multiarch/qemu-user-static. It allows running containers with different architectures, which can provide local access to the big-endian environment.

Interesting. Let me take a look and see if it lets us locally test code with reasonable amount of effort.

qzhang90:
> > I see. This one could be helpful https://github.com/multiarch/qemu-user-static. It allows running containers with different architectures, which can provide local access to the big-endian environment.
> 
> Interesting. Let me take a look and see if it lets us locally test code with reasonable amount of effort.

Sure, thanks!

hcho3:
@qzhang90 The s390x Ubuntu image works quite nicely. I am able to everything that I can do with a typical Ubuntu installation. A major disadvantage I find is that emulation is quite slow. It takes 11.5 minutes to build XGBoost inside the s390x container, which is 11 times slower than what it usually takes to build XGBoost (1 minute on the same machine).

qzhang90:
> @qzhang90 The s390x Ubuntu image works quite nicely. I am able to everything that I can do with a typical Ubuntu installation. A major disadvantage I find is that emulation is quite slow. It takes 11.5 minutes to build XGBoost inside the s390x container, which is 11 times slower than what it usually takes to build XGBoost (1 minute on the same machine).

@hcho3 I see, emulation can be slow, indeed. Did you try to give more CPUs to the container and build it in parallel by `make -j$(nproc)`?

hcho3:
@qzhang90 Yes, I used Ninja, which uses all CPU cores by default.

qzhang90:
@hcho3 Maybe the first build is slow but the subsequent ones will be faster?

hcho3:
I think the PR is now good. It runs the model compatibility test from `tests/python` to ensure that XGBoost on s390x can read model files generated on amd64.

@trivialfis I'd like to hear your opinion as to whether we should accept this PR. We should be realistic in that none of us developers use big-endian machines and adjust how much support we provide for that.

I don't think I have time to get all the Python and R test work with s390x. So here I'm only running model compatibility tests.

qzhang90:
> I will take a look into the details later. For supporting big endian I think it should be fine.

Hi @trivialfis , just wondering have you got a chance to look into this, and what will be the next steps? Thanks!

src/gbm/gbtree_model.cc
```diff
@@ -10,18 +10,35 @@ namespace gbm {
...
+  } else {
+    auto x = param.ByteSwap();
+    fo->Write(&x, sizeof(x));
+  }
   for (const auto & tree : trees) {
     tree->Save(fo);
   }
   if (tree_info.size() != 0) {
-    fo->Write(dmlc::BeginPtr(tree_info), sizeof(int32_t) * tree_info.size());
+    if (DMLC_IO_NO_ENDIAN_SWAP) {
```
trivialfis, line 24: Can we put the condition into `ByteSwap` function?  If true then `ByteSwap` is a no op.

qzhang90: @trivialfis sorry but can you explain more about `no op`? The intuition here is that `ByteSwap` only needs to be invoked when the condition `DMLC_IO_NO_ENDIAN_SWAP` does not meet, thus it seems to me not that reasonable to put the condition into `ByeSwap`.

qzhang90:
> My question is would it be better to move the predicate into `ByteSwap` function itself:
> 
> ```
> 
> Param ByteSwap() const {
>    if (DMLC_IO_NO_ENDIAN_SWAP) {
>       return *this;
>    }
>    Param ret;
>     .... Swapping each field.
>    return ret;
> }
> ```
> 
> This way it seems less intrusive. But I'm fine with the current approach as I want to set internal serialization to JSON as default in this release.

I agree it is less intrusive, but leaving the condition outside of `ByteSwap` makes the `ByteSwap` function cleaner, and also the code easier to understand, which is: `ByteSwap` should indeed swap and bytes.

src/gbm/gbtree_model.cc
qzhang90: Hi @trivialfis, would you mind me marking this as `resolved` if we agreed on leaving this condition outside of `ByteSwap`?

trivialfis:
No problem.  Will take another look today.  Thanks for the patience.

trivialfis:
@hcho3  I will leave it for you to merge it.  Hopefully we can transit to JSON sooner.

PR #5816, opened 2020-06-20
Remove GK quantile.

trivialfis:
* GK is not used.


hcho3:
> WXQ is not documented and doesn't provide better performance.

Do you have some benchmarks?

trivialfis:
@hcho3

Implementation is not well defined according to:  https://github.com/dmlc/xgboost/pull/5251#issuecomment-584456536

Speed is not improved compared to WQuantile according to: https://github.com/dmlc/xgboost/pull/5251#issuecomment-584970975

But I also ran some bench with approx for 10 boosting rounds with higgs and url, the accuracy drops a little I'm not sure if it's a good thing, as there could be some unexpected regularization due to over pruning and somehow improves the training result.

**Before PR**:
```
{
  "higgs": {
    "xgb-approx": {
      "accuracy": {
        "AUC": 0.7863048187857767,
        "Accuracy": 0.6683204545454545,
        "Log_Loss": 0.5994604972915893,
        "Precision": 0.6274573581031722,
        "Recall": 0.9213437189704226
      },
      "train_time": 123.83011394099594
    }
  }
}

{
  "url": {
    "xgb-approx": {
      "accuracy": {
        "AUC": 0.9937412206162852,
        "Accuracy": 0.9793312549819918,
        "Log_Loss": 0.23446456971505325,
        "Precision": 0.9794169259854371,
        "Recall": 0.989928429820077
      },
      "train_time": 128.89666074500565
    }
  }
}
```
**After PR**:
```
{
  "higgs": {
    "xgb-approx": {
      "accuracy": {
        "AUC": 0.7863032095727018,
        "Accuracy": 0.6681763636363637,
        "Log_Loss": 0.5995105257037553,
        "Precision": 0.6273229415617546,
        "Recall": 0.9214491873787757
      },
      "train_time": 122.97428455900081
    }
  }
}

{
  "url": {
    "xgb-approx": {
      "accuracy": {
        "AUC": 0.9937861863103972,
        "Accuracy": 0.9793875958316118,
        "Log_Loss": 0.23464104552753764,
        "Precision": 0.9797411558496062,
        "Recall": 0.989672822034638
      },
      "train_time": 128.23554962199705
    }
  }
}
```

RAMitchell:
When I extended the quantile tests found WXQ does not pass some tests. The rank of the output quantiles was outside the expected range in many cases, additionally when encountering low cardinality data it will sometimes remove elements from the set even when the cardinality is less than the maximum sketch size - so the sketch is removing information from the data set. I don't understand its implementation so I couldn't directly fix these issues. WQ does not have these problems.

trivialfis:
Restored WXQ.

PR #5819, opened 2020-06-21
Update core.py documentation

Mo-Saif:


trivialfis:
Sorry for the long wait.  It's not necessarily 1 by default.  The default is whatever openmp provides.  If users want to set it in the global environment XGBoost will respect it.

PR #5825, opened 2020-06-24
Introducing DPC++-based plugin (predictor, objective function) supporting oneAPI programming model

vepifanov:
Part of the PR: https://github.com/dmlc/xgboost/pull/5659

Updater target is grow_quantile_histmaker.

Following process intended:

Start with existing implementation of grow_quantile_histmaker and continuously replace OpenMP parts into DPC++ kernels.
As soon as we get whole flow transformed into DPC++ kernels, start analyzing potential optimizations.
Current partial stoppers:

DPC++ compiler is required to build plugin sources and final linkage of the library, so additional compiler/linker flags should be added to building process (beyond the plugin itself).
Data passing between updater/prediction/getGradient calls is wrapped into HostDeviceVector which now supports only CPU/CUDA memory, so additional host-device transfers are required on each iteration.


vepifanov:
What is the intended way of enabling custom predictors?
Registering in a predictor factory is not enough, since it is dispatched by predefined predictor type enum and created manually during updater configuration.

plugin/updater_oneapi/README.md
```diff
@@ -0,0 +1,35 @@
...
+### Algorithms
+| predictor | Description |
+| --- | --- |
+predictor_oneapi | prediction using OneAPI device  |
+
+Python example:
+```python
+param['predictor'] = 'predictor_oneapi'
+param['objective'] = 'reg:squarederror_oneapi'
+
```
napetrov, line 23:
```suggestion
```
```

**Comment by hcho3** [comment]
> Registering in a predictor factory is not enough, since it is dispatched by predefined predictor type enum and created manually during updater configuration.

I think you can add a new entry to the enum and then modify the updater configuration to use the DPC++ predictor. You'll want to add a run-time check to see whether XGBoost was compiled with the DPC++ plugin.

File: plugin/updater_oneapi/regression_obj_oneapi.cc [hunk 2]
```diff
@@ -0,0 +1,180 @@
...
+  const char* DefaultEvalMetric() const override {
+    return Loss::DefaultEvalMetric();
+  }
+
+  void PredTransform(HostDeviceVector<float> *io_preds) override {
+    size_t const ndata = io_preds->Size();
+
+    cl::sycl::buffer<bst_float, 1> io_preds_buf(io_preds->HostPointer(), io_preds->Size());
+
+    qu_.submit([&](cl::sycl::handler& cgh) {
```
trivialfis, line 126: Is it possible to use a global queue and don't block on it?

plugin/updater_oneapi/regression_obj_oneapi.cc
```diff
@@ -0,0 +1,180 @@
...
+
+    bool is_null_weight = info.weights_.Size() == 0;
+    auto scale_pos_weight = param_.scale_pos_weight;
+    if (!is_null_weight) {
+      CHECK_EQ(info.weights_.Size(), ndata)
+          << "Number of weights should be equal to number of data points.";
+
+      cl::sycl::buffer<bst_float, 1> weights_buf(info.weights_.HostPointer(), info.weights_.Size());
+
+      qu_.submit([&](cl::sycl::handler& cgh) {
```
trivialfis, line 79: Can the lambda capture `is_null_weight` so we can merge to lambdas?

plugin/updater_oneapi/regression_obj_oneapi.cc
```diff
@@ -0,0 +1,180 @@
...
+};
+
+// register the objective functions
+DMLC_REGISTER_PARAMETER(RegLossParamOneAPI);
+
+XGBOOST_REGISTER_OBJECTIVE(SquaredLossRegressionOneAPI, LinearSquareLossOneAPI::Name())
+.describe("Regression with squared error with DPC++ backend.")
+.set_body([]() { return new RegLossObjOneAPI<LinearSquareLossOneAPI>(); });
+
+// TODO: Find a way to dispatch names of DPC++ kernels with various template parameters of loss function
```
trivialfis, line 162: If you want we can work it out together.  I don't mind some refactoring for internal code first.

plugin/updater_oneapi/regression_obj_oneapi.cc
vepifanov: Ideally yes, but there can be some issues with passing pointer to weights buffer into kernel in case when this buffer is empty. I will investigate it.

plugin/updater_oneapi/regression_obj_oneapi.cc
vepifanov: This will be one of our further steps, when we move to proper device dispatching (we are still in discussion on how to do it).
However, we need to find a good place to store the single queue for all parts of training flow and/or implement mechanisms to pass the queue between this parts (updater, gradient computation, obj. function, etc.).

plugin/updater_oneapi/regression_obj_oneapi.cc
vepifanov: Current approach looks fine.

This is more of a limitation of a DPC++ compiler - kernel name should be some globally-visible class name, so I can't define it inside like Loss structure. The only way I found up to this point is to define empty classes for each kernel outside Loss structure (somewhere nearby) and pass it as an additional template parameters.

In case of objective function it will be two additional template parameters (one in case of merged kernel).

In general case number of kernels can be much larger and it will also require explicit specialization of all potential combinations of templated classes/functions and it will go messy very fast.

I will probably communicate internally with DPC++ compiler team to find better solution.

plugin/updater_oneapi/regression_obj_oneapi.cc
vepifanov: Merged kernels into one.

plugin/updater_oneapi/regression_obj_oneapi.cc
vepifanov: Implemented multiple loss functions using template parameters for kernels name for now.

vepifanov:
Any additional comments on this PR?

Latest changes - had to move declaration of "objxgboost" target to base CMake file to properly set compile flags (-fsycl) only to necessary plugin files.
Turned out that the set_source_files_properties only works when called in the same file where target was declared.

trivialfis:
Will look into this tomorrow.

plugin/CMakeLists.txt
```diff
@@ -6,3 +6,16 @@ endif (PLUGIN_LZ4)
...
   target_sources(objxgboost PRIVATE ${xgboost_SOURCE_DIR}/plugin/dense_parser/dense_libsvm.cc)
 endif (PLUGIN_DENSE_PARSER)
+
+if (PLUGIN_UPDATER_ONEAPI)
+  
+  set(PLUGINS_SOURCES_ONEAPI 
+    ${xgboost_SOURCE_DIR}/plugin/updater_oneapi/regression_obj_oneapi.cc 
+    ${xgboost_SOURCE_DIR}/plugin/updater_oneapi/predictor_oneapi.cc)
+
+  set(PLUGINS_SOURCES_ONEAPI ${PLUGINS_SOURCES_ONEAPI} PARENT_SCOPE)
```
hcho3, line 16: @vepifanov  Let's make a separate object target `oneapi_plugin`, so that we can call `set_source_files_properties()` on this target. Then we don't have to move `objxgboost`. If you'd like, I can make the change.

vepifanov: That would be nice, thank you!
I thought about making another target, but I'm not sure how to properly transfer necessary flags from base target.

hcho3: @vepifanov I just pulled [the OneAPI Docker image](https://hub.docker.com/r/intel/oneapi-basekit) from Intel. I am getting this error when I invoked CMake:
```
CMake Error at /usr/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:165 (message):
  Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES)
Call Stack (most recent call first):
  /usr/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:458 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.18/Modules/FindOpenMP.cmake:529 (find_package_handle_standard_args)
  CMakeLists.txt:114 (find_package)


-- Configuring incomplete, errors occurred!
```

How did you get OpenMP working with Intel compilers?

hcho3: @vepifanov I found the fix:
```
cmake .. -DPLUGIN_UPDATER_ONEAPI=ON -DCMAKE_C_COMPILER=clang \
    -DCMAKE_CXX_COMPILER=dpcpp -DOpenMP_CXX_LIB_NAMES="libiomp5"  \
    -DOpenMP_C_LIB_NAMES="libiomp5" -DOpenMP_CXX_FLAGS="-openmp"  \
    -DOpenMP_C_FLAGS="-openmp"   \
    -DOpenMP_libiomp5_LIBRARY=/opt/intel/inteloneapi/compiler/latest/linux/compiler/lib/intel64_lin/libiomp5.so
```

hcho3: Good thing that Intel provides a Docker image. We might be able to add oneAPI plugin to CI in the near future.

hcho3: See my latest commit.

hcho3:
+1 to having tests. I'd like to look into adding oneAPI container to CI pipeline soon, as the Intel compiler helped us find a noncompliant bit of C++ (#5921).

trivialfis:
> I was going to adapt existing tests for obj

@vepifanov No problem,  anything that shows the code works.

vepifanov:
Added tests for plugin functionality - it works fine locally on machines with and without gpu devices.

However, I faced some issues during building of googletests using DPCPP compiler:
1. To use math functions DPCPP compiler requires libimf.so to be linked, and that can be fixed by adding -DCMAKE_CXX_STANDARD_LIBRARIES="/opt/intel/inteloneapi/compiler/latest/linux/compiler/lib/intel64_lin/libimf.so" to cmake command line.
2. There are some typo/bug/some auto-generated stuff in math.h header (that I believe comes even from Intel Compiler) which raise warnings about type conversions. This warnings turn into errors due to "-Werror -Wconversion" flags in googletest cmake environment. They can not be overriden by adding "-DCMAKE_CXX_FLAGS=-Wno-error=conversion" to cmake command line, because it inserts if before the googletest flag "-Werror" and it becomes shadowed. So I had to manually disable "-Wconversion" flag in googletest cmake scripts to get things work, which I believe is not the generally appropriate way to solve this issue.

trivialfis:
For the first issue, is there a way to write a cmake module to handle it?

hcho3:
> To use math functions DPCPP compiler requires libimf.so to be linked, and that can be fixed by adding -DCMAKE_CXX_STANDARD_LIBRARIES="/opt/intel/inteloneapi/compiler/latest/linux/compiler/lib/intel64_lin/libimf.so" to cmake command line.

That's odd. I was about to compile `testxgboost` without libimf.so. See https://github.com/dmlc/xgboost/pull/5921#issuecomment-662194822.

hcho3:
I need to see if we can use Docker in GitHub Actions.

vepifanov:
> > To use math functions DPCPP compiler requires libimf.so to be linked, and that can be fixed by adding -DCMAKE_CXX_STANDARD_LIBRARIES="/opt/intel/inteloneapi/compiler/latest/linux/compiler/lib/intel64_lin/libimf.so" to cmake command line.
> 
> That's odd. I was about to compile `testxgboost` without libimf.so. See [#5921 (comment)](https://github.com/dmlc/xgboost/pull/5921#issuecomment-662194822).

New version of oneAPI Toolkit became available just yesterday and looks like it shouldn't contain issues I described (with math.h and libimf.so). With this new version, I managed to build whole library + tests using the command line from PR you mentioned without any errors.

I believe that at this point, this PR is ready for merge.

plugin/updater_oneapi/README.md
```diff
@@ -0,0 +1,36 @@
+# DPC++-based Algorithm for Tree Construction
```
trivialfis, line 1: Please add a notice that the parameter name might change in the future, as we want to have a more integrated experience as time move.

vepifanov: Done.

PR #5831, opened 2020-06-27
Ensure that LoadSequentialFile() actually read the whole file

### LINKED ISSUE #5826: Error while loading model
After train a model for a long time I saved it using

`bst.save_model("xgboostModel.trail04.json")`

I latter tried to load it with the same params to continue training 
```
filemodel = 'xgboostModel.trail04.json'
bst = xgb.train(param, xgTrain, num_boost_round=numOfRounds, evals=watchList, early_stopping_rounds=2000, xgb_model=filemodel)
```
but I get this error:

```
  File "/home/user/programs/pycharm/plugins/python-ce/helpers/pydev/pydevd.py", line 1438, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/home/user/programs/pycharm/plugins/python-ce/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/home/user/projects/temp/pythontmp/lastPointClassifiers.py", line 270, in <module>
    trial04()
  File "/home/user/projects/temp/pythontmp/lastPointClassifiers.py", line 256, in trial04
    xgb_model=filemodel)
  File "/home/user/.local/lib/python3.7/site-packages/xgboost/training.py", line 212, in train
    xgb_model=xgb_model, callbacks=callbacks)
  File "/home/user/.local/lib/python3.7/site-packages/xgboost/training.py", line 37, in _train_internal
    model_file=xgb_model)
  File "/home/user/.local/lib/python3.7/site-packages/xgboost/core.py", line 1175, in __init__
    self.load_model(model_file)
  File "/home/user/.local/lib/python3.7/site-packages/xgboost/core.py", line 1804, in load_model
    self.handle, c_str(os_fspath(fname))))
  File "/home/user/.local/lib/python3.7/site-packages/xgboost/core.py", line 190, in _check_call
xgboost.core.XGBoostError: [19:57:42] /workspace/src/common/json.cc:413: Expecting: ",", got: "

```

When I try the same exact thing but run it over a smaller number of rounds it loads fine and trains like a champ.

```
def Example():
    X, Y = getData(randomize=True)
    X = np.array(X)
    Y = np.array(Y)
    l = int(len(X) * .5)
    xgTrain = xgb.DMatrix(X[:l], label=Y[:l])
    xgTest = xgb.DMatrix(X[l:], label=Y[l:])
    param = {
        'objective': 'multi:softmax',
        'eta': 0.0001,
        'max_depth': 6,
        'nthread': 16,
        'num_class': 2,
        'tree_method': 'gpu_hist',
    }
    watchList = [(xgTrain, 'train'), (xgTest, 'test')]
    numOfRounds = 1000000
    filemodel = 'xgboostModel.trail04.json'

    print("fit()")
    if path.exists(filemodel):
        bst = xgb.train(param, xgTrain, num_boost_round=numOfRounds, evals=watchList, early_stopping_rounds=2000,
                        xgb_model=filemodel)
    else:
        bst = xgb.train(param, xgTrain, num_boost_round=numOfRounds, evals=watchList, early_stopping_rounds=2000)

    bst.save_model(filemodel)
```

hcho3:
Fixes #5826.

On Linux, XGBoost uses the POSIX function `read()` to read the JSON model file. The `read()` function may return fewer bytes than the number of bytes requested, according to the [POSIX standard](https://pubs.opengroup.org/onlinepubs/009695399/functions/read.html):

> The value returned may be less than _nbyte_ if the number of bytes left in the file is less than _nbyte_

Fix: Call `read()` multiple times until we actually read the whole file.

### DIFF
--- a/include/xgboost/json_io.h
+++ b/include/xgboost/json_io.h
@@ -9,12 +9,13 @@
 #include <vector>
 #include <memory>
 #include <string>
-#include <cinttypes>
 #include <utility>
 #include <map>
 #include <limits>
 #include <sstream>
 #include <locale>
+#include <cinttypes>
+#include <cstdio>
 
 namespace xgboost {
 /*
@@ -84,8 +85,10 @@ class JsonReader {
     std::string msg = "Expecting: \"";
     msg += c;
     msg += "\", got: \"";
-    if (got == -1) {
+    if (got == EOF) {
       msg += "EOF\"";
+    } else if (got == 0) {
+      msg += "\\0\"";
     } else {
       msg += std::to_string(got) + " \"";
     }
--- a/include/xgboost/json_io.h
+++ b/include/xgboost/json_io.h
@@ -15,7 +15,6 @@
 #include <sstream>
 #include <locale>
 #include <cinttypes>
-#include <cstdio>
 
 namespace xgboost {
 /*
@@ -85,7 +84,7 @@ class JsonReader {
     std::string msg = "Expecting: \"";
     msg += c;
     msg += "\", got: \"";
-    if (got == EOF) {
+    if (got == -1) {
       msg += "EOF\"";
     } else if (got == 0) {
       msg += "\\0\"";
--- a/src/common/io.cc
+++ b/src/common/io.cc
@@ -7,9 +7,10 @@
 #include <unistd.h>
 #endif  // defined(__unix__)
 #include <algorithm>
-#include <cstdio>
+#include <fstream>
 #include <string>
 #include <utility>
+#include <cstdio>
 
 #include "xgboost/logging.h"
 #include "io.h"
@@ -130,21 +131,12 @@ std::string LoadSequentialFile(std::string fname) {
     bytes_read += result;
   }
   close(fd);
-#else  // defined(__unix__)
-  FILE *f = fopen(fname.c_str(), "r");
-  if (f == NULL) {
-    std::string msg;
-    OpenErr();
-  }
-  fseek(f, 0, SEEK_END);
-  auto fsize = ftell(f);
-  fseek(f, 0, SEEK_SET);
-
-  buffer.resize(fsize + 1);
-  fread(&buffer[0], 1, fsize, f);
-  fclose(f);
-#endif  // defined(__unix__)
   buffer.back() = '\0';
+#else  // defined(__unix__)
+  std::ifstream ifs(fname);
+  buffer = std::string( (std::istreambuf_iterator<char>(ifs) ),
+                        (std::istreambuf_iterator<char>()) );
+#endif  // defined(__unix__)
   return buffer;
 }
 
--- a/src/common/io.cc
+++ b/src/common/io.cc
@@ -123,7 +123,7 @@ std::string LoadSequentialFile(std::string fname) {
 #endif  // defined(__linux__)
   ssize_t bytes_read = 0;
   while (bytes_read < f_size_bytes) {
-    ssize_t result = read(fd, &buffer[bytes_read], f_size_bytes);
+    ssize_t result = read(fd, &buffer[bytes_read], f_size_bytes - bytes_read);
     if (result < 0) {
       close(fd);
       ReadErr();
--- a/src/common/io.cc
+++ b/src/common/io.cc
@@ -120,10 +120,14 @@ std::string LoadSequentialFile(std::string fname) {
 #if defined(__linux__)
   posix_fadvise(fd, 0, 0, POSIX_FADV_SEQUENTIAL);
 #endif  // defined(__linux__)
-  ssize_t bytes_read = read(fd, &buffer[0], f_size_bytes);
-  if (bytes_read < 0) {
-    close(fd);
-    ReadErr();
+  ssize_t bytes_read = 0;
+  while (bytes_read < f_size_bytes) {
+    ssize_t result = read(fd, &buffer[bytes_read], f_size_bytes);
+    if (result < 0) {
+      close(fd);
+      ReadErr();
+    }
+    bytes_read += result;
   }
   close(fd);
 #else  // defined(__unix__)
--- a/src/common/io.cc
+++ b/src/common/io.cc
@@ -132,14 +132,6 @@ std::string LoadSequentialFile(std::string fname) {
   }
   close(fd);
   buffer.back() = '\0';
-  // Trim NUL letters at the end
-  int64_t i;
-  for (i = static_cast<int64_t>(buffer.size()); i >= 0; --i) {
-    if (buffer[i - 1] != '\0') {
-      break;
-    }
-  }
-  buffer = buffer.substr(0, i);
 #else  // defined(__unix__)
   std::ifstream ifs(fname);
   buffer = std::string((std::istreambuf_iterator<char>(ifs)),
--- a/src/common/io.cc
+++ b/src/common/io.cc
@@ -109,34 +109,17 @@ std::string LoadSequentialFile(std::string fname) {
                  };
 
   std::string buffer;
-#if defined(__unix__)
-  struct stat fs;
-  if (stat(fname.c_str(), &fs) != 0) {
-    OpenErr();
-  }
-
-  size_t f_size_bytes = fs.st_size;
-  buffer.resize(f_size_bytes + 1);
-  int32_t fd = open(fname.c_str(), O_RDONLY);
-#if defined(__linux__)
-  posix_fadvise(fd, 0, 0, POSIX_FADV_SEQUENTIAL);
-#endif  // defined(__linux__)
-  ssize_t bytes_read = 0;
-  while (bytes_read < f_size_bytes) {
-    ssize_t result = read(fd, &buffer[bytes_read], f_size_bytes - bytes_read);
-    if (result < 0) {
-      close(fd);
-      ReadErr();
-    }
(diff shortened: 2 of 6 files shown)


src/common/io.cc
```diff
@@ -120,27 +121,22 @@ std::string LoadSequentialFile(std::string fname) {
...
+  buffer.back() = '\0';
 #else  // defined(__unix__)
-  FILE *f = fopen(fname.c_str(), "r");
-  if (f == NULL) {
-    std::string msg;
-    OpenErr();
-  }
-  fseek(f, 0, SEEK_END);
-  auto fsize = ftell(f);
-  fseek(f, 0, SEEK_SET);
```
hcho3, line 137: Note: According to the [MSDN documentation](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/ftell-ftelli64?view=vs-2019), `ftell()` may not yield the correct byte offset when the file was opened in the text mode.

Fix. We do away with `fseek()` and `ftell()` entirely and use `std::istreambuf_iterator` to read the whole file into `std::string`. The fix only applies on Windows. On Linux, we still use `read()`, along with `stat()` to measure the file size.

src/common/io.cc
```diff
@@ -120,27 +121,30 @@ std::string LoadSequentialFile(std::string fname) {
...
-
-  buffer.resize(fsize + 1);
-  fread(&buffer[0], 1, fsize, f);
-  fclose(f);
+  buffer = buffer.substr(0, i);
+#else  // defined(__unix__)
+  std::ifstream ifs(fname);
```
trivialfis, line 144: Please note that this is terribly slow.

hcho3: Slower than `std::fread()`?

hcho3: In https://social.msdn.microsoft.com/Forums/en-US/87e128e2-5320-4796-b4a2-36248f86a8ce/huge-performance-hit-when-using-ifstream-vs-fopen?forum=Vsexpressvc, `ifstream` shows competitive performance with respect to `std::fread()`.

include/xgboost/json_io.h
```diff
@@ -84,8 +85,10 @@ class JsonReader {
     std::string msg = "Expecting: \"";
     msg += c;
     msg += "\", got: \"";
-    if (got == -1) {
+    if (got == EOF) {
```
trivialfis, line 88: Please don't remove `-1` case.

src/common/io.cc
```diff
@@ -120,27 +121,30 @@ std::string LoadSequentialFile(std::string fname) {
...
+    bytes_read += result;
   }
   close(fd);
-#else  // defined(__unix__)
-  FILE *f = fopen(fname.c_str(), "r");
-  if (f == NULL) {
-    std::string msg;
-    OpenErr();
+  buffer.back() = '\0';
+  // Trim NUL letters at the end
```
trivialfis, line 135: Where is the NULL inserted?

src/common/io.cc
```diff
@@ -120,27 +121,30 @@ std::string LoadSequentialFile(std::string fname) {
...
+    }
   }
-  fseek(f, 0, SEEK_END);
-  auto fsize = ftell(f);
-  fseek(f, 0, SEEK_SET);
-
-  buffer.resize(fsize + 1);
-  fread(&buffer[0], 1, fsize, f);
-  fclose(f);
+  buffer = buffer.substr(0, i);
```
trivialfis, line 142: Why is it a substr?

tests/cpp/common/test_io.cc
```diff
@@ -39,5 +44,50 @@ TEST(IO, FixedSizeStream) {
     ASSERT_EQ(huge_buffer, out_buffer);
   }
 }
+
+
+#if SIZE_MAX == 0xFFFFFFFFFFFFFFFF  // Only run this test on 64-bit system
+TEST(IO, LoadSequentialFile) {
```
trivialfis, line 50: Could you please clarify the bug a bit more?  I'm confused.

> The value returned may be less than nbyte if the number of bytes left in the file is less than nbyte

Can we rephrase it to: maximum returned value is actual bytes left in the file being read?  If yes, then what's the condition to trigger the bug that we are trying to fix?  Since as long as we don't try to read extra non-existing bytes, the returned value should be correct no matter how large is the file right?  And here we estimate the file size by:
```
  size_t f_size_bytes = fs.st_size;
  buffer.resize(f_size_bytes + 1);
```
So, if `fs.st_size` is a correct value, then we are reading exactly the number of bytes in file. Why would we end up with an half loaded buffer?


Looking at the test, it seems the real cause is (quoting from the link you provided):

> If the value of nbyte is greater than {SSIZE_MAX}, the result is implementation-defined.

Can we add an optional parameter to the load function specifying batch size for test, instead of creating a huge file?

trivialfis: Also, as now we have replaced the number printing algorithm, generated file should be much much smaller.

include/xgboost/json_io.h
hcho3: No, on some systems, EOF is not necessarily -1.

tests/cpp/common/test_io.cc
hcho3: The bug has nothing to do with SSIZE_MAX. The bug has to do with the behavior of read(). When we call read() function with nbytes=2.3GB, it does not actually read 2.3GB. So we must call read() twice or more to get all the bytes into the buffer; otherwise we end up with half full buffer.

src/common/io.cc
hcho3: My bad. It should be resize() so that no copy is made.

tests/cpp/common/test_io.cc
hcho3: > Can we add an optional parameter to the load function specifying batch size for test, instead of creating a huge file?

Not sure what you mean here. How can we test the behavior of read() with nbytes=2.3GB without creating a tempfile of size 2.3GB?

include/xgboost/json_io.h
hcho3: Sorry, you are right. We are reading from a char array, so any EOF that appears would necessarily be -1. On ARM chips, EOF is actually bigger than char, so this case won’t apply.

src/common/io.cc
hcho3: Removed.

src/common/io.cc
hcho3: Never mind. I’ll remove this change.

tests/cpp/common/test_io.cc
hcho3: > Can we rephrase it to: maximum returned value is actual bytes left in the file being read?

No. The standard lists a few possible scenarios where read() returns fewer bytes than the number of bytes requested (nbytes parameter). One possible scenario is to have a short file. For this particular bug, we do have a complete file, yet read() returns fewer bytes. I honestly don’t understand why read() gives fewer than 2.3GB.

This behavior is quite well known. The prevailing wisdom in UNIX community is to wrap read() in a while loop.

hcho3:
@trivialfis I removed the extra changes, as you requested. Can I get another round of review?

src/common/io.cc
```diff
@@ -120,27 +121,22 @@ std::string LoadSequentialFile(std::string fname) {
 #if defined(__linux__)
   posix_fadvise(fd, 0, 0, POSIX_FADV_SEQUENTIAL);
 #endif  // defined(__linux__)
-  ssize_t bytes_read = read(fd, &buffer[0], f_size_bytes);
-  if (bytes_read < 0) {
-    close(fd);
-    ReadErr();
+  ssize_t bytes_read = 0;
+  while (bytes_read < f_size_bytes) {
```
hcho3, line 125: @trivialfis Take a look at this while loop. This is a commonly used construct in UNIX programs.

src/common/io.cc
```diff
@@ -120,27 +121,22 @@ std::string LoadSequentialFile(std::string fname) {
 #if defined(__linux__)
   posix_fadvise(fd, 0, 0, POSIX_FADV_SEQUENTIAL);
 #endif  // defined(__linux__)
-  ssize_t bytes_read = read(fd, &buffer[0], f_size_bytes);
```
hcho3, line 123: @trivialfis Previously, we called `read()` only once. For a file containing 2896075944 bytes (2.69 GB), `read()` only reads 2147479552 bytes (1.99 GB), and the variables would be set as follows:
```
bytes_read = 2147479552, f_size_bytes = 2896075944
```

This PR ensures that `bytes_read` would be identical to `f_size_bytes`.

src/common/io.cc
trivialfis: > std::ifstream shows competitive performance with respect to std::fread().

Interesting as I just tried on Linux, it's about 10x slower.  But I guess when file is that huge JSON would be the bottleneck here.

tests/cpp/common/test_io.cc
trivialfis: > Not sure what you mean here. How can we test the behavior of read() with nbytes=2.69 GB without creating a tempfile of size 2.69 GB?

What I'm worrying about is the behaviour is dependent on other factors like available memory.  So 2.69 GB might not be a correct test.

PR #5853, opened 2020-07-03
[BLOCKING] Thread-safe prediction by making the prediction cache thread-local.

boxdot:
Also remove prediction cache mutex, which did not help: a race
condition still occured whenever the cache was cleaned but at the
same time entries pointing into the cache were alive.

### DIFF
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -1384,6 +1384,5 @@ XGBOOST_REGISTER_TREE_UPDATER(QuantileHistMaker, "grow_quantile_histmaker")
     []() {
       return new QuantileHistMaker();
     });
-
 }  // namespace tree
 }  // namespace xgboost
--- a/tests/cpp/test_learner.cc
+++ b/tests/cpp/test_learner.cc
@@ -3,6 +3,7 @@
  */
 #include <gtest/gtest.h>
 #include <vector>
+#include <thread>
 #include "helpers.h"
 #include <dmlc/filesystem.h>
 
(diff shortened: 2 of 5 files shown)


trivialfis:
@boxdot Could you please elaborate on this a bit more?  How can I trigger a race condition in current code base.

> entries pointing into the cache

The `PredictionCacheEntry`?  The data has to go out of scope before the entry does right?  In which case data passed into `Learner::Predict` goes out of scope before the entry does?

boxdot:
Sorry for the late response, I was on vacation. I added a test to pin data races by using thread sanitizer. Looks like there are still data races. When I have them all I'll describe the problems.

trivialfis:
@boxdot Be careful that Thread sanitizer doesn't do well with openmp.  It will generate a lots of false positive.

boxdot:
> @boxdot Could you please elaborate on this a bit more? How can I trigger a race condition in current code base.

I added a test which triggers a race.

> The `PredictionCacheEntry`? The data has to go out of scope before the entry does for triggering a race right? Could you please guide me through the case that, data passed into `Learner::Predict` goes out of scope before the entry does?

```c++
this->PredictRaw(data.get(), &prediction, training, ntree_limit);
```

in `learner.cc:1018` modifies the cache entry. This can happen from different threads at the same time. Even the cache retrieval method `Cache` is thread-safe, the read/write access to cache entries is not. If you want, I can also add my thread sanitizer output.

> @boxdot Be careful that Thread sanitizer doesn't do well with openmp. It will generate a lots of false positive.

Thanks, I also figured that out. :) I run thread sanitizer with disabled OpenMP, and now it looks good.

hcho3:
> Testing rejected. Reason: Today's spending (56.52 USD) exceeded the budget (50.00 USD) allocated for today! The spending limit gets reset every midnight UTC. You can monitor the spending at the dashboard https://xgboost-ci.net/dashboard/.

It looks like we've hit today's cloud budget. The tests hosted on Jenkins will not run. We'll re-run these tests in 16 hours.

tests/cpp/test_learner.cc
```diff
@@ -176,6 +177,52 @@ TEST(Learner, JsonModelIO) {
...
 
+// Crashes the test runner if there are race condiditions.
+//
+// Build with additional cmake flags to enable thread sanitizer
+// which definitely catches problems. Note that OpenMP needs to be
+// disabled, otherwise thread sanitizer will also report false
+// positives.
+//
+// ```
+// -DCMAKE_C_FLAGS="-fsanitize=thread -g -O1"
```
trivialfis, line 188: XGBoost comes with tsan support in CMake. `-DUSE_SANITIZER=ON`, `-DENABLED_SANITIZERS=thread`

trivialfis:
> in learner.cc:1018 modifies the cache entry. This can happen from different threads at the same time. Even the cache retrieval method Cache is thread-safe, the read/write access to cache entries is not.

That helped a lot.  Looks good to me in general.  Thanks for fixing the data race.

trivialfis:
@hcho3 Marked blocking.  I will try to push to this if needed.

tests/cpp/test_learner.cc
boxdot: Thanks. Adjusted the docs.

trivialfis:
Looking into the failing tests.

trivialfis:
@boxdot I made the cache local to each learner.  Could you please take a look?

tests/cpp/test_learner.cc
```diff
@@ -207,7 +207,7 @@ TEST(Learner, MultiThreadedPredict) {
   std::vector<std::thread> threads;
   for (int32_t thread_id = 0;
        thread_id < 2 * std::thread::hardware_concurrency(); ++thread_id) {
-    threads.push_back(std::thread([learner, p_data] {
+    threads.emplace_back(std::thread([learner, p_data] {
```
boxdot, line 210: I guess you can remove `std::thread` ctor, too?

src/tree/updater_quantile_hist.cc
```diff
@@ -1372,9 +1372,5 @@ XGBOOST_REGISTER_TREE_UPDATER(QuantileHistMaker, "grow_quantile_histmaker")
     []() {
       return new QuantileHistMaker();
     });
-
-template struct QuantileHistMaker::Builder<float>;
-template struct QuantileHistMaker::Builder<double>;
```
boxdot: JFYI: This was need on my compiler (g++ (GCC) 10.1.1). Did not follow though why.

trivialfis:
Thanks, let me keep fixing it.

src/tree/updater_quantile_hist.cc
trivialfis: @boxdot  Thanks, we recently fixed it on master branch.  So the PR creates a conflict.

tests/cpp/test_learner.cc
trivialfis: Removed.

boxdot:
> @boxdot I made the cache local to each learner. Could you please take a look?

Ah, very nice, this is much better. My colleague, @VeaaC, pointed out that it might be a problem with the Singleton implementation (no modern thread local), that it would leak unused cached. This is solved now, as far as I understand.

trivialfis:
@boxdot Thanks for fixing the race condition!

PR #5870, opened 2020-07-07
fix device sketch with weights in external memory mode

### LINKED ISSUE #5866: Training fails when using external memory version for large datasets with instance weights
I am attempting to train an XGBoost model using a large dataset that I cannot completely load in memory. So I decided to use the external memory feature of XGBoost training, like so:

    dtrain = xgb.DMatrix(f"data/train.libsvm#train.cache", feature_names=feature_names)

Now I also need to be able to specify instance weights, so I tried to do that by specifying the weights in a separate `train.libsvm.weight` file:

```
$ head -5 data/train.libsvm.weight
5.28486226776928e-7
5.28486226776928e-7
5.28486226776928e-7
5.28486226776928e-7
5.28486226776928e-7
```

However, training fails with the following error:

```
[22:15:43] 11252555x174 matrix with 44859377 entries loaded from data/train.libsvm#train.cache
[22:15:46] 11252555 weights are loaded from data/train.libsvm.weight
  File "train.py", line 104, in <module>
    train(sys.argv[1], sys.argv[2], sys.argv[3], sys.argv[4])
  File "train.py", line 66, in train
    model = xgb.train(params, dtrain, num_rounds, watchlist)
  File "/home/praveen/auto-test-web/auto-test-web/src/ml/venv/lib/python3.8/site-packages/xgboost/training.py", line 208, in train
    return _train_internal(params, dtrain,
  File "/home/praveen/auto-test-web/auto-test-web/src/ml/venv/lib/python3.8/site-packages/xgboost/training.py", line 75, in _train_internal
    bst.update(dtrain, i, obj)
  File "/home/praveen/auto-test-web/auto-test-web/src/ml/venv/lib/python3.8/site-packages/xgboost/core.py", line 1367, in update
    _check_call(_LIB.XGBoosterUpdateOneIter(self.handle,
  File "/home/praveen/auto-test-web/auto-test-web/src/ml/venv/lib/python3.8/site-packages/xgboost/core.py", line 190, in _check_call
xgboost.core.XGBoostError: [22:15:46] /workspace/src/tree/updater_gpu_hist.cu:952: Exception in gpu_hist: [22:15:46] /workspace/src/common/hist_util.cu:287: Check failed: weights.size() == page.offset.Size() - 1 (11252555 vs. 921785
```

So from the error message, it appears the weights file is recognized and weights for all 1125255 instances are loaded. However, since we are training in batches (using the external memory feature), we only load 921785 instances and the requirement that the size of the weights vector should equal that of the training dataset doesn't hold.

I have also tried specifying weights in the LibSVM input file directly - by replacing the `label` entry with `label:weight`, but I get the exact same error. 

I'm using XGBoost version 1.1.0.

rongou:
Turns out the code already does the right thing, it's just the `CHECK` that's making an incorrect assumption.

Fixes #5866 

@trivialfis @hcho3

### DIFF
--- a/tests/cpp/common/test_hist_util.cu
+++ b/tests/cpp/common/test_hist_util.cu
@@ -209,6 +209,24 @@ TEST(HistUtil, DeviceSketchMultipleColumnsExternal) {
   }
 }
 
+// See https://github.com/dmlc/xgboost/issues/5866.
+TEST(HistUtil, DeviceSketchExternalMemoryWithWeights) {
+  int bin_sizes[] = {2, 16, 256, 512};
+  int sizes[] = {100, 1000, 1500};
+  int num_columns = 5;
+  for (auto num_rows : sizes) {
+    auto x = GenerateRandom(num_rows, num_columns);
+    dmlc::TemporaryDirectory temp;
+    auto dmat =
+        GetExternalMemoryDMatrixFromData(x, num_rows, num_columns, 100, temp);
+    dmat->Info().weights_.HostVector() = GenerateRandomWeights(num_rows);
+    for (auto num_bins : bin_sizes) {
+      auto cuts = DeviceSketch(0, dmat.get(), num_bins);
+      ValidateCuts(cuts, dmat.get(), num_bins);
+    }
+  }
+}
+
 template <typename Adapter>
 void ValidateBatchedCuts(Adapter adapter, int num_bins, int num_columns, int num_rows,
                          DMatrix* dmat) {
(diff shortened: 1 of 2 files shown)


tests/cpp/common/test_hist_util.cu
```diff
@@ -209,6 +209,24 @@ TEST(HistUtil, DeviceSketchMultipleColumnsExternal) {
...
 }
 
+// See https://github.com/dmlc/xgboost/issues/5866.
+TEST(HistUtil, DeviceSketchExternalMemoryWithWeights) {
+  int bin_sizes[] = {2, 16, 256, 512};
+  int sizes[] = {100, 1000, 1500};
+  int num_columns = 5;
+  for (auto num_rows : sizes) {
+    auto x = GenerateRandom(num_rows, num_columns);
+    dmlc::TemporaryDirectory temp;
```
hcho3, line 219: Please move this out of the for loop.

rongou: Done.

PR #5871, opened 2020-07-08
[CI] Simplify CMake build with modern CMake techniques

hcho3:
* Starting CMake 3.12, it is now possible to refer to object libraries in `target_link_libraries()` command. When a shared library links with an object library, the objects and compilation/link flags associated with the object library get automatically propagated to the shared library. So we don't need variables like `LINKED_LIBRARIES_PRIVATE`, `XGBOOST_OBJ_SOURCES`, and `XGBOOST_DEFINITIONS`. Automatic propagation of flags is less error-prone than using the variables to copy the flags. See [More Modern CMake](https://github.com/Bagira80/More-Modern-CMake).

* Use `target_sources()` instead of `PLUGIN_SOURCES` and `XGBOOST_OBJ_SOURCES`.

* Make sure that the LZ4 and Dense parser plugins can be built. #4879 broke the build for the plugins, but we didn't catch it because the CI wasn't testing these plugins. For now, add the plugins to the CPU-only build.

This PR will also be useful for #5861, or whenever we want to add a new C++ dependency really.

### DIFF
--- a/R-package/CMakeLists.txt
+++ b/R-package/CMakeLists.txt
@@ -6,8 +6,8 @@ file(GLOB_RECURSE R_SOURCES
   ${CMAKE_CURRENT_LIST_DIR}/src/*.c)
 # Use object library to expose symbols
 add_library(xgboost-r OBJECT ${R_SOURCES})
-
-set(R_DEFINITIONS
+target_compile_definitions(xgboost-r
+  PUBLIC
   -DXGBOOST_STRICT_R_MODE=1
   -DXGBOOST_CUSTOMIZE_GLOBAL_PRNG=1
   -DDMLC_LOG_BEFORE_THROW=0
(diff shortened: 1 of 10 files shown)


plugin/CMakeLists.txt
```diff
@@ -1,9 +1,8 @@
 if (PLUGIN_LZ4)
-  set(PLUGIN_SOURCES ${PLUGIN_SOURCES}
```
hcho3, line 2: Notice the typo: `PLUGIN_SOURCES` should have been `PLUGINS_SOURCES`.

trivialfis:
I think we have some feature requests for having one common library that can be reused on distribution.

hcho3:
The R binding is tricky, because it makes changes to `libxgboost` by setting unique flags. So R binding cannot share the common library.

R-package/CMakeLists.txt
```diff
@@ -6,36 +6,32 @@ file(GLOB_RECURSE R_SOURCES
...
-set(R_DEFINITIONS
+target_compile_definitions(xgboost-r
+  PUBLIC
   -DXGBOOST_STRICT_R_MODE=1
   -DXGBOOST_CUSTOMIZE_GLOBAL_PRNG=1
   -DDMLC_LOG_BEFORE_THROW=0
   -DDMLC_DISABLE_STDIN=1
   -DDMLC_LOG_CUSTOMIZE=1
   -DRABIT_CUSTOMIZE_MSG_
   -DRABIT_STRICT_CXX98_)
```
hcho3, line 17: These flags are unique to R and they make changes to `objxgboost`. The resulting `libxgboost.so` is substantially different from what the Python and JVM bindings need.

hcho3: To share a common library, we'd need to carefully extract R-specific code into a separate CMake target, so that the R flags only modify the separate target and not the common library. This may be nontrivial effort.

trivialfis: For cran yes, for conda maybe no.  Let me take a look, see how much effort does it take.

hcho3: Can you review this PR as it is? Let's follow up with another PR for the common library discussion.

hcho3:
@trivialfis Here it is:
```
 linux-vdso.so.1 (0x00007ffd9add0000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f3854ff3000)
	libgomp.so.1 => /usr/lib/x86_64-linux-gnu/libgomp.so.1 (0x00007f3854dc4000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f3854bbc000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f38549b8000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f385462f000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f3854291000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f3854079000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3853c88000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f3865ca5000)
```

trivialfis:
I used to be able to hide libgomp and nccl, seems impossible now.  I tried keeping them private but no success.  object library is really difficult to deal with do I prefer normal shared library in usual cases.  But then Windows seems to be very unfriendly to shared library.

PR #5872, opened 2020-07-08
[Doc] Document that CUDA 10.0 is required [skip ci]

hcho3:


hcho3:
FYI, we should consider how long we should support Compute Capability 3.5.

RAMitchell:
Looks like kaggle kernels are no longer using cc 3.5: https://www.kaggle.com/product-feedback/83643

Seems reasonable for us to deprecate in the near future.

PR #5882, opened 2020-07-12
Add SHAP summary plot using ggplot2

tcuongd:
Hi! Big fan of the SHAP plots in the R package, especially since the `shap` package is only in Python. I was thinking it would be worth adding a summary plot that compares the SHAP contributions for each feature, similar to the picture below (taken from https://github.com/slundberg/shap):
image

This PR adds `ggplot2` functionality to do something similar. This is what the new function would produce on the Agaricus dataset:
```
xgb.ggplot.shap.summary(agaricus.test$data, contr, model = bst, top_n = 12)
```
image

Let me know if you guys think this would be worth having :)

I've also moved some of the data pre-processing and checks from `xgb.plot.shap` to a new utility function, so that I could reuse it for the summary plot function. Not sure if it's too messy like this, keen to get your thoughts.


trivialfis:
Restarted the travis job.

tcuongd:
> Is there anyway that we can test this by adding an unittest?

Thanks for the advice! I've added some tests to `test_helpers.R` to ensure that the schema of the data produced by `xgb.shap.data` and `prepare.ggplot.shap.data` match what is expected by the plotting functions. It seems like for the plots themselves we're currently just calling the plot function to see that it completes successfully?

Do you reckon those tests are OK or we should do more (e.g. testing different argument combinations)?

tcuongd:
I've fixed the lint errors :)

trivialfis:
ping @hcho3

fitzLab-AL:
> Hi! Big fan of the SHAP plots in the R package, especially since the `shap` package is only in Python. I was thinking it would be worth adding a summary plot that compares the SHAP contributions for each feature, similar to the picture below (taken from https://github.com/slundberg/shap):
> image
> 
> This PR adds `ggplot2` functionality to do something similar. This is what the new function would produce on the Agaricus dataset:
> 
> ```
> xgb.ggplot.shap.summary(agaricus.test$data, contr, model = bst, top_n = 12)
> ```
> 
> image
> 
> Let me know if you guys think this would be worth having :)
> 
> I've also moved some of the data pre-processing and checks from `xgb.plot.shap` to a new utility function, so that I could reuse it for the summary plot function. Not sure if it's too messy like this, keen to get your thoughts.

Can someone kindly tell me what the gray shaded points mean in the SHAP summary plot?

tcuongd:
@fitzLab-AL We center the feature values using `(x - mu) / sd`, then set the colour limits for the shap summary to `(-3, 3)`. So the grey points (uncoloured) refer to feature values that are more than 3 standard deviations away from the mean feature value.

fitzLab-AL:
Perfect - thanks for your reply!

PR #5836, opened 2020-06-29
[jvm-packages] update spark dependency to 3.0.0

wbo4958:

PR #5890, opened 2020-07-14
[CI] update spark version to 3.0.0

wbo4958:
The voting result about the strategy of spark 3.0 supporting from https://github.com/dmlc/xgboost/pull/5836#issuecomment-657345701 is "only support Spark 3.0+ and scala 2.12 since xgboost 1.2 release". So I made this PR to update the CI related spark version to 3.0.0


wbo4958:
hi @hcho3, I've made this PR to update spark version of CI to 3.0.0, But seems the line https://github.com/dmlc/xgboost/blob/master/tests/ci_build/Dockerfile.jvm_cross#L25 failed. Seems no spark 3.0.0 in S3. Could you help on that

Spark 3.0.0 image can be found in https://archive.apache.org/dist/spark/spark-3.0.0/

wbo4958:
Hi @hcho3, I have a concern if we set to the official spark address because it **maybe** blocked by the website. we have experienced this. So maybe there is some reason why the original spark has been put into S3.

hcho3:
> maybe there is some reason why the original spark has been put into S3.

The original reason was that we needed Spark 2.4.3 built with Scala 2.12 and the official URL used Scala 2.11.

> it maybe blocked by the website. we have experienced this.

Got it. It may be wise to self-host the binary then.

hcho3:
@wbo4958 Actually I think it's fine, because we aggressively cache Docker layers and so the Spark 3.0.0 installation would be cached too. The Spark binary will be downloaded only infrequently, when Docker container gets rebuilt due to security patches.

wbo4958:
@hcho3 ok, thx.

wbo4958:
@CodingCat @trivialfis Could you help to review this PR?

Issue #5891, opened 2020-07-14
[CI] Better utilization of CI resources and other CI improvements


hcho3:
Run-away cloud cost of our Jenkins CI server has been a pressing issue (#5176). It is hosted on AWS, which charges by the hour. #5884 finally created the mechanism to enforce a daily budget via throttling.

We now have a dashboard page to keep track of daily spending: https://xgboost-ci.net/dashboard/
![Screen Shot 2020-07-14 at 3 54 49 AM](https://user-images.githubusercontent.com/2532981/87417955-c5518e00-c585-11ea-9fb3-48eda7df0567.png)

Now it is time to extract savings and ensure that **we are using limited CI resources on where it matters**.

State of the CI: The free credits from AWS ran out this month, so we now have to start drawing from the Open Collective account, which currently has 10531.16 USD. If we limit ourselves to spending 33 USD per day, the balance will last 287 days.

- [ ] Make all tests conditional on the presence of a GitHub comment (e.g. `run tests`). Right now, tests run automatically, and there are many cases where automatically starting tests is wasteful.
- [ ] Skip tests for all draft pull requests.
- [ ] Allow "rolling over" left-over allowance from previous day. For example, if the daily budget is 33 USD and we spent only 10 USD yesterday, we should be able to spend up to 56 USD (33 USD + 23 USD roll over). The spending pattern is quite spiky:
![Screen Shot 2020-07-14 at 4 21 34 AM](https://user-images.githubusercontent.com/2532981/87420098-81f91e80-c589-11ea-8940-fad5f39f23bb.png)
- [x] Migrate some tests to free services. I suspect this may have limited impact, since any tests using GPUs need to run in Jenkins.
- [x] Pin point which tests cost the most. I was surprised to learn that Windows jobs cost more than 50% of the expenses, even though we run more tests in Linux:
![Screen Shot 2020-07-14 at 4 23 14 AM](https://user-images.githubusercontent.com/2532981/87420276-d7cdc680-c589-11ea-8995-7741fa18e2ce.png)
  * #5892 reduces the number of tests for Windows target by half.
  * #5904 saves the cost of Windows test pipeline by up to 66%.
- [ ] Related: write a summary report of the AWS expenses for the last 6 months.
- [ ] Speed up C++ builds. It takes ~ 10 min on Linux and ~ 15 min on Windows to build XGBoost with GPU support.
- [ ] Try to get more funding, which is easier said than done.

Other CI improvements, outside of Jenkins
[...]

PR #5892, opened 2020-07-14
[CI] Reduce load on Windows CI pipeline

hcho3:
As discussed in #5891, tests for the Windows platform cost a disproportionate amount of the expenses, even though a majority of the tests overall run in Linux:
Screen Shot 2020-07-14 at 4 23 14 AM

Until a better solution is found, we reduce the load by removing CUDA 10.0 target from the Jenkins pipeline for Windows.


hcho3:
@trivialfis I'm quite willing to move away from AppVeyor, since it only allows one test job at a time. I will add items in the list.

PR #5893, opened 2020-07-15
Simplify the data backends.

trivialfis:
This PR removes data handler and just use plain old conditions with glue functions.  It doesn't simplify the logic for gluing various data sources, but merely makes the code easier to track.  After this PR, we will have explicit list of supported data for `DMatrix`, `DeviceQuantileDMatrix` and corresponding meta info.  3 entry functions are:

- `dispatch_data_backend`
- `dispatch_meta_backend`
- `dispatch_device_quantile_dmatrix_set_data`

After this PR, we need to renew our document for what's supported for `DMatrix` and meta info.

Another important issue in our `DMatrix` initialization code is, for CPU input we accept only floating point value and ulong sparse index.  Whenever there's a mismatch from user provided data, `ctypes` makes a copy.  Some feedback on XGBoost consuming too much CPU memory actually come from this.  Also we made a lots of code for flattening the data and copying if data is a view.  It would be very helpful if we can add new C API that handles these details in a general way without making any data copy.  CC @ShvetsKS @SmirnovEgorRu on this.

One last thing is, as we need to recognize categorical data from dataframe in the future, this PR makes it easier.  When we start to support categorical data, all we need to change is adding an additional boolean parameter for whether should we honour the type info in dataframe.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -4,13 +4,14 @@
 """Core XGBoost Library."""
 import collections
 # pylint: disable=no-name-in-module,import-error
-from collections.abc import Mapping  # Python 3
+from collections.abc import Mapping
 # pylint: enable=no-name-in-module,import-error
 import ctypes
 import os
 import re
 import sys
 import json
+import warnings
 
 import numpy as np
 import scipy.sparse
@@ -267,7 +268,6 @@ def _convert_unknown_data(data, meta=None, meta_type=None):
             raise TypeError('Can not handle data from {}'.format(
                 type(data).__name__)) from e
     else:
-        import warnings
         warnings.warn(
             'Unknown data type: ' + str(type(data)) +
             ', coverting it to csr_matrix')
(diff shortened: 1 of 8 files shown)


python-package/xgboost/core.py
```diff
@@ -706,10 +629,8 @@ def set_group(self, group):
         group : array like
             Group size of each group
         """
-        if _has_cuda_array_interface(group):
-            self.set_interface_info('group', group)
-        else:
-            self.set_uint_info('group', group)
+        from .data import dispatch_meta_backend
```
RAMitchell, line 632: Any reason not to use a top-level import?

trivialfis: Will change it to top level.

trivialfis:
@hcho3 Could you please take a look.  I will try to address all comments at once.

python-package/xgboost/core.py
trivialfis: Actually no.  I did this because it can avoid circular module import.

python-package/xgboost/data.py
```diff
@@ -1,553 +1,660 @@
...
-from .core import DataIter
-from .compat import lazy_isinstance, STRING_TYPES, os_fspath, os_PathLike
+from .core import c_array, _LIB, _check_call, c_str
+from .core import DataIter, DeviceQuantileDMatrix, DMatrix
+from .compat import lazy_isinstance, os_fspath, os_PathLike
 
 c_bst_ulong = ctypes.c_uint64   # pylint: disable=invalid-name
 
 
-class DataHandler(abc.ABC):
```
hcho3, line 17: So we are completely abandoning `DataHandler` class?

python-package/xgboost/data.py
```diff
@@ -1,553 +1,660 @@
...
+    if _is_cupy_csc(data):
+        raise TypeError('cupyx CSC is not supported yet.')
+    if _is_dlpack(data):
+        return _from_dlpack(data, missing, threads, feature_names,
+                            feature_types)
+    if _is_dt_df(data):
+        _warn_unused_missing(data, missing)
+        return _from_dt_df(data, missing, threads, feature_names,
+                           feature_types)
+    if _has_array_protocol(data):
```
hcho3, line 528: How do we handle data types with `__array__` protocol? Does it get converted into a NumPy array?

trivialfis: Not really, I can't push a cupy array to numpy array.  It's called `array_like`, we need to add extra dispatching logic for it.  I don't want to do it in this PR.

python-package/xgboost/data.py
trivialfis: Yes.  Just plain old function dispatching.  Not clean, but at least debugger friendly ...

python-package/xgboost/data.py
hcho3: Got it. Do you plan to write another PR for dispatching array-like data?

trivialfis: @hcho3 Will do it later.  It's required for passing scikit-learn validations.

PR #5854, opened 2020-07-04
Add DMatrix usage examples to c-api-demo

alexanderGugel:
I had a bit of a hard time using the C API given that there aren't that many usage examples, so maybe this will be useful for people.

This PR extends the C API demo to construct a dense matrix using `XGDMatrixCreateFromMat`. I'm planning on adding additional examples for `XGDMatrixCreateFromCSREx`, `XGDMatrixCreateFromCSCEx`, but unfortunately I haven't been able to get those working yet.

The matrix being constructed is equivalent to the first line of the test data.z

### Test Plan

```
alexanders-mbp:c-api alex$ make run
cc -O3 -I../../include -I../../dmlc-core/include -I../../rabit/include -L../../lib -o c-api-demo c-api-demo.c -lxgboost
LD_LIBRARY_PATH=../../lib ./c-api-demo
[17:32:50] 6513x127 matrix with 143286 entries loaded from ../data/agaricus.txt.train
[17:32:50] 1611x127 matrix with 35442 entries loaded from ../data/agaricus.txt.test
[0]     train-error:0.014433    test-error:0.016139
[1]     train-error:0.014433    test-error:0.016139
[2]     train-error:0.014433    test-error:0.016139
[3]     train-error:0.008598    test-error:0.009932
[4]     train-error:0.001228    test-error:0.000000
[5]     train-error:0.001228    test-error:0.000000
[6]     train-error:0.001228    test-error:0.000000
[7]     train-error:0.001228    test-error:0.000000
[8]     train-error:0.001228    test-error:0.000000
[9]     train-error:0.001228    test-error:0.000000
y_pred: 0.0239 0.9544 0.0239 0.0239 0.0490 0.1056 0.9544 0.0288 0.9544 0.0242 
y_test: 0.0000 1.0000 0.0000 0.0000 0.0000 0.0000 1.0000 0.0000 1.0000 0.0000 
Dense Matrix Example (XGDMatrixCreateFromMat): 0.0239 
```

PR #5896, opened 2020-07-16
Simplify CSRAdapter

alexanderGugel:
Whilst working on #5854, I noticed that the nelem argument supplied to
`XGDMatrixCreateFromCSREx` seems to have no effect whatsoever.

`CSRAdapterBatch` currently doesn't seem to store this field, let alone
do anything with it.

As such, it seems like it can be removed entirely.

This PR removes references to it from CSRAdapter and CSRAdapterBatch.
The signature of the C API functions has not been modified to ensure
backwards compatibility.


RAMitchell (approved):

trivialfis:
I need to check that whether this is a bug or a redundant parameter.  On distributed environment, a worker can has a data partition with missing columns at the end, in which case this parameter can be useful.

PR #5897, opened 2020-07-16
Add option to enable all compiler warnings in GCC/Clang

hcho3:
`ENABLE_ALL_WARNINGS` option will add `-Wall -Wextra` flag to all CMake targets, enabling all compiler warnings from GCC/Clang. Also, the Jenkins CI will now perform Linux builds with `ENABLE_ALL_WARNINGS` turned on.


src/CMakeLists.txt
```diff
@@ -70,6 +70,11 @@ if (MSVC)
   )
 endif (MSVC)
 
+if (ENABLE_ALL_WARNINGS)
+  target_compile_options(objxgboost PUBLIC
+    $<IF:$<COMPILE_LANGUAGE:CUDA>,-Xcompiler=-Wall -Xcompiler=-Wextra,-Wall -Wextra>)
```
hcho3, line 75: Note: NVCC doesn't accept `-Wall` option. We need to pass `-Wall` to the host compiler, so use `-Xcompiler=-Wall` instead.

PR #5904, opened 2020-07-17
Further improvements and savings in Jenkins pipeline

hcho3:
- [x] Reduce the number of machine images (AMIs) with Windows OS down to 1. All Windows jobs will run in the same kind of worker. **This change reduces the unit cost of the Windows test pipeline by 2/3 (66%).** To understand how this dramatic saving is achieved, we should understand [how AWS bills for EC2 machines](https://aws.amazon.com/premiumsupport/knowledge-center/ec2-instance-hour-billing/). Quote from the official doc:

> Each partial instance hour consumed is billed per-second for Linux instances and as a full hour for all other instance types.

For example, if we launch a Linux EC2 instance, run it for 10 minutes and shut it down, then we only pay 1/6 of the hourly price. On the other hand, if we launch a Windows instance and shut it down 10 minutes later, then we have to pay the full hourly price, since 10 minutes gets rounded up to the next full hour. The difference is more dramatic if we launch multiple machines. Using two Linux instances for 10 minutes each costs 1/3 of the hourly price, whereas using two Windows instances for 10 minutes each cost twice the hourly price!

Now let's look at how the Windows testing pipeline is Jenkins is affected by the hourly pricing. Since all steps in the pipeline are at most 15 minutes long, we end up wasting a lot of money:

  * Before this PR.
Screen Shot 2020-07-17 at 7 18 06 PM

Cost for running the pipeline = Two hours C5a4xlarge worker-hour + One G4dnxlarge worker-hour = $1.352 * 2 + $0.71 = **$3.414**

  * After this PR.
Screen Shot 2020-07-17 at 7 26 51 PM

Cost for running the pipeline = One G4dn2xlarge worker-hour = **$1.12**

- [x] Do not publish artifacts if the job is associated with a pull request. This measure saves space in S3.
- [x] Build CUDA code for Compute Capability 7.5 only, if the job is associated with a pull request. This measure speeds up the build.
- [x] Build nightly XGBoost4J SNAPSHOT JARs with Scala 2.12 only, now that Spark 3.0.0 dropped support for Scala 2.11.
- [x] Show skipped Python tests on Windows
- [x] Make Graphviz optional for Python tests

### DIFF
--- a/Jenkinsfile-win64
+++ b/Jenkinsfile-win64
@@ -125,21 +125,19 @@ def TestWin64() {
     bat "nvcc --version"
     echo "Running C++ tests..."
     bat "build\\testxgboost.exe"
-    echo "Installing Python wheel..."
-    bat "conda activate && (python -m pip uninstall -y xgboost || cd .)"
-    bat """
-    conda activate && for /R %%i in (python-package\\dist\\*.whl) DO python -m pip install "%%i"
-    """
     echo "Installing Python dependencies..."
+    def env_name = 'win64_' + UUID.randomUUID().toString().replaceAll('-', '')
+    bat "conda env create -n ${env_name} --file=tests/ci_build/win64_conda_env.yml"
+    echo "Installing Python wheel..."
     bat """
-     conda activate && conda install -y hypothesis jsonschema && conda upgrade scikit-learn pandas numpy hypothesis  && python -m pip install cupy-cuda101
+    conda activate ${env_name} && for /R %%i in (python-package\\dist\\*.whl) DO python -m pip install "%%i"
     """
     echo "Running Python tests..."
-    bat "conda activate && python -m pytest -v -s -rxXs --fulltrace tests\\python"
+    bat "conda activate ${env_name} && python -m pytest -v -s -rxXs --fulltrace tests\\python"
     bat """
-    conda activate && python -m pytest -v -s -rxXs --fulltrace -m "(not slow) and (not mgpu)" tests\\python-gpu
+    conda activate ${env_name} && python -m pytest -v -s -rxXs --fulltrace -m "(not slow) and (not mgpu)" tests\\python-gpu
     """
-    bat "conda activate && python -m pip uninstall -y xgboost"
+    bat "conda env remove --name ${env_name}"
     deleteDir()
   }
 }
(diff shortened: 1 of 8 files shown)


hcho3:
> Let's keep all the computer capabilities for Linux node for now

They are all there for the master branch. Do you need them to run for pull requests too?

hcho3:
> the pre-build binary

Does asyncio require a wheel with multiple arch code in it?

trivialfis:
They are running on Google cloud.

trivialfis:
For testing the asyncio PR.

hcho3:
@trivialfis This PR stops publishing nightly wheels for pull requests. My plan was to make nightlies only for the master and release branches.

trivialfis:
Got it.  I will try to help them building XGBoost locally.

hcho3:
@trivialfis Should I delay the change about artifact publishing until you are done with asyncio PR?

tests/python/test_with_sklearn.py
```diff
@@ -436,7 +436,7 @@ def test_sklearn_api_gblinear():
     assert err < 0.5
 
 
-@pytest.mark.skipif(**tm.no_matplotlib())
+@pytest.mark.skipif(**tm.no_multiple(tm.no_matplotlib(), tm.no_graphviz()))
```
trivialfis, line 439: You can use `skipif` multiple times.

Jenkinsfile-win64
```diff
@@ -137,13 +131,13 @@ def TestWin64GPU(args) {
     conda activate && for /R %%i in (python-package\\dist\\*.whl) DO python -m pip install "%%i"
     """
     echo "Installing Python dependencies..."
-    def cuda_short_ver = args.cuda_target.replaceAll('_', '')
     bat """
-     conda activate && conda install -y hypothesis && conda upgrade scikit-learn pandas numpy hypothesis  && python -m pip install cupy-${cuda_short_ver}
+     conda activate && conda install -y hypothesis jsonschema && conda upgrade scikit-learn pandas numpy hypothesis  && python -m pip install cupy-cuda101
```
trivialfis, line 135: Can we write this down to a  pip requirement file?

PR #5909, opened 2020-07-19
[Doc] Document new objectives and metrics available on GPUs

hcho3:


hcho3:
@trivialfis We don't support `survival:cox` and `aucpr` yet?

PR #5832, opened 2020-06-27
[WIP] [CI] Use Ninja to speed up builds

hcho3:
According to [this talk from the last year](https://slides.com/onqtam/faster_builds), Ninja is significantly faster than MSBuild.

Let the CI run with Ninja and see how much time saving is realized.

PR #5917, opened 2020-07-20
Setup github action.

trivialfis:
Existing tests are still here.  I plan to remove appveyor once this PR is approved as a replacement.  Also, it's possible to replace some Jenkins CPU tests.  As proposed by @hcho3, we can run all default tests on action and travis like linting and some CPU tests, and run Jenkins tests only when approved.  This PR should be able to get us closer to that goal.

## Added Tests

* Add tests for R package on Windows, including mingw64, msvc with windows lastest and 2016.
~* Add CPP and Python tests on Windows latest and Ubuntu latest.~
* Add Java tests on Windows latest, Windows 2016 and Ubuntu latest. (no spark)

The test matrix can be changed, suggestions are welcomed.

## Skipped tests

- `SparsePage.PushCSCAfterTranspose'` in gtest is skipped due to an error in external memory.  So far I can't reproduce it on my local machine.  Will investigate further with a tracking issue.
- pydatatable test is skipped due to failing to install on Windows.  I can try adding it back on Ubuntu and open an issue there.
- 32 bit build is not tested.  Maybe we can setup c++ tests for 32 bit, not sure if that's required.  But for other language bindings it's quite troublesome to setup 32 bit toolchain.
- mingw tests are only used for R.  Any other Windows tests are compiled with msvc.  That's because Rtools comes with a copy of mingw and is easy to setup. Please let me know if we should run mingw for all language bindings.
- Only testing against latest R.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,6 @@
 # This is a basic workflow to help you get started with Actions
 
-name: XGoost-CI
+name: XGBoost-CI
 
 # Controls when the action will run. Triggers the workflow on push or pull request
 # events but only for the master branch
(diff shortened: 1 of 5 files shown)


hcho3:
Link for failing external memory test: https://xgboost-ci.net/blue/organizations/jenkins/xgboost-win64/detail/PR-5917/2/pipeline

trivialfis:
This is a combination of 2 bugs.  The first one is in test, which sets the number of features too small.  Normally this is prevented in Learner via DMatrix validation.  But as in unittest we use predictor directly, hence bypassing that check.  The second one is the external memory on Windows is not splitting the data correctly so data is returned in a single batch.  Hence with enough data in a batch, a memory write error happens in gpu predictor.

trivialfis:
I fixed the first one, second one is a to do item now.

trivialfis:
On the other hand, this is not reproducible on my machine: https://github.com/dmlc/xgboost/pull/5917/checks?check_run_id=892509713

trivialfis:
Removed cpp and Python tests for now to bypass failing tests on Windows.

tests/cpp/helpers.cc
```diff
@@ -357,11 +359,15 @@ std::unique_ptr<DMatrix> CreateSparsePageDMatrix(
     row_count += batch.Size();
   }
 #if defined(_OPENMP)
-  EXPECT_GE(batch_count, 2);
-  EXPECT_EQ(row_count, dmat->Info().num_row_);
+  if (omp_get_max_threads() != 1) {
+    EXPECT_GE(batch_count, 2) << " Threads: " << omp_get_max_threads();
+    EXPECT_EQ(row_count, dmat->Info().num_row_);
+  }
```
hcho3, line 365: FYI, this change might help us transition to Ninja build (#5832)

trivialfis: The root cause is in external memory.  We need to find a time to put a new impl together.

hcho3: Time to face the monster, then.

.github/workflows/main.yml
```diff
@@ -1,23 +1,53 @@
...
+      run: |
+        cd jvm-packages
+        mvn test -pl :xgboost4j_2.12
+
+
   test-with-R:
     runs-on: ${{ matrix.config.os }}
 
-    name: Test R on OS ${{ matrix.config.os }}, R (${{ matrix.config.r }})
+    name: Test R on OS ${{ matrix.config.os }}, R ${{ matrix.config.r }}, Compiler ${{ matrix.config.compiler }}, Build ${{ matrix.config.build }}
```
hcho3, line 37: Quite nice of the R community to publish GitHub Actions like `r-lib/actions/setup-r`. It would have been quite painful to set everything up otherwise.

tests/ci_build/test_r_package.py
```diff
@@ -0,0 +1,103 @@
...
+        if args.compiler == 'mingw':
+            mingw_bin = get_mingw_bin()
+            CXX = os.path.join(mingw_bin, 'g++.exe')
+            CC = os.path.join(mingw_bin, 'gcc.exe')
+            cmd = ['R.exe', 'CMD', 'INSTALL', str(os.path.curdir)]
+            env = os.environ.copy()
+            env.update({'CC': CC, 'CXX': CXX})
+            subprocess.check_call(cmd, env=env)
+        elif args.compiler == 'msvc':
+            cmd = ['R.exe', 'CMD', 'INSTALL', str(os.path.curdir)]
```
hcho3, line 38: I didn't know that you could use `R CMD INSTALL` with Visual Studio compiler. Good to know.

tests/cpp/helpers.cc
hcho3: Shall I open a new issue to keep track of the problem with external memory?

hcho3:
On a second thought, MacOS VM on Travis CI is not particularly fast. Often it takes a while until a VM gets provisioned. I should think about moving some MacOS tests that don't actually require MacOS High Sierra specifically.

Issue #5933, opened 2020-07-23
Make json configuration locale-independent 


sh1ng:
Custom json serialization like  https://github.com/dmlc/xgboost/blob/970b4b3fa2001e389910d3f6fe2af0a2966c3bd7/src/learner.cc#L102 
is fragile 

```
import locale
import xgboost
import numpy as np
locale.setlocale(locale.LC_ALL, 'sk_SK')
model = xgboost.XGBClassifier(booster='gblinear')
X = np.random.rand(100,10)
y = np.random.rand(100) > 0.5
model.fit(X, y)
print(model)
print(model.get_booster().save_config())
model.fit(X, y)
```

```
XGBClassifier(base_score='0,500000', booster='gblinear', colsample_bylevel=None,
              colsample_bynode=None, colsample_bytree=None, gamma=None,
              gpu_id=-1, importance_type='gain', interaction_constraints=None,
              learning_rate=0.5, max_delta_step=None, max_depth=None,
              min_child_weight=None, missing=nan, monotone_constraints=None,
              n_estimators=100, n_jobs=0, num_parallel_tree=None,
              objective='binary:logistic', random_state=0, reg_alpha=0,
              reg_lambda=0, scale_pos_weight=1, subsample=None,
              tree_method=None, validate_parameters=1, verbosity=None)
{"learner":{"generic_param":{"enable_experimental_json_serialization":"0","gpu_id":"-1","gpu_page_size":"0","n_gpus":"0","n_jobs":"0","nthread":"0","random_state":"0","seed":"0","seed_per_iteration":"0","validate_parameters":"1"},"gradient_booster":{"gblinear_train_param":{"max_row_perbatch":"18446744073709551615","tolerance":"0","updater":"shotgun"},"name":"gblinear","updater":{"linear_train_param":{"alpha":"0","eta":"0.5","feature_selector":"cyclic","lambda":"0","learning_rate":"0.5","reg_alpha":"0","reg_lambda":"0"}}},"learner_model_param":{"base_score":"0,500000","num_class":"0","num_feature":"10"},"learner_train_param":{"booster":"gblinear","disable_default_eval_metric":"0","dsplit":"auto","objective":"binary:logistic"},"metrics":[],"objective":{"name":"binary:logistic","reg_loss_param":{"scale_pos_weight":"1"}}},"version":[1,1,1]}
Traceback (most recent call last):
[...]

PR #5934, opened 2020-07-23
[BLOCKING] Remove to_string.

trivialfis:
Closes https://github.com/dmlc/xgboost/issues/5933 .


trivialfis:
@hcho3  Because for parameters we utilize the JSON utilities from dmlc core, which returns a string.  The model parameter is a special case, I mimic the behavior from dmlc core.

hcho3:
@trivialfis So `obj["base_score"]` must be a `std::string` and not `JsonNumber`?

trivialfis:
Sadly yes, as you know all XGBoost parameters are passed into XGBoost as string.  It's also saved as string.

PR #5123, opened 2019-12-15
[R] Basic implementation for R JSON serialization.

trivialfis:
* Change `xgb.save.raw' into full serialization instead of simple model.
* Add `xgb.load.raw' for unserialization.
* Add `xgb.config' getter and setter.

This might be breaking change as previously the `xgb.save.raw` function strips out some parameters due to binary IO.  Now it's a full serialization of all internal states.

PR #5624, opened 2020-05-01
Change type of hist buffer to float

ShvetsKS:
copy of https://github.com/dmlc/xgboost/pull/5523

This PR is supposed to change internal type of GradStats to avoid conversion for hist method. Also bug related to numerical instability in NeedReplace method.

Accuracy ~ the same as was with double type, and ex. for mnist data set was even improved:

mnist | log-loss
-- | --
Master | 0.07304085
This PR | 0.07301824

Similar changes were provided in PR:#4529 but full scope of changes was reverted in #5008.

Performance improvements (1.5x for BuildHist):

santander | full train | InitData | BuildHist | SyncHist | PredictRaw
-- | -- | -- | -- | -- | --
Master | 179.71 | 47.24 | 58.01 | 14 | 46.48
This PR | 162.94 | 47.51 | 38.12 | 8.08 | 54.7

mnist| full train | BuildHist | SyncHist | ApplySplit
-- | -- | -- | -- | --
Master | 78.511 | 34.82 | 23.29 | 1.28
This PR | 62.92 | 22.98 | 9.89 | 1.14

PR #5940, opened 2020-07-24
[BLOCKING] [R] Add a compatibility layer to load Booster object from an old RDS file

### LINKED ISSUE #5794: Predict error in R as of 1.1.1
R version: 3.6.1 (Action of the Toes)
xgboost version: 1.1.1.1

This error can be produced when attempting to call predict on an xgboost model developed pre-1.0

Error:
` Error in predict.xgb.Booster(model, data) : 
  [11:24:23] amalgamation/../src/learner.cc:506: Check failed: mparam_.num_feature != 0 (0 vs. 0) : 0 feature is supplied.  Are you using raw Booster interface? `

### LINKED ISSUE #5806: [Doc] [R] Discourage use of saveRDS() for long-term, archival storage of models
We do our best to ensure that models from older versions of XGBoost can be read into newer versions of XGBoost (**"backward compatibility"**). Our automated testing farm has a suite of tests to ensure backward compatibility.

It has been reported that the Booster object saved with `saveRDS()` function in R cannot be read back if XGBoost is upgraded. See #5794. In short, technical limitation of `saveRDS()` is such that the generated RDS file is not backward compatible.

The documentation needs to be clarified for better user experience. Couple things:
1. **Point out that `saveRDS()` is a short-term temporary storage.** Use `xgb.save()` for long-term storage.
2. **Start promoting JSON format.** It's more robust, and besides, in the worst case one can always use jsonlite or the like to manually parse the model.
3. **Provide migration examples / scripts.** Something like #5794, where users can gently transition from `saveRDS()` to long-term storage. See https://github.com/dmlc/xgboost/issues/5794#issuecomment-645254796, which shows an example script to convert an RDS file from 0.90 to JSON format (1.1.0+).

### LINKED ISSUE #5939: [BLOCKING] [R] Two R methods discards errors from C API functions
C API functions indicate errors by their return values. All callers of C API functions should check the return value and handle error if the return value is -1. There are two places in `R-package/src/xgboost_R.cc` where this is not done:

https://github.com/dmlc/xgboost/blob/a4de2f68e492065f5b90e83789a65388d349f0df/R-package/src/xgboost_R.cc#L376-L381

https://github.com/dmlc/xgboost/blob/a4de2f68e492065f5b90e83789a65388d349f0df/R-package/src/xgboost_R.cc#L398-L405

These function calls should have been wrapped with `CHECK_CALL()` macro to handle errors. Currently, errors from these two functions are silently discarded.

In #5794, the error message was cryptic
```
  [06:35:16] amalgamation/../src/learner.cc:506: Check failed: mparam_.num_feature != 0 (0 vs. 0) : 0 feature is supplied.  Are you using raw Booster interface?
Stack trace:
  [bt] (0) /home/redacted/R/x86_64-pc-linux-gnu-library/3.6/xgboost/libs/xgboost.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x64) [0x7f2da04ae154]
  [bt] (1) /home/redacted/R/x86_64-pc-linux-gnu-library/3.6/xgboost/libs/xgboost.so(xgboost::LearnerConfiguration::ConfigureNumFeatures()+0xf0) [0x7f2da054d220]
  [bt] (2) /home/redacted/R/x86_64-pc-linux-gnu-library/3.6/xgboost/libs/xgboost.so(xgboost::LearnerConfiguration::Configure()+0x356) [0x7f2da05aaba6]
  [bt] (3) /home/redacted/R/x86_64-pc-linux-gnu-library/3.6/xgboost/libs/xgboost.so(xgboost::LearnerImpl::Predict(std::shared_ptr<xgboost::DMatrix>, bool, xgboost::HostDeviceVector<float>*, unsigned int, bool, bool, bool, bool, bool)+0x7d) [0x7f2da0569e1d]
  [bt] (4) /home/redacted/R/x86_64-pc-linux-gnu-library/3.6/xgboost/libs/xgboost.so(XGBoosterPredict+0xf0) [0x7f2da04c27e0]
  [bt] (5) /home/redacted/R/x86_64-pc-linux-gnu-library/3.6/xgboost/libs/xgboost.so(XGBoosterPredict_R+0x89) [0x7f2da04a9799]
  [bt] (6) /opt/R/3.6.3/lib/R/lib/libR.so(+0x131bdd) [0x7f2dcf047bdd]
  [bt] (7) /opt/R/3.6.3/lib/R/lib/libR.so(Rf_eval+0x380) [0x7f2dcf051580]
  [bt] (8) /opt/R/3.6.3/lib/R/lib/libR.so(+0x13d121) [0x7f2dcf053121]
```
This message is misleading, as it sounds as if the error had occurred in the prediction function. In fact, the real error occurred earlier, when the model was loaded. When I inserted `CHECK_CALL()` macro in `R-package/src/xgboost_R.cc`, I got the correct error message:
```
Error in xgb.unserialize(modelfile) : 
  [03:19:04] ../src/learner.cc:864: Check failed: header == serialisation_header_: 

  If you are loading a serialized model (like pickle in Python) generated by older
  XGBoost, please export the model by calling `Booster.save_model` from that version
  first, then load it back in current version.  There's a simple script for helping
  the process. See:

    https://xgboost.readthedocs.io/en/latest/tutorials/saving_model.html

  for reference to the script, and more details about differences between saving model and
  serializing.


Stack trace:
  [bt] (0) /home/phcho/R/x86_64-pc-linux-gnu-library/3.6/xgboost/libs/xgboost.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x7c) [0x7f1f596f737c]
  [bt] (1) /home/phcho/R/x86_64-pc-linux-gnu-library/3.6/xgboost/libs/xgboost.so(xgboost::LearnerIO::Load(dmlc::Stream*)+0x81b) [0x7f1f5980bc3b]
  [bt] (2) /home/phcho/R/x86_64-pc-linux-gnu-library/3.6/xgboost/libs/xgboost.so(XGBoosterUnserializeFromBuffer+0x8c) [0x7f1f596faaec]
```

hcho3:
Fixes #5794
Fixes #5939 
Fixes #5806

**Problem**. In #5794, many R users have reported that XGBoost 1.1.1 could not load their models they had trained with an earlier version of XGBoost. They all used the built-in `saveRDS()` method to save the XGBoost model. The error message was as follows:
```
Error in predict.xgb.Booster(model, data) : [11:24:23] amalgamation/../src/learner.cc:506:
Check failed: mparam_.num_feature != 0 (0 vs. 0) : 0 feature is supplied. Are you using raw Booster interface?
```
I learned that `saveRDS()` is widely used in the R community, and that many users have saved their old models using [the RDS format](https://stat.ethz.ch/R-manual/R-devel/library/base/html/readRDS.html). Users face the painful choice between upgrading to latest XGBoost (and enjoying new features) vs. preserving access to their old models in RDS files.

**Diagnosis**. b56c9028418cea06645efcae6646ea7ba76c5069 (pull request #5123) enhanced the way `xgb.Booster` R objects are saved (serialized) to bytes, by saving additional details such as training hyperparameters. This change made old RDS files inaccessible, as the new deserialization logic (`XGBoosterUnserializeFromBuffer_R()`) was not compatible with the bytes generated by the old serialization logic (`XGBoosterModelToRaw_R()`).

**Fix**. Add a compatibility layer in `xgb.unserialize()` (See Note below). The compatibility layer first tries to call `XGBoosterUnserializeFromBuffer_R()`. If the call succeeds, then the model requested is from XGBoost 1.1.0+ and the new deserialization logic is used. On the other hand, if the call fails, then catch the error and call `XGBoosterLoadModelFromRaw_R()` to use the old deserialization logic. We use regular expression to catch the correct error.

Example: Suppose we have a RDS file that saved a XGBoost model from XGBoost 1.0.0 or 0.90:
```R
require(xgboost)
packageVersion("xgboost")   # "0.90.0.1" or "1.0.0.1"

bst <- xgboost::xgboost(data = agaricus.train$data, label = agaricus.train$label,
               max_depth = 2, eta = 1, nthread = 2, nrounds = 2,
               objective = "binary:logistic")
saveRDS(bst, file = "model_xgb100.rds")
```

The latest XGBoost, together with the pull request, will be able to read the model from the RDS file:
```R
require(xgboost)
packageVersion("xgboost")   # "1.2.0.1" (development version, compiled from the source)

bst <- readRDS("model_xgb100.rds")
pred <- predict(bst, newdata = agaricus.test$data)  # this line will actually work
```

Output:
```
Loading required package: xgboost
[1] ‘1.2.0.1’
Warning message:
In value[[3L]](cond) :
  The model had been generated by XGBoost version 1.0.0 or earlier and was loaded from a RDS file.
  We strongly ADVISE AGAINST using saveRDS() / readRDS() functions, to ensure that your model
  can be read in current and upcoming XGBoost releases.
  Please use xgb.save() instead to archive models long term.
  For more details and explanation, see https://xgboost.readthedocs.io/en/latest/tutorials/saving_model.html
```

Considerations:
* **Should we (developers) continue to allow `saveRDS()`, given that it works like pickle and has many disadvantages of pickle?**: I think it is best to use `xgb.save()` to save models for long-term archival storage. XGBoost now supports JSON serialization, so the model can be saved as a JSON file. That said, we should provide a **smooth migration path**. Many existing R applications use `saveRDS()` and `readRDS()`, and it will be a while until they are re-written to use `xgb.save()` / `xgb.load()`. I added a prominent warning to encourage users to move away from `saveRDS()`. This is important, since some of the new features we'll add in the future will require the use of JSON serialization format.

Note: How does RDS interact with XGBoost? When `saveRDS()` is called with `xgb.Booster` object, the object's native handle `object$handle` is not saved, since you cannot serialize a pointer address. Instead, the raw byte sequence `object$raw` is serialized to the RDS file. The object always holds the byte sequence that's equivalent to the native handle. (This invariant is enforced by calling `xgb.Booster.complete()` every time the native handle is modified.) Later, when we call `readRDS()`, the raw byte sequence `object$raw` is loaded back, and the native handle `object$handle` is set to null. The handle is lazily initialized when operations requiring the handle are used, such as `predict()`. The handle is constructed from the bytes `object$raw`.


hcho3:
TODO. Add model compatibility test for R package.

R-package/R/xgb.unserialize.R
```diff
@@ -6,7 +6,23 @@
...
+    .Call(XGBoosterUnserializeFromBuffer_R, handle, buffer),
+    error = function(e) {
+      error_msg <- conditionMessage(e)
+      m <- regexec("(src/learner.cc:[0-9]+): Check failed: (header == serialisation_header_)",
+                   error_msg, perl = TRUE)
+      groups <- regmatches(error_msg, m)[[1]]
+      if (length(groups) == 3) {
+        warning(paste("Loading model from a RDS file from XGBoost version 1.0.0 or earlier. ",
+                      "We strongly ADVISE AGAINST using saveRDS() / readRDS() functions, to ",
+                      "ensure that your model can be read in current and upcoing XGBoost ",
```
hcho3, line 19: Note: Fix typo

trivialfis:
Let me take a look later today.

hcho3:
I'm writing a compatibility test suite now. I just found out that RDS saved from XGBoost 1.1.1 cannot be read into XGBoost 1.2.0 (compiled from the master branch):
```
terminate called after throwing an instance of 'std::out_of_range'                    
  what():  map::at
```

Will investigate.

hcho3:
https://github.com/dmlc/xgboost/commit/cd3d14ad0e4e851cc855d045db1800855347cd76 (PR #5624) added `hist_maker_param_` field to `QuantileHistMaker`, and `QuantileHistMaker::LoadConfig()` is attempting to load the `hist_maker_param_` field. Since 1.1.1 didn't have this field, it results in an error.

Do we promise backward compatibility for internal configuration? If not, should it be saved by `saveRDS()`?

trivialfis:
I don't think it's possible to keep the compatibility for serialization format as I wrote about it on the model persistent tutorial.  We can throw some errors for user experience, but I don't think it's worth of trying to promise the compability.  For R this seems to be just an issue in c++ (but not true) because the R layer is not changing frequently.  But for Python layer the biggest issue is actually on Python byte code.  And R has a similar issue, just not exposed because we do not change it often.

hcho3:
Got it. Then let's throw a warning in `QuantileHistMaker::LoadConfig()`. We can treat this issue like Pickle and binary models: we support loading it for now but encourage everyone to move away.

trivialfis:
That's another difficult topic.  We support serialization because it's useful.  For Python, the pickle is used for distributed training, dask returns the finished model from worker to client by pickle.  We can't throw a warning on this.  So my expectation is we throw a warning only when version is different.  Luckily we do save the version field.

hcho3:
> So my expectation is we throw a warning only when version is different.

Precisely. Warnings only when version is mismatched.

hcho3:
TODO. Update R docs to discourage `saveRDS()`.

hcho3:
Added a note about `saveRDS()` in the CRAN doc of XGBoost:

Screen Shot 2020-07-25 at 8 43 48 PM

DavorJ:
Is there any reason why `xgboost::xgb.load.raw()` returns `xgb.Booster.handle` and  `xgboost::xgb.load()` a `xgb.Booster`? I would expect both to return `xgb.Booster`. I ask this because some functions like `xgboost::xgb.importance()` accept only `xgb.Booster`.

Furthermore, I use `saveRDS`/`serialize`/etc on the xgboost raw binary stream, like so: `saveRDS(xgboost::xgb.save.raw(xgb.model))`. Then loading is done as: `xgboost::xgb.load.raw(loadRDS(...))`. Is this approach deprecated also?

Note: reason I do this is because I store `xgboost::xgb.save.raw()` output in a database, not file.

hcho3:
@DavorJ

> Is there any reason why `xgboost::xgb.load.raw()` returns `xgb.Booster.handle` and `xgboost::xgb.load()` a `xgb.Booster`?

You can convert `xgb.Booster.handle` to `xgb.Booster` by calling `xgb.Booster.complete()`. From [the CRAN XGBoost manual](https://cran.r-project.org/web/packages/xgboost/xgboost.pdf):

> xgb.Booster.complete: Restore missing parts of an incomplete xgb.Booster object.

> Furthermore, I use `saveRDS`/`serialize`/etc on the xgboost raw binary stream, like so: `saveRDS(xgboost::xgb.save.raw(xgb.model))`. Then loading is done as: `xgboost::xgb.load.raw(loadRDS(...))`. Is this approach deprecated also?

No, that approach is fine, as `xgb.save.raw()` and `xgb.load.raw()` are future-proof. I updated the instructions in the [the CRAN XGBoost manual](https://cran.r-project.org/web/packages/xgboost/xgboost.pdf) to mention your use case:

> Use `xgb.save.raw` to save the XGBoost model as a sequence (vector) of raw bytes in a future-proof manner. Future releases of XGBoost will be able to read the raw bytes and re-construct them corresponding model. To read the model back, use `xgb.load.raw`. The `xgb.save.raw` function is useful if you’d like to persist the XGBoost model as part of another R object.

PR #5948, opened 2020-07-27
[BLOCKING] Fix evaluate root split.

trivialfis:
Run cuda-memcheck on CI.  Might take some time to complete.


trivialfis:
@hcho3 I reverted the change of adding cuda memcheck for now.  We can come back to this after the release.

PR #5949, opened 2020-07-28
Initial support for one hot split.

trivialfis:
This PR aims to have a working pipeline for categorical data, but the support is very limited at current form.  To run tests, one needs to:

- Use Python interface.
- Use `gbtree`.
- Use `gpu_hist` tree method.  Other tree methods are coming.
- Use `DMatrix`, `DeviceQuantileDMatrix` is not yet supported.
- Specify `enable_categorical` for `DMatrix`.
- Do **not** use weights.
- Use pandas with categorical feature type.
- Set `gpu_predictor` explicitly.
- Use JSON for model persistent.
- Specify `enable_experimental_json_serialization` even if you don't use pickle.

## Limitations
- The support is limited to 1 vs rest categorical split.  Other categorical specializations are coming.
- There's no mapping between categorical value and histogram bin.  So memory usage might be sub-optimal when categories are sparse.

### DIFF
--- a/include/xgboost/feature_map.h
+++ b/include/xgboost/feature_map.h
@@ -82,7 +82,9 @@ class FeatureMap {
     if (!strcmp("q", tname)) return kQuantitive;
     if (!strcmp("int", tname)) return kInteger;
     if (!strcmp("float", tname)) return kFloat;
-    LOG(FATAL) << "unknown feature type, use i for indicator and q for quantity";
+    if (!strcmp("categorical", tname)) return kInteger;
+    LOG(FATAL) << "unknown feature type, use i for indicator, q for quantity "
+                  "and categorical for categorical split.";
     return kIndicator;
   }
   /*! \brief name of the feature */
--- a/python-package/xgboost/data.py
+++ b/python-package/xgboost/data.py
@@ -175,20 +175,24 @@ _pandas_dtype_mapper = {
 }
 
 
-def _transform_pandas_df(data, feature_names=None, feature_types=None,
+def _transform_pandas_df(data, enable_categorical,
+                         feature_names=None, feature_types=None,
                          meta=None, meta_type=None):
     from pandas import MultiIndex, Int64Index
-    from pandas.api.types import is_sparse
+    from pandas.api.types import is_sparse, is_categorical
+
     data_dtypes = data.dtypes
-    if not all(dtype.name in _pandas_dtype_mapper or is_sparse(dtype)
+    if not all(dtype.name in _pandas_dtype_mapper or is_sparse(dtype) or
+               (is_categorical(dtype) and enable_categorical)
                for dtype in data_dtypes):
         bad_fields = [
             str(data.columns[i]) for i, dtype in enumerate(data_dtypes)
             if dtype.name not in _pandas_dtype_mapper
         ]
 
-        msg = """DataFrame.dtypes for data must be int, float or bool.
-                Did not expect the data types in fields """
+        msg = """DataFrame.dtypes for data must be int, float, bool or categorical.  When
+                categorical type is supplied, DMatrix parameter
+                `enable_categorical` must be set to `True`."""
         raise ValueError(msg + ', '.join(bad_fields))
 
     if feature_names is None and meta is None:
@@ -207,6 +211,8 @@ def _transform_pandas_df(data, feature_names=None, feature_types=None,
             if is_sparse(dtype):
                 feature_types.append(_pandas_dtype_mapper[
                     dtype.subtype.name])
+            elif is_categorical(dtype) and enable_categorical:
+                feature_types.append('categorical')
             else:
                 feature_types.append(_pandas_dtype_mapper[dtype.name])
 
@@ -215,15 +221,21 @@ def _transform_pandas_df(data, feature_names=None, feature_types=None,
             'DataFrame for {meta} cannot have multiple columns'.format(
                 meta=meta))
 
-    dtype = meta_type if meta_type else np.float32
     data = np.ascontiguousarray(data.values, dtype=dtype)
+    dtype = meta_type if meta_type else np.float32
+    try:
+        data = data.values.astype(dtype)
+    except ValueError as e:
+        raise ValueError('Data must be convertable to float, even ' +
+                         'for categorical data.') from e
 
     return data, feature_names, feature_types
 
 
-def _from_pandas_df(data, missing, nthread, feature_names, feature_types):
+def _from_pandas_df(data, enable_categorical, missing, nthread,
+                    feature_names, feature_types):
     data, feature_names, feature_types = _transform_pandas_df(
-        data, feature_names, feature_types)
+        data, enable_categorical, feature_names, feature_types)
     return _from_numpy_array(data, missing, nthread, feature_names,
                              feature_types)
 
@@ -498,7 +510,8 @@ def _has_array_protocol(data):
 
 
 def dispatch_data_backend(data, missing, threads,
-                          feature_names, feature_types):
+                          feature_names, feature_types,
+                          enable_categorical=False):
     '''Dispatch data for DMatrix.'''
     if _is_scipy_csr(data):
         return _from_scipy_csr(data, missing, feature_names, feature_types)
@@ -514,7 +527,7 @@ def dispatch_data_backend(data, missing, threads,
     if _is_tuple(data):
         return _from_tuple(data, missing, feature_names, feature_types)
     if _is_pandas_df(data):
-        return _from_pandas_df(data, missing, threads,
+        return _from_pandas_df(data, enable_categorical, missing, threads,
                                feature_names, feature_types)
     if _is_pandas_series(data):
         return _from_pandas_series(data, missing, threads, feature_names,
@@ -644,7 +657,8 @@ def dispatch_meta_backend(matrix: DMatrix, data, name: str, dtype: str = None):
         _meta_from_numpy(data, name, dtype, handle)
         return
     if _is_pandas_df(data):
-        data, _, _ = _transform_pandas_df(data, meta=name, meta_type=dtype)
+        data, _, _ = _transform_pandas_df(data, False, meta=name,
+                                          meta_type=dtype)
         _meta_from_numpy(data, name, dtype, handle)
         return
     if _is_pandas_series(data):
--- a/src/tree/updater_gpu_hist.cu
+++ b/src/tree/updater_gpu_hist.cu
@@ -19,7 +19,9 @@
 #include "../common/io.h"
 #include "../common/device_helpers.cuh"
 #include "../common/hist_util.h"
+#include "../common/bitfield.h"
 #include "../common/timer.h"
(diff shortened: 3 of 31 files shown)


include/xgboost/feature_map.h
```diff
@@ -82,6 +82,7 @@ class FeatureMap {
     if (!strcmp("q", tname)) return kQuantitive;
     if (!strcmp("int", tname)) return kInteger;
     if (!strcmp("float", tname)) return kFloat;
+    if (!strcmp("categorical", tname)) return kInteger;
     LOG(FATAL) << "unknown feature type, use i for indicator and q for quantity";
```
trivialfis, line 86: Fix this for cat.

python-package/xgboost/data.py
```diff
@@ -209,14 +216,19 @@ def _transform_pandas_df(data, feature_names=None, feature_types=None,
...
 
     return data, feature_names, feature_types
 
 
-def _from_pandas_df(data, missing, nthread, feature_names, feature_types):
+def _from_pandas_df(data, allow_cat, missing, nthread,
+                    feature_names, feature_types):
     data, feature_names, feature_types = _transform_pandas_df(
-        data, feature_names, feature_types)
+        data, allow_cat, feature_names, feature_types)
```
trivialfis, line 231: Make the parameter name consistent.

src/common/categorical.h
```diff
@@ -0,0 +1,34 @@
...
+XGBOOST_DEVICE bst_cat_t AsCat(T const& v) {
+  return static_cast<bst_cat_t>(v);
+}
+
+inline XGBOOST_DEVICE bool IsCat(Span<FeatureType const> ft, bst_feature_t fidx) {
+  return !ft.empty() && ft[fidx] == FeatureType::kCategorical;
+}
+
+template <typename Comp>
+inline XGBOOST_DEVICE bool GoLeft(float l, float r, bool is_cat, Comp comp) {
```
trivialfis, line 25: Add doc.

src/common/quantile.cu
```diff
@@ -326,10 +332,11 @@ void SketchContainer::Prune(size_t to) {
                                 d_columns_ptr_in[column_id + 1] -
                                     d_columns_ptr_in[column_id]);
     idx -= d_columns_ptr_out[column_id];
+    auto is_cat = !ft.empty() && ft[column_id] == FeatureType::kCategorical;
```
trivialfis, line 335: Use utility function.

src/tree/gpu_hist/histogram.cu
```diff
@@ -203,6 +203,22 @@ void BuildGradientHistogram(EllpackDeviceAccessor const& matrix,
   grid_size = common::DivRoundUp(grid_size,
       common::DivRoundUp(num_groups, num_groups_threshold));
 
+  // thrust::host_vector<int32_t> index(matrix.row_stride * matrix.n_rows);
```
trivialfis, line 206: Remove commented code.

src/tree/updater_gpu_hist.cu
```diff
@@ -568,11 +591,24 @@ struct GPUHistMakerDevice {
     auto right_weight = node_value_constraints[candidate.nid].CalcWeight(
                             param, candidate.split.right_sum) *
                         param.learning_rate;
-    tree.ExpandNode(candidate.nid, candidate.split.findex,
-                    candidate.split.fvalue, candidate.split.dir == kLeftDir,
-                    base_weight, left_weight, right_weight,
-                    candidate.split.loss_chg, parent_sum.GetHess(),
-                     candidate.split.left_sum.GetHess(), candidate.split.right_sum.GetHess());
+    auto is_cat = candidate.split.is_cat;
+    // LOG(WARNING) << "Candidate: " << candidate << std::endl;
```
trivialfis, line 595: Remove.

tests/cpp/tree/gpu_hist/test_evaluate_splits.cu
```diff
@@ -45,6 +45,50 @@ TEST(GpuHist, EvaluateSingleSplit) {
                   parent_sum.GetHess());
 }
 
+TEST(GpuHist, EvaluateCategoricalSplit) {
```
trivialfis, line 48: More complicated tests.

tests/cpp/predictor/test_predictor.cc
```diff
@@ -134,4 +134,43 @@ void TestInplacePrediction(dmlc::any x, std::string predictor,
   learner->SetParam("gpu_id", "-1");
   learner->Configure();
 }
+
+void TestCategoricalPrediction(std::string name) {
```
trivialfis, line 138: More complicated tests.

tests/python/testing.py
```diff
@@ -40,6 +40,17 @@ def no_dt():
             'reason': 'Datatable is not installed.'}
 
 
+def no_cupy():
+    reason = 'Matplotlib is not installed.'
```
trivialfis, line 44: Fix this.

trivialfis:
@hcho3 Right now I'm reusing the split cond in `RegTree` for categorical split.  But once we go beyond one hot split, the split condition can be a vector containing multiple categories.  We need to come up with a better structure with a JSON format.  We need to have more discussion around this otherwise the model format might subject to change.

hcho3:
@trivialfis We will want to add additional fields to the JSON schema to indicate categorical splits. For example, LightGBM stores `decision_type`, `cat_threshold` and `cat_boundaries` fields. The `decision_type[i]` tells us whether the i-th internal node is a categorical or a numerical split. The `cat_threshold` and `cat_boundaries` together store categories associated with the left child of each categorical split. 
https://github.com/dmlc/treelite/blob/7f01e631da8687189473ad6b177ba0615b19496b/src/frontend/lightgbm.cc#L516-L528

It should not be too difficult to add new fields to the current JSON schema. (I'm assuming that vector with multiple categories will be JSON only and won't support legacy binary serialization.)

trivialfis:
@hcho3 

> will be JSON only and won't support legacy binary serialization

Yes.  But at the same time we can't break the binary format.  For example, we can't add anything to `RegTree::Node`.  Also before this PR is merged, I think we need to set JSON as the default pickle format, as Python interface goes through a serialization at the end of training to release GPU memory.

> It should not be too difficult to add new fields to the current JSON schema.

I think so.  Just make sure that we have considered enough different cases.

hcho3:
Got it. Let's discuss about how to add the necessary fields without touching `RegTree::Node`. One idea is to relegate `RegTree::Node` as an external facing interface and convert from to `RegTree::Node` to the real node structure that includes the new fields. This will slow down deserializing binary models. Another idea is to set the `info_` field of `RegTree::Node` to NaN, to indicate a categorical split, and use the [payload field of NaN](https://stackoverflow.com/questions/19800415/why-does-ieee-754-reserve-so-many-nan-values) to indicate where the extra information can be looked up.

tests/cpp/common/test_hist_util.h
```diff
@@ -170,7 +153,9 @@ inline void ValidateColumn(const HistogramCuts& cuts, int column_idx,
     // Each value should get its own bin
     int i = 0;
     for (auto v : unique) {
-      ASSERT_EQ(cuts.SearchBin(v, column_idx), cuts.Ptrs()[column_idx] + i);
+      // Categorical data doesn't discard the first unique value (min value), so bin is
+      // shifted to right by 1.
+      EXPECT_EQ(cuts.SearchBin(v, column_idx), cuts.Ptrs()[column_idx] + i + (is_cat ? 1 : 0));
```
trivialfis, line 158: Revert this.

trivialfis:
@hcho3 I changed the categories in tree into bitfield.

trivialfis:
I reverted change of min value to reduce the size of this PR.

src/tree/updater_gpu_hist.cu
```diff
@@ -568,11 +621,32 @@ struct GPUHistMakerDevice {
...
                         param.learning_rate;
-    tree.ExpandNode(candidate.nid, candidate.split.findex,
-                    candidate.split.fvalue, candidate.split.dir == kLeftDir,
-                    base_weight, left_weight, right_weight,
-                    candidate.split.loss_chg, parent_sum.GetHess(),
-                     candidate.split.left_sum.GetHess(), candidate.split.right_sum.GetHess());
+
+    auto is_cat = candidate.split.is_cat;
+    if (is_cat) {
+      auto cat = common::AsCat(candidate.split.fvalue);
```
hcho3, line 627: Note to myself: in one-hot encoded setting, there is only one matching category in every categorical split. However, the `split_categories_` structure can later store multiple matching categories per split.

hcho3: Another reminder to myself: Treelite must support JSON format of XGBoost.

src/tree/tree_model.cc
```diff
@@ -777,8 +862,17 @@ void RegTree::SaveModel(Json* p_out) const {
...
   std::vector<Json> conds(n_nodes);
   std::vector<Json> default_left(n_nodes);
+  std::vector<Json> split_type(n_nodes);
 
-  for (int32_t i = 0; i < n_nodes; ++i) {
+  std::vector<Json> cat_seg_begin(n_nodes);
+  std::vector<Json> cat_seg_size(n_nodes);
+  std::vector<Json> categories(n_nodes);
+
+  auto& self = *this;
```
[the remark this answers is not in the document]
trivialfis: > Why not use the customary this pointer directly? 

I think at some point I use `operator[]` then removed the usage.  The model IO is revised in https://github.com/dmlc/xgboost/pull/6028 and stores actual categories instead of bitset.  I will address related comments in there.  Rebasing this PR becomes quite difficult.

src/tree/tree_model.cc
```diff
@@ -792,6 +886,20 @@ void RegTree::SaveModel(Json* p_out) const {
...
     default_left[i] = n.DefaultLeft();
+
+    if (self.GetSplitTypes().size() == static_cast<size_t>(n_nodes)) {
+      CHECK_EQ(self.split_categories_segments_.size(), param.num_nodes);
+      split_type[i] = static_cast<I>(self.NodeSplitType(i));
+      auto beg = self.split_categories_segments_.at(i).beg;
+      auto size = self.split_categories_segments_.at(i).size;
+      cat_seg_begin[i] = static_cast<I>(beg);
+      cat_seg_size[i] = static_cast<I>(size);
+      auto node_categories = self.GetSplitCategories().subspan(beg, size);
```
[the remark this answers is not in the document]
trivialfis: Will clean it up.

include/xgboost/span.h
```diff
@@ -515,6 +527,7 @@ class Span {
   }
 
   XGBOOST_DEVICE reference operator[](index_type _idx) const {
+    SPAN_LT(_idx, size());
```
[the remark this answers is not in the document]
trivialfis: Removed in https://github.com/dmlc/xgboost/pull/6028 .

trivialfis:
All merged.

Issue #5952, opened 2020-07-28
Custom multiclass eval metric receives transformed predictions


guyko81:
Hi,

I'm having issues with custom multiclass objective function's eval metric, as the eval always receives transformed predictions, e.g. the argmax value, not the raw predictions.
It was stated in issue #4288 that we can use the example from [xgboost/demo/guide-python/custom_softmax.py](xgboost/demo/guide-python/custom_softmax.py) but to me it's not clear how to use custom predict function for eval metric. 

If you can show how to use it correctly I would appreciate much. 

Thanks

PR #5954, opened 2020-07-29
[Breaking] Fix custom metric for multi output.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/5952 .

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -154,9 +154,9 @@ LearnerModelParam::LearnerModelParam(
 
 struct LearnerTrainParam : public XGBoostParameter<LearnerTrainParam> {
   // data split mode, can be row, col, or none.
-  DataSplitMode dsplit;
+  DataSplitMode dsplit {DataSplitMode::kAuto};
   // flag to disable default metric
-  bool disable_default_eval_metric;
+  bool disable_default_eval_metric {false};
   // FIXME(trivialfis): The following parameters belong to model itself, but can be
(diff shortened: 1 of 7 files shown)


trivialfis:
Removed blocking.  This will be a breaking change.

hcho3:
@trivialfis So this is to be consistent with #5564? Can we include it in this release?

src/learner.cc
```diff
@@ -156,7 +156,7 @@ struct LearnerTrainParam : public XGBoostParameter<LearnerTrainParam> {
   // data split mode, can be row, col, or none.
   DataSplitMode dsplit;
   // flag to disable default metric
-  int disable_default_eval_metric;
+  bool disable_default_eval_metric;
```
hcho3, line 159: The parameter doc should also be updated.

trivialfis:
>  So this is to be consistent with #5564?

Yes.

> Can we include it in this release? 

Yes, I want to include it.  Just not sure about how others feel with a last minute breaking change.

> The release note can indicate that it's breaking.

Will mark it.

hcho3:
> Just not sure about how others feel with a last minute breaking change.

In this case, we already have a precedent (#5564) where custom objectives gets the margin scores. It will be worthwhile to achieve the same behavior in custom metrics too. Also, this PR now adds unit tests for custom metrics with multiple outputs.

trivialfis:
@hcho3 Thanks, remarked as blocking.

PR #5961, opened 2020-07-30
Update XGBoost + Dask overview documentation

jrbourbeau:
This PR contains a few documentation updates for the XGBoost Dask tutorial. In particular, this adds `import`s at the top of the first code snippet (useful for those who are new to Dask and may not know where `LocalCluster` comes from) as well as a few minor updates to the prose on this page. Most of the changes here are subjective and I'm happy to roll back changes as needed.


jrbourbeau:
Thanks for the speedy review!

PR #5964, opened 2020-07-31
[R] Provide better guidance for persisting XGBoost model

hcho3:
Update the CRAN manual of XGBoost to provide clear guidance about persisting XGBoost model. 

[Rendered CRAN manual](https://github.com/dmlc/xgboost/files/5004447/bab841fb-b8b4-4c21-abbd-eccaa6e10de8.pdf)

cc @leedrake5


R-package/R/utils.R
```diff
@@ -308,18 +308,60 @@ xgb.createFolds <- function(y, k = 10)
...
+#' xgb.save(bst, 'xgb.model')
+#' bst2 <- xgb.load('xgb.model')
+#'
+#' # Save as a stand-alone file (JSON); load it with xgb.load()
+#' xgb.save(bst, 'xgb.model.json')
+#' bst2 <- xgb.load('xgb.model.json')
+#'
+#' # Save as a raw byte vector; load it with xgb.load.raw()
+#' xgb_bytes <- xgb.save.raw(bst)
+#' bst2 <- xgb.load.raw(xgb_bytes)
```
hcho3, line 351: Note: Currently `xgb.load.raw()` uses the binary serialization method, but we can transparently replace it with the JSON serialization method in the near future.

R-package/R/utils.R
```diff
@@ -308,18 +308,60 @@ xgb.createFolds <- function(y, k = 10)
...
+#' \code{\link{xgb.load}}.
+#'
+#' Use \code{\link{xgb.save.raw}} to save the XGBoost model as a sequence (vector) of raw bytes
+#' in a future-proof manner. Future releases of XGBoost will be able to read the raw bytes and
+#' re-construct the corresponding model. To read the model back, use \code{\link{xgb.load.raw}}.
+#' The \code{\link{xgb.save.raw}} function is useful if you'd like to persist the XGBoost model
+#' as part of another R object.
+#'
+#' For more details and explanation about model persistence and archival, consult the page
 #' \url{https://xgboost.readthedocs.io/en/latest/tutorials/saving_model.html}.
```
hcho3, line 334: Here is the link.

hcho3: Do you want me to edit saving_model.rst ?

trivialfis: Yes.

hcho3: Done.

hcho3:
Will wait for @leedrake5's review.

leedrake5:
I'm restricting comments to the compatibility note, because I think the other functions are self explanatory. I would recommend adding serialize to the compatibility note since it is a closely related function. Recomended text: 

> Use xgb.save to save the XGBoost model as a stand-alone file. You may opt into the JSON format by specifying the JSON extension. To read the model back, use xgb.load.
> Use xgb.save.raw to save the XGBoost model as a sequence (vector) of raw bytes in a future- proof manner. Future releases of XGBoost will be able to read the raw bytes and re-construct the corresponding model. To read the model back, use xgb.load.raw. The xgb.save.raw function is useful if you’d like to persist the XGBoost model as part of another R object.
> Use xgb.serialize to save the XGBoost model as a sequence (vector) of raw bytes in a machine-specific, but not future-proof manner. Future releases of XGBoost may not be able to read this configuration, but it will record machine training parameters such as the number of GPUs used for compatibility with an instance of XGBoost on a particular device. 


Feel free to edit or disregard - it is only a minor change.

hcho3:
@leedrake5 I do not want to say "Use xgb.serialize ..." in the compatibility note because it may give an impression that it's an acceptable form of long-term archival, when it's not. Instead, I'll add a paragraph as follows:

> Use xgb.save to save the XGBoost model as a stand-alone file. You may opt into the JSON format by specifying the JSON extension. To read the model back, use xgb.load.
> Use xgb.save.raw to save the XGBoost model as a sequence (vector) of raw bytes in a future- proof manner. Future releases of XGBoost will be able to read the raw bytes and re-construct the corresponding model. To read the model back, use xgb.load.raw. The xgb.save.raw function is useful if you’d like to persist the XGBoost model as part of another R object.
> **Note: Do not use the xgb.serialize function to store models long-term. It persists not only the model but also internal configurations and parameters, and its format is not stable across multiple XGBoost versions.**

hcho3:
@trivialfis I'd like to merge this, if you are okay with the new paragraph about `xgb.serialize()`.

PR #5969, opened 2020-08-01
Fix missing data warning.

trivialfis:
Not critical as the missing field is converted to nan in `DMatrix.__init__`.  Just fixing the correctness of the helper function and making it self-contained.

### DIFF
--- a/tests/python/test_dmatrix.py
+++ b/tests/python/test_dmatrix.py
@@ -22,7 +22,7 @@ class TestDMatrix(unittest.TestCase):
             data._warn_unused_missing('uri', None)
             data._warn_unused_missing('uri', np.nan)
 
-        assert len(record) == 0
+            assert len(record) == 0
 
         with pytest.warns(None) as record:
             x = rng.randn(10, 10)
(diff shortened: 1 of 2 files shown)


tests/python/test_dmatrix.py
```diff
@@ -12,6 +13,29 @@
...
+    def test_warn_missing(self):
+        from xgboost import data
+        with pytest.warns(UserWarning):
+            data._warn_unused_missing('uri', 4)
+
+        with pytest.warns(None) as record:
+            data._warn_unused_missing('uri', None)
+            data._warn_unused_missing('uri', np.nan)
+
+        assert len(record) == 0
```
hcho3, line 25: Which `record` is this referring to?

trivialfis: @hcho3 At line 21.

hcho3: Is it legal to refer to a `with` variable past its scope? It's as if:
```
with open(..) as f:
   ...
print('foobar', file=f)
```
which seems undefined.

trivialfis: Let me move that inside.

trivialfis: Done.

trivialfis:
@hcho3 Thanks for the speedy review.

PR #5972, opened 2020-08-03
[WIP][jvm-packages]add trainDistributed for RDD[Watches]

wbo4958:


hcho3:
Please provide some context. Is this related to #5950 ?

wbo4958:
This PR is related to https://github.com/dmlc/xgboost/pull/5950

FelixYBW:
Any update of this PR?

"XGBoost native may need to do some sync between different workers (right now, the sync is not enabled, but I suppose it will be enabled recently" why do we need the sync between workers?

Spark3.0 added columnar API, so we can use the columnar batch to exchange data between spark and ML/DL libraries. We plan to add this support as well.

PR #5974, opened 2020-08-04
Force tree order in JSON.

trivialfis:
It's not a bug.  Right now the order is implicitly enforced, this PR makes it explicit.  Marked as blocking because with explicit behaviour, we can be sure there's no chance for mistake.

* Put in omp parallel.  On large number of trees this is about 20% faster, not very impressive so I can revert it if needed.

### DIFF
--- a/src/gbm/gbtree_model.cc
+++ b/src/gbm/gbtree_model.cc
@@ -49,7 +49,7 @@ void GBTreeModel::SaveModel(Json* p_out) const {
     auto const& tree = trees[t];
     Json tree_json{Object()};
     tree->SaveModel(&tree_json);
-    tree_json["id"] = Integer(t);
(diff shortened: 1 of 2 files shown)


trivialfis:
@hcho3 I can remove the blocking tag if you think it's risky.

hcho3:
@trivialfis Is the `Json` class tested for concurrent accesses? I'm a bit concerned about the use of OpenMP here. We could remove OpenMP from this PR to reduce the risk.

trivialfis:
>  Is the Json class tested for concurrent accesses

No.  But each JSON object is created as a local variable to the thread (inside for loop).

> We could remove OpenMP from this PR to reduce the risk.

Sounds good.  Let me remove it.

trivialfis:
@hcho3 OMP loop is removed.

PR #5978, opened 2020-08-04
Rabit update.

trivialfis:
https://github.com/dmlc/rabit/pull/143

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -381,7 +381,6 @@ object XGBoost extends Serializable {
     val attempt = TaskContext.get().attemptNumber.toString
     rabitEnv.put("DMLC_TASK_ID", taskId)
     rabitEnv.put("DMLC_NUM_ATTEMPT", attempt)
-    rabitEnv.put("DMLC_WORKER_STOP_PROCESS_ON_ERROR", "false")
     val numRounds = xgbExecutionParam.numRounds
     val makeCheckpoint = xgbExecutionParam.checkpointParam.isDefined && taskId.toInt == 0
     try {
@@ -997,4 +996,3 @@ private[spark] class LabeledPointGroupIterator(base: Iterator[XGBLabeledPoint])
     group
   }
 }
-
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
@@ -303,8 +303,7 @@ class XGBoostClassificationModel private[ml](
         private val batchIterImpl = rowIterator.grouped($(inferBatchSize)).flatMap { batchRow =>
           if (batchCnt == 0) {
             val rabitEnv = Array(
-              "DMLC_TASK_ID" -> TaskContext.getPartitionId().toString,
-              "DMLC_WORKER_STOP_PROCESS_ON_ERROR" -> "false").toMap
+              "DMLC_TASK_ID" -> TaskContext.getPartitionId().toString).toMap
             Rabit.init(rabitEnv.asJava)
           }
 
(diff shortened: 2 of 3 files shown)


jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -381,7 +381,6 @@ object XGBoost extends Serializable {
     val attempt = TaskContext.get().attemptNumber.toString
     rabitEnv.put("DMLC_TASK_ID", taskId)
     rabitEnv.put("DMLC_NUM_ATTEMPT", attempt)
-    rabitEnv.put("DMLC_WORKER_STOP_PROCESS_ON_ERROR", "false")
```
CodingCat, line 384: well, as I commented in https://github.com/dmlc/rabit/pull/143, this is something we do need to implement current error handling strategy

regarding https://github.com/dmlc/xgboost/issues/4826 , what's your current plan? I think we could replace "kill SparkContext" with "kill the corresponding job" in the case of a task failure in JVM layer

trivialfis: @CodingCat It's still on early planing stage and I'm obviously not the right person to talk about spark, so my thought will be largely based on experience with dask.  For spark specific things I usually just forward to @wbo4958 ;-).  I have a few plans that I haven't decided on which one to go yet.  They can be roughly listed as follow:

- We drop the support on single node recovery, and relies on Spark to recover from training failures.  My goal is to have XGBoost fail gracefully without interrupting others.  On dask [a Python exception should do it.](https://distributed.dask.org/en/latest/resilience.html#user-code-failures)  If we do this I want to replace rabit with more matured solutions like MPI.

> "kill the corresponding job" in the case of a task failure in JVM layer

Sounds good to me.  Whatever that doesn't interrupt the others.

- We continue the support on single node recovery, but instead we uses call sequence number as cache key.  My major concern on this is multi-threaded application.  It's not common for users to create DMatrix on their own with the dask interface, but I haven't seen people try to access DMatrix by themselves to bypass dask data loading procedure to avoid extra memory overhead and data balancing.  If they do it  asynchronously then we need serious tests on rabit's thread safety guarantee (which doesn't exist).

trivialfis:
@CodingCat Surprise, all tests passed ...

From https://github.com/dmlc/rabit/pull/143#issuecomment-668713796:

> if you remove this, the test may not be runnable

Not runnable you mean being skipped, or might fail?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
trivialfis: Either way, the `exit(1)` call has to go, as we need a proper exception.

CodingCat:
> @CodingCat Surprise, all tests passed ...
> 
> From [dmlc/rabit#143 (comment)](https://github.com/dmlc/rabit/pull/143#issuecomment-668713796):
> 
> > if you remove this, the test may not be runnable
> 
> By not runnable, did you mean the test might be skipped, or might fail?

yeah, it's a surprise , I remember the reason I add this is to handle wrong metrics defined in Spark layer (https://github.com/dmlc/rabit/pull/97)

by default (at that moment), rabit will stop the process, and spark unit test just got interrupted....so I add this flag to completely delegate post action to spark layer


oh, I just found your PR in rabit side is to change the default behavior of rabit to what I "configured to" in Spark side, so it should pass all tests then

trivialfis:
@CodingCat I'm not sure how important the single point model recovery it's on the CPU side.  As commented in https://github.com/dmlc/xgboost/pull/5978#discussion_r465225297 , if we were to continue investing on the model recovery, then we might want to merge rabit into xgboost.  Otherwise I would like a create a thin layer on top of more matured MPI solution and drop rabit.  A good candidate would be [UCX](https://github.com/openucx/ucx)

CodingCat:
> @CodingCat I'm not sure how important the single point model recovery it's on the CPU side. Right now it's not that high a priority on GPU, which may change in the future with growing size of clusters. As commented in [#5978 (comment)](https://github.com/dmlc/xgboost/pull/5978#discussion_r465225297) , if we were to continue investing on the model recovery, then we might want to merge rabit into xgboost. Otherwise I would like a create a thin layer on top of more matured MPI solution and drop rabit. A good candidate would be [UCX](https://github.com/openucx/ucx)

I don't think single point recovery is critical, we are not using it, and other ML/DL frameworks all adopted checkpoint-based fault handling mechanism....

regarding communication layer replacement, +1, but we may want to make them co-exist for 1-2 versions? just in case the bug exists and we have a fallback? (some experience in Spark community is that they have the original akka and self-made RPC framework co-exist for 1 version but make self-made one as default, then in the next one, drop akka completely)

trivialfis:
Sure.  There are 2 ways we can go with replacing rabit.

- Just drop rabit, start the communication layer in XGBoost from scratch.
- Merge rabit into XGBoost and uses its backend facility.

It's not a trivial task, as large portion of our codebase is built around rabit.

PR #5996, opened 2020-08-10
[jvm-packages] Fix deterministic partitioning with dataset containing Double.NaN

Totoketchup:
The functions featureValueOfSparseVector or featureValueOfDenseVector could return a Float.NaN if the input vector was containing any missing values. This would make fail the partition key computation and most of the rows would end up in the same partition and therefore leading to memory issues on one of the executor. We fixed this by avoid returning a NaN and simply use the row HashCode in this case.
We added a test to ensure that the repartition is indeed now uniform on input dataset containing missing values by checking that the partitions size variance is below a certain threshold.

Another remark about the computation of the key:
- If the values contained in your DataFrame are only below let's say 1e-4 / 1e-5, then this computation:
`rowHashCode.toLong + featureValue` is equivalent to `rowHashCode.toLong`, wouldn't it better to make it more random by computing `featureValue.toString().hashCode()` ? It would even fix what I did here too. WDYT ?

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
@@ -103,7 +103,8 @@ object DataUtils extends Serializable {
       case sparseVector: SparseVector =>
         featureValueOfSparseVector(rowHashCode, sparseVector)
     }
-    math.abs((rowHashCode.toLong + featureValue).toString.hashCode % numPartitions)
+    val nonNaNFeatureValue = if (featureValue.isNaN) { 0.0f } else { featureValue }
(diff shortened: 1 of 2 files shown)


Totoketchup:
Hello @CodingCat, I see that you are the one that designed the deterministic repartitioning for XGboost looking at  https://github.com/dmlc/xgboost/issues/4786.

I have a questiion about the computation of the partition key:

Why `math.abs(row.hashCode) % numWorkers` would not be enough to compute the key ? Looking at the implementation of row.hashCode (https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/Row.scala) we can see that it already uses the values inside the row to compute this hash. What does it bring to get a feature from the Row and add it to the computation of the key ?

hcho3:
I'm looking at the failing tests now: #6014

hcho3:
Restarting the tests.

PR #6011, opened 2020-08-13
[CI] Cancel builds on subsequent pushes

hcho3:
If a branch or a pull requests receives successive pushes in a short duration of time, Jenkins may be running multiple builds. Only the latest build should be carried out and the others cancelled, to conserve CI resources.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,6 +40,7 @@ pipeline {
         script {
           checkoutSrcs()
           commit_id = "${GIT_COMMIT}"
+          cancelPreviousBuilds()
         }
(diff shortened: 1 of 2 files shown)


hcho3:
I just trying pushing two commits in a row, and the fix is working. The previous build was cancelled with message
```
Superseded by xgboost/PR-6011#4
```

hcho3:
I also enabled the equivalent setting in AppVeyor as well. The feature is called ["Rolling Builds"](https://www.appveyor.com/docs/build-configuration/#rolling-builds).

PR #6013, opened 2020-08-13
[Doc] Add dtreeviz as a showcase example of integration with 3rd-party software

hcho3:
https://github.com/parrt/dtreeviz


hcho3:
@trivialfis They rely on `tree_to_dataframe()` and `predict()` methods, which we already have tests for.

trivialfis:
Yup, but having a simple demo that can be tested on our side seems really nice and assuring during development.  It's not a heavy dependency anyway.  I can approve this PR and track it in a new issue if that's what you want.

hcho3:
Sure, let's open a new issue for it.

trivialfis:
https://github.com/dmlc/xgboost/issues/6015

PR #6019, opened 2020-08-17
[jvm-packages] cancel job instead of killing SparkContext

wbo4958:
This PR changes the default behavior that kills SparkContext. Instead, This PR
cancels jobs when coming across task failed. That means the SparkContext is
still alive even some exceptions happen.

This PR is regarding https://github.com/dmlc/xgboost/issues/4826 and https://github.com/dmlc/xgboost/pull/5978

@CodingCat @trivialfis @hcho3 Could you help to review this PR?

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -626,9 +626,8 @@ object XGBoost extends Serializable {
       (booster, metrics)
     } catch {
       case t: Throwable =>
-        // if the job was aborted due to an exception
+        // if the job was aborted due to an exception, just throw the exception
         logger.error("the job was aborted due to ", t)
-        trainingData.sparkContext.stop()
         throw t
     } finally {
       uncacheTrainingData(xgbExecParams.cacheTrainingSet, transformedTrainingData)
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -632,7 +632,7 @@ object XGBoost extends Serializable {
       (booster, metrics)
     } catch {
       case t: Throwable =>
-        // if the job was aborted due to an exception, just throw the exception
+        // if the job was aborted due to an exception
         logger.error("the job was aborted due to ", t)
         if (xgbExecParams.killSparkContext) {
           trainingData.sparkContext.stop()
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -76,7 +76,8 @@ private[this] case class XGBoostExecutionParams(
     earlyStoppingParams: XGBoostExecutionEarlyStoppingParams,
     cacheTrainingSet: Boolean,
     treeMethod: Option[String],
-    isLocal: Boolean) {
+    isLocal: Boolean,
+    killSparkContext: Boolean) {
 
   private var rawParamMap: Map[String, Any] = _
 
@@ -220,6 +221,9 @@ private[this] class XGBoostExecutionParamsFactory(rawParams: Map[String, Any], s
     val cacheTrainingSet = overridedParams.getOrElse("cache_training_set", false)
       .asInstanceOf[Boolean]
 
+    val killSparkContext = overridedParams.getOrElse("kill_spark_context", true)
+      .asInstanceOf[Boolean]
+
     val xgbExecParam = XGBoostExecutionParams(nWorkers, round, useExternalMemory, obj, eval,
       missing, allowNonZeroForMissing, trackerConf,
       timeoutRequestWorkers,
@@ -228,7 +232,8 @@ private[this] class XGBoostExecutionParamsFactory(rawParams: Map[String, Any], s
       xgbExecEarlyStoppingParams,
       cacheTrainingSet,
       treeMethod,
-      isLocal)
+      isLocal,
+      killSparkContext)
     xgbExecParam.setRawParamMap(overridedParams)
     xgbExecParam
   }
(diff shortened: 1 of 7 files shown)


jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
```diff
@@ -98,31 +98,30 @@ private[spark] class TaskFailedListener extends SparkListener {
     taskEnd.reason match {
       case taskEndReason: TaskFailedReason =>
         logger.error(s"Training Task Failed during XGBoost Training: " +
-            s"$taskEndReason, stopping SparkContext")
-        TaskFailedListener.startedSparkContextKiller()
+            s"$taskEndReason, cancelling all jobs")
+        TaskFailedListener.cancelAllJobs()
```
CodingCat, line 102: 1. I don't think we should change the default behavior of killing sparkcontext, over the year, a lot of fault recovery strategy in prod env have been built based on killing context behavior...instead, we should provide both behavior. Even more, cancel the job is just for a small percentage of customers who train multiple XGB models simultaneously or train various models in the same app.

2. here you called cancelAllJobs, I don't think there is any difference than killing SparkContext, you should call 
image

wbo4958: Thx @CodingCat. for
> I don't think we should change the default behavior of killing sparkcontext, over the year, a lot of fault recovery strategy in prod env have been built based on killing context behavior...instead, we should provide both behavior. Even more, cancel the job is just for a small percentage of customers who train multiple XGB models simultaneously or train various models in the same app.

Yeah, make sense. I will update the change to add a parameter to control for both scenarios.

> here you called cancelAllJobs, I don't think there is any difference than killing SparkContext, you should call

The intent of this PR is to throw an Exception when doing cancelAllJobs. User should try catch
``` scala
try {
val model = xgboostClassifier.fit(dataset)
} catch {
case e: Exception =>
}
```

1. **`cancelAllJobs`**
``` scala
  private[scheduler] def doCancelAllJobs(): Unit = {
    // Cancel all running jobs.
    runningStages.map(_.firstJobId).foreach(handleJobCancellation(_,
      Option("as part of cancellation of all jobs")))
    activeJobs.clear() // These should already be empty by this point,
    jobIdToActiveJob.clear() // but just in case we lost track of some jobs...
  }
```
2. **`cancelJob`**
``` scala
  private[scheduler] def handleJobCancellation(jobId: Int, reason: Option[String]): Unit = {
    if (!jobIdToStageIds.contains(jobId)) {
      logDebug("Trying to cancel unregistered job " + jobId)
    } else {
      failJobAndIndependentStages(
        jobIdToActiveJob(jobId), "Job %d cancelled %s".format(jobId, reason.getOrElse("")))
    }
  }
``` 
From the code, `cancelAllJobs` will finally call the handleJobCancellation just like `cancelJob` does.

wbo4958:
@CodingCat Could you help to review again?

 the new commit introduces a new parameter **kill_spark_context** to control if killing SparkContext or not when training task fails. and adds two test cases for this new feature.

wbo4958:
Hi @CodingCat, Could you help to review it. Many thx

wbo4958:
@CodingCat I would like to say you are indeed a Spark Expert. I didn't realize we can submit jobs simultaneously in a thread.

I will change according to your comments. Thx a lot

hcho3:
The CI has been fixed in the master branch. Rerunning tests now.

wbo4958:
@CodingCat Could you help to review it?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -220,6 +221,9 @@ private[this] class XGBoostExecutionParamsFactory(rawParams: Map[String, Any], s
     val cacheTrainingSet = overridedParams.getOrElse("cache_training_set", false)
       .asInstanceOf[Boolean]
 
+    val killSparkContext = overridedParams.getOrElse("kill_spark_context", true)
```
CodingCat, line 224: kill_spark_context_on_worker_failure?

jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
```diff
@@ -79,7 +83,7 @@ class SparkParallelismTracker(
   def execute[T](body: => T): T = {
     if (timeout <= 0) {
       logger.info("starting training without setting timeout for waiting for resources")
-      body
+      safeExecute(body)
```
CodingCat, line 86: why we have to change to safeExecute()?

jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
```diff
@@ -90,23 +94,58 @@ class SparkParallelismTracker(
...
-            s"$taskEndReason, stopping SparkContext")
-        TaskFailedListener.startedSparkContextKiller()
+            s"$taskEndReason")
+        if (killSparkContext) {
+          logger.error("killing SparkContext")
+          TaskFailedListener.startedSparkContextKiller()
+        } else {
+          val stageId = taskEnd.stageId
+          // find job ids according to stage id and then cancel the job
+          jobIdToStageIds.foreach(t => {
```
CodingCat, line 132:
```
 jobIdToStageIds.foreach { case (jobId, stageIds) => 

...


}

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
wbo4958: Done, Thx for the naming suggestion

jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
wbo4958: The safeExecute wraps TaskFailedListener inside. I don't know why the body was not executed in safeExecute in the previous version, Since it may hang forever if no TaskFailedListener.

jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
wbo4958: Very good suggestion, really thx, Done.

wbo4958:
@hcho3, Hi Phillip, may need your help on this CI failure

hcho3:
Re-triggering the CI. I opened #6074 to keep track of the flaky test.

wbo4958:
Thx @hcho3 , @CodingCat Could you help to review again?

SongBiu:
I wonder if the thread that run "sparkJobThread.start()" start job before another thread that execute the code "sc.addSparkListener, the function "onJobStart" won't be call forever. It's OK?

wbo4958:
> I wonder if the thread that run "sparkJobThread.start()" start job before another thread that execute the code "sc.addSparkListener, the function "onJobStart" won't be call forever. It's OK?

Good finding, I will fix this issue. Thx again

PR #6027, opened 2020-08-18
Make JSON the default full serialization format.

trivialfis:
Saving model is still using binary as default, but serialization like pickle will be default to JSON.


trivialfis:
There's a slight difference in representing deleted node between json and binary model.  For JSON the `IsLeftChild` is meaningless for a deleted node, while for binary the field is preserved.

PR #6033, opened 2020-08-19
[jvm-packages] Clean the way deterministic partitioning is computed

Totoketchup:
We propose to only use the rowHashCode to compute the partitionKey, adding the FeatureValue hashCode might not bring a better load balancing and is makingthe computation slower. MurmurHash3 is known as being quite uniformly distributed and should suffice here.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/DataUtils.scala
@@ -70,50 +70,13 @@ object DataUtils extends Serializable {
     }
   }
 
-  private def featureValueOfDenseVector(rowHashCode: Int, features: DenseVector): Float = {
-    val featureId = {
-      if (rowHashCode > 0) {
-        rowHashCode % features.size
-      } else {
-        // prevent overflow
-        math.abs(rowHashCode + 1) % features.size
-      }
-    }
-    features.values(featureId).toFloat
-  }
-
-  private def featureValueOfSparseVector(rowHashCode: Int, features: SparseVector): Float = {
-    val featureId = {
-      if (rowHashCode > 0) {
-        rowHashCode % features.indices.length
-      } else {
-        // prevent overflow
-        math.abs(rowHashCode + 1) % features.indices.length
-      }
-    }
-    features.values(featureId).toFloat
-  }
-
-  private def calculatePartitionKey(row: Row, numPartitions: Int): Int = {
-    val Row(_, features: Vector, _, _) = row
-    val rowHashCode = row.hashCode()
-    val featureValue = features match {
-      case denseVector: DenseVector =>
-        featureValueOfDenseVector(rowHashCode, denseVector)
-      case sparseVector: SparseVector =>
-        featureValueOfSparseVector(rowHashCode, sparseVector)
-    }
-    val nonNaNFeatureValue = if (featureValue.isNaN) { 0.0f } else { featureValue }
-    math.abs((rowHashCode.toLong + nonNaNFeatureValue).toString.hashCode % numPartitions)
-  }
-
   private def attachPartitionKey(
       row: Row,
       deterministicPartition: Boolean,
       numWorkers: Int,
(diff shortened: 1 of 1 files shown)


hcho3:
@Totoketchup Is this linked to #5996?

Totoketchup:
Yes it is both fixing the problem with Double.NaN and making the deteministic repartitioning faster.Linked to this remark I made: https://github.com/dmlc/xgboost/pull/5996#issuecomment-671256383

CodingCat:
row.hashCode returns the hashCode purely based on the murmur3, we have seen cases that some partition got empty in prod, 

the code here is to add another layer of hash to murmur3, i.e. even murmur3 is the same for two rows, if their particular feature has different value, they will be assigned to different partition 

>  is makingthe computation slower

could you quantify this? how this becomes a bottleneck?

Totoketchup:
Thanks for your answer ;) 

> we have seen cases that some partition got empty in prod.
This looks more like a bug (like we had with NaN values) than something coming from MurMurHash3, no ? (but I might be wrong). By any chance, do still have the dataset ? I could run some analyses to see the load balancing on it. 

What I'll do is that I'll run some load balancing analyses on the test I added last review first to see if we see any big changes in the distribution of rows in the partitions between MurmurHash3 and MurmurHash3 + Feature Value. WDYT ?

> could you quantify this? how this becomes a bottleneck?

For the speed-up I did not run any analyses indeed but I just 'guessed' that accessing the FeatureVector to retrieve a new value, and then computing its hash is adding some computational time. But maybe the impact is really negligible yes. I'll use the test I added to do some benches on 1M lines. This is not really a bottleneck strictly speaking, but if we can gain some time and not lose any load balancing, this is nice.

Totoketchup:
I ran some benchmarks using the test I previously added on `val trainingDF = buildDataFrame(Classification.train)` with 1000 runs and there is a slight speedup in average. But this might be negligible.

```
Current:
  // Max   227,483,731 ns
  // Min    11,876,080 ns
  // Avg    23,015,432 ns

PR:
  // Max    62,906,984 ns
  // Min    11,879,247 ns
  // Avg    18,362,924 ns
```

Totoketchup:
The thing is that for MurmurHash3 we must look at how uniform is the distribution of hashes and not at collisions. For sure we have high level of collision because of the partitioning size and the modulus operator. Even though 2 rows are different with MurmurHash3, they could collide with the modulus operator. Looking at https://stackoverflow.com/questions/11899616/murmurhash-what-is-it, it confirms that MurmurHash has quite a good uniform distribution, so applying the modulus operator on it would still give a good distribution in my opinion.
The scenario you describe where an entire partition is entirely empty is almost impossible.

CodingCat:
> The scenario you describe where an entire partition is entirely empty is almost impossible.

did I misunderstand anything here? You mean what I saw in prod is fake?

Totoketchup:
That's not what I'm saying, deeply sorry if you understood it like this. If you're telling that this happened to you, of course I believe you. What I'm wondering, from what I read about MurmurHash3 (its uniform distribution and its good avalanche effect) what are the number of rows / columns you have in your prod dataset and the number of workers you used to see such a behavior ?

It's not that I'm not believing you, I am, what I think that what might have happened to you is maybe like what we had with Double.NaN where 50% of our dataset was in only one partition.

Totoketchup:
I'd like to go a bit more in the anlyses of such a scenario:

Let's imagine the scenario where there is indeed a partiition completely empty after having used the repartitioning and let's go backward from this situation.

Let's say we have `N` rows and `P` partitions, and name the faulty parition `P_x`.

To know where a Row R_i is going, we compute `hash(R_i) % P`.
Now, what we have in this case is that for all rows `R_i: hash(R_i) % P != P_x`.  Which means that in this case the hash function is **never** returning a number where the remainder is never `x` but all other numbers in [1, P]. 
From this, we can see that this would be a huge flaw in MurmurHash3, right ?
Another thing is that by just changing P, then the whole behavior would change and your empty partition is not existing anymore.

What I mean by this analysis is that what you are observing might not be due to the hash function itself.
Do you agree with me ?

Totoketchup:
@hcho3 @CodingCat Could we move forward on this? :)

CodingCat:
merged thanks

PR #6038, opened 2020-08-20
GPUTreeShap

RAMitchell:
Initial GPU acceleration for shap values.

The following shows around 14x speed up with a 1080Ti vs AMD Ryzen 7, 8 core.

```python
import xgboost as xgb
import numpy as np
import time

rng = np.random.RandomState(5)
n = 1000
m = 1000
X = rng.rand(m, n)
y = rng.rand(m)
m_tests = [100000]
predictors = ["cpu_predictor", "gpu_predictor"]
model = xgb.XGBRegressor(tree_method='gpu_hist', max_depth=10, n_estimators=500,
                         predictor="gpu_predictor",
                         gpu_id=0,
                         min_child_weight=0.01)
model.fit(X, y)
results = {p: [] for p in predictors}
for m_test in m_tests:
    X_test = rng.rand(m_test, n)
    dtest = xgb.DMatrix(X_test)
    for p in predictors:
        model.get_booster().set_param({"predictor": p})
        start = time.perf_counter()
        xgb_shap = model.get_booster().predict(dtest, pred_contribs=True)
        results[p].append(time.perf_counter() - start)

print(results)
```
```
{'cpu_predictor': [26.5260732], 'gpu_predictor': [1.7506337000000016]}
```


tests/python-gpu/test_gpu_prediction.py
```diff
@@ -4,13 +4,19 @@
...
 import xgboost as xgb
+from hypothesis import given, strategies, assume, settings, note
 
 sys.path.append("tests/python")
 import testing as tm
 from test_predict import run_threaded_predict  # noqa
 
 rng = np.random.RandomState(1994)
 
+shap_parameter_strategy = strategies.fixed_dictionaries({
```
trivialfis, line 15: Can we have test for `num_parallel_tree` too?

src/predictor/gpu_predictor.cu
```diff
@@ -447,6 +456,61 @@ class GPUPredictor : public xgboost::Predictor {
...
+    const auto margin = p_fmat->Info().base_margin_.ConstDeviceSpan();
+    CHECK(margin.empty());
+    float base_score = model.learner_model_param->base_score;
+    auto d_phis = phis.data().get();
+    dh::LaunchN(
+        generic_param_->gpu_id,
+        p_fmat->Info().num_row_ * model.learner_model_param->num_output_group,
+        [=] __device__(size_t idx) {
+          d_phis[(idx + 1) * contributions_columns - 1] =
+              margin.empty() ? base_score : margin[idx];
```
trivialfis, line 496: Wouldn't you get out of bound error for forest or multi-class model?  Let's add a check before entering the kernel.

src/predictor/gpu_predictor.cu
```diff
@@ -510,6 +564,50 @@ class GPUPredictor : public xgboost::Predictor {
...
 
+  std::vector<gpu_treeshap::PathElement> ExtractPaths(
+      const gbm::GBTreeModel& model, size_t tree_limit) {
+    std::vector<gpu_treeshap::PathElement> paths;
+    size_t path_idx = 0;
+    CHECK_LE(tree_limit, model.trees.size());
+    for (auto i = 0ull; i < tree_limit; i++) {
+      const auto& tree = *model.trees.at(i);
+      size_t group = model.tree_info[i];
+      for (auto j = 0ull; j < tree.GetNodes().size(); j++) {
```
trivialfis, line 575: What happen if there's pruned node?

src/predictor/gpu_predictor.cu
RAMitchell: Don't think so, what check do you want specifically? I added a test for margin.

src/predictor/gpu_predictor.cu
RAMitchell: Good catch, added a check for this.

src/predictor/gpu_predictor.cu
trivialfis: `CHECK_EQ(margin.Size(), p_fmat->Info().num_row_)` or `CHECK_EQ(margin.Size(), p_fmat->Info().num_row_ * num_groups)`?  I can't be sure.

trivialfis (approved):

trivialfis:
Any chance you can write a demo in XGBoost?  I know we can look into treeshap repository, but it would be nice to have one here that we can run in CI.

slundberg:
I didn't trace each line out, but this looks good to me. My one comment is that with deep trees you can sometimes get numerical stability issues (like >50). It looks like the GPU is using the same precision as the CPU, but if it ever uses lower precision for the weights in the alg. that is something to watch out for.

src/predictor/gpu_predictor.cu
```diff
@@ -510,6 +563,50 @@ class GPUPredictor : public xgboost::Predictor {
...
+        const auto& nodes = tree.GetNodes();
```
hcho3, line 575: Shouldn't this declaration be moved out to the first `for` loop, since it doesn't depend on `j`?

src/predictor/gpu_predictor.cu
```diff
@@ -510,6 +563,50 @@ class GPUPredictor : public xgboost::Predictor {
...
+            auto parent = nodes[child.Parent()];
+            CHECK(parent.LeftChild() == child_idx ||
+                  parent.RightChild() == child_idx);
+            bool is_left_path = parent.LeftChild() == child_idx;
+            bool is_missing_path = (!parent.DefaultLeft() && !is_left_path) ||
+                                   (parent.DefaultLeft() && is_left_path);
+            float lower_bound = is_left_path ? -inf : parent.SplitCond();
+            float upper_bound = is_left_path ? parent.SplitCond() : inf;
+            paths.emplace_back(path_idx, parent.SplitIndex(), group,
+
```
hcho3, line 595: Remove this extra blank line.

src/predictor/gpu_predictor.cu
```diff
@@ -510,6 +563,50 @@ class GPUPredictor : public xgboost::Predictor {
...
+                               lower_bound, upper_bound, is_missing_path,
```
hcho3, line 596: @RAMitchell  Is the feature bounds  (`lower_bound`, `upper_bound`) used for detecting duplicate features in each tree path? I took a brief look at GPUTreeshap, and that's what I inferred those bounds to be for. It's pretty cool that we don't have to "unwind" paths as in the original SHAP algorithm.

Reminder to myself: Make time to study GPUTreeshap code.

hcho3: Also, does a unique traversal path in a tree consist of a sequence of `gpu_treeshap::PathElement`?

tests/python-gpu/test_gpu_prediction.py
```diff
@@ -185,3 +193,24 @@ def predict_df(x):
...
+    @settings(deadline=None)
+    def test_shap(self, num_rounds, dataset, param, all_rows):
+        param.update({"predictor": "gpu_predictor", "gpu_id": 0})
+        param = dataset.set_params(param)
+        dmat = dataset.get_dmat()
+        bst = xgb.train(param, dmat, num_rounds)
+        if all_rows:
+            test_dmat = xgb.DMatrix(dataset.X, dataset.y, dataset.w, dataset.margin)
+        else:
+            test_dmat = xgb.DMatrix(dataset.X[0:1, :])
```
hcho3, line 208: Wouldn't it be better to take a subsample with multiple rows, instead of picking the first row?

tests/python/testing.py
```diff
@@ -206,16 +207,23 @@ def get_sparse():
 
 
 @strategies.composite
-def _dataset_and_weight(draw):
+def _dataset_weight_margin(draw):
     data = draw(_unweighted_datasets_strategy)
     if draw(strategies.booleans()):
         data.w = draw(arrays(np.float64, (len(data.y)), elements=strategies.floats(0.1, 2.0)))
+    if draw(strategies.booleans()):
```
hcho3, line 214: Why duplicate an `if` condition?

src/predictor/gpu_predictor.cu
RAMitchell: Yes, any number of splits on the same feature can be represented by these bounds.

> Also, does a unique traversal path in a tree consist of a sequence of gpu_treeshap::PathElement?

Yes

tests/python-gpu/test_gpu_prediction.py
RAMitchell: This is for hypothesis to minimise failing examples to a single row. For example if I break something major while developing a can examine the result for a very simple example.

tests/python/testing.py
RAMitchell: The samples from a random variable, so we can get four possible combinations of DMatrix with weights and base_margin on/off.

tests/python-gpu/test_gpu_prediction.py
hcho3: Got it.

PR #6042, opened 2020-08-20
Remove win2016 jvm github action test.

trivialfis:
It's constantly failing on PR with network errors.  After the removal we still have win-latest.  Suggestions are welcomed.


hcho3:
Thanks! Using the latest Windows looks reasonable to me.

PR #6061, opened 2020-08-26
Make driver reusable.

trivialfis:
* Move driver into tree directory.
* Remove `DeviceSplitCandidate`.

Now `GradStats` is used in some places.  Will post benchmark later.


trivialfis:
Stalled.  I will approach it in a different PR.

PR #6062, opened 2020-08-26
Unify set index data.

trivialfis:


trivialfis:
@ShvetsKS

trivialfis:
@hcho3 @RAMitchell @SmirnovEgorRu Please help reviewing.

RAMitchell (commented):
The reorganisation looks good. Does the lambda get compiled inline? The difference between a function call and inlining could be large here.

trivialfis:
@RAMitchell  I believe so, examining it with gdb:

```
(gdb) print get_offset
$2 = {__offsets = 0x555555ca1630}
(gdb) disassemble get_offset
Value can't be converted to integer.
```

Issue #6092, opened 2020-09-07
CMake should not install librabit.a when running `make install`


hcho3:
Currently, the `install` target generated by the CMake script will install librabit.a (Rabit static lib) into the prefix. This is unnecessary, since libxgboost.so already contains all Rabit's object code.

TODO: Refactor Rabit's CMake script.

hcho3:
Fixed in #6096

PR #6096, opened 2020-09-08
Refactor rabit tests

trivialfis:
* Merge RABIT google tests into XGBoost.
* Run standalone tests in CTest.
* Run tests on Jenkins.
* Simplify CMakeLists.txt, fixing #6092 .


trivialfis:
Somehow there are duplicated tests defined in cc files and cpp files.

trivialfis:
@hcho3 I will do a drive by fix for #6092 .

rabit/CMakeLists.txt
```diff
@@ -94,89 +63,21 @@ if (RABIT_BUILD_TESTS)
     add_executable(${test} test/${test}.cc)
     target_link_libraries(${test} rabit_mock_static)
     set_target_properties(${test}  PROPERTIES CXX_STANDARD 11 CXX_STANDARD_REQUIRED ON)
```
hcho3, line 65:
```suggestion
    set_target_properties(${test}  PROPERTIES CXX_STANDARD 14 CXX_STANDARD_REQUIRED ON)
```

hcho3 (changes requested):
I did a first round of review.

You also need to change line 287 of the root CMakeLists.txt: We should install `rabit` along with `objxgboost` when `BUILD_STATIC_LIB` is on. So add it to the variable `INSTALL_TARGETS`. The reason is that Rabit is a dependency of `xgboost`, and when you install a static lib, you need to install all of its dependencies.

hcho3:
I'll do another round of review tomorrow, as I am about to go to bed now.

trivialfis:
@hcho3 Thanks for the explanation.  Let me do some tests, I think it's possible to link multiple static libraries into one.

hcho3:
> I think it's possible to link multiple static libraries into one.

You can turn rabit into an object library, so that librabit.a won't need to be installed. Even here, we still need to declare rabit in the export set of INSTALL(TARGETS) command.

trivialfis:
Got it, so I need to restore the installation logic.

hcho3:
No, you can just use the installation logic in the **root** CMakeLists.txt. No need to maintain installation logic in two places.

trivialfis:
@hcho3 Thanks again.  Modified the script to install it along with objxgboost when building static library.

hcho3 (approved):

PR #6101, opened 2020-09-09
Merge logging facilities for rabit and xgboost.

trivialfis:
* Enable clang-tidy on CI.
* Remove most of the debug logging.
* Remove unused utilities.
* Move standalone tests CMake config into top level for linking.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -341,7 +341,6 @@ endif (GOOGLE_TEST)
 
 if (GOOGLE_TEST AND (NOT WIN32))
   # rabit mock based integration tests
-  list(REMOVE_ITEM rabit_libs "rabit_mock_static") # remove here to avoid installing it
   set(tests lazy_recover local_recover model_recover)
 
   foreach(test ${tests})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -339,6 +339,26 @@ if (GOOGLE_TEST)
     PASS_REGULAR_EXPRESSION ".*test-rmse:0.087.*")
 endif (GOOGLE_TEST)
 
+if (GOOGLE_TEST AND (NOT WIN32))
+  # rabit mock based integration tests
+  list(REMOVE_ITEM rabit_libs "rabit_mock_static") # remove here to avoid installing it
+  set(tests lazy_recover local_recover model_recover)
+
+  foreach(test ${tests})
+    add_executable(${test} rabit/test/${test}.cc)
+    set_output_directory(${test} ${xgboost_BINARY_DIR})
+    target_link_libraries(${test} rabit_mock_static xgboost)
+    set_target_properties(${test}  PROPERTIES CXX_STANDARD 14 CXX_STANDARD_REQUIRED ON)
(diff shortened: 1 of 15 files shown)


CMakeLists.txt
```diff
@@ -339,6 +339,26 @@ if (GOOGLE_TEST)
...
+    add_test(NAME ${test} COMMAND ${test} WORKING_DIRECTORY ${xgboost_BINARY_DIR})
+  endforeach()
+
+  if(RABIT_BUILD_MPI)
+    add_executable(speed_test_mpi test/speed_test.cc)
+    target_link_libraries(speed_test_mpi rabit_mpi)
+    add_test(NAME speed_test_mpi COMMAND speed_test_mpi WORKING_DIRECTORY ${xgboost_BINARY_DIR})
+  endif(RABIT_BUILD_MPI)
+endif (GOOGLE_TEST AND (NOT WIN32))
+
```
hcho3, line 361: Can we keep this block in rabit/CMakeLists.txt ? What is the benefit of moving it to the top CMakeLists.txt ?

trivialfis: @hcho3 

> Can we keep this block in rabit/CMakeLists.txt ? What is the benefit of moving it to the top CMakeLists.txt ?

Because

> Move standalone tests CMake config into top level for linking.

Now the logging facility comes from XGBoost, linking to libxgboost.so is required for these standalone tests.

hcho3: For the sake of modularity, I prefer that this block would stay in rabit/CMakeList.txt. Do we lose access to `xgboost` target from rabit/CMakeLists.txt ?

hcho3: Also, do we actually need to link against libxgboost ? I thought logging facilities came from libdmlc.a (`dmlc` target).

hcho3: Never mind, we actually want `xgboost/logging.h`, not `dmlc/logging.h`. So we do need libxgboost.

trivialfis: Yup.  The `verbosity` is defined in XGBoost so we need the logging facility from XGBoost instead of dmlc.

hcho3: Can we still move the block back to rabit/CMakeLists.txt? Does it interfere with linking with `xgboost`?

trivialfis: Let me think about it.  As now rabit is part of XGBoost, so they will have intertwined linkage.  I'm actually considering removing this static library.

hcho3: I agree, let's think about it a bit more.

Some choices I can think of: 

Option 1. Remove the `rabit` CMake target entirely and merge all the source and header files to `objxgboost`.
Option 2. Separate the logging portion of XGBoost and create a new CMake object library target `objxgboost_logging`. Then link this library to both Rabit and objxgboost.

@trivialfis Would you like me to review this pull request as it is, for now?

trivialfis: > Would you like me to review this pull request as it is, for now?

No, I converted it to draft.  It's already having linkage issue on Jenkins and I don't know how to resolve it.

PR #6105, opened 2020-09-09
Enable building rabit on Windows

trivialfis:
No test is performed.  Just trying to get rid of the `rabit_empty` target.


hcho3:
Let’s remove the Rabit Makefile, since XGBoost doesn’t use Makefile.

trivialfis:
@hcho3 We have R package ..

hcho3:
Ah so Rabit is not part of the amalgamation and it needs a Makefile. Got it.

trivialfis:
It seems autotools is picking mingw over msvc.

trivialfis:
Turns out I can remove the makefile.

rabit/include/rabit/internal/socket.h
```diff
@@ -21,31 +24,92 @@
...
-                       int timeout) { return WSAPoll ( pfd, nfds, timeout ); }
 #else
+
 #include <sys/poll.h>
 using SOCKET = int;
 using sock_size_t = size_t;  // NOLINT
-const int kInvalidSocket = -1;
 #endif  // defined(_WIN32)
 
+#define IS_MINGW() defined(__MINGW32__)
```
RAMitchell, line 50: Ifdef'ing everything seems quite a messy solution. Any alternatives?

RAMitchell (commented):
Let me see if I understand. This PR is removing the dummy library that used to be linked into the windows Xgboost. Macros are being used instead for features that don't work on windows. Does rabit actually work in any way on windows or is it still disabled?

rabit/include/rabit/internal/socket.h
trivialfis: I don't have any other options.

RAMitchell: I can think of anything better, as you said at least it's constrained to one file.

rabit/include/rabit/internal/utils.h
```diff
@@ -180,6 +162,19 @@ inline std::FILE *FopenCheck(const char *fname, const char *flag) {
   return fp;
 }
 }  // namespace utils
+
+// Can not use std::min on Windows with msvc due to:
```
hcho3, line 166: We could define `NOMINMAX` macro to disable `MIN` and `MAX` macros from Windows.h.

tests/ci_build/test_r_package.py
```diff
@@ -26,22 +26,13 @@ def get_mingw_bin():
...
-            cmd = ['R.exe', 'CMD', 'INSTALL', str(os.path.curdir)]
-            env = os.environ.copy()
-            env.update({'CC': CC, 'CXX': CXX})
-            subprocess.check_call(cmd, env=env)
-        elif args.compiler == 'msvc':
-            cmd = ['R.exe', 'CMD', 'INSTALL', str(os.path.curdir)]
-            env = os.environ.copy()
-            # autotool favor mingw by default.
-            env.update({'CC': 'cl.exe', 'CXX': 'cl.exe'})
-            subprocess.check_call(cmd, env=env)
```
hcho3, line 42: Why are we removing MSVC from the R package test?

hcho3: Ah so we are requiring CMake for MSVC compiler?

rabit/include/rabit/internal/utils.h
trivialfis: That's a bit weird.  Can we just stick with current approach?  That feels like an invitation to header include ordering issues.

tests/ci_build/test_r_package.py
trivialfis: Yup, I tried to debug some errors and found that it's picking mingw disregarding the env.

rabit/include/rabit/internal/utils.h
hcho3: The current approach is fine.

hcho3 (approved):

Issue #5905, opened 2020-07-17
[R] Remove stringi dependency


vspinu:
I am building a predictor for amazon lambda where the hard limit of all unpacked dependencies is 250MB. The xgboost + deps is already 92MB mostly due to the 56MB of stringi. 

I have looked at the code-base briefly and all of the use-cases of stringi seem to be easily replaceable by base R functionally. 

Would you be ok with that? I can have a look into a PR.

trivialfis:
Surely.  But are you compiling everything from source? If so, did you strip out the debug symbols in stringi and xgboost?

trivialfis:
I noticed that for CPU only build, with gcc-9, the shared object is only 5.5MB.

vspinu:
I am installing from CRAN archive. This is how it looks:

```
g++ -std=gnu++11 -I"/home/vspinu/bin/R-4.0.0-bin/include" -DNDEBUG -I./include -I./dmlc-core/include -I./rabit/include -I. -DXGBOOST_STRICT_R_MODE=1 -DDMLC_LOG_BEFORE_THROW=0 -DDMLC_ENABLE_STD_THREAD=1 -DDMLC_DISABLE_STDIN=1 -DDMLC_LOG_CUSTOMIZE=1 -DXGBOOST_CUSTOMIZE_LOGGER=1 -DRABIT_CUSTOMIZE_MSG_ -DRABIT_STRICT_CXX98_  -I/usr/local/include  -fopenmp -DDMLC_CMAKE_LITTLE_ENDIAN=1 -pthread -fpic  -g -O2  -c xgboost_R.cc -o xgboost_R.o
g++ -std=gnu++11 -I"/home/vspinu/bin/R-4.0.0-bin/include" -DNDEBUG -I./include -I./dmlc-core/include -I./rabit/include -I. -DXGBOOST_STRICT_R_MODE=1 -DDMLC_LOG_BEFORE_THROW=0 -DDMLC_ENABLE_STD_THREAD=1 -DDMLC_DISABLE_STDIN=1 -DDMLC_LOG_CUSTOMIZE=1 -DXGBOOST_CUSTOMIZE_LOGGER=1 -DRABIT_CUSTOMIZE_MSG_ -DRABIT_STRICT_CXX98_  -I/usr/local/include  -fopenmp -DDMLC_CMAKE_LITTLE_ENDIAN=1 -pthread -fpic  -g -O2  -c xgboost_custom.cc -o xgboost_custom.o
[...]

PR #6109, opened 2020-09-10
[R] Remove stringi dependency

vspinu:
Per discussion in #5905


R-package/tests/testthat/test_model_compatibility.R
```diff
@@ -73,12 +75,15 @@ test_that("Models from previous versions of XGBoost can be loaded", {
...
     name <- m[3]
 
-    if (endsWith(model_file, '.rds')) {
-      booster <- readRDS(model_file)
-    } else {
-      booster <- xgb.load(model_file)
-    }
-    predict(booster, newdata = pred_data)
-    run_booster_check(booster, name)
+    ## don't print C level compatibility warnings
```
hcho3, line 78: Please keep the compatibility warnings. The warnings are expected.

hcho3 (changes requested):
Please keep `verbose = 1`. The verbose output helps us diagnose should any issues arise. All other tests also has verbose outputs.

R-package/tests/testthat/test_model_compatibility.R
vspinu: This is exactly why I suppressed them, because they are expected. They just pollute the test output.

vspinu:
The train output makes it difficult to see what's going on during local development. But ok, if this is standard practice here.

R-package/tests/testthat/test_model_compatibility.R
hcho3: For this test case, we can add `gives_warning()` to ensure that this test throws a warning. I can do that in a future pull request.

vspinu: I have inhibited only C level warnings here, the R level warnings are still there and testhat still complains. So it's redundat. But ok, I have removed for now. It's for another PR indeed.

hcho3 (approved):
Thanks a lot!

hcho3:
@vspinu I opened #6114 to suppress warnings in test_model_compatibility.R.

PR #6112, opened 2020-09-11
[WIP] Drop single point model recovery.

trivialfis:


rabit/src/allreduce_base.h
```diff
@@ -35,6 +37,22 @@ class Datatype {
 }
 namespace rabit {
 namespace engine {
+
+template <typename Fn, typename... Args>
+auto MakeTimeoutTask(Fn&& fn, int32_t timeout, int32_t rank) {
```
trivialfis, line 42: The old time out implementation is incorrect.  Exception needs to  be thrown at main thread.

trivialfis:
## RABIT parameter handling.

Currently on master branch, the test is only passing because of an unrelated error:

```
AssertError:buf should have data from resbuf, rabit is configured to keep process running
```

This exception has nothing to do with RABIT timeout.  Once `rabit_robust` is removed, this
assertion will no longer exist and the test will fail as it's testing timeout facility.  I
dig into it and found the parameter "rabit_timeout" is not being passed down into rabit at
all.  Tracing the line changes, I found that #5082 has removed these lines in
`xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala`:

```
    val xgbRabitParams = xgbParamsFactory.buildRabitParams.asJava
    ...
            tracker.getWorkerEnvs().putAll(xgbRabitParams)
```
so currently non of the rabit parameter is correctly handled.  This PR adds back these 2 lines.

## Timeout

The old implementation of timeout used a async worker thread to throw the exception, which is incorrect as exception is caught by main thread.

trivialfis:
Closing to avoid spamming.

PR #6114, opened 2020-09-13
[R] Check warnings explicitly for model compatibility tests

hcho3:
* Suppress warnings in test_model_compatibility.R, as they clutter test results.
* Furthermore, check that the correct types of warnings are thrown in test_model_compatibility.R.

@vspinu Can you review?


trivialfis (commented):
I think we can remove that warning in longer term.  Just as an example, if I unify the `single_precision_histogram` for tree methods, your RDS breaks.

hcho3:
Yes, the warning is meant to encourage users to move away from `saveRDS()` and use `xgb.save()` instead.

trivialfis:
@hcho3 What if we make a try catch block in serialization functions?

R-package/tests/testthat/test_model_compatibility.R
```diff
@@ -72,13 +72,34 @@ test_that("Models from previous versions of XGBoost can be loaded", {
...
+    })
+    if (compareVersion(model_xgb_ver, '1.0.0.0') < 0) {
+      # Expect a C++ warning when a model was generated in version < 1.0.x
+      m <- grepl(paste0('.*Loading model from XGBoost < 1\\.0\\.0, consider saving it again for ',
+                        'improved compatibility.*'), cpp_warning, perl = TRUE)
+      expect_true(length(m) > 0 && all(m))
+    } else if (is_rds && model_xgb_ver == '1.1.1.1') {
+      # Expect a C++ warning when a model is loaded from RDS and it was generated by version 1.1.x
+      m <- grepl(paste0('.*Attempted to load interal configuration for a model file that was ',
+                        'generated by a previous version of XGBoost.*'), cpp_warning, perl = TRUE)
```
vspinu, line 101: There is a typo "interal" here, but I guess it is how it is in the C code.

R-package/tests/testthat/test_model_compatibility.R
```diff
@@ -72,13 +72,34 @@ test_that("Models from previous versions of XGBoost can be loaded", {
...
     name <- m[3]
+    is_rds <- endsWith(model_file, '.rds')
 
-    if (endsWith(model_file, '.rds')) {
-      booster <- readRDS(model_file)
-    } else {
-      booster <- xgb.load(model_file)
+    cpp_warning <- capture.output({
+      # Expect an R warning when a model is loaded from RDS and it was generated by version < 1.1.x
+      if (is_rds && compareVersion(model_xgb_ver, '1.1.1.1') < 0) {
```
vspinu, line 79: I am a bit confused by this logic. So for ver >= 1.1.1.1, `predict` and `run_booster_check` no longer throw warnings when the model is loaded from RDS? Wouldn't you want to discourage it regardless and throw warning?

R-package/tests/testthat/test_model_compatibility.R
```diff
@@ -72,13 +72,34 @@ test_that("Models from previous versions of XGBoost can be loaded", {
...
+        predict(booster, newdata = pred_data)
+        run_booster_check(booster, name)
+      }
+    })
+    if (compareVersion(model_xgb_ver, '1.0.0.0') < 0) {
+      # Expect a C++ warning when a model was generated in version < 1.0.x
+      m <- grepl(paste0('.*Loading model from XGBoost < 1\\.0\\.0, consider saving it again for ',
+                        'improved compatibility.*'), cpp_warning, perl = TRUE)
+      expect_true(length(m) > 0 && all(m))
+    } else if (is_rds && model_xgb_ver == '1.1.1.1') {
```
vspinu, line 98: Version is hardcoded here, so for new models this if-else won't be triggered at all?

vspinu (commented):
I still see some internal warnings. Maybe those could be gotten rid of as well?

```
⠏ |   0       | update trees in an existing model[10:46:30] WARNING: ../..//amalgamation/../src/gbm/gbtree.cc:81: DANGER AHEAD: You have manually specified `updater` parameter. The `tree_method` parameter will be ignored. Incorrect sequence of updaters will produce undefined behavior. For common uses, we recommend using`tree_method` parameter instead.
⠋ |   1       | update trees in an existing model[10:46:30] WARNING: ../..//amalgamation/../src/gbm/gbtree.cc:81: DANGER AHEAD: You have manually specified `updater` parameter. The `tree_method` parameter will be ignored. Incorrect sequence of updaters will produce undefined behavior. For common uses, we recommend using`tree_method` parameter instead.
[10:46:30] WARNING: ../..//amalgamation/../src/gbm/gbtree.cc:81: DANGER AHEAD: You have manually specified `updater` parameter. The `tree_method` parameter will be ignored. Incorrect sequence of updaters will produce undefined behavior. For common uses, we recommend using`tree_method` parameter instead.
[10:46:30] WARNING: ../..//amalgamation/../src/gbm/gbtree.cc:81: DANGER AHEAD: You have manually specified `updater` parameter. The `tree_method` parameter will be ignored. Incorrect sequence of updaters will produce undefined behavior. For common uses, we recommend using`tree_method` parameter instead.
⠇ |   9       | update trees in an existing model[10:46:30] WARNING: ../..//amalgamation/../src/gbm/gbtree.cc:81: DANGER AHEAD: You have manually specified `updater` parameter. The `tree_method` parameter will be ignored. Incorrect sequence of updaters will produce undefined behavior. For common uses, we recommend using`tree_method` parameter instead.
[10:46:30] WARNING: ../..//amalgamation/../src/gbm/gbtree.cc:81: DANGER AHEAD: You have manually specified `updater` parameter. The `tree_method` parameter will be ignored. Incorrect sequence of updaters will produce undefined behavior. For common uses, we recommend using`tree_method` parameter instead.
✔ |  22       | update trees in an existing model [0.3 s]
```

Also there is a download message:

```
⠏ |   0       | Models from previous versions of XGBoost can be loadedtrying URL 'https://xgboost-ci-jenkins-artifacts.s3-us-west-2.amazonaws.com/xgboost_r_model_compatibility_test.zip'
Content type 'application/zip' length 63026 bytes (61 KB)
==================================================
downloaded 61 KB
```

It can be disabled by `quiet = TRUE` argument of `download.file`, (optionally followed with `on.exit(unlink(zip_file, force = TRUE))` in order to avoid it showing in git status).

R-package/tests/testthat/test_model_compatibility.R
hcho3: That's correct. When RDS is used, XGBoost package does not actually know that RDS is used. So the warning is only generated when a material change in the model representation is detected. There was a major change in 1.1.1.1, and a minor one in 1.2.0.1. The major change is covered by the R warning ("The model had been generated by XGBoost version 1.0.0 or earlier and was loaded from a RDS file") and the minor change is covered by the C++ warning ("Attempted to load internal configuration for a model file that was generated by a previous version of XGBoost").

R-package/tests/testthat/test_model_compatibility.R
hcho3: Thanks. I'll fix the typo.

R-package/tests/testthat/test_model_compatibility.R
hcho3: That's also correct, since no C++ warning is currently generated for the models that were generated by version 1.2.0.x.

R-package/tests/testthat/test_model_compatibility.R
hcho3: The official documentation has a clear guidance against RDS: https://www.rdocumentation.org/packages/xgboost/versions/1.2.0.1/topics/a-compatibility-note-for-saveRDS-save

vspinu (approved):

PR #6117, opened 2020-09-14
Allow kwargs in dask predict

RAMitchell:


trivialfis (commented):
So list arguments are not used?

RAMitchell:
I think requiring explicit kwargs is safer as inputs can't be unintentionally used for the wrong argument.

trivialfis (approved):
LGTM.

PR #6120, opened 2020-09-15
Handle duplicated values in sketching.

trivialfis:
* Reuse the pruning code for initialization.
* Run a 3 passes procedure to accumulate weights in duplicated values.


trivialfis:
Avoid spamming the list.

PR #6124, opened 2020-09-16
Aggregate weights in duplicated elements for GPU GK.

trivialfis:
See comments in the patch.

I tried to accumulate all the weights from duplicated entries and it became quite messy with multiple batches.  To achieve that, we first need to refactor `hist_util.cuh` and make `ExtractCuts` part of `QuantileContainer`, then add the scanned weights from all previous batches to `rmin` and `rmax` of new entries.


trivialfis:
On a second thought, let me do some refactoring first.

PR #6128, opened 2020-09-17
Improve JSON format for categorical features.

trivialfis:
* This PR reduces the saved categories to a single vector instead of vector per-node.

### DIFF
--- a/src/tree/tree_model.cc
+++ b/src/tree/tree_model.cc
@@ -772,6 +772,65 @@ void RegTree::Save(dmlc::Stream* fo) const {
   }
 }
 
+void RegTree::LoadCategoricalSplit(std::vector<Json> *p_split_type,
+                                   std::vector<Json> *p_categories_segments,
+                                   std::vector<Json> *p_categories,
+                                   bst_node_t i) {
+  auto& split_type = *p_split_type;
+  auto& categories_segments = *p_categories_segments;
+  auto& categories = *p_categories;
+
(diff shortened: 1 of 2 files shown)


hcho3:
@trivialfis Do you have performance numbers? How much does this PR improve serialization performance?

src/tree/tree_model.cc
```diff
@@ -772,6 +776,92 @@ void RegTree::Save(dmlc::Stream* fo) const {
...
+      categories_nodes.emplace_back(i);
+      auto begin = categories.size();
+      categories_segments.emplace_back(static_cast<Integer::Int>(begin));
+      auto segment = split_categories_segments_[i];
+      auto node_categories =
+          this->GetSplitCategories().subspan(segment.beg, segment.size);
+      common::KCatBitField const cat_bits(node_categories);
+      for (size_t i = 0; i < cat_bits.Size(); ++i) {
+        if (cat_bits.Check(i)) {
+          categories.emplace_back(static_cast<Integer::Int>(i));
```
hcho3, line 851: @trivialfis What's the performance implication of saving individual categories? Is it better than saving bitmaps directly?

trivialfis: No, it's not.  If saving bitmap is preferred I can make the change.

hcho3: No, if saving individual categories produces acceptable performance, let's keep it. It's easier to parse by a human.

hcho3 (approved):

PR #6129, opened 2020-09-17
Implement intrusive ptr

trivialfis:
Previously the PR is hosted on dmlc-core.  Let's merge it here for XGBoost first so we can get better perf for next release with JSON now being the default serialization format.

@hcho3

### DIFF
--- a/include/xgboost/intrusive_ptr.h
+++ b/include/xgboost/intrusive_ptr.h
@@ -1,0 +1,187 @@
+/*!
+ * Copyright (c) by Contributors 2020
+ * \file intrusive_ptr.h
+ * \brief Implementation of Intrusive Ptr.
+ */
+#ifndef DMLC_INTRUSIVE_PTR_H_
+#define DMLC_INTRUSIVE_PTR_H_
+
+#include <atomic>
+#include <cinttypes>
+#include <functional>
+
+namespace xgboost {
+/*!
+ * \brief Helper class for embedding reference counting into client objects.
+ */
+class IntrusivePtrCell {
+ private:
+  std::atomic<int32_t> count_;
+  template <typename T> friend class IntrusivePtr;
+
+  std::int32_t IncRef() noexcept {
+    return count_.fetch_add(1, std::memory_order_relaxed);
+  }
+  std::int32_t DecRef() noexcept {
+    return count_.fetch_sub(1, std::memory_order_release);
+  }
+  bool IsZero() const { return Count() == 0; }
+
+ public:
+  IntrusivePtrCell() noexcept : count_{0} {}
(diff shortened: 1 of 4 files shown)


hcho3:
@trivialfis Do you have some benchmark results?

trivialfis:
The model is trained with covtype dataset, generated 21000 trees in total with
max_depth set to 6.  The size of saved JSON file is about 96.1MB.

|                                        | Load               | Save               |
|:---------------------------------------|:-------------------|:-------------------|
| Before adding categorical data support | 4.2141406536102295 | 3.429635524749756  |
| Master                                 | 4.652015209197998  | 3.8093554973602295 |
| With new categorical split format      | 4.365228891372681  | 3.491779327392578  |
| With intrusive ptr                     | 2.460336446762085  | 2.5140161514282227 |

include/xgboost/intrusive_ptr.h
```diff
@@ -0,0 +1,187 @@
...
+#ifndef XGBOOST_INTRUSIVE_PTR_H_
+#define XGBOOST_INTRUSIVE_PTR_H_
+
+#include <atomic>
+#include <cinttypes>
+#include <functional>
+
+namespace xgboost {
+/*!
+ * \brief Helper class for embedding reference counting into client objects.
```
hcho3, line 15: We should document why it's okay to use `std:: memory_order_relaxed` when incrementing the reference count.
```suggestion
 * \brief Helper class for embedding reference counting into client objects.
 * See https://www.boost.org/doc/libs/1_74_0/doc/html/atomic/usage_examples.html
 * for discussions of memory order.
```

include/xgboost/intrusive_ptr.h
```diff
@@ -0,0 +1,187 @@
...
+  int32_t Count() const { return count_.load(std::memory_order_relaxed); }
+};
+
+/*!
+ * \brief User defined function for returing embedded reference count.
+ */
+template <typename T> IntrusivePtrCell &IntrusivePtrRefCount(T const *ptr) noexcept;
+
+/*!
+ * \brief Implementation of Intrusive Pointer.
```
hcho3, line 41:
```suggestion
 * \brief Implementation of Intrusive Pointer, a smart pointer that points to an object
 *        with an embedded reference counter. The underlying object must implement
 *        a friend function IntrusivePtrRefCount() that returns the ref counter
 *        (of type IntrusivePtrCell). Embedding the ref counter in the managed object
 *        improves performance, compared to allocating a separate ref counter.
```

hcho3 (approved):

include/xgboost/intrusive_ptr.h
trivialfis: Let me add an example usage.

PR #6131, opened 2020-09-18
Use caching allocator from rmm instead of cub.

trivialfis:


src/common/device_helpers.cuh
```diff
@@ -406,10 +406,14 @@ struct XGBDefaultDeviceAllocatorImpl : XGBBaseDeviceAllocator<T> {
   }
 #if defined(XGBOOST_USE_RMM) && XGBOOST_USE_RMM == 1
   XGBDefaultDeviceAllocatorImpl()
-    : SuperT(rmm::mr::get_current_device_resource(), cudaStream_t{0}) {}
+    : SuperT(rmm::mr::get_current_device_resource(), cudaStream_t{nullptr}) {}
```
hcho3, line 409: Is CUDA stream a pointer?

hcho3 (approved):

src/common/device_helpers.cuh
trivialfis: Yup.  It's an opaque pointer.

hcho3: Thanks. All this time I thought it was an integer like a file descriptor. I learned something new today.

PR #6132, opened 2020-09-18
[dask] Support more meta data.

trivialfis:
* label_lower_bound
* label_upper_bound
* base_margin

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -191,6 +191,9 @@ class DaskDMatrix:
                  label=None,
                  missing=None,
                  weight=None,
+                 base_margin=None,
+                 label_lower_bound=None,
+                 label_upper_bound=None,
                  feature_names=None,
                  feature_types=None):
         _assert_dask_support()
@@ -221,7 +224,9 @@ class DaskDMatrix:
     def __await__(self):
         return self._init.__await__()
 
-    async def map_local_data(self, client, data, label=None, weights=None):
+    async def map_local_data(self, client, data, label=None, weights=None,
+                             base_margin=None,
+                             label_lower_bound=None, label_upper_bound=None):
         '''Obtain references to local data.'''
 
         def inconsistent(left, left_name, right, right_name):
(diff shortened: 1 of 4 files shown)


trivialfis:
@hcho3 I modified the test for survival training to use it for dask.  Could you please help taking a look?

RAMitchell (approved):

python-package/xgboost/dask.py
```diff
@@ -191,6 +191,9 @@ def __init__(self,
                  label=None,
                  missing=None,
                  weight=None,
+                 base_margin=None,
+                 label_lower_bound=None,
+                 label_upper_bound=None,
```
hcho3, line 196: These new parameter should be added to the docstring.

tests/python/test_with_dask.py
```diff
@@ -482,16 +481,59 @@ def test_predict():
...
-            shap = xgb.dask.predict(client, model=booster, data=dtrain, pred_contribs=True)
+            shap = xgb.dask.predict(client, model=booster, data=dtrain,
+                                    pred_contribs=True)
             assert shap.ndim == 2
             assert shap.shape[0] == kRows
             assert shap.shape[1] == kCols + 1
 
 
+def test_aft_survival():
+    # survival doesn't handle empty dataset well.
```
hcho3, line 497: I did not know this. How can we fix this?

hcho3 (commented):
Thanks for adding survival analysis in the Dask test. Some comments:

tests/python/test_with_dask.py
trivialfis: @hcho3 There are checks:
``` c++
    CHECK_NE(info.labels_lower_bound_.Size(), 0U)
      << "labels_lower_bound cannot be empty";
    CHECK_NE(info.labels_upper_bound_.Size(), 0U)
      << "labels_upper_bound cannot be empty";
```

in `survival_metric.cu::EvalEWiseSurvivalBase::Eval`.

hcho3: Should we remove the checks and have the objective return zero?

trivialfis: Depends on the synchronization I think.

hcho3: AFT is an element-wise objective, so it should behave like `reg:squarederror` when it comes to handling empty datasets. Let me create a new issue for this.

hcho3: Opened #6139

hcho3 (approved):

PR #6136, opened 2020-09-20
[jvm-packages]add xgboost4j-gpu/xgboost4j-spark-gpu module to facilitate release

CodingCat:
this diff adds two new (dummy) modules just to facilitate the release of gpu-enabled jvm components to maven central


hcho3 (approved):

trivialfis (commented):
Do we need to make change to the release script?

CodingCat:
> Do we need to make change to the release script?

we don't have a release script now, currently I am just download -gpu artifacts from s3 and publish to maven central

but that's a good point, I will add script soon

PR #6145, opened 2020-09-22
Revert #5755

hcho3:
Temporarily revert #5755 to prevent OOM error in the [e2e mortgage notebook](https://github.com/rapidsai-community/notebooks-contrib/blob/branch-0.14/intermediate_notebooks/E2E/mortgage/mortgage_e2e.ipynb).

### DIFF
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -539,8 +539,7 @@ XGB_DLL int XGBoosterPredictFromDense(BoosterHandle handle, float *values,
   CHECK_EQ(cache_id, 0) << "Cache ID is not supported yet";
   auto *learner = static_cast<xgboost::Learner *>(handle);
 
-  std::shared_ptr<xgboost::data::DenseAdapter> x{
-    new xgboost::data::DenseAdapter(values, n_rows, n_cols)};
+  auto x = xgboost::data::DenseAdapter(values, n_rows, n_cols);
   HostDeviceVector<float>* p_predt { nullptr };
   std::string type { c_type };
   learner->InplacePredict(x, type, missing, &p_predt);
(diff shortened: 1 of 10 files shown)


trivialfis:
I'm still looking into it.  Expect some results today.  Reverting it might break quantile dmatrix completely.

hcho3:
@trivialfis At least it does not break the gtest. Let's see whether the Python tests would pass too.

RAMitchell:
The relevant PR removes the specialised copy for column wise data and instead uses this:
```
// Here the data is already correctly ordered and simply needs to be compacted
// to remove missing data
template <typename AdapterT>
void CopyDataToDMatrix(AdapterT* adapter, common::Span<Entry> data,
                       int device_idx, float missing,
                       common::Span<size_t> row_ptr) {
  auto& batch = adapter->Value();
  auto transform_f = [=] __device__(size_t idx) {
    const auto& e = batch.GetElement(idx);
    return Entry(e.column_idx, e.value);
  };  // NOLINT
  auto counting = thrust::make_counting_iterator(0llu);
  thrust::transform_iterator<decltype(transform_f), decltype(counting), Entry>
      transform_iter(counting, transform_f);
  dh::XGBCachingDeviceAllocator<char> alloc;
  thrust::copy_if(
      thrust::cuda::par(alloc), transform_iter, transform_iter + batch.Size(),
      thrust::device_pointer_cast(data.data()), IsValidFunctor(missing));
}
```

I wonder if the copy_if is responsible for the extra memory allocation.

RAMitchell:
Thrust is using n words to store predicates for the output location of copy_if: https://github.com/thrust/thrust/blob/94e0b1c63fadd5c574c45f05267e45ac136e78ef/thrust/system/detail/generic/copy_if.inl#L64

We can write this more efficiently by redirecting the output of an exclusive scan purely with iterators.

Does someone want to try this? Otherwise I can implement myself.

trivialfis:
Let me give a try.

trivialfis:
I don't think this is necessary anymore.  @hcho3 Thanks for bisecting the code.

PR #6150, opened 2020-09-23
Fall back to CUB allocator if RMM memory pool is not set up

hcho3:
Follow-up to #6131. 

Context:
Previously, the CUB's caching allocator was used in performance-critical loops, to avoid the overhead of `cudaMalloc()`. #6131 removed the CUB allocator with RMM, since RMM's memory pool was conflicting with the CUB allocator pool and causing OOM. 
The pull request is now causing another problem: when RMM is configured without a memory pool, training is significantly slower, as the performance-critical loops are now using plain `cudaMalloc()` instead of the CUB's caching allocator.

Fix: Detect whether RMM is configured with a memory pool. If so, `XGBCachingDeviceAllocatorImpl` will use RMM's memory pool. If not, `XGBCachingDeviceAllocatorImpl` will use CUB's caching allocator.


src/common/device_helpers.cuh
```diff
@@ -402,57 +403,69 @@ struct XGBDefaultDeviceAllocatorImpl : XGBBaseDeviceAllocator<T> {
...
+  }
+#if defined(XGBOOST_USE_RMM) && XGBOOST_USE_RMM == 1
+  XGBCachingDeviceAllocatorImpl()
+      : SuperT(rmm::mr::get_current_device_resource(), cudaStream_t{nullptr}),
+        use_cub_allocator_(true) {
+    char* symbol = abi::__cxa_demangle(typeid(*SuperT::resource()).name(),
+                                       nullptr, nullptr, nullptr);
+    CHECK(symbol);
+    std::string t{symbol};
+    if (t.find("pool_memory_resource") != std::string::npos) {
```
harrism, line 461: It may cause your users frustration if "pool_memory_resource" is the only resource you recognize if they have carefully constructed a resource that works better than a pool_memory_resource for their workflow.  RMM has composable resources, so for example a user might be using a `logging_resource_adapter` wrapped around a `binning_memory_resource` (or upcoming `arena_memory_resource`) with multiple different other resources in use for its bins.

hcho3: @harrism Do `binning_memory_resource` or `arena_memory_resource` allocate a large chunk of memory ahead of time, like `pool_memory_resource`?

hcho3:
```suggestion
    if (t.find("pool_memory_resource") != std::string::npos || t.find("binning_memory_resource") != std::string::npos
          || t.find("arena_memory_resource") != std::string::npos ) {
```

@harrism We can exhaustively list all the allocators that have comparable performance as a pool allocator. It suffices that the full type string includes any of them as substring.

harrism: They are fast suballocators (depending on the pattern, faster than pool_mr), if that's what you are asking. 

My point is that you cannot exhaustively list them all, because they are composable.

hcho3: > you cannot exhaustively list them all, because they are composable.

The symbol string is a fully qualified class name, such as:
```
rmm::mr::logging_resource_adapter<rmm::mr:: binning_memory_resource>
```
so checking for the substring `binning_memory_resource` will trigger the condition.

src/common/device_helpers.cuh
```diff
@@ -402,57 +403,69 @@ struct XGBDefaultDeviceAllocatorImpl : XGBBaseDeviceAllocator<T> {
...
+      GetGlobalCachingAllocator().DeviceFree(ptr.get());
+    } else {
+      SuperT::deallocate(ptr, n);
+    }
+  }
+#if defined(XGBOOST_USE_RMM) && XGBOOST_USE_RMM == 1
+  XGBCachingDeviceAllocatorImpl()
+      : SuperT(rmm::mr::get_current_device_resource(), cudaStream_t{nullptr}),
+        use_cub_allocator_(true) {
+    char* symbol = abi::__cxa_demangle(typeid(*SuperT::resource()).name(),
```
trivialfis, line 457: Memory leak.  We need to call free after using `symbol`.

src/common/device_helpers.cuh
```diff
@@ -402,57 +403,69 @@ struct XGBDefaultDeviceAllocatorImpl : XGBBaseDeviceAllocator<T> {
...
 #endif  // defined(XGBOOST_USE_RMM) && XGBOOST_USE_RMM == 1
 };
 
-#if defined(XGBOOST_USE_RMM) && XGBOOST_USE_RMM == 1
-template <typename T>
-using XGBCachingDeviceAllocatorImpl = XGBDefaultDeviceAllocatorImpl<T>;
-#else
 /**
- * \brief Caching memory allocator, uses cub::CachingDeviceAllocator as a back-end and logs
- * allocations if verbose. Does not initialise memory on construction.
```
trivialfis, line 419: This note is still useful,  I bumped into bugs because `dh::caching_device_vector` doesn't initialize memory.

hcho3: Fixed.

src/common/device_helpers.cuh
hcho3: Fixed.

src/common/device_helpers.cuh
```diff
@@ -402,57 +403,71 @@ struct XGBDefaultDeviceAllocatorImpl : XGBBaseDeviceAllocator<T> {
...
+    if (t.find("pool_memory_resource") != std::string::npos
+        || t.find("binning_memory_resource") != std::string::npos
+        || t.find("arena_memory_resource") != std::string::npos) {
+      use_cub_allocator_ = false;
+    }
   }
-};
 #endif  // defined(XGBOOST_USE_RMM) && XGBOOST_USE_RMM == 1
+ private:
+  bool use_cub_allocator_;
```
trivialfis, line 469: Is this an uninitialized variable when rmm is not used?

hcho3: The constructor initializes it to `true`.

hcho3: Actually, I forgot that the constructor only kicks in when RMM plugin is turned on. Let me fix that.

hcho3: Fixed in the latest commit.

hcho3:
@trivialfis Can I get another round of review?

trivialfis:
I will run some simple benchmarks today.

trivialfis (approved):

PR #6162, opened 2020-09-24
Add back support for scipy.sparse.coo_matrix

hcho3:
See https://discuss.xgboost.ai/t/sparse-matrix-no-longer-supported-in-scikit-learn/1850?u=hcho3. Previous versions of XGBoost used to support the COO sparse matrix but the latest (1.2.0) does not support it.

This pull request restores the support.


trivialfis (approved):

trivialfis:
I think in the old code, we have procedure for converting unknown data to csr using scipy.  I left that code out in the last refactor.

Issue #5875, opened 2020-07-08
Compatibility with shap package


RAMitchell:
Changes to xgboost binary serialisation in 1.1 broke integration with the shap package. See https://github.com/slundberg/shap/issues/1302

This issue tracks the problem from the xgboost side. I have proposed shap migrates to json serialisation and then we add shap tests to our CI.

PR #6163, opened 2020-09-25
Update GPUTreeshap

RAMitchell:
- Enable feature interactions

- Reduce python test time

- Add integration test with shap library

### DIFF
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -553,7 +553,7 @@ class GPUPredictor : public xgboost::Predictor {
   }
 
   void PredictContribution(DMatrix* p_fmat,
-                           std::vector<bst_float>* out_contribs,
+                           HostDeviceVector<bst_float>* out_contribs,
                            const gbm::GBTreeModel& model, unsigned ntree_limit,
                            std::vector<bst_float>* tree_weights,
                            bool approximate, int condition,
@@ -564,6 +564,7 @@ class GPUPredictor : public xgboost::Predictor {
     }
 
     dh::safe_cuda(cudaSetDevice(generic_param_->gpu_id));
+    out_contribs->SetDevice(generic_param_->gpu_id);
     uint32_t real_ntree_limit =
         ntree_limit * model.learner_model_param->num_output_group;
     if (real_ntree_limit == 0 || real_ntree_limit > model.trees.size()) {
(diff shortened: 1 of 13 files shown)


src/predictor/gpu_predictor.cu
```diff
@@ -602,13 +602,74 @@ class GPUPredictor : public xgboost::Predictor {
...
-          phis.data().get() + batch.base_rowid * contributions_columns);
+          phis.data().get() + batch.base_rowid * contributions_columns, phis.size());
     }
     dh::safe_cuda(cudaMemcpy(contribs.data(), phis.data().get(),
                                   sizeof(float) * phis.size(),
                                   cudaMemcpyDefault));
   }
 
+  void PredictInteractionContributions(DMatrix* p_fmat,
+                                       std::vector<bst_float>* out_contribs,
```
trivialfis, line 613: Should we change to host device vector instead?

src/predictor/gpu_predictor.cu
```diff
@@ -602,13 +602,74 @@ class GPUPredictor : public xgboost::Predictor {
...
+                                       const gbm::GBTreeModel& model,
+                                       unsigned ntree_limit,
+                                       std::vector<bst_float>* tree_weights,
+                                       bool approximate) override {
+    if (approximate) {
```
trivialfis, line 618: We need dispatching in gbm get predictor.

src/predictor/gpu_predictor.cu
```diff
@@ -602,13 +602,74 @@ class GPUPredictor : public xgboost::Predictor {
...
+      batch.data.SetDevice(generic_param_->gpu_id);
+      batch.offset.SetDevice(generic_param_->gpu_id);
+      SparsePageView X(batch.data.DeviceSpan(), batch.offset.DeviceSpan(),
+                       model.learner_model_param->num_feature);
+      gpu_treeshap::GPUTreeShapInteractions(
+          X, device_paths.begin(), device_paths.end(), ngroup,
+          phis.data().get() + batch.base_rowid * contributions_columns, phis.size());
+    }
+    dh::safe_cuda(cudaMemcpy(out_contribs->data(), phis.data().get(),
+                                  sizeof(float) * phis.size(),
```
trivialfis, line 669: Format.

tests/python/test_with_shap.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+import numpy as np
+import xgboost as xgb
+import testing as tm
+import unittest
+import pytest
+
+try:
```
trivialfis, line 8: Can we move this into testing.py?

tests/python/test_with_shap.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
```
trivialfis, line 1: Do we need this for Python 3?

tests/python/test_with_shap.py
RAMitchell: I prefer it to stay here, I don't think it will be reused.

src/predictor/gpu_predictor.cu
RAMitchell: There might be of benefit to explicitly failing, maybe the user is expecting the algorithm to use GPU and it is silently switching to CPU.

src/predictor/gpu_predictor.cu
RAMitchell: Done

trivialfis:
Can we close https://github.com/dmlc/xgboost/issues/5875 by this PR?

PR #6181, opened 2020-09-29
Replace some uses of `dh::MakeTransformIterator`.

trivialfis:
* It has unexpected behaviors with other thrust algorithms.


RAMitchell:
You can write a functor if it doesn't compile for windows.

PR #6182, opened 2020-09-29
Use default allocator in sketching.

trivialfis:
* Revert RMM allocator detection on caching allocator.
* Use `thrust::device_vector` for sketching, the perf doesn't seem to have any visible change.

This means even if compiled with RMM, XGBoost will continue to allocate a small runtime memory outside of RMM pool.  This seems to be the cleanest way to proceed.  Other options are discussed in https://github.com/rapidsai/rmm/issues/583 .

cc @hcho3 @RAMitchell


RAMitchell (commented):
The only problem with this fix is making sure we don't break it again in future. This means making sure caching allocators and dh::TemporaryArray are only used for small constant sized allocations across the entire code base.

RAMitchell (approved):

PR #6183, opened 2020-09-29
[Breaking] Change default evaluation metric for classification to logloss / mlogloss

### LINKED ISSUE #6070: Change default eval_metric for binary:logistic objective + add warning for missing eval_metric when early stopping is enabled
I stumbled over the default metric of the `binary:logistic` objective. It seems to be 1-accuracy, which is a rather unfortunate choice. Accuracy is not even a proper scoring rule, see e.g. [Wiki](https://en.wikipedia.org/wiki/Scoring_rule#Proper_scoring_rules).

In my view, it should be "logloss", which is a strictly proper scoring rule in estimating the expectation under the binary objective.

-- Edit

The problem occurs with early stopping without manually setting the `eval_metric`. The default evaluation metric should at least be a strictly consistent scoring rule. 

I am using R with XGBoost version `1.1.1.1`.

lorentzenchr:
Closes #6070.
Change default evaluation metric for binary classification to `"logloss"`, and for muli-class classification to "`mlogloss`".

### DIFF
--- a/R-package/tests/testthat/test_callbacks.R
+++ b/R-package/tests/testthat/test_callbacks.R
@@ -26,7 +26,8 @@ watchlist <- list(train = dtrain, test = dtest)
 
 err <- function(label, pr) sum((pr > 0.5) != label) / length(label)
 
-param <- list(objective = "binary:logistic", max_depth = 2, nthread = 2)
+param <- list(objective = "binary:logistic", eval_metric: "error",
+              max_depth = 2, nthread = 2)
 
 
 test_that("cb.print.evaluation works as expected", {
@@ -105,7 +106,8 @@ test_that("cb.evaluation.log works as expected", {
 })
 
 
-param <- list(objective = "binary:logistic", max_depth = 4, nthread = 2)
+param <- list(objective = "binary:logistic", eval_metric: "error",
+              max_depth = 4, nthread = 2)
 
 test_that("can store evaluation_log without printing", {
   expect_silent(
--- a/R-package/tests/testthat/test_callbacks.R
+++ b/R-package/tests/testthat/test_callbacks.R
@@ -26,7 +26,7 @@ watchlist <- list(train = dtrain, test = dtest)
 
 err <- function(label, pr) sum((pr > 0.5) != label) / length(label)
 
-param <- list(objective = "binary:logistic", eval_metric: "error",
+param <- list(objective = "binary:logistic", eval_metric = "error",
               max_depth = 2, nthread = 2)
 
 
@@ -106,7 +106,7 @@ test_that("cb.evaluation.log works as expected", {
 })
 
 
-param <- list(objective = "binary:logistic", eval_metric: "error",
+param <- list(objective = "binary:logistic", eval_metric = "error",
               max_depth = 4, nthread = 2)
 
 test_that("can store evaluation_log without printing", {
--- a/R-package/tests/testthat/test_callbacks.R
+++ b/R-package/tests/testthat/test_callbacks.R
@@ -238,7 +238,7 @@ test_that("early stopping xgb.train works", {
 test_that("early stopping using a specific metric works", {
   set.seed(11)
   expect_output(
-    bst <- xgb.train(param, dtrain, nrounds = 20, watchlist, eta = 0.6,
+    bst <- xgb.train(param[-2], dtrain, nrounds = 20, watchlist, eta = 0.6,
                      eval_metric = "logloss", eval_metric = "auc",
                      callbacks = list(cb.early.stop(stopping_rounds = 3, maximize = FALSE,
                                                     metric_name = 'test_logloss')))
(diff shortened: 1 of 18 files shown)


hcho3:
This pull request does not add a warning for performing early stopping with a default metric. So we should keep the issue open.

hcho3:
Let me know if you need help.

hcho3:
@lorentzenchr You should update this line too:
https://github.com/dmlc/xgboost/blob/444131a2e6efd21825b10074e62d2b33735058e6/plugin/example/custom_obj.cc#L58-L60
This should fix the error in the Google C++ test. (We use Google Test framework, so hence the name.)

Also, many unit tests for the R package have assertions that checks the default evaluation metrics, so they need to be updated as well. For example:
https://github.com/dmlc/xgboost/blob/444131a2e6efd21825b10074e62d2b33735058e6/R-package/tests/testthat/test_basic.R#L18-L21
There are many of such occurrences in other R unit tests.

hcho3:
@lorentzenchr 

To make it easier, let's just throw a warning whenever no `eval_metric` is explicitly set, regardless of whether early stopping is enabled. This way, we only have to change a single place in the codebase, as follows:
https://github.com/dmlc/xgboost/blob/dda9e1e4879118738d9f9d5094246692c0f6123c/src/learner.cc#L1033-L1036

If you think this is a good idea, I'll add a commit to this pull request.

lorentzenchr:
@hcho3 Now, it is starting to look good. Thanks a lot for your help!

src/learner.cc
```diff
@@ -1031,6 +1031,18 @@ class LearnerImpl : public LearnerIO {
     std::ostringstream os;
     os << '[' << iter << ']' << std::setiosflags(std::ios::fixed);
     if (metrics_.size() == 0 && tparam_.disable_default_eval_metric <= 0) {
+      auto warn_default_eval_metric = [](const std::string& objective, const std::string& before,
```
trivialfis, line 1034: Can we place this warning inside objective function `DefaultEvalMetric`?  Also, are we sure this warning is only emitted once during training?

hcho3: I don't think moving the warning to `ObjFunction::DefaultEvalMetric()` is a good idea, since then the warning would appear in multiple places. I'd like to keep it here so that we can easily remove it later.

> Also, are we sure this warning is only emitted once during training?

Yes. The warning is emitted just before the default evaluation metric gets added to the vector `metrics_`. Once the default metric is in `metrics_`, the warning will not be thrown.

hcho3:
@mayer79  Can you review?

R-package/tests/testthat/test_callbacks.R
```diff
@@ -236,7 +238,7 @@ test_that("early stopping xgb.train works", {
 test_that("early stopping using a specific metric works", {
   set.seed(11)
   expect_output(
-    bst <- xgb.train(param, dtrain, nrounds = 20, watchlist, eta = 0.6,
+    bst <- xgb.train(param[-2], dtrain, nrounds = 20, watchlist, eta = 0.6,
```
hcho3, line 241: What is this for? @lorentzenchr

mayer79:
@hcho3 and @lorentzenchr : I had a look at the changes. They lgtm. I was wondering about the related objective 
"binary:logitraw", which uses "auc" as default. Consistently, this should be "logloss" as well. What do you think?

hcho3:
I prefer to keep the default for binary:logitraw. Is AUC not a proper scoring metric?

R-package/tests/testthat/test_callbacks.R
lorentzenchr: param[2] sets eval_metric. I exclude it here, because it is already specified by kwarg. Otherwise it throws a warnung and the test fails. 

Should I add a comment or alternatively remove the kwarg?

lorentzenchr:
> I prefer to keep the default for binary:logitraw. Is AUC not a proper scoring metric?

Unfortunately, not at all. There is no property/functional of the target distribution that is consistently estimated my maximizing AUC (or nobody has found one yet:smirk:).

mayer79:
> I prefer to keep the default for binary:logitraw. Is AUC not a proper scoring metric?

If I am not wrong, binary:logitraw trains the same model as binary:logistic, just without backtransforming the prediction to probability scale?

R-package/tests/testthat/test_callbacks.R
hcho3: Got it. Let's keep it there.

hcho3:
Yes, binary:logitraw produces models with logit output. Is logloss a proper scoring rule when output is logit?

hcho3:
FYI, I found this link that calls AUC a "semi-proper" scoring rule:  https://stats.stackexchange.com/questions/339919/what-does-it-mean-that-auc-is-a-semi-proper-scoring-rule

lorentzenchr:
@hcho3 Although I find it a bit inconsistent to have different default values of `eval_metric` for `binary:logitraw` and `binary:logistic` (and `reg:logistic`), most use cases are covered with this PR changing the default only for `binary:logistic`. Let's move forward?

hcho3:
@lorentzenchr Sure, we can always come back to `binary:logitraw` later.

Issue #6189, opened 2020-09-30
Reduce compiler warnings for C++


hcho3:
Currently, the C++ compiler generates lots of warnings: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/master/516/pipeline/61. It would be great to reduce the number of warnings.

**Note to new contributors**: Post a comment here if you'd like to work on this issue. Feel free to ping me for help.

sgavil:
Hello, i would like to work on this issue!

hcho3:
@sgavil Great! I'm going to assign this issue to you. Keep in mind that you don't have to eliminate all warnings in a single pull request. It will probably take several pull requests to address all warnings.

hcho3:
See instructions for building XGBoost here: https://xgboost.readthedocs.io/en/latest/build.html#building-on-linux-distributions. Modify the CMake flags as follows:
```
cmake .. -DHIDE_CXX_SYMBOLS=ON
```
in order to show all compiler warnings.

sgavil:
Great! Thank you

vcarpani:
I can also contribute to this, I believe there are enough warnings for 2 different PRs

igormp:
Mind if I jump into that too? GPU-related code has way too many warnings, would be nice to clean up some.

hcho3:
@igormp Great! Yes, your pull request is welcome! Keep in mind that some compiler warnings in the GPU code come from [CUB](https://github.com/nvidia/cub) and [Thrust](https://github.com/nvidia/thrust), so you may also consider submitting pull request there.

igormp:
I've noticed that most warnings (I'd say half of those) come from [dmlc](https://github.com/dmlc/dmlc-core), so maybe some PRs would be welcome there.

hcho3:
@igormp Yes, the compiler warning has been an issue for a while in dmlc. Would you like to take a shot?

igormp:
Sure! It'll try to go through it during this week.

BurakBarlas:
Hi, Can I try to help you in this issue?

hcho3:
@BurakBarlas Yes! Feel free to submit a pull request. Let me know if you need help.

pranav-verma001:
Hello i want to work on this...

trivialfis:
[...]

PR #6197, opened 2020-10-04
Reduced some C++ compiler warnings

sgavil:
Some warnings removed #6189


hcho3:
Syncing this pull request with the latest `master` branch, to fix the test farm.

hcho3:
Thank you for your contribution, and congrats for your first pull request here! I will review once all unit tests complete.

trivialfis (approved):
Thanks for the contribution!  Please rebase to latest master branch.

src/tree/updater_quantile_hist.h
```diff
@@ -442,7 +442,7 @@ class QuantileHistMaker: public TreeUpdater {
     // list of nodes whose histograms would be built explicitly.
     std::vector<ExpandEntry> nodes_for_explicit_hist_build_;
 
-    enum DataLayout { kDenseDataZeroBased, kDenseDataOneBased, kSparseData };
+    enum class DataLayout { kDenseDataZeroBased, kDenseDataOneBased, kSparseData };
```
hcho3, line 445: This line caused compilation to fail. Why was `DataLayout` changed to an enum class? Does it address a compiler warning?

hcho3 (changes requested):

src/tree/updater_quantile_hist.h
sgavil: Yes, it was producing an enum type unscoped warning. I changed the lines where I saw it was used like this: 

https://github.com/dmlc/xgboost/pull/6197/files/b8b00f57582b9e98e28a0506d58011810cbe1762#diff-10aac201acca00858b390ff0d679434fL1239-R1239

hcho3: Got it. Can you fix the failing build for `testxgboost` (C++ tests with Google Test framework)? You can build it by following instructions in https://xgboost.readthedocs.io/en/latest/contrib/unit_tests.html#running-gtest.

hcho3: @sgavil How it is going? Do you need some help?

sgavil: @hcho3 Well, I had to do some refactoring in the names of the unit tests (tbh not sure if I'm able to do that) from example from `ASSERT_EQ(this->data_layout_, RealImpl::kSparseData);` to `ASSERT_EQ(this->data_layout_, RealImpl::DataLayout::kSparseData);` but that belongs to [dmlc-core](https://github.com/dmlc/dmlc-core) and I'm not sure how to proceed with that.

tests/cpp/c_api/test_c_api.cc
```diff
@@ -128,15 +128,15 @@ TEST(CAPI, JsonModelIO) {
   learner->UpdateOneIter(0, p_dmat);
   BoosterHandle handle = learner.get();
 
-  std::string modelfile_0 = tempdir.path + "/model_0.json";
+  std::string modelfile_0 = tempdir.path_ + "/model_0.json";
```
hcho3, line 131: Let's not rename the member fields here, as it's outside the scope of this pull request (reduce compiler warning).

hcho3 (changes requested):

src/tree/updater_quantile_hist.cc
```diff
@@ -1138,9 +1138,9 @@ void QuantileHistMaker::Builder<GradientSumT>::FindSplitConditions(
     // split_cond = -1 indicates that split_pt is less than all known cut points
     CHECK_LT(upper_bound,
              static_cast<uint32_t>(std::numeric_limits<int32_t>::max()));
-    for (uint32_t i = lower_bound; i < upper_bound; ++i) {
-      if (split_pt == gmat.cut.Values()[i]) {
-        split_cond = static_cast<int32_t>(i);
+    for (uint32_t bound = lower_bound; bound < upper_bound; ++bound) {
+      if (split_pt == gmat.cut.Values()[bound]) {
+        split_cond = static_cast<int32_t>(bound);
```
hcho3, line 1143: Why are we renaming the loop variable here? The name `i` is commonly used as a loop variable, and there's no need to rename it here. I also don't think the renaming fits the purpose of this pull request (reduce compiler warning).

hcho3:
@sgavil To my knowledge, a C++ compiler does not throw a warning about naming convention. Please remove all variable renaming, unless it's necessary for removing a compiler warning.

src/tree/updater_quantile_hist.cc
sgavil: It does cause a warning because it is inside another foor loop which has declared another `i`as iterator

hcho3: Got it. We can keep this one then.

hcho3 (approved):

PR #6201, opened 2020-10-05
Loop over copy_if

RAMitchell:
Loop over batches with copy_if in DMatrix construction, because we get an int overflow with thrust. See: https://github.com/NVIDIA/thrust/issues/1302

### DIFF
--- a/src/data/simple_dmatrix.cu
+++ b/src/data/simple_dmatrix.cu
@@ -35,6 +35,7 @@ void CountRowOffsets(const AdapterBatchT& batch, common::Span<bst_row_t> offset,
       thrust::device_pointer_cast(offset.data()));
 }
 
+namespace {
 template <typename AdapterBatchT>
 struct COOToEntryOp {
   AdapterBatchT batch;
@@ -43,6 +44,7 @@ struct COOToEntryOp {
     return Entry(e.column_idx, e.value);
   }
 };
+}  // namespace
 
 // Here the data is already correctly ordered and simply needs to be compacted
 // to remove missing data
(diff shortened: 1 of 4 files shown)


src/data/simple_dmatrix.cu
```diff
@@ -35,24 +35,38 @@ void CountRowOffsets(const AdapterBatchT& batch, common::Span<bst_row_t> offset,
...
-  thrust::transform_iterator<decltype(transform_f), decltype(counting), Entry>
-      transform_iter(counting, transform_f);
   dh::XGBCachingDeviceAllocator<char> alloc;
-  thrust::copy_if(
-      thrust::cuda::par(alloc), transform_iter, transform_iter + batch.Size(),
-      thrust::device_pointer_cast(data.data()), IsValidFunctor(missing));
+  COOToEntryOp<decltype(batch)> transform_op{batch};
+  thrust::transform_iterator<decltype(transform_op), decltype(counting)>
+      transform_iter(counting, transform_op);
+  // We loop over batches because thrust::copy_if cant deal with sizes > 2^31
```
trivialfis, line 58: Should we put this into device_helpers.cuh and replace some other `copy_if`?

src/data/simple_dmatrix.cu
```diff
@@ -35,24 +35,38 @@ void CountRowOffsets(const AdapterBatchT& batch, common::Span<bst_row_t> offset,
       thrust::device_pointer_cast(offset.data()));
 }
 
+template <typename AdapterBatchT>
```
trivialfis, line 38: I prefer anonymous namespace for these helper functors.

src/data/simple_dmatrix.cu
RAMitchell: I'd rather do this later.

trivialfis (approved):

PR #6204, opened 2020-10-06
[Doc] Add info on GPU compiler

igormp:
Add note about the required compiler version for CUDA. Also added a link that gives a short explanation on compute capability version


hcho3 (approved):
Thank you! LGTM

trivialfis (approved):

PR #6207, opened 2020-10-06
Fix ellpack transform iter

trivialfis:
Rework: https://github.com/dmlc/xgboost/pull/6181

I tested it with chunk size greater than MAX_INT on Linux with CUDA 11.  I also compiled it on Windows with CUDA 10.2 (but no test).


trivialfis:
Still having issue.

PR #6212, opened 2020-10-07
Introducing DPC++-based plugin (updater) supporting oneAPI programming model

vepifanov:
Continuation of the original PR: [#5659](https://github.com/dmlc/xgboost/pull/5659)

Updater target is updater_quantile_hist.

Following process intended:
Start with existing implementation of grow_quantile_histmaker and continuously replace OpenMP parts into DPC++ kernels.
As soon as we get whole flow transformed into DPC++ kernels, start analyzing potential optimizations.

Current partial stoppers:
Data passing between updater/prediction/getGradient calls is wrapped into HostDeviceVector which now supports only CPU/CUDA memory, so additional host-device transfers are required on each iteration.


RAMitchell:
4000 lines of code is huge. It looks like the implementation largely duplicates `updater_quantile_hist.cc`. I would suggest finding a way to dispatch separate backends without duplicating the core algorithm structure.

Even having this is a plug-in could be problematic for us. If we refactor any parts of `updater_quantile_hist.cc` or xgboost in general it could cause us to maintain this code base due to failing tests.

hcho3:
Adding to @RAMitchell, it will take substantial refactoring to integrate oneAPI code into the mainline codebase. Mainly, we need to design a unified interface to dispatch to multiple backends.

Historical lesson: the GPU algorithm started out as a plugin. It took a series of refactoring before the GPU backend was integrated into the main codebase.

vepifanov:
> 4000 lines of code is huge. It looks like the implementation largely duplicates updater_quantile_hist.cc.
Yes, I tried to reuse as much of the existing code as possible. It allows to be conformant with existing cpu branch (in terms of available parameters and their usage) and be less dependent on changes in master code.

Generally, some way of dispatching between devices backends inside core algorithm structure instead of having few independent "algorithms" for each target device was our goal too. It was rather difficult to do this from a plugin, so we did an independent "algorithm" as a proof of concept. Now we can look at differences between CPU, GPU(CUDA) and OneAPI branches to come up with some universal algorithm structure which will allow dispatching into particular device kernels.

RAMitchell:
Ok, happy to leave this PR open as a proof of concept and for further discussion, just want to be clear that we are unlikely to merge something this large.

vepifanov:
@RAMitchell, @hcho3 

We continued development of oneAPI plugin inside our fork and until that time we were focused on productization: additional functionality coverage, tests, bug fixing, some optimizations, etc.
Now we have time to return to merging oneAPI functionality into master in some way.

Previously we discussed that the right approach is to make some unified flow with dispatching to different backends to avoid code duplication.
It is a good plan for long term, but I believe it will be hard to achieve without feedbacks from all sides.

To simplify review process and discussion, I will split plugin changes into separate PRs, that I will add in the nearest future.
I tried to reduce code duplication where it is possible, but there are some places that still looks similar to existing CPU code.
This similarities will probably highlight places, that can be slightly reworked during unification to satisfy needs of different backends.

First part of changes: [6814](https://github.com/dmlc/xgboost/pull/6814)

vepifanov:
Second part of changes, containing histogram computation kernels: [2](https://github.com/vepifanov/xgboost/pull/2)

Currently, it is created as a pull request between different branches of my fork to present only the necessary part of changes and avoid reviewing the same code twice.
This new branch contain all changes from previous PR and the new part on top of it.
In case there are no objections, I'm going to follow this approach for next parts of the plugin.
So eventually, if the review will go well, I will recreate PR from the last branch, containing all changes together.

vepifanov:
Third part of changes, containing fixes for broken predictor APIs compatibility and additional objective function: [part 3](https://github.com/vepifanov/xgboost/pull/3)

vepifanov:
Fourth and final part of changes, containing updater implementation, python tests, some bugfixes and fixes of cmake files: [part 4](https://github.com/vepifanov/xgboost/pull/4)

Here is the full list of sequential pull requests. Each consecutive PR add some stuff to the previous one to simplify review process:
 - [part 1](https://github.com/dmlc/xgboost/pull/6814) - data structures based on USM memory for device usage;
 - [part 2](https://github.com/vepifanov/xgboost/pull/2) - histogram computation kernels;
 - [part 3](https://github.com/vepifanov/xgboost/pull/3) - predictor fixes, additional objective function;
 - [part 4](https://github.com/vepifanov/xgboost/pull/4) - updater implementation, python tests, cmake fixes;

Eventually, if everything goes well, I will merge these PRs (4->3->2->1) and then first PR to mainline will contain all the changes.

RAMitchell:
Some general thoughts on this. Appreciate the separation of PRs, it makes it easier to see what's going on. In most places however it is still just duplicating code but replacing OpenMP calls with sycl. As a plug-in it has a reduced burden for us compared to merging to the main code base, however it can still cause problems as changes to internal data structures or apis will cause CI failures in the plug in that we have to resolve. Potentially we could merge it but make CI tests optional, that way it would be your responsibility to keep up with internal changes?

Is it fair to say that this work is targeted at intel GPUs? How available are these to end users?

napetrov:
@RAMitchell  - this is overall DPC++ based support that assume that other hardware vendors can be enabled with same code. This although have certain requirements such as compiler support, Level Zero, etc.  For now only Intel GPUs have this covered fully.  There is certain work was done on NVidia support by  [Codeplay](https://www.codeplay.com/portal/news/2020/02/03/codeplay-contribution-to-dpcpp-brings-sycl-support-for-nvidia-gpus.html), however we have not validate this scenario for XGBoost.  Assuming those details - it's mostly correct to say this focused on Intel GPUs. 

On availability - there is support for integrated graphics starting from GEN9 which is pretty mass avaialbe for last 2-3 years. And in future there is expansion of Intel Xe graphics across different segments.

vepifanov:
Main issue with replacing OpenMP with Sycl calls is that there are some auxiliary structures spread across the updater which needs to be reimplemented/redesigned to use with new device code. Possible approach could be taking out different stages of training flow (histogram computation, splits, partitioning, etc.) into separate classes with common interface, but different implementations for CPU/GPU/oneAPI.

I tried to reduce amount of dependencies on the existing code. For now, it is like interaction constraints and random sampling, beside the public interfaces of updater/predictor.

We are totally fine to keep up the plugin in consistent with other changes. Potentially, we can use our own internal regular CI process to track latest changes in master. However, we need to define set of branches that we should track beside master (like release branches).

RAMitchell:
Sounds good, I think the other outstanding issue here is the addition of a device id parameter, where the gpu_id parameter exists already. We do not want parallel interfaces with the cuda implementation. I'm open to having the cuda interface redesigned, as this kind of grew in an ad-hoc way as we added features. When I mean cuda interface I mean gpu_id parameter, tree_method='gpu_hist', predictor='gpu_predictor' .

If this is addressed I don't think there is much risk for us merging existing work as a plugin, despite the size. To integrate further into xgboost it will require more redesign.

vepifanov:
Sorry, I've missed previous comment and then was away from work for a few days, so the notification went deep into mailbox.

I introduced 'device_id' parameter, since it can represent device of any kind (cpu/gpu/another accelerator) and the name 'gpu_id' is not really suitable for this case in my opinion.

Overall, meanings of 'device_id' and 'gpu_id' and not really aligned now, because 'gpu_id' means index in the list of CUDA devices and 'device_id' means index in the list of SYCL devices.

One option is to use the same alias 'device_id' (or 'gpu_id') for all updaters, but change the behavior of this parameter for each particular updater.
Another options is to expand possible values of 'device_id' from just "0"/"1"/"2" to "cpu:0" / "cuda:0" / "oneapi:0". This may help while introducing unified interfaces, but this will probably require significant changes in XGBoost code. And what is more important, this will require changes in user code as well.

trivialfis:
I'm open to adding a `device_id` as a global parameter:

``` python
with xgboost.config(device_id="CUDA:0"):
    reg = xgboost.XGBRegressor()
```

We already have the config implemented for verbosity thanks to @hcho3.

trivialfis:
Could you please take a look at the notes in https://github.com/dmlc/xgboost/pull/6971 ?  It's in `gbtree.cc`.  I have planned for the future API for selecting devices and also wrote down some difficulties we currently have.  I hope we can avoid making some mistakes with the new OneAPI.

vepifanov:
Thanks for such detailed clarification in the code!

Just to double-check, "`device_id` will decide `tree_method`" means that `device_id` will decide `cpu_hist`/`gpu_hist`/`oneapi_hist` in case when `tree_method` is `hist`?
  
In current implementation of oneAPI plugin, call to Configure() checks the list of SYCL devices and if the device_id corresponds to CPU it creates cpu_updater as a backend and forward all calls to this updater.
In this case, dispatching first by device_id into oneAPI plugin and then dispatching it back to CPU updater by the device type will be a bit confusing.
Idealy, we can leave the new updater solely for oneAPI GPUs/other accelerators, and add oneAPI device type checks into the same place as the general device_id dispatching will be.

Another thing is the data location. Currently we don't have the ability to pass oneAPI GPU data for boosting (as it is for cupy or cuDF), but eventually we will try to support this as well. From my perspective, it should be similar approach to existing ELLPACK format, and maybe even some integration or re-usage of existing code for ELLPACK in DMatrix.
That means, that it will probably require the same heuristics for host-device data transfer for device dispatching.
Therefore, we can think about some generalized approach for checking the applicability of combination of updater-data-device.

Overall I think that the "`device_id` deciding the internal dispatching" is a good approach, since it simplifies the logic of device dispatching from users' perspective.
It also prevents forced introduction of additional parameters in the future due to universal name and string value.

trivialfis:
> Just to double-check, "device_id will decide tree_method" means that device_id will decide cpu_hist/gpu_hist/oneapi_hist in case when tree_method is hist?

Yes, that's my plan.  It will be a huge breaking change but I think this should be how it works in the future.

> Overall I think that the "device_id deciding the internal dispatching" is a good approach, since it simplifies the logic of device dispatching from users' perspective.

Great!  So we have come to same conclusion!

We will remain our support for `gpu_id` for a while.  While at the same time we can start working on the new `device_id`.

vepifanov:
@trivialfis 
I saw you mentioned this PR in [#7049](https://github.com/dmlc/xgboost/pull/7049).

Are there any activity on device dispatching at this point?
As far as I understand, the previous related PR ([#6971](https://github.com/dmlc/xgboost/pull/6971)) was postponed.

trivialfis:
Not yet, my priority right now is categorical data support.  Would be great if I can get some help.

PR #6213, opened 2020-10-08
Clean up warnings

igormp:
Linked to #6189


hcho3:
Currently, our test farm is down. You can run tests locally by following instructions in https://xgboost.readthedocs.io/en/latest/contrib/unit_tests.html.

include/xgboost/c_api.h
```diff
@@ -377,7 +377,7 @@ XGB_DLL int XGDMatrixFree(DMatrixHandle handle);
  * \return 0 when success, -1 when failure happens
  */
 XGB_DLL int XGDMatrixSaveBinary(DMatrixHandle handle,
-                                const char *fname, int silent);
+                                const char *fname);
```
hcho3, line 380: Please do not change the signature of any C API function, to keep backward compatibility.

igormp: Oops, I'll revert that. Is it ok to use variable attributes then to mark it as unused?

hcho3: You can make it unnamed, like: `XGB_DLL int XGDMatrixSaveBinary(DMatrixHandle handle, const char *fname, int)`

igormp:
I guess that's enough changes for a single PR.

I've been thinking, would it be a good idea to have some function overloads for the funcs with unused parameters? This way it'd be possible to transition to a cleaner API without breaking backwards compatibility.

Regarding the Jenkins failures on the windows build, I couldn't understand if that's an issue with my actual code or not, I'd be glad if someone could explain that.

hcho3:
> Regarding the Jenkins failures on the windows build

I've restarted the Windows build. I've seen Visual Studio mysteriously fail with `LINK : fatal error LNK1000: Internal error during ParseCommandLine`. I have no idea why this happens.

> would it be a good idea to have some function overloads for the funcs with unused parameters? This way it'd be possible to transition to a cleaner API without breaking backwards compatibility.

No, for two reasons:
1. The backward compatibility guarantee only applies to the functions listed in `c_api.h`. So for other functions not listed in `c_api.h`, we can simply modify them without using overloading.
2. All functions listed in `c_api.h` has `extern "C"` modifies, indicating that the functions must have a pure C linkage and that the functions must be able to be used by a C compiler. Since there is no function overloading in C, we cannot use function overloading in `c_api.h`.

igormp:
>The backward compatibility guarantee only applies to the functions listed in c_api.h. So for other functions not listed in c_api.h, we can simply modify them without using overloading.

Right, that makes sense. I didn't remove most of the parameters in the PR (and left them as unnamed) because most of those were virtual functions with an `override`, so the compiler didn't like it much when I changed the function signature.

>The usual workaround for 2) is to add a new function suffixed with Ex with new interface. See XGDMatrixSliceDMatrixEx() for example, which is a revised version of XGDMatrixSliceDMatrix(). We add a function with Ex suffix, wait a few releases, and then remove the old function. For example, XGDMatrixCreateFromCSR() function has been removed, and we only have XGDMatrixCreateFromCSREx().

That'd would be easily doable as long as we're somewhat sure that the function signature won't be changed that soon, or that those unused variables will remain like so for the foreseeable future. Should I try to work on that in this same PR?

hcho3:
> Should I try to work on that in this same PR?

For now, let us review this PR as it is.

hcho3 (approved):
LGTM. Thank you for your contribution!

trivialfis:
@igormp 

> I've been thinking, would it be a good idea to have some function overloads for the funcs with unused parameters? This way it'd be possible to transition to a cleaner API without breaking backwards compatibility.

Thanks for the good work!  So now you have went through a large portion of the code base.  If you have any suggestion on the refactoring of current code base we are thrilled to hear it.

trivialfis (approved):
We need to go through our code base, I believe reading these unused variables we can find some bugs or at least some dead code.

igormp:
When I find time during this month, I'll make sure to try to do some extra clean up :)

Issue #6217, opened 2020-10-09
xgboost.dask.train freezes on large dataset but dask_ml.xgboost.train succeeds


mwiebusch78:
After migrating my code froom dask_ml to the new xgboost.dask API the dask job frequently freezes (i.e. stops using any CPU resources but never finishes) after producing the following output:

```
/Users/wiebuschm/.pyenv/versions/latest_xgboost/lib/python3.8/site-packages/distributed/client.py:3530: RuntimeWarning: coroutine 'Client._update_scheduler_info' was never awaited
  self.sync(self._update_scheduler_info)
task [xgboost.dask]:tcp://127.0.0.1:58026 connected to the tracker
task [xgboost.dask]:tcp://127.0.0.1:58022 connected to the tracker
task [xgboost.dask]:tcp://127.0.0.1:58025 connected to the tracker
task [xgboost.dask]:tcp://127.0.0.1:58021 connected to the tracker
task [xgboost.dask]:tcp://127.0.0.1:58026 got new rank 0
task [xgboost.dask]:tcp://127.0.0.1:58022 got new rank 1
task [xgboost.dask]:tcp://127.0.0.1:58025 got new rank 2
task [xgboost.dask]:tcp://127.0.0.1:58021 got new rank 3
/Users/wiebuschm/.pyenv/versions/latest_xgboost/lib/python3.8/site-packages/distributed/client.py:4832: RuntimeWarning: coroutine 'wait_for' was never awaited
  c.close(timeout=3)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
task [xgboost.dask]:tcp://127.0.0.1:58025 connected to the tracker
task [xgboost.dask]:tcp://127.0.0.1:58025 got new rank 2
task [xgboost.dask]:tcp://127.0.0.1:58026 connected to the tracker
task [xgboost.dask]:tcp://127.0.0.1:58026 got new rank 0
task [xgboost.dask]:tcp://127.0.0.1:58021 connected to the tracker
task [xgboost.dask]:tcp://127.0.0.1:58021 got new rank 3
```

I can only reproduce the problem on fairly large datasets and it seems to be intermittent (occasionally it decides to work). When I terminate the process with CTRL-C I get the following stack trace:

```
  File "xgboost_error.py", line 48, in <module>
    booster = xgboost.dask.train(dask_client, params, dm_train, num_boost_round=50)['booster']
[...]

PR #6220, opened 2020-10-10
Catch all standard exceptions in C API.

trivialfis:
Related: https://github.com/dmlc/xgboost/issues/6217

### DIFF
--- a/src/c_api/c_api_error.h
+++ b/src/c_api/c_api_error.h
@@ -16,9 +16,18 @@
 #else  // LOG_CAPI_INVOCATION
 #define API_BEGIN() try {
 #endif  // LOG_CAPI_INVOCATION
+
 /*! \brief every function starts with API_BEGIN();
      and finishes with API_END() or API_END_HANDLE_ERROR */
-#define API_END() } catch(dmlc::Error &_except_) { return XGBAPIHandleException(_except_); } return 0;  // NOLINT(*)
+#define API_END()                                                              \
+  } catch (dmlc::Error & _except_) {                                           \
+    return XGBAPIHandleException(_except_);                                    \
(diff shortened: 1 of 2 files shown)


trivialfis:
It seems on CI environment the exception is not thrown.

hcho3:
@trivialfis Maybe CI containers have an aggressive OOM killer that kills rouge processes allocating huge memory? I suggest that you write a unit test that explicitly throws a standard exception.

hcho3:
See http://coldattic.info/post/40/. Linux kernel can send SIGKILL to a process that goes over a memory limit.

trivialfis:
> I suggest that you write a unit test that explicitly throws a standard exception.

That might be the best option for this PR.  But it defeats the purpose of catching these kind of exception: to avoid abort.  See linked issue.

Actually this is interesting to me, as now the behavior contradicts the C++, see Exceptions in: https://en.cppreference.com/w/cpp/memory/allocator/allocate

trivialfis:
Might be a fixed bug in the std library as I can pass the test instantly with gcc 9.

hcho3:
@trivialfis Linux over-commits memory requests by default, and when memory runs out, the kernels starts killing processes. So catching `std::bad_alloc` might not prevent process abortion. See https://stackoverflow.com/a/51676578.

trivialfis:
@hcho3 Can you review?  I think the PR is still useful as it does work on my platform. ;-) Besides, as we are now compiling Python package with cxx symbols hidden, so any exception will turn into an abort when reaching the C API.  This PR can capture these exceptions and provide a way to output some error messages (borderline but still better than abort).

hcho3 (approved):

Issue #6228, opened 2020-10-13
gpu_hist integer overflows and OOM issues


RAMitchell:
This issue tracks gpu_hist bugs relating to large workloads, uncovered in recent experiments on the mortgage dataset.

- Thrust copy_if has an integer overflow when n_rows*n_cols > 2^31. #6201 implements a workaround by iterating over batches. 

- Memory usage has possibly increased since version 1.0-1.1, leading to OOM on 32gb devices even with the above fix. We should do some analysis of peak memory usage over versions on a large synthetic workload, checking for regressions.

- DaskDeviceQuantileDMatrix has integer overflow bugs related to thrust::inclusive_scan, occurring when dask chunk sizes exceed 2^31.

To prevent this occurring in future we can try unit tests on large sizes, checking for overflow or memory issues. These tests need to be carefully designed to not be flaky (e.g. only run on a machine with sufficient memory) and to run quickly (<1-2 seconds).

hcho3:
@RAMitchell Can we close this now?

jamescolless:
Sorry, just to confirm, does the 2^31-1 limitation also occur in DeviceQuantileDMatrix or only in DaskDeviceQuantileDMatrix?

trivialfis:
Both.  For Dask version, the limitation is for each partition.

trivialfis:
It's fixed in latest master, the fix should go in next release.

PR #6234, opened 2020-10-14
Minimal inclusive scan

RAMitchell:
Attempt at a small inclusive scan implementation to get us around the thrust 2^31 limit (#6228).


src/data/ellpack_page.cu
```diff
@@ -206,15 +206,7 @@ void CopyDataToEllpack(const AdapterBatchT& batch, EllpackPageImpl* dst,
...
-  // 1000 as a safe factor for inclusive_scan, otherwise it might generate overflow and
-  // lead to oom error.
-  // or:
-  // after reduction step 2: cudaErrorInvalidConfiguration: invalid configuration argument
-  // https://github.com/NVIDIA/thrust/issues/1299
-  CHECK_LE(batch.Size(), std::numeric_limits<int32_t>::max() - 1000)
-      << "Known limitation, size (rows * cols) of quantile based DMatrix "
-         "cannot exceed the limit of 32-bit integer.";
-  thrust::inclusive_scan(thrust::cuda::par(alloc), key_value_index_iter,
+  dh::InclusiveScan(key_value_index_iter,
```
trivialfis, line 209: Removing the doc in Python script is also needed.

RAMitchell:
I've got a conda error on the Windows test that seems unrelated, otherwise I think this is ready to go.

python-package/xgboost/core.py
```diff
@@ -870,9 +870,6 @@ class DeviceQuantileDMatrix(DMatrix):
     You can construct DeviceQuantileDMatrix from cupy/cudf/dlpack.
 
     .. versionadded:: 1.1.0
-
```
trivialfis, line 873: Also on dask.

cmake/Utils.cmake
```diff
@@ -144,10 +144,8 @@ function(xgboost_set_cuda_flags target)
   endif (CMAKE_VERSION VERSION_GREATER_EQUAL "3.18")
 
   if (USE_DEVICE_DEBUG)
-    if (CMAKE_BUILD_TYPE MATCHES "Debug")
-      target_compile_options(${target} PRIVATE
-        $<$<COMPILE_LANGUAGE:CUDA>:-G;-src-in-ptx>)
-    endif(CMAKE_BUILD_TYPE MATCHES "Debug")
+    target_compile_options(${target} PRIVATE
+      $<$<AND:$<CONFIG:DEBUG>,$<COMPILE_LANGUAGE:CUDA>>:-G;-src-in-ptx>)
```
hcho3, line 148: Should `CMAKE_BUILD_TYPE` be compared to Debug? Do we want to use `-G` flag for Release builds?

hcho3 (commented):
LGTM overall. How do we plan to test that DDQDM works with elements more than 2^31?

cmake/Utils.cmake
RAMitchell: The generator expression $<CONFIG:DEBUG> takes care of this.

hcho3: Got it.

RAMitchell:
I have this script but it takes > 20s, most of which is in the DMatrix construction time. I don't think that's fast enough for CI.
```python
import numpy as np
import xgboost as xgb
import cupy as cp
import time


# Test for integer overflow or out of memory exceptions
def test_large_input():
    # n*m = 2^31
    n = 1000
    m = (1 << 31) // n
    m //= 2
    X = cp.ones((m, n), dtype=np.float32)
    y = cp.ones(m)
    start = time.time()
    dmat = xgb.DeviceQuantileDMatrix(X, y)
    print(time.time() - start)
    start = time.time()
    xgb.train({"tree_method": "gpu_hist", "max_depth": 1}, dmat, 1)
    print(time.time() - start)
```

RAMitchell:
I tried it again on V100 and the test takes 4s so I've tried enabling it but skipping if there is less than 15gb available device memory.

hcho3:
@RAMitchell Thanks. I've recently change the CI configuration such that each GPU worker runs only one job at a time, so you'll have access to the all 16 GB memory.

hcho3 (approved):
LGTM. Let's see if the large matrix test completes in a reasonable amount of time.

tests/python-gpu/test_large_input.py
```diff
@@ -0,0 +1,21 @@
+import numpy as np
+import xgboost as xgb
+import cupy as cp
+import time
+import pytest
+
+
+# Test for integer overflow or out of memory exceptions
+def test_large_input():
+    available_bytes, _ = cp.cuda.runtime.memGetInfo()
```
hcho3, line 10: Currently, this test will skip when `--use-rmm-pool` is set, as cuPy is not configured with RMM. To make it work with RMM, we'll need to run
```python
cupy.cuda.set_allocator(rmm.rmm_cupy_allocator)
```

Do we want to enable this test with RMM?

hcho3:
Reported CI duration:

* `test-python-gpu-cuda10.2`: 7.82s 
* `test-python-gpu-cuda11.0`: 7.64s
* `test-python-gpu-cuda11.0-cross`: 7.85s

I think we should keep the test.

tests/python-gpu/test_large_input.py
RAMitchell: Don't think this is necessary for now.

PR #6238, opened 2020-10-14
Add info about OS non-determinism to FAQs

mtournier-apixio:
This PR adds to the docs useful information in the FAQs about training deterministic XGBoost models with the same data in different Operating Systems.


trivialfis:
Actually the result might be different on different hardware/compiler/stdlib/os

PR #6249, opened 2020-10-16
[CI] Reduce testing load with RMM

hcho3:
The introduction of the RMM plugin has nearly doubled the size of the Python test suite. Reduce the burden on our CI system by excluding a few heavy unit tests when RMM is enabled (with `--use-rmm-pool` flag to Pytest)

### DIFF
--- a/tests/python-gpu/conftest.py
+++ b/tests/python-gpu/conftest.py
@@ -41,3 +41,16 @@ def local_cuda_cluster(request, pytestconfig):
 
 def pytest_addoption(parser):
     parser.addoption('--use-rmm-pool', action='store_true', default=False, help='Use RMM pool')
+
+def pytest_collection_modifyitems(config, items):
+    if config.getoption('--use-rmm-pool'):
+        blocklist = [
+            'python-gpu/test_gpu_with_dask.py::TestDistributedGPU::test_gpu_hist',
+            'python-gpu/test_gpu_demos.py::test_dask_training',
+            'python-gpu/test_gpu_prediction.py::TestGPUPredict::test_shap',
+            'python-gpu/test_gpu_linear.py::TestGPULinear'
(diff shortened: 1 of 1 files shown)


tests/python-gpu/conftest.py
```diff
@@ -41,3 +41,16 @@ def local_cuda_cluster(request, pytestconfig):
 
 def pytest_addoption(parser):
     parser.addoption('--use-rmm-pool', action='store_true', default=False, help='Use RMM pool')
+
+def pytest_collection_modifyitems(config, items):
+    if config.getoption('--use-rmm-pool'):
+        blocklist = [
+            'python-gpu/test_gpu_with_dask.py::TestDistributedGPU::test_gpu_hist',
```
trivialfis, line 48: We should keep this one.

hcho3: Do we need to run this one with RMM? This is quite expensive test.

trivialfis:
Actually I think the bottle neck might be on the setup phrase.  Adding up the top 20 tests in: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-6248/3/pipeline/207 is about 150 seconds -> 2.5 minutes.  While it took 22 minutes to finish.

hcho3:
The container setup is an amortized cost, as it occurs occasionally whenever Docker images get security updates.

tests/python-gpu/conftest.py
hcho3: Actually we can put it back, if you strongly feel we should keep it.

trivialfis: Thanks for working on this.  Let's put it back for now.  It's a full test for training.  Also it might be useful if we want to log the memory consumption.

PR #6257, opened 2020-10-19
Time the CPU tests on Jenkins.

trivialfis:
#6186 made parallel training as default.  But it seems on Jenkins thread contention is severe when used with sklearn.  I can only reproduce 1 slowdown in a demo.  So timing the tests on CI.

### DIFF
--- a/tests/python/testing.py
+++ b/tests/python/testing.py
@@ -1,5 +1,6 @@
 # coding: utf-8
 import os
+import platform
 from xgboost.compat import SKLEARN_INSTALLED, PANDAS_INSTALLED
 from xgboost.compat import DASK_INSTALLED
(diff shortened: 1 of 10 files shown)


hcho3 (approved):

hcho3:
Merging this now to unblock CI on the master branch.

tests/python/testing.py
```diff
@@ -22,6 +23,10 @@
 memory = Memory('./cachedir', verbose=0)
 
 
+def is_arm():
+    return {'condition': platform.machine().lower().find('arm') != 1,
```
hcho3, line 27: This should be `platform.machine().lower().find('arm') != -1`

PR #6260, opened 2020-10-19
Slice model

### LINKED ISSUE #4052: [feature proposal] Load best model after early stopping
Could you please implement a way to load the best model after ealy stopping? LightGBM is also doing this by default. See here: https://lightgbm.readthedocs.io/en/latest/Python-Intro.html#early-stopping

That would be a great improvement to my hyperparameter optimization workflow.

Thanks
Philip

### LINKED ISSUE #5531: Removing trees from a model
As a follow up to https://github.com/dmlc/xgboost/issues/2281, I think that it would be great if XGBoost supported the removal of trees from a model. 

It would allow the implementation (with callbacks) of adaptive learning rate strategies that need to abandon some trees before adjusting the learning rate. More generally, it would allow to implement various "ensemble pruning" methods.

trivialfis:
This PR is meant the end the confusion around `best_ntree_limit` and unify model slicing.  We have multi-class and random forests, asking users to understand how to set `ntree_limit` is difficult and error prone.

Close #5531 Close #4052.

* Implement the `save_best` option in early stopping.

### DIFF
--- a/doc/python/callbacks.rst
+++ b/doc/python/callbacks.rst
@@ -7,9 +7,9 @@ package.  In XGBoost 1.3, a new callback interface is designed for Python packag
 provides the flexiblity of designing various extension for training.  Also, XGBoost has a
 number of pre-defined callbacks for supporting early stopping, checkpoints etc.
 
-#######################
+
 Using builtin callbacks
-#######################
+-----------------------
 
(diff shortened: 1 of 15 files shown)


trivialfis:
Closing to avoid spam.

PR #6262, opened 2020-10-20
Drop single point model recovery

trivialfis:
Continue https://github.com/dmlc/xgboost/pull/6112

## RABIT parameter handling

Currently on master branch, the rabit time out handling test is only passing because of an unrelated error:

```
AssertError:buf should have data from resbuf, rabit is configured to keep process running
```

This exception has nothing to do with RABIT timeout.  Once `rabit_robust` is removed, this
assertion will no longer exist and the test will fail as it's testing timeout facility.  I
dug into it and found the parameter "rabit_timeout" is not being passed down into rabit at
all.  Tracing back the line changes, I found that #5082 has removed these lines in
`xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala`:

```
    val xgbRabitParams = xgbParamsFactory.buildRabitParams.asJava
    ...
            tracker.getWorkerEnvs().putAll(xgbRabitParams)
```
so currently non of the rabit parameter is handled correctly.  This PR adds back these 2 lines.

## Timeout

The old implementation of timeout used an async worker thread to throw the exception, which is incorrect as exception is caught by main thread.  Timeout is not trivial to implement in real application.  Killing a thread is difficult as it violates all the states shared with main thread.  In this PR I pass the timeout interval to `poll`, which serves as a stop gate.

## Finalization

The `rabit_robust` had a pseudo checkpoint operation during shutdown, which formed a global barrier.  The barrier is now removed along with `rabit_robust`.

## Socket exception

Previously allreduce checks for out of band data on socket, which is now removed.  In most of the applications, OOB data can be safely ignored.

## Other Subtleties

I admit I don't fully understand how are those tests passing before the PR.  For example, if there's a sync at shutdown, some spark failure tests should hang but they didn't.  I suspect it's due to similar reason described in section `RABIT parameter handling`.


trivialfis:
cc @chenqin

RAMitchell (approved):

tests/travis/run_test.sh
```diff
@@ -59,7 +59,7 @@ if [ ${TASK} == "python_test" ]; then
         python -m pip install ./python-package/dist/xgboost-*-py3-none-${TAG}.whl
         conda --version
         python --version
-        python -m pytest -v -s -rxXs --durations=0 --fulltrace tests/python --cov=python-package/xgboost || exit -1
+        python -m pytest -v -s -rxXs --durations=0 --fulltrace tests/python/test_tracker.py --cov=python-package/xgboost || exit -1
```
hcho3, line 62: We should plan to restore the other unit tests for ARM64.

hcho3 (approved):

tests/travis/run_test.sh
trivialfis: Reverted the changes.

PR #6263, opened 2020-10-20
Small optimization on GPU Hist.

trivialfis:
- Remove 1 copy in root node evaluation.
- Use pointer in `inclusive_scan`.
- Remove `SPAN_LT`.  Somehow this has the most significant impact.


RAMitchell (approved):
Lgtm, how much is the perf improvement and in what scenario?

trivialfis:
This has some varying effect on different cards.  On TitanX it's actually slower, on RTX8000 it has close to no effect, on GTX1080ti with higgs it gains about 8 seconds ...

PR #6266, opened 2020-10-21
Revert "Disable JSON full serialization for now."

trivialfis:
Reverts dmlc/xgboost#6248


trivialfis:
Single point model recovery is removed.  The checkpoint operation is only for recording model version now.

hcho3 (approved):

PR #6267, opened 2020-10-21
Skip dask tests on ARM.

trivialfis:


hcho3:
@trivialfis I can take it over, if you'd like.

trivialfis:
@hcho3 Thanks, I didn't update it as the jvm test is probably gonna fail anyway.

trivialfis:
Ah, dask data partitioning is not deterministic so the data  initialization test is also flaky

PR #6269, opened 2020-10-21
Deprecate LabelEncoder in XGBClassifier; Enable cuDF/cuPy inputs in XGBClassifier

### LINKED ISSUE #6232: Support GPU input in `XGBClassifier`; deprecate the use of label encoder
conda, xgboost 1.2.0 , rapids 0.14, Ubuntu 18.04 LTS, GeForce RTX 2080, cuda 10.0.

While the non-sklearn API can take cudf, the sklearn API fails with the error in the title.

Any data set works, just choose for creditcard.csv .    For the target given can choose attached dataset  
[creditcard.csv.zip](https://github.com/dmlc/xgboost/files/5374518/creditcard.csv.zip)

i.e. this works:
```
import cudf
import pynvml
import numpy as np
import pandas as pd
import xgboost as xgb
from math import sqrt
from sklearn.model_selection import KFold, StratifiedKFold, GroupKFold
from sklearn.metrics import mean_squared_error,f1_score, accuracy_score
train = cudf.read_csv("creditcard.csv")
test = cudf.read_csv("creditcard.csv")
y = train['default payment next month']
train = train.drop(['default payment next month'], axis=1)
test = test.drop(['default payment next month'], axis=1)
from cuml import train_test_split
X_train, X_test, y_train, y_test = train_test_split(train, y, test_size=0.2, random_state=123)
dtrain = xgb.DMatrix(X_train,y_train)
dval   = xgb.DMatrix(X_test, y_test)
dtest  = xgb.DMatrix(test)
evallist = [(dval, 'validation'), (dtrain, 'train')]
num_round=5000
trained_model = xgb.train(
                        {
                          'learning_rate': 0.01,
                          'colsample_bytree' : 0.25,
                          'max_depth': 10,
                          'silent': True,
                          'tree_method':'gpu_hist',
                        },
                        dtrain,num_round, evallist,verbose_eval=250)
prediction = trained_model.predict(dtest)
```

but this does not work:

```
import cudf
import pynvml
import numpy as np
import pandas as pd
import xgboost as xgb
from math import sqrt
from sklearn.model_selection import KFold, StratifiedKFold, GroupKFold
from sklearn.metrics import mean_squared_error,f1_score, accuracy_score
train = cudf.read_csv("creditcard.csv")
test = cudf.read_csv("creditcard.csv")
y = train['default payment next month']
train = train.drop(['default payment next month'], axis=1)
test = test.drop(['default payment next month'], axis=1)
from cuml import train_test_split
X_train, X_test, y_train, y_test = train_test_split(train, y, test_size=0.2, random_state=123)

params = {'learning_rate': 0.01, 'colsample_bytree' : 0.25, 'max_depth': 10, 'silent': True, 'tree_method':'gpu_hist','n_estimators': 5000, 'verbose_eval': 250}
model = xgb.XGBClassifier(**params)
eval_set = [(X_test, y_test), (X_train, y_train)]
model.fit(X_train, y_train, eval_set=eval_set)
prediction = model.predict(test)
```

fails with:

```
>>> model.fit(X_train, y_train, eval_set=eval_set)
  File "<stdin>", line 1, in <module>
  File "/home/jon/miniconda3/lib/python3.6/site-packages/xgboost/sklearn.py", line 858, in fit
    self._le = XGBoostLabelEncoder().fit(y)
  File "/home/jon/miniconda3/lib/python3.6/site-packages/sklearn/preprocessing/label.py", line 219, in fit
    y = column_or_1d(y, warn=True)
  File "/home/jon/miniconda3/lib/python3.6/site-packages/sklearn/utils/validation.py", line 749, in column_or_1d
    shape = np.shape(y)
  File "<__array_function__ internals>", line 6, in shape
  File "/home/jon/miniconda3/lib/python3.6/site-packages/cudf/core/series.py", line 734, in __array_function__
    return cudf_func(*args, **kwargs)
TypeError: 'property' object is not callable
>>> 
```

I expect I'm doing something wrong since cudf functionality has been in xgboost for a long time now and I assume many people use the sklearn API, but I cannot see what I'm doing wrong.  Thanks!

hcho3:
Closes #6232

cc @pseudotensor

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -814,10 +814,11 @@ class XGBClassifier(XGBModel, XGBClassifierBase):
                 xgb_options.update({"eval_metric": eval_metric})
 
         if use_label_encoder and not (_is_cudf_df(y) or _is_cudf_ser(y) or _is_cupy_array(y)):
-            warnings.warn('The use of label encoder in XGBClassifier is deprecated. To remove ' +
-                          'this warning, pass option use_label_encoder=False to ' +
-                          'XGBClassifier.fit(). A future release of XGBoost will *remove* label ' +
-                          'encoder from XGBClassifier.',
+            warnings.warn('The use of label encoder in XGBClassifier is deprecated and will be ' +
+                          'removed in a future release. To remove this warning, do the ' +
+                          'following: 1) Pass option use_label_encoder=False to ' +
+                          'XGBClassifier.fit(); and 2) Encode your labels (y) as integers ' +
+                          'starting with 0, i.e. 0, 1, 2, ..., [num_class - 1].',
                           UserWarning)
             self._le = XGBoostLabelEncoder().fit(y)
             label_transform = self._le.transform
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -1036,6 +1036,9 @@ class XGBClassifier(XGBModel, XGBClassifierBase):
     extra_parameters='''
     n_estimators : int
         Number of trees in random forest to fit.
+    use_label_encoder : bool
+        (Deprecated) Use the label encoder from scikit-learn to encode the labels. For new code,
+        we recommend that you set this parameter to False.
 ''')
 class XGBRFClassifier(XGBClassifier):
     # pylint: disable=missing-docstring
@@ -1044,11 +1047,13 @@ class XGBRFClassifier(XGBClassifier):
                  subsample=0.8,
                  colsample_bynode=0.8,
                  reg_lambda=1e-5,
+                 use_label_encoder=True,
                  **kwargs):
         super().__init__(learning_rate=learning_rate,
                          subsample=subsample,
                          colsample_bynode=colsample_bynode,
                          reg_lambda=reg_lambda,
+                         use_label_encoder=use_label_encoder,
                          **kwargs)
 
     def get_xgb_params(self):
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -783,16 +783,37 @@ class XGBClassifier(XGBModel, XGBClassifierBase):
             sample_weight_eval_set=None, feature_weights=None, callbacks=None):
         # pylint: disable = attribute-defined-outside-init,arguments-differ,too-many-statements
 
+        can_use_label_encoder = True
+        label_encoding_check_error = (
+                'The label must consist of integer labels of form 0, 1, 2, ..., [num_class - 1].')
+        label_encoder_deprecation_msg = (
+                'The use of label encoder in XGBClassifier is deprecated and will be ' +
+                'removed in a future release. To remove this warning, do the ' +
+                'following: 1) Pass option use_label_encoder=False when constructing ' +
+                'XGBClassifier object; and 2) Encode your labels (y) as integers ' +
+                'starting with 0, i.e. 0, 1, 2, ..., [num_class - 1].')
+
         evals_result = {}
         if _is_cudf_df(y) or _is_cudf_ser(y):
             import cupy as cp  # pylint: disable=E0401
             self.classes_ = cp.unique(y.values)
+            self.n_classes_ = len(self.classes_)
+            can_use_label_encoder = False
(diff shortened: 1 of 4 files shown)


pseudotensor:
Thanks!

python-package/xgboost/sklearn.py
```diff
@@ -801,8 +816,19 @@ def fit(self, X, y, sample_weight=None, base_margin=None,
             else:
                 xgb_options.update({"eval_metric": eval_metric})
 
-        self._le = XGBoostLabelEncoder().fit(y)
-        training_labels = self._le.transform(y)
+        if self.use_label_encoder and not (_is_cudf_df(y) or _is_cudf_ser(y) or _is_cupy_array(y)):
```
trivialfis, line 819: Why is `and not (_is_cudf_df(y) or _is_cudf_ser(y) or _is_cupy_array(y))` necessary?

trivialfis (commented):
Approve to the general idea.  See minor comment.

Also please track it in https://github.com/dmlc/xgboost/issues/3986 .

python-package/xgboost/sklearn.py
hcho3: It's because we cannot use the label encoder at all when the input is cuDF or cuPy.

trivialfis: Would it be better if we just deprecate it regardless the input:

```
if self.user_label_encoder:
    warning.warn(....)
```

hcho3: For cuDF input, there is nothing to "deprecate," since previous `XGBClassifier` was not working with a cuDF input. Or do you mean that we should always display the warning?

hcho3: We still have to check the type of the label because inserting cuDF into the label encoder will result in an error.

trivialfis: I think there are a few things we need to do.

- Document the pipeline on classification.
- Always give a warning when `use_label_encoder` is set to true.
- Add test cases when the input does not conform to our defined pipeline. (error tests)

hcho3: How should we handle `use_label_encoder=True` with cuDF input? Right now, `XGBClassifier` will silently drop the label encoder to accommodate the cuDF input. Should it be an explicit error if the user sets `use_label_encoder=True` with a cuDF input? (Note that `use_label_encoder=True` is the default setting, to avoid breaking existing code.)

hcho3: > Add test cases when the input does not conform to our defined pipeline

We can add an explicit check as to whether the list of classes (`np.unique(y)`) is actually 0, 1, ..., `[num_class]-1`, and throw an error otherwise.

trivialfis: > Should it be an explicit error if the user sets use_label_encoder=True with a cuDF input?

Yes.

> We can add an explicit check as to whether the list of classes (np.unique(y)) is actually 0, 1, ..., [num_class]-1, and throw an error otherwise.

Sounds good.

hcho3: Eh... If we throw errors for that, then the default setting (use_label_encoder=True) will result in an error for all cuDF inputs. I'd prefer keeping the PR as it currently stands, so that unsuspecting users can use fit() with cuDF inputs.

RAMitchell (approved):

python-package/xgboost/sklearn.py
trivialfis: User of XGBoost using `pandas` before but switch to cuDF might find the behavior surprising as they are not consistent.

hcho3: Got it. Let's throw an error when `use_label_encoder=True` is set and the input is cuDF.

wenfeiy-db:
Could someone explain why we made the deprecation decision for label encoder? I think most native sklearn classifiers do the label transformation internally and call inverse_transform at predict. This causes discrepancy between the xgboost sklearn interface and other sklearn classifiers

hcho3:
This was to accept cudf and cuPy arrays as X.

trivialfis:
Also to avoid saving the scikit learn object in xgboost model

PR #6275, opened 2020-10-22
Remove RABIT CMake targets.

trivialfis:
* Now it's built as part of libxgboost.
* Set correct C API error in RABIT initialization and finalization.


trivialfis:
Suggested by @wbo4958 , I just changed the `XGBoostError` to `Exception`.

hcho3 (approved):

Issue #6276, opened 2020-10-22
[jvm-packages] Flaky test for error handling.


trivialfis:
https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-6269/2/pipeline

After some digging, here are a few issues I found:

- When RABIT fail to initialize/shutdown/TrackerPrint, jvm displays the last error from xgboost instead of rabit, or abort due to lack of API guard.  This is partly addressed by https://github.com/dmlc/xgboost/pull/6275
- On the failing test `train throwing exception if feature size of dataset is different on distributed train`, tracker is shutdown before workers, so RABIT can't send the "shutdown" message to tracker and hence cannot finalize gracefully.
- Catching `XGBoostError` in the test is unreliable.  On the other hand, if we disable killing spark context, a `SparkException` is guaranteed to be thrown after a few retries.

trivialfis:
Running:
```
mvn -Dsuites='ml.dmlc.xgboost4j.scala.spark.MissingValueHandlingSuite*' test
```
is deterministically failing on my machine even went back to before merging RABIT.


```
[INFO] 
[INFO] --- scalatest-maven-plugin:1.0:test (test) @ xgboost4j-spark_2.12 ---
Discovery starting.
Discovery completed in 193 milliseconds.
Run starting. Expected test count is: 8
MissingValueHandlingSuite:
Using Spark's default log4j profile: org/apache/spark/log4j-defaults.properties
20/10/23 02:38:20 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=192.168.1.102, DMLC_TRACKER_PORT=9091, DMLC_NUM_WORKER=4}
task 1 connected to the tracker
[02:38:25] WARNING: /home/fis/Workspace/XGBoost/xgboost/src/gbm/gbtree.cc:137: Tree method is automatically selected to be 'approx' for distributed training.[02:38:25] WARNING: /home/fis/Workspace/XGBoost/xgboost/src/gbm/gbtree.cc:137: Tree method is automatically selected to be 'approx' for distributed training.
task 2 connected to the tracker
task 0 connected to the tracker
task 3 connected to the tracker
task 1 got new rank 0
task 2 got new rank 1
task 0 got new rank 2
task 3 got new rank 3
[...]

PR #6277, opened 2020-10-22
[jvm-package] Try fixing spark failure. [skip ci]

trivialfis:
Based on https://github.com/dmlc/xgboost/pull/6275 .  This is just an attempt on removing the need to handle exception in worker.


trivialfis:
Related: https://github.com/dmlc/xgboost/issues/6276

PR #6287, opened 2020-10-26
[ci] replace 'egrep' with 'grep -E'

jameslamb:
A few weeks ago [on Twitter](https://twitter.com/b0rk/status/1315831992253116417), I learned about [`shellcheck`](https://github.com/koalaman/shellcheck), a linter for shell scripts. To try to learn it tonight, I've been trying it out on a few repos that I know have a lot of shell scripts.

While running it over `xgboost` tonight, I found a minor issue that I think should be fixed: `egrep` is deprecated and can be replaced with `grep -E`.

From [the grep docs](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html) (which are linked by [the shellcheck docs for this issue](https://github.com/koalaman/shellcheck/wiki/SC2196):

> This grep has been enhanced in an upwards-compatible way to provide the exact functionality of the historical egrep and fgrep commands as well. It was the clear intention of the standard developers to consolidate the three greps into a single command.

I've struggled to find exactly when `-E` was added to `grep` but it was at least 6 years ago, so I think it's safe to make this change in the files touched by this PR.

Thanks for your time and consideration

### DIFF
--- a/tests/ci_build/Dockerfile.gpu_build
+++ b/tests/ci_build/Dockerfile.gpu_build
@@ -19,7 +19,7 @@ RUN \
 
 # NCCL2 (License: https://docs.nvidia.com/deeplearning/sdk/nccl-sla/index.html)
 RUN \
-    export CUDA_SHORT=`echo $CUDA_VERSION_ARG | egrep -o '[0-9]+\.[0-9]'` && \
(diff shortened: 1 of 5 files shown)


hcho3 (approved):
Thanks!

PR #6289, opened 2020-10-26
Update plugin instructions for CMake build

funasoul:
Fixed an example of CMake rule so that it can successfully build a plugin.


hcho3 (approved):

hcho3:
Thanks!

Issue #6293, opened 2020-10-27
Support building with cmake with unity build (build speedup)


bsergean:
Hi there,

With cmake > 3.16 there is a nice (optional) feature called unity build which help speedup building large or complex C++ projects. Here is a [write up](https://onqtam.com/programming/2019-12-20-pch-unity-cmake-3-16/) on how this works. Speed ups can be pretty large.

However tweaks are sometimes required to get it to work. Typically if you have 2 constants with the same value in 2 .cpp file, a problem happens. Variables with similar name needs to only be declared once.

Here is how I tried to build xgboost but I got those errors which I have a hard time decipher.

```
$ mkdir build ; cd build
build$ cmake -DCMAKE_UNITY_BUILD=ON -DUSE_OPENMP=OFF -DBUILD_STATIC_LIB=ON .. 
build$ 
build$ make
Scanning dependencies of target dmlc
[  7%] Building CXX object dmlc-core/CMakeFiles/dmlc.dir/Unity/unity_1_cxx.cxx.o
[ 15%] Building CXX object dmlc-core/CMakeFiles/dmlc.dir/Unity/unity_0_cxx.cxx.o
[ 23%] Linking CXX static library libdmlc.a
[ 23%] Built target dmlc
Scanning dependencies of target objxgboost
[ 30%] Building CXX object src/CMakeFiles/objxgboost.dir/Unity/unity_6_cxx.cxx.o
[ 38%] Building CXX object src/CMakeFiles/objxgboost.dir/Unity/unity_5_cxx.cxx.o
In file included from /Users/bsergeant/src/foss/github/xgboost/build/src/CMakeFiles/objxgboost.dir/Unity/unity_5_cxx.cxx:4:
/Users/bsergeant/src/foss/github/xgboost/src/predictor/predictor.cc:12:1: error: explicit specialization of 'Get' after instantiation
DMLC_REGISTRY_ENABLE(::xgboost::PredictorReg);
^
/Users/bsergeant/src/foss/github/xgboost/dmlc-core/include/dmlc/registry.h:236:47: note: expanded from macro 'DMLC_REGISTRY_ENABLE'
  Registry<EntryType > *Registry<EntryType >::Get() {                   \
                                              ^
/Users/bsergeant/src/foss/github/xgboost/src/predictor/cpu_predictor.cc:507:1: note: implicit instantiation first required here
XGBOOST_REGISTER_PREDICTOR(CPUPredictor, "cpu_predictor")
^
[...]

PR #6295, opened 2020-10-27
Support unity build.

trivialfis:
See https://github.com/dmlc/xgboost/issues/6293 for details.  The speedup may help some build machines for distribution, but I don't think it's useful for daily development.  XGBoost's CPU compilation time on my machine is usually within a minute so not much room for improvement.

Update: Will be enabled on Windows CI.

Close https://github.com/dmlc/xgboost/issues/6293 .

### DIFF
--- a/Jenkinsfile-win64
+++ b/Jenkinsfile-win64
@@ -85,7 +85,7 @@ def BuildWin64() {
     bat """
     mkdir build
     cd build
-    cmake .. -G"Visual Studio 15 2017 Win64" -DUSE_CUDA=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DGOOGLE_TEST=ON -DUSE_DMLC_GTEST=ON -DCMAKE_UNITY_BUILD=ON ${arch_flag}
+    cmake .. -G"Visual Studio 15 2017 Win64" -DUSE_CUDA=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DGOOGLE_TEST=ON -DUSE_DMLC_GTEST=ON ${arch_flag}
     """
     bat """
     cd build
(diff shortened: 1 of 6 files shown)


trivialfis:
@hcho3 I tried enabling it on Windows, see if it helps.  ;-)

hcho3:
@trivialfis Thanks. We should also try to get Ninja build working (#5832) to bring the build time under 10 min.

trivialfis:
@hcho3 It's under 10 minutes now.

hcho3:
Interesting. So unity build does help quite a bit on the Windows platform. Let me approve this PR.

hcho3 (approved):

trivialfis:
Close https://github.com/dmlc/xgboost/issues/6293 .

trivialfis:
Em, it breaks external memory on Windows somehow.

hcho3:
@trivialfis Another reason to extinguish race condition and other undefined behaviors from the external memory. This is the same issue as the Ninja PR (#5832)

hcho3:
Or can we drop support for external memory on the Windows platform?

trivialfis:
It's not that easy with external memory.  On the on hand I want to remove all use of threads for loading data, on the other hand the performance will be miserable without pre-fetching.

> drop support for external memory on the Windows platform

Dropping it won't solve the root cause, just hiding the issue.

hcho3:
> Dropping it won't solve the root cause, just hiding the issue.

I agree. I just wanted to unblock this PR and #5832 so that we can start enjoying faster turn-around time.

trivialfis:
@hcho3 Can I merge it by reverting the CI change?

hcho3:
Sure

trivialfis:
@hcho3 Do you think it's suitable to merge without passing power test?  It seems travis is having difficulty on allocating power node. https://travis-ci.org/github/dmlc/xgboost/pull_requests

Issue #5063, opened 2019-11-24
External memory deosn't work without OpenMP.


trivialfis:
GTest `SparsePage.PushCSCAfterTranspose` returns invalid number of batches when XGBoost is not compiled with OpenMP.

Platform: Ubuntu 19.04 X86_64
GCC: 8.3

trivialfis:
Somehow the number of pages is tied to number of threads: https://github.com/dmlc/dmlc-core/blob/552f7de748fbff34f2708b03f930a47ded45d78e/src/data/text_parser.h#L122

Issue #6167, opened 2020-09-25
External Memory Usage with GPU - "Invalid Resource Handle" Error


talipini:
Running xgboost on external memory causes "invalid resource handle" error with GPU.

I have a large dataset that doesn't fit into memory and I was hoping to use external memory to train as explained in https://xgboost.readthedocs.io/en/latest/tutorials/external_memory.html. CPU version seems to work fine but using "gpu_hist" causes this error.  

**terminate called after throwing an instance of 'thrust::system::system_error'
  what():  parallel_for failed: invalid resource handle**

Here is the sample code. create_samples creates a naive dummy 2M samples with 500 features each. Our dataset is much larger but this sample serves the purpose to recreate the error.  

```
import xgboost
import pandas as pd
import numpy as np
import pathlib

def create_samples(sample_file):
    with open(sample_file, 'a') as f:
        for i in range(0,20):
            print(f'Writing chunk - {i}')
            chunk = 100000
            sample_df = pd.DataFrame(np.random.uniform(low=-1.0, high=5, size=(chunk,500)))
            y = np.random.randint(low=1, high=11, size=chunk)
            sample_df.insert(0, column='y',value=y)
            sample_df.to_csv(f,index=False, header=False)

    print(f'Finished creating sample file at {sample_file}')


def run_xgboost(data_file):
    dtrain = xgboost.DMatrix(f"{data_file}?format=csv&label_column=0#traincache")
    params = {
        'objective':'multi:softmax', 
        "num_class": 11,             
        'eval_metric':'merror',
        'nthread': 1,
        'tree_method':'gpu_hist',  
        'sampling_method': 'gradient_based',
    }

    bst = xgboost.train(
        params=params,
        dtrain=dtrain,
        num_boost_round=3
    )
    print("Finished training. ")



if __name__ == '__main__':

    sample_file = "sample.csv" 
    #create dummy sample, if it doesn't exist
    if not pathlib.Path(sample_file).is_file():
        create_samples(sample_file) 
    
    run_xgboost(sample_file)
```

XGBoost Version - 1.3.0-SNAPSHOT (latest nightly build)
GPU: RTX 2080 Ti
Driver: 440.82
CUDA: 10.2

jamescolless:
I'm not sure how to extract a particular error message, but I'm having the same issue with version 1.2.0. A large-ish data set which trains fine using CPU on external memory, but fails when I switch to gpu_hist.

talipini:
[...]

PR #6296, opened 2020-10-27
[WIP] Bypass threaded iter in external memory.

trivialfis:
* Remove the dependency on threaded iterator.
* Make data partitioning deterministic.
* Add Ubuntu to gtest GitHub action job.

Close: https://github.com/dmlc/xgboost/issues/5063

Possibly related: https://github.com/dmlc/xgboost/issues/6167

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-10.15]
+        os: [macos-10.15, ubuntu-latest]
     steps:
     - uses: actions/checkout@v2
       with:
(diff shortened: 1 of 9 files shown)


hcho3:
Does this also fix race condition in external memory? We might be able to transition to Ninja builds fully then?

trivialfis:
> Does this also fix race condition in external memory? We might be able to transition to Ninja builds fully then?

I want to get it fully working and leave the forever beta status.  Still need some time for maturing and adding tests.  In this PR I included some internal files in dmlc core, so opening an early PR in case others have objection.

hcho3:
Great, I really want to remove all non-determinism from external memory implementation. Right now, we can't use Ninja on Windows because apparently using a slightly different build order causes some unit tests to fail.

PR #6302, opened 2020-10-28
Support slicing tree model

### LINKED ISSUE #4052: [feature proposal] Load best model after early stopping

### LINKED ISSUE #5531: Removing trees from a model

trivialfis:
This PR is meant the end the confusion around `best_ntree_limit` and unify model slicing.  We have multi-class and random forests, asking users to understand how to set `ntree_limit` is difficult and error prone.

Close #5531 Close #4052.

* Implement the `save_best` option in early stopping.
* Negative index is not supported.

### DIFF
--- a/doc/python/model.rst
+++ b/doc/python/model.rst
@@ -1,0 +1,27 @@
+#####
+Model
+#####
+
+Slice tree model
+----------------
+
+When ``booster`` is set to ``gbtree`` or ``dart``, XGBoost builds a tree model, which is a
+list of trees and can be sliced into multiple sub-models.
+
+.. code-block:: python
+    from sklearn.datasets import make_classification
+    num_classes = 3
+    X, y = make_classification(n_samples=1000, n_informative=5,
+                               n_classes=num_classes)
+    dtrain = xgb.DMatrix(data=X, label=y)
+    num_parallel_tree = 4
+    num_boost_round = 16
+    total_trees = num_parallel_tree * num_classes * num_boost_round
+
+    # We build a boosted random forest for classification here.
+    booster = xgb.train({
+        'num_parallel_tree': 4, 'subsample': 0.5, 'num_class': 3},
+                        num_boost_round=num_boost_round, dtrain=dtrain)
+
+    # This is the sliced model, containing [3, 7) forests
+    sliced: xgb.Booster = booster[3:7]
--- a/doc/python/model.rst
+++ b/doc/python/model.rst
@@ -24,4 +24,10 @@ list of trees and can be sliced into multiple sub-models.
                         num_boost_round=num_boost_round, dtrain=dtrain)
 
     # This is the sliced model, containing [3, 7) forests
+    # step is also supported with some limitations like negative step is invalid.
     sliced: xgb.Booster = booster[3:7]
+
+
+The sliced model is a copy of selected trees, that means the model itself is immutable
+during slicing.  This feature is the basis of `save_best` option in early stopping
+callback.
--- a/doc/python/model.rst
+++ b/doc/python/model.rst
@@ -27,6 +27,10 @@ list of trees and can be sliced into multiple sub-models.
     # step is also supported with some limitations like negative step is invalid.
     sliced: xgb.Booster = booster[3:7]
 
+    # Access individual tree layer
+    trees = [_ for _ in booster]
+    assert len(trees) == num_boost_round
+
 
 The sliced model is a copy of selected trees, that means the model itself is immutable
 during slicing.  This feature is the basis of `save_best` option in early stopping
--- a/doc/python/model.rst
+++ b/doc/python/model.rst
@@ -16,7 +16,7 @@ list of trees and can be sliced into multiple sub-models.
     dtrain = xgb.DMatrix(data=X, label=y)
     num_parallel_tree = 4
     num_boost_round = 16
-    total_trees = num_parallel_tree * num_classes * num_boost_round
+    # total number of built trees is num_parallel_tree * num_classes * num_boost_round
 
     # We build a boosted random forest for classification here.
     booster = xgb.train({
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -581,12 +581,14 @@ XGB_DLL int XGBoosterCreate(const DMatrixHandle dmats[],
 XGB_DLL int XGBoosterFree(BoosterHandle handle);
 
 /*!
- * \brief Slice a model according to layers.
+ * \brief Slice a model using boosting index. The slice m:n indicates taking all trees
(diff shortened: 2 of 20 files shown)


trivialfis:
Currently I treat out of bound error specially raise `IndexError` but  `ValueError` for other issues.  Should I always raise `IndexError` instead?  Feel free to provide suggestions . ;-)

trivialfis:
Pasting offline conversion with @hcho3 here.  The trees in xgboost can be considered as a 3 dim tensor.  First dim is the number of boosting round, second is number of classes, and the last is size of forest.  This PR supports only slicing the first dim (number of boosted rounds), it's possible to support slicing other dimensions. but to us that seems to be over engineering.

RAMitchell (approved):
This is a great new feature. Any plans to deprecate `ntree_limit` throughout the code base in favour of the new terminology?

trivialfis:
There are other language bindings out there.  I need to go over them to deprecate the parameter.

include/xgboost/gbm.h
```diff
@@ -60,6 +60,16 @@ class GradientBooster : public Model, public Configurable {
    * \param fo output stream
    */
   virtual void Save(dmlc::Stream* fo) const = 0;
+  /*!
+   * \brief Slice the model.
+   * \param layer_begin Begining of boosted tree layer used for prediction.
+   * \param layer_end   End of booster layer. 0 means do not limit trees.
+   * \param out         Output gradient booster
+   */
```
hcho3, line 68:
```suggestion
  /*!
   * \brief Slice a model using boosting index. The slice m:n indicates taking all trees
 *        that were fit during the boosting rounds m, (m+1), (m+2), ..., (n-1).
   * \param layer_begin Begining of boosted tree layer used for prediction.
   * \param layer_end   End of booster layer. 0 means do not limit trees.
   * \param out         Output gradient booster
   */
```

tests/python/test_basic_models.py
```diff
@@ -325,3 +325,69 @@ def validate_model(parameters):
...
+        self.assertRaises(IndexError, lambda: booster[:num_boost_round+1])
+        self.assertRaises(ValueError, lambda: booster[1, 2])  # too many dims
+
+        def assign():
+            booster[...:end] = booster
+        self.assertRaises(TypeError, assign)
+
+    def test_slice(self):
+        self.run_slice('gbtree')
+        self.run_slice('dart')
```
hcho3, line 393: Can we use `@pytest.mark.parameterize` instead?

```python
@pytest.mark.parameterize(booster, ['gbtree', 'dart'])
def run_slice(self, booster):
```

See examples at [Parameterizing tests](https://docs.pytest.org/en/6.1.2/example/parametrize.html)

src/gbm/gbtree.cc
```diff
@@ -398,6 +398,38 @@ void GBTree::SaveModel(Json* p_out) const {
...
+  std::vector<int32_t> &out_trees_info = out_model.tree_info;
+  out_trees_info.resize(layer_trees * n_layers);
+  out_model.param.num_trees = out_model.trees.size();
+  CHECK(this->model_.trees_to_update.empty());
+
+  *out_of_bound = detail::SliceTrees(
+      layer_begin, layer_end, step, this->model_, tparam_, layer_trees,
+      [&](auto const &in_it, auto const &out_it) {
+        auto new_tree =
+            std::make_unique<RegTree>(*this->model_.trees.at(in_it));
```
hcho3, line 426: Do we have assurance that the implicitly generated copy constructor `RegTree(const RegTree&)` behaves correctly?

trivialfis: Added tests with prediction.

tests/python/test_basic_models.py
trivialfis: It seems not being compatible with class  method:
```
TypeError: test_slice() missing 1 required positional argument: 'booster'
```

doc/python/model.rst
```diff
@@ -0,0 +1,38 @@
...
+.. code-block:: python
+
+    from sklearn.datasets import make_classification
+    num_classes = 3
+    X, y = make_classification(n_samples=1000, n_informative=5,
+                               n_classes=num_classes)
+    dtrain = xgb.DMatrix(data=X, label=y)
+    num_parallel_tree = 4
+    num_boost_round = 16
+    total_trees = num_parallel_tree * num_classes * num_boost_round
```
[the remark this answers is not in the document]
trivialfis: Converted into a comment.

include/xgboost/c_api.h
```diff
@@ -580,6 +580,21 @@ XGB_DLL int XGBoosterCreate(const DMatrixHandle dmats[],
...
 XGB_DLL int XGBoosterFree(BoosterHandle handle);
 
+/*!
+ * \brief Slice a model according to layers.
+ *
+ * \param handle Booster to be sliced.
+ * \param begin_layer start of the slice
+ * \param end_layer   end of the slice
+ * \param step        step size of the slice
+ * \param out Sliced booster.
```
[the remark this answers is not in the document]
trivialfis: Added comments.

include/xgboost/gbm.h
trivialfis: Added comments.

tests/python/test_basic_models.py
hcho3: Try `@pytest.mark.parameterize('booster',['gbtree', 'dart'])` (note the quotes around `booster`

hcho3: @trivialfis Also, `TestModels` should not be a subclass of `unittest.TestCase`. https://stackoverflow.com/a/35562401. Try making the class a subclass of `object`:
```python
class TestModels(object):
```

trivialfis: Thanks for the suggestion, done.

hcho3 (approved):

hcho3:
The parametrization looks quite nice. The benefit is that, if a parametrized test fails for only one of the multiple parameters, the test result will clearly indicate which parameter led to failure. I'd like to submit a follow-up PR to introduce more test parametrization where it's appropriate. For example: https://github.com/dmlc/xgboost/blob/29745c6df2d6240579d9af22fd22e35db08acf02/tests/python/test_with_dask.py#L431-L444

PR #6305, opened 2020-10-28
[CI] Use separate Docker cache for each CUDA version

### LINKED ISSUE #6301: [CI] Reduce the number of times Docker images are built
Currently, the CI server heavily uses Docker containers to manage suitable testing environments for the Linux platform. The original intent was to avoid the need for installing packages over and over again, by caching Docker images that were previously built. So the cost of building Docker containers would be amortized over multiple test jobs.

Unfortunately, in recent weeks, the Docker containers are being re-built from scratch very frequently. ~The reason is that, whenever the base container (`nvidia/cuda`) gets updated, all the cached Docker layers get invalidated, forcing our CI server to re-build the container from scratch.~ **Edit. This is not the correct diagnosis. See the latest comment below.**

I recognize the importance of that security patches, but having to re-build entire Docker containers from scratch is tremendously wasteful. (It takes 20-30 mins to complete the build.)

Proposed fix. ~We should have a cron job (or something equivalent) that builds new Docker containers from the Dockerfiles at a preset frequency (say, every 2 weeks). All CI jobs should now always pull pre-build containers, instead of attempting to build containers from Dockerfile. In short, **our containers should be updated less frequently than the base `cuda` container is updated.** The disadvantage is that **updating Dockerfile won't cause rebuilding of the container immediately.**~ **Edit. See the latest comment below.**

@trivialfis

hcho3:
Closes #6301. This will greatly reduce the number of times the Docker container images are re-built.

#6202 introduced a subtle bug that pushed all GPU containers to the same registry `xgb-ci.gpu`, regardless of their CUDA version used. With this patch, GPU containers will be pushed to appropriate registries that are labeled with CUDA versions: `xgb-ci.gpu10.0`, `xgb-ci.gpu10.2`, and `xgb-ci.gpu11.0`.

### DIFF
--- a/tests/ci_build/ci_build.sh
+++ b/tests/ci_build/ci_build.sh
@@ -90,7 +90,7 @@ WORKSPACE="${WORKSPACE:-${SCRIPT_DIR}/../../}"
 DOCKER_IMG_NAME="xgb-ci.${CONTAINER_TYPE}"
 
 # Append cuda version if available
-CUDA_VERSION=$(echo "${CI_DOCKER_BUILD_ARG}" | grep -o -E 'CUDA_VERSION=[0-9]+\.[0-9]+' | grep -o -E '[0-9]+\.[0-9]+')
+CUDA_VERSION=$(echo "${CI_DOCKER_BUILD_ARG}" | grep -o -E 'CUDA_VERSION_ARG=[0-9]+\.[0-9]+' | grep -o -E '[0-9]+\.[0-9]+')
 # Append jdk version if available
 JDK_VERSION=$(echo "${CI_DOCKER_BUILD_ARG}" | grep -o -E 'JDK_VERSION=[0-9]+' | grep -o -E '[0-9]+')
 # Append cmake version if available


hcho3:
One small typo ended up costing a lot of test time, as it turns out. The `ci_build.sh` needs to be cleaned up for legibility and clarity. I will create a follow-up issue for this.

trivialfis (approved):
I confess I don't understand what's happening.  Approving as long as it fixes the issue.

> The ci_build.sh needs to be cleaned up for legibility and clarity. I will create a follow-up issue for this.

When you are on this, can we use a script like Python instead of sh?

hcho3:
> I confess I don't understand what's happening

It has to do with how we currently "cache" Docker containers, as re-building containers at every job would be very expensive. The way we cache them is to push previously built container images to a private repository (Elastic Container Repository from AWS), and then re-use the pre-built images for later testing jobs.

* "Cache hit" occurs when there has been no change in the current Dockerfile or the base image (`nvidia/cuda`), so the pre-built image matches perfectly with the Dockerfile. The docker build will proceed very quickly, and it will emit log saying `from the cache`.

* "Cache miss" occurs when either Dockerfile or the base image (`nvidia/cuda`) was recently updated, so the pre-built image is no longer with sync with Dockerfile. In this case, the CI will re-built the Docker container from scratch.

CircleCI has a good article about Docker caching and how it speeds up CI: https://circleci.com/docs/2.0/docker-layer-caching/

hcho3:
> When you are on this, can we use a script like Python instead of sh?

That's an excellent idea. We can avoid lots of potential bugs with string manipulation.

hcho3:
Note that the CI is now building new containers because we changed the name of the registry 🤦

hcho3:
Temporarily raised the budget limit to 100 USD for today.

Issue #6282, opened 2020-10-23
hang in import of xgboost due to dask even if not using dask class


pseudotensor:
same setup as here:  #6232 
```
Current thread 0x000014970d3ff700 (most recent call first):                                                                                                                                                                                                            
  File "/home/jon/minicondadai/lib/python3.6/_weakrefset.py", line 75 in __contains__                                                                                                                                                                                  
  File "/home/jon/minicondadai/lib/python3.6/abc.py", line 206 in __subclasscheck__                                                                                                                                                                                    
  File "/home/jon/minicondadai/lib/python3.6/typing.py", line 884 in __extrahook__
  File "/home/jon/minicondadai/lib/python3.6/abc.py", line 209 in __subclasscheck__
  File "/home/jon/minicondadai/lib/python3.6/typing.py", line 1154 in __subclasscheck__
  File "/home/jon/minicondadai/lib/python3.6/abc.py", line 228 in __subclasscheck__
  File "/home/jon/minicondadai/lib/python3.6/abc.py", line 228 in __subclasscheck__
  File "/home/jon/minicondadai/lib/python3.6/abc.py", line 193 in __instancecheck__
  File "/home/jon/minicondadai/lib/python3.6/site-packages/dask/config.py", line 96 in update
  File "/home/jon/minicondadai/lib/python3.6/site-packages/dask/config.py", line 99 in update
  File "/home/jon/minicondadai/lib/python3.6/site-packages/dask/config.py", line 99 in update
  File "/home/jon/minicondadai/lib/python3.6/site-packages/dask/config.py", line 497 in update_defaults
  File "/home/jon/minicondadai/lib/python3.6/site-packages/distributed/config.py", line 20 in <module>
  File "<frozen importlib._bootstrap>", line 219 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 678 in exec_module
  File "<frozen importlib._bootstrap>", line 665 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 955 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 971 in _find_and_load
  File "<frozen importlib._bootstrap>", line 219 in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1023 in _handle_fromlist
  File "/home/jon/minicondadai/lib/python3.6/site-packages/distributed/__init__.py", line 1 in <module>
[...]

PR #6309, opened 2020-10-28
Lazy import dask libraries.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/6282 .


hcho3 (approved):

trivialfis:
One downside is type annotation on function is not possible now.

PR #6325, opened 2020-10-30
Clean up warnings

igormp:
Continuing work on #6189.

### DIFF
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -163,7 +163,7 @@ XGB_DLL int XGDMatrixCreateFromCSCEx(const size_t* col_ptr,
                                      const unsigned* indices,
                                      const bst_float* data,
                                      size_t nindptr,
-                                     size_t nelem,
+                                     size_t,
                                      size_t num_row,
                                      DMatrixHandle* out) {
   API_BEGIN();
--- a/src/common/version.cc
+++ b/src/common/version.cc
@@ -16,7 +16,7 @@ namespace xgboost {
 
 const Version::TripletT Version::kInvalid {-1, -1, -1};
 
-Version::TripletT Version::Load(Json const& in, bool check) {
+Version::TripletT Version::Load(Json const& in) {
   if (get<Object const>(in).find("version") == get<Object const>(in).cend()) {
     return kInvalid;
   }
--- a/src/tree/tree_model.cc
+++ b/src/tree/tree_model.cc
@@ -179,7 +179,7 @@ class TextGenerator : public TreeGenerator {
   using SuperT = TreeGenerator;
 
  public:
-  TextGenerator(FeatureMap const& fmap, std::string const& attrs, bool with_stats) :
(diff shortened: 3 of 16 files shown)


hcho3:
Can you address the issues found by the style checker? See the log at https://github.com/dmlc/xgboost/pull/6325/checks?check_run_id=1330040466.

igormp:
Done.
Just out of curiosity, I compared the amount of warnings since the beginning of this month until now, and we went from 2340 warnings to 1889.
If we ignore the ones related to dmlc, then we went from 1584 to 1057, we managed to reduce a third of the warnings, which is pretty nice :partying_face:

src/common/version.cc
```diff
@@ -16,7 +16,7 @@ namespace xgboost {
 
 const Version::TripletT Version::kInvalid {-1, -1, -1};
 
-Version::TripletT Version::Load(Json const& in, bool check) {
+Version::TripletT Version::Load(Json const& in) {
```
trivialfis, line 19: What was I thinking when I added this parameter ...

src/c_api/c_api.cc
```diff
@@ -163,7 +163,7 @@ XGB_DLL int XGDMatrixCreateFromCSCEx(const size_t* col_ptr,
                                      const unsigned* indices,
                                      const bst_float* data,
                                      size_t nindptr,
-                                     size_t nelem,
+                                     size_t,
```
trivialfis, line 166: Please keep this parameter for now.  I'm not sure about these old APIs.

src/tree/tree_model.cc
```diff
@@ -69,20 +69,24 @@ class TreeGenerator {
...
+                                int32_t /*nid*/, uint32_t /*depth*/) const {
     return "";
   }
-  virtual std::string Quantitive(RegTree const& tree, int32_t nid, uint32_t depth) const {
+  virtual std::string Quantitive(RegTree const& /*tree*/,
+                                int32_t /*nid*/, uint32_t /*depth*/) const {
     return "";
   }
-  virtual std::string NodeStat(RegTree const& tree, int32_t nid) const {
+  virtual std::string NodeStat(RegTree const& /*tree*/, int32_t /*nid*/) const {
```
trivialfis, line 84: Thanks for converting them into comments!

src/c_api/c_api.cc
igormp: In a previous PR of mine I already left some parameters as unnamed in the C api, should I revert those as well?

Another option would be to use `__attribute__((unused))`, which should work on both GCC and Clang.

trivialfis: Got it.  Then let's keep the PR as it's.  Thanks for the good work on reducing warnings!

trivialfis (approved):

src/tree/tree_model.cc
igormp: I thought that, since those are virtuals, it'd make sense to leave the name for anyone else doing some override :)

PR #6337, opened 2020-11-03
Use pytest conventions consistently

hcho3:
* ~Adopt parametrized tests.~ Will do this in a later PR.
* Remove all uses of `unittest` and `unittest.TestCase`.
* Use plain `assert` instead of `assertEquals`, `assertGreater` etc, per [the pytest guideline](https://docs.pytest.org/en/stable/assert.html).
* Rename `setUpClass` to `setup_class`. The former was used by `unittest`, and the latter is the correct spelling in `pytest`.
* Replace all occurrences of `assertRaises` with `pytest.raises`. This way, we can use natural notations for function calls.

Note to reviewers: Ignore whitespaces for a small diff.
Screen Shot 2020-11-03 at 2 02 55 AM


tests/python/test_with_sklearn.py
```diff
@@ -28,23 +27,23 @@ def __exit__(self, exc_type, exc_value, traceback):
         shutil.rmtree(self.name)
 
 
-def test_binary_classification():
+@pytest.mark.parametrize('cls', [xgb.XGBClassifier, xgb.XGBRFClassifier])
```
trivialfis, line 30: Do we need `ids` in here?

tests/python/test_monotone_constraints.py
```diff
@@ -61,62 +60,29 @@ def is_correctly_constrained(learner):
...
-class TestMonotoneConstraints(unittest.TestCase):
+class TestMonotoneConstraints:
 
-    def test_monotone_constraints_for_exact_tree_method(self):
-
-        # first check monotonicity for the 'exact' tree method
-        params_for_constrained_exact_method = {
-            'tree_method': 'exact', 'verbosity': 1,
-            'monotone_constraints': '(1, -1)'
+    @pytest.mark.parametrize('tree_method,extra_kwargs',
```
trivialfis, line 65: Why do you need a extra kwargs?

tests/python-gpu/test_gpu_with_sklearn.py
```diff
@@ -33,8 +32,5 @@ def test_gpu_binary_classification():
             assert err < 0.1
 
 
-class TestGPUBoostFromPrediction(unittest.TestCase):
-    cpu_test = twskl.TestBoostFromPrediction()
-
-    def test_boost_from_prediction_gpu_hist(self):
-        self.cpu_test.run_boost_from_prediction('gpu_hist')
+def test_boost_from_prediction_gpu_hist():
+    cpu_test = twskl.test_boost_from_prediction(tree_method='gpu_hist')
```
trivialfis, line 36: This looks weird, are we sure it's running `gpu_hist` and `gpu_hist` only?

trivialfis (commented):
I agree with choosing structs in pytest over unittest.  But I have some tests defined in CPU tests but shared with GPU tests, I'm not sure whether they are running as expected with parametrization.

tests/python/test_monotone_constraints.py
hcho3: We can't set grow_policy when tree_method is not hist.

hcho3:
Since this PR is getting big, I will remove parametrized tests from this PR and save them for a follow-up PR

tests/python/test_with_sklearn.py
hcho3: Not strictly necessary, `ids` produce nicer outputs in the test listing

tests/python-gpu/test_gpu_with_sklearn.py
hcho3: I remove test parametrization from this PR.

hcho3:
@trivialfis Can you review again? I've remove all parametrized tests for now.

trivialfis (approved):

trivialfis:
LGTM given the tests pass.

hcho3:
@trivialfis Can you review the revised `test_slice`? I revised it to test the exact semantics of slicing and not rely on randomness.

trivialfis (approved):
LGTM given the tests are passing.

PR #6342, opened 2020-11-04
Optionaly fail when gpu_id is set to invalid value

honzasterba:
I need a way to find out how many GPUs are available in the system. The current behaviour is that XGBoost just replaces invalid gpu IDs with a valid one. I would like to add a param that allows me to switch this off.

### DIFF
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -580,7 +580,7 @@ class GPUPredictor : public xgboost::Predictor {
       Predictor::Predictor{generic_param} {}
 
   ~GPUPredictor() override {
-    if (generic_param_->gpu_id >= 0) {
+    if (generic_param_->gpu_id >= 0 && generic_param_->gpu_id < common::AllVisibleGPUs()) {
       dh::safe_cuda(cudaSetDevice(generic_param_->gpu_id));
     }
   }
--- a/include/xgboost/generic_parameters.h
+++ b/include/xgboost/generic_parameters.h
@@ -11,6 +11,7 @@
 #include <string>
 
 namespace xgboost {
+
 struct GenericParameter : public XGBoostParameter<GenericParameter> {
   // Constant representing the device ID of CPU.
   static int32_t constexpr kCpuId = -1;
@@ -26,6 +27,8 @@ struct GenericParameter : public XGBoostParameter<GenericParameter> {
   int nthread;
   // primary device, -1 means no gpu.
   int gpu_id;
+  // fail when gpu_id is invalid
+  bool fail_on_invalid_gpu_id {false};
   // gpu page size in external memory mode, 0 means using the default.
   size_t gpu_page_size;
   bool enable_experimental_json_serialization {true};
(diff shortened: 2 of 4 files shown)


src/predictor/gpu_predictor.cu
```diff
@@ -527,7 +527,7 @@ class GPUPredictor : public xgboost::Predictor {
       Predictor::Predictor{generic_param} {}
 
   ~GPUPredictor() override {
-    if (generic_param_->gpu_id >= 0) {
+    if (generic_param_->gpu_id >= 0 && generic_param_->gpu_id < common::AllVisibleGPUs()) {
```
honzasterba, line 530: I am not happy about this and I am open to suggestions. This happens when the check in learner.cc fails and the gpu_id value stays invalid. The line below than causes the whole JVM to halt.
An alternative solution would be to add another flag to generic_param or Predictor named like `gpu_id_is_valid`, but that seems even more cumbersome.
Also I do not understand why we need to `cudaSetDevice` in destructor.

trivialfis (commented):
> I need a way to find out how many GPUs are available in the system.

Not quite sure why you would like xgboost to fail in order to obtain the number of GPUs?  There are other more reliable and reusable way to do it.  Like having a small c extension that just calls into cuda api, or use `pynvml` and check for `CUDA_VISIBLE_DEVICES`.

honzasterba:
That did cross my mind as well, but it seemed reasonable to want xgboost to fail when gpu_id is invalid. You think it is not?

src/predictor/gpu_predictor.cu
trivialfis: I think `thrust` might try to do some other operations during destruction at the time of writing these weird code.  Without a correct global gpu id context something broke.

trivialfis:
> That did cross my mind as well, but it seemed reasonable to want xgboost to fail when gpu_id is invalid. 

I think it makes perfect sense, my question is mostly out of curiosity.  I used to throw the error directly, but the Python pickle file generated on GPU machine may be later loaded on a CPU only machine or machine with fewer GPUs.  So we added the wrapping for `gpu_id` as a workaround for these corners.

src/predictor/gpu_predictor.cu
trivialfis: Will look into this line of code later, see if there's way to remove them.

honzasterba: @trivialfis if there is any research/experimentation I can do to help, please let me know.

honzasterba:
@trivialfis please let me know if there is anything I can do to help get this done

trivialfis (commented):
Please help adding a test so that we don't break it in the future, either Python or C++ is fine.

honzasterba:
test added

trivialfis (approved):
LGTM.  Thanks for your contribution.

hcho3:
@trivialfis Should this be part of 1.3.0 release?

trivialfis:
Yes if there's no conflict.

Issue #5765, opened 2020-06-08
Host IP resolution problem w/ dask on kubernetes or dask-gateway


gforsyth:
Hi!

There are two different failure modes here, but I think the solution to them is the same so I'm going to keep them bundled.  Both of these come up using the bundled `dask` helpers and in both cases (k8s & dask-gateway) I believe the culprit is here:

https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/dask.py#L360-L367

For a dask cluster deployed with a helm chart, the scheduler pod is exposed via a kubernetes service. The service acts as a reverse proxy, providing a static route to the scheduler pod (that would survive the scheduler pod restarting).

Currently, `xgboost.dask` isn't compatible with this deployment pattern:
```
import dask.array as da
from distributed.client import Client
from xgboost.dask import DaskXGBClassifier

client = Client()
print(client.scheduler.address)

xgbc = DaskXGBClassifier()

X = da.random.random((10, 10))
y = da.random.random((10, 1))
xgbc.fit(X, y)

tcp://dask-scheduler:8786
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-12-4607fe3bd004> in <module>
     11 X = da.random.random((10, 10))
     12 y = da.random.random((10, 1))
---> 13 xgbc.fit(X, y)

/opt/conda/lib/python3.7/site-packages/xgboost/dask.py in fit(self, X, y, sample_weights, eval_set, sample_weight_eval_set)
    651         results = train(self.client, params, dtrain,
    652                         num_boost_round=self.get_num_boosting_rounds(),
--> 653                         evals=evals)
    654         self._Booster = results['booster']
    655         # pylint: disable=attribute-defined-outside-init

/opt/conda/lib/python3.7/site-packages/xgboost/dask.py in train(client, params, dtrain, evals, *args, **kwargs)
    382     workers = list(_get_client_workers(client).keys())
    383 
--> 384     rabit_args = _get_rabit_args(workers, client)
    385 
    386     def dispatched_train(worker_addr):

/opt/conda/lib/python3.7/site-packages/xgboost/dask.py in _get_rabit_args(worker_map, client)
    336 
    337     env = client.run_on_scheduler(_start_tracker, host.strip('/:'),
--> 338                                   len(worker_map))
    339     rabit_args = [('%s=%s' % item).encode() for item in env.items()]
    340     return rabit_args

[...]

PR #6343, opened 2020-11-04
[dask] Fix running on k8s.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/5765

* Use `socket` from Python.
* Only use workers associated with input data instead of all workers from client.  This avoids accessing `client.scheduler_info()['workers']`.
* Avoid calling `client.gather` inside task function.

~XGBoost still cannot be run on GKE yet.~  See https://github.com/dask/dask/issues/6800 for details.

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -851,8 +851,9 @@ async def _predict_async(client, model, data, missing=numpy.nan, **kwargs):
     async def map_function(func):
         '''Run function for each part of the data.'''
         futures = []
-        for wid in range(len(worker_map.keys())):
-            worker_addr = list(worker_map.keys())[wid]
+        workers_address = list(worker_map.keys())
+        for wid in range(len(workers_address)):
+            worker_addr = workers_address[wid]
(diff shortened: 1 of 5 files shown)


jameslamb:
Hey @trivialfis , I'm really excited to see this! I have easy access to Dask clusters on EKS, so if you want me to test anything there I'd be happy to.

trivialfis:
@jameslamb Thanks for the offer!  So far I managed to get some examples working on GKE.  But automated pytest is not on the table since everything just times out while waiting for scheduling.  Feel free to test it out, it might help uncovering some other unknown issues on xgboost and dask and be useful to others.  Also the issue is not reproducible on local deployment, so they are cluster specific.  Your tests are definitely gonna be helpful!

I'm not sure about the performance impact and memory usage impact of current workaround, should be trivial even if there's any but I need to try it out.

RAMitchell (approved):

hcho3 (approved):

trivialfis:
A refactoring is coming.

PR #6345, opened 2020-11-05
[docs] [dask] make Dask distributed example copy-pastable

jameslamb:
Hello! I've been getting more familiar with the XGBoost Dask interface, as part of https://github.com/dask/community/issues/104. I was working through the documentation at https://xgboost.readthedocs.io/en/latest/tutorials/dask.html#overview tonight, and I'd like to suggest an improvement to them.

This PR proposes adding some Dask Array code to generate random data, instead of relying on users to figure that out themselves based on a code comment. I think making the examples easy to run by just copying and pasting will make it easier for users to start experimenting with `xgboost.dask`.

This also proposes adding links to the Dask DataFrame and Dask Array documentation, to help users who might not be familiar with those collections.

### DIFF
--- a/doc/tutorials/dask.rst
+++ b/doc/tutorials/dask.rst
@@ -41,24 +41,43 @@ on a dask cluster:
 .. code-block:: python
 
   import xgboost as xgb
+  import dask.array as da
   import dask.distributed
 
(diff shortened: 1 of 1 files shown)


hcho3 (approved):
Nice, thanks for taking interest in XGBoost Dask API.

PR #6351, opened 2020-11-06
Update agaricus-lepiota.fmap .

calxu:
Fix a minor symbol mistake. The following key value is "bruies", there should be no "?".


trivialfis (commented):
That seems to be in part of the `agaricus-lepiota.names`, which is part of data description.

PR #6358, opened 2020-11-09
Thread local memory allocation for BuildHist

ShvetsKS:
This PR allows to use thread local memory for better thread work balance

mortgage, train | train time (master), s | train time (this PR), s | Speedup
-- | -- | -- | --
BuildLocalHist | 10.5 | 5.9 | 1.78
ApplySplit |  6.6 | 4.8 | 1.375
Full train | 22.14 - 3 | 15.62 - 3 | 1.45


trivialfis:
Actually 1 day I went nuts and rewrote all the CPU approximated algorithms.  I think my re-implementation is simpler and has better perf.  Will link a branch later.

ShvetsKS:
> Actually 1 day I went nuts and rewrote all the CPU approximated algorithms (hist, approx, local). I think my re-implementation is simpler and has better perf. Will link a branch later.

It's really interesting. Could you share initial version of these changes to compare performance?

ShvetsKS:
@trivialfis can we consider these optimizations for upcoming release?

trivialfis:
@ShvetsKS Could you please take a look into my impl? https://github.com/trivialfis/xgboost/blob/rebase-local/src/tree/updater_hist.cc  I unified the implementation for all 3 approximated tree methods.  It's not yet done, but very close.  If you agree that it's simpler then I might complete replace the implementation here.

trivialfis:
The unification is necessary for categorical data support.

ShvetsKS:
@trivialfis Sure, thanks for providing your implementation. I'm studying it now :)

ShvetsKS:
@trivialfis, Sorry for long response.
I didn't notice any performance improvements and even significant slowdown was discovered compared to master branch. Could you share yours benchmarks results with perf improvements?

train time, s | higgs1m | airline-ohe | mortgage1q | msranks | Santander | plasticc
-- | -- | -- | -- | -- | -- | -- |
master |  23.8448 | 56.8644 | 23.01579 | 112.164 | 137.42 | 1.2505
trivialfis/rebase-local | 56.4 | 130.8829 | 36.24462 | 189.9914 | 302.611 | 2.196214

Seems current reimplementation of the training flow (in `UpdateTreeWithDriver` function) still uses old performance oriented kernels (BuildHistogram, PartitionBuilder) but supports only `lossguide` training mode.

Possibly the reason of slowdown is `lossguide` vs `depthwise` comparison but for `mortgage1q` measurements I used `lossguide` for both branches and it's strange to see the slowdown.

I'd like to note this PR would be definitely applicable for scope of reimplementation in `trivialfis/rebase-local` also :).
So is it possible to merge it?

And I'd be glad to help to finish code refactoring/unification but not sure that it can be finished before 1.3 release as there is significant problem with perf.

trivialfis:
It supports depthwise.  I don't expect it to be merged anytime soon.  This size of change will take a long  procedure to ensure everything is done right.

I will review the PR as it's later today.

src/common/hist_util.cc
```diff
@@ -552,6 +552,7 @@ void BuildHistDenseKernel(const std::vector<GradientPair>& gpair,
   const BinIdxType* gradient_index = gmat.index.data<BinIdxType>();
```
trivialfis, line 552: I think there's no actual change in this file right?

src/common/row_set.h
```diff
@@ -185,13 +195,13 @@ class PartitionBuilder {
     for (size_t i = 0; i < blocks_offsets_.size()-1; ++i) {
       size_t n_left = 0;
       for (size_t j = blocks_offsets_[i]; j < blocks_offsets_[i+1]; ++j) {
-        mem_blocks_[j].n_offset_left = n_left;
-        n_left += mem_blocks_[j].n_left;
+        mem_blocks_[j][0].n_offset_left = n_left;
```
trivialfis, line 198: Where is elements other than 0 used?

src/common/row_set.h
```diff
@@ -149,25 +149,35 @@ class PartitionBuilder {
       max_n_tasks_ = n_tasks;
     }
   }
+  void AllocateForTask(size_t id) {
```
trivialfis, line 152: Can you document when should these 2 new functions used?

src/common/hist_util.h
```diff
@@ -634,6 +647,7 @@ class ParallelGHistBuilder {
   std::vector<GHistRowT> hist_memory_;
   /*! \brief map pair {tid, nid} to index of allocated histogram from hist_memory_  */
   std::map<std::pair<size_t, size_t>, size_t> tid_nid_to_hist_;
+  std::map<std::pair<size_t, size_t>, int> tid_nid_to_hist_hist_buffer_;
```
trivialfis, line 650: ?

src/common/hist_util.h
```diff
@@ -438,21 +438,31 @@ class HistCollection {
...
-      data_.resize(nbins_ * (nid + 1));
+    if (data_.size() < (nid + 1)) {
+      data_.resize((nid + 1));
     }
 
-    row_ptr_[nid] = nbins_ * n_nodes_added_;
+    row_ptr_[nid] = n_nodes_added_;
     n_nodes_added_++;
   }
+  void AllocateData(bst_uint nid) {
```
trivialfis, line 448: When should these 2 functions used?

trivialfis:
> Sorry for wrong assumption. while (!expand_set.empty()) flow in UpdateTreeWithDriver looks very similar to lossguide

To me that's a good thing as now the code path is unified.

src/common/hist_util.cc
ShvetsKS: yes, reverted

src/common/row_set.h
ShvetsKS: as separate buffers were allocated for each tasks (to allow thread locality) there is no other than 0 usage.
Changed to `.data()` usage

src/common/row_set.h
ShvetsKS: done

src/common/hist_util.h
ShvetsKS: described and a little bit changed.

src/common/hist_util.h
ShvetsKS: it was documented

src/common/row_set.h
ShvetsKS: Another function is redundant, so it was removed

ShvetsKS:
@trivialfis seems failed step "continuous-integration/travis-ci/pr" is similar to failed ones in master branch: https://github.com/dmlc/xgboost/commit/4d1d5d401029b72f29a0227843b40d41f53944c6.

trivialfis:
Fixed in master by @hcho3.

ShvetsKS:
> Fixed in master by @hcho3.

seems all checks are passed

src/common/row_set.h
```diff
@@ -150,24 +150,31 @@ class PartitionBuilder {
...
+    if (mem_blocks_[id].size() == 0) {
+      mem_blocks_[id].resize(1);
+    }
+  }
+
   common::Span<size_t> GetLeftBuffer(int nid, size_t begin, size_t end) {
     const size_t task_idx = GetTaskIdx(nid, begin);
-    return { mem_blocks_.at(task_idx).Left(), end - begin };
+    return { mem_blocks_.at(task_idx).data()->Left(), end - begin };
```
trivialfis, line 162: Actually I'm confused by this, why do you need a vector if you are just using the first element?

ShvetsKS: @trivialfis  Sorry for long response.
vector was changed to unique_ptr

ShvetsKS: @trivialfis is it ok now?

trivialfis (approved):

src/common/row_set.h
```diff
@@ -150,24 +151,32 @@ class PartitionBuilder {
     }
   }
 
+  // allocate thread local memory, should be called for each specific task
+  void AllocateForTask(size_t id) {
+    if (mem_blocks_[id].get() == nullptr) {
+      BlockInfo* local_block_ptr = new BlockInfo;
```
SmirnovEgorRu, line 157: I prefer to add some checks for memory allocation here

SmirnovEgorRu (commented):
Overall, LGTM.
But one minor comment to add a memory check

src/common/row_set.h
ShvetsKS: done

SmirnovEgorRu (approved):

PR #6360, opened 2020-11-09
[jvm-packages] fix potential unit test suites aborted issue

wbo4958:
**Background:**
  1. XGBoost has enabled Stopping SparkContext by default while tasks run into exceptions,
  2. The Stopping SparkContext (`SparkContext.getOrCreate().stop()`) is called in a separate
     thread differring with thread running unit tests
  3. The unit tests are executed one by one in the same thread.
  
  Consider this situation, **TEST B** follows **TEST A** (**TEST A** will throw an exception in the worker)
 -  a) **TEST A** triggers [`SparkContext.getOrCreate().stop()`](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala#L158) which takes long time to be finished
         in a separate thread
-   b) **TEST A** calls [afterEach](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PerTest.scala#L47) which tries to stop SparkContext which is stopping because of **a)**
-  c) **TEST B** runs. first, calling [beforeEach](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PerTest.scala#L45) which creates SparkSession by wrapping a [SparkContext
       called by SparkContext.getOrCreate()](https://github.com/apache/spark/blob/branch-3.0/sql/core/src/main/scala/org/apache/spark/sql/SparkSession.scala#L934), Since SparkContext is a singleTon and kept by
       [activeContext](https://github.com/apache/spark/blob/branch-3.0/core/src/main/scala/org/apache/spark/SparkContext.scala#L2584) which was defined in SparkContext,
  
  But **SparkContext.stop** will clear activeContext in the last of stop code by [SparkContext.clearActiveContext()](https://github.com/apache/spark/blob/branch-3.0/core/src/main/scala/org/apache/spark/SparkContext.scala#L2038)
  **Here is the issue,**
  The newly created Sparksession of **TEST B** will wrap the SparkContext which is stopping by
  **TEST A**, which will result [sparkContext.assertNotStopped()](https://github.com/apache/spark/blob/branch-3.0/sql/core/src/main/scala/org/apache/spark/sql/SparkSession.scala#L103) throw exception and block the
   following unit tests in SparkSession.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PerTest.scala
+++ b/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PerTest.scala
@@ -19,7 +19,7 @@ package ml.dmlc.xgboost4j.scala.spark
 import java.io.File
 
 import ml.dmlc.xgboost4j.{LabeledPoint => XGBLabeledPoint}
-import org.apache.spark.{SparkConf, SparkContext, TaskFailedListener}
+import org.apache.spark.{SparkContext, SparkContextUtils, TaskFailedListener}
 import org.apache.spark.sql._
 import org.scalatest.{BeforeAndAfterEach, FunSuite}
 
@@ -50,6 +50,7 @@ trait PerTest extends BeforeAndAfterEach { self: FunSuite =>
         currentSession.stop()
         cleanExternalCache(currentSession.sparkContext.appName)
         currentSession = null
+        waitSparkContextTotallyStopped
       }
       TaskFailedListener.killerStarted = false
     }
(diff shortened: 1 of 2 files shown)


jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PerTest.scala
```diff
@@ -107,4 +108,34 @@ trait PerTest extends BeforeAndAfterEach { self: FunSuite =>
...
+  //      SparkContext.clearActiveContext()
+  // Here is the issue,
+  // The newly created Sparksession of TEST B will wrap the SparkContext which is stopping by
+  // TEST A, which will result sparkContext.assertNotStopped() throw exception and block the
+  //  following unit tests in SparkSession.
+
+  private def waitSparkContextTotallyStopped: Unit = {
+    var totalWaitedTime = 0L
+    while (!SparkContextUtils.getActiveSparkContext.isEmpty && totalWaitedTime <= 11000) {
+      Thread.sleep(1000)
```
trivialfis, line 136: Is this really a robust solution?

wbo4958: Yes, it's not a robust solution for now, since we really don't know when the SparkContext will be stopped in other thread. What we can do is set a reasonable timeout and check. I just checked the code, seems there is an explicit 21s timeout in the "stop" code. maybe I need to change the timeout to 22s.

CodingCat:
Hey, @wbo4958 thanks for the investigation and the fix , the busy loop is always fragile

I proposed a fix at https://github.com/dmlc/xgboost/pull/6373 based on wait/notify , please review

wbo4958:
Closed  this PR for Nan has a robust solution https://github.com/dmlc/xgboost/pull/6373

Issue #6307, opened 2020-10-28
Flaky test: TestGPUUpdaters::test_external_memory


trivialfis:
https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-6258/4/pipeline/207

```
[2020-10-28T10:53:16.632Z] tests/python-gpu/test_gpu_updaters.py::TestGPUUpdaters::test_gpu_hist_device_dmatrix PASSED
[2020-10-28T10:53:56.001Z] tests/python-gpu/test_gpu_updaters.py::TestGPUUpdaters::test_external_memory double free or corruption (fasttop)
[2020-10-28T10:53:56.001Z] Fatal Python error: Segmentation fault
[2020-10-28T10:53:56.001Z] 
[2020-10-28T10:53:56.001Z] Thread 0x00007f3f329ba700 (most recent call first):
[2020-10-28T10:53:56.001Z]   File "/opt/python/Fatal Python error: Aborted
[2020-10-28T10:53:56.001Z] 
[2020-10-28T10:53:56.001Z] tests/ci_build/test_python.sh: line 42:  4497 Aborted                 (core dumped) pytest -v -s -rxXs --fulltrace --durations=0 -m "not mgpu" ${args} tests/python-gpu
```

https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-6302/3/pipeline
```
[2020-10-28T10:52:27.516Z]         ret : int
[2020-10-28T10:52:27.516Z]             return value from API calls
[2020-10-28T10:52:27.516Z]         """
[2020-10-28T10:52:27.516Z]         if ret != 0:
[2020-10-28T10:52:27.516Z] >           raise XGBoostError(py_str(_LIB.XGBGetLastError()))
[2020-10-28T10:52:27.516Z] E           xgboost.core.XGBoostError: std::bad_alloc
[2020-10-28T10:52:27.516Z] 

[2020-10-28T10:52:27.516Z] /home/ubuntu/.local/lib/python3.7/site-packages/xgboost/core.py:186: XGBoostError

[2020-10-28T10:52:27.516Z] ---------------------------------- Hypothesis ----------------------------------
[2020-10-28T10:52:27.516Z] Falsifying example: test_external_memory(
[2020-10-28T10:52:27.516Z]     self=<test_gpu_updaters.TestGPUUpdaters at 0x7ff94cf55050>,
[...]

PR #6361, opened 2020-11-09
Deterministic external memory.

trivialfis:
Replaces https://github.com/dmlc/xgboost/pull/6317 for addressing comment from @RAMitchell .

Close: #5063
Close: #6307


trivialfis:
@RAMitchell I'm not sure if this is better than the previous approach.  In this PR, the data pool is involved even if external memory is not being used. (single batch file).

PR #6363, opened 2020-11-10
Expose global verbosity in shared library but do not export in R Package

meztez:
There was no way to make xgboost silent from the R Package (none that I could find) so what do think about exposing ConsoleLogger::Configure in R?

Thoughts?

It is mainly to reduce the amount of 
```
...
Loading model from XGBoost < 1.0.0, consider saving it again for improved compatibility
...
```

I get in the server log while people call some R Api repeatedly.

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -7,6 +7,14 @@
 # General helper utilities ----------------------------------------------------
 #
 
+# Set global xgboost verbosity
+xgb.set.global.verbosity <- function(verbosity = c("warning","silent","info","debug")) {
+  verbosity <- match.arg(verbosity)
+  verbosity <- switch(verbosity, "silent" = "0", "warning" = "1", "info" = "2","debug" = "3")
+  .Call(XGSetGlobalVerbosity_R, verbosity)
+  return(invisible(verbosity))
+}
+
 # SQL-style NVL shortcut.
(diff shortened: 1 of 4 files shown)


trivialfis:
I don't think this will be merged.

> Loading model from XGBoost < 1.0.0, consider saving it again for improved compatibility

Consider saving it again?

meztez:
They are going to get there eventually, they just have to go through a lot of them it seems. Some 50-60 assets it seems?
While they are doing that, the messages are clogging the system logs with no way for them to control it.

This is mainly to expose a cli available options in R.

trivialfis:
That sounds bad.  Let me look into it.

meztez:
I told them to use

```R
invisible(capture.output(suppressWarnings({pred <- predict(bst, agaricus.test$data)})))
```

trivialfis:
@meztez Thanks for the workaround.  In fact we have something similar in mind for a while.  I didn't approve the PR is because we want it to be more general with more options other than verbosity, and portable to other language bindings.

trivialfis:
@hcho3 Will follow up on the PR. ;-)

trivialfis:
Related: https://github.com/dmlc/xgboost/issues/5011 https://github.com/dmlc/xgboost/issues/4600

PR #6367, opened 2020-11-10
Update custom_metric_obj.rst

JeanLescut:
Just adding an import that was missing.
We could also remove the typing hint, but I chose to keep it as it is.


trivialfis (approved):
Thanks for catching it.

PR #6372, opened 2020-11-11
[CI] Remove R check from Jenkins

hcho3:
Now that we are running R check in GitHub Actions, remove it from Jenkins.


tests/ci_build/print_r_stacktrace.sh
```diff
@@ -1,19 +0,0 @@
-#!/bin/bash
-
```
trivialfis, line 2: Do we need this script for action?

hcho3: Good point. Let me put it back and update the action yml.

trivialfis (approved):

PR #6373, opened 2020-11-11
[jvm-packages] fix potential unit test suites aborted issue

CodingCat:
problem described https://github.com/dmlc/xgboost/pull/6360 , credits to @wbo4958 for the investigation

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
@@ -150,14 +150,18 @@ object TaskFailedListener {
 
   var killerStarted: Boolean = false
 
+  var sparkContextKiller: Thread = _
+
   val sparkContextShutdownLock = new AnyRef
+
+
 
   private def startedSparkContextKiller(): Unit = this.sparkContextShutdownLock.synchronized {
     if (!killerStarted) {
       killerStarted = true
       // Spark does not allow ListenerThread to shutdown SparkContext so that we have to do it
       // in a separate thread
-      val sparkContextKiller = new Thread() {
+      sparkContextKiller = new Thread() {
         override def run(): Unit = {
           LiveListenerBus.withinListenerThread.withValue(false) {
             sparkContextShutdownLock.synchronized {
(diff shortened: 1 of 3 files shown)


jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
```diff
@@ -146,22 +148,29 @@ class TaskFailedListener(killSparkContext: Boolean = true) extends SparkListener
...
-    if (!killerStarted) {
+    if (!killerStarted.get()) {
       // Spark does not allow ListenerThread to shutdown SparkContext so that we have to do it
       // in a separate thread
       val sparkContextKiller = new Thread() {
         override def run(): Unit = {
           LiveListenerBus.withinListenerThread.withValue(false) {
             SparkContext.getOrCreate().stop()
+            sparkContextShutdownLock.synchronized {
+              killerStarted.set(false)
```
wbo4958, line 165: I don't know whether there is a situation triggerring multiple sparkContextKiller?

one task failure will trigger 1 sparkContextKiller, and after killerStarted.set(false). another task failure will trigger another sparkContextKiller ?

CodingCat: it will not, when we reach here, it means that we have returned from SparkContext.stop() which cleaned all msgs (including TaskFailedEvent)...in the other word, when we want to set this to false, all attempts of starting sparkContextKiller should have been gone through...in the successive attempts, since we haven't set this to false, they will not start new killers

jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/SparkParallelismTracker.scala
```diff
@@ -16,6 +16,8 @@
 
 package org.apache.spark
 
+import java.util.concurrent.atomic.AtomicBoolean
```
wbo4958, line 19: Seems this line should be deleted.

wbo4958 (approved):
LGTM

hcho3:
@wbo4958 @CodingCat Should this be part of 1.3.0 release?

CodingCat:
> @wbo4958 @CodingCat Should this be part of 1.3.0 release?

yeah, hopefully I should merge it within 1 hour

PR #6374, opened 2020-11-11
[R] Fix warnings from R check --as-cran

trivialfis:
* Remove printf in rabit.
* Run `devtools::document` and adjust some doc for linter.


hcho3 (approved):
LGTM, as long as tests pass.

PR #6375, opened 2020-11-11
[dask] Fix union of workers.

trivialfis:
* Fix union of workers.
* Submit with future of `DMatrix` directly, simplifying training.
* Avoid the use of `**kwargs`.


trivialfis:
I want to keep the dask module small as possible.  Right now the training method is shrunk down to 70 LOC, with the heavy part in `DaskDMatrix`.  The difficulty is in preserving order of partitions for `predict` method.

RAMitchell (approved):

trivialfis:
Ended reverting all the simplification due to needing rabit synchronization.

trivialfis:
I tried using `delayed`, but passing a list of `delayed` objects has undesired behavior.

PR #6378, opened 2020-11-12
[R] Fix a crash that occurs with noLD R

### LINKED ISSUE #5935: NoLD issue in xgb.train and/or xgb.attr
CRAN has notified me of a NoLD problem in one of my R-packages that seems to be attributable to XGBoost. Unfortunately I haven't been able to figure out a way to test the NoLD setup that CRAN has without actually submitting the package to CRAN. If you could provide any assistance to help figure out where the issue is coming from so I can report back to CRAN without getting several of my packages archived that we be wonderful.

```
Error in finalizer(env) : 
  Inconsistent 'best_score' values between the closure state: 0.826267 and the xgb.attr: 0.826267
Calls: %>% ... do.call -> <Anonymous> -> xgb.train -> f -> finalizer
```

https://www.stats.ox.ac.uk/pub/bdr/noLD/radiant.model.out

Relevant source code: https://github.com/radiant-rstats/radiant.model/blob/master/R/gbt.R

hcho3:
Closes #5935. Unfortunately, no unit test was added, because it's quite difficult to setup an environment with noLD R (it involves building R from the source with a custom flag).

@vnijs @jameslamb  Can you review?

### DIFF
--- a/R-package/R/callbacks.R
+++ b/R-package/R/callbacks.R
@@ -352,9 +352,15 @@ cb.early.stop <- function(stopping_rounds, maximize = FALSE,
   finalizer <- function(env) {
     if (!is.null(env$bst)) {
       attr_best_score <- as.numeric(xgb.attr(env$bst$handle, 'best_score'))
-      if (best_score != attr_best_score)
-        stop("Inconsistent 'best_score' values between the closure state: ", best_score,
-             " and the xgb.attr: ", attr_best_score)
+      if (best_score != attr_best_score) {
+        # If the difference is too big, throw an error
+        if (abs(best_score - attr_best_score) >= 1e-14) {
+          stop("Inconsistent 'best_score' values between the closure state: ", best_score,
+               " and the xgb.attr: ", attr_best_score)
+        }
+        # If the difference is due to floating-point truncation, update best_score
+        best_score <- attr_best_score
+      }
       env$bst$best_iteration <- best_iteration
       env$bst$best_ntreelimit <- best_ntreelimit
       env$bst$best_score <- best_score


vnijs:
I'm getting errors compiling in the NoLD docker container. Is it possible to test this using something like the below?

```
install.packages("drat", repos="https://cran.rstudio.com")
drat:::addRepo("dmlc")
install.packages("xgboost", repos="http://dmlc.ml/drat/", type = "source")
```

hcho3:
No, `dmlc.ml` does not work any more. You can follow the instructions from the [official doc](https://xgboost.readthedocs.io/en/latest/build.html#installing-the-development-version-linux-mac-osx):
```
git clone --recursive https://github.com/dmlc/xgboost
cd xgboost
mkdir build
cd build
cmake .. -DR_LIB=ON
make -j$(nproc)
make install
```

trivialfis:
Can you try to change `a != b` to `a - b != 0` and see what happens?

hcho3:
> Can you try to change a != b to a - b != 0 and see what happens?

There is no difference. The example fails without this patch, and with this patch the example works.

trivialfis:
I mean change the patch to `best_score - attr_best_score != 0` instead of using an absolute tolerance.

hcho3:
@trivialfis Without the absolute tolerance condition, the example script will fail with 
```
Error in finalizer(env) : 
  Inconsistent 'best_score' values between the closure state: 0.829749 and the xgb.attr: 0.829749
```
The difference between `best_score` and `attr_best_score` is only 1e-16. I added the tolerance check to allow for a slight error due to floating-point arithmetic.

trivialfis (approved):

vnijs:
@hcho3 I was able to build the `fix_nold` branch using your instructions. I can confirm that this works in the NoLD environment for R! 

That said, I would still recommend adding a test for CRAN in case a similar issue pops up in the future. That way it will be easier to track down where this issue started :) Thanks!

hcho3:
@vnijs Thanks for reviewing.

> I would still recommend adding a test for CRAN in case a similar issue pops up in the future

I'm afraid we won't act on this recommendation. Testing with noLD environment requires building R from the source, and that's not something we want to spend our testing budget.

jameslamb:
> Testing with noLD environment requires building R from the source, and that's not something we want to spend our testing budget.

Sorry I'm late to the party! I had never tested in a noLD environment before, so I learned a lot from this. Thanks for including me!

I've never use
d it before, but just noting that there is a noLD docker image from rhub: https://hub.docker.com/r/rhub/debian-gcc-devel-nold/tags. If you wanted, you could set up a job that runs the R build and tests in it only when triggered by a comment, so it doesn't have to be on every PR. For example, you could do that only on releases.

That's what we do for valgrind in LightGBM: https://github.com/microsoft/LightGBM/blob/master/.github/workflows/r_valgrind.yml#L10

vnijs:
@jameslamb I did try rhubs images and tools for NoLD but could never get them to work for some reason. So I eventually "rolled-my-own", more or less. 

@hcho3 I don't think the test needs to be part of xgboost development really as I can see that might be a hassle. Just that there is a (testthat) test that will be checked by CRAN in a NoLD environment when a new release is available. This could be as simple as the reproducible example I provided.

hcho3:
@jameslamb Thanks for the workflow template. I didn't realize that you can trigger a job with a pull request comment. Interesting! This is going to be super useful.

jameslamb:
> @jameslamb Thanks for the workflow template. I didn't realize that you can trigger a job with a pull request comment. Interesting! This is going to be super useful.

@StrikerRUS taught me that 😀

PR #6379, opened 2020-11-12
[dask] Allow empty data matrix in AFT survival

### LINKED ISSUE #6139: Allow empty datasets in `survival:aft` objective and `aft-nloglik` metric
Currently, `survival:aft` objective and `aft-nloglik` metric throws an error when an empty data matrix is passed in. Empty data matrix is a common occurrence when XGBoost is used with Dask, due to quirks in how Dask partitions the data matrix.

The AFT objective and metric should handle empty data matrices like other element-wise objective and metrics, such as `reg:squarederror` and `logloss`.

cc @trivialfis

hcho3:
Closes #6139

### DIFF
--- a/src/metric/survival_metric.cu
+++ b/src/metric/survival_metric.cu
@@ -206,10 +206,6 @@ struct EvalEWiseSurvivalBase : public Metric {
   bst_float Eval(const HostDeviceVector<bst_float>& preds,
                  const MetaInfo& info,
                  bool distributed) override {
-    CHECK_NE(info.labels_lower_bound_.Size(), 0U)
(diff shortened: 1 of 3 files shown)


trivialfis (approved):

trivialfis:
Actually you might need to add a unittest.  The dask interface is no longer using all workers

hcho3:
Any suggestion? Is it sufficient to pass in an empty `DMatrix`?

trivialfis:
Yup.

trivialfis (approved):
LGTM

PR #6380, opened 2020-11-12
[R] Do not convert continuous labels to factors

### LINKED ISSUE #5829: [R] generate.cv.folds() converts all label vectors with objectives not reg:squarederror to factors
Thanks for making such a great tool. I recently made the switch from v0.90 to v1.0.0 and noticed some unexpected behavior. Now that `xgb.cv()` pulls the `label` directly from the `xgb.DMatrix` object, all `label` vectors are converted to factors in `generate.cv.folds()` prior to calling `xgb.createFolds()` when the objective is not `reg:squarederror`. This means that continuous `label` vectors from nonstandard objective functions, like `count:poisson`, `reg:gamma`, `reg:tweedie`, etc., are converted to factors prior to generating the folds. This process can drastically slow down the creation of folds. I think there are a couple of potential solutions (assuming this is not the desired behavior in all cases):

1. Added a parameter in `xgb.cv()` that forces `generate.cv.folds()` to treat `label` as continuous
2. Include objectives with potentially continuous labels in this `generate.cv.folds()` [filter](https://github.com/dmlc/xgboost/blob/master/R-package/R/utils.R#L213)

hcho3:
Closes #5829

cc @benmarchi @jameslamb @mayer79

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -187,6 +187,18 @@ xgb.iter.eval <- function(booster_handle, watchlist, iter, feval = NULL) {
 # Helper functions for cross validation ---------------------------------------
 #
 
+# Possibly convert the labels into factors, depending on the objective.
+# The labels are converted into factors only when the given objective refers to the classification
+# or ranking tasks.
+convert.labels <- function(labels, objective_name) {
+  if (objective_name %in% c('binary:logistic', 'binary:logitraw', 'binary:hinge', 'multi:softmax',
+                            'multi:softprob', 'rank:pairwise', 'rank:ndcg', 'rank:map')) {
+    return(factor(labels))
+  } else {
+    return(labels)
+  }
+}
+
 # Generates random (stratified if needed) CV folds
 generate.cv.folds <- function(nfold, nrows, stratified, label, params) {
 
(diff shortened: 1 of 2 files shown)


trivialfis (commented):
Can we leave this part to the user?  I don't want to encode any c++ parameter in language binding.

hcho3:
Currently `generate.cv.folds` generates stratified folds, where the labels determine how the folds are stratified. So the detection of the task type is necessary.

jameslamb:
> I don't want to encode any c++ parameter in language binding.

I disagree with this. I think it's ok for language bindings to have logic that changes based on what the objective is.

But @hcho3 instead of hard-coding a vector or literals, I think it would be better to give this vector an informative name and store it outside of `generate.cv.folds()`.

As an example, we do that in LightGBM for early stopping: https://github.com/microsoft/LightGBM/blob/1bc27939a43c414c3424f339994d0aa11f3aa3b1/R-package/R/metrics.R#L9

R-package/R/utils.R
```diff
@@ -187,6 +187,18 @@ xgb.iter.eval <- function(booster_handle, watchlist, iter, feval = NULL) {
 # Helper functions for cross validation ---------------------------------------
 #
 
+# Possibly convert the labels into factors, depending on the objective.
+# The labels are converted into factors only when the given objective refers to the classification
+# or ranking tasks.
+convert.labels <- function(labels, objective_name) {
+  if (objective_name %in% c('binary:logistic', 'binary:logitraw', 'binary:hinge', 'multi:softmax',
+                            'multi:softprob', 'rank:pairwise', 'rank:ndcg', 'rank:map')) {
+    return(factor(labels))
```
jameslamb, line 196: I think `as.factor()` would be better. It can be faster, especially if `labels` is already a factor when this is called.

From `?factor`

> as.factor coerces its argument to a factor. It is an abbreviated (sometimes faster) form of factor.

hcho3: Didn't know that! Thanks! I'll make the change.

R-package/R/utils.R
```diff
@@ -208,17 +220,15 @@ generate.cv.folds <- function(nfold, nrows, stratified, label, params) {
     #  - For regression, leave y numeric and do stratification by quantiles.
     if (exists('objective', where = params) &&
```
jameslamb, line 221: I know this code is not new to this PR, but I think `exists()` is dangerous here. If someone explicitly set `objective = NULL`  in `params`, this would evaluate to TRUE.

image

Then when you pass `NULL` into `convert.labels()`, it will result in a hard-to-understand error.

image

Would you be open to a PR that fixes these uses of `exists()`?

hcho3: Yes, let's fix it in a follow-up PR.

jameslamb: ok sounds good! I'd be happy to contribute that, I think it'll be straightforward.

R-package/R/utils.R
hcho3: See my latest commit.

trivialfis (approved):

Issue #6350, opened 2020-11-06
prediction different in v1.2.1 and master branch


7starsea:
I am using the released version 1.2.1 for development due to its easy installation in python and I also write a simple c code for prediction using the c_api.  If linked to v1.2.1 libxgboost.so, the difference of the predictions between python and c is exactly zero. However, if linked to libxgboost.so from the master branch (commit f3a425398 at Nov. 5, 2020), there is a difference.

I would like to deploy the c code in real system using the master branch since I would like to build a static lib, and now the difference of the predictions between v1.2.1 and master branch hinders me.

Thanks.

hcho3:
Can you post an example program so that we can reproduce the bug?

hcho3:
Also, do you see different predictions from the master version of Python package and the C API?

trivialfis:
There are some optimization done on the CPU predictor, might generate different result by different floating point error.  But yeah, do you have a reproducible example?

7starsea:
> Also, do you see different predictions from the master version of Python package and the C API?

I am comparing the v1.2.1 of python package and the master branch of C API.

hcho3:
@7starsea Can you also compare the outputs from the Python and C API, both from the master branch? The issue could be the way C API functions are used in your application.

7starsea:
> @7starsea Can you also compare the outputs from the Python and C API, both from the master branch? The issue could be the way C API functions are used in your application.

I compared the python and c api, both from v1.2.1 and the predictions are exactly the same.

hcho3:
@7starsea  Got it. If you post both Python and C programs that predict from the same model, we'd be able to troubleshoot the issue further.

7starsea:
@hcho3 here is the testing code

https://github.com/7starsea/xgboost-testing

hcho3:
@7starsea I just tried your example and got the following output:
```
difference: [0. 0. 0. 0.] 0.0 0.0
```

I used the latest commit of XGBoost (debeae2509d90ec1d3402a3a185fba7a25113ff1).

7starsea:
> @7starsea I just tried your example and got the following output:
> 
> ```
> difference: [0. 0. 0. 0.] 0.0 0.0
> ```
> 
> I used the latest commit of XGBoost ([debeae2](https://github.com/dmlc/xgboost/commit/debeae2509d90ec1d3402a3a185fba7a25113ff1)).

Interesting, is your python version v1.2.1?

[...]

PR #6384, opened 2020-11-12
Fix predict difference

ShvetsKS:
related issue: https://github.com/dmlc/xgboost/issues/6350

### DIFF
--- a/src/predictor/cpu_predictor.cc
+++ b/src/predictor/cpu_predictor.cc
@@ -52,10 +52,12 @@ inline bst_float PredValueByOneTree(const RegTree::FVec& p_feats,
 inline void PredictByAllTrees(gbm::GBTreeModel const &model, const size_t tree_begin,
                               const size_t tree_end, bst_float* preds,
                               const std::vector<RegTree::FVec> &thread_temp,
-                              const size_t offset, const size_t block_size) {
+                              const size_t offset, const size_t block_size,
+                              const size_t num_group) {
   for (size_t tree_id = tree_begin; tree_id < tree_end; ++tree_id) {
+    const size_t gid = model.tree_info[tree_id];
     for (size_t i = 0; i < block_size; ++i) {
-      preds[i] += PredValueByOneTree(thread_temp[offset + i], model.trees[tree_id]);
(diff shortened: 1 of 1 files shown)


trivialfis (commented):
Could you please provide a brief explanation for what happened?  Also a unittest.  Lastly I see that you are adding an extra allocation, does it impact perf?

trivialfis:
Ah, I saw your reply on the issue.  Will continue there.

ShvetsKS:
> Could you please provide a brief explanation for what happened? Also a unittest. Lastly I see that you are adding an extra allocation, does it impact perf?

The reason of difference in output result is changed sequence of floating point operations:
Before a4ce0ea optimizations we accumulated trees responses for each sample in local variable `psum` with `0` initial value. And than increment `out_pred[i]`.
But in a4ce0ea optimizations we increment `out_pred[i]` directly (initial value 0.5 usually), so the fp error was a little bit different.

There is already implemented cpp tests for cpu predictor (`cpu_predictor->PredictBatch`), there is no significant changes. Or you mentioned about `ReduceResult`?

I checked it on several benchmarks, seems there is no performance affection.

ShvetsKS:
https://xgboost-ci.net/blue/organizations/jenkins/xgboost-win64/detail/PR-6384/3/pipeline#step-89-log-1987
Not sure that it's related to current changes as locally all tests from t`ests/python/test_with_sklearn.py` are passed.

trivialfis:
@ShvetsKS Let's pause on this with comment: https://github.com/dmlc/xgboost/issues/6350#issuecomment-726053585 .  We @RAMitchell @hcho3 agreed that we should not establish the tradition of fixing floating change across versions, as you stated, that will severely limit our development.  We haven't decided how to document or whether do need to document this.

trivialfis:
But to reply your question on the failing CI.  Yeah, it happens all the time that error is only reproducible on CI ...  Something I have been bumping into quite often.  The error in this PR is new to me, so it might be specific to this PR instead of CI glitches.  You can see on our issues list for all flaky tests we have found.

ShvetsKS:
@trivialfis  seems I should close this PR as #6350 was resolved? :)

hcho3:
Let us continue our discussion in #6350.

Issue #6283, opened 2020-10-23
DMatrix creation takes a big part of training and prediction time


RukhovichIV:
# Description of the problem

We did some measurements and found out that DMatrix creation takes a huge part of any operation in XGBoost with both **pandas DataFrame** and **numpy array** with float32 values.
It takes  **up to 90% of the prediction time** on our datasets and **up to 35% of the training time** on it.

|                  | Pandas.DataFrame |                 | Numpy.ascontigiousarray |                 |                      |
|------------------|------------------|-----------------|-------------------------|-----------------|----------------------|
| operation        | time, s          | dmatrix time, % | time, s                 | dmatrix time, % | numpy improvement, s |
| airline, train    | 36.5497          | 14.6913         | 35.7296                 | 8.8368          | 0.8201               |
| airline, predict  | 5.9467           | 97.6645         | 3.6844                  | 79.7643         | 2.2623               |
| higgs, train      | 14.6953          | 3.2722          | 15.1656                 | 2.9845          | -0.4703              |
| higgs, predict    | 1.0596           | 46.8678         | 1.0114                  | 44.8713         | 0.0482               |
| msrank, train     | 77.4849          | 5.6062          | 74.3278                 | 3.5626          | 3.1571               |
| msrank, predict   | 5.6924           | 68.3226         | 4.3200                  | 56.0864         | 1.3724               |
| mortgage, train   | 19.4234          | 32.4388         | 20.8129                 | 21.7833         | -1.3895              |
| mortgage, predict | 6.3151           | 91.5222         | 4.7360                  | 88.5982         | 1.5791               |

# What can be done

- XGBoost already has the `inplace_predict` method, which allows us to do prediction for data in numpy array without copying it. We'd like to discuss the possibility **to extend** this function with pandas DataFrame.
- If we go deeper into the problem, `inplace_predict` is not the default function and it seems like **most of the XGBoost users are not familiar with it**. Also, for obvious reasons, we can't make a similar function for training, which also works longer than it could be. **Is there a need for copying the input data for every DMatrix object? Is it possible to add the additional `no_copy` flag to avoid copying of the input data in case of usage float32 type?**

trivialfis:
For training with GPU hist and GPU input, you can use `DeviceQuantileDMatrix`, which constructs the histogram index inplace.  This is mostly for better memory consumption instead of speed.  You will notice longer construction time with it, but the overall training time should be faster.

[...]

PR #6386, opened 2020-11-12
Disable HT for DMatrix creation

ShvetsKS:
Recently introduced `OmpSetNumThreads` function brings up to 2x slowdown on HW with enabled Hyper-threading.

DMatrix create time| master, s | this PR, s | Speedup
-- | -- | -- | --
mortgage | 11.67 | 5.68 | 2.05
higgs | 0.96 | 0.51 | 1.88
msrank | 4.36 | 3.53 | 1.23

It significantly affects inference stage as was mentioned in https://github.com/dmlc/xgboost/issues/6283.

Reason: there is still overheads on memory allocation\initialization (`InitBudget`, `InitStorage`) so there is no gain from usage HT here.

### DIFF
--- a/src/common/threading_utils.h
+++ b/src/common/threading_utils.h
@@ -160,6 +160,16 @@ inline int32_t OmpSetNumThreads(int32_t* p_threads) {
   omp_set_num_threads(threads);
   return nthread_original;
 }
+inline int32_t OmpSetNumThreadsWithoutHT(int32_t* p_threads) {
+  auto& threads = *p_threads;
+  int32_t nthread_original = omp_get_max_threads();
+  if (threads <= 0) {
+    threads = nthread_original;
+  }
+  omp_set_num_threads(threads);
+  return nthread_original;
+}
+
 }  // namespace common
 }  // namespace xgboost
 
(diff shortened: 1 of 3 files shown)


trivialfis (commented):
So the default setting doesn't account for HT?  Also, please don't change the file permission.

ShvetsKS:
> So the default setting doesn't account for HT? Also, please don't change the file permission.

@trivialfis by default `nthreads == -1` for DMatrix create, and `omp_get_num_procs` returns 2x of hardware threads with enabled HT. So it has negative impact on time of DMatrix creation always. That's why I reverted initial behavior (before introduction `OmpSetNumThreads` function) for DMatrix.

Files permissions were reverted.

ShvetsKS:
All tests were passed for previous commit 4320275
Current run failed due to exceeded budget

ShvetsKS:
@trivialfis would it be better to wait refreshed budget to see all passed tests again?

trivialfis:
Yup.  Will merge once all tests pass.

trivialfis (approved):

PR #6406, opened 2020-11-19
[CI] Upgrade to MacOS Mojave image

hcho3:
Homebrew just added support for MacOS Big Sur (11.0) and dropped High Sierra (10.13), according to their policy of supporting three latest MacOS releases. So we require MacOS Mojave from now on.

@trivialfis This should unblock the CI.

### DIFF
--- a/tests/travis/run_test.sh
+++ b/tests/travis/run_test.sh
@@ -44,7 +44,7 @@ if [ ${TASK} == "python_test" ]; then
       cd ../python-package
       python setup.py bdist_wheel
       cd ..
-      TAG=macosx_10_13_x86_64.macosx_10_14_x86_64.macosx_10_15_x86_64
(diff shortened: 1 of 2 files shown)


tests/travis/run_test.sh
```diff
@@ -44,7 +44,7 @@ if [ ${TASK} == "python_test" ]; then
       cd ../python-package
       python setup.py bdist_wheel
       cd ..
-      TAG=macosx_10_13_x86_64.macosx_10_14_x86_64.macosx_10_15_x86_64
+      TAG=macosx_10_14_x86_64.macosx_10_15_x86_64.macosx_11_0_x86_64
```
trivialfis, line 47: Oh, here it is. ;-)

trivialfis (approved):

tests/travis/run_test.sh
hcho3: Also see `osx_image` in `.travis.yml`.

Issue #6297, opened 2020-10-27
Add an explicit option to use RMM


hcho3:
Moved from https://github.com/rapidsai/rmm/issues/583.

We should ask the user to explicitly declare that they are using RMM allocator. The desired behavior is as follows:

* If `use_rmm` is set to True: XGBoost will use RMM allocator for every allocation, trusting the users to set a suitable device resource for good performance.
* If `use_rmm` is set to False: XGBoost will use CUB allocator as well as plain `cudaMalloc()`. We trust the users to remember setting `use_rmm=True` when they set up an RMM memory pool.
* `use_rmm` should be set to False by default.

PR #6414, opened 2020-11-20
Add global configuration

### LINKED ISSUE #5011: GPU memory monitor fails to log allocation.

hcho3:
- [x] Add management functions for global configuration: `XGBSetGlobalConfig()`, `XGBGetGlobalConfig()`.
- [x] Add Python interface: `set_config()`, `get_config()`, and `config_context()`.
- [x] Add unit tests for Python
- [x] Add R interface: `xgb.set.config()`, `xgb.get.config()`
- [x] Add unit tests for R

Example (Python):
```python
import xgboost as xgb

# Suppress warning caused by model generated with XGBoost version < 1.0.0
with xgb.config_context(verbosity=0):
    bst = xgb.Booster(model_file='./old_model.bin')
```

Example (R):
```R
library(xgboost)

# Set verbosity level to silent (0)
xgb.set.config(verbosity = 0)
# Now global verbosity level is 0
config <- xgb.get.config()
print(config$verbosity)
# Set verbosity level to warning (1)
xgb.set.config(verbosity = 1)
# Now global verbosity level is 1
config <- xgb.get.config()
print(config$verbosity)
```

Closes #6363
Closes #5011

We can add more parameters in a future PR.


python-package/xgboost/__init__.py
```diff
@@ -29,4 +30,5 @@
            'RabitTracker',
            'XGBModel', 'XGBClassifier', 'XGBRegressor', 'XGBRanker',
            'XGBRFClassifier', 'XGBRFRegressor',
-           'plot_importance', 'plot_tree', 'to_graphviz', 'dask']
+           'plot_importance', 'plot_tree', 'to_graphviz', 'dask',
+           'set_config', 'get_config', 'config_context']
```
trivialfis, line 34: Is it necessary to have explicit getter and setter?

python-package/xgboost/config.py
```diff
@@ -0,0 +1,101 @@
...
+    Context manager for global XGBoost configuration
+
+    .. note::
+
+        All settings, not just those presently modified, will be returned to their previous values
+        when the context manager is exited. This is not thread-safe.
+
+    .. versionadded:: 1.3.0
+
+    Parameters
```
trivialfis, line 73: The parameter is `**new_config`, not `verbosity`.

python-package/xgboost/config.py
```diff
@@ -0,0 +1,101 @@
+"""Global configuration for XGBoost"""
+import ctypes
+from contextlib import contextmanager
+
+from .core import _LIB, _check_call
+
+
+def set_config(*, verbosity=None):
```
trivialfis, line 8: Can we avoid listing parameter and use opaque strings similar to `booster.set_param`.

src/c_api/c_api.cc
```diff
@@ -46,6 +47,32 @@ XGB_DLL int XGBRegisterLogCallback(void (*callback)(const char*)) {
   API_END();
 }
 
+XGB_DLL int XGBSetGlobalConfig(const char** name, const char** value, size_t num_param) {
```
trivialfis, line 50: Another option is use JSON.  We have `booster.load/save_config`, maybe we can align with it.

src/global_config.cc
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+#include <dmlc/thread_local.h>
+#include "xgboost/global_config.h"
+
+namespace xgboost {
+
+void GlobalConfiguration::SetArgs(const Args& args) {
+  for (const auto& e : args) {
+    if (e.first == "verbosity") {
```
trivialfis, line 15: I think we don't need to check for str equality.  The `XGBoostParameter` struct can return all unknown parameters and consume all known parameters automatically.

src/logging.cc
```diff
@@ -52,7 +52,14 @@ bool ConsoleLogger::ShouldLog(LogVerbosity verbosity) {
...
 
 void ConsoleLogger::Configure(Args const& args) {
-  param_.UpdateAllowUnknown(args);
+  auto unknown = param_.UpdateAllowUnknown(args);
+  if (!unknown.empty()) {
+    std::ostringstream oss;
+    for (const auto& e : unknown) {
+      oss << e.first << ", ";
+    }
+    LOG(FATAL) << "Unknown parameters: " << oss.str();
```
trivialfis, line 61: Ah, you are already using it.  Can we place this outside of logger?

trivialfis:
Some early review comments, for discussion.

python-package/xgboost/__init__.py
hcho3: I was following sklearn's config manager: https://scikit-learn.org/stable/modules/generated/sklearn.config_context.html

python-package/xgboost/config.py
hcho3: This is similar to sklearn's config manager: https://github.com/scikit-learn/scikit-learn/blob/0fb307bf3/sklearn/_config.py#L85

src/global_config.cc
hcho3: Do you want me to create a new `XGBoostParameter` struct? I didn't do that because I thought you didn't like duplicating parameters. If I create `GlobalConfigurationParams` struct, it will duplicate `verbosity` parameter.

src/logging.cc
hcho3: Actually I will need to remove this check from `ConsoleLogger::Configure()`, as it's failing lots of tests.

python-package/xgboost/config.py
hcho3: It's a bit nicer to explicitly parameters, so that the Python API doc will show it. Otherwise, users won't know which parameters are in the global scope. See https://scikit-learn.org/stable/modules/generated/sklearn.config_context.html

src/c_api/c_api.cc
hcho3: Didn't think of that. Yes, that would reduce the boilerplace a bit.

src/global_config.cc
hcho3: Actually duplicating parameters may be unavoidable, since we are already doing that in the Python layer. Let me go ahead and create `GlobalConfigurationParams`.

python-package/xgboost/config.py
hcho3: Another option is to list global parameters in https://xgboost.readthedocs.io/en/latest/parameter.html.

python-package/xgboost/__init__.py
hcho3: Also, the context manager is implemented in terms of the getter and the setter.

src/global_config.cc
trivialfis: Please don't duplicate parameters.  We can have a deeper look into it later.

hcho3: Got it. I won't create `GlobalConfigurationParams`

python-package/xgboost/config.py
hcho3: I'm going to honor the spirit of avoiding duplication, and will avoid listing parameter in the Python later. Instead, we'll have a documented list in `parameter.rst`.

src/c_api/c_api.cc
hcho3: @trivialfis One issue is that R does not have a built-in JSON parser, so we'd need to add a new dependency on the R package to add global config to the R package. On the other hand, the current approach does not require a new dependency.

python-package/xgboost/config.py
hcho3: See the latest commit, which uses opaque strings like `set_param`.

python-package/xgboost/config.py
hcho3: Updated it to `**new_config`, as the parameters are no longer explicitly listed in Python.

trivialfis:
Is it necessary for this to be blocking?  I thought it's going into the next release?

hcho3:
@trivialfis This PR does not contain `use_rmm` or `gpu_id`. Were we not planning to address #6363 in this release?

trivialfis:
I usually take some time for API changes (addition counts).  Some comments from @RAMitchell would also be nice.  My opinion would be we branch out whatsoever, and back port this PR if it can be merged before rc2 or final release if there's no rc2.  This way we don't have to rush.

src/c_api/c_api.cc
hcho3: Implementing `get_config` in R turns out to be a headache, since I have to write R extension code that returns a dictionary. For the ease of maintenance, I propose that we make `jsonlite` a required dependency of XGBoost, so that `get_config` can simply return a JSON string.

hcho3:
Got it. This PR won't block RC1.

hcho3:
I added the R interface.

hcho3:
@meztez This is replacement for #6363. Can you review?

R-package/tests/testthat/test_config.R
```diff
@@ -0,0 +1,13 @@
+context('Test global configuration')
+
+require(xgboost)
```
jameslamb, line 3: this isn't necessary, and I think you can remove it. `testthat` will already handle loading `{xgboost}`, based on the commands you have in https://github.com/dmlc/xgboost/blob/master/R-package/tests/testthat.R

R-package/NAMESPACE
```diff
@@ -52,6 +53,7 @@ export(xgb.plot.tree)
 export(xgb.save)
 export(xgb.save.raw)
 export(xgb.serialize)
+export(xgb.set.config)
```
jameslamb, line 56: Since you're adding a new dependency, `{jsonlite}`, I expect it to show up in the diff on NAMESPACE.

I think you need to add `importFrom`s to this list: https://github.com/dmlc/xgboost/blob/master/R-package/R/xgboost.R#L78

```r
#' @importFrom jsonlite fromJSON
#' @importFrom toJSON
```

As a side note, I have a personal preference for putting these `importFrom` comments on the code that uses the imports, not grouping them all together in one file. Makes it harder to make mistakes like this, and it means that when code is removed from your package, the imports it carried with it are automatically removed too, without you needing to remember to go update a separate file. But that's bigger than this PR.

jameslamb (commented):
> @jameslamb I'm quite new to the R ecosystem. Does the new addition to the R package look reasonable to you?

Looks pretty good! I left a few suggestions that I think should be addressed, though.


You should feel comfortable adding `{jsonlite}` as a dependency. It doesn't introduce any other required dependencies (https://cran.r-project.org/web/packages/jsonlite/index.html) and is used by a lot of other packages, so it is battled-tested. `{lightgbm}` depends on it for JSON parsing: https://github.com/microsoft/LightGBM/blob/637ff80038e946b3ad7520a49dbdc01ac0ae2c2d/R-package/DESCRIPTION#L51

R-package/NAMESPACE
hcho3: @jameslamb I read that `@importFrom` is not necessary when `::` operator is used to access individual functions: https://kbroman.org/pkg_primer/pages/depends.html

jameslamb: It's something that R people argue about haha.

I personally like to use both. I think of it like this:

* `::` = guarantees the correctness of your code (and makes it slightly faster because R doesn't search for symbols across all attached namespaces)
* `importFrom` = documentation that makes the `NAMESPACE` file readable as a source of truth for *"what are all the packages this package depends on, and which specific objects does it need?"*

R-package/DESCRIPTION
```diff
@@ -64,5 +63,6 @@ Imports:
     methods,
     data.table (>= 1.9.6),
     magrittr (>= 1.5),
+    jsonlite (>= 1.6),
```
jameslamb, line 66: @hcho3 one more thing I forgot to comment on...do you know for sure that the code you've introduced is incompatible with older versions of `{jsonlite}`? Version 1.6 is almost 2 years old so this floor is reasonable, but if you don't have a specific reason for adding this floor I'd change this to `(>= 1.0)`

I think that it's better for libraries to only have floors / ceilings on major versions (like `>=1.0`) unless they know about a specific incompatibility with some minor version. That makes your package work for the largest number of possible users.

R-package/NAMESPACE
hcho3: Got it. Let me add `importForm` for jsonlite functions.

R-package/DESCRIPTION
hcho3: Given that LightGBM is using 1.0+, it should be fine for XGBoost to use it too.

R-package/tests/testthat/test_config.R
hcho3: Removed in my latest commit.

jameslamb (approved):
R changes look good to me!

hcho3:
@RAMitchell Can I get a review from you too?

trivialfis (commented):
Could you please help taking a brief look into using it with dask?  See what's the expected out come.  For example, users will set this value on client process, then what happens on worker process?

hcho3:
@trivialfis 

> users will set this value on client process, then what happens on worker process?

The worker process probably won't receive the new `verbosity` value. There are two options:
* Users should set `verbosity` in the Booster parameter (as before).
* Or users should use `client.run` to call `xgb.set_config` in every worker process.

trivialfis:
Maybe we can get some inspirations from https://docs.dask.org/en/latest/configuration.html#downstream-libraries to merge xgboost's configuration into dask in client process then obtain it on worker process.  WDYT?

trivialfis:
But whichever path we choose, I think it's best we can have a concise description over the expected behavior before merging.

hcho3:
I'm not sure whether all global configuration should always be passed to the workers. For example we certainly do not want to pass `gpu_id` to the workers, when we do add `gpu_id` to global configuration.

If we were to decide on automatic propagation of some global parameters (like `verbosity`), we should clearly document which parameters will be propagated.

trivialfis:
@hcho3 We certainly don't want user to set `gpu_id` on dask interface at all, not even before this PR.   I think in long term we are going to remove those parameters in training and in favor of the new ones defined here.

hcho3:
Interesting, so you are planning to do away with `gpu_id` entirely then? In that case, we can define a precise mechanism for propagating global config from the client to workers.

trivialfis:
@hcho3 I wanted to do that a while ago and vaguely talked about it in https://github.com/dmlc/xgboost/issues/4600 .

trivialfis:
I'm not sure about the "right" design with dask interface.  But hopefully we can make it simple as possible.

hcho3:
@trivialfis I documented the expected behavior of `set_config()` in a Dask cluster.

trivialfis:
I will look deeper on Monday.  ;-)

trivialfis:
Still WIP.  I'm trying to use `dmlc::Parameter` as a base, but it accepts only strings, which makes it a bit difficult to work with.  Will port a routine from our sklearn wrapper to config.

hcho3:
This PR no longer blocks 1.3.0 release. Let's get this feature right, so that we can use it to solve #6297.

doc/parameter.rst
```diff
@@ -16,6 +16,13 @@ Before running XGBoost, we must set three types of parameters: general parameter
   :backlinks: none
   :local:
 
+********************
+Global Configuration
+********************
+The following parameters can be set in the global scope, using ``xgb.config_context()`` (Python) or ``xgb.set.config()`` (R).
+
+* ``verbosity``: Verbosity of printing messages. Valid values of 0 (silent), 1 (warning), 2 (info), and 3 (debug).
```
trivialfis, line 24: Should we deprecate the one used in training?

python-package/xgboost/config.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # Example of using the context manager xgb.config_context().
+        # The context manager will restore the previous value of the global
+        # configuration upon exiting.
+        with xgb.config_context(verbosity=0):
+            # Suppress warning caused by model generated with XGBoost version < 1.0.0
+            bst = xgb.Booster(model_file='./old_model.bin')
+        assert xgb.get_config()['verbosity'] == 2  # old value restored
+    """
+    config_doc.header = header
```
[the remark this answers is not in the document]
hcho3: I kept getting error saying that `header` was an unbound local variable.

python-package/xgboost/config.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+@contextmanager
+@config_doc(header="""
+    Context manager for global XGBoost configuration.
+    """,
+            parameters="""
+    Parameters
+    ----------
+    new_config: Dict[str, str]
```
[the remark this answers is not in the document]
hcho3: No, it should be `[str, Any]`.

python-package/xgboost/dask.py
```diff
@@ -639,14 +639,15 @@ async def _train_async(client,
...
+    _global_config = config.get_config()
 
     def dispatched_train(worker_addr, rabit_args, dtrain_ref, dtrain_idt, evals_ref):
         '''Perform training on a single worker.  A local function prevents pickling.
 
         '''
         LOGGER.info('Training on %s', str(worker_addr))
         worker = distributed.get_worker()
-        with RabitContext(rabit_args):
+        with RabitContext(rabit_args), config.config_context(**_global_config):
```
hcho3, line 650: Does Dask automatically marshal the object `_global_config` to the worker processes? That's pretty cool.

hcho3 (commented):
Thanks for tackling the integration with Dask. It looks good to me.

python-package/xgboost/dask.py
```diff
@@ -639,14 +639,15 @@ async def _train_async(client,
 
     workers = list(_get_workers_from_data(dtrain, evals))
     _rabit_args = await _get_rabit_args(len(workers), client)
+    _global_config = config.get_config()
```
hcho3, line 642: Is it okay that `get_config()` is not async, whereas `_get_rabit_args()` is?

doc/parameter.rst
hcho3: Yes. How soon should we do so?

python-package/xgboost/dask.py
trivialfis: Get rabit args has a client running in it.

python-package/xgboost/dask.py
trivialfis: Cool most of the time.  Not cool when you start worrying about something got pickled. ;-)

doc/parameter.rst
trivialfis: Not sure.  We can discuss it for the next release when stuffs like rmm and gpu_id are ported here.

python-package/xgboost/config.py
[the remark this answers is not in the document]
trivialfis: That's weird.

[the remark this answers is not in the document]
hcho3: You can try removing `config_doc.header = header` and see if you get the same error as I did.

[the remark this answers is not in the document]
trivialfis: @hcho3 Fixed in latest commit.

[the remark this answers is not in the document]
hcho3: Great!

python-package/xgboost/config.py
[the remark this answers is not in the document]
trivialfis: Fixed.

trivialfis:
Added check for unknown parameter.

R-package/R/xgb.config.R
```diff
@@ -0,0 +1,38 @@
+#' Global configuration consists of a collection of parameters that can be applied in the global
+#' scope. See \url{https://xgboost.readthedocs.io/en/latest/parameter.html} for the full list of
```
trivialfis, line 2: Can we use `stable` instead of `latest`?

R-package/R/xgb.config.R
```diff
@@ -0,0 +1,38 @@
...
+#' parameters supported in the global configuration. Use \code{xgb.set.config} to update the
+#' values of one or more global-scope parameters. Use \code{xgb.get.config} to fetch the current
+#' values of all global-scope parameters (listed in
+#' \url{https://xgboost.readthedocs.io/en/latest/parameter.html}).
```
trivialfis, line 6: Ditto.

R-package/R/xgb.config.R
hcho3: Sure. Go ahead.

trivialfis (approved):

PR #6417, opened 2020-11-20
Fix flaky sparse page dmatrix test.

trivialfis:
https://xgboost-ci.net/blue/organizations/jenkins/xgboost-win64/detail/PR-6416/1/


trivialfis:
In a previous PR, I changed the race test to use `std::thread` instead of openmp to enable it on non-omp build, which was to help enabling tests with external memory on non-omp environment.  The PR changes `std::thread` to `std::future`, to avoid the sequential `join`.

hcho3 (approved):

PR #6431, opened 2020-11-23
Update dmlc-core submodule and conform to new API

tpboudreau:
This PR:

1. Updates the branch and commit referenced in the dmlc-core submodule to the latest for that library
2. Makes a small change to XGBoost to allow it to be compiled against the current dmlc-core API

***

Details:

The dmlc-core library made a breaking change to its API, removing an optional output parameter from the public method dmlc::Parameter::UpdateAllowUnknown().

The XGBoost class XGBoostParameter, which is modeled on dmlc::Parameter, also has an UpdateAllowUnknown() method which contains an analogous optional output parameter.  However, XGBoost no longer compiles cleanly against the latest dmlc-core because XGBoostParameter::UpdateAllowUnknown() calls dmlc::Parameter::UpdateAllowUnknown() with the optional argument that it no longer recognizes.

This patch retains the public API for XGBoostParameter::UpdateAllowUnknown(), but always ignores the optional output parameter.

Another option for this update would be to make a breaking change to the public api for XGBoostParameter::UpdateAllowUnknown() and remove (rather than always ignore) the optional parameter.  But this should probably wait for a major release.

The optional parameter (now being ignored) is never used at any call site within the XGBoost codebase nor in any cpp test.  (I'll monitor the CI tests to ensure that no package tests fail.)

***

This patch is in preparation for a planned enhancement to the dmlc-core library (which will necessarily contain this breaking change) for downstream use in XGBoost.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -171,10 +171,15 @@ jobs:
         architecture: 'x64'
     - name: Install Python packages
       run: |
-        python -m pip install wheel setuptools mypy
+        python -m pip install wheel setuptools mypy dask[complete] distributed
     - name: Run mypy
       run: |
-        cd python-package && mypy . || true
+        cd python-package
+        # dask is required to pass, others are not
+        mypy ./xgboost/dask.py ../tests/python/test_with_dask.py --follow-imports=silent
+        mypy ../tests/python-gpu/test_gpu_with_dask.py --follow-imports=silent
+        # If any of the above failed, contributor won't see the next error.
+        mypy . || true
 
   doxygen:
     runs-on: ubuntu-latest
--- a/.gitignore
+++ b/.gitignore
@@ -115,3 +115,13 @@ dask-worker-space/
 
 # Jupyter notebook checkpoints
 .ipynb_checkpoints/
+
+# credentials and key material
+config
+credentials
+credentials.csv
+*.env
+*.pem
+*.pub
+*.rdp
+*_rsa
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,7 @@
 [submodule "dmlc-core"]
 	path = dmlc-core
 	url = https://github.com/dmlc/dmlc-core
+	branch = main
 [submodule "cub"]
 	path = cub
 	url = https://github.com/NVlabs/cub
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -11,6 +11,7 @@ xgb.Booster.handle <- function(params = list(), cachelist = list(),
     if (typeof(modelfile) == "character") {
       ## A filename
       handle <- .Call(XGBoosterCreate_R, cachelist)
+      modelfile <- path.expand(modelfile)
       .Call(XGBoosterLoadModel_R, handle, modelfile[1])
       class(handle) <- "xgb.Booster.handle"
       if (length(params) > 0) {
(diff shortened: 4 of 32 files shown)


hcho3:
@tpboudreau Thanks for your contribution.

> a planned enhancement to the dmlc-core library ... for downstream use in XGBoost.

Would you be able to share what kind of enhancement you are planning?

tpboudreau:
Thanks @hcho3 for the quick feedback.

> Would you be able to share what kind of enhancement you are planning?

Yes, we're exploring the idea of adding functionality to the csv parser in dmlc-core that would allow the user to select a subset of columns present in the input file to be included in each output row.  The initial thought is to add a new URL parameter whose value is a list of column indices to select as included features, something like: `feature_columns=1,3,5,9`.  (This seems the most natural place to specify them -- analogous to the label and weight column params -- but might be unwieldy if the index list gets very long.)  The scope and ergonomics of the feature are still open.

The motivating use case is the situation in which you have a very large csv file, with many columns, but on a particular exercise you require only a small fraction of the columns.  Allowing columns to be subselected on the fly during parsing would then obviate the (current) requirement of externally processing the file to slim it down (and duplicative storing of subsets, etc.)

> XGBoostParameter::UpdateAllowUnknown() is not considered a public API, so we can break it in the next minor release.

In this case, I think I'd suggest removing the optional parameter from the signature right now -- probably less confusing that way.

hcho3:
@tpboudreau Have you considered using weighted feature sampling? This is a new feature that's going to be part of 1.3.0. See example at https://github.com/dmlc/xgboost/blob/master/demo/guide-python/feature_weights.py.

tpboudreau:
The feature weight functionality looks interesting -- thanks for pointing it out.  Some of what's driving this for us is not just the need to logically ignore features (zero weighting), but the hope of reducing memory consumption (and other resource usage) for unused features given the very large files we often use.

That said, I'm not an everyday user of XGBoost, so my intuitions about the impacts of various options are not well developed yet.  I'll keep feature weighting in mind during exploration.

trivialfis:
Personally I suggest using a proper data manipulation library like numpy instead of xgboost (which is a machine learning library) for these kind of tasks.  The scope of DMatrix in xgboost is very limited: being a intermediate storage format for various functionalities in xgboost.

tpboudreau:
@trivialfis, thanks for the input.  I appreciate your desire to keep the DMatrix functionality focused.

I should have mentioned in my earlier sketch of our idea, that we're using XGBoost's external memory feature, which requires (if I understand correctly) the DMatrix to be loaded from a file.  So pre-processing an original csv file with numpy (or other tool) just to copy certain columns to an intermediate csv file that will be used to instantiate a DMatrix is the step we're hoping to avoid.

I could be missing something -- it would be great to learn that there's already a way to do this column selection in one pass of the original file.

trivialfis:
Depends on https://github.com/dmlc/dmlc-core/pull/638 .

hcho3 (approved):

tpboudreau:
Thanks @hcho3 for pushing this (and the dmlc-core memory issue) over the finish line.

hcho3:
@tpboudreau I'm approving this PR, to use the latest dmlc-core. However, I have reservations about adding feature subsamping in the DMatrix. As @trivialfis said, we try not to make DMatrix too bloated.

trivialfis:
Glad to see this merged! Thanks for the effort @tpboudreau @hcho3

PR #6457, opened 2020-12-03
Update the comments

hzy001:
After upgrading XGBoost to the master version, I found that when I set the param num_col to 0 as usual in `XGDMatrixCreateFromFile`, it could not get the correct columns. 
Signed-off-by: Hao Ziyu <haoziyu@qiyi.com>


trivialfis (commented):
I think it's best to declare it never guesses the column.  Right now the only situation that columns need guessing is reading from file like libsvm or csv.  It has been nothing but a headache due to sparsity of data.

trivialfis:
Thanks for submitting the PR.  I will try to follow up.

trivialfis (approved):

PR #6470, opened 2020-12-05
Fix global config default value.

trivialfis:


hcho3:
Please provide some description of why this is necessary.

python-package/xgboost/dask.py
```diff
@@ -771,11 +774,9 @@ async def _predict_async(client, model, data, missing=numpy.nan, **kwargs):
         raise TypeError(_expect([DaskDMatrix, da.Array, dd.DataFrame],
                                 type(data)))
 
-    _global_config = config.get_config()
-
```
hcho3, line 775: What is wrong with this line? Does it receive the config of the worker, not the scheduler? Why did the unit test not catch the problem?

trivialfis:
@hcho3 

The bug is found during debugging a R model compatibility test, where xgboost when silent but the test requires a warning.  The cause is verbosity not being initialized by configuration, which only run after actual functionality like `update_one_iter` or `predict`.   So at the load stage the `verbosity` global parameter is arbitrary value.

After fixing it at C++ level.  The dask test failed.  That's likely because after entering async function, it's running on a different process, which has a different static storage for the global parameter and got reinitialized to 1 on that process after the fix.

As for why wasn't caught before, I don't know.

hcho3:
@trivialfis Thanks for great explanation. The fix now makes sense to me. Approving it now.

hcho3 (approved):

PR #6501, opened 2020-12-14
[doc] [dask] Add example on early stopping with Dask

jameslamb:
I've been testing the new Dask features released in XGBoost 1.3.0 recently. Thanks so much for all the hard work!

I was trying to use early stopping and couldn't figure out what the expectation was for passing validation sets. This PR proposes adding a small section on evaluation + early stopping to the Dask tutorial. I think it might help people who are trying to figure this out, and it brings some attention to this new feature.


trivialfis (approved):
Thanks for improving the doc!  Looks good to me overall.  I fixed a small typo and added a line for obtaining the best model.

Issue #6452, opened 2020-12-01
Load DMatrix from file and feature_names


Roffild:
```python
import pathlib, pandas, xgboost
psave=pathlib.Path(r"xgb_load_test.xgb")
data=pandas.DataFrame({"first": [778, 786], "two": [234, 56], "result": [0, 1]})
d1=xgboost.DMatrix(data.drop(columns="result"), data["result"])
d1.save_binary(psave)
d2=xgboost.DMatrix(psave)
print(d1.feature_names)
print(d2.feature_names)
```
```
['first', 'two']
['f0', 'f1']
```

hcho3:
Currently the binary format for DMatrix does not preserve the feature names. You may want to use alternative formats if you need to preserve feature names.

Roffild:
feature_names are saved in binary format.

hcho3:
> feature_names are saved in binary format

No, it does not. Take a look at https://github.com/dmlc/xgboost/blob/927c316aeb0a948e9635c1ffe54d5e6ed25f6797/src/data/simple_dmatrix.cc#L200

trivialfis:
Looking into it.

trivialfis:
On latest branch this is actually saved, along with feature types.  But after loading the DMatrix back from binary, they got reset from Python constructor.

hcho3:
> On latest branch this is actually saved

Can you point me where this happens? I thought only feature types were being saved.

hcho3:
My bad. Here is the place where feature names are saved:

https://github.com/dmlc/xgboost/blob/512b464cfa109bbd8c4bb88eaf418060656ee8a1/src/data/data.cc#L184-L185

PR #6507, opened 2020-12-15
Fix DMatrix feature names/types IO.

trivialfis:
* Added type hints.

Close https://github.com/dmlc/xgboost/issues/6452 .

### DIFF
--- a/tests/python/test_dmatrix.py
+++ b/tests/python/test_dmatrix.py
@@ -222,6 +222,7 @@ class TestDMatrix:
                 "c": [4, 5]
             })
             m0 = xgb.DMatrix(data.loc[:, ["a", "b"]], data["c"])
+            assert m0.feature_names == ['a', 'b']
             m0.save_binary(path)
             m1 = xgb.DMatrix(path)
             assert m0.feature_names == m1.feature_names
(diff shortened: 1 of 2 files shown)


tests/python/test_dmatrix.py
```diff
@@ -207,6 +211,22 @@ def test_feature_names(self):
...
+    def test_save_binary(self):
+        import pandas as pd
+        with tempfile.TemporaryDirectory() as tmpdir:
+            path = os.path.join(tmpdir, 'm.dmatrix')
+            data = pd.DataFrame({
+                "a": [0, 1],
+                "b": [2, 3],
+                "c": [4, 5]
+            })
+            m0 = xgb.DMatrix(data.loc[:, ["a", "b"]], data["c"])
```
hcho3, line 224:
```suggestion
            m0 = xgb.DMatrix(data.loc[:, ["a", "b"]], data["c"])
            assert m0.feature_names == ['a', 'b']
```
We should also check that the feature names of `m0` are not generic:

hcho3 (changes requested):
Thanks for fixing it. See my comment.

hcho3 (approved):
LGTM, pending tests passing.

Issue #6503, opened 2020-12-15
Categorical data support.


trivialfis:
We started initial experimental support for categorical in xgboost 1.3.  The initial target is to make the one-hot encoding based tree split available for all xgboost components.  Here is a list of to-do items:

## One Hot
- Recognize categorical data in DMatrix.
  + [X] Accept feature types.
  + [X] DMatrix binary.
  + [X] Revert data handler. (#5893)
  + [x] Support for non-dataframe types (https://github.com/dmlc/xgboost/pull/7220)

- Remove min values.
  + [ ] Clarify trivial split (https://github.com/dmlc/xgboost/issues/2914).
  + [ ] Replace trivial split with infinity.

- Device DMatrix.
  + [x] Normal data. (#7042)
  + [x] Iterator. (#7043)

- GPU Sketching for categorical data.
  + [X] Count categories automatically.
  + [x] Weighted sketching. (#6508)
  + [x] Handle merging by removing duplicated entries after merge. (#7043)
  + [X] Ignore num bins in GPU pruning. (#6137)
  + [X] Handle compress index. (#6140)

- GPU Evaluate splits based on feature types.
  + [X] Unify split candidate structs. (https://github.com/dmlc/xgboost/pull/6037)
  + [X] Working implementation.
  + [ ] Handle monotone constraint.

- Tree model (RegTree).
  + [X] Define expand categorical.
  + [x] Tree model dump. (#7036)
  + [x] Feature scores. (#7041)
  + [x] Model as dataframe. (#7065)
  + [x] Split value histogram. (#7065)

- GPU Hist Updater.
  + [X] Evaluate split
  + [X] Apply split

- Prediction.
  + [x] Use predictor in dart (https://github.com/dmlc/xgboost/pull/6693).
  + [x] CPU Predictor. (#7001)
  + [X] GPU Predictor. (#6165)
  + [x] CPU leaf prediction. (#7001)
  + [x] GPU leaf prediction. (#7001)
  + [X] Missing value handling.
  + [x] Shap. (#7053)
  + [x] In-place prediction. (https://github.com/dmlc/xgboost/pull/7322)

- Model IO.
  + [X] Design JSON model schema.
  + [X] Implement it.
  + [X] Update schema.
  + [X] Set JSON to be the default format for pickle serialization.

- CPU Sketching.
  + [X] Merge sparse cut builder and dense cut builder. (#5880)
  + [x] Count categories. (https://github.com/dmlc/xgboost/pull/7221)
[...]

PR #6508, opened 2020-12-15
Support categorical data in GPU weighted sketching.

trivialfis:
Related: https://github.com/dmlc/xgboost/issues/6503 .

### DIFF
--- a/src/common/hist_util.cu
+++ b/src/common/hist_util.cu
@@ -220,10 +220,12 @@ void ProcessBatch(int device, MetaInfo const &info, const SparsePage &page,
 }
 
 void ProcessWeightedBatch(int device, const SparsePage& page,
-                          Span<const float> weights, size_t begin, size_t end,
+                          MetaInfo const& info, size_t begin, size_t end,
                           SketchContainer* sketch_container, int num_cuts_per_feature,
                           size_t num_columns,
                           bool is_ranking, Span<bst_group_t const> d_group_ptr) {
+  auto weights = info.weights_.ConstDeviceSpan();
+
   dh::XGBCachingDeviceAllocator<char> alloc;
   const auto& host_data = page.data.ConstHostVector();
   dh::device_vector<Entry> sorted_entries(host_data.begin() + begin,
@@ -267,9 +269,10 @@ void ProcessWeightedBatch(int device, const SparsePage& page,
                              batch_it, dummy_is_valid,
                              0, sorted_entries.size(),
                              &cuts_ptr, &column_sizes_scan);
-
-  auto const& h_cuts_ptr = cuts_ptr.ConstHostVector();
   auto d_cuts_ptr = cuts_ptr.DeviceSpan();
+  detail::RemoveDuplicatedCategories(device, info, d_cuts_ptr, &sorted_entries,
+                                     &column_sizes_scan);
+  auto const& h_cuts_ptr = cuts_ptr.ConstHostVector();
 
   // Extract cuts
   sketch_container->Push(dh::ToSpan(sorted_entries),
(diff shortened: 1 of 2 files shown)


src/common/hist_util.cu
```diff
@@ -267,9 +269,10 @@ void ProcessWeightedBatch(int device, const SparsePage& page,
                              batch_it, dummy_is_valid,
                              0, sorted_entries.size(),
                              &cuts_ptr, &column_sizes_scan);
-
-  auto const& h_cuts_ptr = cuts_ptr.ConstHostVector();
   auto d_cuts_ptr = cuts_ptr.DeviceSpan();
+  detail::RemoveDuplicatedCategories(device, info, d_cuts_ptr, &sorted_entries,
```
hcho3, line 273: I don't see a unit test for `RemoveDuplicatedCategories`. Can you write on?

trivialfis: It's not exposed in header.  Does `TestCategoricalSketch` suffice?

tests/cpp/common/test_hist_util.cu
```diff
@@ -146,7 +157,8 @@ void TestCategoricalSketch(size_t n, size_t num_categories, int32_t num_bins) {
 }
 
 TEST(HistUtil, DeviceSketchCategoricalFeatures) {
-  TestCategoricalSketch(1000, 256, 32);
+  TestCategoricalSketch(1000, 256, 32, false);
```
hcho3, line 160: How does this test pass, if there are 256 categories and only 32 bins. Then surely `n_uniques` should be 256 and won't be equal to `cuts.TotalBins()`?

src/common/hist_util.cu
hcho3: Maybe. I'd like to understand how `TestCategoricalSketch` is passing. See my comment below.

tests/cpp/common/test_hist_util.cu
trivialfis: Bin values for each feature are the unique values.

trivialfis: There's no sketching if feature type is categorical.

trivialfis: That means `num_bin` is ignored for categorical features.

hcho3: > That means `num_bin` is ignored for categorical features.

Can you point me to the piece of code that detects whether a given feature is categorical? Is it the following snippet?
https://github.com/dmlc/xgboost/blob/b05073bda5a5b0ab6be6281212037e26ed9c0404/src/common/quantile.cu#L626-L635

trivialfis: @hcho3  For the GPU implementation of GK sketching, there are 4 parts:

- hist_utils.cu(h).  This file is the front end of GPU sketch, which extracts sliding windows, remove duplicated categorical values, then push the value into quantile structure.
- Pruning.  This is defined in quantile.cu, the implementation is exactly as described in xgboost paper (more exact than CPU implementation).  For categorical data, the first part has already handle everything, so extra code in pruning for cat data, is to ignore it (don't prune cat features).  Search the use of `IsCat` function.
- Merging.  I haven't don't anything there yet.  It's used by data iter, which is used by `DaskDeviceQuantileDMatrix` and `DeviceQuantileDMatrix`.  Also it's required for distributed setting where we need a global sketch for data.
- Creating the bin values.  That's the code you referenced.

trivialfis: Hope that helps.

hcho3: Thanks for the detailed reply. Searching for `IsCat` function makes it clear how the sketching algo treats categorical features differently.

> I haven't done anything there yet. It's used by data iter, which is used by DaskDeviceQuantileDMatrix and DeviceQuantileDMatrix.

So fair to say that we can't yet use categorical data in DeviceQuantileDMatrix ?

hcho3 (approved):

tests/cpp/common/test_hist_util.cu
trivialfis: @hcho3 DQD is not yet supported.  See items in https://github.com/dmlc/xgboost/issues/6503 .

PR #6510, opened 2020-12-15
[Breaking] Upgrade cuDF and RMM to 0.18 nightlies; require RMM 0.18+ for RMM plugin

hcho3:
Also update the RMM plugin code to account for the updated signature of the template class `rmm::mr::thrust_allocator<T>` (https://github.com/rapidsai/rmm/pull/647).

cc @harrism @jrhemstad

### DIFF
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -429,7 +429,7 @@ struct XGBDefaultDeviceAllocatorImpl : XGBBaseDeviceAllocator<T> {
   }
 #if defined(XGBOOST_USE_RMM) && XGBOOST_USE_RMM == 1
   XGBDefaultDeviceAllocatorImpl()
-    : SuperT(rmm::mr::get_current_device_resource(), cudaStream_t{nullptr}) {}
+    : SuperT(cudaStream_t{nullptr}, rmm::mr::get_current_device_resource()) {}
 #endif  // defined(XGBOOST_USE_RMM) && XGBOOST_USE_RMM == 1
 };
 
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -429,7 +429,7 @@ struct XGBDefaultDeviceAllocatorImpl : XGBBaseDeviceAllocator<T> {
   }
 #if defined(XGBOOST_USE_RMM) && XGBOOST_USE_RMM == 1
   XGBDefaultDeviceAllocatorImpl()
-    : SuperT(cudaStream_t{nullptr}, rmm::mr::get_current_device_resource()) {}
+    : SuperT(rmm::cuda_stream_default, rmm::mr::get_current_device_resource()) {}
 #endif  // defined(XGBOOST_USE_RMM) && XGBOOST_USE_RMM == 1
 };
 
--- a/tests/ci_build/Dockerfile.gpu
+++ b/tests/ci_build/Dockerfile.gpu
@@ -19,7 +19,7 @@ ENV PATH=/opt/python/bin:$PATH
 # Create new Conda environment with cuDF, Dask, and cuPy
 RUN \
     conda create -n gpu_test -c rapidsai-nightly -c rapidsai -c nvidia -c conda-forge -c defaults \
-        python=3.7 cudf=0.17* rmm=0.17* cudatoolkit=$CUDA_VERSION_ARG dask dask-cuda dask-cudf cupy \
+        python=3.7 cudf=0.18* rmm=0.18* cudatoolkit=$CUDA_VERSION_ARG dask dask-cuda dask-cudf cupy \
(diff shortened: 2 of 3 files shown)


src/common/device_helpers.cuh
```diff
@@ -429,7 +429,7 @@ struct XGBDefaultDeviceAllocatorImpl : XGBBaseDeviceAllocator<T> {
   }
 #if defined(XGBOOST_USE_RMM) && XGBOOST_USE_RMM == 1
   XGBDefaultDeviceAllocatorImpl()
-    : SuperT(rmm::mr::get_current_device_resource(), cudaStream_t{nullptr}) {}
+    : SuperT(cudaStream_t{nullptr}, rmm::mr::get_current_device_resource()) {}
```
harrism, line 432: Recommend using `rmm::cuda_stream_default` here.

```suggestion
    : SuperT(rmm::cuda_stream_default, rmm::mr::get_current_device_resource()) {}
```

pseudotensor:
Do things like this preserve backwards compatibility?  i.e. anyone using xgboost with RMM has to use nightly (unstable) rapids 0.18?  Or any rapids will work?

src/common/device_helpers.cuh
hcho3: Is there a semantic difference, or is this a cosmetic change?

harrism: Ah, but this would not be backwards compatible.  To @pseudotensor 's point, I think the change to the parameter order of `rmm::mr::thrust_allocator_adaptor` was a breaking change. If you want backwards compatibility, you will need to detect the version of RMM and ifdef or something...

hcho3: @harrism Should XGBoost aim to support multiple versions of RMM? If so, is there a macro I can use to detect the version of RMM?

hcho3:
@pseudotensor As it stands, no, this PR breaks backward compatibility.

src/common/device_helpers.cuh
harrism: > Is there a semantic difference, or is this a cosmetic change?

It's not cosmetic. In RMM and other RAPIDS C++ APIs we aim for type safety wherever possible. `cuda_stream_view` is a strongly typed wrapper for streams. Without it, you can end up with ambiguity, e.g. `SuperT(0, 0)` or `SuperT(nullptr, nullptr)` is perfectly valid and will compile. Even `SuperT(nullptr, cudaStreamDefault)` will probably compile. We added strongly typed wrappers to RMM to help improve RAPIDS APIs.

> @harrism Should XGBoost aim to support multiple versions of RMM? If so, is there a macro I can use to detect the version of RMM?

That's up to you. No, RMM doesn't currently have a version macro. This is something we could consider if you want to file an issue. Sorry about the breaking change.

pseudotensor:
Seems not good idea to track nightly breaking changes.  xgboost should be more stable, have deprecation stages, etc.  Ya, seems easy to import rmm and check version.  Would be nice if rmm had standard __version__ .  But one can do:

```
>>> from rmm._version import get_versions
>>> get_versions()['version']
'0.14.0'
```

hcho3:
> Seems not good idea to track nightly breaking changes. xgboost should be more stable, have deprecation stages, etc

@pseudotensor 
I agree with you in general. However, the RMM plugin is still marked as experimental and so it is exempt from the usual deprecation policy.

@harrism For now, we will make this particular breaking change. I will file a feature request for detecting the RMM version.

pseudotensor:
> > Seems not good idea to track nightly breaking changes. xgboost should be more stable, have deprecation stages, etc
> 
> @pseudotensor
> I agree with you in general. However, the RMM plugin is still marked as experimental and so it is exempt from the usual deprecation policy.
> 
> @harrism For now, we will make this particular breaking change. I will file a feature request for detecting the RMM version.

Even if experimental in xgboost it doesn't follow xgboost has to track unstable nightly versions of the experimental feature.  That's your choice.

hcho3:
@pseudotensor 

Some considerations:

* RMM plugin is by default turned off from the usual XGBoost installation (`pip install xgboost`). Only the XGBoost Conda package from the `rapidsai` channel will use it.
* There's the issue of not having a way to obtain the RMM version from the C++ layer. Your suggestion works in Python but not in C++.

So in this particular occasion, I vote for the breaking change.

hcho3:
@pseudotensor @harrism I created https://github.com/rapidsai/rmm/issues/663 to request the version string in C++.

trivialfis (approved):
Looks good to me.  I will go with the breaking change.  Requiring nightly dependencies when using nightly xgboost seems fine to me.

trivialfis:
Also please remove the `CI` tag when squashing commits as it's touching the core of xgboost.

hcho3:
Got it. Let me also add a [breaking] prefix, so that later I can remind myself that RMM 0.18+ is required to enable the RMM plugin.

PR #6513, opened 2020-12-16
Drop binary format for memory snapshot.

trivialfis:
* Remove saving binary snapshot.  Loading is preserved for now.


trivialfis:
Note: JSON snapshot was flipped on as default in 1.3.

hcho3 (approved):

PR #6517, opened 2020-12-16
Enable loading model from <1.0.0 trained with objective='binary:logitraw'

### LINKED ISSUE #6509: restoring xgboost 0.82.1 version models (in R) with xgboost 1.1.1.1 or 1.2.01 still failing ?
I am trying to adapt my workflow to later versions of xgboost, however earlier fitted models are causing me grief... I had hoped that the following comments indicated that a new compatability layer would solve this issue, however it seems not ?

>#5794 has been addressed by #5940, and in the upcoming XGBoost version, you can read models from old RDS models again.
>_Originally posted by @hcho3 in https://github.com/dmlc/xgboost/issues/5815#issuecomment-663975193_

I provide a reprex here:
https://github.com/nirguk/xgboost_versions

for example sake, I fit an xgb model on Titanic data with binary:logitraw as objective.
I save this with xgb.save and then
I try to load this in 1.1.1.1 and also in 1.2.0.1  with xgb.load, but receive error:
```
Error in xgb.Booster.handle(modelfile = modelfile) : 
  [22:52:10] amalgamation/../src/objective/./regression_loss.h:89: Check failed: base_score > 0.0f && base_score < 1.0f: base_score must be in (0,1) for logistic loss, got: -0
```

hcho3:
Closes #6509 
* Override `ProbToMargin()` for the `binary:logitraw` objective, since no transformation is necessary.
* Augment model compatibility suite with `binary:logitraw`. I also updated the files `xgboost_model_compatibility_test.zip` and `xgboost_r_model_compatibility_test.zip`.

### DIFF
--- a/R-package/tests/helper_scripts/generate_models.R
+++ b/R-package/tests/helper_scripts/generate_models.R
@@ -2,7 +2,6 @@
 # of saved model files from XGBoost version 0.90 and 1.0.x.
 library(xgboost)
 library(Matrix)
-source('./generate_models_params.R')
 
 set.seed(0)
 metadata <- list(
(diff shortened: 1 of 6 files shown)


trivialfis (approved):
LGTM if all tests are green.

Issue #6496, opened 2020-12-13
Better support for mypy.


trivialfis:
Currently many more Python projects like dask and optuna is using Python type hints. With the Python package of xgboost gaining more and more features, we should also adopt mypy as a safe guard against some type errors and for better code documentation.

trivialfis:
I marked this as good first issue.  Adding type checks for Python is an incremental process, parameter by parameter and function by function.  It would be a good chance to review existing code base and make improvement during the process.  I have made some progress on https://github.com/dmlc/xgboost/pull/6519 with the dask module, which is by far the most dynamic module in xgboost.  Other modules can use it as a base line.

Contributions are welcomed!

jjwang01:
Hi, I was wondering if I could work on this as a first issue. Thanks!

trivialfis:
@jjwang01 Please do. ;-)  ping me if you need any help

jjwang01:
Is it alright if I split this into multiple PRs?

trivialfis:
@jjwang01 As many as you need.  In fact, it's a good idea to have multiple PRs so we can communicate better around details.

jjwang01:
@trivialfis Is there some sort of local build script that I can use to test locally rather than after submitting a commit to PR?

jjwang01:
Hi, just wanted to follow up on this comment.

trivialfis:
Hi, sorry for missing the message.

1. Install mypy in your python envorinment.
2. `cd python-package` in xgboost source directory.
3. Run `mypy .`

trivialfis:
The only module that's lacking the support for mypy is the pyspark interface now.

trivialfis:
Notes:
- Use [TypeGuard](https://docs.python.org/3/library/typing.html#typing.TypeGuard) in Python 3.10 for data dispatching.
- Use [ParamSpec](https://docs.python.org/3/library/typing.html#typing.ParamSpec) in Python 3.10 for decorators.
- Use [collections.abc.Iterator](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterator) to replace `Iterator` (pyspark).

trivialfis:
Will not cover pyspark at 2.0. mypy is not ready for pyspark's deep inheritance style.

michael-gendy-mention-me:
[...]

PR #6519, opened 2020-12-16
[dask] Add type hints.

trivialfis:
Related https://github.com/dmlc/xgboost/issues/6496 .

* Add type hints to dask interface.
* Add a `py.typed` file to python package.  See https://www.python.org/dev/peps/pep-0561/#id18 for details.
* Add various parameters to match the interface with single node.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -176,14 +176,10 @@ jobs:
       run: |
         cd python-package
         # dask is required to pass, others are not
-        printf '\n\n mypy start: --- Following errors until end mark are not required to fix ------\n\n'
-
-        mypy . || true
-
-        printf '\n\n mypy end:  --- Above errors are not required to fix --------------------------\n\n'
-
         mypy ./xgboost/dask.py ../tests/python/test_with_dask.py --follow-imports=silent
         mypy ../tests/python-gpu/test_gpu_with_dask.py --follow-imports=silent
+        # If any of the above failed, contributor won't see the next error.
+        mypy . || true
 
   doxygen:
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -171,10 +171,19 @@ jobs:
         architecture: 'x64'
     - name: Install Python packages
       run: |
-        python -m pip install wheel setuptools mypy
+        python -m pip install wheel setuptools mypy dask[complete] distributed
     - name: Run mypy
       run: |
-        cd python-package && mypy . || true
+        cd python-package
+        # dask is required to pass, others are not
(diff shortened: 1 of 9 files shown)


.github/workflows/main.yml
```diff
@@ -174,7 +174,8 @@ jobs:
         python -m pip install wheel setuptools mypy
     - name: Run mypy
       run: |
-        cd python-package && mypy . || true
+        cd python-package
+        mypy ./xgboost/dask.py ../tests/python/test_with_dask.py --follow-imports=silent
```
hcho3, line 178: Can we still run mypy on other Python files? (We can keep it nonblocking)

trivialfis: Yes.  Let me do that.

trivialfis:
The html doc becomes quite verbose though ...

hcho3:
@trivialfis We can hide the type hints from the doc by using `autodoc_typehints` option in Sphinx: https://www.sphinx-doc.org/en/2.0/usage/extensions/autodoc.html#confval-autodoc_typehints. WDYT?

trivialfis:
Thanks for the reference, let me test it later.

trivialfis:
@hcho3 Do you happen to know how to use smaller font size for API reference?  I compared the doc for xgboost and other projects and the font size in xgboost's doc is significantly larger.  If not then I will continue googling ...

hcho3:
@trivialfis This is due to the choice of the theme. I'm not sure if we should necessarily follow other projects when it comes to the font size. I think our doc is similar to sklearn when it comes to the font size.

hcho3:
@trivialfis Why don't we consider splitting our API doc into multiple pages? Right now, it's a single page with lots of functions and modules, and it might be nice to break it up.

trivialfis:
As can be seen from the screen shot, information like `Optional` and `Union` is omitted in the doc strings, which is less accurate than the checked type hints.  It would be difficult for us to get everything right without proper tools like mypy.

hcho3:
@trivialfis https://github.com/agronholm/sphinx-autodoc-typehints seems promising. It should make the type hints appear in the docstring.

trivialfis:
Let me try it.

trivialfis:
Debugged it a little bit didn't get it to work.  It has some weird issue with import `No module named 'pandas.util'; 'pandas' is not a package`.  The error happens inside dask and happens only when building doc.  Also weirdly if I try to print out the backtrace inside dask, the doc build passed but non of the type hints is used.

hcho3:
@trivialfis I found a way to move the type hints to the description, as follows:
Screen Shot 2020-12-28 at 8 30 53 PM

I found the trick from Optuna (optuna/optuna#2147), which actually moved the type hints to the description area 5 days ago.

hcho3 (approved):
LGTM, as long as tests pass.

PR #6531, opened 2020-12-19
Expand `~` into the home directory on Linux and MacOS

hcho3:
According to https://discuss.xgboost.ai/t/xgb-save-setting-a-file-path/2020/7?u=hcho3, XGBoost could not handle path containing `~`, the shorthand for the home directory on Linux and MacOS platforms. This PR enables XGBoost to handle such paths, by expanding `~` in paths into the full path of the home directory.

Example:
```R
xgb.save(bst, '~/Desktop/xgboost/test.model')
```

The fix uses `os.path.expanduser` (Python) and `path.expand` (R).

### DIFF
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -11,6 +11,7 @@ xgb.Booster.handle <- function(params = list(), cachelist = list(),
     if (typeof(modelfile) == "character") {
       ## A filename
       handle <- .Call(XGBoosterCreate_R, cachelist)
+      modelfile <- path.expand(modelfile)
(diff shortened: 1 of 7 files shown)


trivialfis (approved):
Looked into the Python implementation, seems quite easy to be ported to c, which may be beneficial to other lang bindings.  Approving for now, but in the future I might choose a C++ implementation.

hcho3:
@trivialfis Thanks for the review. Should this be part of the patch release too?

trivialfis:
I don't think that's necessary, let's keep the patch release for regression fixes.

Issue #5349, opened 2020-02-23
Eval displays 'nan' with 'reg:gamma' objective and 'gblinear' booster


gorkemozkaya:
I noticed the tree evals keep displaying 'nan's with  'reg:gamma' objective and 'gblinear' booster. The default eval metric for "reg:gamma" objective is "gamma-nloglik". Below is the notebook that can reproduce the issue: 
https://github.com/gorkemozkaya/ds_debugging/blob/master/xgboost_gamma_regression.ipynb

honzasterba:
I have also found similar behaviour behaviour. Here is the script used to reproduce.

https://gist.github.com/honzasterba/79bfcb95e6c183479b45218a40b8af3b

 Is this a bug?

neemah2o:
@hcho3 , can you verify if this is a bug?

PR #6537, opened 2020-12-21
Clip small positive values in gamma-nloglik

gorkemozkaya:
For the `gamma-nloglik` eval metric, small positive values in the labels are causing `NaN`'s in the outputs, as reported here: https://github.com/dmlc/xgboost/issues/5349. This will add clipping on them, similar to what is done in other metrics like `poisson-nloglik` and `logloss`.


trivialfis (approved):
Thanks for the PR!

PR #6540, opened 2020-12-21
"featue_map" typo changed to  "feature_map"

mburaksayici:
"featue_map" typo changed to  "feature_map"


trivialfis (approved):
Thanks for the catch!

PR #6550, opened 2020-12-24
Multiclass prediction caching

ShvetsKS:


ShvetsKS:
Not related to current changes failed step Win64:Test: TypeError: predict_proba() got an unexpected keyword argument 'X'
https://xgboost-ci.net/blue/organizations/jenkins/xgboost-win64/detail/PR-6550/5/pipeline#step-89-log-1315

as the same problem on cb207a3 commit
https://xgboost-ci.net/blue/organizations/jenkins/xgboost-win64/detail/master/660/pipeline

trivialfis:
Could you please share your bench result?

src/tree/updater_quantile_hist.cc
```diff
@@ -620,7 +636,7 @@ void QuantileHistMaker::Builder<GradientSumT>::Update(
 template<typename GradientSumT>
 bool QuantileHistMaker::Builder<GradientSumT>::UpdatePredictionCache(
     const DMatrix* data,
-    HostDeviceVector<bst_float>* p_out_preds) {
+    HostDeviceVector<bst_float>* p_out_preds, const int gid, const int ngroup) {
```
SmirnovEgorRu, line 639: Can you get ```gid``` and ```ngroup``` from the Builder rather than pass them as arguments?
You have ```p_last_tree_ pointer``` . Probably, it's enough already or need to just add a few fields into builder.

In this case:
1) No needs to have ```UpdatePredictionCacheMulticlass``` function at all.
2) No changes in arguments of` ```UpdatePredictionCache```
3) No changes in basic classes like: ```GBTree``` and ```TreeUpdater```

ShvetsKS:
> Could you please share your bench result?

@trivialfis  surre.
Msrank benchmark (full training time):
73s vs 66s.

After introduced prediction optimizations, PredictRaw  kernel step is not big hotspot in multiclass classification case but we can see 10% perf increasing for whole training stage.

src/tree/updater_quantile_hist.cc
ShvetsKS: We have to provide `gid` at each `UpdatePredictionCache` call. Currently there is no possibility to get group id `gid` in specific Builder and even in `QuantileHistMaker`. It requires extensions for `QuantileHistMaker` with methods like `setGroupIndex()`  for example. But as `UpdatePredictionCache` call is polymorphic we can't avoid changes in basic  `TreeUpdater` class.

include/xgboost/tree_updater.h
```diff
@@ -78,6 +78,17 @@ class TreeUpdater : public Configurable {
     return false;
   }
 
+  virtual bool UpdatePredictionCacheMulticlass(const DMatrix* data,
+                                               HostDeviceVector<bst_float>* out_preds,
+                                               const int gid, const int ngroup) {
```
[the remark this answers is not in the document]
ShvetsKS: done

SmirnovEgorRu (approved):

hcho3 (changes requested):
We may want to test this with `num_parallel_tree` > 1.

hcho3 (approved):
Never mind my previous comment. I see the check `num_new_trees == 1` inside `DoBoost()`. LGTM.

trivialfis (commented):
As usual, please add a test. ;-)

ShvetsKS:
> As usual, please add a test. ;-)

@trivialfis  test was added :)

ShvetsKS:
@trivialfis  seems failed steps are not related to current changes

hcho3 (approved):

trivialfis (approved):

trivialfis:
Thanks for the optimization!

PR #6559, opened 2020-12-30
add credentials to .gitignore

jameslamb:
This PR proposes adding some rules to `.gitignore` to prevent sensitive information like passwords and access keys from being checked into source control.

I've added common extensions for key material, plus the specific files that the AWS CLI docs encourage people to create: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html.


hcho3 (approved):
Thanks!

Issue #6553, opened 2020-12-25
`ntree_limit` not used in dask scikit-learn prediction.


pseudotensor:
Started from discussion here: https://github.com/dmlc/xgboost/issues/6547

This is just start of concern that dask and non-dask do not agree.  I don't know even if they agreed that the use of the internal xgb tree structure will give same results.  So far they do not.

```
from dask.distributed import Client
from dask_cuda import LocalCUDACluster
from dask import dataframe as dd
import xgboost as xgb
import numpy as np
import random


def main(client):

    # data
    import pickle
    params, X, y = pickle.load(open("bad_dask.pkl", "rb"))
    for k in ["accuracy", "debug_verbose", "disable_gpus", "early_stopping_rounds", "early_stopping_threshold",
              "ensemble_level", "interpretability", "lossguide", "model_class_name",
              "monotonicity_constraints", "score_f_name", "silent", "time_tolerance",
              "train_shape", "valid_shape"]:
        del params[k]
    print(params)
    print(X.shape)
    target = "__TARGET__"
    X[target] = y

    # dask
    np.random.seed(1234)
    random.seed(1234)
    dask_df = dd.from_pandas(X, chunksize=5000).persist()
    X = X.drop(target, axis=1)
    yd = dask_df[target]
    Xd = dask_df.drop(target, axis=1)

    dtrain = xgb.dask.DaskDMatrix(client, Xd, yd)
    output = xgb.dask.train(client,
                            params,
                            dtrain,
                            num_boost_round=params['n_estimators'])
    preds = xgb.dask.predict(client, output, dtrain)
    print(preds.compute()[0:10])

    # non-dask
    np.random.seed(1234)
    random.seed(1234)
    dtrain = xgb.DMatrix(X, y)
    output = xgb.train(params,
                            dtrain,
                            num_boost_round=params['n_estimators'])
    preds = output.predict(dtrain)
    print(preds[0:10])

    model = xgb.XGBRegressor(**params)
    model.fit(X, y)
    preds = model.predict(X)
    print(preds[0:10])


if __name__ == '__main__':
    with LocalCUDACluster() as cluster:
        with Client(cluster) as client:
            main(client)
```

gives

```
/home/jon/minicondadai/bin/python /home/jon/h2oai.fullcondatest/bad_dask.py
[...]

Issue #6478, opened 2020-12-09
Add SHAP support in Dask API


hcho3:
Currently, the `predict()` method in the Dask API (`xgboost.dask`) only supports the default prediction mode:

https://github.com/dmlc/xgboost/blob/0ffaf0f5be4c486a1802e56812cc2e0d75042c69/python-package/xgboost/dask.py#L886

It would be nice to also support SHAP prediction with Dask arrays.

hcho3:
Closing, as #6117 already exposes `pred_contribs=True` option to the Dask API.

PR #6567, opened 2021-01-02
[WIP] [dask] Use inplace prediction by default.

trivialfis:
* Use inplace predict for dask skl interface.
* Fix https://github.com/dmlc/xgboost/issues/6553 by using `best_iteration`.  The issue was caused by ignoring `ntree_limit`, but this argument is confusing in the context of forest and multi-class, so I decided to avoid it for any new code.

To-do:
- [ ] Add tests for early stopping.

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -1543,16 +1543,13 @@ class DaskXGBClassifier(DaskScikitLearnBase, XGBClassifierBase):
         output_margin: bool,
         base_margin: Optional[_DaskCollection]
     ) -> _DaskCollection:
-        test_dmatrix = await DaskDMatrix(
-            client=self.client, data=X, base_margin=base_margin,
-            missing=self.missing
+        predt = await super()._predict_async(
+            X=X,
+            output_margin=output_margin,
+            base_margin=base_margin,
+            validate_features=validate_features
         )
-        pred_probs = await predict(client=self.client,
-                                   model=self.get_booster(),
-                                   data=test_dmatrix,
-                                   validate_features=validate_features,
-                                   output_margin=output_margin)
-        return _cls_predict_proba(self.objective, pred_probs, da.vstack)
+        return _cls_predict_proba(self.objective, predt, da.vstack)
 
     # pylint: disable=missing-docstring
     def predict_proba(
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -1258,25 +1258,40 @@ async def _evaluation_matrices(
 
 
 class DaskScikitLearnBase(XGBModel):
-    '''Base class for implementing scikit-learn interface with Dask'''
+    """Base class for implementing scikit-learn interface with Dask"""
 
     _client = None
 
     @_deprecate_positional_args
     async def _predict_async(
-        self, data: _DaskCollection,
+        self,
+        data: _DaskCollection,
         output_margin: bool = False,
         validate_features: bool = True,
-        base_margin: Optional[_DaskCollection] = None
+        base_margin: Optional[_DaskCollection] = None,
     ) -> Any:
-        test_dmatrix = await DaskDMatrix(
-            client=self.client, data=data, base_margin=base_margin,
-            missing=self.missing
-        )
-        pred_probs = await predict(client=self.client,
-                                   model=self.get_booster(), data=test_dmatrix,
-                                   output_margin=output_margin,
-                                   validate_features=validate_features)
+        if base_margin:
+            test_dmatrix = await DaskDMatrix(
+                client=self.client,
+                data=data,
+                base_margin=base_margin,
+                missing=self.missing,
+            )
+            pred_probs = await predict(
+                client=self.client,
+                model=self.get_booster(),
+                data=test_dmatrix,
+                output_margin=output_margin,
+                validate_features=validate_features,
+            )
+        else:
+            pred_probs = await inplace_predict(
+                client=self.client,
+                model=self.get_booster(),
+                data=data,
+                predict_type="margin" if output_margin else "value",
+                missing=self.missing,
+            )
         return pred_probs
 
     def predict(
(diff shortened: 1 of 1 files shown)


hcho3:
Does it mean that pred_contrib is not available on Dask?

pseudotensor:
@trivialfis FYI ntree_limit is a critical aspect of xgboost and using early stopping.  The tests in xgboost make it clear how ntree_limit is connected the value of best_iterations.

pseudotensor:
> Does it mean that pred_contrib is not available on Dask?
> 
> See #6478

FYI the scikit-learn API for dask does not support pred_contribs=True, just output_margin and base_margin.

trivialfis:
In my opinion we can just forget the existence the `ntree_limit`.  Could you please take a look into model slicing?  Let us know if there's any use case for `ntree_limit` that can not be replaced by model slicing.

pseudotensor:
> In my opinion we can just forget the existence the `ntree_limit`. Could you please take a look into model slicing? Let us know if there's any use case for `ntree_limit` that can not be replaced by model slicing.

This?  https://xgboost.readthedocs.io/en/latest/python/model.html

This is even further far away from any sklearn API and has all the same complications w.r.t. tree counts as ntree_limit.

pseudotensor:
https://github.com/dmlc/xgboost/issues/6568 FYI pred_contribs doesn't work with non-sklearn dask API as well.

pseudotensor:
@trivialfis if the proposal is to (by default) always have predict use the best_iteration result, so user gets the optimal early stopping result by default, that is a good proposal.  If normal xgboost also did that, this would be a good change because the best iteration result is the normal thing wanted when doing early stopping, not the total number of trees one ends up with.  If in that case one had to use model slicing in order to get some arbitrary other non-best tree prediction, then that is an ok proposal.  But it has to be made uniform between dask and non-dask.

Also, then one should avoid needing to have to pass ntree_limit=0 for GLM, which is a pretty silly requirement.

trivialfis:
> This is even further far away from any sklearn API and has all the same complications w.r.t. tree counts as ntree_limit.

Really?  To me `ntree_limit` to me is much more complicated.  Do you calculate this value by hand when using cls with forest?

> the proposal is to (by default) always have predict use the best_iteration result, so user gets the optimal early stopping result by default, that is a good proposal. 

So far that's the default for single node scikit interface.  Just making the alignment here.

> But it has to be made uniform between dask and non-dask.

Of course.  We are making progress on the unification.  The 2 interfaces are now much more closer than a few months before.

> Also, then one should avoid needing to have to pass ntree_limit=0 for GLM, which is a pretty silly requirement.

Agree.

pseudotensor:
> > This is even further far away from any sklearn API and has all the same complications w.r.t. tree counts as ntree_limit.
> 
> Really? To me `ntree_limit` to me is much more complicated. Do you calculate this value by hand when using cls with forest?

I mean, in terms of parallel trees and classes, the model slicing is as (or more) complicated.  You still have to do the calculation to get the correct tree count

Any additional lines of code required is point of violation of scikit-learn API.  One should be able to just 1) construct 2) fit 3) predict, not have to slice etc.  It at most should be an argument passed to fit or predict.  Else the whole purpose of scikit-learn API is lost.  There's a reason scikit-learn API is popular, it's simple and clean/understandable without arbitrary extra steps.  This is also why CUML etc. try to conform.

As for how to calculate, the early stopping model returns best_ntree_limit, so there is nothing to compute.  But this is the same as (best_iteration + 1 ) * num_parallel_tree if one needed to know.  Multi-class or not does not enter the equation ever.  See the tests, e.g. like in test_training_continuation you wrote :)

> 
> > the proposal is to (by default) always have predict use the best_iteration result, so user gets the optimal early stopping result by default, that is a good proposal.
> 
> So far that's the default for single node scikit interface. Just making the alignment here.

I see yes, I was unsure about that.  I see that is true for the non-dask API, but I don't see for the dask API.

> 
> > But it has to be made uniform between dask and non-dask.
> 
> Of course. We are making progress on the unification. The 2 interfaces are now much more closer than a few months before.
> 
> > Also, then one should avoid needing to have to pass ntree_limit=0 for GLM, which is a pretty silly requirement.
> 
> Agree.

trivialfis:
For multi class, the ntree limit is:

(best_iteration + 1 ) * num_parallel_tree * num_class

Model slicing is one can use `best_iteration` for whatever model.  No extra calculation is required.so you can forget about the above equation.

I think I can see the confusion.  My point is not about model slicing, rather it's about using `best_iteration` instead of `best_ntree_limit`.   Currently we have the `ntree_limit` in skl interface predict function, the end result is it will be replaced by `niter_limit`(maybe a better name but hopefully I can get it clear for this discussion).  Model slicing will be the underlying mechanism for implementing it.  I mentioned it so that you can use it for now before we make the change.

trivialfis:
@hcho3 

> Does it mean that pred_contrib is not available on Dask?

For skl dask, no.  I don't think right now skl interface has support for it.  Single node or distributed.

pseudotensor:
> For multi class (not binary), the ntree limit is:
> 
> (best_iteration + 1 ) * num_parallel_tree * num_class
> 
> Model slicing is one can use `best_iteration` for whatever model. No extra calculation is required.so you can forget about the above equation.
> 
> I think I can see the confusion. My point is not about model slicing, rather it's about using `best_iteration` instead of `best_ntree_limit`. Currently we have the `ntree_limit` in skl interface predict function, the end result is it will be replaced by `niter_limit`(maybe a better name but hopefully I can get it clear for this discussion). Model slicing will be the underlying mechanism for implementing it. I mentioned it so that you can use it for now before we make the change.

https://github.com/dmlc/xgboost/blob/24b87fc18a04d4aade26e058c6d21066978e98b5/tests/python/test_training_continuation.py#L123-L137

Is this testing code incorrect?

trivialfis:
They are wrong.... Tracking it down the bug dates back to 2016 or maybe even earlier.  `ntree_limit` is incorrectly set.

trivialfis:
Closing in favor of https://github.com/dmlc/xgboost/pull/6638 .

PR #6576, opened 2021-01-06
[dask] Add DaskXGBRanker

trivialfis:
Initial support for distributed LTR using dask.

* Support `qid` in libxgboost.
* Refactor `predict` and `n_features_in_`, `best_[score/iteration/ntree_limit]` to avoid duplicated code.
* Define `DaskXGBRanker`.

The dask ranker doesn't support group structure, instead it uses query id and convert to group ptr internally.


trivialfis:
Question to myself:   Right now the qid is required to be sorted for input.  Maybe we need to perform the sorting ourselves?

tests/python/testing.py
```diff
@@ -209,6 +211,29 @@ def get_sparse():
     return X, y
 
 
+@memory.cache
+def get_mq2008(dpath):
+    from sklearn.datasets import load_svmlight_files
```
RAMitchell, line 216: What is the runtime of these new tests? Some time ago, I noticed a large amount of time is consumed by ranking tests.

RAMitchell (approved):

tests/python/testing.py
trivialfis: [2021-01-07T13:18:15.483Z] 2.04s teardown python/test_with_dask.py::test_dask_ranking

https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-6576/8/pipeline/206

trivialfis:
I need to disable the support for group weight for now.

python-package/xgboost/dask.py
```diff
@@ -1287,21 +1288,13 @@ def predict(
         validate_features: bool = True,
         base_margin: Optional[_DaskCollection] = None
     ) -> Any:
-        '''Predict with `data`.
```
hcho3, line 1290: Why are we removing the docstring for `predict()`?

python-package/xgboost/dask.py
```diff
@@ -1223,61 +1264,21 @@ class DaskScikitLearnBase(XGBModel):
...
-        sample_weight: Optional[_DaskCollection] = None,
-        base_margin: Optional[_DaskCollection] = None,
-        eval_set: List[Tuple[_DaskCollection, _DaskCollection]] = None,
-        eval_metric: Optional[Callable] = None,
-        sample_weight_eval_set: Optional[List[_DaskCollection]] = None,
-        early_stopping_rounds: Optional[int] = None,
-        verbose: bool = True,
-        feature_weights: Optional[_DaskCollection] = None,
-        callbacks: List[TrainingCallback] = None
-    ) -> "DaskScikitLearnBase":
```
hcho3, line 1240: Why are we removing the docstring for `fit()`?

python-package/xgboost/dask.py
```diff
@@ -1522,14 +1486,16 @@ def fit(
...
+        early_stopping_rounds: Optional[int] = None,
         verbose: bool = True,
-        feature_weights: _DaskCollection = None,
+        xgb_model: Optional[Booster] = None,
+        sample_weight_eval_set: Optional[List[_DaskCollection]] = None,
+        feature_weights: Optional[_DaskCollection] = None,
         callbacks: Optional[List[TrainingCallback]] = None
     ) -> "DaskXGBClassifier":
         _assert_dask_support()
+        assert xgb_model is None, '`xgb_model` is not supported, load the model instead.'
```
hcho3, line 1498: Why is `xgb_model` not supported for DaskXGBRanker?

python-package/xgboost/sklearn.py
```diff
@@ -271,26 +308,32 @@ def _wrap_evaluation_matrices(self, X, y, group,
...
+                    eval_group[i] is group and
+                    eval_qid[i] is qid
+                ):
                     evals.append(train_dmatrix)
                 else:
                     m = DMatrix(valid_X,
                                 label=label_transform(valid_y),
                                 missing=self.missing, weight=sample_weight_eval_set[i],
                                 nthread=self.n_jobs)
-                    m.set_info(group=eval_group[i])
```
hcho3, line 293: Nice catch!

python-package/xgboost/dask.py
trivialfis: The doc string comes from inheritance as now we support all parameters.

python-package/xgboost/dask.py
trivialfis: The doc string comes from inheritance as now we support all parameters.

python-package/xgboost/dask.py
trivialfis: Written before making the decision of flushing model.  Will correct that.

trivialfis: Done.

hcho3 (approved):

Issue #6547, opened 2020-12-22
[FEATURE] convert dask -> non-dask model


pseudotensor:
https://github.com/rapidsai/cuml/issues/3140#issuecomment-729934809

Related to this general issue, I think xgboost team does a much better job and ensures dask can be pickled etc.

But I encountered some differences between non-dask and dask trees that makes reading them incompatible.  Specifically we have poor accuracy when trying to read the tree structure to do predictions, but only for dask models.

I know xgboost can be used with treelite etc.  Is it expected that dask models can be used too?

In general, if there is any complication, is there a way to convert a dask model (sckit and raw API) into non-dask form so non-dask tools can be used?

@trivialfis

hcho3:
@pseudotensor I don't see any technical hurdle in converting Dask model to an ordinary XGBoost model JSON file. We already have a callback function to serialize model at every boosting iteration:
https://github.com/dmlc/xgboost/blob/2231940d1d08f5aca23cf53d922642996a098d44/tests/python/test_with_dask.py#L1017-L1033

trivialfis:
Could you please share the script that prediction is different?

trivialfis:
Prediction difference is serious bug, so far I haven't been able to reproduce it.  But if you have a MRE I will not hesitate to fix it and push another patch release.

pseudotensor:
Yes, @trivialfis , I'm not saying exactly that the predictions are off, just the way we access the tree structure via the internal format leads to very different predictions between dask and non-dask.  I'll get a repro of some kind ASAP.

trivialfis:
To answer the original question, the booster returned by dask train function is exactly the same with single node.  But if you want to convert the pickled model between skl interfaces of dask and single node, I don't think that's possible at this point.  They are different Python classes and pickle stores Python bytecode.

pseudotensor:
I think it was deduced that the primary problem here is that the dask model does not use ntree_limit so predictions can be off if trying to actually select the best_iterations model.  If that is solved, this issue is no longer needed, so I'll close.  Thanks!

pseudotensor:
Actually, re-opening.  dask is heavily feature incomplete, esp. for scikit-learn API.  But most critically dask cannot do pred_contribs etc.  So would be good to be able to convert to non-dask so can apply normal operations during prediction time.

@hcho3 , thanks, I'll try your suggestion.

pseudotensor:
[...]

PR #6582, opened 2021-01-07
Support `_estimator_type`.

trivialfis:
For more info, see: https://scikit-learn.org/stable/developers/develop.html#estimator-types

* Model trained from dask can be loaded by single node skl interface.  Close https://github.com/dmlc/xgboost/issues/6547 .

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -16,7 +16,7 @@ from .compat import (SKLEARN_INSTALLED, XGBModelBase,
                      XGBClassifierBase, XGBRegressorBase, XGBoostLabelEncoder)
 
 
-class XGBRankerMixIn:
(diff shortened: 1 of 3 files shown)


python-package/xgboost/sklearn.py
```diff
@@ -442,7 +456,7 @@ def save_model(self, fname: str):
                 meta[k] = v
             except TypeError:
                 warnings.warn(str(k) + ' is not saved in Scikit-Learn meta.')
-        meta['type'] = type(self).__name__
```
hcho3, line 445: This is potentially a breaking change. What would happen if the user tried to load an old model from JSON?

trivialfis: If they are using xgboost, then it's fine.

RAMitchell (approved):

PR #6589, opened 2021-01-09
Suppress hypothesis health check for dask client.

trivialfis:
The client fixture is not reset for each hypothesis test case and it's fine for xgboost.


trivialfis:
This is extracted from https://github.com/dmlc/xgboost/pull/6586 .  Will continue investigating in there.

trivialfis:
Merging to unblock CI.

PR #6590, opened 2021-01-10
Remove instance access to SparsePage.

trivialfis:
This is to prevent race access of `host_device_vector.HostVector()`.  We have fixed a few bugs caused by the `[]` operator in `SparsePage`, but I encountered a new one in `SparsePage.GetTranspose()` today.  This PR is for nuking the `operator[]` to prevent any related bug in the future.

From now on, accessing instances in `SparsePage` must go through a `SparsePageView`, which contains span to host data.  No change is made to GPU code since the `operator[]` is not available on device code.

### DIFF
--- a/include/xgboost/data.h
+++ b/include/xgboost/data.h
@@ -252,15 +252,6 @@ class SparsePage {
   /*! \brief an instance of sparse vector in the batch */
   using Inst = common::Span<Entry const>;
 
-  /*! \brief get i-th row from the batch */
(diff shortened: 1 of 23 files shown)


RAMitchell (approved):
This is an improvement although it doesn't completely solve the issue as `auto page = batch.GetView();` could be caled inside an omp loop. 

I don't have any better ideas.

hcho3 (approved):

Issue #6583, opened 2021-01-07
No option to set base_margin for eval_set in sklearn fit method


MatticeV:
I'm using the Python Scikit-Learn wrapper for XGBoost, for my model I want to set base_margin to log(weights). This is possible for the training data, but eval_set accepts the validation data which is converted to a DMatrix object without setting a base_margin (line 283 in sklearn.py). This means that weights are included in the evaluation of the model and it isn't done on a like-for-like basis.

Is it possible to add a eval_set_base_margin option to the fit method?

Issue #6300, opened 2020-10-28
Flaky test: test_with_dask.py::test_boost_from_prediction[approx]


hcho3:
https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-6294/1/pipeline

#6299 will mark the test as XFAIL.

PR #6591, opened 2021-01-11
Add base_margin for evaluation dataset.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/6583 .
Close https://github.com/dmlc/xgboost/issues/6300 .

* Add `base_margin` to evaluation dataset in skl.
* Add all available meta info `DMatrix` constructor, and deprecate the pos arguments.
* Apply similar changes to dask.
* This PR also unifies the evaluation metric configuration between dask and non-dask code paths.  This adds additional validation and saves computation when training data is used for evaluation.

I will split up this mono PR after getting all the pieces together.  With growing size of meta info, I need to do some refactoring.

Small parts are extracted to:
- https://github.com/dmlc/xgboost/pull/6601
- https://github.com/dmlc/xgboost/pull/6627

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -1437,6 +1437,12 @@ class XGBRanker(XGBModel, XGBRankerMixIn):
             A list of (X, y) tuple pairs to use as validation sets, for which
             metrics will be computed.
             Validation metrics will help us track the performance of the model.
+        eval_group : list of arrays, optional
+            A list in which ``eval_group[i]`` is the list containing the sizes of all
+            query groups in the ``i``-th pair in **eval_set**.
+        eval_qid : list of array_like, optional
+            A list in which ``eval_qid[i]`` is the array containing query ID of ``i``-th
+            pair in **eval_set**.
         sample_weight_eval_set : list, optional
             A list of the form [L_1, L_2, ..., L_n], where each L_i is a list of
             group weights on the i-th validation set.
@@ -1469,12 +1475,6 @@ class XGBRanker(XGBModel, XGBRankerMixIn):
         xgb_model :
             file name of stored XGBoost model or 'Booster' instance XGBoost model to be
             loaded before training (allows training continuation).
-        eval_group : list of arrays, optional
-            A list in which ``eval_group[i]`` is the list containing the sizes of all
-            query groups in the ``i``-th pair in **eval_set**.
-        eval_qid : list of array_like, optional
-            A list in which ``eval_qid[i]`` is the array containing query ID of ``i``-th
-            pair in **eval_set**.
         base_margin_eval_set : list, optional
             A list of the form [M_1, M_2, ..., M_n], where each M_i is an array like
             object storing base margin for the i-th validation set.
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -1393,11 +1393,11 @@ class XGBRanker(XGBModel, XGBRankerMixIn):
         eval_set=None,
         eval_group=None,
         eval_qid=None,
-        sample_weight_eval_set=None,
         eval_metric=None,
         early_stopping_rounds=None,
         verbose=False,
         xgb_model: Optional[Union[Booster, str, XGBModel]] = None,
+        sample_weight_eval_set=None,
         base_margin_eval_set=None,
         feature_weights=None,
         callbacks=None
@@ -1443,16 +1443,6 @@ class XGBRanker(XGBModel, XGBRankerMixIn):
         eval_qid : list of array_like, optional
             A list in which ``eval_qid[i]`` is the array containing query ID of ``i``-th
             pair in **eval_set**.
-        sample_weight_eval_set : list, optional
-            A list of the form [L_1, L_2, ..., L_n], where each L_i is a list of
-            group weights on the i-th validation set.
-
-            .. note:: Weights are per-group for ranking tasks
-
-                In ranking task, one weight is assigned to each query group (not each
-                data point). This is because we only care about the relative ordering of
-                data points within each group, so it doesn't make sense to assign
-                weights to individual data points.
         eval_metric : str, list of str, optional
             If a str, should be a built-in evaluation metric to use. See
             doc/parameter.rst.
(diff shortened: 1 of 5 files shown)


trivialfis:
Note to myself: Compile a list of entry points of meta info.

trivialfis:
@hcho3 CI seems to be running into trouble. ;-(

hcho3:
@trivialfis Can you fix this error?
```
TypeError: Optional[t] requires a single type. Got (<class 'xgboost.core.Booster'>, 'XGBModel')
```

trivialfis:
Yup, I will be working on this today.

python-package/xgboost/sklearn.py
```diff
@@ -556,7 +592,7 @@ def load_model(self, fname):
 
     def _configure_fit(
         self,
-        booster: Optional[Booster],
+        booster: Optional[Booster, "XGBModel"],
```
hcho3, line 595: This line appears to cause the trouble. What is this for? Can we revert it?

trivialfis: This is for the `xgb_model` parameter in `fit` method.  It can be both a `Booster` object and a sklearn wrapper, which should inherit from `XGBModel`.  I will fix that.

trivialfis: Sorry for asking you to look into that,  I meant the jenkins CI, it seems you have restarted it.

hcho3: In that case, you should use `Optional[Union[Booster, "XGBModel"]]`

hcho3: @trivialfis According to https://stackoverflow.com/questions/43022843/nvidia-nvml-driver-library-version-mismatch, rebooting fixes this error:
```
Failed to initialize NVML: Driver/library version mismatch
```
Apparently, the EC2 instances performed an unattended (automatic) upgrades. All I did was to kill all worker instances and restart the failed job. Now the problem is fixed.

trivialfis: @hcho3 I think maybe, some workers are still lingering around?  https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-6591/14/pipeline

hcho3: No, I shut them down.

python-package/xgboost/sklearn.py
```diff
@@ -212,6 +212,105 @@ def adddoc(cls):
...
+    base_margin: Optional[Any],
+    feature_weights: Optional[Any],
+    eval_set: Optional[List[Tuple[Any, Any]]],
+    sample_weight_eval_set: Optional[List[Any]],
+    base_margin_eval_set: Optional[List[Any]],
+    eval_group: Optional[List[Any]],
+    eval_qid: Optional[List[Any]],
+    create_dmatrix: Callable,
+    label_transform: Callable = lambda x: x,
+) -> Tuple[Any, Optional[List[Tuple[Any, str]]]]:
```
hcho3, line 231:
```suggestion
) -> Tuple[EvalMatrix, Optional[List[Tuple[EvalMatrix, str]]]]:
```
Let's define a union type `EvalMatrix` as `Union[DMatrix, "DaskDMatrix"]`. This way, we can specifically annotate the return type of `_wrap_evaluation_matrices`.

tests/python/test_with_dask.py
```diff
@@ -149,68 +149,46 @@ def test_dask_predict_shape_infer() -> None:
 
 
 @pytest.mark.parametrize("tree_method", ["hist", "approx"])
-def test_boost_from_prediction(tree_method: str) -> None:
-    if tree_method == 'approx':
-        pytest.xfail(reason='test_boost_from_prediction[approx] is flaky')
```
hcho3, line 154: Is the test no longer flaky?

hcho3 (commented):
Thanks for consolidating common logic. It looks much nicer.

python-package/xgboost/core.py
```diff
@@ -434,14 +437,26 @@ class DMatrix:                  # pylint: disable=too-many-instance-attributes
...
+        data,
+        label=None,
+        *,
+        weight=None,
+        base_margin=None,
+        group=None,
+        qid=None,
+        label_lower_bound=None,
+        label_upper_bound=None,
+        missing: Optional[float] = None,
```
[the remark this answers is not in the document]
trivialfis: Considering the number of meta info is growing, I think it's reasonable choice to stick with positional only arguments.

python-package/xgboost/sklearn.py
trivialfis: @hcho3 We can't, the `DaskDMatrix` is not defined on this module.  I tried using `TypeVar` but there seems to be failure in type deduction.

tests/python/test_with_dask.py
trivialfis: @hcho3 I rewrote the test.

trivialfis: That doesn't mean it's no longer flaky, we will have to see in the future.

trivialfis:
Since I have spited this PR into smaller part in https://github.com/dmlc/xgboost/pull/6601 and will continue to do so for better review, I will add links to related PRs.

python-package/xgboost/core.py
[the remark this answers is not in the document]
hcho3: > it's reasonable choice to stick with positional only arguments.

Did you mean to say "keyword-only arguments"?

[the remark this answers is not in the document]
trivialfis: Yup, sorry.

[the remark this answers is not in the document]
hcho3: It is possible to delay the breaking change by moving the `missing` parameter to right after `base_margin` so as to preserve the current parameter order. This would give users some time to switch over to keyword-only args.

Is there a good reason for moving the `missing` parameter right now?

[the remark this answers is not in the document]
trivialfis: I will try to move parameters.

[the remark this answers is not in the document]
hcho3: If the re-ordering is to make Dask and non-Dask interface consistent, then I'm okay with the breaking change. Is this the case here?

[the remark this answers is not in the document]
trivialfis: I think you are right that it doesn't have to be breaking change.  Let me try a bit harder after the fixing up the mess I made with the patch rel.

trivialfis:
Added a test for avoiding duplicated DMatrix after rebasing.

trivialfis:
@hcho3 Ready for another round of review.

hcho3 (approved):

PR #6593, opened 2021-01-11
  [breaking] Remove duplicated predict functions, Fix attributes IO. 

trivialfis:
* Use inheritance to remove duplicated code.
* Fix attributes not being restored.
* Rename all `data` to `X`. [breaking]

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -682,10 +682,16 @@ class XGBModel(XGBModelBase):
         self._set_evaluation_result(evals_result)
         return self
 
-    def predict(self, data, output_margin=False, ntree_limit=None,
-                validate_features=True, base_margin=None):
+    def predict(
+        self,
+        X,
+        output_margin=False,
+        ntree_limit=None,
+        validate_features=True,
+        base_margin=None
+    ):
         """
-        Predict with `data`.
+        Predict with `X`.
 
         .. note:: This function is not thread safe.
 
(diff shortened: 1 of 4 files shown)


hcho3:
This PR is only breaking for the users who specified the `data` argument as a keyword argument?

trivialfis:
> This PR is only breaking for the users who specified the data argument as a keyword argument?

Yes.

trivialfis:
This is part of cleaning up the Python code base.  https://github.com/dmlc/xgboost/pull/6594 is also used for cleaning up, but happens to fix an old hidden bug.  Last big refactor is in https://github.com/dmlc/xgboost/pull/6591 . The goal is to bring the dask and sklearn module together.

trivialfis:
@hcho3 This turned up to be another fix for attributes not being restored properly after reading model.

python-package/xgboost/sklearn.py
```diff
@@ -1037,50 +1048,21 @@ def fit(self, X, y, *, sample_weight=None, base_margin=None,
...
-        class_probs = self.get_booster().predict(
-            test_dmatrix,
+    def predict(
+        self,
+        X,
+        output_margin=False,
+        ntree_limit=None,
+        validate_features=True,
+        base_margin=None
+    ):
```
hcho3, line 1058: What happened to the docstring for this `predict()` function?

hcho3 (approved):

python-package/xgboost/sklearn.py
trivialfis: The doc string is inherited from `XGBModel`.

PR #6601, opened 2021-01-12
Define metainfo for all DMatrix interfaces.

trivialfis:
A small part extracted from https://github.com/dmlc/xgboost/pull/6591 to ease reviewing.

* Add check for consistency between 4 different DMatrix types.  They should share a number of common arguments in constructor with correct ordering.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -312,15 +312,18 @@ class DataIter:
                 data, feature_names, feature_types
             )
             dispatch_device_quantile_dmatrix_set_data(self.proxy, data)
-            self.proxy.set_info(label=label, weight=weight,
-                                base_margin=base_margin,
-                                group=group,
-                                qid=qid,
-                                label_lower_bound=label_lower_bound,
-                                label_upper_bound=label_upper_bound,
-                                feature_names=feature_names,
-                                feature_types=feature_types,
-                                feature_weights=feature_weights)
+            self.proxy.set_info(
+                label=label,
+                weight=weight,
+                base_margin=base_margin,
+                group=group,
+                qid=qid,
+                label_lower_bound=label_lower_bound,
+                label_upper_bound=label_upper_bound,
+                feature_names=feature_names,
+                feature_types=feature_types,
+                feature_weights=feature_weights
+            )
         try:
             # Differ the exception in order to return 0 and stop the iteration.
             # Exception inside a ctype callback function has no effect except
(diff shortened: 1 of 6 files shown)


python-package/xgboost/core.py
```diff
@@ -427,21 +430,34 @@ def inner_f(*args, **kwargs):
...
+        label=None,
+        weight=None,
+        base_margin=None,
+        missing: Optional[float] = None,
+        *,
+        group=None,
+        qid=None,
+        label_lower_bound=None,
+        label_upper_bound=None,
+        silent: bool = False,
```
hcho3, line 454: Per my last remark about compatibility, please move this `slient` parameter to line 450 (in between `*` and `group`).

python-package/xgboost/core.py
```diff
@@ -937,39 +984,67 @@ class DeviceQuantileDMatrix(DMatrix):
...
+        label=None,
+        weight=None,
+        base_margin=None,
+        missing=None,
+        *,
+        group=None,
+        qid=None,
+        label_lower_bound=None,
+        label_upper_bound=None,
+        silent=False,
```
hcho3, line 1006: Same comment.

hcho3 (changes requested):
Please address my comment at https://github.com/dmlc/xgboost/pull/6591#discussion_r556381288. You should choose between two alternatives:

1) Break backward compatibility now. This might be reasonable choice if it's necessary to unity Dask and non-Dask interfaces; OR
2) Don't break backward compatibility now; only add deprecation warning. For this alternative, all new parameters should only be added at the end of each function, not in the middle, to avoid breaking existing code using positional args.

My review below assumes that you've chosen Option 2. Let me know if you want to go with Option 1.

trivialfis:
> Please address my comment at #6591 (comment). You should choose between two alternatives:

I will try to avoid breaking the compatibility.

trivialfis:
@hcho3 I limited the breaking changes to `enable_categorical` and dask interface.

python-package/xgboost/core.py
trivialfis: The compatibility for `DMatrix` is kept.  But the `*` is still placed next to `label` to provide a warning for next release.  WDYT?

hcho3: Great! Thanks for keeping the order for the existing parameters.

hcho3 (approved):

PR #6609, opened 2021-01-15
Use simple print in tracker print function.

trivialfis:
The dask training doesn't print out the evaluation result even if one passes the `verbose_eval` argument.  This PR fixes it.  I don't know how to setup a test as the printing is not done in current process so the stdout cannot be easily captured.

### DIFF
--- a/python-package/xgboost/tracker.py
+++ b/python-package/xgboost/tracker.py
@@ -293,7 +293,7 @@ class RabitTracker(object):
             s = SlaveEntry(fd, s_addr)
             if s.cmd == 'print':
                 msg = s.sock.recvstr()
-                logging.info(msg.strip())
+                print(msg.strip(), flush=True)
                 continue
             if s.cmd == 'shutdown':
                 assert s.rank >= 0 and s.rank not in shutdown
--- a/python-package/xgboost/rabit.py
+++ b/python-package/xgboost/rabit.py
@@ -79,8 +79,7 @@ def tracker_print(msg):
     if is_dist != 0:
         _check_call(_LIB.RabitTrackerPrint(c_str(msg)))
     else:
-        sys.stdout.write(msg)
-        sys.stdout.flush()
+        print(msg.strip(), flush=True)
 
 
 def get_processor_name():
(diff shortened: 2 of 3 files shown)


trivialfis:
The downside is the evaluation is printed on scheduler instead of client.

python-package/xgboost/tracker.py
```diff
@@ -293,7 +293,7 @@ def accept_slaves(self, nslave):
             s = SlaveEntry(fd, s_addr)
             if s.cmd == 'print':
                 msg = s.sock.recvstr()
-                logging.info(msg.strip())
```
hcho3, line 296: The `logging` module should be preferred over the simple `print`, since Dask has a good integration with the `logging` module. See related discussion at https://github.com/rapidsai/cuml/issues/3350.

trivialfis: @hcho3 I definitely want to use logging.  But what logging level should be used?  To print out the evaluation history, it should not be info, as the default in Python logging is warning.  But also the evaluation history should not be warning right?

trivialfis: Quoting from https://docs.python.org/3/howto/logging.html :


Task you want to perform | The best tool for the task
-- | --
Display console output for ordinary usage of a command line script or program | print()
Report events that occur during normal operation of a program (e.g. for status monitoring or fault investigation) | logging.info() (or logging.debug() for very detailed output for diagnostic purposes)
Issue a warning regarding a particular runtime event | warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted
Report an error regarding a particular runtime event | Raise an exception
Report suppression of an error without raising an exception (e.g. error handler in a long-running server process) | logging.error(), logging.exception() or logging.critical() as appropriate for the specific error and application domain

I think simple print is a reasonable choice here.

trivialfis: @hcho3 

I don't think we should print something like this:

```
INFO:root:test-data:error:2.71 train-data:error:2.33
...
```

hcho3: Got it. Thanks for the reference.

hcho3 (approved):

Issue #6615, opened 2021-01-18
Incorrect `best_ntree_limit`.


trivialfis:
The fix https://github.com/dmlc/xgboost/pull/6579 got in 1.3.2 was actually wrong.  Here's a brief summary:

The old (before fix) `best_ntree_limit` ignores the `num_class` parameters, which is incorrect.  In before we workarounded it in c++ layer to avoid possible breaking changes on other language bindings.  But the Python interpretation stayed incorrect.  The PR fixed that in Python to consider `num_class`, but didn't remove the old workaround, so tree calculation in predictor is incorrect, see `PredictBatch` in `CPUPredictor`.

Proposal:  Revert the fix for now, and deprecate the parameter in next release.

cc @pseudotensor @hcho3

trivialfis:
R binding is consistent with the old Python behavior, see `callbacks.R`.

hcho3:
I agree that we should deprecate `ntree_limit` argument and encourage the use of model slicing. For now, let's revert the change to minimize the risk of breaking the release.

@trivialfis Do you plan to make another release? Should it be 1.3.2.post0 (hot-fix) or 1.3.3 (another patch release)?

trivialfis:
I will make a 1.3.3 release on Python side.  Thinking about how to add a test that can check for an incorrect behavior.

trivialfis:
PR in https://github.com/dmlc/xgboost/pull/6616 .  Will backport.

PR #6616, opened 2021-01-18
Revert ntree limit fix

trivialfis:
Close https://github.com/dmlc/xgboost/issues/6615 .

### DIFF
--- a/python-package/xgboost/training.py
+++ b/python-package/xgboost/training.py
@@ -112,6 +112,7 @@ def _train_internal(params, dtrain,
     if bst.attr('best_score') is not None:
         bst.best_score = float(bst.attr('best_score'))
         bst.best_iteration = int(bst.attr('best_iteration'))
+        # num_class is handled internally
         bst.set_attr(
             best_ntree_limit=str((bst.best_iteration + 1) * num_parallel_tree)
         )
--- a/python-package/xgboost/training.py
+++ b/python-package/xgboost/training.py
@@ -109,22 +109,18 @@ def _train_internal(params, dtrain,
     else:
         raise ValueError(f'Unknown booster: {booster}')
 
-    num_groups = int(config['learner']['learner_model_param']['num_class'])
-    num_groups = 1 if num_groups == 0 else num_groups
     if bst.attr('best_score') is not None:
         bst.best_score = float(bst.attr('best_score'))
         bst.best_iteration = int(bst.attr('best_iteration'))
         bst.set_attr(
-            best_ntree_limit=str(
-                (bst.best_iteration + 1) * num_parallel_tree * num_groups
-            )
+            best_ntree_limit=str((bst.best_iteration + 1) * num_parallel_tree)
         )
         bst.best_ntree_limit = int(bst.attr("best_ntree_limit"))
     else:
         # Due to compatibility with version older than 1.4, these attributes are added
         # to Python object even if early stopping is not used.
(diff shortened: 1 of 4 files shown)


tests/python/test_predict.py
```diff
@@ -52,12 +58,19 @@ def run_predict_leaf(predictor):
...
             layer = row[start: end]
             for c in range(classes):
-                tree_group = layer[c * num_parallel_tree:
-                                   (c+1) * num_parallel_tree]
+                tree_group = layer[c * num_parallel_tree: (c + 1) * num_parallel_tree]
                 assert tree_group.shape[0] == num_parallel_tree
                 # no subsampling so tree in same forest should output same
                 # leaf.
                 assert np.all(tree_group == tree_group[0])
+
```
trivialfis, line 66: only this test is new.

python-package/xgboost/training.py
```diff
@@ -109,22 +109,19 @@ def _train_internal(params, dtrain,
...
-    num_groups = 1 if num_groups == 0 else num_groups
     if bst.attr('best_score') is not None:
         bst.best_score = float(bst.attr('best_score'))
         bst.best_iteration = int(bst.attr('best_iteration'))
+        # num_class is handled internally
         bst.set_attr(
-            best_ntree_limit=str(
-                (bst.best_iteration + 1) * num_parallel_tree * num_groups
-            )
+            best_ntree_limit=str((bst.best_iteration + 1) * num_parallel_tree)
```
trivialfis, line 117: It's not an exact revert since we also have fix for an old bug for `gblinear` with `ntree_limit`, which is valid.

hcho3: So the `best_ntree_limit` attribute is really a misnomer; it should have been called `best_iteration`. Is my understanding correct?

trivialfis: The C++ layer considers `num_group` as it's a model parameter, and predictor `PredictBatch` has `GBTreeModel` as its argument.  But it doesn't consider `num_parallel_tree`, which is a training parameter instead of model parameter.  So the C++ function for `ntree_limit` is half implementation for `best_iteration`.

hcho3: So it's `best_iteration * num_parallel_tree` then? We should clearly document the meaning of `best_ntree_limit`, e.g.
```
Despite its name, the best_ntree_limit attribute is actually the product (best_iteration * num_parallel_tree).
If you set num_parallel_tree > 1, then best_ntree_limit won't be equal to the best boosting round.
```

trivialfis: Just to clarify and reinforce:  The `best_ntree_limit` equals `num_parallel_tree * best_iteration`.  `num_class` is multiplied inside predictor.

When using classifier, the `best_ntree_limit` equals to `number of trees produced in best iteration / num_class`.

trivialfis: The inplace prediction doesn't have this problem as it can use `best_iteration` directly.

trivialfis: @hcho3 I will try to deprecate the attribute after setting it as a Python `@property` where we can throw proper warnings.

trivialfis: Also, the sklearn model uses this attribute by default when running prediction.

trivialfis: Added a short note on `train` function.

hcho3 (approved):

PR #6625, opened 2021-01-20
Exclude dmlc test on github action.

trivialfis:


trivialfis:
The test in thraeded iterator is flaky os osx as it's a timing based test.

hcho3 (approved):

PR #6638, opened 2021-01-26
[WIP] Rework predict functions.

trivialfis:
Related: https://github.com/dmlc/xgboost/issues/6632  Instead of a `struct`, I went with a JSON string for any extra arguments.  Constructing `struct` in other languages are not necessarily easy.

This is a monolithic PR for early review and discussion, I will split it up into multiple smaller PRs.

## Changes

* Add new function for predicting from DMatrix.
* Use `begin_iteration` and `end_iteration` to replace `ntree_limit`.
* Change Python native, sklearn and dask interfaces for `iteration_range` and deprecate `ntree_limit`.
* Change inplace prediction functions to support `base_margin` and possibly other meta info in the future.
* Use inplace predict by default in sklearn and dask sklearn interfaces.
* Expose inplace predict in C header.
* The prediction output shape is now calculated in C++, with an additional parameter `strict_shape` to allow more explicit dimensions.  For example, when set to `True`, `predict_leaf` might output a 4 dimension array with respected to `(n_samples, n_iterations, n_classes, n_trees_in_forest)`.  Default is `False` so compatibility is kept.
* The adapters for dense and CSR are changed to accept more data types instead of strictly float32, this helps making the `inplace_predict` function to be truly inplace for all supported data types.
* All prediction function are made to be thread safe.

## Tracking extracted PRs.

- https://github.com/dmlc/xgboost/pull/6644
- https://github.com/dmlc/xgboost/pull/6645
- https://github.com/dmlc/xgboost/pull/6648
- https://github.com/dmlc/xgboost/pull/6653
- https://github.com/dmlc/xgboost/pull/6668

## Related issues:

- https://github.com/dmlc/xgboost/issues/6632
- https://github.com/dmlc/xgboost/issues/6621
- https://github.com/dmlc/xgboost/issues/6553

PR #6645, opened 2021-01-27
[dask] Add a dummy sample to infer output shape.

trivialfis:
This is for inferring shape with direct prediction (without DaskDMatrix).
There are a few things that requires known output shape before carrying out
actual prediction, including dask meta data, output dataframe columns.

* Infer output shape based on local prediction.
* Remove set param in predict function as it's not thread safe nor necessary as
we now let dask to decide the parallelism.
* Simplify prediction on `DaskDMatrix`.
* Remove unnecessary serialization.

A small part extracted from https://github.com/dmlc/xgboost/pull/6638 with added test and remove redundant serialization.

### DIFF
--- a/doc/tutorials/dask.rst
+++ b/doc/tutorials/dask.rst
@@ -108,10 +108,9 @@ computation a bit faster when meta information like ``base_margin`` is not neede
   prediction = xgb.dask.inplace_predict(client, output, X)
 
 Here ``prediction`` is a dask ``Array`` object containing predictions from model if input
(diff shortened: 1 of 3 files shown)


trivialfis:
This PR also fixes a performance issue in predict function that the booster might got serialized multiple times in `map_blocks/partitions`.

trivialfis:
This PR should significantly improve the performance for prediction.

@pseudotensor

trivialfis:
https://github.com/dmlc/xgboost/pull/6648 needs to be merged first.

RAMitchell (approved):

PR #6647, opened 2021-01-27
[Breaking] Change default evaluation metric for binary:logitraw objective to logloss

### LINKED ISSUE #6618: Change default evaluation metric for binary:logitraw to logloss, to be consistent with binary:logistic
https://discuss.xgboost.ai/t/why-is-it-so-slow-with-objective-binary-logitraw/2077

> I have observed a big running time difference with objective=“binary:logistic” and objective=“binary:logitraw” in R with xgboost

hcho3:
Closes #6618

cc @lorentzenchr @mayer79

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -1035,16 +1035,18 @@ class LearnerImpl : public LearnerIO {
     os << '[' << iter << ']' << std::setiosflags(std::ios::fixed);
     if (metrics_.size() == 0 && tparam_.disable_default_eval_metric <= 0) {
       auto warn_default_eval_metric = [](const std::string& objective, const std::string& before,
-                                         const std::string& after) {
-        LOG(WARNING) << "Starting in XGBoost 1.3.0, the default evaluation metric used with the "
-                     << "objective '" << objective << "' was changed from '" << before
-                     << "' to '" << after << "'. Explicitly set eval_metric if you'd like to "
-                     << "restore the old behavior.";
+                                         const std::string& after, const std::string& version) {
+        LOG(WARNING) << "Starting in XGBoost " << version << ", the default evaluation metric "
+                     << "used with the objective '" << objective << "' was changed from '"
+                     << before << "' to '" << after << "'. Explicitly set eval_metric if you'd "
+                     << "like to restore the old behavior.";
       };
       if (tparam_.objective == "binary:logistic") {
-        warn_default_eval_metric(tparam_.objective, "error", "logloss");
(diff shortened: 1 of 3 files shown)


src/learner.cc
```diff
@@ -1035,16 +1035,18 @@ class LearnerImpl : public LearnerIO {
     os << '[' << iter << ']' << std::setiosflags(std::ios::fixed);
     if (metrics_.size() == 0 && tparam_.disable_default_eval_metric <= 0) {
       auto warn_default_eval_metric = [](const std::string& objective, const std::string& before,
-                                         const std::string& after) {
-        LOG(WARNING) << "Starting in XGBoost 1.3.0, the default evaluation metric used with the "
-                     << "objective '" << objective << "' was changed from '" << before
-                     << "' to '" << after << "'. Explicitly set eval_metric if you'd like to "
-                     << "restore the old behavior.";
+                                         const std::string& after, const std::string& version) {
+        LOG(WARNING) << "Starting in XGBoost " << version << ", the default evaluation metric "
```
hcho3, line 1039: Note. We should plan to remove this warning soon. Probably right after 1.4.0 ?

hcho3: Does the warning go to `logging` or stdout?

trivialfis: @hcho3 stdout

trivialfis: That's the `_log_calback` function in `xgboost/core.py`.

hcho3: Got it. We could modify `_log_callback` to use the `logging` module. Alternatively, we could simply remove the warning message.

lorentzenchr:
Is there a test for the changed behaviour or a test to adapt because of the change?

src/learner.cc
trivialfis: We can do both in the future, but it will be another breaking change.  To use `logging`, one needs to know the verbosity.  So `_log_callback` should have 2 parameters `_log_callback(msg: str, verbosity: int)` to indicate the level of current message.

hcho3: Got it. Let's remove the warning message after 1.4.0 release.

trivialfis (approved):

hcho3:
> Is there a test for the changed behaviour or a test to adapt because of the change?

Unlike #6183, not much of the existing code is affected by this change, since `binary:logitraw` is not used as often as `binary:logistic`.

lorentzenchr (approved):
LGTM, too 👍

PR #6653, opened 2021-01-28
Enhance inplace prediction.

trivialfis:
* Accept array interface for csr and array.
* Accept an optional proxy dmatrix for metainfo.

Extracted from https://github.com/dmlc/xgboost/pull/6638 .

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -58,21 +58,23 @@ CallbackEnv = collections.namedtuple(
      "evaluation_result_list"])
 
 
-def from_pystr_to_cstr(data):
-    """Convert a list of Python str to C pointer
+def from_pystr_to_cstr(data: Union[str, List[str]]):
+    """Convert a Python str or list of Python str to C pointer
 
     Parameters
     ----------
-    data : list
-        list of str
+    data
+        str or list of str
     """
 
-    if not isinstance(data, list):
-        raise NotImplementedError
-    pointers = (ctypes.c_char_p * len(data))()
-    data = [bytes(d, 'utf-8') for d in data]
-    pointers[:] = data
-    return pointers
+    if isinstance(data, str):
+        return bytes(data, "utf-8")
+    if isinstance(data, list):
+        pointers = (ctypes.c_char_p * len(data))()
+        data = [bytes(d, 'utf-8') for d in data]
+        pointers[:] = data
+        return pointers
+    raise TypeError()
 
 
 def from_cstr_to_pystr(data, length):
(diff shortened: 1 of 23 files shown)


trivialfis:
The old adapters for dense and csr are kept to preserve API for DMatrix.

python-package/xgboost/core.py
```diff
@@ -1579,9 +1611,17 @@ def predict(self,
...
-        '''Run prediction in-place, Unlike ``predict`` method, inplace prediction does
+    def inplace_predict(
+        self,
+        data,
+        iteration_range: Tuple[int, int] = (0, 0),
+        predict_type: str = "value",
+        missing: float = np.nan,
+        validate_features: bool = True,
+        base_margin: Any = None,
+        strict_shape: bool = False
```
trivialfis, line 1622: This parameter may needs more thought.  It's meant to be a placeholder for https://github.com/dmlc/xgboost/pull/6638 .  It affects only the leaf prediction.

python-package/xgboost/core.py
```diff
@@ -959,6 +973,22 @@ def feature_types(self, feature_types: Optional[Union[List[Any], Any]]) -> None:
...
+    inplace_predict).  `owner` specifies whether should the handle be freed by this
+    object.
+
+    """
+    def __init__(self, owner):  # pylint: disable=super-init-not-called
+        self.handle = ctypes.c_void_p()
+        _check_call(_LIB.XGProxyDMatrixCreate(ctypes.byref(self.handle)))
+        self._owner = owner
+
+    def __del__(self):
```
RAMitchell, line 987: This is dangerous. Any other way to do this?

src/data/adapter.h
```diff
@@ -228,6 +228,118 @@ class DenseAdapter : public detail::SingleBatchDataIter<DenseAdapterBatch> {
   size_t num_columns_;
 };
 
+class ArrayAdapterBatch : public detail::NoMetaInfo {
```
RAMitchell, line 231: Maybe leave some docs here as to why duplicate classes exist.

RAMitchell (commented):
There is a lot of effort to support base margin here. Why is it so important? Couldnt you just add it to the out prediction in numpy?

trivialfis:
@RAMitchell no, there's predict transform.  It isn't a lots of effort I think, all the needed tools are there, the PR is just gluing them together.

python-package/xgboost/core.py
trivialfis: Let me think about it.

trivialfis: Simplified it by removing some indirect functions.

RAMitchell (approved):

PR #6656, opened 2021-01-28
Add use_rmm flag to global configuration

### LINKED ISSUE #6297: Add an explicit option to use RMM

hcho3:
Closes #6297. The user should explicitly inform XGBoost if RMM is used. If RMM is used, XGBoost should not allocate from the CUB allocator.

### DIFF
--- a/demo/rmm_plugin/rmm_mgpu_with_dask.py
+++ b/demo/rmm_plugin/rmm_mgpu_with_dask.py
@@ -5,6 +5,9 @@ from dask.distributed import Client
 from dask_cuda import LocalCUDACluster
 
 def main(client):
+    # Inform XGBoost that RMM is used for GPU memory allocation
+    xgb.set_config(use_rmm=True)
+
     X, y = make_classification(n_samples=10000, n_informative=5, n_classes=3)
     X = dask.array.from_array(X)
     y = dask.array.from_array(y)
--- a/demo/rmm_plugin/rmm_mgpu_with_dask.py
+++ b/demo/rmm_plugin/rmm_mgpu_with_dask.py
@@ -14,7 +14,7 @@ def main(client):
     dtrain = xgb.dask.DaskDMatrix(client, X, label=y)
 
     params = {'max_depth': 8, 'eta': 0.01, 'objective': 'multi:softprob', 'num_class': 3,
-              'tree_method': 'gpu_hist'}
+              'tree_method': 'gpu_hist', 'eval_metric': 'merror'}
     output = xgb.dask.train(client, params, dtrain, num_boost_round=100,
                             evals=[(dtrain, 'train')])
     bst = output['booster']
--- a/doc/parameter.rst
+++ b/doc/parameter.rst
@@ -22,6 +22,7 @@ Global Configuration
 The following parameters can be set in the global scope, using ``xgb.config_context()`` (Python) or ``xgb.set.config()`` (R).
 
 * ``verbosity``: Verbosity of printing messages. Valid values of 0 (silent), 1 (warning), 2 (info), and 3 (debug).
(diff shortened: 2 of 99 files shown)


tests/ci_build/conda_env/cpu_test.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
 - pyyaml
 - cpplint
 - pylint
-- numpy
+- numpy=1.19.4
```
hcho3, line 11: This is a workaround, as the latest `shap` package appears to break with NumPy 1.19.5:
https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-6656/4/pipeline/221
```
     try:
>       from .. import _cext
E    ImportError: numpy.core.multiarray failed to import
```

demo/rmm_plugin/rmm_mgpu_with_dask.py
```diff
@@ -5,13 +5,16 @@
...
+    xgb.set_config(use_rmm=True)
+
     X, y = make_classification(n_samples=10000, n_informative=5, n_classes=3)
     X = dask.array.from_array(X)
     y = dask.array.from_array(y)
     dtrain = xgb.dask.DaskDMatrix(client, X, label=y)
 
     params = {'max_depth': 8, 'eta': 0.01, 'objective': 'multi:softprob', 'num_class': 3,
-              'tree_method': 'gpu_hist'}
+              'tree_method': 'gpu_hist', 'eval_metric': 'merror'}
```
trivialfis, line 17: Is this related to rmm?

doc/parameter.rst
```diff
@@ -22,6 +22,7 @@ Global Configuration
 The following parameters can be set in the global scope, using ``xgb.config_context()`` (Python) or ``xgb.set.config()`` (R).
 
 * ``verbosity``: Verbosity of printing messages. Valid values of 0 (silent), 1 (warning), 2 (info), and 3 (debug).
+* ``use_rmm``: Whether to use RAPIDS Memory Manager (RMM) to allocate GPU memory. This option is only applicable when XGBoost is built (compiled) with the RMM plugin enabled. Vaid values are ``true`` and ``false``.
```
trivialfis, line 25: `vaid` -> `valid`.

tests/ci_build/conda_env/cpu_test.yml
hcho3: @trivialfis Have you run into this issue? What's your take on it?

trivialfis: never seen it before.

> What's your take on it?

@RAMitchell worked on the c extension before so maybe have better insight.

demo/rmm_plugin/rmm_mgpu_with_dask.py
hcho3: This updates the demo so that it doesn't display the warning about default evaluation metric.

tests/ci_build/conda_env/cpu_test.yml
hcho3: Using `shap` from conda-forge also fixes the issue, but unfortunately conda-forge has an outdated version of `shap`.

trivialfis (approved):

PR #6662, opened 2021-01-30
Align device id in predict transform with predictor.

trivialfis:


tests/cpp/objective/test_objective.cc
```diff
@@ -17,3 +17,25 @@ TEST(Objective, UnknownFunction) {
     delete obj;
   }
 }
+
+namespace xgboost {
+TEST(Objective, PredTransform) {
+  // Test that show PredTransform uses the same device with predictor.
+  xgboost::GenericParameter tparam;
+  Args args{{"gpu_id", "0"}};  // set to deviec
```
trivialfis, line 25: Typo.

trivialfis:
This PR let `PredTransform` uses the device from prediction vector instead of `gpu_id` parameter.

RAMitchell (approved):

Issue #2914, opened 2017-12-01
Clarification on Trivial Splits


cowden:
I have a question regarding how xgboost splits out missing values.  The exact greedy and approximate tree methods tend to split out missing values by splitting at twice the maximum value on a feature.  Whereas the histogram method splits at the minimum value of a feature.  The factor of 2 seems consistent, so I don't think it has any dependence on the distribution of the feature.

Just for reference, I've been able to consistently see this kind of split in the latest release (v0.60). 

I've found some threads discussing the treatment of missing data [2],[3],[4], but I'm interesting in understanding why 2?

So my question is the following:
Is the decision to exclusively split away missing values at twice the maximum of a variable a consequence/artifact/feature of the split proposal algorithm described in [1], or is it a feature of the implementation?

If it's the latter case, can this parameter be tuned?

[1] https://arxiv.org/abs/1603.02754
[2] https://github.com/dmlc/xgboost/issues/21
[3] https://github.com/dmlc/xgboost/issues/1112
[4] https://github.com/dmlc/xgboost/issues/1842

hcho3:
I've worked on the histogram implementation and here's my take on it:

1) The factor 2 is simply an implementation choice; there's nothing inherent in the algorithm that forces it. It's probably to give enough cushion to ensure all non-missing entries fall on one side of the split. The factor 2 is introduced by a choice in the approximate quantile sketcher:
```
/* Lines 95-101 of src/common/hist_util.cc */
    // push a value that is greater than anything
    if (a.size != 0) {
      bst_float cpt = a.data[a.size - 1].value;
      // this must be bigger than last value in a scale
      bst_float last = cpt + fabs(cpt);
      cut.push_back(last);
    }
```
So the factor 2 is currently not tune-able.

2) As for the choice between minimum and maximum values, we should look at how candidate splits are enumerated. In exact and approximate methods, the candidate splits are scanned from left to right, and then right to left:
```
/* Lines 630-637 of updater_colmaker.cc */
          if (param.need_forward_search(fmat.GetColDensity(fid), ind)) {
            this->EnumerateSplit(c.data, c.data + c.length, +1,
                                 fid, gpair, info, stemp[tid]);  // note +1
          }
          if (param.need_backward_search(fmat.GetColDensity(fid), ind)) {
            this->EnumerateSplit(c.data + c.length - 1, c.data - 1, -1,
                                 fid, gpair, info, stemp[tid]);  // note -1
          }
```
[...]

PR #6666, opened 2021-02-02
Optimizing BoostNewTrees

merleyc:
The first XGBoost iteration (BoostNewTrees method call) takes more time than the subsequent ones. For example, by using batch mode with all available threads in the mortgage dataset, the first iteration takes ~14x more than the others.
The reason is because it initializes the DMatrix data during the first iteration. Specifically, the data manipulation done in PushRowPage method is the hotspot, which currently divides the DMatrix data into OMP threads by columns (features). However, it will be more efficient to split the data into OMP thread by rows instead of columns considering that the recommendation is to have more instances than features*. 
 
**Approach:** Optimized the code to use (1) rows of the input data to separate OMP threads instead of columns and (2) creates a sketches manager to deal with sketches created by different threads, (3) and finally merges the sketches into a single one by using already existing method, called PushSummary. PushSummary was implemented by @canonizer in this [commit](https://github.com/dmlc/xgboost/commit/cc6a5a3666bced8c5090d9f9a25c1325b7d2cf57).

**Results:** For batch mode, the improvement for the first iteration is ~47% and for the total training runtime is ~5.4%. More details in Appendix 1
With this optimization the models and sketches become slightly different but the validation scores (RMSE for regression and F1 & ACC for classification) are similar.

**Future work:** The optimization is focused on dense data. However there is room for improvement when data is not dense but almost dense, since the original code assumes that 1 feature with value 0 makes the dataset sparse. If we add a threshold to consider data dense, even with a few features with value 0, then this optimization could be applied to them as well.

**Appendix 1:**
Regression task using commit c3c8e66fc9d285c265b07d78fb6866770a65347b from Jan 28th 2021. Dataset: Mortgage in CSV format with shape=(31540064, 46)

batchMode

Thank you!


trivialfis:
Could you please try passing this google unittest first: `HistUtil.DenseCutsAccuracyTest` ?

trivialfis:
The failed cpp test is in `tests/cpp/common/test_hist_utils.cc`.

merleyc:
image


> The failed cpp test is in `tests/cpp/common/test_hist_utils.cc`.

merleyc:
Sorry I closed this PR by accident but I just reopened it.

@trivialfis For "google unittest first: HistUtil.DenseCutsAccuracyTest" are you referring to the test "XGBoost-CI / Test Google C++ test (CPU) (macos-10.15)"? 
I have tested it in my machine using and it is passing. Please see below.
Command used: cd build; ctest --exclude-regex AllTestsInDMLCUnitTests --extra-verbose
I have a Centos 7.
image

merleyc:
Hi @trivialfis ,

The optimization changes the way the data is split by OMP_THREADS in the PushRowPage method. This method is called in SketchOnDMatrix which is in the failing test [[HERE](https://github.com/dmlc/xgboost/blob/956beead705118fd2513da79af9ccfe9cef59a0d/tests/cpp/common/test_hist_util.cc#L197)].

The optimization changes slightly the values in the final tree ensemble, maybe is that the reason this test is failing in github?
I cannot reproduce the failing test to fix it since for me it is passing. Can you please check if it is failing for you as well?

Thank you

trivialfis:
Sorry for the late reply.  Let me take a deeper look.

merleyc:
Thanks for taking a deeper look.

Under **"Test Google C++ test (CPU) (macos-10.15)"**, there were 4 tests failing. I did the following changes [here](https://github.com/dmlc/xgboost/blob/956beead705118fd2513da79af9ccfe9cef59a0d/tests/cpp/common/test_hist_util.h) 

Change 1:
BEFORE: int allowable_error = std::max(2, int(expected_bin_weight * 0.3)); 
AFTER: int allowable_error = std::max(10, int(expected_bin_weight * 0.3));

My question is why is the allowable_error should be minimum 2 and not any other value?

Change 2:
BEFORE:ASSERT_EQ(cuts.SearchBin(v, column_idx), cuts.Ptrs()[column_idx] + i);
AFTER: EXPECT_LE(std::abs(int(cuts.SearchBin(v, column_idx) - (cuts.Ptrs()[column_idx] + i))), 1);

Any recommendation on how to fix the error. I know my change in the test might not be right thing to do.

The following tests are failing because of the same comparison **"Test JVM on OS ubuntu-latest & Test JVM on OS windows-latest"** in ScalaBoosterImplSuite.scala:89.
It seems like my change slightly decreased the auc value in some iteration. From the test:
test with quantile histo depthwise *** FAILED *** 
_0.984811 was not greater than or equal to 0.985305 (ScalaBoosterImplSuite.scala:89)_

Any recommendation on how to tackle this? Or is this delta acceptable?

Thank you

trivialfis:
@merleyc All are good questions and I would like to get in depth answers too.  The sketching is tightly connected to split value finding: https://github.com/dmlc/xgboost/issues/2914 .  I will try to investigate and document it in a precise manner.

merleyc:
Thanks for sharing te PR2914. I will wait you get back to me regarding the CI test results and questions. Thanks!

merleyc:
Hi @trivialfis ,
Did you have a chance to check if my fixes in the CI tests are ok?
Thanks!

trivialfis:
Not yet, working on auc.

Issue #6621, opened 2021-01-19
[RFC] Deprecating `ntree_limit` in favor of `best_iteration`.


trivialfis:
The use of `ntree_limit` is error prone in the context of multi-class and random forest.  The calculation is as followed:

```
ntree_limit = num_class * num_parallel_tree * iteration
```

This looks simple, but due to some historical reasons, the `num_class` is handled internally while `num_parallel_tree` is handled in language bindings.  The behavior generates unnecessary confusion.  Right now there are 2 alternatives in xgboost, first one is the `inplace_predict` function, which accepts iteration number directly, second one is model slicing, which allows user to extract a part of model using iteration number.

The deprecation might be a huge breaking change, so I would like to hear some suggestions on how we move forward.  Many thanks!

trivialfis:
cc @hcho3 @RAMitchell

trivialfis:
Closing in favor of https://github.com/dmlc/xgboost/issues/6632 .

PR #6668, opened 2021-02-02
[breaking] Add prediction fucntion for DMatrix and use inplace predict for dask.

trivialfis:
This PR is about to wrap up https://github.com/dmlc/xgboost/pull/6638, only that `sklearn` is still using `DMatrix` for prediction by default instead of `inplace_predict`.

* Add a new API function for predicting on `DMatrix`.  This function aligns with rest of the `InplacePredictFrom*` functions on semantic of function arguments.
* Purge `ntree_limit` from libxgboost, use iteration instead.
* [dask] Use `inplace_predict` by default for dask sklearn models.
* [dask] Run prediction shape inference on worker instead of client.

The breaking change is in the Python sklearn `apply` function, I made it to be consistent with other prediction functions where `best_iteration` is used by default. Also, close https://github.com/dmlc/xgboost/issues/6553

### DIFF
--- a/include/xgboost/learner.h
+++ b/include/xgboost/learner.h
@@ -113,8 +113,8 @@ class Learner : public Model, public Configurable, public dmlc::Serializable {
    * \param data input data
    * \param output_margin whether to only predict margin value instead of transformed prediction
    * \param out_preds output vector that stores the prediction
-   * \param ntree_limit limit number of trees used for boosted tree
-   *   predictor, when it equals 0, this means we are using all the trees
+   * \param layer_begin Begining of boosted tree layer used for prediction.
+   * \param layer_end   End of booster layer. 0 means do not limit trees.
    * \param training Whether the prediction result is used for training
    * \param pred_leaf whether to only predict the leaf index of each tree in a boosted tree predictor
    * \param pred_contribs whether to only predict the feature contributions
@@ -124,7 +124,8 @@ class Learner : public Model, public Configurable, public dmlc::Serializable {
   virtual void Predict(std::shared_ptr<DMatrix> data,
                        bool output_margin,
                        HostDeviceVector<bst_float> *out_preds,
-                       unsigned ntree_limit = 0,
+                       unsigned layer_begin,
+                       unsigned layer_end,
                        bool training = false,
                        bool pred_leaf = false,
                        bool pred_contribs = false,
--- a/include/xgboost/learner.h
+++ b/include/xgboost/learner.h
@@ -113,7 +113,7 @@ class Learner : public Model, public Configurable, public dmlc::Serializable {
    * \param data input data
    * \param output_margin whether to only predict margin value instead of transformed prediction
    * \param out_preds output vector that stores the prediction
-   * \param layer_begin Begining of boosted tree layer used for prediction.
+   * \param layer_begin Beginning of boosted tree layer used for prediction.
    * \param layer_end   End of booster layer. 0 means do not limit trees.
    * \param training Whether the prediction result is used for training
    * \param pred_leaf whether to only predict the leaf index of each tree in a boosted tree predictor
@@ -141,7 +141,7 @@ class Learner : public Model, public Configurable, public dmlc::Serializable {
    * \param          type        Prediction type.
    * \param          missing     Missing value in the data.
    * \param [in,out] out_preds   Pointer to output prediction vector.
-   * \param          layer_begin Begining of boosted tree layer used for prediction.
+   * \param          layer_begin Beginning of boosted tree layer used for prediction.
    * \param          layer_end   End of booster layer. 0 means do not limit trees.
    */
   virtual void InplacePredict(dmlc::any const &x,
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -96,6 +96,24 @@ def from_cstr_to_pystr(data, length):
     return res
 
 
+def _convert_ntree_limit(booster, ntree_limit, iteration_range):
+    if ntree_limit is not None and ntree_limit != 0:
+        warnings.warn(
+            "ntree_limit is deprecated, use `iteration_range` or model "
+            "slicing instead.",
+            UserWarning
+        )
+        if iteration_range is not None and iteration_range[1] != 0:
+            raise ValueError(
+                "Only one of `iteration_range` and `ntree_limit` can be non zero."
+            )
+        num_parallel_tree, num_groups = _get_booster_layer_trees(booster)
+        num_parallel_tree = max([num_parallel_tree, 1])
+        num_groups = max([num_groups, 1])
+        iteration_range = (0, ntree_limit // num_parallel_tree)
+    return iteration_range
+
+
 def _expect(expectations, got):
     """Translate input error into string.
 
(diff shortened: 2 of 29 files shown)


trivialfis:
@RAMitchell @ShvetsKS @SmirnovEgorRu This is a big change, could you please help taking a look?

trivialfis:
To make inplace predict the default for sklearn, we need to have some more adapters like coo, csc.

trivialfis:
Something wrong with the docker build.  Please ignore it for review as the last test run was successful.

python-package/xgboost/core.py
```diff
@@ -1553,17 +1602,33 @@ def predict(self,
...
 
+        iteration_range :
+            Specifies which layer of trees are used in prediction.  For example, if a
+            random forest is trained with 100 rounds.  Specifying `iteration_range=(10,
+            20)`, then only the forests built during [10, 20) (half open set) rounds are
+            used in this prediction.
+
+            .. versionadded:: 1.4.0
+
+        strict_shape :
```
RAMitchell, line 1624: Can you provide a bit more context on why this parameter is needed? Keeping in mind that users can call ndarray.reshape() on the output if they require a strict shape.

src/c_api/c_api.cc
```diff
@@ -604,20 +604,58 @@ XGB_DLL int XGBoosterPredict(BoosterHandle handle,
...
-  learner->Predict(
-      *static_cast<std::shared_ptr<DMatrix>*>(dmat),
-      (option_mask & 1) != 0,
-      &entry.predictions, ntree_limit,
-      static_cast<bool>(training),
-      (option_mask & 2) != 0,
-      (option_mask & 4) != 0,
-      (option_mask & 8) != 0,
-      (option_mask & 16) != 0);
+  ntree_limit = GetLayerTrees(ntree_limit, learner);
```
RAMitchell, line 607: I'm confused about variable names - this is still ntree_limit, wouldn't this now be called layer_end?

RAMitchell (commented):
My understanding of this PR is that the predict c_api is being changed to use a json input, and ntreelimit is being removed in favour of number of iterations.

What happens in the JVM package? I guess it will stay on the old interface until someone updates it.

There are a lot of changes here but most of it is superficial, so it seems generally fine.

python-package/xgboost/core.py
trivialfis: Yup.  It's used to clarify the dimensions and make it consistent.  Just to give an example, so on top of mind, could you name the dimension of predict leaf?  Also how many dimensions of softmax should have with and without output margin?

src/c_api/c_api.cc
trivialfis: That's the old API, I don't want to break it for 1.4.   I will rename the variable.

trivialfis:
> My understanding of this PR is that the predict c_api is being changed to use a json input, and ntreelimit is being removed in favour of number of iterations.

That's right.  A lot of changes are devoted to removing the `ntree_limit` in the codebase.  So the PR looks ugly huge.

> What happens in the JVM package? I guess it will stay on the old interface until someone updates it.

It doesn't use ntree_limit so should be easy to update it.  What's I'm worrying about is the R package, I will make corresponding change, but may need some help from @hcho3 .

python-package/xgboost/core.py
trivialfis: At some point we might lower the calculation down into `Learner` class instead of C API.  Right now I just want to remove the burden from users.

trivialfis:
I'm quite excited to remove `ntree_limit`, it's troubling me for a while.

RAMitchell (approved):

RAMitchell:
@ShvetsKS can you take a look at this also? Thanks

ShvetsKS:
> @ShvetsKS @SmirnovEgorRu would you like to help creating these adapters?

@trivialfis Sure! Sorry for late response. Could you clarify the required time frame for finishing this work?

trivialfis:
> @trivialfis Sure! Sorry for late response. Could you clarify the required time frame for finishing this work?

No problem, I don't have an exact timeline for this, just part of improving the prediction function and memory usage long term roadmap.  I will be working on categorical data and dask for a while so take your time.

At the meanwhile, could you please help taking a look into this PR?  It should set up everything needed for future development.

wphicks (approved):
This generally looks great! I provided just a couple minor suggestions. The only overall comment that I'd make is that there are quite a lot of unrelated formatting changes in here that make this harder to review. No need to pull them out, but in the future it would probably be better to do a big formatting cleanup as a separate PR.

include/xgboost/learner.h
```diff
@@ -113,8 +113,8 @@ class Learner : public Model, public Configurable, public dmlc::Serializable {
    * \param data input data
    * \param output_margin whether to only predict margin value instead of transformed prediction
    * \param out_preds output vector that stores the prediction
-   * \param ntree_limit limit number of trees used for boosted tree
-   *   predictor, when it equals 0, this means we are using all the trees
+   * \param layer_begin Begining of boosted tree layer used for prediction.
```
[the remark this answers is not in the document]
trivialfis: Good catch.

trivialfis:
> No need to pull them out, but in the future it would probably be better to do a big formatting cleanup as a separate PR.

Thanks for the suggestion!  I tried to go through my PR again, I think some places look like formatting change, but they have actual change like parameter addition.  Feel free to point out where the change is not necessary.  @RAMitchell mentioned formatting issue in my previous PRs a few time. ;-)

python-package/xgboost/dask.py
```diff
@@ -1001,28 +995,22 @@ async def _direct_predict_impl(
...
-        # clear the state to avoid gpu_id, gpu_predictor
-        booster = Booster(model_file=booster.save_raw())
-        test_predt = booster.inplace_predict(test_sample, **kwargs)
-    else:
-        m = DMatrix(test_sample)
-        test_predt = booster.predict(m, **kwargs)
+        kwargs = kwargs.copy()
+        if kwargs["predict_type"] == "margin":
+            kwargs["output_margin"] = True
+        del kwargs["predict_type"]
```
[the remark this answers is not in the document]
trivialfis: @wphicks Done.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/6621 .

ShvetsKS:
@trivialfis is it expected that with current changes `PredictBatchByBlockOfRowsKernel` function is called even if subsampling is not used (equal to '1')?
It was discovered that during each training iterations `PredictBatchByBlockOfRowsKernel` is called now

trivialfis:
@ShvetsKS I'm not sure how much overhead it has, but I can make a special condition to avoid calling predict if the number of trees for prediction is 0.  Thanks for catching that!   Is there any other concern?

trivialfis:
@ShvetsKS Added a guard.

ShvetsKS:
> @ShvetsKS I'm not sure how much overhead it has, but I can make a special condition to avoid calling predict if the number of trees for prediction is 0. Thanks for catching that! Is there any other concern?

LGTM. No other concerns.

trivialfis:
@RAMitchell @wphicks @ShvetsKS Thanks for the review!

wphicks:
> Thanks for the suggestion! I tried to go through my PR again, I think some places look like formatting change, but they have actual change like parameter addition. Feel free to point out where the change is not necessary. @RAMitchell mentioned formatting issue in my previous PRs a few time. ;-)

Yep, like I said, not a big deal for this PR, and I'm glad it made it in already :tada:! Just for future reference, I was talking about things like [this](https://github.com/dmlc/xgboost/pull/6668/files#diff-b7920fe48897304cecd722fae109a55a22f6fc69b672ab6c81d5ce9d2c6c2556R118-R119). Those single-quotes didn't need to be converted to double-quotes. As a matter of fact, at a quick glance, I'm not sure that that file needed to be touched at all. Are you by chance using some sort of auto-formatter like black or something like that in your dev environment? Anyway, definitely nothing to sweat about :smile:! Just thought I'd mention it since leaving those out will help folks power through review much quicker in the future.

trivialfis:
@wphicks 

> hose single-quotes didn't need to be converted to double-quotes.

Ah ... you are right, that was black.  Will be more careful next time.

Issue #6649, opened 2021-01-27
Dask XGBoost hangs during training with multiple GPU workers


elaineejiang:
Hi, I am using XGBoost (v.1.1.1) with Dask (v. 2020.12.0). I have a Dask cluster that connects to remote GPU workers using k8s (v.1.14). I've noticed that if I train on multiple GPU workers, the dispatch-training tasks will hang on Rabit initialization. For example, this is what the call stack looks like in one of the workers:

```
Key: dispatched_train-60c6f36b-c5df-4871-9ff1-bf5fc547ccdf-0
File "[...]/ext/public/python/3/7/x/dist/lib/python3.7/threading.py", line 890, in _bootstrap self._bootstrap_inner()

File "[...]/ext/public/python/3/7/x/dist/lib/python3.7/threading.py", line 926, in _bootstrap_inner self.run()

File "[...]/ext/public/python/3/7/x/dist/lib/python3.7/threading.py", line 870, in run self._target(*self._args, **self._kwargs)

File "[...]/ext/public/python/distributed/2020/12/0/dist/lib/python3.7/distributed/threadpoolexecutor.py", line 55, in _worker task.run()

File "[...]/ext/public/python/distributed/2020/12/0/dist/lib/python3.7/distributed/_concurrent_futures_thread.py", line 65, in run result = self.fn(*self.args, **self.kwargs)

File "[...]/ext/public/python/distributed/2020/12/0/dist/lib/python3.7/distributed/worker.py", line 3425, in apply_function result = function(*args, **kwargs)

File "[...]/ext/public/python/xgboost/1/1/1/dist/lib/python3.7/xgboost/dask.py", line 418, in dispatched_train with RabitContext(rabit_args):

File "[...]/ext/public/python/xgboost/1/1/1/dist/lib/python3.7/xgboost/dask.py", line 82, in __enter__ rabit.init(self.args)

File "[...]/ext/public/python/xgboost/1/1/1/dist/lib/python3.7/xgboost/rabit.py", line 27, in init _LIB.RabitInit(len(arr), arr)
```
Here is a reproducible example:
```
# Prereq: Set up Dask cluster and client with >1 GPU workers

import pandas as pd
import numpy as np
import xgboost as xgb
import dask.dataframe as dd

seed = 42
random_state = np.random.RandomState(seed)

df = pd.DataFrame(random_state.random_sample((1000, 4)), columns=list(['x1', 'x2', 'x3', 'y']))
[...]

Issue #6604, opened 2021-01-14
hang in dask xgboost on CPU


pseudotensor:
rapids 0.14
master dmlc and tried 1.3.0
dask cluster with 2 nodes with normal dask_scheduler and dask_workers
2 ec2 nodes

Hi @trivialfis , how do I diagnose this problem?  I understand I'm not giving repro, but it's same kind of thing as usual we have discussed. Main question is logs from dask scheduler, dask workers, and xgboost show no problems, just hangs.

xgboost dask can work for several fits, but hangs at arbitrary times

xgboost stuck here
```
Current thread 0x0000148bdd769700 (most recent call first):
  File "/home/ubuntu/dai-1.9.1-linux-x86_64/python/lib/python3.6/threading.py", line 299 in wait
  File "/home/ubuntu/dai-1.9.1-linux-x86_64/python/lib/python3.6/threading.py", line 551 in wait
  File "/home/ubuntu/dai-1.9.1-linux-x86_64/cpu-only/lib/python3.6/site-packages/distributed/utils.py", line 336 in sync
  File "/home/ubuntu/dai-1.9.1-linux-x86_64/cpu-only/lib/python3.6/site-packages/distributed/client.py", line 832 in sync
  File "/home/ubuntu/dai-1.9.1-linux-x86_64/cpu-only/lib/python3.6/site-packages/xgboost/dask.py", line 1322 in fit
  File "/home/ubuntu/dai-1.9.1-linux-x86_64/cpu-only/lib/python3.6/site-packages/xgboost/core.py", line 422 in inner_f
```

Dask stuck here:
![image](https://user-images.githubusercontent.com/2249614/104538581-ffe9c100-55d0-11eb-81c9-d0f7e01c2e00.png)

I've tried playing with which IP is used, as ec2 has 2 versions, but both hit same problem

Seems kinda similar to problems with rabit that it uses wrong IP and then gets stuck, but no errors appear in this case.

Also, if I try to just re-use the dask scheduler/worker with a separate python interpreter use, that hangs too immediately, like original hang is blocking things.

FYI for my version of xgboost (1.4.0) the line mentioned above for dask.py 1322 is:

```
        return self.client.sync(self._fit_async,
                                X=X,
                                y=y,
                                sample_weight=sample_weight,
                                base_margin=base_margin,
                                eval_set=eval_set,
                                eval_metric=eval_metric,
                                sample_weight_eval_set=sample_weight_eval_set,
[...]

PR #6677, opened 2021-02-04
Fix Dask XGBoost hanging on rabit initialization during training with multi-GPU multi-nodes

elaineejiang:
This helped me resolve the issue in #6649. Please let me know if it makes sense.

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -818,6 +818,8 @@ async def _train_async(
 
         '''
         LOGGER.debug('Training on %s', str(worker_addr))
+        # Initialize rabit without workers first
+        rabit.init()
         worker = distributed.get_worker()
         with RabitContext(rabit_args), config.config_context(**global_config):
             local_dtrain = _dmatrix_from_list_of_parts(**dtrain_ref)
@@ -839,6 +841,11 @@ async def _train_async(
                     LOGGER.info(msg)
                 else:
                     local_param[p] = worker.nthreads
+            
+            # If worker did not receive input data, return without failing
+            if local_dtrain.num_row() == 0:
+                return None
+
             bst = worker_train(params=local_param,
                                dtrain=local_dtrain,
                                num_boost_round=num_boost_round,


trivialfis:
Did you figure why didn't the worker receive data?

python-package/xgboost/dask.py
```diff
@@ -818,6 +818,8 @@ def dispatched_train(
 
         '''
         LOGGER.debug('Training on %s', str(worker_addr))
+        # Initialize rabit without workers first
+        rabit.init()
```
trivialfis, line 822: > Please let me know if it makes sense.

I don't quite understand why does this help.

python-package/xgboost/dask.py
```diff
@@ -839,6 +841,11 @@ def dispatched_train(
                     LOGGER.info(msg)
                 else:
                     local_param[p] = worker.nthreads
+            
+            # If worker did not receive input data, return without failing
+            if local_dtrain.num_row() == 0:
```
trivialfis, line 846: You should let the train continue, otherwise the cluster can hang on waiting for this exited worker for synchronization.

trivialfis (changes requested):

python-package/xgboost/dask.py
elaineejiang: If I let training continue, I get this error `Check failed: mparam_.num_feature != 0 (0 vs. 0) : 0 feature is supplied. Are you using raw Booster interface?` in `learner.cc` if the worker did not receive input data.

trivialfis: @elaineejiang  Thanks for the explanation.  In that case, the synchronization between DMatrix is failing.  During the construction, DMatrix on workers synchronize the number of columns in input data, see https://github.com/dmlc/xgboost/blob/72892cc80d2b656d0d911381ee29a32904308fb4/src/data/simple_dmatrix.cc#L159 .  So as long as one of the worker has any data, it shouldn't be 0.

trivialfis:
To me the issue you encountered seems to be a network issue.  That's quite problematic as each cluster has its own quirks, I don't know what exactly is happening on your cluster.  If I were you I would start with trying to figure out why the connection is refused.

python-package/xgboost/dask.py
elaineejiang: I see, thanks @trivialfis. Just to be clear, I'm only seeing this error when I try to run multiple XGBoost learners in parallel via
```
dask_client.submit(xgb.dask.DaskXGBRegressor([...]).fit(dd, xcols, ycol, wcol)
```
Is this something that's not recommend or not supported?

elaineejiang:
> To me the issue you encountered seems to be a network issue. That's quite problematic as each cluster has its own quirks, I don't know what exactly is happening on your cluster. If I were you I would start with trying to figure out why the connection is refused.

Agreed that this is likely to do with my specific cluster configuration. I have seen similar issues posted (#6604) that haven't been closed yet and was wondering if the initial call to `rabit.init()` can help resolve those issues.

python-package/xgboost/dask.py
trivialfis: I think the correct version should be something similar to:

``` python
classifier_future = client.submit(classifier.fit, X, y, sample_weight=w, eval_set=[(X, y)])
```

This should work.

trivialfis:
> initial call to rabit.init() can help resolve those issues.

It shouldn't.  `init` is supposed to receive all worker information.

python-package/xgboost/dask.py
elaineejiang: Ah so sorry -- I typed that out on the fly and got the syntax wrong. I am seeing
```
Check failed: mparam_.num_feature != 0 (0 vs. 0) : 0 feature is supplied. Are you using raw Booster interface?
```
with what your version.

If that's _supposed_ to work, I guess it's just my network issue?

trivialfis:
I think I can reproduce the issue now with setting `pure` to False.  Will investigate.  So far it's hanged in Python thread mutex, not sure where's applied.

trivialfis:
Em .. my error seems to be different.   dask is wrongly identifying data being created by other client.  It doesn't apply here.

elaineejiang:
I see. I saw in the GPU tests (https://github.com/dmlc/xgboost/blob/master/tests/distributed/distributed_gpu.py) to call rabit initializer:
```
# Always call this before using distributed module
xgb.rabit.init()
```
which is why I decided to add it just to see what would happen. However, I don't know why calling `rabit.init()` without worker args prevents the subsequent call `rabit.init(args)` from hanging. But it's the only thing that's worked on my cluster.

trivialfis:
> But it's the only thing that's worked on my cluster.

It might not be working correctly I think.  A better place to look at tests would be `tests/python-gpu/test_gpu_with_dask.py` and `tests/python/test_with_dask.py`.

I'm looking into parallel model training, might take some time as I vaguely recall there's a similar issue on Spark package.  For now please train 1 model at a time.

elaineejiang:
@trivialfis sounds good, thanks for all the help debugging!

Here's a summary for anyone who might be following. The two issues are:

**1.  Rabit initialization hangs during training on multi-node, multi-GPU**

- My workaround was to add `rabit.init()`, and I was able to see the logs such as:
```
[02/04 19:01:13.564 INFO [xgboost.dask] dask.py:414] Training on tcp://10.101.37.135:30030
[02/04 19:01:13.862 INFO [xgboost.dask] dask.py:414] Training on tcp://10.98.13.197:30030
[02/04 19:01:13.973 INFO [xgboost.dask] dask.py:414] Training on tcp://10.103.98.134:30030
[02/04 19:01:19.461 INFO [xgboost.dask] dask.py:414] Training on tcp://10.99.9.197:30030
```
- I also checked that the GPUs were being utilized
- @trivialfis mentioned it still might not be working properly, so I'll test some more. I also can't explain why calling `rabit.init()` would make a difference, so I'll continue to investigate.

**2.  Synchronization between DMatrix is failing, resulting in `Check failed: mparam_.num_feature != 0 (0 vs. 0) : 0 feature is supplied. Are you using raw Booster interface?`.**
- This only occurs when I'm training multiple models in parallel via `classifier_future = client.submit(classifier.fit, X, y, sample_weight=w, eval_set=[(X, y)])`

trivialfis:
If you are willing to patch xgboost, here is something you can try:

Change the `fit` method of `DaskXGBClassifier`/`DaskXGBRegressor`:

into:
```
        _assert_dask_support()
        args = {k: v for k, v in locals().items() if k != 'self'}
        if self._client is None:
            try:
                with distributed.worker_client() as client:
                    self.client = client
                    return self.client.sync(self._fit_async, **args)
            except ValueError:
                pass
        return self.client.sync(self._fit_async, **args)
```

trivialfis:
Or equivalently you can set the client and launch training on a local function:
```
def my_training_func(cls, X, y):
    with worker_client() as client:
        cls.client = client
        return cls.fit(X, y)

client.submit(my_training_func, X, y)
```

elaineejiang:
I've tried doing the latter example (getting worker client before launching) but I still hit the `0 feature` error.

trivialfis:
Thanks for the reply.  I think for now one will have to train one model at a time.  xgboost relies on MPI like communication framework, if one of the worker is scheduled behind the others it will hang.

elaineejiang:
Thanks @trivialfis for looking into this. Would it be possible to add this request to the roadmap?

trivialfis:
I will put it in there, but not promising it will deliver.  Seems to be a lot of work.

trivialfis:
I will put it in there, but not promising it will deliver.  Seems to be a lot of work.

elaineejiang:
Thanks @trivialfis! Is there an expected release deadline for 1.4.0?

PR #6693, opened 2021-02-08
Use `Predictor` for `dart`.

trivialfis:
* Use normal predictor for dart booster.
* Implement `inplace_predict` for dart.
* Enable `dart` for dask interface now that it's thread-safe.
* categorical data should be working out of box for dart now, but I will do some extra testing later.

The implementation is not very efficient as it has to pull back the data and apply weight for each tree, but still a significant improvement over previous implementation as now we no longer binary search for each sample.

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -754,10 +754,7 @@ class XGBModel(XGBModelBase):
         # Inplace predict doesn't handle as many data types as DMatrix, but it's
         # sufficient for dask interface where input is simpiler.
         params = self.get_params()
-        booster = self.booster
-        if params.get("predictor", None) is None and (
-            booster is None or booster == "gbtree"
-        ):
+        if params.get("predictor", None) is None and self.booster != "gblinear":
             return True
         return False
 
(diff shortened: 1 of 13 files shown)


python-package/xgboost/sklearn.py
```diff
@@ -756,7 +756,7 @@ def _can_use_inplace_predict(self) -> bool:
         params = self.get_params()
         booster = self.booster
         if params.get("predictor", None) is None and (
-            booster is None or booster == "gbtree"
+            booster is None or booster != "gblinear"
```
trivialfis, line 759: Remove "is none"

src/gbm/gbtree.cc
```diff
@@ -765,32 +791,13 @@ class Dart : public GBTree {
...
-  }
-
   // select which trees to drop
   // passing clear=True will clear selection
   inline void DropTrees(bool is_training) {
-    idx_drop_.clear();
     if (!is_training) {
       return;
     }
+    idx_drop_.clear();
```
trivialfis, line 800: Add a comment that this function should be thread-safe when it's not training.

RAMitchell (approved):
LGTM

PR #6648, opened 2021-01-27
Make prediction thread safe.

trivialfis:
As https://github.com/dmlc/xgboost/pull/6645#issuecomment-768359052 fixed a performance issue, the booster object in dask are no longer copied.  So thread safety in predict must be guaranteed.

PR #6696, opened 2021-02-09
Fix perf gap in thread safe prediction

ShvetsKS:
As in https://github.com/dmlc/xgboost/pull/6648 local `RegTree::FVec` storage was introduced to preserve thread safety each training iteration requires buffer initialization in case of subsampling. 

stages (santander dataset) | before https://github.com/dmlc/xgboost/pull/6648 | thread safe https://github.com/dmlc/xgboost/pull/6648 and master | current PR
--| -- | -- | --
full training |136s | 165s | 140s
PredictRaw | 55s | 78s | 59s

this PR introduces threading for local buffers initialization.
But as still some gaps are exist is it possible to have not thread safe `PredictDMatrix` call during training?

### DIFF
--- a/src/predictor/cpu_predictor.cc
+++ b/src/predictor/cpu_predictor.cc
@@ -65,8 +65,8 @@ inline void PredictByAllTrees(gbm::GBTreeModel const &model, const size_t tree_b
 }
 
 template <typename DataView>
-void FVecFill(const size_t block_size, const size_t batch_offset, DataView* batch,
-              const size_t fvec_offset, std::vector<RegTree::FVec>* p_feats, int num_feature) {
+void FVecFill(const size_t block_size, const size_t batch_offset, const int num_feature,
+              DataView* batch, const size_t fvec_offset, std::vector<RegTree::FVec>* p_feats) {
   for (size_t i = 0; i < block_size; ++i) {
     RegTree::FVec &feats = (*p_feats)[fvec_offset + i];
     if (feats.Size() == 0) {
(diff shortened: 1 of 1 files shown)


trivialfis:
I thought we have prediction cache enabled?

ShvetsKS:
> I thought we have prediction cache enabled?

For subsampling case we still don't have prediction caching (seems it's still in progress: https://github.com/dmlc/xgboost/pull/6683, major complexity is related to multiclass classification case as for each group own indices subset is generated.)

trivialfis:
@ShvetsKS Would you like to take a look into the GPU implementation of subsampling?  Or the WIP rewrite for CPU hist?  The subsampling doesn't have to conflict with cache.

trivialfis:
Last time your found it to be slower than master branch, which is expected as I don't have time to incooperate many recent optimization into it.  But I believe my rewrite can at least offer some insight on how to implement some of the features in a different way.

ShvetsKS:
> Last time your found it to be slower than master branch, which is expected as I don't have time to incooperate many recent optimization into it. But I believe my rewrite can at least offer some insight on how to implement some of the features in a different way.

Current changes are applicable even for inference stage as it's better to initialize `RegTree::FVec` thread locally anyway.

ShvetsKS:
problem with continuous-integration/travis-ci/pr  is related with server node
`await ServerNode.close(self)`
seems restarting should help.

trivialfis:
Sorry for the long wait.  Still on vacation, but will try to look into it soon as possible.

trivialfis (approved):

PR #6699, opened 2021-02-10
[dask] remove outdated comment

jameslamb:
I was reading through the Dask module today and noticed a comment that I think is outdated. It mentions CV and Ranking as TODOs, but both of these features exist in `xgboost.dask`.

This PR proposes removing that comment.

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -69,13 +69,6 @@ try:
 except ImportError:
     TrainReturnT = Dict[str, Any]  # type:ignore
 
-# Current status is considered as initial support, many features are not properly
-# supported yet.
-#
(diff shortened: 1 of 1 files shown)


trivialfis:
CV is not implemented yet.  We are hoping that we can use some 3 party libraries to approach HPO.

jameslamb:
oh sorry! I should have checked more carefully. I still think the rest of that comment no longer applies and can be removed. But feel free to close this if you disagree, of course.

trivialfis:
Thanks for the PR.  I'm currently on vacation so might be slow.

jameslamb:
This is VERY not urgent, don't worry. Enjoy your vacation!

PR #6705, opened 2021-02-13
[python-package] Fix class Booster: feature_types = None

Roffild:
```
  in get_dump
    if self.feature_types is None:
AttributeError: 'Booster' object has no attribute 'feature_types'
```


trivialfis (approved):

trivialfis:
Thanks for the fix!

Issue #6712, opened 2021-02-17
[dask] Use client when persisting collections


jose-moralez:
Hi. I'm trying to fit xgboost models concurrently using asynchronous dask clients, however using the collection's persist method like [here](https://github.com/dmlc/xgboost/blob/9f15b9e322ac505189394ff840bda3f52a7abb1f/python-package/xgboost/dask.py#L306) triggers the computation on whatever is the default client at the moment, so code like this:

```python
async def train(params):
  async with Client(asynchronous=True) as client:
      # create X_train, X_valid, y_train, y_valid as dask collections
      # ...
      dtrain = await xgb.dask.DaskDMatrix(client, X_train, y_train)
      dvalid = await xgb.dask.DaskDMatrix(client, X_valid, y_valid)
```
won't neccesarily create the collections in its corresponding `client` and will trigger errors like these:
```
  File "./api_xgb.py", line 71, in train                                                                                                        
    dvalid = await xgb.dask.DaskDMatrix(client, X_valid, y_valid)                                                                                 
  File "/home/jose_morales/miniconda3/envs/fastapi/lib/python3.8/site-packages/xgboost/dask.py", line 257, in map_local_data                      
    data = data.persist()                                                                                                                         
  File "/home/jose_morales/miniconda3/envs/fastapi/lib/python3.8/site-packages/dask/base.py", line 254, in persist
    (result,) = persist(self, traverse=False, **kwargs)
  File "/home/jose_morales/miniconda3/envs/fastapi/lib/python3.8/site-packages/dask/base.py", line 755, in persist
    results = client.persist(
  File "/home/jose_morales/miniconda3/envs/fastapi/lib/python3.8/site-packages/distributed/client.py", line 2944, in persist
    futures = self._graph_to_futures(
  File "/home/jose_morales/miniconda3/envs/fastapi/lib/python3.8/site-packages/distributed/client.py", line 2543, in _graph_to_futures
    dsk = highlevelgraph_pack(dsk, self, keyset)
  File "/home/jose_morales/miniconda3/envs/fastapi/lib/python3.8/site-packages/distributed/protocol/highlevelgraph.py", line 115, in highlevelgrap
h_pack
    "state": _materialized_layer_pack(
    "state": _materialized_layer_pack(
  File "/home/jose_morales/miniconda3/envs/fastapi/lib/python3.8/site-packages/distributed/protocol/highlevelgraph.py", line 40, in _materialized_
layer_pack
    raise ValueError(
[...]

PR #6722, opened 2021-02-22
[dask] Use client to persist collections

jose-moralez:
This attempts to solve #6712 by using the `client` object when persisting collections. This ensures that the futures are computed by the `client` passed to `xgb.dask.DaskDMatrix` and allows for several trainings to happen concurrently on different clusters.

### DIFF
--- a/tests/python/test_with_dask.py
+++ b/tests/python/test_with_dask.py
@@ -652,7 +652,7 @@ async def generate_concurrent_trainings() -> None:
                                 asynchronous=True,
                                 dashboard_address=0) as cluster:
             async with Client(cluster, asynchronous=True) as client:
-                X, y, w = generate_array()
+                X, y, w = generate_array(with_weights=True)
                 dtrain = await DaskDMatrix(client, X, y, weight=w)
                 dvalid = await DaskDMatrix(client, X, y, weight=w)
                 output = await xgb.dask.train(client, {}, dtrain=dtrain)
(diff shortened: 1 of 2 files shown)


jose-moralez:
Hi @trivialfis, here's my proposal for solving #6712. I ran the included test with the current master branch and was able to reproduce the `ValueError: Inputs contain futures that were created by another client.` error. Is including `weights` enough to test that `meta` is being computed correctly as well? Looking forward to your thoughts on this.

tests/python/test_with_dask.py
```diff
@@ -647,6 +645,25 @@ def test_with_asyncio() -> None:
...
 
 
+async def generate_concurrent_trainings() -> None:
+    async def train():
+        async with LocalCluster(n_workers=2,
+                                threads_per_worker=1,
+                                asynchronous=True,
+                                dashboard_address=0) as cluster:
+            async with Client(cluster, asynchronous=True) as client:
+                X, y, w = generate_array()
```
trivialfis, line 655: with_weights = True.

trivialfis (commented):
Thanks for the fix!  A small error in test.

https://github.com/dmlc/xgboost/pull/6726 Will fix the CI issue.

jose-moralez:
Thanks for the fix. Should I wait for #6726 to get merged to update this?

trivialfis:
> Should I wait for #6726 to get merged to update this?

Yup.  Waiting for review.

PR #6728, opened 2021-02-24
Removing the wrong gamma-deviance metric

Roffild:
Corrected version, but without checking the condition `label>0 && pred>0`:
```c++
struct EvalGammaDeviance {
  const char *Name() const {
    return "gamma-deviance";
  }

  XGBOOST_DEVICE bst_float EvalRow(bst_float label, bst_float pred) const {
    const bst_float eps = 1e-16f;
    return 2 * (std::log(pred/(label+eps)) + label/(pred+eps) - 1);
  }
  static bst_float GetFinal(bst_float esum, bst_float wsum) {
    return wsum == 0 ? esum : esum / wsum;
  }
};
```

### DIFF
--- a/src/metric/elementwise_metric.cu
+++ b/src/metric/elementwise_metric.cu
@@ -270,21 +270,6 @@ struct EvalPoissonNegLogLik {
   }
 };
 
-struct EvalGammaDeviance {
-  const char *Name() const {
-    return "gamma-deviance";
-  }
-
-  XGBOOST_DEVICE bst_float EvalRow(bst_float label, bst_float pred) const {
-    bst_float epsilon = 1.0e-9;
-    bst_float tmp = label / (pred + epsilon);
-    return tmp - std::log(tmp) - 1;
-  }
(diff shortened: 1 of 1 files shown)


Roffild (commented):
Approved

trivialfis (changes requested):
Good catch!  But could you please replace it with the correct one instead?  The metrics defined in xgboost have many use like distributed setting, language bindings other than Python.  Also 3 party implementation cannot be saved.

Roffild:
This is part of Tweedie. And everything about Tweedie needs to be redone. I don't understand what formulas were used for Tweedie in XGBoost.

I am guided by these sources:
https://en.wikipedia.org/wiki/Tweedie_distribution#The_Tweedie_deviance

[sklearn/_loss/glm_distribution.py](https://github.com/scikit-learn/scikit-learn/blob/95119c13af77c76e150b753485c662b7c52a41a2/sklearn/_loss/glm_distribution.py#L178)

Issue #3900, opened 2018-11-15
Node not parsed correctly by xgb.model.dt.tree if split criteria is "<Inf"


dshopin:
Hi,

I noticed that in the table produced by `xgb.model.dt.tree` some nodes have NA in the "Feature" column, e. g.:
```
Tree    Node   ID     Feature   Split  Yes      No       Missing    Quality  Cover
0         55      0-55    <NA>    NA     0-NA  0-NA    0-NA      NA         NA
```
Afer looking through the code of the function, I think the issue is due to a regex that parses the text dump. Here's the raw text:

`55:[f279<inf] yes=83,no=84,missing=84,gain=6.56160975,cover=16.25`

The regex is:
```
anynumber_regex <- "[-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?"
branch_rx <- paste0("f(\\d+)<(", anynumber_regex, ")\\] yes=(\\d+),no=(\\d+),missing=(\\d+),", 
                    "gain=(", anynumber_regex, "),cover=(", anynumber_regex, 
                    ")")

```
So, it doesn't capture sub-strings where split criteria is not a number.

dshopin:
And it's the same when split criteria ">NaN", e. g.:

`"206:[f333<nan] yes=227,no=228,missing=228,gain=0.739670038,cover=4.5880127"`

Guessing, this is the reason for the issue https://github.com/dmlc/xgboost/issues/2387 as well.

I currently fixed it in my code by changing `anynumber_regex` to:

`anynumber_regex <- "[-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?|[+-]?[Ii]nf|[nN]a[nN]"`

hcho3:
@dshopin I wonder how `inf` and `nan` appear in splits. As far as I know, XGBoost should only produce real-valued splits.

zhangjinrui2718:
> Hi,
> 
> I noticed that in the table produced by `xgb.model.dt.tree` some nodes have NA in the "Feature" column, e. g.:
> 
> ```
> Tree    Node   ID     Feature   Split  Yes      No       Missing    Quality  Cover
> 0         55      0-55    <NA>    NA     0-NA  0-NA    0-NA      NA         NA
> ```
> 
> Afer looking through the code of the function, I think the issue is due to a regex that parses the text dump. Here's the raw text:
> 
[...]

PR #6740, opened 2021-03-02
[R] Parse splits on inf in xgb.model.dt.tree (#3900)

thatchersj:
As per the discussion on (#3900) splits on `inf` were previously being incorrectly parsed with `NA`s due to a failed regex match.

Since the change (#6109 Sep 2020) to remove `stringi` dependency, the handling of failed regex matches has changed and can now cause a number of different errors (detailed below).

I have added `inf` to the regex and added a simple test case that fails for previous versions.

## Problem

If you have `inf` splits you now get one of three undesirable behaviours detailed below using dummy tree dumps.

Note: This code was run on Windows 10 using xgboost 1.3.2.1 and R 4.0.3.

#### 1. If you have multiple non-inf splits you get a data.table error
This seems to be the most common case, and was the error I stumbled upon that led me here.
```
xgb.model.dt.tree(
  text = c(
    "booster[0]",
    "0:[f1<inf] yes=1,no=3,missing=3,gain=0.1,cover=3",
    "1:[f2<3] yes=2,no=3,missing=3,gain=0.1,cover=1",
    "2:[f1<2] yes=4,no=3,missing=3,gain=0.3,cover=4",
    "3:leaf=0.2,cover=1",
    "4:leaf=0.5,cover=1"
  )
)

# Error in `[.data.table`(td, isLeaf == FALSE, `:=`((branch_cols), { : 
#  Supplied 2 items to be assigned to 3 items of column 'Feature'. If you wish to 'recycle'
#  the RHS please use rep() to make this intent clear to readers of your code.
```

The following cases are both improbable in practice but included for reference, **3** being particularly misleading

#### 2. If you only have only inf splits, you get a subscript out of bounds error
```
xgb.model.dt.tree(
  text = c(
    "booster[0]",
    "0:[f1<inf] yes=1,no=2,missing=2,gain=0.5,cover=4",
    "2:leaf=0.2,cover=1"
  )
)

# Error in do.call(rbind, matches)[, c(2, 3, 5, 6, 7, 8, 10), drop = FALSE] : 
#   subscript out of bounds
```

#### 3. If you only have only 1 non-inf split, then it's details get copied onto all rows
This is the special case of recycling that data.table allows (avoiding the error in **1**)
```
xgb.model.dt.tree(
  text = c(
    "booster[0]",
    "0:[f1<inf] yes=1,no=3,missing=3,gain=0.1,cover=3",
    "1:[f2<3] yes=2,no=3,missing=3,gain=0.5,cover=1",
    "2:leaf=0.2,cover=1",
    "3:leaf=0.5,cover=1"
  )
)

#    Tree Node  ID Feature Split  Yes   No Missing Quality Cover
# 1:    0    0 0-0       2     3  0-2  0-3     0-3     0.5     1
# 2:    0    1 0-1       2     3  0-2  0-3     0-3     0.5     1
# 3:    0    2 0-2    Leaf    NA <NA> <NA>    <NA>     0.2     1
# 4:    0    3 0-3    Leaf    NA <NA> <NA>    <NA>     0.5     1
```

## Cause

As discussed above, the nodes are not parsed correctly as `anynumber_regex` does not match `inf`.

The code change on lines `121-125` of `R-package/R/xgb.model.dt.tree.R` uses 
```
(A)  matches <- regmatches(t, regexec(branch_rx, t))
     #skip some indices with spurious capture groups from anynumber_regex
     xtr <- do.call(rbind, matches)[, c(2, 3, 5, 6, 7, 8, 10), drop = FALSE]
```
to replace the old line
```
(B)  xtr <- stri_match_first_regex(t, branch_rx)[, c(2, 3, 5, 6, 7, 8, 10), drop = FALSE]
```
This code change has altered the behaviour when the regex fails to find a match, for example
```
txt = c(
  "booster[0]",
  "0:[f1<inf] yes=1,no=3,missing=3,gain=0.1,cover=3",
  "1:[f2<3] yes=2,no=3,missing=3,gain=0.5,cover=1",
  "2:leaf=0.2,cover=1",
  "3:leaf=0.5,cover=1"
)

anynumber_regex <- "[-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?"
branch_rx <- paste0(
  "f(\\d+)<(", anynumber_regex, ")\\] yes=(\\d+),no=(\\d+),missing=(\\d+),",
  "gain=(", anynumber_regex, "),cover=(", anynumber_regex, ")"
)

(A) do.call(rbind, regmatches(txt[2:3], regexec(branch_rx, txt[2:3])))[, c(2, 3, 5, 6, 7, 8, 10), drop = FALSE]

#      [,1] [,2] [,3] [,4] [,5] [,6]  [,7]
# [1,] "2"  "3"  "2"  "3"  "3"  "0.5" "1" 

(B) stringi::stri_match_first_regex(txt[2:3], branch_rx)[, c(2, 3, 5, 6, 7, 8, 10), drop = FALSE]

#      [,1] [,2] [,3] [,4] [,5] [,6]  [,7]
# [1,] NA   NA   NA   NA   NA   NA    NA  
# [2,] "2"  "3"  "2"  "3"  "3"  "0.5" "1" 

```

## Solution 

As suggested by @dshopin and seconded by @hcho3 in #3900 I have changed the `anynumber_regex` to include `inf`:
```
anynumber_regex <- "[-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?|[-+]?[Ii]nf"
```

### DIFF
--- a/R-package/R/xgb.model.dt.tree.R
+++ b/R-package/R/xgb.model.dt.tree.R
@@ -95,7 +95,7 @@ xgb.model.dt.tree <- function(feature_names = NULL, model = NULL, text = NULL,
 
   add.tree.id <- function(node, tree) if (use_int_id) node else paste(tree, node, sep = "-")
 
-  anynumber_regex <- "[-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?"
+  anynumber_regex <- "[-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?|[-+]?[Ii]nf"
 
   td <- data.table(t = text)
   td[position, Tree := 1L]
--- a/R-package/R/xgb.model.dt.tree.R
+++ b/R-package/R/xgb.model.dt.tree.R
@@ -95,7 +95,7 @@ xgb.model.dt.tree <- function(feature_names = NULL, model = NULL, text = NULL,
 
   add.tree.id <- function(node, tree) if (use_int_id) node else paste(tree, node, sep = "-")
 
-  anynumber_regex <- "[-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?|[-+]?[Ii]nf"
+  anynumber_regex <- "[-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?|[-+]?[Ii]nf|[nN]a[nN]"
 
   td <- data.table(t = text)
   td[position, Tree := 1L]
--- a/R-package/tests/testthat/test_helpers.R
+++ b/R-package/tests/testthat/test_helpers.R
@@ -281,6 +281,18 @@ test_that("xgb.model.dt.tree throws error for gblinear", {
   expect_error(xgb.model.dt.tree(model = bst.GLM))
 })
 
+test_that("xgb.model.dt.tree works with splits on inf", {
+  tree.dump <- c(
+    "booster[0]",
+    "0:[f1<inf] yes=1,no=3,missing=3,gain=0.1,cover=3",
+    "1:[f2<-inf] yes=2,no=3,missing=3,gain=0.1,cover=1",
+    "2:[f1<2] yes=4,no=3,missing=3,gain=0.3,cover=4",
+    "3:leaf=0.2,cover=1",
+    "4:leaf=0.5,cover=1"
+  )
+  expect_equal(xgb.model.dt.tree(text = tree.dump)$Split, c(Inf, -Inf, 2, NA, NA))
+})
+
 test_that("xgb.importance works with and without feature names", {
   importance.Tree <- xgb.importance(feature_names = feature.names, model = bst.Tree)
   if (!flag_32bit)
--- a/R-package/tests/testthat/test_helpers.R
+++ b/R-package/tests/testthat/test_helpers.R
@@ -293,6 +293,18 @@ test_that("xgb.model.dt.tree works with splits on inf", {
   expect_equal(xgb.model.dt.tree(text = tree.dump)$Split, c(Inf, -Inf, 2, NA, NA))
 })
 
+test_that("xgb.model.dt.tree works with splits on nan", {
+  tree.dump <- c(
+    "booster[0]",
+    "0:[f1<nan] yes=1,no=3,missing=3,gain=0.1,cover=3",
+    "1:[f2<nan] yes=2,no=3,missing=3,gain=0.1,cover=1",
+    "2:[f1<2] yes=4,no=3,missing=3,gain=0.3,cover=4",
+    "3:leaf=0.2,cover=1",
+    "4:leaf=0.5,cover=1"
+  )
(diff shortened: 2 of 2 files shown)


trivialfis:
Thanks for the PR and detailed description, could you please share a reproducible example that produces inf split?

thatchersj:
> Thanks for the PR and detailed description, could you please share a reproducible example that produces inf split?

Short answer may be no. I was surprised to see it and couldn't see anything obvious in the data to have caused it. Given the discussions on #2914 I'd have thought it wasn't possible if I hadn't seen it!

I came across the error on an old client production environment (running R 3.5.2) and have since rolled back the xgboost version to prevent it happening again for now. 
Unfortunately this means I no longer have access to the exact environment that created the model with the inf splits, and my attempts to recreate it on my own dev environment (R 4.0.3) haven't produced any inf splits. Not sure if the R version is likely to be a factor, but I will have a go at recreating the environment locally at some point today and have another dig into the data to see if I can create a minimal reprex. Could be a while as the dataset has >1m rows and >100 features.

trivialfis:
Em, so data contains `inf` but xgboost doesn't throw an error.

thatchersj:
> Em, so data contains `inf` but xgboost doesn't throw an error.

Yes, although the handling of `inf` is weirdly inconsistent. I'll raise a new issue for that

trivialfis:
Hi, I opened a different PR for checking invalid data: https://github.com/dmlc/xgboost/pull/6742 .

trivialfis:
Hi, could you please try latest master branch and see if the `inf` split is still reproducible?  Right now the `DMatrix` should throw an error when data contains `inf` but `missing` is set to other value.

thatchersj:
Sorry, I'm unable to build the package from source so can't test this, however I do disagree with the change made #6742.
I don't see why Inf should be an invalid value for decision tree regression. There seems to be a perfectly reasonable notion of splitting at Inf, and there is well-defined comparison between Inf and any real number. Moreover, this behaviour was previously available in xgboost (see the last example in my ticket you have closed #6741) and #6742 breaks/backs out this functionality.

trivialfis:
I see your point.  Yes, it's possible for a decision tree to split on `inf` as a trivial case, but right now we don't have uniformed handling of `inf` in various tree building algorithm.  I will see if it make sense to revert that commit.

thatchersj:
> I see your point.  Yes, it's possible for a decision tree to split on `inf` as a trivial case, but right now we don't have uniformed handling of `inf` in various tree building algorithm.  I will see if it make sense to revert that commit.

That makes sense, thanks!

trivialfis:
We @RAMitchell @hcho3 talked about the issue with data containing `inf` and using `inf` as split offline.  So these are 2 separate issues.

For the first one, we believe xgboost doesn't need to rush into supporting it right now since we have a `missing` parameter in `DMatrix` for specifying this kind of data, also users can handle it by preprocessing.  Providing full support for `inf` requires careful inspection into various language bindings and internal algorithms.

For the second issue, in the future, it's possible that xgboost can generate `inf` in split for trivial split value, but that would be a different topic.

chyunsu3 (approved):

PR #6743, opened 2021-03-05
[dask] Use `distributed.MultiLock`

trivialfis:
This enables training multiple models in parallel.

* Conditionally import `MultiLock`.
* Use async train directly in scikit learn interface.
* Use `worker_client` when available.

Close https://github.com/dmlc/xgboost/issues/6649 Close https://github.com/dmlc/xgboost/pull/6677 .

Currently depends on https://github.com/dask/distributed/pull/4503 .

### DIFF
--- a/tests/python/test_with_dask.py
+++ b/tests/python/test_with_dask.py
@@ -1286,10 +1286,13 @@ def test_dask_unsupported_features(client: "Client") -> None:
         )
 
 
-@pytest.mark.skipimport("distributed.MultiLock")
 def test_parallel_submits(client: "Client") -> None:
     """Test for running multiple train simultaneously from single clients."""
-    from distributed import MultiLock
+    try:
+        from distributed import MultiLock  # NOQA
+    except ImportError:
+        pytest.skip("`distributed.MultiLock' is not available")
+
     from sklearn.datasets import load_digits
 
     futures = []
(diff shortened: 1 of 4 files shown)


trivialfis:
@hcho3 This is the flaky test I mentioned: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-6743/7/pipeline .

tests/python/test_with_dask.py
```diff
@@ -1285,6 +1286,82 @@ def test_dask_unsupported_features(client: "Client") -> None:
         )
 
 
+def test_parallel_submits(client: "Client") -> None:
+    """Test for running multiple train simultaneously from single clients."""
+    try:
```
RAMitchell, line 1291: https://docs.pytest.org/en/stable/reference.html#pytest-importorskip

RAMitchell (approved):
LGTM, do you want to add some warning about this behaviour on the docs?

tests/python/test_with_dask.py
trivialfis: Not sure if that's usable with sub-module.  Let me try.

trivialfis: Tried it, that doesn't work with from import syntax:
``` python
pytest.mark.skipimport("distributed.MultiLock")
```

PR #6747, opened 2021-03-11
Re-implement ROC-AUC.

trivialfis:
- Binary
- MultiClass
- LTR
- Add documents.

This PR resolves a few issues:
  - Define a value when the dataset is invalid, which can happen if there's an empty dataset, or when the dataset contains only positive or negative values.
  - Define ROC-AUC for multi-class classification.
  - Define weighted average value for distributed setting.
  - A correct implementation for learning to rank task.  The previous implementation is just binary classification with averaging across groups, which doesn't measure ordered learning to rank.


doc/parameter.rst
```diff
@@ -399,7 +399,13 @@ Specify the learning task and the corresponding learning objective. The objectiv
...
     - ``mlogloss``: `Multiclass logloss <http://scikit-learn.org/stable/modules/generated/sklearn.metrics.log_loss.html>`_.
-    - ``auc``: `Area under the curve <http://en.wikipedia.org/wiki/Receiver_operating_characteristic#Area_under_curve>`_. Available for binary classification and learning-to-rank tasks.
+    - ``auc``: `Receiver Operating Characteristic Area under the Curve <http://en.wikipedia.org/wiki/Receiver_operating_characteristic#Area_under_curve>`_.
+      Available for classification and learning-to-rank tasks.
+
+      - When used with binary classification, the objective should be ``binary:logistic``.
+      - When used with multi-class classification, objective must be ``multi:softprob`` as ``multi:softmax`` doesn't output probability.  Also the AUC is calculated by 1-vs-rest with reference class weighted by class prevalence.
+      - When used with LTR task, the AUC is computed by comparing pairs of documents to count correctly sorted pairs.  This corresponds to pairwise learning to rank.  The implementation has some issues with average AUC around groups and distributed not being well defined.
```
trivialfis, line 407: distributed workers

src/common/common.h
```diff
@@ -163,8 +164,10 @@ inline void AssertOneAPISupport() {
...
 
-template <typename Idx, typename V, typename Comp = std::less<V>>
-std::vector<Idx> ArgSort(std::vector<V> const &array, Comp comp = std::less<V>{}) {
+template <typename Idx, typename Container,
+          typename V = typename Container::value_type,
+          typename Comp = std::less<V>>
+std::vector<Idx> ArgSort(Container const &array, Comp comp = std::less<V>{}) {
   std::vector<Idx> result(array.size());
   std::iota(result.begin(), result.end(), 0);
   std::stable_sort(
```
trivialfis, line 173: Use parallel sort.

src/common/device_helpers.cuh
```diff
@@ -1149,4 +1210,94 @@ auto Reduce(Policy policy, InputIt first, InputIt second, Init init, Func reduce
...
+    cub::DeviceRadixSort::SortPairsDescending(
+        storage.data().get(), bytes, values.data(), out.data().get(),
+        sorted_idx.data(), sorted_idx.data(), sorted_idx.size());
+  }
+}
+
+namespace detail {
+// Wrapper around cub sort to enforce size_t as offset type
+template <bool descending, typename KeyT, typename ValueT, typename OffsetIteratorT>
+void DeviceSegmentedRadixSortPair(
```
trivialfis, line 1244: Not needed.

src/metric/auc.cc
```diff
@@ -0,0 +1,320 @@
...
+
+  return std::make_tuple(fp, tp, auc);
+}
+
+/**
+ * Calculate AUC for multi-class classification problem using 1-vs-rest approach.
+ *
+ * TODO(jiaming): Use better algorithms like:
+ *
+ * - Hand, D. J. and Till, R. J. A Simple Generalisation of the Area Under the ROC Curve
```
trivialfis, line 72: Remove this.  AUCu is better.

src/metric/auc.cc
```diff
@@ -0,0 +1,320 @@
...
+
+#pragma omp parallel for reduction(+:sum_auc, sum_w)
+  for (size_t g = 1; g < info.group_ptr_.size(); ++g) {
+    omp_handler.Run([&]() {
+      size_t cnt = info.group_ptr_[g] - info.group_ptr_[g - 1];
+      float w = s_weights.empty() ? 1.0f : s_weights[g - 1];
+      auto g_predts = s_predts.subspan(info.group_ptr_[g - 1], cnt);
+      auto g_labels = s_labels.subspan(info.group_ptr_[g - 1], cnt);
+      float auc;
+      if (g_labels.size() < 3 && invalid_group != info.group_ptr_.size()) {
```
trivialfis, line 178: fix this.

src/metric/auc.cc
```diff
@@ -0,0 +1,320 @@
...
+XGBOOST_REGISTER_METRIC(EvalBinaryAUC, "auc")
+.describe("Receiver Operating Characteristic Area under the curve.")
+.set_body([](const char*) { return new EvalAUC(); });
+
+#if !defined(XGBOOST_USE_CUDA)
+std::tuple<float, float, float>
+GPUBinaryAUC(common::Span<float const> predts, MetaInfo const &info,
+             int32_t device, std::shared_ptr<DeviceAUCCache> *p_cache) {
+  common::AssertGPUSupport();
+  return {0, 0, 0};
```
trivialfis, line 303: use make_tuple.

src/metric/auc.cc
```diff
@@ -0,0 +1,320 @@
...
+float GPUMultiClassAUC(common::Span<float const> predts, MetaInfo const &info,
+                       int32_t device, std::shared_ptr<DeviceAUCCache>* cache) {
+  common::AssertGPUSupport();
+  return 0;
+}
+
+std::tuple<float, float> GPURankingAUC(common::Span<float const> predts,
+                                       MetaInfo const &info, int32_t device) {
+  common::AssertGPUSupport();
+  return {0, 0};
```
trivialfis, line 315: use make_tuple.

src/metric/auc.cu
```diff
@@ -0,0 +1,563 @@
...
+      thrust::cuda::par(alloc), check_it, check_it + group_ptr.size() - 1,
+      [=] __device__(size_t len) { return len < 3; });
+  if (n_invalid > 0) {
+    InvalidGroupAUC();
+  }
+  if (n_invalid == info.group_ptr_.size() - 1) {
+    return {0.0f, 0.0f};
+  }
+
+  dh::device_vector<size_t> sorted_idx(predts.size(), 0);
```
trivialfis, line 462: Use cache.

src/metric/auc.h
```diff
@@ -0,0 +1,41 @@
...
+#define XGBOOST_METRIC_AUC_H_
+#include <cmath>
+#include <memory>
+#include <tuple>
+
+#include "rabit/rabit.h"
+#include "xgboost/base.h"
+#include "xgboost/span.h"
+#include "xgboost/data.h"
+#include "../common/common.h"
```
trivialfis, line 14: Move into `auc.cc`.

src/metric/auc.h
```diff
@@ -0,0 +1,41 @@
+/*!
+ * Copyright 2021 by XGBoost Contributors
+ */
+#ifndef XGBOOST_METRIC_AUC_H_
+#define XGBOOST_METRIC_AUC_H_
+#include <cmath>
+#include <memory>
+#include <tuple>
+
+#include "rabit/rabit.h"
```
trivialfis, line 10: Not needed.

tests/cpp/metric/test_auc.cc
```diff
@@ -0,0 +1,90 @@
...
+  EXPECT_NEAR(GetMetricEval(metric, {1, 0, 0}, {0, 0, 1}), 0.25f, 1e-10);
+
+  // Invalid dataset
+  xgboost::MetaInfo info;
+  info.labels_ = {0, 0};
+  float auc = metric->Eval({1, 1}, info, false);
+  ASSERT_TRUE(std::isnan(auc));
+  info.labels_ = xgboost::HostDeviceVector<float>{};
+  auc = metric->Eval(xgboost::HostDeviceVector<float>{}, info, false);
+  // ASSERT_TRUE(std::isnan(auc));
```
trivialfis, line 35: Restore the test and extend it to multi class.

hcho3:
Does this PR also reimplement the learning to rank task? Can you point us to relevant literature for the new implementation?

trivialfis:
> Does this PR also reimplement the learning to rank task?

Yes.  For the AUC metric.  Some code was pulled from https://github.com/dmlc/xgboost/pull/6465 .

> Can you point us to relevant literature for the new implementation?

I will do that, it's a little bit complicated.

hcho3:
Got it. It would be nice to see how the new implementation is more in line with the latest literature on Learning To Rank (LTR). We can even write a short vignette or technical report to document our new LTR implementation.

trivialfis:
I have adopted some heuristics here.  Will explain in detail once I can get all tests passed.

hcho3:
> I have adopted some heuristics here.

Interesting. Once the implementation is complete and you can explain the heuristics, I'd like to volunteer time to write up a short technical report. This way, we could hopefully remember all the pertinent details and still be able to maintain the LTR objective and metrics.

trivialfis:
A smaller PR is extracted in https://github.com/dmlc/xgboost/pull/6749 .

doc/parameter.rst
```diff
@@ -399,7 +399,15 @@ Specify the learning task and the corresponding learning objective. The objectiv
...
-    - ``auc``: `Area under the curve <http://en.wikipedia.org/wiki/Receiver_operating_characteristic#Area_under_curve>`_. Available for binary classification and learning-to-rank tasks.
+    - ``auc``: `Receiver Operating Characteristic Area under the Curve <http://en.wikipedia.org/wiki/Receiver_operating_characteristic#Area_under_curve>`_.
+      Available for classification and learning-to-rank tasks.
+
+      - When used with binary classification, the objective should be ``binary:logistic`` or similar functions that work on probability.
+      - When used with multi-class classification, objective should be ``multi:softprob`` instead of ``multi:softmax``, as the latter doesn't output probability.  Also the AUC is calculated by 1-vs-rest with reference class weighted by class prevalence.
+      - When used with LTR task, the AUC is computed by comparing pairs of documents to count correctly sorted pairs.  This corresponds to pairwise learning to rank.  The implementation has some issues with average AUC around groups and distributed workers not being well ddefined.
+      - When used in distributed environment, result is weighted average among workers.
```
RAMitchell, line 408: On a single machine the AUC calculation is exact. In a distributed environment the AUC is a weighted average over the AUC of training rows on each node - therefore, distributed AUC is an approximation sensitive to the distribution of data across workers. Use another metric in distributed environments if precision and reproducibility are important.

src/common/ranking_utils.cuh
```diff
@@ -0,0 +1,80 @@
...
+#define XGBOOST_COMMON_RANKING_UTILS_H_
+
+#include <cub/cub.cuh>
+#include "xgboost/base.h"
+#include "device_helpers.cuh"
+#include "./math.h"
+
+namespace xgboost {
+namespace common {
+XGBOOST_DEVICE inline size_t DiscreteTrapezoidArea(size_t n, size_t h) {
```
RAMitchell, line 14: Define n,h

src/metric/auc.cu
```diff
@@ -0,0 +1,547 @@
...
+  /**
+   * Create sorted index for each class
+   */
+  auto d_sorted_idx = dh::ToSpan(cache->sorted_idx);
+  dh::ArgSort<false>(predts, d_sorted_idx);
+
+  /**
+   * Linear scan
+   */
+  dh::caching_device_vector<float> d_auc(1);
```
RAMitchell, line 105: If you specify the fill value as 0 in the caching_device_vector constructor, it should be safely initialised.

src/metric/auc.cu
```diff
@@ -0,0 +1,547 @@
...
+
+    return thrust::make_pair(fp, tp);
+  };  // NOLINT
+  auto d_fptp = dh::ToSpan(cache->fptp);
+  dh::LaunchN(device, d_sorted_idx.size(),
+              [=] __device__(size_t i) { d_fptp[i] = get_fp_tp(i); });
+
+  dh::XGBDeviceAllocator<char> alloc;
+  auto d_unique_idx = dh::ToSpan(cache->unique_idx);
+  dh::Iota(d_unique_idx, device);
```
RAMitchell, line 129: Is an Iota function necessary? What about thrust sequence?

src/metric/auc.cu
```diff
@@ -0,0 +1,547 @@
...
+  __device__ float operator()(size_t i) const {
+    return weights.empty() ? 1.0f : weights[sorted_idx[i]];
+  }
+};
+}  // namespace
+
+/**
+ * A cache to GPU data to avoid reallocating memory.
+ */
+struct DeviceAUCCache {
```
RAMitchell, line 41: How necessary are these cached items? Can you use dh::TemporaryArray to avoid caching? If I remember correctly caching_device_vector currently has synchronisation overhead due to upstream changes in thrust.

RAMitchell (commented):
Awesome to get multiclass and ranking support. Some minor comments. I'm assuming your math is correct.

It might be a good idea to add hypothesis tests comparing the results against sklearn. 

What is performance looking like?

src/metric/auc.cu
trivialfis: > has synchronisation overhead

Yes, happens in `thrust::fill_n`.  I will try to use temp array instead.

> How necessary are these cached items?

Just trying to avoid `cudaMalloc` for every iteration.

src/metric/auc.cu
trivialfis: Written for the `ArgSort`, which is async.

trivialfis:
> It might be a good idea to add hypothesis tests comparing the results against sklearn.

I have some test cases for comparing with sklearn on binary classification and multi-class classification.  Ranking is not supported by sklearn.

I will try to make the data more random with hypothesis.

trivialfis:
@RAMitchell 

I added parametrize test with different sample sizes.  scikit-learn throws an error when the dataset is invalid so the degenerate case is tested separately.

I will run some basic benchmark later.

src/metric/auc.cu
trivialfis: Swapped some caching device vectors with temporary arrays.

trivialfis:
Ran small benchmark on higgs with `Booster.eval`.  There's no performance change for binary classification:

* before
CPU 2.074251174926758
GPU 0.022435665130615234

* after
CPU 2.070319652557373
GPU 0.022629976272583008

RAMitchell (approved):
Nice work.

hcho3 (approved):
LGTM. Some comments:

src/metric/auc.cc
```diff
@@ -0,0 +1,338 @@
...
+  }
+  return auc_sum;
+}
+
+/**
+ * Calculate AUC for 1 ranking group;
+ */
+float GroupRankingAUC(common::Span<float const> predts,
+                      common::Span<float const> labels, float w) {
+  // on ranking, we just count all pairs.
```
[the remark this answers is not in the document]
trivialfis: Actually I want to remove the random sampling in objective and replace it with a top k parameter.

I implement AUC for ranking mostly for compatibility reason.  I don't think it's a good metric for ranking.  Averaging between groups is destroying all the auc properties.

src/metric/auc.cc
```diff
@@ -0,0 +1,338 @@
...
+  std::vector<float> results(n_classes * 3, 0);
+  auto s_results = common::Span<float>(results);
+  auto local_area = s_results.subspan(0, n_classes);
+  auto tp = s_results.subspan(n_classes, n_classes);
+  auto auc = s_results.subspan(2 * n_classes, n_classes);
+
+  if (!info.labels_.Empty()) {
+    dmlc::OMPException omp_handler;
+#pragma omp parallel for
+    for (omp_ulong c = 0; c < n_classes; ++c) {
```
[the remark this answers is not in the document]
trivialfis: I don't know what's the best way for CPU.  The GPU is parallized by the scan operator.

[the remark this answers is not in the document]
trivialfis: Maybe we can look at tensorflow and see how they calculate AUC.

trivialfis:
@hcho3 All style comments are addressed.

src/metric/auc.cc
```diff
@@ -0,0 +1,338 @@
...
+float MultiClassOVR(std::vector<float> const& predts, MetaInfo const& info) {
+  auto n_classes = predts.size() / info.labels_.Size();
+  CHECK_NE(n_classes, 0);
+  auto const& labels = info.labels_.ConstHostVector();
+
+  std::vector<float> results(n_classes * 3, 0);
+  auto s_results = common::Span<float>(results);
+  auto local_area = s_results.subspan(0, n_classes);
+  auto tp = s_results.subspan(n_classes, n_classes);
+  auto auc = s_results.subspan(2 * n_classes, n_classes);
```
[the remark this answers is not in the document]
trivialfis: It would be even better if we have a matrix/tensor library backend.  ;-)  If I were to start a new ML project that might be the first thing I do...

src/metric/auc.cc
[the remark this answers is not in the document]
EvgenShSg: @trivialfis Hi, been watching through xgboost versions and stumbled upon this discussion. Could you please elaborate on this idea of averaging between groups that destroys auc properties? Or could you give a link to read about it? Thx in advance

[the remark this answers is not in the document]
trivialfis: The AUC is "area under the curve", which is a normalized ratio with a geometry interpretation. One cannot average ratios directly.

PR #6749, opened 2021-03-13
Add device argsort.

trivialfis:
Extracted from https://github.com/dmlc/xgboost/pull/6747 .

### DIFF
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -1189,7 +1189,8 @@ template <bool accending, typename IdxT, typename U>
 void ArgSort(xgboost::common::Span<U> values, xgboost::common::Span<IdxT> sorted_idx) {
   size_t bytes = 0;
   Iota(sorted_idx);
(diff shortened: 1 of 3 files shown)


src/common/device_helpers.cuh
```diff
@@ -1179,4 +1184,70 @@ auto Reduce(Policy policy, InputIt first, InputIt second, Init init, Func reduce
   }
   return aggregate;
 }
+
+template <bool accending, typename IdxT, typename U>
+void ArgSort(xgboost::common::Span<U> values, xgboost::common::Span<IdxT> sorted_idx) {
+  size_t bytes = 0;
```
RAMitchell, line 1190: Maybe get into the habit of checking if cub input length is < 2^31. Let's try to avoid our previous troubles with cub.

RAMitchell (approved):

src/common/device_helpers.cuh
trivialfis: Done.

Issue #6754, opened 2021-03-16
XGBRFRegressor much worse for gpu_hist vs hist on small data


arnocandel:
Python 3.6.9

```
pip install xgboost==1.3.3
pip install scikit-learn==0.24.1
```

```
from xgboost import XGBRFRegressor
from sklearn.datasets import load_boston
from sklearn.metrics import mean_squared_error

X, y = load_boston(return_X_y=True)
split=300

X_train = X[:split, :]
y_train = y[:split]

X_test = X[split:, :]
y_test = y[split:]

# CPU
model = XGBRFRegressor(tree_method='hist')
model.fit(X_train, y_train)
preds = model.predict(X_test)
print(mean_squared_error(y_test, preds))

# GPU
model = XGBRFRegressor(tree_method='gpu_hist')
model.fit(X_train, y_train)
preds = model.predict(X_test)
print(mean_squared_error(y_test, preds))
```

GPU method produces inferior results:

```
38.15866831928765
188.78431456187735
```

Note: predicting the mean leads to `114.115` squared error, so `gpu_hist` is much worse than predicting a constant.

x86 Ubuntu 18.04 LTS
i9, Xeon E5 and 3970X CPUs
1080Ti, 2080Ti, 3090 GPUs

same for latest https://s3-us-west-2.amazonaws.com/xgboost-nightly-builds/xgboost-1.4.0_SNAPSHOT%2B1a73a2851143bb8a8a5b9fb307cfa059d3fed578-py3-none-manylinux2010_x86_64.whl

PR #6755, opened 2021-03-16
Fix GPU RF

trivialfis:
Close https://github.com/dmlc/xgboost/issues/6754 .

Fix by copying out the gradient before sampling.

### DIFF
--- a/src/tree/updater_gpu_hist.cu
+++ b/src/tree/updater_gpu_hist.cu
@@ -169,6 +169,7 @@ struct GPUHistMakerDevice {
   std::unique_ptr<RowPartitioner> row_partitioner;
   DeviceHistogram<GradientSumT> hist{};
 
+  dh::caching_device_vector<GradientPair> d_gpair;  // storage for gpair;
   common::Span<GradientPair> gpair;
 
   dh::caching_device_vector<int> monotone_constraints;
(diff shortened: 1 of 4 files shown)


src/tree/updater_gpu_hist.cu
```diff
@@ -269,7 +270,13 @@ struct GPUHistMakerDevice {
...
               GradientPair());
 
-    auto sample = sampler->Sample(dh_gpair->DeviceSpan(), dmat);
+    if (d_gpair.size() != dh_gpair->Size()) {
+      d_gpair.resize(dh_gpair->Size());
+    }
+    thrust::copy(thrust::device, dh_gpair->ConstDevicePointer(),
+                 dh_gpair->ConstDevicePointer() + dh_gpair->Size(),
+                 d_gpair.begin());
+    auto sample = sampler->Sample(dh::ToSpan(d_gpair), dmat);
```
hcho3, line 279: We have to make a copy here since `Sample()` is expected to modify the incoming `gpair` parameter? Mutating gpairs would be bad when we fit multiple trees in a single round.

trivialfis: > modify the incoming gpair parameter?

Yes.

hcho3 (approved):

src/tree/updater_gpu_hist.cu
pseudotensor: Thanks guys!

arnocandel: Fix confirmed. Thanks.

Issue #6768, opened 2021-03-18
R documentation: Construction of dtrain is inconsistent and can be shortened.


ReeceGoding:
The following pattern is often seen in [the R documentation](https://cran.r-project.org/web/packages/xgboost/xgboost.pdf):
```R
data(agaricus.train, package='xgboost')
train <- agaricus.train
dtrain <- xgb.DMatrix(train$data, label=train$label)
```
By my count, this is used in the docs for `dim.xgb.DMatrix`, `dimnames.xgb.DMatrix`, `getinfo`, `print.xgb.DMatrix`, `setinfo`, `slice`, `xgb.DMatrix`, and `xgb.DMatrix.save`. It is also found for a `dtest` variable in `predict.xgb.Booster`'s docs.

However, the following pattern is instead found in the docs for `xgb.create.features` (and again in the construction of`new.dtrain`),  `xgb.cv`, and `xgb.train`:
```R
data(agaricus.train, package='xgboost')
dtrain <- xgb.DMatrix(agaricus.train$data, label = agaricus.train$label)
```
Most commonly, this second pattern is also used for a `dtest` variable as well, making the full block:
```R
data(agaricus.train, package='xgboost')
data(agaricus.test, package='xgboost')

dtrain <- xgb.DMatrix(agaricus.train$data, label = agaricus.train$label)
dtest <- xgb.DMatrix(agaricus.test$data, label = agaricus.test$label)
```
I suggest that these should be made consistent.

The shortest solution that I know is to use:
```R
data(agaricus.train, package='xgboost')
data(agaricus.test, package='xgboost')
dtrain <- with(agaricus.train, xgb.DMatrix(data, label = label))
dtest <- with(agaricus.test, xgb.DMatrix(data, label = label))
```
and its equivalent for the shorter cases:
```R
data(agaricus.train, package='xgboost')
dtrain <- with(agaricus.train, xgb.DMatrix(data, label = label))
```
The only issue that I can see with my suggestion is that R tells me that the following returns `FALSE`:
```R
train <- agaricus.train
dtrain2 <- xgb.DMatrix(train$data, label = train$label)
identical(dtrain, dtrain2)
```
However, I think that is just because the external pointers for the two variables disagree. The fact that `all.equal(dtrain, dtrain2)` is `TRUE` suggests that the issue is very minor.

If we're happy with this, it shouldn't take me long to make a pull request changing the docs to the `with` version.

trivialfis:
[...]

PR #6772, opened 2021-03-19
R documentation: Close issue #6768 (consistency of construction of dtest and dtrain variables)

ReeceGoding:
Almost all of this was just a find and replace job for the relevant constructions. I left any cases that weren't exactly the discussed templates alone. Of particular note, I ignored the `dtest` variable in `predict.xgb.Booster`'s docs, left the examples that used the `train` variable in `dim.xgb.DMatrix` and `dimnames.xgb.DMatrix` unmodified (because they used `train`, which this change would remove), and I ignored both `new.dtrain` and `new.dtest` in the docs for `xgb.create.features` (in fact, in that case, we cannot use `with` at all). However, I did edit the construction of the original `dtrain` and `dtest` variables in `predict.xgb.Booster`'s docs, which turned out to be even more verbose than the examples that I was originally complaining about (they used the `data` argument by name, unlike every other example).

As this was just a find and replace job, I have largely not tested the examples in the documentation that I've changed. However, as mentioned in the original issue, I am very highly confident that this code is exactly equivalent to the original code, only shorter and more consistent. The only examples that have any reasons to break are the ones that used the `train` variable elsewhere, but as mentioned above, I have left those unchanged. As far as I can tell, every other example only used the now-removed `train` and `test` variables in their constructions of `dtrain` and `dtest`.


trivialfis (approved):
Thanks for improving the documents!  It looks good to me.

Issue #6723, opened 2021-02-23
Eval displays 'nan' with 'reg:gamma' objective and 'gbtree' booster


honzasterba:
seems to be related to #5349, but was not fixed by the fix

script to reproduce:

```python
import xgboost as xgb
dtrain = xgb.DMatrix("matrix.part0")
param = {
    'booster': "gbtree",
    'tree_method': "approx",
    'max_depth': 6, 
    "objective": "reg:gamma",
    "lambda": 1.0,
    'gamma': 0.0, 
    'nthread': 16
}
num_round = 50
bst = xgb.train(param, dtrain, num_round, [(dtrain, "train")])
```

script output:
```
[0]	train-gamma-nloglik:nan
[1]	train-gamma-nloglik:nan
[2]	train-gamma-nloglik:nan
[3]	train-gamma-nloglik:nan
[4]	train-gamma-nloglik:nan
[5]	train-gamma-nloglik:nan
[6]	train-gamma-nloglik:nan
[7]	train-gamma-nloglik:nan
[8]	train-gamma-nloglik:nan
[9]	train-gamma-nloglik:nan
...
```

data and script attached
[nan_metrics.zip](https://github.com/dmlc/xgboost/files/6028259/nan_metrics.zip)

tested with: 1.2.0 and 1.3.3, same result

trivialfis:
I need to restore the estimation on dispersion.

trivialfis:
For now please use gamma deviance instead, it should be a good metric as the objective is actually derived from it with log link.

trivialfis:
Your data contains 0.  You need to do some per-processing to remove them.

PR #6778, opened 2021-03-24
Verify strictly positive for gamma regression.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/6723 .

### DIFF
--- a/src/metric/elementwise_metric.cu
+++ b/src/metric/elementwise_metric.cu
@@ -312,7 +312,7 @@ struct EvalGammaNLogLik {
     // b = -std::log(-theta);
     float b = 1.0f;
     // c = 1. / kPsi * std::log(y/kPsi) - std::log(y) - common::LogGamma(1. / kPsi);
-    //     1.0f      * 1.0f             - 1.0f        - 0
+    //   = 1.0f      * std::log(y)      - std::log(y) - 0 = 0
     float c = 0;
     // general form for exponential family.
     return -((y * theta - b) / a + c);
--- a/src/metric/elementwise_metric.cu
+++ b/src/metric/elementwise_metric.cu
@@ -304,13 +304,17 @@ struct EvalGammaNLogLik {
   }
 
   XGBOOST_DEVICE bst_float EvalRow(bst_float y, bst_float py) const {
-    const bst_float eps = 1e-16f;
-    if (y < eps) y = eps;
-    bst_float psi = 1.0;
+    py = std::max(py, 1e-6f);
+    // hardcoded dispersion.
+    float constexpr kPsi = 1.0;
     bst_float theta = -1. / py;
-    bst_float a = psi;
-    bst_float b = -std::log(-theta);
-    bst_float c = 1. / psi * std::log(y/psi) - std::log(y) - common::LogGamma(1. / psi);
+    bst_float a = kPsi;
+    // b = -std::log(-theta);
+    float b = 1.0f;
+    // c = 1. / kPsi * std::log(y/kPsi) - std::log(y) - common::LogGamma(1. / kPsi);
+    //     1.0f      * 1.0f             - 1.0f        - 0
+    float c = 0;
+    // general form for exponential family.
     return -((y * theta - b) / a + c);
   }
   static bst_float GetFinal(bst_float esum, bst_float wsum) {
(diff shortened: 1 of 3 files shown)


src/metric/elementwise_metric.cu
```diff
@@ -304,13 +304,17 @@ struct EvalGammaNLogLik {
   }
 
   XGBOOST_DEVICE bst_float EvalRow(bst_float y, bst_float py) const {
-    const bst_float eps = 1e-16f;
-    if (y < eps) y = eps;
-    bst_float psi = 1.0;
+    py = std::max(py, 1e-6f);
+    // hardcoded dispersion.
+    float constexpr kPsi = 1.0;
```
hcho3, line 309: Do we have a plan to allow for a different value for dispersion in the future?

src/metric/elementwise_metric.cu
```diff
@@ -304,13 +304,17 @@ struct EvalGammaNLogLik {
...
-    bst_float a = psi;
-    bst_float b = -std::log(-theta);
-    bst_float c = 1. / psi * std::log(y/psi) - std::log(y) - common::LogGamma(1. / psi);
+    bst_float a = kPsi;
+    // b = -std::log(-theta);
+    float b = 1.0f;
+    // c = 1. / kPsi * std::log(y/kPsi) - std::log(y) - common::LogGamma(1. / kPsi);
+    //     1.0f      * 1.0f             - 1.0f        - 0
+    float c = 0;
+    // general form for exponential family.
```
hcho3, line 317: Did you manage to locate a reference for this?

src/metric/elementwise_metric.cu
trivialfis: The dispersion should be estimated from data.  We can put more efforts to clarity the choice or add estimation in the code,right now I'm just adding notes.

hcho3: It can be treated as a hyperparameter.

trivialfis: I'm not sure yet,as you can see the metric doesn't use label at all.

trivialfis: The dependency on labels is hidden in dispersion.

hcho3: Wait, why is `std::log(y)` equal to 1.0f?

src/metric/elementwise_metric.cu
```diff
@@ -304,13 +304,17 @@ struct EvalGammaNLogLik {
...
+    float constexpr kPsi = 1.0;
     bst_float theta = -1. / py;
-    bst_float a = psi;
-    bst_float b = -std::log(-theta);
-    bst_float c = 1. / psi * std::log(y/psi) - std::log(y) - common::LogGamma(1. / psi);
+    bst_float a = kPsi;
+    // b = -std::log(-theta);
+    float b = 1.0f;
+    // c = 1. / kPsi * std::log(y/kPsi) - std::log(y) - common::LogGamma(1. / kPsi);
+    //     1.0f      * 1.0f             - 1.0f        - 0
```
hcho3, line 315: If `y` is a positive value, `std::log(y)` is not necessarily equal to 1.0f, I think.

src/metric/elementwise_metric.cu
trivialfis: All exponential distribution can be expressed in this form disregarding parametrisation.

src/metric/elementwise_metric.cu
trivialfis: There are 2 log y, they cancel out.

src/metric/elementwise_metric.cu
trivialfis: Where?

src/metric/elementwise_metric.cu
hcho3: Thanks, I will suggest a small change to clarify this point.

src/metric/elementwise_metric.cu
hcho3: See my next comment, which you just responded.

src/metric/elementwise_metric.cu
trivialfis: The 1.0 is misleading, will remove it.

hcho3, line 315:
```suggestion
    // c = 1. / kPsi * std::log(y/kPsi) - std::log(y) - common::LogGamma(1. / kPsi);
    //   = 1.0f      * std::log(y)      - std::log(y) - 0 = 0
```

hcho3 (approved):

RAMitchell (approved):

Issue #6552, opened 2020-12-25
Training behaviour difference between v1.1.0 and v1.3.1


fordicus:
Merry Christmas, all.

I am suffering from a different behaviour between an old version and 
the newest version of XGBoost. Here are some information.

* 1.3.1 ( pip install xgboost                  ) the newest one
* 1.1.0 ( pip install -Iv xgboost==1.1.0 ) the old one that I used
Ubuntu 18.04.4 LTS
CUDA   11.1.74
QUADRO RTX 8000 48GB

Data file (7.49 GB) can be downloaded from (mega cloud)
https://mega.nz/file/irQXCSbA#rfl53ewh88k_pBsVLw_fA5hC6mz91Vi9yRUrMqEMYno

or it also can be found at (google drive)
https://drive.google.com/file/d/14RCv2COi6rw7JWDkYp9825z7zSHY1ogQ/view?usp=sharing

The issue with the old version was not being able to dump the trained model
while the training itself progresses very nicely.
( https://discuss.xgboost.ai/t/serialisation-of-large-models/1897/2 )

As I am aware that there have been a lot of effort to improve the dump process,
I had to try the newest version. But using the newst version, the issue is that
I cannot see the same training behaviour as the old one. Simply put,
it seems it takes a way longer time in order to see the rounds proceed.
I only terminated with a keyboard interruption.

For 30 minutes, there is no one-single progress of a round using v1.3.1.
Using v1.1.0 and waiting for 30 minutes, the prints at terminal read

[0]  convergence-merror:0.10068
[1]  convergence-merror:0.10052
[2]  convergence-merror:0.09998
[3]  convergence-merror:0.09930
[4]  convergence-merror:0.09852
[5]  convergence-merror:0.09773
[6]  convergence-merror:0.09690
[7]  convergence-merror:0.09608
[8]  convergence-merror:0.09524
[9]  convergence-merror:0.09439
[10] convergence-merror:0.09355
[11] convergence-merror:0.09268
[12] convergence-merror:0.09182
[13] convergence-merror:0.09097
[14] convergence-merror:0.09013
[15] convergence-merror:0.08930
[16] convergence-merror:0.08855
[17] convergence-merror:0.08781
[18] convergence-merror:0.08708
[19] convergence-merror:0.08636
[20] convergence-merror:0.08557
[21] convergence-merror:0.08477

Note that reinstalling the old version after removing the newest recovers 
the previous good training procedure.

[...]

PR #6779, opened 2021-03-24
Memory consumption fix for row-major adapters

ShvetsKS:


ShvetsKS:
related issue: https://github.com/dmlc/xgboost/issues/6552

ShvetsKS:
Step "XGBoost-R-Tests/..." is failed due to connection problems:
"The remote server returned an error: (502) Bad Gateway. Bad Gateway"

ShvetsKS:
@trivialfis thank for removing unused variable :)

RAMitchell:
Is there any solution that does not specialise for row major only?

trivialfis:
I spent some time on reading papers about matrix transpose, see comments in linked issue.

ShvetsKS:
> Is there any solution that does not specialise for row major only?

Yes, as we have to do transpose for columnar-major case there is no common effective solution in my opinion. If it is possible to have resulting columnar-major DMatrix we can some kind of generalize the code effectively.

ShvetsKS:
not sure that [this failure](https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-6779/9/pipeline#step-339-log-278) related to current changes
[2021-03-25T08:11:20.565Z] terminate called after throwing an instance of 'thrust::system::system_error'

[2021-03-25T08:11:20.565Z]   what():  for_each: failed to synchronize: cudaErrorIllegalAddress: an illegal memory access was encountered

trivialfis (approved):
Looks good to me.  Thanks for the quick fix!


@RAMitchell @hcho3 I don't have a better solution for this as it requires a general matrix transpose procedure.  I would like to get it into 1.4.  WDYT?

RAMitchell:
I'm happy to proceed.

PR #6783, opened 2021-03-25
Support configuring constraints by feature names

### LINKED ISSUE #6704: [FEA] Allow users to pass in feature names for monotonicity or interaction constraints.
Hi! I love the monotonicity and interaction constraints in XGBoost, but the current method of specifying them relies on long text strings of all the features. An example for monotone constraints:
`"(0,1,1,0,0,-1...)"` 
Not only can this get long if you have a lot of features, but often you want only a few features constrained with the rest zeros. Further, this method of passing in the constraints is not flexible if you are fitting many models with different feature subsets or features in different orders.

One awesome addition to make this easier would be to allow users pass in constraints as a dictionary:
`{"X1": 1, "X10" -1, ...}` where the keys are feature names and the values are the constraints (and features not present could be assigned the default of `0`). 

The caveat here is that the user must then pass in a dataset that has column names (either a DMatrix or DataFrame-like object).

A few lines of code during the `fit` and `train` methods would then convert this dictionary into the usual string list that is passed to the underlying tree builder using the column names in the data. If the user passes data that does not contain the features, an error could raised.

I am happy to provide some dummy code if it would help.

giladmaya:
Resolves #6704

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -1253,26 +1253,35 @@ class Booster(object):
         feature_idx_mapping = {k: str(v) for v, k in enumerate(self.feature_names or [])}
 
         try:
-            s = '['
+            s = "["
             for constraint in value:
-                s += '[' + ','.join([feature_idx_mapping[feature_name]
-                                     for feature_name in constraint]) + ']'
-            return s + ']'
-        except KeyError:
+                s += (
+                    "["
+                    + ",".join(
+                        [feature_idx_mapping[feature_name] for feature_name in constraint]
+                    )
+                    + "]"
+                )
+            return s + "]"
+        except KeyError as e:
             # pylint: disable=raise-missing-from
-            raise ValueError('Constrained features are not a subset of '
-                                     'training data feature names')
-
+            raise ValueError(
+                "Constrained features are not a subset of training data feature names"
+            ) from e
 
     def _configure_constraints(self, params: Union[Dict, List]) -> Union[Dict, List]:
         if isinstance(params, dict):
-            value = params.get('monotone_constraints')
+            value = params.get("monotone_constraints")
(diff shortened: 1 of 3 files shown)


python-package/xgboost/core.py
```diff
@@ -1233,6 +1234,59 @@ def _configure_metrics(self, params: Union[Dict, List]) -> Union[Dict, List]:
...
+        try:
+            s = '['
+            for constraint in value:
+                s += '[' + ','.join([feature_idx_mapping[feature_name]
+                                     for feature_name in constraint]) + ']'
+            return s + ']'
+        except KeyError:
+            # pylint: disable=raise-missing-from
+            raise ValueError('Constrained features are not a subset of '
+                                     'training data feature names')
```
trivialfis, line 1264:
``` python
except KeyError as e:
  raise ValueError("...") from e
```

python-package/xgboost/core.py
```diff
@@ -1193,6 +1193,7 @@ def __init__(self, params=None, cache=(), model_file=None):
 
         params = params or {}
         params = self._configure_metrics(params.copy())
+        params = self._configure_constraints(params)
```
trivialfis, line 1196: It might be a good idea to make a copy since you are modifying the value.  Python arguments are passed as reference so after training the parameter in users' code will be changed.

trivialfis (commented):
Thanks for the PR!  Looks good to me in general, a few minor suggestions.

hcho3 (approved):

python-package/xgboost/core.py
trivialfis: NVM.  Just saw the copy.

trivialfis:
Thanks for the PR!

Issue #6521, opened 2020-12-16
hang in import of xgboost due to cudf even if not using cudf


pseudotensor:
Similar to: https://github.com/dmlc/xgboost/issues/6282 but on latest 1.4.0 SNAPSHOT, presume in 1.3.0 as well.
```
Current thread 0x00007f45fb09b740 (most recent call first):
  File "<frozen importlib._bootstrap>", line 219 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 922 in create_module
  File "<frozen importlib._bootstrap>", line 571 in module_from_spec
  File "<frozen importlib._bootstrap>", line 658 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 955 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 971 in _find_and_load
  File "/opt/h2oai/dai/cuda-10.0/lib/python3.6/site-packages/cudf/utils/gpu_utils.py", line 10 in validate_setup
  File "/opt/h2oai/dai/cuda-10.0/lib/python3.6/site-packages/cudf/__init__.py", line 5 in <module>
  File "<frozen importlib._bootstrap>", line 219 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 678 in exec_module
  File "<frozen importlib._bootstrap>", line 665 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 955 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 971 in _find_and_load
  File "/opt/h2oai/dai/cuda-10.0/lib/python3.6/site-packages/xgboost/compat.py", line 46 in <module>
  File "<frozen importlib._bootstrap>", line 219 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 678 in exec_module
  File "<frozen importlib._bootstrap>", line 665 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 955 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 971 in _find_and_load
  File "/opt/h2oai/dai/cuda-10.0/lib/python3.6/site-packages/xgboost/core.py", line 21 in <module>
  File "<frozen importlib._bootstrap>", line 219 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 678 in exec_module
  File "<frozen importlib._bootstrap>", line 665 in _load_unlocked
[...]

PR #6785, opened 2021-03-26
Remove dask import in compat

trivialfis:


tests/python/testing.py
```diff
@@ -33,8 +32,13 @@ def no_sklearn():
 
 
 def no_dask():
-    return {'condition': not DASK_INSTALLED,
-            'reason': 'Dask is not installed'}
+    reason = "dask/distributed is not installed."
+    try:
+        import dask             # noqa
+        import distributed      # noqa
```
hcho3, line 38: I'm not sure if importing dask to check its existence is necessarily a good idea. Please see #6521 and #6522

hcho3 (changes requested):

tests/python/testing.py
trivialfis: This is in testing module, are we worrying about import?

PR #6787, opened 2021-03-26
Fix inplace predict missing value.

trivialfis:
* Fix missing value on GPU in-place predict.
* Accept integer as missing value.

### DIFF
--- a/src/c_api/c_api_utils.h
+++ b/src/c_api/c_api_utils.h
@@ -11,6 +11,9 @@
 #include "xgboost/logging.h"
 #include "xgboost/json.h"
 #include "xgboost/learner.h"
+#include "xgboost/c_api.h"
+
+#include "c_api_error.h"
 
 namespace xgboost {
 /* \brief Determine the output shape of prediction.
@@ -140,6 +143,20 @@ inline uint32_t GetIterationFromTreeLimit(uint32_t ntree_limit, Learner *learner
     ntree_limit /= std::max(num_parallel_tree, 1u);
   }
   return ntree_limit;
+}
+
+inline float GetMissing(Json const &config) {
+  float missing;
+  auto const& j_missing = config["missing"];
+  if (IsA<Number const>(j_missing)) {
+    missing = get<Number const>(j_missing);
+  } else if (IsA<Integer const>(j_missing)) {
+    missing = get<Integer const>(j_missing);
+  } else {
+    missing = nan("");
+    LOG(FATAL) << "Invalid missing value: " << j_missing;
+  }
+  return missing;
 }
 }  // namespace xgboost
 #endif  // XGBOOST_C_API_C_API_UTILS_H_
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -171,9 +171,9 @@ struct DeviceAdapterLoader {
     }
     auto value = batch.GetElement(ridx * columns + fidx).value;
     if (is_valid(value)) {
-      return nan("");
+      return value;
     } else {
-      return batch.GetElement(ridx * columns + fidx).value;
+      return nan("");
     }
   }
 };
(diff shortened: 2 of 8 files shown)


trivialfis:
I need to refactor the code around adapters at some point.

src/c_api/c_api_utils.h
```diff
@@ -141,5 +144,19 @@ inline uint32_t GetIterationFromTreeLimit(uint32_t ntree_limit, Learner *learner
   }
   return ntree_limit;
 }
+
+inline float GetMissing(Json const &config) {
+  float missing;
+  auto const& j_missing = config["missing"];
+  if (IsA<Number const>(j_missing)) {
```
hcho3, line 151: Is "nan" a valid `Number`?

src/predictor/gpu_predictor.cu
```diff
@@ -164,7 +169,12 @@ struct DeviceAdapterLoader {
     if (use_shared) {
       return smem[threadIdx.x * columns + fidx];
     }
-    return batch.GetElement(ridx * columns + fidx).value;
+    auto value = batch.GetElement(ridx * columns + fidx).value;
+    if (is_valid(value)) {
+      return nan("");
+    } else {
+      return batch.GetElement(ridx * columns + fidx).value;
```
hcho3, line 176:
```suggestion
    auto value = batch.GetElement(ridx * columns + fidx).value;
    if (is_valid(value)) {
      return value;
    } else {
      return nan("");
```

If `value` is valid, you shouldn't return a NaN.

hcho3 (changes requested):

src/predictor/gpu_predictor.cu
trivialfis: Thanks for catching that!

src/c_api/c_api_utils.h
trivialfis: Yes.  Added test to show that.

src/predictor/gpu_predictor.cu
hcho3: @trivialfis Can you add a unit test to cover this code?

trivialfis: Em .. I need to create a dataset that exceeds the size of shared memory (which is gpu dependent).

trivialfis: Let me do that.

hcho3: If it's too difficult, we can merge this PR as it is.

trivialfis: Done.  It's good to have you as a reviewer so I can try harder on testing.

tests/python-gpu/test_gpu_prediction.py
```diff
@@ -185,6 +190,20 @@ def predict_dense(x):
...
+        # Create a wide dataset
+        X = cp_rng.randn(100, 10000)
+        y = cp_rng.randn(100)
+
+        missing_idx = [i for i in range(0, X.shape[1], 16)]
+        X[:, missing_idx] = missing
+        reg = xgb.XGBRegressor(tree_method="gpu_hist", n_estimators=8, missing=missing)
+        reg.fit(X, y)
+
+        gpu_predt = reg.predict(X)
```
trivialfis, line 202: This is in-place since X is supported by inplace predict and no `predictor` is specified.

tests/python-gpu/test_gpu_prediction.py
```diff
@@ -185,6 +190,20 @@ def predict_dense(x):
...
+        reg.set_params(predictor="cpu_predictor")
+        cpu_predt = reg.predict(X)
```
trivialfis, line 204: This is not inplace as we have specified `cpu_predictor`.  (When predictor is specified we have to fallback to normal predict as we don't know what are the inputs).

src/predictor/gpu_predictor.cu
hcho3: Good to hear that my review helped.

hcho3 (approved):

PR #6798, opened 2021-03-29
Hide symbols in CI build + hide symbols for C and CUDA

trivialfis:
Recently I ran into an issue where `cub::DeviceRadixSort` runs into a cuda invalid value error when used with cuDF.  I believe it's due to mixed C++ symbols exported from both libraries.  ~This might also be the cause of https://github.com/rapidsai/cudf/issues/7722~.  This PR enables "hidden" visibility for all c++ symbols by default on our CI build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,7 +197,9 @@ endif (USE_CUDA)
 #-- Hide all C++ symbols
 if (HIDE_CXX_SYMBOLS)
   foreach(target objxgboost xgboost dmlc)
+    set_target_properties(${target} PROPERTIES C_VISIBILITY_PRESET hidden)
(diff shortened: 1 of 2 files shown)


trivialfis:
@hcho3 Now I'm thinking maybe we set it to default in CMake option?

hcho3 (approved):

trivialfis:
@hcho3 Could you confirm that this will change our nightly build?

hcho3:
@trivialfis The nightly build already uses `-DHIDE_CXX_SYMBOLS=ON`. Setting `C_VISIBILITY_PRESET` and `CUDA_VISIBILITY_PRESET` will help.

trivialfis:
Thanks @hcho3 !

PR #6799, opened 2021-03-29
Typehint for Sklearn.

trivialfis:
TODO:
~- [ ] Remove type from doc string.~  Need to update CI sphinx-build .

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -334,6 +334,13 @@ def _wrap_evaluation_matrices(
     return train_dmatrix, evals
 
 
+_SklObjective = Optional[
+    Union[
+        str, Callable[[np.ndarray, np.ndarray], Tuple[np.ndarray, np.ndarray]]
+    ]
+]
+
+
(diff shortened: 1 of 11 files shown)


python-package/xgboost/sklearn.py
```diff
@@ -903,12 +920,14 @@ def evals_result(self):
         .. code-block:: python
 
             {'validation_0': {'logloss': ['0.604835', '0.531479']},
-            'validation_1': {'logloss': ['0.41965', '0.17686']}}
+             'validation_1': {'logloss': ['0.41965', '0.17686']}}
         """
-        if self.evals_result_:
+        if hasattr(self, "evals_result_"):
```
trivialfis, line 925: Also check for null.

trivialfis:
Ideally we should remove the type annotations in doc strings, but there are a few things we need to do first:
- I found the upgrading sphinx-build can help display types correctly.  I'm not sure whether it's up-to-date on our CI.
- The `array_like` is still an issue.  We need newer version of numpy and Python 3.8 to use `numpy.typing.ArrayLike`.

hcho3 (approved):

PR #6777, opened 2021-03-24
Fix dart inplace prediction with GPU input.

trivialfis:
* On C API there's a check for output prediction being device-readable.  With dart this doesn't hold.
* Added optimization to avoid memory copying whenever possible.  Will post the result later.

PR #6804, opened 2021-03-30
Optimize dart inplace predict perf.

trivialfis:
Similar to optimization done in https://github.com/dmlc/xgboost/pull/6777 , but for in-place prediction.


trivialfis:
For 100 trees higgs:

* Before
GPU Inplace Predict::Duration 7.213954210281372

* After
GPU Inplace Predict::Duration 2.420344591140747

RAMitchell (approved):

PR #6805, opened 2021-03-30
Set `HIDE_CXX_SYMBOLS` to ON by default.

trivialfis:
Close https://github.com/dmlc/xgboost/pull/6798 .


hcho3 (approved):

RAMitchell (approved):

trivialfis:
Actually, I have some concerns over this PR.  Yes, it prevents all sorts of symbol conflicts.  But at the same time, the compiler is allowed to do some weird things when all symbols are hidden.  For example, gdb can no longer look into some code sections due to striped functions.  I have experience on some other projects that toggling visibility can generate different program behavior.  Our tests are working fine, but there will be some unknown subtleties.

PR #6807, opened 2021-03-30
Don't estimate sketch batch size when rmm is used.

trivialfis:


RAMitchell (approved):
This is not ideal, I don't see a way around it though.

PR #6809, opened 2021-03-30
Update document for sklearn model IO.

trivialfis:
* Update the use of JSON.
* Remove unnecessary type cast.

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -4,6 +4,7 @@
 import copy
 import warnings
 import json
+import os
 from typing import Union, Optional, List, Dict, Callable, Tuple, Any
 import numpy as np
 from .core import Booster, DMatrix, XGBoostError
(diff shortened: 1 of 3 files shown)


python-package/xgboost/sklearn.py
```diff
@@ -574,9 +548,6 @@ def load_model(self, fname):
             if k == 'classes_':
                 self.classes_ = np.array(v)
                 continue
-            if k == 'use_label_encoder':
-                self.use_label_encoder = bool(v)
-                continue
```
hcho3, line 579: Why is this line being removed?

python-package/xgboost/sklearn.py
```diff
@@ -542,27 +525,18 @@ def save_model(self, fname: str):
...
 
-        The model is loaded from an XGBoost internal format which is universal
-        among the various XGBoost interfaces. Auxiliary attributes of the
-        Python Booster object (such as feature names) will not be loaded.
-
-        Parameters
-        ----------
-        fname : string
-            Input file name.
-
```
hcho3, line 557: Can we keep the docstring?

hcho3 (changes requested):

python-package/xgboost/sklearn.py
trivialfis: We reuse the doc from `booster.load_model`.  See line 563

trivialfis: bool value is handled by JSON.

hcho3: Given the test, I take your word here.

hcho3 (approved):

RAMitchell (approved):

PR #6811, opened 2021-03-30
Fix approximated predict contribution.

trivialfis:
Add a prediction type for approximated interaction.

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -740,15 +740,17 @@ XGB_DLL int XGBoosterPredict(BoosterHandle handle,
  *
  * \param handle Booster handle
  * \param dmat   DMatrix handle
- * \param c_json_config String encoded predict configuration in JSON format.
+ * \param c_json_config String encoded predict configuration in JSON format, with
+ *                      following available fields in the JSON object:
  *
- *    "type": [0, 5]
+ *    "type": [0, 6]
  *      0: normal prediction
  *      1: output margin
  *      2: predict contribution
- *      3: predict approxmated contribution
+ *      3: predict approximated contribution
  *      4: predict feature interaction
- *      5: predict leaf
+ *      5: predict approximated feature interaction
+ *      6: predict leaf
  *    "training": bool
  *      Whether the prediction function is used as part of a training loop.  **Not used
  *      for inplace prediction**.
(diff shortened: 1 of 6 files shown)


include/xgboost/learner.h
```diff
@@ -36,7 +36,8 @@ enum class PredictionType : std::uint8_t {  // NOLINT
   kContribution = 2,
   kApproxContribution = 3,
   kInteraction = 4,
-  kLeaf = 5
+  kApproxInteraction = 5,
```
hcho3, line 39: Is this a new feature? I don't see the necessary change anywhere in the C++ codebase.

src/c_api/c_api_utils.h
```diff
@@ -71,6 +70,7 @@ inline void CalcPredictShape(bool strict_shape, PredictionType type, size_t rows
     }
     break;
   }
+  case PredictionType::kApproxInteraction:
```
RAMitchell, line 73: Should something be happening here?

RAMitchell (commented):
Did approximate feature interactions get disabled somewhere in a past refactor?

include/xgboost/c_api.h
```diff
@@ -740,15 +740,17 @@ XGB_DLL int XGBoosterPredict(BoosterHandle handle,
...
+ * \param c_json_config String encoded predict configuration in JSON format, with
+ *                      following available fields in the JSON object:
  *
- *    "type": [0, 5]
+ *    "type": [0, 6]
  *      0: normal prediction
  *      1: output margin
  *      2: predict contribution
- *      3: predict approxmated contribution
+ *      3: predict approximated contribution
```
RAMitchell, line 750: This is technically a breaking change right?

trivialfis:
@hcho3 @RAMitchell It's a bug I introduced during the refactoring of the prediction funciton.  I had an oversight that the `approximate` parameter is useful for feature interaction and there was no test for it.  This PR is just adding it back.  Since both changes are in the same release so it's not a breaking change.

hcho3 (approved):

src/c_api/c_api_utils.h
trivialfis: No, the output shape is exactly the same between shap and approx.

PR #6814, opened 2021-03-30
oneAPI DPC++-based plugin: Data Management

vepifanov:
Continuation of the original PRs: [5659](https://github.com/dmlc/xgboost/pull/5659), [6212](https://github.com/dmlc/xgboost/pull/6212).

Part of the oneAPI plugin, containing:
- common USMVector class for storing and synchronization of device memory based on DPC++ USM buffer;
- wrapper for DMatrix to store data on target device and avoid additional data synchronization on each iteration;
- USM-based version of ColumnMatrix to store columnar data on target device;
- USM-based version of RowSet collection to store rows indices on target device;

### DIFF
--- a/plugin/updater_oneapi/column_matrix_oneapi.h
+++ b/plugin/updater_oneapi/column_matrix_oneapi.h
@@ -1,0 +1,281 @@
+/*!
+ * Copyright 2017-2021 by Contributors
+ * \file column_matrix_oneapi.h
+ * \brief Utility for fast column-wise access for oneAPI plugin
+ */
+
+#ifndef XGBOOST_COMMON_COLUMN_MATRIX_ONEAPI_H_
+#define XGBOOST_COMMON_COLUMN_MATRIX_ONEAPI_H_
+
+#include <limits>
+#include <vector>
+#include <memory>
+#include "../../src/common/column_matrix.h"
+#include "hist_util_oneapi.h"
+
+namespace xgboost {
+namespace common {
+
+template <typename BinIdxType>
+class DenseColumnOneAPI: public Column<BinIdxType> {
+ public:
+  DenseColumnOneAPI(ColumnType type, common::Span<const BinIdxType> index,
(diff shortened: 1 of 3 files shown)


trivialfis:
Thanks for continuing the work on this.  Please remove `[WIP]` tag if it's ready for review.

trivialfis:
Sorry, didn't notice you have changed the title (I usually watch the notification from emails, but this doesn't trigger one).  Is there any plan on how to choose the device from Python API?  Like the current `gpu_id` in xgboost?  We can think about implementation later, but it seems to be a good time to talk about basic interfaces like choosing devices.

vepifanov:
In the 3rd pull request (predictor), the "device_id" (name is debatable) is added to GenericParameter, so you can set it from python.
It works similarly to "gpu_id", with regarding to support both CPU/GPU dispatching:
- -1 - means the default strategy of choosing device, usually the priority is SYCL-supporting GPU (if presented) -> then CPU;
- if the device under the "device_id" index is CPU - fallback to CPU updater (grow_quantile_histmaker);
- if the device under the "device_id" index is GPU - go to new oneAPI_GPU updater;

trivialfis:
Sorry, I lost track of which PR is the third PR.  I just grep the plugin directory and couldn't find the token `device_id`.  I may be able to help with the device id part but I need to see some working examples.

vepifanov:
You can find introduction of device_id here - [https://github.com/vepifanov/xgboost/pull/3](https://github.com/vepifanov/xgboost/pull/3).

There is also this [branch](https://github.com/vepifanov/xgboost/tree/release_1.3.0_oneapi) which was used for creating internal package, so it should contain working version of the plugin (including device_id parameter).

trivialfis:
@hcho3 @RAMitchell please help take a look into the latest progress here. ;-)

trivialfis:
Closing as there's a new push.

PR #6817, opened 2021-03-31
More general predict proba.

trivialfis:
* Use `output_margin` for `softmax`.
* Add test for dask binary cls.
* Remove unused output margin.

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -534,6 +534,8 @@ class XGBModel(XGBModelBase):
         self.get_booster().load_model(fname)
         meta = self.get_booster().attr('scikit_learn')
         if meta is None:
+            # FIXME(jiaming): This doesn't have to be a problem as most of the needed
+            # information like num_class and objective is in Learner class.
             warnings.warn(
                 'Loading a native XGBoost model with Scikit-Learn interface.'
             )
@@ -545,6 +547,8 @@ class XGBModel(XGBModelBase):
                 self._le = XGBoostLabelEncoder()
                 self._le.from_json(v)
                 continue
+            # FIXME(jiaming): This can be removed once label encoder is gone since we can
+            # generate it from `np.arange(self.n_classes_)`
             if k == 'classes_':
                 self.classes_ = np.array(v)
                 continue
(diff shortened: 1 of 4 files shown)


trivialfis:
Marked as blocking for 2 reasons.
- I found a weird case that the output shape is incorrectly inferred by dask.  I'm not sure what's the cause yet.  This happens when output prediction is 2-dim (multi-class), which has the same shape as input data and `map_blocks` seems to be ignoring the change of output shape (from n_features to n_classes).  I don't know how exactly is it triggered since before this PR it works fine.  Still investigating.
- `multi:softmax` can be used safely with classifier, which seems to be a nice thing to have.

python-package/xgboost/sklearn.py
```diff
@@ -1262,15 +1263,23 @@ def predict_proba(
             a numpy array of shape array-like of shape (n_samples, n_classes) with the
             probability of each data example being of a given class.
         """
+        # custom obj:      Do nothing as we don't know what to do.
+        # softprob:        Do nothing, output is proba.
+        # softmax:         Use output margin to remove the argmax in PredTransform.
+        # binary:logistic: Expand the prob vector into 2-class matrix after predict.
+        # binary:logitraw: Unsupported, let's deprecate this objective when possible.
```
hcho3, line 1270: I've seen at least a few users using this objective. Should we actually deprecate it?

hcho3 (approved):

RAMitchell (approved):

python-package/xgboost/sklearn.py
trivialfis: It's not clear to me in what scenario is it useful.

trivialfis: If you want to have an untransformed prediction, use `output_margin` instead.

hcho3: So far we've avoided breaking existing model files, i.e. you could load a very old model file and run predictions. Removing support for `binary:logitraw` will make old models files inaccessible. We don't yet have a formal policy for changes affecting old model files. (e.g. Wait 3 major versions?)

trivialfis: I removed the note.

python-package/xgboost/sklearn.py
```diff
@@ -1238,15 +1241,23 @@ def predict_proba(
             a numpy array of shape array-like of shape (n_samples, n_classes) with the
             probability of each data example being of a given class.
         """
+        # custom obj:      Do nothing as we don't know what to do.
+        # softprob:        Do nothing, output is proba.
+        # softmax:         Use output margin to remove the argmax in PredTransform.
+        # binary:logistic: Expand the prob vector into 2-class matrix after predict.
+        # binary:logitraw: Unsupported.
```
hcho3, line 1248:
```suggestion
        # binary:logitraw: Unsupported by predict_proba() 
```

Issue #6822, opened 2021-04-02
[Bug] DeviceQuantileDmatrix failing with memory errors on A100-80GB


ayushdg:
On large datasets, DeviceQuantileDMatrix fails with memory errors that look like some kind of overflow/ptr bug.

Reproducer:
```python
import cudf
import numpy as np
import xgboost as xgb

nrows=51130552
ncols=43
df = cudf.DataFrame()
for i in range(ncols-1):
    df[f"fea{i}"] = np.random.random(nrows)
    df[f"fea{i}"] = df[f"fea{i}"].astype("float32")
df['val'] = [0,1] * (nrows//2)
df['val'] = df['val'].astype("int8")

fea_cols=df.columns.difference(["val"])
dmat = xgb.DeviceQuantileDMatrix(data=df[fea_cols], label=df["val"])
del(df)

dxgb_gpu_params = {
    "max_depth": 8,
    "max_leaves": 2 ** 8,
    "alpha": 0.9,
    "eta": 0.1,
    "gamma": 0.1,
    "learning_rate": 0.1,
    "subsample": 1,
    "reg_lambda": 1,
    "tree_method": "gpu_hist",
    "objective": "binary:logistic",
}
NUM_BOOST_ROUND = 5


res = xgb.train(
    dxgb_gpu_params, dmat, num_boost_round=NUM_BOOST_ROUND,
)
```

The following scripts succeeds on a `V100-32GB` and `A100-40GB` but fails on `A100-80GB`

Stacktrace:
```
Traceback (most recent call last):
  File "xgb_error.py", line 19, in <module>
    dmat = xgb.DeviceQuantileDMatrix(data=df[fea_cols], label=df["val"])
  File "/opt/conda/envs/rapids/lib/python3.8/site-packages/xgboost/core.py", line 946, in __init__
    handle, feature_names, feature_types = init_device_quantile_dmatrix(
  File "/opt/conda/envs/rapids/lib/python3.8/site-packages/xgboost/data.py", line 787, in init_device_quantile_dmatrix
    _check_call(ret)
  File "/opt/conda/envs/rapids/lib/python3.8/site-packages/xgboost/core.py", line 189, in _check_call
    raise XGBoostError(py_str(_LIB.XGBGetLastError()))
[...]

PR #6826, opened 2021-04-06
Use batched copy if.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/6822 .

### DIFF
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -1290,6 +1290,21 @@ void InclusiveScan(InputIteratorT d_in, OutputIteratorT d_out, ScanOpT scan_op,
                                            num_items, nullptr, false)));
 }
 
+template <typename InIt, typename OutIt, typename Predicate>
+void CopyIf(InIt in_first, InIt in_second, OutIt out_first, Predicate pred) {
+  // We loop over batches because thrust::copy_if cant deal with sizes > 2^31
+  // See thrust issue #1302, #6822
+  size_t max_copy_size = std::numeric_limits<int>::max() / 2;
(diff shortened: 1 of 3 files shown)


RAMitchell (approved):

src/common/device_helpers.cuh
```diff
@@ -1290,6 +1290,21 @@ void InclusiveScan(InputIteratorT d_in, OutputIteratorT d_out, ScanOpT scan_op,
                                            num_items, nullptr, false)));
 }
 
+template <typename InIt, typename OutIt, typename Predicate>
+void CopyIf(InIt in_first, InIt in_second, OutIt out_first, Predicate pred) {
+  // We loop over batches because thrust::copy_if cant deal with sizes > 2^31
+  // See thrust issue #1302, #6822
+  size_t max_copy_size = std::numeric_limits<int>::max() / 2;
```
vkuzmin-uber, line 1297: Not sure what standard current compiler supports, but this can be a 'constexpr', since max is. See example:

https://en.cppreference.com/w/cpp/language/constant_expression

vkuzmin-uber: I see "XGBoost now adopts the C++14 standard", it should be allowed.

vkuzmin-uber (commented):
I see it is merged already. Leave comment just in case - constexpr for expressions like this makes sense

src/common/device_helpers.cuh
trivialfis: @vkuzmin-uber Thanks for the review, I will open another PR for constexpr.  Most of the time the compiler should be able to eliminate these constants automatically so did not pay enough attention to it, thanks for pointing it out.

Issue #6791, opened 2021-03-26
[CI] Upgrade to CentOS 7 image for building libxgboost.so


hcho3:
As noted in #6467, CentOS 6 reached end-of-life on Nov 30, 2020 and thus will no longer receive security updates. We should upgrade the worker image to CentOS 7.

**What's the impact of this change?**
* Currently we distribute a pre-built binary wheel for the Linux platform which has `manylinux2010` tag. The tag will need to change to `manylinux2014`. Users will need to use Pip 19.3 or later.
* We will be dropping support for old Linux distributions such as CentOS 6 or RHEL 6.

hcho3:
Will re-visit this after the 1.4.0 release.

PR #6827, opened 2021-04-06
[CI] Pack R package tarball with pre-built xgboost.so (with GPU support)

### LINKED ISSUE #6654: Prebuilt R Package Binaries with GPU Support
More attention is being paid to R on GPU so it would be great to have prebuilt binaries. I understand they cannot be hosted on CRAN, but many users are comfortable using `devtools::install_url()`. So if you could host it on the [releases](https://github.com/dmlc/xgboost/releases) page that would be a big help for two reasons: Many users are not comfortable using cmake. And, more importantly, more and more users use a library manager like `{renv}`, meaning that packages need to be rebuilt for each project. If a binary were available it could be reused for each.

`{catboost}` uses [this approach](https://catboost.ai/docs/installation/r-installation-binary-installation.html) for installing the R package from their [releases page](https://github.com/catboost/catboost/releases).

If I am reading the [installation file](https://github.com/dmlc/xgboost/blob/master/cmake/RPackageInstall.cmake.in#L34) properly, it looks like binary a file is built. Perhaps CI can build this and put the binary on the releases page?

Of course, finding a way to get the GPU version on CRAN is ideal, but I don't know how hard that is. Though there have been other packages on CRAN to make use of the GPU so I imagine it is possible.

Sorry if this has already been requested. I searched the issues but didn't quite see anything about this.

hcho3:
Closes #6654

Usage. Users will now be able to install XGBoost with the following simple commands:
```bash
# Install dependencies
R -q -e "install.packages(c('data.table', 'magrittr', 'jsonlite', 'remotes'))"
# Install XGBoost
R CMD INSTALL ./xgboost.tar.gz
```

No need to use `cmake` or suffer setup issues from `nvcc`. It just works.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -54,7 +54,21 @@ pipeline {
       steps {
         script {
           parallel ([
+            'clang-tidy': { ClangTidy() },
+            'build-cpu': { BuildCPU() },
+            'build-cpu-arm64': { BuildCPUARM64() },
+            'build-cpu-rabit-mock': { BuildCPUMock() },
+            // Build reference, distribution-ready Python wheel with CUDA 10.0
+            // using CentOS 6 image
+            'build-gpu-cuda10.0': { BuildCUDA(cuda_version: '10.0') },
+            // The build-gpu-* builds below use Ubuntu image
+            'build-gpu-cuda10.1': { BuildCUDA(cuda_version: '10.1') },
+            'build-gpu-cuda10.2': { BuildCUDA(cuda_version: '10.2', build_rmm: true) },
+            'build-gpu-cuda11.0': { BuildCUDA(cuda_version: '11.0') },
             'build-gpu-rpkg': { BuildRPackageWithCUDA(cuda_version: '10.0') }
+            'build-jvm-packages-gpu-cuda10.0': { BuildJVMPackagesWithCUDA(spark_version: '3.0.0', cuda_version: '10.0') },
+            'build-jvm-packages': { BuildJVMPackages(spark_version: '3.0.0') },
+            'build-jvm-doc': { BuildJVMDoc() }
           ])
         }
       }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -271,12 +271,14 @@ def BuildRPackageWithCUDA(args) {
     def container_type = 'gpu_build_r_centos6'
     def docker_binary = "docker"
     def docker_args = "--build-arg CUDA_VERSION_ARG=10.0"
-    sh """
-    ${dockerRun} ${container_type} ${docker_binary} ${docker_args} tests/ci_build/build_r_pkg_with_cuda.sh ${commit_id}
-    """
-    echo 'Uploading R tarball...'
-    path = ("${BRANCH_NAME}" == 'master') ? '' : "${BRANCH_NAME}/"
-    s3Upload bucket: 'xgboost-nightly-builds', path: path, acl: 'PublicRead', includePathPattern:'xgboost_r_gpu_linux_*.tar.gz'
+    if (env.BRANCH_NAME == 'master' || env.BRANCH_NAME.startsWith('release')) {
+      sh """
+      ${dockerRun} ${container_type} ${docker_binary} ${docker_args} tests/ci_build/build_r_pkg_with_cuda.sh ${commit_id}
+      """
+      echo 'Uploading R tarball...'
+      path = ("${BRANCH_NAME}" == 'master') ? '' : "${BRANCH_NAME}/"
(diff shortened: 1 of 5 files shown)


tests/ci_build/Dockerfile.gpu_build_r_centos6
```diff
@@ -0,0 +1,106 @@
...
+ENV CPP=/opt/rh/devtoolset-4/root/usr/bin/cpp
+ENV F77=/opt/rh/devtoolset-4/root/usr/bin/gfortran
+
+RUN \
+    wget https://zlib.net/fossils/zlib-1.2.5.tar.gz  && \
+    wget https://sourceware.org/pub/bzip2/bzip2-1.0.6.tar.gz  && \
+    wget http://tukaani.org/xz/xz-5.2.2.tar.gz  && \
+    wget https://ftp.pcre.org/pub/pcre/pcre-8.40.tar.gz  && \
+    wget https://www.openssl.org/source/old/1.0.0/openssl-1.0.0k.tar.gz  && \
+    wget --no-check-certificate https://curl.se/download/curl-7.47.1.tar.gz  && \
```
hcho3, line 38: A few packages have to be built from the source because CentOS 6 is a very old distribution and the system packages are not sufficiently up-to-date to build R 3.3.0. We'll want to update to CentOS 7 after the 1.4.0 release. Tracking issue: #6791.

Jenkinsfile
```diff
@@ -264,6 +265,21 @@ def BuildCUDA(args) {
...
+    unstash name: 'srcs'
+    def container_type = 'gpu_build_r_centos6'
+    def docker_binary = "docker"
+    def docker_args = "--build-arg CUDA_VERSION_ARG=10.0"
+    sh """
+    ${dockerRun} ${container_type} ${docker_binary} ${docker_args} tests/ci_build/build_r_pkg_with_cuda.sh
+    """
+    echo 'Uploading Python wheel...'
+    path = ("${BRANCH_NAME}" == 'master') ? '' : "${BRANCH_NAME}/"
+    s3Upload bucket: 'xgboost-nightly-builds', path: path, acl: 'PublicRead', includePathPattern:'xgboost_r_gpu_linux_*.tar.gz'
```
hcho3, line 279: This section should not execute for pull requests. For now, I'll let this run so as to ensure that the package actually gets uploaded to the bucket correctly.

Jenkinsfile
```diff
@@ -264,6 +265,22 @@ def BuildCUDA(args) {
...
+    unstash name: 'srcs'
+    def container_type = 'gpu_build_r_centos6'
+    def docker_binary = "docker"
+    def docker_args = "--build-arg CUDA_VERSION_ARG=10.0"
+    sh """
+    ${dockerRun} ${container_type} ${docker_binary} ${docker_args} tests/ci_build/build_r_pkg_with_cuda.sh ${commit_id}
+    """
+    echo 'Uploading R tarball...'
+    path = ("${BRANCH_NAME}" == 'master') ? '' : "${BRANCH_NAME}/"
+    s3Upload bucket: 'xgboost-nightly-builds', path: path, acl: 'PublicRead', includePathPattern:'xgboost_r_gpu_linux_*.tar.gz'
```
hcho3, line 279: This section should not execute for pull requests. For now, I'll let this run so as to ensure that the package actually gets uploaded to the bucket correctly.

tests/ci_build/Dockerfile.gpu_build_r_centos6
hcho3: Why choose an old Linux distro? This is so that the resulting `libxgboost.so` is compatible with a wide range of Linux OSes currently in operation. See https://www.python.org/dev/peps/pep-0571/

tests/ci_build/build_r_pkg_with_cuda.sh
```diff
@@ -0,0 +1,30 @@
...
+cd ..
+
+rm xgboost
+rm -v xgboost_rpack/configure
+rm -rfv xgboost_rpack/src
+mkdir -p xgboost_rpack/src
+cp -v lib/xgboost.so xgboost_rpack/src/
+echo 'all:' > xgboost_rpack/src/Makefile
+echo 'all:' > xgboost_rpack/src/Makefile.win
+mv xgboost_rpack/ xgboost/
```
hcho3, line 29: This super wacky hack is found in https://github.com/dmlc/xgboost/blob/master/cmake/RPackageInstall.cmake.in and https://github.com/dmlc/xgboost/blob/master/cmake/RPackageInstallTargetSetup.cmake . This hack lets us bypass the normal build process of R and have R use `xgboost.so` that we've already built.

trivialfis (approved):
Thanks for automating the process!  Would you like to add the comments in code instead of GitHub comment?  Also please help update the document for installation.

PR #6835, opened 2021-04-07
Throw error when using unsupported objective with `predict_proba`.

trivialfis:
This partially reverts https://github.com/dmlc/xgboost/pull/6817 on the use of `softmax`.  Looking into it more thoroughly, the `softmax` function can never appear in training due to automatic configuration in `fit`.  But the objective can still be invalid for predict_proba (like `softmax`) as illustrated in the test as an edge case.  Previously I let those functions pass and may return an incorrect result.  This PR reverts that part and throws a ValueError.  Other than that fixes in linked PR are still useful and kept.


trivialfis:
Closing for now.  Handling this edge case might require more efforts.

Issue #6806, opened 2021-03-30
xgboost on spark version 0.9 got Exception (java.util.NoSuchElementException,key not found: train)


liuyonglang:
FYI，when i use xgboost on spark 0.9 version to fit a model ,wo get this Exception(java.util.NoSuchElementException,key not found: train)。what condition can lead to the problem?  Thanks for your attention
follow is my paramters:

XGBoostSpark: Running XGBoost 0.90 with parameters:
alpha -> 0.29174557458107453
min_child_weight -> 2.7498486288961455
sample_type -> uniform
base_score -> 0.5
colsample_bylevel -> 0.7423861022776388
grow_policy -> lossguide
skip_drop -> 0.0
lambda_bias -> 0.20602638227615486
silent -> 0
scale_pos_weight -> 1.0
seed -> 12345
cache_training_set -> false
features_col -> importance_vec
num_early_stopping_rounds -> 0
label_col -> C25
num_workers -> 2
subsample -> 1.0
lambda -> 1.0
max_depth -> 4
probability_col -> probability
raw_prediction_col -> rawPrediction
tree_limit -> 0
custom_eval -> null
rate_drop -> 0.0
max_bin -> 16
train_test_ratio -> 0.9
use_external_memory -> false
objective -> binary:logistic
eval_metric -> logloss
num_round -> 23
timeout_request_workers -> 1800000
missing -> 0.0
checkpoint_path -> 
tracker_conf -> TrackerConf(0,python)
tree_method -> hist
max_delta_step -> 0.0
eta -> 0.001200144733197779
verbosity -> 1
colsample_bytree -> 0.6367027366379028
normalize_type -> tree
custom_obj -> null
gamma -> 0.0
sketch_eps -> 0.03
nthread -> 1
prediction_col -> prediction
checkpoint_interval -> -1

trivialfis:
Could you please provide a reproducible example?

liuyonglang:
> Could you please provide a reproducible example?

it is like this:
https://github.com/dmlc/xgboost/pull/4380#issuecomment-486047797.
but I dont found what the solution finally.

wbo4958:
@liuyonglang Can you try the latest xgboost release and see if the issue is still there, if yes, pls provide your reproducible sample. thx

liuyonglang:
> @liuyonglang Can you try the latest xgboost release and see if the issue is still there, if yes, pls provide your reproducible sample. thx

[...]

PR #6842, opened 2021-04-11
[jvm-packages] fix "key not found: train" issue

wbo4958:


wbo4958:
to fix https://github.com/dmlc/xgboost/issues/6806

trivialfis:
@CodingCat Could you please help take a look?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -377,7 +377,7 @@ object XGBoost extends Serializable {
     // to workaround the empty partitions in training dataset,
     // this might not be the best efficient implementation, see
     // (https://github.com/dmlc/xgboost/issues/1277)
-    if (watches.toMap("train").rowNum == 0) {
```
CodingCat, line 380: is it possible that there is "train' as the key but it is empty?

I think we'd better keep both

wbo4958: toMap will filter out DMatrix with rowNum <= 0,  https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala#L762

PR #6848, opened 2021-04-13
avoid hang when some workers fail but spark context still live

7haya:
sparkJobThread.start will start the process of applying for resources and submitting jobs
parallelismTracker.execute will add TaskFailedListener to SparkContext
If a Job is submitted before TaskFailedListener is added to SparkContext, it will not be killed when the event TaskFailedListener.onTaskEnd is executed, because it is not added to jobIdToStageIds when it is started. When Track starts again, it will get stuck
My editor ensures that sparkJobThread.start will not be started until TaskFailedListener is added to SparkContext.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -612,11 +612,17 @@ object XGBoost extends Serializable {
             boostersAndMetrics.foreachPartition(() => _)
           }
         }
+        val listener = new TaskFailedListener(xgbExecParams.killSparkContextOnWorkerFailure)
+        sc.addSparkListener(listener)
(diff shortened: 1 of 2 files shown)


CodingCat:
I may be too lazy to think with more details, since the first two lines of safeExecute are adding listener, and the job is executed by https://github.com/dmlc/xgboost/pull/6848/files#diff-c634bdecc113326bfe308870cbd33c3967fa780ca632707b7d252893f8e22f31L618-L619 which refers as "body" in safeExecute function

how the job can be submitted before TaskFailedListener is added to SparkContext?

7haya:
> I may be too lazy to think with more details, since the first two lines of safeExecute are adding listener, and the job is executed by https://github.com/dmlc/xgboost/pull/6848/files#diff-c634bdecc113326bfe308870cbd33c3967fa780ca632707b7d252893f8e22f31L618-L619 which refers as "body" in safeExecute function
> 
> how the job can be submitted before TaskFailedListener is added to SparkContext?

OK...I modified this part of the code based on the old version of the commit, then resolved the conflict and submitted a pull request
The old version of commit will call sparkJobThread.start() before calling parallelismTracker.execute()

PR #6854, opened 2021-04-13
[CI] Upload Doxygen doc to correct destination

hcho3:
The link for the C API doc was not working: https://xgboost.readthedocs.io/en/stable/dev/c__api_8h.html. When we ported the Doxygen build from Jenkins to GitHub Actions, we made a small typo that broke the C API doc.

This commit has already been ported to `release_1.3.0` and `release_1.4.0` and now the links are back online:
* https://xgboost.readthedocs.io/en/release_1.3.0/dev/c__api_8h.html
* https://xgboost.readthedocs.io/en/release_1.4.0/dev/c__api_8h.html

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -279,7 +279,7 @@ jobs:
       run: |
         cd build/
         tar cvjf ${{ steps.extract_branch.outputs.branch }}.tar.bz2 doc_doxygen/
-        python -m awscli s3 cp ./${{ steps.extract_branch.outputs.branch }}.tar.bz2 s3://xgboost-docs/ --acl public-read
+        python -m awscli s3 cp ./${{ steps.extract_branch.outputs.branch }}.tar.bz2 s3://xgboost-docs/doxygen/ --acl public-read
       if: github.ref == 'refs/heads/master' || contains(github.ref, 'refs/heads/release_')
       env:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -279,7 +279,7 @@ jobs:
       run: |
         cd build/
         tar cvjf ${{ steps.extract_branch.outputs.branch }}.tar.bz2 doc_doxygen/
-        python -m awscli s3 cp ./${{ steps.extract_branch.outputs.branch }}.tar.bz2 s3://xgboost-docs/ --acl public-read
+        python -m awscli s3 cp ./${{ steps.extract_branch.outputs.branch }}.tar.bz2 s3://xgboost-docs/doxygen/ --acl public-read
```
hcho3, line 282: ReadTheDocs doesn't have Doxygen installed on their build server, so here we build Doxygen in GitHub Actions and then upload the tarball to the `xgboost-docs` S3 bucket. The sphinx build script expects to find the tarball at a fixed location. See https://github.com/dmlc/xgboost/blob/b9a4f3336af7c78f9f7b66f7c91a419108b7ec83/doc/conf.py#L36-L55

trivialfis: Unrelated to this PR.  But the `call` to shell code can be completely replaced by Python.

trivialfis (approved):

PR #6855, opened 2021-04-13
Removed magrittr dependency in callbacks.R (closes #6852)

### LINKED ISSUE #6852: Remove magrittr dependency in the R documentation for cb.gblinear.history
The documentation for `cb.gblinear.history`, generated [here](https://github.com/dmlc/xgboost/blob/master/R-package/R/callbacks.R), has a dependency on magrittr. While this is not a particular obscure library, it seems strange that we assume familiarity with it. Furthermore, the six calls to the library that I've found in the documentation all appear to be trivial. If I'm not mistaken, they're all of the form `xgb.gblinear.history(foo) %>% matplot(type = 'l')`. This should be trivial to express without magrittr. In fact, earlier on in the documentation, what I believe to be one such equivalent call appears:
```R
coef_path <- xgb.gblinear.history(bst)
matplot(coef_path, type = 'l')
```
So why not just get rid of the magrittr dependency in the documentation, even if just for consistency's sake? I don't believe that it's used in any of our other documentation.

As a final point, I believe that the next version of R is getting its own pipe operator, so even if we do want to use pipes in our documentation, we soon won't need magrittr's.

I should have a pull request ready for this tonight. It seems like a trivial job to do.

ReeceGoding:
To close issue #6852, I've removed the magrittr dependency in the relevant documentation. The simple path of replacing each call like `xgb.gblinear.history(bst) %>% matplot(type = 'l')` with its corresponding `matplot(xgb.gblinear.history(bst), type = 'l')` was taken. For some calls, I get different results when I run the code before and after this change. However, there is sufficient randomness in said code that I get different results each time that I run it, regardless of any changes.

One thing that definitely has changed is that the plot y-axes now have titles matching the argument that they are called with. For example, where we previously had no titles on our y-axes that were generated using magrittr pipes, we now have something like "xgb.gblinear.history(bst)". If this is undesirable (and I don't see why it would be?), then the most consistent solution is to follow the `coef_path <- xgb.gblinear.history(bst)` example, which named its y-axis "coef_path" as a side-effect of the assignment just shown.

Also corrected suspected typo: "forld".

### DIFF
--- a/R-package/R/callbacks.R
+++ b/R-package/R/callbacks.R
@@ -560,7 +560,6 @@ cb.cv.predict <- function(save_models = FALSE) {
 #' #
 #' # In the iris dataset, it is hard to linearly separate Versicolor class from the rest
 #' # without considering the 2nd order interactions:
-#' require(magrittr)
 #' x <- model.matrix(Species ~ .^2, iris)[,-1]
 #' colnames(x)
 #' dtrain <- xgb.DMatrix(scale(x), label = 1*(iris$Species == "versicolor"))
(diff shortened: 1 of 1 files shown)


trivialfis (approved):
Looks good to me.  Thanks for the PR!

PR #6884, opened 2021-04-21
[docs] Add tutorial for XGBoost-Ray

krfricke:
This PR adds a new tutorial for XGBoost-Ray.

The code samples are included and unit tested in the XGBoost-Ray repository.

I added xgboost_ray as a dependency to sphinx so that the documentation of the RayParams can automatically be fetched from the latest XGBoost-Ray release. Let me know if this is fine

cc @trivialfis @hcho3

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -7,3 +7,4 @@ matplotlib>=2.1
 graphviz
 numpy
 recommonmark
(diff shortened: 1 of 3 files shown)


trivialfis (approved):
Thanks for the well written tutorial!

PR #6889, opened 2021-04-21
Ensure predict leaf output 1-dim vector where there's only 1 tree.

trivialfis:
Before the output shape for predict leaf with 1 tree is `(n_samples, )`, after 1.4 it's `(n_samples, 1)`.  The shape for other cases like more than 1 num_boost_round or more than 1 class/num_parallel_tree is unchanged.

We added a parameter called `strict_shape` in 1.4 to get consistent shape across all model types and I recommend using it: https://xgboost.readthedocs.io/en/latest/prediction.html .  When specified, predict leaf always output a 4 dim tensor with clear meaning for each dim.

### DIFF
--- a/doc/prediction.rst
+++ b/doc/prediction.rst
@@ -54,8 +54,8 @@ After 1.4 release, we added a new parameter called ``strict_shape``, one can set
   Output is a 4-dim array with ``(n_samples, n_iterations, n_classes, n_trees_in_forest)``
   as shape.  ``n_trees_in_forest`` is specified by the ``numb_parallel_tree`` during
   training.  When strict shape is set to False, output is a 2-dim array with last 3 dims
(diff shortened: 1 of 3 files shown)


hcho3:
Should this also be part of 1.4.1 patch release?

trivialfis:
I don't have a strong preference.   This seems to be a corner case, where you need to run predict leaf on 1 tree, and use the output as panda input directly.  I can make a small patch to pip if needed.  No other language binding is affected.

trivialfis:
I haven't looked into the xgbes in detail yet, not sure how does it handle situations like having `num_parallel_tree` greater than 1. (I think it might break).  So in general I still suggest others use the `strict_shape` parameter and make sure you are getting what you want.

hcho3 (approved):

PR #6891, opened 2021-04-22
Add an API guard to prevent global variables being changed.

trivialfis:
Fixes the failure on CI where other libraries like cupy and cudf don't necessarily handle the non-default device.

Also, I think it's a good idea to prevent xgboost from changing any global variable in runtimes.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -1939,8 +1939,11 @@ class Booster(object):
                 )
             )
             return _prediction_output(shape, dims, preds, False)
-        if lazy_isinstance(data, "cupy.core.core", "ndarray"):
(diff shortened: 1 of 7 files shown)


python-package/xgboost/core.py
```diff
@@ -1939,8 +1939,11 @@ def inplace_predict(
                 )
             )
             return _prediction_output(shape, dims, preds, False)
-        if lazy_isinstance(data, "cupy.core.core", "ndarray"):
+        if lazy_isinstance(data, "cupy.core.core", "ndarray") or lazy_isinstance(
```
trivialfis, line 1942: This is due to a cupy upgrade on CI.

RAMitchell (approved):

PR #6901, opened 2021-04-23
[CI] Skip external memory gtest on osx.

trivialfis:
On updated libomp@12, the external memory test causes internal error in libomp:

```
1: OMP: Error #131: Thread identifier invalid.
```

So far non of us can reproduce it on outside of CI environment.  This PR tried to switch different xcode versions but all ran into various failures.  So to unblock the CI I just disable 1 c++ test.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,9 +19,6 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: 'true'
-    - uses: maxim-lobanov/setup-xcode@v1
-      with:
-        xcode-version: "10.3"
     - name: Install system packages
       run: |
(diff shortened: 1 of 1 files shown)


trivialfis:
Notes:
- 12.4 is not compatible with libomp 12.0.0
- 11.7 crashes.

trivialfis:
@hcho3  Okay, all 3 major versions are failing.  Since we can't fix the libomp version from homebrew, nor does switching xcode version work.  The remaining options are:
- switch to gcc or open-source clang.
- switch back to travis and try again.

hcho3 (approved):

RAMitchell (approved):

Issue #6896, opened 2021-04-23
R: models can get de-serialized redundantly multiple times


david-cortes:
In the R version, the booster model objects store the underlying C++ objects as raw bytes plus a pointer to an in-memory object.

When one de-serializes a booster object (e.g. by restarting the R session or through `saveRDS` and `readRDS`), the pointer then becomes null, and when the time comes to use the C++ object, the object gets reconstructed from the raw bytes.

However, the pointer does not get updated in the original R object, which means the C++ object will get restored on each function call which is slow and uses more memory. Example:
```r
library(xgboost)
data(mtcars)
xgb_data <- xgb.DMatrix(as.matrix(mtcars[,-1]), label=mtcars[,1,drop=TRUE])
model <- xgb.train(list(objective="reg:squarederror"), xgb_data, nrounds=10)
saveRDS(model, file.path(tempdir(), "model.rds"))
model_new <- readRDS(file.path(tempdir(), "model.rds"))
print(model_new$handle)
```
```
<pointer: (nil)>
attr(,"class")
[1] "xgb.Booster.handle"
```
```r
pred <- predict(model_new, xgb_data)
print(model_new$handle)
```
```
<pointer: (nil)>
attr(,"class")
[1] "xgb.Booster.handle"
```

Would be better if this pointer were to be put in an R environment (e.g. `model$cpp_obj <- as.environment(list(handle=a_restored_handle))`), as then it would be updated by reference instead, keeping the pointer to the de-serialized object in the same object from which a function like `predict` is called and de-serializing only once during the object's lifetime in the R session.

hcho3:
You can call `.complete()` method in order to amortize the cost of deserialization: https://www.rdocumentation.org/packages/xgboost/versions/1.4.1.1/topics/xgb.Booster.complete

Example:
```R
bst <- readRDS("xgb.model.rds")
bst <- xgb.Booster.complete(bst) 
# Now use bst without incurring the cost of deserialization
```

david-cortes:
> You can call `xgb.Booster.complete()` method in order to amortize the cost of deserialization: https://www.rdocumentation.org/packages/xgboost/versions/1.4.1.1/topics/xgb.Booster.complete
> 
> Example:
> 
> ```r
> bst <- readRDS("xgb.model.rds")
> bst <- xgb.Booster.complete(bst) 
> # Now use bst without incurring the cost of deserialization
> ```

[...]

PR #6903, opened 2021-04-24
Update R handles in-place

david-cortes:
Addressing issue https://github.com/dmlc/xgboost/issues/6896

When de-serializing an object in R from raw bytes, the R handles become null as there is no C++ object to which to point to in-memory. In these situations, the C++ objects are recreated from raw bytes as needed, but the R handle objects are not modified in-place, which can lead to de-serializing multiple times.

This PR changes it so that R external pointer objects would be modified in-place in order to avoid redundant operations and memory usage. After the PR:
```r
library(xgboost)
data(mtcars)
xgb_data <- xgb.DMatrix(as.matrix(mtcars[,-1]), label=mtcars[,1,drop=TRUE])
model <- xgb.train(list(objective="reg:squarederror"), xgb_data, nrounds=10)
saveRDS(model, file.path(tempdir(), "model.rds"))
model_new <- readRDS(file.path(tempdir(), "model.rds"))
```
```r
print(model_new$handle)
```
```
<pointer: (nil)>
attr(,"class")
[1] "xgb.Booster.handle"
```
```r
pred <- predict(model_new, xgb_data)
print(model_new$handle)
```
```
<pointer: 0x55609a7541f0>
attr(,"class")
[1] "xgb.Booster.handle"
```

### DIFF
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -1,7 +1,7 @@
 # Construct an internal xgboost Booster and return a handle to it.
 # internal utility function
 xgb.Booster.handle <- function(params = list(), cachelist = list(),
-                               modelfile = NULL) {
+                               modelfile = NULL, handle = NULL) {
   if (typeof(cachelist) != "list" ||
       !all(vapply(cachelist, inherits, logical(1), what = 'xgb.DMatrix'))) {
     stop("cachelist must be a list of xgb.DMatrix objects")
@@ -20,7 +20,7 @@ xgb.Booster.handle <- function(params = list(), cachelist = list(),
       return(handle)
     } else if (typeof(modelfile) == "raw") {
       ## A memory buffer
-      bst <- xgb.unserialize(modelfile)
+      bst <- xgb.unserialize(modelfile, handle)
       xgb.parameters(bst) <- params
       return (bst)
     } else if (inherits(modelfile, "xgb.Booster")) {
@@ -129,7 +129,7 @@ xgb.Booster.complete <- function(object, saveraw = TRUE) {
     stop("argument type must be xgb.Booster")
 
   if (is.null.handle(object$handle)) {
-    object$handle <- xgb.Booster.handle(modelfile = object$raw)
+    object$handle <- xgb.Booster.handle(modelfile = object$raw, handle = object$handle)
   } else {
     if (is.null(object$raw) && saveraw) {
       object$raw <- xgb.serialize(object$handle)
--- a/R-package/R/xgb.unserialize.R
+++ b/R-package/R/xgb.unserialize.R
@@ -4,7 +4,6 @@
 #' @param handle An \code{xgb.Booster.handle} object which will be overwritten with
 #' the new deserialized object. Must be a null handle (e.g. when loading the model through
 #' `readRDS`). If not provided, a new handle will be created.
-#' 
 #' @return An \code{xgb.Booster.handle} object.
 #'
 #' @export
--- a/R-package/R/xgb.unserialize.R
+++ b/R-package/R/xgb.unserialize.R
@@ -3,9 +3,13 @@
 #' @param buffer the buffer containing booster instance saved by \code{\link{xgb.serialize}}
 #'
 #' @export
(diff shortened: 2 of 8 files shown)


david-cortes:
All R tests now passing. The C++ failing test is probably not related to this PR.

trivialfis:
Thanks, I will look into it.  Please ignore the osx test.  The Apple clang is not happy with the new libomp.

R-package/R/xgb.Booster.R
```diff
@@ -129,7 +129,7 @@ xgb.Booster.complete <- function(object, saveraw = TRUE) {
     stop("argument type must be xgb.Booster")
 
   if (is.null.handle(object$handle)) {
-    object$handle <- xgb.Booster.handle(modelfile = object$raw)
```
hcho3, line 132: I'm confused as to how this line is not currently working as desired. The `object$handle` attribute is assigned the return value of `xgb.Booster.handle`, right? Why do we need to also pass `object$handle` as an argument to `xgb.Booster.handle` also?

It seems similar to writing
```cpp
double foobar(double x, const double* y) {
  const double new_val = x + 1;
  *y = new_val;
  return new_val;
}
...
double v;
v = foobar(1.0, &v);
```

hcho3:
@david-cortes Aside: we've issued a guidance saying that users should avoid using `saveRDS()` / `readRDS()` for long-term archival of models. See https://www.rdocumentation.org/packages/xgboost/versions/1.4.1.1/topics/a-compatibility-note-for-saveRDS-save. This is because we don't provide backward compatibility with RDS-persisted models, i.e. you may not be able to read a model that you generated from a previous version of XGBoost. Are you aware of this guidance?

R-package/R/xgb.Booster.R
david-cortes: It's not that it isn't working, but that it won't update the R external pointer object in-place, creating a new one instead. If passed this way (as in the PR), it will just update an existing R object (which holds a null pointer).

E.g.
```r
new_obj <- readRDS(...)
predict(new_obj, new_data) # <- got de-serialized
predict(new_obj, more_data) # <- again de-serialized
...
```
vs.
```r
new_obj <- readRDS(...)
predict(new_obj, new_data) # <- got de-serialized
predict(new_obj, more_data) # <- already de-serialized
```

Currently, when a C++ object gets de-serialized after passing an object with a null pointer handle, the R pointer object to it will be short-lived as there will not be any R object tied to it (it will not be reachable after the function call ends, but it is not immediately garbage-collected).

Since the R pointer object (which is now no longer assigned to anything) is itself small, it's a less likely candidate to get garbage-collected by R, thus postponing C++ object destruction and increasing memory usage when one inadvertenly uses a model with a null handle.

david-cortes:
I'm aware that saveRDS is not recommended, but that doesn't mean it's not useful. For example, when one restarts an R session keeping the variables/objects, these will be auto-saved through `save`/`load`. Or perhaps an xgboost model might be stored inside another object (e.g. from some ML wrapper framework), for which only `save` and `saveRDS` are usable.

hcho3:
@david-cortes Got it. Just keep in mind that you will need to migrate existing RDS files using `xgb.save` / `xgb.load` when upgrading XGBoost to a later version.

R-package/R/xgb.Booster.R
hcho3: I'm still confused as to why calling `predict()` twice will trigger deserialization twice. If `object$handle` is assigned a reference to the new handle, the `object` should still have the reference to the new handle when `predict()` is called the second time?

david-cortes: `object$handle` is an element of an R list, which are copy-on-write, not pass-by-reference. Thus, it will not get updated outside of the function call.

hcho3: > R list, which are copy-on-write, not pass-by-reference

Wow I had no idea. I was thinking R lists are like a Python object with attributes, but I was mistaken.

hcho3 (approved):
Thank you! I got to learn about how objects work in R.

PR #6907, opened 2021-04-27
[jvm-packages] Make it easier to release GPU/CPU code artifacts to Maven Central

hcho3:
When I was running the following commands, 
```
mvn package -Dspark.version=3.0.0 -DskipTests -Duse.cuda=ON -Pgpu-with-gpu-tests
mvn deploy -Prelease -DskipTests
```
I expected to have two JAR files, `xgboost4j_2.12-1.4.1.jar` and `xgboost4j-gpu_2.12-1.4.1.jar`, the second of which contains CUDA code and the first does not. As a result, the second JAR is expected to be bigger than the first (160 MB vs 4.4 MB). But when I ran the commands, I often ended up with two JAR files with equal size (160 MB), indicating a mix-up.

This PR ensures that `xgboost4j-gpu_2.12-1.4.1.jar` gets the bigger `libxgboost4j.so` with CUDA code and that `xgboost4j_2.12-1.4.1.jar` gets the smaller `libxgboost4j.so` without CUDA code.

### DIFF
--- a/jvm-packages/xgboost4j-gpu/src
+++ b/jvm-packages/xgboost4j-gpu/src
@@ -1,1 +1,0 @@
-../xgboost4j/src/
--- a/jvm-packages/create_jni.py
+++ b/jvm-packages/create_jni.py
@@ -84,8 +84,9 @@ if __name__ == "__main__":
 
     print("building Java wrapper")
     with cd(".."):
-        maybe_makedirs("build")
-        with cd("build"):
+        build_dir = 'build-gpu' if cli_args.use_cuda == 'ON' else 'build'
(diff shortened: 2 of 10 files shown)


jvm-packages/xgboost4j-gpu/src
```diff
@@ -1 +0,0 @@
-../xgboost4j/src/
```
hcho3, line 1: I remove the `src/` symlink, so that the `xgboost4j-gpu` package can have an indepedent `resources/` directory that's distinct from the `resources/` directory of the `xgboost4j` package.

PR #6911, opened 2021-04-28
Disable pylint error.

trivialfis:


trivialfis:
Merging to remove 1 CI error.  Will continue to look into others:

- shotgun is still flaky.
- hist dask is flaky.
- Somehow a new error on osx with dask shows up with index error.  Maybe an upgraded numpy or dask has different behaviours on different platforms.
- github action osx failure.  Seems to be internal error in new libomp.

PR #6922, opened 2021-04-30
Reduce 'InitSampling' complexity and set gradients to zero

ShvetsKS:
Complexity of discard operation of MT generator is linear so current InitSampling implementation is not efficient due to unbalanced local thread's discards. In this PR similar [approach](https://github.com/dmlc/xgboost/blob/master/src/tree/gpu_hist/gradient_based_sampler.cu#L26) to GPU sampler was implemented:

- generate new seed for each iteration by global mt19937 rng.
- apply uniform distribution with default_random_engine (linear_congruential_engine) and generated seed.
- set gradients to zero for i-th row with condition rnd(i) > p

results for binary classification for synthetic data [61 x 10723651] with 'sample': 0.8:
|  | Master | This PR |
| ------------- | ------------- | ------------- |
| InitSampling |  27.06s | 0.78s |
| ApplySplit|  10.25s | 11.83s |
| BuildLocalHistograms|  25.9s | 32.76s |
| UpdatePredictionCache|  14.26s | 2.84s |
| Full training  | 87.8s | 58.9s |

Number of threads doesn't affect generated sequence.

### DIFF
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -48,9 +48,11 @@ void QuantileHistMaker::Configure(const Args& args) {
 }
 
 template<typename GradientSumT>
-void QuantileHistMaker::SetBuilder(std::unique_ptr<Builder<GradientSumT>>* builder,
+void QuantileHistMaker::SetBuilder(const size_t n_trees,
+                                   std::unique_ptr<Builder<GradientSumT>>* builder,
                                    DMatrix *dmat) {
   builder->reset(new Builder<GradientSumT>(
+                n_trees,
                 param_,
                 std::move(pruner_),
                 int_constraint_, dmat));
@@ -92,14 +94,15 @@ void QuantileHistMaker::Update(HostDeviceVector<GradientPair> *gpair,
   param_.learning_rate = lr / trees.size();
   int_constraint_.Configure(param_, dmat->Info().num_col_);
   // build tree
+  const size_t n_trees = trees.size();
   if (hist_maker_param_.single_precision_histogram) {
     if (!float_builder_) {
-      SetBuilder(&float_builder_, dmat);
+      SetBuilder(n_trees, &float_builder_, dmat);
     }
     CallBuilderUpdate(float_builder_, gpair, dmat, trees);
   } else {
     if (!double_builder_) {
-      SetBuilder(&double_builder_, dmat);
+      SetBuilder(n_trees, &double_builder_, dmat);
     }
     CallBuilderUpdate(double_builder_, gpair, dmat, trees);
   }
@@ -545,7 +548,6 @@ void QuantileHistMaker::Builder<GradientSumT>::ExpandWithLossGuide(
                          e.best.DefaultLeft(), e.weight, left_leaf_weight,
                          right_leaf_weight, e.best.loss_chg, e.stats.GetHess(),
                          e.best.left_sum.GetHess(), e.best.right_sum.GetHess());
-
       this->ApplySplit({candidate}, gmat, column_matrix, hist_, p_tree);
 
       const int cleft = (*p_tree)[nid].LeftChild();
@@ -589,18 +591,23 @@ void QuantileHistMaker::Builder<GradientSumT>::Update(
     DMatrix *p_fmat, RegTree *p_tree) {
   builder_monitor_.Start("Update");
 
-  const std::vector<GradientPair>& gpair_h = gpair->ConstHostVector();
-
+  std::vector<GradientPair>* gpair_ptr = &(gpair->HostVector());
+  // in case 'num_parallel_trees != 1' no posibility to change initial gpair
+  if (GetNumberOfTrees() != 1) {
+    gpair_local_.resize(gpair_ptr->size());
+    gpair_local_ = *gpair_ptr;
+    gpair_ptr = &gpair_local_;
+  }
   tree_evaluator_ =
       TreeEvaluator(param_, p_fmat->Info().num_col_, GenericParameter::kCpuId);
   interaction_constraints_.Reset();
   p_last_fmat_mutable_ = p_fmat;
 
-  this->InitData(gmat, gpair_h, *p_fmat, *p_tree);
+  this->InitData(gmat, *p_fmat, *p_tree, gpair_ptr);
   if (param_.grow_policy == TrainParam::kLossGuide) {
-    ExpandWithLossGuide(gmat, gmatb, column_matrix, p_fmat, p_tree, gpair_h);
+    ExpandWithLossGuide(gmat, gmatb, column_matrix, p_fmat, p_tree, *gpair_ptr);
   } else {
-    ExpandWithDepthWise(gmat, gmatb, column_matrix, p_fmat, p_tree, gpair_h);
+    ExpandWithDepthWise(gmat, gmatb, column_matrix, p_fmat, p_tree, *gpair_ptr);
   }
 
   for (int nid = 0; nid < p_tree->param.num_nodes; ++nid) {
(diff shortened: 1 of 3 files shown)


src/tree/updater_quantile_hist.cc
```diff
@@ -740,33 +761,19 @@ void QuantileHistMaker::Builder<GradientSumT>::InitSampling(const std::vector<Gr
...
+      uint64_t initial_seed_th = simple_skip(discard_size * tid, initial_seed, base);
+      std::linear_congruential_engine<std::uint_fast64_t, base, 0, (1ULL << 63ULL)> eng(initial_seed_th);
 
-      rnds[tid].discard(discard_size * tid);
       for (size_t i = ibegin; i < iend; ++i) {
-        if (gpair[i].GetHess() >= 0.0f && rnds[tid]() < coin_flip_border) {
-          p_row_indices_used[ibegin + row_offsets_used[tid]++] = i;
-        } else {
+        if (!(gpair[i].GetHess() >= 0.0f && coin_flip(eng))) {
           p_row_indices_unused[ibegin + row_offsets_unused[tid]++] = i;
```
trivialfis, line 770: Since you have 0 gradient for ignored samples, is it still necessary for the partitioner to be aware of sampling?

ShvetsKS: @trivialfis sorry for long response (was trying to fix cpp/python tests).
Practitioner still has to be aware about 'unused' rows to do `UpdatePredictionCache` correctly.

ShvetsKS:
@trivialfis could you take a look at this PR please?

src/common/row_set.h
```diff
@@ -103,9 +103,9 @@ class RowSetCollection {
     size_t* all_begin = dmlc::BeginPtr(row_indices_);
     size_t* begin = all_begin + (e.begin - all_begin);
 
-    CHECK_EQ(n_left + n_right, e.Size());
+    CHECK_LE(n_left + n_right, e.Size());
```
trivialfis, line 106: When is it less?

src/tree/updater_quantile_hist.cc
```diff
@@ -697,7 +701,26 @@ bool QuantileHistMaker::Builder<GradientSumT>::UpdatePredictionCache(
   return true;
 }
 
+/*
+  Right-to-left binary method: https://en.wikipedia.org/wiki/Modular_exponentiation
+*/
+uint64_t SimpleSkip(uint64_t exponent, uint64_t initial_seed, uint64_t base) {
```
trivialfis, line 707: Interesting algorithm. ;-)

src/tree/updater_quantile_hist.cc
```diff
@@ -706,6 +729,8 @@ void QuantileHistMaker::Builder<GradientSumT>::InitSampling(const std::vector<Gr
   unused_rows_.resize(info.num_row_);
   size_t* p_row_indices_used = row_indices->data();
   size_t* p_row_indices_unused = unused_rows_.data();
+  std::vector<GradientPair>& gpair_ref = const_cast<std::vector<GradientPair>&>(gpair);
```
trivialfis, line 732: Can you just make a copy of gradient as a class member and use it through out the current iteration?

src/tree/updater_quantile_hist.cc
```diff
@@ -740,19 +761,34 @@ void QuantileHistMaker::Builder<GradientSumT>::InitSampling(const std::vector<Gr
       const size_t ibegin = tid * discard_size;
       const size_t iend = (tid == (nthread - 1)) ?
                           info.num_row_ : ibegin + discard_size;
-
-      rnds[tid].discard(discard_size * tid);
+      constexpr uint64_t kBase = 16807;
```
trivialfis, line 764: Is this arbitrary?

src/tree/updater_quantile_hist.cc
```diff
@@ -740,19 +761,34 @@ void QuantileHistMaker::Builder<GradientSumT>::InitSampling(const std::vector<Gr
...
+      size_t local_unused_offset = 0;
+      size_t local_used_offset = 0;
       for (size_t i = ibegin; i < iend; ++i) {
-        if (gpair[i].GetHess() >= 0.0f && rnds[tid]() < coin_flip_border) {
-          p_row_indices_used[ibegin + row_offsets_used[tid]++] = i;
-        } else {
-          p_row_indices_unused[ibegin + row_offsets_unused[tid]++] = i;
+        if (!(gpair[i].GetHess() >= 0.0f && coin_flip(eng)) || gpair[i].GetGrad() == 0.0f) {
+          p_row_indices_unused[ibegin + local_unused_offset++] = i;
+          if (is_single_tree) {
```
trivialfis, line 775: Why is single tree different?

src/tree/updater_quantile_hist.cc
```diff
@@ -1072,28 +1115,62 @@ inline std::pair<size_t, size_t> PartitionDenseKernel(const common::DenseColumn<
         }
       }
     }
+    }
   } else {
     for (auto rid : rid_span)  {
+    if (pgh[rid*2] != 0) {
```
trivialfis, line 1121: ?

src/tree/updater_quantile_hist.cc
```diff
@@ -1072,28 +1115,62 @@ inline std::pair<size_t, size_t> PartitionDenseKernel(const common::DenseColumn<
...
+  }
+}
   return {nleft_elems, nright_elems};
 }
 
 // Split row indexes (rid_span) to 2 parts (left_part, right_part) depending
 // on comparison of indexes values (idx_span) and split point (split_cond).
 // Handle sparse columns
-template<bool default_left, typename BinIdxType>
+template<bool default_left, typename BinIdxType, bool check_gradient = false>
```
trivialfis, line 1163: I don't see any check?

src/tree/updater_quantile_hist.cc
trivialfis, line 732: Also, why do you need const cast?

trivialfis (changes requested):
Sorry for the long delay.  I think I don't quite understand the code in CPU hist now.  Could you please simplify it down or do some refactoring first before piling up more code?

- Why does the prediction cache update need to care about used rows?  Since you have looked into the GPU impl, I think the approach there is simpler.  WDYT?
- Please try to avoid making special cases.  Why does the number of trees have anything to do with sampling?
- Why do you need to assign a special task on tid 1?
....

I think most of those are not needed.  Thanks for optimizing it.  But please consider making some cleanups before.

src/common/row_set.h
ShvetsKS: reverted

src/tree/updater_quantile_hist.cc
ShvetsKS: copy is expensive operation for each iteration, and hopefully there is no  need to do a copy for single tree case (when `num_parallel_tree == 1`).
So currently I simplified the `InitSampling` (no dependency on number of trees) and copy is done only for not single trees (`num_parallel_tree != 1`).

src/tree/updater_quantile_hist.cc
ShvetsKS: No. it's similar value as for minstd_rand: https://en.cppreference.com/w/cpp/numeric/random/linear_congruential_engine
By requirements for lcg we have to use primitive root modulo n.

src/tree/updater_quantile_hist.cc
ShvetsKS: for single tree we can change (set to zeros for unused rows) initial gradient vector while for not single tree case we have to reset initial values for each tree assigned for iteration (special case `num_parallel_tree != 1`). So copy is done only for not single tree case and removed this condition from `InitSampling`.

src/tree/updater_quantile_hist.cc
ShvetsKS: It gives possibility to reduce work for partition and building histogram kernel as unused rows were marked by zero gradients. We can exclude these specific rows from build histogram process.

But currently I reverted it as the performance benefits from refactored InitSampling, UpdatePredictionCache (no exact prediction for unused rows) are much bigger than overheads for Partition and BuildLocalHistograms kernels.

src/tree/updater_quantile_hist.cc
ShvetsKS: deleted

src/tree/updater_quantile_hist.cc
ShvetsKS: deleted

ShvetsKS:
> Sorry for the long delay. I think I don't quite understand the code in CPU hist now. Could you please simplify it down or do some refactoring first before piling up more code?
> 
> * Why does the prediction cache update need to care about used rows?  Since you have looked into the GPU impl, I think the approach there is simpler.  WDYT?
> * Please try to avoid making special cases.  Why does the number of trees have anything to do with sampling?
> * Why do you need to assign a special task on tid 1?
>   ....
> 
> I think most of those are not needed. Thanks for optimizing it. But please consider making some cleanups before.

Thanks a lot for review!

Seems after applying your comments code become more obvious and clear.

Regarding prediction cache, there is no any specific handling for 'used' rows, except taking a leaf values from the latest partition.
For more effective partition and build histogram kernels it's better to exclude 'unused' rows indices from  `partition_builder_` (it was done by checking `(pgh[rid*2] != 0)` on root node ). And due to excluding 'unused' rows from partition we had to handle 'unused' rows in `UpdatePredictionCache` kernel exactly.
In refactored implementation I simplified `UpdatePredictionCache` (deleted expensive handling of 'unused' rows) and in addition to simplicity it brings more performance benefits than excluding of  'unused' rows from portion and build hist kernels.

ShvetsKS:
seems failed tests are not related to current changes:
windows build, test: problem with env "AssertionError: wheel_path = xgboost-1.5.0_SNAPSHOT-cp38-cp38-win_amd64.whl"
linux test: failed test in /workspace/tests/python/test_linear.py, but on master branch it's also failed (checked locally, and similar problem here: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/master/858/pipeline)

trivialfis:
I'm running into it too.   @hcho3 might have some insight.

ShvetsKS:
@trivialfis Thanks a lot for update.
I'm also in a process of preparing refactoring of cpu hist: first step is unifying the depthwise and lossguid strategy. But I would prefer to separate this work if possible as current PR doesn't affect code simplicity and clearness.

trivialfis:
Sure.  I would love to see any code simpliciation.  The part of this PR that confuses me is why row partitioner has to understand any unused rows.  From my understanding, as long as the gradient is set to 0, the rest should play out without any specialized handling.  Granted if we try to keep track of the unused rows the performance might be better.  But so far looking at the code base, this optimization contributes a large portion of complexity and I'm not entirely sure whether it's worthy.

ShvetsKS:
as you can see: https://github.com/dmlc/xgboost/pull/6922/files#diff-010c75219801f6b68880c42b0138f3b28517a0addb3055d98539a430ef3f3222L657
the track of unused rows was deleted :)
And currently we keep in mind that zero gradients doesn't impact histogram calculation and meanwhile we still apply partition for 'unused' rows without specialization (no big impact on performance).

ShvetsKS:
maybe this PR should be divided into two parts?
1) changing the type of random number generator to reduce complexity of discard operation
2) setting to zero gradients for 'unused' rows
as most complexity of current changes is due to setting gradients to zero.

trivialfis (commented):
> as most complexity of current changes is due to setting gradients to zero.

Do you have a branch of your refactoring code that you can share?  I'm quite looking forward to it and see how we can improve the current structure.  We still have some time before the next release so I think we don't have to rush into optimization PR. ;-)

src/tree/updater_quantile_hist.cc
ShvetsKS: yes. I applied refactoring for subsampling, thank for proposal!

ShvetsKS:
@trivialfis 

> > as most complexity of current changes is due to setting gradients to zero.
> 
> Do you have a branch of your refactoring code that you can share? I'm quite looking forward to it and see how we can improve the current structure. We still have some time before the next release so I think we don't have to rush into optimization PR. ;-)

I'll cleanup it a little and push draft pr with merging lossguide/depthwise strategies tomorrow.

But I think this PR become simple enough to be merged before refactoring. And it would be simpler to have this pr finished as base for refactoring process (new sampling, setting to zero gradients - it should be taken into account during refactoring).

trivialfis:
Thanks, will revisit today. ;-)

ShvetsKS:
small refactoring: https://github.com/dmlc/xgboost/pull/7007

trivialfis (approved):
Thanks for working on the optimization!

trivialfis:
Hi, out of curiosity, what tools do you use for profiling?  I have been using linux perf, valgrind and the monitor in XGBoost, they seem to contradict each other...

ShvetsKS:
> Hi, out of curiosity, what tools do you use for profiling? I have been using linux perf, valgrind and the monitor in XGBoost, they seem to contradict each other...

Basically I used [VTune Performance Analyzer](https://software.intel.com/content/www/us/en/develop/documentation/vtune-help/top.html) to have a deeper look at observed hotspots.

trivialfis:
Thanks for sharing!

PR #6928, opened 2021-05-01
[R-package] remove dependency on {magrittr} (fixes #6904)

### LINKED ISSUE #6904: [R-package] remove dependency on {magrittr}
I've noticed that the R package has an `Imports`-level dependency on `{magrittr}`: https://github.com/dmlc/xgboost/blob/8760ec48277b345aaaa895b82570c25566fc0503/R-package/DESCRIPTION#L65

It seems that that library is only being used for the pipe operator (`%>%`), and all the uses of that operator are internal to the R package.

Would you be open to a pull request that factors out this dependency? I don't believe that the `%>%` is providing any functional benefit, and removing `{magrittr}` might have the following benefits:

* faster continuous integration runs
* faster installs for users
* lower risk of installation issues from source / CRAN (though to be fair `{magrittr}` is very lightweight)
* lower risk of installation issues for the conda package (https://github.com/conda-forge/xgboost-feedstock/blob/91db3d16c18dcd78baee02c58dc3e5273ca53770/recipe/meta.yaml#L140)
* reduce the cost of `{magrittr}`'s reverse dependency checks

jameslamb:
This PR fixes #6904. As discussed in that issue, it proposes factoring out the R package's `Imports`-level dependency on `{magrittr}`.


ReeceGoding:
Can't the `FUN = function(x) list2mat(x)` in line `callbacks.R` be entirely replaced with just `list2mat`?

jameslamb:
> If I'm not mistaken, can't the `FUN = function(x) list2mat(x)` line in `callbacks.R` be entirely replaced with just `list2mat`?

maybe, but either way I didn't want to touch that as part of this PR. I wanted to make the `%>%`-less code preserve the logic of the existing code with `%>%` and avoid any other refactoring not related to that goal, to make the review as easy as possible.

hcho3:
Will review tomorrow

trivialfis (approved):
Thanks for the PR!  Do we need to change the cmake file too?

jameslamb:
> Thanks for the PR! Do we need to change the cmake file too?

Ah! I see a few missed places running `git grep magrittr`, actually. Sorry for the oversight. I'll remove them right now.

jameslamb:
Ok yes, sorry @trivialfis . I found a few more places that should have it removed: https://github.com/dmlc/xgboost/pull/6928/commits/119607b4298af299f3a9ee4868deafb399f2f027

hcho3 (approved):

hcho3:
Thank you!

jameslamb:
I just merged in latest `master` to get the CI fixes from #6951 and #6952.

trivialfis:
Thanks for the PR!

PR #6930, opened 2021-05-02
[R] don't remove CMakeLists in cleanup

ali5h:
currently installing the R-pacakge will leave the repo in dirty state, since
`CmakeLists.txt` is already checked in. This fixes the `cleanup`
script to not delete this file.

### DIFF
--- a/R-package/cleanup
+++ b/R-package/cleanup
@@ -1,4 +1,3 @@
 #!/bin/sh
 
 rm -f src/Makevars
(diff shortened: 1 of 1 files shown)


trivialfis (approved):

trivialfis:
Merging.  @hcho3 @RAMitchell but we need to figure out what to do with shotgun test.  Maybe we should just limit it to single thread during testing?

Issue #6908, opened 2021-04-27
Use array interface for numpy array inputs


sheecegardezi:
Code to reproduce:

using following  version
xgb.__version__= 1.4.0
System: Ubuntu 20.04 LTS

```
import xgboost as xgb
from sklearn.metrics import mean_squared_error
import pandas as pd
import numpy as np
from sklearn.datasets import load_boston
boston = load_boston()
data = pd.DataFrame(boston.data)
data.columns = boston.feature_names
X, y = data.iloc[:,:-1],data.iloc[:,-1]
xg_reg = xgb.XGBRegressor(objective ='reg:squarederror', colsample_bytree = 0.3, learning_rate = 0.1, max_depth = 5, alpha = 10, n_estimators = 10)
xg_reg.fit(X,y)
preds = xg_reg.predict(X)
rmse = np.sqrt(mean_squared_error(y, preds))
print("RMSE: %f" % (rmse))
```
Following warning is received: 
```
~venv/lib64/python3.6/site-packages/xgboost/data.py:115: UserWarning: Use subset (sliced data) of np.ndarray is not recommended because it will generate extra copies and increase memory consumption
  "memory consumption")
```

hcho3:
This is expected behavior, since you are using a sliced array (`data.iloc[:,:-1]`) as an input to `DMatrix`. The program will work; just keep in mind that the array will be copied and load to increased memory usage.

JoseLlorensRipolles:
Hey @sheecegardezi , 

Althought this issue is closed, I thought I would add that you can make the array contiguous with 
`X = np.ascontiguousarray(X)`
Which will improve memory usage and avoid the warning. 

Would you consider adding this fix for uncontiguous arrays to XGboost @hcho3?  I would like to do it if so. Thank you very much in advance.

GiuseppeLaurenza:
> Hey @sheecegardezi ,
> 
> Althought this issue is closed, I thought I would add that you can make the array contiguous with
> `X = np.ascontiguousarray(X)`
> Which will improve memory usage and avoid the warning.

Hi, it something to do with our data or a modification that XGBoost should do?

rteuwens:
[...]

PR #6934, opened 2021-05-03
Improve warning when using np.ndarray subsets

JoseLlorensRipolles:
The UserWarning prompted when using numpy subsets was a bit difficult to understand. I also added a suggestion on how to fix the warning.

### DIFF
--- a/python-package/xgboost/data.py
+++ b/python-package/xgboost/data.py
@@ -117,9 +117,10 @@ def _maybe_np_slice(data, dtype):
     try:
         if not data.flags.c_contiguous:
             warnings.warn(
-                "Use subset (sliced data) of np.ndarray is not recommended " +
+                "Use of np.ndarray subsets (sliced data) is not recommended " +
                 "because it will generate extra copies and increase " +
-                "memory consumption")
+                "memory consumption. Consider using np.ascontiguousarray to " +
+                "make the array contiguous.")
             data = np.array(data, copy=True, dtype=dtype)
         else:
             data = np.array(data, copy=False, dtype=dtype)


trivialfis (approved):
Thanks for the PR!

JoseLlorensRipolles:
Thank **you** for your awesome work at XGBoost 😄!

rteuwens:
Context: https://github.com/dmlc/xgboost/issues/6908#issuecomment-834658652

Given your suggestion, would this break xgboost?

```py
        if not data.flags.c_contiguous:
            data = np.ascontiguousarray(data)
        data = np.array(data, copy=False, dtype=dtype)
```

Currently, any interaction with sklearn CV split objects results in terminal UserWarning spam.

JoseLlorensRipolles:
> Context: [#6908 (comment)](https://github.com/dmlc/xgboost/issues/6908#issuecomment-834658652)
> 
> Given your suggestion, would this break xgboost?
> 
> ```python
>         if not data.flags.c_contiguous:
>             data = np.ascontiguousarray(data)
>         data = np.array(data, copy=False, dtype=dtype)
> ```
> 
> Currently, any interaction with sklearn CV split objects results in terminal UserWarning spam.

No, it would not break XGBoost, but it would change the structure of the user array without their intent and without giving notice, which could be problematic. In my opinion the copy approach is safer. However, it is true that this warning raises quite a bit when using some sklean methods (Like a gridsearch), so maybe the warning could be dropped altogether.

rteuwens:
> No, it would not break XGBoost, but it would change the structure of the user array without their intent and without giving notice, which could be problematic. In my opinion the copy approach is safer. However, it is true that this warning raises quite a bit when using some sklean methods (Like a gridsearch), so maybe the warning could be dropped altogether.

The error is super annoying and hard to suppress. Limiting the notification to once or dropping it altogether sounds like a welcome change, tbh...

hcho3:
@rteuwens I filed #6997 to remove the warning.

rteuwens:
> @rteuwens I filed #6997 to remove the warning.

Much love! Thank you

PR #6935, opened 2021-05-03
Link to valid tree_method values in docs 

dsaxton:
Within the Python sklearn API reference it's mentioned that `tree_method` should come from the parameters document, but there doesn't appear to be an easy way to find that document. I'm assuming this is what's meant, in which case I think it would be helpful to provide a link in the docstring: https://xgboost.readthedocs.io/en/latest/python/python_api.html#module-xgboost.sklearn

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -106,7 +106,7 @@ __model_doc = f'''
         Specify which tree method to use.  Default to auto.  If this parameter
         is set to default, XGBoost will choose the most conservative option
         available.  It's recommended to study this option from parameters
-        document.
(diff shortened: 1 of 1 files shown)


trivialfis (approved):
Thanks for the PR!

trivialfis:
The tree method doc was added recently.  I believe there are lots of places where we can improve the documents.  Free feel to open issues/prs if you find something confusing or have suggestions.

dsaxton:
> The tree method doc was added recently. I believe there are lots of places where we can improve the documents. Free feel to open issues/prs if you find something confusing or have suggestions.

Sounds good, will submit more PRs if I see anything else :+1:

PR #6946, opened 2021-05-07
[WIP] Fix multiclass auc with empty dataset.

trivialfis:
Will rebase onto master after debugging.  For some reason the allreduce in nccl triggers a memory access error and I haven't figured out the root cause.  I'm curious in seeing what's going to happen on CI.

### DIFF
--- a/src/objective/multiclass_obj.cu
+++ b/src/objective/multiclass_obj.cu
@@ -117,7 +117,8 @@ class SoftmaxMultiClassObj : public ObjFunction {
     std::vector<int>& label_correct_h = label_correct_.HostVector();
     for (auto const flag : label_correct_h) {
       if (flag != 1) {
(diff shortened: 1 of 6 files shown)


src/objective/multiclass_obj.cu
```diff
@@ -117,7 +117,8 @@ class SoftmaxMultiClassObj : public ObjFunction {
     std::vector<int>& label_correct_h = label_correct_.HostVector();
     for (auto const flag : label_correct_h) {
       if (flag != 1) {
-        LOG(FATAL) << "SoftmaxMultiClassObj: label must be in [0, num_class).";
+        LOG(FATAL) << "SoftmaxMultiClassObj: label must be in [0, num_class)."
+                   << "Labels:" << info.labels_.Size();
```
trivialfis, line 121: Revert this

PR #6951, opened 2021-05-11
[CI] Fix CI/CD pipeline broken by latest auditwheel (4.0.0)

hcho3:
The latest auditwheel change the formatting of the name of the file it outputs. See https://github.com/pypa/auditwheel/pull/288. So introduce an extra file renaming operation so that we have a single unique Python wheel in the location `python-package/dist/`.


trivialfis (approved):
LGTM.

PR #6961, opened 2021-05-12
Fix compiler error in x86 building

Livius90:
Following error was fixed in x86 release build:
```
Error C2398: Element '4': conversion from 'xgboost::bst_ulong' to 'size_t' requires a narrowing conversion, xgboost\src\c_api\c_api.cc 738	
Error_C2397: conversion from 'uint64_t' to '_Ty' requires a narrowing conversion, xgboost\src\gbm\gbtree.cc 234
```


Livius90:
As i see `static_cast<size_t>(p_fmat->Info().num_row_)` need instead of `(size_t)(p_fmat->Info().num_row_)` for Jenkins then it would be fine.

Livius90:
I will fix static_cast<size_t> for Jenkins then i start a new pull request later.

trivialfis:
No problem.  Thanks for fixing it!

PR #6964, opened 2021-05-13
Fix compiler error in x86 building - 2nd request

Livius90:
Following `size_t` conversion errors were fixed in x86 release build:
```
Error C2398: Element '4': conversion from 'xgboost::bst_ulong' to 'size_t' requires a narrowing conversion, xgboost\src\c_api\c_api.cc 738	
Error C2397: conversion from 'uint64_t' to '_Ty' requires a narrowing conversion, xgboost\src\gbm\gbtree.cc 234
```


Livius90:
As i see in release-1.4.0 branch the first issue which is "Error C2398" is already fixed.
https://github.com/dmlc/xgboost/commit/a6d1fbf8d1d295a910a847d315eb11e4fa84e599

Need it to merge to master also and my second solution can fix the next issue in gbtree.cc.

trivialfis:
Merged.  Thanks for the fix!

PR #6966, opened 2021-05-13
Simplify list2mat call from lapply in callbacks.R

ReeceGoding:
`list2mat` is a function of one argument, so the call can be simplified.


trivialfis (approved):
Thanks for the PR!  That's a weird line of code. ;-)

PR #6967, opened 2021-05-13
Corrected lapply comment in callbacks.R

ReeceGoding:
The comment was made false by the removal of the pipes. What was originally the first `lapply` is now the second. I've done what I can on my own machine to confirm that the second call is indeed doing the job, although one could argue that it's actually the second and the third that are doing the job.

Given that we're using two `lapply`s for one job, I can't help but think that there is an `mapply` or `Map` way to do this better. However, I've tried my best and come up with nothing.


trivialfis (approved):
Thanks for the fix!

PR #6969, opened 2021-05-18
[dask] Fix prediction on df with latest dask.

trivialfis:
For some unknown reasons in dask 2021.05.0, list is turned into a future for `map_blocks`.  Use tuple instead.

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -1000,7 +1000,7 @@ async def _direct_predict_impl(  # pylint: disable=too-many-branches
     output_shape: Tuple[int, ...],
     meta: Dict[int, str],
 ) -> _DaskCollection:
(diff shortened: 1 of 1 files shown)


jakirkham:
Interesting. This seems like it could be a bug in Dask. Do you have an example? Maybe we can file an issue on [the Dask issue tracker]( https://github.com/dask/dask/issues )

RAMitchell (approved):

trivialfis:
@jakirkham I'm trying to create a reproducible example, this is blocking the CI so have to work around it first, sorry for the confusion.

jakirkham:
Oh no worries. Just trying to see how we can help :)

trivialfis:
@jakirkham I filed an issue on https://github.com/dask/dask/issues/7681 .

PR #6983, opened 2021-05-24
Grouping features for computing feature contributions (SHAP values)

oOTWK:
Categorial features are often preprocessed by one-hot encoding. PredictContribution() computes SHAP values for each feature. The issue is if there are one-hot encoded features, it computes SHAP values for each one-hot (yes or no) feature. But what we really want is the SHAP value of the original categorical feature (group of one-hot encoded features).

In this PR, PredictContribution() takes group_indices (array of int) which contains a group index given a feature index. Then, SHAP values are computed for each group of features. 

Moreover, this is useful when there are too many features and feature contributions are hard to interpret. Grouping features gives better interpretability and also saves computation time and memory of TreeSHAP.

I hope this will be implemented for the GPU version in future releases.

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -726,6 +726,8 @@ XGB_DLL int XGBoosterEvalOneIter(BoosterHandle handle,
  *    The second scenario applies when you are defining a custom objective function.
  * \param out_len used to store length of returning result
  * \param out_result used to set a pointer to array
+ * \param group_indices group index of features for group of feature contributions 
+ * \param num_feat_group the number of feature groups provided in group_indices
  * \return 0 when success, -1 when failure happens
  */
 XGB_DLL int XGBoosterPredict(BoosterHandle handle,
@@ -734,7 +736,9 @@ XGB_DLL int XGBoosterPredict(BoosterHandle handle,
                              unsigned ntree_limit,
                              int training,
                              bst_ulong *out_len,
-                             const float **out_result);
+                             const float **out_result,
+                             int *group_indices = nullptr,
+                             int num_feat_group = 0);
 /*!
  * \brief Make prediction from DMatrix, replacing `XGBoosterPredict`.
  *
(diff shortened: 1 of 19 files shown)


trivialfis:
We are working on the appveyor, please ignore it for now.

RAMitchell (commented):
Nice work, it's a fairly clean modification. Xgboost is working towards native categorical support, I wonder if it might be better to wait for this feature? You wouldn't have to have a new interface for grouping features, xgb would already handle that.

I don't like to keep adding parameters to predict interface, especially when they only apply sometimes.

Mathematically, is the sum of the shapley values of a one hot encoded feature equivalent to the shapley value of its group under this modification?

oOTWK:
Thank you for the review!

> Nice work, it's a fairly clean modification. Xgboost is working towards native categorical support, I wonder if it might be better to wait for this feature? You wouldn't have to have a new interface for grouping features, xgb would already handle that.

Not just for categorical features, but it would be also useful to see how different combinations of features contribute to prediction. Especially, when there are too many features, it is hard to interpret each individual feature.

> I don't like to keep adding parameters to predict interface, especially when they only apply sometimes.

I agree. Can we have a separate API call for calculating feature contributions?

> Mathematically, is the sum of the shapley values of a one hot encoded feature equivalent to the shapley value of its group under this modification?

No, mathematically it's different. That's why `split_index` is replaced by `group_index` in TreeSHAP(). This changes the list of `PathElement`

RAMitchell:
> I agree. Can we have a separate API call for calculating feature contributions?

It does make sense to have a separate call, it's up to you if you want to implement this.

I think you can also achieve this modification without much trouble using GPUTreeShap. GPUTreeShap accepts a user-defined interface to the X matrix - you could group features via this interface, as well as defining a custom split type and it should work.

oOTWK:
Hello @RAMitchell 

I checked native categorical support and this will be certainly helpful for computing Shap values of categorical features. Still, we might want to compute mutual feature contributions of an arbitrary group of features. This will be useful when there are too many features and/or for correlated features.

As you suggested, I tried to achieve this using GPUTreeShap with a custom split type. To do so, all features in the same mutual group have to be merged into one path element because grouping features will reduce the permutations in Shap computation. As there could be possibly more than one feature involved in one path element, EvaluateSplit (to compute one_fraction) needs upper-bounds and lower-bounds of all features involved in the path element. This augmentation makes PathElement variable size and I don't think it is possible to (efficiently) allocate the array of PathElement in __shared__ memory. I'm new to CUDA language. Could you provide some suggestions?

On the other hand, grouping features in cpu TreeShap is a simple modification and it's just a matter of API style. Instead of adding more parameters to Predict interface, how about putting mutual_feature_index to DMatrix as its metadata? So this can be set through DMatrix before calling Predict(dmat, contrib=True).

Thank you.

PR #6986, opened 2021-05-24
[CI] Move appveyor tests to action

trivialfis:
* Drop support for VS14, use VS15 instead.
* Drop support for mingw.
* Remove debug build.
* Split up jvm tests.
* Split up Python tests.


RAMitchell (commented):
So this retains mingw support?

RAMitchell (approved):

hcho3:
Just removed the webhook for AppVeyor. Can you try merging in the `master` branch and try again?

.github/workflows/main.yml
```diff
@@ -101,6 +101,66 @@ jobs:
...
 
+  python-tests:
+    name: Test XGBoost Python package
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        config:
+          - {os: windows-2016, compiler: 'mingw', python-version: "3.8"}
+          - {os: windows-2016, compiler: 'msvc', python-version: "3.8"}
+        python-version: ["3.8"]
```
hcho3, line 112: I don't see this test running in the status page?

hcho3:
Can we add back MinGW? Also, we may want to install MinGW first. See https://github.com/marketplace/actions/install-mingw

trivialfis:
Looked into the action you recommended, trying it now.

trivialfis:
Nope, doesn't work.

hcho3:
Not sure how well we can support MinGW. We should maybe drop it?

trivialfis:
> Not sure how well we can support MinGW. We should maybe drop it?

So far I think the biggest consumer for mingw is R.  Personally I'm happy to drop support for mingw in non-R environment.  Nowadays VC is pretty good, also clang is easy to use on Windows ..

trivialfis:
But since I'm not a Windows user so I can't speak for the others.

hcho3:
@trivialfis Have you tried installing R in GitHub Actions? That should give us access to working MinGW.

.github/workflows/python_tests.yml
```diff
@@ -0,0 +1,101 @@
...
+
+    - name: Build XGBoost with mingw
+      shell: bash -l {0}
+      if: matrix.config.compiler == 'mingw'
+      run: |
+        mkdir build_mingw
+        cd build_mingw
+        cmake ../ -G"Unix Makefiles" -DGOOGLE_TEST=ON  -DUSE_DMLC_GTEST=ON
+        make -j$(nproc)
+        ctest --extra-verbose
```
hcho3, line 79: Please remove this section, since we won't test MinGW.

hcho3 (changes requested):

.github/workflows/python_tests.yml
trivialfis: Removed.

hcho3 (approved):

PR #6987, opened 2021-05-24
Revert OMP guard.

trivialfis:
The guard protects the global variable from being changed by XGBoost.  But this leads to a
bug that the `n_threads` parameter is no longer used after the first iteration.  This is
due to the fact that `omp_set_num_threads` is only called once in `Learner::Configure` at
the beginning of the training process.

The guard is still useful for `gpu_id` since `cudaSetDevice` is called all the time in our code base.


RAMitchell (approved):
Global state is fun.

PR #6996, opened 2021-05-25
[dask] Avoid constructing DaskDMatrix in Predict.

trivialfis:


RAMitchell (approved):

python-package/xgboost/dask.py
```diff
@@ -1486,20 +1486,17 @@ async def _predict_async(
                 base_margin=base_margin,
                 validate_features=validate_features,
             )
-            if isinstance(predts, dd.DataFrame):
-                predts = predts.to_dask_array()
         else:
-            test_dmatrix = await DaskDMatrix(
-                self.client, data=data, base_margin=base_margin, missing=self.missing
```
hcho3, line 1493: Will `base_margin` and `missing` be silently ignored?

trivialfis: How come that I keep forgetting this.  Need to add a test for it to remind myself.  Thanks for catching this!  Will close this PR for now.

PR #7001, opened 2021-05-26
Implement categorical prediction for CPU and GPU predict leaf.

trivialfis:
* Implement categorical prediction for CPU prediction.
* Implement categorical prediction for GPU predict leaf.
* Refactor the prediction functions to have unified get next node.

Related: https://github.com/dmlc/xgboost/issues/6503 .

### DIFF
--- a/include/xgboost/tree_model.h
+++ b/include/xgboost/tree_model.h
@@ -445,6 +445,10 @@ class RegTree : public Model {
                          bst_float right_leaf_weight, bst_float loss_change,
                          float sum_hess, float left_sum, float right_sum);
 
+  bool HasCategoricalSplit() const {
+    return !split_categories_.empty();
+  }
+
   /*!
    * \brief get current depth
    * \param nid node id
@@ -537,13 +541,6 @@ class RegTree : public Model {
     std::vector<Entry> data_;
     bool has_missing_;
   };
-  /*!
-   * \brief get the leaf index
-   * \param feat dense feature vector, if the feature is missing the field is set to NaN
-   * \return the leaf index of the given feature
-   */
-  template <bool has_missing = true>
-  int GetLeafIndex(const FVec& feat) const;
 
   /*!
    * \brief calculate the feature contributions (https://arxiv.org/abs/1706.06060) for the tree
@@ -583,14 +580,6 @@ class RegTree : public Model {
   void CalculateContributionsApprox(const RegTree::FVec& feat,
                                     bst_float* out_contribs) const;
   /*!
-   * \brief get next position of the tree given current pid
-   * \param pid Current node id.
-   * \param fvalue feature value if not missing.
-   * \param is_unknown Whether current required feature is missing.
-   */
-  template <bool has_missing = true>
-  inline int GetNext(int pid, bst_float fvalue, bool is_unknown) const;
-  /*!
    * \brief dump the model in the requested format as a text string
    * \param fmap feature map that may help give interpretations of feature
    * \param with_stats whether dump out statistics as well
@@ -626,6 +615,20 @@ class RegTree : public Model {
     size_t beg {0};
     size_t size {0};
   };
+
+  struct CategoricalSplitMatrix {
+    common::Span<FeatureType const> split_type;
+    common::Span<uint32_t const> categories;
+    common::Span<Segment const> node_ptr;
+  };
+
+  CategoricalSplitMatrix GetCategoriesMatrix() const {
+    CategoricalSplitMatrix view;
+    view.split_type = common::Span<FeatureType const>(this->GetSplitTypes());
+    view.categories = this->GetSplitCategories();
+    view.node_ptr = common::Span<Segment const>(split_categories_segments_);
+    return view;
+  }
 
  private:
   void LoadCategoricalSplit(Json const& in);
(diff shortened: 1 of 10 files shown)


include/xgboost/tree_model.h
```diff
@@ -724,38 +723,5 @@ inline bool RegTree::FVec::IsMissing(size_t i) const {
...
-      } else {
-        return (*this)[pid].RightChild();
-      }
-    }
-  } else {
-    // 35% speed up due to reduced miss branch predictions
-    // The following expression returns the left child if (fvalue < split_cond);
-    // the right child otherwise.
-    return (*this)[pid].LeftChild() + !(fvalue < (*this)[pid].SplitCond());
-  }
```
ShvetsKS, line 757: Why it's not preserved in new `GetNextNode` function?
At least 35% of performance regression will be introduced
Fast check on higgs1m data: **master 0.77s vs this pr: 1.24s** (caused by a lot of newly introduced conditions).

ShvetsKS: if it's possible I could help with reducing the perf degradations :)

trivialfis: Sure, feel free to push to my branch or made PRs. ;-)

I did run some quick benchmarks on this PR with datasets generated by `sklearn.datasets.make_regression` with 1e6 samples and 30 features.  The run time didn't change at all ...

trivialfis: @ShvetsKS Also please be mindful that we have a pruner, I'm not entirely sure whether right node index is exactly left node index + 1.

trivialfis: @ShvetsKS I came up with this branch, feel free to take a look https://github.com/trivialfis/xgboost/tree/cat-cpu-predictor-optimization .

I'm running it on `make_regression` with 1e7 samples and 30 features, the run time is about the same (slightly faster).  Not sure whether it works on you environment.

ShvetsKS: possibly you can't observe regression as dmatrix create takes most of the time for inference on such dimensions, for clear comparison it's better to use `inplace_predict`

ShvetsKS: > @ShvetsKS I came up with this branch, feel free to take a look https://github.com/trivialfis/xgboost/tree/cat-cpu-predictor-optimization .

I'll try to be helpful :)

trivialfis: Actually I'm using `inplace_predict`.  This is the function is used to run the simple bench (it's part of some other random scripts I haven't been able to cleanup yet ...)

``` python
def reg_predict(args):
    X, y = make_regression(args)
    parameters = configure(args, PARAMETERS.copy())
    booster = xgb.Booster(
        model_file=parameters.get('tree_method', 'updater') + "reg.json"
    )
    s = time()
    for i in range(10):
        booster.inplace_predict(X)
    print("inplace_predict", time() - s)
```

trivialfis: @ShvetsKS I have merged my small optimization into this branch.

ShvetsKS: @trivialfis  sorry for long response!
I fixed perf degradations with simple modifications: https://github.com/ShvetsKS/xgboost/commits/cat-cpu-predictor 907ce2e7d98cc1b1a59a41c54a28d90af17416d6

1) moved template parameter setting for `has_categorical` from the loop
2) usage of formula `left node index + 1` was reverted as we have the check for this (https://github.com/dmlc/xgboost/blob/master/src/tree/updater_quantile_hist.cc#L1067) and pruners do not change the node id in case of pruning. Seems there is no reason not to use this formula as in pruning only `SetLeaf` is called and it doesn't change the id on node.
3) small improvement of `GetNextNode` as there is no need to use all nodes as only one node reference is needed

results:
| Data        | master | this pr | fixed perf |
| ------------- |:-------------:| -----:| -----:|
| higgs     | 0.616 | 1.03 | 0.612 |
| airline-ohe      | 0.78 |  0.67 | 0.79 |
| mortgage |2.41|2.71|2.41|
| msrank |1.90|2.57|1.89|
| santander|0.272|0.41|0.23|

trivialfis: > usage of formula left node index + 1 was reverted

Users can run pruner themselves, it doesn't have to be part of hist.

trivialfis: The issue with pruner is, those deleted nodes can be reused in `AllocNode`.

trivialfis: Rest of your optimization looks good and feel free to push to my branch.  ;-)  Thanks for working on this

ShvetsKS: > > usage of formula left node index + 1 was reverted
> 
> Users can run pruner themselves, it doesn't have to be part of hist.

strange there is no any compliance since this approach was introduced long ago.
Could you provide more details or some example with such situations, and I think it's possible to add checks into `AllocNode` for example?

trivialfis: What's the overhead of calling right node method given a `Node` struct (instead of indexing into tree) first?  It should be fully inlined right?

ShvetsKS: sorry, on higgs with `inplace_predict` 1.5x degradation: use `RightChild()` with exact condition - 0.927s, `LeftChild() + 1`without exact condition - 0.615s.
seems the reason is branch prediction miss, not inlining.

trivialfis: Thanks for sharing the perf change.  So far it seems to be correct since after a tree is pruned, there's no way to continue growing it so the `Reuse` function is never used.  As far as I know, non of the existing updater can do it at the moment.

trivialfis: @ShvetsKS I have merged your code into my branch and removed the unused variable `go_left`.  Could you please take another look?

ShvetsKS: @trivialfis thanks a lot. Looks good for me!

trivialfis: Thanks for sharing the optimization!

src/predictor/cpu_predictor.cc
```diff
@@ -344,7 +391,9 @@ class CPUPredictor : public Predictor {
         }
         feats.Fill(page[i]);
         for (unsigned j = 0; j < ntree_limit; ++j) {
-          int tid = model.trees[j]->GetLeafIndex(feats);
+          auto const& tree = *model.trees[j];
+          auto const& cats = tree.GetCategoriesMatrix();
+          bst_node_t tid = GetLeafIndex<true, true>(tree, feats, cats);
```
hcho3, line 396: Should we also add a check for `has_categorical` here? Or are we purposefully removing it as `GetLeaf` is not performance critical?

ShvetsKS: also it's possible `feats` has no missing values

trivialfis: Yes, I'm assuming it's not critical.  Otherwise we will have a lot more specializations

trivialfis: This is the previous default so I'm not slowing it down here.  If optimization is needed (being the computation bottleneck of some algorithms/models), we can come back to it in a different PR that focuses on optimization.

PR #6825, opened 2021-04-05
[CI] Upgrade cuDF and RMM to 0.19 nightly

hcho3:
Might help fix https://github.com/rapidsai/cudf/issues/7722

PR #7012, opened 2021-06-01
[CI] Upgrade cuDF and RMM to 21.06 nightly

hcho3:
Replaces #6825

Ensure that XGBoost works with the upcoming 21.06 release of RAPIDS software.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -62,9 +62,7 @@ pipeline {
             // using CentOS 6 image
             'build-gpu-cuda10.0': { BuildCUDA(cuda_version: '10.0') },
             // The build-gpu-* builds below use Ubuntu image
-            'build-gpu-cuda10.1': { BuildCUDA(cuda_version: '10.1') },
-            'build-gpu-cuda10.2': { BuildCUDA(cuda_version: '10.2', build_rmm: true) },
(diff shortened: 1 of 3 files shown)


hcho3:
I've removed CUDA 10.1 and 10.2 from the test matrix.

trivialfis (approved):
Nice work on handling all these complexities!

hcho3:
@trivialfis Do you think it's a big jump if we require CUDA 11.0+ in the next release? (Meaning, users with CUDA 10.x will need to build XGBoost from the source, with no assurance that things will work due to lack of test coverage)

trivialfis:
I'm fine with that.  If needed we can put a warning in C++ code, probably through the use of  `cudaRuntimeGetVersion()`

hcho3:
Do you mean a deprecation warning?

trivialfis:
Yup, if that's needed.

hcho3:
Not sure if it's possible to load a shared lib if the shared lib was built with CUDA 11.0 and the target machine is using CUDA 10.x. It may result in failing to load `libxgboost.so` at all.

trivialfis:
I think it's possible to load. CUDA links a static library into `libxgboost.so`, which handles different versions.

Issue #6816, opened 2021-03-31
Speed up dask tests.


trivialfis:
https://github.com/microsoft/LightGBM/pull/4088#issuecomment-810711965

jmoralez:
Hi. I'd be happy to give this a go, just have a couple of questions on the scope.
1. Should this target only the tests that take in a `client` fixture or every test that builds a cluster?
2. The main component is creating a cluster fixture that lives throughout the tests. What should the client be?
    * A fixture that uses the cluster fixture.
    * Define `client = Client(cluster)` in every test.
    * Use the `with Client(cluster) as client` context manager in every test.
    * Something else

trivialfis:
Closing since the bottleneck is hypothesis tests.  Thanks for the PR @jmoralez

PR #7020, opened 2021-06-03
[dask] speed up tests

jmoralez:
This aims to reduce the runtime of the dask tests. Following https://github.com/dmlc/xgboost/issues/6816#issuecomment-852831437, the first step was to replace the `client` fixture with one that reuses the same cluster and just creates new clients in every test, and makes the least changes to the existing code. 

There are some other tests that are building clusters instead of using the `client` fixture that could be benefited by this, however adding this only reduced the runtime by about a minute, so I ran `pytest` with `--durations=0` and got this:
```
=============================================================== slowest durations ================================================================
155.91s call     python/test_with_dask.py::TestWithDask::test_approx
138.92s call     python/test_with_dask.py::TestWithDask::test_hist
22.91s call     python/test_with_dask.py::test_from_dask_array
9.20s call     python/test_with_dask.py::TestWithDask::test_feature_weights
7.02s call     python/test_with_dask.py::test_dask_ranking
6.61s call     python/test_with_dask.py::test_with_asyncio
6.44s call     python/test_with_dask.py::test_parallel_submit_multi_clients
```
Will investigate the ones that take the most time.

### DIFF
--- a/tests/python/test_with_dask.py
+++ b/tests/python/test_with_dask.py
@@ -28,7 +28,6 @@ if tm.no_dask()['condition']:
     pytest.skip(msg=tm.no_dask()['reason'], allow_module_level=True)
 
 from distributed import LocalCluster, Client
-from distributed.utils_test import client, loop, cluster_fixture
 import dask.dataframe as dd
 import dask.array as da
 from xgboost.dask import DaskDMatrix
@@ -38,6 +37,20 @@ if hasattr(HealthCheck, 'function_scoped_fixture'):
     suppress = [HealthCheck.function_scoped_fixture]
 else:
     suppress = hypothesis.utils.conventions.not_set  # type:ignore
+
+
+@pytest.fixture(scope='module')
+def cluster():
+    with LocalCluster(
+        n_workers=2, threads_per_worker=2, dashboard_address=None
+    ) as dask_cluster:
+        yield dask_cluster
+
+
+@pytest.fixture
+def client(cluster):
+    with Client(cluster) as dask_client:
+        yield dask_client
 
 
 kRows = 1000


trivialfis:
The slow test is probably not caused by dask but by hypothesis.

jmoralez:
I see. The total runtime for the tests in this PR was [14 minutes](https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-7020/1/pipeline/197#step-225-log-2154) and the current master is around [17 minutes](https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/master/878/pipeline/207#step-237-log-1542). Should I add the `client` fixture to tests that currently build a cluster internally like these?
https://github.com/dmlc/xgboost/blob/7beb2f7fae9589750c8d8dd01d039f90c71fc81f/tests/python/test_with_dask.py#L67
https://github.com/dmlc/xgboost/blob/7beb2f7fae9589750c8d8dd01d039f90c71fc81f/tests/python/test_with_dask.py#L110
That could probably reduce the runtime by a couple more minutes. Or if you have any other suggestions for maybe improving the hypothesis ones I could look into it.

trivialfis:
> hould I add the client fixture to tests that currently build a cluster internally like these?

Some tests use specific number of workers so that they have to define their own cluster.

> Or if you have any other suggestions for maybe improving the hypothesis ones I could look into it.

Sorry I don't have any suggestion, you know these better than me. ;-)

jmoralez:
I ran the tests that take the most time and pretty much all the time is spent training so I don't think there's something that could be improved there. There are a couple of clusters that get created with `kWorkers` and no `threads_per_worker`, I could create an additional fixture like `kCluster` and pass that to those tests so that they reuse that one.

This didn't improve as much as I hoped haha so please let me know if its useful at all.

trivialfis:
> I ran the tests that take the most time and pretty much all the time is spent training so I don't think there's something that could be improved there.

That's fine, the hypothesis tests take longer than we would like but highly effective at catching bugs.

> There are a couple of clusters that get created with kWorkers and no threads_per_worker

Thank you for looking into them!

>  could create an additional fixture like kCluster and pass that to those tests so that they reuse that one.

I will follow up on making those changes since I wrote most of  the tests, I should cleanup my own mess.

> This didn't improve as much as I hoped haha so please let me know if its useful at all.

Of course it's useful and thank you!  After merging the PR we know that we should focus on other places.

jmoralez:
@trivialfis I think this is ready for review, looking forward to your thoughts.

trivialfis (approved):

Issue #5951, opened 2020-07-28
Inplace Predict Java API (xgboost4j)


viswanathk:
Hello.

I see that in the changelog for 1.1.0, a new API was added for thread safe prediction (`inplace_predict`). The current `predict` method exposed through xgboost4j is not thread safe, and hence marked as `synchronized` in the method.

Given that the `inplace_predict` can now do threadsafe prediction, what are the plans of adding this to the Java API? We use this library in a multithreaded Java application, and this can help boost performance.

Thanks,
Viswanath.

trivialfis:
Would you like to open a PR for it?  Inplace prediction means XGBoost has to digest external data directly.  I'm not familiar with the ecosystem of Java so I don't know what data type to support.

trivialfis:
Also after https://github.com/dmlc/xgboost/pull/5853 normal predict function should also be thread safe in xgboost native code.

viswanathk:
Good to know about #5853, but looks like it might take a while before we see it in release, and inplace_predict already does the job. Do you think it will be faster to add the Java wrapper to the code, than #5853 waiting for to make it release?

> Would you like to open a PR for it? Inplace prediction means XGBoost has to digest external data directly. I'm not familiar with the ecosystem of Java so I don't know what data types to support.

I will go over the required changes - although I am not sure on where the gotchas can potentially be.

trivialfis:
It's a blocking PR for the next release and we are now squashing the next release, so I would say pretty soon.


> I will go over the required changes

I can help with issues in C++.  @CodingCat expressed interest in inplace prediction before.  So maybe @CodingCat can provide some suggestions on related topics.

dozaza:
Hi guys,

I'm working an online XGBoost prediction service, and want to use Java Multi thread for better performance.

As C++ code has resolved thread safe feature, simply removing "synchronize" key word in Java "predict" method, could enable multi-thread prediction in Java Environment?

trivialfis:
The C API is thread safe now.  You need to check whether JAVA layer has additional states.

viswanathk:
Java layer doesn't seem to maintain any state that get's updated on the `predict` flow. Raised the PR https://github.com/dmlc/xgboost/pull/6021

Who's the maintainer / reviewer for the Java package?

kururuken:
Any update on this issue?

sihanwang94:
Need this feature +1

PR #7027, opened 2021-06-07
[JVM-Packages] Remove synchronized in java predict function.

trivialfis:
We have made the predict function thread safe in 1.4.

Follow up on https://github.com/dmlc/xgboost/pull/6021 .  Related https://github.com/dmlc/xgboost/issues/5951 .

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
@@ -285,11 +285,11 @@ public class Booster implements Serializable, KryoSerializable {
    * @param predContribs prediction feature contributions
    * @return predict results
    */
-  private synchronized float[][] predict(DMatrix data,
-                                         boolean outputMargin,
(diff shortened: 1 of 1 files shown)


trivialfis:
Hmm, it's not the predict function not being thread-safe, but rather the jvm iterator.

Craigacp:
I think removing the synchronized keyword should be sufficient (at least for our use case in Tribuo). What do you mean by the JVM iterator isn't thread safe?

trivialfis:
It looks like XGBoost is failing to construct `DMatrix` from java iterator for some reasons.

Craigacp:
If you can point me at a test or some example code I can run that down.

trivialfis:
I will take a deeper dive into jvm prediction after https://github.com/dmlc/xgboost/pull/6819 is done.

gangpeng:
@trivialfis 
Hi,  
We would to remove the synchronized API from xgboost4j for only project. We only construct DMatrix from float[] in production, the code path looks thread-safe to me. Is the problem you mentioned only related to iterable?
Thanks!

PR #7028, opened 2021-06-07
Is this the right log loss function?

HanzhangRen:
Seems to be different from p.74 of https://www.slideshare.net/ShangxuanZhang/xgboost.


trivialfis (changes requested):
The one in doc and the one in PPT are the same except for the sign.  Also, I think you have missed 2 minus signs in the rightmost log function.

PR #7031, opened 2021-06-08
Extra wrapper to split data for `eval_set` to use in scikit-learn model_selection and pipeline

manmadetomcat:
Wrapper to split data in order to (along with early stopping):
- Utilise scikit-learn hyperparamter tuning
- Utilise scikit-learn cross validation functions with predefined wrappers
- Place xgboost models in scikit-learn pipelines
- Use scikit-learn model evaluation functions like `sklearn.model_selection.learning_curve`

### DIFF
--- a/python-package/xgboost/__init__.py
+++ b/python-package/xgboost/__init__.py
@@ -16,6 +16,7 @@ from . import dask
 try:
     from .sklearn import XGBModel, XGBClassifier, XGBRegressor, XGBRanker
     from .sklearn import XGBRFClassifier, XGBRFRegressor
+    from .wrapper import get_wrapper
     from .plotting import plot_importance, plot_tree, to_graphviz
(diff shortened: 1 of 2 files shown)


hcho3:
Can you provide a short example of how `get_wrapper` would be used?

manmadetomcat:
> Can you provide a short example of how `get_wrapper` would be used?

You would provide the name of the xgboost model and any model parameters to the `get_wrapper` function and it returns an estimator mostly identical to the xgboost model that splits `X` before calling the model's fit method so early stopping can be used without manually splitting data for `eval_set`.

trivialfis (changes requested):
Thanks for working on this.  However, I don't think we should merge it into XGBoost.  It's just wrapping a call to `model_selection.train_test_split`

Issue #7030, opened 2021-06-08
[CI] Unpin dask version.


trivialfis:
Latest dask is released.

PR #7033, opened 2021-06-09
[CI] Configure RAPIDS, dask, modin

trivialfis:
Close https://github.com/dmlc/xgboost/issues/7030 .


trivialfis:
Hmm, still not propogated.

trivialfis:
@hcho3 Should I simply rerun this after the cache being cleared?

hcho3:
Yes just restart the pipeline.

trivialfis:
I think the mgpu worker still has stalled conda packages.  Also modin complaints about shared memory access...

tests/ci_build/Dockerfile.gpu
```diff
@@ -19,7 +19,7 @@ ENV PATH=/opt/python/bin:$PATH
 # Create new Conda environment with cuDF, Dask, and cuPy
 RUN \
     conda create -n gpu_test -c rapidsai-nightly -c rapidsai -c nvidia -c conda-forge -c defaults \
-        python=3.7 cudf=21.06* rmm=21.06* cudatoolkit=$CUDA_VERSION_ARG dask=2021.05.0 dask-cuda dask-cudf cupy \
+        python=3.7 cudf=21.06* rmm=21.06* cudatoolkit=$CUDA_VERSION_ARG dask>=2021.6.0 dask-cuda dask-cudf cupy \
```
hcho3, line 22: We are currently using the stable 21.06 version of Dask-CUDA, which has a hard dependency on the Dask version 2021.5.1

https://github.com/rapidsai/dask-cuda/blob/ef78a82ac5af8379e17857ae44cb529e3040f1eb/conda/recipes/dask-cuda/meta.yaml#L30

Upgrading to 21.08 nightly version of RAPIDS suite will pick up the latest Dask.

hcho3:
@trivialfis The issue should now be fixed

tests/ci_build/test_python.sh
```diff
@@ -60,6 +60,7 @@ case "$suite" in
   cpu)
     source activate cpu_test
     install_xgboost
+    export RAY_OBJECT_STORE_ALLOW_SLOW_STORAGE=1
```
hcho3, line 63: @trivialfis  It will be probably better to increase the size of `/dev/shm`, as recommended by the error message. I can take care of that. WDYT?

trivialfis: That's a better solution.  Please help increasing the shared memory size.

tests/ci_build/Dockerfile.gpu
trivialfis: Thanks for debugging it.

tests/ci_build/test_python.sh
trivialfis: @hcho3 Maybe we can just go with the current setting and don't think about it, we are not actually doing anything with ray anyway, it's just somehow in the dependency tree.

hcho3 (approved):

Issue #6298, opened 2020-10-28
Flaky test: TestGPUUpdaters::test_categorical


hcho3:
https://xgboost-ci.net/blue/organizations/jenkins/xgboost-win64/detail/master/553/pipeline
https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-6286/2/pipeline

I'm going to file a pull request to temporarily mark the test as XFAIL, as we cannot afford to waste CI resource to restart failed test jobs.

hcho3:
#6299

PR #7036, opened 2021-06-11
Support categorical split in tree model dump.

trivialfis:
Support categorical tree node for text dump.

Related https://github.com/dmlc/xgboost/issues/6503 .
Close https://github.com/dmlc/xgboost/issues/6298 .

### DIFF
--- a/src/predictor/cpu_predictor.cc
+++ b/src/predictor/cpu_predictor.cc
@@ -52,11 +52,6 @@ bst_float PredValue(const SparsePage::Inst &inst,
     if (tree_info[i] == bst_group) {
       auto const &tree = *trees[i];
       bool has_categorical = tree.HasCategoricalSplit();
-
-      auto categories = common::Span<uint32_t const>{tree.GetSplitCategories()};
-      auto split_types = tree.GetSplitTypes();
-      auto categories_ptr =
-          common::Span<RegTree::Segment const>{tree.GetSplitCategoriesPtr()};
       auto cats = tree.GetCategoriesMatrix();
       bst_node_t nidx = -1;
       if (has_categorical) {
--- a/src/tree/tree_model.cc
+++ b/src/tree/tree_model.cc
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2015-2020 by Contributors
+ * Copyright 2015-2021 by Contributors
  * \file tree_model.cc
  * \brief model structure for tree
  */
@@ -74,6 +74,7 @@ class TreeGenerator {
                                 int32_t /*nid*/, uint32_t /*depth*/) const {
     return "";
   }
+  virtual std::string Categorical(RegTree const&, int32_t, uint32_t) const = 0;
   virtual std::string Integer(RegTree const& /*tree*/,
                                 int32_t /*nid*/, uint32_t /*depth*/) const {
     return "";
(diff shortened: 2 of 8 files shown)


src/tree/tree_model.cc
```diff
@@ -92,26 +93,50 @@ class TreeGenerator {
...
     auto const split_index = tree[nid].SplitIndex();
     std::string result;
+    auto is_categorical = tree.GetSplitTypes()[nid] == FeatureType::kCategorical;
     if (split_index < fmap_.Size()) {
       switch (fmap_.TypeOf(split_index)) {
-        case FeatureMap::kIndicator: {
+      case FeatureMap::kCategorical: {
+        auto name = split_index < fmap_.Size()
+                        ? fmap_.Name(split_index)
+                        : "feat" + std::to_string(split_index);
```
hcho3, line 102: Since `name` is only needed for the error message, can you put this inside `if (!is_categorical)` block? Inside, use `LOG(FATAL)` to throw error

hcho3: Also, how come we are not writing anything to `result` for `case FeatureMap::kCategorical:`, like all the other cases?

trivialfis: I kept the possibility that users don't provide a feature map with categorical type.  Categorical is a different split type, all the others are just variations of numerical splits.  So no matter what the feature map says it has to go through a dedicated print procedure.

I can restrict it and ensure a correct feature map is provided.

src/predictor/cpu_predictor.cc
```diff
@@ -52,11 +52,6 @@ bst_float PredValue(const SparsePage::Inst &inst,
     if (tree_info[i] == bst_group) {
       auto const &tree = *trees[i];
       bool has_categorical = tree.HasCategoricalSplit();
-
-      auto categories = common::Span<uint32_t const>{tree.GetSplitCategories()};
-      auto split_types = tree.GetSplitTypes();
-      auto categories_ptr =
-          common::Span<RegTree::Segment const>{tree.GetSplitCategoriesPtr()};
```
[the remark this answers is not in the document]
trivialfis: Yup, the predict instance is never tested and we just keep around for some reasons..

src/tree/tree_model.cc
trivialfis: I remove this logic, now users either not provide a feature map or it has to be consistent with the training data.

trivialfis:
@hcho3 @RAMitchell could you please take a look at how the output should look.  I'm currently printing the categories as sets, but we don't have a `\in` operator here, so I used `=`, which is inaccurate.  A possible alternative is something like `0 = 0 || 1 || 3`.  Here is an example text dump, notice the `0=[0, 14, 32]` part:

```
0:[0={0,14,32}] yes=2,no=1,missing=1,gain=11,cover=2
        1:[f1<1] yes=3,no=4,missing=4,gain=0,cover=0
                3:leaf=0,cover=0
                4:leaf=0,cover=0
        2:[0={0,14,32}] yes=6,no=5,missing=5,gain=11,cover=2
                5:leaf=2,cover=3
                6:leaf=3,cover=4
```
and here's the graphviz output:
figure

trivialfis:
Changed to using colon:

Text dump:
```
0:[0:{0,14,32}] yes=2,no=1,missing=1,gain=11,cover=2
        1:[f1<1] yes=3,no=4,missing=4,gain=0,cover=0
                3:leaf=0,cover=0
                4:leaf=0,cover=0
        2:[0:{0,14,32}] yes=6,no=5,missing=5,gain=11,cover=2
                5:leaf=2,cover=3
                6:leaf=3,cover=4
```

Graphviz:
figure

RAMitchell (approved):
LGTM, the format looks fine to me.

trivialfis:
Hmm .. gpu coordinate descent is failing non-deterministically..

Issue #6091, opened 2020-09-07
Eliminate text parsing from feature importances and evaluation metrics 


hcho3:
Currently, important functions such as feature importances and evaluation metrics rely on parsing of text strings, specifically the text output from the model dump function. For example:

https://github.com/dmlc/xgboost/blob/68c55a37d9bb680fe435f1d011e5fea62be97d22/python-package/xgboost/core.py#L1797-L1832

https://github.com/dmlc/xgboost/blob/68c55a37d9bb680fe435f1d011e5fea62be97d22/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java#L509-L540

https://github.com/dmlc/xgboost/blob/68c55a37d9bb680fe435f1d011e5fea62be97d22/python-package/xgboost/training.py#L85-L91

https://github.com/dmlc/xgboost/blob/68c55a37d9bb680fe435f1d011e5fea62be97d22/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java#L240-L255

Also see https://github.com/dmlc/xgboost/issues/4665#issuecomment-532932603 https://github.com/dmlc/xgboost/issues/4665#issuecomment-532945623

We should aim to eliminate all such uses of text parsing, since a slight change in the text dump will cause all these functions to break.

Proposed replacement:
* Feature importances: Implement a new C++ function that returns a JSON string representing features and their importances.
* Evaluation metrics: Implement a new C++ function that returns a JSON string representing eval set names and their eval metrics.

trivialfis:
Looked into this a little bit.  The implementation isn't difficult, but depends on https://github.com/dmlc/xgboost/pull/6605 due to the use of feature names/types.  I will try to figure out a better way to store those information.

PR #7041, opened 2021-06-15
Implement feature score in GBTree.

trivialfis:
Related: https://github.com/dmlc/xgboost/issues/6091 .

Other than eliminating parsing, this is also for categorical data support.  The old text parsing implementation doesn't understand categorical split outputs.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -2191,7 +2191,9 @@ class Booster(object):
 
         return self.get_score(fmap, importance_type='weight')
 
-    def get_score(self, fmap='', importance_type='weight'):
+    def get_score(
+        self, fmap: os.PathLike = '', importance_type: str = 'weight'
+    ) -> Dict[str, float]:
         """Get feature importance of each feature.
         Importance type can be defined as:
 
@@ -2203,9 +2205,9 @@ class Booster(object):
 
         .. note:: Feature importance is defined only for tree boosters
 
-            Feature importance is only defined when the decision tree model is chosen as base
-            learner (`booster=gbtree`). It is not defined for other base learner types, such
-            as linear learners (`booster=gblinear`).
+            Feature importance is only defined when the decision tree model is chosen as
+            base learner (`booster=gbtree` or `booster=dart`). It is not defined for other
+            base learner types, such as linear learners (`booster=gblinear`).
 
         Parameters
         ----------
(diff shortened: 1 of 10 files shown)


src/c_api/c_api_utils.h
```diff
@@ -181,5 +183,39 @@ class XGBoostAPIGuard {
...
+  if (feature_map.Size() == 0) {
+    // Use the feature names and types from booster.
+    std::vector<std::string> feature_names;
+    learner->GetFeatureNames(&feature_names);
+    std::vector<std::string> feature_types;
+    learner->GetFeatureTypes(&feature_types);
+    for (size_t i = 0; i < n_features; ++i) {
+      feature_map.PushBack(
+          i,
+          maybe(feature_names, i, "f" + std::to_string(i)).data(),
```
wphicks, line 214: For my own education, can you tell me where this `maybe` function comes from and why it's necessary here?

wphicks (approved):
Implementation looks great! Very clean and readable. I had just one question that stems from my own lack of familiarity with the codebase, but everything LGTM.

RAMitchell (approved):

python-package/xgboost/core.py
```diff
@@ -2203,96 +2205,43 @@ def get_score(self, fmap='', importance_type='weight'):
...
+                ctypes.byref(features),
+                ctypes.byref(scores)
+            )
+        )
+        features_arr = from_cstr_to_pystr(features, length)
+        scores_arr = ctypes2numpy(scores, length.value, np.float32)
+        results = {}
+        for feat, score in zip(features_arr, scores_arr):
+            results[feat] = score
+        return results
```
harupy, line 2244: Hi @trivialfis, can we convert `score` to the native python float type here for a use case below?

```python
model  = xgb.train(...)
imp = model. get_score(...)

with open(filepath, "w") as f:
    json.dump(imp, f)  # this line throws if `score` is of `numpy.float32`
```

harupy: Ofcourse, we can do `imp = {k: float(v) for k, v in imp}`

trivialfis: Thanks for pointing that out, I will update with a new PR.

harupy: thanks for the quick action!

PR #7043, opened 2021-06-16
Support categorical data for dask functional interface and DQM.

trivialfis:
Depends on https://github.com/dmlc/xgboost/pull/7042 .

* Implement cat support in GPU sketch merge.
* Add support for DeviceQuantileDMatrix.
* Add support for dask.
* Add support for data iterator.


tests/python-gpu/test_gpu_with_dask.py
```diff
@@ -126,6 +171,56 @@ def run_with_dask_array(DMatrixT: Type, client: Client) -> None:
         inplace_predictions)
 
 
+@pytest.mark.parametrize("rounds", [10])
```
trivialfis, line 174: Better parametrize.

src/common/quantile.cu
```diff
@@ -490,13 +491,37 @@ void SketchContainer::Merge(Span<OffsetT const> d_that_columns_ptr,
...
-            dh::ToSpan(this->Other()), columns_ptr_b_.DeviceSpan());
+  auto feature_types = this->FeatureTypes().ConstDeviceSpan();
+  MergeImpl(device_, this->Data(), this->ColumnsPtr(), that, d_that_columns_ptr,
+            feature_types, dh::ToSpan(this->Other()),
+            columns_ptr_b_.DeviceSpan());
   this->columns_ptr_.Copy(columns_ptr_b_);
   CHECK_EQ(this->columns_ptr_.Size(), num_columns_ + 1);
   this->Alternate();
 
+  if (this->HasCategorical()) {
```
trivialfis, line 502: This is duplicated with `Unique`, try to remove it.

trivialfis:
I think there stalled cache in mgpu worker.

hcho3:
@trivialfis I just cleared the cache.

src/common/quantile.cuh
```diff
@@ -16,6 +16,19 @@ class HistogramCuts;
...
 
+namespace detail {
+struct IsCatOp {
+  XGBOOST_DEVICE bool operator()(FeatureType ft) {
+    return ft == FeatureType::kCategorical;
+  }
+};
+struct SketchUnique {
+  XGBOOST_DEVICE bool operator()(SketchEntry const& a, SketchEntry const& b) const {
+    return a.value - b.value == 0;
```
RAMitchell, line 27: What is this comparator doing?

RAMitchell (approved):

src/common/quantile.cuh
trivialfis: See if they are equal.

RAMitchell: Why not `a.value == b.value`?

trivialfis: If memory is correct, I think that is not guaranteed to work: https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html

PR #7047, opened 2021-06-18
Convert numpy float to Python float in feat score.

trivialfis:
Address issue raised by @harupy in https://github.com/dmlc/xgboost/pull/7041#discussion_r654226008


harupy:
Looks good to me.

PR #7051, opened 2021-06-20
Revert "update libsvm file to start with 1 index"

trivialfis:
This reverts commit a894ab6898f5ed525ca7dec89a79ea1fa7f49f5a.

I went back to 0.70 and the libsvm parser in dmlc-core has always been 0-based.


trivialfis:
Need to fix some indexing mode instead of doing a full  revert.

PR #7053, opened 2021-06-23
Categorical split support for SHAP

trivialfis:
Depends on https://github.com/rapidsai/gputreeshap/pull/26 .

### DIFF
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -415,6 +415,68 @@ class DeviceModel {
   }
 };
 
+struct ShapSplitCondition {
+  ShapSplitCondition() = default;
+  XGBOOST_DEVICE
+  ShapSplitCondition(float feature_lower_bound, float feature_upper_bound,
+                     bool is_missing_branch, common::CatBitField cats)
+      : feature_lower_bound(feature_lower_bound),
+        feature_upper_bound(feature_upper_bound),
+        is_missing_branch(is_missing_branch), categories{std::move(cats)} {
+    assert(feature_lower_bound <= feature_upper_bound);
+  }
+
+  /*! Feature values >= lower and < upper flow down this path. */
+  float feature_lower_bound;
+  float feature_upper_bound;
+  /*! Feature value set to true flow down this path. */
+  common::CatBitField categories;
+  /*! Do missing values flow down this path? */
+  bool is_missing_branch;
+
+  // Does this instance flow down this path?
+  XGBOOST_DEVICE bool EvaluateSplit(float x) const {
+    // is nan
+    if (isnan(x)) {
+      return is_missing_branch;
+    }
+    if (categories.Size() != 0) {
+      auto cat = static_cast<uint32_t>(x);
+      return categories.Check(cat);
+    } else {
+      return x >= feature_lower_bound && x < feature_upper_bound;
+    }
+  }
+
+  // the &= op in bitfiled is per cuda thread, this one loops over the entire
+  // bitfield.
+  XGBOOST_DEVICE static common::CatBitField Intersect(common::CatBitField l,
+                                                      common::CatBitField r) {
+    if (l.Data() == r.Data()) {
+      return l;
+    }
+    if (l.Size() > r.Size()) {
+      thrust::swap(l, r);
+    }
+    for (size_t i = 0; i < r.Bits().size(); ++i) {
+      l.Bits()[i] &= r.Bits()[i];
+    }
+    return l;
+  }
+
+  // Combine two split conditions on the same feature
+  XGBOOST_DEVICE void Merge(ShapSplitCondition other) {
+    // Combine duplicate features
+    if (categories.Size() != 0 || other.categories.Size() != 0) {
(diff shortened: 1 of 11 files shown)


src/predictor/gpu_predictor.cu
```diff
@@ -481,20 +513,45 @@ void ExtractPaths(dh::device_vector<gpu_treeshap::PathElement>* paths,
...
+        if (is_left_path) {
+          for (size_t i = 0; i < node_cats.size(); ++i) {
+            path_cats[i] = ~node_cats[i];
+          }
+        } else {
+          for (size_t i = 0; i < node_cats.size(); ++i) {
+            path_cats[i] = node_cats[i];
+          }
+        }
+        // fixme: distinguish whether it's left path.
```
trivialfis, line 539: Remove this.

src/predictor/gpu_predictor.cu
```diff
@@ -462,14 +465,43 @@ void ExtractPaths(dh::device_vector<gpu_treeshap::PathElement>* paths,
...
+                       return ft == FeatureType::kCategorical;
+                     })) {
+    auto max_elem_it = dh::MakeTransformIterator<size_t>(
+        dh::tbegin(d_cat_node_segments),
+        [] __device__(RegTree::Segment seg) { return seg.size; });
+    size_t max_cat_it =
+        thrust::max_element(thrust::device, max_elem_it,
+                            max_elem_it + d_cat_node_segments.size()) -
+        max_elem_it;
+    max_cat =
```
trivialfis, line 490: Use pinned memory.

src/predictor/gpu_predictor.cu
```diff
@@ -481,20 +513,45 @@ void ExtractPaths(dh::device_vector<gpu_treeshap::PathElement>* paths,
...
-          upper_bound, is_missing_path,     zero_fraction, v};
+
+      float lower_bound = -inf;
+      float upper_bound = inf;
+      gpu_treeshap::Bitset bits;
+      if (common::IsCat(tree.cats.split_type, child.Parent())) {
+        auto path_cats = d_path_categories.subspan(max_cat * output_position, max_cat);
+        size_t size = tree.cats.node_ptr[child.Parent()].size;
+        auto node_cats = tree.cats.categories.subspan(tree.cats.node_ptr[child.Parent()].beg, size);
+        assert(path_cats.size() >= node_cats.size());
```
trivialfis, line 529: span check.

trivialfis:
Please do not merge this one until I can replace the submodule back to rapidsai.

src/predictor/gpu_predictor.cu
```diff
@@ -481,20 +575,44 @@ void ExtractPaths(dh::device_vector<gpu_treeshap::PathElement>* paths,
...
+      float upper_bound = inf;
+      common::CatBitField bits;
+      if (common::IsCat(tree.cats.split_type, child.Parent())) {
+        auto path_cats = d_path_categories.subspan(max_cat * output_position, max_cat);
+        size_t size = tree.cats.node_ptr[child.Parent()].size;
+        auto node_cats = tree.cats.categories.subspan(tree.cats.node_ptr[child.Parent()].beg, size);
+        SPAN_CHECK(path_cats.size() >= node_cats.size());
+        if (is_left_path) {
+          for (size_t i = 0; i < node_cats.size(); ++i) {
+            path_cats[i] = ~node_cats[i];
```
RAMitchell, line 594: `path_cats[i] = is_left_path ? ~node_cats[i] : node_cats[i];`?

RAMitchell (approved):

src/predictor/gpu_predictor.cu
trivialfis: So in the bitset and xgboost predict, values that are set to 1 go to right, 0 go to left, and each bit corresponds to a category.   But when we construct the shap path, there's no left or right in the split condition, only "whether should we go down this path".  As a result, for paths that go to right node, we keep the bits unchanged, but for the paths that go to the left, we flip the bits.

trivialfis: Done.

talieh-tabatabaei:
How Shap values for each one-hot-encoded category are aggregated to generated  the final value for the categorical variables?

oOTWK:
@talieh-tabatabaei When `enable_categorical=True` is enabled, it does not compute Shap values for each one-hot-encoded variable. To enable this for Shap computation, the tree model should be trained with this enabled. In this case, there are no one-hot-encoded variables anymore, it's just one categorical variable itself. So decision variable in the tree model is the categorical variable.

For example, let's say we have Fruit category and the values are Apple, Orange, and Melon. A node of tree could have Fruit==Orange. In one-hot-encoded version, it is equivalent to Is_Orange==1 but in this case, Is_Orange, Is_Apple, and Is_Melon are different variables. In TreeShap, these variables do not invoke `Unwind` to each other. On the other hand, if we just use the categorical variable Fruit, Fruit==Orange and Fruit==Apple and Fruit==Melon all use the same variable Fruit. This does invoke `Unwind` if there is more than one Fruit variable appearing on the current path. As a result, it will only compute Shap value for Fruit variable.

PR #7061, opened 2021-06-24
Fix early stopping behavior with MAPE metric

### LINKED ISSUE #7052: Early stopping with MAPE wrongly chooses the worst iteration as `best_iteration`
On Xgboost python library (version 1.4.2), early-stopping with MAPE should choose an iteration with the minimum MAPE value, however it actually chooses an iteration with the maximum MAPE value.

### Code to reproduce:
```python
from sklearn.datasets import load_boston
from sklearn.model_selection import train_test_split
import xgboost as xgb

boston = load_boston()
X_train, X_val, y_train, y_val = train_test_split(boston.data, boston.target, train_size = 0.7, test_size = 0.3)

# MAPE - wrongly choose the best iteration
m = xgb.XGBRegressor(n_estimators=100)
m.fit(X_train, y_train, eval_set=[(X_val, y_val)], eval_metric=["mape"], early_stopping_rounds=5, verbose=100)

print(m.evals_result())

print(f"Best iteration: {m.best_iteration}")
print(f"Best score: {m.best_score}")
```

### Output:

```
[0]	validation_0-mape:0.67114
[5]	validation_0-mape:0.14712
{'validation_0': OrderedDict([('mape',
               [0.671135, 0.474162, 0.327384, 0.237696, 0.181148, 0.147122])])}
Best iteration: 0
Best score: 0.671135
```

### Expected behaviour:
The best iteration should be 12, and the best score should be 0.108408

hcho3:
Closes #7052

### DIFF
--- a/python-package/xgboost/callback.py
+++ b/python-package/xgboost/callback.py
@@ -549,7 +549,7 @@ class EarlyStopping(TrainingCallback):
             # user to decide.
             maximize_metrics = ('auc', 'aucpr', 'map', 'ndcg', 'auc@',
                                 'aucpr@', 'map@', 'ndcg@')
-            if any(metric.startswith(x) for x in maximize_metrics):
+            if metric != 'mape' and any(metric.startswith(x) for x in maximize_metrics):
                 self.maximize = True
             else:
                 self.maximize = False


python-package/xgboost/callback.py
```diff
@@ -549,7 +549,7 @@ def minimize(new, best):
             # user to decide.
             maximize_metrics = ('auc', 'aucpr', 'map', 'ndcg', 'auc@',
                                 'aucpr@', 'map@', 'ndcg@')
-            if any(metric.startswith(x) for x in maximize_metrics):
+            if metric != 'mape' and any(metric.startswith(x) for x in maximize_metrics):
```
hcho3, line 552: `mape` starts with `map`, so the existing logic would set `maximize=True`.

trivialfis (approved):
We probably want to unify our metric into cost functions ...

Issue #6148, opened 2020-09-22
Training continuation with multiple DMatrix objects


ldesreumaux:
# Issue

If a training is started with a DMatrix object and continued with another DMatrix object that is identical (same dataset...), I should get the same model as if the training had been done with a single DMatrix object.

In the following C++ code, two models are trained. In the first training, UpdateOneIter is called two times with the same DMatrix. In the second training, the two calls to UpdateOneIter are made with two different DMatrix objects that contain the same dataset. I expect the two model dumps (`out_models1` and `out_models2`) to be the same, but they are different.

```C++
#include <xgboost/learner.h>

using namespace xgboost;

void ConfigureBooster(std::shared_ptr<Learner>& booster) {
    std::vector<std::pair<std::string, std::string>> cfg;

    cfg.emplace_back(std::make_pair("tree_method", "hist"));
    cfg.emplace_back(std::make_pair("objective", "binary:logistic"));
    cfg.emplace_back(std::make_pair("subsample", "0.9"));
    cfg.emplace_back(std::make_pair("seed", "42"));
    cfg.emplace_back(std::make_pair("seed_per_iteration", "1"));

    booster->SetParams(cfg);
    booster->Configure();
}

int main(int argc, char** argv) {
    const std::string dataset_path = "../data/census.bin";
    FeatureMap fmap;

    /* Training with 1 DMatrix object */

    std::shared_ptr<DMatrix> dtrain1(DMatrix::Load(dataset_path, true, false));
    std::shared_ptr<Learner> booster1(Learner::Create({dtrain1}));
    ConfigureBooster(booster1);
    booster1->UpdateOneIter(0, dtrain1);
    booster1->UpdateOneIter(1, dtrain1);
    std::vector<std::string> out_models1 = booster1->DumpModel(fmap, true, "text");
    for (std::string out_model : out_models1)
        std::cout << out_model << std::endl;

    /* Training with 2 DMatrix objects (but same dataset!) */

    std::shared_ptr<DMatrix> dtrain2(DMatrix::Load(dataset_path, true, false));
    std::shared_ptr<Learner> booster2(Learner::Create({dtrain2}));
    ConfigureBooster(booster2);
    booster2->UpdateOneIter(0, dtrain2);
    dtrain2.reset(DMatrix::Load(dataset_path, true, false));
    booster2->UpdateOneIter(1, dtrain2);
    std::vector<std::string> out_models2 = booster2->DumpModel(fmap, true, "text");
[...]

PR #7064, opened 2021-06-25
Move `GHistIndex` into `DMatrix`.

trivialfis:
See https://github.com/dmlc/xgboost/issues/4354 for details.  This PR moves `GHistIndex`, we also need to move column matrix in future PRs.  Also related: https://github.com/dmlc/xgboost/issues/5143 , https://github.com/dmlc/xgboost/issues/6148

Few more bonus:
- We can remove the `is_gmat_initialized_` in hist builder.
- We can implement external memory support if desired.
- We can reuse it in approx tree method.

I will update with benchmarks later.

@SmirnovEgorRu @ShvetsKS

### DIFF
--- a/amalgamation/xgboost-all0.cc
+++ b/amalgamation/xgboost-all0.cc
@@ -38,6 +38,7 @@
 #include "../src/data/sparse_page_raw_format.cc"
 #include "../src/data/ellpack_page.cc"
 #include "../src/data/ellpack_page_source.cc"
+#include "../src/data/gradient_index.cc"
 
 // prediction
 #include "../src/predictor/predictor.cc"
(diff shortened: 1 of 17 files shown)


trivialfis:
Here is a quick benchmark, I think there's no performance difference since I'm just moving the code around:

|         | master            | PR                 |
|:--------|:------------------|:-------------------|
| HIGGS   | 181.0964948719969 | 180.65688620665847 |
| Covtype | 70.7731420363416  | 70.98095073600416  |

ShvetsKS:
@trivialfis sorry for long response.
I also remeasured benchmarks and didn't noticed any performance degradations.
Seems this PR simplifies the implementation of quantile cpu DMatrix :)

trivialfis:
Yeah, I'm planning categorical data support for CPU, hoping that I can do some cleanup first instead of piling new code.

PR #7065, opened 2021-06-25
Handle categorical split in model histogram and dataframe.

trivialfis:
* Throw an error in the histogram function if the selected feature is categorical.
* Parse categorical split in model to dataframe function.  Added a column `Category`.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -2400,7 +2400,7 @@ class Booster(object):
             fn = self.feature_names
             if fn is None:
                 # Let xgboost generate the feature names.
-                fn = self.get_fscore().keys()
+                fn = ["f{0}".format(i) for i in range(self.num_features())]
             try:
                 index = fn.index(feature)
                 feature_t = ft[index]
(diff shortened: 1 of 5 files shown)


trivialfis:
I investigated into the last failing test with categorical test in gpu updater, the minor difference in one histogram bin lead to an different tree that can change prediction up to abs(0.5), this is what I got in the evaluation kernel:
```
tid: 3, nidx: 12, feat: 6, max: 89.427734375, bg: -638.998825073, bh: 29.000000000, mg: -0.000015259, mh: 0.000000000, pg: -662.094177246, ph: 31.000000000
tid: 2, nidx: 12, feat: 1, max: 89.426757812, bg: -638.998832703, bh: 29.000000000, mg: -0.000007629, mh: 0.000000000, pg: -662.094177246, ph: 31.000000000
```

There's 1e-5 difference in bg(histogram bin gradient), then xgboost selected a different feature to split on.  I relaxed that test a bit more and added a short note.  But I don't know how the get the test more robust while keeping the accuracy.

RAMitchell (approved):

PR #7066, opened 2021-06-28
update early stop callback check for improvement with tolerance

heyitsmui:
Fixing the improvement with tolerance check, e.g. for maximize, the new score has improvement if it at least improved by {tolerance} compared to the best score, i.e. for the new score to be considered as an improvement, new has to be > best + tolerance, and vice versa for minimize.

### DIFF
--- a/python-package/xgboost/callback.py
+++ b/python-package/xgboost/callback.py
@@ -539,10 +539,10 @@ class EarlyStopping(TrainingCallback):
             return x[0] if isinstance(x, tuple) else x
 
         def maximize(new, best):
-            return numpy.greater(get_s(new) + self._tol, get_s(best))
+            return numpy.greater(get_s(new), get_s(best) + self._tol)
 
         def minimize(new, best):
-            return numpy.greater(get_s(best) + self._tol, get_s(new))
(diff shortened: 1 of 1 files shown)


heyitsmui:
cc @trivialfis @RAMitchell ^

trivialfis:
I think that means something else than tolerance.  In this PR we are not tolerating something instead we are asking for something more.

heyitsmui:
thanks for the response - though I am not quite sure I understand what's considered an improvement here then, e.g. for maximize:

```
def maximize(new, best):
    return numpy.greater(get_s(new) + self._tol, get_s(best))
```
that means if the new score is exactly == current best score, the above will return true (considered as an improvement) because of nonnegative tolerance. Similarly, if the new score is worse than the current best score, and as long as new + tolerance is better than the best score, it will be considered as an improvement which doesn't seem right.

trivialfis:
> the new score is slightly worse than the current best score, but as long as new + tolerance is better than the best score, it will be considered as an improvement which doesn't seem right.

Yes, we are tolerating the error.  I understand your use caes, I'm open to a new parameter for what you are specifying, but maybe we can call it something else?

trivialfis:
min_delta sounds good

PR #7067, opened 2021-06-29
Fix model loading from stream

mpetricek-corp:
Fix bug introduced in 17913713b554d820a8ce94226d854b4a5f1d8bbc (allow loading from byte array)

When loading model from stream, only last buffer read from the input stream is used to construct the model.

This may work for models smaller than 1 MiB (if you are lucky enough to read the whole model at once), but will always fail if the model is larger.

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
@@ -61,7 +61,7 @@ public class XGBoost {
       os.write(buf, 0, size);
     }
     in.close();
-    return Booster.loadModel(buf);
+    return Booster.loadModel(os.toByteArray());
   }
 
   /**


trivialfis:
Thanks for the PR!  Is there a way we can add a test for this?

mpetricek-corp:
> Is there a way we can add a test for this?

Perhaps add a test that creates 1.5MB large model definition and tries to load it? I can try creating such test case, perhaps by crafting some minimal model and (as it is a json) put some generated whitespace inside to make it 1.5 MB large (and it will fail to parse if only first/last 1 MB is seen), then send it via a stream.

wbo4958:
@mpetricek-corp Better to add a unit test for your fix. Thx very much

trivialfis:
Can we generate the model with random data.  To get a large model, easiest way is to use classification with random forest.

trivialfis:
Merged since it's difficult to create unittest and is confirmed to work.

PR #7079, opened 2021-07-05
Extract evaluate splits from CPU hist.

trivialfis:
Other than modularizing the split evaluation function, this PR also removes some more functions including `InitNewNodes` and `BuildNodeStats` among some other unused variables.  Also, scattered code like setting leaf weights is grouped into the split evaluator and `NodeEntry` is simplified and made private.  Another subtle difference with the original implementation is that the modified code doesn't call `tree[nidx].Parent()` to traversal upward.

## perf

I think it's faster after the PR.  URL is only run for 100 rounds, others are 500 rounds.

|        | HIGGS              | covtype           | URL                |
|:-------|:-------------------|:------------------|:-------------------|
| master | 180.24662968399934 | 76.94775568600744 | 196.1527223587036  |
| PR     | 177.9364239120041  | 68.41261254699202 | 180.37877798080444 |

### DIFF
--- a/src/tree/hist/evaluate_splits.h
+++ b/src/tree/hist/evaluate_splits.h
@@ -21,14 +21,12 @@ namespace xgboost {
 namespace tree {
 
 template <typename GradientSumT, typename ExpandEntry> class HistEvaluator {
- public:
+ private:
   struct NodeEntry {
     /*! \brief statics for node entry */
     GradStats stats;
     /*! \brief loss of this node, without split */
     bst_float root_gain{0.0f};
-    /*! \brief weight calculated related to current data */
-    float weight{0.0f};
   };
 
  private:
--- a/src/tree/hist/evaluate_splits.h
+++ b/src/tree/hist/evaluate_splits.h
@@ -33,7 +33,7 @@ template <typename GradientSumT, typename ExpandEntry> class HistEvaluator {
 
  private:
   TrainParam param_;
-  common::ColumnSampler column_sampler_;
+  std::shared_ptr<common::ColumnSampler> column_sampler_;
   TreeEvaluator tree_evaluator_;
   int32_t n_threads_ {0};
   FeatureInteractionConstraintHost interaction_constraints_;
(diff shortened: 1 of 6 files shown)


trivialfis:
cc @ShvetsKS @SmirnovEgorRu

ShvetsKS:
> cc @ShvetsKS @SmirnovEgorRu
> 
> The `unused_rows_` is not used in the builder, but used in the test. I think it's legacy?

yes. It's not used now,  gradients are set to zero instead.

trivialfis:
@ShvetsKS Thanks for the reply.  Could you help review this PR?  I will remove the `unused_rows_` in upcoming refactors as I'm migrating the approx implementation to the fast hist code base.

SmirnovEgorRu (approved):
LGTM

hcho3 (approved):
LGTM. Two comments:
* It's nice that the CPU Hist is becoming more like GPU Hist. Consistency makes the whole codebase more legible.
* "Evaluator" is a nice abstraction. I like how we no longer need a mock for testing split evaluation.

src/tree/updater_quantile_hist.cc
```diff
@@ -484,20 +472,11 @@ void QuantileHistMaker::Builder<GradientSumT>::ExpandTree(
...
-        const int cleft = (*p_tree)[nid].LeftChild();
-        const int cright = (*p_tree)[nid].RightChild();
-        CPUExpandEntry left_node = nodes_to_evaluate[i*2 + 0];
-        CPUExpandEntry right_node = nodes_to_evaluate[i*2 + 1];
-
-        left_node.loss_chg = snode_[cleft].best.loss_chg;
-        right_node.loss_chg = snode_[cright].best.loss_chg;
-
+        CPUExpandEntry left_node = nodes_to_evaluate.at(i * 2 + 0);
+        CPUExpandEntry right_node = nodes_to_evaluate.at(i * 2 + 1);
```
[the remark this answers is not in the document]
trivialfis: It's not critical.

src/tree/hist/evaluate_splits.h
```diff
@@ -0,0 +1,268 @@
...
+    p_best->Update(best);
+
+    return e;
+  }
+
+ public:
+  void EvaluateSplits(const common::HistCollection<GradientSumT> &hist,
+                      GHistIndexMatrix const &gidx, const RegTree &tree,
+                      std::vector<ExpandEntry>* p_entries) {
+    auto& entries = *p_entries;
```
ShvetsKS, line 137: is it possible to enable timer for this kernel like it was before `builder_monitor_.Stop("EvaluateSplits");`?

trivialfis: Done.

ShvetsKS:
LGTM

PR #7081, opened 2021-07-06
Support hessian in host sketch container.

trivialfis:
* Implement some utilities for threads to reduce dependency on `omp_get_max_threads`.
* Merge hessian with sample weights.

### DIFF
--- a/src/common/quantile.cc
+++ b/src/common/quantile.cc
@@ -10,19 +10,21 @@ namespace xgboost {
 namespace common {
 
 HostSketchContainer::HostSketchContainer(std::vector<bst_row_t> columns_size,
-                                         int32_t max_bins, bool use_group)
+                                         int32_t max_bins, bool use_group,
+                                         int32_t n_threads)
     : columns_size_{std::move(columns_size)}, max_bins_{max_bins},
-      use_group_ind_{use_group} {
+      use_group_ind_{use_group}, n_threads_{n_threads} {
   monitor_.Init(__func__);
   CHECK_NE(columns_size_.size(), 0);
   sketches_.resize(columns_size_.size());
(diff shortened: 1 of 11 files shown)


tests/cpp/common/test_hist_util.cc
```diff
@@ -226,6 +226,26 @@ TEST(HistUtil, DenseCutsAccuracyTestWeights) {
   }
 }
 
+TEST(HistUtil, QuantileWithHessian) {
```
trivialfis, line 229: Compare the result with multiplying Hessian directly.

trivialfis:
This is part of refactoring the approx tree method.  In the implementation of current `approx` the sample weight is ignored, but the impl in this PR multiples the sample weight with hessian.

src/common/quantile.cc
```diff
@@ -89,40 +91,94 @@ std::vector<bst_feature_t> HostSketchContainer::LoadBalance(
...
+    auto const &group_weights = info.weights_.HostVector();
+    CHECK_GE(group_ptr.size(), 2);
+    CHECK_EQ(group_ptr.back(), hessian.size());
+    size_t cur_group = 0;
+    for (size_t i = 0; i < hessian.size(); ++i) {
+      results[i] = hessian[i] * group_weights[cur_group];
+      if (i == group_ptr[cur_group + 1]) {
+        cur_group++;
+      }
+    }
```
hcho3, line 112: This portion is sequential for now?

hcho3 (approved):

src/common/quantile.cc
trivialfis: It's either sequential or binary search.

PR #7084, opened 2021-07-06
Update reference to treelite website

JeffHCross:
treelite.io is no longer a valid site and re-directs users to a parked domain. Re-directing to the documentation is safer at this point.


trivialfis:
@hcho3

hcho3 (approved):

PR #7086, opened 2021-07-07
Fix CSR Array adapter batch columns.

trivialfis:
The function isn't used in current codebase.

Part of https://github.com/dmlc/xgboost/pull/7070 .


trivialfis:
Will merge it into another PR.

PR #7089, opened 2021-07-07
Refactor external memory formats.

trivialfis:
* Save base_rowid.
* Return write size.
* Remove unused function.
* Add tests.

Part of https://github.com/dmlc/xgboost/pull/7070 .

### DIFF
--- a/tests/cpp/data/test_ellpack_page_raw_format.cu
+++ b/tests/cpp/data/test_ellpack_page_raw_format.cu
@@ -1,0 +1,45 @@
+/*!
+ * Copyright 2021 XGBoost contributors
+ */
+#include <gtest/gtest.h>
+#include <dmlc/filesystem.h>
+#include <xgboost/data.h>
+
+#include "../../../src/data/sparse_page_source.h"
+#include "../../../src/data/ellpack_page.cuh"
+
+#include "../helpers.h"
+
+namespace xgboost {
+namespace data {
+TEST(EllpackPageRawFormat, IO) {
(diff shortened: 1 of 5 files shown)


tests/cpp/data/test_ellpack_page_raw_format.cu
```diff
@@ -0,0 +1,45 @@
...
+  for (auto const &ellpack : m->GetBatches<EllpackPage>({0, 256})) {
+    auto loaded = page.Impl();
+    auto orig = ellpack.Impl();
+    ASSERT_EQ(loaded->Cuts().Ptrs(), orig->Cuts().Ptrs());
+    ASSERT_EQ(loaded->Cuts().MinValues(), orig->Cuts().MinValues());
+    ASSERT_EQ(loaded->Cuts().Values(), orig->Cuts().Values());
+    ASSERT_EQ(loaded->base_rowid, orig->base_rowid);
+    ASSERT_EQ(loaded->row_stride, orig->row_stride);
+    ASSERT_EQ(loaded->gidx_buffer.HostVector(), orig->gidx_buffer.HostVector());
+  }
```
hcho3, line 42: Are we assuming that `m->GetBatches` would only return a single page? Otherwise, we'd be asserting that multiple pages be identical to `page`.

trivialfis: Yes, this is a simple dmatrix created on line 18.

hcho3: Got it.

hcho3 (approved):

PR #7091, opened 2021-07-08
Check input value is duplicated when quantile queue is full

oOTWK:
When the last value of `queue` is equal to the input value `x`, its weight is added to the weight of the last entry of `queue` instead of creating a new entry. However, this is ignored when the `queue` is full.


trivialfis (approved):
Thanks for the PR!

PR #7097, opened 2021-07-09
fix typo in `XGDMatrixSetFloatInfo` example

ZabelTech:
hi,
this patch fixes a small typo that i noticed while following along the c api examples.


trivialfis (approved):

trivialfis:
Thanks for the fix!

PR #7099, opened 2021-07-10
Optional find dependency in installed cmake config.

trivialfis:
Following up https://github.com/dmlc/xgboost/pull/6159 .  This PR removes `find_dependency` in CMake config and make sure dependencies are private if xgboost is built as shared library (the default).

An issue was found during writing another C demo for external memory, where I have to set the project with `CXX` language to get CMake working (but I'm writing a demo for C).  Also `FindNccl` is a custom script in xgboost which is not available as part of CMake, adding it to dependency will break users' code.

After the PR we simply include gputreeshap header instead of linking it due to a CMake error saying it's not in the export set when built as a static library.

Other than these, I added a simple script using `readelf` to show that the C demo is not linking anything from the dependency set of xgboost.


rabit/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ find_package(Threads REQUIRED)
 
 set(RABIT_SOURCES
   ${CMAKE_CURRENT_LIST_DIR}/src/allreduce_base.cc
-  ${CMAKE_CURRENT_LIST_DIR}/src/c_api.cc)
+  ${CMAKE_CURRENT_LIST_DIR}/src/rabit_c_api.cc)
```
trivialfis, line 7: msvc complains 2 files having same name might produce incorrect result.  So I renamed the c api file and merged  the test_io.cc into xgboost.

hcho3 (approved):

Issue #7039, opened 2021-06-14
Segmentation fault on MacOS with libomp 12.0.0


fractor:
The following code results in a Python segmentation fault when executed with Python 3.x on MacOS with libomp 12.0.0. Downgrading to a previous version of libomp solves it.

```
# xgboost macos seg fault when libomp==12.0.0
# passes with libomp==11.1.0

"""
	verify libomp version
		brew list --version libomp
		libomp 12.0.0
	to install libomp 11.1.0
		wget https://raw.githubusercontent.com/chenrui333/homebrew-core/0094d1513ce9e2e85e07443b8b5930ad298aad91/Formula/libomp.rb
        brew unlink libomp
        brew install --build-from-source ./libomp.rb
        brew list --version libomp

"""

import logging
import numpy as np
import xgboost as xgb

def xgboost_unit_tests():
	feature_arr = np.random.uniform(size=(100,5))
	label_arr = np.random.randint(low=0, high=10, size=(100,1))
	data_arr = np.concatenate((feature_arr, label_arr), axis=1)
	#Run Attribute Rank with -O 2
	run_xgboost(data_arr,data_arr)
	logging.info("XGBoost complete")
	return


def run_xgboost(trainarr, valarr):
	trainfeats = trainarr[:,:-1]
	trainlabels = trainarr[:,-1]
	# first seg fault
	dtrain = xgb.DMatrix(trainfeats, label=trainlabels)
	valfeats = valarr[:,:-1]
	vallabels = valarr[:,-1]
	# second seg fault
	dval = xgb.DMatrix(valfeats, label=vallabels)
	return


# Press the green button in the gutter to run the script.
if __name__ == '__main__':
	xgboost_unit_tests()
```

andy-brainome:
Just to be clear - when everything was paired down to these 40 some lines of code, the seg fault 11 was replaced with the dumping of source code to the screen. i.e. "warnings.warn(".  This line of code has shifted up or down a couple of times.
```
andys@MacBook-Pro:~/work/xgboost-macos-seg-fault-11$ python3 main.py
/Users/andys/.pyenv/versions/3.8.7/lib/python3.8/site-packages/xgboost/data.py:104: UserWarning: Use subset (sliced data) of np.ndarray is not recommended because it will generate extra copies and increase memory consumption
  warnings.warn(
andys@MacBook-Pro:~/work/xgboost-macos-seg-fault-11$
```
[...]

PR #7107, opened 2021-07-14
[CI] Pin libomp to 11.1.0 

hcho3:
Works around #7039 by using libomp 11.1.0 in CI.


hcho3:
@trivialfis This should fix the problem where OpenMP runtime hangs or crashes.

trivialfis (approved):

PR #7109, opened 2021-07-15
[CI] freeze dask version for now.

trivialfis:
Freeze dask to be 06.2 for now to unblock the CI.

I haven't been able to find the root cause of the failure,  will try to bisect it later.  It's just way too weird.


trivialfis:
Seems failing again.

PR #7117, opened 2021-07-21
Forbid overlap in feature interaction constraints

### LINKED ISSUE #7115: last interaction constraint example seems wrong
In https://github.com/dmlc/xgboost/issues/4075#issuecomment-458280942, a very precise definition of feature interaction constraints is given. In my view, the last example in https://xgboost.readthedocs.io/en/latest/tutorials/feature_interaction_constraint.html 

image

is wrong (if the definition is correct). 

Why? The ancestor nodes on the right-most branch have feature set {x_0, x_1}. So the last split should also be on either x_0 or x_1.

Am I wrong?

hcho3:
Closes #7115

I see this as a short-term fix, until we can figure out a better way to handle interaction constraints with overlapping members.

Ideally, we want to keep track of all feature IDs that appear in the path from the root node to each leaf node, like LightGBM: https://github.com/microsoft/LightGBM/issues/4481. Current XGBoost implementation only considers the immediate node that are being split.

cc @mayer79

### DIFF
--- a/src/tree/param.cc
+++ b/src/tree/param.cc
@@ -2,6 +2,7 @@
  * Copyright by Contributors 2019
  */
 #include <iostream>
+#include <algorithm>
 #include <vector>
 #include <utility>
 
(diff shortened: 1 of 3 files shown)


mayer79:
@hcho3: Thanks for the new error message. One thing I wanted to ask: The R interface allows to specify the argument without quotes, i.e. `interaction_constraints = list(c(0, 1), c(2, 3, 4))`. Are quotes in Python mandatory or would a nested list also work?

trivialfis (commented):
- This is a breaking change since we have documented the behaviour and added tests.  It's now a "feature" instead of a bug ...
- I don't have a strong preference as to whether overlapping features are "correct" or should be banned.
- If we were to make the breaking change, maybe we can use `std::set` instead of creating 3 loops?
- If we were to make the breaking change, we also need to update the documents and tests.

mayer79:
> * This is a breaking change since we have documented the behaviour and added tests.  It's now a "feature" instead of a bug ...
> * I don't have a strong preference as to whether overlapping features are "correct" or should be banned.
> * If we were to make the breaking change, maybe we can use `std::set` or `std::unique` instead of creating 3 loops?
> * If we were to make the breaking change, we also need to update the documents and tests.

What about replacing the fatal error by a warning?

hcho3:
Closing for now. We don't yet have a bandwidth to properly change the documented behavior of interaction constraints.

PR #7120, opened 2021-07-21
Remove extra sync for dense data

ShvetsKS:


trivialfis (commented):
Thanks for the quick fix!  Could you please have some tests?

ShvetsKS:
> Thanks for the quick fix! Could you please have some tests?

done

trivialfis (approved):
Will merge once the tests have passed.


TODO: Follow up with tests for other tree methods.

ShvetsKS:
> TODO: Follow up with tests for other tree methods.

just added the parameters for `approx` also

Issue #7098, opened 2021-07-09
nthread inside xgboost() function not working in R


cmeuli07:
The nthread parameter is evidently being ignored in R. When setting nthread = 1 xgboost appears to be defaulting to the maximum number of threads on my machine because my cpu is instantly maxing out. I've included a reprex and my session info below:

Reprex:
```
require(dplyr)
require(xgboost)


df1 <- sample_frac(iris, size = 500, replace = T) %>% 
  mutate(species_num = dense_rank(Species))

xgb2 <- xgboost(data = data.matrix(select(df1, -Species, -Sepal.Length)),
                  label = data.matrix(select(df1, Sepal.Length)),
                  colsample_bytree = 0.5,
                  nrounds = 25000,
                  min_child_weight = 50,
                  max_depth = 14,
                  eta = 0.1,
                  gamma = 0,
                  subsample = 0.8,
                  nthread = 1L,
                  early_stopping_rounds = 24000,
                  verbose = 0
                  )
```

Session Info:
```
> sessionInfo()
R version 4.0.4 (2021-02-15)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19043)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                           LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] xgboost_1.5.0.1 dplyr_1.0.7    

loaded via a namespace (and not attached):
 [1] lattice_0.20-41   fansi_0.4.2       assertthat_0.2.1  utf8_1.1.4        crayon_1.4.1      grid_4.0.4        R6_2.5.0         
 [8] jsonlite_1.7.2    DBI_1.1.1         lifecycle_1.0.0   magrittr_2.0.1    pillar_1.6.1      rlang_0.4.10      rstudioapi_0.13  
[15] data.table_1.14.0 Matrix_1.3-2      vctrs_0.3.8       generics_0.1.0    ellipsis_0.3.2    tools_4.0.4       glue_1.4.2       
[22] purrr_0.3.4       compiler_4.0.4    pkgconfig_2.0.3   tidyselect_1.1.1  tibble_3.1.2  
```

trivialfis:
[...]

PR #7127, opened 2021-07-24
[R] Fix nthread in DMatrix constructor.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/7098 .

Restricting the construction of DMatrix to be single threaded.

### DIFF
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -103,7 +103,9 @@ XGB_DLL SEXP XGDMatrixCreateFromMat_R(SEXP mat, SEXP missing, SEXP n_threads) {
   }
   exc.Rethrow();
   DMatrixHandle handle;
-  CHECK_CALL(XGDMatrixCreateFromMat_omp(BeginPtr(data), nrow, ncol, asReal(missing), &handle, asInteger(n_threads)));
+  CHECK_CALL(XGDMatrixCreateFromMat_omp(BeginPtr(data), nrow, ncol,
+                                        asReal(missing), &handle,
+                                        asInteger(n_threads)));
   ret = PROTECT(R_MakeExternalPtr(handle, R_NilValue, R_NilValue));
   R_RegisterCFinalizerEx(ret, _DMatrixFinalizer, TRUE);
   R_API_END();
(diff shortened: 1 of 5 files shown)


hcho3:
Is it possible to make `XGDMatrixCreateFromMat_R` aware of `nthread` setting?

trivialfis:
That's either a new API or a breaking change.

hcho3:
@trivialfis I don't think changing the signature for `XGDMatrixCreateFromMat_R` would break any user-facing functionality. We'd just need to modify `xgb.DMatrix.R`. What do you think?

trivialfis:
I was told that the C function in R is also considered as part of the interface when working on https://github.com/dmlc/xgboost/pull/6819 .  That's why the old predict C function for R is still here.

RAMitchell (approved):

R-package/src/xgboost_R.cc
```diff
@@ -103,7 +103,7 @@ XGB_DLL SEXP XGDMatrixCreateFromMat_R(SEXP mat, SEXP missing) {
   }
```
hcho3, line 103: The for loop in lines 96-102 is still using multiple threads. We must revise the loop also.

hcho3 (changes requested):

hcho3:
Also, let's create a new API function, `XGDMatrixCreateFromMat_omp_R`, that takes in the extra parameter `nthread`, so that R users do not lose out on the ability to use multiple threads when loading from a matrix.

R-package/src/xgboost_R.cc
trivialfis: It's actually using the default.

> XGDMatrixCreateFromMat_omp_R, that takes in the extra parameter nthread

Maybe a new JSON based API.

hcho3: "The default" is set by `OMP_NUM_THREADS`, not `nthread.

hcho3:
Let's just change the signature for `XGDMatrixCreateFromMat_R`, since this function is only used by the R package.

RAMitchell (approved):

PR #7131, opened 2021-07-26
Allow compilation with nvcc 11.4

robertmaynard:
This allows xgboost to compile with nvcc 11.4

I haven't looked into `dh::ArgSort` fully, but if it requires full 64bit support we can't update to cub 1.12+ due tohttps://github.com/NVIDIA/cub/issues/336. If xgboost is happy with 32bit space we can support cub 1.12+ by updating `ArgSort` to explicitly use `int` for offset type.

### DIFF
--- a/src/data/ellpack_page.cu
+++ b/src/data/ellpack_page.cu
@@ -185,10 +185,7 @@ struct TupleScanOp {
 
 // Change the value type of thrust discard iterator so we can use it with cub
 template <typename T>
-class TypedDiscard : public thrust::discard_iterator<T> {
- public:
-  using value_type = T;  // NOLINT
-};
+using TypedDiscard = thrust::discard_iterator<T>;
 
 // Here the data is already correctly ordered and simply needs to be compacted
 // to remove missing data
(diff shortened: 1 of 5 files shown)


hcho3:
@robertmaynard  XGBoost needs full 64-bit support for `ArgSort`: https://github.com/dmlc/xgboost/blob/44cc9c04eafc10f34306c9f69eee88d297eaac49/src/metric/auc.cc#L49-L50

hcho3 (approved):

src/data/ellpack_page.cu
```diff
@@ -185,10 +185,7 @@ struct TupleScanOp {
 
 // Change the value type of thrust discard iterator so we can use it with cub
 template <typename T>
-class TypedDiscard : public thrust::discard_iterator<T> {
- public:
-  using value_type = T;  // NOLINT
-};
+using TypedDiscard = thrust::discard_iterator<T>;
```
RAMitchell, line 188: The purpose of this class is to change the value type away from void, in order to work with cub/thrust in certain edge cases. If it is no longer necessary, maybe cub/thrust has changed upstream? In which case we don't need the using statement we can just use the discard director directly.

trivialfis: Maybe we can make the change when we drop cuda 11.2 support.  But at the moment we still need it.

trivialfis (approved):
Thanks for the PR!

PR #7132, opened 2021-07-27
[JVM-Packages] Use Python tracker in XGBoost for JVM package.

trivialfis:
The old dmlc tracker is causing test failure https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/master/931/pipeline .  See if we can take this as a chance to merge 2 trackers.


wbo4958:
It's really good to maintain one tracker for all language bindings, But it will make jvm package depends on python binding language then.

+1 for this PR

wbo4958 (approved):

Issue #7136, opened 2021-07-28
EarlyStopping can incorrectly detect metric improvement with positive tolerance


MarkGuryanov:
Hello!

The logic of class [`xgboost.callbacks.EarlyStopping`](https://github.com/dmlc/xgboost/blob/92ae3abc972c100a11d7e75fb6dbeebdb7cefd1e/python-package/xgboost/callback.py#L471) determining whether the metric improved seems to be incorrect if parameter `abs_tol` is graeter than 0. Namely, its private method [`_update_rounds`](https://github.com/dmlc/xgboost/blob/92ae3abc972c100a11d7e75fb6dbeebdb7cefd1e/python-package/xgboost/callback.py#L536) defines two helper functions, [`maximize`](https://github.com/dmlc/xgboost/blob/92ae3abc972c100a11d7e75fb6dbeebdb7cefd1e/python-package/xgboost/callback.py#L541) and [`minimize`](https://github.com/dmlc/xgboost/blob/92ae3abc972c100a11d7e75fb6dbeebdb7cefd1e/python-package/xgboost/callback.py#L544), which erroneously determine the event of metric improvement.

In case when metric is intended to be maximized, the following code decides whether it has improved (became greater than previous best):
```python
def maximize(new, best):
    return numpy.greater(get_s(new) + self._tol, get_s(best))
```
This function returns `True` if `new + tol > best`, but if `tol > 0`, this can be true even when `new <= best` (for example, `new == best - tol/2` so `new + tol == best + tol/2 > best`), but obviously this is not the case of metric improvement.

Similarly, when metric is intended to be minimized, the following code decides whether it has improved (became less than previous best):
```python
def minimize(new, best):
    return numpy.greater(get_s(best) + self._tol, get_s(new))
```
This function returns `True` if `best + tol > new`, but again, in case `tol > 0` and (as an example) `new == best + tol/2` this condition is true, but `new` is actually worse (greater) than `best`.

This means that when new metric value is slightly worse (not more than `abs_tol`) than previous best, `EarlyStopping` considers it as improvement and saves this value as new best (which is actually worse). And this can result in boosting slightly worsening the model each iteration, but `EarlyStopping` will consider it as an improvement, thus skipping the moment when the model was truly best and yielding incorrect `best_iteration` and so on.

Fortunately, this behavior can be fixed by substituting `+` with `-` in these two helper methods, so it becomes:
* `maximize`: improves if `new - tol > best` (so `new > best + tol`);
[...]

PR #7137, opened 2021-07-28
Support `min_delta` in early stopping.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/7136 .
Close https://github.com/dmlc/xgboost/pull/7066 .

### DIFF
--- a/python-package/xgboost/callback.py
+++ b/python-package/xgboost/callback.py
@@ -485,13 +485,8 @@ class EarlyStopping(TrainingCallback):
         Whether to maximize evaluation metric.  None means auto (discouraged).
     save_best
         Whether training should return the best model or the last model.
-    abs_tol
-        Absolute tolerance for early stopping condition.
-
-        .. versionadded:: 1.5.0
-
     min_delta
-        Minimum change in score, can only be set if abs_tol is 0.
+        Minimum absolute change in score to be qualified as an improvement.
 
         .. versionadded:: 1.5.0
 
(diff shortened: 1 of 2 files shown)


MarkGuryanov:
@trivialfis I'm not sure why we need to have both `abs_tol` and `min_delta`. Maybe I misunderstand the idea of `abs_tol`, but I thought of it as minimum change in the monitored metric in the direction of improvement to qualify as a substantial improvement. In this case we must not tolerate the event when metric gets worse and surely we must not consider this as an "improvement", because as I've described it in #7136, if we do this, the model will continue to train while metric on evaluation set gets slightly worse each step, and in the long run it will lead to overfitting, effectively nullifying the early stopping mechanism. Thus we must expect the metric to be strictly greater than previous best to consider it as an "improvement", 

It will be useful to rename parameter `abs_tol` to `min_delta` and fix the calculations (either by switching `+` to `-` or as it is done in #7066 - mathematically it is the same), but having both of them and requiring at least one to be 0 will be quite misleading for users IMHO.

Also, looking into APIs of early stopping callbacks from other frameworks (for example, [pytorch-lightning](https://pytorch-lightning.readthedocs.io/en/latest/api/pytorch_lightning.callbacks.early_stopping.html#pytorch_lightning.callbacks.early_stopping.EarlyStopping) and [tensorflow](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/EarlyStopping#args)), I see only one parameter `min_delta` and no tolerance parameter.

trivialfis:
My initial thought is just to allow some turbulence in validation results during training.  If that's not useful and doesn't make sense I can remove the `abs_tol` completely.

trivialfis:
Your argument make sense to me.  Will follow up with the removal.

trivialfis:
Removed.

MarkGuryanov (approved):
I think everything is fine now. Thank you!

RAMitchell (approved):

Issue #6719, opened 2021-02-21
Support parquet or other binary format for external memory


PeterPanonGit:
Currently XGBoost external memory only support libsvm and csv file as input file. However these types of text files are really slow to write and read when the data size is large. Is there a roadmap for xgboost to support more efficient data input format for external memory such as parquet, hdf5, etc?

trivialfis:
I'm trying to move the external memory support to 3 party libraries by making `DMatrix` to support iterators, a implementation for GPU is available https://github.com/dmlc/xgboost/blob/master/demo/guide-python/data_iterator.py .  Others still need some more work.

trivialfis:
Please check out the new iterator interface, with which you can load the data with any library you want.

PR #7139, opened 2021-07-28
Add parquet build option

bridgream:
With the [newly added Parquet support to dmlc-core](https://github.com/dmlc/dmlc-core/pull/653), we only need to add the build flag in CMakeLists to enable this feature in XGBoost as well. 
Related: #6719 , although we already have an iterator interface solution.


trivialfis:
Dealing with MSVC is always fun. ;-)

bridgream:
Thanks for helping! Let me know if extra tests or edits are needed.

trivialfis:
Hmm .. dmlc unittest is not playing well with unity build for some reason.

trivialfis:
I just built XGBoost with latest dmlc core using unity build and VC 2019.  Maybe we should upgrade the version on CI

trivialfis:
Thanks for the great work!  I managed to update dmlc core so this PR is closed instead.

PR #7148, opened 2021-07-30
[jvm-packages] Change to new predict API

wbo4958:
1. deprecated all old predict APIs
2. deprecated treeLimit param
3. add new params: training/iterationBegin/iterationEnd/iterationEnd/strictShape

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Tensor.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Tensor.java
@@ -1,0 +1,156 @@
+/*
+ Copyright (c) 2021 by Contributors
+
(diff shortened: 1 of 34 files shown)


jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Tensor.java
```diff
@@ -0,0 +1,156 @@
...
+
+package ml.dmlc.xgboost4j.java;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A Tensor to hold the prediction result.
+ */
+public class Tensor {
```
Craigacp, line 25: Could this class have a direct accessor for each element which does the appropriate indexing? And some documentation which states it's row-major, as otherwise the `getRawResult` function isn't really documented.

PR #7159, opened 2021-08-06
Fix prediction configuration.

trivialfis:
After the `predictor` parameter is added to the constructor, this configuration was broken.


trivialfis:
@RAMitchell @hcho3 Could you please take a quick look?

RAMitchell (approved):

PR #7160, opened 2021-08-06
Include `cpack` to generate source packages

ajschmidt8:
This PR adds `include(CPack)` to CMakeLists.txt to support generating source packages for `xgboost`.

cc: @trxcllnt


trivialfis (approved):
Thanks for the PR!

Issue #7156, opened 2021-08-04
Segmentation Fault with CV2 import 


8ern4ard:
The following code creates a Fatal Python error: Segmentation fault. 
- XGBoost 1.4.0 and opencv 4.5.2 in conda on Ubuntu 20.04.2 LTS (conda-forge)

It does not do so when (i) cv2 is not imported or (ii) n_jobs  = 1 or 2 on a machine with 2 CPUs or (iii) n_job = 1 on a machine with 16 CPUs. Also, I can't reproduce it on MacOS BigSur (same package versions, also conda-forge). 

I tried to build XGBoost from source, with no luck. 
In my specific case the workaround is to only import cv2 within the function that needs it, 
so by the time the code reaches XGBoost it is out of scope. 

gdb and faulthandler traces below.

**Crash location**
As shown below it originates in /xgboost/data.py", line 155 in _from_numpy_array:
```
_check_call(_LIB.XGDMatrixCreateFromMat_omp(
        flatten.ctypes.data_as(ctypes.POINTER(ctypes.c_float)),
        c_bst_ulong(data.shape[0]),
        c_bst_ulong(data.shape[1]),
        ctypes.c_float(missing),
        ctypes.byref(handle),
        ctypes.c_int(nthread)))
```

**Demo Code**
```
import cv2

import xgboost
import numpy as np

import faulthandler
faulthandler.enable()

model = xgboost.XGBClassifier(n_jobs=3)

X = np.asarray([[0, 0, 0], [1, 1, 1], [2, 2, 2]])
y = np.asarray([0, 1, 2])
model.fit(X, y)
print(model.predict(X))
```

**faulthandler:**
```
Current thread 0x00007f8d07a9b740 (most recent call first):
  File "/home/ubuntu/miniconda3/envs/automap21/lib/python3.9/site-packages/xgboost/data.py", line 155 in _from_numpy_array
  File "/home/ubuntu/miniconda3/envs/automap21/lib/python3.9/site-packages/xgboost/data.py", line 556 in dispatch_data_backend
  File "/home/ubuntu/miniconda3/envs/automap21/lib/python3.9/site-packages/xgboost/core.py", line 538 in __init__
  File "/home/ubuntu/miniconda3/envs/automap21/lib/python3.9/site-packages/xgboost/core.py", line 433 in inner_f
[...]

PR #7161, opened 2021-08-06
Work around a segfault observed in SparsePage::Push()

hcho3:
Work around #7156, by removing `resize()` of the thread-local vector inside the OpenMP block.

### DIFF
--- a/.github/workflows/r_tests.yml
+++ b/.github/workflows/r_tests.yml
@@ -4,6 +4,7 @@ on: [push, pull_request]
 
 env:
   R_PACKAGES: c('XML', 'igraph', 'data.table', 'ggplot2', 'DiagrammeR', 'Ckmeans.1d.dp', 'vcd', 'testthat', 'lintr', 'knitr', 'rmarkdown', 'e1071', 'cplm', 'devtools', 'float', 'titanic')
+  GITHUB_PAT: ${{ secrets.PUBLIC_REPO_ACCESS_GITHUB_PAT }}
 
 jobs:
   lintr:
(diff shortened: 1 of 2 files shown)


trivialfis (commented):
Could you please explain why is this an issue?  Is opencv  changing some setting in number of threads?

hcho3:
@trivialfis I don't know yet. I filed this PR to document a workaround I happened to find.

trivialfis:
Asking because we are using thread local vector in many places. I'm worrying this might happen in other places.

trivialfis:
But Looks good to me, as long as the tests can pass.

hcho3:
@trivialfis It turns out that it's sufficient to avoid `resize()` of the thread-local vector inside the OpenMP block. (I'm removing the reduction clause because MSVC doesn't support it for the `max` operator.)

We may want to scan the code for all other occurrences of thread-local vectors and see if we are resizing them inside the OpenMP block.

hcho3:
One possibility is that an empty vector could have an optimized representation that differs from the representation of an non-empty vector. So calling `max_columns_vector[tid].resize()` may cause re-allocation of `max_columns_vector` (due to the change of representation) and thus lead to the mutation of the shared object.

trivialfis:
Hmm ... might not be necessary, we don't know the cause, let's don't make too many changes based on guesses.

trivialfis:
> max_columns_vector[tid].resize() may cause re-allocation of max_columns_vector (due to the change of representation)

It can be small vector optimization, but that's thread local right?  It can not change its parent container.

hcho3:
@trivialfis FYI this PR will also fix the R tests on GitHub Actions.

trivialfis (approved):

PR #7188, opened 2021-08-25
[WIP] Add C++ API for external applications

### LINKED ISSUE #4895: [Feature request] Python-like interface for modern C++

hcho3:
Closes #4895

Goals:
* Ease of use
* Closely model the Python frontend in its design
* Ability to support a variety of C++ matrix libraries, such as Armadillo and Eigen.
* Use the C API for implementation.
  - This will de-couple the C++ API layer from XGBoost internals.
  - The C API is well tested with the Python package, so the maintenance of the Python package also ensures that the C++ API would work.
* Sufficiently flexible so that some property testing can be migrated from pytest to gtest (for the sake of speed).

Non-goals:
* API stability (at least not now)

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -1,5 +1,5 @@
 /*!
- * Copyright (c) 2015~2021 by Contributors
+ * Copyright (c) 2015-2021 by Contributors
  * \file c_api.h
  * \author Tianqi Chen
  * \brief C API of XGBoost, used for interfacing to other languages.
--- a/include/xgboost/cxx_api.h
+++ b/include/xgboost/cxx_api.h
@@ -1,0 +1,89 @@
+/*!
+ * Copyright (c) 2015-2021 by Contributors
+ * \file cxx_api.h
+ * \brief C++ API of XGBoost, which is designed to make it easy for other C++ applications to use
+ * XGBoost.
+ * \author Hyunsu Cho
+ */
+#ifndef XGBOOST_CXX_API_H_
+#define XGBOOST_CXX_API_H_
+
+#include <string>
+#include <utility>
+#include <memory>
+#include <limits>
+#include <exception>
+#include <cstdint>
+
+namespace xgboost {
+namespace cxx_api {
+
+class DMatrixInternal;
+class BoosterInternal;
+
+static_assert(std::numeric_limits<float>::has_quiet_NaN,
+              "Must have a non-signaling NaN");
+
+class XGBoostException : public std::exception {
+ public:
+  XGBoostException(const std::string& msg) : msg_(msg) {}
+	const char* what() const noexcept {
+    return msg_.c_str();
+  }
+
+ private:
+  std::string msg_;
+};
+
+class DMatrix {
+ public:
+  DMatrix();
+  DMatrix(const DMatrix&) = delete;
+  DMatrix(DMatrix&&);
+  ~DMatrix();
+  static DMatrix CreateFromMat(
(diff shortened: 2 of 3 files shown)


trivialfis (changes requested):
Thanks for exploring this.  However, I think we don't need to rush into it and should give it more thoughts.

Designing an API is nontrivial, we have been maintaining the Python and R APIs for a while and there are certainly some drawbacks in the original design that we don't want to repeat.  For examples I personally find the `DMatrix` class not very user friendly and should be considered as internal.  Secondly the booster object is too general, we may want to have something more specific like ranker, regressor ...  Also from the perspective of C++ language, we might want a header only package with compatibility with the existing ecosystem along with newer c++ standard.  For instance `std::optional` can be a good friend considering the number of optional parameters that XGBoost has.  I'm just scratching the surface here.  In short, I think if we were to create a C++ package, it should be out of tree for a while until the design stablize.

hcho3:
> I think if we were to create a C++ package, it should be out of tree for a while until the design stablize

Do you mean that I should create a separate GitHub repository? Then we can adopt a later C++ standard and use things like std::optional.

As for the DMatrix, I'm not sure if it can be avoided entirely in the C++ API, since unlike Python, C++ doesn't have a canonical array library like NumPy and CuPy.

Correcting past mistakes is a worthy goal, but it goes against the stated goal of consistency with the Python interface. Do you think this goal should be abandoned? Alternatively, we could shoot for C++ API that looks like the sklearn API.

trivialfis:
> Do you mean that I should create a separate GitHub repository? 

Yes.  But I think we should give it some more thought before starting.

> Then we can adopt a later C++ standard and use things like std::optional.

It's just a suggestion, we can also use simpler things:

``` c++
auto cls = xgboost::Classifier().set_param(tree_method, "gpu_hist").set_param("max_depth", 6)
```

This is also a good way to handle optional parameters.  But `std::optional` seems to be easier for other functions like `predict`.  Or, we can use both where it's appropriate.

> C++ doesn't have a canonical array library like NumPy and CuPy

C++ template is duck type, just like Python. ;-)  Also, we can use optional include like

``` c++
#include <xgboost/cudf_adapter.h>
```

> Do you think this goal should be abandoned?

I don't think that's necessary, Python has sklearn interface and we can be consistent with that.

> MLPack looks kinda like sklearn 

It's a good choice I think.

Actually, I have been thinking about this for a while, but really don't have time to prioritize it.  Maybe we can draft a design doc first?

hcho3:
Sure, let me draft a design doc. Once we can align on goals and features, I can start working on a prototype in a separate GitHub repository.

> C++ template is duck type

The problem remains that different array libraries expose their internals differently. I personally think we should just pick a single array library that is easy to use and also can be installed easily from Conda.

I will address other issues in the design doc. Stay tuned.

trivialfis:
> I personally think we should just pick a single array library that is easy to use and also can be installed easily from Conda.

Let me draft something out this week.

hcho3:
> Let me draft something out this week

I will start a new doc today and share it with you. Then you can add more to it.

trivialfis:
Closing for now.

PR #7199, opened 2021-08-28
Fix auc.cu compilation having no atomicAdd() instance overload.

cbalint13:
This small PR fix compile error of ```auc.cu```:

* The error appears on any cuda == 11.x variants regardless compute arch flags.
* The error appears multiple places in auc.cu when about invoking index sorting:
```
/home/cbalint/rpmbuild/BUILD/xgboost/src/metric/auc.cu(99): here

/usr/local/cuda-11.4/bin/../targets/x86_64-linux/include/cub/device/dispatch/../../
agent/agent_radix_sort_onesweep.cuh(657): error: 
no instance of overloaded function "atomicAdd" matches the argument list
            argument types are: (std::ptrdiff_t *, int)
          detected during:
            instantiation of "cub::AgentRadixSortOnesweep<AgentRadixSortOnesweepPolicy, IS_DESCENDING
```

In basic ```std::ptrdiff_t``` (an elegant intention) is replaced by ```int32_t``` to satisfy cub library.
Another possible solution is a local atomicAdd(pairdiff_t *, int) overload like #7198 .

Cc @trivialfis

### DIFF
--- a/src/common/device_helpers.cuh
+++ b/src/common/device_helpers.cuh
@@ -1371,22 +1371,22 @@ void ArgSort(xgboost::common::Span<U> keys, xgboost::common::Span<IdxT> sorted_i
 
   if (accending) {
     void *d_temp_storage = nullptr;
-    safe_cuda((cub::DispatchRadixSort<false, KeyT, ValueT, std::ptrdiff_t>::Dispatch(
+    safe_cuda((cub::DispatchRadixSort<false, KeyT, ValueT, int32_t>::Dispatch(
         d_temp_storage, bytes, d_keys, d_values, sorted_idx.size(), 0,
         sizeof(KeyT) * 8, false, nullptr, false)));
     TemporaryArray<char> storage(bytes);
     d_temp_storage = storage.data().get();
-    safe_cuda((cub::DispatchRadixSort<false, KeyT, ValueT, std::ptrdiff_t>::Dispatch(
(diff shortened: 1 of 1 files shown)


trivialfis:
By default XGB should be using the cub submodule right?

cbalint13:
> By default XGB should be using the cub submodule right?

 Yes it should, my bad to not mention this.

* Was so used to **not have local cub**, i failed to [mention](https://download.copr.fedorainfracloud.org/results/rezso/ML/fedora-rawhide-x86_64/02349935-xgboost/xgboost.spec) it.
* If PR is still acceptable, maybe is an opportunity to get rid of local CUB ans switch to nvidia's built-in.

Thank you @trivialfis !

trivialfis:
Thanks for the PR, I'm worried about the 32-bit type.  We spent lots of effort debugging integer overflow before, you can see those wrappers in the same file looping over standard thrust functions.  If 32 bit is enough then maybe we can just use the standard interface from cub instead of `cub::xxxDispatch` functions.

trivialfis:
I think @canonizer has insight into whether the integer atomic might overflow when the number of total entries is large than INT32_MAX.

trivialfis:
This might help https://github.com/dmlc/xgboost/pull/7232 .

cbalint13:
> This might help #7232 .

* I confirm that #7232 also fix the issue reported.

I close this very proposal.

Thank you !

PR #7192, opened 2021-08-26
Optimizations for partition, build hist, evaluate, sync kernels

ShvetsKS:
Here is several optimizations of cpu hist method mostly applicable for large datasets like Airline(100M), Higgs(10m), Epsilon(2k X 400k) to increase hw utilization (as for such large data it's better to read bin matrix in not random sequence while building histograms that's why optimized partition was implemented). Summary of this PR:

- sync and evaluate kernels were merged (helped to improve "elavuate+sync" kernels performance for wide histogram (bosch, santander, epsilon) by 2x), added guided scheduling instead of static as each feature has unique number of bins.
- improved partition kernel: allowed to prepare work for thread build hist with **not random** memory access for nodes levels below root, reduced merge indexes time.
- improved build hist kernel with supporting different strategies: reading by rows/columns to keep cache locality, simple columnar sampling support.
- improved objective function calculation (reduced number of blocks and avoided overheads of `operator[]` helped to decrease objective func. computation time almost 5x).
- improved `UpdatePredictionCache` as optimized partition consists position info (node ids) for each training row.

Final results:

|            | airline | epsilon | higgs | bosch (consist missings)  | santander (consist missings) | url (consist missings)| mortgage | airline-ohe | msrank |
|------------|---------|---------|-------|--------|-----------|-------|----------|-------------|--------|
| master, s  | 403.61  | 772.716 | 53.73 | 44.102 | 121.94    | 93.28 | 11.35    | 35.04       | 66.44  |
| this PR, s | 109.49  | 248.44  | 23.77 | 30.3   | 25.16     | 52.3 | 6.9      | 23.58       | 46.14  |
| speedup    | 3.69    | 3.11    | 2.26  | 1.47   | 4.85      | 1.79 | 1.64     | 1.49        | 1.44   |

There is no difference in log loss, rmse, ... metrics for disabled `colsampling_by_node` sampling.
Only small difference in case of enabled 'colsampling_by_node' is just due to changed rng call sequence (that's why changes were introduced into `feature_weights.py` demo).

Of course refactoring is required, but I wanted to start it after a brief overview.

PR #7206, opened 2021-09-01
Reduced span overheads in objective function calculate

ShvetsKS:
it's part of https://github.com/dmlc/xgboost/pull/7192

Before this changes spans were created for each individual sample. Additional overheads of span `operator[]` were also avoided by usage raw pointers.

|       GetGradient     | airline | higgs | 
|------------|---------|---------|
| master, s  | 50.62 | 4.99 | 
| this PR, s | 16.34 | 1.64 | 
| speedup    | 3.1 | 3.04 |

### DIFF
--- a/src/objective/regression_obj.cu
+++ b/src/objective/regression_obj.cu
@@ -73,29 +73,46 @@ class RegLossObj : public ObjFunction {
     additional_input_.HostVector().begin()[1] = scale_pos_weight;
     additional_input_.HostVector().begin()[2] = is_null_weight;
 
-    common::Transform<>::Init([] XGBOOST_DEVICE(size_t _idx,
+    const size_t nthreads = omp_get_max_threads();
+    bool on_device = device >= 0;
+    const size_t n_data_blocks = std::max(static_cast<size_t>(1),
+                                          (on_device ? ndata : nthreads));
+    const size_t block_size = ndata/n_data_blocks + !!(ndata%n_data_blocks);
+    common::Transform<>::Init([block_size, ndata] XGBOOST_DEVICE(
+                           size_t data_block_idx,
                            common::Span<float> _additional_input,
                            common::Span<GradientPair> _out_gpair,
                            common::Span<const bst_float> _preds,
                            common::Span<const bst_float> _labels,
                            common::Span<const bst_float> _weights) {
+          const bst_float* preds_ptr = _preds.data();
+          const bst_float* labels_ptr = _labels.data();
+          const bst_float* weights_ptr = _weights.data();
+          GradientPair* out_gpair_ptr = _out_gpair.data();
+          const size_t begin = data_block_idx*block_size;
+          const size_t end = std::min(ndata, begin + block_size);
           const float _scale_pos_weight = _additional_input[1];
           const bool _is_null_weight = _additional_input[2];
 
-          bst_float p = Loss::PredTransform(_preds[_idx]);
-          bst_float w = _is_null_weight ? 1.0f : _weights[_idx];
-          bst_float label = _labels[_idx];
-          if (label == 1.0f) {
-            w *= _scale_pos_weight;
+          for (size_t idx = begin; idx < end; ++idx) {
+            bst_float p = Loss::PredTransform(preds_ptr[idx]);
+            bst_float w = _is_null_weight ? 1.0f : weights_ptr[idx];
+            bst_float label = labels_ptr[idx];
+            if (label == 1.0f) {
+              w *= _scale_pos_weight;
+            }
+            if (!Loss::CheckLabel(label)) {
+              // If there is an incorrect label, the host code will know.
+              _additional_input[0] = 0;
+            }
+            out_gpair_ptr[idx] = GradientPair(Loss::FirstOrderGradient(p, label) * w,
+                                              Loss::SecondOrderGradient(p, label) * w);
+            if (out_gpair_ptr[idx].GetHess() < 0.0f) {
+              out_gpair_ptr[idx] = GradientPair(0);
+            }
           }
-          if (!Loss::CheckLabel(label)) {
-            // If there is an incorrect label, the host code will know.
-            _additional_input[0] = 0;
-          }
-          _out_gpair[_idx] = GradientPair(Loss::FirstOrderGradient(p, label) * w,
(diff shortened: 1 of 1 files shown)


src/objective/regression_obj.cu
```diff
@@ -73,29 +73,43 @@ class RegLossObj : public ObjFunction {
     additional_input_.HostVector().begin()[1] = scale_pos_weight;
     additional_input_.HostVector().begin()[2] = is_null_weight;
 
-    common::Transform<>::Init([] XGBOOST_DEVICE(size_t _idx,
+    const size_t nthreads = omp_get_max_threads();
+    const size_t n_data_blocks = nthreads;
```
trivialfis, line 77: Can we simply use the raw pointer optimization and remove the thread block?  This function is universal to CPU and GPU. (please don't make a separated implementation ;-) )

src/objective/regression_obj.cu
```diff
@@ -73,29 +73,43 @@ class RegLossObj : public ObjFunction {
...
+            bst_float label = labels_ptr[idx];
+            if (label == 1.0f) {
+              w *= _scale_pos_weight;
+            }
+            if (!Loss::CheckLabel(label)) {
+              // If there is an incorrect label, the host code will know.
+              _additional_input[0] = 0;
+            }
+            out_gpair_ptr[idx] = GradientPair(Loss::FirstOrderGradient(p, label) * w,
+                                            Loss::SecondOrderGradient(p, label) * w);
```
trivialfis, line 106: Indent.

trivialfis (changes requested):
Thanks for spiting up the PR.

src/objective/regression_obj.cu
trivialfis: Or alternatively you can modify the transform function.  But I prefer to keep things simple considering the objective should be a small fraction of overall runtime.

src/objective/regression_obj.cu
ShvetsKS: fixed

src/objective/regression_obj.cu
ShvetsKS: seems no need to modify transform func, added simple support of old behaviour for gpu device (`n_data_blocks == ndata`)

ShvetsKS:
Not sure that https://app.travis-ci.com/github/dmlc/xgboost/jobs/537364519 problem in Travis CI is related to current changes.

trivialfis:
Restarted.

trivialfis:
Hmm ... not sure what's new on Travis.

ShvetsKS:
similar failed test (`tests\python\test_with_sklearn.py::test_estimator_reg`) in master: https://xgboost-ci.net/blue/organizations/jenkins/xgboost-win64/detail/master/948/pipeline, so problem in step `Jenkins Win64: Test` is not related to current changes

trivialfis:
Yeah, handling the CI is the hardest part of maintenance ... it never lacks surprises.

napetrov:
@trivialfis  - should we proceed then with merging this PR?

trivialfis:
Will make some changes before merging.  Thanks for the patience.

ShvetsKS:
> Will make some changes before merging. Thanks for the patience.

@trivialfis seems all checks are passed. Could you take a look at applied comments and help with defining the expected changes?

trivialfis (commented):
Sorry for the slow response.  I can merge it as it's but I have to be honest that I'm quite reluctant to do so.  I generated 1e8 samples with a single column for testing the gradient procedure.  The max_depth is set to 1.  I compared the result between using thread block and using a plain parallel loop, the results are 3.21 sec and 7.59 sec.  While the `UpdateOneIter` takes 117 seconds, and again that's with a single feature.  I'm not sure why would you make this specialization on a general elementwise transform function for 4  seconds.

src/objective/regression_obj.cu
```diff
@@ -73,29 +73,46 @@ class RegLossObj : public ObjFunction {
...
+            if (label == 1.0f) {
+              w *= _scale_pos_weight;
+            }
+            if (!Loss::CheckLabel(label)) {
+              // If there is an incorrect label, the host code will know.
+              _additional_input[0] = 0;
+            }
+            out_gpair_ptr[idx] = GradientPair(Loss::FirstOrderGradient(p, label) * w,
+                                              Loss::SecondOrderGradient(p, label) * w);
+            if (out_gpair_ptr[idx].GetHess() < 0.0f) {
```
trivialfis, line 110: Could you please share which objective is generating negative gradient?

ShvetsKS:
> Sorry for the slow response. I can merge it as it's but I have to be honest that I'm quite reluctant to do so. I generated 1e8 samples with a single column for testing the gradient procedure. The max_depth is set to 1 and I have 12 threads. I compared the result between using thread block and using a plain parallel loop, the results are 3.21 sec and 7.59 sec. While the `UpdateOneIter` takes 117 seconds, and again that's with a single feature and a single node. I'm not sure why would you make this specialization on a general elementwise transform function for 4 seconds.
> 
> Update: I forgot the gradient index generation, which seems to be taking the majority of time.

For airline dataset for example this PR helps to save 34s (with default params of [benchmark](https://github.com/NVIDIA/gbm-bench )). For current/master version it's not big improvement but with optimized partition, hist building, evaluate and sync kernels `GetGradient` function becomes one of the hotspots.

trivialfis:
Opened a small PR on https://github.com/ShvetsKS/xgboost/pull/3 .

src/objective/regression_obj.cu
ShvetsKS: mse, log loss can generate negative gradients (for hessian log-loss e.g.).
But this hessian check was added to avoid similar [here](https://github.com/dmlc/xgboost/blob/master/src/tree/updater_quantile_hist.cc#L483) checks in next PRs (there is no need to read gradient/hessian vector additionally during updater call).

ShvetsKS:
> Opened a small PR on [ShvetsKS#3](https://github.com/ShvetsKS/xgboost/pull/3) .

thanks a lot for corrections!

trivialfis:
No problem, could you please check https://github.com/ShvetsKS/xgboost/pull/3#discussion_r714239072 .

trivialfis (approved):

PR #7208, opened 2021-09-02
Partition optimization

ShvetsKS:
it's part of #7192

Current implementation of partition allow to read bin matrix data with less random access

full train| airline, ~100m | higgs, ~10m
-- | -- | --
master, s | 403 | 53.7
this PR, s | 170s | 32.5
speedup | 2.37 | 1.65


ShvetsKS:
@trivialfis Can we start detailed review of this PR as this part is related to partition optimizations only?
Optimizations of histogram, evaluate and sync kernels are prepared in separate branches

tests/cpp/tree/hist/test_histogram.cc
```diff
@@ -170,54 +170,55 @@ void TestSyncHist(bool is_distributed) {
...
-    const size_t parent_id = tree[node.nid].Parent();
-    const size_t subtraction_node_id =
-        nodes_for_explicit_hist_build_[node_id].nid;
-    auto parent_hist = histogram.Histogram()[parent_id];
-    auto sibling_hist = histogram.Histogram()[subtraction_node_id];
-
-    check_hist(parent_hist, this_hist, sibling_hist, 0, total_bins);
-    ++node_id;
-  }
+  // const common::OptPartitionBuilder* p_opt_partition_builder;
```
trivialfis, line 173: I think this is a sign of working in progress?

ShvetsKS: my mistake, sorry!

ShvetsKS: tests were aligned to optimized partition kernel usage

ShvetsKS:
@trivialfis failed steps `Jenkins Linux: Test` and `Jenkins Linux: Deploy` have similar logs as latest commit in master branch^ https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/release_1.5.0/1/pipeline
seems there is no any failed checks/tests affected by current changes.

trivialfis:
Yup, I'm aware of the failure.  It's blocking the release.

napetrov:
@trivialfis  - are we considering this PR to be integrated prior to release?

trivialfis:
No, I don't think it's a good idea to integrate this amount of changes.  I looked briefly before and I think it can use some cleanups.  I'm on holiday right now.  Will provide detailed review and possible changes once I come back.

napetrov:
@trivialfis  - ok, thanks. Have good  holidays!

ShvetsKS:
similar fails in master branch [here](https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/master/1003/pipeline):
docker: Error response from daemon: OCI runtime create failed
@trivialfis  should I try to restart tests again?

trivialfis:
Restarted.

@hcho3 The issue seems to be lingering.

trivialfis (changes requested):
Hi, I appreciate your excellent optimization.  But I do think there's a way to write more readable code.

The review is not in detail.  A new partitioner that implements `std::partition` spans 550 LOC without any comment and with 29 public variables is a bit too much for me.  Also, the old one seems to be still here for some reason.  I can't merge a PR with this level of complexity without seriously thinking about a redesign/rewrite. There are some good reading materials in c++ core guidelines and are highly recommended:

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#main
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-performance

ShvetsKS (commented):
@trivialfis thanks a lot for review and sorry for long response!
I applied part of comments but related to code complexity, number of input arguments, and reworking in general are still in progress.

src/common/threading_utils.h
```diff
@@ -15,6 +15,11 @@
 namespace xgboost {
 namespace common {
 
+template <typename T>
```
trivialfis, line 18: Excellent!  Now we have a small utility that can be used everywhere

src/common/hist_builder.h
```diff
@@ -0,0 +1,156 @@
...
+                 const uint32_t row_end,
+                 const GHistIndexMatrix& gmat,
+                 const BinIdxType* numa_data,
+                 uint16_t* nodes_ids,
+                 std::vector<std::vector<GradientSumT>>* p_hists,
+                 const uint16_t* mapping_ids) {
+    const size_t nrows = row_end - row_begin;
+    const size_t no_prefetch_size = Prefetch::NoPrefetchSize(nrows);
+
+    if (is_root) {
```
trivialfis, line 129: May I ask what's the difference between the `is_root` can existing `contiguousBlock`?

src/common/hist_builder.h
```diff
@@ -0,0 +1,156 @@
...
+    }
+  }
+
+  uint32_t GetNumBins() const {
+      return nbins_;
+  }
+
+ private:
+  /*! \brief number of threads for parallel computation */
+  size_t nthread_ { 0 };
```
trivialfis, line 149: Is this used anywhere?

src/common/hist_builder.h
```diff
@@ -0,0 +1,156 @@
...
+template<typename FPType, bool do_prefetch,
+         typename BinIdxType, bool is_root,
+         bool any_missing>
+void BuildHistKernel(const std::vector<GradientPair>& gpair,
+                     const uint32_t* rows,
+                     const uint32_t row_begin,
+                     const uint32_t row_end,
+                     const GHistIndexMatrix& gmat,
+                     const BinIdxType* numa_data,
+                     uint16_t* nodes_ids,
```
trivialfis, line 55: Could you please share why do you need to pass nodes_id for building histogram instead of passing the histogram for this particular node?  This way we can have a small scope for this function and makes it easier to test.

Also, I think the node should have 32-bit type `bst_node_t`.

src/common/hist_builder.h
```diff
@@ -0,0 +1,156 @@
...
+                           // to work with gradient pairs as a singe row FP array
+  std::vector<std::vector<FPType>>& hists = *p_hists;
+
+  for (size_t i = row_begin; i < row_end; ++i) {
+    const size_t ri = is_root ? i : rows[i];
+    const size_t icol_start = any_missing ? row_ptr[ri] : ri * n_features;
+    const size_t icol_end =  any_missing ? row_ptr[ri+1] : icol_start + n_features;
+    const size_t row_size = icol_end - icol_start;
+    const size_t idx_gh = two * ri;
+    const uint32_t nid = is_root ? 0 : mapping_ids[nodes_ids[ri]];
```
trivialfis, line 76: It's unclear to me what is this mapping doing.

src/common/hist_builder.h
```diff
@@ -0,0 +1,156 @@
...
+
+  uint32_t GetNumBins() const {
+      return nbins_;
+  }
+
+ private:
+  /*! \brief number of threads for parallel computation */
+  size_t nthread_ { 0 };
+  /*! \brief number of all bins over all features */
+  uint32_t nbins_ { 0 };
```
trivialfis, line 151: This also doesn't seem to be used within the class?

src/common/hist_builder.h
```diff
@@ -0,0 +1,156 @@
...
+  template <typename T>
+  static constexpr size_t GetPrefetchStep() {
+    return Prefetch::kCacheLineSize / sizeof(T);
+  }
+};
+
+template<typename FPType, bool do_prefetch,
+         typename BinIdxType, bool is_root,
+         bool any_missing>
+void BuildHistKernel(const std::vector<GradientPair>& gpair,
```
trivialfis, line 49: Excellent that it's now in a header so we don't have to define those specializations!

trivialfis (commented):
Hi, I took another look, but only at the histogram building kernel this time.  Can we start by working on this function first?  Merging smaller functions makes review and test easier, also has the benefit of decoupling different code sections.  What do you think?

src/common/hist_builder.h
ShvetsKS: Sure.
Changes in `hist_builder.h` were introduced due to changes in partition builder.
Before optimizations "full" bin matrix went through the memory for each node on the same level during histogram building: potentially for each node on specific level we have to read the bin matrix almost from the beginning to the end with random strides. The bandwidth of the memory is significantly dropped in such conditions.
So to mitigate this problem in optimized code we have "sorted" row indecies (but no sorting is done exactly! :) ) for each specific level and as a side effect (rows are not grouped by nodes) we have to know what `node_id` specific row is belong to. `node_id` is required to know what node's histogram should be updated with specific bin matrix row.

Sorry if description is not clear, please let me know and I'll try to add more details.

As for 32-bit type usage, now it works for max_depth <= 16 perfectly. I have to check the performance impact before changing it from 16 to 32 bit and I hope there won't be any reason to add template parameter for it.

src/common/hist_builder.h
ShvetsKS: This is simple mapping of the real node id (in the currently build tree) to node number on current depth (`0 <= nid < 2^current_depth`). This mapping allows to re-use allocated histograms.

src/common/hist_builder.h
ShvetsKS: deleted. Thanks!

src/common/hist_builder.h
ShvetsKS: contiguous block is mostly possible on the root node and it's really rare case when nodes with depth > 0 have contiguous blocks so there is no need to handle such cases specifically.
Also `is_root` parameter handles the reset to zero `node_ids` buffer. As described above we need to track the node ids for each row (rows are not grouped for each node) so we have to reset nod_ids buffer to zero at the start of each tree building

src/objective/regression_obj.cu
```diff
@@ -109,6 +109,9 @@ class RegLossObj : public ObjFunction {
             }
             out_gpair_ptr[idx] = GradientPair(Loss::FirstOrderGradient(p, label) * w,
                                               Loss::SecondOrderGradient(p, label) * w);
+            if (out_gpair_ptr[idx].GetHess() < 0.0f) {
```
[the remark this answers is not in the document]
ShvetsKS: deleted :)

ShvetsKS:
> Hi, I took another look, but only at the histogram building kernel this time. Can we start by working on this function first? Merging smaller functions makes review and test easier, also has the benefit of decoupling different code sections. What do you think?

@trivialfis Thanks a lot for provided review!
I'd love to follow your proposal, but main changes in histogram builder (handling of node ids, mapping, memory allocation etc.) are introduced to support changes in partition builder and not applicable without it. Seems only the movement of build hist kernels  to header file to avoid template specialization can be decoupled. But I'll try to simplify it anyway.

src/common/hist_builder.h
trivialfis: Thank you for the detailed explanation!  That is an insightful optimization.

For future references with these types of optimization, I think you should put that on the front page of PR before anything else and in the code comment, and potentially open an RFC like https://github.com/dmlc/xgboost/issues/7308 before making the change, so we can be clear about the design choice. (I write design doc from time to time just to keep myself clear about the changes).  For this PR, I asked myself again, there's no way I can guess what you are trying to do by looking at the code.

So, let me try to summarize your summary ;-)  Feel free to correct me if I'm wrong.  Instead of asking which row should we process given a node index, you ask which node should we process given a row index.  This optimization is only applicable when the depth-wise grow policy is used, so the code is split and specialized.

Can you design an intuitive interface for the partitioner and the histogram builder so that the code can be easily reused?  I can give it some thoughts if you can confirm my summary.

> As for 32-bit type usage, now it works for max_depth <= 16 perfectly.

I don't think this optimization is necessary.

src/common/opt_partition_builder.h
```diff
@@ -204,6 +204,13 @@ class OptPartitionBuilder {
     if (!all_dense) {
       std::vector<size_t>& local_states = states[tid];
       std::vector<bool>& local_default_flags = default_flags[tid];
+      if (row_indices_begin >= row_indices_end) {
```
trivialfis, line 207: Hey, I saw that you have pushed a fix, presumably found during some local testing.  But really, there's no way I can guess what you are trying to fix.  These few lines look like a hacky workaround to some edge cases.  It's fine, I do that all the time, but can you think of a more general design that doesn't have these conditions?

trivialfis:
I'm happy to schedule a call if needed.  I think we have some different points of view and would be great to share more.

src/common/hist_builder.h
ShvetsKS: > Instead of asking which row should we process given a node index, you ask which node should we process given a row index. This optimization is only applicable when the depth-wise grow policy is used, so the code is split and specialized.

It's clear and right summary :)
Yes. I had to specialize the partition builder related code for depth-wise and loss-guided grow polices, as for loss-guided policy we have to know exactly the set of rows splited to left and to right child nodes.

>Can you design an intuitive interface for the partitioner and the histogram builder so that the code can be easily reused? I can give it some thoughts if you can confirm my summary.

Would it be applicable if the `OptPartitionBuilder` class supported the `depthwise` and `lossguided` policies (strategy pattern) instead of `is_loss_guided` template usage?
As for histogram builder, there is no significant changes in `BuildHistKernel`, only `node_id` usage which is applicable for both growing policies.

src/common/opt_partition_builder.h
ShvetsKS: I changed this fix. Thanks for comment!
Seems, currently it doesn't look like some hacky workaround ;)
And I agree that clear and more general design should be proposed for `OptPartitionBuilder`, possibly with strategy patter applying.

ShvetsKS:
> I'm happy to schedule a call if needed. I think we have some different points of view and would be great to share more.

That's great! We could discuss the delivery plan for these optimizations.
I'm free for a call at any time of 6'th and 7'th November (kirshvets@mail.ru - for direct connection).
Or we can arrange a discussion some time later, after applying comments at least related to:
1) code simplicity, number of class members, input arguments etc.
2) possibly strategy pattern implementation in partition builder

trivialfis:
> I'm free for a call at any time of 6'th and 7'th November (kirshvets@mail.ru - for direct connection).

Thanks, sent an email.

> Or we can arrange a discussion some time later, after applying comments at least related to:

Happy to make the schedule whenever you see fit.  Let's continue the discussion offline. ;-)

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostGeneralSuite.scala
```diff
@@ -115,7 +115,7 @@ class XGBoostGeneralSuite extends FunSuite with TmpFolderPerSuite with PerTest {
     val eval = new EvalError()
     val training = buildDataFrame(Classification.train)
     val testDM = new DMatrix(Classification.test.iterator)
-    val paramMap = Map("eta" -> "1", "gamma" -> "0.5", "max_depth" -> "0",
+    val paramMap = Map("eta" -> "1", "gamma" -> "0.5", "max_depth" -> "6",
```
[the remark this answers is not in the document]
ShvetsKS: reverted back

ShvetsKS:
@trivialfis is there a way to handle this [problem](https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-7208/26/pipeline):
"Public key for epel-release-7-14.noarch.rpm is not installed"

ShvetsKS:
@trivialfis, similar fails were observed [here](https://github.com/dmlc/xgboost/pull/7879#partial-pull-merging)

trivialfis:
Opened a PR: https://github.com/dmlc/xgboost/pull/7884 to see if we can fix it.

trivialfis:
Thank you for continuing the work on optimization. But looking at the code I think my comment will be similar to previous ones. Can we work on smaller chunks of code first and merge the rest incrementally?

ShvetsKS:
> Thank you for continuing the work on optimization. But looking at the code I think my comment will be similar to previous ones. Can we work on smaller chunks of code first and merge the rest incrementally?

All changes in this PR are related to each other: new partition builder required to introduce changes into histogram building kernels, and each part doesn't work separetly.
As common histogram building kernels are used in 'approx' and 'hist' methods, I had to introduce changes in 'approx' builder also. Now new optimized partition builder is applicable for both methods and even common `RowPartitioner` was introduced to avoid existed code duplication :)
Do you think it's better to split changes in 'approx' and 'hist' methods into separate PR's?

Now I see only three not addressed comments related only to 'opt_partition_builder.h' about number of fileds and parameters, sorry for not addressing it yet. I'll try to simplify the code.

razdoburdin:
Hello, everyone.
I have started decomposition of this PR into the smaller parts. The first one is presented here: https://github.com/dmlc/xgboost/pull/7991

PR #7216, opened 2021-09-08
Fix memory leak in C demo.

trivialfis:


trivialfis:
Merging for a one-liner fix.  Also, I have run sanitizer with all the C API demos.

PR #7195, opened 2021-08-27
[jvm-packages][xgboost4j-gpu] Add option to load data from GPU memory

wbo4958:
This PR adds GPU support for XGBoost4j-gpu by leveraging the latest CUDF jar release.

With this PR, Users can do the ETL for the local CSV/Parquet/Orc files by CUDF jar, then build
the GpuDMatrix and do the training.

All of the above stages are accelerating by GPUs.

PR #7217, opened 2021-09-09
Expose DMatrix API for CUDA columnar and array.

trivialfis:
* Use JSON encoded configurations.
* Expose them into header file.

Preparing for https://github.com/dmlc/xgboost/pull/7195 .

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -210,6 +210,36 @@ XGB_DLL int XGDMatrixCreateFromDT(void** data,
                                   DMatrixHandle* out,
                                   int nthread);
 
+/*!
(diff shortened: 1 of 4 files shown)


RAMitchell (commented):
Looks good to me. Only concern is that these are hard to remove once added. Can any api functions be deprecated or removed?

RAMitchell (approved):

trivialfis:
> Can any api functions be deprecated or removed?

These are constructors of DMatrix.  If we need to remove/deprecate some functions, the most likely cause would be that a specific data format is no longer supported by XGBoost.  We are also transitioning from old API functions that accept a fixed parameter set to JSON-based configuration, so the new functions are more extendable.

Issue #7210, opened 2021-09-06
XGBoost4J: model predicts different probabilities depending on number of training samples


rebsp:
### Steps to reproduce
We have trained a binary classifier on a numerical feature vector of size `(3_900_000, 686)`. The training happened with the Python XGBoost library and the sklearn-API (`XGBClassifier`). If we load the model with XGBoost4J in our Java environment, we get different predictions compared to loading the model in Python and predicting with the same input vector. The worst part is that the probabilities are not just a bit off but it is actually flipping from predict 1 to predict 0.

Probability of class 1 with large model in Java:
![image](https://user-images.githubusercontent.com/6863864/132208569-866381c9-a8da-4d42-afea-3ee5af710521.png)
Probability of class 1 with large model in Python:
![image](https://user-images.githubusercontent.com/6863864/132208832-3626188a-d842-4dda-bdea-f62c7ff88dc6.png)

If we repeat the model training with the same parameters but fewer rows (e. g. `(1_000_000, 686)`) the probabilities match between Java and Python.

Probability of class 1 with small model in Java:
![image](https://user-images.githubusercontent.com/6863864/132208907-3b0f23fd-504b-4131-a7f4-63cb75da21ef.png)
Probability of class 1 with small model in Python:
![image](https://user-images.githubusercontent.com/6863864/132208977-a4ef41ca-3987-448d-b068-546f06f5093e.png)

Unfortunately I cannot hand out a sample code example because of IP restrictions, but I could provide the two models through DM's or email for debugging purposes if need be.

For validation purposes, we have converted the model into [ONNX format](https://onnxruntime.ai/docs/reference/api/java-api.html) and tested again. This time the models yield the same results independent of training sample size, thus confirming our assumption that the XGBoost4J library is the culprit.

### Environment
* OS
  tested with Windows 10, Ubuntu 18.04
* Python
  tested with 3.7.10, 3.8.10
* XGBoost
  1.4.1
* Java SDK
  Liberica 11.0.11 (LTS)

Thanks for looking into this, much appreciated!

rebsp:
possibly related to https://github.com/dmlc/xgboost/issues/7103#issuecomment-879587422 ?

> This numerical error in the inference stage is more serious when the data size is pretty large, e.g., m = 10000000.

trivialfis:
Thanks for opening the issue.  Will look into it.

trivialfis:
Could you please share your model with me?  I have my email address in my github profile.

[...]

PR #7225, opened 2021-09-14
[jvm-packages] Deprecate constructors with implicit missing value.

trivialfis:
During debugging https://github.com/dmlc/xgboost/issues/7210 , I found the java implementation has a constructor the uses 0 as the default value for `missing`, which doesn't make sense since 0 is a meaningful input for XGBoost.  `NaN` would be a better choice and is in fact used by both Python and R.  However, I don't think it's a good idea for us to swap the missing value silently.  Therefore, I propose that we deprecate those constructors and let users specify `missing` explicitly.

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/DMatrix.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/DMatrix.java
@@ -119,6 +119,33 @@ public class DMatrix {
 
   /**
    * create DMatrix from dense matrix
+   *
+   * @param data data values
+   * @param nrow number of rows
+   * @param ncol number of columns
+   * @throws XGBoostError native error
+   *
+   * @deprecated Please specify the missing value explicitly
+   */
+  @Deprecated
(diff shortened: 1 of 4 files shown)


trivialfis:
@wbo4958 @CodingCat @hcho3

RAMitchell (commented):
It's a bit annoying to have to specify it when you have dense data. I would argue it should just be NaN like everywhere else, but defer to jvm users.

trivialfis:
Hey @Craigacp , could you please help take a look?

Craigacp:
Looks fine to me. You could modify the javadoc in `DMatrix.java` to say `Please specify the missing value explicitly using {@link DMatrix(float[], int, int, float)` which will insert a link to the correct constructor.

I've only ever used the sparse DMatrix constructors though, so I'm not sure how changing the default will affect users.

trivialfis:
@Craigacp Thanks for the suggestion!

@RAMitchell I'm worried that changing the `missing` value silently can bring some unindented results for users.  Debugging it isn't easy considering users cannot see what's inside the `DMatrix`.

wbo4958:
LGTM,

wbo4958 (approved):

PR #7232, opened 2021-09-15
Enable compiling with system cub.

trivialfis:
~- This works only with CUDA 11.4, for CUDA 11.2 there are some obscure errors with inclusive scan and custom iterators.~

- Tested with all CUDA 11.x.
- Workaround cub scan by using discard iterator in AUC.
- Limit the size of Argsort when compiled with CUDA cub.

### DIFF
--- a/src/metric/auc.cu
+++ b/src/metric/auc.cu
@@ -230,6 +230,14 @@ float ScaleClasses(common::Span<float> results, common::Span<float> local_area,
   return auc_sum;
 }
 
+namespace {
+struct Triple2Pair {
+  auto __device__ operator()(thrust::tuple<uint32_t, float, float> const &t) {
+    return thrust::make_pair(thrust::get<1>(t), thrust::get<2>(t));
+  }
+};
+}  // anonymous namespace
+
 /**
  * MultiClass implementation is similar to binary classification, except we need to split
  * up each class in all kernels.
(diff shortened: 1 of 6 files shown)


trivialfis:
Tested with 11.2 and 11.4 with and without using system cub.

src/metric/auc.cu
```diff
@@ -331,24 +331,25 @@ float GPUMultiClassAUCOVR(common::Span<float const> predts, MetaInfo const &info
...
       thrust::make_counting_iterator(0), [=] __device__(size_t i) {
-        uint32_t class_id = i / n_samples;
-        return thrust::make_tuple(class_id, d_fptp[i].first, d_fptp[i].second);
+        return thrust::make_tuple(i, d_fptp[i].first, d_fptp[i].second);
       });
   // shrink down to pair
   auto fptp_it_out = thrust::make_transform_output_iterator(
-      dh::tbegin(d_fptp), [=] __device__(Triple const &t) {
-        return thrust::make_pair(thrust::get<1>(t), thrust::get<2>(t));
+      dh::TypedDiscard<Triple>{}, [d_fptp] __device__(Triple const &t) {
```
trivialfis, line 338: Originally this reduces the triple to pair, cub in CUDA 11.0~11.2 doesn't handle changed type.  So here I just write the result in lambda and use discard iterator to keep the output type unchanged.

hcho3 (approved):

wbo4958:
@trivialfis Could you cherry-pick this PR to 1.4.0 branch?

trivialfis:
No, I'm planning for a new release

PR #7242, opened 2021-09-17
[WIP] Add typehint to data.py

trivialfis:


trivialfis:
I think it's unnecessary to build a prototype for everything. Will close this now and come back to it once we have full coverage of typehint on tests.

PR #7245, opened 2021-09-18
[R] Avoid symbol naming conflicts with other packages

david-cortes:
I am experiencing an issue with some packages crashing when used in the same session. I'm unable to provide a reproducer, and I don't know exactly what is causing it or which package(s) are at fault, but I suspect it might be due to a symbol naming conflict in the compiled shared objects.

As it is right now, XGBoost registers all its symbols in the R dynamic object when it loads, which can cause issues if multiple packages are loaded and define symbols with the same name. This PR limits the exported symbol visibility to only the necessary - see https://cran.r-project.org/doc/manuals/R-exts.html#Controlling-visibility

Note that it modifies the R file `src/Makevars.in` by adding the R-defined `$(CXX_VISIBILITY)` variable. I don't know how that plays out when using the CMake build system, as I guess it doesn't interact with R in that case, but the R package at least builds fine with `make Rbuild`.

### DIFF
--- a/R-package/src/xgboost-win.def
+++ b/R-package/src/xgboost-win.def
@@ -1,0 +1,3 @@
+LIBRARY xgboost.dll
+EXPORTS
+ R_init_xgboost
--- a/R-package/src/Makevars.in
+++ b/R-package/src/Makevars.in
@@ -17,7 +17,7 @@ endif
 $(foreach v, $(XGB_RFLAGS), $(warning $(v)))
 
 PKG_CPPFLAGS=  -I$(PKGROOT)/include -I$(PKGROOT)/dmlc-core/include -I$(PKGROOT)/rabit/include -I$(PKGROOT) $(XGB_RFLAGS)
-PKG_CXXFLAGS= @OPENMP_CXXFLAGS@ @ENDIAN_FLAG@ -pthread
+PKG_CXXFLAGS= @OPENMP_CXXFLAGS@ @ENDIAN_FLAG@ -pthread $(CXX_VISIBILITY)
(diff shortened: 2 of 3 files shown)


david-cortes:
The failing check is from the python package so unrelated to this PR.

trivialfis (commented):
We have a similar option in CMakeLists.  I think it's a good choice.

But need to investigate whether R_init_xgboost needs decoration since it exposes a C ABI instead of CXX.

trivialfis (approved):
Thanks for the PR.  I haven't looked into the win def yet.  @hcho3 @RAMitchell

david-cortes:
> We have a similar option in CMakeLists. I think it's a good choice.
> 
> But need to investigate whether R_init_xgboost needs decoration since it exposes a C ABI instead of CXX.

Yes, that's a good point, as it doesn't get affected by the `PKG_CXXFLAGS`. It is probably not required and the package works fine without it, but I think it doesn't change anything if it's kept there.

R-package/src/xgboost-win.def
```diff
@@ -0,0 +1,3 @@
+LIBRARY xgboost.dll
+EXPORTS
+ R_init_xgboost
```
hcho3, line 3: I don't think this is necessary, since the `R_init_xgboost` function already has the decoration `__declspec(dllexport)`.

david-cortes: I think that decoration gets ignored when building it through the R toolchain.

hcho3: Good point! Forgot that only Visual Studio recognizes this tag.

PR #7251, opened 2021-09-23
[CI] Use RAPIDS nightly 21.10

hcho3:
Ensure that XGBoost is compatible with the upcoming release of RAPIDS 21.10.


trivialfis:
There's a problem with the latest rmm.  RMM uses a custom version of thrust and brings into include path of XGBoost.  Currently I test the CUDA version, but not the thrust version.

hcho3:
@trivialfis So XGBoost isn't yet compatible with RMM 21.10 🤔

hcho3:
Superseded by #7254

Issue #6659, opened 2021-01-29
Memory usage jumps to 50G when trying to predict


honzasterba:
I have a fairly small booster and data-set, when trying to do prediction on this dataset the memory usage jumps to 50GB.
Here is the code to reproduce:

```
import xgboost as xgb
dtest = xgb.DMatrix("dmatrix.bin")
bst1 = xgb.Booster() 
bst1.load_model('booster.bin')
ypred_h1 = bst1.predict(dtest)
```

Data used to reproduce attached.

[data.zip](https://github.com/dmlc/xgboost/files/5893793/data.zip)

trivialfis:
Just confirming the data loading is correct, your data has 3781180 columns?

honzasterba:
the original training dataset has less than 100 columns, but there are some high cardinality categoricals which due to 1-hot encoding lead to this many columns in the xgboost training set
also should be noted that this is a regression since 1.3.0, with 1.2.0 I did not see this memory spike

trivialfis:
@ShvetsKS Would you like to help taking a look?  I think the thread optimization spikes up the memory usage.  A better way to handle this might be taking some thoughts on extreme sparse dataset.

Right now you can try setting `nthread` to 1 explicitly, or use GPU predictor.

trivialfis:
```
bst.set_param({"nthread": 1})
# or
bst.set_param({"predictor": "gpu_predictor"})
```

honzasterba:
setting nthread to 1 helped to work around the issue

ShvetsKS:
@trivialfis memory usage was increased as currently we [process](https://github.com/dmlc/xgboost/blob/master/src/predictor/cpu_predictor.cc#L189) `kBlockOfRowsSize` observations per each tree to keep cache locality (1 observation was processed before).
I think there are at least tree possible options:
1) add possibility to change `kBlockOfRowsSize` via user provided parameters, and set different from default 64 value
2) implement automatic L1/L2 cache fitting (for current example `kBlockOfRowsSize` would be equal to 1)
3) and as you proposed better way to handle extra sparse datasets with specific prediction implementation

PR #7255, opened 2021-09-24
Avoid thread block with sparse data.

trivialfis:
The 0.5 threshold is just a heuristic not a result of a rigorous benchmark.  For now, I need to close https://github.com/dmlc/xgboost/issues/6659 since sparse datasets are quite usual with one-hot encoding.

@ShvetsKS

### DIFF
--- a/src/predictor/cpu_predictor.cc
+++ b/src/predictor/cpu_predictor.cc
@@ -254,7 +254,8 @@ class CPUPredictor : public Predictor {
                       int32_t tree_end) const {
     const int threads = omp_get_max_threads();
     constexpr double kDensityThresh = .5;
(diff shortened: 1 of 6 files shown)


ShvetsKS:
@trivialfis LGTM in general. But is it possible in future to have the possibility to disable "memory saving mode" for sparse case (is performance is dropped)?

trivialfis:
@ShvetsKS Thanks for the quick response.  In general, I'm open to setting the threshold to smaller values when needed or simply use binary search for extremely sparse datasets.  I think we can find a good threshold for handling those datasets.  Also now that we have a new prediction API that accepts JSON input, you can add a new parameter if needed as the last option.

PR #7267, opened 2021-09-27
[Breaking] Remove `max_delta_step` from Poisson.

trivialfis:
User can use the parameter with same name `max_delta_step` in tree method to control leaf weights, no
need to manipulate hessian values.


trivialfis:
@RAMitchell It might cause problems in hessian as you mentioned due to small values.  But having it might also cause trouble if the hessian is indeed small.  It's a hyper-parameter that needed tuning.

To me, both `max_depth_step` serve the same purpose.  Say if you have a small hessian, the weight of leaf is roughly this:
```
min(G/H, delta)
```
If you have samll hessian, G/H will be large, then delta chunks the output weight.  The `max_delta_step` in objective increases the hessian to reduce G/H.  So they are both trying to reduce the leaf weights.

Issue #7224, opened 2021-09-14
Problem for (de)serialization of Spark implementation of custom objective and eval


nicovdijk:
I have been trying to get (de)serialization of the Spark version of XGBoost in combination with a customObj and customEval working. Serialization works, but loading the model results in the following error:

org.json4s.package$MappingException: No constructor for type ObjectiveTrait, JObject(List())

This error is related to the json4s package. Based on some tests I have run, it seems that we are missing a type hint for the customObj class. I have been trying to add this type hint, but the implementation of the `jsonEncode` method of the `CustomObjParam` class explicitly uses `NoTypeHints`. Comparing the `CustomObjParam` implementation:
```scala
override def jsonEncode(value: ObjectiveTrait): String = {
    import org.json4s.jackson.Serialization
    implicit val formats = Serialization.formats(NoTypeHints)
    compact(render(Extraction.decompose(value)))
}
```
with the general Spark implementation (here of the `IntArrayParam`):
```scala
override def jsonEncode(value: Array[Int]): String = {
    import org.json4s.JsonDSL._
    compact(render(value.toSeq))
}
```
I suspect that changing the line with `NoTypeHints` to something else would allow using type hints. Even better would be to make this work without the need for type hints.

The following test for the Spark version of XGBoost should demonstrate the issue:
```scala
import ml.dmlc.xgboost4j.scala.spark.XGBoostRegressor

import org.json4s._
import org.json4s.jackson.Serialization

import org.apache.spark.{SparkConf, SparkContext}

import some.package.MyCustomObj

class XGBoostTests {
    val sparkConfig = new SparkConf()
        .setAppName("app")
        .setMaster("local")
        .set("spark.ui.enabled", "false")
    val sparkContext = new SparkContext(sparkConfig)

    implicit val formats: Formats = Serialization.formats(ShortTypeHints(List(classOf[MyCustomObj])))

    def testSerialization(): Unit = {
        // XGBoost model
        val xgbRegressor = new XGBoostRegressor()
            .setFeaturesCol("features")
            .setLabelCol("label")
            .setCustomObj(new MyCustomObj)
        // Try saving and loading.
        System.out.println("**** Saving and loading ****");
        xgbRegressor.write.overwrite().save("./temp-xgb-model");
        println("**************** Succesfully saved *****************")
        val xgbRegressor2 = XGBoostRegressor.load("./temp-xgb-model");
        println("**************** Succesfully loaded *****************")
    }

}
```

[...]

PR #7274, opened 2021-09-28
[XGBoost4J-Spark] Problem for (de)serialization of Spark implementation of custom objective and eval

nicovdijk:
Suggested solution to the issue in the title.


trivialfis:
Thanks for the PR!

@wbo4958 Could you please take a look when you are available?

nicovdijk:
Looking forward to getting some feedback.

Also, I'm just realizing that my implementation does not add type hints
when loading is the first thing that happens (now the type hints are added
when instantiating the custom obj/eval for the first time). I guess one
could make sure one always creates an instance before loading, but there
must be a better way. I would appreciate any suggestions. Thanks!

wbo4958:
@nicovdijk What issue does your PR try to solve? Could you add more descriptions?

nicovdijk:
> @nicovdijk What issue does your PR try to solve? Could you add more descriptions?

This PR is a solution to issue #7224 with the same name.

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PersistenceSuite.scala
```diff
@@ -133,6 +133,41 @@ class PersistenceSuite extends FunSuite with TmpFolderPerSuite with PerTest {
...
+  test("test persistence of XGBoostClassifier and XGBoostClassificationModel " +
+    "using custom Obj and Eval") {
+    val eval = new EvalError()
+    val trainingDF = buildDataFrame(Classification.train)
+    val testDM = new DMatrix(Classification.test.iterator)
+
+    val paramMap = Map("eta" -> "0.1", "max_depth" -> "6", "silent" -> "1",
+      "custom_obj" -> new CustomObj, "custom_eval" -> new EvalError,
+      "num_round" -> "10", "num_workers" -> numWorkers)
+    val xgb = new XGBoostClassifier(paramMap)
```
wbo4958, line 145: This "xgb" seems not to be used?

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/CustomObj.scala
```diff
@@ -0,0 +1,99 @@
...
+ limitations under the License.
+ */
+
+package ml.dmlc.xgboost4j.scala.spark
+
+import ml.dmlc.xgboost4j.java.XGBoostError
+import ml.dmlc.xgboost4j.scala.{DMatrix, ObjectiveTrait}
+import ml.dmlc.xgboost4j.scala.spark.params.CustomObjParam._
+import org.apache.commons.logging.LogFactory
+import org.json4s.ShortTypeHints
```
wbo4958, line 23: this is not used. better to remove it

wbo4958 (commented):
@nicovdijk I just figured out what your PR tries to fix. Overall, You PR can fix this issue. But it's not straight forward for users who would like to customize their obj/eval, it requires users explicitly set the type hint.

Could we have callback way to do it by xgboost?

nicovdijk:
> @nicovdijk I just figured out what your PR tries to fix. Overall, You PR can fix this issue. But it's not straight forward for users who would like to customize their obj/eval, it requires users explicitly set the type hint.
> 
> Could we have callback way to do it by xgboost?

I would like to automate it, yes, but I'm not sure what is the best way.
- We could implement adding the type hint in abstract classes SparkCustomObjective and SparkCustomEval, see the new commit. What do you think?

Also:
- I was also wondering how persistence works in xgboost4j. Is it possible there to save and load including a custom objective and eval? I don't see any persistence test, but maybe I'm not looking in the right place...
- The type hints are now added when an instance of a custom objective or eval is created. How can I make sure the hints are added when creating the classes? Right now, saving a model, then restarting the java program and trying to load would not work...

PS The comments have been fixed (I don't seem to have a proper linter).

wbo4958:
XGBoost2MLlibParams https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/GeneralParams.scala#L260 will check and set the param, Could we add some matches for CustomObjParam and CustomEvalParam and set the TypeHints here?

wbo4958:
> > XGBoost2MLlibParams https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/GeneralParams.scala#L260 will check and set the param, Could we add some matches for CustomObjParam and CustomEvalParam and set the TypeHints here?
> 
> That works nicely, thanks for the suggestion.
> 
> The only problem left is that we need to make sure that the type hint gets added, also before creating an instance for the first time (now it gets added when we save first). Any suggestions?
> 
> Maybe one would like to create a dummy instance somewhere when creating the class itself. However, I don't really how this should work for someone using XGBoost..

We could add define an Interface (Trait) for this

nicovdijk:
> We could add define an Interface (Trait) for this

I'm not sure I understand.. Ideally we should register the type hint before instantiation. However, it seems to me that we cannot do that for someone using XGBoost. They would have to do that themselves.

We would like to do the registration when defining the class. In python, for instance, you can run some code when importing a module, but here I'm not sure how to accomplish this. I don't know how a Trait can help me. Can you explain more?

wbo4958:
> would like a general TypeHint adding trait. However, modifying ObjectiveTrait in that way would also force users that do not use Spark to implement that method... Is that a good idea?
> 
> Also, I still don't see how that solves the problem about adding type hints before instantiation of the corresponding class. Sorry if I'm slow..

1. End users customize obj, then they must implement the method adding type hint. So the customized obj has the instance of type hint.
2.  We can add a case to match CustomObjParam and add another method to set the type hint  from the customized obj into the CustomObjParam in XGBoost2MLlibParams


Please correct me.

wbo4958:
Hi @nicovdijk, Personally, this still seems like a workaround instead of a solution.  I will try to make a commit locally to see if my suggestion can work.

nicovdijk:
Ok, I'm still not really sure what you mean, so I would like to see any suggestion you have, thanks!

wbo4958:
@nicovdijk Thx for your feed back. Yeah, I just realize the ObjectiveTrait is only in XGBoost4j.

BTW, Have you tested the scenario just load the model with customized obj/eval without any other train process?

nicovdijk:
> Ok, Could you refine your PR, then we can review it
Sorry, what refinements exactly eould you like?

wbo4958:
@trivialfis Please help to review it, and please change the [XGBoost-spark] to [jvm-packages]

wbo4958:
> > > There is one last thing I would like to add. We use the Spark API in a Java project. Using traits with implementations is a bit cumbersome then. Can we add an abstract class `SparkCustomObjective` or `JavaCustomObjective` (or any other name) to add support for that? It should probably `extends IObjective with TypeHintsTrait`.
> > 
> > 
> > In that case, SparkCustomObjective will break the public API.
> 
> Just to be clear: I would add a Scala abstract class that would be easier to use in Java code, nothing more. On the other hand, I guess there is a work-around by manually using the `SavedTypeHints` object to add `TypeHints`.

You mean xgboost4j introduce the trait from xgboost4j-spark?

nicovdijk:
> You mean xgboost4j introduces the trait from xgboost4j-spark?

No, I only want to add the following to xgboost4j-spark:
```scala
abstract class SparkCustomEval extends EvalTrait with TypeHintsTrait

abstract class SparkCustomObjective extends ObjectiveTrait with TypeHintsTrait
```

Then in Java, you use SparkCustomEval directly (you cannot have multiple inheritance in Java).

wbo4958:
I got the clue from https://stackoverflow.com/questions/61034133/json4s-how-to-deserialize-json-with-fulltypehints-w-o-explicitly-setting-typehi

wbo4958:
@trivialfis Could you help to review it

abugh:
@trivialfis @nicovdijk Hi I'm a user of XGBoost and get caught by this issue for a long time. Just as @wbo4958  said, is it possible for XGBoost to solve it so that XGboost can be integrated as a `stage` of a Spark `PipelineMode` when we are using custom objectives, just as what we can do when using objectives in XGBoost.  For now, we must use XGBoostClassificationModel.load(), or else it will cause errors. It's really inconvenient for the users.

nicovdijk:
@abugh, I believe this merged PR does exactly that. Is this not the case?

wbo4958:
@abugh, Yeah, Please try the newest release

abugh:
@nicovdijk  But I see that in the `jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PersistenceSuite.scala` unit-test, you are still using `XGBoostClassificationModel .save()` and `XGBoostClassificationModel .load()` .

How about if we transfer XGBoostClassificationModel into a stage of Spark PipelineModel, just like this `val pipelineModel = new Pipeline().setStages(Array(XGBoostClassificationModelxxx)).fit(xxx)`? and then use `pipelineModel.save` and `PipelineModel.load` to save and load. Do you mean that the newest xgboost support this now? I didn't see such case in unit-test in this PR. 
Thanks for replying.

abugh:
Now I'm using XGBoost-Spark 1.3.1, with scala 2.1.2 . When I'm using XGBoostRegressionModel with Customized Objectives, I can't even load the model with XGBoostRegressionModel.load() and (I used XGBoostRegressionModel.save() to save the model). 
I thought that the customized objectives are the core module of the XGBoost, but seems that it doesn't work for me even when I'm doing the most standard operations. Is there anything wrong in my operations? Or is there anyway for me to solve this problem except by upgrading the xgboost?
If I must upgrade the xgboost version, which version at least I need to upgrade to?

abugh:
And I see that in the latest version(1.5.1), the unit-test about customized objectives is replaced by using integrated objective.
https://github.com/dmlc/xgboost/blob/release_1.5.1/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/PersistenceSuite.scala

nicovdijk:
@abugh, this PR is not included in the 1.5.0 release or the patch release 1.5.1. It is included in the master branch. I do not know why it's not included yet nor when it will be. 

I have even tested saving and loading in a pipeline in master - it works (which is expected, since it's just using the save and load implementation of the xgboost stage).

abugh:
@nicovdijk Really thanks for your applying. So do you mean that loading in a pipeline is same to loading in XGBoostRegressionModel? Or else, If loading in XGBoostRegressionModel works well,  loading in a pipeline should also work well?

`I have even tested saving and loading in a pipeline in master - it works` Yeah I think it makes sense, as masters branch has already admit this PR, and as you said, this PR is just used to solve this problem. But if I'm using the release Patch(1.5.1 or 1.3.1)m it doesn't work. Any suggestion you can give for me if I can only use release patch for some reasons?

abugh:
Oh! good news. I change to use 1.5.1 release patch, seems it works now when using loading in XGBoostRegressionModel(loading in pipelineMode isn't tested yet, but I guess it works.)

I don't know the reason, as you mentioned that 1.5.1 release patch doesn't admit this PR. But I just share this information to you.

nicovdijk:
> @nicovdijk Really thanks for your applying. So do you mean that loading in a pipeline is same to loading in XGBoostRegressionModel? Or else, If loading in XGBoostRegressionModel works well, loading in a pipeline should also work well?
> 
> `I have even tested saving and loading in a pipeline in master - it works` Yeah I think it makes sense, as masters branch has already admit this PR, and as you said, this PR is just used to solve this problem. But if I'm using the release Patch(1.5.1 or 1.3.1)m it doesn't work. Any suggestion you can give for me if I can only use release patch for some reasons?

@trivialfis, do you know if/when the code in this PR gets released?

abugh:
And also could you tell me which spark version is needed when using xgboost4j-spark_1.51 with Scala 2.12?

abugh:
> Oh! good news. I change to use 1.5.1 release patch, seems it works now when using loading in XGBoostRegressionModel(loading in pipelineMode isn't tested yet, but I guess it works.)
> 
> I don't know the reason, as you mentioned that 1.5.1 release patch doesn't admit this PR. But I just share this information to you.

Unfortunately, seems that the successful run was a coincidence. I can't recurrent it even with the same conditions, and not know the reason.

It would be nice if you can tell me the plan/schedule to release this PR in the release patch. I will wait for it.

Issue #7257, opened 2021-09-24
Is there are some ways to know on which subsample of data the XGBoost tree was fitted?


PowderL:
I am working on data science with xgboost. Recently, I want to diagnose the model with the self-defined method.  And I want to know if there exists any way to get the subset of data sampled from training data for each xgboost tree.

trivialfis:
Unfortunately, the information is not available to the users.

trivialfis:
It's actually straightforward to implement the sampling yourself.  Just define a custom objective and set the hessian value of not selected samples to 0.

PowderL:
> It's actually straightforward to implement the sampling yourself. Just define a custom objective and set the hessian value of not selected samples to 0.

OK, thanks so much, I really appreciate for your reply.

PowderL:
> It's actually straightforward to implement the sampling yourself. Just define a custom objective and set the hessian value of not selected samples to 0.

OK, thanks so much, I really appreciate for your reply. And, is there any way to get the random seed like the  random state in the numpy package to sample the train data for each xgboost tree?

PowderL:
> I am working on data science with xgboost. Recently, I want to diagnose the model with the self-defined method. And I want to know if there exists any way to get the subset of data sampled from training data for each xgboost tree.

PR #7286, opened 2021-10-05
Is there are some ways to know which subsample of data used by each XGBoost tree for fitting?

PowderL:
As I knew, in the training process of XGBoost, Boostrap that randomly samples a subset of training data with or without replacement for every single tree is executed. Is there any way to get the subsample of training data for each tree? Or, via obtaining the random numbers for sampling even the seed of the series of random numbers, we can get the subset of training data. I have asked the questions a week ago, I have asked the questions a week ago and got extraordinary suggestions [(7257).](https://github.com/dmlc/xgboost/issues/7257).  But it seems that the present version of XGBoost seems to not support this. Is it possible the subsequent version of XGBoost supporting this?


trivialfis:
Please don't open a PR for questions.  It consumes CI resources.  I answer questions in [issues](https://github.com/dmlc/xgboost/issues) and @hcho3 help maintain the [form](https://discuss.xgboost.ai/)

Currently, there's no plan to expose the subsampling process to the user interface.  It's implemented as an internal procedure in tree methods.  Also, CPU and GPU use different random engines.  To expose such an interface, XGBoost has to accept user-defined functions, possibly in a different language as callbacks, which is not trivial nor efficient.  As mentioned in the issue, you can implement the subsample yourself easily.

PR #7297, opened 2021-10-08
Re-implement PR-AUC.

trivialfis:
The new implementation supports binary/multi-class classification and learning to rank to binary relevance.  Also, it handles empty datasets and irregular datasets by returning NaN instead of making an exception.  Lastly, the GPU implementation has the same functionality as the CPU one instead of being ranking only.

* Perf

n_samples = 1e7
runs = 16
n_classes = 8 (multi-class only)

|                    | Master             | AUCPR              |
|:-------------------|:-------------------|:-------------------|
| CPU ROC-AUC Binary | 8.385259628295898  | 8.928396701812744  |
| CPU PR-AUC Binary  | 9.372312784194946  | 9.454226732254028  |
| GPU ROC-AUC Binary | 0.7207620143890381 | 0.711883544921875  |
| GPU PR-AUC Binary  | NA                 | 0.8523569107055664 |
| CPU ROC-AUC Multi  | 64.72866940498352  | 66.78449654579163  |
| CPU PR-AUC Multi   | NA                 | 67.58437657356262  |
| GPU ROC-AUC Multi  | 8.162241697311401  | 8.218487024307251  |
| GPU PR-AUC Multi   | NA                 | 8.903043985366821  |

* Related

Close https://github.com/dmlc/xgboost/issues/6561 .
Close https://github.com/dmlc/xgboost/issues/6272 .
Close https://github.com/dmlc/xgboost/issues/6551 .
Close https://github.com/dmlc/xgboost/issues/6692 .

PR #7300, opened 2021-10-10
Fix weighted samples in multi-class AUC.

trivialfis:


trivialfis:
The fix is extracted from https://github.com/dmlc/xgboost/pull/7297 for backporting.

Issue #7287, opened 2021-10-05
Lack of reproducibility despite using a consistent seed


deanforce:
Hello,
We’ve been having some discussion concerning a lack of reproducibility despite using a consistent seed in xgboost when employing Bayesian optimization for the hyperparameters. After some digging, the issue seems to be in the xgboost package itself, but I’m still unclear as to what’s driving this. Could you please take a look? a sample syntax to reproduce the problem is available here: https://github.com/AnotherSamWilson/ParBayesianOptimization/issues/43 2

Thank you!

trivialfis:
I think I got the result now.  It's caused by non-deterministic openmp reduction  Same cause as https://github.com/dmlc/xgboost/issues/5019 .

Issue #5019, opened 2019-11-06
Inconsistent evaluation metric for tweedie


guiguilegui:
Hello everyone,

Using the tweedie distribution doesn't give consistent evaluation metric results (nloglikelihood), even when a seed is used.

Example when used on R 3.6.1 with xgboost_0.90.0.2:
```
df <- data.frame(
	y = c(1000, 0, 0, 0, 0, 1000),
	x1 = c('A', 'B', 'C', 'A', 'B', 'C')
)

mm <- model.matrix(~ x1, data = df)[,-1]
dtt <- xgb.DMatrix(mm, label = df$y)

set.seed(9370707)
xgb <- 
	sapply(
		1:100,
			function(x) {
			set.seed(9370707)
			xgb.train(
				data = dtt, 
				objective = 'reg:tweedie',
				nrounds = 1, 
				watchlist = list(test = dtt),
				seed = 9370707
			)$evaluation_log$test_tweedie_nloglik[1]
			}
	)

table(xgb)
#xgb
#701.132507 701.132568 
#        62         38
```

PR #7303, opened 2021-10-10
Deterministic result for element-wise/mclass metrics.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/7287 .
Close https://github.com/dmlc/xgboost/issues/5019 .


trivialfis:
The CPU AUC also uses openmp reduction.  I will keep it out of scope for this PR for now and do a complete scan of its usage.

Issue #2087, opened 2017-03-08
Multiple output regression


miguelmartin75:
How do I perform multiple output regression? Or is it simply not possible?

My current assumption is that I would have to modify the code-base such that XGMatrix supports a matrix as labels and that I would have to create a custom objective function.

My end goal would be to perform regression to output two variables (a point) and to optimise euclidean loss. Would I be better off to make two seperate models (one for x coordinates and one for y coordinates).

Or... would I be better off using a random forest regressor within sklearn or some other alternative algorithm?

khotilov:
Multivariate/multilabel regression is not currently implemented #574 #680
Tianqi had added some relevant placeholder data structures to gbtree learner, but no one had time, I guess, to work the machinery out.

jindongwang:
Pity, since many competitions are with multi-outputs

MarkusBonsch:
This would be a really nice feature to have.

joel-thomas-wilson:
Do we have any updates on this?

JacobKempster:
I agree - this feature would be extremely valuable (exactly what I need right now...)

cp9612:
Would like to see this feature coming

veonua:
any reason why it is closed?

hcho3:
@veonua See #3439.

loretoparisi:
In the meanwhile there is any alternative, like any ensemble of single output models?

jimmywan:
> In the meanwhile there is any alternative, like any ensemble of single output models like:

https://scikit-learn.org/stable/modules/generated/sklearn.multioutput.MultiOutputRegressor.html

cmottet:
I am going to also weigh and say that having such feature would be extremely handy. The MultiOutputRegressor mentioned above is a nice wrapper to build multiple models at once and it does work well for predicting target variables that are independent from one another. However, if the target variables are highly correlated, then you really want to build one model that predicts a vector.

MxNl:
A year has passed soon since the last comment :-). This is why I want to repeat the wish to have such an interesting feature. I would be happy to see this. Thanks anyway for all your work.

hcho3:
Reopening for visibility.

kk26269:
> Multivariate/multilabel regression is not currently implemented #574 #680
> Tianqi had added some relevant placeholder data structures to gbtree learner, but no one had time, I guess, to work the machinery out.

[...]

PR #7309, opened 2021-10-11
Initial support for multi-output regression.

trivialfis:
* Implemented as 1 model per target.
* Not available for classification yet.

I will follow up with R and dask implementations if this PR is approved.

Related: https://github.com/dmlc/xgboost/issues/2087 .

Todos:
- [x] Verify data types other than numpy array.
- [ ] Have better specification for return shape.

### DIFF
--- a/demo/guide-python/multioutput_regression.py
+++ b/demo/guide-python/multioutput_regression.py
@@ -1,0 +1,45 @@
+"""A demo adopted from scikit-learn for multi-output regression.
+
+https://scikit-learn.org/stable/auto_examples/ensemble/plot_random_forest_regression_multioutput.html#sphx-glr-auto-examples-ensemble-plot-random-forest-regression-multioutput-py
+
+"""
+import numpy as np
+import xgboost as xgb
+import argparse
+from matplotlib import pyplot as plt
+
+
+def plot_predt(y, y_predt, name):
+    s = 25
+    plt.scatter(y[:, 0], y[:, 1], c="navy", s=s,
+                edgecolor="black", label="data")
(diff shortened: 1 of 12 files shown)


Craigacp:
That's exciting. We'd be interested in a XGBoost4J wrapper too, and I might be able to write such a thing once you're happy with the native API. We currently fake this up by training a separate Booster for each output dimension.

trivialfis:
TO-DO for initial support of multi-target model training.  

- [ ] A typed tensor view that can be used in GPU kernels: https://github.com/dmlc/xgboost/pull/7365
- [ ] Typed tensor storage.
- [ ] Type erased tensor for accepting user inputs.  This is available as the array interface class, we just need to extend it to a higher dimension.
- [ ] Convert labels and base margin into matrix/tensor with proper shape attributes.
- [ ] Define the tree index for target, similar to what we currently have as group idx.

trivialfis:
Most of the basic infrastructures are in place now.  Need more thought on classification problem due to different number of classes per target.

trivialfis:
The initial support is merged in https://github.com/dmlc/xgboost/pull/7514 .

trivialfis:
@Craigacp I think for multi-target classification models, xgboost needs a new interface for and potentially lots of refactoring.  I will focus on regression for now.  Thank you for joining the discussion and feel free to test the new feature. ;-)

aniruddhghatpande:
@trivialfis  Is this contain support for all multi output regression ? @Craigacp were you able to write the wrapper for XGBoost4J ?

Craigacp:
I haven't looked into it. I've had less time to work on XGBoost related things for the past couple of years.

PR #7359, opened 2021-10-25
Update spark dependency to spark 3.2

WeichenXu123:
Current xgboost4j released version built with spark 3.0, but the binary jar is incompatible with spark 3.2
This PR update the spark dependency to be spark 3.2


wbo4958:
@trivialfis, looks like we can close this PR

trivialfis:
Thanks you both!

PR #7365, opened 2021-10-26
Implement a general array view.

trivialfis:
* Replace existing matrix and vector view.

Makes https://github.com/dmlc/xgboost/pull/7309 easier.  Once we go to multi-target multi-class, predict leaf will have 5-dim output.

### DIFF
--- a/src/gbm/gblinear.cc
+++ b/src/gbm/gblinear.cc
@@ -243,7 +243,10 @@ class GBLinear : public GradientBooster {
     // The bias is the last weight
     out_scores->resize(model_.weight.size() - learner_model_param_->num_output_group, 0);
     auto n_groups = learner_model_param_->num_output_group;
-    MatrixView<float> scores{out_scores, {learner_model_param_->num_feature, n_groups}};
+    linalg::TensorView<float, 2> scores{
+        *out_scores,
+        {learner_model_param_->num_feature, n_groups},
+        GenericParameter::kCpuId};
     for (size_t i = 0; i < learner_model_param_->num_feature; ++i) {
       for (bst_group_t g = 0; g < n_groups; ++g) {
         scores(i, g) = model_[i][g];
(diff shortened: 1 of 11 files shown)


src/gbm/gblinear.cc
```diff
@@ -243,10 +243,15 @@ class GBLinear : public GradientBooster {
...
     out_scores->resize(model_.weight.size() - learner_model_param_->num_output_group, 0);
     auto n_groups = learner_model_param_->num_output_group;
-    MatrixView<float> scores{out_scores, {learner_model_param_->num_feature, n_groups}};
+    linalg::TensorView<float, 2> scores{
+        *out_scores,
+        {learner_model_param_->num_feature, n_groups},
+        GenericParameter::kCpuId};
     for (size_t i = 0; i < learner_model_param_->num_feature; ++i) {
+      auto feature_score = scores.Slice(i, linalg::All());
+      CHECK_EQ(feature_score.Shape().size(), 1);
```
hcho3, line 252: Is this check worth the overhead in a hot loop?

tests/cpp/common/test_linalg.cc
```diff
@@ -28,11 +29,85 @@ TEST(Linalg, Vector) {
   size_t kRows = 31, kCols = 77;
   HostDeviceVector<float> storage;
   auto m = MakeMatrixFromTest(&storage, kRows, kCols);
-  auto v = VectorView<float>(m, 3);
+  // auto v = VectorView<float>(m, 3);
```
hcho3, line 32:
```suggestion
```

src/gbm/gblinear.cc
trivialfis: Let me remove it.

trivialfis:
Added a bit more code comments and an optimization that indexing uses only pointer instead of span.

trivialfis:
@hcho3 Could you please take another look when you have time?

hcho3 (approved):

PR #7214, opened 2021-09-08
Rewrite approx

trivialfis:
Based on https://github.com/dmlc/xgboost/pull/7183

This PR rewrites the approx tree method to use codebase from hist for better performance and prepare for categorical data support.

The rewrite has many benefits:
- Support for both `max_leaves` and `max_depth`.
- Support for `grow_policy`.
- Support for mono constraint.
- Support for feature weights.
- Support for easier bin configuration (`max_bin`).
- Support for categorical data.
- Faster performance for most of the datasets. (many times faster)
- Support for prediction cache.
- Significantly better performance for external memory.
- Unites the code base between approx and hist.

TODOs:
- [x] Figure out how to handle `MatchThreadsToNodes`.
- [x] Handle different sizes of batches.
- [x] Handle empty partition gracefully.
- [x] Add tests.
- [x] Distributed training.
- [x] Tests for build policies and related parameters (max_leave, max_depth).
- [x] Fix task initialization.

Close https://github.com/dmlc/xgboost/issues/7244 .

PR #7372, opened 2021-10-27
Support external memory in CPU histogram building.

trivialfis:
Extracted from https://github.com/dmlc/xgboost/pull/7214

The perf change for the complete PR is in https://github.com/dmlc/xgboost/pull/7214#issuecomment-952971039 .  Might have some small overhead (within variance), but the benefit is hist/approx will be able to support both categorical data and external memory in the future.

* Support external memory in histogram building.
* Remove unused `SubtractionTrick` function.  The builder class can be completely removed, but I will keep the cleanup out of the scope.

### DIFF
--- a/src/common/hist_util.cc
+++ b/src/common/hist_util.cc
@@ -133,74 +133,108 @@ struct Prefetch {
 
 constexpr size_t Prefetch::kNoPrefetchSize;
 
-
-template<typename FPType, bool do_prefetch, typename BinIdxType, bool any_missing = true>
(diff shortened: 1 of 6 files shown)


trivialfis:
@ShvetsKS Could you please take a look into this PR?  We can set `hist` to default once we have external memory support (it will be feature complete against all the other tree methods).

trivialfis:
@ShvetsKS  friendly ping.

trivialfis:
I will merge this PR next week if there's no objection.

Issue #7083, opened 2021-07-06
Base margin support for multi-class classifier


trivialfis:
A tracking issue for the feature.

PR #7381, opened 2021-10-29
Support multi-class with base margin.

trivialfis:
* cudf
* cupy
* numpy
* pandas
* modin
* dask

This is already partially supported but never properly tested.  So the only possible way to use it is calling `numpy.array.flatten` with `base_margin` before passing it into XGBoost.  This PR adds proper support for most of the data types along with tests.

Close https://github.com/dmlc/xgboost/issues/7083 .


trivialfis:
Interesting that I couldn't reproduce the error in my setup.  Might be an issue with CUDA 10

hcho3 (approved):

PR #7385, opened 2021-10-31
Pass infomation about objective to tree methods.

trivialfis:
This PR adds a new struct for obtaining some information about the objective. The struct is not used yet but is required for the future development of categorical split.  In https://github.com/dmlc/xgboost/pull/7214 , the optimal partitioning is only available for binary classification and regression.  Copying inlined comment here:
```
 * The task field is useful for tree split finding, also for some metrics like auc.
 * While const_hess is useful for algorithms like adaptive tree where one needs to update
 * the leaf value after building the tree.  Lastly, knowing whether hessian is constant
 * can allow some optimizations like skipping the quantile sketching.
```

* Define `ObjInfo`.
* Pass it to tree methods.

### DIFF
--- a/tests/cpp/tree/test_gpu_hist.cu
+++ b/tests/cpp/tree/test_gpu_hist.cu
@@ -275,7 +275,8 @@ void TestHistogramIndexImpl() {
   int constexpr kNRows = 1000, kNCols = 10;
 
   // Build 2 matrices and build a histogram maker with that
-  tree::GPUHistMakerSpecialised<GradientPairPrecise> hist_maker, hist_maker_ext;
+  tree::GPUHistMakerSpecialised<GradientPairPrecise> hist_maker{{ObjInfo::kRegression, true}},
+      hist_maker_ext{{ObjInfo::kRegression, true}};
   std::unique_ptr<DMatrix> hist_maker_dmat(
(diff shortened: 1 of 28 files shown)


trivialfis:
I will pass it to metric once https://github.com/dmlc/xgboost/pull/7364 is approved or rejected to avoid conflicts.

tests/cpp/tree/test_gpu_hist.cu
```diff
@@ -275,7 +275,8 @@ void TestHistogramIndexImpl() {
   int constexpr kNRows = 1000, kNCols = 10;
 
   // Build 2 matrices and build a histogram maker with that
-  tree::GPUHistMakerSpecialised<GradientPairPrecise> hist_maker, hist_maker_ext;
+  tree::GPUHistMakerSpecialised<GradientPairPrecise> hist_maker{{ObjInfo::kRegression, true}},
+      hist_maker_ext{{ObjInfo::kRegression, true}};
```
hcho3, line 279: Should all C++ tests use an objective with identical Hessians?

trivialfis: @hcho3 Let me set it to false by default since that's what most of the objectives have.

hcho3 (approved):

PR #7390, opened 2021-11-02
Handle `OMP_THREAD_LIMIT` in environment.

trivialfis:
Fixes R win builder.

cc @hetong007 

Details, `OMP_THREAD_LIMIT` is an environment variable defined by OpenMP standard.  Users can set it to limit the number of threads in each thread group.  However, the `omp_get_max_threads()` function used by XGBoost is not affected by this env var.  So when the env var is set, XGBoost naively thought it can obtain `omp_get_max_threads()` number of threads, while the received number of threads is much smaller.

I ran it on win builder for a quick test https://win-builder.r-project.org/vCrjkrlXZE0l/00check.log we should run it again after backporting to 1.5 branch.

### DIFF
--- a/R-package/R/xgboost.R
+++ b/R-package/R/xgboost.R
@@ -9,8 +9,8 @@ xgboost <- function(data = NULL, label = NULL, missing = NA, weight = NULL,
                     early_stopping_rounds = NULL, maximize = NULL,
                     save_period = NULL, save_name = "xgboost.model",
                     xgb_model = NULL, callbacks = list(), ...) {
-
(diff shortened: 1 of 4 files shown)


trivialfis:
Also, the cause of https://github.com/dmlc/xgboost/issues/7156 is now known.  This would be proper fix.

trivialfis:
cc @hcho3 I feel like I just did the impossible.  Please review. ;-)

hcho3 (approved):

hcho3:
Nicely done!

trivialfis:
Hmm, msvc doesn't implement this environment variable.

PR #7393, opened 2021-11-03
CPU evaluation for cat data.

trivialfis:
* Implementation for one hot based.
* Implementation for partition based for regression. (LightGBM)

Currently, the evaluation function is using parameters from numerical split.  We should set up a new set of training parameters later.  But before that, I want to have a working implementation and some experiment results first.

Extracted from https://github.com/dmlc/xgboost/pull/7214 .

### DIFF
--- a/src/tree/hist/evaluate_splits.h
+++ b/src/tree/hist/evaluate_splits.h
@@ -6,13 +6,16 @@
 
 #include <algorithm>
 #include <memory>
+#include <numeric>
 #include <limits>
 #include <utility>
 #include <vector>
 
+#include "xgboost/task.h"
 #include "../param.h"
 #include "../constraints.h"
 #include "../split_evaluator.h"
+#include "../../common/categorical.h"
 #include "../../common/random.h"
 #include "../../common/hist_util.h"
 #include "../../data/gradient_index.h"
@@ -36,13 +39,13 @@ template <typename GradientSumT, typename ExpandEntry> class HistEvaluator {
   int32_t n_threads_ {0};
   FeatureInteractionConstraintHost interaction_constraints_;
   std::vector<NodeEntry> snode_;
+  ObjInfo task_;
 
   // if sum of statistics for non-missing values in the node
   // is equal to sum of statistics for all values:
   // then - there are no missing values
   // else - there are missing values
-  bool static SplitContainsMissingValues(const GradStats e,
-                                         const NodeEntry &snode) {
+  bool static SplitContainsMissingValues(const GradStats e, const NodeEntry &snode) {
     if (e.GetGrad() == snode.stats.GetGrad() &&
         e.GetHess() == snode.stats.GetHess()) {
       return false;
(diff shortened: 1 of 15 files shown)


src/tree/hist/evaluate_splits.h
```diff
@@ -91,49 +96,118 @@ template <typename GradientSumT, typename ExpandEntry> class HistEvaluator {
       iend = static_cast<int32_t>(cut_ptr[fidx]) - 1;
     }
 
+    auto calc_bin_value = [&](auto i) {
```
trivialfis, line 99: This can be split into multiple functions, but then we have have lots of duplicated code.

src/tree/hist/evaluate_splits.h
```diff
@@ -167,12 +241,37 @@ template <typename GradientSumT, typename ExpandEntry> class HistEvaluator {
...
+          auto n_bins = cut.Ptrs().at(fidx + 1) - cut.Ptrs()[fidx];
+          if (common::UseOneHot(n_bins, param_.max_cat_to_onehot, task_)) {
+            EnumerateSplit<+1, kOneHot>(cut, {}, histogram, fidx, nidx, evaluator, best);
+            EnumerateSplit<-1, kOneHot>(cut, {}, histogram, fidx, nidx, evaluator, best);
+          } else {
+            auto const &cut_ptr = cut.Ptrs();
+            std::vector<size_t> sorted_idx(n_bins);
+            std::iota(sorted_idx.begin(), sorted_idx.end(), 0);
+            auto feat_hist = histogram.subspan(cut_ptr[fidx], n_bins);
+            std::stable_sort(sorted_idx.begin(), sorted_idx.end(), [&](size_t l, size_t r) {
```
trivialfis, line 258: Can't use argsort as we don't have a cpu transform iter.

tests/cpp/tree/hist/test_evaluate_splits.cc
```diff
@@ -39,7 +36,7 @@ template <typename GradientSumT> void TestEvaluateSplits() {
   std::iota(row_indices.begin(), row_indices.end(), 0);
   row_set_collection.Init();
 
-  auto hist_builder = GHistBuilder<GradientSumT>(n_threads, gmat.cut.Ptrs().back());
+  auto hist_builder = GHistBuilder<GradientSumT>(omp_get_max_threads(), gmat.cut.Ptrs().back());
```
hcho3, line 39: Shouldn't we call function `OmpGetThreadLimit()` to query `OMP_THREAD_LIMIT`?

tests/cpp/tree/hist/test_evaluate_splits.cc
```diff
@@ -108,5 +105,142 @@ TEST(HistEvaluator, Apply) {
...
+  using GradientSumT = double;
+  std::vector<FeatureType> ft(kCols, FeatureType::kCategorical);
+
+  TrainParam param;
+  param.UpdateAllowUnknown(Args{{"min_child_weight", "0"}, {"reg_lambda", "0"}});
+
+  size_t n_cats{8};
+
+  auto dmat =
+      RandomDataGenerator(kRows, kCols, 0).Seed(3).Type(ft).MaxCategory(n_cats).GenerateDMatrix();
```
hcho3, line 120: I really like how we can generate random data with a [fluent interface](https://en.wikipedia.org/wiki/Fluent_interface).

hcho3 (approved):
LGTM

tests/cpp/tree/hist/test_evaluate_splits.cc
trivialfis: It's a c++ unittest, I think we don't have to worry about that too much.  We on the other hand, we need for thorough  integration tests for that env, probably in another PR.

PR #7395, opened 2021-11-04
Add note about CRAN release [skip ci]

trivialfis:


trivialfis:
Hi @hetong007 , could you please take a look and fill in the blanks when you are available?  Next time I can do more thorough checks before tagging a release. ;-)

hetong007:
Thanks for the notes. I think r-hub and win-builder are sufficient and they have covered almost all submission tests in the previous releases. This time, the reverse-dependency is kinda tricky and infeasible to check on our side (at least I don't know a convinient way) for every release, unless it has significant breaking changes.

trivialfis:
Excellent!  I will continue working on the reverse dependecies.

PR #7398, opened 2021-11-05
Update dmlc-core and use data iter for GPU sampling tests.

trivialfis:
* Add parquet build.
* Fix unity build.

Close https://github.com/dmlc/xgboost/pull/7139


hcho3 (approved):

trivialfis:
Pushed a commit to use the new external memory interface in the gradient sampling test.  It bypasses dmlc threaded iter.

hcho3:
Should this change part of the PR that updates dmlc-core?

trivialfis:
Good point.  I will split it up since more tests need the conversion.

trivialfis:
I still need the updated dmlc-core to confirm that it's working correctly ...

hcho3 (approved):
LGTM. Please rename this PR to indicate that the external memory test was revised to use the data iterator.

PR #7399, opened 2021-11-06
Expose build info

trivialfis:
To-do:
- [x] C API.
- [x] Test.


trivialfis:
This PR adds a function to capture some of the build flags used in XGBoost and expose them as runtime information.  This way the users can see the capability of binary they are using.

hcho3 (approved):

PR #7412, opened 2021-11-10
Bump R cache version on GH action.

trivialfis:


trivialfis:
Unrelated,  there's a new dep from igraph.

PR #7453, opened 2021-11-19
Add range-based slicing to tensor view.

trivialfis:
* Decouple the implementation with xgboost to make it reusable.
* Add range-based slicing to tensor view.
* Calculate CContiguous explicitly.
* Remove span-based iterators.


RAMitchell (approved):
There is still some use of span here right? So this implementation is still coupled with xgb.

trivialfis:
Yes, but the span class by itself is quite reusable.

Issue #7454, opened 2021-11-19
`AttributeError` with fitting model on Dask Array backed by `scipy.sparse.csr_matrix`


jrbourbeau:
I came across a use case where attempting to fit a `DaskXGBClassifier` on a Dask Array whose partitions are `scipy.sparse.csr_matrix`s (as is returned by Dask-ML's `HashingVectorizer`) results in a `AttributeError: divisions not found` error (full traceback included below).

From doing some initial debugging it appears the underlying issue is that during the fitting process we end up passing a `list` of sparse matrices to Dask's `dd.multi.concat` here

https://github.com/dmlc/xgboost/blob/d33854af1b4f783c5230bb21aff7234b16f409f7/python-package/xgboost/dask.py#L207

However, `dd.multi.concat` expects a `list` of Dask DataFrames, which is where the `AttributeError: divisions not found` is coming from (Dask DataFrames have a `.divisions` attribute which `dd.multi.concat` assumes exists).

Here's an example code snippet which should reproduce the issue when using the latest `xgboost` (1.5.0) and `dask` (2021.11.2) / `distributed` (2021.11.2) releases:

```python
import dask.dataframe as dd
import dask_ml.feature_extraction.text
import pandas as pd
import sklearn.datasets
from dask.distributed import Client
from xgboost.dask import DaskXGBClassifier

if __name__ == "__main__":

    with Client():
        # Create Dask DataFrame from sklearn 20newsgroups dataset
        bunch = sklearn.datasets.fetch_20newsgroups()
        df = dd.from_pandas(
            pd.DataFrame({"text": bunch.data, "target": bunch.target}), npartitions=25
        )

        # Create features with dask-ml's `HashingVectorizer``
        vect = dask_ml.feature_extraction.text.HashingVectorizer()
        X = vect.fit_transform(df["text"])

        # Format classification labels
        y = df["target"].to_dask_array()

        # Train XGBoost classifier
        clf = DaskXGBClassifier()
        print(f"{X = }")
        print(f"{y = }")
        clf.fit(X, y)  # Results in `AttributeError: divisions not found`

```

<details>
<summary>Full traceback:</summary>

```
Traceback (most recent call last):
  File "/Users/james/projects/coiled/evangelism-private/mongodb-with-coiled/test.py", line 28, in <module>
    clf.fit(X, y)
  File "/Users/james/mambaforge/envs/dask-nlp/lib/python3.9/site-packages/xgboost/dask.py", line 1817, in fit
[...]

PR #7457, opened 2021-11-21
Support scipy sparse in dask.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/7454 .

pydata sparse is not yet supported.  Would be great if we can add a new adapter in c++.


hcho3 (approved):

jrbourbeau:
Thanks @trivialfis!

PR #7467, opened 2021-11-22
Add approx partitioner.

trivialfis:
Extracted from https://github.com/dmlc/xgboost/pull/7214 .

### DIFF
--- a/src/tree/updater_approx.h
+++ b/src/tree/updater_approx.h
@@ -1,0 +1,146 @@
+/*!
+ * Copyright 2021 XGBoost contributors
+ *
+ * \brief Implementation for the approx tree method.
+ */
+#ifndef XGBOOST_TREE_UPDATER_APPROX_H_
+#define XGBOOST_TREE_UPDATER_APPROX_H_
+
+#include <limits>
+#include <utility>
+#include <vector>
+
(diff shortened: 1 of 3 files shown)


src/tree/updater_approx.h
```diff
@@ -0,0 +1,146 @@
...
+ public:
+  bst_row_t base_rowid = 0;
+
+  static auto SearchCutValue(bst_row_t ridx, bst_feature_t fidx, GHistIndexMatrix const &index,
+                             std::vector<uint32_t> const &cut_ptrs,
+                             std::vector<float> const &cut_values) {
+    int32_t gidx = -1;
+    auto const &row_ptr = index.row_ptr;
+    auto get_row_ptr = [&](size_t ridx) { return row_ptr.at(ridx - index.base_rowid); };
+
```
RAMitchell, line 40: This function looks the same as other places in the code base. Can (or should) it be extracted?

RAMitchell (approved):

src/tree/updater_approx.h
trivialfis: Right now there's one similar function in histogram builder, but that one is specialized over first page as it's in the hot loop of building histogram.

PR #7471, opened 2021-11-23
[R] Export function xgb.shap.data

hcho3:
https://discuss.xgboost.ai/t/xgb-shap-data-not-found/2557?u=hcho3

Even though the utility function `xgb.shap.data` was visible in the R-package documentation ([here](https://www.rdocumentation.org/packages/xgboost/versions/1.5.0.2/topics/xgb.shap.data)), it was not explicitly published from `NAMESPACE`. The function should be exported so that it can be actually used.


trivialfis (commented):
Does the user actually need to call that function in order to use SHAP related features?  Just curious since it's a new public API function.

hcho3:
@trivialfis It's supposed to be an internal function, but for some reason it is already part of the public documentation.

trivialfis:
If no one can use it, let's remove the public document instead. ;-)

trivialfis:
I will leave it for you to decide.  I'm not familiar with R so I worried about a new function to maintain.  If you think it's best to export the function then this PR looks good to me.

hcho3:
Closing this, as `xgb.shap.data` was marked as internal in #6374, and the document has been regenerated. https://github.com/dmlc/xgboost/blob/master/R-package/man/xgb.plot.shap.Rd no longer shows `xgb.shap.data`

PR #7474, opened 2021-11-23
Fix github macos package upload.

trivialfis:


trivialfis:
@hcho3 Another small fix for the config field.

PR #7475, opened 2021-11-23
Fix macos package upload.

trivialfis:
This splits up the macos and win tests so they can use different conda env.


hcho3 (approved):

trivialfis:
The upload on github action seems to be successful, but I couldn't find the wheel on s3 bucket. ;-(

PR #7481, opened 2021-11-24
Fix external memory, gpu_hist and subsampling combination bug. (#7476)

### LINKED ISSUE #7476: External memory, gpu_hist and subsampling combination bug
Hi,

I've detected that xgboost segfaults when I try to use subsampling with external memory and gpu_hist as a parameter.

Segfault happens when the number of batches is greater than 1 and subsampling is less than 1.0.

Here I send the script that causes the segfault.

I have a solution for the bug, a pull request follows quickly.

[external_memory_3.zip](https://github.com/dmlc/xgboost/files/7589866/external_memory_3.zip)

Best Regards

GinkoBalboa:
- The error happens because when reading from external memory the batch is
  reassembled for every new iteration. The variable `original_page_` is
  initialized from the first batch, when the constructor of `GradiendBasedSample`
  is called. After iterating through the batches the original memory is not
  accessible, so when trying to access the memory pointed by `original_page_`
  causes an error.

- The solution is instead of accessing data from the `original_page_`, to access
  the data from the first page of the available batch.

fix #7476

### DIFF
--- a/src/tree/gpu_hist/gradient_based_sampler.cu
+++ b/src/tree/gpu_hist/gradient_based_sampler.cu
@@ -185,12 +185,10 @@ GradientBasedSample UniformSampling::Sample(common::Span<GradientPair> gpair, DM
   return {dmat->Info().num_row_, page_, gpair};
 }
 
-ExternalMemoryUniformSampling::ExternalMemoryUniformSampling(EllpackPageImpl const* page,
-                                                             size_t n_rows,
+ExternalMemoryUniformSampling::ExternalMemoryUniformSampling(size_t n_rows,
                                                              BatchParam batch_param,
                                                              float subsample)
-    : original_page_(page),
-      batch_param_(std::move(batch_param)),
+    : batch_param_(std::move(batch_param)),
       subsample_(subsample),
       sample_row_index_(n_rows) {}
 
@@ -218,15 +216,17 @@ GradientBasedSample ExternalMemoryUniformSampling::Sample(common::Span<GradientP
                     sample_row_index_.begin(),
                     ClearEmptyRows());
 
+  auto batch_iterator = dmat->GetBatches<EllpackPage>(batch_param_);
+  auto first_page = (*batch_iterator.begin()).Impl();
   // Create a new ELLPACK page with empty rows.
   page_.reset();  // Release the device memory first before reallocating
   page_.reset(new EllpackPageImpl(
-      batch_param_.gpu_id, original_page_->Cuts(), original_page_->is_dense,
-                                  original_page_->row_stride, sample_rows));
+      batch_param_.gpu_id, first_page->Cuts(), first_page->is_dense,
+                           first_page->row_stride, sample_rows));
 
   // Compact the ELLPACK pages into the single sample page.
   thrust::fill(dh::tbegin(page_->gidx_buffer), dh::tend(page_->gidx_buffer), 0);
-  for (auto& batch : dmat->GetBatches<EllpackPage>(batch_param_)) {
+  for (auto& batch : batch_iterator) {
     page_->Compact(batch_param_.gpu_id, batch.Impl(), dh::ToSpan(sample_row_index_));
   }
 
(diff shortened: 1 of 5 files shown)


src/tree/gpu_hist/gradient_based_sampler.cu
```diff
@@ -218,15 +216,17 @@ GradientBasedSample ExternalMemoryUniformSampling::Sample(common::Span<GradientP
...
   page_.reset(new EllpackPageImpl(
-      batch_param_.gpu_id, original_page_->Cuts(), original_page_->is_dense,
-                                  original_page_->row_stride, sample_rows));
+      batch_param_.gpu_id, first_page->Cuts(), first_page->is_dense,
+                           first_page->row_stride, sample_rows));
 
   // Compact the ELLPACK pages into the single sample page.
   thrust::fill(dh::tbegin(page_->gidx_buffer), dh::tend(page_->gidx_buffer), 0);
-  for (auto& batch : dmat->GetBatches<EllpackPage>(batch_param_)) {
+  for (auto& batch : batch_iterator) {
```
trivialfis, line 229: Is it possible that the fit page is being duplicated?

trivialfis (commented):
Thanks for the fix, removing a state is great!  One question is inlined in comment.


Could you please add a test here: https://github.com/dmlc/xgboost/blob/820e1c01ef3cdbbe9f5483ab85d24b97d79138db/tests/python/test_data_iterator.py#L66 so that we won't make the same mistake again in the future.  Thanks for the nice work!

src/tree/gpu_hist/gradient_based_sampler.cu
GinkoBalboa: Do you mean the new constructor called with first_page data and then again using the same page in the iterator? 

If that is the question, then I think the constructor only uses first_page to get some info on how to reserve memory for page_. Later, in the iterator, the real processing is performed so first_page must appear again there.

GinkoBalboa:
> Thanks for the fix, removing a state is great! One question is inlined in comment.
> 
> Could you please add a test here:
> 
> https://github.com/dmlc/xgboost/blob/820e1c01ef3cdbbe9f5483ab85d24b97d79138db/tests/python/test_data_iterator.py#L66
> so that we won't make the same mistake again in the future. Thanks for the nice work!

I've added the test you suggested. Just copied the existing code from run_data_iterator test into a new test, and added subsample value less than 1. 

Cheers!

trivialfis:
Let me take a closer look and fix the test

GinkoBalboa:
Ok, thanks for helping me. Please, take your time, don't hurry, I clicked "request review" by mistake :-)

GinkoBalboa:
- I put the test into a separate function to check out where exactly fails. 
- I merged the latest change, because I was thinking maybe the new CI changes could mean something.
- Noticed after the last fail that the failure was because of greater differences between predicted and expected, so I increased the relative tolerance from 1e-2 to 5e-2. I think the error is greater because we use subsampling. Hope this time the test will pass.

trivialfis:
Thank you for the nice work.  I will get back to this tomorrow.  I don't have access to my device at the moment.

GinkoBalboa:
Ok, now I understand what you meant when you told me to add test. Nice work, thank you.

I just added the relaxing of rtol when subsample < 1.0. This is the main reason for test fail. I added scaled rtol, lets see if it will pass.

trivialfis:
Will take a closer look.  The quantile is created differently with batched data, so testing prediction result is not reliable.

trivialfis (approved):
The fix itself looks good.  Will merge once all tests are green.  Handling floating-point errors in these kinds of tests are always messy...

Thanks for the good work!

trivialfis:
Opened an issue for the major cause of floating-point error: https://github.com/dmlc/xgboost/issues/7488  Will follow up next week. (I'm currently out of the office, sorry).

trivialfis:
Thanks for the fix and sorry for the long delay.

PR #7487, opened 2021-11-26
Updated CMake version requirement in build.rst

danmarinescu:
The documentation states that to build from source you need CMake 3.13 or higher. However, according to https://github.com/dmlc/xgboost/blob/master/CMakeLists.txt#L1 CMake 3.14 or higher is required. This PR updates the documentation to state that CMake 3.14 or higher is required


trivialfis (approved):
Thanks for the PR!

Issue #7484, opened 2021-11-25
 Socket Connect Error:Connection refused


kitty-eu-org:
### error
`When I was using spark xgboost, an error occurred：
  retry connect to ip(retry time 1): [ip]
  retry connect to ip(retry time 1): [ip]
  retry connect to ip(retry time 1): [ip]
  retry connect to ip(retry time 1): [ip]
  retry connect to ip(retry time 1): [ip]
  retry connect to ip(retry time 1): [ip]
  retry connect to ip(retry time 1): [ip]
  retry connect to ip(retry time 1): [ip]
  retry connect to ip(retry time 1): [ip]
  retry connect to ip(retry time 1): [ip]
  retry connect to ip(retry time 2): [ip]
  retry connect to ip(retry time 2): [ip]
  retry connect to ip(retry time 2): [ip]
  retry connect to ip(retry time 2): [ip]
  retry connect to ip(retry time 2): [ip]
  retry connect to ip(retry time 2): [ip]
  retry connect to ip(retry time 2): [ip]
  retry connect to ip(retry time 2): [ip]
  retry connect to ip(retry time 2): [ip]
  retry connect to ip(retry time 2): [ip]
  retry connect to ip(retry time 3): [ip]
  retry connect to ip(retry time 3): [ip]
  retry connect to ip(retry time 3): [ip]
  retry connect to ip(retry time 3): [ip]
  retry connect to ip(retry time 3): [ip]
  retry connect to ip(retry time 3): [ip]
  retry connect to ip(retry time 3): [ip]
  retry connect to ip(retry time 3): [ip]
  retry connect to ip(retry time 3): [ip]
  retry connect to ip(retry time 3): [ip]
  retry connect to ip(retry time 4): [ip]
  retry connect to ip(retry time 4): [ip]
  retry connect to ip(retry time 4): [ip]
  retry connect to ip(retry time 4): [ip]
  retry connect to ip(retry time 4): [ip]
  retry connect to ip(retry time 4): [ip]
  retry connect to ip(retry time 4): [ip]
  retry connect to ip(retry time 4): [ip]
  retry connect to ip(retry time 4): [ip]
  retry connect to ip(retry time 4): [ip]
  connect to (failed): [ip]
  connect to (failed): [ip]
  Socket Connect Error:Connection refused
  Socket Connect Error:Connection refused`
### env
- ubuntu
- spark2.3.1
- scala2.11
- xgboost realease0.9.0
- Firewall off
[...]

PR #7495, opened 2021-12-03
add System.exit(*)

hezhaozhao-git:
xgboost has changed RabitTracker.java . I think that once the tracker server process dies, the entire java program should also die, because all subsequent calculations depend on the tacker process.
You can check the [issue](https://github.com/dmlc/xgboost/issues/7484) for details

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/RabitTracker.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/RabitTracker.java
@@ -1,5 +1,7 @@
 package ml.dmlc.xgboost4j.java;
 
+
+
 import java.io.*;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


wbo4958:
I don't think it's a good solution to add "System.exit()" since it will abort the whole process which may cause some trouble for the end-users.

Some users train an XGBoost model in a thread and have many other pipelines after training no matter if the training process is successful. If the whole process is aborted, it's really painful for users.  We should avoid that.

hezhaozhao-git:
> I don't think it's a good solution to add "System.exit()" since it will abort the whole process which may cause some trouble for the end-users.
> 
> Some users train an XGBoost model in a thread and have many other pipelines after training no matter if the training process is successful. If the whole process is aborted, it's really painful for users. We should avoid that.

But if the tracker process dies, then the final training will still fail, because the training process relies on the tracker to communicate, and if the jvm process does not die, it will quickly cover up the error message of the tracker process. I think this is not a good idea.

wbo4958:
XGBoost is just a library, and Users just call some APIs from XGBoost. If XGBoost aborts the users process which is really not an ideal way. We should leave that choice to the user. 
Eg, users can decide whether to abort the process or not by themself. Some users detect if xgboost failed to train, they may switch to another Machine Library for training. and XGBoost should not block their whole pipeline.

hezhaozhao-git:
> 



> XGBoost is just a library, and Users just call some APIs from XGBoost. If XGBoost aborts the users process which is really not an ideal way. We should leave that choice to the user. Eg, users can decide whether to abort the process or not by themself. Some users detect if xgboost failed to train, they may switch to another Machine Library for training. and XGBoost should not block their whole pipeline.

Oh, yes, I understand what you said, System.exit() pr I will close it, but I think tracker.py needs to be improved, because it will affect the opening of the training task and the task being trained, what do you think how?

PR #7514, opened 2021-12-16
Initial support for multioutput regression.

trivialfis:
Close https://github.com/dmlc/xgboost/pull/7309 .

* Add num target model parameter, which is configured from input labels.
* Change elementwise metric and indexing for weights.
* Add demo.
* Add tests.

### DIFF
--- a/src/objective/regression_obj.cu
+++ b/src/objective/regression_obj.cu
@@ -108,7 +108,7 @@ class RegLossObj : public ObjFunction {
 
           for (size_t idx = begin; idx < end; ++idx) {
             bst_float p = Loss::PredTransform(preds_ptr[idx]);
-            bst_float w = _is_null_weight ? 1.0f : weights_ptr[idx % n_targets];
+            bst_float w = _is_null_weight ? 1.0f : weights_ptr[idx / n_targets];
             bst_float label = labels_ptr[idx];
             if (label == 1.0f) {
               w *= _scale_pos_weight;
--- a/src/objective/regression_obj.cu
+++ b/src/objective/regression_obj.cu
@@ -56,9 +56,9 @@ class RegLossObj : public ObjFunction {
     return Loss::Info();
   }
 
-  uint32_t Targets(DMatrix const* m) const override {
+  uint32_t Targets(MetaInfo const& info) const override {
     // Multi-target regression.
-    return std::max(static_cast<size_t>(1), m->Info().labels.Shape(1));
+    return std::max(static_cast<size_t>(1), info.labels.Shape(1));
   }
 
   void GetGradient(const HostDeviceVector<bst_float>& preds,
--- a/src/objective/regression_obj.cu
+++ b/src/objective/regression_obj.cu
@@ -56,6 +56,11 @@ class RegLossObj : public ObjFunction {
     return Loss::Info();
   }
 
+  uint32_t Targets(DMatrix const* m) const override {
+    // Multi-target regression.
+    return std::max(static_cast<size_t>(1), m->Info().labels.Shape(1));
+  }
+
   void GetGradient(const HostDeviceVector<bst_float>& preds,
                    const MetaInfo &info, int,
                    HostDeviceVector<GradientPair>* out_gpair) override {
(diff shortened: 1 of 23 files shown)


src/objective/regression_obj.cu
```diff
@@ -101,7 +108,7 @@ class RegLossObj : public ObjFunction {
 
           for (size_t idx = begin; idx < end; ++idx) {
             bst_float p = Loss::PredTransform(preds_ptr[idx]);
-            bst_float w = _is_null_weight ? 1.0f : weights_ptr[idx];
+            bst_float w = _is_null_weight ? 1.0f : weights_ptr[idx % n_targets];
```
hcho3, line 111: Are weights now per-target? Previously, weights are per-instance.

trivialfis: Will fix it.

trivialfis: Thanks for the catch.

tests/cpp/metric/test_multiclass_metric.cc
```diff
@@ -11,6 +11,7 @@ inline void CheckDeterministicMetricMultiClass(StringView name, int32_t device)
 
   HostDeviceVector<float> predts;
   MetaInfo info;
+
```
trivialfis, line 14: Remove this.

src/objective/regression_obj.cu
```diff
@@ -83,8 +88,10 @@ class RegLossObj : public ObjFunction {
     // for better performance.
     const size_t n_data_blocks = std::max(static_cast<size_t>(1), (on_device ? ndata : nthreads));
     const size_t block_size = ndata / n_data_blocks + !!(ndata % n_data_blocks);
+    auto const n_targets = std::max(info.labels.Shape(1), static_cast<size_t>(1));
+
```
trivialfis, line 92: We should use more ndarray friendly ways to express the calculation in the future.

hcho3:
Can we throw an error when the user attempt to use other objectives with multi-output labels?

trivialfis:
> Can we throw an error when the user attempt to use other objectives with multi-output labels?

Added a check along with test.  Thanks for the suggestion.

trivialfis:
The support is primitive.  We need to expand it to other regression objectives and add more documents.

hcho3 (approved):

Craigacp:
Is this API suitable for wrapping up in XGBoost4J, or do you want to build out a more general one before doing that?

trivialfis:
The API is good for regression and binary classification and should be stable unless major issue is found.  So it's good enough to start looking into language bindings.  I will try to add xgboost4j support later.  Will keep you posted on the progress.

It's unlikely that we will implement multi class multi target based on existing interface so no need to worry about it at the moment. :-)

trivialfis:
@Craigacp Apologies for the slow progress. I looked into the JVM packages and I can't build a complete stack for jvm from spark down to basic java wrapper.  The required change is not trivial since we will have 2-dim inputs for both base_margin and label, I'm not sure what's the best way to implement that for jvm packages.

Craigacp:
Do you have a branch somewhere with the current state of it? I'd assumed that DMatrix would grow to accept a 2d matrix for the targets and then most of the rest of the changes would be plumbing, but if there's something that needs some design effort I can take a look.

trivialfis:
@Craigacp The master branch contains support for Python.  If you are interested in a code walk I'm happy to chat offline.

> I'd assumed that DMatrix would grow to accept a 2d matrix for the targets

Yes, that's pretty much the only requirement for regression.  For multi-label classification some configuration needs to be done to make sure `binary:logistic` is used and no `num_class` is passed in.

trivialfis:
On Python we use a JSON string to represent the input memory buffer, so we have complete information about shape and strides, etc.  I'm not sure what's the best way to handle these numeric data types on JVM.

Craigacp:
Ok, I should have some time next week to look through how the python code works. I might take you up on that code walk through once I've familiarised myself with it a bit.

The length of the header array should tell us the number of examples, and then for multidimensional things we can require that the target array not be sparse and have explicit zeros (though that might be pretty wasteful for multi-label classification). Then the target array is of known shape `[numExamples, numOutputDimensions]` linearised into a single vector (row-wise) and that should be enough information for the C API. But there might be a better way to do it.

PR #7520, opened 2021-12-18
[TEST][MINOR] Add skipif no_matplotlib for test_callbacks_demo

xwu99:
callbacks.py is using matplotlib, the test will be failed if matplotlib is not available.


trivialfis (approved):
Thanks for the fix!

PR #7523, opened 2021-12-20
[jvm-package] Clean up the legacy gpu support tests

wbo4958:
This PR is to remove gpu-with-gpu-test profile which previously tested the CPU load + GPU train. But with rapids plugin added, the related tests have been removed, instead, we added more tests for GPU load + GPU train.

BTW, this PR will leave GpuTestSuite unused, Do we need to clean this up?


trivialfis:
> BTW, this PR will leave GpuTestSuite unused, Do we need to clean this up?

If it's no longer useful feel free to remove them.

trivialfis (approved):
Thanks for the cleanup!  Looks good as long as all tests are green.

PR #7529, opened 2021-12-22
Clarify the behavior of invalid categorical value handling.

trivialfis:


trivialfis:
In this PR it's treated as the same as a missing value. Maybe we can go to a simpler route like always left.

hcho3 (approved):

Issue #7480, opened 2021-11-24
Docs: Use uppercase omega for overall model complexity cost, lowercase omega for per tree complexity cost


vertexcite:
Documentation suggestion:
For https://github.com/dmlc/xgboost/blob/master/doc/tutorials/model.rst, use lowercase omega for per tree complexity cost, and uppercase omega for overall model complexity cost.  Currently, uppercase omega used for both.  

Similarly, https://raw.githubusercontent.com/dmlc/web-data/master/xgboost/model/step_fit.png should have :math:`\Omega(\theta)` (i.e. uppercase omega for overall model complexity cost: function applied to parameters theta, not individual tree f).

hcho3:
Thanks for your suggestion. Would you be willing to submit a pull request?

vertexcite:
@hcho3 Apologies for missing your reply until now.  I've created a pull request for the .rst doc.  Do you know how the image https://raw.githubusercontent.com/dmlc/web-data/master/xgboost/model/step_fit.png  was produced?

trivialfis:
Closed by https://github.com/dmlc/xgboost/pull/7532

PR #7532, opened 2021-12-23
[doc] Lowercase omega for per tree complexity

vertexcite:
As suggested on issue #7480

I think it should be fine to merge.

### DIFF
--- a/doc/tutorials/model.rst
+++ b/doc/tutorials/model.rst
@@ -97,11 +97,13 @@ Mathematically, we can write our model in the form
 
   \hat{y}_i = \sum_{k=1}^K f_k(x_i), f_k \in \mathcal{F}
 
-where :math:`K` is the number of trees, :math:`f` is a function in the functional space :math:`\mathcal{F}`, and :math:`\mathcal{F}` is the set of all possible CARTs. The objective function to be optimized is given by
+where :math:`K` is the number of trees, :math:`f_k` is a function in the functional space :math:`\mathcal{F}`, and :math:`\mathcal{F}` is the set of all possible CARTs. The objective function to be optimized is given by
 
 .. math::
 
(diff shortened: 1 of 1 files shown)


doc/tutorials/model.rst
```diff
@@ -97,11 +97,13 @@ Mathematically, we can write our model in the form
 
   \hat{y}_i = \sum_{k=1}^K f_k(x_i), f_k \in \mathcal{F}
 
-where :math:`K` is the number of trees, :math:`f` is a function in the functional space :math:`\mathcal{F}`, and :math:`\mathcal{F}` is the set of all possible CARTs. The objective function to be optimized is given by
+where :math:`K` is the number of trees, :math:`f_k` is a function in the functional space :math:`\mathcal{F}`, and :math:`\mathcal{F}` is the set of all possible CARTs. The objective function to be optimized is given by
```
trivialfis, line 100: I think `f_k` is unnecessary since it's referring to the general function in the space.

trivialfis (commented):
Thanks for the improvements.  One small question in comment.

doc/tutorials/model.rst
vertexcite: Thanks @trivialfis.  My rationale was:`f` isn't in the general function space ``:math:`\mathcal{F}` `` until it has its subscript.

Since " ``:math:`f_k` `` is a function in the functional space ``:math:`\mathcal{F}` `` "
is just a translation of " `f_k \in \mathcal{F}` " from the previous line, another alternative would be to just omit it, so that the line reads:

" where ``:math:`K` `` is the number of trees, and ``:math:`\mathcal{F}` `` is the set of all possible CARTs. "

trivialfis: Thank you for the explanation.

trivialfis (approved):

PR #7536, opened 2021-12-25
Remove macro definitions of snprintf and vsnprintf

ldesreumaux:
These macros can create problems in projects that include xgboost and use vsnprintf/snprintf.

vsnprintf and snprintf are supported since Visual Studio 14, so we don't need these macros anymore.


trivialfis (approved):

gomlfx:
Thank god I just have Autogluon for this now! LOL

Issue #7158, opened 2021-08-06
No supported methods for JSON serialization of trained XGBoost model without using disk


amohar2:
XGBoost's Learning API for Python provides the following ways to save/load the model:
1. save_config/load_config: this will save/load the internal Booster configuration parameters as JSON. However, it seems that save_config does not save all the required information to load Booster object from scratch. For example, the following code:

`import xgboost
from sklearn.datasets import make_classification

x, y = make_classification(50, 21, 6, n_classes=2)
xtest, ytest = make_classification(10, 21, 6, n_classes=2)
param = {
    'eta': 0.3,
    'max_depth': 3,
    'objective': 'multi:softprob',
    'num_class': 3}
dtrain = xgboost.DMatrix(x, label=y)
dtest = xgboost.DMatrix(xtest)
booster = xgboost.train(param, dtrain)
booster.predict(dtest)

saved_config = booster.save_config()

booster = xgboost.Booster()
booster.load_config(saved_config)
booster.predict(dtest)`

will produce the following error:
Traceback (most recent call last):
  File "test_xgb.py", line 23, in <module>
    booster.predict(dtest)
  File "python3.8/site-packages/xgboost/core.py", line 1749, in predict
    _check_call(
  File "python3.8/site-packages/xgboost/core.py", line 210, in _check_call
    raise XGBoostError(py_str(_LIB.XGBGetLastError()))
xgboost.core.XGBoostError: [23:39:08] xgboost/src/learner.cc:602: Check failed: mparam_.num_feature != 0 (0 vs. 0) : 0 feature is supplied.  Are you using raw Booster interface?

2. save_model/load_model: using these two functions, we can save the model from disk and load it back successfully. However, in my particular application, I want to avoid writing the saved model to disk, and save_model() does not provide this option.

3. save_raw/load_model: using save_raw(), we can save the model into a buffer in memory, and load it back with load_model(). It seems that save_raw() always returns a bytes type object that will be used in load_model(). However, similar to pickle, this will create security issues, since malicious code can be executed using bytes type objects.

Are there any other serialize/deserialize functions in XGBoost Learning/Scikit-Learn APIs that address the following requirements:
1. save the model as string or JSON in memory (not bytes/bytearray)
2. can be used to initialize the XGBoost model at a later time

hcho3:
Currently, no, there is no serialization method that meets all your requirements.

trivialfis:
> save the model as string or JSON in memory (not bytes/bytearray)

[...]

PR #7145, opened 2021-07-29
[WIP] Use model slicing for copy.

trivialfis:
Related: https://github.com/dmlc/xgboost/issues/7138 https://github.com/dmlc/xgboost/issues/6697

Avoid serializing into JSON during copy.  By default, XGBoost makes a copy after training for releasing memory.  This PR optimizes the copy function to use model slicing instead of serialization.  This helps when creating many large models using scikit-learn grid search.

Issue #6697, opened 2021-02-09
Introduced cpu performance gaps disscussion


ShvetsKS:
Two performance problems were discovered in recently merged PRs:
1) thread safety prediction: https://github.com/dmlc/xgboost/pull/6648

santander training stage | before #6648 | thread safe #6648 and master | https://github.com/dmlc/xgboost/pull/6696 PR
-- | -- | -- | --
full training | 136s | 165s | 140s
PredictRaw | 55s | 78s | 59s

Would it be possible to have not thread safe `PredictDMatrix` call to reduce overheads for subsampling cases and have initialization buffer only once on first training iteration? I'd appreciate for any ideas.

2) drop binary format: https://github.com/dmlc/xgboost/pull/6513 leads to stable 10% perf regression, observed for Higgs1m dataset after this commit (16s vs 14.3s full training).
Before that we have an option to save old behavior via setting `enable-experimental-json-serialization` to `False`  but currently there is no such possibility.
Could you share your thoughts about best way to mitigate it?

trivialfis:
I'm confused that why is thread-safe prediction has any impact on training?  I thought we have enabled prediction cache for bosth regression and multi-class on CPU?

trivialfis:
If the prediction happens during evaluation, we can use thread local static storage I believe.


As for the JSON thing, it's occurred during the end of training, which is used to release memory.

ShvetsKS:
> I'm confused that why is thread-safe prediction has any impact on training? I thought we have enabled prediction cache for both regression and multi-class on CPU?

Yes prediction caching is enabled for regression, binary and multiclass classification but for `subsample==1` case only

ShvetsKS:
> As for the JSON thing, it's occurred during the end of training, which is used to release memory.

is it possible to have some alternative `enable-experimental-json-serialization` to reduce observed gaps?

trivialfis:
Thanks for the reply.  I think it's possible to change the subsampling implementation to allow cache.

As for the serialization time, I will try to find a way to remove it, either by BSON or by a better way to release memory.

trivialfis:
The thread safety was mostly for dask interface, also some other feature requests.  In dask, the prediction is done on each block of data in parallel.

ShvetsKS:
> The thread safety was mostly for dask interface, also some other feature requests. In dask, the prediction is done on each block of data in parallel.

[...]

Issue #4060, opened 2019-01-17
Ran into an issue with numerical precision and thresholds


slundberg:
For a (not-sharable) dataset I was working on XGBoost chose thresholds that were very close to machine epsilon from actual training set feature values (less than `np.finfo(np.float32).eps` away). This lead to the JSON dump of the model being wrong, since the conversion to base-10 destroyed the information needed to route down the correct path. This meant the only way to read a tree that would behave the same as XGBoost was the use the raw memory dump.

I am posting this issue so this danger is noted for others, and in case it would be worth seeing in the future why XGBoost is choosing such unstable thresholds. It seems like just asking for trouble when exporting models and when running them on different machines, since an epsilon change in a feature value causes a potentially meaningful change in the model output (since a sample now goes down different paths).

hcho3:
Would adding more decimal digits in the dump help? But I do agree that the choice of thresholds is quite weird. It would be nice if we can reproduce the issue with a public dataset.

slundberg:
It might, but since base 10 and 2 don't line up exactly it would be tricky to ensure it was always right. Finding an example where the JSON is wrong would take a bit more work, but below is a simple example that shows the kind of very tight thresholds XGBoost creates (note it depends on the latest benchmark branch of shap to read the values without using the JSON output):

```python
import xgboost
import shap
import numpy as np

X,y = shap.datasets.adult()

model = xgboost.XGBClassifier(n_estimators=1, max_depth=4, random_state=0)
model.fit(X, y)

explainer = shap.TreeExplainer(model)

# Find close thresholds
Xv = X.values
for i in range(explainer.model.thresholds.shape[0]):
    for j in range(explainer.model.num_nodes[i]):
        for k in range(Xv.shape[0]):
            v = Xv[k,explainer.model.features[i,j]]
            t = explainer.model.thresholds[i,j]
            if abs(v - t) < 1e-5:
                print(
                    "The value of feature",
                    explainer.model.features[i,j],
                    "for sample",k,"is",v,"and a split threshold for that feature is",
                    t,". That's a difference of",v - t,"."
                )
```
output:
```
The value of feature 0 for sample 8381 is 85.0 and a split threshold for that feature is 84.99999 . That's a difference of 7.62939453125e-06 .
The value of feature 0 for sample 20463 is 85.0 and a split threshold for that feature is 84.99999 . That's a difference of 7.62939453125e-06 .
[...]

PR #7545, opened 2022-01-08
[RFC] Add support for Universal Binary JSON

trivialfis:
This is an RFC for adding binary JSON format (https://ubjson.org/) to XGBoost for
serialization to resolve the performance issues.  In the original proposal of revising the
serialization format, one of the reasons we choose JSON is that there are multiple
specifications of binary equivalence.

## Motivation

The serialization can occur in multiple places, including saving model, pickle,
transferring the model between distributed workers, releasing memory at the end of
training (copy).  We have adopted JSON format in the past, which cleanups the
serialization format, but also introduced performance overhead.  As originally proposed,
we will add a binary implementation when needed.  The binary JSON can close the gap
between the current JSON format and our old binary format, allowing us to phase out support
for the old binary models.

## Universal Binary JSON
Universal Binary JSON (https://ubjson.org/) is a specification for serializing JSON
document into the binary format with a focus on performance and ease of implementation.

So far I have found 2 promising candidates.  First is the BSON specification.  It's widely
adopted and well maintained with big projects like mongoDB utilizing it.  However it's
complicated, requires known document size, and most importantly it has a limit on the
total document size as 16 megabytes. (https://docs.mongodb.com/manual/reference/limits/).
Another one is the ubjson adopted in this RFC, which is efficient for array objects but has
lesser implementations.  For instance non of the existing Python implementations have full
support for the specification.  It has the concept of a typed container (array and object),
which particularly suits the need for XGBoost since XGBoost has many arrays including
tree weights, node indices, etc.  Also, it's a relatively simple specification that we can
implement easily.

## Implementation

The serializer is easily added on top of existing JSON infrastructure since the JSON
implementation in XGBoost has a modularized reader and writer.  Also, I have changed model
serialization to using typed containers wherever appropriate.

## Performance

Benchmark is carried out on a model trained on airline dataset with 1000 boosting rounds
and 12 max\_depth.  The following result is a total of 1000 calls of `Booster.__copy__`.  In
total 1e6 tree copies are carried out in the test.

|      | Master             | PR                 |
|:-----|:-------------------|:-------------------|
| Time | 1279.9649624824524 | 53.655569553375244 |

Close https://github.com/dmlc/xgboost/pull/7145
Close https://github.com/dmlc/xgboost/issues/6697
Close https://github.com/dmlc/xgboost/issues/7158
Close https://github.com/dmlc/xgboost/issues/4060

## TO-DOs
- [x] Remove the experimental tag for JSON model format.
- [x] Investigate JVM support.
- [x] More documents.
- [ ] Add support to the SHAP package.
- [x] Add warnings according to the roadmap.
- [x] Roadmap for deprecating the old binary format.


hcho3:
I approve the idea of adopting UBJSON to speed up serialization.

> Close #7158

Do you plan to implement a method to serialize Booster object as a string or byte-buffer (without using the disk)?

trivialfis:
We already have that feature, in R that's xgb.save.raw

trivialfis:
Hmm .. the SHAP package is using that raw buffer ...

hcho3:
@trivialfis There isn't an equivalent feature in Python.

trivialfis:
@hcho3 It's a function that calls `Learner.SaveModel`.  This PR changes it to use UBJSON from old binary format, which I might change back due to breaking.

trivialfis:
But the idea is that it's quite easy to change the format for that function.

hcho3:
@trivialfis Can we add an optional argument to `save_raw()`, like `save_raw(dump_format='json')` ? We can set the argument default to `binary` for now, but after a while we can change the default to `json` or `ubjson`.

trivialfis:
Yes, we can.  But we need a new C api function.

hcho3:
Got it. Let's add a new C API function.

trivialfis:
Added a new API function along with support for R and Python.

trivialfis:
@hcho3 I plan to gradually remove the support for the old binary model.  A simple roadmap would be

- Release 1.6: Add ubjson to python/R/jvm, emit a warning about the model format.  The jvm part might be more difficult, we will see.
- Release 2.0: Continue the warning.
- Release 2.1: Continue the warning.
- Release 2.2: Remove support for saving to old binary format.  But keep the loading part with a warning.
- Release 2.3: Remove support for loading old binary format.

I will open an issue for the roadmap if it's feasible.

hcho3:
@trivialfis Can you put up a roadmap issue for phasing out the binary serialization format?

RAMitchell:
This looks very nice. Can we expect dask performance to improve? In the past I remember model serialisation in a multi-process setting to be a significant bottleneck.

trivialfis:
> Can we expect dask performance to improve? In the past I remember model serialisation in a multi-process setting to be a significant bottleneck.

We have optimization on dask to avoid model serialization, specifically, users can pass a future object to xgb.dask.predict.  But for training large models this can still be significant, for instance https://github.com/dmlc/xgboost/issues/5474#issuecomment-976561515 .

jvm-packages/xgboost4j/src/native/xgboost4j.cpp
```diff
@@ -662,20 +662,24 @@ JNIEXPORT jint JNICALL Java_ml_dmlc_xgboost4j_java_XGBoostJNI_XGBoosterLoadModel
...
+ * Signature: (JLjava/lang/String;[[B)I
  */
-JNIEXPORT jint JNICALL Java_ml_dmlc_xgboost4j_java_XGBoostJNI_XGBoosterGetModelRaw
-  (JNIEnv * jenv, jclass jcls, jlong jhandle, jobjectArray jout) {
+JNIEXPORT jint JNICALL Java_ml_dmlc_xgboost4j_java_XGBoostJNI_XGBoosterSaveModelToBuffer
+  (JNIEnv * jenv, jclass jcls, jlong jhandle, jstring jformat, jobjectArray jout) {
   BoosterHandle handle = (BoosterHandle) jhandle;
+  const char *format = jenv->GetStringUTFChars(jformat, 0);
   bst_ulong len = 0;
   const char* result;
```
trivialfis, line 673: Initialize it.

jvm-packages/xgboost4j/src/native/xgboost4j.cpp
```diff
@@ -662,20 +662,24 @@ JNIEXPORT jint JNICALL Java_ml_dmlc_xgboost4j_java_XGBoostJNI_XGBoosterLoadModel
...
-  int ret = XGBoosterGetModelRaw(handle, &len, &result);
-  JVM_CHECK_CALL(ret);
+  xgboost::Json config {xgboost::Object{}};
+  config["format"] = std::string{format};
+  std::string config_str;
+  xgboost::Json::Dump(config, &config_str);
 
+  int ret = XGBoosterSaveModelToBuffer(handle, config_str.c_str(), &len, &result);
```
trivialfis, line 679: Early return.

src/c_api/c_api.cc
```diff
@@ -900,45 +894,110 @@ XGB_DLL int XGBoosterPredictFromCUDAColumnar(
...
   } else {
     std::unique_ptr<dmlc::Stream> fi(dmlc::Stream::Create(fname, "r"));
     static_cast<Learner*>(handle)->LoadModel(fi.get());
   }
   API_END();
 }
 
-XGB_DLL int XGBoosterSaveModel(BoosterHandle handle, const char* c_fname) {
+namespace {
+void WarnOldModel() {
```
trivialfis, line 920: Typo

src/c_api/c_api_utils.h
```diff
@@ -241,5 +241,25 @@ inline void GenerateFeatureMap(Learner const *learner,
 }
 
 void XGBBuildInfoDevice(Json* p_info);
+
+template <typename JT>
+inline auto RequiredArg(Json in, std::string const &key, StringView func) {
```
trivialfis, line 246: Should accept and return references.

doc/tutorials/saving_model.rst
```diff
@@ -2,16 +2,18 @@
...
-the future until JSON format is no-longer experimental and has satisfying performance.
-This tutorial aims to share some basic insights into the JSON serialisation method used in
-XGBoost.  Without explicitly mentioned, the following sections assume you are using the
-JSON format, which can be enabled by providing the file name with ``.json`` as file
-extension when saving/loading model: ``booster.save_model('model.json')``.  More details
-below.
+open format that can be easily reused.  Later in XGBoost 1.6.0, additional support for
+`Universal Binary JSON <https://ubjson.org/>`__ is added as an optimization for more
+efficient model IO.  They have the same document structure with different representations,
+and we will refere them collectively as the JSON format. This tutorial aims to share some
```
trivialfis, line 11: typo.

trivialfis:
All merged as separated PRs.

PR #7548, opened 2022-01-10
Silent some warnings for unused variable.

trivialfis:
Some variables are not used if XGBoost is not compiled with CUDA.


RAMitchell (approved):
Could maybe do with a comment. This is unusual C++ code so it helps the reader understand why this exists.

trivialfis:
Thanks, added a short comment.

PR #7531, opened 2021-12-23
External memory support for hist

trivialfis:
* Generate column matrix from gHistIndex.
* Avoid synchronization with the sparse page once the cache is written.
* Cleanups: Remove member variables/functions, change the update routine to look like approx and gpu_hist.
* Remove pruner.

Close https://github.com/dmlc/xgboost/issues/7354

PR #7551, opened 2022-01-10
Restore functionality of max_depth=0 in hist

### LINKED ISSUE #7543: `max_depth=0` doesn't work since 1.5.0
According to doc, `max_depth=0` means no limit on depth when `grow_policy` is `lossguide` and `tree_method` is `hist`.

See this sample code:
```python
import numpy as np
import xgboost as xgb

x = np.array([[0], [1], [2], [3]])
y = np.array([0, 1, 2, 3])

model = xgb.XGBRegressor(
    n_estimators=1,
    eta=1,
    tree_method="hist",
    grow_policy="lossguide",
    reg_lambda=0,
    max_leaves=128,
    max_depth=0,
).fit(x, y)
assert np.array_equal(model.predict(x), y)
```

The trained model is expected to have one single tree, 4 leaves with depth of 2. The final assertion passes with 1.3.1 and 1.4.2, but fails with 1.5.0 and 1.5.1. When it fails, the single tree in the model has depth of 1.

hcho3:
Closes #7543

### DIFF
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -263,7 +263,7 @@ void QuantileHistMaker::Builder<GradientSumT>::ExpandTree(
       ApplySplit<any_missing>(nodes_for_apply_split, gmat, column_matrix, p_tree);
       SplitSiblings(nodes_for_apply_split, &nodes_to_evaluate, p_tree);
 
-      if (depth < param_.max_depth) {
+      if (param_.max_depth == 0 || depth < param_.max_depth) {
         size_t i = 0;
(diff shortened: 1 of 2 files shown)


trivialfis (approved):
Thanks for the fix.

https://github.com/dmlc/xgboost/pull/7531 unifies the pipeline with `CPUExpandEntry::ChildIsValid` for this purpose, but that PR depends on the rewrite of approx so a quick fix seems to be the best way.

trivialfis:
Tried to backport this PR but there's a conflict.  Will keep it to next major release.

PR #7565, opened 2022-01-14
Fix tree param feature type.

trivialfis:
Changed from `int` to `uint32_t`.


trivialfis:
The issue was found when I was optimizing https://github.com/dmlc/xgboost/pull/7556 .

RAMitchell (approved):

PR #7566, opened 2022-01-14
Deprecate positional arg for sklearn init.

trivialfis:
* Deprecate positional arg for sklearn init.
* Move `max_cat_to_onehot` to proceed the `enable_categorical`.
* Mark it as experimental.


trivialfis:
Will open different PR for more specific changes.

PR #7581, opened 2022-01-19
[dask] Add scheduler address to dask config.

trivialfis:
See added doc for use case.

- Add user configuration.
- Bring back to the logic of using scheduler address from dask.  This was removed when we were trying to support GKE, now we bring it back and let xgboost try it if direct guess or host IP from user config failed.

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -32,6 +32,7 @@ Optional dask configuration
 import platform
 import logging
 import collections
+import socket
 from contextlib import contextmanager
 from collections import defaultdict
 from threading import Thread
@@ -151,11 +152,25 @@ def _multi_lock() -> Any:
     return MultiLock
 
 
-def _start_tracker(n_workers: int, host_ip: Optional[str]) -> Dict[str, Any]:
-    """Start Rabit tracker """
-    env: Dict[str, Union[int, str]] = {'DMLC_NUM_WORKER': n_workers}
-    host = get_host_ip(host_ip)
-    rabit_context = RabitTracker(hostIP=host, n_workers=n_workers, use_logger=False)
+def _start_tracker(
+    n_workers: int, addr_from_dask: str, addr_from_user: Optional[str]
+) -> Dict[str, Any]:
+    """Start Rabit tracker"""
+    env: Dict[str, Union[int, str]] = {"DMLC_NUM_WORKER": n_workers}
+    try:
+        rabit_context = RabitTracker(
+            hostIP=get_host_ip(addr_from_user), n_workers=n_workers, use_logger=False
+        )
+    except socket.error as e:
+        if e.errno != 99:  # not a bind error
+            raise
+        LOGGER.warning(
+            f"Failed to bind address: {get_host_ip(addr_from_user)}, try"
+            f" {addr_from_dask} instead."
+        )
+        rabit_context = RabitTracker(
+            hostIP=addr_from_dask, n_workers=n_workers, use_logger=False
+        )
     env.update(rabit_context.worker_envs())
 
     rabit_context.start(n_workers)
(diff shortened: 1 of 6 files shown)


RAMitchell (commented):
I can't review this :) Can a dask person take a look?

python-package/xgboost/dask.py
```diff
@@ -136,11 +152,26 @@ async def __aexit__(self, *args: Any, **kwargs: Any) -> None:
...
+        )
+    except socket.error as e:
+        if e.errno != 99 or addr_from_dask is None:  # not a bind error
+            raise
+        LOGGER.warning(
+            "Failed to bind address '%s', trying to use '%s' instead.",
+            str(get_host_ip(addr_from_user)),
+            str(addr_from_dask),
+        )
+        rabit_context = RabitTracker(
```
kevingerman, line 172: This statement is guarded in a try/catch block above.  Repeating it in the except clause exposes the module to a catastrophic exception.  Would it be better to make this function recursive on except with a base case that checks that addr_from_user does not equal addr_from_dask before recursing on addr_from_dask.?

python-package/xgboost/dask.py
```diff
@@ -805,12 +837,43 @@ def _dmatrix_from_list_of_parts(
...
+) -> List[bytes]:
+    """Get rabit context arguments from data distribution in DaskDMatrix.
+
+    """
+    # There are 3 possible different addresses:
+    # 1. Provided by user via dask.config
+    # 2. Guessed by xgboost `get_host_ip` function
+    # 3. From dask scheduler
+    # We try 1 and 3 if 1 is available, otherwise 2 and 3.
+    valid_config = ["scheduler_address"]
```
jacobtomlinson, line 851: When extending the Dask config system we typically create a new YAML with default values in and extend the Dask config with it.

See how we do it in [dask-ctl](https://github.com/dask-contrib/dask-ctl/blob/main/dask_ctl/config.py).

python-package/xgboost/dask.py
```diff
@@ -805,12 +837,43 @@ def _dmatrix_from_list_of_parts(
...
+    # See if user config is available
+    if dconfig is not None:
+        for k in dconfig:
+            if k not in valid_config:
+                raise ValueError(f"Unknown configuration: {k}")
+        host_ip: Optional[str] = dconfig.get("scheduler_address", None)
```
jacobtomlinson, line 857: I'm curious why this `dconfig` object is being passed around instead of accessing `dask.config.get("xgboost.scheduler_address")` directly.

kevingerman (commented):
Looks good except for the comment about how the RabitTracker constructor fallback is handled.   Just don't want calls in an except block that can cause another exception.

jacobtomlinson (commented):
The Dask things in here generally seem fine to me. A couple of comments about how the config system is being used.

python-package/xgboost/dask.py
trivialfis: I couldn't access it on the worker side for some reasons.  I read the doc from https://docs.dask.org/en/stable/configuration.html , it's not quite clear to me how to make sure everything is synced.

python-package/xgboost/dask.py
trivialfis: Thank you for sharing, let me look into that.

python-package/xgboost/dask.py
```diff
@@ -174,6 +205,7 @@ def __init__(self, args: List[bytes]) -> None:
 
     def __enter__(self) -> None:
         rabit.init(self.args)
+        assert rabit.is_distributed()
```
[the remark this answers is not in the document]
trivialfis: Just trying to make sure xgboost is initialized correctly.

python-package/xgboost/dask.py
trivialfis: I also read this section https://docs.dask.org/en/stable/configuration.html#downstream-libraries .  In XGBoost dask is loaded lazily so having dask in init might be less desirable due to the size of dependency.  So I went on and read this:


> However, downstream libraries may choose alternative solutions, such as isolating their configuration within their library, rather than using the global dask.config system. All functions in the dask.config module also work with parameters, and do not need to mutate global state.

So far using this parameter passing approach seems to be easiest way to get 1 parameter in.

python-package/xgboost/dask.py
trivialfis: Thanks for the suggestion.  Let me try to factor out this part of the code.

python-package/xgboost/dask.py
jacobtomlinson: That makes sense. Generally the Dask config should be available on the scheduler/workers but that can be deployment dependent.

python-package/xgboost/dask.py
trivialfis: @kevingerman I converted it into a recursive function, but the exception is still raised and depends on dask to handle it.  I think throwing the socket error might be the best way to handle it, dask scheduler/workers will continue to run while xgboost training will be stopped.  What do you think?

kevingerman (approved):

PR #7603, opened 2022-01-25
Add type hints to callback, compat, and config.py

jjwang01:
Adding type hints to callback.py, compat.py, and config.py as part of https://github.com/dmlc/xgboost/issues/6496.

### DIFF
--- a/python-package/xgboost/callback.py
+++ b/python-package/xgboost/callback.py
@@ -17,7 +17,7 @@ import numpy
 from . import rabit
 from .core import Booster, DMatrix, XGBoostError, _get_booster_layer_trees
 from .compat import STRING_TYPES
-
+from .training import _PackedBooster
 
 _Score = Union[float, Tuple[float, float]]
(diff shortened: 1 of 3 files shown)


trivialfis (commented):
Thanks for the PR, could you please take a look into the failing tests?

I think adding types to callback might be tricky at the moment as we haven't exposed the packed booster to the public interface yet.

jjwang01:
Yep, looks like all the error traces are complaining about the `_PackedBooster` object. Perhaps I should set the type hints for the respective variables I've added `_PackedBooster` hints to as `Any`? I believe that only `Booster` will cause a compile time error with the `_PackedBooster` object.

trivialfis:
@jjwang01 Can we revert the changes in callback.py and keep the remaining?  I plan to improve the packed booster and expose it to the public in the future.  So right now we might just leave it as it's.

trivialfis:
Closing as these modules are fully typed.

Issue #7583, opened 2022-01-20
Number of threads used by predict.xgb.Booster (in R) cannot be specified and depends on previous runs


jiho:
In my environment (reduced to the `xgboost` relevant packages):

```
> library("xgboost")
> devtools::session_info()
─ Session info 
 setting  value
 version  R version 4.1.2 (2021-11-01)
 os       Ubuntu 20.04.3 LTS
 system   x86_64, linux-gnu
 ui       RStudio
 date     2022-01-20
 rstudio  2021.09.1+372 Ghost Orchid (server)

─ Packages 
 package     * version date (UTC) lib source
 jsonlite      1.7.3   2022-01-17 [1] CRAN (R 4.1.2)
 Matrix        1.4-0   2021-12-08 [4] CRAN (R 4.1.2)
 xgboost     * 1.5.0.2 2021-11-21 [1] CRAN (R 4.1.2)

> parallel::detectCores()
[1] 72
```

I see no way of setting `nthread` for `predict()`. I was seeing weird changes in timing of prediction and investigated with the following minimal example:

```R
# make up fake data
set.seed(1)

n <- 10^3
train <- data.frame(x1=rnorm(n), x2=rnorm(n))
train$y <- 1*train$x1^2 + 2*train$x2^2 + runif(n)
train <- as.matrix(train)

n <- 5*10^5
test <- data.frame(x1=rnorm(n), x2=rnorm(n))
test <- as.matrix(test)

save(train, test, file="data.Rdata")
```

Then run each section below in a **new R session** to make sure the run is not influenced by past state. This is important.

### Train single threaded -> predict single threaded

```R
library("xgboost")
load("data.Rdata")
m1 <- xgb.train(
  params=list(eta=0.05, max_depth=4, nthread=1),
  data=xgb.DMatrix(train[,1:2], label=train[,3]),
  nrounds=1000, verbose=FALSE)
system.time(predict(m1, newdata=test))
system.time(predict(m1, newdata=test))
```
which gives
```
> system.time(predict(m1, newdata=test))
   user  system elapsed 
  6.988   0.018   5.139 
> system.time(predict(m1, newdata=test))
   user  system elapsed 
  7.199   0.015   5.029 
```
and overall 100% CPU usage (peaking at ~200% but I suppose this is the overhead of switching between R and the C backend).

### Train multi-threaded -> predict multi-threaded

```R
load("data.Rdata")
library("xgboost")
m5 <- xgb.train(
[...]

Issue #7308, opened 2021-10-11
[RFC] Unify device configuration.


trivialfis:
This is a continuation of https://github.com/dmlc/xgboost/issues/4600

# Overview
## Use global configuration
From my perspective, this method is cleaner and covers both `DMatrix` and `Booster` so it's listed first.  An easier-to-implement solution is described in the next section.

Define a new `device` parameter for XGBoost as global configuration and remove existing parameters including `gpu_hist`, `gpu_id`, and `predictor`.  For the native Python interface, it will look like this:

``` python
with xgboost.config_context(device="CUDA:0"):
    Xy = xgb.DMatrix(X, y)
    xgboost.train({"tree_method": "hist"}, Xy)
    xgboost.predict(Xy)
```

The above code snippet should run on the first CUDA device, using GPU implementation of `hist` tree method. Also, the prediction should run on the same device regardless of the location of input data. The scikit-learn interface will look like this:

``` python
clf = xgb.XGBClassifier(device="CUDA:0", tree_method="hist")
```

while the config context is created internally in each function of `XGBClassifier`.  For R users, we also have the `xgb.set_config` function that changes global parameters.

JVM packages are lagging behind. But in theory, we can have something similar.  For the Java binding, we can define functions that are similar to R or Python `xgb.set_config` to set the global parameter.  For Scala binding, we have high-level estimators like `XGBClassifier` in Python so we can handle the configuration internally.

Last but not least, the C interface is the basis of all other interfaces, so its implementation should be trivial.

For handling existing code, my suggestion would be simply to throw an informative error.  For example, if the user has specified `gpu_hist`, then we require `device` also to be set.

## Alternative solution

This might be more practical in short term.  The `device` parameter doesn't have to be a global parameter.  Like the currently available `gpu_id`, which is a parameter for the `booster` object.  Hence we can keep it that way and reuse the `gpu_id` parameter.  This is still a breaking change due to other removed parameter but require lesser changes.  For the native Python interface, it will look like this:

``` python
Xy = xgb.DMatrix(X, y)
booster = xgboost.train({"tree_method": "hist", "gpu_id": "CUDA:0"}, Xy)
# or
booster = xgboost.train({"tree_method": "hist", "gpu_id": "0"}, Xy)
# for compatibility reason.

booster.set_param({"gpu_id": "CPU"})
# Use CPU for prediction.
booster.predict(Xy)
```

# Motivation

[...]

PR #7608, opened 2022-01-26
Remove `omp_get_max_threads`

trivialfis:
This is the one last PR for removing omp global variable.

* Add context object to the `DMatrix`.  This bridges `DMatrix` with https://github.com/dmlc/xgboost/issues/7308 .
* Require context to be available at the construction time of booster.
* Add `n_threads` support for R csc DMatrix constructor.
* Remove `omp_get_max_threads` in R glue code.
* Remove threading utilities that rely on omp global variable.

Close https://github.com/dmlc/xgboost/issues/7583 .

### DIFF
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -331,7 +331,6 @@ class LearnerConfiguration : public Learner {
     generic_parameters_.UpdateAllowUnknown(args);
 
     ConsoleLogger::Configure(args);
-    common::OmpSetNumThreads(&generic_parameters_.nthread);
 
     // add additional parameters
     // These are cosntraints that need to be satisfied.
--- a/src/learner.cc
+++ b/src/learner.cc
@@ -521,9 +521,7 @@ class LearnerConfiguration : public Learner {
     return cfg_;
   }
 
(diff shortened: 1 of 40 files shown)


src/learner.cc
```diff
@@ -331,7 +331,6 @@ class LearnerConfiguration : public Learner {
     generic_parameters_.UpdateAllowUnknown(args);
 
     ConsoleLogger::Configure(args);
-    common::OmpSetNumThreads(&generic_parameters_.nthread);
```
trivialfis, line 334: From now on XGBoost will no longer mutate the omp global variable.

trivialfis:
Tried to eliminate omp parallel clause, but apple clang is not quite happy about the use of lambda function.

hcho3 (commented):
Overall LGTM. Who owns the `ctx` object? I hope we don't end up with a dangling object, since many components take a pointer to `ctx` object that's owned by someone else.

trivialfis:
@hcho3 Both the learner and DMatrix own a ctx. They are the first to be initialized and the last to be destroyed.

Issue #7578, opened 2022-01-19
[FEA] Make XGBoost4j-Spark to support PySpark


wbo4958:
For now, Both XGBoost4j-Spark and XGBoost4j-Spark-GPU don't support XGBoost run on PySpark env, which may cause users hard to use XGBoost4j since they may know nothing about scala language.

There were two PRs about how to make XGBoost support PySpark. https://github.com/dmlc/xgboost/pull/4656, and https://github.com/dmlc/xgboost/pull/5658/.  I personally prefer the #4656 which composes the XGBoost4j-Spark python wrapper into the XGBoost4j-Spark.jar, so, in this case, the users do not need to add extra jars when submitting the XGBoost applications.

@thesuperzapper Could you continue to finish your previous PR?

candalfigomoro:
> I personally prefer the #4656 which composes the XGBoost4j-Spark python wrapper into the XGBoost4j-Spark.jar, so, in this case, the users do not need to add extra jars when submitting the XGBoost applications.

What about including it within the xgboost python package? Obviously it should gracefully fail when the xgboost4j-spark JAR is unavailable.

> @thesuperzapper Could you continue to finish your previous PR?

From the PR code:

`## EXCLUDED: trackerConf=None,`

Could you add a way to set the Scala tracker instead of the Python one?

In Java/Scala I set:

`xgbClassifier.set("trackerConf", new TrackerConf(0, "scala"));`

trivialfis:
My personal preference is we can get a design doc first.  I spent the last 2 days learning about spark and was able to build a Python interface using the new arrow serialization format in pyspark (`df.mapInPandas`) along with support for normal rdd (much slower).  So there are 2 possible paths forward, we can build it on top of existing jvm package, or we can start with the Python package.  I help maintain the Python package so I'm biased.  More inputs are welcomed.

candalfigomoro:
> My personal preference is we can get a design doc first. I spent the last 2 days learning about spark and was able to build a Python interface using the new arrow serialization format in pyspark (`df.mapInPandas`) along with support for normal rdd (much slower). So there are 2 possible paths forward, we can build it on top of existing jvm package, or we can start with the Python package. I help maintain the Python package so I'm biased. More inputs are welcomed.

While you can train python xgboost on a single node and perform a parellel inference with .mapInPandas (in pyspark >= 3.0) or a @pandas_udf (in pyspark >= 2.3), the problem is that the training step is not distributed.

[...]

PR #7609, opened 2022-01-27
Initialize PySpark API

wbo4958:
To fix https://github.com/dmlc/xgboost/issues/7578


trivialfis (changes requested):
Thank you for following up on this feature request.

Can we maintain it in a separated repository right now?  I think this wrapper is working in progress.

Cc @hcho3 @RAMitchell

trivialfis:
Let's have a design doc first.

PR #7611, opened 2022-01-28
use exit hook to remove temp file

MichaelChirico:
This guarantees the removal will trigger for unexpected early exits


trivialfis (approved):
Thank you for the PR.

PR #7612, opened 2022-01-28
[Doc] Add instruction to install XGBoost for Apple Silicon using Conda

hcho3:
https://github.com/conda-forge/xgboost-feedstock/pull/79 added a build for `osx-arm64`. This build is a cross-build, using Azure Pipelines to build a binary for Apple Silicon. Given the high complexity of setting up a cross-build, we won't be able to replicate it ourselves. As a result, we won't upload a M1-native Python wheel on PyPI. Instead, we'll tell Apple Silicon users to use Conda to install XGBoost.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -41,6 +41,14 @@ into permission errors.  Python pre-built binary capability for each platform:
 | Windows           | |tick|  |  |cross|             |
 +-------------------+---------+----------------------+
 
+If you are using **Apple Silicon**, please use the Conda packaging manager to install XGBoost:
+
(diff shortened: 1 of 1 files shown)


trivialfis (approved):

hcho3:
@trivialfis I just found https://github.com/pypa/cibuildwheel, which is a tool made by Python Packaging Authority to build wheel on multiple platforms, including Apple Silicon. I will be testing this with Treelite first and see how this goes.

trivialfis:
Fingers crossed. ;-)

PR #7615, opened 2022-01-28
(R) Accept CSR data for predictions

### LINKED ISSUE #7610: (R) Cannot predict on CSR data
In the R interface, it is not possible to use the predict function on sparse CSR data like in the other interfaces, which greatly reduces the usability of sparse data support as then any single-row prediction for sparse data would become `O(ncols)` rather than `O(nnz)` or `O(nnz*log(nnz))`.

The R interface uses `Matrix` for sparse matrix classes - the package matrix has classes `dgRMatrix` for CSR format and `dsparseVector` for single-row CSR data, but currently the predict function takes only CSC (`dgCMatrix`).
```r
library(xgboost)
library(Matrix)
data(mtcars)
X <- as.matrix(mtcars[, -1])
y <- mtcars[, 1]
model <- xgboost(as(X, "CsparseMatrix"), y, nrounds=5, verbose=0)
predict(model, as(X, "RsparseMatrix"))
```
```
Error in xgb.DMatrix(newdata, missing = missing) : 
  xgb.DMatrix does not support construction from S4
```

vs.
```python
from sklearn.datasets import load_boston
X, y = load_boston(return_X_y=True)
model = XGBRegressor(n_estimators=5).fit(X, y)
pred = model.predict(csr_matrix(X))
```
(which works correctly)

david-cortes:
Fixes #7610

This PR wraps the C function to create a DMatrix from CSR data in the R interface, allowing to pass prediction data in CSR format. The code is a copycat version of the existing CSC wrapper.

### DIFF
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -162,7 +162,11 @@ xgb.Booster.complete <- function(object, saveraw = TRUE) {
 #' Predicted values based on either xgboost model or model handle object.
 #'
 #' @param object Object of class \code{xgb.Booster} or \code{xgb.Booster.handle}
-#' @param newdata takes \code{matrix}, \code{dgCMatrix}, local data file or \code{xgb.DMatrix}.
+#' @param newdata takes \code{matrix}, \code{dgCMatrix}, \code{dgRMatrix}, \code{dsparseVector},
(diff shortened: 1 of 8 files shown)


trivialfis (approved):
Thank you for the feature! The PR looks good to me.

PR #7624, opened 2022-02-01
VM-Packages: Adding support for detecting musl-based Linux

snoopcheri:
This PR adds support for detecting musl-based Linux operating systems, when the NativeLibLoader tries to determine the path of the XGBoost4j shared library to load.

The solution for detecting a musl-based Linux system is based on how the the [sqlite-jdbc](https://github.com/xerial/sqlite-jdbc) project does it [here](https://github.com/xerial/sqlite-jdbc/blob/ada2b147dc858ef557a995e2f4ef25f7c65fc6a7/src/main/java/org/sqlite/util/OSInfo.java#L115-L131): it lists the memory-mapped files of a Linux system and checks whether any name contains the string "musl". If that's the case, the OS is considered a musl-based Linux system.

The path of the shared library for a regular Linux-based system for x86_64 architecture is still linux/x86_64/libxgboost4j.so. For a musl-based system this would now be linux-musl/x86_64/libxgboost4j.so.

This even works for Arm-based versions, where the path would be linux-musl/aarch64/libxgboost4j.so.

Note: This feature was initially discussed in the public forums [here](https://discuss.xgboost.ai/t/xgboost4j-support-for-alpine-linux-and-or-macos-arm64-planned/2623).

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java
@@ -21,8 +21,13 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Locale;
+import java.util.stream.Stream;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
@@ -34,6 +39,8 @@ import org.apache.commons.logging.LogFactory;
 class NativeLibLoader {
   private static final Log logger = LogFactory.getLog(NativeLibLoader.class);
 
+  private static Path mappedFilesBaseDir = Paths.get("/proc/self/map_files");
+
   /**
    * Supported OS enum.
    */
@@ -41,12 +48,18 @@ class NativeLibLoader {
     WINDOWS("windows"),
     MACOS("macos"),
     LINUX("linux"),
+    LINUX_MUSL("linux-musl"),
     SOLARIS("solaris");
 
     final String name;
 
     OS(String name) {
       this.name = name;
+    }
+
+    @VisibleForTesting
+    static void setMappedFilesBaseDir(Path baseDir) {
+      mappedFilesBaseDir = baseDir;
     }
 
     /**
@@ -61,13 +74,39 @@ class NativeLibLoader {
       } else if (os.contains("win")) {
         return WINDOWS;
       } else if (os.contains("nux")) {
-        return LINUX;
+        return isMuslBased() ? LINUX_MUSL : LINUX;
       } else if (os.contains("sunos")) {
         return SOLARIS;
       } else {
         throw new IllegalStateException("Unsupported OS:" + os);
       }
     }
+
+    /**
+     * Checks if the Linux OS is musl based. For this, we check the memory-mapped
+     * files and see if one of those contains the string "musl".
+     *
+     * @return true if the Linux OS is musl based, false otherwise.
+     */
+    static boolean isMuslBased() {
+      try (Stream<Path> dirStream = Files.list(mappedFilesBaseDir)) {
+        return dirStream
+          .map(OS::toRealPath)
+          .anyMatch(s -> s.toLowerCase().contains("musl"));
+      } catch (IOException ignored) {
+        // ignored
+      }
+      return false;
+    }
+
+    private static String toRealPath(Path path) {
+      try {
+        return path.toRealPath().toString();
+      } catch (IOException e) {
+        return "";
+      }
+    }
+
   }
 
   /**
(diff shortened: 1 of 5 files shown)


jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java
```diff
@@ -61,13 +74,39 @@ static OS detectOS() {
...
       } else if (os.contains("sunos")) {
         return SOLARIS;
       } else {
         throw new IllegalStateException("Unsupported OS:" + os);
       }
     }
+
+    /**
+     * Checks if the Linux OS is musl based. For this, we check the memory-mapped
+     * files and see if one of those contains the string "musl".
```
Craigacp, line 87: Could you update this so it says it checks the filenames, rather than the file contents. It's a little ambiguous.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java
```diff
@@ -34,21 +39,29 @@
...
     SOLARIS("solaris");
 
     final String name;
 
-    private OS(String name) {
+    OS(String name) {
       this.name = name;
     }
 
+    @VisibleForTesting
```
Craigacp, line 60: Is this annotation necessary? It is package private, so it should be visible for the test code already. There aren't any other usages of guava in the xgboost4j library so this now makes it a runtime dependency whereas before it was optional.

jvm-packages/xgboost4j/src/test/java/ml/dmlc/xgboost4j/java/ArchDetectionTest.java
```diff
@@ -0,0 +1,85 @@
...
+        {"amd64", X86_64},
+        {"aarch64", AARCH64},
+        {"arm64", AARCH64},
+        {"sparc64", SPARC}
+      });
+    }
+
+    @Test
+    public void testArch() throws Exception {
+      restoreSystemProperties(() -> {
```
Craigacp, line 67: Is it necessary to add a dependency for the 5 line method this saves?

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java
snoopcheri: Fair point.

CLARIFIED (c3cc638f)

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java
snoopcheri: I've added it as an indication for the reader why this method has relaxed visibility - or in this case even, why it even exists (it's only used by the test). It's not required to execute test though. I can remove it.

REMOVED (303bfcda)

jvm-packages/xgboost4j/src/test/java/ml/dmlc/xgboost4j/java/ArchDetectionTest.java
snoopcheri: I guess those things are always debatable. I've used it since it's just a test dependency and lightweight. But it can easily be replaced by custom code. So I did that.

CHANGED (cf0e7dc2)

snoopcheri:
@Craigacp I've integrated your feedback. Please have another look. Thanks! :)

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java
Craigacp: I understand the desire to document it, but in this case it's probably better to just leave a comment rather than pull in guava.

Craigacp:
Looks good. Could you update the error messages around here too - https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java#L132? I'm not sure what the situation is wrt musl and OpenMP, but if you get an `UnsatisfiedLinkError` at that line then it's not likely to be an OpenMP issue so the error message is probably wrong. 

Also it might be nice to log the detected platform at level fine or lower as now it's computed rather than being a direct function of the JVM properties (as if I've mmapped a library where the filename has "musl" in then I'll get an extremely impenetrable error message out of XGBoost if I'm actually using glibc, and logging the detected platform/OS will make it more obvious whats going on).

snoopcheri:
@Craigacp 

Sorry for the delay! But real life and stuff.. :)

I've added the following things:

* 6afc19fd adds a debug log which prints the detected filename that lead the system to believe it's musl-based
* 2e137888 is my attempt to give better hints to the end user, in case the shared library cannot be loaded. For Linux-based systems, it now also mentions that the musl-detection might have gone wrong. I'm not really sure what OpenMP AND musl really means. So I didn't go into details here.

Lemme know what you think!

Craigacp (approved):
LGTM. @trivialfis @CodingCat could one of you please review & merge this? It'll make it easier to deploy XGBoost4J in docker containers which use musl. We've hit this issue before in Tribuo and have a [note in our docs](https://github.com/oracle/tribuo/blob/main/Common/XGBoost/src/main/java/org/tribuo/common/xgboost/XGBoostTrainer.java#L65) that XGBoost4J doesn't work on top of musl libc, so it'll be nice to have a real solution available (even if users have to compile it themselves).

trivialfis (approved):
Thank you for the fix!  LGTM.

PR #7370, opened 2021-10-27
FairXGBoost: Fairness-aware Classification in XGBoost

s-ravichandran:
Implementation of fairness-aware classification based on [our paper](https://arxiv.org/abs/2009.01442)

Discussion at https://github.com/dmlc/xgboost/issues/7282

The regularizer proposed in the paper aims at achieving demographic parity (by reducing disparate impact) for minority groups. For a detailed overview of what demographic parity means, please refer to [this wonderful blog post](http://research.google.com/bigpicture/attacking-discrimination-in-ml/) 

Contributions:

* Added a mechanism to store the group identifier (sensitive feature) in memory by adding a MetaInfo field
* Support for reading the sensitive feature in both DMatrix and DaskDMatrix
* Added dataset and demo for [ProPublica COMPAS recidivism analysis](https://github.com/propublica/compas-analysis)

PR #7634, opened 2022-02-06
[dask] Cleanup dask module.

trivialfis:
* Add a new utility for mapping function onto workers.
* Use Python dict to map the partitions instead of relying on the order of a list.
* Better type hints.
* Unify the type for feature names.
* Cleanup the iterator.

- TODOs:
- [x] Restore support for DMatrix ID.


hcho3 (approved):
LGTM. Will this PR help #7370?

trivialfis:
Yes, now adding meta info is easier as they are managed by a dictionary, we don't need to worry about referencing the wrong partition.

PR #7636, opened 2022-02-08
Add cran maintainer.

trivialfis:
Sorry for the long delay. I will submit the CRAN package once the maintainer field is changed.

I read the cran policy:

> When a new maintainer wishes to take over a package, this should be accompanied by the written agreement of the previous maintainer (unless the package has been formally orphaned). 

@hetong007 Would you like to continue the submission (which would be great) or would you like to share the maintenance (I'm happy to do so but might not be the best person)?

### DIFF
--- a/R-package/DESCRIPTION
+++ b/R-package/DESCRIPTION
@@ -29,6 +29,7 @@ Authors@R: c(
   person("XGBoost contributors", role = c("cph"),
          comment = "base XGBoost implementation")
   )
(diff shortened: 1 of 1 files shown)


hetong007:
In recent years, I haven't been a user of R or contributing code to an R package, thus I may not be the right person to manage the R package for XGBoost. 

I'm 100% supportive if you take over the CRAN release duty, and I'll definitely be available for any issue in the process.

trivialfis:
Thank you for the support!  I will go ahead and merge the change.

PR #7638, opened 2022-02-08
Prepare external memory support for hist.

trivialfis:
Extracted from https://github.com/dmlc/xgboost/pull/7531 .

This PR prepares the `GHistIndexMatrix` to host the column matrix which is used by the hist tree method by accepting `sparse_threshold` parameter.

Some cleanups are made to ensure the correct batch param is being passed into `DMatrix` along with some additional tests for correctness of `SimpleDMatrix`.

### DIFF
--- a/src/data/gradient_index.h
+++ b/src/data/gradient_index.h
@@ -4,12 +4,14 @@
  */
 #ifndef XGBOOST_DATA_GRADIENT_INDEX_H_
 #define XGBOOST_DATA_GRADIENT_INDEX_H_
+#include <memory>
 #include <vector>
-#include "xgboost/base.h"
-#include "xgboost/data.h"
+
 #include "../common/categorical.h"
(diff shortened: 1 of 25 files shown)


src/data/gradient_index.h
```diff
@@ -117,7 +123,19 @@ class GHistIndexMatrix {
  */
 inline bool RegenGHist(BatchParam old, BatchParam p) {
   // parameter is renewed or caller requests a regen
-  return p.regen || (old.gpu_id != p.gpu_id || old.max_bin != p.max_bin);
+  if (!(p != BatchParam{})) {
```
RAMitchell, line 126: p==BatchParam{} ?

RAMitchell (approved):

src/data/gradient_index.h
trivialfis: Added an additional operator.

Issue #7653, opened 2022-02-14
Cgroups awareness for XGBoost when using n_jobs=-1


daschnerm:
The performance of XGBoost can be significantly affected when running with default parameters in a container whose CPU is limited using cgroups.

Projects such as https://mybinder.org/ and other container-based projects, use cgroups to limit resource usage. This leads to a large performance impact, when default parameters (n_jobs=-1) are used: XGBoost reads the host CPU count using OpenMP, which leads to over-subscription and causes containers to be throttled immediately.  In a small test using XGBClassifier with the Wisconsin breast cancer dataset (569 rows and 32 columns) and a container limited using cgroups to 1 CPU core, the executing time is about 90x slower compared to specifying ``n_jobs = container_cpus``

| n_jobs | Execution time | Container CPUs | Host CPUs |
|--------|----------------|----------------|-----------|
| -1     | 28s            | 1              | 16        |
| 1      | 0.3s           | 1              | 16        |

Looking at the number of threads spawned during the test, one can see, that indeed the host CPUs are used for determining the number of threads.

Therefore, I suggest, that:
- XGboost should respect limits imposed by cgroups, when using default parameters / specifying n_jobs=-1
- The limits may be overwritten by ``OMP_THREAD_LIMIT``, if the variable is set

[sklearn](https://scikit-learn.org/stable/whats_new/v1.0.html) for example, respect the limits imposed by cgroups and can therefore correctly determine the number of suitable cores:

>[...] take cgroups quotas into account when deciding the number of threads used by OpenMP. This avoids performance problems caused by over-subscription when using those classes in a docker container for instance 

An implemented could look similar to the one of [Sklearn](https://github.com/scikit-learn/scikit-learn/blob/0de3b0d1eaacee9f7b15cabc05752cba945c7621/sklearn/utils/_openmp_helpers.pyx#L50):

- min(openmp.omp_get_max_threads(), cpu_count())
- , where cpu_count = min(physical_cpu_cout(), container_cpu_count())

container_cpu_count could be implemented similar to [joblib](https://github.com/joblib/joblib/blob/754433f617793bc950be40cfaa265a32aed11d7d/joblib/externals/loky/backend/context.py#L195), by reading  `/sys/fs/cgroup/cpu/cpu.cfs_quota_us`  and `/sys/fs/cgroup/cpu/cpu.cfs_period_us` to calculate the usable CPU cores XGBoost using ``int(math.ceil(cfs_quota_us / cfs_period_us))``.

[...]

PR #7654, opened 2022-02-14
Honor CPU counts from CFS.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/7653 .

Need some testing.

### DIFF
--- a/src/common/threading_utils.cc
+++ b/src/common/threading_utils.cc
@@ -32,7 +32,11 @@ int32_t GetCfsCPUCount() noexcept {
     }
     char value[64];
     CHECK(::read(fd, value, sizeof(value)) < signed(sizeof(value)));
-    return std::atoi(value);
+    try {
+      return std::stoi(value);
+    } catch (std::exception const&) {
+      return -1;
+    }
   };
   // complete fair scheduler from Linux
   auto const cfs_quota(read_int("/sys/fs/cgroup/cpu/cpu.cfs_quota_us"));
--- a/src/common/threading_utils.cc
+++ b/src/common/threading_utils.cc
@@ -11,7 +11,7 @@
 namespace xgboost {
 namespace common {
 /**
- * \brief Get thread limit from docker container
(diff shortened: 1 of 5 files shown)


trivialfis:
I verified that the PR can correctly detect the docker argument `--cpus`.

trivialfis:
@hcho3 Could you please take a look?  Your expertise regarding container env would be really helpful.

src/common/threading_utils.cc
```diff
@@ -0,0 +1,47 @@
...
+  // swap limt /sys/fs/cgroup/memory.memsw.limit_in_bytes
+
+  auto read_int = [](char const* const file_path) noexcept {
+    auto const fd(::open(file_path, O_RDONLY, 0));
+    if (fd == -1) {
+      return -1;
+    }
+    char value[64];
+    CHECK(::read(fd, value, sizeof(value)) < signed(sizeof(value)));
+    return std::atoi(value);
```
hcho3, line 35: Can we use `std::stoi` here instead? It will throw an exception upon encountering an invalid value. Is this line part of a performance-critical path?

`std::atoi` may have an [undefined behavior](https://en.cppreference.com/w/cpp/string/byte/atoi):
> If no conversion can be performed, ​0​ is returned.
> If the value of the result cannot be represented, i.e. the converted value falls out of range of the corresponding return type, behavior is undefined.

hcho3:
```suggestion
    try {
        return std::stoi(value);
    } catch (std::exception&) {
        return -1;
    }
```

hcho3 (changes requested):

src/common/threading_utils.cc
trivialfis: Thank you for sharing the reference.  Changed to stoi.

trivialfis:
@hcho3 Could you please take another look?

hcho3 (approved):

PR #7670, opened 2022-02-17
[R] Fix broken links.

trivialfis:
* Regenerate documents.


trivialfis:
Added myself as one of the authors to remove the cran check note.

PR #7673, opened 2022-02-18
Work around nvcc error.

trivialfis:
This fixes the failing build on master branch.

* Remove kernel check.

### DIFF
--- a/include/xgboost/span.h
+++ b/include/xgboost/span.h
@@ -81,24 +81,25 @@ namespace common {
 #if defined(_MSC_VER)
 
 // Windows CUDA doesn't have __assert_fail.
-#define KERNEL_CHECK(cond)                                                     \
-  do {                                                                         \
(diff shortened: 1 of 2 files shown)


hcho3:
Is this strictly a workaround for a bug in nvcc?

trivialfis:
It complains a CUDA kernel (row partitioner) is using an expected number of captures.  Removing the kernel check works.  Specifying all the captures all works.  As a result, I brought a `CUDA_KERNEL_CHECK` that can be used to see what's happening during compilation (replace the `KERNEL_CHECK` with `CUDA_KERNEL_CHECK` will result in a compilation error, which should be a bug in nvcc).  Also, I brought the second workaround to that specific kernel.

 https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/master/1210/

hcho3 (approved):

Issue #7669, opened 2022-02-17
`xgb.plot.tree` stops with "Non-tree model detected"


Kodiologist:
Consider this R script:

```
library(xgboost)

m = xgboost::xgboost(
    data = as.matrix(data.frame(x = c(0, 1))),
    label = c(1, 2),
    nrounds = 1)
print(xgb.plot.tree(model = m))
```

The result is:

```
Error in xgb.model.dt.tree(feature_names = feature_names, model = model,  : 
  Non-tree model detected! This function can only be used with tree models.
```

But if you set `nrounds` to `2`, it seems to work without complaint.

Possibly related: #2876, [a Stack Overflow question](https://stackoverflow.com/questions/42670033/r-getting-non-tree-model-detected-this-function-can-only-be-used-with-tree-mo).

trivialfis:
Thank you for raising the issue.  The parsing code for the tree dump doesn't work with a tree that has only the root node.  Will try to fix it.

PR #7689, opened 2022-02-21
[R] Fix parsing decision stump.

trivialfis:
* Fix in dataframe construction.
* Fix in plotting.

Still need tests for more sophisticated cases like mixed tree types.

Close https://github.com/dmlc/xgboost/issues/7669

### DIFF
--- a/R-package/R/xgb.model.dt.tree.R
+++ b/R-package/R/xgb.model.dt.tree.R
@@ -134,7 +134,10 @@ xgb.model.dt.tree <- function(feature_names = NULL, model = NULL, text = NULL,
   ]
 
   # assign feature_names when available
-  if (!is.null(feature_names) && !is.na(td$Feature)) {
+  is_stump <- function() {
+    return(length(td$Feature) == 1 && is.na(td$Feature))
+  }
+  if (!is.null(feature_names) && !is_stump()) {
     if (length(feature_names) <= max(as.numeric(td$Feature), na.rm = TRUE))
       stop("feature_names has less elements than there are features used in the model")
     td[isLeaf == FALSE, Feature := feature_names[as.numeric(Feature) + 1]]
--- a/R-package/R/xgb.model.dt.tree.R
+++ b/R-package/R/xgb.model.dt.tree.R
@@ -125,7 +125,7 @@ xgb.model.dt.tree <- function(feature_names = NULL, model = NULL, text = NULL,
       xtr[, 3:5] <- add.tree.id(xtr[, 3:5], Tree)
       if (length(xtr) == 0) {
         as.data.table(
-          list(Feature = NA, Split = NA, Yes = NA, No = NA, Missing = NA, Quality = NA, Cover = NA)
+          list(Feature = "NA", Split = "NA", Yes = "NA", No = "NA", Missing = "NA", Quality = "NA", Cover = "NA")
         )
       } else {
         as.data.table(xtr)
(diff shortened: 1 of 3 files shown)


trivialfis:
@hetong007 Could you please take a look into this fix?  I'm a little bit confused by the regex parsing.

hetong007:
Sure. What exactly is the confusing part?

trivialfis:
@hetong007 For instance this comment in the code:

> # skip some indices with spurious capture groups from anynumber_regex

I'm not entirely sure what's it skipping. In general, I would have more confidence about the correctness of the code if we use the JSON dump along with jsonlite for parsing.  But right now in the R API, the user can directly pass a text string input so I can't make the change otherwise it would be breaking.

trivialfis:
Would be great if you can review the PR and see if the fix makes sense.

hetong007:
I looked into the code with an example. 

The to-parse text is `"f0<0.5] yes=1,no=2,missing=1,gain=0.0366666317,cover=2"`, with `branch_rx="f(\\d+)<([-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?)\\] yes=(\\d+),no=(\\d+),missing=(\\d+),gain=([-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?),cover=([-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?)"`. 

After the `regmatches` call, we have the results as

```
[[1]]
 [1] "f0<0.5] yes=1,no=2,missing=1,gain=0.0366666317,cover=2"
 [2] "0"
 [3] "0.5"
 [4] ""
 [5] "1"
 [6] "2"
 [7] "1"
 [8] "0.0366666317"
 [9] ""
[10] "2"
[11] ""
```

So basically the skipped indices contains either the input or an empty string, thus the list `c(2, 3, 5, 6, 7, 8, 10)` just slices those informative ones out.

Overall I think your fix here makes sense, with one inline comment.

R-package/R/xgb.model.dt.tree.R
```diff
@@ -116,16 +116,28 @@ xgb.model.dt.tree <- function(feature_names = NULL, model = NULL, text = NULL,
...
+  td[
+    isLeaf == FALSE,
+    (branch_cols) := {
+      matches <- regmatches(t, regexec(branch_rx, t))
+      # skip some indices with spurious capture groups from anynumber_regex
+      xtr <- do.call(rbind, matches)[, c(2, 3, 5, 6, 7, 8, 10), drop = FALSE]
+      xtr[, 3:5] <- add.tree.id(xtr[, 3:5], Tree)
+      if (length(xtr) == 0) {
+        as.data.table(
+          list(Feature = "NA", Split = "NA", Yes = "NA", No = "NA", Missing = "NA", Quality = "NA", Cover = "NA")
```
hetong007, line 128: Maybe we can set `Feature = NA, ...` instead of `"NA"`, as `NA` is a constant keyword, while `"NA"` is a normal string.

hetong007 (approved):

trivialfis:
@hetong007 Thank you for digging into this.  However, I reverted the change as by the end of the function, there are procedures to convert each column to the correct type using `as.numeric` and `as.integer`.  We need to use string value `NA` first, as the parsed values are strings and need to be consistent with the columns initialized by strings `NA`.

hetong007:
> @hetong007 Thank you for digging into this! Your explanation of the code is helpful.
> 
> On the issue of using constant instead of string value `NA`, I reverted the change as by the end of the function, there are procedures to convert each column to the correct type using `as.numeric` and `as.integer`. We need to use string value `NA` first, as the parsed values are strings and need to be consistent with the columns initialized by strings `NA`.

Cool! This makes sense.

Issue #4372, opened 2019-04-15
Forbid negative Hessian values


hcho3:
Currently, each tree updater tests if each data point has negative Hessian value and ignores the data point with negative Hessian. This complicates the updater code. We should have extra validation steps so as to ensure that all Hessian values are non-negative. The updater code then can be simplified by removing the Hessian check.

hcho3:
Would regularization cause data points to be ignored?

trivialfis:
@hcho3 Good point.

PR #7659, opened 2022-02-15
Update documents and tests.

trivialfis:
* Revise documents after recent refactoring and cat support.
* Add tests for behavior of max_depth and max_leaves.

PR #7695, opened 2022-02-23
Support categorical data for hist.

trivialfis:
* Extract partitioner from hist.
* Implement categorical data support by passing the gradient index directly into the partitioner.
* Organize/update document.
* Remove code for negative hessian.

Extracted from https://github.com/dmlc/xgboost/pull/7659 .

Close https://github.com/dmlc/xgboost/issues/4372.


RAMitchell (approved):
Looks good to me, needs further review from a CPU performance perspective.

Issue #6759, opened 2021-03-17
non-linear memory usage for wide data


arnocandel:
```
from xgboost import XGBRegressor
import numpy as np
import time

for cols in [500000, 600000]:
    shape = (1000, cols)
    print("shape: %s" % str(shape))

    X = np.random.rand(*shape)
    y = np.random.rand(shape[0])

    # GPU
    t = time.time()
    try:
        model = XGBRegressor(tree_method='gpu_hist', n_estimators=1, max_bin=64)
        model.fit(X, y)
        print("GPU fit took: %s" % (time.time() - t))
    except:
        print("OOM")
```

on RTX 2080Ti 11GB:
```
shape: (1000, 500000)
GPU fit took: 11.824267148971558
shape: (1000, 600000)
OOM
```

on RTX 3090 24GB:
```
shape: (1000, 500000)
GPU fit took: 8.36401629447937
shape: (1000, 600000)
OOM
```

One would expect that 20% more data would fit into more than 2x memory.

Also, sometimes RTX 3090 takes up to 60s for same operation, idle system.
```
arno@rippa:/data$ python xgb-wide.py 
shape: (1000, 500000)
GPU fit took: 60.621460914611816
shape: (1000, 600000)
OOM
arno@rippa:/data$ python xgb-wide.py 
shape: (1000, 500000)
GPU fit took: 12.049131631851196
shape: (1000, 600000)
OOM
arno@rippa:/data$ python xgb-wide.py 
shape: (1000, 500000)
GPU fit took: 7.468593120574951
shape: (1000, 600000)
OOM
```

tested with latest nightly:
https://s3-us-west-2.amazonaws.com/xgboost-nightly-builds/xgboost-1.4.0_SNAPSHOT%2B4f75f514ce48936e654c6baddccb7fb864e16af2-py3-none-manylinux2010_x86_64.whl

trivialfis:
The internal data representation of xgboost is a CSR array, so the memory usage would be high for wide datasets.  We might be able to ingest CPU data directly (bypassing the CSR array) for training in the future and do more optimization for this kind of dataset.

arnocandel:
thanks - any suggestions how to avoid the non-linearity (500k cols fits into 11GB card, 600k cols doesn't fit into 24GB card)?

trivialfis:
I will run some tests tomorrow.

trivialfis:
[...]

PR #7698, opened 2022-02-24
Use argsort for GPU sketching.

trivialfis:
* Sort `uint32_t` index instead of `xgboost::Entry`, which is half of the size.
* Replace the memory tracking code with something simpler.

I think this is the most memory-efficient way to perform the sort, unless we go down even further to `uint16_t`.

For dense cupy array using `DeviceQuantileDMatrix`, this can save up to 33% GPU memory for f64 and 40% memory for f32 inputs.

Close https://github.com/dmlc/xgboost/issues/6759 .

### DIFF
--- a/src/common/hist_util.cu
+++ b/src/common/hist_util.cu
@@ -51,64 +51,26 @@ size_t RequiredSampleCuts(bst_row_t num_rows, bst_feature_t num_columns,
   return result;
 }
 
-size_t RequiredMemory(bst_row_t num_rows, bst_feature_t num_columns, size_t nnz,
-                      size_t num_bins, bool with_weights) {
-  size_t peak = 0;
-  // 0. Allocate cut pointer in quantile container by increasing: n_columns + 1
-  size_t total = (num_columns + 1) * sizeof(SketchContainer::OffsetT);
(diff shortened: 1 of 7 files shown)


trivialfis:
However, the performance also drops significantly.

trivialfis:
Converting to draft as the performance of `thrust::sort` decreases significantly due to increased register use.

hcho3:
@trivialfis How bad is the performance drop? Do you have a benchmark?

trivialfis:
@hcho3 Using 1e6 rows, with 320 columns and 32 batches with `DeviceQuantileDMatrix`, the initialization time is increased from 5.2566 seconds to 15.3669 seconds.  My test is on an old GTX1080TI.  On RTX8000 should not be a problem.  But still, that's a 3x increase.

trivialfis:
Register per-thread for the sort is increased from 48 to 128.

trivialfis:
Try to have something similar in https://github.com/dmlc/xgboost/pull/7705 .  But keep getting memory error inside merge sort. Investigating.

trivialfis:
https://github.com/NVIDIA/thrust/issues/1622

trivialfis:
Closing for now.

PR #7702, opened 2022-02-24
Cache binary search result for GPU cat evaluation.

trivialfis:
This slightly reduces the time used for the thrust stable sort.  Using the cat_in_the_dat example in demo, the time is reduced from about 4.2 seconds to 3 seconds (32 rounds only).  It's a small dataset so might not be representative.

To have good performance for GPU, I think in the future we either have to choose a different algorithm that doesn't require sorting, or we continue our path to fuse the kernels for multiple nodes.

### DIFF
--- a/src/tree/gpu_hist/evaluator.cu
+++ b/src/tree/gpu_hist/evaluator.cu
@@ -54,6 +54,18 @@ void GPUHistEvaluator<GradientSumT>::Reset(common::HistogramCuts const &cuts,
           cudaMemsetAsync(split_cats_.data().get(), '\0', split_cats_.size() * sizeof(CatST)));
 
       cat_sorted_idx_.resize(cuts.cut_values_.Size() * 2);  // evaluate 2 nodes at a time.
+
+      fidx_.resize(cat_sorted_idx_.size());
+      auto d_fidxes = dh::ToSpan(fidx_);
+      auto it = thrust::make_counting_iterator(0ul);
+      auto values = cuts.cut_values_.ConstDeviceSpan();
+      auto ptrs = cuts.cut_ptrs_.ConstDeviceSpan();
+      thrust::transform(thrust::cuda::par(alloc), it, it + fidx_.size(), fidx_.begin(),
+                        [=] XGBOOST_DEVICE(size_t i) {
+                          auto is_left = i < values.size();
+                          auto fidx = dh::SegmentId(ptrs, i);
+                          return fidx;
+                        });
     }
   }
 }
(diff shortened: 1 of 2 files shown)


trivialfis:
To give an intuition of how's expansive the sorting is, for the `cat_in_the_dat` example:

```
 Time (%)  Total Time (ns)  Instances   Avg (ns)     Med (ns)    Min (ns)   Max (ns)   StdDev (ns)                                                  Name                                                
 --------  ---------------  ---------  -----------  -----------  ---------  ---------  -----------  ----------------------------------------------------------------------------------------------------
     89.3    1,861,547,537      1,023  1,819,694.6  1,809,669.0  1,496,088  2,395,355     82,436.0  void cub::DeviceMergeSortBlockSortKernel<(bool)0, cub::DeviceMergeSortPolicy<thrust::device_ptr<uns…
      2.3       48,190,073        991     48,627.7     25,537.0      9,249    144,901     40,198.0  void dh::LaunchNKernel<xgboost::tree::GPUHistEvaluator<xgboost::detail::GradientPairInternal<double…
      1.9       40,098,547      1,023     39,197.0     23,169.0     13,888    379,343     52,766.3  void xgboost::tree::SharedMemHistKernel<xgboost::detail::GradientPairInternal<double>, (bool)1>(xgb…
```

RAMitchell:
Surely this sort step is also expensive for CPU implementations?

trivialfis:
@RAMitchell We will prepare some larger benchmarks to measure.  I might want to update the gbm-bench with categorical data.

src/tree/gpu_hist/evaluator.cu
```diff
@@ -54,6 +54,18 @@ void GPUHistEvaluator<GradientSumT>::Reset(common::HistogramCuts const &cuts,
...
       cat_sorted_idx_.resize(cuts.cut_values_.Size() * 2);  // evaluate 2 nodes at a time.
+
+      fidx_.resize(cat_sorted_idx_.size());
+      auto d_fidxes = dh::ToSpan(fidx_);
+      auto it = thrust::make_counting_iterator(0ul);
+      auto values = cuts.cut_values_.ConstDeviceSpan();
+      auto ptrs = cuts.cut_ptrs_.ConstDeviceSpan();
+      thrust::transform(thrust::cuda::par(alloc), it, it + fidx_.size(), fidx_.begin(),
+                        [=] XGBOOST_DEVICE(size_t i) {
+                          auto is_left = i < values.size();
```
RAMitchell, line 65: Unused?

src/tree/gpu_hist/evaluator.cu
```diff
@@ -64,33 +76,33 @@ common::Span<bst_feature_t const> GPUHistEvaluator<GradientSumT>::SortHistogram(
...
+      return lfidx < rfidx;  // not the same feature
+    }
+    if (common::IsCat(input.feature_types, lfidx)) {
+      auto lw = evaluator.CalcWeightCat(input.param, input.gradient_histogram[l]);
+      auto rw = evaluator.CalcWeightCat(input.param, input.gradient_histogram[r]);
+      return lw < rw;
+    }
+    return l < r;
+  };
+  thrust::stable_sort(thrust::cuda::par(alloc), dh::tbegin(sorted_idx), dh::tend(sorted_idx), comp);
```
RAMitchell, line 105: I wonder if this sort is slow because the comparator is repeatedly performing slow gather memory accesses and evaluating CalcWeightCat? This comparator is evaluated nlogn times. What if you pre-calculate these in a tuple (e.g. by using a transformer iterator as input to sort) and allow thrust to sort the tuples of precaculated values?

trivialfis: Yes, that's incredibly helpful!  Will close this PR now and submit a new one.

PR #7703, opened 2022-02-25
Remove compatibility check in hist; add version check in LoadConfig()

trivialfis:
We won't support this use as we are moving into ubjson as serialization format and phasing out the old binary model.


src/learner.cc
```diff
@@ -406,8 +407,29 @@ class LearnerConfiguration : public Learner {
...
+                 << "For more information, visit "
+                 << "https://xgboost.readthedocs.io/en/latest/tutorials/saving_model.html";
+#else  // XGBOOST_STRICT_R_MODE
+      LOG(FATAL) << "You are using XGBoost " << Version::String(Version::Self())
+                 << " and attempting to load model from XGBoost "
+                 << Version::String(origin_version) << " via pickle.load() or a similar "
+                 << "mechanism. For long-term archival of models, use the save_model() method. "
+                 << "For more information, visit "
+                 << "https://xgboost.readthedocs.io/en/latest/tutorials/saving_model.html";
+#endif  // XGBOOST_STRICT_R_MODE
```
hcho3, line 431: @trivialfis  This PR will also disallow loading from old Python pickles. Can you review the error meesage?

src/learner.cc
```diff
@@ -19,6 +19,7 @@
 #include <stack>
 #include <utility>
 #include <vector>
+#include <tuple>
```
hcho3, line 22: Unnecessary include

hcho3:
Closing in favor of #7864

PR #7705, opened 2022-02-25
Optimize GPU evaluation function for categorical data.

trivialfis:
* Use transform the pre-compute some of the values.

The time used for the `cat-in-the-dat.py` demo is reduced by about half from 4.2 seconds to about 2.1 seconds.

Suggested by @RAMitchell .


trivialfis:
Unsure whether sorting a transform iterator is supported.

RAMitchell (approved):

PR #7707, opened 2022-02-28
Add type hints to core.py

bridgream:
I attempted to add type hints to core.py. I wasn't able to eliminate all mypy errors, but I hoped to make incremental changes because adding type hints seemed very intrusive and could easily trigger conflicts. 

Related to #6496

### DIFF
--- a/python-package/xgboost/_typing.py
+++ b/python-package/xgboost/_typing.py
@@ -1,0 +1,54 @@
+import ctypes
+import os
+from typing import Optional, List, Any, TypeVar, Union, Dict
+
+import numpy as np
+
+# os.PathLike/string/numpy.array/scipy.sparse/pd.DataFrame/dt.Frame/
+# cudf.DataFrame/cupy.array/dlpack
+DataType = Any
+
+# xgboost accepts some other possible types in practice due to historical reason, which is
+# lesser tested.  For now we encourage users to pass a simple list of string.
+FeatNamesT = Optional[List[str]]
+
+ArrayLike = Any
+PathLike = Union[str, os.PathLike]
+list_or_dict = Union[List, Dict]
+cupy_t = ArrayLike  # maybe need a stub for cupy arrays
+numpy_or_cupy_t = Union[np.ndarray, cupy_t]
+
+# ctypes
+# c_bst_ulong corresponds to bst_ulong defined in xgboost/c_api.h
+c_bst_ulong = ctypes.c_uint64
+
+ctype_t = Union[
+    ctypes.c_void_p, ctypes.c_char_p, ctypes.c_int,
+    ctypes.c_float, ctypes.c_uint, ctypes.c_size_t
+]
+
+# supported numeric types
+c_numeric_t = Union[
+    ctypes.c_float, ctypes.c_double, ctypes.c_uint,
+    ctypes.c_uint64, ctypes.c_int32, ctypes.c_int64
+]
+
+# c pointer types
+# real type should be, as defined in typeshed
+# but this has to be put in a .pyi file
+# c_str_ptr_t = ctypes.pointer[ctypes.c_char]
+c_str_ptr_t = ctypes.Array
+# c_str_pptr_t = ctypes.pointer[ctypes.c_char_p]
+c_str_pptr_t = ctypes.Array
+# c_float_ptr_t = ctypes.pointer[ctypes.c_float]
+c_float_ptr_t = ctypes.Array
+
+# c_numeric_ptr_t = Union[
+#  ctypes.pointer[ctypes.c_float], ctypes.pointer[ctypes.c_double],
+#  ctypes.pointer[ctypes.c_uint], ctypes.pointer[ctypes.c_uint64],
+#  ctypes.pointer[ctypes.c_int32], ctypes.pointer[ctypes.c_int64]
+# ]
+c_numeric_ptr_t = ctypes.Array
+
+# template parameter
+T = TypeVar("T")
--- a/python-package/xgboost/_typing.py
+++ b/python-package/xgboost/_typing.py
@@ -14,21 +14,20 @@ FeatNamesT = Optional[List[str]]
 
 ArrayLike = Any
 PathLike = Union[str, os.PathLike]
-list_or_dict = Union[List, Dict]
-cupy_t = ArrayLike  # maybe need a stub for cupy arrays
-numpy_or_cupy_t = Union[np.ndarray, cupy_t]
+CupyT = ArrayLike  # maybe need a stub for cupy arrays
+NdarrayOrCupyT = Union[np.ndarray, CupyT]
 
 # ctypes
 # c_bst_ulong corresponds to bst_ulong defined in xgboost/c_api.h
 c_bst_ulong = ctypes.c_uint64
 
-ctype_t = Union[
+CTypeT = Union[
     ctypes.c_void_p, ctypes.c_char_p, ctypes.c_int,
     ctypes.c_float, ctypes.c_uint, ctypes.c_size_t
 ]
 
 # supported numeric types
-c_numeric_t = Union[
+CNumericT = Union[
     ctypes.c_float, ctypes.c_double, ctypes.c_uint,
     ctypes.c_uint64, ctypes.c_int32, ctypes.c_int64
 ]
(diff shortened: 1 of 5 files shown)


bridgream:
CI failed because I used `typing.ParamSpec`, which was [only introduced in Python 3.10](https://www.python.org/dev/peps/pep-0612/). I'll seek to fix that.

python-package/xgboost/core.py
```diff
@@ -776,7 +793,7 @@ def set_float_info_npy2d(self, field: str, data) -> None:
         from .data import dispatch_meta_backend
         dispatch_meta_backend(self, data, field, 'float')
 
-    def set_uint_info(self, field: str, data) -> None:
+    def set_uint_info(self, field: str, data: np.ndarray) -> None:
```
trivialfis, line 796: Please use array like.

python-package/xgboost/core.py
```diff
@@ -27,12 +28,18 @@
 # lesser tested.  For now we encourage users to pass a simple list of string.
 FeatNamesT = Optional[List[str]]
 
+list_or_dict = TypeVar("list_or_dict", List, Dict)
+ArrayLike = np.ndarray | Sequence
```
trivialfis, line 32: I think for now we can just use Any.  I have been working on defining the protocol of input data: https://github.com/dmlc/xgboost/pull/7242 We accept a lot more types and ndarray.

python-package/xgboost/core.py
```diff
@@ -91,7 +98,7 @@ def _convert_ntree_limit(
     return iteration_range
 
 
-def _expect(expectations, got):
+def _expect(expectations: Sequence[Any], got: Sequence[Any]) -> str:
```
trivialfis, line 101: Maybe `Sequence[Type]`?

python-package/xgboost/core.py
```diff
@@ -237,18 +254,18 @@ def _cuda_array_interface(data) -> bytes:
...
+def ctypes2numpy(cptr: ctypes.pointer, length: int, dtype: Type[np.number]) -> np.ndarray:
     """Convert a ctypes pointer array to a numpy array."""
-    ctype = _numpy2ctypes_type(dtype)
-    if not isinstance(cptr, ctypes.POINTER(ctype)):
+    ctype: Type[supported_ctypes] = _numpy2ctypes_type(dtype)
+    if not isinstance(cptr, ctypes.POINTER(ctype)):  # should we use `issubclass`?
         raise RuntimeError(f"expected {ctype} pointer")
     res = np.zeros(length, dtype=dtype)
-    if not ctypes.memmove(res.ctypes.data, cptr, length * res.strides[0]):
+    if not ctypes.memmove(res.ctypes.data, cptr, length * res.strides[0]):  # "memmove" does not return a value?
```
trivialfis, line 263: I think that's probably a bug in mypy.

python-package/xgboost/core.py
```diff
@@ -790,7 +807,7 @@ def set_uint_info(self, field: str, data) -> None:
         from .data import dispatch_meta_backend
         dispatch_meta_backend(self, data, field, 'uint32')
 
-    def save_binary(self, fname, silent=True) -> None:
+    def save_binary(self, fname: str | os.PathLike, silent: bool = True) -> None:
```
trivialfis, line 810: Consider using Union instead.  This requires `from future import __annotations__`, which has different behavior.

python-package/xgboost/core.py
```diff
@@ -806,7 +823,7 @@ def save_binary(self, fname, silent=True) -> None:
                                              c_str(fname),
                                              ctypes.c_int(silent)))
 
-    def set_label(self, label) -> None:
+    def set_label(self, label: Collection) -> None:
```
trivialfis, line 826: Not sure why is this a `Collection`?  Maybe ArrayLike?

python-package/xgboost/core.py
bridgream: Agreed; I checked the usages of `_expect` and realized `Sequence[Type]` was a better choice.

python-package/xgboost/core.py
bridgream: I'm not sure if that different behavior matters (except for the explicit import), but since `typing.Union` is available for Python <3.10, I agree it's better to stick with that.

python-package/xgboost/core.py
bridgream: My bad; I was confused by `dispatch_meta_backend` but apparently I missed the docstring...

bridgream:
The CI build still failed due to `TypeError: '_ctypes.PyCArrayType' object is not subscriptable`. Could it be because I used a platform specific annotation `Union[bytes, ctypes.Array[ctypes.c_char_p]]`?

bridgream:
I was able to eliminate most mypy errors from 80+ to 19 (hopefully that's not because I introduced too many `Any`). I think most of the rest errors would involve more intrusive changes or changes of other files.

python-package/xgboost/core.py
bridgream: You were right. By the way, the maintainers of typeshed were able to fix this bug in a flash.

bridgream:
It turns out we cannot include `ctypes.pointer[]` in any part of the Python scripts. However, to properly annotate typed C pointers, it seems we have to use this grammar. In fact, [typeshed has a class interface for](https://github.com/python/typeshed/blob/master/stdlib/ctypes/__init__.pyi#L175) `ctypes.pointer` [but this is not recognized by Python](https://github.com/python/mypy/issues/7540). 

Maybe we should separate the definitions of types from source code by putting them in a stub file?

trivialfis:
Apologies for the slow response, will come back to this tomorrow.

trivialfis:
> Maybe we should separate the definitions of types from source code by putting them in a stub file?

That sounds like a good idea, we need to have some stubs later.

bridgream:
@trivialfis Thanks for following up! I've done a quick demo using stub (basically I just used `stubgen` and filled types that weren't automatically inferred, so that I was able to make mypy satisfied) and erased previous work (I made a backup of the previous commit, though). 

Some issues: 
1. With the existence of `core.pyi`, mypy simply ignores all other files including `core.py` itself (which is expected behavior). Is this still helpful? I guess we also wanted mypy to help us do static check for implementation? 
2. Should we generate a stub file for each .py file, or just one `__init__.pyi` enumerating all APIs? 
3. I've added a `_typing.py` to collect all typing-related declaration but maybe I should follow `typing.py` as in [this draft PR](https://github.com/dmlc/xgboost/pull/7242). Could you please offer some advice?

trivialfis:
Thank you for experimenting with this! Does it make sense to simply use `ctypes.pointer` for now?  I think python typing is still evolving and we don't have to be too harsh on ourselves at the moment.  It's unlikely that we have a perfect solution for everything in short term.

trivialfis:
Hi, please let me know when it's ready for review. ;-)

bridgream:
Hi @trivialfis, after some of our work, there are still 22 mypy errors in core.py but I find them quite tricky to tackle. The errors can be generated with
```
cd python-package
mypy xgboost/core.py --ignore-missing-imports
```

Besides the errors that relate to ctypes, some errors are due to the flexible user interface. For example, the setter `feature_names` starting from core.py:1004 takes either a `str`, a `List[str]` or `None`. Even if there's code to deal with each case, mypy cannot understand the implementation but simply throws errors. 

While I don't think simply adding `# type: ignore` is a good solution, I'm not sure whether we still should solve all the errors now. Otherwise, given the improvement we've already done, I think it's a good time to review and possibly merge.

python-package/xgboost/_typing.py
```diff
@@ -0,0 +1,62 @@
...
+DataType = Any
+
+# xgboost accepts some other possible types in practice due to historical reason, which is
+# lesser tested.  For now we encourage users to pass a simple list of string.
+FeatureNames = Optional[List[str]]
+
+ArrayLike = Any
+PathLike = Union[str, os.PathLike]
+CupyT = ArrayLike  # maybe need a stub for cupy arrays
+NdarrayOrCupyT = Union[np.ndarray, CupyT]
```
trivialfis, line 19: I think we might be better off to return `Any` for prediction.  The return type is actually a dependent type, which means given the input type, the return type is known so it's not a union.  From a static type language's point of view (like c++, rust), the union can be changed at runtime, but the return type is determined at compile time.

trivialfis (commented):
Thank you for the nice work.  One comment inlined.

python-package/xgboost/_typing.py
trivialfis: mypy doesn't support dependent type yet, you can find some related feature requests in their issue list.  To provide a concrete example, if the users use a cupy array as input to predict, then they know the output is a cupy array.  However, using `Union` the users will be asked to handle both numpy array and cupy array.

bridgream: I agree. By the way, should we use string literal for cupy types, e.g. `"cupy.ndarray"` instead of `CupyT` here https://github.com/bridgream/xgboost/blob/mypy/python-package/xgboost/core.py#L276

trivialfis: @bridgream I think it will show an error as mypy couldn't find cupy on the test environment (cupy is an optional dependency)

trivialfis (approved):
Thank you for the nice work!  Will merge as part of 1.6.

bridgream:
Great, thank you! I'll continue to improve type hints in other Python scripts.

python-package/xgboost/_typing.py
trivialfis: On a second thought, overloading might help, but the change might be tedious as cupy is an optional dependency.

bridgream: Maybe we should follow your solution in [this PR](https://github.com/dmlc/xgboost/pull/7242/files#diff-9748ea5ca0c5306e9992e33ff13359e683c4b9dc62b5a291b35b2966899aca85) by creating an interface. Since xgboost only touches some basic interface, this shouldn't take too much work. We could then introduce `@overload` to achieve better hints. 
I can try that next.

Issue #7712, opened 2022-03-01
Feature shape mismatch error in xgboost 1.5.0 for linux-s390x


pradghos:
We have observed `Feature shape mismatch` error in `xgboost 1.5.0` for linux-s390x and it can be easily reproducible with below test case - 

**Failure log**:
```python 
pytest -v tests/python/test_with_sklearn.py::test_XGBClassifier_resume

tests/python/test_with_sklearn.py::test_XGBClassifier_resume FAILED                                                                                    [100%]

========================================================================== FAILURES ==========================================================================
_________________________________________________________________ test_XGBClassifier_resume __________________________________________________________________

    def test_XGBClassifier_resume():
        from sklearn.datasets import load_breast_cancer
        from sklearn.metrics import log_loss

        with TemporaryDirectory() as tempdir:
            model1_path = os.path.join(tempdir, 'test_XGBClassifier.model')
            model1_booster_path = os.path.join(tempdir, 'test_XGBClassifier.booster')

            X, Y = load_breast_cancer(return_X_y=True)

            model1 = xgb.XGBClassifier(
                learning_rate=0.3, random_state=0, n_estimators=8)
            model1.fit(X, Y)

>           pred1 = model1.predict(X)

.... 

            if not hasattr(data, "shape"):
                raise TypeError(
                    "`shape` attribute is required when `validate_features` is True."
                )
            if len(data.shape) != 1 and self.num_features() != data.shape[1]:
>               raise ValueError(
                    f"Feature shape mismatch, expected: {self.num_features()}, "
                    f"got {data.shape[1]}"
                )
E               ValueError: Feature shape mismatch, expected: 0, got 30

```

There are other test case failure also for the same issue in `xgboost 1.5` ; However above test cases worked fine with `xgboost 1.3.3` in linux-s390x.

pradghos:
`features.value` is getting truncated to 0 in `num_features()` is getting truncated in python layer. 

```python
> /opt/miniconda3/envs/test_xgb_mster/lib/python3.9/site-packages/xgboost/core.py(2039)inplace_predict()
-> if not hasattr(data, "shape"):
(Pdb) n
> /opt/miniconda3/envs/test_xgb_mster/lib/python3.9/site-packages/xgboost/core.py(2043)inplace_predict()
-> if len(data.shape) != 1 and self.num_features() != data.shape[1]:
(Pdb) step num_features()
--Call--
[...]

PR #7715, opened 2022-03-03
Fix for Feature shape mismatch error

pradghos:
- Addressing https://github.com/dmlc/xgboost/issues/7712
- Corrected integer types
- also referred current implementation of `num_row()`, `num_col()`


trivialfis (approved):
Thank you for the fix!

pradghos:
Thanks @trivialfis !

PR #7722, opened 2022-03-09
Initialize TreeUpdater ctx_ with nullptr

hmchen-github:
MemorySanitizer build complains that https://github.com/dmlc/xgboost/blob/a62a3d991d8826719e13f6e7b502b9c0409bbaac/src/tree/updater_prune.cc#L27 use-of-uninitialized-value.


trivialfis (approved):
Thank you for the fix

PR #7723, opened 2022-03-09
Run tests in a temporary directory

hmchen-github:
Fix some tests to run in a temporary directory in case the root
directory is not writable. Note that most of tests are already
running in the temporary directory, so this PR just makes them
consistent.


trivialfis (approved):
Thank you for the fix.

PR #7730, opened 2022-03-14
Correct xgboost-config directory for inclusion in other projects

wphicks:
Currently, `xgboost-config.cmake` is created in the `CMAKE_CURRENT_BINARY_DIR` directory, but we attempt to install it from the `CMAKE_BINARY_DIR` directory. For downstream projects attempting to consume xgboost from e.g. CPM, this prevents the install from going through successfully.


trivialfis (approved):
Thank you for the fix!

Issue #7729, opened 2022-03-14
[Bug]: Training with sparse data on a GPU results in bad model performance


JovanVeljanoski:
Hi, 

I believe I've stumbled on a bug when training an `xgboost.XGBClassifier` model on a GPU using sparse data. 
If I train a model on a CPU, i get a certain score (precision, f1 etc..). Using the same hyperparameters (mostly defaults), training a model on the GPU i get substantially worse performance (worse f1, worse precision etc..). 

I tracked it to be somehow related to passing `scipy.sparse._csr.csr_matrix` as a training data. I include a working example below:

```python
import xgboost
import numpy as np
from sklearn.metrics import classification_report

print(xgboost.__version__)
print('')

X = np.load('./Xtrain.npy', allow_pickle=True).item()
y = np.load('./ytrain.npy', allow_pickle=True)

booster_cpu = xgboost.XGBClassifier(learning_rate=0.1, 
                                    max_depth=7, 
                                    n_estimators=1000, 
                                    objective='binary:logistic', 
                                    tree_method='hist', 
                                    use_label_encoder=False, 
                                    eval_metric='logloss')
booster_cpu.fit(X, y)
pred_cpu = booster_cpu.predict(X)


booster_gpu = xgboost.XGBClassifier(learning_rate=0.1, 
                                    max_depth=7, 
                                    n_estimators=1000, 
                                    objective='binary:logistic', 
                                    tree_method='gpu_hist', 
                                    use_label_encoder=False, 
                                    eval_metric='logloss')
booster_gpu.fit(X, y)
pred_gpu = booster_gpu.predict(X)

print('Predictions via the CPU model')
print(classification_report(y, pred_cpu))
print(' ')
print('Predictions via the GPU model')
print(classification_report(y, pred_gpu))
```

This returns 

```
1.5.1

Predictions via the CPU model
              precision    recall  f1-score   support

           0       1.00      1.00      1.00     95269
           1       0.98      0.90      0.94      4731

    accuracy                           0.99    100000
   macro avg       0.99      0.95      0.97    100000
weighted avg       0.99      0.99      0.99    100000

 
Predictions via the GPU model
              precision    recall  f1-score   support

           0       0.97      0.73      0.83     95269
           1       0.10      0.62      0.17      4731

    accuracy                           0.72    100000
   macro avg       0.54      0.67      0.50    100000
weighted avg       0.93      0.72      0.80    100000
```

[...]

Issue #5301, opened 2020-02-10
Libsvm with unordered features breaks gpu_hist


RAMitchell:
A libsvm file with a row such as below creates undefined behaviour
`0 1:3400 0:170`
The gpu_hist algorithm assumes that rows of data are ordered by feature. It uses binary search over sparse rows to find specific matrix elements, and the binary search fails due to incorrect ordering. There may be other cases of this assumption inside xgboost that I am not aware of.

The bug is very subtle and manifests in slightly incorrect training results where the algorithm designates elements as missing because the binary search can't find them.

We need some way to detect such cases and log a fatal error, preferably across all xgboost algorithms. We can modify the dmlc parser to check for this case, although it is not strictly wrong according to any specification, or we could detect it somewhere in xgboost constructors.

thvasilo:
I think I may have seen this issue in one of the CPU algorithms (most probably approx) when I was creating libsvm files by hand to test out stuff.

The libsvm format assumes that features are order as well or is that only in XGBoost?

RAMitchell:
The libsvm format checker does seem to indicate that features should be ordered, see at the end of this readme: https://github.com/cjlin1/libsvm/tree/master/tools

I think we can do the validation directly in dmlc parser.

PR #7731, opened 2022-03-14
Sort sparse page index when constructing DMatrix.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/7729 .
Close https://github.com/dmlc/xgboost/issues/5301 .

GPU DMatrix constructor is unaffected as it doesn't support cupyx yet, all the other inputs are dense.

This might hit performance, another way to fix the issue with scipy sparse is to sort the index in Python instead of the low-level constructor, however, it won't prevent issues with other inputs.  Lastly, we can simply check for sorted data and raise an error for unsorted inputs.

### DIFF
--- a/include/xgboost/data.h
+++ b/include/xgboost/data.h
@@ -201,6 +201,9 @@ struct Entry {
   inline static bool CmpValue(const Entry& a, const Entry& b) {
     return a.fvalue < b.fvalue;
   }
+  static bool CmpIndex(Entry const& a, Entry const& b) {
+    return a.index < b.index;
+  }
   inline bool operator==(const Entry& other) const {
     return (this->index == other.index && this->fvalue == other.fvalue);
   }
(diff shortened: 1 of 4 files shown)


trivialfis:
Here is a quick benchmark with 0.5 sparsity for sparse dataset.  Both dense and sparse has shape 1e6 * 1e3.  The difference is trivial when sorting is not needed and only check is performed.

|             | Master            | PR                |
|:------------|:------------------|:------------------|
| Sparse sort | NA                | 4.370167255401611 |
| Sparse      | 3.301400899887085 | 3.301400899887085 |
| Normal      | 5.396663665771484 | 5.622706651687622 |

trivialfis:
Here is a quick benchmark with 0.5 sparsity for sparse dataset.  Both dense and sparse has shape 1e6 * 1e3.  The difference is trivial when sorting is not needed and only check is performed.

|             | Master            | PR                |
|:------------|:------------------|:------------------|
| Sparse sort | NA                | 4.370167255401611 |
| Sparse      | 3.301400899887085 | 3.301400899887085 |
| Normal      | 5.396663665771484 | 5.622706651687622 |

RAMitchell (approved):

PR #7732, opened 2022-03-16
Dev/oneapi objectives support

razdoburdin:
In continuous of #5659, #6212 and #7724.

Here I present a realization for support of oneAPI objectives and oneAPI predictor. Being applied together with the change from #7724, this change will allow using of oneAPI devices for prediction and for calculation of objectives. This request does not contain the oneAPI updater. Updater is the most complex part, so in my mind, it is better to discuss the general concept based on the simpler examples.

For building the code with the plugin support, one should do the following steps (DPCPP compiler is required):
```bash
$ mkdir build
$ cd build
$ cmake .. -DPLUGIN_UPDATER_ONEAPI=ON
$ make
```
For using the plugin, one should specify the parameter
```device_selector=oneapi:gpu``` or ```device_selector=oneapi:gpu:0```


razdoburdin:
The functionality has been already added by https://github.com/dmlc/xgboost/pull/9846

razdoburdin:
The functionality is already added by https://github.com/dmlc/xgboost/pull/9846

PR #7735, opened 2022-03-16
Small cleanup to hist tree method.

trivialfis:
* Remove special optimization using number of bins.
* Remove 1-based index for column sampling.
* Remove data layout.
* Unify update prediction cache.

### DIFF
--- a/src/common/row_set.h
+++ b/src/common/row_set.h
@@ -55,8 +55,6 @@ class RowSetCollection {
   /*! \brief return corresponding element set given the node_id */
   inline const Elem& operator[](unsigned node_id) const {
     const Elem& e = elem_of_each_node_[node_id];
-    CHECK(e.begin != nullptr)
-        << "access element that is not in the set";
     return e;
   }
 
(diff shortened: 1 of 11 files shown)


trivialfis:
The data layout is used only for root initialization. As for the logic of choosing the smallest number of bins, I replaced it by using the first feature.  Lastly, for the 1 based feature sampling, I don't think it's used in practice, or at least we will have to add proper tests and usage examples before the code is rotten.

No performance change in this PR.

src/common/row_set.h
```diff
@@ -75,14 +73,10 @@ class RowSetCollection {
     CHECK_EQ(elem_of_each_node_.size(), 0U);
 
     if (row_indices_.empty()) {  // edge case: empty instance set
-      // assign arbitrary address here, to bypass nullptr check
```
RAMitchell, line 78: Hmmmmmm...

RAMitchell (approved):
LGTM

PR #7736, opened 2022-03-17
Demo for using custom objective with multi-target regression.

trivialfis:
@StatMixedML Feel free to take a look, would love to hear more inputs about the future development of multi-target models.


RAMitchell (approved):
Nice example.

PR #7742, opened 2022-03-22
Add Type Hints for Python Package

bridgream:
Following https://github.com/dmlc/xgboost/pull/7707, I plan to resolve most mypy errors across the python package if not all. 

We should probably populate compat.py with stubs for even more packages, such as cupy.

### DIFF
--- a/python-package/xgboost/_typing.py
+++ b/python-package/xgboost/_typing.py
@@ -11,7 +11,7 @@ DataType = Any
 
 # xgboost accepts some other possible types in practice due to historical reason, which is
 # lesser tested.  For now we encourage users to pass a simple list of string.
-FeatureInfo = Optional[Sequence[str]]
+FeatureInfo = Sequence[str]
 FeatureNames = FeatureInfo
 FeatureTypes = FeatureInfo
 
--- a/python-package/xgboost/_typing.py
+++ b/python-package/xgboost/_typing.py
@@ -1,7 +1,7 @@
 """Shared typing definition."""
 import ctypes
 import os
-from typing import Optional, Any, TypeVar, Union, Type, Sequence, Callable
+from typing import Any, TypeVar, Union, Type, Sequence, Callable, List, Dict
 
 # os.PathLike/string/numpy.array/scipy.sparse/pd.DataFrame/dt.Frame/
 # cudf.DataFrame/cupy.array/dlpack
@@ -14,6 +14,7 @@ DataType = Any
 FeatureInfo = Sequence[str]
 FeatureNames = FeatureInfo
 FeatureTypes = FeatureInfo
+ParamType = Union[List, Dict]  # better be sequence but need to remove isinstance(..., list) first
 
 ArrayLike = Any
 PathLike = Union[str, os.PathLike]
@@ -22,6 +23,9 @@ NumpyOrCupy = Any
 NumpyDType = Union[str, Type[np.number]]
 
 FloatCompatible = Union[float, np.float32, np.float64]
+
+# callables
+FPreProcCallable = Callable
 
 # ctypes
 # c_bst_ulong corresponds to bst_ulong defined in xgboost/c_api.h
--- a/python-package/xgboost/_typing.py
+++ b/python-package/xgboost/_typing.py
@@ -11,8 +11,9 @@ DataType = Any
 
 # xgboost accepts some other possible types in practice due to historical reason, which is
 # lesser tested.  For now we encourage users to pass a simple list of string.
-FeatureNames = Optional[Sequence[str]]
-FeatureTypes = Optional[Sequence[str]]
+FeatureInfo = Optional[Sequence[str]]
+FeatureNames = FeatureInfo
+FeatureTypes = FeatureInfo
 
 ArrayLike = Any
 PathLike = Union[str, os.PathLike]
--- a/python-package/xgboost/_typing.py
+++ b/python-package/xgboost/_typing.py
@@ -1,7 +1,7 @@
 """Shared typing definition."""
 import ctypes
 import os
-from typing import Optional, Any, TypeVar, Union, Type, Sequence
+from typing import Optional, Any, TypeVar, Union, Type, Sequence, Callable
 
 # os.PathLike/string/numpy.array/scipy.sparse/pd.DataFrame/dt.Frame/
 # cudf.DataFrame/cupy.array/dlpack
@@ -65,3 +65,4 @@ CNumericPtr = ctypes.pointer
 
 # template parameter
 _T = TypeVar("_T")
+_F = TypeVar("_F", bound=Callable[..., Any])
--- a/python-package/xgboost/_typing.py
+++ b/python-package/xgboost/_typing.py
@@ -1,10 +1,12 @@
 """Shared typing definition."""
 import ctypes
 import os
-from typing import Optional, Any, TypeVar, Union, Sequence
+from typing import Optional, List, Any, TypeVar, Union, Type, Sequence
 
 # os.PathLike/string/numpy.array/scipy.sparse/pd.DataFrame/dt.Frame/
 # cudf.DataFrame/cupy.array/dlpack
+import numpy as np
+
 DataType = Any
 
 # xgboost accepts some other possible types in practice due to historical reason, which is
@@ -16,6 +18,7 @@ ArrayLike = Any
 PathLike = Union[str, os.PathLike]
 CupyT = ArrayLike  # maybe need a stub for cupy arrays
 NumpyOrCupy = Any
+NumpyDType = Union[str, Type[np.number]]
 
 # ctypes
 # c_bst_ulong corresponds to bst_ulong defined in xgboost/c_api.h
--- a/python-package/xgboost/_typing.py
+++ b/python-package/xgboost/_typing.py
@@ -1,7 +1,7 @@
 """Shared typing definition."""
 import ctypes
 import os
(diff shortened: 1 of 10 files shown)


python-package/xgboost/_typing.py
```diff
@@ -1,20 +1,24 @@
...
 # os.PathLike/string/numpy.array/scipy.sparse/pd.DataFrame/dt.Frame/
 # cudf.DataFrame/cupy.array/dlpack
+import numpy as np
+
 DataType = Any
 
 # xgboost accepts some other possible types in practice due to historical reason, which is
 # lesser tested.  For now we encourage users to pass a simple list of string.
 FeatureNames = Optional[List[str]]
+FeatureTypes = Optional[List[str]]
```
trivialfis, line 15: We might want to use Sequence instead.

python-package/xgboost/data.py
```diff
@@ -161,12 +162,12 @@ def _maybe_np_slice(data: np.ndarray, dtype) -> np.ndarray:
 
 
 def _from_numpy_array(
-    data,
-    missing,
-    nthread,
+    data: DataType,
+    missing: float,
```
trivialfis, line 166: float is actually not accurate, we can also take np.float32 and related types

trivialfis (commented):
Thank you for working on this, please let me know if you need any help

python-package/xgboost/_typing.py
bridgream: I think it's feasible and am able to pass mypy tests, although internally both `FeatureNames` and `FeatureTypes` are `list`s.

python-package/xgboost/data.py
bridgream: How about using `FloatCompatible = Union[float, np.float32, np.float64]` (do we also need to add `np.float16`?)

bridgream:
Sorry I wasn't able to follow up recently... Can I ask what `cat_codes` is? For example, https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/core.py#L1151

python-package/xgboost/_typing.py
```diff
@@ -1,20 +1,27 @@
...
 # os.PathLike/string/numpy.array/scipy.sparse/pd.DataFrame/dt.Frame/
 # cudf.DataFrame/cupy.array/dlpack
+import numpy as np
+
 DataType = Any
 
 # xgboost accepts some other possible types in practice due to historical reason, which is
 # lesser tested.  For now we encourage users to pass a simple list of string.
-FeatureNames = Optional[List[str]]
+FeatureInfo = Optional[Sequence[str]]
```
bridgream, line 14: Should we change to `Sequence[str]` here, and add `Optional` in the signatures when needed? Some functions require `feature_names` to be _not_ None, for example, https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/core.py#L2562-L2563

python-package/xgboost/_typing.py
trivialfis: Thank you for making the change!

trivialfis:
@bridgream 

> Sorry I wasn't able to follow up recently... Can I ask what cat_codes is? For example

No worries, mypy needs to mature and we are not rushing into it.  cat code is the encoded categorical feature from cudf. Let's just mark it as a list for now.

python-package/xgboost/_typing.py
trivialfis: I think mypy suggests using `Sequence` for function input and more specific type like `List` and `Tuple` as function return types. https://mypy.readthedocs.io/en/stable/generics.html?highlight=variance#variance-of-generic-types

bridgream: Not sure if we're talking about the same thing... I meant to use `FeatureInfo = Sequence[str]` here, and when referring to it, use `Optional[FeatureInfo]`
I've also drafted a commit to better illustrate my idea: https://github.com/bridgream/xgboost/commit/5e3d4aff364a959f784ff316aa66b52484b20448

This isn't something major, though...

trivialfis:
Hi, @bridgream could you please resolve the conflicts?

bridgream:
Hi @trivialfis , could you please restart the Travis CI job? It looks the job failed at random.

trivialfis:
Please ignore it for now.  It's SSH policy is changed.

python-package/xgboost/core.py
```diff
@@ -1603,7 +1609,7 @@ def attr(self, key: str) -> Optional[str]:
             return py_str(ret.value)
         return None
 
-    def attributes(self) -> Dict[str, str]:
+    def attributes(self) -> Dict[str, Optional[str]]:
```
trivialfis, line 1612: I think dict[str, str] should be correct. Have you found a case where only the key is available?

python-package/xgboost/config.py
```diff
@@ -132,7 +140,7 @@ def get_config():
     set_config: Set global XGBoost configuration
     get_config: Get current values of the global configuration
             """)
-def config_context(**new_config):
+def config_context(**new_config: Any) -> Iterator[None]:
```
trivialfis, line 143: Is it a more general generator or an iterator?

python-package/xgboost/_typing.py
```diff
@@ -1,21 +1,32 @@
...
 DataType = Any
 
 # xgboost accepts some other possible types in practice due to historical reason, which is
 # lesser tested.  For now we encourage users to pass a simple list of string.
-FeatureNames = Optional[Sequence[str]]
-FeatureTypes = Optional[Sequence[str]]
+FeatureInfo = Sequence[str]
+FeatureNames = FeatureInfo
+FeatureTypes = FeatureInfo
+ParamType = Union[List, Dict]  # better be sequence but need to remove isinstance(..., list) first
```
trivialfis, line 17: Let's try to avoid the `Type` postfix. Just call it `Param` or `Parameters`.

python-package/xgboost/core.py
```diff
@@ -1347,7 +1353,7 @@ class Booster:
 
     def __init__(
         self,
-        params: Optional[Dict] = None,
+        params: Optional[Union[Dict, List]] = None,
```
trivialfis, line 1356: I think you have defined an alias in the _typing module.

python-package/xgboost/core.py
```diff
@@ -2560,8 +2566,10 @@ def _validate_features(self, data: DMatrix) -> None:
             )
         # Booster can't accept data with different feature names
         if self.feature_names != data.feature_names:
-            dat_missing = set(self.feature_names) - set(data.feature_names)
-            my_missing = set(data.feature_names) - set(self.feature_names)
+            dat_missing = set(cast(Sequence[str], self.feature_names)) - \
+                          set(cast(Sequence[str], data.feature_names))
+            my_missing = set(cast(Sequence[str], data.feature_names)) - \
```
trivialfis, line 2571: cast(FeatureNames).

python-package/xgboost/training.py
```diff
@@ -28,7 +33,7 @@ def _assert_new_callback(
 
 
 def _configure_custom_metric(
-    feval: Optional[Metric], custom_metric: Optional[Metric]
+    feval: FEvalCallable, custom_metric: FEvalCallable
```
trivialfis, line 36: What's the difference between this and the original `Metric`?

trivialfis (commented):
Seems we are a lot closer to fully integrated typing. ;-)

python-package/xgboost/core.py
bridgream: Yes, the case is right above this function at line 1591: 
`def attr(self, key: str) -> Optional[str]:`
`attr(key)` is called in `attributes()` at line 1626:
`return {n: self.attr(n) for n in attr_names}`

python-package/xgboost/config.py
bridgream: At first glance, it seems that `Generator` is more proper than `Iterator`, because `config_context` is implemented as a `Generator` here. However, as `config_context` is essentially a context manager, it's not wrong to annotate the return type as an `Iterator`. In fact, `Iterator` is more general than `Generator`, as `Generator` is a subclass of `Iterator`:
`assert issubclass(Generator, Iterator)`

For mypy, however, the choice remains controversial as tracked by this open issue: https://github.com/python/typeshed/issues/2772

On a side note, mypy says
`The return type of a generator function should be "Generator" or one of its supertypes`
when I use a wrong type hint for the function.

python-package/xgboost/core.py
bridgream: Maybe this can be fixed together with the previous comment, by introducing `BoosterParam`.

python-package/xgboost/training.py
bridgream: It indeed looks unnecessary to introduce a separate `FEvalCallable` as I review the changes I've made. Let me inline it then.

bridgream:
Rebased master. There are 15 mypy errors remaining but they all look a bit complex to tackle.

python-package/xgboost/config.py
trivialfis: > In fact, Iterator is more general than Generator,

Thank you for sharing. I need to remind myself about coroutines. In that case Iterator should be fine.

trivialfis (approved):

trivialfis:
@bridgream Hi, do you think it's ready to be merged?

bridgream:
Hi @trivialfis , thanks for the commits! Yes, I believe we're ready.

trivialfis:
Great work!

Issue #7721, opened 2022-03-09
not providing reproducible predictions after saving in bin format


trevorwilf:
I am not sure if this is an issue or not.  It may be expected bahavior.  I posted a question here about it:
[https://stackoverflow.com/questions/71401330/xgboost-not-providing-reproducible-predictions-after-saving-in-bin-format](url)

essentially, when I use xgb.save using custom parameters the model fails
R version: 4.1.2
XGBoost version 1.5.2.1

The methodology is as follows (logistic-regression, gbtree):
``` 
bst <- xgboost(
                params = best.params
                , data = dtrain
                , nrounds = nrounds
                , early_stopping_rounds = early_stopping_rounds
                , nthread = nthread
                , num_parallel_tree = num_parallel_tree
                , eval_metric =  eval_metric
                , verbose = 2  
                , print_every_n = 1
)

min(predict(bst, dtest))
max(predict(bst, dtest))
xgb.save(bst, savefilemodelloc)
``` 
this produces:
min = 0.17932555079
max = 0.78802382946

now I read the bin back in
```
remove(bst)
bst <- xgb.load(savefilemodelloc)
min(predict(bst, dtest))
max(predict(bst, dtest))
```
this produces:
min = 0.49377295375
max = 0.50564271212

this is being run on the exact same data set, and is producing no where near the same results.  I have tried rebuilding the model several times with nearly identical results.

I have discovered that if I load my parameters back into the model manually it works.
for example
```
remove(bst)
bst <- xgb.load(savefilemodelloc)
xgb.parameters(bst) <- best.params
min(predict(bst, dtest))
max(predict(bst, dtest))
```
this now produces:
min = 0.17932555079
max = 0.78802382946

Is this expected behavior?  

One person has suggested that I am using a deprecated method for saving models, but I could not find another way where I could pass the model between xgboost versions.  Also, I may need to eventually pass this model to python.  I thought this would ensure the best compatibility but I have not tested it.

trivialfis:
Hi, could you please provide a reproducible example that I can run?

PR #7751, opened 2022-03-25
Move `num_parallel_tree` to model parameter.

trivialfis:
The size of forest should be a property of model itself instead of a training
hyper-parameter.

Close https://github.com/dmlc/xgboost/issues/7721 .


RAMitchell (approved):
Makes sense.

Issue #4356, opened 2019-04-10
reading nan or corrupted libsvm file leads to surprising results


kretes:
Reading a libsvm file where some points have 'nan' values leads to surprising results.

1. The DMatrix read by xgboost is different than one read by `sklearn.datasets.load_svmlight_file` when the file contains 'nan'.
Version from `sklearn.datasets.load_svmlight_file` yields the same result for 'nan' as for missing values.
2. When training using such dataset we surprisingly yield better results in terms of ndcg.

I prepared a snippet that reproduces this issue which is runnable under google collab:

https://gist.github.com/kretes/ae8a9948af7d68cb7f8c0757a44ddfd9

You can see there that results obtained by reading 'nan' file directly by xgb are different that the ones obtained from 'empty' file, while reading 'nan' file with `sklearn.datasets.load_svmlight_file` treats nan the same as missing

kretes:
Additional thing is that this is not a bad metric monitoring. When we performed ndcg evaluation on predictions from 'nan' dataset - it was aligned with this from xgboost. This leads to a hypothesis of some kind of label leakage.

I verified and it seems true - I've updated the gist here https://gist.github.com/kretes/8ea87a44aa80228c3ea8006f8fb9c4e7 and when I read a DMatrix with the same X's but different y's -> predictions from model are different, while they should obviously be the same.

trivialfis:
I need to revisit the parser from dmlc core.

kretes:
@trivialfis should I summarize that in a ticket in the dmlc-core ?

trivialfis:
Thanks.  No need, it's just I'm quite busy currently.   Will look into this.

trivialfis:
@hcho3 dmlc core defines its own `strtof` for performance, but doesn't support `nan` or `inf`.  I will see if it's possible to add the support or just generate a warning.

hcho3:
@trivialfis The latest dmlc-core actually has support for `nan` and `inf`: https://github.com/dmlc/dmlc-core/commit/8afbca699d9d9b720024e32d47a49f29b59c4c9e

trivialfis:
Awesome, then I will try it tomorrow see if it can close this.

CodingCat:
...every time for such a title to jump out from my inbox, I need to take a second look....

PR #7753, opened 2022-03-25
Remove text loading in basic walk through demo.

trivialfis:


RAMitchell (commented):
Any particular reason to change this? One downside can be that `load_svmlight_file` is much slower.

trivialfis:
@RAMitchell For the Python package, I want to move away from the text parser in dmlc, or at least it should not be the first option users think of (as in the basic walkthrough). If we want to advertise its performance benefit, we should first fix related issues like: https://github.com/dmlc/xgboost/issues/4356 , and probably move the maintenance from dmlc-core to XGBoost.

hcho3 (approved):

PR #7759, opened 2022-03-27
R package: Supporting non-standard libomp locations during installation from source

### LINKED ISSUE #7757: R package does not install when the OpenMP library is in a non-standard place
Hello,

I'm trying to install XGBoost from the current CRAN package version (1.5.2.1) on MacOS Catalina 10.15.7 with a Hombrew-installed libomp. 

The installation fails in my case because my Homebrew installation - and thus libomp - is in a non-standard place (`/opt/brew` rather than `/usr/local`). The reason for the failure is that the `configure.ac` of the R package does not make use  of existing variables such as `CPPFLAGS` or `LDFLAGS` when testing if  it is supported. However,  the R build environment provides those; using them in `configure.ac` for compiling the test program works fine and solves the problem in my case.

I believe this might also be an issue on Macs with an ARM processor; as far as I know, Homebrew installs itself in `/opt/homebrew` in that case, and the XGBoost installation should fail in a similar way.
  
Is there any reason why these R-provided environment variables are not used in the `configure.ac`? If not, do you accept a pull request?

Many thanks,

   Christian.

cmarquardt:
This pull request adds CPPFLAGS and LDFLAGS to the testing for OpenMP during the R installation from a source package and should fix #7757.

The real change is in `R-package/configure.ac`; most changes in `R-package/configure` are because I have a more recent version of `autoconf` than the one used for the previous version of the file.

These changes work when applied against the xgboost package on CRAN. I haven't fully understood how to generate the source package for R in the full xgboost source code tree, though; there's a top-level Makefile with a target `Rpack` copying files into a subdirectory `xgboost`. The contents of that directory look like an R package, but no tar.gz or similar is generated. 

When trying to do an `R CMD INSTALL .` from inside the directory created with `make Rpack`, the compilation fails with C++ build errors for `amalgamation/xgboost-all0.cc`. However, the compilation also fails without my changes in `master`.  Thus, I believe it's unrelated to the modification proposed in this pull request - but I have no working knowledge of C++, unfortunately. And maybe it's not the right way to get the R source package anyway. The `cmake` build of the R package appears to work fine, but as far as I understand that won't be affected by the changes in this PR anyway.


trivialfis (approved):
Thank you for the fix!

Issue #6851, opened 2021-04-13
Add support for pandas nullable types in feature matrix


freddyaboulton:
I would like to use [pandas nullable types](https://pandas.pydata.org/docs/user_guide/integer_na.html) with xgboost. Currently, xgboost will raise a ValueError if any of the features use the new nullable pandas types. 

Repro with xgb `1.2.1` and pandas `1.2.3`:

```python
from xgboost import XGBClassifier
import pandas as pd
import pytest

for dtype in ['Int64', 'Float64', 'boolean']:
    
    xgb = XGBClassifier()

    # ValueError if only features use nullable types
    df = pd.DataFrame({"a": pd.Series([True, False, True, False], dtype=dtype),
                       "b": pd.Series([4, 5, 6])})
    y = pd.Series([1, 0, 1, 0])
    
    with pytest.raises(ValueError):
        xgb.fit(df, y)


# No ValueError if use "old" types
for dtype in ['int', 'float', 'bool']:
    
    df = pd.DataFrame({"a": pd.Series([True, False, True, False], dtype=dtype)})
    y = pd.Series([1, 0, 1, 0])

    xgb = XGBClassifier()
    xgb.fit(df, y)
```

As a user, I would like this feature so I can build a data processing pipeline using the latest pandas capabilities.

Thank you!

trivialfis:
Took a quick look.  I think a better way is to work on https://github.com/dmlc/xgboost/pull/5667 so that we can have masked arrays.

freddyaboulton:
Thanks for taking a look @trivialfis ! So a feature matrix with pandas nullable types will be converted to a masked array? What's the status of #5667 ?

trivialfis:
> So a feature matrix with pandas nullable types will be converted to a masked array?

That's what I'm hoping and I suspect that's how it's implemented for nullable arrays.

> What's the status of #5667 ?

I assigned myself to work on it based on the existing work by @zhangzhang10 .  But arrow can trunk the data in an arbitrary way so it's a bit messy, I will try to follow up.

Ark-kun:
XGBoost still fails when encountering nullable floating-point dtypes.

gsheni:
@Ark-kun I believe this now works
```python
from xgboost import XGBClassifier
import pandas as pd
import pytest

xgb = XGBClassifier()

# ValueError if only features use nullable types
df = pd.DataFrame({"a": pd.Series([True, False, True, False], dtype='boolean'),
[...]

PR #7760, opened 2022-03-28
Support pandas nullable types.

trivialfis:
* Check for nullable type input.

cuDF uses nullable type by default, no special support is needed.

Close https://github.com/dmlc/xgboost/issues/6851 .

### DIFF
--- a/python-package/xgboost/data.py
+++ b/python-package/xgboost/data.py
@@ -247,6 +247,7 @@ be set to `True`.""" + err
     raise ValueError(msg)
 
 
+# pylint: disable=too-many-locals
 def _transform_pandas_df(
     data: DataFrame,
     enable_categorical: bool,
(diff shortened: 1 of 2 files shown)


hcho3 (approved):

python-package/xgboost/data.py
```diff
@@ -251,11 +257,26 @@ def _transform_pandas_df(
...
+        is_sparse,
+        is_categorical_dtype,
+        is_integer_dtype,
+        is_bool_dtype,
+    )
+
+    nullable_alias = {"Int16", "Int32", "Int64"}
+
+    # dtype: pd.core.arrays.numeric.NumericDtype
+    def is_nullable_dtype(dtype: Any) -> bool:
```
Ark-kun, line 270: Why is this function needed?
It does not seem to make any difference when the same types are in `_pandas_dtype_mapper`.

python-package/xgboost/data.py
```diff
@@ -293,6 +316,9 @@ def _transform_pandas_df(
                     .cat.codes.astype(np.float32)
                     .replace(-1.0, np.NaN)
                 )
+            elif is_nullable_dtype(dtype):
+                # Converts integer <NA> to float NaN
+                transformed[data.columns[i]] = data[data.columns[i]].astype(np.float32)
```
Ark-kun, line 321: Is it correct to convert Int64 to float32?

python-package/xgboost/data.py
```diff
@@ -251,11 +257,26 @@ def _transform_pandas_df(
...
     import pandas as pd
-    from pandas.api.types import is_sparse, is_categorical_dtype
+    from pandas.api.types import (
+        is_sparse,
+        is_categorical_dtype,
+        is_integer_dtype,
+        is_bool_dtype,
+    )
+
+    nullable_alias = {"Int16", "Int32", "Int64"}
```
Ark-kun, line 267: `boolean` is nullable as well. Either it should be in this set or this set should eb renamed.

PR #7781, opened 2022-04-04
[jvm-packges] set the correct objective if user doesn't explicitly set it

wbo4958:
This PR deletes the default value of "objective" and sets it correctly if the user does not explicitly set it.

For example, If the user does not see objective 
- XGBoostClassifier will set the objective according to the num class
- XGBoostRegressor will default the objective to "reg:squarederror"

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
@@ -178,17 +178,15 @@ class XGBoostClassifier (
     }
 
     if (_numClasses == 2) {
-      if (getObjective != "binary:logistic") {
-        throw new IllegalArgumentException("when num class = 2, the objective must be set to" +
-          s" binary:logistic, found $getObjective")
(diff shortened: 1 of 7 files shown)


jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
```diff
@@ -171,6 +171,27 @@ class XGBoostClassifier (
...
   override protected def train(dataset: Dataset[_]): XGBoostClassificationModel = {
+    val _numClasses = getNumClasses(dataset)
+    if (isDefined(numClass) && $(numClass) != _numClasses) {
+      throw new Exception("The number of classes in dataset doesn't match " +
+        "\'num_class\' in xgboost params.")
+    }
+
+    if (_numClasses == 2) {
+      if (getObjective != "binary:logistic") {
+        logWarning(s"Incorrect objective [${getObjective}] for classification " +
```
trivialfis, line 182: I think it's not necessary to change the user setting.

wbo4958:
@trivialfis please help to review it. Thx

trivialfis (approved):

PR #7788, opened 2022-04-06
[jvm-packages] Update jackson-databind to 2.12.6

hcho3:
Continue #7773, to address a security advisory.

https://github.com/advisories/GHSA-57j2-w4cx-62h2

@trivialfis @wbo4958

PR #7791, opened 2022-04-08
[jvm-packages] remove the dep of com.fasterxml.jackson

wbo4958:
There is no need to add jackson for handling array inferface, since
the json of array interface is really simple.

To fix https://github.com/dmlc/xgboost/pull/7788


trivialfis (approved):
LGTM, assuming all tests can pass.

PR #7812, opened 2022-04-15
[POC] Experimental support for l1 error.

trivialfis:
Support adaptive tree, a feature supported by both sklearn and lightgbm.  The tree leaf is recomputed based on residue of labels and predictions after construction.

For l1 error, the optimal value is the median (50 percentile).

This is marked as experimental support for the following reasons:
- The value is not well defined for distributed training, where we might have empty leaves for local workers. Right now I just use the original leaf value for computing the average with other workers, which might cause significant errors.
- The GPU implementation is not ideal due to sampling support.  The partition is calculated from node position, if a node doesn't have any valid sample then it's missing in the partitioner and we will have to fill it back.


src/common/stats.cuh
```diff
@@ -0,0 +1,128 @@
+/*!
+ * Copyright 2022 by XGBoost Contributors
+ */
+#ifndef XGBOOST_COMMON_STATS_CUH_
+#define XGBOOST_COMMON_STATS_CUH_
+
+#include <thrust/iterator/counting_iterator.h>
+#include <thrust/iterator/permutation_iterator.h>
+#include <thrust/sort.h>
```
trivialfis, line 9: Not needed.

src/common/partition_builder.h
```diff
@@ -279,6 +280,38 @@ class PartitionBuilder {
     return blocks_offsets_[nid] + begin / BlockSize;
   }
 
+  // Copy row partitions into global cache for reuse in objective
+  template <typename Sampledp>
+  void LeafPartition(Context const* ctx, RegTree const& tree, RowSetCollection const& row_set,
```
trivialfis, line 285: Add test with sampling

src/common/common.h
```diff
@@ -189,15 +251,38 @@ std::vector<Idx> ArgSort(Container const &array, Comp comp = std::less<V>{}) {
...
 
   XGBOOST_DEVICE float operator[](size_t i) const { return weights.empty() ? dft : weights[i]; }
 };
+
+
+/**
+ * Last index of a group in a CSR style of index pointer.
+ */
+template <typename Idx>
+XGBOOST_DEVICE size_t LastOf(size_t group, common::Span<Idx> indptr) {
```
trivialfis, line 278: Remove it.

src/common/common.h
```diff
@@ -189,15 +251,38 @@ std::vector<Idx> ArgSort(Container const &array, Comp comp = std::less<V>{}) {
   return result;
 }
 
+template <typename V, typename Comp = std::less<V>>
```
trivialfis, line 254: Test

src/common/common.h
```diff
@@ -7,19 +7,20 @@
 #define XGBOOST_COMMON_COMMON_H_
 
 #include <xgboost/base.h>
+#include <xgboost/linalg.h>
```
trivialfis, line 10: Not needed.

include/xgboost/tree_model.h
```diff
@@ -734,5 +735,18 @@ inline bool RegTree::FVec::IsMissing(size_t i) const {
 inline bool RegTree::FVec::HasMissing() const {
   return has_missing_;
 }
+
+struct RowIndexCache {
+  HostDeviceVector<size_t> row_index;
```
trivialfis, line 740: Doc

tests/python-gpu/test_gpu_updaters.py
```diff
@@ -3,7 +3,7 @@
 import gc
 import pytest
 import xgboost as xgb
-from hypothesis import given, strategies, assume, settings, note
+from hypothesis import given, strategies, assume, settings, note, reproduce_failure, seed
```
trivialfis, line 6: Not needed.

tests/cpp/common/test_stats.cc
```diff
@@ -0,0 +1,58 @@
...
+ */
+#include <gtest/gtest.h>
+#include <xgboost/generic_parameters.h>
+
+#include "../../../src/common/stats.h"
+
+namespace xgboost {
+namespace common {
+TEST(Stats, Percentil) {
+  {
```
trivialfis, line 12: Name

src/gbm/gbtree.cc
```diff
@@ -216,53 +217,60 @@ void CopyGradient(HostDeviceVector<GradientPair> const* in_gpair, int32_t n_thre
...
+                            ObjFunction const* obj, size_t gidx,
+                            std::vector<std::unique_ptr<RegTree>>* p_trees) {
+  if (!obj || !obj->Task().UpdateTreeLeaf()) {
+    return;
+  }
+  auto& trees = *p_trees;
+  for (size_t tree_idx = 0; tree_idx < trees.size(); ++tree_idx) {
+    auto row_idx = updaters_.back()->GetRowIndexCache(tree_idx);
+    // distinguish the difference between multi-class and multi-target.
+    obj->UpdateTreeLeaf(row_idx, p_fmat->Info(), predictions, trees[tree_idx].get());
```
trivialfis, line 230: Not supported

src/objective/regression_obj.cu
```diff
@@ -675,5 +670,191 @@ XGBOOST_REGISTER_OBJECTIVE(TweedieRegression, "reg:tweedie")
...
 
+namespace detail {
+void UpdateLeafValues(std::vector<float>* p_quantiles, RowIndexCache const& row_index,
+                      RegTree* p_tree) {
+  auto& tree = *p_tree;
+  auto& quantiles = *p_quantiles;
+  auto const& h_node_idx = row_index.node_idx.HostVector();
+
+  size_t n_leaf{h_node_idx.size()};
+  rabit::Allreduce<rabit::op::Max>(&n_leaf, 1);
```
trivialfis, line 681: We can gather the number of invalid nodes before allreduce.

src/tree/gpu_hist/row_partitioner.cuh
```diff
@@ -153,21 +179,147 @@ class RowPartitioner {
...
-   * position as an argument and return the new position for this training
-   * instance.
+   * \param p_out_row_indices Row partitions for each leaf.
+   * \param op Device lambda. Should provide the row index and current position as an
+   * argument and return the new position for this training instance.
    */
-  template <typename FinalisePositionOpT>
-  void FinalisePosition(FinalisePositionOpT op) {
+  template <typename FinalisePositionOpT, typename Sampledp>
+  void FinalisePosition(Context const* ctx, RegTree const* p_tree, size_t n_leaf, ObjInfo task,
```
trivialfis, line 187: Document the expected output from this function and possible edge cases.

src/tree/gpu_hist/row_partitioner.cuh
```diff
@@ -2,11 +2,37 @@
...
 #include "xgboost/base.h"
 #include "../../common/device_helpers.cuh"
+#include "xgboost/generic_parameters.h"
+#include "xgboost/task.h"
+#include "xgboost/tree_model.h"
 
 namespace xgboost {
 namespace tree {
+namespace detail {
+inline void FillMissingLeaf(std::vector<bst_node_t> const& missing, RowIndexCache* p_row_indices) {
```
trivialfis, line 16: Call for better ideas.

src/tree/updater_quantile_hist.cc
```diff
@@ -228,12 +244,14 @@ void QuantileHistMaker::Builder<GradientSumT>::ExpandTree(
     expand_set = driver.Pop();
   }
 
+  this->LeafPartition(tree, gpair_h, p_out_row_indices);
```
trivialfis, line 247: Make this optional

src/tree/updater_approx.cc
```diff
@@ -231,6 +242,8 @@ class GloablApproxBuilder {
       driver.Push(best_splits.begin(), best_splits.end());
       expand_set = driver.Pop();
     }
+
+    this->LeafPartition(tree, hess, p_out_row_indices);
```
trivialfis, line 246: Make this optional.

trivialfis:
Missing features in this PR:
- Support for multi-output models.
- Initial estimation before training.

include/xgboost/tree_updater.h
```diff
@@ -78,6 +78,13 @@ class TreeUpdater : public Configurable {
   }
 
   virtual char const* Name() const = 0;
+  /*!
+   * \brief Get the partition of rows based on the last iteration.
+   */
+  virtual common::Span<RowIndexCache const> GetRowIndexCache(size_t tree_idx) const {
```
trivialfis, line 84: Maybe we should skip the step silently?

tests/python/test_with_dask.py
```diff
@@ -1285,7 +1289,12 @@ def is_stump():
         def minimum_bin():
             return "max_bin" in params and params["max_bin"] == 2
 
-        if minimum_bin() and is_stump():
+        # See note on `ObjFunction::UpdateTreeLeaf.
```
trivialfis, line 1292: Typo

tests/python/test_with_dask.py
```diff
@@ -1305,9 +1314,10 @@ def test_hist(
            dataset=tm.dataset_strategy)
     @settings(deadline=None, suppress_health_check=suppress, print_blob=True)
     def test_approx(
-            self, client: "Client", params: Dict, dataset: tm.TestDataset
+        self, client: "Client", params: Dict, dataset: tm.TestDataset
     ) -> None:
         num_rounds = 30
+        # params["eta"] = 0.1
```
trivialfis, line 1320: Remove it.

RAMitchell (commented):
Looks significantly better to me without trying to share the internal row representations inside the updaters.

include/xgboost/tree_updater.h
```diff
@@ -49,18 +49,25 @@ class TreeUpdater : public Configurable {
    *  existing trees.
    */
   virtual bool CanModifyTree() const { return false; }
+  /*!
+   * \brief Wether the out_position in `Update` is valid. This determines whether adaptive
+   *        tree can be used.
+   */
+  virtual bool HasNodePosition() const { return false; }
```
RAMitchell, line 56: In theory we can make every updater support node position right? In xgboost we are using a kind of 'fake object-oriented programming'. We define an interface of functionality that all tree updaters should implement, but then make some updaters implement this functionality and others not.

trivialfis: @RAMitchell We have exact updater and pruner.  Surely, we can work on them too, but that's just too much for 1 PR I think.

src/tree/updater_gpu_hist.cu
```diff
@@ -641,7 +681,8 @@ struct GPUHistMakerDevice {
 
         int left_child_nidx = tree[candidate.nid].LeftChild();
         int right_child_nidx = tree[candidate.nid].RightChild();
-        // Only create child entries if needed
+        // Only create child entries if needed_
+        p_tree->GetNumLeaves();
```
trivialfis, line 685: Remove this one.

src/tree/updater_gpu_hist.cu
```diff
@@ -719,41 +764,36 @@ class GPUHistMakerSpecialised {
...
-    maker.reset(new GPUHistMakerDevice<GradientSumT>(device_,
-                                                     page,
-                                                     info_->feature_types.ConstDeviceSpan(),
-                                                     info_->num_row_,
-                                                     param_,
-                                                     column_sampling_seed,
-                                                     info_->num_col_,
-                                                     batch_param));
+    dh::safe_cuda(cudaSetDevice(ctx_->gpu_id));
+    info_->feature_types.SetDevice(ctx_->gpu_id);
```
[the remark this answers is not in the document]
trivialfis: The change for `device_` -> `ctx_->gpu_id` might not be strictly related, I changed it when I ran into a segfault during the test of `gpu_id` being specified to 1.  I can revert those if needed, but I think these changes are easy to skip.

src/tree/gpu_hist/row_partitioner.cuh
```diff
@@ -149,23 +153,47 @@ class RowPartitioner {
...
    */
-  template <typename FinalisePositionOpT>
-  void FinalisePosition(FinalisePositionOpT op) {
+  template <typename FinalisePositionOpT, typename Sampledp>
+  void FinalisePosition(Context const* ctx, RegTree const* p_tree, size_t n_leaf, ObjInfo task,
+                        HostDeviceVector<bst_node_t>* p_out_row_indices, FinalisePositionOpT op,
+                        Sampledp sampledp) {
     auto d_position = position_.Current();
     const auto d_ridx = ridx_.Current();
+    if (!task.UpdateTreeLeaf()) {
```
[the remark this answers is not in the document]
trivialfis: I can merge it if the performance doesn't change that much.

src/common/partition_builder.h
```diff
@@ -279,6 +281,29 @@ class PartitionBuilder {
...
+  template <typename Sampledp>
+  void LeafPartition(Context const* ctx, RegTree const& tree, RowSetCollection const& row_set,
+                     std::vector<bst_node_t>* p_position, Sampledp sampledp) const {
+    auto& h_pos = *p_position;
+    h_pos.resize(row_set.Data()->size(), std::numeric_limits<bst_node_t>::max());
+
+    auto p_begin = row_set.Data()->data();
+    ParallelFor(row_set.Size(), ctx->Threads(), [&](size_t i) {
+      auto const& node = row_set[i];
+      if (node.node_id < 0 || !tree[node.node_id].IsLeaf()) {
```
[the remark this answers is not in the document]
trivialfis: On the CPU, all nodes are being kept track of including internal nodes. For node_id < 0, it should be an internal node and we can skip the `IsLeaf` check, I will see if there's any edge case I missed.

src/tree/updater_gpu_hist.cu
```diff
@@ -476,13 +499,29 @@ struct GPUHistMakerDevice {
...
+                                    cudaMemcpyHostToDevice));
+      auto d_nodes = dh::ToSpan(nodes);
+      dh::LaunchN(d_ridx.size(), [=] XGBOOST_DEVICE(size_t idx) mutable {
+        bst_node_t nidx = d_position[idx];
+        auto weight = d_nodes[nidx].LeafValue();
+        out_preds_d(d_ridx[idx]) += weight;
+      });
+    } else {
+      // Avoid copying nodes by using evaluator to get leaf weight on-the-fly, this is
+      // useful when tree leaf is not updated after tree construction.
```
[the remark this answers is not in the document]
trivialfis: Haven't run the full benchmark yet, copying nodes is probably fine, but not sure about the memory allocation.

src/tree/gpu_hist/row_partitioner.cuh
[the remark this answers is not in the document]
trivialfis: For most of the changes, I try to avoid memory allocation, will have to run full benchmarks to check the actual changes in performance.

src/common/partition_builder.h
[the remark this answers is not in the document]
trivialfis: Turned the condition into a check.

trivialfis:
This is an initial benchmark with gbm-bench.  I haven't run the benchmark with the newly added l1 error, which can use some optimizations in quantile computation and will benchmark after the related work is finished.

|                   | bosch             |                    | higgs              |                    | years              |                   |
|:------------------|:------------------|--------------------|:-------------------|--------------------|:-------------------|-------------------|
|                   | Time              | AUC                | Time               | AUC                |                    | MSE               |
| CPU Hist (master) | 79.08961106099741 | 0.6902547809021118 | 141.6087378029988  | 0.8399291373085628 | 30.87248679699769  | 79.77742340435458 |
| CPU Hist (PR)     | 77.38803030599956 | 0.6902547809021118 | 137.95923503099766 | 0.8399291373085628 | 24.732420017997356 | 79.77742340435458 |
| GPU Hist (master) | 21.8356087760003  | 0.6887260447943401 | 23.57480316100191  | 0.839534647606739  | 10.975372369997785 | 80.31760322683175 |
| GPU Hist (PR)     | 21.80807627499962 | 0.6887260388302106 | 24.510682308002288 | 0.8395346476950757 | 10.876273957001104 | 79.91012003330812 |

RAMitchell (approved):

s-banach:
Hey @trivialfis, looks like great work!
Sorry I can't read C++ very well, so I hope I can ask for clarification here.
The 1.7.0 release notes claim you can optimize "without a valid hessian".
Does that mean the hessian is ignored completely when choosing splits?

It seems to be implicit in the xgboost paper than the hessian is supposed to be nonnegative.
I wonder if this approach would allow custom objectives where the hessian may sometimes be negative?

trivialfis:
@s-banach 

At the moment XGBoost uses the sample weight as hessian (default to 1) for l1, and recomputes the leaf values based on input labels after growing the tree.

> I wonder if this approach would allow custom objectives where the hessian may sometimes be negative

We haven't exposed the feature to custom objectives yet unless you are writing C++ extensions.

kayhman:
Hi @trivialfis,

This is a very interesting approach. How do you update the value after the tree is built ? 

The changelog mention line search, but I see no mention of that in the code. I guess that this is the method UpdateTreeLeaf of the objectives that does the job (in UpdateTreeLeafHost ?), but I don't understand what mathematical operation is done.

Can you provide more details?

Best regards,

kayhman:
ok, I see. The line search is done thought the `alpha` parameter, and you use the `alpha`-quantile as leaf value?

And as far as I can see, alpha is always set to 0.5. 

Is that right?

trivialfis:
For l1, the optimal is median, which is 0.5 quantile.

kayhman:
ok, thanks for the answer. So no needs for line search in this case.

PR #7823, opened 2022-04-20
[jvm-packages] move the dmatrix building into rabit context

wbo4958:
For multi-node training for the whole gpu pipeline, the dmatrix building needs to be put in the rabit context.


trivialfis:
Stopped the build due to hang.

wbo4958:
Headache, I can run it on two different machines locally without any issue.

wbo4958:
@trivialfis, for some reason, one test hangs in CI env. So I just skip this test in order to make this PR merge.

I will add barrier execution mode which launch all the xgboost training task at the same time and add barrier when training to fix the sparkcontext being killed issue in the following PR.

trivialfis (approved):

Issue #7728, opened 2022-03-14
Single precision GPU hist is flaky with cali dataset.


trivialfis:
The issue is likely caused by floating-point error along with the global rng used by sampling.  Will investigate once we can remove the global rng.

https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-7512/53/pipeline

```
[...]

PR #7828, opened 2022-04-21
Remove single_precision_histogram

RAMitchell:
Removing the `single_precision_histogram` due to major observed accuracy issues. In particular many tests can fail due to numerical error accumulating in gain calculation, resulting in leaf nodes being opened that contain no training data at all.

In this PR I have removed a now redundant class in gpu_hist and slightly refactored the TreeUpdater base class to always require a global context for construction.


trivialfis:
Close https://github.com/dmlc/xgboost/issues/7728 .

src/tree/updater_gpu_hist.cu
```diff
@@ -532,6 +529,13 @@ struct GPUHistMakerDevice {
 
   void ApplySplit(const GPUExpandEntry& candidate, RegTree* p_tree) {
     RegTree& tree = *p_tree;
+
+    // Sanity check - have we created a leaf with no training instances?
+    if (!rabit::IsDistributed() && row_partitioner) {
```
trivialfis, line 534: Any potential solution if this check occurs?

trivialfis: My mistake.  I ran into issues with empty nodes in partitioner `UpdatePosition`, where `left_count` can be 0 due to floating-point errors.  In this `ApplySplit` function, if we check on the left/right child instead of the parent (candidate), then we might run into trouble.

tests/python-gpu/test_gpu_updaters.py
```diff
@@ -3,7 +3,7 @@
 import gc
 import pytest
 import xgboost as xgb
-from hypothesis import given, strategies, assume, settings, note
+from hypothesis import given, strategies, assume, settings, note, reproduce_failure
```
trivialfis, line 6: I think this is forgotten.

trivialfis (commented):
Apologies for the conflict.  Do you need help with rebasing?

trivialfis (approved):

PR #7832, opened 2022-04-22
"irrevelant" to "irrelevant"

forestkey:
As we are moving forward, the gap between them is becoming more and more "irrevelant". I guess "irrevelant" should be "irrelevant"


trivialfis (approved):
Thank you for the fix!

Issue #7835, opened 2022-04-22
[FEA] [JVM-Packages] Add barrier execution mode for support


wbo4958:
This FEA is based on https://github.com/dmlc/xgboost/pull/5625 and https://github.com/dmlc/xgboost/issues/4793.

## Why we need barrier execution mode

The Rabit hanging issue, for example,

1. Rabit.init() is like a barrier, and it will wait for other workers for synchronization. If one worker died before Rabit.init(), then it will hang.
2. During training, if one work died, it will call Rabit.finalize(), other workers may wait on allreduce, then it hang.

How xgboost resolve this issue. It is killing SparkContext, which is a really bad user experience. On the other hand, Killing SparkContext is depending on SparkListener, sometimes even if the task failed, xgboost can't receive the notification in some cases (I don't know why), so the killing threads will not be started, then it will hang forever.

What xgboost can benefit from barrier execution mode.

1. the xgboost tasks can be launched at once

Previously, if there is available slot, Task scheduled will offer some (not all) xgboost task with the resource, so some tasks may run in advance, while others may still wait for the resource. 

2. abort the barrier stage if one task failed.

The one feature of barrier stage is if one task failed, Spark DAG/TaskScheduler will kill all other running tasks, which does not depend on any spark listener. With this, xgboost will not be hanged forever. At the same time, we can get rid of xgboost sparkcontext killer. 

## Side effect

Spark does not support barrier mode and dynamic allocation at the same time. But xgboost may have some un-expected behavior when enabling dynamic allocation, see https://github.com/dmlc/xgboost/pull/5625#issuecomment-622651681. So I think it is ok.

PR #7836, opened 2022-04-24
[Breaking][jvm-packages] add barrier execution mode

wbo4958:
To fix https://github.com/dmlc/xgboost/issues/7835. Credit to @[WeichenXu123](https://github.com/WeichenXu123)

By introducing the barrier execution mode, we don't need to kill SparkContext when some xgboost tasks failed, instead, Spark will abort the whole barrier stage which does not depend on SparkListener, we will never encounter the xgboost task hang issue. So in this PR, the killSparkContextOnWorkerFailure parameter is deleted.

Since this PR has deleted Spark SparkParallelismTracker, the timeoutRequestWorkers parameter is not needed anymore. But one test `cross-version model loading (0.82)` will fail if we just delete timeoutRequestWorkers, we just keep this parameter in this PR. Will file a following up PR to delete timeoutRequestWorkers parameters.


wbo4958:
@trivialfis could you help to review it?

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostGeneralSuite.scala
```diff
@@ -375,13 +374,14 @@ class XGBoostGeneralSuite extends FunSuite with TmpFolderPerSuite with PerTest {
...
-    val trainingDF = buildDataFrame(Seq(XGBLabeledPoint(1.0f, 1, Array(), Array())))
+    val trainingDF = buildDataFrame(Seq(XGBLabeledPoint(1.0f, 4,
+      Array(0, 1, 2, 3), Array(0, 1, 2, 3))))
     val xgb = new XGBoostClassifier(paramMap)
-    intercept[XGBoostError] {
-      val model = xgb.fit(trainingDF)
+    intercept[SparkException] {
+      xgb.fit(trainingDF)
     }
   }
```
trivialfis, line 386: Are these changes strictly related to the support of barrier mode?

trivialfis (commented):
Thank you for working on this, being able to remove code and improve robustness is exciting!

- Could you please share more details on what's the breaking change and how users should adapt to it?
- Could you please keep some RABIT mock tests here to show that exceptions are indeed being well handled?

wbo4958:
> Thank you for working on this, being able to remove code and improve robustness is exciting!
> 
> * Could you please share more details on what's the breaking change and how users should adapt to it?

killSparkContextOnWorkerFailure parameter is deleted. Users are supposed not to use this parameter.

> * Could you please keep some RABIT mock tests here to show that exceptions are indeed being well handled?

Not exactly. the deleted case [test("test SparkContext should not be killed ")](https://github.com/dmlc/xgboost/pull/7836/files#diff-83dc1ab309d0aaa9e7ec44586d99b82cee3c34daa70592fccaca834c6acc27afL121) is only for `kill_spark_context_on_worker_failure` parameter, and there already has the rabbit mock tests https://github.com/dmlc/xgboost/pull/7836/files#diff-83dc1ab309d0aaa9e7ec44586d99b82cee3c34daa70592fccaca834c6acc27afR85

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostGeneralSuite.scala
wbo4958: Completely not. I just make this tests more like the test description.

trivialfis (approved):

PR #7845, opened 2022-04-26
[jvm-packages] model loading should be compatible with old models

wbo4958:
For the parameters not used, XGBoost should delete them, instead of
keeping it, which will cause some issue for the old model.

For example, the kill_spark_context_on_worker_failure is deleted,
but when loading the model saved in 1.6.0, (in which has
kill_spark_context_on_worker_failure parameter).
XGBoost will throw an exception


wbo4958:
@trivialfis , this PR should also be back-ported to 1.6.1

Thx

trivialfis (approved):

PR #7846, opened 2022-04-27
[jvm-package] remove the coalesce in barrier mode

wbo4958:
Barrier mode doesn't allow coalesce operation, which should fail
all tests of xgboost4j-spark-gpu, but it doesn't. That's because
the test file is extremly small which result in only 1 PartitionedFile
and finally bypass the checking.

This PR makes at least multi-PartitionedFiles for a test file, and
then the existing tests can cover this PR


wbo4958:
@trivialfis, please also port this PR back to 1.6.1

Thx

trivialfis (approved):

PR #7848, opened 2022-04-29
[CI] make container be able to re-attached

wbo4958:
When re-starting the container, it will fail in entrypoint.sh which
will exit when adding an existing group or user


hcho3 (approved):

hcho3:
Thanks!

PR #7852, opened 2022-04-29
Booster parameters typing

amotzop:
* Added a `TypedDict` describing the possible Booster parameters, and added it as a type hint.
* Added typing to the `training.py` file.

### DIFF
--- a/python-package/xgboost/_typing.py
+++ b/python-package/xgboost/_typing.py
@@ -1,7 +1,20 @@
 """Shared typing definition."""
 import ctypes
 import os
-from typing import Optional, Any, TypeVar, Union, Sequence
+import numpy as np
+from typing import (
+    Optional,
+    Any,
+    TypeVar,
+    Union,
+    Sequence,
+    Literal,
+    Tuple,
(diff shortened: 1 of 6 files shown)


trivialfis (commented):
Thank you for working on this, but I would like to avoid having another copy of parameter list in Python, which will complicate the maintenence.

hcho3:
I agree with @trivialfis. XGBoost has multiple interfaces (Python, Java, R), and we want to keep the list of parameters in one place.

amotzop:
I understand.
I just wanted to ask what do you think about the following:
1. Commiting just the extra typing information in `training.py`?
2. Adding code to automatically generate the `Parameters` TypedDict from the existing list of parameters (and it will be just an alias to `Dict[str,Any]` before the generation)?

I'm asking because I think that generally such typing information can really help find bugs/mistakes in code and so I'll like to find some way to both have it and keep it easily maintainable.

trivialfis:
Thank you for the suggestion.  There's a list in sklearn API with type information.

PR #7857, opened 2022-04-30
Always use partition based categorical splits.

trivialfis:
My mistake, we are always using regression trees.

### DIFF
--- a/src/common/categorical.h
+++ b/src/common/categorical.h
@@ -12,7 +12,6 @@
 #include "xgboost/data.h"
 #include "xgboost/parameter.h"
 #include "xgboost/span.h"
-#include "xgboost/task.h"
 
 namespace xgboost {
 namespace common {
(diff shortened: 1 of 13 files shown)


src/common/categorical.h
```diff
@@ -83,7 +83,7 @@ inline void InvalidCategory() {
  * \brief Whether should we use onehot encoding for categorical data.
  */
 XGBOOST_DEVICE inline bool UseOneHot(uint32_t n_cats, uint32_t max_cat_to_onehot, ObjInfo task) {
```
RAMitchell, line 85: Remove task argument. I think with this change task can be removed from a lot of function arguments.

trivialfis: Thank you for reminding me of that. All removed.

RAMitchell (approved):

PR #7858, opened 2022-05-02
Small refactor to categoricals

RAMitchell:
Current implementation blocks any changes to the order in which nodes are opened in the tree updater. When a one hot encoded categorical split occurs, this information was not stored on the device. In order for the update position function to work correctly this information had been copied to the device beforehand inside the ApplySplit member function. Therefore the update position call had a dependence on the ApplySplit function for the same node being called directly before.

This PR changes the one hot encoding path to store categorical split information on the device in the same way as sort based splits, making it available on the device.

I changed the vectors storing this categorical split information for all nodes to expand when necessary instead of being initialised according to the maximum tree size. This prevents OOM in cases where the user specifies very deep trees which end up being sparse.

### DIFF
--- a/src/tree/gpu_hist/evaluator.cu
+++ b/src/tree/gpu_hist/evaluator.cu
@@ -44,11 +44,6 @@ void GPUHistEvaluator<GradientSumT>::Reset(common::HistogramCuts const &cuts,
     node_categorical_storage_size_ =
         common::CatBitField::ComputeStorageSize(cuts.MaxCategory() + 1);
     CHECK_NE(node_categorical_storage_size_, 0);
-    // We need to allocate for all nodes since the updater can grow the tree layer by
-    // layer, all nodes in the same layer must be preserved until that layer is
-    // finished.  We can allocate one layer at a time, but the best case is reducing the
-    // size of the bitset by about a half, at the cost of invoking CUDA malloc many more
-    // times than necessary.
     split_cats_.resize(node_categorical_storage_size_);
     h_split_cats_.resize(node_categorical_storage_size_);
     dh::safe_cuda(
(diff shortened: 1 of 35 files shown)


src/tree/gpu_hist/evaluator.cu
```diff
@@ -30,46 +30,45 @@ void GPUHistEvaluator<GradientSumT>::Reset(common::HistogramCuts const &cuts,
...
-      // size of the bitset by about a half, at the cost of invoking CUDA malloc many more
-      // times than necessary.
-      split_cats_.resize(param.MaxNodes() * bit_storage_size);
-      h_split_cats_.resize(split_cats_.size());
-      dh::safe_cuda(
-          cudaMemsetAsync(split_cats_.data().get(), '\0', split_cats_.size() * sizeof(CatST)));
+    node_categorical_storage_size_ =
+        common::CatBitField::ComputeStorageSize(cuts.MaxCategory() + 1);
+    CHECK_NE(node_categorical_storage_size_, 0);
+    // We need to allocate for all nodes since the updater can grow the tree layer by
```
trivialfis, line 47: Does this comment still hold? "all nodes"

src/tree/updater_gpu_hist.cu
```diff
@@ -567,27 +565,12 @@ struct GPUHistMakerDevice {
...
-        split_cats.resize(common::CatBitField::ComputeStorageSize(max_cat + 1), 0);
-        CHECK_LE(split_cats.size(), h_cats.size());
-        std::copy(h_cats.data(), h_cats.data() + split_cats.size(), split_cats.data());
-
-        node_categories.resize(candidate.split.split_cats.Bits().size());
-        dh::safe_cuda(cudaMemcpyAsync(
-            node_categories.data().get(), candidate.split.split_cats.Data(),
-            candidate.split.split_cats.Bits().size_bytes(), cudaMemcpyDeviceToDevice));
-      }
+      CHECK_GT(candidate.split.split_cats.Bits().size(), 0);
```
trivialfis, line 568: Nice cleanup!

PR #7862, opened 2022-05-05
updated winning solution under readme.md

amitalokbera:
Added one more winning solution which uses XGBoost


trivialfis (approved):
Thank you for the update!

PR #7863, opened 2022-05-05
Parameter typing

amotzop:
* Added script to automatically generate a `TypedDict` for the possible Booster parameters for the python API from the doc-strings.
* Added some typing to the file `training.py`.


amotzop:
I'm not sure how to integrate this script into the package generation process.
If someone can direct me to the relevant place I would be grateful.

trivialfis (commented):
Thank you for taking another look at the parameters. I still don't think it's necessary to have a typed list for parameters in native API.  If you want to have type checking for parameters, scikit-learn interface might be a better choice.

amotzop:
I think that if we want to have types for the parameters it's better to be consistent with it's use.
Also, there are some cases where one has to use the native API, and for those cases I believe it helps to have the right typing.

trivialfis:
> Also, there are some cases where one has to use the native API, and for those cases I believe it helps to have the right typing.

For training, the scikit-learn interface should be sufficient. For other cases, one can always get back the booster by using `get_booster()` member function.

trivialfis:
Thank you for working on this. As mentioned previously, I don't think we need a list of typed parameter for native API. For future reference, please open an issue for major changes to discuss the prospect of the planned features.

PR #7864, opened 2022-05-05
Allow loading old models from RDS

### LINKED ISSUE #7860: [RFC] Support reading old models from RDS, on best effort basis
Note. the discussion is solely discussing the use of RDS in the context of R programming. However, many points will apply to Python pickle as well.

## Motivation

In the past, we issued the following annoucement to tell R users to avoid using RDS to save models:

> Do not use **saveRDS** or **save** for long-term archival of models. Instead, use **xgb.save** or **xgb.save.raw**.
> It is a common practice to use the built-in **saveRDS** function (or **save**) to persist R objects to the disk. While it is possible to persist xgb.Booster objects using **saveRDS**), it is not advisable to do so if the model is to be accessed in the future. If you train a model with the current version of XGBoost and persist it with **saveRDS**, the model is not guaranteed to be accessible in later releases of XGBoost. To ensure that your model can be accessed in future releases of XGBoost, use **xgb.save** or **xgb.save.raw** instead.

https://www.rdocumentation.org/packages/xgboost/versions/1.6.0.1/topics/a-compatibility-note-for-saveRDS-save

While we (XGBoost developers) still believe that it's best to use `xgb.save`, we realize that not everyone is able to use `xgb.save`. Third-party libraries such as `mlr3` and `caret` use XGBoost internally and save the resulting model as an RDS. It is not realistic to rewrite the code of these third-party libraries to use `xgb.save` instead of `saveRDS`.

To accommodate this common use case, we should try to support reading old models from RDS files.

## Current behavior
Currently, loading an old model (fit with an old version of XGBoost) from an RDS file often results in an error. (For example, models fit with XGBoost 1.1.0.)

## Proposed behavior
**We should ensure that old models (fit with an old version of XGBoost) should be accessible from the latest XGBoost, even when they come from RDS files.** We aim for graceful degradation, where some of the latest features of XGBoost remain unavailable if an old model is read from RDS.

Users will encounter the following warning about mismatched versions:

> This model was trained with an older version of XGBoost. Some features may not be available as a result. For example, you won't be able to use `xgb.config` to query the training hyperparameters.

Rest assured, users will be able to do everything that's important, such as `predict` and `xgb.dump`. The behavior is as if the user had used `xgb.save` and `xgb.load` instead.

## Proposed implementation
The XGBoost model object consists of two parts:
1. **Model parameters**, such as decision trees.
2. **Configuration**, such as hyperparameters and other internal objects

We guarantee backward compatibility for the first part, the model parameters. When users call `xgb.save`, only the model parameters gets saved, not the configuration. "Backward compatibility" means that users can later call `xgb.load` in later versions of XGBoost to access the model.

On the other hand, we don't guarantee backward compatibility for the second part, configuration. The configuration is closely tied to internals of XGBoost and is subject to frequent change. When users call `saveRDS` to save the model into an RDS file, both model parameters and configuration are saved. This is why users encounter errors when they call `readRDS` to read the old model. More precisely, a later version XGBoost cannot read a configuration from an older version of XGBoost.

To achieve the desired behavior, we should simply **ignore configuration if it cannot be read**. The logic would be as follows:
1. Each XGBoost model remembers the version of XGBoost with which it was trained.
2. The user calls `loadRDS`.
3. Let X = XGBoost version used by the model; and Y = XGBoost version currently running.
4. If X == Y, then all is good. Load both model parameters and configuration.
5. If X < Y, then load only the model parameters; ignore configuration. By ignoring the configuration, we avoid errors and users can access the old model.

When the configuration is ignored, some features of XGBoost will not be available. For example, all training hyperparameters will be discarded. Therefore, we throw a warning to the user about mismatched versions.

## Alternative implementations considered
Option 1. Drop configuration automatically after the model is fit.
This will disable such useful features as `xgb.config`.

Option 2. Try to load configuration, until it fails.
This option won't work if we make a substantial change in the configuration that goes beyond adding or removing a field.

@trivialfis

hcho3:
Replaces #7703

Closes #7860

@trivialfis Can we include this in 1.6.1 patch release?


R-package/tests/testthat/test_model_compatibility.R
```diff
@@ -95,16 +96,13 @@ test_that("Models from previous versions of XGBoost can be loaded", {
         run_booster_check(booster, name)
       }
     })
-    if (compareVersion(model_xgb_ver, '1.0.0.0') < 0) {
-      # Expect a C++ warning when a model was generated in version < 1.0.x
-      m <- grepl(paste0('.*Loading model from XGBoost < 1\\.0\\.0, consider saving it again for ',
+    if (is_rds && compareVersion(model_xgb_ver, '1.1.1.1') >= 0) {
+      # Expect a C++ warning when a model is loaded from RDS and it was generated by old XGBoost`
+      print(cpp_warning)
```
hcho3, line 101: Remove

src/learner.cc
```diff
@@ -406,8 +406,18 @@ class LearnerConfiguration : public Learner {
   }
 
   void LoadConfig(Json const& in) override {
+    // If configuration is loaded, ensure that the model came from the same version
     CHECK(IsA<Object>(in));
-    Version::Load(in);
+    auto origin_version = Version::Load(in);
+
+    if (!Version::Same(origin_version)) {
```
trivialfis, line 413: Could you please take a look into the `ModelMsg()` function? I would like to keep the difference between "model" and RDS serialization.

trivialfis:
> @trivialfis Can we include this in 1.6.1 patch release?

Is it necessary? I thought it's not a bug fix but an improvement?

hcho3:
> it's not a bug fix but an improvement?

You are right. I removed the blocking tag. This can be part of the upcoming release.

src/learner.cc
hcho3: I'm okay with just using ModelMsg() in the warnings.

hcho3: Replaced with `ModelMsg()`

trivialfis (approved):

Issue #7875, opened 2022-05-08
error: identifier "cudaEventWaitDefault" is undefined


jnoorman:
After a clean reinstall and recursive pull getting this error trying to build on ubuntu 20.04 for CUDA

[ 62%] Building CXX object src/CMakeFiles/objxgboost.dir/tree/updater_sync.cc.o
[ 63%] Building CXX object src/CMakeFiles/objxgboost.dir/__/rabit/src/allreduce_base.cc.o
[ 64%] Building CXX object src/CMakeFiles/objxgboost.dir/__/rabit/src/rabit_c_api.cc.o
[ 64%] Building CXX object src/CMakeFiles/objxgboost.dir/__/rabit/src/engine.cc.o
[ 65%] Building CUDA object src/CMakeFiles/objxgboost.dir/c_api/c_api.cu.o
[ 66%] Building CUDA object src/CMakeFiles/objxgboost.dir/common/common.cu.o
[ 67%] Building CUDA object src/CMakeFiles/objxgboost.dir/common/device_helpers.cu.o
/xgboost/src/c_api/../data/../common/device_helpers.cuh(1610): error: identifier "cudaEventWaitDefault" is undefined

/xgboost/src/common/device_helpers.cuh(1610): error: identifier "cudaEventWaitDefault" is undefined

1 error detected in the compilation of "/tmp/tmpxft_0002514b_00000000-12_device_helpers.compute_75.cpp1.ii".
make[2]: *** [src/CMakeFiles/objxgboost.dir/build.make:934: src/CMakeFiles/objxgboost.dir/common/device_helpers.cu.o] Error 1
make[2]: *** Waiting for unfinished jobs....
1 error detected in the compilation of "/tmp/tmpxft_00025132_00000000-12_c_api.compute_75.cpp1.ii".
make[2]: *** [src/CMakeFiles/objxgboost.dir/build.make:908: src/CMakeFiles/objxgboost.dir/c_api/c_api.cu.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:292: src/CMakeFiles/objxgboost.dir/all] Error 2
make: *** [Makefile:152: all] Error 2

trivialfis:
XGBoost requires CUDA 11.x.

PR #7876, opened 2022-05-08
Update CUDA version requirement in CMake script.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/7875 .


hcho3 (approved):
My bad for missing this bit. Thanks!

PR #7892, opened 2022-05-12
[breaking] Drop single precision hist.

trivialfis:
Follow up for https://github.com/dmlc/xgboost/pull/7828 .


tests/cpp/tree/hist/test_histogram.cc
```diff
@@ -172,7 +167,7 @@ void TestSyncHist(bool is_distributed) {
   n_ids[1] = 5;
   for (size_t i : n_ids) {
     auto this_hist = histogram.Histogram()[i];
-    GradientSumT *p_hist = reinterpret_cast<GradientSumT *>(this_hist.data());
+    double *p_hist = reinterpret_cast<double_t *>(this_hist.data());
```
hcho3, line 170:
```suggestion
    double *p_hist = reinterpret_cast<double *>(this_hist.data());
```

hcho3 (approved):

tests/cpp/tree/hist/test_histogram.cc
trivialfis: Ah, auto completion ...

Issue #7890, opened 2022-05-12
Quantile DMatrix for CPU.


trivialfis:
This is part of https://github.com/dmlc/xgboost/issues/7690 to reduce memory usage of hist tree method.

- [x] Implement Quantile DMatrix for CPU. (https://github.com/dmlc/xgboost/pull/8130)
- [x] Implement reference DMatrix. (https://github.com/dmlc/xgboost/pull/8130)
- [x] Interoperability between CPU and GPU. (https://github.com/dmlc/xgboost/pull/8136 https://github.com/dmlc/xgboost/pull/8215)
- [x] Make it default in sklearn interface. (https://github.com/dmlc/xgboost/pull/8220)

PR #7897, opened 2022-05-13
[WIP] Remove bin compression in column matrix.

trivialfis:
Memory usage is slightly increased, but makes https://github.com/dmlc/xgboost/issues/7890 easier to implement as we don't need to dispatch the data type in other places like predictor.

- n_samples: 1e6
- n_features: 32

- todo: profile result.

### DIFF
--- a/src/common/column_matrix.h
+++ b/src/common/column_matrix.h
@@ -31,10 +31,7 @@ enum ColumnType : uint8_t { kDenseColumn, kSparseColumn };
     to reduce the memory usage. */
 class Column {
  public:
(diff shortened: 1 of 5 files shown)


trivialfis:
The previous data is incorrect.  The memory usage increases significantly when the data is close to dense.

|        | 0.95  | 0.5       | 0.05     | 0         |
|:-------|:------|:----------|:---------|:----------|
| PR     | 9.9MB | 1707.2MB  | 3777.3MB | 1891.5 MB |
| Master | 9.9MB | 1534.0 MB | 2811.9MB | 1822.6 MB |

PR #7901, opened 2022-05-14
Fix and cleanup for column matrix.

trivialfis:
* Fix missed type dispatching for dense columns with missing values.
* Code cleanup to reduce special cases.
* Reduce memory usage.

For 1e6 * 32 sparse matrix with sparsity at 0.05 (very close to dense), the memory usage is reduced from 1.1 GB to 870MB.


trivialfis:
Screenshot from 2022-05-15 01-23-46
Screenshot from 2022-05-15 01-24-07

RAMitchell (approved):

PR #6127, opened 2020-09-17
Predict improvement

ShvetsKS:
the main idea of this PR is keeping cache locality for predicted rows.

Results:

Data set | predict time (master), s | predict time (this PR), s | Speedup
-- | -- | -- | --
 higgs1m | 4.12 | 1.32 | 3.12
 msrank |  13.9 | 6.07 |  2.29

without Dmatrix create:
Data set | predict time (master), s | predict time (this PR), s | Speedup
-- | -- | -- | --
 higgs1m | 4.12 - 0.25 | 1.32 - 0.25 | 3.62
 msrank |  13.9 - 2.6 | 6.07 - 2.6 |  3.25

Data set | logloss (master) | logloss (this PR) 
-- | -- | -- 
 higgs1m | 0.4036675 | 0.4036675 
 msrank |  0.8550638 | 0.8550638

PR #7902, opened 2022-05-14
Remove tree structure assumption in GetNextNode

drorspei:
There is an assumption in predictor/predict_fn.h:GetNextNode that right children are always exactly one after their matching left children.

This isn't enforced elsewhere in the library, and can be false when loading trees from external sources such as json.

This pull request changes the single line that makes the assumption with a line that doesn't make the assumption.

I've also opened an issue regarding the current incorrect logic here: #7903

### DIFF
--- a/src/predictor/predict_fn.h
+++ b/src/predictor/predict_fn.h
@@ -22,7 +22,7 @@ inline XGBOOST_DEVICE bst_node_t GetNextNode(const RegTree::Node &node, const bs
                  ? node.LeftChild()
                  : node.RightChild();
     } else {
-      return node.LeftChild() + !(fvalue < node.SplitCond());
+      return  fvalue < node.SplitCond() ? node.LeftChild() : node.RightChild();
     }
   }
 }


src/predictor/predict_fn.h
```diff
@@ -22,7 +22,7 @@ inline XGBOOST_DEVICE bst_node_t GetNextNode(const RegTree::Node &node, const bs
                  ? node.LeftChild()
                  : node.RightChild();
     } else {
-      return node.LeftChild() + !(fvalue < node.SplitCond());
+      return  fvalue < node.SplitCond() ? node.LeftChild() : node.RightChild();
```
hcho3, line 25: This change will lead to degraded performance and therefore cannot be accepted. Assuming that the right child node is +1 offset from the left child saves one function call and leads to up to 35% performance improvement (https://github.com/dmlc/xgboost/pull/6127).

In general, users are able to generate arbitrary JSON payload that will lead to undefined behavior when loaded by XGBoost. We cannot protect against all possible failure modes arising from this use case. Users should only load from JSON files that originate from the only trusted source: `Booster.save_model()`.

hcho3 (changes requested):

hcho3:
@trivialfis Shall we add a warning in the doc about loading JSON from an external source?

trivialfis:
Yes, please do.

src/predictor/predict_fn.h
drorspei: Cool cool
I'll note that I don't think the 35% is over this version, but over an older version that is a lot different in terms of output assembly.
There's also an answered question by hcho3 in the pull request regarding the performance of exactly the change in this pull request.
Basically on any modern compiler and architecture the output assembly of this line should be a conditional move and not a branch jump.

trivialfis:
@ShvetsKS Could you please take a look into this when you are available?

ShvetsKS:
@drorspei you can reproduce degraded performance with simple run: "python xgboost_hist_method_bench.py --dataset higgs1m", from benchmark repo https://github.com/dmlc/xgboost-bench
**master**:  XGBOOST predict (train data) = 1.2422 sec
**current PR**: XGBOOST predict (train data) = 2.0260 sec

Also I have to note for 'hist' method there is a [check](https://github.com/dmlc/xgboost/blob/master/src/tree/updater_quantile_hist.cc#L378) to control expected behaviour

hcho3:
I will file a separate PR to document the danger of loading JSON from an external source.

PR #7904, opened 2022-05-14
Rename misspelled config parameter for pseudo-Huber

hcho3:
Follow-up to https://github.com/dmlc/xgboost/pull/7727.

This is a breaking change in the config. The impact is limited as https://github.com/dmlc/xgboost/pull/7864 disabled loading config for old models.

### DIFF
--- a/doc/model.schema
+++ b/doc/model.schema
@@ -207,7 +207,7 @@
         }
       }
     },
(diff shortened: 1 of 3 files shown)


trivialfis (approved):
Thank you for the fix. My mistake ..

PR #7905, opened 2022-05-15
Correct a mistake in Setting Parameters section

### LINKED ISSUE #7878: Is there a mistake in the Python Package Introduction?
In the [Setting Parameters](https://xgboost.readthedocs.io/en/stable/python/python_intro.html#setting-parameters) section of the Python Package Introduction of XGBoost documentation, the code for specifing validations set to watch performance goes like this:

`evallist = [(dtest, 'eval'), (dtrain, 'train')]`

I followed this code and found early stopping rounds did not work, as it was targeting the train score rather than the eval score. It seems that the order should be reversed:

`evallist = [(dtrain, 'train'), (dtest, 'eval')]`

SzeYeung1:
Users who follow the code of specify validations set ( evallist = [(dtest, 'eval'), (dtrain, 'train')] ) in the Setting Parameters section will make the train score rather than the eval score as the criteria of early stopping, therefore propose reverse the order as 、evallist = [(dtrain, 'train'), (dtest, 'eval')]、.

Closes #7878

### DIFF
--- a/doc/python/python_intro.rst
+++ b/doc/python/python_intro.rst
@@ -147,7 +147,7 @@ XGBoost can use either a list of pairs or a dictionary to set :doc:`parameters <
 
   .. code-block:: python
 
-    evallist = [(dtest, 'eval'), (dtrain, 'train')]
+    evallist = [(dtrain, 'train'), (dtest, 'eval')]
 
 Training
 --------


hcho3:
Thanks!

PR #7912, opened 2022-05-17
Use adapter to initialize column matrix.

trivialfis:
* Add an adapter for SparsePage. Might seem weird, but we already have something similar in GPU code and predictor. We can unify the code further by treating SparsePage as an ordinary adapter batch.
* Implement initilaization for column matrix from adapter.
* More precise tests.

### DIFF
--- a/src/common/column_matrix.h
+++ b/src/common/column_matrix.h
@@ -303,7 +303,7 @@ class ColumnMatrix {
         auto line = batch.GetLine(rid);
         for (size_t i = 0; i < line.Size(); ++i) {
           auto coo = line.GetElement(i);
-          if (data::IsValidFunctor{missing}(coo)) {
+          if (data::IsValidFunctor {missing}(coo)) {
             auto fid = coo.column_idx;
             const uint32_t bin_id = row_index[k];
             get_bin_idx(bin_id, rid, fid);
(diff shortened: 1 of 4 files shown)


trivialfis:
This is to implement https://github.com/dmlc/xgboost/issues/7890 .

src/common/column_matrix.h
```diff
@@ -286,22 +297,18 @@ class ColumnMatrix {
...
-        for (size_t i = ibegin; i < iend; ++i, ++j) {
-          const uint32_t bin_id = row_index[i];
-          auto fid = inst[j].index;
-          get_bin_idx(bin_id, rid, fid);
+        auto line = batch.GetLine(rid);
+        for (size_t i = 0; i < line.Size(); ++i) {
+          auto coo = line.GetElement(i);
+          if (data::IsValidFunctor {missing}(coo)) {
+            auto fid = coo.column_idx;
+            const uint32_t bin_id = row_index[k];
```
hcho3, line 308: What's the performance implication of abstracting the element access? Is the overhead acceptable?

hcho3: Also,  wouldn't it be inefficient to create the functor in a tight-loop? Let's create the functor instance outside the loop.

trivialfis: > Is the overhead acceptable?

It's not the bottleneck. The major issue with this function is the `missing_flags_` that's a boolean vector, which is not thread-safe. https://github.com/dmlc/xgboost/pull/7208 is likely to rewrite many of these anyway, right now I just want to make sure some of the features are merged before over optimization.

trivialfis: > right now I just want to make sure some of the features are merged before optimization.

I'm not against optimization (and like them), just trying to show a big picture of what we need in these data structures before we go nuts with optimization.

hcho3 (approved):

PR #7919, opened 2022-05-19
Do not initialise data structures to maximum possible tree size.

RAMitchell:
This change results in significant performance improvements when the tree is sparse, i.e. the user specifies some large max depth, but final tree size is significantly smaller than the maximum possible size.

Performance is otherwise unchanged.

Fraud dataset at max_depth=22 with gpu_hist:
before - 18.4165s
after - 1.7270s

### DIFF
--- a/src/tree/split_evaluator.h
+++ b/src/tree/split_evaluator.h
@@ -13,6 +13,7 @@
 #include <utility>
 #include <vector>
 #include <limits>
+#include <algorithm>
 
 #include "xgboost/tree_model.h"
 #include "xgboost/host_device_vector.h"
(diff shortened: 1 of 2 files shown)


src/tree/split_evaluator.h
```diff
@@ -157,6 +159,15 @@ class TreeEvaluator {
     if (!has_constraint_) {
       return;
     }
+
+    auto max_nidx = std::max(leftid, rightid);
+    if (lower_bounds_.Size() <= max_nidx) {
```
trivialfis, line 164: Is the memory correctly initialized?

trivialfis (approved):

PR #7921, opened 2022-05-19
VM-Packages: Auto-detection of MUSL is replaced by system properties

### LINKED ISSUE #7915: Enhance detection of MUSL-based Linux systems (xgboost4j)
### Context

With the release 1.6 of xgboost, a [feature](https://github.com/dmlc/xgboost/pull/7624) was added that would automatically detect under Linux, whether it is a regular or a MUSL-based system. Depending on this, the proper Native shared library would be loaded (MUSL-based systems cannot load shared libraries using `glibc`).

The check whether a Linux system is MUSL-based is done by inspecting memory-mapped files in `/proc/self/map_files`. This is the same logic as the sqllite-jdbc project is doing [here](https://github.com/xerial/sqlite-jdbc/blob/ada2b147dc858ef557a995e2f4ef25f7c65fc6a7/src/main/java/org/sqlite/util/OSInfo.java#L115-L131).

### Problem

While this typically works well, there are cases, where this directory is not accessible. There are at least two cases:
* since `/proc/self/map_files` was added in Linux kernel 3.3, for older kernels this directory simply does not exist
* some versions of the kernel do not allow access to this directory for non-root users

This problem has also been discussed on the sqlite-jdbc forums. See [here](https://github.com/xerial/sqlite-jdbc/issues/706#issuecomment-1015191818) for a good summary of the problem.

The [current implementation in xgboost](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java#L102-L106) assumes, that if this directory cannot be read for some reason, the Linux system is not MUSL-based.

__Unfortunately, this assumption is not always correct.__

### Proposal
We need another way to determine whether a Linux system is MUSL-based or not. It seems, there is no simple auto-detection, that works 100% for all cases and environments. Therefore, I propose that the user can give a hint to xgboost, whether the Linux system is MUSL-based or not.

(1)
In the class [NativeLibraryLoader](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java#L102-L106) we check for the presence of a custom environment variable like `IS_MUSL`. If this env variable is present and its value is truthy (`true` or `1`), xgboost assumes the system is MUSL-based, no matter whether memory-mapped files in `/proc/self/map_files` can be read or not.

(2)
If this custom environment variable is not present, then detection falls back to the current solution.

This could be simplified by defining that you always have to configure the `IS_MUSL` environment variable, if you're running on a MUSL-based system. But since it seems to work most of the time, it still seems convenient for most users to keep the current detection mechanism.

### How does this work for the end user?

* If you're running xgboost on a regular (glibc-based) Linux system, there's nothing to consider.
* If you're running xgboost on a MUSL-based Linux system and auto-detection detects this correctly, there's nothing to consider.
* If you're running xgboost on a MUSL-based Linux system and auto-detection fails to detect this, you'll get an ugly failure when xgboost is trying to load the shared library. The error message should then include a clue, that you may want to use the `IS_MUSL` environment variable to override this.

Please let me know what you think! I can create a PR with the suggested change.

snoopcheri:
This PR removes auto-detection of MUSL-based Linux systems in favor of system properties the user can set to configure a specific path for a native library.

At the moment, there's only a single library with the name `xgboost4j` that we need. A user who wants to configure a specific path, has to set the system property `xgboostruntime.native.xgboost4j`. If this system property is present, its value is used as a library path inside the JAR.

A user who runs xgboost4j on a MUSL-based Linux system, will get an additional error message `Alternatively, if your Linux OS is musl-based, you should set the path for the native library xgboost4j via the system property xgboostruntime.native.xgboost4j`.

Resolves #7915

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java
@@ -31,6 +31,8 @@ import java.util.stream.Stream;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import static ml.dmlc.xgboost4j.java.NativeLibLoader.LibraryPathProvider.getLibraryPathFor;
+
 /**
  * class to load native library
  *
@@ -149,8 +151,19 @@ class NativeLibLoader {
     }
   }
 
+  static class LibraryPathProvider {
+
+    private static final String nativeResourcePath = "/lib";
+
+    static String getLibraryPathFor(OS os, Arch arch, String libName) {
+      return nativeResourcePath + "/" +
+        getPlatformFor(os, arch) + "/" +
+        System.mapLibraryName(libName);
+    }
+
+  }
+
   private static boolean initialized = false;
-  private static final String nativeResourcePath = "/lib";
   private static final String[] libNames = new String[]{"xgboost4j"};
 
   /**
@@ -168,16 +181,14 @@ class NativeLibLoader {
     if (!initialized) {
       OS os = OS.detectOS();
       Arch arch = Arch.detectArch();
-      String platform = os.name + "/" + arch.name;
       for (String libName : libNames) {
         try {
-          String libraryPathInJar = nativeResourcePath + "/" +
-              platform + "/" + System.mapLibraryName(libName);
+          String libraryPathInJar = getLibraryPathFor(os, arch, libName);
           loadLibraryFromJar(libraryPathInJar);
         } catch (UnsatisfiedLinkError ule) {
           String failureMessageIncludingOpenMPHint = "Failed to load " + libName + " " +
               "due to missing native dependencies for " +
-              "platform " + platform + ", " +
+              "platform " + getPlatformFor(os, arch) + ", " +
               "this is likely due to a missing OpenMP dependency";
 
           switch (os) {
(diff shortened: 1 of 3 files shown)


jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/NativeLibLoader.java
```diff
@@ -149,8 +108,43 @@ static Arch detectArch() {
     }
   }
 
+  /**
+   * Utility class to determine the path of a native library.
+   */
+  static class LibraryPathProvider {
```
Craigacp, line 114: Is there a reason why this is a separate class rather than adding these fields & methods to `NativeLibLoader`?

snoopcheri: I think it has a clear purpose, reduces the scope for `nativeResourcePath` and `customNativeLibraryPathPropertyPrefix` and aligns nicely with the corresponding test.

snoopcheri: @Craigacp Is this OK with you?

Craigacp: I think spinning a class file for two fields and two methods isn't worth the effort, but it also doesn't really matter either way. The new test file needs a license header, and aside from that everything looks ok.

snoopcheri: Good point! I've added the license info.

(b4954f81)

trivialfis (approved):
LGTM! Thank you for working on this.

PR #7925, opened 2022-05-20
Batch UpdatePosition calls

RAMitchell:
Also removes position information from inside RowPartitioner class. We don't actually need it anywhere.

FinalisePosition has been moved outside the class, and now calculates positions as well as leaf values, which can be used for UpdatePredictionCache.

I have removed prediction caching for external memory in gpu_hist.


RAMitchell:
Some thoughts on further re-factoring:

- Updaters should not see the task. Their behaviour should not depend on the objective function.
- The prediction cache update is perhaps best returned from the tree->Update call alongside position information. It is very easy to calculate the prediction when you are calculating the row positions. Alternatively GBTree can just calculate this using the positions obtained from the update.
- Our lives would be made much easier if every updater consistently supported position information and prediction caching, without exceptions.
- The gradient based sampling needs to transparently show which samples it has selected to enable the above for gpu_hist.

RAMitchell:
Single GPU benchmarks, max_depth=20

| dataset | master      | batch-position-scan |
|---------|-------------|---------------------|
| airline | 5014.992213 |         4659.965927 |
| bosch   | 94.40516737 |          92.6324375 |
| covtype | 190.9750907 |         162.9997489 |
| epsilon | 986.6671753 |         979.8350237 |
| fraud   | 1.618839191 |         1.577903566 |
| higgs   | 2577.564037 |         2177.161882 |
| year    | 2251.341695 |         2081.802971 |

RAMitchell:
There is a performance regression for airline dataset (many rows) with max_depth=8. One of the batched kernels shows poor performance and needs to be profiled/optimised.

Issue #5444, opened 2020-03-27
Consistency of min_child_weight parameter


RAMitchell:
The `min_child_weight` parameter (default value 1.0) has different effects based on scaling of objective functions. I noticed this when developing a new objective function that had a small Hessian and the tree was not able to grow with default parameters. Objectives like squared error and logistic loss will be regularised very differently as a consequence. For example using logistic loss where the hessian values can be much smaller, it can require a much larger number of training instances to split. In #2483 it is noted that the hessian in the case of logistic loss is proportional to variance, however this is not true of other objectives in general.

This is relevant to the task of finding good default parameters across a range of objectives (#4986).

One obvious solution is normalising all objective functions in some consistent way.

Another solution is deprecating min_child_weight and moving to a parameter like min_child_instances, regularising based on the amount of training data without respect to the objective function.

@trivialfis has also proposed implementing multiclass objective functions via vector leaves, if we do this the hessian will be a vector and it is not obvious how to correctly apply min_child_weight.

QuantHao:
I think one problem of replace min_child_weight with min_child_instances is that: how to deal with sample weight? A good questions is raised and answered at [a LightGBM issue](https://github.com/microsoft/LightGBM/issues/2870).

PR #7932, opened 2022-05-24
[WIP] Implement `min_child_samples` for `hist` and `approx`.

trivialfis:
Related: https://github.com/dmlc/xgboost/issues/5444

I will work on `gpu_hist` as a follow-up, some more changes are still pending for the evaluation function for categorical data support.


trivialfis:
The implementation is similar to LightGBM with hessian as an estimation of counts. But it has its drawbacks: https://github.com/microsoft/LightGBM/issues/3634 https://github.com/microsoft/LightGBM/issues/5236

RAMitchell:
I think the criticism of this heuristic is quite valid, it makes the result unpredictable. The assumption that the hessian is uniformly distributed seems dubious. I think the only reliable way to calculate this is unfortunately to include it in the histogram calculation.

RAMitchell:
After working on the row partitioning I realized that if we have count information it makes this step much faster. So if we counted rows in the histogram we lose some time, but gain some back in row partitioning.

PR #7934, opened 2022-05-24
Handle missing values in one hot splits.

trivialfis:


trivialfis:
The test is reused in GPU as well.

RAMitchell (approved):

PR #7937, opened 2022-05-25
Unify the cat split storage for CPU.

trivialfis:
Use the bitset inside the split entry for both OHE split and partition-based split.


hcho3 (approved):

hcho3:
Merging this now to unblock #7943

PR #7938, opened 2022-05-25
[jvm-packages] remove default parameters

wbo4958:
This PR is to remove the most booster default parameters, which will be defaulted by xgboost native.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -166,7 +166,8 @@ private[this] class XGBoostExecutionParamsFactory(rawParams: Map[String, Any], s
   def buildXGBRuntimeParams: XGBoostExecutionParams = {
     val nWorkers = overridedParams("num_workers").asInstanceOf[Int]
     val round = overridedParams("num_round").asInstanceOf[Int]
-    val useExternalMemory = overridedParams("use_external_memory").asInstanceOf[Boolean]
+    val useExternalMemory = overridedParams
+      .getOrElse("use_external_memory", false).asInstanceOf[Boolean]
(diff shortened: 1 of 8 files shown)


wbo4958:
@trivialfis could you help to check the python test CI?

trivialfis:
Please ignore it for now. The hypothesis package is updated and caught the error where xgboost failed to parse some parameters in double precision.

trivialfis (commented):
Could you please set this parameter to true https://github.com/dmlc/xgboost/blob/755d9d460924ab0afdf287d6ff3a01ae701a657a/include/xgboost/generic_parameters.h#L39 and check if there's any warning during training? If not then you can pass it from the scala package to enable the validation.

wbo4958:
> Could you please set this parameter to true
> 
> https://github.com/dmlc/xgboost/blob/755d9d460924ab0afdf287d6ff3a01ae701a657a/include/xgboost/generic_parameters.h#L39
> 
> and check if there's any warning during training? If there is no warning then you can pass it from the scala package to enable the validation.

Yeah, there are some WARNINGs,

``` console
Parameters: { "allow_non_zero_for_missing", "dmlc_worker_connect_retry", "features_col", "handle_invalid", "label_col", "missing", "num_early_stopping_rounds", "num_round", "num_workers", "prediction_col", "probability_col", "rabit_ring_reduce_threshold", "rabit_timeout", "raw_prediction_col", "silent", "tracker_conf", "train_test_ratio", "tree_limit", "use_external_memory" } might not be used.
```

Most parameters unused are coming from the JVM parameter default values, and we should delete these parameters before passing into native. I also have filed an item (AKA an followup) to unify these parameters.

trivialfis:
Hmm, rabit timeout is useful.

trivialfis (approved):

trivialfis:
> Hmm, rabit timeout is useful.

I take it back. It's not useful for XGBoost training, this should be removed after rabit initialization.

trivialfis:
@wbo4958 Could you please rebase the branch to master?

PR #7940, opened 2022-05-25
[jvm-packages] add format option when saving a model

wbo4958:
This PR adds a format option which allows the booster to be dumped into json or ubj.

For now, we still use "deprecated" format as the default when dumping the booster model, but we're able to choose the different booster formats like json, ubj to be dumped.

eg,

``` scala
model.write.option("format", "json").save(modelJsonPath)
```

or 

``` scala
model.write.option("format", "ubj").save(modelJsonPath)
```

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/utils/XGBoostReadWrite.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/utils/XGBoostReadWrite.scala
@@ -16,12 +16,16 @@
 
 package ml.dmlc.xgboost4j.scala.spark.utils
 
+import ml.dmlc.xgboost4j.java.{Booster => JBooster}
+
 import org.apache.spark.ml.util.MLWriter
 
 private[spark] abstract class XGBoostWriter extends MLWriter {
 
+  /** Currently it's using the "deprecated" format as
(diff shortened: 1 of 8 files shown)


wbo4958:
@trivialfis please help to review it.

trivialfis (commented):
Thank you for the work on the file format. Could you please update the document: https://github.com/dmlc/xgboost/blob/master/doc/tutorials/saving_model.rst ?

wbo4958:
Yeah, I will add it in the followup PR. Thx

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/utils/XGBoostReadWrite.scala
```diff
@@ -0,0 +1,31 @@
...
+package ml.dmlc.xgboost4j.scala.spark.utils
+
+import ml.dmlc.xgboost4j.java.{Booster => JBooster}
+
+import org.apache.spark.ml.util.MLWriter
+
+private[spark] abstract class XGBoostWriter extends MLWriter {
+
+  /** Currently it's using the "deprecated" format as
+   * default, which will be changed into `ubj` in future releases. */
```
trivialfis, line 26: changed into `json` in future releases.

trivialfis (approved):

PR #7945, opened 2022-05-26
Fix incomplete type hints for `verbose`

### LINKED ISSUE #7933: Scikit-Learn/Dask API: incorrect/incomplete type hint and documentation for `verbose`
The `fit()` method of all classes in the python scikit-learn and dask API describe the `verbose` parameter as:
> verbose (Optional[bool]) – If verbose and an evaluation set is used, writes the evaluation metric measured on the validation set to stderr.

However, `verbose` is set as `verbose_eval` in the Learning API ([example for XGBModel](https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/sklearn.py#L1002)), which also allows an integer value in order to adjust the frequency of printing evaluation metrics:
> verbose_eval (Optional[Union[bool, int]]) – Requires at least one item in evals. If verbose_eval is True then the evaluation metric on the validation set is printed at each boosting stage. If verbose_eval is an integer then the evaluation metric on the validation set is printed at every given verbose_eval boosting stage. The last boosting stage / the boosting stage found by using early_stopping_rounds is also printed. Example: with verbose_eval=4 and at least one item in evals, an evaluation metric is printed every 4 boosting stages, instead of every boosting stage.

Given that there isn't any other use of `verbose` in the scikit-learn or dask API apart from passing it to `verbose_eval` that would prevent the use of an integer value for the purpose mentioned above, I propose to update the type hint and documentation of `verbose` to match the description of `verbose_eval`.

tsabsch:
Fixes #7933


python-package/xgboost/sklearn.py
```diff
@@ -938,8 +938,11 @@ def fit(
                 Use `early_stopping_rounds` in :py:meth:`__init__` or
                 :py:meth:`set_params` instead.
         verbose :
-            If `verbose` and an evaluation set is used, writes the evaluation metric
-            measured on the validation set to stderr.
+            If `verbose` is True and an evaluation set is used, the evaluation metric
+            measured on the validation set is printed to stderr at each boosting stage.
```
trivialfis, line 942: It's printed to stdout.

trivialfis (changes requested):
Thank you for the fix! One small issue with the doc.

python-package/xgboost/sklearn.py
tsabsch: thx, I didn't verify this myself. Updated now for both occurrences.

trivialfis (approved):

Issue #7906, opened 2022-05-15
[RFC] Making R interface more idiomatic


david-cortes:
I notice that there is a version 2.0 of xgboost in the plans, which among other things, is expected to include support for categorical features in the R interface.

Given that this is a major version release and as such is expected to introduce potentially breaking changes, I think this is a good opportunity to make the R interface more in line with base R and core/popular R modeling packages. Many people (including myself) find the R interface of xgboost to be inconvenient and unidiomatic, but changing the interface for `xgboost()` from its current state would be a rather big breaking change and would probably break lots of user scripts that depend on `xgboost()`.

In short, `xgboost()` does not work with the most common data types used in R (`data.frame`) and does not follow R conventions in terms of e.g. function arguments. For people who are familiar with base R and with other R packages, there are many ways in which the R interface of xgboost could be improved for a better end-user experience, such as:

* Offering an x/y interface as well as a formula interface.
* Accepting data frames as inputs and handling categorical/factor variables from data frames.
* Accepting `factor` variables as "y".
* Accepting non-standard evaluation for column names (e.g. passing the weight variable as a column name without quotes).
* Using base-1 numeration for integers as R does instead of base-0.
* Controlling prediction types through a `type` argument.
* Making the naming of function arguments more consistent with base R and core packages - for example, naming the weights as `weights` instead of `weight`, like base R does.
* Changing default arguments by, for example, not dumping the model to a file in disk by default.

Among many others.

Would this project accept big breaking PRs for the R interface (particularly for `xgboost()` and `predict.xgb.Booster()`) for the 2.0 release that would make it more similar to base R and other R packages?

RAMitchell:
I don't think any current active maintainers are big R users so we welcome input. Could we just build a new interface behind a different namespace until it's ready? I don't think there's a need to immediate replace the old interface in a short space of time.

trivialfis:
> Would this project accept big breaking PRs for the R interface (particularly for xgboost() and predict.xgb.Booster()) for the 2.0 release that would make it more similar to base R and other R packages?

I would like to welcome these changes. The concern about breaking changes can be handled by running reverse dependency checks.

mayer79:
I suggest to keep `xgboost()` as it is and instead call the new function differently, e.g. `xgboost2()`, or use a slim method dispatch with `xgboost.formula()`. The former has the advantage to accompany it with a corresponding `predict2()` function. Otherwise, great work @david-cortes.

PR #7947, opened 2022-05-26
[R] Use `type` argument to control prediction types

david-cortes:
ref #7906 

In R packages for statistical modeling, the `predict` functions for model objects tend to allow switching between different prediction types such as probabilities or classes through a parameter `type` (for example, base R's `predict.glm` does it like that).

`xgboost` offers several mutually exclusive parameters to control prediction types (`outputmargin`, `precontrib`, `predinteraction`, `predleaf`). This PR merges them all into a single `type` parameter in line with base R and other popular packages, adding an option to predict classes for classification objectives.

Note however that objective `multi:softmax` is problematic in this scenario, since it doesn't allow switching between probabilities and classes on the fly, so an exception about it is mentioned in the docs.

Right now the code logic is to check the objective and determine if and how it should determine the predicted class from the outputs, but:
* I wasn't able to find any C-level function to extract the booster's objective, and am not sure that the way is done in this PR is ideal.
* It currently lists a set of objectives which allow class prediction from the original output, but I think a better approach would be to check if the objective starts with `binary` or `multi` - problem with such approach right now is that `multi:softmax` doesn't output probabilities.

Disclaimer: this is likely to cause several breakages in R packages that depend on `xgboost`.

### DIFF
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -188,7 +188,9 @@ xgb.Booster.complete <- function(object, saveraw = TRUE) {
 #'             \item "contrib" will return feature contributions to individual predictions (see Details).
 #'             \item "interaction" will return contributions of feature interactions to individual
 #'                   predictions (see Details).
-#' }
+#'             }
+#'
+#'             Note that, when using custom objective functions, "link" and "response" will default to "raw".
 #' @param missing Missing is only used when input is dense matrix. Pick a float value that represents
 #'        missing values in data (e.g., sometimes 0 or some other extreme value is used).
 #' @param ntreelimit Deprecated, use \code{iterationrange} instead.
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -167,19 +167,35 @@ xgb.Booster.complete <- function(object, saveraw = TRUE) {
 #'
 #'        For single-row predictions on sparse data, it's recommended to use CSR format. If passing
 #'        a sparse vector, it will take it as a row vector.
+#' @param type Type of prediction to make:\itemize{
+#'             \item "link" will return the predicted values according to the objective function being optimized for
+#'                   (after applying the corresponding link function) - for example, for
+#'                   \code{objective="binary:logistic"}, it will return predicted probabilities, while for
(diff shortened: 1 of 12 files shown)


trivialfis:
Thank you for the work on aligning XGBoost with the R ecosystem. I will take a deeper look later to see if it's possible to keep the compatibility with some deprecation warnings. Otherwise we will have to run a reverse dependency check and talk to other projects before cran submission for the next release.

trivialfis:
@hetong007 Would you great if you can help review the changes.

hetong007:
In terms of the potential breaks, may I suggest that (at least in the next release) we keep both the new argument `type` and the old set of arguments? In this case, users can work with `predict` with either way. Just when users predict with the old arguments, `predict` throws out a deprecation warning.

Moreover, by setting a higher priority to `type` we resolve the case where the new and the old are conflicting.

david-cortes:
Added deprecation warnings and acceptance of the old argument types.

R-package/R/xgb.Booster.R
```diff
@@ -375,23 +395,78 @@ predict.xgb.Booster <- function(object, newdata, missing = NA, outputmargin = FA
...
+    old_args_for_type <- list(
+      "outputmargin" = "margin",
+      "predleaf" = "leaf",
+      "predcontrib" = "contrib",
+      "predinteraction" = "interaction"
+    )
+
+    for (arg in names(old_args_for_type)) {
+      if (arg %in% extra_argnames) {
+        warning(
```
hetong007, line 412: My only concern with `warning` is that it might break CRAN checks of reverse dependencies. We need to talk to their maintainers before pushing this change to CRAN.

david-cortes: It most definitly will. Good thing is, next version is going to be a 2.0, which signals a major release and thus potentially breaking changes.

hetong007: In concept for general development this is true, while for CRAN it still regardlessly breaks reverse dependencies and got Ripleyed.

mayer79: Changing the interface of `predict()` will definitively cause chaos, not only on CRAN.

david-cortes:
Renamed the types to be more in line with base R.

PR #7948, opened 2022-05-27
Handle missing categorical value in CPU evaluator.

trivialfis:


src/tree/hist/evaluate_splits.h
```diff
@@ -119,13 +119,90 @@ class HistEvaluator {
...
+   * missing values is resided on that partition:
+   *
+   *   | Forward  | Backward |
+   *   |----------+----------|
+   *   | A [BCDE] | [ABCD] E |
+   *   | AB [CDE] | [ABC] DE |
+   *   | ABC [DE] | [AB] CDE |
+   *   | ABCD [E] | [A] BCDE |
+   */
+  template <int d_step>
```
hcho3, line 138: Add comment: in this function, d_step=+1 indicates the backward direction, and d_step=-1 indicates the forward direction.

hcho3 (approved):
I like how the logic for partition is now cleanly separated from the numerical enumeration. LGTM.

src/tree/hist/evaluate_splits.h
trivialfis: Actually +1 is the forward, it's just a convention from one hot split that the gradient is being added to the right node, which might be confusing.

hcho3: In that case, switch the two columns in the example table.

trivialfis: Done.

hcho3 (approved):

PR #7950, opened 2022-05-27
[CI] Use private mirror to host llvm-openmp tarballs

hcho3:
Currently, Mac CD pipelines are failing with error
```
CondaHTTPError: HTTP 403 FORBIDDEN for url <https://anaconda.org/conda-forge/llvm-openmp/11.1.0/download/osx-64/llvm-openmp-11.1.0-hda6cdc1_1.tar.bz2>
```
Curiously, I was able to manually download from the same URL. It could be that GitHub Actions workers are being throttled by Conda-forge.

To fix this error, host the same file on our own private mirror (an S3 bucket).


trivialfis:
Can we try the latest libomp instead of this frozen version?

hcho3:
@trivialfis This workaround was introduced fairly recently, to avoid segfaults due to mixing of two versions of libomp from Conda and Homebrew. Not sure if anything has changed yet.

PR #7952, opened 2022-05-30
set log level as ERROR when trackerProcess has some stderr output

yangjiandan:
Generally, in order to avoid too many spark logs, we usually do
`val sc = new SparkContext(conf)
       sc.setLogLevel("WARN")`

Using XGBoostClassifier by Spark Application, there are nothing error logs when lacking python3 in cluster environment, 
and we can see the following log when set log info level.
I think it is more useful to find error when loging stderr using error level.
> 22/05/23 15:11:02 WARN XGBoostSpark: train_test_ratio is deprecated since XGBoost 0.82, we recommend to explicitly pass a training and multiple evaluation datasets by passing 'eval_sets' and 'eval_set_names'
22/05/23 15:11:03 INFO RabitTracker$TrackerProcessLogger:   File "/mnt/sata05/hadoop/yarn/local/usercache/ubd_obx_test/appcache/application_1653287501496_0002/container_e29_1653287501496_0002_01_000001/tmp/tracker4901044343178211783.py", line 24
22/05/23 15:11:03 INFO RabitTracker$TrackerProcessLogger:     def __init__(self, sock: socket.socket) -> None:
22/05/23 15:11:03 INFO RabitTracker$TrackerProcessLogger:                            ^
22/05/23 15:11:03 INFO RabitTracker$TrackerProcessLogger: SyntaxError: invalid syntax
22/05/23 15:11:03 INFO RabitTracker$TrackerProcessLogger: Tracker Process ends with exit code 1
22/05/23 15:11:03 INFO SparkContext: Starting job: collect at XGBoost.scala:431
22/05/23 15:11:03 INFO DAGScheduler: Registering RDD 149 (repartition at DataUtils.scala:96) as input to shuffle 9
22/05/23 15:11:03 INFO DAGScheduler: Got job 24 (collect at XGBoost.scala:431) with 1 output partitions
22/05/23 15:11:03 INFO DAGScheduler: Final stage: ResultStage 34 (collect at XGBoost.scala:431)
22/05/23 15:11:03 INFO DAGScheduler: Parents of final stage: List(ShuffleMapStage 33)
22/05/23 15:11:03 INFO DAGScheduler: Missing parents: List(ShuffleMapStage 33)
22/05/23 15:11:03 INFO DAGScheduler: Submitting ShuffleMapStage 33 (MapPartitionsRDD[149] at repartition at DataUtils.scala:96), which has no missing parents

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/RabitTracker.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/RabitTracker.java
@@ -58,8 +58,12 @@ public class RabitTracker implements IRabitTracker {
           trackerProcessLogger.error(line);
         }
         trackerProcess.get().waitFor();
-        trackerProcessLogger.info("Tracker Process ends with exit code " +
-                trackerProcess.get().exitValue());
+        int exitValue = trackerProcess.get().exitValue();
+        if (exitValue != 0) {
+          trackerProcessLogger.error("Tracker Process ends with exit code " + exitValue);
+        } else {
+          trackerProcessLogger.info("Tracker Process ends with exit code " + exitValue);
+        }
       } catch (IOException ex) {
         trackerProcessLogger.error(ex.toString());
       } catch (InterruptedException ie) {
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/RabitTracker.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/RabitTracker.java
@@ -55,7 +55,7 @@ public class RabitTracker implements IRabitTracker {
       String line;
       try {
         while ((line = reader.readLine()) != null) {
-          trackerProcessLogger.info(line);
+          trackerProcessLogger.error(line);
         }
         trackerProcess.get().waitFor();
         trackerProcessLogger.info("Tracker Process ends with exit code " +


yangjiandan:
@superbobry could you help me to reivew this patch, thanks.

trivialfis (commented):
That might not be the only possible warning emitted during tracker initialization. I think a better way forward for us is to have a unified tracker implementation instead of launching Python process inside the JVM runtime.

yangjiandan:
> That might not be the only possible warning emitted during tracker initialization. I think a better way forward for us is to have a unified tracker implementation instead of launching Python process inside the JVM runtime.

I do agree with you. Unified tracker implementation is a long term plan. Could we first fix known emitted warning to avoid other encounter the same problem.
I look through `RabitTracker`, and do not find other possible warning emitted, could you give me some hints.

yangjiandan:
> That might not be the only possible warning emitted during tracker initialization. I think a better way forward for us is to have a unified tracker implementation instead of launching Python process inside the JVM runtime.

I do agree with you. Unified tracker implementation is a long term plan. Could we first fix known emitted warning to avoid other encounter the same problem.
I look through `RabitTracker`, and do not find other possible warning emitted, could you give me some hints.

trivialfis:
None at the moment. But this logic means we can not add any in the future. Do you think turning a non-zero return code into an error might be a better solution?

yangjiandan:
> None at the moment. But this logic means we can not add any in the future. Do you think turning a non-zero return code into an error might be a better solution?

Thanks for your quickly response. I mean we could first fix current issue,  my colleague and I spent a whole day to find reason of spark application failure,  and finally found lacking python3 is the root cause. Do you have some solutions to resolve this.

yangjiandan:
If you don't think this patch is necessary，please feel free to close this mr.

yangjiandan:
> Maybe we can turn this into an error if the exit code is not zero?

yes, that is more reasonableI， I am newer to xgboost and can not understand the whole project. Thanks a lot for you opinion. I will update this patch later according to your.

yangjiandan:
@trivialfis do you have bandwidth to review my code again? please feel free to tell me if there are issues in my code.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/RabitTracker.java
```diff
@@ -55,11 +55,15 @@ public void run() {
       String line;
       try {
         while ((line = reader.readLine()) != null) {
-          trackerProcessLogger.info(line);
+          trackerProcessLogger.error(line);
```
trivialfis, line 58: Could you please remove this error since we have the exit code check now?

yangjiandan: return value is not enough for user to find root cause, maybe stderr is more helpful.  what do you think about?

yangjiandan: There is none code printing something to stderr in tracker.py, so it should have some error during executing tracker.py, I think it's reasonable to use error level during loging strerr message.

trivialfis: If this error is kept then the exit code validation is dead code and will never be run. We have to pick one of them.

yangjiandan: `trackerProcessLogger.error(line) just logging by ERROR level, it does not affect the rest logic of code`

trivialfis: My mistake. Look good  to me then.

trivialfis (approved):

PR #7956, opened 2022-05-31
[jvm-packages] re-write xgboost read/write

wbo4958:
1. Removed the duplicated Default XGBoost read/write code which is copied from
  spark 2.3.x
2. Put some utils into util package


wbo4958:
I also verified https://github.com/dmlc/xgboost/pull/7359#issuecomment-953591695

trivialfis (approved):

PR #7966, opened 2022-06-02
[jvm-packages] add jni for setting feature name and type

wbo4958:
This PR adds jni APIs for XGDMatrixGetStrFeatureInfo and XGDMatrixSetStrFeatureInfo to set/get feature information including feature_name and feature_type which will be typically used in categorical feature supporting. I will file a followup to support categorical feature in jvm-packages.


wbo4958:
@trivialfis please help to review it

trivialfis (approved):

PR #7992, opened 2022-06-14
[Breaking] Remove rabit support for custom reductions and `grow_local_histmaker` updater

rongou:
Right now cpu tree methods rely on `rabit::Reducer` to perform custom reductions, but in reality they are just sums of gradients and hessians. Neither NCCL nor the new federated learning gRPC supports custom reductions, so removing this would move us closer to a common interface.

Additionally `rabit::SerializeReducer` supports reduction on arbitrary data structure, but it's only used by `updater_histmaker.cc`, which is no longer being referenced anywhere in the code base, so both are removed.


rongou:
@trivialfis @RAMitchell 

Not sure if this is the best approach, please take a look and let me know what you think. Next step I need to do some deeper refactoring to get rid of `rabit::SerializeReducer`.

rongou:
@trivialfis @RAMitchell turns out `rabit::SerializeReducer` is only used by `updaters_histmaker.cc`, which doesn't seem to be in use any more, so removed both.

I'm not sure why the R tests are failing. Is it caused by this change?

trivialfis:
The R test is unrelated. I'm trying to unblock the CI.

trivialfis:
I'm happy to remove the hist maker. It's old code to implement global_approx and local_approx. I rewrote the former but the latter was kept in this hist_maker. We don't usually use it but I mentioned it in the doc https://github.com/dmlc/xgboost/blob/master/doc/treemethod.rst#approximated-solutions so it's a breaking change non the less. Please annotate the PR with `[breaking]` and mention the change in the title so that we can include the change in release news.

rongou:
@trivialfis done.

trivialfis (approved):
LGTM. Please rebase to master branch.

Issue #7995, opened 2022-06-15
[CI] Disable modin test for now.


trivialfis:
Issue for tracking error raised from modin and ray along with latest pandas: https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-7960/5/pipeline/180/ .

PR #7996, opened 2022-06-15
Disable modin tests for now.

trivialfis:
https://github.com/dmlc/xgboost/issues/7995


trivialfis:
Seems to be working now.

PR #8001, opened 2022-06-16
Use 22.04 for RMM.

trivialfis:
22.06 is not released yet.


trivialfis:
Merged to remove cmake error in CI.

Issue #8016, opened 2022-06-21
[JVM-packages] Executor crashes when running CrossValidator on XGBoost4j-spark-GPU


wbo4958:
When setting the task GPU to be fractional, the XGBoost CrossValidor job will fail with the failure that the executor crashes with SIGSEGV error.

After debugging with @trivialfis (credits to jiaming), it turned out the API `XGBoosterSetParam` is not thread-safe. Dring the phase of transforming, XGBoost will run several transforming tasks at a time, and each of them will set the "gpu_id" and "predictor" parameters, so if several tasks (multi-threads) call  `XGBoosterSetParam` simultaneously, it may cause the memory to be corrupted and cause SIGSEGV.

``` console
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
        set = {__val = {18446744067266837215, 140643996305479, 140642338238808, 2000, 140642338239008, 140643872133392, 140642338238967, 140643872141344, 3, 2251803447816704, 12884901888, 7756222594670153728,
            140642338238808, 7756222594670153728, 37, 622203}}
        pid = <optimized out>
        tid = <optimized out>
        ret = <optimized out>
#1  0x00007fea3b651859 in __GI_abort () at abort.c:79
        save_stage = 1
        act = {__sigaction_handler = {sa_handler = 0x0, sa_sigaction = 0x0}, sa_mask = {__val = {140643996521704, 140642338239312, 140643996009913, 140643996008416, 0, 140638704107520, 140643997435552,
              140643997418656, 140643995948022, 140643992509728, 140643871492960, 140642338239200, 140643990474580, 1, 140642338239568, 140643994956076}}, sa_flags = 310912640, sa_restorer = 0x7fe9d89c7300}
        sigs = {__val = {32, 0 <repeats 15 times>}}
#2  0x00007fea3a9dbf1f in ?? () from /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/server/libjvm.so
No symbol table info available.
#3  0x00007fea3b31d143 in ?? () from /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/server/libjvm.so
No symbol table info available.
#4  0x00007fea3b16ed30 in JVM_handle_linux_signal () from /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/server/libjvm.so
No symbol table info available.
[...]

PR #8017, opened 2022-06-21
[jvm-packages] fix executor crashing issue when transforming on xgboost4j-spark-gpu

wbo4958:
the API XGBoosterSetParam is not thread-safe. Dring the phase of transforming,
XGBoost runs several transforming tasks at a time, and each of them will set
the "gpu_id" and "predictor" parameters, so if several tasks (multi-threads)
all XGBoosterSetParam simultaneously, it may cause the memory to be corrupted
and cause SIGSEGV.

This PR first gets the booster from broadcast and set to the correct gpu_id
and predictor, and then all transforming tasks will use the same booster to
do the transforming.

To fix https://github.com/dmlc/xgboost/issues/8016


jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
```diff
@@ -331,26 +330,26 @@ class XGBoostClassificationModel private[ml](
     }
   }
 
-  private[scala] def producePredictionItrs(broadcastBooster: Broadcast[Booster], dm: DMatrix):
+  private[scala] def producePredictionItrs(broadcastBooster: Booster, dm: DMatrix):
```
trivialfis, line 333: the name `broadcastBooster` can be changed to `booster`.

PR #8020, opened 2022-06-21
PySpark XGBoost integration

WeichenXu123:

PySpark XGBoost integration.

```
class SparkXGBRegressor(_SparkXGBEstimator):
    """
    SparkXGBRegressor is a PySpark ML estimator. It implements the XGBoost regression
    algorithm based on XGBoost python library, and it can be used in PySpark Pipeline
    and PySpark ML meta algorithms like CrossValidator/TrainValidationSplit/OneVsRest.

    SparkXGBRegressor automatically supports most of the parameters in
    `xgboost.XGBRegressor` constructor and most of the parameters used in
    `xgboost.XGBRegressor` fit and predict method (see `API docs <https://xgboost.readthedocs\
    .io/en/latest/python/python_api.html#xgboost.XGBRegressor>`_ for details).

    SparkXGBRegressor doesn't support setting `gpu_id` but support another param `use_gpu`,
    see doc below for more details.

    SparkXGBRegressor doesn't support setting `base_margin` explicitly as well, but support
    another param called `base_margin_col`. see doc below for more details.

    SparkXGBRegressor doesn't support following params:
      `gpu_id`, `enable_categorical`, `use_label_encoder`, `n_jobs`, `nthread`,
      `validate_features`, `output_margin`, `base_margin` param.

    callbacks:
        The export and import of the callback functions are at best effort.
        For details, see :py:attr:`xgboost.spark.SparkXGBRegressor.callbacks` param doc.
    missing:
        The parameter `missing` in SparkXGBRegressor has different semantics with
        that in `xgboost.XGBRegressor`. For details, see
        :py:attr:`xgboost.spark.SparkXGBRegressor.missing` param doc.
    validationIndicatorCol
        For params related to `xgboost.XGBRegressor` training
        with evaluation dataset's supervision, set
        :py:attr:`xgboost.spark.SparkXGBRegressor.validationIndicatorCol`
        parameter instead of setting the `eval_set` parameter in `xgboost.XGBRegressor`
        fit method.
    weightCol:
        To specify the weight of the training and validation dataset, set
        :py:attr:`xgboost.spark.SparkXGBRegressor.weightCol` parameter instead of setting
        `sample_weight` and `sample_weight_eval_set` parameter in `xgboost.XGBRegressor`
        fit method.
    xgb_model:
        Set the value to be the instance returned by
        :func:`xgboost.spark.SparkXGBRegressorModel.get_booster`.
    num_workers:
        Integer that specifies the number of XGBoost workers to use.
        Each XGBoost worker corresponds to one spark task.
    use_gpu:
        Boolean that specifies whether the executors are running on GPU
        instances.
    base_margin_col:
        To specify the base margins of the training and validation
        dataset, set :py:attr:`xgboost.spark.SparkXGBRegressor.base_margin_col` parameter
        instead of setting `base_margin` and `base_margin_eval_set` in the
        `xgboost.XGBRegressor` fit method. Note: this isn't available for distributed
        training.
```
```
    >>> from xgboost.spark import SparkXGBRegressor
    >>> from pyspark.ml.linalg import Vectors
    >>> df_train = spark.createDataFrame([
    ...     (Vectors.dense(1.0, 2.0, 3.0), 0, False, 1.0),
    ...     (Vectors.sparse(3, {1: 1.0, 2: 5.5}), 1, False, 2.0),
    ...     (Vectors.dense(4.0, 5.0, 6.0), 2, True, 1.0),
    ...     (Vectors.sparse(3, {1: 6.0, 2: 7.5}), 3, True, 2.0),
    ... ], ["features", "label", "isVal", "weight"])
    >>> df_test = spark.createDataFrame([
    ...     (Vectors.dense(1.0, 2.0, 3.0), ),
    ...     (Vectors.sparse(3, {1: 1.0, 2: 5.5}), )
    ... ], ["features"])
    >>> xgb_regressor = SparkXGBRegressor(max_depth=5, missing=0.0,
    ... validation_indicator_col='isVal', weight_col='weight',
    ... early_stopping_rounds=1, eval_metric='rmse')
    >>> xgb_reg_model = xgb_regressor.fit(df_train)
    >>> xgb_reg_model.transform(df_test)
```

```
class SparkXGBClassifier(_SparkXGBEstimator, HasProbabilityCol, HasRawPredictionCol):
    """
    SparkXGBClassifier is a PySpark ML estimator. It implements the XGBoost classification
    algorithm based on XGBoost python library, and it can be used in PySpark Pipeline
    and PySpark ML meta algorithms like CrossValidator/TrainValidationSplit/OneVsRest.

    SparkXGBClassifier automatically supports most of the parameters in
    `xgboost.XGBClassifier` constructor and most of the parameters used in
    `xgboost.XGBClassifier` fit and predict method (see `API docs <https://xgboost.readthedocs\
    .io/en/latest/python/python_api.html#xgboost.XGBClassifier>`_ for details).

    SparkXGBClassifier doesn't support setting `gpu_id` but support another param `use_gpu`,
    see doc below for more details.

    SparkXGBClassifier doesn't support setting `base_margin` explicitly as well, but support
    another param called `base_margin_col`. see doc below for more details.

    SparkXGBClassifier doesn't support setting `output_margin`, but we can get output margin
    from the raw prediction column. See `rawPredictionCol` param doc below for more details.

    SparkXGBClassifier doesn't support `validate_features` and `output_margin` param.

    Parameters
    ----------
    callbacks:
        The export and import of the callback functions are at best effort. For
        details, see :py:attr:`xgboost.spark.SparkXGBClassifier.callbacks` param doc.
    missing:
        The parameter `missing` in SparkXGBClassifier has different semantics with
        that in `xgboost.XGBClassifier`. For details, see
        :py:attr:`xgboost.spark.SparkXGBClassifier.missing` param doc.
    rawPredictionCol:
        The `output_margin=True` is implicitly supported by the
        `rawPredictionCol` output column, which is always returned with the predicted margin
        values.
    validationIndicatorCol:
        For params related to `xgboost.XGBClassifier` training with
        evaluation dataset's supervision,
        set :py:attr:`xgboost.spark.SparkXGBClassifier.validationIndicatorCol`
        parameter instead of setting the `eval_set` parameter in `xgboost.XGBClassifier`
        fit method.
    weightCol:
        To specify the weight of the training and validation dataset, set
        :py:attr:`xgboost.spark.SparkXGBClassifier.weightCol` parameter instead of setting
        `sample_weight` and `sample_weight_eval_set` parameter in `xgboost.XGBClassifier`
        fit method.
    xgb_model:
        Set the value to be the instance returned by
        :func:`xgboost.spark.SparkXGBClassifierModel.get_booster`.
    num_workers:
        Integer that specifies the number of XGBoost workers to use.
        Each XGBoost worker corresponds to one spark task.
    use_gpu:
        Boolean that specifies whether the executors are running on GPU
        instances.
    base_margin_col:
        To specify the base margins of the training and validation
        dataset, set :py:attr:`xgboost.spark.SparkXGBClassifier.base_margin_col` parameter
        instead of setting `base_margin` and `base_margin_eval_set` in the
        `xgboost.XGBClassifier` fit method. Note: this isn't available for distributed
        training.
```
```
    >>> from xgboost.spark import SparkXGBClassifier
    >>> from pyspark.ml.linalg import Vectors
    >>> df_train = spark.createDataFrame([
    ...     (Vectors.dense(1.0, 2.0, 3.0), 0, False, 1.0),
    ...     (Vectors.sparse(3, {1: 1.0, 2: 5.5}), 1, False, 2.0),
    ...     (Vectors.dense(4.0, 5.0, 6.0), 0, True, 1.0),
    ...     (Vectors.sparse(3, {1: 6.0, 2: 7.5}), 1, True, 2.0),
    ... ], ["features", "label", "isVal", "weight"])
    >>> df_test = spark.createDataFrame([
    ...     (Vectors.dense(1.0, 2.0, 3.0), ),
    ... ], ["features"])
    >>> xgb_classifier = SparkXGBClassifier(max_depth=5, missing=0.0,
    ...     validation_indicator_col='isVal', weight_col='weight',
    ...     early_stopping_rounds=1, eval_metric='logloss')
    >>> xgb_clf_model = xgb_classifier.fit(df_train)
    >>> xgb_clf_model.transform(df_test).show()
```

Summarize some important follow-ups:

1. Apply DMatrix data iterator interface
2. model summary / explanation
3. Classifier uses CPU functions like softmax, how to address it ?
4. Support XGB random forest classifer / regressor
5. Support XGB ranker
6. Support PySpark SparseVector (we need wait pyspark 3.4 out)
7. Update tests, remove unittest dependency.
8. Add type annotations for xgboost.spark APIs
9. Add demo/guide-python and doc/tutorials
10. Fix the macOS github action on python xgboost.spark.

I will take 3 / 6 / 7 / 9 / 10, other items I would like to assign them to other community guys.


python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,810 @@
...
+            params["objective"] = "multi:softprob"
+            params["num_class"] = num_classes
+        else:
+            params["objective"] = "reg:squarederror"
+
+        if self.getOrDefault(self.use_gpu):
+            params["tree_method"] = "gpu_hist"
+            # TODO: fix this. This only works on databricks runtime.
+            #  On open-source spark, we need get the gpu id from the task allocated gpu resources.
+            params["gpu_id"] = 0
```
WeichenXu123, line 343: An important TODO

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,810 @@
...
+import pandas as pd
+from scipy.special import expit, softmax
+from pyspark.ml import Estimator, Model
+from pyspark.ml.param.shared import HasFeaturesCol, HasLabelCol, HasWeightCol, \
+    HasPredictionCol, HasProbabilityCol, HasRawPredictionCol, HasValidationIndicatorCol
+from pyspark.ml.param import Param, Params, TypeConverters
+from pyspark.ml.util import MLReadable, MLWritable
+from pyspark.sql.functions import col, pandas_udf, countDistinct, struct
+from pyspark.sql.types import ArrayType, FloatType
+from xgboost import XGBClassifier, XGBRegressor
```
WeichenXu123, line 14: @trivialfis 
In PySpark-XGBoost integration, the historical code we reuse some API from `xgboost.sklearn` (e.g. the XGBClassifier, XGBRegressor),
on second thought, I think it is better to make the implementation only rely on the xgboost core API, what do you think ?

python-package/xgboost/spark/__init__.py
```diff
@@ -0,0 +1,16 @@
+"""XGBoost: eXtreme Gradient Boosting library.
+
+Contributors: https://github.com/dmlc/xgboost/blob/master/CONTRIBUTORS.md
```
trivialfis, line 3: This can be replaced by something more specific to pyspark interface.

python-package/xgboost/spark/data.py
```diff
@@ -0,0 +1,262 @@
+import os
+from typing import Iterator
+import numpy as np
+import pandas as pd
+from scipy.sparse import csr_matrix
+from xgboost import DMatrix
+
+
+# Since sklearn's SVM converter doesn't address weights, this one does address weights:
+def _dump_libsvm(features, labels, weights=None, external_storage_precision=5):
```
trivialfis, line 10: I'm not entirely sure why do you need to dump data into libsvm. If it's about external memory, we should go with data iterator instead.

python-package/xgboost/spark/data.py
```diff
@@ -0,0 +1,262 @@
...
+        if train and has_validation:
+            pdf_val = pdf.loc[pdf["validationIndicator"], :]
+            pdf = pdf.loc[~pdf["validationIndicator"], :]
+
+        num_feature_dims = len(pdf["values"].values[0])
+
+        expected_feature_dims = _check_feature_dims(num_feature_dims,
+                                                    expected_feature_dims)
+
+        values_list.append(pdf["values"].to_list())
```
trivialfis, line 167: Not an issue that needs to be fixed in the initial PR, but I think the `to_list` is not necessary and we can use some more optimized procedures in pandas or numpy.

trivialfis (commented):
Thank you for getting this started! I haven't looked into the details yet, for the initial work let's get some simple things like code style aligned first to avoid massive changes later.

- Consider using the same document style as the rest of the library. (For instance, replace `:param:` notation).
- Consider using the `black` formatter.

Please add a simple demonstration in `demo/guide-python` so that we can run it and gain a better intuition on how things work. A tutorial in `doc/tutorials` is also extremely welcomed!

python-package/xgboost/spark/data.py
WeichenXu123: > dump data into libsvm

this is an option (turn off by default), this approach can reduce memory. If there's data iterator API, we don't need this.

> go with data iterator

In current code we don't use DMatrix data iterator (we didn't find related API ). Does xgboost python support DMatrix data iterator API now ?

python-package/xgboost/spark/core.py
wbo4958: Yeah, we can add it to the following PR

python-package/xgboost/spark/data.py
trivialfis: It supports a form of custom iterator for external memory: https://xgboost.readthedocs.io/en/stable/tutorials/external_memory.html#data-iterator . I will work on the `DeviceQuatnileDMatrix` support after this initial merge, which uses a very similar interface. But for the first PR, I don't think external memory support is necessary.

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,810 @@
...
+# Put pyspark specific params here, they won't be passed to XGBoost.
+# like `validationIndicatorCol`, `baseMarginCol`
+_pyspark_specific_params = [
+    'featuresCol', 'labelCol', 'weightCol', 'rawPredictionCol',
+    'predictionCol', 'probabilityCol', 'validationIndicatorCol'
+                                       'baseMarginCol'
+]
+
+_unsupported_xgb_params = [
+    'gpu_id',  # [ML-12862]
```
wbo4958, line 40: May I ask what "[ML-12862]" means?

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,810 @@
...
+    # [ML-12913], for classification, we can use rawPrediction as margin
+    'output_margin',
+    'validate_features',  # [ML-12923]
+    'base_margin'  # [ML-12689]
+}
+
+_created_params = {"num_workers", "use_gpu"}
+
+
+class _XgboostParams(HasFeaturesCol, HasLabelCol, HasWeightCol,
```
wbo4958, line 59: Could we unify the parameter format? seems we have both formats like "featuresCol" and "max_bin"

python-package/xgboost/spark/core.py
WeichenXu123: sorry that's wrong label, I will remove it.

wbo4958:
I was thinking how do we support categorical data in this PR? @trivialfis

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,810 @@
...
+            has_base_margin = True
+            select_cols.append(
+                col(self.getOrDefault(self.baseMarginCol)).alias("baseMargin"))
+
+        dataset = dataset.select(*select_cols)
+        # create local var `xgb_model_creator` to avoid pickle `self` object to remote worker
+        xgb_model_creator = self._get_xgb_model_creator()  # pylint: disable=E1111
+        fit_params = self._gen_fit_params_dict()
+
+        if self.getOrDefault(self.num_workers) > 1:
```
wbo4958, line 474: May I ask why needs to consider different situations for num_workers? Could we just use `_fit_distributed` ?

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,810 @@
...
+    def _repartition_needed(self, dataset):
+        """
+        We repartition the dataset if the number of workers is not equal to the number of
+        partitions. There is also a check to make sure there was "active partitioning"
+        where either Round Robin or Hash partitioning was actively used before this stage.
+        """
+        if self.getOrDefault(self.force_repartition):
+            return True
+        try:
+            num_partitions = dataset.rdd.getNumPartitions()
```
wbo4958, line 313: I guess if not forcing repartition, there may be some data skew issue.

python-package/xgboost/spark/core.py
WeichenXu123: These params `HasFeaturesCol, HasLabelCol, HasWeightCol` comes from reusing pyspark ML params. These param name has been defined as "featureCol" etc. in pyspark ML param module. We cannot change the param name.

But, other params are coming from xgboost library (e.g. max_depth, max_bin), shall we make them to be camel-Case style format ?

python-package/xgboost/spark/core.py
WeichenXu123: yes that's the reason we do repartitioning.

python-package/xgboost/spark/data.py
WeichenXu123: I marked a TODO there and we can address it in follow-up PR.

python-package/xgboost/spark/core.py
WeichenXu123: If num_workers == 1, we don't need to run rabit workers, and we can directly run xgboost single node routine. It would be faster. Is my understanding correct ?

tests/python/test_spark/xgboost_local_cluster_test.py
```diff
@@ -0,0 +1,465 @@
...
+
+    def test_regressor_distributed_basic(self):
+        regressor = XgboostRegressor(num_workers=self.n_workers, n_estimators=100)
+        model = regressor.fit(self.reg_df_train_distributed)
+        pred_result = model.transform(self.reg_df_test_distributed).collect()
+        for row in pred_result:
+            self.assertTrue(np.isclose(row.expected_label, row.prediction, atol=1e-3))
+
+    @unittest.skip
+    def test_check_use_gpu_param(self):
```
WeichenXu123, line 280: Q: shall I setup a CI test to run over GPU cluster ?

python-package/xgboost/spark/core.py
trivialfis: It should make no difference for the internal of XGBoost.

tests/python/test_spark/xgboost_local_cluster_test.py
trivialfis: Yes. Please place GPU related tests under tests/python-gpu/

python-package/xgboost/spark/core.py
wbo4958: Yeah, force_repartition is defaulting to false, so if the users run their application and feel pretty slow, how they know the performance issue is caused by "data skew" or "xgboost native training itself"

python-package/xgboost/spark/core.py
wbo4958: Looks like the RABIT tracker is running on the the same process as Python driver, If there is only one RABIT worker, IMO, it won't cause the performance issue. Instead, we need to maintain two "same" branches.

python-package/xgboost/spark/core.py
wbo4958: Yeah, IMO, we shouldn't change the parameters from xgboost library, since there is already a detailed description of the parameters, and better not to add another "same meaning" but "different format style" parameter.

``` python
class _XgboostParams(HasFeaturesCol, HasLabelCol, HasWeightCol,
                     HasPredictionCol, HasValidationIndicatorCol,
                     HasArbitraryParamsDict, HasBaseMarginCol):
```

I mean, could we re-define `HasFeaturesCol, HasLabelCol ...` in XGBoost pyspark wrapper instead of using the builit-in ones. I don't know if it makes sense.

python-package/xgboost/spark/utils.py
```diff
@@ -0,0 +1,191 @@
...
+    """Get or create spark session. Note: This function can only be invoked from driver side."""
+    if pyspark.TaskContext.get() is not None:
+        # This is a safety check.
+        raise RuntimeError(
+            "_get_spark_session should not be invoked from executor side."
+        )
+    return SparkSession.builder.getOrCreate()
+
+
+def _getConfBoolean(sqlContext, key, defaultValue):
```
wbo4958, line 146: _get_conf_boolean ?

python-package/xgboost/spark/utils.py
```diff
@@ -0,0 +1,191 @@
...
+def _get_rabit_args(context, n_workers):
+    """
+    Get rabit context arguments to send to each worker.
+    """
+    env = _start_tracker(context, n_workers)
+    rabit_args = [("%s=%s" % item).encode() for item in env.items()]
+    return rabit_args
+
+
+def get_host_ip(context):
```
wbo4958, line 116: _get_host_ip ?

python-package/xgboost/spark/utils.py
```diff
@@ -0,0 +1,191 @@
...
+        # Remove parameters without a default value and those in the unsupported_set
+        if (
+            parameter.default is not parameter.empty
+            and parameter.name not in unsupported_set
+        ):
+            filtered_params_dict[parameter.name] = parameter.default
+    return filtered_params_dict
+
+
+class HasArbitraryParamsDict(Params):
```
wbo4958, line 34: Could we add a separate params file to hold the customized Param?

python-package/xgboost/spark/__init__.py
```diff
@@ -0,0 +1,21 @@
+"""PySpark XGBoost integration interface
+"""
+
+try:
+    import pyspark
+except ImportError:
+    raise RuntimeError("xgboost spark python API requires pyspark package installed.")
+
+from .estimator import (
+    XgboostClassifier,
```
wbo4958, line 10: Could we prefix XGB which should be aligned with XGBoost other modules?

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,877 @@
...
+            .mapPartitions(lambda x: x)
+            .collect()[0][0]
+        )
+        result_xgb_model = self._convert_to_model(cloudpickle.loads(result_ser_booster))
+        return self._copyValues(self._create_pyspark_model(result_xgb_model))
+
+    def _fit(self, dataset):
+        self._validate_params()
+        # Unwrap the VectorUDT type column "feature" to 4 primitive columns:
+        # ['features.type', 'features.size', 'features.indices', 'features.values']
```
wbo4958, line 476: Looks like vector_to_array just converts the Vector to Arrays, so the comments here seem not to be used any more?

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,877 @@
...
+        features_col = col(self.getOrDefault(self.featuresCol))
+        label_col = col(self.getOrDefault(self.labelCol)).alias("label")
+        features_array_col = vector_to_array(features_col, dtype="float32").alias(
```
wbo4958, line 479: Hi @WeichenXu123, I was thinking the internal of this PR is actually using array col for features by converting the vector to the array, which means users must vectorize their features first, and then de-vectorize by this PR, looks like it's a little bit redundant. How about considering accepting any type of feature col. eg,

1. feature col is array type, we don't need to to anything
2. feature col is vector type, we just call vector_to_array
3. feature col is an array of numeric cols names (AKA, column-wise), we can create the array col by "functions.array"

python-package/xgboost/spark/core.py
WeichenXu123: This is an issue, what about turn on repartition by default ?

tests/python/test_spark/xgboost_local_cluster_test.py
WeichenXu123: @trivialfis Where can I see the GPU CI machine config ? (e.g. how many GPU we can use on the CI machine)

Ideally, my GPU test I hope it can run on a machine with 4 GPUs (I will setup a spark cluster with local-cluster mode and run spark jobs with 4 executors, each spark executor will be allocated with 1 GPU.)

python-package/xgboost/spark/core.py
WeichenXu123: yes make sense.
I will improve later.

python-package/xgboost/spark/core.py
WeichenXu123: This comment comes from supporting pyspark Sparse Vector (it is not included in this PR for now, but we will add it later), not relates to this PR. I will remove it.

python-package/xgboost/spark/core.py
WeichenXu123: > could we re-define HasFeaturesCol, HasLabelCol ...

This might add a lot of code, I prefer to keep reusing HasFeaturesCol.., what about in xgboost estimator we link "feature_col"  param name to "featureCol"  ?

tests/python/test_spark/xgboost_local_cluster_test.py
trivialfis: @hcho3 May have more insight into the CI infrastructure.

> y GPU test I hope it can run on a machine with 4 GPUs

The mgpu instance should be g4dn.12xlarge on aws, which comes with 4 GPUs.

trivialfis: @hcho3 May have more insight into the CI infrastructure.

> y GPU test I hope it can run on a machine with 4 GPUs

The mgpu instance should be g4dn.12xlarge on aws, which comes with 4 GPUs. You can mark the spark tests with `mgpu` pytest mark: https://github.com/dmlc/xgboost/blob/f470ad3af98e6632d81596c7241aa294b4a7648f/tests/python-gpu/conftest.py#L59 . https://github.com/dmlc/xgboost/blob/f470ad3af98e6632d81596c7241aa294b4a7648f/tests/pytest.ini#L3

WeichenXu123: Thank you! @trivialfis 
An additional question, how to build xgboost python package with GPU support on my local machine ? `python setup.py bdist_wheel` is the command correct ?

trivialfis: That command doesn't build with GPU support. If you want to build the wheel without installing XGBoost, then you need to do it in 2 steps: https://xgboost.readthedocs.io/en/stable/build.html#building-with-gpu-support .

``` sh
cd ./build
cmake -DUSE_CUDA=ON -DUSE_NCCL=ON /path/to/xgboost/
make -j$(nproc)
```

after which, you can create the binary wheel for Python:

```
cd python-package
python setup.py bdist_wheel
```

trivialfis: Since you are writing Python code, I think the 2 step approach might be desired. You can compile C++ code once then run
``` sh
pip install . -e
```
and you don't have to run any build procedure anymore as long as you are testing locally.

python-package/xgboost/spark/core.py
wbo4958: Yeah, I agree with you, let's default to true.

wbo4958:
another question, how to support the categorical data?

python-package/xgboost/spark/core.py
wbo4958: Could we just leave HasXXXCol here and hack them in the constructor? eg,

users pass features_col and then convert to featuresCol internal which should be a black box? Does that make sense?

trivialfis:
> another question, how to support the categorical data?

We will work on it later. I assume we will simply pickle the feature types to executors.

python-package/xgboost/spark/core.py
WeichenXu123: @wbo4958 
On second thought, we have `_repartition_needed` , when `force_repartition` is False, it will check whether the dataframe need repartition, if it find the dataset has the same partition number with num_worker and has a "repartition" operator as the last step of query plan, then it will skip repartition, if check failed it will still do repartition.

python-package/xgboost/spark/core.py
WeichenXu123: > feature col is array type, we don't need to to anything
Supported.
> feature col is vector type, we just call vector_to_array
Supported.
> feature col is an array of numeric cols names (AKA, column-wise), we can create the array col by "functions.array"
On second thought, I suggest user to use `pyspark.ml.transform.VectorAssembler` to assemble these columns into a vector column. This is a common patten in pyspark ML.
User can create a pyspark ML pipeline contains transforming stages (do assembling and other tranformations) and xgboost estimator stage.


About @trivialfis 's proposal:
> Prediction doesn't need data concatenation. This can also remove specializations in data processing procedures.

This I also suggest to use `pyspark.ml.transform.VectorAssembler`, and create pyspark ML pipeline so that it can ensure applying the same assembling transformation during training and prediction.

python-package/xgboost/spark/core.py
WeichenXu123: I already fixed it in this PR.

python-package/xgboost/spark/data.py
WeichenXu123: addressed in this PR.

WeichenXu123:
@trivialfis @wbo4958 

PR status:

> We need to filter out the spark-specific parameters before passing them into XGBoost train function: `Parameters: { "baseMarginCol", "force_repartition" } might not be used.`

Fixed.

> Prediction is thread-safe in XGBoost.

Removed wrong comment.

> Prediction doesn't need data concatenation. This can also remove specializations in data processing procedures.

Addressed.

> We use pytest, unittest.TestCase is not needed, Python2 compatibility code is not needed.

Minor issues, we can address it in follow-up PR.

> Not entirely sure about the extensive use of double. XGBoost returns f32 most of the time.

Now always convert feature values into float32 before training.

> Classifier uses CPU functions like softmax.

@trivialfis Could you elaborate using which one to replace it ? 

> xgb_model_creator unused for _fit_distributed.

Fixed.

> DMatrix construction should happen inside rabit context.

Fixed

> Model summary and Model explanation

Let's add them in follow-up PR.

> how to support the categorical data?

I already added "feature_types" param in this PR so that we can support it now.

> Use DMatrix data iterator API to reduce memory consumption

Let's do it in follow-up PR.

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,834 @@
...
+    )
+    force_repartition = Param(
+        Params._dummy(),
+        "force_repartition",
+        "A boolean variable. Set force_repartition=true if you "
+        + "want to force the input dataset to be repartitioned before XGBoost training."
+        + "Note: The auto repartitioning judgement is not fully accurate, so it is recommended"
+        + "to have force_repartition be True.",
+    )
+    feature_names = Param(
```
wbo4958, line 140: I am not sure if we need to define this parameter on the pyspark side, since both "feature_names" and "feature_types" are already in xgboost.

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,834 @@
...
+        Params._dummy(),
+        "feature_names",
+        "A list of str to specify feature names."
+    )
+    feature_types = Param(
```
wbo4958, line 145: same as "feature_names"

python-package/xgboost/spark/estimator.py
```diff
@@ -0,0 +1,208 @@
...
+    _SparkXGBEstimator,
+    SparkXGBClassifierModel,
+    SparkXGBRegressorModel,
+    _set_pyspark_xgb_cls_param_attrs,
+)
+
+
+class SparkXGBRegressor(_SparkXGBEstimator):
+    """
+    XgboostRegressor is a PySpark ML estimator. It implements the XGBoost regression
```
wbo4958, line 13: Seems the doc is not updated?

python-package/xgboost/spark/estimator.py
```diff
@@ -0,0 +1,208 @@
...
+    def _pyspark_model_cls(cls):
+        return SparkXGBRegressorModel
+
+
+_set_pyspark_xgb_cls_param_attrs(SparkXGBRegressor, SparkXGBRegressorModel)
+
+
+class SparkXGBClassifier(_SparkXGBEstimator, HasProbabilityCol, HasRawPredictionCol):
+    """
+    XgboostClassifier is a PySpark ML estimator. It implements the XGBoost classification
```
wbo4958, line 109: Seems the doc is not updated?

wbo4958:
Seems there are still mixed different python styles.

what I am thinking is hacking some special parameters, like SparkXGBClassifier only accepts like "features_col" instead of "featuresCol".

then adding some check in the constructor to set this parameters into the real paramter, for example,

``` python
            if k == "feature_col":
                self.setParams("featureCols", v)
```

I don't know if this makes sense.  if it is, we can add it in the following up.

python-package/xgboost/spark/__init__.py
```diff
@@ -0,0 +1,21 @@
+"""PySpark XGBoost integration interface
+"""
+
+try:
+    import pyspark
+except ImportError:
+    raise RuntimeError("xgboost spark python API requires pyspark package installed.")
```
trivialfis, line 7:
```suggestion
except ImportError as e:
    raise ImportError("pyspark package needs to be installed to use this module") from e
```

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,834 @@
...
+    _get_spark_session,
+)
+from .params import (
+    HasArbitraryParamsDict,
+    HasBaseMarginCol,
+)
+
+from pyspark.ml.functions import array_to_vector, vector_to_array
+from pyspark.sql.types import \
+    ArrayType, DoubleType, FloatType, IntegerType, LongType, ShortType
```
trivialfis, line 49:
```suggestion
from pyspark.sql.types import (
    ArrayType, DoubleType, FloatType, IntegerType, LongType, ShortType
)
```

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,834 @@
...
+        + "to have force_repartition be True.",
+    )
+    feature_names = Param(
+        Params._dummy(),
+        "feature_names",
+        "A list of str to specify feature names."
+    )
+    feature_types = Param(
+        Params._dummy(),
+        "feature_names",
```
trivialfis, line 147:
```suggestion
        "feature_types",
```

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,834 @@
...
+            return self._convert_to_classifier(booster)
+        else:
+            return None  # check if this else statement is needed.
+
+    def _query_plan_contains_valid_repartition(self, dataset):
+        """
+        Returns true if the latest element in the logical plan is a valid repartition
+        """
+        num_partitions = dataset.rdd.getNumPartitions()
+        query_plan = dataset._sc._jvm.PythonSQLUtils.explainString(
```
trivialfis, line 397: Could you please add some comments on how it works? Seems you are reaching to internal API and it would be great to have a brief note.

trivialfis (commented):
Overall looks good to me, some small suggestions in the comments. Could you please pass the tests on CI? Let us know if there's anything we can help.

For pylint, you can go to the source root and run `LINT_LANG=python make lint`

python-package/xgboost/spark/core.py
WeichenXu123: `feature_names` is not in xgboost.sklearn.XGBModel constructor params. So I need define it in pyspark side.
`feature_types` is in xgboost.sklearn.XGBModel constructor params, I remove it from pyspark param list.

WeichenXu123:
Found one DMatrix param `feature_weights` the PR haven't support it. Will update it later.

WeichenXu123:
Summarize some important follow-ups:

1) Apply DMatrix data iterator interface
2) model summary / explanation
3) Classifier uses CPU functions like softmax, how to address it ?
4) Support XGB random forest classifer / regressor
5) Support XGB ranker
6) Support PySpark SparseVector (we need wait pyspark 3.4 out)
7) Update tests, remove unittest dependency.

I will take 3 / 6 / 7, other items I would like to assign them to other community guys.

trivialfis:
@WeichenXu123 Thanks for the summary. Would you like to finish the initial setup for the GPU tests? @hcho3 is the maintainer of XGBoost's CI infrastructure.

python-package/xgboost/spark/core.py
wbo4958: Looks good to me. Sorry for overloop

wbo4958:
LGTM. thx @WeichenXu123

WeichenXu123:
@trivialfis @hcho3 
Because python tests now need to run pyspark, it need to setup JDK in python CI and set JAVA_HOME as well. How to setup java in the python CI properly ?

tests/python-gpu/conftest.py
```diff
@@ -58,5 +58,6 @@ def pytest_collection_modifyitems(config, items):
     # mark dask tests as `mgpu`.
     mgpu_mark = pytest.mark.mgpu
     for item in items:
-        if item.nodeid.startswith("python-gpu/test_gpu_with_dask.py"):
+        if item.nodeid.startswith("python-gpu/test_gpu_with_dask.py") or \
+                item.nodeid.startswith("python-gpu/test_spark_with_gpu/test_spark_with_gpu.py"):
```
WeichenXu123, line 62: @hcho3 Is this setting for GPU test correct ?
What I need is:
run this file `python-gpu/test_spark_with_gpu/test_spark_with_gpu.py` in a machine with 4 GPUs.

trivialfis: @WeichenXu123 Let me take a look tomorrow.

hcho3: @WeichenXu123 Yes, this setting is correct.

WeichenXu123:
@trivialfis One additional question, when call `booster.predict`, we cannot set "missing" param if `X` is not DMatrix type,
in this case, will xgboost use the training DMatrix missing value to run prediction ?

hcho3:
> what's the CI machine memory size for mgpu_mark ?

We use `g4dn.12xlarge` machine, which has 192 GB memory.

hcho3:
@WeichenXu123 I'm getting this error while running the PySpark test on my machine:
```
22/07/07 01:19:54 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
22/07/07 01:19:55 ERROR Utils: Failed to create directory /opt/python/envs/gpu_test/lib/python3.8/site-packages/pyspark/work
java.nio.file.AccessDeniedException: /opt/python/envs/gpu_test/lib/python3.8/site-packages/pyspark/work
```
Is it possible to set a different location for a temp directory?

Also, do we plan to support Windows? If not, let's disable PySpark tests on Windows.

WeichenXu123:
@hcho3 

I haven't test it on windows, I think we can do supporting windows in follow-up PR if needed.

I found CI test printing error like:
```
RuntimeError: Python in worker has different version 3.9 than that in driver 3.8, PySpark cannot run with different minor versions. Please check environment variables PYSPARK_PYTHON and PYSPARK_DRIVER_PYTHON are correctly set.
```

This means pyspark worker use a different python executable with pyspark driver, to fix this, we can add the 2 environment variables (make them point to the same python executable file path):
```
export PYSPARK_DRIVER_PYTHON=/path/to/python
export PYSPARK_PYTHON=/path/to/python
```

> Is it possible to set a different location for a temp directory?
I will update it.

WeichenXu123:
@hcho3 
> 22/07/07 01:19:54 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
22/07/07 01:19:55 ERROR Utils: Failed to create directory /opt/python/envs/gpu_test/lib/python3.8/site-packages/pyspark/work
java.nio.file.AccessDeniedException: /opt/python/envs/gpu_test/lib/python3.8/site-packages/pyspark/work

Would you provide stack trace for this ? 
I want to know which code raise the error but I haven't found it . My local test works fine.

tests/python/test_spark/utils_test.py
```diff
@@ -117,6 +121,10 @@ def setUpClass(cls):
             {
                 "spark.master": "local-cluster[2, 2, 1024]",
                 "spark.python.worker.reuse": "false",
+                "spark.driver.host": "127.0.0.1",
+                "spark.task.maxFailures": "1",
+                "spark.sql.execution.pyspark.udf.simplifiedTraceback.enabled": "false",
+                "spark.sql.pyspark.jvmStacktrace.enabled": "true",
```
WeichenXu123, line 129: These 2 configs prevent spark hiding scala side stacktrace in error message.

hcho3:
@WeichenXu123 See https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-8020/15/pipeline/183

WeichenXu123:
> @WeichenXu123 See https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-8020/15/pipeline/183

Got it.
This is directory created by pyspark executor.
If pyspark is installed by `pip install`, then it should be only used as "local" mode or client connecting to spark cluster.
My test case uses "local-cluster" mode, it need to create pyspark executor processes, we should download pyspark tallball to deploy it, instead of installing with "pip install pyspark"

Let me update testing code.

wbo4958:
``` console
22-07-07T06:43:08.018Z] tests/python-gpu/test_spark_with_gpu/test_spark_with_gpu.py::test_sparkxgb_classifier_with_gpu Setting default log level to "WARN".

[2022-07-07T06:43:08.018Z] To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).

[2022-07-07T06:43:08.018Z] 22/07/07 06:43:07 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable

[2022-07-07T06:43:08.673Z] 22/07/07 06:43:08 ERROR Worker: Failed to setup worker resources: 

[2022-07-07T06:43:08.673Z] org.apache.spark.SparkException: Resource script: test_spark_with_gpu/discover_gpu.sh to discover gpu doesn't exist!
```

failed to find the "discover_gpu.sh" @WeichenXu123

WeichenXu123:
> ```
> test_spark_with_gpu/discover_gpu.sh
> ```

@hcho3 
I put the file in `tests/python-gpu/test_spark_with_gpu/discover_gpu.sh`. 
How to get the repository root path in CI so that I can set the correct full path in spark config ?

WeichenXu123:
@hcho3 Are all CI issues addressed ? My local tests all passed. But my local environment is slightly different with CI.

wbo4958:
@WeichenXu123 what kind of CI issue here?

I also can pass the python gpu test locally, suppose it works

WeichenXu123:
> @WeichenXu123 what kind of CI issue here?
> 
> I also can pass the python gpu test locally, suppose it works

Could you approval all CI runs and see whether they can pass

wbo4958:
The CI for the newest PR is so far so good. But it failed for windows, that's ok for now, we can skip the windows platform for now.

WeichenXu123:
> The CI for the newest PR is so far so good. But it failed for windows, that's ok for now, we can skip the windows platform for now.

Yes. Could you help me disable run spark test on windows ? I am not sure how to modify the CI script.

trivialfis:
let me try.

python-package/xgboost/spark/core.py
```diff
@@ -0,0 +1,880 @@
+# type: ignore
+"""Xgboost pyspark integration submodule for core code."""
+# pylint: disable=import-error, consider-using-f-string, too-many-arguments, too-many-locals
+# pylint: disable=invalid-name, fixme, too-many-lines, unbalanced-tuple-unpacking, no-else-return
+# pylint: disable=protected-access, logging-fstring-interpolation, no-name-in-module
+# pylint: disable=wrong-import-order, ungrouped-imports, too-few-public-methods, broad-except
+# pylint: disable=too-many-statements
```
trivialfis, line 7: Please fix as many errors as possible and skip only if it conflicts with designs that cannot be fixed in near future. Some are easy like invalid-name and consider-using-f-string . Linter might be annoying but it's necessary.

wbo4958:
I simply add "--ignore-glob" in Jenkinsfile-win64, but seems the Jenkinsfile-win64 is not updated. see https://github.com/dmlc/xgboost/pull/8055/commits/26ed6025110c1899c6c7410b3f42302a32374daa

trivialfis:
Please fix the hang on github action: https://github.com/dmlc/xgboost/runs/7237846264?check_suite_focus=true .

Feel free to ignore other tests for now. modin test and categorical data tests are unrelated. I will fix them in a separate PR.

trivialfis:
> Shall we disable MacOS support for now?

I will leave that to @WeichenXu123 to decide. If we disable osx tests then we need to add a check inside the spark module to raise an error when it's being run on windows/osx.

WeichenXu123:
> > Shall we disable MacOS support for now?
> 
> I will leave that to @WeichenXu123 to decide. If we were to disable macos tests then we need to add a check inside the spark module to raise an error/warning when it's being run on windows/osx.

Let me try to fix MacOS test , the test passed on my local Macbook.

.github/workflows/python_tests.yml
```diff
@@ -131,6 +131,7 @@ jobs:
     - name: Install Python package
       shell: bash -l {0}
       run: |
+        echo JAVA_HOME=$JAVA_HOME
```
WeichenXu123, line 134: @hcho3 @trivialfis 
I am curious for this action where it installs java and which version it uses.

WeichenXu123:
@trivialfis @hcho3 What's the memory limitation for the github action machine for running "Test XGBoost Python package on macos-10.15" ?
I guess the github action machine might have too low memory which causes spark executor crash and causes the spark jobs hangs when running tests.

WeichenXu123:
@hcho3 Could you assign me the permission to kill/rerun github actions ? so that I can rerun test faster.

hcho3:
@WeichenXu123 Once your first PR is merged, you'll have permissions to run and restart tests. Right now, you have limited permissions as a first-time contributor.

Can we merge this PR without MacOS support for now?

WeichenXu123:
> @WeichenXu123 Once your first PR is merged, you'll have permissions to run and restart tests. Right now, you have limited permissions as a first-time contributor.
> 
> Can we merge this PR without MacOS support for now?

We can. The CI hanging is very weird, local run I don't find any issue. Probably due to the CI machine resource limitation.

WeichenXu123:
@trivialfis @wbo4958 @hcho3 
All issues fixed, is this PR ready to merge ?

wbo4958:
Thx @WeichenXu123 for your hard work, @trivialfis Would you help to merge it?

WeichenXu123:
> Thank you for the excellent work on the pyspark interface! Will merge it once CI is fixed.

Thank you. @trivialfis There are some failure tests which is not related to my PR. Are they fixed in master branch ? If so I can merge master and rerun CI.

hcho3:
Please wait until the CI is fixed. ETA by tomorrow

wbo4958 (approved):

WeichenXu123:
@hcho3 Is CI fixed ? Thank you!

hcho3:
@WeichenXu123 I restarted the tests. Hopefully we can merge this PR soon

PR #8022, opened 2022-06-22
Fix compiler warnings.

trivialfis:
- Remove/fix unused parameters
- Remove deprecated code in rabit.
- Update dmlc-core.

Using nvcc 11.6 and gcc 9.4.0, only 1 warning remains:

```
hist_util.cc:43:89: warning: ‘void* memset(void*, int, size_t)’ clearing an object of non-trivial type ‘class xgboost::detail::GradientPairInternal<double>’; use assignment or value-initialization instead [-Wclass-memaccess]
   43 |   memset(hist.data() + begin, '\0', (end - begin) * sizeof(xgboost::GradientPairPrecise));
      |                                                                                         ^
```

This one requires some benchmarking. We can prepare for turning warnings into errors on CI once we start using a newer version of CUDA.

### DIFF
--- a/rabit/include/rabit/internal/socket.h
+++ b/rabit/include/rabit/internal/socket.h
@@ -397,7 +397,7 @@ class TCPSocket : public Socket{
    */
   inline void Create(int af = PF_INET) {
 #if !IS_MINGW()
-    sockfd = socket(PF_INET, SOCK_STREAM, 0);
+    sockfd = socket(af, SOCK_STREAM, 0);
     if (sockfd == kInvalidSocket) {
       Socket::Error("Create");
     }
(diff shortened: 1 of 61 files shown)


trivialfis:
https://xgboost-ci.net/blue/organizations/jenkins/xgboost/detail/PR-8022/4/pipeline

rabit/include/rabit/internal/socket.h
```diff
@@ -397,7 +397,7 @@ class TCPSocket : public Socket{
    */
   inline void Create(int af = PF_INET) {
 #if !IS_MINGW()
-    sockfd = socket(PF_INET, SOCK_STREAM, 0);
+    sockfd = socket(af, SOCK_STREAM, 0);
```
RAMitchell, line 400: !!!!???

RAMitchell (approved):

PR #8028, opened 2022-06-24
Fix multiple typos

jorislimonier:
Fix 4 "graphiz" instead of "graphviz".


trivialfis:
Thank you for the fix!

PR #8039, opened 2022-06-29
Force auc.cc to be statically linked

kiwiwarmnfuzzy:
So that its registration logic can be run, and thus 'auc' metric function can be found.

discuss.xgboost.ai discussion: [Should src/metric/auc.{h,cc,cu} be forced to be linked statically?](https://discuss.xgboost.ai/t/should-src-metric-auc-h-cc-cu-be-forced-to-be-linked-statically/2830)

### DIFF
--- a/src/metric/auc.cc
+++ b/src/metric/auc.cc
@@ -25,6 +25,8 @@
 
 namespace xgboost {
 namespace metric {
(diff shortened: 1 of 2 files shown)


trivialfis (approved):
Thank you for the fix!

PR #8050, opened 2022-07-05
Quantile Dmatrix for CPU.

trivialfis:
https://github.com/dmlc/xgboost/issues/7890

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -1260,7 +1260,9 @@ class QuantileDMatrix(DMatrix):
 
     ref :
         The training dataset that provides quantile information, needed when creating
-        validation/test dataset with ``QuantileDMatrix``.
+        validation/test dataset with ``QuantileDMatrix``. Supplying the training DMatrix
+        as a reference means that the same quantisation applied to the training data is
+        applied to the validation/test data
 
     """
 
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -1146,7 +1146,7 @@ class DMatrix:  # pylint: disable=too-many-instance-attributes
 
         Parameters
         ----------
-        feature_types : list or None
+        feature_types :
             Labels for features. None will reset existing feature names
 
         """
@@ -1189,7 +1189,7 @@ class DMatrix:  # pylint: disable=too-many-instance-attributes
 
 
 class _ProxyDMatrix(DMatrix):
-    """A placeholder class when DMatrix cannot be constructed (DeviceQuantileDMatrix,
+    """A placeholder class when DMatrix cannot be constructed (QuantileDMatrix,
     inplace_predict).
 
     """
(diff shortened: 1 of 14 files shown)


trivialfis:
TODOs:
- [ ] Tests
- [ ] Cleanup for various components that need to handle incomplete meta info and optional base row idx.
- [ ] Benchmark for accuracy, performance, and memory.

trivialfis:
Small benchmarks with 10 boosting iterations to test the initialization time.

|                   | QuantileDMatrix    | DMatrix            |
|:------------------|:-------------------|:-------------------|
| Sparse-1e6-64-0.1 | 3.5805992519890424 | 3.432847997988574  |
| Sparse-1e6-64-0.5 | 2.4399433140060864 | 2.2847362619941123 |
| Sparse-1e6-64-0.8 | 1.6939350479806308 | 1.6097231970052235 |
| Sparse-1e6-64-0.9 | 1.2425977539969608 | 1.2289739979896694 |
| HIGGS             | 11.658786264015362 | 11.978758318000473 |

WeichenXu123:
@trivialfis Is this near complete ? Once it merge would you start the data iteration optimizaton work for xgboost pyspark estimator ? :)

trivialfis:
It's pending for review. I will work on some other aspects around this data structure before promoting it to the default of spark pipeline.

python-package/xgboost/core.py
```diff
@@ -1234,17 +1234,33 @@ def _set_data_from_csr(self, csr: scipy.sparse.csr_matrix) -> None:
...
-    quantisation. This DMatrix is primarily designed to save memory in training from
-    device memory inputs by avoiding intermediate storage. Set max_bin to control the
-    number of bins during quantisation.  See doc string in :py:obj:`xgboost.DMatrix` for
-    documents on meta info.
+class QuantileDMatrix(DMatrix):
+    """A DMatrix variant that generates quantilized data directly from input for
+    ``hist`` and ``gpu_hist`` tree methods. This DMatrix is primarily designed to save
+    memory in training by avoiding intermediate storage. Set ``max_bin`` to control the
+    number of bins during quantisation, which should be consistent with the training
+    parameter ``max_bin``. When ``QuantileDMatrix`` is used for validation/test dataset,
```
RAMitchell, line 1242: Would be good to have slightly more information here.
e.g.
"Supplying the training DMatrix as a reference means that the same quantisation applied to the training data is applied to the validation/test data."

python-package/xgboost/core.py
```diff
@@ -1339,6 +1369,20 @@ def _init(self, data: DataType, enable_categorical: bool, **meta: Any) -> None:
         self.handle = handle
 
 
+class DeviceQuantileDMatrix(QuantileDMatrix):
```
RAMitchell, line 1372: Is this completely deprecated? Does that mean QuantileDMatrix automatically uses the device for device inputs?

RAMitchell (approved):

python-package/xgboost/core.py
trivialfis: Yes.  The c++ layer does the dispatching.

python-package/xgboost/core.py
trivialfis: Thank you for the suggestion. Added as part of the document for `ref`.

PR #8052, opened 2022-07-05
Fix mypy error with the latest dask.

trivialfis:
Dask is adding type hints to its codebase and as the result, checks in XGBoost can be
performed more rigorously.

- Remove compatibility with old dask version where multi lock was missing.
- Restrict input of `X` to be non-series.
- Adopt latest definition of `Delayed`.
- Avoid passing optional `host_ip`.
- Avoid deprecated `worker.nthreads`.

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -68,20 +68,20 @@ from .sklearn import xgboost_model_doc
 from .sklearn import _cls_predict_proba
 from .sklearn import XGBRanker
 
+
 if TYPE_CHECKING:
     from dask import dataframe as dd
     from dask import array as da
-    from dask import delayed as ddelayed
     import dask
     import distributed
 else:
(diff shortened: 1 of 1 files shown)


python-package/xgboost/dask.py
```diff
@@ -384,6 +356,7 @@ async def _map_local_data(
         label_upper_bound: Optional[_DaskCollection] = None,
     ) -> "DaskDMatrix":
         """Obtain references to local data."""
+        from dask.delayed import Delayed
```
hcho3, line 359: Why are we moving the import into the local scope? Is it possible to keep the import at the top of the module?

trivialfis: There's a `delayed` function in the `delayed` module. The lazy loader used in xgboost dask module cannot handle this so we have to import it. On the other hand we don't want to add dask as a hard dependency, so I moved it to local.

hcho3 (approved):

Issue #7778, opened 2022-03-31
Federated Learning RFC


rongou:
Here is a design proposal for initial support of Federated Learning. Please take a look and share you feedback. Thanks!

Partially addresses #5430.
@piiswrong @hcho3 @trivialfis @RAMitchell @tqchen @podcastinator

## Motivation
In many machine learning use cases, data owners (hospitals, banks, etc.) cannot share their private data with each other or with third parties, due to privacy regulations such as GDPR, HIPAA, and CCPA. Federated learning can be used to train a joint model without the need to aggregate data in a centralized location.

For certain types of problems (tabular data, time series forecasting), XGBoost is shown to be competitive against deep learning models, while requiring less tuning and computation. With the additional overhead in a federated setting, XGBoost is well suited for real world applications.

## Goals
For the initial version, the goals are limited and modest:
* Enhance XGBoost to support federated learning.
* Support using [NVFlare](https://github.com/NVIDIA/NVFlare) to coordinate the learning process, but the design should be amenable to support other federated learning platforms.
* The initial effort will focus on horizontal federated learning, i.e. each data partition contains all the features, including labels.
* Efficiency: training speed should be close to traditional distributed training.
* Accuracy: should be close to centralized learning.

## Non-Goals
* Should not maintain a separate fork of XGBoost.
* The initial version will not support vertical federated learning, i.e. features are distributed between different data owners.
* Initially we will assume the federated environment is non-adversarial, and will not provide strong privacy guarantees. This will be improved upon in later iterations.
* We will not support data owners dropping out during the learning process.

## Assumptions
Federated learning can refer to a wide range of problems, from a few close partners collaborating in a relatively secure environment, to millions of mobile devices connected over the Internet without encryption. For this design we assume we will start from the former, and can implement more security and privacy features with additional effort.

A typical use case is not that much different from the current distributed setting:
* A few trusted partners jointly train a model.
* Data is roughly evenly distributed among the participants.
* Reasonably fast network connection between each participant and a central trusted party.
* It is permissible to transmit histograms of features between participants, but not raw features directly.

## Risks

The current XGBoost codebase is fairly complicated and hard to modify. Some code refactoring needs to happen first, before support for federated learning can be added. Care must be taken to not break existing functionality, or make regular training harder.

## Design
[...]

PR #8057, opened 2022-07-08
Common interface for collective communication

rongou:
Part of #7778.


trivialfis (commented):
Thank you for the great work on cleaning up the communication interface. Do you think we can have some of these abstractions under the rabit umbrella? Rabit has a Python interface that we can use for other purposes like synchronizing user-defined metrics. (used in the callback module)

trivialfis:
Another benefit is that sometimes nccl has issues with users' network setup and the error is difficult to interpret (unhandled system error) and is only raised during training.  I want to have a Python interface unified with rabit for GPU communication that we can easily test and debug.

rongou:
Yes I do plan to hook this up with a python interface. Do you think we should keep the existing rabit api as is and just swap out the implementation? Or deprecate rabit and have a new interface, say, `collective`?

trivialfis:
I think we  can just do a swap. But since you have been working on it you might have better plan.

rongou:
@trivialfis @RAMitchell finally got the CI to pass. This is more or less the complete new unified api. I haven't changed the callers yet. Would you mind taking a quick look to see if it makes sense at a high level? Thanks!

trivialfis:
Thank you for the great work here! I will dive into the PR tomorrow.

python-package/xgboost/collective.py
```diff
@@ -0,0 +1,217 @@
...
+        the name of processor(host)
+    """
+    mxlen = 256
+    length = ctypes.c_ulong()
+    buf = ctypes.create_string_buffer(mxlen)
+    _LIB.XGCommunicatorGetProcessorName(buf, ctypes.byref(length), mxlen)
+    return buf.value
+
+
+T = TypeVar("T")                # pylint:disable=invalid-name
```
trivialfis, line 102: We can get this from `_typing.py`.

src/collective/communicator_factory.cc
```diff
@@ -0,0 +1,61 @@
...
+
+CommunicatorFactory::CommunicatorFactory(CommunicatorType type, Communicator* communicator)
+    : type_{type}, communicator_{communicator} {}
+
+void CommunicatorFactory::Init(int argc, char* argv[]) {
+  if (instance_) {
+    LOG(FATAL) << "Communicator factory can only be initialized once.";
+  }
+
+  auto type = GetTypeFromEnv();
```
trivialfis, line 26: Is it better to get it from env than using argv?

src/collective/communicator_factory.cu
```diff
@@ -0,0 +1,77 @@
...
+#if defined(XGBOOST_USE_FEDERATED)
+#include "../../plugin/federated/federated_communicator.h"
+#endif
+
+namespace xgboost {
+namespace collective {
+
+thread_local std::unique_ptr<CommunicatorFactory> CommunicatorFactory::instance_{};
+
+CommunicatorFactory::CommunicatorFactory(CommunicatorType type, Communicator* communicator)
```
trivialfis, line 19: Maybe we can use dmlc registry to help define the factory?

include/xgboost/c_api.h
```diff
@@ -1386,4 +1388,86 @@ XGB_DLL int XGBoosterFeatureScore(BoosterHandle handle, const char *json_config,
...
+/*!
+ * \brief initialize the collective communicator,
+ *  call this once before using anything
+ *  The additional arguments is not necessary.
+ *  Usually the communicator will detect settings
+ *  from environment variables.
+ * \param argc number of arguments in argv
+ * \param argv the array of input arguments
+ */
+XGB_DLL int XGCommunicatorInit(int argc, char *argv[]);
```
trivialfis, line 1401: We have been using JSON for this type of argument. I think it's better than argv since the input can be easily verified.

src/collective/communicator_factory.cc
```diff
@@ -0,0 +1,61 @@
...
+  auto const arg = GetTypeFromArgs(argc, argv);
+  if (arg != CommunicatorType::kUnknown) {
+    type = arg;
+  }
+  switch (type) {
+    case CommunicatorType::kRabit: {
+      RabitCommunicatorFactory factory{argc, argv};
+      auto* comm = factory.Create();
+      instance_.reset(new CommunicatorFactory(type, comm));
```
trivialfis, line 35: So `CommunicatorFactory` is a factory of factories?

python-package/xgboost/collective.py
```diff
@@ -0,0 +1,217 @@
...
+from typing import Any, TypeVar, Optional, cast, List, Union
+
+import numpy as np
+
+from .core import _LIB, c_str, _check_call
+
+LOGGER = logging.getLogger("[xgboost.collective]")
+
+
+def _init_collective() -> None:
```
trivialfis, line 15: Is this necessary?

python-package/xgboost/collective.py
```diff
@@ -0,0 +1,217 @@
...
+    """Get the processor name.
+
+    Returns
+    -------
+    name : str
+        the name of processor(host)
+    """
+    mxlen = 256
+    length = ctypes.c_ulong()
+    buf = ctypes.create_string_buffer(mxlen)
```
trivialfis, line 97: XGBoost has a thread-local store for returning buffer. Take a look at the C API on how various strings and floating point buffers are returned.

python-package/xgboost/collective.py
```diff
@@ -0,0 +1,217 @@
...
+        the result of broadcast.
+    """
+    rank = get_rank()
+    length = ctypes.c_ulong()
+    if root == rank:
+        assert data is not None, 'need to pass in data when broadcasting'
+        s = pickle.dumps(data, protocol=pickle.HIGHEST_PROTOCOL)
+        length.value = len(s)
+    # run first broadcast
+    _check_call(_LIB.XGCommunicatorBroadcast(ctypes.byref(length),
```
trivialfis, line 127: Maybe we can simplify this down to 1 broadcast with the internal return buffer?

src/collective/communicator_factory.cu
```diff
@@ -0,0 +1,77 @@
...
+  }
+}
+
+void CommunicatorFactory::Finalize() { instance_.reset(); }
+
+DeviceCommunicator* CommunicatorFactory::GetDeviceCommunicator(int device_ordinal) {
+  if (!device_communicator_) {
+#ifdef XGBOOST_USE_NCCL
+    if (type_ != CommunicatorType::kFederated) {
+      device_communicator_.reset(new NcclDeviceCommunicator(device_ordinal, communicator_.get()));
```
trivialfis, line 64: Out of curiosity, nccl can use host network right?

trivialfis (commented):
Thank you for the great work on unifying the API!

Overall I have a couple of design questions:
- Do we need two layers of factories?
- Is the communicator expected to be short-lived? (created on demand) while the factory is expected to be long-lived (throughout the whole training session).
- How does user access the device communicator? Is it necessary to call `DeviceCommunicator` or we can just treat it as any other comm at the interface level?

src/collective/communicator_factory.cu
rongou: Yeah I think it's possible, but we need to figure out how to configure the parameters first. Will come back to this.

rongou (commented):
> Do we need two layers of factories?

You mean the separation between communicator and device communicator? They are separate because the behaviors are different, the device communicator always assumes device pointers.

Or if you mean the factory itself, it's really just a singleton that provides access to the communicator:
`CommunicatorFactory::GetInstance()->GetCommunicator()`

This probably can be simplified with the dmlc registry.

> Is the communicator expected to be short-lived? (created on demand) while the factory is expected to be long-lived (throughout the whole training session).

Both are expected to be long-lived singletons.

> How does user access the device communicator? Is it necessary to call DeviceCommunicator or we can just treat it as any other comm at the interface level?

`CommunicatorFactory::GetInstance()->GetDeviceCommunicator(gpu_id)`

As mentioned above, the behavior is different.

rongou:
@trivialfis switched to use JSON to configure the communicators. Please take another look. Thanks!

Next I'll try to switch to the dmlc registry.

trivialfis:
> Next I'll try to switch to the dmlc registry.

I don't have a strong preference for this. Feel free to create your own factory methods if needed. I just wanted to avoid code duplication due to GPU support.

src/collective/communicator_factory.cu
rongou: Ended up not using the registry since we need some logic to pick which communicator to use, but got rid of the factories.

include/xgboost/c_api.h
rongou: Done.

src/collective/communicator_factory.cc
rongou: Got rid of the factories.

rongou:
@trivialfis got rid of the factories. Please take another look. Should we try to merge this? So far this is a non-breaking change. Once I start modifying the callers, it'd probably become a breaking change.

include/xgboost/c_api.h
```diff
@@ -1386,4 +1388,118 @@ XGB_DLL int XGBoosterFeatureScore(BoosterHandle handle, const char *json_config,
                                   bst_ulong *out_dim,
                                   bst_ulong const **out_shape,
                                   float const **out_scores);
+
+/*!
+ * \brief Initialize the collective communicator.
```
trivialfis, line 1393: Can we mark this as experimental? I see that we might be making breaking changes in the future as we might try to unify some other features like CPU/GPU communicator interface, or clean up the input parameters.

plugin/federated/federated_server.cc
```diff
@@ -71,32 +73,35 @@ class AllreduceFunctor {
   void Accumulate(std::string& buffer, std::string const& input, DataType data_type,
                   ReduceOperation reduce_operation) const {
     switch (data_type) {
-      case DataType::CHAR:
-        Accumulate(&buffer[0], reinterpret_cast<char const*>(input.data()), buffer.size(),
+      case DataType::INT8:
```
trivialfis, line 76: Thank you for cleaning up the types!

python-package/xgboost/collective.py
```diff
@@ -0,0 +1,244 @@
...
+    np.dtype('int64'): 4,
+    np.dtype('uint64'): 5,
+    np.dtype('float32'): 6,
+    np.dtype('float64'): 7
+}
+
+
+@unique
+class Op(IntEnum):
+    """Supported operations for rabit."""
```
trivialfis, line 191: Do we still call it rabit?

python-package/xgboost/collective.py
```diff
@@ -0,0 +1,244 @@
...
+    -------
+    result :
+        The result of allreduce, have same shape as data
+
+    Notes
+    -----
+    This function is not thread-safe.
+    """
+    if not isinstance(data, np.ndarray):
+        raise Exception('allreduce only takes in numpy.ndarray')
```
trivialfis, line 219: TypeError

python-package/xgboost/collective.py
```diff
@@ -0,0 +1,244 @@
...
+        data: np.ndarray, op: Op
+) -> np.ndarray:
+    """Perform allreduce, return the result.
+
+    Parameters
+    ----------
+    data :
+        Input data.
+    op :
+        Reduction operators, can be MAX or SUM
```
trivialfis, line 207: Not `MIN`? I think we can omit `can be ...` in the doc since we have the Python type hint.

src/collective/communicator.cc
```diff
@@ -0,0 +1,59 @@
...
+    type = CommunicatorType::kRabit;
+  }
+  type_ = type;
+  switch (type) {
+    case CommunicatorType::kRabit: {
+      communicator_.reset(RabitCommunicator::Create(config));
+      break;
+    }
+    case CommunicatorType::kMPI:
+      LOG(FATAL) << "Not implemented yet.";
```
trivialfis, line 39: If you don't plan to implement it, we can just remove it.

src/collective/communicator.cu
```diff
@@ -0,0 +1,40 @@
...
+#include "device_communicator.cuh"
+#include "device_communicator_adapter.cuh"
+#ifdef XGBOOST_USE_NCCL
+#include "nccl_device_communicator.cuh"
+#endif
+
+namespace xgboost {
+namespace collective {
+
+thread_local int Communicator::device_ordinal_{-1};
```
trivialfis, line 14: We want to do something for this parameter, could you please take a look at https://github.com/dmlc/xgboost/issues/7308 ? One issue in my experimental implementation is that we need to consider multi-threaded environments like Python async functions. In such environments, it's difficult to enforce synchronization of these thread local parameters.

trivialfis (commented):
The PR looks good to me overall. Left my comments in the code review.

As for the interface, I hope that we can eventually merge the interfaces for CPU and GPU in the future by having some dispatching mechanism for recognizing input data device ordinal. I think it's doable in recent future.

python-package/xgboost/collective.py
rongou: We don't currently use `MIN`, but added it back since it's not too much code.

python-package/xgboost/collective.py
rongou: Done.

src/collective/communicator.cc
rongou: Do you know if people actually use the current MPI integration with Rabit? It's probably not too hard to add it if there is demand for it.

src/collective/communicator.cu
rongou: The proposed python interface in #7308 looks good.

Here the device id doesn't have to be per thread, assuming we have a global configuration that sets it, and it's available when we call `Communicator::Init()`. But the communicators do need to be thread local since they are not thread safe.

src/collective/communicator.cc
rongou: Removed for now.

rongou:
@trivialfis can this be merged? Is there anything else you want me to change? Thanks!

trivialfis (approved):
Let's merge it once the CI is back to healh.  Thank you for the great work.  Could you please share more details about the future plan for federated learning?

rongou:
The main goal is to include federated learning in a release so users can just `pip install xgboost` to get the functionality. A few steps left to do:
* Add  the JNI wrapper for the communicator api
* Change all the callers to use communicator instead of rabit
* Change CI to build with federated learning enabled by default
* Release with federated learning included

Once this is done, we can probably do more with Rabit (rewrite, swap out), unify host/device apis.

For federated learning, we may look at things like homomorphic encryption, differential privacy, and vertical federated learning (features are split between participants).

trivialfis:
Restarted jenkins linux.

lidh15:
> homomorphic encryption, differential privacy



> The main goal is to include federated learning in a release so users can just `pip install xgboost` to get the functionality. A few steps left to do:
> 
> * Add  the JNI wrapper for the communicator api
> * Change all the callers to use communicator instead of rabit
> * Change CI to build with federated learning enabled by default
> * Release with federated learning included
> 
> Once this is done, we can probably do more with Rabit (rewrite, swap out), unify host/device apis.
> 
> For federated learning, we may look at things like homomorphic encryption, differential privacy, and vertical federated learning (features are split between participants).

do we have homomorphic encryption or differential privacy in the plugin now?

PR #8068, opened 2022-07-13
[PySpark] add gpu support for spark local mode

wbo4958:
This PR is to support gpu training (use_gpu=True) on Spark local mode.

Please note that, 

1. Spark local mode does not support GPU configuration, so it will fail if the user adds related GPU configurations.
2. For now, we only support num_workers=1 when running on GPU on Spark local mode.

This feature is needed since it's really convenient for local debugging.

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -53,6 +53,8 @@ from .utils import (
     _get_rabit_args,
     _get_args_from_message_list,
     _get_spark_session,
+    _is_local,
+    _get_gpu_id,
 )
 from .params import (
     HasArbitraryParamsDict,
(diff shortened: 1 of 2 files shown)


wbo4958:
@trivialfis How to run python tests locally? is there a doc for that?

trivialfis:
```
pytest ./tests/python
```
You can check the doc for pytest.

python-package/xgboost/spark/core.py
```diff
@@ -281,19 +283,40 @@ def _validate_params(self):
...
+                        "The spark cluster does not support gpu configuration for local mode. "
+                        + "Please delete spark.executor.resource.gpu.amount and "
+                        + "spark.task.resource.gpu.amount"
+                    )
+
+                if self.getOrDefault(self.num_workers) > 1:
+                    raise ValueError(
+                        "Training XGBoost on the spark local mode only supports num_workers = 1, "
+                        + "and only primary GPU device will be used."
+                    )
```
WeichenXu123, line 304: I think we can still support multiple GPU workers in spark local mode.
We can get partition id from spark TaskContext, and use the partition id as the gpu_id for the corresponding spark task. WDYT ?

wbo4958: Thx. Hmm, there will be only 1 process for spark local mode, If we support the scenario of num_workers > 1, then all the GPU training tasks will run on the same process, which seems not to be supported due to the NCCL issue. @trivialfis Could you confirm that?

trivialfis: It should work but let's not invite trouble.

wbo4958: Yeah, I agree. The GPU supporting for the local mode is mostly used in local debugging.

WeichenXu123 (approved):

python-package/xgboost/spark/core.py
WeichenXu123: @wbo4958 
A correction:
> local mode pyspark all the GPU training tasks will run on the same process

This is not true.
For pyspark, in barrier mode pyspark task, each task will be run in an individual python process.

wbo4958: Oh right, my bad. I previously thought it was the JVM side. Ok, let me add it support.

trivialfis (approved):

trivialfis:
Please address the linter error.

PR #8070, opened 2022-07-14
[CI] Try installing modin through conda.

trivialfis:


trivialfis:
Will open a different PR.

PR #8072, opened 2022-07-14
[py-skl] Use sklearn `check_random_state`.

trivialfis:
This aligns with sklearn rs handling.


trivialfis:
Closing as we need to support 64bit seed.

PR #8073, opened 2022-07-14
Refactor split valuation kernel

RAMitchell:
Increases split evaluation kernel throughput on my V100 from 150 GB/s to 400 GB/s. 

These gains come from:
- using cuda fast division operations
- avoiding shared memory broadcasts, instead using warp shuffle.
- not computing parents gain for every possible split (this is constant per node)

The fastest possible throughput I have achieved is 700 GB/s, however it becomes a little complicated to get this fast. Current version I think is fast enough, placing the bottleneck entirely on histogram computation.


src/tree/split_evaluator.h
```diff
@@ -118,21 +118,32 @@ class TreeEvaluator {
       return ::xgboost::tree::CalcWeight(param, stats);
     }
 
-    XGBOOST_DEVICE float
-    CalcGainGivenWeight(ParamT const &p, tree::GradStats const& stats, float w) const {
+    // Fast floating point division instruction on device
+    XGBOOST_DEVICE float Divide(float a, float b) const {
```
trivialfis, line 122: Can we extract this as an independent function?

src/tree/gpu_hist/evaluate_splits.cu
```diff
@@ -22,202 +22,190 @@ XGBOOST_DEVICE float LossChangeMissing(const GradientPairPrecise &scan,
...
-  auto begin = feature_histogram.data();
-  auto end = begin + feature_histogram.size();
-  for (auto itr = begin; itr < end; itr += BLOCK_THREADS) {
-    bool thread_active = itr + threadIdx.x < end;
-    // Scan histogram
-    GradientSumT bin = thread_active ? *(itr + threadIdx.x) : GradientSumT();
-    local_sum += bin;
+
+template <int BLOCK_THREADS, typename GradientSumT>
+class EvaluateSplitAgent {
```
trivialfis, line 36: Would be great if you put some notes here about the assumption of block size equals the warp size.

src/tree/gpu_hist/evaluate_splits.cu
```diff
@@ -22,202 +22,190 @@ XGBOOST_DEVICE float LossChangeMissing(const GradientPairPrecise &scan,
...
-  // For loop sums features into one block size
-  auto begin = feature_histogram.data();
-  auto end = begin + feature_histogram.size();
-  for (auto itr = begin; itr < end; itr += BLOCK_THREADS) {
-    bool thread_active = itr + threadIdx.x < end;
-    // Scan histogram
-    GradientSumT bin = thread_active ? *(itr + threadIdx.x) : GradientSumT();
-    local_sum += bin;
+
+template <int BLOCK_THREADS, typename GradientSumT>
```
trivialfis, line 35: Is `GradientSumT` still needed? I think we have removed single precision floating point histogram.

trivialfis (commented):
The performance improvement looks exciting! Some questions in the comments.

RAMitchell:
Depth 8 benchmarks:

| dataset | master      | eval        |
| airline | 90.88661192 | 89.93374479 |
| bosch   | 12.88504644 | 12.46627029 |
| covtype | 18.01187677 | 17.60098921 |
| epsilon | 46.48386218 | 43.91058178 |
| fraud   | 1.315704659 | 1.237399099 |
| higgs   | 17.19260674 | 17.27932671 |
| year    | 7.047273015 | 6.841409724 |

src/tree/split_evaluator.h
RAMitchell: I'm not expecting to use it anywhere else at this moment, so I think it should stay unless you have something specific in mind A kernel needs to be heavily bottlenecked by arithmetic before this makes a difference, and I can't think of other places in xgboost.

src/tree/gpu_hist/evaluate_splits.cu
RAMitchell: I think I'll throw in a static assert with an informative message in case someone tries to change this.

src/tree/gpu_hist/evaluate_splits.cu
RAMitchell: Will remove.

trivialfis (approved):

PR #8084, opened 2022-07-15
Check cuDF lazily

grafail:
In some cases the "import cudf" can raise a RuntimeError or other errors while loading. Doing a lazy check is safer for these cases. Similiar to #7752.

Example of RuntimeError that could be generated by import:
```python
  File "cuda/_cuda/ccuda.pyx", line 3671, in cuda._cuda.ccuda._cuInit
  File "cuda/_cuda/ccuda.pyx", line 435, in cuda._cuda.ccuda.cuPythonInit
RuntimeError: Failed to dlopen libcuda.so
```

### DIFF
--- a/python-package/xgboost/data.py
+++ b/python-package/xgboost/data.py
@@ -745,11 +745,7 @@ def _from_cudf_df(
 
 
 def _is_cudf_ser(data: DataType) -> bool:
-    try:
-        import cudf
-    except ImportError:
-        return False
-    return isinstance(data, cudf.Series)
+    return lazy_isinstance(data, "cudf.core.series", "Series")
 
 
 def _is_cupy_array(data: DataType) -> bool:


bdice:
Hi @grafail, would you be able to share a full traceback and cudf version info, and also the version of cuda-python? I’d like to dig into this further. This is being raised by cuda-python and we try to guard against it in cudf but might be missing a case. Thanks!

trivialfis (approved):
Thank you for the PR! It would be great if you can help share a reproducible example with cuDF developers.

grafail:
Hi @bdice. This issue can by recreated by just importing cudf on `rapidsai/rapidsai-core:22.06-cuda11.2-base-ubuntu18.04-py3.8`, while not using the [nvidia runtime](https://github.com/NVIDIA/nvidia-container-runtime) in docker.
It seems to not create an issue on `rapidsai/rapidsai-core:21.10-cuda11.2-base-ubuntu18.04-py3.8` though.

bdice:
@grafail  Thanks for the additional information!

PR #8096, opened 2022-07-19
[Python] Require black and isort for new Python files.

trivialfis:
- Require black and isort for spark and dask module.

These files are relatively new and are more conform to the black formatter. We will
convert the rest of the library as we move forward.

Other libraries including dask/distributed and optuna use the same formatting style and
have a more strict standard. The black formatter is indeed quite nice, automating it can
help us unify the code style.

- Gather Python checks into a single script.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -146,24 +146,6 @@ jobs:
       run: |
         make lint
 
-  mypy:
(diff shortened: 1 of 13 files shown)


hcho3:
Never mind my previous comment. `make lint` should eventually be changed to use Black, but not now.

PR #8103, opened 2022-07-21
Prepare gradient index for Quantile DMatrix.

trivialfis:
- Extract `PushBatchImpl` for reusing code. This is the common function between pushing a sparse page and a adapter, the former doesn't need to count row size and it known to be the single page.
- Implement `GetFvalue` for prediction.

Extracted from: https://github.com/dmlc/xgboost/pull/8050 .


trivialfis:
After this, there will be another 2 PRs, one for C++ iterative DMatrix and another one for the Python interface.

RAMitchell (approved):

Issue #8108, opened 2022-07-22
Xgboost pyspark: Support pyspark Sparse Vector


WeichenXu123:
Xgboost pyspark: Support pyspark Sparse Vector

WeichenXu123:
@trivialfis 

@xiangrui hope xgboost 2.0 can support this on databricks. 
The feature requires a `unwrap_udt` API which now only available on databricks.
Can we add this feature now ? I will checking whether it is on databricks environment and if yes  then imports the databricks-specific `unwrap_udt` function.

Btw, when apache/spark 3.4 comes out, `unwrap_udt` API will be added and then apache/spark can also support this feature.

WeichenXu123:
@trivialfis

> Can we add this feature now ? In the code it can check whether it is on databricks environment and if yes then import the databricks-specific unwrap_udt function.

What do you think ?

trivialfis:
https://github.com/dmlc/xgboost/issues/8108#issuecomment-1192705836 sounds good

trivialfis:
hi @WeichenXu123 is it possible for us to use `mapInArrow` instead of `mapInPandas`? What's the current status of the API and does it support device memory? https://arrow.apache.org/docs/cpp/memory.html#devices

WeichenXu123:
> hi @WeichenXu123 is it possible for us to use `mapInArrow` instead of `mapInPandas`? What's the current status of the API and does it support device memory? https://arrow.apache.org/docs/cpp/memory.html#devices

CC @hyukjin.kwon

WeichenXu123:
Got reply from @HyukjinKwon:

> 
it should support
although nobody tested it
because it exposes the native PyArrow batch instances
if that supports, it should work

We can try it as a follow-up optimization.



[6:26](https://databricks.slack.com/archives/DJ0AR42B1/p1660213594575069)
because it exposes the native PyArrow batch instances
[6:26](https://databricks.slack.com/archives/DJ0AR42B1/p1660213600313039)
if that supports, it should work




[6:26](https://databricks.slack.com/archives/DJ0AR42B1/p1660213594575069)
because it exposes the native PyArrow batch instances
[6:26](https://databricks.slack.com/archives/DJ0AR42B1/p1660213600313039)
if that supports, it should work

PR #8117, opened 2022-07-26
[pyspark] support a list of feature column names

wbo4958:
This PR introduces a new Param storing a list of feature column names,  so with it, users do not need to vectorize or create an array feature column beforehand, instead, they need to specify feature_col with a list of column names.

Eg,

``` python
feature_col_names = [x.name for x in raw_df.schema if x.name != label]
xgb = SparkXGBClassifier(features_col=feature_col_names, label_col=label)
```

why introduce this new parameter?

The internal of XGBoost is wrapping the feature columns into a vector or an array column on the JVM side, but it will unwrap the vector/array column into each feature column on the python side. So if we just pass each feature column, which can improve the performance.

But for now, there are some limitations that users must enable `use_gpu` when using this way and it may break some spark ml pipelines eg, `OneVsRest`


wbo4958:
@trivialfis @WeichenXu123 Could you help to review this PR? Thx

trivialfis (commented):
Thank you for the PR. Could you please:
- add tests
- Add a demo about how to use it with other spark ml utilities.

wbo4958:
> Thank you for the PR. Could you please:
> 
> * add tests
> * Add a demo or test that shows how to use it with other spark ml utilities.

Done @trivialfis @WeichenXu123

python-package/xgboost/spark/core.py
```diff
@@ -315,6 +320,23 @@ def _validate_params(self):
...
 
+def _validate_and_convert_feature_col_as_float_col(
+    dataset, features_col_name: list
+) -> list:
+    """feature column names must be IntegralType or float or double types"""
+    feature_cols = []
+    for c in features_col_name:
+        if isinstance(dataset.schema[c].dataType, DoubleType):
+            feature_cols.append(col(c).cast(FloatType()).alias(c))
+        elif isinstance(dataset.schema[c].dataType, (FloatType, IntegralType)):
```
WeichenXu123, line 331: What's IntegralType type ? Do you mean IntegerType ?

python-package/xgboost/spark/core.py
```diff
@@ -315,6 +320,23 @@ def _validate_params(self):
                     )
 
 
+def _validate_and_convert_feature_col_as_float_col(
```
WeichenXu123, line 323: _validate_and_convert_feature_col_as_float_col_list

python-package/xgboost/spark/core.py
```diff
@@ -315,6 +320,23 @@ def _validate_params(self):
...
+    dataset, features_col_name: list
+) -> list:
+    """feature column names must be IntegralType or float or double types"""
```
WeichenXu123, line 326: Values in feature columns must be...

python-package/xgboost/spark/core.py
```diff
@@ -315,6 +320,23 @@ def _validate_params(self):
                     )
 
 
+def _validate_and_convert_feature_col_as_float_col(
+    dataset, features_col_name: list
```
WeichenXu123, line 324: features_col_names

python-package/xgboost/spark/core.py
```diff
@@ -341,6 +363,22 @@ def _validate_and_convert_feature_col_as_array_col(dataset, features_col_name):
     return features_array_col
 
 
+def _validate_and_convert_feature_col(
```
WeichenXu123, line 366: Is the function being used ?

python-package/xgboost/spark/core.py
```diff
@@ -262,6 +260,13 @@ def _validate_params(self):
                 "Therefore, that parameter will be ignored."
             )
 
+        if self.getOrDefault(self.features_cols) and not self.getOrDefault(
+            self.use_gpu
+        ):
+            raise ValueError(
+                "XGBoost accepts a list of feature column names only when use_gpu is enabled."
+            )
```
WeichenXu123, line 268: We should emit warning say if "features_cols" param set, then "features_col" param is igored.

python-package/xgboost/spark/core.py
```diff
@@ -583,9 +636,7 @@ def _train_booster(pandas_df_iter):
             evals_result = {}
             with RabitContext(_rabit_args, context):
                 dtrain, dvalid = create_dmatrix_from_partitions(
-                    pandas_df_iter,
-                    None,
-                    dmatrix_kwargs,
+                    pandas_df_iter, features_cols_names, dmatrix_kwargs
```
WeichenXu123, line 639: What if features_cols_names conflicts with label / weight / base_margin column name ?

trivialfis:
Could you please add a test for cross validation fro sparkml as well?

python-package/xgboost/spark/core.py
wbo4958: @trivialfis, could you correct me if xgboost supports IntegralType (Byte/Integer/Long/Short) ?

trivialfis: Yes, sort of. XGBoost will convert the data to float internally.

trivialfis: BTW, XGBoost can convert it on the fly without creating a copy of data.

wbo4958: Thx @trivialfis, Yeah, if the type is short/byte, it will reduce the size of the shuffle write.

python-package/xgboost/spark/core.py
wbo4958: Done

python-package/xgboost/spark/core.py
wbo4958: Done

python-package/xgboost/spark/core.py
wbo4958: Done

python-package/xgboost/spark/core.py
wbo4958: Thx, removed.

python-package/xgboost/spark/core.py
wbo4958: Done

python-package/xgboost/spark/core.py
wbo4958: Thx, I tried, it worked, no exception happened.

wbo4958:
@trivialfis could you help to review again. Thx

python-package/xgboost/spark/core.py
```diff
@@ -315,6 +320,23 @@ def _validate_params(self):
...
 
 
+def _validate_and_convert_feature_col_as_float_col_list(
+    dataset, features_col_names: list
+) -> list:
+    """Values in feature columns must be integral types or float/double types"""
+    feature_cols = []
+    for c in features_col_names:
+        if isinstance(dataset.schema[c].dataType, DoubleType):
+            feature_cols.append(col(c).cast(FloatType()).alias(c))
```
trivialfis, line 330: I don't think we need to do the cast here.  XGBoost can handle double type.

python-package/xgboost/spark/core.py
```diff
@@ -685,6 +727,38 @@ def read(cls):
...
+    def _get_feature_col(self, dataset) -> (list, Optional[list]):
+        """XGBoost model trained with features_cols parameter can also predict
+        vector or array feature type. But first we need to check features_cols
+        and then featuresCol
+        """
+
+        feature_col_names = self.getOrDefault(self.features_cols)
+        features_col = []
+        if (
+            len(feature_col_names)
```
trivialfis, line 739: This condition looks weird?

```
len(feature_col_names) > 0 >= len(...)
```
is it correct?

python-package/xgboost/spark/core.py
```diff
@@ -497,10 +525,19 @@ def _fit(self, dataset):
         self._validate_params()
         label_col = col(self.getOrDefault(self.labelCol)).alias(alias.label)
 
-        features_array_col = _validate_and_convert_feature_col_as_array_col(
-            dataset, self.getOrDefault(self.featuresCol)
-        )
-        select_cols = [features_array_col, label_col]
+        select_cols = [label_col]
+        features_cols_names = None
+        if len(self.getOrDefault(self.features_cols)):
```
trivialfis, line 530:
```suggestion
        if self.getOrDefault(self.features_cols):
```

trivialfis (commented):
Thank you for the PR!  Some questions in the comments.

python-package/xgboost/spark/core.py
wbo4958: casting here can reduce the shuffle write size.

python-package/xgboost/spark/core.py
wbo4958: yes. the pylint required this

trivialfis:
``` python
all(c in dataset.columns for c in feature_col_names)
```

wbo4958: Done, I use another set way to check.

python-package/xgboost/spark/core.py
wbo4958: Done, Thx

trivialfis:
note to myself: need to update the linter script.

trivialfis (approved):

python-package/xgboost/spark/core.py
WeichenXu123: Q:
Do we need to support integer type here ?

WeichenXu123 (approved):
LGTM, leave one question though.

WeichenXu123:
@trivialfis @wbo4958 
I will file PR for https://github.com/dmlc/xgboost/issues/8108 once this PR merged.
And hope my next PR will go into xgboost 2.0 .
Thanks!

wbo4958:
@trivialfis @WeichenXu123 Thx for reviewing.

PR #8129, opened 2022-08-01
Fix compatibility with latest cupy.

trivialfis:


trivialfis:
Frozen mypy due to https://github.com/python/typeshed/issues/8351 .

PR #8131, opened 2022-08-01
In PySpark Estimator example use the model with validation_indicator

praateekmahajan:
The example was not using `xgb_regressor2` or `xgb_classifier2` which actually has `validation_indicator_col` property defined.


praateekmahajan:
@WeichenXu123

trivialfis (approved):
Thank you for the fix

PR #8141, opened 2022-08-04
Use config_context in sklearn interface.

trivialfis:
Related https://github.com/dmlc/xgboost/issues/7308

- Test for nested `config_context`.
- Test for thread safety of global config.
- Use config context in sklearn interface.
- Format the sklearn module.

### DIFF
--- a/demo/dask/README.rst
+++ b/demo/dask/README.rst
@@ -1,3 +1,5 @@
+.. _dask-examples:
+
 XGBoost Dask Feature Walkthrough
 ================================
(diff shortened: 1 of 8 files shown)


trivialfis:
I ran formatter on the sklearn module since the change is already spanning through the module anyway.

trivialfis:
I ran formatter on the sklearn module since the change is already spanning through the module anyway.

hcho3 (approved):

PR #8145, opened 2022-08-09
[pyspark] Make Xgboost estimator support using sparse matrix as optimization

WeichenXu123:
Closes https://github.com/dmlc/xgboost/issues/8108

Make Xgboost estimator support using sparse matrix as optimization.

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -547,8 +547,22 @@ class _SparkXGBEstimator(Estimator, _SparkXGBParams, MLReadable, MLWritable):
 
         select_cols = [label_col]
         features_cols_names = None
-        if self.getOrDefault(self.enable_sparse_data_optim):
+        enable_sparse_data_optim = self.getOrDefault(self.enable_sparse_data_optim)
+        if enable_sparse_data_optim:
             from pyspark.ml.linalg import VectorUDT
+
+            if self.getOrDefault(self.missing) != 0.0:
+                # If DMatrix is constructed from csr / csc matrix, then inactive elements
+                # in csr / csc matrix are regarded as missing value, but, in pyspark, we
+                # are hard to control elements to be active or inactive in sparse vector column,
+                # some spark transformers such as VectorAssembler might compress vectors
+                # to be dense or sparse format automatically, and when a spark ML vector object
+                # is compressed to sparse vector, then all zero value elements become inactive.
+                # So we force setting missing param to be 0 when enable_sparse_data_optim config
+                # is True.
+                raise ValueError(
+                    "If enable_sparse_data_optim is True, missing param != 0 is not supported."
+                )
 
             if self.getOrDefault(self.features_cols):
                 raise ValueError(
@@ -565,7 +579,13 @@ class _SparkXGBEstimator(Estimator, _SparkXGBParams, MLReadable, MLWritable):
                 )
 
             unwrap_udf = _get_unwrap_udf_fn()
-
+            features_unwrapped_vec_col = unwrap_udf(col(features_col_name))
+            select_cols.extend([
+                features_unwrapped_vec_col.type.alias("featureVectorType"),
+                features_unwrapped_vec_col.size.alias("featureVectorSize"),
+                features_unwrapped_vec_col.indices.alias("featureVectorIndices"),
+                features_unwrapped_vec_col.values.alias("featureVectorValues"),
+            ])
         else:
             if self.getOrDefault(self.features_cols):
                 features_cols_names = self.getOrDefault(self.features_cols)
@@ -629,7 +649,7 @@ class _SparkXGBEstimator(Estimator, _SparkXGBParams, MLReadable, MLWritable):
             "feature_types": self.getOrDefault(self.feature_types),
             "feature_names": self.getOrDefault(self.feature_names),
             "feature_weights": self.getOrDefault(self.feature_weights),
-            "missing": self.getOrDefault(self.missing),
+            "missing": float(self.getOrDefault(self.missing)),
         }
         booster_params["nthread"] = cpu_per_task
         use_gpu = self.getOrDefault(self.use_gpu)
@@ -667,7 +687,8 @@ class _SparkXGBEstimator(Estimator, _SparkXGBParams, MLReadable, MLWritable):
             evals_result = {}
             with RabitContext(_rabit_args, context):
                 dtrain, dvalid = create_dmatrix_from_partitions(
-                    pandas_df_iter, features_cols_names, gpu_id, dmatrix_kwargs
+                    pandas_df_iter, features_cols_names, gpu_id, dmatrix_kwargs,
+                    enable_sparse_data_optim=enable_sparse_data_optim,
                 )
                 if dvalid is not None:
                     dval = [(dtrain, "training"), (dvalid, "validation")]
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -366,7 +366,7 @@ def _validate_and_convert_feature_col_as_array_col(dataset, features_col_name):
     return features_array_col
 
 
-def _get_unwrap_udf_fn():
+def _get_unwrap_udt_fn():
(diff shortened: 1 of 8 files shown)


WeichenXu123:
CC @wbo4958 @trivialfis The PR is ready for review. :)

python-package/xgboost/spark/core.py
```diff
@@ -527,17 +547,69 @@ def _fit(self, dataset):
...
+                # some spark transformers such as VectorAssembler might compress vectors
+                # to be dense or sparse format automatically, and when a spark ML vector object
+                # is compressed to sparse vector, then all zero value elements become inactive.
+                # So we force setting missing param to be 0 when enable_sparse_data_optim config
+                # is True.
+                raise ValueError(
+                    "If enable_sparse_data_optim is True, missing param != 0 is not supported."
+                )
+
+            if self.getOrDefault(self.features_cols):
```
trivialfis, line 567: @wbo4958 Would you like to take a look into this?

python-package/xgboost/spark/core.py
```diff
@@ -527,17 +547,69 @@ def _fit(self, dataset):
...
+            if not isinstance(features_col_datatype, VectorUDT):
+                raise ValueError(
+                    "If enable_sparse_data_optim is True, the feature column values must be "
+                    "`pyspark.ml.linalg.Vector` type."
+                )
+
+            unwrap_udt = _get_unwrap_udt_fn()
+            features_unwrapped_vec_col = unwrap_udt(col(features_col_name))
+
+            # After a `pyspark.ml.linalg.VectorUDT` type column being unwrapped, it becomes
```
trivialfis, line 584: Thank you for the detailed comments!

python-package/xgboost/spark/data.py
```diff
@@ -139,6 +140,76 @@ def append_m(part: pd.DataFrame, name: str, is_valid: bool) -> None:
...
+                        # we need to check the values field to get vector length.
+                        vec_size = len(vec_values)
+                        csr_indices = np.arange(vec_size, dtype=np.int32)
+                        csr_values = vec_values
+
+                    if n_features == 0:
+                        n_features = vec_size
+                    assert n_features == vec_size
+
+                    # remove zero elements from csr_indices / csr_values
```
trivialfis, line 176: Is this necessary? When `missing` is set to 0, XGBoost can remove those values.

trivialfis (commented):
Thank you for the work on sparse data support. Could you please fix the CI errors?

python-package/xgboost/spark/data.py
WeichenXu123: I remember the DMatrix if data using csc/csr matrix, then it ignore the missing argument but regard all inactive element in the sparse matrix as missing values.

If so, then keep zero elements and removing them represents 2 different semantic:
Keep these zero means it will be regarded as "zero" value feature,
Remove these zero elements means it will be regarded as missing value.

Is my understanding correct ?

trivialfis: The ctor for CSR matrix should be able to handle missing values (but not for the CSC, which would raise a warning). https://github.com/dmlc/xgboost/blob/20d1bba1bb8b3aa0ba3762e99e21117b416a8567/python-package/xgboost/data.py#L35

trivialfis: which is a good reminder that I should clear the difference.

WeichenXu123: @trivialfis 
OK, so for missing handling, CSR is the same with dense input (respect "missing" param), but CSC is different (ignore "missing" param and regard inactive elements as missing), right ?
We should document this in DMatrix doc.

python-package/xgboost/spark/core.py
```diff
@@ -527,17 +547,69 @@ def _fit(self, dataset):
 
         select_cols = [label_col]
         features_cols_names = None
-        if self.getOrDefault(self.features_cols):
-            features_cols_names = self.getOrDefault(self.features_cols)
-            features_cols = _validate_and_convert_feature_col_as_float_col_list(
-                dataset, features_cols_names
-            )
-            select_cols.extend(features_cols)
+        enable_sparse_data_optim = self.getOrDefault(self.enable_sparse_data_optim)
```
wbo4958, line 550: the _fit function is almost 200 lines, which is super huge, could we split this function?

python-package/xgboost/spark/core.py
```diff
@@ -527,17 +547,69 @@ def _fit(self, dataset):
...
+        if enable_sparse_data_optim:
+            from pyspark.ml.linalg import VectorUDT
+
+            if self.getOrDefault(self.missing) != 0.0:
```
wbo4958, line 554: could we put this checking into the _validate_params?

python-package/xgboost/spark/core.py
WeichenXu123: I am working on another Ranker estimator PR. We can do this refactor after these feature PRs merged. Otherwise fixing conflicts is annoying.

python-package/xgboost/spark/core.py
WeichenXu123: @trivialfis 
According to your comment: https://github.com/dmlc/xgboost/pull/8145#discussion_r943352000
Seemingly we don't need to add this restriction ?

trivialfis: I think we need the restriction that `missing` must be 0. Otherwise there will be two missing/invalid value, 0s removed by spark. `missing` removed by xgboost.

WeichenXu123: @trivialfis 

OK. So that CSR input is still different with dense input when handling "missing",
For CSR input, both inactive elements (removed elements in CSR) or elements of `value==${missing param value}` are regarded as missing value.

We should document this carefully. Dense / CSC / CSR are different on missing logic.

python-package/xgboost/spark/data.py
trivialfis: I will update the CSC implementation instead.

wbo4958:
BTW, pyspark 3.4 has not been released. Do we need to wait it?

WeichenXu123:
@wbo4958 No. The code still works on spark < 3.4, (but only this feature flag turn on it will raise error)
And I also add test on spark 3.4 (I install a spark 3.4 dev version in CI)

Is this PR good to merge ?

wbo4958:
sure, make sense. Thx

python-package/xgboost/spark/data.py
```diff
@@ -101,11 +102,55 @@ def reset(self) -> None:
...
+
+    n_features = 0
+
+    for vec_type, vec_size_, vec_indices, vec_values in zip(
+        part.featureVectorType,
+        part.featureVectorSize,
+        part.featureVectorIndices,
+        part.featureVectorValues,
+    ):
+        if vec_type == 0:
```
wbo4958, line 117: Please correct me if I'm wrong. now that the missing is 0, do we still really need the sparse vector? per my understanding, if one instance has a missing value, then the whole instance will be removed.

wbo4958:
Overall, LGTM, only one question left from me.

python-package/xgboost/spark/data.py
WeichenXu123: @trivialfis Is it true ? If so then training on sparse data makes no sense. Almost all instances will be removed ?

WeichenXu123: I think no, @wbo4958 You can check my test case `test_regressor_with_sparse_optim` and `test_classifier_with_sparse_optim`, every training instance contains missing value "0", but the generated model generates good prediction results.

WeichenXu123:
CC @trivialfis Thanks!

trivialfis (approved):
Looks good to me!

python-package/xgboost/spark/data.py
wbo4958: @trivialfis what does "updating CSC implementation" mean? so for current implementation, it indeed remove the whole instance if one element in the instance is missing value?

Issue #8146, opened 2022-08-09
Make data set available in Python package


ravimakhija:
Hi - this is a feature proposal for the Python package portion of `xgboost`. If there is interest, I would be happy to try implementing and provide a PR. 

#### Details:

The Getting Started documentation shows an example of loading the "agaricus" data set by using a file path. 

Link: https://xgboost.readthedocs.io/en/stable/get_started.html

I believe the use of a local-specific file path for the data provides unnecessary friction for Python users when initially trying out `xgboost`. Hence, the enhancement idea is to include the "agaricus" data set in the python package. This might work similar to how `sklearn` provides data sets, e.g. `data = sklearn.datasets.load_diabetes()`. For `xgboost` the interface might look something like `data = xgboost.datasets.load_agaricus()`. 

I would look forward to any feedback on this feature proposal.

trivialfis:
I think it might be better to change that getting started script to use some other easier-to-access datasets instead of copying the data into XGBoost Python wheel.

ravimakhija:
@trivialfis  Thanks for your thoughts! So long as there is no specific reason to use the "agricus" data set, that sounds good to me. 

**Two implementation options might be:**

Option 1: Just use `sklearn.datasets` directly to get a data set that works well for showcasing `xgboost`. 
Option 2: Download a data set from the web, possibly by using something like `wget` via Python

**Additional Thoughts:**

Option 1 is probably the best/quickest way to do it. However, if we want to keep the data set in sync for all examples (now or perhaps down the road), then we would also have to make sure there is an analogous approach for the other programming language options for the chosen data set. I'll try to look into that. 

Option 2 has the disadvantage that it requires having a reliable data source, and for new users to trust that data source. Also, it ends up using the user's file path which will likely require the user to do a slight modification before running the code. However, it's probably a more programming language-agnostic method.

trivialfis:
I think `sklearn.datasets` is an excellent choice. As for other languages, the XGBoost R package has the data packaged already so it should be fine.

PR #8153, opened 2022-08-10
Simplify Python getting started example

ravimakhija:
This PR aims to resolve issue #8146. 

#### Current behavior: 

The current Python [getting started example](https://xgboost.readthedocs.io/en/stable/get_started.html) requires the user to manually set the local file path to the Agaricus data set in libsvm format before being able to run the example code. 

#### New behavior: 

This PR leverages `sklearn` to load the data set. Hence, as long as the user has `sklearn` installed, after a `pip install xgboost`, the user should be able to run the getting started example as is (reducing friction, especially for new users). 

#### Additional Details:

- The well-known Iris data set is used, since the Agaricus data set is not available in `sklearn.datasets`. 
- The `xgboost.XGBClassifier` class is used here rather than the `xgboost.fit` function to train the model as the latter would require an extra step of converting form `numpy` arrays to `xgboost.DMatrix`.

### DIFF
--- a/doc/get_started.rst
+++ b/doc/get_started.rst
@@ -19,16 +19,18 @@ Python
 
 .. code-block:: python
 
-  import xgboost as xgb
-  # read in data
-  dtrain = xgb.DMatrix('demo/data/agaricus.txt.train')
-  dtest = xgb.DMatrix('demo/data/agaricus.txt.test')
-  # specify parameters via map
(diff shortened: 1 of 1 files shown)


trivialfis (approved):
This looks great!  Personally, I want to move XGBoost closer to sklearn and introduce the native interface only if necessary. ;-)

PR #8158, opened 2022-08-12
Simplify device sketch.

trivialfis:
* Use a single sketch container for all batches.


trivialfis:
Doesn't work due to incorrect row estimation.

PR #8139, opened 2022-08-04
Use CUDA docker image 11.0.3 tag.

trivialfis:

PR #8162, opened 2022-08-12
[backport] Update CUDA docker image and NCCL. (#8139)

trivialfis:


trivialfis:
Maybe I should backport everything around the CI instead.

PR #8173, opened 2022-08-14
[pyspark] Add param validation for "objective" and "eval_metric" param, and remove invalid booster params

WeichenXu123:
* Added param validation for "objective" and "eval_metric" param
* Removed invalid booster params passed to `xgboost.train`
* Allow regressor setting other objective value
* Handle the case running on spark GPU cluster but `spark.task.resource.gpu.amount` config is not set.


WeichenXu123:
CC @trivialfis @wbo4958

trivialfis (approved):

trivialfis:
Please fix the formatting error: https://github.com/dmlc/xgboost/runs/7824147126?check_suite_focus=true .

python-package/xgboost/spark/core.py
```diff
@@ -267,6 +267,18 @@ def _validate_params(self):
                 "If features_cols param set, then features_col param is ignored."
             )
 
+        if self.getOrDefault(self.objective) is not None:
```
wbo4958, line 270: It seems not make sense (to me) to check the type of a parameter that already has the typeConverter

WeichenXu123: The "objective" param is automatically generated, currently it does not have type converter.

python-package/xgboost/spark/core.py
```diff
@@ -345,7 +353,14 @@ def _validate_params(self):
                 )
             else:
                 # checking spark non-local mode.
-                if not gpu_per_task or int(gpu_per_task) < 1:
+                if not gpu_per_task:
+                    # For spark cluster with GPU config, the "spark.task.resource.gpu.amount"
+                    # config might not be set. In this case, the default gpu per task is 1.
+                    # If running on no GPU available, in spark task calling `_get_gpu_id` will
+                    # raise gpu not found error.
+                    gpu_per_task = "1"
```
WeichenXu123, line 361: @wbo4958 A notable fix.

WeichenXu123:
CC @trivialfis @wbo4958 Done.

python-package/xgboost/spark/core.py
WeichenXu123: My fault, reverted this.

python-package/xgboost/spark/core.py
wbo4958: I still hope we can find a common solution for the type converter in the following PR

wbo4958 (approved):

WeichenXu123:
@trivialfis @wbo4958 We can merge this. The test failures are not related to my PR.

PR #8179, opened 2022-08-17
[CI] Prune unused archs from libnccl

hcho3:
This further reduces the size of the compiled binary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,6 +141,10 @@ if (USE_CUDA)
   if ((${CMAKE_CUDA_COMPILER_VERSION} VERSION_GREATER_EQUAL 11.4) AND (NOT BUILD_WITH_CUDA_CUB))
     set(BUILD_WITH_CUDA_CUB ON)
   endif ()
+
+  if (USE_NCCL)
+    add_custom_target(PruneLibNccl nvprune ${GEN_CODE} /usr/lib64/libnccl_static.a -o libnccl_static.a)
+  endif (USE_NCCL)
 endif (USE_CUDA)
 
 if (FORCE_COLORED_OUTPUT AND (CMAKE_GENERATOR STREQUAL "Ninja") AND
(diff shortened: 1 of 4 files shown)


trivialfis (changes requested):
Thank you for working on this. I think this needs to be a CI specific change. People might not want XGBoost to prune the nccl on their system.

CMakeLists.txt
```diff
@@ -141,6 +141,10 @@ if (USE_CUDA)
   if ((${CMAKE_CUDA_COMPILER_VERSION} VERSION_GREATER_EQUAL 11.4) AND (NOT BUILD_WITH_CUDA_CUB))
     set(BUILD_WITH_CUDA_CUB ON)
   endif ()
+
+  if (USE_NCCL)
+    add_custom_target(PruneLibNccl nvprune ${GEN_CODE} /usr/lib64/libnccl_static.a -o libnccl_static.a)
```
RAMitchell, line 146: Should this path be hardcoded?

RAMitchell (commented):
Agree with Jiaming that users self installing probably don't need this, and it could easily break their install process.

Is it possible to create the option PRUNE_NCCL which performs the prune then modifies the nccl library path variable in find_nccl.

hcho3:
> People might not want XGBoost to prune the nccl on their system.

The custom target is entirely optional, as it's not part of the `all` target. So you need to explicitly opt in: `ninja PruneLibNccl`

hcho3:
> Is it possible to create the option PRUNE_NCCL which performs the prune then modifies the nccl library path variable in find_nccl.

This is hard, as you'd need to modify the nccl path as a result of building a target.

hcho3:
@trivialfis @RAMitchell I revised this pull request so that the changes are restricted to scripts in the CI directory `tests/ci_build`. Can you review again?

RAMitchell:
Thanks. I was curious if it can be done in pure cmake with a bit less hacking, here is my attempt: #8183. Not sure if it's any better.

RAMitchell:
Let's go with this version I think.

RAMitchell (approved):

trivialfis (approved):

PR #8181, opened 2022-08-17
Allow insecure gRPC connections for federated learning

rongou:
For NVFLARE integration, we plan to run a local gRPC proxy, which doesn't really need SSL.

@trivialfis


trivialfis:
Restarted the CI.

trivialfis (approved):

PR #8193, opened 2022-08-22
[WIP] Fix GPU categorical bug.

RAMitchell:
This fixes an off by one error in GPU partition based split calculation and changes the direction of the categorical split condition. The off by one error resulted in one category consistently being put on the wrong side, so training loss can diverge significantly. 

Before, a training instance that belongs to the set of a categorical split would take the right branch, now it takes the left branch. For me this feels like a simplification, as matching the split condition always goes left, and also avoids complication in split evaluation with the direction of scan and missing value direction.

I have not adjusted the CPU code yet, so CPU tests should fail, but this should be relatively simple.

### DIFF
--- a/src/common/categorical.h
+++ b/src/common/categorical.h
@@ -49,8 +49,6 @@ inline XGBOOST_DEVICE bool InvalidCat(float cat) {
 }
 
 /* \brief Whether should it traverse to left branch of a tree.
- *
- *  For one hot split, go to left if it's NOT the matching category.
  */
 template <bool validate = true>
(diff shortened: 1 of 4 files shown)


trivialfis:
Thank you for diving into this issue. Will follow up tomorrow.

hcho3:
> a training instance that belongs to the set of a categorical split would take the right branch, now it takes the left branch.

This is a breaking change. Can you mark this pull request as breaking?

trivialfis:
I'm not fully into the idea of switching the split direction yet. The original split direction is designed to be aligned with OHE, which helps user expectations and our tests.

PR #8204, opened 2022-08-26
Optimization/applysplit/common row partitioner

razdoburdin:
Hello all,
I continue to work on decomposition of https://github.com/dmlc/xgboost/pull/7208 into smaller and clearer parts. I tried to take into account the feedback from https://github.com/dmlc/xgboost/pull/7991

In the original PR https://github.com/dmlc/xgboost/pull/7208 @ShvetsKS introduced the CommonRowPartitioner. It is the common partitioner for "hist" and "approx" methods.

In the current PR I have generalized the ApproxRowPartitioner and HistRowPartitioner in the class [CommonRowPartitioner](src/tree/common_row_partitioner.h ) without applying any optimizations. Such change looks like the possible first move in the implementation of CPU optimizations from the original PR.

I used a logic from HistRowPartitioner in my CommonRowPartitioner  to both "hist" and "approx" methods. To the best of my understanding it should work fine for both cases. But maybe you see any potential problems with such a solution?


src/tree/updater_quantile_hist.cc
```diff
@@ -19,17 +17,167 @@
...
+#include "common_row_partitioner.h"
 #include "constraints.h"
+#include "hist/histogram.h"
 #include "hist/evaluate_splits.h"
 #include "param.h"
 #include "xgboost/logging.h"
 #include "xgboost/tree_updater.h"
 
 namespace xgboost {
+struct RandomReplace {
```
RAMitchell, line 29: Is this used somewhere?

src/data/gradient_index.cc
```diff
@@ -49,12 +49,8 @@ GHistIndexMatrix::GHistIndexMatrix(DMatrix *p_fmat, bst_bin_t max_bins_per_feat,
   this->columns_ = std::make_unique<common::ColumnMatrix>();
 
   // hessian is empty when hist tree method is used or when dataset is empty
-  if (hess.empty() && !std::isnan(sparse_thresh)) {
```
RAMitchell, line 52: Is removing this condition safe? If so, why?

RAMitchell (commented):
Looks good to me generally, assuming all tests are passing for approx after these changes.

src/tree/updater_quantile_hist.cc
razdoburdin: Yes. I didn't change this logic yet. So it is used in the [same place](https://github.com/dmlc/xgboost/blob/662289eec9a1b5bc3cb702ab43d38bc470c61a82/src/tree/updater_quantile_hist.cc#L441) as in the master branch

src/data/gradient_index.cc
razdoburdin: Removing of this condition is required for correct initialization of column matrix for 'approx' method. We don't see any problem with removing this condition. Do you see some potential problem in this case?

src/tree/updater_quantile_hist.cc
razdoburdin: > Looks good to me generally, assuming all tests are passing for approx after these changes.

The only failed test is python linting. All tests for 'approx' passed.

RAMitchell: Please remove it. We will get to that later.

src/data/gradient_index.cc
trivialfis: Hi, the `approx` tree method in its current form doesn't use the column matrix. Unlike `hist`, the `approx` needs to generate the histogram index for every iteration for most of the objective functions (like binary classification), and generating a column matrix is quite expensive for both time and memory so I skipped it here and defined a custom partitioner.

src/tree/updater_quantile_hist.cc
razdoburdin: Sorry,
your suggestion is not clear for me. The RandomReplce structure is currently used in [QuantileHistMaker::Builder::InitSampling](https://github.com/dmlc/xgboost/blob/662289eec9a1b5bc3cb702ab43d38bc470c61a82/src/tree/updater_quantile_hist.cc#L441). So it can not be deleted without changing the other code.
Can you please clarify your suggestion?

RAMitchell: I mean only in this pr. Add it later when you need it. Otherwise we are reviewing things where we don't know what they are used for. It's just about keeping your prs small and specific.

razdoburdin: done. Thanks for your advice.

src/data/gradient_index.cc
razdoburdin: I see, thanks.
So, what is your opinion about using the same partitioner both for hist and approx methods?

RAMitchell: I did notice that for the current PartitionBuilder, there are methods only used by approx method and methods only used by hist. So there is not actually much code reuse (maybe just the LeafPartition method?).

razdoburdin: You are right, the `LeafPartition` has different realizations for `hist` and `approx` methods. However the most part of the code (`UpdatePostion` and constructor) is used by both method.

razdoburdin:
> 
Hi, what is your decision about this PR?

RAMitchell (commented):
Sorry for late reply. Feel free to ping us if we are slow. LGTM, except I think the check on column matrix initialisation may need to be added back.

trivialfis:
Using column matrix for partition with approx is quite inefficient if `const_hess` is false.

razdoburdin:
Hi all, 
I have returned the check for column matrix initialization back. Unfortunately amount of reusable code becomes less in this case.

RAMitchell (approved):

trivialfis (commented):
Is the column matrix still initialized for approx tree method? Asking since as mentioned previously the construction of it is quite costly for approx since it happens every iteration.

razdoburdin:
> Is the column matrix still initialized for approx tree method? Asking since as mentioned previously the construction of it is quite costly for approx since it happens every iteration.

Sorry, it was my fault. I returned check only in one place of code, not in both. The ColumnMatrix is not initialized for approx method now.

trivialfis:
Will take a deeper look.

src/common/column_matrix.h
```diff
@@ -167,6 +171,7 @@ class ColumnMatrix {
    */
   void InitFromSparse(SparsePage const& page, const GHistIndexMatrix& gmat, double sparse_threshold,
                       int32_t n_threads) {
+    is_initialized_ = true;
```
trivialfis, line 174: We have a few places for initialization. From adapter, from ghist index matrix, and from external memory.

src/tree/common_row_partitioner.h
```diff
@@ -0,0 +1,233 @@
...
+      const size_t n_right = partition_builder_.GetNRightElems(i);
+      CHECK_EQ((*p_tree)[nid].LeftChild() + 1, (*p_tree)[nid].RightChild());
+      row_set_collection_.AddSplit(nid, (*p_tree)[nid].LeftChild(), (*p_tree)[nid].RightChild(),
+                                  n_left, n_right);
+    }
+  }
+
+  void UpdatePosition(GenericParameter const* ctx, GHistIndexMatrix const& gmat,
+                      std::vector<CPUExpandEntry> const& nodes, RegTree const* p_tree) {
+    auto const& column_matrix = gmat.Transpose();
```
trivialfis, line 95: Out of curiosity, if there's no column matrix generation for approx, how does this partitioner handle approx?

razdoburdin: To avoid column matrix generation, I reduced code sharing between `hist` and `approx `methods. Currently partitioner use different logic for case with initialized column matrix and for the opposite one. Here is the [decision point](https://github.com/razdoburdin/xgboost/blob/fc260e09935f34dcfd780e5a08226ac3716dc43f/src/common/partition_builder.h#L169)

trivialfis:
Thank you for the great work on unifying the partitioner!

PR #8212, opened 2022-08-30
Add `max_cat_threshold` to GPU and handle missing cat values.

trivialfis:
related: https://github.com/dmlc/xgboost/pull/8193

- Port the `max_cat_threshold` parameter to GPU.
- Use a 2-pass scan on GPU instead of a pseudo-bin, with missing values being handled.


RAMitchell (commented):
This PR is redoing #8193? 

One problem here is that the missing value reduction is still being calculated on the constructor of EvaluateSplitAgent but not used. Ideally I think we want numerical and one hot splits all doing the same thing (although not necessary for this PR). This evaluate split kernel is quite delicate in terms of performance. If you change everything to do forward and backwards passes you might find it slows down significantly or even speeds up. Epsilon is a good test case for numerical splits.

If we merge this I think you should take the edge case tests from #8193, although I reversed the split direction in that PR so they need to be inverted. Those test cases guard against the bugs I saw in the categorical airline dataset.

trivialfis:
> This PR is redoing https://github.com/dmlc/xgboost/pull/8193?

In a slightly different way.

> One problem here is that the missing value reduction is still being calculated on the constructor of EvaluateSplitAgent but not used

Yes, that's a problem. I can skip that by spliting up the Agent class.

> If we merge this I think you should take the edge case tests from

The 2 approaches are not entirely the same. Using 2-pass scan helps with an edge case where we need to avoid including the last bin into split calculation. We would like to avoid spliting only on missing value.

RAMitchell:
Splitting only on missing values seems valid.

trivialfis:
Copied the tests from https://github.com/dmlc/xgboost/pull/8193 and added some checks for default direction.

tests/cpp/tree/gpu_hist/test_evaluate_splits.cu
```diff
@@ -17,29 +18,293 @@ auto ZeroParam() {
...
+    EvaluateSplitInputs input{2, 0, parent_sum, dh::ToSpan(feature_set),
+                              dh::ToSpan(feature_histogram)};
+    DeviceSplitCandidate result = evaluator.EvaluateSingleSplit(input, shared_inputs).split;
+    EXPECT_EQ(result.dir, kLeftDir);
+    EXPECT_FLOAT_EQ(result.loss_chg, 0.0f);
+    EXPECT_FLOAT_EQ(result.left_sum.GetGrad() + result.right_sum.GetGrad(), parent_sum.GetGrad());
+    EXPECT_FLOAT_EQ(result.left_sum.GetHess() + result.right_sum.GetHess(), parent_sum.GetHess());
+  }
+  // With 3.0/3.0 missing values
+  // Forward, first 2 categories are selected, while the last one go to left along with missing value
```
RAMitchell, line 133: This definitely seems weird to me to arbitrarily keep one category along with the missing values and have a worse loss.

If we can solve the numerical issues I definitely think missing only splits should be allowed.

trivialfis: Sounds good. Looking forward to your findings!

Issue #8186, opened 2022-08-21
[pyspark] Document model format.


trivialfis:
https://discuss.xgboost.ai/t/save-pyspark-model-load-it-on-other-language/2894

DavidMaister:
Hey, I managed to solve this so I'll share how I've solved. 

- The SparkXGBClassifier saving the model, saves a directory with parquet files.
- Read these folder with parquet files on pyspark, write these object as json.
- In python, load this json and we want to change the format because it doesnt work the original one.
```
import json
f = open('pyspark_raw_json.json')
pyspark_raw_json = json.load(f)
with open('pyspark_json_converted.json', 'w', encoding='utf-8') as f:
    json.dump(json.loads(list(pyspark_raw_json.values())[0]), f)
```
- Then I am able to load the model in xgboost
```
loaded_model = xgb.XGBClassifier()
loaded_model.load_model('pyspark_model_converted.json')
```

wbo4958:
I think we need to change the "save" to the model that xgboost can use directly. Let me resolve this.

PR #8219, opened 2022-09-02
[pyspark] make the model saved by pyspark compatible

wbo4958:
Users can't directly load the model using xgboost python package trained by pyspark. it requires much effort to do that, see https://github.com/dmlc/xgboost/issues/8186. This PR first saves the model in JSON format and then writes it to txt file. Then the user can easily load the model by

``` python
import xgboost as xgb
bst = xgb.Booster()

YOUR_MODEL_PATH="xxx"
bst.load_model(YOUR_MODEL_PATH/model/part-00000)
```


wbo4958:
@trivialfis Could you help to check a python test failed 

``` console

[2022-09-05T12:59:12.999Z] =================================== FAILURES ===================================

[2022-09-05T12:59:12.999Z] ____________________________ test_gpu_data_iterator ____________________________

[2022-09-05T12:59:12.999Z] 

[2022-09-05T12:59:12.999Z] cls = <class '_pytest.runner.CallInfo'>
```

wbo4958:
@WeichenXu123 @trivialfis Could you help to review this PR?

trivialfis:
Will look into it tomorrow.

wbo4958:
@WeichenXu123 @trivialfis Any feedback for this issue?

wbo4958:
Hi @WeichenXu123 @trivialfis, could you help to review it?

trivialfis:
Can we document the function `get_booster(self)` and let the user extract the booster?

wbo4958:
Yeah, We can, but the issue will be the same with the one JVM package previously encountered. Most users dump the mode by the spark way, they may don't like to do another `get_booster.save_model` again. So the model may be moved to another machine (or another team without any knowledge of spark) without spark cluster deployed since users may just want to load the model with python package and do some prediction. in that case, it's really un-convenient for users. This PR is supposed not to introduce any side effects, so I think it's ok to be merged.

python-package/xgboost/spark/model.py
```diff
@@ -222,11 +216,11 @@ def saveImpl(self, path):
...
         xgb_model = self.instance._xgb_sklearn_model
         _SparkXGBSharedReadWrite.saveMetadata(self.instance, path, self.sc, self.logger)
-        model_save_path = os.path.join(path, "model.json")
-        ser_xgb_model = serialize_xgb_model(xgb_model)
-        _get_spark_session().createDataFrame(
-            [(ser_xgb_model,)], ["xgb_sklearn_model"]
-        ).write.parquet(model_save_path)
+        model_save_path = os.path.join(path, "model")
+        xgb_model_file = dump_model_to_json_file(xgb_model)
+        # The json file written by Spark base on `booster.save_raw("json").decode("utf-8")`
```
trivialfis, line 221: Why not?

wbo4958: There are some " \" " in the json file which can't be loaded by xgboost. Do you want to check more?

python-package/xgboost/spark/model.py
```diff
@@ -21,34 +21,28 @@ def _get_or_create_tmp_dir():
     return xgb_tmp_dir
 
 
-def serialize_xgb_model(model):
+def dump_model_to_json_file(model) -> str:
```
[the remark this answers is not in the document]
wbo4958: Done

python-package/xgboost/spark/model.py
trivialfis: I will take a look tomorrow

python-package/xgboost/spark/model.py
```diff
@@ -222,11 +216,11 @@ def saveImpl(self, path):
...
-        model_save_path = os.path.join(path, "model.json")
-        ser_xgb_model = serialize_xgb_model(xgb_model)
-        _get_spark_session().createDataFrame(
-            [(ser_xgb_model,)], ["xgb_sklearn_model"]
-        ).write.parquet(model_save_path)
+        model_save_path = os.path.join(path, "model")
+        xgb_model_file = save_model_to_json_file(xgb_model)
+        # The json file written by Spark base on `booster.save_raw("json").decode("utf-8")`
+        # can't be loaded by XGBoost directly.
+        _get_spark_session().read.text(xgb_model_file).write.text(model_save_path)
```
WeichenXu123, line 223:
```
_get_spark_session().read.text(xgb_model_file).
```
This line is not correct.
spark.read.text(path) the path must be a distributed file system path which all spark executor can access.

WeichenXu123: You can use distributed FS API to copy local file `xgb_model_file` into the model saved path (a hadoop FS path)

wbo4958: wow, right. you're correct, @WeichenXu123 Good findings. Could you point me to what is the "distributed FS API"? Really appreciate it.

WeichenXu123: You can use this:
https://arrow.apache.org/docs/python/generated/pyarrow.fs.HadoopFileSystem.html

But, this does not support DBFS (databricks filesystem), we need support databricks case as well.
Databricks mount `dbfs:/xxx/xxx` to local file system `/dbfs/xxx/xxx`.

WeichenXu123: The example code in the PR description 
```
import xgboost as xgb
bst = xgb.Booster()

# Basically, YOUR_MODEL_PATH should be like "xxxx/model/xxx.txt"
YOUR_MODEL_PATH="xxx"
bst.load_model(YOUR_MODEL_PATH)
```
seems does not wok ? If the path is a distributed FS path ?

WeichenXu123:
Do we really need this PR ?
User can load pyspark model and then call `pyspark_model.booster` to get the raw booster model.

python-package/xgboost/spark/model.py
wbo4958: @trivialfis No need anymore, I just found another way to do it.

python-package/xgboost/spark/model.py
wbo4958: @WeichenXu123, I use the RDD to save the text file, it should work with all kinds of hadoop-compatible FS..

wbo4958:
> Do we really need this PR ? User can load pyspark model and then call `pyspark_model.booster` to get the raw booster model.

Yeah, guess the scenario, the data scientist who does not know spark gets a model saved by xgboost-spark and wants to load it by xgboost python package, what does he/she can do? 

Although we can doc it, trust me, not everyone would like to read the whole doc carefully. Previously, XGBoost-JVM has the same issue, so I changed that.

wbo4958:
Hi @hcho3, what does "Pending" mean for pipelines like xgboost-ci/pr?

python-package/xgboost/spark/model.py
```diff
@@ -222,11 +200,11 @@ def saveImpl(self, path):
...
         _SparkXGBSharedReadWrite.saveMetadata(self.instance, path, self.sc, self.logger)
-        model_save_path = os.path.join(path, "model.json")
-        ser_xgb_model = serialize_xgb_model(xgb_model)
-        _get_spark_session().createDataFrame(
-            [(ser_xgb_model,)], ["xgb_sklearn_model"]
-        ).write.parquet(model_save_path)
+        model_save_path = os.path.join(path, "model")
+        booster = xgb_model.get_booster().save_raw("json").decode("utf-8")
+        _get_spark_session().sparkContext.parallelize([booster], 1).saveAsTextFile(
+            model_save_path
```
WeichenXu123, line 206: Interesting idea, but how to control the saved file name ?

WeichenXu123: Does the `booster` string contain "\n" character ?

wbo4958: It is always part-00000

WeichenXu123: Let's document the file name "part-00000" is the model json file.

and pls add a test to ensure the model json file does not contain `\n` character and document the reason.

wbo4958: Just checked the code, the file name is defined by https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/internal/io/SparkHadoopWriter.scala#L225.

``` scala
  override def initWriter(taskContext: NewTaskAttemptContext, splitId: Int): Unit = {
    val numfmt = NumberFormat.getInstance(Locale.US)
    numfmt.setMinimumIntegerDigits(5)
    numfmt.setGroupingUsed(false)

    val outputName = "part-" + numfmt.format(splitId)
    val path = FileOutputFormat.getOutputPath(getConf)
    val fs: FileSystem = {
      if (path != null) {
        path.getFileSystem(getConf)
      } else {
        // scalastyle:off FileSystemGet
        FileSystem.get(getConf)
        // scalastyle:on FileSystemGet
      }
    }
...
```

 here the **splitId** is the TaskContext.partitionId(). In our case, there is only 1 partition, so the file name is "part-00000"

WeichenXu123: yes I know that. My point is can we customize the file name to make it more user-friendly.
Not a must though.

hcho3:
> what does "Pending" mean for pipelines like xgboost-ci/pr?

The CI pipeline doesn't run until one of the admins (like me) give approval. We do this to save the CI costs.

python-package/xgboost/spark/model.py
trivialfis: that's the internal behavior of pyspark, not sure if it's a good idea to rely on it.

wbo4958: Yeah, If you guys insist, I can use the FileSystem java API to achieve it by py4j.

WeichenXu123: > Yeah, If you guys insist, I can use the FileSystem java API to achieve it by py4j.

No need to do that, it makes code hard to maintain, your current code is fine.

tests/python/test_spark/test_spark_local.py
```diff
@@ -427,6 +435,12 @@ def setUp(self):
     def get_local_tmp_dir(self):
         return self.tempdir + str(uuid.uuid4())
 
+    def assert_model_compatible(self, model: XGBModel, model_path: str):
+        bst = xgb.Booster()
+        path = glob.glob(f"{model_path}/**/model/part-00000", recursive=True)[0]
+        bst.load_model(path)
+        self.assertEqual(model.get_booster().save_raw("json"), bst.save_raw("json"))
```
WeichenXu123, line 442: add a test to assert model file does not include `\n` char.

wbo4958: Yeah, per my understanding, seems we don't need to do this, since if there is "\n", the assertion must be failed `self.assertEqual(model.get_booster().save_raw("json"), bst.save_raw("json"))`

trivialfis:
I will leave the approval to @WeichenXu123 . Could you please add document as well? About the get_booster and your workaround for the model serialization.

wbo4958:
Sure, I will add the doc in the following PR along with how to leverage RAPIDS to accelerate xgboost pyspark.

@trivialfis @hcho3 could you trigger the CI of this PR

PR #8215, opened 2022-09-01
Copy data from Ellpack to GHist.

trivialfis:
This allows XGBoost to train with `hist` tree method on CPU with GPU-initialized `QuantileDMatrix`.

Related: https://github.com/dmlc/xgboost/issues/7890 .

PR #8220, opened 2022-09-02
Make `QuantileDMatrix` default to sklearn esitmators.

trivialfis:
Stacked on https://github.com/dmlc/xgboost/pull/8215 .

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -726,10 +726,9 @@ def _create_quantile_dmatrix(
     if parts is None:
         msg = f"worker {worker.address} has an empty DMatrix."
         LOGGER.warning(msg)
-        import cupy
 
         d = QuantileDMatrix(
-            cupy.zeros((0, 0)),
+            numpy.empty((0, 0)),
             feature_names=feature_names,
             feature_types=feature_types,
             max_bin=max_bin,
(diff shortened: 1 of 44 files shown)


trivialfis:
Todo:
- [ ] Handle empty DMatrix from CUDA inputs.
- [ ] Better tests.

python-package/xgboost/dask.py
```diff
@@ -1544,15 +1543,21 @@ def inplace_predict(  # pylint: disable=unused-argument
 
 
 async def _async_wrap_evaluation_matrices(
-    client: Optional["distributed.Client"], **kwargs: Any
+    client: Optional["distributed.Client"],
+    tree_method: Optional[str],
+    max_bin: Optional[int],
+    **kwargs: Any,
```
hcho3, line 1549: It appears that `tree_method` and `max_bin` will no longer be in the `kwargs` dictionary. Can you make sure this will not cause undesirable behavior? For example, is `kwargs` passed to `xgb.train`?

python-package/xgboost/sklearn.py
```diff
@@ -904,6 +908,17 @@ def _duplicated(parameter: str) -> None:
...
         return model, metric, params, early_stopping_rounds, callbacks
 
+    def _create_dmatrix(self, ref: Optional[DMatrix], **kwargs: Any) -> DMatrix:
+        # Use `QuantileDMatrix` to save memory.
+        if self.tree_method in ("hist", "gpu_hist"):
+            try:
+                return QuantileDMatrix(
+                    **kwargs, ref=ref, nthread=self.n_jobs, max_bin=self.max_bin
+                )
+            except TypeError:  # `QuantileDMatrix` supports lesser types than DMatrix
```
RAMitchell, line 918: Which cases are not supported out of curiosity? Just wondering if this is going to run into the exception regularly in some cases.

src/data/iterative_dmatrix.cc
```diff
@@ -14,6 +14,37 @@
...
+
+  auto d = MakeProxy(proxy_)->DeviceIdx();
+
+  StringView msg{"All batch should be on the same device."};
+  if (batch_param_.gpu_id != Context::kCpuId) {
+    CHECK_EQ(d, batch_param_.gpu_id) << msg;
+  }
+
+  batch_param_ = BatchParam{d, max_bin};
+  batch_param_.sparse_thresh = 0.2;  // default from TrainParam
```
RAMitchell, line 37: If we change the default, this hardcoded value will be forgotten.

RAMitchell (approved):

python-package/xgboost/sklearn.py
trivialfis: datatable, CSC, arrow.  Also, the DMatrix has a dispatcher to convert unknown types to scipy csr.

src/data/iterative_dmatrix.cc
trivialfis: Let me fix it.

python-package/xgboost/dask.py
trivialfis: Training parameters are obtained via `params = self.get_xgb_params()`, which you can find in the `async def _fit_async` function.

src/data/iterative_dmatrix.cc
trivialfis: Changed to obtain the value from train param.

Issue #7725, opened 2022-03-11
Implement IPv6 support for distributed XGBoost


ikriv:
Currently distributed XGBoost does not support IPv6, this is explicitly asserted in rabit/include/rabit/internal/socket.h
Our internal network is IPv6 only, which means we can't run XGBoost out of the box.

IPv6 is 20+ years old, and IPv4 is being phased out more and more as the time progresses. I think it is very important to enable IPv6 in XGBoost for more successful adoption.

trivialfis:
Hi @ikriv , thank you for raising the issue.  Would you like to work on this?

trivialfis:
I will take a deeper look into this.

trivialfis:
I will try to implement it after 1.6.

Note to myself:
- [ ] Check NCCL support.
- [ ] Change socket wrapper to support IPV6.
- [ ] Implement it in Python tracker.
- [ ] Possible change in Scala tracker.

ikriv:
Thank you so much, @trivialfis! I think Python tracker is already good (at least it does not complain loudly), it's the socket wrapper that insists on IPv4

trivialfis:
Not yet.  I will update this thread.

trivialfis:
Hi, can anyone here help with providing guidance on testing IPv6 on a public cloud like AWS or GCP? I'm not an expert on network administrating and got lost in those cloud documents.

The draft PR is here: https://github.com/dmlc/xgboost/pull/8225 .

trivialfis:
I tested the code using GCP VM instances. Feel free to try it if you are interested in this.

zhangleiray007:
Q: is there any plan for ipv6 support on pyspark :)

trivialfis:
Would be great if there people with expertise can help work on this. It takes some effort to setup a test environment in the CI and we don't have any admin.

CrazyBinXXX:
Is spark ipv6 supported now? If not I'd like to contribute.

PR #8225, opened 2022-09-05
Initial support for IPv6

trivialfis:
- Merge rabit socket into XGBoost.
- Dask interface support.
- Add test for socket.
- Add more compatibility code with mysys32, including correct linking, and enable most of
  the system function calls.
- Workaround windows includes in tests. dmlc filesystem includes `windows.h`, which
  conflicts with winsock2.h.

Current status:
This PR enables initial support for IPv6 with dask on Linux. Some effort is put into Windows support but we don't support distributed training on Windows in general. As for macos, RABIT allreduce returns error but I don't have a macos device for development or testing.

I ran some basic tasks using dask on GCP debian-based instance with dual stack and verified that IPv6 is indeed used.

https://github.com/dmlc/xgboost/issues/7725 .


trivialfis:
The use of `collective` is to be consistent with https://github.com/dmlc/xgboost/pull/8057

rabit/src/allreduce_base.h
```diff
@@ -201,8 +201,8 @@ class AllreduceBase : public IEngine {
     }
   };
   /*! \brief translate errno to return type */
-  inline static ReturnType Errno2Return() {
-    int errsv = utils::Socket::GetLastError();
+  static ReturnType Errno2Return() {
+    int errsv = errno;
```
trivialfis, line 205: fix this.

trivialfis:
@rongou Would you like to take a look into this?

doc/tutorials/dask.rst
```diff
@@ -504,6 +503,35 @@ dask config is used:
...
 
+
+************
+IPv6 Support
+************
+
+.. versionadded:: 2.0.0
+
+XGBoost has initial IPv6 support for the dask interface on Linux. Due to most of the
+cluster support for IPv6 is partial (duo stack instead of IPv6 only), we require
```
rongou, line 514: dual stack

rongou (approved):

RAMitchell (commented):
I can't really say much about the implementation as I don't have experience here.

It does seem a bit unusual to me that we are doing this ourselves and not using a library. Would using another framework for communication give us IPv6?

If you are confident with this PR you should go ahead with it.

trivialfis:
@RAMitchell We use to treat rabit as a separate library. ;-) But as suggested by @rongou , we can switch to something well maintained for allreduce implementation.

PR #8226, opened 2022-09-06
Drop use of deleted virtual function to support older MacOS

hcho3:
Deleted virtual functions are not supported in earlier versions of MacOS (<10.15). See https://github.com/JuliaPackaging/Yggdrasil/pull/5430.


trivialfis (commented):
Is it possible to use macro instead of removing it completely?

trivialfis (approved):

trivialfis (approved):

Issue #8221, opened 2022-09-02
[pyspark] SparkXGBClassifier failed to train with early_stopping_rounds and validation_indicator_col


Faaany:
Following the official SparkXGBClassifier example shown [here ](https://xgboost.readthedocs.io/en/latest/python/python_api.html#module-xgboost.spark), I try to run XGBoost training on my own dataset and get the following error: 

```bash
22/09/02 17:05:15 ERROR PythonRunner: Python worker exited unexpectedly (crashed)
org.apache.spark.api.python.PythonException: Traceback (most recent call last):
  File "/opt/spark-3.3.0-bin-hadoop3/python/lib/pyspark.zip/pyspark/worker.py", line 666, in main
    eval_type = read_int(infile)
  File "/opt/spark-3.3.0-bin-hadoop3/python/lib/pyspark.zip/pyspark/serializers.py", line 595, in read_int
    raise EOFError
EOFError
        at org.apache.spark.api.python.BasePythonRunner$ReaderIterator.handlePythonException(PythonRunner.scala:559)
        at org.apache.spark.api.python.PythonRunner$$anon$3.read(PythonRunner.scala:765)
        at org.apache.spark.api.python.PythonRunner$$anon$3.read(PythonRunner.scala:747)
        at org.apache.spark.api.python.BasePythonRunner$ReaderIterator.hasNext(PythonRunner.scala:512)
        at org.apache.spark.InterruptibleIterator.hasNext(InterruptibleIterator.scala:37)
        at scala.collection.Iterator.foreach(Iterator.scala:943)
        at scala.collection.Iterator.foreach$(Iterator.scala:943)
        at org.apache.spark.InterruptibleIterator.foreach(InterruptibleIterator.scala:28)
        at scala.collection.generic.Growable.$plus$plus$eq(Growable.scala:62)
        at scala.collection.generic.Growable.$plus$plus$eq$(Growable.scala:53)
        at scala.collection.mutable.ArrayBuffer.$plus$plus$eq(ArrayBuffer.scala:105)
        at scala.collection.mutable.ArrayBuffer.$plus$plus$eq(ArrayBuffer.scala:49)
        at scala.collection.TraversableOnce.to(TraversableOnce.scala:366)
        at scala.collection.TraversableOnce.to$(TraversableOnce.scala:364)
        at org.apache.spark.InterruptibleIterator.to(InterruptibleIterator.scala:28)
[...]

PR #8231, opened 2022-09-08
[pyspark] Fix xgboost spark estimator dataset repartition issues

WeichenXu123:

Fix xgboost spark estimator dataset repartition issues

1. Fix issue of repartition generating empty partitions, via repartition on `rand(1)` column.
2. For num_worker=1, if setting force_repartition=True, does not raise error but still repartition to 1 partition.
3. When `validationIndicatorCol` set, always trigger repartition.

Closes https://github.com/dmlc/xgboost/issues/8221

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -20,7 +20,7 @@ from pyspark.ml.param.shared import (
     HasWeightCol,
 )
 from pyspark.ml.util import MLReadable, MLWritable
-from pyspark.sql.functions import col, countDistinct, pandas_udf, struct, rand
+from pyspark.sql.functions import col, countDistinct, pandas_udf, rand, struct
 from pyspark.sql.types import (
     ArrayType,
     DoubleType,
@@ -168,7 +168,7 @@ class _SparkXGBParams(
         Params._dummy(),
         "repartition_random_shuffle",
         "A boolean variable. Set repartition_random_shuffle=true if you want to random shuffle "
-        "dataset when repartitioning is required. By default is True."
+        "dataset when repartitioning is required. By default is True.",
     )
     feature_names = Param(
         Params._dummy(), "feature_names", "A list of str to specify feature names."
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -541,9 +541,6 @@ class _SparkXGBEstimator(Estimator, _SparkXGBParams, MLReadable, MLWritable):
         query_plan = dataset._sc._jvm.PythonSQLUtils.explainString(
             dataset._jdf.queryExecution(), "extended"
         )
-        get_logger(self.__class__.__name__).warning(
-            f"debug-repartition: \n{query_plan}\n"
-        )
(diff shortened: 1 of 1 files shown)


WeichenXu123:
CC @trivialfis @wbo4958 Thanks!

python-package/xgboost/spark/core.py
```diff
@@ -691,7 +684,10 @@ def _fit(self, dataset):
             )
 
         if self._repartition_needed(dataset):
-            dataset = dataset.repartition(num_workers)
+            # Repartition on `monotonically_increasing_id` column to avoid repartition
+            # result unbalance. Directly using `.repartition(N)` might result in some
```
wbo4958, line 688: May I ask, In which scenario the `data.repartition(num_workers)` might result in some empty partitions?

WeichenXu123: I think it is because of the round robin repartition issue which causes no records shuffled to some partitions

wbo4958: Does the "round robin" really cause empty partition? I even can't think of a real case that can cause this issue.

Have you ever encountered the data.repartition(N) to cause the empty partition?

BTW, Does it cause the performance issue here we call repartition twice?

WeichenXu123: I am still debugging...

WeichenXu123:
CC @wbo4958

python-package/xgboost/spark/core.py
```diff
@@ -690,8 +681,19 @@ def _fit(self, dataset):
...
+            and self.getOrDefault(self.validationIndicatorCol)
+        ):
+            # If validationIndicatorCol defined, we always repartition dataset
+            # to balance data, because user might unionise train and validation dataset,
+            # without shuffling data then some partitions might contain only train or validation
+            # dataset.
+            # Repartition on `rand` column to avoid repartition
+            # result unbalance. Directly using `.repartition(N)` might result in some
+            # empty partitions.
+            dataset = dataset.repartition(num_workers, rand(1))
```
wbo4958, line 695: I still prefer to add a parameter to control if random shuffle. Since my PR https://github.com/dmlc/xgboost/pull/8245 can detect if the training data is empty DMatrix, so we can add some log printing about data skew and how to resolve it by enabling xxx parameter. Does that make sense?

hcho3:
Retriggering the CI, as we recently transitioned away from Jenkins

WeichenXu123:
@wbo4958 Is the PR good to merge ? (except the lint errors)

trivialfis:
Hi, could you please share the current status of this fix? Is it still necessary after @wbo4958 's PR?

wbo4958:
No, we can file the following PR  for it. This PR looks good to me.

wbo4958 (approved):

WeichenXu123:
@trivialfis @wbo4958 Ready for merging ! :)

wbo4958:
+1 ed

wbo4958 (approved):

trivialfis (approved):

PR #8218, opened 2022-09-02
Optimization/buildhist/hist util

razdoburdin:
Hi,
this small PR provides performance improvement for BuildHistKernel by helping the compiler to generate faster binary. For performance measurements I booked the c6i.12xlarge (24 cores, hyperthreading is off) instance on on Amazon Web Services and used the benchmarks from [here](https://github.com/IntelPython/scikit-learn_bench).

Each dataset was executed five times. The average run time is calculated by the formula
average = (run1 + run2 + run3 + run4 + run5 - min - max) / 3. I.e. the fastest and the slowest runs are cuted off, then other three are averaged. The summary of measurements are in the following table:
(Note that speed up value of the mlsr dataset is not a representative, due to runtime speed of this dataset is very unstable).

|  Dataset name       | master run1    | master run2    | master run3    | master run4    | master run5    | master average  | this PR run1    | this PR run2   | this PR run3    | this PR run4    | this PR run5    | this PR average   | speedup |
| --------------------- | -------- | -------- | -------- | -------- | -------- | -------- | -------- | -------- | -------- | -------- | -------- | -------- | ------- |
| abalone               | 4.10E-01 | 4.03E-01 | 4.11E-01 | 3.99E-01 | 4.05E-01 | **4.06E-01** | 4.02E-01 | 4.01E-01 | 3.94E-01 | 3.97E-01 | 3.96E-01 | **3.98E-01** | **1.02**    |
| airline-ohe           | 4.09E+01 | 4.20E+01 | 4.00E+01 | 3.99E+01 | 4.20E+01 | **4.10E+0**1 | 3.37E+01 | 3.37E+01 | 3.37E+01 | 3.27E+01 | 3.39E+01 | **3.37E+01** | **1.22**    |
| higgs1m               | 1.04E+01 | 1.04E+01 | 1.04E+01 | 1.04E+01 | 1.04E+01 | **1.04E+01** | 9.26E+00 | 9.27E+00 | 9.27E+00 | 9.32E+00 | 9.23E+00 | **9.27E+00** | **1.12**    |
| letters               | 1.04E+01 | 1.04E+01 | 1.04E+01 | 1.02E+01 | 1.02E+01 | **1.03E+01** | 1.01E+01 | 1.00E+01 | 1.00E+01 | 1.00E+01 | 1.00E+01 | **1.00E+01** | **1.03**    |
[...]

PR #8233, opened 2022-09-08
Optimization/buildhist/colwisebuildhist

razdoburdin:
Hi,
this PR is a part of https://github.com/dmlc/xgboost/pull/7192. It continues the optimization of BuildHistKernel, that was started in https://github.com/dmlc/xgboost/pull/8218.
Here I introduce a column wise building of histogram and a dispatcher for choice between the row wise and the column wise kernels. This optimization allow to improve runtime speed for several datasets up to 2x. For performance measurements I booked the c6i.12xlarge (24 cores, hyperthreading is off) instance on on Amazon Web Services and used the benchmarks from [here](https://github.com/IntelPython/scikit-learn_bench).

Methodology of the benchmarking is the same as in https://github.com/dmlc/xgboost/pull/8218. The only two datasets from the list affecting by this optimization are santander and epsilon. For this reason only them are shown in the following table:
|                       | master    | this PR   | speedup |
| --------------------- | -------- | -------- | ------- |
| santander             | 1.02E+02 | 4.89E+01 | 2.09    |
| epsion                | 1.84E+02 | 1.03E+02 | 1.78    |

I am looking forward for your review and comments!

upg:
I have deleted the optimization related to santander dataset for simplification of the review.

### DIFF
--- a/src/common/column_matrix.cc
+++ b/src/common/column_matrix.cc
@@ -23,10 +23,12 @@ void ColumnMatrix::InitStorage(GHistIndexMatrix const& gmat, double sparse_thres
   gmat.GetFeatureCounts(feature_counts.data());
 
   // classify features
+  any_sparse_column_ = false;
   for (bst_feature_t fid = 0; fid < nfeature; ++fid) {
     if (static_cast<double>(feature_counts[fid]) < sparse_threshold * nrow) {
       type_[fid] = kSparseColumn;
       all_dense_column = false;
+      any_sparse_column_ = true;
     } else {
       type_[fid] = kDenseColumn;
     }
--- a/src/common/column_matrix.cc
+++ b/src/common/column_matrix.cc
@@ -23,12 +23,10 @@ void ColumnMatrix::InitStorage(GHistIndexMatrix const& gmat, double sparse_thres
   gmat.GetFeatureCounts(feature_counts.data());
 
   // classify features
-  any_sparse_column_ = false;
   for (bst_feature_t fid = 0; fid < nfeature; ++fid) {
     if (static_cast<double>(feature_counts[fid]) < sparse_threshold * nrow) {
       type_[fid] = kSparseColumn;
       all_dense_column = false;
-      any_sparse_column_ = true;
     } else {
       type_[fid] = kDenseColumn;
     }
--- a/src/common/hist_util.cc
+++ b/src/common/hist_util.cc
@@ -264,40 +264,21 @@ void BuildHistKernel(const std::vector<GradientPair> &gpair,
   }
 }
 
-template <bool do_prefetch, bool first_page, bool any_missing>
-void BuildHistDispatch(const std::vector<GradientPair> &gpair,
-                       const RowSetCollection::Elem row_indices, const GHistIndexMatrix &gmat,
-                       GHistRow hist, bool read_by_column) {
-  switch (gmat.index.GetBinTypeSize()) {
-  case kUint8BinsTypeSize:
-    BuildHistKernel<do_prefetch, uint8_t, first_page, any_missing>
-                   (gpair, row_indices, gmat, hist, read_by_column);
-    break;
-  case kUint16BinsTypeSize:
-    BuildHistKernel<do_prefetch, uint16_t, first_page, any_missing>
-                   (gpair, row_indices, gmat, hist, read_by_column);
-    break;
-  case kUint32BinsTypeSize:
-    BuildHistKernel<do_prefetch, uint32_t, first_page, any_missing>
-                   (gpair, row_indices, gmat, hist, read_by_column);
-    break;
-  default:
-    CHECK(false);  // no default behavior
-  }
-}
-
 template <bool do_prefetch, bool any_missing>
 void BuildHistDispatch(const std::vector<GradientPair> &gpair,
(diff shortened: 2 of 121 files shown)


liuwenyi2:
Hi,
Your work is very interesting. I‘d like to test your optimization effect. Could you please tell me the source version of XGBoost?

razdoburdin:
Hi,
for performance measurements I have used the master branch in the dmlc/xgboost repo.
You can easily reproduce the calculations with it.
If you need any help, you are welcome!

liuwenyi2:
> Hi, for performance measurements I have used the master branch in the dmlc/xgboost repo. You can easily reproduce the calculations with it. If you need any help, you are welcome! From: wenyi Liu ***@***.***> Sent: Saturday, September 10, 2022 11:34 AM To: dmlc/xgboost ***@***.***> Cc: Razdoburdin, Dmitry ***@***.***>; Author ***@***.***> Subject: Re: [dmlc/xgboost] Optimization/buildhist/colwisebuildhist (PR #8233) Hi, Your work is very interesting. I‘d like to test your optimization effect. Could you please tell me the source version of XGBoost?

much appreciated!
Since I can only compile XGboost-1.1.0 on clusterIs.
I want to reproduce your calculations with 1.1.0 to do some measurements.
I'm trying to do it.

src/common/hist_util.cc
```diff
@@ -204,75 +221,161 @@ void BuildHistKernel(const std::vector<GradientPair> &gpair,
...
+    ColsWiseBuildHistKernel<BinIdxType, first_page, any_missing>
+                           (gpair, row_indices, gmat, hist, hbm);
+  } else {
+    RowsWiseBuildHistKernel<do_prefetch, BinIdxType, first_page, any_missing>
+                           (gpair, row_indices, gmat, hist);
+  }
+}
+
+template <bool do_prefetch, bool first_page, bool any_missing, class GHistBuildingManager>
+void BuildHistDispatch(const std::vector<GradientPair> &gpair,
```
trivialfis, line 286: Can we use the `DispatchBinType` instead?

src/common/column_matrix.cc
```diff
@@ -23,10 +23,12 @@ void ColumnMatrix::InitStorage(GHistIndexMatrix const& gmat, double sparse_thres
   gmat.GetFeatureCounts(feature_counts.data());
 
   // classify features
+  any_sparse_column_ = false;
```
trivialfis, line 26: `any_sparse_column_ = !all_dense_column`

src/tree/hist/histogram.h
```diff
@@ -16,6 +19,19 @@
 
 namespace xgboost {
 namespace tree {
+
+struct ColSample {
```
trivialfis, line 23: I think using `TrainParam` directly seems to be simpiler.

src/tree/hist/histogram.h
```diff
@@ -26,9 +42,15 @@ class HistogramBuilder {
...
   BatchParam param_;
   int32_t n_threads_{-1};
+
+  ColSample colsample_;
+  std::shared_ptr<common::ColumnSampler> column_sampler_;
+
   size_t n_batches_{0};
   // Whether XGBoost is running in distributed environment.
   bool is_distributed_{false};
+  // Addhoch colsample threshold level
```
trivialfis, line 52: `adhoc`.

src/tree/hist/histogram.h
```diff
@@ -59,10 +84,17 @@ class HistogramBuilder {
                             GHistIndexMatrix const &gidx,
                             std::vector<ExpandEntry> const &nodes_for_explicit_hist_build,
                             common::RowSetCollection const &row_set_collection,
-                            const std::vector<GradientPair> &gpair_h) {
+                            const std::vector<GradientPair> &gpair_h, int depth) {
     const size_t n_nodes = nodes_for_explicit_hist_build.size();
     CHECK_GT(n_nodes, 0);
 
+    const common::ColumnMatrix& column_matrix = gidx.Transpose();
+    const bool column_sampling =
```
trivialfis, line 92: Could you please explain the logic here? there are missing values but no sparse column in column matrix and bytree/bylevel lesser than threshold while there's no bynode? Is this necessary?

Also, this special case seems to be difficult to test.

trivialfis (commented):
Thank you for the work on optimization. Some questions in the comments. Can we omit the optimization on column sampling for now and just build the histogram for all columns?

razdoburdin:
> 
Hi, I have removed column sampling optimization for a while. Hope it will help with a review. Thanks for the suggestion!

src/common/hist_util.cc
razdoburdin: done

trivialfis (commented):
Could you please try to enable some tests for each of the case? I don't want to break anything in here by accident. ;-)

razdoburdin:
> Could you please try to enable some tests for each of the case? I don't want to break anything in here by accident. ;-)

As far as column wise building is just another way to calculate the same result as the row wise building,  I suggest to reuse the existing tests for histogram building.
I added force_read_by_column flag to BuildHist. It allows to use disable automatic choice of building strategy and to use column wise building instead. I duplicated the related tests with the new flag, so they now test both strategies.

trivialfis (approved):

PR #8234, opened 2022-09-08
Prepare for improving Windows networking compatibility.

trivialfis:
* Include dmlc filesystem indirectly as dmlc/filesystem.h includes windows.h, which conflicts with winsock2.h
* Define `NOMINMAX` conditionally.
* Link the winsock library when mysys32 is used.
* Add config file for read the doc.

Extracted from https://github.com/dmlc/xgboost/pull/8225 .

### DIFF
--- a/tests/cpp/filesystem.h
+++ b/tests/cpp/filesystem.h
@@ -1,0 +1,9 @@
+/*!
+ * Copyright (c) 2022 by XGBoost Contributors
+ */
+#pragma once
+
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif  // WIN32_LEAN_AND_MEAN
+#include "dmlc/filesystem.h"
--- a/tests/cpp/filesystem.h
+++ b/tests/cpp/filesystem.h
@@ -1,9 +1,15 @@
 /*!
  * Copyright (c) 2022 by XGBoost Contributors
  */
-#pragma once
 
+#ifndef XGBOOST_TESTS_CPP_FILESYSTEM_H
+#define XGBOOST_TESTS_CPP_FILESYSTEM_H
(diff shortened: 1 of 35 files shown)


tests/cpp/filesystem.h
```diff
@@ -0,0 +1,9 @@
+/*!
+ * Copyright (c) 2022 by XGBoost Contributors
+ */
+#pragma once
```
hcho3, line 4: Please use include guard macros instead. CPPLint requires it.

tests/cpp/filesystem.h
```diff
@@ -0,0 +1,9 @@
...
+
+#ifndef WIN32_LEAN_AND_MEAN
```
hcho3, line 6: Where is this macro used?

hcho3 (changes requested):
Let me submit a pull request to dmlc-core to hide all implementations details from the `dmlc/filesystem.h` header. For now, this PR looks fine.

tests/cpp/filesystem.h
trivialfis: Inside `windows.h`.

trivialfis:
> Let me submit a pull request to dmlc-core to hide all implementations details from the dmlc/filesystem.h header. For now, this PR looks fine.

Thank you! I'm not sure about the effect of that as I don't quite understand the internal of windows.h, so I tried to make this change local to XGBoost since it's only used in unittest.

trivialfis:
The error is a problem because we are using unity build on windows CI, so all files are exposed.

tests/cpp/filesystem.h
trivialfis: Changed to using macro guard.

hcho3 (approved):

PR #8239, opened 2022-09-09
Bump r cache version.

trivialfis:


trivialfis:
We should wait for the binary package for windows to be available.

PR #8242, opened 2022-09-10
JNI wrapper for the collective communicator

rongou:
We're probably not going to support federated learning in Spark, but since we'll switch the c++ code to use the communicator interface, this is also needed to keep the java/scala code functioning.


rongou:
@trivialfis

trivialfis:
Started buildkite.

trivialfis (commented):
Looks good to me overall.  Please document its experimental status.

rongou:
Added a note about the api being experimental.

trivialfis:
Restarted failed tests.

trivialfis (approved):

wbo4958:
Looks like Communicator.java is quite the same as Rabit.java. So does Communicator is dedicated to federated learning? Can it be replaced with the existing Rabit? Looks like we need to re-design this part to abstract the common parts. But that can be in a following-up PR.

rongou:
@wbo4958 yes it's more a less a drop-in replacement for Rabit. My next step is to replace all the rabit calls with communicator, and then delete Rabit.java. These changes don't affect the JVM side in terms of behavior, but are needed since we are changing the c++ code.

rongou:
@trivialfis can this be merged? Thanks!

PR #8246, opened 2022-09-14
Use quantised gradients in gpu_hist histograms

RAMitchell:
Floating point addition and subtraction is causing small numerical errors in the learning process, sometimes causing the algorithm to create splits on 'phantom' gradients. We can solve this by consistently working with gradients in a high precision quantised format, only converting back to floating point for the final gain or weight calculations.

This PR is the first of 2 PRs, converting histogram related calculations to use quantised gradients.

A second PR will address the split evaluation part.


trivialfis:
Thank you for working on resolving this issue. What's the current status of the floating point error?

RAMitchell:
This does not fix anything yet, I need to do this second part which affects the calculations in split evaluation.

This spark test failure looks unrelated to this PR.

include/xgboost/base.h
```diff
@@ -259,10 +259,61 @@ class GradientPairInternal {
...
 /*! \brief High precision gradient statistics pair */
 using GradientPairPrecise = detail::GradientPairInternal<double>;
-/*! \brief Fixed point representation for gradient pair. */
-using GradientPairInt32 = detail::GradientPairInternal<int>;
-/*! \brief Fixed point representation for high precision gradient pair. */
-using GradientPairInt64 = detail::GradientPairInternal<int64_t>;
+
+/*! \brief Fixed point representation for high precision gradient pair. Has a different interface so
+ * we don't accidentally use it in gain calculations.*/
+class GradientPairInt64 {
```
trivialfis, line 265: I think this duplicated can be removed with some type checking in the `XGBOOST_DEVICE explicit GradientPairInternal(const GradientPairInternal<T2> &g)` constructor and similar functions.

In c++ 17:
``` c++
template <typename... T>
struct is_grad_convertible {
  using type = std::conditional_t<(std::is_floating_point<T>::value && ...) ||
                                      (std::is_integral<T>::value && ...),
                                  std::true_type, std::false_type>;
};

// in function templates:
template <typename T2, std::enable_if_t<is_grad_convertible<T, T2>::value>* = nullptr>
GradientPairInternal(const GradientPairInternal<T2> &g);
```

tests/cpp/tree/gpu_hist/test_histogram.cu
```diff
@@ -95,11 +94,25 @@ TEST(Histogram, GPUDeterministic) {
   std::vector<int> shm_sizes{48 * 1024, 64 * 1024, 160 * 1024};
   for (bool is_dense : is_dense_array) {
     for (int shm_size : shm_sizes) {
-      TestDeterministicHistogram<GradientPairPrecise>(is_dense, shm_size);
+      TestDeterministicHistogram(is_dense, shm_size);
     }
   }
 }
 
+void ValidateCategoricalHistogram(size_t n_categories, common::Span<GradientPairInt64> onehot,
```
trivialfis, line 102: Can we merge the CPU and GPU versions back if we can deduplicate the gradient definition or use a template for this function?

include/xgboost/base.h
RAMitchell: It's not the copying that's the problem. When I began to change the gradient types the following statements still compile with integer gradients and they really shouldn't:
`if(grad.GetHess() >= min_child_weight)`
`evaluator.CalcGain(...., grad,...)`
So by changing the interface name these statements for longer compile and it is not possible to accidentally use the quantised value in calculations.

It's a bit annoying to duplicate some code but I think it's worth it. I kept finding bugs from the above situation until I made this change.

tests/cpp/tree/gpu_hist/test_histogram.cu
RAMitchell: As mentioned above, it's intentional to force the integer gradients to be handled differently.

trivialfis:
I restarted the CI.

trivialfis:
The pyspark error seems to be related.

RAMitchell:
I think I need to reduce over the distributed workers in the histogram rounding bound calculation.

trivialfis:
> I think I need to reduce over the distributed workers in the histogram rounding bound calculation.

hmm  .. why?

RAMitchell:
Allreduce is now happening on quantised gradients. We would get the wrong results if each worker provided integer gradients quantised in a different scale.

RAMitchell:
@trivialfis can I please get another review?

src/tree/gpu_hist/histogram.cu
```diff
@@ -107,35 +112,55 @@ HistRounding<GradientSumT> CreateRoundingFactor(common::Span<GradientPair const>
...
+  auto h = gpair.GetQuantisedHess();
+
+  AtomicAdd64As32(dst_ptr, g);
+  AtomicAdd64As32(dst_ptr + 1, h);
 }
 
-template HistRounding<GradientPairPrecise> CreateRoundingFactor(
-    common::Span<GradientPair const> gpair);
-template HistRounding<GradientPair> CreateRoundingFactor(common::Span<GradientPair const> gpair);
+// Global 64 bit integer atomics at the time of writing do not benefit from being separated into two
```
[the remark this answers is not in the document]
RAMitchell: Yes much slower.

src/tree/gpu_hist/histogram.cu
```diff
@@ -66,36 +66,41 @@ struct Clip : public thrust::unary_function<GradientPair, Pair> {
     auto ph = Pclip(x.GetHess());
 
     auto ng = Nclip(x.GetGrad());
-    auto nh = Nclip(x.GetHess());
+    auto nh = Pclip(x.GetHess());
```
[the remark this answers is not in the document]
RAMitchell: Ah I see. I thought it was strange to clip hessian negative as it can't be negative. Changing it back.

trivialfis (approved):

Issue #8249, opened 2022-09-17
Feature request: Send null features down both branches


aeftimia:
My understanding is that XGBoost currently handles null values as separate datatypes, and will split on whether the data is null or not before moving on to what the value is if not null. This is very reasonable if the existence of a null value has predictive power over the target variable or even if it is uncorrelated. But consider a demand forecasting problem in which a feature went missing around peak season. An XGBoost model built on this data could learn that missing values of this feature indicate high demand. If the feature being null were a boolean variable, we would of course discard it to prevent the model from considering this conclusion, but XGBoost's approach effectively forces the use of this feature. Since XGBoost is built on decision trees, it is in a unique position to simply ignore null values with a [modified tree learning algorithm](http://mlwiki.org/index.php/Decision_Tree_(Data_Mining)#Handling_Missing_Values). The idea is null features should be treated as weighted leaves in both splits and weighed according to the proportion of leaves with non-null data in each branch.

Of course, this only works if non-null values of the feature generally provide intended predictive power. Going back to the forecasting example, if a feature went live during peak season, then went back offline, null values of the feature could still become an indicator of high demand. That said, dropping the feature entirely in this scenario is probably more sensible. Overall, I'd rather have the option of dropping a mostly useless feature than let a rare instance of its absence potentially spoil its predictive power entirely.

trivialfis:
XGBoost doesn't treat the null value as a separate type (although I'm not entirely sure what "separate datayptes" means). Whether the missing value should go to the left branch or the right branch of a split depends on which direction gives more gain. This is decided per-node.

aeftimia:
Oh my understanding was xgboost has a separate split for whether the data is null (e.g. is of a separate "null type", and if not, proceed with dealing with it as a normal float). Based on your description, it sounds branches are determined without using null values, then null values are assigned to the branch that minimizes loss. I think this would still have the same pitfall in the scenario I described. So the request remains to have the option of sending null values down both branches in a smart way to avoid overfitting on the existence of null values.

trivialfis:
It's interesting to think about how to modify the algorithm to combat overfitting. Thank you for the reference, I will look into the wiki deeper and see the relationship between the proposed modification and XGBoost tree split. At the same time, please feel free to look into XGBoost's implementation: https://github.com/dmlc/xgboost/blob/70df36c99cf7112319a02c25f7aac3f39a718f33/src/tree/hist/evaluate_splits.h#L211

aeftimia:
[...]

PR #8256, opened 2022-09-21
Send null values down both branches

aeftimia:
# WIP

Potential means of handling missing data that avoids overfitting. See feature-request #8249 and [this link](http://mlwiki.org/index.php/Decision_Tree_(Data_Mining)#Handling_Missing_Values) for the basic concept. 

I added a new boolean into the training parameters (better names are, of course, welcome) and took a crack at refactoring evaluate_splits.h to incorporate the algorithm when searching for splits with continuous features. The idea is instead of simply doing a backward sweep when there is a discrepancy between `left_sum` and the total gradient and hessian, do a backward sweep while adding `1/(iend-ibegin)` times that discrepancy to the `left_sum` for each iteration in the sweep. This way, the residual gradients and Hessians from null values are added to `left_sum` incrementally throughout the sweep.

If how the code was refactored seems acceptable, I can repeat the pattern on methods that handle categorical and one hot encoded feature, then move on to inference and testing. 

@trivialfis

### DIFF
--- a/src/tree/hist/evaluate_splits.h
+++ b/src/tree/hist/evaluate_splits.h
@@ -207,13 +207,10 @@ class HistEvaluator {
   // Enumerate/Scan the split values of specific feature
   // Returns the sum of gradients corresponding to the data points that contains
   // a non-missing value for the particular feature fid.
-  template <int d_step>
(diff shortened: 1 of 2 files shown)


trivialfis:
Thank you for working on this. I will look deeper into the PR

PR #8257, opened 2022-09-21
[Breaking] Switch from rabit to the collective communicator

rongou:
This PR switches the `Rabit` api to `Communicator`, which gives us more flexibility in the collective communication implementation. It's a breaking change, but to most uses, it's a straightforward swap.


trivialfis (commented):
Thank you for the work on swapping out rabit. When it's ready, could you please split up the PR into smaller ones and start with internal C++ changes?

rongou:
@trivialfis the issue is this is a breaking change. Once we change the c++ portion, we have to change the python and java apis too, otherwise the communicator would be uninitialized.

rongou:
@wbo4958 any ideas about the JVM test failures? They seem to pass on my local desktop.

wbo4958:
I will check it today.

wbo4958:
I can repro it locally, which will make life easy.

rongou:
@wbo4958 Thanks for the help with the debugging. Resetting the mocklist seems to have fixed it.

@trivialfis I think this PR is ready for review. It touches a lot of files, but mostly it's a one-to-one swap from rabit to communicator. Thanks!

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -22,7 +22,7 @@ import scala.collection.mutable
 import scala.util.Random
 import scala.collection.JavaConverters._
 
-import ml.dmlc.xgboost4j.java.{IRabitTracker, Rabit, XGBoostError, RabitTracker => PyRabitTracker}
+import ml.dmlc.xgboost4j.java.{Communicator, IRabitTracker, XGBoostError, RabitTracker => PyRabitTracker}
```
trivialfis, line 25: Where is the `IRabitTracker` used?

src/collective/communicator.cu
```diff
@@ -16,7 +17,7 @@ thread_local std::unique_ptr<DeviceCommunicator> Communicator::device_communicat
 
 void Communicator::Finalize() {
   communicator_->Shutdown();
-  communicator_.reset(nullptr);
```
trivialfis, line 19: Could you please share under which case the communicator can still be called after being shut down? I think a nullptr can be a guard for unintentional calls.

src/collective/noop_communicator.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "communicator.h"
+
+namespace xgboost {
+namespace collective {
+
+/**
+ * A no-op communicator, used for non-distributed training.
+ */
+class NoOpCommunicator : public Communicator {
```
trivialfis, line 15: I see that you have already added checks for non-distributed env in various communicator implementations. Is this still necessary?

src/metric/auc.cu
```diff
@@ -46,7 +45,7 @@ struct DeviceAUCCache {
   dh::device_vector<size_t> unique_idx;
   // p^T: transposed prediction matrix, used by MultiClassAUC
   dh::device_vector<float> predts_t;
-  std::unique_ptr<dh::AllReducer> reducer;
+  collective::DeviceCommunicator* communicator;
```
trivialfis, line 48: If this is now a global instance, we don't have to maintain a pointer to it.

python-package/xgboost/dask.py
```diff
@@ -158,7 +158,7 @@ def _try_start_tracker(
         if isinstance(addrs[0], tuple):
             host_ip = addrs[0][0]
             port = addrs[0][1]
-            rabit_context = RabitTracker(
+            rabit_tracker = RabitTracker(
```
trivialfis, line 161: hmm .. so we still need rabit tracker for downstream projects. Could you please share how federated learning communicates the worker addresses across all workers?

src/collective/communicator.cc
```diff
@@ -12,14 +13,10 @@
 namespace xgboost {
 namespace collective {
 
-thread_local std::unique_ptr<Communicator> Communicator::communicator_{};
+thread_local std::unique_ptr<Communicator> Communicator::communicator_{new NoOpCommunicator()};
```
trivialfis, line 16: See other comments on the no-op, I have some concerns that we will use this accidentally, we have issues where XGBoost failed to establish a working communicator group but proceed with distributed training without explicit error. I haven't been able to tackle those issues due to the complicated network setup others use. But still, it's something that we should have in mind.

trivialfis (commented):
Thank you for the great work on swapping out rabit. Could you please layout the plan for future PRs for 1.7 release? I'm trying to estimate an ETA.

python-package/xgboost/dask.py
rongou: Yes if we use the rabit communicator, which is the default, we still need to start a tracker. I imagine if we switch to something like gloo, then we can get rid of rabit completely.

For federated learning, since we have to start a gRPC server first, we just pass the server address (host:port) to each client.

src/collective/communicator.cc
rongou: On this line https://github.com/dmlc/xgboost/blob/master/rabit/src/engine.cc#L73, rabit provides a default engine if it's not initialized. We have some code that depends on this behavior. The `NoOpCommunicator` is one way to replicate this behavior.

I think the user has to tell us if they are doing distributed training, either by entering the `CommunicatorContext`, or calling `Init` directly. Otherwise we'd have no way of knowing, and can't just return a `nullptr` in case they are not in the distributed mode.

src/collective/communicator.cu
rongou: I think we have some tests that mix distributed training and local training.

src/collective/noop_communicator.h
rongou: As mentioned above, this is needed to replicate the existing rabit behavior.

src/metric/auc.cu
rongou: We still need the device id to get the communicator, which in this class is only passed in during `Init` and not saved. We can probably clean this up once we have better device id management.

rongou (commented):
After this PR, I think the only code change needed is #8279, and then we just need to weak the CI to build with federated learning enabled.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
rongou: A couple of private functions https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala#L351.

trivialfis: Thank you for sharing.

trivialfis (approved):

src/metric/auc.cu
rongou: Ended up removing these pointers.

trivialfis:
Could you please merge master again?

rongou:
It's up to date.

rongou:
@trivialfis can this be merged? Thanks!

rongou:
@trivialfis @hcho3 can this be merged now? Thanks!

hcho3:
@gnaggnoyil You can choose between Rabit and Federated by passing `{"xgboost_communicator": "[tracker type]"}` to `xgboost.collective.init()`.

As for `xgboost.rabit`, we got rid of the submodule in 1.7.0, but it broke some downstream projects. So we plan to release patch release 1.7.1 to restore it.

Issue #8251, opened 2022-09-21
Misguided Implementation of Saabas' method in `Predict(approx_contribs=True)`


nalzok:
Despite citing the original blog post [Interpreting random forests](https://blog.datadive.net/interpreting-random-forests/), the current implementation of Saabas' method in `RegTree::CalculateContributionsApprox` seems misguided.

https://github.com/dmlc/xgboost/blob/7d43e74e71c1a0f6526fda78d46395c883d14cb0/src/tree/tree_model.cc#L1214-L1218

In particular, the `mean_values` parameter is filled by the function `FillNodeMeanValues`, which as its name suggests, simply calculates the average response within a particular node.

https://github.com/dmlc/xgboost/blob/7d43e74e71c1a0f6526fda78d46395c883d14cb0/src/predictor/cpu_predictor.cc#L260-L275

However, I argue that we should rename the parameter to `weight_values` and calculate the prediction/weight at each internal node as if it was a leaf. To quote the blog post by Saabas,

> What's novel here is that you can see the breakdown of the prediction, written down in terms of value changes along the prediction path, together with feature names that “caused” every value change due to being in the guard (the numbers are approximate due to rounding).

The point of Saabas' method is to see how the prediction changes as we go deeper into the tree. For random forests, the prediction/weight coincides with the average value within that node, i.e. $-\frac{G_j}{H_j}$, which is what the current implementation calculates. However, for XGBoost, the value should be $-\frac{\operatorname{ThresholdL1}(G_j-\alpha)}{H_j+\lambda}$, where $\alpha$ and $\lambda$ are L1 and L2 regularization parameters respectively.

https://github.com/dmlc/xgboost/blob/7d43e74e71c1a0f6526fda78d46395c883d14cb0/src/tree/param.h#L246-L258

To sum up, I believe the `RegTree::CalculateContributionsApprox` method should use the weight/prediction value in each node to calculate, instead of `mean_values`. I'm more than happy to open a PR to address that, but please let me know if the weights are stored somewhere so that we can avoid some duplicated computation.

trivialfis:
@RAMitchell

nalzok:
I would like to point out that `RegTree::CalculateContributions` has a similar problem

https://github.com/dmlc/xgboost/blob/7d43e74e71c1a0f6526fda78d46395c883d14cb0/src/tree/tree_model.cc#L1417-L1421

[...]

PR #2003, opened 2017-02-03
Add prediction of feature contributions

mcuelenaere:
This implements the idea described at http://blog.datadive.net/interpreting-random-forests/
which tries to give insight in how a prediction is composed of its feature contributions
and a bias.

The calculation of the node values is based on the findings of @SauceCat at https://github.com/dmlc/xgboost/issues/1986.

Open issues:
 1. I added this functionality to `XGBoosterPredict()` as another value to `option_mask`. However it is mutually exclusive with the `PredictLeaf()` and `PredictRaw()` options, so it feels wrong to being able to pass `0x07` to this function (which is no different from `0x04`)
 1. to calculate the values at each node, I multiply the learning rate with the `base_weight` property. However, the `learning_rate` value seems to be unavailable in `GBTree` (or `RegTree` for that matter). So the code calculates it by dividing the leaf value with the base weight, which is pretty hacky.
 1. only the Python bindings have been updated to expose this new functionality

PR #8264, opened 2022-09-23
Fix mean_values for Saabas method

nalzok:
Address https://github.com/dmlc/xgboost/issues/8251

### DIFF
--- a/src/predictor/cpu_predictor.cc
+++ b/src/predictor/cpu_predictor.cc
@@ -274,13 +274,23 @@ float FillNodeMeanValues(RegTree const *tree, bst_node_t nidx, std::vector<float
   return result;
 }
 
-void FillNodeMeanValues(RegTree const* tree, std::vector<float>* mean_values) {
+void FillNodeMeanValues(RegTree const* tree, std::vector<float>* mean_values, bool approximate) {
   size_t num_nodes = tree->param.num_nodes;
   if (mean_values->size() == num_nodes) {
     return;
   }
   mean_values->resize(num_nodes);
-  FillNodeMeanValues(tree, 0, mean_values);
+
+  if (approximate) {
+    // prediction value as if the node is a leaf
+    auto &node_mean_values = *mean_values;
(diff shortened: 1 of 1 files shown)


nalzok:
Looks like I have passed all tests except the following one

https://github.com/dmlc/xgboost/blob/8d247f0d64bee392857b330b20bfe6086798eb7b/tests/cpp/predictor/test_cpu_predictor.cc#L74-L86

I'm not sure what's going on here. Is that because I assumed `base_weight` has been populated before calling `PredictContribution`, but it's not in this case?

src/predictor/cpu_predictor.cc
```diff
@@ -274,13 +274,23 @@ float FillNodeMeanValues(RegTree const *tree, bst_node_t nidx, std::vector<float
...
     return;
   }
   mean_values->resize(num_nodes);
-  FillNodeMeanValues(tree, 0, mean_values);
+
+  if (approximate) {
+    // prediction value as if the node is a leaf
+    auto &node_mean_values = *mean_values;
+    for (bst_node_t nidx = 0; nidx < num_nodes; ++nidx) {
+      node_mean_values[nidx] = tree->Stat(nidx).base_weight;
```
RAMitchell, line 288: I think `tree->Stat(nidx).base_weight` is not adjusted for learning rate. So this prediction would be much too large.

nalzok: That's true, but unfortunately, we cannot access the training parameters (i.e. `learning_rate`) here. Do you have a better idea than the following hack?

https://github.com/dmlc/xgboost/pull/2003/commits/87d96d2c13663122346216c8eba0829dfdcbf175#diff-0c9f721872945f814f6d5e9ade9cd184b7e1882ce6a138f183f5af07c8520c65R596-R600

trivialfis:
@RAMitchell Do you think there's a better approach?

RAMitchell:
I think the current implementation is correct as we don't claim to implement the Saabas method.

PR #8269, opened 2022-09-25
Obtain CSR matrix from DMatrix.

trivialfis:
* Obtain CSR matrix from DMatrix.
* Obtain gradient index from Quantile DMatrix.

This is mostly for testing at a higher level. Right now we rely on training a booster to infer that the DMatrix is correctly constructed. With this PR, we can ease the testing process. Also, this has been [requested before](https://github.com/dmlc/xgboost/pull/4759).

The return value from Quantile DMatrix is histogram index instead of cut values as we shift the cut value to include `min_values`, which is not very useful for language bindings.

Unlike most of XGBoost C functions, the caller of C API is required to allocate the memory itself instead of using thread local memory from XGBoost. This is to avoid allocating a huge memory buffer that can not be freed until exiting the thread.

External memory is not supported.

Close https://github.com/dmlc/xgboost/pull/4759 .


include/xgboost/c_api.h
```diff
@@ -761,6 +761,39 @@ XGB_DLL int XGDMatrixNumRow(DMatrixHandle handle,
...
+ *
+ * \param handle the handle to the DMatrix
+ * \param out The output of number of non-missing values
+ *
+ * \return 0 when success, -1 when failure happens
+ */
+XGB_DLL int XGDMatrixNumNonMissing(DMatrixHandle handle, bst_ulong *out);
+
+/*!
+ * \brief Get the predictors from DMatrix as CSR matrix.  If this is a quantized DMatrix,
```
RAMitchell, line 777: Could it be useful to return quantised float values instead of integers?

We could return values such that `bst.predict(quantile_dmat) == bst.predict(xgb.DMatrix(quantile_dmat.get_data()))`

trivialfis: Yes, we can. That's a good argument. I skipped restoring the values because it contains artificially created min and max values, which might be confusing.

trivialfis: Changed the return value to be cut values.

trivialfis:
@RAMitchell Please take another look when you are available.

RAMitchell (approved):

PR #8272, opened 2022-09-26
Init estimation for regression.

trivialfis:


trivialfis:
Some small refactors are coupled in this PR to cleanly split up CPU impl and CUDA impl for utilities.

src/objective/regression_loss.h
```diff
@@ -39,6 +52,15 @@ struct LinearSquareLoss {
 
   static const char* Name() { return "reg:squarederror"; }
   static ObjInfo Info() { return {ObjInfo::kRegression, true, false}; }
+
+  static bool InitEstimation(Context const* ctx, MetaInfo const& info,
+                             linalg::Tensor<float, 1>* base_margin) {
+    CheckInitInputs(info);
+    base_margin->Reshape(1);
+    auto out = base_margin->HostView();
+    out(0) = WeightedMean(ctx, info);
```
RAMitchell, line 61: If you wanted to make a more general implementation that would work for the other loss functions, maybe you could just calculate the gradient/hessian, sum them up and calculate base margin as w=G/H.

trivialfis: We don't have prediction at this stage, as a result there's no gradient. This initialization is to establish the first prediction.

trivialfis: @RAMitchell I believe the scale pos(itive) weight is exclusively for logistic, and I agree that we can remove it.

RAMitchell: You could just predict 0 and get the gradients.

trivialfis:
I will change the init code to all objectives in followups.

trivialfis:
Will extract a small portion of this PR if I can't get everything working in time.

RAMitchell (approved):

PR #8274, opened 2022-09-26
Use integer gradients in gpu_hist split evaluation

RAMitchell:
Part 2 of #8246


tests/cpp/tree/test_gpu_hist.cu
```diff
@@ -162,81 +162,6 @@ inline GradientQuantizer DummyRoundingFactor() {
   return GradientQuantizer(dh::ToSpan(gpair));
 }
 
-// TODO(trivialfis): This test is over simplified.
-TEST(GpuHist, EvaluateRootSplit) {
```
RAMitchell, line 166: I removed this test because it specifies a dense dataset but then also has missing gradients, which now triggers a run time check. It was easier to remove it than rewrite it.

src/tree/gpu_hist/evaluate_splits.cu
```diff
@@ -59,67 +62,67 @@ class EvaluateSplitAgent {
...
         rounding(shared_inputs.rounding),
-        parent_sum(dh::LDGIterator<GradientPairPrecise>(&inputs.parent_sum)[0]),
-        param(shared_inputs.param),
-        evaluator(evaluator),
+        parent_sum(dh::LDGIterator<GradientPairInt64>(&inputs.parent_sum)[0]),
+        param(shared_inputs.param), evaluator(evaluator),
         missing(parent_sum - ReduceFeature()) {
-    static_assert(kBlockSize == 32,
-                  "This kernel relies on the assumption block_size == warp_size");
+    static_assert(
```
trivialfis, line 88: that's awesome!

trivialfis (approved):
Please fix the clang tidy error.

Issue #8083, opened 2022-07-15
Apply dmatrix iteration iterface in PySpark xgboost and support external memory mode


WeichenXu123:
Apply dmatrix iteration iterface in PySpark xgboost and support external memory mode.

WeichenXu123:
@trivialfis When will you create first PR "Apply dmatrix iteration iterface in PySpark xgboost " ? If you are busy, I can help to create PR. Thank you.

trivialfis:
@WeichenXu123 Thank you for the work. Feel free to continue the work on external memory. I just didn't want to have too many conflicts between the implementation of ext memory and the quantile dmatrix. (See the WIP iterator in https://github.com/dmlc/xgboost/pull/8088 ).

WeichenXu123:
@trivialfis To reduce conflicts, I would like to wait https://github.com/dmlc/xgboost/pull/8088 merged first. :)

trivialfis:
I need to apply the quantile dmatrix to spark as well.

trivialfis:
I agree. I kept the feature here mostly for experimental support.

trivialfis:
But quantile DMatrix can still be used for reducing memory usage without sacrificing performance.

PR #8284, opened 2022-09-28
[pyspark] Use quantile dmatrix.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/8083 .


trivialfis:
@WeichenXu123 @wbo4958 Please take a look when you are available.

python-package/xgboost/spark/core.py
```diff
@@ -758,17 +759,19 @@ def _train_booster(pandas_df_iter):
...
 
             gpu_id = None
-            if use_gpu:
-                gpu_id = context.partitionId() if is_local else _get_gpu_id(context)
-                booster_params["gpu_id"] = gpu_id
 
+            if use_qdm:
                 # max_bin is needed for qdm
                 if (
                     features_cols_names is not None
```
wbo4958, line 766: So the QuantileDMatrix is only used for `features_cols_names is not None` ?

python-package/xgboost/spark/data.py
```diff
@@ -240,34 +258,65 @@ def make(values: Dict[str, List[np.ndarray]], kwargs: Dict[str, Any]) -> DMatrix
...
     # worker do the AllReduce when constructing DMatrix, or else it may hang
     # forever.
-    dvalid = make(valid_data, kwargs) if has_validation_col else None
-
-    if dvalid is not None:
-        assert dvalid.num_col() == dtrain.num_col()
+    if has_validation_col:
+        if use_qdm:
+            it = PartIter(valid_data, gpu_id, **meta)
+            dvalid: Optional[DMatrix] = QuantileDMatrix(it, **params, ref=dtrain)
```
wbo4958, line 316: can this handle the empty data?

python-package/xgboost/spark/data.py
```diff
@@ -240,32 +258,66 @@ def make(values: Dict[str, List[np.ndarray]], kwargs: Dict[str, Any]) -> DMatrix
...
+        for k, v in kwargs.items():
+            if k in non_data_keys:
+                non_data_params[k] = v
+            else:
+                meta[k] = v
+        return meta, non_data_params
+
+    meta, params = split_params()
+
+    if feature_cols is not None:  # rapidsai plugin
```
wbo4958, line 288: basically, feature_cols is working for w and w/o rapids plugin

python-package/xgboost/spark/data.py
```diff
@@ -240,32 +258,66 @@ def make(values: Dict[str, List[np.ndarray]], kwargs: Dict[str, Any]) -> DMatrix
...
+        cache_partitions(iterator, append_dqm)
+        assert gpu_id is not None
```
wbo4958, line 290: Could we put the check into very begging for quick throwing?

python-package/xgboost/spark/data.py
trivialfis: For non-rapids setup, feature cols is not supported.

python-package/xgboost/spark/data.py
trivialfis: will move it.

python-package/xgboost/spark/data.py
trivialfis: Added tests.

trivialfis:
@wbo4958 Could you please take another look?

python-package/xgboost/spark/data.py
wbo4958: I don't think so, without the rapids plugin, the combination of (features_cols + gpu_hist) still can work.

wbo4958:
LGTM

RAMitchell (approved):

trivialfis:
Apologies for the new changes. For some reason, the pytest mark doesn't work with test cases derived from the python unittest module.

python-package/xgboost/spark/data.py
```diff
@@ -149,24 +153,52 @@ def _read_csr_matrix_from_unwrapped_spark_vec(part: pd.DataFrame) -> csr_matrix:
...
-    """Create DMatrix from spark data partitions. This is not particularly efficient as
-    we need to convert the pandas series format to numpy then concatenate all the data.
+    """Create DMatrix from spark data partitions.
 
     Parameters
     ----------
     iterator :
         Pyspark partition iterator.
+    feature_cols:
+        A sequence of feqture names, used only when rapids plugin is enabled.
```
wbo4958, line 187: feqture -> feature. this parameter can be used even without rapid plugin.

python-package/xgboost/spark/data.py
```diff
@@ -228,6 +260,10 @@ def append_dqm(part: pd.DataFrame, name: str, is_valid: bool) -> None:
 
     def make(values: Dict[str, List[np.ndarray]], kwargs: Dict[str, Any]) -> DMatrix:
         if len(values) == 0:
+            get_logger("XGBoostPySpark").warning(
```
wbo4958, line 263: previously, the warning is only printed for empty training data. while this PR also prints it for validation data.

wbo4958 (commented):
Overall, LGTM. only some minor nits.

python-package/xgboost/spark/data.py
trivialfis: I haven't really done any test for it and it will likely trigger an assert error. We have `DMatrix` and `QuantileDMatrix` to support, I will leave that to the next release.

python-package/xgboost/spark/data.py
trivialfis: It's still good to have a warning, the support is "best effort" and is not yet extensively tested. If something goes wrong at least users will have some clues to debug.

PR #8288, opened 2022-09-29
Reduce time for some multi-gpu tests

RAMitchell:


trivialfis (commented):
Thank you for working on this. A few notes for the change of fixture:
- The local cuda cluster fixture is enabled with rmm support https://github.com/dmlc/xgboost/blob/b14c44ee5e4f941c59de3d91633f4b0acdff5b23/tests/python-gpu/conftest.py#L30 .
- I think the CI has 4 GPUs.

hcho3 (approved):

PR #8301, opened 2022-10-04
[CI] Enable federated learning plugin in buildkite

rongou:


tests/buildkite/build-cuda.sh
```diff
@@ -22,7 +22,7 @@ command_wrapper="tests/ci_build/ci_build.sh gpu_build_centos7 docker --build-arg
 echo "--- Build libxgboost from the source"
 $command_wrapper tests/ci_build/prune_libnccl.sh
 $command_wrapper tests/ci_build/build_via_cmake.sh -DUSE_CUDA=ON -DUSE_NCCL=ON \
-  -DUSE_OPENMP=ON -DHIDE_CXX_SYMBOLS=ON -DUSE_NCCL_LIB_PATH=ON \
+  -DUSE_OPENMP=ON -DPLUGIN_FEDERATED=ON -DHIDE_CXX_SYMBOLS=ON -DUSE_NCCL_LIB_PATH=ON \
```
hcho3, line 25: Is gRPC a runtime dependency? If so, we shouldn't enable the federated learning plugin in the Python wheel on PyPI.

hcho3: We can built a separate artifact with gRPC for testing purposes

tests/ci_build/Dockerfile.rmm
```diff
@@ -11,6 +11,14 @@ RUN \
...
     apt-get update && \
     apt-get install -y wget unzip bzip2 libgomp1 build-essential ninja-build git && \
+    # Install gRPC for Federated Learning
+    git clone -b v1.49.1 https://github.com/grpc/grpc.git && \
+    pushd grpc && \
+    git submodule update --init && \
+    cmake -S . -B build -GNinja && \
+    cmake --build build --target install && \
+    popd && \
+    rm -fr grpc && \
```
hcho3, line 21: It could be better to use `grpc-cpp` package from Conda, as we won't have to build from the source

rongou: Tried this, but looks like the `libprotobuf` conda package doesn't install the cmake modules: https://github.com/conda-forge/libprotobuf-feedstock/issues/68

rongou:
As discussed offline, @hcho3 is taking this over. Feel free to close it when you are ready. Thanks!

PR #8310, opened 2022-10-06
Scala 2.13 and Spark 3.2.1

vincenzobaz:
This PR introduces Scala 2.13 and Spark 3.2.1 to the repository.

The biggest caveat is the impossibility to port the flink module this Flink does not have a 2.13 version.
A solution would be to use the Flink Java API however that requires more important refactoring since Flink ml does not exist for Java.


wbo4958:
Could we create another branch to support scala 2.13? or let the main branch support both scala 2.12 and scala 2.13, then we can pre-build xgboost-2.12 and xgboost-2.13?

When bumping xgboost dep to spark 3.x, XGBoost committers chose not to support spark 2.x, this is not good. So I hope the current XGBoost can support all spark 3.x versions. On the other hand, XGBoost contains not only xgboost project but also xgboost-gpu project, and xgboost-gpu requires [RAPIDS plugin](https://nvidia.github.io/spark-rapids/) which has not supported scala 2.12 yet.

trivialfis (commented):
@wbo4958 Maybe we can merge this after 1.7 is branched out?

wbo4958:
I really hope XGBoost can support all the spark 3.x versions for both scala 2.12 and scala 2.13. This is really a huge breaking PR that makes xgboost can run on spark 3.0 and spark 3.1 versions. and also xgboost4j-spark-gpu can't work.

trivialfis:
Sure, is there a viable approach?

vincenzobaz:
Thank you for the feedback. I am not sure why `xgboost4j-spark-gpu` does not work unfortunately.

I agree that this PR brings a lot of breaking changes. A possibility would be to keep more modern Spark support in master and move the legacy version to a dedicated branch. Publishing both _2.12 and _2.13 is probably the way to go.
I would suggest adopting sbt for the build of the jvm modules since it can handle correctly java and Scala builds and cross version development / publication is easy.

I upgraded Spark to the first version to support 2.13. However I think it could be useful to upgrade in `master` too so to reduce the divergence between the 2.13 and 2.12 versions of the codebase.

wbo4958:
Thx @vincenzobaz, I'd like to quote this comment https://github.com/NVIDIA/spark-rapids/discussions/6739#discussioncomment-3829057

dotbg:
@trivialfis this MR is redundant now as 2.13 was merged in #9099

vincenzobaz:
Thank you @dotbg , I appreciate the great work!

dotbg:
@vincenzobaz thanks. This PR helped me a lot!

Issue #8307, opened 2022-10-05
Flaky test on Device Quantile DMatrix.


trivialfis:
https://buildkite.com/xgboost/xgboost-ci/builds/330#0183a5d6-e959-4e3b-97ad-32da7bc9ae4a

```
You can reproduce this example by temporarily adding 
@reproduce_failure('6.56.0', b'AAAAAAABAAAAXwAChQ==') as a decorator on 
your test case
```

PR #8312, opened 2022-10-06
[CI] Refactor tests to reduce CI time.

RAMitchell:
Also xfail failing test from #8307.


RAMitchell:
CPU python tests 24m:42s->14m:26s.
GPU python tests 5m:27s->1m:54s.

RAMitchell:
Test XGBoost on Windows: 21min->12min

hcho3:
The failing MacOS tests are fixed in #8321

tests/python/testing.py
```diff
@@ -204,7 +204,6 @@ def next(self, input_data: Callable) -> int:
             label=self.y[self.it].copy(),
             weight=self.w[self.it].copy() if self.w else None,
         )
-        gc.collect()  # clear up the copy, see if XGBoost access freed memory.
```
trivialfis, line 207: is this change related? I put it there along with the "copy" function above to help prevent XGBoost from accessing freed memory.

tests/python/test_with_sklearn.py
```diff
@@ -12,7 +12,7 @@
 
 rng = np.random.RandomState(1994)
 
-pytestmark = pytest.mark.skipif(**tm.no_sklearn())
+pytestmark = [pytest.mark.skipif(**tm.no_sklearn()), pytest.mark.timeout(30)]
```
trivialfis, line 15: Which test might timeout?

tests/python/test_with_dask.py
```diff
@@ -64,8 +66,8 @@ def client(cluster):
         yield dask_client
 
 
-kRows = 1000
-kCols = 10
+kRows = 300
```
trivialfis, line 69: 1000 is pretty trivial, right? Is it becoming a bottleneck? If the data is too small dask might put them into a single worker.

trivialfis:
Apologies for the inconvenience, the failing test should be fixed by https://github.com/dmlc/xgboost/pull/8323 .

tests/python/testing.py
RAMitchell: The repeated calls to gc are the main cause of slowness in external memory tests. Although I wonder if removing this line is causing the segfault in mgpu tests.

tests/python/test_with_sklearn.py
RAMitchell: Times taken from CI:

```
36.49s call     python/test_with_sklearn.py::test_parameter_tuning
34.58s call     python/test_with_sklearn.py::test_RFECV
```

I made the test under 30s now, but I am placing timeouts to prevent others from adding any slower tests than currently exist.

RAMitchell: Ideally we will be targeting <5s for python tests with some few exceptions.

trivialfis: Got it, thank you for reducing the time!

tests/python/testing.py
trivialfis: That would be really interesting to understand why.

tests/python/test_with_dask.py
RAMitchell: Changed this one back. There is not too much difference.

RAMitchell:
I'm still getting intermittent segfaults in the mgpu tests somehow.

trivialfis (approved):

PR #8316, opened 2022-10-06
Better error message when world size and rank are set as strings

rongou:
When configuring the federated communicator, only allow world size and rank to be set as integers, and provide a more meaningful error message when they are set as strings.

### DIFF
--- a/tests/cpp/plugin/test_federated_communicator.cc
+++ b/tests/cpp/plugin/test_federated_communicator.cc
@@ -40,9 +40,6 @@ class FederatedCommunicatorTest : public ::testing::Test {
   }
 
   void TearDown() override {
-    while (!server_) {
-      sleep(1);
-    }
     server_->Shutdown();
     server_thread_->join();
   }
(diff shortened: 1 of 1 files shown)


rongou:
@trivialfis @hcho3

trivialfis (commented):
not entirely sure whether this is necessary. I personally prefer concise parameters instead of the javascript tradition, the string parameters in XGBoost are more of a legacy from dmlc::Parameter.

rongou:
I think it's just a little surprising to a user that even though she has specified the world rank and size, and happen to use strings, we complain about them not being set. Maybe we should enforce the rule that they should be integers? That'll give a more meaningful error message.

rongou:
@trivialfis please take another look.

tests/cpp/plugin/test_federated_communicator.cc
```diff
@@ -40,6 +40,9 @@ class FederatedCommunicatorTest : public ::testing::Test {
   }
 
   void TearDown() override {
+    while (!server_) {
```
trivialfis, line 43: This seems odd, could you please share when it helps?

trivialfis (commented):
I made some changes to use existing functions for checking arguments, could you please take a look?

The call to sleep seems weird?

tests/cpp/plugin/test_federated_communicator.cc
rongou: This was somehow needed to get the create test to work. Seems more trouble than its worth. Removed.

trivialfis (approved):

PR #8319, opened 2022-10-07
Using column_sampler for optimization of ColWiseBuildHist

razdoburdin:
Hi,
this PR is a part of https://github.com/dmlc/xgboost/pull/7192. 
It continues the optimization of BuildHistKernel, which was started at https://github.com/dmlc/xgboost/pull/8233.
In https://github.com/dmlc/xgboost/pull/8233 I removed the optimization related to column_sampler to simplify the reviewing process. Here I return it. 

The methodology of the benchmarking is the same as in https://github.com/dmlc/xgboost/pull/8218. The only dataset affected by optimization is `santander`. Training speed for `santander `increases twice after applying this optimization. You can find the exact numbers at https://github.com/dmlc/xgboost/pull/8233.

I am looking forward for your review and comments!

### DIFF
--- a/src/common/hist_util.cc
+++ b/src/common/hist_util.cc
@@ -137,37 +137,50 @@ constexpr size_t Prefetch::kNoPrefetchSize;
 struct RuntimeFlags {
   const bool first_page;
   const bool read_by_column;
+  const bool column_sampling;
   const BinTypeSize bin_type_size;
 };
 
 template <bool _any_missing,
           bool _first_page = false,
           bool _read_by_column = false,
+          bool _column_sampling = false,
           typename BinIdxTypeName = uint8_t>
 class GHistBuildingManager {
  public:
   constexpr static bool kAnyMissing = _any_missing;
   constexpr static bool kFirstPage = _first_page;
   constexpr static bool kReadByColumn = _read_by_column;
+  constexpr static bool kColumnSampling = _column_sampling;
   using BinIdxType = BinIdxTypeName;
 
  private:
   template <bool new_first_page>
   struct SetFirstPage {
-    using Type = GHistBuildingManager<kAnyMissing, new_first_page, kReadByColumn, BinIdxType>;
+    using Type = GHistBuildingManager<kAnyMissing, new_first_page, kReadByColumn,
+                                      kColumnSampling, BinIdxType>;
   };
 
   template <bool new_read_by_column>
   struct SetReadByColumn {
-    using Type = GHistBuildingManager<kAnyMissing, kFirstPage, new_read_by_column, BinIdxType>;
+    using Type = GHistBuildingManager<kAnyMissing, kFirstPage, new_read_by_column,
+                                      kColumnSampling, BinIdxType>;
+  };
+
+  template <bool new_column_sampling>
+  struct SetColumnSampling {
+    using Type = GHistBuildingManager<kAnyMissing, kFirstPage, kReadByColumn,
+                                      new_column_sampling, BinIdxType>;
   };
 
   template <typename NewBinIdxType>
   struct SetBinIdxType {
-    using Type = GHistBuildingManager<kAnyMissing, kFirstPage, kReadByColumn, NewBinIdxType>;
+    using Type = GHistBuildingManager<kAnyMissing, kFirstPage, kReadByColumn,
+                                      kColumnSampling, NewBinIdxType>;
   };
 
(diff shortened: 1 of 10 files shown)


hcho3:
Please ignore the failing MacOS tests from GitHub Actions. We're working to fix it.

razdoburdin:
Hi @trivialfis ,
what is your opinion about this optimization?

trivialfis:
Apologies for the slow response, I will get to this tomorrow.

razdoburdin:
> Apologies for the slow response, I will get to this tomorrow.

Hi,
have you any decision about this PR?

src/common/hist_util.cc
```diff
@@ -284,13 +300,14 @@ void ColsWiseBuildHistKernel(const std::vector<GradientPair> &gpair,
   };
 
   const size_t n_features = gmat.cut.Ptrs().size() - 1;
-  const size_t n_columns = n_features;
+  const size_t n_columns = kColumnSampling ? fids.size() : n_features;
```
trivialfis, line 303: Is `fids` empty if there's no sampling?

src/tree/hist/histogram.h
```diff
@@ -76,6 +84,20 @@ class HistogramBuilder {
       buffer_.Reset(this->n_threads_, n_nodes, space, target_hists);
     }
 
+    constexpr float kColsampleTh = 0.1;
+    bool column_sampling = (column_sampler_ != nullptr) &&
```
trivialfis, line 88: When column sampler is nullptr?

src/tree/hist/histogram.h
```diff
@@ -76,6 +84,20 @@ class HistogramBuilder {
       buffer_.Reset(this->n_threads_, n_nodes, space, target_hists);
     }
 
+    constexpr float kColsampleTh = 0.1;
```
trivialfis, line 87: why 0.1?

src/tree/hist/histogram.h
```diff
@@ -76,6 +84,20 @@ class HistogramBuilder {
...
+    bool column_sampling = (column_sampler_ != nullptr) &&
+                           (train_param_.colsample_bytree < kColsampleTh ||
+                            train_param_.colsample_bylevel < kColsampleTh);
```
trivialfis, line 90: how about bynode? Is it used?

trivialfis (commented):
Apologies for the delay. May I ask the specific objective of this PR? Are you trying to optimize the case to build for less columns when sampling is enabled?

I'm concerned with optimizations that don't generalize and requires edge cases. I haven't been able to track which test can confirm the correctness of the code.

src/common/hist_util.cc
razdoburdin: Fids is empty is case of condition [here](https://github.com/dmlc/xgboost/pull/8319/files#diff-8217fa4498a3341ec017b9661c7cc75123cc88fb058c865db9a7f71558a01cb0R91) is false.

src/tree/hist/histogram.h
razdoburdin: I set it to nullptr in [tests ](https://github.com/dmlc/xgboost/pull/8319/files#diff-4bdebd351f1f1f377c74e6764eaf56ff9c6e1b5f9a123ab51665003f164b56d2R49)of the histogram builder without using this optimization.

src/tree/hist/histogram.h
razdoburdin: It is an ad-hoc threshold value.

src/tree/hist/histogram.h
razdoburdin: Not now, maybe later, one can investigate these options more deeply.

razdoburdin:
> Apologies for the delay. May I ask the specific objective of this PR? Are you trying to optimize the case to build for less columns when sampling is enabled?
yes, that was the idea

> I'm concerned with optimizations that don't generalize and requires edge cases. I haven't been able to track which test can confirm the correctness of the code.
It is reasonable. I added a [subcase](https://github.com/dmlc/xgboost/pull/8319/files#diff-42fecfa97f179a4084d1ec4e4a86e845882f64d6acf1d46d03b480010b31dee5R47) for `TestEvaluateSplits ` that forces using the code with` kColumnSampling = true`. The idea is to fill `fids `vector manually. It is enough for pushing the execution to the branch with `kColumnSampling = true`.

trivialfis:
Let me try to look deeper into this PR. The optimization is definitely important but I would like to keep lesser conditional branches than it currently has.

PR #8324, opened 2022-10-08
[ci] reduce pyspark test time

wbo4958:
Without this PR, It takes 70s to finish test_classifier_with_cross_validator 
`70.03s call     python/test_spark/test_spark_local.py::XgboostLocalTest::test_classifier_with_cross_validator`
With this PR, it only takes 17s to be finished `17.43s call     python/test_spark/test_spark_local.py::XgboostLocalTest::test_classifier_with_cross_validator`

I also tried enabling pytest with parallelly running by pytest-xdist and it worked well with pyspark tests, and the time can be reduced from `4m49.335s` to `2m19.574s`.

I don't know if we need to enable pytest-xdist for python tests?

BTW, I also found some tests may have overlaps, like `test_classifier_distributed_weight_eval` and `test_regressor_distributed_weight_eval`, seems we don't need to test all kinds of estimators @WeichenXu123, please correct me.

### DIFF
--- a/tests/test_distributed/test_with_spark/test_spark_local.py
+++ b/tests/test_distributed/test_with_spark/test_spark_local.py
@@ -697,13 +697,14 @@ class XgboostLocalTest(SparkTestCase):
         self.assert_model_compatible(model.stages[0], tmp_dir)
 
     def test_classifier_with_cross_validator(self):
-        xgb_classifer = SparkXGBClassifier()
+        xgb_classifer = SparkXGBClassifier(n_estimators=1)
         paramMaps = ParamGridBuilder().addGrid(xgb_classifer.max_depth, [1, 2]).build()
         cvBin = CrossValidator(
             estimator=xgb_classifer,
             estimatorParamMaps=paramMaps,
             evaluator=BinaryClassificationEvaluator(),
             seed=1,
+            parallelism=4,
             numFolds=2,
         )
         cvBinModel = cvBin.fit(self.cls_df_train_large)


trivialfis (commented):
Not sure about the effect of xdist on other tests as xgboost already uses all the cores for some tests, which may lead to contention.

RAMitchell:
Thank you for working on this! It definitely looked like there was significant overlap in the tests and many could be combined to reduce the overhead.

It may be possible to use xdist just for spark tests, I suspect it will cause problems for other tests but it might be worth experimenting. Maybe we could set OMP_NUM_THREADS before running the tests. The main goal here is to reduce the time of the CI cpu python tests step, which currently stands at around 25mins.

wbo4958:
Yeah, Let's fix them one by one. Currently, this PR can reduce the most time-consuming test.

wbo4958:
@trivialfis please help to trigger the CI for the build. Thx

WeichenXu123 (approved):

hcho3:
I started the CI

NorthSummer:
> Without this PR, It takes 70s to finish test_classifier_with_cross_validator `70.03s call python/test_spark/test_spark_local.py::XgboostLocalTest::test_classifier_with_cross_validator` With this PR, it only takes 17s to be finished `17.43s call python/test_spark/test_spark_local.py::XgboostLocalTest::test_classifier_with_cross_validator`
> 
> I also tried enabling pytest with parallelly running by pytest-xdist and it worked well with pyspark tests, and the time can be reduced from `4m49.335s` to `2m19.574s`.
> 
> I don't know if we need to enable pytest-xdist for python tests?
> 
> BTW, I also found some tests may have overlaps, like `test_classifier_distributed_weight_eval` and `test_regressor_distributed_weight_eval`, seems we don't need to test all kinds of estimators @WeichenXu123, please correct me.

Hi, could you please tell me how to count running time of the functions in the c++ codebase? Thanks!

PR #8296, opened 2022-09-30
[WIP] Enable federated learning support in the CI

rongou:

PR #8325, opened 2022-10-09
[CI] Test federated learning plugin in the CI

hcho3:
Close #8301 
Close #8296 

Notes:
* We test federated learning only in the BuildKite CI pipelines. We don't test federated learning in GitHub Actions, as it only provide 2-core CPUs, so it will take a long time to build gRPC from the source. On the other hand, we have access to a beefy machine (16-core CPUs) on BuildKite.
* gRPC is installed into Docker containers. We don't want to build gRPC from the source over and over again, as it consists of 2000 files (!). Build once and cache it.
* We build both CPU and GPU binaries with federated learning support.
* We do not enable federated learning for Windows platform or ARM64 arch.


CMakeLists.txt
```diff
@@ -67,6 +68,8 @@ address, leak, undefined and thread.")
 option(PLUGIN_DENSE_PARSER "Build dense parser plugin" OFF)
 option(PLUGIN_RMM "Build with RAPIDS Memory Manager (RMM)" OFF)
 option(PLUGIN_FEDERATED "Build with Federated Learning" OFF)
+option(USE_GRPC_FROM_SYSTEM "Use gRPC from the system; if OFF, CMake will download
```
trivialfis, line 71: Can we just leave the grpc installation to system package managers like conda or apt? I'm not entirely whether it's a good idea to maintain dependency with cmake ourselves, it's a build tool after all ...

hcho3: Conda decided to build `grpc-cpp` with GCC 11 (!), so it's incompatible with the compiler we use. 

The system package managers are a better bet. The only issue is that they'll probably give you an older version of gRPC.

@rongou Is it okay if we use an older version of gRPC?

tests/cpp/plugin/test_federated_adapter.cu
```diff
@@ -54,7 +54,7 @@ TEST_F(FederatedAdapterTest, DeviceAllReduceSum) {
   for (auto rank = 0; rank < kWorldSize; rank++) {
     threads.emplace_back(std::thread([rank] {
       FederatedCommunicator comm{kWorldSize, rank, kServerAddress};
-      DeviceCommunicatorAdapter adapter{rank, &comm};
+      DeviceCommunicatorAdapter adapter{0, &comm};
```
trivialfis, line 57: Is the correct?

CMakeLists.txt
hcho3: CentOS 7 yum: gRPC 1.20
Ubuntu 20.04 apt: gRPC 1.16

trivialfis: Maybe we can just leave the grpc/protobuf installation to the CI script and keep the cmake file lean?

tests/cpp/plugin/test_federated_adapter.cu
hcho3: Using `rank` will error out because we now use only a single GPU for gtest.

CMakeLists.txt
hcho3: Got it. I'll update the doc to show how to install gRPC in a Conda environment.

trivialfis: We can probably make a list of (optional) build dependencies in our building from the source page and leave it for others to figure out how they want to install it. There are lots of package managers out there.

## Optional build dependencies (not runtime python deps, which is a much larger set):
- cudatoolkit
- nccl
- grpc
- OPENMP
- rmm
- nvtx
- HDFS
- S3, AZUER filesystem packages?
- ONEAPI ?
- MPI?
- sanitizers?

hcho3: Good idea.

As for which version of gRPC we should use in our CI, let's wait for @rongou's response

tests/cpp/plugin/test_federated_adapter.cu
trivialfis: Got it, out of curiosity, after setting the rank to 0, is the test still testing what it's supposed to be testing?

hcho3: It still creates two threads, one for server and another for client

trivialfis: Good to know. Please add a comment to it, otherwise, someone might try to "fix" it. ;-)

hcho3:
For an unknown reason, invoking the gtest binary via the ctest driver causes a crash, whereas invoking the gtest binary directly is successful.

hcho3:
Nevermind, `FederatedCommunicatorTest.Allreduce` is failing with
```
terminate called after throwing an instance of 'std::runtime_error'
what():  Allreduce RPC failed
```

I have no idea why this is happening.

trivialfis:
@rongou Hi, is this the last PR for federated learning planned for 1.7?

trivialfis:
@hcho3 I couldn't reproduce the error with a bare metal machine, maybe it's a docker-specific constraint on the network?

CMakeLists.txt
rongou: I don't think we are using a lot of newer features in gRPC. The only issue is the CMake plugin in gRPC seems to be fairly recent (see https://www.f-ax.de/dev/2020/11/08/grpc-plugin-cmake-support.html). If you can get the code to build with older versions then it should be fine.

rongou:
@trivialfis yes this should be the last PR for 1.7. #8316 is a nice to have.

As for `find_package(Protobuf CONFIG REQUIRED)`, I think this is needed for the grpc plugin to work properly (see https://www.f-ax.de/dev/2020/11/08/grpc-plugin-cmake-support.html).

CMakeLists.txt
```diff
@@ -117,6 +118,20 @@ endif (BUILD_STATIC_LIB AND (R_LIB OR JVM_BINDINGS))
 if (PLUGIN_RMM AND (NOT BUILD_WITH_CUDA_CUB))
   message(SEND_ERROR "Cannot build with RMM using cub submodule.")
 endif (PLUGIN_RMM AND (NOT BUILD_WITH_CUDA_CUB))
+if (PLUGIN_FEDERATED)
+  if (CMAKE_CROSSCOMPILING)
+    message(SEND_ERROR "Cannot cross compile with federated learning support")
+  endif ()
+  if (BUILD_STATIC_LIB)
+    message(SEND_ERROR "Cannot build static lib with federated learning support")
```
rongou, line 126: Why not? This should work, right?

CMakeLists.txt
```diff
@@ -117,6 +118,20 @@ endif (BUILD_STATIC_LIB AND (R_LIB OR JVM_BINDINGS))
 if (PLUGIN_RMM AND (NOT BUILD_WITH_CUDA_CUB))
   message(SEND_ERROR "Cannot build with RMM using cub submodule.")
 endif (PLUGIN_RMM AND (NOT BUILD_WITH_CUDA_CUB))
+if (PLUGIN_FEDERATED)
+  if (CMAKE_CROSSCOMPILING)
+    message(SEND_ERROR "Cannot cross compile with federated learning support")
```
rongou, line 123: Not sure about this. Shouldn't federated learning also work on non-intel Mac?

hcho3: Cross-compiling gRPC is nontrivial and we don't have bandwidth to do it right now.

CMakeLists.txt
hcho3: We don't have an adequate test coverage to test static libs. Let's come back to it later.

CMakeLists.txt
trivialfis: I managed to get the code to build on conda env. Linking dynamic library failed for some reasons, but linking the static library works.

hcho3:
I fixed https://github.com/dmlc/xgboost/pull/8325#issuecomment-1273189445 by installing gRPC in a Conda environment. My guess is that I accidentally overwrote some headers in the system when I installed gRPC globally.

plugin/federated/CMakeLists.txt
```diff
@@ -1,26 +1,48 @@
...
-    PLUGIN "protoc-gen-grpc=${grpc_cpp_plugin_location}")
+add_library(federated_proto_lib STATIC)
+target_sources(federated_proto_lib PRIVATE
+  ${federated_srcs} ${federated_hdrs}
+  ${federated_grpc_srcs} ${federated_grpc_hdrs})
+target_link_libraries(federated_proto_lib PUBLIC
+  ${_PROTOBUF_LIBPROTOBUF} ${_GRPC_GRPCPP})
+target_include_directories(federated_proto_lib PUBLIC
+  ${CMAKE_CURRENT_BINARY_DIR})
+xgboost_target_properties(federated_proto_lib)
```
trivialfis, line 41: Are these changes still necessary now that compiling grpc is no longer part of CMake?

plugin/federated/README.md
```diff
@@ -5,13 +5,20 @@ This folder contains the plugin for federated learning. Follow these steps to bu
 
 Install gRPC
 ------------
+We highly recommend installing gRPC in a local environment, such as a Conda environment,
```
trivialfis, line 8: Can we just refer to the gRPC document instead of writing our own?

tests/ci_build/Dockerfile.cpu
```diff
@@ -28,7 +28,17 @@ ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
 
 # Create new Conda environment
 COPY conda_env/cpu_test.yml /scripts/
-RUN mamba env create -n cpu_test --file=/scripts/cpu_test.yml
+RUN mamba env create -n cpu_test --file=/scripts/cpu_test.yml && \
+    mamba create -y -n build_env -c conda-forge python=3.9
```
trivialfis, line 32: is 3.9 needed? The XGBoost package claims support for 3.8 and most (all?) other tests are on 3.8.

plugin/federated/README.md
hcho3: Done

hcho3:
@rongou Can you review the changes I made to gtest? I had to randomize the server port, because reusing the same port twice often resulted into a  connection failure.

rongou (approved):

plugin/federated/CMakeLists.txt
hcho3: I reverted the changes.

hcho3:
GTest is finally fixed :tada:

trivialfis (approved):

PR #8326, opened 2022-10-09
add port in trackerConf for spark mode

Francis-W:
allow customize pyTracker / sparkTracker port binding for spark mode.in some case , user need give a fixed port number. 
for example run SparkDriver pod with istio sidecar. add annotation
`traffic.sidecar.istio.io/excludeInboundPorts: {portNumber}`


trivialfis:
cc @wbo4958

Issue #8202, opened 2022-08-25
XGBoost External Memory Version Load SVM File Into DMatrix Error


vandykai:
```
import os
from typing import List, Callable
import xgboost
from sklearn.datasets import load_svmlight_file

class Iterator(xgboost.DataIter):
  def __init__(self, svm_file_paths: List[str]):
    self._file_paths = svm_file_paths
    self._it = 0
    # XGBoost will generate some cache files under current directory with the prefix
    # "cache"
    super().__init__(cache_prefix=os.path.join(".", "cache"))

  
def next(self, input_data: Callable):
    """Advance the iterator by 1 step and pass the data to XGBoost.  This function is
    called by XGBoost during the construction of ``DMatrix``

    """
    if self._it == len(self._file_paths):
      # return 0 to let XGBoost know this is the end of iteration
      return 0

    # input_data is a function passed in by XGBoost who has the exact same signature of
    # ``DMatrix``
    X, y = load_svmlight_file(self._file_paths[self._it])
    input_data(X, y)
    self._it += 1
    # Return 1 to let XGBoost know we haven't seen all the files yet.
    return it

  def reset(self):
    """Reset the iterator to its beginning"""
    self._it = 0

it = Iterator(["my.svm"])
Xy = xgboost.DMatrix(it)
```

my.svm file:
```
1 0:0.45 2:0.45 3:0.89
0 0:0.45 2:0.45 3:0.89
1 0:0.45 2:0.45 3:0.89
```

error message:
ValueError: feature_names must be unique

```
xgboost.__version__
1.6.1
```

trivialfis:
Please use keyword arguments for the `input_data`.

PR #8327, opened 2022-10-10
Require keyword args for data iterator.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/8202 .


trivialfis:
Python can raise a TypeError without the wrapper:
```
TypeError: DataIter._next_wrapper.<locals>.data_handle() takes 1 positional argument but 2 were given
```

But with the wrapper, the message is more informative.

RAMitchell (approved):

PR #8328, opened 2022-10-10
[doc] Display survival demos in sphinx doc. [skip ci]

trivialfis:


trivialfis:
Here is the gallery: https://xgboost--8328.org.readthedocs.build/en/8328/python/survival-examples/index.html .

RAMitchell (approved):

PR #8342, opened 2022-10-13
Avoid blank lines with federated training

rongou:
Right now we use `LOG(CONSOLE) << msg` to print messages during federated training. But the evaluation callback already puts a newline in the message, resulting in blank lines being printed.

@trivialfis


trivialfis (approved):

trivialfis:
hmm , is the CI triggered?

PR #8344, opened 2022-10-15
[pyspark] Cleanup data processing.

trivialfis:
* Enable additional combinations of ctor parameters.
* Unify procedures for QuantileDMatrix and DMatrix.

Close https://github.com/dmlc/xgboost/issues/8341

@wbo4958 Could you please share why is this necessary? Is it possible to use the normal `features` column with `gpu_hist` at least in theory (maybe with lower initialization performance)? https://github.com/dmlc/xgboost/blob/80e10e02aba45cc54795e075720896c9eec9519e/python-package/xgboost/spark/core.py#L282

### DIFF
--- a/python-package/xgboost/spark/data.py
+++ b/python-package/xgboost/spark/data.py
@@ -209,16 +209,25 @@ def create_dmatrix_from_partitions(  # pylint: disable=too-many-arguments
     def append_m(part: pd.DataFrame, name: str, is_valid: bool) -> None:
         nonlocal n_features
         if name == alias.data or name in part.columns:
-            if name == alias.data and feature_cols is not None:
-                array = part[feature_cols]
-            else:
+            if (
+                name == alias.data
+                and feature_cols is not None
+                and part[feature_cols].shape[0] > 0  # guard against empty partition
+            ):
+                array: Optional[np.ndarray] = part[feature_cols]
+            elif part[name].shape[0] > 0:
                 array = part[name]
                 array = stack_series(array)
+            else:
+                array = None
 
-            if name == alias.data:
+            if name == alias.data and array is not None:
                 if n_features == 0:
                     n_features = array.shape[1]
                 assert n_features == array.shape[1]
+
+            if array is None:
+                return
 
             if is_valid:
                 valid_data[name].append(array)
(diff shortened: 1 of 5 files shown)


python-package/xgboost/spark/data.py
```diff
@@ -208,10 +208,14 @@ def create_dmatrix_from_partitions(  # pylint: disable=too-many-arguments
 
     def append_m(part: pd.DataFrame, name: str, is_valid: bool) -> None:
         nonlocal n_features
-        if name in part.columns and part[name].shape[0] > 0:
```
wbo4958, line 211: why get rid of the check "part[name].shape[0] > 0"

wbo4958:
one comment.

python-package/xgboost/spark/data.py
trivialfis: The empty data tests are passing.

rongou:
Not sure, maybe the server didn't have the time to start? The CI instances may be pretty overloaded. I wonder if we should add a wait in the test `SetUp`.

rongou:
Sent PR #8351 that might help with this failure.

python-package/xgboost/spark/data.py
wbo4958: what? really? I will check it, please hold on.

wbo4958:
> Could you please share why is this necessary? Is it possible to use the normal features column with gpu_hist at least in theory (maybe with lower initialization performance)?

@trivialfis Once databricks suggested only enabling feature_cols when use_gpu is enabled. I think it's ok to enable it right now..

python-package/xgboost/spark/data.py
wbo4958: We must add check `part[feature_cols].shape[0] > 0`, or else, stack_series will throw exception. The latest change is ok.

wbo4958:
LGTM

wbo4958 (approved):

trivialfis:
> Once databricks suggested only enabling feature_cols when use_gpu is enabled. I think it's ok to enable it right now..

We will have to leave it to the next release. Making extensive tests for various combination of parameters is not trivial.

PR #8338, opened 2022-10-13
[Don't merge] Root cause for slowdown in CPU tests

hcho3:
See #8334. It appears that building the Python wheel with federated learning support leads to massively longer run-time for some CPU tests. I've verified it on my machine.

Submitting this PR to measure the difference in performance.

PR #8350, opened 2022-10-17
Speed up tests with federated learning enabled

### LINKED ISSUE #8334: Bunch of CPU tests are timing out
https://buildkite.com/organizations/xgboost/analytics/suites/xgboost-pytest-cpu/runs/4025651f-449b-4765-bf89-ba34f3f9af0c

rongou:
Fixes #8334 
Fixes #8338

Enabling federated learning slowed down some cpu-only tests, mainly because we increased the number of symbols drastically in `libxgboost.so`, and the `XGBoostAPIGuard` throwing an exception and ignoring it is very expensive with symbol lookups. Fixes:
* When federated learning is enabled, pass `-Wl,--exclude-libs,ALL` to the linker to ignore symbols in `.a` libraries (see https://sourceware.org/binutils/docs/ld/Options.html).
* Don't throw exception and then ignore it in `XGBoostAPIGuard`.
* Hide c++ symbols when building gRPC (see https://gcc.gnu.org/wiki/Visibility).

Final result running `python/test_updaters.py::TestTreeMethod::test_categorical_ohe`:
* Federated learning disabled: 14.33 seconds
* Federated learning enabled: 13.99 seconds

Thanks to @eordentlich for finding the root cause.

@trivialfis @hcho3

### DIFF
--- a/plugin/federated/CMakeLists.txt
+++ b/plugin/federated/CMakeLists.txt
@@ -29,5 +29,5 @@ target_link_libraries(federated_client INTERFACE federated_proto)
 
 # Rabit engine for Federated Learning.
 target_sources(objxgboost PRIVATE federated_server.cc)
-target_link_libraries(objxgboost PRIVATE federated_client)
+target_link_libraries(objxgboost PRIVATE federated_client "-Wl,--exclude-libs,ALL")
 target_compile_definitions(objxgboost PUBLIC -DXGBOOST_USE_FEDERATED=1)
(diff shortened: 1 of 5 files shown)


hcho3 (approved):
LGTM, assuming the CI passes

trivialfis:
Thank you for the fix! That is quick! Please help set the `enable` parameter to `True`: https://github.com/dmlc/xgboost/blob/031d66ec27d1d99dfc145ffae4f2ac8895305ac1/python-package/xgboost/testing.py#L44.

trivialfis (approved):

hcho3:
The CPU tests are now very fast:
image

hcho3:
@rongou @eordentlich The 1.7 release is unblocked now. Thanks!

PR #8356, opened 2022-10-18
XGBRanker documentation: few clarifications

luca-s:
I would like to propose these updates to the documentation of XGBRanker to clarify few aspects:
- `predict` method can work on multiple groups. I found several online resources  that suggests to split the input data by group and call `predict` on each group individually.  I believe this is a misconception derived by the `group`/`qid` parameter requirement in the `fit` method.
- `qid` parameter in `fit` method needs an example on how it can be used. I couldn't find an online example that makes use of it.

I do not believe my text is particularly good, but if you could just consider the intent and maybe find a better way of writing it, than it would be great as well.

Thanks.

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -1785,7 +1785,7 @@ class XGBRFRegressor(XGBRegressor):
     end_note="""
         .. note::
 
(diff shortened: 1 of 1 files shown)


trivialfis (approved):
Thank you for the improvements!

PR #8369, opened 2022-10-20
Use bst_ulong instead of size_t for CSR conversion

yzhliu:
Some platform, e.g., WebAssembly, has strict typing checking. The mismatch will cause crash.

And they should be `size_t` to be consistent with https://github.com/dmlc/xgboost/blob/master/src/c_api/c_api.cc#L392 and https://github.com/dmlc/xgboost/blob/master/src/data/proxy_dmatrix.cc#L19

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -2307,7 +2307,7 @@ class Booster:
                     _array_interface(csr.indptr),
                     _array_interface(csr.indices),
                     _array_interface(csr.data),
-                    ctypes.c_size_t(csr.shape[1]),
+                    c_bst_ulong(csr.shape[1]),
                     from_pystr_to_cstr(json.dumps(args)),
                     p_handle,
                     ctypes.byref(shape),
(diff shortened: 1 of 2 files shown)


trivialfis (commented):
Thank you for the fix! But can we keep the use of bst_ulong? It's defined as uint64_t, which is more strict than `size_t`. There's `c_bst_ulong` in the `_typing.py` module.

yzhliu:
@trivialfis I'm ok with changing python side to `c_bst_ulong`. But do we really want to do that? cuz all the downstreams are `size_t`, examples like https://github.com/dmlc/xgboost/blob/master/src/data/adapter.h#L352 and `ArrayInterface.Shape()` used by it. IMO using `size_t` would be more general that works for old 32bit machines and future 128bit machine?

not strong opinion though.

trivialfis:
I think we will have improved consistency (which is currently lacking for XGBoost) with other C API functions if we use uint64_t. Using uint64_t **here** to record the shape of the dataset doesn't have practical implications for different memory address sizes. We can optionally define `bst_ulong` as `uint128_t` if the future Skynet decides that having a single memory address space to index exabytes of memory is necessary. ;-)

A side note, c++ `ptrdiff_t` is usually defined as `int64_t` on 64-bit machines, as a result, that's the practical limit of memory indexing size.

trivialfis (approved):
Thank you for the fix!

PR #8373, opened 2022-10-20
compat.py: Ensure Python is >= v3.8

cclauss:
https://pypi.org/project/flake8-2020 says...
```
./python-package/xgboost/compat.py:13:8: YTT201 `sys.version_info[0] == 3` referenced (python4), use `>=`
assert sys.version_info[0] == 3, "Python 2 is no longer supported."
       ^
```

### DIFF
--- a/python-package/xgboost/compat.py
+++ b/python-package/xgboost/compat.py
@@ -10,7 +10,7 @@ import numpy as np
 
 from ._typing import _T
 
-assert sys.version_info[0] == 3, "Python 2 is no longer supported."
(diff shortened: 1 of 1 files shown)


hcho3:
I don't think this is necessary, since we already require a specific Python version in `setup.py`:
https://github.com/dmlc/xgboost/blob/28a466ab518fbb1fccef777bf4efb0b9d8c79514/python-package/setup.py#L373

cclauss:
Python 4 would pass the setup.py test but fail the compat.py test.

hcho3:
We'll come back to this when Python 4 is  released.

PR #8375, opened 2022-10-20
[pyspark] Add type hint to basic utilities.

trivialfis:


python-package/xgboost/spark/utils.py
```diff
@@ -130,13 +126,13 @@ def _get_max_num_concurrent_tasks(spark_context):
...
 
-def _is_local(spark_context) -> bool:
+def _is_local(spark_context: SparkContext) -> bool:
     """Whether it is Spark local mode"""
     # pylint: disable=protected-access
     return spark_context._jsc.sc().isLocal()
 
 
-def _get_gpu_id(task_context) -> int:
+def _get_gpu_id(task_context: BarrierTaskContext) -> int:
```
wbo4958, line 135: Can we change it to TaskContext instead of BarrierTaskContext?

trivialfis: Changed.

trivialfis: Reverted, `getTaskInfos` is exclusive to barrier task context.

wbo4958: Yeah, only _get_host_ip needs BarrerTaskContext. for others, TaskContext is ok.

trivialfis: Since get host ip is called by others, unless we want to perform a down cast we need to pass the type down.

wbo4958 (approved):

PR #8380, opened 2022-10-21
Fix CUDA async stream.

trivialfis:


trivialfis:
The isusse was found when running tests on a sm86 machine with cuda 1.8.

RAMitchell (approved):

PR #8384, opened 2022-10-25
Fix building XGBoost with libomp 15

hcho3:
The Homebrew package for libomp 15 is [now keg-only](https://github.com/Homebrew/homebrew-core/issues/112107#issuecomment-1278042927), meaning that we now need to specify extra build flags.
```
==> Caveats
libomp is keg-only, which means it was not symlinked into /opt/homebrew,
because it can override GCC headers and result in broken builds.

For compilers to find libomp you may need to set:
  export LDFLAGS="-L/opt/homebrew/opt/libomp/lib"
  export CPPFLAGS="-I/opt/homebrew/opt/libomp/include"
```
This PR attempts to fix building XGBoost with libomp 15.

This should be also back-ported to the R package.

### DIFF
--- a/R-package/configure.ac
+++ b/R-package/configure.ac
@@ -29,7 +29,7 @@ fi
 if test `uname -s` = "Darwin"
 then
   OPENMP_CXXFLAGS='-Xclang -fopenmp'
-  OPENMP_LIB='-lomp'
+  OPENMP_LIB='-I/opt/homebrew/opt/libomp/include -lomp -L/opt/homebrew/opt/libomp/lib'
   ac_pkg_openmp=no
   AC_MSG_CHECKING([whether OpenMP will work in a package])
   AC_LANG_CONFTEST([AC_LANG_PROGRAM([[#include <omp.h>]], [[ return (omp_get_max_threads() <= 1); ]])])
--- a/R-package/configure.ac
+++ b/R-package/configure.ac
@@ -28,7 +28,12 @@ fi
 
 if test `uname -s` = "Darwin"
 then
-  HOMEBREW_LIBOMP_PREFIX=`brew --prefix libomp`
+  if command -v brew &> /dev/null
+    HOMEBREW_LIBOMP_PREFIX=`brew --prefix libomp`
+  else
+    # Homebrew not found
+    HOMEBREW_LIBOMP_PREFIX=''
+  fi
   OPENMP_CXXFLAGS="-Xpreprocessor -fopenmp -I${HOMEBREW_LIBOMP_PREFIX}/include"
   OPENMP_LIB="-lomp -L${HOMEBREW_LIBOMP_PREFIX}/lib"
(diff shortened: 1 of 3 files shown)


R-package/configure.ac
```diff
@@ -29,7 +29,7 @@ fi
 if test `uname -s` = "Darwin"
 then
   OPENMP_CXXFLAGS='-Xclang -fopenmp'
-  OPENMP_LIB='-lomp'
+  OPENMP_LIB='-I/opt/homebrew/opt/libomp/include -lomp -L/opt/homebrew/opt/libomp/lib'
```
trivialfis, line 32: hmm .. hardcoded path might not be preferable. Are we sure that users should be using libomp with R considering rtools contains a complete toolchain.

hcho3: You're right. It is a bad idea to hardcode the path. Homebrew actually uses a different location for Intel Macs and for M1 Macs.

Intel Macs: `/usr/local/opt/`
M1 Macs: `/opt/homebrew/opt/`

hcho3:
@trivialfis Can I get another round of review?

R-package/configure.ac
```diff
@@ -28,8 +28,15 @@ fi
 
 if test `uname -s` = "Darwin"
 then
-  OPENMP_CXXFLAGS='-Xclang -fopenmp'
-  OPENMP_LIB='-lomp'
+  if command -v brew &> /dev/null
+  then
+    HOMEBREW_LIBOMP_PREFIX=`brew --prefix libomp`
```
trivialfis, line 33: Actually, can we even assume homebrew is available? I don't feel comfortable around changes in the R package as CRAN has one of the strictest and most specialized environment requirements.

trivialfis: NVM, I notice the condition now.

R-package/configure.ac
```diff
@@ -28,8 +28,15 @@ fi
...
+  else
+    # Homebrew not found
+    HOMEBREW_LIBOMP_PREFIX=''
+  fi
+  OPENMP_CXXFLAGS="-Xpreprocessor -fopenmp -I${HOMEBREW_LIBOMP_PREFIX}/include"
```
trivialfis, line 38: what's the difference between -Xclang and -Xpreprocessor ?

trivialfis (approved):

R-package/configure.ac
hcho3: No idea. I only know that Xprocessor works on my Macbook but Xclang doesn't.

hcho3:
This should now unblock the CI.

R-package/configure.ac
hcho3: I think it has to do with the fact that we have to specify extra flags to the preprocessor specifically, due to non-standard location of OpenMP.

Issue #8134, opened 2022-08-03
Conflicts between dask and pyspark GPU tests.


trivialfis:
https://github.com/dmlc/xgboost/pull/8117 rearranges the order of dask and pyspark GPU tests so that dask tests are run after pyspark tests.  Without the changes, cuDF would allocate data on random devices resulting in non-deterministic errors like each column in a dataframe is allocated on a different device.

This issue tracks the current status of those tests.

PR #8395, opened 2022-10-26
Extract dask and spark test into distributed test.

trivialfis:
- Move test files.
- Run spark and dask separately to prevent conflicts.
- Gather common code into testing module.


trivialfis:
Close https://github.com/dmlc/xgboost/issues/8134 .

tests/test_distributed/test_federated/test_federated.py
```diff
@@ -80,7 +81,7 @@ def run_test(with_ssl: bool = True, with_gpu: bool = False) -> None:
 
 
 if __name__ == '__main__':
-    run_test(with_ssl=True, with_gpu=False)
-    run_test(with_ssl=False, with_gpu=False)
-    run_test(with_ssl=True, with_gpu=True)
-    run_test(with_ssl=False, with_gpu=True)
+    run_federated(with_ssl=True, with_gpu=False)
```
trivialfis, line 84: We should convert these into pytest and run them on CI.

trivialfis:
Should I put the tests as part of XGBoost module?
Pros:
- Users can run the tests, especially helpful if they are troubleshooting distributed environment.
- Distributors can run the tests. This way packaging systems like conda doesn't need to download the git repository or invent their own tests.
- Easier imports when writing tests.

Cons:
- Size of the Python package will grow.
- We need some heavy changes to annotate the tests with Python type hint and make the linter happy.

Some other libraries like dask/distributed, cudf, and cuml are integrating the tests into their Python module. For more implications, see https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#choosing-a-test-layout-import-rules .

trivialfis:
Let's keep the existing layout for now. If the need ever comes up we can make the change then, git can recognize moving files.

hcho3 (approved):

PR #8397, opened 2022-10-27
[CI] Cleanup github action tests.

trivialfis:
- Merge doxygen build with sphinx.
- Use mamba on non-windows Github Action.
~- Set warning as error for sphinx.~


trivialfis:
### Environment setup time
13 minutes on master branch tests:
https://github.com/dmlc/xgboost/actions/runs/3327587462/jobs/5502578459
https://github.com/dmlc/xgboost/actions/runs/3335816298/jobs/5520228525
https://github.com/dmlc/xgboost/actions/runs/3319727165/jobs/5485242261

less than 2 minutes on PR:
https://github.com/dmlc/xgboost/actions/runs/3336969955/jobs/5522745052

doc/conf.py
```diff
@@ -144,7 +144,6 @@
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 exclude_patterns = ['_build']
-html_extra_path = ['./tmp']
```
hcho3, line 147: We need to include `tmp` in the path since we put the JVM and C API doc in it. See https://github.com/dmlc/xgboost/blob/f626799d251c20e183a8764a4c869514e3302c80/doc/conf.py#L36-L55

hcho3 (changes requested):

doc/conf.py
trivialfis: I reverted the change, but the warning as error option is also removed as the tmp directory is not always available on PR branches due to httperror.

hcho3 (approved):

PR #8263, opened 2022-09-22
Enable OpenSSF Scorecard Github Action

joycebrum:
Closes #8255 

As discussed in the #8255 here are the changes to enable the [Scorecard GitHub Action](https://github.com/ossf/scorecard-action) and the [badge](https://openssf.org/blog/2022/09/08/show-off-your-security-score-announcing-scorecards-badges/) to the repo.

The changes are:

- A configuration file in .github/workflow to enable the action
- the badge in the README file

Thanks for considering it and any doubts or concers I'll be happy to help.

Issue #8255, opened 2022-09-21
Enable OpenSSF Scorecard Github Action 


joycebrum:
Hello, I'm working on behalf of Google and the [Open Source Security Foundation][ossf] to help essential open-source projects improve their supply-chain security. Given the relevance and impact that XGBoost has on countless projects, the OpenSSF has identified it as one of the 100 most critical open source projects.

Would you consider adopting an OpenSSF tool called [Scorecards][sc]? Scorecards runs dozens of automated security checks to help maintainers better understand their project's supply-chain security posture. It is developed by the OpenSSF, [in partnership with GitHub][sc-blog].

I see that XGBoost project already follow many best practices regarding security issues that Scorecard checks, such as the Binary-Artifacts, CI-Tests, Code-Review and Vulnerabilities criterias. But still there plenty of criterias you could work on in order to improve the project's security and the Scorecard would help you going through this task. 

To simplify maintainers' lives, the OpenSSF has also developed the [Scorecard GitHub Action][sc-gha]. It is very lightweight and runs on every change to the repository's main branch. The results of its checks are available on the project's [security dashboard](https://github.com/dmlc/xgboost/security), and include suggestions on how to solve any issues (see examples below). The Action does not run or interact with any workflows, but merely parses them to identify possible vulnerabilities. This Action has been adopted by 1800+ projects already.

Would you be interested in a PR which adds this Action? Optionally, it can also publish your results to the OpenSSF REST API, which allows a [badge][badge] with the project's score to be added to its README.

The **changes involved** would be:

- A configuration file to enable the Scorecard Github Action on .github/workflows/scorecard.yml
- (optional) the badge added to the README file.

![Code scanning dashboard with multiple alerts, including Code-Review and Token-Permissions][img-security]

![Detail of a Token-Permissions alert, indicating the specific file and remediation steps][img-detail]

[badge]: https://openssf.org/blog/2022/09/08/show-off-your-security-score-announcing-scorecards-badges/
[ossf]: https://openssf.org/
[sc]: https://github.com/ossf/scorecard
[sc-blog]: https://github.blog/2022-01-19-reducing-security-risk-oss-actions-opensff-scorecards-v4/
[sc-gha]: https://github.com/ossf/scorecard-action
[img-security]: https://user-images.githubusercontent.com/15221358/190184391-84ca1844-259a-4b3b-9c86-74adadbea7f1.png
[...]

PR #8402, opened 2022-10-28
Fix Scorecard Github Action not working 

joycebrum:
Regarding #8255 and https://github.com/dmlc/xgboost/pull/8263#issuecomment-1294613038

I've created a issue to the scorecard team to take a closer look in the error https://github.com/ossf/scorecard-action/issues/998 but it seemed to be fixed in the 2.0.6, as far as I've been able to test.

This PR is upgrading the Scorecard Github Action from 2.0.3 to 2.0.6


hcho3:
Thank you!

PR #8407, opened 2022-11-01
Add back xgboost.rabit in 1.7

### LINKED ISSUE #8406: BUG: xgboost 1.7 no longer has attribute rabit
See modin-project/modin#5171 for context-- https://github.com/modin-project/modin depends on xgboost and uses xgboost.rabit, but as of xgboost 1.7, xgboost no longer has `rabit`, which modin uses here: https://github.com/modin-project/modin/blob/5825856a4ee19ff0dd5d1beaf6090bf601d0b411/modin/experimental/xgboost/utils.py#L95-L102

AFAICT this is a breaking change that according to semantic version should not have gone into xgboost 1.7.

**Steps to reproduce**:

```shell
pip install xgboost==1.6
python -c "import xgboost; print(xgboost.rabit)"
pip install xgboost==1.7
python -c "import xgboost; print(xgboost.rabit)"
```

I'm on:

- macOS Monterey 12.4
- MacBook Pro (16-inch, 2019)
- Processor: 2.3 GHz 8-core intel core i9
- Memory: 16 GB 2667 MHz DDR4
- Python: 3.10.6

hcho3:
Closes #8406 by making `xgboost.rabit` available in 1.7.

### DIFF
--- a/python-package/xgboost/__init__.py
+++ b/python-package/xgboost/__init__.py
@@ -4,7 +4,7 @@ Contributors: https://github.com/dmlc/xgboost/blob/master/CONTRIBUTORS.md
 """
 
 from . import tracker  # noqa
-from . import collective, dask
+from . import collective, dask, rabit
 from .core import (
     Booster,
     DataIter,
--- a/python-package/xgboost/rabit.py
+++ b/python-package/xgboost/rabit.py
@@ -1,0 +1,268 @@
+"""Compatibility shim for xgboost.rabit; to be removed in 2.0"""
+import ctypes
+import warnings
+from enum import IntEnum, unique
+import logging
+import pickle
+from typing import Any, TypeVar, Callable, Optional, cast, List, Union
+
+import numpy as np
+
+from .core import _LIB, c_str, _check_call
+
+LOGGER = logging.getLogger("[xgboost.rabit]")
+
+
+def _deprecation_warning():
+    return (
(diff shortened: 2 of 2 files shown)


hcho3:
@trivialfis @rongou Can I get a review for this?

rongou:
@hcho3 the problem is this won't work for xgboost any more since the c++ code depends on the communicators. If we have to keep the existing api, one option is to delegate everything to the `xgboost.collective` module.

hcho3:
@rongou Doesn't `xgboost.collective` fall down to Rabit in the default setting? For example, `xgboost.collective.init()` calls `rabit::Init` internally.

rongou:
Yes but it needs to set up the `RabitCommunicator`.

hcho3:
@rongou I see. Do you have bandwidth to set up the delegation to `xgboost.collective`? I can give it a try, but I'm afraid I'd get it wrong, given the substantial difference between `xgboost.collective.init` (takes in dictionary) and `xgboost.rabit.init` (takes in list of strings)

rongou:
Ok I'll take a look.

hcho3:
Thanks! Let me know if there's anything I can help. Also, feel free to file a new pull request when you're done.

rongou:
Do we just want to add back `init`/`finalize` to fix `modin`, or do you want to add all other functions too?

hcho3:
@rongou Let's try to add all functions.

PR #8427, opened 2022-11-05
Fix 1.7.1 version file.

trivialfis:
We should run cmake after changing the cmake version configuration.


hcho3:
Oops, my bad.

hcho3 (approved):

hcho3:
Can we document this in one of the dev scripts?

trivialfis:
@hcho3 It's in https://github.com/dmlc/xgboost/blob/master/doc/contrib/release.rst . (step 2) It's difficult to spot, any suggestions?

I will try to make a more automated script in the future.

Issue #8424, opened 2022-11-04
Vertical Federated Learning RFC


rongou:
## Motivation
XGBoost 1.7.0 introduced the initial support for Federated Learning. However, only horizontal federated learning is supported. Training samples are assumed to be split horizontally, i.e. each participant has a subset of samples, but with all the features and labels. In many real world applications, data is split vertically: each participant has all the samples, but only a partial list of features, and not all participants have access to the labels. It would be useful to support vertical federated learning in XGBoost.

## Goals
* Enhance XGBoost to support vertical federated learning.
* Support using [NVFlare](https://github.com/NVIDIA/NVFlare) to coordinate the learning process, but the design should be amenable to support other federated learning platforms.
* Efficiency: training speed should be close to traditional distributed training.
* Accuracy: should be close to centralized learning.

## Non-Goals
* Initially we will assume the federated environment is non-adversarial, and will not provide strong privacy guarantees. This will be improved upon in later iterations.
* We will not support data owners dropping out during the learning process.

## Assumptions
In vertical federated learning, before model training, the participants need to jointly compute the common IDs of their private sets. This is called private set intersection (PSI). For XGBoost, we assume this is already done, and users may rely on some other library/framework for this step.

Similar to horizontal federated learning, we make some simplifying assumptions::
* A few trusted partners jointly train a model.
* Reasonably fast network connection between each participant and a central trusted party.

## Risks
The current XGBoost codebase is fairly complicated and hard to modify. Some code refactoring needs to happen first, before support for vertical federated learning can be added. Care must be taken to not break existing functionality, or make regular training harder.

## Design
LightGBM, a gradient boosting library similar to XGBoost, supports “feature parallel” distributed learning.

![feature_parallel](https://user-images.githubusercontent.com/497101/200077543-0a1806fe-1b25-4ddc-84ca-fc2738909ce4.png)

Conceptually, feature parallelism is similar to vertical federated learning. A possible design is to first enhance XGBoost distributed training to support feature parallelism, and then build vertical federated learning on top of it. This would benefit the wider user community, thus greatly reducing the risks involved in refactoring XGBoost’s code base.

### Feature Parallelism
XGBoost has an internal training parameter called `DataSplitMode`, which can be set to `auto`, `col`, and `row`. However, it’s currently not exposed to the end user, and can only be set to `row` for distributed training.

In order to support column-based data split for distributed training, we need to do the following:
[...]

PR #8434, opened 2022-11-07
Use `DataSplitMode` to configure data loading

rongou:
This is a pure refactoring to leverage the `DataSplitMode` enum to configure data loading, so that we'd have more flexibility (e.g. adding column-wise split support).

Part of #8424

### DIFF
--- a/src/cli_main.cc
+++ b/src/cli_main.cc
@@ -115,6 +115,7 @@ struct CLIParam : public XGBoostParameter<CLIParam> {
         .add_enum("auto", 0)
         .add_enum("col", 1)
         .add_enum("row", 2)
+        .add_enum("none", 3)
         .describe("Data split mode.");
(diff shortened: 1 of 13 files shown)


rongou:
@trivialfis

src/cli_main.cc
```diff
@@ -157,8 +158,14 @@ struct CLIParam : public XGBoostParameter<CLIParam> {
     if (name_pred == "stdout") {
       save_period = 0;
     }
-    if (dsplit == 0 && collective::IsDistributed()) {
-      dsplit = 2;
+    if (dsplit == 0) {
+      if (collective::IsFederated()) {
+        dsplit = 3;
```
trivialfis, line 163: Can we use the enum symbol here instead of integer?

rongou: Done.

trivialfis (approved):
LGTM, assuming all tests pass.

PR #8436, opened 2022-11-08
[doc] Update C tutorial. [skip ci]

trivialfis:
- Use rst references instead of doxygen links.
- Replace deprecated functions.

Close https://github.com/dmlc/xgboost/issues/8400 .


trivialfis:
I'm thinking that we might just drop the online Doxygen doc and use sphinx doc exclusively. For now, one issue with `breathe` is that examples are not linked in the rendered document, which might not be a big issue for the C API document.

hcho3 (approved):

PR #8437, opened 2022-11-08
SO_DOMAIN do not support on AIX, using getsockname instead

zheddie:
getsockopt do not support SO_DOMAIN on AIX/IBM i . Using the getsockname() instead to get the domain info. thanks.

### DIFF
--- a/include/xgboost/collective/socket.h
+++ b/include/xgboost/collective/socket.h
@@ -287,11 +287,22 @@ class TCPSocket {
 #elif defined(__APPLE__)
     return domain_;
 #elif defined(__unix__)
+#ifndef __PASE__
     std::int32_t domain;
     socklen_t len = sizeof(domain);
     xgboost_CHECK_SYS_CALL(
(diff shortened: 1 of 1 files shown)


trivialfis (commented):
Thank you for the fix, does it work if the socket is not yet binded?

trivialfis:
You can check the result based on cpp unittest `testxgboost --gtest_filter=Socket.Basic`.

zheddie:
@trivialfis  thanks. Can you help me how to build out the testxgboost utility?

zheddie:
> Thank you for the fix, does it work if the socket is not yet binded?

Yes. the domain could be retrieved just after the socket been created by socket() routine.

trivialfis:
XGBoost requires googletest to build its c++ tests.

```
cmake path/to/xgboost -DGOOGLE_TEST=ON
make
./testxgboost

trivialfis (approved):

zheddie:
BTW, how can we make sure this fix tobe integrated in next release/fixpack ? @trivialfis  thanks.

trivialfis:
@zheddie We will.

PR #8428, opened 2022-11-05
[R] Fix CRAN test notes.

trivialfis:
- Limit the number of used CPU cores in examples.
- Add document for the constraint.
- Bring back the cleanup script.

Replaces https://github.com/dmlc/xgboost/pull/8423 .
Related: https://github.com/dmlc/xgboost/issues/8366#issuecomment-1304096969 .

PR #8440, opened 2022-11-08
[backport] [R] Fix CRAN test notes. (#8428)

trivialfis:
- Limit the number of used CPU cores in examples.
- Add a note for the constraint.
- Bring back the cleanup script.


hcho3 (approved):
mypy is failing mysteriously. Can we ignore this for the purpose of the R package release?

trivialfis:
We will ignore it for now. Curiously, I can't reproduce the errors on my device using the same version of mypy.

PR #8442, opened 2022-11-08
Always link federated proto statically

### LINKED ISSUE #8439: Target `federated_proto` is not installed, resulting in missing library dependency missing
When `PLUGIN_FEDERATED` is set to `ON`, target `federated_proto` is built and linked to the main `xgboost` target. However, no installation rule is specified for `federated_proto`. Thus, on a Linux platform, when `BUILD_SHARED_LIBS` is set to `ON`, the installed `libxgboost.so` will link with `libfederated_proto.so` but cannot find it in its installation path.

Steps to reproduce:

```
cd xgboost && mkdir build && cd build
cmake -DPLUGIN_FEDERATED=ON -DBUILD_WITH_CUDA_CUB=ON -DUSE_CUDA=ON -DUSE_NCCL=ON \
    -DCMAKE_INSTALL_PREFIX=install -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release ..
cmake --build . --config Release
cmake --install . --config Release
ldd install/lib/libxgboost.so | grep federated
```

The output for the last command will then be
```
        libfederated_proto.so => not found
```
which is not expected.

This issue can be constantly reproduced with current master(8e76f5f). My environment is Archlinux, with all packages installed from official repository and having version as of date 2022-11-07. Required packages are installed using `pacman -S base-devel cmake cuda nccl grpc`. CUDA version is 11.8 and GCC version is 12.2.

rongou:
Fixes #8439

Including the federated proto only adds a few MBs to `libxgboost.so`, so it's a little silly to have a separate `.so` for it.

### DIFF
--- a/plugin/federated/CMakeLists.txt
+++ b/plugin/federated/CMakeLists.txt
@@ -6,7 +6,7 @@ find_package(gRPC CONFIG REQUIRED)
 message(STATUS "Found gRPC: ${gRPC_CONFIG}")
 
 # Generated code from the protobuf definition.
-add_library(federated_proto federated.proto)
+add_library(federated_proto STATIC federated.proto)
 target_link_libraries(federated_proto PUBLIC protobuf::libprotobuf gRPC::grpc gRPC::grpc++)
 target_include_directories(federated_proto PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
 xgboost_target_properties(federated_proto)


rongou:
@trivialfis

hcho3 (approved):

trivialfis (approved):

PR #8445, opened 2022-11-09
Fix R package build on CI.

trivialfis:
@hcho3 Is it possible to trigger a build for the R package on the PR branch instead of the master branch?

### DIFF
--- a/tests/buildkite/build-rpkg-win64-gpu.ps1
+++ b/tests/buildkite/build-rpkg-win64-gpu.ps1
@@ -5,12 +5,12 @@ $ErrorActionPreference = "Stop"
 Write-Host "--- Build XGBoost R package with CUDA"
 
 nvcc --version
+$arch_flag = "-DGPU_COMPUTE_VER=75"
+
+bash tests/ci_build/build_r_pkg_with_cuda_win64.sh $Env:BUILDKITE_COMMIT
+if ($LASTEXITCODE -ne 0) { throw "Last command failed" }
+
 if ( $is_release_branch -eq 1 ) {
-  $arch_flag = "-DGPU_COMPUTE_VER=75"
-
-  bash tests/ci_build/build_r_pkg_with_cuda_win64.sh $Env:BUILDKITE_COMMIT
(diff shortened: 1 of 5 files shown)


hcho3:
@trivialfis You can comment out the `if` condition in this script:
https://github.com/dmlc/xgboost/blob/a83748eb4521581f9f37fc40ae5651c150d720fc/tests/buildkite/build-rpkg-win64-gpu.ps1#L8-L21

hcho3:
@trivialfis We can always build the R package on PRs to ensure that we don't break the build accidentally.

hcho3:
See my latest commit.

trivialfis:
@hcho3 Thank you for the reference. Will try to fix the Linux build as well.

hcho3:
Yeah, skipping the R package build on PRs was not such a bright idea.

trivialfis:
For the longer term, I think we need to make cmake a system requirement https://github.com/dmlc/xgboost/blob/a83748eb4521581f9f37fc40ae5651c150d720fc/R-package/DESCRIPTION#L70

hcho3:
Pretty sure all issues are fixed. Fingers crossed.

hcho3 (approved):

tests/buildkite/build-rpkg-win64-gpu.ps1
```diff
@@ -5,12 +5,12 @@ $ErrorActionPreference = "Stop"
 Write-Host "--- Build XGBoost R package with CUDA"
 
 nvcc --version
-if ( $is_release_branch -eq 1 ) {
-  $arch_flag = "-DGPU_COMPUTE_VER=75"
+$arch_flag = "-DGPU_COMPUTE_VER=75"
```
hcho3, line 8: This line is actually not used. In a follow-up PR, I will find ways to only build `sm_75` to speed up builds for pull requests. (For `master`, still need to build all `sm`s.)

hcho3:
The CI is now healthy again. Thanks!

trivialfis:
@hcho3 Thank you for the nice work!

Issue #8377, opened 2022-10-20
[DOC] Add documentation about training on GPUs and doing inference on CPU-only systems


beckernick:
Some users want to train their on systems with GPUs, save the model, and then load it in production on CPU-only systems to do inference. This use case is fully supported, but some users may not be aware of it and/or think that the results may differ (beyond floating point accumulation errors). Examples include [this Twitter thread](https://twitter.com/MattiasLamotte/status/1582857806759153665) and several issues in the past few months (https://github.com/dmlc/xgboost/issues/8362, https://github.com/dmlc/xgboost/issues/8148, and https://github.com/dmlc/xgboost/issues/8047 )

This might be a great thing to add to the XGBoost docs (maybe in the [Model IO section](https://xgboost.readthedocs.io/en/stable/tutorials/saving_model.html#introduction-to-model-io)).

Happy to put up a PR for discussion if there's interest.

trivialfis:
Thank you for the suggestion and the offer @beckernick ! Maybe we can write the doc here: https://xgboost.readthedocs.io/en/stable/gpu/index.html ? I'm not sure where people get the idea of the model is exclusive to the hardware that generated it, maybe some part of XGBoost's existing document is misleading?

beckernick:
Hmm, I think that could be a good place, too. Happy to defer to your suggestion. Feel free to assign this to me.

And that's a great question -- I'm honestly not sure. I wonder if people have just been having general challenges with cross-environment (or cross-XGBoost version) model save/load issues and suspecting something GPU related rather than challenges of compatibility with pickle. This might slowly fade as people see https://xgboost.readthedocs.io/en/stable/tutorials/saving_model.html#a-note-on-backward-compatibility-of-models-and-memory-snapshots .

trivialfis:
Thank you for looking into this. I also opened a PR for cleaning up the GPU document: https://github.com/dmlc/xgboost/pull/8378 . Feel free to add a brief summary of CPU-GPU interoperability.

beckernick:
Sorry I wasn't able to get back to this in time for that PR. Will open a follow up PR for discussion

PR #8450, opened 2022-11-09
Explain CPU/GPU model interoperability in GPU documentation

beckernick:
This PR:
- Adds a brief section to the GPU documentation about how models trained on GPUs can be used on CPU-only systems for inference.

Closes https://github.com/dmlc/xgboost/issues/8377


trivialfis (approved):
Thank you for the PR!

PR #8432, opened 2022-11-07
Work with newer thrust and libcudacxx

robertmaynard:

PR #8454, opened 2022-11-10
Work with newer thrust and libcudacxx for master

robertmaynard:
Port https://github.com/dmlc/xgboost/pull/8432 to master


trivialfis:
Thank you!

PR #8460, opened 2022-11-11
Hash-Pinning Dependencies

### LINKED ISSUE #8438: Hash-Pinning Dependencies
To contextualization, the [Pinned Dependencie Check](https://github.com/ossf/scorecard/blob/2cbf5afd5460b51fd40939f8c44b32543b1a0bcb/docs/checks.md#pinned-dependencies) is a Scorecard Check to help you to identify dependencies that are not pinned by hash.

I've seen that changing the dependencies to be hash-pinned would be a troublesome work for you to do, considering that there are lots of dependencies to change and that it would be a manual job.

I would like to help you on this update. Let me know if it's OK for you that I open a PR with the changes to hash-pinning the dependencies.

joycebrum:
Closes #8438

### DIFF
--- a/.github/workflows/jvm_tests.yml
+++ b/.github/workflows/jvm_tests.yml
@@ -15,16 +15,16 @@ jobs:
         os: [windows-latest, ubuntu-latest, macos-11]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@e2f20e631ae6d7dd3b768f56a5d2af784dd54791 # v2.5.0
       with:
         submodules: 'true'
 
-    - uses: actions/setup-python@v2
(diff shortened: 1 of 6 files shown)


joycebrum:
Yes it is ready, I was waiting for some actions to run before changing it to ready.

I decided to break it into more PRs, because I'm still trying to understand the best approach to deal with the Dockerfiles and the python dependencies mentioned in the actions/dockerfiles, I'm concerned that they would be relevant to tests and pinning it would affect the tests.

So at first we should only pin these safe ones and later when I got more details on what can or cannot be pinned I can open another issue to continue.

Thank you for the concerns in the issue discussion, all of them were pretty relevant.

PR #8462, opened 2022-11-12
Alter xgb.train() to return results of multiple evaluation metrics (Release 1.6.0)

MrJerryTAO:
In the latest XGBoost R package, Version 1.6.0.1 by April 2022, the fundamental function xgb.train() does not produce evaluation results of more than one metric although the help document suggests that multiple entries in the eval_metric = list() argument are allowed. However, the function xgb.cv() enables returning multiple evaluation metrics. Therefore, after comparing the source code of these two functions, I found a straightforward fix to alter xgb.train(), which enables returning results of multiple metrics for both training and testing data sets. See my R script at https://github.com/MrJerryTAO/MrJerryTAO.git.


hcho3:
I'm closing this in favor of #8657. Generally we apply all changes to the `master` branch.

PR #8469, opened 2022-11-16
[jvm-packages] [pyspark] Make cuDF optional in PySpark package

### LINKED ISSUE #8467: Xgboost regressor training with GPU does not work when python environment does not have "cudf" package
Testing on xgboost==1.7.1

Reproducing code:

```
# run on a spark cluster configured with GPU.

df_train = spark.createDataFrame([
    (Vectors.dense(1.0, 2.0, 3.0), 0, False, 1.0),
    (Vectors.sparse(3, {1: 1.0, 2: 5.5}), 1, False, 2.0),
    (Vectors.dense(4.0, 5.0, 6.0), 0, True, 1.0),
    (Vectors.sparse(3, {1: 6.0, 2: 7.5}), 1, True, 2.0),
] * 100, ["features", "label", "isVal", "weight"])
df_test = spark.createDataFrame([
    (Vectors.dense(1.5, 2.0, 3.0), ),
    (Vectors.sparse(3, {1: -1.0, 2: 5.5}), ),
] * 100, ["features"])

from xgboost.spark import SparkXGBRegressor
xgb_regressor = SparkXGBRegressor(
    num_workers=2,
    max_depth=5, missing=0.0, use_gpu=True,
    validation_indicator_col='isVal', weight_col='weight',
    early_stopping_rounds=1, eval_metric='rmse'
)
xgb_reg_model = xgb_regressor.fit(df_train)
xgb_reg_model.transform(df_test).collect()
```

Error:
```
Py4JJavaError: An error occurred while calling z:org.apache.spark.api.python.PythonRDD.collectAndServe.
: org.apache.spark.SparkException: Job aborted due to stage failure: Could not recover from a failed barrier ResultStage. Most recent failure reason: Stage failed because barrier task ResultTask(71, 0) finished unsuccessfully.
org.apache.spark.api.python.PythonException: 'ModuleNotFoundError: No module named 'cudf''. Full traceback below:
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-824c7a6f-bc83-452f-baf0-520e7f6f9371/lib/python3.9/site-packages/xgboost/spark/core.py", line 809, in _train_booster
    booster = worker_train(
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-824c7a6f-bc83-452f-baf0-520e7f6f9371/lib/python3.9/site-packages/xgboost/spark/data.py", line 309, in create_dmatrix_from_partitions
    dtrain = make_qdm(train_data, gpu_id, meta, None, params)
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-824c7a6f-bc83-452f-baf0-520e7f6f9371/lib/python3.9/site-packages/xgboost/spark/data.py", line 167, in make_qdm
    m = QuantileDMatrix(it, **params, ref=ref)
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-824c7a6f-bc83-452f-baf0-520e7f6f9371/lib/python3.9/site-packages/xgboost/core.py", line 620, in inner_f
    return func(**kwargs)
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-824c7a6f-bc83-452f-baf0-520e7f6f9371/lib/python3.9/site-packages/xgboost/core.py", line 1386, in __init__
    self._init(
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-824c7a6f-bc83-452f-baf0-520e7f6f9371/lib/python3.9/site-packages/xgboost/core.py", line 1445, in _init
    it.reraise()
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-824c7a6f-bc83-452f-baf0-520e7f6f9371/lib/python3.9/site-packages/xgboost/core.py", line 488, in reraise
    raise exc  # pylint: disable=raising-bad-type
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-824c7a6f-bc83-452f-baf0-520e7f6f9371/lib/python3.9/site-packages/xgboost/core.py", line 478, in _handle_exception
    return dft_ret
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-824c7a6f-bc83-452f-baf0-520e7f6f9371/lib/python3.9/site-packages/xgboost/core.py", line 534, in <lambda>
    return self._handle_exception(lambda: self.next(input_data), 0)
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-824c7a6f-bc83-452f-baf0-520e7f6f9371/lib/python3.9/site-packages/xgboost/spark/data.py", line 99, in next
    data=self._fetch(self._data[alias.data]),
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-824c7a6f-bc83-452f-baf0-520e7f6f9371/lib/python3.9/site-packages/xgboost/spark/data.py", line 85, in _fetch
    import cudf  # pylint: disable=import-error
ModuleNotFoundError: No module named 'cudf'
```

hcho3:
Closes #8467

Not using cuDF may have adverse performance implication, but at least it's better not to crash due to missing `cudf`.

### DIFF
--- a/python-package/xgboost/compat.py
+++ b/python-package/xgboost/compat.py
@@ -43,6 +43,11 @@ except ImportError:
     pandas_concat = None
     PANDAS_INSTALLED = False
 
+
+# cuDF
+CUDF_INSTALLED = importlib.util.find_spec("cudf") is not None
+
+
 # sklearn
 try:
     from sklearn.base import BaseEstimator as XGBModelBase
--- a/python-package/xgboost/spark/data.py
+++ b/python-package/xgboost/spark/data.py
@@ -5,7 +5,7 @@ from typing import Any, Callable, Dict, Iterator, List, Optional, Sequence, Tupl
 import numpy as np
 import pandas as pd
 from scipy.sparse import csr_matrix
-from xgboost.compat import concat
+from xgboost.compat import CUDF_INSTALLED, concat
 
 from xgboost import DataIter, DMatrix, QuantileDMatrix
 
@@ -81,7 +81,7 @@ class PartIter(DataIter):
         if not data:
             return None
 
-        if self._device_id is not None:
+        if self._device_id is not None and CUDF_INSTALLED:
             import cudf  # pylint: disable=import-error
             import cupy as cp  # pylint: disable=import-error
 


hcho3:
@WeichenXu123 Would you like to test this patch?

WeichenXu123 (approved):

WeichenXu123:
@hcho3 Btw, for the released xgboost 1.7.1, does it compatible with all of cuda 11.3 / 11.4 / 11.5 ?

hcho3:
Yes. We build our binaries using CUDA 11.0.

WeichenXu123:
> Yes. We build our binaries using CUDA 11.0.

awesome!
We hope to have a patch release 1.7.2 including this patch.

hcho3:
@WeichenXu123 Have you had a chance to test this patch? Did it work?

WeichenXu123:
> @WeichenXu123 Have you had a chance to test this patch? Did it work?

works

hcho3:
Great! Let's merge this after @trivialfis approves.

wbo4958:
Interesting, I suppose the "vector" features and use_gpu will not reach that piece of code.  @trivialfis

WeichenXu123:
Hold on.. with your change, although GPU training works without `cudf` installed, but there's a performance regression (in my test on taxi dataset about 2x slower) compared with the code when I contributed the xgboost spark estimator.
@hcho3 @wbo4958 

I think when `cudf` not installed, we should set `use_qdm=False`

wbo4958:
I am also concerned about what if one worker has installed cudf while others don't install cudf. It will be messed up. 
If it is reasonable about defining a parameter to control use_qdm ?

WeichenXu123:
@wbo4958 New PR created https://github.com/dmlc/xgboost/pull/8471

trivialfis (commented):
> Hold on.. with your PR change, although GPU training works without cudf installed, but there's a performance regression (in my test on taxi dataset about 2x slower) compared with the code when I contributed the xgboost spark estimator.

This will depend on the size of the dataset and generated model. For a small model, the overhead might be significant due to expensive initialization using `QuantileDMatrix` on CPU. For a larger model, the cost can be amortized.

I think the overhead is fine. If you want to use GPU pipeline fully you need GPU-based data storage like cudf. There's little benefit to working around that. Based on our discussion @wbo4958 and other stakeholders, we might make cuDF a hard requirement for pyspark-rapidsai in the future anyway with the help of cuda IPC.

We can move ahead with this PR and let users know there's overhead between CPU/GPU data conversion.

> I am also concerned about what if one worker has installed cudf while others don't install cudf. It will be messed up.

This is a bug in the cluster. We can set a check before the construction of DMatrix to make sure all workers are doing the same thing.

hcho3:
> we might make cuDF a hard requirement for pyspark-rapidsai in the future

I hope we can make it easier to install cuDF with pip.

trivialfis:
> I hope we can make it easier to install cuDF with pip.

It requires an additional pypi index, but shouldn't be too difficult.

hcho3:
> It requires an additional pypi index, but shouldn't be too difficult.

@WeichenXu123 had some trouble installing cuDF using pip. See https://github.com/dmlc/xgboost/issues/8467

WeichenXu123:
> @trivialfis
> 
> > It requires an additional pypi index, but shouldn't be too difficult.
> 
> @WeichenXu123 had some trouble installing cuDF using pip. See #8467

I already uses the NVIDA index for pip install cuDF, but the installed version only supports cuda >= 11.5, see my comments in https://github.com/dmlc/xgboost/issues/8467

WeichenXu123:
@trivialfis 

> This will depend on the size of the dataset and generated model. For a small model, the overhead might be significant due to expensive initialization using QuantileDMatrix on CPU. For a larger model, the cost can be amortized.
> 
> I think the overhead is fine. If you want to use GPU pipeline fully you need GPU-based data storage like cudf. There's little benefit to working around that. Based on our discussion @wbo4958 and other stakeholders, we might make cuDF a hard requirement for pyspark-rapidsai in the future anyway with the help of cuda IPC.
> 
> We can move ahead with this PR and let users know there's overhead between CPU/GPU data conversion.


Currently cuDF cannot support databricks runtime 2.0 (to be released soon), so we need a version that works well without cuDF and does not introduce performance regression, i.e. in this case, for GPU training, we should still use plain DMatrix instead of QuantileDMatrix.

trivialfis:
Hi all, I'm in favor of merging this PR with an additional check that every worker has the same required package or performing the same steps based on https://github.com/dmlc/xgboost/pull/8469#pullrequestreview-1183596175 . What do you think?

WeichenXu123:
@trivialfis 

> Hi all, I'm in favor of merging this PR with an additional check that every worker has the same required package or performing the same steps based on [#8469 (review)](https://github.com/dmlc/xgboost/pull/8469#pullrequestreview-1183596175) . What do you think?

Can we make xgboost use `DMatrix` when cuDF is not installed ? To avoid the performance regression here https://github.com/dmlc/xgboost/pull/8469#issuecomment-1316770295 :)

trivialfis:
The QDM doesn't depend on cuDF.  With cuDF the performance can be better but it's not a hard requirement.

WeichenXu123:
But QDM without cuDF performs slower than DMatrix when model is small (it should be a common case)

wbo4958:
I guess @WeichenXu123 's PR is ok to introduce a parameter use_qdm (default to true). But  I still don't want to check the cudf in the driver side. and meanwhile, I'd also like xgboost to throw an exception with the message (to set use_qdm to false) when no cudf installed. Does that make sense?

trivialfis:
I don't prefer the extra parameter as I don't think it's necessary.

Allow me to summarize the issues and conflicts here.
- When cuDF is not available, QDM slows down performance for small models.
- cuDF pypi support is limited to latest CUDA versions with enhanced compatibility, as a result, we prefer not to make it a hard dependency for pyspark-xgboost GPU pipeline. Databricks' current runtime doesn't have the latest CUDA version.
- Optional QDM may introduce inconsistency between pyspark and other Python interfaces including sklearn and dask.
- DMatrix increases performance on CPU input and GPU model but decreases memory usage  #efficiency.

Can we meet at a middle ground? We make QDM optional by checking whether cuDF is available to workers and use rabit allreduce to check all workers share the same result from the predicate. After the work on pyspark with CUDA IPC and an upgrade to databricks' runtime, we revert the change and proceed with using cuDF as default?

trivialfis:
My lack of support for the additional parameters is its combinatory increase in complexity. We have `use_gpu`, `tree_method`, `use_qdm`, `feature_cols` all affecting the choice. The logic is unnecessarily complex for long-term maintenance.

WeichenXu123:
Then what about add a special handling for databricks , if on databricks runtime (DATARICKS_RUNTIME_VERSION environ exists), then use DMatrix

trivialfis:
@WeichenXu123 Do you prefer XGBoost checking databricks env over checking the availability of cuDF?

WeichenXu123:
> @WeichenXu123 Do you prefer XGBoost checking databricks env over checking the availability of cuDF?

It's OK too. :)

trivialfis:
Can we continue this PR? Checking cuDF seems to be simpler and users have some control.

WeichenXu123:
> Can we continue this PR? Checking cuDF seems to be simpler and users have some control.

What's the final decision ? Based on this PR, for databricks runtime, add an additional check and use DMatrix instead ?

trivialfis:
Let's make QDM optional based on the cuDF check.

WeichenXu123:
@trivialfis 

> Let's make QDM optional based on the cuDF check.

Got it. I will update my PR.

WeichenXu123:
@trivialfis 
My PR updated according to your latest suggestion: https://github.com/dmlc/xgboost/pull/8471

PR #8471, opened 2022-11-16
[jvm-packages] [pyspark] Make QDM optional based on cuDF check

WeichenXu123:

Closes https://github.com/dmlc/xgboost/issues/8467
Closes https://github.com/dmlc/xgboost/pull/8469

If cuDF is installed and using hist / gpu_hist tree method, use QDM, otherwise use DMatrix.

### DIFF
--- a/python-package/xgboost/compat.py
+++ b/python-package/xgboost/compat.py
@@ -44,10 +44,6 @@ except ImportError:
     PANDAS_INSTALLED = False
 
 
-# cuDF
-CUDF_INSTALLED = importlib.util.find_spec("cudf") is not None
-
-
 # sklearn
 try:
     from sklearn.base import BaseEstimator as XGBModelBase
@@ -75,6 +71,18 @@ except ImportError:
 
     XGBKFold = None
     XGBStratifiedKFold = None
+
+
+def is_cudf_installed():
+    """Check cuDF installed or not"""
+    # Checking by `importing` instead of check `importlib.util.find_spec("cudf") is not None`
+    # because user might install cudf successfully but importing cudf raises issues (e.g. saying
+    # running on mismatched cuda version)
+    try:
+        import cudf
+        return True
+    except ImportError:
+        return False
 
 
 class XGBoostLabelEncoder(LabelEncoder):
--- a/python-package/xgboost/compat.py
+++ b/python-package/xgboost/compat.py
@@ -73,7 +73,7 @@ except ImportError:
     XGBStratifiedKFold = None
 
 
-logger = logging.getLogger(__name__)
+_logger = logging.getLogger(__name__)
 
 
 def is_cudf_installed() -> bool:
--- a/python-package/xgboost/compat.py
+++ b/python-package/xgboost/compat.py
@@ -73,7 +73,10 @@ except ImportError:
     XGBStratifiedKFold = None
 
 
-def is_cudf_installed():
+logger = logging.getLogger(__name__)
+
+
+def is_cudf_installed() -> bool:
     """Check cuDF installed or not"""
     # Checking by `importing` instead of check `importlib.util.find_spec("cudf") is not None`
     # because user might install cudf successfully but importing cudf raises issues (e.g. saying
@@ -82,6 +85,7 @@ def is_cudf_installed():
         import cudf
         return True
     except ImportError:
+        _logger.exception("Importing cuDF failed, use DMatrix instead of QDM")
         return False
 
 
--- a/python-package/xgboost/compat.py
+++ b/python-package/xgboost/compat.py
@@ -76,13 +76,13 @@ except ImportError:
 _logger = logging.getLogger(__name__)
 
 
-def is_cudf_installed() -> bool:
-    """Check cuDF installed or not"""
-    # Checking by `importing` instead of check `importlib.util.find_spec("cudf") is not None`
-    # because user might install cudf successfully but importing cudf raises issues (e.g. saying
-    # running on mismatched cuda version)
+def is_cudf_available() -> bool:
+    """Check cuDF package available or not"""
+    if importlib.util.find_spec("cudf") is None:
+        return False
     try:
         import cudf
+
         return True
     except ImportError:
         _logger.exception("Importing cuDF failed, use DMatrix instead of QDM")
--- a/python-package/xgboost/compat.py
+++ b/python-package/xgboost/compat.py
@@ -43,6 +43,11 @@ except ImportError:
     pandas_concat = None
     PANDAS_INSTALLED = False
 
+
+# cuDF
+CUDF_INSTALLED = importlib.util.find_spec("cudf") is not None
+
+
 # sklearn
 try:
     from sklearn.base import BaseEstimator as XGBModelBase
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -37,7 +37,7 @@ from xgboost.training import train as worker_train
 
 import xgboost
 from xgboost import XGBClassifier, XGBRanker, XGBRegressor
-from xgboost.compat import CUDF_INSTALLED
+from xgboost.compat import is_cudf_installed
 
 from .data import (
     _read_csr_matrix_from_unwrapped_spark_vec,
@@ -57,7 +57,6 @@ from .params import (
     HasEnableSparseDataOptim,
     HasFeaturesCols,
     HasQueryIdCol,
-    UseQuantileDMatrix,
 )
 from .utils import (
     CommunicatorContext,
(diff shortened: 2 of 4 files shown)


WeichenXu123:
CC @trivialfis @wbo4958 @hcho3

python-package/xgboost/spark/core.py
```diff
@@ -755,7 +758,12 @@ def _fit(self, dataset):
             k: v for k, v in train_call_kwargs_params.items() if v is not None
         }
         dmatrix_kwargs = {k: v for k, v in dmatrix_kwargs.items() if v is not None}
-        use_qdm = booster_params.get("tree_method", None) in ("hist", "gpu_hist")
+
+        if self.isDefined(self.use_quantile_dmatrix):
+            use_qdm = self.getOrDefault(self.use_quantile_dmatrix)
+        else:
+            use_qdm = CUDF_INSTALLED and \
```
wbo4958, line 765: do we need to check CUDF in driver side?

WeichenXu123: we can check CUDF in executor side, then different worker might have different use_qdm config if they have different python env.

python-package/xgboost/compat.py
```diff
@@ -43,6 +43,11 @@ def lazy_isinstance(instance: Any, module: str, name: str) -> bool:
     pandas_concat = None
     PANDAS_INSTALLED = False
 
+
+# cuDF
+CUDF_INSTALLED = importlib.util.find_spec("cudf") is not None
```
trivialfis, line 48: `importlib` is not a python standard library.

python-package/xgboost/spark/core.py
trivialfis, line 765: Can we assume all workers have the same python env? Otherwise, we have a rabbit hole.

trivialfis (commented):
I don't think an independent parameter is needed. XGBoost should be able to decide what data structure should be used based on input.

python-package/xgboost/compat.py
hcho3: It is part of CPython? https://docs.python.org/3/library/importlib.html

python-package/xgboost/spark/core.py
WeichenXu123: Then which logic shall we use ?

Only when all workers has CUDF installed, then enable `use_qdm` ?

WeichenXu123: @trivialfis I think independent parameter is added for the case: different worker has different python environment .

wbo4958: please don't check cudf in the driver side. Even driver side doesn't have cudf, workers still can have cudf installed and use the QDM.

WeichenXu123: @wbo4958 Suppose some worker has cudf and some does not, If some worker use QDM, some do not, what will happen ?

python-package/xgboost/compat.py
trivialfis: Interesting, thank you for the reference.

python-package/xgboost/spark/core.py
trivialfis: let's continue the discussion in https://github.com/dmlc/xgboost/pull/8469#pullrequestreview-1183596175

WeichenXu123: Replied https://github.com/dmlc/xgboost/pull/8469#issuecomment-1319468447

WeichenXu123:
@trivialfis PR updated ! :)

python-package/xgboost/spark/core.py
```diff
@@ -56,6 +57,7 @@
     HasEnableSparseDataOptim,
     HasFeaturesCols,
     HasQueryIdCol,
+    UseQuantileDMatrix,
```
wbo4958, line 60: seems this is not needed any more?

python-package/xgboost/spark/core.py
```diff
@@ -755,7 +757,11 @@ def _fit(self, dataset):
             k: v for k, v in train_call_kwargs_params.items() if v is not None
         }
         dmatrix_kwargs = {k: v for k, v in dmatrix_kwargs.items() if v is not None}
-        use_qdm = booster_params.get("tree_method", None) in ("hist", "gpu_hist")
+
+        # If cuDF is not installed, then using DMatrix instead of QDM,
+        # because without cuDF, DMatrix performs better than QDM.
+        use_qdm = CUDF_INSTALLED and \
```
wbo4958, line 763: I'm still confused why do we still need to check the cudf on the driver side? If we have packed all the xgboost dependencies including CUDF into a zip and specify it by `--py-files` when spark-submitting. Even if the driver doesn't have cudf installed, the training task can still use qdm.

WeichenXu123: OK I can move the check to worker side. :)

WeichenXu123: @wbo4958 @trivialfis Updated!

python-package/xgboost/compat.py
```diff
@@ -72,6 +73,18 @@ def lazy_isinstance(instance: Any, module: str, name: str) -> bool:
     XGBStratifiedKFold = None
 
 
+def is_cudf_installed():
+    """Check cuDF installed or not"""
+    # Checking by `importing` instead of check `importlib.util.find_spec("cudf") is not None`
```
trivialfis, line 78: I'm not sure if it's a good idea to hide an important error.

python-package/xgboost/spark/data.py
```diff
@@ -5,7 +5,7 @@
 import numpy as np
 import pandas as pd
 from scipy.sparse import csr_matrix
-from xgboost.compat import concat
+from xgboost.compat import CUDF_INSTALLED, concat
```
trivialfis, line 8: Where's this constant defined?

trivialfis (changes requested):

python-package/xgboost/compat.py
WeichenXu123: what about print a warning here ?

trivialfis: Sounds good.

python-package/xgboost/spark/data.py
WeichenXu123: removed.

python-package/xgboost/spark/core.py
```diff
@@ -775,6 +771,12 @@ def _train_booster(pandas_df_iter):
 
             gpu_id = None
 
+            # If cuDF is not installed, then using DMatrix instead of QDM,
+            # because without cuDF, DMatrix performs better than QDM.
+            # Note: Checking `is_cudf_installed` in spark worker side because
+            # spark worker might has different python environment with driver side.
+            use_qdm = use_hist and is_cudf_installed()
```
trivialfis, line 778: `use_hist` includes both cpu hist and gpu hist.

trivialfis: Please fix the linter error as well.

WeichenXu123: > use_hist includes both cpu hist and gpu hist.

Yes. Any issues on this ?

trivialfis: CPU hist should continue to use qdm with or without cudf. cuDF is a GPU only library.

WeichenXu123: ok got it.

WeichenXu123: PR updated!

trivialfis (approved):

python-package/xgboost/compat.py
```diff
@@ -72,6 +73,22 @@ def lazy_isinstance(instance: Any, module: str, name: str) -> bool:
     XGBStratifiedKFold = None
 
 
+_logger = logging.getLogger(__name__)
+
+
+def is_cudf_available() -> bool:
+    """Check cuDF package available or not"""
+    if importlib.util.find_spec("cudf") is None:
+        return False
```
WeichenXu123, line 82: This is for avoiding printing importing error if `cuDF` not installed. (specifically for databricks runtime)

python-package/xgboost/compat.py
```diff
@@ -72,6 +73,22 @@ def lazy_isinstance(instance: Any, module: str, name: str) -> bool:
...
+    try:
+        import cudf
+
+        return True
+    except ImportError:
+        _logger.exception("Importing cuDF failed, use DMatrix instead of QDM")
```
trivialfis, line 88: If an exception is raised, I don't think we can continue to use DMatrix instead of QDM right? Have you tested the code to verify that this statement holds true?

WeichenXu123: uh ?

We can simply test:
```
>>> def f1():
...     try:
...             raise ValueError()
...             return True
...     except ValueError:
...             return False
... 
>>> 
>>> f1()
False
```

trivialfis: I meant `        _logger.exception("Importing cuDF failed, use DMatrix instead of QDM")`, This will reraise the exception right?

WeichenXu123: @trivialfis No. It only log message and exception stack, does not reraise exception.  See test:

```
>>> import logging
>>> 
>>> def f1():
...     try:
...         raise ValueError()
...         return True
...     except ValueError:
...         logging.exception("log exception.")
...         return False
... 
>>> print(f"f1 return value: {f1()}")
ERROR:root:log exception.
  File "<stdin>", line 3, in f1
ValueError
f1 return value: False
```
The `f1 return value: False` line is printed.

trivialfis (approved):

python-package/xgboost/compat.py
trivialfis: Got it, thank you for the explanation.

Issue #7642, opened 2022-02-09
Performance of converting datatable frames to DMatrix


oleksiyskononenko:
Dear developers, I'm now looking into a performance issue when converting datatable frames to XGBoost DMatrix. Here is a simple python script that I wrote to benchmark datatable->XGBoost vs numpy- XGBoost performance:

```python
import datatable as dt
import xgboost as xgb
import numpy as np
import time

nrows = 10**8
ncols = 1
DATA = [range(nrows)] * ncols
print("Shape: (%d; %d)" % (nrows, ncols))

NP = np.array(DATA)
t0 = time.time()
XGB = xgb.DMatrix(NP)
tnp = time.time() - t0

DT = dt.Frame(np.transpose(NP))
t0 = time.time()
XGB = xgb.DMatrix(DT)
tdt = time.time() - t0

print("t(DT->XG)/t(NP->XG): ", tdt/tnp)
```

I can see that conversion of one-column numpy array takes somewhat less time than the conversion of a one-column datatable frame
```
Shape: (100000000; 1)
t(DT->XG):  3.133333921432495
t(NP->XG):  2.115936040878296
t(DT->XG)/t(NP->XG):  1.480826386478058
```

when I keep the number of elements the same, but go for 10 columns dt->XGBoost conversion becomes even slower
```
Shape: (10000000; 10)
t(DT->XG):  2.4032959938049316
t(NP->XG):  1.1607439517974854
t(DT->XG)/t(NP->XG):  2.0704790148449845
```
when I go for 100 columns, still keeping the number of elements the same, it becomes even worse
```
Shape: (1000000; 100)
t(DT->XG):  3.0988237857818604
t(NP->XG):  0.9062139987945557
t(DT->XG)/t(NP->XG):  3.41952760595611
```

For some other multicolumn data I found that `t(DT->XG)/t(NP->XG)` ratio could be as bad as 20. For some smaller data it could be around 1.

Do you have any ideas as to why performance of dt->XGBoost could be so bad comparing to numpy? First, I thought it could be related to the number of threads used. But it seems that the conversion is single-threaded for both np->XGBoost and dt->XGBoost, no matter how many threads I ask for in `xgb.DMatrix()`. Thanks in advance for any help on this issue.

P.S. For the benchmarks above I'm using the latest packages from pypi repo: XGBoost `1.5.2`,  numpy `1.22.2` and datatable `1.0.0`.

trivialfis:
[...]

PR #8472, opened 2022-11-18
Take datatable as row major input.

trivialfis:
Try to avoid a transform with a dense table.

Close https://github.com/dmlc/xgboost/issues/7642

Using 1e6x16:
MA: t(DT->XG)/t(NP->XG):  2.3788813664780304
PR: t(DT->XG)/t(NP->XG):  0.8311551647390493

### DIFF
--- a/python-package/xgboost/data.py
+++ b/python-package/xgboost/data.py
@@ -460,12 +460,9 @@ def _from_pandas_series(
 
 
 def _is_dt_df(data: DataType) -> bool:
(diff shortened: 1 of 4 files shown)


trivialfis:
I'm trying to bring the CSC API up to date either by eliminating it and relying on CSR API or by building a new API similar to CSR.

This PR removes the need to use column-major format for datatable, which might ease the process if we were to remove direct support for CSC.

RAMitchell (approved):

PR #8478, opened 2022-11-22
[jvm-packages] fix rabit params not passed though in xgboost4j-spark/XGBoost.scala

ebullientfly:
1. fix rabit params not passed though in xgboost4j-spark/XGBoost.scala
2. use a unified way to set and get rabit params


trivialfis:
cc @wbo4958 .

wbo4958:
Sry for the late response, 

> fix rabit params not passed though in xgboost4j-spark/XGBoost.scala

@ebullientfly, I just read the PR which seems like a re-work on how to start Rabit Tracker and pass the Rabit envs. It's not a "fix", right?

PR #8486, opened 2022-11-25
Support null value in CUDA array interface.

trivialfis:
- Fix for potential null value in array interface (the python namespace, not the actual data).
- Fix incorrect check on mask stride.

The error is found when running cuDF tests with nightly cuDF (22.12) where the `strides` for the array interface is set to `NONE`.


trivialfis:
Will add some tests to ensure we have some coverage even without the latest cuDF.

wphicks (approved):

hcho3 (approved):

Issue #8485, opened 2022-11-25
Support half type from cupy.


trivialfis:
On compute 6.0 and greater, CUDA supports float16 and we can use it with array interface.

PR #8487, opened 2022-11-25
Support half type from cupy.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/8485 .


trivialfis:
Latest GCC and clang have f16 support on x86 as well. But we need special macro and compiler flag for them, and we won't be able to use those versions on our CI in recent future so it's not implemented yet.

hcho3 (approved):

PR #8494, opened 2022-11-29
Add an in-memory collective communicator

rongou:
This is mainly useful for testing so that we don't have to spin up a rabit tracker or a gRPC server.

Also extracted the core multi-threaded logic into a handler, and refactored the federated gRPC server to use it.

Part of #8424


rongou:
@trivialfis @hcho3

trivialfis (approved):

PR #8496, opened 2022-11-29
[pyspark] check use_qdm across all the workers

wbo4958:
This PR is a followup of https://github.com/dmlc/xgboost/pull/8471 which can't resolve the issue that some workers have cudf while some others don't.

This PR gathers use_qdm infos from all workers and does the cudf-available check.


wbo4958:
@WeichenXu123 @trivialfis please help to review it

python-package/xgboost/spark/core.py
```diff
@@ -780,14 +781,21 @@ def _train_booster(pandas_df_iter):
...
+                _rabit_args = _get_rabit_args(context, num_workers)
+
+            worker_message = {
+                "rabit_msg": _rabit_args,
+                "use_qdm": use_qdm,
+            }
+
+            messages = context.allGather(message=json.dumps(worker_message))
+
+            use_qdm = all(json.loads(x)["use_qdm"] for x in messages)
```
trivialfis, line 795: All workers should be consistent. Let's use a check instead of a workaround.

wbo4958: We can't assume the user's env, especially for standalone mode. users need to install the xgboost dep in every worker manually. And this check is really cheap since the message is attached to the original allGather API getting rabit message.  but it can really fix the "inconsistent" issue

wbo4958: And this PR also removes the first barrier() which is not needed, since allGather achieves the same goal.

trivialfis: We would like the users to have a consistent environment.

> users need to install the xgboost dep in every worker manually

I'm sure a user can automate the process.

My suggestion is not to remove the allgather, but instead to **assert** that all workers have the same environment.

wbo4958: what do you mean "not to remove the allgather"?

@WeichenXu123 what's your opinion about *but instead to assert that all workers have the same environment. If it doesn't, we will throw an exception and terminate the job.*

wbo4958: @WeichenXu123, how about throwing an exception when detecting in-consistent env?

wbo4958: The code is simple and won't introduce any side-effects including extra overhead.

WeichenXu123: @wbo4958 Sounds good!

wbo4958:
@hcho3 @trivialfis please help to trigger the CI

wbo4958:
@hcho3 @trivialfis please help to trigger the CI. Thx

wbo4958:
@WeichenXu123 @trivialfis please help to review it again. Thx

WeichenXu123 (approved):

trivialfis (approved):

PR #8497, opened 2022-11-30
[pyspark] sort qid for SparkRanker

### LINKED ISSUE #8491: [pyspark] SparkXGBRanker does not work on dataframe with multiple partitions
I have experienced the same issue mentioned in this discussion forum https://discuss.xgboost.ai/t/sparkxgbranker-does-not-work-on-parallel-workers/2986, and did not see any issue raised in Github, so I just do it here

"SparkXGBRanker from 1.7.0 release requires data to be sorted by qid. It works fine if we have one worker and sorted dataframe. However with multiple workers data comes to them unordered and raises exception:

org.apache.spark.api.python.PythonException: 'xgboost.core.XGBoostError: [17:46:40] …/src/data/data.cc:486: Check failed: non_dec: qid must be sorted in non-decreasing order along with data. 

... 

Is there any ways to prepare df in sorted order for workers? Or sorting should be done on each worker?"

One trivial solution is that given a df with ['qid', 'label', 'features'], 
df = df.repartition(1)
df = df.sort(df.qid.asc())

repartition into 1 and do the sorting can make the SparkXGBRanker run without any error as far as I explore, but such expensive operation really slow down the total processing time, could you add support for such case? Thanks in advance!

wbo4958:
To fix #8491

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -1,7 +1,7 @@
 # type: ignore
 """Xgboost pyspark integration submodule for core code."""
 # pylint: disable=fixme, too-many-ancestors, protected-access, no-member, invalid-name
-# pylint: disable=too-few-public-methods, too-many-lines
+# pylint: disable=too-few-public-methods, too-many-lines, too-many-branches
 import json
 from typing import Iterator, Optional, Tuple
 
@@ -728,6 +728,10 @@ class _SparkXGBEstimator(Estimator, _SparkXGBParams, MLReadable, MLWritable):
                 dataset = dataset.repartition(num_workers, rand(1))
             else:
                 dataset = dataset.repartition(num_workers)
+
+        if self.isDefined(self.qid_col) and self.getOrDefault(self.qid_col):
+            # XGBoost requires qid to be sorted for each partition
+            dataset = dataset.sortWithinPartitions(alias.qid)
 
         train_params = self._get_distributed_train_params(dataset)
         booster_params, train_call_kwargs_params = self._get_xgb_train_call_args(
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -731,7 +731,7 @@ class _SparkXGBEstimator(Estimator, _SparkXGBParams, MLReadable, MLWritable):
 
         if self.isDefined(self.qid_col) and self.getOrDefault(self.qid_col):
             # XGBoost requires qid to be sorted for each partition
-            dataset = dataset.sortWithinPartitions(alias.qid)
+            dataset = dataset.sortWithinPartitions(alias.qid, ascending=True)
 
         train_params = self._get_distributed_train_params(dataset)
         booster_params, train_call_kwargs_params = self._get_xgb_train_call_args(
--- a/tests/test_distributed/test_with_spark/test_spark_local.py
+++ b/tests/test_distributed/test_with_spark/test_spark_local.py
@@ -1134,25 +1134,8 @@ class XgboostRankerLocalTest(SparkTestCase):
                 (Vectors.dense(1.0, 2.0, 3.0), 0, 6),
                 (Vectors.dense(4.0, 5.0, 6.0), 1, 6),
                 (Vectors.dense(9.0, 4.0, 8.0), 2, 6),
-                (Vectors.sparse(3, {1: 1.0, 2: 5.5}), 0, 5),
-                (Vectors.sparse(3, {1: 6.0, 2: 7.5}), 1, 5),
-                (Vectors.sparse(3, {1: 8.0, 2: 9.5}), 2, 5),
-                (Vectors.dense(1.0, 2.0, 3.0), 0, 4),
-                (Vectors.dense(4.0, 5.0, 6.0), 1, 4),
-                (Vectors.dense(9.0, 4.0, 8.0), 2, 4),
-                (Vectors.sparse(3, {1: 1.0, 2: 5.5}), 0, 3),
-                (Vectors.sparse(3, {1: 6.0, 2: 7.5}), 1, 3),
-                (Vectors.sparse(3, {1: 8.0, 2: 9.5}), 2, 3),
-                (Vectors.dense(1.0, 2.0, 3.0), 0, 2),
-                (Vectors.dense(4.0, 5.0, 6.0), 1, 2),
-                (Vectors.dense(9.0, 4.0, 8.0), 2, 2),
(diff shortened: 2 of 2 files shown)


wbo4958:
@WeichenXu123 @trivialfis please help to review it.

hcho3 (approved):

python-package/xgboost/spark/core.py
```diff
@@ -729,6 +729,10 @@ def _fit(self, dataset):
             else:
                 dataset = dataset.repartition(num_workers)
 
+        if self.isDefined(self.qid_col) and self.getOrDefault(self.qid_col):
+            # XGBoost requires qid to be sorted for each partition
+            dataset = dataset.sortWithinPartitions(alias.qid)
```
WeichenXu123, line 734: Nit: add `ascending=True` explicitly.

tests/test_distributed/test_with_spark/test_spark_local.py
```diff
@@ -1124,3 +1093,80 @@ def test_early_stop_param_validation(self):
...
+                (Vectors.sparse(3, {1: 8.0, 2: 9.5}), 2, 3),
+                (Vectors.dense(1.0, 2.0, 3.0), 0, 2),
+                (Vectors.dense(4.0, 5.0, 6.0), 1, 2),
+                (Vectors.dense(9.0, 4.0, 8.0), 2, 2),
+                (Vectors.sparse(3, {1: 1.0, 2: 5.5}), 0, 1),
+                (Vectors.sparse(3, {1: 6.0, 2: 7.5}), 1, 1),
+                (Vectors.sparse(3, {1: 8.0, 2: 9.5}), 2, 1),
+                (Vectors.dense(1.0, 2.0, 3.0), 0, 0),
+                (Vectors.dense(4.0, 5.0, 6.0), 1, 0),
+                (Vectors.dense(9.0, 4.0, 8.0), 2, 0),
```
WeichenXu123, line 1154: nit: do we need hardcode so long data list ?
we can hardcode 4 rows and use `[ ... ] * 100` instead.

WeichenXu123 (approved):

tests/test_distributed/test_with_spark/test_spark_local.py
```diff
@@ -1124,3 +1093,80 @@ def test_early_stop_param_validation(self):
...
+        pred_result = model.transform(self.ranker_df_test).collect()
+
+        for row in pred_result:
+            assert np.isclose(row.prediction, row.expected_prediction, rtol=1e-3)
+
+    def test_ranker_qid_sorted(self):
+        ranker = SparkXGBRanker(qid_col="qid", num_workers=2)
+        assert ranker.getOrDefault(ranker.objective) == "rank:pairwise"
+        model = ranker.fit(self.ranker_df_train_1)
+        model.transform(self.ranker_df_test).collect()
```
trivialfis, line 1172: what's the purpose of this test?

tests/test_distributed/test_with_spark/test_spark_local.py
wbo4958: Done.

python-package/xgboost/spark/core.py
wbo4958: Done

tests/test_distributed/test_with_spark/test_spark_local.py
wbo4958: to test if the SparkRanker will throw exception

tests/test_distributed/test_with_spark/test_spark_local.py
```diff
@@ -1124,3 +1093,80 @@ def test_early_stop_param_validation(self):
...
+                (Vectors.dense(9.0, 4.5, 8.0), 0, 2.36570),
+                (Vectors.sparse(3, {1: 1.0, 2: 6.0}), 1, -1.87988),
+                (Vectors.sparse(3, {1: 6.0, 2: 7.0}), 1, -0.30612),
+                (Vectors.sparse(3, {1: 8.0, 2: 10.5}), 1, 2.44826),
+            ],
+            ["features", "qid", "expected_prediction"],
+        )
+        self.ranker_df_train_1 = self.session.createDataFrame(
+            [
+                (Vectors.sparse(3, {1: 1.0, 2: 5.5}), 0, 9),
```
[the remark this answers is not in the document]
wbo4958: yeah, the qid is the descending order. without the fix, it will throw exception `../src/data/data.cc:486: Check failed: non_dec: qid must be sorted in non-decreasing order along with data.`

WeichenXu123 (approved):

wbo4958:
@hcho3 please help to merge it. thx

trivialfis:
@wbo4958 Could you please change the tests to NOT use hardcoded results?

trivialfis:
https://github.com/dmlc/xgboost/pull/8497#discussion_r1035757638

wbo4958:
> [#8497 (comment)](https://github.com/dmlc/xgboost/pull/8497#discussion_r1035757638)

Hi @trivialfis, For this case, the test I added is to check if the pyspark application will be crashed, so it's ok, I think, to hardcode the data. Since I think the data is so straightforward to show the scenario which can crash the process.

tests/test_distributed/test_with_spark/test_spark_local.py
```diff
@@ -1124,3 +1093,63 @@ def test_early_stop_param_validation(self):
...
+        )
+
+    def test_ranker(self):
+        ranker = SparkXGBRanker(qid_col="qid")
+        assert ranker.getOrDefault(ranker.objective) == "rank:pairwise"
+        model = ranker.fit(self.ranker_df_train)
+        pred_result = model.transform(self.ranker_df_test).collect()
+
+        for row in pred_result:
+            assert np.isclose(row.prediction, row.expected_prediction, rtol=1e-3)
```
trivialfis, line 1149: @wbo4958 This is not only checking exception.

wbo4958: This test is moved from https://github.com/dmlc/xgboost/pull/8497/files#diff-3b3ca1f9bd10767b61c3eab170a027b67408881dcf57e4e992c2caa47d660ff5L386-L407, I didn't change it

trivialfis: Ah ... That's a headache, I'm blocked by these tests and don't know how to recreate them...

wbo4958: yes, we can have the following PR to refactor these tests by not hardcoding them

PR #8501, opened 2022-12-01
[jvm-packages] Set up dependabot

hcho3:
This may help us keep dependencies up to date.

PR #8507, opened 2022-12-01
[jvm-packages] Configure dependabot properly

hcho3:
Following up with #8501

* Specify the group ID in dependency specification.
* Update all `pom.xml` files.
* Pin Spark and Scala to 3.0 and 2.12, respectively. Allow automatic updates for patch releases.

cc @trivialfis @wbo4958 @WeichenXu123 We are setting up a bot to automatically file pull requests to update dependencies of the JVM packages.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,60 +10,84 @@ updates:
     schedule:
       interval: "daily"
     ignore:
-      # Pin Scala version
+      # Pin Scala version to 2.12.x
       - dependency-name: "org.scala-lang:scala-compiler"
+        versions: [">= 2.13.0"]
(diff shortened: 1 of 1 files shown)


trivialfis:
cc @wbo4958 I  not familiar with the ecosystem there.

hcho3:
Note: This PR assumes the status quo, where XGBoost4J and XGBoost4J-Spark depends on Spark 3.0 and Scala 2.12. In the future, we should find a way to support the latest Spark (3.3) and Scala (2.13).

wbo4958:
Yeah, in order to support 2.13, we need to re-work some code.

PR #8524, opened 2022-12-03
[jvm-packages] Bump version of akka packages

hcho3:


eejbyfeldt:
Was the license https://www.lightbend.com/blog/why-we-are-changing-the-license-for-akka in akka >= 2.7.0 considered when doing this bump? Since the new license for akka is not open source this is might be problematic for some users of this library.

PR #8540, opened 2022-12-05
feat(py, plot_importance): + values_format as arg

gianfa:
Adds the ability to format the value printed near the bars on the `plot_importance` plot, directly from function arguments.

`plot_importance` is a very convenient graph, but the values it prints are too noisy, being full precision.  
This forces you to have to construct your own clean plot yourself from scratch.  
With this improvement you can continue to use `plot_importance` and have more control over the values printed on the bars.


trivialfis (approved):

trivialfis:
Thank you for the new feature!

Issue #8012, opened 2022-06-20
Address all lintr warnings.


trivialfis:
https://github.com/dmlc/xgboost/runs/6965009976?check_suite_focus=true

jameslamb:
👋 Could I work on this one?

I originally set up LightGBM's `{lintr}` stuff (https://github.com/microsoft/LightGBM/pull/2437) and recently updated it for `{lintr}` v3.0.0 (https://github.com/microsoft/LightGBM/pull/5294), so I'm pretty familiar it.

trivialfis:
Thank you for the offer! Actually, I would love to hear your opinion on https://github.com/dmlc/xgboost/issues/7906 and https://github.com/microsoft/LightGBM/pull/4968 before spending a non-trivial amount of time on the R package. I started to learn more about R and find some of the issues in the existing codebase of various packages including xgb and lgb. I saw https://github.com/microsoft/LightGBM/pull/4968 , which is a great improvement for R users I believe. But it's using R6 and might bring significant breaking change.

trivialfis:
Forgot to ping @jameslamb  ;-)

jameslamb:
Sorry for the delay in responding!

<hr>

> before spending a non-trivial amount of time on the R package

I think getting `{lintr}` linting running in CI again is worth doing regardless of any future design work. In my experience, most of the fixes for `{lintr}` warnings are quick, easy, and don't impact the public API of the library. It's also great to have prior to doing new development because `{lintr}` gives you some free minimal coverage of code that doesn't yet have tests.

<hr>

> to hear your opinion on https://github.com/dmlc/xgboost/issues/7906 and https://github.com/microsoft/LightGBM/pull/4968 

**Short answer**

I agree with @RAMitchell (https://github.com/dmlc/xgboost/issues/7906#issuecomment-1127024958). XGBoost should take on this list of proposed changes in the form of new public APIs, and then consider a lengthy deprecation cycle.

**Long answer**

In LightGBM, we've been accepting PRs with some of the breaking changes proposed in https://github.com/microsoft/LightGBM/pull/4968 (I asked its author to break them into smaller pieces, so we could do higher-quality reviews and merge things gradually). The only reason that work isn't complete is because of a severe lack of contributor/maintainer availability in LightGBM, which has meant not much time left for things other than fixing CI, improving packaging details, and supporting new versions of dependencies.

[...]

PR #8545, opened 2022-12-05
[R] [ci] move linting code out of package

jameslamb:
Contributes to #8012.

This PR proposes the following changes to the way this project uses `{lintr}` to lint its R code.

* moves the `run_lint.R` out of the package and up into `tests/ci_build/`
    - *this script is only used during development, so it doesn't need to be included in the package distribution*
    - *doing this reduces the size of the built package on CRAN, and removes two `Suggests` dependencies (`{lintr}` and `{crayon}`), reducing the effort required for those other projects to test new releases*
* modifies the linting code so that `.Rmd` files are also checked
* completely removes use of `{crayon}` in the linting code
    - *this is just used to color terminal output, which I think is not worth the risk of CI builds that could be caused by failure to install that library + the added CI time to download and install it*

### How I tested this

```shell
Rscript ./tests/ci_build/lint_r.R "$(pwd)"
```

This shows 1285 linting errors, which II'm planning to propose fixes for in follow-up PRs.

### Notes for Reviewers


R-package/tests/helper_scripts/install_deps.R
```diff
@@ -15,10 +15,8 @@ pkgs <- c(
   "Ckmeans.1d.dp",
   "vcd",
   "testthat",
-  "lintr",
```
trivialfis, line 18: Is it no longer necessary to install lintr?

trivialfis (approved):
Thank you for working on this! Question inlined in comments.

Also, apologies for the failing CI. GitHub action seems to be breaking down ...

R-package/tests/helper_scripts/install_deps.R
jameslamb: ah sorry! that was a mistake, I've added this back in in https://github.com/dmlc/xgboost/pull/8545/commits/b392fbbd464fc109fd08e069155acadc8768d248

PR #1142, opened 2016-04-24
Enable flake8 for Python

sinhrks:
Enable ``flake8`` in addition to ``lint``, which performs static analysis of Python code. 

I could find missing import of ``pickle`` and ``XGBoostError``, and fixed in this PR.

PR #8546, opened 2022-12-05
[python] remove unnecessary conversions between data structures

jameslamb:
I ran [`flake8`](https://flake8.pycqa.org/en/latest/) over the code in the Python package, with [`flake8-comprehensions`](https://pypi.org/project/flake8-comprehensions/) installed.

```shell
pip install flake8 flake8-comprehensions
flake8 \
    --ignore=B,F,E,W \
    --exclude='python-package/build' \
    .
```

This found a few opportunities for small optimizations. For example, the results of `set([1, 2, 3])` and `{1, 2, 3}` are equivalent, but the `set([])` form has the extra cost of construction an intermediate list (which is immediately thrown away).

This PR proposes resolving those warnings, for some tiny efficiency improvements.

```text
./python-package/setup.py:282:51: C400 Unnecessary generator - rewrite as a list comprehension.
./python-package/xgboost/core.py:1163:48: C405 Unnecessary tuple literal - rewrite as a set literal.
./python-package/xgboost/core.py:1511:18: C402 Unnecessary generator - rewrite as a dict comprehension.
./python-package/xgboost/tracker.py:144:23: C405 Unnecessary list literal - rewrite as a set literal.
./python-package/xgboost/tracker.py:258:23: C405 Unnecessary list literal - rewrite as a set literal.
./python-package/xgboost/training.py:518:18: C402 Unnecessary generator - rewrite as a dict comprehension.
```


python-package/xgboost/training.py
```diff
@@ -514,8 +514,6 @@ def cv(
         params = dict(params)
         if 'eval_metric' in params:
             params['eval_metric'] = _metrics
-    else:
-        params = dict((k, v) for k, v in params.items())
```
jameslamb, line 518: `flake8-comprehensions` warned about the use of `dict()` here, but when I looked more closely, it seems like this line isn't necessary at all.

It's been in `xgboost` for 6+ years (since #1142), so maybe something has changed since then.

But if the code enters this point, it's already guaranteed that `params` is a dictionary.

https://github.com/dmlc/xgboost/blob/67752e39679ec3a5a9e3a4a4b93967931e4b7d62/python-package/xgboost/training.py#L512-L518

And for something that is already a dictionary, I can't think of a situation where re-iterating over all the contents like this would produce a different dictionary.

```shell
params = {"a": 1, "b": [2, 3], "c": None}

new_params = dict((k, v) for k, v in params.items())

params == new_params
# True
```

trivialfis: Thank you for catching that!

python-package/xgboost/core.py
```diff
@@ -1160,7 +1160,7 @@ def feature_names(self, feature_names: Optional[FeatureNames]) -> None:
                 raise ValueError(msg)
             # prohibit to use symbols may affect to parse. e.g. []<
             if not all(isinstance(f, str) and
-                       not any(x in f for x in set(('[', ']', '<')))
+                       not any(x in f for x in {'[', ']', '<'})
```
jameslamb, line 1163:
```suggestion
                       not any(x in f for x in ['[', ']', '<'])
```

`pylint` rightly caught that since this is iterator, not doing a set comparison, an iterable type should be used here instead of a set.

>  /home/runner/work/xgboost/xgboost/python-package/xgboost/core.py:1163: convention (C0208, use-sequence-for-iteration, DMatrix.feature_names) Use a sequence type when iterating over values

build link: https://github.com/dmlc/xgboost/actions/runs/3635213908/jobs/6134104007

trivialfis: hmm, the change seems to be causing failure in tests.

jameslamb: Interesting, I'm not sure why. I can't see how iterating over a list of strings should be different (in a way that matters for this code) than iterating over a set of strings.

I just merged latest `master` into this branch, in case the test failures aren't related to this PR's changes. If they still fail, I'll try to reproduce this locally.

jameslamb: Ok yes, I was able to reproduce the failing tests locally.

```shell
cd python-package
pip install .
cd ..
pytest tests/python/test_early_stopping.py
```

I'll try to figure out which of these changes is to blame, sorry about that!

python-package/xgboost/training.py
jameslamb: @trivialfis I found that this change was actually to blame for the failure of `tests/python/test_early_stopping.py`, not the one from this comment thread: https://github.com/dmlc/xgboost/pull/8546#discussion_r1042219019.

That `dict((k, v) ...` code was serving a second purpose...it ensured that the rest of `xgboost.cv()` worked on *a copy* of `params`, so calls like `params.pop()` wouldn't have side-effects on the passed-in dictionary.

I fixed that in https://github.com/dmlc/xgboost/pull/8546/commits/849ea9221367fcb5ce8175befb523801b9f0d66a.

------

There is one remaining test Python test failure (`python/test_with_sklearn.py::test_pandas_input`) on macOS ([build link](https://github.com/dmlc/xgboost/actions/runs/3654519571/jobs/6175015371)) and one on Windows ([build link](https://github.com/dmlc/xgboost/actions/runs/3654519571/jobs/6175016596)).

```text
assert isinstance(
            clf_isotonic.calibrated_classifiers_[0].base_estimator, xgb.XGBClassifier
        )
E       AttributeError: '_CalibratedClassifier' object has no attribute 'base_estimator'
```

I think that might not be related to this PR's changes, given that I also see it on this other PR which only touches the R package: #8565 ([build link](https://github.com/dmlc/xgboost/actions/runs/3652941683/jobs/6171873398)).

I wonder if that's related to the newest release of `scikit-learn` supporting pandas as an input and output (https://github.com/scikit-learn/scikit-learn/issues/5523). 

I noticed, for example, that the failing Mac build here is using `scikit-learn=1.2.0`, which was just released today ([PyPI link](https://pypi.org/project/scikit-learn/#history)).

trivialfis:
let me fix the failure with updated sklearn tomorrow.

python-package/xgboost/training.py
jameslamb: Hopefully #8580 will fix this.

trivialfis:
apologies for the push. I need to bring in the CI fixes and hope you don't mind.

jameslamb:
no problem, do what you need! And let me know if I can help in any way.

PR #8547, opened 2022-12-05
Support `Series` and Python primitives in `inplace_predict` and QDM

trivialfis:
- Support `pd.Series` in inplace predict.
- Support list and tuple in inplace predict.
- Support `pd.Series` in QDM.

### DIFF
--- a/tests/python/test_predict.py
+++ b/tests/python/test_predict.py
@@ -5,7 +5,7 @@ import numpy as np
 import pandas as pd
 import pytest
 from scipy import sparse
-from xgboost.testing.data import np_dtypes
+from xgboost.testing.data import np_dtypes, pd_dtypes
 from xgboost.testing.shared import validate_leaf_output
 
 import xgboost as xgb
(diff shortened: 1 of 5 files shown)


tests/python/test_predict.py
```diff
@@ -231,6 +231,7 @@ def test_base_margin(self):
         from_dmatrix = booster.predict(dtrain)
         np.testing.assert_allclose(from_dmatrix, from_inplace)
 
+    @pytest.mark.skipif(**tm.no_pandas())
```
hcho3, line 234: Is this necessary?

trivialfis: Yes. The test also uses pandas in `np_dtypes` for testing differences between np.dtype and pandas types. For instance, what happen if np.bool is passed in? Is it nullable in a pandas dataframe?

hcho3: Still not sure. Neither `np_dtypes` nor `test_dtypes` has any reference to Pandas?

trivialfis: https://github.com/dmlc/xgboost/blob/0c38ca7f6e329dbb062939dcb843d0dc229675b4/python-package/xgboost/testing/data.py#L46

hcho3 (approved):

PR #8558, opened 2022-12-06
Properly await async method client.wait_for_workers

mrocklin:
Previously this async method was called but never awaited, resulting in a confusing message:

```
RuntimeWarning: coroutine 'Client._wait_for_workers' was never awaited
  client.wait_for_workers(n_workers)
```

Now we await it.

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -853,7 +853,7 @@ async def _get_rabit_args(
         sched_addr = None
 
     # make sure all workers are online so that we can obtain reliable scheduler_info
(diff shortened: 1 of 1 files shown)


trivialfis (approved):
Thank you for the fix! Will port it to 1.7.2.

trivialfis:
@mrocklin Hi, I will push a `type: ignore` to workaround the mypy check for now. But I think the annotation for `wait_for_workers` is inaccurate and it should not return `None`.

mrocklin:
Thank you @trivialfis .  Please let me know if there is anything more that I should do here.

PR #8564, opened 2022-12-07
[R] resolve brace_linter warnings

jameslamb:
Contributes to #8012 .

Resolves the following `{lintr}` warnings.

> style: [brace_linter] Opening curly braces should never go on their own line and should always be followed by a new line.
> style: [brace_linter] There should be a space before an opening curly brace.
> style: [brace_linter] Closing curly-braces should always be on their own line, unless they are followed by an else.
> style: [brace_linter] Any function spanning multiple lines should use curly braces.


trivialfis (approved):
Thank you for the fixes!

PR #8565, opened 2022-12-07
[R] resolve line_length_linter warnings

jameslamb:
Contributes to #8012 .

This fixes warnings about long lines in the project's R code. It fixes 1013 of the 1285 warnings `{lintr}` is currently raising ([link to recent CI build I got that value from](https://github.com/dmlc/xgboost/actions/runs/3635197925/jobs/6134073268)).

I chose an arbitrary value (150 characters) that could be achieved without needing to touch too many lines. I'm proposing that this be accepted to at least prevent any longer lines from making it into the project's R code.

Reducing this ceiling to, say, 125 or 100 characters would involve touching a LOT more of the project's R code, and I think that should be done by maintainers if you want to enforce a stricter limit.

### How I tested this

Ran the following to re-generate the docs and double-check that adding `{}` and changing indentation in roxygen comments wouldn't actually change their formatting.

```shell
cd R-package
R CMD INSTALL --with-keep.source .
Rscript -e "roxygen2::royxgenize(load = 'installed')"
Rscript -e "?xgboost::xgb.train"
```

### DIFF
--- a/R-package/DESCRIPTION
+++ b/R-package/DESCRIPTION
@@ -64,6 +64,6 @@ Imports:
     methods,
     data.table (>= 1.9.6),
     jsonlite (>= 1.0),
-RoxygenNote: 7.2.2
+RoxygenNote: 7.2.3
 Encoding: UTF-8
 SystemRequirements: GNU make, C++14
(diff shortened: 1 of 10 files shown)


trivialfis:
> I chose an arbitrary value (150 characters) that could be achieved without needing to touch too many lines.

I don't want to go into a debate on the subject of the "optimal number of characters a line should have" and I usually just stick to the linter's default. :-) But for the XGBoost R package, I think your choice of threshold makes sense and will improve the existing codebase.

trivialfis (approved):

PR #8576, opened 2022-12-08
Update NVFlare demo to work with the latest release

rongou:
A few small changes to conform to the latest NVFlare release.


rongou:
@trivialfis @hcho3

trivialfis (approved):
Thank you for the work! Is this run on the CI? If not, how difficult it's to run it?

rongou:
Probably not too hard, but @hcho3 would be a better judge.

trivialfis:
if we can have all dependencies from conda and pip, that should be quite easy.

PR #8579, opened 2022-12-09
Fix sklearn test that calls a removed field

rongou:
Seeing this error in the CI:
```console
E       AttributeError: '_CalibratedClassifier' object has no attribute 'base_estimator'
20852
```
Looks like it's been removed in the sklearn version.


rongou:
@trivialfis @hcho3

hcho3:
Thanks!

PR #8580, opened 2022-12-10
Make test_with_sklearn compatible with different versions of sklearn

rongou:
Looks like we may be running different versions of `sklearn` on github actions vs buildkite.

@trivialfis @hcho3


hcho3:
Given that this is strictly for testing, I'd prefer that we just update sklearn in BuildKite side

rongou:
Do you just need to rebuild the container?

hcho3:
Yes, a container rebuild should be sufficient

rongou:
Are you going to do it? I don't think I have access.

hcho3:
Yes

hcho3:
Fixed in #8583

Issue #8573, opened 2022-12-08
Valid categorical values incorrectly treated as missing in prediction


adamreeve:
Code to reproduce:
```python
import xgboost
import pandas as pd
import tempfile

model_json = """{
  "version": [1, 7, 1],
  "learner": {
    "feature_names": ["x0"],
    "feature_types": ["c"],
    "objective": {
      "name": "reg:squarederror",
      "reg_loss_param": {
        "scale_pos_weight": "1"
      }
    },
    "attributes": {},
    "learner_model_param": {
      "base_score": "0.0",
      "num_class": "0",
      "num_feature": "1"
    },
    "gradient_booster": {
      "name": "gbtree",
      "model": {
        "gbtree_model_param": {
          "num_trees": "1",
          "size_leaf_vector": "0"
        },
        "tree_info": [0],
        "trees": [
          {
            "id": 0,
            "tree_param": {
              "num_deleted": "0",
              "num_feature": "1",
              "num_nodes": "3",
              "size_leaf_vector": "0"
            },
            "parents": [2147483647, 0, 0],
            "left_children": [1, -1, -1],
            "right_children": [2, -1, -1],
            "split_type": [1, 0, 0],
            "split_indices": [0, 0, 0],
            "split_conditions": [NaN, 1.0, 2.0],
            "default_left": [0, 0, 0],
            "categories_nodes": [0],
            "categories_segments": [0],
            "categories_sizes": [3],
            "categories": [0, 3, 31],
            "loss_changes": [0.0, 0.0, 0.0],
            "base_weights": [0.0, 0.0, 0.0],
            "sum_hessian": [0.0, 0.0, 0.0]
          }
        ]
      }
    }
  }
}"""

booster = xgboost.Booster()
with tempfile.TemporaryDirectory() as tempdir:
    model_path = tempdir + "/model.json"
    with open(model_path, 'w') as f:
        f.write(model_json)
    booster.load_model(model_path)
    data = pd.DataFrame({"x0": list(range(33))}).astype("category")
    expected = pd.Series([(2.0 if x in [0, 3, 31] else 1.0) for x in data['x0']])
    preds = booster.inplace_predict(data)
    for i, x in enumerate(data['x0']):
        if preds[i] != expected[i]:
            print(f"Incorrect prediction for x={x}. Expected {expected[i]} but got {preds[i]}")
```
This prints:
```
Incorrect prediction for x0=32. Expected 1.0 but got 2.0
```
[...]

PR #8585, opened 2022-12-12
[WIP] Save `num_categories` in Learner.

trivialfis:
- Introduce additional fields in both DMatrix and Learner for the number of categories.
- Pass it down using `LearnerModelParam`.
- Support user-supplied information for categorical features via a new type.

Close https://github.com/dmlc/xgboost/issues/8573 .


trivialfis:
This is not necessary. I will port sso

PR #8591, opened 2022-12-13
Fix inference with categorical feature.

trivialfis:
Always go left with categories that are greater than the ones store in tree.

Related: https://github.com/dmlc/xgboost/pull/7529#issuecomment-999274606

Close https://github.com/dmlc/xgboost/issues/8573 .

https://github.com/dmlc/xgboost/pull/8585 is not necessary. Assuming the categories are encoded, knowing the greatest one is sufficient for inference as those exceeded the bitset is always the ones that are not chosen and should go to the left branch. The bug was introduced in https://github.com/dmlc/xgboost/pull/7529 for unnecessary reason.

### DIFF
--- a/src/common/categorical.h
+++ b/src/common/categorical.h
@@ -55,14 +55,12 @@ inline XGBOOST_DEVICE bool InvalidCat(float cat) {
  */
 inline XGBOOST_DEVICE bool Decision(common::Span<uint32_t const> cats, float cat) {
   KCatBitField const s_cats(cats);
-  // FIXME: Size() is not accurate since it represents the size of bit set instead of
-  // actual number of categories.
   if (XGBOOST_EXPECT(InvalidCat(cat), false)) {
     return true;
   }
 
   auto pos = KCatBitField::ToBitPos(cat);
-  // If input categorical is larger than the size of bit field, it implies that the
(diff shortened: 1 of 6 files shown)


src/common/categorical.h
```diff
@@ -48,20 +48,23 @@ inline XGBOOST_DEVICE bool InvalidCat(float cat) {
...
  *
- *  For one hot split, go to left if it's NOT the matching category.
+ *   Go to left if it's NOT the matching category, which matches one-hot encoding.
  */
-template <bool validate = true>
-inline XGBOOST_DEVICE bool Decision(common::Span<uint32_t const> cats, float cat, bool dft_left) {
+inline XGBOOST_DEVICE bool Decision(common::Span<uint32_t const> cats, float cat) {
   KCatBitField const s_cats(cats);
   // FIXME: Size() is not accurate since it represents the size of bit set instead of
   // actual number of categories.
```
hcho3, line 59:
```suggestion
```
This comment is no longer necessary since `Size()` is no longer used.

hcho3 (changes requested):

src/common/categorical.h
trivialfis: Thank you for the good catch.

hcho3 (approved):

PR #8597, opened 2022-12-14
Added ranges for alpha and lambda in docs

edjeordjian:


doc/parameter.rst
```diff
@@ -121,10 +121,12 @@ Parameters for Tree Booster
 * ``lambda`` [default=1, alias: ``reg_lambda``]
 
   - L2 regularization term on weights. Increasing this value will make model more conservative.
+  - range: [0,∞]
```
trivialfis, line 124: Thank you for the improvement, could you please consider using a LaTeX symbol with `:math:`?

trivialfis: it's `:math:` ,not `math:`.

edjeordjian: Thank you for your patience 😅

trivialfis (approved):
Thank you for the improvement!

PR #8605, opened 2022-12-16
Refactor PySpark tests.

trivialfis:
- Convert classifier tests to pytest tests.
- Replace hardcoded tests.

Extracted from https://github.com/dmlc/xgboost/pull/8272 .


RAMitchell (approved):
LGTM. My only concern is that the time does not increase dramatically.

trivialfis:
The test is mostly handcrafted and hence small.

PR #8607, opened 2022-12-16
Implement fit stump.

trivialfis:
Extracted from https://github.com/dmlc/xgboost/pull/8272 .


trivialfis:
The output of `FitStump` is the weight for a multi-target tree without regularization. For now, https://github.com/dmlc/xgboost/pull/8272 calculates uses mean value of this weight for multi-target.

RAMitchell (approved):

PR #8610, opened 2022-12-17
[R] remove unused compiler flag RABIT_CUSTOMIZE_MSG_

jameslamb:
Proposes removing compiler flag `-DRABIT_CUSTOMIZE_MSG_` in CMake-based builds of the R package.

It has been in the R package here since #759 and was used in `rabit` to control whether or not to override some logging functions:

https://github.com/dmlc/rabit/blob/f307acebdc8b165e6d50dce2e287a1bc3eec8870/include/rabit/internal/utils.h#L82

But seems to no longer be used.

Since `rabit` is now vendored into XGBoost (since #759), I just tested that like this:

```shell
git grep RABIT_CUSTOMIZE
```


trivialfis (approved):
Thank you for spotting that unused flag and digging into it!

PR #8614, opened 2022-12-19
[R] remove unused imports in tests

jameslamb:
This PR proposes removing all uses of `require()` in the R package's tests.

Most of those are unnecessary. For example, when using `{testthat}` it isn't necessary to load `{testthat}` or the library being tested in test files.

### Benefit of this change

Simplifies the tests by removing unnecessary code.

Moves `{xgboost}` closer to the standard for managing optional dependencies prescribed in Writing R Extensions. From "Suggested Packages" in that doc ([link](https://cran.r-project.org/doc/manuals/R-exts.html#Suggested-packages)):

> ... someone wanting to run the ... tests ... may not have a suggested package available... using `if(requireNamespace("pkgname"))` is preferred, if possible.

### What's wrong with `require()`?

Despite its name, `require()` does not actually fail if the library you ask it to load isn't installed.

```shell
Rscript -e "require('does.not.exist')"
echo $?
# 0
```

`require(pkg)` is more expensive than `requireNamespace(pkg)` because it both loads and attaches `pkg`, while also being less strict than `library(pkg)` because it warns but does not raise an exception if `pkg` is not installed.

For this reasons, either `requireNamespace()` or `library()` is almost always preferable to `require()`. See `?requireNamespace` or https://r-pkgs.org/dependencies-mindset-background.html#sec-dependencies-namespace for more discussion.

### How I tested this

Installed `{xgboost}` on this branch.

```shell
R CMD INSTALL --with-keep.source R-package/
```

Then removed suggested packages `{vcd}` and `{titanic}` and ran the tests.

```shell
Rscript -e "remove.packages(c('vcd', 'titanic'))"
cd R-package/tests
Rscript testthat.R
```

The resulted in the following results

```text
[ FAIL 0 | WARN 2 | SKIP 21 | PASS 1010 ]
```

And a summary explaining the reason for tests being skipped.

```text
── Skipped tests  ──────────────────────────────────────
• Optional testing dependency 'titanic' not found. (1)
• Optional testing dependency 'vcd' not found. (20)
```

Next, I installed those two packages and ran the tests again.

```shell
Rscript -e "install.packages(c('titanic', 'vcd'), repos = 'https://cran.r-project.org')"
Rscript testthat.R
```

And saw all the tests succeed + around 300 more run.

```text
[ FAIL 0 | WARN 2 | SKIP 1 | PASS 1326 ]
```

### Notes for Reviewers

Thanks for your time and consideration! Please let me know if you'd prefer that I slow down these small maintenance contributions on the R package, or batch them up into larger PRs.

### DIFF
--- a/R-package/tests/testthat/test_basic.R
+++ b/R-package/tests/testthat/test_basic.R
@@ -1,6 +1,3 @@
-require(xgboost)
-library(Matrix)
-
 context("basic functions")
 
 data(agaricus.train, package = 'xgboost')
--- a/R-package/tests/testthat/test_callbacks.R
+++ b/R-package/tests/testthat/test_callbacks.R
@@ -1,9 +1,4 @@
 # More specific testing of callbacks
-
-require(xgboost)
-require(data.table)
-require(titanic)
-
 context("callbacks")
 
 data(agaricus.train, package = 'xgboost')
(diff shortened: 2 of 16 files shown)


trivialfis (approved):
Thank you for cleaning up the imports and the educational description! The PR looks good to me.

PR #8616, opened 2022-12-20
Initial support for multi-target tree.

trivialfis:
This is a rough PR for early reviews and discussions, it contains bugs and unfinished code.

- Add a new multi-target tree structure that's embedded into the existing regtree.
- Add a new builder to hist.
- Add a new evaluator to hist.

I try to reuse as much existing code as possible. For instance, there's no structural change to the histogram builder and the implementation just iterates over a list of builders for each target. However, this might change in the future as we want a more integrated implementation. The evaluation code has to be rewritten for performance. Lastly, there are some optimization techniques for the multi-target tree when the number of targets is huge. Some of the known methods include summarizing the gradient, selecting the gradient, projecting the gradient, optimizing for sparse gradient, etc. I haven't implemented any of those yet, the PR is for the core multi-target tree structure.

There are other cases where we have vector leaf but not multi-target tree grower. For instance, we might want the leaf to be a linear model, or it might contain extra parameters for a probability distribution. These will require new tree training algorithms, but the tree structure is largely the same. The PR is a proof of concept.

The implementation is not as efficient as the single-target one, which doesn't represent the theoretical performance of the strategy.

For small testing datasets, using vector-leaf might lead to significant overfit.

## what's working

- Multi-target regression and single-target-multi-class classification training using `hist` and `gbtree` with most of the tree parameters except for the mono constraint. Numeric feature only.
- Inference on CPU including QDM, inplace, DMatrix for normal outputs. (no leaf index, SHAP, etc)

## what's not working
everything else.

## Related
- https://github.com/dmlc/xgboost/pull/5460
- https://github.com/dmlc/xgboost/issues/2087

### DIFF
--- a/doc/parameter.rst
+++ b/doc/parameter.rst
@@ -226,7 +226,7 @@ Parameters for Tree Booster
     list is a group of indices of features that are allowed to interact with each other.
     See :doc:`/tutorials/feature_interaction_constraint` for more information.
 
-* ``multi_strategy``, [default = ``composite``]
+* ``multi_strategy``, [default = ``one_output_per_tree``]
 
   .. versionadded:: 2.0.0
 
@@ -235,8 +235,8 @@ Parameters for Tree Booster
   - The strategy used for training multi-target models, including multi-target regression
   and multi-class classification. See :doc:`/tutorials/multioutput` for more information.
 
-    - ``composite``: One model for each target.
-    - ``monolithic``:  Use multi-target trees.
+    - ``one_output_per_tree``: One model for each target.
+    - ``multi_output_tree``:  Use multi-target trees.
 
 .. _cat-param:
 
--- a/doc/parameter.rst
+++ b/doc/parameter.rst
@@ -226,6 +226,18 @@ Parameters for Tree Booster
     list is a group of indices of features that are allowed to interact with each other.
     See :doc:`/tutorials/feature_interaction_constraint` for more information.
 
+* ``multi_strategy``, [default = ``composite``]
+
+  .. versionadded:: 2.0.0
+
+  .. note:: This parameter is working-in-progress.
+
+  - The strategy used for training multi-target models, including multi-target regression
+  and multi-class classification. See :doc:`/tutorials/multioutput` for more information.
+
+    - ``composite``: One model for each target.
+    - ``monolithic``:  Use multi-target trees.
+
 .. _cat-param:
 
 Parameters for Categorical Feature
--- a/include/xgboost/linalg.h
+++ b/include/xgboost/linalg.h
@@ -530,17 +530,17 @@ class TensorView {
   /**
    * \brief Number of items in the tensor.
    */
-  LINALG_HD [[nodiscard]] std::size_t Size() const { return size_; }
+  [[nodiscard]] LINALG_HD std::size_t Size() const { return size_; }
   /**
    * \brief Whether this is a contiguous array, both C and F contiguous returns true.
    */
-  LINALG_HD [[nodiscard]] bool Contiguous() const {
+  [[nodiscard]] LINALG_HD bool Contiguous() const {
     return data_.size() == this->Size() || this->CContiguous() || this->FContiguous();
   }
   /**
    * \brief Whether it's a c-contiguous array.
    */
-  LINALG_HD [[nodiscard]] bool CContiguous() const {
+  [[nodiscard]] LINALG_HD bool CContiguous() const {
     StrideT stride;
     static_assert(std::is_same<decltype(stride), decltype(stride_)>::value);
     // It's contiguous if the stride can be calculated from shape.
@@ -550,7 +550,7 @@ class TensorView {
   /**
    * \brief Whether it's a f-contiguous array.
    */
-  LINALG_HD [[nodiscard]] bool FContiguous() const {
+  [[nodiscard]] LINALG_HD bool FContiguous() const {
     StrideT stride;
     static_assert(std::is_same<decltype(stride), decltype(stride_)>::value);
     // It's contiguous if the stride can be calculated from shape.
--- a/python-package/xgboost/testing/__init__.py
+++ b/python-package/xgboost/testing/__init__.py
@@ -10,7 +10,6 @@ import os
 import platform
 import socket
 import sys
-import zipfile
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
 from io import StringIO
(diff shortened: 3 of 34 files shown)


src/predictor/cpu_predictor.cc
```diff
@@ -310,14 +300,8 @@ void PredictBatchByBlockOfRowsKernel(DataView batch, gbm::GBTreeModel const &mod
 
     FVecFill(block_size, batch_offset, num_feature, &batch, fvec_offset, p_thread_temp);
     // process block of rows through all trees to keep cache locality
-    if (model.learner_model_param->IsVectorLeaf()) {
```
trivialfis, line 313: This is to not rely on the model parameter, which is not serialized into JSON model.

include/xgboost/linalg.h
```diff
@@ -530,17 +530,17 @@ class TensorView {
   /**
    * \brief Number of items in the tensor.
    */
-  LINALG_HD [[nodiscard]] std::size_t Size() const { return size_; }
+  [[nodiscard]] LINALG_HD std::size_t Size() const { return size_; }
```
trivialfis, line 533: clangd is not quite happy about the place of c++ attribute when running in CUDA mode.

src/common/quantile.h
```diff
@@ -352,19 +352,6 @@ struct WQSummary {
       prev_rmax = data[i].rmax;
     }
   }
-  // check consistency of the summary
```
trivialfis, line 355: Unused function.

src/gbm/gbtree.cc
```diff
@@ -27,9 +28,11 @@
 #include "xgboost/host_device_vector.h"
 #include "xgboost/json.h"
 #include "xgboost/logging.h"
+#include "xgboost/model.h"
 #include "xgboost/objective.h"
 #include "xgboost/predictor.h"
-#include "xgboost/string_view.h"
+#include "xgboost/string_view.h"  // for StringView
```
trivialfis, line 34: I kept the custom string view for now. Some changes in c++20 string_view might be useful, we can back-port it to xgboost when needed.

src/tree/updater_quantile_hist.cc
```diff
@@ -105,6 +108,212 @@ void UpdateTree(common::Monitor *monitor_, linalg::MatrixView<GradientPair const
...
+      hists.push_back(&histogram_builder_[t].Histogram());
+    }
+    for (auto const &gmat : p_fmat->GetBatches<GHistIndexMatrix>(HistBatch(param_))) {
+      evaluator_->EvaluateSplits(*p_tree, hists, gmat.cut, best_splits);
+      break;
+    }
+    monitor_->Stop(__func__);
+  }
+
+  void LeafPartition(RegTree const &tree, linalg::MatrixView<GradientPair const> gpair,
```
trivialfis, line 291: This is not used yet. We need some work on L1 and quantile regression for estimating vector leaf.

tests/ci_build/lint_python.py
```diff
@@ -230,6 +236,11 @@ def main(args: argparse.Namespace) -> None:
     parser.add_argument("--format", type=int, choices=[0, 1], default=1)
     parser.add_argument("--type-check", type=int, choices=[0, 1], default=1)
     parser.add_argument("--pylint", type=int, choices=[0, 1], default=1)
+    parser.add_argument(
+        "--fix",
```
trivialfis, line 240: A new argument for convenience.

tests/python-gpu/test_gpu_updaters.py
```diff
@@ -32,6 +32,19 @@ def train_result(param, dmat: xgb.DMatrix, num_rounds: int) -> dict:
     return result
 
 
+class TestGPUUpdatersMulti:
```
trivialfis, line 35: I have extracted all the multi-target/class datasets into an independent hypothesis, other than the test for CPU hist, no testing logic is changed.

python-package/xgboost/testing/__init__.py
```diff
@@ -352,137 +352,6 @@ def __repr__(self) -> str:
         return self.name
 
 
-@memory.cache
```
trivialfis, line 355: pylint complains about the file being too huge (>1000 loc). I moved some of the data fetchers into `testing/data.py`.

trivialfis:
Not sure if this is useful, but you can do it just for fun:
```
def alternate(plot_result: bool) -> None:
    """Draw a circle with 2-dim coordinate as target variables."""
    from xgboost.callback import TrainingCallback

    class ResetStrategy(TrainingCallback):
        def before_iteration(self, model, epoch: int, evals_log) -> bool:
            strategy = "monolithic" if epoch % 2 == 0 else "composite"
            model.set_param({"multi_strategy": strategy})
            return False

    X, y = gen_circle()
    # Train a regressor on it
    reg = xgb.XGBRegressor(
        tree_method="hist",
        n_estimators=4,
        n_jobs=1,
        max_depth=8,
        subsample=0.6,
        callbacks=[ResetStrategy()]
    )
    reg.fit(X, y, eval_set=[(X, y)])
```

doc/parameter.rst
```diff
@@ -226,6 +226,18 @@ Parameters for Tree Booster
...
+
+  .. versionadded:: 2.0.0
+
+  .. note:: This parameter is working-in-progress.
+
+  - The strategy used for training multi-target models, including multi-target regression
+  and multi-class classification. See :doc:`/tutorials/multioutput` for more information.
+
+    - ``composite``: One model for each target.
+    - ``monolithic``:  Use multi-target trees.
```
RAMitchell, line 239: I think the names could be better.
e.g. 'multi_output_tree' and 'one_output_per_tree'.

RAMitchell (approved):

doc/parameter.rst
trivialfis: Done.

s-banach:
(Just seeing this for the first time, haven't put much brain power into it yet.)
Will it be easy to use this where the multiple targets are different parameters of a random variable?
E.g. shape and scale of a gamma variable.

StatMixedML:
@s-banach 

Essentially, multi target trees can be used if there are more than one parameter to predict. This can be useful if you want to model all parameters of a univariate and multivariate parametric distribution, see


[Multi-Target XGBoostLSS Regression](https://arxiv.org/abs/2210.06831)
[Distributional Gradient Boosting Machines](https://arxiv.org/abs/2204.00778)
[XGBoostLSS - An extension of XGBoost to probabilistic forecasting](https://arxiv.org/abs/1907.03178)

PR #8623, opened 2022-12-21
Support bitwise allreduce operations in the communicator

rongou:
We likely need bitwise allreduce operations for feature parallel training.

Part of #8424


rongou:
@trivialfis @hcho3

trivialfis (approved):
Looks good to me!

PR #8624, opened 2022-12-25
[R] [CI] add more linting checks

jameslamb:
Proposes adding the following additional linting checks on R code in the project:

* `any_duplicated_linter()`
    - *encourages the use of `anyDuplicated(x)` over `any(duplicated(x))`*
    - *this built-in is faster in most cases, because it doesn't need to compute all duplicates*
* `any_is_na_linter()`
    - *encourages the use of `anyNA(x)` over `is.na(x)`*
    - *this built-in is faster in most cases, because it doesn't need to compute whether or not every value in `x` is `NA`*
* `sprintf_linter()`
    - *detects mistakes in string formatting with `sprintf()`*
* `unreachable_code_linter()`
    - *detects code that will never be executed, e.g. after an unconditional `stop()` or `return()`*
* `vector_logic_linter()`
    - *warns against the common mistake of using vectorized logical operators like `|` and `&&` (which return a vector of results) when a single logical is expected*

Other changes in this PR are to address the following errors raised by these linters.

> warning: [unreachable_code] Code and comments coming after a top-level return() or stop() should be removed.
>
> warning: [vector_logic] Conditional expressions require scalar logical operators (&& and ||)
>
> warning: [any_is_na] anyNA(x) is better than any(is.na(x))

### How I tested this

```shell
Rscript ./tests/ci_build/lint_r.R $(pwd)

R CMD INSTALL R-package/
cd R-package/tests
Rscript testthat.R
```

### Notes for Reviewers

I chose this set of additional linters because they address efficiency and correctness, so they provide some user-facing benefit.

I might propose adding more in the future ([we use many more in LightGBM](https://github.com/microsoft/LightGBM/blob/08f0853db2cf609bad43d5df5a4d92c3fe2d7caf/.ci/lint_r_code.R#L31)), but don't want to draw away too much of maintainers' attention here.

### DIFF
--- a/R-package/R/xgb.DMatrix.R
+++ b/R-package/R/xgb.DMatrix.R
@@ -328,7 +328,6 @@ setinfo.xgb.DMatrix <- function(object, name, info, ...) {
     return(TRUE)
   }
   stop("setinfo: unknown info name ", name)
-  return(FALSE)
 }
 
 
@@ -418,7 +417,7 @@ print.xgb.DMatrix <- function(x, verbose = FALSE, ...) {
   cat(infos)
   cnames <- colnames(x)
   cat('  colnames:')
-  if (verbose & !is.null(cnames)) {
+  if (verbose && !is.null(cnames)) {
     cat("\n'")
     cat(cnames, sep = "','")
     cat("'")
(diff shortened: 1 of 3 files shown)


trivialfis (approved):
Looks good! Always good to see improvements in the codebase.

PR #8626, opened 2022-12-25
Update custom_metric_obj.rst

mrbaloglu:
For the codeblock given in line 291, in the softprob_obj method definition, the variable 'classes' is not defined. It seems it should be defined as given in the changes.


trivialfis (changes requested):
Thank you for the PR. But I think the label is one dimension.

mrbaloglu:
I agree, then, we should have 'classes = predt.shape[1]' as we see that predt is two dimensional from the nested loop below.

trivialfis (approved):
Apologies for the slow reply and thank you for the improvement. Looks good to me.

Issue #8615, opened 2022-12-19
macOS incompatibility 


worstseed:
This issue has been reported by multiple users when they attempted to install and use JAHS-Bench-201 whose surrogates are based on XGBoost on macOS. Here are a few [details](https://github.com/automl/jahs_bench_201/issues/6#issuecomment-1345633609):

MacBook Pro 15", 2018
MacOS 13.0 (Ventura) but happened on MacOS 12.* (Monterey) as well
MacBook Pro 13'', M2, 2022 (Monterey)

I created a new conda environment (conda 4.13.0) with Python 3.7.15 (but the same error with Python 3.10.) and installed JAHS-Bench-201 with `pip install jahs-bench`. Running `python -m jahs_bench_examples.minimal` leads to the following error:

```
[20:03:24] WARNING: /Users/runner/work/xgboost/xgboost/src/gbm/gbtree.cc:386: Loading from a raw memory buffer on CPU only machine.  Changing tree_method to hist.
Traceback (most recent call last):
  File "/usr/local/Caskroom/miniconda/base/envs/jahs_bench_test/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Caskroom/miniconda/base/envs/jahs_bench_test/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/Caskroom/miniconda/base/envs/jahs_bench_test/lib/python3.7/site-packages/jahs_bench_examples/minimal.py", line 13, in <module>
    run()
  File "/usr/local/Caskroom/miniconda/base/envs/jahs_bench_test/lib/python3.7/site-packages/jahs_bench_examples/minimal.py", line 5, in run
    benchmark = jahs_bench.Benchmark(task="cifar10", kind="surrogate", download=True)
  File "/usr/local/Caskroom/miniconda/base/envs/jahs_bench_test/lib/python3.7/site-packages/jahs_bench/api.py", line 135, in __init__
    loaders[kind]()
  File "/usr/local/Caskroom/miniconda/base/envs/jahs_bench_test/lib/python3.7/site-packages/jahs_bench/api.py", line 148, in _load_surrogate
    self._surrogates[o] = XGBSurrogate.load(pth) if not self._lazy else \
  File "/usr/local/Caskroom/miniconda/base/envs/jahs_bench_test/lib/python3.7/site-packages/jahs_bench/surrogate/model.py", line 462, in load
[...]

PR #8632, opened 2023-01-04
Fix loading GPU pickle with a CPU-only xgboost distribution.

trivialfis:
We can handle loading the pickle on a CPU-only machine if the XGBoost is built with CUDA enabled (Linux and Windows PyPI package), but not if the distribution is CPU-only (macOS PyPI package).

To reproduce the error: train a model with GPU hist. Load it with an XGBoost build that doesn't have CUDA capability.

Close https://github.com/dmlc/xgboost/issues/8615 .

### DIFF
--- a/src/gbm/gbtree.cc
+++ b/src/gbm/gbtree.cc
@@ -28,6 +28,7 @@
 #include "xgboost/logging.h"
 #include "xgboost/objective.h"
 #include "xgboost/predictor.h"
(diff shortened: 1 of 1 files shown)


trivialfis:
This is a quick fix meant for backporting, once I clear the backlog for other PRs I will focus on https://github.com/dmlc/xgboost/issues/7308 instead.

RAMitchell (approved):

PR #8647, opened 2023-01-06
[R] remove unused define XGBOOST_CUSTOMIZE_LOGGER

jameslamb:
It looks to me like define `XGBOOST_CUSTOMIZE_LOGGER` does not have any effect on compilation of this project.

This PR proposes removing it.

### How I tested this

Checked for references like this:

```shell
# check files in source control
git grep XGBOOST_CUSTOMIZE_LOGGER

# check files pulled in from git submodules
grep -R XGBOOST_CUSTOMIZE_LOGGER
```

Only found the ones removed in the diff here.

Built the R package and ran the tests.

```shell
R CMD INSTALL R-package/
cd R-package/tests
Rscript testthat.R
```

### Notes for Reviewers

According to the blame ([link](https://github.com/dmlc/xgboost/blame/2b88099c7406b014b2c6babd71cdf2e3eaa9cb7a/include/xgboost/base.h#L34)) this was added back in 2016, in #759. I suspect that some changes since then made the define unnecessary.


trivialfis:
Thank you for the cleanup. I need to take a deeper look into this change. Since CRAN prohibits the use of stdout/stderr in C libraries, we redirect the logging to `Rprintf`. We need to be prudent to not break the CRAN policy.

jameslamb:
Totally understand!

I can help with that. I know the policy you're talking about...CRAN requires that compiled code write to R's output stream (e.g. via the use of `Rprintf` and its variants) instead of directly to stdout.

In XGBoost today, that behavior is guarded by the `XGBOOST_STRICT_R_MODE` define.

https://github.com/dmlc/xgboost/blob/06a1cb6e03395c4f2ea0d6c8bc76f5bf91b061ba/R-package/src/xgboost_custom.cc#L9-L14

https://github.com/dmlc/xgboost/blob/06a1cb6e03395c4f2ea0d6c8bc76f5bf91b061ba/src/logging.cc#L14-L21

If the change in this PR caused the R package to violate that policy, the CI jobs running `R CMD check --as-cran` would be failing.

Right here on this line `checking compiled code ...`: https://github.com/dmlc/xgboost/actions/runs/3852619686/jobs/6564903824#step:7:349

https://github.com/dmlc/xgboost/blob/06a1cb6e03395c4f2ea0d6c8bc76f5bf91b061ba/.github/workflows/r_tests.yml#L123

https://github.com/dmlc/xgboost/blob/06a1cb6e03395c4f2ea0d6c8bc76f5bf91b061ba/tests/ci_build/test_r_package.py#L121

trivialfis (approved):
Apologies for the delay, thank you for cleaning up the code!

PR #8648, opened 2023-01-06
[Breaking][jvm-packages] Bump rapids version to 22.12.0

wbo4958:
This PR bumps spark version to 3.1.1 and the rapids version to 22.12.0, which results in the latest xgboost can't run with the old rapids packages.


trivialfis (approved):
Thank you for the PR! XGBoost usually prioritizes compatibility with the latest dependencies over older versions, so this kind of breaking is fine.

wbo4958:
@trivialfis Fixed, thx

Issue #8663, opened 2023-01-12
CRAN packages not using the correct compiler in configure scripts


trivialfis:
That is

C50 Cubist GPBayes RDieHarder RPostgreSQL RcppGSL RhpcBLASctl Rmixmod
Rrdrand SimInf SymTS eaf excursions fRLR ff float getip git2r gsl gslnls
hdf5r island memuse mmap ore poismf randtoolbox rayrender rgeolocate
rmatio rnetcarto rngWELL scrypt survSNP xgboost

These show

checking for gcc... gcc

even on platforms using clang as the compiler.  Please re-read

https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Configure-and-cleanup

There need be no 'gcc' compiler, it might be faked (macOS) or very old
(Solaris).  The aim of a configure script is to test things using the
same setup as will be used to compile code in the package.

In several cases you are using a C compiler in configure but a C++
compiler to compile the code: they usually have access to different
headers.  Looks like this is the case in at least

GPBayes RcppGSL Rmixmod fRLR ff rayrender xgboost

The logs showing

        checking for gcc option to accept ISO C89... none needed

need configure rebuilt with a modern autoconf, which will check C11 (and
that or C17 are the default in all the current compilers we know of --
however, that old autoconf test will fail when C23 comes into use).

Please correct before 2023-01-26 to safely retain your package on CRAN.

--
Brian D. Ripley,                  [ripley@stats.ox.ac.uk](mailto:ripley@stats.ox.ac.uk)
Emeritus Professor of Applied Statistics, University of Oxford

hcho3:
Since we've had trouble getting R tests working on XGBoost 1.7, I am skeptical that we'd be able to make the fix in time. We should prepare for the possibility of having XGBoost pulled from CRAN.

trivialfis:
I think I can give it a go, at least for 1.6

hcho3:
Maybe we should consider using an alternative distribution mechanism, like R-universe.

trivialfis:
I think it's fine, working on other distributions is not less time consuming than fixing the issues on CRAN. I will take another look into the tests today and see how much work is needed.

jameslamb:
I agree with @trivialfis , it is really worth trying to stay on CRAN. And especially for `{xgboost}`... if `{xgboost}` is removed from CRAN, all of these other packages will have to either quickly release a new version not depending on `{xgboost}` or also be removed:

[...]

PR #8669, opened 2023-01-12
[R] Get CXX flags from R CMD config.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/8663 

Log for configuration:
```
checking Backtrace lib... 
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether the compiler supports GNU C++... yes
checking whether g++ -std=gnu++14 accepts -g... yes
checking for g++ -std=gnu++14 option to enable C++11 features... none needed
checking for backtrace in -lexecinfo... no
checking endian... 
configure: creating ./config.status
config.status: creating src/Makevars
** libs
Makevars:16: -DXGBOOST_STRICT_R_MODE=1
Makevars:16: -DDMLC_LOG_BEFORE_THROW=0
Makevars:16: -DDMLC_ENABLE_STD_THREAD=1
Makevars:16: -DDMLC_DISABLE_STDIN=1
Makevars:16: -DDMLC_LOG_CUSTOMIZE=1
Makevars:16: -DXGBOOST_CUSTOMIZE_LOGGER=1
```

All credit to @jameslamb and the work on LightGBM!

### DIFF
--- a/R-package/configure
+++ b/R-package/configure
@@ -620,11 +620,11 @@ OPENMP_LIB
 OPENMP_CXXFLAGS
 OBJEXT
 EXEEXT
-ac_ct_CC
+ac_ct_CXX
 CPPFLAGS
 LDFLAGS
-CFLAGS
-CC
+CXXFLAGS
+CXX
 target_alias
 host_alias
 build_alias
@@ -671,11 +671,12 @@ enable_option_checking
       ac_precious_vars='build_alias
 host_alias
 target_alias
-CC
-CFLAGS
+CXX
+CXXFLAGS
 LDFLAGS
 LIBS
-CPPFLAGS'
+CPPFLAGS
+CCC'
 
 
 # Initialize some variables set by options.
@@ -1291,8 +1292,8 @@ if test -n "$ac_init_help"; then
   cat <<\_ACEOF
 
 Some influential environment variables:
-  CC          C compiler command
-  CFLAGS      C compiler flags
+  CXX         C++ compiler command
+  CXXFLAGS    C++ compiler flags
   LDFLAGS     linker flags, e.g. -L<lib dir> if you have libraries in a
               nonstandard directory <lib dir>
   LIBS        libraries to pass to the linker, e.g. -l<library>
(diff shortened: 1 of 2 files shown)


hcho3 (approved):

jameslamb (commented):
Excellent!

This project uses a C compiler (`$CC`) to test for OpenMP, I think you may also need to do one of the following.

```shell
# get C compiler from R's configuration
CC=`"${R_HOME}/bin/R" CMD config CCC`

# ... or switch the OpenMP test to using a C++ compiler, like LightGBM does
# (note the use of CXX instead of CC)

${CXX} -o conftest conftest.c ${CPPFLAGS} ${LDFLAGS} ${OPENMP_LIB} ${OPENMP_CXXFLAGS} 2>/dev/null && ./conftest && ac_pkg_openmp=yes
```

LightGBM reference: https://github.com/microsoft/LightGBM/blob/23403a7c2c4d38961506968291f13c3f344ccb6f/R-package/configure.ac#L140

Otherwise, the detection of OpenMP in `configure` might not match what happens at compilation time. For example, in the situation where someone on a Mac is using `~/.R/Makevars` to force R to use `gcc` instead of `clang`.

I'm not sure which of the two approaches is preferable for XGBoost. In LightGBM we chose the `CXX` approach because we use only C++ source files. But XGBoost has this one `.c` file (https://github.com/dmlc/xgboost/blob/master/R-package/src/init.c) so I'm not sure.

trivialfis:
@jameslamb Thank you for catching that! I updated the c-related flags, currently running tests on rhub with fedora clang and macos cran setup.

trivialfis:
macos test looks good: https://builder.r-hub.io/status/xgboost_2.0.0.1.tar.gz-02f421441fd7716d01e5980d4b4a689d

trivialfis:
Macos compilation seems fine https://builder.r-hub.io/status/xgboost_2.0.0.1.tar.gz-8c4a9c12ea429fc21922b8c6381bffa2

PR #8676, opened 2023-01-13
Initial support for column-split cpu predictor

rongou:
When data is split by column, produce predictions in two passes by leveraging bit vectors.

Part of #8424


rongou:
@trivialfis @hcho3

trivialfis:
merging the master branch should fix the CI error.

src/predictor/cpu_predictor.cc
```diff
@@ -284,16 +283,236 @@ void FillNodeMeanValues(RegTree const* tree, std::vector<float>* mean_values) {
...
 
-class CPUPredictor : public Predictor {
- protected:
-  // init thread buffers
-  static void InitThreadTemp(int nthread, std::vector<RegTree::FVec> *out) {
-    int prev_thread_temp_size = out->size();
-    if (prev_thread_temp_size < nthread) {
-      out->resize(nthread, RegTree::FVec());
+// init thread buffers
+static void InitThreadTemp(int nthread, std::vector<RegTree::FVec> *out) {
```
trivialfis, line 287: Use an anonymous namespace instead of static if it's not intended to be used outside of the TU.

src/predictor/cpu_predictor.cc
```diff
@@ -284,16 +283,236 @@ void FillNodeMeanValues(RegTree const* tree, std::vector<float>* mean_values) {
...
+    }
+  }
+
+  template <typename DataView, size_t block_of_rows_size>
+  void PredictBatchKernel(DataView batch, std::vector<bst_float> *out_preds) {
+    int32_t const num_group = model_.learner_model_param->num_output_group;
+
+    CHECK_EQ(model_.param.size_leaf_vector, 0) << "size_leaf_vector is enforced to 0 so far";
+    // parallel over local batch
+    const auto nsize = static_cast<bst_omp_uint>(batch.Size());
```
trivialfis, line 468: No need to set `bst_omp...` explicitly, use `auto` and `ParallelFor` will do the casting:
``` c++
common::Parallel(n_blocks, n_threads, [&](auto block_id) {}); // auto block_id has the same type as `n_blocks`.
```

src/predictor/cpu_predictor.cc
```diff
@@ -284,16 +283,236 @@ void FillNodeMeanValues(RegTree const* tree, std::vector<float>* mean_values) {
...
+  uint32_t const tree_begin_;
+  uint32_t const tree_end_;
+
+  std::vector<std::size_t> tree_sizes_{};
+  std::vector<std::size_t> tree_offsets_{};
+  std::size_t bits_per_row_{};
+  std::vector<RegTree::FVec> feat_vecs_{};
+
+  std::size_t n_rows_;
+  std::vector<BitVector::value_type> decision_storage_{};
```
trivialfis, line 508: Could you please add some comments about the layout of the storage? How to index it?

src/predictor/cpu_predictor.cc
rongou: Done.

src/predictor/cpu_predictor.cc
rongou: Done.

trivialfis (approved):

PR #8681, opened 2023-01-14
[R] Fix threads used to create DMatrix in predict.

trivialfis:


trivialfis:
Hopefully, this is the one last fix needed to address the CRAN release issue.

hcho3 (approved):

Issue #5496, opened 2020-04-08
Rpackage Version 1.0.0.2 crashes RStudio


tbstockton:
Running the R package Version 1.0.0.2 crashes RStudio. This occurs  at the xgb.DMatrix call in the xgbtrain example. As well as when using saved xgboost objects fit under v0.90.0.2 or fitting new models.

data(agaricus.train, package='xgboost')
data(agaricus.test, package='xgboost')
dtrain <- xgb.DMatrix(agaricus.train$data, label = agaricus.train$label)

R.version
platform       x86_64-apple-darwin15.6.0   
arch           x86_64                      
os             darwin15.6.0                
system         x86_64, darwin15.6.0        
status                                     
major          3                           
minor          5.2                         
year           2018                        
month          12                          
day            20                          
svn rev        75870                       
language       R                           
version.string R version 3.5.2 (2018-12-20)
nickname       Eggshell Igloo 

RStudio Version 1.2.5033

trivialfis:
I can't reproduce on Linux.  @hcho3 Could you please help taking a look?

tbstockton:
Sure, glad to help.  v0.90.0.2 worked fine on the described MACOS system. v1.0.0.2 crashes on the xgb.train example when running xgb.DMatrix().  Also crashes when I load and predict using a v0.90.0.2 xgboost object build on the below Ubuntu.

It seems to run fine under RStudio server on:
NAME="Ubuntu"
VERSION="18.04.4 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.4 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic

trivialfis:
@tbstockton What happens if you simply run it with `Rscript` without using studio?

trivialfis:
I also tried on RStudio, didn't reproduce it.

hcho3:
@tbstockton I just tried it on my Macbook Pro (2019, MacOS Catalina) and did not reproduce the issue.

I installed R and RStudio using Homebrew:
```
brew install r
[...]

Issue #1715, opened 2016-10-28
OMP: Error #15: Initializing libiomp5.dylib, but found libiomp5.dylib already initialized.


symPhysics:
For bugs or installation issues, please provide the following information.
The more information you provide, the more easily we will be able to offer
help and advice.
## Environment info

Operating System: Mac OSX Sierra 10.12.1

Compiler:

Package used (python):

`xgboost` version used: xgboost 0.6a2

If you are using python package, please provide
1. The python version and distribution: 
   Pythong 2.7.12
2. The command to install `xgboost` if you are not installing from source
   pip install xgboost
## Steps to reproduce
1.  from xgboost import XGBClassifier
    import numpy as np
    import matplotlib.pyplot as plt
    x = np.array([[1,2],[3,4]])
    y = np.array([0,1])
    clf = XGBClassifier(base_score = 0.005)
    clf.fit(x,y)
    plt.hist(clf.feature_importances_)
   ## What have you tried?
   
   See the error message:
    "OMP: Error #15: Initializing libiomp5.dylib, but found libiomp5.dylib already initialized.
   OMP: Hint: This means that multiple copies of the OpenMP runtime have been linked into the program. That is dangerous, since it can degrade performance or cause incorrect results. The best thing to do is to ensure that only a single OpenMP runtime is linked into the process, e.g. by avoiding static linking of the OpenMP runtime in any library. As an unsafe, unsupported, undocumented workaround you can set the environment variable KMP_DUPLICATE_LIB_OK=TRUE to allow the program to continue to execute, but that may cause crashes or silently produce incorrect results. For more information, please see http://www.intel.com/software/products/support/."

I tried:
import os
os.environ['KMP_DUPLICATE_LIB_OK']='True'

It can do the job for me. But it is kind of ugly.

---

I know it might be not the problem of xgboost, but I'm pretty sure this problem happened after I upgrade xgboost using 'pip install xgboost'.    I post the issue here to see if someone had the same problem as me.  I have very little knowledge about OpenMP.  Please help!  
Thanks in advance!

staubda:
I'm having the same issue when I try to use both xgboost and matplotlib.

OS X 10.10.5
Python 2.7.12
xgboost 0.6
matplotlib 1.5.1

jimthompson5802:
I've encountered the same problem with xgboost and matplotlib.  Same symptom...kernel dies with the same error message as reported by @symPhysics 

Mac OS X 10.12.3

[...]

PR #8684, opened 2023-01-15
[R] fix OpenMP detection on macOS

jameslamb:
Detection of OpenMP on macOS in the R package's `configure` script is currently not working.

This PR proposes switching from a C compiler to a C++ compiler in `R-package/configure` to fix that.

### Why using a C compiler breaks this check

This code in `configure.ac`...

https://github.com/dmlc/xgboost/blob/e7d612d22ca7bf82ada8f581a40253e3cf7230b4/R-package/configure.ac#L23

https://github.com/dmlc/xgboost/blob/e7d612d22ca7bf82ada8f581a40253e3cf7230b4/R-package/configure.ac#L57

... results in `autoconf` generating a file `conftest.cpp`

https://github.com/dmlc/xgboost/blob/e7d612d22ca7bf82ada8f581a40253e3cf7230b4/R-package/configure#L2745

https://github.com/dmlc/xgboost/blob/e7d612d22ca7bf82ada8f581a40253e3cf7230b4/R-package/configure#L2849

... but then XGBoost's `configure` file is looking instead for a file named `conftest.c`

https://github.com/dmlc/xgboost/blob/e7d612d22ca7bf82ada8f581a40253e3cf7230b4/R-package/configure.ac#L58

### How I tested this

On my mac tonight (macOS 12.2.1, R 4.2.2, compiling with `clang++` 13.0.0), I ran the following on `master`

```shell
cd R-package
R CMD INSTALL .
```

And saw the following in logs

```text
checking whether OpenMP will work in a package... no
*****************************************************************************************
         OpenMP is unavailable on this Mac OSX system. Training speed may be suboptimal.
         To use all CPU cores for training jobs, you should install OpenMP by running

             brew install libomp
*****************************************************************************************
```

@trivialfis noted the same thing on recent R Hub checks, https://github.com/dmlc/xgboost/pull/8669#issuecomment-1381000453

Repeating that on this branch, the R package compiles, links to OpenMP, and its unit tests succeed.

```shell
cd tests
Rscript testthat.R
```

### Notes for Reviewers

Sorry for not noticing this while reviewing #8669. It wasn't until today that I had time to look closely and test.

In LightGBM, we try to catch issues like this by `grep`-ing the build logs: https://github.com/microsoft/LightGBM/blob/5df7d516f3e64e8431aecfea892a850afeb787bb/.ci/test_r_package.sh#L250-L267. I'm not proposing that here, but you might want to consider it in the future.

### DIFF
--- a/R-package/configure
+++ b/R-package/configure
@@ -2101,8 +2101,6 @@ CXX14STD=`"${R_HOME}/bin/R" CMD config CXX14STD`
 CXX="${CXX14} ${CXX14STD}"
 CXXFLAGS=`"${R_HOME}/bin/R" CMD config CXXFLAGS`
 
-CC=`"${R_HOME}/bin/R" CMD config CC`
-CFLAGS=`"${R_HOME}/bin/R" CMD config CFLAGS`
 CPPFLAGS=`"${R_HOME}/bin/R" CMD config CPPFLAGS`
 
 LDFLAGS=`"${R_HOME}/bin/R" CMD config LDFLAGS`
@@ -2857,7 +2855,7 @@ main (void)
   return 0;
 }
 _ACEOF
-  ${CC} -o conftest conftest.c ${CPPFLAGS} ${LDFLAGS} ${OPENMP_LIB} ${OPENMP_CXXFLAGS} 2>/dev/null && ./conftest && ac_pkg_openmp=yes
+  ${CXX} -o conftest conftest.cpp ${CPPFLAGS} ${LDFLAGS} ${OPENMP_LIB} ${OPENMP_CXXFLAGS} 2>/dev/null && ./conftest && ac_pkg_openmp=yes
   { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: ${ac_pkg_openmp}" >&5
 printf "%s\n" "${ac_pkg_openmp}" >&6; }
   if test "${ac_pkg_openmp}" = no; then
--- a/R-package/configure
+++ b/R-package/configure
@@ -2101,6 +2101,8 @@ CXX14STD=`"${R_HOME}/bin/R" CMD config CXX14STD`
 CXX="${CXX14} ${CXX14STD}"
 CXXFLAGS=`"${R_HOME}/bin/R" CMD config CXXFLAGS`
 
+CC=`"${R_HOME}/bin/R" CMD config CC`
+CFLAGS=`"${R_HOME}/bin/R" CMD config CFLAGS`
 CPPFLAGS=`"${R_HOME}/bin/R" CMD config CPPFLAGS`
 
 LDFLAGS=`"${R_HOME}/bin/R" CMD config LDFLAGS`
--- a/R-package/configure.ac
+++ b/R-package/configure.ac
@@ -15,6 +15,8 @@ CXX14STD=`"${R_HOME}/bin/R" CMD config CXX14STD`
 CXX="${CXX14} ${CXX14STD}"
 CXXFLAGS=`"${R_HOME}/bin/R" CMD config CXXFLAGS`
 
+CC=`"${R_HOME}/bin/R" CMD config CC`
+CFLAGS=`"${R_HOME}/bin/R" CMD config CFLAGS`
 CPPFLAGS=`"${R_HOME}/bin/R" CMD config CPPFLAGS`
 
 LDFLAGS=`"${R_HOME}/bin/R" CMD config LDFLAGS`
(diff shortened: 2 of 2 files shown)


R-package/configure
```diff
@@ -2101,8 +2101,6 @@ CXX14STD=`"${R_HOME}/bin/R" CMD config CXX14STD`
 CXX="${CXX14} ${CXX14STD}"
 CXXFLAGS=`"${R_HOME}/bin/R" CMD config CXXFLAGS`
 
-CC=`"${R_HOME}/bin/R" CMD config CC`
```
trivialfis, line 2104: We might need the CC flag here due to function registration in `R-package/src/init.c`.

trivialfis (commented):
Thank you for the fix! Small question inlined in the comments.

I don't have a macos device, @hcho3 please help look into this. ;-)

R-package/configure
hcho3: Yes, let's retain this line.

R-package/configure
```diff
@@ -2101,8 +2101,6 @@ CXX14STD=`"${R_HOME}/bin/R" CMD config CXX14STD`
...
-CFLAGS=`"${R_HOME}/bin/R" CMD config CFLAGS`
```
hcho3, line 2105: Let's retain this line too.

hcho3 (changes requested):
LGTM. But please keep the two lines with `CC` and `CFLAGS`.

hcho3:
@trivialfis I think we should include this in our CRAN submission, if you plan to re-submit it.

jameslamb:
Sure, sorry about that! Just pushed https://github.com/dmlc/xgboost/pull/8684/commits/2f59118245242fa71b00944d4dba291f47136e46 adding the C compiler back.

And regenerated `configure` like this:

```shell
autoconf \
    --output configure \
    ./configure.ac
```

trivialfis (approved):

trivialfis:
> In LightGBM, we try to catch issues like this by grep-ing the build logs

That's a really good suggestion. Will open an issue for this.

jameslamb:
Just like that error says, you probably have multiple versions of OpenMP installed on your Mac.

You could find them like this:

```shell
find /usr -name 'libomp.dylib' -type f
find ${HOME} -name 'libomp.dylib' -type f
```

One common cause of this is the use of `conda`, which may put many copies of OpenMP onto your system ([see this discussion in the LightGBM docs](https://lightgbm.readthedocs.io/en/v3.3.2/FAQ.html#lightgbm-crashes-randomly-with-the-error-like-initializing-libiomp5-dylib-but-found-libomp-dylib-already-initialized).

For example, on my mac I see the following running those commands

```text
/usr/local/Cellar/libomp/15.0.7/lib/libomp.dylib
${HOME}/miniconda3/pkgs/llvm-openmp-14.0.4-ha654fa7_0/lib/libomp.dylib
${HOME}/miniconda3/pkgs/llvm-openmp-15.0.6-h61d9ccf_0/lib/libomp.dylib
${HOME}/miniconda3/lib/libomp.dylib
${HOME}/miniconda3/envs/xgboost-dev/lib/libomp.dylib
${HOME}/miniconda3/envs/pycharm-lightgbm/lib/libomp.dylib
${HOME}/miniconda3/envs/pycharm-lightgbm/lib/python3.9/site-packages/sklearn/.dylibs/libomp.dylib
```

If you're using R from `conda`, you can see the suggestions from #1715 for how to work around that.

If not, and you want to be sure R processes always find the Homebrew one, you could run something like the following.

```shell
export LD_LIBRARY_PATH="/usr/local/opt/libomp/lib:${LD_LIBRARY_PATH}"
```

Or achieve similar in `~/.bashrc`, `~/.bashprofile`, etc.

----

Please also note that I'm not a maintainer here. But I suspect that maintainers would prefer you search the open issues for similar error messages (like #5496 which you commented on a few years ago), and then if you don't find what you're looking for, open a new issue (instead of commenting on a closed pull request) with questions like this. And ideally providing a [minimal, reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) showing exactly what you did and describing your environment.

leedrake5:
@jameslamb Thanks for the very helpful feedback, but apologies if the original message went astray, just wanted to report the problem I ran into following the installation instructions here. I think you've correctly identified the problem on my setup, currently working to resolve that.

barracuda156:
It is still broken, and gives a wrong recommendation:
```
checking whether OpenMP will work in a package... no
*****************************************************************************************
         OpenMP is unavailable on this Mac OSX system. Training speed may be suboptimal.
         To use all CPU cores for training jobs, you should install OpenMP by running\n
             brew install libomp
```
P. S. Perhaps “should” should not be used with an optional package manager.

PR #8690, opened 2023-01-17
Make sure input numpy array is aligned.

trivialfis:
- use `np.require` to specify that the alignment is required.
- scipy csr as well.
- validate input pointer in `ArrayInterface`.

### DIFF
--- a/src/data/array_interface.h
+++ b/src/data/array_interface.h
@@ -402,10 +402,10 @@ class ArrayInterface {
     data = ArrayInterfaceHandler::ExtractData(array, n);
     static_assert(allow_mask ? D == 1 : D >= 1, "Masked ndarray is not supported.");
 
-    this->DispatchCall([&](auto const *values) {
+    this->DispatchCall([&](auto const *data_typed_ptr) {
       auto ptr = reinterpret_cast<uintptr_t>(data);
-      CHECK_EQ(ptr % std::alignment_of<std::remove_pointer_t<decltype(values)>>::value, 0)
-          << "Input pointer misalignment.";
+      auto alignment = std::alignment_of<std::remove_pointer_t<decltype(data_typed_ptr)>>::value;
+      CHECK_EQ(ptr % alignment, 0) << "Input pointer misalignment.";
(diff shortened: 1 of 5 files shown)


src/data/array_interface.h
```diff
@@ -400,6 +401,13 @@ class ArrayInterface {
 
     data = ArrayInterfaceHandler::ExtractData(array, n);
     static_assert(allow_mask ? D == 1 : D >= 1, "Masked ndarray is not supported.");
+
+    this->DispatchCall([&](auto const *values) {
```
hcho3, line 405:
```suggestion
    this->DispatchCall([&](auto const *data_casted) {
```
to indicate that the parameter is really a casted version of `data`.

tests/python/test_dmatrix.py
```diff
@@ -327,7 +327,7 @@ def test_sparse_dmatrix_csr(self):
         nrow = 100
         ncol = 1000
         x = rand(nrow, ncol, density=0.0005, format='csr', random_state=rng)
-        assert x.indices.max() < ncol - 1
+        assert x.indices.max() <= ncol - 1
```
hcho3, line 330:
```suggestion
        assert x.indices.max() < ncol
```
I think this is slightly more idiomatic.

hcho3 (changes requested):

src/data/array_interface.h
trivialfis: @hcho3 Not casted version is `void const*`, which is actually casted from input pointer (user-input-ptr -> void const* -> user-input-ptr)

hcho3: How about `data_typed_ptr`?

trivialfis: sounds good.

trivialfis: done.

tests/python/test_dmatrix.py
trivialfis: done.

hcho3 (approved):

PR #8695, opened 2023-01-18
[python-package] remove unused functions in xgboost.data

jameslamb:
Proposes removing two unused internal functions in the Python package:

* `xgboost.data._check_complex()`
* `xgboost.data._to_data_type()`

I'm unsure when the uses of these functions were removed, but I'm pretty confident that they're no longer used.

```shell
grep -R '_check_complex'
```

### DIFF
--- a/python-package/xgboost/data.py
+++ b/python-package/xgboost/data.py
@@ -52,14 +52,6 @@ def _warn_unused_missing(data: DataType, missing: Optional[FloatCompatible]) -> 
         )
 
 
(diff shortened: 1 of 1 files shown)


trivialfis (commented):
Thank you again for cleaning up the code! Let me double-check the `_check_complex`, it was required by sklearn estimator tests, I'm not sure why it's passing now.

trivialfis (approved):
Just checked, the sklearn estimator check `parametrize_with_checks` no longer passes complex data for validation.

PR #8706, opened 2023-01-22
Add sample_weight to eval_metric

BenEfrati:
Hi,
After https://github.com/dmlc/xgboost/pull/6751 provide a custom eval_metric via fit is deprecated.

In our eval_metric function we call to `get_weight()` from dmatrix.
When passing the metric function via constructor or `set_params` method we're unable to get sample_weight.

It might be useful for other sklearn metric function that accept sample_weight argument

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -119,28 +119,14 @@ def _metric_decorator(func: Callable) -> Metric:
 
     def inner(y_score: np.ndarray, dmatrix: DMatrix) -> Tuple[str, float]:
         y_true = dmatrix.get_label()
-        return func.__name__, func(y_true, y_score)
-
-    return inner
-
-def _sample_weight_metric_decorator(func: Callable) -> Metric:
-    """Decorate a metric function from sklearn.
-
-    Converts an metric function that uses the typical sklearn metric signature so that it
-    is compatible with :py:func:`train`
-
-    """
-
-    def inner(y_score: np.ndarray, dmatrix: DMatrix) -> Tuple[str, float]:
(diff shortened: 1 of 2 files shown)


idanmoradarthas:
This is an important ability!

trivialfis (commented):
Thank you for the new feature! Let's not add an additional parameter and pass the weight when present.

BenEfrati:
> Thank you for the new feature! Let's not add an additional parameter and pass the weight when present.

Agree,
Are you ok with the try except approach?

trivialfis:
I think we can patch the `_metric_decorator` as follow:
``` python
def _metric_decorator(func: Callable) -> Metric:
    """Decorate a metric function from sklearn.

    Converts a metric function that uses the typical sklearn metric signature so that
    it is compatible with :py:func:`train`

    """

    def inner(y_score: np.ndarray, dmatrix: DMatrix) -> Tuple[str, float]:
        y_true = dmatrix.get_label()
        weight = dmatrix.get_weight()
        if weight.size == 0:
            return func.__name__, func(y_true, y_score)
        else:
            return func.__name__, func(y_true, y_score, sample_weight=weight)

    return inner
```
and add a test.

trivialfis (commented):
Could you please help add a test in https://github.com/dmlc/xgboost/blob/master/tests/python/test_with_sklearn.py ?

BenEfrati:
> Could you please help add a test in https://github.com/dmlc/xgboost/blob/master/tests/python/test_with_sklearn.py ?

Sure

trivialfis (approved):
Thank you for the nice feature!

PR #8734, opened 2023-01-29
[backport] Make sure input numpy array is aligned. (#8690)

trivialfis:
- use `np.require` to specify that the alignment is required.
- scipy csr as well.
- validate input pointer in `ArrayInterface`.


trivialfis:
that's interesting, misalignment with distributed GPU inputs.

PR #8741, opened 2023-01-30
Use `bash -l {0}` as the default so you don't have to repeat it in GitHub Actions

hamelsmu:
use `bash -l {0}` as the default so you don't have to repeat it

Documentation: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaults


trivialfis (approved):
Thank you for sharing this nice cleanup!

PR #8742, opened 2023-01-31
Support cpu quantile sketch with column-wise data split

rongou:
When doing quantile sketch with column-wise split data, we only need to sketch columns available locally, there is no need to synchronize the cuts across all the workers. Later on when we evaluate the splits, we also only need to evaluate on locally available columns, and then synchronize across the workers to find the global best split.

This PR also changes the `DMatrix::SliceCol` method slightly since we always split the columns into n (roughly) equal parts, so that we don't have to repeat the slice index calculation everywhere.

### DIFF
--- a/src/common/quantile.cc
+++ b/src/common/quantile.cc
@@ -139,80 +139,6 @@ struct QuantileAllreduce {
     return worker_values.subspan(feat_beg, feat_size);
   }
 };
-
-/**
- * \brief Merge all categories from other workers.
- */
-void AllreduceCategories(Span<FeatureType const> feature_types, int32_t n_threads,
-                         std::vector<std::set<float>> *p_categories) {
-  auto &categories = *p_categories;
-  auto world_size = collective::GetWorldSize();
-  auto rank = collective::GetRank();
-  if (world_size == 1) {
-    return;
-  }
-
-  // CSC indptr to each feature
-  std::vector<size_t> feature_ptr(categories.size() + 1, 0);
-  for (size_t i = 0; i < categories.size(); ++i) {
-    auto const &feat = categories[i];
-    feature_ptr[i + 1] = feat.size();
-  }
-  std::partial_sum(feature_ptr.begin(), feature_ptr.end(), feature_ptr.begin());
-  CHECK_EQ(feature_ptr.front(), 0);
-
-  // gather all feature ptrs from workers
-  std::vector<size_t> global_feat_ptrs(feature_ptr.size() * world_size, 0);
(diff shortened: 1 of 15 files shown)


rongou:
@trivialfis @hcho3

src/common/quantile.cc
```diff
@@ -390,7 +396,7 @@ void SketchContainerImpl<WQSketch>::MakeCuts(HistogramCuts* cuts) {
   cuts->min_vals_.HostVector().resize(sketches_.size(), 0.0f);
   std::vector<typename WQSketch::SummaryContainer> final_summaries(reduced.size());
 
-  ParallelFor(reduced.size(), n_threads_, Sched::Guided(), [&](size_t fidx) {
```
rongou, line 393: It's weird, the guided schedule is causing those test failures. Switching to auto fixes them. Maybe some bug in OpenMP?

rongou: Hmm, never mind, looks like auto also fails sometimes. Probably some kind of race condition in this parallel for.

rongou: Ok, finally found the root cause.

rongou:
@trivialfis @hcho3 The CI is mostly passing now, the two failures are probably not related to this PR. Please take a look. Thanks!

src/common/quantile.cc
```diff
@@ -283,7 +285,9 @@ void SketchContainerImpl<WQSketch>::AllReduce(
   collective::Allreduce<collective::Operation::kMax>(&n_columns, 1);
   CHECK_EQ(n_columns, sketches_.size()) << "Number of columns differs across workers";
 
-  AllreduceCategories(feature_types_, n_threads_, &categories_);
+  if (!col_split_) {
```
trivialfis, line 288: Maybe we should `LOG(FATAL)` in all reduce categories if this is a column split matrix?

trivialfis (commented):
overall looks good to me. one small question inlined in comments.

src/common/quantile.cc
rongou: `AllreduceCategories` is a freestanding function that doesn't have access to `this->col_split_`. Maybe we should make it a member function?

rongou: Made it a private member function. PTAL

trivialfis (approved):
Looks good to me! Please merge from the latest master branch to resolve CI issue.

rongou:
Not sure why the dask test is failing. I can't reproduce it locally. But it does look a little suspicious since it's `test_quantile_dmatrix`. Will try to dig a bit more.

trivialfis:
It's unrelated. I have a fix.

Issue #8090, opened 2022-07-17
[python] 'setup.py install' is deprecated


jameslamb:
## Description

XGBoost's Python package relies on subclassing some of the commands from `setuptools`, with XGBoost-specific logic to build a shared library.

https://github.com/dmlc/xgboost/blob/e28f6f66577eb31451e5d3fbf4be0a21626f7372/python-package/setup.py#L343-L348

This approach, and specifically support for invocations like `python setup.py install`, is considered deprecated in `setuptools` and may be removed in the future.

See, for example, any recent run of the `XGBoost-Python-Wheels` CI job ([example build link](https://github.com/dmlc/xgboost/runs/7374372443?check_suite_focus=true)).

You'll find a warning like the following.

> python3.8/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.

## Specific Request

I'm not really sure. I noticed this in LightGBM (https://github.com/microsoft/LightGBM/issues/5061), and wanted to document it here as well.

I might attempt to resolve this in LightGBM by trying to switch to [`scikit-build`](https://scikit-build.readthedocs.io/en/latest/), but that isn't something we've worked on or even really discussed in detail yet (https://github.com/microsoft/LightGBM/issues/5061#issuecomment-1066276131).

## References

See much more documentation on this in the description and comments at https://github.com/microsoft/LightGBM/issues/5061.

trivialfis:
Thank you for raising the issue!  Let me take a closer look into scikit-build.

jameslamb:
I just got back from SciPy 2022 and there was a lot of excitement about it there. This proposal just received a grant: https://iscinumpy.gitlab.io/post/scikit-build-proposal/.

Anyway yeah, happy to help! If we make any progress on it over in LightGBM updates will be posted at https://github.com/microsoft/LightGBM/issues/5061.

As for the urgency of this...I believe that this PEP 632 being accepted ([link](https://peps.python.org/pep-0632/)) means that projects have until Python 3.12 to resolve deprecation warnings like the one mentioned in this issue's write-up.

PR #8748, opened 2023-02-01
Require black formatter for the python package.

trivialfis:
We have been using the black formatter for selected files to enable the requirement incrementally. However, it's difficult to enforce new code and the diff in new PRs often contains unnecessary formatting changes due to regional formatting.

This PR enables the black formatter on the whole package so that we can move on.

The change brought by this PR is relatively small since we have been working on aligning the code to the black formatter for a while. The most significantly impacted file is setup.py, but I think it's fine since we might be moving away from it according to https://github.com/dmlc/xgboost/issues/8090 . I hope we can make the change to ease future development. ;-)

### DIFF
--- a/python-package/setup.py
+++ b/python-package/setup.py
@@ -19,17 +19,17 @@ sys.path.insert(0, CURRENT_DIR)
 # requires using CMake directly.
 USER_OPTIONS = {
     # libxgboost options.
-    'use-openmp': (None, 'Build with OpenMP support.', 1),
-    'use-cuda':   (None, 'Build with GPU acceleration.', 0),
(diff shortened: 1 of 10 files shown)


hcho3:
I hope to make progress on #8694. I've been using pre-commit for cuML and RAFT, and it's super convenient to fix formatting issues while running `git commit`.

trivialfis:
@hcho3 that looks nice. we also have linter on the CI, which might be helpful.

The PR is to enforce the linter on the core package, which will be checked by the CI.

RAMitchell (approved):

PR #8749, opened 2023-02-01
Fix GPU L1 error.

trivialfis:
The getter was accidentally removed in 90cce3823.

Also, partially revert the optimization in the prediction cache. The prediction cache is violated after leaf value is updated by the objective function.


trivialfis:
working on this, need anther fix in prediction cache.

Issue #1433, opened 2016-08-02
Quantile Regression and Support for Prediction Intervals


nnorton24:
I know that sklearn.ensemble.GradientBoostingRegressor supports quantile regression and the production of [prediction intervals](http://scikit-learn.org/stable/auto_examples/ensemble/plot_gradient_boosting_quantile.html). Are there any plans for the XGBoost package to offer similar support?

hetong007:
Quantile regression is currently not supported. 

It can be achieved by defining the objective function on user's side:  https://github.com/dmlc/xgboost/blob/master/R-package/demo/custom_objective.R

arbackus:
Quantile regression with xgboost requires custom gradient and hessian functions. Here is an implementation in Python: http://www.bigdatarepublic.nl/regression-prediction-intervals-with-xgboost/

joseortiz3:
Note [that](http://www.bigdatarepublic.nl/regression-prediction-intervals-with-xgboost/) implementation is not very useful for most users, since it specifies grid-searching three parameters (very costly) to get a quantile estimate. In sklearn, you define only the quantile value and you are given a very robust and reliable estimate of that quantile.

Sandy4321:
any progress ?

link above is broken 
http://www.bigdatarepublic.nl/regression-prediction-intervals-with-xgboost/

trivialfis:
Related https://github.com/unit8co/darts/issues/1134 .

BraulioPi:
??

trivialfis:
Suppported by the Python package, and R package (still in master).

PR #8750, opened 2023-02-01
Initial support for quantile loss.

trivialfis:
- Add quantile obj and metric.

Related https://github.com/dmlc/xgboost/issues/1433 .

I will keep that issue open after the PR is finished (assuming there's no major issue in this PR). Some more work are still needed for support in other interfaces.


src/learner.cc
```diff
@@ -786,6 +787,10 @@ class LearnerConfiguration : public Learner {
     auto& args = *p_args;
     args = {cfg_.cbegin(), cfg_.cend()};  // renew
     obj_->Configure(args);
+
```
trivialfis, line 790: Remove this

RAMitchell (approved):

Issue #8744, opened 2023-01-31
Support for individual tree predictions mapped to label space


andermi:
As described [here](https://stackoverflow.com/questions/74062129/how-to-get-individual-trees-prediction-value-for-xgboost-regressor),
please add documentation (or add an api function) for how to get each tree's prediction in label space for the regressor in order to do residual analysis with the labels per tree

trivialfis:
Hi, I opened https://github.com/dmlc/xgboost/pull/8752 , hope that helps!

PR #8752, opened 2023-02-02
[doc] Add demo for inference using individual tree.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/8744 .


andermi:
Thank you for providing this! Would this method also work for using `reg:squarederror` rather than logistic? And in that case, what might I use in place of `logit` to get raw leaf values?

trivialfis:
@andermi You don't need to do anything for squarederror, not even the base margin and output margin in this demo. The leaf value is the output value and they can be summed in Python if you want.

RAMitchell (approved):

PR #8761, opened 2023-02-07
Add quantile metric.

trivialfis:
At the moment, the metric doesn't return the quantile value as part of its name like `quantile@0.5` (metrics like `NDCG` would display `ndcg@3`). It's just `quantile`. Eventually, we need a better interface for metrics to support non-trivial distance measures. This PR only handles the basic implementation of mean pinball loss.

- Add quantile loss as metric, currently simply named `quantile`. We can call it something more "formal" like mean_pinball_error if it's deemed better.
- Add quantile parameter, which requires parsing vector of floats. We have similar structures for tree param with mono and interaction constraints, but for integer instead.

Extracted from https://github.com/dmlc/xgboost/pull/8750 .

### DIFF
--- a/R-package/src/Makevars.in
+++ b/R-package/src/Makevars.in
@@ -91,6 +91,7 @@ OBJECTS= \
     $(PKGROOT)/src/common/survival_util.o \
     $(PKGROOT)/src/common/threading_utils.o \
     $(PKGROOT)/src/common/ranking_utils.o \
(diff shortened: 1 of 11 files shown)


RAMitchell (changes requested):
Can we do a small ground truth test against sklearn?

trivialfis:
Added a fix to parameter validation.

RAMitchell (approved):

Issue #8708, opened 2023-01-25
XGBRanker with sample_weight makes python interpreter died and restarted.


xbanke:
Sorry for submitting this issue. I really donot know how to solve this problem.
As the doc said, in ranking jobs, the `sample_weight` is assigned to each group but each data point, so I think the shape of sample_weight should be the same with `group`. But this make python iterpreter died and auto restarted when I run `fit`.
With some tracking, the problem maybe happend to the function `self._create_dmatrix`. It looks like that the shape of `sample_weight` needed to be the same with `y`.
Also, I tried with the shape of sample_weight equal to labels `y`, it raises `XGBoostError: [13:02:51] ../src/data/data.cc:683: Check failed: group_ptr_.size() == weights_.Size() + 1 (34 vs. 1001) : Size of weights must equal to number of groups when ranking group is used.`

```py
# python version: 3.9
# xgboost version: 1.7.3

import numpy as np
import pandas as pd
import xgboost as xgb

N = 1000
dates = pd.date_range('20100101', periods=N)
X = pd.DataFrame(np.random.randn(N, 5), columns=list('ABCDE'), index=dates)
y = pd.Series(np.random.randint(0, 10, size=N), index=dates, name='label')
group = X.groupby(dates + pd.offsets.MonthEnd(0)).size()
sample_weight = pd.Series(np.arange(len(group)), index=group.index)  # shape equal to group

ranker = xgb.XGBRanker(objective='rank:pairwise', max_depth=3, learning_rate=0.1, booster='gbtree', tree_method='hist', n_jobs=4, n_estimators=50, enable_categorical=False, random_state=42)

ranker.fit(X=X, y=y, group=group, sample_weight=sample_weight, verbose=True)  # it died here without any error info

```

xbanke:
when to release the next version?

trivialfis:
Soon.

PR #8762, opened 2023-02-07
Fix ranking with quantile dmatrix and group weight.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/8708 .

- Fix mixing QDM with device and host inputs.
- Fix group weight for host QDM.

### DIFF
--- a/src/data/iterative_dmatrix.cc
+++ b/src/data/iterative_dmatrix.cc
@@ -63,6 +63,13 @@ void GetCutsFromRef(std::shared_ptr<DMatrix> ref_, bst_feature_t n_features, Bat
     }
   };
   auto ellpack = [&]() {
+    // workaround ellpack being initialized from CPU.
+    if (p.gpu_id == Context::kCpuId) {
+      p.gpu_id = ref_->Ctx()->gpu_id;
+    }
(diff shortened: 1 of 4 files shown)


src/data/iterative_dmatrix.cc
```diff
@@ -63,6 +63,13 @@ void GetCutsFromRef(std::shared_ptr<DMatrix> ref_, bst_feature_t n_features, Bat
     }
   };
   auto ellpack = [&]() {
+    // workaround ellpack being initialized from CPU.
+    if (p.gpu_id == Context::kCpuId) {
```
RAMitchell, line 67: I don't understand what is happening here, and it looks weird with two statements having the same condition.

trivialfis: If the device is on CPU, we try the device ordinal in reference dmatrix. After the assignment from the reference dmatrix, if the device is still on cpu, we just use 0 as a workaround.

This happens when both the current qdm and the reference dmatrix are initialized on CPU but we train the booster on GPU.

I'm working to unify the device ordinal, we will have a single context from booster instead. This PR is quick fix for backport.

RAMitchell (approved):

Issue #8766, opened 2023-02-09
[spark] Set DMatrix `enable_categorical` to True if user provide `feature_types` param


WeichenXu123:
If user provide `feature_types` param, we should set `enable_categorical` to True when constructing DMatrix.

WeichenXu123:
@trivialfis Is my understanding correct ? I will file PR for it once you confirmed it.

trivialfis:
Correct. Please submit a PR.

WeichenXu123:
@trivialfis Thank you!

PR #8772, opened 2023-02-09
Fix feature types param

WeichenXu123:
Fix feature types param:
If user provides feature_types param, we should set 'enable_categorical' param as `True` when constructing `DMatrix`.

closes https://github.com/dmlc/xgboost/issues/8766

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -147,7 +147,7 @@ _unsupported_predict_params = {
 _unsupported_params_hint_message = {
     "enable_categorical":
         "`xgboost.spark` estimators do not have 'enable_categorical' param, but you can "
-        "set `feature_types` params and mark categorical feature with 'c' string. "
+        "set `feature_types` param and mark categorical features with 'c' string. "
 }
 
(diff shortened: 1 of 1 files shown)


WeichenXu123:
CC @trivialfis 

Can we backport this fix to branch 1.7 ?

trivialfis:
@WeichenXu123 Yeah, small enough. Let me know when it's ready.

WeichenXu123:
@trivialfis Yes it's ready for review. Just some minor linter error to fix. 

Q: How to add a test ? Can we check whether a DMatrix object that is constructed with `enable_categorical=True` ?

trivialfis:
Do we have test for categorical data?

WeichenXu123:
> Do we have test for categorical data?

We don't have it for now. I can add one.

trivialfis (commented):
Looks good to me! Please fix linter errors.

Considering that we have no test for categorical data and it's experimental, I think we can push the test to the next PR. Otherwise it would be difficult to backport since we have changed a lots of tests in spark.

trivialfis (approved):

PR #8773, opened 2023-02-09
fix inconsistent doc and minor typo for external memory

jaywu109:
Current [doc](https://xgboost.readthedocs.io/en/stable/tutorials/external_memory.html) is using `input_data` without **keyword arguments**.
https://github.com/dmlc/xgboost/blob/199c421d60d375eda169cd1715f5beba95b5b86d/doc/tutorials/external_memory.rst#L49
As shown below, it mentioned it's simplified version of `demo/guide-python/external_memory.py`
https://github.com/dmlc/xgboost/blob/199c421d60d375eda169cd1715f5beba95b5b86d/doc/tutorials/external_memory.rst#L66
but it's actually different from the file mentioned as follows:
https://github.com/dmlc/xgboost/blob/199c421d60d375eda169cd1715f5beba95b5b86d/demo/guide-python/external_memory.py#L67

I've found errors may occur like the one in #8202 if following the current doc. Not sure if it's better to update the doc to make it consistent with the `demo/guide-python/external_memory.py` to avoid any possible error for the users.

### DIFF
--- a/doc/tutorials/external_memory.rst
+++ b/doc/tutorials/external_memory.rst
@@ -63,7 +63,7 @@ constructor.
   booster = xgboost.train({"tree_method": "approx"}, Xy)
 
 
-The above snippet is a simplifed version of ``demo/guide-python/external_memory.py``.  For
+The above snippet is a simplified version of ``demo/guide-python/external_memory.py``.  For
 an example in C, please see ``demo/c-api/external-memory/``.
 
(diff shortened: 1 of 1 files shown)


trivialfis (approved):
Thank you for the fix!

FYI, the function requires keyword arguments in the latest xgboost (master branch) now.

PR #8783, opened 2023-02-10
Generalize prediction cache.

trivialfis:
Generalize prediction cache for other uses.

* Extract most of the functionality into `DMatrixCache`.
* Limit the size of the prediction cache to 32 items. This should be plenty. When the number is exceeded, it's usually because the user forgets to free the input DMatrix during inference, and by limiting the size of the cache, we are actually improving performance for those cases.
* Move API entry struct to an independent file to reduce dependency on the `predictor.h` file.
* Add a test.

I'm working on learning to rank related changes. One hurdle I run into is how to cache sorted indexes inside Metric. Currently, the cox metric adds a member function to meta info, while the AUC has an ad-hoc cache. This PR proposes we reuse the prediction cache in these places.

### DIFF
--- a/include/xgboost/cache.h
+++ b/include/xgboost/cache.h
@@ -4,13 +4,13 @@
 #ifndef XGBOOST_CACHE_H_
 #define XGBOOST_CACHE_H_
 
-#include <xgboost/logging.h>
+#include <xgboost/logging.h>  // CHECK_EQ
 
-#include <cstddef>        // std::size_t
-#include <memory>         // std::weak_ptr,std::shared_ptr,std::make_shared
-#include <queue>          // std:queue
-#include <unordered_map>  // std::unordered_map
-#include <vector>         // std::vector
+#include <cstddef>            // std::size_t
+#include <memory>             // std::weak_ptr,std::shared_ptr,std::make_shared
+#include <queue>              // std:queue
+#include <unordered_map>      // std::unordered_map
+#include <vector>             // std::vector
 
 namespace xgboost {
 class DMatrix;
@@ -70,16 +70,20 @@ class DMatrixCache {
     this->CheckConsistent();
   }
 
-  void Clear() {
+  void ClearExcess() {
     this->CheckConsistent();
     while (queue_.size() >= max_size_) {
       auto p_fmat = queue_.front();
       queue_.pop();
       container_.erase(p_fmat);
     }
+    this->CheckConsistent();
   }
 
  public:
+  /**
+   * \param cache_size Maximum size of the cache.
+   */
   explicit DMatrixCache(std::size_t cache_size) : max_size_{cache_size} {}
   /**
    * \brief Cache a new DMatrix if it's no in the cache already.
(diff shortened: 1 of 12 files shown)


include/xgboost/cache.h
```diff
@@ -0,0 +1,130 @@
...
+    }
+    while (!remained.empty()) {
+      auto p_fmat = remained.front();
+      queue_.push(p_fmat);
+      remained.pop();
+    }
+    this->CheckConsistent();
+  }
+
+  void Clear() {
```
hcho3, line 73:
```suggestion
  void ClearExcess() {
```
To make it clear that only entries that exceed the `max_size_` limit would be deleted.

include/xgboost/cache.h
```diff
@@ -0,0 +1,130 @@
...
+   *        returning.
+   */
+  decltype(container_) const& Container() {
+    this->ClearExpired();
+    return container_;
+  }
+
+  std::shared_ptr<CacheT> Entry(DMatrix const* m) const {
+    CHECK(container_.find(m) != container_.cend());
+    CHECK(container_.at(m).ref.lock());
```
hcho3, line 125:
```suggestion
    CHECK(!container_.at(m).ref.expired());
```
The `lock()` operation involves an atomic operation, according to https://en.cppreference.com/w/cpp/memory/weak_ptr/lock. Checking `expired()` may be more efficient.

include/xgboost/predictor.h
```diff
@@ -1,95 +1,64 @@
...
-  PredictionCacheEntry& Entry(DMatrix* m);
-  /* \brief Get a const reference to the underlying hash map.  Clear expired caches before
-   *        returning.
-   */
-  decltype(container_) const& Container();
+  PredictionContainer() : DMatrixCache<PredictionCacheEntry>{DefaultSize()} {}
+  PredictionCacheEntry& Cache(std::shared_ptr<DMatrix> m, int32_t device) {
+    this->CacheItem(m);
+    auto p_cache = this->container_.find(m.get());
+    p_cache->second.Value().predictions.SetDevice(device);
```
hcho3, line 59:
```suggestion
    if (device != Context::kCpuId) {
      p_cache->second.Value().predictions.SetDevice(device);
    }
```
We shouldn't call `SetDevice` when the device is a CPU.

tests/cpp/test_cache.cc
```diff
@@ -0,0 +1,52 @@
...
+  explicit CacheForTest(std::size_t k) : i{k} {}
+};
+}  // namespace
+
+TEST(DMatrixCache, Basic) {
+  std::size_t constexpr kRows = 2, kCols = 1, kCacheSize = 4;
+  DMatrixCache<CacheForTest> cache(kCacheSize);
+
+  auto add_cache = [&]() {
+    auto p_fmat = RandomDataGenerator(kRows, kCols, 0).GenerateDMatrix();
```
hcho3, line 26:
```suggestion
    // Create a lambda function here, so that p_fmat gets deleted upon the
    // end of the lambda. This is to test how the cache handle expired
    // cache entries.
    auto p_fmat = RandomDataGenerator(kRows, kCols, 0).GenerateDMatrix();
```

hcho3 (changes requested):

trivialfis:
@hcho3 Thank you for the review, all comments are addressed.

hcho3 (approved):

Issue #8774, opened 2023-02-09
XGBoost eval_metric and prediction results are different


aParsecFromFuture:
```
def acc(y_true, y_pred):
        return accuracy_score(y_true, y_pred.round())

  params = {
      'objective': 'binary:logistic',
      'enable_categorical': True,
      'tree_method': 'hist',
      'eval_metric': acc
  }

  model = xgb.XGBClassifier(**params)

  model.fit(X_train, y_train, 
            eval_set=[(X_valid, y_valid)])
  
  y_pred = model.predict(X_valid)
  score = accuracy_score(y_valid, y_pred)
  print(score)
```

Accuracy score should be same score of the last iteration but I get 73% and 69%. Validation data has 10k rows. Any idea?

trivialfis:
I can't reproduce using the latest xgboost. Could you please share a reproducible example?

aParsecFromFuture:
I get the same behavior with built-in error metric. (1 - error) is not equal to accuracy score.

Class distribution: (0: 32496, 1: 20523)

<img width="678" alt="Screenshot 2023-02-14 at 22 31 37" src="https://user-images.githubusercontent.com/30957924/218838491-81a0ac93-1b23-4e84-99b6-d23c315ab2eb.png">

trivialfis:
Let me get back to this tomorrow. I have an account with the same email address on my github profile.

aParsecFromFuture:
> Let me get back to this tomorrow. I have an account with the same email address on my github profile.

I shared the notebook with trivialfis user. Good luck.

trivialfis:
```
This is a limited-participation competition. Only invited users may participate.
```

aParsecFromFuture:
> This is a limited-participation competition. Only invited users may participate.

I'm going to open it publicly for today. Let me know when your review is done.
Link: https://www.kaggle.com/greysky/temporary-notebook-it-will-be-deleted

trivialfis:
thank you for sharing, I can see the notebook, it's the dataset that I can't access. I need to participate in the competition before seeing the dataset, but the competition is invite only.
![Screenshot from 2023-02-15 16-05-22](https://user-images.githubusercontent.com/16746409/218969104-1c188e92-039e-4002-aeb5-ab3c6187cbf0.png)

aParsecFromFuture:
> thank you for sharing, I can see the notebook, it's the dataset that I can't access. I need to participate in the competition before seeing the dataset, but the competition is invite only.
[...]

PR #8809, opened 2023-02-15
Fix CPU bin compression with categorical data.

trivialfis:
* The bug causes maximum category to be lesser than 256 or the maximum number of bins when the input data is dense.

Close https://github.com/dmlc/xgboost/issues/8774


trivialfis:
@hcho3 Could you please take a look when you are available? This should be the one last fix for 1.7.4. ;-)

hcho3 (approved):

PR #8811, opened 2023-02-15
Support column split in histogram builder

rongou:
When data is split column-wise, we don't need to sync the histograms across workers, since each worker has a distinct set of features. We can evaluate splits locally and find the globally best split later.

Also added a convenience method `IsColumnSplit` in `DMatrix`.


rongou:
@trivialfis @hcho3

trivialfis (approved):

Issue #6955, opened 2021-05-12
Ranking prediction are all same value 


1a1a11a:
I am using the XGBoost C API to do some ranking tasks and I have tried to use `rank:pairwise`, `rank:map`, and `rank:ndcg` as the objective, however, when I use `rank:map`, and `rank:ndcg` to train and predict, all the predicted values are constant, but it is not a problem when I use `rank:pairwise`. Does anyone know what could be wrong in my setup? Or is there a better way other than digging into the source code? Thank you!

1a1a11a:
I have checked that the relevance in the training data are all from 0 to 1, and the inference data are not constant. I think the training has some problems, the metric does not change at all from the first iteration to 10 iterations (while it changes if I use 
regression or rank:pairwise.

creat89:
For NDCG, and I'm guessing MAP too, the relevance scores must be integers. See:

https://github.com/dmlc/xgboost/issues/6352#issuecomment-735595924

or:

https://discuss.xgboost.ai/t/very-large-ndcg-result/1712

Issue #6709, opened 2021-02-15
Mismatch OS


gprovencher-cdpq:
Hi,

I observe that the rankings obtained using XGBRanker are not the same whether I use Windows or Linux. (I use xgboost 1.3.3)

`from xgboost import XGBRanker
import numpy as np
import pandas as pd

# Training set.
seed = 1
np.random.seed(seed)
input_df = pd.DataFrame({
    "feature_1": np.random.rand(1000),
    "feature_2": np.random.rand(1000)
})

# Target column has a simple relation with respect to the feature columns.
input_df["target"] = input_df["feature_1"] * input_df["feature_2"]
 
feature_columns = ["feature_1", "feature_2"]
target_columns = "target"
    
estimator = XGBRanker(
        base_score=0.5, booster="gbtree", colsample_bylevel=1.0,
        colsample_bynode=1.0, colsample_bytree=1.0, gamma=0.5, gpu_id=-1,
        importance_type='gain', interaction_constraints='',
        learning_rate=0.3, max_delta_step=1, max_depth=3,
        min_child_weight=5, missing=np.nan, monotone_constraints='()',
        n_estimators=50, n_jobs=1, num_parallel_tree=1,
        random_state=0, reg_alpha=0, reg_lambda=1.0,
        scale_pos_weight=None, subsample=1.0, tree_method="exact",
        validate_parameters=False, verbosity=0, nthread=1
)
 
X = input_df[feature_columns].copy()
y = input_df[target_columns].copy()
group = np.array([len(X)])
 
estimator.fit(X=X, y=y, group=group, sample_weight=None, base_margin=None, 
              eval_set=None, sample_weight_eval_set=None, 
              eval_group=None, eval_metric=None, 
              early_stopping_rounds=None, verbose=False, xgb_model=None, 
              feature_weights=None, callbacks=None)

print("Feature importance:", estimator.feature_importances_)

# Dataset for predictions.
np.random.seed(seed + 1)
to_predict_df = pd.DataFrame({
    "feature_1": np.random.rand(1000),
    "feature_2": np.random.rand(1000)
})
to_predict_df["result"] = estimator.predict(data=to_predict_df)

to_predict_df["answer"] = to_predict_df["feature_1"] * to_predict_df["feature_2"]
to_predict_df["answer_rank"] = to_predict_df["answer"].rank().astype(int)
to_predict_df["result_rank"] = to_predict_df["result"].rank().astype(int)
[...]

Issue #6707, opened 2021-02-15
ndcg@k metric is inaccurate


PascalIversen:
Hey,
I am using xgboost version 1.3.3 and I am getting invalid ndcg@k metric values. The values are >1 while the ndcg@k has to be in [0,1] by definition.

Example:
```python
# creating dummy ranking data
import numpy as np
groups = [60, 50, 32, 22, 22 , 22 , 20, 12]
labels = []
features = []
for group in groups:
  label_group = list(np.argsort(np.argsort(np.random.randn(group))))
  labels = labels + label_group
  for label in label_group:
    features.append([label+10, np.random.randn(), np.random.randn()]) 
features = np.stack(features)


# fitting a ranking model

import xgboost as xgb
import pandas

model = xgb.XGBRanker(
        objective="rank:pairwise",
        max_depth= 10,
        n_estimators=1000,
         verbosity=2)

ranker = model.fit(X=features,
        y=labels, group=groups,
        eval_set=[(features, labels)],
        eval_group=[groups], eval_metric="ndcg@10",  early_stopping_rounds=2, verbose=True)
```

```
[16:15:01] INFO: ../src/tree/updater_prune.cc:101: tree pruning end, 146 extra nodes, 0 pruned nodes, max_depth=10
[0]	validation_0-ndcg@10:1.32964
[16:15:01] INFO: ../src/tree/updater_prune.cc:101: tree pruning end, 112 extra nodes, 0 pruned nodes, max_depth=10
[1]	validation_0-ndcg@10:112.34529
[16:15:01] INFO: ../src/tree/updater_prune.cc:101: tree pruning end, 130 extra nodes, 0 pruned nodes, max_depth=10
[2]	validation_0-ndcg@10:1.69752
[16:15:01] INFO: ../src/tree/updater_prune.cc:101: tree pruning end, 92 extra nodes, 0 pruned nodes, max_depth=10
```

lcschv:
Hey, I have faced the same issue. After some investigation I think the issue is that the computation of the ideal ranking is wrong.

For instance, let a query _q_, and a set of relevant documents for this query RelevantDocs = {d1, d2, d3}. Lets assume that the relevance grades of each document is the following: {d1=2, d2=1, d3=1}.

In order to compute the ideal ranking, we have to sort the Relevant Documents in the best way possible (based only on the set of relevant documents).

So, in this case, the ideal ranking is
[...]

Issue #6352, opened 2020-11-08
rank:ndcg does not work in Python


PascalIversen:
Hey,

`rank:ndcg` produces constant results, while `rank:pairwise` works as expected.

```python
import numpy as np
import pandas as pd
import random
import xgboost as xgb

# creating some artificial ranking data

gsizes = [10, 100, 50]
params = [1, 2, 3]
relevances = []
features_1 = []
features_2 = []
qids = []
for gsize, param in zip(gsizes, params):
    
    range_ = list(range(gsize))
    random.shuffle(range_)
    
    relevances.extend(np.abs(np.random.randn(gsize)) + param*np.array(range_))
    feature_1 = np.array(range_)*param*2+14 + np.random.randn(gsize)
    feature_2 = np.array(range_)*param**2 + np.random.randn(gsize)**2
    features_1.extend(feature_1)
    features_2.extend(feature_2)
    qids.extend([param]*gsize)

ranker = xgb.XGBRanker(objective='rank:ndcg',
                       base_score=0.1,
                       verbosity = 1,
                       n_estimators=200)
inputs = np.array([features_1, features_2, qids]).T
fitted = ranker.fit(X=inputs , y=relevances, group=gsizes)
fitted.predict(inputs)[:10]

array([-0.9008925, -0.9008925, -0.9008925, -0.9008925, -0.9008925,
       -0.9008925, -0.9008925, -0.9008925, -0.9008925, -0.9008925],
      dtype=float32)
```

The results when I use pairwise:
```python
fitted.predict(inputs)[:10]
array([-4.8152976 , -1.3937551 ,  0.36378622,  1.7936848 , -2.8243895 ,
       -9.020457  , -7.9117856 ,  3.2185328 , -7.152711  , -3.2654085 ],
      dtype=float32)
```
which has an NDCG of 1.

Environment:
Python 3.7.6
xgboost 1.2.0

Best,

Pascal

trivialfis:
Thanks for submitting the issue.  Let me take a look tomorrow.  The implementation involves sampling to reduce time complexity so might take some time to debug.

trivialfis:
Had a brief look into it.  After weighting the gradient all split gains become negative in this example.  Will continue looking to ensure this is not a bug.

PascalIversen:
[...]

PR #8822, opened 2023-02-17
Rework learning to rank.

trivialfis:
Early PR, would like to get some comments from the community as I'm also new to the LTR.

- Rework the lambdamart implementation for all three objectives.
- Default to NDCG.
- Try to implement the position debiasing described by https://github.com/dmlc/xgboost/issues/6143 .
- Make GPU computation deterministic for both metric and objective.
- Add support for truncation.
- Check the input label.
- Fix ranking metrics with weights.
- Add documentation for all objectives.
- Support sklearn cross-validation.

## Debiasing
I'm new to learning to rank in general. It seems using simulated clicks is a common practice in unbiased LTR development. I tried to make a version of the click simulation using svmrank as described in the paper and showed in https://github.com/ULTR-Community/ULTRA.git , the one in the demo is just svmrank replaced by xgboost. I haven't really looked into the ULTRA repository, my simulation code might be incorrect.

I have tested the debiasing method with MSLR-WEB10K with simulated clicks, it leads to significant over-fitting.

As suggested by @thvasilo , we can have an additional field `position` for handling missing documents. Other than missing docs coming from practical reasons like corrupted user logs, we can have missing docs simply because they are preserved as the test set. I partially implemented the support and then removed it. The position should be in sync with the NDCG calculation, if we were to use the additional position info to calculate the bias, we should also use it to calculate the delta NDCG. This makes the code a bit more complicated and I'm not sure how useful it's in practice. I don't have access to any real user click data for testing. Would be great if I can compare the simulation result with a real dataset.

Lastly, I don't have access to the yahoo dataset, my request to that dataset was denied. The yahoo dataset was used in the paper as the only benchmarking dataset. Would be great if someone can help and compare the implementation with the results in the paper and the original implementation by the authors.

## Closes

Close https://github.com/dmlc/xgboost/issues/6143
Close https://github.com/dmlc/xgboost/issues/6955
Close https://github.com/dmlc/xgboost/issues/5561
Close https://github.com/dmlc/xgboost/issues/6709
Close https://github.com/dmlc/xgboost/issues/6707
Close https://github.com/dmlc/xgboost/issues/6352

## Related:
- RFC for revising LTR in XGBoost: https://github.com/dmlc/xgboost/issues/6450
- WIP LightGBM implementation of debiasing https://github.com/microsoft/LightGBM/pull/5393
- Original WIP LightGBM implementation: https://github.com/microsoft/LightGBM/pull/4531 .

## Results
[mq2007.csv](https://github.com/dmlc/xgboost/files/10925168/mq2007.csv)
[mq2008.csv](https://github.com/dmlc/xgboost/files/10925170/mq2008.csv)
[mslr10k.csv](https://github.com/dmlc/xgboost/files/10925172/mslr10k.csv)
[mslr30k.csv](https://github.com/dmlc/xgboost/files/10925175/mslr30k.csv)

### DIFF
--- a/doc/tutorials/learning_to_rank.rst
+++ b/doc/tutorials/learning_to_rank.rst
@@ -1,0 +1,199 @@
+################
+Learning to Rank
+################
+
+**Contents**
+
+.. contents::
+  :local:
+  :backlinks: none
+
+********
+Overview
+********
+Often in the context of information retrieval, learning-to-rank aims to train a model that arranges a set of query results into an ordered list `[1] <#references>`__. For surprivised learning-to-rank, the predictors are sample documents encoded as feature matrix, and the labels are relevance degree for each sample. Relevance degree can be multi-level (graded) or binary (relevant or not). The training samples are often grouped by their query index with each query group containing multiple query results.
+
+XGBoost implements learning to rank through a set of objective functions and performance metrics. The default objective is ``rank:ndcg`` based on the ``LambdaMART`` `[2] <#references>`__ algorithm, which in turn is an adaptation of the ``LambdaRank`` `[3] <#references>`__ framework to gradient boosting trees. For a history and a summary of the algorithm, see `[5] <#references>`__. The implementation in XGBoost features deterministic GPU computation, distributed training, position debiasing and two different pair construction strategies.
+
+************************************
+Training with the Pariwise Objective
+************************************
+``LambdaMART`` is a pairwise ranking model, meaning that it compares the relevance degree for every pair of samples in a query group and calculate a proxy gradient for each pair. The default objective ``rank:ndcg`` is using the surrogate gradient derived from the ``ndcg`` metric. To train a XGBoost model, we need an additional sorted array called ``qid`` for specifying the query group of input samples. An example input would look like this:
+
++-------+-----------+---------------+
+|   QID |   Label   |   Features    |
++=======+===========+===============+
+|   1   |   0       |   :math:`x_1` |
++-------+-----------+---------------+
+|   1   |   1       |   :math:`x_2` |
++-------+-----------+---------------+
+|   1   |   0       |   :math:`x_3` |
++-------+-----------+---------------+
+|   2   |   0       |   :math:`x_4` |
++-------+-----------+---------------+
+|   2   |   1       |   :math:`x_5` |
++-------+-----------+---------------+
+|   2   |   1       |   :math:`x_6` |
++-------+-----------+---------------+
+|   2   |   1       |   :math:`x_7` |
++-------+-----------+---------------+
+
+Notice that the samples are sorted based on their query index in a non-decreasing order. In the above example, the first three samples belong to the first query and the next four samples belong to the second. For the sake of simplicity, we will use a synthetic binary learning-to-rank dataset in the following code snippets, with binary labels representing whether the result is relevant or not, and randomly assign the query group index to each sample. For an example that uses a real world dataset, please see :ref:`sphx_glr_python_examples_learning_to_rank.py`.
+
+.. code-block:: python
+
+  from sklearn.datasets import make_classification
+  import numpy as np
+
+  import xgboost as xgb
+
+  # Make a synthetic ranking dataset for demonstration
+  X, y = make_classification(random_state=rng)
+  rng = np.random.default_rng(1994)
+  n_query_groups = 3
+  qid = rng.integers(0, 3, size=X.shape[0])
+
+  # Sort the inputs based on query index
+  sorted_idx = np.argsort(qid)
+  X = X[sorted_idx, :]
+  y = y[sorted_idx]
+
+The simpliest way to train a ranking model is by using the scikit-learn estimator interface. Continuing the previous snippet, we can train a simple ranking model without tuning:
+
+.. code-block:: python
+
+  ranker = xgb.XGBRanker(tree_method="hist", lambdarank_num_pair_per_sample=8, objective="rank:ndcg", lambdarank_pair_method="topk")
+  ranker.fit(X, y, qid=qid)
+
+Please note that, as of writing, there's no learning-to-rank interface in scikit-learn. As a result, the :py:class:`xgboost.XGBRanker` class does not fully conform the scikit-learn estimator guideline and can not be directly used with some of its utility functions. For instances, the ``auc_score`` and ``ndcg_score`` in scikit-learn don't consider query group information nor the pairwise loss. Most of the metrics are implemented as part of XGBoost, but to use scikit-learn utilities like :py:func:`sklearn.model_selection.cross_validation`, we need to make some adjustments in order to pass the ``qid`` as an additional parameter for :py:meth:`xgboost.XGBRanker.score`. Given a data frame ``X`` (either pandas or cuDF), add the column ``qid`` as follows:
+
+.. code-block:: python
+
+  df = pd.DataFrame(X, columns=[str(i) for i in range(X.shape[1]))
+  df["qid"] = qid
+  ranker.fit(df, y)  # No need to pass qid as a separate argument
+
+  from sklearn.model_selection import StratifiedGroupKFold, cross_val_score
+  # Works with cv in scikit-learn, along with HPO utilities like GridSearchCV
+  kfold = StratifiedGroupKFold(shuffle=False)
+  cross_val_score(ranker, df, y, cv=kfold, groups=df.qid)
+
+The above snippets build a model using ``LambdaMART`` with the ``NDCG@8`` metric. The outputs of a ranker are relevance scores:
+
+.. code-block:: python
+
+  scores = ranker.predict(X)
+  sorted_idx = np.argsort(scores)[::-1]
+  # Sort the relevance scores from most relevant to least relevant
+  scores = scores[sorted_idx]
+
+
+*************
+Position Bias
+*************
+
+.. versionadded:: 2.0.0
+
+.. note::
+
+   The feature is considered experimental. This is a heated research area, and your input is much appreciated!
+
+Obtaining real relevance degrees for query results is an expensive and strenuous, requiring human labelers to label all results one by one. When such labeling task is infeasible, we might want to train the learning-to-rank model on user click data instead, as it is relatively easy to collect. Another advantage of using click data directly is that it can reflect the most up-to-date user preferences `[1] <#references>`__. However, user clicks are often biased,  as users tend to choose results that are displayed in higher positions. User clicks are also noisy, where users might accidentally click on irrelevant documents. To ameliorate these issues, XGBoost implements the ``Unbiased LambdaMART`` `[4] <#references>`__ algorithm to debias the position-dependent click data. The feature can be enabled by the ``lambdarank_unbiased`` parameter; see :ref:`ltr-param` for related options and :ref:`sphx_glr_python_examples_learning_to_rank.py` for a worked example with simulated user clicks.
+
+****
+Loss
+****
+
+XGBoost implements different ``LambdaMART`` objectives based on different metrics. We list them here as a reference. Other than those used as objective function, XGBoost also implements metrics like ``pre`` (for precision) for evaluation. See :doc:`parameters </parameter>` for available options and the following sections for how to choose these objectives based of the amount of effective pairs.
+
+* NDCG
+
+`Normalized Discounted Cumulative Gain` ``NDCG`` can be used with both binary relevance and multi-level relevance. If you are not sure about your data, this metric can be used as the default. The name for the objective is ``rank:ndcg``.
+
+
+* MAP
+
+`Mean average precision` ``MAP`` is a binary measure. It can be used when the relevance label is 0 or 1. The name for the objective is ``rank:map``.
+
+
+* Pairwise
+
+The `LambdaMART` algorithm scales the logistic loss with learning to rank metrics like ``NDCG`` in the hope of including ranking information into the loss function. The ``rank:pairwise`` loss is the original version of the pairwise loss, also known as the `RankNet loss` `[7] <#references>`__ or the `pairwise logistic loss`. Unlike the ``rank:map`` and the ``rank:ndcg``, no scaling is applied (:math:`|\Delta Z_{ij}| = 1`).
+
+Whether scaling with a LTR metric is actually more effective is still up for debate; `[8] <#references>`__ provides a theoretical foundation for general lambda loss functions and some insights into the framework.
+
+******************
+Constructing Pairs
+******************
+
+There are two implemented strategies for constructing document pairs for :math:`\lambda`-gradient calculation. The first one is the ``mean`` method, another one is the ``topk`` method. The preferred strategy can be specified by the ``lambdarank_pair_method`` parameter.
+
+For the ``mean`` strategy, XGBoost samples ``lambdarank_num_pair_per_sample`` pairs for each document in a query list. For example, given a list of 3 documents and ``lambdarank_num_pair_per_sample`` is set to 2, XGBoost will randomly sample 6 pairs, assuming the labels for these documents are different. On the other hand, if the pair method is set to ``topk``, XGBoost constructs about :math:`k \times |query|` number of pairs with :math:`|query|` pairs for each sample at the top :math:`k = lambdarank\_num\_pair` position. The number of pairs counted here is an approximation since we skip pairs that have the same label.
+
+*********************
+Obtaining Good Result
+*********************
+
+Learning to rank is a sophisticated task and an active research area. It's not trivial to train a model that generalizes well. There are multiple loss functions available in XGBoost along with a set of hyperparameters. This section contains some hints for how to choose hyperparameters as a starting point. One can further optimize the model by tuning these hyperparameters.
+
+The first question would be how to choose an objective that matches the task at hand. If your input data has multi-level relevance degrees, then either ``rank:ndcg`` or ``rank:pairwise`` should be used. However, when the input has binary labels, we have multiple options based on the target metric. `[6] <#references>`__ provides some guidelines on this topic and users are encouraged to see the analysis done in their work. The choice should be based on the number of `effective pairs`, which refers to the number of pairs that can generate non-zero gradient and contribute to training. `LambdaMART` with ``MRR`` has the least amount of effective pairs as the :math:`\lambda`-gradient is only non-zero when the pair contains a non-relevant document ranked higher than the top relevant document. As a result, it's not implemented in XGBoost. Since ``NDCG`` is a multi-level metric, it usually generate more effective pairs than ``MAP``.
+
+However, when there are sufficiently many effective pairs, it's shown in `[6] <#references>`__ that matching the target metric with the objective is of significance. When the target metric is ``MAP`` and you are using a large dataset that can provide a sufficient amount of effective pairs, ``rank:map`` can in theory yield higher ``MAP`` value than ``rank:ndcg``.
+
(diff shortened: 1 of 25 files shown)


chen1st:
Now ndcg@n can only be an evaluation metrics,  why it can't be an objective.

trivialfis:
@chen1st that's part of the PR, to support truncation level.

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+  y = y[sorted_idx]
+
+The simpliest way to train a ranking model is by using the sklearn estimator interface. Continuing the previous snippet, we can train a simple ranking model without tuning:
+
+.. code-block:: python
+
+  ranker = xgb.XGBRanker(tree_method="hist", lambdarank_num_pair_per_sample=8, objective="rank:ndcg", lambdarank_pair_method="topk")
+  ranker.fit(X, y, qid=qid)
+
+Please note that, as of writing, there's no learning to rank interface in sklearn. As a result, the :py:class:`xgboost.XGBRanker` does not fully conform the sklearn estimator guideline and can not be directly used with some of its utility functions. For instances, the ``auc_score`` and ``ndcg_score`` in sklearn don't consider group information nor the pairwise loss. Most of the metrics are implemented as part of XGBoost, but to use sklearn utilities like :py:func:`sklearn.model_selection.cross_validation`, we need to make some adjustments in order to pass the `qid` as an additional parameter for :py:meth:`xgboost.XGBRanker.score`. The `X` for :py:class:`xgboost.XGBRanker` may contain a special column called ``qid`` when it's a pandas dataframe or a cuDF dataframe:
```
hcho3, line 68:
```suggestion
Please note that, as of writing, there's no learning-to-rank interface in scikit-learn. As a result, the :py:class:`xgboost.XGBRanker` class does not fully conform the scikit-learn estimator guideline and can not be directly used with some of its utility functions. For instances, the ``auc_score`` and ``ndcg_score`` in scikit-learn don't consider query group information nor the pairwise loss. Most of the metrics are implemented as part of XGBoost, but to use scikit-learn utilities like :py:func:`sklearn.model_selection.cross_validation`, we need to make some adjustments in order to pass the ``qid`` as an additional parameter for :py:meth:`xgboost.XGBRanker.score`. Given a data frame ``X`` (either pandas or cuDF), add the column ``qid`` as follows:
```
I revised the last sentence to make it a command, so that it follows clearly from the previous sentence.

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+
+.. code-block:: python
+
+  df = pd.DataFrame(X, columns=[str(i) for i in range(X.shape[1]))
+  df["qid"] = qid
+  ranker.fit(df, y)  # No need to pass qid as a separate argument
+
+  from sklearn.model_selection import StratifiedGroupKFold, cross_val_score
+  # Works with cv in sklearn, along with HPO utilities like grid search cv.
```
hcho3, line 77:
```suggestion
  # Works with cv in scikit-learn, along with HPO utilities like GridSearchCV
```

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+  scores = ranker.predict(X)
+  sorted_idx = np.argsort(scores)[::-1]
+  # Sort the relevance scores from most relevant to least relevant
+  scores = scores[sorted_idx]
+
+
+*************
+Position Bias
+*************
+Real relevance degree for query result is difficult to obtain as it often requires human judegs to examine the content of query results. When such labeled data is absent, we might want to train the model on ground truth data like user clicks. Another upside of using click data directly is that it can relect the up-to-date relevance status `[1] <#references>`__. However, user clicks are often nosiy and biased as users tend to choose results displayed in higher position. To ameliorate this issue, XGBoost implements the ``Unbiased LambdaMART`` `[4] <#references>`__ algorithm to debias the position-dependent click data. The feature can be enabled by the ``lambdarank_unbiased`` parameter, see :ref:`ltr-param` for related options and :ref:`sphx_glr_python_examples_learning_to_rank.py` for a worked example with simulated user clicks.
```
hcho3, line 94:
```suggestion
Obtaining real relevance degrees for query results is an expensive and strenuous task, as it requires human labelers to label all results one by one. When such labeling task is infeasible, we might want to train the learning-to-rank model on user click data instead, as it is relatively easy to collect. Another advantage of using click data directly is that it can reflect the most up-to-date user preferences `[1] <#references>`__. However, user clicks are often biased,  as users tend to choose results that are displayed in higher positions. User clicks are also noisy, where users accidentally click on irrelevant documents. To ameliorate this issue, XGBoost implements the ``Unbiased LambdaMART`` `[4] <#references>`__ algorithm to debias the position-dependent click data. The feature can be enabled by the ``lambdarank_unbiased`` parameter; see :ref:`ltr-param` for related options and :ref:`sphx_glr_python_examples_learning_to_rank.py` for a worked example with simulated user clicks.
```
Bunch of style improvements. Also treated bias and noise as distinct issues.

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+`Normalized Discounted Cumulative Gain` ``NDCG`` can be used with both binary relevance and multi-level relevance. If you are not sure about your data, this metric can be used as the default. The name for the objective is ``rank:ndcg``.
+
+
+* MAP
+`Mean average precision` ``MAP`` is a binary measure. It can be used when the relevance label is 0 or 1. The name for the objective is ``rank:map``.
+
+
+* Pairwise
+
+The `LambdaMART` algorithm scales the logistic loss with learning to rank metrics like ``NDCG`` in the hope of including ranking infomation into the loss function. The ``rank:pairwise`` loss is the orginal version of the pairwise loss, also known as the `RankNet loss` `[7] <#references>`__ or the `pairwise logistic loss`. Unlike the ``rank:map`` and the ``rank:ndcg``, no scaling is applied (:math:`|\Delta Z_{ij}| = 1`).
```
hcho3, line 112:
```suggestion
The `LambdaMART` algorithm scales the logistic loss with learning to rank metrics like ``NDCG`` in the hope of including ranking information into the loss function. The ``rank:pairwise`` loss is the original version of the pairwise loss, also known as the `RankNet loss` `[7] <#references>`__ or the `pairwise logistic loss`. Unlike the ``rank:map`` and the ``rank:ndcg``, no scaling is applied (:math:`|\Delta Z_{ij}| = 1`).
```

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Whether scaling with a LTR metric is actually more effective is still up for debate, `[8] <#references>`__ provides a theoretical foundation for general lambda loss functions and some insights into the framework.
```
hcho3, line 114:
```suggestion
Whether scaling with a LTR metric is actually more effective is still up for debate; `[8] <#references>`__ provides a theoretical foundation for general lambda loss functions and some insights into the framework.
```

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+
+******************
+Constructing Pairs
+******************
+
+There are two implemented strategies for constructing document pairs for :math:`\lambda`-gradient calculation. The first one is the ``mean`` method, another one is the ``topk`` method. The preferred strategy can be specified by the ``lambdarank_pair_method`` parameter.
+
+For the ``mean`` strategy, XGBoost samples ``lambdarank_num_pair_per_sample`` pairs for each document in a query list. For example, given a list of 3 documents and ``lambdarank_num_pair_per_sample`` is set to 2, XGBoost will randomly sample 6 pairs assuming the labels for these documents are different. On the other hand, if the pair method is set to ``topk``, XGBoost constructs about :math:`k \times |query|` number of pairs with :math:`|query|` pairs for each sample at the top :math:`k = lambdarank\_num\_pair` position. The number of pairs counted here is an approximation since we skip pairs that have the sample label.
```
hcho3, line 122:
```suggestion
For the ``mean`` strategy, XGBoost samples ``lambdarank_num_pair_per_sample`` pairs for each document in a query list. For example, given a list of 3 documents and ``lambdarank_num_pair_per_sample`` is set to 2, XGBoost will randomly sample 6 pairs, assuming the labels for these documents are different. On the other hand, if the pair method is set to ``topk``, XGBoost constructs about :math:`k \times |query|` number of pairs with :math:`|query|` pairs for each sample at the top :math:`k = lambdarank\_num\_pair` position. The number of pairs counted here is an approximation since we skip pairs that have the same label.
```

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+
+*********************
+Obtaining Good Result
+*********************
+
+Learning to rank is a sophisticated task and a field of heated research. It's not trivial to train a model that generalizes well. There are multiple loss functions available in XGBoost along with a set of hyper-parameters. This section contains some hints for how to choose those parameters as a starting point. One can further optimize the model by tuning these parameters.
```
hcho3, line 128:
```suggestion
Learning to rank is a sophisticated task and an active research area. It's not trivial to train a model that generalizes well. There are multiple loss functions available in XGBoost along with a set of hyperparameters. This section contains some hints for how to choose hyperparameters as a starting point. One can further optimize the model by tuning these hyperparameters.
```
It's common to refer to `hyperparameter` as a single word in academic papers.

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+
+The first question would be how to choose an objective that matches the task at hand. If your input data is multi-level relevance degree, then either ``rank:ndcg`` or ``rank:pairwise`` should be used. However, when the input is binary we have multiple options based on the target metric. `[6] <#references>`__ provides some guidelines on this topic and users are encouraged to see the analysis done in their work. The choice should be based on the number of `effective pairs`, which refers to the number of pairs that can generate non-zero gradient and contribute to training. `LambdaMART` with ``MRR`` has the least amount of effective pairs as the :math:`\lambda`-gradient is only non-zero when the pair contains a non-relevant document ranked higher than the top relevant document. As a result, it's not implemented in XGBoost. Since ``NDCG`` is a multi-level metric, it usually generate more effective pairs than ``MAP``.
```
hcho3, line 130:
```suggestion
The first question would be how to choose an objective that matches the task at hand. If your input data has multi-level relevance degrees, then either ``rank:ndcg`` or ``rank:pairwise`` should be used. However, when the input has binary labels, we have multiple options based on the target metric. `[6] <#references>`__ provides some guidelines on this topic and users are encouraged to see the analysis done in their work. The choice should be based on the number of `effective pairs`, which refers to the number of pairs that can generate non-zero gradient and contribute to training. `LambdaMART` with ``MRR`` has the least amount of effective pairs as the :math:`\lambda`-gradient is only non-zero when the pair contains a non-relevant document ranked higher than the top relevant document. As a result, it's not implemented in XGBoost. Since ``NDCG`` is a multi-level metric, it usually generate more effective pairs than ``MAP``.
```

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+
+However, when there's a sufficient amount of effective pairs, it's shown in `[6] <#references>`__ that matching the target metric with the objective is of significance. When the target metric is ``MAP`` and you are using a large dataset that can provide a sufficient amount of effective pairs, ``rank:map`` can in theory yield higher ``MAP`` value than the ``rank:ndcg``.
```
hcho3, line 132:
```suggestion
However, when there are sufficiently many effective pairs, it's shown in `[6] <#references>`__ that matching the target metric with the objective is of significance. When the target metric is ``MAP`` and you are using a large dataset that can provide a sufficient amount of effective pairs, ``rank:map`` can in theory yield higher ``MAP`` value than ``rank:ndcg``.
```

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+
+The choice of pair method (``lambdarank_pair_method``) and the number of pairs for each sample (``lambdarank_num_pair_per_sample``) is similar, as the mean-``NDCG`` considers more pairs than ``NDCG@10``, it can generate more effective pairs and provide more granularity. Also, using the ``mean`` strategy can help the model generalize with random sampling. However, one might want to focus the training on the top :math:`k` documents instead of using all pairs in practice, the tradeoff should be made based on the user's goal.
```
hcho3, line 134:
```suggestion
The consideration of effective pairs also applies to the choice of pair method (``lambdarank_pair_method``) and the number of pairs for each sample (``lambdarank_num_pair_per_sample``). For example, the mean-``NDCG`` considers more pairs than ``NDCG@10``, so the former generates more effective pairs and provides more granularity than the latter. Also, using the ``mean`` strategy can help the model generalize with random sampling. However, one might want to focus the training on the top :math:`k` documents instead of using all pairs, to better fit their real-world application.
```

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+
+When using mean value instead of targeting a specific position by calculating the target metric (like ``NDCG``) over the whole query list, user can specify how many pairs they want in each query by setting the ``lambdarank_num_pair_per_sample`` and XGBoost will randomly sample this amount of pairs for each element in the query group (:math:`|pairs| = |query| \times num\_pairsample`). Often time, setting it to 1 can produce reasonable result, with higher value producing more pairs (with the hope that a reasonable amount of them being effective). On the other hand, if you are prioritizing the top :math:`k` documents, the ``lambdarank_num_pair_per_sample`` should be set to slightly higher than :math:`k` (with a few more documents) to obtain a good training result.
```
hcho3, line 136:
```suggestion
When using the mean strategy for generating pairs, where the target metric (like ``NDCG``) is computed over the whole query list, users can specify how many pairs should be generated per each document, by setting the ``lambdarank_num_pair_per_sample``. XGBoost will randomly sample ``lambdarank_num_pair_per_sample`` pairs for each element in the query group (:math:`|pairs| = |query| \times num\_pairsample`). Often, setting it to 1 can produce reasonable results. In cases where performance is inadequate due to insufficient number of effective pairs being generated, set ``lambdarank_num_pair_per_sample`` to a higher value. As more document pairs are generated, more effective pairs will be generated as well.

On the other hand, if you are prioritizing the top :math:`k` documents, the ``lambdarank_num_pair_per_sample`` should be set slightly higher than :math:`k` (with a few more documents) to obtain a good training result.
```

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+
+In summary, to start off the training, if you have a large dataset, consider using the target-matching objective, otherwise ``NDCG`` or the RankNet loss (``rank:pairwise``) might be preferred. With the same target metric, use the ``lambdarank_num_pair_per_sample`` to specify the top :math:`k` documents for training if your dataset is large, and use the mean value version otherwise. Lastly, ``lambdarank_num_pair_per_sample`` can be used to control the amount of pairs for both methods.
```
hcho3, line 138:
```suggestion
**Summary** If you have large amount of training data:

* Use the target-matching objective.
* Choose the ``topk`` strategy for generating document pairs (if it's appropriate for your application).

On the other hand, if you have comparatively small amount of training data:

* Select ``NDCG`` or the RankNet loss (``rank:pairwise``).
* Choose the ``mean`` strategy for generating document pairs, to obtain more effective pairs.

For any method chosen, you can modify ``lambdarank_num_pair_per_sample`` to control the amount of pairs generated. 
```

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+
+********************
+Distributed Training
+********************
+XGBoost implements distributed learning-to-rank with integration of multiple frameworks including dask, spark, and pyspark. The interface is similar to single node. Please refer to document of the respective XGBoost interface for details. Scattering a query group onto multiple workers is theoretically sound but can affect the model accuracy. For most of the use cases, the small discrepancy is not an issue since when distributed training is involved the dataset is usually large. As a result, users don't need to partition the data based on group information. Given the dataset is correctly sorted, XGBoost can aggregate sample gradients accordingly.
```
hcho3, line 143:
```suggestion
XGBoost implements distributed learning-to-rank with integration of multiple frameworks including Dask, Spark, and PySpark. The interface is similar to the single-node counterpart. Please refer to document of the respective XGBoost interface for details. Scattering a query group onto multiple workers is theoretically sound but can affect the model accuracy. For most of the use cases, the small discrepancy is not an issue, as the amount of training data is usually large when distributed training is used. As a result, users don't need to partition the data based on query groups. As long as each data partition is correctly sorted by query IDs, XGBoost can aggregate sample gradients accordingly.
```

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+
+*******************
+Reproducbile Result
```
hcho3, line 146:
```suggestion
Reproducible Result
```

doc/tutorials/learning_to_rank.rst
```diff
@@ -0,0 +1,177 @@
...
+*******************
+
+Like any other tasks, XGBoost should generate reproducbile results given the same hardware and software environments, along with data partitions if distributed interface is used. Even when the underlying environment has changed, the result should still be consistent. However, when the ``lambdarank_pair_method`` is set to ``mean``, XGBoost uses sampling, and the random number generator used on Windows (MSVC) is different from the one used on other platforms like Linux (GCC, Clang), the output varies significantly between these platforms.
```
hcho3, line 149:
```suggestion
Like any other tasks, XGBoost should generate reproducible results given the same hardware and software environments (and data partitions, if distributed interface is used). Even when the underlying environment has changed, the result should still be consistent. However, when the ``lambdarank_pair_method`` is set to ``mean``, XGBoost uses random sampling, and results may differ depending on the platform used. The random number generator used on Windows (Microsoft Visual C++) is different from the ones used on other platforms like Linux (GCC, Clang), so the output varies significantly between these platforms.
```

hcho3 (changes requested):
I reviewed the documentation part of this pull request. Nice discussion on learning-to-rank, including the method for choosing different hyperperameters.
The demo code is fantastic, as the built-in click data simulator (`PDM`) greatly enhances understanding.

I have some stylistic recommendations.

doc/tutorials/learning_to_rank.rst
trivialfis: Good suggestion!

doc/tutorials/learning_to_rank.rst
trivialfis: Maybe I should mention other bias as well, for instance, presentation bias.

doc/tutorials/learning_to_rank.rst
trivialfis: A side note, we will remove the requirement on sorted data in coming PRs.

python-package/xgboost/testing/data.py
```diff
@@ -340,3 +343,242 @@ def get_mq2008(
...
+    """Simulate click data with position bias model. There are other models available in
+    `ULTRA <https://github.com/ULTR-Community/ULTRA.git>`_ like the cascading model.
+
+    References
+    ----------
+    Unbiased LambdaMART: An Unbiased Pairwise Learning-to-Rank Algorithm
+
+    """
+
+    def __init__(self, eta: float) -> None:
```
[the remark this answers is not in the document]
trivialfis: I think the default parameter is not necessary.

python-package/xgboost/testing/data.py
```diff
@@ -340,3 +343,242 @@ def get_mq2008(
         y_valid,
         qid_valid,
     )
+
+
+ClickFold = namedtuple("ClickFold", ("X", "y", "q", "s", "c", "p"))
```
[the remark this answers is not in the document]
trivialfis: Will change it.

trivialfis:
whew ...

PR #8823, opened 2023-02-18
Fixes compilation errors on MSVC x86 targets

mxmauro:
This PR aims to fix compilation issues found when targeting 32-bit on MSVC.

`__popcnt64` is not available on x86 and, for the rest of the fixes, MSVC does not like narrowing without explicit conversion.

### DIFF
--- a/include/xgboost/linalg.h
+++ b/include/xgboost/linalg.h
@@ -151,7 +151,7 @@ inline LINALG_HD int Popc(uint64_t v) {
   return __popcll(v);
 #elif defined(__GNUC__) || defined(__clang__)
   return __builtin_popcountll(v);
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && _defined(_M_X64)
   return __popcnt64(v);
 #else
   return NativePopc(v);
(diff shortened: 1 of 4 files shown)


trivialfis (commented):
Thank you for the fix! Could you please change the c-style cast to `static_cast` instead?

mxmauro:
Sure!

trivialfis:
Hi, could you please fix the c++ lint error as well?

mxmauro:
Hi, yes but no idea how.

There are other bugs I'm getting when compiling other projects generated by cmake related to linter which calls python. The problem is the line `string(REPLACE " " ";" EXCLUDE_PATH ${EXCLUDE_PATH})` or `dmlc-core\cmake\lint.cmake`, because does not include quotes and `EXCLUDE_PATH` is not defined, I get this error:

`string sub-command replace requires four arguments`

mxmauro:
Hi @trivialfis didn't realize there were a lot of checks and found the lint issue (line too long). Now it is fixed.

Have a nice weekend.

trivialfis (approved):
Thank you for the nice fixes!

mxmauro:
You are welcome :)

PR #8826, opened 2023-02-21
Update eval_metric validation to support list of strings

mzzhang95:
Hey guys, I find that SparkXGB estimators do not support multiple evaluation metric because of param validation. 

In this PR, I deleted the param validation for eval_metrics.

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -314,6 +314,7 @@ class _SparkXGBParams(
                 raise ValueError("Only string type 'objective' param is allowed.")
 
         if self.getOrDefault(self.eval_metric) is not None:
+            if not (
                 isinstance(self.getOrDefault(self.eval_metric), str)
                 or (isinstance(self.getOrDefault(self.eval_metric), List)
                     and all(isinstance(metric, str) for metric in self.getOrDefault(self.eval_metric)))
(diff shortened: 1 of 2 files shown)


hcho3:
Did you test this pull request on your machine?

mzzhang95:
> Did you test this pull request on your machine?

Yes, I have tested it. It works well. After deleting these three lines, I am able to train a SparkXGB model with multiple eval_metrics.

wbo4958:
@mzzhang95 Seems eval_metric can accept str or list[str], could you add the list[str] support instead of getting rid of the validation?

mzzhang95:
> @mzzhang95 Seems eval_metric can accept str or list[str], could you add the list[str] support instead of getting rid of the validation?

I added the check back. Now it will check both str and List[str] for eval_metric.

trivialfis (commented):
Thank you for the fix! Could you please help add a simple test in https://github.com/dmlc/xgboost/blob/d9688f93c7bbdc31fcebd89f00144a18d8660fcf/tests/test_distributed/test_with_spark/test_spark_local.py#L448 ?

We use `pytest` for running python tests.

mzzhang95:
I added two unit tests for string type and list type ``eval_metric`` with ``SparkXGBoostClassifier`. Do you also wanna include regressor and ranker?

And when could this PR be merged? Will there be a release in recent? Thank!

trivialfis (approved):

trivialfis:
> And when could this PR be merged? 

I should be able to merge it once all tests pass.

> Will there be a release in recent? Thank!

Patch release for critical bug fixes only. The next major release will take some time.

PR #8828, opened 2023-02-22
Support column-split in row partitioner

rongou:
The approach is similar to what's done in `cpu_predictor.cc`: first collect decision/missing bits, then aggregate them across all the workers, and finally use them to do the partitioning.

For now only support the `approx` method.

### DIFF
--- a/src/common/partition_builder.h
+++ b/src/common/partition_builder.h
@@ -201,6 +201,11 @@ class PartitionBuilder {
     SetNRightElems(node_in_set, range.begin(), n_right);
   }
 
+  /**
+   * @brief When data is split by column, we don't have all the features locally on the current
+   * worker, so we go through all the rows and mark the bit vectors on whether the decision is made
+   * to go right, or if the feature value used for the split is missing.
+   */
   void MaskRows(const size_t node_in_set, std::vector<xgboost::tree::CPUExpandEntry> const &nodes,
                 const common::Range1d range, GHistIndexMatrix const& gmat,
                 const common::ColumnMatrix& column_matrix,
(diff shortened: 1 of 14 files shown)


rongou:
@trivialfis @hcho3

src/common/partition_builder.h
```diff
@@ -199,6 +201,100 @@ class PartitionBuilder {
...
+    std::size_t nid = nodes[node_in_set].nid;
+    bst_feature_t fid = tree[nid].SplitIndex();
+    bool is_cat = tree.GetSplitTypes()[nid] == FeatureType::kCategorical;
+    auto node_cats = tree.NodeCats(nid);
+
+    auto const& index = gmat.index;
+    auto const& cut_values = gmat.cut.Values();
+    auto const& cut_ptrs = gmat.cut.Ptrs();
+
+    auto gidx_calc = [&](auto ridx) {
```
trivialfis, line 224: Please consider some refactoring to reuse this function with `partition()`, maybe with `GHistIndexMatrix::GetFValue` as well.

src/tree/common_row_partitioner.h
```diff
@@ -205,6 +244,14 @@ class CommonRowPartitioner {
         ctx, tree, this->Partitions(), p_out_position,
         [&](size_t idx) -> bool { return gpair[idx].GetHess() - .0f == .0f; });
   }
+
+ private:
```
trivialfis, line 248: Maybe we should extract these into a different class or struct with its own utility member functions? Also, the `CommonRowPartitioner` has private members, please place them together.

trivialfis (approved):

PR #8831, opened 2023-02-22
[spark][test] xgboost 1.7 build against spark 3.4

WeichenXu123:
xgboost 1.7 build against spark 3.4

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
@@ -28,7 +28,9 @@ import org.apache.spark.sql.functions._
 import scala.collection.{Iterator, mutable}
 
 import org.apache.spark.ml.param.ParamMap
-import org.apache.spark.ml.util.{DefaultXGBoostParamsReader, DefaultXGBoostParamsWriter, XGBoostWriter}
+import org.apache.spark.ml.util.{
+  DefaultXGBoostParamsReader, DefaultXGBoostParamsWriter, XGBoostDatasetUtils, XGBoostWriter
+}
 import org.apache.spark.sql.types.StructType
 
 class XGBoostClassifier (
@@ -166,7 +168,7 @@ class XGBoostClassifier (
   }
 
   override protected def train(dataset: Dataset[_]): XGBoostClassificationModel = {
-    val _numClasses = getNumClasses(dataset)
+    val _numClasses = XGBoostDatasetUtils.getNumClasses(dataset, this.getLabelCol)
     if (isDefined(numClass) && $(numClass) != _numClasses) {
       throw new Exception("The number of classes in dataset doesn't match " +
         "\'num_class\' in xgboost params.")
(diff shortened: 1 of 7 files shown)


zhengruifeng (approved):

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
```diff
@@ -166,7 +168,7 @@ class XGBoostClassifier (
   }
 
   override protected def train(dataset: Dataset[_]): XGBoostClassificationModel = {
-    val _numClasses = getNumClasses(dataset)
+    val _numClasses = XGBoostDatasetUtils.getNumClasses(dataset, this.getLabelCol)
```
WeichenXu123, line 171: "removed getNumClasses method" causes breaking change.
Besides this issue, other part xgboost4j codes are compatible with spark 3.4.

The breaking change is in spark 3.4 and it haven't been released, 
@zhengruifeng will make a fix in apache/spark
so that we can avoid custom build xgboost4j-spark for databricks runtime.

wbo4958:
May I have a question? Can the xgboost jars built against spark 3.4 run on older spark like 3.1/3.2/3.3?

WeichenXu123:
> May I have a question? Can the xgboost jars built against spark 3.4 run on older spark like 3.1/3.2/3.3?

Unfortunately we cannot ensure this.
The ideal solution is to make CI runs against 3.2/3.3/3.4 to detect incompatible changes.
Good news is spark mllib is in maintenance mode, usually it would not have big change in future.

WeichenXu123:
> make CI runs against 3.2/3.3/3.4 to detect incompatible changes.

To saving resources, we can run these CIs only when we want to release a new version.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala
WeichenXu123: Spark side fixing PR is filed: https://github.com/apache/spark/pull/40119

wbo4958:
I just checked there is no getNumClasses API of DatasetUtils in the spark311 which means, it should fail if we compile the xgboost jar on spark 340 and run on spark311.

Could you rework this to make xgboost run on any kind of spark 3.x?

WeichenXu123:
> I just checked there is no getNumClasses API of DatasetUtils in the spark311 which means, it should fail if we compile the xgboost jar on spark 340 and run on spark311.
> 
> Could you rework this to make xgboost run on any kind of spark 3.x?

This PR is just for testing, we should not merge it.
See https://github.com/dmlc/xgboost/pull/8831#discussion_r1114169398
Spark already adds back the `getNumClasses` API to Model class. So xgboost code does not need any change and it will work fine on spark 3.4

WeichenXu123:
CC @wbo4958 Could you help me trigger the test `buildkite/xgboost-ci/pr ` ? I just want to verify xgboost 1.7 python package works fine with spark 3.4

trivialfis:
triggered all linux tests.

trivialfis:
Hi, I merged https://github.com/dmlc/xgboost/pull/9083 and https://github.com/dmlc/xgboost/pull/9039 Is there any additional items need to be checked?

PR #8845, opened 2023-02-25
Implement parallel logical algorithms.

trivialfis:

### DIFF
--- a/src/common/algorithm.h
+++ b/src/common/algorithm.h
@@ -3,11 +3,11 @@
  */
 #ifndef XGBOOST_COMMON_ALGORITHM_H_
 #define XGBOOST_COMMON_ALGORITHM_H_
-#include <algorithm>          // for upper_bound, stable_sort, sort, max, all_of, none_of, min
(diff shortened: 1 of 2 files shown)


RAMitchell (commented):
Is this faster at all?

trivialfis:
@RAMitchell Good point.

For simple conditions like `v > 0.0f`, the difference is negligible. For more complicated checks like whether the input is an integer and larger than 0 (NDCG label), the difference is somewhat visible but still not an issue.

```
std::vector<float> data(100000000);

parallel none_of:      0.059212s
std::none_of:  0.140824s
```

I simply implemented it as these two functions when used for data validation, it has to walk through the entire dataset. Seems not necessary.

RAMitchell:
Feel free to merge if you want. My experience in the past is that simple operations on data don't benefit much. Multithreading then adds a bit of complexity and makes debugging harder.

trivialfis:
Make sense, thank you for the review. Will remove the code from the LTR branch.

PR #8847, opened 2023-02-25
Support column split in `approx` tree method

rongou:
Since each worker has a distinct set of columns/features, and we already build histograms locally (#8811) and partition rows collaboratively (#8828), the only change remaining is to handle split finding correctly. We first find the local best splits using the existing approach, then do a round of `allgather` to collect best splits from all the workers, and update splits to be globally best splits.

### DIFF
--- a/include/xgboost/data.h
+++ b/include/xgboost/data.h
@@ -529,6 +529,11 @@ class DMatrix {
     return Info().num_nonzero_ == Info().num_row_ * Info().num_col_;
   }
 
+  /*! \brief Whether the data is split row-wise. */
+  bool IsRowSplit() const {
+    return Info().data_split_mode == DataSplitMode::kRow;
+  }
+
   /*! \brief Whether the data is split column-wise. */
   bool IsColumnSplit() const {
     return Info().data_split_mode == DataSplitMode::kCol;
--- a/src/data/data.cc
+++ b/src/data/data.cc
@@ -912,6 +912,7 @@ DMatrix* DMatrix::Load(const std::string& uri, bool silent, DataSplitMode data_s
     if (!cache_file.empty()) {
       LOG(FATAL) << "Column-wise data split is not support for external memory.";
     }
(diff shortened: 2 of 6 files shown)


rongou:
@trivialfis @hcho3

trivialfis:
>  the only change remaining is to handle split finding correctly.

That's exciting! Do you have a complete example that I can run? Would love to try it out.

rongou:
At least on a small dataset it seems to produce identical results as row split:
```python
import filecmp
import multiprocessing

import xgboost as xgb
from xgboost import RabitTracker


def train(split, rank):
    dtrain = xgb.DMatrix('demo/data/agaricus.txt.train', data_split_mode=split)
    dtest = xgb.DMatrix('demo/data/agaricus.txt.test', data_split_mode=split)
    param = {"max_depth": 2, "eta": 1, "objective": "binary:logistic"}
    watchlist = [(dtest, "eval"), (dtrain, "train")]
    num_round = 2
    bst = xgb.train(param, dtrain, num_boost_round=num_round, evals=watchlist)
    if rank == 0:
        bst.save_model(f'agaricus.model.{split}.json')


def run_worker(rabit_env, rank):
    with xgb.collective.CommunicatorContext(**rabit_env):
        print("Training with row split")
        train(0, rank)
        print("Training with column split")
        train(1, rank)


def main():
    world_size = 2
    tracker = RabitTracker(host_ip='127.0.0.1', n_workers=world_size)
    tracker.start(world_size)

    workers = []
    for rank in range(world_size):
        worker = multiprocessing.Process(target=run_worker, args=(tracker.worker_envs(), rank))
        workers.append(worker)
        worker.start()
    for worker in workers:
        worker.join()
        assert worker.exitcode == 0

    result = filecmp.cmp('agaricus.model.0.json', 'agaricus.model.1.json', shallow=False)
    print(f'Two models are equal: {result}')


if __name__ == "__main__":
    main()
```

trivialfis (approved):

Issue #8683, opened 2023-01-14
[CI] Flaky JVM test.


trivialfis:
https://github.com/dmlc/xgboost/actions/runs/3920444405/jobs/6702077160

```
2023-01-14T22:04:46.8896271Z XGBoostRegressorSuite:
2023-01-14T22:04:46.9548459Z - XGBoost-Spark XGBoostRegressor output should match XGBoost4j *** FAILED ***
2023-01-14T22:04:46.9549885Z   ml.dmlc.xgboost4j.java.XGBoostError: [22:04:46] /home/runner/work/xgboost/xgboost/src/learner.cc:1444: Check failed: learner_model_param_.num_feature == p_fmat->Info().num_col_ (126 vs. 36) : Number of columns does not match number of features in booster.
2023-01-14T22:04:46.9550707Z Stack trace:
2023-01-14T22:04:46.9551425Z   [bt] (0) /tmp/libxgboost4j7459547976508297557.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x85) [0x7f76e088bb25]
2023-01-14T22:04:46.9552069Z   [bt] (1) /tmp/libxgboost4j7459547976508297557.so(xgboost::LearnerImpl::ValidateDMatrix(xgboost::DMatrix*, bool) const+0x116) [0x7f76e0a88276]
2023-01-14T22:04:46.9552790Z   [bt] (2) /tmp/libxgboost4j7459547976508297557.so(xgboost::LearnerImpl::UpdateOneIter(int, std::shared_ptr<xgboost::DMatrix>)+0xb8) [0x7f76e0a90728]
2023-01-14T22:04:46.9599461Z   [bt] (3) /tmp/libxgboost4j7459547976508297557.so(XGBoosterUpdateOneIter+0x73) [0x7f76e0896843]
2023-01-14T22:04:46.9606162Z   [bt] (4) [0x7f7722dc26bd]
2023-01-14T22:04:46.9606504Z   at ml.dmlc.xgboost4j.java.XGBoostJNI.checkCall(XGBoostJNI.java:48)
2023-01-14T22:04:46.9606916Z   at ml.dmlc.xgboost4j.java.Booster.update(Booster.java:173)
2023-01-14T22:04:46.9607352Z   at ml.dmlc.xgboost4j.java.XGBoost.trainAndSaveCheckpoint(XGBoost.java:217)
[...]

PR #8851, opened 2023-02-27
Fix cache with gc

trivialfis:
Close https://github.com/dmlc/xgboost/issues/8683 .

The previous cache uses static thread local memory implemented as part of the standard library. However, the Scala binding uses a different thread to call the `dispose` method, which calls into the booster destructor. As a result, the constructor and destructor are invoked by different threads, violating the cache cleanup process.

A short-lived lock is required for enabling thread safety with the new DMatrix cache.

- Make DMatrixCache thread safe.
- Remove the use of thread-local memory.


tests/cpp/test_cache.cc
```diff
@@ -52,4 +54,63 @@ TEST(DMatrixCache, Basic) {
     }
   }
 }
+
+TEST(DMatrixCache, MultiThread) {
```
trivialfis, line 58: I have run this test a few thousand times using the `gtest_repeat`, along with thread sanitizer.

RAMitchell (approved):

PR #8853, opened 2023-02-28
[CI] Require C++17 to build XGBoost

### LINKED ISSUE #8737: [RFC] Require C++17 standard.
Related: https://github.com/microsoft/LightGBM/issues/5691

It appears that a future version of R may require the R package to use C++17. This is a sign that C++17 is being widely adopted.

My proposal is to require C++17 standard to build XGBoost.

Supporting factors
* Sufficiently modern C++ compilers support C++17, such as GCC 5+ and Visual Studio 2019+.
* The disruption should be minimal, since our CI already tests XGBoost with C++17 (this is necessary for the RMM plugin).

On the side of the CI, I will need to re-build the Windows worker to use Visual Studio 2019 (right now it's using 2017).

@dmlc/xgboost-committer

hcho3:
Also upgrade Visual Studio to 2019.

Closes #8737


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
```
hcho3, line 1: CMake 3.18 is needed to use CUDA with C++17. See https://cmake.org/cmake/help/latest/prop_tgt/CUDA_STANDARD.html

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
...
+cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
```
hcho3, line 1: CMake 3.18 is needed to use CUDA with C++17. See https://cmake.org/cmake/help/latest/prop_tgt/CUDA_STANDARD.html

RAMitchell (approved):

trivialfis (commented):
Looks good to me assuming tests pass (except for clang-tidy, which we can work on it later).

PR #8858, opened 2023-03-01
Take advantage of C++17 features

rongou:
Using `std::as_const` and `std::exclusive_scan`.


rongou:
@hcho3

rongou:
Hmm looks like the R windows pipeline is still dong `gnu++14`?

trivialfis:
let me fix it.

trivialfis:
https://github.com/dmlc/xgboost/pull/8860

rongou:
@hcho3 looks like the gcc version is too old for `std::exclusive_scan`?

hcho3:
Using GCC 9 for now, since some R packages don't build with GCC 10. See https://github.com/kr-colab/discoal/issues/21#issue-665436177

hcho3:
Disabling sanitizer test due to #8863

trivialfis:
please don't disable the sanitizer test for the use of `exclusive_scan`.

trivialfis:
As mentioned in previous offline chats, we should expect some compiler issues with c++17 considering we support multiple bindings with different specifications of compilers (especially for R mingw). We won't be fully utilizing c++-17 for a period of time depending on the choice of compilers.

trivialfis (approved):

PR #8859, opened 2023-03-01
Support sklearn cross validation for ranker.

trivialfis:
- Add a convention for X to include a special `qid` column.

sklearn utilities consider only `X`, `y`, and `sample_weight` for supervised learning algorithms, but we need an additional qid array for ranking.

It's important to be able to support the cross-validation function in sklearn since all other tuning functions like grid search are based on cross-validation.

I plan to deprecate the use of `group`. `qid` is the most widely used convention in learning to rank implementations.

### DIFF
--- a/python-package/xgboost/callback.py
+++ b/python-package/xgboost/callback.py
@@ -23,7 +23,13 @@ from typing import (
 import numpy
 
 from . import collective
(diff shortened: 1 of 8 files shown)


trivialfis:
The convention is primarily for the `score` function since `cross_validate` accepts a `fit_params` parameter.

RAMitchell (commented):
Nice work!

PR #8860, opened 2023-03-01
Update c++ requirement to 17 for the R package.

trivialfis:


rongou (approved):

rongou:
Don't we also need to change these?
https://github.com/dmlc/xgboost/blob/master/R-package/src/Makevars.in#L6
https://github.com/dmlc/xgboost/blob/master/R-package/src/Makevars.win#L6

hcho3:
I wonder why the CI passed. Anyway, can you file a new PR to fix it?

Issue #8854, opened 2023-03-01
[CI] Re-enable clang-tidy


hcho3:
In #8853, we disabled clang-tidy to unblock the CI. Clang-tidy began to fail because it has additional requirements for projects using the C++17 standard.
We would need to update the C++ codebase substantially to comply with the styling guideline of clang-tidy.

@trivialfis

hcho3:
This line should be un-commented to re-enable clang-tidy:
https://github.com/dmlc/xgboost/blob/6d8afb22188aa89cbcc720542e0fa331227ff7cb/tests/buildkite/pipeline.yml#L25-L29

PR #8861, opened 2023-03-02
Restore clang tidy test.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/8854 .

- Upgrade clang-tidy to 15.
- Fix clang-tidy warnings.

~One caveat of upgrading the clang version is that it might require a curand header.~

### DIFF
--- a/src/common/charconv.cc
+++ b/src/common/charconv.cc
@@ -495,12 +495,10 @@ class PowerBaseComputer {
                              static_cast<int32_t>(IEEE754::kFloatBias) -
                              static_cast<int32_t>(IEEE754::kFloatMantissaBits) -
                              static_cast<int32_t>(2);
-      static_assert(static_cast<int32_t>(1) -
-                            static_cast<int32_t>(IEEE754::kFloatBias) -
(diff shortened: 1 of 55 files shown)


src/common/charconv.cc
```diff
@@ -911,7 +911,7 @@ from_chars_result FromCharFloatImpl(const char *buffer, const int len,
   // the bias and also special-case the value 0.
   int32_t shift = (f_e2 == 0 ? 1 : f_e2) - exp_b2 - IEEE754::kFloatBias -
                   IEEE754::kFloatMantissaBits;
-  assert(shift >= 0);
+  assert(shift >= 1);
```
hcho3, line 914: Why the change of condition here?

trivialfis: Tell clang-tidy that `shift` must be greater or equal to zero. I have verified the condition using `CHECK` and the `RoundTripExhaustive` JSON test.

hcho3: > Tell clang-tidy that shift must be greater or equal to zero.

But the new condition says `shift >= 1`?

trivialfis: I will just put a nolint there.

PR #8866, opened 2023-03-03
Support learning rate for zero-hessian objectives.

trivialfis:
- Extract the train param into gbtree.
- Pass learning rate into the objective.


trivialfis:
let me add some more tests .

RAMitchell (approved):

PR #8872, opened 2023-03-06
Support F order for the tensor type.

trivialfis:
- Add F order support for tensor and view. Gradients for multi-class and multi-target are stored in F order when updating trees.
- Use parameter pack for automatic type cast. (avoid excessive static cast for shape).

### DIFF
--- a/src/objective/quantile_obj.cu
+++ b/src/objective/quantile_obj.cu
@@ -64,8 +64,7 @@ class QuantileRegression : public ObjFunction {
     out_gpair->SetDevice(ctx_->gpu_id);
     out_gpair->Resize(n_targets * info.num_row_);
     auto gpair =
-        linalg::MakeTensorView(ctx_->IsCPU() ? out_gpair->HostSpan() : out_gpair->DeviceSpan(),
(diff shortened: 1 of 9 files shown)


src/objective/quantile_obj.cu
```diff
@@ -80,15 +79,8 @@ class QuantileRegression : public ObjFunction {
...
-                                          {static_cast<std::size_t>(n_samples),
-                                           static_cast<std::size_t>(alpha.size()),
-                                           static_cast<std::size_t>(n_targets / alpha.size())});
-
-          // std::tie is not available for cuda kernel.
-          std::size_t sample_id = std::get<0>(idx);
-          std::size_t quantile_id = std::get<1>(idx);
-          std::size_t target_id = std::get<2>(idx);
+          auto [sample_id, quantile_id, target_id] =
+              linalg::UnravelIndex(i, n_samples, alpha.size(), n_targets / alpha.size());
```
trivialfis, line 83: @hcho3 I think your c++-17 work is really helpful.

hcho3: Awesome!

RAMitchell (approved):

PR #8889, opened 2023-03-09
Pass obj info by reference instead of by value.

trivialfis:
- Pass obj info into tree updater as const pointer.

This way we don't have to initialize the learner model param before configuring gbm, hence breaking up the dependency of configurations.

### DIFF
--- a/include/xgboost/tree_updater.h
+++ b/include/xgboost/tree_updater.h
@@ -1,5 +1,5 @@
-/*!
- * Copyright 2014-2022 by XGBoost Contributors
+/**
+ * Copyright 2014-2023 by XGBoost Contributors
  * \file tree_updater.h
  * \brief General primitive for tree learning,
  *   Updating a collection of trees given the information.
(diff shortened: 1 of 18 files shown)


include/xgboost/tree_updater.h
```diff
@@ -91,22 +90,23 @@ class TreeUpdater : public Configurable {
...
 
-  /*!
+  /**
    * \brief Create a tree updater given name
    * \param name Name of the tree updater.
    * \param ctx A global runtime parameter
+   * \param task Infomation about the objective.
    */
-  static TreeUpdater* Create(const std::string& name, Context const* ctx, ObjInfo task);
+  static TreeUpdater* Create(const std::string& name, Context const* ctx, ObjInfo const* task);
```
hcho3, line 101:
```suggestion
  static TreeUpdater* Create(const std::string& name, Context const* ctx, ObjInfo const& task);
```

Have you considered using a constant reference instead? I don't see a reason for allowing a `nullptr` here.

hcho3 (changes requested):

include/xgboost/tree_updater.h
trivialfis: using a pointer can ask the compiler to check whether I'm accidentally passing it by value. If we change it to reference, we need to disable the default copy constructor and assignment operator, then define an explicit copy method to have the same effect, which seems to be a little too much for a simple struct.

hcho3: Got it, then pointer is fine.

hcho3 (approved):

PR #8890, opened 2023-03-09
Define multi-strategy parameter.

trivialfis:

- Define a new field for LearnerTrainParam to support both model-per-target and vector leaf.

The naming is `compo` for composite and `mono` for monolithic. It's probably not the best name, please make your suggestions. ;-)

Sorry for polluting the change with [include-what-you-use](https://github.com/include-what-you-use/include-what-you-use) comments. I can remove them if it makes review difficult.

I plan to slowly integrate it into the xgboost code base to have a clear view of what we are using and to eliminate unneeded includes to reduce compile time. If reviewers prefer having a dedicated PR for header changes, please let me know in the comments. ;-)

### DIFF
--- a/include/xgboost/learner.h
+++ b/include/xgboost/learner.h
@@ -324,7 +324,8 @@ struct LearnerModelParam {
   // this one as an immutable copy.
   LearnerModelParam(Context const* ctx, LearnerModelParamLegacy const& user_param,
                     linalg::Tensor<float, 1> base_margin, ObjInfo t, MultiStrategy multi_strategy);
-  LearnerModelParam(LearnerModelParamLegacy const& user_param, ObjInfo t, MultiStrategy multi_strategy);
+  LearnerModelParam(LearnerModelParamLegacy const& user_param, ObjInfo t,
+                    MultiStrategy multi_strategy);
   LearnerModelParam(bst_feature_t n_features, linalg::Tensor<float, 1> base_score,
                     std::uint32_t n_groups, bst_target_t n_targets, MultiStrategy multi_strategy)
       : base_score_{std::move(base_score)},
@@ -336,7 +337,9 @@ struct LearnerModelParam {
   [[nodiscard]] linalg::TensorView<float const, 1> BaseScore(std::int32_t device) const;
 
   void Copy(LearnerModelParam const& that);
-  [[nodiscard]] bool IsVectorLeaf() const noexcept { return multi_strategy == MultiStrategy::kMonolithic; }
+  [[nodiscard]] bool IsVectorLeaf() const noexcept {
+    return multi_strategy == MultiStrategy::kMonolithic;
+  }
   [[nodiscard]] bst_target_t OutputLength() const noexcept { return this->num_output_group; }
   [[nodiscard]] bst_target_t LeafLength() const noexcept {
     return this->IsVectorLeaf() ? this->OutputLength() : 1;
--- a/include/xgboost/learner.h
+++ b/include/xgboost/learner.h
@@ -8,29 +8,33 @@
 #ifndef XGBOOST_LEARNER_H_
 #define XGBOOST_LEARNER_H_
 
-#include <dmlc/io.h>          // Serializable
-#include <xgboost/base.h>
-#include <xgboost/context.h>  // Context
-#include <xgboost/feature_map.h>
-#include <xgboost/host_device_vector.h>
-#include <xgboost/linalg.h>  // Tensor
-#include <xgboost/model.h>
-#include <xgboost/task.h>
+#include <dmlc/io.h>          // for Serializable
+#include <xgboost/base.h>     // for bst_feature_t, bst_target_t, bst_float, Args, GradientPair
+#include <xgboost/context.h>  // for Context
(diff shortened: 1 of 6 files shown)


src/learner.cc
```diff
@@ -779,9 +810,17 @@ class LearnerConfiguration : public Learner {
     if (obj_ == nullptr || tparam_.objective != old.objective) {
       obj_.reset(ObjFunction::Create(tparam_.objective, &ctx_));
     }
+
+    bool has_nc {cfg_.find("num_class") != cfg_.cend()};
```
trivialfis, line 814: This change is made to fix the removed lines above. The change limits the inserted parameter into this function instead of propagating it through the entire library.

src/c_api/c_api_utils.h
```diff
@@ -12,10 +12,11 @@
 #include <vector>
 
 #include "xgboost/c_api.h"
-#include "xgboost/data.h"  // DMatrix
+#include "xgboost/data.h"         // DMatrix
+#include "xgboost/feature_map.h"  // for FeatureMap
```
trivialfis, line 16: The `feature map` is no longer included in `learner.h`. This is to fix missing includes.

src/learner.cc
```diff
@@ -306,18 +328,26 @@ struct LearnerTrainParam : public XGBoostParameter<LearnerTrainParam> {
...
-        .set_default("gbtree")
-        .describe("Gradient booster used for training.");
+    DMLC_DECLARE_FIELD(booster).set_default("gbtree").describe(
+        "Gradient booster used for training.");
     DMLC_DECLARE_FIELD(objective)
         .set_default("reg:squarederror")
         .describe("Objective function used for obtaining gradient.");
+    DMLC_DECLARE_FIELD(multi_strategy)
+        .add_enum("compo", Strategy::kComposite)
+        .add_enum("mono", Strategy::kMono)
```
hcho3, line 346:
```suggestion
        .add_enum("composite", Strategy::kComposite)
        .add_enum("monolithic", Strategy::kMonolithic)
```
Why not spell them out fully? Given the wide use of IDEs, we can afford to be a bit verbose when it comes to naming.

include/xgboost/learner.h
```diff
@@ -274,9 +281,16 @@ class Learner : public Model, public Configurable, public dmlc::Serializable {
 };
 
 struct LearnerModelParamLegacy;
+/**
+ * \brief Strategy for building multi-target models.
+ */
+enum class Strategy : std::int32_t {
+  kComposite = 0,
+  kMono = 1,
+};
```
hcho3, line 290:
```suggestion
enum class MultiTargetStrategy : std::int32_t {
  kComposite = 0,
  kMonolithic = 1,
};
```

Let's be more specific in the name

hcho3 (changes requested):
Here's my suggestion for naming.

src/learner.cc
```diff
@@ -226,35 +244,38 @@ struct LearnerModelParamLegacy : public dmlc::Parameter<LearnerModelParamLegacy>
...
+                                     Strategy multi_strategy)
+    : num_feature{user_param.num_feature},
+      num_output_group{
+          std::max(static_cast<std::uint32_t>(user_param.num_class), user_param.num_target)},
+      task{t},
+      multi_strategy{multi_strategy} {
+  if (user_param.num_class > 1 && user_param.num_target > 1) {
+    LOG(FATAL) << "multi-target-multi-class is not yet supported. Output classes:"
+               << user_param.num_class << ", output targets:" << user_param.num_target;
+  }
```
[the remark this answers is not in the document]
trivialfis: I set the lower bound to 1 in the parameter declaration.

src/learner.cc
```diff
@@ -974,9 +1014,6 @@ class LearnerIO : public LearnerConfiguration {
     if (!DMLC_IO_NO_ENDIAN_SWAP) {
       mparam_ = mparam_.ByteSwap();
     }
-    if (mparam_.num_target == 0) {
-      mparam_.num_target = 1;
-    }
```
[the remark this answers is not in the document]
trivialfis: I set the lower bound to 1 in the parameter declaration.

include/xgboost/learner.h
```diff
@@ -287,30 +301,48 @@ struct LearnerModelParam {
...
+        num_feature{n_features},
+        num_output_group{std::max(n_groups, n_targets)},
+        multi_strategy{multi_strategy} {}
 
   linalg::TensorView<float const, 1> BaseScore(Context const* ctx) const;
-  linalg::TensorView<float const, 1> BaseScore(int32_t device) const;
+  [[nodiscard]] linalg::TensorView<float const, 1> BaseScore(std::int32_t device) const;
 
   void Copy(LearnerModelParam const& that);
+  [[nodiscard]] bool IsVectorLeaf() const noexcept { return multi_strategy == Strategy::kMono; }
```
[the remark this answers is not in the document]
trivialfis: It will take some time to mature. The current implementation is less than alpha.

src/learner.cc
trivialfis: Done.

hcho3 (approved):

PR #8911, opened 2023-03-14
Check inf in data for all types of DMatrix.

trivialfis:


trivialfis:
Ran into segfault using GPU hist, caused by a CPU quantile DMatrix constructed from a dataset containing inf.

RAMitchell (approved):

PR #8920, opened 2023-03-15
[jvm-packages] Use akka 2.6

### LINKED ISSUE #8919: [jvm-packages] akka license change.
https://github.com/dmlc/xgboost/pull/8524#issuecomment-1470017657

hcho3:
Closes #8919

The akka package changed its license to a non-free license starting version 2.7. This may cause issues for some users.

@eejbyfeldt  
@trivialfis

### DIFF
--- a/jvm-packages/xgboost4j-gpu/pom.xml
+++ b/jvm-packages/xgboost4j-gpu/pom.xml
@@ -41,13 +41,13 @@
         <dependency>
             <groupId>com.typesafe.akka</groupId>
             <artifactId>akka-actor_${scala.binary.version}</artifactId>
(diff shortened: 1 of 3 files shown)


trivialfis (commented):
I don't understand the terms in the new license, was thinking about asking rapids for help. But reverting seems to be a good choice...

eejbyfeldt:
@hcho3 Thanks for considering this. It should be noted that we are currently not users of xgboost but just exploring the possibility of using it, but the usage of akka >= 2.7.0 would be a blocker for us.

trivialfis (approved):

PR #8968, opened 2023-03-24
Support prediction cache for multi-target tree.

trivialfis:
- Fix prediction range.


trivialfis:
Let me add support for mixed trees

PR #8985, opened 2023-03-27
More tests for column split and vertical federated learning

rongou:
Added some more tests for the learner and fit_stump, for both column-wise distributed learning and vertical federated learning.

Also moved the `IsRowSplit` and `IsColumnSplit` methods from the `DMatrix` to the `MetaInfo` since in some places we only have access to the `MetaInfo`. Added a new convenience method `IsVerticalFederatedLearning`.

Some refactoring of the testing fixtures.


rongou:
@trivialfis @hcho3

trivialfis (approved):

PR #8995, opened 2023-03-29
Fix aft_loss_distribution documentation

rongou:
Turns out `aft_loss_distribution` is not really a loss, but a parameter to the `survival:aft` loss.


rongou:
@trivialfis @hcho3

hcho3 (approved):

PR #8999, opened 2023-03-30
Portable macro definition.

trivialfis:
Fix the following warning from gcc:
```
//rabit/src/allreduce_base.o
   In file included from ../..//rabit/include/rabit/internal/socket.h:8,
                    from ../..//rabit/src/allreduce_base.h:22,
                    from ../..//rabit/src/allreduce_base.cc:14:
   ../..//include/xgboost/collective/socket.h:65:22: warning: this use of "defined" may not be portable [-Wexpansion-to-defined]
      65 | #if xgboost_IS_MINGW()
         |                      ^
```

### DIFF
--- a/include/xgboost/collective/socket.h
+++ b/include/xgboost/collective/socket.h
@@ -1,11 +1,11 @@
-/*!
- * Copyright (c) 2022 by XGBoost Contributors
+/**
+ * Copyright (c) 2022-2023, XGBoost Contributors
  */
 #pragma once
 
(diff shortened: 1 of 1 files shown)


trivialfis:
I still don't know where is the non-reproducible failure coming from https://buildkite.com/xgboost/xgboost-ci/builds/2088#0187320b-aa43-4e7c-958a-639d57dc7428 . I reproduced it only once on an ARM device.

RAMitchell (approved):

PR #9002, opened 2023-03-30
Make objectives work with vertical distributed and federated learning

rongou:
Went through all the objectives and made sure they work with both column-split distributed training and vertical federated learning.

### DIFF
--- a/src/objective/adaptive.h
+++ b/src/objective/adaptive.h
@@ -36,13 +36,15 @@ inline void FillMissingLeaf(std::vector<bst_node_t> const& maybe_missing,
 }
 
 inline void UpdateLeafValues(std::vector<float>* p_quantiles, std::vector<bst_node_t> const& nidx,
-                             float learning_rate, RegTree* p_tree) {
+                             MetaInfo const& info, float learning_rate, RegTree* p_tree) {
   auto& tree = *p_tree;
   auto& quantiles = *p_quantiles;
   auto const& h_node_idx = nidx;
 
   size_t n_leaf{h_node_idx.size()};
-  collective::Allreduce<collective::Operation::kMax>(&n_leaf, 1);
+  if (info.IsRowSplit()) {
+    collective::Allreduce<collective::Operation::kMax>(&n_leaf, 1);
+  }
   CHECK(quantiles.empty() || quantiles.size() == n_leaf);
   if (quantiles.empty()) {
     quantiles.resize(n_leaf, std::numeric_limits<float>::quiet_NaN());
@@ -52,12 +54,16 @@ inline void UpdateLeafValues(std::vector<float>* p_quantiles, std::vector<bst_no
   std::vector<int32_t> n_valids(quantiles.size());
   std::transform(quantiles.cbegin(), quantiles.cend(), n_valids.begin(),
                  [](float q) { return static_cast<int32_t>(!std::isnan(q)); });
-  collective::Allreduce<collective::Operation::kSum>(n_valids.data(), n_valids.size());
+  if (info.IsRowSplit()) {
+    collective::Allreduce<collective::Operation::kSum>(n_valids.data(), n_valids.size());
+  }
   // convert to 0 for all reduce
   std::replace_if(
       quantiles.begin(), quantiles.end(), [](float q) { return std::isnan(q); }, 0.f);
   // use the mean value
-  collective::Allreduce<collective::Operation::kSum>(quantiles.data(), quantiles.size());
+  if (info.IsRowSplit()) {
+    collective::Allreduce<collective::Operation::kSum>(quantiles.data(), quantiles.size());
+  }
   for (size_t i = 0; i < n_leaf; ++i) {
     if (n_valids[i] > 0) {
       quantiles[i] /= static_cast<float>(n_valids[i]);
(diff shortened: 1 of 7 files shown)


rongou:
@trivialfis @hcho3

src/objective/quantile_obj.cu
```diff
@@ -35,7 +35,7 @@ class QuantileRegression : public ObjFunction {
   bst_target_t Targets(MetaInfo const& info) const override {
     auto const& alpha = param_.quantile_alpha.Get();
     CHECK_EQ(alpha.size(), alpha_.Size()) << "The objective is not yet configured.";
-    CHECK_EQ(info.labels.Shape(1), 1) << "Multi-target is not yet supported by the quantile loss.";
+    CHECK_LE(info.labels.Shape(1), 1) << "Multi-target is not yet supported by the quantile loss.";
```
trivialfis, line 38: Under which case this can be 0? Would be great if we can make sure it's greater or equal to 1 and only allow the first dimension (rows) to be zero.

trivialfis (commented):
I'm not quite convinced that only one of the participants can have access to labels. We will have to maintain lots of these conditions in the future.

src/objective/quantile_obj.cu
rongou: Fixed.

tests/cpp/plugin/test_federated_learner.cc
```diff
@@ -13,66 +13,119 @@
...
+
+    Json model{Object{}};
+    learner->SaveModel(&model);
+
+    RunWithFederatedCommunicator(kWorldSize, server_address_, &VerifyBaseScoreAndModel, kRows,
+                                 kCols, objective, base_score, model);
+  }
+};
+
+TEST_F(FederatedLearnerTest, RegSquaredError) { TestObjective("reg:squarederror"); }
```
trivialfis, line 94: I think this https://github.com/dmlc/xgboost/blob/4caca2947df22fa76af27bbaad379aafe9f6c51d/tests/cpp/objective/test_objective.cc#L30 would be more future-proofed.

tests/cpp/test_learner.cc
```diff
@@ -608,31 +608,104 @@ TEST_F(InitBaseScore, InitWithPredict) { this->TestInitWithPredt(); }
...
+
+    Json model{Object{}};
+    learner->SaveModel(&model);
+
+    auto constexpr kWorldSize{3};
+    RunWithInMemoryCommunicator(kWorldSize, &TestColumnSplit, dmat, objective, base_score, model);
+  }
+};
+
+TEST_F(ColumnSplit, RegSquaredError) { this->TestBaseScoreAndModel("reg:squarederror"); }
```
trivialfis, line 674: Ditto.

src/objective/quantile_obj.cu
```diff
@@ -167,8 +170,10 @@ class QuantileRegression : public ObjFunction {
     common::Mean(ctx_, *base_score, &temp);
     double meanq = temp(0) * sw;
 
-    collective::Allreduce<collective::Operation::kSum>(&meanq, 1);
-    collective::Allreduce<collective::Operation::kSum>(&sw, 1);
+    if (info.IsRowSplit()) {
```
trivialfis, line 173: Maybe we should make a wrapper function for `Allreduce` so that newcomers to the code base can understand these conditions?  I think we need some separation for communication logic and machine learning algorithm implementation. I'm open to any idea, at the moment it's quite difficult for a "machine learning person" to add a new algorithm and make sure it works correctly with different distributed system requirements. I can see that these conditions are necessary for metrics as well since they require labels. Is there a way to make this less intrusive by abstracting the logic under collective instead of exposing it to the machine learning part?

src/objective/adaptive.h
```diff
@@ -97,8 +103,39 @@ inline void UpdateTreeLeaf(Context const* ctx, HostDeviceVector<bst_node_t> cons
...
+        }
+        collective::Broadcast(static_cast<void*>(leaf_values.data()),
+                              leaf_values.size() * sizeof(bst_float), 0);
+      } else {
+        std::vector<bst_float> leaf_values(p_tree->GetNumLeaves());
+        collective::Broadcast(static_cast<void*>(leaf_values.data()),
+                              leaf_values.size() * sizeof(bst_float), 0);
+        auto i = 0;
+        auto& tree = *p_tree;
+        for (auto nid = 0; nid < tree.NumNodes(); nid++) {
```
trivialfis, line 128: The mapping between `GetNodes` and `nid` is not clear to me. There's a update leaf value function called by but host and device after calculating the quantiles and it has access to leaf value, consider moving the logic there.

Also, please consider abstracting this logic away from the main algorithm. I find it difficult to consider all possible combinations of conditions.

trivialfis (commented):
Initial review. I'm a bit concerned about the generality of this dispatching. Is there a way to hide it in the collective instead of exposing it to the main algorithms?

Looking for ideas, @RAMitchell @hcho3 .

src/objective/adaptive.h
rongou: Done.

src/objective/quantile_obj.cu
rongou: The problem is `Allreduce` is a low level primitive, but column-wise split and vertical federated learning are higher level semantic changes, so they don't always map to each other. At the level of the communicator, I don't think we can determine whether we should skip a particular `Allreduce` call, even if we know what distributed/federated environment we are in.

Maybe we can come up with a clever mechanism to make this cleaner. Perhaps as a follow up?

tests/cpp/plugin/test_federated_learner.cc
rongou: Done.

tests/cpp/test_learner.cc
rongou: Done.

PR #9006, opened 2023-04-03
Convert federated tests into test suite.

trivialfis:
- Add specialization for learning to rank.

After #8822, only integer label is accepted for `ndcg` and `pairwise`, and only binary is accepted for `map`. This PR assigns binary label for LTR, along with which, the tests are converted into gtest parametrized tests.


trivialfis:
cc @rongou

rongou (approved):

PR #9007, opened 2023-04-03
Fix IsBinaryRel() check

cbalint13:
This small PR fixes wrong binary checks for labels.

* The behaviour ~~bug~~ was introduced by https://github.com/dmlc/xgboost/pull/8931
* The checks ~~should be~~ now are in range ```1.0+eps > label > 0.0-eps```, with ```eps = 1e-6```

----

The errors was cought on TVM autotune process:
```
File "/usr/lib64/python3.11/site-packages/tvm/autotvm/tuner/xgboost_cost_model.py"
, line 538, in after_iteration
    bst_eval = model.eval_set(self.evals, epoch, feval)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/lib64/python3.11/site-packages/xgboost/core.py", line 1995, in eval_set
    _check_call(
File "/usr/lib64/python3.11/site-packages/xgboost/core.py", line 270, in _check_call
xgboost.core.XGBoostError: [15:59:59] /builddir/build/BUILD/xgboost/src/common/ranking_utils.h:378: 
Check failed: is_binary: MAP can only be used with binary labels.
```

---

Thank you,
~Cristian.

Cc: @trivialfis , @RAMitchell , please help with the review.

### DIFF
--- a/src/common/ranking_utils.h
+++ b/src/common/ranking_utils.h
@@ -362,7 +362,7 @@ template <typename AllOf>
 bool IsBinaryRel(linalg::VectorView<float const> label, AllOf all_of) {
   auto s_label = label.Values();
   return all_of(s_label.data(), s_label.data() + s_label.size(), [] XGBOOST_DEVICE(float y) {
-    return std::abs(y - 1.0f) < kRtEps || std::abs(y - 0.0f) < kRtEps;
+    return ((y - 1.0f) < kRtEps) && ((y - 0.0f) > -kRtEps);
   });
 }
 /**
@@ -373,7 +373,6 @@ bool IsBinaryRel(linalg::VectorView<float const> label, AllOf all_of) {
  */
 template <typename AllOf>
 void CheckMapLabels(linalg::VectorView<float const> label, AllOf all_of) {
-  auto s_label = label.Values();
   auto is_binary = IsBinaryRel(label, all_of);
   CHECK(is_binary) << "MAP can only be used with binary labels.";
 }


trivialfis:
Hi, the check was for `y \in {0, 1}` or `y == 0.0f || y == 1.0f`, it's written this way for being compatible with floating point.

cbalint13:
> Hi, the check was for `y \in {0, 1}` or `y == 0.0f || y == 1.0f`, it's written this way for being [compatible with](https://isocpp.org/wiki/faq/newbie#floating-point-arith) floating point.

@trivialfis 

* So you mean ```y``` can be only 0.0 (with small eps margin) **or** ```y``` can be only 1.0 (with small eps margin) ?

If so,  I should reconsider the use case.

cbalint13:
> In that case, we can threshold the label using 0.5. But it's not as rigorous as one might prefer.

I'll do so, will see if resulted MAP impacts beaviour of TVM's internal model evaluation.
To close the discussion I leave here a  way of inplace ```binarization``` as per your suggestion.

```
 dtrain = xgb.DMatrix(data, label=label)
 lbl = dtrain.get_label()
 lbl[lbl>=.5] = 1.0
 lbl[lbl<0.5] = 0.0
 dtrain.set_label(lbl)
```

Thanks @trivialfis !
~Cristian.

tqchen:
@cbalint13 perhaps a better way would be to move away from the map metric. Since likely we need continuous regression ranking. I assume pair-wise or other metrics could still work for such cases.

@trivialfis can you confirm if that is the case?

tqchen:
It would still be useful to have rank values being continuous. The particular use case in TVM was cost model prediction, where we only care about the relative rank of the cost but not their accurate values. 

As a result, the ability to be able to specify the ground truth label sample, rather than the pairs is useful as there can be many such pairs in total.

cbalint13:
@trivialfis ,

> In the last couple PRs, I have changed NDCG to integer only, meaning it accepts only discrete labels, with an additional option for user to specify custom gain function, which can be continuous. The custom gain function still needs a few more PRs to be merged.

I am interested to track it, is there a tracker for the upcoming custom gain function (as pr or issue) ?

trivialfis:
https://github.com/dmlc/xgboost/pull/8822 is the WIP PR, I'm extracting small parts of it for review and merge. 

For continuous rank value, NDCG with custom gain seems to be the right choice. Will keep you updated with the progress.

PR #9020, opened 2023-04-10
Make sure metrics work with column-wise distributed training

rongou:
This is mostly moving things around so that we can test the metrics under distributed training, for both CPU and GPU, row split and column split.


rongou:
@trivialfis I'll follow up with some refactoring to make the logic for row vs. column split, and horizontal vs vertical federated learning more encapsulated.

tests/cpp/metric/test_auc.cc
```diff
@@ -1,261 +1,68 @@
+#include "test_auc.h"
+
 #include <xgboost/metric.h>
-#include "../helpers.h"
 
 namespace xgboost {
 namespace metric {
 
-TEST(Metric, DeclareUnifiedTest(BinaryAUC)) {
-  auto ctx = xgboost::CreateEmptyGenericParam(GPUIDX);
```
trivialfis, line 8: Is it necessary to move all these code to headers? Maybe copying the function declaration is enough?

tests/cpp/helpers.h
```diff
@@ -39,6 +39,18 @@
...
+#if defined(__CUDACC__)
+#define DeclareUnifiedDistributedTest(name) MGPU ## name
+#else
+#define DeclareUnifiedDistributedTest(name) name
+#endif
+
+#if defined(__CUDACC__)
+#define ALL_GPUS (xgboost::common::AllVisibleGPUs())
+#else
+#define ALL_GPUS (3)
```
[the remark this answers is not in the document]
rongou: This is actually for the CPU tests. Maybe should call this `WORLD_SIZE` to make it clearer?

tests/cpp/metric/test_auc.cc
rongou: I think the problem is since we are including the `.cc` file in `.cu`, we'd get duplicate symbols if we put the implementation in the `.cc` file.

trivialfis: Since we are declaring all the test bodies in headers, I think we can remove the cc file include trick right?

tests/cpp/helpers.h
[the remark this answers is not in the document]
rongou: Done.

tests/cpp/metric/test_auc.cc
rongou: In most cases the tests are the same across CPU, GPU, row split distributed, and column split distributed, but in a few places (I think they are all in `elementwise`), the row-split distributed results are actually different from single node, which makes sense since we are summing some intermediate data across all workers.

Maintaining the tests shouldn't be too different from before, they are just in the `.h` files instead of `.cc`. The distributed tests are mostly guardrails to make sure we don't accidentally break them.

trivialfis: Thank you for sharing, I understand the difference created by row-split, but for column split, the result should be the same as single node right?

rongou: Right, for column-split distributed training, metric calculation should be the same as single node.

trivialfis: Would testing that property be easier than the current tests? My thought is we can use something similar to the objective tests for federated learning instead of adding new parameters to the existing metric tests.

rongou: The objective tests are at a higher level, we only run through each objective once to obtain the result, so it's simpler to compare it with the result from column split. The metric tests are lower level, we are calculating the metric result repeatedly under different conditions, so it's harder to compare the results. I suppose we could refactor the tests more to make it easier to do comparisons, but with the current setup, this seems like the simplest approach.

trivialfis: Apologies for the many questions, can we make it high-level? What's the downside of doing so
```
for m : all_metrics:
    local_data = random()
    column_split_data = copy(local_data)
    column_split_data.info.column_split = true

    expected_result = m.evaluate(local_data)
    column_split_result = run_with_communicator(world_size=3, []() { return m.evaluate(column_split_data) })
    assert_equal(expected_result, column_split_result)
```

rongou: The main issue is for each metric, we are testing many variations of inputs (each `GetMetricEval` call is different test case). In order to compare single node vs column split, we'd have to either:
* Only test a single input per metric, a happy path, thus missing a lot of corner cases.
* Or still test all the inputs, which means repeating all the tests, introducing a lot of duplication.

Since we've already calculated the expected metric outputs, there is really no need to compare column-split results with single-node ones.

This is slightly different from the objective tests, where the outputs depend on the randomly generated data. For metric tests we are manually specifying predictions and labels, so the outputs are fixed.

trivialfis (approved):

tests/cpp/metric/test_auc.cc
trivialfis: Got it. For future references, it would be great if we can hide or make it more automatic (like having a metric test helper that can generate different test cases). Personally, I would like to keep simple things like metric to be beginner friendly.

PR #9021, opened 2023-04-11
Replace setup.py with pyproject.toml

### LINKED ISSUE #8090: [python] 'setup.py install' is deprecated

hcho3:
Closes #8090

* Create `pyproject.toml`
* Implement a custom build backend (see below) in `packager` directory. Build logic from `setup.py` has been refactored and migrated into the new backend.
* Tested: `pip wheel .` (build wheel), `python -m build --sdist .` (source distribution)

TODOs
- [x] Update CI to use new build commands.
- [x] Write a doc for building wheels and sdists. Also show how to set CMake args:
```
pip install -v . --config-setting use_system_libxgboost=True
```
- [x] Handle CMake args in `packager/pep517.py`.
- [x] Test editable installation.
- [x] Add typing annotation to `packager`.

Context
* Directly calling `setup.py` for installing the Python package is deprecated. `python setup.py install` will throw a warning:
```
SetuptoolsDeprecationWarning: setup.py install is deprecated.
Use build and pip and other standards-based tools.
```
* The new replacement for `setup.py` is PEP 517 and PEP 621.
  - PEP 621: The package's metadata is now specified in `pyproject.toml`. The TOML file is declarative and cannot contain arbitrary code unlike `setup.py`. 
  - PEP 517: It is now possible to specify a **build backend** in `pyproject.toml`:
```
[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"
```

* Users now can choose between a variety of build tools: `setuptools`, `poetry`, `hatch` etc. In addition, individual Python packages can specify a custom (bespoke) build backend by specifying `backend-path`:
```
[build-system]
requires = ["packaging"]
backend-path = ["xgboost"]
build-backend = "packager.pep517"
```
A build backend implements two methods:
  - `def build_wheel(wheel_directory, config_settings=None, metadata_directory=None)`
  - `def build_sdist(sdist_directory, config_settings=None)`

### DIFF
--- a/doc/contrib/python_packaging.rst
+++ b/doc/contrib/python_packaging.rst
@@ -77,7 +77,7 @@ completes quickly:
   
   $ pip install xgboost-2.0.0-py3-none-linux_x86_64.whl  # Completes quickly
 
-.. rubric:: Foonotes
+.. rubric:: Footnotes
 
 .. [#shared_lib_name] The name of the shared library file will differ
    depending on the operating system in use. See :ref:`build_shared_lib`.
--- a/python-package/packager/nativelib.py
+++ b/python-package/packager/nativelib.py
@@ -28,7 +28,6 @@ def _lib_name() -> str:
 
 def build_libxgboost(
     cpp_src_dir: pathlib.Path,
-    *,
     build_dir: pathlib.Path,
     build_config: BuildConfiguration,
 ) -> pathlib.Path:
@@ -111,7 +110,6 @@ def build_libxgboost(
 
 def locate_local_libxgboost(
     toplevel_dir: pathlib.Path,
-    *,
     logger: logging.Logger,
 ) -> Optional[pathlib.Path]:
     """
@@ -126,7 +124,6 @@ def locate_local_libxgboost(
 
 def locate_or_build_libxgboost(
     toplevel_dir: pathlib.Path,
-    *,
     build_dir: pathlib.Path,
     build_config: BuildConfiguration,
 ) -> pathlib.Path:
--- a/python-package/packager/pep517.py
+++ b/python-package/packager/pep517.py
@@ -147,7 +147,7 @@ def build_editable(
         )
 
     if locate_local_libxgboost(TOPLEVEL_DIR, logger=logger) is None:
-        raise AssertionError(
+        raise RuntimeError(
             "To use the editable installation, first build libxgboost with CMake. "
             "See https://xgboost.readthedocs.io/en/latest/build.html for detailed instructions."
         )
--- a/python-package/packager/sdist.py
+++ b/python-package/packager/sdist.py
@@ -8,7 +8,7 @@ from .util import copy_with_logging, copytree_with_logging
 
 
 def copy_cpp_src_tree(
-    cpp_src_dir: pathlib.Path, target_dir: pathlib.Path, *, logger: logging.Logger
+    cpp_src_dir: pathlib.Path, target_dir: pathlib.Path, logger: logging.Logger
 ) -> None:
     """Copy C++ source tree into build directory"""
 
--- a/python-package/packager/util.py
+++ b/python-package/packager/util.py
@@ -7,7 +7,7 @@ import shutil
 
 
 def copytree_with_logging(
-    src: pathlib.Path, dest: pathlib.Path, *, logger: logging.Logger
+    src: pathlib.Path, dest: pathlib.Path, logger: logging.Logger
 ) -> None:
     """Call shutil.copytree() with logging"""
     logger.info("Copying %s -> %s", str(src), str(dest))
@@ -15,7 +15,7 @@ def copytree_with_logging(
 
 
 def copy_with_logging(
-    src: pathlib.Path, dest: pathlib.Path, *, logger: logging.Logger
+    src: pathlib.Path, dest: pathlib.Path, logger: logging.Logger
 ) -> None:
     """Call shutil.copy() with logging"""
     if dest.is_dir():


hcho3:
@jameslamb You might be interested in this work.

jameslamb:
Thanks for tagging me! I'm very interested in this and happy to provide a review or answer questions if you want some extra help.

For reference, in LightGBM I'm pursuing using `scikit-build-core` instead of having a custom build backend: https://github.com/microsoft/LightGBM/pull/5759. But only because I don't know much about the custom build backends. I'll read through the approach you took here.

trivialfis (commented):
Just out of curiosity, is it possible that we split the XGBoost package into two like what the conda forge package does? One as libxgboost and another one as (py)xgboost.

CMake is now on [pypi](https://pypi.org/project/cmake/), which has two implications. Firstly it's a C++ project with only a CLI interface. The mentioned libxgboost would be similar: a C++ project with only a C interface. Secondly, I am not sure if there is a distinction between build-time and run-time dependency for toml projects. If so then we can have cmake as a build-time dependency for the source distribution.

jameslamb:
> I am not sure if there is a distinction between build-time and run-time dependency for toml projects

There is! This is one of the nice benefits of the new backends... you don't have to have something like `setuptools` or `wheel` as a runtime dependency.

Also on the topic of CMake being available on PyPI, two very important things to consider:

1. anyone running `pip install cmake` on a platform for which there isn't a `CMake` wheel is going end up compiling it from source, which can be really painful and definitely more difficult than installing it from package managers with broader platform support like `brew` on macOS, `apt` on various Linux flavors
2. because of that, `scikit-build-core` has the behavior "if an already-installed `cmake` is found on `PATH`, use that... otherwise, install the one from PyPI"

vyasr:
> Secondly, I am not sure if there is a distinction between build-time and run-time dependency for toml projects. 

It's worth noting that this is in fact one of the motivating reasons for the creation of the modern build infrastructure, namely PEP 517 as previously mentioned as well as [PEP 518](https://peps.python.org/pep-0518/), which defines the `build-system` part of `pyproject.toml`. The canonical issue this was designed to fix is a package that requires numpy and/or Cython to be installed in order to run setup.py in order to provide numpy headers (`np.get_include`) or Cythonize files, respectively. Quoting the PEP:

> But when a project chooses to use setuptools, the use of an executable file like setup.py becomes an issue. You can’t execute a setup.py file without knowing its dependencies, but currently there is no standard way to know what those dependencies are in an automated fashion without executing the setup.py file where that information is stored. It’s a catch-22 of a file not being runnable without knowing its own contents which can’t be known programmatically unless you run the file.

The `build-system` section of `pyproject.toml` allows specifying the requirements for the build. PEP 517 builds occur in a self-contained virtual environment where all build requirements are installed. Run requirements are a separate list that is installed into the user's environment along with the built package after a build is complete (when using `pip install .` to build locally) or a pre-built binary is found (`pip install ${pkg}`). The contained build is analogous to how conda builds work (in venvs instead of conda envs).

trivialfis:
It's not necessary to be useful, just a way for us to split c++ packaging from the Python package.

hcho3:
@trivialfis Let's not split the packaging. It will introduce headaches, since different virtual env managers for Python install native libs in different locations. We don't want to make `libpath.py` even more complicated than it is now.

hcho3:
@vyasr I managed to use `hatchilng` backend (with some customization). As a result, I was able to remove the boilerplate code, such as the metadata generation. Now XGBoost's build backend only contains the necessary customization. Thanks for your advice!

hcho3:
@trivialfis Editable installation is now supported. Run: `pip install -e ."

hcho3:
@jameslamb To give you some context, I chose a happy middle path between implementing a backend from scratch and using an existing backend. I built on top of `hatchling`, which led to modular code with minimal amount of boilerplate.

trivialfis (commented):
Thank you for working on the Python build system @hcho3 , I don't have enough expertise to make technical comments. It would be great if you can update the document for the new build system https://github.com/dmlc/xgboost/blob/master/doc/build.rst#building-python-package-from-source .

hcho3:
Good news, I am able to remove `insert_vcomp140.py` and fold it into the new build backend.

trivialfis:
> Good news, I am able to remove insert_vcomp140.py and fold it into the new build backend.

We might want to make it optional. I'm not sure what's the implication of doing that for systems like conda.

hcho3:
@trivialfis Good point, let me add a build option then.

hcho3:
I am removing `vcomp140.dll` from the XGBoost wheel. I just tested the wheel on a minimal installation of Windows, and `import xgboost` fails with

> Error message(s): ["Could not find module 'C:\\Users\\Administrator\\Downloads\\foobar\\lib\\site-packages\\xgboost\\lib\\xgboost.dll' (or one of its dependencies). Try using the full path with constructor syntax."]

I tried a couple things to salvage the error, but none worked.

* Preload dependent DLLs using [the approach taken by scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/66a4d9639e27bc04c21bac4eb259fe28b1dad4af/build_tools/github/vendor.py#L48). The approach only works with compiled extension but not vanilla C lib that's loaded via `ctypes.cdll.LoadLibrary()`.
* Use `delvewheel` to patch the wheel. The import still failed.
* Pass `winmode` arg to `ctypes.CDLL`. No avail.
* I tested both the new wheel and 1.7.5 version. All failed.

Conclusion: Let's not bundle `vcomp140.dll` at all, since it doesn't work. Instead, let's tell Windows users to install Visual C++ Redistributable package. This is what LightGBM does. Alternatively, they can use Conda which handles this automatically.

trivialfis:
note to myself: need to update the release script.

hcho3:
@trivialfis I updated `dev/release-artifacts.py`. Can you take a look?

tests/ci_build/lint_python.py
```diff
@@ -219,6 +219,8 @@ def main(args: argparse.Namespace) -> None:
                 "tests/ci_build/test_r_package.py",
                 "tests/ci_build/test_utils.py",
                 "tests/ci_build/change_version.py",
+                # Python packaging
+                "python-package/packager/",
```
trivialfis, line 223: maybe merge this with `python-package/xgboost` and just call mypy with `python-package`?

trivialfis (approved):
The change looks good to me. Thank you again for working on this! I'm not entirely sure how all the tools interact with each other. Seems complicated as there's code generation during build/install. Would be great if others can take a look as well.

tests/ci_build/lint_python.py
hcho3: @trivialfis I thought you didn't want to distribute `packager` as part of the Python wheel?

trivialfis: Apologies for the ambiguity, I meant running the `mypy` on `xgboost/python-package` on the CI, not distributing `packager`  with xgboost.


https://github.com/dmlc/xgboost/blob/191d0aa5cfac87cfc122ebbe8b14da4177309d13/tests/ci_build/lint_python.py#L201

hcho3: If `packager` is not under directory `xgboost/python-package`, we'd need to specify `packager` as a separate entry. But if we move `packager` to `xgboost/python-package`, then we'd be distributing `packager` with XGBoost.

I prefer the first alternative.

hcho3: My bad, it's sufficient to call `mypy python-package/`

python-package/packager/pep517.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def write_hatch_config(dest_dir: pathlib.Path, *, logger: logging.Logger) -> None:
+    """Write a small custom hook for Hatch, to set a custom tag."""
+    template = (
+        "from hatchling.builders.hooks.plugin.interface import BuildHookInterface\n"
+        "class CustomBuildHook(BuildHookInterface):\n"
+        "    def initialize(self, version, build_data):\n"
+        "        build_data['tag'] = '{tag}'\n"
+    )
```
vyasr, line 76: You should be able to write this a little more neatly using textwrap.dedent and a multiline string to get the desired formatting
```suggestion
    template = textwrap.dedent("""
        from hatchling.builders.hooks.plugin.interface import BuildHookInterface
        class CustomBuildHook(BuildHookInterface):
            def initialize(self, version, build_data):
                build_data['tag'] = '{tag}'"""
    )
```

python-package/packager/pep517.py
```diff
@@ -0,0 +1,191 @@
...
+    hatch_build_file = dest_dir / "hatch_build.py"
+    logger.info("Writing %s", str(hatch_build_file))
+    with open(hatch_build_file, "w", encoding="utf-8") as f:
+        f.write(template.format(tag=get_tag()))
```
vyasr, line 80: Could skip the format by using an f-string above instead of embedding a format specifier `{tag}` to be replaced later.

python-package/packager/pep517.py
```diff
@@ -0,0 +1,191 @@
...
+    metadata_directory: Optional[str] = None,
+) -> str:
+    """Build an editable installation. We mostly delegate to Hatchling."""
+    logger = logging.getLogger("xgboost.packager.build_editable")
+
+    if config_settings:
+        raise NotImplementedError(
+            "XGBoost's custom build backend doesn't support config_settings option "
+            f"when building editable installation. {config_settings=}"
+        )
```
vyasr, line 180: More of a future problem than anything worth dealing with now, but in line with my earlier suggestion to split non-CMake and CMake config options it may eventually be worth splitting the config settings that are for the Python vs C++ build since the former may be supportable for editable builds too.

python-package/packager/pep517.py
```diff
@@ -0,0 +1,191 @@
...
+
+        copytree_with_logging(
+            TOPLEVEL_DIR / "xgboost", workspace / "xgboost", logger=logger
+        )
+        copytree_with_logging(
+            TOPLEVEL_DIR / "packager", workspace / "packager", logger=logger
+        )
+
+        temp_cpp_src_dir = workspace / "cpp_src"
+        copy_cpp_src_tree(cpp_src_dir, target_dir=temp_cpp_src_dir, logger=logger)
```
vyasr, line 161: IMHO it seems kind of arbitrary to define the copying of the cpp subtree in a separate function in another module when the lines above this are all hardcoding paths to copy anyway. I would personally prefer to see this function inlined, but up to you.

python-package/packager/nativelib.py
```diff
@@ -0,0 +1,158 @@
...
+
+    libxgboost = locate_local_libxgboost(toplevel_dir, logger=logger)
+    if libxgboost is not None:
+        return libxgboost
+    if build_config.use_system_libxgboost:
+        # Find libxgboost from system prefix
+        sys_prefix = pathlib.Path(sys.prefix).absolute().resolve()
+        libxgboost = sys_prefix / "lib" / _lib_name()
+        if not libxgboost.exists():
+            raise AssertionError(
```
vyasr, line 142:
```suggestion
            raise RuntimeError(
```
or maybe a ValueError?

tests/buildkite/build-cpu-arm64.sh
```diff
@@ -18,7 +18,7 @@ $command_wrapper bash -c "cd build && ctest --extra-verbose"
 
 echo "--- Build binary wheel"
 $command_wrapper bash -c \
-  "cd python-package && rm -rf dist/* && python setup.py bdist_wheel --universal"
+  "cd python-package && rm -rf dist/* && pip wheel --no-deps -v . --wheel-dir dist/"
```
vyasr, line 21: Minor note, since you're using `pip wheel --no-deps` everywhere you may be able to use `python -m build -w` as a drop-in replacement for the most part.

vyasr: You may also be able to use that in your documentation and CI scripts, but I don't have a strong preference either way there.

python-package/packager/pep517.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def get_requires_for_build_editable(
+    config_settings: Optional[Dict[str, Any]] = None
+) -> List[str]:
+    """A PEP 517 method. Delegate to Hatchling"""
+    return hatchling.build.get_requires_for_build_editable(config_settings)
+
+
+def write_hatch_config(dest_dir: pathlib.Path, *, logger: logging.Logger) -> None:
```
vyasr, line 69: Rather than writing this file as a temporary, can't you just commit the `hatch_build.py` file to the repo with the `get_tag` function in that file to be used by the `initialize` function. Then the file would always be present and it would be copied into the temporary dir in `build_wheel` too, right?

vyasr (changes requested):
This looks good! I'm not too familiar with hatch, so I needed to dig a bit through that backend to review some of the more internal components, but this approach seems pretty reasonable to me.

doc/contrib/python_packaging.rst
```diff
@@ -0,0 +1,53 @@
...
+#########################
+
+
+.. contents:: Contents
+  :local:
+
+.. _binary_wheels:
+
+***************************************************
+How to build binary wheels and source distributions
```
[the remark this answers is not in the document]
hcho3: @vyasr This document is specifically written for XGBoost developers. So running `pip wheel` is expected to produce a binary wheel.

doc/contrib/python_packaging.rst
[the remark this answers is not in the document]
hcho3: This document is not meant for general audience. It's specifically meant for XGBoost developers. Let me revise the title to `How to package XGBoost Python package` to make this clear.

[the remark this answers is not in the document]
hcho3: I'll revise other parts of this document so that it's more geared towards packaging XGBoost specifically.

doc/contrib/python_packaging.rst
```diff
@@ -0,0 +1,53 @@
...
+***************************************************
+
+Binary wheels and source distributions (sdist for short) are two main
```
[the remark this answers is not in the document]
hcho3: I should remove the "binary" bit from this sentence here, since the sentence applies generally to all Python packages.

I'll also add a sentence somewhere saying that, for XGBoost specifically, `pip wheel` produces a binary wheel due to the existence of a compiled binary.

python-package/packager/build_config.py
```diff
@@ -0,0 +1,57 @@
...
+        """Parse config_settings from Pip (or other PEP 517 frontend)"""
+        if config_settings is not None:
+            for field_name in [x.name for x in dataclasses.fields(self)]:
+                self._set_config_setting(config_settings, field_name)
+
+    def get_cmake_args(self) -> List[str]:
+        """Convert build configuration to CMake args"""
+        cmake_args = []
+        for field_name in [x.name for x in dataclasses.fields(self)]:
+            if field_name in ["use_system_libxgboost"]:
```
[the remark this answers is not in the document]
hcho3: We don't plan to add more CMake flags (or any flags) to the build configuration. The existing CMake flags were moved over due to legacy uses. XGBoost core supports many more CMake flags, and the best way for a developer to use these flags is to invoke CMake manually inside a `build/` directory.

doc/install.rst
```diff
@@ -16,15 +16,28 @@ Stable Release
 Python
 ------
 
-Pre-built binary are uploaded to PyPI (Python Package Index) for each release.  Supported platforms are Linux (x86_64, aarch64), Windows (x86_64) and MacOS (x86_64, Apple Silicon).
+Pre-built binary wheels are uploaded to PyPI (Python Package Index) for each release. Supported platforms are Linux (x86_64, aarch64), Windows (x86_64) and MacOS (x86_64, Apple Silicon).
 
 .. code-block:: bash
 
+  # Pip 21.3+ is required
```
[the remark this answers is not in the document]
hcho3: XGBoost requires an in-tree build, which became the default since Pip 21.3+.

From Pip's changelog:

> 21.3 (2021-10-11)
> Deprecations and Removals
> * In-tree builds are now the default. --use-feature=in-tree-build is now ignored. --use-deprecated=out-of-tree-build may be used temporarily to ease the transition. ([#10495](https://github.com/pypa/pip/issues/10495))

python-package/packager/build_config.py
```diff
@@ -0,0 +1,57 @@
...
+    use_s3: bool = False
+    # Whether to enable the dense parser plugin
+    plugin_dense_parser: bool = False
+    # Special option: See explanation below
+    use_system_libxgboost: bool = False
+
+    def _set_config_setting(
+        self, config_settings: Dict[str, Any], field_name: str
+    ) -> None:
+        if field_name in config_settings:
```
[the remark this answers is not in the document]
hcho3: Good point, we should probably throw an error if an unknown setting is passed in.

python-package/packager/build_config.py
```diff
@@ -0,0 +1,57 @@
...
+                self._set_config_setting(config_settings, field_name)
+
+    def get_cmake_args(self) -> List[str]:
+        """Convert build configuration to CMake args"""
+        cmake_args = []
+        for field_name in [x.name for x in dataclasses.fields(self)]:
+            if field_name in ["use_system_libxgboost"]:
+                continue
+            cmake_option = field_name.upper()
+            cmake_value = "ON" if getattr(self, field_name) else "OFF"
```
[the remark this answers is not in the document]
hcho3: > all the getattr usage in this file is probably unnecessary since the attributes will always exist

No, I'm not checking the existence of each attribute. I am simply testing each attribute's Boolean value for True/False. Let me revise the statement to
```python
cmake_value = "ON" if getattr(self, field_name) == True else "OFF"
```
to make it clear.

python-package/packager/nativelib.py
```diff
@@ -0,0 +1,158 @@
...
+    else:
+        if shutil.which("ninja"):
+            build_tool = "ninja"
+        else:
+            build_tool = "make"
+        generator = "-GNinja" if build_tool == "ninja" else "-GUnix Makefiles"
+        try:
+            _build(generator=generator, build_tool=build_tool)
+        except subprocess.CalledProcessError as e:
+            logger.info("Failed to build with OpenMP. Exception: %s", str(e))
```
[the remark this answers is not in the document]
hcho3: Some systems (like MacOS) don't have OpenMP installed by default. So in this case, installing XGBoost from the source should fall back to building with OpenMP turned off.

python-package/packager/pep517.py
hcho3: I put it as a separate module, so that we can easily add more subdirectories for C++ code in the future. If we put the copying logic in `pep517.py`, it would give us more cognitive overhead when we add more C++ files in the future.

python-package/packager/nativelib.py
```diff
@@ -0,0 +1,158 @@
...
+        if "CIBW_TARGET_OSX_ARM64" in os.environ:
+            cmake_cmd.append("-DCMAKE_OSX_ARCHITECTURES=arm64")
+
+        logger.info("CMake args: %s", str(cmake_cmd))
+        subprocess.check_call(cmake_cmd, cwd=build_dir)
+
+        if system() == "Windows":
+            subprocess.check_call(
+                ["cmake", "--build", ".", "--config", "Release"], cwd=build_dir
+            )
```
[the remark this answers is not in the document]
hcho3: The `build_tool` parameter is actually not needed, since the logic for choosing the build tool can be brought into `_build()` function. Let me update the function.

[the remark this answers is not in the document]
hcho3: As for Windows, a typical setup doesn't have Ninja or Make available as a simple system command. (I think Visual Studio has Ninja and Make somewhere in its installation, but you can't access them easily.)

python-package/packager/nativelib.py
```diff
@@ -0,0 +1,158 @@
...
+            nproc = os.cpu_count()
+            assert build_tool is not None
+            subprocess.check_call([build_tool, f"-j{nproc}"], cwd=build_dir)
+
+    if system() == "Windows":
+        for generator in (
+            "-GVisual Studio 17 2022",
+            "-GVisual Studio 16 2019",
+            "-GVisual Studio 15 2017",
+            "-GMinGW Makefiles",
```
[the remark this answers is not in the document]
hcho3: No, we want to hardcore the list because we have not tested other generators for the Windows platform.

python-package/packager/pep517.py
hcho3: Good point, let me do that.

python-package/packager/pep517.py
hcho3: I don't think we want to add more options any time soon.

CMakeLists.txt
```diff
@@ -268,8 +269,13 @@ if (JVM_BINDINGS)
...
 endif (JVM_BINDINGS)
 
-set_output_directory(runxgboost ${xgboost_SOURCE_DIR})
-set_output_directory(xgboost ${xgboost_SOURCE_DIR}/lib)
+if (KEEP_BUILD_ARTIFACTS_IN_BINARY_DIR)
+  set_output_directory(runxgboost ${xgboost_BINARY_DIR})
+  set_output_directory(xgboost ${xgboost_BINARY_DIR}/lib)
+else ()
+  set_output_directory(runxgboost ${xgboost_SOURCE_DIR})
+  set_output_directory(xgboost ${xgboost_SOURCE_DIR}/lib)
```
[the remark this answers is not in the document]
hcho3: `set_output_directory()` is a custom CMake function, which does change the output direcotyr of the final library:
https://github.com/dmlc/xgboost/blob/ef13dd31b1dc073f4732544739db5a252ae5b6e5/cmake/Utils.cmake#L56-L73

As for the reason for putting `libxgboost.so` under the source directory, this is due to legacy reasons. Other language bindings expect `libxgboost.so` to be under `lib/` subdirectory of the source directory.

python-package/packager/nativelib.py
```diff
@@ -0,0 +1,158 @@
...
+    else:
+        raise NotImplementedError(f"System {system()} not supported")
+    return name
+
+
+def build_libxgboost(
+    cpp_src_dir: pathlib.Path,
+    *,
+    build_dir: pathlib.Path,
+    build_config: BuildConfiguration,
```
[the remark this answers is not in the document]
hcho3: I personally like using keyword-only arguments, because I am being explicitly about which arguments correspond to which argument. Do you feel that this adds complexity?

python-package/packager/nativelib.py
```diff
@@ -0,0 +1,158 @@
...
+            except subprocess.CalledProcessError as e:
+                logger.info(
+                    "Tried building with generator %s but failed with exception %s",
+                    generator,
+                    str(e),
+                )
+                # Empty build directory
+                shutil.rmtree(build_dir)
+                build_dir.mkdir()
+                continue
```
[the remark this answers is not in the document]
hcho3: I will add a terminating `else` clause.

python-package/packager/pep517.py
```diff
@@ -0,0 +1,191 @@
...
+import hatchling.build
+
+from .build_config import BuildConfiguration
+from .nativelib import locate_local_libxgboost, locate_or_build_libxgboost
+from .sdist import copy_cpp_src_tree
+from .util import copy_with_logging, copytree_with_logging
+
+
+@contextmanager
+def cd(path: Union[str, pathlib.Path]) -> Iterator[str]:  # pylint: disable=C0103
```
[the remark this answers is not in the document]
hcho3: Nice, let me add a note about Python 3.11.

doc/install.rst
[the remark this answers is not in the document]
vyasr: Got it. Looks like it can also be done optionally as of 21.1, but no need to change the requirement for that reason.

python-package/packager/build_config.py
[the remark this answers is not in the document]
vyasr: Sounds good.

python-package/packager/build_config.py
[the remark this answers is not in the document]
vyasr: Ah you are right, I don't know why I was thinking this had anything to do with attributes being none rather than simply accessing them by name.

python-package/packager/nativelib.py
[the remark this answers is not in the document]
vyasr: I don't mind keyword-only arguments, but I find them odd when they come without defaults. More importantly, in this case I'm not sure why you would have one positional argument and then have two keyword-only arguments with no defaults. That's the main thing that makes for a confusing API IMO. I would make them all keyword-only if that's your preference.

CMakeLists.txt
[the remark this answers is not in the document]
vyasr: I see, OK if that is needed for legacy use cases that is fine.

python-package/packager/nativelib.py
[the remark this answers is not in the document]
vyasr: Yep I realized that it was being captured in the closure of the nested function but I was confused about why the build tool could be specified on Linux but not Windows. I know little about Windows though so that seems like a good reason to me.

python-package/packager/nativelib.py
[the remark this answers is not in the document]
vyasr: Got it.

python-package/packager/nativelib.py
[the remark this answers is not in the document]
vyasr: Got it.

python-package/packager/pep517.py
vyasr: OK that's fine then.

python-package/packager/pep517.py
vyasr: OK.

doc/contrib/python_packaging.rst
```diff
@@ -0,0 +1,83 @@
...
+the wheel into the current Python environment. Since the wheel already
+contains a pre-built copy of ``libxgboost.so``, it does not have to be
+built at the time of install. So ``pip install`` with the binary wheel
+completes quickly:
+
+.. code-block:: console
+  
+  $ pip install xgboost-2.0.0-py3-none-linux_x86_64.whl  # Completes quickly
+
+.. rubric:: Foonotes
```
vyasr, line 80:
```suggestion
.. rubric:: Footnotes
```

python-package/packager/pep517.py
```diff
@@ -0,0 +1,157 @@
...
+        raise NotImplementedError(
+            "XGBoost's custom build backend doesn't support config_settings option "
+            f"when building editable installation. {config_settings=}"
+        )
+
+    if locate_local_libxgboost(TOPLEVEL_DIR, logger=logger) is None:
+        raise AssertionError(
+            "To use the editable installation, first build libxgboost with CMake. "
+            "See https://xgboost.readthedocs.io/en/latest/build.html for detailed instructions."
+        )
```
vyasr, line 153: Out of curiosity, what happens if you remove this? Is it possible to use an editable installation that also builds libxgboost? In that situation running `pip install -e` would need to rebuild the library as well.

Not saying this feature should be implemented, just want to understand the current setup of the backend to figure out how best to work with this in the future for other RAPIDS libraries too.

doc/contrib/python_packaging.rst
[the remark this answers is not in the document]
vyasr: Understood.

vyasr (approved):
This is looking good! There are a couple of outstanding comments, but almost everything has been addressed and nothing is left that I would block merging on. Great to see this work here!

vyasr:
Note that I don't have permissions to resolve threads on this PR so I tried to leave comments or clearly indicate where I thought things were resolved.

python-package/packager/pep517.py
hcho3: > what happens if you remove this?

Editable installation won't build `libxgboost.so`, so `import xgboost` will error out.

> Is it possible to use an editable installation that also builds libxgboost?

I decided against it, as there is no reliable way for `packager` module to detect changes in the C++ code. If we want to re-build `libxgboost.so` upon changes made to C++, `scikit-build-core` will probably be a better choice.

I ultimately decided against `scikit-build-core` because:
* It doesn't yet support editable installation.
* I couldn't customize it enough to fit the current use cases of XGBoost.
* Much of what `scikit-learn-core` offers is not useful for XGBoost, since XGBoost does not use compiled extensions. XGBoost uses `ctypes` exclusively to access C API functions from `libxgboost.so`.

The calculus will be different for RAPIDS libraries which uses compiled extensions extensively. IMHO, if you use compiled extensions a lot, it's better to use `scikit-build-core` and give up on editable installation.

doc/contrib/python_packaging.rst
hcho3: Good catch!

python-package/packager/pep517.py
jameslamb: For what it's worth, `scikit-build-core` just added experimental support for editable installs: https://github.com/scikit-build/scikit-build-core/releases/tag/v0.3.0

Not sure exactly what "experimental" means, to be fair, and to what extent that's acceptable for a project as widely-used as XGBoost. But just wanted to note it.

I am planning to pursue `scikit-build-core` for LightGBM (https://github.com/microsoft/LightGBM/pull/5759), to not have to have any lingering Python glue score that's exposed to the changes in Python packaging. That is still very much a work in progress though, so I'm definitely not recommending that you switch to that over `hatchling`... just noting the link as something you may want to watch as we all figure out the `setup.py`-free world together 😊

python-package/packager/nativelib.py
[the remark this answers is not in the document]
hcho3: I removed all `*` from function signatures.

python-package/packager/pep517.py
hcho3: @jameslamb Guess my knowledge was outdated :) In the future, I'll definitely consider `scikit-learn-core` as a potential alternative.

vyasr: I will be migrating RAPIDS to use `scikit-build-core` at some point in the not-too-distant future. We're already using `scikit-build`, so the transition is the most natural way to get us onto a PEP 517 build. 

Support for editable installs is a must for RAPIDS, and is one of the reasons I haven't attempted the migration yet. [That feature request is something they're aware of](https://github.com/scikit-build/scikit-build-core/issues/114) and are working on. [My understanding is that the current experimental support](https://github.com/scikit-build/scikit-build-core/pull/212) only covers some possible cases around library changes but not all. I will definitely be keeping a close eye on this. I expect full support for editable installs is a precondition of a 1.0 release, and the project has been moving at a reasonable pace so it shouldn't be too long before that exists.

That said, at this stage `scikit-build-core` is definitely less mature than hatchling and I can't fault your choice since you seem to have been able to get everything working fairly painlessly :slightly_smiling_face:

jameslamb: @vyasr @hcho3 @trivialfis FYI I have a PR ready over in LightGBM for using `scikit-build-core`: https://github.com/microsoft/LightGBM/pull/5759. If you have time and interested, I'd appreciate any comments you have on it.

We don't care about editable installs for `lightgbm`, and I'm willing to go contribute to `scikit-build-core` (both code and discussions) as we discover issues in it, to help that tool mature.

vyasr: Thanks for the ping! I've started testing the migration for a couple RAPIDS libraries as well, see https://github.com/rapidsai/rmm/pull/1287 and https://github.com/rapidsai/cudf/pull/13531. Editable installs are already in significantly better shape than they were a couple of months ago FWIW.

jameslamb: Nice! I learned a lot about `scikit-build-core` in this process, `@` me any time on those RAPIDS PRs if you want an opinion or another set of eyes.

PR #9022, opened 2023-04-11
[spark] Make spark model have the same UID with its estimator

WeichenXu123:
Make spark model have the same UID with its estimator.

Because all other pyspark models obey the rule that uses the same UID with estimator, and the UID is used in param copying (e.g. CrossValidator copies param list to estimator, and UID is used for param owner validation), so we need to make the UID correct.

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -931,7 +931,11 @@ class _SparkXGBEstimator(Estimator, _SparkXGBParams, MLReadable, MLWritable):
         result_xgb_model = self._convert_to_sklearn_model(
             bytearray(booster, "utf-8"), config
         )
-        return self._copyValues(self._create_pyspark_model(result_xgb_model))
+        spark_model = self._create_pyspark_model(result_xgb_model)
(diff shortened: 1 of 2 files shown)


WeichenXu123:
Can we backport this fix to xgboost 1.7.6 version ?

trivialfis:
I don't have any plan for 1.7.6 at the moment, is this fix significant? In what scenario this can cause issue?

WeichenXu123:
> I don't have any plan for 1.7.6 at the moment, is this fix significant? In what scenario this can cause issues?

Only some very edge case (a pyspark pipeline contains xgboost estimator, and crossvalidator tunes the pipeline, and we set some params related to xgboost model prediction).

It is not an critical bug so 1.7.6 release is not urgent for . :)

WeichenXu123:
@trivialfis Could you trigger all CI runs and then merge the PR once CI passed? Thankyou!

WeichenXu123:
Reminder: Backport this commit to 1.7 branch :)

trivialfis (approved):

Issue #9044, opened 2023-04-17
[jvm-packages] Remove scala implementation of rabit tracker.


trivialfis:
It's marked as experimental and unmaintained. So far no one has raised interest in working on it. It's easier for us to keep the codebase unified.

If we were to have a new implementation of tracker, I think it should be written in a way that can be shared by multiple interfaces.

PR #9045, opened 2023-04-18
[breaking] [jvm-packages] Remove scala-implemented tracker.

trivialfis:
close https://github.com/dmlc/xgboost/issues/9044 .


trivialfis:
cc @wbo4958 .

dotbg:
@trivialfis  may I suggest removing the whole `ml.dmlc.xgboost4j.scala.rabit` package as it becomes dead code.

wbo4958:
I indeed see someone is using the scala tracker, so I don't know if it's appropriate to delete it.

dotbg:
@wbo4958 well,  it says "  * Note that this implementation is currently experimental, and should be used at your own risk." in documentation. I canupdate it instead of removing, if you prefer. However, I know nothing about it, so I'd have to ask for help. In case, this is preferred way of tackling this issue, please, rephrase the ticket and I can lend a hand

trivialfis:
I think we will remove it instead. It's not ideal for us to keep maintenance on two different implementations of the tracker. At some point, we should move the implementation into a shared core.

wbo4958:
LGTM if the CI/CD can pass

wbo4958 (approved):

hygt:
@trivialfis @wbo4958 FYI this change is causing hard to debug errors if your codebase was using the Scala tracker, as `TrackerConf(0L, "scala")` means that the Python tracker will try to bind to the `scala` address, which fails without a clear error message. It would be nice to mention that somewhere in the release note, or better yet, avoid using stringly typed configuration.

trivialfis:
The Python tracker is next to remove. We will rework the networking module

Issue #4628, opened 2019-07-02
[jvm-packages] spark hangs when training is run in quick succession


thesuperzapper:
I am getting an infinite hang when I run the following code a few times in quick succession:
```
import ml.dmlc.xgboost4j.scala.spark.XGBoostClassifier

val dataPath = "__SPARK_HOME_LOCATION__/data/mllib/sample_binary_classification_data.txt"
val data = spark.read.format("libsvm").option("vectorType", "dense").load(dataPath)
val xgbClassifier = new XGBoostClassifier()

xgbClassifier.fit(data).transform(data).show()
```

**Steps to reproduce:**
1. open spark-shell with xgboost jars
2. Run the above code
3. quickly rerun the last line until a hang happens

**Other information:**
When the hang happens I only get the tracker message, and nothing after that, I have to kill the spark job. (If I wait between runs, they always succeed. )
```
Tracker started, with env={DMLC_NUM_SERVER=0, DMLC_TRACKER_URI=XXX.XXX.XXX.XXX, DMLC_TRACKER_PORT=9096, DMLC_NUM_WORKER=2}
```

**My environment:**
* XGBoost Master
* Spark 2.4.3
* (Happens in both: __Zeppelin__ and __Spark-Shell__)

CodingCat:
Currently, parallel training for multiple models is not supported

thesuperzapper:
@CodingCat

In this case, I am letting each training finish, so it's only one model.

However, you're almost certainly right about the cause. That is, I think there is some "cool down" after training each model, in which you can't train another without hanging.

thesuperzapper:
@CodingCat 
After further investigation,  this crash is caused by running the same training within **exactly 60sec** of the last time you ran it.

**Steps to reproduce:**
1. Download Spark 2.4.3 [from here](https://www-us.apache.org/dist/spark/spark-2.4.3/spark-2.4.3-bin-hadoop2.7.tgz), and extract it.
2. Download the following jars from maven:
  * https://repo1.maven.org/maven2/ml/dmlc/xgboost4j/0.90/xgboost4j-0.90.jar
  * https://repo1.maven.org/maven2/ml/dmlc/xgboost4j-spark/0.90/xgboost4j-spark-0.90.jar
  * https://repo1.maven.org/maven2/com/typesafe/akka/akka-actor_2.11/2.3.11/akka-actor_2.11-2.3.11.jar
[...]

PR #9054, opened 2023-04-19
Stop using Rabit in prediction

austinzh:
As discuss in https://github.com/dmlc/xgboost/issues/4628
[PreXGBoost.scala:transformDataset](https://github.com/dmlc/xgboost/blob/a68580e2a7d84d7206165ab1f3c911886d45dcba/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/PreXGBoost.scala#L190) cause succession  training fail because it can not shutdown Rabit if iterator did not reach the end.

We shutdown Rabit only on the next() being call and hasNext() is false. As it show below.
```
        override def next(): Row = {
          val ret = batchIterImpl.next()
          if (!batchIterImpl.hasNext) {
            Rabit.shutdown()
          }
          ret
        }
```
But when we call show() or limit() instead of collect() or write() iterator won't reach hasNext() == false

And as for prediction, we don't need Rabit anyway, so there is no point to start it at all.
This PR is to remove it from prediction.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/PreXGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/PreXGBoost.scala
@@ -1,5 +1,5 @@
 /*
- Copyright (c) 2021-2022 by Contributors
+ Copyright (c) 2021-2023 by Contributors
 
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
@@ -22,7 +22,6 @@ import java.util.ServiceLoader
 import scala.collection.JavaConverters._
 import scala.collection.{AbstractIterator, Iterator, mutable}
 
-import ml.dmlc.xgboost4j.java.Communicator
 import ml.dmlc.xgboost4j.scala.{Booster, DMatrix}
 import ml.dmlc.xgboost4j.scala.spark.util.DataUtils.PackedParams
 import ml.dmlc.xgboost4j.scala.spark.params.XGBoostEstimatorCommon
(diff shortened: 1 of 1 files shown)


jvm-packages/dev/Dockerfile
```diff
@@ -21,10 +21,10 @@ FROM centos:7
 # Install all basic requirements
 RUN \
     yum -y update && \
-    yum install -y bzip2 make tar unzip wget xz git centos-release-scl yum-utils java-1.8.0-openjdk-devel && \
+    yum install -y which bzip2 make tar unzip wget xz git centos-release-scl yum-utils java-1.8.0-openjdk-devel && \
     yum-config-manager --enable centos-sclo-rh-testing && \
     yum -y update && \
-    yum install -y devtoolset-7-gcc devtoolset-7-binutils devtoolset-7-gcc-c++ && \
+    yum install -y devtoolset-7-gcc devtoolset-7-binutils devtoolset-7-gcc-c++ gcc-gfortran python3-devel openblas-devel lapack-devel pkgconfig  && \
     # Python
```
trivialfis, line 28: Not sure if we need to change the docker file here.

trivialfis (commented):
@wbo4958 Please help take a look.

jvm-packages/dev/Dockerfile
austinzh: I follow the README.md in jvm-package.
./xgboost/jvm-packages/dev/build-linux.sh will fail for 
1. there is no `which` command 
2. with openblas-devel `scipy` can not be install

wbo4958 (commented):
Please update the Copyright (Copyright (c) 2021-2023 by Contributors) and delete the unused import (import ml.dmlc.xgboost4j.java.Communicator)

wbo4958:
BTW, why are you changing the docker file? if the docker file didn't work on your side, I guess, it shouldn't work on the CI/CD side either, but seems the CI/CD is working as expected?

wbo4958:
Huh, And I checked GpuPreXGBoost transform which doesn't enable the rabit tracker. Seems I have forgotten to remove it for the CPU transform.

trivialfis:
I think the docker file is no longer used on CI. We will delete it instead and fix the readme

austinzh:
> I think the docker file is no longer used on CI. We will delete it instead and fix the readme

Sure, I will revert the Dockerfile then. Just keep things simple.

trivialfis:
Please ignore the failure on Windows for now.

wbo4958 (approved):
LGTM. Thx

PR #9060, opened 2023-04-20
More collective aggregators

rongou:
Added more collective functions to encapsulate the behavior difference between row-wise and column-wise split.

Still need to clean up how we deal with column split in other places.


rongou:
@trivialfis

trivialfis (approved):

PR #9081, opened 2023-04-24
More refactoring to take advantage of collective aggregators

rongou:
Also added `ShouldHaveLabels()` to encode when we expect to have labels.

### DIFF
--- a/src/collective/aggregator.h
+++ b/src/collective/aggregator.h
@@ -31,18 +31,16 @@ namespace collective {
  * @param buffer The buffer storing the results.
  * @param size The size of the buffer.
  * @param function The function used to calculate the results.
- * @param args Arguments to the function.
  */
-template <typename Function, typename T, typename... Args>
-void ApplyWithLabels(MetaInfo const& info, T* buffer, size_t size, Function&& function,
-                     Args&&... args) {
+template <typename Function>
+void ApplyWithLabels(MetaInfo const& info, void* buffer, size_t size, Function&& function) {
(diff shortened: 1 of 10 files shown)


src/collective/aggregator.h
```diff
@@ -31,18 +31,16 @@ namespace collective {
  * @param buffer The buffer storing the results.
  * @param size The size of the buffer.
  * @param function The function used to calculate the results.
- * @param args Arguments to the function.
  */
-template <typename Function, typename T, typename... Args>
-void ApplyWithLabels(MetaInfo const& info, T* buffer, size_t size, Function&& function,
-                     Args&&... args) {
+template <typename Function>
+void ApplyWithLabels(MetaInfo const& info, void* buffer, size_t size, Function&& function) {
```
trivialfis, line 36: Is type erasion necessary? (void*)

rongou: When doing the broadcast, we don't need to know the type, only the size (https://github.com/dmlc/xgboost/blob/master/src/collective/communicator-inl.h#L128), so in here we don't really care about the type either.

rongou:
@trivialfis @hcho3 Any ideas about the R test errors? Doesn't seem to be related to this PR.

hcho3:
@rongou Please ignore it for now. We are aware of the issue (https://github.com/dmlc/xgboost/issues/9091). Not sure when I'd get time to fix it.

trivialfis (approved):

PR #9102, opened 2023-04-26
Rely on gRPC to generate random port

rongou:
To avoid occasionally binding to a port still in use.


rongou:
@trivialfis @hcho3

trivialfis (commented):
Is it possible to just query port 0 on unix?

rongou:
No 0 just tells the server builder to pick a random unused port. The client needs the actual port number to connect.

trivialfis (approved):

PR #9103, opened 2023-04-26
Add demo for vertical federated learning

rongou:
Moved the current nvflare demo under `horizontal`.


rongou:
@trivialfis

trivialfis (approved):
LGTM. Out of curiosity,is there any difficulty for running it on the CI?

rongou:
For the horizontal one, it's probably not that hard since it only uses the small mushroom dataset. But for the vertical one we are downloading Higgs which is probably too big for CI.

@hcho3 should we try to run the demo (at least the horizontal one) on the CI? I think it would involve installing nvflare in the python environment.

trivialfis:
We can probably add a CPU test later. At the moment, I think it's quite difficult to setup the dependencies using only conda. (Correct me if I'm wrong). I will wait for conda to get rid of the autotools build for protobuf.

PR #9120, opened 2023-05-04
Support `hist` in the partition builder under column split

rongou:
First step in supporting column split for `hist`.


rongou:
@trivialfis

src/common/partition_builder.h
```diff
@@ -218,7 +232,27 @@ class PartitionBuilder {
         }
       }
     } else {
-      LOG(FATAL) << "Column data split is only supported for the `approx` tree method";
+      auto pred_hist = [&](auto ridx, auto bin_id) {
```
trivialfis, line 235: Logically, is there any difference from https://github.com/dmlc/xgboost/blob/07b2d5a26d96888fa4cec4dd1b1ef0b31d0c3fbb/src/common/partition_builder.h#L126 ?

rongou: Yes they are the same, but since it's a closure, can't be easily reused. Maybe we should extract into a standalone function?

trivialfis: The partitioner is becoming less and less manageable now. I don't have any good suggestions on how to refactor it at the moment and can approve the PR as-is. But would be great if there's a way to simplify/modularize it.

rongou: Would it be helpful to separate out the column-split logic into its own class?

trivialfis: Maybe next time. Currently, we have total 8 combinations for `tree_method, categorical, column-split`, along with these, we also need to take multi-output into account. I'm trying to think of a way to make the code easier to understand and write a test suite specifically for partition builder.

rongou:
@trivialfis looks like the CI failure is unrelated. Can this be merged now? Thanks!

trivialfis (approved):

trivialfis:
I have restarted all the tests.

Issue #9107, opened 2023-04-28
Allreduce failed and boundary errors since v1.7.0 with dask and rabit


efocht:
Since v1.7.0 I'm getting allreduce problems when training with dask. Using LocalCluster with 4 - 16 workers, 1 thread per worker. Example below is from 16 workers.

The dataset is huge and there are many features, I wasn't able to create a reproducer. Instead I instrumented the code in `rabit/src/allreduce_base.cc` as you can see in the patch https://github.com/efocht/xgboost/commit/161ed3cd7dc7c762edf7f46a49d1bde2c45c3836

The problems start after a while, I can see that allreduce works fine many times but suddenly things go wrong:
```
[11:21:29] TryAllreduceRing byte_size=8 count=553842
[11:21:29] TryAllreduceRing byte_size=8 count=502914
[11:21:29] TryAllreduceRing byte_size=8 count=502914
[11:21:29] TryAllreduceRing byte_size=8 count=502914
[11:21:29] TryAllreduceRing byte_size=8 count=496548
[11:21:29] TryAllreduceRing byte_size=8 count=509280
[11:21:29] TryAllreduceRing byte_size=8 count=496548
[11:21:29] TryAllreduceRing byte_size=8 count=502914
[11:21:29] TryAllreduceRing byte_size=8 count=509280
[11:21:29] TryAllreduceRing byte_size=8 count=509280
[11:21:29] TryAllreduceRing byte_size=8 count=515646
[11:21:29] TryAllreduceRing byte_size=8 count=509280
[11:21:29] TryAllreduceRing byte_size=8 count=502914
[11:21:29] TryAllreduceRing byte_size=8 count=509280
[11:21:29] TryAllreduceRing byte_size=8 count=496548
[11:21:29] TryReduceScatterRing ERROR end len = -1
[11:21:29] TryReduceScatterRing ERROR end len = -1
[11:21:29] TryReduceScatterRing ERROR end len = -1
[11:21:29] TryReduceScatterRing ERROR end len = -1
[11:21:29] TryReduceScatterRing ERROR end len = -1
[11:21:29] TryReduceScatterRing ERROR end len = -1
[11:21:29] TryReduceScatterRing ERROR end len = -1
[11:21:29] TryReduceScatterRing ERROR end len = -1
[11:21:29] TryReduceScatterRing ERROR end len = -1
[11:21:29] TryReduceScatterRing ERROR end len = -1
[...]

PR #9122, opened 2023-05-04
Fix monotone constraints on CPU.

trivialfis:
Fix https://github.com/dmlc/xgboost/issues/9107 .

- Fix use after free on CPU.
- Validate the size of the parameter.

### DIFF
--- a/src/tree/hist/evaluate_splits.h
+++ b/src/tree/hist/evaluate_splits.h
@@ -412,6 +412,7 @@ class HistEvaluator {
     tree_evaluator_.AddSplit(candidate.nid, left_child, right_child,
                              tree[candidate.nid].SplitIndex(), left_weight,
                              right_weight);
+    evaluator = tree_evaluator_.GetEvaluator();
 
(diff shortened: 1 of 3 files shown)


src/tree/hist/evaluate_splits.h
```diff
@@ -412,6 +412,7 @@ class HistEvaluator {
     tree_evaluator_.AddSplit(candidate.nid, left_child, right_child,
                              tree[candidate.nid].SplitIndex(), left_weight,
                              right_weight);
+    evaluator = tree_evaluator_.GetEvaluator();
```
hcho3, line 415: Does `AddSplit()` change the evaluator object in `tree_evaluator_`?

trivialfis: Yes, my oversight. It resize the storage according to number of nodes. The tests couldn't detect the error as it happens rarely due to pre-allocation and over-allocation for efficient GPU training.

hcho3 (approved):

PR #9144, opened 2023-05-08
Wait for data CUDA stream instead of sync.

trivialfis:


src/data/array_interface.cu
```diff
@@ -1,11 +1,15 @@
-/*!
- * Copyright 2021 by Contributors
+/**
+ * Copyright 2021-2023, XGBoost Contributors
  */
+#include <cinttypes>  // for int64_t
```
hcho3, line 4:
```suggestion
#include <cstdint>  // for int64_t
```

hcho3 (approved):

PR #9153, opened 2023-05-11
Fix `_defined` typo

StephanTLavavej:
Attempting to build XGBoost with MSVC fails catastrophically because there's a preprocessor line saying `_defined` which is malformed. This is the only occurrence in the whole codebase and it appears to be a simple typo.

After fixing this, compilation appears to succeed (with compiler warnings).


hcho3:
Thanks!

PR #9156, opened 2023-05-15
[pyspark] support typing for model.py

wbo4958:
This PR adds typing for model.py of xgboost.spark,  and I will create the followup PR to support typing of core.py of xgboost.spark


python-package/xgboost/spark/model.py
```diff
@@ -1,28 +1,73 @@
...
+        SparkXGBClassifier,
+        SparkXGBClassifierModel,
+        SparkXGBRanker,
+        SparkXGBRankerModel,
+        SparkXGBRegressor,
+        SparkXGBRegressorModel,
+    )
+
+    # xgboost types
+    XGB_ESTIMATOR = Union[XGBClassifier, XGBRanker, XGBRegressor]
```
trivialfis, line 39: Use `XGBModel` class instead. It's the parent class for all referenced classes here.

wbo4958:
Seems the failure `TestTreeMethod.test_categorical_ames_housing` is not caused by this PR? @trivialfis any thoughts on this?

python-package/xgboost/spark/model.py
```diff
@@ -1,28 +1,73 @@
...
+    ]
+    SPARK_XGB_MODEL_TYPE = Union[
+        Type[SparkXGBClassifierModel],
+        Type[SparkXGBRegressorModel],
+        Type[SparkXGBRankerModel],
+    ]
+
+    SPARK_XGB_INSTANCE = TypeVar(
+        "SPARK_XGB_INSTANCE", SPARK_XGB_ESTIMATOR, SPARK_XGB_MODEL
+    )
```
trivialfis, line 57: - Please consider the inheritance structure here instead of using Union.
- Please use `CamelCase` for types.

python-package/xgboost/spark/model.py
```diff
@@ -1,28 +1,73 @@
...
 from pyspark.ml.util import DefaultParamsReader, DefaultParamsWriter, MLReader, MLWriter
 from pyspark.sql import SparkSession
 
 from xgboost.core import Booster
 
 from .utils import get_class_name, get_logger
 
-
-def _get_or_create_tmp_dir():
+if TYPE_CHECKING:
```
trivialfis, line 27: Why type checking only? These are hard dependencies of the spark module, you can import them freely. The spark module is not part of package import: https://github.com/dmlc/xgboost/blob/7375bd058b48714d689d8404f7fc76b094412886/python-package/xgboost/__init__.py#L18 . As a result, what you import here won't affect users that don't import the spark module.

trivialfis:
> Seems the failure TestTreeMethod.test_categorical_ames_housing is not caused by this PR? @trivialfis any thoughts on this?

Please ignore it. It's the mighty internet error.

python-package/xgboost/spark/model.py
wbo4958: @trivialfis any thoughts on that?

python-package/xgboost/spark/model.py
wbo4958: @trivialfis any thoughts on that?

python-package/xgboost/spark/model.py
trivialfis: I think it's the same issue one would encounter for any typed language, the modules need to be written in a structured way. Python duck typing hides the issue, now that we want to annotate the code, some refactoring needs to be done.

wbo4958:
I will create a followup PR to support typing for core.py

trivialfis (approved):

PR #9199, opened 2023-05-25
Avoid thrust logical operation.

trivialfis:
Thrust implementation of `thrust::all_of/any_of/none_of` adopts an early stopping strategy to bailout early by dividing the input into small batches. This is not ideal for data validation as we expect all data to be valid. The strategy leads to excessive kernel launches and stream synchronization.

A small optimization from https://github.com/dmlc/xgboost/pull/9194 .

### DIFF
--- a/src/data/device_adapter.cuh
+++ b/src/data/device_adapter.cuh
@@ -237,8 +237,8 @@ bool NoInfInData(AdapterBatchT const& batch, IsValidFunctor is_valid) {
   // intervals to early stop. But we expect all data to be valid here, using small
   // intervals only decreases performance due to excessive kernel launch and stream
   // synchronization.
-  auto valid = thrust::reduce(thrust::cuda::par(alloc), value_iter, value_iter + batch.Size(), true,
-                              thrust::logical_and<>{});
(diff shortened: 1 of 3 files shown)


src/data/device_adapter.cuh
```diff
@@ -221,13 +221,24 @@ size_t GetRowCounts(const AdapterBatchT batch, common::Span<size_t> offset,
...
+    }
+    // check that there's no inf in data.
+    return !std::isinf(v);
+  });
+  dh::XGBCachingDeviceAllocator<char> alloc;
+  // The default implementation in thrust optimizes any_of/none_of/all_of by using small
+  // intervals to early stop. But we expect all data to be valid here, using small
+  // intervals only decreases performance due to excessive kernel launch and stream
+  // synchronization.
+  auto valid = dh::Reduce(thrust::cuda::par(alloc), value_iter, value_iter + batch.Size(), true,
```
rongou, line 240: Would it be simpler to use `thrust::transform_reduce`?

trivialfis: We have trouble with using thrust reduce directly as it doesn't support input larger than INT_MAX.

rongou: Never mind, looks like we need the index to iterate.

rongou (approved):

PR #9200, opened 2023-05-25
Fix running MGPU gtests

rongou:
Right now if I run all the gtests together (`./testxgboost`) on a multi-gpu machine I get an error. This is not caught by the CI since we run single-gpu and multi-gpu tests separately.

### DIFF
--- a/src/collective/communicator.cu
+++ b/src/collective/communicator.cu
@@ -12,19 +12,22 @@
 namespace xgboost {
 namespace collective {
 
-thread_local int Communicator::device_ordinal_{-1};
 thread_local std::unique_ptr<DeviceCommunicator> Communicator::device_communicator_{};
 
 void Communicator::Finalize() {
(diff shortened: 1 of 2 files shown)


rongou:
@trivialfis @hcho3

trivialfis:
Curiously, I have been running the gtest suite over and over again in recent days.

rongou:
On a multi-gpu machine? It shouldn't have worked since we short circuit nccl initialization with a single gpu: https://github.com/dmlc/xgboost/blob/master/src/collective/nccl_device_communicator.cuh#L27. So when we switch to an MGPU test, GPU 0 would be uninitialized.

trivialfis:
Yes, 2-GPU machine.

trivialfis:
NVM. I was using an old commit.

trivialfis (approved):

PR #9210, opened 2023-05-27
Remove unused variables.

trivialfis:


trivialfis:
cc @rongou

src/collective/communicator-inl.h
```diff
@@ -252,7 +252,7 @@ inline AllgatherVResult<T> AllgatherV(std::vector<T> const &inputs,
 
   // Calculate input offsets (std::exclusive_scan).
   std::vector<std::size_t> offsets(all_sizes.size());
-  for (auto i = 1; i < offsets.size(); i++) {
+  for (std::size_t i = 1; i < offsets.size(); i++) {
```
rongou, line 255: What's the reason for not using `auto` here?

trivialfis: It's `int` by default, which generates a signed comparison warning when compared against `std::size_t` returned by `offsets.size()`.

rongou: Would it be simpler to append `u`?
```
for (auto i = 1u; i < offsets.size(); i++) {
```

trivialfis: I'm not sure which one is simpler. Matching types seems simple. ;-)

rongou (approved):

PR #9212, opened 2023-05-29
Skip optional synchronization in thrust.

trivialfis:
- Use `par_nosync` from thrust 2.0 (bundled in CTK 12.0).


trivialfis:
We still need to find a way to skip synchronization in the device vector when the memory is cached. I don't know how to change the allocator to return an appropriate system yet.

trivialfis:
@rongou Could you please take a look when you are available?

rongou (approved):

RAMitchell (approved):

PR #9223, opened 2023-05-31
TreeRefresher note word spelling modification

gklab:


trivialfis (approved):
Thank you for the fix!

PR #9234, opened 2023-06-01
[CI] Automatically bump Rapids version in containers

hcho3:
TODO:
- [x] set up a workflow to submit an automated pull request, using https://github.com/marketplace/actions/create-pull-request


trivialfis (commented):
Does this affect Python tests? I have been using a conda environment file for development with RMM, maybe it's useful to upstream it here.

hcho3:
> Does this affect Python tests?

Yes.

> I have been using a conda environment file for development with RMM

Currently, we don't use environment files because environment files don't support a parameter. Since we are adopting an automated solution to keep RAPIDS up to date, we can work around this restriction:

* Move all Conda envs into environment files. Make sure to pin Rapids packages to specific versions: `rapids=23.04`, `rmm=23.04` and so forth.
* The automated script should periodically check for the latest RAPIDS release and update the Conda env files appropriately.

What do you think of this idea?

trivialfis:
For some reason, no. It keeps downloading individual components like nsight-system and cudart as 12.1.

hcho3:
Got it. Then let me merge this PR as it is.

PR #9246, opened 2023-06-05
Remove document and demo in RABIT.

trivialfis:
They are quite outdated. One can find the document in the original repository.


hcho3 (approved):
I love it when we remove tons of unused code :)

PR #9253, opened 2023-06-05
Rearranged maven profiles so that scala-2.13 artifacts are published without gpu-related libraries

dotbg:
I've noticed that the 2.13 is missing in the S3 snapshot folder.

### DIFF
--- a/jvm-packages/pom.xml
+++ b/jvm-packages/pom.xml
@@ -301,14 +301,6 @@
                     <url>https://s3.amazonaws.com/xgboost-maven-repo/release</url>
                 </repository>
             </repositories>
-            <modules>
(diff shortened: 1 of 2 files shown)


jvm-packages/pom.xml
```diff
@@ -301,14 +301,6 @@
                     <url>https://s3.amazonaws.com/xgboost-maven-repo/release</url>
                 </repository>
             </repositories>
-            <modules>
```
dotbg, line 304: Please, note, that this profile is only responsible for the repositories now. 
Release has to be done specifying profiles `default` and `gpu` **in** **addition** to this one

trivialfis:
cc @hcho3 @wbo4958

PR #9256, opened 2023-06-07
cpu_predictor has_categorical error

gklab:


trivialfis (approved):
Thank you for the fix!

gklab:
The code can be simplified here.

gklab:
@trivialfis

trivialfis:
Hi, can we revert the simplification and keep the fix?

gklab:
yes

gklab:
> Hi, can we revert the simplification and keep the fix?

OK. May I ask what is the original intention of not simplifying this code?

gklab:
> Hi, can we revert the simplification and keep the fix?

Interesting, I had no idea that my simplification of this code would cause so many validations to fail.

gklab:
> Hi, can we revert the simplification and keep the fix?

Oh, I'm a rookie in C++, I don't know that template parameters should use constants

trivialfis (approved):
Nice fix, thank you!

PR #9270, opened 2023-06-07
More tests for cpu predictor with column split

rongou:
Fixed `PredictLeaf`. Added checks for methods not yet supported.


rongou:
@trivialfis

PR #9281, opened 2023-06-09
updater_quantile_hist.cc annotation adjustment

gklab:


trivialfis:
Many of the annotations are generated by include-what-you-use. It turns out it's challenging to keep up-to-date as a naive application of iwyu doesn't always give the correct output and requires manual adjustment. It would be great if we could adopt a complete solution based on iwyu with header mapping.

gklab:
> 

Well, automatic plus manual adjustment is really not a long-term solution.
I try to research it.

gklab:
> Many of the annotations are generated by include-what-you-use. It turns out it's challenging to keep up-to-date as a naive application of iwyu doesn't always give the correct output and requires manual adjustment. It would be great if we could adopt a complete solution based on iwyu with header mapping.

Well, automatic plus manual adjustment is really not a long-term solution.
I try to research it.

trivialfis:
Closing now as the changes are out-dated.

PR #9283, opened 2023-06-09
expand_entry remove redundant methods

gklab:
This function does not seem to be used in the code.


trivialfis (approved):
Thank you for the cleanup!

Issue #9170, opened 2023-05-17
[pyspark] unify transform for classification and core framework.


wbo4958:
I will file a PR for this to unify the transform to reuse some code.

CarlaFernandez:
Hello @wbo4958 , perhaps you could take a look at my recent issue #9175 since you'll be working on the `transform`?

wbo4958:
@CarlaFernandez, ok, I will check it.

PR #9292, opened 2023-06-12
[pyspark] rework transform to reuse same code

wbo4958:
To fix https://github.com/dmlc/xgboost/issues/9170

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -1129,15 +1129,15 @@ class _SparkXGBModel(Model, _SparkXGBParams, MLReadable, MLWritable):
 
         return pred_contrib_col_name
 
-    def _out_schema(self) -> str:
-        """The returned type of the user-defined function. The value must
+    def _out_schema(self) -> Tuple[bool, str]:
+        """Return the bool to indicate if it's a single prediction, true is single prediction,
+        and the returned type of the user-defined function. The value must
         be a DDL-formatted type string."""
 
-        return (
-            f"{pred.prediction} double, {pred.pred_contrib} array<double>"
-            if self._get_pred_contrib_col_name() is not None
-            else "double"
-        )
+        if self._get_pred_contrib_col_name() is not None:
+            return False, f"{pred.prediction} double, {pred.pred_contrib} array<double>"
+
+        return True, "double"
 
     def _get_predict_func(self) -> Callable:
         """Return the true prediction function which will be running on the executor side"""
@@ -1146,7 +1146,7 @@ class _SparkXGBModel(Model, _SparkXGBParams, MLReadable, MLWritable):
         pred_contrib_col_name = self._get_pred_contrib_col_name()
 
         def _predict(
-            model: XGBModel, X: ArrayLike, base_margin: Optional[np.ndarray]
+            model: XGBModel, X: ArrayLike, base_margin: Optional[ArrayLike]
         ) -> Union[pd.DataFrame, pd.Series]:
             data = {}
             preds = model.predict(
@@ -1169,7 +1169,7 @@ class _SparkXGBModel(Model, _SparkXGBParams, MLReadable, MLWritable):
     def _post_transform(self, dataset: DataFrame, pred_col: Column) -> DataFrame:
         """Post process of transform"""
         prediction_col_name = self.getOrDefault(self.predictionCol)
-        single_pred = "," not in self._out_schema()
+        single_pred, _ = self._out_schema()
 
         if single_pred:
             if prediction_col_name:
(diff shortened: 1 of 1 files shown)


wbo4958:
@WeichenXu123 @trivialfis could you please help to review this PR? Previously, SparkXGBClassifier totally overrides the whole _transform function which causes duplicated common code, so this PR tries to unify them.

python-package/xgboost/spark/core.py
```diff
@@ -1092,12 +1093,87 @@ def _get_feature_col(
             )
         return features_col, feature_col_names
 
+    def _get_pred_contrib_col_name(self) -> str:
+        """Return the pred_contrib_col col name"""
+        pred_contrib_col_name = ""
```
trivialfis, line 1098: Let's make this return `Optional[str]` instead of `str.  A `None` is easier to recognize than an empty string.

python-package/xgboost/spark/core.py
```diff
@@ -1092,12 +1093,87 @@ def _get_feature_col(
...
+                base_margin=base_margin,
+                validate_features=False,
+                **predict_params,
+            )
+            data[pred.prediction] = pd.Series(preds)
+
+            result = data[pred.prediction]
+            if pred_contrib_col_name:
+                contribs = pred_contribs(model, X, base_margin)
+                data[pred.pred_contrib] = pd.Series(list(contribs))
```
trivialfis, line 1138: Not sure how it works. Wouldn't this be super slow?

python-package/xgboost/spark/core.py
```diff
@@ -1092,12 +1093,87 @@ def _get_feature_col(
...
+                result = pd.DataFrame(data=data)
```
trivialfis, line 1139: This violates the type for the result. You can return `pd.DataFrame` here.

python-package/xgboost/spark/core.py
```diff
@@ -1092,12 +1093,87 @@ def _get_feature_col(
...
+
+            return result
+
+        return _predict
+
+    def _post_transform(self, dataset: DataFrame, pred_col: Column) -> DataFrame:
+        """Post process of transform"""
+        prediction_col_name = self.getOrDefault(self.predictionCol)
+        single_pred = "," not in self._out_schema()
```
trivialfis, line 1148: That's a bit, hmm, unconventional. Is there a way to refactor this to make it more "standard"?

python-package/xgboost/spark/core.py
```diff
@@ -1112,20 +1188,9 @@ def _transform(self, dataset: DataFrame) -> DataFrame:
         features_col, feature_col_names = self._get_feature_col(dataset)
         enable_sparse_data_optim = self.getOrDefault(self.enable_sparse_data_optim)
```
trivialfis, line 1189: Is `enable_sparse_data_optim` used?

trivialfis (commented):
Does the refactor make the code cleaner or easier to understand? I find it quite hacky, but might be a general issue with PySpark-based libraries.

python-package/xgboost/spark/core.py
wbo4958: The _out_schema has typing hint with str returned which is a DDL formatted string, so if there're many columns, it must have at least a ",". So it's conventional when we check if it is a single column according to if  "," is in the schema.

python-package/xgboost/spark/core.py
wbo4958: Yeah, It will be used in the common framework, pls refer to https://github.com/dmlc/xgboost/pull/9292/files#diff-8900fd245992de81e0a7b8cb7fcd18c37bae4473e8672e9b11fd7010cac358cfR1198

python-package/xgboost/spark/core.py
wbo4958: Good suggestion. Done

python-package/xgboost/spark/core.py
wbo4958: any thoughts to rework this?  using contribs.tolist() ?

python-package/xgboost/spark/core.py
wbo4958: Good suggestion. Done.

python-package/xgboost/spark/core.py
trivialfis: We are meta-programming by manipulating strings here.

python-package/xgboost/spark/core.py
```diff
@@ -1092,12 +1093,86 @@ def _get_feature_col(
...
+        )
+
+    def _get_predict_func(self) -> Callable:
+        """Return the true prediction function which will be running on the executor side"""
+
+        predict_params = self._gen_predict_params_dict()
+        pred_contrib_col_name = self._get_pred_contrib_col_name()
+
+        def _predict(
+            model: XGBModel, X: ArrayLike, base_margin: Optional[np.ndarray]
```
trivialfis, line 1124: `base_margin` is not necessarily `np.ndarray`. Let's stick with `ArrayLike`.

python-package/xgboost/spark/core.py
trivialfis: Maybe just use the `pred_contrib_col_name` as a predicate?

wbo4958: fixed this issue.

python-package/xgboost/spark/core.py
wbo4958: Done

trivialfis (approved):

PR #9299, opened 2023-06-14
[doc] How to build xgboost4j on macOS

jaceklaskowski:
A new section how to build xgboost4j locally on macOS (that turns out to be similar to Linux, but it was not obvious initially when I first looked at README.md and hence the change).

It also turned out I don't need Docker to build locally. I'm not against Docker but thought it's going to be more "welcoming" for devs who are for less deps and merely want to use the good ol' `mvn` 😉

### DIFF
--- a/jvm-packages/README.md
+++ b/jvm-packages/README.md
@@ -146,14 +146,16 @@ Spark assumes that the dataset is 1-based indexed. However, when you do predicti
 
 You can build/package xgboost4j locally with the following steps:
 
-**Linux:**
+### Linux
+
 1. Ensure [Docker for Linux](https://docs.docker.com/install/) is installed.
 2. Clone this repo: `git clone --recursive https://github.com/dmlc/xgboost.git`
 3. Run the following command:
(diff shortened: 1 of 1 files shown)


trivialfis:
Hi, thank you for the PR. There's already a section in https://github.com/dmlc/xgboost/blob/master/doc/build.rst#id27 . Would be great if you can opt to improve it instead of starting a new doc.

jaceklaskowski:
I'm OK with the decision but wonder what's going on with `jvm-packages/README.md` then that mentions Linux and Windows only and makes an impression that macOS is left aside (something I experience when I first saw the repo). `jvm-packages/README.md` is the very first doc many devs would read first before `doc/build.rst` (if this doc page is going to be read ever).

trivialfis:
It's a remanent of old documents. We expect users to visit https://xgboost.readthedocs.io/en/stable/ instead of digging into readme files. I will followup with a cleanup to these files.

trivialfis:
The old document is now removed.

Issue #9301, opened 2023-06-15
[doc] Variable 'label' is not defined in the pyspark application example


georgeothon:
Hi, I found an undefined variable in the pyspark estimator application example. The variable 'label' was used instead of the variable 'label_name' to generate the feture list.

![xgboost_doc](https://github.com/dmlc/xgboost/assets/60243072/e65f8703-b853-4273-879f-0f83851d807d)

PR #9302, opened 2023-06-15
[doc] Variable 'label' is not defined in the pyspark application example #9301

georgeothon:
Hi, I found an undefined variable in the pyspark estimator application example. The variable 'label' was used instead of the variable 'label_name' to generate the feture list.
xgboost_doc


trivialfis (approved):
Thank you for the fix!

PR #9312, opened 2023-06-16
Use dh::CUDAStream in NCCL communicator

rongou:
@trivialfis

### DIFF
--- a/src/collective/nccl_device_communicator.cu
+++ b/src/collective/nccl_device_communicator.cu
@@ -1,0 +1,227 @@
+/*!
+ * Copyright 2023 XGBoost contributors
+ */
+
+#include "nccl_device_communicator.cuh"
+
+namespace xgboost {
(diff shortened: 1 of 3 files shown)


trivialfis:
No need to rush into it. We merged the NCCL changes. ;-)

I think it will take some time for XGBoost use replace the default CUDA stream and use async streams everywhere.

rongou:
You think the test failures are legit?

trivialfis:
Not sure if it's the CUDA driver warning or the nccl change. The NCCL change seems more likely considering the driver warning was present in previous builds as well

trivialfis:
It's fine to not use that stream for now. I think we will pass the booster context into allreduce in the future. The context contains a CUDA stream (legacy stream for now)

rongou:
Yeah probably need some effort to get everything to use the async stream. Closing this for now.

PR #9315, opened 2023-06-19
Use ptr from mmap for GHistIndexMatrix and ColumnMatrix. 

trivialfis:
The basic idea is to define a reference counted view on opaque data. Either host memory or mmap can back the data. One problem with this approach is that the underlying resource might be read-only, but we cannot reflect this on the view. We can use `std::variant` to hold a const and a non-const version of the vectors, but the code can get messy with many accessors. Currently, the pages returned by `GetBatches` are const references, providing some safety.

- Define a resource for holding various types of memory pointers.
- Define ref vector for holding resources.
- Align all IO operations with data pages to 8 bytes.
- Swap the underlying resources for GHistM and ColumnM.
- Add documentation for current status.
- s390x support is removed. It should work if you can compile XGBoost, all the old workaround code does is to get GCC to compile.

Stacked on top of https://github.com/dmlc/xgboost/pull/9282.

### DIFF
--- a/doc/tutorials/external_memory.rst
+++ b/doc/tutorials/external_memory.rst
@@ -28,7 +28,8 @@ development. Like the :py:class:`~xgboost.QuantileDMatrix` with
 supplied by the user. However, unlike the :py:class:`~xgboost.QuantileDMatrix`, external
 memory will not concatenate the batches unless GPU is used (it uses a hybrid approach,
 more details follow). Instead, it will cache all batches on the external memory and fetch
-them on-demand.
+them on-demand.  Go to the end of the document to see a comparison between
+`QuantileDMatrix` and external memory.
 
 *************
 Data Iterator
@@ -208,7 +209,9 @@ construction of `QuantileDmatrix` with data chunks. On the other hand, if it's p
 :py:class:`~xgboost.DMatrix`, it instead enables the external memory feature. The
 :py:class:`~xgboost.QuantileDmatrix` concatenates the data on memory after compression and
 doesn't fetch data during training. On the other hand, the external memory `DMatrix`
-fetches data batches from external memory on-demand.
+fetches data batches from external memory on-demand.  Use the `QuantileDMatrix` (with
+iterator if necessary) when you can fit most of your data in memory. The training would be
+an order of magnitute faster than using external memory.
 
 ****************
 Text File Inputs
--- a/doc/tutorials/external_memory.rst
+++ b/doc/tutorials/external_memory.rst
@@ -22,6 +22,14 @@ GPU-based training algorithm. We will introduce them in the following sections.
 
    The feature is still experimental as of 2.0. The performance is not well optimized.
 
+The external memory support has gone through multiple iterations and is still under heavy
+development. Like the :py:class:`~xgboost.QuantileDMatrix` with
+:py:class:`~xgboost.DataIter`, XGBoost loads data batch-by-batch using a custom iterator
+supplied by the user. However, unlike the :py:class:`~xgboost.QuantileDMatrix`, external
+memory will not concatenate the batches unless GPU is used (it uses a hybrid approach,
+more details follow). Instead, it will cache all batches on the external memory and fetch
+them on-demand.
+
 *************
 Data Iterator
 *************
@@ -113,10 +121,11 @@ External memory is supported by GPU algorithms (i.e. when ``tree_method`` is set
 ``gpu_hist``). However, the algorithm used for GPU is different from the one used for
 CPU. When training on a CPU, the tree method iterates through all batches from external
 memory for each step of the tree construction algorithm. On the other hand, the GPU
-algorithm concatenates all batches into one and stores it in GPU memory. To reduce overall
-memory usage, users can utilize subsampling. The good news is that the GPU hist tree
-method supports gradient-based sampling, enabling users to set a low sampling rate without
-compromising accuracy.
+algorithm uses a hybrid approach. It iterates through the data during the beginning of
+each iteration and concatenates all batches into one in GPU memory. To reduce overall
+memory usage, users can utilize subsampling. The GPU hist tree method supports
+`gradient-based sampling`, enabling users to set a low sampling rate without compromising
+accuracy.
 
 .. code-block:: python
 
@@ -134,6 +143,8 @@ see `this paper <https://arxiv.org/abs/2005.09148>`_.
    When GPU is running out of memory during iteration on external memory, user might
    recieve a segfault instead of an OOM exception.
 
+.. _ext_remarks:
+
 *******
 Remarks
 *******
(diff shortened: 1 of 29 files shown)


src/data/sparse_page_source.h
```diff
@@ -79,30 +94,77 @@ class TryLockGuard {
...
 
   using Ring = std::vector<std::future<std::shared_ptr<S>>>;
   // A ring storing futures to data.  Since the DMatrix iterator is forward only, so we
   // can pre-fetch data in a ring.
   std::unique_ptr<Ring> ring_{new Ring};
-  dmlc::OMPException exec_;
+  // Catching exception in pre-fetch threads to prevent segfault. Not always work though,
+  // OOM error can be delayed due to lazy commit. On the bright side, if mmap is used then
+  // OOM error should be rare.
+  ExceHandler exec_;
```
trivialfis, line 167: typo.

doc/tutorials/external_memory.rst
```diff
@@ -142,17 +153,62 @@ When using external memory with XBGoost, data is divided into smaller chunks so
...
+*******************************
+Compared to the QuantileDMatrix
+*******************************
+
+Passing an iterator to the :py:class:`~xgboost.QuantileDmatrix` enables direct
+construction of `QuantileDmatrix` with data chunks. On the other hand, if it's passed to
+:py:class:`~xgboost.DMatrix`, it instead enables the external memory feature. The
+:py:class:`~xgboost.QuantileDmatrix` concatenates the data on memory after compression and
+doesn't fetch data during training. On the other hand, the external memory `DMatrix`
+fetches data batches from external memory on-demand.
```
trivialfis, line 211:
```suggestion
fetches data batches from external memory on-demand. Use the `QuantileDMatrix` (with iterator if necessary) when you can fit most of your data in memory. The training would be an order of magnitute faster than using external memory.
```

doc/tutorials/external_memory.rst
```diff
@@ -22,6 +22,14 @@ GPU-based training algorithm. We will introduce them in the following sections.
 
    The feature is still experimental as of 2.0. The performance is not well optimized.
 
+The external memory support has gone through multiple iterations and is still under heavy
+development. Like the :py:class:`~xgboost.QuantileDMatrix` with
+:py:class:`~xgboost.DataIter`, XGBoost loads data batch-by-batch using a custom iterator
+supplied by the user. However, unlike the :py:class:`~xgboost.QuantileDMatrix`, external
+memory will not concatenate the batches unless GPU is used (it uses a hybrid approach,
+more details follow). Instead, it will cache all batches on the external memory and fetch
+them on-demand.
```
trivialfis, line 31:
```suggestion
them on-demand. Go to the end of the document to see a comparison between `QuantileDMatrix` and external memory.
```

src/common/column_matrix.h
```diff
@@ -131,22 +142,65 @@ class DenseColumnIter : public Column<BinIdxT> {
...
 
   template <typename ColumnBinT, typename BinT, typename RIdx>
   void SetBinSparse(BinT bin_id, RIdx rid, bst_feature_t fid, ColumnBinT* local_index) {
     if (type_[fid] == kDenseColumn) {
       ColumnBinT* begin = &local_index[feature_offsets_[fid]];
       begin[rid] = bin_id - index_base_[fid];
-      // not thread-safe with bool vector.  FIXME(jiamingy): We can directly assign
-      // kMissingId to the index to avoid missing flags.
-      missing_flags_[feature_offsets_[fid] + rid] = false;
+      // not thread-safe with bool vector.
```
trivialfis, line 200:
```suggestion
      // not thread-safe with bitfield.
```

src/common/io.h
```diff
@@ -129,18 +135,209 @@ inline std::string ReadAll(std::string const &path) {
...
+  ~MmapResource() noexcept(false) override;
+
+  [[nodiscard]] void* Data() override;
+  [[nodiscard]] std::size_t Size() const override;
+};
+
+/**
+ * @param Alignment for resource read stream and aligned write stream.
+ */
+constexpr std::size_t IOAlignment() {
```
trivialfis, line 219: We might want to make this a compile-time configurable parameter in the future. Currently, 8-byte alignment is sufficient for all types used by the external memory.

src/common/ref_resource_view.h
```diff
@@ -0,0 +1,158 @@
...
+   * @param ptr  The pointer to view.
+   * @param n    The length of the view.
+   * @param mem  The owner of the pointer.
+   * @param init Initialize the view with this value.
+   */
+  RefResourceView(value_type* ptr, size_type n, std::shared_ptr<common::ResourceHandler> mem,
+                  T const& init)
+      : RefResourceView{ptr, n, mem} {
+    if (n != 0) {
+      std::fill_n(ptr_, n, init);
```
trivialfis, line 55: Need some more thoughts for GPU integration.

src/data/ellpack_page_raw_format.cu
```diff
@@ -1,51 +1,55 @@
...
+  bool Read(EllpackPage* page, common::ResourceReadStream* fi) override {
     auto* impl = page->Impl();
     if (!ReadHistogramCuts(&impl->Cuts(), fi)) {
       return false;
     }
-    fi->Read(&impl->n_rows);
-    fi->Read(&impl->is_dense);
-    fi->Read(&impl->row_stride);
-    fi->Read(&impl->gidx_buffer.HostVector());
+    if (!fi->Read(&impl->n_rows)) {
```
trivialfis, line 25: The read method is non-throwing based on the convention set by dmlc core. Therefore we always check the return type and use it to indicate failure. This might not be a good fit for XGBoost.

src/data/gradient_index_format.cc
```diff
@@ -1,38 +1,49 @@
...
+#include "sparse_page_writer.h"           // for SparsePageFormat
+
+namespace xgboost::data {
 class GHistIndexRawFormat : public SparsePageFormat<GHistIndexMatrix> {
  public:
-  bool Read(GHistIndexMatrix* page, dmlc::SeekStream* fi) override {
+  bool Read(GHistIndexMatrix* page, common::ResourceReadStream* fi) override {
+    CHECK(fi);
+
     if (!ReadHistogramCuts(&page->cut, fi)) {
```
trivialfis, line 21: This doesn't use `mmap` as the cuts are stored in the host device vector. We need to change it in the future. We might drop the requirement of storing the cuts.

src/common/io.h
```diff
@@ -129,18 +135,209 @@ inline std::string ReadAll(std::string const &path) {
...
+/**
+ * @brief Wrap resource into a dmlc stream.
+ *
+ *  This class is to facilitate the use of mmap. Caller can optionally use the `Read()`
+ *  method or the `Consume()` method. The former copies data into output, while the latter
+ *  makes copy only if it's a primitive type.
+ *
+ *  Input is required to be aligned to IOAlignment().
+ */
+class ResourceReadStream : public dmlc::SeekStream {
```
trivialfis, line 236: Rename to emphasize `Aligned`.

src/common/io.h
```diff
@@ -129,18 +135,209 @@ inline std::string ReadAll(std::string const &path) {
...
+
+/**
+ * @param Alignment for resource read stream and aligned write stream.
+ */
+constexpr std::size_t IOAlignment() {
+  // For most of the pod types in XGBoost, 8 byte is sufficient.
+  return 8;
+}
+
+// FIXME(Jiamingy): Remove the dependency on `dmlc::SeekStream` once all page types
```
trivialfis, line 224: Not too difficult to do, we can implement it in this PR if necessary.

src/common/io.h
```diff
@@ -129,18 +136,244 @@ inline std::string ReadAll(std::string const &path) {
   return content;
 }
 
+struct MMAPFile;
+
+/**
+ * @brief Handler for one-shot resource. Unlike `std::pmr::*`, the resource handler is
+ *        fixed once it's constructed. User cannot use mutable operations like resize.
```
trivialfis, line 143:
```suggestion
 *        fixed once it's constructed. Users cannot use mutable operations like resize
 *        without acquiring the specific resource first.
```

RAMitchell (approved):

PR #9317, opened 2023-06-19
Cleanup to prepare for using mmap pointer in external memory.

trivialfis:
Extracted from https://github.com/dmlc/xgboost/pull/9315 .

- Update SparseDMatrix comment.
- Use a pointer in the bitfield. We will replace the `std::vector<bool>` in `ColumnMatrix` with bitfield.
- Clean up the page source. The timer is removed as it's not accurate once we swap the mmap pointer into the page.

### DIFF
--- a/src/common/bitfield.h
+++ b/src/common/bitfield.h
@@ -1,5 +1,5 @@
-/*!
- * Copyright 2019 by Contributors
+/**
+ * Copyright 2019-2023, XGBoost Contributors
  * \file bitfield.h
  */
 #ifndef XGBOOST_COMMON_BITFIELD_H_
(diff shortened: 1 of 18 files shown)


src/common/bitfield.h
```diff
@@ -86,13 +90,15 @@ struct BitFieldContainer {
 
  public:
   BitFieldContainer() = default;
-  XGBOOST_DEVICE explicit BitFieldContainer(common::Span<value_type> bits) : bits_{bits} {}
-  XGBOOST_DEVICE BitFieldContainer(BitFieldContainer const& other) : bits_{other.bits_} {}
```
hcho3, line 90: Is it your intention that we shouldn't allow copy-construction of `BitFieldContainer` on device?

src/common/bitfield.h
```diff
@@ -70,8 +73,9 @@ struct BitFieldContainer {
   };
 
  private:
-  common::Span<value_type> bits_;
-  static_assert(!std::is_signed<VT>::value, "Must use unsiged type as underlying storage.");
+  value_type* bits_{nullptr};
```
hcho3, line 76: Can you explain why we should use a raw pointer here? You mentioned the use of `std::vector<bool>` in `ColumnMatrix`. Does the span behave badly when `value_type == bool` ?

trivialfis: It's used in hot loops, the check used in span can be expensive.

src/common/bitfield.h
trivialfis: When `default` is specified, nvcc will treat it as host-device function automatically.

hcho3 (approved):

PR #9319, opened 2023-06-21
Adding type hinting tests

michael-gendy-mention-me:
As part of [this issue](https://github.com/dmlc/xgboost/issues/6496#issuecomment-1558022930), we want to add mypy support to our tests. This PR will be type hinting a range of tests for the Python package

### DIFF
--- a/tests/python/test_basic.py
+++ b/tests/python/test_basic.py
@@ -2,11 +2,18 @@ import json
 import os
 import tempfile
 from pathlib import Path
+from typing import (
+    Any,
+    Dict,
+    List,
+    Tuple,
+)
 
 import numpy as np
 import pytest
 
 import xgboost as xgb
+from xgboost.core import Booster
 from xgboost import testing as tm
 
 dpath = 'demo/data/'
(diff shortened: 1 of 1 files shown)


trivialfis (commented):
Thank you for the work on type hint @michael-gendy-mention-me !

> I'm happy to continue for remaining tests, but I'm not sure how much value adding type hinting for the tests is generating. Do you see this as important and should I therefore continue with more files?

Indeed, I believe it's essential. By having type hints in tests, we can check the type-hint of the xgboost library from a user's perspective thoroughly.

Based on this, my priority for type hint implementation would be:
- Core module.
- Scripts in `demo/guide-python`.
- Other scripts like dask and survival training in `demo`.
- Tests.

I put demo over tests just to add value as soon as possible. The type hint might make some of the demonstration scripts easier to read. Others can freely choose what they want to work on.

tests/python/test_basic.py
```diff
@@ -271,14 +272,14 @@ def test_DMatrix_init_from_path(self):
...
+        data: np.ndarray = np.random.randn(100, 2)
+        target: np.ndarray = np.array([0, 1] * 50)
+        features: List[str] = ['Feature1', 'Feature2']
 
-        dm = xgb.DMatrix(data, label=target, feature_names=features)
+        dm: xgb.DMatrix = xgb.DMatrix(data, label=target, feature_names=features)
 
         # save, assert exists, remove file
-        binary_path = Path("dtrain.bin")
+        binary_path: Path = Path("dtrain.bin")
```
trivialfis, line 282: I'm happy with these changes. But for the ease of working on type hints and consistency with the rest of the library, trivially deducible types don't need annotation.

trivialfis (commented):
Please add the file to:

- https://github.com/dmlc/xgboost/blob/4066d682611d8de85529deca9a952a636abbda2e/tests/ci_build/lint_python.py#L62
- https://github.com/dmlc/xgboost/blob/4066d682611d8de85529deca9a952a636abbda2e/tests/ci_build/lint_python.py#L16 (optional, but would be great to sort out for formatting when we get a chance).

Install mypy, black, and isort. Run the script in the project root. Feel free to disable `--pylint` and let the CI run it for time saving.

tests/python/test_basic.py
```diff
@@ -16,32 +18,31 @@
 class TestBasic:
     def test_compat(self):
```
trivialfis, line 19: mypy doesn't check the function if there's no annotation in the signature.
```suggestion
    def test_compat(self) -> None:
```

Anjas2005:
hey @trivialfis  is this issue still going on can i help to resolve this
If Yes what would you like me to do?

trivialfis:
@Anjas2005 thank you for volunteering. please open a different PR.

PR #9336, opened 2023-06-26
Minor documentation update: learning_to_rank.rst

jasjung:
Was not able to run the original code block in the documentation. It references `rng` before it was assigned. Also, random_state does not accept numpy.random.RandomState instance.


trivialfis (approved):
Thank you for the fix! Please note that the document is added recently, if you want to use the feature mentioned in the doc you need nightly builds.

PR #9344, opened 2023-06-28
Choose predictor only when it's training.

trivialfis:
We will remove the parameter `predictor`. After the removal, the predictor will be dispatched based on the device context. However, we keep the heuristic for choosing the CPU predictor for the first prediction round during training due to memory pressure.


trivialfis:
Will focus on https://github.com/dmlc/xgboost/pull/9129 instead.

PR #9345, opened 2023-06-28
Remove meaningless conditional judgments

gklab:


src/common/numeric.h
```diff
@@ -32,9 +32,7 @@ void RunLengthEncode(Iter begin, Iter end, std::vector<Idx>* p_out) {
       out.push_back(i);
```
gklab, line 32: According to the above for loop
`out.push_back(n);`
This line also redundant?
Or is there something wrong with my understanding?
@trivialfis

trivialfis: A quick counter-example comes to mind, what if n==0?

gklab: > A quick counter-example comes to mind, what if n==0?

Thanks for your answer！

PR #9346, opened 2023-06-28
Fix device communicator dependency

rongou:
Having a circular dependency between communicator and device communicator seems to cause address sanitizer errors when destructing them.


rongou:
@trivialfis

trivialfis (commented):
Thank you for the fix. Please address the tidy error.

rongou:
> Thank you for the fix. Please address the tidy error.

Done.

trivialfis (approved):

PR #9348, opened 2023-06-29
Use mmap in external memory demo.

trivialfis:


trivialfis:
On second thought, let's not complicate the demo.

Issue #6706, opened 2021-02-13
Python build: --use-system-libxgboost cannot find library


adamjstewart:
Hi, I'm a developer for the [Spack](https://spack.io) package manager. Unlike other package managers, Spack installs each package to a different directory, allowing users to install multiple versions of the same library. I'm working on overhauling our build recipes for `xgboost` and `py-xgboost` and noticed the following difficulties:

1. `python setup.py build` does not accept the `--use-system-libxgboost` flag, even though the build phase is where we decide whether or not to build the C++ library
2. `setup.py` cannot find `libxgboost.(so|dylib)` at build-time, as it is in a custom location
3. `xgboost/libpath.py` cannot find `libxgboost.(so|dylib)` at run-time, as it is in a custom location

I can get around issue 1 by only running `python setup.py install`, but all other Python packages in Spack have a 2-phase installation process with `python setup.py build && python setup.py install`. I can get around issues 2 and 3 by patching the source code to tell it where the library is found, but this is fragile if the code changes from version to version. I can get around all 3 issues by building the entire xgboost library when I build the Python bindings, but that leads to a lot of code duplication between the two Spack packages.

Some suggestions for how this could be fixed:

3. Add a `--use-system-libxgboost` flag to the `build` phase. This should be fairly easy and non-controversial.
4. Change the `--use-system-libxgboost` flag to optionally accept a path at which the library can be found
5. Store the path passed to `--use-system-libxgboost` at build-time and write it to `xgboost/libpath.py` so it can be found at run-time

There are also a couple of alternatives we could use:

6. Search `LIBRARY_PATH` at build-time and `(DY)?LD_LIBRARY_PATH` at run-time like other C/C++ libraries do
7. Allow users to pass in a directory in `setup.cfg` or `site.cfg`. This is how NumPy handles non-Python dependencies like BLAS/LAPACK.

I've numbered each issue and suggestion separately so it will be easier to comment on them.

trivialfis:
Thanks for the detailed description and solutions.  I'm open to sol 3, 4, 5.  Would you like to open a PR?

adamjstewart:
I'll see what I can do!

trivialfis:
That certainty looks like an interesting new package manager.  I have been using nix and conda and some others.  I should try Spack too.

adamjstewart:
[...]

PR #9349, opened 2023-06-30
Enable Installation of Python Package with System lib in a Virtual Environment

oliverholworthy:
## Goal

Enable Installation of Python Package with System lib in a Virtual Environment

## Details

- Replaces `sys.prefix` with `sys.base_prefix`
  - these are different when inside a virtual environment. Using  `sys.base_prefix` instead of `sys.prefix` enables installation of package in a virtual environment when the xgboost is already installed on the system

> **sys.base_prefix**
> Set during Python startup, before site.py is run, to the same value as `prefix`. If not running in a virtual environment, the values will stay the same; if site.py finds that a virtual environment is in use, the values of `prefix` and `exec_prefix` will be changed to point to the virtual environment, whereas `base_prefix` and `base_exec_prefix` will remain pointing to the base Python installation (the one which the virtual environment was created from).

## Motivation

Setting up tests to run inside a virtual environment in an image that has xgboost pre-installed as a system lib. (e.g. `nvcr.io/nvidia/tensorflow:23.06-tf2-py3`). When in a virtual environment, the `sys.prefix` changes to point to the `lib` directory inside the virtualenvironment instead of the system lib path. 
- Installing xgboost with the `use_system_libxgboost=True` in a virtual environment
- Using a `tox` virtual environment with `sitepackages=true` to pickup an xgboost installation present on the system already

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -275,7 +275,7 @@ jobs:
 
       - name: Install ninja
         run: |
-          apt-get update && apt-get install -y ninja-build
+          sudo apt-get update && sudo apt-get install -y ninja-build
 
       - name: Build XGBoost on Ubuntu
         run: |
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -108,6 +108,37 @@ jobs:
         cd ..
         python -c 'import xgboost'
 
+  python-system-installation-test-on-Linux:
+    runs-on: ${{ matrix.os }}
+    name: Test installing XGBoost Python with System Installation on Linux
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: ["3.8"]
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: 'true'
+    - name: Install Ubuntu system dependencies
+      run: |
+        sudo apt-get install -y --no-install-recommends ninja-build
+    - name: Build and install XGBoost
+      shell: bash -l {0}
+      run: |
+        mkdir build
+        cd build
+        cmake ..
+        make -j$(nproc)
+        cd ..
+        cp lib/* $(python  -c 'import sys; print(sys.base_prefix)')/lib
+        pip install virtualenv
+        virtualenv venv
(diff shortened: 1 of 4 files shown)


trivialfis (commented):
Thank you for the work on improved support with venv. However, I got a bit confused by the paths now. Maybe we need a new option that's more general as discussed in https://github.com/dmlc/xgboost/issues/6706, specifically https://github.com/dmlc/xgboost/issues/6706#issuecomment-813783117. Without a new option, it's difficult to handle the diversity of packaging methods. In the case of venv, the shared can be in `/usr/lib` or `myvenv/lib`, or `myvenv/lib/python3.x/site-packages/xgboost/lib`.

hcho3 (changes requested):
There are times when we need to use sys.prefix. Example: libxgboost is installed inside a Conda enviornment.

trivialfis:
I think mamba always copies the python executable, so the base_prefix and the prefix are the same:
```
Python 3.10.12 | packaged by conda-forge | (main, Jun 23 2023, 22:40:32) [GCC 12.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.base_prefix
'/home/jiamingy/.anaconda/envs/xgboost_dev'
>>> sys.prefix
'/home/jiamingy/.anaconda/envs/xgboost_dev'
```
I expect conda to have the same behavior.

hcho3:
It would be nice to make paths user configurable

hcho3 (approved):
I realize that this PR doesn't break Conda. For now this change is fine. In a future PR, we can add user configuration for paths.

trivialfis (approved):

trivialfis:
I canceled the job for the stalling test. It's caused by an issue in column-split test, which has been fixed in master branch. Will rebase the PR.

trivialfis:
@hcho3 @oliverholworthy After reading the mamba document: https://mamba.readthedocs.io/en/latest/advanced_usage/more_concepts.html#linking I think this PR might be problematic if there are users who prefer soft-links.

hcho3:
Thanks, I opened #9376 as a reminder.

PR #9356, opened 2023-07-04
Support exporting cut values

trivialfis:
Deviating from the rest of the API functions, the value is returned as a json array interface. The choice is optional, but I want to start the discussion of this choice as we will return more complicated arrays in the future including CUDA arrays.

- Add functions to export cut values in C and Python.
- Move ellpackpage into `src` for accessing `HistogramCuts`.
- Tests.

Downstream projects can use the information to compress the tree model if needed.

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -822,7 +822,7 @@ XGB_DLL int XGDMatrixNumNonMissing(DMatrixHandle handle, bst_ulong *out);
  * @since 1.7.0
  *
  * \param handle the handle to the DMatrix
- * \param config Json configuration string. At the moment it should be an empty document,
+ * \param config JSON configuration string. At the moment it should be an empty document,
  *               preserved for future use.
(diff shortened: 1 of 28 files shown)


RAMitchell (approved):
Is this a feature request?

trivialfis:
No. I'm looking into enabling inference libraries to use the quantization result to compress the model.

hcho3 (approved):
LGTM. (I only reviewed the Python portion)

Currently, Treelite performs `unique` operation to get the cut values from thresholds. This functionality will be useful, as we can skip the `unique` operation.

trivialfis:
@hcho3 Please note that this doesn't work with approx, which changes cut value for each iteration

trivialfis:
One will have to get the value for every iteration. Currently, this has to be done through some hacks.

hcho3:
For now, Treelite can continue to scan the thresholds manually to compute the smallest integer width that will fit all thresholds.

PR #9362, opened 2023-07-06
Define the new `device` parameter.

trivialfis:
Related: https://github.com/dmlc/xgboost/issues/7308

A new `device` parameter for replacing the previous `gpu_id`. The new parameter can take three different values and two additional aliases:
- `cpu`
- `cuda`
- `cuda:<ordinal>` with `<ordinal>` as an integer
- `gpu` alias to cuda
- `gpu:<ordinal>` alias to cuda

The compatibility with `gpu_id` is kept. A warning is emitted when it's used. Along with the new parameter, `gpu_hist` is sunset.

After this PR, we still have many necessary changes in documents and demonstrations. I will leave them as follow-up PRs.

TODOs:
- [x] "GPU"
- [x] lower case.
- [x] Spark
- [x] PySpark
- [x] Test custom obj with deprecated `gpu_hist` and the new `hist` + `device`. (prediction cache, gradient)
- [x] Test inplace predict with mismatched data type and device.

### DIFF
--- a/doc/parameter.rst
+++ b/doc/parameter.rst
@@ -68,8 +68,8 @@ General Parameters
     + ``cpu``: Use CPU.
     + ``cuda``: Use a GPU (CUDA device).
     + ``cuda:<ordinal>``: ``<ordinal>`` is an integer that specifies the ordinal of the GPU (which GPU do you want to use if you have more than one devices).
-    + ``gpu``: Same as ``cuda``.
-    + ``gpu:<ordinal>``: Same as ``cuda:<ordinal>``.
+    + ``gpu``: Default GPU device selection from the list of available and supported devices. Only ``cuda`` devices are supported currently.
+    + ``gpu:<ordinal>``: Default GPU device selection from the list of available and supported devices. Only ``cuda`` devices are supported currently.
 
 Parameters for Tree Booster
 ===========================


trivialfis:
@WeichenXu123 Would you like to take a look at the changes in the pyspark module?

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/rapids/spark/GpuPreXGBoost.scala
```diff
@@ -280,7 +280,7 @@ object GpuPreXGBoost extends PreXGBoostProvider {
             // - gpu id
             // - predictor: Force to gpu predictor since native doesn't save predictor.
             val gpuId = if (!isLocal) XGBoost.getGPUAddrFromResources else 0
-            booster.setParam("gpu_id", gpuId.toString)
+            booster.setParam("device", gpuId.formatted("cuda:%d"))
```
wbo4958, line 283: a concise way to format string, s"cuda:$gpuId"

wbo4958:
it's good from spark side.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/rapids/spark/GpuPreXGBoost.scala
trivialfis: Done. Thank you for the suggestion!

trivialfis:
regex hangs on mingw ...

doc/gpu/index.rst
```diff
@@ -22,21 +22,22 @@ Supported parameters
...
 
-  param['gpu_id'] = 0
+  param["device"] = "cuda:0"
   param['tree_method'] = 'gpu_hist'
 
 .. code-block:: python
   :caption: With Scikit-Learn interface
 
-  XGBRegressor(tree_method='gpu_hist', gpu_id=0)
+  XGBRegressor(tree_method='gpu_hist', device="cuda")
```
RAMitchell, line 40: What if I specify gpu_hist and device cpu?

trivialfis: You will receive a warning saying `gpu_hist` is deprecated. There's a comment in the test, to summarize, we should emit an error. However, it's difficult to detect whether the device is set to CPU because it's the default or because it's specified by the user. As a result, we let it slip.

doc/parameter.rst
```diff
@@ -59,6 +59,18 @@ General Parameters
...
+* ``device`` [default= ``cpu``]
+
+  .. versionadded:: 2.0.0
+
+  - Device for XGBoost to run. User can set it to one of the following values:
+
+    + ``cpu``: Use CPU.
+    + ``cuda``: Use a GPU (CUDA device).
+    + ``cuda:<ordinal>``: ``<ordinal>`` is an integer that specifies the ordinal of the GPU (which GPU do you want to use if you have more than one devices).
+    + ``gpu``: Same as ``cuda``.
```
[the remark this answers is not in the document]
trivialfis: Absolutely, that's why we have `gpu` and `cuda`. It's equal to `cuda` "for now", in the future others can make variants.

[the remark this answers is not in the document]
trivialfis: @napetrov Feel free to share your suggestions. ;-) Would like to get more opinions.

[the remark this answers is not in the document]
napetrov: I mean from API perspective users would be writing code assuming GPU device are always CUDA, but this would results in breaking change when there would be another GPU backend.

Might be it worth to point that this is not the same but just a convenient way  for selecting default GPU device - so in long-term this would result in GPU dispatching regardless of particular HW.   i.e. now this is not same as 'cuda' but is default GPU device selector although it can select only from ['cuda'].  In this way we would clearly define expectations from API and would allow extensions here

napetrov, line 71:
```suggestion
    + ``gpu``: Default GPU device selection from the list of available and supported devices. Only ''cuda'' devices are supported currently.
```

Something along this lines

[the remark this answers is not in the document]
trivialfis: Thank you for sharing, will change the document as suggested.

trivialfis: Done.

PR #9363, opened 2023-07-06
Unify the hist tree method for different devices.

trivialfis:
- Dispatch the tree method based on the current context.
- Deprecate the use of `gpu_hist`.

### DIFF
--- a/src/common/error_msg.cc
+++ b/src/common/error_msg.cc
@@ -1,0 +1,36 @@
+/**
+ * Copyright 2023 by XGBoost contributors
+ */
+#include "error_msg.h"
+
+#include "xgboost/logging.h"
+
+namespace xgboost::error {
+void WarnDeprecatedGPUHist() {
+  bool static thread_local logged{false};
+  if (logged) {
+    return;
+  }
+  auto msg =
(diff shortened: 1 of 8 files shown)


src/common/error_msg.cc
```diff
@@ -0,0 +1,36 @@
...
+  }
+  auto msg =
+      "The tree method `gpu_hist` is deprecated since 2.0.0. To use GPU training, set the `device` "
+      R"(parameter to CUDA instead.
+
+    E.g. tree_method = "hist", device = "CUDA"
+
+)";
+  LOG(WARNING) << msg;
+  logged = true;
```
hcho3, line 22: Does the `device` parameter exist already? Then why do the C++ tests use `gpu_id` ?

trivialfis: Not yet, it's on this PR: https://github.com/dmlc/xgboost/pull/9362 . It's difficult to split up the changes without referencing each other.

hcho3: Got it. Let me approve this PR for now.

hcho3 (approved):

PR #9364, opened 2023-07-06
Set feature_names and feature_types in jvm-packages

jinmfeng001:
1. Support setting and getting feature_name and feature_type in DMatrix.scala.
2. Support setting and getting feature info(feature_name and feature_type) to and from booster.
3. Change booster Serialization format to **json**, so that the feature_name and feature_type can be loaded back correctly.
4. Add params to support setting feature_names and feature_types which will be passed to native functions.


jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/GeneralParams.scala
```diff
@@ -177,6 +177,21 @@ private[spark] trait GeneralParams extends Params {
 
   final def getSeed: Long = $(seed)
 
+  /** Feature's name, it will be set to DMatrix and Booster, and in the final native json model.
+   * In native code, the parameter name is feature_name.
+   * */
+  final val featureNames = new StringArrayParam(this, "feature_names",
+  "an array of feature name")
```
trivialfis, line 184:
```suggestion
  "an array of feature names")
```

jvm-packages/xgboost4j/src/native/xgboost4j.cpp
```diff
@@ -1148,3 +1148,69 @@ JNIEXPORT jint JNICALL Java_ml_dmlc_xgboost4j_java_XGBoostJNI_XGDMatrixGetStrFea
...
+
+  std::vector<std::string> features;
+  std::vector<char *> features_char;
+
+  for (bst_ulong i = 0; i < feature_num; ++i) {
+    jstring jfeature = (jstring)jenv->GetObjectArrayElement(jfeatures, i);
+    const char *s = jenv->GetStringUTFChars(jfeature, 0);
+    features.push_back(std::string(s, jenv->GetStringLength(jfeature)));
+    if (s != nullptr) jenv->ReleaseStringUTFChars(jfeature, s);
+    if (features.back().length() == 0) features.pop_back();
```
trivialfis, line 1175: Why is it necessary to pop it? We might want to raise an error if there's an empty string.

jvm-packages/xgboost4j/src/native/xgboost4j.cpp
```diff
@@ -1148,3 +1148,69 @@ JNIEXPORT jint JNICALL Java_ml_dmlc_xgboost4j_java_XGBoostJNI_XGDMatrixGetStrFea
...
+  }
+
+  for (size_t i = 0; i < features.size(); ++i) {
+    features_char.push_back(&features[i][0]);
```
trivialfis, line 1179: `features[i].c_str()`

jvm-packages/xgboost4j/src/native/xgboost4j.cpp
```diff
@@ -1148,3 +1148,69 @@ JNIEXPORT jint JNICALL Java_ml_dmlc_xgboost4j_java_XGBoostJNI_XGDMatrixGetStrFea
...
+    JNIEnv *jenv, jclass jclz, jlong jhandle, jstring jfield,
+    jobjectArray jfeatures) {
+  BoosterHandle handle = (BoosterHandle)jhandle;
+
+  const char *field = jenv->GetStringUTFChars(jfield, 0);
+
+  bst_ulong feature_num = (bst_ulong)jenv->GetArrayLength(jfeatures);
+
+  std::vector<std::string> features;
+  std::vector<char *> features_char;
```
trivialfis, line 1168:
```suggestion
  std::vector<char const*> features_char;
```

jvm-packages/xgboost4j/src/native/xgboost4j.cpp
```diff
@@ -1148,3 +1148,69 @@ JNIEXPORT jint JNICALL Java_ml_dmlc_xgboost4j_java_XGBoostJNI_XGDMatrixGetStrFea
...
+    if (s != nullptr) jenv->ReleaseStringUTFChars(jfeature, s);
+    if (features.back().length() == 0) features.pop_back();
+  }
+
+  for (size_t i = 0; i < features.size(); ++i) {
+    features_char.push_back(&features[i][0]);
+  }
+
+  int ret = XGBoosterSetStrFeatureInfo(
+      handle, field, (const char **)dmlc::BeginPtr(features_char), feature_num);
```
trivialfis, line 1183: With the previous suggestions, the type cast should not be necessary.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
```diff
@@ -744,7 +789,7 @@ private static long[] dmatrixsToHandles(DMatrix[] dmatrixs) {
   private void writeObject(java.io.ObjectOutputStream out) throws IOException {
     try {
       out.writeInt(version);
-      out.writeObject(this.toByteArray());
+      out.writeObject(this.toByteArray("json"));
```
trivialfis, line 792: Is it possible that we don't make this change in this PR? We want to have consistent a default for all language bindings.

trivialfis (commented):
Thank you for working on the feature!

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
jinmfeng001: We have to change it to `json` format, because the default format [SaveModel](https://github.com/dmlc/xgboost/blob/master/src/learner.cc#L1088-L1134) and [LoadModel](https://github.com/dmlc/xgboost/blob/master/src/learner.cc#L956-L1084) ignores the feature_name and feature_type fields. If we don't want to change to json format, we need to modify the SaveModel and LoadModel functions. Since the default format will be deprecated, changing format here is simple and this function is only used when serialization/deserialization between spark workers, I think we can just change it here.

jvm-packages/xgboost4j/src/native/xgboost4j.cpp
jinmfeng001: If the the feature name and feature type is "", I want to treat it as invalid so just remove it. But looks like the "" is a valid name for DMatrix and Booster, I keep the "" values now.

jvm-packages/xgboost4j/src/native/xgboost4j.cpp
jinmfeng001: Changed as requested.

jvm-packages/xgboost4j/src/native/xgboost4j.cpp
jinmfeng001: Changed as requested.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
trivialfis: Thank you for sharing, let's change it to `ubj` instead, which is more efficient.

jinmfeng001: Changed to `ubj` format, thank you.

trivialfis:
cc @wbo4958 .

Issue #9360, opened 2023-07-05
Support np arrays for Vertical Federated Learning


edumugi:
Experimenting with the federated plugin (#8424) in a vertical environment I noticed that giving data as a numpy array to the DMatrix is not supported yet.

It should be a really easy fix since we only need to support the _data_split_mode_ arg in the **_from_numpy_array** function.

Steps would be:

- Add new arg _data_split_mode_ to **_from_numpy_array** in file python-package/xgboost/data.py
- Read _data_split_mode_ from config in function **XGDMatrixCreateFromDense** inside file src/c_api/c_api.cc

Should I open a PR for these changes?

trivialfis:
The federated learning support is still working in progress. cc @rongou .

rongou:
Yes, feel free to send @trivialfis and me a PR.

trivialfis:
Thank you for the feature!

PR #9365, opened 2023-07-06
Support numpy vertical

edumugi:
Implement changes discussed in #9360.

@rongou @trivialfis

### DIFF
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -467,7 +467,7 @@ XGB_DLL int XGDMatrixCreateFromDense(char const *data,
       static_cast<DataSplitMode>(OptionalArg<Integer, int64_t>(config, "data_split_mode", 0));
   xgboost_CHECK_C_ARG_PTR(out);
   *out = new std::shared_ptr<DMatrix>(
-      DMatrix::Create(&adapter, missing, n_threads, std::string(""), data_split_mode));
+      DMatrix::Create(&adapter, missing, n_threads, "", data_split_mode));
   API_END();
 }
 
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -463,9 +463,9 @@ XGB_DLL int XGDMatrixCreateFromDense(char const *data,
   auto config = Json::Load(StringView{c_json_config});
   float missing = GetMissing(config);
   auto n_threads = OptionalArg<Integer, int64_t>(config, "nthread", 0);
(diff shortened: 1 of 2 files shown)


src/c_api/c_api.cc
```diff
@@ -464,7 +464,10 @@ XGB_DLL int XGDMatrixCreateFromDense(char const *data,
   float missing = GetMissing(config);
   auto n_threads = OptionalArg<Integer, int64_t>(config, "nthread", 0);
   xgboost_CHECK_C_ARG_PTR(out);
-  *out = new std::shared_ptr<DMatrix>(DMatrix::Create(&adapter, missing, n_threads));
+  auto data_split_mode =
```
rongou, line 467: This line should be above the check on `out`.

edumugi: Fixed!

src/c_api/c_api.cc
```diff
@@ -463,8 +463,11 @@ XGB_DLL int XGDMatrixCreateFromDense(char const *data,
   auto config = Json::Load(StringView{c_json_config});
   float missing = GetMissing(config);
   auto n_threads = OptionalArg<Integer, int64_t>(config, "nthread", 0);
+  auto data_split_mode =
+      static_cast<DataSplitMode>(OptionalArg<Integer, int64_t>(config, "data_split_mode", 0));
   xgboost_CHECK_C_ARG_PTR(out);
-  *out = new std::shared_ptr<DMatrix>(DMatrix::Create(&adapter, missing, n_threads));
+  *out = new std::shared_ptr<DMatrix>(
+      DMatrix::Create(&adapter, missing, n_threads, std::string(""), data_split_mode));
```
rongou, line 470: You can just pass in the empty string `""` here, no need to construct a `std::string`.

rongou (approved):

src/c_api/c_api.cc
edumugi: Is it OK like this?

trivialfis (commented):
@rongou Any recommendations for writing tests?

rongou:
Not sure. We haven't been testing column-split data loading on the python side. Perhaps we can add a method to `DMatrix`, something like `get_data_split_mode`, and verify it after we construct the numpy DMatrix.

trivialfis:
Got it, I will leave it for now.

trivialfis (approved):

PR #9370, opened 2023-07-08
Support column split with GPU quantile

rongou:
For now only testing numerical values. Need a followup to also test categorical features.


rongou:
@trivialfis

trivialfis (approved):

PR #9373, opened 2023-07-10
go: Add Go xgboost bindings

tmc:
This adds a good chunk of a start of Go bindings for XGBoost.

### DIFF
--- a/go/Makefile
+++ b/go/Makefile
@@ -1,0 +1,11 @@
+.PHONY: test
+test:
+	go test ./... 
+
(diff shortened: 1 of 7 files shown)


trivialfis (commented):
Hi, thank you for the initial work on the go binding. At this point, I think you will need to maintain it in a separate repository like the Julia and the Swift bindings. We can reference the binding in the doc once the binding is mature and there's a certain level of commitment: https://github.com/dmlc/xgboost/blob/master/doc/index.rst .

PR #9384, opened 2023-07-14
Support column split in gpu hist updater

rongou:


rongou:
@trivialfis I'm working on enabling column-split with the gpu hist tree method. Only added some tests here, but they seem to hang on some cuda calls. Any idea on how to debug this? Do you see anything obviously wrong?

trivialfis:
@rongou GPU doesn't support multi-target yet.

rongou:
@trivialfis I think this is finally working. Please take a look. Thanks!

trivialfis (approved):
Looks good to me!

Would you like to produce a document that briefly introduces how everything works with column split in the future?

rongou:
> Would you like to produce a document that briefly introduces how everything works with column split in the future?

Yes probably should create a tutorial.

trivialfis (approved):

trivialfis:
For some reason, this PR has optimized the GPU hist with the Higgs dataset for about 20 percent of improvement, tested with both RTX8000 and 4070 ti super.

trivialfis:
NVM, git bisect is incorrect.

PR #9385, opened 2023-07-14
[jvm-packages] Add the new `device` parameter.

trivialfis:
Related https://github.com/dmlc/xgboost/issues/7308 .

This PR is to add the `device` parameter for the Spark package. For the core implementation PR, please see https://github.com/dmlc/xgboost/pull/9362 .

### DIFF
--- a/jvm-packages/xgboost4j-gpu/src/test/java/ml/dmlc/xgboost4j/gpu/java/BoosterTest.java
+++ b/jvm-packages/xgboost4j-gpu/src/test/java/ml/dmlc/xgboost4j/gpu/java/BoosterTest.java
@@ -77,7 +77,8 @@ public class BoosterTest {
         put("objective", "binary:logistic");
         put("num_round", round);
         put("num_workers", 1);
-        put("tree_method", "gpu_hist");
+        put("tree_method", "hist");
+	put("device", "cuda");
         put("max_bin", maxBin);
       }
     };
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
@@ -73,7 +73,7 @@ private[scala] case class XGBoostExecutionParams(
     xgbInputParams: XGBoostExecutionInputParams,
     earlyStoppingParams: XGBoostExecutionEarlyStoppingParams,
     cacheTrainingSet: Boolean,
-    treeMethod: Option[String],
+    device: Option[String],
     isLocal: Boolean,
     featureNames: Option[Array[String]],
     featureTypes: Option[Array[String]]) {
@@ -180,6 +180,11 @@ private[this] class XGBoostExecutionParamsFactory(rawParams: Map[String, Any], s
         " as 'hist', 'approx', 'gpu_hist', and 'auto'")
       treeMethod = Some(overridedParams("tree_method").asInstanceOf[String])
     }
+    val device: Option[String] = overridedParams.get("device") match {
+      case None => None
+      case Some(dev: String) => if (treeMethod == "gpu_hist") Some("cuda") else Some(dev)
+    }
+
     if (overridedParams.contains("train_test_ratio")) {
       logger.warn("train_test_ratio is deprecated since XGBoost 0.82, we recommend to explicitly" +
         " pass a training and multiple evaluation datasets by passing 'eval_sets' and " +
@@ -228,7 +233,7 @@ private[this] class XGBoostExecutionParamsFactory(rawParams: Map[String, Any], s
       inputParams,
       xgbExecEarlyStoppingParams,
       cacheTrainingSet,
-      treeMethod,
+      device,
       isLocal,
       featureNames,
       featureTypes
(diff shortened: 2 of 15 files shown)


trivialfis:
cc @wbo4958 .

RAMitchell (approved):

trivialfis:
@dotbg Could you please help take a look into the PR when you are available? I'm not an expert in Scala/Spark.

dotbg:
@trivialfis  At the first glance the code looks ok. I wonder whether the `gpu` is a better name for the device option.

trivialfis:
@dotbg `gpu` is available as well, and it's currently equivalent to `cuda`. We keep it as a placeholder for the future if anyone were to have implementation for different GPU devices.

jvm-packages/xgboost4j-gpu/src/test/java/ml/dmlc/xgboost4j/gpu/java/BoosterTest.java
```diff
@@ -77,7 +77,8 @@ public void testBooster() throws XGBoostError {
         put("objective", "binary:logistic");
         put("num_round", round);
         put("num_workers", 1);
-        put("tree_method", "gpu_hist");
+        put("tree_method", "hist");
+	put("device", "cuda");
```
dotbg, line 81:
```suggestion
        put("device", "cuda");
```

dotbg: will the `put("tree_method", "gpu_hist");`  also work?

dotbg:
I see, in this case, it may be good(but not a showstopper) to align naming in other places as well.

trivialfis:
> will the put("tree_method", "gpu_hist"); also work?

It should work but with a warning. I will add a test tomorrow.

> I see, in this case, it may be good(but not a showstopper) to align naming in other places as well.

Could you please elaborate on this? Apologies for not having providing a clearer context before. I'm currently walking through each interface https://github.com/dmlc/xgboost/issues/7308#issuecomment-1435248520 . The syntax is documented in [here](https://github.com/dmlc/xgboost/blob/master/doc/parameter.rst#general-parameters), which was part of https://github.com/dmlc/xgboost/pull/9362 .

I think the flink interface doesn't support GPU yet. But I will double-check the native Java interface and the Scala interface, I think they don't have hard-coded parameters that require changes (feel free to correct me).

The R interface doesn't have hard-coded parameters, and the CRAN package doesn't support GPU.

The Python interface is mostly handled in previous PRs, I will have some more specialized handling for PySpark. The naming of parameters are consistent.

dotbg:
@trivialfis well, one of my concerns is that the package names will contain `gpu`, not `cuda`. It is not a huge deal if documented properly. But it would be great to document what is required if someone decides to make code `OpenCL` comiant

trivialfis:
I think this part should be fine, we have documents on how GPU support is achieved for both general XGB packages and the JVM packages, along with notes for CUDA being the only option at the moment.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -180,6 +180,17 @@ private[this] class XGBoostExecutionParamsFactory(rawParams: Map[String, Any], s
         " as 'hist', 'approx', 'gpu_hist', and 'auto'")
       treeMethod = Some(overridedParams("tree_method").asInstanceOf[String])
     }
+    val device: Option[String] = overridedParams.get("device") match {
+      case None => None
+      case Some(dev: String) => if (treeMethod == "gpu_hist") Some("cuda") else Some(dev)
+    }
+    if (!device.isEmpty) {
+      require(
+        !device.contains(":"),
```
wbo4958, line 189: this is not going to work.

wbo4958:
```scala
val device = overridedParams.get("device").map(dev => if (treeMethod == "gpu_hist") "cuda" else dev)

device.foreach(str => require(!str.contains(":"), "Please don't specify the device ordinal as GPUs are managed by Spark."))
```

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -318,7 +329,7 @@ object XGBoost extends Serializable {
       val externalCheckpointParams = xgbExecutionParam.checkpointParam
 
       var params = xgbExecutionParam.toMap
-      if (xgbExecutionParam.treeMethod.exists(m => m == "gpu_hist")) {
+      if (xgbExecutionParam.device.exists(m => (m == "cuda" || m == "gpu"))) {
```
wbo4958, line 332: so both device = "cuda" and device= "gpu" are supported?

trivialfis: Correct.

wbo4958 (approved):

Issue #6703, opened 2021-02-12
Add option to retrieve the outcome of `validate_parameters` to Python API


KoenSR:
The `xgboost.train` method in the Python API checks for the validity of the parameters provided to the `params` argument by default and raises a warning if there are any unused parameters. 

It is not possible to easily retrieve whether this warning was raised or not. As testing the validity of input parameters is important in testing ML pipelines, I would suggest adding a feature to the Python API that can be used to retrieve the results of the parameter validation.

trivialfis:
I think a better way might be rerouting the warning into Python `logging.warn` so that you can catch it.  WDYT?

svenharris:
I think it would be useful the API allowed you to pass a set of parameters into something like `assert_parameters_valid` which would check names/types and possibly the specific combination of parameters are valid together, and raise and error if the configuration is not valid or contains parts that don't get used:

```
In[0]:  xgb.assert_parameters_valid({"max_depth": "four"})

Traceback (most recent call last):
AssertionError: max_depth parameter must be integer between 1 and ...


In[1]:  xgb.assert_parameters_valid({"sketch_eps": 0.05, "tree_method": "hist})

Traceback (most recent call last):
AssertionError: "sketch_eps" only used for tree_method="approx"
```

I don't know if something like this exist internally already or if errors/warnings get raised on the fly? The documentation is helpful but I think having such a function to assert against would give a lot more certainty particularly in production scenarios

trivialfis:
The parameter validation is not precise, which should be improved in future code refactoring.  Right now xgboost doesn't raise a warning on your second example.

As for the assert function, I think the easiest way to do it is calling `booster.set_param` then call any api function, like `booster.save_config`.  Following code should print a warning.  Maybe we can reroute that warning into `logging.warn` so you can catch it with Python `catch` keyword.
``` python
booster.set_param(...)
booster.save_config()
```

The reason for not having a direct function in `xgboost` module is because most of the parameters are local to the booster instead of global variables.

trivialfis:
From above example, the json string returned by `booster.save_config` contains all the used parameter.  I think it's also possible for user to check it from there.  Just please note that parameters are dependent on each other, like `max_depth` only appears in the config when the booster is chosen to be `gbtree` or `dart`.

PR #9387, opened 2023-07-14
Turn warning messages into Python warnings.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/6703 .

The change also makes it easier for me to test various warnings. See the updated test of parameter validation for an example.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -157,7 +157,7 @@ def _log_callback(msg: bytes) -> None:
     if smsg.find("WARNING:") != -1:
         warnings.warn(smsg, UserWarning)
         return
-    print(py_str(msg))
+    print(smsg)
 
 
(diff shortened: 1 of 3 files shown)


python-package/xgboost/core.py
```diff
@@ -153,6 +153,10 @@ def _expect(expectations: Sequence[Type], got: Type) -> str:
 
 def _log_callback(msg: bytes) -> None:
     """Redirect logs from native library into Python console"""
+    smsg = py_str(msg)
+    if smsg.find("WARNING:") != -1:
+        warnings.warn(smsg, UserWarning)
+        return
     print(py_str(msg))
```
hcho3, line 160:
```suggestion
    print(smsg)
```

trivialfis: Done.

PR #9390, opened 2023-07-17
[pyspark] Handle the `device` parameter in pyspark.

trivialfis:
- Handle the new `device` parameter in pyspark.
- Deprecate the old `use_gpu` parameter.

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -203,6 +203,12 @@ class _SparkXGBParams(
         "The device type for XGBoost workers. Available options are cpu,cuda and gpu.",
         TypeConverters.toString,
     )
+    use_gpu = Param(
+        Params._dummy(),
+        "use_gpu",
+        "Deprecated, use `device` instead.",
+        TypeConverters.toBoolean,
+    )
     force_repartition = Param(
         Params._dummy(),
         "force_repartition",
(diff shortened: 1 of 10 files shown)


trivialfis:
cc @WeichenXu123 .

WeichenXu123:
emm,

This is a breaking change, it breaks our customer code unexpectedly, can we just deprecate `use_gpu` param instead of removing it suddenly ?

trivialfis:
@WeichenXu123 Restored the compatibility with an warning when `use_gpu` is set to non-None.

python-package/xgboost/spark/core.py
```diff
@@ -198,11 +197,16 @@ class _SparkXGBParams(
...
+        "device",
+        "The device type for XGBoost workers. Available options are cpu,cuda and gpu.",
+        TypeConverters.toString,
+    )
     use_gpu = Param(
         Params._dummy(),
         "use_gpu",
-        "A boolean variable. Set use_gpu=true if the executors "
-        + "are running on GPU instances. Currently, only one GPU per task is supported.",
+        "Deprecated, use `device` instead.",
```
WeichenXu123, line 209: Let's keep original doc but append Deprecated

trivialfis: Done, the notice is prepended.

PR #9392, opened 2023-07-17
[CI] [Don't merge] Test how BuildKite handles a PR that's also a branch

hcho3:


hcho3:
BuildKite doesn't block on this PR since it's also a regular branch. So #9034 was not really effective, and the CI runs for all dependabot PRs :(

PR #9396, opened 2023-07-18
Switch to per-thread default stream

rongou:
The legacy default stream does a lot of implicit synchronization, which may cause performance issues. In addition, it seems to interfere with our multi-gpu tests using the in-memory communicator.

The per-thread default stream is created for each cpu thread. More importantly, it doesn't implicitly synchronize with other streams, which makes it easier to reason about and debug concurrency issues. RAPIDS libraries like cuDF are released with per-thread default stream enabled.


rongou:
@trivialfis @RAMitchell @hcho3

src/tree/gpu_hist/row_partitioner.cuh
```diff
@@ -221,7 +221,6 @@ class RowPartitioner {
   dh::device_vector<int8_t> tmp_;
   dh::PinnedMemory pinned_;
   dh::PinnedMemory pinned2_;
-  cudaStream_t stream_;
```
trivialfis, line 224: I think this is used for async copy

trivialfis: Got it. Thank you for sharing.

trivialfis (approved):

Issue #9322, opened 2023-06-22
code change


celestinoxp:
I recently ran into a bug and lost a lot of time figuring out which features were involved.

the problem is here, because it only alerts that there are features with the same names, but it doesn't say which ones (in my case it was necessary to know that)

xgboost\core.py", line 1155, in feature_names
     `raise ValueError('feature_names must be unique')`
ValueError: feature_names must be unique

I propose that this code be changed to
 ```
duplicates = [name for name in feature_names if feature_names.count(name) > 1] #new line
raise ValueError('feature_names must be unique. Duplicates found: {}'.format(duplicates)) #line changed
```

Thus, it is possible to know which features have the same names, and avoid wasting time...

more details: https://github.com/pycaret/pycaret/issues/3615

I'm using Windows 11
Anaconda
python 3.10
xgboost 1.7.6

trivialfis:
Thank you for the suggestion! Would you like to open a PR for the improved error message?

celestinoxp:
> Thank you for the suggestion! Would you like to open a PR for the improved error message?

no thanks, if you find this post useful, can you update the code?

trivialfis:
Got it. I will update the feature validation code.

PR #9401, opened 2023-07-20
[breaking] Remove support for single string feature info.

trivialfis:
- Input must be a sequence of strings.
- Improve validation error message.
- Don't reset feature type along with feature name.

Close https://github.com/dmlc/xgboost/issues/9322 .

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -1234,11 +1234,10 @@ class DMatrix:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
     @property
     def feature_names(self) -> Optional[FeatureNames]:
-        """Get feature names (column labels).
+        """Labels for features (column labels).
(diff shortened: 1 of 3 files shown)


python-package/xgboost/core.py
```diff
@@ -1244,55 +1261,54 @@ def feature_names(self, feature_names: Optional[FeatureNames]) -> None:
```
hcho3, line 1261: Can you move this docstring to the getter? The docstring in the setter isn't visible in the generated doc. See https://stackoverflow.com/a/16025754

hcho3 (changes requested):

python-package/xgboost/core.py
trivialfis: Done.

hcho3 (approved):

Issue #8880, opened 2023-03-07
Property best_iteration not reflecting the best iteration when trained without early stopping


flaviopereira-maersk:
I think I found what I consider to be an issue related to the _best_iteration_ property of a fitted model. When training a model without early stopping, the _best_iteration_ property gets filled with the last iteration number, even though it may not be the best iteration. This can be misleading, as it implies that the last iteration was the best iteration, even when that's not the case.

To find the true best iteration, one can perform an _argmin_ on the evals_result list of RMSEs after the training is complete. This will reveal that there are better iterations than the last one - which is the one stored in the best_iteration property. The workaround for this issue is to use _early_stopping_rounds_ with a value greater than or equal to _num_estimators_.

One potential improvement to this issue would be to make the _best_iteration_ property always have the iteration number where the _eval_metric_ was the best. This would make the property more informative and accurate and would help avoid confusion and errors caused by relying on the last iteration number as the best iteration.

Also, docs say the _best_iteration_ property only has a value when early stopping is active, which is not true.

trivialfis:
Thank you for raising the issue. Let me take a closer look into it.

PR #9403, opened 2023-07-20
Define `best_iteration` only if early stopping is used.

trivialfis:
This is the behavior specified by the document but not honored in the actual code. The improvement calculated in early stopping is a bit more complicated than regular monitoring as it has options like `min_delta`. Leaving it to early stopping to prevent inconsistent behaviors.

- Don't set the attributes if there's no early stopping.
- Cleanup the code for callbacks and replace assertions with proper exceptions.
- Assign the attributes when early stopping `save_best` is used.
- Turn the attributes into Python properties.

Close https://github.com/dmlc/xgboost/issues/8880 .

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -230,9 +230,9 @@ __model_doc = f"""
     subsample : Optional[float]
         Subsample ratio of the training instance.
     sampling_method :
-        Sampling method. Used only by the GPU version of `hist` tree method.
-          - `uniform`: select random training instances uniformly.
-          - `gradient_based` select random training instances with higher probability
+        Sampling method. Used only by the GPU version of ``hist`` tree method.
+          - ``uniform``: select random training instances uniformly.
+          - ``gradient_based`` select random training instances with higher probability
             when the gradient and hessian are larger. (cf. CatBoost)
     colsample_bytree : Optional[float]
         Subsample ratio of columns when constructing each tree.
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -230,10 +230,10 @@ __model_doc = f"""
     subsample : Optional[float]
         Subsample ratio of the training instance.
     sampling_method :
-        Sampling method. Used only by `gpu_hist` tree method.
+        Sampling method. Used only by the GPU version of `hist` tree method.
           - `uniform`: select random training instances uniformly.
-          - `gradient_based` select random training instances with higher probability when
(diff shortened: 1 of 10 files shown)


python-package/xgboost/sklearn.py
```diff
@@ -230,10 +230,10 @@ def task(i: int) -> float:
     subsample : Optional[float]
         Subsample ratio of the training instance.
     sampling_method :
-        Sampling method. Used only by `gpu_hist` tree method.
+        Sampling method. Used only by the GPU version of `hist` tree method.
```
hcho3, line 233:
```suggestion
        Sampling method. Used only by the GPU version of ``hist`` tree method.
```

python-package/xgboost/sklearn.py
```diff
@@ -230,10 +230,10 @@ def task(i: int) -> float:
...
           - `uniform`: select random training instances uniformly.
```
hcho3, line 234:
```suggestion
          - ``uniform``: select random training instances uniformly.
```

python-package/xgboost/sklearn.py
```diff
@@ -230,10 +230,10 @@ def task(i: int) -> float:
...
-          - `gradient_based` select random training instances with higher probability when
-            the gradient and hessian are larger. (cf. CatBoost)
+          - `gradient_based` select random training instances with higher probability
+            when the gradient and hessian are larger. (cf. CatBoost)
```
hcho3, line 236:
```suggestion
          - ``gradient_based`` select random training instances with higher probability
            when the gradient and hessian are larger. (cf. CatBoost)
```

tests/python/test_callback.py
```diff
@@ -216,20 +207,27 @@ def test_early_stopping_save_best_model(self):
...
-            cls.fit(X, y, eval_set=[(X, y)], callbacks=[early_stop])
+            cls.fit(X, y, eval_set=[(X, y)])
 
         # No error
         early_stop = xgb.callback.EarlyStopping(rounds=early_stopping_rounds,
                                                 save_best=False)
         xgb.XGBClassifier(
-            booster='gblinear', n_estimators=10, eval_metric=tm.eval_error_metric_skl
-        ).fit(X, y, eval_set=[(X, y)], callbacks=[early_stop])
+            booster='gblinear',
```
hcho3, line 226:
```suggestion
            booster="gblinear",
```

hcho3 (changes requested):

trivialfis:
@hcho3 All changes applied.

hcho3 (approved):

Issue #9377, opened 2023-07-11
[jvm package] xgboost model training early stop at minimum auc value


henghamao:
jvm package cited in pom.xml:
```
<dependency>
            <groupId>ml.dmlc</groupId>
            <artifactId>xgboost4j_2.12</artifactId>
            <version>1.7.6</version>
  </dependency> 
```
The model training code:
```
public void train(DMatrix trainMatrix){
        int round = 100; 
        int earlyStop = 10;
            
        HashMap<String, Object> mapParams = new HashMap<>(); 
        mapParams.put("eta", 0.1); 
        mapParams.put("objective", "binary:logistic");
        mapParams.put("max_depth", 3); 
        mapParams.put("eval_metric", "auc"); 
        mapParams.put("silent", 0);
        
        HashMap<String, DMatrix> mapWatches = new HashMap<>(); 
        mapWatches.put("selTrain", trainMatrix ); 

        try{
            booster = XGBoost.train(trainMatrix, mapParams, round, mapWatches, null, null, null, earlyStop); 
            System.out.println(booster.getAttrs() ); //打印训练的结果
            saveModel();
        }
        catch (Exception e) {
            e.printStackTrace();
            log.error("Error to train XGBoost Model", e.getMessage());
        }
    }
```
The model stop at the fist round which had the minimum auc value, while we expected the model to stop at the maximun auc value.

trivialfis:
cc @wbo4958

wbo4958:
sure, I will check it.

wbo4958:
@henghamao, could you add `mapParams.put("maximize_evaluation_metrics", true);` and retry?

henghamao:
Thanks！
By adding the line "mapParams.put("maximize_evaluation_metrics", true);", the training stopped at the maxmize AUC.
```
[10:07:09] [91] selTrain-auc:0.61312234112458341

[10:07:09] [92] selTrain-auc:0.61317036169818728

[10:07:09] [93] selTrain-auc:0.61344198401778538

[10:07:09] [94] selTrain-auc:0.61385812077231050

[10:07:09] [95] selTrain-auc:0.61395870490803284

[10:07:09] [96] selTrain-auc:0.61397982565356402

[10:07:09] [97] selTrain-auc:0.61431707070547215

[10:07:10] [98] selTrain-auc:0.61436486693250414

[10:07:10] [99] selTrain-auc:0.61446896184963551

{best_score=0.61446893, best_iteration=99}
[...]

PR #9404, opened 2023-07-21
[jvm-packages] automatically set the max/min direction for best score

wbo4958:
To fix https://github.com/dmlc/xgboost/issues/9377

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
@@ -17,6 +17,8 @@ package ml.dmlc.xgboost4j.java;
 
 import java.io.*;
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -29,6 +31,11 @@ import org.apache.hadoop.fs.FileSystem;
  */
 public class XGBoost {
   private static final Log logger = LogFactory.getLog(XGBoost.class);
+
+  public static final String[] MAXIMIZ_METRICES = {
+    "auc", "aucpr", "pre", "pre@", "map", "ndcg",
+    "auc@", "aucpr@", "map@", "ndcg@",
+  };
 
   /**
    * load model from modelPath
@@ -158,7 +165,7 @@ public class XGBoost {
     //collect eval matrixs
     String[] evalNames;
     DMatrix[] evalMats;
-    float bestScore;
+    float bestScore = 1;
     int bestIteration;
     List<String> names = new ArrayList<String>();
     List<DMatrix> mats = new ArrayList<DMatrix>();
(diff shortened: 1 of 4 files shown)


jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -360,16 +376,50 @@ static boolean shouldEarlyStop(int earlyStoppingRounds, int iter, int bestIterat
...
       String maximize = String.valueOf(params.get("maximize_evaluation_metrics"));
-      assert(maximize != null);
       return Boolean.valueOf(maximize);
-    } catch (Exception ex) {
-      logger.error("maximize_evaluation_metrics has to be specified for enabling early stop," +
-              " allowed value: true/false", ex);
-      throw ex;
     }
+
+    if (params.get("eval_metric") != null) {
```
trivialfis, line 405: If the user has specified metric name, it should reflect in the log, could you please help confirm that?

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -360,16 +376,50 @@ static boolean shouldEarlyStop(int earlyStoppingRounds, int iter, int bestIterat
...
+      // user has special metric name
+      metricName = String.valueOf(params.get("eval_metric"));
+    } else {
+      // infer the metric name from log
+      metricName = getMetricNameFromlog(evalInfo, evalNames);
+    }
+
+    assert metricName != null;
+
+    if (!"mape".equals(metricName)) {
```
trivialfis, line 415: why mape?

trivialfis (commented):
Hey, thank you for the fix!

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
wbo4958: Yes, if the user sets the eval_matric, the log will contain it.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
wbo4958: I just followed the python way to handle the direction, https://github.com/dmlc/xgboost/blob/f7f673b00c15458fb4dd74a2a0d2ba80369c5faf/python-package/xgboost/callback.py#L384

trivialfis: Ah, that's the `mape` starting with `map`.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
trivialfis: Do we still need to check whether `eval_metric` is null given that we can obtain all the info from the log?

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
wbo4958: yes,

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
wbo4958: good point. But let's just keep it there. no need to infer it again.

trivialfis (approved):

PR #9414, opened 2023-07-24
Initial GPU support for the approx tree method.

trivialfis:


trivialfis:
The `approx` tree method on GPU is still at early stage with many low-hanging fruits for optimization. For instance, when input is on the CPU, 60 percent of the training time is spent on copying data from the CPU to GPU for batched ellpack construction.

RAMitchell (approved):

PR #9416, opened 2023-07-25
Retry switching to per-thread default stream

rongou:
Added a cmake option to turn it off if something goes wrong. Fixed the failing mgpu pytests caused by thread safety issues with `cub::CachingDeviceAllocator`.


rongou:
@trivialfis @hcho3

trivialfis (approved):

trivialfis:
Nice fix!

PR #9421, opened 2023-07-26
Raise if expected workers are not alive in `xgboost.dask.train`

### LINKED ISSUE #9419: `xgboost.dask.train` gets stuck indefinitely if worker holding data from `DaskDMatrix` left
In the current design, `DaskDMatrix` records the workers storing the individual partitions of the dataset. `xgboost.dask.train` then expects that number of workers to be around in order to start training: https://github.com/dmlc/xgboost/blob/7579905e187fbbc2cec48299f7ea36a0365ba4d8/python-package/xgboost/dask.py#L854 If a worker left for some reason (e.g. a worker ran out of memory or a spot instance was interrupted) and does not return, `xgboost.dask.train` will get stuck indefinitely. 

**Reproducer:**

```python3
import dask.array as da
import xgboost as xgb 

from distributed import Worker
from distributed.utils_test import gen_cluster, async_poll_for


@gen_cluster(client=True)
async def test_worker_left(c, s, a, b):
    async with Worker(s.address):
        dx = da.random.random((300, 10)).rechunk(chunks=(10, None))
        dy = da.random.random((300,)).rechunk(chunks=(10,))
        d_train = await xgb.dask.DaskDMatrix(
            c, dx, dy,
        )
    await async_poll_for(lambda: len(s.workers) == 2, timeout=5)
    await xgb.dask.train( 
        c,
        {},
        d_train,
        evals=[(d_train, "train")],
    )
```

Versions:
* `dask=2023.7.1`
* `distributed=2023.7.1`

### LINKED ISSUE #9420: `xgboost.dask.train` gets stuck indefinitely if worker holding data from `DaskDMatrix` restarts
Similar to #9419, the current design of `DaskDMatrix` is problematic because it records the workers storing the individual partitions of the dataset. If a worker left, but enough workers are around to satisfy 
https://github.com/dmlc/xgboost/blob/7579905e187fbbc2cec48299f7ea36a0365ba4d8/python-package/xgboost/dask.py#L85
(.e.g., because the worker restarted and returned with a new address or there are additional workers), then `xgboost.dask.train` starts the training phase and pins a training task to the worker that already left the cluster. This will cause the task to transition into the `no-worker` state where it will remain indefinitely and `xgboost.dask.train` will get stuck.

**Reproducer:**

```python3
import dask.array as da
import xgboost as xgb 

from distributed import Nanny
from distributed.utils_test import gen_cluster


@gen_cluster(client=True, Worker=Nanny)
async def test_worker_restarted(c, s, a, b):
    dx = da.random.random((300, 10)).rechunk(chunks=(10, None))
    dy = da.random.random((300,)).rechunk(chunks=(10,))
    d_train = await xgb.dask.DaskDMatrix(
        c, dx, dy,
    )
    await c.restart_workers([a.worker_address])
    await xgb.dask.train( 
        c,
        {},
        d_train,
        evals=[(d_train, "train")],
    )
```
Versions:

* `dask=2023.7.1`
* `distributed=2023.7.1`

A possible fix for #9419 and this issue would be to replace `wait_for_workers` with a check that the expected workers are around, and raise an error if not.

hendrikmakait:
* Closes #9419
* Closes #9420

This PR checks whether the workers expected to hold the data from `DaskDMatrix` are still alive so that training jobs can then be scheduled on those workers. This avoids several deadlock scenarios.

Note that this PR does not make `xgboost.dask.train` resilient against an expected worker leaving _after_ the check.

### DIFF
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -905,8 +905,9 @@ def _filter_empty(
     raise ValueError("None of the workers can provide a valid result.")
 
 
-def _check_workers_are_alive(workers: List[str], client: "distributed.Client") -> None:
-    current_workers = client.scheduler_info()["workers"].keys()
+async def _check_workers_are_alive(workers: List[str], client: "distributed.Client") -> None:
+    info = await self.scheduler.identity()
+    current_workers = info["workers"].keys()
     missing_workers = set(workers) - current_workers
     if missing_workers:
         raise RuntimeError(f"Missing required workers: {missing_workers}")
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -906,7 +906,7 @@ def _filter_empty(
 
 
 async def _check_workers_are_alive(workers: List[str], client: "distributed.Client") -> None:
-    info = await self.scheduler.identity()
+    info = await client.scheduler.identity()
     current_workers = info["workers"].keys()
     missing_workers = set(workers) - current_workers
     if missing_workers:
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -905,7 +905,9 @@ def _filter_empty(
     raise ValueError("None of the workers can provide a valid result.")
 
 
-async def _check_workers_are_alive(workers: List[str], client: "distributed.Client") -> None:
+async def _check_workers_are_alive(
+    workers: List[str], client: "distributed.Client"
+) -> None:
     info = await client.scheduler.identity()
     current_workers = info["workers"].keys()
     missing_workers = set(workers) - current_workers
--- a/python-package/xgboost/dask.py
+++ b/python-package/xgboost/dask.py
@@ -905,8 +905,7 @@ def _filter_empty(
     raise ValueError("None of the workers can provide a valid result.")
 
 def _check_workers_are_alive(workers: List[str], client: "distributed.Client"):
-    info = client.scheduler_info()
-    current_workers = info["workers"].keys()
+    current_workers = client.scheduler_info()["workers"].keys()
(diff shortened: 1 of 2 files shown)


fjetter (approved):
LGTM

python-package/xgboost/dask.py
```diff
@@ -907,6 +905,13 @@ def _filter_empty(
     raise ValueError("None of the workers can provide a valid result.")
 
 
+def _check_workers_are_alive(workers: List[str], client: "distributed.Client") -> None:
```
trivialfis, line 908: Not entirely sure what replacing wait with check does on some clusters. The wait was there for some cluster types like GKE, where the worker can take some time to be in sync with the scheduler somehow and `client.scheduler_info()["workers"].` was not particularly reliable.

trivialfis (commented):
Thank you for raising the issue, some questions in comments.

python-package/xgboost/dask.py
hendrikmakait: @trivialfis: You raise a good point here. When using an async client, `client.scheduler_info()` would get refreshed by a periodic callback. So this _could_ lead to a situation where `client.scheduler_info()` is outdated. I've updated the PR to request the data directly from the scheduler, ensuring that `workers` is up-to-date. Generally, if a worker is alive and has executed work (e.g., loaded partitions of the `DaskDMatrix`), the scheduler will be aware of the worker.

trivialfis:
Could you please take a look into the failing Python tests?

hendrikmakait:
> Could you please take a look into the failing Python tests?

Sure, I'll try to reproduce it locally.

hendrikmakait:
@trivialfis: The tests should work now.

trivialfis (approved):

PR #9430, opened 2023-08-01
Clean up MGPU C++ tests

rongou:
Currently the MGPU gtests require multiple GPUs to run, and are skipped if only one GPU is found. This is mostly because NCCL doesn't like having multiple communicators on a single GPU. Other than the tests specifically testing NCCL, they can run on a single GPU if

* they don't rely on NCCL
* or we switch to the `DeviceCommunicatorAdapter` with single GPU

This makes testing distributed GPU logic easier on a single GPU. The CI pipeline with multiple GPUs available will still run the full setup with NCCL.

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -27,21 +27,21 @@ jobs:
           persist-credentials: false
 
       - name: "Run analysis"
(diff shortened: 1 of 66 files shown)


rongou:
@trivialfis @hcho3

trivialfis (approved):

PR #9436, opened 2023-08-03
Explicitly specify libcudart_static in CMake config

hcho3:
* `nvcc` defaults to statically linking `libcudart`, so this PR merely codifies the current behavior. This is useful in case where we were to use clang to build XGBoost with GPU support, as clang uses dynamic linking for `libcuart` unless explicitly stated otherwise.
* Patch RMM's CMake config to use the static cudart.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -232,6 +232,15 @@ add_subdirectory(${xgboost_SOURCE_DIR}/plugin)
 
 if (PLUGIN_RMM)
   find_package(rmm REQUIRED)
+
+  # Patch the rmm targets so they reference the static cudart
+  # Remove this patch once RMM stops specifying cudart requirement
+  # (since RMM is a header-only library, it should not specify cudart in its CMake config)
+  get_target_property(rmm_link_libs rmm::rmm INTERFACE_LINK_LIBRARIES)
+  list(REMOVE_ITEM rmm_link_libs CUDA::cudart)
+  list(APPEND rmm_link_libs CUDA::cudart_static)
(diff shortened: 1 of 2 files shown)


hcho3:
@trivialfis Can I get a review?

CMakeLists.txt
```diff
@@ -232,6 +232,15 @@ add_subdirectory(${xgboost_SOURCE_DIR}/plugin)
 
 if (PLUGIN_RMM)
   find_package(rmm REQUIRED)
+
+  # Patch the rmm targets so they reference the static cudart
+  # Remove this patch once RMM stops specifying cudart requirement
+  # (since RMM is a header-only library, it should not specify cudart in its CMake config)
```
trivialfis, line 238: even though it's header-only, having cudart as it's link interface in CMake target makes sense as cudart is indeed its dependency.

hcho3: The point is that the application should be able to choose between static cudart and dynamic cudart.

hcho3: @vyasr Can you comment on how RMM will handle cudart in the future?

hcho3: @trivialfis  See https://github.com/rapidsai/rmm/issues/1320

trivialfis: @hcho3 @jakirkham Feel free to merge if this unblocks a critical issue with conda packaging. We can continue the discussion on the rmm thread.

jakirkham:
Thanks Philip & Jiaming! 🙏

PR #9438, opened 2023-08-04
Switch back to the GPUIDX macro

rongou:
CLion seems to have trouble running individual tests with the `GetGPUId()` function although it works correctly from the command line. Switching back to the macro seems to fix it.


rongou:
@trivialfis

trivialfis (approved):

Issue #9372, opened 2023-07-08
[FR] Xgboost training run out of memory in some cases, can we add memory threshold config to prevent OOM ?


WeichenXu123:
One of our customer is facing a issue of Xgboost training run out of memory :

The customer trains xgboost model in distributed way, and when setting xgboost "max_depth" to a high value, the training routine easily runs out of memory.

The issue is, they found, for some of their dataset, using a specific "max_depth" value it works fine, but for some other dataset, using the same  "max_depth"  but OOM occurs, 

we hope to set a larger "max_depth" in most cases for better model accuracy, but, we also need to ensure preventing OOM happening, then this is a pain point,

Can we add a param like "xgboost_train_worker_cpu_memory_usage_threshold" and "xgboost_train_worker_GPU_memory_usage_threshold", and xgboost training worker tracks its memory usage, when it finds it exceeds the threshold, then it stop increasing the model depth and finalize the model and then stop training ?

Related ticket: https://github.com/dmlc/xgboost/issues/9342

trivialfis:
Same issue with https://github.com/dmlc/xgboost/issues/9342 I'm looking into it, might take some time.

Issue #9342, opened 2023-06-27
xgboost.dask.train (cpu) high memory usage when increasing max_depth


ncclementi:
We are trying to understand memory usage for xgboost.dask.train and what is the expected behavior as max_depth increases.  

Running a LocalCluster in my laptop mac m1 - 16GB memory
dask version = 2023.6.0
xgboost version = 1.7.6

Notes: I tried to run max_depth 20 and it killed my laptop. 

Reproducer (fabricated example, to showcase problem)

```python
import dask
import xgboost
from dask.distributed import Client
from distributed.diagnostics import MemorySampler

client = Client()
ms = MemorySampler()

dtypes = {}

for i in range(40):
    dtypes["f-%d" % i] = float
    
for i in range(40):
    dtypes["i-%d" % i] = int
    
df = dask.datasets.timeseries(
    start="2000-01-01",
    end="2001-01-01",
    dtypes=dtypes,
    partition_freq="4d",
    freq="10s",
)

df = df.persist()  # this is a 1.9GiB dataset 
X = df.drop("i-1", axis=1)
y = df["i-1"]

d_train = xgboost.dask.DaskDMatrix(
    None, X, y, enable_categorical=True
)


#this will give the curve corresponding to 18 but you can run it with different values and get the plot below
with ms.sample("md_18"):

    model = xgboost.dask.train(
        None,
    {
            "max_depth": 18, 
        },
        d_train,
        evals=[(d_train, "train")],
    )
    
ms.plot(align=True, grid=True)   
```

![md_6-12-18](https://github.com/dmlc/xgboost/assets/7526622/eb8d718d-92ec-438d-8494-21671138bf61)

We noticed that if we include `num_parallel_trees` things are different, but it's not clear to us why, note that the cases for 18 and 20 were taking too long to finish, so we cut them off (they were over 10 min on the 7th iteration)

```python
    model = xgboost.dask.train(
        None,
    {
            "max_depth": 18, 
            "num_parallel_tree": 100, 
        },
        d_train,
        evals=[(d_train, "train")],
    )
```

`num_parallel_trees = 100`
![md_12-18-20_npt100](https://github.com/dmlc/xgboost/assets/7526622/615ea1b8-2f8e-41f7-9583-66daeb596603)

trivialfis:
[...]

PR #9440, opened 2023-08-04
Bound the size of the histogram cache.

trivialfis:
Second attempt for https://github.com/dmlc/xgboost/pull/9432 . Less intrusive and smaller overhead when the size doesn't overflow.

- Bound the size of the histogram cache. Currently, the value is 1 << 16 nodes.
- Unify the code path between hist, multi-hist, and approx.

Close https://github.com/dmlc/xgboost/issues/9342
Close https://github.com/dmlc/xgboost/issues/9372

Aside from the raised issue, this also helps multi-target tree as memory usage can be significant when the number of targets is large.

todos:
- [x] Test distributed
- [x] unittest
- [x] Benchmark for both exceeding the bound and within the bound.
- [x] Test mixed batch for QDM and external memory.

### notes
- When I try to run tests with `max_depth=20`, the CPU utilization is low and consistently under 60 percent. Further profiling is needed in the future.
- There's no performance change when the size is within bound.
- Memory usage is bounded, and OOM can be avoided when the tree is deep. Tested with max_depth=20, without the PR, the process is killed on my 128GB memory desktop.
- The parameter for limiting the size is internal at the moment, as we can't reuse the parameter for GPU yet.

### DIFF
--- a/src/common/threading_utils.h
+++ b/src/common/threading_utils.h
@@ -7,13 +7,14 @@
 #include <dmlc/common.h>
 #include <dmlc/omp.h>
 
-#include <algorithm>
-#include <cstdint>  // for int32_t
-#include <cstdlib>  // for malloc, free
-#include <limits>
+#include <algorithm>    // for min
+#include <cstddef>      // for size_t
+#include <cstdint>      // for int32_t
+#include <cstdlib>      // for malloc, free
+#include <functional>   // for function
 #include <new>          // for bad_alloc
-#include <type_traits>  // for is_signed
-#include <vector>
+#include <type_traits>  // for is_signed, conditional_t
+#include <vector>       // for vector
 
 #include "xgboost/logging.h"
 
@@ -25,6 +26,8 @@ inline int32_t omp_get_thread_limit() __GOMP_NOTHROW { return 1; }  // NOLINT
 
 // MSVC doesn't implement the thread limit.
 #if defined(_OPENMP) && defined(_MSC_VER)
+#include <limits>
+
 extern "C" {
 inline int32_t omp_get_thread_limit() { return std::numeric_limits<int32_t>::max(); }  // NOLINT
 }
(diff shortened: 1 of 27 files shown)


hcho3 (approved):

src/common/threading_utils.h
```diff
@@ -84,8 +87,8 @@ class BlockedSpace2d {
   // dim1 - size of the first dimension in the space
   // getter_size_dim2 - functor to get the second dimensions for each 'row' by row-index
   // grain_size - max size of produced blocks
-  template <typename Func>
-  BlockedSpace2d(std::size_t dim1, Func getter_size_dim2, std::size_t grain_size) {
+  BlockedSpace2d(std::size_t dim1, std::function<std::size_t(std::size_t)> getter_size_dim2,
+                 std::size_t grain_size) {
```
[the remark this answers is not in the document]
trivialfis: Just wanted the compiler to do checks for me. ;-)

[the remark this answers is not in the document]
trivialfis: And more self-explanatory API.

src/common/threading_utils.h
```diff
@@ -129,22 +132,22 @@ class BlockedSpace2d {
...
-  std::vector<size_t> first_dimension_;
+  std::vector<std::size_t> first_dimension_;
 };
 
 
 // Wrapper to implement nested parallelism with simple omp parallel for
-template <typename Func>
-void ParallelFor2d(const BlockedSpace2d& space, int nthreads, Func func) {
+inline void ParallelFor2d(BlockedSpace2d const& space, std::int32_t n_threads,
+                          std::function<void(std::size_t, Range1d)> func) {
```
hcho3, line 141: Using `std::function` in a tight loop may have a performance implication, since `func` is no longer inlined.

trivialfis: This is not used in a tight loop, it's used to create parallel loops.

trivialfis: My fault, it's used in a tight loop.

trivialfis: But some benchmarks with higgs dataset show no difference before and after the PR. I can revert the change if necessary.

hcho3: You can just revert this line. The constructor can remain the same since it's not in a tight loop.

PR #9443, opened 2023-08-08
Handle UTF-8 paths correctly on Windows platform 

### LINKED ISSUE #9406: Handle UTF-8 paths correctly on Windows platform
I've got an error while using xgb.load() to load a JSON model file which was saved by xgb.save().
Like this: 
Error in xgb.Booster.handle(modelfile = modelfile) 
I found that if the model file path did not contain Chinese characters and it will work perfect!
The version of xgboost package is 1.7.5.1.
By the way, if I use xgb.save() with the extension 'model' and load the model file, it also works.
Hoping this issue will be fixed in the future release.

hcho3:
Closes #9406

The fix is limited in scope. It does not fix every file I/O but only the parts that save and load the Booster object.
* `XGBoosterSaveModel` uses the DMLC stream, which already correctly handles UTF-8 paths.
* `XGBoosterLoadModel` is revised to handle UTF-8 paths, by using the C++17 filesystem API.

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -1221,7 +1221,7 @@ XGB_DLL int XGBoosterPredictFromCudaColumnar(BoosterHandle handle, char const *v
  * \brief Load model from existing file
  *
  * \param handle handle
- * \param fname File URI or file name.
+ * \param fname File URI or file name. The string must be UTF-8 encoded.
  * \return 0 when success, -1 when failure happens
  */
 XGB_DLL int XGBoosterLoadModel(BoosterHandle handle,
@@ -1230,7 +1230,7 @@ XGB_DLL int XGBoosterLoadModel(BoosterHandle handle,
  * \brief Save model into existing file
  *
  * \param handle handle
- * \param fname File URI or file name.
+ * \param fname File URI or file name. The string must be UTF-8 encoded.
  * \return 0 when success, -1 when failure happens
  */
 XGB_DLL int XGBoosterSaveModel(BoosterHandle handle,
--- a/tests/cpp/c_api/test_c_api.cc
+++ b/tests/cpp/c_api/test_c_api.cc
@@ -178,7 +178,7 @@ TEST(CAPI, JsonModelIO) {
   learner->UpdateOneIter(0, p_dmat);
   BoosterHandle handle = learner.get();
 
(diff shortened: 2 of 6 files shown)


tests/cpp/c_api/test_c_api.cc
```diff
@@ -162,7 +163,7 @@ TEST(CAPI, ConfigIO) {
 TEST(CAPI, JsonModelIO) {
   size_t constexpr kRows = 10;
   size_t constexpr kCols = 10;
-  dmlc::TemporaryDirectory tempdir;
```
hcho3, line 165: The current implementation of `dmlc::TemporaryDirectory` is not capable of handling UTF-8 paths. We should deprecate `dmlc::TemporaryDirectory` in favor of the standard function `std::filesystem::temp_directory_path`.

trivialfis (approved):

include/xgboost/c_api.h
```diff
@@ -1221,7 +1221,7 @@ XGB_DLL int XGBoosterPredictFromCudaColumnar(BoosterHandle handle, char const *v
  * \brief Load model from existing file
  *
  * \param handle handle
- * \param fname File URI or file name.
+ * \param fname File URI or file name. The string must be UTF-8 encoded.
```
trivialfis, line 1224: Out of curiosity, what happens if it's not encoded in utf-8?

hcho3: If the path is ASCII, it's automatically a valid UTF-8 string.
If the path is not ASCII and is encoded with another encoding, you will get an undefined behavior.

That said, all current language bindings encode paths using UTF-8:

* Java: https://github.com/dmlc/xgboost/blob/7ce090e775e017f7850a279befddb1d694c0c9fe/jvm-packages/xgboost4j/src/native/xgboost4j.cpp#L690C30-L690C47
* Python: https://github.com/dmlc/xgboost/blob/7ce090e775e017f7850a279befddb1d694c0c9fe/python-package/xgboost/core.py#L2502 (`c_str` uses the UTF-8 encoding)
* R: I will need to file a follow-up PR to adopt UTF-8 encoding for `xgb.save` and `xgb.load` https://github.com/dmlc/xgboost/blob/7ce090e775e017f7850a279befddb1d694c0c9fe/R-package/R/xgb.save.R#L46

PR #9444, opened 2023-08-08
[CI] Check compiler warnings.

trivialfis:


tests/ci_build/build_via_cmake.sh
```diff
@@ -24,7 +24,7 @@ fi
 rm -rf build
 mkdir build
 cd build
-cmake .. ${cmake_args} -DGOOGLE_TEST=ON -DUSE_DMLC_GTEST=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DENABLE_ALL_WARNINGS=ON -GNinja ${cmake_prefix_flag} -DHIDE_CXX_SYMBOLS=ON
+cmake .. ${cmake_args} -DGOOGLE_TEST=ON -DUSE_DMLC_GTEST=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DENABLE_ALL_WARNINGS=ON -DCMAKE_COMPILE_WARNING_AS_ERROR=ON -GNinja ${cmake_prefix_flag} -DHIDE_CXX_SYMBOLS=ON
```
hcho3, line 27: How about warnings coming from dmlc-core? Would they also count as errors?

trivialfis: I think so, yes. I helped removed all GCC warnings in dmlc before.

PR #9451, opened 2023-08-09
Add `save_raw` method to `XGBModel`

judahrand:
This method was missing from `XGBModel` but would be very useful.

### DIFF
--- a/tests/python/test_with_sklearn.py
+++ b/tests/python/test_with_sklearn.py
@@ -990,6 +990,98 @@ def test_save_load_model():
         assert clf.best_score == score
 
 
+def save_raw_load_model(raw_format):
+    from sklearn.datasets import load_digits
+    from sklearn.model_selection import KFold
+
+    digits = load_digits(n_class=2)
+    y = digits['target']
+    X = digits['data']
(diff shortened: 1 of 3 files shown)


tests/python/test_with_sklearn.py
```diff
@@ -990,6 +990,98 @@ def test_save_load_model():
...
 
 
+def save_raw_load_model(raw_format):
+    from sklearn.datasets import load_digits
+    from sklearn.model_selection import KFold
+
+    digits = load_digits(n_class=2)
+    y = digits['target']
+    X = digits['data']
+    kf = KFold(n_splits=2, shuffle=True, random_state=rng)
```
trivialfis, line 1000: Considering this is model IO, maybe we can use a lighter-weight test without grid search and with early stopping enabled so that the test can cover more IO-related features in a more efficient way?

trivialfis (commented):
Thank you for the feature! Could you please look at the GitHub action errors (not buildkite)?

tests/python/test_with_sklearn.py
judahrand: Yeah, that's fair. I'd just copy pasta'd the tests for `save_model`. I'll look at simplifying this when I get a chance.

Issue #9452, opened 2023-08-09
Large memory spike after introduction of histogram size bound


fjetter:
https://github.com/dmlc/xgboost/pull/9440 introduced a bound to histogram sizes but the behavior is not entirely what I would expect.

For small-ish values of `max_depth` (I tested 12) I am indeed observing a reduction of memory usage and an improvement in end to end runtime performance when running on dask (looks like network is much more consistently utilized which is nice). Good job!

However, when dialing up the depth of the tree, I am actually observing a higher memory usage than before.

The below plots show cluster memory usage when running on a dask cluster with four workers.

Stable (xgboost 1.7.6)

![image](https://github.com/dmlc/xgboost/assets/8629629/c97450fe-201f-4a28-befa-6d940f98f86f)

Nightly (merge commit of https://github.com/dmlc/xgboost/pull/9440)

![image](https://github.com/dmlc/xgboost/assets/8629629/25f54dd2-3573-490e-b662-3f864f2428e9)

Not only is the new version exhibiting a very steep spike that is higher than what is on stable, the usage also seems to grow with increasing number of boosting rounds. On larger examples I could observe this to be even more severe.

To generate these plots I was using the same code as provided in  but was running the examples on Coiled. Similar examples should work when running on a LocalCluster

```python
import dask
import xgboost
from dask.distributed import Client
from distributed.diagnostics import MemorySampler

cluster = coiled.Cluster(
    n_workers=4,
    worker_vm_types=["r6i.xlarge"]
)
client = Client(cluster)

ms = MemorySampler()

dtypes = {}

for i in range(40):
    dtypes["f-%d" % i] = float
    
for i in range(40):
    dtypes["i-%d" % i] = int
    
df = dask.datasets.timeseries(
    start="2000-01-01",
    end="2001-01-01",
    dtypes=dtypes,
    partition_freq="4d",
    freq="10s",
)

df = df.persist()  # this is a 1.9GiB dataset 
X = df.drop("i-1", axis=1)
y = df["i-1"]

d_train = xgboost.dask.DaskDMatrix(
    None, X, y, enable_categorical=True
)

for num_boost_round in [5, 10, 15]:
    for max_depth in [18]:
        print(f"{max_depth=}_{num_boost_round=}")
        with ms.sample(f"stable-{max_depth=}_{num_boost_round=}"):
            model = xgboost.dask.train(
                None,
                {
                    "max_depth": max_depth, 
                },
                d_train,
[...]

PR #9455, opened 2023-08-09
Use `realloc` for histogram cache and expose the cache limit.

trivialfis:
Following are plots with 4 local workers from `LocalCluster`. The initial spike in the plot is not accurate. Please note that memory usage with `LocalCluster` does not represent real world usage. The histogram cache is built for each worker.

The memory spike observed in https://github.com/dmlc/xgboost/issues/9452 is caused by `std::vector::resize`. I'm not sure which one is actually better for a longer term.

The memory usage can not be directly compared to 1.7.x as we have a new estimation method for intercept. To disable it, set the `base_score` to `0.5` (default in 1.7).

I have exposed the limit as a parameter to the user interface so that there's at least an option when things get dire.

Close https://github.com/dmlc/xgboost/issues/9452 .

- Default configuration of max cache:

figure-16

- Cache size limited to 4096:

figure-12

### DIFF
--- a/doc/parameter.rst
+++ b/doc/parameter.rst
@@ -226,6 +226,15 @@ Parameters for Tree Booster
     - ``one_output_per_tree``: One model for each target.
     - ``multi_output_tree``:  Use multi-target trees.
 
+* ``max_cached_hist_node``, [default = 65536]
+
+  Maximum number of cached nodes for CPU histogram.
+
+  .. versionadded:: 2.0.0
+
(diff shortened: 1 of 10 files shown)


hcho3 (approved):

trivialfis:
BTW, I did some benchmark, the performance is unchanged.

trivialfis:
cc @WeichenXu123 for awareness.

PR #9457, opened 2023-08-10
[R] remove default values in internal utility functions

jameslamb:
Some of the internal-only functions in `R-package/R/utilities.R` have default values for some arguments.

This PR proposes removing those defaults, to force any calling code to specify values for all arguments. It also proposes switching the calls to those functions to using keyword arguments.

Taken together, I think these changes make bugs of the form "forgot to pass through configuration" and "passed through arguments in the wrong order" slightly less likely, and make it a bit easier to read through and understand the code.


trivialfis (approved):
Thank you for the cleanups!

PR #9458, opened 2023-08-10
[Doc] fix typos in documentation

jameslamb:
I was looking for some quick ways to help out tonight, so went through the documentation in `doc/` with `aspell check`.

This PR proposes fixes to the typos I found.


trivialfis (approved):
Thank you for sharing how to do the check! I should use the checker more often.

PR #9460, opened 2023-08-10
Fix pyspark parameter.

trivialfis:
- Don't pass the `use_gpu` parameter to the learner.
- Fix GPU approx with PySpark.

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -352,7 +352,7 @@ class _SparkXGBParams(
         tree_method = self.getOrDefault(self.getParam("tree_method"))
         if tree_method == "exact":
             raise ValueError(
-                f"The `{tree_method}` tree method is not supported on GPU."
+                "The `exact` tree method is not supported for distributed systems."
             )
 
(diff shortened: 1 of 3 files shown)


trivialfis:
@WeichenXu123 @wbo4958 Could you please help review when you are available?

WeichenXu123 (approved):

python-package/xgboost/spark/core.py
```diff
@@ -349,9 +350,7 @@ def _validate_params(self) -> None:
             )
 
         tree_method = self.getOrDefault(self.getParam("tree_method"))
-        if (
-            self.getOrDefault(self.use_gpu) or use_cuda(self.getOrDefault(self.device))
-        ) and not _can_use_qdm(tree_method):
+        if tree_method == "exact":
```
wbo4958, line 353: according to the raise msg, shouldn't add "use_gpu and tree_method=="exact"?

trivialfis: Good catch, changed to `not supported on distributed systems`.

PR #9461, opened 2023-08-10
[R] remove default values in internal booster manipulation functions

jameslamb:
Similar to #9457.

Proposes removing default values in internal functions `xgb.handleToBooster()` and `xgb.Booster.handle()`, to force calling code to supply all arguments and therefore reduce the risk of "forgot to pass something and silently fell back to a default" types of bugs.


trivialfis (approved):
Thank you for the refactoring!

jameslamb:
Ok thanks very much, please do tell me if that ever changes. I want to be respectful of your time and the project's limited resources.

Issue #9352, opened 2023-06-30
get_dump(dump_format='json') and dump_model(..., dump_format='json') return invalid JSON strings when special characters are present in feature names


KWiecko:
#### Rationale:

At the end of the model training workflow we need to convert `xgboost` booster to `coremltools` MLModel. For that we use `coremltools` converter, which under the hood calls `get_dump(dump_format='json')` to perform conversion. It looks like `get_dump(dump_format='json')` returns invalid JSON strings if special characters are present in feature names.

#### Detailed bug description:

Feature names which contain 'non standard' characters (e.g. double quotes ' " ', tab '\t', newline '\n' symbols) are not properly dumped to JSON by `get_dump(dump_format='json')` and `dump_model(..., dump_format='json')` methods. All feature names (no matter which characters they contain) seem to just be wrapped with double quotes which leads to invalid JSON strings if a feature name contains a 'special' character.

I'm not 100% sure but my guess is that `JsonGenerator` class (localed in `xgboost/src/tree/tree_model.cc`) is responsible for dumping trees to JSONs so maybe replacing "" wrapping  (`..., "split": "{fname}", ...`)  with a json.dumps() statement would solve this issue?

`save_model('*.json')` always dumps model to a valid JSON representation.

### Environment

- xgboost: 1.4.2 and 1.7.6
- ubuntu 22.04, Amazon Linux 2
- installed via pip

### Minimum reproducible example:

```python
import numpy as np
import json
import xgboost as xgb


if __name__ == '__main__':

    x = np.random.rand(100, 3)
    y = np.random.rand(100, 1)

    # set dummy params
    params = {'objective': 'reg:squarederror', 'eval_metric': 'rmse', 'max_depth': 4}

    # following feature names will not cause problems when calling .get_dump(dump_format='json') or dump_model(..., dump_format='json')
    correctly_dumped_feature_names = ['feature 0', 'feature 1', 'feature 2']
    dm_correctly_dumped_feature_names = xgb.DMatrix(x, label=y, feature_names=correctly_dumped_feature_names)
    correctly_dumped_booster = xgb.train(params=params, dtrain=dm_correctly_dumped_feature_names, num_boost_round=3)

    # get dump and attempt to iterate over trees JSONs
    correct_model_dump = correctly_dumped_booster.get_dump(dump_format='json')

    for tree_json_index, tree_json in enumerate(correct_model_dump):
        # all trees correctly loaded
        json.loads(tree_json)

[...]

PR #9474, opened 2023-08-11
Handle special characters in JSON model dump.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/9352 .

- Reuse the string handling logic in `Json`.

The test coverage for special characters is not quite complete at the moment. Expect some failures with features that need to parse the dump output. On top of my mind for things to test:
- Dataframe dump on Python and R.
- Feature importance.

I tested model plot, which uses the `dot` format and it works fine with special characters.

### DIFF
--- a/src/common/json.cc
+++ b/src/common/json.cc
@@ -1,23 +1,29 @@
-/*!
- * Copyright (c) by Contributors 2019-2022
+/**
+ * Copyright 2019-2023, XGBoost Contributors
  */
 #include "xgboost/json.h"
 
-#include <dmlc/endian.h>
+#include <array>             // for array
+#include <cctype>            // for isdigit
+#include <cmath>             // for isinf, isnan
+#include <cstdio>            // for EOF
+#include <cstdlib>           // for size_t, strtof
+#include <cstring>           // for memcpy
+#include <initializer_list>  // for initializer_list
+#include <iterator>          // for distance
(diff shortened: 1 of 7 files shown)


src/common/json.cc
```diff
@@ -57,12 +63,12 @@ void JsonWriter::Visit(JsonObject const* obj) {
 }
 
 void JsonWriter::Visit(JsonNumber const* num) {
-  char number[NumericLimits<float>::kToCharsSize];
-  auto res = to_chars(number, number + sizeof(number), num->GetNumber());
+  std::array<char, NumericLimits<float>::kToCharsSize> number;
+  auto res = to_chars(number.data(), number.data() + number.size(), num->GetNumber());
```
hcho3, line 67: Is it possible to use `std::to_chars`, now that we are using C++17? Not blocking this release, let's consider using `std::to_chars` in the future.

hcho3 (approved):

src/common/json.cc
trivialfis: The implementation is unspecified by the standard. MSVC can produce different result than GCC.

hcho3: Even for integers?

trivialfis: Not for interger. I will do some cleanups after 2.0

PR #9476, opened 2023-08-14
[R] remove more uses of default values in internal functions

jameslamb:
Similar to #9457 and #9461.

Proposes removing default values in internal functions in the R package, to force calling code to supply all arguments and therefore reduce the risk of "forgot to pass something and silently fell back to a default" types of bugs.


trivialfis (approved):
Thank you for the cleanups!

PR #9478, opened 2023-08-14
[jvm-packages] throw exception when tree_method=approx and device=cuda

wbo4958:


trivialfis (commented):
Thank you for the work on validation. One minor change in comment.

trivialfis (approved):

PR #9479, opened 2023-08-14
Updated versions. Reorganised dependencies.

dotbg:
I've cleaned up the root pom and moved the compile and runtime dependencies to the corresponding `xgboost4j[-gpu]` projects. Everything else depends on these and, thus, the dependency tree should be more transparent for users.


trivialfis (approved):

hcho3:
@dotbg Thank you!

dotbg:
Not at all.

PR #9488, opened 2023-08-15
Fix federated learning demos and tests

SYangster:
Fixing nvflare federated learning demos and tests


SYangster:
@rongou

trivialfis (approved):
Thank you for the fix!

Side note: I have tested the demo locally with this PR, but it would be great if we can run automated tests on CI in the future.

PR #9501, opened 2023-08-17
Fix safe_xgboost macro on c++

### LINKED ISSUE #9500: safe_xgboost macro (defined in c api docs) has syntax errors
Description:
The C / C++ api documentation specifies a macro "safe_xgboost" to be used with each language. This macro is incorrect for the C++ code, containing at least one syntax error.

One syntax error is that after XGBGetLastError(), two closing parethesis are used, however, there is only one corresponding opening parethesis.

Another syntax error I found was in the use of "Exception". This was not recognized in my environment, however I fixed it by using std::runtime_error instead.

TZProgrammer:
Closes #9500


trivialfis (approved):
Thank you for the fix!

Issue #8087, opened 2022-07-17
[py-skl] Move feature_weights to constructor.


trivialfis:
Similar to feature types, we should move anything that's not strictly related to data to the constructor for the scikit-learn interface. Also, it will fix the pyspark interface:

```
 WARNING: /home/jiaming/Workspace/XGBoost-dev/xgboost/src/learner.cc:628: 
Parameters: { "feature_weights" } might not be used.

  This could be a false alarm, with some parameters getting used by language bindings but
  then being mistakenly passed down to XGBoost core, or some parameter actually being used
  but getting flagged wrongly here. Please open an issue if you find any such cases.

```

PR #9506, opened 2023-08-21
Move feature weight to skl parameters.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/8087 .


trivialfis:
@hcho3 Please help take a look when you are available. Would be great if we could get this into 3.0 as well.

trivialfis:
cc @rongou

rongou (approved):

PR #9514, opened 2023-08-23
Small cleanup to rabit tracker.

trivialfis:
Looking at the code, I'm not entirely sure what it does. The function `_get_tree` and `get_ring` return a valid topology and share a close structure. The removed code is actually scrambling the topology.


trivialfis:
that's weird, the ring allreduce assumes a specific structure yet it asks the tracker to assign rank.

PR #9519, opened 2023-08-24
[pyspark] Support stage-level scheduling for training

wbo4958:
XGBoost training task requires taking the whole GPU exclusively, so we have limited there's only 1 task running at the same time even for the ETL stages, which typically decreases the parallelism of spark tasks, and finally has some side-effect for the whole performance of xgboost end-2-end training.

The stage-level scheduling we are using in this PR does not change executor resources and hence does not require dynamic allocation to be enabled.  The original executors are retained. The features used there, however, are new in Spark 3.4.  We also changed CPU task resources therein to ensure only 1 task per gpu just for that barrier stage.


wbo4958:
@WeichenXu123 @trivialfis @eordentlich please help to review, Thx

python-package/xgboost/spark/core.py
```diff
@@ -886,6 +891,121 @@ def _get_xgb_parameters(
...
+        # Each training task requires cpu cores > total executor cores//2 + 1 which can
+        # make the tasks be sent to different executors.
+        #
+        # Please note that we can't set task_cpu_cores to the value which is smaller than
+        # total executor cores/2 because only task_gpu_amount can't make sure the tasks be
+        # sent to different executor even task_gpus=1.0
+
+        # Spark-rapids is a project to leverage GPUs to accelerate spark SQL and it can
+        # really help the performance. If spark-rapids is enabled. we don't allow other
+        # ETL gpu tasks running alongside training tasks to avoid OOM
```
trivialfis, line 983: I'm a bit confused by this comment. Let me try to rephrase it and see if I got it right.

```
To send the tasks to different executors, the task must require at least `total executor cores // 2 + 1` CPU cores.

Without configuring the `task_cpu_cores` to a larger value, we can't make sure tasks are sent to different executors even if `task_gpu_amount` is set to 1.

When the `spark-rapids` plugin is enabled for SQL acceleration, we need to prevent other GPU-based ETL tasks from running during training to avoid OOM errors.
```

python-package/xgboost/spark/core.py
```diff
@@ -886,6 +891,121 @@ def _get_xgb_parameters(
...
+        task_cores = (
+            int(executor_cores)
+            if "com.nvidia.spark.SQLPlugin" in spark_plugins
+            and "true" == spark_rapids_sql_enabled
+            else (int(executor_cores) // 2) + 1
+        )
+
+        # task_gpus means how many gpu slots the task requires in a single GPU,
+        # it doesn't mean how many gpu shares it would like to require, so we
+        # can set it to any value of (0, 0.5] or 1.
```
trivialfis, line 996: but not [0.5, 1)?

trivialfis (commented):
Thank you for the excellent feature! You might need to update the [document](https://github.com/dmlc/xgboost/blob/master/doc/tutorials/spark_estimator.rst#accelerate-the-whole-pipeline-for-xgboost-pyspark) as well.

python-package/xgboost/spark/core.py
```diff
@@ -886,6 +891,121 @@ def _get_xgb_parameters(
...
+        # Please note that we can't set task_cpu_cores to the value which is smaller than
+        # total executor cores/2 because only task_gpu_amount can't make sure the tasks be
+        # sent to different executor even task_gpus=1.0
+
+        # Spark-rapids is a project to leverage GPUs to accelerate spark SQL and it can
+        # really help the performance. If spark-rapids is enabled. we don't allow other
+        # ETL gpu tasks running alongside training tasks to avoid OOM
+        spark_plugins = ss.conf.get("spark.plugins", " ")
+        assert spark_plugins is not None
+        spark_rapids_sql_enabled = ss.conf.get("spark.rapids.sql.enabled", "true")
```
WeichenXu123, line 986:
```suggestion
        spark_rapids_sql_enabled = ss.conf.get("spark.rapids.sql.enabled", "true").lower()
```

python-package/xgboost/spark/core.py
```diff
@@ -886,6 +891,121 @@ def _get_xgb_parameters(
...
+            return rdd
+
+        ss = _get_spark_session()
+
+        # executor_cores will not be None
+        executor_cores = ss.sparkContext.getConf().get("spark.executor.cores")
+        assert executor_cores is not None
+
+        # Each training task requires cpu cores > total executor cores//2 + 1 which can
+        # make the tasks be sent to different executors.
```
WeichenXu123, line 975: This is hacky, and it causes CPU cores wasted, assuming one xgboost train worker can only use one cpu core, but you allocate multiple CPU cores to it.

wbo4958: Yes, that's correct. It will waste some CPU cores. Let's assume executor_cores=12, then the training task will require 7, but there are still 5 CPU cores left for the other spark task.

The ideal way to limit task number should rely on task_gpu_amount, but the current spark implementation has not supported it. We can report an issue for Spark and once Spark fixes that issue, we can rework this.

python-package/xgboost/spark/core.py
wbo4958: Yeah, spark will throw exception if it's in [0.5, 1)

wbo4958:
I did the performance test

## Env:

Spark standalone cluster with 2 Nodes each having 12 CPU cores and 64G mem, and 1 TITAN V GPU with 12 G mem.

## Dataset:

Mortgage,  total 255, 962, 232 rows, 27 features + 1 label

## Result,

- W/O stage-level scheduling,

``` console
    .config("spark.executor.memory", "60g")
    .config("spark.executor.cores", 12)
    .config("spark.task.cpus", 12)
    .config("spark.executor.resource.gpu.amount", 1)
    .config("spark.task.resource.gpu.amount", 1)
    .config("spark.sql.execution.arrow.maxRecordsPerBatch", 2000000) 
```

The training times are:  532.36, 534.46, average: 533.41 


- W/ stage-level scheduling

``` console
    .config("spark.executor.memory", "60g")
    .config("spark.executor.cores", 12)
    .config("spark.task.cpus", 1)
    .config("spark.executor.resource.gpu.amount", 1)
    .config("spark.task.resource.gpu.amount", 0.08)
    .config("spark.sql.execution.arrow.maxRecordsPerBatch", 2000000) 
```

The training times are: 383.34, 383.11, average: 383.225

The ratio is 533.41/383.225 = 1.39, which means the PR can have 39% speedup.

The ETL phase is just reading the parquet file and doing the repartition, if we involve more real ETL operations, I guess, the stage-level scheduling PR can bring more speedup.

python-package/xgboost/spark/core.py
wbo4958: Yes, you're correct.

python-package/xgboost/spark/core.py
WeichenXu123: Q:

> make the tasks be sent to different executors.

Why we need this rule ? one executor can hold multiple spark tasks, any issues it causes ?

wbo4958: This PR has bypassed this scenario that the executor.gpu.amount > 1, since it requires some special configuration to make it work. 

https://github.com/dmlc/xgboost/pull/9519/files#diff-8900fd245992de81e0a7b8cb7fcd18c37bae4473e8672e9b11fd7010cac358cfR929-R937

WeichenXu123: @wbo4958 

Then I think we can directly set executor.gpu.amount=1 and executor.cores=1 when we start spark application ?

wbo4958: @WeichenXu123, we can't use gpu resources to limit the task number. I had a design doc (supporting stage-level scheduling) for spark-rapids-ml with a similar design pattern with xgboost. the doc link can be found at https://docs.google.com/document/d/1C6s1nqNkOl0roNtRA5x1JPUuM2jykuR00JyMzgt0HvU/edit?usp=sharing which explained why we can't use GPU resources to do the task limitation.

eordentlich:
Any update on the status of this PR?

python-package/xgboost/spark/core.py
wbo4958: Done.

wbo4958:
Hi @WeichenXu123, Could you help to review it again? Thx very much.

python-package/xgboost/spark/core.py
WeichenXu123: This is bug:

Repro code:
```
import org.apache.spark.resource.{ResourceProfileBuilder, TaskResourceRequests}

val rdd = sc.range(0, 100, 1, 4)
var rdd1 = rdd.repartition(3)
val treqs = new TaskResourceRequests().cpus(1).resource("gpu", 1.0)
val rp = new ResourceProfileBuilder().require(treqs).build
rdd1 = rdd1.mapPartitions(iter => {Thread.sleep(30000); iter})withResources(rp)
rdd1.collect()  # 3 spark tasks run in parallel on one executor
```

```
spark-shell --master spark://192.168.31.236:7077 --conf spark.executor.cores=12 --conf spark.executor.resource.gpu.amount=1
```

Let's see whether spark team can address the bug.
If it can't be fixed for now, let's use current workaround.

wbo4958: Hi @WeichenXu123, Yeah, looks like it is a spark bug. But Using the task.cpus to limit the task number in this PR can still achieve the same goal as using the task.gpu.amount, except wasting some cpus cores. Can we merge this PR first, and change it after spark fixes this bug?

WeichenXu123 (approved):
LGTM,

once https://github.com/dmlc/xgboost/pull/9519#discussion_r1356038670 is fixed, please file a followup PR to remove workaround code.

wbo4958:
Hi @trivialfis, Could you help trigger CI ?

python-package/xgboost/spark/core.py
trivialfis: why?

python-package/xgboost/spark/core.py
```diff
@@ -894,6 +899,127 @@ def _get_xgb_parameters(
...
+            else (int(executor_cores) // 2) + 1
+        )
+
+        # task_gpus means how many gpu slots the task requires in a single GPU,
+        # it doesn't mean how many gpu shares it would like to require, so we
+        # can set it to any value of (0, 0.5] or 1.
+        task_gpus = 1.0
+
+        # treqs = TaskResourceRequests().cpus(task_cores).resource("gpu", task_gpus)
+        # rp = ResourceProfileBuilder().require(treqs).build
```
trivialfis, line 1013: commented code.

python-package/xgboost/spark/core.py
```diff
@@ -894,6 +899,127 @@ def _get_xgb_parameters(
...
+
+        self.logger.info(
+            "XGBoost training tasks require the resource(cores=%s, gpu=%s).",
+            task_cores,
+            task_gpus,
+        )
+        # return rdd.withResources(rp)
```
trivialfis, line 1020: commented code.

python-package/xgboost/spark/core.py
trivialfis: Could you please update the comments with a cleaner explanation?

trivialfis (commented):
Could you please provide an example usage that gets everything right (with no warning), and add comments on the example for each configuration with an explanation for why the configuration is set in this way? If I were a user I would find it quite difficult to address all potential warnings.

python-package/xgboost/spark/core.py
wbo4958: Good catch. Done

python-package/xgboost/spark/core.py
wbo4958: Good catch. Done.

python-package/xgboost/spark/core.py
wbo4958: It's a bug in spark, I deleted the comment.

wbo4958:
> Could you please provide an example usage that gets everything right (with no warning), and add comments on the example for each configuration with an explanation for why the configuration is set in such a way? If I were a user I would struggle to understand and address all potential warnings.

I will file a follow-up PR to add related doc

python-package/xgboost/spark/core.py
```diff
@@ -894,6 +900,121 @@ def _get_xgb_parameters(
...
+            task_gpu_amount = sc.getConf().get("spark.task.resource.gpu.amount")
+
+            if task_gpu_amount is None:
+                # The ETL tasks will not grab a gpu when spark.task.resource.gpu.amount is not set,
+                # but with stage-level scheduling, we can make training task grab the gpu.
+                return False
+
+            if float(task_gpu_amount) == float(executor_gpus):
+                self.logger.warning(
+                    "The configuration of cores (exec=%s, task=%s, runnable tasks=1) will "
```
trivialfis, line 960: is this CPU cores or GPU cores (I don't know what a GPU core is), but the condition is testing the number of GPUs.

python-package/xgboost/spark/core.py
```diff
@@ -894,6 +900,121 @@ def _get_xgb_parameters(
...
+
+        task_cores = (
+            int(executor_cores)
+            if "com.nvidia.spark.SQLPlugin" in spark_plugins
+            and "true" == spark_rapids_sql_enabled.lower()
+            else (int(executor_cores) // 2) + 1
+        )
+
+        # Each training task requires cpu cores > total executor cores//2 + 1 which can
+        # make sure the tasks be sent to different executors.
```
trivialfis, line 1002: Is this trying to make sure that no two tasks can fit into the same executor due to excessive CPU requirements? However, due to a bug in Spark that ignores GPU requirements, we have to hack the CPU cores for GPU tasks.

Am I correct?

python-package/xgboost/spark/core.py
```diff
@@ -894,6 +900,121 @@ def _get_xgb_parameters(
...
+
+            if ss.version < "3.4.0":
+                self.logger.warning(
+                    "Stage-level scheduling in xgboost requires spark version 3.4.0+"
+                )
+                return True
+
+            if not _is_standalone_or_localcluster(sc):
+                self.logger.warning(
+                    "Stage-level scheduling in xgboost requires spark standalone or "
```
trivialfis, line 920: Can it be used on other types of clusters (maybe in the future)? Is there any cloud service like EMR or Dataproc can be used with this?

python-package/xgboost/spark/core.py
```diff
@@ -894,6 +900,121 @@ def _get_xgb_parameters(
...
+
+            if int(executor_cores) == 1:
+                # there will be only 1 task running at any time.
+                self.logger.info(
+                    "Stage-level scheduling in xgboost requires spark.executor.cores > 1 "
+                )
+                return True
+
+            if int(executor_gpus) > 1:
+                # For spark.executor.resource.gpu.amount > 1, we suppose user knows how to configure
```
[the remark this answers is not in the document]
wbo4958: a follow PR will be filed to add related doc

python-package/xgboost/spark/core.py
wbo4958: No, it's testing how many tasks will run at the same time on the executor.

task_gpu_amount is a fraction value. we can use executor_gpus / task_gpu_amount to get the concurrent tasks.

python-package/xgboost/spark/core.py
wbo4958: Both CPU and GPU can limit the concurrent tasks. but using GPU is more optimization.  See https://github.com/dmlc/xgboost/pull/9519#discussion_r1356038670

python-package/xgboost/spark/core.py
wbo4958: Yes. from spark 3.5.1, we can enable this feature on spark standalone/ k8s / yarn cluster.

trivialfis: Do we need a condition to check for 3.5.1?

python-package/xgboost/spark/core.py
trivialfis: See you are raising a warning, could you please revise the message so that users understand what you mean by "cores"?

python-package/xgboost/spark/core.py
trivialfis: > **Both CPU and GPU can limit the concurrent tasks. but using GPU is more optimization. See https://github.com/dmlc/xgboost/pull/9519#discussion_r1356038670**

From the comment, my understanding is that GPU cannot be used to limit the concurrent tasks due to a bug in spark.

python-package/xgboost/spark/core.py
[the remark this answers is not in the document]
trivialfis: Thanks!

python-package/xgboost/spark/core.py
```diff
@@ -894,6 +900,121 @@ def _get_xgb_parameters(
...
+            if executor_cores is None or executor_gpus is None:
+                self.logger.warning(
+                    "Stage-level scheduling in xgboost requires spark.executor.cores, "
+                    "spark.executor.resource.gpu.amount to be set."
+                )
+                return True
+
+            if int(executor_cores) == 1:
+                # there will be only 1 task running at any time.
+                self.logger.info(
```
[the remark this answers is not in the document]
trivialfis: @wbo4958 Let's draw the line here:
- warning for things that users should fix/address.
- info for things that "good to know", but totally fine if I don't know.

python-package/xgboost/spark/core.py
wbo4958: yes.

python-package/xgboost/spark/core.py
wbo4958: 3.5.1 is not released yet. I will file another PR once it's available.

python-package/xgboost/spark/core.py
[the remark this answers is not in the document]
wbo4958: Changed to info

trivialfis (approved):

PR #9524, opened 2023-08-26
Small cleanup to tracker initialization.

trivialfis:
- Remove recover related code.
- Clean startup, no need to consider previously connected nodes.


trivialfis:
cc @rongou

rongou (approved):

PR #9525, opened 2023-08-26
[rabit] Drop support for MPI backend.

trivialfis:
### Motivation
- We have no test for the MPI backend.
- Nowadays, people use distributed frameworks for machine learning, they usually have their own communication channels.
- Even if the MPI support is desirable, the C++ interface of MPI is deprecated and we have to rewrite the implementation anyway.

### changes
- add checks in cmake
- Remove mpi related code.


trivialfis:
cc @hcho3 @RAMitchell

RAMitchell (approved):

PR #9527, opened 2023-08-28
Use the new `DeviceOrd` in the linalg module.

trivialfis:


src/tree/updater_gpu_hist.cu
```diff
@@ -511,8 +511,8 @@ struct GPUHistMakerDevice {
     }
 
     CHECK(p_tree);
-    dh::safe_cuda(cudaSetDevice(ctx_->gpu_id));
-    CHECK_EQ(out_preds_d.DeviceIdx(), ctx_->gpu_id);
+    dh::safe_cuda(cudaSetDevice(ctx_->Ordinal()));
+    CHECK(out_preds_d.Device().IsCUDA());
```
hcho3, line 515: Shouldn't this line compare the device of `out_preds_d` with `ctx_->Device()`?

hcho3 (approved):

hcho3:
Nice job abstracting the `gpu_id`. I have one question.

src/tree/updater_gpu_hist.cu
trivialfis: It's just CUDA, we won't be using GPU hist if context is CPU.

hcho3: What if GPU id mismatches?

trivialfis: That's a bug and might segfault due to accessing the wrong memory

hcho3: Yes, so shouldn't we have `CHECK_EQ(out_preds.Device(), ctx_->Device())` ?

trivialfis: I will have an additional check for matching context.

trivialfis: Done.

Issue #9521, opened 2023-08-25
cudaErrorIllegalAddress: an illegal memory access was encountered


Sion1225:
Hello.
While using XGBoost, I have been encountering persistent errors.

```error
File "/home/siwon/.local/lib/python3.10/site-packages/xgboost/core.py", line 271, in _check_call
    raise XGBoostError(py_str(_LIB.XGBGetLastError()))
xgboost.core.XGBoostError: [20:09:18] ../src/tree/updater_gpu_hist.cu:799: Exception in gpu_hist: [20:09:18] ../src/c_api/../data/../common/common.h:46: ../src/tree/gpu_hist/row_partitioner.cuh: 295: cudaErrorIllegalAddress: an illegal memory access was encountered
Stack trace:
  [bt] (0) /home/siwon/.local/lib/python3.10/site-packages/xgboost/lib/libxgboost.so(+0x738dca) [0x7f68a1d1fdca]
  [bt] (1) /home/siwon/.local/lib/python3.10/site-packages/xgboost/lib/libxgboost.so(+0x73ca89) [0x7f68a1d23a89]
  [bt] (2) /home/siwon/.local/lib/python3.10/site-packages/xgboost/lib/libxgboost.so(+0xaf6920) [0x7f68a20dd920]
  [bt] (3) /home/siwon/.local/lib/python3.10/site-packages/xgboost/lib/libxgboost.so(+0xaf74d1) [0x7f68a20de4d1]
  [bt] (4) /home/siwon/.local/lib/python3.10/site-packages/xgboost/lib/libxgboost.so(+0xafbfe1) [0x7f68a20e2fe1]
  [bt] (5) /home/siwon/.local/lib/python3.10/site-packages/xgboost/lib/libxgboost.so(+0xafcf66) [0x7f68a20e3f66]
  [bt] (6) /home/siwon/.local/lib/python3.10/site-packages/xgboost/lib/libxgboost.so(+0x41b919) [0x7f68a1a02919]
  [bt] (7) /home/siwon/.local/lib/python3.10/site-packages/xgboost/lib/libxgboost.so(+0x41c790) [0x7f68a1a03790]
  [bt] (8) /home/siwon/.local/lib/python3.10/site-packages/xgboost/lib/libxgboost.so(+0x47ee17) [0x7f68a1a65e17]



Stack trace:
[...]

PR #9529, opened 2023-08-28
Fix GPU categorical split memory allocation.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/9521 .

### DIFF
--- a/src/tree/updater_gpu_common.cuh
+++ b/src/tree/updater_gpu_common.cuh
@@ -64,19 +64,12 @@ struct DeviceSplitCandidate {
   // split.
   bst_cat_t thresh{-1};
 
-  common::CatBitField split_cats;
   bool is_cat { false };
 
   GradientPairInt64 left_sum;
   GradientPairInt64 right_sum;
 
   XGBOOST_DEVICE DeviceSplitCandidate() {}  // NOLINT
-
-  template <typename T>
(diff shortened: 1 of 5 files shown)


src/tree/updater_gpu_common.cuh
```diff
@@ -64,20 +64,13 @@ struct DeviceSplitCandidate {
   // split.
   bst_cat_t thresh{-1};
 
-  common::CatBitField split_cats;
```
hcho3, line 67: Trying to understand what this PR does.
Does the reference `split_cats` get invalidated as more categories are added? Why is it better to get the categories from the evaluator directly?

trivialfis: Correct. It's invalidated when more nodes are added.

hcho3: So same reason as why you shouldn't carry around references to vector elements?  Got it.

hcho3 (approved):

src/tree/updater_gpu_common.cuh
trivialfis: I think it was introduced during optimization and refactoring in 1.7. But should be fixed now. I will do some more tests myself. The error was only reproducible using a sanitizer.

PR #9531, opened 2023-08-29
[rabit] Improved connection handling.

trivialfis:
- Enable timeout with connect.
- Report connection error from the system.
- Handle retry for both tracker connection and peer connection.

### DIFF
--- a/include/xgboost/collective/result.h
+++ b/include/xgboost/collective/result.h
@@ -1,0 +1,126 @@
+/**
+ *  Copyright 2023, XGBoost Contributors
+ */
+#pragma once
+
+#include <memory>   // for unique_ptr
+#include <sstream>  // for stringstream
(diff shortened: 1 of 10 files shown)


include/xgboost/collective/result.h
```diff
@@ -0,0 +1,135 @@
...
+    return ss.str();
+  }
+};
+}  // namespace detail
+
+/**
+ * @brief An error type that's easier to handle than throwing dmlc exception. We can
+ *        record and propagate the system error code.
+ */
+struct Result {
```
trivialfis, line 75: At some point in the future, we need to propagate the error to Python or other language bindings for delegating the error handling to higher-level frameworks like dask. At the moment, a functional form of error handling is easier to handle than exceptions.

rongou (approved):

PR #9536, opened 2023-08-31
Fix inplace predict with fallback when base margin is used.

trivialfis:
- Copy meta info from proxy DMatrix.
- Use `std::call_once` to emit less warnings for thread-based prediction.


tests/cpp/gbm/test_gbtree.cu
```diff
@@ -58,21 +58,6 @@ void TestInplaceFallback(Context const* ctx) {
   HostDeviceVector<float>* out_predt{nullptr};
   ConsoleLogger::Configure(Args{{"verbosity", "1"}});
   std::string output;
-  // test whether the warning is raised
```
trivialfis, line 61: The test is removed as we are only emitting the warning once.

RAMitchell (approved):

PR #9540, opened 2023-09-04
[R] reformat build scripts

jameslamb:
Proposes some small changes to the build scripts in the R package, with to make them a bit easier to read (in my opinion) and to make diffs slightly easier to review (via breaking long blocks of arguments out into one argument per line).

**this PR is only moving code around and should not change behavior**.

I'm sorry for a style-only PR, but I do really think this will make it a bit easier for contributors to understand how the R package is built. If you'd rather not review such style-only PRs, please let me know and I'll stop submitting things like this.

### Notes for Reviewers

We've been using similar formatting in `{lightgbm}` for a few years without issue, e.g. https://github.com/microsoft/LightGBM/blob/1881a501782e0ab1e7618b3ab6ff8871ee311860/R-package/src/Makevars.win.in#L18-L22.

In that project, we also use `cmakelint` to standardize formatting in `CMakeLists.txt` and `*.cmake` scripts... do you have any interest in adopting that tool here in XGBoost? If so I'd be happy to put up a separate PR proposing its use in CI here.


trivialfis (approved):
Thank you for cleaning up the code!
> but I do really think this will make it a bit easier for contributors to understand how the R package is built

I agree. Small changes add up. Thank you for working on the R package.

> n that project, we also use cmakelint to standardize formatting in CMakeLists.txt and *.cmake scripts

That sounds awesome!

PR #9546, opened 2023-09-05
[R] remove unused headers

jameslamb:
Proposes removing some unnecessary headers in the R package's R-to-C/C++ interface.

These unnecessarily introduce a risk of name collisions and other types of compilation errors, and make it a bit more difficult to understand what the R package is relying on from R itself.

### Notes for Reviewers

For background on how routine registration works in R packages (e.g. how the `R-package/src/init.c` works), see https://cran.r-project.org/doc/manuals/R-exts.html#Registering-native-routines.

For a summary of what's in `<R.h>` and `<Rinternals.h>` from R, see https://cran.r-project.org/doc/manuals/R-exts.html#Organization-of-header-files.

### DIFF
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -20,7 +20,6 @@
 #include "../../src/common/threading_utils.h"
 
 #include "./xgboost_R.h"  // Must follow other includes.
-#include "Rinternals.h"
 
 /*!
  * \brief macro to annotate begin of api
--- a/R-package/src/init.c
+++ b/R-package/src/init.c
@@ -5,7 +5,6 @@
  * and edited to conform to xgboost C linter requirements. For details, see
  * https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Registering-native-routines
  */
-#include <R.h>
 #include <Rinternals.h>
 #include <stdlib.h>
 #include <R_ext/Rdynload.h>


R-package/src/xgboost_R.cc
```diff
@@ -20,7 +20,6 @@
 #include "../../src/common/threading_utils.h"
 
 #include "./xgboost_R.h"  // Must follow other includes.
-#include "Rinternals.h"
```
jameslamb, line 23: This one was also possibly-problematic because some compilers will treat it as "search for a file named `Rinternals.h` in the project directory".

e.g., from the `gcc` docs (https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html):

> `#include <file>`
> *This variant is used for system header files. It searches for a file named file in a standard list of system directories. You can prepend directories to this list with the `-I` option (see [Invocation](https://gcc.gnu.org/onlinedocs/cpp/Invocation.html)).*

> `#include "file"`
> *This variant is used for header files of your own program. It searches for a file named file first in the directory containing the current file, then in the quote directories and then the same directories used for `<file>`. You can prepend directories to the list of quote directories with the -iquote option.*

trivialfis:
I'm not entirely sure what happened to GitHub actions, restarted a job for testing.

jameslamb:
GitHub had a fairly significant outage today, which affected services like GitHub Actions: https://www.githubstatus.com/incidents/smdz34v7j8q0

trivialfis (approved):

trivialfis:
Apologies for the slow response.

jameslamb:
No problem at all, a 2.0 release is a lot more important than this PR 😂

PR #9553, opened 2023-09-06
Add tests for gpu_approx with column split and vertical federated

rongou:
Turns out this is already working.


rongou:
@trivialfis

trivialfis (approved):

PR #9562, opened 2023-09-08
More support for column split in gpu predictor

rongou:
* Added support for predict leaf
* More testing


rongou:
@trivialfis I think this is finally working now. Please take a look. Thanks!

src/predictor/gpu_predictor.cu
```diff
@@ -701,27 +712,39 @@ __global__ void PredictByBitVectorKernel(
...
-    for (auto tree_idx = tree_begin; tree_idx < tree_end; tree_idx++) {
-      auto const tree_group = d_tree_group[tree_idx];
-      TreeView d_tree{tree_begin,          tree_idx,           d_nodes,
-                      d_tree_segments,     d_tree_split_types, d_cat_tree_segments,
-                      d_cat_node_segments, d_categories};
-      bst_uint out_prediction_idx = row_idx * num_group + tree_group;
-      d_out_predictions[out_prediction_idx] += GetLeafWeightByBitVector(
-          row_idx, d_tree, decision_bits, missing_bits, num_nodes, tree_offset);
-      tree_offset += d_tree.d_tree.size();
+    if (num_group == 1) {
```
trivialfis, line 726: Is it necessary to distinguish whether n_groups is 1? I think the indexing should be similar if not the same right?
```
num_row * 1 + 0 # only one group
```

rongou: This is based on the original `PredictKernel` (line 299). I guess it's just a performance optimization to reduce global memory access to `d_out_predictions`?

trivialfis (approved):

PR #9568, opened 2023-09-11
Remove internal use of gpu_id.

trivialfis:
Device communicator is still using integer ordinal, which will be changed by upcoming PRs.

### DIFF
--- a/src/common/host_device_vector.cu
+++ b/src/common/host_device_vector.cu
@@ -25,8 +25,8 @@ void SetCudaSetDeviceHandler(void (*handler)(int)) {
 template <typename T>
 class HostDeviceVectorImpl {
  public:
-  HostDeviceVectorImpl(size_t size, T v, int device) : device_(device) {
-    if (device >= 0) {
+  HostDeviceVectorImpl(size_t size, T v, DeviceOrd device) : device_(device) {
+    if (device.IsCUDA()) {
       gpu_access_ = GPUAccess::kWrite;
       SetDevice();
       data_d_->resize(size, v);
@@ -37,8 +37,8 @@ class HostDeviceVectorImpl {
 
   // Initializer can be std::vector<T> or std::initializer_list<T>
   template <class Initializer>
-  HostDeviceVectorImpl(const Initializer& init, int device) : device_(device) {
-    if (device >= 0) {
+  HostDeviceVectorImpl(const Initializer& init, DeviceOrd device) : device_(device) {
+    if (device.IsCUDA()) {
       gpu_access_ = GPUAccess::kWrite;
       LazyResizeDevice(init.size());
       Copy(init);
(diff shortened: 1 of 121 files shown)


hcho3:
Why are we dropping gtest from the CI?

trivialfis:
@hcho3 windows build is failing with cmake find gtest. (same error on master branch). Since the test is building the Python package, I figure we don't need to build c++ tests as we are not running it.

hcho3:
Are we not running gtest on Windows CI?

trivialfis:
only on buildkite.

hcho3:
Ah, got it

src/common/host_device_vector.cu
```diff
@@ -138,7 +138,7 @@ class HostDeviceVectorImpl {
     } else {
       auto ptr = other->ConstDevicePointer();
       SetDevice();
-      CHECK_EQ(this->DeviceIdx(), other->DeviceIdx());
+      CHECK_EQ(this->Device().ordinal, other->Device().ordinal);
```
hcho3, line 141:
```suggestion
      CHECK(this->Device() == other->Device());
```

src/common/host_device_vector.cu
```diff
@@ -156,24 +156,25 @@ class HostDeviceVectorImpl {
...
     if (device_ == device) { return; }
-    if (device_ >= 0) {
+    if (device_.IsCUDA()) {
       LazySyncHost(GPUAccess::kNone);
     }
 
-    if (device_ >= 0 && device >= 0) {
-      CHECK_EQ(device_, device) << "New device ordinal is different from previous one.";
+    if (device_.IsCUDA() && device.IsCUDA()) {
+      CHECK_EQ(device_.ordinal, device.ordinal)
```
hcho3, line 166: Is it possible for two `DeviceOrd` objects to be different when they have the same positive value in the `ordinal` field ?

src/predictor/gpu_predictor.cu
```diff
@@ -1048,11 +1044,8 @@ class GPUPredictor : public xgboost::Predictor {
     if (tree_weights != nullptr) {
       LOG(FATAL) << "Dart booster feature " << not_implemented;
     }
-    CHECK(!p_fmat->Info().IsColumnSplit())
-        << "Predict contribution support for column-wise data split is not yet implemented.";
```
hcho3, line 1052: Is SHAP support available for column-wise split data now?

hcho3 (approved):

src/common/host_device_vector.cu
trivialfis: Not possible at the moment. Unless there's new device type other than CUDA.

src/predictor/gpu_predictor.cu
trivialfis: The check was removed between PR rebases. I will bring it back.

PR #9576, opened 2023-09-13
Utilities and cleanups for socket.

trivialfis:
- Use c++-17 nodiscard and nested ns.
- Add bind method to socket.
- Remove rabit parameters.


trivialfis:
cc @rongou

rongou (approved):

PR #9577, opened 2023-09-13
[R] Fix method name.

trivialfis:
r-devel thinks this is an unregistered s3 method.

```
Mismatches for apparent methods not registered:
format:
  function(x, ...)
format.eval.string:
  function(iter, eval_res, eval_err)
See section ‘Registering S3 methods’ in the ‘Writing R Extensions’
manual.
```

### DIFF
--- a/R-package/R/callbacks.R
+++ b/R-package/R/callbacks.R
@@ -70,7 +70,7 @@ cb.print.evaluation <- function(period = 1, showsd = TRUE) {
         i == env$begin_iteration ||
         i == env$end_iteration) {
       stdev <- if (showsd) env$bst_evaluation_err else NULL
-      msg <- format.eval.string(i, env$bst_evaluation, stdev)
+      msg <- format_eval_string(i, env$bst_evaluation, stdev)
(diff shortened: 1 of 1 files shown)


trivialfis:
cc @jameslamb

jameslamb:
We faced this in LightGBM's last attempt to release to CRAN too: https://github.com/microsoft/LightGBM/pull/5988

Sorry, I should have come and checked for it here.

By the way, if CRAN also complains about tests and examples using too many threads, you might find these links useful: https://github.com/microsoft/LightGBM/issues/5987

jameslamb (approved):
This should fix that NOTE from CRAN. It's conventional in R to use a `.` to prefix things that are intended to be internal, so I might have named this `.format_eval_string()`... but the difference isn't too important.

trivialfis:
Thank you for sharing and the suggestions!

> It's conventional in R to use a . to prefix things that are intended to be internal, so I might have named this .format_eval_string()

I will make the change, anything that makes the code better.

PR #9591, opened 2023-09-18
[R] Fix number of threads used in examples.

trivialfis:
- Set threads for data.table
- Disable parallel model IO in XGBoost for R.
- Some helper scripts.

### DIFF
--- a/R-package/.Rbuildignore
+++ b/R-package/.Rbuildignore
@@ -4,3 +4,5 @@
 ^.*\.Rproj$
 ^\.Rproj\.user$
 README.md
+^doc$
+^Meta$
--- a/R-package/R/callbacks.R
+++ b/R-package/R/callbacks.R
@@ -557,14 +557,18 @@ cb.cv.predict <- function(save_models = FALSE) {
 #'
 #' @examples
 #' #### Binary classification:
-#' #
+#'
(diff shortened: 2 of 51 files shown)


trivialfis:
tests also have a limit for 2 threads.

trivialfis:
```
Running R code in ‘testthat.R’ had CPU time 3.6 times elapsed time
* checking for unstated dependencies in vignettes ... OK
* checking package vignettes in ‘inst/doc’ ... OK
* checking re-building of vignette outputs ... [32s/11s] NOTE
Re-building vignettes had CPU time 3 times elapsed time
```

trivialfis:
@jameslamb @hcho3 Would be great if you could help review the PR when you are available, I think we are close to passing the pre-test now.

jameslamb (approved):
Changes look good to me!

I agree that the benefit of getting `{xgboost}` v2.0.0 onto CRAN are large enough that it's worth hard-coding in this 2-thread limit for model I/O.

trivialfis:
We still need to go through the reverse dependency check and talk to downstream projects. So, it might take a long time.

trivialfis:
Thank you for all the helpful information! @jameslamb .

trivialfis:
Changed base to master from 2.0.

PR #9596, opened 2023-09-20
Workaround poll on macos.

trivialfis:


trivialfis:
cc @rongou .

rabit/src/allreduce_base.cc
```diff
@@ -549,7 +549,7 @@ AllreduceBase::TryAllreduceTree(void *sendrecvbuf_,
       break;
     }
     // select must return
-    auto poll_res = watcher.Poll(timeout_sec);
+    auto poll_res = watcher.Poll(timeout_sec, false);  // fail on macos
```
rongou, line 552: What does "fail on macox" mean? The old code was failing or the new code?

trivialfis: The old code is failing but did not check. The new code checks for those errors https://github.com/dmlc/xgboost/blob/8c676c889d380a32ed61ee2b69596c3f7ef154b8/rabit/include/rabit/internal/socket.h#L96 .

trivialfis: It seems those errors don't affect the actual output on macos. The underlying reason however, is unclear to me. Maybe I should try to get a mac device somehow.

rongou: You can start a macos instance on aws.

rongou (approved):

rabit/src/allreduce_base.cc
trivialfis: I will make more checks for macos once the old rabit code can be replaced.

PR #9600, opened 2023-09-20
Load configuration before loading model.

trivialfis:
Initialize the context before loading the model. Partly addresses https://github.com/dmlc/xgboost/issues/9497#issuecomment-1728436631 .


trivialfis:
Looking at the failing tests, it takes more effort to untangle the configurations.

PR #9602, opened 2023-09-21
Use array interface for testing numpy arrays.

trivialfis:


python-package/xgboost/data.py
```diff
@@ -165,7 +165,7 @@ def _is_scipy_coo(data: DataType) -> bool:
 
 
 def _is_numpy_array(data: DataType) -> bool:
```
shwina, line 167: Perhaps now a more appropriate name for this is `_is_numpy_array_like`?

trivialfis: Thank you for the suggestion, renamed.

hcho3 (approved):

PR #9620, opened 2023-09-29
Add python api to slice columns in DMatrix

rongou:
Expose the C++ `DMatrix.SliceCol()` method in Python, mostly for testing purposes.


trivialfis:
I'd we were to focus on in-memory inputs, can we use the input for slicing like numpy? I really want to keep the DMatrix as simple as possible.

rongou:
Yeah I think we can try to slice the scipy matrix directly, but probably need better support for the in-memory types first. Will work on that.

PR #9628, opened 2023-10-04
Support column-wise data split with in-memory inputs

rongou:


include/xgboost/c_api.h
```diff
@@ -144,9 +144,7 @@ XGB_DLL int XGDMatrixCreateFromFile(const char *fname, int silent, DMatrixHandle
  *            See :doc:`/tutorials/input_format` for more info.
  *          \endverbatim
  *   - silent (optional): Whether to print message during loading. Default to true.
- *   - data_split_mode (optional): Whether to split by row or column. In distributed mode, the
- *     file is split accordingly; otherwise this is only an indicator on how the file was split
- *     beforehand. Default to row.
+ *   - data_split_mode (optional): Whether the file was split by row or column beforehand. Default to row.
```
trivialfis, line 147:
```suggestion
 *   - data_split_mode (optional): Whether the file was split by row or column beforehand for distributed computing. Default to row.
```

python-package/xgboost/core.py
```diff
@@ -1231,6 +1231,16 @@ def num_nonmissing(self) -> int:
         _check_call(_LIB.XGDMatrixNumNonMissing(self.handle, ctypes.byref(ret)))
         return ret.value
 
+    def data_split_mode(self) -> DataSplitMode:
+        """Get the data split mode of the DMatrix.
+
+        .. versionadded:: 2.0.1
```
trivialfis, line 1237:
```suggestion
        .. versionadded:: 2.1.0
```

python-package/xgboost/core.py
```diff
@@ -1298,7 +1308,10 @@ def feature_names(self, feature_names: Optional[FeatureNames]) -> None:
 
         # validate feature name
         feature_names = _validate_feature_info(
-            feature_names, self.num_col(), "feature names"
+            feature_names,
```
trivialfis, line 1311: I think we don't need to support `slice`, it's mostly used by the `cv` function, which is not supported for distributed training. I would like to remove the `slice` function when we can deprecate the `cv` function and provide document for alternatives.

rongou: Sounds good.

trivialfis (approved):

Issue #9630, opened 2023-10-06
[CI] enforce cmakelint checks in CI


jameslamb:
## Description

In #9540, I mentioned the use of `cmakelint` ([cmake-lint/cmake-lint](https://github.com/cmake-lint/cmake-lint)) to standardize formatting of `CMakeLists.txt` and `*.cmake` scripts.

It can be run like this:

```shell
cmake_files=$(
    find . -name CMakeLists.txt -o -path "./cmake/*.cmake" \
    | grep -v dmlc-core \
    | grep -v gputreeshap
)
cmakelint \
    --linelength=120 \
    --filter=-convention/filename,-package/stdargs,-readability/wonkycase \
    ${cmake_files}
```

As of latest `master` (https://github.com/dmlc/xgboost/commit/4d7a187cb0ebc2194abc1efe0aaee6f06af937c5), that yields 338 errors.

<details><summary>list of errors (click me)</summary>

```text
./demo/c-api/basic/CMakeLists.txt:6: Extra spaces between 'if' and its () [whitespace/extra]
./demo/c-api/basic/CMakeLists.txt:10: Expression repeated inside endif; better to use only endif() [readability/logic]
./demo/c-api/inference/CMakeLists.txt:7: Extra spaces between 'if' and its () [whitespace/extra]
./demo/c-api/inference/CMakeLists.txt:11: Expression repeated inside endif; better to use only endif() [readability/logic]
./CMakeLists.txt:11: Extra spaces between 'if' and its () [whitespace/extra]
./CMakeLists.txt:13: Extra spaces between 'endif' and its () [whitespace/extra]
./CMakeLists.txt:13: Expression repeated inside endif; better to use only endif() [readability/logic]
./CMakeLists.txt:89: Extra spaces between 'if' and its () [whitespace/extra]
./CMakeLists.txt:91: Extra spaces between 'endif' and its () [whitespace/extra]
./CMakeLists.txt:91: Expression repeated inside endif; better to use only endif() [readability/logic]
./CMakeLists.txt:92: Extra spaces between 'if' and its () [whitespace/extra]
./CMakeLists.txt:94: Extra spaces between 'endif' and its () [whitespace/extra]
./CMakeLists.txt:94: Expression repeated inside endif; better to use only endif() [readability/logic]
./CMakeLists.txt:95: Extra spaces between 'if' and its () [whitespace/extra]
./CMakeLists.txt:97: Extra spaces between 'endif' and its () [whitespace/extra]
[...]

PR #9637, opened 2023-10-08
[CI] standardize endif() calls in CMake scripts

jameslamb:
Contributes to #9630.

Fixes 181 of the 338 `cmakelint` errors reported in #9630, all of these two forms:

```shell
Extra spaces between 'endif' and its () [whitespace/extra]
Expression repeated inside endif; better to use only endif() [readability/logic]
```

`cmakelint` prefers this:

```cmake
if (USE_DEVICE_DEBUG AND NOT (USE_CUDA))
  message(SEND_ERROR "`USE_DEVICE_DEBUG` must be enabled with `USE_CUDA` flag.")
endif()
```

to this:

```cmake
if (USE_DEVICE_DEBUG AND NOT (USE_CUDA))
  message(SEND_ERROR "`USE_DEVICE_DEBUG` must be enabled with `USE_CUDA` flag.")
endif (USE_DEVICE_DEBUG AND NOT (USE_CUDA))
```

Including conditions in `endif()` makes CMake scripts harder to read, in my (and `cmakelint`'s) opinion, and doesn't offer any other benefits.

`CMake`'s docs even say that support for conditions is only included for backwards compatibility:

> *Per legacy, the [`else()`](https://cmake.org/cmake/help/latest/command/else.html#command:else) and [`endif()`](https://cmake.org/cmake/help/latest/command/endif.html#command:endif) commands admit an optional `<condition>` argument. If used, it must be a verbatim repeat of the argument of the opening if command.*

ref: https://cmake.org/cmake/help/latest/command/if.html


trivialfis (approved):
Looks good to me! Thank you for linting the CMake code! Hopefully, we can integrate the linter into the CI pipeline later.

PR #9650, opened 2023-10-11
[coll] Implement a new tracker and a communicator.

trivialfis:
The new tracker and communicators communicates through the use of JSON documents. Along with which, communicators are aware of each others.


trivialfis:
cc @rongou

rongou:
Wouldn't JSON be pretty slow, especially for large amount of data? Have you done any early benchmarking to compare it to the existing implementation?

trivialfis:
cc @rongou the JSON is only used for communication between worker and the tracker, not used for actual data

rongou (approved):

PR #9661, opened 2023-10-12
Skip check for pollhup.

trivialfis:
This seems to be a Windows-specific issue, one end is closed and the other end might receive a pollhup even if the data is already sent and can be fully read by the other end.

Fix flaky https://buildkite.com/xgboost/xgboost-ci-windows/builds/3636 .


trivialfis:
cc @rongou .

PR #9673, opened 2023-10-13
Update learning_to_rank.rst; Correct qid sort in snippet

lynntf:
`qid` should be sorted in the first snippet.


trivialfis (approved):
Thank you for  the fix!

PR #9674, opened 2023-10-13
[CI] address cmakelint warnings about whitespace

jameslamb:
Contributes to #9630.
Similar to #9637.

Addresses the following types of errors from `cmakelint`:

```text
Extra spaces between 'else' and its () [whitespace/extra]
Extra spaces between 'if' and its () [whitespace/extra]
Lines should be <= 120 characters long [linelength]
Mismatching spaces inside () after command [whitespace/mismatch]
Tab found; please use spaces [whitespace/tabs]
Weird indentation; use 2 spaces [whitespace/indent]
```


trivialfis (approved):
Looks good!

PR #9677, opened 2023-10-14
CMake CUDA features

chuckatkins:
This adds a few useful features to the CMake code wrt how it interacts with CUDA:

1. Remove old policies and conditional branches for `CMAKE_VERSION < 3.18`
2. Allow the builtin `CMAKE_CUDA_ARCHITECTURES` variable to be used instead of `GPU_COMPUTE_VER`, falling back to `GPU_COMPUTE_VER` if CMake is too old or `CMAKE_CUDA_ARCHITECTURES` is not specified.
3. Add a `USE_CUDA_LTO` option to enable device-code link-time-optimization.
4. Allow `CUDA_HOST_COMPILER` and `CUDA_RUNTIME_LIBRARY` to be user-overridden.

The above features are implemented in such a way to preserve existing behavior when they are not specified or enabled.


cmake/Utils.cmake
```diff
@@ -176,9 +175,17 @@ function(xgboost_set_cuda_flags target)
 
   set_target_properties(${target} PROPERTIES
     CUDA_STANDARD 17
-    CUDA_STANDARD_REQUIRED ON
-    CUDA_SEPARABLE_COMPILATION OFF
-    CUDA_RUNTIME_LIBRARY Static)
+    CUDA_STANDARD_REQUIRED ON)
+
+  if(CMAKE_INTERPROCEDURAL_OPTIMIZATION AND DEFINED CMAKE_CUDA_DEVICE_LINK_OPTIONS_IPO)
```
trivialfis, line 180: This seems to be an internal flag of CMake, I can't find any document other than its internal use https://github.com/Kitware/CMake/blob/734a009dcf0b58e9b109e9aec20f61cbc6708504/Modules/Compiler/NVIDIA-CUDA.cmake#L17 .

cmake/Utils.cmake
```diff
@@ -176,9 +175,17 @@ function(xgboost_set_cuda_flags target)
...
+    set_target_properties(${target} PROPERTIES
+      INTERPROCEDURAL_OPTIMIZATION ON
+      CUDA_SEPARABLE_COMPILATION ON)
```
trivialfis, line 183: Why is `CUDA_SEPARABLE_COMPILATION` tied to host LTO?

trivialfis (commented):
Thank you for working on modernizing the CMake code! The `CMAKE_CUDA_ARCHITECTURES` change looks good to me, and I agree the LTO might be useful. But I have a couple questions about related changes in comments.

cmake/Utils.cmake
chuckatkins: The `CMAKE_CUDA_DEVICE_LINK_OPTIONS_IPO` variable containing the LTO compiler flags for CUDA which isn't defined otherwis.  It seemed more clear than just checking for v3.25 without context but since it is documented in the release notes for v3.25 it's probably better to just check the CMAke version with a comment explaining that CUDA LTO was enabled in that CMake version.  I'll update accordingly.

cmake/Utils.cmake
chuckatkins: The implementation of LTO in both nvcc and CMake builds upon separable compilation.  In CMake LTO for device code is only enabled if SEPARABLE_COMPILATION is also enabled, otherwise it only takes effect for the hose code.  It doesn't seems to be explicitly spelled out as such in either the CMake or CUDA documentation but they do both need to be set.  From an implementation standpoint, CMake will use different gencode arguments: `code=lto_XX` during the compilation phase and `sm_XX` during a separate device linking stage to generate `cuda_device.o` which is then linked with the resulting host objects.  Again probably something worth attaching a comment to for clarity.

chuckatkins: So really `CUDA_SEPARABLE_COMPILATION` has nothing to do with host LTO but is required for device LTO.  There's no CMake level option to enable IPO / LTO on a per-language basis; CMAKE_INTERPROCEDURAL_OPTIMISATION` sets the default for all languages.  I could add a configure-time option to control only device LTO, initialized by `CMAKE_INTERPROCEDURAL_OPTIMIZATION` and `USE_CUDA`; that would allow for the additional scenarios of host-LTO without device-LTO and device-LTO without host-LTO.  Would that be prefered?

chuckatkins: So working this a little more, I don't think there's a way to only enable device lto without host lto but it can be disabled giving valid configurations for lto of host-only, host and device, or neither.

trivialfis: I don't think that's necessary. Just couldn't find documents on what you are currently using and a bit concerned on using lesser known features or relying on other project's internal behaviour.

trivialfis: As far as the changes with CUDA, I'm quite happy to see changes for using the `cmake_cuda_architecture` along with setting the new policy for CMake, these are well-documented features and I assume the upstream has tests for them to keep the behaviour stable.

> giving valid configurations for lto of host-only, host and device, or neither.

I'm OK with the host only option if we can leave out the un-documented behavior of CMake. We can open an issue to keep track of enabling CUDA version. I was told that separable compilation has performance implication on the final code (for the worse in some cases), so I'm a bit hesitant to enable it without proper benchmarking first.

chuckatkins: > As far as the changes with CUDA, I'm quite happy to see changes for using the `cmake_cuda_architecture` along with setting the new policy for CMake, these are well-documented features and I assume the upstream has tests for them to keep the behaviour stable.

I had to make a slight change to not use `cmake_dependent_option` for it since apparently it forces the value into a `BOOL` which is obviously not what we want here.


> > giving valid configurations for lto of host-only, host and device, or neither.
> 
> I'm OK with the host only option if we can leave out the un-documented behavior of CMake. We can open an issue to keep track of enabling CUDA version. I was told that separable compilation has performance implication on the final code (for the worse in some cases), so I'm a bit hesitant to enable it without proper benchmarking first.

I updated the LTO commit so the three different options work accordingly.  In general the nvcc docs show a runtime performance hit when enabling separable compilation because the optimization passes don't look across translation units.  But the way it's implemented here is to only enable separable compilation with device LTO, which is required for it to function.  So without device LTO enabled separable compilation is still turned off.

chuckatkins:
@trivialfis I think this is good to go now

trivialfis:
Let me take another look tomorrow, thank you for the patience!

CMakeLists.txt
```diff
@@ -7,6 +7,8 @@ cmake_policy(SET CMP0079 NEW)
 cmake_policy(SET CMP0076 NEW)
 set(CMAKE_POLICY_DEFAULT_CMP0063 NEW)
 cmake_policy(SET CMP0063 NEW)
+set(CMAKE_POLICY_DEFAULT_CMP0069 NEW)
```
robertmaynard, line 10: Lets just yank all these chuck, as we have a  `cmake_minimum_required` of 3.18

CMakeLists.txt
```diff
@@ -171,10 +192,22 @@ if(USE_CUDA)
   set(CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER})
   message(STATUS "Configured CUDA host compiler: ${CMAKE_CUDA_HOST_COMPILER}")
 
+  if(POLICY CMP0104)
+    cmake_policy(SET CMP0104 NEW)
```
robertmaynard, line 196: Is there a code path I am not seeing where we aren't getting 3.18 as our minimum cmake version?

CMakeLists.txt
```diff
@@ -171,10 +192,22 @@ if(USE_CUDA)
...
+  endif()
+
+  if(GPU_COMPUTE_VER OR NOT CMAKE_CUDA_ARCHITECTURES)
```
robertmaynard, line 199: Since we always have cmake 3.18+ which has `CMAKE_CUDA_ARCHITECTURES` why don't we transform the value in `GPU_COMPUTE_VER` to `CMAKE_CUDA_ARCHITECTURES`?

Roughly the check would be:
```cmake
if(DEFINED GPU_COMPUTE_VER  AND NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
  transform....
endif()

File: cmake/Utils.cmake [hunk 6]
```diff
@@ -85,6 +85,10 @@ endfunction()
 # Generate nvcc compiler flags given a list of architectures
 # Also generates PTX for the most recent architecture for forwards compatibility
 function(format_gencode_flags flags out)
+  if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.18" AND CMAKE_CUDA_ARCHITECTURES)
```
robertmaynard, line 88:
```suggestion
  if(DEFINED CMAKE_CUDA_ARCHITECTURES)
```

cmake/Utils.cmake
```diff
@@ -104,7 +108,6 @@ function(format_gencode_flags flags out)
   endif()
 
   if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.18")
-    cmake_policy(SET CMP0104 NEW)
```
robertmaynard, line 107: Nit: Can we just yank the conditional here since we can rely on 3.18?

robertmaynard (changes requested):

CMakeLists.txt
robertmaynard: Awesome, thanks

CMakeLists.txt
chuckatkins: I was going for minimal change but that's a better end result anyways, especially after stripping out the `< 3.18` `else` branches

CMakeLists.txt
```diff
@@ -168,15 +185,24 @@ endif()
 if(USE_CUDA)
   set(USE_OPENMP ON CACHE BOOL "CUDA requires OpenMP" FORCE)
   # `export CXX=' is ignored by CMake CUDA.
-  set(CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER})
+  set(CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER} CACHE FILEPATH
+    "The compiler executable to use when compiling host code for CUDA or HIP language files.")
+  mark_as_advanced(CMAKE_CUDA_HOST_COMPILER)
   message(STATUS "Configured CUDA host compiler: ${CMAKE_CUDA_HOST_COMPILER}")
 
+  set(CMAKE_CUDA_RUNTIME_LIBRARY Static CACHE STRING
```
robertmaynard, line 193: Don't we just want to specify a value for this after `enable_language(CUDA)` or is this fixing some bug I am not aware of?

chuckatkins: It's currently explicitly being set as `Static` directly in the target property so making it a cache variable defaulting to `Static` preserves the existing behavior but lets the user pass in `-DCMAKE_CUDA_RUNTIME_LIBRARY=Shared` if they want to (we do :-) ). If it waits until after `enable_language` you wouldn't be able to tell if the value being used is user specified (which we want to keep) or a detected default (which we want to override with `Static` to maintain current behavior). At least I think so?

robertmaynard: You should be able to remove the cache variable entirely and just use `-DCMAKE_CUDA_RUNTIME_LIBRARY=Shared` . The internal default that cmake computes will go into a different variable and should be used only if no user value for `CMAKE_CUDA_RUNTIME_LIBRARY` exists.

robertmaynard: I now see where xgboost was explicitly using the per property var., I would go with:
```
if(NOT DEFINED CMAKE_CUDA_RUNTIME_LIBRARY)
  set(CMAKE_CUDA_RUNTIME_LIBRARY Static)
endif()
```
That way we don't influence other projects that consume xgboost via a source build ( FetchContent, CPM, etc ) with a cache variable.

The `CMAKE_CUDA_HOST_COMPILER` is a different since it is used during compiler detection, and therefore needs to be consistent across all targets. I am okay with the current approach for that.

robertmaynard (approved):
Looks great!

trivialfis:
Hi, could you please take a look into the errors in the CI?

chuckatkins:
> Hi, could you please take a look into the errors in the CI?

Looks like `tests/ci_build/prune_libnccl.sh` is using some of the modified bits that needed to be changed.  Should be fixed now.

chuckatkins:
> > Hi, could you please take a look into the errors in the CI?
> 
> Looks like `tests/ci_build/prune_libnccl.sh` is using some of the modified bits that needed to be changed. Should be fixed now.

@trivialfis @robertmaynard coincidentally, in fixing this I think this might have come a cross a bug in `nvprune` and it's argumet parsing.  Initially the changes to compute `CMAKE_CUDA_ARCHITECTURES` would generate something like `60-real;70-real;80` to generate cubin for 60, 70, and 80 and ptx for 80.  CMake translates this to `nvcc` flags `--generate-code=arch=compute_60,code=[sm_60] --generate-code=arch=compute_70,code=[sm_70] --generate-code=arch=compute_80,code=[compute_80,sm_80]`.  According to `nvprune --help` it's supposed to accept the same `--generate-code` code flags as `nvcc` but it chokes on the last `--generate-code=arch=compute_80,code=[compute_80,sm_80]` with multiple elements in the `code=` portion.  Adjusting the CMake code to instead compute `60-real;70-real;80-real;80-virtual` changes the `nvcc` flags for 80 to `--generate-code=arch=compute_80,code=[sm_80] --generate-code=arch=compute_80,code=[compute_80]` which is effectively the same thing but parsed by `nvprune` without issue.

robertmaynard:
> > > Hi, could you please take a look into the errors in the CI?
> > 
> > 
> > Looks like `tests/ci_build/prune_libnccl.sh` is using some of the modified bits that needed to be changed. Should be fixed now.
> 
> @trivialfis @robertmaynard coincidentally, in fixing this I think this might have come a cross a bug in `nvprune` and it's argumet parsing. Initially the changes to compute `CMAKE_CUDA_ARCHITECTURES` would generate something like `60-real;70-real;80` to generate cubin for 60, 70, and 80 and ptx for 80. CMake translates this to `nvcc` flags `--generate-code=arch=compute_60,code=[sm_60] --generate-code=arch=compute_70,code=[sm_70] --generate-code=arch=compute_80,code=[compute_80,sm_80]`. According to `nvprune --help` it's supposed to accept the same `--generate-code` code flags as `nvcc` but it chokes on the last `--generate-code=arch=compute_80,code=[compute_80,sm_80]` with multiple elements in the `code=` portion. Adjusting the CMake code to instead compute `60-real;70-real;80-real;80-virtual` changes the `nvcc` flags for 80 to `--generate-code=arch=compute_80,code=[sm_80] --generate-code=arch=compute_80,code=[compute_80]` which is effectively the same thing but parsed by `nvprune` without issue.

I will file a bug on nvprune about this, it does look like a regression of support

trivialfis (approved):
Thank you for your excellent work on enabling LTO and removing the old GPU arch!

PR #9685, opened 2023-10-17
More in-memory input support for column split

rongou:
- test loading from uri
- test pandas
- catch exceptions when running worker functions in a thread
- reformatted `test_dmatrix.py` and `test_with_pandas.py` with `black`

### DIFF
--- a/python-package/xgboost/testing/__init__.py
+++ b/python-package/xgboost/testing/__init__.py
@@ -950,7 +950,7 @@ def run_with_rabit(world_size: int, test_fn: Callable) -> None:
         try:
             with xgb.collective.CommunicatorContext(**rabit_env):
                 test_fn()
-        except Exception as e:
+        except Exception as e:  # pylint: disable=broad-except
             exception_queue.put(e)
 
     tracker = RabitTracker(host_ip="127.0.0.1", n_workers=world_size)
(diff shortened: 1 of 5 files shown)


python-package/xgboost/testing/__init__.py
```diff
@@ -943,12 +944,17 @@ def project_root(path: str) -> str:
 
 
 def run_with_rabit(world_size: int, test_fn: Callable) -> None:
-    tracker = RabitTracker(host_ip="127.0.0.1", n_workers=world_size)
-    tracker.start(world_size)
+    exception_queue: queue.Queue = queue.Queue()
```
trivialfis, line 947: Do we need a lock here? Or GIL is handling it?

tests/python/test_with_pandas.py
```diff
@@ -358,3 +416,235 @@ def test_pyarrow_type(self, DMatrixT: Type[xgb.DMatrix]) -> None:
             if y is not None:
                 np.testing.assert_allclose(m_orig.get_label(), m_etype.get_label())
                 np.testing.assert_allclose(m_etype.get_label(), y.values)
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="does not run on windows")
+class TestPandasColumnSplit:
+    @staticmethod
+    def verify_pandas():
```
trivialfis, line 424: Any chance we can reuse the pandas test instead of copying the code?

rongou: Done.

trivialfis (approved):

PR #9687, opened 2023-10-18
Fix libpath logic for Windows

hcho3:
Ref: https://github.com/conda-forge/xgboost-feedstock/pull/132#issuecomment-1767719122

### DIFF
--- a/python-package/xgboost/libpath.py
+++ b/python-package/xgboost/libpath.py
@@ -31,16 +31,15 @@ def find_lib_path() -> List[str]:
     ]
 
     if sys.platform == "win32":
-        if platform.architecture()[0] == "64bit":
-            dll_path.append(os.path.join(curr_path, "../../windows/x64/Release/"))
-            # hack for pip installation when copy all parent source
-            # directory here
-            dll_path.append(os.path.join(curr_path, "./windows/x64/Release/"))
-        else:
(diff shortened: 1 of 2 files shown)


python-package/xgboost/libpath.py
```diff
@@ -31,16 +31,15 @@ def find_lib_path() -> List[str]:
...
-        if platform.architecture()[0] == "64bit":
-            dll_path.append(os.path.join(curr_path, "../../windows/x64/Release/"))
-            # hack for pip installation when copy all parent source
-            # directory here
-            dll_path.append(os.path.join(curr_path, "./windows/x64/Release/"))
-        else:
-            dll_path.append(os.path.join(curr_path, "../../windows/Release/"))
-            # hack for pip installation when copy all parent source
-            # directory here
-            dll_path.append(os.path.join(curr_path, "./windows/Release/"))
```
trivialfis, line 43: Are these old paths not useful anymore? (like editable installation)

hcho3: I don't think so. I use Windows as a daily driver and I didn't need these paths.

We can add them back when they are needed.

trivialfis (approved):

hcho3:
I will merge this after verifying that this patch fixes the Conda build in https://github.com/conda-forge/xgboost-feedstock/pull/132

PR #9691, opened 2023-10-18
Add 'sycl' devices to the context

razdoburdin:
Hi,
I want to restart discussion about perspectives of adding `sycl `support as plugin for XGBoost. 

The code of the plugin itself is quite complicated and `sycl `specific. So, I think the best way in the current step is adding a link to intel repo., where we can maintain the plugin by ourself. What is your opinion?

Even in case of link to the intel repo. Supporting of new devices will require some changes in the main part of the code, especially in the content of `Context `class. The current PR add `sycl` devices to the `Contex `object.
Supported values of the 'device' parameter are:
`device = "sycl"` -- use default `sycl` device
`device = "sycl:cpu"` -- use default `sycl` CPU
`device = "sycl:gpu"` -- use default `sycl` GPU
`device = "sycl:gpu:{N}"` -- use `sycl` GPU with index `{N}`.

### DIFF
--- a/include/xgboost/context.h
+++ b/include/xgboost/context.h
@@ -22,6 +22,9 @@ struct CUDAContext;
 struct DeviceSym {
   static auto constexpr CPU() { return "cpu"; }
   static auto constexpr CUDA() { return "cuda"; }
+  static auto constexpr SYCL_default() { return "sycl"; }
+  static auto constexpr SYCL_CPU() { return "sycl:cpu"; }
+  static auto constexpr SYCL_GPU() { return "sycl:gpu"; }
 };
 
 /**
@@ -33,12 +36,19 @@ struct DeviceOrd {
   static bst_d_ordinal_t constexpr CPUOrdinal() { return -1; }
   static bst_d_ordinal_t constexpr InvalidOrdinal() { return -2; }
 
-  enum Type : std::int16_t { kCPU = 0, kCUDA = 1 } device{kCPU};
-  // CUDA device ordinal.
+  enum Type : std::int16_t { kCPU = 0, kCUDA = 1,
+                             kSyclDefault = 2, kSyclCPU = 3, kSyclGPU = 4} device{kCPU};
+  // CUDA or Sycl device ordinal.
   bst_d_ordinal_t ordinal{CPUOrdinal()};
 
   [[nodiscard]] bool IsCUDA() const { return device == kCUDA; }
   [[nodiscard]] bool IsCPU() const { return device == kCPU; }
+  [[nodiscard]] bool IsSyclDefault() const { return device == kSyclDefault; }
+  [[nodiscard]] bool IsSyclCPU() const { return device == kSyclCPU; }
+  [[nodiscard]] bool IsSyclGPU() const { return device == kSyclGPU; }
+  [[nodiscard]] bool IsSycl() const { return (IsSyclDefault() ||
+                                              IsSyclCPU() ||
+                                              IsSyclGPU()); }
 
   constexpr DeviceOrd() = default;
   constexpr DeviceOrd(Type type, bst_d_ordinal_t ord) : device{type}, ordinal{ord} {}
@@ -60,6 +70,31 @@ struct DeviceOrd {
   [[nodiscard]] static constexpr auto CUDA(bst_d_ordinal_t ordinal) {
     return DeviceOrd{kCUDA, ordinal};
   }
+  /**
+   * @brief Constructor for SYCL.
+   *
+   * @param ordinal SYCL device ordinal.
+   */
+  [[nodiscard]] constexpr static auto SYCL_default(bst_d_ordinal_t ordinal = -1) {
+    return DeviceOrd{kSyclDefault, ordinal};
+  }
+  /**
+   * @brief Constructor for SYCL CPU.
+   *
+   * @param ordinal SYCL CPU device ordinal.
+   */
+  [[nodiscard]] constexpr static auto SYCL_CPU(bst_d_ordinal_t ordinal = -1) {
+    return DeviceOrd{kSyclCPU, ordinal};
+  }
+
+  /**
+   * @brief Constructor for SYCL GPU.
+   *
+   * @param ordinal SYCL GPU device ordinal.
+   */
+  [[nodiscard]] constexpr static auto SYCL_GPU(bst_d_ordinal_t ordinal = -1) {
+    return DeviceOrd{kSyclGPU, ordinal};
+  }
 
   [[nodiscard]] bool operator==(DeviceOrd const& that) const {
     return device == that.device && ordinal == that.ordinal;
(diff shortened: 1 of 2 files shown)


trivialfis:
Hi @razdoburdin , good to see you again, and thank you for continuing the work on sycl support!

Would be great if you could share some more info:

- What's sync:cpu? How is it different from the OpenMP CPU, feature-wise and performance-wise?
- How to add a plugin in the Intel repo? Is it a static plugin (intel-variant of xgboost in conda-forge, for instance) or a dynamic plugin (dlopen).
- As for the plugin, I think it's a reasonable choice to have a separate repo, or in the `xgboost/plugin/sycl` directory. Either way, we need to have some CI infra in XGBoost so that we don't break the code or the logic to link the Intel repo immediately.

src/context.cc
```diff
@@ -94,29 +94,45 @@ DeviceOrd CUDAOrdinal(DeviceOrd device, bool) {
...
   if (!std::regex_match(input, pattern)) {
     fatal();
   }
 #endif  // defined(__MINGW32__)
 
   // handle alias
-  std::string s_device = std::regex_replace(input, std::regex{"gpu"}, DeviceSym::CUDA());
+  std::string s_device = input;
+  if (!std::regex_match(s_device, std::regex("sycl(:cpu|:gpu)?(:-1|:[0-9]+)?")))
+    s_device = std::regex_replace(s_device, std::regex{"gpu"}, DeviceSym::CUDA());
```
napetrov, line 123: Shouldn't this mean 
User specify any GPU on the system -> xgboost check available devices and pickup GPU device that i can execute on. 
I.e. in case there is only CUDA device - it will be used
In case there is only SYCL device - it will be used
The question what to do in case both are available.. Might be warning about non deterministic selection or might be even error. Not sure how realistic this case is

razdoburdin: I guess in the future it should work like you have described. But let's postpone this fix for future.

napetrov:
Would add couple cents. 
On sycl:cpu - there are 2 parts here. 
1st - is how well the SYCL would work on CPU. Overall this is not ideal and would require level_zero support and not just opencl runtime as done currently. 
2nd - this would require sycl kernels tuning for CPUs 

So we will not be working on 2nd until 1st would get improved.  So we should keep this as an option to run sycl code on CPUs, but not recommend this to users at this point.  But this might be usefull for validation/development purposes at this point 

On CI - we in meantime would work on some form of GPU systems availability via https://cloud.intel.com/ so code also can be tested on Intel GPUs

trivialfis:
> Would you like to have a call next week to discuss the integration stages?

That would be great!

include/xgboost/context.h
```diff
@@ -74,6 +109,12 @@ struct DeviceOrd {
         return DeviceSym::CPU();
       case DeviceOrd::kCUDA:
         return DeviceSym::CUDA() + (':' + std::to_string(ordinal));
+      case DeviceOrd::kSyclDefault:
```
trivialfis, line 112: What is "sync:{ordinal}"?

include/xgboost/context.h
```diff
@@ -74,6 +109,12 @@ struct DeviceOrd {
...
+        return DeviceSym::SYCL_default() + (':' + std::to_string(ordinal));
+      case DeviceOrd::kSyclCPU:
+        return DeviceSym::SYCL_CPU() + (':' + std::to_string(ordinal));
```
trivialfis, line 115: What is "sycl:cpu:ordinal"

include/xgboost/context.h
```diff
@@ -60,6 +70,31 @@ struct DeviceOrd {
   [[nodiscard]] static constexpr auto CUDA(bst_d_ordinal_t ordinal) {
     return DeviceOrd{kCUDA, ordinal};
   }
+  /**
+   * @brief Constructor for SYCL.
+   *
+   * @param ordinal SYCL device ordinal.
+   */
+  [[nodiscard]] constexpr static auto SYCL_default(bst_d_ordinal_t ordinal = -1) {
```
trivialfis, line 78: Please follow the function naming convention "SYCLDefault"/"SyclDefault"

include/xgboost/context.h
razdoburdin: It means using CPU with the specific index for multi CPU systems.

include/xgboost/context.h
razdoburdin: It means using the SYCL device with index `{ordinal}`. It can be CPU or GPU, depending on user's system settings.

include/xgboost/context.h
```diff
@@ -22,6 +22,9 @@ struct CUDAContext;
 struct DeviceSym {
   static auto constexpr CPU() { return "cpu"; }
   static auto constexpr CUDA() { return "cuda"; }
+  static auto constexpr SYCL_default() { return "sycl"; }
+  static auto constexpr SYCL_CPU() { return "sycl:cpu"; }
+  static auto constexpr SYCL_GPU() { return "sycl:gpu"; }
```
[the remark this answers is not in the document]
razdoburdin: SYCL try to run on GPU. If GPU isn't available for some reason, it launches on CPU.

include/xgboost/context.h
razdoburdin: done

include/xgboost/context.h
trivialfis: multi-socket or multi-core?

razdoburdin: multi-socket

src/context.cc
```diff
@@ -94,29 +94,45 @@ DeviceOrd CUDAOrdinal(DeviceOrd device, bool) {
...
 - cuda
-- cuda:<device ordinal>  # e.g. cuda:0
+- cuda:<device ordinal>            # e.g. cuda:0
 - gpu
-- gpu:<device ordinal>   # e.g. gpu:0
+- gpu:<device ordinal>             # e.g. gpu:0
+- sycl
+- sycl:<device ordinal>            # e.g. sycl:0
+- sycl:<cpu, gpu>
+- sycl:<cpu, gpu>:<device ordinal> # e.g. sycl:gpu:0
```
trivialfis, line 103: Let's remove the error message change for now and get it back when there's a public ready feature that can be enabled. At this point, the message is only going to confuse users.

trivialfis (commented):
Overall looks good to me, please keep the change internal for now.

src/context.cc
razdoburdin: done

trivialfis:
Please fix the tidy error

razdoburdin:
> Please fix the tidy error

fixed

trivialfis:
@razdoburdin Apologies for my mistake on not catching it, could you please help fix the CI by conditioning the regex match based on __ming32__ ? It hangs on regex match.

https://github.com/dmlc/xgboost/actions/runs/6662511294/job/18106961526
https://github.com/dmlc/xgboost/actions/runs/6655456686/job/18085766529

razdoburdin:
> @razdoburdin Apologies for my mistake of not catching it, could you please help fix the CI by conditioning the regex match based on `__MING32__` ? It hangs on regex match.
> 
got it, will fix asap

PR #9695, opened 2023-10-19
[coll] Define interface for bridging.

trivialfis:
* Define the basic interface that will shared by nccl, federated and native.


trivialfis:
cc @rongou .

rongou (approved):

PR #9698, opened 2023-10-22
[jvm-packages] Remove hard dependency on libjvm

### LINKED ISSUE #9697: [jvm-packages] Replace JNI with Java Native Access
**Problem**
Currently, XGBoost4J uses Java Native Interface (JNI) to call the native code in `libxgboost4j.so`. JNI has served our needs so far, but its use causes a tight coupling with the particular JDK that XGBoost4J was built with.

For example, when building XGBoost4J on Mac mini M1, I used Azul JDK 8 (*):
```
$ export JAVA_HOME=/Library/Java/JavaVirtualMachines/zulu-8.jdk/Contents/Home/
$ mkdir build
$ cd build
$ cmake .. -GNinja -DJVM_BINDINGS=ON -DUSE_OPENMP=OFF
```
This leads to a binary that has a hard dependency on `libjvm.dylib`:
```
$ otool -L libxgboost4j.dylib

libxgboost4j.dylib:
        @rpath/libxgboost4j.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libjvm.dylib (compatibility version 1.0.0, current version 1.0.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1500.65.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.100.3)
```

As a result, when I try to load the binary on another MacOS machine without Azul JDK 8, the import fails with an error:
```
$ java -version   # This machine has OpenJDK 21
openjdk version "21" 2023-09-19
OpenJDK Runtime Environment Homebrew (build 21)
$ python -c "import ctypes; ctypes.cdll.LoadLibrary('./libxgboost4j.dylib')"
  File "<string>", line 1, in <module>
  File "/Users/phcho/mambaforge/lib/python3.10/ctypes/__init__.py", line 452, in LoadLibrary
    return self._dlltype(name)
  File "/Users/phcho/mambaforge/lib/python3.10/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: dlopen(./libxgboost4j.dylib, 0x0006): Library not loaded: @rpath/libjvm.dylib
Reason: tried: '/Library/Java/JavaVirtualMachines/zulu-8.jdk/Contents/Home/jre/lib/server/libjvm.dylib' (no such file)  ...
```
The error message suggests a tight coupling of XGBoost4J with a particular JDK. **The coupling makes it difficult for us to distribute pre-built binaries targeting the MacOS Apple Silicon platform.** Users may have installed Azul JDK 8 (whose support is due to end in 2026), OpenJDK 11, or OpenJDK 21 (the latest version).

**Proposal**. Replace JNI with [Java Native Access (JNA)](https://github.com/java-native-access/jna). JNA works like `ctypes` in Python, so we would be able to call functions directly from `libxgboost.so`. JNA is used by many java software such as IntelliJ IDEA, and it is tested with multiple versions of JDK and platforms. The removal of JNI stub code will also reduce the boilerplate in the code base.

**Alternatives**. We can tell users to install a particular JDK as we have done so far. So for example, users of MacOS Apple Silicon will be asked to install Azul JDK 8.

@trivialfis  @wbo4958 @WeichenXu123 @Craigacp @drnguyen @terrytangyuan Your feedback will be highly appreciated.

(*) OpenJDK 8 supports Intel only and is not available for Apple Silicon.

hcho3:
Closes #9697

Follow the footstep of https://github.com/microsoft/onnxruntime/pull/2873 to remove the hard link dependency on libjvm. The JVM process will load the correct sets of libraries prior to loading libxgboost4j, so we don't need to explicitly link Java libraries again.

With this fix, XGBoost4J will no longer be tightly coupled with any particular JDK.

Before:
```
$ otool -L libxgboost4j.dylib
libxgboost4j.dylib:
        @rpath/libxgboost4j.dylib (compatibility version 0.0.0, current version 0.0.0)
        /opt/homebrew/opt/openjdk/libexec/openjdk.jdk/Contents/Home/lib/server/libjvm.dylib (compatibility version 1.0.0, current version 1.0.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1500.65.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.100.3)
```

After:
```
$ otool -L libxgboost4j.dylib
libxgboost4j.dylib:
        @rpath/libxgboost4j.dylib (compatibility version 0.0.0, current version 0.0.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1500.65.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.100.3)
```

### DIFF
--- a/jvm-packages/CMakeLists.txt
+++ b/jvm-packages/CMakeLists.txt
@@ -25,4 +25,3 @@ target_include_directories(xgboost4j
   ${PROJECT_SOURCE_DIR}/rabit/include)
 
 set_output_directory(xgboost4j ${PROJECT_SOURCE_DIR}/lib)
-target_link_libraries(xgboost4j PRIVATE ${JAVA_JVM_LIBRARY})


terrytangyuan (approved):
LGTM!

PR #9702, opened 2023-10-23
Serialize expand entry for allgather.

trivialfis:


trivialfis:
cc @rongou

rongou (commented):
Overall looks good, but aren't you worried about the size blowup with JSON? Especially with federated learning when the data may need to be shipped over relatively slow internet connections.

trivialfis:
If the performance becomes a problem, we can shrink the size of the JSON object to plain binary by removing object key strings and packing everything into an array, XGB can work with binary JSON.

trivialfis:
On the other hand, I think GPU cat is ot yet supported, correct me if I'm wrong. I couldn't find the synchronization for GPU cat bits.

rongou:
> If the performance becomes a problem, we can shrink the size of the JSON object to plain binary by removing object key strings and packing everything into an array, XGB can work with binary JSON.

Fair enough.

> On the other hand, I think GPU cat is ot yet supported, correct me if I'm wrong. I couldn't find the synchronization for GPU cat bits.

I'm not sure. We do have some tests for category features, maybe they are not exhaustive enough.

rongou (approved):

PR #9724, opened 2023-10-25
[pyspark]  unify the way for determining whether  runs on the GPU.

wbo4958:
The current method of determining whether to run the training or transformation process on the GPU is based on the settings "use_gpu" or "device=cuda/gpu". However, if users simply set "tree_method=gpu_hist", XGBoost will not utilize the GPU. 

This PR aims to address this problem and includes some minor code refactoring as well.

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -347,7 +347,7 @@ class _SparkXGBParams(
     def _validate_gpu_params(self) -> None:
         """Validate the gpu parameters and gpu configurations"""
 
-        if use_cuda(self.getOrDefault(self.device)) or self.getOrDefault(self.use_gpu):
+        if self._run_on_gpu():
             ss = _get_spark_session()
             sc = ss.sparkContext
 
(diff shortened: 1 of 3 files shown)


wbo4958:
@trivialfis could you help to review it.

python-package/xgboost/spark/core.py
```diff
@@ -473,6 +473,15 @@ def _validate_params(self) -> None:
...
         self._validate_gpu_params()
 
+    def _run_on_gpu(self) -> bool:
+        """If train or transform on the gpu according to the parameters"""
+
+        return (
+            use_cuda(self.getOrDefault(self.device))
+            or self.getOrDefault(self.use_gpu)
+            or self.getOrDefault(self.getParam("tree_method")) == "gpu_hist"
+        )
```
trivialfis, line 483: Did we support the `gpu_hist` with Pyspark in 1.7? I thought the `use_gpu` was the only way to get GPU support with PySpark even before 2.0.

wbo4958: Yes, pyspark always supports gpu_hist

wbo4958: Do you want gpu_hist is not supported in pyspark?

trivialfis (approved):

PR #9726, opened 2023-10-26
[coll] Add nccl.

trivialfis:
Different from the existing one for allgather-v, this uses ring-based custom implementation instead of rotating broadcasts. We can use a similar path for bit-wise allreduce, I will leave that as a future to-do item.

### DIFF
--- a/src/collective/coll.cu
+++ b/src/collective/coll.cu
@@ -1,0 +1,203 @@
+/**
+ * Copyright 2023, XGBoost Contributors
+ */
+#if defined(XGBOOST_USE_NCCL)
+#include <cstdint>  // for int8_t, int64_t
+
+#include "../common/cuda_context.cuh"
+#include "../common/device_helpers.cuh"
+#include "../data/array_interface.h"
+#include "allgather.h"  // for AllgatherVOffset
+#include "coll.cuh"
+#include "comm.cuh"
+#include "nccl.h"
+#include "xgboost/collective/result.h"  // for Result
+#include "xgboost/span.h"               // for Span
+
+namespace xgboost::collective {
+Coll* Coll::MakeCUDAVar() { return new NCCLColl{}; }
+
+NCCLColl::~NCCLColl() = default;
+namespace {
+Result GetNCCLResult(ncclResult_t code) {
+  if (code == ncclSuccess) {
+    return Success();
+  }
+
+  std::stringstream ss;
+  ss << "NCCL failure: " << ncclGetErrorString(code) << ".";
+  if (code == ncclUnhandledCudaError) {
+    // nccl usually preserves the last error so we can get more details.
+    auto err = cudaPeekAtLastError();
+    ss << "  CUDA error: " << thrust::system_error(err, thrust::cuda_category()).what() << "\n";
+  } else if (code == ncclSystemError) {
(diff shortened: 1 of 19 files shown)


trivialfis:
cc @rongou .

src/collective/coll.cu
```diff
@@ -0,0 +1,203 @@
...
+  auto nccl = dynamic_cast<NCCLComm const*>(&comm);
+  CHECK(nccl);
+  auto send = data.subspan(comm.Rank() * size, size);
+  return Success() << [&] {
+    return GetNCCLResult(
+        ncclAllGather(send.data(), data.data(), size, ncclInt8, nccl->Handle(), nccl->Stream()));
+  } << [&] { return nccl->Block(); };
+}
+
+[[nodiscard]] Result NCCLColl::AllgatherV(Comm const& comm, common::Span<std::int8_t const> data,
```
rongou, line 185: I think the original implementation is based on this: https://arxiv.org/abs/1812.05964. Have you looked at the performance implications?

trivialfis: Not yet, it's quite difficult to set up a benchmark for the existing one since it doesn't support thread-based multi-gpu and doesn't have a Python interface.

trivialfis: I have implemented both algorithms for both CPU and GPU. For GPU indeed the bcast is faster for small-size communication. I think I will leave both implementations here for now and defer the formal benchmark.

rongou: Yeah for GPU the memcpy can be expensive.

src/collective/coll.cu
```diff
@@ -0,0 +1,250 @@
...
+                                          common::Span<std::int64_t const> sizes,
+                                          common::Span<std::int64_t> recv_segments,
+                                          common::Span<std::int8_t> recv, AllgatherVAlgo algo) {
+  auto nccl = dynamic_cast<NCCLComm const*>(&comm);
+  CHECK(nccl);
+  // get worker offset
+  detail::AllgatherVOffset(sizes, recv_segments);
+  // copy data
```
rongou, line 226: Is this still needed for the broadcast implementation?

trivialfis: Yes, we need to copy the data from send buffer to recv buffer

src/collective/coll.cu
trivialfis: Haven't looked through nsys, are talking about memcpy inside ncclSend/Recv?

src/collective/coll.cu
rongou: The original implementation doesn't have a copy, right? I think this is only needed for the ring implementation.

trivialfis: I can try to detect the pointer range and see whether send and recv overlap, then use inplace when possible. Is this what you mean?

rongou: In the broadcast implementation, there are two separate parameters `data` and `recv`, so data is broadcast into recv, no need to copy first; while in the ring implementation there is only `recv`. I think all you need to do here is to move the memcpy under the `kRing` case.

trivialfis: Ah, got it, thank you for the explanation!

rongou (approved):

PR #9732, opened 2023-10-30
[coll] Federated comm.

trivialfis:
Implement communicator and server based on `Comm` and `Tracker`. The collective implementation will be part of the next PR.


trivialfis:
cc @rongou .

rongou (approved):

PR #9737, opened 2023-10-31
[CI] Enable gmock in gtest

hcho3:
@trivialfis You should be able to use `<gmock/gmock.h>` header now.


trivialfis (approved):
Did not expect it to be a one-liner. Thank you!

PR #9738, opened 2023-10-31
[coll] Add federated coll.

trivialfis:
- Define a new data type, the proto file is copied for now.
- Merge client and communicator into `FederatedColl`.
- Define CUDA variant.
- Migrate tests for CPU, add tests for CUDA.

The new implementation no longer uses std::string.


trivialfis:
cc @rongou .

rongou (approved):

PR #9740, opened 2023-10-31
[R] Avoid in-place modifications to inputs when plotting importances

david-cortes:
This PR addresses two issues:
* Calling `xgb.plot.importance` performs in-place modifications to the input it receives, by adding a new column `Importance`.
* It currently states that it will
> aggregate, just in case when the values were not yet summed up by feature

but this is not the case: while it aggregates, if there is a duplicated entry (Feature), the output will still have it more than once, only with the importances summed, and the plot will have two bars for the same feature.

I'm not entirely sure if this is the most correct behavior though - perhaps a better thing to do would be to not atempt to aggregate when there are duplicated features, and just plot them like they are in the input.

The first behavior (modifying the input it receives in-place) is quite bothersome however.

### DIFF
--- a/R-package/R/xgb.plot.importance.R
+++ b/R-package/R/xgb.plot.importance.R
@@ -87,7 +87,13 @@ xgb.plot.importance <- function(importance_matrix = NULL, top_n = NULL, measure 
   }
 
   # also aggregate, just in case when the values were not yet summed up by feature
-  importance_matrix <- importance_matrix[, Importance := sum(get(measure)), by = Feature]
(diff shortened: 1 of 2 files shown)


trivialfis (approved):
Thank you for removing the inplace operation. This is more in line with the R tradition of immutable variables.

> I'm not entirely sure if this is the most correct behavior though

For my preference, I would probably just raise an error saying that features should be unique. I don't have strong opinion on this, feel free to change it if it helps remove confusion.

Will merge once buildkite is green. Please let me know if there are further changes.

PR #9745, opened 2023-11-01
[coll] Improvements and fixes for tracker and allreduce.

trivialfis:
- Allow tracker to wait.
- Fix allreduce type cast
- Return args from federated tracker.


trivialfis:
cc @rongou .

rongou (approved):

PR #9748, opened 2023-11-01
[CI] Improve CI for Mac M1

hcho3:
* Don't create Conda env
* Build XGBoost4J lib


CMakeLists.txt
```diff
@@ -32,6 +32,9 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
     message(FATAL_ERROR "Need Clang 9.0 or newer to build XGBoost")
   endif()
 endif()
+if(APPLE)
+  set(CMAKE_MACOSX_DEPLOYMENT_TARGET 10.15)
```
trivialfis, line 36: Is it possible to use this as an independent command line argument in CI instead of hardcoding it into the source code?

hcho3: Sure, let me create a PR.

PR #9755, opened 2023-11-03
[doc] Draft for language binding consistency. [skip ci]

trivialfis:


trivialfis:
Hi all, this is a draft for documenting some consistency issues that hopefully we can address in the future. Please help review and make suggestions.

cc @ExpandingMan @david-cortes

david-cortes:
@trivialfis Thanks for the info. Also wondering about the recommended way of handling the `.cv` functions and details like creating one big `DMatrix` and slicing it vs. creating separate ones out of sliced data. I see for example that the R interface does a lot of things outside of the core C++ library.

trivialfis:
@david-cortes IMHO, XGBoost doesn't need to have a CV function in its interface. The existing implementation pre-dates my involvement in XGBoost. To me, the CV function should be part of a more general hyper-parameter optimization library or a model evaluation library instead of XGBoost. My opinion is based on the fact that cross-validation is a general statistic technique instead of a gradient-boosting-specific technique.

Ideal case aside, if we were to have a specialized feature version for XGB, (like working with survival data, custom early stopping strategy, etc), I think we can directly work with input data like `data.table` and `np.ndarray` instead of  `DMatrix`.

trivialfis:
> I think we can directly work with input data like data.table and np.ndarray instead of DMatrix. 

That said, it's totally reasonable to have a different project than XGBoost, or upstream some changes to existing projects like scikit-learn.

david-cortes:
(Note: comment is unrelated to the PR in discussion)

Generally, it makes sense to leave cross-validation to a dedicated framework, but there are cases where a dedicated `cv` function can exploit optimizations specific to the algorithm that a higher-level interface wouldn't.

I'm not familiar with what xgboost does internally in each situation, but for example in catboost, binning data in the histogram method (especially when there are categorical features) can take a lot of time, and an efficient `cv` routine can be exploited to do this only once instead of once per data fold.

trivialfis:
> Generally, it makes sense to leave cross-validation to a dedicated framework, but there are cases where a dedicated cv function can exploit optimizations specific to the algorithm that a higher-level interface wouldn't.

That's a very good point! We don't have this sort of optimization yet but can be done if necessary. Basically, one would obtain a `QuantileDMatrix` for the full dataset and use it as a reference for the others. However, I'm a bit confused, wouldn't this leak information about the validation data into training? Might not matter in practice, but seems like an odd choice.

david-cortes:
> > Generally, it makes sense to leave cross-validation to a dedicated framework, but there are cases where a dedicated cv function can exploit optimizations specific to the algorithm that a higher-level interface wouldn't.
> 
> That's a very good point! We don't have this sort of optimization yet but can be done if necessary. Basically, one would obtain a `QuantileDMatrix` for the full dataset and use it as a reference for the others. However, I'm a bit confused, wouldn't this leak information about the validation data into training? Might not matter in practice, but seems like an odd choice.
> 
> ```python
> Xy = QuantileDMatrix(X, y)  # <- contains the bining for the whole dataset.
> X_train, X_test, y_train, y_test = train_test_split(X, y)
> Xy_train = QuantileDMatrix(X_train, y_train, ref=Xy)
> Xy_test = QuantileDMatrix(X_test, y_test, ref=Xy)
> ```

Yes, there is some small data leakage when doing cross-validation with the same quantization for all folds, and it gives slightly different results than doing it the "correct" way, but as sample sizes increase, this difference shrinks.

Although I see that for example lightgbm authors want to [move away from this CV behavior](https://github.com/microsoft/LightGBM/issues/4319), so that's why I was asking whether there would be a recommendation for xgboost's CV function. Although I also think in the case of xgboost, there's less pre-processing that in e.g. catboost so the time and memory gains would be much smaller.

ExpandingMan:
The only part of the draft that I have significant concerns about is the default parameter values.  Some of the default values are problematic, in particular see [this issue](https://github.com/dmlc/XGBoost.jl/issues/88) in which the default value for number of threads can cause programs to silently hang if they exceed the number of Julia process threads.  I would not be at all surprised if that were not the only such issue (are there any parameters that reference an index? I can't think of any off the top of my head, but if so they will be different for 0 and 1-based languages).  I think it's reasonable to require that language bindings can always accept a null argument to reproduce the `libxgboost` defaults.  I also think it's reasonable to have a policy of preferring `libxgboost` defaults where they do not cause undesired side effects, but declaring it as a strict policy seems risky.

On a somewhat less important note, iteration number seems like it could present some concerns.  Currently, in the julia wrapper, it looks like we start with round `1` and don't translate it, which, if I'm reading your documentation corrrectly, sounds like it might actually be a bug.  This can of course be easily fixed, but I think on the Julia side we would prefer to keep iteration numbers 1-based for consistency with indexing.

Finally, I agree that the wrapper packages should be *separate* from additional general machine learning stuff like CV.  I've encountered a few people who claimed we should add it, but there's a handful of packages with CV functionality in Julia, and I'd imagine that the number of such packages in Python must be astronomical.  Early stopping, on the other hand, I don't see the same way, as it's much harder to provide a general interface for that works across different ML algorithms (arguably such a thing could not even be logically consistent).  I do think it would be a good idea for the core library to add early stopping if the developers are so inclined.

trivialfis:
> Some of the default values are problematic

We can always make exceptions. This is a guideline, not a policy, as shared in the first paragraph.

> On a somewhat less important note, iteration number seems like it could present some concerns.

Is it possible to translate it to 0-based under the hood?

> I do think it would be a good idea for the core library to add early stopping if the developers are so inclined.

This might not happen in the near future. Having ES inside the core would mean we will move all the logic of callback, custom objective, metrics into c++ core, which is quite challenging to say the least.

ExpandingMan:
> Is it possible to translate it to 0-based under the hood?

Yes, definitely, I was just a bit unsure whether that would comply with your proposed guidelines because of reporting and all.  If we can do the same thing we can do with array indices, I have no concern here.

> This might not happen in the near future. Having ES inside the core would mean we will move all the logic of callback, custom objective, and metrics into C++ core, which is quite challenging, to say the least.

Oh well.  I had a feeling it might be quite involved, but thought I'd mention it just in case.

doc/contrib/consistency.rst
```diff
@@ -0,0 +1,62 @@
...
+
+- If the field is small, we can save it as model attribute (which is a key-value structure). These attributes are ignored by all other language bindings and mostly an ad-hoc storage.
+
+- Lastly, we should use the UBJSON as the default output format when given a chance (not to be burdened by the old binary format).
+
+*********************
+Training Continuation
+*********************
+
+There are cases where we want to train a model based on the previous model, for boosting trees, it's either adding new trees or modifying the existing trees. This can be normal model update, error recovery, or other special cases we don't know of yet. When it happens, the training iteration should start from 0, not from the lasting boosted rounds of the model. 0 is a special iteration number, we perform some extra checks like whether the label is valid during that iteration. These checks can be expensive but necessary for eliminating silent errors. Keeping the iteration starts from zero allows us to perform these checks only once for each input data.
```
trivialfis, line 29:
```suggestion
There are cases where we want to train a model based on the previous model, for boosting trees, it's either adding new trees or modifying the existing trees. This can be normal model update, error recovery, or other special cases we don't know of yet. When it happens, the training iteration should start from 0, not from the last boosted rounds of the model. 0 is a special iteration number, we perform some extra checks like whether the label is valid during that iteration. These checks can be expensive but necessary for eliminating silent errors. Keeping the iteration starts from zero allows us to perform these checks only once for each input data.
```

doc/contrib/consistency.rst
```diff
@@ -0,0 +1,62 @@
...
+
+******************
+Default Parameters
+******************
+
+Like many other machine learning libraries, XGBoost all of them can either be inferred from the data or have default values. Bindings should not make copies of these default values and let the XGBoost core to decide. When the parameter key is not passed into the C++ core, XGBoost will pick the default accordingly. These default are not necessarily optimal, but they are there for consistency. If there's a new choice of default parameter, we can change it inside the core and it will be automatically propagated to all bindings. Given the same set of parameters and data, various bindings should stride to produce the same model. One exception is the `num_boost_rounds`, which exists only in high level bindings and has various alias like ``n_estimators``. It's default value is close to arbitrary at the moment, we haven't been able to get a good default yet.
```
trivialfis, line 50:
```suggestion
Like many other machine learning libraries, all parameters from XGBoost can either be inferred from the data or have default values. Bindings should not make copies of these default values and let the XGBoost core decide. When the parameter key is not passed into the C++ core, XGBoost will pick the default accordingly. These defaults are not necessarily optimal, but they are there for consistency. If there's a new choice of default parameter, we can change it inside the core and it will be automatically propagated to all bindings. Given the same set of parameters and data, various bindings should strive to produce the same model. One exception is the `num_boost_rounds`, which exists only in high-level bindings and has various alias like ``n_estimators``. Its default value is close to arbitrary at the moment, we haven't been able to get a good default yet.
```

doc/contrib/consistency.rst
```diff
@@ -0,0 +1,62 @@
...
...y a machine learning library providing boosting algorithm implementation. Some other implementations might perform some sort of data manipulation implicitly like deciding the coding of the data, transforming the data according to some heuristic before training. We prefer to keep these operations based on necessities instead of convenience to keep the scope of the project well defined. Whenever possible, we should leave these features to 3-party libraries and consider how a user can compose their pipeline. For instance, XGBoost itself should not perform ordinal encoding for categorical data, users will pick an encoder that fits their use cases (like out-of-core implementation, distributed implementation, known mapping, etc). If some transformations are decided to be part of the algorithm, we can have it inside the core instead of the language binding. Examples would be target-encoding or sketching the response variables. If we were to support them, we can have it inside the core implementation as part of the ML algorithm. This aligns with the same principals of default parameter, various bindings should provide similar (if not the same) results given same set of parameters and data.
```
trivialfis, line 62:
```suggestion
XGBoost is mostly a machine learning library providing boosting algorithm implementation. Some other implementations might perform some sort of data manipulation implicitly like deciding the coding of the data, and transforming the data according to some heuristic before training. We prefer to keep these operations based on necessities instead of convenience to keep the scope of the project well-defined. Whenever possible, we should leave these features to 3-party libraries and consider how a user can compose their pipeline. For instance, XGBoost itself should not perform ordinal encoding for categorical data, users will pick an encoder that fits their use cases (like out-of-core implementation, distributed implementation, known mapping, etc). If some transformations are decided to be part of the algorithm, we can have it inside the core instead of the language binding. Examples would be target-encoding or sketching the response variables. If we were to support them, we could have it inside the core implementation as part of the ML algorithm. This aligns with the same principles of default parameters, various bindings should provide similar (if not the same) results given the same set of parameters and data.
```

PR #9757, opened 2023-11-03
Change default metric for gamma regression to deviance.

trivialfis:
- Cleanup the gamma implementation.
- Use deviance instead since the objective is derived from deviance.

### DIFF
--- a/src/objective/regression_loss.h
+++ b/src/objective/regression_loss.h
@@ -13,9 +13,7 @@
 #include "xgboost/logging.h"
 #include "xgboost/task.h"  // ObjInfo
 
-namespace xgboost {
-namespace obj {
-// common regressions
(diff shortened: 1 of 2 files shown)


trivialfis:
We can perhaps get a better initial estimation.

RAMitchell (approved):

saima1704:
gamma <- function (preds, drain)
  labels <- getinfo(dtrain_sim, "label")
  preds <- exp(preds) 
    # Gradient (First Order)
  grad <- 1.0 - labels / preds
  # Hessian (Second Order)
  hess <- labels / preds
  return(list(grad = grad, hess = hess))
}
May I kindly ask if this custom loss function is equivalent to reg:gamma in r?

PR #9770, opened 2023-11-07
Use `unique_ptr` for cub allocator.

trivialfis:
Make sanitizer happy.


trivialfis:
Will be part of https://github.com/dmlc/xgboost/pull/9772 .

PR #9772, opened 2023-11-07
[coll] Pass context to various functions.

trivialfis:
In the future, the `Context` object would be required for collective operations, this PR passes the context object to some required functions to prepare for swapping out the implementation.


trivialfis:
cc @rongou

rongou (approved):

PR #9775, opened 2023-11-08
[jvm-packages] support stage-level scheduling

wbo4958:
This PR enables the stage-level scheduling when running xgboost on spark 3.4.0+ standalone or local cluster mode.


trivialfis (approved):
Thank you for the work on stage-level scheduling! Do we need to follow up with documents?

wbo4958:
Yeah. Let me file a follow-up PR to update the doc.

Issue #9765, opened 2023-11-07
[jvm-package] Upload XGBoost4J with Scala 2.13 to Maven Central


hcho3:
#9099 added support for Scala 2.13 to XGBoost4J, but I forgot to upload `xgboost4j_2.13-2.0.1.jar` (and other artifacts) to Maven Central. I will do this by end of this week.

cc @dotbg @wbo4958

dotbg:
thanks a lot!

hcho3:
XGBoost 2.0.2 is now available on Maven Central:
* https://repo1.maven.org/maven2/ml/dmlc/xgboost-jvm_2.12/2.0.2/
* https://repo1.maven.org/maven2/ml/dmlc/xgboost-jvm_2.13/2.0.2/
* https://repo1.maven.org/maven2/ml/dmlc/xgboost4j_2.12/2.0.2/
* https://repo1.maven.org/maven2/ml/dmlc/xgboost4j_2.13/2.0.2/

PR #9776, opened 2023-11-08
[jvm-packages] Add Scala version suffix to xgboost-jvm package

### LINKED ISSUE #9774: [jvm-packages] xgboost-jvm metapackage is not suffixed with Scala version
In past versions, the top-level metapackage `xgboost-jvm` was always suffixed with Scala version.
For example, from https://repo1.maven.org/maven2/ml/dmlc/xgboost-jvm_2.12/ :

Screenshot 2023-11-08 000427

However, the latest XGBoost (2.0.1) does not have the suffix:
Screenshot 2023-11-08 000529

As a result, downstream projects may not be able to correctly locate artifacts for a given Scala version. Now that we want to distribute XGBoost JARs for both Scala 2.12 and Scala 2.13, the issue is quite critical.

I am preparing a new patch release 2.0.2 to fix this issue.

hcho3:
Closes #9774. We'd need to release a new patch release (2.0.2) in order to upload artifacts with correct metadata.

@wbo4958 @dotbg Can you review this PR? I found the problem with `artifactId` while working on #9765.


dev/prepare_jvm_release.py
```diff
@@ -47,57 +50,76 @@ def cd(path):
...
+        "--release-version",
+        type=str,
+        required=True,
+        help="Version of the release being prepared",
+    )
     args = parser.parse_args()
 
-    if sys.platform != "darwin" or platform.machine() != "x86_64":
-        raise NotImplementedError("Please run this script using an Intel Mac")
+    if sys.platform != "darwin" or platform.machine() != "arm64":
```
dotbg, line 92: What is the reason the intel macs are excluded?

dev/prepare_jvm_release.py
```diff
@@ -115,60 +137,98 @@ def main():
                 cp(file, f"{xgboost4j_spark}/src/test/resources")
 
         print("====Creating directories to hold native binaries====")
-        for os_ident, arch in [("linux", "x86_64"), ("windows", "x86_64"), ("macos", "x86_64")]:
+        for os_ident, arch in [
+            ("linux", "x86_64"),
+            ("windows", "x86_64"),
+            ("macos", "x86_64"),
```
dotbg, line 143: this is not aligned with the line # 92

dotbg (commented):
Thanks a log for picking this up. I have few comments, but these are not related to the jdk build

dev/prepare_jvm_release.py
hcho3: The native binary for Intel Mac is built by the CI already. So I need to run the prep script on M1 mac to build the binary for M1. You can look at the content of the JAR to see that there are multiple libxgboost4j.dylib.

Issue #9204, opened 2023-05-26
Eliminate host synchronization.


trivialfis:
## column sampler
- [ ] https://github.com/dmlc/xgboost/blob/614f47c477472118a20838b4880ab1ab51572c3c/src/common/random.h#L177
- [ ] Need device implementation of sampling.

## gradient
Validate data once at the beginning.
- [ ] https://github.com/dmlc/xgboost/blob/614f47c477472118a20838b4880ab1ab51572c3c/src/objective/regression_obj.cu#L142

PR #9785, opened 2023-11-15
Implement column sampler in CUDA.

trivialfis:
Related https://github.com/dmlc/xgboost/issues/9204 .

- CUDA implementation.
- Extract the broadcasting logic, we will need the context parameter after revamping the collective implementation.
- Some changes to the event loop for fixing a deadlock in CI.
- Move argsort into algorithms.cuh, add support for cuda stream.

### DIFF
--- a/src/collective/loop.cc
+++ b/src/collective/loop.cc
@@ -117,12 +117,12 @@ void Loop::Process() {
       break;
     }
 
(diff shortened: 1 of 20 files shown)


trivialfis:
cc @rongou .

rongou (approved):

PR #9805, opened 2023-11-22
Cleanup code for distributed training.

trivialfis:
- Merge `GetNcclResult` into nccl stub.
- Split up utilities from the main dask module.
- Let Channel return `Result` to accommodate nccl channel.
- Remove old `use_label_encoder` parameter.


trivialfis:
cc @rongou .

rongou (approved):

Issue #9810, opened 2023-11-27
Roadmap for new R interface


david-cortes:
ref https://github.com/dmlc/xgboost/issues/9734
ref https://github.com/dmlc/xgboost/issues/9475

This issue is intended as a roadmap tracker for progress in bringing xgboost's R interface up to date and discussions around these tasks and coordination.

From the previous tasks, here I've made a list of potential tasks to take on, but I might be missing some things, and I've put the biggest task (new `xgboost()` function) under a single bullet point while in practice it'll likely involve multiple rounds of PRs. Please feel free to add more tasks to this list.

I've taken the liberty of classifying these issues in terms of whether they'd be blockers for releasing a new xgboost version or not, albeit some people might disagree with my assessments.

* [ ] (Blocker) Enable categorical features for current `DMatrix` constructors (`matrix`, `dgCMatrix`, `dgRMatrix`).
* [ ] (Blocker) Add support for creating DMatrices from R `data.frame` objects, automatically setting `factor` variables to be of categorical type in the DMatrix.
    * Note: these objects are a list of arrays which aren't necessarily in a single memory chunk, and which can have types `int` (int32_t), `double` (float64), and potentially `int64_t` from package `bit64`.
    * I guess this and the first point could be done in the same PR since they might be touching similar code sections.
* [ ] (Blocker) Enable multi-output input labels and predictions.
* [ ] (Low priority) Add a mechanism to create a `DMatrix` object from `arrow` objects (from package "arrow"). Like for data frames, should automatically recognize categorical columns from the categorical arrow type.
    * Note: the idea here is to exploit functions that work directly on arrow format, **without** converting to base R arrays (which do not support all the arrow types) along the way.
* [ ] Add an interface to create `QuantileDMatrix` objects from R, accepting the same kinds of inputs as `DMatrix` (`data.frame`, `matrix`, `dgCMatrix`, `dgRMatrix`, `arrow` if implemented, maybe `float::float32`), and also auto-recognizing categorical features for objects that have them (data frames and arrow tables).
* [ ] (Low priority) Add methods to get additional info from `DMatrix` objects that are currently missing from the R package, such as `get_quantile_cut` (guess this is just a call to `XGDMatrixGetQuantileCut`?).
* [ ] (Blocker) Move more `DMatrix` parameters that reference data towards `xgb.DMatrix()` function arguments, such as `qid`, `group`, `label_lower_bound`, `label_upper_bound` , etc.
[...]

PR #9816, opened 2023-11-28
[R] Use array interface for dense DMatrix creation

david-cortes:

This PR switches the underlying C function for DMatrix creation from dense R matrices to the `XGDMatrixCreateFromDense` function used in the python interface.

Along the way, it also fixes a potential memory leak in case of failures, in which a failure to allocate an R externalptr would lead to a longjump without freeing the allocated C++ object from its pointer.

I see there was already some mechanism to create the kind of array JSON that numpy uses in the DMatrix creators from sparse matrices, but couldn't find any equivalent for dense matrices. There also seems to be a dedicated module to create JSONs, but they look simple enough for `snprintf`.

The JSON parser for the config by the way seems to have issues with `nan` and `inf`, so this PR makes a conversion to `NaN` and `Infinity` which seem to be accepted.

### DIFF
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -74,14 +74,8 @@ namespace {
       jconfig["missing"] = R_NaInt;
     }
   } else {
-    const double missing_as_double = Rf_asReal(missing);
     // missing specified
-    if (std::isinf(missing_as_double)) {
-      jconfig["missing"] = (missing_as_double < 0) ? -std::numeric_limits<float>::infinity()
-                                                   : std::numeric_limits<float>::infinity();
-    } else {
-      jconfig["missing"] = Rf_asReal(missing);
-    }
+    jconfig["missing"] = Rf_asReal(missing);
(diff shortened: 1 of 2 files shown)


trivialfis (commented):
Thank you for working on the dense matrix, let me make some changes tomorrow for using the JSON class in XGBoost.

trivialfis:
Hi @david-cortes , I have made some changes to use the `linalg` module in XGBoost for generating the strings, could you please take a look when you are available? I'm open to suggestions and feel free to ping me if you find some of the code confusing.

david-cortes:
> Hi @david-cortes , I have made some changes to use the `linalg` module in XGBoost for generating the strings, could you please take a look when you are available? I'm open to suggestions and feel free to ping me if you find some of the code confusing.

Thanks for the changes. I've made some very small modifications on top and I guess it should be good to go now.

trivialfis (approved):
Nice work! Will merge once the CI is green.

PR #9818, opened 2023-11-28
[R] Update docs about fitting from CSR

david-cortes:

The docs had a note about not supporting DMatrix objects created from CSR when fitting models. The problem has been solved by now so the note is not applicable anymore.


trivialfis (approved):
Thank you for the fix!

PR #9824, opened 2023-11-29
[R] Move more functions to array interface

david-cortes:
From comment https://github.com/dmlc/xgboost/issues/9734#issuecomment-1832600306

This PR moves more R functions to the new array interface.

Here I've created a copy-paste of the matrix function for 1d vectors. I'm not sure why it needs `[[nodiscard]]` qualifiers though.

Along the way, it also solves a potential memory leak in the previous PR in which the call to `CHECK_CALL` could potentially lead to leaked memory of `std::string` objects that are used for the array interface.

### DIFF
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -59,6 +59,32 @@ namespace {
   return "";
 }
 
+[[nodiscard]] std::string MakeArrayInterfaceFromRVector(SEXP R_vec) {
+  const size_t vec_len = Rf_xlength(R_vec);
+
+  // Lambda for type dispatch.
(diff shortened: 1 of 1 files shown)


trivialfis:
> I'm not sure why it needs [[nodiscard]] qualifiers though.

Not required, it's just a suggestion from clang tidy to annotate functions with nodiscard when possible. We currently disable the suggestion to avoid massive change, but eventually we will turn it on. Feel free to ignore it if it's not useful.

> potentially lead to leaked memory of std::string

That's difficult to get right. I think I will forget about it after a period of time without writing R c code. Is it possible to return errors in a more linear fashion?

trivialfis (approved):
Thank you for the work on using array interface!

PR #9828, opened 2023-11-30
Support dataframe data format in native XGBoost.

trivialfis:
- Implement a columnar adapter.
- Refactor Python pandas handling code to avoid converting into a single numpy array.
- Add support in R for transforming columns.

This is not yet as efficient as we would like since we can't handle missing value indicators for each column. I will leave that as a future to-do instead.

Categorical data should work with R factor now, minus the uncertainty about how we should suggest people to use a consistent encoder.

related: https://github.com/dmlc/xgboost/issues/9810 .

### DIFF
--- a/R-package/R/xgb.DMatrix.R
+++ b/R-package/R/xgb.DMatrix.R
@@ -19,7 +19,8 @@
 #' @param missing a float value to represents missing values in data (used only when input is a dense matrix).
 #'        It is useful when a 0 or some other extreme value represents missing values in data.
 #' @param silent whether to suppress printing an informational message after loading from a file.
-#' @param feature_names Set names for features.
+#' @param feature_names Set names for features. Overrides column names in data
+#'        frame and matrix.
 #' @param nthread Number of threads used for creating DMatrix.
 #' @param group Group size for all ranking group.
 #' @param qid Query ID for data samples, used for ranking.
@@ -32,6 +33,8 @@
 #' If a DMatrix gets serialized and then de-serialized (for example, when saving data in an R session or caching
 #' chunks in an Rmd file), the resulting object will not be usable anymore and will need to be reconstructed
 #' from the original source of data.
+#' @param enable_categorical Experimental support of specializing for
+#'        categorical features. JSON/UBJSON serialization format is required.
 #'
 #' @examples
 #' data(agaricus.train, package='xgboost')
@@ -58,19 +61,28 @@ xgb.DMatrix <- function(
   qid = NULL,
   label_lower_bound = NULL,
   label_upper_bound = NULL,
-  feature_weights = NULL
+  feature_weights = NULL,
+  enable_categorical = FALSE
 ) {
   if (!is.null(group) && !is.null(qid)) {
     stop("Either one of 'group' or 'qid' should be NULL")
   }
+  cnames <- NULL
+  ctypes <- NULL
   if (typeof(data) == "character") {
-    if (length(data) > 1)
-      stop("'data' has class 'character' and length ", length(data),
-           ".\n  'data' accepts either a numeric matrix or a single filename.")
+    if (length(data) > 1) {
+      stop(
+        "'data' has class 'character' and length ", length(data),
+        ".\n  'data' accepts either a numeric matrix or a single filename."
+      )
+    }
     data <- path.expand(data)
     handle <- .Call(XGDMatrixCreateFromFile_R, data, as.integer(silent))
   } else if (is.matrix(data)) {
-    handle <- .Call(XGDMatrixCreateFromMat_R, data, missing, as.integer(NVL(nthread, -1)))
+    handle <- .Call(
+      XGDMatrixCreateFromMat_R, data, missing, as.integer(NVL(nthread, -1))
+    )
+    cnames <- colnames(data)
   } else if (inherits(data, "dgCMatrix")) {
     handle <- .Call(
       XGDMatrixCreateFromCSC_R,
@@ -103,6 +115,40 @@ xgb.DMatrix <- function(
       missing,
       as.integer(NVL(nthread, -1))
     )
+  } else if (is.data.frame(data)) {
+    ctypes <- sapply(data, function(x) {
+      if (is.factor(x)) {
+        if (!enable_categorical) {
+          stop(
+            "When factor type is used, the parameter `enable_categorical`",
+            " must be set to TRUE."
+          )
+        }
+        "c"
+      } else if (is.integer(x)) {
+        "int"
+      } else if (is.logical(x)) {
+        "i"
+      } else {
+        if (!is.numeric(x)) {
+          stop("Invalid type in dataframe.")
+        }
+        "float"
+      }
+    })
+    ## as.data.frame somehow converts integer/logical into real.
+    data <- as.data.frame(sapply(data, function(x) {
+      if (is.factor(x)) {
+        ## XGBoost uses 0-based indexing.
+        as.numeric(x) - 1
+      } else {
+        x
+      }
+    }))
+    handle <- .Call(
+      XGDMatrixCreateFromDF_R, data, missing, as.integer(NVL(nthread, -1))
+    )
+    cnames <- colnames(data)
   } else {
     stop("xgb.DMatrix does not support construction from ", typeof(data))
   }
@@ -119,7 +165,11 @@ xgb.DMatrix <- function(
   if (!is.null(base_margin)) {
     setinfo(dmat, "base_margin", base_margin)
   }
+  if (!is.null(cnames)) {
+    setinfo(dmat, "feature_name", cnames)
+  }
   if (!is.null(feature_names)) {
+    ## override cnames
     setinfo(dmat, "feature_name", feature_names)
   }
   if (!is.null(group)) {
@@ -136,6 +186,9 @@ xgb.DMatrix <- function(
   }
   if (!is.null(feature_weights)) {
     setinfo(dmat, "feature_weights", feature_weights)
+  }
+  if (!is.null(ctypes)) {
+    setinfo(dmat, "feature_type", ctypes)
   }
 
   return(dmat)
--- a/R-package/R/xgb.DMatrix.R
+++ b/R-package/R/xgb.DMatrix.R
@@ -67,7 +67,6 @@ xgb.DMatrix <- function(
   if (!is.null(group) && !is.null(qid)) {
     stop("Either one of 'group' or 'qid' should be NULL")
   }
-  cnames <- NULL
   ctypes <- NULL
   if (typeof(data) == "character") {
     if (length(data) > 1) {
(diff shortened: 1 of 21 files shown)


RAMitchell (approved):

python-package/xgboost/data.py
```diff
@@ -407,89 +412,83 @@ def is_pd_sparse_dtype(dtype: PandasDType) -> bool:
...
-    else:
-        transformed = data
-
-    def cat_codes(ser: pd.Series) -> pd.Series:
+    def cat_codes(ser: pd.Series) -> np.ndarray:
         if is_pd_cat_dtype(ser.dtype):
-            return ser.cat.codes
-        assert is_pa_ext_categorical_dtype(ser.dtype)
+            return _ensure_np_dtype(
+                ser.cat.codes.astype(np.float32).replace(-1.0, np.NaN).values,
```
david-cortes, line 424: Isn't `to_numpy()` the recommended method here?

R-package/R/xgb.DMatrix.R
```diff
@@ -81,6 +107,9 @@ xgb.DMatrix <- function(data, info = list(), missing = NA, silent = FALSE, nthre
   if (!is.null(cnames)) {
     setinfo(dmat, "feature_name", cnames)
   }
+  if (!is.null(ctypes)) {
+    setinfo(dmat, "feature_type", ctypes)
```
david-cortes, line 111: Worth adding to the docs that these are automatically determined for data frames.

trivialfis: I will add the `enable_categorical` parameter similar to the one in Python. Do you prefer `enable_categorical` or `enable.categorical` or something else?

david-cortes: I think `enable_categorical` would be better but don't have any strong preference.

python-package/xgboost/data.py
trivialfis: I'm not sure what's recommended, we are handling all the types explicitly here. I will change it to use `to_numpy` to save the trouble of calling `astype`.

R-package/src/xgboost_R.cc
```diff
@@ -222,6 +222,64 @@ XGB_DLL SEXP XGDMatrixCreateFromMat_R(SEXP mat, SEXP missing, SEXP n_threads) {
...
+  }
+
+  Json jinterface{std::move(array)};
+  auto sinterface = Json::Dump(jinterface);
+  Json jconfig{xgboost::Object{}};
+  jconfig["missing"] = asReal(missing);
+  jconfig["nthread"] = asInteger(n_threads);
+  auto sconfig = Json::Dump(jconfig);
+
+  CHECK_CALL(XGDMatrixCreateFromColumnar(sinterface.c_str(), sconfig.c_str(), &handle));
```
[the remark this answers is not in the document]
trivialfis: Thank you for catching that!

> Better to do everything that uses them in a braced section, and check the result code outside of those braces.

Excellent play book, will follow that.

R-package/src/xgboost_R.cc
```diff
@@ -222,6 +222,64 @@ XGB_DLL SEXP XGDMatrixCreateFromMat_R(SEXP mat, SEXP missing, SEXP n_threads) {
...
+  DMatrixHandle handle;
+
+  using xgboost::Json;
+
+  auto make_vec = [&](auto const *ptr, std::int32_t len) {
+    auto v = xgboost::linalg::MakeVec(ptr, len);
+    return xgboost::linalg::ArrayInterface(v);
+  };
+
+  std::vector<Json> array;
```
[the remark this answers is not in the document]
trivialfis: Will change it, thank you for the suggestion.

R-package/R/xgb.DMatrix.R
```diff
@@ -73,6 +76,29 @@ xgb.DMatrix <- function(data, info = list(), missing = NA, silent = FALSE, nthre
...
       as.integer(NVL(nthread, -1))
     )
+  } else if (is.data.frame(data)) {
+    ctypes <- unlist(lapply(data, function(x) {
+      if (is.factor(x)) {
+        "c"
+      } else if (is.integer(x) || is.logical(x)) {
+        "i"
+      } else {
+        "q"
```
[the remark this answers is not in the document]
trivialfis: added a stop.

R-package/src/xgboost_R.cc
```diff
@@ -222,6 +222,64 @@ XGB_DLL SEXP XGDMatrixCreateFromMat_R(SEXP mat, SEXP missing, SEXP n_threads) {
...
+  Json jinterface{std::move(array)};
+  auto sinterface = Json::Dump(jinterface);
+  Json jconfig{xgboost::Object{}};
+  jconfig["missing"] = asReal(missing);
+  jconfig["nthread"] = asInteger(n_threads);
+  auto sconfig = Json::Dump(jconfig);
+
+  CHECK_CALL(XGDMatrixCreateFromColumnar(sinterface.c_str(), sconfig.c_str(), &handle));
+
+  ret = Rf_protect(R_MakeExternalPtr(handle, R_NilValue, R_NilValue));
```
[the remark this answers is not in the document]
trivialfis: Moved the allocation.

trivialfis:
For some reason, even if just creating DMatrix (simplest possible task), the np array created from arrow extension is incredibly inefficient. I have checked the resulting DMatrices have the same hash. In addition, they run in the exact same code path inside libxgboost.

trivialfis:
Got it, it's caused by meta info instead of the covariate.

R-package/R/xgb.DMatrix.R
```diff
@@ -73,6 +82,40 @@ xgb.DMatrix <- function(data, info = list(), missing = NA, silent = FALSE, nthre
...
+          stop(
+            "When factor type is used, the parameter `enable_categorical`",
+            " must be set to TRUE."
+          )
+        }
+        "c"
+      } else if (is.integer(x)) {
+        "int"
+      } else if (is.logical(x)) {
+        "i"
```
david-cortes, line 98: Question there: I see this is also being used in the pandas adapter.

In R, boolean (logical) types are represented as C `int`, where possible values are FALSE (zero), NA (-INT_MAX), and TRUE (everything else), while python's bool type has only True and False.

I see you mention in a comment later that these get converted to numeric type, but the C++ code still checks for integer/logical-typed columns.

What would happen with these missing values encoded as -INT_MAX if the columns are supplied in their original types?

include/xgboost/c_api.h
```diff
@@ -159,6 +159,16 @@ XGB_DLL int XGDMatrixCreateFromURI(char const *config, DMatrixHandle *out);
 XGB_DLL int XGDMatrixCreateFromCSREx(const size_t *indptr, const unsigned *indices,
                                      const float *data, size_t nindptr, size_t nelem,
                                      size_t num_col, DMatrixHandle *out);
+/**
+ * @brief Create a DMatrix from columnar data. (table)
+ *
+ * @param data   See @ref XGBoosterPredictFromColumnar for details.
+ * @param config See @ref XGDMatrixCreateFromDense for details.
```
david-cortes, line 166: Something I'm wondering here: if this config already conveys the information about whether a column has integer type, is it actually needed to make a distinction between `q` and `int` in `feature_types`?

R-package/src/xgboost_R.cc
```diff
@@ -222,6 +222,69 @@ XGB_DLL SEXP XGDMatrixCreateFromMat_R(SEXP mat, SEXP missing, SEXP n_threads) {
...
+  using xgboost::Json;
+
+  auto make_vec = [&](auto const *ptr, std::int32_t len) {
+    auto v = xgboost::linalg::MakeVec(ptr, len);
+    return xgboost::linalg::ArrayInterface(v);
+  };
+
+  std::int32_t rc{0};
+  {
+    auto n_features = Rf_length(df);
```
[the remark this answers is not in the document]
trivialfis: Will change, didn't know the existance of that function.

R-package/R/xgb.DMatrix.R
```diff
@@ -73,6 +82,40 @@ xgb.DMatrix <- function(data, info = list(), missing = NA, silent = FALSE, nthre
       missing,
       as.integer(NVL(nthread, -1))
     )
+  } else if (is.data.frame(data)) {
+    ctypes <- unlist(lapply(data, function(x) {
```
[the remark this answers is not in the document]
trivialfis: Thank you for the suggestion, will change.

R-package/R/xgb.DMatrix.R
trivialfis: Thank you for the suggestion, I removed the `unlist` as suggested in https://github.com/dmlc/xgboost/pull/9828#discussion_r1420623833 .

trivialfis:
cc @david-cortes DF for label and base margin is still not yet supported. These are useful for multi-output/multi-label problems. But we can work on them later.

trivialfis:
Added the `cnames` configuration back for the matrix.

trivialfis:
@david-cortes Could you please help take another look?

R-package/R/xgb.DMatrix.R
```diff
@@ -58,19 +61,28 @@ xgb.DMatrix <- function(
...
+      )
+    }
     data <- path.expand(data)
     handle <- .Call(XGDMatrixCreateFromFile_R, data, as.integer(silent))
   } else if (is.matrix(data)) {
-    handle <- .Call(XGDMatrixCreateFromMat_R, data, missing, as.integer(NVL(nthread, -1)))
+    handle <- .Call(
+      XGDMatrixCreateFromMat_R, data, missing, as.integer(NVL(nthread, -1))
+    )
+    cnames <- colnames(data)
```
david-cortes, line 85: I think this would not be required anymore given that `feature_names` already makes the same function call.

david-cortes:
LGTM. Left two small comments.

R-package/R/xgb.DMatrix.R
```diff
@@ -58,19 +61,28 @@ xgb.DMatrix <- function(
   qid = NULL,
   label_lower_bound = NULL,
   label_upper_bound = NULL,
-  feature_weights = NULL
+  feature_weights = NULL,
+  enable_categorical = FALSE
```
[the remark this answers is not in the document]
trivialfis: > Question: do I understand it correctly that this parameter is only used to auto-detect categorical features from data frames, but would otherwise play no role if e.g. the user were to manually set this field in the DMatrix later through setinfo, for example?

Correct. It's more or less a guard to prevent surprise since XGBoost didn't accept categorical data before, which might cause issues in silence if we suddenly accept it.

R-package/R/xgb.DMatrix.R
trivialfis: Thank you for the explanation, removed.

R-package/R/xgb.DMatrix.R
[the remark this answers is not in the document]
trivialfis: I don't have strong preference on the naming, we have an introductory document for cat data in the tutorials, feel free to add additional explanation.

Issue #7616, opened 2022-01-28
R interface could use inplace prediction


david-cortes:
The Python interface where possible uses `inplace_predict`, which is fater than creating a `DMatrix` object for making predictions. The R interface OTOH always uses `DMatrix`, which requires 2 redundant copies of the data. It should be possible to use functions such as `XGBoosterPredictFromDense` in the R interface too, avoiding an extra data copy, and wrapping them should be even simpler since R does not support as many array configurations as numpy (e.g. no strided array views).

trivialfis:
Do you think we should add a parameter for the predict function to use inplace prediction optionally?  The prediction using `DMatrix` can be cached as we can use its memory address as the cache key so it's not always slower than inplace prediction.

david-cortes:
> Do you think we should add a parameter for the predict function to use inplace prediction optionally? The prediction using `DMatrix` can be cached as we can use its memory address as the cache key so it's not always slower than inplace prediction.

Bu whether it's cached or not, it first needs to produce a data copy when it casts the `double` R array to a `float` vector, and then when it copies that `float` vector to the DMatrix that's being created.

trivialfis:
Would you like to work on the interface?  :-). The R package can use a lot of help.

david-cortes:
I'll pass on this specific feature.

PR #9829, opened 2023-11-30
[R] Use inplace predict

david-cortes:
This PR switches the `predict` method to use inplace prediction when possible (dense matrix / CSR matrix as inputs).

It seems that there are quite a lot of limitations around when it is or isn't possible to use these functions which are not documented in the readthedocs page, so I am not sure if the conditions that I am adding here to check for whether inplace predict is supported are correct.

I'm also not sure what's the most efficient way of dealing with all the JSON manipulation here.

Along the way, since it is refactoring some of the earlier code introduced for processing the "missing" parameter, it also modifies the function that creates DMatrix objects from CSR/CSC. It doesn't make a practical difference there though, since currently only classes inheriting from "DsparseMatrix" (float64 values) are supported, but might come handy in the future if other sparse matrix classes (like "LsparseMatrix") become supported in the future.

### DIFF
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -341,9 +341,22 @@ predict.xgb.Booster <- function(object, newdata, missing = NA, outputmargin = FA
                                 predleaf = FALSE, predcontrib = FALSE, approxcontrib = FALSE, predinteraction = FALSE,
                                 reshape = FALSE, training = FALSE, iterationrange = NULL, strict_shape = FALSE, ...) {
   object <- xgb.Booster.complete(object, saveraw = FALSE)
+  config <- jsonlite::fromJSON(xgb.config(object))
 
-  if (!inherits(newdata, "xgb.DMatrix")) {
-    config <- jsonlite::fromJSON(xgb.config(object))
+  use_as_dense_matrix <- FALSE
+  use_as_csr_matrix <- FALSE
+  inplace_predict_supported <- !predcontrib && !predinteraction && !predleaf && is.null(ntreelimit)
+  if (inplace_predict_supported) {
+    if (config$learner$learner_train_param$booster != "gbtree") {
+      inplace_predict_supported <- FALSE
+    }
+  }
+  if (inplace_predict_supported && is.matrix(newdata)) {
+    use_as_dense_matrix <- TRUE
+  } else if (inplace_predict_supported && inherits(newdata, "dgRMatrix")) {
+    use_as_csr_matrix <- TRUE
+    csr_data <- list(newdata@p, newdata@j, newdata@x, ncol(newdata))
+  } else if (!inherits(newdata, "xgb.DMatrix")) {
     nthread <- strtoi(config$learner$generic_param$nthread)
     newdata <- xgb.DMatrix(
       newdata,
@@ -414,9 +427,21 @@ predict.xgb.Booster <- function(object, newdata, missing = NA, outputmargin = FA
     args$type <- set_type(6)
   }
 
-  predts <- .Call(
-    XGBoosterPredictFromDMatrix_R, object$handle, newdata, jsonlite::toJSON(args, auto_unbox = TRUE)
-  )
+  json_conf <- jsonlite::toJSON(args, auto_unbox = TRUE)
+  if (use_as_dense_matrix) {
+    predts <- .Call(
+      XGBoosterPredictFromDense_R, object$handle, newdata, missing, json_conf
+    )
+  } else if (use_as_csr_matrix) {
+    predts <- .Call(
+      XGBoosterPredictFromCSR_R, object$handle, csr_data, missing, json_conf
+    )
+  } else {
+    predts <- .Call(
+      XGBoosterPredictFromDMatrix_R, object$handle, newdata, json_conf
+    )
+  }
+
   names(predts) <- c("shape", "results")
   shape <- predts$shape
   ret <- predts$results
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -85,25 +85,27 @@ namespace {
   return "";
 }
 
-[[nodiscard]] std::string MakeJsonConfigForArray(SEXP missing, SEXP n_threads, SEXPTYPE arr_type) {
-  using namespace ::xgboost;  // NOLINT
-  Json jconfig{Object{}};
-
+void AddMissingToJson(xgboost::Json *jconfig, SEXP missing, SEXPTYPE arr_type) {
   const SEXPTYPE missing_type = TYPEOF(missing);
   if (Rf_isNull(missing) || (missing_type == REALSXP && ISNAN(Rf_asReal(missing))) ||
       (missing_type == LGLSXP && Rf_asLogical(missing) == R_NaInt) ||
       (missing_type == INTSXP && Rf_asInteger(missing) == R_NaInt)) {
     // missing is not specified
     if (arr_type == REALSXP) {
-      jconfig["missing"] = std::numeric_limits<double>::quiet_NaN();
+      (*jconfig)["missing"] = std::numeric_limits<double>::quiet_NaN();
     } else {
-      jconfig["missing"] = R_NaInt;
+      (*jconfig)["missing"] = R_NaInt;
     }
   } else {
     // missing specified
-    jconfig["missing"] = Rf_asReal(missing);
+    (*jconfig)["missing"] = Rf_asReal(missing);
   }
+}
 
+[[nodiscard]] std::string MakeJsonConfigForArray(SEXP missing, SEXP n_threads, SEXPTYPE arr_type) {
+  using namespace ::xgboost;  // NOLINT
+  Json jconfig{Object{}};
+  AddMissingToJson(&jconfig, missing, arr_type);
   jconfig["nthread"] = Rf_asInteger(n_threads);
   return Json::Dump(jconfig);
 }
@@ -234,7 +236,7 @@ XGB_DLL SEXP XGDMatrixCreateFromCSC_R(SEXP indptr, SEXP indices, SEXP data, SEXP
   Json jconfig{Object{}};
   // Construct configuration
   jconfig["nthread"] = Integer{threads};
-  jconfig["missing"] = xgboost::Number{asReal(missing)};
+  AddMissingToJson(&jconfig, missing, TYPEOF(data));
   std::string config;
   Json::Dump(jconfig, &config);
   CHECK_CALL(XGDMatrixCreateFromCSC(sindptr.c_str(), sindices.c_str(), sdata.c_str(), nrow,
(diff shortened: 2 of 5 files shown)


david-cortes:
fixes https://github.com/dmlc/xgboost/issues/7616

david-cortes:
I see there's a failing MSVC job, but it doesn't seem to contain any test failure, only warnings. No idea what went wrong there.

trivialfis:
Always fun to have R failing on Windows when one thinks everything is ready. ;-) I restarted the CI and it seems to be failing again. I can help debug it on Monday next week.

david-cortes:
Let's hope the R update in the msvc job helps here.

trivialfis:
Let me try to reproduce it.

R-package/R/xgb.Booster.R
```diff
@@ -341,9 +341,22 @@ predict.xgb.Booster <- function(object, newdata, missing = NA, outputmargin = FA
...
-
-  if (!inherits(newdata, "xgb.DMatrix")) {
-    config <- jsonlite::fromJSON(xgb.config(object))
+  config <- jsonlite::fromJSON(xgb.config(object))
+
+  use_as_dense_matrix <- FALSE
+  use_as_csr_matrix <- FALSE
+  inplace_predict_supported <- !predcontrib && !predinteraction && !predleaf && is.null(ntreelimit)
+  if (inplace_predict_supported) {
+    if (config$learner$learner_train_param$booster != "gbtree") {
```
trivialfis, line 350: dart is supported as well. Only gblinear is not supported.

trivialfis:
I can reproduce the error on my machine, and somehow magically got it working by `cout << res_code << endl` in the new prediction function. The log from MSVC looks suspicious.
Other than the error, the prediction function needs to consider base margin as well, which requires the use of proxy DMatrix. I will look into the QDM PR for the proxy implementation in R.
```
 C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.37.32822\include\stdarg.h(29,9): warning C4068: unknown pragma 'Rf_warning' [D:\a\xgboost\xgboost\build\R-package\xgboost-r.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\ucrt\corecrt_malloc.h(13,9): warning C4068: unknown pragma 'Rf_warning' [D:\a\xgboost\xgboost\build\R-package\xgboost-r.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\ucrt\corecrt_malloc.h(14,9): warning C4068: unknown pragma 'Rf_warning' [D:\a\xgboost\xgboost\build\R-package\xgboost-r.vcxproj]
```

david-cortes:
BTW, while trying to add base margin here, I ran into an issue which manifested itself with both inplace predictions and dmatrix predictions: https://github.com/dmlc/xgboost/issues/9869

trivialfis:
> wouldn't setting a proxy dmatrix inside the prediction function be slower and/or consume the same amount of memory

No, the proxy DMatrix is just a reference to the original data, so, don't free the data without first freeing the proxy dmatrix ...

trivialfis:
With hindsight, the inplace prediction can be completely implemented with proxy DMatrix without those special parameters for `csr/dense`, but at this point, there's no real benefit to making such breaking changes.

david-cortes:
Added `base_margin` as a parameter to `predict`, plus a function to use the in-place prediction on data frames.

david-cortes:
@trivialfis After adding support for categorical features for data frames, I just realized that passing them to `predict` will now require re-identifying which of those earlier factor columns were categorical, given that they need to be manually converted to base-0 indexing.

What would be the best way to retrieve that information (which columns are categorical) from the booster?

trivialfis:
@david-cortes In Python, the booster has the same `feature_type` parameter, search `XGBoosterGetStrFeatureInfo` in core.py.

david-cortes:
@trivialfis Thanks for the tip. Would be ideal if you could investigate the failing MSVC job here so that we can merge this PR soon, as I'm finding some of the code here will be needed for the QuantileDMatrix and other PRs too.

I unfortunately don't know how best to debug an MSVC-compiled module that needs to be imported in R, as it can't use typical tools like sanitizers with a regular R build.

trivialfis:
I can try, but I admit I don't use Windows very often.

hcho3:
Is this PR independent of #9902?

I use Windows as a daily driver, so should be able to help debug the error in this PR.

trivialfis:
> Is this PR independent of https://github.com/dmlc/xgboost/pull/9902?

yes.

trivialfis:
Tried to use the sanitizer bundled in vs, but ran into a compiler internal error ...

david-cortes:
> Tried to use the sanitizer bundled in vs, but ran into a compiler internal error ...

Does that compiler error still happen if you comment out the R-specific code introduced here?

trivialfis:
I have made no progress on the Windows error, we will have to delay this PR for a while.

david-cortes:
> Discussion with @hcho3 , we are likely to move away from MSVC for the R package, along with which, we will drop support for building R package with GPU on Windows using MSVC. We haven't ruled out the possibility of using rtools and CMake to build GPU support, in addition, recommending WSL2 is a potential option as well.

Is that just for the R package, or also for the python package for windows? If just for R, have there been MSVC-specific issues like this before? Curious about what makes it problematic.

trivialfis:
Could you please help rebase the PR?

david-cortes:
Updated, but I'm now not 100% sure if everything is correct.

I suddenly started getting an error in this test and the one right after it after merging the latest master:
https://github.com/dmlc/xgboost/blob/c5d0608057f3c0b335fd31594c65a3a3ca4afca3/R-package/tests/testthat/test_basic.R#L142

Before, it was using `ntreelimit`, which I guess meant that the code was converting to DMatrix, while now it's actually using inplace predict.

It looks like predictions from in-place match those from DMatrix and from `xgb.train` up to only about ~1e-7 in this particular case (dart booster), while elsewhere they seem to match closer.

***********************************

BTW, would be ideal if you could merge the PR about data iterators before this one:
https://github.com/dmlc/xgboost/pull/9913

trivialfis:
> Updated, but I'm now not 100% sure if everything is correct.

Please let me know how things are going. ;-) Is there anything I can help?

david-cortes:
> > Updated, but I'm now not 100% sure if everything is correct.
> 
> Please let me know how things are going. ;-) Is there anything I can help?

PR is ready for review. Would be ideal if you could look into this issue with loss of floating point precision for dart. Since the difference between inplace and dmatrix is below $10^{-6}$ I guess it might not be a correctness problem, but it's odd that it gets a larger difference with dart than with gbtree.

R-package/R/xgb.Booster.R
```diff
@@ -77,26 +77,44 @@ xgb.get.handle <- function(object) {
...
 #'        local data file, or `xgb.DMatrix`.
-#'        For single-row predictions on sparse data, it is recommended to use the CSR format.
-#'        If passing a sparse vector, it will take it as a row vector.
-#' @param missing Only used when input is a dense matrix. Pick a float value that represents
-#'        missing values in data (e.g., 0 or some other extreme value).
+#'
+#'        For single-row predictions on sparse data, it's recommended to use CSR format. If passing
+#'        a sparse vector, it will take it as a row vector.
+#'
+#'        Note that, for repeated predictions, one might want to create a DMatrix to pass here instead
```
trivialfis, line 89: not sure what you mean by repeated predictions, prediction using the same data repeatedly, or with different data? For the latter case, inplace prediction can be faster due to the overhead of allocating large chunks of memory for DMatrix.

R-package/R/xgb.Booster.R
```diff
@@ -77,26 +77,44 @@ xgb.get.handle <- function(object) {
...
+#'              the operation slower than in an equivalent `matrix` object.
+#'        \item The order of the columns must match with that of the data from which the model was fitted
+#'              (i.e. columns will not be referenced by their names, just by their order in the data).
+#'        \item If the model was fitted to data with categorical columns, these columns must be of
+#'              `factor` type here, and must use the same encoding (i.e. have the same levels).
+#'        \item If `newdata` contains any `factor` columns, they will be converted to base-0
+#'              encoding (same as during DMatrix creation) - hence, one should not pass a `factor`
+#'              under a column which during training had a different type.
+#'        }
+#' @param missing Float value that represents missing values in data (e.g., 0 or some other extreme value).
```
trivialfis, line 103: We usually use `nan` as missing value, since XGBoost can handle 0 just as any other real number.

This is also a headache for us since many sparse matrix implementations assume 0 as missing.

R-package/R/xgb.Booster.R
```diff
@@ -287,16 +311,77 @@ xgb.get.handle <- function(object) {
...
                                 reshape = FALSE, training = FALSE, iterationrange = NULL, strict_shape = FALSE,
-                                validate_features = FALSE, ...) {
+                                validate_features = FALSE, base_margin = NULL, ...) {
   if (validate_features) {
     newdata <- validate.features(object, newdata)
   }
-  if (!inherits(newdata, "xgb.DMatrix")) {
+  is_dmatrix <- inherits(newdata, "xgb.DMatrix")
+  if (is_dmatrix && !is.null(base_margin)) {
+    warning("'base_margin' is ignored when passing 'xgb.DMatrix' as input.")
```
trivialfis, line 320: Maybe an error instead? I don't see any case that this is not an error.

R-package/R/xgb.Booster.R
```diff
@@ -287,16 +311,77 @@ xgb.get.handle <- function(object) {
...
+  }
+
+  use_as_df <- FALSE
+  use_as_dense_matrix <- FALSE
+  use_as_csr_matrix <- FALSE
+  n_row <- NULL
+  if (!is_dmatrix) {
+
+    inplace_predict_supported <- !predcontrib && !predinteraction && !predleaf
```
trivialfis, line 329: We should be able to support leaf in the near future.

R-package/src/xgboost_R.cc
```diff
@@ -1247,7 +1247,60 @@ XGB_DLL SEXP XGBoosterEvalOneIter_R(SEXP handle, SEXP iter, SEXP dmats, SEXP evn
   return mkString(ret);
 }
 
-XGB_DLL SEXP XGBoosterPredictFromDMatrix_R(SEXP handle, SEXP dmat, SEXP json_config)  {
+namespace {
+
+struct ProxyDmatrixError : public std::exception {};
+
+struct ProxyDmatrixWrapper {
```
trivialfis, line 1254: is this related to the R implementation?

R-package/R/xgb.Booster.R
david-cortes: Yes, default value is R's `NA`, which gets translated to either `NAN` or `R_NaInt` depending on the input type.

R-package/src/xgboost_R.cc
david-cortes: Yes, it's used in order to break out of a C++ block, so that C++ objects get destructed before throwing an R error.

R-package/R/xgb.Booster.R
david-cortes: Updated the wording.

R-package/R/xgb.Booster.R
david-cortes: Changed to error.

trivialfis (approved):

trivialfis:
Thank you for the work on inplace prediction! The R package is looking much better now thanks to your great work.

PR #9839, opened 2023-12-03
[R] Enable multi-output objectives

david-cortes:

This PR enables multi-column features in `DMatrix`, thereby allowing multi-output objectives.

Note that it switches some of the earlier code that was there to use the recently introduced helpers to create array interface strings from R objects, so as to avoid unnecessary memory copies and support integer data types.


trivialfis (approved):
Thank you for the work on multi-output!

trivialfis:
Let me look into the issue.

PR #9800, opened 2023-11-21
Add support inference on SYCL devices

razdoburdin:
In continuation of https://github.com/dmlc/xgboost/pull/9691
This PR adds experimental support of prediction with SYCL devices.

The key changes are:
- Added the new predictor: `sycl_predictor`, which is calling if SYCL device parameter is specified.
- Currently only the `PredictBatch` method has a native SYCL implementation.
- For building with SYCL support, user should specify the key `-DPLUGIN_SYCL=ON` for `cmake.`
- C++ unit tests for SYCL were rewrited for better unification with another tests in the repo.
- Python tests for SYCL prediction were added.
- CI configuration files were renewed, SYCL tests were added.

PR #9846, opened 2023-12-04
Sycl implementation for objective functions

razdoburdin:
In continuation of https://github.com/dmlc/xgboost/pull/9800
Replaces of https://github.com/dmlc/xgboost/pull/7732

This PR adds SYCL implementation and corresponding tests for `multi:softmax`, `multi:softprob`, `reg:squarederror`, `reg:squaredlogerror`, `reg:logistic`, `binary:logistic`, `binary:logitraw` objectives.

The dispatching is organized as follows. In case of execution device is set to `sycl`, the original name of objective function is replaced by `original_name_sycl`. In case of requited function hasn't sycl implementation, the original name is used.


plugin/sycl/objective/multiclass_obj.cc
```diff
@@ -0,0 +1,258 @@
...
+ * \tparam Iterator The type of the iterator.
+ */
+template<typename Iterator>
+inline Iterator FindMaxIndex(Iterator begin, Iterator end) {
+  Iterator maxit = begin;
+  for (Iterator it = begin; it != end; ++it) {
+    if (*it > *maxit) maxit = it;
+  }
+  return maxit;
+}
```
trivialfis, line 73: Any chance we can use the one in `common/math.h`?

plugin/sycl/objective/multiclass_obj.cc
```diff
@@ -0,0 +1,258 @@
...
+
+struct SoftmaxMultiClassParam : public XGBoostParameter<SoftmaxMultiClassParam> {
+  int num_class;
+  // declare parameters
+  DMLC_DECLARE_PARAMETER(SoftmaxMultiClassParam) {
+    DMLC_DECLARE_FIELD(num_class).set_lower_bound(1)
+        .describe("Number of output class in the multi-class classification.");
+  }
+};
```
trivialfis, line 82: Maybe create a header instead of copying?

plugin/sycl/objective/regression_loss.h
```diff
@@ -0,0 +1,155 @@
...
+  // duplication is necessary, as __device__ specifier
+  // cannot be made conditional on template parameter
+  static bst_float PredTransform(bst_float x) { return Sigmoid(x); }
+  static bool CheckLabel(bst_float x) { return x >= 0.0f && x <= 1.0f; }
+  static bst_float FirstOrderGradient(bst_float predt, bst_float label) {
+    return predt - label;
+  }
+  static bst_float SecondOrderGradient(bst_float predt, bst_float label) {
+    const bst_float eps = 1e-16f;
+    return std::max(predt * (1.0f - predt), eps);
```
trivialfis, line 85: So .. is this `sycl::max` or `std::max`? If the compiler can handle `std::max` then we can avoid lots of copy and paste.

plugin/sycl/objective/regression_loss.h
```diff
@@ -0,0 +1,155 @@
...
+    return 1.0f;
+  }
+  static bst_float ProbToMargin(bst_float base_score) { return base_score; }
+  static const char* LabelErrorMsg() { return ""; }
+  static const char* DefaultEvalMetric() { return "rmse"; }
+
+  static const char* Name() { return "reg:squarederror_sycl"; }
+
+  static ObjInfo Info() { return {ObjInfo::kRegression, true, false}; }
+};
```
trivialfis, line 44: Looks like the same as the original one.

plugin/sycl/objective/regression_loss.h
```diff
@@ -0,0 +1,155 @@
...
+  static T FirstOrderGradient(T predt, T label) { return predt - label; }
+  template <typename T>
+  static T SecondOrderGradient(T predt, T label) {
+    const T eps = T(1e-16f);
+    return std::max(predt * (T(1.0f) - predt), eps);
+  }
+  static bst_float ProbToMargin(bst_float base_score) {
+    CHECK(base_score > 0.0f && base_score < 1.0f)
+        << "base_score must be in (0,1) for logistic loss, got: " << base_score;
+    return -logf(1.0f / base_score - 1.0f);
```
trivialfis, line 99: > // TODO(razdoburdin): SYCL does not fully support std math inside offloaded kernels

Would be great if we could avoid copying when the function is supported.

src/learner.cc
```diff
@@ -528,6 +528,9 @@ class LearnerConfiguration : public Learner {
     auto const& objective_fn = learner_parameters.at("objective");
     if (!obj_) {
       CHECK_EQ(get<String const>(objective_fn["name"]), tparam_.objective);
+      if (ctx_.IsSycl()) {
+        tparam_.objective = ObjFunction::GetSyclImplementationName(tparam_.objective);
+      }
```
trivialfis, line 533: Is it possible that the dispatching happens inside `ObjFunction::Create`? Also, what happens if a user saves the model, loads it back on a different machine that doesn't have a sycl device?

trivialfis (commented):
Hi, thank you for the work on objective functions. Some initial comments for discussion. It would be great if we could reduce the amount of copied code.

> Find a better way to dispatch names of SYCL kernels with various

If you have ideas on how to refactor the code, please share. ;-)

src/learner.cc
razdoburdin: It is a good idea. In this case we wouldn't have problems with saved models.

plugin/sycl/objective/multiclass_obj.cc
razdoburdin: done

trivialfis (approved):

PR #9847, opened 2023-12-04
[R] Bump requirement to 4.3.2.

trivialfis:
https://github.com/dmlc/xgboost/pull/9835

### DIFF
--- a/R-package/DESCRIPTION
+++ b/R-package/DESCRIPTION
@@ -58,7 +58,7 @@ Suggests:
     float,
     titanic
 Depends:
-    R (>= 3.3.0)
+    R (>= 4.3.2)
 Imports:
     Matrix (>= 1.1-0),
     methods,


trivialfis:
cc @david-cortes .

david-cortes:
> I don't know the context of this PR (I'm sure I've missed a comment on one of the threads), so I apologize if I'm saying something that's already been considered but just in case... I don't think a version of `{xgboost}` with a requirement `R (>= 4.3)` could be released to CRAN currently.
> 
> r-oldrel on CRAN is still R 4.2.x.
> 
> See my comments in this related LightGBM discussion: [microsoft/LightGBM#6213 (review)](https://github.com/microsoft/LightGBM/pull/6213#pullrequestreview-1749635843).

There's no requirement to have the latest version of a package running under oldrel. As an example, this package depends on R >= 4.3 and doesn't get any issue with the checks, which are running on an older version of the package for oldrel:
https://cran.r-project.org/web/packages/outliertree/index.html

jameslamb:
Oh interesting! I always thought it was necessary that packages on CRAN pass checks on all of these: https://cran.r-project.org/web/checks/check_flavors.html

Sorry for being a distraction, and thanks for teaching me something new 😊

trivialfis:
Need to spend some time on Windows to get MSVC to work with R 4.3. Hopefully can find a workaround somehow.

https://github.com/dmlc/xgboost/actions/runs/7096217610/job/19314372806?pr=9847

david-cortes:
Curious that it fails like that. The offending header has this disclaimer:
> This definition uses an anonymous structure, which is defined in C11 (but
not C99).  It is, however, supported at least by GCC, clang and icc.  The
private_data_c member should never be used in code, but tells the compiler
about type punning when accessing the .r and .i elements, so is safer to use
when interfacing with Fortran COMPLEX*16 or directly C99 _Complex double
(PR#18430).

Is it perhaps a matter of using a different C++ standard?

jameslamb:
```text
syntax error: missing ';' before identifier 'private_data_c' 
```

^ We encountered this in LightGBM when upgrading to R4.3 in that project's CI.

See https://github.com/microsoft/LightGBM/pull/6061

We fixed it by adding `-DR_LEGACY_COMPLEX` for MSVC-based builds.

https://github.com/microsoft/LightGBM/blob/f5b6bd60d9d752c8e5a75b11ab771d0422214bb4/CMakeLists.txt#L386-L392

trivialfis:
Ah! Thank you both for looking into this, it has been most helpful!

trivialfis:
Hi, please help review the PR when you are available.

jameslamb (approved):
Given that you want to move to `R >= 4.3`.

The fix to the `if args.compiler == "msvc"` block in `test_r_package.py` especially looks like it might fix some test coverage that's accidentally not been working for a while (maybe those builds were falling back to MinGW?).

But could you link me to the discussion that led to wanting to enforce `R >= 4.3` in `DESCRIPTION`? Sorry if I missed it, I know there's been a lot of activity on the R package here recently.

To the best of my understanding, making that change will have at least 2 negative side effects I can think of:

* forcing all of `{xgboost}`'s strong reverse dependencies to either drop support for older version of R or move their `{xgboost}` support from from strong (`Imports`, `Depends`) to conditional (`Suggests`)
    - *if they want to use newer versions of `{xgboost}`*
* preventing anyone using `R < 4.3` from even installing the newer versions of the library

For example `{shapviz}` ([link](https://cran.r-project.org/web/packages/shapviz/index.html)) and `{modeltime}` ([link](https://cran.r-project.org/web/packages/modeltime/index.html)) currently have strong dependencies on `{xgboost}` and support `R >= 3.6.0`.

R 4.3.x has only been out for 8 months at this point ([R release history](https://cran.r-project.org/bin/windows/base/old/)).

If these concerns have already been considered, I apologize for the extra noise and for slowing you down unnecessarily.

jameslamb:
Ok that addresses my concerns, just wanted to be sure it had been considered before. Thank you!

jameslamb:
Ahhh got it, thanks!

I think I didn't understand that PRs merged to `master` were considered to be contributing to that new R interface. I was assuming anything merged to `master` could end up being pushed to CRAN the next time XGBoost does a release, and that a new release could happen in days to weeks from now, not months.

Thanks for correcting my understanding.

trivialfis:
I did submit an update for 1.7 to keep XGB on CRAN though.

mayer79:
Two comments

1. Requiring R >= 4.3 is uncommonly strict. For instance, {ggplot2} requires R 3.4.4, {data.table} the even older 3.1.0. Companies often stick to a certain R version until there is sufficient pressure to update. This does not mean we should not do it, but there needs to be very good reasons to do so.
2. Unlike in Python, R packages that use XGBoost are forced to work with its current CRAN version. It is not possible to specify "xgboost < 2" in some requirement file. Put differently, all packages that use XGBoost will also require R 4.3. I am not 100% sure if this also applies to "suggested" packages.

david-cortes:
> there need to be very good reasons to do so.

There are very good reasons to: R 4.3 introduces functionality that can be used to define custom serialization for more object types.

For xgboost, this means it will possible (after a few more PRs) to only trigger serialization on-demand, which has multiple benefits:
* Reduced memory usage.
* Faster fitting times (assuming a model that doesn't get immediately serialized).
* No need to call functions like `xgb.Booster.complete` from the user side (in turn preventing mistakes such as de-serializing models at every `predict` call in a forked process, or timing the `predict` method incorrectly).
* Reduced chances of generating inconsistencies between the R metadata and the C++ objects when something fails or gets interrupted.

trivialfis:
For me, this is quiet convincing

> Reduced chances of generating inconsistencies between the R metadata and the C++ objects when something fails or gets interrupted.

The serialization has been a foot gun since forever.

PR #9849, opened 2023-12-04
[R] Enable vector-valued parameters

david-cortes:

This PR adds support for passing parameters that might consist of a vector of values instead of a single value, thereby enabling support for multi-quantile regression.

I see in the python interface that parameter `eval_metric` behaves quite differently as it requires passing entries by triggering multiple calls to `XGBoosterSetParam` but with the same parameter name, instead of passing a json list as a single parameter, so that's the approach I am following here.

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -93,6 +93,14 @@ check.booster.params <- function(params, ...) {
     interaction_constraints <- sapply(params[['interaction_constraints']], function(x) paste0('[', paste(x, collapse = ','), ']'))
     params[['interaction_constraints']] <- paste0('[', paste(interaction_constraints, collapse = ','), ']')
   }
+
+  # for evaluation metrics, should generate multiple entries per metric
+  if (NROW(params[['eval_metric']]) > 1) {
+    eval_metrics <- as.list(params[["eval_metric"]])
+    names(eval_metrics) <- rep("eval_metric", length(eval_metrics))
+    params_without_ev_metrics <- within(params, rm(eval_metric))
+    params <- c(params_without_ev_metrics, eval_metrics)
+  }
   return(params)
 }
 
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -98,7 +98,7 @@ check.booster.params <- function(params, ...) {
   if (NROW(params[['eval_metric']]) > 1) {
     eval_metrics <- as.list(params[["eval_metric"]])
     names(eval_metrics) <- rep("eval_metric", length(eval_metrics))
-    params_without_ev_metrics <- within(params, rm(eval_metric))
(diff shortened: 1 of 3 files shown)


R-package/R/utils.R
```diff
@@ -93,6 +93,14 @@ check.booster.params <- function(params, ...) {
...
     params[['interaction_constraints']] <- paste0('[', paste(interaction_constraints, collapse = ','), ']')
   }
+
+  # for evaluation metrics, should generate multiple entries per metric
+  if (NROW(params[['eval_metric']]) > 1) {
+    eval_metrics <- as.list(params[["eval_metric"]])
+    names(eval_metrics) <- rep("eval_metric", length(eval_metrics))
+    params_without_ev_metrics <- within(params, rm("eval_metric"))
+    params <- c(params_without_ev_metrics, eval_metrics)
+  }
```
trivialfis, line 103: A bit confused here, how does the existing code handle multiple evaluation metrics? I think the PR #8657 was supposed to handle it, but I could be wrong.

trivialfis (commented):
Thank you for the work on quantile regression! One small question in the comment.

> instead of passing a json list as a single parameter

That would be possible. The JSON parameter was introduced long after the initial parameter implementation.

R-package/R/utils.R
david-cortes: Up this point, if following the code in the master branch, the object `p` would also have a list with one entry per metric, PLUS another entry where the metrics are together.

For example, the code in the test would produce something like this for variable `p` up to the point before the `lapply`:
```
$eval_metric
$eval_metric[[1]]
[1] "error"

$eval_metric[[2]]
[1] "auc"

$eval_metric[[3]]
[1] "logloss"


$eval_metric
[1] "error"

$eval_metric
[1] "auc"

$eval_metric
[1] "logloss"
```

The call to `lapply` (current master branch) as it is in this PR would take the first entry from the multi-valued one, so it will transform it like this:
```
$eval_metric
[1] "error"

$eval_metric
[1] "error"

$eval_metric
[1] "auc"

$eval_metric
[1] "logloss"
```

After the changes in this PR, it will only have the non-repeated ones. Since the `lapply` code is being changed here in way in which multi-valued entries now produce a different input to the C function, I thought the easiest way would be to remove the multi-valued entry.

I experimented with passing them as a JSON list, but that doesn't seem to have the intended effect - for example, `xgb.config` produces this error:
```
Error in xgboost::xgb.config(object) : 
  [21:31:12] ../..//src/metric/metric.cc:49: Unknown metric function ["error", "auc", "logloss"]
```

trivialfis: Thank you for the explanation, I can reproduce the repeated values.

trivialfis (approved):

Issue #7118, opened 2021-07-21
Survival AFT analysis using xgb.cv


mikecyterski:
Hello folks. Has anyone used Survival Analysis with AFT using xgb.cv in R? I see examples online using xgb.train, where you can set lower and upper limits on the label (response variable) within the xgb.DMatrix, but I keep getting errors in my code when trying to do this with xgb.cv. I set the lower and upper label bounds in this fashion:

y_lower_bound -> dataset[,1]
y_upper_bound -> dataset[,2]
setinfo(covar_train, 'label_lower_bound', y_lower_bound)
setinfo(covar_train, 'label_upper_bound', y_upper_bound)

But when I make the call to xgb.cv, I get an error about the label:

Error in xgb.cv(params = params, covar_train, early_stopping_rounds = 10,  : 
  Labels must be provided for CV either through xgb.DMatrix, or through 'label=' when 'data' is matrix

covar_train is my DMatrix of model covariates. If I "getinfo" about the label of covar_train, it comes back "NULL", which isn't surprising, because I haven't set it, but how to indicate to xgb.cv to use the upper and lower label bounds in conjunction with the 'survival:aft' objective function and the 'aft-nloglik' evaluation metric? I've run very similar code using xgb.train, and that seems to work fine.

If anyone can share code that implements the Survival AFT model using xgb.cv, that'd be much appreciated!

trivialfis:
cc @hcho3

mayer79:
I once had the same problem.  IMHO, `xgb.cv` does not deal with `label_lower_bound` and `label_upper_bound`. My solution was to do the cross-validation without `xgb.cv`.

mikecyterski:
Thanks MM. Someone else recommended writing my own cross-validation routine but I was waiting to see if there was an easier solution. Would you mind sharing your code?

PR #9863, opened 2023-12-08
[R] Fix xgb.cv() for AFT models

mayer79:
This PR fixes https://github.com/dmlc/xgboost/issues/7118

In order to keep the `xgb.cv()` API as it is:

- `data` must be an `xgb.DMatrix` object
- which contains the two infos 'label_lower_bound' and 'label_upper_bound'.

Automatic stratified splitting is deactivated with a warning. @david-cortes this is something we need to keep in mind for multioutput regressions as well.

### DIFF
--- a/R-package/R/xgb.cv.R
+++ b/R-package/R/xgb.cv.R
@@ -134,13 +134,13 @@ xgb.cv <- function(params = list(), data, nrounds, nfold, label = NULL, missing 
 
   check.custom.obj()
   check.custom.eval()
-  
+
   # AFT uses 'label_lower/upper_bound' instead of 'label'
   if (is.character(params$objective) && params$objective == 'survival:aft') {
     if (!inherits(data, 'xgb.DMatrix')) {
       stop("Objective 'survival:aft' requires the data to be an 'xgb.DMatrix'.")
     }
-    if (is.null(getinfo(data, name = 'label_lower_bound')) || 
+    if (is.null(getinfo(data, name = 'label_lower_bound')) ||
         is.null(getinfo(data, name = 'label_upper_bound'))) {
       stop("Objective 'survival:aft' requires 'label_lower_bound' and 'label_upper_bound.")
     }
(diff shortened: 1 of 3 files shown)


R-package/R/xgb.cv.R
```diff
@@ -134,17 +134,29 @@ xgb.cv <- function(params = list(), data, nrounds, nfold, label = NULL, missing
...
-    if (!is.null(label))
-      warning("xgb.cv: label will be ignored, since data is of type xgb.DMatrix")
-    cv_label <- getinfo(data, 'label')
+  
+  # AFT uses 'label_lower/upper_bound' instead of 'label'
+  if (is.character(params$objective) && params$objective == 'survival:aft') {
+    if (!inherits(data, 'xgb.DMatrix')) {
+      stop("Objective 'survival:aft' requires the data to be an 'xgb.DMatrix'.")
+    }
+    if (is.null(getinfo(data, name = 'label_lower_bound')) ||
```
david-cortes, line 143: Something I am wondering: most of these `getinfo` used throughout the function are only checking whether the DMatrix contains that field or not, but each call involves creating a full data copy just to check that it exists. @trivialfis Is there perhaps some function that could be used to just check if the DMatrix has the field or not? Do functions like `XGDMatrixGetFloatInfo` actually create a data copy when calling them?

david-cortes:
Thanks for looking into this. Left a small comment.

Although I am also thinking: if it constructs a DMatrix directly from `data` and `labels` - wouldn't it also require just a few lines to allow such a transformation with bounds as function arguments?

david-cortes:
> We _could_ add more arguments to `xgb.cv()`. I was actually thinking in the other direction: let the function work only with xgb.DMatrix input and remove the "label" argument. Why? The current signature of xgb.cv() is very incomplete. For instance there is no "weight" argument.

Yes, that makes sense too - then we wouldn't need to update things in two places if new parameters come out.

david-cortes:
There's now a function `xgb.DMatrix.hasinfo` which could be used here to check whether the DMatrix has a given field without involving data copies:
https://github.com/dmlc/xgboost/blob/ff3d82c006083c01f7e1d9796564d32844ea952c/R-package/R/xgb.DMatrix.R#L223

mayer79:
Closed in favour of #10031

PR #9864, opened 2023-12-10
[Draft] [R] Add QuantileDMatrix creation from dense matrices

david-cortes:

This PR adds a wrapper over QuantileDMatrix for dense matrix inputs.

I am not very sure that I'm implementing all of this correctly so would be ideal if a maintainer could take a deep look.

I have some doubts about how this class (QuantileDMatrix) is meant to be created:
* Should feature types / names / etc. be added to the proxy dmatrix as part of the iterations?
* Should labels and/or bounds be set as part of the iterations? If not, are categorical features supported for this class?
* If the call to `XGQuantileDMatrixCreateFromCallback` fails, returning a non-zero code, does the proxy dmatrix still need to be freed manually through `XGDMatrixFree`?
* Is there something that would currently work with a regular DMatrix but not with a QuantileDMatrix, and which would need to be checked on the R side instead of passing things to the C functions?

Also a small note: there's a PR which [hasn't yet been merged](https://github.com/dmlc/xgboost/pull/9829) as it's currently failing MSVC jobs, which introduces some refactorings around the handling of 'missing' parameter in the JSON configs. After such refactoring is introduced, it should also be used as part of the code introduced here, as otherwise missing values in integer matrices will not be handled correctly.


trivialfis:
Let me briefly answer some the of questions first,  then try to come up with improved documents.

> Should feature types / names / etc. be added to the proxy dmatrix as part of the iterations?

yes.

> Should labels and/or bounds be set as part of the iterations? If not, are categorical features supported for this class?

yes, it should be part of the iterations. Yes, categorical features are supported.

> If the call to XGQuantileDMatrixCreateFromCallback fails, returning a non-zero code, does the proxy dmatrix still need to be freed manually through XGDMatrixFree?

Yes, everything created successfully should be freed accordingly. If the proxy is created successfully, it should be freed, regardless of how it's used.

> Is there something that would currently work with a regular DMatrix but not with a QuantileDMatrix,

The `QDM` works with the `hist` tree method (default) for training and normal prediction. It's used to conserve memory for training.

> and which would need to be checked on the R side instead of passing things to the C functions?

I usually just let the C functions emit errors.

R-package/R/xgb.DMatrix.R
```diff
@@ -58,19 +77,35 @@ xgb.DMatrix <- function(
   qid = NULL,
   label_lower_bound = NULL,
   label_upper_bound = NULL,
-  feature_weights = NULL
+  feature_weights = NULL,
+  as_quantile_dmatrix = FALSE,
+  ref = NULL,
+  max_bin = NULL
```
trivialfis, line 83: It's probably better to split it into a sub class.

david-cortes:
Now I'm confused.

If I understand it correctly, creating a QuantileDMatrix from a dense input involves setting the proxy dmatrix 4 times.

From some quick tests, it seems to be possible to pass `label` by calling `XGDMatrixSetInfoFromInterface` _after_ the QuantileDMatrix has been created with argument `data` only (at least for cases without categorical data).

If the label were to be set on the proxy dmatrix at every iterator call, wouldn't that result in more work being done compared to setting it after? Aren't there conversions and memory copies happening inside those calls?

Also, just to be sure - if one is supposed to set the feature names and types on the proxy dmatrix at every iterator call, does that mean that one can pass different subsets of columns in different iterations?

**********************

By the way, what would happen in a case in which `XGQuantileDMatrixCreateFromCallback` returns an error code, and then `XGDMatrixFree` after that also returns an error code? Can that happen? How/what should be reported to the user then?

trivialfis:
Let me write some documents for the QDM internal, it's a special case for the iterator interface, which is also used by external memory support. As a result, some of the things seem redundant, but we kept it to simplify the code so that it can work with both use cases without any special handling. So, to answer the question of whether it can accept a subset of samples, yes, but it's for external memory support.

> By the way, what would happen in a case in which XGQuantileDMatrixCreateFromCallback returns an error code, and then XGDMatrixFree after that also returns an error code? Can that happen? 

If you are trying to free the proxy, then should succeed assuming the proxy was successfully created. If you are trying to free the QDM, then it should fail.

> How/what should be reported to the user then?

If `FromCallback` returns -1, then it's likely that something went wrong during iteration, which means an error in the `next` or `reset` function. In this case:
- Report the errors from callbacks. No need to free the unsuccessfully created QDM, (no harm in doing it either since it will fail anyway). See: https://github.com/dmlc/xgboost/blob/b3700bbb3fc057f4e6948a493b92c3e6ec28cc51/src/c_api/c_api.cc#L312 , if the `Create` static method fails, there's no memory leak since there's no pointer being allocated.
- If no error from `reset/next` callbacks, then report the error from `FromCallback` as usual like the original `DMatrix` class.

trivialfis:
>  it can accept a subset of samples, yes, but it's for external memory support.

And yes, the QDM indeed can take on batches on initialization. See `quantile_data_iterator.py` in `demo`. It's lesser known but can be helpful when a user doesn't have the memory capacity to construct the QDM while holding the original data in memory.

trivialfis:
Ref: https://github.com/dmlc/xgboost/issues/9734#issuecomment-1849023783

> Would it be possible to have C functions to create these QuantileDMatrices from array_interface strings for dense/csr data sources + required parameters? Is there some limitation that would make such an approach not work?

it can no longer take on iterators as described in https://github.com/dmlc/xgboost/pull/9864#issuecomment-1851120592.

R-package/R/xgb.DMatrix.R
```diff
@@ -78,9 +102,18 @@ xgb.DMatrix <- function(
...
     handle <- .Call(XGDMatrixCreateFromFile_R, data, as.integer(silent))
   } else if (is.matrix(data)) {
-    handle <- .Call(
-      XGDMatrixCreateFromMat_R, data, missing, as.integer(NVL(nthread, -1))
-    )
+    if (!as_quantile_dmatrix) {
+      handle <- .Call(XGDMatrixCreateFromMat_R, data, missing, nthread)
+    } else {
+      if (!is.null(ref)) {
+        if (!inherits(ref, "xgb.QuantileDMatrix")) {
```
trivialfis, line 109: It can be a normal DMatrix as well.

R-package/src/xgboost_R.cc
```diff
@@ -519,6 +519,97 @@ XGB_DLL SEXP XGDMatrixNumCol_R(SEXP handle) {
...
+  CHECK_CALL(res_code2);
+  CHECK_CALL(res_code1);
+
+  R_SetExternalPtrAddr(ret, out_dmat);
+  R_RegisterCFinalizerEx(ret, _DMatrixFinalizer, TRUE);
+  R_API_END();
+
+  UNPROTECT(1);
+  return ret;
+}
```
trivialfis, line 611: Depending on your roadmap, it might be desirable to implement this in R instead of C++, the data iterator is the common interface to QDM and external memory, I will share some old unpublished documents soon.

david-cortes: I think it'd be better to use a dedicated C++-only route for single-iteration QuantileDMatrix, and then later on implement a customizable DataIterator in R.

R-package/R/xgb.DMatrix.R
david-cortes: Fixed.

R-package/R/xgb.DMatrix.R
david-cortes: Yes, since the code paths will be entirely different, better to have a dedicated function `xgb.QuantileDMatrix` like in the python interface.

trivialfis:
Is this still draft?

david-cortes:
> Is this still draft?

Yes, and I'm going to create a new MR as I see this kind of functionality would need to be coded differently.

david-cortes:
Closing in favor of bigger PR that follows the same approach as in the python interface: https://github.com/dmlc/xgboost/pull/9913

PR #9867, opened 2023-12-11
Flexible find protobuf.

trivialfis:
- Allow using non-config based find package.

I use the protobuf from conda-forge for development, which doesn't have the cmake config file.

### DIFF
--- a/plugin/federated/CMakeLists.txt
+++ b/plugin/federated/CMakeLists.txt
@@ -8,7 +8,7 @@ if(NOT Protobuf_FOUND)
 endif()
 if(NOT Protobuf_FOUND)
   # let CMake emit error
-  find_package(Protobuf CONFIG)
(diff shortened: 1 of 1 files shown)


plugin/federated/CMakeLists.txt
```diff
@@ -1,7 +1,16 @@
...
 set(protobuf_BUILD_SHARED_LIBS TRUE)
-find_package(Protobuf CONFIG REQUIRED)
+
+find_package(Protobuf CONFIG)
+if(NOT Protobuf_FOUND)
+  find_package(Protobuf)
+endif()
+if(NOT Protobuf_FOUND)
+  # let CMake emit error
+  find_package(Protobuf CONFIG)
```
rongou, line 11: Shouldn't this be required?

rongou (approved):

PR #9876, opened 2023-12-12
SYCL inference optimization

razdoburdin:
This PR improves inference speed on sycl devices. Perf measurements demonstrate up to 2x speed up.


trivialfis (approved):
Looks good assuming all tests pass.

Issue #9869, opened 2023-12-11
[R] Predictions with base margin are off by constant number


david-cortes:
ref https://github.com/dmlc/xgboost/issues/9810

It seems in the R interface, adding a base margin to a DMatrix somehow results in the predictions being off by exactly 1.48.

Example:
```r
library(xgboost)
data("agaricus.train")
x <- agaricus.train$data[1:100, ]
y <- agaricus.train$label[1:100]
n_threads = 1
dm_labelled <- xgb.DMatrix(
    data = x,
    label = y,
    nthread = n_threads
)
dm_only_x <- xgb.DMatrix(
    data = x,
    nthread = n_threads
)
dm_w_base_margin <- xgb.DMatrix(
    data = x,
    nthread = n_threads,
    base_margin = rep(0, nrow(x))
)
params <- list(
    objective = "binary:logistic",
    tree_method = "hist",
    nthread = n_threads
)
model <- xgb.train(
    data = dm_labelled,
    params = params,
    nrounds = 5L
)
pred_wo_margin <- predict(model, dm_only_x, outputmargin = TRUE)
pred_w_margin <- predict(model, dm_w_base_margin, outputmargin = TRUE)
print(pred_wo_margin[1] - pred_w_margin[1])
```
```
[1] -1.48
```

In contrast, the parameter seems to work as expected on the python interface:
```python
import numpy as np
rng = np.random.default_rng(seed=123)
X = rng.standard_normal(size=(100,10))
y = rng.integers(2, size=100)
import xgboost as xgb
dmat_labelled = xgb.DMatrix(data=X, label=y)
dmat_only_x = xgb.DMatrix(data=X)
dmat_w_base_margin = xgb.DMatrix(data=X, base_margin=np.zeros(y.shape[0]))
model = xgb.train(
    dtrain=dmat_labelled,
    params={
        "objective" : "binary:logistic",
        "tree_method" : "hist",
    },
    num_boost_round = 5
)
pred_wo_margin = model.predict(dmat_only_x, output_margin=True)
pred_w_margin = model.predict(dmat_w_base_margin, output_margin=True)
print(pred_w_margin[0] - pred_wo_margin[0])
```

trivialfis:
Hmm, there are some inconsistencies between the base margin handling and base score handling, I will take a deeper look.

david-cortes:
[...]

PR #9882, opened 2023-12-13
[doc] Brief introduction to `base_score`.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/9869 .

### DIFF
--- a/doc/tutorials/intercept.rst
+++ b/doc/tutorials/intercept.rst
@@ -1,0 +1,77 @@
+#########
+Intercept
+#########
+
+.. versionadded:: 2.0.0
+
+Since 2.0.0, XGBoost supports estimating the model intercept (named ``base_score``)
+automatically based on targets upon training. The behavior can be controlled by setting
+``base_score`` to a constant value. The following snippet disables the automatic
+estimation:
+
+.. code-block:: python
+
+    import xgboost as xgb
+
+    reg = xgb.XGBRegressor()
+    reg.set_params(base_score=0.5)
+
+Other than the ``base_score``, users can also provide global bias via the data field
+``base_margin``, which is a vector or a matrix depending on the task.
+
+.. code-block:: python
+
+    import xgboost as xgb
+    from sklearn.datasets import make_regression
+
+    X, y = make_regression()
+
+    reg = xgb.XGBRegressor()
+    reg.fit(X, y)
+    # Request for raw prediction
+    m = reg.predict(X, output_margin=True)
+
+    reg_1 = xgb.XGBRegressor()
+    # Feed the prediction into the next model
+    reg.fit(X, y, base_margin=m)
+    reg.predict(X, base_margin=m)
+
+
+It specifies the bias for each sample and can be used for stacking an XGBoost model on top
+of other models, see :ref:`sphx_glr_python_examples_boost_from_prediction.py` for a worked
+example. When ``base_margin`` is specified, it automatically overrides the ``base_score``
+parameter. If you are stacking XGBoost models, then the usage should be relatively
+straightforward, with the previous model providing raw prediction and a new model using
+the prediction as bias. For more customized inputs, users need to take extra care of the
+link function. Let :math:`F` be the model and :math:`g` be the link function, since
+``base_score`` is overridden when sample-specific ``base_margin`` is available, we will
+omit it here:
+
+.. math::
+
+   g(E[y_i]) = F(x_i)
+
+
+When base margin :math:`b` is provided, it's added to the raw model output :math:`F`:
+
+.. math::
+
+   g(E[y_i]) = F(x_i) + b_i
+
+and the output of the final model is:
+
+
+.. math::
+
+   g^{-1}(F(x_i) + b_i)
+
+Using the gamma deviance objective ``reg:gamma`` as an example, which has a log link
+function, hence:
+
+.. math::
+
+   \ln{(E[y_i])} = F(x_i) + b_i \\
+   E[y_i] = \exp{(F(x_i) + b_i)}
+
+As a result, if you are feeding outputs from models like GLM with a corresponding
+objective function, make sure the outputs are not yet transformed by the inverse link.
--- a/doc/tutorials/intercept.rst
+++ b/doc/tutorials/intercept.rst
@@ -16,6 +16,9 @@ estimation:
     reg = xgb.XGBRegressor()
     reg.set_params(base_score=0.5)
 
+In addition, here 0.5 represents the value after applying the inverse link function. See
+the end of the document for a description.
+
 Other than the ``base_score``, users can also provide global bias via the data field
 ``base_margin``, which is a vector or a matrix depending on the task.
 
@@ -75,3 +78,20 @@ function, hence:
 
 As a result, if you are feeding outputs from models like GLM with a corresponding
 objective function, make sure the outputs are not yet transformed by the inverse link.
+
+In the case of ``base_score`` (intercept), if you access the estimation through
+:py:meth:`~xgboost.Booster.save_config`, XGBoost returns the value
+:math:`g^{-1}(base_score)`. With logistic regression and the logit link function, given
+the ``base_score`` as 0.5, :math:`logit(0.5) = 0.0` is added to the raw model output:
+
+.. math::
+
+   E[y_i] = g^{-1}{(F(x_i) + g(intercept))}
+
+This is more intuitive if you remove the model and consider only the intercept, which is
+estimated before the model is fitted:
+
+.. math::
+
+   E[y_i] = g^{-1}{g(intercept))} \\
+   E[y_i] = intercept
--- a/doc/tutorials/intercept.rst
+++ b/doc/tutorials/intercept.rst
@@ -20,7 +20,9 @@ In addition, here 0.5 represents the value after applying the inverse link funct
 the end of the document for a description.
 
 Other than the ``base_score``, users can also provide global bias via the data field
-``base_margin``, which is a vector or a matrix depending on the task.
+``base_margin``, which is a vector or a matrix depending on the task. With multi-output
+and multi-class, the ``base_margin`` is a matrix with size ``(n_samples, n_targets)`` or
+``(n_samples, n_classes)``.
 
 .. code-block:: python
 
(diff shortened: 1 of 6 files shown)


trivialfis:
cc @david-cortes

david-cortes:
Thanks for looking into this.

However, I am still feeling that the docs do not explain what `base_score` is or how it is obtained.

Suppose that I fit a model to the first 100 rows of the agaricus data:

```r
library(xgboost)
data("agaricus.train")
x <- agaricus.train$data[1:100, ]
y <- agaricus.train$label[1:100]
dm_labelled <- xgb.DMatrix(data = x, label = y)
model <- xgb.train(
    data = dm_labelled,
    params = list(objective = "binary:logistic"),
    nrounds = 1L
)
```

In this case, 13% of the rows belong to the "positive" class:
```r
print(mean(y))
```
```
[1] 0.13
```

The model's `base_score` as returned by the json is as follows:
```r
model_json <- jsonlite::fromJSON(xgb.config(model))
base_score <- as.numeric(model_json$learner$learner_model_param$base_score)
print(base_score)
```
```
[1] 0.1854274
```

It does seem to be the case that `base_margin` overrides the intercept, and gets added to the raw scores as described in this PR.

In this case however, the intercept that gets added seems to correspond to what would be obtained by applying the link function to `base_score`, **not** to `base_score` itself:

$$
\log(\frac{0.1854274}{1 - 0.1854274}) = -1.48
$$

```r
dm_only_x <- xgb.DMatrix(data = x)
dm_w_base_margin <- xgb.DMatrix(data = x, base_margin = rep(0, nrow(x)))
pred_wo_margin <- predict(model, dm_only_x, outputmargin = TRUE)
pred_w_margin <- predict(model, dm_w_base_margin, outputmargin = TRUE)
print(pred_wo_margin[1] - pred_w_margin[1])
```
```
[1] -1.48
```

Note also that the optimal intercept in this case would be a slightly different number than what the model estimated:
```r
optimal_intercept <- log(sum(y == 1)) - log(sum(y == 0))
print(optimal_intercept)
```
```
[1] -1.900959
```

doc/tutorials/intercept.rst
```diff
@@ -74,4 +77,21 @@ function, hence:
    E[y_i] = \exp{(F(x_i) + b_i)}
 
 As a result, if you are feeding outputs from models like GLM with a corresponding
-objective function, make sure the outputs are not yet transformed by the inverse link.
\ No newline at end of file
+objective function, make sure the outputs are not yet transformed by the inverse link.
+
+In the case of ``base_score`` (intercept), if you access the estimation through
+:py:meth:`~xgboost.Booster.save_config`, XGBoost returns the value
+:math:`g^{-1}(base_score)`. With logistic regression and the logit link function, given
```
trivialfis, line 84: Incorrect, will fix.

doc/tutorials/intercept.rst
```diff
@@ -74,4 +77,21 @@ function, hence:
...
+
+   E[y_i] = g^{-1}{(F(x_i) + g(intercept))}
+
+This is more intuitive if you remove the model and consider only the intercept, which is
+estimated before the model is fitted:
+
+.. math::
+
+   E[y_i] = g^{-1}{g(intercept))} \\
+   E[y_i] = intercept
```
trivialfis, line 97: Need to mention how it's obtained.

david-cortes:
Thanks again for looking into it.

Would be ideal if you could also update the docs for `base_score` in the [parameters section](https://xgboost.readthedocs.io/en/stable/parameter.html).

There's still a few other issues that could be worth mentioning in this doc:
* What's the behavior if the user supplies `base_score` as a training parameter **AND** passes a DMatrix with `base_margin` to `xgb.train`? It seems it then keeps the `base_score` but doesn't apply it to that training data.
* While `base_score` is restricted to a single number, can `base_margin` have multiple dimensions if e.g. one is using a multi-output model? It seems to work as expected in python, so I'm thinking of updating the R interface to allow for it.

I am also thinking: if the current behavior around `base_score` is due to the old binary model serialization format, shouldn't it be possible to introduce a new field in the model, say "intercept", which would be auto-filled from "base_score" if not present in a serialized model, and otherwise be set to the actual intercept and allow per-output intercepts for e.g. multi-output and multinomial logistic?

trivialfis:
> What's the behavior if the user supplies base_score as a training parameter AND passes a DMatrix with base_margin to xgb.train? It seems it then keeps the base_score but doesn't apply it to that training data.

The base margin overrides the base score, but this is not "by design", it's more like a user error.

> While base_score is restricted to a single number, can base_margin have multiple dimensions if e.g. one is using a multi-output model? 

Yes, will update the doc.

>  if the current behavior around base_score is due to the old binary model serialization format

Yes, it's limited by the binary model serialization and we can't use a vector value.

> which would be auto-filled from "base_score" if not present in a serialized model, and otherwise be set to the actual intercept and allow per-output intercepts for e.g. multi-output and multinomial logistic?

I thought about it, but the current serialization code is just messy and I don't want to modify the binary model format in any way, considering that we are deprecating it anyway.

trivialfis:
Also as a reminder, I should make a specialization for binary classification to use frequency instead of Newton.

david-cortes:
Thanks for the info. I guess that clarifies the behavior of the intercept around the logistic objective.

I'm still curious about this method of estimation and the usages though.

Suppose now that the objective is `count:poisson`, and I create a small example where there's only an intercept:
```python
import numpy as np
import xgboost as xgb
rng = np.random.default_rng(seed=123)
X = rng.standard_normal(size=(100,10))
y = 1 + rng.poisson(10, size=100)
model = xgb.train(
    dtrain=xgb.DMatrix(X, label=y),
    params={"objective" : "count:poisson", "gamma" : 1e4},
    num_boost_round=1,
)
pred = model.predict(xgb.DMatrix(X))
import json
base_score = json.loads(
    model.save_raw("json").decode()
)["learner"]["learner_model_param"]["base_score"]
```

In this case, we have:
```python
np.mean(y)
```
```
10.74
```

I understand that here the function should work as

$$
\hat{y} = \exp(F)
$$

Hence, `base_score` should be `mean(y)` and the intercept should be the logarithm of that (inverse of the link function), but:
```python
float(base_score)
```
```
126.05785
```

```python
np.log(float(base_score))
```
```
4.8367409285690615
```

Then, the predictions are constant, but they do not seem to correspond to the intercept, nor to be close to the actual values of `y`:
```python
pred[0]
```
```
109.99812
```

```python
model.predict(xgb.DMatrix(X), output_margin=True)[0]
```
```
4.7004633
```

What's happening in this case?

trivialfis:
The poisson case is a result of poor Newton optimization. We use 0 as the starting point, which has a significant error compared to the actual mean.

david-cortes:
But if that was the actual intercept, even if off by some amount, shouldn't the predictions be equal to `base_score`, since the model doesn't have any further conditions adding towards it?

Also, if one applies one-step Newton starting from zero, even then, the results would not quite match with the earlier `base_score`:
```python
import numpy as np
rng = np.random.default_rng(seed=123)
X = rng.standard_normal(size=(100,10))
y = 1 + rng.poisson(10, size=100)

def grad_poisson(x):
    return np.exp(x) - y

def hess_poisson(x):
    return np.exp(x)

x0 = np.zeros_like(y)
x1_newton = -hess_poisson(x0).sum() / grad_poisson(x0).sum()
print(x1_newton)
```
```
0.1026694045174538
```

By the way, if you are applying the inverse of the link function to `base_score`, then the optimal `base_score` should be the mean of `y` whenever that link function is what GLM theory calls a "cannonical" link function for the given likelihood.

That applies to: `reg:squarederror`, `reg:gamma`, `binary:logistic`, `count:poisson`. I'm not 100% sure but I think it also applies to tweedie and AFT.

trivialfis:
> But if that was the actual intercept, even if off by some amount, shouldn't the predictions be equal to base_score, since the model doesn't have any further conditions adding towards it?

There's still an extra tree stump after the base_score since there's at least one iteration being fitted. Hopefully, the script helps illustrate:
``` python
import json

import numpy as np
import xgboost as xgb

rng = np.random.default_rng(seed=123)
X = rng.standard_normal(size=(100, 10))
y = 1 + rng.poisson(10, size=100)
Xy = xgb.DMatrix(X, label=y)

model = xgb.train(
    dtrain=Xy,
    params={"objective": "count:poisson", "min_split_loss": 1e5},  # no split
    num_boost_round=1,  # but still has one tree stump
    evals=[(Xy, "train")],
)

config = json.loads(model.save_config())["learner"]["learner_model_param"]
intercept = float(config["base_score"])
print("intercept:", intercept)

jmodel = json.loads(str(model.save_raw(raw_format="json"), "ascii"))
jtrees = jmodel["learner"]["gradient_booster"]["model"]["trees"]
assert len(jtrees) == 1
# split_condition actually means weight for leaf, and split threshold for internal split
# nodes
assert len(jtrees[0]["split_conditions"]) == 1
leaf_weight = jtrees[0]["split_conditions"][0]
print(leaf_weight)  # raw prediction

output_margin = np.log(intercept) + leaf_weight
predt_0 = np.exp(output_margin)

print(np.log(intercept))
print(predt_0)

predt_1 = model.predict(xgb.DMatrix(X))
print(predt_1[0])
np.testing.assert_allclose(predt_0, predt_1)
```

> Also, if one applies one-step Newton starting from zero, even then, the results would not quite match with the earlier

That's caused by an old parameter to control step size, which I want to remove but did not get the approval: https://github.com/dmlc/xgboost/pull/7267

@RAMitchell would be great if we could revisit that PR.

> By the way, if you are applying the inverse of the link function to base_score, then the optimal base_score should be the mean of y whenever that link function is what GLM theory calls a "cannonical" link function for the given likelihood.

Thank you for sharing! I will learn more about it. I will do some searching. On the other hand, I think the canonical link for gamma is the inverse link, even though the log link is often used in practice. Not sure if the mean value still applies.

david-cortes:
Thanks for explaining the situation here.

It still leaves me wondering about one last detail though: in the case above, is it that the tree had no split conditions and it created a terminal node to conform to a tree structure, or is it that trees can have their own intercept regardless of whether they have subnodes or not?

***********************

About gamma objective: you are right, my mistake - it's not using the cannonical link.

That being said, I am fairly certain that the optimal intercept is still equal to the logarithm of the mean.

I haven't checked any proofs but here's a small example to support the claim:
```python
import numpy as np
from sklearn.metrics import mean_gamma_deviance
from scipy.optimize import minimize

rng = np.random.default_rng(seed=123)
y = rng.standard_gamma(5, size=100)

def fx(c):
    yhat = np.exp(np.repeat(c, y.shape[0]))
    return mean_gamma_deviance(y, yhat)

x0 = 0.
res = minimize(fx, x0)["x"]
np.testing.assert_almost_equal(res, np.log(y.mean()))
```

trivialfis:
Open an issue to keep track of the status https://github.com/dmlc/xgboost/issues/9899

trivialfis:
Is there anything I need to modify from the documentation s perspective?

david-cortes:
> Is there anything I need to modify from the documentation s perspective?

I cannot think of anything that'd be missing after all the last changes. Thanks for this addition. Would be ideal to merge now.

PR #9883, opened 2023-12-13
[Draft] Overview of scaling effort for developers.

trivialfis:
I wrote this a long time ago, it's more or less just me trying to think about the API and potential work I can do instead of being useful. Maybe it can provide some context now that @david-cortes is working on the iterator interface for R.

For a smaller demo on how iterator works with external memory, see `demo/c-api/external-memory` instead. The QDM works in a similar way with an additional parameter `ref`.


trivialfis:
I will drop the PR if it's not informative, not good at writing documents. ;-(

PR #9893, opened 2023-12-14
[jvm-packages] Fix POM for xgboost-jvm metapackage

hcho3:
Ref: #9787 #9797

Maven cannot expand the variable `${scala.binary.version}` in the module name.

Also update `dev/prepare_jvm_release.py`:
* The script can now run on any machine, since all native libraries will be downloaded from the CI pipeline.
* To upload packages targeting Scala 2.13, the maintainer needs to use `sed` to replace all occurrences of `xgboost-jvm_2.12` with `xgboost-jvm_2.13`.

PR #9897, opened 2023-12-15
Update change_scala_version.py to also change scala.version property

hcho3:
Follow up to https://github.com/dmlc/xgboost/pull/9893

The script `change_scala_version.py` should also update the `scala.version` property, since the user may not be use the profile `scala-2.12` when running Maven.

cc @trivialfis @wbo4958

### DIFF
--- a/dev/change_scala_version.py
+++ b/dev/change_scala_version.py
@@ -29,11 +29,20 @@ def main(args):
             replaced_scalaver = False
             replaced_scala_binver = False
             for line in lines:
-                line = re.sub(
-                    r"<artifactId>xgboost-jvm_[0-9\.]*",
-                    f"<artifactId>xgboost-jvm_{scala_ver}",
-                    line,
-                )
+                for artifact in [
+                    "xgboost-jvm",
+                    "xgboost4j",
+                    "xgboost4j-gpu",
+                    "xgboost4j-spark",
+                    "xgboost4j-spark-gpu",
+                    "xgboost4j-flink",
+                    "xgboost4j-example",
+                ]:
+                    line = re.sub(
+                        f"<artifactId>{artifact}_[0-9\\.]*",
+                        f"<artifactId>{artifact}_{scala_ver}",
+                        line,
+                    )
                 # Only replace the first occurrence of scala.version
                 if not replaced_scalaver:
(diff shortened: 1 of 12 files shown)


trivialfis:
I'm not familiar with the ecosystem and cannot comment on this.

wbo4958:
I got below error using the script.
``` console
sh.ErrorReturnCode_2: 

  RAN: /usr/bin/sed -i 's/<artifactId>xgboost-jvm_[0-9\.]*/<artifactId>xgboost-jvm_2.13/g' -e 's/<artifactId>xgboost-jvm_[0-9\.]*/<artifactId>xgboost-jvm_2.13/g' -e '0,/<scala.version>/ s/<scala.version>[0-9\.]*/<scala.version>2.13.11/' -e '0,/<scala.binary.version>/ s/<scala.binary.version>[0-9\.]*/<scala.binary.version>2.13/' jvm-packages/pom.xml

  STDOUT:


  STDERR:
/usr/bin/sed: can't read s/<artifactId>xgboost-jvm_[0-9\.]*/<artifactId>xgboost-jvm_2.13/g: No such file or direct
```

hcho3:
Are you using a Mac? I was able to run the script on a Ubuntu box without any issue.

wbo4958:
I was using Ubuntu to do the test

hcho3:
@wbo4958 Can you try again with the updated script?

wbo4958:
It can work. But I'm curious seems it only updated the xgboost_jvm instead of updating the sub modules,

For example, the artifactId in xgboost4j-spark has not been changed, is that expected?
``` xml
<artifactId>xgboost4j-spark_${scala.binary.version}</artifactId>
```

hcho3:
@wbo4958 This is fine. The property `scala.binary.version` is correctly loaded from the parent `xgboost4j-jvm_2.12`. I have tested it with the command
```
# Use staging repository https://oss.sonatype.org/content/repositories/mldmlc-1114
mvn dependency:get -DgroupId=ml.dmlc -DartifactId=xgboost4j_2.13 -Dpacking=jar \
    -Dversion=2.0.3 -DremoteRepositories=https://oss.sonatype.org/content/repositories/mldmlc-1114
```

Issue #9865, opened 2023-12-10
[R] Checking whether DMatrix has field implies making full data copies of said field


david-cortes:
Follow up from https://github.com/dmlc/xgboost/pull/9863
ref https://github.com/dmlc/xgboost/issues/9810

In methods such as `print.xgb.DMatrix`:
https://github.com/dmlc/xgboost/blob/562352101d5dd1630eded1522b9b4ed9b07e8692/R-package/R/xgb.DMatrix.R#L517

The code checks whether a DMatrix has a given field by retrieving it.

This causes a full (maybe even two?) data copy of the underlying field being retrieved, but the data there is not used: it's only being checked that it exists.

Would be more efficient to use some way of checking that the DMatrix has a field taht would not involve copying all of the data in that field.

trivialfis:
We don't have this method yet. For the print method you referenced, I think it's easy to just get all meta info and filter it by null values. We need all the available fields anyway.

As for the AFT PR, I think we can merge it for now, but wouldn't it be better for the long term that we can avoid processing the `DMatrix` and move toward the high-level interface? If so, we can just get the label/bounds from user inputs instead of the DMatrix.

PR #9901, opened 2023-12-15
[R] Refactor logic for DMatrix fields

david-cortes:
fixes https://github.com/dmlc/xgboost/issues/9865

This PR refactors a bit the logic for DMatrix fields:
* It removes forced coercions, since those should be handled automatically by the new 'array_interface' when the dtype is supported.
* It replaces `length` with `NROWS` in case later on multi-dimensional inputs become supported. Right now, if the input for some field has incorrect dimension, an error will be thrown from the C++ side anyway.
* It adds a function to check whether a DMatrix object has a given field without generating a memory copy of said field, and uses this function where appropriate.
* It adds a visible note where possible when the DMatrix has a blank pointer, which can happen when one calls R's serialization methods on it. Note that before this PR, since methods like `print` call `getinfo`, an error would have been thrown from the C++ side instead.


R-package/R/xgb.DMatrix.R
```diff
@@ -199,6 +202,35 @@ xgb.DMatrix <- function(
...
+#' # Fields can be added after construction
+#' setinfo(dm, "label", 1:5)
+#' xgb.DMatrix.hasinfo(dm, "label")
+#' @export
+xgb.DMatrix.hasinfo <- function(object, info) {
+  if (!inherits(object, "xgb.DMatrix")) {
+    stop("Object is not an 'xgb.DMatrix'.")
+  }
+  if (.Call(XGCheckNullPtr_R, object)) {
+    warning("xgb.DMatrix object is invalid. Must be constructed again.")
```
trivialfis, line 228: Can it happen?

R-package/R/xgb.DMatrix.R
```diff
@@ -418,26 +450,32 @@ setinfo <- function(object, ...) UseMethod("setinfo")
 #' @rdname setinfo
 #' @export
 setinfo.xgb.DMatrix <- function(object, name, info, ...) {
+  .internal.setinfo.xgb.DMatrix(object, name, info, ...)
+  attr(object, "fields")[[name]] <- TRUE
```
trivialfis, line 454: Niche case, what happens if I set a Null value?

R-package/R/xgb.DMatrix.R
david-cortes: Yes, it can happen when the object is serialized and then de-serialized, for example:
```r
library(xgboost)
x <- matrix(1:10, nrow=5)
dm <- xgb.DMatrix(x)
dm2 <- dm |> serialize(NULL) |> unserialize()
xgb.DMatrix.hasinfo(dm2, "label")
```
```
[1] FALSE
Warning message:
In xgb.DMatrix.hasinfo(dm2, "label") :
  xgb.DMatrix object is invalid. Must be constructed again.
```

R-package/R/xgb.DMatrix.R
david-cortes: Then it would error out from the C++ side:
```
Error in .internal.setinfo.xgb.DMatrix(object, name, info, ...) :
The length of labels must equal to the number of rows in the input data
```

trivialfis (approved):

PR #9902, opened 2023-12-15
[R] Avoid memory copies in `predict`

david-cortes:

This PR modifies the reshaping code in the `predict` method to avoid generating deep memory copies when not needed.

This passes the tests but I'm not 100% sure that very possible case is covered so would be ideal if a maintainer could take a deep look at the changes.


trivialfis:
Can this PR wait? I'm on PTO at the moment and reply might be slow. I would like to take a deeper look into R predict function.

david-cortes:
> Can this PR wait? I'm on PTO at the moment and reply might be slow. I would like to take a deeper look into R predict function.

Sure, no hurry here.

trivialfis:
I think we can probably lower the entire reshaping process to C, what do you think?

david-cortes:
> I think we can probably lower the entire reshaping process to C, what do you think?

Do you mean inside `XGBoosterPredictFromDMatrix`, or inside the R wrapper over it? If the former, I think it'd be quite helpful, especially if it adds a parameter 'column_major' or similar in the JSON. The latter, I don't think there'd be much of a performance difference with the R code here.

trivialfis:
> I don't think there'd be much of a performance difference with the R code here.

It's more about easier-to-follow code flow instead of performance. XGBoost's prediction function is mostly optimized for batch prediction where the input size is non-trivial, for latency-sensitive applications one needs to use a specialized inference library, which is significantly different.

david-cortes:
> > I don't think there'd be much of a performance difference with the R code here.
> 
> It's more about easier-to-follow code flow instead of performance. XGBoost's prediction function is mostly optimized for batch prediction where the input size is non-trivial, for latency-sensitive applications one needs to use a specialized inference library, which is significantly different.

Yes, perhaps the R interface could be made simpler if all this manipulation were moved to the C++ side with an option for column-major outputs, but that'd be a very different type of PR. Are you willing to introduce that kind of functionality?

I by the way think you'll find plenty of R users serving xgboost models for single-row predictions through `predict.xgb.Booster`.

trivialfis (approved):

trivialfis:
We will get to the predict function after changes for inplace prediction.

PR #9903, opened 2023-12-18
[R] Catch C++ exceptions

david-cortes:

This PR adds a catch clause for C++ exceptions in the R interface.

Currently, the code involves constructing objects such as `std::vector`, which can potentially throw exceptions like `std::bad_alloc`. At the moment, those exceptions would not be catched anywhere, and since the functions have C linkage, this means that such exceptions could potentially result in memory corruption and similar from the user side when triggered.

Note that producing an R error involves a C long jump, so any C++ objects that deal with memory outside of the function stack, including the exception object, need to be destructed before throwing the error.

There's one thing I wasn't quite sure about though: there's a built-in error class `dmlc::Error` which can contain error messages. Right now, those are passed to the R error function by accessing `.what()`, but I am not sure if the memory for that exception also needs to be freed with a similar workaround as the one introduced here for standard C++ exceptions.

### DIFF
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -138,21 +138,37 @@ SEXP SafeMkChar(const char *c_str, SEXP continuation_token) {
 }
 }  // namespace
 
+char cpp_ex_msg[256];
+struct RRNGStateController {
+  RRNGStateController() {
+    GetRNGstate();
+  }
+
+  ~RRNGStateController() {
+    PutRNGstate();
+  }
+};
+
 /*!
  * \brief macro to annotate begin of api
  */
 #define R_API_BEGIN()                           \
-  GetRNGstate();                                \
-  try {
+  try {                                         \
+    RRNGStateController rng_controller{};
 /*!
  * \brief macro to annotate end of api
  */
 #define R_API_END()                             \
   } catch(dmlc::Error& e) {                     \
-    PutRNGstate();                              \
-    error(e.what());                            \
+    Rf_error("%s", e.what());                   \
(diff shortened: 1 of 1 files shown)


R-package/src/xgboost_R.cc
```diff
@@ -138,21 +138,37 @@ SEXP SafeMkChar(const char *c_str, SEXP continuation_token) {
...
+    Rf_error("%s", e.what());                   \
+  } catch(std::exception &e) {                  \
+    std::strncpy(cpp_ex_msg, e.what(), 256);    \
+    goto throw_cpp_ex_as_R_err;                 \
   }                                             \
-  PutRNGstate();
+  if (false) {                                  \
+    throw_cpp_ex_as_R_err:                      \
+    Rf_error("%s", cpp_ex_msg);                 \
+  }
```
trivialfis, line 171: is this for debugging?

david-cortes: No, it's to show the C++ exception message as an R error.

david-cortes: To clarify: when an exception is thrown, it gets to that condition through a `goto` statement. Would be good to hear other ideas if you think there's a better way to do this - I myself couldn't think of any other way.

trivialfis: Would be great if leave some comments for explaining these few lines of code, it's not very intuitive to see some code being meaningful inside a condition block that's not reachable under normal c control flow.

david-cortes: Added a comment above.

trivialfis (approved):

PR #9905, opened 2023-12-19
Correct name of function for setting data frames in proxy dmatrix

david-cortes:
In a previous PR: https://github.com/dmlc/xgboost/pull/9828

A C function `XGProxyDMatrixSetDataColumnar` was introduced, but what was added to the public C header was instead `XGProxyDMatrixSetDataCudaColumnar`, which overlaps with a previous function name.

This PR renames it to match with the function which was introduced in the PR.

### DIFF
--- a/include/xgboost/c_api.h
+++ b/include/xgboost/c_api.h
@@ -532,7 +532,7 @@ XGProxyDMatrixSetDataCudaArrayInterface(DMatrixHandle handle,
  *
  * @return 0 when success, -1 when failure happens
  */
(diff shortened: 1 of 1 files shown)


david-cortes:
Failing job seems unrelated:

> Error: Unexpected HTTP status code '500' when retrieving versions from 'https://static.azul.com/zulu/bin/'. <!DOCTYPE html>

trivialfis:
Those are different functions, I will take a look.

trivialfis (approved):

Issue #9875, opened 2023-12-12
[R] More compact Roxygen


mayer79:
Roxygen is now closer to Markdown. E.g., instead of the verbose

```
 \itemize{
   \item \code{Features}
   ...
}
```

one can write

```

- `Features`
- ...
```

This makes the docstrings easier to read. 

I can work on this, if nobody has any objections.

- [x] callbacks.R
- [x] utils.R
- [ ] xgb.Booster.R
- [ ] xgb.config.R
- [ ] xgb.create.features.R  
- [ ] xgb.cv.R
- [ ] xgb.DMatrix.R         
- [ ] xgb.DMatrix.save.R    
- [ ] xgb.dump.R    
- [x] xgb.ggplot.R  
- [x] xgb.importance.R
- [x] xgb.load.R
- [x] xgb.load.raw.R
- [x] xgb.model.dt.tree.R
- [x] xgb.plot.deepness.R
- [x] xgb.plot.importance.R
- [x] xgb.plot.multi.trees.R 
- [x] xgb.plot.shap.R     
- [x] xgb.plot.tree.R
- [ ] xgb.save.R
- [x] xgb.save.raw.R
- [ ] ~xgb.serialize.R~
- [ ] xgb.train.R
- [ ] ~xgb.unserialize.R~
- [ ] xgboost.R

trivialfis:
> I can work on this, if nobody has any objections.

Thank you for volunteering! Feel free to open a PR.

trivialfis:
@mayer79 Try to use this in the XGB repo, it's using roxygen:
```
python ./tests/ci_build/test_r_package.py --task=doc
```

Yes, it does build the package.

david-cortes:
> @trivialfis : Beginners question: How do you generate the rd help files in a convenient way?
> 
> For other packages, I usually run `devtools::document()` or `devtools::document("R-package")`.. But that starts building the library and stops with
> 
> "fatal error: common.h: No such file or directory #include <dmlc/common.h"

I think you might have cloned the repository without the submodules. You need to use `clone --recursive`.

Maybe you could try something like this:
```shell
git submodule update --init --recursive
```

jameslamb:
> `python ./tests/ci_build/test_r_package.py --task=doc`

You can avoid needing to rebuild + reinstall every time by passing `load = 'installed'`, like this:

```shell
# build and install the library for the first time
R CMD INSTALL --with-keep.source ./R-package

# re-generate the docs
[...]

PR #9906, opened 2023-12-19
[R] improve docstrings for "xgb.Booster.R"

mayer79:
This PR introduces Markdown docstrings. It is the first step towards #9875. 

1. To keep the PR not overly large, the changes are focussing on the file "xgb.Booster.R" which mainly contains the `predict()` function.

2. Besides switching to Markdown, I am also cleaning small inconsistencies in the docstrings, change some wordings, and format some examples a bit neater.

3. Quite unrelated, the third commit removes a warning in an example related to an inexisting parameter `silent`.

I am aware that such reviews cost time. At least, I am not changing anything in the functions.

### DIFF
--- a/R-package/DESCRIPTION
+++ b/R-package/DESCRIPTION
@@ -63,7 +63,8 @@ Imports:
     Matrix (>= 1.1-0),
     methods,
     data.table (>= 1.9.6),
-    jsonlite (>= 1.0),
+    jsonlite (>= 1.0)
+Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.2.3
 Encoding: UTF-8
 SystemRequirements: GNU make, C++17
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -177,7 +177,7 @@ xgb.Booster.complete <- function(object, saveraw = TRUE) {
 #' @param object Object of class `xgb.Booster` or `xgb.Booster.handle`.
 #' @param newdata Takes `matrix`, `dgCMatrix`, `dgRMatrix`, `dsparseVector`,
 #'        local data file, or `xgb.DMatrix`.
-#'        For single-row predictions on sparse data, it is recommended to use the CSR format. 
+#'        For single-row predictions on sparse data, it is recommended to use the CSR format.
 #'        If passing a sparse vector, it will take it as a row vector.
 #' @param missing Only used when input is a dense matrix. Pick a float value that represents
 #'        missing values in data (e.g., 0 or some other extreme value).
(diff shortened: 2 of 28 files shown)


trivialfis:
cc @david-cortes .

trivialfis:
Hi @mayer79 , thank you for the work on using markdown syntax, could you please take a look into the CI failures?

mayer79:
One problem was from existing square brackets in the documentation. They are parsed differently under Markdown. This should now be fixed.

Here is an example how the help will show to useres:

### Old

image

### New

image

trivialfis (approved):
Looks good to me, would be great if @david-cortes could help take a look as well.

trivialfis:
Unrelated note, would be interesting if roxygen can output markdown without converting it into rd, we can plug the document into sphinx then XGBoost's document site can display R API just like the Python API.

david-cortes:
LGTM.

While you're at it, perhaps you could also look into misrendered things like this:
image

I think it was there before, but could also benefit from correct tags and formatting.

david-cortes:
> Unrelated note, would be interesting if roxygen can output markdown without converting it into rd, we can plug the document into sphinx then XGBoost's document site can display R API just like the Python API.

I don't know if that's possible, but looks like the R docs can otherwise be rendered as a similar-looking HTML (don't know how though). For example, lightgbm has a page for R docs which is auto-generated from the roxygen code: https://lightgbm.readthedocs.io/en/latest/R/reference/

mayer79:
> While you're at it, perhaps you could also look into misrendered things like this:
> 
> I think it was there before, but could also benefit from correct tags and formatting.

In what .rd file is it? The aim of this and upcoming PRs is definitively to also fix things liket his.

david-cortes:
> > While you're at it, perhaps you could also look into misrendered things like this:
> > I think it was there before, but could also benefit from correct tags and formatting.
> 
> In what .rd file is it? The aim of this and upcoming PRs is definitively to also fix things liket his.

That particular screenshot was from `xgb.ggplot.shap.summary`.

trivialfis:
> I don't know if that's possible, but looks like the R docs can otherwise be rendered as a similar-looking HTML (don't know how though).

We can convert Rmd files into md files, then consume it using sphinx using knit, just wondering if that's possible with Rd files as well. Not important at this point, just something nice to have in the future. Apologies for the distraction.

mayer79:
@trivialfis 

There is a git pages workflow that uses the {pkgdown} package to generate HTML files. Here is an example such a help page of such a function: https://modeloriented.github.io/shapviz/reference/sv_dependence.html

mayer79:
@david-cortes : I will go over this help file in the next PR. Hope to fix the majority of such things as well.

trivialfis:
Thank you for sharing, I will look into it.

Issue #9472, opened 2023-08-11
Deprecate text file input.


trivialfis:
I propose we deprecate the use of text file input and remove the text parsers in XGBoost, including the `libsvm` parser and `csv` parser from dmlc core. Nowadays, there's a wealth of third-party libraries focus on feature engineering that can handle these formats with high efficiency. Loading the data inside XGBoost does not provide much value as users are likely need to perform tasks like cross-validation and hyper-parameter optimization.

At the moment, there are three use cases for the text input:
- CLI. I propose the removal of the CLI in https://github.com/dmlc/xgboost/issues/9471 .
- External memory: We have largely replaced the external memory with a custom data iterator. Even with text input, the underlying implementation uses a data iterator.
- Federated learning: I believe we will move toward memory input as we progress for better integration with frameworks like nvflare.

trivialfis:
Reminder: remove any reference in document including the c tutorial.

saxena-vinayak36:
I think I can contribute to this aspect

trivialfis:
I think the work to deprecate it is onerous, might not be a good place to start your first PR. :-)

PR #9913, opened 2023-12-20
[R] Add data iterator, quantile dmatrix, and missing `feature_types`

david-cortes:
closes https://github.com/dmlc/xgboost/pull/9864

(note that it has some overlap with https://github.com/dmlc/xgboost/pull/9829 - whichever is merged first will create conflicts in the other)

This PR adds R wrappers over the following:
* Data iterators, for building DMatrix objects from external memory.
* QuantileDMatrix, built using the same data iterator from the R side.
* Parameter `feature_types` in DMatrix construction.

Additionally, it updates the docs for `feature_types` in the python interface in order to reflect what the current code does.


R-package/R/xgb.DMatrix.R
```diff
@@ -195,10 +299,491 @@ xgb.DMatrix <- function(
...
+#' (i.e. taking it back to the first batch, called before and after the sequence of batches
+#' has been consumed).
+#'
+#' Note that, after resetting the iterator, the batches will be accessed again, so the same data
+#' (and in the same order) must be passed in subsequent iterations.
+#' @return An 'xgb.DataIterator' object, containing the same inputs supplied here, which can then
+#' be passed to \link{xgb.ExternalDMatrix}.
+#' @seealso \link{xgb.ExternalDMatrix}, \link{xgb.ProxyDMatrix}.
+#' @export
+xgb.DataIterator <- function(env = new.env(), f_next, f_reset) {
```
trivialfis, line 435: I will look closer tomorrow, but in the meanwhile, let's keep the naming consistent between R and Python.

david-cortes: Renamed to `xgb.DataIter`.

R-package/NAMESPACE
```diff
@@ -32,6 +32,10 @@ export(slice)
 export(xgb.DMatrix)
 export(xgb.DMatrix.hasinfo)
 export(xgb.DMatrix.save)
+export(xgb.DataIter)
+export(xgb.ExternalDMatrix)
+export(xgb.ProxyDMatrix)
```
trivialfis, line 37: I'm not sure if we need to export the proxy DMatrix, users are unlikely to interact with it.

R-package/NAMESPACE
```diff
@@ -32,6 +32,10 @@ export(slice)
 export(xgb.DMatrix)
 export(xgb.DMatrix.hasinfo)
 export(xgb.DMatrix.save)
+export(xgb.DataIter)
+export(xgb.ExternalDMatrix)
```
trivialfis, line 36: Let's not expose this as an independent class for now. One day we might support external storage with QuantileDMatrix.

What do you think if we make the `data` parameter of `DMatrix` to be compatible with the iterator?

R-package/R/xgb.DMatrix.R
```diff
@@ -1,13 +1,41 @@
...
-#'        a \code{dsparseVector} object (only when making predictions from a fitted model, will be
-#'        interpreted as a row vector), or a character string representing a filename.
+#' Construct an 'xgb.DMatrix' object from a given data source, which can then be passed to functions
+#' such as \link{xgb.train} or \link{predict.xgb.Booster}.
+#'
+#' Function 'xgb.QuantileDMatrix' will construct a DMatrix with quantization for the histogram
+#' method already applied to it, which can be used to reduce memory usage (compared to using a
+#' a regular DMatrix first and then creating a quantization out of it) when using the histogram
+#' method (`tree_method = "hist"`, which is the default algorithm), but is not usable for the
+#' sorted-indices method (`tree_method = "exact"`).
```
trivialfis, line 10: nor approx.

R-package/R/xgb.DMatrix.R
```diff
@@ -1,13 +1,41 @@
...
+#' responsibility to ensure that factor columns have the same levels as the ones from which the DMatrix
+#' was constructed.
+#'
+#' Other column types are not supported.
+#' \item CSR matrices, as class `dgRMatrix` from package `Matrix`.
+#' \item CSC matrices, as class `dgCMatrix` from package `Matrix`. These are \bold{not} supported for
+#' 'xgb.QuantileDMatrix'.
+#' \item Single-row CSR matrices, as class `dsparseVector` from package `Matrix`, which is interpreted
+#' as a single row (only when making predictions from a fitted model).
+#' \item Text files in SVMLight / LibSVM formats, passed as a path to the file. These are \bold{not}
```
trivialfis, line 34: We can stop supporting it in DMatrix as well: https://github.com/dmlc/xgboost/issues/9472

R-package/R/xgb.DMatrix.R
```diff
@@ -18,11 +46,32 @@
...
-#'        It is useful when a 0 or some other extreme value represents missing values in data.
+#' @param missing a float value to represents missing values in data (not used when creating DMatrix
+#'        from text files).
+#'        It is useful to change when a zero, infinite, or some other extreme value represents missing
+#'        values in data.
 #' @param silent whether to suppress printing an informational message after loading from a file.
 #' @param feature_names Set names for features. Overrides column names in data
 #'        frame and matrix.
+#'
+#'        Note: columns are not referenced by name when calling `predict`, so the column order there
```
trivialfis, line 57: I think this is an area that I should work on, to make sure the ordering is correct, at least for dataframe.

R-package/R/xgb.DMatrix.R
```diff
@@ -195,10 +299,491 @@ xgb.DMatrix <- function(
...
+#' # After construction, can be used as a regular DMatrix
+#' params <- list(nthread = 1, objective = "reg:squarederror")
+#' model <- xgb.train(data = dm, nrounds = 2, params = params)
+#'
+#' # Predictions can also be called on it, and should be the same
+#' # as if the data were passed differently.
+#' pred_dm <- predict(model, dm)
+#' pred_mat <- predict(model, as.matrix(mtcars[, -1]))
+#' @export
+xgb.ExternalDMatrix <- function(
```
trivialfis, line 706: I'm not sure whether should make a different class at the interface level for external memory. Personally, I would prefer simply having DM and QDM to accept an iterator to avoid having confusion.

R-package/R/xgb.DMatrix.R
trivialfis: I just remember the `valid_features` parameter in Python, there's a check in there to verify the columns are consistent.

R-package/R/xgb.DMatrix.R
david-cortes: Are there definitive plans to remove it? I see it hasn't been removed from the C++ side yet.

R-package/R/xgb.DMatrix.R
david-cortes: I think this sort of check should definitely not be enabled by default in a prediction function, since it can potentially add a lot of latency when serving a model.

My idea was to have such a check implemented in the new `xgboost()` interface (with a separate class for the model object), but would be ideal if it could be done in C++ without having to copy the names to R character vectors.

That being said, in e.g. R's `predict.glm`, if the data contains the column names but in a different order, then the input data would get re-ordered instead of throwing an error.

R-package/R/xgb.DMatrix.R
trivialfis: Not yet, I need to keep the work in sync with the work for federated learning, which uses text data for testing. But it's for certain that we want to deprecate it.

R-package/NAMESPACE
trivialfis: That makes sense, I have been struggling with this since the first review and have been stalling the decision (my bad). Will look closer in the new class.

R-package/R/xgb.DMatrix.R
david-cortes: Then let's better leave that for a subsequent PR after this and in-place predict have been merged, since both touch code sections dealing with this text format, like docs and such.

R-package/R/xgb.DMatrix.R
trivialfis: The check in Python is optional but enabled by default. I'm not sure how to do it in the new `xgboost` interface though since the check is for inference while the `xgboost()` function is for training.

david-cortes: My idea was to make the output from `xgboost()` be of a different class than `xgb.train`, just like the python interface has scikit-learn classes which are different from `xgb.Booster` - then, this class coming out of `xgboost()` would have a different `predict` method which could do extra things, such as add class names, extra checks, etc.

david-cortes: Although, if the data has column names and so does the booster, it should be possible to add an extra parameter in `predict.xgb.Booster` to optionally make it behave like R's `predict.glm` which reorders columns and fails if they aren't present.

R-package/R/xgb.DMatrix.R
```diff
@@ -195,10 +300,491 @@ xgb.DMatrix <- function(
...
+#'     y_batch <- iterator_env[["y"]][1:16]
+#'   } else {
+#'     x_batch <- iterator_env[["x"]][17:32, ]
+#'     y_batch <- iterator_env[["y"]][17:32]
+#'   }
+#'
+#'   # Function 'xgb.ProxyDMatrix' must be called manually
+#'   # at each batch with all the appropriate attributes,
+#'   # such as feature names and feature types.
+#'   xgb.ProxyDMatrix(
```
trivialfis, line 672: Is it possible that we hide this function and pass a closure to the user like the Python code? I think it's easier to use if the user doesn't need to worry about an additional function and an opaque handle. R is a high level language and it would great if we can hide these C artifacts.

R-package/R/xgb.DMatrix.R
trivialfis: Got it, we will do the check in your way.

R-package/R/xgb.DMatrix.R
```diff
@@ -195,10 +300,491 @@ xgb.DMatrix <- function(
...
+#'
+#' Might be created as either a regular 'DMatrix', or as a 'QuantileDMatrix'.\itemize{
+#' \item If created as a regular 'DMatrix', the data is accessed on-demand as needed, multiple
+#' times, without being concatenated (but note that fields like 'label' \bold{will} be concatenated
+#' from multiple calls to the data iterator).
+#' \item If created as a 'QuantileDMatrix', the quantized representation of the full data will
+#' be created in memory, concatenated from multiple calls to the data iterator. The quantized
+#' version is typically lighter than the original data, so there might be cases in which this
+#' representation could potentially fit in memory even if the full data doesn't.
+#' }
```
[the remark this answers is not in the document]
david-cortes: Yes, it does sound quite confusing, but I don't think the python interface is the best reference here.

If I understand it correctly, the issue here is that there's two routes in this PR which give the same object type:
* `xgb.QuantileDMatrix` + in-memory data.
* `xgb.ExternalDMatrix` + `xgb.DataIter` + `as_quantile_dmatrix=TRUE`.

But if this were to be implemented the same way as in python (accepting `xgb.DataIter` as `data` instead of having a dedicated class for it), then there would be:
* A function `xgb.DMatrix` which could still produce different output types depending on its inputs (regular DMatrix vs. the 'ExternalDMatrix' introduced here).
* A parameter `missing` which is auto-modified for integer inputs in some cases but not others.
* Multiple parameters that can contradict each other (e.g. feature names).

If the idea is to avoid inconsistencies, then I think the best option would be to have only two functions:
* `xgb.DMatrix`, taking only in-memory data, and accepting a parameter `as_quantile_dmatrix` in order to produce class `xgb.QuantileDMatrix` as output.
* `xgb.ExternalDMatrix`, taking only `xgb.DataIter` as input, and also accepting a parameter `as_quantile_dmatrix`.

Either that, or 4 functions after an 'ExternalQuantileDMatrix' gets implemented.

But I can also guess that:
* There will be much fewer people using `xgb.ExternalDMatrix` than the others.
* People using `xgb.ExternalDMatrix` will likely be the more advanced users.
* People using `xgb.ExternalDMatrix` will probably be more worried about getting the data iterators right and wondering about their usage than about other DMatrix properties.

Which is why I chose this kind of separation.

If there was an ExternalQuantileDMatrix type implemented, perhaps it'd make more sense to have 4 possible constructors, but I'm not sure about what's the best way to define these constructors and classes with the current behavior.

R-package/R/xgb.DMatrix.R
david-cortes: I don't think a closure is the best option here.

Another option could be to have the user return the result of this `xgb.ProxyDMatrix` (or `NULL` to signal end of batch), but let this be just a simple list and defer processing for later instead of processing things just after `xgb.ProxyDMatrix` is called.

It would however forego the possibility of doing a manual clean up of whatever environment generated the data inside the iterator if the user wishes to do so immediately before the next batch, but perhaps it's a more logical design.

R-package/R/xgb.DMatrix.R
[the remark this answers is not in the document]
trivialfis: > but I don't think the python interface is the best reference here

I agree. For background, the interface of Python is mostly caused by the use of text file input like `data="path/to/csv"`, which has the same problem with the iterator with confusing parameters.

> A parameter missing which is auto-modified for integer inputs in some cases but not others.

Not entirely sure what this means. I think as long as a missing value is accepted, integer will be considered as float.

Thank you for listing all the possible options, and apologies for the slow replies lately, will be more active starting tomorrow.

In general, I would still like to differentiate `DMatrix` and `QuantileDMatrix` since they have different functionalities and accept a different set of parameters, only QDM accepts `max_bin`, a single parameter `as_quantile_dmatrix` might not be sufficient to reflect and clarify the difference, and my preference is still to have two different classes for them.

On the other hand, `ExternalMemDMatrix` is quite similar to `DMatrix` with the same set of parameters, and can be used "almost" all the places where DMatrix can be used.

The question is how to integrate the concept of external memory into the mix. I propose the following functions and classes, `ExternalMemoryDMatrix` as implemented in this PR, but without the support for being a QDM. In addition, a QDM constructor that accepts only the iterator `xgb.QuantileDMatrix.from_iterator(iter) -> xgb.QuantileDMatrix`.

The reasoning behind this is that the `xgb.QuantileDMatrix` and `xgb.QuantileDMatrix.from_iterator` return the same type with the exact same underlying data (in c++), it's just how it's constructed in the first place and so there's no need for a new class in R. Then we use different constructor to represent the difference of construction process. As for the external memory version of DMatrix, as you have suggested, they behave differently, we will continue with your approach of creating a new class for it.

What do you think?

R-package/R/xgb.DMatrix.R
trivialfis: > I don't think a closure is the best option here.

Could you please elaborate on this?

R-package/R/xgb.DMatrix.R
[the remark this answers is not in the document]
david-cortes: Sounds good.

So that'd leave the following functions with the following output classes:
* `xgb.DMatrix` -> `xgb.DMatrix`
* `xgb.QuantileDMatrix` -> `xgb.QuantileDMatrix`
* `xgb.QuantileDMatrix.from_iterator` -> `xgb.QuantileDMatrix`
* `xgb.ExternalDMatrix` -> `xgb.ExternalDMatrix`

Is that a correct understanding?

> I think as long as a missing value is accepted, integer will be considered as float.

Not quite - integer matrices are also accepted, and are passed as such to the array_interface parser. In such case, the missing value is auto-adjusted in the C++ code according to the type of the input, but if the user passes a data iterator, the missing value is passed before the iterator to the first C function, and thus cannot be converted later on once the data is seen and the dtype is determined (with integer missing value in R being `-INT_MAX`).

R-package/R/xgb.DMatrix.R
david-cortes: > > I don't think a closure is the best option here.
> 
> Could you please elaborate on this?

They are akward to use, hard to document, and hard to examine as a user. Functions, environments and lists are more common and easier to use. Or do you think that a closure could bring some advantage here that would not be achieved by e.g. letting the user return the result of `xgb.ProxyDMatrix` ? (assuming that the handle is left out of the signature)

R-package/R/xgb.DMatrix.R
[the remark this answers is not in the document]
trivialfis: > Is that a correct understanding?

Yes.

> and thus cannot be converted later on once the data is seen and the dtype is determined

That's interesting, did not think of that.

R-package/R/xgb.DMatrix.R
trivialfis: I don't have preference on whether a closure or a function should be used, it's just I would like to hide/abstract many things as possible, and sometimes closure is the way to do it. If there are other approaches, I'm happy to take it.

R-package/R/xgb.DMatrix.R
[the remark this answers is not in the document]
david-cortes: Renamed.

R-package/R/xgb.DMatrix.R
david-cortes: Factored out the `proxy_handle` parameter. Now the user needs to call a function `xgb.ProxyDMatrix` inside of the callback (which doesn't take any handle) and return its result.

Issue #9908, opened 2023-12-20
Tests for custom objective/metric with serialization.


trivialfis:
- [ ] Pickle sklearn.
- [ ] Pickle booster.
- [ ] save_model sklearn.
- [ ] save_model booster.

PR #9924, opened 2023-12-26
[R] On-demand serialization + standardization of attributes

david-cortes:

This PR changes the serialization logic of `xgb.Booster` objects to trigger only on-demand, by using an R altrep list class on which serialization methods are implemented, and not all of the expected altrep methods are implemented so as to avoid potential unwanted conversions that might lose the serializers.

Since the serialization logic changes with this PR, the way in which attributes are kept in model objects also needed to be changed:
* Now there is a clear division between R-specific attributes, which can be arbitrary objects and are accessible and settable through `attributes(model)`; and C-level attributes which are kept in the model JSON, which can be accessed and set through `xgb.attributes(model)`.
* Some attributes which were previously part of the R class were now moved to C-level attributes.
* As serialization is now on-demand and R attributes are optional, there's no further need for an `xgb.Booster.handle` class. I removed it as part of this PR, since it would not have any use-case where it'd have some advantage over `xgb.Booster` without optional attributes.

This separation means that now accessing attributes is not as simple as calling `model$<attribute>` - now one needs to explicitly use either `attributes` or `xgb.attributes`.

Since the logic for keeping track of attributes was changed here, this PR **required doing changes throughout pretty much all the R code**. As it is right now, all of the tests are passing, but **I am not entirely confident that this PR won't break something not covered in the tests**, and I am not sure if I have updated all the docs that became outdated after these changes.

The changes here also now make xgboost incompatible with {caret}, so I've removed all the references to it. I also removed suggestions around it since this package is not in active development anymore as it was superceded by {tidymodels}. Haven't tested {mlr} but I have a feeling it might also break.

A couple notes about the PR and about many things I noticed - **would be ideal if maintainers could open independent issues about some of these if needed**:

### Prioritization

**As this PR creates merge conflicts with all others and it's the hardest to keep track of, would be very helpful to merge it before others like inplace predict or quantile dmatrices or roxygen updates**.

### Shallow and deep copies

Before this PR, serialization and de-serialization was triggered multiple times throughout the runs of different functions. Doing this was very inefficient, and had the potential to create inconsistencies between the R attributes and the C booster.

After this PR, there are no such duplications - objects are updated in-place, at the C level. It's now also possible to incrementally update a booster in-place through `xgb.train`, which is controlled by an optional parameter `training_continuation`, but I wasn't sure how this parameter should play along with xgboost parameter `process_type`.

* Note that there's one edge case that I didn't know how to cover: if the user sends an interrupt signal, a booster that's being modified in-place will be left in an inconsistent state between the R attributes (like evaluation log) and the C booster. The R attributes are nevertheless not used in `predict` or similar methods.

Since one might now need to make copies of the booster, a helper function `xgb.copy.Booster` was also implemented in the public interface. I could not find any C-level function to duplicate a booster so I used the ubj to-bytes serializer for it, which I am guessing is not the most efficient way.

### Information lost during serialization

It seems that the serialization functions that save models to disk (e.g. `XGBoosterSaveModel`), regardless of the format that they end up using (json/ubj), will never save custom user attributes that one sets to the booster outside of training, except for some particular attributes like 'niter'; unlike the serialization functions that save them to bytes like `XGBoosterSaveModelToRaw` which keep all of the custom attributes.

Additionally, these serializers also seem to lose feature names and types if they were set in the booster through `XGBoosterSetStrFeatureInfo`. As such, I modified some of the tests to manually remove the feature names for the comparisons, but I have a feeling that something here could be improved.

### Serialization compatibility

After this PR, it will not be possible to load models that were saved with R serializers like `saveRDS` with a previous xgboost version, and it will not be possible to load models saved with `saveRDS` after this PR in previous xgboost versions. I updated the compatibility note to mention the breakpoint being xgboost version 2.1.0, which I suppose will be the next release.

There's a test which downloads serialized files from the internet and which will need to be updated. For now, I simply modified the test to skip the Rds files, but in the future would be more logical to update the files there.

I'm also thinking that it might be better to commit those model files here and bundle them in the R package instead of downloading them when the tests are run - if the files are updated now, unless the download link is changed, then the same test running on older xgboost versions will fail - will also make testing faster as there won't be a need to download files from the internet.

### Serialization advise

XGBoost's own docs advise users to not use R-specific serializers, but lots of the functionalities from the public interface actually **require** using R serializers, such as function `xgb.gblinear.history`, which uses attributes from `callbacks` that aren't part of the standard C booster and thus do not get saved in functions like `XGBoosterSaveModel` or `XGBoosterSaveModelToRaw`. Moreover, from the point above, even C-level information is lost when using to-disk serializers, such as the feature names in the booster, which can also lead to very unexpected behaviors (e.g. they are saved with `xgb.save.raw`, but not with `xgb.save`).

I was thinking that perhaps the booster could add serialized versions of the R attributes (as obtained by `base::serialize`), but those raw bytes would not conform to a JSON-loadable string, would add latency and memory usage, are not usable in other interfaces, among other minuses.

It leaves me wondering what should be the actual suggestion in the doc, given that a user following that advise might find that things actually break after the fact, and there's advantages to using `xgb.save.raw` + R's `writeBin` compared to directly using `xgb.save`.

### Plotting multi-valued trees

I am not sure if the plotting functionalities need any kind of adjustment for models like multi-quantile regression, multi-target objectives, and so on. I didn't add any modification here.

In the case of multi-quantile regression, functions seem to produce some output, but I am not sure about the correctness of such output, since there's one value per leaf visible there. Should this perhaps produce an error at the C++ level?

For mutli-output regression, functions like `xgb.dump` will error out from the C side - I would guess that multi-quantile and multi-class might share some structural similarities but am not familiar with tree plotting.

### Retrieving booster attributes

As part of this PR, I added internal functions to access fields that are part of the C booster, such as the booster type.

I implemented them by producing full dumps of the booster JSON through `XGBoosterSaveJsonConfig`, then parsing them with R's `jsonlite`, and accessing the field in that parsed JSON.

This is rather inefficient, but I couldn't find any C-level function to extract only one particular field from the booster JSON. Would be quite useful to add more functions to retrieve and set commonly used attributes like the booster type and the number of threads, or to retrieve only one particular attribute from the JSON if given a path like `field1.subfield2.<etc>`.

### Missing functions in core library

Both the R and Python interfaces now resort to parsing text dumps in order to extract information such as model coefficients in a linear booster. If such functionalities are going to be required throughout different interfaces, would be ideal to create C-level accessors for them that would for example avoid the loss of precision from the conversion from float to string and back.

Furthermore, some of these parsings broke after adding feature names, and the regexes needed to be updated. I am not sure that I've pushed updated everywhere necessary.

### Attribute 'niter'

Handling of this attribute was rather strange before this PR. This attribute was kept both in the booster attributes and in the R attributes, with the caveat that the C one used base-0 indexing, while the R one used base-1. Further, they were not updated in synch everywhere.

After this PR, I preferred to remove this attribute, sticking instead to 'nrounds' as returned by function `XGBoosterBoostedRounds`.

There is one particular issue with this function however: after calling `XGBoosterSetParam` on a fitted booster, depending on what the parameters there contain, function `XGBoosterBoostedRounds` will afterwards return zero, but the trees will still be there when looking at e.g. `learner.gradient_booster.gbtree_model_param.num_trees` in the JSON config.

### Linear intercepts

When creating a `gblinear` booster, it will have two intercepts:
* One given by `base_score`.
* Another given by the last coefficient in the JSON.

Both of those intercepts are added together in the prediction, which is quite confusing. Would be ideal if this booster type could auto-adjust itself after the fact to contain only one intercept - for example by forcibly setting `base_score` to zero and adding it to the last coefficient.

It would also improve things for some objective like `multi:softmax` as it gets an automatically added a `base_score` of 0.5 which, logically speaking, does not help with convergence / loss decrease.

### Legacy R code

Changing things throughout multiple places made me notice that the logic for callbacks is very hard to follow and rather unidiomatic - for example, it requires defining variables inside the function call that do not seem to be used and one cannot tell from a look at that function if they could be removed or not. I am guessing it will also be the hardest thing to review in this PR, and the one that's most prone to errors.

Since these callbacks are part of the public interface (meaning: the user can pass custom callbacks), it'd be ideal to:
* Document them (current docs do not cover key aspects like the contents of the environments).
* Change the logic from using environments where variables are assigned to arbitrary names (e.g. the user might not know that there's a `bst` variable in the environment that needs to be accessed), towards using function keyword arguments.
    * Ideally, there should be a factory function like `xgb.Callback` that would take arguments like `before_training`, `after_training`, `before_iteration`, `after_iteration`, each taking functions expecting a defined function signature.
    * It should also standardize the logic for keeping things from a callback in booster attributes or not.
* Avoid usage of global variables and `<<-` setters - better to keep a shared R `environment`-class variable for such purposes.

### R-specific functionality

I also noticed that there are some functionalities in the R interface that the Python one lacks, such as function `xgb.gblinear.history`. I think this function could also be moved into a C-level functionality to keep that coefficient history as an internal booster property.


trivialfis:
Thank you for introducing the new serialization! This is a major upgrade to the existing interface and I'm excited about it.

I will look into the PR deeper in the coming days and learn more about the changes.

Since this is related to serialization, I will assist in writing some testing guidelines, as mentioned earlier, serialization has been a foot gun and we still have issues with it today. I would like to be extra careful from the beginning.

trivialfis:
Related https://github.com/dmlc/xgboost/issues/9908

david-cortes:
Looks like the windows build from buildkite is still using R 3.6:
```
-- Found LibR: C:/Program Files/R/R-3.6.3
...
C:\buildkite-agent\builds\buildkite-windows-cpu-autoscaling-group-i-0667cc80194b35b5c-1\xgboost\xgboost-ci-windows\R-package\src\xgboost_R.cc(679,27): error C3861: 'R_make_altlist_class': identifier not found [C:\buildkite-agent\builds\buildkite-windows-cpu-autoscaling-group-i-0667cc80194b35b5c-1\xgboost\xgboost-ci-windows\build\R-package\xgboost-r.vcxproj]
C:\buildkite-agent\builds\buildkite-windows-cpu-autoscaling-group-i-0667cc80194b35b5c-1\xgboost\xgboost-ci-windows\R-package\src\xgboost_R.cc(681,3): error C3861: 'R_set_altlist_Elt_method': identifier not found [C:\buildkite-agent\builds\buildkite-windows-cpu-autoscaling-group-i-0667cc80194b35b5c-1\xgboost\xgboost-ci-windows\build\R-package\xgboost-r.vcxproj]
```

hcho3:
What's the minimum R version do we need for ALTREP? I can update the R version on the Windows side.

R-package/src/xgboost_R.cc
```diff
@@ -540,15 +546,147 @@ XGB_DLL SEXP XGDMatrixNumCol_R(SEXP handle) {
...
+XGB_DLL SEXP XGPointerEqComparison(SEXP obj1, SEXP obj2) {
+  return Rf_ScalarLogical(R_ExternalPtrAddr(obj1) == R_ExternalPtrAddr(obj2));
+}
+
 // functions related to booster
-void _BoosterFinalizer(SEXP ext) {
-  if (R_ExternalPtrAddr(ext) == NULL) return;
-  CHECK_CALL(XGBoosterFree(R_ExternalPtrAddr(ext)));
-  R_ClearExternalPtr(ext);
+static void _BoosterFinalizer(SEXP R_ptr) {
```
hcho3, line 563: Instead of using `static`, can we put all the internal functions in an anonymous namespace?

R-package/src/xgboost_R.cc
```diff
@@ -540,15 +546,147 @@ XGB_DLL SEXP XGDMatrixNumCol_R(SEXP handle) {
...
+
+/* Note: the idea for separating this function from the one above is to be
+able to trigger all R allocations first before doing non-R allocations. */
+static void XGBAltrepSetPointer(SEXP R_altrepped_obj, BoosterHandle handle) {
+  SEXP R_ptr = R_altrep_data1(R_altrepped_obj);
+  R_SetExternalPtrAddr(R_ptr, handle);
+  R_RegisterCFinalizerEx(R_ptr, _BoosterFinalizer, TRUE);
+}
+
+const char *ubj_json_format_str = "{\"format\": \"ubj\"}";
```
hcho3, line 601: This should also be put inside an anonymous namespace, to prevent possible name collision.

hcho3 (changes requested):

david-cortes:
> What's the minimum R version do we need for ALTREP? I can update the R version on the Windows side.

It's R 4.3.

hcho3:
> I'm also thinking that it might be better to commit those model files here and bundle them in the R package instead of downloading them when the tests are run - if the files are updated now, unless the download link is changed, then the same test running on older xgboost versions will fail - will also make testing faster as there won't be a need to download files from the internet.

Currently, we have a policy of not including binary files in the git repository, since git isn't suited for handling diffs in binary (non-text) files. Some alternatives:

* Use Git LFS. This can get expensive pretty fast, as GitHub charges per storage and per bandwidth.
* Keep hosting model files externally, but version the files, so that old tests don't break when we update the model files later.

hcho3:
Does it mean that the new XGBoost R package will only be compatible with R 4.3 ? We probably should document the requirement.

david-cortes:
> Does it mean that the new XGBoost R package will only be compatible with R 4.3 ? We probably should document the requirement.

It's already stated in the `DESCRIPTION` file, so a user installing it with `install.packages` will not be able to install the new xgboost version in an older R version.

hcho3:
Sorry, I missed the conversation in #9847, as I was gone for vacation. I will go ahead and update the R version in the Windows CI.

R-package/src/xgboost_R.cc
david-cortes: Moved to a namespace.

R-package/src/xgboost_R.cc
david-cortes: Moved to the same namespace as above.

R-package/R/utils.R
```diff
@@ -363,6 +360,14 @@ NULL
 #' accessible in later releases of XGBoost. To ensure that your model can be accessed in future
 #' releases of XGBoost, use \code{\link{xgb.save}} or \code{\link{xgb.save.raw}} instead.
 #'
+#' Currently, it is not possible to use R serializers like `readRDS` to load an XGBoost. model
+#' saved with an XGBoost. version lower than 2.1.0, and it's not possible to load an XGBoost. model
+#' saved with R serializers like `readRDS` under XGBoost. version 2.1.0 when using an older version
+#' of XGBoost.
```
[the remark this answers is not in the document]
david-cortes: I am not sure where to obtain such information, and don't think it'd be very helpful if those docs are only going to be mentioned in the newer xgboost version, which will have 'x' in all previous version boxes.

I've nevertheless reworded the notice a bit to make it more clear that there's mutual incopatibilities before/after 2.1.0.

david-cortes:
> > I'm also thinking that it might be better to commit those model files here and bundle them in the R package instead of downloading them when the tests are run - if the files are updated now, unless the download link is changed, then the same test running on older xgboost versions will fail - will also make testing faster as there won't be a need to download files from the internet.
> 
> Currently, we have a policy of not including binary files in the git repository, since git isn't suited for handling diffs in binary (non-text) files. Some alternatives:
> 
> * Use Git LFS. This can get expensive pretty fast, as GitHub charges per storage and per bandwidth.
> * Keep hosting model files externally, but version the files (by adding suffix), so that old tests don't break when we update the model files later.

There's also the option of pre-downloading it as part of the R package build script; and/or to look for the file under somewhere like `inst` when running tests without creating a package plus adding it to `.gitignore`, and save the file there instead of `tmp` if it detects such a folder.

hcho3 (approved):
No objections from me

R-package/src/xgboost_R.cc
```diff
@@ -562,33 +705,104 @@ XGB_DLL SEXP XGBoosterCreate_R(SEXP dmats) {
...
+XGB_DLL SEXP XGBoosterBoostedRounds_R(SEXP handle) {
+  SEXP out = Rf_protect(Rf_allocVector(INTSXP, 1));
+  R_API_BEGIN();
+  CHECK_CALL(XGBoosterBoostedRounds(R_ExternalPtrAddr(handle), INTEGER(out)));
+  R_API_END();
+  Rf_unprotect(1);
+  return out;
+}
+
+/* Note: R's integer class is 32-bit-and-signed only, while xgboost
```
trivialfis, line 793: It's probably fine to return the signed integer in R since and throw an error if it overflows. Overflow can only happen if the model is trained using a different language binding since R cannot handle data size larger than signed integer.

R-package/src/xgboost_R.cc
```diff
@@ -562,33 +705,104 @@ XGB_DLL SEXP XGBoosterCreate_R(SEXP dmats) {
...
+  for (bst_ulong idx = 0; idx < len; idx++) {
+    SET_STRING_ELT(out, idx, Rf_mkChar(out_features[idx]));
+  }
+  Rf_unprotect(2);
+  return out;
+  R_API_END();
+  return R_NilValue; /* <- should not be reached */
+}
+
+XGB_DLL SEXP XGBoosterBoostedRounds_R(SEXP handle) {
```
trivialfis, line 784: Can we extract some of the unrelated changes into a smaller PR and merge them first? Makes reviews easier. ;-)

trivialfis (commented):
Huge thanks for the work on serialization! It will take some time to process it (I'm still on PTO). Can we extract some of the unrelated changes and merge them first?

In addition, a brief introduction to how it works as code comment would be nice. Something that can get new comers up to speed.

R-package/src/xgboost_R.cc
david-cortes: It'd be quite hard to do so, because this piece of information (which replaces both of the older 'niter') is used internally in combination with other attributes from C and R.

It's also not possible to make this PR without these sort of changes, because these attributes are now set and accessed differently, and the PR by default leaves the booster clean of custom C attributes, which requires this functionality.

R-package/src/xgboost_R.cc
david-cortes: Objects from `arrow` support larger dimensions. If we add them later on it will be possible to have overflows.

david-cortes:
> In addition, a brief introduction to how it works as code comment would be nice. Something that can get new comers up to speed.

There's some small code comments before the class definition (`xgboost_R.cc:572`), and some documentation about how to use the R object in the 'details' section of the docs for `xgb.train`. Not sure what or where else to add for comments.

trivialfis:
Will get to this PR once https://github.com/dmlc/xgboost/pull/9947 is sorted out.

hcho3:
Any other blockers for this PR?

trivialfis:
I'm looking into it today. Want understand it better before having change in IO.

trivialfis:
> I could not find any C-level function to duplicate a booster so I used the ubj to-bytes serializer for it, which I am guessing is not the most efficient way

It's fine, the UBJ is reasonably efficient, and we don't want to perform copy too often.

> Additionally, these serializers also seem to lose feature names and types if they were set in the booster through

Could you please try that again now that the default format is UBJ?

> Would be quite useful to add more functions to retrieve and set commonly used attributes like the booster type and the number of threads

Internally, the parameters are scattered across components, we have to do the walk anyway, with or without JSON.

> There is one particular issue with this function however: after calling XGBoosterSetParam on a fitted booster

Could you please share an example for this issue? Are you switching the booster parameter from gbtree to gblinear?

trivialfis:
I have added additional tests for the cases you mentioned: https://github.com/dmlc/xgboost/pull/9962

- Custom attribute `{"foo": "bar"}`
- Feature names and types, set with and without DMatrix.
- The number of boosted rounds, checked with and without changing parameters and updating type.

Could you please take a look and share what I'm missing there?

david-cortes:
1. Did some testing around saving model to disk. Can confirm that everything is saved **when using JSON or UBJ**. Will update the docs, examples, and checkpoint callback here. Will also move the `setinfo` for the booster to the beginning instead of the end so that the checkpoint callback will get that info.
2. For the number of rounds, on further inspection, it seems to happen when replacing the rounds in an update (parameter `process_type=update`), which I guess it's the right thing to do if previous rounds are going to be overwritten. Although I see that Python and R handle the round numbering quite differently in the case of non-replacing updates.

trivialfis:
Excellent! Will look into other changes now that we are certain the model IO itself is working as expected.

david-cortes:
@trivialfis I've pushed some updates around the serialization format and removed the workarounds from the tests. Could you check that the docs that are now in `xgb.save` reflect the way that things work internally?

david-cortes:
> After cleaning up for using UBJ as default (removing code that's used to handle the issues mentioned in the PR description), the next item for this PR would be to make sure both model IO that discards hyperparameter and model serialization that preserves hyperparameter are working correctly.

I think the current tests already cover those.

R-package/NAMESPACE
```diff
@@ -26,17 +28,18 @@ export(cb.save.model)
...
 export(xgb.DMatrix.save)
 export(xgb.attr)
 export(xgb.attributes)
 export(xgb.config)
+export(xgb.copy.Booster)
 export(xgb.create.features)
 export(xgb.cv)
 export(xgb.dump)
 export(xgb.gblinear.history)
+export(xgb.get.Booster.nrounds)
```
trivialfis, line 42: It's no a big issue, but would be great to align the naming with `num_boosted_rounds`:
- Python `num_boosted_rounds`.
- Java `getNumBoostedRound`.
- Scala `getNumBoostedRound`.

R-package/R/xgb.Booster.R
```diff
@@ -1,180 +1,88 @@
-# Construct an internal xgboost Booster and return a handle to it.
+# Construct an internal xgboost Booster and get its current number of rounds.
 # internal utility function
-xgb.Booster.handle <- function(params, cachelist, modelfile, handle) {
+# Note: the number of rounds in the C booster gets reset to zero when updating
+# the parameters through 'xgb.parameters', hence the need to return it from
+# this function when using it for training continuation
```
trivialfis, line 5: Is this comment valid?

R-package/R/xgb.Booster.R
```diff
@@ -1,180 +1,88 @@
...
+xgb.Booster <- function(params, cachelist, modelfile, training_continuation) {
```
trivialfis, line 6: I'm not sure about the necessity of `training_continuation` parameter.

> It's now also possible to incrementally update a booster in-place through xgb.trai

Making a copy is not too bad, it conforms to the R semantic of immutable data. A single copy is quite cheap compared to the training itself. You can verify this by generating large models and serializing with UBJSON in memory (not writing to file).

Edit: it seems you are copying the booster anyway by calling a `duplicate` function?

R-package/R/xgb.Booster.R
```diff
@@ -1,180 +1,88 @@
...
-      class(handle) <- "xgb.Booster.handle"
+      .Call(XGBoosterLoadModel_R, xgb.get.handle(bst), enc2utf8(modelfile[1]))
+      niter <- xgb.nrounds(bst)
       if (length(params) > 0) {
-        xgb.parameters(handle) <- params
+        xgb.parameters(bst) <- params
       }
-      return(handle)
-    } else if (typeof(modelfile) == "raw") {
+      return(list(bst = bst, niter = niter))
```
trivialfis, line 22: Do we need to pack the niter into a  list with bst?

R-package/R/xgb.Booster.R
```diff
@@ -1,180 +1,88 @@
...
+    } else if (is.raw(modelfile)) {
       ## A memory buffer
-      bst <- xgb.unserialize(modelfile, handle)
+      bst <- xgb.unserialize(modelfile)
```
trivialfis, line 25: Can we use the function `xgb.load.raw`, which corresponds to the `xgb.save.raw` function?

In general, we have three sets of IO procedures:
- Save the model to a file, without hyper-parameters like the number of threads.
```
XGBoosterSaveModel
XGBoosterLoadModel
```
`xgb.save` is using this one. Preferably we can rename it to `save.model`. No strong opinion there.

- Save the model to a buffer, without hyper-parameters like the number of threads.
```
XGBoosterSaveModelToBuffer
XGBoosterLoadModelFromBuffer
```
`xgb.save.raw` is using this one.
- Save everything to a buffer, including hyper-parameters.
```
XGBoosterSerializeToBuffer
XGBoosterUnserializeFromBuffer
```
`xgb.serialize` is using this one.

We need to make sure the right function is called.

R-package/R/xgb.Booster.R
```diff
@@ -1,180 +1,88 @@
...
+      niter <- xgb.nrounds(bst)
       xgb.parameters(bst) <- params
-      return(bst)
+      return(list(bst = bst, niter = niter))
```
trivialfis, line 28: If I understand correctly, you want to get this attribute on-demand and remove R attributes so that everything is consistent.

R-package/R/xgb.Booster.R
```diff
@@ -1,180 +1,88 @@
...
     } else if (inherits(modelfile, "xgb.Booster")) {
       ## A booster object
-      bst <- xgb.Booster.complete(modelfile, saveraw = TRUE)
-      bst <- xgb.unserialize(bst$raw)
-      xgb.parameters(bst) <- params
-      return(bst)
+      if (training_continuation == "copy") {
```
trivialfis, line 31: Please see the previous comment about the necessity of this parameter.

R-package/R/xgb.Booster.R
```diff
@@ -1,180 +1,88 @@
...
-#' if (file.exists("xgb.model.rds")) file.remove("xgb.model.rds")
-#' # the handle is invalid:
-#' print(bst1$handle)
-#'
-#' bst1 <- xgb.Booster.complete(bst1)
-#' # now the handle points to a valid internal booster model:
-#' print(bst1$handle)
-#'
-#' @export
-xgb.Booster.complete <- function(object, saveraw = TRUE) {
```
trivialfis, line 135: Excellent! I'm always worrying that this function is causing trouble.

R-package/R/xgb.Booster.R
```diff
@@ -597,14 +490,25 @@ predict.xgb.Booster.handle <- function(object, ...) {
 xgb.attr <- function(object, name) {
   if (is.null(name) || nchar(as.character(name[1])) == 0) stop("invalid attribute name")
   handle <- xgb.get.handle(object)
-  .Call(XGBoosterGetAttr_R, handle, as.character(name[1]))
+  out <- .Call(XGBoosterGetAttr_R, handle, as.character(name[1]))
+  if (!NROW(out) || !nchar(out)) {
+    return(NULL)
+  }
+  if (!is.null(out)) {
+    if (name %in% c("best_iteration", "best_ntreelimit", "best_score")) {
```
trivialfis, line 498: Interesting, I should update the Python interface to follow the change here.

R-package/R/xgb.Booster.R
```diff
@@ -641,31 +540,22 @@ xgb.attributes <- function(object) {
...
+#' of R attributes, doesn't follow typical copy-on-write semantics for assignment - i.e. all references
+#' to the same booster will also get updated.
 #'
-#' @param object Object of class `xgb.Booster`.
+#' See \link{xgb.copy.Booster} for an example of this behavior.
+#' @param object Object of class `xgb.Booster`. \bold{Will be modified in-place} when assigning to it.
 #' @param value A JSON string.
-#'
+#' @return `xgb.config` will return the parameters as JSON text, which can be parsed through
+#' e.g. `jsonlite::fromJSON`.
```
trivialfis, line 558: I regret that I have made the interface like this, I think it's much better if we can call jsonlite inside this function. What do you think?

R-package/R/xgb.Booster.R
```diff
@@ -751,28 +642,292 @@ xgb.config <- function(object) {
...
     .Call(XGBoosterSetParam_R, handle, names(p[i]), p[[i]])
   }
-  if (is(object, 'xgb.Booster') && !is.null(object$raw)) {
-    object$raw <- xgb.serialize(object$handle)
+  return(object)
+}
+
+#' @rdname getinfo
+#' @export
+getinfo.xgb.Booster <- function(object, name) {
```
trivialfis, line 650: Is there any function in R that can be integrated with this info? Like `colnames`.

R-package/R/xgb.Booster.R
```diff
@@ -751,28 +642,292 @@ xgb.config <- function(object) {
...
+#' number of rounds to zero.
+#' @export
+xgb.get.Booster.nrounds <- function(model) {
+  return(xgb.nrounds(model))
 }
 
 # Extract the number of trees in a model.
-# TODO: either add a getter to C-interface, or simply set an 'ntree' attribute after each iteration.
 # internal utility function
+xgb.nrounds <- function(bst) {
```
trivialfis, line 700: See previous comment about naming.

R-package/R/xgb.DMatrix.R
```diff
@@ -357,9 +356,22 @@ dimnames.xgb.DMatrix <- function(x) {
...
 #' Note that, while 'qid' cannot be retrieved, it's possible to get the equivalent 'group'
 #' for a DMatrix that had 'qid' assigned.
 #'
+#' \bold{Important}: when calling `setinfo`, the objects are modified in-place. See
+#' \link{xgb.copy.Booster} for an idea of this in-place assignment works.
+#'
+#' Be aware that, when a booster object is serialized to disk through functions like
+#' \link{xgb.save}, information that was set on it will be lost, but such information
+#' would be kept as part of the serialization when using to-bytes serializers like
+#' \link{xgb.save.raw}, and when using R serializers like \link{saveRDS}.
```
trivialfis, line 374: please update the document here.

trivialfis (commented):
First round of review for the booster type for discussion. I still need to look into the rest.

trivialfis:
I'm quite stressed whenever there's a change in the model IO, so this will take time to go through, thank you very much for the excellent work on modernizing the booster class for R!

R-package/R/xgb.Booster.R
david-cortes: I don't know of any. I don't think `colnames` would be appropriate, since the booster doesn't hold the data and doesn't have dimensions when used calling `dim` on it.

Moreover, if one can slice it to subset boosting rounds, I don't think it'd be logical to then have the column names refer to something entirely different.

Perhaps @mayer79 or @jameslamb could give some suggestions here.

david-cortes: Actually, on second thought, there is `stats::variable.names`. Will add it as an option here.

R-package/R/xgb.DMatrix.R
david-cortes: Updated.

R-package/NAMESPACE
david-cortes: Changed to `xgb.get.num.boosted.rounds`.

R-package/R/xgb.train.R
```diff
@@ -371,27 +399,32 @@ xgb.train <- function(params = list(), data, nrounds, watchlist = list(),
...
   )
-  bst <- xgb.handleToBooster(handle = handle, raw = NULL)
 
   # extract parameters that can affect the relationship b/w #trees and #iterations
-  num_class <- max(as.numeric(NVL(params[['num_class']], 1)), 1)
-  num_parallel_tree <- max(as.numeric(NVL(params[['num_parallel_tree']], 1)), 1)
+  # Note: it might look like these aren't used, but they need to be defined in this
+  # environment for the callbacks for work correctly.
+  num_class <- max(as.numeric(NVL(params[['num_class']], 1)), 1) # nolint
+  num_parallel_tree <- max(as.numeric(NVL(params[['num_parallel_tree']], 1)), 1) # nolint
```
trivialfis, line 426: are these still useful after the `ntree_limit` is removed?

R-package/R/xgb.train.R
```diff
@@ -148,17 +148,43 @@
...
+#'        are kept as R attributes, and thus do not get saved when using non-R serializaters like
+#'        \link{xgb.save} (but are kept when using R serializers like \link{saveRDS}).
 #' @param ... other parameters to pass to \code{params}.
 #' @param label vector of response values. Should not be provided when data is
 #'        a local data file name or an \code{xgb.DMatrix}.
 #' @param missing by default is set to NA, which means that NA values should be considered as 'missing'
 #'        by the algorithm. Sometimes, 0 or other extreme value might be used to represent missing values.
 #'        This parameter is only used when input is a dense matrix.
 #' @param weight a vector indicating the weight for each row of the input.
+#' @param keep_extra_attributes Whether to keep extra R attributes in the booster object which
```
trivialfis, line 177: is this still necessary after your refactor for callbacks? Can we keep callback-specific attributes to callbacks themselves instead of the booster? (a question instead of a request).

R-package/src/init.c
```diff
@@ -15,9 +15,17 @@ Check these declarations against the C/Fortran source code.
...
 
 /* .Call calls */
+extern void XGBInitializeAltrepClass(DllInfo *info);
+extern SEXP XGDuplicate_R(SEXP);
+extern SEXP XGDuplicateAttrib(SEXP, SEXP);
+extern SEXP XGPointerEqComparison(SEXP, SEXP);
 extern SEXP XGBoosterTrainOneIter_R(SEXP, SEXP, SEXP, SEXP, SEXP);
 extern SEXP XGBoosterCreate_R(SEXP);
-extern SEXP XGBoosterCreateInEmptyObj_R(SEXP, SEXP);
+extern SEXP XGBoosterCopyInfoFromDMatrix(SEXP, SEXP);
```
trivialfis, line 24: Let's keep the naming consistent for R-specific C functions, `_R()`.

R-package/R/xgb.Booster.R
david-cortes: Removed.

It was BTW only duplicating the R attributes at the end, not the whole object.

R-package/src/xgboost_R.h
```diff
@@ -174,13 +202,37 @@ XGB_DLL SEXP XGDMatrixGetDataAsCSR_R(SEXP handle);
...
+ * \brief copy information about features from a DMatrix into a Booster
+ * \param booster R 'externalptr' pointing to a booster object
+ * \param dmat R 'externalptr' pointing to a DMatrix object
+ */
+XGB_DLL SEXP XGBoosterCopyInfoFromDMatrix(SEXP booster, SEXP dmat);
+
+/*!
+ * \brief handle R 'externalptr' holding the booster object
+ * \param field field name
+ * \param features features to set for the field
```
trivialfis, line 215: I think this is feature or features based on the dispatching logic you have in the R C function.

R-package/R/xgb.Booster.R
david-cortes: Changed to `xgb.load.raw`.

R-package/R/xgb.Booster.R
david-cortes: Yes.

R-package/R/xgb.Booster.R
david-cortes: Removed.

R-package/R/callbacks.R
```diff
@@ -440,8 +440,12 @@ cb.save.model <- function(save_period = 0, save_name = "xgboost.model") {
       stop("'save_model' callback requires the 'bst' booster object in its calling frame")
 
     if ((save_period > 0 && (env$iteration - env$begin_iteration) %% save_period == 0) ||
-        (save_period == 0 && env$iteration == env$end_iteration))
-      xgb.save(env$bst, sprintf(save_name, env$iteration))
+        (save_period == 0 && env$iteration == env$end_iteration)) {
+      suppressWarnings({
```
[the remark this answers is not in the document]
trivialfis: Got it, thank you for sharing, maybe a simple comment for future comers?

R-package/R/xgb.Booster.R
david-cortes: Switched to R lists instead of JSON text.

R-package/R/xgb.Booster.R
david-cortes: Removed.

R-package/src/init.c
david-cortes: Renamed.

R-package/src/xgboost_R.cc
```diff
@@ -597,6 +605,19 @@ XGB_DLL SEXP XGDMatrixNumCol_R(SEXP handle) {
   return ScalarInteger(static_cast<int>(ncol));
 }
 
+XGB_DLL SEXP XGDuplicate_R(SEXP obj) {
+  return Rf_duplicate(obj);
```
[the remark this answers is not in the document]
david-cortes: AFAIK it's not possible to call R's own C functions like that.

R-package/src/xgboost_R.cc
```diff
@@ -665,14 +686,136 @@ XGB_DLL SEXP XGDMatrixGetDataAsCSR_R(SEXP handle) {
...
+}
+
+const char *ubj_json_format_str = "{\"format\": \"ubj\"}";
+
+SEXP XGBAltrepSerializer_R(SEXP R_altrepped_obj) {
+  R_API_BEGIN();
+  BoosterHandle handle = R_ExternalPtrAddr(R_altrep_data1(R_altrepped_obj));
+  char const *serialized_bytes;
+  bst_ulong serialized_length;
+  CHECK_CALL(XGBoosterSaveModelToBuffer(
```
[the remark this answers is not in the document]
david-cortes: No, that's handled by R, since those are R attributes that aren't elements/slots in the object. I've added a test about it just in case.

R-package/src/xgboost_R.cc
```diff
@@ -665,14 +686,136 @@ XGB_DLL SEXP XGDMatrixGetDataAsCSR_R(SEXP handle) {
...
+  if (!deep) {
+    SEXP out = Rf_protect(XGBMakeEmptyAltrep());
+    R_set_altrep_data1(out, R_altrep_data1(R_altrepped_obj));
+    Rf_unprotect(1);
+    return out;
+  } else {
+    SEXP out = Rf_protect(XGBMakeEmptyAltrep());
+    char const *serialized_bytes;
+    bst_ulong serialized_length;
+    CHECK_CALL(XGBoosterSaveModelToBuffer(
```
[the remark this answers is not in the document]
david-cortes: No, that's again handled by R.

R-package/R/xgb.Booster.R
trivialfis: Then is it necessary to cache it in a list?

R-package/src/xgboost_R.h
david-cortes: But that's the "field" content.

R-package/R/xgb.train.R
david-cortes: It's not necessary, but a user that is aiming for minimal footprint (e.g. when serving the model) might want not to add any of these.

R-package/R/xgb.train.R
david-cortes: No, but they will be removed in the follow-up PR.

R-package/R/xgb.Booster.R
david-cortes: Added `variable.names` method for `xgb.Booster`.

R-package/R/callbacks.R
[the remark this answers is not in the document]
david-cortes: Added a comment.

trivialfis:
Getting close to 4am for me now, will continue tomorrow.

R-package/R/xgb.Booster.R
david-cortes: Yes, it's necessary to return it there, because in the case of updates, it gets reset to zero, but then there is a condition that checks:
```r
if (is_update && nrounds > niter_init)
    stop("nrounds cannot be larger than ", niter_init, " (nrounds of xgb_model)")
```

So, if one calls `XGBoosterBoostedRounds` after setting the parameters (which resets the rounds to zero in the case of updates), it will return zero and thus will not be possible to check that condition.

Or is there perhaps a better way of obtaining the current number of boosting iterations in order to check for that condition when performing an update?

R-package/R/xgb.Booster.R
david-cortes: Updated the comment.

david-cortes:
@trivialfis Looks like this repository commited `.Rnw`  files, which are auto-generated by `knitr` from the `.Rmd` files.

R-package/src/xgboost_R.cc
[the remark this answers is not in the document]
trivialfis: Got it, just got curious.

R-package/R/xgb.unserialize.R
```diff
@@ -1,23 +1,14 @@
...
-  } else {
-    if (!is.null.handle(handle))
-      stop("'handle' is not null/empty. Cannot overwrite existing handle.")
-    .Call(XGBoosterCreateInEmptyObj_R, cachelist, handle)
-  }
+  bst <- .Call(XGBoosterCreate_R, cachelist)
   tryCatch(
-    .Call(XGBoosterUnserializeFromBuffer_R, handle, buffer),
+    .Call(XGBoosterUnserializeFromBuffer_R, xgb.get.handle(bst), buffer),
     error = function(e) {
```
trivialfis, line 12: I think we can remove this warning now since all the compatibility is gone.

R-package/R/xgb.train.R
trivialfis: Got it, thank you for sharing.

R-package/R/xgb.save.R
```diff
@@ -1,26 +1,38 @@
...
-#' @param fname name of the file to write.
+#' @param model Model object of \code{xgb.Booster} class.
+#' @param fname Name of the file to write.
+#'
+#' Note that the extension of this file name determined the serialization format to use:\itemize{
+#' \item Extension ".ubj" will use the universal binary JSON format (recommended).
+#' This format uses binary types for e.g. floating point numbers, thereby preventing any loss
+#' of precision when converting to a human-readable JSON text or similar.
+#' \item Extension ".json" will use plain JSON, which is a human-readable format.
+#' \item Extension ".model" will use a \bold{deprecated} binary format. This format will
```
trivialfis, line 13:
```suggestion
#' \item Extension ".deprecated" will use a \bold{deprecated} binary format. This format will
```

R-package/R/xgb.train.R
trivialfis: Do we need to concern about these attributes when serving the model?

R-package/src/xgboost_R.cc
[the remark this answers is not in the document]
trivialfis: Got it, thank you for sharing.

R-package/R/xgb.Booster.R
trivialfis: Nice!

R-package/R/xgb.Booster.R
trivialfis: Thank you for the change!

R-package/R/xgb.Booster.R
trivialfis: Since this function is a part of the public interface, I think we should think of a way to avoid unnecessary publicly accessible attributes. I don't have a good solution to the update process yet (has been like this for a while), but a quick workaround as an internal hack could be setting it as an internal booster attribute and deleting it after the check.

david-cortes: Function `xgb.Booster` is not within the exported namespace, it's only meant for internal usage.

R-package/R/xgb.train.R
david-cortes: No, they are not used in `predict` nor importance nor plotting, which is why I wanted to make them optional.

But on second thought, it should be pretty straightforward for a user to remove them, so I've removed this parameter.

david-cortes:
> > Looks like this repository commited .Rnw files, which are auto-generated by knitr from the .Rmd files.
> 
> Feel free to remove it if it's not required to be part of the repository.

Removed.

R-package/R/xgb.unserialize.R
trivialfis: Sounds good!

R-package/R/xgb.Booster.R
trivialfis: thank you for sharing, then it's fine.

trivialfis (approved):

david-cortes:
@trivialfis any comments about the tests that download a file with serialized models and where `.Rds` cases are now skipped?

trivialfis:
Thank you for the excellent work! Will merge once the buildkite tests pass. (probably tomorrow).

trivialfis:
> any comments about the tests that download a file with serialized models and where .Rds cases are now skipped?

No thoughts yet, we will upload new files once the refactoring is finished and restore those tests. I think we will continue to try to avoid having any binary file in the git repo.

R-package/R/utils.R
```diff
@@ -374,10 +379,6 @@ NULL
 #' The \code{\link{xgb.save.raw}} function is useful if you'd like to persist the XGBoost model
 #' as part of another R object.
 #'
-#' Note: Do not use \code{\link{xgb.serialize}} to store models long-term. It persists not only the
-#' model but also internal configurations and parameters, and its format is not stable across
-#' multiple XGBoost versions. Use \code{\link{xgb.serialize}} only for checkpointing.
-#'
```
hcho3, line 380: What's the replacement for `xgb.serialize`? Should users use `saveRDS` to fully preserve all attributes?

hcho3 (approved):

hcho3:
I find the new ALTREP serializer more ergonomic, with the clean handling of the external handles.

Question: Is the R serializer used in `saveRDS` considered stable? For example, can I serialize a list consisting of multiple attributes using `saveRDS` and read it in a later R version?

trivialfis:
> Yes, users should rely on R serializers like saveRDS in order to keep the R attributes, just like with regular R objects.

Thank you for sharing. Although currently in this PR the parameters are handled by R attributes as "extra attributes", but later on, I think I will revise the serialization functions in C to use something similar to the `xgb.serialize` and `xgb.unserialize` function for parameters.

trivialfis:
@david-cortes One example would be:
``` R
library(xgboost)
nrounds <- 1

set.seed(1994)

n_samples <- 10

d <- cbind(
  x1 = rnorm(n_samples),
  x2 = rnorm(n_samples),
  x3 = rnorm(n_samples)
)
y <- d[, "x1"] + d[, "x2"]^2 +
  ifelse(d[, "x3"] > .5, d[, "x3"]^2, 2^d[, "x3"]) +
  rnorm(n_samples)

y <- rbinom(n = n_samples, size = 1, prob = 1 / 2)

dtrain <- xgb.DMatrix(data = d, label = y)

booster <- xgb.train(
  params = list(
    tree_method = "approx",
    objective = "binary:logistic"
  ),
  data = dtrain,
  nrounds = nrounds
)
# approx
xgb.config(booster)$learner$gradient_booster$gbtree_train_param$tree_method

saveRDS(booster, "model.rds")
booster0 <- readRDS("model.rds")
# auto
xgb.config(booster0)$learner$gradient_booster$gbtree_train_param$tree_method
```
The parameters are lost during RDS IO.

trivialfis:
> But that's also the case for UBJ serialization:

We have two distinguished cases as mentioned previously:
- saving the model.
- saving everything.

The `xgb.save` saves the model, and nothing else. It's used to "export" a model for other tasks like inference. `xgb.serialize` is used to save everything, that the user can continue training without any intervention.

david-cortes:
> > But that's also the case for UBJ serialization:
> 
> We have two distinguished cases as mentioned previously:
> 
> * saving the model.
> * saving everything.
> 
> The `xgb.save` saves the model, and nothing else. It's used to "export" a model for other tasks like inference, explanation. `xgb.serialize` is used to save everything, that the user can continue training without any intervention, including things like whether to use GPU.

Thanks for pointing this out.

What would be the python equivalent of this function?

trivialfis:
pickle.

trivialfis:
specifically, `__setstate__` and `__getstate__`.

trivialfis:
I'm quite excited about this PR, and am comfortable with merging it regarding the interface change, this way we can unblock other PRs and avoid too many rebase work. Do you want to merge it and leave the RDS issue in a follow-up PR or do you want to have it solved in this one (I can help with the serialization)?

david-cortes:
I've added a small note about the R attribute `params`.

I am wondering however if there's some C-level function that would automatically gather and serialize all C attributes that aren't handled by `XGBoosterSaveModelToBuffer`.

This step could be done in the same ALTREP serializer if we know what else needs to be there, but I see that the python class has many things that it keeps in attributes and I'm not sure which ones are C things that need to be re-assigned. Does it need to also gather `xgb.parameters`, for example? (there's currently no getter for that in R).

> Do you want to merge it and leave the RDS issue in a follow-up PR

Let's better do it in this PR to keep all serialization-related things together.

trivialfis:
@david-cortes See https://github.com/dmlc/xgboost/pull/9924#discussion_r1445104757 .

```
XGBoosterSerializeToBuffer
XGBoosterUnserializeFromBuffer
```

It saves everything that's defined in C, including parameters and attributes. It's used by the now removed `xgb.serialize` R function, and: https://github.com/dmlc/xgboost/blob/01c471155627f9d81c230d0a16a7655c9fdd16c4/python-package/xgboost/core.py#L1811

trivialfis:
I think we can use it here: https://github.com/dmlc/xgboost/pull/9924/files#r1445141247 .

david-cortes:
Changed it to use `XGBoosterSerializeToBuffer` and `XGBoosterUnserializeFromBuffer` instead of `XGBoosterSaveModelToBuffer`.

Seems to pass the tests and work as expected.

hcho3:
We should probably retain the note about backward compatibility: if saveRDS is used, the model may not be fully usable in a future version of XGBoost.

david-cortes:
> We should probably retain the note about backward compatibility: if saveRDS is used, the model may not be fully usable in a future version of XGBoost.

Is that because `XGBoosterSerializeToBuffer` doesn't have such guarantee? If so, what happens when a user tries to load an incompatible model? Is there such an incompatible serialized booster that I could download from somewhere?

trivialfis:
When loading from a different version of XGBoost, the internal parameters are automatically discarded, in which case, it's the same as `xgb.save`.

david-cortes:
> When loading from a different version of XGBoost, the internal parameters are automatically discarded, in which case, it's the same as `xgb.save`.

Ok, then I guess it's no big deal, and the compatibility note makes sense to keep.

trivialfis:
See https://github.com/dmlc/xgboost/issues/9734#issuecomment-1786427206

hcho3:
Then let's drop the warning about saveRDS ?

david-cortes:
> Then let's drop the warning about saveRDS ?

I've rewritten the compatibility note to reflect the current situation per my understanding. Would be ideal if you and @trivialfis could take a look at it. It doesn't have a warning to _not_ use R's serializers, but describes a bit the differences.

I think it's still important to have it, since now we have an incompatibility with models that were created before this PR, and an incompatibility with older versions of `qs` which is also used for serialization.

R-package/R/utils.R
```diff
@@ -349,21 +349,41 @@ xgb.createFolds <- function(y, k) {
...
+#' \item Additional R-specific attributes  - e.g. results of callbacks, such as evaluation logs,
+#' which are kept as a `data.table` object, accessible through `attributes(model)$evaluation_log`
+#' if present.
+#' }
+#'
+#' The first ones (configuration attributes) do not have the same compatibility guarantees as
+#' attributes that are set and accessed through \link{xgb.attributes} - that is, such attributes
+#' might be lost after loading the booster in a different XGBoost version, regardless of the
+#' serializer that was used. Note that these are saved when using \link{xgb.save}, but not when
+#' using \link{xgb.save.raw}.
```
hcho3, line 373: I don't think either `xgb.save` or `xgb.save.raw` preserves configuration attributes.

@trivialfis Any thoughts?

hcho3:
Thanks for the new wording. I recognize that some use cases do call for the use of `saveRDS`.

R-package/R/utils.R
trivialfis: `xgb.save` is the same as `xgb.save.raw`, the former writes to a file while the latter writes to a memory buffer. You are right that they don't save configuration. Only the removed `xgb.serialize` does.

trivialfis: See https://github.com/dmlc/xgboost/pull/9924#discussion_r1445104757

david-cortes: My bad, I've updated the note.

R-package/R/utils.R
```diff
@@ -352,16 +349,45 @@ xgb.createFolds <- function(y, k) {
...
+#' \link{save} or \link{saveRDS} to serialize an XGBoost R model, but XGBoost also provides
+#' its own serializers with perhaps better compability guarantees and which allow loading
+#' said models in other language bindings of XGBoost.
+#'
+#' Note that an `xgb.Booster` object, outside of its core components, might also keep:\itemize{
+#' \item Additional model configuration attributes (accessible through \link{xgb.config}),
+#' which might be used during model fitting but are not used in e.g. `predict`, feature importance,
+#' or plotting methods.
+#' \item Additional R-specific attributes  - e.g. results of callbacks, such as evaluation logs,
+#' which are kept as a `data.table` object, accessible through `attributes(model)$evaluation_log`
```
trivialfis, line 365:
```suggestion
#' @description
#'
#' When it comes to serializing XGBoost models, it's possible to use R serializers such as
#' \link{save} or \link{saveRDS} to serialize an XGBoost R model, but XGBoost also provides
#' its own serializers with better compatibility guarantees, which allow loading
#' said models in other language bindings of XGBoost.
#'
#' Note that an `xgb.Booster` object, outside of its core components, might also keep:\itemize{
#' \item Additional model configuration (accessible through \link{xgb.config}),
#' which includes model fitting parameters like `max_depth` and runtime parameters like `nthread`.
#' These are not necessarily useful for prediction/importance/plotting.
#' \item Additional R-specific attributes  - e.g. results of callbacks, such as evaluation logs,
#' which are kept as a `data.table` object, accessible through `attributes(model)$evaluation_log`
```

R-package/R/utils.R
```diff
@@ -352,16 +349,45 @@ xgb.createFolds <- function(y, k) {
...
+#' purposes, such as keeping track of evaluation metrics as the model was fit, or saving the R
+#' call that produced the model, but are otherwise not used for prediction / importance / plotting / etc.
+#' These R attributes are only preserved when using R's own serializers.
+#'
+#' Note that XGBoost models in R starting from version `2.1.0` and onwards, and XGBoost models
+#' before version `2.1.0`; have a very different R object structure and are incompatible with
+#' each other. Hence, models that were saved with R serializers live `saveRDS` or `save` before
+#' version `2.1.0` will not work with latter `xgboost` versions and vice versa. Be aware that
+#' the structure of R model objects could in theory again in the future, so XGBoost's serializers
+#' should be preferred for very long-term storage.
```
trivialfis, line 387:
```suggestion
#' The first one (configurations) does not have the same compatibility guarantees as
#' the model itself, including attributes that are set and accessed through \link{xgb.attributes} - that is, such configuration
#' might be lost after loading the booster in a different XGBoost version, regardless of the
#' serializer that was used. These are saved when using \link{saveRDS}, but will be discarded
#' if loaded into an incompatible XGBoost version. They are not saved when using XGBoost's
#' serializers from its public interface including \link{xgb.save} and \link{xgb.save.raw}.
#'
#' The second ones (R attributes) are not part of the standard XGBoost model structure, and thus are
#' not saved when using XGBoost's own serializers. These attributes are only used for informational
#' purposes, such as keeping track of evaluation metrics as the model was fit, or saving the R
#' call that produced the model, but are otherwise not used for prediction / importance / plotting / etc.
#' These R attributes are only preserved when using R's serializers.
#'
#' Note that XGBoost models in R starting from version `2.1.0` and onwards, and XGBoost models
#' before version `2.1.0`; have a very different R object structure and are incompatible with
#' each other. Hence, models that were saved with R serializers live `saveRDS` or `save` before
#' version `2.1.0` will not work with latter `xgboost` versions and vice versa. Be aware that
#' the structure of R model objects could in theory change again in the future, so XGBoost's serializers
#' should be preferred for long-term storage.
```

R-package/R/utils.R
trivialfis: Mostly just trying to avoid referring to configuration as attributes, in my opinion, it causes less confusion.

PR #9929, opened 2023-12-27
[R] Add missing DMatrix functions

david-cortes:

This PR adds R wrappers over some DMatrix functions that were not available from the R interface:
* `XGDMatrixGetQuantileCut`
* `XGDMatrixNumNonMissing`
* `XGDMatrixGetDataAsCSR`

For the first one, I see that xgboost returns a JSON following the array interface from NumPy, which includes the pointer address as an unsigend 64-bit integer.

I see however that xgboost's JSON parser uses **signed** 64-bit integers (both internally and in the JSON code from this PR).
* Is it guaranteed that a platform won't have a larger pointer address space than 64-bits?
* If the pointer address happens to be beyond $2^{63}$, wouldn't the conversion trigger undefined C++ behavior? I have a feeling that this sort of conversion might be catched by UBSAN or valgrind, and CRAN might ask to fix it later on.


trivialfis:
> Is it guaranteed that a platform won't have a larger pointer address space than 64-bits

2^63 is a big number, I don't think we will have memory of that size in the foreseeable future, but do prove me wrong!

david-cortes:
> > Is it guaranteed that a platform won't have a larger pointer address space than 64-bits
> 
> 2^63 is a big number, I don't think we will have memory of that size in the foreseeable future, but do prove me wrong!

But regardless of the size of the data, if one calls allocators like `malloc` or `new`, those will output random addresses within the machine's range, which will then be passed to the JSON parser.

trivialfis:
> For the first one, I see that xgboost returns a JSON following the array interface from NumP

For context, other returns use thread local memory, which is not released until the thread exits, which might be undesirable for the main feature matrix.

david-cortes:
> > > Is it guaranteed that a platform won't have a larger pointer address space than 64-bits
> > 
> > 
> > 2^63 is a big number, I don't think we will have memory of that size in the foreseeable future, but do prove me wrong!
> 
> But regardless of the size of the data, if one calls allocators like `malloc` or `new`, those will output random addresses within the machine's range, which will then be passed to the JSON parser.

Actually, I am not so sure about how this sort of thing works.

Is it guaranteed that a platform will always provide a contiguous numeration for a process' address space within the machine's RAM range, or can it be randomized within the whole range of the `size_t` range for a specific process?

trivialfis:
> Is it guaranteed that a platform will always provide a contiguous numeration for a process' address space within the machine's RAM range, or can it be randomized/reassigned within the whole range of the size_t range for a specific process?

It can be randomized, for security reasons I think (not an expert).

However, since `std::ptrdiff_t` is a signed integer, most of the implementations restrict the possible value of pointers.

R-package/R/xgb.DMatrix.R
```diff
@@ -528,6 +528,135 @@ setinfo.xgb.DMatrix <- function(object, name, info, ...) {
...
+#' As such, the first and last bin will be outside of the range of the data, so as to include
+#' all of the observations there.
+#'
+#' If a given column has 'n' bins, then there will be 'n+1' cuts / borders for that column,
+#' which will be output in sorted order from lowest to highest.
+#'
+#' Different columns can have different numbers of bins according to their range.
+#' @param dmat An `xgb.DMatrix` object, as returned by \link{xgb.DMatrix}.
+#' @param output Output format for the quantile cuts. Possible options are:\itemize{
+#' \item `"list"` will return the output as a list with one entry per column, where
```
trivialfis, line 546: I think we are moving toward the markdown syntax?

david-cortes:
Ok, then I guess using signed int64 should not pose any kind of problem for the foreseeable future.

R-package/R/xgb.DMatrix.R
```diff
@@ -528,6 +528,135 @@ setinfo.xgb.DMatrix <- function(object, name, info, ...) {
   stop("setinfo: unknown info name ", name)
 }
 
+#' @title Get Quantile Cuts from DMatrix
```
trivialfis, line 531: Is it necessary to return different formats? I think it might be better off to pick one that's most efficient and let the users handle the rest of the conversion.

R-package/src/xgboost_R.cc
```diff
@@ -540,6 +669,75 @@ XGB_DLL SEXP XGDMatrixNumCol_R(SEXP handle) {
...
+  if (std::max(nrows, ncols) > std::numeric_limits<int>::max()) {
+    Rf_error("%s", "Error: resulting DMatrix data does not fit into R 'dgRMatrix'.");
+  }
+
+  SET_VECTOR_ELT(out, 0, Rf_allocVector(INTSXP, nrows + 1));
+  SET_VECTOR_ELT(out, 1, Rf_allocVector(INTSXP, nnz));
+  SET_VECTOR_ELT(out, 2, Rf_allocVector(REALSXP, nnz));
+  SET_VECTOR_ELT(out, 3, Rf_ScalarInteger(ncols));
+
+  std::unique_ptr<bst_ulong[]> indptr(new bst_ulong[nrows + 1]);
```
trivialfis, line 722: why not just `std::vector`? Or better yet, copy the data directly into `out`?

R-package/src/xgboost_R.cc
```diff
@@ -136,6 +162,109 @@ SEXP SafeMkChar(const char *c_str, SEXP continuation_token) {
...
+  const bool is_int = std::is_same<DestDType, int>::value;
+  SEXP out = Rf_protect(
+    is_int? SafeAllocInteger(len, continuation_token) : SafeAllocReal(len, continuation_token));
+  void *ptr_out = is_int? static_cast<void*>(INTEGER(out)) : static_cast<void*>(REAL(out));
+  XGCopyTypedArray<SourceDType, DestDType>(src, ptr_out, len, stride);
+  Rf_unprotect(1);
+  return out;
+}
+
+SEXP XGArrayStrToROneDimensional(const char *array_str, SEXP ctoken) {
```
trivialfis, line 193: We have some code in XGBoost to deal with the array interface `src/data/array_interface.h`.

R-package/src/xgboost_R.cc
david-cortes: `std::vector` zero-initializes the memory, while `new[]` doesn't. The types are different so it's not possible to copy it directly in the R vectors.

R-package/src/xgboost_R.cc
david-cortes: But would it allow simplifying something here? I see it has the same kind of fields so for example there still would need to be a `switch` to make the conversion to R types.

R-package/R/xgb.DMatrix.R
david-cortes: I think its good to have a user-friendly format and a more efficient format depending on the use-case.

R-package/R/xgb.DMatrix.R
david-cortes: Perhaps @mayer79 could revisit this one later.

R-package/src/xgboost_R.cc
trivialfis: Thank you for the explanation.

R-package/src/xgboost_R.cc
trivialfis: Let me give it a try tomorrow.

R-package/R/xgb.DMatrix.R
trivialfis: I'm not sure `csc` is a good format for this return, 0 is a valid value in cuts. XGBoost can split with 0 as the threshold.

R-package/R/xgb.DMatrix.R
trivialfis: I'm not sure `csc` is a good format for this return, 0 is a valid value in cuts. XGBoost can split with 0 as the threshold.

david-cortes: These R CSC matrices can have explicit zeros when one creates them by manually assigning to the slots instead of using the checked constructors.

trivialfis: The problem is extra zeros are padded into the end.

trivialfis: It's not exactly a CSC matrix, or a matrix at all. I'm not comfortable with coercing it into one without good use cases.

david-cortes: Yes, that's why I set the "list" format to be the default.

david-cortes: My idea for this CSC format was to be able to use the `print` method from `Matrix`, which displays the information in a way that's visually easier to parse.

trivialfis: Is it necessary to be done in this function? I prefer using helper functions for wrangling data. XGBoost itself should try return its data with minimal ambiguity.

david-cortes: Removed this CSC format.

R-package/R/xgb.DMatrix.R
```diff
@@ -526,6 +526,111 @@ setinfo.xgb.DMatrix <- function(object, name, info) {
...
+#'
+#' # Now can get the quantile cuts
+#' xgb.get.DMatrix.qcut(dm)
+#' @export
+xgb.get.DMatrix.qcut <- function(dmat, output = c("list", "arrays")) { # nolint
+  stopifnot(inherits(dmat, "xgb.DMatrix"))
+  output <- head(output, 1L)
+  stopifnot(output %in% c("list", "arrays"))
+  res <- .Call(XGDMatrixGetQuantileCut_R, dmat)
+  if (output == "arrays") {
```
trivialfis, line 578: The default is list and arrays instead of list or array? `c("list", "arrays")`.

trivialfis:
Changes in my commit:
- Use array interface handler.
- Check the return from XGBoost is contiguous.
- Remove the empty name in `out_names`.

david-cortes:
> Brought back the extra name in `out_names`, not sure how it works.

The R C function `mkNamed` doesn't have a 'length' parameter, so it determines the end of the array by looking for an empty string.

david-cortes:
This is the error from the windows job:
> array_interface.obj : error LNK2005: "public: static void __cdecl xgboost::ArrayInterfaceHandler::SyncCudaStream(__int64)" (?SyncCudaStream@ArrayInterfaceHandler@xgboost@@SAX_J@Z) already defined in xgboost_R.obj [C:\buildkite-agent\builds\buildkite-windows-cpu-autoscaling-group-i-08f7f49d2fc54b768-1\xgboost\xgboost-ci-windows\build\xgboost.vcxproj]

Looks like the header might need some `ifdef` guard or similar.

R-package/R/xgb.DMatrix.R
david-cortes: It takes only the first argument. It's quite common in R function signatures to set the default to a vector with all possible values and take only the first one, such as in method `predict.glm`, or even in other xgboost functions like `xgb.dump`.

R-package/src/xgboost_R.cc
```diff
@@ -136,6 +162,36 @@ SEXP SafeMkChar(const char *c_str, SEXP continuation_token) {
...
+  // R supports only int and double.
+  bool is_int =
+      xgboost::DispatchDType(array.type, [](auto t) { return std::is_integral_v<decltype(t)>; });
+  bool is_float = xgboost::DispatchDType(
+      array.type, [](auto v) { return std::is_floating_point_v<decltype(v)>; });
+  CHECK(is_int || is_float) << "Internal error: Invalid DType.";
+  // Allocate memory in R
+  SEXP out =
+      Rf_protect(is_int ? SafeAllocInteger(array.n, ctoken) : SafeAllocReal(array.n, ctoken));
+  CHECK(array.is_contiguous) << "Return by XGBoost should be contiguous";
```
david-cortes, line 178: I think this one could be moved to one line before.

trivialfis:
> Looks like the header might need some ifdef guard or similar.

No, it's just MSVC being MSVC, I will try to work around it.

R-package/src/xgboost_R.cc
trivialfis: Wiil move the check.

R-package/R/xgb.DMatrix.R
trivialfis: Thank you for the detailed explanation!

trivialfis (approved):

david-cortes:
@trivialfis On a second look, there's one potential issue in the new array interface wrapper: all integer types are translated into R's `INTEGER`, which is a signed int32, but the arrays might contain values outside of the range of this type.

Maybe it should add a check for whether the maximum value (which would be the last entry in the `indptr` array) is within the range of a signed 32-bit integer, and use `float64` otherwise.

PR #9930, opened 2023-12-27
[R] Enable 'dot' dump format

david-cortes:

This PR allows using the 'dot' format for dumping models to text, and generating plots from these dumps like in the python interface.

Note that this doesn't solve all the issues with plotting when there are categorical or named features, as other functions rely on `xgb.model.dt.tree` which as of this PR still doesn't support all possible xgboost model dumps.


trivialfis (approved):
Thank you for working on the plotting function!

trivialfis:
Will merge once the CI is back.

Issue #9618, opened 2023-09-28
[CI] Failing tests are not reported accordingly on Windows.


trivialfis:
https://buildkite.com/xgboost/xgboost-ci-windows/builds/3566#018adc46-b6b7-423a-937c-4f35bbb67575

hcho3:
I need to study how exit codes work in PowerShell.

trivialfis:
> I need to study how exit codes work in PowerShell.

Or write Python instead. ;-)

PR #9932, opened 2023-12-28
[CI] Update machine images

hcho3:
* Use R 4.3 in the Windows machine
* Rebuild machine images using latest base OS images
* Update the launch parameter according to the latest Elastic AWS Stack of BuildKite.
* Enable `BUILDKIT_INLINE_CACHE` option to enable caching layers from pulled containers. This is needed for latest Docker, which uses BuildKit (https://github.com/moby/moby/issues/39003).
* `dlltool.exe` was failing due to long file names. 1) Error out early if `dlltool` fails. 2) Use a short file name for the temporary file.
* Speed up RPM by setting `ulimit` option. See https://github.com/docker/buildx/issues/379
* Use latest NVIDIA Container Toolkit. As a side effect, the legacy command `nvidia-docker` is no longer available. Now we should use `docker run --gpus all` instead.

### DIFF
--- a/cmake/modules/FindLibR.cmake
+++ b/cmake/modules/FindLibR.cmake
@@ -65,7 +65,9 @@ function(create_rlib_for_msvc)
 
   execute_process(COMMAND ${DLLTOOL_EXE}
     "--input-def" "${CMAKE_CURRENT_BINARY_DIR}/R.def"
-    "--output-lib" "${CMAKE_CURRENT_BINARY_DIR}/R.lib")
+    "--output-lib" "${CMAKE_CURRENT_BINARY_DIR}/R.lib"
+    COMMAND_ECHO STDOUT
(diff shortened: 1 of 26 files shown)


trivialfis:
Could you please help take a look into https://github.com/dmlc/xgboost/issues/9618 , it would be great if we can restore the Windows tests.

hcho3:
> Could you please help take a look into https://github.com/dmlc/xgboost/issues/9618

I will take a look at it once we unblock the rest of the CI.

hcho3:
@trivialfis Sorry this is taking a bit longer than I expected. Please refrain from running CI jobs until the CI is straightened out.

trivialfis:
Please disable the allocator test, it doesn't always work.

hcho3:
Merging this to unblock the CI

PR #9935, opened 2023-12-29
[R] Remove parameters and attributes related to `ntree` and rebase `iterationrange`

david-cortes:

After the introduction of newer tree modalities such as multi-quantile regression, the code for determining the number of trees in a model is no longer correct, and other sections that rely on it might produce incorrect results. I see that parameters referring to number of trees have been deprecated in favor of parameters referring to number of iterations, so I'm making the switch here.

This PR:
* Removes the deprecated argument `ntree_limit` in `predict`.
* Removes the 'ntree' attribute and internal accessor which are no longer needed.
* Fixes incorrect usage of `num_class` to determine prediction shapes, as now there are more ways in which `predict` can output multi-dimensional results.
* Changes the format of `iterationrange` to match with R's sequences/ranges.

I've based it off from the current last commit of previous PR https://github.com/dmlc/xgboost/pull/9924 which is a requisite for the changes introduced here.

I'm not sure how to make the PR show the diff w.r.t to that other PR, as it's not a branch of this repository, and I cannot open a PR here to merge to a branch on my own repository, so this PR will need to be rebased later on in order to make it mergable (I think the kind of commits that github generates will also mess up `git merge` later on so a rebase will anyway be needed).

### DIFF
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -282,9 +282,8 @@ predict.xgb.Booster <- function(object, newdata, missing = NA, outputmargin = FA
 
 
   if (!is.null(iterationrange)) {
-    if (is.character(iterationrange) &&
-        length(iterationrange) == 1 &&
-        iterationrange == "all") {
+    if (is.character(iterationrange)) {
+      stopifnot(iterationrange == "all")
       iterationrange <- c(0, 0)
     } else {
       iterationrange[1] <- iterationrange[1] - 1 # base-0 indexing
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -89,7 +89,6 @@ xgb.get.handle <- function(object) {
 #' @param outputmargin Whether the prediction should be returned in the form of original untransformed
 #'        sum of predictions from boosting iterations' results. E.g., setting `outputmargin=TRUE` for
 #'        logistic regression would return log-odds instead of probabilities.
-#' @param ntreelimit Deprecated, use `iterationrange` instead.
 #' @param predleaf Whether to predict pre-tree leaf indices.
 #' @param predcontrib Whether to return feature contributions to individual predictions (see Details).
 #' @param approxcontrib Whether to use a fast approximation for feature contributions (see Details).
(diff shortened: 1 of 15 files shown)


hcho3:
> I'm not sure how to make the PR show the diff w.r.t to that other PR, as it's not a branch of this repository, and I cannot open a PR here to merge to a branch on my own repository, so this PR will need to be rebased later on in order to make it mergable

It's fine. It's common for us to open pull requests that include the commits from another pull request. Let's try to merge https://github.com/dmlc/xgboost/pull/9924 soon.

david-cortes:
@trivialfis Would be ideal if you could review this PR next.

R-package/R/xgb.Booster.R
```diff
@@ -99,11 +98,17 @@ xgb.get.handle <- function(object) {
 #'        or `predinteraction` is `TRUE`.
 #' @param training Whether the predictions are used for training. For dart booster,
 #'        training predicting will perform dropout.
-#' @param iterationrange Specifies which trees are used in prediction. For
-#'        example, take a random forest with 100 rounds.
-#'        With `iterationrange=c(1, 21)`, only the trees built during `[1, 21)` (half open set)
-#'        rounds are used in this prediction. The index is 1-based just like an R vector. When set
-#'        to `c(1, 1)`, XGBoost will use all trees.
+#' @param iterationrange Sequence of rounds/iterations from the model to use for prediction, specified by passing
+#'        a two-dimensional vector with the start and end numbers in the sequence (same format as R's `seq` - i.e.
```
trivialfis, line 102: Thank you for making it R-compatible. These indexing schemes have been a headache.

R-package/R/xgb.Booster.R
```diff
@@ -275,25 +280,22 @@ predict.xgb.Booster <- function(object, newdata, missing = NA, outputmargin = FA
...
-    ## only iteration range, handle 1-based indexing
-    iterationrange <- c(iterationrange[1] - 1, iterationrange[2] - 1)
-  } else if (ntreelimit != 0 && !is.null(iterationrange)) {
-    ## both are specified, let libgxgboost throw an error
+  if (!is.null(iterationrange)) {
+    if (is.character(iterationrange) &&
+        length(iterationrange) == 1 &&
+        iterationrange == "all") {
+      iterationrange <- c(0, 0)
+    } else {
```
trivialfis, line 289: since the only valid value is "all", maybe we can check:
``` R
if (is.character(iterationrange)) {
    stopifnot(iterationrange == "all")
    iterationrange <- c(0, 0)
} else {
    ...
}
```

trivialfis (commented):
I'm looking into https://github.com/dmlc/xgboost/pull/9948 as well, would be great if there's a common reindexing function either in C or in R that handles all the translation and can be tested independently.

I find it difficult to reason that all places that need indexing are performing the translation consistently.

R-package/R/xgb.Booster.R
david-cortes: Modified.

david-cortes:
> I'm looking into #9948 as well, would be great if there's a common reindexing function either in C or in R that handles all the translation and can be tested independently.
> 
> I find it difficult to reason that all places that need indexing are performing the translation consistently.

Do you have a list of places where such reindexing should be applied?

As I see it, currently there is:
* `best_iteration` - but this one follows base-0 in the C attribute and base-1 in the R attribute, so handling would be quite different from the others.
* `iterationrange` - not too hard to do the conversion, since it always amounts to just subtracting from the initial index.
* Booster slicing - but this one requires a very different approach that the others do not share; and if it were to be re-implemented by taking an integer array instead of start/end/step, the handling would also be completely different if we want it to follow R's idioms.
* Class index in the linear coefficients callback (currently base-0) - but I think this one would make more sense to match with the numbers in `label` which are base-0.

trivialfis:
Thank you for sharing, it might be difficult to gather everything into one place. An additional concern is the categorical data.

It's not just 0-based indexing v.s. 1-based indexing, exclusive and inclusive on the end of a range is also problematic, would be great if we could at least find a place to document all the difference so that we can lookup in the future.

david-cortes:
> Thank you for sharing, it might be difficult to gather everything into one place. An additional concern is the categorical data.
> 
> It's not just 0-based indexing v.s. 1-based indexing, exclusive and inclusive on the end of a range is also problematic, would be great if we could at least find a place to document all the differences so that we can lookup in the future.

You mean in one of those `.rst` files from the readthedocs page for developers?

trivialfis:
> You mean in one of those .rst files from the readthedocs page for developers?

sounds good!

trivialfis (approved):

PR #9939, opened 2023-12-30
[R] Remove unusable 'feature_names' argument and make 'model' first argument in inspection functions

david-cortes:

As a follow up after previous PR is merged: https://github.com/dmlc/xgboost/pull/9924

(note that, like other PRs in progress, this one includes all of the commits from the PR above, and will need to be rebased after that one is merged)

In the previous PR:
* Feature names were added as C-level booster attributes, automatically added during model fitting.
* As a side effect of having these feature names, they now show up in the model dumps.
* Since the function that creates a table out of the model trees uses these dumps, it's not possible to override those feature names.

Some of the functions related to plotting and importances had a parameter `feature_names` which was passed to the trees-to-table function, which now is not usable and is not even required if the model already has the features names on it.

This PR removes this function argument when not applicable, and switches the `model` argument to be the first argument in the signature, which is a lot more convenient when using those functions as a user.


mayer79:
Very convenient, thx, I also found it strange that `model` was not the first argument.

trivialfis (commented):
> Feature names were added as C-level booster attributes, automatically added during model fitting.

Is there any test that shows hat happens if the input is a matrix? (for dump and plot)

david-cortes:
> > Feature names were added as C-level booster attributes, automatically added during model fitting.
> 
> Is there any test that shows hat happens if the input is a matrix? (for dump and plot)

Matrices can also have both column and row names. Nevertheless, there are tests for these functions in which the booster has the feature names removed:
https://github.com/dmlc/xgboost/blob/1168a688726aca01e6d52dd931bb63df06e3f110/R-package/tests/testthat/test_helpers.R#L54

PR #9940, opened 2023-12-31
[R] Minor improvements for evaluation printing

david-cortes:

This PR makes some small changes in the evaluation metrics printing:
* It uses a plus-minus sign for the standard deviation instead of '+', since that's what the standard deviation is meant to represent.
* It prints 'round' before the boosting round number in order to make these prints visually different from R's `print`.
* It uses R's `sd` function to calculate standard deviations, which correctly applies a bias correction for sampling bias and uses more numerically robust techniques than the current difference of squares.

### DIFF
--- a/R-package/R/callbacks.R
+++ b/R-package/R/callbacks.R
@@ -766,7 +766,7 @@ xgb.gblinear.history <- function(model, class_index = NULL) {
   enames <- names(eval_res)
   if (is.null(enames))
     stop('evaluation results must have names')
-  iter <- sprintf('[round: %d]\t', iter)
+  iter <- sprintf('[%d]\t', iter)
   if (!is.null(eval_err)) {
     if (length(eval_res) != length(eval_err))
       stop('eval_res & eval_err lengths mismatch')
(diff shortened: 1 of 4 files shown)


R-package/R/callbacks.R
```diff
@@ -766,11 +766,12 @@ xgb.gblinear.history <- function(model, class_index = NULL) {
   enames <- names(eval_res)
   if (is.null(enames))
     stop('evaluation results must have names')
-  iter <- sprintf('[%d]\t', iter)
+  iter <- sprintf('[round: %d]\t', iter)
```
trivialfis, line 769: I don't think this is necessary.

david-cortes: The thing is that if you print them as `[<number>] <space>`  then they look quite similar to R's own printing of character vectors, which can be a bit confusing.

trivialfis: That's probably fine for now, we would like to keep the output consistent across bindings at the moment. Besides, the output format is well-known for all other libraries.

david-cortes: Removed the `round: ` part.

I think the python interface could also show a plusminus sign instead of plus. Since python2 is not supported, I don't think it should create any issue there.

Issue #7547, opened 2022-01-09
[Roadmap] Phasing out the support for old binary format.


trivialfis:
XGBoost has a custom binary model format that has been used since day 1.  Later in 1.0, we
introduced the JSON format as an alternative, which has a schema and has better
extensibility.  The JSON format has been used as a default format for memory snapshot
serialization (pickle, rds, etc) and has extra features including categorical data support,
extra data feature names, and features types.  However, for performance and compatibility
reasons we have continued the support for the old binary format.  In 1.6 we plan to add
universal binary JSON as an extension to the current JSON format also as a replacement for the old
binary format.

## Motivation
The old binary format is essentially copying internal structures like parameters, tree
nodes into a memory buffer, so it has a fixed memory layout that's difficult to change and
debug.  If we look at the `Learner` class it's full of conditions to work around some
issues in binary format accumulated over the past.  These issues root from the situation
that we can not change the binary output in any way, which also has an indirect impact on
how we write code.  For instance, we can not change the `RegTree` structure due to how the
node is stored in the output and it's the very core of XGBoost.  To overcome these issues
and clear some room for future development we need to phase out its use.

## Roadmap

If the Universal Binary JSON implementation is accepted, I propose following the roadmap
for phasing out the support of the old binary format:

- 1.6: Add UBJSON and use it as the default format for full serialization (pickle, rds) as
  default.  Emit warning when users are loading old JSON format.  This is necessary since
  the `default_left` is changed from boolean to integer.
- 2.0: Emit warning when users save or load binary format.
- 2.1: Continue the warning about the binary format.
- 2.2: Continue the warning, save the model as JSON or UBJ as default if the format is not specified.
- 2.3: Remove support for saving old binary format, remove support for old JSON model.
- 2.4: Remove support for loading old binary format.

trivialfis:
@hcho3

hcho3:
> This is necessary since
the default_left is changed from boolean to integer.

How necessary is this? Was the default_left changed to improve performance?

trivialfis:
Yes.  Most of the improvement comes from the typed array where we can omit the construction of `Json` struct.  But there's no typed array for boolean.

Actually, there is, but it's useless.  The representation of boolean is `T` and `F` characters, they are both type and value at the same time.  So if we were to have a typed boolean array, the whole array would be either true or false.

[...]

PR #9947, opened 2024-01-03
Save model in ubj as the default.

trivialfis:
Related https://github.com/dmlc/xgboost/issues/7547 .

- Set default to UBJ in C.
- Set default to UBJ in JVM.
- Extract Python serialization tests.
- Remove changing `verbosity` in tests so that I can capture warnings.
- Cleanup tests to reduce the amount of warnings caused by the change of default.


trivialfis:
@wbo4958 Would be great if you could help take a look when you are available.

python-package/xgboost/testing/__init__.py
```diff
@@ -668,8 +668,7 @@ def weight_margin(draw: Callable) -> TestDataset:
 
 # A strategy for drawing from a set of example datasets. May add random weights to the
 # dataset
-@memory.cache
```
trivialfis, line 671: Removed since the composite strategy cannot be pickled. joblib has been issuing warnings for a while.

src/c_api/c_api.cc
```diff
@@ -1313,10 +1313,8 @@ XGB_DLL int XGBoosterLoadModel(BoosterHandle handle, const char* fname) {
 
 namespace {
 void WarnOldModel() {
-  if (XGBOOST_VER_MAJOR >= 2) {
```
trivialfis, line 1316: The version is lowered since we have released an extra 1.7 release. The original plan was to go from 1.6 to 2.0

wbo4958 (commented):
Still need to update the comments of https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/main/scala/org/apache/spark/ml/util/XGBoostReadWrite.scala#L34

wbo4958:
Otherwise, LGTM.

trivialfis:
Thank you for sharing!

trivialfis:
Still need to refresh the document, will open another PR once the R package is sorted out.

Issue #9944, opened 2024-01-02
Booster slicing doesn't take all valid increasing sequences


david-cortes:
When trying to slice a booster, it will not accept all formats that python would otherwise accept, and it's not clear what exactly it considers acceptable ranges - for example, if the step is not equal to one, it requires that the `end` be `step` more than the last returned round instead of just +1.

Example:
```python
import numpy as np, xgboost as xgb
rng = np.random.default_rng(seed=123)
y = rng.standard_normal(size=100)
X = rng.standard_normal(size=(100,10))
dm = xgb.DMatrix(X, y)
model = xgb.train(
    dtrain=dm,
    params={},
    num_boost_round=20
)
```

Fails:
```python
model[0:4:3]
```

Succeeds:
```python
model[0:6:3]
```

In contrast, python lists and numpy arrays would take both of those.

david-cortes:
Additionally:

* If the end of the slice (which is not included in the output) goes past the number of rounds in the model, regardless of what's the last number that the sequence would involve, it will throw an out-of-bounds error.

Here it has 20 rounds, but cannot slice like this:
```python
import numpy as np, xgboost as xgb
rng = np.random.default_rng(seed=123)
y = rng.standard_normal(size=100)
X = rng.standard_normal(size=(100,10))
dm = xgb.DMatrix(X, y)
model = xgb.train(
    dtrain=dm,
    params={},
    num_boost_round=20
)
model[0:21:3]
```

Note that this slice would stop at 18, which is within the range of the model's rounds:
```python
np.arange(20)[0:21:3]
```
```
array([ 0,  3,  6,  9, 12, 15, 18])
```

* When the underlying C function `XGBoosterSlice` returns code -2 (index out of bound), it will not set a message in `XGBGetLastError()`. The case above has an error message, but that error is thrown from the python wrapper, not from C++.

trivialfis:
The slicing is not as rich as Python or numpy. We at the time thought this might be a bit over-engineering for a booster class. It seems we should implement the full slicing just for less confusion. The easiest way to do it is just to convert a booster into a list of boosters and then let Python do the rest of the indexing. However, the conversion can be expensive.

david-cortes:
[...]

PR #9948, opened 2024-01-03
[R] Enable booster slicing

david-cortes:

To be rebased after PR https://github.com/dmlc/xgboost/pull/9924 is merged, as it contains the commits from it.

This PR adds a wrapper over the C function to slice booster rounds.

The slicing logic is made in such a way that it maches with R's syntax for sequences, but there were some issues from the C-level function as described in previous issue: https://github.com/dmlc/xgboost/issues/9944

so the tests aren't very comprehensive


trivialfis:
apologies for the difficulty of stacking PRs, I will try to look into https://github.com/dmlc/xgboost/pull/9924 as soon as possible.

R-package/NAMESPACE
```diff
@@ -62,6 +64,7 @@ export(xgb.plot.tree)
 export(xgb.save)
 export(xgb.save.raw)
 export(xgb.set.config)
+export(xgb.slice.Booster)
```
trivialfis, line 67: Is this still needed after having the s3 method?

david-cortes: It's not needed, but I think it helps with discoverability compared to having just an S3 method.

trivialfis: Thank you for sharing, sounds good!

trivialfis (approved):
Overall looks good to me! Can merge once rebase is done.

david-cortes:
Sorry - I'm not sure what happened here. Looks like the PR somehow closed itself after force-pushing, and shows it has being merged already, which it isn't. Looks like I cannot reopen it from here either.

PR #9949, opened 2024-01-03
[R] Ensure translation from array_interface uses type of sufficient width

david-cortes:

As a follow up from https://github.com/dmlc/xgboost/pull/9929

This PR makes a check before copying data into an R array of 'int' type (32-bit, signed) that the data being copied into it can fit into the corresponding data type, and chooses a `float64` (R's 'real') otherwise, which has a wider range.


david-cortes:
No idea why the C++ tests are failing here, since this PR only touches R-specific files which aren't tested there.

trivialfis:
I don't think it's possible to map integers to floating points exactly, we can throw an error if there's overflow.

david-cortes:
> I don't think it's possible to map integers to floating points exactly, we can throw an error if there's overflow.

`float64` has integral precision up to 2^53.

trivialfis (approved):

PR #9957, opened 2024-01-07
[R] Refactor callback structure and attributes

david-cortes:

To be merged after the following previous PRs which are dependencies here:
* https://github.com/dmlc/xgboost/pull/9924
* https://github.com/dmlc/xgboost/pull/9935
* https://github.com/dmlc/xgboost/pull/9948

This PR refactors the structure and logic of the callbacks in order to match those of the python interface, which allows more flexibility and feels a lot easier to manipulate.

After this PR, callbacks have their own environment which they can manipulate without affecting the environment of the training function call, and can execute functions at different stages of the process (before/after training and before/after round), with a signature for those functions which is now standardized and documented, and an example added there for creating custom user callbacks.

After this, the logic for the attributes of the booster object also changes, as now callbacks can leave arbitrary attributes in the booster through their finalizers which will be kept with the same name as the callback, and the names of some existing attributes are now moved to a sub-list within the name of the callback that produces them - e.g. before, the CV prediction left an entry `pred`, which now will be under `cv_predict$pred` instead.

Most of the code here is straight copy-paste from the previous callbacks, with some minor modifications:
* Possibility to slice the booster up to the best round when using early stopping.
* Changing the default format for saving model snapshots to JSON.
* Minor efficiency improvements like pre-allocating arrays for the expected size of the artifacts.


R-package/R/callbacks.R
```diff
@@ -1,563 +1,923 @@
...
+#' Note that, for \link{xgb.cv}, this will be the full data, while data for the specific
+#' folds can be found in the `model` object.
+#'
+#' \item watchlist The evaluation watchlist, as passed under argument `watchlist` to
+#' \link{xgb.train}.
+#'
+#' For \link{xgb.cv}, this will always be `NULL`.
+#'
+#' \item begin_iteration Index of the first boosting iteration that will be executed
+#' (base-1 indexing).
```
trivialfis, line 81: I think it should always be 1 and `num_boost_rounds` should always be the number of *new* trees to add. This helps XGBoost to provide additional checks during the first iteration with new data, and fewer conditions are easier to understand.

R-package/R/callbacks.R
```diff
@@ -1,563 +1,923 @@
...
-#' \code{bst_evaluation},
-#' \code{iteration}.
+#' This will typically be '1', but when using training continuation, depending on the
+#' parameters for updates, boosting rounds will be continued from where the previous
+#' model ended, in which case this will be larger than 1.
 #'
-#' @seealso
-#' \code{\link{callbacks}}
+#' \item end_iteration Index of the last boostign iteration that will be executed
+#' (base-1 indexing, inclusive of this end).
```
trivialfis, line 88: Is this useful somehow other than pre-allocating a vector to store the evaluation history? These types of vectors/lists are quite cheap that don't need pre-allocation, and users who write callback usually know exactly the number of boosting rounds (they also write the training code).  I think it's not necessary to have an additional parameter with a condition that we (and the user) need to check whether it matches the number of boosting rounds.

R-package/R/callbacks.R
```diff
@@ -1,563 +1,923 @@
...
 #'
+#' It should match with argument `nrounds` passed to \link{xgb.train} or \link{xgb.cv}.
+#'
+#' Note that boosting might be interrupted before reaching this last iteration, for
+#' example by using the early stopping callback \link{xgb.cb.early.stop}.
+#'
+#' \item iteration Index of the iteration number that is being executed (first iteration
```
trivialfis, line 95: Currently, the Python interface calls it epoch, can we keep it consistent here?

R-package/R/callbacks.R
```diff
@@ -1,563 +1,923 @@
...
+#' will be the same as parameter `begin_iteration`, then next one will add +1, and so on).
+#'
+#' \item iter_feval Evaluation metrics for the `watchlist` that was supplied, either
+#' determined by the objective, or by parameter `feval`.
+#'
+#' For \link{xgb.train}, this will be a named vector with one entry per element in
+#' `watchlist`, where the names are determined as 'watchlist name' + '-' + 'metric name' - for
+#' example, if `watchlist` contains an entry named "tr" and the metric is "rmse",
+#' this will be a one-element vector with name "tr-rmse".
```
trivialfis, line 104: Can we pass the evaluation result as a proxy instead of the actual data to callbacks?

R-package/R/callbacks.R
```diff
@@ -1,563 +1,923 @@
...
+#' \item call
+#' \item params
+#' \item niter
+#' \item nfeatures
+#' \item folds
+#' }
+#'
+#' When using the built-in early stopping callback (\link{xgb.cb.early.stop}), said callback
+#' will always be executed before the others, as it sets some booster C-level attributes
+#' that other callbacks might also use. Otherwise, the order of execution will match with
```
trivialfis, line 153: Maybe we need an internal callback that does this and split it up with the early stopping? Along with the comment for evaluation history instead of actual data, see `CallbackContainer` in Python.

trivialfis (commented):
Thank you for working on the callback function interface! It's one of the most important parts of language bindings.

After briefly looking into the code, I'm more interested in the interface than the implementation details. Looking at the documents, I think some of the options and parameters are not necessary, it would be great if we could try to make the interface surface as small as possible. For instance, if some of the parameters are passed by the user in the first place, then maybe they can use closure to capture those parameters in callbacks.
In addition, some more design choices need to be made, for instance, less difference between CV and normal training would be great. XGBoost has a historically complex codebase and interface, it would be great if we could reduce the amount of conditional behaviors like "if x is used in y, then it's different than the behavior when using x in z"

R-package/R/callbacks.R
david-cortes: Yes, it is useful if e.g. one wants to:
* Create a custom step size calculator or otherwise modify some other parameter(s) accordingly.
* Target a certain number of snapshots during training instead of every N rounds or so.
* Create re-usable callbacks for xgboost in other packages.
* Create a different type of booster with an external optimizer like [this guy did here](https://github.com/catboost/catboost/issues/1994#issuecomment-1898816616).

Among others.

It's also useful when it comes to creating custom code to combine results of previous a callback with new results after an update (e.g. it would help to determine whether to append or replace).

R-package/R/callbacks.R
david-cortes: It is passed. But the user might _also_ want to use the actual data in a custom callback.

R-package/R/callbacks.R
david-cortes: But if it's implemented as a callback, it'd still need to be executed before other callbacks.

R-package/R/callbacks.R
david-cortes: There's also a functionality that updates trees without adding new ones.

david-cortes:
> less difference between CV and normal training would be great

That's part of what this PR modifies: now in both cases, callbacks add attributes under their own (the callback's) name, and these attributes match between the two, with exceptions like the prediction-saving which is only for CV, or the coefficients history which saves one result per booster.

Or do you have some other type of changes in mind that could be adopted here?

> it would be great if we could try to make the interface surface as small as possible

Yes, it feels like some things aren't really helpful, but for the most part, the functionalities here are also present in the python interface (e.g. python also has a `LearningRateScheduler`), so I didn't remove anything when switching to this new structure but perhaps removals could be done in subsequent PRs.

> if some of the parameters are passed by the user in the first place

Most of these callbacks are added automatically based on arguments to `xgb.train` or `xgb.cv`. The only ones that are required to be passed by the user are the coefficients history and the 'reset parameters' callbacks.

david-cortes:
> For instance, if some of the parameters are passed by the user in the first place, then maybe they can use closure to capture those parameters in callbacks

One remark here: the user _might_ know these parameters from passing them to `xgb.train`, but if someone wants to develop a reusable custom callback that could be simply added to the list the same way as xgboost's (e.g. to offer it as a package), then it'd be helpful to have things that might seem redundant, such as the evaluation data or the labels in the DMatrix.

R-package/R/callbacks.R
trivialfis: That makes sense.

R-package/R/callbacks.R
trivialfis: they can get it from the model itself. let's keep this part consistent across language bindings.

trivialfis:
> then it'd be helpful to have things that might seem redundant, such as the evaluation data or the labels in the DMatrix.

That's a valid use case, but I think callbacks that need access to these data, especially data that don't change throughout the training session, can make them parameters of the constructor instead of the hook methods.

R-package/R/callbacks.R
david-cortes: But isn't that how it's currently done? Only difference is that it's part of the public interface so the user can choose to pass it manually.

R-package/R/callbacks.R
david-cortes: The python interface also passes it, as a parameter named `epoch`.

For example, this is the current python signature for the equivalent function to run before each round:
```python
def before_iteration(
        self,
        model: _Model,
        epoch: int,
        dtrain: DMatrix,
        evals: Optional[List[Tuple[DMatrix, str]]],
    ) -> bool
```

david-cortes:
> > then it'd be helpful to have things that might seem redundant, such as the evaluation data or the labels in the DMatrix.
> 
> That's a valid use case, but I think callbacks that need access to these data, especially data that don't change throughout the training session, can make them parameters of the constructor instead of the hook methods.

Then there's cases like external-memory DMatrices, higher-level interfaces that might apply some processing on their own (e.g. encoding of categorical labels), etc.

R-package/R/callbacks.R
```diff
@@ -1,563 +1,923 @@
...
 #'
-#' When a callback function has \code{finalize} parameter, its finalizer part will also be run after
-#' the boosting is completed.
+#' If passing `NULL` for this or for the other function inputs, then no function will be executed.
 #'
-#' WARNING: side-effects!!! Be aware that these callback functions access and modify things in
-#' the environment from which they are called from, which is a fairly uncommon thing to do in R.
+#' If passing a function, it will be called with parameters supplied as non-named arguments
+#' matching the function signatures that are shown in the default value for each function argument.
+#' @param f_before_iter A function that will be executed before each boosting round.
```
trivialfis, line 22: Does it return a boolean value?

R-package/R/callbacks.R
trivialfis: > This will typically be '1', but when using training continuation, depending on the
#' parameters for updates

It's passed in Python, but I think it should start from 0/1 instead of the previous boosted number.

R-package/R/callbacks.R
david-cortes: No, this function `f_before_iter` doesn't need to return anything if the user defines it.

The idea is that there can be cases in which some user function might need to execute before instead of after, or there might be actions that happen in both.

For example:
* A user callback that modifies things like the step sizes would likely fit better under `f_before_iter`.
* A user callback that uses some EM-type algorithm would likely fit better under `f_after_iter`.

R-package/R/callbacks.R
david-cortes: But imagine again a case in which a user is setting up a custom step size sequence: if this were to be reset to zero on training continuation, then the user callback would give out different results when training in one go than when using continuation.

R-package/R/callbacks.R
trivialfis: I'm mostly comparing the callback between R and Python, with the hope that we can minimize the difference when approapriate, the one in Python does return a boolean value for the same reason as the `f_after_iter`.

R-package/R/callbacks.R
trivialfis: That's a good point, we might want to break the compatibility in other language bindings for this change.

R-package/R/callbacks.R
trivialfis: We will do it your way.

My concern is that, often time the real data is difficult to protect, and might not be fully available. Should we pass the DMatrix or the original data, before or after processing like converting to numeric? Also, what happens with distributed training?

That said, I think your use case makes sense and will proceed as it's.

R-package/R/callbacks.R
david-cortes: Modified so that it would interpret the return value as a signal of whether to stop the training or not.

R-package/R/callbacks.R
david-cortes: I don't think those should be concerns, because the user can call `getinfo` on the DMatrix to obtain the data in a standardized format and use that in the callbacks.

As far as the function interface is aware, any input data other than `y` already needs to be passed as DMatrix for these functions, and that's what functions are meant to work with.

trivialfis (approved):

Issue #9956, opened 2024-01-05
Use UBJ in Python checkpoint.


trivialfis:
https://github.com/dmlc/xgboost/blob/38dd91f49137305e5822f459ac338a1497f2e400/python-package/xgboost/callback.py#L595
https://github.com/dmlc/xgboost/blob/38dd91f49137305e5822f459ac338a1497f2e400/python-package/xgboost/core.py#L2622

PR #9958, opened 2024-01-08
Use UBJ in Python checkpoint.

trivialfis:
- Update document.

Close https://github.com/dmlc/xgboost/issues/9956 .

### DIFF
--- a/doc/python/python_api.rst
+++ b/doc/python/python_api.rst
@@ -88,22 +88,18 @@ Callback API
 
 .. autoclass:: xgboost.callback.EvaluationMonitor
     :members:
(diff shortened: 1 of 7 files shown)


doc/python/python_api.rst
```diff
@@ -88,22 +88,18 @@ Callback API
 
 .. autoclass:: xgboost.callback.EvaluationMonitor
     :members:
-    :inherited-members:
```
trivialfis, line 91: We don't want documents from the abstract class in concrete callback classes.

RAMitchell (approved):

Issue #9960, opened 2024-01-08
Mistake in vector leaf train


oOTWK:
There is a logic mistake in the vector leaf training at `ApplyTreeSplit` https://github.com/dmlc/xgboost/blob/master/src/tree/hist/evaluate_splits.h#L705.

The error is in the gain calculation. The current implementation looks like the following.

```
CalcWeight(*param_, left_sum, param_->learning_rate, left_weight);
CalcWeight(*param_, right_sum, param_->learning_rate, right_weight);
gain_[left_child] = CalcGainGivenWeight(*param_, left_sum, left_weight);
gain_[right_child] = CalcGainGivenWeight(*param_, right_sum, right_weight);
```

Weight is calculated as base_weight * learning_rate and that weight is used to calculate the gain. The learning rate should be excluded from the gain calculation.

PR #9978, opened 2024-01-10
Fix gain calculation in multi-target tree.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/9960 .

### DIFF
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -149,6 +149,9 @@ class MultiTargetHistBuilder {
   }
 
   void InitData(DMatrix *p_fmat, RegTree const *p_tree) {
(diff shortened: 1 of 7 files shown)


trivialfis:
@rongou I have disabled the test for now. We need to refactor the evaluation to serialize the entries before synchronization.

src/tree/updater_quantile_hist.cc
```diff
@@ -149,6 +149,9 @@ class MultiTargetHistBuilder {
   }
 
   void InitData(DMatrix *p_fmat, RegTree const *p_tree) {
+    if (collective::IsDistributed()) {
```
trivialfis, line 152: @hcho3 Added a check to disable the distributed training.

hcho3 (approved):

PR #9980, opened 2024-01-10
fixes #9972 FieldEntry ctor specialisation syntax error

### LINKED ISSUE #9972: FieldEntry ctor specialisation syntax error in macro DECLARE_FIELD_ENUM_CLASS
The syntax for declaring the specialisation of `FieldEntry` in macro `DECLARE_FIELD_ENUM_CLASS` is valid with GCC in C++17 standards compliance mode, but not C++20.
To be more precise, GCC up to 10.5 accepts the syntax in C++20 mode; from 11 onwards, this causes an error.
https://godbolt.org/z/K8fGc98jq
The fix is to simply replace in the macro the ctor declaration from `FieldEntry<EnumClass>() {` to `FieldEntry() {` which compiles in both 17 and 20 mode.

rpopescu:
fixes #9972 FieldEntry ctor specialisation syntax error in macro DECLARE_FIELD_ENUM_CLASS

### DIFF
--- a/include/xgboost/parameter.h
+++ b/include/xgboost/parameter.h
@@ -53,7 +53,7 @@ namespace parameter {  \
 template <>  \
 class FieldEntry<EnumClass> : public FieldEntry<int> {  \
  public:  \
-  FieldEntry<EnumClass>() {  \
+  FieldEntry() {  \
     static_assert(  \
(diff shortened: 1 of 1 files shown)


trivialfis (approved):
Thank you for the fix!

Issue #9856, opened 2023-12-06
[CI] Flaky timeout test.


trivialfis:
https://github.com/dmlc/xgboost/actions/runs/7119127358/job/19383556299
https://github.com/dmlc/xgboost/actions/runs/7386261666/job/20092505010
https://github.com/dmlc/xgboost/actions/runs/7401120631/job/20136169837?pr=9949
https://github.com/dmlc/xgboost/actions/runs/7528975440/job/20492275487

PR #9990, opened 2024-01-15
Fix error handling in the event loop.

trivialfis:
Make sure the thread worker cannot exit without the `stop_` flag being set.

Close https://github.com/dmlc/xgboost/issues/9856 .


trivialfis:
cc @rongou

rongou (approved):

PR #9993, opened 2024-01-16
Add (failing) tests for RABIT.

trivialfis:


trivialfis:
Will open again after the refactor.

PR #10011, opened 2024-01-26
SYCL. Add some basic functional for sycl implementation of partition builder.

razdoburdin:
Hi,
I continue adding of `sycl `support (https://github.com/dmlc/xgboost/pull/9691, https://github.com/dmlc/xgboost/pull/9846, https://github.com/dmlc/xgboost/pull/9800).
By this PR I am starting adding the training in the `sycl `devices. All training is to large for a single PR. For that reason I plan to add it step by step together with the related tests. Here I add some basic logic related to `PartitionBuilder`. 

I have also made some changes in sycl-related CMakeFile to build sycl tests with icpx.

### DIFF
--- a/plugin/sycl/common/partition_builder.h
+++ b/plugin/sycl/common/partition_builder.h
@@ -31,10 +31,6 @@ namespace common {
 // The builder is required for samples partition to left and rights children for set of nodes
 class PartitionBuilder {
  public:
-  static constexpr size_t maxLocalSums = 256;
-  static constexpr size_t subgroupSize = 16;
-
-
   template<typename Func>
   void Init(::sycl::queue* qu, size_t n_nodes, Func funcNTaks) {
     qu_ = qu;
@@ -53,11 +49,6 @@ class PartitionBuilder {
     }
   }
 
-  size_t GetSubgroupSize() {
-    return subgroupSize;
-  }
-
-
   size_t GetNLeftElems(int nid) const {
     return result_rows_[2 * nid];
   }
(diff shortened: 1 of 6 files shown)


plugin/sycl/common/partition_builder.h
```diff
@@ -0,0 +1,101 @@
+/*!
+ * Copyright 2017-2023 XGBoost contributors
```
napetrov, line 2: should it be 2024? Or just creation year

tests/cpp/CMakeLists.txt
```diff
@@ -14,8 +14,38 @@ if(USE_CUDA)
...
+    ${xgboost_SOURCE_DIR}/include
+    ${xgboost_SOURCE_DIR}/dmlc-core/include
+    ${xgboost_SOURCE_DIR}/rabit/include)
+
+  target_compile_definitions(plugin_sycl_test PUBLIC -DXGBOOST_USE_SYCL=1)
+
+  target_link_libraries(plugin_sycl_test PUBLIC -fsycl)
+
+  set_target_properties(plugin_sycl_test PROPERTIES
+  COMPILE_FLAGS -fsycl
```
napetrov, line 36: should we also have "-fsycl-device-code-split=per_kernel" as options?

plugin/sycl/common/partition_builder.h
razdoburdin: fixed

tests/cpp/CMakeLists.txt
razdoburdin: > should we also have "-fsycl-device-code-split=per_kernel" as options?

I didn't observe any perf differences, at least for the last compiler.

trivialfis (approved):

PR #10016, opened 2024-01-30
[R] Rename ProxyDMatrix -> DataBatch

david-cortes:

As a follow up from https://github.com/dmlc/xgboost/pull/9913

I had second thoughts on the naming using internally and externally.

Currently, the `DataIter` callback needs to return a `ProxyDMatrix`, and internally, it uses a `ProxyDMatrixHandle`, which is quite unintuitive.

This PR renames them so that the object that needs to be passed to the iterator is now `DataBatch`, and the internal object (which is not exposed to the user) is named `ProxyDMatrix`.

Along the way, it fixes some tests to not accept function arguments that were removed during internal changes in the previous PR.


trivialfis (approved):

trivialfis:
The change makes sense. Please help rebase the PR.

david-cortes:
> The change makes sense. Please help rebase the PR.

This is very odd. I don't get any merge conflict when merging the latest master.

PR #10018, opened 2024-01-30
[R] Remove `enable_categorical` as parameter

david-cortes:

The Python DMatrix class introduced a parameter `enable_categorical` which needs to be manually changed from its default value in order to auto-detect categorical features in data frames.

From earlier comments, I understand the idea behind this parameter was to ensure backwards compatibility with processing of data frames in previous versions of xgboost for python, which used to be accepted but with all columns taken as numerical.

In R, there hasn't been any release of xgboost where data frames would have been accepted by DMatrix constructors, so there is no historical compatibility issue that would call for having this parameter.

This PR removes it, making categorical column detection always trigger for data frames. Note that the user can still override it by manually passing `feature_types`, just like in the python interface.


trivialfis (approved):
looks good!

PR #10019, opened 2024-01-30
[R] Document handling of indexes

david-cortes:

This PR adds some notes about handling of index numbers in R, as requested in [previous comment](https://github.com/dmlc/xgboost/pull/9935#issuecomment-1891670106).


doc/contrib/coding_guide.rst
```diff
@@ -224,3 +224,27 @@ masked array at the moment (no specialized adapter).  One should be careful abou
...
+
+XGBoost, being written in C++, uses base-0 indexing and considers ranges / sequences to be inclusive of the left end but not the right one - for example, a range (0, 3) would include the first three elements, numbered 0, 1, and 2.
+
+The Python interface uses this same logic, since this is also the way that indexing in Python works, but other languages like R have different logic. In R, indexing is base-1 and ranges / sequences are inclusive of both ends - for example, to refer to the first three elements in a sequence, the interval would be written as (1, 3), and the elements numbered 1, 2, and 3.
+
+In order to provide a more idiomatic R interface, XGBoost adjusts adjusts its user-facing R interface to follow this and similar R conventions, but internally, it needs to convert all these numbers to the format that the C interface uses. This is made more problematic by the fact that models are meant to be serializable and loadable in other interfaces, which will have different indexing logic.
```
trivialfis, line 238:
```suggestion
In order to provide a more idiomatic R interface, XGBoost adjusts its user-facing R interface to follow this and similar R conventions, but internally, it needs to convert all these numbers to the format that the C interface uses. This is made more problematic by the fact that models are meant to be serializable and loadable in other interfaces, which will have different indexing logic.
```

doc/contrib/coding_guide.rst
```diff
@@ -224,3 +224,27 @@ masked array at the moment (no specialized adapter).  One should be careful abou
...
+- Categorical features are defined in R as a ``factor`` type which encodes with base-1 indexing. When categorical features are passed as R ``factor`` types, the conversion is done automatically to base-0 indexing, but if the user whishes to manually supply categorical features as already-encoded integers, then those integers need to already be in base-0 encoding.
+- Categorical labels for DMatrices do not undergo any extra processing - the user must supply base-0 encoded labels.
+- A function to retrieve class-specific coefficients when using the linear coefficients history callback takes a class index parameter, which also does not undergo any conversion (i.e. user must pass a base-0 index), in order to match with the label logic - that is, the same class index will refer to the class encoded with that number in the DMatrix ``label`` field.
+
```
trivialfis, line 249: How about evaluation output:
```
[0]: rmse ...
[1]: rmse ...
```

doc/contrib/coding_guide.rst
```diff
@@ -224,3 +224,27 @@ masked array at the moment (no specialized adapter).  One should be careful abou
...
+New additions to the R interface that take on indexable elements should be mindful of these conventions and try to mimic R's behavior as much as possible.
```
trivialfis, line 250: How about the feature index from the tree plot and JSON model? I think for the JSON model, the output will continue to be 0-based, not sure about what to do with the plot. In addition, the categorical features in tree dump and plot.

trivialfis (commented):
Thank you for writing the document, I think this would be useful to users as well instead of being contributors only.

doc/contrib/coding_guide.rst
david-cortes: Added, alongside with other cases that were missing.

doc/contrib/coding_guide.rst
david-cortes: Added.

david-cortes:
> Thank you for writing the document, I think this would be useful to users as well instead of being contributors only.

Are you suggesting to move this to some other file?

trivialfis:
Yes, perhaps somewhere as part of the R document like a vignette since it's quite R-specific; otherwise, creating a new file like `doc/index-base. rst` also works.

david-cortes:
> Yes, perhaps somewhere as part of the R document like a vignette since it's quite R-specific; otherwise, creating a new file like `doc/index-base. rst` also works.

Moved it to a new file under `R-package`, but I'm not sure that I specified the RST format correctly.

Also added a few notes about base-0 encodings in the DMatrix docs.

trivialfis (approved):

david-cortes:
@trivialfis @hcho3 Looks like the docs didn't render correctly, and didn't include this page in the index. I am not familiar with RST format. Could you look into it?
image
image

trivialfis:
I will look into it tomorrow night.

hcho3:
@trivialfis I am looking at this now.

hcho3:
@david-cortes https://github.com/dmlc/xgboost/pull/10030 should fix the doc.

Issue #10027, opened 2024-02-01
Global configuration could manage global number of threads


david-cortes:
XGBoost has a global configuration state:
https://xgboost.readthedocs.io/en/stable/parameter.html#global-configuration

Some places in the code try to retrieve a global number of threads, such as here:
https://github.com/dmlc/xgboost/blob/4dfbe2a8935ca45321e906ab8ee62b5d592d1fe8/src/gbm/gbtree_model.cc#L112

The default number of threads is determined from OpenMP, which is a reasonable thing to do:
https://github.com/dmlc/xgboost/blob/4dfbe2a8935ca45321e906ab8ee62b5d592d1fe8/src/common/threading_utils.cc#L100

But being a global configuration, would be more logical to allow users to modify it through the functions that are named as such.

It would make things better for the R interface, since currently there are workarounds like these which result in degraded performance for users:
https://github.com/dmlc/xgboost/blob/4dfbe2a8935ca45321e906ab8ee62b5d592d1fe8/src/gbm/gbtree_model.cc#L120

If it were a global option, should be possible to adopt a mechanism like LightGBM's which changes the number of configured threads before tests and examples, but doesn't have the lines that change the threads in the user-visible part of the examples, so that it executes with reduced threads on CRAN but doesn't suggest doing so to the user.

Example:
https://github.com/microsoft/LightGBM/blob/252828fd86627d7405021c3377534d6a8239dd69/R-package/R/lgb.Dataset.R#L783

trivialfis:
After working on distributed systems, I'm a bit afraid of this global variable, it's global and it's thread local. This results in the fact that it needs special handling for distributed systems, and it fails without possible fix when user launches a new thread without setting the value again.

david-cortes:
> After working on distributed systems, I'm a bit afraid of this global variable, it's global and it's thread local. This results in the fact that it needs special handling for distributed systems, and it fails without possible fix when user launches a new thread without setting the value again.

But a failure here would be no issue, since then it could just default back to OMP's config.

The idea would be to use this as a means to override the OMP config, so that it would check for this first, and then if it's not set, would look at the OMP config.

trivialfis:
> since then it could just default back to OMP's config.

Depending on your perspective, this change of behavior can be considered a bug. In general, global variable are not very desirable.

> The idea would be to use this as a means to override the OMP config

[...]

PR #10028, opened 2024-02-02
[R] Don't cap global number of threads

david-cortes:
ref https://github.com/dmlc/xgboost/issues/10027

This PR removes manual capping of number of threads in XGBoost functions that take it from a global config instead of being supplied as argument.

CRAN has a requirement of not using more than 2 threads in tests and examples, so it substitutes this workaround by manual setting of OMP number of threads config, which is done by adding a dependency `RhpcBLASctl` which does provides such functionality, and adds silent calls to it before the tests that might execute something with the global number of OMP threads, the same way it was done in LightGBM: https://github.com/microsoft/LightGBM/pull/6226

Note: I am not entirely sure that I added silent calls to this function that sets the OMP threads in all of the examples where it's needed.

### DIFF
--- a/R-package/R/xgb.config.R
+++ b/R-package/R/xgb.config.R
@@ -4,7 +4,17 @@
 #' values of one or more global-scope parameters. Use \code{xgb.get.config} to fetch the current
 #' values of all global-scope parameters (listed in
 #' \url{https://xgboost.readthedocs.io/en/stable/parameter.html}).
+#' @details
+#' Note that some functions might use a globally-configured number of threads, which is not
+#' managed by this configuration. Typically, XGBoost methods accept an `nthreads` parameter
+#' controlling the number of parallel  threads that will be used, but some functionalities
+#' might execute functions before this parameter can be supplied.
 #'
+#' In such cases, the global number of threads is taken from the configured number of OpenMP
+#' (OMP) threads, which if not set, will default to the maximum number of available threads.
+#' The number of OMP threads in turn can be configured through an environment variable
+#' `OMP_NUM_THREADS` (which needs to be set before R is started), or alternatively, can be
+#' changed in an R session through package `RhpcBLASctl`, by calling function `RhpcBLASctl::omp_set_num_threads`.
 #' @rdname xgbConfig
 #' @title Set and get global configuration
 #' @name xgb.set.config, xgb.get.config
--- a/R-package/demo/basic_walkthrough.R
+++ b/R-package/demo/basic_walkthrough.R
@@ -55,6 +55,8 @@ print(paste("test-error=", err))
 # save model to binary local file
 xgb.save(bst, "xgboost.model")
 # load binary model to R
+# Function doesn't take 'nthreads', but can be set like this:
+RhpcBLASctl::omp_set_num_threads(1)
 bst2 <- xgb.load("xgboost.model")
 pred2 <- predict(bst2, test$data)
 # pred2 should be identical to pred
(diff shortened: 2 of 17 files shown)


david-cortes:
Regarding the failing CI job:
```
Quitting from lines 498-508 [loadModel] (xgboostPresentation.Rmd)
Error: processing vignette 'xgboostPresentation.Rmd' failed with diagnostics:
there is no package called 'RhpcBLASctl'
```

Looks like the vignette builder is not installing all the packages mentioned under `Suggests`.

hcho3 (approved):

trivialfis:
> Looks like the vignette builder is not installing all the packages mentioned under Suggests.

Please help update the CI script: `/R-package/tests/helper_scripts/install_deps.R`.

R-package/R/xgb.config.R
```diff
@@ -4,7 +4,17 @@
 #' values of one or more global-scope parameters. Use \code{xgb.get.config} to fetch the current
 #' values of all global-scope parameters (listed in
 #' \url{https://xgboost.readthedocs.io/en/stable/parameter.html}).
+#' @details
+#' Note that some functions might use a globally-configured number of threads, which is not
+#' managed by this configuration. Typically, XGBoost methods accept an `nthreads` parameter
+#' controlling the number of parallel  threads that will be used, but some functionalities
+#' might execute functions before this parameter can be supplied.
```
trivialfis, line 11: The only feature that has problems with local `nthread` is the model serialization, let's be more specific here to reduce confusion.

R-package/R/xgb.config.R
```diff
@@ -4,7 +4,17 @@
...
 #'
+#' In such cases, the global number of threads is taken from the configured number of OpenMP
+#' (OMP) threads, which if not set, will default to the maximum number of available threads.
+#' The number of OMP threads in turn can be configured through an environment variable
+#' `OMP_NUM_THREADS` (which needs to be set before R is started), or alternatively, can be
+#' changed in an R session through package `RhpcBLASctl`, by calling function `RhpcBLASctl::omp_set_num_threads`.
```
trivialfis, line 17: I think a simple note that in addition to the `nthread` parameter, XGBoost honors global configuration set by `OMP_NUM_THREADS` or `RhpcBLASctl` should suffice.

R-package/demo/basic_walkthrough.R
```diff
@@ -55,6 +55,8 @@ print(paste("test-error=", err))
 # save model to binary local file
 xgb.save(bst, "xgboost.model")
 # load binary model to R
+# Function doesn't take 'nthreads', but can be set like this:
+RhpcBLASctl::omp_set_num_threads(1)
```
trivialfis, line 59: Please note that this also modifies all the thread numbers for the rest of the demo, including prediction.

trivialfis (commented):
Overall, it looks good; I like the small utility library. Please see the comments around the documentation.

R-package/demo/basic_walkthrough.R
david-cortes: That shouldn't be a problem, since the rest of the demo also needs to follow the same thread limitations.

R-package/R/xgb.config.R
david-cortes: Updated.

R-package/R/xgb.config.R
david-cortes: Simplified.

david-cortes:
R tests are now passing.

trivialfis (approved):

PR #10029, opened 2024-02-02
[R] Use same RNG algorithm as in other interfaces

david-cortes:

This PR removes the custom RNG option for the R package that uses R's own RNG in place of XGBoost's default RNG.

In order to still mimic the old behavior, it will draw a random seed through R's RNG system when the `seed` parameter is not manually specified, but otherwise will take the random seed (whether manually passed or automatically drawn) with the default random engine.

With this change, running the same call to `xgb.train` with the same data and parameters should produce the same results across interfaces.

*******************

By the way, I see that XGBoost is using a Mersenne-Twister generator. This family of RNG algorithms has been found to [not hold all the statistical properties that one expects from an RNG algorithm](https://arxiv.org/abs/1910.06437), and many large projects (e.g. NumPy, DuckDB, Highway, among many others) have consequently moved towards newer RNGs like [Xoshiro family](https://prng.di.unimi.it) and [PCG family](https://www.pcg-random.org/index.html).

It might not matter much given the limited amount of random numbers that are drawn in XGBoost, but could be a potential improvement to look after.


trivialfis:
@hcho3 is the CI working correctly? I can't see the dashboard for some reason.

hcho3:
The Lambda function is broken for some reason

hcho3:
CI is now back to health. Fix: https://github.com/hcho3/xgboost-devops/commit/9e789e519381bad2de8f13bd0fa98b3d3a3e6e51

trivialfis:
Excellent! Thank you for the fix!

trivialfis (approved):

PR #10031, opened 2024-02-04
[R] Make `xgb.cv` work with `xgb.DMatrix` only, adding support for survival and ranking fields

david-cortes:
fixes https://github.com/dmlc/xgboost/issues/7118
closes https://github.com/dmlc/xgboost/pull/9863

This PR, which includes chunks from the [previous one](https://github.com/dmlc/xgboost/pull/9863) by @mayer79 , modifies `xgb.cv` to work only with `xgb.DMatrix` inputs, and switches the `stratified` parameter to have different defaults depending on whether the task is about classification or regression.

With this change, it is now possible to run `xgb.cv` for survival and ranking objectives. In the case of ranking objectives, split is done based on the groups, the same way as it is done in python. Unlike the python interface however, it always defaults to a group-based split for such cases, regardless of whether the user passes `stratified` or not.

I realized while implementing these changes that the DMatrix slicing method had not incorporated the new quick field-checking attibute that's now part of DMatrices - this is added here too.


hcho3:
Let me review tomorrow

david-cortes:
@trivialfis PR is ready for review.

R-package/R/utils.R
```diff
@@ -235,33 +235,34 @@ convert.labels <- function(labels, objective_name) {
 }
 
 # Generates random (stratified if needed) CV folds
-generate.cv.folds <- function(nfold, nrows, stratified, label, params) {
+generate.cv.folds <- function(nfold, nrows, stratified, label, group, params) {
+  if (NROW(group)) {
+    return(generate.group.folds(nfold, group))
```
hcho3, line 240: It appears that we don't support stratified CV when `group` is non-empty. Can we throw a warning about the use of un-stratified splitting when `group` is non-empty?

R-package/R/utils.R
```diff
@@ -235,33 +235,34 @@ convert.labels <- function(labels, objective_name) {
...
+  }
+  objective <- params$objective
+  if (!is.character(objective)) {
+    stratified <- FALSE
```
hcho3, line 244:
```suggestion
    warning("Will use unstratified splitting (custom objective used)")
    stratified <- FALSE
```

hcho3 (changes requested):
Apologies for the delay. The pull request looks good to me.

R-package/R/utils.R
david-cortes: Changed it to throw an error rather than a warning.

david-cortes: Changed it to a warning now.

hcho3 (approved):

PR #10033, opened 2024-02-06
update python intro doc

UncleLLD:
update  some spelling errors in  `python_intro.rst`


trivialfis (commented):
Thank you for the fix, please see inline comment.

PR #10058, opened 2024-02-21
update python docs some mistakes

UncleLLD:
* `doc/python/callbacks.rst`，the original code reported an error as follows:
image

after change `Valid` into `Train`, it is ok

* `doc/python/sklearn_estimator.rst`, the original code reported an error as follows:
image

after change `Validation_0` into `validation_0`, it is ok

### DIFF
--- a/doc/python/callbacks.rst
+++ b/doc/python/callbacks.rst
@@ -36,7 +36,7 @@ inside iteration loop.  You can also pass this callback function directly into X
     # Specify which dataset and which metric should be used for early stopping.
     early_stop = xgb.callback.EarlyStopping(rounds=early_stopping_rounds,
                                             metric_name='CustomErr',
-                                            data_name='Train')
+                                            data_name='Valid')
 
     booster = xgb.train(
         {'objective': 'binary:logistic',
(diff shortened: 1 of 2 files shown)


doc/python/callbacks.rst
```diff
@@ -49,7 +49,7 @@ inside iteration loop.  You can also pass this callback function directly into X
         verbose_eval=False)
 
     dump = booster.get_dump(dump_format='json')
-    assert len(early_stop.stopping_history['Valid']['CustomErr']) == len(dump)
+    assert len(early_stop.stopping_history['Train']['CustomErr']) == len(dump)
```
trivialfis, line 52: apologies for the mistake, I think we should modify the previous the `data_name` in line 39 to `data_name="Valid"` instead since it doesn't make sense to early stop with the training data.

UncleLLD: > apologies for the mistake, I think we should modify the previous the `data_name` in line 39 to `data_name="Valid"` instead since it doesn't make sense to early stop with the training data.

yes, you are right! I have changed it again, please review, thanks!

trivialfis (approved):
Thank you for the fix!

PR #10059, opened 2024-02-21
Use UBJSON for serializing splits for vertical data split.

trivialfis:
Explicitly serialize the split entry into UBJSON instead of relying on implicit binary representation.
- Restore the test for vector leaf.
- Use F64 for storing gradient.
- A new allgather V to avoid using special characters.

cc @rongou .

### DIFF
--- a/src/collective/communicator-inl.h
+++ b/src/collective/communicator-inl.h
@@ -1,5 +1,5 @@
 /**
- * Copyright 2022-2023 by XGBoost contributors
+ * Copyright 2022-2024, XGBoost contributors
  */
 #pragma once
 #include <string>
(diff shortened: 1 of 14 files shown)


trivialfis:
I think the failure is unrelated, will try to fix it in a different PR.

src/collective/communicator-inl.h
```diff
@@ -192,6 +192,9 @@ inline std::vector<T> AllgatherV(std::vector<T> const &input) {
   return result;
 }
 
+[[nodiscard]] std::vector<std::vector<char>> VectorAllgatherV(
```
rongou, line 195: Add some comments?

rongou (approved):

src/collective/communicator-inl.h
trivialfis: Done.

PR #10064, opened 2024-02-22
[CI] Fix JVM tests on GH Action

trivialfis:


hcho3:
Need some help?

trivialfis:
Thank you, I'm trying to resolve the JVM build issue on Windows https://github.com/dmlc/xgboost/pull/10059 Not sure what's updated on GitHub, just adding some logs at the moment and see why the Python package builds fine but the JVM package doesn't.

hcho3:
The latest Windows runner image `20240218.2.0` is somehow broken. The CI was fine for the previous version `20240211.1.0`.

trivialfis:
Considering the Python package is working fine, it might be caused by a conflict in environment variables between the `setup-java` action and the new image, but I can't be sure, just making random trial and error at the moment. It seems we can't downgrade or freeze the image version https://stackoverflow.com/questions/64882287/specify-runner-version-within-github-actions

hcho3:
I was able to fix the pipeline. For some reason, you can run `cmake .. -G"Visual Studio 17 2022"` outside `create_jni.py`, but running the same command inside `create_jni.py` leads to an error. It looks like `subprocess.run` does not have access to all environment variables.

hcho3:
Merging this to unblock the CI.

PR #10066, opened 2024-02-22
update spark_estimator doc

jmoralez:
Updates the documentation on the spark estimator to use `transform` instead of `predict`. Also fixes the name of the initial estimator.


trivialfis (approved):
Thank you for the fix!

Issue #10065, opened 2024-02-22
Allow users to set the logger level in XGBoost-PySpark


danmar3:
Hi, currently using XGBoost-PySpark in notebooks generates several log messages. I have not been able to turn them off. For example, when calling .transform, the notebook gets spammed with several messages like:

```
2024-02-22 08:44:49,270 INFO XGBoost-PySpark: predict_udf Do the inference on the CPUs
2024-02-22 08:44:49,275 INFO XGBoost-PySpark: predict_udf Do the inference on the CPUs
2024-02-22 08:44:49,310 INFO XGBoost-PySpark: predict_udf Do the inference on the CPUs
2024-02-22 08:44:49,403 INFO XGBoost-PySpark: predict_udf Do the inference on the CPUs
2024-02-22 08:44:49,411 INFO XGBoost-PySpark: predict_udf Do the inference on the CPUs
2024-02-22 08:44:49,535 INFO XGBoost-PySpark: predict_udf Do the inference on the CPUs
2024-02-22 08:44:49,603 INFO XGBoost-PySpark: predict_udf Do the inference on the CPUs
2024-02-22 08:44:49,616 INFO XGBoost-PySpark: predict_udf Do the inference on the CPUs
```

Currently, every time get_logger is called ([here](https://github.com/dmlc/xgboost/blob/252e018275195b5e6a7fc49f069dccb2ec2d8c46/python-package/xgboost/spark/core.py#L1415)), the logger information is set to INFO [here](https://github.com/dmlc/xgboost/blob/252e018275195b5e6a7fc49f069dccb2ec2d8c46/python-package/xgboost/spark/utils.py#L103C5-L103C27). 
This does not allow the user to set the logging level, as it is always re-set to 'INFO'.

I think this can be solved by removing the setLevel line [here](https://github.com/dmlc/xgboost/blob/252e018275195b5e6a7fc49f069dccb2ec2d8c46/python-package/xgboost/spark/utils.py#L103C5-L103C27).

Thank you

trivialfis:
Hmm, we need to find a way to unify all the logging levels.

trivialfis:
There's XGB logging, Python logging, spark logging, among some others.

wbo4958:
Let me have a PR to fix this issue.

wbo4958:
[...]

PR #10077, opened 2024-02-27
[pyspark] rework the log

wbo4958:
To fix https://github.com/dmlc/xgboost/issues/10065

It will show the log like below by default

```
2024-02-29 10:29:05,749 INFO XGBoost-PySpark: _fit Running xgboost-2.1.0-dev on 1 workers with
INFO:XGBoost-PySpark:Leveraging cuda:0 to train with QDM: off
2024-02-29 10:29:09,637 INFO XGBoost-PySpark: _fit Finished xgboost training!
2024-02-29 10:29:11,410 INFO XGBoost-PySpark: predict_udf CUDF or Cupy is unavailable, fallback the inference on the CPUs
```

users can disable the log by 

```
import logging

logging.getLogger('XGBoost-PySpark').setLevel('WARN') 
```

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -95,9 +95,9 @@ from .utils import (
     deserialize_xgb_model,
     get_class_name,
     get_logger,
+    get_logger_level,
     serialize_booster,
     use_cuda,
-    get_logger_level,
 )
 
 # Put pyspark specific params here, they won't be passed to XGBoost.
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -97,6 +97,7 @@ from .utils import (
     get_logger,
     serialize_booster,
     use_cuda,
+    get_logger_level,
 )
 
 # Put pyspark specific params here, they won't be passed to XGBoost.
@@ -180,6 +181,8 @@ Pred = namedtuple(
 pred = Pred("prediction", "rawPrediction", "probability", "predContrib")
 
 _INIT_BOOSTER_SAVE_PATH = "init_booster.json"
+
+_LOG_TAG = "XGBoost-PySpark"
 
 
 class _SparkXGBParams(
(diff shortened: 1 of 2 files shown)


trivialfis (commented):
I think this doesn't resolve the issue since the verbosity is still set by the get logger function?

wbo4958:
> I think this doesn't resolve the issue since the verbosity is still set by the get logger function?

It doesn't matter to just show a few lines of log by default, which can help debug.

trivialfis:
You link the issue and claim to fix it?

wbo4958:
> You link the issue and claim to fix it?

Yeah, I replied to it in the issue.

danmar3:
Hi @wbo4958 , thanks for looking into this. Reducing the number of lines printed is great, however I wanted to make an additional suggestion. How about in addition to the current solution, we also update the [get_logger](https://github.com/dmlc/xgboost/blob/252e018275195b5e6a7fc49f069dccb2ec2d8c46/python-package/xgboost/spark/utils.py#L100) as follows:

```python3
# make level optional
def get_logger(name: str, level: Optional[str] = None) -> logging.Logger:
    """Gets a logger by name, or creates and configures it for the first time."""
    logger = logging.getLogger(name)
    # <HERE> only update the level if not None
    if level is not None:
        logger.setLevel(level)
    # If the logger is configured, skip the configure
    if not logger.handlers and not logging.getLogger().handlers:
        handler = logging.StreamHandler(sys.stderr)
        formatter = logging.Formatter(
            "%(asctime)s %(levelname)s %(name)s: %(funcName)s %(message)s"
        )
        handler.setFormatter(formatter)
        logger.addHandler(handler)
    return logger
```

I believe users in general do not expect a logger level to be reset. It would be great if users could set the logger configuration at the beginning of their application/notebook, so the user chooses which logs to see/hide. Users will still be able to see the logs if we would like so by calling logging.getLogger('XGBoost-PySpark').setLevel('INFO') at the beginning of the notebook.

Thanks!

wbo4958:
@danmar3 Good suggestion, let me refine this PR for it.

trivialfis (approved):

Issue #9987, opened 2024-01-15
Vertical Federated Learning with Secure Features (secure inference and encrypted training) RFC


ZiyueXu77:
## Motivation
Current XGBoost introduced the support for Federated Learning, both horizontal and vertical. However, their capability in supporting secure features is limited. Based on basic arithmetic operations - addition and multiplication - that is supported by common Homomorphic Encryption schemes (Paillier, BFV/BGV, or CKKS), the current horizontal and vertical pipeline cannot be integrated. The reason is server and/or clients need to perform operations that are not supported by HE schemes, including division and argmax.
It will be useful to implement a variation of current vertical federated learning XGBoost to provide a solution with secure features. 

## Secure Pattern
With the basic HE operations of addition (and potentially multiplication), for vertical FL, a feasible solution is "[SecureBoost](https://arxiv.org/pdf/1901.08755.pdf)", for which all unsupported operations are performed from label-owner (server) side. The main purpose for this pipeline is to protect the label information from leaking to passive clients, and to protect each parties' feature value information from leaking to others. However in this case, passive parties will partially expose their cumulative histogram information to the active party (although one of the added features will be to minimize the potential leakage).

The main difference from our current HE solution for horizontal deep learning pipelines is:
* For horizontal applications, aggregation server has no data (and for most FL schemes, performs no training), mostly performing aggregation only. Hence in theory, server is the "minor contributor" from model training perspective, and clients have a concern of leakage to server. So in this case, the protection is against server.
* For vertical XGBoost application, label owner / active party holds the label, which can be considered the most valuable asset for the whole process. Therefore, the active party in this case is the "major contributor" from model training perspective, and it will have a concern of leaking this information to passive clients. So in this case, the protection is mainly against passive clients.

Other secure patterns for XGBoost, including horizontal and vertical schemes, mostly will need more sophisticated HE support, and we will put those for future work. 

## Goals
* Enhance XGBoost to support secure vertical federated learning.
* Support using [NVFlare](https://github.com/NVIDIA/NVFlare) to coordinate the learning process, but the design should be amenable to support other federated learning platforms.
* Support using any arbitrary encryption library, decoupled from xgboost via a secure interface.
* Efficiency: training speed should be close to alternative distributed training design.
* Accuracy: should be close to alternative vertical pipeline.
 
## Non-Goals
[...]

PR #10079, opened 2024-02-27
Implement secure boost scheme - secure evaluation and validation (during training) without local feature leakage

ZiyueXu77:
For implementing Vertical Federated Learning with Secure Features, as discussed in
https://github.com/dmlc/xgboost/issues/9987
This part is independent from the encryption and the alternative vertical pipeline. The purpose is to avoid leaking the real cut value information from participants. Hence add as a separate PR. 
This PR is based on #10037, which should be reviewed and merged first.


trivialfis:
hi, what's the relationship between this PR and https://github.com/dmlc/xgboost/pull/10037 ?

trivialfis:
Thank you for sharing, do you want me to set up a different feature branch than the master for you to play with? It seems much is still in flux and is heavily modifying the core, I can review PR for the different branches and the CI can run as well.

ZiyueXu77:
> Thank you for sharing, do you want me to set up a different feature branch than the master for you to play with? It seems much is still in flux and is heavily modifying the core, I can review PR for the different branches and the CI can run as well.

sure! that will be helpful, thanks.

ZiyueXu77:
> Done https://github.com/dmlc/xgboost/tree/vertical-federated-learning . You can change the github PR base branch by using the "Edit" next to the PR title.

Thanks!

ZiyueXu77:
Hi @trivialfis , the method implementation part for secure inference is ready, I added detailed information to our [RFC](https://github.com/dmlc/xgboost/issues/9987) under the section "Design for Secure Inference - avoid leakage of feature cut value".  @YuanTingHsieh will add / make modifications to the unit testing. Thanks!

src/common/quantile.cc
```diff
@@ -445,29 +449,27 @@ void SketchContainerImpl<WQSketch>::MakeCuts(Context const *ctx, MetaInfo const
       max_cat = std::max(max_cat, AddCategories(categories_.at(fid), p_cuts));
```
trivialfis, line 449: Based on my understanding, categorical features are not yet supported right?

src/common/quantile.cc
```diff
@@ -361,21 +361,23 @@ void SketchContainerImpl<WQSketch>::AllReduce(
...
+  // add the cut points
   auto &cut_values = cuts->cut_values_.HostVector();
-  // if empty column, fill the cut values with 0
+  // if secure and empty column, fill the cut values with NaN
   if (secure && (required_cuts_original == 0)) {
     for (size_t i = 1; i < required_cuts; ++i) {
-      cut_values.push_back(0.0);
+      cut_values.push_back(std::numeric_limits<double>::quiet_NaN());
     }
+    return std::numeric_limits<double>::quiet_NaN();
```
trivialfis, line 380: I think the function is returning an indicator `last_value = nan` or `last_value != nan`, which can be simply done by simply returning `true` and `false`.

src/tree/common_row_partitioner.h
```diff
@@ -106,26 +106,39 @@ class CommonRowPartitioner {
...
-          split_cond = static_cast<bst_bin_t>(bound);
+      if (is_index) {
+        // if the split_pt is already recorded as a bin_id, use it directly
+        (*split_conditions)[i] = static_cast<int32_t>(split_pt);
+        // at this point, each participants received the best split index,
+        // therefore can recover the split_pt from bin_id, update tree info
+        auto split_pt_local = vals[split_pt];
+        // make updates to the tree, replacing the existing index
+        // with cut value, so as to be consistent with the tree model format
+        const_cast<RegTree::Node&>(tree.GetNodes()[nidx]).SetSplit(fidx, split_pt_local);
```
trivialfis, line 126: Unless there's really good reason (like interfacing legacy C libraries), const_cast is discouraged, please consider making the necessary cleanup/refactor instead of pushing an experimental edit.

Aside, could you please help look into whether it's possible to extract this float -> integer bin index logic, and move the tree split logic into the evaluator? Currently, it's the only place the modifies a tree, and it would be great if we could keep it that way.

src/tree/common_row_partitioner.h
```diff
@@ -194,7 +207,12 @@ class CommonRowPartitioner {
     std::vector<int32_t> split_conditions;
     if (column_matrix.IsInitialized()) {
       split_conditions.resize(n_nodes);
-      FindSplitConditions(nodes, *p_tree, gmat, &split_conditions);
+      if (is_secure_) {
+        // in secure mode, the split index is kept instead of the split value
+        FindSplitConditions(nodes, *p_tree, gmat, &split_conditions, true);
```
trivialfis, line 212: The condition looks exactly the same as: `FindSplitConditions(nodes, *p_tree, gmat, &split_conditions, is_secure_);`

src/tree/hist/evaluate_splits.h
```diff
@@ -303,22 +303,35 @@ class HistEvaluator {
...
+                                                         GradStats{right_sum}) - parent.root_gain);
+          if (!is_secure_) {
+            split_pt = cut_val[i];  // not used for partition based
+            best.Update(loss_chg, fidx, split_pt, d_step == -1, false, left_sum, right_sum);
+          } else {
+            // secure mode: record the best split point, rather than the actual value
+            // since it is not accessible at this point (active party finding best-split)
+            best.Update(loss_chg, fidx, i, d_step == -1, false, left_sum, right_sum);
+          }
         } else {
```
trivialfis, line 315: Do you think a [policy class](https://en.wikipedia.org/wiki/Modern_C%2B%2B_Design#Policy-based_design) might help here? Or maybe there are other efficient ways to handle these conditions? I'm loosing track of these conditions, considering that we have three enumeration functions:
- numeric
- partition
- one hot
Then we have three split modes:
- column
- row
- column + secure
So, in combination, 9 potential cases, and we haven't counted vector leaf yet. Need to find a better way to manage these conditions.

src/common/quantile.cc
ZiyueXu77: right, need to find a proper testing data with categorical features

src/common/quantile.cc
ZiyueXu77: good point! will make updates to bool

src/tree/common_row_partitioner.h
ZiyueXu77: yeah... I tried to find a way to make updates to the split value, but ended up with this solution, will revisit to see if there are alternative ways, maybe adding a function to tree node spec

src/tree/hist/evaluate_splits.h
ZiyueXu77: agreed, will make some updates

src/tree/common_row_partitioner.h
ZiyueXu77: thanks! will check

src/common/quantile.cc
ZiyueXu77: updated

src/tree/common_row_partitioner.h
ZiyueXu77: moved this "split_value" recovery process to evaluate_splits.h, right after the global best split allgather

src/tree/common_row_partitioner.h
ZiyueXu77: after the above change, this is no longer needed

src/tree/hist/evaluate_splits.h
ZiyueXu77: Another thing regarding this mode combinations: potentially with the upcoming processor interface we will be able to enable encrypted horizontal, shall we further add a row + secure mode, adding a 4th one for  
`enum class DataSplitMode : int { kRow = 0, kCol = 1, kColSecure = 2 };`? (or maybe there are better options?)

ZiyueXu77:
> Hi, could you please share how to run some high level tests?

Sure, this is what I am using for testing:
https://github.com/ZiyueXu77/NVFlare/tree/secureboost/examples/advanced/xgboost_secure
will share the data link with you

ZiyueXu77:
Another general challenge for any vertical pipelines: at inference time, all parties need to be online, and as our model records the "global feature index", the "order" of the clients need to remain the same. We may need some mechanisms to ensure this order.

src/tree/hist/evaluate_splits.h
trivialfis: For my preference, I would have put it in the `CommunicatorContext` configuration for whether the channel is encrypted.

trivialfis:
> the "order" of the clients need to remain the same. We may need some mechanisms to ensure this order.

I will leave that to nvflare.

trivialfis (commented):
The code looks good to me overall. We can merge it once we have some basic unittests.

As for integration tests in Python with nvflare (in future PRs), we can assert that
- models are different for different workers.
- predictions are the same
- evaluation result are the same
- only works if the 0th worker has the label.

I highly recommend using the hypothesis test framework (see python tests in xgboost and search the term hypothesis).

ZiyueXu77:
> > the "order" of the clients need to remain the same. We may need some mechanisms to ensure this order.
> 
> I will leave that to nvflare.
agree, makes sense

ZiyueXu77:
> The code looks good to me overall. We can merge it once we have some basic unittests.
> 
> As for integration tests in Python with nvflare (in future PRs), we can assert that
> 
> * models are different for different workers.
> * predictions are the same
> * evaluation result are the same
> * only works if the 0th worker has the label.
> 
> I highly recommend using the hypothesis test framework (see python tests in xgboost and search the term hypothesis).

Thanks! @YuanTingHsieh , could you add the unit tests according to @trivialfis 's suggestions?

trivialfis:
> could you add the unit tests according to @trivialfis 's suggestions

Those points are all for integration tests not for small unittest. I think the integration tests in Python with nvflare will take more effort, we don't need to rush it in this PR.

trivialfis:
Hi, is there any update?

ZiyueXu77:
> Hi, is there any update?

Thanks for asking! :) @YuanTingHsieh has been busy with a related NVFlare release in the past two weeks, now the release is close to finish, he will have time to work on this soon.

trivialfis:
That should be unrelated, will look into this PR today.

trivialfis (approved):

ZiyueXu77:
Hi @trivialfis , thanks for the updates, just merged it. 
Everything passed, except an error for R regarding "matrix":
`2024-04-29T13:40:51.6394593Z make: *** [Makefile:291: Matrix.ts] Error 1`

trivialfis:
Triggered the rest of the CI.

ZiyueXu77:
Hi @trivialfis , there are 3 failed checks, but I think they align with the rebase merge, shall we just merge this? Thanks!

trivialfis:
Please fix errors on buildkite.

ZiyueXu77:
And the other two errors are

`File "/__w/xgboost/xgboost/tests/ci_build/test_r_package.py", line 166, in check_rpackage
    with open(rcheck_dir / "00install.out", "r") as fd:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'xgboost.Rcheck/00install.out'` 

`_bk;t=1714406293923An error occurred (RepositoryAlreadyExistsException) when calling the CreateRepository operation: The repository with name 'xgb-ci.clang_tidy11.8' already exists in the registry with id '492475357299'`

I don't quite get what they indicate.

trivialfis:
Let's ignore the Windows error for now and get the Linux buildkite ci to pass: https://buildkite.com/xgboost/xgboost-ci/builds/5116#018f2a11-c525-41a1-a9d7-81bfdb1f701c

ZiyueXu77:
> Let's ignore the Windows error for now and get the Linux buildkite ci to pass: https://buildkite.com/xgboost/xgboost-ci/builds/5116#018f2a11-c525-41a1-a9d7-81bfdb1f701c

Thanks! addressed the warning in clang-tidy

trivialfis:
Excellent! Please continue to fix other errors in the Linux CI (at the moment, it's the memory leak reported by the leak sanitizer).

PR #10092, opened 2024-03-07
make DMatrix subclassable that supports slice mehod

xbanke:
Currently, subclass of `DMatrix` could not slice normally. `slice` method on subclass will return a `DMatrix` instance not the subclass instance.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -1295,7 +1295,7 @@ class DMatrix:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 ctypes.c_int(1 if allow_groups else 0),
             )
         )
-        return DMatrix(handle)
+        return self.__class__(handle)
 
     @property
     def feature_names(self) -> Optional[FeatureNames]:


trivialfis (commented):
It's probably not a safe assumption that the subclass implements an init method that takes only a C handle.

Is there any reason that you don't use numpy and alike for slicing data? If you want to save memory, the Quantile DMatrix can be used. In general, the DMatrix class is intended to be a temporary storage for XGBoost instead of a general data manipulation tool.

xbanke:
In my case, I want the DMatrix been attached the index. This make coding simple and convenient. In most of the cases, I use pandas.DataFame that has a datetime index as the raw data, and I need do slice by date range. There is some code sample.

```py
from typing import Any, Self
from functools import wraps
import pandas as pd
import xgboost as xgb


class MyDMatrix(xgb.DMatrix):
    def __init__(self, data: Any, *args: Any, index: pd.Index | None = None, **kwargs: Any) -> None:
        super().__init__(self, data, *args, **kwargs)
        self._set_index(index)

    def _set_index(self, idx: pd.Index | None = None) -> None:
        n_rows = self.num_row()
        if idx is None:
            idx = pd.Index(np.arange(n_rows))
        else:
            assert len(idx) == n_rows
        self._index = idx 

    @property
    def index(self) -> pd.Index:
        return self._index

    def slice(self, idx: Any) -> Self:
        index = self.index[idx]
        dmatrix = super().slice(idx)
        dmatrix._set_index(index)
        return dmatrix

    def get_rindex(self, start: pd.Timestamp | None = None, end: pd.Timestamp | None = None) -> list[int]:
        ...


class MyBooster(xgb.Booster):
    
    def predict(
        self,
        ds: MyDMatrix,
        **kwargs: Any,
    ) -> pd.DataFrame:
        preds = super().predict(ds, **kwargs)
        if len(preds.shape) == 1:
            preds = preds.reshape((-1, 1))
        return pd.DataFrame(preds, index=ds.index)


@wraps(xgb.train)
def train(params: dict[str, Any], dtrain: MyDMatrix, num_boost_round: int = 10, **kwargs:Any) -> MyBooster:
    booster = xgb.train(params=params, dtrain=dtrain, num_boost_round=num_boost_round, **kwargs)
    return MyBooster(model_file=booster)
```

One more thing, `xgb.Booster` and `xgb.DMatrix` are both pure python class not C class from cython or somewhere, I think it's not a big problem to subclass them. If anyone really want to do that just like me, he/she just needs overriding the slice method in case of `DMatrix`

trivialfis:
Thank you for sharing, I think the users would indeed need to override the slice method as you have suggested. The approach used in this PR assumes a correct implementation of the `__init__` method from the subclass, which, from my perspective, is more difficult to do than an override.

Will close the PR now. Feel free to continue the discussion.

PR #10093, opened 2024-03-07
Support graphviz plot for multi-target tree.

trivialfis:
When the number of targets is less than or equal to 3, the full leaf value is displayed, otherwise, it's printed as `leaf=[1.1, 1.2, ..., 3.4]`. The threshold is currently hard coded, for detailed inspection of leaf value, we will cover other forms of tree dump later.

### DIFF
--- a/src/tree/tree_model.cc
+++ b/src/tree/tree_model.cc
@@ -767,8 +767,8 @@ XGBOOST_REGISTER_TREE_IO(GraphvizGenerator, "dot")
 constexpr bst_node_t RegTree::kRoot;
 
 std::string RegTree::DumpModel(const FeatureMap& fmap, bool with_stats, std::string format) const {
-  if (format != "dot") {
-    MTNotImplemented();
+  if (this->IsMultiTarget() && format != "dot") {
+    LOG(FATAL) << format << " tree dump " << MTNotImplemented();
   }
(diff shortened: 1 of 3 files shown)


src/tree/tree_model.cc
```diff
@@ -733,7 +767,9 @@ XGBOOST_REGISTER_TREE_IO(GraphvizGenerator, "dot")
 constexpr bst_node_t RegTree::kRoot;
 
 std::string RegTree::DumpModel(const FeatureMap& fmap, bool with_stats, std::string format) const {
-  CHECK(!IsMultiTarget());
+  if (format != "dot") {
+    MTNotImplemented();
+  }
```
hcho3, line 772:
```suggestion
  CHECK_EQ(format, "dot") << "Dump format " << format << MTNotImplemented();
```

hcho3, line 772:
```suggestion
  CHECK_EQ(format, "dot") << "Dump format " << format << MTNotImplemented();
```

hcho3 (changes requested):

src/tree/tree_model.cc
trivialfis: No, we don't require the check for scalar trees

trivialfis: Fixed the error message.

hcho3 (approved):

Issue #10099, opened 2024-03-08
`rank:pairwise` conflict with `ndcg_exp_gain` = `False`


xbanke:
If `ndcg_exp_gain` = `True`, the degress must be less or equal to 31.

Is there a way that totally compatible with version `1.7.6` if objective is `rank:pairwise`?

trivialfis:
May I assume that you are using the NDCG metric with the pairwise objective?

xbanke:
> May I assume that you are using the NDCG metric with the pairwise objective?

No, I have my own metric: correlation coefficient. So it does nothing with NDCG

xbanke:
Thank you very much.

PR #10100, opened 2024-03-08
Fix pairwise objective with NDCG metric along with custom gain.

trivialfis:
- Allow setting `ndcg_exp_gain` for `rank:pairwise`.

This is useful when using pairwise for objective but ndcg for metric.

- Add default configuration in pairwise to pass the `ndcg_exp_gain` to metrics.

Close https://github.com/dmlc/xgboost/issues/10099 .

cc @xbanke After this PR, you can run the ranknet loss with NDCG and custom label by turning `ndcg_exp_gain` off. The switch is still necessary since the NDCG metric also validates the input label.


xbanke:
When will this been merged?

hcho3 (approved):

PR #10032, opened 2024-02-05
[R] Rename `watchlist` -> `evals`

david-cortes:
ref https://github.com/dmlc/xgboost/issues/9810
To be merged after previous PR: https://github.com/dmlc/xgboost/pull/9957

This PR renames the parameter 'watchlist' in the R interface to 'evals' , like it is named in the python interface.

The name 'watchlist' is quite unintuitive, since one tends to call this names along the lines of "evaluation set", "validation data", etc.

PR #10110, opened 2024-03-11
[R] warning for deprecated watchlist argument in xgb.train()

mayer79:
Regarding the renaming of watchlist to evals in `xgb.train()?  as per https://github.com/dmlc/xgboost/pull/10032 , we might consider raising a warning.

Often, the main training logic in R is to use `xgb.train()` with early stopping on the watchlist. By renaming the `watchlist` argument, such cases will accidently use the training data for early stopping, which means that the model does not stop training until the maximum possible number of rounds is reached. This happens silently, so we might consider this warning.


trivialfis:
@david-cortes

trivialfis (approved):

PR #10057, opened 2024-02-20
SYCL. Add functional for sycl implementation of RowSetCollection.

razdoburdin:
Hi,
I continue adding of sycl support (https://github.com/dmlc/xgboost/pull/9691, https://github.com/dmlc/xgboost/pull/9846, https://github.com/dmlc/xgboost/pull/9800, https://github.com/dmlc/xgboost/pull/10011, https://github.com/dmlc/xgboost/pull/10045).
By this PR I add the next part of the training-related code for the sycl devices. Here I add logic related to RowSetCollectionan and the related tests.
This PR itself doesn't add any new functional, but I plan to combine all related parts one by one to a launchable sycl-training.

PR #10119, opened 2024-03-12
SYCL. Add functional for split evaluation

razdoburdin:
Hi,
I continue adding of sycl support (https://github.com/dmlc/xgboost/pull/10080, https://github.com/dmlc/xgboost/pull/10057, etc).
By this PR I add the next part of the training-related code for the sycl devices. Here I add logic related to split evaluation and the related tests.

### DIFF
--- a/plugin/sycl/tree/param.h
+++ b/plugin/sycl/tree/param.h
@@ -36,9 +36,7 @@ struct TrainParam {
   float reg_alpha;
   float max_delta_step;
 
-
   TrainParam() {}
-
 
   explicit TrainParam(const xgboost::tree::TrainParam& param) {
     reg_lambda = param.reg_lambda;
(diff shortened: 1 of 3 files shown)


plugin/sycl/tree/param.h
```diff
@@ -0,0 +1,104 @@
+/*!
+ * Copyright 2014-2023 by Contributors
```
napetrov, line 2: 2024

plugin/sycl/tree/split_evaluator.h
```diff
@@ -0,0 +1,210 @@
+/*!
+ * Copyright 2018-2023 by Contributors
```
napetrov, line 2: 2024

napetrov (approved):
Looks good

plugin/sycl/tree/param.h
razdoburdin: done

plugin/sycl/tree/split_evaluator.h
razdoburdin: done

plugin/sycl/tree/param.h
```diff
@@ -0,0 +1,104 @@
...
+
+#include <CL/sycl.hpp>
+
+namespace xgboost {
+namespace sycl {
+namespace tree {
+
+
+/*! \brief Wrapper for necessary training parameters for regression tree to access on device */
+struct TrainParam {
```
trivialfis, line 30: Can we reuse the existing one?

plugin/sycl/tree/param.h
```diff
@@ -0,0 +1,104 @@
...
+    reg_lambda = param.reg_lambda;
+    reg_alpha = param.reg_alpha;
+    min_child_weight = param.min_child_weight;
+    max_delta_step = param.max_delta_step;
+  }
+};
+
+/*! \brief core statistics used for tree construction */
+template<typename GradType>
+struct GradStats {
```
trivialfis, line 50: Can we reuse the existing one?

plugin/sycl/tree/split_evaluator.h
```diff
@@ -0,0 +1,210 @@
...
+      } else if (w < lower[nodeid]) {
+        return lower[nodeid];
+      } else if (w > upper[nodeid]) {
+        return upper[nodeid];
+      } else {
+        return w;
+      }
+    }
+
+    inline GradType Sqr(GradType a) const { return a * a; }
```
trivialfis, line 149: Can we use the one in common/math.h ?

trivialfis (commented):
Hi, thank you for the continued work. It would be great if we could reduce the amount of copied code when possible.

plugin/sycl/tree/param.h
razdoburdin: Unfortunately, no. The original one has fields of type std::vector, that are not copyable to sycl-device.

plugin/sycl/tree/split_evaluator.h
razdoburdin: done

PR #10120, opened 2024-03-12
Use `std::uint64_t` for row index.

trivialfis:
- Use std::uint64_t instead of size_t to avoid implementation defined type.

  It's difficult to define template instantiation with implementation defined type.

- Rename to `bst_idx_t`, to account for other types of indexing, including directly indexing data as a 1-d array.

- Small cleanup to the base header.

### DIFF
--- a/src/data/adapter.h
+++ b/src/data/adapter.h
@@ -73,11 +73,11 @@ constexpr size_t kAdapterUnknownSize = std::numeric_limits<size_t >::max();
 
 struct COOTuple {
   COOTuple() = default;
-  XGBOOST_DEVICE COOTuple(size_t row_idx, size_t column_idx, float value)
+  XGBOOST_DEVICE COOTuple(bst_idx_t row_idx, bst_idx_t column_idx, float value)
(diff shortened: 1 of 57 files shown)


src/data/adapter.h
```diff
@@ -178,29 +174,25 @@ class CSRAdapterBatch : public detail::NoMetaInfo {
 
 class CSRAdapter : public detail::SingleBatchDataIter<CSRAdapterBatch> {
  public:
-  CSRAdapter(const size_t* row_ptr, const unsigned* feature_idx,
-             const float* values, size_t num_rows, size_t num_elements,
-             size_t num_features)
-      : batch_(row_ptr, feature_idx, values, num_rows, num_elements,
-               num_features),
+  CSRAdapter(const size_t* row_ptr, const unsigned* feature_idx, const float* values,
```
hcho3, line 177: Should we keep using `size_t` for `row_ptr`?

trivialfis: I think we can remove this class in the future since we are changing to the array interface. But I will leave it untouched for this PR

hcho3 (approved):

PR #10073, opened 2024-02-23
Processor Interface

nvidianz:
Processor interface with the dummy implementation.

PR #10124, opened 2024-03-15
Implement secure boost scheme phase 2 - enable secure xgboost via processor interface

ZiyueXu77:
Add extensions to the vertical pipeline implemented in phase 1 to call Processor interface at proper locations to enable secure features:
* Before broadcasting g/h pairs, call interface to prepare buffer
* Communicate GHistIndexMatrix and RowSetCollection to interface for local gRPC handler in response to AllGather call


ZiyueXu77:
This PR is dependent on https://github.com/dmlc/xgboost/pull/10073, which should be merged first

src/learner.cc
```diff
@@ -493,6 +493,14 @@ class LearnerConfiguration : public Learner {
 
     this->ConfigureMetrics(args);
 
+    xgboost::processing::ProcessorLoader loader;
+    processor_instance = loader.load("dummy");
```
ZiyueXu77, line 497: Hi @trivialfis, to enable secureboost, we need to initiate this instance of processor, which will take some parameters, currently they are hard-coded. We wonder if you have some suggestion on where to inject the parameters? I thought about via training params, but kind of feeling awkward as it is only related to a particular federated scheme.

ZiyueXu77 (commented):
Question about processor interface config

ZiyueXu77:
Hi @trivialfis, this PR is ready for review, we have some remaining failed checks due to two issues:
3 windows checks: Cannot open include file: 'dlfcn.h' -> @nvidianz is looking into this
2 mac_os: Error: The version '3.8' with architecture 'arm64' was not found for macOS 14.4.1. -> this seems to be a recent issue impacting everyone.

src/processing/processor.h
```diff
@@ -0,0 +1,132 @@
...
+
+namespace processing {
+
+const char kLibraryPath[] = "LIBRARY_PATH";
+const char kMockProcessor[] = "mock";
+const char kLoadFunc[] = "LoadProcessor";
+
+/*! \brief An processor interface to handle tasks that require external library through plugins */
+class Processor {
+ public:
```
shaojian-ant, line 20: Here is my understanding about these interfaces. Please confirm if it's correct:
ProcessGHPairs: active party encrypts g&h pairs
HandleGHPairs: passive parties save encrypted g&h pairs in local
ProcessAggregation: passive parties each cacluate it's histogram based on encrypted g&h pairs, the output is ciphertext
HandleAggregation: active party decrypts the allgathered historgram ciphertexts
ProcessHistograms: active party encrypted allgathered & decrypted histograms?
HandleHistograms: active parth decrypts the allgathered historgram ciphertexts?

I have two more questions:
1. Where are ProcessHistograms and HandleHistograms needed in secure xgboost?
2. Passive parties need a public key to perform HE addition calculations. The public key is generated by the active party. So I think there should be two more interfaces like ProcessPublicKey and HandlePublicKey.

ZiyueXu77: Hi @shaojian-ant , thanks for the question!
"ProcessHistograms" and "HandleHistograms" are for the upcoming secure horizontal scheme, I will remove these two from this PR to avoid confusion.
Under federated mode, the HE is supposed to be performed outside xgboost via the interface, hence the key management, etc. will not be part of xgboost lib (and hence in xgboost we only have a "mock" interface inplementation to illustrate the functionality)

tests/cpp/processing/test_processor.cc
```diff
@@ -0,0 +1,86 @@
...
+ public:
+    void SetUp() override {
+        auto loader = processing::ProcessorLoader();
+        processor_ = loader.load(processing::kMockProcessor);
+        processor_->Initialize(true, {});
+    }
+
+    void TearDown() override {
+        processor_->Shutdown();
+        processor_ = nullptr;
```
shaojian-ant, line 20: Whether processor_  is always on the heap,  and need :  delete processor_

src/processing/processor.h
```diff
@@ -0,0 +1,112 @@
...
+    /*!
+     * \brief Handle buffers with encoded pairs received from broadcast
+     *
+     * \param size Output buffer size
+     * \param The encoded buffer
+     * \param The encoded buffer size
+     *
+     * \return The encoded buffer
+     */
+    virtual void* HandleGHPairs(size_t *size, void *buffer, size_t buf_size) = 0;
```
shaojian-ant, line 61: Whether the returned pointer from HandGHPairs need be freed by FreeBuffer method?

src/processing/processor.h
shaojian-ant: Thanks for your reply!!!

src/processing/processor.h
ZiyueXu77: Thanks for the question! Indeed we would need properly free the buffer, will update our method. BTW, is there a preferred location in xgboost where the memory collections are performed at the end of a training session?

trivialfis: No, XGBoost is mostly following RAII of C++.

ZiyueXu77:
Hi @trivialfis, this PR is ready for review, currently we have 6 failed checks, but I think they are out of the scope of this PR, we could have more issues when triggering buildkite:
- 3 for windows, as we do not support windows at this time
- 2 for R, due to Error: `Failed to get R release: Failed to get Ghostscript`
- 1 for pylint, related to spark.core (is this a new thing?) 
```
Run python tests/ci_build/lint_python.py --format=0 --type-check=0 --pylint=1
************* Module xgboost.spark.core
xgboost/spark/core.py:1450:5[7](https://github.com/dmlc/xgboost/actions/runs/9134298149/job/25119550353?pr=10124#step:7:8): E0606: Possibly using variable 'base_margin_col' before assignment (possibly-used-before-assignment)
```

ZiyueXu77:
combined with https://github.com/dmlc/xgboost/pull/10231, close this one

PR #10112, opened 2024-03-11
Revamp the rabit implementation.

trivialfis:
# motivation
With the increasing complexity of the networking module due to support for vertical and horizontal federated learning and GPU-based training for scaling, the existing rabit module is no longer sufficient. We have multiple dimensions of features to support:

- data split.
- federated learning.
- GPU acceleration.

Among the above features, GPU acceleration and federated learning require loading optional external libraries. Lastly, we are trying to support resilience.

# features
This PR replaces the original RABIT implementation with a new one, which has already been partially merged into XGBoost. The new one features:
- Federated learning for both CPU and GPU.
- NCCL.
- More data types.
- A unified interface for all the underlying implementations.
- Improved timeout handling for both tracker and workers.
- Exhausted tests with metrics (fixed a couple of bugs along the way).
- A reusable tracker for Python and JVM packages.

**todos**:
- [x] JVM
- [x] Standardize the naming of worker parameters and tracker methods.
- [x] Worker sortby.

# working in progress

Retry is still in progress. This is to provide essential support for handling exception (e.g., a network error or an OOM). Segfault handling has to be done with additional cooperation with the distributed framework and is out of scope for this work.

# note for review
- Breaking changes are made to the tracker in Python and JVM interfaces.
- communicator parameters are reworked. For a list of parameters, see the document in the C header.
- Federated tracker requires an additional `n_workers` parameter.

# Related
Close https://github.com/dmlc/xgboost/issues/8191
Close https://github.com/dmlc/xgboost/issues/4981
Close https://github.com/dmlc/xgboost/issues/4781

PR #10152, opened 2024-03-27
Update collective implementation.

trivialfis:
- Cleanup resource during `Finalize` to avoid handling threads in destructor.
- Calculate the size for allgather automatically.
- Use simple allgather for small (smaller than the number of worker) allreduce.

Extracted from https://github.com/dmlc/xgboost/pull/10112 .


trivialfis:
cc @rongou .

rongou (approved):

PR #10174, opened 2024-04-10
SYCL. Add basic functional for QuantileHistMaker

razdoburdin:
Hi,
I continue adding of sycl support (https://github.com/dmlc/xgboost/pull/10138, https://github.com/dmlc/xgboost/pull/10119  etc).
By this PR I add the next part of the training-related code for the sycl devices. Here I add basic logic related to hist maker and the related tests. Now it is possible to create sycl hist maker with `TreeUpdater::Create`.

### DIFF
--- a/plugin/sycl/tree/updater_quantile_hist.h
+++ b/plugin/sycl/tree/updater_quantile_hist.h
@@ -1,0 +1,102 @@
+/*!
+ * Copyright 2017-2024 by Contributors
+ * \file updater_quantile_hist.h
+ */
+#ifndef PLUGIN_SYCL_TREE_UPDATER_QUANTILE_HIST_H_
+#define PLUGIN_SYCL_TREE_UPDATER_QUANTILE_HIST_H_
+
+#include <dmlc/timer.h>
+#include <xgboost/tree_updater.h>
+
+#include "../data/gradient_index.h"
+#include "../common/hist_util.h"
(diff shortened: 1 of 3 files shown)


razdoburdin:
Hi @trivialfis , could we merge it?

plugin/sycl/tree/updater_quantile_hist.h
```diff
@@ -0,0 +1,105 @@
...
+      LOG(WARNING) << "Attempted to load interal configuration for a model file that was generated "
+        << "by a previous version of XGBoost. A likely cause for this warning is that the model "
+        << "was saved with saveRDS() in R or pickle.dump() in Python. We strongly ADVISE AGAINST "
+        << "using saveRDS() or pickle.dump() so that the model remains accessible in current and "
+        << "upcoming XGBoost releases. Please use xgb.save() instead to preserve models for the "
+        << "long term. For more details and explanation, see "
+        << "https://xgboost.readthedocs.io/en/latest/tutorials/saving_model.html";
+      this->hist_maker_param_.UpdateAllowUnknown(Args{});
+    }
+  }
```
trivialfis, line 77: I think this is not necessary, we drop the configuration in learner if the versions don't match.

razdoburdin: done

PR #10179, opened 2024-04-12
Update python3statement URL

tqa236:
See https://github.com/python3statement/python3statement.github.io/issues/292.

To avoid having to pay for the domain indefinitely the maintainers are now redirecting to GitHub Pages.


trivialfis (approved):
Thank you for updating the doc!

PR #10199, opened 2024-04-17
[coll] Improve event loop.

trivialfis:
- Add a test for blocking calls.
- Do not require the queue to be empty after wake up, this frees up the thread to answer blocking call.
- Handle EOF in read.
- Improve the error message in the result. Allow concatenation of multiple results.

Extracted from https://github.com/dmlc/xgboost/pull/10112


trivialfis:
cc @rongou .

rongou (approved):

PR #10203, opened 2024-04-18
[coll] Add global functions.

trivialfis:
Tests are still in https://github.com/dmlc/xgboost/pull/10112 since we haven't swapped out the global implementation in this PR yet.


trivialfis:
cc @rongou .

rongou (approved):

PR #10208, opened 2024-04-18
[coll] Implement shutdown for tracker and comm.

trivialfis:
- Allow force shutdown for the tracker.
- Implement shutdown notice for error handling thread in comm.

Extracted from https://github.com/dmlc/xgboost/pull/10112 .


trivialfis:
cc @rongou .

PR #10220, opened 2024-04-24
[pyspark] Sort workers by task ID.

trivialfis:
The PySpark interface uses partition ID as task ID. Unlike the ordering of workers, partitions should be deterministic.


wbo4958:
LGTM.

PR #10219, opened 2024-04-24
Pull CI changes into vertical federated learning branch.

trivialfis:

PR #10225, opened 2024-04-25
[CI] Use native arm64 worker in GHAction to build M1 wheel

hcho3:
See https://github.com/dmlc/xgboost/pull/10219#issuecomment-2075847580


hcho3:
We might be able to retire the Mac Mini from BuildKite entirely, to minimize maintenance in the future.

hcho3:
@trivialfis Can you take a look at failing mypy?

trivialfis:
It's probably a bug in mypy, I run into it as well after upgrading mypy.

trivialfis:
https://github.com/python/mypy/issues/17166

trivialfis (approved):

PR #10226, opened 2024-04-26
[pyspark] clear temp storage to release GPU memory before each iteration

wbo4958:
When setting the current data batch (already copied to GPU from CPU) to the DMatrixProxy, it still holds the previous data batch during building QDM, which will increase the peak GPU  memory usage. So this PR tried to clear the temporary storage before each iteration.

I tested it on my local standalone cluster with 1 worker node with 12G GPU memory, XGBoost can train (4800000, 750)  with this PR, while XGBoost can train only (4300000, 750) without this PR.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -632,8 +632,12 @@ class DataIter(ABC):  # pylint: disable=too-many-instance-attributes
             )
             self._data_ref = ref
 
+        def _next() -> int:
+            self.clear()
+            return self.next(input_data)
+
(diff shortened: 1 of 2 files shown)


trivialfis:
Thank you for working on this. Is it possible to make this change universal instead of spark-only? The temporary data is a reference to transformed features like encoded categories or converted data types, it's to make sure they don't get garbage collected before the next iteration.

trivialfis:
Closing as implemented.

Issue #10221, opened 2024-04-24
NumPy 2.0 support


jakirkham:
NumPy 2.0 is coming out soon ( https://github.com/numpy/numpy/issues/24300 ). NumPy 2.0.0rc1 packages for conda & wheels came out 3 weeks back ( https://github.com/numpy/numpy/issues/24300#issuecomment-2030603395 )

To prepare for NumPy 2.0, it might be worthwhile to start testing against NumPy 2 in CI

NumPy has put out [a migration guide]( https://numpy.org/devdocs/numpy_2_0_migration_guide.html#numpy-2-migration-guide ). More details are in [the release notes]( https://numpy.org/devdocs/release/2.0.0-notes.html )

If XGBoost make use of NumPy's C API (and produces wheels that use it), having a release of XGBoost with wheels built against NumPy 2.0.0rc1 would be helpful to ensure NumPy 1 & 2 compatible wheels (as wheels built against NumPy 1 won't be compatible with NumPy 2). More details in [this NumPy 2 ABI doc]( https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-abi-handling )

Also as NumPy is tracking ecosystem support for NumPy 2.0, it would be helpful to share XGBoost current support status in issue (with any plans): https://github.com/numpy/numpy/issues/26191

jakirkham:
cc @hcho3

trivialfis:
Thank you for the heads up. Will do some tests, I think most of of XGBoost should be fine other than some removed symbols

PR #10252, opened 2024-05-06
Fixes for numpy 2.0.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/10221 .

Ran some basic tests. After the PR, XGBoost should handle new numpy inputs correctly. For dependencies like cupy, dask, and cudf, we will wait until numpy 2.0 is officially out.


hcho3:
Thanks for working on it so quickly!

jakirkham:
Thank you both! 🙏

Do we have a rough idea what release this might land in?

trivialfis:
We don't have an eta yet.

PR #10253, opened 2024-05-06
[jvm-packagaes] Freeze spark to 3.4.1 for now.

trivialfis:
The newer spark version for CPU conflicts with the more conservative version used by rapids.

We need JSON support for JVM packages for passing various XGBoost parameters in https://github.com/dmlc/xgboost/pull/10112  . However, different spark versions require different jackson versions. This creates conflicts between packages.


trivialfis:
cc @wbo4958 .

hcho3 (approved):

Issue #10180, opened 2024-04-13
Pandas 2.2: Index.format is deprecated


hcho3:
This line
https://github.com/dmlc/xgboost/blob/732e27cebc0c4085d0d9cc57f278c531cf68a8b0/python-package/xgboost/data.py#L942

produces the following warning with Pandas 2.2:
```
xgboost/data.py:942: FutureWarning: Index.format is deprecated and will be removed in a future version.
Convert using index.astype(str) or index.map(formatter) instead.
```

The deprecation was introduced in https://github.com/pandas-dev/pandas/pull/55439

PR #10266, opened 2024-05-10
Fixes for the latest pandas.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/10180 .


hcho3 (commented):
Is this change compatible with Pandas 1.5 ?

trivialfis:
@hcho3 The used API was there https://pandas.pydata.org/pandas-docs/version/1.5/reference/api/pandas.Index.map.html , but I haven't tested it. Do we need to support this particular version of pandas?

hcho3:
I think a best effort support for past version should be sufficient

hcho3 (approved):

Issue #10105, opened 2024-03-10
Docs for custom objectives could cover more details


david-cortes:
From reading the docs, it's quite unclear what exactly is or isn't supported as custom objectives in XGBoost.

For example, one might wonder after reading the docs:
* Does it support convex objectives only, or can it also solve non-convex (with potentially negative Hessian) objectives? From some experiments, it looks like it clips negative Hessians to zero but would be ideal if the docs could mention the logic.
* Does it work with Fisher's method instead of Newton's? (It should, the more so since it clips Hessians) I think it could mention it as an alternative, especially if the objective is non-convex.
* What is or isn't supported for multi-output objectives? I think it could have a big and clear disclaimer that the methods aren't meant for vector-valued objectives with dependencies between targets (e.g. multinomial logistic).

trivialfis:
> Does it support convex objectives only, or can it also solve non-convex

No, it doesn't. We kind of workaround it by clipping it, but there's no proof that it can converge. So, yes, XGBoost can run, whether the result is useful is a different problem.

> Does it work with Fisher's method instead of Newton's?

Yes, you can use the expected Hessian. Again, I don't have a proof for it to work, but experiments with forecasting seem promising result.

> What is or isn't supported for multi-output objectives

As long as you can come up with an approximation of the Hessian that's 0 or 1 dim per-sample (<= 2-dim for the whole dataset), it should be supported. For instance, the softmax cross entropy uses the diagonal of the Hessian.

trivialfis:
I can work on these things later. Thank you for raising the issue.

david-cortes:
> For instance, the softmax cross entropy uses the diagonal of the Hessian.

But is that the best approach for multinomial logistic? Would such a procedure be guaranteed to converge for general optimization?

Take for example GLMNET: in its multinomial variant in which it solves for one class at a time, it recalculates the gradients and hessians after every single-class newton iteration, instead of after a full cycle over all classes like XGBoost does:
https://www.jstatsoft.org/article/view/v033i01

And in its equivalent of multiple classes solved at the same time, it uses a diagonal upper bound of the hessian instead of the true diagonal:
https://arxiv.org/pdf/1311.6529.pdf

Did some experiments and the diagonal Hessian seems to indeed perform substantially better than their upper bound, but don't know if it has the same theoretical properties around convergence.

[...]

PR #10283, opened 2024-05-13
Add more detailed explanations and tutorial for advanced objectives

david-cortes:
This PR expands the docs for custom objectives to cover some important details that are missing about what is or isn't supported.

It also adds a new example guide for advanced usage in which it uses Dirichlet regression as a custom objective (which doesn't meet the requirements of custom objectives for the Hessian), explaining the workarounds that can be used for adapting a model like this as explained in [an earlier discussion](https://github.com/dmlc/xgboost/issues/10105) so that it'd fit into xgboost's framework.

**Note:** I'm not 100% sure that the .rst file will render correctly. Would be ideal if someone could take a look before merging.

**Note2:** the tutorial copy-pasted some data from a publication, which was in turn taken from an R package. I added a citation there but I'm not sure if it requires some other additional form of attribution.


trivialfis:
> Would be ideal if someone could take a look before merging.

You can find the "docs/readthedocks.org:xgboost" entry in the CI test list.

david-cortes:
> > Would be ideal if someone could take a look before merging.
> 
> You can find the "docs/readthedocs.org:xgboost" entry in the CI test list.

Thanks, I just checked and it looks correctly rendered.

PR is ready for review.

trivialfis (commented):
Thank you for this didactic document! I will try to go through it in the coming days and hope you don't mind if I take my time to understand everything in the doc.

david-cortes:
Added also R code for all of the examples, but running them will require merging this PR first : https://github.com/dmlc/xgboost/pull/10323

trivialfis:
Apologies for the slow reply lately, will get to this ASAP.

doc/tutorials/advanced_custom_obj.rst
```diff
@@ -0,0 +1,723 @@
...
+        return(diag.bound.Ehess)
+    }
+
+(*note: the calculation can be made more efficiently than what is shown here
+by not calculating the full matrix, and in R, by making the rows be the last
+dimension and transposing after the fact*)
+
+With all these pieces in place, one can now frame this model into the format
+required for XGBoost's custom objectives - note that in the Python interface,
+even though the result is 2D, it must be returned as a 1D array where the
```
trivialfis, line 409: This is no longer true after 2.1. Python can handle matrix inputs.

trivialfis (commented):
Thank you for the tutorial. I haven't been able to calculate the hessian myself, but I don't want to be the bottleneck here for any longer. The overall concept looks good to me. One small correction in the comments.

trivialfis:
Will merge once the R test can pass.

david-cortes:
> Will merge once the R test can pass.

The failing R test is about the roxygen note in the `DESCRIPTION` file:
https://github.com/dmlc/xgboost/blob/00264eb72bfe1d5dfc8df00563eaa1c08e6dc15d/R-package/DESCRIPTION#L69

It won't pass because the CI is using version 7.3.2, while the master branch says 7.3.1.

trivialfis:
Thanks for sharing; I will update it on Monday once I can access my desktop.

PR #10287, opened 2024-05-14
Support doc link for the sklearn module.

trivialfis:
See https://scikit-learn.org/stable/developers/develop.html#developer-api-for-html-representation for more info.

It's a small enhancement for having a link to XGBoost's documentation site when an object is rendered in a jupyter notebook, see the upper right corner of the object description in the figure below.

Screenshot from 2024-08-01 23-28-19

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -795,6 +795,26 @@ class XGBModel(XGBModelBase):
     def __sklearn_is_fitted__(self) -> bool:
         return hasattr(self, "_Booster")
 
+    @property
+    def _doc_link_module(self) -> str:
+        return "xgboost"
+
+    @property
+    def _doc_link_template(self) -> str:
+        from .core import _py_version
+
+        ver = _py_version()
+        rel = "latest" if ver.endswith("-dev") else "stable"
+        module = self.__class__.__module__
+        # All sklearn estimators are forwarded to the top level module in both source
+        # code and sphinx api doc.
+        if module == "xgboost.sklearn":
+            module = module.split(".")[0]
+        name = self.__class__.__name__
+
+        base = "https://xgboost.readthedocs.io/en"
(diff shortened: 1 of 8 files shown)


trivialfis:
hmm, we might need a newer Python for this.

hcho3:
> hmm, we might need a newer Python for this.

How new do we need it?

tests/ci_build/conda_env/linux_cpu_test.yml
```diff
@@ -14,7 +14,7 @@ dependencies:
 - pylint
 - numpy
 - scipy
-- scikit-learn
+- scikit-learn>=1.4.1
```
hcho3, line 17: Should we also add `scikit-learn>=1.4.1` to `pyproject.toml`?

python-package/xgboost/sklearn.py
```diff
@@ -795,6 +797,32 @@ def _more_tags(self) -> Dict[str, bool]:
...
+    def _doc_link_template(self) -> str:
+        ver = _py_version()
+        (major, minor, _), post = _parse_version(ver)
+
+        if post == "dev":
+            rel = "latest"
+        else:
+            # RTD tracks the release branch, we don't have different branch patch
+            # release.
+            rel = f"{major}.{minor}.0"
```
hcho3, line 814:
```suggestion
            rel = f"release_{major}.{minor}.0"
```

hcho3 (changes requested):

tests/ci_build/conda_env/linux_cpu_test.yml
trivialfis: No need, for older versions, the code has no effect.

python-package/xgboost/sklearn.py

trivialfis:
Hmm, after the removal of the deterministic repartitioning script, the dask test is quite flaky. Will try to fix it in a different PR.

hcho3 (approved):

PR #10288, opened 2024-05-14
Use a thread pool for external memory.

trivialfis:
Currently, XGBoost uses `std::async` for launching data pre-fetch threads. This approach is convenient but chokes profilers like nsys as it launches a new thread for every batch of data. This PR implements a simple thread pool to use a fixed number of threads.


trivialfis:
@rongou Could you please help take a look into this PR? Other than external memory, I think the thread pool might be able to help significantly reduce the complexity of the collective event loop.

rongou (approved):

PR #10306, opened 2024-05-21
[coll] Keep the tracker alive during initialization error.

trivialfis:
This prevents the tracker from throwing additional errors, which breaks the assumption in the jvm package that errors are expected to be thrown in the executor.


trivialfis:
cc @rongou .

hcho3 (approved):

PR #10282, opened 2024-05-13
python_tests: Upgrade from macos-11 to macos-13

cclauss:
`macos-13` is the last GitHub Actions image for Intel x64 because `macos-14` is Apple Silicon ARM64.

GitHub Actions blog:
> Over the next 12 weeks, jobs using the `macos-latest` runner label will migrate from macOS 12 (Monterey) to macOS 14 (Sonoma).

https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image

PR #10320, opened 2024-05-23
[CI] Upgrade github workflows to use latest Conda setup action

hcho3:
Closes #10282


trivialfis:
I kept the sanitizer thread here for test. Feel free to remove it if you don't think that's necessary.

PR #10325, opened 2024-05-24
SYCL. Optimize gradients calculations.

razdoburdin:
Hi, this PR make optimize gradients calculations on sycl devices by removing some unnecessary data copying and memory reallocations.


trivialfis (commented):
In the longer term, do you think it's possible to refactor the code to make it resemble the underlying math expression? Maybe some wrapper functions like this: https://github.com/dmlc/xgboost/blob/949f0622295613c0e8020d9a0f166b41aedc737f/src/objective/quantile_obj.cu#L83 ? I hope we can keep the objective relatively simple so that more users can understand how an objective is defined.

plugin/sycl/objective/multiclass_obj.cc
```diff
@@ -66,51 +84,62 @@ class SoftmaxMultiClassObj : public ObjFunction {
...
-    ::sycl::buffer<bst_float, 1> weights_buf(is_null_weight ? NULL : info.weights_.HostPointer(),
-                                             is_null_weight ? 1 : info.weights_.Size());
+    bst_float* preds_ptr = preds_.Data();
+    bst_float* labels_ptr = labels_.Data();
+    bst_float* weights_ptr = weights_.Data();
+    GradientPair* out_gpair_ptr = out_gpair_.Data();
+
+    int label_correctness_flag = 1;
+    const size_t wg_size = 32;
+    const size_t nBatch = ndata / kBatchSize + (ndata % kBatchSize > 0);
```
trivialfis, line 94: Is it possible that we can have a separate function or a set of helpers for calculating batches that are independent of the objective functions? It would be great if we could start the objective function with a simpler structure and avoid things that split out things that are not strictly related to the math itself (like batch, memcpy, etc).

razdoburdin: Hi, I have introduced the `BatchProcessingHelper `class. It conceals most of the details regarding batch processing implementation.

trivialfis (approved):

PR #10337, opened 2024-05-28
[R] Rename `BIAS` -> `(Intercept)`

david-cortes:

This PR renames the automatically added feature "BIAS" to "(Intercept)" which is what R uses as the name for biases / intercepts.


trivialfis (approved):

mayer79:
Joining the party (too) late: This change will break all CRAN packages that use XGBoost with SHAP.

PR #10350, opened 2024-05-30
Fix typo & format in C API documentation

astariul:
Fix typo : `Univeral` -> `Universal`

Also, the format of the Doxygen was wrong : the list `format` is supposed to be displayed within the `config` argument, but in the documentation, it's displayed outside : 

Screenshot 2024-05-30 at 10 51 31 PM

---

Disclaimer : I couldn't build the docs locally, I couldn't test my changes.


trivialfis (approved):
Thank you for the fix, you can find the build result in the list of CI tests `docs/readthedocs.org:xgboost`

PR #10351, opened 2024-05-30
[CI] Add nightly CI job to test against dev version of deps

### LINKED ISSUE #10206: [CI] Set up a nightly pipeline to test with dev versions of RAPIDS
Currently, XGBoost is tested with latest stable version of RAPIDS. As a result, the latest XGBoost may not be fully compatible with an upcoming version of RAPIDS. By the time RAPIDS releases a new version, it may end up packaging a version of XGBoost that's not fully compatible with the new RAPIDS version. See https://github.com/dmlc/xgboost/pull/10175 for an example.

We should set up a CI pipeline that tests XGBoost with development version of RAPIDS on a nightly basis.

hcho3:
Closes #10206 

Sample run: https://buildkite.com/xgboost/xgboost-ci-nightly/builds/9
(Temporarily enabled for this PR, but the CI pipeline would normally run on a nightly basis)


tests/ci_build/Dockerfile.gpu_dev_ver
```diff
@@ -0,0 +1,51 @@
...
+
+ENV PATH=/opt/mambaforge/bin:$PATH
+
+# Create new Conda environment with cuDF, Dask, and cuPy
+RUN \
+    export NCCL_SHORT_VER=$(echo "$NCCL_VERSION_ARG" | cut -d "-" -f 1) && \
+    mamba create -y -n gpu_test -c rapidsai-nightly -c nvidia -c conda-forge \
+        python=3.10 "cudf>$RAPIDS_VERSION_ARG" "rmm>$RAPIDS_VERSION_ARG" cudatoolkit=$CUDA_VERSION_ARG \
+        "nccl>=${NCCL_SHORT_VER}" \
+        dask=2024.1.1 \
```
hcho3, line 30: Not sure when we can release this pin on `dask`.

trivialfis (approved):

hcho3:
I'll try to solicit one more review from the build team.

tests/ci_build/Dockerfile.gpu_dev_ver
```diff
@@ -0,0 +1,51 @@
...
+ENV DEBIAN_FRONTEND noninteractive
+SHELL ["/bin/bash", "-c"]   # Use Bash as shell
+
+# Install all basic requirements
+RUN \
+    apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/3bf863cc.pub && \
+    apt-get update && \
+    apt-get install -y wget unzip bzip2 libgomp1 build-essential openjdk-8-jdk-headless && \
+    # Python
+    wget -nv -O conda.sh https://github.com/conda-forge/miniforge/releases/download/22.11.1-2/Mambaforge-22.11.1-2-Linux-x86_64.sh && \
```
bdice, line 19: Can you migrate to miniforge and use a newer version? (mambaforge isn't needed anymore, as miniforge contains mamba.)

tests/ci_build/Dockerfile.gpu_dev_ver
bdice: I think the dask version would be tied to the RAPIDS release version.

tests/ci_build/Dockerfile.gpu_dev_ver
```diff
@@ -0,0 +1,51 @@
...
+    bash conda.sh -b -p /opt/mambaforge
+
+ENV PATH=/opt/mambaforge/bin:$PATH
+
+# Create new Conda environment with cuDF, Dask, and cuPy
+RUN \
+    export NCCL_SHORT_VER=$(echo "$NCCL_VERSION_ARG" | cut -d "-" -f 1) && \
+    mamba create -y -n gpu_test -c rapidsai-nightly -c nvidia -c conda-forge \
+        python=3.10 "cudf>$RAPIDS_VERSION_ARG" "rmm>$RAPIDS_VERSION_ARG" cudatoolkit=$CUDA_VERSION_ARG \
```
bdice, line 28: RAPIDS needs a different channel order. Also, why are we using `>` here instead of `=`?
```suggestion
    mamba create -y -n gpu_test -c rapidsai-nightly -c conda-forge -c nvidia \
        python=3.10 "cudf=$RAPIDS_VERSION_ARG" "rmm=$RAPIDS_VERSION_ARG" cudatoolkit=$CUDA_VERSION_ARG \
```

tests/ci_build/Dockerfile.gpu_dev_ver
```diff
@@ -0,0 +1,51 @@
...
+        "nccl>=${NCCL_SHORT_VER}" \
+        dask=2024.1.1 \
+        "dask-cuda>$RAPIDS_VERSION_ARG" "dask-cudf>$RAPIDS_VERSION_ARG" cupy \
```
bdice, line 31:
```suggestion
        "dask-cuda=$RAPIDS_VERSION_ARG" "dask-cudf=$RAPIDS_VERSION_ARG" cupy \
```

tests/buildkite/pipeline-nightly.yml
```diff
@@ -0,0 +1,37 @@
...
+  DISABLE_RELEASE: "1"
+    # Skip uploading artifacts to S3 bucket
+    # Also, don't build all CUDA archs; just build sm_75
+  USE_DEPS_DEV_VER: "1"
+    # Use dev versions of RAPIDS and other dependencies
+steps:
+  #### -------- CONTAINER BUILD --------
+  - label: ":docker: Build containers"
+    commands:
+      - "tests/buildkite/build-containers.sh gpu_build_centos7"
```
bdice, line 15: RAPIDS no longer supports CentOS 7. If this command requires RAPIDS, it will need to use something with glibc >=2.28 (Rocky Linux 8 has 2.28, Ubuntu 20.04 has 2.31).

tests/buildkite/build-containers.sh
```diff
@@ -20,7 +20,7 @@ case "${container}" in
   cpu)
     ;;
 
-  gpu)
+  gpu | gpu_dev_ver)
```
jameslamb, line 23: It looks to me like this and `gpu_build_centos7` are identical... could those cases be consolidated to just `gpu*`?

(pending the outcome of this other discussion: https://github.com/dmlc/xgboost/pull/10351#discussion_r1622403527)

tests/ci_build/Dockerfile.gpu_dev_ver
jameslamb: +1, I think and I think you're already going to get the RAPIDS pins via having `dask-cuda` in this environment, since it depends on `rapids-dask-dependency`.

https://github.com/rapidsai/dask-cuda/blob/f1a14050ec9827c9401c6ca90617b16c358029dd/pyproject.toml#L23

https://github.com/rapidsai/rapids-dask-dependency/blob/e2938ebb44d9d28a0fc5b8c5ebcc29a57bd5b3ea/pyproject.toml#L14

I recommend removing `dask` from this list of requirements and just taking whatever versions get pulled in via `dask-cuda` and `dask-cudf`.

tests/ci_build/Dockerfile.gpu_dev_ver
```diff
@@ -0,0 +1,51 @@
...
+        numpy pytest pytest-timeout scipy scikit-learn pandas matplotlib wheel python-kubernetes urllib3 graphviz hypothesis \
+        "pyspark>=3.4.0" cloudpickle cuda-python && \
+    mamba clean --all && \
```
jameslamb, line 34:
```suggestion
    mamba clean --all --yes && \
```

I think this needs a `--yes` to prevent an interactive prompt, right? I haven't been able to find the logs for a run of this new job, not sure if that's caused a problem.

jameslamb (commented):
Thanks for working on this! I put up a couple of suggestions (and agree with all of @bdice 's suggestions).

tests/buildkite/pipeline-nightly.yml
hcho3: This container is solely used for building XGBoost for manylinux2014 standard. Should we migrate XGBoost to glibc >= 2.28 too?

tests/ci_build/Dockerfile.gpu_dev_ver
hcho3: I'm using `>` here because the variable `$RAPIDS_VERSION_ARG` indicates the latest stable version. https://github.com/dmlc/xgboost/blob/d2d01d977a5ba07e55ed9230e2bd83539e81d184/tests/buildkite/conftest.sh#L27

hcho3: Let me add a comment here.

tests/buildkite/pipeline-nightly.yml
hcho3: Opened https://github.com/dmlc/xgboost/issues/10357. Since CentOS 7 will reach EOL soon, we should move to rockylinux8.

tests/ci_build/Dockerfile.gpu_dev_ver
hcho3: Removing `dask` from the list did the trick. Thanks!

tests/ci_build/Dockerfile.gpu_dev_ver
bdice: You probably want a solution that will use the _nightly_ version like 24.06 rather than the _previous stable_ version like 24.04.

If you do that, you can use `=X.Y.*`. That's how we handle it in RAPIDS:

```
- librmm==24.8.*
```

https://github.com/rapidsai/cudf/blob/1354abdb7a4f9eb58bfc6e359c49d0baabacb4e1/conda/environments/all_cuda-122_arch-x86_64.yaml#L49

tests/buildkite/update-rapids.sh
```diff
@@ -4,7 +4,10 @@ set -euo pipefail
 
 LATEST_RAPIDS_VERSION=$(gh api repos/rapidsai/cuml/releases/latest --jq '.name' | sed -e 's/^v\([[:digit:]]\+\.[[:digit:]]\+\).*/\1/')
 echo "LATEST_RAPIDS_VERSION = $LATEST_RAPIDS_VERSION"
+DEV_RAPIDS_VERSION=$(date +%Y-%m-%d -d "20${LATEST_RAPIDS_VERSION//./-}-01 + 2 month" | cut -c3-7 | tr - .)
+echo "DEV_RAPIDS_VERSION = $DEV_RAPIDS_VERSION"
 
 PARENT_PATH=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 
 sed -i "s/^RAPIDS_VERSION=[[:digit:]]\+\.[[:digit:]]\+/RAPIDS_VERSION=${LATEST_RAPIDS_VERSION}/" $PARENT_PATH/conftest.sh
+sed -i "s/^DEV_RAPIDS_VERSION=[[:digit:]]\+\.[[:digit:]]\+/DEV_RAPIDS_VERSION=${DEV_RAPIDS_VERSION}/" $PARENT_PATH/conftest.sh
```
hcho3, line 13: Note. This script runs periodically to update `RAPIDS_VERSION` and `DEV_RAPIDS_VERSION` variables in `conftest.sh`.

tests/ci_build/Dockerfile.gpu_dev_ver
hcho3: @bdice Didn't realize that there might be two nightly versions. I updated the PR to specify the nightly version exactly.

hcho3:
@trivialfis I remember you pinning `dask` to 2024.1.1 in #10125. Did you have a change to find out why the latest version fails with `CancelledError` ?

trivialfis:
Unfortunately, no. Debugging dask issue is above my head.

hcho3:
In that case, let's add the pin `dask=2024.1.1` to the requirements of the pip wheel and Conda package.

trivialfis:
I don't think XGBoost is causing the error and there's no compatibility issue here. We can pin dask on the CI, but I don't think it is necessary to pin it in the XGBoost release package.

hcho3:
I worry about the latest dask breaking XGBoost code. Do you think if the error is limited to a specific test only, or is it potentially affecting other parts of XGBoost?

trivialfis:
I don't know, we had a sync with the dask team previously. Apparently it's not easy to debug. I will follow-up on this one and see where we are.

PR #10389, opened 2024-06-04
Partial fixups for federated tests

crockeea:
When I tried to run federated learning tests following the directions in https://github.com/dmlc/xgboost/blob/master/plugin/federated/README.md, I ran into several problems.

1. The path to the tests is wrong in the README; that's fixed in this PR.
2. All calls to `run_federated_server` swap the order of the first two arguments. I decided to change the order of the parameters rather than change each call to the function. 
3. `run_federated_server` doesn't supply the key/certificate paths when creating a `FederatedTracker`.

Note that I still cannot get the tests to work even with these changes; now they fail without a helpful error:
```
[21:45:11] Insecure federated server listening on 0.0.0.0:9091, world size 2
  File "/home/ubuntu/xgboost/tests/test_distributed/test_federated/test_federated.py", line 84, in <module>
    run_federated(with_ssl=False, with_gpu=False)
  File "/home/ubuntu/xgboost/tests/test_distributed/test_federated/test_federated.py", line 69, in run_federated
    raise Exception("Error starting Federated Learning server")
Exception: Error starting Federated Learning server
```

### DIFF
--- a/plugin/federated/README.md
+++ b/plugin/federated/README.md
@@ -10,7 +10,7 @@ Test Federated XGBoost
 ----------------------
 ```shell
 # Under xgboost source tree.
-cd tests/distributed/test_federated
+cd tests/test_distributed/test_federated
 # This tests both CPU training (`hist`) and GPU training (`gpu_hist`).
 ./runtests-federated.sh
 ```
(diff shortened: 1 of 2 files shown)


trivialfis:
Apologies for the slow response here. We are still working on it, and things are in flux. Will take a deeper look into these and push some documents after https://github.com/dmlc/xgboost/issues/9987 is done.

PR #10397, opened 2024-06-06
[coll] Small cleanup to federated comm.

trivialfis:
- Remove the extra ctor.
- Use the default timeout in coll.


trivialfis:
cc @rongou .

rongou (approved):

Issue #10379, opened 2024-06-03
Compatibility with latest Dask


hcho3:
Currently, some of the unit tests fail with latest Dask (https://github.com/dmlc/xgboost/pull/10125).

For now we've pinned the Dask version to `2024.1.1`. We should find out what's causing the issue with the latest Dask and make a fix in either Dask or XGBoost.

trivialfis:
Specifically, after training and run prediction, the prediction might hang or raise a cancelled error.

crusaderky:
About the CancelledError, see my comment on https://github.com/dask/dask/issues/11179

trivialfis:
Closing since we have removed the pin.

PR #10398, opened 2024-06-06
Avoid default tokenization in Dask

rjzamora:
I spent some time investigating a `concurrent.futures._base.CancelledError`  error in [`test_gpu_with_dask.py::TestDistributedGPU::test_categorical`](https://github.com/dmlc/xgboost/blob/f5815b69826e6a6c479796f52aa50acd1809a32c/tests/test_distributed/test_with_dask/test_with_dask.py#L409) and eventually bisected the cause to https://github.com/dask/dask/pull/10883

That PR changed how `Booster` objects are tokenized by dask, and effectively moved from using a random `uuid`-based approach to a deterministic hash. I haven't had a chance to figure out exactly why the new deterministic hash causes problems, but this PR opts out of the new approach for now.

cc @trivialfis 

Addresses part of https://github.com/dmlc/xgboost/issues/10379

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -7,6 +7,7 @@ import json
 import os
 import re
 import sys
+import uuid
 import warnings
 import weakref
 from abc import ABC, abstractmethod
@@ -3143,3 +3144,10 @@ class Booster:
                 UserWarning,
             )
         return nph_stacked
+
+    def __dask_tokenize__(self):
+        # TODO: Implement proper tokenization to avoid
+        # unnecessary re-computation in Dask. However,
+        # default tokenzation causes problems after
+        # https://github.com/dask/dask/pull/10883
+        return uuid.uuid4()
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -3145,9 +3145,8 @@ class Booster:
             )
         return nph_stacked
 
-    def __dask_tokenize__(self):
-        # TODO: Implement proper tokenization to avoid
-        # unnecessary re-computation in Dask. However,
-        # default tokenzation causes problems after
+    def __dask_tokenize__(self) -> uuid.UUID:
+        # TODO: Implement proper tokenization to avoid unnecessary re-computation in
+        # Dask. However, default tokenzation causes problems after
         # https://github.com/dask/dask/pull/10883
         return uuid.uuid4()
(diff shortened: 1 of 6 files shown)


tests/test_distributed/test_gpu_with_dask/test_gpu_with_dask.py
rjzamora, line 1: All changes in this file are related to the recent deprecation of `dask_cudf.from_dask_dataframe`.

trivialfis:
Thank you for the fix! We recommend users to scatter the booster before running prediction, this way we can reuse the same booster already on the workers for multiple prediction instead of transferring it repeatedly. Does the PR change that?

rjzamora:
> We recommend users to scatter the booster before running prediction, this way we can reuse the same booster already on the workers for multiple prediction calls instead of transferring it repeatedly. Does the PR change that?

This PR *should* produce the same behavior as `dask<2024.2.1`. You should still be able to scatter the model ahead of time and use the futures for multiple prediction calls. However, if you scatter the same model twice, dask won't recognize that the model hasn't changed, and so the same model will need to be transferred again (with different key names).

I'm not entirely sure what is causing the "cancelled" error for `dask>=2024.2.1`, but it seems likely that dask's attempt to tokenize `Booster` objects deterministically is a bit "off". It seems like different/retrained models can be incorrectly hashed to the same token. For example, if a model is retrained, you would want dask to use new keys to represent the corresponding futures. In `dask>=2024.2.1`, this does not seem to be the case (In `test_empty_partition`, I added a break point and found that both `bst["booster"]` and `bst_empty["booster"]` tokenize to the same value).

It may also be the case that deterministic tokenization is working "fine" for `Booster` objects, but xgboost happens to be using dask in a way that only works if a newly scattered model is always treated as unique.

trivialfis:
Thank you for the detailed explanation!

python-package/xgboost/core.py
```diff
@@ -3143,3 +3144,9 @@ def get_split_value_histogram(
                 UserWarning,
             )
         return nph_stacked
+
+    def __dask_tokenize__(self) -> uuid.UUID:
+        # TODO: Implement proper tokenization to avoid unnecessary re-computation in
+        # Dask. However, default tokenzation causes problems after
+        # https://github.com/dask/dask/pull/10883
+        return uuid.uuid4()
```
crusaderky, line 3152: Two consecutive calls to tokenize() will return two different values, and this can cause problems. I would advise caching the output: https://github.com/dask/dask/issues/11179#issuecomment-2168094529

rjzamora: Thanks @crusaderky ! The purpose of this PR was to effectively roll back behavior to "match" tokenization before 10883. However, I agree that it makes sense to cache the value.

Issue #10370, opened 2024-06-02
Update the default CTK to 12.4.


trivialfis:
It should be able to work on all CTK 12.x versions. However, do know that once the default CTK is upgraded to 12, xgboost won't work on CTK 11.x.

hcho3:
Latest gRPC is not compatible with CUDA 11.8 (https://github.com/tensorflow/tensorflow/issues/63356). 

TODO: Upgrade gRPC while upgrading CUDA to 12.4.

trivialfis:
@hcho3 How difficult do you think it would be to upgrade the CTK version? Is it possible to upgrade without solving https://github.com/dmlc/xgboost/issues/10400 first?

PR #10399, opened 2024-06-07
[CI] Migrate to rockylinux8 / manylinux_2_28_x86_64

### LINKED ISSUE #10357: [CI] Migration to manylinux_2_28_x86_64 and rockylinux8
XGBoost uses manylinux2014 to package Python wheels. The manylinux2014 standard requires the use of CentOS 7, and CentOS 7 will reach its end-of-life on June 30, 2024.

Fix: We should use the `rockylinux8` image when building the Python wheel, and tag it with platform tag `manylinux_2_28_x86_64`

hcho3:
Closes #10357

Also:
* Use `gcc-toolset-*` instead of `devtoolset-*`. For now, use GCC 10 to support CUDA 11.8. We will migrate to CUDA 12.x in the near future (#10370).
* Use latest version of CMake and Maven.
* Remove the unused image `s390x`.


tests/ci_build/build_via_cmake.sh
```diff
@@ -24,7 +24,9 @@ fi
 rm -rf build
 mkdir build
 cd build
-cmake .. ${cmake_args} -DGOOGLE_TEST=ON -DUSE_DMLC_GTEST=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DENABLE_ALL_WARNINGS=ON -DCMAKE_COMPILE_WARNING_AS_ERROR=ON -GNinja ${cmake_prefix_flag} -DHIDE_CXX_SYMBOLS=ON -DBUILD_DEPRECATED_CLI=ON
+# Disable CMAKE_COMPILE_WARNING_AS_ERROR option temporarily until
+# https://github.com/dmlc/xgboost/issues/10400 is fixed
+cmake .. ${cmake_args} -DGOOGLE_TEST=ON -DUSE_DMLC_GTEST=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DENABLE_ALL_WARNINGS=ON -DCMAKE_COMPILE_WARNING_AS_ERROR=OFF -GNinja ${cmake_prefix_flag} -DHIDE_CXX_SYMBOLS=ON -DBUILD_DEPRECATED_CLI=ON
```
hcho3, line 29: See https://github.com/dmlc/xgboost/issues/10400

hcho3:
It looks like we have to upgrade to CUDA 12.x to build latest gRPC: https://github.com/tensorflow/tensorflow/issues/63356

tests/buildkite/build-cuda.sh
```diff
@@ -40,7 +40,7 @@ $command_wrapper python tests/ci_build/rename_whl.py  \
   --commit-hash ${BUILDKITE_COMMIT}  \
   --platform-tag ${WHEEL_TAG}
 
-echo "--- Audit binary wheel to ensure it's compliant with manylinux2014 standard"
+echo "--- Audit binary wheel to ensure it's compliant with manylinux_2_28 standard"
```
jameslamb, line 43:
```suggestion
echo "--- Audit binary wheel to ensure it's compliant with ${WHEEL_TAG} standard"
```

tests/ci_build/Dockerfile.jvm_cross
```diff
@@ -1,5 +1,5 @@
-FROM ubuntu:18.04
-ARG JDK_VERSION=8
+FROM ubuntu:22.04
+ARG JDK_VERSION=11
```
jameslamb, line 2: I think this should probably also be in the PR title (assuming that becomes a changelog note). That might be a breaking change for Java users, like anyone using XGBoost4J, right?

jameslamb (approved):
Left some cosmetic recommendations, but overall the structure makes sense to me!

If it helps with the decision... in LightGBM we've been using `manylinux_2_28` for the Python package since November 2022 (https://github.com/microsoft/LightGBM/pull/5580) and I don't recall getting any reports about it being a problem.

tests/ci_build/Dockerfile.jvm_cross
hcho3: Changed it back to `8`, to be consistent with
https://github.com/dmlc/xgboost/blob/c9f5fcaf21ea25f78538b672cc60004fce9f524d/tests/buildkite/conftest.sh#L30

hcho3:
Getting a weird link error:
```
/opt/rh/gcc-toolset-12/root/usr/bin/ld: /workspace/xgboost: hidden symbol `_ZNSt10filesystem9_Dir_base7advanceEbRSt10error_code' isn't defined
/opt/rh/gcc-toolset-12/root/usr/bin/ld: final link failed: bad value
```
https://buildkite.com/xgboost/xgboost-ci/builds/5501#018ff4b1-a587-4032-bb80-8b7fadcafa79/132-657

According to https://bugzilla.redhat.com/show_bug.cgi?id=1929043, upgrading to `gcc-toolset-10` should fix the issue.

trivialfis:
I can't reproduce the failure on Windows.

hcho3:
@trivialfis What do you think? Should we consider releasing 2.1.0 with `manylinux_2_28` standard?
Doing so would ensure that XGBoost doesn't depend on core libs that will soon be outdated. LightGBM has been using `manylinux_2_28` for a while now, so the risk of breaking things is likely low.

trivialfis (approved):
Thanks for managing the binary build! Please help backport it to 2.1.

Issue #9471, opened 2023-08-11
Deprecate the command line interface.


trivialfis:
Currently, the CLI  is rarely used as the vast majority of the users want access to the ecosystem of other language bindings. Even basic things like cross-validation are quite difficult to perform using the CLI.

For inference, there are low-latency options like inplace_predict with proper C API, not to mention third-party libraries like FIL that focus on model deployment.

I propose that we move on and deprecate the CLI in order to focus on more promising features.

saxena-vinayak36:
can anyone tell me how to merge my pr with this?

saxena-vinayak36:
I think I can contribute to this.

trivialfis:
H @saxena-vinayak36 i, this issue has already been closed.

PR #10403, opened 2024-06-08
[WIP] Diagnose failing CLI tests on Windows

hcho3:
https://github.com/dmlc/xgboost/actions/runs/9425984805/job/25968341700


trivialfis:
Considering that the CLI is being deprecated https://github.com/dmlc/xgboost/issues/9471, maybe we can drop the Windows test. But the JVM test is a bit more tricky.

PR #10404, opened 2024-06-10
[CI] Fix JVM tests on Windows

### LINKED ISSUE #10344: Use CMAKE_MSVC_RUNTIME_LIBRARY to consistently handle linking mode for MSVC runtime
On the Windows platform, the MSVC (Microsoft Visual C++) runtime can be linked either statically or dynamically. Currently, XGBoost makes this choice by manually editing raw build flags: https://github.com/dmlc/xgboost/blob/bc6c993aaa68b3f67849a28dfe0218e951f1f42d/cmake/Utils.cmake#L29-L51 This approach is brittle and prone to weird breakage, especially when XGBoost is used by a downstream project.

A more robust method is to use the built-in flag `CMAKE_MSVC_RUNTIME_LIBRARY` to control the linking mode for the MSVC runtime. See https://github.com/dmlc/treelite/blob/260a5d2c22830ec2bdaa5fc6dc186aa6f3b551e7/CMakeLists.txt#L46-L54 for an example.

hcho3:
https://github.com/dmlc/xgboost/actions/runs/9470870529/job/26092755865

Closes #10344


hcho3:
Update: The `xgboost4j.dll` binary built by the CI pipeline is corrupted. I was able to reproduce the same error on my Windows machines.

Steps taken:
1. Download `xgboost4j.dll` from https://s3-us-west-2.amazonaws.com/xgboost-nightly-builds/fix_win/libxgboost4j/xgboost4j_6384bce6aae53b791187fd0db21e574a7a6c1100.dll
2. Rename the file to `xgboost4j.dll` and put it under `jvm-packages/xgboost4j/src/main/resources/lib/windows/x86_64`.
3. Open the Powershell and run:
```powershell
$env:MAVEN_SKIP_NATIVE_BUILD=1
mvn test -B -pl :xgboost4j_2.12
```

hcho3:
Darn, got the same error on BuildKite :(

https://buildkite.com/xgboost/xgboost-ci-windows/builds/5550#01901269-c3f8-4792-8425-19dc8d6f0bb1

hcho3:
@Craigacp Can you help? I'm at my wit's end because I can't figure out why the JVM fails to load `xgboost4j.dll` with DLL initialization error. Have you seen this error before? I see this error both on GitHub Actions and a self-hosted CI server.

Craigacp:
I can take a look. Is the dll you linked [here](https://github.com/dmlc/xgboost/pull/10404#issuecomment-2159319569) the one to test out?

hcho3:
@Craigacp Yes, that would be perfect. Thanks for your help !

hcho3:
I did some digging into the generated DLL and here's my finding: the JVM (`java.exe`) and the Python interpreter (`python.exe`) are loading incompatible versions of core libraries into memory at the same time, leading to the initialization error for `xgboost4j.dll`.

Currently, `xgboost4j.dll` depends on the following core libraries:
* kernel32.dll
* **msvcp140.dll**
* vcomp140.dll
* ws2_32.dll
* concrt140.dll
* **vcruntime140.dll**
* **vcruntime140_1.dll**

(Obtained via `dumpbin /dependents xgboost4.dll`)

At startup, `java.exe` and `python.exe` load different versions of **msvcp140.dll**, **vcruntime140.dll**, and **vcruntime140_1.dll**.

| Core lib | `java.exe` | `python.exe` (Miniconda) | `python.exe` (Vanilla Python from python.org) |
|----|----|----|----|
| msvcp140.dll | 14.36.32532.0 | 14.40.33810.0 | 14.40.33810.0 |
| vcruntime140.dll | 14.36.32532.0 | 14.40.33810.0 | 14.38.33126.1 |
| vcruntime140_1.dll | 14.36.32532.0 | 14.40.33810.0 | 14.38.33126.1 |

Craigacp:
I think there's a third option, which is have Maven directly execute the CMake build for the JNI binding. Or even have CMake execute the Maven build once it's built the native library (this is how we do Java builds for ONNX Runtime).

hcho3:
> have CMake execute the Maven build once it's built the native library (this is how we do Java builds for ONNX Runtime).

I like this idea. Let me give it a shot.

Craigacp:
The relevant bits are all here (though it runs Gradle rather than Maven, and the Gradle build is forced to look in a specific place for binaries which we might need to copy for Maven). https://github.com/microsoft/onnxruntime/blob/main/cmake/onnxruntime_java.cmake

hcho3:
Given that 2.1.0 release is just around the corner, I am choosing the least disruptive option (Option 2).

hcho3:
Even when I tried running Java and Python separately, I still get the same error (`DLL initialization failed`). We may need to consider excluding Windows binary in the upcoming release.

@trivialfis @Craigacp @jameslamb

hcho3:
As fate has it, the DLL error had **nothing** to do with duplication of core libraries. The DLL error was actually caused by improper mixing of two modes of linking for the C runtime.

Background: On Windows, you can choose to link with the C runtime dynamically (`/MD` build option) or statically (`/MT`). When building an application, it is imperative that all components are built with the same choice of runtime linking. You will get an undefined behavior when some components are built with `/MD` and others with `/MT`.
The current method of choosing between `/MD` and `/MT` was not sufficiently robust (#10344). So when the MSVC was updated, some of the components got the `/MD` flag when the correct default for XGBoost is `/MT`.

Fix. Use a more robust method to enforce the `/MT` option for all components. See #10344.

hcho3:
@trivialfis Looks like the CI is unblocked now. Can you review?

trivialfis (approved):
Huge thanks for the fix!

PR #10408, opened 2024-06-11
[pyspark] Allow to avoid repartition 

wbo4958:
The current xgboost pyspark will not repartition the dataset only when the last operator of input dataset is repartition and the repartitioned number is equal to num_workers. Overall, this looks good. But we should also keep in mind that repartition is really expensive especially for the GPU  case, the repartition probably dominates the most of training time.  

There is a real xgboost case that reading from file directly using spark and then fit the data into xgboost estimator.  So for this kind of case, we can make the partition number equal to num_workers by playing with some spark configurations, and the data partitions should have been well balanced by spark. So I think, it's safe for xgboost to skip the repartition internally  which can really improve the whole xgboost end to end time.

And on the other hand, xgboost already supports the force_repartition, so if user really would like to enable the repartition, they can set force_repartition to true.

### DIFF
--- a/doc/tutorials/spark_estimator.rst
+++ b/doc/tutorials/spark_estimator.rst
@@ -267,7 +267,7 @@ An example submit command is shown below with additional spark configurations an
     --conf spark.task.cpus=1 \
     --conf spark.executor.resource.gpu.amount=1 \
     --conf spark.task.resource.gpu.amount=0.08 \
-    --packages com.nvidia:rapids-4-spark_2.12:23.04.0 \
+    --packages com.nvidia:rapids-4-spark_2.12:24.04.1 \
     --conf spark.plugins=com.nvidia.spark.SQLPlugin \
     --conf spark.sql.execution.arrow.maxRecordsPerBatch=1000000 \
     --archives xgboost_env.tar.gz#environment \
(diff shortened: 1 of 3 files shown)


python-package/xgboost/spark/core.py
```diff
@@ -705,19 +705,19 @@ def _query_plan_contains_valid_repartition(self, dataset: DataFrame) -> bool:
...
-        num_partitions = dataset.rdd.getNumPartitions()
+
         assert dataset._sc._jvm is not None
         query_plan = dataset._sc._jvm.PythonSQLUtils.explainString(
             dataset._jdf.queryExecution(), "extended"
         )
         start = query_plan.index("== Optimized Logical Plan ==")
         start += len("== Optimized Logical Plan ==") + 1
+
+        query_plan[start : start + len("Repartition")] == "Repartition"
```
trivialfis, line 716: Please remove the commented code.

doc/tutorials/spark_estimator.rst
```diff
@@ -276,3 +276,11 @@ An example submit command is shown below with additional spark configurations an
...
 are required. More configuration options can be found in the RAPIDS link above along with
 details on the plugin.
+
+Advanced Usage
+==============
+
+XGBoost needs to repartition to the num_workers to ensure there will be num_workers training
+tasks running at the same time, but repartition is a costly operation. To avoid repartition,
+users can set ``spark.sql.files.maxPartitionNum`` and ``spark.sql.files.minPartitionNum``
+to num_workers.
```
trivialfis, line 286: Could you please elaborate on when a user might want to force reparittion dataset as well?

trivialfis (changes requested):

python-package/xgboost/spark/core.py
wbo4958: Done.

trivialfis (approved):

PR #10412, opened 2024-06-13
Update learning_to_rank.rst

nocluebutalotofit:
After
qid = qid[sorted_idx]
qid is sorted.
Applying the sorted_idx again when passing it to fit() ranker.fit(X, y, qid=qid[sorted_idx])
causes:
xgboost.core.XGBoostError: [15:45:42] C:\b\abs_0fh_d4x2ng\croot\xgboost-split_1713973188995\work\cpp_src\src\data\data.cc:532: Check failed: non_dec: `qid` must be sorted in non-decreasing order along with data.


trivialfis (approved):
Thank you for the fix!

Issue #10415, opened 2024-06-14
[Breaking Change][jvm-package] Rewrite XGBoost JVM Package


wbo4958:
## Reason
 
The XGBoost JVM package is the Java language binding for the XGBoost library. It is supposed to be a lightweight, thin wrapper around XGBoost. However, the current XGBoost JVM implementation is quite heavy-weight. For example, it groups the dataset using RDD for ranking, implements ranking within the XGBoostRegressor, samples the dataset for training and testing, and zips the training and evaluation datasets, duplicates some code usage. All of these additional features make the XGBoost JVM codebase difficult to read and maintain. Additionally, it is missing support for the latest XGBoost parameters and does not properly handle dense/sparse data usage.

## Goal

1. Create a new XGBoostRanker for ranking problem.
2. Remove "grouping" for ranking problem.
3. Remove the trainTestRatio and its implementation
4. Remove "zip" train and eval dataset (add a new Boolean validation column to indicate if the instance is for training or for evaluating)
5. Catch up the latest parameters.
6. Add setter/getter for all parameters.
7. Support XGBoost style parameters when defining the parameters, (Like `final val baseScore = new DoubleParam(this, "base_score", "The initial `)
8.  Support dense/sparse when the input is vector type
9.  Support array input for both CPU and GPU
10.  Support columnar input for CPU ???
11. Avoid repartition if the number of input partittions is equal to num_workers.
12. Fix surefire issue
13. More scalastyle checking.
14. Shade the xgboost4j/xgboost4j-spark into a single jar
15. Shade the xgboost4j-gpu/xgboost4j-spark-gpu into a single jar

trivialfis:
These might be helpful:
https://github.com/dmlc/xgboost/blob/master/doc/contrib/featuremap.rst
https://github.com/dmlc/xgboost/blob/master/doc/contrib/community.rst

PR #10416, opened 2024-06-14
[jvm-packages] Rewrite xgboost jvm packages

wbo4958:
To fix https://github.com/dmlc/xgboost/issues/10415


trivialfis:
Hi, huge thanks for working on this.

Could you please do a small write-up for the new implementation? We haven't heard about the details of it yet. It would be great if you could write an introduction in a tour style with examples, along with potential breaking changes, etc.

trivialfis:
Merging into the jvm branch. We will do the reviews and tests when the branch is ready.

Issue #10371, opened 2024-06-02
Update Python requirement to 3.10.


trivialfis:
This helps us match the base requirement of numpy, along with which some sklearn features also depend on newer Python versions.

PR #10434, opened 2024-06-17
[breaking] Bump Python requirement to 3.10.

trivialfis:
- Bump the Python requirement.
- Fix type hints.
- Use loky to avoid deadlock.
- Workaround cupy-numpy compatibility issue on Windows caused by the `safe` casting rule.
- Simplify the repartitioning logic to avoid dask errors.

Close https://github.com/dmlc/xgboost/issues/10371


hcho3 (approved):

trivialfis:
all modin tests are broken due to updated numpy.

trivialfis:
cupy is running into error as well.

trivialfis:
dask-expr appears to be running into trouble with the latest pandas.

trivialfis:
- Moved the make sparse function into `data.py` to fix pylint error, also the definition is more consistent.
- Removed the deterministic partitioning function in dask test due to dask internal error. This might make the test flaky, we will see.

trivialfis:
Err, immediately fail after removing the partitioning function.

trivialfis:
Couldn't reproduce the hang, will try it on aws later.

trivialfis:
@hcho3 Could you please help take another look?

hcho3 (approved):

PR #10440, opened 2024-06-18
[CI] Stop vendoring libomp.dylib in MacOS Python wheels

hcho3:
Adopt https://github.com/microsoft/LightGBM/pull/6391 to achieve the following goals:

* Stop vendoring libomp in the MacOS Python wheels. Currently, we vendor libomp 11.1.0, a 3-year old version.
* Allow XGBoost to work with libomp from a variety of sources, e.g. Conda or Homebrew.
  - `libxgboost.dylib` no longer has dependency on hardcoded path of `libomp.dylib`. So it can be used in conjunction with any installation of `libomp.dylib`.
  - If a copy of `libomp.dylib` is already present in memory, XGBoost will use that copy. This way, we can avoid myriad problems caused by having multiple incompatible versions of libomp (#7039).

One disadvantage of this approach is that users will now get an `OSError` if OpenMP is not present in the system. This is not too bad, since the error message already directs users to install OpenMP:
https://github.com/dmlc/xgboost/blob/b9e5229ff203d938250802c00700eefccf853381/python-package/xgboost/core.py#L224-L230

Before this commit (note the hard-coded path for libomp.dylib):
```
$ otool -L ./lib/libxgboost.dylib
./lib/libxgboost.dylib:
        @rpath/libxgboost.dylib (compatibility version 0.0.0, current version 0.0.0)
        /opt/homebrew/opt/libomp/lib/libomp.dylib (compatibility version 5.0.0, current version 5.0.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1500.65.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.100.3)
```

After this commit:
```
$ otool -L ./lib/libxgboost.dylib
./lib/libxgboost.dylib:
        @rpath/libxgboost.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libomp.dylib (compatibility version 5.0.0, current version 5.0.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1500.65.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.100.3)
```

@jameslamb Thanks for inspiration!

### DIFF
--- a/.github/workflows/python_wheels.yml
+++ b/.github/workflows/python_wheels.yml
@@ -28,6 +28,10 @@ jobs:
     - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
       with:
         submodules: 'true'
+    - name: Set up homebrew
+      uses: Homebrew/actions/setup-homebrew@68fa6aeb1ccb0596d311f2b34ec74ec21ee68e54
+    - name: Install libomp
+      run: brew install libomp
     - uses: conda-incubator/setup-miniconda@a4260408e20b96e80095f42ff7f1a15b27dd94ca  # v3.0.4
       with:
         miniforge-variant: Mambaforge
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -367,6 +367,13 @@ if(JVM_BINDINGS)
   xgboost_target_defs(xgboost4j)
 endif()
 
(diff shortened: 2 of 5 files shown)


hcho3:
TODO: Update CI to build libxgboost.dylib without bundling libomp.

PR #10447, opened 2024-06-18
Link `CMAKE_DL_LIBS` when `dlopen` is used.

trivialfis:


hcho3:
Shouldn't it be part of 2.1.0?

trivialfis:
I can backport it if needed. I think some dependencies in the grpc is linking this library so xgboost doesn't have to. In addition, dlopen is only used for pypi distribution, which happens to have grpc by default as well.

PR #10449, opened 2024-06-19
[CI] [jvm-packages] Build libxgboost4j.dylib on M1 MacOS with OpenMP support

hcho3:
Currently, the CI pipeline builds `libxgboost4j.dylib` (M1, arm64) with OpenMP turned off.
We should enable it for M1 as well.


hcho3:
Yes, it affects user binaries. You should put this in the maintenance, since it doesn't add a new feature

PR #10456, opened 2024-06-19
[R] Redesigned `xgboost()` interface skeleton

david-cortes:

This PR starts off the redesigned interface for `xgboost()`, by adding a bare-bones version that replaces the old one:
* Uses x/y interface instead of `xgb.DMatrix`.
* Enforces a match between the type and shape of `y` and the objective.
* Makes the `objective` optional, picking a default according to the type of `y`.
* Allows column name matching in parameters that refer to columns.
* Uses `xgb.QuantileDMatrix` whenever possible.
* Keeps extra metadata about `y` (column names and factor levels).
* Inherits from `xgb.Booster`, but has an additional class `xgboost`, so that the current methods can be used on it.

But:
* Doesn't yet make use of the additional metadata that it keeps in other methods, beyond the new `print` that's added here.
* Doesn't yet have a more idiomatic `predict` method.
* Has not yet updated examples and guides to use this function.
* Doesn't yet deal with the unidiomatic passing of `params`.
* Doesn't yet offer some functionalities that it should, like custom metrics.
* Has not yet updated the serialization doc for this new class.

Among others. It should nevertheless provide a base to work on.

A couple items I am not so sure about - would be helpful to get reviews on these:
* Does the function in this PR correctly capture all the possible cases in which there can be more than one score per leaf? (e.g. multinomial, multi-task).
* Is it treating `Surv` objects correctly for all cases?
* Where can I find reference papers to add to the docs explaining how XGBoost implements things like monotonicity constraints and interaction contrstraints? I see there's the original paper that introduces XGBoost that is mentioned in the docs, but I imagined that a lot more might have been written by the same authors over time.

CCing @mayer79 for review.


trivialfis:
Thank you for sharing the early work. I will look into the new example and introduction for learning about the new interface.

david-cortes:
> Thank you for sharing the early work. I will look into the new example and introduction for learning about the new interface.

To clarify: there's no intro and only a very small example, since there's no specialized `predict` yet for this class, for example.

trivialfis:
Would you like me to add a feature branch in xgboost for you to iterate on?

david-cortes:
> Would you like me to add a feature branch in xgboost for you to iterate on?

Thanks but I don't think there's any need for that. Or are you planning a 2.1.1 release soon?

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+        stop("Internal error.")
+      }
+
+      out$dmatrix_args <- list(
+        label = y[, 1L] * (2 * (y[, 2L] - 0.5))
+      )
+
+    } else {
+      if (y_attr$type == "left") {
+        lb <- ifelse(
```
mayer79, line 304: Instead of the ifelse() maybe

lb <- ub <- y[, 1L]
lb[y[, 2L] == 0] <- 0

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+          0,
+          y[, 1L]
+        )
+        ub <- y[, 1L]
+        out$dmatrix_args <- list(
+          label_lower_bound = lb,
+          label_upper_bound = ub
+        )
+      } else if (y_attr$type == "right") {
+        lb <- y[, 1L]
```
mayer79, line 315: Maybe

```r
lb <- ub <- y[, 1L]
ub[y[, 2L] == 0] <- Inf
```

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+
+check.nthreads <- function(nthreads) {
+  if (is.null(nthreads)) {
+    return(1L)
+  }
+  if (!inherits(nthreads, c("numeric", "integer")) || !NROW(nthreads)) {
+    stop("'nthreads' must be a positive scalar value.")
+  }
+  if (length(nthreads) > 1L) {
+    nthreads <- head(nthreads, 1L)
```
mayer79, line 504:
```suggestion
    nthreads <- nthreads[1L]
```

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+  }
+  if (is.na(nthreads) || nthreads < 1) {
```
mayer79, line 506:
```suggestion
  if (is.na(nthreads) || nthreads < 1L) {
```

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+      stop(
+        "'feature_weights' does not match in length with columns of 'x' (",
+        length(feature_weights), " vs. ", ncol(x), ")."
+      )
+    }
+    if (NROW(names(feature_weights)) && NROW(colnames(x))) {
+      matched <- match(colnames(x), names(feature_weights))
+      matched <- matched[!is.na(matched)]
+      matched <- matched[!duplicated(matched)]
+      if (length(matched) > 0 && length(matched) < length(feature_weights)) {
```
mayer79, line 574:
```suggestion
      if (length(matched) > 0L && length(matched) < length(feature_weights)) {
```

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+    if (NROW(monotone_constraints) > ncol(x)) {
+      stop(
+        "'monotone_constraints' contains more entries than there are columns in 'x' (",
+        NROW(monotone_constraints), " vs. ", ncol(x), ")."
+      )
+    }
+
+    if (is.list(monotone_constraints)) {
+
+      if (!NROW(names(monotone_constraints))) {
```
mayer79, line 601: We need the names multiple times
```r
nms <- names(monotone_constraints)
```

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+          paste(
+            names(monotone_constraints)[duplicated(names(monotone_constraints))] |> head(),
+            collapse = ", "
+          )
+        )
+      }
+      if (NROW(setdiff(names(monotone_constraints), colnames(x)))) {
+        stop(
+          "'monotone_constraints' contains column names not present in 'x': ",
+          paste(head(names(monotone_constraints)), collapse = ", ")
```
mayer79, line 622:
```suggestion
          paste(utils::head(names(monotone_constraints)), collapse = ", ")
```

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+          stop(
+            "Passed a character vector for 'interaction_constraints', but 'x' ",
+            "has no column names to match them against."
+          )
+        }
+        out <- match(idx, cnames) - 1L
+        if (anyNA(out)) {
+          stop(
+            "'interaction_constraints' contains column names not present in 'x': ",
+            paste(head(idx[which(is.na(out))]), collapse = ", ")
```
mayer79, line 697:
```suggestion
            paste(utils::head(idx[which(is.na(out))]), collapse = ", ")
```

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+          stop("Column indices for 'interaction_constraints' must follow base-1 indexing.")
+        }
+        if (max(idx) > ncol(x)) {
+          stop("'interaction_constraints' contains invalid column indices.")
+        }
+        if (is.numeric(idx)) {
+          if (any(idx != floor(idx))) {
+            stop(
+              "'interaction_constraints' must contain only integer indices. Got non-integer: ",
+              paste(head(idx[which(idx != floor(idx))]), collapse = ", ")
```
mayer79, line 715:
```suggestion
              paste(utils::head(idx[which(idx != floor(idx))]), collapse = ", ")
```

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+      )
+    }
+    if (NROW(names(feature_weights)) && NROW(colnames(x))) {
+      matched <- match(colnames(x), names(feature_weights))
+      matched <- matched[!is.na(matched)]
+      matched <- matched[!duplicated(matched)]
+      if (length(matched) > 0 && length(matched) < length(feature_weights)) {
+        stop(
+          "'feature_weights' names do not contain all columns of 'x'. Missing: ",
+          head(setdiff(colnames(x), names(feature_weights)))
```
mayer79, line 577:
```suggestion
          utils::head(setdiff(colnames(x), names(feature_weights)))
```

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+        )
+      }
+      if (!NROW(colnames(x))) {
+        stop("If passing 'monotone_constraints' as a named list, 'x' must have column names.")
+      }
+      if (anyDuplicated(names(monotone_constraints))) {
+        stop(
+          "'monotone_constraints' contains duplicated names: ",
+          paste(
+            names(monotone_constraints)[duplicated(names(monotone_constraints))] |> head(),
```
mayer79, line 614:
```suggestion
            names(monotone_constraints)[duplicated(names(monotone_constraints))] |> utils::head(),
```

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+print.xgboost <- function(x, ...) {
+  cat("XGBoost model object\n")
+  cat("Call:\n  ")
+  print(attributes(x)$call)
+  cat("Objective: ", attributes(x)$params$objective, "\n", sep = "")
+  cat("Number of iterations: ", xgb.get.num.boosted.rounds(x), "\n", sep = "")
+  cat("Number of features: ", xgb.num_feature(x), "\n", sep = "")
+
+  printable_head <- function(v) {
+    v_sub <- head(v, 5L)
```
mayer79, line 974:
```suggestion
    v_sub <- utils::head(v, 5L)
```

mayer79 (approved):
Awesome work @david-cortes !

The auto-selection of the objective is really neat.

While reading the code, I added some detail stuff (not that they are important, it was just to keep me busy).

Some additional comments:

- I'd prefer helper functions like `process.x.and.col.args()` to be named like `_process_x_and_col_args()`.
- We will also need to deal with `X` (store names of features, store factor levels of character/factor columns).

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+    }
+    if (!is.numeric(w)) {
+      w <- as.numeric(w)
+    }
+    if (length(w) != nrow_y) {
+      stop(
+        "'weights' must be a 1-d vector with the same length as 'y' (",
+        length(w), " vs. ", nrow_y, ")."
+      )
+    }
```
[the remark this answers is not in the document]
david-cortes: This is already checked in the core library.

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+      stop(
+        "Got numeric 'y' - supported objectives for this data are: ",
+        paste(.REGRESSION_OBJECTIVES(), collapse = ", "),
+        ". Was passed: ",
+        objective
+      )
+    }
+
+    n_targets <- 1L
+    if (objective == "reg:quantileerror" && NROW(params$quantile_alpha) > 1) {
```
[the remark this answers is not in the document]
david-cortes: It is not necessarily the case that `NROW` will return an integer type, so there isn't always an advantage in declaring the comparison as integer.

R-package/R/xgboost.R
david-cortes: Retrieving the names does not generate a deep copy of them.

R-package/R/xgboost.R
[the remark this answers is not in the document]
mayer79: > It is not necessarily the case that `NROW` will return an integer type, so there isn't always an advantage in declaring the comparison as integer.

It either returns length() or dim()[1], both are ints. But we are discussing nano seconds ;).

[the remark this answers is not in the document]
david-cortes:
```r
class(NROW(1:3000000000))
```
```
[1] "numeric"
```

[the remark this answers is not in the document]
mayer79: I see! But that's quite a large number of quantiles to model ;-).

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+        "Survival objectives are only supported for types: ",
+        paste(supported_surv_types, collapse = ", "),
+        ". Was passed: ",
+        y_attr$type
+      )
+    }
+
+    if (is.null(objective)) {
+      objective <- "survival:aft"
+    } else {
```
trivialfis, line 257: Nice change for survival training!

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+          Inf,
+          y[, 1L]
+        )
+        out$dmatrix_args <- list(
+          label_lower_bound = lb,
+          label_upper_bound = ub
+        )
+      } else if (y_attr$type == "interval") {
+        out$dmatrix_args <- list(
+          label_lower_bound = ifelse(y[, 3L] == 2, 0, y[, 1L]),
```
trivialfis, line 327: Could you please add some comments for these indexing?

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+        length(w), " vs. ", nrow_y, ")."
+      )
+    }
+    lst_args$dmatrix_args$weight <- w
+  }
+  return(lst_args)
+}
+
+check.nthreads <- function(nthreads) {
+  if (is.null(nthreads)) {
```
trivialfis, line 497: Is this going to make single thread the default for R xgboost?

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+    return(1L)
+  }
+  if (!inherits(nthreads, c("numeric", "integer")) || !NROW(nthreads)) {
+    stop("'nthreads' must be a positive scalar value.")
+  }
+  if (length(nthreads) > 1L) {
+    nthreads <- head(nthreads, 1L)
+  }
+  if (is.na(nthreads) || nthreads < 1) {
```
trivialfis, line 506: XGBoost takes 0 as using all available cores.

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+  if (is.numeric(nthreads)) {
+    if (floor(nthreads) != nthreads) {
+      stop("'nthreads' must be an integer.")
+    }
+  }
+  return(as.integer(nthreads))
+}
+
+check.can.use.qdm <- function(x, params) {
+  if (inherits(x, "sparseMatrix") && !inherits(x, "dgRMatrix")) {
```
trivialfis, line 518: We can support COO internally, for CSC, maybe we can simply transpose in R. Do you think it can help the make code less complex?

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+    stop("'x' cannot be NULL.")
+  }
+  if (inherits(x, "xgb.DMatrix")) {
+    stop("Cannot pass 'xgb.DMatrix' as 'x' to 'xgboost()'. Try 'xgb.train()' instead.")
+  }
+  supported_x_types <- c("data.frame", "matrix", "dgCMatrix", "dgRMatrix")
+  if (!inherits(x, supported_x_types)) {
+    stop(
+      "'x' must be one of the following classes: ",
+      paste(supported_x_types, collapse = ", "),
```
trivialfis, line 551: Do we need to worry about `data.table` and potentially other libraries that inherit from these known types in this error message?

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+    if (is.list(feature_weights)) {
+      feature_weights <- unlist(feature_weights)
+    }
+    if (!inherits(feature_weights, c("numeric", "integer"))) {
+      stop("'feature_weights' must be a numeric vector or named list matching to columns of 'x'.")
+    }
+    if (length(feature_weights) != ncol(x)) {
+      stop(
+        "'feature_weights' does not match in length with columns of 'x' (",
+        length(feature_weights), " vs. ", ncol(x), ")."
```
trivialfis, line 567: Is it possible that we can use the check in c++ https://github.com/dmlc/xgboost/blob/513d7a7d84678fc4e75c18899fe52263258f0290/src/data/data.cc#L785 ?

R-package/R/xgboost.R
```diff
@@ -1,32 +1,1022 @@
...
+  }
+  if ("process_type" %in% names(params)) {
+    if (params$process_type != "default") {
+      stop("Non-default 'process_type' is not supported for 'xgboost()'. Try 'xgb.train()'.")
+    }
+  }
+
+  params_function_args <- c(
+    "objective", "verbose", "verbosity", "nthread", "seed",
+    "monotone_constraints", "interaction_constraints"
```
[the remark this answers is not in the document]
david-cortes: They are meant to be passed as arguments to the function `xgboost()` instead of being passed through `params`.

R-package/R/xgboost.R
david-cortes: Changed the condition and added it to the docs.

R-package/R/xgboost.R
david-cortes: No, the default is to use all threads, as obtained when calling `parallel::detectCores()`.

R-package/R/xgboost.R
david-cortes: Added a comment block above.

R-package/R/xgboost.R
david-cortes: No, because all `data.table` objects are also `data.frame`s.
```r
class(data.table())
````
```
[1] "data.table" "data.frame"
```

R-package/R/xgboost.R
david-cortes: Maybe, but what would be the benefit in such case if the data still needs to be duplicated?

david-cortes: Modified it to accept COO and CSC for QDM, with a casting for now until direct QDM construction from COO is added.

R-package/R/xgboost.R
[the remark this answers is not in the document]
trivialfis: `monotone_constraints` and `interaction_constraints` might be better as a part of normal params? This way it feels more consistent. It's difficult from a user's perspective to understand why some of these parameters are different.

R-package/R/xgboost.R
trivialfis: I'm aware of this, hence the question, is it possible that this error message might lead the user to think `data.frame` is not supported? Or it's common knowledge in the R world?

david-cortes: No, because if `class(x)` contains `data.frame` as an entry, `inherits` will return `TRUE`, so there will be no error message in a derived class like `data.table` or `tibble`. And if it is a class like `arrow` or polars `DataFrame` which do not inherit from `data.frame`, then the error will show as expected, since they are not supported.

R-package/R/xgboost.R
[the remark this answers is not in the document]
david-cortes: From my perspective:
* Some parameters I chose to put as function arguments in order to give them a default value visible to the user (e.g. `nthreads`) or make the default value change according to the data (e.g. `objective`).
* Some parameters are independent of the data (e.g. regularization) or work regardless of whether the data supports them or not (e.g. max categ. to one-hot), while others refer specifically to columns in the data (e.g. interaction constraints) and would not work with a different dataset.
* Ideally, later on there should be a function like `xgb.control` or similar that would list the parameters and offer auto-completion by having them as function arguments (e.g. like `VGAM::vglm.control` or `glmx::glmx.control`). Those train-control parameters typically involve things that are independent of the data.
* Alternatively, and as is more common in R, _all_ parameters could be moved to function arguments (e.g. like `ranger::ranger` or `glmnet::glmnet`), but perhaps that would make the docs very hard to use.

> It's difficult from a user's perspective to understand why some of these parameters are different.

Yes, it's a bit arbitrary, but that is also the situation in the other interfaces - e.g. the number of boosting rounds is a function argument, as is `verbose_eval` in the python interface. It also feels like parameters such as `seed` should be in the train control list, but that would have other side effects due to the logic introduced in an earlier PR (https://github.com/dmlc/xgboost/pull/10029).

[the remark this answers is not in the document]
david-cortes: Perhaps for v3 you could consider having a division between data-dependent parameters (e.g. `feature_weights` is passed under `DMatrix`, while `monotone_contraints` is passed under `params`) and data-independent parameters.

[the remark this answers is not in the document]
trivialfis: Thank you for the detailed explanation. Please think of me as a newcomer instead of a maintainer. Or think of the future GitHub issues we will need to reply to when someone puts the parameter in the wrong place and asks why.

>  but that is also the situation in the other interfaces

I think that's only true for the low-level interface, and some of them are only in this way because we think of it as an unnecessary breaking change. For higher-level interfaces like the scikit-learn, or Spark, they have a relatively consistent interface. For the scikit-learn, we are moving the parameters according to the scikit-learn estimator guideline where the constructor takes hyper-parameter and sometimes feature-related data like feature types, while the `fit` method takes data(sample)-dependent parameters that can be split for early stopping, etc.

> default value visible to the user

We can compromise on this, maybe the document is sufficient?

> Alternatively, and as is more common in R, all parameters could be moved to function arguments

That sounds reasonable to me. Personally, I prefer consistency, which has been a pain point for XGBoost for a while now. Would be great to see improvement in this area.

> Perhaps for v3 you could consider having a division between data-dependent parameters (e

Hmm, since this is a new interface already, maybe it's better to do it right the first time if we know where the issue is?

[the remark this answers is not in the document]
david-cortes: Ok then, I'll move all parameters to function arguments, leaving them as undocumented `...` for now (to be changed in a later PR). We'll see how the docs and examples end up looking later on.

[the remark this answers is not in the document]
david-cortes: Modified it by removing `params` and making all parameters as function arguments, by using `...` for now.

trivialfis (approved):

trivialfis:
Thank you for the excellent work, will merge once the CI passes.

Issue #10461, opened 2024-06-20
[R] CRAN issues with `noRemap`


david-cortes:
In the CRAN checks for the last released version of XGBoost:
https://www.stats.ox.ac.uk/pub/bdr/noRemap/xgboost.out

There is now a check `noRemap` with this description:
> Checks with -DR_NO_REMAP used for C++ code
> 
> Tests using _R_CXX_NO_REMAP_= true, which compiles C++ code with
R_NO_REMAP defined.  Otherwise as for the fedora-gcc checks:
see https://www.stats.ox.ac.uk/pub/bdr/Rconfig/r-devel-linux-x86_64-fedora-gcc
> 
> It is planned that this will become the default in due course, first
for CRAN incoming checks from 2024-04-14.

This gives compilation errors in XGBoost in all of the calls to R C-level functions that are not prefixed with `Rf_`, such as this:
```
xgboost_R.cc:684:28: error: 'allocVector' was not declared in this scope; did you mean 'Rf_allocVector'?
  684 |   out_shape_sexp = PROTECT(allocVector(INTSXP, out_dim));
      |                            ^~~~~~~~~~~
```

I am guessing that perhaps these short-hand no-Rf function names could be enabled by manually adding some define. Perhaps @jameslamb could comment here since LightGBM doesn't appear to suffer from these issues in the checks.

Otherwise, it might be a good idea to change all usages of R C-level functions to their equivalents that have `Rf_` prefix (e.g. `mkChar` -> `Rf_mkChar`).

jameslamb:
Thanks for the `@`.

In LightGBM, we have for a long time used this pattern:

```cpp
#define R_NO_REMAP
#define R_USE_C99_IN_CXX
#include <Rinternals.h>
```

ref: https://github.com/microsoft/LightGBM/blob/5cd95a5b161d7630731d50e9ac529c6bf3dc809f/R-package/src/lightgbm_R.h#L10-L12

Combined with, as @david-cortes mentioned above, using the `Rf_` prefix for things included from the R headers (e.g. `Rf_allocVector()`).

I suspect that's why `{lightgbm}` is not seeing this issue on CRAN. It should be safe for `{xgboost}` to do the same.

I think the best solution is just to do that (define `R_NO_REMAP` + use the `Rf_` prefix). I'd be happy to put up a PR with that change.

trivialfis:
[...]

PR #10465, opened 2024-06-20
[R] Use `Rf_` prefix in C-level R functions

david-cortes:
ref https://github.com/dmlc/xgboost/issues/10461

This PR changes the calls to R C-level functions to use the function versions that have names starting with `Rf_`, as it will be required at CRAN in the future. Along the way, it forcibly adds the macro `R_NO_REMAP` so that introduction of further non-Rf calls would turn into errors.

Note that the CRAN issue is for the current released version there, so if they request changes, it will be necessary to create a similar PR specifically for the CRAN release. At this point the branches differ too much for a simple backport of this PR, but the required changes will become evident once macro `R_NO_REMAP` gets added.

### DIFF
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -352,7 +352,7 @@ void _DMatrixFinalizer(SEXP ext) {
 
 XGB_DLL SEXP XGBSetGlobalConfig_R(SEXP json_str) {
   R_API_BEGIN();
-  CHECK_CALL(XGBSetGlobalConfig(CHAR(asChar(json_str))));
+  CHECK_CALL(XGBSetGlobalConfig(CHAR(Rf_asChar(json_str))));
   R_API_END();
   return R_NilValue;
 }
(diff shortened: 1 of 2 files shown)


trivialfis (commented):
Out of curiosity, are there any prefixed functions for `STRING_ELT`, `CHAR`, etc?

david-cortes:
> Out of curiosity, are there any prefixed functions for `STRING_ELT`, `CHAR`, etc?

I don't see any such `Rf_` versions in the header:
https://github.com/wch/r-source/blob/0196c9d7cef6c21cad02eed747f11acef036743e/src/include/Rinternals.h

jameslamb:
This looks like the right fix to me, thanks for doing it here.

It looks like changing `PROTECT()` and `UNPROTECT()` isn't strictly necessary, as those look like they'd always resolve to the `Rf_*` equivalents regardless of the use of `R_NO_REMAP`.

```cpp
/* Pointer Protection and Unprotection */
#define PROTECT(s)	Rf_protect(s)
#define UNPROTECT(n)	Rf_unprotect(n)
#define UNPROTECT_PTR(s)	Rf_unprotect_ptr(s)
```

https://github.com/wch/r-source/blob/0196c9d7cef6c21cad02eed747f11acef036743e/src/include/Rinternals.h#L382-L384

But doing that type of change does reduce the risk of conflicts from other, not-related-to-R macros named `PROTECT` and `UNPROTECT` (which are fairly generic names).

If you have time/interest, I'd welcome a PR in LightGBM also making the change `PROTECT -> Rf_protect` and `UNPROTECT -> Rf_unprotect`.

trivialfis (approved):

PR #10487, opened 2024-06-25
[CI] Add CI pipeline to build libxgboost4j.so targeting Linux ARM64

hcho3:
https://discuss.xgboost.ai/t/status-of-linux-arm64-support-in-xgboost4j/3657

* Add CI pipeline to build libxgboost4j.so targeting Linux ARM64. Target glibc 2.17.
* Clean up M1 pipeline
* Create a separate pipeline to build libxgboost4j.so (Linux x86_64) that's compatible with glibc 2.17. We don't want to require glibc 2.28 for the Maven release.

### DIFF
--- a/tests/buildkite/build-jvm-linux-arm64.sh
+++ b/tests/buildkite/build-jvm-linux-arm64.sh
@@ -14,7 +14,7 @@ $command_wrapper bash -c \
   "cd build && cmake .. -GNinja -DJVM_BINDINGS=ON -DUSE_OPENMP=ON && ninja -v"
 ldd lib/libxgboost4j.so
 
-echo "--- Upload libxgboost4j.dylib"
+echo "--- Upload libxgboost4j.so"
 pushd lib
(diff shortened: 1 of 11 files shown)


tests/buildkite/build-jvm-linux-arm64.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# Build XGBoost4J binary
+echo "--- Build libxgboost4j.so"
+set -x
+mkdir build
+$command_wrapper bash -c \
+  "cd build && cmake .. -GNinja -DJVM_BINDINGS=ON -DUSE_OPENMP=ON && ninja -v"
+ldd lib/libxgboost4j.so
+
+echo "--- Upload libxgboost4j.dylib"
```
hcho3, line 17:
```suggestion
echo "--- Upload libxgboost4j.so"
```

Should fix typo

hcho3:
The CI pipeline can use a big spring cleaning. I will clean it once I finalize the JVM release.

PR #10488, opened 2024-06-26
[EM] Allow staging ellpack on host for GPU external memory.

trivialfis:
- New parameter `on_host`.
- Abstract format creation and stream creation into policy classes.

### DIFF
--- a/src/data/ellpack_page_source.cu
+++ b/src/data/ellpack_page_source.cu
@@ -1,29 +1,153 @@
 /**
  * Copyright 2019-2024, XGBoost contributors
  */
+#include <thrust/host_vector.h>
+
 #include <memory>
 
+#include "../common/common.h"  // for safe_cuda
+#include "../common/cuda_pinned_allocator.h"
 #include "ellpack_page.cuh"
 #include "ellpack_page.h"  // for EllpackPage
 #include "ellpack_page_source.h"
 
 namespace xgboost::data {
-void EllpackPageSource::Fetch() {
-  dh::safe_cuda(cudaSetDevice(device_.ordinal));
+struct EllpackHostCache {
+  thrust::host_vector<std::int8_t, common::cuda::pinned_allocator<std::int8_t>> cache;
+
+  void Resize(std::size_t n) {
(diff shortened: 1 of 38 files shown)


trivialfis:
cc @rongou .

src/data/ellpack_page_source.cu
```diff
@@ -1,29 +1,161 @@
...
+  bst_idx_t bound_{0};
+
+ public:
+  explicit EllpackHostCacheStreamImpl(std::shared_ptr<EllpackHostCache> cache)
+      : cache_{std::move(cache)} {}
+
+  [[nodiscard]] bst_idx_t Write(void const* ptr, bst_idx_t n_bytes) {
+    auto n = cur_ptr_ + n_bytes;
+    if (n > cache_->cache.size()) {
+      cache_->Resize(n, dh::DefaultStream());
```
rongou, line 41: Allocating pinned memory seems pretty expensive. I wonder if it's worth calculating the total size and allocate the cache in one shot.

trivialfis: That requires an additional iteration over the disk cache, which is not cheap either.

trivialfis: The vector class uses dynamic table internally, the amortized cost should be reasonable, we can look into the initialization cost after removing the disk cache entirely.

trivialfis: Alternatively, as you suggested, we can let system pagable memory to back the main cache.

rongou: In my test, allocating a 120GB pinned memory buffer takes 5-6 seconds. But as a first pass, this is probably fine.

rongou (approved):

src/data/ellpack_page_source.cu
trivialfis: It's expensive indeed. If we extrapolate on that speed assuming it's linear, then 500 GB would be 21 seconds, probably fine for the initial implementation. We will look into malloc as the main cache.

PR #10499, opened 2024-06-28
[jvm-packages] handle null to avoid potential NullPointerException

ShangjinTang:
This PR add null checks for `jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/TrackerProperties.java`.

It retrieves the current thread's class loader using `Thread.currentThread().getContextClassLoader()` and verifies if it is not null.

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/TrackerProperties.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/TrackerProperties.java
@@ -22,10 +22,12 @@ public class TrackerProperties {
     InputStream inputStream = null;
 
     try {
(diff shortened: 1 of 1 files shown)


trivialfis:
Hi, thank you for the PR. Could you please elaborate on under what circumstances it would be null, and what happens if it's null after this PR?

ShangjinTang:
Hello,

This is a failure that we scanned from our static code check tool and marked as "must be fixed".
I also searched from web, some developers have occured that `Thread.currentThread().getContextClassLoader()` might be null (although I cannot reproduce the issue).
I am using the xgboost4j library on Android, so the entire process would crash if contextClassLoader is null.

Thanks

wbo4958:
Looks like TrackerProperties.java is not used anymore after tracker reworking.

PR #10529, opened 2024-07-02
[EM] Move prefetch in reset into the end of the iteration.

trivialfis:
This helps make the prefetching continuous through multiple iterations. The reset function is called at the beginning of an iteration, which as the name suggests, resets the data fetching process. However, this adds one additional fetch at the beginning, and the use of the first batch must wait for this fetch. This PR makes the last iteration responsible for fetching the first batch of the data while the last iteration is still being run, as a result, the DMatrix will be ready when it's iterated again. The result is if we set the number of prefetched batches to 1, the computation and memory copy can be performed in lockstep.

The difficulty is in supporting multiple use cases. For hist, the source doesn't need to be kept in sync after the initial cache build, but it is required before the build. For approx, the source is always synced.


trivialfis:
cc @rongou

rongou (approved):

PR #10534, opened 2024-07-02
Federated plugin for histogram.

trivialfis:
Rebased version https://github.com/dmlc/xgboost/pull/10410 . After all the extracted PRs, this PR:
- Split up the histogram builder into policies, one for normal build, and another for encrypted build.
- Each policy is responsible for both sample split and column split.
- Federated learning without encryption is handled by the normal build.


trivialfis:
cc @rongou This is ready for review now.

rongou (approved):

PR #10538, opened 2024-07-03
Cache GPU histogram kernel configuration.

trivialfis:
API calls like `cudaFuncSetAttribute` can be blocked by memory transfer. This PR caches the configured kernels.

Only the first call to reset is necessary since the feature group is only initialized once. I keep the change simple and let GPU hist call histogram reset once per iteration to align with other member variables.

Will merge the quantizer and feature group into it in the future.


trivialfis:
Apologies for the slow reply lately. Will get to this ASAP.

hcho3 (approved):

PR #10540, opened 2024-07-03
[fed] Provide the initial federated plugin interface.

trivialfis:
Extracted from https://github.com/dmlc/xgboost/pull/10534 . See the https://github.com/dmlc/xgboost/pull/10410 for existing limitations. The most outstanding issue of the interface is the lack of support for multi-class. To enable this, we need to further refactor the tree methods, which is out of scope for the initial implementation. Most of the tests are in the main PR.

We will be targeting the feature branch for now. The merge into the master branch will start once the feature branch is complete.

### DIFF
--- a/plugin/federated/federated_plugin.h
+++ b/plugin/federated/federated_plugin.h
@@ -1,0 +1,361 @@
+/**
+ * Copyright 2024, XGBoost contributors
+ *
+ * @brief This file defines the interface required for a federated plugin to implement.
+
+ * For federated learning, operations for the gradient and gradient histogram are
+ * performed in the encrypted space, and the encryption is provided by a third-party
+ * plugin. The interface is split into four sections:
+ *
+ *   - Library handle.
+ *   - Gradient encryption.
+ *   - Build histogram for vertical federated learning.
+ *   - Build histogram for horizontal federated learning.
+ *
+ * See below function prototypes for details. All prototypes are for C functions that are
+ * suitable for `dlopen`.
+ */
+#pragma once
+
+#include <algorithm>   // for copy_n
+#include <cstdint>     // for uint8_t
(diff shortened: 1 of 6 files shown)


trivialfis:
cc @rongou .

plugin/federated/federated_plugin.h
```diff
@@ -0,0 +1,361 @@
...
+
+ public:
+  ~FederatedPluginMock() = default;
+
+  [[nodiscard]] common::Span<std::uint8_t> EncryptGradient(
+      common::Span<float const> data) override {
+    grad_.resize(data.size_bytes());
+    auto casted =
+        common::Span{reinterpret_cast<std::uint8_t const *>(data.data()), data.size_bytes()};
+    std::copy_n(casted.data(), casted.size(), grad_.data());
```
rongou, line 227: Do we still plan to use the mock plugin for unencrypted federated learning? I wonder if we can somehow avoid these types of copying.

trivialfis: Not at the moment, it's only used for running tests. Any suggestions? I don't what this mock myself, but haven't found a way to run test without one.

trivialfis: Probably we can write the histogram build in a way that can reuse the existing histogram implementation. Will look into it tomorrow.

trivialfis: Hmm, then the tests wouldn't be reliable. Also, the plugin itself cannot communicate, so we can't hand the entire histogram build to the plugin. 
Any suggestion for design would be great.

rongou: For testing it's probably fine.

rongou (approved):

PR #10541, opened 2024-07-04
Avoid the use of size_t in the partitioner.

trivialfis:
size_t is implementation-defined, which causes issue when we want to pass pointer or span.


hcho3 (approved):

trivialfis:
Will take a bit of work, also removed the const cast on the way.

trivialfis:
- Use `Span` instead of `Elem` where `node_id` is not needed.
- Remove the `const_cast`.
- Make sure the constness is not removed in the `Elem` by making it reference only.

trivialfis:
@hcho3 could you please help take another look when you are available?

hcho3 (approved):

PR #10542, opened 2024-07-04
[fed] Replace secure split with query in coll.

trivialfis:
Extracted from https://github.com/dmlc/xgboost/pull/10534 . This PR removes the secure col split and prefers querying whether a plugin is available.

We will be targeting the feature branch for now. The merge into the master branch will start once the feature branch is complete.


trivialfis:
cc @rongou

rongou (approved):

PR #10543, opened 2024-07-04
SYCL. Reorder if-else statements to allow using of cpu branches for sycl-devices

razdoburdin:
Hi.
While working on `sycl` plugin migration to xgboost 2.1.0, I faced the same dispatching problem as in https://github.com/dmlc/xgboost/pull/9682. 
Current PR simply rearrange the if-else statements to execute `cpu` code instead of `cuda `code for `sycl `devices.
Such change doesn't affect any `cpu/cuda` logic, but is helpful for `sycl `development.


trivialfis (approved):
In the future, we can use the `DispatchDevice` method more.

PR #10544, opened 2024-07-04
Avoid thrust vector initialization.

trivialfis:
- Extract all vector wrappers into a different header to reduce the size of the helper.
- Add a wrapper for rmm device uvector.
- Split up the `Resize` method for HDV.

During construction or resize, the thrsut `device_vector` launches an additional kernel and a stream sync to initialize the content. We can't provide a different synchronization policy to thrust. This PR skips the initialization when rmm is used. rmm has a `device_uvector` which skips the initialization. To use it, we wrapped a rmm memory resource into the xgboost memory logger.

### DIFF
--- a/include/xgboost/host_device_vector.h
+++ b/include/xgboost/host_device_vector.h
@@ -135,7 +135,9 @@ class HostDeviceVector {
 
   void SetDevice(DeviceOrd device) const;
 
-  void Resize(size_t new_size, T v = T());
+  void Resize(std::size_t new_size);
+  /** @brief Resize and initialize the data if the new size is larger than the old size. */
+  void Resize(size_t new_size, T v);
 
   using value_type = T;  // NOLINT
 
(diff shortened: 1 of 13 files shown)


trivialfis:
Need to fix the initialization error: https://buildkite.com/xgboost/xgboost-ci/builds/5919#01907e66-bb36-4c38-8642-270bbb6a4026 .

trivialfis:
cc @rongou .

include/xgboost/host_device_vector.h
```diff
@@ -135,7 +135,9 @@ class HostDeviceVector {
 
   void SetDevice(DeviceOrd device) const;
 
-  void Resize(size_t new_size, T v = T());
+  void Resize(std::size_t new_size);
+  /** @brief Resize and initialize the data if the new size is larger than the old size. */
+  void Resize(size_t new_size, T v);
```
rongou, line 140: Should this also be `std::size_t` to be consistent?

rongou (approved):

include/xgboost/host_device_vector.h
trivialfis: Good catch, will fix.

PR #10549, opened 2024-07-05
[EM] Handle base idx in GPU histogram.

trivialfis:
- support base index in histogram.
- initialize row partitions with base index


trivialfis:
cc @rongou .

rongou (approved):

PR #10564, opened 2024-07-10
[TEST] Drop async code in dask.

trivialfis:
Debug a hang in dask prediction.


trivialfis:
Closing as the bug is fixed.

PR #10566, opened 2024-07-10
[fed] Split up federated test CMake file.

trivialfis:
- Collect all federated test files into the same directory.
- Independently list the files.


trivialfis:
cc @rongou .

rongou (approved):

PR #10291, opened 2024-05-15
Add heu processor library files (for secure boost scheme phase 2)

shaojian-ant:
Add the [heu](https://github.com/secretflow/heu) library implementation for the processor interface of secure xgboost. The processor interface is defined in [#10124](https://github.com/dmlc/xgboost/pull/10124). @ZiyueXu77 @trivialfis

The heu relative code is [here](https://github.com/shaojian-ant/xgb-processing/tree/main), and work is still in progress.

PR #10570, opened 2024-07-11
Add heu library files

shaojian-ant:
This PR replaces [#10291 ], which changes the target branch from dmlc:master to dmlc:vertical-federated-learning


shaojian-ant:
migrated to [#10570 ]

Issue #10314, opened 2024-05-22
Flaky column split tests.


trivialfis:
https://buildkite.com/xgboost/xgboost-ci/builds/5300#018f9f25-a37c-4240-8245-ad8209ede209

Some C++ tests for column splits are flaky. Some of these tests are added in https://github.com/dmlc/xgboost/pull/10112 , which increases the coverage but also causes some random issues on the CI. They are not critical, but we need to investigate how to make these tests deterministic.

trivialfis:
@rongou I think I have found the root cause for this, the bit field is not thread-safe for CPU https://github.com/dmlc/xgboost/blob/1ca4bfd20e442f0a26a90e5f65218bb4af0e7c16/src/tree/common_row_partitioner.h#L57 .

PR #10572, opened 2024-07-11
Fix column split race condition.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/10314 .


trivialfis:
cc @rongou .

rongou (approved):
Good catch!

PR #10582, opened 2024-07-12
Avoid caching allocator for large allocations.

trivialfis:
- Row partitioner.
- Sort for quantile.
- Change the row partitioner to use the reset method to avoid allocation.
- Reuse the definition of row index type.


trivialfis:
cc @rongou .

PR #10587, opened 2024-07-15
[Doc] fix readme

RektPunk:
The cran badge is breaking now, so it changes the address to prevent it:
[CRAN Status Badge](http://cran.r-project.org/web/packages/xgboost) to [CRAN Status Badge](https://cran.r-project.org/web/packages/xgboost)

Just in case, I also changed other HTTP to HTTPS.


trivialfis (approved):
Thank you for the fix and for adding TLS.

PR #10592, opened 2024-07-16
[WIP] [EM] Support resource view in ellpack.

trivialfis:
This PR changes ellpack to use the resource view as the backing storage similar to CPU counterparts. This will help us enable more options for GPU external memory including `mmap`, normal malloc, device malloc, pinned memory, managed memory, etc.


trivialfis:
Closing to avoid unnecessary noise.

PR #10593, opened 2024-07-16
[CI] Reduce the frequency of dependabot PRs

hcho3:
cc @wbo4958


trivialfis:
I will defer this to @wbo4958 . I'm not sure what's the issue with dependent bots.

PR #10595, opened 2024-07-17
Refactor `DeviceUVector`.

trivialfis:
Create a wrapper instead of using inheritance to avoid inconsistent interface of the class (The use of `device_ptr`).

We will probably have something similar to `rmm::device_uvector` in the future if the need arises, but I would like to keep it as a thin wrapper for now.


trivialfis:
cc @rongou

rongou (approved):

PR #10602, opened 2024-07-17
[EM] Support mmap backed ellpack.

trivialfis:
This PR changes ellpack to use the resource view as the backing storage similar to CPU counterparts. This will help us enable more options for GPU external memory including `mmap`, normal malloc, device malloc, pinned memory, and managed memory. For now, only mmap is used. The PR supports both HMM and non-HMM systems.

In addition, some refactoring to split up wrappers of the cuda runtime API to make them usable with a host compiler.

I haven't done any profiling yet as I need another branch to run real batch-based workflow, the madvise and prefetching are mostly placeholders. Will experiment with various backends after this PR.

- Support resource view in ellpack.
- Define the CUDA version of MMAP resource.
- Define the CUDA version of malloc resource.
- Refactor cuda runtime API wrappers, and add memory access related wrappers.
- gather windows macros into a single header.

To-do 
- [x] Dispatch from ellpack source.


trivialfis:
cc @rongou

src/common/resource.cu
```diff
@@ -0,0 +1,43 @@
...
+      n_{length} {
+  auto device = dh::CurrentDevice();
+  dh::safe_cuda(
+      cudaMemAdvise(handle_->base_ptr, handle_->base_size, cudaMemAdviseSetReadMostly, device));
+  dh::safe_cuda(cudaMemAdvise(handle_->base_ptr, handle_->base_size,
+                              cudaMemAdviseSetPreferredLocation, device));
+  dh::safe_cuda(
+      cudaMemAdvise(handle_->base_ptr, handle_->base_size, cudaMemAdviseSetAccessedBy, device));
+  dh::safe_cuda(
+      cudaMemPrefetchAsync(handle_->base_ptr, handle_->base_size, device, dh::DefaultStream()));
```
rongou, line 28: Does this always happen after the data is initialized on the cpu?

trivialfis: Yes, for now. We might optionally only prefetch the gradient index part and leave other scalars on host.

rongou (approved):

PR #10608, opened 2024-07-18
Move device histogram storage into `histogram.cuh`.

trivialfis:
Split up the `updater_gpu_hist.cu` file for easier unittests. Currently, the test file directly includes the `updater_gpu_hist.cu` file and accesses the internal of the updater, which makes it difficult to make modification. Up coming PRs will try to revise some of tests to avoid such direct accesses.


trivialfis:
cc @rongou 
Note: failing tests are unrelated.

rongou (approved):

PR #10609, opened 2024-07-18
[CI] Fix test environment.

trivialfis:


trivialfis:
The ubjson is removed from macos. I opened a PR there https://github.com/Iotic-Labs/py-ubjson/pull/20 see if there's response.

trivialfis:
There's another failure in buidkite with cupy. Will look into it in a different PR.

tests/ci_build/Dockerfile.i386
```diff
@@ -1,7 +1,6 @@
 FROM i386/debian:sid
 
 ENV DEBIAN_FRONTEND noninteractive
-SHELL ["/bin/bash", "-c"]   # Use Bash as shell
```
hcho3, line 4: Can we bring the line back without the comment?
```
SHELL ["/bin/bash", "-c"]
```
This should work.

trivialfis: Done. Thank you for the suggestion. That's a weird bug.

hcho3 (approved):

hcho3:
Merging to unblock the CI.

trivialfis:
Thanks, was about to merge it. Wondering what's the issue with the freebsd test. Hasn't been the most stable action workflow.

hcho3:
> Wondering what's the issue with the freebsd test. Hasn't been the most stable action workflow.

We may consider removing gtest from the FreeBSD tests.

trivialfis:
Will do some debugging with column split. From limited experience, qemu tends to hang if an application oversubscribes the system threads, and column split related tests certainly oversubscribe the threads since all workers are using all available threads.

Issue #10613, opened 2024-07-19
Crash on GPU with large wide dataset.


trivialfis:
Will update the issue once I can ping down the cause, likely overflow somewhere.

PR #10615, opened 2024-07-20
Fix integer overflow.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/10613 .

This was introduced by #10549 by accident. Before that PR, the index was directly passed into the iterator, which accepts a `size_t` as the indexing argument.  #10549 created an intermediate variable, which has the same type as the row index (std::uint32_t), resulting in a smaller type.

The PR also does some small cleanups.


hcho3:
Should it be part of the 2.1.1 patch release?

trivialfis:
No need, it's a recent regression.

trivialfis:
cc @rongou

rongou (approved):

PR #10624, opened 2024-07-22
[CMake] Explicitly link with CCCL (standalone or CTK)

hcho3:
Related: https://github.com/conda-forge/xgboost-feedstock/pull/167


CMakeLists.txt
```diff
@@ -221,6 +222,10 @@ if(USE_CUDA)
   add_subdirectory(${PROJECT_SOURCE_DIR}/gputreeshap)
 
   find_package(CUDAToolkit REQUIRED)
+
+  if(BUILD_WITH_EXTERNAL_CCCL)
+    find_package(CCCL REQUIRED CONFIG COMPONENTS cub)
```
jakirkham, line 227: Would it make sense to always do this (iow dropping `if`)? Think this would pick up the CTK CCCL if no other is found

jakirkham, line 227: Do we need Thrust here too?

```suggestion
    find_package(CCCL REQUIRED CONFIG COMPONENTS cub thrust)
```

CMakeLists.txt
```diff
@@ -427,6 +432,9 @@ else()
     set(INSTALL_TARGETS xgboost)
   endif()
 endif()
+if(BUILD_WITH_EXTERNAL_CCCL)
+  list(APPEND INSTALL_TARGETS _CUB_CUB _CUB_libcudacxx)
```
jakirkham, line 436: Do we need `libcudacxx` in `find_package` as well?

trivialfis (commented):
Do we want to replace thrust as well?

jakirkham (commented):
Thanks Hyunsu! 🙏

Had a couple questions below

CMakeLists.txt
trivialfis: We are not using it at the moment.

CMakeLists.txt
hcho3: `find_package(CCCL)` errors out if `cccl` isn't installed explicitly, since CMake can't locate `cccl-config.cmake` or `FindCCCL.cmake`.

jakirkham: Can we please double check this?

In the Conda provided CTK, we make sure [the CCCL config files are packaged]( https://github.com/conda-forge/cuda-cccl-feedstock/blob/c78dce02651dad7f8f42bcd2156016f747c48b2a/recipe/meta.yaml#L41-L44 )

Would think this is the case in other CTK packages of CCCL. Though even if that were not the case, it would be good to discover and raise with the right folks to fix

CMakeLists.txt
jakirkham: Interesting ok. So in what way is `libcudacxx` being used on this line?

hcho3: This is an artifact of the install script. Without this line, CMake complains:
```
CMake Error: install(EXPORT "XGBoostTargets" ...) includes target "_CUB_libcudacxx" which requires target "_Thrust_libcudacxx" that is not in any export set.
```

CMakeLists.txt
hcho3: Yes, when I tried building it locally without installing `cccl`, I got an error about missing `cccl-config.cmake`. I wonder why, especially because the CTK does contain `cccl-config.cmake`:
```
$ find /usr/local/cuda/targets/x86_64-linux/lib/cmake/cccl/
/usr/local/cuda/targets/x86_64-linux/lib/cmake/cccl/
/usr/local/cuda/targets/x86_64-linux/lib/cmake/cccl/cccl-config.cmake
/usr/local/cuda/targets/x86_64-linux/lib/cmake/cccl/cccl-config-version.cmake
```

cmake/Utils.cmake
```diff
@@ -104,6 +104,11 @@ function(xgboost_set_cuda_flags target)
     target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_NVTX=1)
   endif()
 
+  if(BUILD_WITH_EXTERNAL_CCCL)
+    target_compile_definitions(${target} PRIVATE -DTHRUST_IGNORE_CUB_VERSION_CHECK=1)
```
trivialfis, line 108: If we are using external cccl as a whole package, do we still need this flag?

hcho3: Removed.

CMakeLists.txt
jakirkham: Awesome, thanks Hyunsu! 🙏

Please feel free to mark this thread resolved

CMakeLists.txt
jakirkham: Think we may want to use `target_link_libraries` to accomplish this

For [example]( https://github.com/NVIDIA/cccl/blob/1b16af7ccc2f3aeb4c94a743c951bae3e0586052/examples/example_project/README.md?plain=1#L39-L41 ):

> ```cmake
> # Alternatively, if you're building a library
> add_library(your_library your_file.cu)
> target_link_libraries(your_library PRIVATE CCCL::CCCL)
> ```

Alternatively [with Thrust]( https://github.com/NVIDIA/cccl/blob/1b16af7ccc2f3aeb4c94a743c951bae3e0586052/thrust/thrust/cmake/README.md?plain=1#L20-L22 ) (note `target_link_libraries` probably should use `private` here too):

```cmake
find_package(Thrust REQUIRED CONFIG)
thrust_create_target(Thrust)
target_link_libraries(MyProgram Thrust)
```

hcho3: Hmm, somehow with the latest changes, I don't need to add install targets to avoid CMake error. Problem solved, I suppose?

jakirkham: Was wondering if that might be the case. Great, thanks Hyunsu! 🙏

hcho3:
@jakirkham It seems that for CUDA 11.8, CTK doesn't have the CMake config for CCCL:
```
]$ find /usr/local/cuda/lib64/cmake/
/usr/local/cuda/lib64/cmake/
/usr/local/cuda/lib64/cmake/libcudacxx
/usr/local/cuda/lib64/cmake/libcudacxx/libcudacxx-config.cmake
/usr/local/cuda/lib64/cmake/libcudacxx/libcudacxx-header-search.cmake
/usr/local/cuda/lib64/cmake/libcudacxx/libcudacxx-config-version.cmake
/usr/local/cuda/lib64/cmake/thrust
/usr/local/cuda/lib64/cmake/thrust/README.md
/usr/local/cuda/lib64/cmake/thrust/FindTBB.cmake
/usr/local/cuda/lib64/cmake/thrust/thrust-config-version.cmake
/usr/local/cuda/lib64/cmake/thrust/thrust-config.cmake
/usr/local/cuda/lib64/cmake/cub
/usr/local/cuda/lib64/cmake/cub/cub-config-version.cmake
/usr/local/cuda/lib64/cmake/cub/cub-config.cmake
```

Do we need CTK 12.0+ to access the built-in CCCL? cc @bdice

jakirkham:
That sounds plausible

Can we just fallback to Thrust & CUB in that case?

CMakeLists.txt
```diff
@@ -221,6 +221,23 @@ if(USE_CUDA)
...
   find_package(CUDAToolkit REQUIRED)
+  find_package(CCCL CONFIG)
+  if(NOT CCCL_FOUND)
+    message(STATUS "Standalone CCCL not found. Attempting to use CCCL from CUDA Toolkit...")
+    find_package(CCCL CONFIG
+      HINTS ${CUDAToolkit_LIBRARY_DIR}/cmake)
+    if(NOT CCCL_FOUND)
+      message(STATUS "Could not locate CCCL from CUDA Toolkit. Using Thrust and CUB from CUDA Toolkit...")
+      find_package(Thrust CONFIG REQUIRED
+        HINTS ${CUDAToolkit_LIBRARY_DIR}/cmake)
```
jakirkham, line 232: Think we also need to call [`thrust_create_target`]( https://github.com/NVIDIA/cccl/blob/1b16af7ccc2f3aeb4c94a743c951bae3e0586052/thrust/thrust/cmake/README.md?plain=1#L14-L23 )

```suggestion
      find_package(Thrust CONFIG REQUIRED
        HINTS ${CUDAToolkit_LIBRARY_DIR}/cmake)
      thrust_create_target(Thrust)
```

jakirkham (commented):
Thanks Hyunsu! 🙏

Added a suggestion below

bdice:
> Do we need CTK 12.0+ to access the built-in CCCL? cc @bdice

The better way to get CCCL is via GitHub: https://github.com/NVIDIA/cccl?tab=readme-ov-file#github

That way you can use a CCCL version that works with all the supported CTK versions, rather than having to make exceptions for every CUDA version in the build scripts. I think you need at least CCCL 2.2.0 to make things work properly with CMake, which shipped with CUDA 12.3.0. If you fetch CCCL from GitHub via CPM (or use a conda package of CCCL) you can get a more recent version that will support both CUDA 11 and 12. Currently RAPIDS is using CCCL 2.5.0.

PR #10632, opened 2024-07-24
[dask] Disable `broadcast` in the `scatter` call.

trivialfis:
As noted in the `scatter` document, when AMM is enabled (default), the `broadcast` flag is invalid).

Huge thanks to @rjzamora @dantegd for debugging this.


hcho3 (approved):

rjzamora (approved):
Seems good to me, thanks!

It seems *possible* that `broadcast=True` will scale a bit better when the number of workers in large. However, I think it makes sense to keep the fix simple (as this PR does) until you find clear motivation to optimize further.

PR #10633, opened 2024-07-24
Allow building with CCCL that's newer than CTK

hcho3:
Follow-up to #10624

* When using an external CCCL (e.g. Conda), ensure that the Thrust/CUB headers from the external CCCL are chosen instead of the ones from the CTK. This way, we can build XGBoost with CCCL that's newer than the CTK.
  - The include paths need to be correctly ordered.
     (Bad) `-I/usr/local/cuda/targets/x86_64-linux/include -I$CONDA_PREFIX/include`
     (Good) `-I$CONDA_PREFIX/include -I/usr/local/cuda/targets/x86_64-linux/include`

With this PR, we can build XGBoost with CCCL 2.5.0 + CTK 11.8.

### DIFF
--- a/cmake/Utils.cmake
+++ b/cmake/Utils.cmake
@@ -104,7 +104,9 @@ function(xgboost_set_cuda_flags target)
     target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_NVTX=1)
   endif()
 
-  target_link_libraries(${target} PRIVATE CCCL::CCCL)
+  target_link_libraries(${target}
+    INTERFACE CCCL::CCCL
+    PUBLIC CUDA::cudart_static)
   target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_CUDA=1)
   target_include_directories(
     ${target} PRIVATE
@@ -235,7 +237,6 @@ macro(xgboost_target_link_libraries target)
 
   if(USE_CUDA)
     xgboost_set_cuda_flags(${target})
-    target_link_libraries(${target} PUBLIC CUDA::cudart_static)
   endif()
 
   if(PLUGIN_RMM)
--- a/cmake/Utils.cmake
+++ b/cmake/Utils.cmake
@@ -104,6 +104,8 @@ function(xgboost_set_cuda_flags target)
     target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_NVTX=1)
   endif()
 
+  # Use CCCL we find before CUDA Toolkit to make sure we get newer headers as intended
+  # The CUDA Toolkit includes its own copy of CCCL that often lags the latest releases (and would be picked up otherwise)
   target_link_libraries(${target}
     PRIVATE CCCL::CCCL CUDA::cudart_static)
   target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_CUDA=1)
--- a/cmake/Utils.cmake
+++ b/cmake/Utils.cmake
@@ -105,8 +105,7 @@ function(xgboost_set_cuda_flags target)
   endif()
 
   target_link_libraries(${target}
-    INTERFACE CCCL::CCCL
-    PUBLIC CUDA::cudart_static)
+    PRIVATE CCCL::CCCL CUDA::cudart_static)
   target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_CUDA=1)
   target_include_directories(
     ${target} PRIVATE
--- a/cmake/Utils.cmake
+++ b/cmake/Utils.cmake
@@ -105,9 +105,19 @@ function(xgboost_set_cuda_flags target)
   endif()
 
   # Use CCCL we find before CUDA Toolkit to make sure we get newer headers as intended
-  # The CUDA Toolkit includes its own copy of CCCL that often lags the latest releases (and would be picked up otherwise)
-  target_link_libraries(${target}
-    PRIVATE CCCL::CCCL CUDA::cudart_static)
+  # The CUDA Toolkit includes its own copy of CCCL that often lags the latest releases
+  # (and would be picked up otherwise)
+  if(BUILD_STATIC_LIB)
+    # If the downstream user is statically linking with libxgboost, it needs to
+    # explicitly link with CCCL and CUDA runtime.
+    target_link_libraries(${target}
+      PUBLIC CCCL::CCCL CUDA::cudart_static)
+  else()
+    # If the downstream user is dynamically linking with libxgboost, it does not
+    # need to link with CCCL and CUDA runtime.
+    target_link_libraries(${target}
+      PRIVATE CCCL::CCCL CUDA::cudart_static)
+  endif()
   target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_CUDA=1)
   target_include_directories(
     ${target} PRIVATE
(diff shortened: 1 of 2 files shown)


src/common/quantile.h
```diff
@@ -71,6 +71,14 @@ struct WQSummary {
          << "value: " << e.value;
       return os;
     }
+
+    XGBOOST_DEVICE inline bool operator==(Entry const& rhs) const {
```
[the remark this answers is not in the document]
hcho3: Thanks for providing the context. I removed the workaround.

cmake/Utils.cmake
```diff
@@ -104,7 +104,9 @@ function(xgboost_set_cuda_flags target)
     target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_NVTX=1)
   endif()
 
-  target_link_libraries(${target} PRIVATE CCCL::CCCL)
+  target_link_libraries(${target}
+    INTERFACE CCCL::CCCL
```
trivialfis, line 108: Does this propagate into the C interface?

hcho3: My bad. A quick experiment with the C API demo reveals that the installed CMake config "leaks" CCCL requirement to the consumer of libxgboost. Let me rectify this.

cmake/Utils.cmake
```diff
@@ -235,7 +236,6 @@ macro(xgboost_target_link_libraries target)
 
   if(USE_CUDA)
     xgboost_set_cuda_flags(${target})
-    target_link_libraries(${target} PUBLIC CUDA::cudart_static)
```
jakirkham, line 238: As we are using `PUBLIC` here, am guessing we want to do the same in this change

cmake/Utils.cmake
```diff
@@ -104,7 +104,8 @@ function(xgboost_set_cuda_flags target)
     target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_NVTX=1)
   endif()
 
-  target_link_libraries(${target} PRIVATE CCCL::CCCL)
+  target_link_libraries(${target}
+    PRIVATE CCCL::CCCL CUDA::cudart_static)
```
jakirkham, line 108: This would maintain the same linking behavior as before

```suggestion
    PRIVATE CCCL::CCCL PUBLIC CUDA::cudart_static)
```

jakirkham (commented):
Thanks Hyunsu! 🙏

Had a minor suggestion below

cmake/Utils.cmake
hcho3: @jakirkham No, setting this `PUBLIC` will adversely affect the downstream users of the C API.

hcho3: Actually, maybe downstream users need to install CUDA runtime too? Let me take another look.

jakirkham: Ok happy to be corrected here. Just thought we wanted the same behavior

cmake/Utils.cmake
jakirkham: Since CCCL is header only. There shouldn't be any linking

Though please let me know if I'm missing something

cmake/Utils.cmake
hcho3: Yeah, another test with the C API demo (https://github.com/dmlc/xgboost/tree/master/demo/c-api/basic) reveals that having `CUDA::cudart_static` as a `PUBLIC` dependency will affect the downstream user.

Error
```
[1/1] Linking C executable api-demo
FAILED: api-demo
: && /usr/bin/cc  -rdynamic CMakeFiles/api-demo.dir/c-api-demo.o -o api-demo -L/usr/local/lib64 -Wl,-rpath,/usr/local/lib64  /usr/local/lib64/libxgboost.so  -lCUDA::cudart_static && :
/usr/bin/ld: cannot find -lCUDA::cudart_static
```

I considered revising the CMake config to add `find_dependency(CUDAToolkit)`. It fixed the build, but added an unneeded static lib to the downstream lib:
```
[2/2] : && /usr/bin/cc  -rdynamic CMakeFiles/api-demo.dir/c-api-demo.o -o api-demo -L/usr/local/lib64   -L/usr/local/cuda-11.8/targets/x86_64-linux/lib -Wl,-rpath,/usr/local/lib64  /usr/local/lib64/libxgboost.so  /usr/local/cuda-11.8/targets/x86_64-linux/lib/libcudart_static.a  -lpthread  -ldl  -lrt && :
```

So I think we should make `CUDA::cudart_static` a `PRIVATE` dependency.

cmake/Utils.cmake
hcho3: If `CCCL::CCCL` is an interface dep, the downstream user that uses `xgboost::xgboost` from `find_package(XGBoost)` will be asked to link with the CMake target `CCCL::CCCL`. Whether CCCL is a header-only lib is not relevant here.

hcho3: @trivialfis Does XGBoost use any Thrust or CUB structures in the public API ? My guess is that we don't.

trivialfis: No, XGBoost defines a C API. Anything written in C++ is not part of it.

We have a document about the C++ headers in the `include` directory but have explicitly emphasized that this is not a stable API. In addition, the only dependency in those headers is dmlc-core.

hcho3: Thanks, then we can keep CCCL and CUDA runtime as `PRIVATE` deps.

jakirkham: Ok sounds good. Thanks for the thorough investigations on both of these topics! 🙏

Let's just make sure we are doing the same thing in conda-forge that is being done here. This will help us stay aligned and reduce conda-forge patches over time

Somewhat tangential question, but is Python using the C or C++ interface? Guessing C (through `ctypes`) though wanted to check

cmake/Utils.cmake
trivialfis: I believe @hcho3 is right here.

> maybe downstream users need to install CUDA runtime too? Let me take another look.

No, XGBoost runs on CPU-only machines even if compiled with CUDA. We statically link a CUDA runtime (cudart) shim into XGBoost shared object, the user should not need CUDA dependency.

jakirkham: Ok agreed. Thanks for looking into this! 🙏

cmake/Utils.cmake
trivialfis: > but is Python using the C or C++ interface

C

cmake/Utils.cmake
```diff
@@ -104,7 +104,8 @@ function(xgboost_set_cuda_flags target)
     target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_NVTX=1)
   endif()
 
-  target_link_libraries(${target} PRIVATE CCCL::CCCL)
+  target_link_libraries(${target}
```
trivialfis, line 107: Please help add a comment about why the ordering is of significance here.

jakirkham: Something like this?

```suggestion
  # Use CCCL we find before CUDA Toolkit to make sure we get newer headers as intended
  # The CUDA Toolkit includes its own copy of CCCL that often lags the latest releases (and would be picked up otherwise)
  target_link_libraries(${target}
```

trivialfis: Looks good!

jakirkham:
FWIW LGTM

Thank you both! 🙏

trivialfis (approved):

PR #10639, opened 2024-07-25
[jvm-packages] [breaking] rework xgboost4j-spark and xgboost4j-spark-gpu

wbo4958:
# About this PR

## Features and Fixes

- Introduce an abstract XGBoost Estimator:
  - Create an abstract base class that encapsulates common functionality for all XGBoost estimators.
  - This will promote code reuse and maintain consistency across different estimators.

- Update to the latest XGBoost parameters
  - Add all XGBoost parameters supported in XGBoost4j-spark base on [this](https://xgboost.readthedocs.io/en/stable/parameter.html)
  - Add setter and getter for these parameters.
  - Remove the deprecated parameters

- Address the missing value handling:
  - 0 is a meaningful value in machine learning case, the old version enforces 0 as the missing value which may result in XGBoost model inaccurate.

- Remove any ETL operations in XGBoost
  - As a thin wrapper of XGBoost4j, it shouldn't wrap much ETL in XGBoost4j itself to make code hard to maintain and read.

- Rework the GPU plugin:
  - Rework the GPU plugin to improve its stability and make it more readable and maintainable according to the new design pattern of xgboost4j-spark

- Expand sanity tests for CPU and GPU consistency:
  - Increase the coverage of sanity tests to verify that the training and transform results are consistent between CPU and GPU implementations.
  - This will help identify and prevent any discrepancies or bugs related to the GPU support.

## Breaking changes

- Remove some unused parameters like Rabit related parameters and train_test_ratio. etc.
- Separate XGBoost-spark parameter from the whole XGBoost parameters, and make the constructor of XGBoost estimator only accept the XGBoost parameters instead of parameters for xgboost4j-spark. Eg,
  ``` scala
  val xgbParams: Map[String, Any] = Map(
        "max_depth" -> 5,
        "eta" -> 0.2,
        "objective" -> "binary:logistic"
      )
  val est = new XGBoostClassifier(xgbParams)
        .setNumRound(10)
        .setDevice("cpu")
        .setNumWorkers(2)
  ``` 
   or

    ``` scala
    val est = new XGBoostClassifier(xgbParams)
          .setMaxDepth(5)
          .setEta(0.2)
          .setObjective("binary:logistic")
          .setNumRound(10)
          .setDevice("cpu")
          .setNumWorkers(2)
    ```

## Test this PR

- Unit tests For scala 2.12

``` bash
mvn clean package
mvn -P gpu clean package
```

- Unit tests For scala 2.13
``` bash
cd xgboost
python dev/change_scala_version.py --scala-version 2.13

cd jvm-packages
mvn clean package
mvn -P gpu clean package
```


wbo4958:
Hi @trivialfis @hcho3 @dotbg, Could you kindly help review this PR, Thx very much.

wbo4958:
Hi there, The CI got passed, could you @trivialfis @hcho3 @dotbg help review this PR? Thx very much.

hcho3:
Apologies for the delay, I will look at it this Friday.

trivialfis (commented):
I tried to go through this rework but I'm afraid it's beyond my expertise. Could you please update the examples and documents so that we can get a feel of what's going on with the new interface? Also, we might need to schedule an online interview, @hcho3 might want to join as well.

hcho3:
> Also, we might need to schedule an online interview, @hcho3 might want to join as well.

Yes, I would love to join.

hcho3:
Some questions about build artifacts and packaging:
* Do we continue to build two variants of `libxgboost4j.so`, one with CUDA and one without?
* Now that `xgboost4j-gpu` package is gone, where to we put `libxgboost4j.so` (CUDA variant)? Does it go inside the `xgboost4j-spark-gpu` package?
* How long do we plan to distribute Spark 2.12 and 2.13 variants?

hcho3:
@trivialfis Do you have a particular agenda in mind for the online interview? Would it be primarily a Q&A session? I can set up an Outlook invite to choose the best time for all of us.

trivialfis:
For breaking changes PR, I am mostly concerned about the user interface, hence the request for documentation:
- How does the service loader work. What's the sequence of events for a CPU pipeline and a GPU pipeline. Also, the combination of rapids SQL and XGBoost parameters (4 cases with on/off for both sides).
- What are the necessary changes from users, can we submit deprecation warnings, or can we submit informative errors, or just hard break.
- The sequence of events when a sparse vector is used. The memory usage profile of various types of data.
- What are the removed ETL, how does it change user's code.
- What migration guide is needed.
- What are the upcoming changes.
- What are the difference between scala and pyspark.
- how well does it integrate with SparkML training utilities like CV.

And we will need to have a clear picture for both CPU and GPU. As for internal code structure, i don't have any concern. It looks readable to me and I'm fairly sure that @wbo4958 does a better job than me for JVM based projects.

wbo4958:
Sorry for delay, I was on vacation these days.

wbo4958:
> Some questions about build artifacts and packaging:
> 
> * Do we continue to build two variants of `libxgboost4j.so`, one with CUDA and one without?
[Bobby] Yes, we still need to build tow libxgboost4j.so,
> * Now that `xgboost4j-gpu` package is gone, where to we put `libxgboost4j.so` (CUDA variant)? Does it go inside the `xgboost4j-spark-gpu` package?
[Bobby] the libxgboost4j.so with cuda will reside in xgboost4j package. xgboost4j-spark-gpu now depends on xgboost4j/xgboost4j-spark

> * How long do we plan to distribute Spark 2.12 and 2.13 variants?
[Bobby] what do you mean how long?

wbo4958:
@trivialfis @hcho3, yeah, let's have a online review session.

hcho3:
> libxgboost4j.so with cuda will reside in xgboost4j package

CPU users will complain about the size of xgboost4j package if we choose this route.

wbo4958:
> > libxgboost4j.so with cuda will reside in xgboost4j package
> 
> CPU users will complain about the size of xgboost4j package if we choose this route.

Not exactly. When compiling the CPU versions, there's no cuda things in libxgboost4j.so which will be put into public/nightly places. But for GPU version, we first compile libxgboost4j.so with CUDA and put it into xgboost4j package, but we don't distribute this xgboost4j to the maven repo, instead, we will only distribute xgboost4j-spark-gpu package to the maven, which is a uber package including xgboost4j/xgboost4j-spark.

wbo4958:
Hey @trivialfis @hcho3, CI passed, Could you help review it? Thx

doc/jvm/xgboost_spark_migration.rst
```diff
@@ -0,0 +1,140 @@
...
+      .setRabitTrackerTimeout(100)
+      .setRabitTrackerHostIp("192.168.0.2")
+      .setRabitTrackerPort(19203)
+
+- rabitRingReduceThreshold
+- rabitTimeout
+- rabitConnectRetry
+- singlePrecisionHistogram
+- lambdaBias
+- interactionConstraints
```
trivialfis, line 139: Is this feature interaction?

doc/jvm/xgboost_spark_migration.rst
```diff
@@ -0,0 +1,140 @@
...
+    .setNumRound(5)
+    .setNumWorkers(1)
+    .setFeaturesCol("feature_column")
+    .setLabelCol("label_column")
+    .setEta(1)
+    .setMaxDepth(6)
+    .setObjective("binary:logistic")
+
+*****************
+Unused Parameters
```
trivialfis, line 98: Are these removed? Or they are kept but have no effect hence "unused"?

doc/jvm/xgboost_spark_migration.rst
```diff
@@ -0,0 +1,140 @@
...
+      <artifactId>xgboost-spark_${scala.binary.version}</artifactId>
+      <version>3.0.0</version>
+  </dependency>
+
+When submitting the XGBoost application to the Spark cluster, you only need to specify the single `xgboost-spark` package.
+
+.. code-block:: bash
+
+  spark-submit \
+    --jars xgboost-spark_2.12-3.0.0.jar \
```
trivialfis, line 31: Do you want to mention the GPU package as well?

doc/jvm/index.rst
```diff
@@ -38,6 +38,7 @@ Contents
   XGBoost4J-Spark-GPU Tutorial <xgboost4j_spark_gpu_tutorial>
```
trivialfis, line 38: Do we need to change this tutorial? In the offline review, there's a breaking change for the jar names.

doc/jvm/xgboost_spark_migration.rst
```diff
@@ -0,0 +1,140 @@
+######################################################
+Migration Guide: How to migrate from XGBoost Spark 3.x
+######################################################
+
+XGBoost Spark underwent significant modifications beginning with version 3.0,
```
trivialfis, line 5:
```suggestion
XGBoost Spark underwent significant modifications in version 3.0,
```

doc/jvm/xgboost_spark_migration.rst
```diff
@@ -0,0 +1,140 @@
...
+which may cause compatibility issues with existing user code.
+
+This guide will walk you through the process of updating your code to ensure
+it's compatible with XGBoost Spark 3.0 and later versions.
+
+**********************
+XGBoost Spark Packages
+**********************
+
+XGBoost Spark 3.0 introduced a single uber package named xgboost-spark_2.12-3.0.0.jar, which bundles
```
trivialfis, line 15: not entirely sure whether it's necessary to be an Uber package or the implication of this.

doc/jvm/xgboost_spark_migration.rst
```diff
@@ -0,0 +1,140 @@
...
+  val regressor = new XGBoostRegressor().setObjective("rank:ndcg")
+
+  // after 3.0
+  val ranker = new XGBoostRanker()
+
+******************************
+XGBoost Constructor Parameters
+******************************
+
+XGBoost Spark now categorizes parameters into two groups: XGBoost-Spark parameters and XGBoost parameters.
```
trivialfis, line 54: Where can I find which parameter belongs to the spark parameter group?

doc/jvm/xgboost_spark_migration.rst
```diff
@@ -0,0 +1,140 @@
...
+When constructing an XGBoost estimator, only XGBoost-specific parameters are permitted. XGBoost-Spark specific
```
trivialfis, line 55: What's the motivation for this?

trivialfis (commented):
Could you please help address the implication for the following items:
- Remove any ETL operations in XGBoost

doc/jvm/xgboost_spark_migration.rst
wbo4958: They're removed.

doc/jvm/xgboost_spark_migration.rst
wbo4958: Good catch, I will add it back.

doc/jvm/xgboost_spark_migration.rst
wbo4958: Yes, Good idea.

doc/jvm/index.rst
wbo4958: I will have a following PR for the whole jvm doc updating.

doc/jvm/xgboost_spark_migration.rst
wbo4958: I think we have resolved this confusion.

doc/jvm/xgboost_spark_migration.rst
wbo4958: no where,  user can use setter functions.

doc/jvm/xgboost_spark_migration.rst
wbo4958: 1. XGBoost-specific parameters can be got from https://xgboost.readthedocs.io/en/stable/parameter.html and we don't have a spark-specific parameters to refer. 
2. (Not a strong reason) The xgboost-specific parameters can be sent to xgboost library without doing filtering, if involving xgboost-spark parameters, we need to add a whiltelist that blocks them to be passed to xgboost library.
3. Spark ML uses setters for all parameters, and don't accept any parameters during constructing.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,299 @@
...
+      }
+      new QuantileDMatrix(colBatchIter, missing, maxBin, nthread)
+    }
+
+    estimator.getEvalDataset().map { evalDs =>
+      val evalProcessed = preprocess(estimator, evalDs)
+      ColumnarRdd(train.toDF()).zipPartitions(ColumnarRdd(evalProcessed.toDF())) {
+        (trainIter, evalIter) =>
+          val trainDM = buildQuantileDMatrix(trainIter)
+          val evalDM = buildQuantileDMatrix(evalIter)
```
trivialfis, line 149: You can't build a quantile DMatrix for evaluation without a reference DMatrix.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,299 @@
...
+    val rdd = ColumnarRdd(dataset.asInstanceOf[DataFrame]).mapPartitions { tableIters =>
+      // booster is visible for all spark tasks in the same executor
+      val booster = bBooster.value
+      val originalSchema = bOriginalSchema.value
+
+      // UnsafeProjection is not serializable so do it on the executor side
+      val toUnsafe = UnsafeProjection.create(originalSchema)
+
+      synchronized {
+        val device = booster.getAttr("device")
```
trivialfis, line 183: Not sure why `device` is an attribute of the booster, we usually only have things like `best_iteration` as an attribute.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,299 @@
...
+        if (device != null && device.trim.isEmpty) {
+          booster.setAttr("device", "cuda")
```
trivialfis, line 185: Don't set it to attribute, it goes into the model file, which contains runtime independent model components like trees.

jvm-packages/xgboost4j-spark-gpu/src/test/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPluginSuite.scala
```diff
@@ -0,0 +1,515 @@
...
+        "objective" -> "binary:logistic"
+      )
+      val features = Array("c1", "c2")
+      val estimator = new XGBoostClassifier(xgbParams)
+        .setFeaturesCol(features)
+        .setMissing(0.2f)
+        .setAlpha(0.97)
+        .setLeafPredictionCol("leaf")
+        .setContribPredictionCol("contrib")
+        .setNumRound(1)
```
trivialfis, line 53: Is there a get num rounds in this test?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/Utils.scala
```diff
@@ -14,19 +14,57 @@
...
+  }
+
+  private[spark] implicit class MLVectorToXGBLabeledPoint(val v: Vector) extends AnyVal {
+    /**
+     * Converts a [[Vector]] to a data point with a dummy label.
+     *
+     * This is needed for constructing a [[ml.dmlc.xgboost4j.scala.DMatrix]]
+     * for prediction.
+     */
+    // TODO support sparsevector
```
trivialfis, line 52: I see `SparseVector` being matched in the next statement, is it not supported?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostEstimator.scala
```diff
@@ -0,0 +1,622 @@
...
+ * Indicate what to be predicted
+ *
+ * @param predLeaf    predicate leaf
+ * @param predContrib predicate contribution
+ * @param predRaw     predicate raw
+ * @param predTmp     predicate probability for classification, and raw for regression
+ */
+private[spark] case class PredictedColumns(
+    predLeaf: Boolean,
+    predContrib: Boolean,
```
trivialfis, line 414: Does spark support shap feature interaction?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/TreeBoosterParams.scala
```diff
@@ -0,0 +1,238 @@
...
+    ParamValidators.gt(0))
+
+  final def getMaxCachedHistNode: Int = $(maxCachedHistNode)
+
+  setDefault(eta -> 0.3, gamma -> 0, maxDepth -> 6, minChildWeight -> 1, maxDeltaStep -> 0,
+    subsample -> 1, samplingMethod -> "uniform", colsampleBytree -> 1, colsampleBylevel -> 1,
+    colsampleBynode -> 1, lambda -> 1, alpha -> 0, treeMethod -> "auto", scalePosWeight -> 1,
+    processType -> "default", growPolicy -> "depthwise", maxLeaves -> 0, maxBins -> 256,
+    numParallelTree -> 1, maxCachedHistNode -> 65536)
+
```
trivialfis, line 229: Is it necessary to copy the default here? Is it possible to have an Optional type that indicates not-set and let xgboost handle the default?

trivialfis (commented):
Could you please help address the implication for the following items:
- Remove any ETL operations in XGBoost

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/TreeBoosterParams.scala
wbo4958: This is just default value and if user doesn't set them explicitly, they will not be passed to xgboost.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostEstimator.scala
wbo4958: As far as I know, the answer is no.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/Utils.scala
wbo4958: The SparseVector is still supported, and it will be converted to dense to build DMatrix

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
wbo4958: The attribute is just a flag which is used to indicate if the "device" has been set by booster to avoid duplicating set.

trivialfis: Can the normal parameter be used as the flag?

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
wbo4958: the booster is a global variable for all transform tasks. Just utilize this for checking if device has been set. Not a big deal.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
wbo4958: what does that mean? 

The original is doing so.

trivialfis: https://github.com/dmlc/xgboost/blob/7bccc1ea2c81f7532483064af94ac8c7aba07828/tests/python/test_quantile_dmatrix.py#L189

wbo4958:
> Could you please help address the implication for the following items:
> 
> * Remove any ETL operations in XGBoost


For example, 

ETL includes "caching dataset", and collect group for ranking issue.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/TreeBoosterParams.scala
trivialfis: So, we can set arbitrary values here?

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
trivialfis: Can the normal parameter be used instead?

wbo4958: It will be so complicated if using normal parameter which involves distributed multi-tasks (multi-thread). This is the simplest way to achieve the goal.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
wbo4958: What is Xy here? Is pyspark doing this?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/Utils.scala
trivialfis: Is the missing value being filled with 0s? If so, what happens if the 0s indeed represent missing values? Also, this is a breaking change I believe, which should be part of the migration guide.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
wbo4958: see the above comment

jvm-packages/xgboost4j-spark-gpu/src/test/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPluginSuite.scala
wbo4958: Good catch, hmm, I can add it.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/TreeBoosterParams.scala
wbo4958: Yeah, that's true. But those values are the defaults from xgboost python parameters page.

trivialfis: Yeah, we need to "remember" changing this whenever we change the internal, but we always forget, hence the outdated parameter list.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
wbo4958: I just checked, pyspark is also not doing this.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
trivialfis: Why does a getter for the normal parameter involve distributed multi-tasks? Just get the Json configuration.

doc/jvm/xgboost_spark_migration.rst
wbo4958: Done

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
trivialfis: The `QDM` needs the training data as a reference to obtain the valid quantile for the evaluation dataset. Otherwise, the histogram cuts are built using only the evaluation dataset, which is incorrect.

doc/jvm/xgboost_spark_migration.rst
wbo4958: We don't have a page for that, but I think the API docs can have these.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
wbo4958: the code could run on different tasks (multi threads) which will have race condition.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/Utils.scala
wbo4958: No exactly, missing value can be any value including 0. The sparsevector will be converted to densevector, and the missing value will be passed to the xgboost when constructing the DMatrix.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/TreeBoosterParams.scala
wbo4958: If we don't set the default values, the getter function will throw exception. That's really painful.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,299 @@
...
+    /** build QuantileDMatrix on the executor side */
+    def buildQuantileDMatrix(iter: Iterator[Table]): QuantileDMatrix = {
+      val colBatchIter = iter.map { table =>
+        withResource(new GpuColumnBatch(table)) { batch =>
+          new CudfColumnBatch(
+            batch.select(indices.featureIds.get),
+            batch.select(indices.labelId),
+            batch.select(indices.weightId.getOrElse(-1)),
+            batch.select(indices.marginId.getOrElse(-1)),
+            batch.select(indices.groupId.getOrElse(-1)));
```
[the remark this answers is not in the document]
wbo4958: The old JVM is using HasGroupCol, so that's the history.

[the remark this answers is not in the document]
trivialfis: That's not the same as `qid` right?

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/Utils.scala
trivialfis: Yes, during the conversion from sparse to dense vector, what value is being filled into the missing entries of the input sparse vector?

trivialfis:
@WeichenXu123 @rongou  @Craigacp It would be great if we could get some help reviewing the PR when you are available.

rongou (approved):
Not too familiar with this part of the code base, but overall looks good.

Craigacp:
I haven't written any Spark or Scala code since Spark 1.4 so I don't think I'll have much ability to review this PR.

dotbg:
I've started reviewing the spark-related code, but it takes a while

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
trivialfis: This might help https://github.com/dmlc/xgboost/pull/10692

wbo4958: Thx, I will fix it.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
wbo4958: I mean,  we need a **global** variable which should not be rewritten by task. the **booster** happens to be a global one, so here we use it. Or else, we need to create another global variable by blockmanager.

trivialfis: I mean using the get JSON config from the booster instead of using the get attr.

- get config:
https://github.com/dmlc/xgboost/blob/773ded684b15fb3381c73f326cddbf6e5f81dccf/src/c_api/c_api.cc#L982
- get attr:
https://github.com/dmlc/xgboost/blob/773ded684b15fb3381c73f326cddbf6e5f81dccf/src/c_api/c_api.cc#L1569

The device is stored in the configuration and it's thread-safe to get it. On the other hand, the `attr` is preserved for booster properties instead of runtime parameter.

terrytangyuan (commented):
@hcho3 Thanks for mentioning this PR to me. Unfortunately I haven't done much serious Scala dev for a long time so I'll defer the review to others.

dotbg (commented):
I left a few comments, but still have rather long way to go further

doc/jvm/index.rst
wbo4958: It's Done, Mark as resolved.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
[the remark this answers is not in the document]
wbo4958: I think it's the same with qid

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
wbo4958: Done. fixed.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/params/ParamMapConversion.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.spark.ml.param._
+
+private[spark] trait ParamMapConversion extends NonXGBoostParams {
+
+  /**
+   * Convert XGBoost parameters to Spark Parameters
+   *
+   * @param xgboostParams XGBoost style parameters
+   */
+  def xgboost2SparkParams(xgboostParams: Map[String, Any]): Unit = {
```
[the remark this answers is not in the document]
wbo4958: Yes, in XGBoostEstimatorSuite

``` scala
class XGBoostEstimatorSuite extends AnyFunSuite with PerTest with TmpFolderPerSuite {

  test("params") {
    val df = smallBinaryClassificationVector
    val xgbParams: Map[String, Any] = Map(
      "max_depth" -> 5,
      "eta" -> 0.2,
      "objective" -> "binary:logistic"
    )
    val estimator = new XGBoostClassifier(xgbParams)
      .setFeaturesCol("features")
      .setMissing(0.2f)
      .setAlpha(0.97)
      .setLeafPredictionCol("leaf")
      .setContribPredictionCol("contrib")
      .setNumRound(1)

    assert(estimator.getMaxDepth === 5)
    assert(estimator.getEta === 0.2)
    assert(estimator.getObjective === "binary:logistic")
    assert(estimator.getFeaturesCol === "features")
    assert(estimator.getMissing === 0.2f)
    assert(estimator.getAlpha === 0.97)

    estimator.setEta(0.66).setMaxDepth(7)
    assert(estimator.getMaxDepth === 7)
    assert(estimator.getEta === 0.66)

    val model = estimator.train(df)
    assert(model.getMaxDepth === 7)
    assert(model.getEta === 0.66)
    assert(model.getObjective === "binary:logistic")
    assert(model.getFeaturesCol === "features")
    assert(model.getMissing === 0.2f)
    assert(model.getAlpha === 0.97)
    assert(model.getLeafPredictionCol === "leaf")
    assert(model.getContribPredictionCol === "contrib")
  }
```

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/QuantileDMatrix.scala
```diff
@@ -0,0 +1,112 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+package ml.dmlc.xgboost4j.scala
+
+import _root_.scala.collection.JavaConverters._
```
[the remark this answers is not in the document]
wbo4958: Done. Thx for it.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,299 @@
...
+    val conf = dataset.sparkSession.conf
+    val hasRapidsPlugin = conf.get("spark.sql.extensions", "").split(",").contains(
+      "com.nvidia.spark.rapids.SQLExecPlugin")
+    val rapidsEnabled = try {
+      conf.get("spark.rapids.sql.enabled").toBoolean
+    } catch {
+      // Rapids plugin has default "spark.rapids.sql.enabled" to true
+      case _: NoSuchElementException => true
+      case _: Throwable => false // Any exception will return false
+    }
```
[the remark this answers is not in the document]
wbo4958: The original will set the rapidsEnable to false when it's not NoSuchElementException. But the suggested way will set it as true.

wbo4958:
Hi @hcho3 @trivialfis, Could you help review it again. Thx

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,302 @@
...
+        new QuantileDMatrix(colBatchIter, missing, maxBin, nthread)
+      )
+    }
+
+    estimator.getEvalDataset().map { evalDs =>
+      val evalProcessed = preprocess(estimator, evalDs)
+      ColumnarRdd(train.toDF()).zipPartitions(ColumnarRdd(evalProcessed.toDF())) {
+        (trainIter, evalIter) =>
+          val trainDM = buildQuantileDMatrix(trainIter)
+          val evalDM = buildQuantileDMatrix(evalIter)
```
trivialfis, line 152: This is still not using the reference matrix right?

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
trivialfis: any update?

jvm-packages/xgboost4j-spark-gpu/src/test/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPluginSuite.scala
trivialfis: any update?

doc/jvm/xgboost_spark_migration.rst
```diff
@@ -0,0 +1,162 @@
...
+        <version>3.0.0</version>
+    </dependency>
+
+* For GPU
+
+  .. code-block:: xml
+
+    <dependency>
+        <groupId>ml.dmlc</groupId>
+        <artifactId>xgboost-spark-gpu_${scala.binary.version}</artifactId>
```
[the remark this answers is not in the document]
wbo4958: Yeah, we still need it, since we will have two uber jars, one for CPU without gpu stuffs, the other is for GPU with gpu stuffs.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
wbo4958: Done, I didn't use XGBoosterGetAttr and XGBoosterSaveJsonConfig. I just add a flag in booster to check if device is set. that's really simple.

jvm-packages/xgboost4j-spark-gpu/src/test/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPluginSuite.scala
wbo4958: Done.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/Utils.scala
wbo4958: Done.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
wbo4958: Good finding. Done.

wbo4958:
Hi @trivialfis, I've fixed all the comments.

trivialfis:
Where is the definition of xgboost ranker?

wbo4958:
@trivialfis, I will add it in the following PR, I didn't want to blow up this PR.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,303 @@
...
+        withResource(new GpuColumnBatch(table)) { batch =>
+          new CudfColumnBatch(
+            batch.select(indices.featureIds.get),
+            batch.select(indices.labelId),
+            batch.select(indices.weightId.getOrElse(-1)),
+            batch.select(indices.marginId.getOrElse(-1)),
+            batch.select(indices.groupId.getOrElse(-1)));
+        }
+      }
+      ref.map(r => new QuantileDMatrix(colBatchIter, r, missing, maxBin, nthread)).getOrElse(
```
eordentlich, line 142: Does quantiledmatrix construction require communication between workers to determine the bin boundaries and where is that set up?

eordentlich (commented):
Still have more to go through but here are a batch of comments (some questions).

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
trivialfis: Yes, there are a couple of allreduces for the data shape and the quantiles.

wbo4958: Yes, it needs the communication, and it happened in the xgboost library.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,302 @@
...
+                } else {
+                  try {
+                    currentBatch = new ColumnarBatch(
+                      GpuColumnVectorUtils.extractHostColumns(table, dataTypes),
+                      table.getRowCount().toInt)
+                    val rowIterator = currentBatch.rowIterator().asScala.map(toUnsafe)
+                      .map(converter(_))
+                    model.predictInternal(booster, dm, pred, rowIterator).toIterator
+                  } finally {
+                    dm.delete()
```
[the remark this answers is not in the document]
wbo4958: Yes, XGBoost will return all the prediction values on the input dm.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,302 @@
...
+                if (dm == null) {
+                  Iterator.empty
+                } else {
+                  try {
+                    currentBatch = new ColumnarBatch(
+                      GpuColumnVectorUtils.extractHostColumns(table, dataTypes),
+                      table.getRowCount().toInt)
+                    val rowIterator = currentBatch.rowIterator().asScala.map(toUnsafe)
+                      .map(converter(_))
+                    model.predictInternal(booster, dm, pred, rowIterator).toIterator
```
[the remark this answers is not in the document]
wbo4958: the rowIterator includes all the input columns. in predictInternal which will zip the input data frame and the predicted data frame.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,302 @@
...
+
+        def loadNextBatch(): Unit = {
+          closeCurrentBatch()
+          if (tableIters.hasNext) {
+            val dataTypes = originalSchema.fields.map(x => x.dataType)
+            iter = withResource(tableIters.next()) { table =>
+              // Create DMatrix
+              val featureTable = new GpuColumnBatch(table).select(featureIds)
+              if (featureTable == null) {
+                throw new RuntimeException("Something wrong for feature indices")
```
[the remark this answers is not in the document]
wbo4958: Hmm, Good idea. Done.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostPlugin.scala
```diff
@@ -0,0 +1,49 @@
...
+ limitations under the License.
+ */
+package ml.dmlc.xgboost4j.scala.spark
+
+import java.io.Serializable
+
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.{DataFrame, Dataset}
+
+trait XGBoostPlugin extends Serializable {
```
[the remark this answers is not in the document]
wbo4958: When upstreaming the GPU solution to the XGBoost, the reviewer required the XGBoostPlugin to be extended by other companies, other libraries. I think it's reasonable. So we define the XGBoostPlugin and leveraged ServiceLoader to load it. But for now, no other solution for xgboost, so we have restricted it to GpuXGBoostPlugin in XGBoostEstimator.scala

``` scala
  private val plugin: Option[XGBoostPlugin] = {
    val classLoader = Option(Thread.currentThread().getContextClassLoader)
      .getOrElse(getClass.getClassLoader)

    val serviceLoader = ServiceLoader.load(classOf[XGBoostPlugin], classLoader)

    // For now, we only trust GpuXGBoostPlugin.
    serviceLoader.asScala.filter(x => x.getClass.getName.equals(
      "ml.dmlc.xgboost4j.scala.spark.GpuXGBoostPlugin")).toList match {
      case Nil => None
      case head :: Nil =>
        Some(head)
      case _ => None
    }
  }
```

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/QuantileDMatrix.scala
```diff
@@ -36,6 +36,27 @@ class QuantileDMatrix private[scala](
     this(new JQuantileDMatrix(iter.asJava, missing, maxBin, nthread))
   }
 
+  /**
+   * Create QuantileDMatrix from iterator based on the array interface
+   *
+   * @param iter       the XGBoost ColumnBatch batch to provide the corresponding array interface
+   * @param refDMatrix The training dataset that provides quantile information, needed
```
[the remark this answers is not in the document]
wbo4958: Good catch. Done.

jvm-packages/xgboost4j-spark-gpu/src/main/java/ml/dmlc/xgboost4j/java/QuantileDMatrix.java
```diff
@@ -35,11 +35,39 @@ public QuantileDMatrix(
...
       int nthread) throws XGBoostError {
+    this(iter, null, missing, maxBin, nthread);
+  }
+
+  /**
+   * Create QuantileDMatrix from iterator based on the cuda array interface
+   *
+   * @param iter       the XGBoost ColumnBatch batch to provide the corresponding cuda array
+   *                   interface
+   * @param refDMatrix The training dataset that provides quantile information, needed when
```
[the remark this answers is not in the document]
wbo4958: Good catch. Done.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,299 @@
...
+    // TODO support array/vector feature
+    estimator.getFeaturesCols.foreach { name =>
+      val col = estimator.castIfNeeded(dataset.schema, name)
+      selectedCols.append(col)
+    }
+    val input = dataset.select(selectedCols.toArray: _*)
+    estimator.repartitionIfNeeded(input)
+  }
+
+  // visiable for testing
```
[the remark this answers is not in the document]
wbo4958: Good catch. Done.

doc/jvm/xgboost_spark_migration.rst
```diff
@@ -0,0 +1,163 @@
+######################################################
+Migration Guide: How to migrate from XGBoost Spark 3.x
```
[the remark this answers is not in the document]
wbo4958: changed it to XGBoost4j-Spark 3.x

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
```diff
@@ -431,126 +213,97 @@ object XGBoost extends XGBoostStageLevel {
...
-          var optionWatches: Option[() => Watches] = None
-
-          // take the first Watches to train
-          if (iter.hasNext) {
-            optionWatches = Some(iter.next())
+      val rabitEnv = tracker.getWorkerArgs()
+
+      val boostersAndMetrics = input.barrier().mapPartitions { iter =>
+        require(iter.hasNext, "Couldn't get DMatrix")
+        val watches = iter.next()
```
eordentlich, line 265: It looks like QuantileDMatrix is computed here, but the Communicator is not initialized till inside the trainBooster call below.   So still not fully clear if communication is happening for the QuantileDMatrix per response to my previous comment on this.

wbo4958: You're right, it's my bad. the communicator should be initialized before creating quantile dmatrix. I just made sure the local tests on only 1 GPU can pass. This issue definitely needs to be found in multi-GPUs environment.

trivialfis:
I will look into the ranking later. In theory, as long as the partitions for each local workers are sorted, the gradient is considered correct.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala
wbo4958: The newest commit has fixed this issue. Thx.

doc/jvm/xgboost_spark_migration.rst
```diff
@@ -0,0 +1,162 @@
+########################################################
+Migration Guide: How to migrate to XGBoost-Spark jvm 3.x
+########################################################
+
+XGBoost-Spark jvm packages underwent significant modifications in version 3.0,
+which may cause compatibility issues with existing user code.
+
+This guide will walk you through the process of updating your code to ensure
+it's compatible with XGBoost Spark 3.0 and later versions.
```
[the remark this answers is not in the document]
wbo4958: Done.

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,302 @@
...
+      val originalSchema = bOriginalSchema.value
+
+      // UnsafeProjection is not serializable so do it on the executor side
+      val toUnsafe = UnsafeProjection.create(originalSchema)
+
+      synchronized {
+        val device = booster.getAttr("device")
+        if (device != null && device.trim.isEmpty) {
+          booster.setAttr("device", "cuda")
+          val gpuId = if (!isLocal) XGBoost.getGPUAddrFromResources else 0
```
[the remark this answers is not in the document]
wbo4958: The new commit has changed it to 

``` scala
    if (runtimeParams.runOnGpu) {
      val gpuId = if (runtimeParams.isLocal) {
        TaskContext.get().partitionId() % runtimeParams.numWorkers
      } else {
        getGPUAddrFromResources
      }
      logger.info("Leveraging gpu device " + gpuId + " to train")
      params = params + ("device" -> s"cuda:$gpuId")
    }
```

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,302 @@
...
+
+    validate(estimator, dataset)
+
+    val train = preprocess(estimator, dataset)
+    val schema = train.schema
+
+    val indices = estimator.buildColumnIndices(schema)
+
+    val maxBin = estimator.getMaxBins
+    val nthread = estimator.getNthread
```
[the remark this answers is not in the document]
wbo4958: Yes.

jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/TrainTestData.scala
```diff
@@ -65,18 +66,40 @@ trait TrainTestData {
 object Classification extends TrainTestData {
   val train: Seq[XGBLabeledPoint] = getLabeledPoints("/agaricus.txt.train", 126, zeroBased = false)
   val test: Seq[XGBLabeledPoint] = getLabeledPoints("/agaricus.txt.test", 126, zeroBased = false)
+
+  Random.setSeed(10)
+  val randomWeights = Array.fill(train.length)(Random.nextFloat())
+  val trainWithWeight = train.zipWithIndex.map { case (v, index) =>
+    XGBLabeledPoint(v.label, v.size, v.indices, v.values,
+      randomWeights(index), v.group, v.baseMargin)
```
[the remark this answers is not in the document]
wbo4958: There're some tests that need the same weights to build DMatrix directly instead of building Dataframe

``` scala
  test("XGBoost-Spark binary classification output with weight should match XGBoost4j") {
    val trainingDM = new DMatrix(Classification.trainWithWeight.iterator)
    trainingDM.setWeight(Classification.randomWeights)
    val trainingDF = buildDataFrame(Classification.trainWithWeight)
    val testDF = buildDataFrame(Classification.test)
    val paramMap = Map("objective" -> "binary:logistic")
    checkResultsWithXGBoost4j(trainingDM, testDM, trainingDF, testDF,
      5, paramMap, Some("weight"))
  }
```

jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/GpuXGBoostPlugin.scala
```diff
@@ -0,0 +1,302 @@
...
+  /**
+   * Whether the plugin is enabled or not, if not enabled, fallback
+   * to the regular CPU pipeline
+   *
+   * @param dataset the input dataset
+   * @return Boolean
+   */
+  override def isEnabled(dataset: Dataset[_]): Boolean = {
+    val conf = dataset.sparkSession.conf
+    val hasRapidsPlugin = conf.get("spark.sql.extensions", "").split(",").contains(
```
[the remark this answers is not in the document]
wbo4958: Yes, good suggestion. Done. Thx

trivialfis:
Huge thanks to @dotbg @eordentlich for the reviews and for catching some issues early. I wouldn't be able to do that myself. Could you please help approve the PR for spark hanlding?

I don't have any more comments about logic related to XGBoost itself.

eordentlich (commented):
Fixes looks great.   2 more minor comments.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressor.scala
```diff
@@ -16,405 +16,91 @@
...
-      super.transformSchema(schema)
-    } else {
-      transformSchemaWithFeaturesCols(false, schema)
+  /**
+   * Validate the parameters before training, throw exception if possible
+   */
+  override protected[spark] def validate(dataset: Dataset[_]): Unit = {
+    super.validate(dataset)
+
+    // If the objective is set explicitly, it must be in binaryClassificationObjs and
```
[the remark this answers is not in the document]
wbo4958: Done

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostEstimator.scala
```diff
@@ -0,0 +1,641 @@
...
+    var predTmp = false
+    this match {
+      case p: XGBProbabilisticClassifierParams[_] => // classification case
+        predRaw = addToSchema(p.rawPredictionCol)
+        predTmp = addToSchema(p.probabilityCol, Some(TMP_TRANSFORMED_COL))
+
+        if (isDefinedNonEmpty(predictionCol)) {
+          // Let's use transformed col to calculate the prediction
+          if (!predTmp) {
+            // Add the transformed col for predition
```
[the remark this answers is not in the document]
wbo4958: Thx, Done.

eordentlich (approved):
👍

wbo4958:
Hi @trivialfis, could you help merge it?

trivialfis (approved):
Thank you for the nice rework! Looking forward to the complete implementation.

PR #10642, opened 2024-07-27
Allow external configuration of endianness in R package build

jakirkham:
Rework R package's `configure.ac` to accept and use the `USE_LITTLE_ENDIAN` environment variable. If set, this allows overriding the endianness used for the R package built. If not set, then this falls back to `autoconf`'s own endianness check. If it is still not clear what endianness to use, the build then errors and asks for it to be specified.

For more context conda-forge builds have needed to [patch out the existing configuration logic for endianness and set their own value]( https://github.com/conda-forge/xgboost-feedstock/blob/44f54392837b13980c3a0d68d7b03c78b041b3c6/recipe/patches/0001-Force-endian-flag-in-cross-compilation-mode.patch#L1 ) as the current check isn't always correct. This would allow the conda-forge builds to drop that patch and simply specify `USE_LITTLE_ENDIAN` when needed (a more maintainable solution).


jakirkham:
Noticed a couple failures here. In looking into them, am curious does this go back to `dmlc-core`. Do we need to update `dmlc-core` as well?

On a somewhat related note, it appears `dmlc-core` is using some deprecated endianness logic. More details in this issue: https://github.com/dmlc/dmlc-core/issues/688

hcho3:
Added `AC_CONFIG_HEADERS()` macro to avoid the warning
```
configure.ac:87: warning: AC_C_BIGENDIAN should be used with AC_CONFIG_HEADERS
```
This warning causes the R build to fail: https://github.com/dmlc/xgboost/actions/runs/10152770673/job/28074838045?pr=10642

hcho3 (approved):

jakirkham:
Thanks Hyunsu! 🙏

Issue #5386, opened 2020-03-03
Allow python/R bindings to use an external libxgboost


dpryan79:
Would it be possible to change the R/python bindings so they can use precompiled xgboost libraries? In conda, as an example, we distribute the standard libraries separate from the python and R bindings, since users rarely need all of them. An example patch for the python bindings can be found here, though it should be modified to make it more generic: https://github.com/conda-forge/xgboost-feedstock/blob/master/recipe/0001-conda-Unbundle-libxgboost.-dll-dylib-so.patch

Xref: https://github.com/conda-forge/xgboost-feedstock/pull/47

trivialfis:
It would be very useful for distribution purpose.  I rewrote the Python setup.py script recently, would you like to take a look and see how can we find the library dynamically?

trivialfis:
There are specializations in R package to make it comply to CRAN policy, for examples `stdout/stderr` is forbidden, some math functions are replaced.  Particularly CMake is no usable.

trivialfis:
For Python package this is done in: https://github.com/dmlc/xgboost/pull/6362

jakirkham:
Awesome! Thank you Jiaming 😄

dpryan79:
Thanks!

jakirkham:
Handling the last bit of the associated conda-forge patch in PR: https://github.com/dmlc/xgboost/pull/10643

PR #10643, opened 2024-07-27
Add `Library\mingw-w64` to Windows search path

jakirkham:
Fixes https://github.com/dmlc/xgboost/issues/5386
Fixes https://github.com/conda-forge/xgboost-feedstock/issues/51

When searching for `libxgboost` on Windows, this adds `Library\mingw-w64` to the search path after other paths

This currently is done as [a patch in conda-forge]( https://github.com/conda-forge/xgboost-feedstock/blob/44f54392837b13980c3a0d68d7b03c78b041b3c6/recipe/patches/0003-Use-mingw-w64-path.patch#L11-L35 ). Given this is just checking one more location (and after checking the usual ones), though it made sense to go ahead and upstream this one

### DIFF
--- a/python-package/packager/nativelib.py
+++ b/python-package/packager/nativelib.py
@@ -138,6 +138,9 @@ def locate_or_build_libxgboost(
             sys_prefix / "Library",
             sys_prefix / "Library" / "bin",
             sys_prefix / "Library" / "lib",
+            sys_prefix / "Library" / "mingw-w64",
+            sys_prefix / "Library" / "mingw-w64" / "bin",
+            sys_prefix / "Library" / "mingw-w64" / "lib",
         ]
         sys_prefix_candidates = [
             p.expanduser().resolve() for p in sys_prefix_candidates
--- a/python-package/xgboost/libpath.py
+++ b/python-package/xgboost/libpath.py
@@ -38,6 +38,9 @@ def find_lib_path() -> List[str]:
                 os.path.join(sys.base_prefix, "Library"),
                 os.path.join(sys.base_prefix, "Library", "bin"),
                 os.path.join(sys.base_prefix, "Library", "lib"),
+                os.path.join(sys.base_prefix, "Library", "mingw-w64"),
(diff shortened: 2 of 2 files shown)


python-package/packager/nativelib.py
```diff
@@ -138,6 +138,9 @@ def locate_or_build_libxgboost(
             sys_prefix / "Library",
             sys_prefix / "Library" / "bin",
             sys_prefix / "Library" / "lib",
+            sys_prefix / "Library" / "mingw-w64",
+            sys_prefix / "Library" / "mingw-w64" / "bin",
```
trivialfis, line 142: Could you please share under which condition libxgboost is on the bin path? Also, is mingw ever used for compiling the Python package? Asking since mingw has/had a broken Unix socket implementation, hence distributed computing is disabled there.

hcho3: Conda-forge uses Visual Studio to build the Python package. See the logs at https://github.com/conda-forge/xgboost-feedstock/runs/27951580159.

@jakirkham I am inclined to exclude this patch from the upstream because the use of `mingw-w64` appears to be a convention specific to Conda-forge and not generally applicable to all users.

trivialfis: If conda forge uses visual studio, maybe we can simply drop the patch from conda forge as well?

hcho3: No, Conda-forge needs this patch. Libraries are installed under the mingw-w64 directory. I tried removing it and got errors.

trivialfis: In the bin directory?

hcho3: Yes. The patch is a product of trial and error. I had to add it to get https://github.com/conda-forge/xgboost-feedstock/pull/132 working.

trivialfis: That's unexpected. Thank you for sharing. I'm fine with merging it then.

jakirkham: Thank you both! 🙏

PR #10649, opened 2024-07-30
[CI] Fix Python wheel workflow.

trivialfis:


tests/ci_build/build_python_wheels.sh
```diff
@@ -15,13 +15,13 @@ if [[ "$platform_id" == macosx_* ]]; then
     if [[ "$platform_id" == macosx_arm64 ]]; then
         # MacOS, Apple Silicon
         wheel_tag=macosx_12_0_arm64
-        cpython_ver=39
+        cpython_ver=310
```
trivialfis, line 18: We can move this particular configuration out of the script and let the caller decide the version.

PR #10657, opened 2024-07-31
Update the release script to download xgboost-cpu

hcho3:
Download three wheels for `xgboost-cpu`.


trivialfis (approved):

trivialfis:
Opened an issue about the type error: https://github.com/python/mypy/issues/17617 Not familiar with async type enough to make a fix yet, but appears to be a bug in mypy as it should recognize something other than the `None` type.

hcho3:
Merging it now to unblock the CI.

PR #10677, opened 2024-08-05
Cleanup GPU Hist tests.

trivialfis:
- Remove GPU Hist gradient sampling test. The same properties are tested in the gradient sampler test suite.
- Move basic histogram tests into the histogram test suite.
- Remove the header inclusion of the `updater_gpu_hist.cu` in tests.

Follow up of #10608


trivialfis:
cc @rongou .

rongou (approved):

PR #10678, opened 2024-08-05
[dask] Reduce the flakiness of tests.

trivialfis:
- Use a dedicated cluster with a single worker.


trivialfis:
I can't reproduce the error on my device. Will try to loosen the test instead.

trivialfis:
Dask is moving partitions between workers even though the size of each partition is held fixed.

trivialfis:
The old repartition script was broken after `dask-expr`, I'm temping to use a single worker for this test.

hcho3 (approved):

PR #10680, opened 2024-08-06
Added SVE implementation to improve the performance on ARM architecture

divya2108:
**Motivation**:  This pull request aims to improve the performance of training algorithm of XGBoost on **ARM** architecture by leveraging **SVE** intrinsics.

**Brief** **description**:
1. This change of including SVE intrinsics improves the performance by **55**% as compared to the ARM default. 
2. The modified function iterates over a row of data and updates a histogram based on the given indices and offsets.
3. The accuracy has been verified after the modifications.

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,12 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
   endif()
 endif()
 
+include(${xgboost_SOURCE_DIR}/cmake/CheckSVEsupport.cmake)
+check_xgboost_sve_support()
+if(XGBOOST_COMPILER_HAS_ARM_SVE)
+    add_compile_definitions(XGBOOST_SVE_COMPILER_SUPPORT)
+endif()
+
 include(${xgboost_SOURCE_DIR}/cmake/PrefetchIntrinsics.cmake)
 find_prefetch_intrinsics()
 include(${xgboost_SOURCE_DIR}/cmake/Version.cmake)
@@ -353,13 +359,6 @@ target_include_directories(xgboost
   $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>)
 #-- End shared library
-
-include(${xgboost_SOURCE_DIR}/cmake/CheckSVEsupport.cmake)
-check_xgboost_sve_support()
-if(XGBOOST_ARM_SVE_HARDWARE_SUPPORT)
-    target_compile_definitions(objxgboost PUBLIC XGBOOST_SVE_SUPPORT_DETECTED)
-    target_compile_options(objxgboost PRIVATE ${XGBOOST_SVE_FLAGS})
-endif()
 
 #-- CLI for xgboost
 if(BUILD_DEPRECATED_CLI)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,9 +40,6 @@ endif()
 
 include(${xgboost_SOURCE_DIR}/cmake/CheckSVEsupport.cmake)
 check_xgboost_sve_support()
-if(XGBOOST_COMPILER_HAS_ARM_SVE)
-    add_compile_definitions(XGBOOST_SVE_COMPILER_SUPPORT)
-endif()
 
 include(${xgboost_SOURCE_DIR}/cmake/PrefetchIntrinsics.cmake)
 find_prefetch_intrinsics()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -270,6 +270,51 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "OS400")
   set(CMAKE_CXX_ARCHIVE_CREATE "<CMAKE_AR> -X64 qc <TARGET> <OBJECTS>")
 endif()
 
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+  include(CheckCSourceCompiles)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=armv8-a+sve")
+  check_c_source_compiles("
+  #if defined(__aarch64__) && defined(__ARM_FEATURE_SVE)
+  #include <arm_sve.h>
+  int main() {
+    svfloat64_t a;
+    a = svdup_n_f64(0);
+    return 0;
+  }
+  #endif
+  " COMPILER_HAS_ARM_SVE)
+
+  if(COMPILER_HAS_ARM_SVE)
+    message(STATUS "ARM SVE compiler support detected")
+    set(SOURCE_CODE "
+    #include <sys/prctl.h>
+    int main() {
+      int ret = prctl(PR_SVE_GET_VL);
+      return ret >= 0 ? 0 : 1;
+    }
+    ")
+    file(WRITE ${CMAKE_BINARY_DIR}/check_sve_support.c "${SOURCE_CODE}")
+    try_run(RUN_RESULT COMPILE_RESULT
+      ${CMAKE_BINARY_DIR}/check_sve_support_output
+      ${CMAKE_BINARY_DIR}/check_sve_support.c
+    )
+
+    if(RUN_RESULT EQUAL 0)
+      message(STATUS "ARM SVE hardware support detected")
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=armv8-a+sve")
+      string(APPEND CMAKE_CXX_FLAGS " -DSVE_SUPPORT_DETECTED")
+    else()
+      message(STATUS "ARM SVE hardware support not detected")
+    endif()
+  else()
+    message(STATUS "ARM SVE compiler support not detected")
+  endif()
+
+  set(CMAKE_C_FLAGS "${ORIGINAL_CMAKE_C_FLAGS}")
+else()
+  message(STATUS "Not an aarch64 architecture")
+endif()
+
 if(USE_NCCL)
   find_package(Nccl REQUIRED)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -270,51 +270,6 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "OS400")
   set(CMAKE_CXX_ARCHIVE_CREATE "<CMAKE_AR> -X64 qc <TARGET> <OBJECTS>")
 endif()
 
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
-  include(CheckCSourceCompiles)
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=armv8-a+sve")
-  check_c_source_compiles("
-  #if defined(__aarch64__) && defined(__ARM_FEATURE_SVE)
-  #include <arm_sve.h>
-  int main() {
-    svfloat64_t a;
-    a = svdup_n_f64(0);
-    return 0;
-  }
-  #endif
-  " COMPILER_HAS_ARM_SVE)
-
-  if(COMPILER_HAS_ARM_SVE)
-    message(STATUS "ARM SVE compiler support detected")
-    set(SOURCE_CODE "
-    #include <sys/prctl.h>
-    int main() {
-      int ret = prctl(PR_SVE_GET_VL);
-      return ret >= 0 ? 0 : 1;
-    }
-    ")
-    file(WRITE ${CMAKE_BINARY_DIR}/check_sve_support.c "${SOURCE_CODE}")
-    try_run(RUN_RESULT COMPILE_RESULT
-      ${CMAKE_BINARY_DIR}/check_sve_support_output
-      ${CMAKE_BINARY_DIR}/check_sve_support.c
-    )
-
-    if(RUN_RESULT EQUAL 0)
-      message(STATUS "ARM SVE hardware support detected")
-      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=armv8-a+sve")
-      string(APPEND CMAKE_CXX_FLAGS " -DSVE_SUPPORT_DETECTED")
-    else()
-      message(STATUS "ARM SVE hardware support not detected")
-    endif()
-  else()
-    message(STATUS "ARM SVE compiler support not detected")
-  endif()
-
-  set(CMAKE_C_FLAGS "${ORIGINAL_CMAKE_C_FLAGS}")
-else()
-  message(STATUS "Not an aarch64 architecture")
-endif()
-
 if(USE_NCCL)
   find_package(Nccl REQUIRED)
 endif()
@@ -398,6 +353,13 @@ target_include_directories(xgboost
   $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>)
 #-- End shared library
+
+include(${xgboost_SOURCE_DIR}/cmake/CheckSVEsupport.cmake)
+check_xgboost_sve_support()
+if(XGBOOST_ARM_SVE_HARDWARE_SUPPORT)
+    target_compile_definitions(objxgboost PUBLIC XGBOOST_SVE_SUPPORT_DETECTED)
+    target_compile_options(objxgboost PRIVATE ${XGBOOST_SVE_FLAGS})
+endif()
 
 #-- CLI for xgboost
 if(BUILD_DEPRECATED_CLI)
--- a/cmake/CheckSVEsupport.cmake
+++ b/cmake/CheckSVEsupport.cmake
@@ -1,7 +1,7 @@
 function(check_xgboost_sve_support)
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
     include(CheckCSourceCompiles)
-    
+
     # Save the original C_FLAGS to restore later
     set(ORIGINAL_C_FLAGS "${CMAKE_C_FLAGS}")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=armv8-a+sve")
--- a/cmake/CheckSVEsupport.cmake
+++ b/cmake/CheckSVEsupport.cmake
@@ -1,0 +1,56 @@
+function(check_xgboost_sve_support)
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+    include(CheckCSourceCompiles)
+    
+    # Save the original C_FLAGS to restore later
+    set(ORIGINAL_C_FLAGS "${CMAKE_C_FLAGS}")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=armv8-a+sve")
+
+    # Check if the compiler supports ARM SVE
+    check_c_source_compiles("
+    #if defined(__aarch64__) && defined(__ARM_FEATURE_SVE)
+    #include <arm_sve.h>
+    int main() {
+        svfloat64_t a;
+        a = svdup_n_f64(0);
+        return 0;
+    }
+    #endif
+    " XGBOOST_COMPILER_HAS_ARM_SVE)
+
+    if(XGBOOST_COMPILER_HAS_ARM_SVE)
+        message(STATUS "ARM SVE compiler support detected")
+
+        # Check for hardware support
+        set(SOURCE_CODE "
+        #include <sys/prctl.h>
+        int main() {
+            int ret = prctl(PR_SVE_GET_VL);
+            return ret >= 0 ? 0 : 1;
+        }
+        ")
+        file(WRITE ${CMAKE_BINARY_DIR}/check_sve_support.c "${SOURCE_CODE}")
+        try_run(RUN_RESULT COMPILE_RESULT
+            ${CMAKE_BINARY_DIR}/check_sve_support_output
+            ${CMAKE_BINARY_DIR}/check_sve_support.c
+        )
+
+        if(RUN_RESULT EQUAL 0)
+            message(STATUS "ARM SVE hardware support detected")
+            # Apply the SVE flags and definitions specifically to the xgboost target
+            set(XGBOOST_ARM_SVE_HARDWARE_SUPPORT TRUE PARENT_SCOPE)
+            set(XGBOOST_SVE_FLAGS "-march=armv8-a+sve" PARENT_SCOPE)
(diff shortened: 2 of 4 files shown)


trivialfis:
Thank you for the PR! I'm not an expert in SIMDs, is it guaranteed to have aligned pointers and padded memory allocation for the intrinsics?

divya2108:
Hi @trivialfis 
The code has been thoroughly validated to ensure alignment and padding issues are addressed. The datatypes have not been altered from the scalar code; instead, the original scalar operations have been translated into SIMD using equivalent SVE intrinsics and there are no compile-time errors. 
All potential accuracy issues have been resolved and verified on widely used datasets like Iris, Airlines delay and breast cancer detection.

divya2108:
Specialized allocators like aligned_alloc() doesn't help with SVE intrinsics because:
1. ARM's SVE SIMD architecture handles data processing in parallel, which inherently considers data alignment. For example for a 256 bit vector length system, we load 8 float elements (8*32) through VLA(vector length agnostic) instructions into a SVE register.
2. Most of the instructions including widening and narrowing instructions helps take care of the data alignment.

divya2108:
Hi @trivialfis,
Additionally, SVE also provides predicate registers enabling key features such as:
a) **Per-lane predication** that allows SIMD instructions to be executed conditionally on specific lanes of a SIMD register
b) **Predicate-driven loop control and management** that helps to manage data that does not align perfectly with the vector length.

Mousius:
> Thank you for the detailed info. Could you please help explain why it works without the use of specialized allocators like https://en.cppreference.com/w/c/memory/aligned_alloc ? It's important for us to know the logic for future maintenance.

Hi @trivialfis,

As @divya2108 mentioned, SVE has predication support.

These lines create masks which limit the load/stores from going out of bounds:
https://github.com/dmlc/xgboost/blob/5194c1726bfa7f5d1b10856d87889f4fc1871638/src/common/hist_util.cc#L265-L266

SVE is also happy to do unaligned loads and stores.

trivialfis:
Thank you for the explanation! I will take a deeper look.

src/common/hist_util.cc
```diff
@@ -252,13 +257,52 @@ void RowsWiseBuildHistKernel(Span<GradientPair const> gpair, Span<bst_idx_t cons
...
-    for (size_t j = 0; j < row_size; ++j) {
-      const uint32_t idx_bin =
-          two * (static_cast<uint32_t>(gr_index_local[j]) + (kAnyMissing ? 0 : offsets[j]));
-      auto hist_local = hist_data + idx_bin;
-      *(hist_local) += pgh_t[0];
-      *(hist_local + 1) += pgh_t[1];
-    }
+    #if defined(SVE_SUPPORT_DETECTED)
+      svfloat64_t pgh_t0_vec = svdup_n_f64(pgh_t[0]);
+      svfloat64_t pgh_t1_vec = svdup_n_f64(pgh_t[1]);
```
trivialfis, line 262: It seems you don't need the `pgh_t` in the SVE code section. Could you please use `p_gpair` and have the names of the loaded vectors, such as `svfloat64_t grad, svfloat64_t hess`, for readability?

trivialfis (commented):
Started looking into this PR today. Thank you for working on using the arm intrinsic, but could you please add **detailed** code comments and extract the code into an independent section (like a function that can be inlined)? Most people here (me included) have a background closer to data science instead of low-level programming.

divya2108:
> Started looking into this PR today. Thank you for working on using the arm intrinsic, but could you please add **detailed** code comments and extract the code into an independent section (like a function that can be inlined)? Most people here (me included) have a background closer to data science instead of low-level programming.

Hi @trivialfis, Thank you for suggesting the appropriate changes. I have made the modifications as recommended. Could you please review the updated changes?

maajidkhann:
The CMake logic looks right. It only compiles SVE code when the compiler supports it and during the runtime it triggers the SVE code only when the hardware supports SVE (I see there's a runtime HW check for SVE ISA). Changes LGTM.

Mousius:
> The CMake logic looks right. It only compiles SVE code when the compiler supports it and during the runtime it triggers the SVE code only when the hardware supports SVE (I see there's a runtime HW check for SVE ISA). Changes LGTM.

Can you point me to where the runtime check happens? As far as I can tell, this only works if the build environment supports compiling and running SVE. 

The new path is conditionally compiled with `#ifdef XGBOOST_SVE_SUPPORT_DETECTED` with no fallback at runtime here:
https://github.com/dmlc/xgboost/pull/10680/files#diff-def34075edb2b3bdb6dc7b5ebcffd518793520fd4fffd70870b12f076a3cb481R305-R308

This makes me think this is only for users compiling from sources on a specific piece of hardware. If we wanted this to work in the generically distributed wheel, we'd have to do the SVE runtime check instead of the `#ifdef`.

Correct me if I'm wrong 😸

maajidkhann:
> > The CMake logic looks right. It only compiles SVE code when the compiler supports it and during the runtime it triggers the SVE code only when the hardware supports SVE (I see there's a runtime HW check for SVE ISA). Changes LGTM.
> 
> Can you point me to where the runtime check happens? As far as I can tell, this only works if the build environment supports compiling and running SVE.
> 
> The new path is conditionally compiled with `#ifdef XGBOOST_SVE_SUPPORT_DETECTED` with no fallback at runtime here: https://github.com/dmlc/xgboost/pull/10680/files#diff-def34075edb2b3bdb6dc7b5ebcffd518793520fd4fffd70870b12f076a3cb481R305-R308
> 
> This makes me think this is only for users compiling from sources on a specific piece of hardware. If we wanted this to work in the generically distributed wheel, we'd have to do the SVE runtime check instead of the `#ifdef`.
> 
> Correct me if I'm wrong 😸

I agree with you. I found the HW detection logic here: https://github.com/dmlc/xgboost/pull/10680/files#diff-5650b69c609ef22dea88915eb256a6838341248d3ddfd17430388f7f7e58c4feR24

But this is just for compile time. I think similar logic need to be used during runtime and a runtime check is required.

Since there's already a working SVE HW detection logic, should be easy to reintroduce it in the source code file.

CC @divya2108

trivialfis:
Is SVE guaranteed to be available for ARM implementation?

divya2108:
> Is SVE guaranteed to be available for ARM implementation?

No, SVE is not guaranteed to be available on all ARM implementations. While ARMv8-A architecture, which includes SVE support, is present in newer processors like Graviton3, Graviton4, Grace, it is not mandatory for all ARM CPUs to implement SVE. The code in [hist_util.cc](https://github.com/dmlc/xgboost/blob/master/src/common/hist_util.cc) checks for SVE support at runtime to ensure that the target hardware supports it & runs the default code otherwise.

divya2108:
> > > The CMake logic looks right. It only compiles SVE code when the compiler supports it and during the runtime it triggers the SVE code only when the hardware supports SVE (I see there's a runtime HW check for SVE ISA). Changes LGTM.
> > 
> > 
> > Can you point me to where the runtime check happens? As far as I can tell, this only works if the build environment supports compiling and running SVE.
> > The new path is conditionally compiled with `#ifdef XGBOOST_SVE_SUPPORT_DETECTED` with no fallback at runtime here: https://github.com/dmlc/xgboost/pull/10680/files#diff-def34075edb2b3bdb6dc7b5ebcffd518793520fd4fffd70870b12f076a3cb481R305-R308
> > This makes me think this is only for users compiling from sources on a specific piece of hardware. If we wanted this to work in the generically distributed wheel, we'd have to do the SVE runtime check instead of the `#ifdef`.
> > Correct me if I'm wrong 😸
> 
> I agree with you. I found the HW detection logic here: https://github.com/dmlc/xgboost/pull/10680/files#diff-5650b69c609ef22dea88915eb256a6838341248d3ddfd17430388f7f7e58c4feR24
> 
> But this is just for compile time. I think similar logic need to be used during runtime and a runtime check is required.
> 
> Since there's already a working SVE HW detection logic, should be easy to reintroduce it in the source code file.
> 
> CC @divya2108

Yes, I agree. Thank you for bringing this to notice. 
I have added a SVE hardware check at runtime. Now it is generically compiled and falls back on the default code if SVE hardware support is not detected.  

I have verified this by building the code on different architectures. Here is a summary for more clarity: 
image

rageshhajela16:
@trivialfis Thanks for the initial review and your comments. Can you please suggest any additional feedback which might need further evaluation or clarifications from our side. Thanks. cc: @divya2108

trivialfis:
Sorry for the slow reply, got stuck at some other work lately. One question, is it possible to reduce the call frequency of `check_sve_hw_support` to maybe once per training session?

divya2108:
> Sorry for the slow reply, got stuck at some other work lately. One question, is it possible to reduce the call frequency of `check_sve_hw_support` to maybe once per training session?

Yes, it's possible to reduce the frequency of calls to check_sve_hw_support by implementing a caching mechanism that checks the SVE hardware support status only once at the beginning of a training session. I have stored the result and it is being reused throughout the session.

divya2108:
Hi @trivialfis, just wanted to follow up on the code review. Let me know if you need any additional details or clarifications.

rageshhajela16:
> Hi @trivialfis, just wanted to follow up on the code review. Let me know if you need any additional details or clarifications.

Hi @trivialfis , we have pushed all the necessary changes. Kindly review and let us know for any additional details or modifications required. Thanks.

trivialfis:
Apologies for the slow response, will look into this. Thank you very much for your patience!

trivialfis:
@hcho3 I see you have assigned yourself to the PR. Thank you for volunteering! I recently got access to a Grace machine and might be able to do some tests there.

hcho3:
Sorry it must have been by mistake. I will try to look at the PR for the next few days however.

CMakeLists.txt
```diff
@@ -38,6 +38,12 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
   endif()
 endif()
 
+include(${xgboost_SOURCE_DIR}/cmake/CheckSVEsupport.cmake)
+check_xgboost_sve_support()
+if(XGBOOST_COMPILER_HAS_ARM_SVE)
+    add_compile_definitions(XGBOOST_SVE_COMPILER_SUPPORT)
```
hcho3, line 44: We should prefer target-scoped primitives to global-scoped primitives in CMake. See https://cliutils.gitlab.io/modern-cmake/chapters/basics.html#targets-are-your-friend. I will go ahead and revise the pull request to replace `add_compile_definitions` with `target_compile_definitions`.

cmake/CheckSVEsupport.cmake
```diff
@@ -0,0 +1,32 @@
+function(check_xgboost_sve_support)
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+    include(CheckCSourceCompiles)
+
+    # Save the original C_FLAGS to restore later
+    set(ORIGINAL_C_FLAGS "${CMAKE_C_FLAGS}")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=armv8-a+sve")
```
hcho3, line 7: Rather than modifying `CMAKE_C_FLAGS` directly, we should use `CMAKE_REQUIRED_FLAGS` instead, which is [explicitly designed to influence the behavior of `check_c_source_compiles`](https://cmake.org/cmake/help/v3.31/module/CheckCSourceCompiles.html).

Example: https://github.com/facebook/rocksdb/blob/c0be6a4b90a1f616969b2a808035ebf334894a37/CMakeLists.txt#L309-L342

Let me update the pull request to use `CMAKE_REQUIRED_FLAGS`.

src/common/hist_util.cc
```diff
@@ -15,19 +16,28 @@
...
+#define PREFETCH_READ_T0(addr) \
+  do {                         \
+  } while (0)
 #endif  // defined(XGBOOST_MM_PREFETCH_PRESENT)
 
 namespace xgboost::common {
-HistogramCuts::HistogramCuts() {
-  cut_ptrs_.HostVector().emplace_back(0);
-}
+HistogramCuts::HistogramCuts() { cut_ptrs_.HostVector().emplace_back(0); }
```
hcho3, line 40: I see lots of unsubstantial formatting changes. We should apply clang-format with the same `.clang-format` configuration.

chyunsu3 (changes requested):

src/common/hist_util.cc
```diff
@@ -186,6 +188,107 @@ class GHistBuildingManager {
...
+        if (ret == -1) {
+            cached_sve_support = 0;
+        } else {
+            cached_sve_support = 1;
+        }
+    }
+    return cached_sve_support;
+}
+
+static int sve_enabled = check_sve_hw_support();
```
hcho3, line 289: Does the value of a global static variable valid when accessed from multiple threads? It might be better to thread-local storage instead.

@trivialfis Any thoughts on this topic?

chyunsu3 (commented):
We should ensure that the CI pipeline tests XGBoost with SVE intrinsic.

Two kinds of tests are needed:
1. **End-to-end test**. Build XGBoost with SVE and run pytests. We can do this easily, using the ARM worker machine in the CI.
2. **Micro test**. Write a gtest that compares the result of the histogram kernel with and without SVE enabled. For this we need a way to temporarily disable SVE feature at runtime.

trivialfis:
> We should ensure that the CI pipeline tests XGBoost with SVE intrinsic.

Is it enabled on the CI?

src/common/hist_util.cc
trivialfis: I will work on it. Still learning the code.

CMakeLists.txt
divya2108: > We should prefer target-scoped primitives to global-scoped primitives in CMake. See https://cliutils.gitlab.io/modern-cmake/chapters/basics.html#targets-are-your-friend. I will go ahead and revise the pull request to replace `add_compile_definitions` with `target_compile_definitions`.

Hey @hcho3, thanks for reviewing the code. Will you be making this change and pushing it?

trivialfis: I have made the change and pushed it.

trivialfis: I have made the change and pushed it.

rageshhajela16:
Thanks @trivialfis for your review and contributions to this implementation! Please let us if you would like us to contribute any additional fixes based on review comments. We would like to confirm with you before proceeding to avoid any duplicate work. Thanks again for your time in review of this PR. We appreciate! cc: @divya2108

trivialfis:
@hcho3 Do you think it's possible to have this in the pip wheel?

trivialfis:
Could you please share the CPU you were using for the benchmarks? I ran a benchmark on a Grace machine  (I work for NVIDIA) with synthetic data, and the performance is actually lower.

My synthetic data:
- n_samples: 67108864
- n_features: 256

Training parameters:
- 64 iterations
- 6 max depth

```
* SVE disabled

[63]    Train-rmse:18.88612
Qdm train (sec) ended in:  123.9944839477539 seconds.
Trained for 64 iterations.
{'load-batches': {'load (sec)': 7.225183725357056}, 'load-all': {'concat (sec)': 1.3589859008789062e-05}, 'Qdm': {'Train-DMatrix (sec)': 29.529566526412964, 'train (sec)': 123.9944839477539}}

* SVE enabled

[63]    Train-rmse:18.88612
Qdm train (sec) ended in:  154.86435317993164 seconds.
Trained for 64 iterations.
{'load-batches': {'load (sec)': 7.193156003952026}, 'load-all': {'concat (sec)': 1.430511474609375e-05}, 'Qdm': {'Train-DMatrix (sec)': 29.482257604599, 'train (sec)': 154.86435317993164}}
```

hcho3:
> Do you think it's possible to have this in the pip wheel?

Yes, it should be possible.

divya2108:
> Could you please share the CPU you were using for the benchmarks? I ran a benchmark on a Grace machine (I work for NVIDIA) with synthetic data, and the performance is actually lower. I have verified that the row-wise kernel is being used.
> 
> My synthetic data:
> 
> * n_samples: 67108864
> * n_features: 256
> 
> Training parameters:
> 
> * 64 iterations
> * 6 max depth
> 
> Compilers:
> 
> * g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
> 
> ```
> * SVE disabled
> 
> [63]    Train-rmse:18.88612
> Qdm train (sec) ended in:  123.9944839477539 seconds.
> Trained for 64 iterations.
> {'load-batches': {'load (sec)': 7.225183725357056}, 'load-all': {'concat (sec)': 1.3589859008789062e-05}, 'Qdm': {'Train-DMatrix (sec)': 29.529566526412964, 'train (sec)': 123.9944839477539}}
> 
> * SVE enabled
> 
> [63]    Train-rmse:18.88612
> Qdm train (sec) ended in:  154.86435317993164 seconds.
> Trained for 64 iterations.
> {'load-batches': {'load (sec)': 7.193156003952026}, 'load-all': {'concat (sec)': 1.430511474609375e-05}, 'Qdm': {'Train-DMatrix (sec)': 29.482257604599, 'train (sec)': 154.86435317993164}}
> ```
> 
> It's okay to be slower on certain platforms, we can look for a way to disable it. But I would like to get some understanding of how the performance works for your platform as well.

These are the machine and dataset details which I used:

AWS Graviton3, ARM-based CPU

Dataset details: kaggle higgs boson dataset (250000 samples, 32 features)

Training parameters: 120 iterations, 6 max_depth

compiler: g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

divya2108:
Hi @trivialfis, just wanted to follow up on the PR review. Let me know if there’s anything I can do to help or clarify. Thanks!

trivialfis:
apologies for the slow response here. I need to test it on aws, and find a way to disable it for grace.

trivialfis:
> can you please help to review and try it! Let us know your feedback. Thanks

Could you please elaborate on what I should try?

divya2108:
> apologies for the slow response here. I need to test it on aws, and find a way to disable it for grace.

Hi @trivialfis, I have added a change to disable SVE for SVE128 supported hardware (grace) and run the default Neon flow in that.  case. Now the SVE implementation works for ARM CPU's having vector length >=256.

And also seems like the changes you pushed in the past seem to be missing now somehow? I can't locate it here in remote and even locally. Can you help me point to the source of these commits (Can't even find them in your xgboost forked repo), I can cherry-pick it back on top of my branch.

trivialfis:
Let me take a look tomorrow. Seems there's a force push that overrode the previous changes.

trivialfis:
Will try to redo some of the changes and run test.

trivialfis:
Tested again on grace today, small overhead for dispatching, I think I can work around it:
```
master:

[63]	Train-rmse:18.88612
Qdm train (sec) ended in:  97.99142909049988 seconds.
Trained for 64 iterations.
{'load-batches': {'load (sec)': 7.013079643249512}, 'load-all': {'concat (sec)': 1.7404556274414062e-05}, 'Qdm': {'Train-DMatrix (sec)': 30.664090633392334, 'train (sec)': 97.99142909049988}}

sve:

[63]	Train-rmse:18.88612
Qdm train (sec) ended in:  103.17199659347534 seconds.
Trained for 64 iterations.
{'load-batches': {'load (sec)': 7.116881370544434}, 'load-all': {'concat (sec)': 1.6450881958007812e-05}, 'Qdm': {'Train-DMatrix (sec)': 30.612082481384277, 'train (sec)': 103.17199659347534}}
```
Will test again on AWS with graviton.

trivialfis:
I just tested the default build on aws `c8g.x16large`, the `vector_length` is 128?

divya2108:
> I just tested the default build on aws `c8g.x16large`, the `vector_length` is 128 and the extension is not used there?

Hi @trivialfis. As discussed above in my comment, the current logic, checks for SVE VL of the hardware
and if the VL < 256, then it falls back to the default flow (Neon).
 
So, in the machine you tested earlier (Grace) and the c8g.x16large (Graviton4) machine you tested today,
both have SVE128 in the hardware and as per our current logic, it will fall back to default flow (Neon).
 
Curious to know, if overhead remained the same even in c8g.x16large or it came down? Did you try any workarounds
around that in the code with the dispatcher?
 
I believe, if you now run the code on machines with SVE>=256, you would observe the boost in performance because
of SVE ACLE code. One good machine to test SVE goodness would be m7g.8xlarge (Graviton 3) which has SVE256 in the hardware.

trivialfis:
@divya2108 Do you expect the extension would provide speedup for graviton 4 if we remove the 256-bit restriction? If so I can run the test again. Otherwise, I don't think we will merge the extension only for graviton3.

Mousius:
> @divya2108 Do you expect the extension would provide speedup for graviton 4 if we remove the 256-bit restriction? If so I can run the test again. Otherwise, I don't think we will merge the extension only for graviton3 unless there are other latest and accessible platforms that can benefit from this.

It wouldn't be only Graviton 3, A64FX has 512-bit SVE, and we can incrementally improve this to detect the streaming SVE vector length on devices where that's >128-bit.

trivialfis:
Apologies for the slow response here.

Would be great if there's a latest hardware that we can benchmark on and see the merit. Otherwise it's quite unlikely that we will merge a assembly like PR.

We love performance, but most people here are data scientists instead of performance engineers, it takes some really good use cases to justify the code complexity. As you can see, we couldn't attach anyone to help maintain the C++ codebase, let alone an assembly function.

divya2108:
Revisiting this PR with the motivation to ensure XGBoost remains performant by default across evolving CPU architectures, particularly as Arm adoption continues to accelerate. We’d like to highlight that Fujitsu has an upcoming [processor](https://github.com/fujitsu/FUJITSU-MONAKA/blob/main/doc/FUJITSU-MONAKA_specification_v1.0.pdf) supporting 256-bit SVE, and our implementation has demonstrated meaningful speedups on such hardware.
image

 
Given the prior interest in exploring architecture-specific optimizations, we kindly request your consideration for merging this PR. It is implemented using ACLE intrinsics and is fully guarded with architecture checks, ensuring zero impact on platforms without SVE support.
 
We fully understand the concerns around increased maintenance overhead for low-level optimizations. To that end, we’re committed to helping maintain and support this code going forward. If there are any specific improvements, validation steps, or additional benchmarks you'd like us to provide to help move this forward, we’re more than happy to assist.
 
Thank you for your time and consideration.

trivialfis:
Thank you for sharing, how did you get access to this machine?

divya2108:
> Thank you for sharing, how did you get access to this machine?

FUJITSU-MONAKA is not yet publicly available. Our mention of it was to highlight the relevance of forward-looking architecture support, as it will be one of the first commercial processors with 256-bit SVE.
For our current benchmarking and validation, we’ve used AWS Graviton3, which supports 256-bit SVE and shares architectural similarities with FUJITSU-MONAKA. While not identical, it provides a practical and accessible platform to test and validate SVE-based optimizations, helping ensure the PR delivers meaningful performance improvements on real-world hardware.

PR #10685, opened 2024-08-09
[CI] use key=value form for Dockerfile ENV statements

jameslamb:
Was looking at recent PRs, and noticed a warning in #10654.

> Legacy key/value format with whitespace separator should not be used
LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format
More info: https://docs.docker.com/go/dockerfile/rule/legacy-key-value-format/

Per those Docker docs, it seems that syntax found in some of this project's Dockerfiles is considered deprecated and may be removed in a future Docker release.

```dockerfile
# deprecated
ENV VER 10

# preferred syntax
ENV VER=10
```

This proposes fixing those here. I found them like this:

```shell
git grep 'ARG .* .*'
git grep 'ENV .* .*'
```


trivialfis (approved):
Looks good!

PR #10687, opened 2024-08-09
[R] prefer startsWith() to substr() or regular expressions

jameslamb:
Base R has a function `startsWith()`, which behaves similarly to the Python string method `.startswith()` (but vectorized 😁).

This can be faster than other methods for checking the start of a string, like using `substr(x) == "pattern"` or `grep()` or `grepl()`. For what it's worth, I also think it's a little easier to read.

This proposes adding a `{lintr}` linter that suggests doing that, and fixing two such cases in `{xgboost}`. Specifically these:

```text
xgboost/R-package/R/utils.R:107:7: warning: [string_boundary] Use startsWith() to detect an initial substring. Doing so is more readable and more efficient.
      substr(NVL(params[['objective']], 'x'), 1, 6) == 'multi:' &&
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

[[2]]
xgboost/R-package/R/xgb.model.dt.tree.R:112:37: warning: [string_boundary] Use !is.na(x) & startsWith(x, string) to detect a fixed initial substring, or, if missingness is not a concern, just startsWith. Doing so is more readable and more efficient.
  td <- td[Tree %in% trees & !grepl('^booster', t)]
                                    ^~~~~~~~~~
```

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -104,7 +104,7 @@ check.booster.params <- function(params, ...) {
 
   # for multiclass, expect num_class to be set
   if (typeof(params[['objective']]) == "character" &&
-      substr(NVL(params[['objective']], 'x'), 1, 6) == 'multi:' &&
+      startsWith(NVL(params[['objective']], 'x'), 'multi:') &&
       as.numeric(NVL(params[['num_class']], 0)) < 2) {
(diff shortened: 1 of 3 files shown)


trivialfis (approved):
Thank you for the PR!

PR #10682, opened 2024-08-07
[EM] CPU implementation for external memory QDM.

trivialfis:
The PR works on CPU implementation using QDM with an external data source. File IO for the `SparsePage` is quite expensive, especially when we work on GPU (coming). In addition, the reference DMatrix can help speed up CV significantly since quantile sketching is the most expensive part of constructing a QDM.
- A new DMatrix type.
- Extract common code into a new QDM base class.

Not yet working:
- Not exposed to the interface yet, will wait for the GPU implementation.
- ~No meta info yet, still working on the source.~
- Exporting data to CSR is not supported yet.

Most of the changes are for refactoring, the new DMatrix type shares lots of code with `IterativeDMatrix` and `SparsePageDMatrix`.

On a system with 128GB memory, 24 threads, and PCIe4 NVME (w/o RAID0), I have tested a 514GB float32 synthetic dataset with 32 batches and 512 features, using only 6 rounds of training:
```
ExtMemQuantileDMatrix duration: 1578.2151629924774
UpdateOneIter: 1091.35s, 6 calls
```
Still has plenty of room for improvement. But I consider the external memory with supported objectives (not based on quantiles) "usable" now. Training a 514GB dataset on a single consumer device is not trivial.

PR #10689, opened 2024-08-09
[EM] Add GPU version of the external memory QDM.

trivialfis:
The second part of https://github.com/dmlc/xgboost/pull/10682 .

This PR adds basic GPU support for the  external memory QDM, some features still need to be worked on:

TO-DO:
- Interpolation between CPU and GPU.
- Exporting data to CSR.
- Expose the features to the interface.

In addition, we will change the GPU updater to support batch-based training in the coming PRs. I haven't decided whether we should support the page concatenation in the future. It will depend on the difficulty of supporting two batch-based training paradigms.


trivialfis:
cc @rongou .

rongou (approved):

PR #10691, opened 2024-08-09
[EM] Enable access to the number of batches.

trivialfis:
- Expose `NumBatches` in `DMatrix`.
- Small cleanup for removing legacy CUDA stream and ~force CUDA context initialization~.
- Purge old external memory data generation code.

The CUDA uninitialized error was observed in CTK 12.2, tested CTK 12.5 on a different machine but it's not observable there. Let's keep the stream sync here to avoid potential errors, it only runs at the beginning of a thread.

A subtle difference is that, after the PR, if an external DMatrix contains only a single batch, then the `SingleColBlock` would return true. Currently, it returns true regardless of where it actually contains a single block when the data type is external memory.

*update*: Removed the force initialization for now due to CI running CTK 11.8.

### DIFF
--- a/src/data/sparse_page_source.cu
+++ b/src/data/sparse_page_source.cu
@@ -18,4 +18,14 @@ void DevicePush(DMatrixProxy *proxy, float missing, SparsePage *page) {
   cuda_impl::Dispatch(proxy,
                       [&](auto const &value) { CopyToSparsePage(value, device, missing, page); });
 }
+
+void InitNewThread::operator()() const {
+  *GlobalConfigThreadLocalStore::Get() = config;
+  // For CUDA 12.2, we need to force initialize the CUDA context by synchronizing the
+  // stream when creating a new thread in the thread pool. While for CUDA 11.8, this
+  // action might cause an insufficient driver version error for some reason. Lastly, it
+  // should work with CUDA 12.5 without any action being taken.
+
+  // dh::DefaultStream().Sync();
+}
 }  // namespace xgboost::data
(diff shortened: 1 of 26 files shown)


trivialfis:
Okay, so the CUDA error is observable on the CI as well.

trivialfis:
Note to myself: The GPU test is invalid in the master branch, update prediction cache simply returns false due to external memory not being supported.

trivialfis:
Will wait for https://github.com/dmlc/xgboost/pull/10697 and test.

trivialfis:
cc @rongou .

src/data/sparse_page_source.cu
```diff
@@ -18,4 +18,14 @@ void DevicePush(DMatrixProxy *proxy, float missing, SparsePage *page) {
...
 }
+
+void InitNewThread::operator()() const {
+  *GlobalConfigThreadLocalStore::Get() = config;
+  // For CUDA 12.2, we need to force initialize the CUDA context by synchronizing the
+  // stream when creating a new thread in the thread pool. While for CUDA 11.8, this
+  // action might cause an insufficient driver version error for some reason. Lastly, it
+  // should work with CUDA 12.5 without any action being taken.
+
+  // dh::DefaultStream().Sync();
```
rongou, line 29: Is this still needed?

rongou (approved):

src/data/sparse_page_source.cu
trivialfis: Yes, it's needed for 12.2, but it seems fine with 12.5. However, currently the CI is running 11.8, which doesn't work with the initialization.

PR #10693, opened 2024-08-11
[coll] Reduce the amount of open files (socket).

trivialfis:
Reduce the chance of hitting `Too many open files`.


trivialfis:
cc @rongou .

rongou (approved):

PR #10694, opened 2024-08-12
[EM] Support ExtMemQdm in the GPU predictor.

trivialfis:
- Remove the unused `entry_start` parameter.
- Support batch-local row index in ellpack.


trivialfis:
Tried to reproduce it using docker, but got a driver initialization error instead, whereas the CI shows illegal CUDA memory access.

trivialfis:
cc @rongou .

rongou (approved):

PR #10700, opened 2024-08-13
[multi] Implement weight feature importance.

trivialfis:
- Only weight is implemented as the vector-leaf tree doesn't have hessian or gain at the moment.
- Mixed tree types is not supported yet.


trivialfis:
Will try to support mixed trees.

hcho3 (approved):

PR #10701, opened 2024-08-13
SYCL. Add implementation for UpdatePredictionCache; Connect sycl-updater with leraner.

razdoburdin:
Hi,
I continue adding of sycl support (https://github.com/dmlc/xgboost/pull/10690, https://github.com/dmlc/xgboost/pull/10681 etc).
By this PR I add the next part of the training-related code for the sycl devices. Here I add functional for updating of prediction cache. And since this is the last part required for single device fitting, this PR also connect sycl-updater with xgboost-learner. So it makes possible to launch fitting on sycl-device.

### DIFF
--- a/plugin/sycl/tree/updater_quantile_hist.cc
+++ b/plugin/sycl/tree/updater_quantile_hist.cc
@@ -4,7 +4,6 @@
  */
 #include <vector>
 #include <memory>
-#include <utility>
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wtautological-constant-compare"
@@ -29,12 +28,6 @@ void QuantileHistMaker::Configure(const Args& args) {
   const DeviceOrd device_spec = ctx_->Device();
   qu_ = device_manager.GetQueue(device_spec);
 
-  // initialize pruner
-  if (!pruner_) {
-    pruner_.reset(TreeUpdater::Create("prune", ctx_, task_));
-  }
-  pruner_->Configure(args);
-
   param_.UpdateAllowUnknown(args);
   hist_maker_param_.UpdateAllowUnknown(args);
 
@@ -56,7 +49,6 @@ void QuantileHistMaker::SetPimpl(std::unique_ptr<HistUpdater<GradientSumT>>* pim
                 ctx_,
                 qu_,
                 param_,
-                std::move(pruner_),
                 int_constraint_, dmat));
   if (collective::IsDistributed()) {
     LOG(FATAL) << "Distributed mode is not yet upstreamed for sycl";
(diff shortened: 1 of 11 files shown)


tests/cpp/tree/test_prediction_cache.h
```diff
@@ -0,0 +1,92 @@
+#include <gtest/gtest.h>
```
trivialfis, line 1: Please include the license header, along with C header guard.

razdoburdin: done

plugin/sycl/tree/updater_quantile_hist.cc
```diff
@@ -27,21 +29,115 @@ void QuantileHistMaker::Configure(const Args& args) {
...
+
   param_.UpdateAllowUnknown(args);
   hist_maker_param_.UpdateAllowUnknown(args);
+
+  bool has_fp64_support = qu_.get_device().has(::sycl::aspect::fp64);
+  if (hist_maker_param_.single_precision_histogram || !has_fp64_support) {
+    if (!hist_maker_param_.single_precision_histogram) {
+      LOG(WARNING) << "Target device doesn't support fp64, using single_precision_histogram=True";
+    }
+    hist_precision_ = HistPrecision::fp32;
```
trivialfis, line 46: A bit of warning, this can lead to non-converging models. It's the reason we removed this option from XGBoost long time ago.

plugin/sycl/tree/updater_quantile_hist.cc
```diff
@@ -27,21 +29,115 @@ void QuantileHistMaker::Configure(const Args& args) {
   const DeviceOrd device_spec = ctx_->Device();
   qu_ = device_manager.GetQueue(device_spec);
 
+  // initialize pruner
+  if (!pruner_) {
+    pruner_.reset(TreeUpdater::Create("prune", ctx_, task_));
```
trivialfis, line 34: The pruner is not using sycl, do we need it? The existing hist tree updater doesn't use the pruner. Instead, it checks whether a node is splittable before growing, it's not only more efficient due to less computation, but also simplier.

plugin/sycl/tree/updater_quantile_hist.cc
razdoburdin: Thanks! I will check and fix it later

plugin/sycl/tree/updater_quantile_hist.cc
razdoburdin: Make sense. I have removed pruner from sycl.

trivialfis (approved):
We can merge the PR, but it seems you are using a really old XGBoost version as reference? Might be a good idea to use the latest hist method for easing future maintenance and integration.

razdoburdin:
> We can merge the PR, but it seems you are using a really old XGBoost version as reference? Might be a good idea to use the latest hist method for easing future maintenance and integration.

Yes, we will work on deeper synchronization of the plugin with actual xgboost.

PR #10702, opened 2024-08-13
SYCL. Fix sampling for devices without fp64 support

razdoburdin:
Currently sampling for `sycl`-devices uses `oneDPL `library for generation of bernoulli distribution. Unfortunately, `oneDPL `implicitly uses `fp64 `for bernoulli, that makes impossible to launch fitting on `sycl `devices without `fp64 `support.
This PR fix this issue by adding specialized code for such cases.


trivialfis:
Out of curiosity, what type of device doesn't implement f64?

razdoburdin:
> Out of curiosity, what type of device doesn't implement f64?
Mostly integrated graphics

trivialfis (approved):

PR #10707, opened 2024-08-14
[EM] Enable prediction cache for GPU.

trivialfis:
- Use `UpdatePosition` for all nodes and skip `FinalizePosition` when external memory is used.
- Create `encode/decode` for node position, this is just as a refactor.
- Reuse code between update position and finalization.

The choice to skip the finalization is just because it's the simplest way and the most efficient. The finalization requires an additional iteration of the data, which can hurt external memory performance.

The cache is disabled when external memory sampling is used as the partitioner doesn't have valid row index.

### DIFF
--- a/src/common/categorical.h
+++ b/src/common/categorical.h
@@ -1,20 +1,17 @@
 /**
- * Copyright 2020-2023, XGBoost Contributors
+ * Copyright 2020-2024, XGBoost Contributors
  * \file categorical.h
  */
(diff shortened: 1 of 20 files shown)


trivialfis:
cc @rongou .

rongou (approved):

trivialfis:
Some tests are difficult to do without https://github.com/dmlc/xgboost/pull/10691 . We care about whether the data contains only a single batch, while whether the batch is stored on disk doesn't matter.

PR #10711, opened 2024-08-16
Add managed memory allocator.

trivialfis:
For experimenting with external memory.

### DIFF
--- a/src/common/cuda_pinned_allocator.h
+++ b/src/common/cuda_pinned_allocator.h
@@ -8,8 +8,8 @@
 
 #include <cuda_runtime.h>
 
-#include <cstddef>
-#include <limits>
+#include <cstddef>  // for size_t
+#include <limits>   // for numeric_limits
 
 #include "common.h"
 
@@ -30,7 +30,7 @@ struct PinnedAllocPolicy {
   using value_type = T;            // NOLINT: The type of the elements in the allocator
 
   size_type max_size() const {  // NOLINT
-    return (std::numeric_limits<size_type>::max)() / sizeof(value_type);
+    return std::numeric_limits<size_type>::max() / sizeof(value_type);
   }
 
   pointer allocate(size_type cnt, const_pointer = nullptr) {  // NOLINT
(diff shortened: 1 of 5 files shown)


trivialfis:
cc @rongou .

src/common/cuda_pinned_allocator.h
```diff
@@ -1,93 +1,114 @@
...
 template <typename T>
-class pinned_allocator;
+struct PinnedAllocPolicy {
+  using pointer = T*;              // NOLINT: The type returned by address() / allocate()
+  using const_pointer = const T*;  // NOLINT: The type returned by address()
+  using size_type = std::size_t;   // NOLINT: The type used for the size of the allocation
+  using value_type = T;            // NOLINT: The type of the elements in the allocator
+
+  size_type max_size() const {  // NOLINT
+    return (std::numeric_limits<size_type>::max)() / sizeof(value_type);
```
rongou, line 33: Shouldn't it just be:
```cpp
return std::numeric_limits<size_type>::max() / sizeof(value_type);
```

src/common/cuda_pinned_allocator.h
```diff
@@ -1,93 +1,114 @@
...
 template <typename T>
-class pinned_allocator {
+struct ManagedAllocPolicy {
+  using pointer = T*;              // NOLINT: The type returned by address() / allocate()
+  using const_pointer = const T*;  // NOLINT: The type returned by address()
+  using size_type = std::size_t;   // NOLINT: The type used for the size of the allocation
+  using value_type = T;            // NOLINT: The type of the elements in the allocator
+
+  size_type max_size() const {  // NOLINT
+    return (std::numeric_limits<size_type>::max)() / sizeof(value_type);
```
rongou, line 57: ditto

rongou (approved):

src/common/cuda_pinned_allocator.h
trivialfis: Yeah, that's from the old code, haven't done the cleanup yet, will change it.

src/common/cuda_pinned_allocator.h
trivialfis: Done.

PR #10717, opened 2024-08-19
[cmake] prefer 'cmake -B' and 'cmake --build' everywhere

### LINKED ISSUE #10699: [docs] [cmake] switch to 'cmake --build'?
## Description

Many CI scripts and docs throughout this project use a pattern like this for CMake:

https://github.com/dmlc/xgboost/blob/43704549a2746fcf134a89b3153502a02a94a295/doc/build.rst?plain=1#L130-L133

Would you support a series of PRs (or maybe just one, depending on how involved it gets) changing all such cases to this pattern?

```shell
cmake -B build -S . -DUSE_CUDA=ON
cmake --build build --target all -j4
```

## Benefits of this work

* reduce friction in building
  - *(2 commands instead of 4)*
* makes it easier for users to change generators
  - *(e.g. to Ninja instead of GNU Make)*

## Approach

Replace all uses of `mkdir build && cd build && cmake ..` or similar with the corresponding 2 `cmake` commands (`cmake` followed by `cmake --build`).

## Notes

I did this in LightGBM a few months ago. We haven't received any complaints about it, and I've personally found it a bit easier to work with: https://github.com/microsoft/LightGBM/pull/6368

But won't be offended if you say "no thank you, we don't want this" 😅

jameslamb:
Fixes #10699

Everywhere that I could find a pattern like this:

```shell
mkdir build
cd build
cmake -GNinja ..
ninja -v
```

I changed it to this:

```shell
cmake -B build -S . -G Ninja
cmake --build build -v
```

## Notes for Reviewers

I reverted this to only changing docs, based on https://github.com/dmlc/xgboost/pull/10717#issuecomment-2298191276.

### How I found these

```shell
git grep -E ' \.\. '
git grep -E ' \.\.$'
git grep -E 'cmake.* \.\.'
git grep 'mkdir build'
git grep -E 'make \-j'
git grep -E 'ninja '
```

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -26,8 +26,6 @@ jobs:
           pkg install -y cmake git ninja googletest
 
         run: |
-          mkdir build
-          cd build
-          cmake .. -GNinja -DGOOGLE_TEST=ON
-          ninja -v
-          ./testxgboost
+          cmake -B build -S . -GNinja -DGOOGLE_TEST=ON
+          cmake --build build -v
+          ./build/testxgboost
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -26,6 +26,8 @@ jobs:
           pkg install -y cmake git ninja googletest
 
         run: |
-          cmake -B build -S . -GNinja -DGOOGLE_TEST=ON
-          cmake --build build -v
-          ./build/testxgboost
+          mkdir build
+          cd build
+          cmake .. -GNinja -DGOOGLE_TEST=ON
+          ninja -v
+          ./testxgboost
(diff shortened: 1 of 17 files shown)


trivialfis:
Thank you for working on simplifying the build logic. I think we don't have to be thorough for this change, most of us understand what's going on with the cmake command, and typing out `ninja -v` makes it clear that we are not using `make` (which is slower due to its way of dependency traversal). Maybe it's good enough to focus on documentation as a starting point. ;-)

jameslamb:
Thanks @trivialfis 

I've reverted some changes here so this PR now only affects the docs.

Fine with me if you'd like to consider #10699 closed after just these changes (or even if you've changed your mind and don't want these docs changes either, in which case we can just close the PR + issue). I won't be offended, I don't want these PRs to be annoying 😅

jameslamb:
Ok no problem, thanks for considering it.

PR #10721, opened 2024-08-19
Check cub errors.

trivialfis:
- Make sure cuda error returned by cub scan is caught.
- Avoid temporary buffer allocation in thrust device vector.


trivialfis:
cc @rongou .

rongou (approved):

PR #10730, opened 2024-08-21
Use latest cmake for clang-tidy.

trivialfis:
- Install cmake using pip.
- Fix compile command generation.
- Clean up the tidy script and remove the need to load the yaml file.
- Fix modernized type traits.
- Fix span class. Polymorphism support is dropped as it was broken anyway.

The new cmake is required by the latest GPU tree shap.

### DIFF
--- a/include/xgboost/json.h
+++ b/include/xgboost/json.h
@@ -222,6 +222,14 @@ class JsonObject : public Value {
   ~JsonObject() override = default;
 };
 
+namespace detail {
+template <typename T, typename U>
+using IsSameT = std::enable_if_t<std::is_same_v<std::remove_cv_t<T>, std::remove_cv_t<U>>>;
+
+template <typename T>
+using IsF64T = std::enable_if_t<std::is_same_v<T, double>>;
+}  // namespace detail
+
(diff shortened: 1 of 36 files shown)


trivialfis:
@hcho3 I haven't updated clang-tidy itself on the CI yet, but I have run `clang-tidy-19` locally and the test should pass.

include/xgboost/json.h
```diff
@@ -223,6 +222,14 @@ class JsonObject : public Value {
   ~JsonObject() override = default;
 };
 
+namespace detail {
+template <typename T, typename U>
+using IsSameT = std::enable_if_t<std::is_same_v<std::remove_cv_t<T>, std::remove_cv_t<U>>>;
```
trivialfis, line 227: Only extracting this due to a potential bug in clang-tidy-19, can't be sure yet. It keeps asking me to use modernized type traits if I don't add these helpers.

hcho3:
> Do you want me to update the clang-tidy as well?

Yes, that would be great.

trivialfis:
@hcho3 Done.

hcho3 (approved):

PR #10734, opened 2024-08-22
[EM] Avoid resizing host cache.

trivialfis:
- Add SAM allocator and resource.
- Use page-based cache instead of stream-based cache.

Data copying can be expensive when the memory is tracked by CUDA. This PR aims to avoid any resize operation in the cache by directly caching the page instead of writing it as a stream.


trivialfis:
cc @rongou .

rongou (approved):

Issue #10312, opened 2024-05-22
Hang inside NCCL.


trivialfis:
https://buildkite.com/xgboost/xgboost-ci-multi-gpu/builds/5027#018f9e76-44ae-4979-bd6d-c9aa5e0a617d

We ran into something like this before we moved into process-based multi-GPU training. The issue appears to be back now we have thread-based tests again.

trivialfis:
CUDA is not quite friendly to valgrind, need a different method to debug this.

trivialfis:
Not exactly an XGBoost issue, but we still need a workaround.

trivialfis:
The fix requires a kernel driver that supports CTK12.5+.

hcho3:
@trivialfis Should we use CTK 12.5 in our CI ?

trivialfis:
@hcho3 It can prevent a flaky CPP MGPU test timeout error. Please help update the MGPU test CI when you are available.

PR #10735, opened 2024-08-23
[coll] Improve column split tests with named threads.

trivialfis:
- Name some of the threads for easier debugging.
- Use the `MGPU` prefix for running tests on the CI.

Tried to debug https://github.com/dmlc/xgboost/issues/10312 , the hang comes from inside CTK host malloc, I don't have a fix yet. Upstreaming some of the enhancements.


trivialfis:
cc @rongou .

rongou (approved):

PR #10736, opened 2024-08-23
[EM] Pass batch parameter into extmem format.

trivialfis:
- Allow customization for format read.
- Customization for the number of pre-fetch batches.


src/data/gradient_index_page_source.cc
```diff
@@ -37,6 +37,7 @@ void ExtGradientIndexPageSource::Fetch() {
     CHECK_GE(source_->Iter(), 1);
     CHECK_NE(cuts_.Values().size(), 0);
     HostAdapterDispatch(proxy_, [this](auto const& value) {
+      CHECK(this->proxy_->Ctx()->IsCUDA()) << "All batches must use the same device type.";
```
trivialfis, line 40:
```suggestion
      CHECK(this->proxy_->Ctx()->IsCPU()) << "All batches must use the same device type.";
```

trivialfis:
cc @rongou

rongou (approved):

PR #10737, opened 2024-08-24
musl libc support

SamuelMarks:
Warnings are bad! - Found one in your library deep within another:
```
#17 120.2   In file included from /tmp/wd/postgresml/pgml-extension/target/debug/build/xgboost-sys-a1e545b7b843ddb4/out/xgboost/rabit/include/rabit/internal/socket.h:38,
#17 120.2                    from /tmp/wd/postgresml/pgml-extension/target/debug/build/xgboost-sys-a1e545b7b843ddb4/out/xgboost/src/collective/loop.cc:15:
#17 120.2   /usr/include/sys/poll.h:1:2: warning: #warning redirecting incorrect #include <sys/poll.h> to <poll.h> [-Wcpp]
#17 120.2       1 | #warning redirecting incorrect #include <sys/poll.h> to <poll.h>
#17 120.2         |  ^~~~~~~
#17 120.2   In file included from /tmp/wd/postgresml/pgml-extension/target/debug/build/xgboost-sys-a1e545b7b843ddb4/out/xgboost/rabit/include/rabit/internal/socket.h:38,
#17 120.2                    from /tmp/wd/postgresml/pgml-extension/target/debug/build/xgboost-sys-a1e545b7b843ddb4/out/xgboost/src/collective/socket.cc:14:
#17 120.2   /usr/include/sys/poll.h:1:2: warning: #warning redirecting incorrect #include <sys/poll.h> to <poll.h> [-Wcpp]
#17 120.2       1 | #warning redirecting incorrect #include <sys/poll.h> to <poll.h>
#17 120.2         |  ^~~~~~~
#17 120.2   In file included from /tmp/wd/postgresml/pgml-extension/target/debug/build/xgboost-sys-a1e545b7b843ddb4/out/xgboost/rabit/include/rabit/internal/socket.h:38,
#17 120.2                    from /tmp/wd/postgresml/pgml-extension/target/debug/build/xgboost-sys-a1e545b7b843ddb4/out/xgboost/src/collective/tracker.cc:4:
#17 120.2   /usr/include/sys/poll.h:1:2: warning: #warning redirecting incorrect #include <sys/poll.h> to <poll.h> [-Wcpp]
#17 120.2       1 | #warning redirecting incorrect #include <sys/poll.h> to <poll.h>
#17 120.2         |  ^~~~~~~
```

EDIT0: Another one, resolved it same way they did https://github.com/gperftools/gperftools/issues/693' ; a simple `#define mmap64 mmap`

EDIT1: To replicate, here's a `Dockerfile`:
```dockerfile
FROM alpine
RUN apk add --no-cache cmake gcc g++ make musl-dev
WORKDIR /src/xgboost
COPY . .
RUN cmake -S . -B build && \
    cmake --build build
```

### DIFF
--- a/include/xgboost/collective/poll_utils.h
+++ b/include/xgboost/collective/poll_utils.h
@@ -35,13 +35,7 @@
 
 #if !defined(_WIN32)
 
-#define _GNU_SOURCE
-#include <features.h>
-#ifndef __USE_GNU
-    #define __MUSL__
-#endif
-#undef _GNU_SOURCE
-
+#include "xgboost/unixdefs.h"
 #ifdef __MUSL__
 #include <poll.h>
 #else
--- a/include/xgboost/collective/poll_utils.h
+++ b/include/xgboost/collective/poll_utils.h
@@ -35,7 +35,18 @@
 
 #if !defined(_WIN32)
 
+#define _GNU_SOURCE
+#include <features.h>
+#ifndef __USE_GNU
+    #define __MUSL__
+#endif
+#undef _GNU_SOURCE
+
+#ifdef __MUSL__
+#include <poll.h>
+#else
 #include <sys/poll.h>
+#endif
 
 using SOCKET = int;
 using sock_size_t = size_t;  // NOLINT
--- a/include/xgboost/unixdefs.h
+++ b/include/xgboost/unixdefs.h
@@ -1,0 +1,14 @@
+#if defined(UNIXDEFS_H) && !defined(_MSC_VER) && \
+    !defined(__MINGW32__) && !defined(__CYGWIN__) && \
+    !defined(_WIN32) && !defined(__WINDOWS__) && \
+    !defined(__MUSL__)
+#define UNIXDEFS_H
+
+#define _GNU_SOURCE
+#include <features.h>
+#ifndef __USE_GNU
+    #define __MUSL__
+#endif
+#undef _GNU_SOURCE
+
+#endif
--- a/src/common/io.cc
+++ b/src/common/io.cc
@@ -7,13 +7,7 @@
 #include <sys/mman.h>  // for mmap, mmap64, munmap, madvise
 #include <unistd.h>    // for close, getpagesize
 
-#define _GNU_SOURCE
-#include <features.h>
-#ifndef __USE_GNU
-    #define __MUSL__ 
-#endif
-#undef _GNU_SOURCE
-
+#include "xgboost/unixdefs.h"
 #ifdef __MUSL__
 #define mmap64 mmap
 #endif
--- a/src/common/io.cc
+++ b/src/common/io.cc
@@ -7,6 +7,17 @@
 #include <sys/mman.h>  // for mmap, mmap64, munmap, madvise
 #include <unistd.h>    // for close, getpagesize
 
+#define _GNU_SOURCE
+#include <features.h>
+#ifndef __USE_GNU
+    #define __MUSL__ 
+#endif
+#undef _GNU_SOURCE
+
+#ifdef __MUSL__
+#define mmap64 mmap
+#endif
+
 #else
 
 #include <xgboost/windefs.h>


include/xgboost/unixdefs.h
```diff
@@ -0,0 +1,14 @@
+#if defined(UNIXDEFS_H) && !defined(_MSC_VER) && \
+    !defined(__MINGW32__) && !defined(__CYGWIN__) && \
+    !defined(_WIN32) && !defined(__WINDOWS__) && \
+    !defined(__MUSL__)
+#define UNIXDEFS_H
+
+#define _GNU_SOURCE
+#include <features.h>
```
trivialfis, line 8: Could you please share why we need this header?

include/xgboost/unixdefs.h
```diff
@@ -0,0 +1,14 @@
+#if defined(UNIXDEFS_H) && !defined(_MSC_VER) && \
+    !defined(__MINGW32__) && !defined(__CYGWIN__) && \
+    !defined(_WIN32) && !defined(__WINDOWS__) && \
+    !defined(__MUSL__)
+#define UNIXDEFS_H
+
+#define _GNU_SOURCE
```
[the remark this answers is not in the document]
SamuelMarks: https://stackoverflow.com/a/5583764
> Defining `_GNU_SOURCE` has nothing to do with license and everything to do with writing (non-)portable code.

include/xgboost/unixdefs.h
SamuelMarks: Feature flags. Can put it in another header if you think it works better semantically there.

include/xgboost/collective/poll_utils.h
```diff
@@ -35,7 +35,12 @@
 
 #if !defined(_WIN32)
 
+#include "xgboost/unixdefs.h"
+#ifdef __MUSL__
+#include <poll.h>
+#else
 #include <sys/poll.h>
```
trivialfis, line 42: Apologies for the wrong import. I think `#include <poll.h>` should work fine given https://man7.org/linux/man-pages/man2/poll.2.html

src/common/io.cc
```diff
@@ -7,6 +7,11 @@
 #include <sys/mman.h>  // for mmap, mmap64, munmap, madvise
 #include <unistd.h>    // for close, getpagesize
 
+#include "xgboost/unixdefs.h"
+#ifdef __MUSL__
+#define mmap64 mmap
```
trivialfis, line 12: We can use `mmap` as default and let the 32-bit system raise an overflow error when the file size is larger than 2GB.

trivialfis (commented):
Thank you for the replies. Looking closer, I think the musl folks are right and there's no need to make specialization for musl. Please see the inlined comments for the proposed changes.

SamuelMarks:
> Thank you for the replies. Looking closer, I think the musl folks are right and there's no need to make specialization for musl. Please see the inlined comments for the proposed changes.

@trivialfis Yes that is correct; if you use `poll.h` and `mmap` then you won't need this PR. As for the `mmap` fallout, you can also check the other direction, if `mmap64` is available then `_LARGEFILE64_SOURCE` is defined or more correctly `#if defined(_FILE_OFFSET_BITS) && _FILE_OFFSET_BITS >= 64` then you can use that symbol. See also: https://pubs.opengroup.org/onlinepubs/009695399/utilities/c99.html#tag_04_12_13_03

Otherwise sure you can just switch to `mmap` over `mmap64` for everyone. It's your call*.
*pun intended?

trivialfis:
> Yes that is correct; if you use poll.h and mmap then you won't need this PR

Sounds great! That should make things much simpler. Would you like to help make the fix?

> As for the mmap fallout, you can also check the other direction

I don't think that's necessary, users of 32-bit platforms can reduce the batch size to be smaller than 2 GB.

SamuelMarks:
@trivialfis Sure thing done in #10767

Issue #10726, opened 2024-08-20
[R] XGBoost 2 deprecation error


mayer79:
I am testing the new R package (version 2.2.0.0). 

The deprecation warnings give errors for objects like "valids" or "watchlist" that contain external pointers.

```r
library(xgboost)

dtrain <- xgb.DMatrix(iris[2:4], label = iris[, 1])
fit <- xgb.train(
  params = list(), data = dtrain, watchlist = list(train = dtrain), nrounds = 10
)

# Gives (sorry, german...)
Fehler in parse(text = paste(new_par, "<-", pars[[pars_par]])) : 
  <text>:1:10: Unerwartete(s) '<'
1: evals <- <
             ^
Zusätzlich: Warnmeldung:
In check.deprecation(...) : 'watchlist' ist veraltet.
Benutzen Sie stattdessen 'evals'
Siehe help("Deprecated") und help("xgboost-deprecated").
```

The warning is fine, but the error seems a bit harsh.

mayer79:
Ah, I see this is already discussed here: https://github.com/dmlc/xgboost/pull/10563/files

PR #10563, opened 2024-07-10
[R] Gracefully handle deprecated parameters

hcho3:
Reproducer:
```R
require(xgboost)
data(agaricus.train, package='xgboost')
data(agaricus.test, package='xgboost')
train <- agaricus.train
test <- agaricus.test
dtrain <- xgb.DMatrix(data = train$data, label=train$label)
dtest <- xgb.DMatrix(data = test$data, label=test$label)
watchlist <- list(train=dtrain, test=dtest)
bst <- xgb.train(data=dtrain, max_depth=2, eta=1, nthread = 2, nrounds=2,
                 watchlist=watchlist, objective = "binary:logistic")
```

The current `master` version fails with cryptic error
```
Loading required package: xgboost
Error in parse(text = paste(new_par, "<-", pars[[pars_par]])) :
  <text>:1:10: unexpected '<'
1: evals <- <
             ^
Calls: xgb.train -> check.deprecation -> eval -> parse
In addition: Warning message:
In check.deprecation(...) : 'watchlist' is deprecated.
Use 'evals' instead.
See help("Deprecated") and help("xgboost-deprecated").
4: parse(text = paste(new_par, "<-", pars[[pars_par]]))
3: eval(parse(text = paste(new_par, "<-", pars[[pars_par]])), envir = env)
2: check.deprecation(...)
1: xgb.train(data = dtrain, max_depth = 2, eta = 1, nthread = 2,
       nrounds = 2, watchlist = watchlist, objective = "binary:logistic")
```
The error occurs because the use of `eval(parse(...))` does not correctly handle data frame values.
This pull request replaces the expression with `assign(...)`.

Also: Explicitly filter out deprecated parameters from the dot params. In the reproducer script, not filtering out `watchlist` leads to error `Error: No method asJSON S3 class: xgb.DMatrix` because `watchlist` was still part of the `params` list and R tried to serialize the `params` list as JSON.

After the fix, the reproducer script successfully runs, with only a warning:
```
Loading required package: xgboost
[1]     train-logloss:0.233156  test-logloss:0.226460
[2]     train-logloss:0.136542  test-logloss:0.137763
Warning message:
In check.deprecation(...) : 'watchlist' is deprecated.
Use 'evals' instead.
See help("Deprecated") and help("xgboost-deprecated").
```

PR #10741, opened 2024-08-24
[R] Safer assignment during deprecation checks

mayer79:
Fixes https://github.com/dmlc/xgboost/issues/10726 by using a more stable assignment of deprecated parameters to the paramter list.


mayer79:
Ah, I see this is already fixed here: https://github.com/dmlc/xgboost/pull/10563/files

Issue #10742, opened 2024-08-24
[R] Warning when plotting trees with DiagrammeR


david-cortes:
Not sure what's the reason here, but there's a warning coming from DiagrammeR when trying to plot multiple trees:

```r
library(xgboost)
data(iris)
model <- xgboost(iris[,-5], iris$Species, nrounds=3)
xgb.plot.multi.trees(model)
```
```
Column 2 ['No'] of item 2 is missing in item 1. Use fill=TRUE to fill with NA (NULL for list columns), or use.names=FALSE to ignore column names. use.names='check' (default from v1.12.2) emits this message and proceeds as if use.names=FALSE for  backwards compatibility. See news item 5 in v1.12.2 for options to control this message.
```

PR #10743, opened 2024-08-24
[R] Fix warning about mismatched names when joining

david-cortes:
fixes https://github.com/dmlc/xgboost/issues/10742

This PR fixes a warning (which actually turn out to come from data.table, not from DiagrammeR) about mismatched column names when concatenating inputs.

I'm not 100% if this is doing what the function is meant to do though, so would be ideal if someone could take a deeper look.


trivialfis:
@mayer79 Could you please help review this PR when you are available? I'm not familiar with the DiagrammeR package.

mayer79 (approved):
The original code tried to rbind two data.frames with different column names. This is fixed.

david-cortes:
> The original code tried to rbind two data.frames with different column names. This is fixed.

But is that what it was supposed to be doing?

mayer79:
> > The original code tried to rbind two data.frames with different column names. This is fixed.
> 
> But is that what it was supposed to be doing?

I think yes, but I have never actively used this function.

PR #10744, opened 2024-08-24
[R] Allow passing data.frame to SHAP

david-cortes:

This PR modifies the SHAP-related functions to accept common input type that the prediction functions can take as input.


R-package/R/xgb.ggplot.R
```diff
@@ -102,6 +102,20 @@ xgb.ggplot.deepness <- function(model = NULL, which = c("2x1", "max.depth", "med
 #' @export
 xgb.ggplot.shap.summary <- function(data, shap_contrib = NULL, features = NULL, top_n = 10, model = NULL,
                                     trees = NULL, target_class = NULL, approxcontrib = FALSE, subsample = NULL) {
+  has_categ <- FALSE
```
trivialfis, line 105: What happens if we don't normalize the values?

david-cortes: I don't know. But that normalization is not applicable to categorical columns.

trivialfis: Was thinking if normalization is optional, we might disable it for categorical features automatically.

david-cortes: The rest of the function would still error out. I think it could be adjusted (it's a matter of switching the plot from 'numeric scale' to 'discrete values'), but I'm not familiar with this xgboost functionality so don't know if the plot would make sense.

Does the python interface have an analog for plotting shap values?

trivialfis: No, in Python the plotting is handled by the shap package, which doesn't support categorical features.

david-cortes: Perhaps @mayer79 could chime in.

mayer79 (approved):

R-package/R/xgb.ggplot.R
trivialfis: Thank you for the explanation! Would be great if we can share it in the error message as well instead of a unsupported error.

david-cortes: Modified it to ignore categorical features when producing these plots.

trivialfis (approved):

trivialfis (approved):

PR #10747, opened 2024-08-24
[R] Add training metrics monitoring for `xgboost()`

david-cortes:

I just realized that the new `xgboost()` interface was not doing anything with the `verbose` parameter. This PR adds an option to use the same x/y training data for metrics monitoring, which gets activated when passing a non-silent verbosity level.


R-package/R/xgboost.R
```diff
@@ -932,11 +935,16 @@ xgboost <- function(
 
   fn_dm <- if (use_qdm) xgb.QuantileDMatrix else xgb.DMatrix
   dm <- do.call(fn_dm, lst_args$dmatrix_args)
+  evals <- list()
```
trivialfis, line 938: Hmm, I missed this the last time. How does one use validation with this `X/y` interface?

david-cortes: There's no validation. This is just monitoring on the same training data.

I guess  evaluation on a different dataset could be added later, but it would need be more involved - for example, it would need to re-encoded categorical columns in the evaluation data using the same categories as in 'x', would need to check whether the classes in 'y' match, and so on.

trivialfis: Thank you for the explanation. Might be necessary though, considering validation is pretty much the only effective way in keeping ML training in check. I doubt that anyone who uses ML training aside from DL doesn't have validation during training.

david-cortes: There's `xgb.cv` though. There could be an `xgboost.cv` (or `cv.xgboost`) to do it better, or somethng like 'eval_data_fraction' (better yet if implemented in the core library), but I think those could be added later after CRAN release blockers are addressed.

PR #10750, opened 2024-08-24
[R] Remove demos

david-cortes:
ref https://github.com/dmlc/xgboost/issues/9734#issuecomment-2308471749

This PR removes the R package demos for XGBoost.

Reason being: it's uncommon nowadays for people to look at demos. XGBoost has vignettes , in-docs examples, and tutorials in the docs website; all of which are more likely to be read by users. Moreover, some of the functionalities in the demo have become outdated with the changes that XGBoost has implemented since they were written (e.g. it now has support for categorical features and data frames), and some of the demos have already been removed due to being broken (e.g. interaction with the library 'caret', which is deprecated already and not meant to be used).

It might be worth it in the future to add some of those back as additional snippets in vignettes, but given the limited manpower available, removing them is going to make progress on the R interface faster.


trivialfis:
Hi, could you please help remove the related CI tests as well?

david-cortes:
> Hi, could you please help remove the related CI tests as well?

Removed.

trivialfis (approved):

PR #10753, opened 2024-08-26
[EM] Return a full DMatrix instead of a Ellpack from the GPU sampler.

trivialfis:
- Slice meta info using GPU.
- Work with `DMatrix` intead of `Ellpack` in the hist updater.
This way we can support page concatenation and batch training at the same time.

todos:
- [x] Test `NumNonMissing`.


trivialfis:
cc @rongou .

rongou (approved):

PR #10755, opened 2024-08-26
Ensure that `pip check` does not fail due to bad platform tag

### LINKED ISSUE #10705: `xgboost` python package fails on `pip check` with `pip 24.2` on `osx-arm64`
`pip 24.2` introduces a check on the tag within the wheel file.

`xgboost` has a hatch custom build hook that overrides the creation of the tag. 

See: https://github.com/dmlc/xgboost/blob/v2.1.1/python-package/pyproject.toml#L48
Which should trigger: https://github.com/dmlc/xgboost/blob/9c9db1259240bffe9040ed7ca6e3fb2c1bda80e4/python-package/hatch_build.py#L15

While building the version `2.1.1` of the package for the `conda` ecosystem, the resulting tag for the wheel file on `osx-arm64` has the `Tag: py3-none-macosx_11_1_arm64` but among the platform tags that are checked, there is no `macosx_11_1_arm64` platform tag and the pip check fails with:

```
xgboost 2.1.1 is not supported on this platform
```

As far as I understand from the issues and PRs relative to this problem, the idea of `pip` for the `Darwin` platform, was to flatten the version to `macosx_11_0` for all of the `11.x`.

This seems to occur for Python version `3.12` but not on `[3.8, 3.9, 3.10, 3.11]`.

### Workaround

I disabled the custom build hook for `osx-arm64` and, it seems to work fine.

### Other information

- Full WHEEL content for osx-arm64:
  
  ```
  Wheel-Version: 1.0
  Generator: hatchling 1.21.1
  Root-Is-Purelib: true
  Tag: py3-none-macosx_11_1_arm64
  ```
- compatible tags for the platform: 
  
  ```
  ['macosx_13_0_arm64', 'macosx_13_0_universal2', 'macosx_12_0_arm64', 'macosx_12_0_universal2', 'macosx_11_0_arm64', 'macosx_11_0_universal2', 'macosx_10_16_universal2', 'macosx_10_15_universal2', 'macosx_10_14_universal2', 'macosx_10_13_universal2', 'macosx_10_12_universal2', 'macosx_10_11_universal2', 'macosx_10_10_universal2', 'macosx_10_9_universal2', 'macosx_10_8_universal2', 'macosx_10_7_universal2', 'macosx_10_6_universal2', 'macosx_10_5_universal2', 'macosx_10_4_universal2']
  ```

- `pip check`: https://github.com/pypa/pip/blob/e98cc5ce078d8c8afd6804ff4e61aa2b12d05715/src/pip/_internal/commands/check.py#L29-L34

### Other references

See: https://github.com/pypa/packaging/issues/435
See PR: https://github.com/AnacondaRecipes/xgboost-feedstock/pull/20

hcho3:
Closes #10705

`sysconfig.get_platform` does not report an accurate platform tag; to use a correct platform tag, use `pip`.


python-package/pyproject.toml
```diff
@@ -1,6 +1,7 @@
 [build-system]
 requires = [
-    "hatchling>=1.12.1"
+    "hatchling>=1.12.1",
+    "pip>=22.0"
```
hcho3, line 4: This installs a copy of the latest `pip` inside an isolated build environment; it does not replace the `pip` in the host system.

hcho3:
@lorepirri Can I get a review for this fix?

python-package/hatch_build.py
```diff
@@ -3,16 +3,16 @@
 Here, we customize the tag of the generated wheels.
 """
 
-import sysconfig
 from typing import Any, Dict
 
 from hatchling.builders.hooks.plugin.interface import BuildHookInterface
+from pip._vendor.packaging.tags import platform_tags
```
jakirkham, line 9: Since this is just [coming from `packaging`]( https://github.com/pypa/packaging/blob/c385b58c4d0e4d66f5c8273ebb155801c92aadf3/src/packaging/tags.py#L499 ), should we use that directly

```suggestion
from packaging.tags import platform_tags
```

python-package/pyproject.toml
```diff
@@ -1,6 +1,7 @@
 [build-system]
 requires = [
-    "hatchling>=1.12.1"
+    "hatchling>=1.12.1",
+    "pip>=22.0",
```
jakirkham, line 4: Appears `packaging` added `platform_tags` to the public API in commit ( https://github.com/pypa/packaging/commit/ba07d8287b4554754ac7178d177033ea3f75d489 ), which is part of [21.1]( https://github.com/pypa/packaging/releases/tag/21.1 )

```suggestion
    "packaging>=21.1",
```

jakirkham (commented):
Thanks Hyunsu! 🙏

Am wondering if just using `packaging` directly might be more straightforward. Included a suggestion along these lines below

Please let me know what you think 🙂

python-package/hatch_build.py
hcho3: No, we can't know for sure if `packaging` is available on the user's system. It's better to count down `pip`, since it's directly used for building and packaging wheels.

hcho3: I wonder if `pip` carries a significantly modified version of `packaging`. In the end, we have to generate a tag that will be accepted by `pip`. I don't want a situation where we use `packaging` to generate a tag and have it rejected by `pip`.

jakirkham: I don't think so. AIUI `pip` adds several things to a `_vendor` directory to avoid having dependencies on them, which would result in cycles in the dependency graph

In fact [their vendoring policy]( https://github.com/pypa/pip/blob/ac63d0fa126fbfb01263b1597c2b425ba720d59f/src/pip/_vendor/README.rst ) says

> * Vendored libraries **MUST** not be modified except as required to
  successfully vendor them.
> * Vendored libraries **MUST** be released copies of libraries available on
  PyPI.
> ...

python-package/pyproject.toml
jakirkham: Now that we are using `packaging` instead, think we can mark this thread resolved

tests/buildkite/remove_nccl_dep.patch
```diff
@@ -1,14 +1,12 @@
...
-@@ -30,8 +30,7 @@ classifiers = [
- ]
+@@ -30,7 +30,6 @@ classifiers = [
  dependencies = [
      "numpy",
--    "scipy",
--    "nvidia-nccl-cu12 ; platform_system == 'Linux' and platform_machine != 'aarch64'"
-+    "scipy"
+     "scipy",
+-    "nvidia-nccl-cu12 ; platform_system == 'Linux' and platform_machine != 'aarch64'",
```
jakirkham, line 9: As a note to reviewers, GitHub PR diffs of patches can look confusing

Would recommend looking at [the patch on Hyunsu's fork]( https://github.com/hcho3/xgboost/blob/fix_pip_tag/tests/buildkite/remove_nccl_dep.patch ) (also included below)

This is just dropping the `nvidia-nccl-cu12` line

```patch
diff --git python-package/pyproject.toml python-package/pyproject.toml
index 20d3f9974..953087ff4 100644
--- python-package/pyproject.toml
+++ python-package/pyproject.toml
@@ -30,7 +30,6 @@ classifiers = [
 dependencies = [
     "numpy",
     "scipy",
-    "nvidia-nccl-cu12 ; platform_system == 'Linux' and platform_machine != 'aarch64'",
 ]
 
 [project.urls]
```

jakirkham (approved):

Issue #10704, opened 2024-08-13
[CI] Column split tests might hang.


trivialfis:
A tracking issue for column split tests.

trivialfis:
https://buildkite.com/xgboost/xgboost-ci-windows/builds/6395

PR #10756, opened 2024-08-27
Increase timeout in C++ tests from 1 to 5 seconds.

trivialfis:
To avoid CI failures on FreeBSD.

- Increase timeout from 1 second to 5 seconds for the poll and recv.
- Use the name "ColumnSplit" uniformly across all tests for easier filtering.

Related https://github.com/dmlc/xgboost/issues/10704 .


trivialfis:
I have been running column split tests for half an hour now, seem stable.

hcho3 (approved):

PR #10759, opened 2024-08-27
Use trailing commas in `pyproject.toml`

jakirkham:
As [trailing commas are valid in TOML]( https://github.com/toml-lang/toml/blob/dc5af251cd348ee053d9355359743767e3d0b123/toml.md?plain=1#L667C33-L668C55 ), adding them to these lists cuts down on diff line noise when adding and removing dependencies. Also simplifies downstream patching ( like in build contexts: https://github.com/conda-forge/xgboost-feedstock/pull/191#discussion_r1732031076 )

cc @hcho3

### DIFF
--- a/python-package/pyproject.toml
+++ b/python-package/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "hatchling>=1.12.1"
+    "hatchling>=1.12.1",
 ]
 backend-path = ["."]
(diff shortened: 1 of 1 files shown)


hcho3:
I will make the change in #10755. Thanks!

jakirkham:
Ah ok. Didn't see that. Thanks Hyunsu! 🙏

PR #10760, opened 2024-08-28
[EM] Have one partitioner for each batch.

trivialfis:
- Initialize one partitioner for each batch.
- Collect partition size during initialization.
- Support base ridx in the finalization.

todos:
- [x] Support concatenated pages.
- [x] Check performance impact. (no impact)


trivialfis:
cc @rongou .

rongou (approved):

PR #10764, opened 2024-08-29
[EM] Refactor GPU histogram builder.

trivialfis:
- Expose the maximum number of cached nodes to be consistent with the CPU implementation. Also easier for testing.
- Extract the subtraction trick for easier testing.
- Split up the `GradientQuantiser` to avoid circular dependency.

todos:
- [x] perf check
- [x] document


trivialfis:
cc @rongou .

rongou (approved):

PR #10765, opened 2024-08-29
[EM] Avoid stream sync in quantile sketching.

trivialfis:
Use context for thrust policy. This is to prepare for overlapping the quantile calculation with data loading.


trivialfis:
cc @rongou

rongou (approved):

PR #10767, opened 2024-08-30
POSIX compliant `poll.h` and `mmap` over `sys/poll.h` and `mmap64`

SamuelMarks:
Closes #10737 

Basic build testing confirms this works with both musl and glibc:
```dockerfile
FROM alpine
RUN apk add --no-cache cmake gcc g++ make musl-dev
WORKDIR /src/xgboost
COPY . .
RUN cmake -S . -B build && \
    cmake --build build
```

```dockerfile
FROM debian:bookworm-slim
RUN apt-get update -qq && \
    apt-get install -y cmake gcc g++ make libstdc++6 && \
    rm -rf /var/lib/apt/lists/*
WORKDIR /src/xgboost
COPY . .
RUN cmake -S . -B build && \
    cmake --build build
```

### DIFF
--- a/include/xgboost/collective/poll_utils.h
+++ b/include/xgboost/collective/poll_utils.h
@@ -35,7 +35,7 @@
 
 #if !defined(_WIN32)
 
-#include <sys/poll.h>
+#include <poll.h>
 
 using SOCKET = int;
 using sock_size_t = size_t;  // NOLINT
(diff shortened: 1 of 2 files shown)


trivialfis (approved):
Thank you for the fix!

snoopcheri:
Great to see this fixed, as we stumbled over this problem ourselves as well!
I assume this will be part of the 2.2.x release?

trivialfis:
We will probably make a patch release.

PR #10796, opened 2024-09-03
Update custom_metric_obj.rst

CSantos01:
Added the square to the derivative in the hessian


trivialfis (approved):
Thank you for the fix!

PR #10798, opened 2024-09-04
fix xgb.model.dt.tree in case where all leaves are negatives

jonibr22:
If all leaves on the model are negative values, xgb.model.dt.tree throws an error saying that the given model is a non-tree model.
However, a model where all leaves are negative is possible if i'm using poisson/gamma objective with few number of rounds.


trivialfis (approved):

trivialfis:
Thank you for the fix!

PR #10801, opened 2024-09-04
C API: Cache CUDA availability so it's not re-evaluated every API call

chuckatkins:


src/c_api/c_api.cu
```diff
@@ -59,15 +66,27 @@ void XGBBuildInfoDevice(Json *p_info) {
...
-  // Not calling `safe_cuda` to avoid unnecessary exception handling overhead.
-  // If errors, do nothing, assuming running on CPU only machine.
-  cudaGetDevice(&device_id_);
+  try {
+    if (MIGHT_HAVE_CUDA.load(std::memory_order_relaxed)) {
+      // Not calling `safe_cuda` to avoid unnecessary exception handling overhead.
+      // If errors, do nothing, assuming running on CPU only machine.
+      cudaGetDevice(&device_id_);
+    }
+  } catch (dmlc::Error const&) {
```
trivialfis, line 75: Where is this error being raised if `safe_cuda` is not used?

trivialfis:
Closing due to stalled.

Issue #9376, opened 2023-07-11
[Python] Make lib paths configurable


hcho3:
Currently, we use heuristics to determine the path of libxgboost.so: https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/libpath.py. This heuristics may fail in edge cases, e.g. when `sys.base_prefix` does not refer to a suitable system directory.

We should allow users to specify a custom directory for libxgboost.so. It can be done via an environment variable or a configuration file.

monego:
One "edge case" is Guix, where base_prefix resolves to the interpreter location:

```python
import sys
print(sys.base_prefix)
# /gnu/store/d0mzc34bs02g6fm7dy5xq9ggfws81sna-python-3.10.7
```

Another one would be Nix, but they seem to be rebuilding libxgboost with the Python package while Guix keeps them separate.

The workarounds are symlinking libxgboost.so somewhere in the Python package or manually patching libpath.py. It would be good to set an environment variable instead of having to work it around.

PR #10802, opened 2024-09-04
Made lib paths configurable via environment variable

iiradia:
#9376 This PR adds documentation to clarify the use of the `XGBOOST_LIBRARY_PATH` environment variable, allowing users to set a custom directory for the `libxgboost.so` shared library. This is helpful for cases where the default system path does not correctly resolve the library location.


trivialfis (commented):
Thank you for working on this! Do you think it's possible to encode this information during build time instead of runtime?

PR #10810, opened 2024-09-08
[EM] Refactor ellpack construction.

trivialfis:
- Remove the calculation of n_symbols in the accessor.
- Pass the context into various ctors.
- Pack initialization steps into parameter list.
- Specialization for dense data to prepare for further compression.


trivialfis:
cc @rongou

rongou (approved):

PR #10811, opened 2024-09-09
[EM] Avoid synchronous calls and unnecessary ATS access.

trivialfis:
- Pass context into various functions.
- Factor out some CUDA algorithms.
- Use ATS only for update position.


trivialfis:
cc @rongou .

rongou (approved):

PR #10813, opened 2024-09-09
[EM] Multi-level quantile sketching for GPU.

trivialfis:
This helps reduce memory usage when handling multiple batches. This is a variant of the CPU implementation. The basic idea is that when the stream size cannot be known in advance, we store the sketches in multiple levels, with each level growing in exponential size.

No new test since there are already multiple existing tests.

- Fix and old error in the quantile data iterator demo.

### DIFF
--- a/demo/guide-python/quantile_data_iterator.py
+++ b/demo/guide-python/quantile_data_iterator.py
@@ -105,16 +105,20 @@ def main():
     assert m_with_it.num_row() == m.num_row()
     # Tree meethod must be `hist`.
     reg_with_it = xgboost.train(
-        {"tree_method": "hist", "device": "cuda"}, m_with_it, num_boost_round=rounds
+        {"tree_method": "hist", "device": "cuda"},
+        m_with_it,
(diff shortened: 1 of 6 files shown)


demo/guide-python/quantile_data_iterator.py
```diff
@@ -105,17 +105,21 @@ def main():
...
     reg = xgboost.train(
-        {"tree_method": "hist", "device": "cuda"}, m, num_boost_round=rounds
+        {"tree_method": "hist", "device": "cuda"},
+        m,
+        num_boost_round=rounds,
+        evals=[(m, "Train")],
     )
     predict = reg.predict(m)
 
-    numpy.testing.assert_allclose(predict_with_it, predict, rtol=1e6)
```
trivialfis, line 117: Removed since `1e6` was an error from the beginning, the result is not the same. Actually, the one with the iterator has a better result.

trivialfis:
cc @rongou .

rongou (approved):

PR #10814, opened 2024-09-09
Fix misleading error when feature names are missing during inference

shlomota:


trivialfis (approved):
Thank you for the fix!

shlomota:
@trivialfis - is this better?
I see it's blocked?

trivialfis (approved):

PR #10824, opened 2024-09-17
fix c++20

joequant:


hcho3 (changes requested):
We current use the C++17 standard in our build. Therefore, the pull request is not necessary.

hcho3:
This pull request cannot be accepted because it will break the current C++17 build. We will adopt `char8_t` type when we formally adopt the C++20 standard in our builds. Thanks for your understanding.

PR #10828, opened 2024-09-19
[coll] Use loky for rabit op tests.

trivialfis:
- Fix CI IPv6 test error.
- Fix flaky QDM test.

### DIFF
--- a/python-package/xgboost/testing/updater.py
+++ b/python-package/xgboost/testing/updater.py
@@ -218,8 +218,13 @@ def check_extmem_qdm(
         )
 
     booster_it = xgb.train({"device": device}, Xy_it, num_boost_round=8)
-    X, y, w = it.as_arrays()
-    Xy = xgb.QuantileDMatrix(X, y, weight=w)
+    it = tm.IteratorForTest(
+        *tm.make_batches(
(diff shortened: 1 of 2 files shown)


trivialfis:
cc @hcho3 Please help review the PR for unblocking the CI when you are available.

python-package/xgboost/testing/updater.py
```diff
@@ -218,8 +218,13 @@ def check_extmem_qdm(
         )
 
     booster_it = xgb.train({"device": device}, Xy_it, num_boost_round=8)
-    X, y, w = it.as_arrays()
-    Xy = xgb.QuantileDMatrix(X, y, weight=w)
+    it = tm.IteratorForTest(
+        *tm.make_batches(
+            n_samples_per_batch, n_features, n_batches, use_cupy=device != "cpu"
+        ),
+        cache=None,
```
hcho3, line 225:
```suggestion
        cache="cache",
        on_host=on_host,
```
Did you mean to re-create the same iterator?

trivialfis: They are not the same, the second one doesn't cache data to the disk.

PR #10840, opened 2024-09-23
[coll] Avoid all-to-all connection.

trivialfis:
- Use a fixed tree topo for broadcast by always sending the data to the 0th node.

XGBoost always broadcast with 0th node as root anyway.

- Reduce the amount of connections.

### DIFF
--- a/src/collective/tracker.cc
+++ b/src/collective/tracker.cc
@@ -2,6 +2,8 @@
  * Copyright 2023-2024, XGBoost Contributors
  */
 
+#include "tracker.h"
+
 #if defined(__unix__) || defined(__APPLE__)
 #include <netdb.h>       // gethostbyname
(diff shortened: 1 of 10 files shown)


src/collective/tracker.cc
```diff
@@ -155,6 +157,7 @@ Result RabitTracker::Bootstrap(std::vector<WorkerProxy>* p_workers) {
   for (auto const& w : workers) {
     worker_error_handles_.emplace_back(w.Host(), w.ErrorPort());
   }
+  LOG(CONSOLE) << "[tracker]: Bootstrap " << workers.size() << " workers.";
```
trivialfis, line 160: Will turn this into a debug log controlled by the `verbosity` parameter once we have all tests done.

trivialfis:
Replaced by https://github.com/dmlc/xgboost/pull/12075

PR #10842, opened 2024-09-24
SYCL. Implementation of HostDeviceVector

razdoburdin:
Hi,
this PR makes two changes into `sycl`-plugin.
1. `sycl::queue` is always passed by pointer now. This change doesn't affect anything, just unify the code.
2. `Sycl `implementation of `HostDeviceVector` and related tests are introduced. The `HostDeviceVector` is not yet used in the plugin. In the next step I would integrate it to the code.

### DIFF
--- a/plugin/sycl/common/hist_util.cc
+++ b/plugin/sycl/common/hist_util.cc
@@ -19,15 +19,15 @@ namespace common {
  * \brief Fill histogram with zeroes
  */
 template<typename GradientSumT>
-void InitHist(::sycl::queue qu, GHistRow<GradientSumT, MemoryType::on_device>* hist,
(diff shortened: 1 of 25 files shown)


trivialfis:
Hi, thank you for continuing work on sycl. Do you see any path forward to integrate this into main XGBoost? While it's nice to be isolated as a plugin in the early stage, but copying everything seems to make it difficult for future integration. It can't be a plugin forever right? Maybe we can have some brief notes on how to integrate in the future?

napetrov:
@trivialfis  i think with previous changes this would be feature complete for base stuff and we can start process for moving to the main XGBoost.   
Might be usefull to meet and discuss steps and details there.

trivialfis (approved):

PR #10843, opened 2024-09-24
[EM] Improve memory estimation for quantile sketching.

trivialfis:
- Add basic estimation for RMM.
- Re-estimate after every sub-batch.
- Some debug logs for memory usage.
- Fix the locking mechanism in the memory allocator logger.


trivialfis:
cc @rongou .

rongou (approved):

PR #10849, opened 2024-09-25
[R] Rename `ExternalDMatrix` -> `ExtMemDMatrix`.

trivialfis:
We will have an external memory version of `QuantileDMatrix`, which is named `ExtMemQuantileDMatrix` to reflect the use of external memory with data being fetched on-demand during training instead of simply using an iterator for construction.


trivialfis:
@david-cortes @mayer79 Could you please help take a look when you are available?

mayer79:
LGTM

PR #10850, opened 2024-09-25
Use CUDA virtual memory for pinned memory allocator.

trivialfis:
This is to avoid obtaining CUDA memory-related global locks (like those used by `cudaFree`), and to fix https://github.com/dmlc/xgboost/issues/10312 . Only works with `cudart>=12.5`. The implementation is based on this blog [post](https://developer.nvidia.com/blog/introducing-low-level-gpu-virtual-memory-management/), with additional support for host numa allocation.

Running MGPU column split tests, haven't been able to observe any hang with the PR. (yet)

- Add a grow-only virtual memory allocator.
- Define a driver API wrapper. Split up the runtime API wrapper.

### DIFF
--- a/src/common/threading_utils.cc
+++ b/src/common/threading_utils.cc
@@ -9,10 +9,12 @@
 #include <fstream>     // for ifstream
 #include <string>      // for string
 
-#include "common.h"           // for DivRoundUp
+#include "common.h"  // for DivRoundUp
 
 #if defined(__linux__)
 #include <pthread.h>
+#include <sys/syscall.h>  // for SYS_getcpu
+#include <unistd.h>       // for syscall
 #endif
 
 namespace xgboost::common {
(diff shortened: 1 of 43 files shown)


trivialfis:
At the moment, this is not being tested on the CI as the feature requires CTK12.5+ to be stable.

trivialfis:
cc @hcho3 @rongou .

src/common/threading_utils.cc
```diff
@@ -118,6 +120,14 @@ std::int32_t OmpGetNumThreads(std::int32_t n_threads) {
   return n_threads;
 }
 
+[[nodiscard]] bool GetCpuNuma(unsigned int* cpu, unsigned int* numa) {
```
trivialfis, line 123: This is not used, I keep it to check the numa node when debugging.

rongou (approved):

PR #10861, opened 2024-09-30
[EM] Add basic distributed GPU tests.

trivialfis:
- Split Hist and Approx tests in unittests.
- Basic GPU tests for distributed.


trivialfis:
cc @rongou .

I'm working on optimizing ellpack to better handle relatively dense but not dense data. Need to upstream some tests and refactoring first to avoid an over-complicated PR.

rongou (approved):

PR #10870, opened 2024-10-01
[EM] Treat ellpack as dense matrix when there's no compression.

trivialfis:
This enables XGBoost to use shared memory to build histograms. In addition, it lowers memory usage and speeds up external memory for data that cannot be compressed using the ellpack format.

- From now on, only datasets with at least one missing value for every sample are considered sparse.
- We need to distinguish fully dense, mostly dense, and sparse. This is similar to what the CPU implementation currently does.
- Some cleanups


trivialfis:
cc @rongou

rongou (approved):

PR #10873, opened 2024-10-05
Remove all use of custom thrust exec.

trivialfis:
- Avoid `cudaMalloc` in thrust.
- Avoid synchronization.

I have checked regression model training using gdb, and there's no more `cudaMalloc` calls in thrust. But we might still have lingering thrust calls that don't have any execution policy in other model types. Also, NCCL might allocate memory using cuda driver API.


trivialfis:
cc @rongou

rongou (approved):

PR #10876, opened 2024-10-08
[EM] Improve external iterator.

trivialfis:
- Use bool instead of int in the Python interface.
- Unify the internal advance operator for the ext qdm between CPU and GPU.
- Avoid the use of `n_batches` calculated from the external iterator and use the cache size instead. We will add support for internal page concatenation for ellpack to handle irregular batch sizes from distributed frameworks.


trivialfis:
cc @rongou .

rongou (approved):

PR #10878, opened 2024-10-08
Make the span class trivially copyable.

trivialfis:
This way, we can use the uvector in more places and catch some more unused variables.


trivialfis:
cc @rongou

rongou (approved):

PR #10884, opened 2024-10-11
SYCL. Minimize host-device memory synchronization for ColumnSampler

razdoburdin:
This simple PR improve training speed-up on sycl-devices for `plastic `dataset on ~1.4x and for `santander `on ~1.2x by reducing the host-device memory synchronizations.

### DIFF
--- a/src/common/random.h
+++ b/src/common/random.h
@@ -177,7 +177,10 @@ class ColumnSampler {
     }
     Reset();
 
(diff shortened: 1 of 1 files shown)


trivialfis (approved):
Will merge. But I can't guarantee this type of optimization won't break in the future though, it's not uncommon to pre-fetch data to the destinated device.

PR #10887, opened 2024-10-12
[EM] Concatenate ellpack pages for ExtMemQdm.

trivialfis:
The concatenation only happens if ExtMemQdm is used with a host cache. We will work on other types of cache for consistency in the future. For now, the host cache in ExtMemQdm is the most important use case.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -537,7 +537,7 @@ class DataIter(ABC):  # pylint: disable=too-many-instance-attributes
             This is an experimental parameter.
 
     min_cache_page_bytes :
-        The minimum number of bytes of each cached pages Only used for on-host cache
+        The minimum number of bytes of each cached pages. Only used for on-host cache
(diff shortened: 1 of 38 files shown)


python-package/xgboost/core.py
```diff
@@ -536,16 +536,34 @@ class DataIter(ABC):  # pylint: disable=too-many-instance-attributes
 
             This is an experimental parameter.
 
+    min_cache_page_bytes :
+        Only used for on-host cache with GPU. The minimum number of bytes of each cached
+        pages. When using GPU-based external memory with data cached in the host memory,
+        XGBoost can concatenate the pages internally to increase the batch size for the
+        GPU. The page concatenation is enabled by default when all conditions are
+        satisfied and is set to about a 1/8 of the total device memory. Users can
+        manually set the values based on the actual hardware. If it's set to 0, then no
```
trivialfis, line 545:
```suggestion
        manually set the value based on the actual hardware. If it's set to 0, then no
```

trivialfis:
cc @rongou .

rongou (approved):

PR #10895, opened 2024-10-15
[EM] Push for higher device memory utilization.

trivialfis:
- Allow some pages to be in device memory for the validation dataset.
- Allow bounding the stream length for quantile sketching.
- Rename the parameter `extmem_concat_pages` to `extmem_single_page` to avoid confusion with the cache optimization.

Along with recent optimizations in the sketching, we should be able to close https://github.com/dmlc/xgboost/issues/6759 .


trivialfis:
cc @rongou .

rongou (approved):

PR #10902, opened 2024-10-16
[BP] Ensure that `pip check` does not fail due to bad platform tag (#10755)

trivialfis:


trivialfis:
@hcho3 Could you please help check whether the patch is correct? I'm a bit confused by how this works.

hcho3 (approved):
The patch looks good to me

PR #10918, opened 2024-10-22
[EM] Add CPU categorical feature support and data validations.

trivialfis:
- Add support for extmem with categorical features for the CPU implementation.
- Validate the feature types for each batch.


trivialfis:
cc @rongou .

rongou (approved):

PR #10923, opened 2024-10-23
Define guard macros to prevent windows.h from conflicting with winsock2.h

hcho3:
When building XGBoost on Windows with CUDA 12.5 and standalone CCCL (*), I get the following error:
```
error : invalid redeclaration of type name "sockaddr"
```
(from https://github.com/hcho3/xgboost-ci-ng/actions/runs/11471864036/job/31923485369)

The error occurs because `windows.h` includes `winsock.h`, which conflicts with `winsock2.h`. See https://stackoverflow.com/questions/1372480/c-redefinition-header-files-winsock2-h. Where does `windows.h` come from? Most likely from the standalone CCCL, as the error didn't occur when using the vendored CCCL from the CTK.

Fix. Define macros `_WINSOCKAPI_` so that `windows.h` doesn't include `winsock.h`. In addition, define macros `NOMINMAX`, `WIN32_LEAN_AND_MEAN`, and `_WINSOCKAPI_` as part of the CMake build so that they go into effect before CCCL headers are included.

Also: Use `INVALID_SOCKET` instead of `-1` to represent invalid sockets on Windows. See https://learn.microsoft.com/en-us/windows/win32/winsock/socket-data-type-2.

(\*) I use CCCL 2.6.1 from GitHub because CUDA 12.5 ships with a buggy version of CCCL (https://github.com/NVIDIA/cccl/issues/1956)

### DIFF
--- a/include/xgboost/collective/socket.h
+++ b/include/xgboost/collective/socket.h
@@ -99,6 +99,7 @@ inline auto ThrowAtError(StringView fn_name, std::int32_t errsv = LastError()) {
 using SocketT = SOCKET;
 #else
 using SocketT = int;
+#define INVALID_SOCKET -1
 #endif  // defined(_WIN32)
 
 #if !defined(xgboost_CHECK_SYS_CALL)
@@ -276,7 +277,7 @@ class TCPSocket {
   SockDomain domain_{SockDomain::kV4};
 #endif
 
-  constexpr static HandleT InvalidSocket() { return -1; }
+  constexpr static HandleT InvalidSocket() { return INVALID_SOCKET; }
 
   explicit TCPSocket(HandleT newfd) : handle_{newfd} {}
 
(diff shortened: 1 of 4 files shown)


include/xgboost/collective/socket.h
```diff
@@ -99,6 +99,7 @@ inline auto ThrowAtError(StringView fn_name, std::int32_t errsv = LastError()) {
 using SocketT = SOCKET;
 #else
 using SocketT = int;
+#define INVALID_SOCKET -1
```
trivialfis, line 102: Is this defined when windows is used?

trivialfis (approved):

trivialfis (approved):

hcho3:
@trivialfis Can you review the latest commit? CUDA virtual memory isn't used in Windows so one of gtest had to be updated.

trivialfis (approved):

PR #10924, opened 2024-10-23
[EM][doc] Update document for the GPU implementation.

trivialfis:
Previously I planned to add interpolation between CPU and GPU inputs. After some thought, I believe it's not worth the effort for now. If one is using external memory, the data is likely to be large. In such a case, the GPU implementation can use up all the memory in both the host and device, making another copy of data is unwise.

### DIFF
--- a/doc/tutorials/external_memory.rst
+++ b/doc/tutorials/external_memory.rst
@@ -67,7 +67,7 @@ external memory training, users need to define a data iterator with 2 class meth
 
       """
       if self._it == len(self._file_paths):
-        # return 0 to let XGBoost know this is the end of the iteration
+        # return False to let XGBoost know this is the end of the iteration
         return False
 
       # input_data is a function passed in by XGBoost and has the exact same signature of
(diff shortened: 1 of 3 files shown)


doc/tutorials/external_memory.rst
```diff
@@ -54,42 +54,43 @@ external memory training, users need to define a data iterator with 2 class meth
...
       super().__init__(cache_prefix=os.path.join(".", "cache"))
 
-    def next(self, input_data: Callable):
+    def next(self, input_data: Callable) -> bool:
       """Advance the iterator by 1 step and pass the data to XGBoost. This function is
       called by XGBoost during the construction of ``DMatrix``
 
       """
       if self._it == len(self._file_paths):
         # return 0 to let XGBoost know this is the end of the iteration
```
trivialfis, line 70:
```suggestion
        # return False to let XGBoost know this is the end of the iteration
```

trivialfis:
cc @rongou

rongou (approved):

PR #10925, opened 2024-10-24
[jvm-packages] fix transform performance issue

wbo4958:
The latest xgboost spark has a transform performance regression. It  took about 220s to predict  300, 000 rows in a single task, With this PR, it just took a few seconds for the same spark application code, same spark submit script, same cluster.

### DIFF
--- a/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostEstimator.scala
+++ b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostEstimator.scala
@@ -136,6 +136,7 @@ private[spark] trait XGBoostEstimator[
 
   /**
    * Sort partition for Ranker issue.
+   *
    * @param dataset
    * @return
    */
@@ -320,6 +321,7 @@ private[spark] trait XGBoostEstimator[
       trainRDD.zipPartitions(evalRDD) { (left, right) =>
         new Iterator[Watches] {
           override def hasNext: Boolean = left.hasNext
+
           override def next(): Watches = {
             val trainDMatrix = buildDMatrix(left)
             val evalDMatrix = buildDMatrix(right)
@@ -332,6 +334,7 @@ private[spark] trait XGBoostEstimator[
       trainRDD.mapPartitions { iter =>
         new Iterator[Watches] {
           override def hasNext: Boolean = iter.hasNext
+
           override def next(): Watches = {
             val dm = buildDMatrix(iter)
             new Watches(Array(dm), Array(Utils.TRAIN_NAME), None)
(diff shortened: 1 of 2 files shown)


wbo4958:
Hi @trivialfis, Could you help review it.

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostEstimator.scala
```diff
@@ -527,24 +530,54 @@ private[spark] trait XGBoostModel[M <: XGBoostModel[M]] extends Model[M] with ML
...
+    //    tmpIters.foldLeft(originalRowIter) { case (accIter, nextIter) =>
+    //      // Zip the accumulated iterator with the next iterator
+    //      accIter.zip(nextIter).map { case (a: Row, b: Row) =>
+    //        Row.fromSeq(a.toSeq ++ b.toSeq)
+    //      }
+    //    }
+
+    tmpIters.size match {
+      case 4 =>
+        originalRowIter.zip(tmpIters(0)).zip(tmpIters(1)).zip(tmpIters(2)).zip(tmpIters(3)).map {
```
trivialfis, line 559: Could you please add some comments what you are doing here? This is not the easiest to read code with symbols `a,b,c,d,e` being concatenated.

trivialfis (changes requested):

jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostEstimator.scala
wbo4958: This piece of code is an optimization for 
``` scala
    // This is not so efficient considering that toSeq from first iterators will be called
    // many times.
    //    tmpIters.foldLeft(originalRowIter) { case (accIter, nextIter) =>
    //      // Zip the accumulated iterator with the next iterator
    //      accIter.zip(nextIter).map { case (a: Row, b: Row) =>
    //        Row.fromSeq(a.toSeq ++ b.toSeq)
    //      }
    //    }
```

to zip some iterators to a new iterator. For example

``` scala
iter1 = Seq("a", "b", c").iterator
iter2 = Seq("1", "2", "3").iterator
iter3 = Seq("xa", "xb", "xc").iterator
```

`iter1.zip(iter2).zip(iter3).map ...` could get a new iterator with content like 


``` scala
Seq(("a", "1", "xa"),   
        ("b", "2", "xb"),
        ("c", "3", "xc"))`
```

trivialfis: That's unfortunate. Reminds me of lisp ` (car (cdr (car ...)))`. Are you certain that there's no way to make this more explicit (like using a helper function with a name).?

wbo4958: No, we can't do that since we have no idea about the elements in the Array,  It could be originalRowIterator, rawPredictions, or originalRowIterator, leaf predictions.

trivialfis: Would love to learn more about Scala one day, the macro implementation there looks interesting.

trivialfis (approved):

PR #10927, opened 2024-10-24
Simplify the GPU memory logger.

trivialfis:
Simplify the logger to only keep tracker of the current and peak memory usage. The current locking mechanism doesn't guarantee the ordering of the lock and it's difficult to fix.

We can use rmm or nsys for detailed memory profiling in the future.


trivialfis:
cc @rongou .

rongou (approved):

PR #10930, opened 2024-10-25
Add support for CUDA virtual device memory.

trivialfis:
This builds upon the existing host-based virtual memory allocator. The feature is not used yet. It's only here as a backup solution if there's no sound mitigation for the memory fragmentation in the CUDA memory pool.


trivialfis:
cc @rongou

rongou (approved):

PR #10935, opened 2024-10-27
[breaking] Remove dask from the default import.

trivialfis:
This aligns with the spark package and simplifies the imports significantly. We will work on improved learning to rank support and basic resilience.

This is a breaking change as:

``` python
xgboost.dask.DaskDMatrix()
```
will stop working. One needs to import the dask module independently:
``` python
from xgboost.dask import DaskDMatrix
from xgboost import dask as dxgb
```


trivialfis:
@wphicks For both simplification with consistency with the spark package.

wphicks (approved):

PR #10937, opened 2024-10-28
[jvm-packages] support array features type for cpu

wbo4958:
This PR introduces a new feature that can support array input type for both training and transform. Users can not only train with vector but also train with array, and the model also can predict the array or vector input.


wbo4958:
Hi @trivialfis, could you help review it. Thx

trivialfis (approved):

PR #10945, opened 2024-10-30
[coll] Support timeout for nccl calls.

trivialfis:
- Use nccl asynchronous communicator.
- Abort the communicator when timeout.
- Update NCCL on CI for cpp tests.

nccl 2.23+ is required. XGBoost will continue to work with older versions of nccl, but upon timeout, it will crash, which is still an improvement over hang.

todo:
- [x] perf check.


trivialfis:
Thus far, the behavior is not quite reliable and the documentation around the async communicator is sparse. I will push back the work on this, right now it's mostly for testing.

tests/ci_build/Dockerfile.gpu
```diff
@@ -43,6 +43,9 @@ RUN set -ex; \
     chmod +x /usr/local/bin/gosu && \
     gosu nobody true
 
+RUN \
```
trivialfis, line 46: @hcho3 I won't merge this PR any time soon. However, I found the NCCL version used for the CPP test is quite old and is from the base image.

trivialfis:
cc @rongou .

rongou (approved):

PR #10978, opened 2024-11-03
[jvm-packages] Update javadoc plugin.

trivialfis:
supersede https://github.com/dmlc/xgboost/pull/10976 .


trivialfis:
@wbo4958 Could you please help take a look into the errors when you are available? I'm not sure how to resolve it during mvn test run:
```
[WARNING] Javadoc Warnings
[WARNING] /workspace/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/DMatrix.java:20: error: cannot find symbol
[WARNING] import ml.dmlc.xgboost4j.LabeledPoint;
[WARNING] ^
[WARNING] symbol:   class LabeledPoint
[WARNING] location: package ml.dmlc.xgboost4j
[WARNING] /workspace/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/DMatrix.java:38: error: cannot find symbol
[WARNING] public DMatrix(Iterator<LabeledPoint> iter, String cacheInfo) throws XGBoostError {
[WARNING] ^
[WARNING] symbol:   class LabeledPoint
```

wbo4958:
I'm really not a fan of updating the plugin or dependency frequently. that could bring more work for us since the new release of the plugin probably has some issue which may cause compiling and testing fail.

hcho3:
@wbo4958 Should we disable the dependabot entirely? It was added because some users use security scanners that scans CVEs in the dependencies.

trivialfis:
I will look into the fix.

wbo4958:
Hi @hcho3, I'm wondering if the javadoc compiling is working before? I just tried it without updating the version of maven-javadoc-plugin and it didn't work for the latest master HEAD

wbo4958:
The problem here is the javadoc couldn't find the scala class (LabeledPoint)

wbo4958:
To be honest, I didn't find any solution for this issue. I'm wondering if we need to move LabeledPoint from scala to java to fix this issue.

trivialfis:
Yeah, I thought about that as well, was thinking if you know some magic for a simpler fix

hcho3:
Yes, Javadoc hasn't been able to generate document for LabeledPoint for a while. The status quo is to ignore the warning and get a Javadoc without any documentation for LabeledPoint.

wbo4958:
Hi @hcho3, What's the final decision on this issue? should we move the LabeledPoint to java or just ignore the warning?

hcho3:
We can probably ignore the warning for now.

PR #10973, opened 2024-11-02
[coll] Fix tracker port and thread init.

trivialfis:

PR #10983, opened 2024-11-05
[coll] Expose configuration.

trivialfis:
This PR adds a configuration parameter to the dask interface for specifying communicator configurations.

Stacked on top of https://github.com/dmlc/xgboost/pull/10973 .

~todos~
- [ ] ~spark~

### DIFF
--- a/python-package/xgboost/collective.py
+++ b/python-package/xgboost/collective.py
@@ -75,7 +75,8 @@ class Config:
                 return v
             if not isinstance(v, type(typ)):
                 raise TypeError(
-                    f"Invalid type for configuration {v}, expecting {typ}, got {type(v)}"
+                    f"Invalid type for configuration `{key}`, "
+                    f"expecting {type(typ).__name__}, got {type(v).__name__}."
                 )
             return v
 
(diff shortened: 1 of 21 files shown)


python-package/xgboost/collective.py
```diff
@@ -15,7 +16,41 @@
 LOGGER = logging.getLogger("[xgboost.collective]")
 
 
-def init(**args: Any) -> None:
+_ArgVals: TypeAlias = Optional[Union[int, str]]
+_Args: TypeAlias = Dict[str, _ArgVals]
+
+
+@dataclass
+class Config:
```
wbo4958, line 24: Could we rename it make it more readable?

wbo4958: And could we add json support in it?

trivialfis: `xgboost.collective.Config` seems readable to me.

trivialfis:
@wbo4958 Might be related https://github.com/dmlc/xgboost/issues/10986 .

python-package/xgboost/collective.py
trivialfis: Removed the converters. We can use the utilities from `dataclasses`, see the collective test for an exmaple.

trivialfis:
cc @rongou .

rongou (approved):

PR #10984, opened 2024-11-06
[spark] Make xgboost spark support large model size

WeichenXu123:
Spark RDD can't support one line with very long content.

To make large size model training / saving / loading works,
I split model json string to chunks when collecting model in training, and modify saving / loading code too.


python-package/xgboost/spark/core.py
```diff
@@ -1095,12 +1095,15 @@ def _train_booster(
...
-                    data={
-                        "config": [booster.save_config()],
-                        "booster": [booster.save_raw("json").decode("utf-8")],
-                    }
-                )
+                config = booster.save_config()
+                yield pd.DataFrame({"data": [config]})
+                booster = booster.save_raw("json").decode("utf-8")
+
+                chunk_size = 4096 * 1024
```
wbo4958, line 1102: Could we make chunk_size to be global variable, since the below somewhere is still using it.

wbo4958:
LGTM for the functionality except the CI issue.

python-package/xgboost/spark/core.py
```diff
@@ -1110,7 +1110,7 @@ def _run_job() -> Tuple[str, str]:
             rdd = (
                 dataset.mapInPandas(
                     _train_booster,  # type: ignore
-                    schema="config string, booster string",
+                    schema="data string",
```
wbo4958, line 1113: good finding

wbo4958:
Could you run `python tests/ci_build/lint_python.py --format=1 --type-check=1 --pylint=1` to check the python format

WeichenXu123:
I can't fully understand the linter error:
```
xgboost/spark/core.py:1162: error: Incompatible types in assignment (expression has type "str", variable has type "Booster")  [assignment]
xgboost/spark/core.py:1164: error: Argument 1 to "len" has incompatible type "Booster"; expected "Sized"  [arg-type]
Found 2 errors in 1 file (checked 41 source files)
```
@wbo4958 any ideas ?

trivialfis:
@WeichenXu123 XGBoost's Python package uses Python typehint. In the following line:
``` python
booster = booster.save_raw("json").decode("utf-8")
```
The `booster` was a `xgboost.Booster` object, the `decode("utf-8")` however, returns a string. Assigning a string to a `Booster` type violates static typing.

python-package/xgboost/spark/core.py
```diff
@@ -597,6 +597,9 @@ def _get_unwrapped_vec_cols(feature_col: Column) -> List[Column]:
 )
 
 
+_MODEL_CHUNK_SIZE = 4096 * 1024
```
wbo4958, line 600: I was wondering if 4M is an appropriate size for the normal size model? @trivialfis any idea about it?

WeichenXu123: the chunk size is for making spark processing easier, i.e. one chunk becomes one row in spark dataframe. So the chunk_size should be determined by the proper spark row size right ?

wbo4958: Yeah, make sense, so what's the proper spark row size that can make spark process easier?

If the model size is like 1G, then it will be split to 256 rows. should that have some perf issue?

WeichenXu123: no perf issue. very safe. it just optimizes the row size so that it improves pyspark dataframe ser / deser processing.

WeichenXu123: 4MB is a good row size according to our experience

wbo4958:
LGTM if the CI can pass

wbo4958 (approved):

WeichenXu123:
@trivialfis Can we make a patch release to include this fix ? We have several customers facing the issue. thanks!

PR #10985, opened 2024-11-06
[dask] Prevent the training from hang due to aborted worker.

trivialfis:
Set `allow_other_workers=True` in `client.submit`. See code comments for the rationale.

### DIFF
--- a/python-package/xgboost/dask/__init__.py
+++ b/python-package/xgboost/dask/__init__.py
@@ -33,13 +33,11 @@ Optional dask configuration
 """
 import collections
 import logging
-import platform
 from collections import defaultdict
 from contextlib import contextmanager
-from functools import partial, update_wrapper
+from functools import partial, update_wrapper, wraps
 from threading import Thread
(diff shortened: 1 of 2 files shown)


python-package/xgboost/dask/__init__.py
```diff
@@ -506,13 +506,46 @@ async def map_worker_partitions(
...
+            # Turn result into a list for bag construction
+            worker = distributed.get_worker()
+
+            if worker.address != address:
+                raise ValueError(
+                    f"Invalid worker address: {worker.address}, expecting {address}. "
+                    "This is likely caused by one of the workers died and Dask "
+                    "re-scheduled a different one. Resilience is not yet supported."
+                )
+            return [func(*args, **kwargs)]
```
hendrikmakait, line 523: Will the functions that run on existing workers wait indefinitely for the dead worker to appear, or will they start failing at some point?

wphicks (approved):

PR #10993, opened 2024-11-12
SYCL. Unify calculations for objectives reg:absoluteerror, reg:quantileerror, binary:hinge

razdoburdin:
Hi, I work on unification of sycl-plugin with the main part of the framework.
This PR unify calculations for some objective functions: `reg:absoluteerror`, `reg:quantileerror` and `binary:hinge`

The procedure of unification occurs in the following way:
1. source files with objective functions are compiled with sycl compiler in case sycl-plugin is enabled.
2. Instead of the original implementation of `linalg::ElementWiseKernel` the sycl realisation is used.


src/common/linalg_op.h
```diff
@@ -49,6 +49,7 @@ void ElementWiseKernelHost(linalg::TensorView<T, D> t, std::int32_t n_threads, F
 }
 
 #if !defined(XGBOOST_USE_CUDA)
+#if !defined(XGBOOST_USE_SYCL)
```
trivialfis, line 52:
```
#if !(use_cuda or use_sycl)
```

src/metric/elementwise_metric.cu
```diff
@@ -45,7 +45,7 @@ namespace {
 template <typename Fn>
 PackedReduceResult Reduce(Context const* ctx, MetaInfo const& info, Fn&& loss) {
   PackedReduceResult result;
-  auto labels = info.labels.View(ctx->Device());
+  auto labels = info.labels.View(ctx->Device().IsSycl() ? DeviceOrd::CPU() : ctx->Device());
```
trivialfis, line 48: I assume you want to use `IsCPU` instead of `IsSycl`?

plugin/sycl/common/linalg_op.h
```diff
@@ -40,7 +44,44 @@ ::sycl::event GroupWiseKernel(::sycl::queue* qu, int* flag_ptr,
...
+template<typename Fn, typename TupleType, size_t ... I>
+auto call(Fn&& fn, TupleType t, std::index_sequence<I ...>) {
+     return fn(std::get<I>(t) ...);
+}
+
+template<typename Fn, typename TupleType>
+auto call(Fn&& fn, TupleType t) {
+    static constexpr auto size = std::tuple_size<TupleType>::value;
+    return call(fn, t, std::make_index_sequence<size>{});
+}
```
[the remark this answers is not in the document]
razdoburdin: It is, but `std::apply` can not be executed in sycl-kernel.

src/common/linalg_op.h
razdoburdin: done

src/metric/elementwise_metric.cu
razdoburdin: No. This function doesn't have sycl-specific implementation yet. So for the proper execution on `cpu `the data should be transferred back to host.
I plan to add sycl-implementation for it in the future. Didn't do it by single PR just to avoid unreadable PR size.

razdoburdin:
CI failures are now reasoned by exceeding the daily budget: 
image

Issue #10997, opened 2024-11-13
XGBoost Python API ignoring random state 


mbarsacchi:
Hello! 
I've noticed that, since version 2.1.0, Xgboost Python API has been ignoring the random seed, causing the results to be deterministic (even if `colsample_by_node is applied`  - this issue does not apply to `subsample`).

You can reproduce by running this
```
from numpy import loadtxt
from xgboost import XGBClassifier


# data available https://raw.githubusercontent.com/jbrownlee/Datasets/master/pima-indians-diabetes.data.csv
dataset = loadtxt('pima-indians-diabetes.csv', delimiter=",")


X = dataset[:,0:8]
Y = dataset[:,8]

model1 = XGBClassifier(n_estimators=2, learning_rate=0.04, max_depth=9,tree_method="hist", colsample_bynode=0.5, random_state=42)
model1.fit(X, Y)

model2 = XGBClassifier(n_estimators=2, learning_rate=0.04, max_depth=9,tree_method="hist", colsample_bynode=0.5, random_state=43)
model2.fit(X, Y)

assert list(model1.feature_importances_) != list(model2.feature_importances_)
```
The feature importances differ on version 2.0.3, but they do not on 2.1.0. No matter what `random_state` I use, I get the same results.

I've managed to narrow it down to this specific commit: https://github.com/dmlc/xgboost/commit/fedd9674c8441a26b94f9d4061b11a1be4cb0e7c

Checking out the commit before (178cfe70a8bf6ca0081032cf0de47e5219d0cf08) and compiling still works as intended (two importances are different). Once you check out the commit (linked before - and any of the following one) and compile, the assert fails.

--- 
This is the process I followed every time:
```
git checkout <commit-hash>
git submodule update
cmake -B build -S . -GNinja
cmake --build build
# Install as editable installation
cd ./python-package
pip install -e .
```

trivialfis:
Thank you for sharing, will look into it. Perhaps the sampler is initialized before the global rng

trivialfis:
Opened a PR for the fix: https://github.com/dmlc/xgboost/pull/10998 Will be the next patch release.

PR #10998, opened 2024-11-14
Fix rng for the column sampler.

trivialfis:
- Use `GlobalRandom`
- Small cleanup for the GPU ctors.

Close https://github.com/dmlc/xgboost/issues/10997 .

### DIFF
--- a/src/common/random.h
+++ b/src/common/random.h
@@ -233,7 +233,7 @@ class ColumnSampler {
 };
 
 inline auto MakeColumnSampler(Context const* ctx) {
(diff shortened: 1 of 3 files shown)


src/common/random.h
```diff
@@ -233,7 +233,7 @@ class ColumnSampler {
 };
 
 inline auto MakeColumnSampler(Context const* ctx) {
-  std::uint32_t seed = common::GlobalRandomEngine()();
```
trivialfis, line 236: For the longer term, we should remove this global rng.

rongou (approved):

PR #11002, opened 2024-11-19
Small enhancements for doc and error messages.

trivialfis:
- Use the same setup_rmm for external memory demos.
- Display timeout seconds in nccl timeout.


trivialfis:
cc @rongou .

rongou (approved):

PR #11003, opened 2024-11-19
[pyspark] Support collective.Conf in spark

wbo4958:
This PR is a followup of https://github.com/dmlc/xgboost/pull/10983

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -259,12 +259,12 @@ class _SparkXGBParams(
     collective_conf = Param(
         Params._dummy(),
         "collective_conf",
-        "xgboost.collective.Config. The communicator configuration.",
+        "xgboost.collective.Config. The collective configuration.",
         TypeConverters.identity,
     )
 
     def set_collective_conf(self, value: Config) -> "_SparkXGBParams":
-        """Set communicator configuration"""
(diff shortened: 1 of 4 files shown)


python-package/xgboost/spark/core.py
```diff
@@ -257,21 +258,20 @@ class _SparkXGBParams(
         "launched on the driver side; otherwise, it will be launched on the executor side.",
         TypeConverters.toBoolean,
     )
-    tracker_host_ip = Param(
+    tracker = Param(
```
wbo4958, line 261: the name "tracker" is ok?

trivialfis: I would probably continue to use the name `coll_cfg` just to be consistent. Besides, the configuration is for both the tracker and the communicators.

wbo4958: sure, I changed it to `collective_conf`

trivialfis: I know everyone has a naming preference ...  if you use a new name, please modify the existing ones in Dask for consistency.

wbo4958: OK, I just changed it to coll_cfg to be compatible with dask.

trivialfis (approved):
Thank you for the PR!

PR #11007, opened 2024-11-19
[Dask] Sort and partition for ranking.

trivialfis:
- Implement automatic local sort.
- Implement partitioning by query ID.
- Document for distributed ranking.

Please see the changes in the tutorial about the new features. In summary, we can now have two different ways to handle query groups in the Dask interface, depending on whether a global sort is desired.

Please note that, this PR ensures the accuracy of the model at a heavy performance price. A global sort is costly and pushing huge amount of small partitions into the `QuantileDMatrix` is also inefficient.

*related*:
- [ ] https://github.com/dmlc/XGBoost.jl/issues/160
- [ ] https://github.com/dmlc/xgboost/pull/9491

In addition to this PR, I need to work on a distributed version of group weight.


trivialfis:
cc @rongou .

rongou (approved):

PR #11014, opened 2024-11-22
Handle `cudf.pandas` proxy objects properly

galipremsagar:
`cudf.pandas` proxy objects will not pass `is_cudf....` checks and rather they tend to only pass `is_pandas...`. But the pandas code-flows have many `to_numpy` calls that are introducing slow-downs incase of `cudf.pandas` objects. This PR bypasses this problem by introducing a mechanism to extract proper `cudf` objects out of `cudf.pandas` objects. That way there is no slow-down.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -2681,7 +2681,7 @@ class Booster:
         )
 
         if _is_cudf_pandas_df(data) or _is_cudf_pandas_ser(data):
-            data = data._fsproxy_fast
+            data = data._fsproxy_fast  # pylint: disable=protected-access
 
         enable_categorical = True
         if _is_arrow(data):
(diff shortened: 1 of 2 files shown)


hcho3 (approved):

hcho3:
@trivialfis Can you review this?

trivialfis:
Added this to the 2.1.3 release. Will push some changes at Monday for `lazy_isinstance` along with a test.

trivialfis:
Changes from my commit:
- Added support for DMatrix.
- Use `lazy_isinstance`.

trivialfis:
Please help test and review the change.

galipremsagar:
> Please help test and review the change.

Can confirm this works locally.

PR #11016, opened 2024-11-25
SYCL. Unify regression objective calculation

razdoburdin:
In continue for https://github.com/dmlc/xgboost/pull/10993.

This PR unify calculations of regression objectives, so sycl-specific implementation is no longer needed.


trivialfis (approved):

trivialfis:
Let's ignore it for this PR.

PR #11023, opened 2024-11-27
[jvm-packages] LTR: distribute the features with same group into same partition

wbo4958:
To fix https://github.com/dmlc/xgboost/issues/6713

This PR leverages repartitionByRange to replace repartition for ranking problem, which can guarantee the feature instances in the same group can be distributed into the same partition.


trivialfis:
@wbo4958 The PR looks good to me. How difficult would it be to run a test with an actual dataset? Would love to see the accuracy as well. Consider the ms dataset https://github.com/dmlc/xgboost/pull/11007/files .

trivialfis (approved):

PR #11047, opened 2024-12-03
[pyspark] LTR: distribute the features with same group into same partition

wbo4958:
To fix https://github.com/dmlc/xgboost/issues/6713 in xgboost pyspark


trivialfis:
I need to work on the qid weight a little bit, should not take too much time.

trivialfis (approved):

PR #11065, opened 2024-12-06
[R] Add evaluation set and early stopping for `xgboost()`

david-cortes:

This PR adds the evaluation set and early stopping functionality to the new `xgboost()` function.

Due to the data processings that this new interface does (e.g. encoding of factors for both 'x' and 'y'), it only allows specifying the evaluation data as a subset of the x/y data, either as a random fraction or by selected row indices, as otherwise taking lists of data would require a lot of hassle with encodings, variable reorderings, and so on.


R-package/R/xgb.train.R
```diff
@@ -147,9 +147,15 @@
 #' @param print_every_n Print each nth iteration evaluation messages when `verbose>0`.
 #'   Default is 1 which means all messages are printed. This parameter is passed to the
 #'   [xgb.cb.print.evaluation()] callback.
-#' @param early_stopping_rounds If `NULL`, the early stopping function is not triggered.
-#'   If set to an integer `k`, training with a validation set will stop if the performance
-#'   doesn't improve for `k` rounds. Setting this parameter engages the [xgb.cb.early.stop()] callback.
+#' @param early_stopping_rounds Number of boosting rounds after which training will be stopped
```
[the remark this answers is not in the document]
david-cortes: It's the same in R. Updated the docs.

R-package/R/xgboost.R
```diff
@@ -512,6 +516,9 @@ check.can.use.qdm <- function(x, params) {
       return(FALSE)
     }
   }
+  if (NROW(eval_set)) {
```
trivialfis, line 519: What does this imply? If `eval_set` has a valid number of rows then we can't use qdm?

david-cortes: Yes, because it then slices the DMatrix that gets created.

trivialfis: Does the slicing need to happen after the DMatrix is created?

david-cortes: Yes, because otherwise there'd be issues with things like needing to make sure categorical 'y' and features have the same encodings between the two sets, objects from package `Matrix` returning a different class when they are sliced, and so on.

trivialfis: Hmm, so from your perspective the DMatrix is more suitable for slicing than built-in classes...

david-cortes: Unfortunately yes.

david-cortes:
Just realized that `verbosity` wasn't being passed from `xgboost()` to `params`. Fixed it here, so that there's only one verbosity parameter to control.

R-package/R/xgboost.R
trivialfis: That's indeed unfortunate.

trivialfis: I have a strong preference for not using the DMatrix slice method. The DMatrix can be sliced only because it's a CSR for the simplest case. But it's considered an internal cache for XGBoost instead of something that can be manipulated from outside. For instance, with external memory, the DMatrix actually represents an async fetcher of a disk file.

david-cortes: Well, the alternative would be to:
* Restructure the code such that 'x' and 'y' get processed earlier.
* Make it so that characters in 'x' get converted to factor before passing them to `xgb.DMatrix`, and so that `data.frames` get their subclasses (like `data.table`) removed beforehand.
* Add a custom slicer for `Matrix` classes. Either that, or additional castings of storage format, or pull an extra dependency for efficient slicing.

But it'd end up being inefficient either way. The moreso considering that on the R side, the slicing would happen with a vector of random indices on one of the following:
* A column-major matrix of 8-byte elements (likely slower than a CSR).
* A list of vectors (what a dataframe is behind the scenes).
* A CSC matrix, which first will get converted to that format from CSR (`Matrix` doesn't slice CSR directly), and the slice getting converted to COO.

It could in theory be more efficient to do the slicing in R for base `matrix` classes, but probably not so much for the others.

trivialfis: Any suggestion for the future implementation of CV optimization previously mentioned? It's designed for the the Qdm and the external memory version of Qdm.

david-cortes: Sounds good and quite helpful for the CV function indeed. But I don't think it's very relevant here, since unlike `xgb.cv`, (a) it's only doing two slicing operations and only one quantization/binning, (b) it doesn't accept an arbitrary `xgb.DMatrix` - instead, it creates it internally from a small subset of allowed classes.

Hence, it doesn't need to consider special cases like external memory or distributed mode, and there isn't too much room for improvement in terms of speed savings.

trivialfis: Makes sense. We will have a second cv function in the future for high-level inputs (like  data.table and iterator etc).

trivialfis (approved):
Will merge after the CI is back online.

PR #11069, opened 2024-12-07
Use static dim check for the array interface handler.

trivialfis:


hcho3:
Are there not cases where we want to use a dynamic (run-time) argument to `Shape` ?

trivialfis:
@hcho3 The `ArrayInterface` has static dim since all XGBoost inputs have a known upper bound on dimension.

hcho3 (approved):

PR #11070, opened 2024-12-07
[breaking] [py] Drop support for datatable.

trivialfis:
Related https://github.com/microsoft/LightGBM/issues/6662 .

Not to be confused with the R `data.table`, this is the Python version.

cc @jameslamb 

We are working on auto re-coding for categorical features with dataframe inputs, it's unlikely that we will have the same treatment for datatable. It's better to remove it so that we can have consistent behaviors across all dataframe implementations.


jameslamb (approved):
Agree, I think it's a good time for this, for all the reasons mentioned in that LightGBM issues you linked to.

I did this just to check:

```shell
git grep -i -E 'datatable|h2o'
```

I see a few other uses that should be removed:

https://github.com/dmlc/xgboost/blob/4b2001ed11bf9c51c5c1031bd486b16fa59a160e/python-package/pyproject.toml#L46

https://github.com/dmlc/xgboost/blob/4b2001ed11bf9c51c5c1031bd486b16fa59a160e/python-package/xgboost/testing/__init__.py#L155-L156

https://github.com/dmlc/xgboost/blob/4b2001ed11bf9c51c5c1031bd486b16fa59a160e/doc/python/python_intro.rst#L169-L171

trivialfis:
Yup, will prune all the related code and references.

trivialfis:
Will merge if there's no objection.

hcho3 (approved):

PR #11072, opened 2024-12-07
[R] Add parameters constructor

david-cortes:

This PR adds a utility function to return a list of XGBoost parameters, with docs about all of them copied from the .rst files with slight modifications considering that they are meant exclusively for the R package. I'm not entirely sure whether all of them are applicable though.

The idea behind this constructor function is to offer an IDE-friendly way of creating and discovering XGBoost parameters, with full in-package documentation as is typical in R packages. It should be possible to reuse these same docs later on in `xgboost()` as needed through `@inheritParams`, but for now I'm only using this for `xgb.train` and `xgb.cv`.

Since this introduces a new function with a somewhat similar name as a completely different method in the package, it also renames that method in the process in order to avoid confusions and IDE miscompletions.


trivialfis:
Since this PR concerns the user interface, would be great if we could get some help for review from an R user's perspective @mayer79 @jameslamb

R-package/R/xgb.train.R
```diff
@@ -512,3 +381,435 @@ xgb.train <- function(params = list(), data, nrounds, evals = list(),
...
+#' values. Note that this function will not perform any validation on the supplied arguments.
+#'
+#' If passing `NULL` for a given parameter (the default for all of them), then the default
+#' value for that parameter will be used. Default values are automatically determined by the
+#' XGBoost core library upon calls to [xgb.train()] or [xgb.cv()], and are subject to change
+#' over XGBoost library versions.
+#' @return A list with the entries that were passed non-NULL values. It is intended to
+#' be passed as argument `params` to [xgb.train()] or [xgb.cv()].
+#' @export
+#' @param objective (default=reg:squarederror)
```
mayer79, line 403:
```suggestion
#' @param objective (default="reg:squarederror")
```

mayer79:
Very good idea!

R-package/R/xgb.train.R
```diff
@@ -512,3 +381,435 @@ xgb.train <- function(params = list(), data, nrounds, evals = list(),
...
+#'
+#' Supported values are:
+#' - `"reg:squarederror"`: regression with squared loss.
+#' - `"reg:squaredlogerror"`: regression with squared log loss \eqn{\frac{1}{2}[log(pred + 1) - log(label + 1)]^2}.  All input labels are required to be greater than -1.  Also, see metric `rmsle` for possible issue  with this objective.
+#' - `"reg:logistic"`: logistic regression, output probability
+#' - `"reg:pseudohubererror"`: regression with Pseudo Huber loss, a twice differentiable alternative to absolute loss.
+#' - `"reg:absoluteerror"`: Regression with L1 error. When tree model is used, leaf value is refreshed after tree construction. If used in distributed training, the leaf value is calculated as the mean value from all workers, which is not guaranteed to be optimal.
+#'
+#'   Version added: 1.7.0
+#' - `"reg:quantileerror"`: Quantile loss, also known as "pinball loss". See later sections for its parameter and [Quantile Regression](https://xgboost.readthedocs.io/en/latest/python/examples/quantile_regression.html#sphx-glr-python-examples-quantile-regression-py) for a worked example.
```
trivialfis, line 418: This is a Python example, do we want to include it here?

david-cortes: I'd say python examples are still relevant.

trivialfis (approved):

jameslamb:
> *Since this PR concerns the user interface, would be great if we could get some help for review from an R user's perspective @mayer79 @jameslamb*

Thanks for the `@`, sorry I did not get to it before this was merged.

> *The idea behind this constructor function is to offer an IDE-friendly way of creating and discovering XGBoost parameters, with full in-package documentation as is typical in R packages*

I think it's a great idea, and would definitely be helpful for interactive development.

Some suggestions (to consider in follow-ups, given that this was already merged):

* it would be good to have explicit unit tests on this new function, not just rely on implicit coverage from other tests. For example:
  - *confirming that only non-`NULL` items show up in the return value*
  - *confirming that calling this with 0 arguments returns an empty list*
* I understand that existing names with `.` in them need to be preserved for backwards compatibility, but I'd stop adding new ones (e.g. I'd prefer `xgb_params` to `xgb.params`)
  - *to reduce the risk of "R thinks this is an S3 method" confusion like the ones that led to #9577*
* I'd have asked to change the line wrapping of the roxygen comments instead of ignoring the project's preferred line length with `# nolint` comments
   - *or at least, asked that the linter be allowed to cover the function signature, which it currently cannot because of this placement:*

https://github.com/dmlc/xgboost/blob/85ccb8f88a00c131b508b780750bd17b0ac98e04/R-package/R/xgb.train.R#L811

I also would have asked for a little research supporting it being safe to change `xgb.parameters` to `xgb.model.parameters` without a deprecation cycle. For example, it looks to me like nothing on CRAN is currently using `xgb.parameters` (https://github.com/search?q=org%3Acran%20%22xgb.parameters%22&type=code), so I guess it at least won't break anything on CRAN. (I don't recall where the discussion about `xgboost2` vs. releasing this as a new version of `xgboost` ended, disregard this last comment if the plan is to release a new package).

trivialfis:
@jameslamb Thank you for the suggestion, I have converted your comment into a new tracking issue https://github.com/dmlc/xgboost/issues/11083

PR #11074, opened 2024-12-07
[R] Error out on unrecognized arguments and `...` parameters

david-cortes:

closes https://github.com/dmlc/xgboost/pull/10563

As a follow up from https://github.com/dmlc/xgboost/pull/11072

(requires changes from that PR)

This PR modifies the logic of xgboost functions to:

* Accept training parameters only under the `params` argument (save for `xgboost()`).
* Error out when receiving an argument that is not used.
* Throw a warning instead of an error when the parameter is recognized as deprecated.
* Change `feval` to `custom_metric` like it was done in the python package.
* Use the checked `xgb.params` constructor for all tests and examples whenever possible.

# Motivation

After all the changes in the R interface, we've now refocused `xgb.train` to be a low-level function that should stay as close as possible to the xgboost core library and the other language bindings and which should be preferred for production usage, while we now have `xgboost()` which is the user-friendly version following R idioms and meant for interactive usage.

Given that `xgb.train()` is a more low-level and production-oriented function, one would not want to find surprising or strange behaviors on it, such as silently accepting a parameter which is not used, or having undocumented `...` parameters, or bumping into undefined behavior when something is passed as both `...` and `params`.

As another motivation, the python package does not allow `*args` or `**kwargs` to pass parameters to `xgb.train` - this would make the R interface calls look more similar to those of other interfaces.

Even in XGBoost's own test suite, there have been many tests that were using removed arguments and no one noticed due to this behavior of accepting all arguments. The changes here allowed catching a few more tests that had unused parameters.

Since it might not be obvious that some arguments have been renamed, and this will be the first CRAN release to use the renamed versions, it would be quite helpful to throw informative errors to users and library authors such that they would know how to change their code.

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -71,20 +71,12 @@ NVL <- function(x, val) {
 
 # Merges booster params with whatever is provided in ...
 # plus runs some checks
-check.booster.params <- function(params, ...) {
+check.booster.params <- function(params) {
   if (!identical(class(params), "list"))
     stop("params must be a list")
 
   # in R interface, allow for '.' instead of '_' in parameter names
   names(params) <- gsub(".", "_", names(params), fixed = TRUE)
-
-  # merge parameters from the params and the dots-expansion
-  dot_params <- list(...)
-  names(dot_params) <- gsub(".", "_", names(dot_params), fixed = TRUE)
-  if (length(intersect(names(params),
-                       names(dot_params))) > 0)
-    stop("Same parameters in 'params' and in the call are not allowed. Please check your 'params' list.")
-  params <- c(params, dot_params)
 
   # providing a parameter multiple times makes sense only for 'eval_metric'
   name_freqs <- table(names(params))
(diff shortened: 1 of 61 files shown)


R-package/R/utils.R
```diff
@@ -535,44 +529,88 @@ NULL
...
-  idx_lut <- all_match[idx_pars]
-  # which of idx_lut were the exact matches?
-  ex_match <- depr_par_lut[idx_lut, 1] %in% names(pars)
-  for (i in seq_along(idx_pars)) {
-    pars_par <- names(pars)[idx_pars[i]]
-    old_par <- depr_par_lut[idx_lut[i], 1]
-    new_par <- depr_par_lut[idx_lut[i], 2]
-    if (!ex_match[i]) {
-      warning("'", pars_par, "' was partially matched to '", old_par, "'")
+check.deprecation <- function(
```
trivialfis, line 563: Is there any test for this function?

david-cortes: The function is not tested directly as it uses the parent frame, but the tests currently there check that the function does what it is supposed to when called from inside `xgb.train`.

For example, if `evals` is not passed (either through `evals` or through `watchlist` which this function is automatically swapping), there would be no evaluation log left in the booster object.

trivialfis (approved):

Issue #9925, opened 2023-12-26
[R] Cannot plot trees with categorical splits


david-cortes:
ref https://github.com/dmlc/xgboost/issues/9810

Currently, attempting to plot trees that have categorical splits in R will result in an error:
```r
library(xgboost)
set.seed(123)
y <- rnorm(100)
x <- sample(3, size=100*3, replace=TRUE) |> matrix(nrow=100)
x <- x - 1
dm <- xgb.DMatrix(data=x, label=y)
setinfo(dm, "feature_type", c("c", "c", "c"))
model <- xgb.train(
    data=dm,
    params=list(
        tree_method="hist",
        max_depth=3
    ),
    nrounds=2
)
xgb.plot.tree(model=model)
````
```
Error in do.call(rbind, matches)[, c(2, 3, 5, 6, 7, 8, 10), drop = FALSE] :
subscript out of bounds
```

This is due to the regexes used to parse the dumps not having been updated for the format used in categorical splits:
https://github.com/dmlc/xgboost/blob/a197899161fa70e681101de4232745fdfe737804/R-package/R/xgb.model.dt.tree.R#L123

david-cortes:
@mayer79 Perhaps you would like to work on this issue?

trivialfis:
I hope we can remove the regex if possible. XGB can output graphviz dump. I can help add other formats if necessary.

david-cortes:
> I hope we can remove the regex if possible. XGB can output graphviz dump. I can help add other formats if necessary.

@trivialfis would be very helpful to add a format "table" which would output the same as python function `trees_to_dataframe`. Then we can get rid of the regexes in both interfaces and avoid needing to update when vector leaves are implemented.

trivialfis:
Yes, I did a proof of concept before, but didn't submit a PR because at the time I was wondering how to export the data to arrow. I can make another attempt.

david-cortes:
> Yes, I did a proof of concept before, but didn't submit a PR because at the time I was wondering how to export the data to arrow. I can make another attempt.

I don't think arrow is necessary here - these tables are going to be rather small in most cases, so perhaps a plain JSON with one entry per column in the table would do.

trivialfis:
> I don't think arrow is necessary here

[...]

PR #11077, opened 2024-12-07
[R] Throw explicit error when using categorical features in functions which don't support them

david-cortes:
ref https://github.com/dmlc/xgboost/issues/9925

This PR adds an explicit error message when trying to call `xgb.model.dt.tree` and functions that depend on it (currently `xgb.plot.multi.trees`) on a model that has categorical features, as it fails to extract the splits from those.

It doesn't actually fix the issue - i.e. this PR doesn't fix the functions to make them work with categorical data - but instead just makes the error message understandable.

Should ideally be reverted once there is a C-level function to produce tables from trees with proper support for categorical splits.


trivialfis (approved):
Will merge once the CI is green.

david-cortes:
On second thought, I'm thinking it better to remove the `text` argument for trees-to-table:
* Once a C-level function to get the table is added, it'd work from a booster object only instead of text dumps, so better to have the breaking change now before a CRAN release.
* Having this `text` argument doesn't allow throwing an informative error when the model has categorical features.

trivialfis:
> Once a C-level function to get the table is added

It might take a while, though. I'm working on storing the `factor` in the booster, as there are complaints about this being a foot gun.

trivialfis:
Could you please help resolve the conflicts?

david-cortes:
> Could you please help resolve the conflicts?

Solved.

trivialfis (approved):

PR #11001, opened 2024-11-19
Next-generation CI/CD pipelines with RunsOn

hcho3:
Suggestion to reviewers: Go to https://github.com/hcho3/xgboost/tree/xgboost-ci-ng to easily browse the workflow definitions and scripts. Also pay attention to the following:
1. Overall organization of workflows
2. How containers are defined and constructed

See https://xgboost--11001.org.readthedocs.build/en/11001/contrib/ci.html for the overview.

-------

In #8142, we migrated our CI/CD pipelines from Jenkins to BuildKite. BuildKite has served us well in the past 2 years.
BuildKite was superior to Jenkins in the following ways:
* Provision of cloud components are now documented in scripts that are stored as part of the XGBoost git repository.
* We are no longer subject to frequent failures and security issues from Jenkins plug-ins.

However, over time, the BuildKite setup has entailed the following disadvantages:
1. We ended up with lots of infra scripts and glue code that provision cloud components. (See `tests/buildkite/infrastructure`.) The infra scripts require specialized knowledge of the cloud, and currently I am the only person who understands these scripts.
4. The provision of worker images was not fully automated. The setup steps were documented in bootstrap scripts, but someone had to manually run them. Also, the BuildKite's CloudFormation template required that we specify exact IDs for machine images, meaning that we had to re-provision the entire cloud stack every time we re-generate the machine images. In short: **It was cumbersome to generate new machine images**, so over time, we tended to not update machine images, which led to bit rots.
5. Developers who are familiar with GitHub Actions face substantial learning curve when configuring BuildKite, as BuildKite is a more niche application.
6. Over time, we added more and more files in `tests/buildkite` and `tests/ci_build`, and it was unclear as to what kinds of files belonged to each directory. (#9896)

I propose a new solution for hosting our CI/CD pipelines: [RunsOn](https://runs-on.com/). RunsOn is a web app that lets host GitHub Action runners on Amazon EC2. By migrating to RunsOn, we address all the issues described above.
1. We got rid of (almost) all specialized infra scripts. The only remaining infra script is the Packer build definition in `ops/packer`, but at least this bit is easy to understand for newcomers. (It basically provisions an EC2 instance, runs a few commands, and saves the machine image for later use.)
[...]

PR #11079, opened 2024-12-08
Improve design and address comments in the new CI

hcho3:
Follow up to #11001 

Apply feedback from #11046. See #11046 to learn about the rationale for changes.


hcho3:
@trivialfis Most of the changes here are stylistic and do not affect the development process too much. However, this pull request contains one major change that significantly changes the development flow.

Currently, we build Docker containers on the fly as part of the CI run for the pull request. But after this PR, we will start to have a separate CI pipeline to build and push CI containers in https://github.com/dmlc/xgboost-devops.

## Why the change?
**To make the code more legible and easier to reason about**. The current script `docker_build.sh` is hard to understand because it does **three things** at once:
1) build a container
2) fetch a container from a private registry (for caching); and
3) push the new container back to the registry.

Often a workflow would call `docker_build.sh` just to fetch a pre-built container from the registry. This makes it confusing to newcomers, as noted by @jameslamb in https://github.com/dmlc/xgboost/pull/11046#discussion_r1871795114 and https://github.com/dmlc/xgboost/pull/11046#discussion_r1871816405.

This PR separates the concerns by having each script do one thing only.
* `docker_build.sh` (xgboost-devops repo) builds a container and pushes it to the registry.
* `docker_run.py` (xgboost repo) fetches the container from the registry.

**To ensure that the container stays up to date**. In the current setup, we use the Docker registry as a caching layer for containers. This setup suffers from two problems.

1. **False cache hit**: New versions of packages are released but the containers are not refreshed in time because the Dockerfile did not get a (textual) change. For example, the `xgb-ci.gpu` container installs `scikit-learn` (no constraints specified). Even though a new version of `scikit-learn` (1.6.0) was released today, the container was not refreshed because the Dockerfile for the container, `Dockerfile.gpu`, was unchanged. This behavior is a side effect of how Docker caches builds.
2. **Containers get re-built at unexpected time interval**: Whenever the base image gets updated, the whole Docker container gets re-built. The re-build tends to happen at an unfortunate timing, causing us to wait longer for the CI execution for time-critical PRs.

In the new setting, `xgboost-devops` repo hosts a dedicated CI pipeline to build CI containers at regular scheduled time. (It is triggered by `push`, `pull_request`, as well as `schedule`.) So the first issue is addressed. The second issue is also addressed, as PRs on the XGBoost repo will spend a predictable amount of time pulling the containers from the registry.

I've written a guide to show how CI containers are updated: https://xgboost--11079.org.readthedocs.build/en/11079/contrib/ci.html#making-changes-to-ci-containers. Can you take a look at it and see if it makes sense?

ops/pipeline/build-test-jvm-packages.sh
```diff
@@ -21,8 +23,10 @@ do
   fi
 done
 
+CONTAINER_TAG=${DOCKER_REGISTRY_URL}/xgb-ci.jvm:main
```
jameslamb, line 26: I would probably have called this `IMAGE_URI` instead of `CONTAINER_TAG`, since this is the fully-qualified identifier for an image. The "tag" in this case is just `:main`.

In general, I recommend following this type of naming for the variables involved here:

* `IMAGE_REGISTRY` = the place to pull the images from, like
  - *(e.g. `docker.io`, `quay.io`, `ghcr.io`, an ECR URL)*
* `IMAGE_REPO` = the name of the image, called a "repository" on some registries
  - *(e.g. `xgb-ci.jvm`, `dask/dask-notebook`)*
* `IMAGE_TAG` = ID that separates images from the same registry and repo
  - *(e.g. `2024.10`, `main`, `stable`)*
* `IMAGE_URI` = fully-qualified identifier of the form `${IMAGE_REGISTRY}/${IMAGE_REPO}:${IMAGE_TAG}`
  - *(e.g. `quay.io/pypa/manylinux2014_aarch64:latest`)*

I know this might seem minor and pedantic, but being precise about these terms (and about the difference between "container" and "image") is helpful for others reading these scripts and config files.

python-package/pyproject.toml
```diff
@@ -85,3 +85,9 @@ class-attribute-naming-style = "snake_case"
 
 # Allow single-letter variables
 variable-rgx = "[a-zA-Z_][a-z0-9_]{0,30}$"
+
+[tool.pydistcheck]
+inspect = true
+ignore = ["compiled-objects-have-debug-symbols"]
```
jameslamb, line 91: Thanks for trying this out! If other checks in `pydistcheck` are not important to you or cause issues, you can add a section like this to opt in to only the ones you care about:

```toml
select = [
  "distro-too-large-compressed"
]
```

I modeled this off of the way `ignore` and `select` work in `ruff` ... if you populate `select`, it turns the behavior into "only run the checks listed in `select`, and skip all others".

That's what I did in PRs for https://github.com/rapidsai/build-planning/issues/110, to keep it tightly focused on that one purpose of checking file sizes.

If you want to keep all the other checks (https://pydistcheck.readthedocs.io/en/latest/check-reference.html) then great, I hope they're helpful, but just mentioning this so you know it's possible to limit the scope of what this tool is checking.

jameslamb (commented):
Responding to the `@` from https://github.com/dmlc/xgboost/pull/11046#issuecomment-2528338561 ... sorry I couldn't get to it sooner!

I know this is already merged, but left a few small suggestions you might want to. consider for followup. All very minor... I don't have any other significant structural changes to request.

This looks awesome! Thanks so much for giving me the opportunity to review and for talking through all the comments I left on #11046 with me.

PR #11089, opened 2024-12-11
Extract array interface handlers.

trivialfis:
- Extract helper functions into a different module.
- Define typed dictionary for type check.


wphicks (approved):
Looks great!

Issue #11088, opened 2024-12-11
Auto encoding for categorical data during inference.


trivialfis:
We are working on automatic re-encoding for categorical features during inference. This teaches the booster to handle data encoded differently than the training dataset and eliminates the need for a scikit-learn pipeline for data encoding when using DataFrame inputs.

- [ ] Python native
- [ ] Dask
- [ ] PySpark
- [ ] R
- [ ] Scala/Spark

Related:
- https://github.com/dmlc/xgboost/issues/9676

trivialfis:
@david-cortes May I ask how the `tmp` is kept alive during the construction of QDM?

https://github.com/dmlc/xgboost/blob/f4f3bd4e61a2375f9ea8cbb2dd1d3af7d5b09cf4/R-package/R/xgb.DMatrix.R#L181

The `setData` from the proxy DMatrix only holds a reference (a C pointer) to the input data, if the data is released early then it's a use-after-free error. But testing the `qdm.from_iterator` seems fine.

david-cortes:
@trivialfis In that function, the DMatrix is set in the line right before `tmp` gets deleted. Or are you saying that the data still needs to be kept after the DMatrix has been set? If so, at which point is it safe to deallocate the data?

Regarding the feature: since the idea is to have this feature in different interfaces, how would it work behind the scenes?

Would be ideal if the categorical encodings could get saved in the booster and be used in plots/trees-to-tables/jsons/etc. (https://github.com/dmlc/xgboost/issues/9927). Better yet if it's a standardized C-level attribute so that the encodings could survive transfers from one interface to another.

I see some potential difficulties though:
* Pandas allows arbitrary python `object` for categorical encodings (not json-friendly).
* Different data formats (pandas, arrow, polars, etc.) might have different limitations for what kind of values they allow as categories (e.g. strings-only vs. also integers).

trivialfis:
>  Or are you saying that the data still needs to be kept after the DMatrix has been set

It needs to be kept until the next `next` call or `reset` call of the iterator. I got a bit confused since the data is related right after the `setData` call, but no access error. So, in theory, it should be kept after set, and until the next iteration.

> Regarding the feature: since the idea is to have this feature in different interfaces, how would it work behind the scenes?

We will store the levels in the booster as you suggested. Things will be handled in C++, we might allow users to optionally disable the encoder for performance reasons (searching through levels is not cheap in the context of inference, especially with strings).

[...]

PR #11092, opened 2024-12-11
[R] Ensure `ProxyDMatrix` creation keeps data until next iteration

david-cortes:
ref https://github.com/dmlc/xgboost/issues/11088

This PR modifies the logic of the ProxyDMatrix loop to ensure that whatever data it gets fed will be kept alive until the next iteration.

### DIFF
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -682,12 +682,12 @@ XGB_DLL SEXP XGProxyDMatrixCreate_R() {
 
 XGB_DLL SEXP XGProxyDMatrixSetDataDense_R(SEXP handle, SEXP R_mat) {
   R_API_BEGIN();
-  R_SetExternalPtrProtected(handle, R_mat);
   DMatrixHandle proxy_dmat = R_ExternalPtrAddr(handle);
   int res_code;
   {
     std::string array_str = MakeArrayInterfaceFromRMat(R_mat);
     res_code = XGProxyDMatrixSetDataDense(proxy_dmat, array_str.c_str());
+    R_SetExternalPtrProtected(handle, R_mat);
   }
   CHECK_CALL(res_code);
   R_API_END();
@@ -696,7 +696,6 @@ XGB_DLL SEXP XGProxyDMatrixSetDataDense_R(SEXP handle, SEXP R_mat) {
 
 XGB_DLL SEXP XGProxyDMatrixSetDataCSR_R(SEXP handle, SEXP lst) {
   R_API_BEGIN();
-  R_SetExternalPtrProtected(handle, lst);
   DMatrixHandle proxy_dmat = R_ExternalPtrAddr(handle);
   int res_code;
   {
@@ -709,6 +708,7 @@ XGB_DLL SEXP XGProxyDMatrixSetDataCSR_R(SEXP handle, SEXP lst) {
                                         array_str_indices.c_str(),
                                         array_str_data.c_str(),
                                         ncol);
+    R_SetExternalPtrProtected(handle, lst);
   }
   CHECK_CALL(res_code);
   R_API_END();
@@ -717,12 +717,12 @@ XGB_DLL SEXP XGProxyDMatrixSetDataCSR_R(SEXP handle, SEXP lst) {
 
 XGB_DLL SEXP XGProxyDMatrixSetDataColumnar_R(SEXP handle, SEXP lst) {
   R_API_BEGIN();
-  R_SetExternalPtrProtected(handle, lst);
   DMatrixHandle proxy_dmat = R_ExternalPtrAddr(handle);
   int res_code;
   {
     std::string sinterface = MakeArrayInterfaceFromRDataFrame(lst);
     res_code = XGProxyDMatrixSetDataColumnar(proxy_dmat, sinterface.c_str());
+    R_SetExternalPtrProtected(handle, lst);
   }
   CHECK_CALL(res_code);
   R_API_END();
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -682,6 +682,7 @@ XGB_DLL SEXP XGProxyDMatrixCreate_R() {
 
 XGB_DLL SEXP XGProxyDMatrixSetDataDense_R(SEXP handle, SEXP R_mat) {
   R_API_BEGIN();
+  R_SetExternalPtrProtected(handle, R_mat);
   DMatrixHandle proxy_dmat = R_ExternalPtrAddr(handle);
   int res_code;
   {
@@ -695,6 +696,7 @@ XGB_DLL SEXP XGProxyDMatrixSetDataDense_R(SEXP handle, SEXP R_mat) {
 
 XGB_DLL SEXP XGProxyDMatrixSetDataCSR_R(SEXP handle, SEXP lst) {
   R_API_BEGIN();
+  R_SetExternalPtrProtected(handle, lst);
   DMatrixHandle proxy_dmat = R_ExternalPtrAddr(handle);
   int res_code;
   {
@@ -715,6 +717,7 @@ XGB_DLL SEXP XGProxyDMatrixSetDataCSR_R(SEXP handle, SEXP lst) {
 
 XGB_DLL SEXP XGProxyDMatrixSetDataColumnar_R(SEXP handle, SEXP lst) {
   R_API_BEGIN();
+  R_SetExternalPtrProtected(handle, lst);
   DMatrixHandle proxy_dmat = R_ExternalPtrAddr(handle);
   int res_code;
   {
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -739,7 +739,7 @@ struct _RDataIterator {
   SEXP proxy_dmat;
 
   _RDataIterator(
-    SEXP f_next, SEXP f_reset, SEXP calling_env, SEXP continuation_token, SEP proxy_dmat) :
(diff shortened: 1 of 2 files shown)


R-package/src/xgboost_R.cc
```diff
@@ -682,6 +682,7 @@ XGB_DLL SEXP XGProxyDMatrixCreate_R() {
 
 XGB_DLL SEXP XGProxyDMatrixSetDataDense_R(SEXP handle, SEXP R_mat) {
   R_API_BEGIN();
+  R_SetExternalPtrProtected(handle, R_mat);
```
trivialfis, line 685: Hmm, so when is the data released?

david-cortes: My understanding is that the external pointer object has only one slot for a protected `SEXP` object. So a call to this function would make the previous object in that slot become unprotected:
https://github.com/wch/r-source/blob/872770ededfc26429adca09fab0803e1b78b6eac/src/main/memory.c#L3855
https://github.com/wch/r-source/blob/872770ededfc26429adca09fab0803e1b78b6eac/src/main/memory.c#L1292

But I'm not 100% sure about it.

david-cortes:
Looks like `xgb.ExtMemDMatrix` and `xgb.QuantileDMatrix.from_iterator` will  also need similar protection mechanisms.

R-package/src/xgboost_R.cc
trivialfis: Going through C macros is not my strong suit ...

Do you think it's difficult to store the data in the `ProxyDMatrixWrapper` you wrote in `xgboost_R.cc`? Currently, it's not used in the R code, but we can make the R proxy dmatrix creator return the wrapper instead of the real proxy dmatrix handler.

trivialfis: In Python, I store the temporary data in the iterator, but anywhere is fine.

david-cortes: Going through the code again, it looks like both `xgb.ExtMemDMatrix` and `xgb.QuantileDMatrix.from_iterator` will end up calling one of these 3 functions in the end, so please ignore the latter comment.

> Do you think it's difficult to store the data in the ProxyDMatrixWrapper you wrote in xgboost_R.cc?

I think it could be quite tricky, because the public R C API doesn't allow protecting/unprotecting individual objects (unless by the logic of putting an unprotected object inside a protected container like list/externalptr) - it only offers `UNPROTECT(<last N>)`.

It'd have to use a list object and be constantly assigning the same element to it in order to achieve the same effect, which wouldn't have any advantage compared to the mechanism in this PR of assigning to the "protected" slot of the externalptr, **assuming that it works the way we expect it to**.

trivialfis: > assuming that it works the way we expect it to.

That's internal to R right? Even if it works today, it can break at the next release and we don't have a way to test that.

david-cortes: > > assuming that it works the way we expect it to.
> 
> That's internal to R right? Even if it works today, it can break at the next release and we don't have a way to test that.

No, that's part of the public C API of R, which has pretty strong guarantees. Internal API calls in package code would disqualify packages from being hosted at CRAN.

And either way, it looks like it should end up using the macro from the earlier link that decreases the reference count, unless you manually happen to define internal-use variables before including the R header:
https://github.com/wch/r-source/blob/872770ededfc26429adca09fab0803e1b78b6eac/src/include/Rinternals.h#L191

trivialfis: > No, that's part of the public C API of R

hmm, I meant the "external pointer object has only one slot for a protected SEXP object.  So a call to this function would make the previous object in that slot become unprotected" behavior.

david-cortes:
@trivialfis Just to make sure: in this PR, the data is being held right up until the next call to `XGProxyDMatrixSetData*`.

But it gets unprotected _before_ the next call with the next batch of data.

Is that the correct behavior or should it be the other way around? i.e. old batch kept alive **until current batch** setdata finished.

trivialfis:
> Is that the correct behavior or should it be the other way around? i.e. old batch kept alive until current batch setdata finished.

The previous batch can be freed once the `next` or the `reset` is called. The events happen in the following sequence:

1. iter->next (in R)
2. generate temp data
3. set_data called inside the `next` , exit the next call
4. xgboost consumes the data in C
5. iter->next (in R)
6. generate temp data
7. set_data called inside the `next`, exit `next` call.
8. xgboost consumes the data in C
...

The `free` for data generated by 2 can happen after 5 and before 8. If called before 6, then we can save some memory as there's only one batch of temporary data can be in the memory. The saved memory doesn't matter much for most of the cases.

david-cortes:
> > Is that the correct behavior or should it be the other way around? i.e. old batch kept alive until current batch setdata finished.
> 
> The previous batch can be freed once the `next` or the `reset` is called. The events happen in the following order:
> 
> 1. iter->next (in R)
> 2. generate temp data
> 3. set_data called inside the `next` , exit the next call
> 4. xgboost consumes the data in C
> 5. iter->next (in R)
> 6. generate temp data
> 7. set_data called inside the `next`, exit `next` call.
> 8. xgboost consumes the data in C
>    ...
> 
> The `free` for data generated by 2 can happen between 5 and 8. If freed before 6, then we can save some memory as there's only one batch of temporary data can be in the memory. The saved memory doesn't matter much in most cases.

But when is (4) finalized? In this case, the only interactions with the DMatrix object are through `setdata`. Or am I missing some additional call?

trivialfis:
> But when is (4) finalized? 

When 5 happens.

david-cortes:
> > But when is (4) finalized?
> 
> When 5 happens.

Meaning: inside the loop in `*FromCallback` before calling `XGDMatrixCallbackNext(DataIterHandle)` ?

trivialfis:
>  the only interactions with the DMatrix object are through setdata. Or am I missing some additional call?

No missing call.

In the previous comment, both the `setdata` and generation of temp data happen INSIDE the `next` call:
```
function iter.next() {
    # free here
    temp = process_df()
    set_data(temp)
    # free here works as well, but we can have two temp data in the memory.
}

R-package/src/xgboost_R.cc
trivialfis: got it, thank you for sharing.

david-cortes:
I'm still not getting when exactly would the DMatrix be finished with the next batch, but I've added some calls to unprotect the previous batch right before the call to the R-side 'next'. It should in theory allow freeing the earlier batch before the next one is allocated in memory when using external memory.

Please take a look to see if that's the correct place.

trivialfis:
> I'm still not getting when exactly would the DMatrix be finished with the next batch

may I ask where is the confusion? Apologies for my poor explanation. If a new batch is requested by XGBoost, then the old batch is finished, and XGBoost requests a new batch by calling the next function.

david-cortes:
> > I'm still not getting when exactly would the DMatrix be finished with the next batch
> 
> may I ask where is the confusion? Apologies for my poor explanation. If a new batch is requested by XGBoost, then the old batch is finished, and XGBoost requests a new batch by calling the next function. Not entirely sure where it got complicated.

Got it now. So then the current logic as of the last commit should work in such a way that it doesn't require two batches to co-exist in memory.

But wouldn't it be more logical to make it so that the data is consumed during the call to `setdata` and not referenced after that call, considering that there aren't any further user functions running between the two?

trivialfis:
The previous batch can be freed inside anywhere of this function:

https://github.com/dmlc/xgboost/blob/376009ca13d250473b0dee2dec3d1148eac6963c/R-package/R/xgb.DMatrix.R#L557

preferably before:

https://github.com/dmlc/xgboost/blob/376009ca13d250473b0dee2dec3d1148eac6963c/R-package/R/xgb.DMatrix.R#L569

since we will have two temporary batches in memory if we free after this line. But it's fine.

trivialfis:
> But wouldn't it be more logical to make it so that the data is consumed during the call to setdata 

It couldn't, the proxy dmatrix is just a proxy to the user input as the name suggests, and doesn't do anything other than hold the reference to the data. Consider it a "data view" that XGBoost understands, instead of a container of the data. I should probably call it `DataReference` or something like that instead of `ProxyDMatrix`.

david-cortes:
> The previous batch can be freed inside anywhere of this function:
> 
> https://github.com/dmlc/xgboost/blob/376009ca13d250473b0dee2dec3d1148eac6963c/R-package/R/xgb.DMatrix.R#L557
> 
> preferably before:
> 
> https://github.com/dmlc/xgboost/blob/376009ca13d250473b0dee2dec3d1148eac6963c/R-package/R/xgb.DMatrix.R#L569
> 
> since we will have two temporary batches (old and new `tmp`) in memory if we free after this line. But it's fine.

Then the current placement should be fine, since this function:
https://github.com/dmlc/xgboost/blob/376009ca13d250473b0dee2dec3d1148eac6963c/R-package/R/xgb.DMatrix.R#L556

Is only called from the C++ side here:
https://github.com/dmlc/xgboost/blob/376009ca13d250473b0dee2dec3d1148eac6963c/R-package/src/xgboost_R.cc#L747

.. through wrappers like these:
https://github.com/dmlc/xgboost/blob/376009ca13d250473b0dee2dec3d1148eac6963c/R-package/R/xgb.DMatrix.R#L365

trivialfis (approved):
Thank you for the fix! Looks great.

david-cortes:
Looks like this didn't work out correctly. Now it ends up with a null pointer derefence somewhere.

trivialfis:
Maybe we lower the data transformation into C and store the temporary data in the proxy wrapper.

PR #11095, opened 2024-12-12
[R] Change behavior around renamed / removed arguments

david-cortes:

In a previous PR (https://github.com/dmlc/xgboost/pull/11074), many function arguments were removed.

Currently, the logic is that passing parameters through `...` results in errors, which is the behavior we want to have for functions like `xgb.train`. But to make a potential CRAN release easier, perhaps a first release of the new interface could be more lenient.

This PR changes the default behavior to instead accept arguments that were present in the latest CRAN release of xgboost, but issuing a warning. It adds an option `xgboost.strict_mode` that would turn these warnings into errors as a workaround, without needing to keep them in signatures.

Ideally we could remove this whole system soon after we land a new release on CRAN.

_Note: it assumes that this other PR https://github.com/dmlc/xgboost/pull/11077 should have been merged_


xximwon (approved):

trivialfis:
I think we will have a new cran package named something like `xgboost3` instead of reusing the existing package.

> Note: it assumes that this other PR https://github.com/dmlc/xgboost/pull/11077 should have been merged

Do you want to merge that PR first?

david-cortes:
> Would you like to merge that PR first?

Yes, would be ideal to merge that PR.

> We will have a new cran package named something like xgboost3 instead of reusing the existing package. So, a clean break is preferable.

I nervetheless think a warning with option to turn into error could be a good trade off for the very first release for reverse dependencies.

david-cortes:
In particular, if one a user ends up with parallel installs of `xgboost` and `xgboost3`, there might be name clashes on dependencies that use xgboost functions as they adapt; and if released under the same name `xgboost`, it'd give a much longer time period for dependencies to adapt.

xximwon (approved):

xximwon (approved):

xximwon (approved):

xximwon (approved):

trivialfis (approved):

PR #11097, opened 2024-12-13
Various small cleanups.

trivialfis:
- Add deduction guide for Span.
- Merge data split mode setter into the `MetaInfo` column sync method.
- Check CUDA warning for cross space function calls.


trivialfis:
cc @rongou

rongou (approved):

PR #11098, opened 2024-12-14
Initial implementation of the ordinal recoder.

trivialfis:
There is no integration yet, just the recoder. Retrieving and storing the data in xgboost is more complicated than the recoder itself, will upstream it in future PRs.

The recoder still uses some utilities in XGBoost like the `Span` class and an iterator. If we want to extract it to a different project, we can find a different implementation of these utilities. Other things like error handling and memory allocation can be customized through the policy class.

Tests require a container class in XGBoost. We can merge the container into the encoder module if needed. At the moment, the encoder is view-only and doesn't own any memory. After the Python and R integration is finished, more extensive and sophisticated tests will be done.

ref https://github.com/dmlc/xgboost/issues/11088


trivialfis:
cc @rongou .

rongou (approved):

trivialfis:
Merging, we will revisit when other projects want to reuse the code.

PR #11104, opened 2024-12-15
[R] Move gc data protection to R side

david-cortes:
ref https://github.com/dmlc/xgboost/pull/11092#issuecomment-2543526600

This PR moves the logic for protection of R data from the garbage collector towards the R side.

I'm not sure whether data holding _fields_ of the DMatrix also needs to be kept protected after setting it in the proxy dmatrix, but this PR does so just in case.

### DIFF
--- a/R-package/R/xgb.DMatrix.R
+++ b/R-package/R/xgb.DMatrix.R
@@ -354,8 +354,7 @@ xgb.QuantileDMatrix <- function(
   data_iterator <- .single.data.iterator(iterator_env)
 
   env_keep_alive <- new.env()
-  env_keep_alive$keepalive1 <- NULL
-  env_keep_alive$keepalive2 <- NULL
+  env_keep_alive$keepalive <- NULL
 
   # Note: the ProxyDMatrix has its finalizer assigned in the R externalptr
   # object, but that finalizer will only be called once the object is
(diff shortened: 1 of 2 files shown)


trivialfis:
Let me try to create a prototype for the https://github.com/dmlc/xgboost/issues/11088 and see how this should work. I need to store the categories somewhere.

trivialfis:
Should we revert https://github.com/dmlc/xgboost/pull/11092 before applying this PR? The previous one seems redundant now.

david-cortes:
> Should we revert #11092 before applying this PR? The previous one seems redundant now.

It is undoing the changes from it already.

But it leaves me wondering: what's the difference between `xgb.ExtMemDMatrix` and `xgb.QuantileDMatrix.from_iterator`? Both of them seem to apply exclusively to histogram-based models, and both of them seem to access the data the same amount of times.

Is `xgb.ExtMemDMatrix` perhaps meant to leave the iterator alive throughout the call to `xgb.train`?

trivialfis:
The `xgb.ExtMemDMatrix` stores a cache on disk, whereas the `QDM.from_iterator` concatenates everything in memory and saves memory by using the compressed histogram bin index instead of raw input.

In addition, the former applies to both `approx` and `hist`, while the latter is `hist` only.

https://github.com/dmlc/xgboost/blob/master/doc/tutorials/external_memory.rst#compared-to-the-quantiledmatrix

trivialfis:
> I'm not sure whether data holding fields of the DMatrix

If you meant things like labels and weights, no, they don't need any protection.

david-cortes:
> > I'm not sure whether data holding fields of the DMatrix
> 
> If you meant things like labels and weights, no, they don't need any protection.

Thanks, modified the PR to **not** keep those protected after setting them.

trivialfis:
Will take a deeper look tomorrow.

trivialfis (approved):

PR #11112, opened 2024-12-17
Optimize memory reallocations

razdoburdin:
This small PR optimizes memory allocations during `CommonRowPartitioner `creation. In the current realization the memory for `RowSetCollection `is allocated in each iteration. This PR allows to reuse the already allocated memory if possible. 

The update is the most beneficial for datasets with high-number of rows.
Here are the perf gains for execution on single- and 2-socket machines, each socket has 56 cores CPU. HT is not used. 
image


razdoburdin:
The CI failure looks not related to PRs content. 
image

trivialfis (approved):

PR #11113, opened 2024-12-17
Add important details about external memory to docs

david-cortes:
Not 100% sure that I'm getting all the details right, so @trivialfis please take a careful look.

### DIFF
--- a/doc/tutorials/external_memory.rst
+++ b/doc/tutorials/external_memory.rst
@@ -26,9 +26,9 @@ The external memory support has undergone multiple development iterations. Like 
 data batch-by-batch using a custom iterator supplied by the user. However, unlike the
 :py:class:`~xgboost.QuantileDMatrix`, external memory does not concatenate the batches
 (unless specified by the ``extmem_single_page``) . Instead, it caches all batches in the
-external memory and fetch them on-demand. Go to the end of the document to see a
-comparison between :py:class:`~xgboost.QuantileDMatrix` and the external memory version of
-:py:class:`~xgboost.ExtMemQuantileDMatrix`.
+external memory to disk (in a compressed format) and fetches them on-demand. Go to the
+end of the document to see a comparison between :py:class:`~xgboost.QuantileDMatrix` and
+the external memory version of :py:class:`~xgboost.ExtMemQuantileDMatrix`.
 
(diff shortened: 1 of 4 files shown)


doc/tutorials/external_memory.rst
```diff
@@ -26,9 +26,9 @@ The external memory support has undergone multiple development iterations. Like
 data batch-by-batch using a custom iterator supplied by the user. However, unlike the
 :py:class:`~xgboost.QuantileDMatrix`, external memory does not concatenate the batches
 (unless specified by the ``extmem_single_page``) . Instead, it caches all batches in the
-external memory and fetch them on-demand. Go to the end of the document to see a
-comparison between :py:class:`~xgboost.QuantileDMatrix` and the external memory version of
-:py:class:`~xgboost.ExtMemQuantileDMatrix`.
+external memory to disk (in a compressed format) and fetches them on-demand. Go to the
```
trivialfis, line 29: We use  `external memory` to refer to disk. For the CPU, the "main memory" is the memory we usually talk about, and external memory can refer to anything else from disk to network storage. For XGBoost, only disk is supported. As a result, in XGBoost "external memory" and "disk" are the same thing from the perspective of a CPU-based algorithm.

For GPU, the main memory is the device memory, whereas CPU memory and disk are "external" to GPU.

python-package/xgboost/core.py
```diff
@@ -446,6 +446,9 @@ class DataIter(ABC):  # pylint: disable=too-many-instance-attributes
     cache_prefix :
         Prefix to the cache files, only used in external memory.
 
+        Note that using this class for non-quantile external memory **will cache data
+        on disk** under the path passed here.
```
trivialfis, line 450:
```suggestion
        Note that using this class for external memory **will cache data
        on disk** under the path passed here.
```

doc/tutorials/external_memory.rst
trivialfis: Maybe a brief intro with a few sentences. But for this PR, could you please help revise or revert the change?

david-cortes: Undid the changes to the .rst file.

trivialfis (approved):

PR #11115, opened 2024-12-18
[jvm-packages] Support external memory

wbo4958:


trivialfis:
Added some changes. It's not quite optimized yet, the host cache might be slow due to growing the stream. I will be on holiday until next year, so pushing as much as possible tonight.

Do we plan to do something similar for pyspark?

PR #11118, opened 2024-12-18
Remove inconsistency in return types between `predict` and `inplace_predict`

galipremsagar:
I see that `inplace_predict` returns a `cupy` array but `predict` returns a numpy array always. Not sure if this is intentional. Can one of the maintainers comment on this?

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -1294,10 +1294,6 @@ class XGBModel(XGBModelBase):
                         base_margin=base_margin,
                         validate_features=validate_features,
                     )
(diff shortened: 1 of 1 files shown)


trivialfis:
If you are using the sklearn interface, it always returns a numpy array until https://github.com/dmlc/xgboost/pull/8950 can be resolved. As for the lower level `predict/inplace_predict`, yes, it's intentional. The `inplace_predict` was introduced much later, and we wanted to align it with user inputs (DF v.s. Array). There's no alignment for the `predict` method as the input is a `DMatrix`.

galipremsagar:
Thanks @trivialfis ! Closing this PR.

PR #11126, opened 2024-12-20
Warning for polars lazy frame.

trivialfis:
We might eventually reuse the booster `device` config for the `collect` call. But for now, a warning should be enough. Otherwise, we need to check whether the polars is built with GPU support in addition to retrieving the JSON configuration from XGBoost for every inference call.


trivialfis:
cc @rongou .

rongou (approved):

PR #11127, opened 2024-12-23
[jvm-packages] Supports external memory

wbo4958:
This PR introduces a new parameter "useExternalMemory" to indicate whether to use external memory when build QuantileDMatrix. If it's set to true, then the cached file will be cached into the directory specified by `spark.local.dir`, if `spark.local.dir` is not set, "/tmp/ will be used for the caching.

### DIFF
--- a/jvm-packages/xgboost4j/src/native/xgboost4j-gpu.cu
+++ b/jvm-packages/xgboost4j/src/native/xgboost4j-gpu.cu
@@ -6,6 +6,7 @@
 
 #include "../../../../src/common/cuda_pinned_allocator.h"
 #include "../../../../src/common/device_vector.cuh"  // for device_vector
+#include "../../../../src/common/json_utils.h"
 #include "../../../../src/data/array_interface.h"
 #include "jvm_utils.h"  // for CheckJvmCall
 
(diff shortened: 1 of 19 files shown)


jvm-packages/xgboost4j/src/native/xgboost4j-gpu.cu
```diff
@@ -391,6 +392,116 @@ class DataIteratorProxy {
   };
 };
 
+// An iterator proxy for external memory.
+class ExternalMemoryIteratorProxy {
```
trivialfis, line 396: This is the same as the original one with only `label` is supported?

jvm-packages/xgboost4j/src/native/xgboost4j-gpu.cu
```diff
@@ -391,6 +392,116 @@ class DataIteratorProxy {
...
+  jobject jiter_;
+  JNIEnv *jenv_;
+  DMatrixHandle proxy_;
+  int jni_status_;
+  jobject last_batch_ {nullptr};
+
+ public:
+  explicit ExternalMemoryIteratorProxy(jobject jiter): jiter_(jiter) {
+    XGProxyDMatrixCreate(&proxy_);
```
trivialfis, line 405: These function calls need to be checked.

PR #11131, opened 2024-12-28
Update C API tutorial

ayoub317:
Set C++17 as the minimum required version for the tutorial

The current code relies on features introduced in C++17, such as nested namespaces and std::is_same_v.
Without C++17, the compilation results in warnings like:
"nested namespace definition is a C++17 extension; define each namespace separately", and errors like:
"no template named 'is_same_v' in namespace 'std'".
To fix this, I've set C++17 as the minimum required version to ensure compatibility.

Additionally, I made a small fix to the matrix initialization in the tutorial.


trivialfis:
Hi, thank you for the PR. I think the tutorial itself is C-only and doesn't need c++ version configuration.

ayoub317:
Hello, my bad ! Thank you for pointing that out.

trivialfis (approved):

PR #11133, opened 2024-12-29
[python-package][PySpark] Expose Training and Validation Metrics

### LINKED ISSUE #11132: [python-package][PySpark] Expose Training and Validation Metrics
Unlike the JVM binding, where, after training the XGBoost Spark model, we can retrieve a summary of the training and evaluation sets passed, this functionality is not currently available in the PySpark Python XGBoost binding.

In the JVM package, this is defined through [this](https://github.com/dmlc/xgboost/blob/bec2d3236081dd7b429ddb791ac1e6df202e27f3/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostTrainingSummary.scala) class. We obtain the summary while training the model, which first passes through [this method](https://github.com/dmlc/xgboost/blob/bec2d3236081dd7b429ddb791ac1e6df202e27f3/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java#L157C17-L157C24), and in the end reaches [this](https://github.com/dmlc/xgboost/blob/bec2d3236081dd7b429ddb791ac1e6df202e27f3/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoost.scala#L254) line. After that, when creating the Spark[Classification|Regression|Ranker]Model, we pass through a constructor like [this](https://github.com/dmlc/xgboost/blob/bec2d3236081dd7b429ddb791ac1e6df202e27f3/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostRegressor.scala#L58) one.

I followed a similar approach to expose and retrieve these metrics in the PySpark XGBoost binding, and I will be submitting a review soon. Any feedback is welcome.

ayoub317:
Closes #11132


python-package/xgboost/spark/xgboost_training_summary.py
```diff
@@ -0,0 +1,43 @@
+"""Xgboost training summary integration submodule."""
+
+from dataclasses import dataclass, field
+from typing import Dict, List
+
+
+@dataclass
+class _XGBoostTrainingSummary:
```
wbo4958, line 8: Could be `XGBoostTrainingSummary` ?

python-package/xgboost/spark/xgboost_training_summary.py
```diff
@@ -0,0 +1,43 @@
+"""Xgboost training summary integration submodule."""
```
wbo4958, line 1: how about renaming xgboost_training_summary.py to summary.py?

python-package/xgboost/spark/xgboost_training_summary.py
ayoub317: DONE

python-package/xgboost/spark/xgboost_training_summary.py
ayoub317: DONE

wbo4958:
Hi @ayoub317, Could you be able to add some unit tests for this feature?

ayoub317:
Hi @wbo4958,

Yes, with pleasure ! Could you please provide some pointers on where I should define these tests ?

I had a quick look at the test codebase, and I would assume the following:

The tests for the regressor and classifier could be placed under:
xgboost -> tests -> test_distributed -> test_with_spark -> test_spark_local.py -> in the class [TestPySparkLocal](https://github.com/dmlc/xgboost/blob/master/tests/test_distributed/test_with_spark/test_spark_local.py#L461)

The test for the ranker could be placed under:
xgboost -> tests -> test_distributed -> test_with_spark -> test_spark_local.py -> in the class [TestPySparkLocalLETOR](https://github.com/dmlc/xgboost/blob/master/tests/test_distributed/test_with_spark/test_spark_local.py#L1781)

Does that sound like a good choice ?

Thanks !

wbo4958:
The path you pasted should be ok for adding new testing.

ayoub317:
Note that it seems the failures on the CI are not related to these commits, and the linting issue encountered with the last test commit could not be avoided due to the current linting setup.

Black will complain about the imports not being formatted as follows:
```
from xgboost.spark import (
    SparkXGBClassifier,
    SparkXGBClassifierModel,
    SparkXGBRanker,
    SparkXGBRankerModel,
    SparkXGBRegressor,
    SparkXGBRegressorModel,
)
```
Meanwhile, isort will complain that they are not formatted like this:
```
from xgboost.spark import (SparkXGBClassifier, SparkXGBClassifierModel,
                           SparkXGBRanker, SparkXGBRankerModel,
                           SparkXGBRegressor, SparkXGBRegressorModel)
```

To resolve this, I created a pyproject.toml file located under the xgboost/tests/test_distributed/ folder and added the following configuration to it:
```
[tool.isort]
profile = "black"
```

This resolved the problem locally but not on the CI

ayoub317:
Thanks @wbo4958 ! I pushed another commit adding the tests. Any feedback is welcome.

python-package/xgboost/spark/core.py
```diff
@@ -1148,7 +1151,7 @@ def _train_booster(
                 if dvalid is not None:
                     dval = [(dtrain, "training"), (dvalid, "validation")]
                 else:
-                    dval = None
+                    dval = [(dtrain, "training")]
```
wbo4958, line 1154: @trivialfis, Could you check this is ok by enabling it by default?

tests/test_distributed/test_with_spark/test_xgboost_summary.py
```diff
@@ -0,0 +1,233 @@
...
+    SparkXGBClassifierModel,
+    SparkXGBRanker,
+    SparkXGBRankerModel,
+    SparkXGBRegressor,
+    SparkXGBRegressorModel,
+)
+
+from .test_spark_local import spark as spark_local
+
+logging.getLogger("py4j").setLevel(logging.INFO)
```
wbo4958, line 21: is this for debug?

tests/test_distributed/test_with_spark/test_xgboost_summary.py
```diff
@@ -0,0 +1,233 @@
+import logging
```
wbo4958, line 1: I'm wondering, if we could put the tests in this file into the existing test_spark_local.py and reuse the existing test data?

tests/test_distributed/test_with_spark/test_xgboost_summary.py
```diff
@@ -0,0 +1,233 @@
...
+        xgb_model: Union[
+            SparkXGBClassifierModel, SparkXGBRankerModel, SparkXGBRegressorModel
+        ]
+    ) -> None:
+        assert hasattr(xgb_model.training_summary, "validation_objective_history")
+        assert isinstance(xgb_model.training_summary.validation_objective_history, dict)
+        assert not xgb_model.training_summary.validation_objective_history
+
+    @staticmethod
+    def assert_non_empty_training_objective_history(
```
wbo4958, line 101: I'm wondering if we could get the evaluate_results from xgboost itself and the training summary from xgboost-pyspark on the same dataset, and then check if they are equal? You can see some tests in test_spark_local.py are doing same comparison.

tests/test_distributed/test_with_spark/test_xgboost_summary.py
ayoub317: Yes, and since it was also set in test_spark_local.py, I kept it. Do you prefer that we remove it ?

tests/test_distributed/test_with_spark/test_xgboost_summary.py
ayoub317: Yes, I can move them there without much effort. Let me know if you'd like me to proceed with that.
However, in my humble opinion, it’s better to keep them in this separate file, and here’s the rationale :
The test_spark_local.py file already exceeds 1800 lines of code, which makes it increasingly difficult to read, maintain and navigate. As new features are added to PySpark XGBoost, this file will only continue to grow, compounding the problem.
I think refactoring the tests to organize them by key features, rather than bundling everything under the TestPySparkLocal class would be a better long-term approach.

If we decide to keep the tests in this file, I can either leave the examples here as they are, or, as you suggested, for better modularity and data reuse, we could import them from test_spark_local.py. Another option is to store all shared data in a separate file, allowing both test_spark_local.py and test_xgboost_summary.py to import what they need from it.

Let me know what you think, I have no strong opinion on this.

tests/test_distributed/test_with_spark/test_xgboost_summary.py
ayoub317: Yes, absolutely. I tested this on a simple DataFrame locally, and the results matched perfectly. We should definitely add such tests, I’ll take care of that !
That said, I believe this is more of a test for the PySpark XGBoost implementation itself rather than the training summary we expose. Just a note : adding these comparisons will slightly increase the complexity of the test logic.

tests/test_distributed/test_with_spark/test_xgboost_summary.py
wbo4958: Yeah, That's good point. Originally, I would like to separate the tests per the estimators. like XGBoostClassifier/Regressor/Ranker, instead of per features. So you can share the same dataset for different features.

ayoub317: Perfect ! I like that approach more than what I proposed ! I might work on it sometime soon if no one else does, especially since I have some other features I would like us to introduce.

ayoub317:
Thanks for the review @wbo4958. I added a new commit to address what we discussed.
It seems the failures in the CI are not related to this PR.

python-package/xgboost/spark/core.py
wbo4958: Hi @trivialfis, I wonder if this default eval dataset is necessary?

wbo4958 (approved):
LGTM,

python-package/xgboost/spark/core.py
trivialfis: Will look into this.

trivialfis: Looks good to me, it's unlikely someone will train a model without any evaluation.

trivialfis (approved):
Looks good to me.

trivialfis:
Let me fix the CI first https://github.com/dmlc/xgboost/pull/11162 .

PR #11137, opened 2025-01-01
Update model.rst

migolan:


trivialfis (commented):
The index is local to the sum.

migolan:
> The index is local to the sum.

Of course, but everywhere else in this document, the `f` functions are enumerated with a subscript index `k`, and `i` is used to denote data samples. There were several other occurrences of this in the documentation before and they have been fixed.

trivialfis (approved):

PR #11141, opened 2025-01-04
[R] Enable prefetch on windows configure script

david-cortes:
ref https://github.com/dmlc/xgboost/pull/11124

This PR adds prefetching to the compilation flags on windows when supported.

Since CRAN only uses MinGW for windows builds and the msvc build here was dropped, it enables the GC builtin unconditionally, since no other compiler will be used on windows.

### DIFF
--- a/R-package/configure.win
+++ b/R-package/configure.win
@@ -1,0 +1,21 @@
+R_EXE="${R_HOME}/bin${R_ARCH_BIN}/R.exe"
+CXX=`"${R_EXE}" CMD config CXX`
+
+cat > test.cpp <<EOL
+#include <xmmintrin.h>
+int main() {
+  char data = 0;
+  const char* address = &data;
+  _mm_prefetch(address, _MM_HINT_NTA);
+  return 0;
+}
+EOL
+
+XGBOOST_MM_PREFETCH_PRESENT=""
+${CXX} -o test test.cpp 2>/dev/null && ./test && XGBOOST_MM_PREFETCH_PRESENT="-DXGBOOST_MM_PREFETCH_PRESENT=1"
+rm -f ./test
+rm -f ./test.cpp
+
+sed \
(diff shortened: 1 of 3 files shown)


david-cortes:
@trivialfis I'm not sure what to do about the CMake scripts that are failing in the CI here. This PR creates a file "Makevars.win.in" from which "Makevars.win" is created, but I guess the configure script that generates "Makevars.win" will not get executed by CMake.

trivialfis:
Looks like apple clang internal error for the Python test. On the other hand, do we need to worry about R `stringi`? https://github.com/dmlc/xgboost/actions/runs/12609778121/job/35143648760?pr=11141

david-cortes:
> Looks like apple clang internal error for the Python test. On the other hand, do we need to worry about R `stringi`? https://github.com/dmlc/xgboost/actions/runs/12609778121/job/35143648760?pr=11141

No idea about the stringi error. But besides the apple issue, there's a problem with the cmake build trying to use a fixed `Makevars.win`, which I am not sure how to solve. Perhaps it could create that file through an additional target or similar, but don't know what'd be the best way.

trivialfis:
That's weird, I will look into it.

trivialfis:
>  I'm not sure what to do about the CMake scripts that are failing in the CI here.

Could you please point me to the CI failure? I couldn't find it.

david-cortes:
Looks like the error got solved. This was the log:
https://github.com/dmlc/xgboost/actions/runs/12609648878/job/35143402945

However I don't see any windows cmake R job in the CI, and I'm not sure if it'd work under such a combination (windows + cmake + R + non-msvc compiler).

trivialfis:
I have done some tests today with Rtools43. Need some updates to get cmake+mingw+Windows working. Not sure if it worths the effort to maintain.

trivialfis (approved):

Issue #11143, opened 2025-01-06
CMake instructions for R are outdated


david-cortes:
The docs about building the R package from source with CMake here:
https://xgboost.readthedocs.io/en/latest/build.html#building-r-package-from-source
```
cmake -B build -S . -G"Visual Studio 16 2019" -A x64 -DUSE_CUDA=ON -DR_LIB=ON -DR_VERSION=4.0.0
```

.. still have instructions for MSVC, which is not supported anymore as we verified that it **doesn't** compile ALTREP classes correctly and leads to crashes .

It also has outdated instructions for what to add to path:
```
C:\rtools40\usr\bin
```

.. which uses a hard-coded `rtools40`, but that's not the latest version anymore (newer ones have default paths like `rtools43`, or perhaps higher for the latest version). Perhaps it could provide a command to obtain that path from R itself it that's possible, and an example for setting the `$PATH` variable through `set` for windows.

Further, the second path it suggests to add to `$PATH`:
```
C:\rtools40\mingw64\bin
```

.. is empty under the latest versions of RTools.

_Note: I'm not very familiar with CMake or windows, so I'm leaving this issue for the maintainers here._

trivialfis:
Let's drop support for that combination on windows and focus on the r default. It's better to have one working method than few methods that may or may not work.

It's a bit sad that we dropped MSVC though, mingw's network module is quite broken the last time I tried. I will want distributed training on Windows.

trivialfis:
I will update the doc.

david-cortes:
> Let's drop support for that combination on windows and focus on the r default. It's better to have one working method than few methods that may or may not work.
> 
> It's a bit sad that we dropped MSVC though, mingw's network module is quite broken the last time I tried. I will want distributed training on Windows.

What about clang? Also, if there's no CMake build for windows, what would be the way to build the GPU-enabled package?

trivialfis:
GPU build on Windows requires MSVC or clang (I haven't tested clang), googling shows some people try to use nvcc with mingw, but let's not take that too serious. Cmake build will continue to be supported on Linux.

Clang should work, at least apple clang and clang on Linux are good. I haven't tested it on Windows. I don't use Windows very often, even with Windows, I only use the WSL.

For now, the Windows version of the R package is limited to single node CPU implementation. It's not too bad, people can still use WSL.

PR #11145, opened 2025-01-07
[doc] Cleanup document for building from source. [skip ci]

trivialfis:
- Use `ninja` uniformly and refer to CMake doc for other options. The latest CMake is cross platform, no need to add special instructions for each OS.
- Drop support for MSVC with the R package.
- Stop documenting methods we don't test, like mingw with the Python package.

Close https://github.com/dmlc/xgboost/issues/11143#issuecomment-2573647151 .

### DIFF
--- a/doc/build.rst
+++ b/doc/build.rst
@@ -2,10 +2,9 @@
 Building From Source
 ####################
 
-This page gives instructions on how to build and install XGBoost from the source code on various
-systems.  If the instructions do not work for you, please feel free to ask questions at
-`the user forum <https://discuss.xgboost.ai>`_.
-
+This page gives instructions on how to build and install XGBoost from the source code on
+various systems.  If the instructions do not work for you, please feel free to ask
+questions at `GitHub <https://github.com/dmlc/xgboost/issues>`__.
 
 .. note:: Pre-built binary is available: now with GPU support
 
@@ -19,23 +18,14 @@ systems.  If the instructions do not work for you, please feel free to ask quest
 *************************
 Obtaining the Source Code
 *************************
-To obtain the development repository of XGBoost, one needs to use ``git``.
 
-.. note:: Use of Git submodules
-
-  XGBoost uses Git submodules to manage dependencies. So when you clone the repo, remember to specify ``--recursive`` option:
+To obtain the development repository of XGBoost, one needs to use ``git``. XGBoost uses
+Git submodules to manage dependencies. So when you clone the repo, remember to specify
+``--recursive`` option:
 
   .. code-block:: bash
 
     git clone --recursive https://github.com/dmlc/xgboost
-
-For windows users who use github tools, you can open the git shell and type the following command:
-
-.. code-block:: batch
-
-  git submodule init
-  git submodule update
-
 
 .. _build_shared_lib:
 
@@ -44,7 +34,7 @@ Building the Shared Library
 ***************************
 
 This section describes the procedure to build the shared library and CLI interface
-independently.  For building language specific package, see corresponding sections in this
+independently. For building language specific package, see corresponding sections in this
 document.
 
 - On Linux and other UNIX-like systems, the target library is ``libxgboost.so``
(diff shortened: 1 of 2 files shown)


trivialfis:
cc @david-cortes @hcho3

trivialfis:
In the future, we might merge the Python script for building jvm packages into CMake scripts and eliminate the Python dependency.

doc/build.rst
```diff
@@ -459,32 +383,24 @@ If you want to use XGBoost4J-Spark, replace ``xgboost4j`` with ``xgboost4j-spark
...
-
-Building with GPU support
-==========================
-If you want to build XGBoost4J that supports distributed GPU training, run
-
-.. code-block:: bash
-
-  mvn -Duse.cuda=ON install
+- OpenMP on MacOS: See :ref:`running_cmake_and_build` for installing ``openmp``. The flag
+  -``mvn Duse.openmp=OFF`` can be used to disable OpenMP support.
```
hcho3, line 394:
```suggestion
  -``mvn -Duse.openmp=OFF`` can be used to disable OpenMP support.
```

hcho3 (changes requested):

hcho3 (approved):

doc/build.rst
```diff
@@ -317,51 +290,33 @@ simplest way to install the R package after obtaining the source code is:
   cd R-package
   R CMD INSTALL .
 
-As an alternative, the package can also be loaded through ``devtools::load_all()``
-from the same subfolder ``R-package`` in the repository's root, and by extension, can
-be installed through RStudio's build panel if one adds that folder ``R-package`` as an
-R package project in the RStudio IDE.
+Use the environment variable ``MAKEFLAGS=-j$(nproc)`` if you want to speedup the build. As
```
david-cortes, line 293: This doesn't work for R packages. `MAKEFLAGS` needs to be in the user file `~/.R/Makevars`.

doc/build.rst
```diff
@@ -317,51 +290,33 @@ simplest way to install the R package after obtaining the source code is:
...
-    - ``C:\rtools40\mingw64\bin``
+If you want to use the CMake build for greater flexibility around compile flags, the
+earlier snippet can be replaced by:
 
-  * If you are using R 3.x with RTools 3.x:
+.. code-block:: bash
 
-    - ``C:\Rtools\bin``
-    - ``C:\Rtools\mingw_64\bin``
+  cmake -B build -S . -DR_LIB=ON -GNinja
```
david-cortes, line 309: Wouldn't this make it use the MSVC compiler by default?

trivialfis: There's a warning in the end of this section, MSVC with R is not supported.

doc/build.rst
trivialfis: I'm running this command at the moment with Ubuntu 22.04,  and R 4.4.2, it's effective.

trivialfis: `MAKEFLAGS=-j$(nproc) R CMD INSTALL`

david-cortes: That's curious (doesn't work for me with `install.packages`). But would that work if you use devtools to install?

doc/build.rst
david-cortes: Perhaps it could be helpful to provide instructions for a clang install with a tool like conda to build the package with GPU support.

david-cortes:
@trivialfis Doesn't the CMake build for R require setting up other variables like the R version and similar?

doc/build.rst
trivialfis: Yes, it works.

``` R
devtools::load_all(path = "/home/jiamingy/workspace/xgboost_dev/xgboost/R-package")
```

``` bash
MAKEFLAGS=-j$(nproc) Rscript ./basic.R
```

doc/build.rst
trivialfis: Let me add a note that cmake is not supported either. I tried to make it work with Windows and the latest rtools, it's quite a hack. I don't expect users to do something similar.

The mingw is hidden deep in the rtools43 package.

trivialfis:
> Doesn't the CMake build for R require setting up other variables like the R version and similar?

On Linux, it's quite simple, only the `-DR_LIB=ON` is required. On Windows, I don't know.

david-cortes:
> > Doesn't the CMake build for R require setting up other variables like the R version and similar?
> 
> On Linux, it's quite simple, only the `-DR_LIB=ON` is required. On Windows, I don't know.

First error that I run into on windows:
```
CMake Error at cmake/modules/FindLibR.cmake:44 (message):
  LIBR_LIB_DIR was not set!
```

> Let me add a note that cmake on windows with R is not supported either.

But then how is `xgboost_r_gpu_linux_2.1.3.tar.gz` generated?

> I don't expect users to do something similar.

Perhaps suggesting installs of cmake, clang, and so on from conda could result in easier instructions?

trivialfis:
> Perhaps suggesting installs of cmake, clang, and so on from conda could result in easier instructions?

I use conda on Linux for installing gcc indeed, but have no experience with conda on Windows. Even on Linux, it takes some experience with the underlying runtime like locking:
```
- sysroot_linux-64==2.28
- gcc_linux-64=13.*
```
I don't expect Windows to be easier. In addition, I actually use `mamba-forge` instead of conda, last time I tried it on Windows, the cmd shell was quite broken, and I couldn't get powershell to work.

PR #11148, opened 2025-01-07
[R] Error out when trying to compile with MSVC

david-cortes:
ref https://github.com/dmlc/xgboost/pull/11145

This PR makes it throw a compilation error with a clear message when a user tries to build the R package with the MSVC compiler. Hopefully this should save users some headaches as they'd otherwise run into runtime segfaults if the package gets compiled with MSVC.

### DIFF
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -27,6 +27,10 @@
 
 #include "./xgboost_R.h"  // Must follow other includes.
 
+#ifdef _MSC_VER
+#error "Compilation of R package with MSVC is not supported due to issues handling R headers"
+#endif
+
 namespace {
 
 /* Note: this class is used as a throwable exception.


trivialfis:
The error looks good, but did you compile the R package with MSVC successfully? With the latest R, I couldn't compile the package at all with MSVC being confused by R's header files.

david-cortes:
> The error looks good, but did you compile the R package with MSVC successfully? With the latest R, I couldn't compile the package at all with MSVC being confused by R's header files.

I haven't tried compiling with MSVC. It fails due to other reasons before reaching the compilation stage.

trivialfis:
> It fails due to other reasons before reaching the compilation stage.

Somehow feels right.

I revised the CMake script for R a few years ago and got it running on the CI with an old R version. I think it's time for us to move away from that now. I couldn't get MSVC to work with the latest R.

trivialfis (approved):

PR #11151, opened 2025-01-07
[R] Update deprecated and removed parameter lists for `xgboost()`

david-cortes:
ref https://github.com/dmlc/xgboost/pull/11109
ref https://github.com/dmlc/xgboost/pull/11095

This PR adds previous arguments of `xgboost()` to the lists of deprecations and removals.

### DIFF
--- a/R-package/R/utils.R
+++ b/R-package/R/utils.R
@@ -560,6 +560,29 @@ deprecated_train_params <- list(
   ),
   removed = character()
 )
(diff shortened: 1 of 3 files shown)


trivialfis:
I will try to run a reverse dependency check tomorrow and share the result, so we can decide how to launch the new package.

trivialfis (approved):

PR #11166, opened 2025-01-14
[doc] Reference the R doc in sphinx document site.

trivialfis:
**todos**
- [x] Registry tag
- [x] Fix `pkgdown` error due to missing xgboost.
- [x] Check release branch.
- [x] Ensure consistency of documents in master and release branches.


trivialfis:
The integration is quite primitive. We can expand the integration after https://github.com/dmlc/xgboost/pull/11123 is merged if needed.

Thank you @jameslamb for sharing the use of `pkgdown`, cool project ;-)

[Spark](https://github.com/apache/spark/tree/master/R/pkg/pkgdown) and [LightGBM](https://github.com/microsoft/LightGBM/blob/60b0155ac573a8ad5994c74c49e05854281e2469/R-package/vignettes/basic_walkthrough.Rmd#L6-L14) both take a similar approach for document integration.

trivialfis:
https://github.com/readthedocs/readthedocs.org/issues/11928

doc/R-package/Makefile
```diff
@@ -4,6 +4,7 @@ PKGROOT=../../R-package
 # ADD The Markdown to be built here, with suffix md
 discoverYourData.md: $(PKGROOT)/vignettes/discoverYourData.Rmd
 xgboostPresentation.md: $(PKGROOT)/vignettes/xgboostPresentation.Rmd
+xgboostfromJSON.md: $(PKGROOT)/vignettes/xgboostfromJSON.Rmd
 
 # General Rules for build rmarkdowns, need knitr
 %.md:
```
david-cortes, line 10: Default option for this is to render to .html. The call needs to knitr/rmarkdown be modified to render to .md.

trivialfis:
I need some help from @hcho3 after he is back. We might want to cache the R pkgdown result in an S3 bucket, similar to the JVM documents. Building a whole chain of pkgdown dependencies on the RTD worker is not ideal.

doc/R-package/index.rst
```diff
@@ -9,22 +9,28 @@ XGBoost R Package
...
+Getting Started
+***************
+
+- Checkout the :doc:`Installation Guide </install>` contains instructions to install xgboost, and :doc:`Tutorials </tutorials/index>` for examples on how to use XGBoost for various tasks.
+- `XGBoost R tutorials and API reference <../r_docs/index.html>`__
+- `CRAN document <https://cran.r-project.org/web/packages/xgboost/xgboost.pdf>`__
 
 *********
-Tutorials
+Vignettes
```
david-cortes, line 21: Here it could try to embed the .md if it's somehow possible to generate them with the Makefile.

trivialfis:
still working in progress, need to upstream a docker image to the ops repo.

trivialfis:
https://github.com/dmlc/xgboost-devops/pull/6

doc/R-package/index.rst
```diff
@@ -13,7 +13,17 @@ You have found the XGBoost R Package!
 Get Started
 ***********
 * Checkout the :doc:`Installation Guide </install>` contains instructions to install xgboost, and :doc:`Tutorials </tutorials/index>` for examples on how to use XGBoost for various tasks.
-* Read the `API documentation <https://cran.r-project.org/web/packages/xgboost/xgboost.pdf>`_.
+* Read the `API documentation <../r_docs/R-package/docs/index.html>`__
```
david-cortes, line 16: Perhaps it could mention that these are the docs for the latest github version, while the ones at CRAN are for the latest release. Or will it have a version switcher like the sphinx docs?

doc/R-package/Makefile
trivialfis: Based on the tests so far, the result is md.

doc/R-package/index.rst
trivialfis: Done.

doc/R-package/index.rst
trivialfis: Done.

doc/conf.py
```diff
@@ -35,7 +37,35 @@
 release = xgboost.__version__
 
 
-def run_doxygen():
+# Witchcraft alert:
```
trivialfis, line 40: took me a while to figure out that the existing code was broken after the CI revamp. Would be great if we could use a package manager like conda to handle R packages instead of spawning out a new CI workflow.

trivialfis:
@david-cortes Please help review the output document: https://xgboost--11166.org.readthedocs.build/en/11166/ when you are available.

cc @hcho3 @jameslamb for CI changes, related: https://github.com/dmlc/xgboost-devops/pull/6

I spawned a new CI workflow for building R documents with `pkgdown`, along with some fixes to the existing jvm doc fetcher.

CUDA failures are not related, not sure what's updated there. Please ignore them for now.

doc/requirements.txt
```diff
@@ -10,8 +10,8 @@ numpy
 scipy
 myst-parser
 ray[train]
-xgboost_ray
```
trivialfis, line 13: this fetches an old version of xgboost from pypi.

doc/conf.py
```diff
@@ -67,26 +97,52 @@ def run_doxygen():
...
+
+
+def get_sha(branch: str) -> str | None:
+    sha = os.getenv("READTHEDOCS_GIT_COMMIT_HASH", default=None)
+    if sha is not None:
+        return sha
+
+    if branch == "master":
+        res = subprocess.run(["git", "rev-parse", "master"], stdout=subprocess.PIPE)
+    else:
```
trivialfis, line 124: Do we need to check release branch?

ops/pipeline/build-r-docs.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if [[ -z ${BRANCH_NAME:-} ]]
+then
+  echo "Make sure to define environment variable BRANCH_NAME."
+  exit 1
+fi
+
+source ops/pipeline/get-docker-registry-details.sh
+
+IMAGE_URI=${DOCKER_REGISTRY_URL}/xgb-ci.cpu_build_r_doc:PR-6
```
trivialfis, line 13: Need to change this back to main.

david-cortes:
> @david-cortes Please help review the output document: https://xgboost--11166.org.readthedocs.build/en/11166/ when you are available. (many like `## Error: object 'model' not found` still needs to be fixed)
> 
> cc @hcho3 @jameslamb for CI changes, related: [dmlc/xgboost-devops#6](https://github.com/dmlc/xgboost-devops/pull/6)
> 
> I spawned a new CI workflow for building R documents with `pkgdown`, along with some fixes to the existing jvm doc fetcher.
> 
> CUDA failures are not related, not sure what's updated there. Please ignore them for now.

I think the core failure is:
> ## Error in library(xgboost): there is no package called 'xgboost'

Since the vignette is not being built through `R CMD`, it probably needs to install the generated artifact first, which should solve most of the errors. Either that, or load it through `devtools::load_all()` in the same Rscript call that builds the vignettes.

Also, looks like the TOC requires at least one deeper level:
image

Currently appears to be limited at two. Perhaps something like this would do:
```
.. toctree::
  :maxdepth: 3
```

------------------

Some suggestions for the pkgdown site:
* It's not obvious that one needs to click "Reference" to get to function documentations. Perhaps it could have it as a more obvious link in the description of the landing page, like it does for the sphinx links:
image
* Title "XGBoost R Package for Scalable GBM" doesn't quite give the impression that this is the documentation package for the R package. Perhaps it could use a more obvious title like "XGBoost R Package Documentation".

david-cortes:
These links point to "latest" version for some reason, even though the version being shown has the PR number:
image

david-cortes:
Looks like the pkgdown site is not rendering R equations correctly - could it be missing some plugin?
image

This is how it looks like in the in-package documentation in RStudio:
image

trivialfis:
> Also, looks like the TOC requires at least one deeper level:

Hmm, not sure.  I intentionally kept the navigation bar succinct.

> Title "XGBoost R Package for Scalable GBM" doesn't quite give the impression 

The landing page for the pkgdown wite is from the R package README file i believe.  We can revise that file.

> Could also have a warning that the version switcher only has that R documentation package starting at 3.0

We add documents all the time, I think it's ok to not have this warning. Otherwise we need to annotate many more new docs.

> Since the vignette is not being built through R CMD

Yes, I need to add a installation call before the knitr call. Will do it tomorrow. Great to have some initial feedback. 

> Looks like the pkgdown site is not rendering R equations correctly - could it be missing some plugin?

Good catch, need to investigate.

> These links point to "latest" version for some reason, even though the version being shown has the PR number:

It's the read me file I think? Need to double check,  it's hard linked, will change it back to stable if possible.

ops/pipeline/build-r-docs-impl.sh
```diff
@@ -0,0 +1,28 @@
...
+set -euo pipefail
+
+branch_name=$1
+
+echo "R_LIBS_USER: ${R_LIBS_USER}"
+
+gosu root chown -R $UID:$GROUPS ${R_LIBS_USER}
+
+cd R-package
+MAKEFLAGS=-j$(nproc) Rscript ./tests/helper_scripts/install_deps.R
```
david-cortes, line 18: Some of these R dependencies have system dependencies - see here for example: https://github.com/dmlc/xgboost/pull/11101/files#diff-17796461dde560588f341bc929041c6576bc9ffffa049121918f7d950ae0ee18

david-cortes:
> It's the read me file I think? Need to double check, it's hard linked, will change it back to stable if possible.

If it _needs_ to be hard-coded, guess "latest" is a better choice. But other internal links elsewhere point to the same version that one is currently browsing.

david-cortes:
> I intentionally kept the navigation bar succinct.

I think it'd be helpful to have more levels. Especially for the python package docs.

trivialfis:
Found a related issue for the tex rendering https://github.com/r-lib/pkgdown/issues/2704

> Some of these R dependencies have system dependencies 

System dependencies are handled by the docker image.

> But other internal links elsewhere point to the same version that one is currently browsing.

Will double check.

trivialfis:
- The tex rendering is fixed, albeit not pretty. We need a fix in pkgdown instead.
- Changed the title in the `R-package/README.md`, and did a small cleanup there as well. User documents are linked to stable, while development-related items are linked to the latest.
- vignettes are fixed by installing xgboost.
- API link now points to the reference page instead of the landing page.
- Grouped the functions in reference page.


I will leave other documentation modifications for future PRs and focus on getting the page to build.

trivialfis:
@hcho3 Do you think it's possible to use a github webhook to trigger readthedocs build, preferably called inside a github action?

Currently, the JVM docs (and R docs in this PR) are built by github actions and downloaded by the read the docs build worker. However, there's no existing mechanism to synchronize the two services. As a result, it's necessary to manually trigger a rebuild in RTD to render the latest document.

related:
- https://docs.readthedocs.io/en/stable/guides/setup/git-repo-manual.html#manual-integration-setup
- https://github.com/marketplace/actions/workflow-webhook-action

R-package/pkgdown/_pkgdown.yml
```diff
@@ -0,0 +1,93 @@
+url: https://github.com/dmlc/xgboost
+
+template:
+  bootstrap: 5
+  math-rendering: mathjax
```
david-cortes, line 5: @trivialfis What about the `katex` with the workaround that is mentioned in the issue you linked?
https://github.com/r-lib/pkgdown/issues/2704#issuecomment-2307055568

Looks like it should solve the issue of symbols like sqrt showing random glitches.

R-package/pkgdown/_pkgdown.yml
```diff
@@ -0,0 +1,93 @@
...
+      - "xgb.cb.print.evaluation"
+      - "xgb.cb.reset.parameters"
+      - "xgb.cb.save.model"
+  - title: Low-level Training Functions
+    desc: Low-level Training Functions with DMatrix and Booster
+    contents:
+      - "xgb.params"
+      - "xgb.train"
+      - "xgb.cv"
+      - "print.xgb.cv.synchronous"
```
david-cortes, line 89: Also `predict.xgb.Booster` and `print.xgb.Booster`.

R-package/pkgdown/_pkgdown.yml
```diff
@@ -0,0 +1,93 @@
+url: https://github.com/dmlc/xgboost
+
+template:
+  bootstrap: 5
+  math-rendering: mathjax
+
+reference:
```
david-cortes, line 7: Could the CI job be configured in such a way that it'd throw a warning if some function is exported but not listed in this file?

R-package/pkgdown/_pkgdown.yml
trivialfis: Will test it.

R-package/pkgdown/_pkgdown.yml
trivialfis: These are in the booster section. Is that correct?

R-package/pkgdown/_pkgdown.yml
trivialfis: Pkgdown throws an error if there's a missing symbol.

R-package/pkgdown/_pkgdown.yml
david-cortes: Yes. Not sure where they'd fit better though.

R-package/pkgdown/_pkgdown.yml
trivialfis: Let's not do that. Injecting npm dependencies with a fixed version is outside my comfort zone. I know nothing about web development and would like to avoid it in XGBoost.

R-package/pkgdown/_pkgdown.yml
trivialfis: Seems logical to put them in the booster section as they are booster s3 methods.

doc/conf.py
hcho3: Not sure why it's there

trivialfis: We need to get the commit hash for the correct artifact. Otherwise we need to specialize the GitHub workflow for release branch

trivialfis:
> It would break existing links to https://xgboost.readthedocs.io, unfortunately.

Hmm, that's not ideal. Maybe we have to find a way to build the doc using RTD. Will look into it when I get back.

hcho3:
We might be able to set up a redirect to keep the old links working. See https://docs.readthedocs.com/platform/stable/user-defined-redirects.html#migrating-your-documentation-to-another-domain

hcho3:
@trivialfis I found out that ReadTheDocs offers a dedicated REST API to trigger doc builds. See https://docs.readthedocs.com/platform/stable/api/v3.html#post--api-v3-projects-(string-project_slug)-versions-(string-version_slug)-builds-

I just tried it and it works flawlessly.

hcho3:
One more improvement: when the CI pipeline builds tarballs for JVM and R docs, it should tag tarballs with GitHub hashes so that ReadTheDocs can download the correct version.

trivialfis:
> just tried it and it works flawlessly.

That's great news!

> tag tarballs with GitHub hashes 

I'm currently using git hashes instead just because I'm familiar with it. Not sure how GitHub hashes work.

hcho3:
I meant git hashes, not GitHub hashes.

trivialfis:
PR cannot read gh secrets, debugging using a different branch https://github.com/dmlc/xgboost/tree/trigger-rtd .

trivialfis:
The CI change is ready for review now. Thank you @hcho3 for all the helps!

Some notes:
- RTD is run twice for each PR. Let me test if this still works if I disable the PR webhook.
- We might want to cache the R dependencies in the container image. We can copy the R dependencies specification there or just let docker fetch it during image update.

trivialfis:
Disabled the automatic PR build. See if the GA can trigger it.

trivialfis:
Turned back on. It will have to be built twice for now.

doc/contrib/docs.rst
```diff
@@ -81,6 +81,40 @@ Build the Python Docs using pip and Conda
...
+
+The ``xgboost/doc/tmp`` is part of the ``html_extra_path`` sphinx configuration specified
+in the ``conf.py`` file, which informs sphinx to copy the extracted html files to the
+build directory. Following is a list of environment variables used by the fetchers in
+``conf.py``:
+
+ - READTHEDOCS: Read the docs flag. Build the full documentation site including R, JVM and
+   C doc when set to ``True`` (case sensitive).
+ - XGBOOST_R_DOCS: Local path for pre-build R document, used for development.
+ - XGBOOST_JVM_DOCS: Local path for pre-build JVM document, used for development.
```
hcho3, line 111:
```suggestion
 - ``READTHEDOCS``: Read the docs flag. Build the full documentation site including R, JVM and
   C doc when set to ``True`` (case sensitive).
 - ``XGBOOST_R_DOCS``: Local path for pre-built R document, used for development.
 - ``XGBOOST_JVM_DOCS``: Local path for pre-built JVM document, used for development.
```

hcho3 (approved):

hcho3 (approved):

trivialfis:
Hmm, triggering the build for the master branch doesn't update the "latest" document.

trivialfis:
Opened an issue: https://github.com/readthedocs/readthedocs.org/issues/11993

PR #11172, opened 2025-01-17
Fix categorical data check.

trivialfis:
The original plan was to add cat support after https://github.com/dmlc/xgboost/issues/11088 is done. Now the feature has delayed, we need to fix the check.


trivialfis:
cc @rongou

rongou (approved):

PR #11173, opened 2025-01-18
Provide error message when a pool has taken all memory.

trivialfis:
Without the check, XGBoost emits a range error, something like `CHECK(begin < end)`.


trivialfis:
cc @rongou .

rongou (approved):

PR #11174, opened 2025-01-18
[dask] Warn about the default setting of `OMP_NUM_THREADS`.

trivialfis:


trivialfis:
https://github.com/dask/distributed/blob/09855eeeea75d05182e5d48a96cd6507498b1b51/distributed/distributed.yaml#L197

trivialfis:
Will use global config instead.

PR #11182, opened 2025-01-23
Optimization of PushRowPage for high number of cpu cores

razdoburdin:
This PR adds optimal row-wise parallelization for `PushRowPage`.
The current realization is ineffective for multithread CPUs in case of a dataset has large number of rows and small number of columns. Current PR fixes this bottleneck. I observed up to 15% perf improvements for large datasets like airline on system with 2x56 cores CPU.

**upg:**
Some illustrations of changes in parallel processing:

We have a sparse input matrix. Each cell corresponds to a feature, and we process each cell individually, placing the results into the output array.
image

The original implementation processes data column by column. This approach is suboptimal when there are many CPU cores but only a few columns.
image

In the initial optimization, each thread processed a subset of rows but included all columns. This approach required significant memory overhead because buffers for intermediate results had to be used.
image

In the current version, I have reorganized the logic. Now, the input matrix is analyzed by the `LoadBalance` function. If a column is too large, it is split across multiple threads. This reduces buffer sizes since each thread now processes only a single feature.
image


trivialfis (commented):
Thank you for the optimization! Does the existing test cover both cases? If not could you please consider extracting the code for small unit tests? Also, would you like to share your benchmark results?

razdoburdin:
> Thank you for the optimization! Does the existing test cover both cases? If not could you please consider extracting the code for small unit tests? Also, would you like to share your benchmark results?

Here is the speed-up in comparison with the current master branch.
image

As for unit-tests, it is not easy to cover such a case. I will try to figure it out.

trivialfis:
One particular caution about similar optimizations, thread local buffers can cost significant amount of memory usage when the core count is high. For large server CPUs with hundreds of CPUs this can be server. We partially reverted a similar optimization in the CPU predictor which uses thread local blocks due to this reason.

trivialfis:
Ref https://github.com/dmlc/xgboost/issues/6659

razdoburdin:
Sorry for a long delay. I have reorganized the optimization to reduce memory overhead. I also added some tests to verify the changes.

razdoburdin:
> Thank you for continuing the optimization work. Could you please share the latest profiling result including performance and memory usage on the datasets that you are targeting?

Here is the benchmarking for the last version of the PR. The numbers in the table >1 mean that this PR is faster than master branch, and consuming more memory.
image

src/common/quantile.h
```diff
@@ -748,37 +748,79 @@ std::vector<bst_idx_t> CalcColumnSize(Batch const &batch, bst_feature_t const n_
...
 template <typename Batch, typename IsValid>
-std::vector<bst_feature_t> LoadBalance(Batch const &batch, size_t nnz, bst_feature_t n_columns,
-                                       size_t const nthreads, IsValid&& is_valid) {
-  /* Some sparse datasets have their mass concentrating on small number of features.  To
-   * avoid waiting for a few threads running forever, we here distribute different number
-   * of columns to different threads according to number of entries.
+WLBalance LoadBalance(Batch const &batch, size_t nnz, bst_feature_t n_columns,
+                      size_t const nthreads, IsValid&& is_valid) {
+  /* Some datasets have long columns. It is beneficial to split such columns between threads and 
+   * than collect the result if number of threads is high enourth. In this case, each thread being
```
trivialfis, line 772:
```suggestion
   * than collect the result if the number of threads is high enough. In this case, each thread being
```

src/common/quantile.h
```diff
@@ -748,37 +748,79 @@ std::vector<bst_idx_t> CalcColumnSize(Batch const &batch, bst_feature_t const n_
...
-    if (count > entries_per_thread) {
-      current_thread++;
-      count = 0;
-      cols_ptr.at(current_thread) = cols_ptr[current_thread - 1];
+  size_t count = 0;
+  for (size_t column_idx  = 0; column_idx < n_columns; ++column_idx) {
+    size_t n_entries = entries_per_columns[column_idx];
+
+    if (n_entries > 0) {
+      size_t n_splits =  std::min(nthreads * n_entries / total_entries, n_entries);
```
trivialfis, line 794: Why is `n_splits` calculated in this way?

src/common/quantile.h
```diff
@@ -748,37 +748,79 @@ std::vector<bst_idx_t> CalcColumnSize(Batch const &batch, bst_feature_t const n_
...
+    std::vector<size_t> columns;
+    size_t split_idx = 0;
+    size_t n_splits = 1;
+
+    ThreadWorkLoad() : columns() {}
+  };
+
+  std::vector<ThreadWorkLoad> baskets;
+  std::vector<bool> is_column_splited;
+  bool has_splitted = false;
```
trivialfis, line 764: `is_spitted`/`has_been_xxx`?

src/common/quantile.h
```diff
@@ -840,46 +882,126 @@ class SketchContainerImpl {
...
+           *  collect info from the other ones.
+           */
+          size_t column_idx = wl.columns.front();
+
+          typename WQSketch::SummaryContainer main_summary;
+          main_summary.Reserve(sketches_[column_idx].limit_size);
+          typename WQSketch::SummaryContainer split_summary;
+          split_summary.Reserve(2 * sketches_[column_idx].limit_size);
+          typename WQSketch::SummaryContainer comb_summary;
+          comb_summary.Reserve(3 * sketches_[column_idx].limit_size);
```
trivialfis, line 987: How are these `2 *` and `3 *` sizes calculated?

trivialfis (commented):
Hmm, I don't quite understand the code. Do you think it's necessary to create this level of complexity for the gain?

src/common/quantile.h
razdoburdin: The goal is to assign for processing of the columns amount of threads being proportional to number of entries in the column.

src/common/quantile.h
razdoburdin: The size of split_summary size guaranties absence of intermediate pruning while calculating summary. While the size of comb_summary is the sum of sizes of main_summary and split_summary

razdoburdin:
> Hmm, I don't quite understand the code. Do you think it's necessary to create this level of complexity for the gain?
The level of complexity required for this optimization has increased since I started working on it. 😞
Overall, I'm trying to improve scalability for modern high-end CPUs, but it's still not good enough.

image

I suggest we postpone this PR and revisit it once other hotspots are resolved—at that point, the importance of this fix will become more significant.

trivialfis:
Thank you for sharing. Yes, please take your time and feel free to open refactoring PRs before optimization PRs. This way we can (hopefully) build some abstractions and utilities before optimization.

PR #11184, opened 2025-01-26
[JVM-Packages] Allow XGBoost jvm package run on GPU without spark-rapids

wbo4958:
When I tried to run a simple xgboost4j spark  gpu example with GPU, xgboost throws exception indicating "Couldn't find ai/rapids/Table" which is in spark-rapids. Which is not correct, since xgboost4j-spark-gpu shouldn't be coupled with spark-rapids.

This PR fixes this issue.


wbo4958:
Hi @trivialfis, Please help review it. Thx.

trivialfis (approved):

Issue #10994, opened 2024-11-12
Test fails with Dask 2024.11.0+


hcho3:
https://github.com/dmlc/xgboost/actions/runs/11753771153/job/32747003155

```
E                       distributed.client.FutureCancelledError: ('_argmax-06657a445bd2e0d811c6ff48d5860817', 24) cancelled for reason: unknown.

../../mambaforge/envs/linux_cpu_test/lib/python3.10/site-packages/distributed/client.py:2427: FutureCancelledError
-------------------------------------------------- Captured log setup --------------------------------------------------
INFO     distributed.http.proxy:proxy.py:85 To route to workers diagnostics web server please install jupyter-server-proxy: python -m pip install jupyter-server-proxy
INFO     distributed.scheduler:scheduler.py:1755 State start
INFO     distributed.scheduler:scheduler.py:4225   Scheduler at:     tcp://127.0.0.1:45365
INFO     distributed.scheduler:scheduler.py:4240   dashboard at:  http://127.0.0.1:35259/status
INFO     distributed.scheduler:scheduler.py:8115 Registering Worker plugin shuffle
INFO     distributed.nanny:nanny.py:368         Start Nanny at: 'tcp://127.0.0.1:44537'
INFO     distributed.nanny:nanny.py:368         Start Nanny at: 'tcp://127.0.0.1:32881'
INFO     distributed.scheduler:scheduler.py:4579 Register worker <WorkerState 'tcp://127.0.0.1:41799', name: 0, status: init, memory: 0, processing: 0>
INFO     distributed.scheduler:scheduler.py:6171 Starting worker compute stream, tcp://127.0.0.1:41799
INFO     distributed.core:core.py:883 Starting established connection to tcp://127.0.0.1:51116
INFO     distributed.scheduler:scheduler.py:4579 Register worker <WorkerState 'tcp://127.0.0.1:42755', name: 1, status: init, memory: 0, processing: 0>
INFO     distributed.scheduler:scheduler.py:6171 Starting worker compute stream, tcp://127.0.0.1:42755
INFO     distributed.core:core.py:883 Starting established connection to tcp://127.0.0.1:51122
INFO     distributed.scheduler:scheduler.py:5925 Receive client connection: Client-1b5e6d99-a113-11ef-bd45-00c0cab020ae
INFO     distributed.core:core.py:883 Starting established connection to tcp://127.0.0.1:51132
-------------------------------------------------- Captured log call ---------------------------------------------------
[...]

PR #11194, opened 2025-01-30
[CI] Unpin Dask, adopt CUDA 12.8 and RAPIDS 24.12

hcho3:
Related: https://github.com/dmlc/xgboost-devops/pull/9, dmlc/xgboost-devops#11

Use latest Dask, CUDA 12.8, and RAPIDS 24.12.

Also it's now easy to update the image tag in all pipelines simultaneously.


hcho3:
@trivialfis We have a problem here: Latest RAPIDS cannot be installed with Dask 2024.10.0, and yet our tests cannot pass with latest Dask. 😢

jakirkham:
Thanks Hyunsu! 🙏

What are the failures we are seeing?

hcho3:
> What are the failures we are seeing?

See:
* #10994
* https://github.com/dmlc/xgboost/actions/runs/13057002018/job/36432128685?pr=11194 (from #11194)

> E                       distributed.client.FutureCancelledError: ('_argmax-16ef6ea6596412a85b175e90dab51ecb', 44) cancelled for reason: unknown.

/opt/miniforge/envs/gpu_test/lib/python3.10/site-packages/distributed/client.py:2427: FutureCancelledError

trivialfis:
> We have a problem here: Latest RAPIDS cannot be installed with Dask 2024.10.0, and yet our tests cannot pass with latest Das

I assume it's caused by dask dataframe. Can we disable some of the tests for now? Or use arrays in place of dataframes? I think it will take some time for dask to go over all the issues. It's unlikely we halt the ci for that.

hcho3:
Yes, I'll probably skip the problematic test, as done in the 2.1 branch: https://github.com/dmlc/xgboost/blob/62e7923619352c4079b24303b367134486b1c84f/tests/test_distributed/test_gpu_with_dask/test_gpu_with_dask.py#L388-L393

tests/test_distributed/test_gpu_with_dask/test_gpu_with_dask.py
```diff
@@ -10,6 +10,7 @@
 import pytest
 from hypothesis import given, note, settings, strategies
 from hypothesis._settings import duration
+from packaging.version import parse as parse_version
```
trivialfis, line 13: This is not in the stdlib, do we need to specify it in our testing environments?

trivialfis (approved):
Looks good to me, one minor question.

tests/test_distributed/test_gpu_with_dask/test_gpu_with_dask.py
hcho3: Currently, our CI already has `packaging` installed (due to it being a dependency of other packages we use). However, if we were to specify test dependencies in the pyproject.toml, we should definitely list `packaging` as part of the `[test]` suite of optional dependencies.

PR #11197, opened 2025-02-02
[R] Add migration guide for new xgboost

david-cortes:

This PR adds a short migration guide for the new XGBoost highlighting the main differences from the versions that are currently on CRAN.

It also adds links to this guide throughout the docs for `xgboost()` and `xgb.train()`, and re-writes the header of the `xgb.train` docs as currently it wasn't very informative.


trivialfis:
Thank you for working on this, will look into it once I can get the doc build sorted out.

trivialfis:
it would be great if @mayer79 could help take a look as well. ;-)

doc/R-package/migration_guide.rst
```diff
@@ -0,0 +1,45 @@
...
+    - Arguments that are meant to be consumed by the DMatrix constructor must be passed directly to ``xgb.DMatrix`` instead (e.g. argument for categorical features or for feature names).
+    - Some arguments have been renamed (e.g. previous 'watchlist' is now 'evals', in line with the Python package).
+    - The format of the callbacks to pass to ``xgb.train`` has largely been re-written. See the documentation of ``xgb.Callback`` for details.
+
+- Function ``xgb.DMatrix()``:
+    - This function now accepts 'data.frame' inputs and determines which features are categorical from their types - anything with type 'factor' or 'character' will be considered as categorical. Note that when passing data to the 'predict' method, the 'factor' variables must have the same encoding (i.e. same levels) as XGBoost will not re-encode them for you.
```
mayer79, line 24: Factors are not re-encoded at prediction time. But what about characters?

david-cortes: They are not recoded either: https://github.com/dmlc/xgboost/blob/88c8d1ae59e48f49adec9c94dcb9e617fee70b13/R-package/R/xgb.Booster.R#L98

mayer79: Unrelated to the PR: this means that single-row predictions with character columns will be completely wrong, right? I feel we should think of fixing this already before the release. 

We would need to 

- turn characters into factors,
- store a named list with names and levels of character/factor columns,
- and maybe a vector of feature names (to allow users to predict on data frames with more/shuffled columns).

`predict()` would need these info to make safe predictions. 

I am aware that we have already discussed on relateed things, but it keeps popping up in my mind.

mayer79:
This will be very useful, thanks a lot. No idea where this bold sentence is coming from...

mayer79 (approved):

doc/R-package/migration_guide.rst
david-cortes: It will have to waiting until support for storing categorical metadata gets implemented in the core library.

trivialfis: https://github.com/dmlc/xgboost/issues/11088 will have to be part of the next release. Too much work for 3.0.

trivialfis:
> For some reason the RST renderer decided to put one point in bold:

I see the issue is fixed.

trivialfis (approved):

PR #11216, opened 2025-02-07
Use REST API to trigger RTD build.

trivialfis:
This PR uses the REST API for RTD to trigger document build for branches in XGBoost's main repo.

It relies on a private key to trigger the build. However, there's no straightforward way to read GitHub secrets for PRs due to security concerns, the trigger happens only for the main repo's branches. PR will continue to have documents missing.

For branches in the main repo, the document is built twice for now, once due to automatic webhook, and another one due to the CI trigger.

Yes, this is quite hacky.

Extracted from https://github.com/dmlc/xgboost/pull/11166 for easier review. Some additional notes are in that PR.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -1,0 +1,29 @@
+name: FreeBSD
+
+on: [push, pull_request]
+
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    name: A job to run test in FreeBSD
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: 'true'
+      - name: Test in FreeBSD
+        id: test
+        uses: vmactions/freebsd-vm@v1
+        with:
+          usesh: true
+          prepare: |
+            pkg install -y cmake git ninja googletest bash
+          run: |
+            bash ops/pipeline/test-freebsd.sh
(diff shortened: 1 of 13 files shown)


doc/conf.py
```diff
@@ -1,18 +1,10 @@
...
-# containing dir.
-#
-# Note that not all possible configuration values are present in this
-# autogenerated file.
-#
-# All configuration values have a default; values that are commented out
-# serve to show the default.
+"""Sphinx configuration.
+
+See `doc/contrib/docs.rst <https://xgboost.readthedocs.io/en/stable/contrib/docs.html>`__
```
trivialfis, line 3: This is still in the https://github.com/dmlc/xgboost/pull/11166

trivialfis:
@hcho3 

The PR fetch for jvm doc doesn't work at the moment. The doc has to be built as a dmlc branch.
- PR doesn't have permission to read GitHub secret to trigger a rtd build.
- By default, it should fallback to the using artifacts from the master branch. But it doesn't have the git commit hash for the master branch to identify the s3 bucket path.

Either we keep only one copy of the master branch artifact to avoid git hash, or we need to do some checkouts inside the RTD builder. But this is getting a bit confusing for me now.

hcho3:
@trivialfis We can actually access the RTD token from a PR, as follows:
1) Run the RTD trigger job using RunsOn, not GH Action
2) Use AWS Secret Manager to securely pull the RTD token into the RunsOn worker. Each RunsOn worker (which is an EC2 VM instance) has permissions configured to access certain AWS resources.

Let me set this up today.

trivialfis:
Thank you for looking into this. Please let me know if are busy. I will try to learn more about the AWS secret manager.

hcho3:
@trivialfis Now PRs can also trigger new builds in RTD.

hcho3:
I registered my key in the secrets manager.

Issue #11218, opened 2025-02-07
GPU external memory doesn't support specifying `cuda:ordinal` yet.


trivialfis:
Need some additional changes to support specifying `ordinal`.

PR #11219, opened 2025-02-07
Support CUDA ordinal for external memory.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/11218 .

Initialize the global device ordinal for new threads.


trivialfis:
cc @rongou .

rongou (approved):

PR #11232, opened 2025-02-10
[R] Switch to `xgb.train` inside `xgboost()` when passing DMatrix and no "y"

david-cortes:
ref https://github.com/dmlc/xgboost/issues/9734#issuecomment-2648010780

This PR adds an ugly workaround for supporting cases in which `xgboost()` is called with an `xgb.DMatrix` containing the `label` field, since this used to work in previous versions.

Note: I'm dropping the PR here in case we need it for compatibility. Not yet sure if it this would be good to merge.


bastistician:
Given this was a common error in reverse dependency checks, it seems prudent to catch that and at least give some advice to ease the changeover for users and developers. (I cannot comment on the specific fallback code.)

I'd also recommend making the (temporary) "xgboost.strict_mode" mechanism settable via an environment variable, so developers can enable it easily to test their packages, including when running R CMD check.

bastistician:
FWIW, I can confirm the result of @trivialfis in https://github.com/dmlc/xgboost/issues/9734#issuecomment-2657065306 that this PR causes new errors with message

> argument "x" is missing, with no default

leading to a total of 49 revdeps with new problems rather than 44, see [worse_20250216.txt](https://github.com/user-attachments/files/18816685/worse_20250216.txt).

PR #11248, opened 2025-02-12
Check for invalid cuDF input.

trivialfis:
Make sure there's no mask in cuDF-based meta data. Otherwise, there's a JSON serialization error for the mask object.

### DIFF
--- a/tests/python-gpu/test_from_cudf.py
+++ b/tests/python-gpu/test_from_cudf.py
@@ -382,3 +382,20 @@ def test_from_cudf_iter(enable_categorical):
     predict = reg.predict(m)
     predict_with_it = reg_with_it.predict(m_it)
     np.testing.assert_allclose(predict_with_it, predict)
+
+
+def test_invalid_meta() -> None:
+    df = cudf.DataFrame({"f0": [0, 1, 2], "f1": [2, 3, 4], "y": [None, 1, 2]})
+    y = df["y"]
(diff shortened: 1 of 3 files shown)


tests/python-gpu/test_from_cudf.py
```diff
@@ -382,3 +382,20 @@ def test_from_cudf_iter(enable_categorical):
...
+
+
+def test_invalid_meta() -> None:
+    df = cudf.DataFrame({"f0": [0, 1, 2], "f1": [2, 3, 4], "y": [None, 1, 2]})
+    y = df["y"]
+    X = df.drop(["y"], axis=1)
+    with pytest.raises(ValueError, match="Missing value"):
+        xgb.DMatrix(X, y)
+    with pytest.raises(ValueError, match="Missing value"):
+        xgb.QuantileDMatrix(X, y)
```
hcho3, line 394: Do we have a guide for a workaround in this use case? Is it enough to make a copy of `X` and `y`? If so, maybe we should add an example for that?

hcho3: Ah I just missed the `None` in the `y` column. So if we remove `None` from the example, you won't get an error?

trivialfis: Correct.

hcho3 (approved):

Issue #11185, opened 2025-01-26
Incompatibility Between QuantileDMatrix and xgboost.cv Function


hugocool:
I have encountered an issue when attempting to perform cross-validation using the `xgboost.cv` function with a `QuantileDMatrix` as input. The operation results in the following error:

```
XGBoostError: [17:29:55] /workspace/src/data/iterative_dmatrix.h:88: Slicing DMatrix is not supported for Quantile DMatrix.
Stack trace:
  [bt] (0) /opt/pysetup/.venv/lib/python3.10/site-packages/xgboost/lib/libxgboost.so(+0x22dcbc) [0x7f7553e2dcbc]
  [bt] (1) /opt/pysetup/.venv/lib/python3.10/site-packages/xgboost/lib/libxgboost.so(+0x525d31) [0x7f7554125d31]
  [bt] (2) /opt/pysetup/.venv/lib/python3.10/site-packages/xgboost/lib/libxgboost.so(XGDMatrixSliceDMatrixEx+0x15b) [0x7f7553d448cb]
  [bt] (3) /usr/lib/x86_64-linux-gnu/libffi.so.8(+0x7e2e) [0x7f76ce294e2e]
  [bt] (4) /usr/lib/x86_64-linux-gnu/libffi.so.8(+0x4493) [0x7f76ce291493]
  [bt] (5) /opt/pyenv/versions/3.10.12/lib/python3.10/lib-dynload/_ctypes.cpython-310-x86_64-linux-gnu.so(+0x12ea5) [0x7f76cdcf0ea5]
  [bt] (6) /opt/pyenv/versions/3.10.12/lib/python3.10/lib-dynload/_ctypes.cpython-310-x86_64-linux-gnu.so(+0xc940) [0x7f76cdcea940]
  [bt] (7) /opt/pyenv/versions/3.10.12/lib/libpython3.10.so.1.0(_PyObject_MakeTpCall+0x8c) [0x7f76d4175e8c]
  [bt] (8) /opt/pyenv/versions/3.10.12/lib/libpython3.10.so.1.0(_PyEval_EvalFrameDefault+0x8f84) [0x7f76d4122ed4]
```

**Steps to Reproduce:**

1. Create a `QuantileDMatrix` from a dataset.
2. Attempt to perform cross-validation using `xgboost.cv` with this `QuantileDMatrix`.

**Expected Behavior:**

The `xgboost.cv` function should either:

- Support cross-validation with `QuantileDMatrix`.
[...]

Issue #11013, opened 2024-11-22
ValueError: could not convert string to float


Checunmily:
Hello, recently I tried xgb.cv and got an error when assigning save_best=True in xgb.callback.EarlyStopping().
I just used the demo privided at https://xgboost.readthedocs.io/en/latest/python/examples/cross_validation.html 
Here is the full code I used:
```
import os

import numpy as np

import xgboost as xgb

# load data in do training
CURRENT_DIR = os.path.dirname(__file__)
dtrain = xgb.DMatrix(
    os.path.join(CURRENT_DIR, "../data/agaricus.txt.train?format=libsvm")
)
param = {"max_depth": 2, "eta": 1, "objective": "binary:logistic"}
num_round = 2

print("running cross validation, disable standard deviation display and save best")
# do cross validation, this will print result out as
# [iteration]  metric_name:mean_value
res = xgb.cv(
    param,
    dtrain,
    num_boost_round=10,
    nfold=5,
    metrics={"error"},
    seed=0,
    callbacks=[
        xgb.callback.EvaluationMonitor(show_stdv=False),
        xgb.callback.EarlyStopping(3, save_best=True),
    ],
)
print(res)
```
Finally I will got a 

> ValueError: could not convert string to float: '(0.001228171312096744, 0.0010411026299985337)'

When I tracing back, I found the function 

```
def best_score(self) -> float:
    """Get best_score."""
    return float(cast(float, self.cvfolds[0].bst.attr("best_score")))
```

 at 

> ~/.conda/envs/sklearn-env/lib/python3.12/site-packages/xgboost/training.py", line 271

 will return a float but 

> self.cvfolds[0].bst.attr("best_score")

 just give a str like 

> (0.001228171312096744, 0.0010411026299985337)

When I trying to skip this step I got another error:
```
File "~/.conda/envs/sklearn-env/lib/python3.12/site-packages/xgboost/callback.py", line 464, in after_training
    model = model[: best_iteration + 1]
            ~~~~~^^^^^^^^^^^^^^^^^^^^^^
TypeError: '_PackedBooster' object is not subscriptable
```
Did I miss something? How can I fix it? Thanks in advance.

trivialfis:
The `cv` function doesn't return a model, hence the `save_best` has no effect. It should not be used with cv.

PR #11255, opened 2025-02-13
Add checks for invalid inputs for `cv`.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/11013 .
Close https://github.com/dmlc/xgboost/issues/11185 .


david-cortes:
@trivialfis I guess this check would also need to be implemented in the corresponding R function.

hcho3 (approved):

trivialfis:
I will add a check for the R package in a follow-up PR

PR #11258, opened 2025-02-14
[DASK] Avoid persisting intermediate result

hendrikmakait:
This PR avoids persisting (materializing) an intermediate result during creation of the DMatrix. This materialization step could otherwise lead to a temporary increase in the memory footprint.


trivialfis:
Closing as the `to_delayed` is replaced.

PR #11257, opened 2025-02-14
Use https link for R document check.

trivialfis:

PR #11259, opened 2025-02-17
[R] Restric threads in DMatrix creation from examples and tests

david-cortes:
ref https://github.com/dmlc/xgboost/pull/11257#issuecomment-2660761981

Per the comment above, this PR modifies the calls to `xgb.DMatrix` in examples, test and vignettes to limit the amount of threads, in order to conform to CRAN policies.

### DIFF
--- a/R-package/tests/helper_scripts/generate_models.R
+++ b/R-package/tests/helper_scripts/generate_models.R
@@ -40,7 +40,7 @@ generate_regression_model <- function() {
   print('Regression')
   y <- rnorm(metadata$kRows)
 
-  data <- xgb.DMatrix(X, label = y)
+  data <- xgb.DMatrix(X, label = y, nthread = 1)
   params <- list(tree_method = 'hist', num_parallel_tree = metadata$kForests,
                  max_depth = metadata$kMaxDepth)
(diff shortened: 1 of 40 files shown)


R-package/tests/helper_scripts/generate_models.R
```diff
@@ -40,7 +40,7 @@ generate_regression_model <- function() {
   print('Regression')
   y <- rnorm(metadata$kRows)
 
-  data <- xgb.DMatrix(X, label = y)
+  data <- xgb.DMatrix(X, label = y, nthread = 1)
```
trivialfis, line 43: I didn't think this file is used in tests.

david-cortes: Is the file used for something outside of the R package tests, or can it be deleted? I couldn't find any usage anywhere, and the tests still pass if it gets deleted.

trivialfis: The model is generated once, then future tests use the generated models for testing compatibility.

david-cortes: Then I guess it doesn't hurt to limit threads there too, considering that the booster object also stores number of threads.

trivialfis (approved):

PR #11267, opened 2025-02-20
[R] Fix verbosity for training monitor.

trivialfis:
`monitor_training` should be considered for setting verbosity.

### DIFF
--- a/R-package/R/xgboost.R
+++ b/R-package/R/xgboost.R
@@ -1120,7 +1120,7 @@ xgboost <- function(
   min_split_loss = NULL,
   reg_lambda = NULL,
   weights = NULL,
-  verbosity = if (is.null(eval_set) && !monitor_training) 0L else 1L,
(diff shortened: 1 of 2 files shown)


trivialfis:
cc @david-cortes

david-cortes:
@trivialfis But with this PR, there would be no way of silently producing the table with evaluation metrics without printing them along the way.

trivialfis:
> But with this PR, there would be no way of silently producing the table with evaluation metrics without printing them along the way.

Any suggestion?

david-cortes:
> > But with this PR, there would be no way of silently producing the table with evaluation metrics without printing them along the way.
> 
> Any suggestion?

What would be the issue with leaving it like it is right now?

trivialfis:
No particular issue, just caught me by surprise that monitor evaluation doesn't output anything.

trivialfis:
Closing as non-issue.

Issue #11274, opened 2025-02-21
Quantile regression low CPU usage


jmpanfil:
When training models with `reg:quantileerror`, I am noticing very low CPU usage, maxing out around 9% as opposed to the usual high (70%+) CPU utilization when training more traditional models with `reg:tweedie`, `reg:squarederror`, etc. I am running training on Databricks using DBR 15.4LTS ML that has XGBoost 2.0.3 preinstalled. I am running the regular Python version (not pyspark or XGBoost4J) on a single node machine with 256GB RAM and 64 core CPU. Here's a screenshot of CPU utilization mid-training:

![Image](https://github.com/user-attachments/assets/90af3caa-5d02-4bc3-90ac-042eb6262512)

There is a momentary spike to 34% before settling to 9% for all of training. Compare to consistent 70% CPU utilization for `reg:squarrederror`:

![Image](https://github.com/user-attachments/assets/19d9d04f-8259-4758-927e-2148802205e1)

Both examples use the same code and hyperparameters except for objective.

```
import xgboost as xgb
xgb.train(...)
```
I have tested training a single quantile vs multiple, and it has no effect on CPU utilization. Input data is a `DMatrix` with underlying data being a highly sparse `scipy.sparse.csr_matrix`.

Is this behavior expected due to something inherent to quantile regression?

Thanks for your time as always.

trivialfis:
Hi, thank you for rasing the issue. Could you please share:
- The shape of the input data.
- The quantiles you are targeting. (`quantile_alpha`)
- The density of the data (n-non-missing/(n_samples * n_features).

trivialfis:
Doing a quick test with the following snippet using the nightly and 2.1.4, the CPU looks reasonably occupied:
``` python
import numpy as np
import xgboost as xgb
from xgboost.testing import make_sparse_regression
from time import time

alpha = np.array([0.05, 0.5, 0.95])

s = time()
X, y = make_sparse_regression(
    n_samples=int(2**20), n_features=256, sparsity=0.7, as_dense=False
)
e = time()
print(e - s)

s = time()
booster = xgb.train(
    {"quantile_alpha": [0.05, 0.5, 0.95], "objective": "reg:quantileerror"},
    num_boost_round=128,
    dtrain=xgb.QuantileDMatrix(X, y),
)
e = time()
print(e - s)
```

jmpanfil:
Hi @trivialfis,

[...]

PR #11275, opened 2025-02-21
Use parallel sort for quantile calculation when appropriate.

trivialfis:
- Drop the duplicated `omp_in_parallel` check.
- Use parallel sort instead of parallel leaf values based on a heuristic.

Close https://github.com/dmlc/xgboost/issues/11274


trivialfis:
cc @rongou .

PR #11279, opened 2025-02-24
Correct computation of metrics in multi-quantile regression.

### LINKED ISSUE #11278: Incorrect evaluation when using quantile regression with multiple quantiles
The current algorithm in element_wise_metric.cu does not account for the fact that when using quantile regression with multiple quantiles, there are as many predictions per label as there are quantiles. As a result only a fraction of 1 / (# of quantiles) of the number of elements in the dataset are actually used when computing the metric and the rest are ignored.

t-ens:
A quick-fix I used to get the evaluation metric working and some tests which catch the error.

closes #11278


trivialfis (approved):
Thank you for the fix!

Issue #11198, opened 2025-02-03
Accuracy degradation on bosch dataset


razdoburdin:
Hi,
I have observed the accuracy degradation for `bosch-production-line-performance` dataset after this PR https://github.com/dmlc/xgboost/pull/10298.
So, for commit ab8aa6f126156ad7fffb56a4a9baae06c0228f85 the accuracy is **0.844** for 100 estimators, but for ab228cc78d054fc6652728fdcfc229ee0e4de822 it drops to **0.762**.

trivialfis:
Could you please share your code for the benchmark? Is the accuracy from validation dataset or from the training dataset?

trivialfis:
Initial result from my test:
```
* master
[99]    Train-logloss:0.02704   Train-auc:0.89945       Valid-logloss:0.03350   Valid-auc:0.71049
Train acc 0.9946429979387481
Valid acc 0.9942825668976844
Duration: 38.64595603942871

* 2.1
[99]    Train-logloss:0.02409   Train-auc:0.91872       Valid-logloss:0.03095   Valid-auc:0.71511
Train acc 0.9948367330847816
Valid acc 0.9943873189226086
Duration: 40.097050189971924
```

razdoburdin:
Sorry for a log delay:
Here is a small reproducer:

```
import numpy as np
import xgboost as xgb

X_train = np.load('./data/bosch_x_train.npy')
X_test = np.load('./data/bosch_x_test.npy')
y_train = np.load('./data/bosch_y_train.npy')
y_test = np.load('./data/bosch_y_test.npy')

dtrain = xgb.DMatrix(X_train, y_train)
dtest = xgb.DMatrix(X_test, y_test)

params = {
    'booster': 'gbtree',
    'device': 'cpu',
    'tree_method': 'hist',
    'objective': 'binary:logistic',
    'max_depth': 8,
    'learning_rate': 0.1,
    'scale_pos_weight': 173.63348001466812
}

booster = xgb.train(params, dtrain, 100)

y_pred = booster.predict(dtrain)
y_pred = (y_pred >= 0.5).astype(np.int32)
print(f'train acc = {1 - np.sum(np.abs(y_pred - y_train)) / y_train.size}')
```

I have double checked the accuracy for current `master` branch and for `release_2.1.0`, and observed the degradation from **0.844** to **0.762**.

trivialfis:
Hmm, so there's a scale pos weight. The mean-based initialization method doesn't consider that.

[...]

PR #11280, opened 2025-02-24
Fix init with scale pos weight.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/11198

Use one-step Newton if scale pos weight is used.

Actually, ignoring the scale pos weight should *improve* the training performance since the weight explicitly instructs xgboost to create bias. I think the discrepancy for the bosch dataset is caused by a different initialization point for the Newton iteration.

One-step Newton is quite close to optimal for logistic loss. Also, the result isn't changed for unweighted data.

### DIFF
--- a/src/objective/regression_obj.cu
+++ b/src/objective/regression_obj.cu
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2024, XGBoost Contributors
+ * Copyright 2015-2025, XGBoost Contributors
  * \file regression_obj.cu
  * \brief Definition of single-value regression and classification objectives.
  * \author Tianqi Chen, Kailong Chen
@@ -9,7 +9,6 @@
 #include <algorithm>
 #include <cmath>
 #include <cstdint>  // std::int32_t
-#include <memory>
 #include <vector>
 
 #include "../common/common.h"
@@ -23,6 +22,7 @@
 #include "./regression_loss.h"
 #include "adaptive.h"
 #include "init_estimation.h"  // FitIntercept
+#include "regression_param.h"
 #include "xgboost/base.h"
 #include "xgboost/context.h"  // Context
 #include "xgboost/data.h"     // MetaInfo
@@ -34,8 +34,6 @@
 #include "xgboost/parameter.h"
 #include "xgboost/span.h"
 #include "xgboost/tree_model.h"  // RegTree
-
-#include "regression_param.h"
 
 #if defined(XGBOOST_USE_CUDA)
 #include "../common/cuda_context.cuh"  // for CUDAContext
@@ -102,7 +100,7 @@ class RegLossObj : public FitInterceptGlmLike {
     }
   }
   // 0 - scale_pos_weight, 1 - is_null_weight
-  RegLossObj(): additional_input_(2) {}
+  RegLossObj() : FitInterceptGlmLike{false}, additional_input_(2) {}
 
   void Configure(const std::vector<std::pair<std::string, std::string> >& args) override {
     param_.UpdateAllowUnknown(args);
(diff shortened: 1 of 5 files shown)


trivialfis:
Please help review when you are available @razdoburdin @david-cortes

david-cortes:
@trivialfis I cannot find much info in the docs about what `scale_pos_weight` does.

Does it set weights for observations of the positive class in the same way as passing weights to the DMatrix does? If so, then the intercept for it should be obtainable by a weighted mean instead. And what's more, you shouldn't even need to do the calculation with a vector of weights, since the unweighted mean could be adjusted after-the-fact if you know what the scaling should be.

Otherwise, from what I see in the issue description, the performance increase from this change would be just a coincidence - this is an imbalanced dataset and one-step Newton happens to drive the number closer to zero, so if the imbalance were to be towards the other side it should have the opposite effect.

Also from the issue: it looks like the parameters being tried are very suboptimal since the test accuracy (from a quick look at the data description, haven't seen it in detail) appears to be below what you'd obtain from constant predictions. A better metric to follow for such purposes would be the training logloss (which is what xgboost is optimizing for), or the test AUROC after tuning the hyperparameters.

trivialfis:
@david-cortes You are correct, as mentioned in the PR description, ignoring the weight should improve the training loss instead.

We can change it to mean-adjusting for logistic

david-cortes:
> @david-cortes You are correct, as mentioned in the PR description, ignoring the weight should improve the training loss instead.
> 
> We can change it to mean-adjusting for logistic

But if `scale_pos_weight` acts as weights, shouldn't it also be accounted in the logloss calculations?

trivialfis:
> But if scale_pos_weight acts as weights, shouldn't it also be accounted in the logloss calculations?

Currently not. I don't have a strong preference for this, since:
- Validation datasets should not be affected, this should be considered a training hyper-parameter.
- One can make an argument that hyper-parameter should not prevent metrics from making accurate estimations of model performance. Biased (weighted) estimation might not be desirable. But then we have sample weight, which acts differently. So, I will keep it as it's for now.

david-cortes:
@trivialfis Does `scale_pos_weight` have an effect on other functionalities that depend on a calculation of the objective function during training? For example, `min_split_loss`, leafwise growth policy, `reg_lambda`, etc.

If so, then it sounds like you might want to consider having an option to account for it in the **training** metrics calculations too. Could be helpful when the number is large enough that it substantially changes what the model is optimizing for.

trivialfis:
No, it affects only the gradient. It's a really old parameter, I don't think it's used consistently like sample weight, but so far proven useful.

trivialfis:
@david-cortes Please help take another look when you are available.

src/objective/regression_obj.cu
```diff
@@ -187,6 +185,31 @@ class RegLossObj : public FitInterceptGlmLike {
...
+    }
+
+    // Special handling for the scale_pos_weight parameter
+    auto w = this->param_.scale_pos_weight;
+    auto m = info.labels.Shape(0);
+    auto h_s = out.HostView();
+
+    for (std::size_t i = 0, n = h_s.Size(); i < n; ++i) {
+      // revert the mean back to sum, which is the number of positive samples
+      auto n_pos = h_s(i) * m;
```
david-cortes, line 203: If it has sample weights, here it would need to use the sum of weights instead of the number of rows.

src/objective/regression_obj.cu
```diff
@@ -187,6 +185,31 @@ class RegLossObj : public FitInterceptGlmLike {
...
+  void InitEstimation(MetaInfo const& info, linalg::Vector<float>* base_score) const override {
+    linalg::Vector<float> out;
+    FitInterceptGlmLike::InitEstimation(info, &out);
+    if (std::abs(this->param_.scale_pos_weight - 1.0f) <= kRtEps) {
+      common::Mean(this->ctx_, out, base_score);
+      return;
+    }
+
+    // Special handling for the scale_pos_weight parameter
+    auto w = this->param_.scale_pos_weight;
```
david-cortes, line 197: Since these are scalar calculations, they could be done in higher precision (fp64 or even 'long double') without loss of speed.

david-cortes:
@trivialfis I haven't looked at the code but: does `scale_pos_weight` apply also to objectives like `reg:logistic`?

trivialfis:
> I haven't looked at the code but: does scale_pos_weight apply also to objectives like reg:logistic?

Yes, but in regression, there are no "positive samples" or "negative samples", so one should not use this parameter.

src/objective/regression_obj.cu
trivialfis: I'm leaning toward using one-step newton instead. I don't want too much workaround for a single parameter in the initialization step.

david-cortes:
> > I haven't looked at the code but: does scale_pos_weight apply also to objectives like reg:logistic?
> 
> Yes, but in regression, there are no "positive samples" or "negative samples", so one should not use this parameter.

In _theory_ it shouldn't, but you could still generalize the calculation for decimal values:

$s \times y \log {p} + (1 - y) \log (1 - p)$

If `scale_pos_weight` gets applied to `reg:logistic`, the optimal intercept there should also need to change - adjustment would be exactly the same.

src/objective/regression_obj.cu
david-cortes: I would say the mean initialization approach is quite valuable in other important ways too.

For example, in the issue, it would make the model have an expected value of $\mathbb{E}[y] = 0.5$ by design, which would not be the case with a one-step Newton initialization.

david-cortes:
@trivialfis I'm thinking it might be better to throw an error when the data has both sample weights and `scale_pos_weight`.

trivialfis:
@david-cortes That's an option as well. But I'm preparing for a new release, let's not introduce a last-minute breaking for now. ;-)

> 

> If it has sample weights, here it would need to use the sum of weights instead of the number of rows.

I will revert to the Newton method if `scale_pos_weight` is used to avoid further complicating things.

trivialfis:
Reverted to using newton method. Will revisit it in the next release.

trivialfis:
Since this PR uses the same setting in 2.1, merging to branch out to 3.0.

PR #11289, opened 2025-02-26
[wip] Update release scripts

trivialfis:
- Generate the variants instead of using patches.
- Update links to s3 buckets.
- Build the glibc warning into XGBoost instead of using a patch. This might affect others who build from source but it should be fine since we don't want to support old glibc anyway.

I'm using simple string match and replace to generate files from templates. But we can switch to jinja if necessary (when things are more complicated than we can handle, say, many cuda versions).

In the long term, I would like to gather these scripts into a manager project.

**todo**
- [ ] Need to check the wheels with `twine check`.
- [x] `xgboost-cpu` is missing for win64.
- [x] Is it expected that the PyPI tag `macosx_10_15_x86_64.macosx_11_0_x86_64.macosx_12_0_x86_64` is now simply `macosx_10_15_x86_64`?


trivialfis:
@hcho3 Could you please help take a look when you are available? I'm not sure how to proceed with the win64 xgboost-cpu along with the changed macos wheel.

ops/script/pypi_variants.py
```diff
@@ -0,0 +1,76 @@
...
+
+import argparse
+import os
+
+from test_utils import PY_PACKAGE, ROOT
+
+IN_PATH = os.path.join(PY_PACKAGE, "pyproject.toml.in")
+OUT_PATH = os.path.join(PY_PACKAGE, "pyproject.toml")
+
+WHL_CPU = """
```
trivialfis, line 11: We can probably add this to the toml file unconditionally.

hcho3:
>  Is it expected that the PyPI tag macosx_10_15_x86_64.macosx_11_0_x86_64.macosx_12_0_x86_64 is now simply macosx_10_15_x86_64?

Yes, the new tag should be sufficient. I added a test to ensure that the wheel can be installed into latest MacOS.

I also added a CI pipeline to build `xgboost-cpu` on Windows.

hcho3 (approved):

PR #11291, opened 2025-02-27
[dask] Fix with the latest Dask.

trivialfis:
- Use `futures_of` instead of `to_delayed` to break apart a Dask collection.
- Avoid the C API guard that initializes CUDA context to avoid racing against Dask setting `CUDA_VISIBLE_DEVICES`.

Close https://github.com/dmlc/xgboost/issues/11290 .


trivialfis:
I can't reproduce the mgpu error.

hcho3:
Do you need help?

trivialfis:
All tests pass and all demos fail. How is that possible.

TomAugspurger (approved):
The dask changes look good to me.

Apologies for overcomplicating things in https://github.com/dmlc/xgboost/issues/10994#issuecomment-2688071320. It may still be worth making the initial `client.persist(d)` a bit more complicated to ensure that all the parts are persisted on the right workers (if it's causing any performance issues), but that should be done in its own PR.

trivialfis:
> It may still be worth making the initial client.persist(d) a bit more complicated to ensure that all the parts are persisted on the right workers

Thank you for the suggestion. I will follow up with some experiments.

hcho3 (approved):

PR #11292, opened 2025-02-27
SYCL. Fix multiclass for sycl iGPUs

razdoburdin:
The PR https://github.com/dmlc/xgboost/pull/11029 generalizes calculations of multiclass objectives for sycl. But some sycl devices (integrated graphics) don't support fp64 calculations. This fix adds dispatching to gradient calculations. And returns the basic xgboost functional to iGPUs.


trivialfis:
Does iGPU with sycl have a software-simulated f64 implementation? Seems quite restrictive to not have a standard floating point.

napetrov:
This is less of iGPUs but particular generation of ARC systems, with newer Batlemage generation having fp64 support.

trivialfis:
That's great news, we can drop the workarounds for old generation.

PR #11297, opened 2025-02-28
Implement the container for categories.

trivialfis:

- Implement a container type for categories. It can handle string and other numeric types.

No integration yet. I will work on the DMatrix interface once the storage part is in.


trivialfis:
cc @rongou

rongou (approved):

PR #11303, opened 2025-03-03
Store categories from pandas.

trivialfis:
- Add Python glue code to fetch categories from pandas.
- Add a getter to retrieve the categories from `DMatrix`.

Other DMatrix types are not supported yet.


trivialfis:
I need to either find a fix for the hang or disable the test.

trivialfis:
cc @rongou 

tidy error is fixed in https://github.com/dmlc/xgboost/pull/11306

python-package/xgboost/core.py
```diff
@@ -1240,6 +1251,70 @@ def get_quantile_cut(self) -> Tuple[np.ndarray, np.ndarray]:
         assert data.dtype == np.float32
         return indptr, data
 
+    def get_categories(self) -> Optional[Dict[str, "pa.DictionaryArray"]]:
```
trivialfis, line 1254: We might one day generalize this into an `arrow encoder/decoder`. For now, it's just glue code.

src/c_api/c_api.cc
```diff
@@ -718,6 +720,56 @@ XGB_DLL int XGDMatrixGetStrFeatureInfo(DMatrixHandle handle, const char *field,
   API_END();
 }
 
+XGB_DLL int XGBDMatrixGetCategories(DMatrixHandle handle, char const **out) {
```
trivialfis, line 723: The function is hidden due to WIP.

rongou (approved):

PR #11307, opened 2025-03-04
[WIP] Drop the deprecated binary format.

trivialfis:


trivialfis:
@hcho3 Could you please help re-upload the test models to s3, with binary models removed? Also, the RMM build is failing ;-(

chyunsu3:
@trivialfis Should we also remove RDS model files from  xgboost_r_model_compatibility_test.zip ? It's probably using the legacy binary format.

trivialfis:
Please do. @hcho3 .

chyunsu3:
Done

trivialfis:
Thank you!

trivialfis:
@hcho3 Could you please share the latest download link?

trivialfis:
Let me generate some new models using 3.0 instead

trivialfis:
Models generated by the new script using 3.0.2 and 2.1.4: 
[xgboost_model_compatibility_test.zip](https://github.com/user-attachments/files/21205008/xgboost_model_compatibility_test.zip)

chyunsu3:
Are we dropping compatibility for JSON models from XGBoost 1.x?

As for the download link, I uploaded the zip file to the same s3 bucket as before. Were you not be able to access it?

trivialfis:
Let me work on tests for some older models.

trivialfis:
@hcho3 I'm running into 403 forbidden using the old URL: https://productionresultssa16.blob.core.windows.net/actions-results/9dfffb2c-1a7f-4d1b-9f58-01d24823ae27/workflow-job-run-22352302-b8b0-53f6-8324-ec1e5ebbad73/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-07-13T15%3A29%3A15Z&sig=IFsfl5ivJrYF2mMcSfGKK2wHuCwCVhepqX1JnzM2b2w%3D&ske=2025-07-14T03%3A03%3A48Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-07-13T15%3A03%3A48Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-05-05&sp=r&spr=https&sr=b&st=2025-07-13T15%3A19%3A10Z&sv=2025-05-05

trivialfis:
Tests on CRAN are also failing https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-gcc/xgboost-00check.html .

chyunsu3:
My bad, let me change the access setting.

trivialfis:
I uploaded new models. After this PR, Python and R will test with the same set of models. The old models are still in S3, just in case we need them. Also, we can't remove the R test models until the CRAN package is updated.

trivialfis:
@hcho3 Could you please help take a look when you are available?

trivialfis:
I need to let the R check skip the compatibility check when the link is down to make CRAN tests more resilient.

tests/python/generate_models.py
```diff
@@ -1,152 +1,211 @@
...
-    ranker.fit(X, y, g, sample_weight=w)
-    ranker.save_model(skl_bin('ltr'))
-    ranker.save_model(skl_json('ltr'))
-
-
-def write_versions():
-    versions = {'numpy': np.__version__,
-                'xgboost': version}
-    with open(os.path.join(target_dir, 'version'), 'w') as fd:
-        fd.write(str(versions))
```
chyunsu3, line 141: Why are we dropping `write_versions()`? Is it because the file name for each model artifact already contains the version number?

src/cli_main.cc
```diff
@@ -329,29 +329,46 @@ class CLI {
...
   void LoadModel(std::string const& path, Learner* learner) const {
-    if (common::FileExtension(path) == "json") {
-      auto buffer = common::LoadSequentialFile(path);
-      CHECK_GT(buffer.size(), 2);
-      CHECK_EQ(buffer[0], '{');
-      Json in{Json::Load({buffer.data(), buffer.size()})};
+    auto ext = common::FileExtension(path);
+    auto read_file = [&]() {
+      auto str = common::LoadSequentialFile(path);
+      CHECK_GE(str.size(), 3);  // "{}\0"
```
chyunsu3, line 335: Does `common::LoadSequentialFile` return a string with `\0`? I thought the terminating `\0` won't be part of the string?

hcho3 (commented):
Left some questions, but otherwise looks good

tests/python/generate_models.py
trivialfis: Yes. Also the models we test are from multiple versions.(1-3)

src/cli_main.cc
trivialfis: It returns a vector of char as it consumes binary inputs. I removed the requirement for `\0` and added a small test for the JSON parser.

PR #11310, opened 2025-03-05
Test dataset for dask dataframe with str columns.

trivialfis:
- Move the datagen function and add str inputs.

Ref: https://github.com/dmlc/xgboost/issues/11088 .


trivialfis:
cc @rongou .

rongou (approved):

PR #11316, opened 2025-03-07
[doc] Move the model slicing document for R example.

trivialfis:
We might use extensions like https://github.com/executablebooks/sphinx-tabs to have examples for multiple languages in the future.


trivialfis:
cc @david-cortes

Issue #11261, opened 2025-02-17
How to set delta_metric to identity in pairwise objetive


jaguerrerod:
delta_metric have been introduced from version 2 refactoring:
https://github.com/dmlc/xgboost/blob/600be4dbb54ec5cda4cb61a2bb8fc4bce6f3fc58/src/objective/lambdarank_obj.h#L111-L126

respect the way of computing gradient and hessians in 1.7 versions:
https://github.com/dmlc/xgboost/blob/36eb41c960483c8b52b44082663c99e6a0de440a/src/objective/rank_obj.cu#L876

Where is defined delta and which is its default value in master version?

My cases of use don't need any delta function to overweight top elements of each query as I'm optimizing spearman correlation of the whole query.
I would like not use delta function to weight pairs. 
Is possible disable it or set to identity function?

trivialfis:
If you are referring to the rank net loss, then simple `rank:pairwise ` should suffice.

jaguerrerod:
If that is the case, then I can't explain the performance drop I'm seeing in version 3.0.0 compared to 1.7.8.  
My dataset has very little signal (predictions reach a correlation of 0.03). The queries are large (5,000 observations).  
Something significant changed in the refactoring introduced in version 2.0 that consistently reduces performance.  
With 3.0.0, correlation reaches 0.022 after just a few iterations and quickly starts overfitting, dropping below 0.02.  
With 1.7.8, the model's performance on the test dataset improves continuously up to 0.03, requiring many thousands of trees to reach that level.  
Is there any change in sampling, weighting, or the calculation of gradients/hessians introduced in the refactoring that could explain this?  

I'm using this parameters to try to reproduce the behaviour of 1.7.8 in 3.0.0 (full query optimization by rank without normalization and pairing method 'mean' as metric is spearman correlation):

 ```
 booster = 'gbtree',
  objective = 'rank:pairwise',
  tree_method = 'hist',
  device = 'cuda',
  lambdarank_pair_method = 'mean',
  lambdarank_num_pair_per_sample = 200,
  lambdarank_normalization = FALSE,
```

I'll try to upload a portion of the data as a dataset on Kaggle along with code to reproduce the issue.

trivialfis:
`lambdarank_num_pair_per_sample` is too large. Could you please experiment with `1`?

jaguerrerod:
I have trained the same model with the same data using versions 1.7.8 and 3.0.0.  
[...]

PR #11322, opened 2025-03-10
Normalize the sampling-based ltr with num pairs instead of grad.

trivialfis:
https://github.com/dmlc/xgboost/issues/11261

This PR replaces the gradient-based normalization with the `n_pairs`-based normalization for `mean` pair method.

### DIFF
--- a/src/common/ranking_utils.h
+++ b/src/common/ranking_utils.h
@@ -115,6 +115,7 @@ struct LambdaRankParam : public XGBoostParameter<LambdaRankParam> {
   }
 
   [[nodiscard]] bool HasTruncation() const { return lambdarank_pair_method == PairMethod::kTopK; }
+  [[nodiscard]] bool IsMean() const { return lambdarank_pair_method == PairMethod::kMean; }
 
   // Used for evaluation metric and cache initialization, iterate through top-k or the whole list
   [[nodiscard]] auto TopK() const {
--- a/src/common/ranking_utils.h
+++ b/src/common/ranking_utils.h
@@ -180,7 +180,8 @@ class RankingCache {
   HostDeviceVector<std::size_t> y_sorted_idx_cache_;
   // Cached labels sorted by the model
   HostDeviceVector<float> y_ranked_by_model_;
-  // store rounding factor for objective for each group
+  // Rounding factor for CUDA deterministic floating point summation. One rounding factor
+  // for each ranking group.
   linalg::Vector<GradientPair> roundings_;
   // rounding factor for cost
   HostDeviceVector<double> cost_rounding_;
@@ -214,6 +215,9 @@ class RankingCache {
     }
     if (!info.weights_.Empty()) {
       CHECK_EQ(Groups(), info.weights_.Size()) << error::GroupWeight();
+    }
+    if (param_.HasTruncation()) {
+      CHECK_GE(param_.NumPair(), 1);
     }
   }
   [[nodiscard]] std::size_t MaxPositionSize() const {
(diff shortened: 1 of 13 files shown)


trivialfis:
@hcho3 Please help take a look when you are available. Would be great if I can backport it to 3.0

src/common/ranking_utils.h
```diff
@@ -215,6 +216,9 @@ class RankingCache {
     if (!info.weights_.Empty()) {
       CHECK_EQ(Groups(), info.weights_.Size()) << error::GroupWeight();
     }
+    if (param_.HasTruncation()) {
```
hcho3, line 219: How is this condition related to the use of the ``mean`` method? My understanding is that the normalization factor is ``n_pairs`` for the ``mean`` method and the gradient if other method is used?

src/objective/lambdarank_obj.cc
```diff
@@ -225,10 +225,21 @@ class LambdaRankObj : public FitIntercept {
     };
 
     MakePairs(ctx_, iter, p_cache_, g, g_label, g_rank, loop);
-    if (sum_lambda > 0.0 && param_.lambdarank_normalization) {
-      double norm = std::log2(1.0 + sum_lambda) / sum_lambda;
-      std::transform(g_gpair.Values().data(), g_gpair.Values().data() + g_gpair.Size(),
-                     g_gpair.Values().data(), [norm](GradientPair const& g) { return g * norm; });
+    if (param_.lambdarank_normalization) {
+      double norm = 1.0;
+      if (param_.HasTruncation()) {
```
hcho3, line 230: How is this condition related to the use of the ``mean`` method? My understanding is that the normalization factor is ``n_pairs`` for the ``mean`` method and the gradient if other method is used?

From a quick reading of the source code, it seems like `HasTruncation()` is true if the LambdaRank pair method is ``topk``:

https://github.com/dmlc/xgboost/blob/9ab7567de4c6a197756c4726ee63966e38f33232/src/common/ranking_utils.h#L117

Are we pretty sure that there won't be any other pairing method besides ``topk`` and ``mean`` ?

trivialfis: Any suggestions? Use a `switch` instead may help?

hcho3: Can we create a new method that returns ``true`` if the pair method is ``mean``? I think it's natural to have such a method because it requires a special logic for normalization.

trivialfis: Done.

src/objective/lambdarank_obj.cu
```diff
@@ -269,22 +271,35 @@ void CalcGrad(Context const* ctx, MetaInfo const& info, std::shared_ptr<ltr::Ran
...
   thrust::for_each_n(ctx->CUDACtx()->CTP(), thrust::make_counting_iterator(0ul), d_gpair.Size(),
                      [=] XGBOOST_DEVICE(std::size_t i) mutable {
                        auto g = dh::SegmentId(d_gptr, i);
-                       auto sum_lambda = thrust::get<2>(d_max_lambdas[g]);
-                       // Normalization
-                       if (sum_lambda > 0.0 && norm) {
-                         double norm = std::log2(1.0 + sum_lambda) / sum_lambda;
+                       if (need_norm) {
+                         double norm = 1.0;
+                         if (has_truncation) {
```
hcho3, line 283: Can we use `IsMean()` here to be consistent with `lambdarank_obj.cc`?

trivialfis: Yes, seems a bit of waste since we have a compile time parameter there. But it's fine, I will change it.

hcho3: Got it, then comments are sufficient. Let's keep it as it is.

hcho3 (approved):

PR #11323, opened 2025-03-10
[Don't merge] Test CI containers with latest RAPIDS

hcho3:
Ref: https://github.com/dmlc/xgboost-devops/pull/14


hcho3:
I'd like to unblock the upgrade by disabling the failing tests. This is so that we can start testing latest RAPIDS

trivialfis:
hmm, rmm removed its shared library, and we have seen similar dask failures before.

I'm happy to fix them later if we only skip tests for the dev branch of rapids, but keep them for the stable branch.

PR #11345, opened 2025-03-18
Drop support for CUDA sm 50 and sm 60

trivialfis:
```
nvcc warning : Support for offline compilation for architectures prior to '<compute/sm/lto>_75' will be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
```

sm 70 is kept at the moment since it's Turing and used at CI.


hcho3 (changes requested):
Please update https://github.com/dmlc/xgboost/blob/dec7f5896e50191f5b5fb2314c09c2f76304a47b/doc/gpu/index.rst?plain=1#L7 to indicate the requirement for Compute Capability 7.0

trivialfis:
Up the CUDA requirement to 12.8 since we are using it in CI, in addition thrust from older CTK doesn't compile correctly.

hcho3:
We should keep it as 12.0, because users can use the binary wheel.

doc/gpu/index.rst
```diff
@@ -4,7 +4,7 @@ XGBoost GPU Support
 
 This page contains information about GPU algorithms supported in XGBoost.
 
-.. note:: CUDA 12.0, Compute Capability 5.0 required (See `this list <https://en.wikipedia.org/wiki/CUDA#GPUs_supported>`_ to look up compute capability of your GPU card.)
+.. note:: CUDA 12.8, Compute Capability 6.0 required (See `this list <https://developer.nvidia.com/cuda-gpus>`_ to look up compute capability of your GPU card.)
```
hcho3, line 7:
```suggestion
.. note:: CUDA 12.0, Compute Capability 7.0 required (See `this list <https://developer.nvidia.com/cuda-gpus>`_ to look up compute capability of your GPU card.)
```

hcho3 (changes requested):

doc/gpu/index.rst
```diff
@@ -4,7 +4,7 @@ XGBoost GPU Support
 
 This page contains information about GPU algorithms supported in XGBoost.
 
-.. note:: CUDA 12.0, Compute Capability 5.0 required (See `this list <https://en.wikipedia.org/wiki/CUDA#GPUs_supported>`_ to look up compute capability of your GPU card.)
+.. note:: CUDA 12.0, Compute Capability 7.0 required (See `this list <https://developer.nvidia.com/cuda-gpus>`_ to look up compute capability of your GPU card.)
```
trivialfis, line 7: Actually no. Thinking more about it, we can stay with 5.0. It's just not in the default compilation targets. I don't think we need to state the sm in doc as we simply support what CUDA supports. This SM limitation applies to the binary distribution on PyPI and maven, it's not applicable to source build and other distribution methods like conda.

trivialfis:
Closing for now, we should use the `CMAKE_CUDA_ARCHITECTURES=all-major` or something similar instead.

PR #11350, opened 2025-03-20
Simplify `winsock2` workarounds

jakirkham:
On Windows, make sure to `#include <winsock2.h>` before `#include <windows.h>`. Otherwise `#include <windows.h>` will `#include <winsock.h>`, which is only for legacy support.

xref: https://github.com/dmlc/xgboost/pull/10923


hcho3:
It seems like the new workaround didn't quite work out? Can we keep the status quo for now?

trivialfis:
I think it's quite difficult to keep all the includes in the correct order. We have dependencies like dmlc core and grpc etc.

PR #11352, opened 2025-03-20
[jvm-packages] support columnar input for cpu pipeline

wbo4958:
This PR allow xgboost cpu pipeline train on columnar input, so with this PR, xgboost could train on columnar/array/vector.


wbo4958:
Hi @trivialfis, Any thoughts on the failures of CI?

trivialfis (approved):

PR #11351, opened 2025-03-20
Use RMM's pached CCCL

jakirkham:
Make sure to search for RMM if it will be used. This should pick up the patched CCCL from RMM.

If RMM is not being used and this is a CUDA build, search for CCCL explicitly.

PR #11353, opened 2025-03-20
Use RMM's pached CCCL (2.1.x)

jakirkham:
Backports to XGBoost 2.1.x PR: https://github.com/dmlc/xgboost/pull/11351

Make sure to search for RMM if it will be used. This should pick up the patched CCCL from RMM.

If RMM is not being used and this is a CUDA build, search for CCCL explicitly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -224,24 +224,6 @@ if(USE_CUDA)
   add_subdirectory(${PROJECT_SOURCE_DIR}/gputreeshap)
 
   find_package(CUDAToolkit REQUIRED)
-  find_package(CCCL CONFIG)
-  if(NOT CCCL_FOUND)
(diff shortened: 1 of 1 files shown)


jakirkham:
Using this as the basis for a patch in RAPIDS

If we do not plan any additional XGBoost 2.1.x releases, then this can just stay open for now

hcho3:
@jakirkham Do we still need this pull request, now that https://github.com/rapidsai/xgboost-feedstock/pull/85 is merged?

jakirkham:
Let's leave this here for now. Looks like we may still need this a bit longer for XGBoost CUDA 11 builds

trivialfis:
This has been a year now, can we close it?

jakirkham:
Yeah am ok to close given RAPIDS has moved on to XGBoost 3

PR #11354, opened 2025-03-20
Use RMM's pached CCCL (3.0.x)

jakirkham:
Backports to XGBoost 3.0.x PR: https://github.com/dmlc/xgboost/pull/11351

Make sure to search for RMM if it will be used. This should pick up the patched CCCL from RMM.

If RMM is not being used and this is a CUDA build, search for CCCL explicitly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,30 +229,6 @@ if(USE_CUDA)
   endif()
 
   find_package(CUDAToolkit REQUIRED)
-  find_package(CCCL CONFIG)
-  if(CCCL_FOUND)
-    message(STATUS "Standalone CCCL found.")
-  else()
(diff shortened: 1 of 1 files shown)


jakirkham:
cc @trivialfis

trivialfis (approved):
Will merge once the CI is green.

jakirkham:
Thanks Jiaming! 🙏

Sounds good

<hr>

For the one Windows failure am just seeing this:

Screenshot 2025-03-20 at 10 04 55 AM

Is there somewhere we should be looking for logs?

jakirkham:
Most CI runs pass. Seeing a couple CI failures when running the test suite. However that is well after the change here. So think they may not be related

Issue #11355, opened 2025-03-20
[R] Memory leak when storing Booster objects in a list?


fabvass:
Hi there,

Apologies if this was already addressed somewhere else, but I could not find an Issue here  that touches exactly on this.

So, while trying XGBoost 3.0 in R enabled for GPU, I noted that, storing the Booster that is returned by `xgboost()` inside a `list` leads to what seems to be a GPU memory leak. Clearly, this appears to be related to Boosters now being R 'ALTLIST' with external pointers.

To illustrate the problem, consider the following set-up:

```` R
##### Showing GPU memory not being released by XGBoost
library(xgboost)
xgb.set.config(verbosity = 0)
dat <- data.matrix(mtcars)
y <- ifelse(dat[,2] <= 6, 1, 0)
x <- dat[,-2]
S <- 100
outl <- list()
````

Now, if I simply re-train a model on those data in a loop, here's the GPU memory accumulation that I get:

```` R
### With XGBoost 3.0.0 and no gc():

# GPU memory in use before the loop: ~0.58 Gb
t0 <- Sys.time()
for(i in 1:S)
{
    model <- xgboost(x=x, y=as.factor(y),
                     nrounds = 3,
                     objective = "binary:logistic",
                     device = "cuda",
                     tree_method = "hist")
    outl[[i]] <- model
}
t1 <- Sys.time()
t1 - t0 # runs in ~4.35 seconds
yhat <- ifelse(predict(outl[[S]], x)>.5,1,0)
sum(y==yhat)/length(y) # returns ~0.999
# GPU peak during the loop execution: ~4.1 Gb
# GPU memory in use after the loop: ~3.9 Gb
````

Calling garbage collection explicitly, as suggested [here](https://github.com/dmlc/xgboost/issues/6261), does not help:

```` R
### With XGBoost 3.0.0 plus gc():

# GPU memory in use before the loop: ~0.58 Gb
t0 <- Sys.time()
for(i in 1:S)
{
    model <- xgboost(x=x, y=as.factor(y),
                     nrounds = 3,
                     objective = "binary:logistic",
                     device = "cuda",
                     tree_method = "hist")
    outl[[i]] <- model
    rm(model)
    gc()
}
t1 <- Sys.time()
t1 - t0 # runs in ~17.61 seconds
yhat <- ifelse(predict(outl[[S]], x)>.5,1,0)
sum(y==yhat)/length(y) # returns ~0.999
[...]

PR #11357, opened 2025-03-22
[R] Implement Booster reset.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/11355


trivialfis:
cc @david-cortes

R-package/R/xgb.train.R
```diff
@@ -390,7 +390,7 @@ xgb.train <- function(params = xgb.params(), data, nrounds, evals = list(),
     call = match.call(),
     params = params
   )
-
+  bst <- xgb.reset.Booster(bst)
```
david-cortes, line 393: Should also be called at the end of `xgb.cv` when retaining the boosters.

R-package/R/xgb.Booster.R
```diff
@@ -1206,6 +1206,14 @@ xgb.copy.Booster <- function(model) {
   return(.Call(XGDuplicate_R, model))
 }
 
+xgb.reset.Booster <- function(model) {
```
david-cortes, line 1209: Does this function also clean up memory produced during predictions? If so, perhaps it could also get exported.

R-package/src/xgboost_R.cc
```diff
@@ -306,18 +306,24 @@ object, need to be destructed before triggering the R error.
...
-    std::strncpy(cpp_ex_msg, e.what(), 512);    \
-    goto throw_cpp_ex_as_R_err;                 \
-  }                                             \
-  if (false) {                                  \
-    throw_cpp_ex_as_R_err:                      \
-    Rf_error("%s", cpp_ex_msg);                 \
+#define R_API_END()                                  \
+  }                                                  \
+  catch (std::exception & e) {                       \
+    cpp_ex_msg[MsgSize - 1] = 0;                     \
```
david-cortes, line 320: The issue of messages potentially being longer could also be solved by using `%.s` instead of `%s`.

R-package/src/xgboost_R.cc
```diff
@@ -306,18 +306,24 @@ object, need to be destructed before triggering the R error.
 In order to preserve the error message, it gets copied to a temporary
 buffer, and the R error section is reached through a 'goto' statement
 that bypasses usual function control flow. */
-char cpp_ex_msg[512];
+namespace {
+constexpr std::size_t MsgSize = 512;
+char cpp_ex_msg[MsgSize];
```
david-cortes, line 311: Would this be compliant with standard C++17?

R-package/R/xgb.Booster.R
trivialfis: Let's keep it internal for now. I don't want others to abuse it as it has a huge overhead.

R-package/src/xgboost_R.cc
trivialfis: Just need to make sure it's null terminated. This is looks straight forward.

R-package/R/xgb.train.R
trivialfis: Thank you for the catch, will fix it for CV.

R-package/src/xgboost_R.cc
david-cortes: No, but I don't know if every standards-compliant compiler out there would accept this pattern (array size determined by an `std::size_t` constexpr variable).

R-package/R/xgb.train.R
trivialfis: I can't find where the `bst_folds` is returned in the CV function, could you please help give a pointer?

R-package/src/xgboost_R.cc
trivialfis: Yes, it's valid c++ code. It's used very often

R-package/R/xgb.train.R
david-cortes: They are returned through the callback `xgb.cb.cv.predict`:
https://github.com/dmlc/xgboost/blob/0500992cccd1a695fe0030184fb84d0a5f3d703b/R-package/R/callbacks.R#L882

trivialfis: But the `save_models` is false: https://github.com/dmlc/xgboost/blob/0500992cccd1a695fe0030184fb84d0a5f3d703b/R-package/R/xgb.cv.R#L196 .

david-cortes: Yes, that's the default, but the callbacks are all exported so a user can pass the callback directly with non-default arguments.

trivialfis: Thank you for pointing that out. I added a reset to the CV function. It's a bit hacky though, since I'm relying on creating side effects using pointers

david-cortes (approved):

trivialfis:
I have tested the memory usage. Please use nightly build for now.

PR #11367, opened 2025-03-26
Split up grid size of different kernel configurations.

trivialfis:
They are the same in practice, just to be safe.


trivialfis:
cc @rongou .

rongou (approved):

PR #11378, opened 2025-03-28
rename param.hh/cc to hist_param.hh/cc to fix xcode build

gamante91:
When building object library `objxgboost` with xcode, for example in the context of an iOS library/app, xcode is generating all object files in the same place, regardless of the actual path.
This creates collisions when two compilation units have the same name, for example
- `src/tree/hist/param.cc`
- `src/tree/param.cc`

Generate the following object files, notice the appended hashes only for files that have collisions:
```
./build/objxgboost.build/Debug-iphoneos/Objects-normal/arm64/in_memory_handler.o
./build/objxgboost.build/Debug-iphoneos/Objects-normal/arm64/timer.o
./build/objxgboost.build/Debug-iphoneos/Objects-normal/arm64/ellpack_page.o
./build/objxgboost.build/Debug-iphoneos/Objects-normal/arm64/comm_group.o
./build/objxgboost.build/Debug-iphoneos/Objects-normal/arm64/param-20b7432b394539ccbe53282f1bfc13b8.o
./build/objxgboost.build/Debug-iphoneos/Objects-normal/arm64/param-b1b2b65c7e82fe30091c637960712a86.o
```

When linking everything together, the linker cannot find the original file and reports the following error:
```
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: can't open file: build/objxgboost.build/Debug-iphoneos/Objects-normal/arm64/param.o (No such file or directory)
```

This seems to happen only for xcode generator I've found online some issues related to xcode generator not working well with object libraries, for example:
- https://gitlab.kitware.com/cmake/cmake/-/issues/22415
- https://github.com/tensorflow/tensorflow/issues/15369

The immediate fix is to just rename files that have the same name in order to avoid collisions when creating object files
- `src/tree/hist/param.cc` → `src/tree/hist/hist_param.cc`
- `src/tree/hist/param.h` → `src/tree/hist/hist_param.h`


gamante91:
> Could you please share the error? We have CI tests on macos for both x86 and arm using apple clang, thus far no related error is observed.
@trivialfis error shared, this only happens for xcode generator and object libraries (`objxgboost` in this case)

trivialfis:
Need https://github.com/dmlc/xgboost/pull/11382 to unblock the CI.

trivialfis (approved):

PR #11389, opened 2025-04-07
Fix mypy errors.

trivialfis:


hcho3 (approved):

trivialfis:
@hcho3 In the last few months, it has not been easy to get the Windows worker. The weird thing is that if I cancel the test and restart it, the chance of getting the worker is significantly higher. If we simply wait for it without restarting, it will time out.

PR #11390, opened 2025-04-07
Optimization of data initialization for large sparce datasets

razdoburdin:
This PR speed-ups data initialization for large sparce datasets being executed on multi-core CPUs by parallelizing the execution. 
For `bosch` dataset this PR improve fitting time on 1.3x for 2x56cores system.

To avoid the race condition, I have also switched from using bitfields as missing flag to uint8_t.


trivialfis:
Note to myself:

>  I have also switched from using bitfields as missing flag to uint8_t.

That is increasing memory usage.

trivialfis:
Apologies, coming back from a trip. Will look into the optimization.

src/common/column_matrix.h
```diff
@@ -366,19 +371,69 @@ class ColumnMatrix {
...
-        auto line = batch.GetLine(rid);
-        for (size_t i = 0; i < line.Size(); ++i) {
-          auto coo = line.GetElement(i);
-          if (is_valid(coo)) {
-            auto fid = coo.column_idx;
-            const uint32_t bin_id = row_index[k];
-            SetBinSparse(bin_id, rid + base_rowid, fid, local_index);
-            ++k;
+
+      dmlc::OMPException exc;
```
trivialfis, line 375: Could you please add some comments to provide a high-level summary of the steps achieved in this section of the code?

trivialfis (commented):
Could you please provide some data on the effect of memory usage where there are semi-dense columns?

src/data/gradient_index.h
```diff
@@ -233,7 +233,7 @@ class GHistIndexMatrix {
   void PushAdapterBatchColumns(Context const* ctx, Batch const& batch, float missing,
                                size_t rbegin);
 
-  void ResizeIndex(const size_t n_index, const bool isDense);
+  void ResizeIndex(const size_t n_index, const bool isDense, int n_threads = 1);
```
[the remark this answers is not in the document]
razdoburdin: I fixed the code, no default value now.

src/common/column_matrix.h
```diff
@@ -214,15 +206,28 @@ class ColumnMatrix {
     }
   }
 
+  template <typename ColumnBinT, typename BinT, typename RIdx>
+  void SetBinSparse(BinT bin_id, RIdx rid, bst_feature_t fid, ColumnBinT* local_index, size_t nnz) {
+    if (type_[fid] == kDenseColumn) {
+      ColumnBinT* begin = &local_index[feature_offsets_[fid]];
+      begin[rid] = bin_id - index_base_[fid];
```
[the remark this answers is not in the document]
razdoburdin: I moved the first line outside branches. The second one differs.

src/common/column_matrix.h
```diff
@@ -214,15 +206,28 @@ class ColumnMatrix {
...
  public:
   // get number of features
   [[nodiscard]] bst_feature_t GetNumFeature() const {
     return static_cast<bst_feature_t>(type_.size());
   }
 
   ColumnMatrix() = default;
-  ColumnMatrix(GHistIndexMatrix const& gmat, double sparse_threshold) {
-    this->InitStorage(gmat, sparse_threshold);
+  ColumnMatrix(GHistIndexMatrix const& gmat, double sparse_threshold, int n_threads = 1) {
```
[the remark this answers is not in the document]
razdoburdin: fixed

src/common/column_matrix.h
razdoburdin: I added some comments in the code, to make it more clear

trivialfis:
Got the following results from synthesized dense data, memory usage measured by `cgmemtime`.
```
* master

[7]     Train-rmse:33.39066
Qdm train (sec) ended in: 25.732778310775757 seconds.
Trained for 8 iterations.
{'BenchIter': {'GetTrain (sec)': 27.98164939880371}, 'Qdm': {'Train-DMatrix-Iter (sec)': 91.10861659049988, 'train (sec)': 25.732778310775757}}

user: 1809.226 s
sys:   28.329 s
wall: 119.860 s
child_RSS_high:   37892596 KiB
group_mem_high:   37677792 KiB

* opt pr

[7]     Train-rmse:33.39066
Qdm train (sec) ended in: 25.054997444152832 seconds.
Trained for 8 iterations.
{'BenchIter': {'GetTrain (sec)': 28.075414180755615}, 'Qdm': {'Train-DMatrix-Iter (sec)': 93.2668731212616, 'train (sec)': 25.054997444152832}}

user: 1807.715 s
sys:   31.093 s
wall: 121.895 s
child_RSS_high:   45232596 KiB
group_mem_high:   45032396 KiB
```

That's a 20 percent increase `(45032396 - 37677792) / 37677792)` in memory usage for dense data. Are you sure you want this PR to go in? Asking since the memory usage has been a pain point for XGBoost for a very long time. We receive issues mostly about memory usage instead of computation time, so we care about it a lot.

razdoburdin:
I was able to return to bitfield representation for missing indicator without loosing thread-safe access. It requires quite careful data management, but allow to combine benefits of parallelization and low memory consumption. Some additional memory should be allocated in this case for data alignment, but it is less than 4 bytes per feature in worse case.

trivialfis (commented):
Apologies for the slow response, will do some tests myself. Please see inlined comments.

src/common/column_matrix.h
```diff
@@ -195,34 +236,42 @@ class ColumnMatrix {
     }
   };
 
-  void InitStorage(GHistIndexMatrix const& gmat, double sparse_threshold);
+  void InitStorage(GHistIndexMatrix const& gmat, double sparse_threshold, int n_threads);
 
   template <typename ColumnBinT, typename BinT, typename RIdx>
   void SetBinSparse(BinT bin_id, RIdx rid, bst_feature_t fid, ColumnBinT* local_index) {
```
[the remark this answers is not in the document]
razdoburdin: The original SetBinSparse is also [used](https://github.com/dmlc/xgboost/blob/61b387834e24ef603c07626ed040d30d2aa16c0f/src/common/column_matrix.h#L530)

src/common/column_matrix.h
```diff
@@ -366,19 +423,100 @@ class ColumnMatrix {
...
+       * each underlying word of the bitfiled should be processed by a single thread.
+       * So we need to align the row-blocks.
+       */
+      block_size = DivRoundUp(block_size, MissingIndicator::BitFieldT::kValueSize) *
+                   MissingIndicator::BitFieldT::kValueSize;
+      /*
+       * If base_rowid > 0 we need to shift the blocks boundaries.
+       * Otherwise the two threads may operate with the single word of bitfield.
+       */
+      size_t shift = MissingIndicator::BitFieldT::kValueSize -
```
trivialfis: I don't quite understand how this shifting works. Could you please help clarify it? For starters, this should represent the number of samples each thread needs to shift. How is it related to the bit field value size? How is it related to the module? Why set it to 0 when it equals the value size?

razdoburdin: > I don't quite understand how this shifting works. Could you please help clarify it? For starters, this should represent the number of samples each thread needs to shift. How is it related to the bit field value size? How is it related to the module? Why set it to 0 when it equals the value size?

if `base_rowid > 0`, we add few samples to thread 0 in a way to make sure each next thread will start from a word boundary.

For instance, if the first batch had 35 rows, `base_rowid` for the second batch would be 35. The 36th row's missing flag is at bit 3 of the second word in our bitfield (word[1]). If we simply divided the work evenly, Thread 0 might process rows 35-66 and Thread 1 rows 67-98.
This would cause both threads to access and modify word[1] and word[2],  leading to a race condition.

`base_rowid % MissingIndicator::BitFieldT::kValueSize` calculates the starting position of our new data within a 32-bit word.
For `base_rowid` = 35, the result is 3. This means we are 3 bits into the word. 32 - 3 gives 29. This shift value represents the number of rows (or bits) the first thread must process to get to the next clean word boundary. After processing these 29 rows, the next row to process will be at the start of a new 32-bit word, allowing subsequent blocks to be aligned.

If `base_rowid` is perfectly divisible by 32, the calculation becomes 32.
A shift of 32 is unnecessary because we are already at a perfect boundary. So we set shift = 0.

A have modified the comment to make this logic more clear.

trivialfis: You are talking about the external memory in XGBoost. Is the bitfield shared across multiple batches of data? Otherwise, the

> the 36th row's missing flag is at bit 3

should not be true since there should be a different column matrix for each batch.

razdoburdin: > You are talking about the external memory in XGBoost. Is the bitfield shared across multiple batches of data? Otherwise, the
> 
> > the 36th row's missing flag is at bit 3
> 
> should not be true since there should be a different column matrix for each batch.

In the original code, the bitfield is allocated for the total amount of elements, but each batch uses it's own part. I didn't touch this part in my PR.

trivialfis: Let me do some digging tomorrow.

trivialfis: I did some small tests, and the size of each bitfield is the same as the input batch size. The column matrix is not shared between batches for external memory.

It's a bit confusing, the batch-based input is used by both the `QuantileDMatrix` and the `ExtMemQuantileDMatrix`. The former is in-core and the latter is out-of-core. In-core training always has `base_rowid` equal to 0. For the QDM, it will push for multiple batches into a single column matrix, and for the ExtMemQdm, it pushes only a single batch into each column matrix.

trivialfis: For the two different cases, see:
- https://github.com/dmlc/xgboost/blob/master/demo/guide-python/quantile_data_iterator.py (change it to CPU)
- https://github.com/dmlc/xgboost/blob/master/demo/guide-python/external_memory.py

razdoburdin: Got it.
I have made some more tests. It looks like the value of `base_rowid` one operates with in the `SetIndexMixedColumns `can differs from the `base_rowid` in `GHistIndexMatrix`. 

In case of external memory the function call goes the following:
`ExtGradientIndexPageSource::Fetch()` calls `PushAdapterBatchColumns` with rbegin = 0 https://github.com/dmlc/xgboost/blob/e4406daa58a15c4a08521e630a822bb70f1c366d/src/data/gradient_index_page_source.cc#L64

`GHistIndexMatrix::PushAdapterBatchColumns` calls `PushBatch `with the same rbegin = 0 https://github.com/dmlc/xgboost/blob/e4406daa58a15c4a08521e630a822bb70f1c366d/src/data/gradient_index.cc#L122

`ColumnMatrix::PushBatch` calls SetIndexMixedColumns with base_rowid = rbegin = 0 https://github.com/dmlc/xgboost/blob/e4406daa58a15c4a08521e630a822bb70f1c366d/src/common/column_matrix.h#L287

So, the value of `base_rowid ` inside `SetIndexMixedColumns` is 0 in case of external memory even if `base_rowid` in gmat is non-zero, and the shift for bitfield is also zero in this case.

trivialfis: Let me do more tests, probably need some cleanup/comments.

trivialfis: Can we postpone this PR a little bit? It's an optimization for data initialization, and I find it quite difficult to understand. I will merge the inference PR.

razdoburdin: We can postpone it, but I think it shouldn't be postponed for a long time. Overwise, resolving of merge conflicts will be extremely tricky.

src/common/column_matrix.h
```diff
@@ -366,19 +423,107 @@ class ColumnMatrix {
...
-        auto line = batch.GetLine(rid);
-        for (size_t i = 0; i < line.Size(); ++i) {
-          auto coo = line.GetElement(i);
-          if (is_valid(coo)) {
-            auto fid = coo.column_idx;
-            const uint32_t bin_id = row_index[k];
-            SetBinSparse(bin_id, rid + base_rowid, fid, local_index);
-            ++k;
+
+      // Parallel sparse batch processing
```
Vika-F: I think adding the description of the whole idea might be beneficial.
```suggestion
      // Parallel sparse batch processing
      //
      // This section processes the input batch in parallel across multiple threads.
      //
      // rows [base_rowid, batch_size + base_rowid) are divided into `n_threads` blocks.
      // Threads process the assigned blocks of rows in parallel.
      //
      // As the indicator of the missing elements is stored in a bitfield, to ensure thread-safe
      // access to the bitfield, each underlying word of the bitfield
      // of size MissingIndicator::BitFieldT::kValueSize should be processed by
      // a single thread. Therefore, we align the row-blocks accordingly.
      //
      // block_size - size of the row block assigned to each thread, divisible by the kValueSize.
      // shift - adjustment applied to the starting row (base_rowid) of each thread (except for the 0-th thread)
      //         to ensure that each thread starts processing from a word boundary in the bitfield.
      //
      // 0-th thread processes rows [base_rowid, base_rowid + shift + block_size)
      // 1-st thread processes rows [base_rowid + shift +     block_size, base_rowid + shift + 2 * block_size)
      // 2-nd thread processes rows [base_rowid + shift + 2 * block_size, base_rowid + shift + 3 * block_size)
      // ...
      // (n_threads-1)-th thread processes rows [base_rowid + shift + (n_threads-1) * block_size, base_rowid + batch_size)
      //
      // Computations are done in two passes:
      // 1) Counting non-zero elements per feature per thread to determine their offsets for the next step.
      //    a) Counting non-zero elements per feature per thread.
      //    b) Aggregating counts to determine offsets.
      // 2) Placing elements into the sparse structure using calculated offsets of non-zero elements.
```

src/common/column_matrix.h
```diff
@@ -366,19 +423,107 @@ class ColumnMatrix {
...
+      dmlc::OMPException exc;
+      std::vector<size_t> n_elements((n_threads + 1) * n_features, 0);
+      std::vector<size_t> k_offsets(n_threads + 1, 0);
```
Vika-F: Description of the purpose of those arrays:
```suggestion
            std::vector<size_t> n_elements((n_threads + 1) * n_features, 0);  // number of non-zero elements per feature per thread
                                                                        // n_elements[tid * n_features + fid] =
                                                                        //   number of non-zero elements of feature fid processed by thread tid
                                                                        // n_elements[n_threads * n_features + fid] =
                                                                        //   total number of non-zero elements of feature fid
      std::vector<size_t> k_offsets(n_threads + 1, 0);        // offsets of non-zero elements for each thread
                                                              // k_offsets[0] = 0;
                                                              // k_offsets[tid] - starting offset of non-zero elements processed by thread tid
```

src/common/column_matrix.h
```diff
@@ -366,19 +423,107 @@ class ColumnMatrix {
...
+      size_t block_size = DivRoundUp(batch_size, n_threads);
+
+      /*
+       * We use bitfield as a missing indicator. To ensure thread safe access to the bitfield
+       * each underlying word of the bitfiled should be processed by a single thread.
+       * So we need to align the row-blocks.
+       */
+      block_size = DivRoundUp(block_size, MissingIndicator::BitFieldT::kValueSize) *
+                   MissingIndicator::BitFieldT::kValueSize;
```
Vika-F: Improving the readability a bit:
```suggestion
      const auto word32size = MissingIndicator::BitFieldT::kValueSize;

      size_t block_size = DivRoundUp(batch_size, n_threads);   // preliminary block size
      // align block_size to be multiple of kValueSize, so that each thread processes full words in the bitfield
      block_size = DivRoundUp(block_size, word32size) * word32size;
```

src/common/column_matrix.h
```diff
@@ -366,19 +423,107 @@ class ColumnMatrix {
...
+      /*
+       * To prevent race conditions on the bitfield, we ensure each thread operates on
+       * distinct 32-bit words. If a data batch (starting at `base_rowid`) doesn't align
+       * with a word boundary, this `shift` is calculated. It represents the number of rows
+       * the first thread must process to reach the next aligned word. This guarantees all
+       * subsequent thread workloads start on a clean boundary, making parallel updates safe.
+       */
+      size_t shift = MissingIndicator::BitFieldT::kValueSize -
```
Vika-F:
```suggestion
      // calculate shift to align thread starting positions to word boundaries in the bitfield
      const size_t shift = base_rowid - (base_rowid / word32size) * word32size;
```

src/common/column_matrix.h
```diff
@@ -366,19 +423,107 @@ class ColumnMatrix {
...
+                     (base_rowid % MissingIndicator::BitFieldT::kValueSize);
+      /*
+       * If `base_rowid` is already on a word boundary, the calculation results in
+       * `kValueSize`. In this case, no shift is needed.
+       */
+      if (shift == MissingIndicator::BitFieldT::kValueSize) {
+        shift = 0;
+      }
```
Vika-F: This if is not needed in case of the change in `shift` calculation.

src/common/column_matrix.h
```diff
@@ -366,19 +423,107 @@ class ColumnMatrix {
...
+                }
+                k_offsets[tid + 1] += 1;
+              }
+            }
           }
+        });
+      }
+      exc.Rethrow();
+
+      // Parallel feature processing to aggregate counts & calculate offsets.
```
Vika-F, line 524:
```suggestion
      // Parallel feature processing to aggregate counts & calculate offsets.
      //
      // Compute the number of non-zero elements per feature per thread (n_elements)
      // and the offsets to non-zero elements per thread (k_offsets).
      //
      // The final values of n_elements and k_offsets will be calculated after counting the number of non-zero elements per thread.
```

src/common/column_matrix.h
```diff
@@ -366,19 +423,107 @@ class ColumnMatrix {
...
+            n_elements[tid * n_features + fid];
         }
+        num_nonzeros_[fid] = n_elements[n_threads * n_features + fid];
+      });
+      std::partial_sum(k_offsets.cbegin(), k_offsets.cend(), k_offsets.begin());
+
+      // Parallel row processing to place data using offsets into sparse structure.
+      #pragma omp parallel num_threads(n_threads)
+      {
+        std::vector<size_t> nnz_offsets(n_features, 0);
```
Vika-F, line 545:
```suggestion
                std::vector<size_t> nnz_offsets(n_features, 0);   // offsets of non-zero elements per feature for the current thread
                                                          // nnz_offsets[fid] =
                                                          //   number of non-zero elements of feature fid already processed by this thread
```

src/common/column_matrix.h
```diff
@@ -366,19 +423,107 @@ class ColumnMatrix {
...
+
+          size_t k = 0;
+          for (size_t rid = begin; rid < end; ++rid) {
+            const auto& line = batch.GetLine(rid);
+            for (size_t i = 0; i < line.Size(); ++i) {
+              auto coo = line.GetElement(i);
+              if (is_valid(coo)) {
+                auto fid = coo.column_idx;
+                const uint32_t bin_id = row_index[k_offsets[tid] + k];
+                size_t nnz = n_elements[tid * n_features + fid] + nnz_offsets[fid];
```
Vika-F, line 567:
```suggestion
                const uint32_t bin_id = row_index[k_offsets[tid] + k];  // get the correct offset for this thread
                size_t nnz = n_elements[tid * n_features + fid] + nnz_offsets[fid]; // calculate the correct nnz for this feature and this thread
```

src/common/column_matrix.h
razdoburdin: for example, with base_rowid = 42 old version gives shift = 22, your update gives shift = 10

Vika-F: You are right.
It can be something like this, but I am not sure that it makes the code more understandable.
```
const size_t base_rowid_mod32 = base_rowid % word32size;
const size_t shift = word32size * bool(base_rowid_mod32) - base_rowid_mod32;
```
At least shift is const in this case.

Vika-F:
@trivialfis Some comments describing the logic in the parallel part of the code were added. Can you please take a look one more time at this PR?

trivialfis:
Thank you for the ping, I will go through it (promise).

I enjoy your optimization PRs very much. But is it possible to create some abstractions that are more friendly to ordinary data scientists?

trivialfis:
I'm trying to split up the PR to make it a little more digestible for myself. The first one is here: https://github.com/dmlc/xgboost/pull/11974. I have updated the author field to reflect the origin of the changes.

PR #11396, opened 2025-04-08
Share Python updater tests between CPU and GPU.

trivialfis:
Prepare to remove the `gpu_hist` parameter.

- I skipped the dask restart test to avoid timeout (hang). The test is too flaky.
- Move some tests into the package and replace the `gpu_hist` parameter with the `device` parameter.

We will do some refactorings before we can eventually remove this parameter as it's ubiquitous in the codebase.


trivialfis:
wip: skipped the dask restart test to avoid timeout (hang).

trivialfis:
cc @rongou

rongou (approved):

PR #11391, opened 2025-04-07
Allow fallback for CUDA virtual memory.

trivialfis:
- Parse `nvidia-smi` query to check system driver version.

todo:
- [x] Test.

Close https://github.com/dmlc/xgboost/issues/11397 .

PR #11398, opened 2025-04-09
Make CUDA version check compatible with future versions.

trivialfis:
Extracted from https://github.com/dmlc/xgboost/pull/11391 .


trivialfis:
cc @rongou .

PR #11405, opened 2025-04-11
[dask] Disable the flaky `worker_left` test.

trivialfis:
Error resilience support relies on the timeout option in rabit. Direct handling is not well supported. To avoid hanging in shutdown, we can change the communication protocol or wait for Dask to implement something similar to the Spark barrier. For now, we will disable the test.

https://github.com/dmlc/xgboost/actions/runs/14381427963/job/40337545246

### DIFF
--- a/tests/test_distributed/test_with_dask/test_with_dask.py
+++ b/tests/test_distributed/test_with_dask/test_with_dask.py
@@ -2191,7 +2191,7 @@ class TestDaskCallbacks:
     clean_kwargs={"processes": False, "threads": False},
     allow_unclosed=True,
 )
-@pytest.mark.skip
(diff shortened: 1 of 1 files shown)


tests/test_distributed/test_with_dask/test_with_dask.py
```diff
@@ -2191,6 +2191,7 @@ def test_callback(self, client: "Client") -> None:
     clean_kwargs={"processes": False, "threads": False},
     allow_unclosed=True,
 )
+@pytest.mark.skip
```
hcho3, line 2194:
```suggestion
@pytest.mark.skip(reason="dmlc/xgboost#11405: test_worker_left is flaky")
```

hcho3 (changes requested):

PR #11410, opened 2025-04-16
Use Sphinx Tabs to separate Python/R code in tutorial for advanced objectives

david-cortes:
Just discovered this non-built-in plugin for Sphinx which allows putting code snippets into tabs to separate different languages.

I'm using it here to separate the Python and R code snippets in the tutorial for advanced objectives only, but would be ideal to make multi-language versions of more tutorials in the future.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -239,6 +239,7 @@ extensions = [
     "sphinx.ext.intersphinx",
     "sphinx_gallery.gen_gallery",
     "sphinx_issues",
+    "sphinx_tabs.tabs",
     "breathe",
     "myst_parser",
 ]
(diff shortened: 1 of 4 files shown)


trivialfis:
Thank you for looking into this, but we might run into issues with pdf (latex) output according to https://github.com/executablebooks/sphinx-tabs/issues/150 .

david-cortes:
Are online docs built in PDF form? I couldn't find any PDF link in ReadTheDocs, although I see that the makefile has options for it.

david-cortes:
And actually, it looks like it renders correctly in latextpdf - this is how the local build ends up looking for me:
image

trivialfis:
> Are online docs built in PDF form? I couldn't find any PDF link in ReadTheDocs,

Go to https://xgboost.readthedocs.io/ , click the triangle in the black box at the bottom right corner, then you will see a `Downloads` section in an expanded box.

Thank you for confirming the build, will look into the PR.

david-cortes:
Added also for the slicing tutorial.

PR #11420, opened 2025-04-20
Make all tutorials multi-language by adding R code examples

### LINKED ISSUE #11413: Make all tutorials multi-language
ref https://github.com/dmlc/xgboost/pull/11410

After the PR above gets merged, we'll have a mechanism to show language tabs throughout the online docs, but it would only be used in 2 out of 25 tutorials. This could be a nice way to make all the current tutorials multi-language, by having a tab for 'R' next to the tab for 'Python' - example:

Image

I think most of the tutorials would be straightforward to write as R code (save for ones that don't apply, like distributed mode), but it's quite a lot of work. Leaving the idea here in case someone wants to work on porting at least some tutorials (CC @mayer79 @jameslamb @dfsnow).

Keerthi421:
PR :Make all tutorials multi-language by adding R code examples
Description :
This PR addresses issue #11413 by adding R language examples alongside the existing Python examples in XGBoost tutorials. This makes the documentation more accessible to users who prefer working in R rather than Python.
Implementation
Added language tabs using the sphinx-panels extension
Created equivalent R code examples matching the functionality of existing Python examples
Ensured consistent code style between Python and R implementations
Tested all R code examples to verify functionality
The implementation follows the plan outlined in PLAN.md, providing a systematic approach to convert all tutorials to support multiple languages.

Example
Before this PR, tutorials only provided Python examples. After this PR, users can switch between Python and R examples using the tabbed interface.
Related Issues
Fixes #11413

### DIFF
--- a/MULTILANGUAGE_README.md
+++ b/MULTILANGUAGE_README.md
@@ -1,0 +1,62 @@
+# XGBoost Multi-Language Tutorials
+
+This project aims to make all XGBoost tutorials multi-language, by adding R code examples alongside the existing Python examples. This will allow users to choose their preferred language when reading the documentation.
+
+## Background
+
+XGBoost is a popular machine learning library that supports multiple programming languages, including Python, R, Java, and more. However, most of the current tutorials only provide examples in Python. This project addresses [Issue #11413](https://github.com/dmlc/xgboost/issues/11413) by making all tutorials multi-language.
+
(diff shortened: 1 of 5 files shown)


trivialfis:
While the issue is a good place to use LLM, but at least one should verify the intent of the issue before submitting a PR.

PR #11426, opened 2025-04-23
[EM] Add file read stream.

trivialfis:
Planning to experiment with combining mmap and normal file read.

- Add a new stream backend that directly reads files in addition to the existing `mmap`.
- Some refactoring to the `nvtx` wrappers.

### DIFF
--- a/src/common/cuda_rt_utils.h
+++ b/src/common/cuda_rt_utils.h
@@ -5,10 +5,6 @@
 #include <cstddef>  // for size_t
 #include <cstdint>  // for int32_t
 
-#if defined(XGBOOST_USE_NVTX)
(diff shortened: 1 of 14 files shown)


trivialfis:
@hcho3 Could you please help take a look at the `cibuildwheel` requirement update when you are available? It seems we need to either upgrade the Python version for all CI pipelines, or keep two versions of Python with one for XGBoost (3.10) and one for cibuildwheel (3.11).

trivialfis:
NVM, the `cibuildwheel` warning is unrelated

trivialfis:
cc @rongou

rongou (approved):

Issue #11427, opened 2025-04-23
Avoid nan/inf in JSON output.


trivialfis:
These are not in the standard JSON specification.

PR #11428, opened 2025-04-23
Replace nan with denormal floating point.

trivialfis:
- Remove nan as init value for tree models.
- Avoid zero division in linear models.

Close https://github.com/dmlc/xgboost/issues/11427 .

### DIFF
--- a/src/linear/coordinate_common.h
+++ b/src/linear/coordinate_common.h
@@ -65,10 +65,11 @@ inline double CoordinateDelta(double sum_grad, double sum_hess, double w,
  * \return  The weight update.
  */
 inline double CoordinateDeltaBias(double sum_grad, double sum_hess) {
-  if (std::fpclassify(sum_hess) == FP_ZERO) {
-    return 0;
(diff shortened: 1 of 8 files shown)


trivialfis:
@hcho3 Please help take a look when you are available.

src/linear/coordinate_common.h
```diff
@@ -64,6 +65,9 @@ inline double CoordinateDelta(double sum_grad, double sum_hess, double w,
  * \return  The weight update.
  */
 inline double CoordinateDeltaBias(double sum_grad, double sum_hess) {
+  if (std::fpclassify(sum_hess) == FP_ZERO) {
+    return 0;
+  }
```
hcho3, line 70: The division may still be invalid even when `sum_hess` is a small nonzero value, due to overflow. We should just compute `-sum_grad / sum_hess` and then check if the resulting value is infinity, using `std::isinf`.

hcho3 (changes requested):

hcho3 (approved):

PR #11443, opened 2025-05-04
Specialize for GPU dense histogram.

trivialfis:
Branching is quite expensive on Hopper. This PR tries to reduce the amount of runtime branching.


trivialfis:
cc @rongou

rongou (approved):

PR #11446, opened 2025-05-07
[EM] Drop support for the device cache.

trivialfis:
We will move toward a more flexible cache model in which the DMatrix can split each page into host cache and device cache instead of storing the entire page in the device.


trivialfis:
cc @rongou

rongou (approved):

PR #11448, opened 2025-05-10
[WIP] Avoid R API calls within OpenMP region

hcho3:
https://github.com/dmlc/xgboost/issues/11431#issuecomment-2868004921

### DIFF
--- a/R-package/src/xgboost_R.cc
+++ b/R-package/src/xgboost_R.cc
@@ -247,15 +247,18 @@ XGB_DLL SEXP XGDMatrixSetInfo_R(SEXP handle, SEXP field, SEXP array) {
   auto ctx = DMatrixCtx(R_ExternalPtrAddr(handle));
   if (!strcmp("group", name)) {
     std::vector<unsigned> vec(len);
+    const int* array_ptr = INTEGER_RO(array);
     xgboost::common::ParallelFor(len, ctx->Threads(), [&](xgboost::omp_ulong i) {
-      vec[i] = static_cast<unsigned>(INTEGER(array)[i]);
+      vec[i] = static_cast<unsigned>(array_ptr[i]);
     });
     CHECK_CALL(
         XGDMatrixSetUIntInfo(R_ExternalPtrAddr(handle), CHAR(Rf_asChar(field)), BeginPtr(vec), len));
   } else {
     std::vector<float> vec(len);
-    xgboost::common::ParallelFor(len, ctx->Threads(),
-                                 [&](xgboost::omp_ulong i) { vec[i] = REAL(array)[i]; });
+    const double* array_ptr = REAL_RO(array);
+    xgboost::common::ParallelFor(len, ctx->Threads(), [&](xgboost::omp_ulong i) {
+      vec[i] = static_cast<float>(array_ptr[i]);
+    });
     CHECK_CALL(
         XGDMatrixSetFloatInfo(R_ExternalPtrAddr(handle), CHAR(Rf_asChar(field)), BeginPtr(vec), len));
   }
@@ -397,9 +400,11 @@ XGB_DLL SEXP XGBoosterBoostOneIter_R(SEXP handle, SEXP dtrain, SEXP grad, SEXP h
   int len = Rf_xlength(grad);
   std::vector<float> tgrad(len), thess(len);
   auto ctx = BoosterCtx(R_ExternalPtrAddr(handle));
+  const double* grad_ptr = REAL_RO(grad);
+  const double* hess_ptr = REAL_RO(hess);
   xgboost::common::ParallelFor(len, ctx->Threads(), [&](xgboost::omp_ulong j) {
-    tgrad[j] = REAL(grad)[j];
-    thess[j] = REAL(hess)[j];
+    tgrad[j] = static_cast<float>(grad_ptr[j]);
+    thess[j] = static_cast<float>(hess_ptr[j]);
   });
   CHECK_CALL(XGBoosterBoostOneIter(R_ExternalPtrAddr(handle),
                                  R_ExternalPtrAddr(dtrain),


david-cortes:
Would it perhaps be better to backport the PR that fixed the same issue in the master branch?
https://github.com/dmlc/xgboost/pull/9823

hcho3:
Is that PR after the massive rewrite of the R package? If so, it probably won't apply to the 1.7 branch

david-cortes:
> Is that PR after the massive rewrite of the R package? If so, it probably won't apply to the 1.7 branch

I think it was before the changes to .R files, but I don't remember. The thing is that it also fixed other issues along the way.

david-cortes:
Actually looking at the PR: might not be easy to backport, let's better stick with this one.

jvm-packages/xgboost4j/pom.xml
```diff
@@ -6,10 +6,10 @@
     <parent>
         <groupId>ml.dmlc</groupId>
         <artifactId>xgboost-jvm_2.12</artifactId>
-        <version>1.7.11</version>
+        <version>1.7.12</version>
```
david-cortes, line 9: Since this is R-only, I guess changes here are not needed.

R-package/tests/testthat/test_helpers.R
```diff
@@ -288,6 +288,72 @@ test_that("xgb.model.dt.tree throws error for gblinear", {
...
+      )
+    ][
+      order(Gain, decreasing = TRUE)
+    ]
+    imp
+  }
+  expect_equal(importance_from_dump(), importance, tolerance = 1e-6)
+
+  ## decision stump
+  m <- xgboost::xgboost(
```
david-cortes, line 346: I'm not sure if it'd be correct to use `xgboost::` in tests. It's certainly not needed as the whole namespace is imported.

R-package/tests/testthat/test_feature_weights.R
```diff
@@ -0,0 +1,33 @@
+library(xgboost)
+
+context("feature weights")
+
+n_threads <- 2
+
+test_that("training with feature weights works", {
```
david-cortes, line 7: I'd say better not to add extra tests for this short-lived release, since they increase the chances of rejections and extra fixes being needed.

hcho3: This is merely restoring the tests removed in the previous commit.

trivialfis:
Thank you for working on this! Please let me know if there's anything I can help. (testing, submitting packages, etc).

hcho3:
@trivialfis Can you help with submitting the package with the fix?

trivialfis:
> Can you help with submitting the package with the fix?

Is this ready? (still marked as draft)

Since 1.7.11 never went through, do you think we need to bump to 1.7.12?

hcho3:
> Lastly, did you manage to tackle issues with the changed openblas? There seems to be floating point accuracy related test failures.

No, I was never able to reproduce the issue.

david-cortes:
Per CRAN recommendations:
https://cran.r-project.org/web/packages/policies.html
> Increasing the version number at each submission reduces confusion so is preferred even when a previous submission was not accepted.


> There seems to be floating point accuracy related test failures.

I don't think that's the case. The failures looked like completely different results than would be expected, not small numerical inaccuracies.

> No, I was never able to reproduce the issue.

Perhaps one way could be by building R with flag `--enable-strict-barrier`, like in the email:
https://stat.ethz.ch/pipermail/r-package-devel/2025q2/011652.html

hcho3:
> Perhaps one way could be by building R with flag --enable-strict-barrier, like in the email:

I already tried it and could not reproduce the other test failures.

@trivialfis Have the maintainer gotten back to you about reproducing the OpenBLAS environment? If not, let's ask for permission to remove the failing tests

hcho3:
> How many test failures did you reproduce using the --enable-strict-barrier flag?

Only the tests that were failing due to memory protection error, e.g. test_feature_weights.R. These tests crash immediately when run with --enable-strict-barrier.

I will get back to you with the full list of tests that are affected by the memory issue.

david-cortes:
> > Have the maintainer gotten back to you about reproducing the OpenBLAS environment? If not, let's ask for permission to remove the failing tests
> 
> Need to separate the issues here. I'm not sure whether the issue fixed by this PR is the cause of all test failures.
> 
> How many test failures did you reproduce using the `--enable-strict-barrier` flag? We have multiple test failures. Did you manage to reproduce all of them using this flag, some of them, or none of them? Could you please share the check.log for `enable-strict-barrier`?
> 
> Are you using docker to build the R variant? If so, I can run tests as well.

In the case of memory corruptions, since all the tests are executed in the same process, it might not necessarily be the case that problems come from the specific test that fails, nor that failures would be deterministically reproducible.

> Have the maintainer gotten back to you about reproducing the OpenBLAS environment?

They seldomly answer emails so I wouldn't count on getting a reply in due time.

hcho3:
@trivialfis Only the test in `test_feature_weights.R` is affected by the memory protection bug. Other tests run fine with `--enable-strict-barrier` flag enabled, but `test_feature_weights.R` crashes with segfault:
```
...
Fatal error: Wrong thread calling 'RunFinalizers'
Fatal error: Wrong thread calling 'RunFinalizers'
Fatal error: Wrong thread calling 'RunFinalizers'
Segmentation fault (core dumped)
```

I'd say the following course of action:
1. Make a new submission with the memory fix (this PR) and see if it also fixes other tests.
2. If the other tests still fail, we'd go ahead and remove the failing tests and re-submit, since we have no way to reproduce them.

trivialfis:
Asked for help again in the public list.

trivialfis:
It looks like CRAN maintainers accepted my previous package that removed the tests.

PR #11449, opened 2025-05-11
Add Possibility to Load XGBoost Models as PySpark Models

### LINKED ISSUE #11400: Load XGBoost model made in Python to spark
I got a xgboost model made in python (xgb_model.json) and I need to load it to Spark and on Scala too. When I try to do it it searches for /metadata in the path. What do I do?

ayoub317:
Closes #11400 

The default values for `device`, `use_gpu` and `tree_method` were added due to this [function call](https://github.com/dmlc/xgboost/blob/69268dff557ebd79629d1a698d3a58059fc7a914/python-package/xgboost/spark/core.py#L1060), and eventually this [one](https://github.com/dmlc/xgboost/blob/69268dff557ebd79629d1a698d3a58059fc7a914/python-package/xgboost/spark/core.py#L507) which assumes they already exist in `_paramMap` or `_defaultParamMap`.

### DIFF
--- a/python-package/xgboost/spark/core.py
+++ b/python-package/xgboost/spark/core.py
@@ -1239,6 +1239,61 @@ class _SparkXGBModel(Model, _SparkXGBParams, MLReadable, MLWritable):
         assert self._xgb_sklearn_model is not None
         return self._xgb_sklearn_model.get_booster()
 
+    @classmethod
+    def _load_model_as_sklearn_model(cls, model_path: str) -> XGBModel:
+        """
+        Subclasses should override this method and
+        returns a _SparkXGBModel subclass
+        """
+        raise NotImplementedError()
+
+    @classmethod
+    def convert_sklearn_model_to_spark_xgb_model(
+        cls,
+        xgb_sklearn_model: XGBModel,
+        training_summary: Optional[XGBoostTrainingSummary] = None,
+    ) -> "_SparkXGBModel":
+        """
+        Convert a sklearn model to pyspark xgboost model.
+        """
+        spark_xgb_model = cls(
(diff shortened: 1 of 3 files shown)


trivialfis:
cc @WeichenXu123 @wbo4958

python-package/xgboost/spark/core.py
```diff
@@ -1239,6 +1239,61 @@ def get_booster(self) -> Booster:
...
+    @classmethod
+    def _load_model_as_sklearn_model(cls, model_path: str) -> XGBModel:
+        """
+        Subclasses should override this method and
+        returns a _SparkXGBModel subclass
+        """
+        raise NotImplementedError()
+
+    @classmethod
+    def convert_sklearn_model_to_spark_xgb_model(
```
wbo4958, line 1251: will this function be exposed to users?

trivialfis (commented):
Thank you for working on this. But what's the difference between the approach taken here and the following snippet?
``` python
from xgboost.spark import SparkXGBRegressorModel

reg = XGBRegressor()

SparkXGBRegressorModel(reg, None)
```

ayoub317:
Hello @trivialfis ,
The code snippet you provided doesn't seem to load an XGBoost model and make predictions. I expect the loading part to work, but an error to occur during the prediction without this commit. It also I think packages the loading as an sklearn model and the conversion to a Spark model in a nice way.

python-package/xgboost/spark/core.py
ayoub317: I think it is good to expose it to users so that if they have already loaded or trained an sklearn model, and maybe want to make predictions on a large dataset they can do it without the need to save the model to disk and then load it back with the `load_model` method.

wbo4958:
LGTM.

PR #11453, opened 2025-05-12
[EM] Initial support for splitting up the host cache.

trivialfis:
- Add a new parameter to specify the portion of the cache.
- Split the host cache into host + device cache.

Limitations:
- Direct access to the cache page has not yet been implemented.
- More work is needed to find an optimal split policy. For now, auto means host only.


trivialfis:
@rongou

PR #11456, opened 2025-05-13
Reduce flakiness of the pyspark conf test.

trivialfis:
https://github.com/dmlc/xgboost/actions/runs/14984163779/job/42095164840?pr=11451
https://github.com/dmlc/xgboost/actions/runs/14991077836/job/42115037318?pr=11453

Find an available port prior to running the test. We can't guarantee the port will continue to be available during the test, but this should reduce the chance of using an occupied port.


trivialfis:
cc @wbo4958

Issue #11461, opened 2025-05-14
`xgboost.dask` broken for larger clusters


jrbourbeau:
In `distributed=2025.4.0`, `client.scheduler_info()` started truncating the number of worker info returned by default https://github.com/dask/distributed/pull/9045 (this scaled very poorly for large clusters). 

Today while running [this example](https://github.com/coiled/examples/blob/main/xgboost.ipynb) on a large Dask cluster on Coiled with `xgboost=3.0.1` I got this error:

```
KeyError                                  Traceback (most recent call last)
Cell In[5], line 20
     17 d_train = xgboost.dask.DaskDMatrix(None, X_train, y_train, enable_categorical=True)
     19 print("Training ...")
---> 20 model = xgboost.dask.train(
     21     None,
     22     {"tree_method": "hist"},
     23     d_train,
     24     num_boost_round=4,
     25     evals=[(d_train, "train")],
     26 )
     28 print("Scoring ...")
     29 predictions = xgboost.dask.predict(None, model, X_test)

File [~/miniforge3/envs/examples/lib/python3.10/site-packages/xgboost/core.py:729](http://localhost:8888/~/miniforge3/envs/examples/lib/python3.10/site-packages/xgboost/core.py#line=728), in require_keyword_args.<locals>.throw_if.<locals>.inner_f(*args, **kwargs)
    727 for k, arg in zip(sig.parameters, args):
    728     kwargs[k] = arg
--> 729 return func(**kwargs)

File [~/miniforge3/envs/examples/lib/python3.10/site-packages/xgboost/dask/__init__.py:927](http://localhost:8888/~/miniforge3/envs/examples/lib/python3.10/site-packages/xgboost/dask/__init__.py#line=926), in train(client, params, dtrain, num_boost_round, evals, obj, early_stopping_rounds, xgb_model, verbose_eval, callbacks, custom_metric, coll_cfg)
    925 client = _get_client(client)
    926 args = locals()
--> 927 return client.sync(
    928     _train_async,
    929     global_config=config.get_config(),
    930     dconfig=_get_dask_config(),
    931     **args,
    932 )

File [~/miniforge3/envs/examples/lib/python3.10/site-packages/distributed/utils.py:376](http://localhost:8888/~/miniforge3/envs/examples/lib/python3.10/site-packages/distributed/utils.py#line=375), in SyncMethodMixin.sync(self, func, asynchronous, callback_timeout, *args, **kwargs)
    374     return future
    375 else:
[...]

PR #11462, opened 2025-05-14
Dask 2025.4.0 scheduler info compatibility

jrbourbeau:
Closes https://github.com/dmlc/xgboost/issues/11461 per @trivialfis suggestion here https://github.com/dmlc/xgboost/issues/11461#issuecomment-2881427489

### DIFF
--- a/python-package/xgboost/dask/__init__.py
+++ b/python-package/xgboost/dask/__init__.py
@@ -919,12 +919,11 @@ def train(  # pylint: disable=unused-argument
 
     """
     client = _get_client(client)
-    args = locals()
     return client.sync(
         _train_async,
         global_config=config.get_config(),
         dconfig=_get_dask_config(),
-        **args,
+        **locals(),
     )
 
 
--- a/python-package/xgboost/dask/__init__.py
+++ b/python-package/xgboost/dask/__init__.py
@@ -254,15 +254,11 @@ class CommunicatorContext(collective.CommunicatorContext):
         super().__init__(**args)
 
         worker = distributed.get_worker()
-        with distributed.worker_client() as client:
-            info = client.scheduler_info()
-            w = info["workers"][worker.address]
-            wid = w["id"]
         # We use task ID for rank assignment which makes the RABIT rank consistent (but
         # not the same as task ID is string and "10" is sorted before "2") with dask
-        # worker ID. This outsources the rank assignment to dask and prevents
+        # worker address. This outsources the rank assignment to dask and prevents
         # non-deterministic issue.
-        self.args["DMLC_TASK_ID"] = f"[xgboost.dask-{wid}]:" + str(worker.address)
+        self.args["DMLC_TASK_ID"] = f"[xgboost.dask]:{worker.address}"
 
 
 def _get_client(client: Optional["distributed.Client"]) -> "distributed.Client":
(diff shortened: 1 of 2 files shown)


jrbourbeau:
@trivialfis this test 

https://github.com/dmlc/xgboost/blob/9ad4e2445bb2fa873f63d141b90d453ef61b7e32/tests/python/test_tracker.py#L160-L181

is explicitly checking the previous worker id is in the task id used for determining the rank. Do you have a suggestion on how to update this test? Or should it be removed?

trivialfis:
> Do you have a suggestion on how to update this test? Or should it be removed?

Based on https://github.com/dmlc/xgboost/issues/11461#issuecomment-2881510696 , can we use `worker.name` as a replacement for the current wid?

jrbourbeau:
> Based on https://github.com/dmlc/xgboost/issues/11461#issuecomment-2881510696 , can we use worker.name as a replacement for the current wid?

Yeah, fair point. I just want to reiterate whether the current `wid` (and the equivalent `worker.name`) are integers like 0, 1, 2, 3, ... depends on how the workers were created. `distributed.LocalCluster` sets them to be integers but other deployments can set those names to be strings (or any other hashable object). 

That said, that's a separate issue. I've gone ahead and updated this PR to avoid the `scheduler_info()` call altogether by using `worker.name` for the current `wid`. That should keep the current behavior.

trivialfis:
Please ignore the canceled error. I don't know how it happens, it's dark magic to me.

On the other hand, we have seen this https://github.com/dmlc/xgboost/actions/runs/15050046963/job/42303211874?pr=11462 a couple of times before. Would you like to make a guess? (not expecting you to spend time debugging it). It's a flaky error that looks like dask is having trouble with parameter packs:
```
     |   File "/workspace/tests/test_distributed/test_gpu_with_dask/test_gpu_with_dask.py", line 194, in run_gpu_hist
    |     history = dxgb.train(
    |   File "/home/runner/.local/lib/python3.10/site-packages/xgboost/core.py", line 772, in inner_f
    |     return func(**kwargs)
    |   File "/home/runner/.local/lib/python3.10/site-packages/xgboost/dask/__init__.py", line 923, in train
    |     return client.sync(
    |   File "/opt/miniforge/envs/gpu_test/lib/python3.10/site-packages/distributed/utils.py", line 363, in sync
    |     return sync(
    |   File "/opt/miniforge/envs/gpu_test/lib/python3.10/site-packages/distributed/utils.py", line 439, in sync
    |     raise error
    |   File "/opt/miniforge/envs/gpu_test/lib/python3.10/site-packages/distributed/utils.py", line 409, in f
    |     awaitable = func(*args, **kwargs)
    | TypeError: _train_async() got an unexpected keyword argument 'args'
    +------------------------------------
```

trivialfis (approved):

jrbourbeau:
Btw, do you have a sense for when this change might be included in a release? I see it just missed the 3.0.1 patch release unfortunately

trivialfis:
We can do another one if necessary, cc @hcho3 .

jrbourbeau:
I'm not sure how much of a burden it is to release, but I'd certainly appreciate a release with this update as soon as makes sense for you all (most `xgboost.dask` usage that I'm aware of in the wild runs on clusters with more than 5 workers, so would benefit from this update).

PR #11465, opened 2025-05-15
Implement double-buffer for ellpack.

trivialfis:
This helps us to have the external memory cache in two different places.

- Make the accessor a variant type.
- Most of the changes are for calling a visit function instead of actual code changes.

todos:
- [x] Test the new iterator.
- [x] Test for external memory.
- [x] Alignment.
- [x] Profiling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -163,3 +163,4 @@ Rplots.pdf
 
 # nsys
 *.nsys-rep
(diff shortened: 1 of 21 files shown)


trivialfis:
cc @rongou

rongou (approved):

PR #11471, opened 2025-05-19
Fix VM fallback logic on WSL2

hcho3:
On WSL2, `nvidia-smi` returns a driver version of form `573.24`, with two components instead of three.

I tested this change on my work laptop (Dell Precision 5480, NVIDIA RTX A1000 6GB Laptop GPU, Windows 11 with WSL2).


trivialfis (approved):

hcho3:
I will run some smoke tests with pytest before merging.

hcho3:
Side note: WSL2 is surprising pleasant to use as a development environment, especially with major IDEs supporting Remote Development feature.

PR #11474, opened 2025-05-21
[jvm-packages] ExtenralMemory: Overlap the caching time

wbo4958:
This PR tries to overlap the caching time (caching cudf table to the disk when enabling external memory).

The previous design is

1. cache the cudf table to the disk
2. delivery the cudf table to the xgboost to build QuantileDMatrix incrementally

So, they are running in sequence which is time-consuming.

This PR moves the caching process to run in a separate thread, which can be overlapped when building QuantileDMatrix.

I did some test locally with total 13.1G dataset, 

With this PR, it's going to eliminate the 90%+ caching time.

### DIFF
--- a/jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/ExternalMemory.scala
+++ b/jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/ExternalMemory.scala
@@ -22,7 +22,8 @@ import java.util.concurrent.Executors
 
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.{Await, ExecutionContext, Future}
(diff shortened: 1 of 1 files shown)


jvm-packages/xgboost4j-spark-gpu/src/main/scala/ml/dmlc/xgboost4j/scala/spark/ExternalMemory.scala
```diff
@@ -145,19 +146,17 @@ private[spark] class DiskExternalMemoryIterator(val parent: String) extends Exte
...
-    while (count < 120) {
-      val futureOpt = taskFutures.get(path)
-      val exist = new File(path).exists()
-      if (futureOpt.isDefined && futureOpt.get.isCompleted && exist) {
-        return
-      }
-      count += 1
-      Thread.sleep(50)
+    // If timeout, it's going to throw exception
+    val success = Await.result(futureOpt.get, 6.seconds)
```
trivialfis, line 156: 6 seconds is not a lot when data is large and system is busy, why do we need timeout here? Can writing to disk hang?

PR #11483, opened 2025-05-25
Update install doc for Conda

hcho3:
* There is no longer Conda package called `py-xgboost-gpu`. Use variants instead, e.g. `py-xgboost=*=cuda*`.
* The GPU variant is now available on Windows. I just tested it on my Windows workstation.
* Recommend Miniforge instead of Miniconda.

Preview: https://xgboost--11483.org.readthedocs.build/en/11483/install.html#conda

@trivialfis Unfortunately, this isn't part of the 3.0.2 tag, so the `stable` doc on RTD won't display the updated Conda installation method. I'm inclined to create a new tag `3.0.2-post0` so that the RTD shows the updated install doc. What do you think?

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -104,25 +104,25 @@ Conda should be able to detect the existence of a GPU on your machine and instal
 
 .. code-block:: bash
 
-   # CPU only
-   conda install -c conda-forge py-xgboost-cpu
-   # Use NVIDIA GPU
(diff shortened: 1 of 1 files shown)


trivialfis:
> I'm inclined to create a new tag 3.0.2-post0 so that the RTD shows the updated install doc. What do you think?

Sounds good to me.

trivialfis (approved):

PR #11487, opened 2025-05-27
Benchmark ext mem connect

wbo4958:


trivialfis:
I don't think we are going anywhere with the Python wrapper of the Java package.

PR #11504, opened 2025-06-09
[CI] Test with Rapids 25.06

hcho3:
https://github.com/dmlc/xgboost-devops/pull/27

* The current test suite already passes with Rapids 25.06.
* Update XGBoost to build with CCCL 3.0. (Currently, we don't yet run pytests with Rapids 25.08. We only test the C++ build.)


chyunsu3:
@trivialfis Can I get a review on the change?

PR #11519, opened 2025-06-20
optimize CPU inference with Array-Based Tree Traversal

razdoburdin:
This PR introduces optimization for CPU inference. For each tree, the top N levels are transformed into a compact array-based layout. This allows for a branchless node indexing rule: idx = 2 * idx + int(val < split_cond). To minimize memory overhead, this transformation from the standard tree structure to the array layout is performed on-the-fly for each block of data being processed.  Even with this additional calculations, improved data locality in the cache-friendly array layout leads to inference speed up to ~2x (x1.4 on average). 
image


trivialfis:
Thank you for the optimization on the inference. Please unmark the "draft" status and ping me when the PR is ready for testing.

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,250 @@
...
+  DefaultLeftType default_left_;
+  IsCatType is_cat_;
+  CatSegmentType cat_segment_;
+
+  std::array<bst_feature_t, kNodesCount> split_index_;
+  std::array<float, kNodesCount> split_cond_;
+  std::array<bst_node_t, kNodesCount + 1> nidx_in_tree_;
+
+  inline static bool IsLeaf(const RegTree& tree, bst_feature_t nidx) {
+    return tree[nidx].IsLeaf();
```
Vika-F: Please add this in order to prevent possible errors in compile time.
```suggestion
    static_assert(std::is_same_v<RegTree, TreeType>); 
    return tree[nidx].IsLeaf();
```

Same applies to other similar cases.

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,273 @@
...
+ * \tparam has_categorical if the tree has categorical features
+ *
+ * \tparam any_missing if the class is able to process missing values
+ * 
+ * \tparam kNumDeepLevels number of tree leveles being unrolled into array-based structure
+ */
+template <class TreeType, bool has_categorical, bool any_missing, int kNumDeepLevels>
+class ArrayTreeLayout {
+ private:
+  constexpr static size_t kNodesCount = (1u << kNumDeepLevels) - 1;
```
Vika-F, line 32:
```suggestion
  /* Number of nodes in the array based representation of the top levels of the tree
   */
  constexpr static size_t kNodesCount = (1u << kNumDeepLevels) - 1;
```

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,273 @@
...
+ * 
+ */
+  template <int depth = 0>
+  void inline Populate(const TreeType& tree, RegTree::CategoricalSplitMatrix const &cats,
+                       bst_node_t nidx_array = 0, bst_node_t nidx = 0) {
+    if constexpr (depth == kNumDeepLevels + 1) {
+      return;
+    } else if constexpr (depth == kNumDeepLevels) {
+        /* We save the node index in the origianl tree to able to continue processing
+         * for nodes not egligable for array layout optimisation.
```
Vika-F:
```suggestion
        /* We store the node index in the original tree to ensure continued processing
         * for nodes that are not eligible for array layout optimization. 
```

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,273 @@
...
+        nidx_in_tree_[nidx_array - kNodesCount] = nidx;
+    } else {
+      if (IsLeaf(tree, nidx)) {
+        split_index_[nidx_array]  = 0;
+
+        /* 
+         * if the tree is not fully populated, we can reduce transfering costs.
+         * the values for unpopulated part of the tree are set in a way to guarantie
+         * that a moove will always done in "right" direction.
+         * here we exploiting that comparison with nan always results to false.
```
Vika-F:
```suggestion
         * If the tree is not fully populated, we can reduce transfer costs.
         * The values for the unpopulated parts of the tree are set to ensure
         * that any move will always proceed in the "right" direction.
         * This is achieved by exploiting the fact that comparisons with NaN always result in false.
```

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,273 @@
...
+                                   std::array<common::Span<uint32_t const>, kNodesCount>,
+                                   struct Empty>;
+
+  DefaultLeftType default_left_;
+  IsCatType is_cat_;
+  CatSegmentType cat_segment_;
+
+  std::array<bst_feature_t, kNodesCount> split_index_;
+  std::array<float, kNodesCount> split_cond_;
+  std::array<bst_node_t, kNodesCount + 1> nidx_in_tree_;
```
Vika-F, line 56:
```suggestion
  /* The nodes at tree levels 0, 1, ..., kNumDeepLevels - 1 are unrolled into an array-based structure.
   *  If the tree has additional levels, this array stores the node indices of the sub-trees at level kNumDeepLevels.
   *  This is necessary to continue processing nodes that are not eligible for array-based unrolling.
   *  The number of sub-trees packed into this array is equal to the number of nodes at tree level kNumDeepLevels,
   *  which is calculated as (1u << kNumDeepLevels) == kNodesCount + 1.
   */
  std::array<bst_node_t, kNodesCount + 1> nidx_in_tree_;
```

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,273 @@
...
+        split_index_[nidx_array]  = SplitIndex(tree, nidx);
+        split_cond_[nidx_array]   = SplitCond(tree, nidx);
+
+        /*
+         * LeftChild is used to find if the node is leaf, so it is a valid value,
+         * howerwer RightChild can be invalid in some exotic case.
+         * The tree with invalid right-child can be correctly processed by a classical method,
+         * if the split conditions are propper.
+         * But for array layout invalid RightChild, even unreachable, will lead to memory corruption.
+         * Add check to prevent it.
```
Vika-F:
```suggestion
         * LeftChild is used to determine if a node is a leaf, so it is always a valid value.
         * However, RightChild can be invalid in some exotic cases.
         * A tree with an invalid RightChild can still be correctly processed using classical methods
         * if the split conditions are correct.
         * However, in an array layout, an invalid RightChild, even if unreachable, can lead to memory corruption.
         * A check should be added to prevent this.
```

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,273 @@
...
+
+#include <limits>
+#include <vector>
+
+namespace xgboost::predictor {
+
+/**
+ * @brief The class holds the array-based representation of the top levels of a single tree.
+ * 
+ * \tparam TreeType The type of the origianl tree (RegTree or MultiTargetTree)
```
Vika-F: Strange mix of `@brief` and `\tparam`.
It would be more coherent to use either (`@brief` and `@tparam`) or (`\brief` and `\tparam`)

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,273 @@
...
+   */
+  constexpr static int kMaxNumDeepLevels = 6;
+  static_assert(kNumDeepLevels <= kMaxNumDeepLevels);
+
+  ArrayTreeLayout(const TreeType& tree, RegTree::CategoricalSplitMatrix const &cats) {
+    Populate(tree, cats);
+  }
+
+  /**
+   * @brief
```
Vika-F: Also incoherent use of `@brief` and `\param`

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,273 @@
...
+   * Traverse top levels of the tree for an entire block_size.
+   * In array layout is orginised to garantie that
+   * if the node at the current level has index nidx, than
+   * the node index for left child at the next level is always 2*nidx
+   * the node index for right child at the next level is always 2*nidx+1
+   * This greatly improve data locality
```
Vika-F:
```suggestion
   * Traverse the top levels of the tree for the entire block_size.
   * In the array layout, it is organized to guarantee that
   * if a node at the current level has index nidx, then
   * the node index for the left child at the next level is always 2*nidx, and
   * the node index for the right child at the next level is always 2*nidx+1.
   * This greatly improves data locality.
```

src/predictor/cpu_predictor.cc
```diff
@@ -319,6 +402,19 @@ void PredictBatchByBlockOfRowsKernel(DataView const &batch, gbm::GBTreeModel con
   auto const n_features = model.learner_model_param->num_feature;
   auto const n_blocks = common::DivRoundUp(n_samples, kBlockOfRowsSize);
 
+  /* Precalculate depth for each tree.
+   * This values are requared only for ArrayLyout optimization,
```
Vika-F:
```suggestion
   * This values are required only for ArrayLyout optimization,
```

src/predictor/array_tree_layout.h
razdoburdin: done

src/predictor/array_tree_layout.h
razdoburdin: done

razdoburdin:
> The next possible step would be to convert the trees into array-based representation only once, and not to do it for each block of data.

it sounds reasonable and will further improve perf (by cost of increasing memory consumption).

razdoburdin:
> Thank you for the optimization on the inference. Please unmark the "draft" status and ping me when the PR is ready for testing.

hi @trivialfis, the PR is ready for review.

trivialfis:
cc @hcho3

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,281 @@
...
+  std::array<float, kNodesCount> split_cond_;
+  /* The nodes at tree levels 0, 1, ..., kNumDeepLevels - 1 are unrolled into an array-based structure.
+   *  If the tree has additional levels, this array stores the node indices of the sub-trees at level kNumDeepLevels.
+   *  This is necessary to continue processing nodes that are not eligible for array-based unrolling.
+   *  The number of sub-trees packed into this array is equal to the number of nodes at tree level kNumDeepLevels,
+   *  which is calculated as (1u << kNumDeepLevels) == kNodesCount + 1.
+   */
+  std::array<bst_node_t, kNodesCount + 1> nidx_in_tree_;
+
+  inline static bool IsLeaf(const RegTree& tree, bst_node_t nidx) {
```
trivialfis: No need for the `inline` keyword, same goes for all following methods.

src/predictor/cpu_predictor.cc
```diff
@@ -84,56 +109,113 @@ bst_node_t GetLeafIndex(MultiTargetTree const &tree, const RegTree::FVec &feat,
...
+void PredictByAllTrees(gbm::GBTreeModel const &model, bst_tree_t const tree_begin,
+                       bst_tree_t const tree_end, std::size_t const predict_offset,
+                       std::vector<RegTree::FVec> const &thread_temp, std::size_t const offset,
+                       std::size_t const block_size, linalg::MatrixView<float> out_predt,
+                       const std::vector<int>& tree_depth, bool any_missing) {
+  /*
+   * We use transforming trees to array layout for each block of data to avoid memory overheads.
+   * It makes the array layout inefficient for block_size == 1
+   */ 
+  const bool use_array_tree_layout = block_size > 1;
```
trivialfis, line 193: What happens if this is a small online inference call? The input size could be a few samples per call.

src/predictor/cpu_predictor.cc
```diff
@@ -59,19 +59,44 @@ bst_node_t GetLeafIndex(RegTree const &tree, const RegTree::FVec &feat,
...
+                        bst_node_t* p_nidx, int depth, int gid) {
+  auto const &cats = tree.GetCategoriesMatrix();
+  if constexpr (use_array_tree_layout) {
+    ProcessArrayTree<RegTree, has_categorical, any_missing>
+        (tree, cats, thread_temp, offset, block_size, p_nidx, depth);
+  }
+  for (std::size_t i = 0; i < block_size; ++i) {
+    bst_node_t nidx = 0;
+    if constexpr (use_array_tree_layout) {
+      nidx = p_nidx[i];
```
trivialfis, line 87: unused?

trivialfis (commented):
Still trying to understand the code, will give it a try later. In the meanwhile, could you please craft some specific unittests for the new inference algorithm?

src/predictor/cpu_predictor.cc
razdoburdin: The default (the old one) realization will be used

src/predictor/array_tree_layout.h
razdoburdin: done

src/predictor/cpu_predictor.cc
razdoburdin: The optimized array_layout processing is effective only for nodes, that are close to root. For other nodes, we still use the original method.

razdoburdin:
> Still trying to understand the code, will give it a try later. In the meanwhile, could you please craft some specific unittests for the new inference algorithm?

I added some unit-tests.

src/predictor/cpu_predictor.cc
```diff
@@ -319,6 +406,19 @@ void PredictBatchByBlockOfRowsKernel(DataView const &batch, gbm::GBTreeModel con
   auto const n_features = model.learner_model_param->num_feature;
   auto const n_blocks = common::DivRoundUp(n_samples, kBlockOfRowsSize);
 
+  /* Precalculate depth for each tree.
+   * This values are required only for ArrayLyout optimization,
```
trivialfis:
```suggestion
   * These values are required only for the ArrayLayout optimization,
```

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,300 @@
...
+  IsCatType is_cat_;
+  CatSegmentType cat_segment_;
+
+  std::array<bst_feature_t, kNodesCount> split_index_;
+  std::array<float, kNodesCount> split_cond_;
+  /* The nodes at tree levels 0, 1, ..., kNumDeepLevels - 1 are unrolled into an array-based structure.
+   *  If the tree has additional levels, this array stores the node indices of the sub-trees at level kNumDeepLevels.
+   *  This is necessary to continue processing nodes that are not eligible for array-based unrolling.
+   *  The number of sub-trees packed into this array is equal to the number of nodes at tree level kNumDeepLevels,
+   *  which is calculated as (1u << kNumDeepLevels) == kNodesCount + 1.
```
trivialfis, line 53: What happens if the tree is not well balanced and is more like a linked list than a tree?

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,300 @@
...
+   */
+  std::array<bst_node_t, kNodesCount + 1> nidx_in_tree_;
+
+  static bool IsLeaf(const RegTree& tree, bst_node_t nidx) {
```
trivialfis: Is there a benefit of doing this C++ overloading rather than the simpler `tree.IsLeaf`? How much faster are we seeing?

src/predictor/array_tree_layout.h
razdoburdin: In this case we add dummy nodes with `nan` as split condition. In this dummy node the decision is always "go right" (see also comments https://github.com/razdoburdin/xgboost/blob/b0eaa856e1246416f7f9538bcc004e7723d9b997/src/predictor/array_tree_layout.h#L154), the left child are not initialized. 

So with array layout we have to allocated all nodes, but keep some of the unpopulated in case the tree is pure balanced.

Initial tree:
image

Tree with dummy (nan-valued nodes):
image

src/predictor/array_tree_layout.h
razdoburdin: I did the overload to handle both `RegTree` and `MultiTargetTree` cases. Is there a better option?

trivialfis:
I'm trying to cleanup the CPU predictor. I will update this PR once it is finished.

trivialfis:
I need to fix a perf regression caused by the new ordinal encoder.

trivialfis:
> I need to fix a perf regression caused by the new ordinal encoder.

This has been fixed. I will look deeper into this PR.

src/predictor/array_tree_layout.h
```diff
@@ -0,0 +1,300 @@
...
+ private:
+  /* Number of nodes in the array based representation of the top levels of the tree
+   */
+  constexpr static size_t kNodesCount = (1u << kNumDeepLevels) - 1;
+
+  struct Empty {};
+  using DefaultLeftType =
+        typename std::conditional_t<any_missing,
+                                   std::array<uint8_t, kNodesCount>,
+                                   struct Empty>;
```
trivialfis:
```suggestion
                                   Empty>;
```

src/predictor/array_tree_layout.h
trivialfis: Thank you for sharing, that makes sense.

src/predictor/array_tree_layout.h
trivialfis: Use `RegTree` without extracting the Multi-target tree when populating the buffer, and delegate the dispatching to `RegTree::LeftChild(bst_node_t nidx)` instead of using the `RegTree::Node::LeftChild`. There's a check inside the `RegTree::LeftChild`:

``` c++
  [[nodiscard]] bst_node_t LeftChild(bst_node_t nidx) const {
    if (IsMultiTarget()) {
      return this->p_mt_tree_->LeftChild(nidx);
    }
    return (*this)[nidx].LeftChild();
  }
```

trivialfis:
Thank you for expanding the tree layout. In the future (when you can prioritize it), do you think it's possible to create and store the layout inside the `RegTree` structure as an opt-in method call? My reasoning is as follows:

- The existing RegTree and the multi-target tree already use a very similar layout, minus the dummy nodes. It might be easier/cleaner to do it there.
- We can avoid complicating the predictor too much.
- We can cache the result in the regtree structure to avoid repeated initialization.

You can define a `std::unique_ptr<ArrayTree>` inside the `RegTree`, set it to `nullptr`. Define a method to create the array tree when needed, and reset it back to `nullptr` if any non-const method is called.

razdoburdin:
> Thank you for expanding the tree layout. In the future (when you can prioritize it), do you think it's possible to create and store the layout inside the `RegTree` structure as an opt-in method call? My reasoning is as follows:
> 
> * The existing RegTree and the multi-target tree already use a very similar layout, minus the dummy nodes. It might be easier/cleaner to do it there.
> * We can avoid complicating the predictor too much.
> * We can cache the result in the regtree structure to avoid repeated initialization.
> 
> You can define a `std::unique_ptr<ArrayTree>` inside the `RegTree`, set it to `nullptr`. Define a method to create the array tree when needed, and reset it back to `nullptr` if any non-const method is called.

Do you think memory overhead (about 1KB per tree) is acceptable for storing the layout? If so, it would be the natural next optimization step.

trivialfis:
> Do you think memory overhead (about 1KB per tree) is acceptable for storing the layout? 

I think this should be fine since the size of the layout is bound by depth. The implementation here falls back to the original tree after certain level is reached.

src/predictor/array_tree_layout.h
razdoburdin: done

razdoburdin:
> > Do you think memory overhead (about 1KB per tree) is acceptable for storing the layout?
> 
> I think this should be fine since the size of the layout is bound by depth. The implementation here falls back to the original tree after certain level is reached.

Can we merge the current implementation and postpone buffering of the layout?

trivialfis:
> Can we merge the current implementation and postpone buffering of the layout?

We can. I will look into this PR.

trivialfis (approved):
Thank you for the excellent optimization!

I can understand the code (mostly), and it should be cleaner after merging into the regtree. I will merge this PR once the CI is green.

trivialfis:
Revisiting this thread since I have been looking into the predictor.

I think fast inference should be left to a dedicated library, or at least a dedicated class in XGBoost. There are too many potential memory layouts with different tradeoff. Also, we need a better way to ensure thread safety and thread-safe memory allocation. The booster is not the right place to do it as it's mutable and contains too many unneeded data for inference.

This note is mostly for future reference if there's a plan to continue the optimization. (No change request)

trivialfis:
That said, optimizations on inference are welcome. Just mentioning that if more memory layout or cache is on the pipeline, consider creating something outside of the booster class.

PR #11531, opened 2025-06-26
[CI] [WheelNext] Build variant wheels

hcho3:
WheelNext is a proposal to improve Python wheels to better support specialized hardware and reduce the size of individual wheels. In particular, we are keen to adopt [the Wheel Variant proposal](https://wheelnext.dev/proposals/pepxxx_wheel_variant_support/). Also see [the presentation](https://wheelnext.dev/summits/2025_03/assets/WheelNext%20Community%20Summit%20-%2012%20-%20Wheel%20Variants.pdf).

Use the prototype in https://github.com/wheelnext/pep_xxx_wheel_variants to build a variant wheel. A variant wheel is just like a regular wheel, but with extra variant metadata like `nvidia :: cuda :: 12`.

### DIFF
--- a/ops/pipeline/build-variant-wheels-impl.sh
+++ b/ops/pipeline/build-variant-wheels-impl.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+## Build Python wheels using Wheel Variant prototype (WheelNext)
+## Companion script for ops/pipeline/build-variant-wheels.sh
+
+set -eo pipefail
+
+set -x
+gosu root chown -R $(id -u):$(id -g) /opt/miniforge/envs /opt/miniforge/pkgs/cache
+gosu root chown $(id -u):$(id -g) /opt/miniforge/pkgs
+set +x
+
+mamba create -y -n wheelnext python=3.13 python-build
+
+source activate wheelnext
+
+# Cannot set -u before Conda env activation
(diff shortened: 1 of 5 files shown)


ops/pipeline/build-variant-wheels-impl.sh
```diff
@@ -0,0 +1,23 @@
...
+
+source activate wheelnext
+
+# Cannot set -u before Conda env activation
+set -xu
+
+python -m pip install -v \
+  git+https://github.com/wheelnext/pep_xxx_wheel_variants.git@f3b287090f8a6f510b0e1723896e1c7e638f6bff#subdirectory=pep_xxx_wheel_variants
+pip config set --site global.index-url https://variants-index.wheelnext.dev/
+variantlib make-variant -f python-package/dist/xgboost-*.whl \
```
trivialfis, line 22: Could you please provide some comments on what this line does? Skimming the pipeline, this follows a CUDA build, I don't see any potential variant here since there's no CPU build and NCCL is still installed.

chyunsu3: This is an initial stage, where we build a single variant for the CUDA build. In the future, I plan to build the CPU variant and other variants.

PR #11533, opened 2025-06-26
[EM] Use memory pool if it's available.

trivialfis:


trivialfis:
Closing as it's too slow.

PR #11538, opened 2025-06-27
[doc][EM] Add a brief introduction to NUMA.

trivialfis:
~Add a utility to help set the CPU affinity.~

### DIFF
--- a/doc/tutorials/external_memory.rst
+++ b/doc/tutorials/external_memory.rst
@@ -306,7 +306,7 @@ To obtain the node ID, you can check the machine topology via ``nvidia-smi``:
 The column ``NUMA Affinity`` lists the NUMA node ID for each GPU. In the example output
 shown below, the `GPU0` is associated with the `0` node ID::
 
-	    GPU0    GPU1    NIC0    NIC1    NIC2    NIC3    CPU Affinity    NUMA Affinity   GPU NUMA ID
+            GPU0    GPU1    NIC0    NIC1    NIC2    NIC3    CPU Affinity    NUMA Affinity   GPU NUMA ID
     GPU0     X      NV18    NODE    NODE    NODE    SYS     0-71            0               2
     GPU1    NV18     X      SYS     SYS     SYS     NODE    72-143          1               10
     NIC0    NODE    SYS      X      PIX     NODE    SYS


demo/guide-python/external_memory.py
```diff
@@ -44,14 +46,29 @@
...
 
+def set_cpu_affinity() -> None:
+    """Set CPU affinity for NUMA."""
+    try:
+        import pynvml as nm
+
+        nm.nvmlInit()
+
+        hdl = nm.nvmlDeviceGetHandleByIndex(0)
+        nm.nvmlDeviceSetCpuAffinity(hdl)
```
pentschev: From the [nvmlDeviceSetCpuAffinity docs](https://docs.nvidia.com/deploy/nvml-api/group__nvmlAffinity.html#group__nvmlAffinity_1g1c52b8aefbf804bcb0ab07f5d872c2a1), this only sets the calling _thread_ affinity, and not the entire process. It's important to note that many applications/libraries use multiple threads and thus having affinity limited to only the threads that call this function may be suboptimal if, for example, a third-party library spawns a separate thread that handles data and thus may cause that data to be populated to a different NUMA node. If you work on the basis of one-GPU-per-process then you may be able to [set the whole process NUMA affinity using `os.sched_setaffinity` like you've seen in Dask-CUDA](https://github.com/rapidsai/dask-cuda/blob/6c3223d7f82f4b2e7eb209a8485b1e32cb5357a7/dask_cuda/plugins.py#L14-L27).

trivialfis: I'm considering using the Linux scheduling procedures instead of the NVML function.

I have tested using the nvml one, and it does significantly improve the performance for the child threads. But as you have pointed out, this seems ambiguous.

pentschev: Yeah, from Python that would simply be `os.sched_setaffinity`, I believe it directly wraps the OS' affinity scheduling.

trivialfis: Added a utility for setting the affinity with the `os` module.

demo/guide-python/distributed_extmem_basic.py
```diff
@@ -13,8 +13,9 @@
 If `device` is `cuda`, following are also needed:
 
 - cupy
-- python-cuda
 - rmm
+- pynvml (optional)
```
pentschev: Although the Python module is still named `pynvml`, the package has been deprecated ([see warning at the top](https://github.com/gpuopenanalytics/pynvml)). I would suggest instead documenting the new package name [nvidia-ml-py](https://pypi.org/project/nvidia-ml-py/).

demo/guide-python/external_memory.py
```diff
@@ -23,6 +23,7 @@
 
 - cupy
 - rmm
+- pynvml (optional)
```
pentschev: Same here, nvidia-ml-py.

python-package/xgboost/utils.py
```diff
@@ -0,0 +1,162 @@
...
+            hdl,
+            math.ceil(cnt / _MASK_SIZE),
+        )
+        cpumask = _BitField64(affinity)
+        cpus = list(filter(cpumask.check, range(cnt)))
+
+        nm.nvmlShutdown()
+        return cpus
+    except ImportError:
+        warnings.warn("Failed to import nvml.", UserWarning)
```
pentschev:
```suggestion
        warnings.warn("Failed to import pynvml.", UserWarning)
```

python-package/xgboost/utils.py
```diff
@@ -0,0 +1,162 @@
...
+        if i in dash_pos:
+            uuid += "-"
+        h = hex(0xFF & np.int32(prop.uuid.bytes[i]))
+        assert h[:2] == "0x"
+        h = h[2:]
+
+        while len(h) < 2:
+            h = "0" + h
+        uuid += h
+    return uuid
```
pentschev: Why? https://github.com/rapidsai/dask-cuda/blob/6c3223d7f82f4b2e7eb209a8485b1e32cb5357a7/dask_cuda/utils.py#L752-L775

python-package/xgboost/utils.py
```diff
@@ -0,0 +1,162 @@
...
+
+        nm.nvmlInit()
+
+        ordinal = ordinal if ordinal is not None else current_device()
+        cpus = get_device_cpu_affinity(device)
+        os.sched_setaffinity(0, cpus)
+
+        nm.nvmlShutdown()
+    except ImportError:
+        warnings.warn("Failed to import nvml. CPU affinity is not set.", UserWarning)
```
pentschev:
```suggestion
        warnings.warn("Failed to import pynvml. CPU affinity is not set.", UserWarning)
```

python-package/xgboost/utils.py
```diff
@@ -0,0 +1,162 @@
...
+        device = "cuda"
+
+    split = device.split(":")
+    if len(split) == 1:
+        ordinal = 0
+    elif len(split) == 2:
+        ordinal = int(split[1])
+    else:
+        raise ValueError(f"Invalid device: {device}")
+    return ordinal
```
pentschev: Seems overly complicated, why not just take an integer representing an index or the UUID? Also done in https://github.com/rapidsai/dask-cuda/blob/6c3223d7f82f4b2e7eb209a8485b1e32cb5357a7/dask_cuda/utils.py#L752-L775.

trivialfis: Primarily for consistency with XGBoost's notation for devices: "cuda:0", "cpu", I'm trying not to come up with a new one.

pentschev: Ah ok, makes sense. I didn't remember how XGBoost notation looked like.

python-package/xgboost/utils.py
trivialfis: Thank you for pointing it out. Got a bit too used to cudart.

demo/guide-python/external_memory.py
trivialfis: Done.

demo/guide-python/distributed_extmem_basic.py
trivialfis: Done.

trivialfis:
@pentschev I have simplified the utility using nvml and changed the package name in the documentation.

python-package/xgboost/utils.py
trivialfis: Thinking again, I need this cudart version as XGBoost should prefer the CUDA device enumeration instead of the nvml device enumeration.

hcho3 (approved):
I assume this facility is only available for Linux? (So not Windows, not MacOS)

trivialfis:
Update:
- Documented that this is Linux-only.
- Use CUDA for device enumeration. We obtain the device ordinal and corresponding UUID from CUDA, then use the UUID to get the affinity. This way, we can honor the `CUDA_VISIBLE_DEVICES` environment variable.
- Manually tested on an NVL system.

trivialfis:
Hold this PR for now. I think the CPU affinity alone is not sufficient when the memory is under pressure.

trivialfis:
Expanded the document for using `numactl`, and removed all the utilities. Removing the utilities since CPU affinity alone is not sufficient when memory is under pressure. I don't want to maintain specialized code that is not effective.

PR #11540, opened 2025-06-30
Avoid the use of `restrict` in the evaluator.

trivialfis:
It somehow caches the written value in the registry, resulting in an incorrect value in the second iteration of the scan.

No performance loss is observed.


trivialfis:
cc @rongou .

rongou (approved):

Issue #11510, opened 2025-06-12
[discussion] How to support spark 4.0


wbo4958:
Spark community has released the latest [spark 4.0](https://spark.apache.org/docs/latest/quick-start.html), which requires a minimum JAVA 17, and scala 2.13. So I filed this issue to discuss how do we support spark 4.0 in XGBoost JVM packages.

I would like to propose dropping spark 3.x support in main branch

Two reasons, 1). The xgboost jvm currently is built against scala 2.12/java 8 by default which is not working in spark 4.0. There may be some compatible issues between spark 4.0 and spark 3.x in terms of spark ML library.  2). spark 4.0 introduced a brand new Spark-Connect-ML which is not residing in spark 3.x

wbo4958:
@WeichenXu123 @trivialfis @hcho3, what's your opinion?

chyunsu3:
It depends on whether we should continue supporting Spark 3.x. As a maintainer of CI infra, I'd prefer if we transition to Spark 4 right away, but of course we need to think of the interests of major users too.

chyunsu3:
While we're at it, let's also consider using JVM 22 so that we can move away from Java Native Interface (JNI).

trivialfis:
I couldn't find information about aarch64 support status for the FFM API. It's an important consideration to have.

chyunsu3:
Linux aarch64 should be well supported.

From https://openjdk.org/jeps/454:

> While the Linker interface is neutral with respect to calling conventions, the native linker is optimized for the calling conventions of many platforms:
> * Linux/x64
> * Linux/AArch64
> * Linux/RISC-V
> * Linux/PPC64
> * Linux/s390
> * macOS/x64
> * macOS/AArch64
> * Windows/x64
> * Windows/AArch64
> * AIX/ppc64

trivialfis:
Sounds good! I will look into the FFM unless someone beats me to it.

wbo4958:
Yeah, personally, I'd like not to support Spark 3.x in xgboost master, we can branch out a new branch for spark 3.x just for bug fix.

wbo4958:
I have a concern about aggressively upgrading to JVM 22 and removing JNI. Spark 4.0 supports a minimum java 17, I don't know if there is an issue when running xgboost with new FFIM on the spark 4.0 and java 17 platform? I guess, there will be issues.

chyunsu3:
Yeah, for now, let's aim for JVM 17.

[...]

PR #11545, opened 2025-07-03
[jvm-packages] Fix JVM docs build, by making `LabeledPoint` a Java class

hcho3:
Unblock the CI by fixing the failing JVM docs build.

For some reason, the JavaDoc plugin is no longer able to resolve references to `LabeledPoint`, which is currently a Scala class.
To get the docs fixed, re-implement `LabeledPoint` as a Java class instead. This results into lots of boilerplate. When we transition to JVM 17+ (per #11510), we will be able to use a [record class](https://www.baeldung.com/java-record-keyword) instead.


chyunsu3:
@wbo4958 @WeichenXu123 Can I get a review on this?

trivialfis (approved):

PR #11556, opened 2025-07-11
[EM] Check whether memory policy is set.

trivialfis:
- Add topology-related utilities to check whether the memory policy is bind.
- Add document about `hwloc` and the adaptive cache.


trivialfis:
cc @rongou .

PR #11567, opened 2025-07-17
Use string view when appropriate.

trivialfis:
- Add tests for string utilities.


chyunsu3:
Is it possible to replace `xgboost::StringView` with `std::string_view`, now that we are using C++17?

trivialfis:
I did a trick with the custom string view and JSON https://github.com/dmlc/xgboost/pull/8332
Might need to port it to the STL one. Let's put it in the backlog for now.

hcho3 (approved):

PR #11576, opened 2025-07-22
[EM] More rigorous check for potential cross socket access.

trivialfis:
- Move the get cpu function.
- Use `has_normal_memory` and `has_cpu` to avoid false positives.
- Check invalid node bind.
- Workaround old CTK for the host cache.
- Cleanup version functions.


trivialfis:
cc @rongou .

rongou (approved):

PR #11591, opened 2025-07-26
[enc] Expose categories container into Python.

trivialfis:
- Make pyarrow optional.
- Add a Python class for the container.


trivialfis:
cc @rongou .

rongou (approved):

PR #11594, opened 2025-07-28
Small cleanup to the proxy dmatrix.

trivialfis:
- Remove the auto-dispatch for CUDA methods.


trivialfis:
cc @rongou .

rongou (approved):

PR #11605, opened 2025-08-01
[enc] Support training continuation with sklearn.

trivialfis:

- Support training continuation with the sklearn interface.
- Avoid duplicated re-coding with the validation dataset.
The validation dataset is already re-coded during construction. There's no need to recode it during inference.
- Cleanup the lazy imports.


trivialfis:
cc @rongou

PR #11606, opened 2025-08-04
Make CUDA lineinfo optional.

trivialfis:
Ref https://github.com/dmlc/xgboost/issues/11604 .

Before:
```
$ strip libxgboost.so 
$ du -h libxgboost.so 
445M    libxgboost.so
```

After:
```
$ strip libxgboost.so 
$ du -h libxgboost.so 
148M    libxgboost.so
```

### DIFF
--- a/cmake/Utils.cmake
+++ b/cmake/Utils.cmake
@@ -112,13 +112,11 @@ function(xgboost_set_cuda_flags target)
   if(USE_DEVICE_DEBUG)
     target_compile_options(${target} PRIVATE
       $<$<AND:$<CONFIG:DEBUG>,$<COMPILE_LANGUAGE:CUDA>>:-G;-src-in-ptx>)
-  else()
-    target_compile_options(${target} PRIVATE
-      $<$<COMPILE_LANGUAGE:CUDA>:-lineinfo>)
   endif()
 
   if(USE_NVTX)
     target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_NVTX=1)
+    target_compile_options(${target} PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:-lineinfo>)
   endif()
 
   # Use CCCL we find before CUDA Toolkit to make sure we get newer headers as intended


trivialfis:
cc @hcho3

cmake/Utils.cmake
```diff
@@ -112,13 +112,11 @@ function(xgboost_set_cuda_flags target)
   if(USE_DEVICE_DEBUG)
     target_compile_options(${target} PRIVATE
       $<$<AND:$<CONFIG:DEBUG>,$<COMPILE_LANGUAGE:CUDA>>:-G;-src-in-ptx>)
-  else()
-    target_compile_options(${target} PRIVATE
-      $<$<COMPILE_LANGUAGE:CUDA>:-lineinfo>)
```
chyunsu3, line 117: Can we create a new CMake option to enable the lineinfo? Or is it redundant given the `USE_DEVICE_DEBUG` option?

trivialfis: It's redundant when debug mode is enabled.

After this PR , it will be enabled when nvtx is enabled since the lineinfo is most relevant for profiling.

Please note that, it's not necessary to enable nvtx for profiling. I'm just grouping them under one flag.

We can remove this nvcc flag entirely and rely on the user to supply a cmake compiler flag. I don't have a strong opinion on this choice, remove the flag, split up the flag, or group the flag under nvtx.

hcho3 (approved):

chyunsu3:
I ran some experiments on my end.

CMake options: `-DUSE_DLOPEN_NCCL=ON -DPLUGIN_FEDERATED=OFF` (all SMs enabled)

Before (e4406daa58a15c4a08521e630a822bb70f1c366d):
* Uncompressed: 832 MB
* Compressed: 572 MB

After (dcf8afe6ef718dc2127961d4346381f136cd85b8):
* Uncompressed: 257 MB (70% savings)
* Compressed: 167 MB (70% savings)

PR #11610, opened 2025-08-05
Compile with CTK13.

trivialfis:
- Support CTK13 in C++, run C++ tests.
- No Python test yet. We need an update from rapids.


src/tree/gpu_hist/histogram.cu
```diff
@@ -289,22 +289,12 @@ namespace {
 constexpr std::int32_t kBlockThreads = 1024;
 constexpr std::int32_t kItemsPerThread = 8;
 constexpr std::int32_t ItemsPerTile() { return kBlockThreads * kItemsPerThread; }
+template <auto Ker>
+using DeduceKernelT = std::remove_reference_t<decltype(Ker)>;
 }  // namespace
 
 // Use auto deduction guide to workaround compiler error.
-template <typename Accessor,
```
trivialfis, line 295: For some reason, there's a sub-object warning if I use the template deduction and the generated class signature is off.

PR #11611, opened 2025-08-05
Small cleanups for predictors.

trivialfis:
- Reuse the proxy DMatrix dispatch logic for inplace-predict.
- Add block-based loop.

More is to come.


trivialfis:
cc @rongou

PR #11620, opened 2025-08-07
Support leaf prediction with QDM on CPU.

trivialfis:
- Introduce thread tmp to group related code.
- Unify the checks for the proxy DMatrix.
- Remove unneeded dispatch for the GPU predictor.
- Using the above utilities, we support the leaf prediction with QDM on CPU.

The GPU version has been implemented before.

**todos**:
- [x] https://github.com/dmlc/xgboost/pull/11621
- [x] Tests.


trivialfis:
cc @rongou

PR #11621, opened 2025-08-07
Fix CPU inplace predict performance regression.

trivialfis:
nnz is unknown with inplace predict, always use blocked prediction. This was recently introduced.


trivialfis:
cc @rongou

rongou (approved):

PR #11625, opened 2025-08-09
[mt] Implement tree view.

trivialfis:
This is similar to the `TreeView` currently used in the `gpu_predictor.cu`, but for multi-target trees. I plan to reuse this new tree view for CPU and GPU.


trivialfis:
cc @rongou .

include/xgboost/multi_target_tree_model.h
```diff
@@ -19,12 +19,50 @@
 
 namespace xgboost {
 struct TreeParam;
+/**
+ * @brief A view to the @MultiTargetTree suitable for both host and device..
```
rongou: Extra period.

include/xgboost/multi_target_tree_model.h
```diff
@@ -19,12 +19,50 @@
...
+ */
+struct MultiTargetTreeView {
+  static bst_node_t constexpr InvalidNodeId() { return -1; }
+
+  common::Span<bst_node_t const> left;
```
rongou: Are these spans all the same size? I wonder if it's worth specifying a template parameter

```
template<size_t N = std::dynamic_extent>
struct MultiTargetTreeView {

  common::Span<bst_node_t const, N> left;
  common::Span<bst_node_t const, N> right;
  ...
```

rongou (approved):

include/xgboost/multi_target_tree_model.h
trivialfis: Yeah, they have the same size. We can replace the spans with pointers if needed.

trivialfis:
@hcho3 Could you please help take a look at the JVM CI errors when you are available?

include/xgboost/multi_target_tree_model.h
trivialfis: I changed them into pointers, the extra template parameter won't be needed.

Issue #9043, opened 2023-04-17
[Roadmap] Multiple outputs.


trivialfis:
Since the XGBoost 1.6, we have been working on having multi-output support for the tree model. In 2.0, we will have the initial implementation for the vector-leaf-based multi-output model. This issue is a tracker for future development and for related discussion. The original feature request is here: https://github.com/dmlc/xgboost/issues/2087 .

## Implementation Optimization
- [ ] Use f-order for the gradient. Currently, the gradient has one column for each target but is written in C-order. The transformation takes about one-fifth of the training time.
- [ ] Use f-order for the custom objective.
- [ ] Improve array type dispatching by moving the dispatch logic from per-element to per-array. This enables us to have a more efficient custom objective interface.

## Algorithmic Optimization
We are still looking for potential algorithmic optimization for vector-leaf and here's the pool of candidates. We need to survey all available options. Feel free to share if you have ideas or paper recommendations.
- [ ] [Sketch boost](https://openreview.net/forum?id=WSxarC8t-T).
- [ ] https://arxiv.org/abs/2201.06239

## GPU Implementation
- [ ] Evaluation
- [ ] Histogram
- [ ] Prediction

## Documentation
- [ ] Derive the approximated hessian in the context of boosting trees.

## Multi-task
- [ ] Multi-task xgboost. This is not yet decided. I think it's wise to at least do some exploration before forging the rest of the implementation since we will have a very different interface if we need to consider multi-task. Related: https://github.com/dmlc/xgboost/issues/7693 .

## Features
- [ ] Tree SHAP
- [ ] Plotting
- [ ] Model text dump (JSON, txt, graphviz)
- [ ] Tree data frame.
- [ ] Categorical feature.
- [ ] Constraints
- [ ] Approx tree method
- [ ] Exact tree method
- [ ] Loss weight

## Learning to rank
We can have a ranking model to consider multiple criteria. This might require multi-task to be supported.

## Quantile regression
- [ ] l1
- [ ] quantile

## Distributed
- [ ] Dask
- [ ] PySpark
- [ ] Spark
- [ ] Flink?
- [ ] Federated

# Binding
- [ ] R
- [ ] Scala
- [x] Python
- [ ] Java
- [ ] C

# HPO
- [ ] Check compatibility with major HPO frameworks.

# Other extensions
- [ ] Sparse label
- [ ] Early stopping for each target?

## Applications
- https://arxiv.org/abs/2210.06831
- [ ] FIL

[...]

PR #11630, opened 2025-08-11
[mt] Fix demo evaluation metric.

trivialfis:
- Fix metric.
- Cleanup the CPU implementation using the new utility.


trivialfis:
cc @rongou

PR #11641, opened 2025-08-19
[EM] Retry if quantile merge failed.

trivialfis:
Workaround the C++/CUDA's lack of support for `realloc`. This can help the external memory iterator to consume more batches before running into OOM.


trivialfis:
cc @rongou

rongou (approved):

PR #11644, opened 2025-08-20
[CI] Disable spot pricing for GPU workers

hcho3:
Ensure that the GPU workers don't get interrupted by always using the On-Demand instances. Previously, we were using Spot instances, which are cheaper but have a non-zero probability of interruption.


chyunsu3:
@trivialfis FYI

PR #11651, opened 2025-08-23
[mt] Define the base score as an array.

trivialfis:
- Reuse the array definition from the quantile loss.
- The base score is still a scalar, this PR only handles the storage format.
- I added a compatibility check to return the `base_score` as a scalar in the JSON config when a single-target model is used, to avoid breaking changes. ~This may not be a good idea.~ I will remove it in future PRs as I need to clean up all related tests.

Ref https://github.com/dmlc/xgboost/issues/9043 .


trivialfis:
cc @rongou

rongou (approved):

Issue #10807, opened 2024-09-06
[CI] Upload wheel variants for CUDA 12 and 13


hcho3:
Currently, we build two wheel variants: `xgboost-cpu` (which excludes GPU code) and `xgboost` (where the GPU code targets CUDA 12.4).  In #10729, `xgboost` is found to conflict with another package using CUDA 11.

Following the practices of RAPIDS, we should distribute separate wheels targeting CUDA 11 and CUDA 12.

**Proposal**. Build four wheel variants.
* `xgboost-cpu`: excludes the GPU code
* `xgboost-cu11`: builds GPU code with CUDA 11, depends on `nvidia-nccl-cu11`.
* `xgboost`: builds GPU code with CUDA 12, depends on `nvidia-nccl-cu12`.
* `xgboost-cu12`: a stub package directing users to install `xgboost`. Something like https://pypi.org/project/cuml. The stub package can be replaced with a real one when the main package `xgboost` transitions to CUDA 13.

**Prerequisites**
* https://github.com/pypi/support/issues/4695
* Need to file another issue to increase file size for `xgboost-cu11` to 200 MiB.
* Reduce the wheel size by dropping some archs, like `sm_50`.
* #10803

hcho3:
@jameslamb Can you review my proposal? I kept the `xgboost` name, which would work for the majority of the use cases. Users would opt into `xgboost-cpu` and `xgboost-cu11` as special needs arise.

jameslamb:
@hcho3 let me start by saying I am SO SORRY it took over a month to get back to you!

Thanks for redirecting me here from #10803. I've read this and the motivating issue (#10729) now. I support this proposal.

I think the approach you've proposed is a good one... it minimizes disruption to existing users of `xgboost`, preserves `pip install xgboost` "just working" on CUDA 12 systems, and I do think it'd fix the issue from #10729.

I still think you should exploring modifying the `packager/` code for this purpose before involving `rapids-dependency-file-generator` (for the reasons I mentioned in https://github.com/dmlc/xgboost/issues/10803#issuecomment-2333181102).

---

> *I kept the `xgboost` name, which would work for the majority of the use cases.*

Thank you! This was what I was most worried about in https://github.com/dmlc/xgboost/issues/10803#issuecomment-2333181102, I think because there you described "re-naming" the package, which I assumed meant that `xgboost` might go away.

---

[...]

PR #11662, opened 2025-08-27
[CI] Build and test XGBoost with CUDA 13

hcho3:
- [x] Update CMake configuration to support building with CUDA 13
- [x] Build Docker container(s) with CUDA 13
- [x] Set up a CI workflow to build and test XGBoost with CUDA 13

Related: https://github.com/dmlc/xgboost-devops/pull/41


cmake/Utils.cmake
```diff
@@ -69,7 +69,9 @@ function(compute_cmake_cuda_archs archs)
 
   # Set up defaults based on CUDA varsion
   if(NOT CMAKE_CUDA_ARCHITECTURES)
-    if(CUDA_VERSION VERSION_GREATER_EQUAL "12.8")
+    if(CUDA_VERSION VERSION_GREATER_EQUAL "13.0")
+      set(CMAKE_CUDA_ARCHITECTURES 75 80 90 100 120)
```
chyunsu3, line 73: CUDA 13 dropped support for sm50, sm60, and sm70.

trivialfis:
Is this related to openmp CUDA backend?

chyunsu3:
> Is this related to openmp CUDA backend?

I don't think so. The patch is aimed at resolving the following error in the CMake configuration:
```
CMake Error at /usr/share/cmake-4.1/Modules/FindPackageHandleStandardArgs.cmake:227 (message):
  Could NOT find OpenMP_CUDA (missing: OpenMP_CUDA_FLAGS
  OpenMP_CUDA_LIB_NAMES)
Call Stack (most recent call first):
  /usr/share/cmake-4.1/Modules/FindPackageHandleStandardArgs.cmake:591 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-4.1/Modules/FindOpenMP.cmake:652 (find_package_handle_standard_args)
  CMakeLists.txt:264 (find_package)

-- Configuring incomplete, errors occurred!
```
The error is raised whenever CMake attempts to configure `CMAKE_CUDA_ARCHITECTURES` with 50 or 60, neither of which is supported in CUDA 13. (So the bit about OpenMP is a red herring.) Setting `CMAKE_CUDA_ARCHITECTURES` to `75, 80, 90, 100, 120` for CUDA 13 gets rid of the error.

chyunsu3:
TODO: Run some GPU pytests with CUDA 13

chyunsu3:
@trivialfis Can you review?

ops/pipeline/test-python-wheel-cuda13-impl.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+## Companion script for ops/pipeline/test-python-wheel-cuda13.sh
+
+set -eo pipefail
+# Cannot set -u before Conda env activation
+
+# Set up Conda env
+gosu root chown -R $(id -u):$(id -g) /opt/miniforge/envs /opt/miniforge/pkgs/cache
+gosu root chown $(id -u):$(id -g) /opt/miniforge/pkgs
+mamba create -y -n gpu_test python=3.12 pytest cupy scipy numpy pandas scikit-learn joblib hypothesis
```
chyunsu3, line 10: In the future, we want to move Conda envs out of the container image. See https://github.com/dmlc/xgboost-devops/issues/18#issuecomment-2791128407

I am using the CUDA 13 pipeline as a trial.

ops/pipeline/build-cuda13.sh
```diff
@@ -0,0 +1,58 @@
...
+then
+  export BUILD_ONLY_SM75=1
+else
+  export BUILD_ONLY_SM75=0
+fi
+
+set -x
+
+# Remove nvidia-nccl-cu12 from the list of Python deps
+# nvidia-nccl-cu13 is not yet available on PyPI
```
jameslamb, line 36: @hcho3 even when `nvidia-nccl-cu13` has wheels up on PyPI, what will be the plan for `xgboost` wheels to differentiate between CUDA 12 and CUDA 13?

Back in #10807 we'd talked about ways to support multiple major CUDA versions at the same time but I don't believe anything was implemented for that, right?

chyunsu3: We will need to create a new namespace `xgboost-cu13`. This can be done pretty quickly, since the wheel size is under 100 MB now.

chyunsu3: Updated proposal: https://github.com/dmlc/xgboost/issues/10807#issuecomment-3255218348

chyunsu3:
I will follow up with another PR to implement https://github.com/dmlc/xgboost/issues/10807#issuecomment-3255218348

PR #11664, opened 2025-08-28
[EM] Fix race in the iterator accessor policy.

trivialfis:
Model accuracy is not affected, but a less efficient accessor might be used due to a race condition.

- Copy the parameter.

Fix https://github.com/dmlc/xgboost/actions/runs/17274795128/job/49029841092?pr=11663


trivialfis:
cc @rongou

rongou (approved):

Issue #11440, opened 2025-05-02
xgb.train "unknown error" with External Memory and large dataset


simpsus:
My setup: Windows, python 3.11.11, xgboost 3.0.0

Expected behaviour: Chunks of external libsvm data get created by DataIter, cache file gets generated by xgboost, booster is trained

Experienced behaviour: Chunks get created, cache gets created, train crashes with error message:

`XGBoostError: [19:01:05] C:\actions-runner\_work\xgboost\xgboost\src\common\io.cc:217: Check failed: ptr != nullptr (0000000000000000 vs. nullptr) : Failed to map: .\xgb_cache-000001B4A34AD5A0.gradient_index.page. unknown error`

**Minimal Example**

The example is a few lines long. I tried to keep it as short as possible. The data is random data generated. The dataset is large (5mio x 2000) and the data chunks are defined through a binning variable (`era_col` in the iter).

The iterator is mostly taken from the tutorial page with the only addition of the chunking and saving of the chunks of the dataset to external memory (through the sklearn function). 

The local variables are then deleted and `xgb.train` is called with a few hyper parameters.

```
import numpy as np
import pandas as pd
import xgboost as xgb
from sklearn.datasets import dump_svmlight_file, load_svmlight_file
import os
from pathlib import Path
import gc

class EraChunkDataIter(xgb.DataIter):
    def __init__(
        self,
        df,
        era_col,
        cache_dir,
        target_col,
        features,
        feature_set_name="temp",
        chunk_size=100,
        ):
        self.file_paths = []
        # Get sorted unique eras and chunk them
        eras = sorted(df[era_col].unique())
        era_chunks = [
            eras[i : i + chunk_size] for i in range(0, len(eras), chunk_size)
        ]
        # Create and save each chunk
        for idx, chunk in enumerate(era_chunks):
            file_path = os.path.join(
                cache_dir,
                f"era_chunk_{idx}_{target_col}_{feature_set_name}.libsvm",
            )
            if Path(file_path).exists():
                print(
                    f"{target_col}: Saving chunk {idx}/{len(era_chunks)}: {file_path} already exists, skipping..."
                )
            else:
                print(
                    f"{target_col}: Saving chunk {idx}/{len(era_chunks)} to {file_path}"
                )
                chunk_df = df[df[era_col].isin(chunk)]
                X = chunk_df[features].values.astype(np.float32)
                y = chunk_df[target_col].values.astype(np.float32)
[...]

Issue #11419, opened 2025-04-19
Windows R CI: `#warning Please include winsock2.h before windows.h`


jakirkham:
Seeing the following warning in [Windows R CI builds]( https://github.com/dmlc/xgboost/actions/runs/14520759243/job/40740880057#step:9:41 ) (also [attached log]( https://github.com/user-attachments/files/19819938/gh_run_14520759243_job_40740880057_win_r.txt ) for posterity):

```cpp
g++  -std=gnu++17 -I"C:/R/include" -DNDEBUG -I../..//include -I../..//dmlc-core/include -I../../ -DXGBOOST_BUILTIN_PREFETCH_PRESENT=1 -DXGBOOST_MM_PREFETCH_PRESENT=1 -DXGBOOST_STRICT_R_MODE=1 -DDMLC_LOG_BEFORE_THROW=0 -DDMLC_ENABLE_STD_THREAD=0 -DDMLC_DISABLE_STDIN=1 -DDMLC_LOG_CUSTOMIZE=1    -I"c:/rtools45/x86_64-w64-mingw32.static.posix/include"   -fopenmp -DDMLC_CMAKE_LITTLE_ENDIAN=1 -pthread    -O2 -Wall  -mfpmath=sse -msse2 -mstackrealign    -c ../..//src/metric/elementwise_metric.cc -o ../..//src/metric/elementwise_metric.o
In file included from ../..//include/xgboost/collective/socket.h:23,
                 from ../../src/common/../collective/loop.h:19,
                 from ../../src/common/../collective/comm.h:14,
                 from ../../src/common/../collective/broadcast.h:8,
                 from ../../src/common/random.h:22,
                 from xgboost_custom.cc:6:
C:/rtools45/x86_64-w64-mingw32.static.posix/include/winsock2.h:15:2: warning: #warning Please include winsock2.h before windows.h [-Wcpp]
```

trivialfis:
The `windows.h` include is probably from dmlc.

trivialfis:
Should be fixed by https://github.com/dmlc/xgboost/pull/11859

PR #11667, opened 2025-08-30
Cleanup IO functions and add IO for GPU QDM.

trivialfis:
- Remove the `windows.h` from `io.h`. Closes https://github.com/dmlc/xgboost/issues/11419
- Remove the use of the dmlc temporary directory to avoid Windows-specific function calls (to avoid windows.h)
- Implement IO for the QDM, GPU-only at the moment. This is not exposed to the user and is intended solely for development purposes.
- Support memory mapping large files on Windows. Closes https://github.com/dmlc/xgboost/issues/11440


trivialfis:
cc @rongou

PR #11672, opened 2025-09-03
[test] Synthesize the California housing dataset.

trivialfis:
Avoid errors caused by internet connection https://github.com/dmlc/xgboost/actions/runs/17426781017/job/49475553158?pr=11671

Parameters are estimated using `sklearn.GaussianMixture` and `scipy.stats.norm`.


trivialfis:
cc @rongou

PR #11685, opened 2025-09-12
[doc] Add R code blocks for the intercept introduction.

trivialfis:
- Change Python examples to use classification consistently.
- Add R code blocks.
Todos
- [ ] Check existing examples with the built-in data.

Ref https://github.com/dmlc/xgboost/issues/11413


trivialfis:
cc @david-cortes I'm using the new interface here, would be great if you could help take a look.

doc/tutorials/intercept.rst
```diff
@@ -24,22 +40,51 @@ Other than the ``base_score``, users can also provide global bias via the data f
...
+
+        X, y = make_classification()
+
+        clf = xgb.XGBClassifier()
+        clf.fit(X, y)
+        # Request for raw prediction
+        m = clf.predict(X, output_margin=True)
+
+        clf_1 = xgb.XGBClassifier()
+        # Feed the prediction into the next model
```
david-cortes, line 57: Here it could leave as a small comment that passing the base margin disables the intercept.

doc/tutorials/intercept.rst
```diff
@@ -145,49 +190,100 @@ Example
...
+    .. code-tab:: py
 
-    import numpy as np
-    from scipy.special import logit
-    from sklearn.datasets import make_classification
-    from xgboost import train, DMatrix
+        import numpy as np
+        from scipy.special import logit
+        from sklearn.datasets import make_classification
+        from xgboost import train, DMatrix
```
david-cortes: Not a big thing but: most of other examples and tutorials use `xgb.train` instead of importing `train` directly.

doc/tutorials/intercept.rst
```diff
@@ -145,49 +190,100 @@ Example
...
+        X, y = make_classification(random_state=2025)
+
+    .. code-tab:: r R
+
+        library(xgboost)
+
+        # Load built-in dataset
+        data(agaricus.train, package = "xgboost")
+        X <- agaricus.train$data
+        y <- agaricus.train$label
```
david-cortes, line 213: Perhaps it could convert to factor here, so that it wouldn't be needed multiple times later.

trivialfis: It's fine, the code is split into multiple snippets, doing it explicitly helps emphasize the factor call for binary classification. I'm sure users know how to optimize it away when writing actual code.

doc/tutorials/intercept.rst
trivialfis: Added a comment for both languages.

doc/tutorials/intercept.rst
trivialfis: Changed it to `xgb.train/DMatrix`.

PR #11689, opened 2025-09-17
Destructor `~Loop()` may throw exception (violates C++ standard and guidelines)

lyriccoder:
### Problem description

The destructor of `Loop` is declared as `noexcept(false)` and directly calls `Stop()`:

```cpp
~Loop() noexcept(false) {
    this->Stop();
}
```

Inside `Stop()`, there is an unconditional rethrow of a stored exception:

```cpp
if (curr_exce_) {
    std::rethrow_exception(curr_exce_);
}
```

This means the destructor may throw exceptions.

### Why this is a problem

[C++ Standard (ISO/IEC 14882:2017, 18.1, paragraph 7)](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf)
If a destructor invoked during stack unwinding exits with an exception, `std::terminate()` is called.

In other words, if an exception is already in flight and another one is thrown from the destructor, the program will unconditionally crash.

### Coding guidelines

1. [CERT C++ [ERR55-CPP]: Do not throw an exception from a destructor.](https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL57-CPP.+Do+not+let+exceptions+escape+from+destructors+or+deallocation+functions)
2. AUTOSAR C++14 [A15-5-1]: Destructors shall not exit with an exception.

The current implementation directly violates these rules.

Expected behavior

Destructors must not propagate exceptions. They should either:

- Catch and handle/log exceptions internally, or
- Guarantee Stop() is noexcept, or
- Require explicit Stop() invocation by the user before object destruction.

This aligns with the C++ standard, CERT, and AUTOSAR guidelines and prevents hard crashes due to `std::terminate()`.

### DIFF
--- a/cmake/Utils.cmake
+++ b/cmake/Utils.cmake
@@ -118,7 +118,9 @@ function(xgboost_set_cuda_flags target)
 
   if(USE_NVTX)
     target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_NVTX=1)
-    target_compile_options(${target} PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:-lineinfo>)
+    if(NOT USE_DEVICE_DEBUG)
+      target_compile_options(${target} PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:-lineinfo>)
+    endif()
   endif()
 
   # Use CCCL we find before CUDA Toolkit to make sure we get newer headers as intended
--- a/include/xgboost/linalg.h
+++ b/include/xgboost/linalg.h
@@ -646,6 +646,17 @@ auto MakeVec(T *ptr, size_t s, DeviceOrd device = DeviceOrd::CPU()) {
 }
 
 template <typename T>
+auto MakeVec(DeviceOrd device, common::Span<T> s) {
+  return linalg::TensorView<T, 1>{s, {s.size()}, device};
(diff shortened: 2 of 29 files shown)


trivialfis:
It's a real issue, but the fix is quite off ...

lyriccoder:
> It's a real issue, but the fix is quite off ...

can u suggest the new one? I will rewrite, just do not have the idea how to fix it in another way...  @trivialfis Is the updated fix ok for you?

I added a `boolean` parameter `from_destructor` to the `Stop()` method to change behavior based on the caller:
`Result Stop(bool from_destructor = false);`

How it works:

1. Normal calls (`Stop()` or `Stop(false)`):

- Preserves original behavior
- Still throws exceptions via std::rethrow_exception()
- Maintains existing API contract for users


2. Destructor calls (`Stop(true)`):

- Converts exceptions to Result return values instead of throwing
- Uses try-catch to safely extract exception information
- Returns error details through the Result type

lyriccoder:
@trivialfis do u like the new changes?

trivialfis:
> @trivialfis do u like the new changes?

I'm guessing that you are using an LLM to write the patch. I'm not opposed to the practice, but please do try to understand the change.

PR #11691, opened 2025-09-18
Explanation for isMaximizeEvaluation.

kunaljani1100:
The function `isMaximizeEvaluation(String evalInfo, String[] evalNames, Map<String, Object> params)` is a function that is used to compute whether the evaluation metrics are maximization metrics or not. Since this is a public method that is accessible by developers, it is important for them to understand the use of this function before using this function in their model development.

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
@@ -386,6 +386,32 @@ public class XGBoost {
   }
 
   // visiable for testing
+
+  /**
+   * This methods decides whether the evaluation metrics are to be maximized or
+   * not. The decision is made in the following order:
+   * <ul>
+   *   <li>If the {@code params} object contains the key "maximize_evaluation_metrics",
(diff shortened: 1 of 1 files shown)


trivialfis (commented):
Hmm, we don't need to translate the entire function implementation into the document, those are implementation details. The function name is self-explanatory

kunaljani1100:
Removed the implementation instructions from the document.

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/XGBoost.java
```diff
@@ -386,6 +386,16 @@ private static String getMetricNameFromlog(String evalInfo, String[] evalNames)
   }
 
   // visiable for testing
+
+  /**
+   * Decides whether the evaluation metrics are to be maximized or not.
+   *
+   * @param evalInfo The evaluation log string from which the metric name is inferred.
+   * @param evalNames The names of the evaluation matrices.
+   * @param params The parameters that contain information regarding whethe the
```
trivialfis:
```suggestion
   * @param params The parameters that contain information regarding whether the
```

trivialfis (approved):

PR #11692, opened 2025-09-19
Extending documentation on custom objective

KonstantinBurkin:
This PR adds documentation to clarify how to use custom objective. Current example of custom objective is implemented using `xgb.DMatrix` and `xgb.train` (low-level version). That is not a popular way to use xgboost models. I suggest adding documentation with an example where:
- objective function is defined with numpy arrays only
- objective function is used with XGBRegressor model
- It is stated that objective for XGBRegressor usually accepts vectors in such order: (true_values, predicted_values). Whereas objective for xgb.train expects vectors in such order: (predicted_values, true_values).

### DIFF
--- a/doc/tutorials/custom_metric_obj.rst
+++ b/doc/tutorials/custom_metric_obj.rst
@@ -113,7 +113,7 @@ prediction or the other way around is important.  If you find the training error
 instead of down, this might be the reason.
 
 *****************************
-Version 2: hign-level API
+Version 2: high-level API
 *****************************
 
 .. code-block:: python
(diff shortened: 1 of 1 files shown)


trivialfis (commented):
Hi, the `dtrain` has type `DMatrix`, it contains the labels, along with other things like weight and qid, feature matrix. It's not the "true value", as it's a container for all data XGBoost consumes. Also, there's a section for the scikit-learn interface along with a demo `custom_softmax.py`.

KonstantinBurkin:
I agree with your statement concerning dtrain type. But I could not find examples with scikit-learn interface. I want to add these examples and mention that the order of arguments is different for this case (`predictions`, `y_test`), not (`dtrain`, `predictions`).

trivialfis:
> But I could not find examples with scikit-learn interface.

https://xgboost.readthedocs.io/en/stable/tutorials/custom_metric_obj.html#scikit-learn-interface Next to the section you are modifying. Not elaborate enough, but there's a section. ;-)

> order of arguments is different for this case

There shouldn't be an ordering for these parameters. They are different parameters, with different types, there's no ordering for things that are different.

KonstantinBurkin:
I found it, thanks! Then I close my PR.

PR #11697, opened 2025-09-25
Fix -Wdeprecated-redundant-constexpr-static-def violations

r-barnes:
This code pattern is deprecated as of C++17.


trivialfis:
Closing as compilation failed and no further fix is pushed.

PR #11698, opened 2025-09-25
Bump version to 3.1.0-rc1.

trivialfis:


python-package/pyproject.toml
```diff
@@ -59,7 +59,7 @@ follow_imports = "silent"
 
 [tool.pylint.main]
 ignore = ["tests"]
-extension-pkg-whitelist = ["numpy", "cuda"]
```
trivialfis, line 62: @hcho3 Is it expected that this got changed

chyunsu3: Not sure why. It should not affect the functionality, but we should go ahead and fix the `pyproject.toml.in`.

PR #11706, opened 2025-09-26
Extract the stream, event, and stream view.

trivialfis:
Aside from cleaning up the device helpers, we need to use them in CPP files in addition to cu files.


trivialfis:
cc @rongou

PR #11711, opened 2025-09-29
[jvm-packages] support inferBatchSize - allow for sparse vectors - remove synchonized 

jensgoossens-tomtom:
Hi - I'm trying to incorporate some changes based on limitations we're currently having with the scala/java package.

- InferBatchSize is hardcoded in transform, while the parameter is present - updated to use that instead.
- the internal predict still is synchronized (added in 2016) while the documentation states [that predictions are threadsafe ](https://xgboost.readthedocs.io/en/stable/prediction.html#thread-safety)

- SparseVector support: when providing a SparseVector as features, it now fails because of the internal conversion to Dense and the assertion happening. We internally have a case where the model is being trained with sparse vector optim set to true in Python and we load that model into scala/spark. This makes the transform method unusable and I would otherwise have to resort to some custom, more low level implementation. I acknowledge this is a quickfix, support can be much more elaborate, the caller providing a SparseVector now has to be aware his model should be trained with sparse vector optimization (I don't even know if that's supported in scala as I couldn't find the representative parameter that exists in Python in Scala)

### DIFF
--- a/jvm-packages/xgboost4j/src/test/java/ml/dmlc/xgboost4j/java/BoosterImplTest.java
+++ b/jvm-packages/xgboost4j/src/test/java/ml/dmlc/xgboost4j/java/BoosterImplTest.java
@@ -897,4 +897,34 @@ public class BoosterImplTest {
     Booster booster = trainBooster(trainMat, testMat);
     TestCase.assertEquals(booster.getNumFeature(), 126);
   }
+
+  @Test
+  public void testConcurrentPredict() throws InterruptedException, XGBoostError, ExecutionException, TimeoutException {
+    DMatrix trainMat = new DMatrix(this.train_uri);
+    DMatrix testMat = new DMatrix(this.test_uri);
+    Booster booster = trainBooster(trainMat, testMat);
+
+    float[][] expectedPredictions = booster.predict(testMat);
+
+    ExecutorService executor = Executors.newFixedThreadPool(10);
+    List<CompletableFuture<Void>> futures = new ArrayList<>();
+    
+    //10 threads - each calling predict 50 times
+    for (int t = 0; t < 10; t++) {
+      futures.add(CompletableFuture.runAsync(() -> {
+        try {
+          for (int i = 0; i < 50; i++) {
+            float[][] predictions = booster.predict(testMat);
+            assertArrayEquals(expectedPredictions, predictions);
+          }
+        } catch (XGBoostError e) {
+          throw new RuntimeException(e);
+        }
+      }, executor));
+    }
+
+    CompletableFuture.allOf(futures.toArray(new CompletableFuture[0]))
+      .get(30, TimeUnit.SECONDS);
+    executor.shutdown();
+  }
 }


trivialfis:
> the internal predict still is synchronized (added in 2016) while the documentation states

If memory serves, we tried to remove it a few times. It's not the XBGoost prediction method causing trouble, it's the java iterator throwing out errors when using Spark without the synchronized predict method.

jensgoossens-tomtom:
> > the internal predict still is synchronized (added in 2016) while the documentation states
> 
> If memory serves, we tried to remove it a few times. It's not the XBGoost prediction method causing trouble, it's the java iterator throwing out errors when using Spark without the synchronized predict method.

I have not encountered any errors running it with these changes

jensgoossens-tomtom:
> > the internal predict still is synchronized (added in 2016) while the documentation states
> 
> If memory serves, we tried to remove it a few times. It's not the XBGoost prediction method causing trouble, it's the java iterator throwing out errors when using Spark without the synchronized predict method.

@trivialfis I've added a test in BoosterImplTest which does concurrent calls to the predict method to verify it works concurrently.

trivialfis:
@wbo4958

wbo4958:
Hi @jensgoossens-tomtom, Can you separate this PR into 3 PRs,

1. one to batch size configurable
2. one for synchronized 
3. support sparse vector

wbo4958:
> > the internal predict still is synchronized (added in 2016) while the documentation states
> 
> If memory serves, we tried to remove it a few times. It's not the XBGoost prediction method causing trouble, it's the java iterator throwing out errors when using Spark without the synchronized predict method.

To be honest, I don't know the history.  Were the errors using spark OOM issue? since without synchronized, all tasks will predict concurrently.

wbo4958:
Hi @jensgoossens-tomtom,  For SparseVector, I don't think your change is going to work. Previously, we do the same thing as your PR, but we found that the feature will be in wrong position which resulted in wrong model when training. Similar result for prediction. @trivialfis could explain more about this limitation.

jensgoossens-tomtom:
> Hi @jensgoossens-tomtom, For SparseVector, I don't think your change is going to work. Previously, we do the same thing as your PR, but we found that the feature will be in wrong position which resulted in wrong model when training. Similar result for prediction. @trivialfis could explain more about this limitation.

True - this might only work for our specific case, where we trained the model in Python with
enable_sparse_data_optim set to true. 
(I tested this locally, but hard to mimic in the branch) which I why i also stated in the description:

>  I acknowledge this is a quickfix, support can be much more elaborate, the caller providing a SparseVector now has to be aware his model should be trained with sparse vector optimization (I don't even know if that's supported in scala as I couldn't find the representative parameter that exists in Python in Scala)

If there is no jvm support to train with sparse vectors, I might just discard this portion of the PR as it's to risky and maybe confusing

jensgoossens-tomtom:
> To be honest, I don't know the history. Were the errors using spark OOM issue? since without synchronized, all tasks will predict concurrently.

If that would be the case, users should tweak their spark setup, batch size and others for memory management instead of this synchronization imho

jensgoossens-tomtom:
Closing in favour of separate PRs:
#11745
#11746

PR #11712, opened 2025-09-29
SYCL. Fixes for the current version.

razdoburdin:
This small PR fixes bugs in sycl plugin.

Can we also merge it to release_3.1.0 branch?


trivialfis:
I will backport it, would you like to add a test?

razdoburdin:
> I will backport it, would you like to add a test?

Currently we are able to launch only sycl-cpu tests in public repos, but they don't catch 100% of failures.
This was of the failures, visible on GPU only :(

PR #11717, opened 2025-09-30
[WheelNext] Use correct variant specification

hcho3:
Use `nvidia :: cuda_version_upper_bound :: 13` to indicate `CUDA version < 13`.

In a future PR, we should build another variant wheel targeting CUDA 13.

@DEKHTIARJonathan


DEKHTIARJonathan:
I don't think you need this one btw. CUDA forward compat guarantee should allow you to not set the upper bound @hcho3.

I would test a cuda13 artifact on a cuda12 machine, but it should work.
And if it does -> I would remove this line entirely

PR #11720, opened 2025-10-02
Remove the deprecated CLI.

trivialfis:
Deprecated in 2.1, https://github.com/dmlc/xgboost/pull/9485 .

- [x] Fix the JVM packages.
- [x] Update the list of removed features.

I moved the regression data (`machine.data`) into the `demo/data/regression` directory for the JVM tests. I would like to use synthetic data for JVM tests instead of vendoring data, but that will have to be in future PRs. There are too many subtle changes.

In addition, we need to move the README file for demos into a proper Sphinx document for easier maintenance.

### DIFF
--- a/dev/prepare_jvm_release.py
+++ b/dev/prepare_jvm_release.py
@@ -144,16 +144,16 @@ def main():
 
     with cd("jvm-packages/"):
         print("====Copying resources for testing====")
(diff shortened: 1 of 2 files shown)


trivialfis:
@hcho3 Please help take a look when you are available.

PR #11725, opened 2025-10-06
SYCL. Fixes for the upcoming v3.1

razdoburdin:
This massive PR contains the main part of sycl-related fixes for the upcoming xgboost 3.1.

I have tried to make as little changes in the main part of the code as possible.

### DIFF
--- a/include/xgboost/linalg.h
+++ b/include/xgboost/linalg.h
@@ -591,13 +591,13 @@ auto MakeTensorView(Context const *ctx, Order order, common::Span<T, ext> data, 
 
 template <typename T, typename... S>
 auto MakeTensorView(Context const *ctx, HostDeviceVector<T> *data, S &&...shape) {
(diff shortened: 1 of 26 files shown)


trivialfis:
I will look into merging this and check whether it's possible to backport.

cc @hcho3

hcho3 (commented):
The changes look benign. However, I'm a bit concerned about the increased complexity for device dispatching logic in the main code base; now the CPU updater needs to call `IsSycl()`, even when the CPU updater has nothing to do with SYCL.

Is it possible to make `ctx_->Device()` decay gracefully to `cpu` when called inside a CPU updater?

razdoburdin:
That make sense.

I plan to create sycl-specific implementations for all requared parts, it would allow to remove all this checks like
`ctx->Device().IsSycl() ? DeviceOrd::CPU() : ctx->Device()`

PR #11729, opened 2025-10-08
[mt] Initial work on vector leaf for the GPU predictor.

trivialfis:
Only batch prediction with DMatrix using numeric data is implemented. Other features require some refactoring to unify the launchers.


trivialfis:
cc @rongou

PR #11731, opened 2025-10-09
fix: best_iteration missing in xgb.cv results object

kapsner:
Version 3.1.0's R-documentation of `xgb.cv` still lists "best_iteration" as element of the resulting object, however, it currently (at least when using early-stopping) just can be accessed from `cv_results$early_stop$best_iteration`.

As I was asked to update the `mllrnrs` R package to the new xgboost version, I would like to know, which "best_iteration" I can rely on in my code?

Thanks in advance,
Lorenz

### DIFF
--- a/R-package/R/xgb.cv.R
+++ b/R-package/R/xgb.cv.R
@@ -297,6 +297,9 @@ xgb.cv <- function(params = xgb.params(), data, nrounds, nfold,
     folds = folds
   )
   ret <- c(ret, cb_outputs)
+  if ("early_stop" %in% cb_names && !("best_iteration" %in% names(ret)) {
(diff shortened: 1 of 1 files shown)


trivialfis:
cc @david-cortes

david-cortes:
Thanks for pointing this out. This is a mistake in the documentation - the 'cv' object is not meant to have that attribute. It should be part of a subattribute as mentioned in the next paragraph from those same docs.

I've created a PR to correct the docs:
https://github.com/dmlc/xgboost/pull/11732

kapsner:
Thanks for the timely response. Then I will close this MR now.

PR #11732, opened 2025-10-09
[R] Correct docs about attributes of cv object

david-cortes:
closes https://github.com/dmlc/xgboost/pull/11731

This PR updates the docs of the CV object by removing an attribute which is now under a callback name instead of under the object directly, expanding the in-doc example to make it more clear.


trivialfis:
Thank you for the quick fix!

PR #11735, opened 2025-10-10
Small refactor for the GPU predictor.

trivialfis:
This is to help the larger one: https://github.com/dmlc/xgboost/pull/11730 . The objective is to phase out the `DeviceModel` and the `TreeView` that are specific to the GPU predictor. We will go through multiple PRs to achieve that as components are too interconnected.


trivialfis:
cc @rongou

PR #11740, opened 2025-10-11
Javadocs for TrackerProperties.java

kunaljani1100:
Adding the javadocs for the class `TrackerProperties.java` since there is no information available to the user on how to use this class to read the XGBoost Tracker properties.

Documentation explains to the user that the `xgboost-tracker.properties` must be present in the `resources` directory to ensure that the tracker properties are read correctly without throwing any errors.

Documentation has also been shown on how the host IP address should be declared in `xgboost-tracker.properties` so that it can be read properly using the `getHostIp()` function.


trivialfis:
Hmm, I think we can remove this class after earlier refactoring. cc @wbo4958

kunaljani1100:
Closing this PR since it this class will be removed, so documentation for this class is not necessary.

PR #11744, opened 2025-10-12
Multi target memory reallocation optimization

siqi-he:
This PR extends the memory reallocation optimization introduced in #11112 to `MultiTargetHistBuilder`.
In addition, this pr also adds a `resize()` call to both `MultiTargetHistBuilder` and `HistUpdater` as safeguards in situations when number of pages decreases. 

I did a quick benchmark using a few datasets from https://www.csie.ntu.edu.tw/~cjlin/libsvmtools/datasets/multilabel.html that aren't too small while I could fit in my ec2 instance (c7i.24xl), and the changes seems consistent:

|                      | speed up |
| :------------------- | :------- |
| BlogCatalog          | 1.01     |
| Flickr               | 1.17     |
| siam-competition2007 | 1.00     |
| mediamill            | 1.10     |
| YouTube              | 1.04     |

### DIFF
--- a/src/tree/updater_quantile_hist.cc
+++ b/src/tree/updater_quantile_hist.cc
@@ -136,7 +136,8 @@ class MultiTargetHistBuilder {
     monitor_->Start(__func__);
     std::size_t page_id{0};
     for (auto const &page : p_fmat->GetBatches<GHistIndexMatrix>(ctx_, HistBatch(this->param_))) {
-      this->partitioner_.at(page_id).UpdatePosition(this->ctx_, page, applied, p_tree);
+      this->partitioner_.at(page_id).UpdatePosition(this->ctx_, page, applied,
+                                                    p_tree->HostMtView());
       page_id++;
     }
     monitor_->Stop(__func__);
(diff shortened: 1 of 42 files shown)


siqi-he:
Failed tests seem to be unrelated to changes in this pr, pls see below:
Screenshot 2025-10-12 at 11 06 20 PM
Screenshot 2025-10-12 at 11 07 12 PM

src/tree/updater_quantile_hist.cc
```diff
@@ -151,15 +151,23 @@ class MultiTargetHistBuilder {
...
+      if (page_idx < partitioner_.size()) {
+        partitioner_[page_idx].Reset(ctx_, page.Size(), page.base_rowid,
+                                     p_fmat->Info().IsColumnSplit());
+      } else {
+        partitioner_.emplace_back(ctx_, page.Size(), page.base_rowid,
+                                  p_fmat->Info().IsColumnSplit());
+      }
+      page_idx++;
     }
+    partitioner_.resize(page_idx);
```
trivialfis, line 171: Why do you need to resize this after the loop?

siqi-he: it's meant to be a safeguard if page size decreases between different calls. if the number of pages has decreased, `LeafPartition` which iterates `partitioner_` with `for (auto const &part : partitioner_)` could touch on stale data. this wasn't a problem previously since we were clearing `partitioner_` every time we call `InitData`. 
it does seem to me the scenario above is extremely rare, but since adding a resize in most situations won't do anything, i figured it is probably ok to include it. they will be redundant if the number of pages never decreases.

trivialfis: Thank you for sharing. In that case, would you like to help add a test?

siqi-he: done. i've added tests for both `HistUpdater` and `MultiTargetHistBuilder` that stale partitioners would not write to `p_out_position` especially when we perform an update using a data with large number of pages first then followed by an update using a data with smaller number of pages. let me know if there is any issue

PR #11745, opened 2025-10-13
[jvm-packages] use inferBatchSizeParameter instead of hardcoded value in XGBoostEstimator

jensgoossens-tomtom:
Small PR to remove the configurable TODO in XGBoostEstimator.
Instead of using the hardcoded value, simply access getInferBatchSize - which has the same default as the current hardcoded value, but is now configurable.


trivialfis:
cc @wbo4958

wbo4958:
Wait, Could you be able to add a unit test for this?

jensgoossens-tomtom:
> Wait, Could you be able to add a unit test for this?
Would you like me to test the default values and setting the inferBatchSize via params works?
Testing that it actually uses different batch size internally is a bit is hard to test in a unit test based on how the code is currently structured

PR #11746, opened 2025-10-13
[jvm-packages] remove synchronized from predict

jensgoossens-tomtom:
The internal predict still is synchronized (added in 2016) while the documentation states [that predictions are threadsafe](https://xgboost.readthedocs.io/en/stable/prediction.html#thread-safety)

I've added a concurrency test to see there are no failures / predictions stay the same

### DIFF
--- a/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
+++ b/jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
@@ -363,10 +363,10 @@ public class Booster implements Serializable, KryoSerializable {
    * @return predict two dimensional array of results, where each row corresponds to a prediction.
    */
   private float[][] predict(DMatrix data,
-                                         boolean outputMargin,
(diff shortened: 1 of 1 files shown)


trivialfis:
cc @wbo4958

jvm-packages/xgboost4j/src/main/java/ml/dmlc/xgboost4j/java/Booster.java
```diff
@@ -362,7 +362,7 @@ public String evalSet(DMatrix[] evalMatrixs, String[] evalNames, IEvaluation eva
    * @param predContribs prediction feature contributions
    * @return predict two dimensional array of results, where each row corresponds to a prediction.
    */
-  private synchronized float[][] predict(DMatrix data,
+  private float[][] predict(DMatrix data,
```
wbo4958, line 365: format

jensgoossens-tomtom: Fixed

PR #11747, opened 2025-10-13
[mt] Implement model dump for all formats.

trivialfis:
- Use the new tree view.
- Remove the dmlc registry.


trivialfis:
cc @rongou

PR #11750, opened 2025-10-14
Provide device storage for scalar tree.

trivialfis:
- Cleanup various accessors.
- Use the view object for all const access.

We still use the CPU storage everywhere. Future PRs will use the device storage in the predictor.


trivialfis:
cc @rongou

PR #11752, opened 2025-10-15
Replace the device model.

trivialfis:
- Support leaf, QDM, inplace prediction for multi-target.
- Replace device model with tree-internal storage.
- Cleanup predict kernel.


trivialfis:
cc @rongou

PR #11759, opened 2025-10-20
Share the model view for CPU and GPU.

trivialfis:
- Extract the device model view and share it with CPU.
- Ensure the model is locked when pulling tree data.
- Use host device vector for storing the tree group.

This PR eliminates another data copy (tree group), and ensures thread safety when the model is trained using GPU but inference is performed with CPU.


trivialfis:
cc @rongou

Issue #11758, opened 2025-10-19
Loading old binary xgboost model in version 3.1 raises error that is hard to understand. Perhaps it should tell the user to resave the model?


ikrommyd:
Edit:

This seems to come from the fact that the original model file was in the old binary format that xgboost 3.1 removed.
I'm therefore changing this issue to a "feature request" of a better error message.

Original issue:

I have a test model that we use in ci that could be loaded fine until version 3.1.
In version 3.1, I get the following error:
```py
In [1]: import xgboost

In [2]: xgboost.__version__
Out[2]: '3.1.0'

In [3]: xgboost.Booster(model_file="/tmp/xgboost_example.xgb")
---------------------------------------------------------------------------
XGBoostError                              Traceback (most recent call last)
Cell In[3], line 1
----> 1 xgboost.Booster(model_file="/tmp/xgboost_example.xgb")

File ~/micromamba/envs/awkward-coffea/lib/python3.13/site-packages/xgboost/core.py:2019, in Booster.__init__(self, params, cache, model_file)
   2017     self.__dict__.update(state)
   2018 elif isinstance(model_file, (str, os.PathLike, bytearray)):
-> 2019     self.load_model(model_file)
   2020 elif model_file is None:
   2021     pass

File ~/micromamba/envs/awkward-coffea/lib/python3.13/site-packages/xgboost/core.py:3052, in Booster.load_model(self, fname)
   3049 if isinstance(fname, str) or is_pathlike(fname):
   3050     # assume file name, cannot use os.path.exist to check, file can be from URL.
   3051     fname = os.fspath(os.path.expanduser(fname))
-> 3052     _check_call(_LIB.XGBoosterLoadModel(self.handle, c_str(fname)))
   3053 elif isinstance(fname, bytearray):
   3054     buf = fname

File ~/micromamba/envs/awkward-coffea/lib/python3.13/site-packages/xgboost/core.py:323, in _check_call(ret)
    312 """Check the return value of C API call
    313
    314 This function will raise exception when error occurs.
   (...)    320     return value from API calls
    321 """
    322 if ret != 0:
--> 323     raise XGBoostError(py_str(_LIB.XGBGetLastError()))

XGBoostError: [14:45:01] /workspace/src/c_api/c_api.cc:1529: Check failed: str[0] == '{' (b vs. {) :
Stack trace:
[...]

PR #11760, opened 2025-10-20
Check for removed binary format.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/11758


trivialfis:
cc @hcho3

ikrommyd (approved):
Thanks! I think this can help users who try to load old models understand better what's happening!

PR #11781, opened 2025-10-29
[mt] Implement vector leaf for a decision stump on GPU.

trivialfis:
The implementation is bare minimum at the moment and has slow performance.

- Implement a basic histogram kernel.
- Implement a single-node evaluation function.


trivialfis:
cc @rongou

Issue #11763, opened 2025-10-20
Multinomial logistic intercepts applied on incorrect scale


david-cortes:
It appears the multinomial logistic objective is calculating a base score in the scale of the reponse variable (probabilities) and applying it as-is instead of transforming it to the 'margin' scale.

This shoud predict the class proportions for every observation:
```python
import numpy as np
import xgboost as xgb

y = np.r_[
    np.zeros(25),
    np.ones(25),
    np.repeat(2, 50)
]
rng = np.random.default_rng(seed=123)
X = rng.standard_normal(size=(100,10))
model = xgb.XGBClassifier(
    n_estimators=1,
    gamma=1e10,
    min_child_weight=1e10,
).fit(X, y)
model.predict_proba(X[:5])
```
```
array([[0.30450436, 0.30450436, 0.39099133],
       [0.30450436, 0.30450436, 0.39099133],
       [0.30450436, 0.30450436, 0.39099133],
       [0.30450436, 0.30450436, 0.39099133],
       [0.30450436, 0.30450436, 0.39099133]], dtype=float32)
```

It looks like the intercepts are calculated correctly if seen in the scale of the response variable:
```python
import json
json.loads(
    model.get_booster().save_config()
)["learner"]["learner_model_param"]["base_score"]
```
```
'[2.5E-1,2.5E-1,5E-1]'
```

.. but the probabilities are calculated by applynig softmax on that, without converting those numbers to the 'margin' scale:
```python
from scipy.special import softmax
softmax(
    json.loads(
        json.loads(
            model.get_booster().save_config()
        )["learner"]["learner_model_param"]["base_score"]
    )
)
```
```
array([0.30450434, 0.30450434, 0.39099132])
```

And as a result, the predictions are biased.

trivialfis:
It's a biased estimation indeed, only used for being better than the old universal 0.5 intercept. I did not add the transform as there's no exact invert for the softmax function. (can't remember what's the algebra term for functions that have $y = f^{-1} \cdot f(y)$ ) If we were to select a constant, the problem seems to be more complicated without significant benefit. For instance, we need to apply the softmax and inverse during model load, which needs to reproduce the user input somehow.

Suggestions are welcome.

david-cortes:
I suggest the approach used by glmnet where the intercepts are constrained to sum to zero:
```python
import numpy as np
from scipy.special import softmax
probs = np.array([0.25, 0.25, 0.5])
raw = np.log(probs)
[...]

PR #11786, opened 2025-10-31
Fix intercept initialization for multinomial logistic objective

david-cortes:
closes https://github.com/dmlc/xgboost/issues/11763

### DIFF
--- a/python-package/xgboost/testing/updater.py
+++ b/python-package/xgboost/testing/updater.py
@@ -144,6 +144,8 @@ def check_init_estimation(tree_method: str, device: Device) -> None:
     intercept = run_clf(X, y)
 
     np.testing.assert_allclose(np.sum(softmax(intercept)), 1.0)
+    np.testing.assert_allclose(np.sum(intercept), 0.0, atol=1e-6)
+
     assert np.all(softmax(intercept) > 0)
     np_int = (
         np.histogram(
--- a/python-package/xgboost/testing/with_skl.py
+++ b/python-package/xgboost/testing/with_skl.py
@@ -202,6 +202,7 @@ def run_intercept(device: Device) -> None:
     assert len(result) == 4
 
     assert (softmax(result) >= 0.0).all()
+    np.testing.assert_allclose(sum(result), 0.0, atol=1e-6)
     np.testing.assert_allclose(sum(softmax(result)), 1.0)
 
     # Tests for user input


david-cortes:
Not sure how to handle the GPU segfaults. Perhaps some maintainer could take over.

trivialfis:
Thank you for working on this. I will look into it.

trivialfis:
I fixed the GPU access for the PR. But still need to go through all the failing CPU tests.

src/objective/multiclass_obj.cu
```diff
@@ -209,6 +210,14 @@ class SoftmaxMultiClassObj : public ObjFunction {
...
     CHECK_GE(sum_weight, kRtEps);
     linalg::VecScaDiv(this->ctx_, intercept, sum_weight);
+
+    // Transform it back to margin
+    linalg::Vector<float> mean_intercepts;
+    CHECK_EQ(base_score->Size(), n_classes);
+    common::Mean(this->ctx_, *base_score, &mean_intercepts);
+    auto d_mean = mean_intercepts.View(this->ctx_->Device());
+    TransformKernel(this->ctx_, intercept,
+                    [=] XGBOOST_DEVICE(float v) { return log(v) - d_mean(0); });
```
david-cortes: It'd make more sense to take the mean after applying logarithms, so that they would end up summing to zero, the same way as other libraries handle it.

trivialfis: Yes, I will handle that. Thank you for pointing out.

trivialfis:
@david-cortes Please help take another look when you are available.

python-package/xgboost/testing/updater.py
```diff
@@ -141,21 +142,24 @@ def run_clf(
...
+
+    assert np.all(softmax(intercept) > 0)
     np_int = (
         np.histogram(
             y, bins=np.concatenate([np.unique(y), np.array([np.finfo(np.float32).max])])
         )[0]
         / y.shape[0]
     )
-    np.testing.assert_allclose(intercept, np_int)
+    np.testing.assert_allclose(softmax(intercept), np_int, rtol=1e-6)
```
david-cortes, line 156: Shouldn't this be comparing against `y.mean(axis=0)`?

trivialfis: y is class index

PR #11789, opened 2025-11-03
[mt] Implement partitioning for GPU.

trivialfis:
XGBoost can now grow a full vector-leaf tree using a GPU. However, the implementation is still minimal and has low performance.

I would like to explore some algorithmic optimizations before adding any more complexity to the code.

ref #9043


trivialfis:
cc @rongou

PR #11794, opened 2025-11-06
[enc] Expose the categories container to C.

trivialfis:


src/c_api/c_api.cc
```diff
@@ -1763,12 +1739,6 @@ XGB_DLL int XGBoosterDumpModelExWithFeatures(BoosterHandle handle,
   API_END();
 }
 
-/**
- * Experimental (3.1), hidden.
- */
-/**
- * See @ref XGDMatrixGetCategories
- */
 XGB_DLL int XGBoosterGetCategories(DMatrixHandle handle, char const * /*config*/,
```
trivialfis: Fix type 
```suggestion
XGB_DLL int XGBoosterGetCategories(BoosterHandle handle, char const * /*config*/,
```

chyunsu3:
Can you give me some more context on this PR? I thought the Booster object already supported exporting the categories container as Arrow arrays? How does this PR change things?

trivialfis:
> Can you give me some more context on this PR?

I hid the C functions in the last release in case of unexpected issues. It took me a while to figure out all the states, like whether one needs to re-code data during training continuation, and when input is a DMatrix. This PR exposes these functions to the C header, but they are still marked as experimental.

hcho3 (approved):

PR #11798, opened 2025-11-07
[mt] Split up gradient types for the GPU hist.

trivialfis:
- Support using reduced gradient for tree structure exploration.

The PR adds a gradient container with two different gradient types: one for tree splits and the other for leaf values. This is an optimization for vector-leaf to reduce the overhead of finding the tree structure. Currently, the interface is exposed through a custom objective function by creating a specialized tree objective.

An alternative approach would be to decouple the dimension reduction function from the objective function by introducing an additional parameter, such as `grad_reducer`. This is actually the first approach I tried.

I don't have a strong opinion on which one we should choose. The objective approach facilitates associating the reduced gradient with the prediction and label more easily, but may be less modularized.


trivialfis:
cc @rongou

PR #11800, opened 2025-11-10
Fix ranking tests with weight norm.

trivialfis:
The test might fail if the objective cache is renewed. This PR makes sure the objective and its cache is recreated for each test case, and multiple the normalization factor into the expected results.


trivialfis:
cc @rongou

PR #11801, opened 2025-11-10
[wip] Support arrow-backed pandas categorical columns.

trivialfis:
Don't merge. I'm not sure if we need to support it at the moment, as arrow-backed data is still experimental in Pandas, and the documentation on how to create a categorical feature is limited.

### DIFF
--- a/python-package/xgboost/testing/ordinal.py
+++ b/python-package/xgboost/testing/ordinal.py
@@ -335,7 +335,7 @@ def run_basic_predict(DMatrixT: Type, device: Device, tdevice: Device) -> None:
     assert_allclose(device, predt0, predt2)
 
 
-def run_cat_predict(device: Device) -> None:
+def run_cat_predict(device: Device, use_arrow: bool) -> None:
     """Basic tests for re-coding during prediction."""
     Df, _ = get_df_impl(device)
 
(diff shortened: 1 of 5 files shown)


python-package/xgboost/testing/ordinal.py
```diff
@@ -335,15 +335,32 @@ def run_basic_predict(DMatrixT: Type, device: Device, tdevice: Device) -> None:
...
+        c_list = ["cdef", "abc", "def"]
+        if use_arrow:
+            import pandas as pd
+            import pyarrow as pa
+
+            c_typ = pa.DictionaryArray.from_arrays(
+                pa.array([0, 1, 2]),
+                pa.array(["cdef", "abc", "def"], type=pa.large_utf8()),
+            )
+            c_ser = pd.Series(c_typ, dtype=pd.ArrowDtype(c_typ.type))
```
mroeschke, line 356: When constructing `c_ser` from `c_typ`, I would suggest doing something like

```
c_array = pd.arrays.ArrowExtensionArray(c_typ)
c_ser = pd.Series(c_array)
```

mroeschke:
> the documentation on how to create a categorical feature is limited

Yeah currently there isn't great support for Arrow-based categorical data in pandas. I think pandas should be able to hold the data and some common APIs across all types may be supported, but categorical specific APIs are not currently supported

python-package/xgboost/testing/ordinal.py
trivialfis: Thank you for pointing that out. I need to create a new test, this one doesn't work, as there's no `cat` attribute for arrow dictionary-backed columns (the existing test relies on this interface).

trivialfis:
> but categorical specific APIs are not currently supported

Thank you for sharing, noted. I will watch for pandas' future updates.

PR #11802, opened 2025-11-10
Cleanup linalg kernels.

trivialfis:
- Split up two different transform kernels.
- Handle the dispatching in the `linalg_op.h` header instead of the client code.
- Use thrust transform for transform kernels. Helps with memory IO.

The objective is to handle more linalg related operations like the ones used in https://github.com/dmlc/xgboost/pull/11786 .


trivialfis:
cc @rongou

rongou (approved):

Issue #11804, opened 2025-11-11
Dask worker removal causes crash after training completes


adamreeve:
I consistently get a crash after the final training iteration when using Dask. This is reproducible using the example code at https://xgboost.readthedocs.io/en/stable/python/dask-examples/gpu_training.html

When running this, I get:
```
Using DaskQuantileDMatrix
[20:47:31] Task [xgboost.dask-0]:tcp://127.0.0.1:39803 got rank 0
[20:47:31] Task [xgboost.dask-1]:tcp://127.0.0.1:41247 got rank 1
[20:47:33] [0]  train-rmse:2.84356
[20:47:33] [1]  train-rmse:2.79851
[20:47:33] [2]  train-rmse:2.75492
[20:47:33] [3]  train-rmse:2.71145
2025-11-11 20:47:35,620 - distributed.scheduler - WARNING - Removing worker 'tcp://127.0.0.1:39803' caused the cluster to lose already computed task(s), which will be recomputed elsewhere: {('uniform-from-dask-array-1df0a43faf0d656b0d4794e39ba42714', 4), 'dict-5ae98066-0603-498d-975e-c7a9fabec6fc', 'dict-4169c838-7a09-479a-b53a-3435de0a492a', 'dict-b6d54058-0c05-4645-8461-419017411166', 'dict-7d911c65-146a-4dc0-bff2-27444a132580', ('from-dask-array-1df0a43faf0d656b0d4794e39ba42714', 7), 'dict-5f48dca7-2b8e-46a8-858d-7f804f071678', 'dict-5ccb8d20-6cf0-4c16-87e3-a59cb23c51ad', 'dict-13825add-3269-45e2-b57b-7fb58bac9560', ('uniform-from-dask-array-1df0a43faf0d656b0d4794e39ba42714', 1), 'dict-881f8ed8-2aef-4d71-8966-c92beac004c1'} (stimulus_id='handle-worker-cleanup-1762894055.6203396')
2025-11-11 20:47:35,621 - distributed.scheduler - ERROR - Removing worker 'tcp://127.0.0.1:39803' caused the cluster to lose scattered data, which can't be recovered: {'Booster-317a20c46f8943fbac42423eeda8e498'} (stimulus_id='handle-worker-cleanup-1762894055.6203396')
Traceback (most recent call last):
  File "/home/adam/xgb-dask/main.py", line 83, in <module>
    from_ddqdm = using_quantile_device_dmatrix(client, X, y).compute()
[...]

PR #11806, opened 2025-11-12
Fix loading nccl 2.28.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/11804

The latest nccl pypi package no longer has the `__file__` attribute in its modules (it no longer has modules). We will use the `__path__` from the namespace instead.


trivialfis:
cc @rongou

PR #11808, opened 2025-11-12
Make block_size of BuildHistKernel adaptive

razdoburdin:
Current version of xgboost utilizes fixed block_size = 256 for hist building.

This PR make this value an adaptive function of model parameters and CPU cache size. The change is important mostly for ColsWiseBuildHistKernel and demonstrates up to 2x speed-up for epsilon dataset.


trivialfis:
Thank you for the optimizations! The code looks reasonable, but please add comments when the PR is ready for review. (and ping me).

src/common/cache_manager.cc
```diff
@@ -0,0 +1,104 @@
+/**
+ * Copyright 2021-2025, XGBoost Contributors
+ */
+#include "cache_manager.h"
+#include "xgboost/logging.h"
+
+#include <cstdint>     // for uint64_t
+
+#if defined(__x86_64__)
+void RunCpuid(uint32_t eax, uint32_t ecx, uint32_t* abcd) {
```
Vika-F: It would be safer in this variant:
```suggestion
void RunCpuid(uint32_t eax, uint32_t ecx, uint32_t (& abcd)[4]) {
```

src/common/cache_manager.cc
```diff
@@ -0,0 +1,104 @@
...
+#define _CPUID_GET_LINE_SIZE(__ebx) (__extract_bitmask_value(__ebx /*11:0*/, 0x7ffU, 0) + 1)
+
+#define _CPUID_GET_PARTITIONS(__ebx) (__extract_bitmask_value(__ebx /*21:11*/, 0x3ff800U, 11) + 1)
+
+#define _CPUID_GET_WAYS(__ebx) (__extract_bitmask_value(__ebx /*31:22*/, 0xffc00000U, 22) + 1)
+
+#define _CPUID_CACHE_INFO 0x4U
+
+bool GetCacheInfo(int cache_num, int * type, int * level, int64_t * sets,
+                  int * line_size, int * partitions, int * ways) {
```
Vika-F: References are more suitable here, as those values cannot be NULL.
```suggestion
bool GetCacheInfo(int cache_num, int & type, int & level, int64_t & sets,
                  int & line_size, int & partitions, int & ways) {
```

src/common/cache_manager.cc
```diff
@@ -0,0 +1,104 @@
...
+  bool trust_cpuid = !under_hypervisor;
+  return trust_cpuid;
+}
+
+constexpr int kCpuidTypeNull = 0;
+constexpr int kCpuidTypeData = 1;
+constexpr int kCpuidTypeInst = 2;
+constexpr int kCpuidTypeUnif = 3;
+
+bool DetectDataCaches(int cache_sizes_len, int64_t* cache_sizes) {
```
Vika-F: Similar suggestion here, and in SetDefaultCaches:
```suggestion
bool DetectDataCaches(std::array<int64_t, CacheManager::kMaxCacheSize> & cache_sizes) {
```

src/tree/hist/histogram.h
```diff
@@ -238,8 +241,41 @@ common::BlockedSpace2d ConstructHistSpace(Partitioner const &partitioners,
...
+    //Prefetch
+    l1_row_foot_print += 2 * sizeof(float);
+    std::size_t idx_bin_size = n_columns * sizeof(uint32_t);
+
+    bool hist_fit_to_l1 = (hist_size + offsets_size + idx_bin_size) < usable_l1_size;
+
+    std::size_t occupied_space = (hist_fit_to_l1 ? hist_size : 0) + offsets_size + idx_bin_size;
+    space_in_l1_for_rows = usable_l1_size > occupied_space ? usable_l1_size - occupied_space : 0;
+  }
+  std::size_t block_size = std::max<std::size_t>(1, space_in_l1_for_rows / l1_row_foot_print);
```
Vika-F: Previously block_size was always 256 rows, which is quite large. And now it is 1 row in case no more rows fit into L1. Won't this change affect the performance in the case when there are no enough space for rows in L1?
Should it be max(256, space_in_l1_for_rows / l1_row_foot_print) ?
Or maybe L2 size should be used to calculate the block_size?

razdoburdin: I think, cacheline_size / (2 * sizeof(float) = 8, would be the best value in this case. Using L2 would result to a huge block_size (~1e4-1e5) and produce potential underutilization of CPU cores (blocks are processed in parallel, and if blocks a very big, than some cores would be out of job).

src/common/cache_manager.cc
razdoburdin: non-const references aren't accepted by linter

razdoburdin:
Hi @trivialfis , this PR is ready for review.

trivialfis (commented):
Would you like to explain the cache info in code comments? Also, the construction of hist space on how and why it depends on the cache size?

razdoburdin:
> Would you like to explain the cache info in code comments? Also, the construction of hist space on how and why it depends on the cache size?

done

src/common/cache_manager.cc
```diff
@@ -0,0 +1,109 @@
...
+bool DetectDataCaches(int64_t* cache_sizes) {
+  int cache_num = 0, cache_sizes_idx = 0;
+  while (cache_sizes_idx < kMaxCacheSize) {
+    int type, level, line_size, partitions, ways;
+    int64_t sets, size;
+    bool trust_cpuid =
+      GetCacheInfo(cache_num++, &type, &level, &sets, &line_size, &partitions, &ways);
+    if (!trust_cpuid) return trust_cpuid;
+
+    if (type == kCpuidTypeNull) break;
```
trivialfis: Is the `cache_sizes[idx]` valid if we break here?

src/tree/hist/histogram.h
```diff
@@ -249,6 +320,18 @@ common::BlockedSpace2d ConstructHistSpace(Partitioner const &partitioners,
...
+  common::CacheManager cache_manager_;
+
+  bool ReadByColumn(const GHistIndexMatrix &gidx, bool force_read_by_column) const {
+    if (force_read_by_column) return true;
+
+    auto nbins = gidx.cut.Ptrs().back();
+    size_t hist_size = 2 * sizeof(double) * nbins;
+    const bool hist_fit_to_l2 = 0.8 * cache_manager_.L2Size() > hist_size;
+
+    bool read_by_column = !hist_fit_to_l2 && gidx.IsDense();
```
trivialfis, line 343: Could you please add a comment on how this decision is made?

src/tree/hist/histogram.h
```diff
@@ -238,8 +241,76 @@ common::BlockedSpace2d ConstructHistSpace(Partitioner const &partitioners,
...
+    * Rows do not need to remain in L1 across iterations, but choosing a
+    * reasonable block_size allows the histogram buffer and offsets to stay in L1,
+    * which gives a small performance benefit.
+    */
+
+    /* First step: estimate the size of the histogram and the offsets vector. */
+    std::size_t n_bins = gidx.cut.Ptrs().back();
+    std::size_t n_columns = gidx.cut.Ptrs().size() - 1;
+    bool any_missing = !gidx.IsDense();
+    std::size_t hist_size = 2 * sizeof(double) * n_bins;
```
trivialfis: Consider using `sizeof(GradientPair)` and `sizeof(GradientPairPrecise)` instead of `sizeof(float) * 2` (for all sizeof calls in this PR).

src/tree/hist/histogram.h
```diff
@@ -238,8 +241,76 @@ common::BlockedSpace2d ConstructHistSpace(Partitioner const &partitioners,
...
+  if (read_by_column) {
+   /* In this case, an accurate block_size estimate is performance-critical.
+    * For column-wise histogram construction, each column is processed over the
+    * same block of rows. If the block fits in L1, the row data are loaded once
+    * and reused across all columns; otherwise, the cache must be refilled for
+    * each column.
+    */
+
+    /* First step: determine whether one histogram column fits into L1.
+    * The maximum number of elements in a column is 2^8, 2^16, or 2^32,
```
trivialfis: Could you please elaborate on what it means to be the maximum number of elements in a (histogram) column? I thought that's the number of histogram bins?

src/common/cache_manager.cc
razdoburdin: In this case we use default values from `SetDefaultCaches`

src/tree/hist/histogram.h
razdoburdin: you are right, bins is a correct term. I have fixed the description.

src/tree/hist/histogram.h
razdoburdin: done

src/tree/hist/histogram.h
razdoburdin: done

src/tree/hist/histogram.h
trivialfis: Thank you for updating the comments. It's still not clear to me what it means to have "maximum number of bins" in a column. So, what happens if I specify the training parameter `max_bin=53`?

razdoburdin: You are right, it is better to use max_bin in this case, otherwise the estimation would be too conservative. I have updated the code.

src/common/cache_manager.cc
trivialfis: If this loop breaks, the function returns true, then this line does not execute:
```
  if (!trust_cpuid) SetDefaultCaches();
```
are we using the default values?

src/tree/hist/histogram.h
```diff
@@ -238,8 +242,74 @@ common::BlockedSpace2d ConstructHistSpace(Partitioner const &partitioners,
...
+    * For column-wise histogram construction, each column is processed over the
+    * same block of rows. If the block fits in L1, the row data are loaded once
+    * and reused across all columns; otherwise, the cache must be refilled for
+    * each column.
+    */
+
+    /* First step: determine whether one histogram column fits into L1.
+     * Note: column-wise kernel is used for dense data only.
+     */
+    std::size_t hist_col_size = 2 * sizeof(double) * max_bin;
```
trivialfis: https://github.com/dmlc/xgboost/pull/11808#discussion_r2527723071

src/tree/hist/histogram.h
trivialfis: Hmm, I made the comment for line 286, which is not done.

src/tree/hist/histogram.h
trivialfis: > You are right, it is better to use max_bin in this case, otherwise the estimation would be too conservative

I didn't make any suggestion? I was curious about the constraint and where the numbers `2^8, 2^16` originate or what they are for.

razdoburdin: ok, I have assumed that max_bin would be 2^8, 2^16 or 2^32 as a limit cases for BinTypeSize = 1, 2 or 4. It is better to use the exact max_bin value to have more accurate estimation.

src/tree/hist/histogram.h
razdoburdin: fixed

src/tree/hist/histogram.h
razdoburdin: fixed

src/common/cache_manager.cc
razdoburdin: if the loop breaks, it means CPU doesn't have all 4 cache levels, but all values being already read are correct. I have made some refactoring to make this part more clear.

src/common/cache_manager.cc
```diff
@@ -0,0 +1,110 @@
...
+  return true;
+}
+#endif  // defined(__x86_64__)
+
+namespace xgboost::common {
+
+/* Detect CPU cache sizes at runtime using CPUID.
+ * CPUID cannot be used reliably on:
+ * 1. non-x86_64 architectures
+ * 2. virtualized environments (CPUID may report incorrect cache sizes)
```
trivialfis: May I ask, does this pretty much rule out most of cloud instances?

razdoburdin: yes :(
but we don't have any good way to find real cache sizes in this case.

src/common/cache_manager.cc
```diff
@@ -0,0 +1,110 @@
...
+  *ways              = _CPUID_GET_WAYS(ebx);
+
+  bool trust_cpuid = !under_hypervisor;
+  return trust_cpuid;
+}
+
+constexpr int kCpuidTypeNull = 0;
+constexpr int kCpuidTypeData = 1;
+constexpr int kCpuidTypeInst = 2;
+constexpr int kCpuidTypeUnif = 3;
```
trivialfis: `kCpuidTypeData`, `kCpuidTypeUnif` are not used.

trivialfis:
@razdoburdin I shared a WIP benchmark result with your GitHub email address `dmitry.razdoburdin@intel.com`, please take a look when you are available. I highlighted the regression cases.

razdoburdin:
> @razdoburdin I shared a WIP benchmark result with your GitHub email address `dmitry.razdoburdin@intel.com`, please take a look when you are available. I highlighted the regression cases.

haven't received them. Could you please duplicate to d.razdoburdin@gmail.com ?

trivialfis:
[CPU l1 hist.csv](https://github.com/user-attachments/files/23650286/CPU.l1.hist.csv)
I will upload a CSV instead. ;-(

razdoburdin:
> [CPU l1 hist.csv](https://github.com/user-attachments/files/23650286/CPU.l1.hist.csv) I will upload a CSV instead. ;-(

Could you also share HW details? Is it a virtualized environment? ARM or x86?

trivialfis:
> Could you also share HW details? Is it a virtualized environment? ARM or x86?

Just my personal desktop:
- AMD Ryzen 9 7900X3D 12-Core Processor. I can find an Intel server if needed, but I need to use machines from work.
- Bare metal. No VM or container.

razdoburdin:
> * AMD Ryzen 9 7900X3D 12-Core Processor. I can find an Intel server if needed, but I need to use machines from work.
> * Bare metal. No VM or container.

I see. 7900X3D has great L3 per core capacity, I didn't take into account before. I have upgraded the code.

trivialfis:
| branch       | n_samples_per_batch | n_features | n_batches | sparsity | size (GB) | max_bin | n_rounds | rmse        | DMatrix-Train | Train       |
|--------------|---------------------|------------|-----------|----------|-----------|---------|----------|-------------|---------------|-------------|
| L1 hist      | 1048576             | 245        | 32        | 0        | 30.625    | 257     | 128      | 14.84709263 | 89.5329814    | 438.3597314 |
| L1 hist + L3 | 1048576             | 245        | 32        | 0        | 30.625    | 257     | 128      | 14.84709263 | 87.28941536   | 422.529346  |
| master       | 1048576             | 245        | 32        | 0        | 30.625    | 257     | 128      | 14.84709263 | 89.73098493   | 284.3891058 |

razdoburdin:
> branch	n_samples_per_batch	n_features	n_batches	sparsity	size (GB)	max_bin	n_rounds	rmse	DMatrix-Train	Train
> L1 hist	1048576	245	32	0	30.625	257	128	14.84709263	89.5329814	438.3597314
> L1 hist + L3	1048576	245	32	0	30.625	257	128	14.84709263	87.28941536	422.529346
> master	1048576	245	32	0	30.625	257	128	14.84709263	89.73098493	284.3891058

ok, i hope I have found the reason.
AMD stores the topology in another leaf. I have added the vendor switch. I have also removed under_hypervisor flag, as far as current realization automatically fallback to default values if virtualized environment is unable to report cache size. 

Unfortunately I don't have any 7900X3D to verify the perf by myself.

trivialfis:
Excellent! I can confirm that the regression is now fixed. Would you like to share your benchmark results for the datasets that you have tested?

I will run some more tests to deliberately disable cpuid (hence using the default value), just in case there's a regression for cloud users.

>  as far as current realization automatically fallback to default values if virtualized environment is unable to report cache size.

Could you please elaborate on how the current code detects the case and performs fallback? Is it guaranteed that under VM, `if (type == kCpuidTypeNull) break` is true? Asking since previously there was an explicit check for VM; now that the check is gone, it looks like the loop will finish all the way down to the bottom cache level. (I should probably create a VM to test this ....)

trivialfis:
Hi, I will be on holiday next week. Response might be slow.

I don't want to make this PR difficult, and the CPU implementation could benefit greatly from optimizations. Please feel free to create specialized code for targeted sets of CPUs, make sure the specialization is well-scoped, say within 20 lines of code, and doesn't regress other CPUs.

razdoburdin:
> Excellent! I can confirm that the regression is now fixed. Would you like to share your benchmark results for the datasets that you have tested?

here are benchmarks results, I have made on my 56-cores machine. `epsilon` is the only case with ColWiseHistBuild kernel in use.
image

razdoburdin:
Could you please elaborate on how the current code detects the case and performs fallback? Is it guaranteed that under VM, `if (type == kCpuidTypeNull) break` is true? Asking since previously there was an explicit check for VM; now that the check is gone, it looks like the loop will finish all the way down to the bottom cache level. (I should probably create a VM to test this ....)

All cache sizes are initialized by `-1` by default.
In case some cache level doesn't exist (or VM is configured not to report it) the condition `(type == kCpuidTypeNull)` would break the execution and some elements of `cache_size` array would still be equal to `-1`. The getters in `CacheManager` class check if the corresponding element is `-1` and return the default value in these case.

razdoburdin:
hi @trivialfis,

what is your opinion about this?

trivialfis:
Running some tests (picking up from the last week). Out of curiosity, is the Linux `/sys/devices/system/cpu/cpu0/cache/index2/size` a reliable source of information? In addition, what happens to CPUs with efficient/performance cores, or what happens with CPUs that have different dies (for example, amd 3d cache)?

razdoburdin:
> Out of curiosity, is the Linux `/sys/devices/system/cpu/cpu0/cache/index2/size` a reliable source of information?

In the best of my understanding it utilizes similar to cpuid by kernel loading, so the results should be the same if the kernel is new.

> In addition, what happens to CPUs with efficient/performance cores

cpuid would report L1/L2 corresponding to the logical core, that executes the cpuid. So not 100% optimal values for all cores.

>, or what happens with CPUs that have different dies (for example, amd 3d cache)?

it would report a total L3 size per CPU (for example in case of 2-dies with 16MB each, the reported value would be 32MB).

trivialfis:
@razdoburdin Could you please help take a look into the sycl error? Seems broken by dependency updates.

razdoburdin:
> @razdoburdin Could you please help take a look into the sycl error? Seems broken by dependency updates.

yes, I will take a look

razdoburdin:
> @razdoburdin Could you please help take a look into the sycl error? Seems broken by dependency updates.

sycl build is broken by this PR: https://github.com/dmlc/xgboost/pull/11802
I will make a separate fix

trivialfis:
？ The CI was passing in that PR?

> I will make a separate fix

Thank you, let me know if I can help

Issue #11811, opened 2025-11-15
Early stopping callback results in indeterministic behavior in distributed XGBoost training


unw9527:
First of all, thanks for open sourcing this great project. I am doing distributed XGBoost training with _early stopping rounds_ and a customized callback X. I expect callback X to be called for the **same** number of times across all workers, but with early stopping, different workers call X different times.

Further investigation shows that the function param `early_stopping_rounds` will create an early stopping callback which is added to the XGBoost callback list [here](https://github.com/dmlc/xgboost/blob/9255b9cd5b8964c95db5a5b987715db5b611e6da/python-package/xgboost/training.py#L188-L189). Then this callback list (with my callback X also in it) will be turned into a set [here](https://github.com/dmlc/xgboost/blob/9255b9cd5b8964c95db5a5b987715db5b611e6da/python-package/xgboost/callback.py#L163), and this set will be iterated after each iteration [here](https://github.com/dmlc/xgboost/blob/9255b9cd5b8964c95db5a5b987715db5b611e6da/python-package/xgboost/callback.py#L268).

However, this is problematic because the order of a Python set is _inherently unordered_ and the iteration uses `any()` to short-circuit as soon as it encounters the first True value. Then when early stopping callback returns True, callback X **may or may not** have been called, resulting in inconsistent behavior across different workers.

IMO iterating through a set is error-prone. If we want to remove duplicates we should use an OrderedDict to preserve the order.

trivialfis:
Thank you for investigating into the issue and sharing it! I will create a fix.

unw9527:
Hi @trivialfis, do we know when this fix will be released?

trivialfis:
> Hi @trivialfis, do we know when this fix will be released?

I don't have an ETA yet, but I hope to publish it in this month, if nothing else goes wrong.

trivialfis:
Keeping this open. We might want to raise an error when there are duplicated callbacks as suggested by @unw9527

trivialfis:
> I don't have an ETA yet, but I hope to publish it in this month, if nothing else goes wrong.

3.1.2 is now available on pypi.

unw9527:
> 3.1.2 is now available on pypi.

Thank you!

PR #11812, opened 2025-11-17
Fix ordering of Python callbacks.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/11811

### DIFF
--- a/tests/python/test_callback.py
+++ b/tests/python/test_callback.py
@@ -1,7 +1,7 @@
 import os
 import tempfile
 from collections import namedtuple
-from typing import Tuple, Union, Optional
+from typing import Optional, Tuple, Union
 
 import numpy as np
 import pytest


trivialfis:
cc @rongou

rongou (approved):

python-package/xgboost/callback.py
```diff
@@ -160,7 +160,7 @@ def __init__(
         output_margin: bool = True,
         is_cv: bool = False,
     ) -> None:
-        self.callbacks = set(callbacks)
+        self.callbacks = list(dict.fromkeys(callbacks))
```
unw9527, line 163: Curious, is `dict.fromkeys()` here just for removing duplicates? Why do we even allow duplicates in the callback list? IMO it should be better just throw an error when there are duplicates instead of silently removing them.

trivialfis: > Curious, is dict.fromkeys() here just for removing duplicates? 

Yes, it's just to remove duplicates.

I agree. I don't recall the exact case, but when I rewrote the callback interface years ago, some other parameter tuning methods unnecessarily appended their callbacks to a list repeatedly.

I can try to raise an error in the next major release, this patch is supposed to go into a patch release for bug fix only.

unw9527: Thanks for explanation. When will this patch release be out?

trivialfis: Let's continue the discussion in the original thread https://github.com/dmlc/xgboost/issues/11811#issuecomment-3544185264 .

PR #11814, opened 2025-11-17
[mt] Split up the tree model weights, support extmem.

trivialfis:
- Store base weight and leaf weight differently, as the size of base weight can be smaller.
- Basic tests for external memory.

Ref https://github.com/dmlc/xgboost/issues/9043 .

- [x] MemCostBytes
- [x] Copy
- [x] Document about the storage format.


trivialfis:
cc @rongou

Issue #11813, opened 2025-11-17
Parameter `enable_categorical` not loaded properly from json file


printsg:
Hi, 

I encountered an issue when trying to read XGBClassifier from json file. The model originally trained with `enable_categorical=True`, subsequently stored to json and loaded has this parameter set `False`. 

Below is an example to demonstrate the issue.

```
import pandas as pd
import numpy as np
import xgboost as xgb

# 1. Create Synthetic Data
np.random.seed(42)
N = 100
df = pd.DataFrame({
    # The feature we want XGBoost to treat as categorical
    'city': np.random.choice(['London', 'Paris', 'Tokyo', 'New York'], size=N),
    'age': np.random.randint(20, 60, size=N),
})

# Create a binary target that depends on the city
df['target'] = np.where(df['city'].isin(['London', 'New York']), 1, 0)

# Convert the column directly to the 'category' dtype
df['city'] = df['city'].astype('category')

# Separate features (X) and target (y)
X = df.drop('target', axis=1)
y = df['target']

# Initialize XGBoost Classifier with native categorical support
xgb_model = xgb.XGBClassifier(
    enable_categorical=True,
    tree_method='hist',
    eval_metric='logloss',
    random_state=42
)

# Train the model on the full dataset
xgb_model.fit(X, y)
print(f"xgb_model.enable_categorical={xgb_model.enable_categorical}")

# Save model to json
xgb_model.save_model("categorical-model.json")

# Load model from json
xgb_model_2 = xgb.XGBClassifier()
xgb_model_2.load_model("categorical-model.json")
print("Loading from json...")
print(f"xgb_model.enable_categorical={xgb_model_2.enable_categorical}")
```

`xgboost.__version__ == 3.1.1'

trivialfis:
Thank you for raising the issue. I will look into it.

PR #11816, opened 2025-11-18
Infer the `enable_categorical` during model load.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/11813


trivialfis:
cc @rongou

PR #11820, opened 2025-11-19
[mt] Cleanup partitioners.

trivialfis:
- Share the initialization code between sc tree and mt tree updaters.
- Share the temporary buffer between partitioners.
- Implement external memory support for reduced gradient.


trivialfis:
cc @rongou

PR #11824, opened 2025-11-19
[CI] Disable spot instances for all workers

hcho3:
Spot instances are prone to interruptions. Disable them so that our CI jobs don't get interrupted randomly.

cc @trivialfis


trivialfis:
Thank you for this!

PR #11828, opened 2025-11-20
[ci] Avoid fetching GTest with CMake.

trivialfis:
Prevent failure: https://github.com/dmlc/xgboost/actions/runs/19546987883/job/55969434054

### DIFF
--- a/.github/workflows/cuda13.yml
+++ b/.github/workflows/cuda13.yml
@@ -1,0 +1,87 @@
+name: XGBoost CI (CUDA 13)
+
+on: [push, pull_request]
+
+permissions:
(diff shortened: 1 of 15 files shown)


chyunsu3:
Do you have a reason to believe that fetching gtest from Apt would lead to less flakiness than fetching it from GitHub?

trivialfis:
No reason. But in general, having it handled by a dependency manager makes things easier. For instance, we can cache them in a docker image if needed. Also, we don't have to build from source everytime. I would like to replace the dmlc gtest whenever the chance arises.

chyunsu3:
If so, let's use Conda to install gtest. This way, we can also use it for MacOS and Windows as well.

PR #11829, opened 2025-11-21
[mt] Support learning rate.

trivialfis:
- Use the learning rate for leaf weights.
- Hide the MT implementation behind the debug flag.


trivialfis:
cc @rongou

PR #11841, opened 2025-12-03
Small optimization for the transform kernel.

trivialfis:
Tag the input copyable for newer cccl versions to let thrust stage the data.


trivialfis:
cc @rongou

PR #11847, opened 2025-12-07
[R] Fix memory safety issues

david-cortes:
Fixes a few issues detected by rchk.

@trivialfis You'll likely be asked to resubmit a fixed version of xgboost to CRAN after a few days. This PR will need to be backported if there isn't any 3.2 release by then. There might be more required changes from other checks, so let's wait for their email before resubmitting.


trivialfis:
Thank you for the quick fix!

trivialfis:
> There might be more required changes from other checks, so let's wait for their email before resubmitting.

Sounds good! Thank you for carrying it to the finish line.

PR #11848, opened 2025-12-07
[R] Remove calls to R's global RNG

david-cortes:
At some point, xgboost was using R's random number generators from C++, but those calls were removed in favor of pre-drawing a seed from the  R side to pass to C++ random number generators when needed.

But some calls to the RNG state were still left in the codebase. This PR removes them since they aren't needed anymore.


trivialfis:
Thank you for cleaning up the codebase. Could you please help fix the lint error?

PR #11855, opened 2025-12-10
[mt] Support building histogram with shared memory.

trivialfis:
- Use shared memory if possible.
- Build histogram for multiple nodes by partitioning the grid at the block level.
- Remove gradient transpose.


trivialfis:
cc @rongou

PR #11862, opened 2025-12-11
[mt] Implement prediction cache.

trivialfis:
- Implement the prediction cache for vector leaf.
- Skip histogram build for leaf nodes.


trivialfis:
cc @rongou

PR #11863, opened 2025-12-12
[CI] Set up a nightly pipeline to test XGBoost with latest CCCL/RMM

hcho3:
Example CI run: https://github.com/dmlc/xgboost/actions/runs/20156973678, where the pipeline detects breakage with CCCL 3.2.0 RC1 and RMM 26.02 nightly.

There are two jobs in the pipeline:
* `test-latest-cccl` builds XGBoost with the latest version of CCCL (RC included). RMM is turned off for this job. (This is to avoid incompatibilities between latest CCCL and latest RMM.)
* `test-latest-rmm` builds XGBoost with the latest development version of RMM. The job uses the same CCCL version as rest of Rapids.

cc @csadorf @jakirkham @bdice 

The pipeline will run on a nightly basis. In addition, it's also possible to manually trigger the pipeline via the workflow dispatch.

@trivialfis Eventually this testing will be moved to CCCL and cuML for the sake of better visibility. For now, we host it in the XGBoost repo. It is set up as a nightly pipeline so that it doesn't block daily development.

### DIFF
--- a/ops/pipeline/query-latest-cccl.sh
+++ b/ops/pipeline/query-latest-cccl.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+## Query latest version of CCCL using GitHub CLI
+## Note: RC version may be selected if available
+
+set -euo pipefail
+
+tmpfile="$(mktemp /tmp/abc-script.XXXXXX)"
+cat >"$tmpfile" <<EOL
+import fileinput
+
+from packaging.version import InvalidVersion, Version
+
+versions = []
+for e in fileinput.input():
+    try:
+        tag = e.strip()
+        versions.append((tag, Version(tag)))
+    except InvalidVersion:
+        pass
+print(max(versions, key=lambda x : x[1])[0])
+EOL
+export CCCL_VERSION=$(
+  gh api repos/NVIDIA/cccl/tags --paginate --jq '.[].name' | python3 "$tmpfile"
+)
+echo "--- Latest CCCL version: ${CCCL_VERSION}"
+rm "$tmpfile"
--- a/.github/workflows/cccl_nightly.yml
+++ b/.github/workflows/cccl_nightly.yml
@@ -1,0 +1,71 @@
+name: Test XGBoost with latest CCCL and RMM
+on:
+  workflow_dispatch:  # Can be manually triggered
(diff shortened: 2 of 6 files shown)


bdice:
Great. Both of the failures seen right now can be fixed. The thrust::pair issues indicate a missing header, and you should migrate to cuda::std::pair as well. The RMM issues are from a movement in headers, to `rmm/mr/per_device_resource.hpp`. Happy to review fixes for those separately.

Is it possible to run these as optional PR CI jobs too (failure is okay)?

chyunsu3:
@trivialfis Hi, would you be okay if we added optional CI jobs to pull requests to check for CCCL compatibility? The CI job will be hosted as a separate workflow apart from other jobs.
Implication: we often won't see the "green check marks" when merging pull requests. 
Any suggestion would be welcome.

trivialfis:
> Eventually this testing will be moved to CCCL and cuML for the sake of better visibility.

Could you please help me catch up to speed? What happened with that conversation and what's changed? Why the change of plan?

chyunsu3:
[Here](https://docs.google.com/document/d/1XKLq23chfb3Ad5MPETTahmKIB3Dv5ENuk88Wbj4N3Gk/edit?usp=sharing) is the detailed plan for integration testing. 

I will be on vacation between Dec 13 - Jan 10 and wanted to get a CI running before heading off. If I were to try setting up CI pipelines in CCCL and cuML, I won't be able to get it done in time.

trivialfis (approved):

ops/pipeline/query-latest-cccl.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+## Query latest version of CCCL using GitHub CLI
+## Note: RC version may be selected if available
+
+set -euo pipefail
+
+tmpfile="$(mktemp /tmp/abc-script.XXXXXX)"
```
trivialfis, line 7: Would like to work on simplifying the CI one day. The amount of bash scripts and their complexity is getting a bit too much now. Also, the failure rate of the CI is relatively high at the moment.

jakirkham:
Thanks everyone! 🙏

PR #11871, opened 2025-12-12
Fix building with RMM 26.02

hcho3:
* Include header `rmm/version_config.hpp` and `cuda/version` to get access to version macros.
* Use correct names for RMM version macros: `RMM_VERSION_MAJOR` and `RMM_VERSION_MINOR`.
* Accommodate MR API changes in https://github.com/NVIDIA/cccl/pull/5536 and https://github.com/NVIDIA/cccl/pull/5313

cc @bdice @jakirkham

### DIFF
--- a/src/common/device_vector.cuh
+++ b/src/common/device_vector.cuh
@@ -7,10 +7,11 @@
 #include <thrust/device_vector.h>            // for device_vector
 
 #if defined(XGBOOST_USE_RMM) && XGBOOST_USE_RMM == 1
+#include <cuda/version>                       // for CCCL_MAJOR_VERSION
 #include <cuda/memory_resource>               // for async_resource_ref
 #include <cuda/stream_ref>                    // for stream_ref
 
-#include <rmm/version_config.hpp>
+#include <rmm/version_config.hpp>             // for RMM_VERSION_MAJOR
 
 // TODO(hcho3): Remove this guard once we require Rapids 25.12+
 #if (RMM_VERSION_MAJOR == 25 && RMM_VERSION_MINOR == 12) || RMM_VERSION_MAJOR >= 26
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -336,6 +336,10 @@ if(PLUGIN_RMM)
   list(APPEND rmm_link_libs CUDA::cudart_static)
   set_target_properties(rmm::rmm PROPERTIES INTERFACE_LINK_LIBRARIES "${rmm_link_libs}")
 
+  target_compile_definitions(objxgboost PUBLIC
```
trivialfis: I think RMM has a `version_config.hpp` header file after installation, did they remove it?

src/common/device_vector.cuh
```diff
@@ -294,13 +303,21 @@ class ThrustAllocMrAdapter : public thrust::device_malloc_allocator<T> {
   pointer allocate(size_type n) {  // NOLINT(readability-identifier-naming)
     auto n_bytes = xgboost::common::SizeBytes<T>(n);
     auto s = cuda::stream_ref{::xgboost::curt::DefaultStream()};
+#if (CCCL_MAJOR_VERSION == 3 && CCCL_MINOR_VERSION >= 1) || CCCL_MAJOR_VERSION > 3
+    auto p = static_cast<T *>(mr_.allocate(s, n_bytes, std::alignment_of_v<T>));
+#else  // (CCCL_MAJOR_VERSION == 3 && CCCL_MINOR_VERSION >= 1) || CCCL_MAJOR_VERSION > 3
     auto p = static_cast<T *>(mr_.allocate_async(n_bytes, std::alignment_of_v<T>, s));
+#endif  // (CCCL_MAJOR_VERSION == 3 && CCCL_MINOR_VERSION >= 1) || CCCL_MAJOR_VERSION > 3
```
chyunsu3, line 310: The macro is getting pretty messy. We may want to consider writing an adapter class like `cccl_resource_ref` in https://github.com/rapidsai/rmm/pull/2017

@trivialfis What do you think?

trivialfis: It's probably fine now. They flipped to name to adapt to the asynchronous world, it's unlikely to happen again.

chyunsu3:
@trivialfis Can you approve the PR if it looks good?

trivialfis (approved):

src/common/device_vector.cuh
bdice: Yes. And RMM will be dropping all this complexity in 26.02 or 26.04 and will require CCCL >=3.2.0 at that time.

tests/cpp/helpers.cc
```diff
@@ -30,16 +30,18 @@
 #include <memory>
 #include <vector>
 
+#include "rmm/version_config.hpp"
```
bdice, line 33: Probably want `<>` here?
```suggestion
#include <rmm/version_config.hpp>
```

jakirkham: It looks like the same pattern repeats below

PR #11878, opened 2025-12-16
[wip][EM] Iterate through batches for custom objective.

trivialfis:
This PR provides an interface for the Python package to iterate through the labels when calculating the gradient. However, the data is still concatenated internally. In addition, we are still returning meta info through the host.

- Implement batched info iterator.
- Implement a push method for the gradient container.


trivialfis:
Closing for now.

The idea here was to iterate through the gradient for external memory training, since the size of the gradient matrix for vector leaf can be large. But the overhead and the complexity combined (but mostly the complexity) makes me have a second thought.

We can re-visit this if the memory usage of the gradient matrix becomes a bottleneck.

PR #11880, opened 2025-12-17
Fix predict() ignoring base_margin in xgb.DMatrix

### LINKED ISSUE #11872: Behaviour of `base_margin` during prediction
Hi,

I wanted to better undestand the behavior of using a `base_margin` during training and prediction of an `xgboost` model. I read the [very nice tutorial](https://xgboost.readthedocs.io/en/stable/tutorials/intercept.html#) and tried to extend it in the case where we have a simple train/test split. As you will see in the below code, I have two models that are exactly the same (one trained with `base_score = 0.5` and one with `base_margin = 0`) - which makes sense according to the tutorial - but upon prediction, using (or not using) the `base_margin` results in different predictions. Is that expected behavior? Maybe I have misundestood something?

``` r
library(xgboost)
library(testthat)

# Load data, split train/test
data(agaricus.train, package = "xgboost")
X = agaricus.train$data
y = agaricus.train$label

set.seed(42)
train_ids = sample(x = 1:nrow(X), size = 4000)
test_ids = setdiff(1:nrow(X), train_ids)

xgb_data_train = xgboost::xgb.DMatrix(data = X[train_ids, ], label = y[train_ids])
xgb_data_test  = xgboost::xgb.DMatrix(data = X[test_ids, ], label = y[test_ids])

# m1 model with no base_margin (default base_score = 0.5)
m1 = xgboost::xgb.train(
  data = xgb_data_train,
  params = list(base_score = 0.5, objective = "binary:logistic"),
  nrounds = 1)

# TRAIN ----
# does offset change trained model? => YES (seems also independent of nrounds)
# try base_margin = 1 => NOT the same as base_score = 0.5
xgboost::setinfo(xgb_data_train, "base_margin", rep(1, length(train_ids)))
#> [1] TRUE
m2 = xgboost::xgb.train(
  data = xgb_data_train,
  params = list(objective = "binary:logistic"),
  nrounds = 1)
testthat::expect_equal(xgb.dump(m1), xgb.dump(m2)) # fails
#> Error: Expected `xgb.dump(m1)` to equal `xgb.dump(m2)`.
#> Differences:
#> 8/16 mismatches
#> x[4]: "3:leaf=0.541463435"
#> y[4]: "3:leaf=0.360770881"
#> 
#> x[7]: "11:leaf=-0.41538465"
#> y[7]: "11:leaf=-0.712710977"
#> 
#> x[8]: "12:leaf=0.495652169"
#> y[8]: "12:leaf=0.323709249"
#> 
#> x[9]: "8:leaf=-0.594997048"
#> y[9]: "8:leaf=-1.10756671"
#> 
#> x[12]: "9:leaf=0.582978785"
#> y[12]: "9:leaf=0.395674318"

# try model m3 with base_margin = 0 => same as base_score = 0.5!!!
xgboost::setinfo(xgb_data_train, "base_margin", rep(0, length(train_ids)))
#> [1] TRUE

m3 = xgboost::xgb.train(
  data = xgb_data_train,
  # base_score doesn't matter
  params = list(base_score = 0.2, objective = "binary:logistic"),
  nrounds = 1)
testthat::expect_equal(xgb.dump(m1), xgb.dump(m3)) # models ARE THE SAME!

# PREDICT ----
# predict with no base_margin
p1 = predict(m1, newdata = xgb_data_test)
# predict with no base_margin, different model
p2 = predict(m2, newdata = xgb_data_test)
# predict with no base_margin, same model as m1 but trained with an offset (base_margin = 0)
p3 = predict(m3, newdata = xgb_data_test)
# ALL results different!!!
expect_false(all(p1 == p2))
expect_false(all(p1 == p3)) # WHY??? same models here!!! and same test data?

# add base_margin = 0 to predictions
xgboost::setinfo(xgb_data_test, "base_margin", rep(0, length(test_ids)))
#> [1] TRUE
# predict with base_margin
p11 = predict(m1, newdata = xgb_data_test)
# predict with base_margin, different model
p22 = predict(m2, newdata = xgb_data_test)
# predict with base_margin, same model as m1 but trained with an offset (base_margin = 0)
p33 = predict(m3, newdata = xgb_data_test)

expect_equal(p1, p11)
expect_equal(p2, p22)
expect_equal(p3, p33)
expect_false(all(p11 == p33)) # WHY???
```

<sup>Created on 2025-12-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Sanidhyavijay24:
Fixes #11872 

**The Issue ->**
Previously, `predict.xgb.Booster` prevented passing `base_margin  when the input was `xgb.DMatrix`. Additionally, if `base_margin` was NULL (default), the function failed to check internal `DMatrix` attributes, causing predictions to ignore the margin entirely.

**The Fix ->**

- Removed the `stop()` block restriction.
- Added logic to apply `base_margin` via `setinfo` if passed explicitly.
- Added auto-detection to retrieve `getinfo(newdata, "base_margin")` if the argument is NULL.

**Test ->**
Added a regression test verifying that changing the margin inside a `DMatrix` alters the prediction, and that explicit arguments work.

### DIFF
--- a/R-package/R/xgb.Booster.R
+++ b/R-package/R/xgb.Booster.R
@@ -364,7 +364,7 @@ predict.xgb.Booster <- function(object, newdata, missing = NA, outputmargin = FA
   is_dmatrix <- inherits(newdata, "xgb.DMatrix")
   if (is_dmatrix) {
     rnames <- NULL
-    
+
     # FIX: If user passed a margin argument, apply it to the DMatrix.
     if (!is.null(base_margin)) {
       setinfo(newdata, "base_margin", base_margin)


Sanidhyavijay24:
**Note on Commit History ->**
This branch includes commits from my previous PR because I branched off my local master. However, the Files Changed tab is clean and correctly shows only the relevant changes for this fix (`xgb.Booster.R` and `test_basic.R`).

I can rebase/clean up the history if preferred, but "Squash and Merge" should handle it automatically.
Sorry for the mess!.

R-package/R/xgb.Booster.R
```diff
@@ -362,14 +362,13 @@ predict.xgb.Booster <- function(object, newdata, missing = NA, outputmargin = FA
...
-      "'base_margin' is not supported when passing 'xgb.DMatrix' as input.",
-      " Should be passed as argument to 'xgb.DMatrix' constructor."
-    )
-  }
   if (is_dmatrix) {
     rnames <- NULL
+    
+    # FIX: If user passed a margin argument, apply it to the DMatrix.
+    if (!is.null(base_margin)) {
+      setinfo(newdata, "base_margin", base_margin)
```
david-cortes, line 370: This modifies the DMatrix in-place and propagates outside the function call. It's not supposed to work like that.

Sanidhyavijay24: Hmm.... **You are right** , modifying the user's `newdata` object in-place is an unsafe side effect since xgb.DMatrix uses reference semantics.

To support the `base_margin` argument safely (without modifying the original object) , I can only think of two options to support this argument safely:

**Option 1: Slice (Temporary Copy)**
We create a slice of the `DMatrix`. This creates a new handle, allowing us to set the margin on the temporary object without affecting the original: ->
```
if (!is.null(base_margin)) {
  # Create a slice (copy) to avoid side effects on the original object
  newdata <- slice(newdata, seq_len(nrow(newdata)))
  setinfo(newdata, "base_margin", base_margin)
}
```

**BUT** this will cause a slight performance overhead of creating a new handle.

**Option 2: Drop Argument Support**
We simply remove the stop() block but do not apply the argument automatically. Users must call setinfo themselves before prediction if they use DMatrix.

Which approach do you think will work??

david-cortes: I think this is a bug at the C++ level that would not be solveable from R alone, unless copying the data like in your option 1:
https://github.com/dmlc/xgboost/issues/11872#issuecomment-3666848941

I'd prefer to have a fix at the C++ level if possible (CC @trivialfis).

Sanidhyavijay24: Agreed, a C++ fix for cache invalidation is the correct root solution for the caching/correctness issue.

However, even with a C++ fix, we still face the R-level API design question you raised earlier -> Should `predict(dmat, base_margin=...)` modify the `dmat` object permanently?

Applying the argument directly causes a side effect on the user's object (mutating the `DMatrix`). **Option 1 (Slice)** handles this by keeping the margin ephemeral (only for that prediction call).

I am happy to pause this PR while @trivialfis investigates the C++ side, but I think the **"Slice"** approach will still be needed to support this argument safely without side effects in R.

david-cortes: > Should `predict(dmat, base_margin=...)` modify the dmat object permanently?

No, and shouldn't work at all. The base margin should either be passed to the DMatrix directly, or `predict` should be passed a non-DMatrix object. The C-level functions do not support a prediction from DMatrix plus base margin, and the R interface that works with DMatrices instead of R matrices/dfs is meant to be a low-level design that stays close to C and mimics the other language bindings, where that doesn't work either.

david-cortes: For reference: https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.Booster.predict

Sanidhyavijay24: Thanks for the reference seeing that the Python `Booster.predict` also doesn't support the argument makes the design parity clear.

Since we shouldn't change the code/API, I will drop the PR logic.

Though would you like me to add a Warning to the R documentation (`predict.xgb.Booster.Rd`) instead? I could explicitly note that modifying `base_margin` on a reused `xgb.DMatrix` might be ignored due to backend caching, and that creating a fresh object is the safe workaround.

Let me know if you want that doc update, or if I should just close this!

david-cortes: Ideally, the issue with incorrect usage of caches should get fixed at some point, so I don't think it's necessary to add it to the documentation. So yes, better to close it.

trivialfis:
Doing some tests here: https://github.com/dmlc/xgboost/pull/11885 Not sure how practical it is to reject all changes after DMatrix construction.

PR #11881, opened 2025-12-18
[mt] Refactor the Python boost function.

trivialfis:
- Forward the gradient calculation to the boost function.
- Cleanup circular module dependencies.
- Add the parameter `iteration` to the custom objective prototypes to align with the internal obj definition and to prepare for potential support of batched gradient.
- Avoid exposing the split gradient as parameters.

Ref:https://github.com/dmlc/xgboost/issues/9043


trivialfis:
cc @rongou

rongou (approved):

PR #11882, opened 2025-12-18
[EM] Drop support of GPU external memory page concatenation.

trivialfis:
This PR drops the support of the `extmem_single_page` parameter and prefer standard iterator-based external memory implementation.


trivialfis:
cc @rongou

PR #11883, opened 2025-12-18
[mt] Support feature selection.

trivialfis:
- Column sampling, including feature weights.
- Cleanup the `Init` method of the column sampler to avoid cross device copies.
- Small optimization for the scan kernel with a warp-based implementation.

Feature constraint is not working yet. I will take a deeper look after handling other simpler parameters.


trivialfis:
cc @rongou

rongou (approved):

PR #11885, opened 2025-12-18
[wip] Invalidate the DMatrix cache after modification.

trivialfis:
- [ ] high level test.
- [ ] Make sure the cache is not trashed.

### DIFF
--- a/include/xgboost/cache.h
+++ b/include/xgboost/cache.h
@@ -101,7 +101,7 @@ class DMatrixCache {
 
   void ClearExcess() {
     this->CheckConsistent();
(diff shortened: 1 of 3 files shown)


david-cortes:
How about deleting caches instead of erroring out on modifications?

trivialfis:
Yes, we can do that. But I would like to see the impact of making the DMatrix immutable first. I will update the PR to work with the cache.

trivialfis:
This is not going to be an easy change. Internally, the DMatrix doesn't have a concept of when the construction is "completed" and hence there's no easy way to check whether a modification has been made.

PR #11889, opened 2025-12-24
[mt] Assign a thread for each target when building the histogram. 

trivialfis:
~Stacked on https://github.com/dmlc/xgboost/pull/11883~

This PR optimizes the histogram kernel for vector leaf. It provides a launch bound for the multi-target histogram kernel, and uses one thread for each target gradient. In addition, it tunes the occupancy by reducing register pressure and provides different policies for shared memory usage.

### DIFF
--- a/src/tree/gpu_hist/histogram.cu
+++ b/src/tree/gpu_hist/histogram.cu
@@ -372,9 +372,6 @@ void DispatchCudaSm(std::int32_t device, Fn&& fn) {
  * now, we define a different version as a playground, without risking regressing the
  * established one.
  *
- * Alternative to the minBlocks for launch bounds, we might use the `__maxnreg__` to limit
- * registery.
- *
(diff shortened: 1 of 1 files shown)


src/tree/gpu_hist/histogram.cu
```diff
@@ -307,62 +302,109 @@ __global__ void __launch_bounds__(kBlockThreads)
...
+
 /**
  * @brief Kernel for multi-target histogram.
  *
  * Eventually, we should merge the kernel with the existing single-target version. For
  * now, we define a different version as a playground, without risking regressing the
  * established one.
  *
+ * Alternative to the minBlocks for launch bounds, we might use the `__maxnreg__` to limit
+ * registery.
```
trivialfis: Typo

trivialfis:
cc @rongou

PR #11894, opened 2025-12-28
[mt] Use target-major histogram layout.

trivialfis:
- Pre-calculate the int64 gradient and use a f-major layout (target-major). This sacrifices memory usage to get better performance for GPUs with low f64 capacity. We might want to make it optional in the future, or enable batched training for the gradient.
- Use a target-major layout for the histogram, similar to the CPU implementation.

The PR significantly improves the vector-leaf performance, especially for a large number of targets. We are getting close to a point where we can unify the histogram kernels by replacing the existing single-target version with the vector leaf version. I will do this once the overall vector leaf implementation is mature enough.

### DIFF
--- a/src/tree/gpu_hist/histogram.cu
+++ b/src/tree/gpu_hist/histogram.cu
@@ -408,8 +408,6 @@ __device__ void HistKernelOneNodeTarget(Accessor const& matrix, FeatureGroup con
   Idx const ridx_size = d_ridx_iter.size();
   auto const d_ridx = d_ridx_iter.data();
 
-  extern __align__(cuda::std::alignment_of_v<GradientPairInt64>) __shared__ char shmem[];
-
   if constexpr (Policy::kSharedMem) {
     dh::BlockFill(smem_hist, group.num_bins, GradientPairInt64{});
     __syncthreads();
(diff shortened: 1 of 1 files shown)


src/tree/gpu_hist/histogram.cu
```diff
@@ -363,92 +384,70 @@ void DispatchCudaSm(std::int32_t device, Fn&& fn) {
...
+}
+
+// Build the histogram for a single target in a single node.
 template <typename Policy, typename Accessor, typename RidxIterSpan>
-__global__ __launch_bounds__(HistBound::kBlockThreads, HistBound::kMinBlocks) void HistKernel(
-    Accessor const matrix, FeatureGroupsAccessor const feature_groups, RidxIterSpan* d_ridx_iters,
-    common::Span<std::uint32_t const> blk_ptr, common::Span<GradientPairInt64>* node_hists,
-    GradientPair const* d_gpair, bst_target_t n_targets, GradientQuantiser const* d_roundings) {
-  FeatureGroup group = feature_groups[blockIdx.y];
+__device__ void HistKernelOneNodeTarget(Accessor const matrix, FeatureGroup const& group,
```
trivialfis: Use a reference.

src/tree/gpu_hist/multi_evaluate_splits.cu
```diff
@@ -475,10 +487,15 @@ std::ostream &DebugPrintHistogram(std::ostream &os, common::Span<GradientPairInt
 
   std::vector<GradientPairInt64> h_node_hist(node_hist.size());
   dh::CopyDeviceSpanToVector(&h_node_hist, node_hist);
```
trivialfis: Remove debug code

trivialfis:
cc @rongou

PR #11895, opened 2025-12-29
Implement vector load for GPU compressed buffer.

trivialfis:
~Stacked on https://github.com/dmlc/xgboost/pull/11894~

Mainly to improve performance for the vector leaf.

- Extract the byteswap implementation, add nvcc and msvc support.
- Use vector load in the compressed iterator.

*todos*
- [x] Increase buffer padding.
- [x] Test double iterator


trivialfis:
We might have to update ctk for a newer version of CCCL and CUDA intrinsic.

trivialfis:
cc @rongou 

The windows test is currently failing due to old ctk version (12.5).

rongou (approved):

PR #11914, opened 2026-01-07
[mt] Basic support for regularization.

trivialfis:

- l1/l2
- min_split_loss
- max_delta_step
- Make split gradient optional even if the tree objective is used.
- Small cleanups.

The PR removes the debug flag requirement to ease testing. But the feature is still working in progress.


trivialfis:
cc @rongou

trivialfis:
https://github.com/dmlc/xgboost/blob/f5ff41e21087fb70184da7f91969ce851302fa8f/src/tree/param.h#L240

Both conditions are evaluated to true, then the `&&` returns false.

PR #11918, opened 2026-01-11
Small cleanups for updater tests.

trivialfis:


trivialfis:
cc @hcho3 could you please help take a look into the JVM packages docker failure when you are back?

PR #11924, opened 2026-01-14
[ci] Test using sccache.

trivialfis:
without cache:
- https://github.com/dmlc/xgboost/actions/runs/20999210760/job/60364305284?pr=11924
- https://github.com/dmlc/xgboost/actions/runs/20999210760/job/60364305328?pr=11924

with cache:
- https://github.com/dmlc/xgboost/actions/runs/20999584533/job/60365602486?pr=11924
- https://github.com/dmlc/xgboost/actions/runs/20999584533/job/60365602484?pr=11924

### DIFF
--- a/.github/workflows/misc.yml
+++ b/.github/workflows/misc.yml
@@ -21,15 +21,20 @@ jobs:
       run:
         shell: bash -l {0}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6.0.1
         with:
           submodules: 'true'
       - uses: dmlc/xgboost-devops/actions/miniforge-setup@main
         with:
           environment-name: cpp_test
           environment-file: ops/conda_env/cpp_test.yml
+      - uses: trivialfis/sccache-gha/actions/sccache@main
+        with:
+          cache-key-prefix: ${{ github.job }}
       - name: Build and test XGBoost
         run: bash ops/pipeline/build-test-cpu-nonomp.sh
+      - name: Show sccache stats
+        run: sccache --show-stats
 
   c-api-demo:
     name: Test installing XGBoost lib + building the C API demo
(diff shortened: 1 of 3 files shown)


trivialfis:
cc @hcho3 

I created the sscache composite [action](https://github.com/trivialfis/sccache-gha) to enable caching compilation result. But most of the XGBoost CI jobs explicitly invoke docker to run scripts, which makes it difficult to use a custom action in tests. Is it possible to let GA handle the use of images like https://github.com/open-mpi/pyhwloc/blob/c37741a9dd471e2b45ec9d4d12502d668f4c3fc1/.github/workflows/main.yml#L44 ?

chyunsu3:
@trivialfis

> Is it possible to let GA handle the use of images

We will need to figure out how to handle authentication to pull docker images from AWS ECR.

Also, we will need to revamp the [CI doc](https://xgboost.readthedocs.io/en/stable/contrib/ci.html#examples-useful-tasks-for-local-development), to make it clear 1) how to identify the right container; and 2) how to invoke the container locally. The doc should look more like https://docs.rapids.ai/resources/reproducing-ci/#reproducing-ci-jobs. We should still keep the `docker_run.py` script for local development purposes, but the CI pipelines will now be using `container:` directive instead of `docker_run.py`.

trivialfis:
> We will need to figure out how to handle authentication to pull docker images from AWS ECR.

The ghcr [seems](https://docs.github.com/en/billing/concepts/product-billing/github-packages#how-use-of-github-packages-is-measured) to be free for public packages, can we use that? Assuming we don't actually need private registry.

trivialfis:
We should be able to get macos build working as well, if the composite action is considered the right way to approach this.

chyunsu3:
GHCR is outside AWS, so using GHCR from EC2 workers will incur extra traffic charges. (ECR currently costs zero traffic charges.)
As an alternative, we can continue to use ECR and use `aws-actions/amazon-ecr-login` action to handle authentication. https://github.com/aws-actions/amazon-ecr-login

trivialfis:
I would like to merge my repo into the xgboost devops once we can get a CUDA build to run. For now, I will keep it as-is for easier testing.

PR #11926, opened 2026-01-14
Add Booster.compute_leaf_similarity() method

### LINKED ISSUE #11919: Find similar observations using leaf node matching
I'm proposing a feature to find which training samples are "similar" to a prediction sample, based on whether they end up in the same leaf nodes across trees.

The idea is simple: if two observations consistently land in the same leaves across many trees, the model "sees" them as similar.

I already built a function to do this for my use case. I was doing a prediction to assess sales potential for some geographies and the model was predicting way too high for a zip code. I couldn't figure out why from the features or feature importances alone.

So I wrote a wrapper that checks: for this zip code, which training observations land in the same leaf nodes most often? Turns out a tiny zip code on the other side of the world matched 83% of the time. Looking at the features, they were strangely similar. It was an outlier and later I excluded it form the training.

I would never have found this by looking at Euclidean distance or raw feature values. The model knew these were similar. I just needed a way to ask it.

### This is how it works

1. Use `pred_leaf=True` to get leaf indices for each tree
2. For each tree, check if query and reference land in same leaf (boolean)
3. Average across trees → similarity score between 0 and 1 (I also weight trees by the variance of their leaf predictions, meaning that trees where all leaves predict ~same value aren't very discriminative.)

### What the API could look like

```python
# High-level
similar_idx, scores = model.find_similar(
    query=X_query,
    reference=X_train,
    k=5,
)

# Or lower-level on booster
query_leaves = booster.predict(DMatrix(X_query), pred_leaf=True)
ref_leaves = booster.predict(DMatrix(X_train), pred_leaf=True)
similarity = booster.compute_leaf_similarity(query_leaves, ref_leaves)
```

## Why this is useful

- **Debugging predictions**: "why is this prediction so high?" → find similar training samples and inspect them
- **Finding bad training data**: outliers in training can affect predictions in unexpected places
- **Explaining to stakeholders**: "this prediction is similar to these 5 historical cases" is easier to trust than a black box number

I haven't used it but I found out Random Forests have something similar to this, proximity matrices. Would be nice to have this in XGBoost.

## Questions

- Is this something that fits in XGBoost's scope, or better as a separate utility?
- Any concerns about scaling to large datasets? (the leaf prediction is fast, similarity is just broadcasting)
- Happy to put together a PR if there's interest

mfdel:
## Description

This PR adds a new method `Booster.compute_leaf_similarity()` to compute similarity between observations based on leaf node co-occurrence across trees, similar to Random Forest proximity matrices.

Closes #11919

## API

```python
similarity = booster.compute_leaf_similarity(data, reference, weight_type="gain")
```

**Parameters:**
- `data` (DMatrix): Query dataset (m samples)
- `reference` (DMatrix): Reference dataset (n samples)  
- `weight_type` (str): `"gain"` (default) or `"cover"`

**Returns:** `ndarray` of shape (m, n) with values in [0, 1]

## Formula

$$S_w(a, b) = \frac{\sum_{t=1}^{T} w_t \cdot \mathbf{1}[\phi_t(a) = \phi_t(b)]}{\sum_{t=1}^{T} w_t}$$

Where:
- $\phi_t(x)$ = leaf index for sample $x$ in tree $t$
- $w_t$ = weight of tree $t$

## Weight Types

Following the suggestion in #11919 to reuse feature importance definitions:

- **`"gain"`** (default): Sum of loss reduction across all splits in the tree. Trees that contribute more to model improvement are weighted higher.
- **`"cover"`**: Sum of hessian values across all splits. For regression (hessian=1), this equals sample count. For classification (hessian=p(1-p)), this emphasizes trees that process more uncertain samples.

## Implementation Notes

- Pure Python using `pred_leaf=True` and `trees_to_dataframe()`
- Row-by-row computation for memory efficiency with large datasets
- Properties: self-similarity = 1.0, symmetric when data==reference, range [0,1]

## Tests

Added `tests/python/test_leaf_similarity.py` with tests for:
- Shape and range validation
- Self-similarity (diagonal = 1.0)
- Both weight types work correctly
- Default is `gain`
- Invalid weight_type raises `ValueError`

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -2981,6 +2981,70 @@ class Booster:
             "Data type:" + str(type(data)) + " not supported by inplace prediction."
         )
 
+    def compute_leaf_similarity(
+        self,
+        data: DMatrix,
+        reference: DMatrix,
+        weight_type: str = "gain",
+    ) -> np.ndarray:
+        """Compute similarity between observations based on leaf node co-occurrence.
+
+        Two samples are similar if they land in the same leaf nodes across trees.
+        This is similar to Random Forest proximity matrices.
+
+        Parameters
+        ----------
+        data :
+            Query dataset (m samples).
+        reference :
+            Reference dataset (n samples).
+        weight_type :
+            How to weight trees: "gain" (by loss improvement) or "cover"
+            (by hessian sum, approximately sample count for regression).
+
+        Returns
+        -------
+        similarity : ndarray of shape (m, n)
+            Similarity scores in [0, 1].
+        """
+        if weight_type not in ("gain", "cover"):
+            raise ValueError(
+                f"weight_type must be 'gain' or 'cover', got '{weight_type}'"
+            )
+
+        query_leaves = self.predict(data, pred_leaf=True)
+        ref_leaves = self.predict(reference, pred_leaf=True)
+
+        if query_leaves.ndim == 1:
+            query_leaves = query_leaves.reshape(-1, 1)
+        if ref_leaves.ndim == 1:
+            ref_leaves = ref_leaves.reshape(-1, 1)
+
+        n_trees = query_leaves.shape[1]
+
+        trees_df = self.trees_to_dataframe()
+        split_nodes = trees_df[trees_df["Feature"] != "Leaf"]
+        col = "Gain" if weight_type == "gain" else "Cover"
+        tree_weights = split_nodes.groupby("Tree")[col].sum()
+
+        weights = np.zeros(n_trees, dtype=np.float32)
+        for tree_id, w in tree_weights.items():
+            if tree_id < n_trees:
+                weights[int(tree_id)] = w
+
+        if weights.sum() == 0:
+            weights = np.ones(n_trees, dtype=np.float32)
+
+        total_weight = weights.sum()
+        m, n = len(query_leaves), len(ref_leaves)
+
+        similarity = np.zeros((m, n), dtype=np.float32)
+        for i in range(m):
+            matches_i = query_leaves[i] == ref_leaves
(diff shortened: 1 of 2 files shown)


trivialfis:
Thank you for opening the PR, I will see if I can run some experiments with it and avoid the use of the tree to dataframe (that method itself is quite difficult to extend).

trivialfis:
note to myself:
**todos**:
- [ ] Test multi-class/target.
- [ ] Gradient boosting random forest.
- [ ] Non-DMatrix inputs.
- [ ] `strict_shape`.

trivialfis:
Just an update: I'm experimenting with the similarity matrix. I need to learn more about how well XGBoost is measuring the samples (not very well at the moment)

Figure_2-xgb-boost

trivialfis:
@mfdel Could you please grant me permission to push to this branch for some code refactoring?

image

ZhuYizhou2333:
I am willing to further expand and optimize this method. The issues I observed during testing are that multi-target paths are not currently supported and that a dependency on pandas is introduced. I hope to design a more suitable method to calculate tree weights to achieve better compatibility and performance. There is only one question to ask: is it acceptable to add new C API interfaces to output the weights? I believe adding minimal C API interfaces is necessary because there is currently a lack of an interface to output the "weight" information of different trees.

I think after accessing the new C API interface, the current method should be compatible with the vast majority of configurations, only `gblinear` is not supported because it uses a non-tree model.

ZhuYizhou2333:
https://github.com/dmlc/xgboost/pull/12157 This is the new PR I submitted, which includes a minimal C API and a faster similarity calculation implementation. It does not use the tree to dataframe. I also conducted performance tests: compared with random forest, and evaluated the performance of various weighting methods.

PR #11927, opened 2026-01-15
[ci] Test registry login.

trivialfis:
- Use github action container tag for building the CPU tests.
- Add sccache to the CPU build.
- Use a GA job to login.

without cache:
- https://github.com/dmlc/xgboost/actions/runs/21019521430/job/60431504428

with cache:
- https://github.com/dmlc/xgboost/actions/runs/21019783113/job/60432272700?pr=11927

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,6 +25,7 @@ jobs:
         id: login-ecr
         uses: aws-actions/amazon-ecr-login@v2.0.1
         with:
+          mask-password: 'false'
           registries: '492475357299'
     outputs:
       docker_registry: ${{ steps.login-ecr.outputs.registry }}


chyunsu3:
Let's update the CI container to include `curl`.

trivialfis:
let me do that after basic things work (still trying to put something together).

@hcho3 Would be great if you can help supervise in case if I'm doing something damaging (like unsafe use of credential or launching too many workers just for login, or using the wrong runner, etc).

trivialfis:
cc @hcho3 I moved the CPU gtest builds to using the container tag and added sccache support for them. I haven't been able to go through the CUDA build yet, which is a lot more complicated. It would be great if I can get an initial review first (and some help with the CUDA build/packaging).

hcho3 (approved):
Using `container:` directive from the GitHub Actions looks nice, because it lets us keep only `build-cpu-impl.sh` and remove `build-cpu.sh`. Let's merge this now and create another PR to change the CUDA build.

trivialfis:
Let me do one more test and see if removing the `mask=false` works with AWS worker. ;-)

trivialfis:
> Let me do one more test and see if removing the mask=false works with AWS worker. ;-)

nope.

PR #11930, opened 2026-01-15
Fix build with latest rmm.

trivialfis:
Also need https://github.com/NVIDIA/cccl/issues/7246 to have a clean build. I prefer using the MR as a member variable instead of asking RMM to lookup a map for every alloc/dealloc. Will followup on this one.

- [x] restore the workflow.


chyunsu3:
We need to fix the logic for detecting the latest RMM. Right now, it uses `VERSION` file of `rapidsai/rmm` repo, but it is better to query the latest git tag instead.

chyunsu3:
@trivialfis Is this ready for review?

trivialfis:
Not yet.

trivialfis:
@hcho3 The PR is ready to go.

PR #11931, opened 2026-01-15
Update dmlc core, disable CUDA async pool on Windows.

trivialfis:


trivialfis:
Disabled the Windows CUDA async pool, couldn't workaround the compiler issue.

src/common/device_vector.cuh
trivialfis:
```suggestion
  void SetAsync(bool use_async_pool) {
#if !defined(xgboost_IS_WIN)
    this->use_async_pool_ = use_async_pool;
#endif
  }
```

PR #11932, opened 2026-01-16
[CI] Test using sccache on Windows.

trivialfis:
- Use ninja instead of msbuild on Windows.
- Use sccache.
- Use a composite [action](https://github.com/trivialfis/sccache-gha/blob/main/actions/msvc-dev-env/action.yml) to setup the msvc developer environment.
- Remove custom cccl now that we have a newer CTK.

todos
- [ ] GPU build is failing without explicit error. Need to get some help from experts. Will defer it.
- [x] cache doesn't work with the current AMI due to missing gzip.

with cache:
- https://github.com/dmlc/xgboost/actions/runs/21064213791/job/60577578500?pr=11932

without cache:
- https://github.com/dmlc/xgboost/actions/runs/21060488513/job/60565431055

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -73,8 +73,7 @@ jobs:
           cache-key-prefix: ${{ github.job }}-${{ matrix.build_suite }}
       - name: Build and test
         run: bash ops/pipeline/build-cpu.sh ${{ matrix.build_suite }}
-      - name: Show sccache stats
-        run: sccache --show-stats
+      - run: sccache --show-stats
 
   build-cuda:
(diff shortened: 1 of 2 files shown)


trivialfis:
@hcho3 It would be great if you could help take a look into the composite [action](https://github.com/trivialfis/sccache-gha/blob/main/actions/msvc-dev-env/action.yml) implementation when you are available. Just double checking it's not doing something dangerous.

chyunsu3:
@trivialfis Seems fine.

trivialfis:
> Seems fine.

Excellent! Please help review the PR when you are available.

PR #11935, opened 2026-01-17
[R] Fix more memory safety issues

david-cortes:
Fixes another issue found by rchk.


trivialfis:
Thank you for the quick fix. Let me try to run rchk locally next week 

https://github.com/kalibera/rchk

PR #11936, opened 2026-01-17
[mt] Add missing CPU tests.

trivialfis:


trivialfis:
Merged tests into https://github.com/dmlc/xgboost/pull/11965

PR #11937, opened 2026-01-18
[ci] Use GHA container for CUDA Linux builds.

trivialfis:
- Split up auditwheel due to the need for a different container.
- Merge cuda13 workflow into main to avoid login.
- Install dependencies for auditwheel.

I will try to refactor and pipelines after this PR and see if I can unify some of the pipelines.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,7 +68,7 @@ jobs:
           echo "UBSAN_OPTIONS=print_stacktrace=1:log_path=ubsan_error.log" >> $GITHUB_ENV
           # Work around https://github.com/google/sanitizers/issues/1614
           sysctl vm.mmap_rnd_bits=28
-      - uses: trivialfis/sccache-gha/actions/sccache@main
+      - uses: dmlc/xgboost-devops/actions/sccache@main
         with:
           cache-key-prefix: ${{ github.job }}-${{ matrix.build_suite }}
       - name: Build and test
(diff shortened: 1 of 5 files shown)


chyunsu3:
Enabling "magic cache" will speed up `actions/cache` by redirecting the cache to S3. See https://runs-on.com/caching/magic-cache/ for more details.

trivialfis:
Reran the job to check the cache statistic: https://github.com/dmlc/xgboost/actions/runs/21197376345/job/60982825580?pr=11937

Looks good.

trivialfis:
> Enabling "magic cache" will speed up actions/cache by redirecting the cache to S3

We might need to configure the cache to limit the amount of cached files, though. I rely on the eviction policy to enable sccache. By default, the cache is FIFO. The sccache action uses the git hash as the key (which is similar to a random hash) and relies on the fact that when multiple keys match, it picks the newest one. Without a proper eviction policy, we will have a very large cache due to multiple unique hashes from all the commits.

chyunsu3:
> We might need to configure the cache to limit the amount of cached files, though.

According to the guide, a lifecycle policy of 10 days applies, so files older than 10 days will be automatically deleted.

trivialfis:
@hcho3 Would you like me to migrate this PR to https://github.com/dmlc/xgboost-devops/pull/71 ?

chyunsu3:
Yes

trivialfis:
Got it, I will wait for that PR then.

trivialfis:
I will see if I can optimize these tests after this PR:
```
2026-01-21T08:27:45.2881030Z 270.46s call     python-gpu/test_gpu_updaters.py::TestGPUUpdaters::test_external_memory
2026-01-21T08:27:45.2881981Z 243.55s call     python-gpu/test_gpu_prediction.py::TestGPUPredict::test_shap
2026-01-21T08:27:45.2883088Z 226.12s call     python-gpu/test_gpu_prediction.py::TestGPUPredict::test_shap_interactions
2026-01-21T08:27:45.2884059Z 170.16s call     python-gpu/test_gpu_ordinal.py::test_training_continuation
```

Issue #11884, opened 2025-12-18
xgboost4j-spark-gpu fails to run on multi-GPU server with GPUs in exclusive process mode and spark-rapids plugin


eordentlich:
Exceptions result when running as above for both driver and executor processes.    Stack traces for both includes
`
/workspace/src/common/cuda_rt_utils.cc: 60: cudaErrorDevicesUnavailable: CUDA-capable device(s) is/are busy or unavailable
`
indicating XGBoost code is attempting to create CUDA contexts even though the corresponding Spark driver and executor processes should either not be using GPUs (Spark driver) or have been assigned other device ids by Spark.

On the driver side, it seems one source (might be more than one with the first one triggering the error) might be coming from RabitTracker initialization which uses the common [InitNewThread function](https://github.com/dmlc/xgboost/blob/cd7f9974e996a5b30e9cd3530510a5ae291c6bbf/src/global_config.cc#L17) whose logic seems to setCurrentDevice on the default device.   The source on the executor side is not clear at this point, but it is XGBoost code.

When switching GPUs to default process mode, I believe this behavior (depending on root cause above for executors, i.e. is it coming from main executor thread or new thread) can lead to race conditions in current stream/thread device settings with spark-rapids, resulting in spurious cuda kernels launched by executors on non-assigned device 0, in turn causing errors like: 
```
Caused by: ai.rapids.cudf.CudfException: after determining tmp storage requirements for exclusive_scan: cudaErrorInvalidDevice: invalid device ordinal
```
first noted in https://github.com/NVIDIA/spark-rapids-examples/issues/565

Specifically, executor processes launching kernels on assigned device and device 0 can be observed in nsys traces.

trivialfis:
Thank you for sharing the issue. Do you intend to use XGBoost with GPUs in your current setting? Or do you prefer XGBoost to not touch any CUDA API and exclusively use CPU?

eordentlich:
Intention is to use GPUs on executors, but not on the driver.   Actually was working on root causing https://github.com/NVIDIA/spark-rapids-examples/issues/565 .  After observing unexpected device usage by executor processes attempted to get more insight using exclusive process mode and encountered the noted behavior, which was unexpected as executor processes should not be changing devices when spark-rapids is enabled.    Seems the spark driver should also not be creating spurious CUDA contexts.

trivialfis:
[...]

PR #11939, opened 2026-01-19
[jvm-packages] Set GPU device id explicitly at very beginning for training

wbo4958:
Ref to: https://github.com/dmlc/xgboost/issues/11884

This PR adds JNI functions to manage CUDA device selection for XGBoost Spark, and and allows Spark executors to explicitly set and acquire GPU devices before XGBoost training, relying on Spark's resource scheduler to prevent GPU conflicts between executors.

### DIFF
--- a/jvm-packages/xgboost4j/src/native/xgboost4j-gpu.cu
+++ b/jvm-packages/xgboost4j/src/native/xgboost4j-gpu.cu
@@ -1,5 +1,5 @@
 /**
- * Copyright 2021-2026, XGBoost Contributors
+ * Copyright 2021-2025, XGBoost Contributors
  */
 #include <jni.h>
 #include <xgboost/c_api.h>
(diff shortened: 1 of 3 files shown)


wbo4958:
Hi @trivialfis, Please help review it.

jvm-packages/xgboost4j/src/native/xgboost4j-gpu.cu
```diff
@@ -509,3 +509,35 @@ int QdmFromCallback(JNIEnv *jenv, jobject jdata_iter, jlongArray jref, char cons
...
+}  // anonymous namespace
+
+extern "C" {
+
+/*
+ * Class:     ml_dmlc_xgboost4j_java_XGBoostJNI
+ * Method:    CudaSetDevice
+ * Signature: (I)I
+ */
+JNIEXPORT jint JNICALL Java_ml_dmlc_xgboost4j_java_XGBoostJNI_CudaSetDevice(JNIEnv *, jclass,
```
trivialfis: Please consolidate this function into `xgboost4j.cpp`, then use the `SetDevice` function from `cuda_rt_utils.h` file, which can handle CUDA build and CPU-only build

wbo4958: Done.

PR #11940, opened 2026-01-19
[mt] Support loss guide.

trivialfis:

To support loss guide, we have to keep a buffer for the entire tree's weight and gradient sum. We might do some optimizations in the future, right now this is focusing on feature coverage.


trivialfis:
cc @rongou

PR #11950, opened 2026-01-22
[mt] Feature importance variants.

trivialfis:
- Gain and coverage.

At the moment, I store the sum hessian for all targets. We might need to store a vector instead. I will leave it as future work if it comes up.

todos:
- [x] Test mixed tree types.


trivialfis:
We can have vector leaf for https://github.com/dmlc/xgboost/pull/11926 as well

trivialfis:
cc @rongou

PR #11954, opened 2026-01-26
Move header files from `include` to `src`.

trivialfis:
We have a semi-working C++ "interface" that's not really an interface. Some files are impossible to use without hacking XGBoost's internals due to forward declarations.

This PR hides the following files to reduce confusion and to reduce the need for broken forward declarations:

- learner.h
- gbm.h
- cache.h
- predictor.h

Aside from automated header sorting, I did not make many changes. This is to ensure Git recognizes this PR as a simple rename.


trivialfis:
cc @RAMitchell in case this conflicts with the new design of SHAP.

chyunsu3:
Have you considered putting internal headers to `include/detail` subdirectory, like RAPIDS does?

trivialfis:
Yes. We will have to add a set of sub directories into `details` like `details/tree/hist`. Do you see any benefits of doing so?

chyunsu3:
Using `src` for internal headers will require lots of preceding `../../..` in header includes, whereas using `include/detail` will not. Granted, we'd need to add subdirectories to `include/detail`, increasing the complexity somewhat.

Or we should probably consider adding `src` to include path, so that we don't need the `../../..` ?

Issue #11944, opened 2026-01-21
train on CUDA and load on CPU but got an EXC_BAD_ACCESS crash


warscape:
trained my XGBoost model on CUDA, saved in json format and then tried to load model on macOS, got an EXC_BAD_ACCESS crash.
The python versions on both platforms are 3.10.x, and XGBoost versions are both 3.0.3.

Full crash log attached.

[Python-2026-01-21-163640.log](https://github.com/user-attachments/files/24761527/Python-2026-01-21-163640.log)

warscape:
Tried to modify model params and save another copy in deprecated format, it can be successfully loaded on macOS but will fail at the predict call.

`new_params = {'device': 'cpu', 'tree_method': 'hist', 'predictor': 'cpu_predictor'}
        model = xgb.XGBClassifier(params =new_params)
        model.load_model(_modelPath)
        model.set_params(**new_params)
        model.save_model(_cpuModelPath)`

Here are crash logs:

[Python-2026-01-21-110924.log](https://github.com/user-attachments/files/24761614/Python-2026-01-21-110924.log)
[Python-2026-01-21-095943.log](https://github.com/user-attachments/files/24761615/Python-2026-01-21-095943.log)

trivialfis:
I couldn't reproduce the issue by loading the GPU-trained model with a CPU-only XGBoost build. By default, the output from `save_model` doesn't contain any device-related information (you can try to search "cuda" or "gpu" in the JSON file), I suspect the issue is caused by something else.

> 'predictor': 'cpu_predictor'

XGBoost 3.x doesn't have a predictor parameter.

@hcho3  could you please take a look with a Mac when you are available?

warscape:
@trivialfis thanks for the quick response!
The predictor parameter has no effect, - I got a warning in console log saying this is an unknown parameter and is ignored.

The xgboost I used on macOS is just xgboost, not xgboost-cpu, and it works well when doing training with CPU.
Do you think I should try xgboost-cpu?

trivialfis:
For macos, xgboost is the same as xgboost-cpu, since there's no CUDA support there.

I'm just trying to reproduce the issue (but haven't been able to).

warscape:
@trivialfis @hcho3 any updates on this issue? 🙏

trivialfis:
I will try set up a temporary CI job and see if we can reproduce there.

trivialfis:
[...]

PR #11955, opened 2026-01-27
[ci] Run cross platform tests

trivialfis:
related: https://github.com/dmlc/xgboost/issues/11944

failure logs:
https://github.com/dmlc/xgboost/actions/runs/21430698435/job/61712473307

### DIFF
--- a/tests/cross-platform/test_cross_platform_model.py
+++ b/tests/cross-platform/test_cross_platform_model.py
@@ -61,6 +61,7 @@ def test_inference(model_path: str) -> None:
 
 
 def main() -> int:
+    """Entry for both training and inference."""
     parser = argparse.ArgumentParser(description="Cross-platform XGBoost model test.")
     parser.add_argument("--train", action="store_true", help="Train model using GPU")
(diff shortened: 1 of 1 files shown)


trivialfis:
I downloaded the model and tested it on a local Linux build and a local Windows build. Both work as expected. Not sure what I'm doing wrong with the macos CI test job here

chyunsu3:
It turns out that the OpenMP configuration logic is buggy. The `find_openmp_macos` function in dmlc-core shadows the `find_openmp_macos` function in XGBoost. As a result, `libxgboost.dylib` is not properly patched to use the correct `libomp.dylib`. We will need dmlc/dmlc-core#719 to fix it.

trivialfis:
!!
That's so very weird. Thank you for looking into this! I wouldn't thought of it. The omp seems tricky, based on the https://github.com/dmlc/xgboost/issues/11944#issuecomment-3812861070 I will see if I can get access to a mac

trivialfis:
@hcho3 Please help take a look when you are available. I will follow up with the macos issues and check whether everything is in good shape (reasonably)

PR #11957, opened 2026-01-27
Fix flaky dask test.

trivialfis:
The `test_n_workers` got flaky with the latest Dask. The PR uses `scatter` to control data placement instead of `submit` and `persist`.

For reviewers:

XGBoost uses the union of all workers obtained from the input data to train a model. This test creates a list of dask arrays residing on different workers, and checks whether XGBoost can obtain the union set. The test was flaky because the `submit` did not put the data in the expected worker from time to time.


trivialfis:
Thank you for the review. I will merge this to unblock the CI.

PR #11958, opened 2026-01-27
[CI] Move remaining style checks to pre-commit

RAMitchell:
I have removed the clang-format file as its not used anywhere in CI - I think this is problematic because clang-format might conflict with the cpplint check in CI. If its useful to keep this, then I can add it to pre-commit.


chyunsu3:
Can we keep clang-format and remove cpplint? I prefer clang-format's output.
@trivialfis What do you think?

trivialfis:
Please keep the clang format. clangd picks up the file. https://clangd.llvm.org/features#formatting

RAMitchell:
I've kept clang-format and added it to pre-commit.

hcho3 (approved):

PR #11959, opened 2026-01-27
Avoid memory usage estimation when the async pool is enabled.

trivialfis:
- Similar treatment with RMM.
- Drop support for old thrust versions.


trivialfis:
cc @rongou please help take a look when you are available.

PR #11961, opened 2026-01-27
make temporary buffers an instance variable

### LINKED ISSUE #11960: [python] TransformedDf temporary buffers leak memory
The current implementation of TransformedDf initializes temporary_buffers as a class level variable with empty list as the default, which causes unbounded memory growth when running many transformations in a single python runtime. I have a project where xgboost inference is run in an http server, and discovered this via a memray profile of our production container.

~I initially was going to submit a patch that just made the buffer an instance variable, but after digging a bit further it seems like the buffer is never actually used (I could be missing something though!).~ I was missing something

So, I'm going to submit 2 separate patch sets -- one where I just change this to an instance variable, the other where I remove it entirely. Let me know which approach you prefer.

Edit: Reading the code in a bit more depth I think the original purpose was to just keep the buf from being gc'd until after consumption, so I'll actually omit the 2nd patchset

jochs:
Fixes #11960 

Maintain memory reference to buffers for the lifespan of the TransformedDf instance instead of accidentally storing them in the classvar -- with this change the buffers should be GC'd when the TransformedDf instance is GC'd instead of sticking around for the remainder of the python runtime lifecycle.

### DIFF
--- a/src/common/cuda_context.cuh
+++ b/src/common/cuda_context.cuh
@@ -1,5 +1,5 @@
 /**
- * Copyright 2022-2025, XGBoost Contributors
+ * Copyright 2022-2026, XGBoost Contributors
  */
 #ifndef XGBOOST_COMMON_CUDA_CONTEXT_CUH_
 #define XGBOOST_COMMON_CUDA_CONTEXT_CUH_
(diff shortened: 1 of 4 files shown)


jochs:
@trivialfis thanks for the help! I think tests.test_distributed.test_with_dask.test_with_dask.TestWithDask.test_empty_quantile_dmatrix needs to be updated to expect an error in this case, same as the gpu equivalent. do you want me to handle that? I want to avoid stepping on your toes while you're helping me out 😅

trivialfis:
Thank you for the fix! I'm more than happy to help.

I will handle the empty DMatrix test: https://github.com/dmlc/xgboost/pull/11963

trivialfis (approved):

PR #11962, opened 2026-01-27
[doc] Update for the CUDA async pool and the R CRAN package.

trivialfis:


trivialfis:
@hcho3 would you like to take a look into this when you are available?

hcho3 (approved):
CUDA async pool looks very nice

PR #11963, opened 2026-01-27
[ci] Fix dask empty worker test.

trivialfis:
The re-coder cannot handle an empty worker. Use one-hot encoded data for the test.

Failure: https://github.com/dmlc/xgboost/actions/runs/21411829626/job/61651364490?pr=11959


trivialfis:
cc @rongou

PR #11964, opened 2026-01-28
Refactor shap tests

RAMitchell:
Separate shap tests into its own file. Create a number of new test cases. 

I will probably eventually create an interpretability folder as we separate these from the predictor internally.


trivialfis:
@RAMitchell Please help check the failure at the master branch: https://github.com/dmlc/xgboost/actions/runs/21471842826/job/61846060499

Issue #3734, opened 2018-09-27
 Different results on groupby data


nirupamkar:
I am using XGB for a classification problem. If I groupby the features and use the count as weight, ideally there should not be any change in the model as loss function should remain same. However I am seeing a change. Please let me know if I am missing something.

np.random.seed(42)
x=pd.DataFrame({"y":[int(np.random.random()>.8) for i in range(100)],
               "x1":[np.round(np.random.random(),1) for i in range(100)],"x2":[np.round(np.random.random(),1) for i in range(100)]})
x["weight"]=1
import xgboost  as xgb

valid_features=["x1","x2"]

xgb_train_data = xgb.DMatrix(x[valid_features], label=x.y.values, 
                                  weight=x.weight.values
                                  )
parameters = {
"boosting_type": 'gbdt', 
"objective" : 'binary:logistic',
    "subsample":.7,
    "random_state":42,
    "num_leaves":64,
    "eval_metric":["logloss"]
}
print "running "
evallist=[(lgb_train_data, 'eval')]
model2 = xgb.train(parameters,
xgb_train_data,10,evallist)

log_loss(x.y,model2.predict(xgb_train_data),sample_weight=x.weight.values)
0.3447492831200361

x1=x.groupby(["x1","x2","y"]).agg({"weight":"sum"}).reset_index()
xgb_train_data = xgb.DMatrix(x1[valid_features], label=x1.y.values, 
                                  weight=x1.weight.values,
                                  )
parameters = {
"boosting_type": 'gbdt', 
"objective" : 'binary:logistic',
    "subsample":.7,
    "seed":42,
    "depth":64,
    "eval_metric":["logloss"]
}
print "running "
evallist=[(lgb_train_data, 'eval')]
model2 = xgb.train(parameters,
xgb_train_data,10,evallist)

log_loss(x1.y,model2.predict(xgb_train_data),sample_weight=x1.weight.values)
0.3524573728069663

1 more interesting observation. If I change subsample=1 the results are matching. So does it indicate the sampling might not be weighted?

hcho3:
Yes, currently sampling due to `subsample` currently does not take account of instance weights.

khotilov:
That reminds me about a git stash that I have made somewhere a while ago when I really needed the subsampling to be weighted... I'll dig it up. It might be more involved now though since we have many updaters.

@nirupamkar even with proper subsampling I would not expect to get the same results (should be rather similar though) since the same values might end up in and out of a subsample.

[...]

PR #11965, opened 2026-01-28
[mt] Initial subsample implementation.

trivialfis:
This PR ports the single-target sampling to the vector leaf. I will follow up with a complete implementation of MVS.

- Add CPU implementation for gradient-based sampling.
- Define the sampler in `sampler.cu/cc` files as it's not exclusively about MVS.
- Support vector leaf. The sampling should be consistent between gradient types. But the scale can differ. As a result, the value gradient might be biased. I don't have a good solution at the moment and assume the dimension reduction algorithm can do a decent job at maintaining bias and variance.

related:
- https://github.com/dmlc/xgboost/issues/3734

### DIFF
--- a/src/tree/hist/sampler.cc
+++ b/src/tree/hist/sampler.cc
@@ -3,12 +3,12 @@
  */
 #include "sampler.h"  // for kDefaultMvsLambda
 
-#include <cmath>      // for sqrt
-#include <cstddef>    // for size_t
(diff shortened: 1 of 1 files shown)


trivialfis:
cc @rongou This is built upon your work:-) I haven't added estimation for the lambda yet. Already too much change for this PR.

Issue #11499, opened 2025-06-04
XGBoost 3.0+ fails with a degenerate data set with identical labels


hcho3:
Starting version 3.0, XGBoost computes a smart default for `base_score`. For example, for the `binary:logistic` objective, XGBoost assigns the mean label to `base_score`.

This logic works most of the time, but fails if the training data contains all identical labels.

Reproducer:
```python
import numpy as np
import xgboost as xgb

dtrain = xgb.DMatrix(np.asarray([[1.0], [1.0]]), label=[1.0, 1.0])
dtest = xgb.DMatrix(np.asarray([[1.0]]), label=[1.0])

bst = xgb.train(
    {"objective": "binary:logistic"},
    dtrain,
    1,
    evals=[(dtest, "validation")],
)
```

Error:
```
Traceback (most recent call last):
  File "/home/phcho/test.py", line 7, in <module>
    bst = xgb.train(
          ^^^^^^^^^^
  File "/home/phcho/miniforge3/lib/python3.12/site-packages/xgboost/core.py", line 729, in inner_f
    return func(**kwargs)
           ^^^^^^^^^^^^^^
  File "/home/phcho/miniforge3/lib/python3.12/site-packages/xgboost/training.py", line 183, in train
    bst.update(dtrain, iteration=i, fobj=obj)
  File "/home/phcho/miniforge3/lib/python3.12/site-packages/xgboost/core.py", line 2246, in update
    _check_call(
  File "/home/phcho/miniforge3/lib/python3.12/site-packages/xgboost/core.py", line 310, in _check_call
    raise XGBoostError(py_str(_LIB.XGBGetLastError()))
xgboost.core.XGBoostError: [10:03:12] /workspace/src/objective/./regression_loss.h:69: Check failed: base_score > 0.0f && base_score < 1.0f: base_score must be in (0,1) for logistic loss, got: 1
Stack trace:
  [bt] (0) /home/phcho/miniforge3/lib/python3.12/site-packages/xgboost/lib/libxgboost.so(+0x2a6ecc) [0x7f8df94a6ecc]
  [bt] (1) /home/phcho/miniforge3/lib/python3.12/site-packages/xgboost/lib/libxgboost.so(+0xeda139) [0x7f8dfa0da139]
  [bt] (2) /home/phcho/miniforge3/lib/python3.12/site-packages/xgboost/lib/libxgboost.so(+0x682723) [0x7f8df9882723]
[...]

PR #11973, opened 2026-01-29
Fix logistic with constant labels.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/11499


trivialfis:
cc @rongou 

I kept the pre-commit failure for now in case the reformatting makes the PR unreadable.

PR #11974, opened 2026-01-29
Enable threaded data initialization for CPU.

trivialfis:
Extracted from https://github.com/dmlc/xgboost/pull/11390 , modified to use the context instead. In addition, a new `ParallelForBlock` is created.

### DIFF
--- a/src/common/threading_utils.h
+++ b/src/common/threading_utils.h
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019-2025, XGBoost Contributors
+ * Copyright 2019-2026, XGBoost Contributors
  */
 #ifndef XGBOOST_COMMON_THREADING_UTILS_H_
 #define XGBOOST_COMMON_THREADING_UTILS_H_
@@ -103,9 +103,7 @@ class BlockedSpace2d {
   }
 
   // Amount of blocks(tasks) in a space
-  [[nodiscard]] std::size_t Size() const {
-    return ranges_.size();
-  }
+  [[nodiscard]] std::size_t Size() const { return ranges_.size(); }
 
   // get index of the first dimension of i-th block(task)
   [[nodiscard]] std::size_t GetFirstDimension(std::size_t i) const {
@@ -135,7 +133,6 @@ class BlockedSpace2d {
   std::vector<Range1d> ranges_;
   std::vector<std::size_t> first_dimension_;
 };
-
 
 // Wrapper to implement nested parallelism with simple omp parallel for
 template <typename Func>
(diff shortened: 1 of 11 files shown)


trivialfis:
@razdoburdin Could you please help review the PR when you are available? I have listed you as the author. Please let me know if you prefer otherwise.

razdoburdin:
> @razdoburdin Could you please help review the PR when you are available? I have listed you as the author. Please let me know if you prefer otherwise.

yes, I will take a look

src/data/gradient_index.cc
```diff
@@ -83,32 +83,34 @@ GHistIndexMatrix::GHistIndexMatrix(Context const *, MetaInfo const &, EllpackPag
...
-                                   common::HistogramCuts cuts, bst_bin_t max_bins_per_feat,
-                                   bool is_dense, double sparse_thresh, std::int32_t n_threads)
+GHistIndexMatrix::GHistIndexMatrix(Context const *ctx, SparsePage const &batch,
+                                   common::Span<FeatureType const> ft, common::HistogramCuts cuts,
+                                   bst_bin_t max_bins_per_feat, bool is_dense, double sparse_thresh)
     : cut{std::move(cuts)},
       max_numeric_bins_per_feat{max_bins_per_feat},
       base_rowid{batch.base_rowid},
       isDense_{is_dense} {
-  CHECK_GE(n_threads, 1);
```
razdoburdin, line 103: Is the `n_threads >= 1` checked elseware ?

src/common/threading_utils.h
```diff
@@ -273,6 +273,24 @@ void ParallelFor1d(Index size, std::int32_t n_threads, Func&& fn) {
...
 }
 
+/** @brief Use n_threads as the number of blocks. */
+template <typename Index, typename Func>
+void ParallelForBlock(Index size, std::int32_t n_threads, Func&& fn) {
+  static_assert(std::is_void_v<std::invoke_result_t<Func, common::Range1d>>);
+  std::size_t blk_size = size / n_threads + (size % n_threads > 0);
+  ParallelFor(n_threads, n_threads, [&](auto tid) {
+    auto blk_beg = tid * blk_size;
+    if (blk_beg >= size) {
```
razdoburdin: This check looks redundant. If (size == 0), blk_size == blk_beg == end == 0, so the (end == blk_beg) check shall work in this case.

razdoburdin (commented):
looks good for me

src/data/gradient_index.cc
trivialfis: As long as the number of threads comes from the `Context` class, it should go through this check: https://github.com/dmlc/xgboost/blob/f5f1e659e1eb39103bde8c37b50c81b7dcf809ce/src/common/threading_utils.cc#L120

src/common/threading_utils.h
trivialfis: It's not for `size==0`. Say `size==1`, but `n_threads==8`, then the `block_size=1`, the second block (thread) would have `block_begin = 1 * 1 = 1`, end would be `end = 2 * 1 = 2`, and `block_begin != end`.

razdoburdin: Ah, I see. 
May be just replace `if (end == blk_beg)` by `if (end <= blk_beg)`?

trivialfis: Thank you for the suggestion, done.

trivialfis:
Haven't done this very often. I made sure the only commit here has @razdoburdin as the author, I hope squashing and merging the commit with github will retain the information.

trivialfis:
No.. the author is switched.

PR #11978, opened 2026-01-30
[CI] Break out python lint script

RAMitchell:
Current pre-commit setup calls the lint_python.py script, which touches all files. Pre-commit is intended to run fast and only touch changed files.

@trivialfis @hcho3 in case we have not fully captured the exceptions in the old file, if you notice pre-commit touch files it shouldn't please just update their exceptions.


RAMitchell:
@trivialfis ruff is good we can try it

PR #11979, opened 2026-01-30
[ci] Use sccache for Windows CPU build.

trivialfis:
We can't cache CUDA build yet. But at least we can try to cache the CPU TUs.

Without cache:
https://github.com/dmlc/xgboost/actions/runs/21515165126/job/61991604592?pr=11979
with cache:
https://github.com/dmlc/xgboost/actions/runs/21515165126/job/62000272026

Only a small difference. The majority of time is spent on CUDA build


trivialfis:
Still failing.

PR #11980, opened 2026-01-30
Update pre-commit clang format to latest.

trivialfis:
https://github.com/pre-commit/mirrors-clang-format/tags

- Update clang format
- Disable formatting code comments

The old clang format doesn't recognize `// clang-format off` comment. I would like to be able to keep some blocks from being formatted.

The formatting for code comments is quite off. Comments are usually intentionally formatted in a way that's easy to read, like this [table](https://github.com/dmlc/xgboost/blob/8ca2f55b31945ff073c47a0268fd49696cf54978/src/data/data.cc#L214). It's better to not let clang-format change it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -232,21 +232,18 @@ if(USE_CUDA)
   endif()
 
   enable_language(CUDA)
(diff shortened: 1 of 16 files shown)


trivialfis:
cc @RAMitchell @hcho3

Issue #11500, opened 2025-06-05
Importing xgboost before torch + openmp causes seg fault


ikrommyd:
I'm on apple sillicon in a fresh environment that I just created.
```
uv venv --python 3.12
source .venv/bin/activate
uv pip install xgboost torch
```
Running the following seg faults:
```py
import xgboost
import torch

M = torch.tensor([[2.0, 0.5],
                  [0.5, 1.0]],
                 dtype=torch.double, device="cpu")

L = torch.linalg.cholesky(M)

print("Cholesky succeeded:", L)
```
Switching the import order (torch before xgboost) fixes the problem. Also doing `export OMP_NUM_THREADS=1` fixes the problem.

From my quick debugging session I think it's that xgboost loads `libomp` from homebrew first while then torch loads it's own version. I don't know if torch or xgboost should do something differently here, I'm just opening the issue to raise awareness.

ikrommyd:
Naturally, if I install the packages via pip in a fresh conda environment and also do `mamba install llvm-openmp` the problem is fixed cause both packages grab the same library

ikrommyd:
```
-> otool -L .venv/lib/python3.12/site-packages/xgboost/lib/libxgboost.dylib | grep libomp
        @rpath/libomp.dylib (compatibility version 5.0.0, current version 5.0.0)

-> otool -L .venv/lib/python3.12/site-packages/torch/lib/libomp.dylib | grep libomp

.venv/lib/python3.12/site-packages/torch/lib/libomp.dylib:
        /opt/homebrew/opt/libomp/lib/libomp.dylib (compatibility version 5.0.0, current version 5.0.0)
```
and indeed when loading
```
-> DYLD_PRINT_LIBRARIES=1 python3 - <<EOF 2>&1 | grep libomp.dylib
import xgboost
import torch
EOF


dyld[58427]: <AE900456-D54F-3657-A54C-21B02B390E20> /opt/homebrew/Cellar/libomp/20.1.5/lib/libomp.dylib
dyld[58427]: <2448B10B-E7F4-3091-9302-765D47BFB0B8> /Users/iason/.venv/lib/python3.12/site-packages/torch/lib/libomp.dylib
```
while with the opposite import order
```
-> DYLD_PRINT_LIBRARIES=1 python3 - <<EOF 2>&1 | grep libomp.dylib
import torch
import xgboost
EOF

[...]

Issue #10946, opened 2024-10-31
libomp: xgboost incorrect LC_RPATH referenced in libxgboost.dylib


Ubiquitous:
Upon installing xgboost (2.1.2) with pip on macOS (15.0.1) I noticed that there was an error related to not finding `libomp` once installed. In the installed `libxgboost.dylib` the rcpath references an absolute 'homebrew' directory (I don't use homebrew).

% otool -l /opt/local/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/xgboost/lib/libxgboost.dylib
```
Load command 15
          cmd LC_RPATH
      cmdsize 48
         path /opt/homebrew/opt/libomp/lib (offset 12)
```
It should likely be changed to use: `@rpath/libomp` unless you expect anyone who doesn't use homebrew to fix the hardcoded path anytime they install xgboost with pip.

trivialfis:
@Ubiquitous Thank you for opening an issue. Would you like to help fix it? We are indeed having a lot of headaches with macos distribution around openmp, your expertise is appreciated.

hcho3:
We merged #10440 to patch `libxgboost.dylib` so that it depends on `@rpath/libomp.dylib`, by running
```
otool install_name_tool -change /opt/homebrew/opt/libomp/lib/libomp.dylib "@rpath/libomp.dylib" libxgboost.dylib
```

It looks like we need to run additional command(s) to patch the rcpath as well.

hcho3:
@Ubiquitous Can you try running the following command and see if that fixes the issue?
```
install_name_tool -add_rpath "@rpath/libomp" /opt/local/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/xgboost/lib/libxgboost.dylib
```

hcho3:
I also want to hear from @jameslamb about how LightGBM solved this issue.

jameslamb:
> *It should likely be changed to use: @rpath/libomp unless you expect anyone who doesn't use homebrew to fix the hardcoded path anytime they install xgboost with pip.*

This doesn't sound quite right to me. Having an RPATH pointing to Homebrew's preferred location is just a hint to the loader. If you don't have OpenMP installed via Homebrew, then there shouldn't be anything found at that path, and the loader should precede to search standard system paths like `/usr/lib`.

Before we go any further... @Ubiquitous can you please share precisely what problem you encountered? How are you getting OpenMP and where is it installed?

[...]

Issue #8946, opened 2023-03-19
OpenMP detection is still broken on MacOS


barracuda156:
Despite an earlier PR claiming to fix it: https://github.com/dmlc/xgboost/pull/8684
It is still broken, at least with GCC:
```
checking whether OpenMP will work in a package... no
*****************************************************************************************
         OpenMP is unavailable on this Mac OSX system. Training speed may be suboptimal.
         To use all CPU cores for training jobs, you should install OpenMP by running\n
             brew install libomp
```
Moreover, it gives a wrong recommendation (`libomp` is Clang-specific).

P. S. IMO it is a bad wording to use “should“ for an optional package manager. One does not need any – `libomp`, while not needed at all with GCC, can be installed directly. Also, there is Macports which has it, not just Homebrew.

barracuda156:
Example of configure that actually works: https://github.com/2005m/kit/issues/36#issuecomment-1450355975
(Prefix gotta be adjusted for the package manager being used.)

jameslamb:
Thanks for tagging me and for the additional details. Are you interested inusing your knowledge of this area to improve the prpject by opening a pull request? If not @trivialfis I'd be happy to look into this in a few days.

barracuda156:
> Thanks for tagging me and for the additional details. Are you interested inusing your knowledge of this area to improve the prpject by opening a pull request? If not @trivialfis I'd be happy to look into this in a few days.

@jameslamb Thank you for responding! I am interested to have this fixed so that we can have `xgboost` for `R` in Macports. Allow me a day, perhaps, I need to make sure build is fixed (which includes 32-bit) and tests pass, and then figure out how to amend OpenMP checks so that they work for a) GCC and Clang and b) Homebrew and Macports cases.

trivialfis:
Thank you for raising an issue! As suggested by @jameslamb , would be great if you can submit a fix, your expertise in this area is really appreciated.

hcho3:
@barracuda156 Thanks for raising the issue. As you noticed, currently XGBoost has only been tested with the Apple Clang + Homebrew combination. It will be great if you can make it work with Macports,

barracuda156:
UPD. Tests mostly pass, there is one failure on `ppc32`, which is a known problem, though it is not immediately obvious how to fix it in this case (`malloc` errors usually happen with executables; those are caused by duplicate `libstdc++` in older macOS with modern GCC installed):

```
[...]

PR #11983, opened 2026-02-03
[wip] Add optional support to use static omp on macos

trivialfis:
- There's `/opt/homebrew/opt/libomp/lib/libomp.a`. The PR adds an option to use it for macos.
- Skip the omp patching if the compiler is not clang.

### DIFF
--- a/cmake/FindOpenMPMacOS.cmake
+++ b/cmake/FindOpenMPMacOS.cmake
@@ -36,7 +36,7 @@ macro(find_openmp_macos)
       set(_omp_static_lib "/opt/local/lib/libomp/libomp.a")
       set(_omp_include_dir "/opt/local/include/libomp")
     else()
-      message(FATAL_ERROR "USE_OPENMP_STATIC=ON but libomp.a not found. "
+      message(FATAL_ERROR "BUILD_WITH_SHARED_OPENMP=OFF (static OpenMP) but libomp.a not found. "
         "Install via: brew install libomp")
     endif()
 


trivialfis:
cc @hcho3 @jameslamb 

I have tested the resulting wheels on an M4 instance on AWS. You can find the wheels at the bottom of the "Summary" page: https://github.com/dmlc/xgboost/actions/runs/21625123435. There is some logic I don't quite understand, including the use of `install_name_tool`. They don't seem to affect the end result when the static lib is used. It would be great if you could review and advise on the next steps.

I'm hoping to resolve the following issues for the Python PyPI package. R with autotools is beyond my scope at the moment. The Conda package should continue to work out of the box.
- https://github.com/dmlc/xgboost/issues/8946
- https://github.com/dmlc/xgboost/issues/10946
- https://github.com/dmlc/xgboost/issues/11500

In addition to the standard runs, I have also tested #11500 without `libomp` from Homebrew at test time.

The paths for macports are copied from https://github.com/dmlc/xgboost/issues/8946#issuecomment-1480939034 I haven't tested it yet.

trivialfis:
cc @barracuda156

barracuda156:
Logic of the option seems backward. Why not make an option for static one (since default is dynamic)?

trivialfis:
Mostly to align the naming with existing flags (`BUILD_WITH_SHARED_NCCL`).

barracuda156:
> Mostly to align the naming with existing flags

I see. Do I get it right that _by default_ there should be no change on macOS? Static build seems to a) assume clang (gcc does not use `libomp` at all, but I am not sure it can use its `libgomp` statically) and b) always pick Homebrew lib as long as it exists, which will create a mess if installation is done with MacPorts while Homebrew is co-installed (it is fragile, I know, but we cannot be sure it is not the case on user-end), but for a non-default config perhaps no big deal.

trivialfis:
Not exactly. This is an opt-in choice. It's disabled by default and XGBoost will continue to work (or not work) as before unless you enable the option. You can enable it if it suits your environment. I don't think there's a solution that can work on all settings automatically.

barracuda156:
> I don't think there's a solution that can work on all settings automatically.

I don’t think that making automatic choices which use package manager-provided dependencies is the right thing in principle (at least outside of CI), regardless of a given package manager being more prevalent at a given moment of time, but at least it makes sense to use Clang-specific stuff conditional on Clang being used. We do not want to try linking to `libomp` with GCC.

trivialfis:
> We do not want to try linking to libomp with GCC.

That makes sense. Thank you for the suggestion.

trivialfis:
Hmm. cross platform tests are failing 
https://github.com/dmlc/xgboost/actions/runs/21666731281/job/62464019930?pr=11983
```
2026-02-04T10:12:29.2921940Z   [117/117] Linking CXX shared library lib/libxgboost.dylib
2026-02-04T10:12:29.3026770Z   ld: warning: object file (/opt/homebrew/opt/libomp/lib/libomp.a[2](kmp_alloc.cpp.o)) was built for newer 'macOS' version (14.0) than being linked (12.0)
```

https://github.com/dmlc/xgboost/actions/runs/21666731281/job/62464020125?pr=11983
```
2026-02-04T10:02:40.6638260Z   ld: warning: object file (/usr/local/Cellar/libomp/21.1.8/lib/libomp.a[2](kmp_alloc.cpp.o)) was built for newer 'macOS' version (14.0) than being linked (10.15)
```

trivialfis:
I couldn't reproduce the error locally

trivialfis:
I come to the conclusion that there's no robust way to handle OpenMP on macOS.

jameslamb:
> I come to the conclusion that there's no robust way to handle OpenMP on macOS.

so did `{data.table}` 😭 

> _After several years of many reports of installation problems on Mac, it's time to gingerly point out that there have been no similar problems on Windows or Linux._

https://github.com/Rdatatable/data.table/blob/1bd88cbcc806bdaa2aa276e12c78fb4ba13cb645/po/R-es.po#L3660-L3667

Honestly, recently I've been thinking of trying to just remove OpenMP completely in LightGBM and replace it with `std::thread`. I am so tired of dealing with this dependency and stuff like https://github.com/microsoft/LightGBM/issues/7141 and https://github.com/microsoft/LightGBM/issues/6595#issuecomment-2273654934

But I have no idea what the performance impact of that would be. For me at least, it'd be a large project 😭

trivialfis:
> trying to just remove OpenMP completely in LightGBM and replace it with std::thread. 

Yeah, that's my thinking as well. I'm willing to take some performance hit on macos to fix this issue once and for all.

PR #11984, opened 2026-02-04
Replace isort/black with ruff

RAMitchell:


.pre-commit-config.yaml
```diff
@@ -12,6 +12,18 @@ repos:
       - id: check-executables-have-shebangs
       - id: check-shebang-scripts-are-executable
       - id: mixed-line-ending
+  - repo: https://github.com/astral-sh/ruff-pre-commit
```
trivialfis, line 15: Does it know where to look for configuration? `pyproject.toml`

RAMitchell: I believe it does find it automatically.

trivialfis (approved):

PR #11985, opened 2026-02-04
Move internal shap algorithms into separate namespace.

RAMitchell:


tests/cpp/predictor/test_shap.cc
```diff
@@ -42,6 +49,49 @@ Args BaseParams(Context const* ctx, std::string objective, std::string max_depth
               {"colsample_bytree", "1"},
               {"device", ctx->IsSycl() ? "cpu" : ctx->DeviceName()}};
 }
+
+gbm::GBTreeModel LoadGBTreeModel(Learner* learner, Context const* ctx,
```
RAMitchell: It is quite annoying to get a GBTreeModel out of a booster.

src/interpretability/shap.cc
```diff
@@ -0,0 +1,417 @@
...
+            p_contribs[ncolumns - 1] += base_score(gid);
+          }
+        }
+        feats.Drop();
+      });
+    }
+  }
+}
+
+void ShapInteractionValuesCPU(Context const *ctx, DMatrix *p_fmat,
```
trivialfis: I have been moving the naming convention to `namespace cpu_impl { void XAlgo(); }` from `void XAlgoCPU()`.

src/predictor/interpretability/shap.cc
```diff
@@ -0,0 +1,417 @@
...
+        feats.Fill(view[i]);
+        for (bst_target_t gid = 0; gid < n_groups; ++gid) {
+          float *p_contribs = &contribs[(row_idx * n_groups + gid) * ncolumns];
+          for (bst_tree_t j = 0; j < tree_end; ++j) {
+            if (h_tree_groups[j] != gid) {
+              continue;
+            }
+            std::fill(this_tree_contribs.begin(), this_tree_contribs.end(), 0);
+            auto const sc_tree = model.trees[j]->HostScView();
+            CalculateContributions(sc_tree, feats, &mean_values[j], this_tree_contribs.data(),
```
trivialfis: I know this is WIP, but let's keep the objective of avoiding code duplication in mind. I tried to find ways to unify the code between CSR DMatrix and the gradient index, but it got a bit messy. But I think unifying the code is worth some extra layers of complexity.

tests/cpp/predictor/test_shap.cc
trivialfis: Yeah, I would like to split up the concept of "model" and everything else like "optimizers/tree builders". But it might be too much for XGBoost at its current state.

src/predictor/interpretability/shap.cc
RAMitchell: Agreed. I did notice that and had planned to do something about it.

RAMitchell:
I am having two major difficulties with this PR. The categorical recoding logic is complicated and I don't want to carry this through to the interpretability module. The other is to simply get the tree ensemble out of learner which seems somehow impossible in the current setup.

trivialfis:
I can look into simplifying the categorical features after holiday.

src/predictor/interpretability/shap.cc
RAMitchell: I put the data accessors for predictor in a header so I could use them in shap.

src/predictor/interpretability/shap.cu
```diff
@@ -0,0 +1,446 @@
...
+      cuda::proclaim_return_type<bool>([=] __device__(PathInfo const& e) { return e.IsLeaf(); }));
+
+  info.resize(end - info.begin());
+  using LenT = decltype(std::declval<PathInfo>().length);
+  auto length_iterator = dh::MakeTransformIterator<LenT>(
+      info.begin(), cuda::proclaim_return_type<LenT>(
+                        [=] __device__(PathInfo const& info) { return info.length; }));
+  dh::caching_device_vector<size_t> path_segments(info.size() + 1);
+  thrust::exclusive_scan(ctx->CUDACtx()->CTP(), length_iterator, length_iterator + info.size() + 1,
+                         path_segments.begin());
```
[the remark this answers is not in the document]
RAMitchell: Copilot found a legitimate bug here.

src/predictor/interpretability/shap.cu
```diff
@@ -0,0 +1,446 @@
...
+    p_dst->resize(src.size());
+    auto d_dst = dh::ToSpan(*p_dst);
+    dh::safe_cuda(cudaMemcpyAsync(d_dst.data(), src.data(), d_dst.size_bytes(), cudaMemcpyDefault,
+                                  ctx->CUDACtx()->Stream()));
+  }
+};
+
+using DeviceModel = GBTreeModelView<dh::DeviceUVector, TreeViewVar, CopyViews>;
+std::mutex s_model_mu;
+
```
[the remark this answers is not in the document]
RAMitchell: Dealing with this in #12022

trivialfis:
I haven't looked into the details yet. But it would be great if we could establish some conventions here:
- Do we need the `gpu_` prefix for CUDA files? My preference is that the `.cu(h)` file extension is sufficient and doesn't need additional annotations.
- Since you are trying to separate the SHAP module, should it be outside of the predictor directory?

RAMitchell:
> Since you are trying to separate the SHAP module, should it be outside of the predictor directory?

That was my first attempt but it depends on all of the data accessors in prediction, in particular recoding.

src/predictor/interpretability/shap.cc
```diff
@@ -0,0 +1,372 @@
...
+  std::vector<std::vector<bst_float>> contribs_tloc(n_threads, std::vector<bst_float>(ncolumns));
+
+  auto device = ctx->Device().IsSycl() ? DeviceOrd::CPU() : ctx->Device();
+  auto base_margin = info.base_margin_.View(device);
+  auto ft = p_fmat->Info().feature_types.ConstHostVector();
+
+  auto process_batches = [&](auto acc) {
+    using AccT = std::decay_t<decltype(acc)>;
+    if (p_fmat->PageExists<SparsePage>()) {
+      for (auto const &page : p_fmat->GetBatches<SparsePage>()) {
```
trivialfis: Is it possible to unify the sparse page and gindex code path?

trivialfis:
> The categorical recoding logic is complicated and I don't want to carry this through to the interpretability module.

Would you like to share a bit more on the complication from a caller's perspective for the SHAP module?

> That was my first attempt but it depends on all of the data accessors in prediction, in particular recoding.

I think you have extracted them out?

trivialfis:
Please let me know if you would like me to give the refactoring a try.

src/predictor/interpretability/shap.cc
RAMitchell: Let me see if I can improve this.

RAMitchell:
I originally wanted to have a separate interpretability folder, but these new functions remain coupled to prediction. The shap values all need to add up to the margin prediction. You could put the data loading accessors elsewhere (data/utils) but I think they end up only used by shap and prediction. Let me know if you have better ideas.

PR #11988, opened 2026-02-04
Add support for expectile regression.

matthieubulte:
Hello! This PR adds expectile regression support. Expectile is a regression method based on an asymmetric squared error, [1], which gives a smooth alternative to quantiles regression. These are implemented in catboost :) 

I'm not extremely familiar with the xgboost codebase, so I did my best to get something working and running from python, and got some help from codex for extra bindings in other languages as well as just finding what else had to be done.

What's in there:
- Add `reg:expectileerror` objective and `expectile` metric with `expectile_alpha` (consistent naming across C++/R/Python).
- Initialize expectile base score using (weighted) mean. One can compute the global expectile instead, but this requires solving an iterative equation, and I wasn't sure if that's something that one wants in every context, so I thought it would be better to skip it.
- Update docs and R tests for multi‑alpha output naming and printed metadata.
- Added unit tests.

The current state is a mix of "it works on my machine" and "unit tests pass", but I'm sure I'm missing some things here and would love to get more guidance!

[1] https://en.wikipedia.org/wiki/Expectile which is not very nice to read, I find the formulation here easier to read https://epub.ub.uni-muenchen.de/31542/1/1471082x14561155.pdf

### DIFF
--- a/tests/python/test_eval_metrics.py
+++ b/tests/python/test_eval_metrics.py
@@ -1,5 +1,7 @@
 import numpy as np
 import pytest
+import xgboost as xgb
+from xgboost import testing as tm
 from xgboost.testing.metrics import (
     check_expectile_error,
     check_precision_score,
@@ -10,9 +12,6 @@ from xgboost.testing.metrics import (
     run_roc_auc_binary,
     run_roc_auc_multi,
 )
-
-import xgboost as xgb
-from xgboost import testing as tm
 
 rng = np.random.RandomState(1337)
 


src/objective/regression_obj.cu
```diff
@@ -394,6 +396,126 @@ XGBOOST_REGISTER_OBJECTIVE(PseudoHuberRegression, "reg:pseudohubererror")
...
+
+  void InitEstimation(MetaInfo const& info, linalg::Vector<float>* base_score) const override {
+    CHECK(!alpha_.Empty());
+    auto n_targets = this->Targets(info);
+    base_score->SetDevice(ctx_->Device());
+    base_score->Reshape(n_targets);
+
+    linalg::Vector<float> label_mean;
+    if (info.weights_.Empty()) {
+      common::SampleMean(ctx_, info.IsColumnSplit(), info.labels, &label_mean);
```
trivialfis, line 474: Is the mean value the optimal estimator for the intercept? In quantile regression, it's sample quantiles, and the median is the optimal estimator when the quantile = 0.5. If there's no close solution, the `FitIntercept` should handle this with one-step newton.

RAMitchell: Mean value is optimal for alpha=0.5. I think there is no formula for the others? I believe it makes sense to intialise all to 0.5 and do one iteration. @matthieubulte would you be willing to do an experiment to see if this intialisation helps?

src/objective/regression_obj.cu
```diff
@@ -394,6 +396,126 @@ XGBOOST_REGISTER_OBJECTIVE(PseudoHuberRegression, "reg:pseudohubererror")
...
+
+    linalg::ElementWiseKernel(ctx_, gpair,
+                              [=] XGBOOST_DEVICE(std::size_t i, std::size_t j) mutable {
+                                auto pred = predt(i, j);
+                                auto label = labels(i, 0);
+                                auto expectile = alpha[j];
+                                auto diff = pred - label;
+                                auto weight_scale = diff >= 0.0f ? (1.0f - expectile) : expectile;
+                                auto sample_weight = weight[i];
+                                auto grad = weight_scale * diff * sample_weight;
```
RAMitchell, line 460: Looks correct.

src/objective/regression_obj.cu
matthieubulte: @trivialfis @RAMitchell indeed, the mean value is optimal for `alpha=0.5`. I was actually wondering if it would be desirable to short-cut to the standard MSE in that case.

I like the idea of doing mean initialization + one step of Newton as you both suggest and happy to run an experiment.

matthieubulte:
@RAMitchell would the convergence test be a unit test, or which form would you want?

https://www.tse-fr.eu/sites/default/files/TSE/documents/doc/wp/2023/wp_tse_1458.pdf claims that the expectile function is monotonic
This paper also looks at some computational considerations for the iterative solving, I'll try to use this for the initialization.

RAMitchell:
A unit test would be good, can be in any language. Just do something like 10 examples, max depth 1, 10 boosting iterations with learning rate high (0.5-1.0) and see if it converges to your pre-computed expectiles for those examples.

matthieubulte:
Thanks a lot for the feedback!

I ended up basing the tests on a `Uniform[0,1]` distribution because there the `alpha-`expectile is equal to
```
y_alpha = sqrt(alpha) / (sqrt(alpha) + sqrt(1 - alpha))
```
See https://www.tse-fr.eu/sites/default/files/TSE/documents/doc/wp/2023/wp_tse_1458.pdf Table C.5 and some wrangling.

I added the requested unit test verifying that we do converge in a very simple setting.

As for the initialization, I ran the expectile training with mean and 1-step Newton over the mean, both computed in python and fed via the `base_score`.

Here's the code, not added to the repo:
```python
  import numpy as np
  import xgboost as xgb

  rng = np.random.default_rng(42)
  n = 200_000
  y = rng.random(n)
  X = np.zeros((n, 1))
  dtrain = xgb.DMatrix(X, label=y)

  def uniform_expectile(alpha: float) -> float:
      s = np.sqrt(alpha)
      t = np.sqrt(1.0 - alpha)
      return s / (s + t)

  def run(alpha: float, num_boost_round: int = 200):
      mean_base = float(y.mean())
      diff0 = mean_base - y
      w = np.where(diff0 >= 0.0, 1.0 - alpha, alpha)
      newton_base = float((w * y).sum() / w.sum())

      params = {
          "objective": "reg:expectileerror",
          "expectile_alpha": alpha,
          "tree_method": "hist",
          "max_depth": 1,
          "eta": 0.2,
          "subsample": 1.0,
          "colsample_bytree": 1.0,
          "lambda": 0.0,
          "alpha": 0.0,
          "min_child_weight": 0.0,
          "eval_metric": "expectile",
          "verbosity": 0,
      }

      def train(base_score: float):
          p = dict(params)
          p["base_score"] = float(base_score)
          evals_result = {}
          booster = xgb.train(
              p,
              dtrain,
              num_boost_round=num_boost_round,
              evals=[(dtrain, "train")],
              evals_result=evals_result,
              verbose_eval=False,
          )
          return booster, evals_result["train"]["expectile"]

      booster_mean, curve_mean = train(mean_base)
      booster_newton, curve_newton = train(newton_base)

      target = uniform_expectile(alpha)
      pred_mean = float(booster_mean.predict(dtrain).mean())
      pred_newton = float(booster_newton.predict(dtrain).mean())

      key_rounds = [0, 1, 2, 5, 10, 20, 50, 100, 199]
      metrics = [(i, curve_mean[i], curve_newton[i]) for i in key_rounds]
      return (mean_base, newton_base, target, pred_mean, pred_newton, curve_mean[-1], curve_newton[-1], metrics)

  for alpha in [0.1, 0.2, 0.5, 0.8, 0.9]:
      mean_base, newton_base, target, pred_mean, pred_newton, m_final, n_final, metrics = run(alpha)
      print(f"alpha={alpha:.1f}, target={target:.6f}, mean_base={mean_base:.6f}, newton_base={newton_base:.6f}")
      print(f"  final pred mean={pred_mean:.6f}, final pred newton={pred_newton:.6f}")
      for i, m, n in metrics:
          print(f"    round {i:3d}: metric mean={m:.6f}, metric newton={n:.6f}")

```

Result: both approaches converge to the same final expectile, but the 1‑step Newton base_score gives a much lower initial expectile metric and reaches the plateau in fewer rounds (especially for extreme alphas).

  - alpha=0.1
    round   0: 0.034396 (mean) vs 0.019239 (1-step Newton)
    round   1: 0.029395 vs 0.019061
    round   2: 0.025943 vs 0.018944
    round   5: 0.020883 vs 0.018786
    round  10: 0.018988 vs 0.018734
    round  20: 0.018730 vs 0.018727

  - alpha=0.9
    round   0: 0.034396 vs 0.019239
    round   1: 0.029395 vs 0.019061
    round   2: 0.025943 vs 0.018944
    round   5: 0.020883 vs 0.018786
    round  10: 0.018988 vs 0.018734
    round  20: 0.018730 vs 0.018727

-> I modified the PR to now run the 1-step Newton.

Let me know if there's anything else you would like to see!

tests/python/test_eval_metrics.py
```diff
@@ -221,3 +222,42 @@ def test_precision_score(self) -> None:
...
     def test_quantile_error(self) -> None:
         check_quantile_error("hist", "cpu")
+
+    @pytest.mark.skipif(**tm.no_sklearn())
+    def test_expectile_error(self) -> None:
+        check_expectile_error("hist", "cpu")
+
+    def test_expectile_uniform_convergence(self) -> None:
+        rng = np.random.default_rng(42)
+        n_samples = 50_000
```
RAMitchell: n_samples is a bit high. Lets make sure the test is lightweight and fast.

RAMitchell (approved):
Very nice PR.

doc/parameter.rst
```diff
@@ -515,6 +518,16 @@ Parameter for using Quantile Loss (``reg:quantileerror``)
 
     .. versionadded:: 2.0.0
 
+Parameter for using Expectile Loss (``reg:expectileerror``)
+===========================================================
+
+* ``expectile_alpha``: A scalar or a list of targeted expectiles. Range: [0, 1]. Required for
+  ``reg:expectileerror``.
+
+    .. versionadded:: 3.2.0
```
trivialfis: Let's move this to 3.3. I will help support vector-leaf/multi-target.

src/objective/regression_obj.cu
```diff
@@ -394,6 +396,157 @@ XGBOOST_REGISTER_OBJECTIVE(PseudoHuberRegression, "reg:pseudohubererror")
...
+                                 &label_mean);
+    }
+    CHECK_EQ(label_mean.Size(), 1);
+
+    auto mean_host = label_mean.HostView();
+    auto h_labels = info.labels.HostView();
+    auto h_weights = info.weights_.ConstHostSpan();
+    auto const& alpha = param_.expectile_alpha.Get();
+
+    std::vector<double> sums(2 * n_targets, 0.0);
```
trivialfis, line 486: Let me handle this for GPU after merging. ;-)

trivialfis (commented):
Thank you for the work on the expectile objective. Please rebase your branch to master and see if the CI can pass.

doc/parameter.rst
matthieubulte: Changed the version in the latest commit. Let me know about the vector-leaf next steps.

tests/python/test_eval_metrics.py
```diff
@@ -221,3 +222,42 @@ def test_precision_score(self) -> None:
...
+        }
+        num_boost_round = 200
+        atol = 1e-2
+
+        for alpha in [0.1, 0.5, 0.9]:
+            params["expectile_alpha"] = alpha
+            booster = xgb.train(params, dtrain, num_boost_round=num_boost_round)
+            pred = float(booster.predict(dtrain).mean())
+            expected = uniform_expectile(alpha)
+        np.testing.assert_allclose(pred, expected, rtol=atol, atol=atol)
```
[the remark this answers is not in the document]
matthieubulte: Good catch, fixed.

src/metric/elementwise_metric.cu
```diff
@@ -519,4 +520,84 @@ class QuantileError : public MetricNoCache {
...
+  }
+
+  const char* Name() const override { return "expectile"; }
+  void LoadConfig(Json const& in) override {
+    auto const& obj = get<Object const>(in);
+    auto it = obj.find("expectile_loss_param");
+    if (it != obj.cend()) {
+      FromJson(it->second, &param_);
+      auto const& name = get<String const>(in["name"]);
+      CHECK_EQ(name, "expectile");
```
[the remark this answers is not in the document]
matthieubulte: Fixed and added a unit test to make sure that serializing -> loading yields the expected results.
I check that the result is correct by computing the loss on 2 data points and comparing that what we know should be the output for the chosen expectile.

src/objective/regression_obj.cu
```diff
@@ -7,6 +7,7 @@
 #include <dmlc/omp.h>
 
 #include <algorithm>  // for all_of
+#include <array>      // for array
```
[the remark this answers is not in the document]
matthieubulte: Good catch, fixed.

tests/python/test_eval_metrics.py
matthieubulte: Lowered to 1000 and re-ran the test a few times manually to make sure it's stable, seems good :)

matthieubulte:
Thanks for the feedback. I merged the latest master and addressed your and copilot's comments.

I still have some failing tests, but I don't think they are related to what I added?

trivialfis:
Let me fix it later today, the R build needs to register the new files.

RAMitchell:
Thanks for fixing review comments. 

For CI, install pre-commit and run like `pre-commit run --from-ref upstream/main --to-ref HEAD`.

Looks like there is also an undefined reference build error in the R package due to this PR.

trivialfis:
I fixed the R build errors. Please help address the rest of the test failures. Feel free to ping me if you run into any issue.

matthieubulte:
Thanks for fixing the R errors.

I think all failing tests were caused by the fact that I inadvertently set `verbosity=0` globally from `test_expectile_uniform_convergence`. 

Should all be good now. At least I was able to reproduce -> get to green on my machine :)

PR #11991, opened 2026-02-05
Cleanup the use of omp clauses

trivialfis:
- Avoid the use of raw omp clauses, use `ParallelFor` and its variants instead.
- Methods like `omp_get_thread_num` are still used, we will replace them later.


trivialfis:
cc @rongou

hcho3 (approved):

PR #11992, opened 2026-02-05
[doc] Update NUMA configuration for external memory.

trivialfis:
- Add NUMA configuration with pyhwloc.
- Remove RMM configuration to cleanup the demo.


trivialfis:
cc @rongou

PR #11993, opened 2026-02-05
[ci] disable pylint import order check.

trivialfis:
Let's leave that to ruff. They don't necessarily agree with each other.


trivialfis:
cc @RAMitchell

PR #11998, opened 2026-02-06
[EM] Support mixed GPU models on SNMG.

trivialfis:
- Use UUID to query C2C info, to support mixed GPU models.
- Drop the full-host-cache path to utilize large GPU memory.


trivialfis:
cc @rongou

PR #12009, opened 2026-02-10
[pyspark] Update pyspark GPU tests

RAMitchell:
Just use a different fixture to launch tests on GPU, we don't need to duplicate the tests.

### DIFF
--- a/tests/test_distributed/test_with_spark/test_spark.py
+++ b/tests/test_distributed/test_with_spark/test_spark.py
@@ -48,11 +48,7 @@ def _probe_gpu_addresses() -> List[str]:
 
     info = xgb.build_info()
     use_cuda = _to_bool(info.get("USE_CUDA", False))
-    use_nccl = _to_bool(info.get("USE_NCCL", False)) or _to_bool(
-        info.get("USE_DLOPEN_NCCL", False)
-    )
-
-    if not (use_cuda and use_nccl):
+    if not use_cuda:
         return []
 
     try:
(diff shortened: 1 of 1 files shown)


tests/test_distributed/test_with_spark/test_spark.py
```diff
@@ -217,34 +278,61 @@ def test_regressor(
...
+                spark_regressor.training_summary.validation_objective_history["rmse"]
+            )
+            assert len(validation_history) > 0
+            assert np.isfinite(validation_history).all()
+            assert np.allclose(
+                spark_regressor._xgb_sklearn_model.best_score,
+                min(validation_history),
+                atol=1e-6,
+            )
+        else:
```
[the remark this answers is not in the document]
RAMitchell: GPU model is not identical here between spark and sklearn API so tests have been loosened.

ops/pipeline/test-python-wheel.sh
```diff
@@ -107,7 +107,8 @@ case "$suite" in
     python -c 'from cupy.cuda import jitify; jitify._init_module()'
     pytest -v -s -rxXs --durations=0 -m 'mgpu' tests/python-gpu
     pytest -v -s -rxXs --durations=0 tests/test_distributed/test_gpu_with_dask
-    pytest -v -s -rxXs --durations=0 tests/test_distributed/test_gpu_with_spark
+    pytest -v -s -rxXs --durations=0 tests/test_distributed/test_with_spark/test_data.py -k dmatrix_ctor_gpu
```
trivialfis, line 110: Is it still possible to separate the GPU tests into a different module and make it easier for others to run, without digging into the CI scripts to find and memorize the right incantation?

I moved many tests into the `python-package/xgboost/testing/` to share tests between CPU and GPU.

RAMitchell: Its possible. I actually have some thoughts about this after my work on legate-boost. I think all gpu/cpu tests should ideally be combined and parameterised on device. If we expect the same behaviour we should have the same tests. It also a matter of duplication. This is obviously not a trivial change so we can discuss more.

trivialfis: Yeah, I did something like this before:
```
import torch as tr

def devices() -> list[tr.device]:
    """Available devices for tests."""
    return [cpu(), gpu()] if tr.cuda.is_available() else [cpu()]

@pytest.mark.parametrize("device", devices())
def my_test(device: tr.device) -> None:
    pass
```
Might be useful.

We can merge if you intend to follow up on this. ;-)

Issue #12008, opened 2026-02-10
Customizable worker port for the collective implementation.


trivialfis:
Currently, XGBoost uses an available port returned by querying the OS. To support a more restricted networking environment, we need to add a user option to specify the worker port. This has been implemented for the tracker, but not for workers yet.

Note:
Does NCCL not yet support a user port for TCP sockets? `NCCL_IB_HCA` accepts a port.

Related:
- https://github.com/dmlc/xgboost/issues/11573
- https://github.com/dmlc/xgboost/issues/11757
- https://github.com/dmlc/xgboost/issues/11551
- https://github.com/dmlc/xgboost/issues/9792

PR #12010, opened 2026-02-10
[coll] Support worker port.

trivialfis:
close: https://github.com/dmlc/xgboost/issues/12008

- Provide a `worker_port` parameter for the `Config`. The `Config` object is an interface that glues the distributed frameworks to the XGBoost collective and doesn't provide any functionality.
- Accept user-provided port number in the collective.

**todos**
- [x] Find a way to support port for each worker. This is necessary for SNMG.


trivialfis:
cc @RAMitchell please help take a look when you are available.

RAMitchell (approved):
I can't say too much about the networking side of things but this pr looks good to me and is not very intrusive.

PR #12011, opened 2026-02-10
Fix SketchBatchNumElements test to clamp with kIntMax

### LINKED ISSUE #9716: SketchBatchNumElements minor bugs
1) In xgboost/blob/master/src/common/hist_util.cu, SketchBatchNumElements,

    auto avail = dh::AvailableMemory(device) * 0.8;
    this is a **double**, and this function returns size_t.
    return value,   return **std::min**(sketch_batch_num_elements, kIntMax);

2) In xgboost/blob/master/tests/cpp/common/test_hist_util.cu, TEST(HistUtil, SketchBatchNumElements),

      auto avail = static_cast<size_t>(dh::AvailableMemory(device) * 0.8);
      this is correct, not double. However, "avail_elem" isn't compared with KIntMax, which can be a very large integer, which cause the following code fails,

      "ASSERT_EQ(batch, avail_elem);"

Thanks.

ashutoshk314:
This PR fixes a mismatch between production and test logic in
SketchBatchNumElements. The test now clamps the expected value
with kIntMax, matching the behavior of the implementation.

Fixes #9716


trivialfis (approved):

trivialfis:
Please revert the pre-commit fixes.

PR #12013, opened 2026-02-11
Fix compilation with the latest CCCL.

trivialfis:
- Fix CMake script to pick up the `cuda-cccl` from conda.
- Replace the `thrust::tuple`, except when it's used by `zip_iterator`. Only thrust>=3.0 uses `cuda::std::tuple` in the `zip_iterator`.

CI log: https://github.com/dmlc/xgboost/actions/runs/21935620994/job/63349074776?pr=12013

**todos**
- [x] Figure out what to do with the cpp lint false positive.
- [x] Revert the CI pipeline changes.

### DIFF
--- a/ops/script/lint_cpp.py
+++ b/ops/script/lint_cpp.py
@@ -70,7 +70,6 @@ class Lint:
                     "-build/c++11",
                     "-build/include,",
                     "+build/namespaces",
-                    "+build/include_what_you_use",
                 ]
             )
         )


trivialfis:
cc @RAMitchell The `cpplint` `include-what-you-use` check doesn't distinguish the `tuple` from `cuda::std::` and from `std::`, and makes wrong suggestions about include. I would like to disable the `include-what-you-use-check` with cpplint instead of working around it.

trivialfis:
cc @hcho3 Please help review when you are available.

hcho3 (approved):

PR #12017, opened 2026-02-14
demo: remove type ignore in custom_softmax plot history

pdwi2020:
## Summary
- remove one `# type: ignore` from `demo/guide-python/custom_softmax.py`
- add explicit axis typing/casts around `matplotlib.pyplot.subplots` return value to satisfy static typing

## Why
This is an incremental typing improvement for Python demos, aligned with ongoing work in #6496.

## Validation
- `python3 -m py_compile demo/guide-python/custom_softmax.py`
- Could not run `mypy` locally because it is not installed in this environment.

Part of #6496

### DIFF
--- a/demo/guide-python/custom_softmax.py
+++ b/demo/guide-python/custom_softmax.py
@@ -12,9 +12,10 @@ detailed tutorial and notes.
 """
 
 import argparse
-from typing import Dict, Tuple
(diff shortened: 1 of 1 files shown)


trivialfis:
I see that this is done by codex. Nothing against it, but is there a human to verify and confirm the change?

PR #12022, opened 2026-02-16
Move predictor mutex into GBTreeModel

RAMitchell:


trivialfis (approved):
My preference is not to have the mutex in the "model" and keep the "model" minimal in terms of mutable states. The model itself is a large state that changes in a "commit" style (in the sense of a database transaction).

My original plan was to create a task queue, similar to Rust's mpsc crate (https://doc.rust-lang.org/std/sync/mpsc/), where one pushes model views and tasks into the queue in a message-passing style, thereby eliminating the need for traditional mutexes. But since I haven't been able to prioritize it, these are just empty words. I will leave it for you to decide.

The PR looks good to me.

RAMitchell:
Noted. I think this is currently better than the alternative, which is having mutexes lying around in predictor.

PR #12029, opened 2026-02-19
[ci] Re-enable dependabot for github actions.

trivialfis:
We disabled it due to the JVM package's Maven dependencies. But I would like to keep the github actions we use up-to-date.


trivialfis:
cc @hcho3

Issue #12023, opened 2026-02-16
[RFC] Proposals for "gblinear"


RAMitchell:
The `booster='gblinear'` functionality in XGBoost has a number of long-standing issues, and I think it’s time to reconsider its role in the broader library.

A selection of existing issues:
- https://github.com/dmlc/xgboost/issues/2108
- https://github.com/dmlc/xgboost/issues/10893
- https://github.com/dmlc/xgboost/issues/9936
- https://github.com/dmlc/xgboost/issues/3747
- https://github.com/dmlc/xgboost/issues/3261
- https://github.com/dmlc/xgboost/issues/4919
- https://github.com/dmlc/xgboost/issues/11473
- https://github.com/dmlc/xgboost/issues/11183
- https://github.com/dmlc/xgboost/issues/10732
- https://github.com/dmlc/xgboost/issues/11106
- https://github.com/dmlc/xgboost/issues/10819
- https://github.com/dmlc/xgboost/issues/5379
- https://github.com/dmlc/xgboost/issues/8848

Historically, XGBoost has focused most feature development on tree models. As a result, many features and parameters are **not supported** by the linear model—often silently and without clear documentation. I also think we have invested relatively little effort into the performance and production-readiness of the `gblinear` solvers. For example, the CPU solver uses a “shotgun”-type method, which is non-deterministic and may cause problems in production settings where determinism is required.

In summary, I don’t think `gblinear` in its current form is sufficiently robust or well supported to justify its inclusion as-is. I see two viable paths: deprecate and remove it, or separate it into a new, explicit interface.

## 1) Deprecation

Issue a deprecation warning and stage removal for the next minor release.

### Advantages
- Less work for maintainers
- Less confusion for users
- More focus on improvements to the core tree functionality
- Opportunities to simplify internal refactors

### Disadvantages
- Existing users—especially production deployments—would need to migrate (e.g., to scikit-learn linear models or R’s GLM tooling).
- These alternatives may not implement the same objectives as XGBoost, and comparisons with XGBoost tree models may require additional effort.

## 2) New interface

Create dedicated Python classes, e.g. `XGBLinearRegressor` / `XGBLinearClassifier`. These classes would expose only the parameters supported by linear models, with more specific documentation and stronger validation (e.g., error on tree-only params). We would also invest effort into making them comparable in features and usability to other popular GLM libraries.

### Advantages
- Reuse existing objectives/metrics in XGBoost
- Continue supporting linear models as a convenience feature within XGBoost

[...]

PR #12030, opened 2026-02-19
Deprecate gblinear

RAMitchell:
Address #12023

### DIFF
--- a/ops/conda_env/win64_test.yml
+++ b/ops/conda_env/win64_test.yml
@@ -11,7 +11,7 @@ dependencies:
 - pytest
 - boto3
 - hypothesis
-- cupy>=13.2
+- cupy>=13.2,<14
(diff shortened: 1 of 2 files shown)


ops/conda_env/win64_test.yml
```diff
@@ -11,7 +11,7 @@ dependencies:
 - pytest
 - boto3
 - hypothesis
-- cupy>=13.2
+- cupy>=13.2,<14
```
trivialfis, line 14: I tested locally with the GPU prediction test module, but couldn't reproduce the failure. We need to unfreeze this, otherwise it will be forgotten.

RAMitchell: Its Windows only. I will clear this up with cupy directly.

Issue #12031, opened 2026-02-19
Make dart booster part of gbtree


RAMitchell:
We are deprecating "booster=gblinear" (#12023 ) , that leaves only "gbtree" and "dart". Internally dart is really just a thin skin over gbtree that modifies tree weights. We can instead make dart just a parameter of gbtree.

This reduces complexity around GBTree/Learner and supports future refactoring.

There are risks however around backwards compatibility and breaking changes, so care is required.

RAMitchell:
Now that I am looking at the documentation for dart I am wondering why dropout is applied for non-training predict automatically? We know when we are training or not so this is an unusual default.

trivialfis:
Can we close this along with https://github.com/dmlc/xgboost/pull/12032 ?

PR #12032, opened 2026-02-19
Merge dart into gbtree

RAMitchell:
Implement #12031 

As a follow up I would like prediction to simply support tree weights, then we don't need any unique predict functions. We can also easily support weights in GPU shap by modifying leaf weights before they go into GPUTreeShap.

### DIFF
--- a/src/predictor/gpu_predictor.cu
+++ b/src/predictor/gpu_predictor.cu
@@ -435,7 +435,8 @@ auto MakeTreeSegments(Context const* ctx, bst_tree_t tree_begin, bst_tree_t tree
 void ExtractPaths(Context const* ctx,
                   dh::device_vector<gpu_treeshap::PathElement<ShapSplitCondition>>* paths,
                   gbm::GBTreeModel const& h_model, DeviceModel const& d_model,
-                  dh::device_vector<uint32_t>* path_categories) {
+                  dh::device_vector<uint32_t>* path_categories,
+                  common::Span<float const> tree_weights) {
   curt::SetDevice(ctx->Ordinal());
 
   // Path length and tree index for all leaf nodes
@@ -482,6 +483,8 @@ void ExtractPaths(Context const* ctx,
   auto d_info = info.data().get();
   auto d_tree_groups = d_model.tree_groups;
   auto d_path_segments = path_segments.data().get();
+  auto d_tree_weights = tree_weights;
+  CHECK(d_tree_weights.empty() || d_tree_weights.size() == d_trees.size());
 
   std::size_t max_cat = 0;
   if (std::any_of(h_model.trees.cbegin(), h_model.trees.cend(),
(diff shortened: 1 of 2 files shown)


src/gbm/gbtree.cc
```diff
@@ -405,48 +468,79 @@ void GBTree::LoadConfig(Json const& in) {
...
+  auto& gbtree = *p_gbtree;
+  gbtree["name"] = String("gbtree");
+  gbtree["model"] = Object();
+  auto& model = gbtree["model"];
   model_.SaveModel(&model);
+  SaveWeightDrop(weight_drop_, &gbtree);
+  if (this->IsDart()) {
+    // Keep legacy field for external consumers during transition.
+    SaveWeightDrop(weight_drop_, &out);
+  }
```
chyunsu3, line 543: We should have an explicit policy about forward and backward compatibility for tree models. That is, can a version X of XGBoost load a model generated by version Y ?

So far, we've operated on the assumption that older models can be read by newer XGBoost (X > Y).

Based on this comment, I suppose we'd make breaking changes to the model format for every major version bump?

trivialfis:
I'm a bit wary about this change from a user's perspective. No objection to code unification.

I really don't want to make unnecessary breaking changes. Aside from this, I think having a different booster type at the parameter level (not necessarily at the internal class level) helps clarify different training paradigms.

RAMitchell:
@trivialfis its a pretty mild change. If you had code that trained a dart booster it would work the same, the selection of booster would then just be unecessary and emit a warning.

I don't see it as a different training paradigm. Its just subsampling on trees and its implementation doesn't justify a separate class at such a core level.

The eventual removal of GBDart and GBLinear actually makes the learner redundant - GBTree and Learner get merged and much of the complexity disappears.

src/gbm/gbtree.cc
RAMitchell: Yes I think a reasonable expectation is that your saved model may not work across major versions. In fact sklearn makes no guarantees at all across different versions.

I see our docs say the following:
`We guarantee backward compatibility for models but not for memory snapshots.`

I think its unreasonable to guarantee backwards compatbility over major versions, this is too strict a promise.

chyunsu3: > a reasonable expectation is that your saved model may not work across major versions

Many users came to expect backwards compatibility for XGBoost models. I've seen people trying to load models from ancient versions of XGBoost. This is reason why it took us so long to drop support for the legacy binary format.

So if XGBoost 4.0 suddenly throws an error when reading XGBoost 3.x model files, many folks will be upset. If we ever decide to make breaking changes, we should have multiple versions where the logic for loading legacy layout would be retained. With a prominent warning, so that users can get used to re-saving models every now and then.

chyunsu3: This discussion (about dropping backwards compatibility) should happen in a separate dedicated issue. Let me create a new issue.

RAMitchell: I think this is a problem with peoples expectations or documentation in this case. I can't think of any other library that makes such a strong promise. I had a look around what others are doing.

Pytorch makes no promise across versions, keras across minor versions only, jax says up to 6 months.

Catboost and LightGBM do not explicitly have a policy, although breakages do occur.

trivialfis: I think we will adhere to the compatibility between versions. Nowadays, model training and deployment are two separate processes handled by different teams. They don't always sync between versions and it's difficult to create such requirement.

> Pytorch makes no promise across versions

Considering that no one is training LLM from scratch now, I'm sure there's a way to manage versions for DL libraries.

trivialfis: There needs to be a deprecation process and buffer releases.

RAMitchell:
This PR needs a bit more work but I think I can do it without affecting backward compatibility.

PR #12033, opened 2026-02-19
Simplifications to Quantile Sketching

RAMitchell:
This is a first pass at reducing spaghetti code in quantile sketching. I have removed the version of sketch only used by the approx tree method. If the first is good enough for hist then it is good enough for approx.

### DIFF
--- a/src/common/quantile.h
+++ b/src/common/quantile.h
@@ -13,7 +13,6 @@
 #include <algorithm>
 #include <cmath>
 #include <cstring>
-#include <iostream>
 #include <set>
 #include <vector>
 
@@ -47,26 +46,10 @@ struct WQSummary {
     // constructor
     XGBOOST_DEVICE Entry(RType rmin, RType rmax, RType wmin, DType value)
         : rmin(rmin), rmax(rmax), wmin(wmin), value(value) {}
-    /*!
-     * \brief debug function,  check Valid
-     * \param eps the tolerate level for violating the relation
-     */
-    void CheckValid(RType eps = 0) const {
-      CHECK(rmin >= 0 && rmax >= 0 && wmin >= 0) << "nonneg constraint";
-      CHECK(rmax - rmin - wmin > -eps) << "relation constraint: min/max";
-    }
     /*! \return rmin estimation for v strictly bigger than value */
     XGBOOST_DEVICE RType RMinNext() const { return rmin + wmin; }
     /*! \return rmax estimation for v strictly smaller than value */
     XGBOOST_DEVICE RType RMaxPrev() const { return rmax - wmin; }
-
-    friend std::ostream &operator<<(std::ostream &os, Entry const &e) {
-      os << "rmin: " << e.rmin << ", "
-         << "rmax: " << e.rmax << ", "
-         << "wmin: " << e.wmin << ", "
-         << "value: " << e.value;
-      return os;
-    }
   };
   /*! \brief data field */
   Entry *data;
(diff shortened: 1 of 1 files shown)


src/common/quantile.h
```diff
@@ -68,64 +68,16 @@ struct WQSummary {
...
-        i = j;
-      }
-    }
-  };
-  /*! \brief data field */
-  Entry *data;
-  /*! \brief number of elements in the summary */
-  size_t size;
+  /*! \brief summary entries */
+  Span<Entry> data;
```
trivialfis: The span class in XGBoost has indexing check, which might not be trivial when in a hot loop.

RAMitchell: I made this PR to complicated I think. I will revert the span change for now and just focus on the unification of the sketch type used between approx/hist.

src/common/quantile.h
```diff
@@ -341,189 +295,131 @@ struct WQSummary {
...
-            mrange += src.data[i].RMaxPrev() - src.data[bid].RMinNext();
-          }
-          bid = i;
-          ++nbig;
-        }
+
+  // push element to the queue, return false if the queue is full and need to be flushed
+  inline bool Push(DType x, RType w) {
+    if (qtail == 0 || queue[qtail - 1].value != x) {
+      if (qtail == queue.size() && queue.size() == 1) {
```
trivialfis, line 236: What's the purpose of this special condition? Why do we resize the buffer under this specific condition?

I tracked the condition to the deleted comment about lazy memory allocation. Could you please keep the comment (probably elaborate on it)?

src/common/quantile.h
```diff
@@ -341,189 +295,131 @@ struct WQSummary {
...
-          if (bid != i - 1) {
-            // accumulate the range of the rest points
-            mrange += src.data[i].RMaxPrev() - src.data[bid].RMinNext();
-          }
-          bid = i;
-          ++nbig;
-        }
+
+  // push element to the queue, return false if the queue is full and need to be flushed
+  inline bool Push(DType x, RType w) {
```
trivialfis: No need for `inline` keyword.

src/common/quantile.h
```diff
@@ -341,189 +295,131 @@ struct WQSummary {
...
-    template <typename TStream>
-    inline void Load(TStream &fi) {  // NOLINT(*)
-      CHECK_EQ(fi.Read(&this->size, sizeof(this->size)), sizeof(this->size));
-      this->Reserve(this->size);
-      if (this->size != 0) {
-        CHECK_EQ(fi.Read(this->data, this->size * sizeof(Entry)), this->size * sizeof(Entry));
-      }
+  }
+  template <typename TStream>
+  inline void Load(TStream &fi) {  // NOLINT(*)
```
trivialfis: The `Save` and `Load` are not used, maybe delete them as well.

src/common/quantile.h
```diff
@@ -341,189 +295,131 @@ struct WQSummary {
...
+      queue[qtail++] = QEntry(x, w);
+      return true;
     }
-    // assert: there cannot be more than n big data points
-    if (nbig >= n) {
-      // see what was the case
-      LOG(INFO) << " check quantile stats, nbig=" << nbig << ", n=" << n;
-      LOG(INFO) << " srcsize=" << src.size << ", maxsize=" << maxsize << ", range=" << range
-                << ", chunk=" << chunk;
-      src.Print();
```
trivialfis, line 407: The `Print` function is no longer used after this PR, maybe remove the definition as well?

RAMitchell: Actually all of these are unused (!!):
WQSummary::MaxError
WQSummary::Query
WQSummary::MaxRank
WQSummary::MakeFromSorted
WQuantileSketch::PushSummary
WQSummary::Entry::CheckValid
WQSummary::CheckValid
WQuantileSketch::CheckValid

src/common/quantile.h
RAMitchell: I think its trying to optimise the case of one sketch element e.g. a one hot encoded column. I think this optimisation might be a bit unecessary - I think even vector push back is fine, the container will quickly get up to size.

PR #12038, opened 2026-02-21
[CI] Avoid push runs caused by dependabot.

trivialfis:
Only run tests on persistent branches.

related: https://github.com/dmlc/xgboost/pull/12034


trivialfis:
cc @hcho3

hcho3 (approved):

PR #12044, opened 2026-02-24
Add tutorial document for federated xgboost

ZiyueXu77:
and update docstrings

### DIFF
--- a/doc/tutorials/federated_learning.rst
+++ b/doc/tutorials/federated_learning.rst
@@ -215,6 +215,7 @@ is the default for row-split).
 
 .. code-block:: python
 
+    import pandas as pd
     import xgboost as xgb
 
     rank = 0  # 0, 1, ..., world_size - 1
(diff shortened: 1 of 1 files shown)


doc/tutorials/federated_learning.rst
```diff
@@ -0,0 +1,846 @@
...
+    communicator_env = {
+        "dmlc_communicator": "federated",
+        "federated_server_address": "localhost:9091",
+        "federated_world_size": 3,
+        "federated_rank": rank,
+    }
+
+    with xgb.collective.CommunicatorContext(**communicator_env):
+        train_path = f"/tmp/xgboost/federated/horizontal/site-{rank + 1}/train.csv"
+        valid_path = f"/tmp/xgboost/federated/horizontal/site-{rank + 1}/valid.csv"
```
trivialfis, line 232: Any plan to remove the use of `csv` text file and consume the in-core data like dataframe directly?

ZiyueXu77: csv may still be a quite common format users adopt, and this is only for example purpose. Ultimately it depends on what xgb.DMatrix supports (and deprecates) - we can easily adjust accordingly at data preparation step.

trivialfis: We would like to remove the text parser from XGBoost and delegate the task of loading data to third party libraries (like pandas)

ZiyueXu77: sure sounds good! let me add a step to make it future proof, I will follow [the current doc](https://xgboost.readthedocs.io/en/release_3.2.0/python/python_intro.html#data-interface) below and use pandas to load the csv

data = pandas.DataFrame(np.arange(12).reshape((4,3)), columns=['a', 'b', 'c'])
label = pandas.DataFrame(np.random.randint(2, size=4))
dtrain = xgb.DMatrix(data, label=label)

ZiyueXu77: updated

trivialfis:
The branch is too old to have any meaningful CI run. I will just merge it into the targeted branch for now.

PR #12046, opened 2026-02-26
Wqsummary use span

RAMitchell:


trivialfis:
Not entirely sure about the benefit of the change, it seems you are replacing the `data, size` in the summary with `span{data, size}, current_size`, and the span here doesn't help much, as the code doesn't use its method very much.

RAMitchell:
The code is poorly written because the extent of the data is not described. We have the number of elements in the sketch but not the limits of the reserved sketch. Span should be used to represent the buffer extents. I will use this in further work.

PR #12047, opened 2026-02-26
Reduce memory in GPU quantile sketch for external memory

RAMitchell:
Accumulation of exponentionally increasing numbers of sketch containers here is unecessary. Simply merging and pruning GK sketches for each batch maintains the error bound.

### DIFF
--- a/src/common/quantile.cu
+++ b/src/common/quantile.cu
@@ -505,6 +505,10 @@ void SketchContainer::AllReduce(Context const *ctx, bool is_column_split) {
   }
 
   timer_.Start(__func__);
+  // Bound local sketch size before exchanging data across workers.
+  auto intermediate_num_cuts = static_cast<bst_idx_t>(num_bins_ * kFactor);
+  this->Prune(ctx, intermediate_num_cuts);
+
   auto d_columns_ptr = this->columns_ptr_.ConstDeviceSpan();
   CHECK_EQ(d_columns_ptr.size(), num_columns_ + 1);
   size_t n = d_columns_ptr.size();


RAMitchell:
I did an experiment:
```
Results:

peak_memory_bytes: baseline 224,375,872 vs candidate 5,219,056 (-97.6740%)
max_rel_rank_error: baseline 0.01122284 vs candidate 0.01026154 (-8.5656%)
mean_rel_rank_error: baseline 0.00470723 vs candidate 0.00428144 (-9.0454%)
Run metadata:

n_rows=131072, n_features=64, n_batches=256, max_quantile_blocks=2147483647
```

src/common/quantile.cuh
```diff
@@ -199,6 +199,7 @@ class SketchContainer {
   HostDeviceVector<FeatureType> const& FeatureTypes() const { return feature_types_; }
 
   Span<OffsetT const> ColumnsPtr() const { return this->columns_ptr_.ConstDeviceSpan(); }
+  void SetNumRows(bst_idx_t n_rows) { this->num_rows_ = n_rows; }
```
trivialfis: Minor issue, can this be handled in the `Push` method instead of a separate method?

RAMitchell: Actually this n_rows member is not needed. I just removed it.

trivialfis:
The use of global num rows is something inherited from the original CPU implementation, along with the safe factor, if memory serves. It makes sense to remove it since the push method would have pruned it already. But it has been challenging for me to keep track of the accuracy for the sketching algorithm. Might be a good time for me to actually go through it again after this PR.

PR #12048, opened 2026-02-27
[mt] Unify the quantisers.

trivialfis:
- Support distributed training.
- Host device access. Host access is for single-target. We will migrate to device-only once we can unify the code base between single-target and multi-target more.


trivialfis:
cc @rongou

PR #12049, opened 2026-02-27
More quantile refactoring

RAMitchell:


RAMitchell:
After benchmarking, I think the performance reduction of around 50% is too high, so I will just refactor instead and we can remove if/when approx gets removed.

PR #12054, opened 2026-03-02
Simplify quantile sketch internals

RAMitchell:


src/common/quantile.h
```diff
@@ -366,12 +401,44 @@ struct Queue {
...
+  WQSummaryContainer(WQSummaryContainer &&src) noexcept
+      : WQSummary<>(Span<WQSummary<>::Entry>{}, 0), space{std::move(src.space)} {
+    this->data = {dmlc::BeginPtr(this->space), this->space.size()};
     this->current_elements = src.current_elements;
+    src.current_elements = 0;
+    src.data = Span<WQSummary<>::Entry>{};
   }
-  WQSummaryContainer() : WQSummary<>(Span<WQSummary<>::Entry>{}, 0) {}
+
+  WQSummaryContainer &operator=(WQSummaryContainer const &src) {
```
trivialfis: Is the copy used in non-test code? Skimming through the code, I saw a right value assignment, but haven't found copying yet.

RAMitchell: Thanks, yeah it can be removed.

PR #12055, opened 2026-03-02
Make linux packaging easier

bkmgit:
- Enable use of system dmlc see https://docs.fedoraproject.org/en-US/packaging-guidelines/#bundling
- Add soname to shared library see https://docs.fedoraproject.org/en-US/packaging-guidelines/#_shared_libraries

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -480,7 +480,7 @@ jobs:
             --train --model-path cross_platform_model.ubj
       - name: Upload cross-platform model artifact
         if: matrix.cuda_version == 12 && matrix.suite == 'gpu'
-        uses: actions/upload-artifact@v6.0.0
+        uses: actions/upload-artifact@v7.0.0
         with:
           name: cross-platform-model
           path: cross_platform_model.ubj
@@ -563,7 +563,7 @@ jobs:
         run: |
           python -m pip install -vvv wheelhouse/*.whl
       - name: Upload wheel artifact
-        uses: actions/upload-artifact@v6.0.0
+        uses: actions/upload-artifact@v7.0.0
         with:
           name: python-wheel-${{ matrix.platform_id }}
           path: wheelhouse/*.whl
@@ -602,7 +602,7 @@ jobs:
           environment-file: ops/conda_env/minimal.yml
       - run: conda install scikit-learn numpy -y
       - name: Download macOS wheel artifact
-        uses: actions/download-artifact@v7.0.0
+        uses: actions/download-artifact@v8.0.0
         with:
           name: python-wheel-${{ matrix.platform_id }}
           path: wheelhouse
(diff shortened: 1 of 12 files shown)


trivialfis:
I understand that distributions frown upon submodules, but the coupling between XGBoost and dmlc core is tie at the moment. We bump dmlc to fix issues there. Are you quite sure using a system one is desirable?

bkmgit:
Yes, this is ok. Both can be upgraded at the same time.

trivialfis:
ref https://github.com/dmlc/dmlc-core/pull/720

I will wait for @hcho3 with the dmlc PR.

CMakeLists.txt
```diff
@@ -379,6 +390,9 @@ if(BUILD_STATIC_LIB)
   add_library(xgboost STATIC)
 else()
   add_library(xgboost SHARED)
+  set_target_properties(xgboost PROPERTIES
```
trivialfis, line 394: Why is this necessary?

CMakeLists.txt
```diff
@@ -388,11 +402,16 @@ target_include_directories(xgboost
 #-- End shared library
 
 # Common setup for all targets
-foreach(target xgboost objxgboost dmlc)
+foreach(target xgboost objxgboost)
   xgboost_target_properties(${target})
   xgboost_target_link_libraries(${target})
   xgboost_target_defs(${target})
 endforeach()
+if(NOT BUILD_WTIH_DMLC_SYSTEM)
```
trivialfis: Why is this needed, what's changed?

CMakeLists.txt
```diff
@@ -54,6 +54,7 @@ option(BUILD_STATIC_LIB "Build static library" OFF)
 option(BUILD_DEPRECATED_CLI "Build the deprecated command line interface" OFF)
 option(FORCE_SHARED_CRT "Build with dynamic CRT on Windows (/MD)" OFF)
 option(BUILD_WITH_GIT_HASH "Add a short git hash to the build info." OFF)
+option(BUILD_WITH_DMLC_SYSTEM "Use system package for dmlc." OFF)
```
trivialfis: maybe `SYSTEM_DMLC` seems better?

CMakeLists.txt
chyunsu3: This adds versioning to *.so files. See https://docs.fedoraproject.org/en-US/packaging-guidelines/#_shared_libraries

CMakeLists.txt
bkmgit: DMLC does not need to be built and installed if it is already on the system and the system version is used.

CMakeLists.txt
bkmgit: Done.

trivialfis:
The GHA test is being throttled by GH?

Issue #7693, opened 2022-02-22
[RFC] Exposing objectives and metrics as part of the API.


trivialfis:
This is for discussion about whether should we expose our internal objective functions to high-level API.  From a Python user's perspective, the functionality would look like what's in sklearn except that it will be a class instead of a function:

``` python
from xgboost.objective import PseudoHuber
import xgboost

xgboost.XGBRegressor(objective=PseudoHuber(delta=10.0), tree_method="hist")
```

when the parameter is not needed (`delta=10.0` in above), one can continue to use the old string representation.

# Motivation
As we are having more and more objectives and metrics, the current approach of using only string names is becoming cumbersome since each of these objectives might have its own hyper-parameter, examples are the [fair xgboost](https://github.com/dmlc/xgboost/pull/7640) we have been working on, also the pseudo-Huber that we need to improve by introducing the slope.  Other examples are https://github.com/dmlc/xgboost/issues/5859 which can have multiple distributions to select from.

Aside from exposing the internal objects, we are also helping others who want to define their own custom objectives.  Currently, we only accept callable as custom objective, which is not sufficient for many cases since the objective also decides the link function and the number of output targets.  If we have a class interface we can help our users to simplify their custom objective functions.

# Other things we might need to explore

We might add more features to the objective in the future, for instance, each objective might have the ability to modify tree leaves after each iteration.  Features like this can help mitigate the effect of zero hessian objectives like MAE.  Can we expose this to custom objectives too?  Also, how do we make sure this restructure plan nicely with 3-party tools like hyper-parameter optimization libraries? Lastly, due to the current objective being an internal structure, even if we expose them to users we might not want to state its stability for a reasonabe period of time.

Kodiologist:
That makes sense to me. Objectives can require a fair amount of customization.

hcho3:
It looks like a natural way to configure parameters for metrics.

trivialfis:
Hi @ExpandingMan , would be great if you can provide some inputs for this proposal. ;-)

ExpandingMan:
My main concern in regard to the Julia wrapper is what this will look like in the C API.  Adding new data structures has the potential to add a lot of complexity for wrappers.

A lot of configuration arguments in the C API currently use JSONs.  Something similar should be pretty straightforward for wrappers to accommodate, but we should also consider how the wrappers are ultimately going to expose this to users.

[...]

PR #12059, opened 2026-03-03
Expose objective functions to the Python interface.

trivialfis:
Ref https://github.com/dmlc/xgboost/issues/7693 https://github.com/dmlc/xgboost/issues/9043

- Expose all objective functions to the Python interface.
- No change in the `Learner` yet. My plan is to make an independent C API to create objectives with the `split_grad` customization. This will have to wait until the plan is clear.
- Metrics are not exposed. I'm focusing on getting vector-leaf to work with built-in objectives.
- Remove the `TreeObjective`, since the linear model is now deprecated.
- `cv` function is not supported yet.
- Naming of the parameters are described in the referenced issue 

I named the classes `RegSquaredError` instead of `SquaredError` to maintain consistency with the C++ naming. But I'm open to suggestions.

### DIFF
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -87,7 +87,7 @@ from .compat import (
     is_pyarrow_available,
     py_str,
 )
-from .objective import Objective, _grad_arrinf, _stringify
+from .objective import Objective, _BuiltInObjective, _grad_arrinf, _stringify
 
 if TYPE_CHECKING:
     from pandas import DataFrame as PdDataFrame
@@ -2196,7 +2196,7 @@ class Booster:
             raise TypeError(f"Invalid training matrix: {type(dtrain).__name__}")
         self._assign_dmatrix_features(dtrain)
 
-        if fobj is None:
+        if fobj is None or isinstance(fobj, _BuiltInObjective):
             _check_call(
                 _LIB.XGBoosterUpdateOneIter(
                     self.handle, ctypes.c_int(iteration), dtrain.handle
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -87,7 +87,7 @@ from .compat import (
     is_pyarrow_available,
     py_str,
 )
-from .objective import Objective, TreeObjective, _grad_arrinf
+from .objective import Objective, _grad_arrinf
 
 if TYPE_CHECKING:
     from pandas import DataFrame as PdDataFrame
@@ -2280,21 +2280,14 @@ class Booster:
         vgrad: Optional[ArrayLike]
         vhess: Optional[ArrayLike]
 
-        if isinstance(fobj, TreeObjective):
-            # full gradient for leaf values
+        if isinstance(fobj, Objective):
             vgrad, vhess = fobj(iteration, y_pred, dtrain)
-            # Reduced gradient for split nodes
             split_grad = fobj.split_grad(iteration, vgrad, vhess)
-            # Switch the role of gradient if there's no split gradient but the tree
-            # objective is used.
             if split_grad is not None:
                 sgrad, shess = split_grad
             else:
                 sgrad, shess = vgrad, vhess
                 vgrad, vhess = None, None
-        elif isinstance(fobj, Objective):
-            sgrad, shess = fobj(iteration, y_pred, dtrain)
-            vgrad, vhess = None, None
         else:
             # Plain callable
             sgrad, shess = fobj(y_pred, dtrain)
--- a/python-package/xgboost/core.py
+++ b/python-package/xgboost/core.py
@@ -87,7 +87,7 @@ from .compat import (
     is_pyarrow_available,
     py_str,
 )
-from .objective import Objective, _grad_arrinf
+from .objective import Objective, _grad_arrinf, _stringify
 
 if TYPE_CHECKING:
     from pandas import DataFrame as PdDataFrame
@@ -2162,13 +2162,11 @@ class Booster:
         elif isinstance(params, str) and value is not None:
             params = [(params, value)]
         for key, val in cast(Iterable[Tuple[str, str]], params):
-            if isinstance(val, np.ndarray):
-                val = val.tolist()
-            elif hasattr(val, "__cuda_array_interface__") and hasattr(val, "tolist"):
-                val = val.tolist()
             if val is not None:
                 _check_call(
-                    _LIB.XGBoosterSetParam(self.handle, c_str(key), c_str(str(val)))
+                    _LIB.XGBoosterSetParam(
+                        self.handle, c_str(key), c_str(_stringify(val))
+                    )
                 )
 
     def update(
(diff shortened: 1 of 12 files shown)


RAMitchell (commented):
Exposing objectives is great - but the split gradient method doesn't make sense to me. It implies that you would want a different method of approximating gradients for each objective type, but it seems the same techniques apply (SVD, random projection).

trivialfis:
The alternative is an independent reducer as sketch boost, which I did took during initial implementation.

My final choice of using the objective as the vehicle to deliver the feature as I saw the technique as using different loss functions for split and leaf values, instead of using the same loss with dimension reduction applied on the gradient. This view was briefly mentioned in the document. The split loss may or may not be derived from dimension reduction techniques.

I'm open to change if the view is considered not productive. The interface is marked as wip, we can change.

Then, it's not difficult to compose either, just define a reducer with the `split_grad` as a mixin for your objective class.

RAMitchell:
> My final choice of using the objective as the vehicle to deliver the feature as I saw the technique as using different loss functions for split and leaf values, instead of using the same loss with dimension reduction applied on the gradient. This view was briefly mentioned in the document. The split loss may or may not be derived from dimension reduction techniques.

Do you have an example or reference for this?

If you want I can help implement a simple random projection then we have just one parameter.

trivialfis:
> Do you have an example or reference for this?

It's a perspective. You can **see** the gradient derived from dimension reduction as a different loss gradient: $\partial L_{split}$, $\partial L_{leaf}$, there's no need to put a strict link between them that says dimension reduction must be employed to obtain one from the other. To come up with a simple example for this [demo](https://github.com/dmlc/xgboost/blob/a1e4e74ba8d0d6c18dad58e7745177777a1538ce/demo/guide-python/multioutput_reduced_gradient.py#L28):
``` python
class LsObjSum(LsObjMean):
    def __init__(self, device: str) -> None:
        super().__init__(device=device)

    def __call__(
        self, iteration: int, y_pred: np.ndarray, dtrain: xgb.DMatrix
    ) -> Tuple[np.ndarray, np.ndarray]:
        self.dtrain = dtrain
        self.y_pred = y_pred
        return super().__call__(iteration, y_pred, dtrain)
    def split_grad(
        self, iteration: int, grad: np.ndarray, hess: np.ndarray
    ) -> Tuple[np.ndarray, np.ndarray]:
       # Derive the gradient from the mean of the output values instead of using the mean of the gradient
        y = self.dtrain.get_label()
        y_pred = self.y_pred
        y_mean = np.mean(y, axis=1)
        p_mean = np.mean(y_pred, axis=1)
        grad = p_mean - y_mean
        hess = np.ones(grad.shape)
        return grad, hess
```

> If you want I can help implement a simple random projection then we have just one parameter.

I'm not sure I want to make random projection the only choice, even if I drop the `split_grad` view. That seems like an unnecessary restriction, out of so many reduction techniques out there. I don't think this simplifies or enhances XGBoost.

trivialfis:
I reverted the changes for the vector-leaf. This PR now only handles exposing the objective functions.

RAMitchell:
This split gradient above is always an underestimate of split gain due to this inequality
image

Random projection is unbiased and will converge in expection, always, for every loss. Doesn't it just solve the problem?

trivialfis:
> This split gradient above is always an underestimate of split gain due to this inequality

It's just an example that one can see the gradient this way.

> Random projection is unbiased and will converge in expection, always, for every loss. Doesn't it just solve the problem?

It's certainly one of the candidates. But my point is, why put in this restriction?

trivialfis:
@RAMitchell I have reverted all changes related to the `split_grad`, this PR now only handles exposing the objective. Please help take another look when you are available. Some notes have been added to the PR description.

python-package/xgboost/testing/objective.py
```diff
@@ -0,0 +1,280 @@
...
+from .data import get_cancer
+from .utils import Device
+
+
+def check_train_regression_objectives(device: Device) -> None:
+    """Test training with regression objective classes."""
+    X, y, _ = make_regression(100, 5, use_cupy=device == "cuda")
+    dm = DMatrix(X, label=y)
+
+    for obj_inst, obj_name in [
```
RAMitchell, line 48: I guess we can't use pytest parameterize with this organisation of tests?

In legate-boost I had parameterisation over device and objectives.

python-package/xgboost/testing/objective.py
```diff
@@ -0,0 +1,280 @@
...
+        (RankPairwise(), "rank:pairwise"),
+        (RankMAP(), "rank:map"),
+    ]:
+        bst = train({"device": device}, dm, num_boost_round=5, obj=obj_inst)
+        cfg = json.loads(bst.save_config())
+        assert cfg["learner"]["objective"]["name"] == obj_name
+        assert obj_inst.name == obj_name
+
+
+def check_equivalence(device: Device) -> None:
```
RAMitchell: Would be nice to see this parameterised over objectives.

python-package/xgboost/core.py
```diff
@@ -2280,21 +2278,14 @@ def train_one_iter(grad: NumpyOrCupy, hess: NumpyOrCupy) -> None:
         vgrad: Optional[ArrayLike]
```
RAMitchell, line 2278: Just curious what the latency differences between using the python interface or internal. One possible simplification could be to always use the python, there is just one code path.

RAMitchell (approved):
LGTM. I assume documentation comes at a later stage? We will probably have to render all the math formulas in the docs.

python-package/xgboost/testing/objective.py
trivialfis: There are a couple of options, like parametrization, pytest subtests, or a simple loop. I don't have a strong preference for this specific test, as it's unlikely we need to isolate the tests to reproduce the issues. If one of them fails, all of the objectives should fail.

python-package/xgboost/core.py
trivialfis: The internal assumes split gradient must be available, and only does an extra leaf value computation at the end of iteration if there's an extra leaf value gradient. This is easy to implement as we only need one extra step.

But that's not intuitive to users since the algorithm creates an extra "split gradient", at least that's the mental model. So, the interface assumes a value gradient is available, as in normal gradient boosting. The assumption is switched here.

trivialfis:
> We will probably have to render all the math formulas in the docs.

I'm thinking if it's possible to reuse the document instead of copying and pasting it into Python files.

python-package/xgboost/testing/objective.py
trivialfis: Done.

python-package/xgboost/core.py
trivialfis: I misread the original question. There's no difference in latency, it's just setting parameter. Yes, I think a single code path would be super nice.

PR #12067, opened 2026-03-09
Remove min_val from quantiles

RAMitchell:
The minimum value in the quantiles is redundant. We can consider the lower bin boundary to be -inf implicitly. This refactor breaks the cached pages across versions but I don't think we need to support this.

### DIFF
--- a/doc/treemethod.rst
+++ b/doc/treemethod.rst
@@ -8,7 +8,7 @@ namely ``updater`` and ``tree_method``.  XGBoost has 3 builtin tree methods, nam
 free standing updaters including ``refresh``, ``prune`` and ``sync``.  The parameter
 ``updater`` is more primitive than ``tree_method`` as the latter is just a
 pre-configuration of the former.  The difference is mostly due to historical reasons that
-each updater requires some specific configurations and might has missing features.  As we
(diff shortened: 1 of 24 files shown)


trivialfis:
> This refactor breaks the cached pages across versions but I don't think we need to support this.

This is totally fine. The cache is temporary, with the memory pointer encoded in its filename. There's no need to check compatibility with magic numbers and version numbers.

src/common/hist_util.cc
```diff
@@ -17,43 +17,54 @@
...
-  auto const &mins = this->MinValues();
-  CHECK_LE(Span{mins}.size_bytes(), WriteVec(fo, mins));
   CHECK_GE(fo->Write(has_categorical_), sizeof(has_categorical_));
   CHECK_GE(fo->Write(max_cat_), sizeof(max_cat_));
 }
 
 [[nodiscard]] HistogramCuts *HistogramCuts::Load(common::AlignedResourceReadStream *fi) {
-  auto p_cuts = new HistogramCuts;
+  auto p_cuts = new HistogramCuts{0};
+  std::uint32_t magic{0};
```
trivialfis: Please remove this.

PR #12068, opened 2026-03-10
Push batch tree weights into predictors

RAMitchell:
This PR is a small step toward the follow-up work to remove GBDart as a separate booster and treat DART as weighted GBTree prediction/training behavior. It moves batch tree weighting into the predictor interface so GBTree::PredictBatchImpl can share one prediction path for both plain GBTree and DART-style weighted prediction. The scope here is intentionally limited to batch prediction; inplace prediction and SHAP are left unchanged for now. This is groundwork for the larger PR that removes GBDart.

### DIFF
--- a/src/gbm/gbtree.cc
+++ b/src/gbm/gbtree.cc
@@ -543,17 +543,19 @@ void GBTree::PredictBatchImpl(DMatrix* p_fmat, PredictionCacheEntry* out_preds, 
   auto cache_version = out_preds->version;
   // We can preserve the cache only when:
   // - prediction is unweighted
-  // - prediction starts from iteration 0, so a cached prefix is usable
+  // - prediction starts from iteration 0, so the result is a cacheable prefix
+  auto preserve_cache = tree_weights == nullptr && layer_begin == 0;
+  // We can reuse the existing cached prefix only when:
+  // - the result itself is cacheable
   // - the requested range does not move backwards past the cached version
-  auto preserve_cache = tree_weights == nullptr && layer_begin == 0 &&
-                        layer_end >= static_cast<bst_layer_t>(cache_version);
+  auto reuse_cache = preserve_cache && layer_end >= static_cast<bst_layer_t>(cache_version);
   // Initialize output when:
-  // - the cache cannot be reused, or
+  // - the cached prefix cannot be reused, or
   // - the cache is valid but still empty
-  auto initialize_output = !preserve_cache || cache_version == 0;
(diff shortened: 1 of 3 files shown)


src/gbm/gbtree.cc
```diff
@@ -737,63 +751,19 @@ class Dart : public GBTree {
...
   void PredictBatch(DMatrix* p_fmat, PredictionCacheEntry* p_out_preds, bool training,
                     bst_layer_t layer_begin, bst_layer_t layer_end) override {
     DropTrees(training);
-    this->PredictBatchImpl(p_fmat, p_out_preds, training, layer_begin, layer_end);
+    auto const* tree_weights = &weight_drop_;
+    std::vector<float> dropped_weights;
+    if (training && !idx_drop_.empty()) {
+      dropped_weights = weight_drop_;
+      for (auto idx : idx_drop_) {
+        dropped_weights.at(idx) = 0.0f;
```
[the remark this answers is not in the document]
RAMitchell: This is not worth it for dart

src/predictor/gpu_predictor.cu
```diff
@@ -661,14 +670,28 @@ class GPUPredictor : public xgboost::Predictor {
...
+    HostDeviceVector<float> weights;
+    auto pred_weights = common::OptionalWeights{1.0f};
+    if (tree_weights != nullptr) {
+      weights.SetDevice(ctx_->Device());
+      weights.HostVector().assign(tree_weights->cbegin() + tree_begin,
+                                  tree_weights->cbegin() + tree_end);
+      pred_weights = common::MakeOptionalWeights(ctx_->Device(), weights);
+    }
+    this->PredictDMatrix(dmat, out_preds, model, tree_begin, tree_end, pred_weights);
+    if (tree_weights != nullptr && tree_begin != tree_end) {
```
trivialfis, line 690: Is this a real concern? Is `PredictDMatrix` using a different CUDA stream than the one used here?

RAMitchell: Weights falls out of scope at the end of the function and can be deleted but the cpu often runs ahead of the gpu. Cuda free synchronizes but we might be using a pool that can reallocate the memory to something else. So there is a technical risk of the memory being reused before the kernel finishes. If all subsequent code was using the same stream it would be fine but we don't strictly have that guarantee.

RAMitchell: I forgot that we only ever use 1 stream. I might remove it.

PR #12069, opened 2026-03-10
[EM] Remove warning for ATS.

trivialfis:
We have a device cache now, the performance overhead is no longer solely decided by the link speed.


trivialfis:
cc @rongou

PR #12075, opened 2026-03-11
[coll] Prevent all-to-all connection.

trivialfis:
Use a static topology instead. The topology is a union of a ring and a tree.

Close https://github.com/dmlc/xgboost/pull/10840

### DIFF
--- a/src/collective/broadcast.cc
+++ b/src/collective/broadcast.cc
@@ -100,10 +100,10 @@ Result Broadcast(Comm const& comm, common::Span<std::int8_t> data, std::int32_t 
   }
 
   // For non-zero root, relay data up to rank 0 through the tree, then broadcast.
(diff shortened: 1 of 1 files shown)


trivialfis:
> What benefits are you expecting from this?

Reduced the number of connections. Before this PR, with N nodes, we had O(N^2) connections. Now it's linear. One can reproduce the connection error locally by over-subscribing the machine. For example, my desktop has 24 logical CPUs, and running 32 workers can cause trouble. In practice, this significantly limits the ability to scale.

Different implementations use different strategies to avoid all-to-all connections. It might be lazy connection, might be a routing table, might be something more sophisticated. The PR picks the easiest one.

> By the way the only major user of cpu all gather now seems to be cpu histogram column mode split sharing.

That's great news! Thank you for working on the quantile sketching!

PR #12079, opened 2026-03-12
Simplify GPU quantile merge path scan

RAMitchell:
This simplifies `MergePath` in the GPU quantile sketch merge.

`merge_by_key` now writes the `(x_step, y_step)` tuples consumed by `exclusive_scan_by_key` directly, instead of first materializing a source indicator and then transforming that indicator into scan input.

Behavior is unchanged. The goal here is to make the merge-path data flow more direct and remove some intermediate bookkeeping in the current implementation. This is a small cleanup only, not the larger future change to the reduction algebra.

**Testing**
- `conda run -n xgboost cmake --build build -j35 --target xgboost testxgboost`
- `./build/testxgboost --gtest_filter='Quantile.*:GPUQuantile.*:MGPUQuantileTest.*:IterativeDMatrix.GPUQuantile*:ExtMemQuantileDMatrix/ExtMemQuantileDMatrixGpu.*'`\n\n30 tests passed, 4 `MGPUQuantileTest` cases skipped due to missing federated/NCCL support in this build.

### DIFF
--- a/src/common/quantile.cu
+++ b/src/common/quantile.cu
@@ -113,13 +113,22 @@ void CopyTo(Span<T> out, Span<U> src) {
 XGBOOST_DEVICE thrust::tuple<uint64_t, uint64_t> MergePartition(Span<SketchEntry const> x,
                                                                 Span<SketchEntry const> y,
                                                                 uint64_t k) {
+  // Find the merge partition for the k-th output within one column.  The merged prefix of
+  // length k contains i entries from x and j entries from y, where k = i + j.
   auto m = static_cast<uint64_t>(x.size());
   auto n = static_cast<uint64_t>(y.size());
+  // Search for i inside the valid merge-partition range.  low/high clamp the partition so
+  // j = k - i always stays within [0, n].
   auto low = k > n ? k - n : 0ul;
   auto high = std::min(k, m);
   auto candidate_it = thrust::make_counting_iterator<uint64_t>(low);
   auto need_more_x = dh::MakeTransformIterator<bool>(candidate_it, [=] __device__(uint64_t i) {
+    // j is the number of elements taken from y when the partition takes i from x.
     auto j = k - i;
+    // Move the boundary right while the last candidate from y still sorts ahead of the
+    // next candidate from x.  The first false value is the first valid merge boundary.
+    // j > 0: there is a left-hand candidate in y.
+    // i < m: there is a right-hand candidate in x.
     return j > 0 && i < m && y[j - 1].value >= x[i].value;
   });
   auto partition_it = thrust::lower_bound(thrust::seq, need_more_x, need_more_x + (high - low + 1),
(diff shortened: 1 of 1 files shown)


RAMitchell:
Copilot found an issue in the previous implementation where reinterpret_cast from a tuple to sketch element is incorrect because they are not aligned the same. I may have to add a temporary buffer for correctness.

RAMitchell:
I rewrote it to remove the issue.

Also I had codex look for similar bugs in xgboost:

Findings

I found two places where reinterpret_cast relies on size but does not prove alignment.

First, in_memory_handler reinterprets std::string storage as typed arrays during allreduce. &buffer->front() is passed into Accumulate, and then cast to int32_t*, int64_t*, float*, double*, etc., without any alignment check:

https://github.com/dmlc/xgboost/blob/master/src/collective/in_memory_handler.cc#L80
https://github.com/dmlc/xgboost/blob/master/src/collective/in_memory_handler.cc#L143-L163
That looks like real UB. std::string gives contiguous byte storage, but it does not guarantee alignment suitable for arbitrary typed loads/stores. This code also divides by sizeof(T) but does not assert bytes % sizeof(T) == 0.

Second, the distributed GPU sketch merge path reconstructs Span<SketchEntry> from an int8_t allgather buffer without checking either segment-size divisibility or pointer alignment:

https://github.com/dmlc/xgboost/blob/master/src/common/quantile.cu#L487-L489
This code uses length_as_bytes / sizeof(SketchEntry) but does not check:

length_as_bytes % sizeof(SketchEntry) == 0
reinterpret_cast<uintptr_t>(raw.data()) % alignof(SketchEntry) == 0
There is already a safer pattern for the equivalent CPU-side deserialization path, which does perform an explicit alignment check before the cast:

https://github.com/dmlc/xgboost/blob/master/src/common/quantile.cc#L153-L155

trivialfis:
CUDA memory allocator has implicit alignment guarantee, unaligned access will result into access error. You can be explicit about it, which is good, but it's not necessarily a critical bug to assume the default alignment.

src/common/quantile.cu
```diff
@@ -112,151 +110,87 @@ void CopyTo(Span<T> out, Span<U> src) {
...
-  return merge_path;
+XGBOOST_DEVICE thrust::tuple<uint64_t, uint64_t> MergePartition(Span<SketchEntry const> x,
+                                                                Span<SketchEntry const> y,
+                                                                uint64_t k) {
+  auto m = static_cast<uint64_t>(x.size());
+  auto n = static_cast<uint64_t>(y.size());
+  auto low = k > n ? k - n : 0ul;
+  auto high = std::min(k, m);
+  auto candidate_it = thrust::make_counting_iterator<uint64_t>(low);
+  auto need_more_x = dh::MakeTransformIterator<bool>(candidate_it, [=] __device__(uint64_t i) {
```
trivialfis, line 125: It would be great if you could add some comments to these conditions `j > 0 && i < m`, along with some (now removed) old comments. It takes effort to wrap my head around what's happening here.

RAMitchell: I've added a bit more here

PR #12081, opened 2026-03-13
Remove separate Dart booster implementation

RAMitchell:
This PR is the next step toward removing DART as a separate booster. It aliases legacy `dart` booster selection to `GBTree`, teaches `GBTree` to load legacy DART model/config JSON directly, and removes the separate `Dart` class and registry entry. Saved models/configs are now canonicalized to `gbtree`, while old DART artifacts continue to load through the compatibility path.


trivialfis:
It would be great if you could log a warning that says one should use the `gbtree` directly with `dropout` instead of a separate `dart`.

PR #12043, opened 2026-02-24
Replace the global random engine.

trivialfis:
- Define the RNG in the `Context` class.
- Replace the global thread-local random engine with the new RNG in the `Context`.

Related:
- https://github.com/dmlc/xgboost/pull/10354
- https://github.com/dmlc/xgboost/pull/12037
- https://github.com/dmlc/xgboost/issues/10324
- https://github.com/dmlc/xgboost/issues/6711

PR #10354, opened 2024-05-31
[WIP] Remove global random engine.

trivialfis:
Move the random engine into the context.

todos:
- [ ] Test serialization with sampling.
- [ ] Test with related issues.

Related:
- https://github.com/dmlc/xgboost/issues/10324
- https://github.com/dmlc/xgboost/issues/6711

Issue #10324, opened 2024-05-24
Hist training with checkpointing is non-deterministic based on subsample


andrew-esteban-imc:
Hi there,

We have found that despite setting a seed for our `hist` training, we get non-deterministic results when resuming training from a checkpoint. A reproducer can be seen below:

```python
import numpy as np
import xgboost as xgb
from sklearn.datasets import make_classification
from sklearn.model_selection import train_test_split
from xgboost.callback import TrainingCallback

# Generate random sample data
X, y = make_classification(n_samples=1000, n_features=20, n_informative=10, n_redundant=10, random_state=42)

# Split the data into training and testing sets
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, random_state=42)


# Ensure model has correct epoch upon restart
class XgbCheckpointCallback(TrainingCallback):
    def __init__(
            self,
            start_epoch: int
    ):
        super().__init__()
        self.start_epoch = start_epoch

    def before_training(self, model: xgb.Booster):
        self._prev_update = model.update

        def update(
                dtrain, iteration: int, fobj=None
        ) -> None:
            return self._prev_update(dtrain, iteration + self.start_epoch, fobj)

        model.update = update
        return model

    def after_training(self, model):
        model.update = self._prev_update
        return model


# Set the parameters for XGBoost training
params = {
    'silent': True,
    "tree_method": "hist",
    "seed": 1,
    "base_score": -0.9,
    "max_depth": 3,
    "learning_rate": 0.02,
    "lambda": 500,
    "subsample": 0.9,
}

# Train the XGBoost model using the hist algorithm
dtrain = xgb.DMatrix(X_train, label=y_train)
dtest = xgb.DMatrix(X_test, label=y_test)

# Train the model for the first run
bst1 = xgb.train(params, dtrain, 10, evals=[(dtrain, 'train'), (dtest, 'test')], early_stopping_rounds=5, feval=None)

# Save the model after the first run
bst1.save_model('first_run_model')

# Train the model for the second run
bst2 = xgb.train(params, dtrain, 5, evals=[(dtrain, 'train'), (dtest, 'test')], early_stopping_rounds=5, feval=None)

# Save the model after the second run
bst2.save_model('second_run_model')

# Resume training from the second run and run for 5 more epochs
bst3 = xgb.train(params, dtrain, 5, evals=[(dtrain, 'train'), (dtest, 'test')], early_stopping_rounds=5, feval=None,
[...]

PR #12083, opened 2026-03-13
Serialize the RNGs

trivialfis:
This PR serializes the random states used by the context and by the column sampler. The column sampler stores a state, as we need to broadcast the seed for distributed training. Please note that the configuration is not stable across releases. The `Learner` class discards the configuration automatically if the versions don't match.

*todos*
- [x] Check accuracy is not changed.
- [x] Check cross platform compatibility.
- [x] Unify the code paths for rng serialization.

Related:
- https://github.com/dmlc/xgboost/pull/12037
- https://github.com/dmlc/xgboost/issues/10324
- https://github.com/dmlc/xgboost/pull/12043

Close:
- https://github.com/dmlc/xgboost/issues/6711
- https://github.com/dmlc/xgboost/pull/10354

### DIFF
--- a/plugin/sycl/tree/hist_row_adder.h
+++ b/plugin/sycl/tree/hist_row_adder.h
@@ -1,30 +1,32 @@
-/*!
- * Copyright 2017-2024 by Contributors
- * \file hist_row_adder.h
+/**
+ * Copyright 2017-2026, XGBoost Contributors
  */
 #ifndef PLUGIN_SYCL_TREE_HIST_ROW_ADDER_H_
 #define PLUGIN_SYCL_TREE_HIST_ROW_ADDER_H_
 
-#include <vector>
 #include <algorithm>
+#include <vector>
 
 namespace xgboost {
 namespace sycl {
 namespace tree {
 
 template <typename GradientSumT>
+class HistUpdater;
+
+template <typename GradientSumT>
 class HistRowsAdder {
  public:
-  virtual void AddHistRows(HistUpdater<GradientSumT>* builder,
-                           std::vector<int>* sync_ids, RegTree *p_tree) = 0;
+  virtual void AddHistRows(HistUpdater<GradientSumT>* builder, std::vector<int>* sync_ids,
+                           RegTree* p_tree) = 0;
   virtual ~HistRowsAdder() = default;
 };
 
 template <typename GradientSumT>
-class BatchHistRowsAdder: public HistRowsAdder<GradientSumT> {
+class BatchHistRowsAdder : public HistRowsAdder<GradientSumT> {
  public:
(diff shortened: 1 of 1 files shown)


python-package/xgboost/testing/continuation.py
```diff
@@ -57,3 +58,97 @@ def recursive_compare(obj_0: T, obj_1: T) -> None:
...
+        "objective": objective,
+        "subsample": subsample,
+        "sampling_method": sampling_method,
+        "colsample_bylevel": colsample_bylevel,
+        "seed_per_iteration": seed_per_iteration,
+        "booster": booster,
+    }
+    if num_class > 1:
+        params["num_class"] = num_class
+    if booster == "dart":
```
RAMitchell: Dart booster is gone.

src/common/random.h
```diff
@@ -180,12 +185,26 @@ class ColumnSampler {
...
   collective::SafeColl(rc);
   auto cs = std::make_shared<common::ColumnSampler>(seed);
   return cs;
 }
+
+void SaveRng(Json* p_out, RandomEngine const& rng);
+void LoadRng(Json const& in, RandomEngine* rng);
+
+// The utility function handles loading configuration before the model is fit. We don't
+// want to initialize the column sampler in a constructor, to avoid creating side effect in
```
RAMitchell: Why is it bad to create a side effect in the constructor but not otherwise?

trivialfis: Think about loading a model or extracting the parameter names for validation can alter the random engine. It's quite difficult to keep the training continuation consistent if anything other than the actual training step can alter the RNG.

python-package/xgboost/testing/continuation.py
trivialfis: I will update

trivialfis:
> Would you consider changing the design of column sampler to require the context rng every time it takes a sample instead of maintaining its own stream? 

Let me give it a go.

src/common/random.h
RAMitchell: OK as a general rule I wanted to go towards using constructors instead of the delayed initialization type pattern that's often used. I think we just need to have good tests like you have done here and be mindful of alterations to global state then constructors - I think it's always possible to create a shared pointer automatically lazy initialization situation if needed.

python-package/xgboost/testing/continuation.py
trivialfis: Done.

trivialfis:
@RAMitchell Removed the `rng` and `ctx` member variables from the column sampler and moved the allocation to the constructor. The column sampler `Init` call is still used lazily (unchanged) as it requires sampling parameters.

Issue #11473, opened 2025-05-21
Mixing booster types silently fails


blitchj:
In the Python interface, either using `train` or the sklearn interface `fit` method. 

I first fit a model with the default `gbtree`. I then fit another model with `booster=gblinear`, and `xgb_model `set to the first booster that used `gbtree`. The expected result would be continued boosting but with the new booster. However, it seems to silently ignore the `xgb_model `parameter.

While I would hope it could continue booting with different boosters, I think it should at least raise a warning if the boosters are mismatched and it is ignoring the parameter.

Reproducible code:
```
import json
from xgboost import XGBClassifier, Booster, DMatrix, train
from sklearn.datasets import make_classification

X, y = make_classification(n_samples=100_000, n_features=40, n_informative=33, n_clusters_per_class=30)

bst = train({"num_boost_round":100, "objective": "binary:logistic"}, DMatrix(X, label=y))
linbst = train({"num_boost_round":100, "objective": "binary:logistic", "booster": "gblinear"}, DMatrix(X, label=y), xgb_model=bst)

print(json.dumps(json.loads(linbst.save_raw(raw_format="json").decode("utf-8")),indent=2))
```
or
```
import json
from xgboost import XGBClassifier, Booster, DMatrix, train
from sklearn.datasets import make_classification

X, y = make_classification(n_samples=100_000, n_features=40, n_informative=33, n_clusters_per_class=30)

sk = XGBClassifier()
sk.fit(X, y)
lin = XGBClassifier(booster="gblinear")
lin.fit(X, y, xgb_model=sk.get_booster())

print(json.dumps(json.loads(lin.get_booster().save_raw(raw_format="json").decode("utf-8")),indent=2))
```
The outputs of these seems to only save the gblinear model, suggesting it is ignoring the xgb_model parameter.
However, if I do the same thing but stacking 2 gbtree models, then the saving shows the first half of the trees from the initial model and the second half of additional trees boosted on top.

trivialfis:
Thank you for raising the issue. The booster object itself doesn't support concatenating models. Users need to boost with the base margin. See https://xgboost.readthedocs.io/en/release_3.0.0/python/examples/boost_from_prediction.html#sphx-glr-python-examples-boost-from-prediction-py for an example.

blitchj:
[...]

PR #12088, opened 2026-03-15
[python-package] Warn when xgb_model booster type mismatches training config

PavelGuzenfeld:
## Summary

When passing an `xgb_model` trained with one booster type (e.g. `gbtree`) to
`xgb.train()` configured with a different booster type (e.g. `gblinear`), the
original model's trees are silently discarded. This can be confusing — users
expect training continuation but get a fresh model instead.

This PR adds a `UserWarning` when the booster types differ:

```
UserWarning: Booster type mismatch: `xgb_model` uses 'gbtree' but training
is configured with booster='gblinear'. The trees from `xgb_model` will not
be carried forward.
```

The check is in `training.py` so it covers both the core `xgb.train()` API
and the sklearn `fit()` wrapper.

## Test plan

- Added `test_booster_type_mismatch_warning` in `test_training_continuation.py`
- Verifies warning is raised on mismatch (gbtree -> gblinear)
- Verifies no warning when types match (gbtree -> gbtree)

Closes https://github.com/dmlc/xgboost/issues/11473

### DIFF
--- a/python-package/xgboost/training.py
+++ b/python-package/xgboost/training.py
@@ -3,7 +3,9 @@
 """Training Library containing training routines."""
 
 import copy
+import json
 import os
+import warnings
 import weakref
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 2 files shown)


RAMitchell:
Thank you for the pr but booster type will be gone soon.

PR #12090, opened 2026-03-16
Make GPU sketch prune emit a correct summary

RAMitchell:
This is a focused cleanup in the GPU sketch `Prune(...)` path.

The main change is to separate two roles that were previously mixed together:
- pruning raw pushed entries
- pruning existing sketch summaries

For the summary-side `Prune(...)` path, this PR now:
- selects source indices for the pruned summary
- removes repeated selected indices within each column
- gathers the selected `SketchEntry` values into the output summary

This is closer to the CPU `SetPrune` structure, where prune selection is based on selected records instead of immediately materializing output and then relying on that materialization for all cleanup.

This PR does **not** remove the later value-based `Unique()` step. It turns out that is still needed for now, especially in distributed / merged-summary cases. The value of this PR is separating out the summary prune implementation so later cleanup steps can be changed more safely and incrementally.

This is also one part of a larger refactoring of the GPU sketching path. The broader direction is to separate:
- summary selection under the prune budget
- summary normalization / cleanup
- container and merge-path cleanup

**Testing**
- `make -C build-make -j35 testxgboost`
- `./build-make/testxgboost --gtest_filter='GPUQuantile.Prune:GPUQuantile.PruneDuplicated:GPUQuantile.MergeBasic:GPUQuantile.MultiMerge:Quantile.DistributedBasic:Quantile.Distributed'`

All 6 tests passed.

### DIFF
--- a/src/common/quantile.cu
+++ b/src/common/quantile.cu
@@ -63,8 +63,8 @@ template <typename EntryFromIndex>
 void SelectPruneIndices(common::Span<SketchContainer::OffsetT const> cuts_ptr,
                         Span<SketchContainer::OffsetT const> columns_ptr_in,
                         Span<FeatureType const> feature_types, Span<size_t> selected_idx,
-                        EntryFromIndex entry_from_index) {
-  dh::LaunchN(selected_idx.size(), [=] __device__(size_t idx) {
+                        EntryFromIndex entry_from_index, cudaStream_t stream) {
+  dh::LaunchN(selected_idx.size(), stream, [=] __device__(size_t idx) {
     size_t column_id = dh::SegmentId(cuts_ptr, idx);
     auto in_begin = columns_ptr_in[column_id];
     auto in_size = columns_ptr_in[column_id + 1] - columns_ptr_in[column_id];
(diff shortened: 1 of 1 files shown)


src/common/quantile.cu
```diff
@@ -52,12 +52,58 @@ __device__ SketchEntry BinarySearchQuery(EntryIter beg, EntryIter end, float ran
...
+        [=] __device__(size_t abs_idx) { return entry_from_index(abs_idx); });
+    selected_idx[cuts_ptr[column_id] + idx] =
+        in_begin + BinarySearchQueryIndex(it, it + in_size, q);
+  });
+}
+
+template <typename EntryFromIndex>
+void GatherPruneEntries(Span<size_t const> selected_idx, Span<SketchEntry> out_cuts,
+                        EntryFromIndex entry_from_index) {
+  dh::LaunchN(selected_idx.size(),
```
trivialfis: Please pass an explicit stream to the `LaunchN`. The context returns the default per-thread stream for now, but I plan to move torward self-managed CUDA stream in the future to aid device ordinal selection, among other benefits.

src/common/quantile.cu
```diff
@@ -412,14 +458,24 @@ void SketchContainer::Prune(Context const *ctx, std::size_t to) {
...
-  this->columns_ptr_.Copy(columns_ptr_b_);
+  dh::device_vector<size_t> selected_idx(out.size());
+  auto d_selected_idx = dh::ToSpan(selected_idx);
+  HostDeviceVector<OffsetT> selected_columns_ptr(columns_ptr_b_.Size());
+  selected_columns_ptr.SetDevice(ctx->Device());
+  auto entry_from_index = [=] __device__(size_t abs_idx) {
+    return in[abs_idx];
+  };  // NOLINT
+  SelectPruneIndices(d_columns_ptr_out, d_columns_ptr_in, ft, d_selected_idx, entry_from_index);
+  auto n_selected = dh::SegmentedUnique(
```
trivialfis, line 472: Not sure what the difference is after this PR? It's still running an unique op, and the existing one is still here.

RAMitchell: Turns out I can't remove unique yet I need a couple more steps. What does PR really does is separate out a prune implementation that kind of does 2 things, processing raw inputs and processing sketches, into two different implementations.

src/common/quantile.cu
RAMitchell: Done

PR #12097, opened 2026-03-17
[WIP] Pass objective args through constructors

RAMitchell:
This PR moves objective initialization toward constructor-time configuration. `ObjFunction::Create(...)` now accepts `Args`, the learner passes configuration when constructing fresh objectives, and stateful objectives initialize from constructor arguments while still supporting `Configure(...)` for existing loaded instances.

As a follow-up within the same branch, the tests were updated to prefer constructor-time objective initialization instead of `Create(...); Configure(...)` where deferred setup was not intentional. The remaining deferred-init cases are now mostly limited to LambdaRank and load/clone-style paths like intercept estimation, which makes the next cleanup step much clearer.

## Verification
- `cmake --build build-cpu-quantile -j35 --target testxgboost`
- `./build-cpu-quantile/testxgboost --gtest_filter='Objective.*:AFTObj.*:*Quantile*:*RegressionObj*:*MultiClass*:*Hinge*'`
- `./build-cpu-quantile/testxgboost --gtest_filter='Objective.CPU_vs_CUDA:SyclObjective.CPUvsSycl:Plugin.ExampleObjective:Predictor.ShapOutputCasesCPU:Predictor.ApproxContribsBasic:Predictor.ShapIterationRange'`

### DIFF
--- a/include/xgboost/objective.h
+++ b/include/xgboost/objective.h
@@ -137,6 +137,7 @@ class ObjFunction : public Configurable {
    * @param ctx  Pointer to the context.
    */
   static ObjFunction* Create(const std::string& name, Context const* ctx, Args const& args = {});
+  static ObjFunction* Create(Context const* ctx, Json const& config);
 };
 
 /*!
--- a/include/xgboost/objective.h
+++ b/include/xgboost/objective.h
@@ -135,7 +135,7 @@ class ObjFunction {
    * @param name Name of the objective.
    * @param ctx  Pointer to the context.
    */
-  static ObjFunction* Create(const std::string& name, Context const* ctx, Args const& args = {});
+  static ObjFunction* Create(const std::string& name, Context const* ctx, Args const& args);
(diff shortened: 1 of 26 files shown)


RAMitchell:
@trivialfis after playing around with objective configuration I think I'm starting to understand the learner configuration process and how we might tackle its fundamental difficulties.

I see the main difficulty in that parameter updates are stored in learner as "deltas" but we don't keep a snapshot of the canonical learner config. If we keep at all times a config as a snapshot of truth and reconcile updates against it I think things will simplify.

trivialfis:
I think that's a rabbit hole once you start handling things like eta decay. Basically, you need to preserve an entire copy of the parameter structure in the learner as a mutable state, with internal dependencies. Then you need to compare the states to check whether changes have occurred in some components. In the end, you might end up something like what we are currently doing: each component handles its own parameter.

RAMitchell:
I noticed that decay issue during this but I think it's still a significant simplification to go from deltas of configurations to whole configurations.

PR #12105, opened 2026-03-18
Prune GPU sketch after each merge

RAMitchell:
This changes the batched GPU sketching path to keep the intermediate sketch bounded after each local merge.

Changes in this PR:
- always materialize the current batch sketch into a temporary buffer in `Push(...)`
- route all batch integration through `Merge(...)`, including the first window
- add early empty-input fast paths to `Merge(...)`
- prune the resident sketch after each local merge using the intermediate cut budget

The main motivation is memory behavior during batched GPU sketching. Previously, each batch was pruned locally and then merged into the resident sketch, but the merged sketch was not re-pruned until the final cut generation stage. That allowed the resident sketch to grow across many batches.

With this change, the local sketch remains bounded between batches, which makes smaller sketch batches a more effective memory-control knob.

This also simplifies the control flow:
- `Merge(...)` owns the empty-input cases
- `Push(...)` no longer needs a separate first-window specialization

Testing:
- `make -C build-make -j35 testxgboost`
- `./build-make/testxgboost --gtest_filter='GPUQuantile.Basic:GPUQuantile.Prune:GPUQuantile.PruneDuplicated:GPUQuantile.MergeEmpty:GPUQuantile.MergeBasic:GPUQuantile.MergeDuplicated:GPUQuantile.MergeCategorical:GPUQuantile.MultiMerge:Quantile.DistributedBasic:Quantile.Distributed:MGPUQuantileTest.AllReduceBasic'`

Result:
- 10 passed
- 1 skipped (`MGPUQuantileTest.AllReduceBasic`, missing federated/NCCL support)

### DIFF
--- a/tests/python/test_data_iterator.py
+++ b/tests/python/test_data_iterator.py
@@ -40,13 +40,12 @@ def _assert_cut_rank_error_within_tolerance(
         end = int(indptr[fidx + 1])
         column_cuts = cuts[beg:end]
         assert np.all(np.diff(column_cuts) >= 0.0)
-        # For tiny weighted sketches, a purely relative tolerance or average bin-mass
-        # floor can be smaller than the rank mass of one observation.
-        acceptable_error = max(
-            total_weight * eps,
-            total_weight / float(column_cuts.shape[0]),
-            max_weight,
+        # For tiny weighted sketches, allow the coarse sketch tolerance plus the mass
+        # of one observation, since a cut can move across one weighted sample.
+        base_tolerance = max(
+            total_weight * eps, total_weight / float(column_cuts.shape[0])
         )
+        acceptable_error = base_tolerance + max_weight
 
         # Ignore the last cut, matching the C++ TestRank helper.
         sorted_idx = np.argsort(x[:, fidx], kind="stable")


RAMitchell:
One thing I've noticed is that sketch error is inconsistent. The cpu single machine sketch prunes to a size like O(1/eps * log(n)) but distributed paths and GPU single machine prune to1/eps. I will eventually make this consistent and make the tests enforce it rigorously.

src/common/hist_util.cu
```diff
@@ -58,16 +59,22 @@ size_t RequiredMemory(bst_idx_t num_rows, bst_feature_t num_columns, size_t nnz,
   // 4. Allocate cut pointer by increasing: n_columns + 1
   total += (num_columns + 1) * sizeof(SketchContainer::OffsetT);
   // 5. Allocate cuts: assuming rows is greater than bins: n_columns * limit_size
-  total += RequiredSampleCuts(num_rows, num_bins, num_bins, nnz) * sizeof(SketchEntry);
-  // 6. Deallocate copied entries by reducing: bytes_per_element * shape.
+  total += cuts_bytes;
```
RAMitchell, line 62: I will get rid of memory estimation shortly.

RAMitchell:
These test failures are happening because they expect equivalence of predictions from the a model built on a single batch versus a many batches. The quantiles after this PR are still very accurate and certainly within tolerance in terms of rank error but the bitwise result is no longer the same as a result I will rewrite the test to check appropriate convergence and rank error of quantiles instead of predictions.

PR #12108, opened 2026-03-19
Fix macOS wheel OpenMP dependency resolution

RAMitchell:
## Summary

This PR fixes the macOS x86_64 wheel CI failure.

The failing job was:

- `Build macOS wheel (macosx_x86_64)`

`delocate` failed while repairing the wheel because `libxgboost.dylib` picked up Homebrew LLVM's `libunwind.1.0.dylib`, and that dependency requires macOS 14.0 even though the wheel target is macOS 10.15.

## Root Cause

On the macOS Intel runner, the wheel build was resolving OpenMP through Homebrew. That introduced a dependency on Homebrew LLVM's `libunwind`, which caused wheel repair to fail with:

```text
delocate.libsana.DelocationError: Library dependencies do not satisfy target MacOS version 10.15:
.../libunwind.1.0.dylib has a minimum target of 14.0

### DIFF
--- a/ops/pipeline/build-python-wheels-macos.sh
+++ b/ops/pipeline/build-python-wheels-macos.sh
@@ -22,7 +22,6 @@ if [[ "$platform_id" == macosx_* ]]; then
         cpython_ver=310
         cibw_archs=x86_64
         export MACOSX_DEPLOYMENT_TARGET=10.15
-        export CIBW_CONFIG_SETTINGS='use_openmp=false'
     else
         echo "Platform not supported: $platform_id"
         exit 3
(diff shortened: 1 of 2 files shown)


trivialfis:
Note that the wheel is used by pypi.

https://github.com/dmlc/xgboost/pull/11983#issuecomment-3850216785

RAMitchell:
I really don't have much experience here so feel free to take over.

trivialfis:
> I really don't have much experience here so feel free to take over.

No one does. ;-( 

I borrowed a macbook and used another mac pro from aws at the time to see how it works ...

Basically, I gave up after wasting a lot of time trying to work around it. To handle it properly, we will either:
- Remove parallel for CPU entirely on macos.
- Use the existing thread pool in XGBoost to replace the OpenMP runtime, inject it into the `common::ParallelFor`.

trivialfis:
cc @hcho3

PR #12111, opened 2026-03-19
[python-package] Replace bare Any with specific types in sklearn helpers

PavelGuzenfeld:
## Summary

Replace generic `Any` annotations with more descriptive types in sklearn.py internal helper functions. Follow-up to #6496.

**Changes:**

| Function | Parameter | Before | After |
|----------|-----------|--------|-------|
| `_wrap_evaluation_matrices` | `X`, `y` | `Any` | `ArrayLike` |
| `_wrap_evaluation_matrices` | `group`, `qid`, `sample_weight`, `base_margin` | `Optional[Any]` | `Optional[ArrayLike]` |
| `_wrap_evaluation_matrices` | `eval_set` | `Optional[Sequence[Tuple[Any, Any]]]` | `Optional[Sequence[Tuple[ArrayLike, ArrayLike]]]` |
| `_wrap_evaluation_matrices` | `*_eval_set`, `eval_group`, `eval_qid` | `Optional[Sequence[Any]]` | `Optional[Sequence[ArrayLike]]` |
| `pick_ref_categories` | `X` | `Any` | `ArrayLike` |
| `invalid_type` | `m` | `Any` | `object` |

The return type of `_wrap_evaluation_matrices` stays `Tuple[Any, ...]` since it is called with both `DMatrix` and `DaskDMatrix` (which does not extend `DMatrix`).

Remaining `Any` usages (`**kwargs: Any`, `Dict[str, Any]` for params) are intentional — they represent genuinely heterogeneous data or standard sklearn patterns.

## Test plan

- [x] CI pre-commit passes
- [x] CI mypy type check passes

### DIFF
--- a/python-package/xgboost/sklearn.py
+++ b/python-package/xgboost/sklearn.py
@@ -649,7 +649,7 @@ def get_model_categories(
 
 
 def pick_ref_categories(
-    X: Any,
+    X: ArrayLike,
     model_cats: Optional[Union[FeatureTypes, Categories]],
     Xy_cats: Optional[Categories],
 ) -> Optional[Union[FeatureTypes, Categories]]:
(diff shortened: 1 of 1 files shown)


python-package/xgboost/sklearn.py
```diff
@@ -1211,7 +1209,7 @@ def custom_metric(m: Callable) -> Metric:
                 wrapped = _metric_decorator(m)
             return wrapped
 
-        def invalid_type(m: Any) -> None:
+        def invalid_type(m: object) -> None:
```
trivialfis, line 1212: why?

PavelGuzenfeld: `object` is the proper top type in Python — it means "accepts anything" while keeping type safety. `Any` disables type checking entirely (both co- and contra-variant), so mypy silently skips any misuse. Since `invalid_type` only calls `type(m)`, which works on every object, `object` is the accurate annotation.

That said, this is a minor stylistic point — happy to revert if you prefer keeping `Any` here.

PavelGuzenfeld:
@trivialfis does the explanation above work for you? Happy to revert the object→Any change if you'd prefer keeping Any here.

PR #12114, opened 2026-03-21
Add Automatic Module Names for Java packages

### LINKED ISSUE #11792: Add an Automatic Module Name for XGBoost
I am trying to use XGBoost from a Java project, but it's difficult to use along with `module-info.java`.   The reason is that there is no module name for XGBoost, and the automatically generated one doesn't work with the Jar name.   

Specifically, Java can't derive the name from `xgboost4j_2.13-3.1.1.jar`.  Normally it would be able to, but the `2.13`, the Scala version, makes it have this error:

```
Tue Nov 04 11:01:11 $ jar --describe-module --file xgboost4j_2.13-3.1.1.jar
Unable to derive module descriptor for: xgboost4j_2.13-3.1.1.jar
xgboost4j.2.13: Invalid module name: '2' is not a Java identifier
```

I believe the correct way to fix this is to add a MANIFEST.MF Entry like they did here: https://github.com/oshi/oshi/issues/1243

I am not familiar enough with Maven to know how to add it [but here is a try](https://github.com/carl-mastrangelo/xgboost/commit/cd902b5569b94e52d69d6126c7f43ea9936aa4ef).   This [blog post](http://branchandbound.net/blog/java/2017/12/automatic-module-name/) describes how as well.

carl-mastrangelo:
This allows using XGBoost in Modular (Java 9+) projects

Fixes #11792

### DIFF
--- a/jvm-packages/xgboost4j/pom.xml
+++ b/jvm-packages/xgboost4j/pom.xml
@@ -91,7 +91,7 @@
                           <goal>exec</goal>
                       </goals>
                       <configuration>
-                          <executable>python3</executable>
+                          <executable>python</executable>
                           <arguments>
                             <argument>create_jni.py</argument>
                             <argument>--log-capi-invocation</argument>


trivialfis:
cc @wbo4958

wbo4958:
Thx @carl-mastrangelo, Can you also add it support for xgboost4j-spark-gpu module?

carl-mastrangelo:
@wbo4958  Done.  Just a heads up, I manually tested the other three Jars, but since the Spark GPU is using shading, it's not exactly clear what the module would be, or if it should be shaded as well.  


Also, since I don't have an Nvidia card, I wasn't able to run the tests.

jvm-packages/xgboost4j-spark-gpu/pom.xml
```diff
@@ -52,6 +52,18 @@
...
                 </configuration>
             </plugin>
+            <plugin>
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-jar-plugin</artifactId>
+              <version>3.4.2</version>
+              <configuration>
+                <archive>
+                  <manifestEntries>
+                    <Automatic-Module-Name>ml.dmlc.xgboost4j.sparkgpu</Automatic-Module-Name>
```
wbo4958: how about "ml.dmlc.xgboost4j.spark" ?

carl-mastrangelo: I'm afraid that would conflict with the other spark module.   My understanding is they need to be unique.

wbo4958: how about ml.dmlc.xgboost4j.spark.gpu ?

carl-mastrangelo: Works for me.  PTAL.

carl-mastrangelo:
@wbo4958  Any other feedback?

wbo4958:
LGTM. Thx for your PR.

wbo4958 (approved):

jvm-packages/xgboost4j/pom.xml
```diff
@@ -91,7 +91,7 @@
                           <goal>exec</goal>
                       </goals>
                       <configuration>
-                          <executable>python</executable>
+                          <executable>python3</executable>
```
trivialfis: Let's keep this as `python` instead of `python3`. We can disucss the need to rename in a different issue or PR.

PR #12115, opened 2026-03-21
Add warning for booster type mismatch (fixes #11473)

### LINKED ISSUE #11473: Mixing booster types silently fails

harshad-dhokane:
## Description                                                                                                                                                                                       
  Fixes #11473                                                                                                                                                                                         
                                                                                                                                                                                                       
  ## Bug                                                                                                                                                                                               
  When using `xgb_model` to continue training with a different booster type than the loaded model, XGBoost silently ignores the mismatch.                                                              
                                                                                                                                                                                                       
  ## Fix                                                                                                                                                                                               
  Added validation in the `train()` function to detect booster type inconsistency and warn the user.                                                                                                   
                                                                                                                                                                                                       
  ## Changes                                                                                                                                                                                           
  - Modified `python-package/xgboost/training.py`                                                                                                                                                      
  - Added `json` and `warnings` imports                                                                                                                                                                
  - Added booster type validation after Booster creation                                                                                                                                               
                                                                                                                                                                                                       
  Fixes #11473

### DIFF
--- a/python-package/xgboost/training.py
+++ b/python-package/xgboost/training.py
@@ -3,7 +3,9 @@
 """Training Library containing training routines."""
 
 import copy
+import json
 import os
+import warnings
 import weakref
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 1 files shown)


chyunsu3:
Thanks, but the gblinear booster type will be removed in the near future.

PR #12118, opened 2026-03-22
Simplify GPU sketch batch sizing

RAMitchell:
This changes GPU sketch batching to use a fixed sliding-window size instead of the previous memory-estimation path.

Changes in this PR:
- remove the memory-estimation-based batch sizing logic
- use a fixed default batch size of `64M` elements, clamped to the input size
- remove the production and test plumbing for overriding the batch size
- remove batching helpers that only supported the old estimator path
- replace estimator-based memory tests with coarse peak-memory regression tests on representative GPU sketch workloads

The goal is to make the batching policy simpler and more stable:
- one fixed batch-size policy
- no dependence on allocator-reported available memory
- less dead code and less testing surface tied to an approximate memory formula

`64M` was chosen from local runtime/peak-memory sweeps as a conservative default that stays in the fast region across the tested shapes without the repeated-merge overhead of smaller batches.

Representative results:

For `65536 x 2048`:
- `16M`: `7.07 GiB`, `469 ms`
- `32M`: `2.38 GiB`, `440 ms`
- `48M`: `1.80 GiB`, `329 ms`
- `64M`: `1.47 GiB`, `287 ms`

For `16384 x 8192`:
- `16M`: `5.21 GiB`, `447 ms`
- `32M`: `3.40 GiB`, `453 ms`
- `48M`: `2.57 GiB`, `383 ms`
- `64M`: `1.98 GiB`, `328 ms`

For `65536 x 512`:
- `16M`: `384 MiB`, `611 ms`
- `32M`: `512 MiB`, `66 ms`
- `64M`: `512 MiB`, `68 ms`
- `80M`: `512 MiB`, `63 ms`

These sweeps show that `16M` is clearly too small, while `64M` consistently avoids the worst repeated-batching overhead and remains near the best runtime/memory region across different column counts.

## Testing
- rebuilt `testxgboost`
- ran focused coverage for GPU sketching, weighted sketching, external-memory sketching, adapter sketching, GPU quantile prune/merge, and all `DeviceSketchWithHessian` cases

Result:
- `61 / 61` passed

### DIFF
--- a/.github/workflows/r_tests.yml
+++ b/.github/workflows/r_tests.yml
@@ -81,6 +81,8 @@ jobs:
         run: |
           # Must run before checkout to have the latest git installed.
           apt update && apt install libcurl4-openssl-dev libssl-dev libssh2-1-dev libgit2-dev libglpk-dev libxml2-dev libharfbuzz-dev libfribidi-dev git librsvg2-dev librsvg2-2 pandoc -y
+          # cmake is required by the R `fs` package to build its bundled libuv.
+          apt install cmake -y
       - name: Trust git cloning project sources
         run: |
           git config --global --add safe.directory "${GITHUB_WORKSPACE}"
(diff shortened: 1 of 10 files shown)


trivialfis:
I'm looking into the R failure.

PR #12128, opened 2026-03-25
Use GPU AllreduceV for distributed quantile sketch exchange

RAMitchell:
## Summary

This changes the distributed GPU quantile sketch exchange from an allgather-style pattern to a GPU `AllreduceV` path built on the existing collective interfaces.

The new implementation packs the sketch into a raw byte payload, performs a variable-size reduction over NCCL, and materializes the reduced sketch back into `SketchContainer`. The reduction uses a tree of point-to-point edges for the reduce phase and the existing broadcast path for the final fan-out.

## Motivation

The previous allgather-style exchange causes communicated bytes and temporary communication storage to grow poorly with worker count because every rank materializes peer payloads.

A tree-shaped reduction keeps the communication pattern closer to the CPU `AllreduceV` path and avoids that replication on the NCCL fast path.

## Design

The main pieces are:
- a GPU `AllreduceV` API aligned with the existing collective API shape in `src/collective/allreduce.h`
- a private CUDA implementation in `src/collective/allreduce_v.cuh`
- quantile-side packing/unpacking in `src/common/quantile.cu`

The current NCCL implementation:
- uses existing `Comm::Chan(peer)` wrappers for tree `Send`/`Recv` edges
- uses the existing broadcast wrapper for the final dissemination
- preserves upstream nonblocking NCCL communicator init
- explicitly waits for communicator progress after p2p edges so the tree is stable with nonblocking NCCL communicators

## Main Changes

- Add GPU `AllreduceV` support to the collective layer.
- Route distributed GPU quantile sketch exchange through that helper instead of the old allgather-style path.
- Keep the quantile reduction logic local to `quantile.cu` by packing the sketch as bytes and reducing with merge + prune semantics.
- Add focused tests for the GPU `AllreduceV` tree shape and repeated-call stability.

## Testing

Ran:
- `MGPUAllreduceVTest.SimpleTree`
- `MGPUAllreduceVTest.RepeatedTree`
- `MGPUAllreduceTest.Timeout`
- `GPUQuantile.Prune`
- `GPUQuantile.MergeBasic`
- `GPUQuantile.MultiMerge`
- `MGPUQuantileTest.AllReduceBasic`

## Follow-up

- Replace the temporary relaxed quantile cut-value tolerance with a check based on sketch/rank guarantees.
- Decide how `AllreduceV` should behave on non-NCCL distributed GPU backends such as federated CUDA communication.

### DIFF
--- a/src/common/quantile.cu
+++ b/src/common/quantile.cu
@@ -374,8 +374,8 @@ void SketchContainer::Push(Context const *ctx, Span<Entry const> entries, Span<s
     return;
   }
   this->Merge(ctx, cuts_ptr, out.subspan(0, n_uniques));
-  auto intermediate_budget = SketchSummaryBudget(num_bins_, rows_seen_);
-  this->Prune(ctx, intermediate_budget);
+  auto intermediate_num_cuts = static_cast<bst_idx_t>(this->IntermediateNumCuts());
+  this->Prune(ctx, intermediate_num_cuts);
 }
 
 size_t SketchContainer::ScanInput(Context const *ctx, Span<SketchEntry> entries,
--- a/src/common/quantile.cuh
+++ b/src/common/quantile.cuh
@@ -66,6 +66,12 @@ class SketchContainer {
     columns_ptr_.Copy(columns_ptr_tmp_);
     entries_.resize(n_entries);
   }
+  [[nodiscard]] std::size_t IntermediateNumCuts() const {
+    auto const base = static_cast<std::size_t>(num_bins_) * kFactor;
+    auto const eps = 1.0 / static_cast<double>(base);
+    auto const per_feature = WQSketch::LimitSizeLevel(std::max<std::size_t>(1, rows_seen_), eps);
+    return per_feature * num_columns_;
+  }
 
   // Get the span of one column.
   Span<SketchEntry> Column(bst_feature_t i) {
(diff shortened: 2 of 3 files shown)


src/collective/allreduce_v.cuh
```diff
@@ -0,0 +1,206 @@
...
+    size.resize(1);
+  }
+};
+
+template <typename T, typename Fn>
+std::enable_if_t<
+    std::is_invocable_v<Fn, dh::device_vector<T> const&, dh::device_vector<T> const&,
+                        dh::device_vector<T>*, cudaStream_t>,
+    Result>
+AllreduceV(NCCLComm const& nccl, dh::device_vector<T>* data, AllreduceVScratch<T>* scratch,
```
trivialfis, line 39: The `NCCLComm` has a set of channels stored in the `Comm` base class, with the usual stream block methods. Do you think we can use them?

I would like to reuse some existing wrappers, like channel and broadcast, instead of going directly to the NCCL handle, if possible.

RAMitchell: I've had a try and it seems to work.

RAMitchell:
I had to include a fallback for federated learning.

trivialfis:
Let me take a look next week.

trivialfis:
> I had to include a fallback for federated learning.

Unfortunately, the federated learning module cannot be made to fit the normal ring/tree structure, as peers are forbidden to see the messages from other peers.

PR #12137, opened 2026-04-01
Use context for CUDA external memory DMatrix.

trivialfis:
The change by https://github.com/dmlc/xgboost/pull/12121 is too significant, starting with a smaller place.

- Use a sync policy to let the caller handle stream sync. This avoids flaky lifetime management in the dtor.


RAMitchell (approved):
I find it a bit weird passing a raw pointer all the time without any guardrails for its state, but I guess we strongly guarantee that there is only ever one context and it always exists.

trivialfis:
> I find it a bit weird passing a raw pointer all the time without any guardrails for its state

That's indeed a significant issue. I would like to use a shared ptr instead at some point.

PR #12141, opened 2026-04-04
Migrate RMM usage to CCCL MR design

bdice:
## Summary

Part of https://github.com/rapidsai/rmm/issues/2011.
Depends on https://github.com/rapidsai/rmm/pull/2361.

This updates `tests/cpp/helpers.cc` for API changes in RMM to the new CCCL memory resource design. The `pool_memory_resource` is no longer templated, `set_per_device_resource` is replaced by `set_per_device_resource_ref`, and the pool now holds its upstream via owning `any_resource` so a separate `cuda_memory_resource` vector is no longer needed. Old RMM code paths are preserved behind version guards for backwards compatibility.


jakirkham:
cc @hcho3 @trivialfis

chyunsu3 (approved):
LGTM. The `#ifdef`'s looks reasonable: it's restricted to a single source file (`helpers.cc`)

trivialfis (approved):

jakirkham:
Thanks Bradley and Hyunsu! 🙏

Given this is just in the tests, do we need a patch for building packages or is that unneeded?

jakirkham:
Also thanks Jiaming! 🙏

Sorry for missing you before 😞

PR #12155, opened 2026-04-10
Expand CPU quantile property tests and remove redundant cases

RAMitchell:
## Summary

This PR continues the CPU quantile test rewrite by expanding the property-style coverage in
`test_quantile.cc` and removing redundant legacy CPU quantile tests.

The new CPU property suites now cover:
- summary-level sketch invariants
- container-level cut invariants
- `SketchOnDMatrix` wrapper invariants
- distributed row-split and column-split properties

## What Changed

- broaden the shared CPU quantile case matrix in `tests/cpp/common/test_quantile_helpers.h`
- add local cut validators used by the new property tests
- replace legacy distributed identity-style tests with property checks against reference data
- remove redundant CPU quantile tests from `test_quantile.cc` and `test_hist_util.cc`

## Testing

- `cmake --build build-cpu --target testxgboost -j35`
- focused CPU quantile slice

Result:
- `140/140` passed

### DIFF
--- a/tests/cpp/common/test_quantile.cc
+++ b/tests/cpp/common/test_quantile.cc
@@ -19,6 +19,7 @@ namespace quantile_test {
 class QuantileSummaryTest : public ::testing::TestWithParam<SummaryCase> {};
 class QuantileContainerTest : public ::testing::TestWithParam<ContainerCase> {};
 class QuantileDistributedContainerTest : public ::testing::TestWithParam<ContainerCase> {};
+class QuantileSketchOnDMatrixTest : public ::testing::TestWithParam<ContainerCase> {};
 
 namespace {
 void TestSummaryInvariants(SummaryCase const& c, WQSummaryContainer const& summary,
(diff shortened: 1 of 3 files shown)


tests/cpp/common/test_quantile.cc
```diff
@@ -324,162 +344,49 @@ void DoTestColSplitQuantile(size_t rows, size_t cols) {
...
-}
-
-TEST(Quantile, ColumnSplitSorted) {
-  constexpr size_t kRows = 4000, kCols = 200;
-  TestColSplitQuantile<true>(kRows, kCols);
+  collective::TestDistributedGlobal(4, [&] { DoPropertyColumnSplitQuantile(kRows, kCols); });
 }
 
-namespace {
-void TestSameOnAllWorkers() {
```
trivialfis, line 417: Is "workers are same" one of the tested invariants?

RAMitchell: Actually this got removed somehow - it should be let me fix it.

PR #12162, opened 2026-04-14
Backport #12094: guard __doc__ assignments for python -OO

PavelGuzenfeld:
Backport of approved/merged PR #12094 to release_2.1.0.

Upstream PR: https://github.com/dmlc/xgboost/pull/12094

release_2.1.0 is the last release line supporting Python 3.8 (>=3.10 required from 3.x). Users on Python 3.8 running with `-OO` hit `AttributeError: 'NoneType' object has no attribute 'replace'` at import time because `assert XGBModel.fit.__doc__ is not None` is stripped under `-OO` and `__doc__` is None.

This cherry-picks the guard change from #12094 onto release_2.1.0.

### DIFF
--- a/python-package/xgboost/dask/__init__.py
+++ b/python-package/xgboost/dask/__init__.py
@@ -2003,7 +2003,8 @@ class DaskXGBClassifier(DaskScikitLearnBase, XGBClassifierBase):
             iteration_range=iteration_range,
         )
 
-    predict_proba.__doc__ = XGBClassifier.predict_proba.__doc__
+    if XGBClassifier.predict_proba.__doc__ is not None:
+        predict_proba.__doc__ = XGBClassifier.predict_proba.__doc__
 
     async def _predict_async(
         self,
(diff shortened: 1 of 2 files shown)


hcho3:
Sorry, we don't think this merits backporting to old releases like 2.1.0

hcho3:
No, we are not willing to accept backports to `release_2.1.0`.

PavelGuzenfeld:
I hear you. We are using Jetson Xavier that is bound to python 3.8.
Is there any later version that supports python 3.8  we might back port to? 
Otherwise one must maintain local version for no good reason.

trivialfis:
Please note that, even if it's back ported, we won't be able to make a public binary release as the CI infrastructure has been revised since then. The 2.1 branch no longer has the CI to build artifacts.

PavelGuzenfeld:
> Please note that, even if it's back ported, we won't be able to make a public binary release as the CI infrastructure has been revised since then. The 2.1 branch no longer has the CI to build artifacts.

That would be very helpful, appreciate it.

PavelGuzenfeld:
Many thanks for the merge @trivialfis :)

PR #12176, opened 2026-04-20
Set device in test helper.

trivialfis:


trivialfis:
The device ordinal is a forever footgun.

PR #11942, opened 2026-01-20
Add Python Android support

mhsmith:
I'm currently targeting the 3.0.0 branch as part of a project I'm working on, but I'll rebase this PR on main once it's complete.

* Requires https://github.com/pypa/cibuildwheel/pull/2695
* Requires https://github.com/numpy/numpy/pull/30412
* Requires https://github.com/scipy/scipy/pull/24341

PR #12181, opened 2026-04-24
Install on Termux/Android

poti1:
I would like to be able to easily `pip install xgboost` on my termux without having to install from source code.

These minor changes allow installation on "Android".


trivialfis:
Based on https://github.com/dmlc/xgboost/pull/11942 , it seems modifying platform tags alone is not sufficient.

poti1:
I see I got 3 of the 4 changes from the other MR.

I'll then close this duplicate MR.

PR #12182, opened 2026-04-25
Compact categorical cut layout to observed codes, harden distributed CatContainer

### LINKED ISSUE #12177: XGBoost silent process kill with sparse polars Categorical codes
# XGBoost silent process kill with sparse polars Categorical codes

**Affects:** XGBoost 3.2.0, `tree_method="hist"`, `enable_categorical=True`
**Platform:** Windows 10 Pro 10.0.19045, 128 GB RAM, 237 GB pagefile
**Symptom:** process exits silently with code `3221226505` (`0xC0000005`,
`STATUS_ACCESS_VIOLATION`). No Python traceback, no C++ exception, no stderr.
**Expected on Linux (not tested):** same over-allocation occurs per C++ analysis,
but glibc commits pages lazily so the process likely survives; RAM usage would
explode and bin values may be wrong.

---

## TL;DR

This bug has been plaguing me for 2 days, spent countless hours to narrow it down.

`xgboost::common::AddCategories` in `src/common/quantile.cc` allocates the
per-feature cut-values buffer with `max_category_code + 1` entries instead of
`n_unique` entries. A single `polars.Categorical` colфmn where one value has a
physical code of `2_526_058` while `n_unique = 51` causes XGBoost to allocate
`~10 MB` for what should be a `~200 B` buffer. On Windows this over-allocation
corrupts the heap and kills the process.

**Workaround:** cast the column to `pl.Enum(sorted(unique_values))` before
fitting. `pl.Enum` enforces compact codes `[0..n_unique-1]` — XGBoost allocates
the correct buffer size.

**Fix:** compact physical codes to `[0..n_unique-1]` inside `AddCategories`
before building the cut-values buffer. See section 4 and
[`fix_add_categories_compact.cc`](./fix_add_categories_compact.cc).

---

## 1. How sparse codes arise in polars

This is not a contrived edge case. The production sequence:

1. `polars >= 1.19` makes the global `StringCache` permanently enabled —
   `disable_string_cache()` is a no-op in 1.33+. The cache assigns a physical
   code to every new unique string, incrementing a global counter.
2. By the time the ML pipeline runs, a low-cardinality `category` column
   (89 distinct strings, loaded from parquet) holds compact codes `0..87`.
3. Casting a high-cardinality string column (~2.5 M unique values) to
   `pl.Categorical` fills the cache with ~2.5 M additional entries, advancing
   the global code counter to ~2 526 057.
4. `fill_null("__MISSING__")` on `category` registers the new string
   `"__MISSING__"` in the polluted cache at code `2_526_058`.

End state for the `category` column:

| physical code | string |
|---|---|
| 0 | `cat_0` |
| 1 | `cat_1` |
| ... | 86 other strings at codes 2..87 |
| **2_526_058** | `__MISSING__` |

`n_unique = 89`. `max_physical_code = 2_526_058`.
XGBoost receives this column and calls `AddCategories` with a set of 89 floats
whose maximum value is `2_526_058`. The bug then fires.

---

## 2. C++ root cause — `AddCategories` in `src/common/quantile.cc`

```cpp
// xgboost/src/common/quantile.cc  (~lines 389-402, simplified)
auto AddCategories(std::set<float> const &categories, HistogramCuts *cuts) {
  if (categories.empty()) return InvalidCat();
  auto &cut_values = cuts->cut_values_.HostVector();
  auto max_cat = *std::max_element(categories.cbegin(), categories.cend());
  CheckMaxCat(max_cat, categories.size());
  for (bst_cat_t i = 0; i <= AsCat(max_cat); ++i) {
    cut_values.push_back(i);   // BUG: loop iterates max_cat+1 times, not categories.size()
  }
  return max_cat;
}
```

The loop runs `max_cat + 1` times regardless of how many distinct categories
were observed. With `max_cat = 2_526_058` and `n_unique = 51`:

| quantity | value |
|---|---|
| `categories.size()` | 89 |
| `max_cat` | 2 526 058 |
| loop iterations | 2 526 059 |
| bytes pushed | `2 526 059 × 4` ≈ **9.64 MB** |
| bytes needed | `89 × 4` = **356 B** |
| over-allocation factor | **~28 000×** |

### 2.1 Why the existing guard does not help

`src/common/categorical.h` rejects codes above `2^24 = 16_777_216`:

```cpp
constexpr inline bst_cat_t OutOfRangeCat() {
  return static_cast<bst_cat_t>(16777217) - 1;
}
```

`2_526_058 << 2^24`, so `CheckMaxCat` passes silently and the loop proceeds.
Lowering the threshold is not the fix — even a max_cat of `100_000` produces
a 400 KB buffer for 51 categories, still 8 000× waste.

### 2.2 Why Windows crashes

On Windows the ~10 MB allocation, combined with the heap state produced by
loading the large Arrow buffers, triggers SEH `0xC0000005`. No Python-level
exception is raised — the OS kills the process instantly. The exact allocator
mechanism (page-guard placement, heap fragmentation pattern) was not
instrumented; the crash is confirmed empirically.

On Linux (not tested) glibc commits pages lazily, so the process likely
survives, but wastes ~10 MB of heap per categorical feature and may produce
wrong predictions if histogram bins are indexed past the legitimate range.

### 2.3 Why the crash is machine-dependent

The crash requires a specific heap layout at the moment of the `AddCategories`
allocation. On a 128 GB Windows machine that has loaded a large multi-column
parquet (~760 MB of Arrow buffers) the heap is fragmented enough that the
over-allocation hits a guard page. On machines with less RAM or a different
allocation history the same over-allocation succeeds silently. The underlying
bug (wrong buffer size) is present on all machines.

---

### 3.1 Pure-synthetic standalone reproducer (no real data needed)

[`repro_xgb_synthetic_v2.py`](./repro_xgb_synthetic_v2.py) — confirmed to
crash on Windows 10 / 128 GB RAM / polars 1.33.1 / xgboost 3.2.0.

```
python repro_xgb_synthetic_v2.py
```

What it does:

1. Generates 2 526 059 unique ASCII strings with heavy-tailed lengths
   (p50 ≈ 40 chars, p99 ≈ 4 800 chars) using parallel worker processes and
   saves them to a temp parquet. Loading via parquet — not from a Python list —
   is essential: it replicates the Arrow buffer allocation pattern of a real
   parquet load, which is what produces the heap state that makes Windows crash.
2. Casts the loaded strings to `pl.Categorical`, priming the global StringCache
   with 2 526 059 entries (keeps the series alive to prevent cache eviction in
   polars >= 1.40).
3. Builds a `category` column (51 unique strings) and casts it through the
   polluted cache → physical codes land above 2 526 000.
4. Calls `fill_null("__MISSING__")` → `__MISSING__` gets code 2 526 109.
5. Calls `XGBClassifier.fit()` → process is silently killed.

```
python repro_xgb_synthetic_v2.py --workaround
```

With `--workaround` the script casts `category` to `pl.Enum(sorted_uniques)`
after `fill_null`. Physical codes collapse to `[0..50]`, XGBoost allocates
204 bytes instead of 10 MB, fit completes normally.

Generation took ~116 s on the prod Windows machine (exact core count unknown). The temp parquet is deleted
afterwards; use `--keep-parquet` to reuse it on subsequent runs.

### 3.2 Diagnostic output from a confirmed crash run

```
polars 1.33.1, xgboost 3.2.0, platform=win32
generating 2_526_059 synthetic primer strings...
  saved ...synthetic_skills.parquet (761.6 MB) in 115.8s
loading primer parquet and priming StringCache...
  StringCache primed (2_526_059 entries) in 3.1s
after category cast: n_unique=51, codes_max=2526108
after fill_null:     n_unique=51, codes_max=2526109

fitting XGB train=(211168, 1) val=(100000, 1) -- expect silent kill (0xC0000005) on Windows
[process exits here with rc=3221226505, no further output]
```

---

## 4. Proposed fix — compact codes before sketching

The root cause is that `AddCategories` treats the physical code value as a bin
index. The fix: build a compact `code → rank` mapping once per feature so the
cut-values buffer has exactly `categories.size()` entries.

```cpp
// BEFORE: buffer sized by max_cat+1
for (bst_cat_t i = 0; i <= AsCat(max_cat); ++i) {
  cut_values.push_back(i);
}

// AFTER: buffer sized by n_unique; downstream indexing uses rank, not code
bst_cat_t rank = 0;
for (float cat : categories) {           // std::set is sorted
  cut_values.push_back(cat);             // preserve physical code for serialisation
  code_to_rank[AsCat(cat)] = rank++;
}
```

Memory impact on this bug's trigger case: `9.64 MB → 356 B` per feature.

Impact on well-formed inputs (compact codes `[0..k-1]`): `code_to_rank[i] == i`
for all `i` — identity mapping, no behavioural change, saved models remain
byte-compatible.

Full patch with unit tests, serialisation notes, and a list of the ~6
downstream indexing sites that need a one-line update is in
[`fix_add_categories_compact.cc`](./fix_add_categories_compact.cc).

Three options were considered:

| option | verdict |
|---|---|
| **1. Compact codes (this patch)** | Preferred. Fixes root cause, O(k log k), zero regression on valid inputs. |
| 2. Size cuts by `n_unique`, keep physical-code indexing | Fixes allocation but leaves ~6 indexing sites inconsistent. |
| 3. Lower `OutOfRangeCat()` threshold | Breaking change; does not fix sub-threshold sparse cases. |

---

## 5. (Temp) workaround (confirmed in production)

Cast every `pl.Categorical` column to `pl.Enum(sorted(unique_values))` before
passing it to XGBoost. `pl.Enum` enforces compact physical codes by
construction — no over-allocation, no crash.

```python
for col in cat_columns:
    uniques = sorted(df[col].unique().drop_nulls().to_list())
    df = df.with_columns(pl.col(col).cast(pl.Enum(uniques)))
```

Should I create a PR?

---

[repro_xgb_synthetic_v2.py](https://github.com/user-attachments/files/26915441/repro_xgb_synthetic_v2.py)

[fix_add_categories_compact.cc.txt](https://github.com/user-attachments/files/26915493/fix_add_categories_compact.cc.txt)

fingoldo:
Closes #12177

## Summary

A polars Categorical built against a primed StringCache holds a few unique strings at very large physical codes (e.g. observed code 2_526_058 with only 51 unique values, per the original reproducer). `AddCategories` / `MakeCuts` emitted a dense `[0..max_observed_code]` cut layout per categorical feature, blowing `cut_values` to `O(max_code)` and triggering a silent process kill on Windows (`STATUS_ACCESS_VIOLATION`, exit code `3221226505` / `0xC0000005`) when the saved-tree bitfield was sized off `cuts.MaxCategory()+1`.

This PR stores one cut per observed code and propagates per-feature compact sizing through `MaxNumBinPerFeat()` and the saved-tree path.

## What changed

- `src/common/quantile.{cc,cu}`: `AddCategories` writes one cut per observed code; column-wise empty-worker emits a placeholder cut so downstream sizing never sees the `-1.f` sentinel.
- `src/data/gradient_index.h`: `MaxNumBinPerFeat()` returns the widest per-feature bin range scanned over `cut.Ptrs()` (correct upper bound under compact non-dense layouts).
- `src/tree/{hist/evaluate_splits.h,updater_gpu_hist.cu}`: saved-tree `cat_bits` sized via the new `common::SizeCatBitsForMaxCode` using the per-feature observed max code.
- `src/data/cat_container_hash.{h,cuh}`: dual independent 64-bit content digest (FNV-1a + secondary mixing). Lets distributed runs detect divergent ref dictionaries before training. Device sibling folds bytes on-device.
- `src/data/quantile_dmatrix.{cc,cu}`: digest gate via `AllreduceRefAgreement` + `AllreduceDigestAndCheck`; CPU and GPU paths share allreduce shape for this gate.
- `src/data/cat_container.{h,cc,cu}`: `Sort()`/`Copy()`/`Save()`/`Load()` are mutex-guarded and idempotent via a persisted `sorted_` flag; back-compat defaults preserved.
- `src/data/adapter.h`, `device_adapter.cuh`: `CachedRefMapping` aliases the reference DMatrix's `CatContainer` instead of allocating a fresh device copy on every dispatch.

## Behavior change

`MaxNumBinPerFeat()` now scans `cut.Ptrs()` per-feature instead of `cut.MaxCategory()+1`. Equal to the old value in dense-codes layouts; strictly smaller (correct) under compact layouts produced by this PR.

## Performance

Sweep `primer_size in {1k, 50k, 200k, 500k}`, `n_real_cats=16`, 2048 rows, 8 boost rounds, `tree_method=hist`. Same `xgboost.dll` build for both:

| primer | CPU train master -> ours | GPU train master -> ours |
|--:|--:|--:|
| 50k  | 0.52s -> 0.04s (12.9x) | 0.76s -> 0.09s (8.5x)  |
| 500k | 5.42s -> 0.28s (19.5x) | 6.78s -> 0.19s (36x)   |

Peak RSS at primer=500k drops ~33% during training and ~39% during construct on both devices (CPU train: 115 -> 76 MB; GPU train: 116 -> 79 MB; CPU construct: 100 -> 60 MB; GPU construct: 99 -> 61 MB). `inplace_predict()` output is byte-identical between master and this PR for every primer size on both CPU and CUDA.

In the dense-codes regime (`max_observed_code == n_unique_observed - 1`, typical sklearn `LabelEncoder` or `pd.factorize()` output) the new path produces the same `cut_values` as master; equivalent in performance and bit-for-bit results.

## Testing

- `testxgboost --gtest_filter='Quantile.AddCategories*:CatContainer.*:CatContainerHash.*'`: 20/20 pass.
- `pytest tests/python/test_with_polars.py`: 15/16 pass. The single failure is `test_polars_basic[QuantileDMatrix]` on the boolean-column path, reproducible on clean `origin/master` (`e0d3dfd5`) with byte-identical error - pre-existing, not introduced here.
- `pytest tests/python-gpu/test_gpu_with_polars.py`: 4/4 pass on my old GTX 1050 Ti.
- `pytest tests/test_distributed/test_with_dask/test_with_dask.py::test_categorical_ref_quantile_dmatrix`: pass on dask `LocalCluster`.
- `tests/test_distributed/test_gpu_with_dask/test_gpu_with_dask.py::test_categorical_ref_quantile_dmatrix`: GPU mirror authored to match the CPU version's structure but not exercised locally - the suite imports `cudf` at module top and RAPIDS has no Windows wheel.

## Pre-existing concerns surfaced (not addressed here)

- `src/tree/gpu_hist/evaluator.cu`: `node_categorical_storage_size_` still sized by global `MaxCategory()+1`. Per-feature evaluator buffers would close the residual runtime memory hit.
- `cat_container.cu` `CopyTo` and `quantile.cu` `CopyTo`/`PruneImpl`/`FixError` use the legacy default stream while the surrounding pipeline runs on `ctx->CUDACtx()->Stream()`.
- `cpu_impl::MakeSketches` calls `SyncFeatureType` (an extra `Allreduce`) with no GPU sibling - pre-existing CPU/GPU pipeline-shape gap; this PR's digest gate itself is shape-symmetric across CPU and GPU branches.
- `proxy_dmatrix.cuh` `DispatchAny`'s "dummy return value" branch via `std::any_cast<CudfAdapter>` throws `std::bad_any_cast` on the `type_error` sentinel path; latent (today no callers opt in).
- `bst_cat_t n_total_cats_`: read unlocked from `HasCategorical()` while `Copy()` writes it under `sort_mu_`. Aligned-int32 reads make this benign on x86/ARM but TSAN-flaggable.
- `MaxNumBinPerFeat()` is now `O(n_features)` per call; `gradient_index.cc::ResizeIndex` already caches per call. A class-level cache would amortize across all callers.

### DIFF
--- a/include/xgboost/data.h
+++ b/include/xgboost/data.h
@@ -230,6 +230,7 @@ class MetaInfo {
   [[nodiscard]] CatContainer const* Cats() const;
   [[nodiscard]] CatContainer* Cats();
   [[nodiscard]] std::shared_ptr<CatContainer const> CatsShared() const;
+  [[nodiscard]] std::shared_ptr<CatContainer> CatsShared();
   /**
    * @brief Setter for categories.
    */
@@ -726,6 +727,7 @@ class DMatrix {
   [[nodiscard]] std::shared_ptr<CatContainer const> CatsShared() const {
     return this->Info().CatsShared();
   }
+  [[nodiscard]] std::shared_ptr<CatContainer> CatsShared() { return this->Info().CatsShared(); }
 
  protected:
   virtual BatchSet<SparsePage> GetRowBatches() = 0;
--- a/ops/conda_env/aarch64_test.yml
+++ b/ops/conda_env/aarch64_test.yml
@@ -26,6 +26,7 @@ dependencies:
 - llvmlite
 - loky>=3.5.1
 - pyarrow
+- polars
 - pyspark>=4.0.0
 - cloudpickle
 - pip:
--- a/ops/conda_env/macos_cpu_test.yml
+++ b/ops/conda_env/macos_cpu_test.yml
@@ -28,4 +28,5 @@ dependencies:
 - awscli
 - loky>=3.5.1
 - pyarrow
+- polars
 - cloudpickle
--- a/ops/conda_env/win64_test.yml
+++ b/ops/conda_env/win64_test.yml
@@ -17,3 +17,4 @@ dependencies:
 - py-ubjson
 - loky>=3.5.1
 - pyarrow
+- polars
--- a/src/common/categorical.h
+++ b/src/common/categorical.h
@@ -21,6 +21,19 @@ XGBOOST_DEVICE bst_cat_t AsCat(T const& v) {
   return static_cast<bst_cat_t>(v);
 }
 
+/**
+ * @brief Storage size for a CatBitField whose largest valid bit is @p max_code.
+ *
+ * Widens to size_t before +1 so max_code near INT32_MAX cannot trigger signed-overflow
+ * UB on bst_cat_t = int32_t.
+ *
+ * @return Storage size in @c CatBitField::value_type units.
+ */
+[[nodiscard]] inline std::size_t SizeCatBitsForMaxCode(bst_cat_t max_code) {
+  CHECK_GE(max_code, 0);
+  return CatBitField::ComputeStorageSize(static_cast<std::size_t>(max_code) + 1);
+}
+
 /* \brief Whether is fidx a categorical feature.
  *
  * \param ft   Feature type for all features.
--- a/src/common/quantile.cc
+++ b/src/common/quantile.cc
@@ -533,14 +533,18 @@ void AddCategories(std::set<float> const &categories, float *max_cat, HistogramC
     InvalidCategory();
   }
   auto &cut_values = cuts->cut_values_.HostVector();
-  // With column-wise data split, the categories may be empty.
-  auto feature_max_cat =
-      categories.empty() ? 0.0f : *std::max_element(categories.cbegin(), categories.cend());
+  if (categories.empty()) {
+    // column-wise split: emit a placeholder cut and treat the synthetic 0.0f as the
+    // observed max so downstream sizing (evaluator.cu MaxCategory()+1) does not see -1
+    cut_values.push_back(0.0f);
+    *max_cat = std::max(*max_cat, 0.0f);
+    return;
+  }
+  auto feature_max_cat = *std::max_element(categories.cbegin(), categories.cend());
   CheckMaxCat(feature_max_cat, categories.size());
   *max_cat = std::max(*max_cat, feature_max_cat);
-  for (bst_cat_t i = 0; i <= AsCat(feature_max_cat); ++i) {
-    cut_values.push_back(i);
-  }
+  // one cut per observed code; categories is sorted ascending
+  cut_values.insert(cut_values.end(), categories.cbegin(), categories.cend());
 }
 
 HistogramCuts HostSketchContainer::MakeCuts(Context const *ctx, MetaInfo const &info) {
--- a/src/common/quantile.cu
+++ b/src/common/quantile.cu
@@ -705,16 +705,23 @@ HistogramCuts SketchContainer::MakeCuts(Context const *ctx, bool is_column_split
     auto column = Span<SketchEntry const>{h_entries.data() + begin, end - begin};
 
     if (IsCat(h_feature_types, i)) {
-      auto column_size = std::max(static_cast<std::size_t>(1), column.size());
-      auto feature_max = column.empty() ? 0.0f : column.back().value;
-      if (std::any_of(column.cbegin(), column.cend(),
-                      [](auto const &entry) { return InvalidCat(entry.value); })) {
-        InvalidCategory();
-      }
-      CheckMaxCat(feature_max, column_size);
-      max_cat = std::max(max_cat, feature_max);
-      for (std::size_t cat = 0; cat <= static_cast<std::size_t>(feature_max); ++cat) {
-        h_out_cut_values.push_back(cat);
+      if (column.empty()) {
+        // column-split worker with no rows: emit a placeholder cut and treat the
+        // synthetic 0.0f as observed max so MaxCategory() is never -1
+        h_out_cut_values.push_back(0.0f);
+        max_cat = std::max(max_cat, 0.0f);
+      } else {
+        auto feature_max = column.back().value;
+        if (std::any_of(column.cbegin(), column.cend(),
+                        [](auto const &entry) { return InvalidCat(entry.value); })) {
+          InvalidCategory();
+        }
+        CheckMaxCat(feature_max, column.size());
+        max_cat = std::max(max_cat, feature_max);
+        // one cut per observed physical code; column sorted ascending
+        for (auto const &entry : column) {
+          h_out_cut_values.push_back(entry.value);
+        }
       }
     } else {
       summary.Reserve(column.size());
--- a/src/data/adapter.cc
+++ b/src/data/adapter.cc
@@ -19,11 +19,12 @@
 
 namespace xgboost::data {
 namespace {
-auto GetRefCats(Json handle) {
-  auto cats = reinterpret_cast<CatContainer const*>(get<Integer const>(handle));
+// Pair the owning ref CatContainer pointer with its host view
+[[nodiscard]] std::pair<CatContainer*, enc::HostColumnsView> GetRefCats(Json handle) {
+  auto cats = reinterpret_cast<CatContainer*>(get<Integer const>(handle));
   CHECK(cats);
   auto h_cats = cats->HostView();
-  return h_cats;
+  return {cats, h_cats};
 }
 }  // anonymous namespace
 
@@ -32,7 +33,9 @@ ColumnarAdapter::ColumnarAdapter(StringView columns) {
 
   if (IsA<Object>(jdf)) {
     // Has reference categories.
-    this->ref_cats_ = GetRefCats(jdf["ref_categories"]);
+    auto [ref_cats_ptr, ref_cats_view] = GetRefCats(jdf["ref_categories"]);
+    this->ref_cats_ptr_ = ref_cats_ptr;
+    this->ref_cats_ = ref_cats_view;
     jdf = jdf["columns"];
   }
 
--- a/src/data/adapter.h
+++ b/src/data/adapter.h
@@ -11,7 +11,8 @@
 #include <cstdint>    // for uint8_t
 #include <limits>     // for numeric_limits
 #include <memory>     // for unique_ptr, make_unique
-#include <utility>    // for move
+#include <mutex>      // for once_flag, call_once
+#include <utility>    // for move, forward
 #include <variant>    // for variant
 #include <vector>     // for vector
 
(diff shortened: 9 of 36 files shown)


trivialfis:
Can we just enforce a dense encoding? Making changes across the code base is quite fragile.

fingoldo:
> Can we just enforce a dense encoding? Making changes across the code base is quite fragile.

I did not think about that, it might be an even better way. I got too excited with my current approach, it seems. I will have more time to implement and benchmark your suggestion by the end of this week. Can you approve running the rest of CI workflows in the meantime (there was one test I forgot to run locally that uncovered a bug, should be fixed by now)?

trivialfis:
Thank you for the reply, we will go through the CI when we have a solution to test (it's approval per commit), I don't think we will merge the PR in its current design.

fingoldo:
> Thank you for the reply, we will go through the CI when we have a solution to test (it's approval per commit), I don't think we will merge the PR in its current design.

1) To make sure I implement the right thing - the approach you have in mind is to auto-compact categorical codes at C++ ingestion (e.g. columnar.cc::GetArrowDictionary / CudfAdapter ctor) so xgboost transparently handles sparse input while keeping the "dense codes inside" invariant. Did I read that right?

2) A few pieces of the current diff are independent of the sparse-codes design and don't need to be discarded with it. Are you OK if I spin them off as small focused follow-up PRs once the main one is rewritten? I think they can be valuable.

_cat_container_hash + AllreduceDigestAndCheck_ - cross-worker check that ref dictionaries agree (surfaces divergent refs at construction instead of silent mistraining)

_ValidateCatStrArrayOffsets_ - host-side Arrow offset validation in the categorical adapter (front==0, monotonic, in-range)

_CatContainer Sort/Copy/Save/Load_ mutex-guarding for thread-safety.

trivialfis:
My approach is just to reject any non-dense category index: https://github.com/dmlc/xgboost/pull/12240 The ETL pipeline can handle it.

PR #12193, opened 2026-04-29
Fix encoding OOV category.

### LINKED ISSUE #12189: [bug] SearchSorted missing equality check causes OOV categories to silently map to wrong training category
**XGBoost version:** 3.2.0

**Environment:** Python 3.11, pandas, CPU

When using native categorical features (`enable_categorical=True`) with a string-valued categorical column, out-of-vocabulary (OOV) category values at inference time are **silently mapped to a wrong training category** instead of raising an error, provided the OOV string is lexicographically less than the last training category.

This is a bug in the `SearchSorted` function introduced as part of the auto re-coder feature ([#11088](https://github.com/dmlc/xgboost/issues/11088), merged in 3.1.0). The function uses `std::lower_bound` but **does not verify that the found position is an exact match**, so any OOV string that falls within the sorted range of training categories returns a valid-looking code for a different category.

**Minimal reproducible example**
```
import xgboost as xgb
import pandas as pd
import numpy as np

# Train with categories A, B, C
train_df = pd.DataFrame({
    'email': pd.Categorical(['A', 'B', 'C', 'A', 'B', 'B']),
    'x':     [1.0, 2.0, 3.0, 1.5, 2.5, 3.5]
})
model = xgb.XGBClassifier(enable_categorical=True, tree_method='hist', n_estimators=10)
model.fit(train_df, np.array([0, 1, 0, 1, 0, 1]))

# 'B2' falls between 'B' and 'C' → silently scores as if it were 'C'
test_pass = pd.DataFrame({
    'email': pd.Categorical(['B2', 'C']),
    'x':     [2.0, 2.0]
})
probs = model.predict_proba(test_pass)
print(probs)

# 'D' > 'C' (last category) → correctly raises XGBoostError
test_fail = pd.DataFrame({
    'email': pd.Categorical(['D', 'C']),
    'x':     [2.0, 2.0]
})
try:
    model.predict_proba(test_fail)
except xgb.core.XGBoostError as e:
    print("Expected error:", e)
```

### **Expected behaviour**

`predict_proba` (or `inplace_predict`) should raise an error for `B2` exactly as it does for `D`, because `B2` was not in the training set.

### **Actual behaviour**

`B2` is silently accepted and mapped to the code for `C` (the first training category whose sorted position `lower_bound` returns). No warning or error is produced.

### **Root cause**

In `src/encoder/ordinal.h`, the `SearchSorted` function:
```
auto ret_it = std::lower_bound(it, it + haystack.size(), SearchKey(), comparator);
if (ret_it == it + haystack.size()) {
    return detail::NotFound();  // only fires when OOV > ALL training categories
}
return *ret_it;                 // ← returns insertion point, not a verified exact match
```

`std::lower_bound` returns the first position where the element is `>=` the needle, not necessarily `==`. The missing equality check means `NotFound()` is never returned for OOV strings that fall within the sorted range, so `ReportMissing` is never called for them.

The fix would be to verify the found element equals the needle before returning the index.

### **Impact**

OOV categorical values at inference time receive **a wrong score** (the score of a different, arbitrarily-chosen training category) with no error or warning. This is a correctness issue for any model using string-typed native categorical features and is particularly hard to detect in production since the model continues to return plausible-looking scores.

trivialfis:
Close https://github.com/dmlc/xgboost/issues/12189

### DIFF
--- a/python-package/xgboost/testing/ordinal.py
+++ b/python-package/xgboost/testing/ordinal.py
@@ -421,6 +421,7 @@ def run_cat_invalid(device: Device) -> None:
 
 
 def run_cat_oov_in_range(device: Device) -> None:
+    """Test OOV categories whose sorted insertion point is inside the training set."""
     Df, _ = get_df_impl(device)
 
     def check(train_df: Any, pred_df: Any, y: np.ndarray) -> None:
--- a/python-package/xgboost/testing/ordinal.py
+++ b/python-package/xgboost/testing/ordinal.py
@@ -420,6 +420,43 @@ def run_cat_invalid(device: Device) -> None:
         )
 
 
+def run_cat_oov_in_range(device: Device) -> None:
+    Df, _ = get_df_impl(device)
+
+    def check(train_df: Any, pred_df: Any, y: np.ndarray) -> None:
+        for DMatrixT in (DMatrix, QuantileDMatrix):
+            Xy = DMatrixT(train_df, y, enable_categorical=True)
+            booster = train({"device": device}, Xy, num_boost_round=1)
+
+            with pytest.raises(
+                ValueError, match="Found a category not in the training"
+            ):
+                booster.inplace_predict(pred_df)
+
+            fmat = DMatrixT(pred_df, enable_categorical=True)
+            with pytest.raises(
+                ValueError, match="Found a category not in the training"
+            ):
+                booster.predict(fmat)
+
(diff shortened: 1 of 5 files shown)


RAMitchell (approved):
I am still not super happy about us doing recoding.

trivialfis:
> I am still not super happy about us doing recoding.

Yeah, at some point I gave up arguing against that.

Issue #11947, opened 2026-01-22
[RFC] Add interpretability API as xgboost.interpret module functions


RAMitchell:
## Summary

We have new work underway on shapley values and other related interpretability concepts. This will add new functionality, however the current feature importance/shapley features are included in the predict API. I propose extending the python api with a module (`xgboost.interpret`) for interpretability, containing stateless functions exposing upcoming features.

These functions accept either a `Booster` or an sklearn-style `XGB*` model, plus `DMatrix`/array-like inputs, and return well-typed results (arrays and  or light-weight result objects).

## Motivation
- Minimize disruption to existing `Booster` / sklearn APIs while adding interpretability features.
- Improve discoverability and documentation (module-level functions are easy to document and test).
- Allow incremental implementation: start as wrappers over existing `predict(pred_contribs=..., pred_interactions=...)`, then evolve internals (esp. top-k) without changing the public API.

## Proposed public API
Add a new module:
- `xgboost/interpret.py`

Functions (accept `Booster | XGBModel` and `DMatrix | array-like | pandas`):
- `shap_values(model, X,*, X_background=None, output_margin=False, iteration_range=None, approx=False, validate_features=True, feature_names=None, return_bias=False)`
- `shap_interactions(model, X, *, X_background=None, output_margin=False, iteration_range=None, approx=False, validate_features=True, feature_names=None)`
- `topk_interactions(model, X, *, X_background=None, k=50, metric="mean_abs", output_margin=False, iteration_range=None, validate_features=True, feature_names=None)` - Note: possibly just fold this into shap_interactions
- `partial_dependence(model, X, *, features, grid_resolution=50, percentiles=(0.05,0.95), grid=None, sample_weights=None, random_state=0, output="prediction", iteration_range=None)`
- Possibly adding shap_values, shap_values methods to booster/sklearn class for convenience

## Dispatch/behavior notes
- Internally normalize `model` to a `Booster` via `model` being `Booster` or having `get_booster()`.
- Normalize `X` to `DMatrix` if needed; respect feature names where possible.
- Initial SHAP implementations can wrap existing `Booster.predict(..., pred_contribs=True/pred_interactions=True)` for compatibility.
- `topk_interactions` should ideally avoid materializing full (n, p, p) tensors; target a C++ implementation to compute aggregated top-k pairs efficiently.

## Return types
Prefer lightweight result objects to keep outputs consistent and extensible:
- `ShapValues(values, base_values, feature_names, model_output, ...)`
[...]

PR #12208, opened 2026-05-12
Add basic xgboost.interpret.shap_values API

RAMitchell:
## Summary

Adds an initial `xgboost.interpret` module with a basic `shap_values` function.

This is a small first step toward #11947. The implementation wraps the existing `Booster.predict(pred_contribs=True)` path and keeps the behavior intentionally narrow while establishing the public module/function entry point.

## Changes

- Add `xgboost.interpret.shap_values`
- Accept either a `Booster` or sklearn-style XGBoost model
- Accept `DMatrix` or array-like inputs
- Return feature SHAP values without the bias column by default
- Support `return_bias=True` to return `(values, bias)`
- Support optional temporary `device=` override, restoring the booster config after prediction
- Reject `X_background` for now, since interventional SHAP is not implemented yet
- Add focused Python tests

## Notes

This does not add generated docs yet. The function has a docstring, but a dedicated Sphinx page can follow once the initial API shape is agreed.

## Testing

```bash
PYTHONPATH=/home/nfs/rorym/xgboost-wt/interpret-basic/python-package \
  conda run -n xgboost python -m pytest tests/python/test_interpret.py
```

Result: `5 passed`

Pre-commit passed during commit, including `ruff`, `ruff format`, and `pylint`.

### DIFF
--- a/python-package/xgboost/interpret.py
+++ b/python-package/xgboost/interpret.py
@@ -4,7 +4,7 @@ from typing import Optional, Tuple, Union
 
 import numpy as np
 
-from ._typing import ArrayLike, IterationRange
+from ._typing import ArrayLike, FloatCompatible, IterationRange
 from .core import Booster, DMatrix
 
 
@@ -12,7 +12,7 @@ def _as_booster(model: object) -> Booster:
     if isinstance(model, Booster):
         return model
     get_booster = getattr(model, "get_booster", None)
-    if get_booster is None:
+    if not callable(get_booster):
         raise TypeError(
             "`model` must be an xgboost.Booster or an object with get_booster()."
         )
@@ -33,13 +33,17 @@ def _get_iteration_range(
     return iteration_range
 
 
-def _as_prediction_dmatrix(model: object, X: Union[DMatrix, ArrayLike]) -> DMatrix:
+def _as_prediction_dmatrix(
+    model: object, X: Union[DMatrix, ArrayLike], missing: Optional[FloatCompatible]
+) -> DMatrix:
     if isinstance(X, DMatrix):
+        if missing is not None:
+            raise ValueError("`missing` must not be specified when `X` is a DMatrix.")
         return X
 
     return DMatrix(
         X,
-        missing=getattr(model, "missing", None),
+        missing=missing if missing is not None else getattr(model, "missing", None),
         nthread=getattr(model, "n_jobs", None),
         feature_types=getattr(model, "feature_types", None),
         enable_categorical=getattr(model, "enable_categorical", False),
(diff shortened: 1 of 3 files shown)


python-package/xgboost/interpret.py
```diff
@@ -0,0 +1,146 @@
...
+    *,
+    device: Optional[str],
+    kwargs: dict,
+) -> np.ndarray:
+    if device is None:
+        return booster.predict(data, **kwargs)
+
+    config = booster.save_config()
+    try:
+        booster.set_param({"device": device})
```
trivialfis: I'm not sure if we should add a new code path to set the device. This setter makes the function mutable

python-package/xgboost/interpret.py
```diff
@@ -0,0 +1,146 @@
...
+
+def shap_values(  # pylint: disable=too-many-arguments
+    model: object,
+    X: Union[DMatrix, ArrayLike],
+    *,
+    X_background: Optional[Union[DMatrix, ArrayLike]] = None,
+    device: Optional[str] = None,
+    output_margin: bool = False,
+    iteration_range: Optional[IterationRange] = None,
+    approx: bool = False,
```
trivialfis: Is the approx still valuable?

python-package/xgboost/interpret.py
```diff
@@ -0,0 +1,146 @@
...
+            "pred_contribs": True,
+            "approx_contribs": approx,
+            "validate_features": validate_features,
+            "iteration_range": _get_iteration_range(model, iteration_range),
+        },
+    )
+
+    values = contribs[..., :-1]
+    bias = contribs[..., -1]
+    if return_bias:
```
trivialfis: Have you considered returning the bias unconditionally?

python-package/xgboost/interpret.py
```diff
@@ -0,0 +1,146 @@
...
+    # argument in the initial API so callers can use the proposed signature.
+    _ = output_margin
+
+    booster = _as_booster(model)
+    data = _as_prediction_dmatrix(model, X)
+    contribs = _predict_contribs(
+        booster,
+        data,
+        device=device,
+        kwargs={
```
trivialfis: That's a weird way of using Python kwargs.

python-package/xgboost/interpret.py
RAMitchell: Its still used elsewhere, but should not be part of this new interface.

python-package/xgboost/interpret.py
RAMitchell: Yeah I don't like the fact that its mutable but if we don't do this its not clear that its GPU accelerated and its unclear how that might work. I'd like to keep it as optional arg but document.

python-package/xgboost/interpret.py
RAMitchell: Agree, its better to return both I think.

trivialfis (approved):

PR #12215, opened 2026-05-14
Add timeout to RTD trigger request

massy-o:
## Summary
- add an explicit timeout to the Read the Docs trigger request

## Why
The CI helper triggers the Read the Docs API with requests.post(). If the service accepts the connection but stalls, the workflow can block on the socket. A bounded timeout lets CI fail clearly instead of hanging indefinitely.

## Testing
- python3 -m py_compile ops/pipeline/trigger-rtd-impl.py
- git diff --check

### DIFF
--- a/ops/pipeline/trigger-rtd-impl.py
+++ b/ops/pipeline/trigger-rtd-impl.py
@@ -32,7 +32,7 @@ def trigger_build(token: str) -> None:
 
     URL = f"https://readthedocs.org/api/v3/projects/xgboost/versions/{branch}/builds/"
     HEADERS = {"Authorization": f"token {token}"}
-    response = requests.post(URL, headers=HEADERS)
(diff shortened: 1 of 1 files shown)


ops/pipeline/trigger-rtd-impl.py
```diff
@@ -32,7 +32,7 @@ def trigger_build(token: str) -> None:
 
     URL = f"https://readthedocs.org/api/v3/projects/xgboost/versions/{branch}/builds/"
     HEADERS = {"Authorization": f"token {token}"}
-    response = requests.post(URL, headers=HEADERS)
+    response = requests.post(URL, headers=HEADERS, timeout=30)
```
massy-o, line 35: Self-review: The request still keeps the existing status-code handling, but the workflow no longer waits indefinitely if the Read the Docs API stalls after accepting the connection.

PR #12196, opened 2026-05-03
[doc] Add R code tabs to custom_metric_obj tutorial

syedali254:
Adds R code tabs alongside existing Python examples in the Custom Objective
and Evaluation Metric tutorial using the `.. tabs::` / `.. code-tab::` Sphinx
plugin introduced in #11410.

Changes:
- Converted 8 Python code blocks to multi-language tab format
- Added equivalent R implementations for custom objective, metric, and
  multi-class examples
- Scikit-learn section left as Python-only with a note for R users

Contributes to #11413

PR #12216, opened 2026-05-18
Add doc tests.

trivialfis:
LLMs make authoring examples easier, but we need a way to validate them in CI. This PR adds tests for some snippets in the `doc`, while others are ignored for now. See notes in the `docs.rst` for more info.

A small sphinx extension is used to extract and execute R code blocks. We will be using it to check https://github.com/dmlc/xgboost/pull/12196 .


trivialfis:
cc @jameslamb not sure if this is useful for you, if so, maybe we could extract something for reusing.

trivialfis:
I will merge this if there's no objection.

chyunsu3 (approved):

doc/xgboost_doc_doctest.py
```diff
@@ -0,0 +1,201 @@
...
+        close_r_session(session_key)
+
+
+def _normalise_language(language: str | None) -> str | None:
+    if language is None:
+        return None
+    language = language.lower()
+    if language in {"py", "python", "python3"}:
+        return "python"
+    if language in {"r", "s", "rlang"}:
```
jameslamb, line 123: handling S language code blocks, wow this is thorough 😅

jameslamb (commented):
Thanks for the `@`, yeah this seems pretty useful! Registering as a Sphinx builder is a great idea.

I have another idea for your consideration, that might not require this much code... what about checking example code into the repo directly, as regular scripts (`.py`, `.R`, etc.) and including them into the docs with `literalinclude`?

```rst
.. literalinclude:: ./_example-scripts/custom_obj.R
    :language: r
```

I use this approach to keep `pydistcheck`'s documented default config up to date while also making it machine-readable for unit tests:

* include in docs: https://github.com/jameslamb/pydistcheck/blob/90315eea97014d278d960d4eb270f69c4ecad1ab/docs/configuration.rst#L40
* doc tests: https://github.com/jameslamb/pydistcheck/blob/90315eea97014d278d960d4eb270f69c4ecad1ab/tests/test_docs.py#L15
* how it renders: https://pydistcheck.readthedocs.io/en/latest/configuration.html#pyproject-toml

With a setup like that, no Sphinx builder required. And testing in CI could be something like this:

```shell
find \
  ./docs/_example_scripts \
  -type f \
  -name '*.R' \
  -exec Rscript --vanilla {} \;
```

Would also make it easier to run the project's linters over that code, to help catch other correctness issues and keep the style consistent.

doc/xgboost_doc_doctest.py
trivialfis: Yea, GPT suggested it. I can't take credit for this. My reaction to its suggestion is basically "why not"...

trivialfis:
@jameslamb Thank you for taking a look and the suggestion. Yes, the literal include could work. I don't have a strong preference.

trivialfis:
Let's go with the doc tests for now since the PR is mostly to aid LLMs writing/translating examples, keeping the snippets in the doc seems easier.

Issue #12211, opened 2026-05-13
`predict.xgboost` silently ignores `predcontrib = T`


Kodiologist:
With `xgboost` 3.2.1.1:

```R
m = xgboost::xgboost(
    x = mtcars[, c("cyl", "disp", "hp")],
    y = mtcars$mpg)
str(predict(m,
    mtcars[, c("cyl", "disp", "hp")],
    predcontrib = T))
# Named num [1:32] 21 21 22.8 21.4 18.7 ...
# - attr(*, "names")= chr [1:32] "Mazda RX4" "Mazda RX4 Wag" "Datsun 710" "Hornet 4 Drive" ...
```

What the user should do is say `type = "contrib"`, because XGBoost's two `predict` methods have inconsistent interfaces for this. But it's hard to catch what's going on because `predict.xgboost` produces no warning or error for unrecognized parameters, and `predcontrib = T` used to work in this context in previous versions of XGBoost.

trivialfis:
Thank you for sharing, I will look into this.

Sanidhyavijay24:
Hello @trivialfis , I have submitted a PR (#12218) to address this. It maps the legacy boolean flags (like `predcontrib`) to the modern type argument and adds a deprecation warning. It also adds a strict check for other arguments in `...` to prevent silent failures in the future.  Also added tests for it.
Let me know if you have any feedback!

david-cortes:
Note that most other libraries work like this. Examples:
```r
data("mtcars")
model <- lm(mpg ~ ., data=mtcars)
pred <- predict(model, mtcars, predcontrib = TRUE)
```
```r
library(glmnet)
data("mtcars")
x <- mtcars[, -1] |> as.matrix()
y <- mtcars$mpg
model <- glmnet(x, y)
pred <- predict(model, x, predcontrib = TRUE)
```
```r
library(VGAM)
data("mtcars")
model <- vglm(mpg ~ ., data=mtcars, family=poissonff)
pred <- predict(model, mtcars, predcontrib = TRUE)
```

Could nevertheless be nice to throw an informative warning.

PR #12218, opened 2026-05-19
Fixed silent ignore of legacy predict flags (#12211)

Sanidhyavijay24:
This PR addresses #12211.
It prevents `predict.xgboost` from silently ignoring legacy boolean flags (like `predcontrib = TRUE`). These flags are now mapped to the modern type argument with a deprecation warning. Additionally, any unknown arguments in `...` will now trigger an error to prevent typos from causing silent failures.

### DIFF
--- a/R-package/R/xgboost.R
+++ b/R-package/R/xgboost.R
@@ -1321,7 +1321,9 @@ xgboost <- function(
 #'
 #' Note that this check might add some sizable latency to the predictions, so it's
 #' recommended to disable it for performance-sensitive applications.
-#' @param ... Not used.
+#' @param ... Legacy boolean flags (e.g., \code{predcontrib}) are supported for
+#'   backward compatibility but are deprecated and map to the \code{type} argument.
+#'   Any other arguments passed to \code{...} will cause an error.
 #' @return Either a numeric vector (for 1D outputs), numeric matrix (for 2D outputs), numeric array
 #' (for 3D and higher), or `factor` (for class predictions). See documentation for parameter `type`
 #' for details about what the output type and shape will be.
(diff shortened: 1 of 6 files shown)


R-package/R/xgboost.R
```diff
@@ -1321,7 +1321,9 @@ xgboost <- function(
 #'
 #' Note that this check might add some sizable latency to the predictions, so it's
 #' recommended to disable it for performance-sensitive applications.
-#' @param ... Not used.
+#' @param ... Legacy boolean flags (e.g., \code{predcontrib}) are supported for
+#'   backward compatibility but are deprecated and map to the \code{type} argument.
+#'   Any other arguments passed to \code{...} will cause an error.
```
david-cortes, line 1326: This is actually not how most libraries work. Consider for example:
```r
data("mtcars")
model <- lm(mpg ~ ., data=mtcars)
pred <- predict(model, mtcars, predcontrib = TRUE)
```

Better to throw a warning than to error out.

R-package/R/xgboost.R
```diff
@@ -1352,6 +1354,47 @@ predict.xgboost <- function(
...
+    )
+    found_legacy <- intersect(names(dots), names(mapping))
+
+    if (length(found_legacy) > 0) {
+      for (legacy_arg in found_legacy) {
+        if (isTRUE(dots[[legacy_arg]])) {
+          type <- mapping[[legacy_arg]]
+          warning(
+            sprintf(
+              "Argument '%s' is deprecated. Please use type = '%s' instead.",
```
david-cortes, line 1374: It's not the case that it's deprecated. `predict.xgboost` is a new function introduced in version 3 which never had this argument. Better to raise a warning without making this choice.

R-package/R/xgboost.R
```diff
@@ -1352,6 +1354,47 @@ predict.xgboost <- function(
...
+      }
+      dots[found_legacy] <- NULL
+    }
+
+    if (length(dots) > 0) {
+      dot_names <- names(dots)
+      if (is.null(dot_names)) {
+        dot_names <- rep("<unnamed>", length(dots))
+      }
+      dot_names[dot_names == ""] <- "<unnamed>"
```
david-cortes, line 1390: Where does this `<unnamed>` come from?

david-cortes (changes requested):

trivialfis:
closing due to lack of response.

PR #12223, opened 2026-05-24
Use philox for faster sampling on GPU.

trivialfis:
This is available in C++ 26, cccl backports it. But we will need a newer CCCL for the `random` header. Marking this PR as WIP since we don't have the latest CCCL for XGBoost.

### DIFF
--- a/src/common/random.cu
+++ b/src/common/random.cu
@@ -8,6 +8,7 @@
 #include "algorithm.cuh"     // for ArgSort
 #include "cuda_context.cuh"  // for CUDAContext
 #include "device_helpers.cuh"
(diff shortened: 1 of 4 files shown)


RAMitchell:
I think the speed won't make a difference as this is the simple linear generator being used currently but the quality of the random samples will be very much improved. Honestly we shouldn't be using LCGs at all.

trivialfis:
@RAMitchell 

> I think the speed won't make a difference

I'm mostly concerning about the call to the `discard` method, with LCG, it needs to loop over the number of discarded states.

RAMitchell (approved):

PR #12224, opened 2026-05-28
Add interpretability SHAP C API

RAMitchell:
## Summary

This adds a dedicated C API entry point for SHAP values:

- `XGBoosterInterpretShapValues`
- separate output buffers for feature values and bias
- JSON config for interpretability options
- nullable background `DMatrixHandle` reserved for interventional SHAP
- Python `xgboost.interpret.shap_values` now routes through the new C API

The initial implementation supports the existing tree-path-dependent SHAP behavior by reusing the current contribution prediction path and splitting the final bias column into a separate output. Interventional SHAP is not implemented yet, but the C API shape now includes the background handle needed for it.

## Testing

- `pre-commit run --files include/xgboost/c_api.h src/c_api/c_api.cc src/common/api_entry.h tests/cpp/c_api/test_c_api.cc python-package/xgboost/interpret.py tests/python/test_interpret.py`
- `build-cpu/testxgboost --gtest_filter=CAPI.InterpretShapValues`
- `PYTHONPATH=/home/nfs/rorym/xgboost-wt/interpret-capi/python-package LD_LIBRARY_PATH=/home/nfs/rorym/xgboost-wt/interpret-capi/lib conda run -n xgboost python -m pytest tests/python/test_interpret.py`

### DIFF
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -1389,7 +1389,6 @@ XGB_DLL int XGBoosterInterpretShapValues(
   auto p_m = *static_cast<std::shared_ptr<DMatrix> *>(dmat);
   auto iteration_begin = OptionalArg<Integer>(config, "iteration_begin", Integer::Int{0});
   auto iteration_end = OptionalArg<Integer>(config, "iteration_end", Integer::Int{0});
-  bool strict_shape = OptionalArg<Boolean>(config, "strict_shape", false);
 
   learner->Predict(p_m, false, &entry.predictions, iteration_begin, iteration_end, false, false,
                    true, false, false);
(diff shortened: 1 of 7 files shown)


src/c_api/c_api.cc
```diff
@@ -1360,6 +1360,86 @@ XGB_DLL int XGBoosterPredictFromDMatrix(BoosterHandle handle, DMatrixHandle dmat
...
+  auto iteration_begin = OptionalArg<Integer>(config, "iteration_begin", Integer::Int{0});
+  auto iteration_end = OptionalArg<Integer>(config, "iteration_end", Integer::Int{0});
+
+  learner->Predict(p_m, false, &entry.predictions, iteration_begin, iteration_end, false, false,
+                   true, false, false);
+
+  std::size_t rows = p_m->Info().num_row_;
+  std::size_t cols = p_m->Info().num_col_;
+  std::size_t groups = learner->Groups();
+  auto const &contribs = entry.predictions.ConstHostVector();
```
trivialfis: Since this is a new API, maybe we should rethink whether CPU-only return is preferable.

RAMitchell: Good idea, let me try it.

src/c_api/c_api.cc
```diff
@@ -1396,47 +1438,29 @@ XGB_DLL int XGBoosterInterpretShapValues(
...
-        std::size_t value_offset = row * cols * groups + col * groups + group;
-        values[value_offset] = contribs[contrib_offset + col];
-      }
-      bias[row * groups + group] = contribs[contrib_offset + cols];
-    }
-  }
+  auto *ctx = learner->Ctx();
+  auto out_ctx = ctx->IsCUDA() ? *ctx : ctx->MakeCPU();
+  auto &values = local.ret_hdv_float;
+  auto &bias = local.ret_hdv_float_1;
```
trivialfis, line 1445: (Interactive) SHAP value has large return buffers. If we could pre-allocate memory in Python then we could avoid drawing memory from thread local, which is a global variable and cannot be easily freed.

RAMitchell:
This needs more work. I need to figure out how to route internally without going through learner->predict.

PR #12228, opened 2026-06-03
Re-parameterize the expectile objective.

trivialfis:
- Require the input alphas to be sorted.
- The raw margin is now based on the smallest alpha. Each expectile output is built by predicting a positive gap.

This way, we avoid crossing the curve. This works with expectile but not with quantile, since expectile has a proper Hessian.


trivialfis:
cc @matthieubulte @RAMitchell

RAMitchell (approved):

PR #12244, opened 2026-06-11
Small fix to categorical splits in vector leaf.

trivialfis:
Use the tree size instead of scalar tree node size for IO.


trivialfis:
Fun CI

PR #12250, opened 2026-06-16
3.3.0 release news.

trivialfis:


trivialfis:
cc @hcho3 @RAMitchell would be great if you could take a look when you are available.

PR #12270, opened 2026-06-27
[doc] Add R code tab to Random Forest tutorial

### LINKED ISSUE #11413: Make all tutorials multi-language

Ppp2048:
## Summary

This PR adds an R code tab to the Random Forest tutorial (`doc/tutorials/rf.rst`) as part of the multi-language documentation effort in #11413.

### Changes

* Converted the Python-only example into Python/R tabs.
* Added an equivalent R example using the preferred `xgboost()` API.
* Kept the Python example unchanged.

Closes #11413

### DIFF
--- a/doc/tutorials/rf.rst
+++ b/doc/tutorials/rf.rst
@@ -46,25 +46,41 @@ use a combination of random forest and gradient boosting strategy. It will perfo
 each round. If early stopping is not enabled, the final model will consist of
 ``num_parallel_tree`` * ``num_boost_round`` trees.
 
-Here is a sample parameter dictionary for training a random forest on a GPU using
-xgboost::
+Here is a sample parameter dictionary for training a random forest on a GPU using XGBoost:
 
-  params = {
-    "colsample_bynode": 0.8,
-    "learning_rate": 1,
-    "max_depth": 5,
-    "num_parallel_tree": 100,
-    "objective": "binary:logistic",
-    "subsample": 0.8,
-    "tree_method": "hist",
-    "device": "cuda",
-  }
+.. tabs::
 
-A random forest model can then be trained as follows::
+    .. code-tab:: py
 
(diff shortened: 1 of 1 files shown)


trivialfis:
tabbed examples are required to be self-contained within a module (similar to jupyter notebook cells) for doc test

david-cortes:
Note sure what other might think but it feels like an overkill to add code tabs for what amounts to converting a python dict to an R list.

david-cortes:
Also it shouldn't close issue #11413 since this is just one out of many tutorials.

Ppp2048:
Yeah but given its one of my first few open Source contribution , didn't want to leave stones unturned.

trivialfis:
Out of curiosity, did you check the snippets in the PR?

david-cortes:
> Out of curiosity, did you check the snippets in the PR?

Checking the snippets now, and unrelated to this PR, but:
* Shouldn't something called "Random Forest" imply sampling **with replacement**? Does XGBoost support that?
* Example uses `"objective": "binary:logistic"`, but if that's the case, wouldn't it still end up finding splits based on gradients and hessians? Those would lead to finding different split points and leaf values than the criteria based on class frequencies from random forest libraries.
* Tipically, trees in random forest tend to be deep, and stop based on criteria like number of samples in leafs and possible impurity decreases instead.

PR #12276, opened 2026-06-30
[mt] GPU implementation of OHE vector leaf.

trivialfis:


RAMitchell (commented):
LGTM, but from codex:

[P1] Scan buffer is under-sized for high-cardinality categorical features.
 MakeSharedInputs passes param_.max_bin as n_bins_per_feat_tar, and EvaluateSplits() sizes scan_buffer_ as n_targets * Features() * n_bins_per_feat_tar * 2. But the new categorical scan writes using the global categorical bin_idx from feature_segments, e.g. region_others[bin_idx * n_targets + t]. Categorical cuts are not necessarily bounded by max_bin; existing tests explicitly cover n_categories = 32 with max_bin = 16 and expect categorical cut storage sized by categories, not max_bin.
Suggested fix: size the multi-target scan buffer by actual total bins per target, for example a host scalar derived from cuts_->TotalBins() / cut_values_.Size(), not Features() * max_bin. Add a regression test with multi_output_tree, categorical data, and n_categories > max_bin.
[P1] Non-OHE categorical features are silently evaluated as OHE.
 The new multi-target categorical branches always run one-hot scan/evaluation and return. That diverges from existing CPU and single-target GPU behavior, which check UseOneHot(...) and otherwise use partition-based categorical splits. The added Python coverage only forces OHE for multi-target, and the partition test path is skipped when multi_target=True.
Suggested fix: either keep categorical vector-leaf GPU unsupported unless every categorical feature satisfies UseOneHot, or implement the partition path for multi-target. Add a test with multi_strategy="multi_output_tree" and max_cat_to_onehot=1 to verify the intended behavior.
Positive notes
The OHE missing-direction algebra itself looks consistent: one region stores “other categories” with missing-right, the other stores the matching bin with missing-left, and the final weight assignment uses dir to reconstruct left/right sums. The new C++ test validates the dense no-missing OHE case and expected vector weights.

trivialfis:
1. Added a check for max bin
2. There's no way to distinguish at the moment. The only fix is to implement the partition-based split.

RAMitchell (approved):

PR #12280, opened 2026-07-01
Include license file in Python wheel metadata

### LINKED ISSUE #12256: Python wheels missing license file
Python wheels typically have a license file in their metadata section, but xgboost wheels do not have that:
Image

Compare against lightgbm for example:
Image

wgu9:
Fixes #12256.

This updates the Python package metadata to use the PEP 639 SPDX license expression and declare the top-level Apache-2.0 license file via `license-files`.

The current custom Hatchling build backend now copies the repository-level `LICENSE` into the temporary wheel/sdist workspace before delegating to Hatchling. The generated stub package path also copies the same file into the Python package build context and includes it in the stub wheel/sdist target.

I also checked this against the scikit-build-core backend shape from #12219: scikit-build-core rejects `license-files` when `project.license` is still the old `{ text = ... }` table, but accepts the SPDX string form used here and emits `License-Expression` plus `License-File` metadata.

Validation:
- `/opt/homebrew/bin/python3.12 -m py_compile ops/script/pypi_variants.py python-package/packager/pep517.py`
- `git diff --check`
- Built a temporary Hatchling wheel fixture and confirmed `License-Expression: Apache-2.0`, `License-File: LICENSE`, and `.dist-info/licenses/LICENSE` in the wheel.
- Built a temporary scikit-build-core wheel fixture and confirmed the same license metadata/file behavior.


chyunsu3:
I will review once #12219 is merged.

wgu9:
Thanks. #12219 has now landed, so I updated this branch on top of the scikit-build-core backend and adjusted the license metadata fix accordingly.

The previous CI failures were from two metadata validation errors in the new backend:
- `license-files = ["LICENSE"]` requires `LICENSE` to exist under `python-package/` at metadata generation time.
- PEP 639 SPDX `project.license` is not compatible with the old `License :: ...` classifier.

The follow-up now stages the repository-level `LICENSE` into `python-package/LICENSE` before wheel/sdist builds and removes the deprecated license classifier from the generated pyproject files.

wgu9:
Pushed a small follow-up for the system-install CI failure. The failing job runs `ops/script/pypi_variants.py` before `packaging` is installed, so I removed the top-level `packaging.version` dependency from that script and use the already-parsed project version string for the stub replacement.

Validation:
- `/opt/homebrew/bin/python3.12 ops/script/pypi_variants.py --use-suffix=na --require-nccl-dep=cu12`
- `/opt/homebrew/bin/python3.12 -S ops/script/pypi_variants.py --use-suffix=na --require-nccl-dep=cu12`
- `git diff --check`

chyunsu3 (approved):

